PR #4900, opened 2022-01-07
BL-10704 Undetected Disconnected state

gmartin7:
* on Checkin we now check connection just after checkin
   in case the Internet failed while checking in

---
This change is [Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/4900)

### DIFF
--- a/src/BloomExe/TeamCollection/TeamCollectionApi.cs
+++ b/src/BloomExe/TeamCollection/TeamCollectionApi.cs
@@ -513,8 +513,6 @@ namespace Bloom.TeamCollection
 					BookHistory.AddEvent(_bookSelection.CurrentSelection, BookHistoryEventType.CheckIn, message);
 					BookHistory.SetPendingCheckinMessage(_bookSelection.CurrentSelection, "");
 
-					_tcManager.CurrentCollection.PutBook(_bookSelection.CurrentSelection.FolderPath, checkin:true);
-
 					Analytics.Track("TeamCollectionCheckinBook",
 						new Dictionary<string, string>(){
 							{"CollectionId", _settings?.CollectionId},
@@ -548,9 +546,10 @@ namespace Bloom.TeamCollection
 							{"BookName", _bookSelection?.CurrentSelection?.Title}
 						});
 				}
-
 				UpdateUiForBook();
 				request.PostSucceeded();
+
+				Application.Idle += OnIdleConnectionCheck;
 			}
 			catch (Exception e)
 			{
@@ -567,6 +566,18 @@ namespace Bloom.TeamCollection
 				NonFatalProblem.ReportSentryOnly(e, $"Something went wrong for {request.LocalPath()} ({_bookSelection?.CurrentSelection?.FolderPath})");
 				request.Failed("checkin failed");
 			}
+		}
+
+		private void OnIdleConnectionCheck(object sender, EventArgs e)
+		{
+			Application.Idle -= OnIdleConnectionCheck;
+
+			// BL-10704: In case the Internet went away while we were trying to CheckIn a book...
+			// This will at least signal to the user in the Dropbox case, that while his checkin
+			// may have succeeded, his colleagues won't know about it until the Internet is up again.
+			// If we don't do it "OnIdle", the book status pane doesn't reflect that we actually did
+			// (probably, assuming we are on Dropbox, anyway) complete the checkin.
+			_tcManager.CheckConnection();
 		}
 
 		// Tell the CollectionSettingsDialog that we should reopen the collection now


JohnThomson (changes requested):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/4900)* status: all files reviewed, 1 unresolved discussion (waiting on @gmartin7)

___
*[src/BloomExe/TeamCollection/TeamCollectionApi.cs, line 557 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/4900#-Msq5TkoBYXkrp_kocPn:-Msq5Tkp8mNAOP_-1i1s:b-vv7ewn) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/6d169e470e82c136a2caff80b858f02d1af5301a/src/BloomExe/TeamCollection/TeamCollectionApi.cs#L557)):*
> ```Smalltalk
> 					request.Failed();
> 					return;
> 				}
> ```

You should check to see what happens if you simply change the condition in the if to 'false'. But I'm pretty sure this is not anywhere near sufficient to deal with this problem.
(a) You probably should deal with an interruption of a LAN connection during checkin as well as an internet failure (Dropbox).
(b) Very likely a LAN connection failure in the middle of the checkin will cause an exception to be thrown. That would need to be handled appropriately.
(c) In the case of a Dropbox connection, do we even care if the internet failed? Changes should be successfully written both to the local book folder and to the local copy of the Dropbox folder. Dropbox will finish the sync when the connection resumes. What different behavior do you want from the checkin button as a result of saying the request failed? If none, why make it fail?
(d) You don't seem to be adding anything at the JavaScript end to respond to the request failing. Is it already doing what you want? That would surprise me...mostly the UI responds to websocket broadcasts when status needs to change.
(e) You don't seem to be doing anything to make sure that the state of book as locally recorded or as shown in the UI reflects the failed checkin, if in fact it did fail.
(f) You don't seem to be doing anything to make the UI show that the collection is disconnected.
(g) In the case of a LAN connection failure, we might want to do various things, depending on exactly where in the process it failed. I think there are four main possibilities. It might already be down when we start, or go down while we are doing preliminary checks, or go down while we are writing the zipped book to a temp file, or just possibly go down while we are trying to replace the old zip file in the LAN folder with the new one. The result is presumably either (i) we changed nothing in the LAN folder, or just created a temp file which is harmless (ii) we successfully updated the LAN folder copy of the book to the checked-in state before the failure, or, just possibly, (iii) we left that LAN folder in a state where the book in question is missing, because we deleted the old one but failed to rename the temp file to take its place. Case (ii) doesn't require doing anything special; this checkin has succeeded (though we might want to display UI indicating that we're disconnected). In case (i), we probably want to give an error message and avoid updating state in a way that would indicate the book is checked in. I'm not sure what we can usefully do in case (iii) but it should be very rare.

gmartin7 (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/4900)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @JohnThomson)

___
*[src/BloomExe/TeamCollection/TeamCollectionApi.cs, line 557 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/4900#-Msq5TkoBYXkrp_kocPn:-MsqZ0WR3rGcDHJDGfNO:bul634g) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/6d169e470e82c136a2caff80b858f02d1af5301a/src/BloomExe/TeamCollection/TeamCollectionApi.cs#L557)):*

You should check to see what happens if you simply change the condition in the if to 'false'. But I'm pretty sure this is not anywhere near sufficient to deal with this problem.
(a) You probably should deal with an interruption of a LAN connection during checkin as well as an internet failure (Dropbox).
(b) Very likely a LAN connection failure in the middle of the checkin will cause an exception to be thrown. That would need to be handled appropriately.
(c) In the case of a Dropbox connection, do we even care if the internet failed? Changes should be successfully written both to the local book folder and to the local copy of the Dropbox folder. Dropbox will finish the sync when the connection resumes. What different behavior do you want from the checkin button as a result of saying the request failed? If none, why make it fail?
(d) You don't seem to be adding anything at the JavaScript end to respond to the request failing. Is it already doing what you want? That would surprise me...mostly the UI responds to websocket broadcasts when status needs to change.
(e) You don't seem to be doing anything to make sure that the state of book as locally recorded or as shown in the UI reflects the failed checkin, if in fact it did fail.
(f) You don't seem to be doing anything to make the UI show that the collection is disconnected.
(g) In the case of a LAN connection failure, we might want to do various things, depending on exactly where in the process it failed. I think there are four main possibilities. It might already be down when we start, or go down while we are doing preliminary checks, or go down while we are writing the zipped book to a temp file, or just possibly go down while we are trying to replace the old zip file in the LAN folder with the new one. The result is presumably either (i) we changed nothing in the LAN folder, or just created a temp file which is harmless (ii) we successfully updated the LAN folder copy of the book to the checked-in state before the failure, or, just possibly, (iii) we left that LAN folder in a state where the book in question is missing, because we deleted the old one but failed to rename the temp file to take its place. Case (ii) doesn't require doing anything special; this checkin has succeeded (though we might want to display UI indicating that we're disconnected). In case (i), we probably want to give an error message and avoid updating state in a way that would indicate the book is checked in. I'm not sure what we can usefully do in case (iii) but it should be very rare.
</blockquote></details>

Done. Sorta. As per our discussion w/Colin.

gmartin7 (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/4900)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @JohnThomson)

___
*[src/BloomExe/TeamCollection/TeamCollectionApi.cs, line 580 at r3](https://reviewable.io/reviews/bloombooks/bloomdesktop/4900#-MsqhzLt6OrwsMtgvn30:-MsqhzLt6OrwsMtgvn31:b8xx7ma) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/e229fd26421ee3c31e221273b715721399091772/src/BloomExe/TeamCollection/TeamCollectionApi.cs#L580)):*
> ```Smalltalk
> 			// If we don't do it "OnIdle", the book status pane doesn't reflect that we actually did
> 			// (probably, assuming we are on Dropbox, anyway) complete the checkin.
> 			_tcManager.CheckConnection();
> ```

The OnIdle solution succeeds in allowing the checkin to update the Book status panel UI BEFORE telling the user that they are in a Disconnected state... which is okay for this one Dropbox situation.

JohnThomson (approved):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/4900)* status: all files reviewed, 1 unresolved discussion (waiting on @JohnThomson)

PR #4901, opened 2022-01-07
BL-9953 Fix origami rule precedence

gmartin7:
This change is [Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/4901)

### DIFF
--- a/src/BloomBrowserUI/bookEdit/css/origami.less
+++ b/src/BloomBrowserUI/bookEdit/css/origami.less
@@ -50,7 +50,10 @@
     border-bottom: 3px dashed transparent;
 }
 
-.origami-layout-mode .bloom-editable {
+// BL-9953 Added 'p' so a competing audioRecording rule won't have more precedence. The competing
+// rule also "targets" the p element and all the text whose color we are concerned about is inside
+// of p elements anyway.
+.origami-layout-mode .bloom-editable p {
     // We want to see some text in the background, but it can't interfer with the forground.
     // If you change this color, be sure to see how it looks when grey background is enabled.
     color: #ececec !important;


JohnThomson (changes requested):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/4901)* status: all files reviewed, 1 unresolved discussion (waiting on @gmartin7)

___
*[src/BloomBrowserUI/bookEdit/css/origami.less, line 53 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/4901#-MsqZubpErxfJUaqsANE:-MsqZubpErxfJUaqsANF:b-asedep) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/80dd3c5696d1e33eb2e3f5d2bad071f3c2e8c80d/src/BloomBrowserUI/bookEdit/css/origami.less#L53)):*
> ```Less
> }
> 
> // BL-9953 Added 'p' so audioRecording rule won't have more precedence.
> ```

But this also makes it mean something different. That might be OK, but if you just want to increase precedence, another option is 

.origami-layout-mode.origami-layout-mode .bloom-editable

This will match exactly the same cases as before, but with one notch more precedence. (Actually I think it's a stronger increase in precedence, since classes are more important that elements.)

If you're sure you only want the paragraph colored, not the whole bloom-editable, I guess the original can stand.

gmartin7 (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/4901)* status: all files reviewed, 1 unresolved discussion (waiting on @JohnThomson)

___
*[src/BloomBrowserUI/bookEdit/css/origami.less, line 53 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/4901#-MsqZubpErxfJUaqsANE:-MsqaE-i3lK96NxSR1VC:bty5b0z) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/80dd3c5696d1e33eb2e3f5d2bad071f3c2e8c80d/src/BloomBrowserUI/bookEdit/css/origami.less#L53)):*

But this also makes it mean something different. That might be OK, but if you just want to increase precedence, another option is 

.origami-layout-mode.origami-layout-mode .bloom-editable

This will match exactly the same cases as before, but with one notch more precedence. (Actually I think it's a stronger increase in precedence, since classes are more important that elements.)

If you're sure you only want the paragraph colored, not the whole bloom-editable, I guess the original can stand.
</blockquote></details>

Yeah, I thought about the fact that I was changing the "target" of the rule, but it's the text in the `.bloom-editable` we're after here anyway (this rule grays it out, if the origami function is active) and the competing rule also "targets" the `p` element, so I figured this was best.

gmartin7 (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/4901)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @JohnThomson)

___
*[src/BloomBrowserUI/bookEdit/css/origami.less, line 53 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/4901#-MsqZubpErxfJUaqsANE:-MsqavdbB4QTb1YFU5i2:bv80yhn) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/80dd3c5696d1e33eb2e3f5d2bad071f3c2e8c80d/src/BloomBrowserUI/bookEdit/css/origami.less#L53)):*

Yeah, I thought about the fact that I was changing the "target" of the rule, but it's the text in the `.bloom-editable` we're after here anyway (this rule grays it out, if the origami function is active) and the competing rule also "targets" the `p` element, so I figured this was best.
</blockquote></details>

I've added something to the comment too.

JohnThomson (approved):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/4901)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @gmartin7)

PR #4918, opened 2022-01-17
BL-10593 Fix origami divider so it works when page is scaled

gmartin7:
* 2 commits; 1st is just a Prettier update, 2nd is actual change

---
This change is [Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/4918)

### DIFF
--- a/src/BloomBrowserUI/lib/split-pane/split-pane.js
+++ b/src/BloomBrowserUI/lib/split-pane/split-pane.js
@@ -8,6 +8,7 @@ Released under the MIT license
 https://raw.github.com/shagstrom/split-pane/master/LICENSE
 
 */
+
 (function($) {
 
     $.fn.splitPane = function() {
@@ -159,6 +160,9 @@ https://raw.github.com/shagstrom/split-pane/master/LICENSE
     }
 
     function createMousemove($splitPane, pageX, pageY) {
+        var pageScale = $splitPane[0].getBoundingClientRect().width / $splitPane[0].offsetWidth;
+        var scaledX = pageX / pageScale;
+        var scaledY = pageY / pageScale;
         var splitPane = $splitPane[0],
             firstComponent = $splitPane.children('.split-pane-component:first')[0],
             divider = $splitPane.children('.split-pane-divider')[0],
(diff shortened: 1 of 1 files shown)


JohnThomson (changes requested):
Is it feasible to back out the prettier changes? IIRC, this is a library we copied and slightly modified. I think it would be best to keep it in a state where the only changes are the minimal ones we need.

> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/4918#-:-Mty1Mc43a9SiKHjdOu2:bne9uw4)* status: all files reviewed, 1 unresolved discussion (waiting on @gmartin7)

___
*[src/BloomBrowserUI/lib/split-pane/split-pane.js, line 271 at r2](https://reviewable.io/reviews/bloombooks/bloomdesktop/4918#-Mty2Yw65NB9FFxmI5NA:-Mty2Yw65NB9FFxmI5NB:b-p2ouwp) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/834695c7160b7c550408813550bc77bfc7e26bd1/src/BloomBrowserUI/lib/split-pane/split-pane.js#L271)):*
> ```JavaScript
> 
>     function createMousemove($splitPane, pageX, pageY) {
>         var pageScale = EditableDivUtils.getPageScale();
> ```

I hate that we are taking a general-purpose library and adding a patch that is specific to a particular context in Bloom.
One alternative is to just set scale to splitPane[0].getBoundingClientRect().width / splitPane[0].offsetWidth. This is, I think, likely to be close enough to fix the problem, though worth testing with small split panes and large scaling factors.
If that's not good enough, you could give this module an exported variable getScale: HTMLElement=>number = (elt) => elt.getBoundingClientRect().width / elt.offsetWidth, use that to get the scale (pass it splitPane[0]), and somewhere in one of our own files configure it to be a function that returns EditableDivUtils.getPageScale();.

gmartin7 (commented):
Trickier than it looks. With Andrew's help I was able to temporarily defeat Git's pre-commit hook which called Prettier. Now to go on to your other question.

> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/4918#-:-MuCKMQW4FsJOt9uoV5i:brdwgnx)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @JohnThomson)

___
*[src/BloomBrowserUI/lib/split-pane/split-pane.js, line 271 at r2](https://reviewable.io/reviews/bloombooks/bloomdesktop/4918#-Mty2Yw65NB9FFxmI5NA:-MuCddUV332s3prIW0Gz:b-3f5p8w) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/834695c7160b7c550408813550bc77bfc7e26bd1/src/BloomBrowserUI/lib/split-pane/split-pane.js#L271)):*

I hate that we are taking a general-purpose library and adding a patch that is specific to a particular context in Bloom.
One alternative is to just set scale to splitPane[0].getBoundingClientRect().width / splitPane[0].offsetWidth. This is, I think, likely to be close enough to fix the problem, though worth testing with small split panes and large scaling factors.
If that's not good enough, you could give this module an exported variable getScale: HTMLElement=>number = (elt) => elt.getBoundingClientRect().width / elt.offsetWidth, use that to get the scale (pass it splitPane[0]), and somewhere in one of our own files configure it to be a function that returns EditableDivUtils.getPageScale();.
</blockquote></details>

1st alternative worked great; thanks!

JohnThomson (approved):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/4918)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @gmartin7)

PR #4931, opened 2022-01-27
BL-10731 Allow Progress to get to socket server on its own

gmartin7:
* removed some Chorus stuff
* added a couple of comments to Program.cs
* setting up singleton
* add static websocket server and port to
   ApplicationContainer
* setup App api so that WebSocketManager can get the
   port string to initialize itself
* WebSocketManager uses async/await to get the port
   string and make sure it's initialized before the
   WebSocketManager gets used
* BloomWebSocketServer defaults to port 8093
* update BrowserProgressDialog to have direct server
   access
* refactoring, especially SpreadsheetApi and attendant
   classes
* got tests to pass

---
This change is [Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/4931)


gmartin7:
We decided not to do this at this time.

PR #4944, opened 2022-02-01
BL-10884 Fix SS Importer to update side classes

gmartin7:
* also removed some usings

---
This change is [Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/4944)

### DIFF
--- a/src/BloomExe/Book/BookStorage.cs
+++ b/src/BloomExe/Book/BookStorage.cs
@@ -2822,7 +2822,7 @@ namespace Bloom.Book
 		/// Save a copy of the specified book in the a folder %temp%/bloom pre-import backups.
 		/// Generate a unique name for the backup as necessary, and return its full path.
 		/// </summary>
-		public static string SaveCopyBeforeImportOverwrite(string bookFolder, string bookFilePath)
+		public static string SaveCopyBeforeImportOverwrite(string bookFolder)
 		{
 			string origFileName = Path.GetFileName(bookFolder);
 			var parentFolder = Path.Combine(Path.GetTempPath(), "bloom pre-import backups");
--- a/src/BloomExe/Program.cs
+++ b/src/BloomExe/Program.cs
@@ -153,6 +153,9 @@ namespace Bloom
 						(ChangeLayoutParameters opts) => ChangeLayoutCommand.Handle(opts),
 						(CreateArtifactsParameters opts) => CreateArtifactsCommand.Handle(opts),
 						(SpreadsheetExportParameters opts) => SpreadsheetExportCommand.Handle(opts),
+						// We don't have a way to get the CollectionSettings object for the Import process.
+						// This means that if we use this CLI version, care should be taken to update the book,
+						// so the pages get the correct "side" classes (side-left, side-right). (BL-10884)
 						(SpreadsheetImportParameters opts) => SpreadsheetImportCommand.Handle(opts),
 						errors =>
 						{
--- a/src/BloomExe/Spreadsheet/SpreadsheetApi.cs
+++ b/src/BloomExe/Spreadsheet/SpreadsheetApi.cs
@@ -91,7 +91,10 @@ namespace Bloom.web.controllers
 			request.PostSucceeded();
 		}
 
-
+		/// <summary>
+		/// This is our primary entry point for importing from a spreadsheet. There is also a CLI command,
+		/// but we shouldn't need that one much, now that we have this on our book thumb context menus.
+		/// </summary>
 		public void HandleImportContentFromSpreadsheet(Book.Book book)
 		{
 			if (!_libraryModel.CollectionSettings.HaveEnterpriseFeatures)
@@ -118,13 +121,12 @@ namespace Bloom.web.controllers
 					SetSpreadsheetFolder(book, spreadsheetFolder);
 				}
 
-				var dom = book.OurHtmlDom;
-				var importer = new SpreadsheetImporter(_webSocketServer, dom, Path.GetDirectoryName(inputFilepath), book.FolderPath);
-				importer.ImportWithProgress(inputFilepath, bookPath);
+				var importer = new SpreadsheetImporter(_webSocketServer, book, Path.GetDirectoryName(inputFilepath));
+				importer.ImportWithProgress(inputFilepath);
 
 				// Review: A lot of other stuff happens in Book.Save() and BookStorage.SaveHtml().
 				// I doubt we need any of it for current purposes, but later we might.
-				XmlHtmlConverter.SaveDOMAsHtml5(dom.RawDom, bookPath);
+				XmlHtmlConverter.SaveDOMAsHtml5(book.OurHtmlDom.RawDom, bookPath);
 				book.ReloadFromDisk(null);
 				_bookSelection.InvokeSelectionChanged(false);
 			}
--- a/src/BloomExe/Spreadsheet/SpreadsheetExporter.cs
+++ b/src/BloomExe/Spreadsheet/SpreadsheetExporter.cs
@@ -12,7 +12,6 @@ using System.Diagnostics;
 using System.Drawing;
 using System.IO;
 using System.Linq;
-using System.Reflection.Metadata.Ecma335;
 using System.Security;
 using System.Windows.Forms;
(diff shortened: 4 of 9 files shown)


hatton (changes requested):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/4944)* status: 0 of 4 files reviewed, 1 unresolved discussion (waiting on @gmartin7 and @hatton)

___
*[src/BloomExe/Spreadsheet/SpreadsheetImporter.cs, line 197 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/4944#-Muqo9Qo8V6ouHUJ0kjW:-Muqo9Qo8V6ouHUJ0kjX:b-4rvdjr) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/93c84c887c6ab1a38be7e056161e8f3666386cc2/src/BloomExe/Spreadsheet/SpreadsheetImporter.cs#L197)):*
> ```Smalltalk
> 		private CollectionSettings GetCollectionSettings()
> 		{
> 			var settingsFilePath = CollectionSettings.FindSettingsFileInFolder(
> ```

It seems unusual for a place wanting access to the current collection settings to get them by looking around at files and then creating a new one. Is that how we normally do it?

gmartin7 (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/4944)* status: 0 of 4 files reviewed, 1 unresolved discussion (waiting on @hatton)

___
*[src/BloomExe/Spreadsheet/SpreadsheetImporter.cs, line 197 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/4944#-Muqo9Qo8V6ouHUJ0kjW:-Muqzwke6ws2o7iFtgWZ:b-b9cujp) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/93c84c887c6ab1a38be7e056161e8f3666386cc2/src/BloomExe/Spreadsheet/SpreadsheetImporter.cs#L197)):*

It seems unusual for a place wanting access to the current collection settings to get them by looking around at files and then creating a new one. Is that how we normally do it?
</blockquote></details>

It is pretty unusual (the above method until now was only used in collection renaming), but usually we have the book or at least a bookInfo that we're dealing with, which we don't seem to in this case. OTOH, one of the classes that creates the `SpreadsheetImporter` already has the collection settings. We could modifying the `ctor` to pass in what we need, but that causes a mess in tests and CLI commands that don't have the collection readily available.

hatton (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/4944)* status: 0 of 4 files reviewed, all discussions resolved (waiting on @hatton)

___
*[src/BloomExe/Spreadsheet/SpreadsheetImporter.cs, line 197 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/4944#-Muqo9Qo8V6ouHUJ0kjW:-MurID_PAGXK_X3LRXG1:b9ttx2e) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/6571ad721ef5d4afff11e2870020445fff2637cc/src/BloomExe/Spreadsheet/SpreadsheetImporter.cs#L197)):*

It is pretty unusual (the above method until now was only used in collection renaming), but usually we have the book or at least a bookInfo that we're dealing with, which we don't seem to in this case. OTOH, one of the classes that creates the `SpreadsheetImporter` already has the collection settings. We could modifying the `ctor` to pass in what we need, but that causes a mess in tests and CLI commands that don't have the collection readily available.
</blockquote></details>

OK, those don't seem like justifications. Sorry if I'm missing it.

JohnThomson (changes requested):
The fix looks good, but we've worked hard to make SS rather comprehensively unit tested. Can you come up with a test that would fail without this? The SpreadsheetImageAndTextImportTests does a fair bit of inserting pages, so it's likely that the DOM it produces has wrong page numbering and side classes without this fix.

> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/4944#-:-MurcrSo2wdcQQflcCCq:b-cstiaj)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @gmartin7)

hatton (changes requested):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/4944)* status: 3 of 8 files reviewed, 2 unresolved discussions (waiting on @gmartin7, @hatton, and @JohnThomson)

___
*[src/BloomExe/Spreadsheet/SpreadsheetApi.cs, line 122 at r2](https://reviewable.io/reviews/bloombooks/bloomdesktop/4944#-Muwj4NwCrfSz3ulIQLp:-Muwj4NwCrfSz3ulIQLq:bz9u8gf) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/70aa6ec7c9fce6732ec2dc757d1a76e29eb8069e/src/BloomExe/Spreadsheet/SpreadsheetApi.cs#L122)):*
> ```Smalltalk
> 
> 				var dom = book.OurHtmlDom;
> 				var importer = new SpreadsheetImporter(_webSocketServer, dom,
> ```

Would there be a problem with simply giving the constructor the book? Why break out the properties of it?
___
*[src/BloomExe/Spreadsheet/SpreadsheetApi.cs, line 124 at r2](https://reviewable.io/reviews/bloombooks/bloomdesktop/4944#-MuwjKjo6uRfsGiV_e0A:-MuwjKjo6uRfsGiV_e0B:bftb2d8) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/70aa6ec7c9fce6732ec2dc757d1a76e29eb8069e/src/BloomExe/Spreadsheet/SpreadsheetApi.cs#L124)):*
> ```Smalltalk
> 				var importer = new SpreadsheetImporter(_webSocketServer, dom,
> 					Path.GetDirectoryName(inputFilepath), book.FolderPath, book.CollectionSettings);
> 				importer.ImportWithProgress(inputFilepath, bookPath);
> ```

If the importer already has the path (or the target book, as suggested above), does it need to be given the `bookPath` again? I would remove that, unless I'm misunderstanding something.

JohnThomson (approved):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/4944)* status: all files reviewed, 2 unresolved discussions (waiting on @gmartin7)

gmartin7 (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/4944)* status: 4 of 9 files reviewed, 2 unresolved discussions (waiting on @hatton and @JohnThomson)

___
*[src/BloomExe/Spreadsheet/SpreadsheetApi.cs, line 122 at r2](https://reviewable.io/reviews/bloombooks/bloomdesktop/4944#-Muwj4NwCrfSz3ulIQLp:-Mv513qd-Mw_eCPpHMLt:b-896fix) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/70aa6ec7c9fce6732ec2dc757d1a76e29eb8069e/src/BloomExe/Spreadsheet/SpreadsheetApi.cs#L122)):*

Would there be a problem with simply giving the constructor the book? Why break out the properties of it?
</blockquote></details>

Done.
___
*[src/BloomExe/Spreadsheet/SpreadsheetApi.cs, line 124 at r2](https://reviewable.io/reviews/bloombooks/bloomdesktop/4944#-MuwjKjo6uRfsGiV_e0A:-Mv515SG4GAaJn5xtuD-:b-896fix) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/70aa6ec7c9fce6732ec2dc757d1a76e29eb8069e/src/BloomExe/Spreadsheet/SpreadsheetApi.cs#L124)):*

If the importer already has the path (or the target book, as suggested above), does it need to be given the `bookPath` again? I would remove that, unless I'm misunderstanding something. 
</blockquote></details>

Done.

JohnThomson (approved):
> Reviewed 4 of 4 files at r3, 1 of 1 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/4944)* status: all files reviewed, 2 unresolved discussions (waiting on @hatton)

PR #4945, opened 2022-02-01
Fix renaming books as well as changing titles (BL-10881)

StephenMcConnel:
This change is [Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/4945)

### DIFF
--- a/src/BloomExe/Book/BookData.cs
+++ b/src/BloomExe/Book/BookData.cs
@@ -1716,6 +1716,10 @@ namespace Bloom.Book
 					info.AllTitles = sb.ToString();
 				}
 			}
+			else
+			{
+				_dom.Title = "";
+			}
 
 			if (BookIsDerivative())
 			{
--- a/src/BloomExe/CollectionTab/LibraryListView.cs
+++ b/src/BloomExe/CollectionTab/LibraryListView.cs
@@ -1079,7 +1079,19 @@ namespace Bloom.CollectionTab
 		{
 			get
 			{
-				return AllBookButtons().FirstOrDefault(b => GetBookInfoFromButton(b).FolderPath == SelectedBook?.BookInfo.FolderPath);
+				if (SelectedBook == null)
+					return null;
+				var allButtons = AllBookButtons();
+				// Search for book button by FolderPath because that is guaranteed to be unique.
+				var btnFound = allButtons.FirstOrDefault(b => GetBookInfoFromButton(b).FolderPath == SelectedBook.BookInfo.FolderPath);
+				if (btnFound == null)
+				{
+					// If search fails, try searching by Id, which should be unique but isn't guaranteed since users can copy folders on disk.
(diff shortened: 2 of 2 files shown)


JohnThomson (approved):
What you have is OK, but I think one change might make it even better

> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/4945#-:-Murc7OwDqzm15gjw2do:b-h1a7he)* status: all files reviewed, 1 unresolved discussion (waiting on @StephenMcConnel)

___
*[src/BloomExe/CollectionTab/LibraryListView.cs, line 1107 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/4945#-MurbLlKBdlRb-_9NJBF:-MurbLlKBdlRb-_9NJBG:b-ymfddd) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/ce513990f214142da4654a82da2984186975d13d/src/BloomExe/CollectionTab/LibraryListView.cs#L1107)):*
> ```Smalltalk
> 			{
> 				// Title must have changed, which has changed the FolderPath.  Try to find the button by book Id.
> 				var buttons = AllBookButtons().Where(b => GetBookInfoFromButton(b).Id == SelectedBook?.BookInfo.Id).Select(b => b).ToList();
> ```

How certain is it that this will be called before anything else calls SelectedButton? What would you think of putting most of the logic into SelectedButton, so that if it finds no match on FolderName, it will look by ID, and (assuming it finds exactly one matching book) just return that?
It would make sense that the logic for FIXING the FolderName would stay here.

StephenMcConnel (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/4945)* status: 1 of 2 files reviewed, 1 unresolved discussion (waiting on @JohnThomson)

___
*[src/BloomExe/CollectionTab/LibraryListView.cs, line 1107 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/4945#-MurbLlKBdlRb-_9NJBF:-MuvSd2Z4I9vNXOo2Rba:b-896fix) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/ce513990f214142da4654a82da2984186975d13d/src/BloomExe/CollectionTab/LibraryListView.cs#L1107)):*

How certain is it that this will be called before anything else calls SelectedButton? What would you think of putting most of the logic into SelectedButton, so that if it finds no match on FolderName, it will look by ID, and (assuming it finds exactly one matching book) just return that?
It would make sense that the logic for FIXING the FolderName would stay here.
</blockquote></details>

Done.

JohnThomson (approved):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/4945)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @StephenMcConnel)

PR #4949, opened 2022-02-04
BL-10891 Get xmatter import working better

gmartin7:
* cover image now gets copied
* after importing we bring the xmatter up-to-date and
   save the book

---
This change is [Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/4949)

### DIFF
--- a/src/BloomExe/Spreadsheet/SpreadsheetApi.cs
+++ b/src/BloomExe/Spreadsheet/SpreadsheetApi.cs
@@ -124,9 +124,10 @@ namespace Bloom.web.controllers
 				var importer = new SpreadsheetImporter(_webSocketServer, book, Path.GetDirectoryName(inputFilepath));
 				importer.ImportWithProgress(inputFilepath);
 
-				// Review: A lot of other stuff happens in Book.Save() and BookStorage.SaveHtml().
-				// I doubt we need any of it for current purposes, but later we might.
-				XmlHtmlConverter.SaveDOMAsHtml5(book.OurHtmlDom.RawDom, bookPath);
+				// The importer now does BringBookUpToDate() which accomplishes everything this did,
+				// plus it may have actually changed the folder (and subsequent 'bookPath')
+				// due to a newly imported title. That would cause this call to fail.
+				//XmlHtmlConverter.SaveDOMAsHtml5(book.OurHtmlDom.RawDom, bookPath);
 				book.ReloadFromDisk(null);
 				_bookSelection.InvokeSelectionChanged(false);
 			}
--- a/src/BloomExe/Spreadsheet/SpreadsheetImporter.cs
+++ b/src/BloomExe/Spreadsheet/SpreadsheetImporter.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using Bloom.Book;
 using System.Collections.Generic;
 using System.Diagnostics;
@@ -22,6 +22,7 @@ namespace Bloom.Spreadsheet
 	public class SpreadsheetImporter
 	{
 		private HtmlDom _destinationDom;
+		private readonly Book.Book _book;
 		private InternalSpreadsheet _sheet;
 		private int _currentRowIndex;
 		private int _currentPageIndex;
@@ -60,9 +61,9 @@ namespace Bloom.Spreadsheet
 			_destinationDom = destinationDom;
 			_dataDivElement = _destinationDom.SafeSelectNodes("//div[@id='bloomDataDiv']").Cast<XmlElement>().First();
 			_pathToBookFolder = pathToBookFolder;
+			// Tests and CLI may not set one or more of these
 			_pathToSpreadsheetFolder = pathToSpreadsheetFolder;
 			_webSocketServer = webSocketServer;
-			// Tests and CLI may not set this
 			_collectionSettings = collectionSettings;
 		}
 
@@ -73,6 +74,7 @@ namespace Bloom.Spreadsheet
 		public SpreadsheetImporter(IBloomWebSocketServer webSocketServer, Book.Book book, string pathToSpreadsheetFolder)
 			: this(webSocketServer, book.OurHtmlDom, pathToSpreadsheetFolder, book.FolderPath, book.CollectionSettings)
 		{
+			_book = book;
 		}
 
 		/// <summary>
@@ -172,7 +174,6 @@ namespace Bloom.Spreadsheet
 			_destLayout = Layout.FromDom(_destinationDom, Layout.A5Portrait);
 			while (_currentRowIndex < _inputRows.Count)
 			{
-
 				var currentRow = _inputRows[_currentRowIndex];
 				string rowTypeLabel = currentRow.MetadataKey;
 
@@ -205,10 +206,15 @@ namespace Bloom.Spreadsheet
 				}
 				_currentRowIndex++;
 			}
-			if (_collectionSettings != null)
-				_destinationDom.UpdatePageNumberAndSideClassOfPages(
-					_collectionSettings.CharactersForDigitsForPageNumbers,
-					_collectionSettings.Language1.IsRightToLeft);
+			// This section is necessary to make sure changes to the dom are recorded.
+			// If we run SS Importer from the CLI (without CollectionSettings), BringBookUpToDate()
+			// will happen when we eventually open the book, but the user gets an updated thumbail and preview
+			// if we do it here for the main production case where we DO have both the CollectionSettings
+			// and the Book itself. Testing is the other situation (mostly) that doesn't use CollectionSettings.
+			if (_collectionSettings != null && _book != null)
+			{
+				_book.BringBookUpToDate(new NullProgress());
+			}
 
 			Progress("Done");
 			return _warnings;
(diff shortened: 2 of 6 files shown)


JohnThomson (changes requested):
Need unit tests that fail without these fixes, please. (In the case of the xmatter one, that currently means you'll need to pass a collection settings in the test so BBUTD runs).
Another option is to say that it's not the SSImporter's job to BBUTD...that should be done by the caller if there isn't some other way it will get done.

> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/4949#-:-Mv67N2S0nADH7b1NFUm:b1pwrtw)* status: all files reviewed, 1 unresolved discussion (waiting on @gmartin7)

___
*[src/BloomExe/Spreadsheet/SpreadsheetImporter.cs, line 209 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/4949#-Mv662tqD_uVgnQTvLI8:-Mv662tqD_uVgnQTvLI9:b-v6k9ut) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/529c20b135ad13503a29a89202f7f82e34623fa5/src/BloomExe/Spreadsheet/SpreadsheetImporter.cs#L209)):*
> ```Smalltalk
> 					_collectionSettings.CharactersForDigitsForPageNumbers,
> 					_collectionSettings.Language1.IsRightToLeft);
> 				_book.BringXmatterHtmlUpToDate(_dest);
> ```

Is this helpful to do even if we don't have collection settings? I suppose if we're running from a command line, BBUTD will happen when we eventually open the book. And doing it unconditionally would probably break unit tests right and left by inserting extra xmatter pages. I think those reasons for not doing it should be commented.

gmartin7 (commented):
I put quite a bit of effort into trying to create a unit test that actually had a Book object. Not easy. I've managed to test that the data div gets updated for cover image and title.

> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/4949#-:-MvQUiMlAx8QnGAJ9CFX:bifsnq6)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @JohnThomson)

___
*[src/BloomExe/Spreadsheet/SpreadsheetImporter.cs, line 209 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/4949#-Mv662tqD_uVgnQTvLI8:-MvKZ6ne8NDStVXeJaTO:b-f4j49k) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/529c20b135ad13503a29a89202f7f82e34623fa5/src/BloomExe/Spreadsheet/SpreadsheetImporter.cs#L209)):*

Is this helpful to do even if we don't have collection settings? I suppose if we're running from a command line, BBUTD will happen when we eventually open the book. And doing it unconditionally would probably break unit tests right and left by inserting extra xmatter pages. I think those reasons for not doing it should be commented.
</blockquote></details>

You were absolutely right. Making the xmatter update and save unconditional breaks 90 tests! And both methods are part of BBUTD. Commented.

JohnThomson (changes requested):
> Reviewed 5 of 5 files at r3.
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/4949)* status: 5 of 6 files reviewed, 3 unresolved discussions (waiting on @gmartin7 and @JohnThomson)

___
*[src/BloomExe/Spreadsheet/SpreadsheetImporter.cs, line 212 at r3](https://reviewable.io/reviews/bloombooks/bloomdesktop/4949#-MvVeWb602IyEEpGF1od:-MvVeWb602IyEEpGF1oe:bocf0ri) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/2b8154985bbf13966b841fff8376cd8e310a477f/src/BloomExe/Spreadsheet/SpreadsheetImporter.cs#L212)):*
> ```Smalltalk
> 			// If we run SS Importer from the CLI (without CollectionSettings), BringBookUpToDate()
> 			// will happen when we eventually open the book, but the user gets a better experience
> 			// if we do it here for the main production case where we DO have both the CollectionSettings
> ```

While you're improving comments...might be useful to say what bad user experience you're guarding against. Then if someone considers taking it out they know what to look for.
___
*[src/BloomTests/Spreadsheet/SpreadsheetImporterTests.cs, line 827 at r2](https://reviewable.io/reviews/bloombooks/bloomdesktop/4949#-MvVdKN-BYlsr4Chj-6x:-MvVdKN-BYlsr4Chj-6y:b-ceo4m5) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/650cebc378a826a680c803cb8058dd8a5b5e0d6e/src/BloomTests/Spreadsheet/SpreadsheetImporterTests.cs#L827)):*
> ```Smalltalk
> 			var unexpectedClass = expectedClass == "side-right" ? "side-left" : "side-right";
> 			AssertThatXmlIn.Element(_contentPages[n]).HasNoMatchForXpath($"self::div[contains(@class, '" + unexpectedClass + "')]");
> 			AssertThatXmlIn.Element(_contentPages[n]).HasSpecifiedNumberOfMatchesForXpath($"self::div[contains(@class, '" + expectedClass + "')]", 1);
> ```

We seem to have lost something useful here. I don't think the new test needs to do such complex insertions as this one, but it would be nice to have some verification that an inserted page and a following pre-existing page both get updated to the right page number and side.
Then again...maybe it's enough just to test, as you almost do, that one page-side class gets fixed. That establishes that the relevant part of BBUTD is being called, and presumably tests elsewhere cover what that code does. If you decide we don't need any more, maybe just comment somewhere why you don't see a need for more elaborate checking of this aspect.
___
*[src/BloomTests/Spreadsheet/SpreadsheetImporterWithBookTests.cs, line 244 at r3](https://reviewable.io/reviews/bloombooks/bloomdesktop/4949#-MvVcKrp0xDnqHqRBkam:-MvVcKrp0xDnqHqRBkan:bpgtvrw) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/2b8154985bbf13966b841fff8376cd8e310a477f/src/BloomTests/Spreadsheet/SpreadsheetImporterWithBookTests.cs#L244)):*
> ```Smalltalk
> 		{
> 			var pageWithJustTextDom = AssertThatXmlIn.Element(_resultElement.SelectSingleNode("//div[@id='dc90dbe0-7584-4d9f-bc06-0e0326060054']") as XmlElement);
> 			// This page we added started out as 'side-right'.
> ```

Is the above a mistake? Because I'd think you'd want a case where the importer needs to change the side class.
Also...if this class is now the only test that import fixes page numbers and sides...seems like we'd want the importer to insert a content page, and check that the following page gets updated appropriately, and also that the inserted page is set up correctly. But maybe you have other tests that do that?

gmartin7 (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/4949)* status: 4 of 6 files reviewed, 3 unresolved discussions (waiting on @JohnThomson)

___
*[src/BloomExe/Spreadsheet/SpreadsheetImporter.cs, line 212 at r3](https://reviewable.io/reviews/bloombooks/bloomdesktop/4949#-MvVeWb602IyEEpGF1od:-MvVi4PmEZAutE-s6jLE:b-896fix) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/2b8154985bbf13966b841fff8376cd8e310a477f/src/BloomExe/Spreadsheet/SpreadsheetImporter.cs#L212)):*

While you're improving comments...might be useful to say what bad user experience you're guarding against. Then if someone considers taking it out they know what to look for.
</blockquote></details>

Done.
___
*[src/BloomTests/Spreadsheet/SpreadsheetImporterTests.cs, line 827 at r2](https://reviewable.io/reviews/bloombooks/bloomdesktop/4949#-MvVdKN-BYlsr4Chj-6x:-MvVjokh6lSZq_rReKSW:bmr22pc) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/650cebc378a826a680c803cb8058dd8a5b5e0d6e/src/BloomTests/Spreadsheet/SpreadsheetImporterTests.cs#L827)):*

We seem to have lost something useful here. I don't think the new test needs to do such complex insertions as this one, but it would be nice to have some verification that an inserted page and a following pre-existing page both get updated to the right page number and side.
Then again...maybe it's enough just to test, as you almost do, that one page-side class gets fixed. That establishes that the relevant part of BBUTD is being called, and presumably tests elsewhere cover what that code does. If you decide we don't need any more, maybe just comment somewhere why you don't see a need for more elaborate checking of this aspect.
</blockquote></details>

I added a comment in the new test class explaining why I don't think this test is necessary any longer.
___
*[src/BloomTests/Spreadsheet/SpreadsheetImporterWithBookTests.cs, line 244 at r3](https://reviewable.io/reviews/bloombooks/bloomdesktop/4949#-MvVcKrp0xDnqHqRBkam:-MvVhfywEkWwtBEfJmky:b-ap3v79) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/2b8154985bbf13966b841fff8376cd8e310a477f/src/BloomTests/Spreadsheet/SpreadsheetImporterWithBookTests.cs#L244)):*

Is the above a mistake? Because I'd think you'd want a case where the importer needs to change the side class.
Also...if this class is now the only test that import fixes page numbers and sides...seems like we'd want the importer to insert a content page, and check that the following page gets updated appropriately, and also that the inserted page is set up correctly. But maybe you have other tests that do that?
</blockquote></details>

Yes, I changed it under you while you were reviewing it. ;)

JohnThomson (approved):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/4949)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @gmartin7)

PR #4970, opened 2022-02-15
Use optipng to compress ePUB and BloomPub images (BL-9378)

StephenMcConnel:
This compression is applied after reducing the image dimensions as
desired.  It can possibly save an additional 10-50% in the size of
the ePUB or BloomPub.

---
This change is [Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/4970)

### DIFF
--- a/build/getDependencies-windows.sh
+++ b/build/getDependencies-windows.sh
@@ -77,7 +77,7 @@ cd -
 #     URL: https://build.palaso.org/viewType.html?buildTypeId=bt396
 #     clean: false
 #     revision: latest.lastSuccessful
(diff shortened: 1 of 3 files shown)


JohnThomson (changes requested):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/4970)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @JohnThomson and @StephenMcConnel)

___
*[src/BloomExe/Book/ImageUpdater.cs, line 230 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/4970#-Mw76fBI5vxyBs66dRwO:-Mw76fBJ9sudzagk6wWQ:b-mikhkn) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/f5487137e4e6f2c1604d7866fec3bc74bda17438/src/BloomExe/Book/ImageUpdater.cs#L230)):*
> ```Smalltalk
> 			if (SIL.PlatformUtilities.Platform.IsWindows)
> 				optipngPath = FileLocationUtilities.LocateExecutable("optipng.exe");
> 			var result = CommandLineRunner.Run(optipngPath, "-force \"" + path + "\"", Encoding.UTF8,
> ```

Do we not need to handle the possibility of not finding it? e.g. even if our installer is still installing it, maybe some antivirus has removed it?

StephenMcConnel:
We've decided to remove optipng instead of trying to use it.

PR #4972, opened 2022-02-16
Add buttons for text and background color (BL-10059)

JohnThomson:
This change is [Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/4972)


JohnThomson:
This is a draft. There's a problem with the color palette box coming up too soon after using the tool once which we'll probably want to work on before merging. But I thought it was worth putting up for other developers to try.

JohnThomson:
I don't think there's any need to review the code under ckeditor/plugins. This is copied unmodified from ckeditor.

hatton:
Tell us about why there are so many files changing?

JohnThomson:
Sorry. That comment was to explain why it was originally a draft. It's ready to review now.
There are many new files because we added a CkEditor plugin that implements the new button, and it has many other plugins that are dependencies which the doc says must be added as well. I haven't tried not doing so. There's no obvious way to get these files and ship them without checking them in. I believe only two files are actually changed (bloomEditing.ts and lib/ckeditor/config.js)

PR #4990, opened 2022-02-21
Update TitleBestForUserDisplay -> NameBestForUserDisplay (BL-10960)

sujeffreyl:
This change is [Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/4990)

### DIFF
--- a/src/BloomExe/CollectionTab/LibraryModel.cs
+++ b/src/BloomExe/CollectionTab/LibraryModel.cs
@@ -178,7 +178,7 @@ namespace Bloom.CollectionTab
 		{
 			// I hope we can get rid of this when we retire the old LibraryListView, but for now we need to keep both views up to date.
 			// optimize: we only need to reload the first (editable) collection; better yet, we only need to add the one new book to it.
-			ReloadCollections();
+			ReloadCollections();       
 		}
 
 		public void UpdateLabelOfBookInEditableCollection(Book.Book book)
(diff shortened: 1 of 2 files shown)


sujeffreyl (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/4990)* status: 0 of 2 files reviewed, 1 unresolved discussion

___
*[src/BloomExe/web/controllers/BookCommandsApi.cs, line 170 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/4990#-MwSQqe55I4F-oEMWbJh:-MwSQqe55I4F-oEMWbJi:be4z1mu) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/94337ba103f263e818005ef4cdfd5bbbd8292c64/src/BloomExe/web/controllers/BookCommandsApi.cs#L170)):*
> ```Smalltalk
> 				if (book == null)
> 					return; // can't get the book, can't improve title
> 				bestTitle = book.NameBestForUserDisplay;
> ```

This is not strictly speaking necessary to fix this particular card, but I've identified it as another reference that missed getting updated.  This function corresponds pretty well with the ImproveAndRefreshBookButtons function from the old LibraryListView.cs (now deleted). We did update the reference in ImproveAndRefreshBookButtons() to be NameBestForUserDisplay instead of TitleBest

sujeffreyl (commented):
BTW, I didn't really directly address anything w/race conditions. (I can't even repro both paths anymore).  Seems like there are multipe websocket events happening 'reload' and 'book' and they weren't all returning the same bookLabel.  I am satisfied now that all paths set (if needed) and get the correct value of bookLabel.

> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/4990#-:-MwSTCTT0G0gCQ8O2ddk:bz7gj2i)* status: 0 of 2 files reviewed, 1 unresolved discussion

PR #5017, opened 2022-03-03
Prevent crashes and warn users about .ttc font files (BH-6108)

StephenMcConnel:
ePUBs were crashing and bloomPubs weren't warning users that these fonts
can't be published.

---
This change is [Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5017)

### DIFF
--- a/DistFiles/localization/en/Bloom.xlf
+++ b/DistFiles/localization/en/Bloom.xlf
@@ -3411,6 +3411,11 @@
         <note>ID: PublishTab.Android.File.Progress.Embedding</note>
         <note>{1} is a number with one decimal place, the number of megabytes the font file takes up</note>
       </trans-unit>
+      <trans-unit id="PublishTab.Android.File.Progress.IncompatibleFontFileFormat" sil:dynamic="true">
+        <source xml:lang="en">This book has text in a font named \"{0}\". Bloom cannot use a font in this font's file format.</source>
+        <note>ID: PublishTab.Android.File.Progress.IncompatibleFontFileFormat</note>
+        <note>{0} is a font name</note>
+      </trans-unit>
       <trans-unit id="PublishTab.Android.File.Progress.LicenseForbids" sil:dynamic="true">
         <source xml:lang="en">This book has text in a font named \"{0}\". The license for \"{0}\" does not permit Bloom to embed the font in the book.</source>
         <note>ID: PublishTab.Android.File.Progress.LicenseForbids</note>
--- a/src/BloomExe/Publish/Android/BloomPubMaker.cs
+++ b/src/BloomExe/Publish/Android/BloomPubMaker.cs
@@ -402,10 +402,12 @@ namespace Bloom.Publish.Android
 			fontsWanted.Remove(defaultFont);
 			fontFileFinder.NoteFontsWeCantInstall = true;
 			var filesToEmbed = new List<string>();
+			var badFonts = new HashSet<string>();
 			foreach (var font in fontsWanted)
 			{
 				var fontFiles = fontFileFinder.GetFilesForFont(font);
-				if (fontFiles.Count() > 0)
+				var badFileType = fontFiles.Count() > 0 && !FontMetadata.fontFileTypesBloomKnows.Contains(Path.GetExtension(fontFiles.First()).ToLowerInvariant());
+				if (fontFiles.Count() > 0 && !badFileType)
 				{
 					filesToEmbed.AddRange(fontFiles);
 					progress.MessageWithParams("PublishTab.Android.File.Progress.CheckFontOK", "{0} is a font name", "Checking {0} font: License OK for embedding.", ProgressKind.Progress, font);
@@ -419,7 +421,11 @@ namespace Bloom.Publish.Android
 						font, sizeToReport);
 					continue;
 				}
-				if (fontFileFinder.FontsWeCantInstall.Contains(font))
+				if (badFileType)
+				{
+					progress.MessageWithParams("IncompatibleFontFileFormat", "{0} is a font name", "This book has text in a font named \"{0}\". Bloom cannot use a font in this font's file format.", ProgressKind.Error, font);
+				}
+				else if (fontFileFinder.FontsWeCantInstall.Contains(font))
 				{
 					//progress.Error("Common.Warning", "Warning");
 					progress.MessageWithParams("LicenseForbids","{0} is a font name", "This book has text in a font named \"{0}\". The license for \"{0}\" does not permit Bloom to embed the font in the book.",ProgressKind.Error, font);
@@ -429,6 +435,7 @@ namespace Bloom.Publish.Android
 					progress.MessageWithParams("NoFontFound", "{0} is a font name", "This book has text in a font named \"{0}\", but Bloom could not find that font on this computer.", ProgressKind.Error, font);
 				}
 				progress.MessageWithParams("SubstitutingAndika", "{0} is a font name", "Bloom will substitute \"{0}\" instead.", ProgressKind.Error, defaultFont, font);
+				badFonts.Add(font);	// need to prevent the bad/missing font from showing up in fonts.css and elsewhere
 			}
 			foreach (var file in filesToEmbed)
 			{
@@ -440,6 +447,8 @@ namespace Bloom.Publish.Android
 			var sb = new StringBuilder();
 			foreach (var font in fontsWanted)
 			{
+				if (badFonts.Contains(font))
+					continue;
 				var group = fontFileFinder.GetGroupForFont(font);
 				if (group != null)
 				{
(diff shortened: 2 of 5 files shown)


JohnThomson (changes requested):
A few minor suggestions.

> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5017#-:-MxZqKDQBPNCZwFMBg7W:b-drw3d6)* status: all files reviewed, 4 unresolved discussions (waiting on @StephenMcConnel)

___
*[src/BloomExe/Publish/Android/BloomPubMaker.cs, line 423 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5017#-MxZlKwAAcp0paZs9xcA:-MxZlKwAAcp0paZs9xcB:b-579m5j) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/323db194ce4c1473dc52d904b80b2c9ede63f182/src/BloomExe/Publish/Android/BloomPubMaker.cs#L423)):*
> ```Smalltalk
> 				if (badFileType)
> 				{
> 					progress.MessageWithParams("IncompatibleFontFileFormat", "{0} is a font name", "This book has text in a font named \"{0}\". The font's file format cannot be used for electronic publishing.", ProgressKind.Error, font);
> ```

I'm wondering if it would be better to say "Bloom cannot use a font in this file format...". We (at least I) don't know for sure that there is no way to use it for electronic publishing.
___
*[src/BloomExe/Publish/Android/BloomPubMaker.cs, line 434 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5017#-MxZkzXi21elyI_yX7bP:-MxZkzXj5dPxZ5i_UZwa:bq5sw47) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/323db194ce4c1473dc52d904b80b2c9ede63f182/src/BloomExe/Publish/Android/BloomPubMaker.cs#L434)):*
> ```Smalltalk
> 					progress.MessageWithParams("NoFontFound", "{0} is a font name", "This book has text in a font named \"{0}\", but Bloom could not find that font on this computer.", ProgressKind.Error, font);
> 				}
> 				// REVIEW/TODO: If we say this, shouldn't we revise fonts.css and defaultLangStyles.css to replace/remove the invalidated font? (and maybe other css files or styles in the html file)
> ```

If Bloom is actually NOT using ANB when it can't find a font like we say here (in any kinds of pubs for which we use this message), we should either fix it immediately or make a card.
___
*[src/BloomExe/Publish/Epub/EpubMaker.cs, line 1991 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5017#-MxZlsA66GhDRv2MI1ar:-MxZlsA66GhDRv2MI1as:batfcaq) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/323db194ce4c1473dc52d904b80b2c9ede63f182/src/BloomExe/Publish/Epub/EpubMaker.cs#L1991)):*
> ```Smalltalk
> 		/// Try to embed the fonts we need.
> 		/// </summary>
> 		private void EmbedFonts(ISet<string> warningMessages)
> ```

Not very obvious this is effectively an out param. I thought this was a well-known design pattern called Collector, but can't find any reference to that online; still, calling it a warningMessageCollector might help? Unless you can find the proper name of the strategy.
Or, since it's currently returning void, if you don't ever pass a non-empty collection, you could just have it return a new set.
___
*[src/BloomExe/Publish/Epub/EpubMaker.cs, line 2012 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5017#-MxZpp8RCJYYlsttPdWr:-MxZpp8RCJYYlsttPdWs:b-vcxr3j) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/323db194ce4c1473dc52d904b80b2c9ede63f182/src/BloomExe/Publish/Epub/EpubMaker.cs#L2012)):*
> ```Smalltalk
> 				{
> 					// If we can't embed the font, no reason to refer to it in the css.
> 					continue;
> ```

Is this where we should embed ANB, if we haven't already, and tell it to use that when it sees this font name?

StephenMcConnel (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5017)* status: 0 of 4 files reviewed, 4 unresolved discussions (waiting on @JohnThomson)

___
*[src/BloomExe/Publish/Android/BloomPubMaker.cs, line 423 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5017#-MxZlKwAAcp0paZs9xcA:-Mx_JqbiBUWS5cyIZrku:b-896fix) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/323db194ce4c1473dc52d904b80b2c9ede63f182/src/BloomExe/Publish/Android/BloomPubMaker.cs#L423)):*

I'm wondering if it would be better to say "Bloom cannot use a font in this file format...". We (at least I) don't know for sure that there is no way to use it for electronic publishing.
</blockquote></details>

Done.
___
*[src/BloomExe/Publish/Android/BloomPubMaker.cs, line 434 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5017#-MxZkzXi21elyI_yX7bP:-Mx_YVEBAT2dj5fDvgkQ:b57428h) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/323db194ce4c1473dc52d904b80b2c9ede63f182/src/BloomExe/Publish/Android/BloomPubMaker.cs#L434)):*

If Bloom is actually NOT using ANB when it can't find a font like we say here (in any kinds of pubs for which we use this message), we should either fix it immediately or make a card.
</blockquote></details>

I've added code to ensure that the bad font doesn't make it into fonts.css and is replaced in defaultLangStyles.css and in style settings in the html.
___
*[src/BloomExe/Publish/Epub/EpubMaker.cs, line 1991 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5017#-MxZlsA66GhDRv2MI1ar:-Mx_Oe7-B7YsBxi8pFmG:b4x7j6x) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/323db194ce4c1473dc52d904b80b2c9ede63f182/src/BloomExe/Publish/Epub/EpubMaker.cs#L1991)):*

Not very obvious this is effectively an out param. I thought this was a well-known design pattern called Collector, but can't find any reference to that online; still, calling it a warningMessageCollector might help? Unless you can find the proper name of the strategy.
Or, since it's currently returning void, if you don't ever pass a non-empty collection, you could just have it return a new set.
</blockquote></details>

Changed to return a new (possibly empty) set of warning messages.
___
*[src/BloomExe/Publish/Epub/EpubMaker.cs, line 2012 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5017#-MxZpp8RCJYYlsttPdWr:-MxaSU1OA4ucsZ_41W-u:bbk3da6) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/323db194ce4c1473dc52d904b80b2c9ede63f182/src/BloomExe/Publish/Epub/EpubMaker.cs#L2012)):*

Is this where we should embed ANB, if we haven't already, and tell it to use that when it sees this font name?
</blockquote></details>

There's a comment just below explaining why we never embed ANB.  If that comment is outdated, then this would be the place to embed it.

JohnThomson (changes requested):
Much better. I would have thought, though, that we probably have a unit test for embedding fonts, and it would be useful to extend it with a test that a ttc font does not get embedded and results in a warning. That may not be feasible...if there's no ttc font we can be sure will be available to exclude...but it 's worth considering.
I think FixCssReferencesForBadFonts could also usefully be unit tested...might be easier if you refactor into a method that cleans up a CSS string and a wrapper that knows where to find and replace the CSS that should be changed.

> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5017#-:-MyJ8QEc6GF2DBe3qAKE:b-2pxtzr)* status: all files reviewed, 1 unresolved discussion (waiting on @StephenMcConnel)

___
*[src/BloomExe/Publish/Android/BloomPubMaker.cs, line 483 at r2](https://reviewable.io/reviews/bloombooks/bloomdesktop/5017#-MyJ6iSn-D1Ewp4P76I8:-MyJ6iSo5Yg3ETL4D8z2:bsym80x) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/978cde4703450542a7c9fe3ae3bef709919235ed/src/BloomExe/Publish/Android/BloomPubMaker.cs#L483)):*
> ```Smalltalk
> 					cssText = cssRegex.Replace(cssText, $"font-family: '{defaultFont}';");
> 				}
> 				// Note that the font may be referred to in customCollectionStyles.css or in a style defined in the HTML.
> ```

This comment feels out of place

StephenMcConnel (commented):
I enhanced the existing EmbedFonts unit test to handle both the .ttc font file and the fixup for CSS references.

> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5017#-:-MyJhQju4ZIevWEmm1JU:b3k20u0)* status: 3 of 5 files reviewed, 1 unresolved discussion (waiting on @JohnThomson)

___
*[src/BloomExe/Publish/Android/BloomPubMaker.cs, line 483 at r2](https://reviewable.io/reviews/bloombooks/bloomdesktop/5017#-MyJ6iSn-D1Ewp4P76I8:-MyJPD_40gbjJsBMO8He:b-ub2rxd) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/978cde4703450542a7c9fe3ae3bef709919235ed/src/BloomExe/Publish/Android/BloomPubMaker.cs#L483)):*

This comment feels out of place
</blockquote></details>

Expanded the comment and moved it to the beginning of the method.

JohnThomson (approved):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5017)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @StephenMcConnel)

PR #5029, opened 2022-03-10
Restore original size and make TeamCollectionDialog resizable (BL-11016)

StephenMcConnel:


StephenMcConnel:
We've decided to make this a pure React dialog rather than Winforms hosted.

PR #5039, opened 2022-03-16
Move "Make Bloom Pack" menu item to collection menu (BL-10365)

StephenMcConnel:
This change is [Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5039)

### DIFF
--- a/DistFiles/localization/en/Bloom.xlf
+++ b/DistFiles/localization/en/Bloom.xlf
@@ -779,6 +779,12 @@
       <trans-unit id="CollectionTab.MakeBloomPackButton">
         <source xml:lang="en">Make Bloom Pack</source>
         <note>ID: CollectionTab.MakeBloomPackButton</note>
+        <note>This string is obsolete as of Bloom version 5.3.</note>
+      </trans-unit>
+      <trans-unit id="CollectionTab.MakeBloomPackOfShellBooks">
+        <source xml:lang="en">Make Bloom Pack of Shell Books</source>
+        <note>Shell Books are books that are meant to be translated.</note>
+        <note>ID: CollectionTab.MakeBloomPackOfShellBooks</note>
       </trans-unit>
       <trans-unit id="CollectionTab.MakeBookUsingThisTemplate">
         <source xml:lang="en">Make a book using this source</source>
--- a/src/BloomBrowserUI/collectionsTab/CollectionsTabPane.tsx
+++ b/src/BloomBrowserUI/collectionsTab/CollectionsTabPane.tsx
@@ -129,6 +129,13 @@ export const CollectionsTabPane: React.FunctionComponent<{}> = () => {
             addEllipsis: true
         },
         {
+            label: "Make Bloom Pack of Shell Books",
+            l10nId: "CollectionTab.MakeBloomPackOfShellBooks",
+            command: "collections/makeShellBooksBloompack",
+            addEllipsis: true,
+            shouldShow: () => collections[0].isShellProject
+        },
+        {
             label: "Make Reader Template Bloom Pack...",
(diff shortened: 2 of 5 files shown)


JohnThomson (changes requested):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5039)* status: all files reviewed, 1 unresolved discussion (waiting on @StephenMcConnel)

___
*[src/BloomExe/CollectionTab/CollectionTabView.cs, line 57 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5039#-MyJBLj83BwqBIs7NNcw:-MyJBLj83BwqBIs7NNcx:bll92vs) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/c184833f84721d3d648e2edfdd1177d7096bbcdf/src/BloomExe/CollectionTab/CollectionTabView.cs#L57)):*
> ```Smalltalk
> 
> 			//TODO splitContainer1.SplitterDistance = _collectionListView.PreferredWidth;
> 			_makeBloomPackButton.Visible = model.IsShellProject;
> ```

This bit of logic has gotten lost.
- Is it no longer desirable to restrict this behavior to source collections?
- If not, take a look at the logic related to deciding whether a book IsSuitableForMakingVernacularBooks or something like that, and make sure that a BloomPack made using this code starting from a vernacular collection actually functions correctly as a collection of shell books. I have a vague recollection that books in a source collection have some different flags related to editability which are overridden in methods like IsEditable if the collection is a source one. We're wanting to lose this source/vernacular distinction, so this may not be any kind of problem, but I wanted you to at least think about the implications of making a bloompack of "shell" books from a non-source collection. One way to investigate would be to look at meta.json in a book newly created in a source collection, put in a shell bloompack, and unpacked, then do the same thing in a vernacular collection. Any diffs in the resulting meta.json that don't have an obvious reason deserve thought.

StephenMcConnel (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5039)* status: all files reviewed, 1 unresolved discussion (waiting on @JohnThomson)

___
*[src/BloomExe/CollectionTab/CollectionTabView.cs, line 57 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5039#-MyJBLj83BwqBIs7NNcw:-MyJsIGH6JRshnaaGfb8:b-f22k7h) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/c184833f84721d3d648e2edfdd1177d7096bbcdf/src/BloomExe/CollectionTab/CollectionTabView.cs#L57)):*

This bit of logic has gotten lost.
- Is it no longer desirable to restrict this behavior to source collections?
- If not, take a look at the logic related to deciding whether a book IsSuitableForMakingVernacularBooks or something like that, and make sure that a BloomPack made using this code starting from a vernacular collection actually functions correctly as a collection of shell books. I have a vague recollection that books in a source collection have some different flags related to editability which are overridden in methods like IsEditable if the collection is a source one. We're wanting to lose this source/vernacular distinction, so this may not be any kind of problem, but I wanted you to at least think about the implications of making a bloompack of "shell" books from a non-source collection. One way to investigate would be to look at meta.json in a book newly created in a source collection, put in a shell bloompack, and unpacked, then do the same thing in a vernacular collection. Any diffs in the resulting meta.json that don't have an obvious reason deserve thought. 
</blockquote></details>

I've now restricted this menu item to source collections, postponing any decision about whether "source collection" is a useful/meaningful concept.  :-)
This is how I interpreted JohnH's comment "let's just preserve the old behavior" in the flowdock discussion.

PR #5058, opened 2022-03-24
React copyright and license dialog (BL-6259)

andrew-polk:
This change is [Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5058)

### DIFF
--- a/DistFiles/localization/en/Bloom.xlf
+++ b/DistFiles/localization/en/Bloom.xlf
@@ -3128,7 +3128,7 @@
         <source xml:lang="en">Comic Portrait</source>
         <note>ID: LayoutChoices.ComicPortrait</note>
       </trans-unit>
-<!--
+<!--5.3
       <trans-unit id="License.About" sil:dynamic="true">
         <source xml:lang="en">About %0</source>
         <note>ID: License.About</note>
--- a/DistFiles/localization/en/Bloom.xlf
+++ b/DistFiles/localization/en/Bloom.xlf
@@ -1122,6 +1122,7 @@
         <note>ID: Common.WhatCausedThisProblem</note>
         <note>This is usually a link to an explanatory document.</note>
       </trans-unit>
+<!--5.3
       <trans-unit id="Copyright.CopyrightHolder" sil:dynamic="true">
         <source xml:lang="en">Copyright Holder</source>
         <note>ID: Copyright.CopyrightHolder</note>
@@ -1155,6 +1156,7 @@
         <source xml:lang="en">Copyright and License</source>
         <note>ID: CopyrightAndLicense</note>
       </trans-unit>
+-->
       <trans-unit id="Download.Completed">
         <source xml:lang="en">Your download ({0}) is complete. You can see it in the 'Books from BloomLibrary.org' section of your Collections.</source>
         <note>ID: Download.Completed</note>
@@ -3126,6 +3128,7 @@
         <source xml:lang="en">Comic Portrait</source>
         <note>ID: LayoutChoices.ComicPortrait</note>
       </trans-unit>
+<!--
       <trans-unit id="License.About" sil:dynamic="true">
         <source xml:lang="en">About %0</source>
         <note>ID: License.About</note>
@@ -3188,6 +3191,7 @@
         <source xml:lang="en">Anyone can use it however they want.</source>
         <note>ID: License.PublicDomain</note>
       </trans-unit>
+-->
       <trans-unit id="LicenseDialog.WindowTitle">
         <source xml:lang="en">Bloom {0}</source>
         <note>ID: LicenseDialog.WindowTitle</note>
--- a/DistFiles/localization/en/Bloom.xlf
+++ b/DistFiles/localization/en/Bloom.xlf
@@ -259,6 +259,7 @@
       <trans-unit id="BookMetadata.License">
         <source xml:lang="en">License</source>
         <note>ID: BookMetadata.License</note>
+        <note>Obsolete as of Bloom 5.3 (replaced by Common.License)</note>
       </trans-unit>
       <trans-unit id="BookMetadata.metapicture">
         <source xml:lang="en">Picture</source>
@@ -958,6 +959,10 @@
         <source xml:lang="en">Copy</source>
         <note>ID: Common.Copy</note>
       </trans-unit>
+      <trans-unit id="Common.Copyright">
+        <source xml:lang="en">Copyright</source>
+        <note>ID: Common.Copyright</note>
+      </trans-unit>
       <trans-unit id="Common.Delete">
         <source xml:lang="en">Delete</source>
         <note>ID: Common.Delete</note>
@@ -1000,6 +1005,10 @@
         <note>A link or button that leads to something that tells the user more about what they just read.</note>
         <note>This is the same as Settings.Enterprise.LearnMore</note>
       </trans-unit>
+      <trans-unit id="Common.License" sil:dynamic="true">
+        <source xml:lang="en">License</source>
+        <note>ID: Common.License</note>
+      </trans-unit>
       <trans-unit id="Common.Loading" sil:dynamic="true">
         <source xml:lang="en">Loading...</source>
         <note>ID: Common.Loading</note>
@@ -1031,6 +1040,10 @@
       <trans-unit id="Common.Optional">
         <source xml:lang="en">optional</source>
         <note>ID: Common.Optional</note>
+      </trans-unit>
+      <trans-unit id="Common.Other" sil:dynamic="true">
+        <source xml:lang="en">Other</source>
+        <note>ID: Common.Other</note>
       </trans-unit>
       <trans-unit id="Common.Pause" sil:dynamic="true">
         <source xml:lang="en">Pause</source>
@@ -1108,6 +1121,39 @@
         <source xml:lang="en">What caused this problem?</source>
         <note>ID: Common.WhatCausedThisProblem</note>
         <note>This is usually a link to an explanatory document.</note>
+      </trans-unit>
+      <trans-unit id="Copyright.CopyrightHolder" sil:dynamic="true">
+        <source xml:lang="en">Copyright Holder</source>
+        <note>ID: Copyright.CopyrightHolder</note>
+      </trans-unit>
+      <trans-unit id="Copyright.CopyrightYear" sil:dynamic="true">
+        <source xml:lang="en">Copyright Year</source>
+        <note>ID: Copyright.CopyrightYear</note>
+      </trans-unit>
+      <trans-unit id="Copyright.FollowSILGuidelines" sil:dynamic="true">
+        <source xml:lang="en">Before publishing as SIL, ensure that you follow [SIL corporate guidelines].</source>
+        <note>ID: Copyright.FollowSILGuidelines</note>
+        <note>The text inside the [square brackets] will become a link to a website.</note>
+      </trans-unit>
+      <trans-unit id="Copyright.IllustratorOrPhotographer" sil:dynamic="true">
+        <source xml:lang="en">Illustrator/Photographer</source>
+        <note>ID: Copyright.IllustratorOrPhotographer</note>
+      </trans-unit>
+      <trans-unit id="Copyright.NotATranslation" sil:dynamic="true">
+        <source xml:lang="en">Not a translation or new version</source>
+        <note>ID: Copyright.NotATranslation</note>
+      </trans-unit>
+      <trans-unit id="Copyright.PublishingAsSIL" sil:dynamic="true">
+        <source xml:lang="en">Publishing as SIL</source>
+        <note>ID: Copyright.PublishingAsSIL</note>
+      </trans-unit>
+      <trans-unit id="Copyright.UseOriginalCopyright" sil:dynamic="true">
+        <source xml:lang="en">Continue to use the same copyright and license as the original book:</source>
+        <note>ID: Copyright.UseOriginalCopyright</note>
+      </trans-unit>
+      <trans-unit id="CopyrightAndLicense" sil:dynamic="true">
+        <source xml:lang="en">Copyright and License</source>
+        <note>ID: CopyrightAndLicense</note>
       </trans-unit>
       <trans-unit id="Download.Completed">
         <source xml:lang="en">Your download ({0}) is complete. You can see it in the 'Books from BloomLibrary.org' section of your Collections.</source>
@@ -2893,11 +2939,6 @@
         <note>ID: FontInformationPane.FontUnsuitable</note>
         <note>This shows in the popup when hovering over a font that Bloom can't legally host on its website.</note>
       </trans-unit>
-      <trans-unit id="FontInformationPane.License">
-        <source xml:lang="en">License</source>
-        <note>ID: FontInformationPane.License</note>
-        <note>This shows in the popup as a link to the font's license.</note>
-      </trans-unit>
       <trans-unit id="FontInformationPane.Styles">
         <source xml:lang="en">Styles</source>
         <note>ID: FontInformationPane.Styles</note>
@@ -3084,6 +3125,68 @@
       <trans-unit id="LayoutChoices.ComicPortrait" sil:dynamic="true">
         <source xml:lang="en">Comic Portrait</source>
         <note>ID: LayoutChoices.ComicPortrait</note>
+      </trans-unit>
+      <trans-unit id="License.About" sil:dynamic="true">
+        <source xml:lang="en">About %0</source>
+        <note>ID: License.About</note>
+        <note>%0 is a shorthand version of the Creative Commons license, such as "CC-BY"</note>
+      </trans-unit>
+      <trans-unit id="License.AllRightsReserved" sil:dynamic="true">
+        <source xml:lang="en">All Rights Reserved</source>
+        <note>ID: License.AllRightsReserved</note>
+      </trans-unit>
+      <trans-unit id="License.AllRightsReserved.Description" sil:dynamic="true">
+        <source xml:lang="en">No one can modify or copy it without explicit permission. We will not be sharing it via Bloom Reader or BloomLibrary.org.</source>
+        <note>ID: License.AllRightsReserved.Description</note>
+      </trans-unit>
+      <trans-unit id="License.CannotChangeLicense" sil:dynamic="true">
+        <source xml:lang="en">The license of the original book does not allow changing the license in new versions.</source>
+        <note>ID: License.CannotChangeLicense</note>
+      </trans-unit>
+      <trans-unit id="License.CopyrightHolderAllows" sil:dynamic="true">
+        <source xml:lang="en">The copyright holder allows others to use the book in this way:</source>
+        <note>ID: License.CopyrightHolderAllows</note>
+      </trans-unit>
+      <trans-unit id="License.CreativeCommons.Additional" sil:dynamic="true">
+        <source xml:lang="en">You may use this space to clarify or grant additional permissions (for example, that translations are allowed), but not to alter the license.</source>
+        <note>ID: License.CreativeCommons.Additional</note>
+      </trans-unit>
+      <trans-unit id="License.CreativeCommons.AllowCommercial" sil:dynamic="true">
+        <source xml:lang="en">use the book in a commercial way</source>
+        <note>ID: License.CreativeCommons.AllowCommercial</note>
+      </trans-unit>
+      <trans-unit id="License.CreativeCommons.CopyForFree" sil:dynamic="true">
+        <source xml:lang="en">copy this book for free</source>
+        <note>ID: License.CreativeCommons.CopyForFree</note>
+      </trans-unit>
+      <trans-unit id="License.CreativeCommons.DifferentLicense" sil:dynamic="true">
+        <source xml:lang="en">apply a different license to new versions of this book</source>
+        <note>ID: License.CreativeCommons.DifferentLicense</note>
+      </trans-unit>
+      <trans-unit id="License.CreativeCommons.Intro" sil:dynamic="true">
+        <source xml:lang="en">Everyone is allowed to:</source>
+        <note>ID: License.CreativeCommons.Intro</note>
+      </trans-unit>
+      <trans-unit id="License.CreativeCommons.ShareAlike" sil:dynamic="true">
+        <source xml:lang="en">make new versions of this book, but they must keep the author, illustrator, and other credits</source>
+        <note>ID: License.CreativeCommons.ShareAlike</note>
+      </trans-unit>
+      <trans-unit id="License.Custom" sil:dynamic="true">
+        <source xml:lang="en">Custom</source>
+        <note>ID: License.Custom</note>
+      </trans-unit>
+      <trans-unit id="License.Custom.AvoidIfPossible" sil:dynamic="true">
+        <source xml:lang="en">Avoid writing a custom license if at all possible. They are difficult to write, interpret, and enforce.</source>
+        <note>ID: License.Custom.AvoidIfPossible</note>
+      </trans-unit>
+      <trans-unit id="License.OriginalLicense" sil:dynamic="true">
+        <source xml:lang="en">The original version of this book is %0</source>
+        <note>ID: License.OriginalLicense</note>
+        <note>%0 is a shorthand version of the license such as "CC-BY"</note>
+      </trans-unit>
+      <trans-unit id="License.PublicDomain" sil:dynamic="true">
+        <source xml:lang="en">Anyone can use it however they want.</source>
+        <note>ID: License.PublicDomain</note>
       </trans-unit>
       <trans-unit id="LicenseDialog.WindowTitle">
         <source xml:lang="en">Bloom {0}</source>
@@ -4188,6 +4291,7 @@
       <trans-unit id="PublishTab.Upload.Copyright">
         <source xml:lang="en">Copyright</source>
         <note>ID: PublishTab.Upload.Copyright</note>
+        <note>Obsolete as of Bloom 5.3 (replaced by Common.Copyright)</note>
       </trans-unit>
       <trans-unit id="PublishTab.Upload.Credits">
         <source xml:lang="en">credits</source>
--- a/DistFiles/localization/en/Palaso.xlf
+++ b/DistFiles/localization/en/Palaso.xlf
@@ -386,93 +386,114 @@
       <trans-unit id="MetadataEditor.AllowCommercialUse">
         <source xml:lang="en">Allow commercial uses of your work?</source>
         <note>ID: MetadataEditor.AllowCommercialUse</note>
+        <note>Obsolete as of Bloom 5.3</note>
       </trans-unit>
       <trans-unit id="MetadataEditor.AllowDerivatives">
         <source xml:lang="en">Allow modifications of your work?</source>
         <note>ID: MetadataEditor.AllowDerivatives</note>
+        <note>Obsolete as of Bloom 5.3</note>
       </trans-unit>
       <trans-unit id="MetadataEditor.CopyrightHolder">
         <source xml:lang="en">Copyright Holder</source>
         <note>ID: MetadataEditor.CopyrightHolder</note>
+        <note>Obsolete as of Bloom 5.3 (moved to Bloom.xlf with a different ID)</note>
       </trans-unit>
       <trans-unit id="MetadataEditor.CopyrightYear">
         <source xml:lang="en">Copyright Year</source>
         <note>ID: MetadataEditor.CopyrightYear</note>
+        <note>Obsolete as of Bloom 5.3 (moved to Bloom.xlf with a different ID)</note>
       </trans-unit>
       <trans-unit id="MetadataEditor.CreativeCommons">
         <source xml:lang="en">Creative Commons</source>
         <note>ID: MetadataEditor.CreativeCommons</note>
+        <note>Obsolete as of Bloom 5.3</note>
       </trans-unit>
       <trans-unit id="MetadataEditor.CreativeCommons.Intergovernmental">
         <source xml:lang="en">Intergovernmental Version</source>
         <note>ID: MetadataEditor.CreativeCommons.Intergovernmental</note>
+        <note>Obsolete as of Bloom 5.3</note>
       </trans-unit>
       <trans-unit id="MetadataEditor.CreatorLabel">
         <source xml:lang="en">Creator</source>
         <note>ID: MetadataEditor.CreatorLabel</note>
+        <note>Obsolete as of Bloom 5.3</note>
       </trans-unit>
       <trans-unit id="MetadataEditor.CustomLicense">
         <source xml:lang="en">Custom</source>
         <note>ID: MetadataEditor.CustomLicense</note>
+        <note>Obsolete as of Bloom 5.3 (moved to Bloom.xlf with a different ID)</note>
       </trans-unit>
       <trans-unit id="MetadataEditor.License">
         <source xml:lang="en">License</source>
         <note>ID: MetadataEditor.License</note>
+        <note>Obsolete as of Bloom 5.3 (moved to Bloom.xlf with a different ID)</note>
       </trans-unit>
       <trans-unit id="MetadataEditor.PublicDomain">
         <source xml:lang="en">The copyright holder waives all rights</source>
         <note>ID: MetadataEditor.PublicDomain</note>
+        <note>Obsolete as of Bloom 5.3</note>
       </trans-unit>
       <trans-unit id="MetadataEditor.PublicDomain_ToolTip_">
         <source xml:lang="en">The copyright holder waives all rights</source>
         <note>ID: MetadataEditor.PublicDomain_ToolTip_</note>
         <note>This tooltip string is identical to the string displayed in the dialog in case there is not enough room in the dialog to display the full string.</note>
+        <note>Obsolete as of Bloom 5.3</note>
       </trans-unit>
       <trans-unit id="MetadataEditor.TitleNoCredit">
         <source xml:lang="en">Copyright and License</source>
         <note>ID: MetadataEditor.TitleNoCredit</note>
+        <note>Obsolete as of Bloom 5.3 (moved to Bloom.xlf with a different ID)</note>
       </trans-unit>
       <trans-unit id="MetadataEditor.TitleWithCredit">
         <source xml:lang="en">Credit, Copyright, and License</source>
         <note>ID: MetadataEditor.TitleWithCredit</note>
+        <note>Obsolete as of Bloom 5.3</note>
       </trans-unit>
       <trans-unit id="MetadataEditor.UnknownLicense">
         <source xml:lang="en">Contact the copyright holder for any permissions</source>
         <note>ID: MetadataEditor.UnknownLicense</note>
+        <note>Obsolete as of Bloom 5.3</note>
       </trans-unit>
       <trans-unit id="MetadataEditor.UnknownLicense_ToolTip_">
         <source xml:lang="en">Contact the copyright holder for any permissions</source>
         <note>ID: MetadataEditor.UnknownLicense_ToolTip_</note>
         <note>This tooltip string is identical to the string displayed in the dialog in case there is not enough room in the dialog to display the full string.</note>
+        <note>Obsolete as of Bloom 5.3</note>
       </trans-unit>
       <trans-unit id="MetadataEditor.YesShareAlike">
         <source xml:lang="en">Yes, as long as others share alike</source>
         <note>ID: MetadataEditor.YesShareAlike</note>
+        <note>Obsolete as of Bloom 5.3</note>
       </trans-unit>
       <trans-unit id="MetadataEditor.YesShareAlike_ToolTip_">
         <source xml:lang="en">Yes, as long as others share alike</source>
         <note>ID: MetadataEditor.YesShareAlike_ToolTip_</note>
         <note>This tooltip string is identical to the string displayed in the dialog in case there is not enough room in the dialog to display the full string.</note>
+        <note>Obsolete as of Bloom 5.3</note>
       </trans-unit>
       <trans-unit id="MetadataEditor.additionalRequestsLabel">
         <source xml:lang="en">Additional Requests</source>
         <note>ID: MetadataEditor.additionalRequestsLabel</note>
         <note>When you choose a Creative Commons License, this label shows over the text box at the bottom.</note>
+        <note>Obsolete as of Bloom 5.3</note>
       </trans-unit>
       <trans-unit id="MetadataEditor.betterLinkLabel1">
         <source xml:lang="en">more info</source>
         <note>ID: MetadataEditor.betterLinkLabel1</note>
         <note>The meaning of  "non-commercial" is vague but important. This hyperlink takes you somewhere that defines it.</note>
+        <note>Obsolete as of Bloom 5.3</note>
       </trans-unit>
       <trans-unit id="MetadataEditor.linkToPublicDomainCC0">
         <source xml:lang="en">about {0} Public Domain</source>
         <note>ID: MetadataEditor.linkToPublicDomainCC0</note>
         <note>{0} is replaced by an untranslatable abbreviation functioning as an adjective ("CC0")</note>
+        <note>Obsolete as of Bloom 5.3</note>
       </trans-unit>
       <trans-unit id="MetadataEditor.linkToWarningAboutRefiningCreativeCommons">
         <source xml:lang="en">Not Enforceable</source>
         <note>ID: MetadataEditor.linkToWarningAboutRefiningCreativeCommons</note>
         <note>While this screen allows it, legally you can't enforce any additions to a creative commons license. This link takes you to the clause that says that. This link is visible only if you choose Creative Commons but then type in  the "Additional Request" box.</note>
+        <note>Obsolete as of Bloom 5.3</note>
       </trans-unit>
       <trans-unit id="ScriptsAndVariants.ScriptsAndVariantsDialog._cancelButton">
         <source xml:lang="en">Cancel</source>
--- a/src/BloomBrowserUI/.storybook/main.js
+++ b/src/BloomBrowserUI/.storybook/main.js
@@ -1,4 +1,8 @@
 module.exports = {
     stories: ["../**/stories.tsx"],
-    addons: ["@storybook/addon-a11y"]
+    addons: [
+        "@storybook/addon-docs",
+        "@storybook/addon-controls",
+        "@storybook/addon-a11y"
+    ]
 };
--- a/src/BloomBrowserUI/.vscode/settings.json
+++ b/src/BloomBrowserUI/.vscode/settings.json
@@ -18,5 +18,5 @@
         "statusBar.noFolderBackground": "#75416E",
         "statussBar.prominentBackground": "#75416E"
     },
-    "cSpell.words": ["xmatter"]
+    "cSpell.words": ["localizable", "Winform", "Winforms", "xmatter"]
 }
--- a/src/BloomBrowserUI/bloomMaterialUITheme.ts
+++ b/src/BloomBrowserUI/bloomMaterialUITheme.ts
@@ -1,7 +1,7 @@
 import { createTheme } from "@material-ui/core/styles";
 
 export const kBloomBlue = "#1d94a4";
-export const kBloomLightBlue = "#1e94a4"; // review: was this meant to be BloomBlue? see use in BooksOfCollection.less
+export const kBloomBlueLight = "#8ecad2"; // This is BloomBlue at 50% transparency on a white background
 export const kBloomPurple = "#96668f";
 const kDialogTopBottomGray = "#F1F3F4";
 const kGreyOnDarkColor = "#988b8b";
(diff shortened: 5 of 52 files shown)


JohnThomson (changes requested):
Lots of stuff here. I haven't finished reviewing EditingView.cs; having a really hard time figuring out how these changes interact with the new API. But you may as well have what I've written. Some comments may be more appropriately directed to JohnH as questions about the design. Some are just "did you think about this" and don't necessarily require an explanation as long as you did :-). Some, especially the issue with react components modifying their props, may be worth discussing. It's just possible you could convince me it's not worth fixing.

> Reviewed 46 of 47 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5058#-:-Mz2BD58CSJM3-IBxG7O:b-xdpzrz)* status: 46 of 47 files reviewed, 33 unresolved discussions (waiting on @andrew-polk and @JohnThomson)

___
*[DistFiles/localization/en/Bloom.xlf, line 2900 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5058#-Mz0zWoR7Ue-nDiv7QNX:-Mz0zWoR7Ue-nDiv7QNY:ba24nbf) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/edd4b71ea68b22363d6496d990f8e17c8322afbc/DistFiles/localization/en/Bloom.xlf#L2900)):*
> ```XML
>         <note>ID: FontInformationPane.License</note>
>         <note>This shows in the popup as a link to the font's license.</note>
>       </trans-unit>
> ```

Why is this one deleted rather than marked obsolete?
___
*[DistFiles/localization/en/Bloom.xlf, line 3132 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5058#-Mz0zf8k6qe7RHk8xQSU:-Mz0zf8k6qe7RHk8xQSV:bh6nddo) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/08e66e8d3f631ae0d1676bc4e891013d55d3c0bf/DistFiles/localization/en/Bloom.xlf#L3132)):*
> ```XML
>       </trans-unit>
>

JohnThomson (changes requested):
> Reviewed 1 of 47 files at r1, 2 of 12 files at r2.
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5058)* status: 38 of 48 files reviewed, 35 unresolved discussions (waiting on @andrew-polk and @JohnThomson)

___
*[src/BloomExe/Edit/EditingView.cs, line 697 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5058#-Mz27SzfEP4Kf1nIyeMM:-Mz7JcYfCfi97oVBd1eQ:b75wk8x) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/08e66e8d3f631ae0d1676bc4e891013d55d3c0bf/src/BloomExe/Edit/EditingView.cs#L697)):*

This name doesn't at all suggest to me launching a dialog. It suggests doing something with data we got back from the dialog. Maybe LaunchCopyrightAndLicenseDialogForMetadataFromImageToolbox? Kind of long, and my brain is getting tired.
</blockquote></details>

More than that, it's almost impossible to understand what's going on here. I could not figure it out even after pulling the commit so I could study the full code in context and find callers. Apparently this is some kind of callback that is passed to the LibPalaso ImageTooboxDialog. Maybe it would make more sense after a careful study of how ImageToolboxDialog uses the callback and what arguments it passes. For some reason the callback is passed originalMetaData and an action for setting new metadata. I'm guessing it needs a comment something lke,
"This function is called by the ImageTooboxDialog after the user chooses an image. The image toolbox remains open. After taking appropriate action to determine the metadata for the image...in our case by launching a dialog...this function may call setNewImageMetaData to have the ImageToolBox save the new metadata in the usual way"

That raises more questions. Is this only called for an impge that has no metadata? No, then there would be no originalMetaData. So are we now going to launch this dialog every time? Or is this only called when the user chooses to edit metadata by pressing a button in the ImageToolboxDialog?
What happens if the user closes the ImageToolboxDialog before calling setNewImageMetaData? (Or what prevents that?). 
___
*[src/BloomExe/Edit/EditingView.cs, line 696 at r2](https://reviewable.io/reviews/bloombooks/bloomdesktop/5058#-Mz7M1dU0DHw9qredpFA:-Mz7M1dU0DHw9qredpFB:b-gsf9wl) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/f8c1c06ecd8961a33ddff851f3fb53101218b31c/src/BloomExe/Edit/EditingView.cs#L696)):*
> ```Smalltalk
> 
> 		private Metadata _originalImageMetadata;
> 		private Action<Metadata> _setNewImageMetadataActionForImageToolbox;
> ```

Comment like, "When the coyright and license dialog is launched as part of choosing an image, this action will be set as the thing to do to save the metadata (into the image) when the user confirms changes in the CL dialog"
___
*[src/BloomExe/Edit/EditingView.cs, line 1237 at r2](https://reviewable.io/reviews/bloombooks/bloomdesktop/5058#-Mz7MzSb2IZIzjViVTtX:-Mz7MzSb2IZIzjViVTtY:b-t3yr8u) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/f8c1c06ecd8961a33ddff851f3fb53101218b31c/src/BloomExe/Edit/EditingView.cs#L1237)):*
> ```Smalltalk
> 
> 				_originalImageMetadata = null;
> 				_setNewImageMetadataActionForImageToolbox = null;
> ```

I'm having a hard time satisfying myself that this is adequate. It's not in a finally. My intuition is that it ought to happen right after the ImageToolboxDialog's ShowDialog() call...and maybe something is needed to make sure it happens even if something goes wrong while showing that dialog? We really don't want this still in place if the user later goes to work on the book metadata.

Here's another idea: does it help at all if instead of making ProcessImageMetadataFromImageToolbox a member function of this class, we make it a private Action object inside the OnChangeImage function? Then the action would have direct access to variables of that function, and perhaps we could somehow make sure that image data is saved this way only if the ImageTooboxDialog is running?

andrew-polk (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5058)* status: 33 of 48 files reviewed, 35 unresolved discussions (waiting on @andrew-polk, @hatton, and @JohnThomson)

___
*[src/BloomBrowserUI/bookEdit/copyrightAndLicense/CopyrightPanel.tsx, line 70 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5058#-Mz1GZfpAkeZZMeildi7:-MzGsmXR0TVFGMZQJDRb:b-6pu0o5) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/08e66e8d3f631ae0d1676bc4e891013d55d3c0bf/src/BloomBrowserUI/bookEdit/copyrightAndLicense/CopyrightPanel.tsx#L70)):*

Should we demand all caps? There may be cultures where Sil is a common name.
Would holder ever be given as something like an email address where finding sil.org would be significant?
Are combinations like SILPNG or SilAust or SIL/RISE ever copyright holders?
</blockquote></details>

Your questions prompted me to just go back to JH's statement in the Figma design, "When the copyright contains 'SIL'".
___
*[src/BloomBrowserUI/bookEdit/copyrightAndLicense/CopyrightPanel.tsx, line 94 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5058#-Mz1I14w9yb_WB8NrYAS:-MzGtZJV7XO94QCz82hI:b-896fix) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/08e66e8d3f631ae0d1676bc4e891013d55d3c0bf/src/BloomBrowserUI/bookEdit/copyrightAndLicense/CopyrightPanel.tsx#L94)):*

Shouldn't the question marks here be exclamations? You can't be certain that originalCopyrightYear will be defined if the object it comes from might not be.
</blockquote></details>

Done.
___
*[src/BloomBrowserUI/bookEdit/copyrightAndLicense/CopyrightPanel.tsx, line 102 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5058#-Mz1J0ZU4Lq9uaYEPa6o:-MzHAtrFDpEhzItSz8_R:b-u5myda) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/08e66e8d3f631ae0d1676bc4e891013d55d3c0bf/src/BloomBrowserUI/bookEdit/copyrightAndLicense/CopyrightPanel.tsx#L102)):*

Worth explaining why this should be used rather than margin?
</blockquote></details>

Improved. Thanks.
___
*[src/BloomBrowserUI/bookEdit/copyrightAndLicense/CopyrightPanel.tsx, line 120 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5058#-Mz1IZfs67XjOv5Piu2I:-MzHBDJsBs_wzrsXUVW9:be8cetr) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/08e66e8d3f631ae0d1676bc4e891013d55d3c0bf/src/BloomBrowserUI/bookEdit/copyrightAndLicense/CopyrightPanel.tsx#L120)):*

I find it vaguely surprising that we want to know the 'author' of an image but not of a book. Is it because we have the book's author recorded elsewhere?
</blockquote></details>

Not sure. That was the previous (and so far current) design. 
@hatton
___
*[src/BloomBrowserUI/bookEdit/copyrightAndLicense/CopyrightPanel.tsx, line 186 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5058#-Mz1JK-vA9Oz9J3P3Uaa:-MzHBK293pPrHZUWkOaJ:b-6n0v2l) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/08e66e8d3f631ae0d1676bc4e891013d55d3c0bf/src/BloomBrowserUI/bookEdit/copyrightAndLicense/CopyrightPanel.tsx#L186)):*

Isn't 'publisher' something distinct from copyright holder?
</blockquote></details>

@hatton  
___
*[src/BloomBrowserUI/bookEdit/copyrightAndLicense/LicensePanel.tsx, line 190 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5058#-Mz1iWYv97emJ-dkQWTO:-MzHCNZREGRrKN2pbMrs:b-896fix) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/08e66e8d3f631ae0d1676bc4e891013d55d3c0bf/src/BloomBrowserUI/bookEdit/copyrightAndLicense/LicensePanel.tsx#L190)):*

Maybe a comment like, These two check boxes govern the allowDerivatives value. Both apply only to CC licenses (enableCheckboxes true). The first determines whether derivatives are allowed at all. If so, the second is enabled and determines whether allowDerivaties should be "sa" or simply "yes"
(It took me several minutes to figure that out, so someone else might appreciate it.)
</blockquote></details>

Done.
___
*[src/BloomBrowserUI/bookEdit/js/BloomHintBubbles.ts, line 439 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5058#-Mz1oHN-C9eqSsmYS3n2:-MzHE1QMA0RVqKox9lpy:b-896fix) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/08e66e8d3f631ae0d1676bc4e891013d55d3c0bf/src/BloomBrowserUI/bookEdit/js/BloomHintBubbles.ts#L439)):*

I get that in some circumstance (presence of data-functionhintclick on some element?) we want an onclick on the href. But I have no idea what it means to set href to javascript: clickWasAlreadyHandled rather than bookMetaDataHandler. I think at least a brief comment is indicated.
</blockquote></details>

Done.

hatton (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5058)* status: 33 of 48 files reviewed, 35 unresolved discussions (waiting on @andrew-polk, @hatton, and @JohnThomson)

___
*[src/BloomBrowserUI/bookEdit/copyrightAndLicense/CopyrightPanel.tsx, line 186 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5058#-Mz1JK-vA9Oz9J3P3Uaa:-MzHOrVx9vWiGxFlqm2Q:b-u3trhw) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/08e66e8d3f631ae0d1676bc4e891013d55d3c0bf/src/BloomBrowserUI/bookEdit/copyrightAndLicense/CopyrightPanel.tsx#L186)):*

@hatton  
</blockquote></details>

@johnthomson  Do you have an example?

andrew-polk (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5058)* status: 33 of 48 files reviewed, 35 unresolved discussions (waiting on @andrew-polk, @hatton, and @JohnThomson)

___
*[src/BloomBrowserUI/react_components/BloomDialog/commonDialogComponents.tsx, line 213 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5058#-Mz21_wf3hF7Q_Gg_vTZ:-MzHUaMZ6IebgbcNHZnP:bcvjgzp) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/08e66e8d3f631ae0d1676bc4e891013d55d3c0bf/src/BloomBrowserUI/react_components/BloomDialog/commonDialogComponents.tsx#L213)):*

Adding '80' (which makes it 50% transparent, provided kBloomBlue continues to be done with a six-character hex code) worked in the one context where I originally did it because it had a near-white background. Even for a one-off I was a bit dubious...it just seemed to be what made my display look most like John's mockup. But it was a rather risky hack even there, which I now think I regret. It would break, for example, if we changed kbloomBlue to a 3-character hex or an rgb macro. How sure are you that this will work well everywhere a NoteBox might be used?
Would we want it to be the same final color if the Notebox was on a dark background, or do we really want transparency to make it less contrastive?
In either case, I think it might be better to define a new color, something like kbloomBlueLight or kBloomBlueLowContrast (depending on whether you want the exact color or the transparency).
(Sorry, this started as my bad.)
</blockquote></details>

Good idea. Done.
___
*[src/BloomBrowserUI/react_components/BloomDialog/commonDialogComponents.tsx, line 230 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5058#-Mz23vYr-eNJuyks0V5m:-MzHQt114N1i4vQrDeIv:b-896fix) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/08e66e8d3f631ae0d1676bc4e891013d55d3c0bf/src/BloomBrowserUI/react_components/BloomDialog/commonDialogComponents.tsx#L230)):*

Now that we have a prop of our own that div will not recognize, we should deconstruct and strip it out and propagate only the other props to avoid adding to our collection of console warning messages.
</blockquote></details>

Done.

andrew-polk (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5058)* status: 28 of 50 files reviewed, 35 unresolved discussions (waiting on @andrew-polk, @hatton, and @JohnThomson)

___
*[src/BloomBrowserUI/bookEdit/copyrightAndLicense/CopyrightAndLicenseDialog.tsx, line 228 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5058#-Mz18RSG70sW43mAZ5zG:-MzHrzmA1MxJenTvATAV:b3a8ret) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/08e66e8d3f631ae0d1676bc4e891013d55d3c0bf/src/BloomBrowserUI/bookEdit/copyrightAndLicense/CopyrightAndLicenseDialog.tsx#L228)):*

Could there be anything in imageUrl that requires encodeUriComponent?
</blockquote></details>

I've convinced myself no. Comment added.
___
*[src/BloomBrowserUI/bookEdit/copyrightAndLicense/CopyrightAndLicenseDialog.tsx, line 230 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5058#-Mz19liB5HUBIGL4aPAy:-MzHuutSEbiPTXbfGoAn:bqybgio) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/08e66e8d3f631ae0d1676bc4e891013d55d3c0bf/src/BloomBrowserUI/bookEdit/copyrightAndLicense/CopyrightAndLicenseDialog.tsx#L230)):*

This feels vaguely smelly. At a low level, if find myself wondering if there isn't some way to find out whether we ought to show the dialog without stringifying the whole result.data when we do. At a higher level, it's weird that half the UI is still in C# (for showing the info if we can't edit). Is that planned to change later? I also wonder about the 'OrDialog' in the name. We're going to unconditionally show copyright and license info, right? It just might or might not be in a dialog that allows it to be edited.
</blockquote></details>

Looks like the stringify wasn't needed in the end (at one point I was seeing that result).

There are multiple messages which can come back from C#. I didn't take on replacing those messages as part of this task. I'm open to further discussion, though.
___
*[src/BloomBrowserUI/bookEdit/copyrightAndLicense/CopyrightAndLicenseDialog.tsx, line 251 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5058#-Mz1BVQbDk94nuC1Uv3P:-MzHoxD60eyt06rzX6wU:buw7kxc) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/08e66e8d3f631ae0d1676bc4e891013d55d3c0bf/src/BloomBrowserUI/bookEdit/copyrightAndLicense/CopyrightAndLicenseDialog.tsx#L251)):*

It feels smelly...maybe unavoidable at this point, but smelly...that this dialog knows it displays in the context of the EditTab document. One day, we might decide that you can run this right from the Upload page if you forgot to set it while editing. Be much nicer to just follow John's new pattern and insert a CopyrightAndLicenceDialogLauncher element anywhere you want to be able to show one. Maybe that can't happen until the top level of the Edit pane is a react component.
Do we even need to go through getEditTabBundleExports().getDocument()? When would this be called that the current document would not be the right one?
</blockquote></details>

Right, we can't do it the other way until Edit tab is react at the root, unfortunately.
When we launch from the image IP button, the edit frame document and our document are not the same. Oddly, when launching from the book copyright hint bubble, they are.
___
*[src/BloomBrowserUI/bookEdit/copyrightAndLicense/CopyrightPanel.tsx, line 54 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5058#-Mz1Ed6H0Lv7CntH2clS:-MzHvQFr4C4YFGcEkqKa:bh8s21u) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/08e66e8d3f631ae0d1676bc4e891013d55d3c0bf/src/BloomBrowserUI/bookEdit/copyrightAndLicense/CopyrightPanel.tsx#L54)):*

If feels wrong for a react component to modify its own props. They are supposed to be how its parent controls IT. I think the usual strategy is to pass an onChange handler if something in the child should modify something belonging to the parent. Could something like that work here? (Also elsewhere)
</blockquote></details>

Ok, I've cleaned up setting of props. Now I just need to get rid of the force render hack at the top level.
___
*[src/BloomExe/web/controllers/CopyrightAndLicenseApi.cs, line 40 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5058#-Mz28lSd1HG-xRO0vDha:-MzHgYEHEIPJf1RjpukR:boyexbg) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/08e66e8d3f631ae0d1676bc4e891013d55d3c0bf/src/BloomExe/web/controllers/CopyrightAndLicenseApi.cs#L40)):*

IIRC, this method sets response headers based on the file extension. You should probably use TempFile.WithExtension() to get one that has the right one.
</blockquote></details>

Done. Thanks.
___
*[src/BloomExe/web/controllers/CopyrightAndLicenseApi.cs, line 46 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5058#-Mz293YJDTmou8_QZmU0:-MzHgV1p3HrA9J_kw4vg:b-cghggh) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/08e66e8d3f631ae0d1676bc4e891013d55d3c0bf/src/BloomExe/web/controllers/CopyrightAndLicenseApi.cs#L46)):*

Log? Sentry?
</blockquote></details>

Done. Doesn't seem worth a sentry call.

andrew-polk (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5058)* status: 24 of 50 files reviewed, 35 unresolved discussions (waiting on @andrew-polk, @hatton, and @JohnThomson)

___
*[src/BloomBrowserUI/bookEdit/copyrightAndLicense/CopyrightAndLicenseDialog.tsx, line 61 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5058#-Mz11p_cDgKQFRplAPrb:-MzHvhrVEHoWmk-AoJ0m:bjtd5pp) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/08e66e8d3f631ae0d1676bc4e891013d55d3c0bf/src/BloomBrowserUI/bookEdit/copyrightAndLicense/CopyrightAndLicenseDialog.tsx#L61)):*

Needs more. It's not obvious, perhaps because it's not immediately obvious that show and showDialog are functions, how this helps to provide external ability to show the dialog. Maybe something like, configure the local function for showing the dialog to be the one derived from our dialogEnvironment which allows the ultimate creator of the dialog to make it visible? Assuming that's accurate.
(Of course this may change. Do we even still need the ability to run from Winforms? Of course edit view isn't all React, so JohnH's new approach may not work.
</blockquote></details>

I've had a go. See what you think.

Yeah, unfortunately his new thing won't work until edit is all react.
___
*[src/BloomBrowserUI/bookEdit/copyrightAndLicense/CopyrightPanel.tsx, line 186 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5058#-Mz1JK-vA9Oz9J3P3Uaa:-MzHxAud1ufvR1ognR_N:b5a9f1d) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/08e66e8d3f631ae0d1676bc4e891013d55d3c0bf/src/BloomBrowserUI/bookEdit/copyrightAndLicense/CopyrightPanel.tsx#L186)):*

@johnthomson  Do you have an example?
</blockquote></details>

We have them as two separate fields in parse.
A few examples (some of which may be bogus) are:

**publisher -- copyright holder**
SIL-PNG -- Education for Life
Chetana -- SIL International
Little Zebra Books	-- African Storybook Project
African Storybook Initiative	-- Yayasan Sulinama
___
*[src/BloomBrowserUI/bookEdit/toolbox/overlay/overlayTool.tsx, line 601 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5058#-Mz1pO3PE6XmaxG7FXfU:-MzHz5uo06ctqiJzVhLp:bpcm5i0) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/08e66e8d3f631ae0d1676bc4e891013d55d3c0bf/src/BloomBrowserUI/bookEdit/toolbox/overlay/overlayTool.tsx#L601)):*

Deserves comment. Surely the version where wrapped labels might not work has some advantage we care about to compensate?
</blockquote></details>

I was hoping the comment in the component would suffice. But open to pushback.
___
*[src/BloomBrowserUI/react_components/muiCheckBox.tsx, line 26 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5058#-Mz1rJGFBdrfumQIVjoG:-MzHy_7HEtBheBb8dT2_:bkg0f8c) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/08e66e8d3f631ae0d1676bc4e891013d55d3c0bf/src/BloomBrowserUI/react_components/muiCheckBox.tsx#L26)):*

(a) can you clarify what a wrapped label is, so a client knows definitely NOT to set this if he has one?
(b) any way to assert or throw or something if the label IS wrapped and this is set?
(c) would I ever want this on a new use of muiCheckbox? If so, how would I know? Or would a new case be grounds for new investigation with a goal of getting rid of this?
(d) could the name possibly indicate the positive side? I'd rather see legacyVersionWhichHandlesX[ButNotWrappedLabels]
</blockquote></details>

(a) is "multiline" better?
(b) not that I know of
(c) any new ones shouldn't use it; the dev will be making the layout already taking into account the extra div, so there will be no problem. I change the name to deprecated; does that help? (Note, I made the default to be false so new callers wouldn't need to know about it.)
(d) there is no positive side

hatton (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5058)* status: 24 of 50 files reviewed, 35 unresolved discussions (waiting on @andrew-polk, @hatton, and @JohnThomson)

___
*[src/BloomBrowserUI/bookEdit/copyrightAndLicense/CopyrightPanel.tsx, line 186 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5058#-Mz1JK-vA9Oz9J3P3Uaa:-MzL22uL5CSvHoVEfaCC:b-n3mezt) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/08e66e8d3f631ae0d1676bc4e891013d55d3c0bf/src/BloomBrowserUI/bookEdit/copyrightAndLicense/CopyrightPanel.tsx#L186)):*

We have them as two separate fields in parse.
A few examples (some of which may be bogus) are:

**publisher -- copyright holder**
SIL-PNG -- Education for Life
Chetana -- SIL International
Little Zebra Books	-- African Storybook Project
African Storybook Initiative	-- Yayasan Sulinama
</blockquote></details>

That strengthens my suspicion around asking people to separate these concepts. It makes zero sense for EFL to be the copyright holder and SIL-PNG the publisher. Little sense for Chetana / SIL. And the Zebra one... that's likely backwards, since ASP republishes Zebra books on their platform. But I'd need a pointer to the book to be sure.

andrew-polk (commented):
Ok, @johnthomson, I think I've addressed everything. At least enough for another look.

> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5058#-:-MzML_w8C-L2DDVcZZfK:bmow5kd)* status: 23 of 52 files reviewed, 35 unresolved discussions (waiting on @hatton and @JohnThomson)

___
*[src/BloomExe/Edit/EditingView.cs, line 697 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5058#-Mz27SzfEP4Kf1nIyeMM:-MzMKnHX09Ud4FqQl6D3:b3umvra) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/08e66e8d3f631ae0d1676bc4e891013d55d3c0bf/src/BloomExe/Edit/EditingView.cs#L697)):*

More than that, it's almost impossible to understand what's going on here. I could not figure it out even after pulling the commit so I could study the full code in context and find callers. Apparently this is some kind of callback that is passed to the LibPalaso ImageTooboxDialog. Maybe it would make more sense after a careful study of how ImageToolboxDialog uses the callback and what arguments it passes. For some reason the callback is passed originalMetaData and an action for setting new metadata. I'm guessing it needs a comment something lke,
"This function is called by the ImageTooboxDialog after the user chooses an image. The image toolbox remains open. After taking appropriate action to determine the metadata for the image...in our case by launching a dialog...this function may call setNewImageMetaData to have the ImageToolBox save the new metadata in the usual way"

That raises more questions. Is this only called for an impge that has no metadata? No, then there would be no originalMetaData. So are we now going to launch this dialog every time? Or is this only called when the user chooses to edit metadata by pressing a button in the ImageToolboxDialog?
What happens if the user closes the ImageToolboxDialog before calling setNewImageMetaData? (Or what prevents that?). 
</blockquote></details>

I've attempted to change some names and add some comments to address these things. See what you think.
___
*[src/BloomExe/Edit/EditingView.cs, line 696 at r2](https://reviewable.io/reviews/bloombooks/bloomdesktop/5058#-Mz7M1dU0DHw9qredpFA:-MzMKtU-9Gzp3_BPzVN-:b-896fix) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/f8c1c06ecd8961a33ddff851f3fb53101218b31c/src/BloomExe/Edit/EditingView.cs#L696)):*

Comment like, "When the coyright and license dialog is launched as part of choosing an image, this action will be set as the thing to do to save the metadata (into the image) when the user confirms changes in the CL dialog"
</blockquote></details>

Done.
___
*[src/BloomExe/Edit/EditingView.cs, line 1237 at r2](https://reviewable.io/reviews/bloombooks/bloomdesktop/5058#-Mz7MzSb2IZIzjViVTtX:-MzMKxzT6y9gu4PpB2D4:bxultuo) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/f8c1c06ecd8961a33ddff851f3fb53101218b31c/src/BloomExe/Edit/EditingView.cs#L1237)):*

I'm having a hard time satisfying myself that this is adequate. It's not in a finally. My intuition is that it ought to happen right after the ImageToolboxDialog's ShowDialog() call...and maybe something is needed to make sure it happens even if something goes wrong while showing that dialog? We really don't want this still in place if the user later goes to work on the book metadata.

Here's another idea: does it help at all if instead of making ProcessImageMetadataFromImageToolbox a member function of this class, we make it a private Action object inside the OnChangeImage function? Then the action would have direct access to variables of that function, and perhaps we could somehow make sure that image data is saved this way only if the ImageTooboxDialog is running?
</blockquote></details>

See what you think after the changes I made and the discussion we had.

JohnThomson (changes requested):
Much improved. A few things I think are still worth a little thought, but nothing major.

> Reviewed 4 of 12 files at r2, 1 of 3 files at r4, 4 of 6 files at r5, 4 of 6 files at r6, 6 of 7 files at r7, 5 of 7 files at r8, 2 of 2 files at r9, 5 of 5 files at r10.
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5058#-:-MzMjsz63q53eqbmqURE:b-vs7wvo)* status: 49 of 52 files reviewed, 6 unresolved discussions (waiting on @andrew-polk and @JohnThomson)

___
*[src/BloomBrowserUI/bookEdit/copyrightAndLicense/CopyrightAndLicenseDialog.tsx, line 251 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5058#-Mz1BVQbDk94nuC1Uv3P:-MzMV6jt9EjVVMJE5Ufi:bgrobmz) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/08e66e8d3f631ae0d1676bc4e891013d55d3c0bf/src/BloomBrowserUI/bookEdit/copyrightAndLicense/CopyrightAndLicenseDialog.tsx#L251)):*

Right, we can't do it the other way until Edit tab is react at the root, unfortunately.
When we launch from the image IP button, the edit frame document and our document are not the same. Oddly, when launching from the book copyright hint bubble, they are.
</blockquote></details>

Might be worth capturing some of that in a comment...particularly an example where the current document is NOT the edit tab document, and also that change will be needed if we ever need to show this when the Edit tab is not active.
___
*[src/BloomBrowserUI/bookEdit/copyrightAndLicense/CopyrightPanel.tsx, line 65 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5058#-Mz1FMyWDILDYBH3fzEn:-MzMVelN-xO7qkwhOu6d:bw60p3z) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/08e66e8d3f631ae0d1676bc4e891013d55d3c0bf/src/BloomBrowserUI/bookEdit/copyrightAndLicense/CopyrightPanel.tsx#L65)):*

Actually, the main point is to make blank invalid. I've just made it do what the existing dialog does now. (any 4-digit number)
</blockquote></details>

That will make things difficult if the document is the original Greek of any of the New Testament or some BC document :-) But then, it would be controversial to give a copyright year for those anyway, since they predate all copyright laws.
___
*[src/BloomBrowserUI/bookEdit/copyrightAndLicense/CopyrightPanel.tsx, line 186 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5058#-Mz1JK-vA9Oz9J3P3Uaa:-MzMaLHhAu9pDCmwRXf-:bx2y0fo) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/08e66e8d3f631ae0d1676bc4e891013d55d3c0bf/src/BloomBrowserUI/bookEdit/copyrightAndLicense/CopyrightPanel.tsx#L186)):*

That strengthens my suspicion around asking people to separate these concepts. It makes zero sense for EFL to be the copyright holder and SIL-PNG the publisher. Little sense for Chetana / SIL. And the Zebra one... that's likely backwards, since ASP republishes Zebra books on their platform. But I'd need a pointer to the book to be sure.
</blockquote></details>

You could talk to someone like Alan Connor. I remember a talk he gave once. IIRC, the distinction is that the copyright holder is (of course) the person who owns the document, controlling what may be done with it. The publisher is the person who takes responsibility for making this content public in a particular way. So, for example, YouVersion and other DBL subscribers may publish a Bible which they are not at all the owner of, and the author of a bloom book is often publishing images of which they are not the copyright holder. I _think_ I recall that SIL is forbidden from being the publisher of Scripture; I don't think we are forbidden from owning it. (Indeed, we hardly could be, since we are automatically the copyright owner of something we create until and unless we assign it to someone else). The same document may have different publishers in different contexts...the copyright holder might license one online publisher and another paper publisher, or even one publisher in the UK and another in the USA. Sometimes the original copyright holder assigns all rights to a publisher, making the two effectively identical, but this is not automatic, nor necessarily a good idea. It is probably even possible to have a split copyright holder, where (for example) someone owns all rights to digital publication and someone else owns all rights to paper publication.
I'm not sure it matters in this context, and therefore not sure why I was worried. If the copyright holder is SIL, they need to be careful about SIL guidelines for publishing.
___
*[src/BloomBrowserUI/bookEdit/copyrightAndLicense/CopyrightPanel.tsx, line 84 at r10](https://reviewable.io/reviews/bloombooks/bloomdesktop/5058#-MzMZg8aDGkrYX2ih6hO:-MzMZg8aDGkrYX2ih6hP:bljh875) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/247cfbeb662d311a5ebe838c1e2db2f941078df7/src/BloomBrowserUI/bookEdit/copyrightAndLicense/CopyrightPanel.tsx#L84)):*
> ```XML
> 
>     useEffect(() => {
>         copyrightInfo.imageCreator = imageCreator?.trim();
> ```

copyrightInfo is a local variable, made to be a clone so it doesn't share anything with any object in state or props or any other render. Modifying it in a useEffect strikes me as highly suspicious. It's just possible that the instance captured by this function that it is modifying is the same instance that, say, the next useEffect passes to props.onChange, but it strikes me as a risky thing to depend on. How about just doing all the trimming at the point where you send copyrightInfo to onChange()?
___
*[src/BloomBrowserUI/bookEdit/copyrightAndLicense/CopyrightPanel.tsx, line 100 at r10](https://reviewable.io/reviews/bloombooks/bloomdesktop/5058#-MzMXtnZ3TOaIQ7tcYj2:-MzMXtnZ3TOaIQ7tcYj3:b-sdn61x) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/247cfbeb662d311a5ebe838c1e2db2f941078df7/src/BloomBrowserUI/bookEdit/copyrightAndLicense/CopyrightPanel.tsx#L100)):*
> ```XML
>         isHolderValid,
>         useOriginalCopyrightAndLicense
>     ]);
> ```

It vaguely bothers me that this onChange will run unnecessarily (I think) on the first render. This would be especially bad if the initial values were defaults just used until we get a real value from the server, or if there were costly side effects of sending the message (such as forcing a preview to regenerate) but I don't think that's the case here. The most likely alternative would be to have wrapper functions for the setCopyright etc functions, which call the state setter but also call a function that handles the props.OnChange call. I'm not sure whether that's a worthwhile change.
___
*[src/BloomBrowserUI/react_components/muiCheckBox.tsx, line 26 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5058#-Mz1rJGFBdrfumQIVjoG:-MzMf5vj53IfPD52ok1G:bw286ed) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/08e66e8d3f631ae0d1676bc4e891013d55d3c0bf/src/BloomBrowserUI/react_components/muiCheckBox.tsx#L26)):*

(a) is "multiline" better?
(b) not that I know of
(c) any new ones shouldn't use it; the dev will be making the layout already taking into account the extra div, so there will be no problem. I change the name to deprecated; does that help? (Note, I made the default to be false so new callers wouldn't need to know about it.)
(d) there is no positive side
</blockquote></details>

Fair enough. Thanks for making it clearer
___
*[src/BloomExe/Edit/EditingView.cs, line 697 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5058#-Mz27SzfEP4Kf1nIyeMM:-MzMgPYw4ftAUP3gapfS:b-n8y2pa) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/08e66e8d3f631ae0d1676bc4e891013d55d3c0bf/src/BloomExe/Edit/EditingView.cs#L697)):*

I've attempted to change some names and add some comments to address these things. See what you think.
</blockquote></details>

I think it's much improved. The real test would be whether someone who hadn't talked it over with you could follow. But I don't see a way to improve it now. It's something of an inherently messy situation, using a library dialog and supplying an action to configure its behavior.
___
*[src/BloomExe/Edit/EditingView.cs, line 1237 at r2](https://reviewable.io/reviews/bloombooks/bloomdesktop/5058#-Mz7MzSb2IZIzjViVTtX:-MzMhS2eAxBRGpCy3_kv:b-4n67xa) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/f8c1c06ecd8961a33ddff851f3fb53101218b31c/src/BloomExe/Edit/EditingView.cs#L1237)):*

See what you think after the changes I made and the discussion we had.
</blockquote></details>

I think you're right but it's still not clear why we postpone it so long. These variables only get set and are only used during a callback that happens while the dialog is open, right? So could we have this, which would be more obviously safe?
try {
result = dlg.ShowDialog();
} finally {
    // These variables get set during a callback from the dialog that allows us to use our own way of
    // editing metadata for an image. It's important that they get cleaned up once the dialog is closed
    // so they don't interfere with future uses of the metadata editing code.
    _originalImageMetadataFromImageToolbox = null;
   _saveNewImageMetadataActionForImageToolbox = null;
}
Of course that assumes they're not used in any of the intermediate code. But I don't see any reason for them to be.
___
*[src/BloomExe/Edit/EditingView.cs, line 696 at r10](https://reviewable.io/reviews/bloombooks/bloomdesktop/5058#-MzMfgX78fFFThvAP5eW:-MzMfgX8EB9xA3eYD4N5:b-4hzk7g) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/247cfbeb662d311a5ebe838c1e2db2f941078df7/src/BloomExe/Edit/EditingView.cs#L696)):*
> ```Smalltalk
> 
> 		private Metadata _originalImageMetadataFromImageToolbox;
> 		// When the coyright and license dialog is launched from the palaso image toolbox,
> ```

coyright => copyright
___
*[src/BloomExe/Edit/EditingView.cs, line 710 at r10](https://reviewable.io/reviews/bloombooks/bloomdesktop/5058#-MzMlAY156yXlyN8ue_w:-MzMlAY156yXlyN8ue_x:b-utiz76) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/247cfbeb662d311a5ebe838c1e2db2f941078df7/src/BloomExe/Edit/EditingView.cs#L710)):*
> ```Smalltalk
> 		// But we couldn't convince ourselves with 100% certainty that the save metadata api
> 		// call would finish before the close dialog api call.
> 		private void LaunchCopyrightAndLicenseDialogFromImageToolbox(Metadata originalMetadata, Action<Metadata> saveNewImageMetadata)
> ```

This is much improved. I think my intuition is that it might be even better to inline it, that is, instead of declaring even a private function, just declare these three lines to be an action at the place where you need to pass it to the dialog. Then it's immediately clearer what it's for and there is no temptation to use it for anything else. The comment could usefully go along, of course. But I won't insist...if you think it's better like this you may well be right.

andrew-polk (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5058)* status: 49 of 52 files reviewed, 6 unresolved discussions (waiting on @andrew-polk and @JohnThomson)

___
*[src/BloomBrowserUI/bookEdit/copyrightAndLicense/CopyrightAndLicenseDialog.tsx, line 251 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5058#-Mz1BVQbDk94nuC1Uv3P:-MzMpVol3qb730BH1lYl:bnqzdsy) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/08e66e8d3f631ae0d1676bc4e891013d55d3c0bf/src/BloomBrowserUI/bookEdit/copyrightAndLicense/CopyrightAndLicenseDialog.tsx#L251)):*

Might be worth capturing some of that in a comment...particularly an example where the current document is NOT the edit tab document, and also that change will be needed if we ever need to show this when the Edit tab is not active.
</blockquote></details>

Turns out that launching it for the images was happening a different way which was causing the modal to not cover the toolbox (it ended up being in just the page iframe). I had fixed that anyway, so now it is consistent here and we can just use document.
___
*[src/BloomBrowserUI/bookEdit/copyrightAndLicense/CopyrightPanel.tsx, line 84 at r10](https://reviewable.io/reviews/bloombooks/bloomdesktop/5058#-MzMZg8aDGkrYX2ih6hO:-MzMtz2L8R4SoteMHT_t:b-896fix) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/247cfbeb662d311a5ebe838c1e2db2f941078df7/src/BloomBrowserUI/bookEdit/copyrightAndLicense/CopyrightPanel.tsx#L84)):*

copyrightInfo is a local variable, made to be a clone so it doesn't share anything with any object in state or props or any other render. Modifying it in a useEffect strikes me as highly suspicious. It's just possible that the instance captured by this function that it is modifying is the same instance that, say, the next useEffect passes to props.onChange, but it strikes me as a risky thing to depend on. How about just doing all the trimming at the point where you send copyrightInfo to onChange()?
</blockquote></details>

Done.
___
*[src/BloomExe/Edit/EditingView.cs, line 1237 at r2](https://reviewable.io/reviews/bloombooks/bloomdesktop/5058#-Mz7MzSb2IZIzjViVTtX:-MzMrn5qDcJ5829gy3C5:b-qu6n94) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/f8c1c06ecd8961a33ddff851f3fb53101218b31c/src/BloomExe/Edit/EditingView.cs#L1237)):*

I think you're right but it's still not clear why we postpone it so long. These variables only get set and are only used during a callback that happens while the dialog is open, right? So could we have this, which would be more obviously safe?
try {
result = dlg.ShowDialog();
} finally {
    // These variables get set during a callback from the dialog that allows us to use our own way of
    // editing metadata for an image. It's important that they get cleaned up once the dialog is closed
    // so they don't interfere with future uses of the metadata editing code.
    _originalImageMetadataFromImageToolbox = null;
   _saveNewImageMetadataActionForImageToolbox = null;
}
Of course that assumes they're not used in any of the intermediate code. But I don't see any reason for them to be.
</blockquote></details>

Good point. Done.
___
*[src/BloomExe/Edit/EditingView.cs, line 696 at r10](https://reviewable.io/reviews/bloombooks/bloomdesktop/5058#-MzMfgX78fFFThvAP5eW:-MzMrdBE4CM9MpaKKZzU:b-896fix) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/247cfbeb662d311a5ebe838c1e2db2f941078df7/src/BloomExe/Edit/EditingView.cs#L696)):*

coyright => copyright
</blockquote></details>

Done.
___
*[src/BloomExe/Edit/EditingView.cs, line 710 at r10](https://reviewable.io/reviews/bloombooks/bloomdesktop/5058#-MzMlAY156yXlyN8ue_w:-MzMrjaq9gsGp_r1DAO7:b-fflhqt) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/247cfbeb662d311a5ebe838c1e2db2f941078df7/src/BloomExe/Edit/EditingView.cs#L710)):*

This is much improved. I think my intuition is that it might be even better to inline it, that is, instead of declaring even a private function, just declare these three lines to be an action at the place where you need to pass it to the dialog. Then it's immediately clearer what it's for and there is no temptation to use it for anything else. The comment could usefully go along, of course. But I won't insist...if you think it's better like this you may well be right.
</blockquote></details>

Great idea. Done.

andrew-polk (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5058)* status: 44 of 52 files reviewed, 6 unresolved discussions (waiting on @JohnThomson)

___
*[src/BloomBrowserUI/bookEdit/copyrightAndLicense/CopyrightPanel.tsx, line 100 at r10](https://reviewable.io/reviews/bloombooks/bloomdesktop/5058#-MzMXtnZ3TOaIQ7tcYj2:-MzQqXzA2c-NqlK2EIcW:b-brdj7l) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/247cfbeb662d311a5ebe838c1e2db2f941078df7/src/BloomBrowserUI/bookEdit/copyrightAndLicense/CopyrightPanel.tsx#L100)):*

It vaguely bothers me that this onChange will run unnecessarily (I think) on the first render. This would be especially bad if the initial values were defaults just used until we get a real value from the server, or if there were costly side effects of sending the message (such as forcing a preview to regenerate) but I don't think that's the case here. The most likely alternative would be to have wrapper functions for the setCopyright etc functions, which call the state setter but also call a function that handles the props.OnChange call. I'm not sure whether that's a worthwhile change.
</blockquote></details>

Discussed and commented.

JohnThomson (approved):
> Reviewed 2 of 3 files at r11, 4 of 5 files at r12, 2 of 2 files at r13, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5058)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @andrew-polk)

PR #5064, opened 2022-03-28
BL-11061 Extract "content" into its own project

hatton:
This change is [Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5064)


andrew-polk:
The added .css, .html, and .js files seem like they shouldn't be added.

hatton:
Thanks. Found 12 files that looked to be based on source files:

image

hatton:
Done, thx.

hatton:
thanks

PR #5091, opened 2022-04-07
Remove unwanted language-specific CSS rules for publishing (BL-11109)

StephenMcConnel:
This change is [Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5091)


JohnThomson (changes requested):
New method looks unit-testable.

> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5091#-:-N-8P1QGBGBZicvn8s_Q:b-ni24sh)* status: all files reviewed, 1 unresolved discussion (waiting on @StephenMcConnel)

___
*[src/BloomExe/Book/HtmlDom.cs, line 1919 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5091#-N-8N9XI6QCdmh1L0QYb:-N-8N9XJDyHAttYOOE2M:bjhxlwb) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/e3a3dc20c59045fc916688c22cc0f3620cde2e5e/src/BloomExe/Book/HtmlDom.cs#L1919)):*
> ```Smalltalk
> 			var wantedLanguages = new HashSet<string>(languagesToInclude);
> 			var unwantedLanguages = new HashSet<string>();
> 			var matches = Regex.Matches(cssText, "[\\r\\n]([-.\\s\\w]*\\[lang=(['\"])([-a-zA-Z]+)\\2\\]\\s*{[^}]*})");
> ```

Might be easier to read as an @ string so you don't have to double the backslashes. (You can double the one double-quote to get a single one.)
I'm slightly surprised you don't have to do something special to get multi-line matches, but I suppose you've tested where the style rule runs over more than one line? Or else that never happens?

StephenMcConnel (commented):
That's why I wrote 4 new unit tests for it.  :-)

> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5091#-:-N-OSOYSDtHSPfkViaLx:bsqu2v)* status: all files reviewed, 1 unresolved discussion (waiting on @JohnThomson)

___
*[src/BloomExe/Book/HtmlDom.cs, line 1919 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5091#-N-8N9XI6QCdmh1L0QYb:-N-OVDCqDHP0I5gMKtik:bs9zyvs) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/799037200532b1dd037b592171ae850c55c1db26/src/BloomExe/Book/HtmlDom.cs#L1919)):*

Might be easier to read as an @ string so you don't have to double the backslashes. (You can double the one double-quote to get a single one.)
I'm slightly surprised you don't have to do something special to get multi-line matches, but I suppose you've tested where the style rule runs over more than one line? Or else that never happens?
</blockquote></details>

Done.  I simplified the first expression by removing an unnecessary grouping.
The C# regex matches over the entire text given it by default.  If newlines are in the text, it can match over them as well.  This is demonstrated by a couple of the unit tests I added.

JohnThomson (changes requested):
Don't know how I missed that.
I did make a couple of suggestions about the tests now I see them. And one change doesn't seem to have made it to the PR.

> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5091#-:-N-Ozwim5WkFlLNhdRlw:byhjz3z)* status: all files reviewed, 3 unresolved discussions (waiting on @StephenMcConnel)

___
*[src/BloomExe/Book/HtmlDom.cs, line 1919 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5091#-N-8N9XI6QCdmh1L0QYb:-N-OztGO3Wf6_I85D7Ct:b-5u2svs) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/799037200532b1dd037b592171ae850c55c1db26/src/BloomExe/Book/HtmlDom.cs#L1919)):*

Done.  I simplified the first expression by removing an unnecessary grouping.
The C# regex matches over the entire text given it by default.  If newlines are in the text, it can match over them as well.  This is demonstrated by a couple of the unit tests I added.
</blockquote></details>

Looks like the change didn't get committed?
___
*[src/BloomTests/Book/HtmlDomTests.cs, line 1808 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5091#-N-OycxoCCEE3mliRTQk:-N-OycxoCCEE3mliRTQl:b-kbk87a) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/799037200532b1dd037b592171ae850c55c1db26/src/BloomTests/Book/HtmlDomTests.cs#L1808)):*
> ```Smalltalk
> }
> 
> ";
> ```

When I want two very similar strings like this in a test, I like to make a template, and then do something like cssText=string.Format(template, "...enc rule", "...zh-CN rule"); desiredResult = string.Format(template, "", ""). Makes it easier to see the intentional differences, and also in case anyone later wants to change the template part. (This is definitely optional, just an idea to consider.)
___
*[src/BloomTests/Book/HtmlDomTests.cs, line 1813 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5091#-N-OxvrOEqCVdtQu9cM6:-N-OxvrOEqCVdtQu9cM7:b-iuw1ut) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/799037200532b1dd037b592171ae850c55c1db26/src/BloomTests/Book/HtmlDomTests.cs#L1813)):*
> ```Smalltalk
> 			// Does doesn't seem to have string equality in its repertoire
> 			Assert.That(output, Does.StartWith(desiredResult));
> 			Assert.That(output, Does.EndWith(desiredResult));
> ```

Try output, Is.EqualTo(desiredResult) :-)
Maybe they think that is more idiomatic than Does.Equal.

JohnThomson (approved):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5091)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @StephenMcConnel)

PR #5095, opened 2022-04-08
Fix bug where NotifyUserOfProblem's buttons don't do anything

sujeffreyl:
The issue happens if NotifyUserOfProblem is called from a thread other than the main thread.

---
This change is [Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5095)

### DIFF
--- a/src/BloomExe/ErrorReporter/HtmlErrorReporter.cs
+++ b/src/BloomExe/ErrorReporter/HtmlErrorReporter.cs
@@ -399,7 +399,17 @@ namespace Bloom.ErrorReporter
 			// ENHANCE: Allow the caller to pass in the control, which would be at the front of this.
 			//System.Windows.Forms.Control control = Form.ActiveForm ?? FatalExceptionHandler.ControlOnUIThread;
 			var control = GetControlToUse();
-			SafeInvoke.InvokeIfPossible("Show Error Reporter", control, false, () =>
+
+			//   For a long time, we used false for this parameter (this causes Control.BeginInvoke to be used instead of Control.Invoke).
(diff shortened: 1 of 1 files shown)


sujeffreyl (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5095)* status: 0 of 1 files reviewed, 1 unresolved discussion

___
*[src/BloomExe/ErrorReporter/HtmlErrorReporter.cs, line 406 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5095#-N-96FSv3Zys6g6K_zCr:-N-96FSv3Zys6g6K_zCs:bkpzrm6) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/d914f9b27fb47a33de4aaa882d0291d411a4e7e1/src/BloomExe/ErrorReporter/HtmlErrorReporter.cs#L406)):*
> ```Smalltalk
> 			// However, that's not right. The return value of this function does depend upon the delegate's completion;
> 			// therefore it should be run synchronously to receive the correct return value.
> 			//   The downside though is that we open ourselves up to a chance of deadlock, if A) the delegate or B) any work ahead of the delegate on the main thread
> ```

Further reading: https://docs.google.com/document/d/1d8G8l_Zsc-MGnhTWQn2isb3YPBCjiQELjJs20dYXr9Q/edit#
This is a tough question; requires a lot of thought.

sujeffreyl:
Decided to go with a different approach for tackling this problem. (Trying not to have the calling thread block by default)

PR #5105, opened 2022-04-12
Make sure socket is available before sending on it (BL-11124)

andrew-polk:
This change is [Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5105)

### DIFF
--- a/src/BloomExe/web/BloomWebSocketServer.cs
+++ b/src/BloomExe/web/BloomWebSocketServer.cs
@@ -145,7 +145,8 @@ namespace Bloom.Api
 						// I don't know if Sending on a closed socket would throw, so we'll catch it in any case
 						try
 						{
-							socket?.Send(eventObject.ToString());
+							if (socket != null && socket.IsAvailable)
+								socket.Send(eventObject.ToString());
 						}
 						catch (Exception error)
 						{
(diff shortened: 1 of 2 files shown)


andrew-polk (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5105)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @hatton)

___
*[src/BloomExe/web/BloomWebSocketServer.cs, line 148 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5105#-N-V7yAjBtKGBliw21fP:-N-ZqjG2DiACBJjc5rFE:b9cer6l) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/8c285fe9643aee4d5180f592374446a2fb9d105c/src/BloomExe/web/BloomWebSocketServer.cs#L148)):*

Also, might an alternative be to swallow this at the Sentry level instead? Just some thoughts, no strong feelings here.
</blockquote></details>

We should probably discuss. I'm not seeing a way to improve things.

Definitely, there is no reason to send on a socket which is not available. So I think this is an improvement. But I would like to know how we could get into this state.

I played around with Bloom for quite a while trying to reproduce the scenario where we can get into this bad state but never did.

We were previously attempting to suppress such an error (the try/catch), but it actually comes out to Sentry as a `TaskScheduler.UnobservedTaskException` event. All such events can be suppressed by Sentry with `DisableTaskUnobservedTaskExceptionCapture`, but I don't think that's what we want to do.

andrew-polk:
Closing in favor of #5109. For now that will just go to master, but then we can decide how to proceed further.

PR #5110, opened 2022-04-14
MP3 option only enabled if book has narration (BL-10991)

sujeffreyl:
This change is [Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5110)

### DIFF
--- a/src/BloomBrowserUI/publish/video/AudioVideoOptionsGroup.tsx
+++ b/src/BloomBrowserUI/publish/video/AudioVideoOptionsGroup.tsx
@@ -46,11 +46,18 @@ export const AudioVideoOptionsGroup: React.FunctionComponent<{
     pageTurnDelay: number;
     onSetPageTurnDelay: (arg: number) => void;
     format: string;
-    setFormat: (f: string) => void;
+    onFormatChanged: (f: string) => void;
 }> = props => {
-    const format = props.format;
-    const setFormat = props.setFormat;
+    // When using props.format directly, it was possible to make VideoFormat render in an inconsistent state.
+    // (It rendered Icon as final value, but Label as intermediate value -> inconsistent rendering!)
+    // Making our own version of format using useState, rather than using props.format directly, makes this go away somehow.
+    // (Maybe because we're making it more explicit to React that the state is being updated and it needs to re-render stuff,
+    // but I don't fully understand why it wasn't working before)
+    const [format, setFormat] = useState(props.format);
+
+    // Stores which formats should be non-selectable.
     const [disabledFormats, setDisabledFormats] = useState<string[]>([]);
+
     const [tooBigMsg, setTooBigMsg] = useState("");
     // Manages visibility of the details popup for the main Format label (that shows in the
     // control when the dropdown is closed).
@@ -62,26 +69,18 @@ export const AudioVideoOptionsGroup: React.FunctionComponent<{
     // we want independent control over the dropdown, but so we can hide the popup details
     // thing when we show the dropdown.
     const [formatDropdownIsOpen, setFormatDropdownIsOpen] = useState(false);
+
+    // When component is mounted, find out which formats should be disabled, then update the state.
     useEffect(() => {
+        // Currently, only mp3 format can be disabled. Everything else is always enabled right now.
         BloomApi.get("publish/av/isMP3FormatSupported", c => {
             const isSupported = c.data as boolean;
-
-            if (isSupported && disabledFormats.includes("mp3")) {
-                // Previously disabled, but now enabled
-                const newDisabledFormats = disabledFormats.filter(
-                    x => x !== "mp3"
-                );
-                setDisabledFormats(newDisabledFormats);
-            } else if (!isSupported && !disabledFormats.includes("mp3")) {
-                // Previously enabled, but now disabled
-                const newDisabledFormats = disabledFormats.slice(0);
-                newDisabledFormats.push("mp3");
-                setDisabledFormats(newDisabledFormats);
-            } else {
-                // disabledFormats is unchanged; don't need to set anything again.
+            if (!isSupported) {
+                setDisabledFormats(prevValue => prevValue.concat(["mp3"]));
             }
         });
     }, []);
+
     useEffect(() => {
         BloomApi.get("publish/av/tooBigForScreenMsg", c => {
             setTooBigMsg(c.data);
@@ -131,7 +130,7 @@ export const AudioVideoOptionsGroup: React.FunctionComponent<{
     ];
 
     useEffect(() => {
-        // Handle any disabled formats
+        // Ensure selection is not disabled
         if (disabledFormats.includes(format)) {
             const firstNonDisabledFormat = formatItems
                 .map(x => x.format)
@@ -142,11 +141,8 @@ export const AudioVideoOptionsGroup: React.FunctionComponent<{
                 return;
             }
 
-            console.log(
-                `Because ${format} is disabled, changing format to ${firstNonDisabledFormat}`
-            );
-
             setFormat(firstNonDisabledFormat);
+            props.onFormatChanged(firstNonDisabledFormat);
         }
     }, [format, disabledFormats]);
 
@@ -205,6 +201,7 @@ export const AudioVideoOptionsGroup: React.FunctionComponent<{
                                         const newFormat = e.target
                                             .value as string;
                                         setFormat(newFormat);
+                                        props.onFormatChanged(newFormat);
                                     }}
                                     style={{ width: 160 }}
                                     renderValue={f => {
--- a/src/BloomBrowserUI/publish/video/AudioVideoOptionsGroup.tsx
+++ b/src/BloomBrowserUI/publish/video/AudioVideoOptionsGroup.tsx
@@ -50,6 +50,7 @@ export const AudioVideoOptionsGroup: React.FunctionComponent<{
 }> = props => {
     const format = props.format;
     const setFormat = props.setFormat;
+    const [disabledFormats, setDisabledFormats] = useState<string[]>([]);
     const [tooBigMsg, setTooBigMsg] = useState("");
     // Manages visibility of the details popup for the main Format label (that shows in the
     // control when the dropdown is closed).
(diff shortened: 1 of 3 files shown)


sujeffreyl (commented):
I suggest reviewing commit-by-commit

> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5110#-:-N-cvBDX4z86L9SwCeaT:b-di24mz)* status: 0 of 3 files reviewed, 1 unresolved discussion

___
*[src/BloomBrowserUI/publish/video/AudioVideoOptionsGroup.tsx, line 55 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5110#-N-cuyfNDHDCJXCGHkzu:-N-cuyfNDHDCJXCGHkzv:b-ixj0qz) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/8b180d2ea05f062c3a4f03b1bcc69260bde7afec/src/BloomBrowserUI/publish/video/AudioVideoOptionsGroup.tsx#L55)):*
> ```XML
>     // Making our own version of format using useState, rather than using props.format directly, makes this go away somehow.
>     // (Maybe because we're making it more explicit to React that the state is being updated and it needs to re-render stuff,
>     // but I don't follow understand why it wasn't working before)
> ```

I don't 100% understand the problem, so I wasn't able to write a great explanation here. If anyone can figure out the problem better and therefore write 1) a better explanation, and/or 2) a better workaround, I would love that.

sujeffreyl (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5110)* status: 0 of 3 files reviewed, 1 unresolved discussion

___
*[src/BloomBrowserUI/publish/video/AudioVideoOptionsGroup.tsx, line 55 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5110#-N-cuyfNDHDCJXCGHkzu:-N-cw7AW5PdJdyX4KCdp:b-xecxdc) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/8b180d2ea05f062c3a4f03b1bcc69260bde7afec/src/BloomBrowserUI/publish/video/AudioVideoOptionsGroup.tsx#L55)):*

I don't 100% understand the problem, so I wasn't able to write a great explanation here. If anyone can figure out the problem better and therefore write 1) a better explanation, and/or 2) a better workaround, I would love that.
</blockquote></details>

I initially thought the problem was due to a race condition, but the order the API calls returned in didn't seem to make any difference.

I thought the problem might've been because props.format is a field of an object, not a useState variable directly... but 1) it looked to me like the update of the avSettings object was happening appropriately, 2) when I tried to change PublishAudioVideo to use function syntaax for setState (real groaner BTW because useApiState did not handle that syntax yet), it still repro'd, and 3) when I changed PublishAudioVideo so that format was just a simple useState<string>, it still repro'd too.

Another weird thing is that I could repro it consistently in Bloom, but not really in my Firefox browser (v99)

sujeffreyl (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5110)* status: 0 of 3 files reviewed, 1 unresolved discussion

___
*[src/BloomBrowserUI/publish/video/AudioVideoOptionsGroup.tsx, line 55 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5110#-N-cuyfNDHDCJXCGHkzu:-N-d3QpEDkO73qGpTiQP:b-7unziw) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/8b180d2ea05f062c3a4f03b1bcc69260bde7afec/src/BloomBrowserUI/publish/video/AudioVideoOptionsGroup.tsx#L55)):*

I initially thought the problem was due to a race condition, but the order the API calls returned in didn't seem to make any difference.

I thought the problem might've been because props.format is a field of an object, not a useState variable directly... but 1) it looked to me like the update of the avSettings object was happening appropriately, 2) when I tried to change PublishAudioVideo to use function syntaax for setState (real groaner BTW because useApiState did not handle that syntax yet), it still repro'd, and 3) when I changed PublishAudioVideo so that format was just a simple useState<string>, it still repro'd too.

Another weird thing is that I could repro it consistently in Bloom, but not really in my Firefox browser (v99)
</blockquote></details>

Another thought that's too complicated for me to trace: props.icon is of type React.Node, whereas props.label is of type string. === would have different implications for these two types. Does that play into it?

andrew-polk (changes requested):
I added some comments (which I've left in place).
But then I experimented myself and pushed a branch which simplifies things and seems to work.
The downside is it doesn't handle future disabling of other formats. But I think we can treat that as yagni. If you think it is worth implementing, I think we can also make that work.
Branch name is AlternateApproachForBL10991.

> Reviewed 3 of 3 files at r1, 2 of 2 files at r2, 2 of 2 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5110#-:-N-dSCNNAuj4Tkgzbju5:b3ld6bk)* status: all files reviewed, 3 unresolved discussions (waiting on @sujeffreyl)

___
*[src/BloomBrowserUI/publish/video/AudioVideoOptionsGroup.tsx, line 55 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5110#-N-cuyfNDHDCJXCGHkzu:-N-dJKxkAVWhqeSAB6fr:bbo2n4q) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/8b180d2ea05f062c3a4f03b1bcc69260bde7afec/src/BloomBrowserUI/publish/video/AudioVideoOptionsGroup.tsx#L55)):*

Another thought that's too complicated for me to trace: props.icon is of type React.Node, whereas props.label is of type string. === would have different implications for these two types. Does that play into it?
</blockquote></details>

I agree it seems like it should work without the state variable.
You tried with using `props.format` rather than the reassigned/cloned `format`?

I would change the name to simply `format`. Or `selectedFormat`.
___
*[src/BloomBrowserUI/publish/video/AudioVideoOptionsGroup.tsx, line 83 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5110#-N-dJ-uh8cgraETrYBrY:-N-dJ-uh8cgraETrYBrZ:b-ufq32i) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/8b180d2ea05f062c3a4f03b1bcc69260bde7afec/src/BloomBrowserUI/publish/video/AudioVideoOptionsGroup.tsx#L83)):*
> ```XML
>             }
>         });
>     }, []);
> ```

I'm missing something.
This only runs once. And disabledFormats is initialized to an empty array.
So I don't understand all this logic of was this, now this.
___
*[src/BloomBrowserUI/publish/video/AudioVideoOptionsGroup.tsx, line 133 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5110#-N-dJrRY8iFBzMWopdxr:-N-dJrRY8iFBzMWopdxs:bg6qc5t) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/8b180d2ea05f062c3a4f03b1bcc69260bde7afec/src/BloomBrowserUI/publish/video/AudioVideoOptionsGroup.tsx#L133)):*
> ```XML
> 
>     useEffect(() => {
>         // Handle any disabled formats
> ```

I don't think this comment is accurate.
I would remove it or say something like `Ensure we don't select a disabled format`



_Code quote:_
```XML
// Handle any disabled formats
```

sujeffreyl (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5110)* status: all files reviewed, 3 unresolved discussions (waiting on @andrew-polk and @sujeffreyl)

___
*[src/BloomBrowserUI/publish/video/AudioVideoOptionsGroup.tsx, line 83 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5110#-N-dJ-uh8cgraETrYBrY:-N-xShiT0fl0n6nzK6Zl:b-bo9ayw) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/8b180d2ea05f062c3a4f03b1bcc69260bde7afec/src/BloomBrowserUI/publish/video/AudioVideoOptionsGroup.tsx#L83)):*

I'm missing something.
This only runs once. And disabledFormats is initialized to an empty array.
So I don't understand all this logic of was this, now this.

</blockquote></details>

It's just more generalized, able to work properly even if the 2nd param of  the useEffect changed in the future

andrew-polk (changes requested):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5110)* status: all files reviewed, 3 unresolved discussions (waiting on @sujeffreyl)

___
*[src/BloomBrowserUI/publish/video/AudioVideoOptionsGroup.tsx, line 83 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5110#-N-dJ-uh8cgraETrYBrY:-N-xW8Yn7g_0cMf92qHj:bxp5ucn) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/8b180d2ea05f062c3a4f03b1bcc69260bde7afec/src/BloomBrowserUI/publish/video/AudioVideoOptionsGroup.tsx#L83)):*

It's just more generalized, able to work properly even if the 2nd param of  the useEffect changed in the future
</blockquote></details>

Worth a comment, then, methinks

sujeffreyl (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5110)* status: 2 of 3 files reviewed, 2 unresolved discussions (waiting on @andrew-polk)

___
*[src/BloomBrowserUI/publish/video/AudioVideoOptionsGroup.tsx, line 55 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5110#-N-cuyfNDHDCJXCGHkzu:-N01FHldCsjjhukRcCTK:b-65inq0) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/8b180d2ea05f062c3a4f03b1bcc69260bde7afec/src/BloomBrowserUI/publish/video/AudioVideoOptionsGroup.tsx#L55)):*

I agree it seems like it should work without the state variable.
You tried with using `props.format` rather than the reassigned/cloned `format`?

I would change the name to simply `format`. Or `selectedFormat`.
</blockquote></details>

OK, changed the name back to format
___
*[src/BloomBrowserUI/publish/video/AudioVideoOptionsGroup.tsx, line 133 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5110#-N-dJrRY8iFBzMWopdxr:-N01EWLw7lOnG_9MAvH4:bb6un1j) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/8b180d2ea05f062c3a4f03b1bcc69260bde7afec/src/BloomBrowserUI/publish/video/AudioVideoOptionsGroup.tsx#L133)):*

I don't think this comment is accurate.
I would remove it or say something like `Ensure we don't select a disabled format`
</blockquote></details>

Done

andrew-polk (approved):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5110)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @sujeffreyl)

PR #5112, opened 2022-04-14
Remove extraneous BOMs introduced by hand-editing (BL-11127)

StephenMcConnel:
This change is [Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5112)

### DIFF
--- a/src/BloomExe/XmlHtmlConverter.cs
+++ b/src/BloomExe/XmlHtmlConverter.cs
@@ -38,7 +38,7 @@ namespace Bloom
 			var dom = new XmlDocument();
 			content = AddFillerToKeepTidyFromRemovingEmptyElements(content);
 
-			//in BL-2250, we found that in previous versions, this method would turn, for example, "<u> </u>" REMOVEWHITESPACE.
+			//in BL-2250, we found that in previous versions, this method would return, for example, "<u> </u>" REMOVEWHITESPACE.
 			//That is fixed now, but this is needed to give to clean up existing books.
 			content = content.Replace(@"REMOVEWHITESPACE", "");
 
@@ -66,6 +66,13 @@ namespace Bloom
 			var removedSvgs = new List<string>();
 			content = RemoveSvgs(content, removedSvgs);
 
+			// Hand editing HTML files can sometimes, with some editors, introduce a BOM after the first character in the HTML file.
+			// These can cause problems in creating PDFs (see https://issues.bloomlibrary.org/youtrack/issue/BL-11127).
+			// Either having a BOM or not having a BOM as the first character in the file doesn't really matter.  (File storage tends
+			// to use UTF-8, which doesn't need BOMs.)  And the tidy processing below strips an initial BOM anyway, so it doesn't
+			// hurt to do it here if it exists.  (Starting the search after the first character is perhaps a trivial optimization.)
+			if (content.IndexOf('\uFEFF', 1) > 0)
+				content = content.Replace("\uFEFF", "");
 
 			//using (var temp = new TempFile())
 			var temp = new TempFile();
--- a/src/BloomTests/XmlHtmlConverterTests.cs
+++ b/src/BloomTests/XmlHtmlConverterTests.cs
@@ -345,5 +345,20 @@ namespace BloomTests
 			// Trim because it may mess with leading or trailing white space in ways we don't care about.
 			Assert.That(xml.Trim(), Is.EqualTo(styleContent.Replace(">", "&gt;").Trim()));
 		}
+
+		[Test]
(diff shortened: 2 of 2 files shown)


hatton (changes requested):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5112)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @hatton and @StephenMcConnel)

___
*[src/BloomExe/XmlHtmlConverter.cs, line 71 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5112#-N-delYZCw82kHnqObC1:-N-delYZCw82kHnqObC2:biucd0t) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/247d2e61384bfcbbfc58df063e23123bd494942d/src/BloomExe/XmlHtmlConverter.cs#L71)):*
> ```Smalltalk
> 			// Hand editing HTML files can sometimes, with some editors, introduce a BOM after the first character in the HTML file.
> 			// These can cause problems in creating PDFs (see https://issues.bloomlibrary.org/youtrack/issue/BL-11127).
> 			if (content.IndexOf('\uFEFF', 1) > 0)
> ```

Probably I'm just misreading this, but the 1st line seems to be trying to overlook a BOM in the 1st character (fine), but then removes it from *everywhere* in the second.

StephenMcConnel (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5112)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @hatton)

___
*[src/BloomExe/XmlHtmlConverter.cs, line 71 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5112#-N-delYZCw82kHnqObC1:-N-dpl_y1Zy10pqNvgra:b-sdh8be) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/247d2e61384bfcbbfc58df063e23123bd494942d/src/BloomExe/XmlHtmlConverter.cs#L71)):*

Probably I'm just misreading this, but the 1st line seems to be trying to overlook a BOM in the 1st character (fine), but then removes it from *everywhere* in the second.
</blockquote></details>

You're reading it correctly.  We really don't care about a BOM as the first character in the file.  If it's there, no harm done, but if it's not there, also no harm done.  All of our file storage uses UTF-8 instead of UTF-16, and BOMs really don't make any sense in the context of UTF-8.  I could make the code more complicated to avoid remove a possible BOM in the the first position if you think it's worthwhile.  something like "content = content[0] + content.Substring(1).Replace("\uFEFF","");".

hatton (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5112)* status: 0 of 1 files reviewed, all discussions resolved (waiting on @hatton)

___
*[src/BloomExe/XmlHtmlConverter.cs, line 71 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5112#-N-delYZCw82kHnqObC1:-N-dwCOW9SCSe2Tv3j3S:b-l86mh2) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/247d2e61384bfcbbfc58df063e23123bd494942d/src/BloomExe/XmlHtmlConverter.cs#L71)):*

You're reading it correctly.  We really don't care about a BOM as the first character in the file.  If it's there, no harm done, but if it's not there, also no harm done.  All of our file storage uses UTF-8 instead of UTF-16, and BOMs really don't make any sense in the context of UTF-8.  I could make the code more complicated to avoid remove a possible BOM in the the first position if you think it's worthwhile.  something like "content = content[0] + content.Substring(1).Replace("\uFEFF","");".
</blockquote></details>

I trust your judgment either way, but lets at least add a comment since it does look like a mistake to step gingerly over the 1st one but then blow it away :-)

StephenMcConnel (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5112)* status: 0 of 1 files reviewed, all discussions resolved (waiting on @hatton)

___
*[src/BloomExe/XmlHtmlConverter.cs, line 71 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5112#-N-delYZCw82kHnqObC1:-N-eFWzuCNtNfYho0DZM:b-xqcf6g) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/247d2e61384bfcbbfc58df063e23123bd494942d/src/BloomExe/XmlHtmlConverter.cs#L71)):*

I trust your judgment either way, but lets at least add a comment since it does look like a mistake to step gingerly over the 1st one but then blow it away :-)
</blockquote></details>

I added some commentary and also a unit test.

PR #5116, opened 2022-04-18
Fix URL encoded filename stored for cover image (BL-11145)

StephenMcConnel:
I thought we killed this bug years ago, but apparently not quite.

---
This change is [Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5116)

### DIFF
--- a/src/BloomExe/Book/Book.cs
+++ b/src/BloomExe/Book/Book.cs
@@ -152,6 +152,7 @@ namespace Bloom.Book
 				InitCoverColor(); // should use the same color as what they saw in the preview of the template/shell
 			}
 			FixBookIdAndLineageIfNeeded();
+			FixUrlEncodedCoverImageIfNeeded();
 			Storage.Dom.RemoveExtraBookTitles();
 			Storage.Dom.RemoveExtraContentTypesMetas();
 			Guard.Against(OurHtmlDom.RawDom.InnerXml=="","Bloom could not parse the xhtml of this document");
@@ -1132,11 +1133,13 @@ namespace Bloom.Book
 			var filename = Path.GetFileName(coverImagePath);
 			if (filename != coverImageFileName)
 			{
+				// Note that setting InnerText or calling SetAttribute automatically XML-encodes any necessary
+				// characters.  Getting InnerText or calling GetAttribute reverses the changes, if any.
 				coverImgElt.InnerText = filename;
-				var urlPathString = UrlPathString.CreateFromUnencodedString(filename);
-				var encoded = urlPathString.UrlEncoded;
-				coverImgElt.SetAttribute("src", encoded);
-				coverImgElt.SetAttribute("alt", $"This picture, {encoded}, is missing or was loading too slowly.");
+				coverImgElt.SetAttribute("src", filename);
+				var localizedFormatString = LocalizationManager.GetString("EditTab.Image.AltMsg", "This picture, {0}, is missing or was loading too slowly.");
+				var altValue = String.Format(localizedFormatString, filename);
+				coverImgElt.SetAttribute("alt", altValue);
 			}
 		}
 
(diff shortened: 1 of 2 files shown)


JohnThomson (changes requested):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5116)* status: 0 of 1 files reviewed, 3 unresolved discussions (waiting on @JohnThomson and @StephenMcConnel)

___
*[`src/BloomExe/Book/Book.cs` line 1895 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5116#-N1jQDOF6QncRtdCgIof:-N1jQDOF6QncRtdCgIog:btl9uv2) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/3685e6003ba655e12144cd62352563839027cf8a/src/BloomExe/Book/Book.cs#L1895)):*
> ```Smalltalk
> 			{
> 				var filepath = UrlPathString.GetFullyDecodedPath(FolderPath, ref text);
> 				if (!Regex.IsMatch(filepath, kUrlEncodedRegex))
> ```

I don't understand the reason for this further test. GetFullyDecodedPath is smart enough to modify text only if that results in finding a file where the unmodified name did not. So at this point text is either unchanged, or improved in the way we want: it now locates a file that is probably correct. So why not just set the inner text? Maybe the file name really does have a % escape in it.
(See also suggestions below. What if we optimized GetFullyDecodedPath to return the path without even doing an existence check if the filename has no percent characters? Then call it unconditionally, but pass it a copy of text, and just abandon all other changes if text is unchanged?)
___
*[`src/BloomExe/Book/Book.cs` line 1900 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5116#-N1jO3pp-GJkJvWS8Nw3:-N1jO3pp-GJkJvWS8Nw4:b-fgkj92) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/3685e6003ba655e12144cd62352563839027cf8a/src/BloomExe/Book/Book.cs#L1900)):*
> ```Smalltalk
> 				if (Regex.IsMatch(src, "%25[0-9a-fA-F][0-9a-fA-F]"))
> 				{
> 					// This looks like it was doubly encoded: remove one layer of encoding.
> ```

This implies that GetAttribute removes one layer, and SetAttribute puts one back; but I can't find any documentation indicating that such is the case. Why will it only remove a layer if it is doubly encoded?
___
*[`src/BloomExe/Book/Book.cs` line 1902 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5116#-N1jPKfM8Yr3EPK8X-bC:-N1jPKfM8Yr3EPK8X-bD:b-uum8f4) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/3685e6003ba655e12144cd62352563839027cf8a/src/BloomExe/Book/Book.cs#L1902)):*
> ```Smalltalk
> 					// This looks like it was doubly encoded: remove one layer of encoding.
> 					src = HttpUtility.UrlDecode(src);
> 					node.SetAttribute("src", src);
> ```

Is it a concern that decoding might produce a character like &<>" that would be illegal in an attribute value?
Should we again check that making the change allows us to find a file where otherwise we would not?
Or possibly, should we only try to fix src and alt if changing the text of the element allowed us to match an existing file (thus pretty strongly diagnosing a book that has this problem)? I hate to see us add a patch just to fix a past problem that will permanently cause problems in the (admittedly unlikely) case that we actually want an escape sequence in one of these attributes. Of course, you almost cover this case by returning if the expected file exists; but it would be stronger still to do it only if urldecoding locates an otherwise missing file.

StephenMcConnel (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5116)* status: 0 of 2 files reviewed, 3 unresolved discussions (waiting on @JohnThomson)

___
*[`src/BloomExe/Book/Book.cs` line 1895 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5116#-N1jQDOF6QncRtdCgIof:-N2Hv_xu9HoXJS1TGHE3:bcbu3ct) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/3685e6003ba655e12144cd62352563839027cf8a/src/BloomExe/Book/Book.cs#L1895)):*

I don't understand the reason for this further test. GetFullyDecodedPath is smart enough to modify text only if that results in finding a file where the unmodified name did not. So at this point text is either unchanged, or improved in the way we want: it now locates a file that is probably correct. So why not just set the inner text? Maybe the file name really does have a % escape in it.
(See also suggestions below. What if we optimized GetFullyDecodedPath to return the path without even doing an existence check if the filename has no percent characters? Then call it unconditionally, but pass it a copy of text, and just abandon all other changes if text is unchanged?)
</blockquote></details>

GetFullyDecodedPath works only for one level of encoding.  The problem book has two levels of encoding.  Perhaps the fix should focus in GetFullyDecodedPath, which is what I've done.
___
*[`src/BloomExe/Book/Book.cs` line 1900 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5116#-N1jO3pp-GJkJvWS8Nw3:-N2HuE942_7i9YLCKgmT:byfows0) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/3685e6003ba655e12144cd62352563839027cf8a/src/BloomExe/Book/Book.cs#L1900)):*

This implies that GetAttribute removes one layer, and SetAttribute puts one back; but I can't find any documentation indicating that such is the case. Why will it only remove a layer if it is doubly encoded?
</blockquote></details>

I don't think that's the case.  I think the comment is misleading you.  But I've rewritten things enough that this comment no longer exists.  :-)
___
*[`src/BloomExe/Book/Book.cs` line 1902 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5116#-N1jPKfM8Yr3EPK8X-bC:-N2I26aV1IZgYYZEbqAD:buxeufq) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/3685e6003ba655e12144cd62352563839027cf8a/src/BloomExe/Book/Book.cs#L1902)):*

Is it a concern that decoding might produce a character like &<>" that would be illegal in an attribute value?
Should we again check that making the change allows us to find a file where otherwise we would not?
Or possibly, should we only try to fix src and alt if changing the text of the element allowed us to match an existing file (thus pretty strongly diagnosing a book that has this problem)? I hate to see us add a patch just to fix a past problem that will permanently cause problems in the (admittedly unlikely) case that we actually want an escape sequence in one of these attributes. Of course, you almost cover this case by returning if the expected file exists; but it would be stronger still to do it only if urldecoding locates an otherwise missing file.
</blockquote></details>

Setting InnerText automatically XML-encodes (or HTML-encodes if you prefer to think that way) the necessary characters.  GetAttribute and SetAttribute do the same thing according to a simple test program I wrote (and ran on both Windows and Linux).

JohnThomson (approved):
Nice! I like this much better.

> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5116#-:-N2O-6_nCtYInSMidLkz:b1ne2yk)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @StephenMcConnel)

PR #5121, opened 2022-04-20
Make sure socket is available before sending on it (BL-11124)

andrew-polk:
This change is [Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5121)


hatton:
What are your thoughts on pushing an invisible-to-users fix to a release, particularly before lots of real-world beta testing?

hatton:
In BL-11124, it seems you were wanting to land this on master.

andrew-polk:
I did push the change to master which added info to the sentry errors.
This one just prevents them (unless running in development).
It seemed safe enough to me to push directly to 5.1 and stop the sentry errors. But I'm definitely open to beta only or beta first or master first.

PR #5124, opened 2022-04-22
Add page range control (BL-11146)

JohnThomson:
Draft because four unit tests fail, and I don't have time to fix them. However, it's working in case anyone wants to play with it.
(Also, want to try using MaxInt for the default number of pages, since the current code will not work well if the BloomPub has more pages than the original book.)

---
This change is [Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5124)


andrew-polk:
Note that I've already reviewed this interactively

andrew-polk:
I ran the tests myself and didn't get any failures.
I'll put a note in the card about using MaxInt.

PR #5133, opened 2022-04-27
BL-11059 Detect Disconnected state in Status panel

gmartin7:
* the specific case is checking out a book just after the
   connection goes bad

---
This change is [Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5133)

### DIFF
--- a/src/BloomBrowserUI/teamCollection/TeamCollectionBookStatusPanel.tsx
+++ b/src/BloomBrowserUI/teamCollection/TeamCollectionBookStatusPanel.tsx
@@ -389,7 +389,7 @@ export const TeamCollectionBookStatusPanel: React.FunctionComponent<IBookTeamCol
                         "teamCollection/attemptLockOfCurrentBook",
                         response => {
                             // Not much to do. Change of state is handled by websocket notifications.
-                            // We want to keep it that way, so we don't have to worry about here about
+                            // We want to keep it that way, so we don't have to worry here about
                             // whether the checkout attempt succeeded or not.
                             setBusy(false);
                         },
--- a/src/BloomExe/CollectionTab/CollectionTabView.cs
+++ b/src/BloomExe/CollectionTab/CollectionTabView.cs
@@ -276,10 +276,16 @@ namespace Bloom.CollectionTab
 
 		/// <summary>
 		/// Set a new TC status image. Called at Idle time or startup, on the UI thread.
+		/// N.B.: It also gets called if the user tries to do something and the TeamCollection suddenly
+		/// recognizes it is in a disconnected state.
 		/// </summary>
 		public void SetTeamCollectionStatus(TeamCollectionManager tcManager)
 		{
 			_tcStatusButton.Update(tcManager.CollectionStatus);
+			// This will cause the CollectionsTabBookPane to reload the status of the book
+			// (and the collection itself), which will trickle down to the status panel.
+			if (tcManager.CollectionStatus == TeamCollectionStatus.Disconnected)
+				_webSocketServer.SendEvent("bookStatus", "reload");
 		}
 
 		private void _tcStatusButton_Click(object sender, EventArgs e)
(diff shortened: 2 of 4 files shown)


andrew-polk (changes requested):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5133)* status: all files reviewed, 2 unresolved discussions (waiting on @gmartin7)

___
*[`src/BloomExe/TeamCollection/TeamCollectionApi.cs` line 357 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5133#-N0glfzdBcsb4nPN1kpo:-N0glfzdBcsb4nPN1kpp:b-93s09u) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/58ea0118d80f53a9fb11db56b4b0a67610c29563/src/BloomExe/TeamCollection/TeamCollectionApi.cs#L357)):*
> ```Smalltalk
> 					// situations apart is that in (1) book.IsSaveable is true and in (2) it is not.
> 					whoHasBookLocked = book.IsSaveable ? CurrentUser : null;
> 				}
> ```

Do we still need to set newLocalBook appropriately?
I think yes. 
But if no, we can get rid of the variable.
___
*[`src/BloomExe/TeamCollection/TeamCollectionManager.cs` line 335 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5133#-N0glxq86bgMjiQdzY4o:-N0glxq86bgMjiQdzY4p:bqidrsd) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/58ea0118d80f53a9fb11db56b4b0a67610c29563/src/BloomExe/TeamCollection/TeamCollectionManager.cs#L335)):*
> ```Smalltalk
> 			CurrentCollectionEvenIfDisconnected.SocketServer = SocketServer;
> 			CurrentCollectionEvenIfDisconnected.TCManager = this;
> 			// Every call to MessageLog.WriteMessage() also raises the TeamCollectionStatusChanged event.
> ```

Doesn't seem like you changed this, but...
Should it? Seems... wrong. Or at least less than ideal that writing a message always raises a status change.

gmartin7 (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5133)* status: 3 of 4 files reviewed, 2 unresolved discussions (waiting on @andrew-polk)

___
*[`src/BloomExe/TeamCollection/TeamCollectionApi.cs` line 357 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5133#-N0glfzdBcsb4nPN1kpo:-N0goJdgBBtbRD3t3kdF:bpnqutt) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/58ea0118d80f53a9fb11db56b4b0a67610c29563/src/BloomExe/TeamCollection/TeamCollectionApi.cs#L357)):*

Do we still need to set newLocalBook appropriately?
I think yes. 
But if no, we can get rid of the variable.
</blockquote></details>

Ah! I thought I had just forgotten to remove the declaration, but then I see it IS used down below. Oops. Fixed.
___
*[`src/BloomExe/TeamCollection/TeamCollectionManager.cs` line 335 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5133#-N0glxq86bgMjiQdzY4o:-N0greFj-ck3M3Klrghx:b-ixlrmg) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/58ea0118d80f53a9fb11db56b4b0a67610c29563/src/BloomExe/TeamCollection/TeamCollectionManager.cs#L335)):*

Doesn't seem like you changed this, but...
Should it? Seems... wrong. Or at least less than ideal that writing a message always raises a status change.
</blockquote></details>

No, I didn't. Unfortunately, WriteMessage or its derivatives are called in about 21 places in the code! Not sure I want to chase down whether a status changed event is necessary or not for each of those!

andrew-polk (approved):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5133)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @gmartin7)

PR #5134, opened 2022-04-27
Show progress dialog while audio/video is processing (BL-11136)

andrew-polk:
This change is [Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5134)


hatton:
Does everything make sense when doing mp3 output?

andrew-polk:
> Does everything make sense when doing mp3 output?

yes, then you just get "Processing audio" and "Finalizing audio"

PR #5139, opened 2022-05-09
BL-9961 Enable multi-paragraph paste from Notepad

gmartin7:
This change is [Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5139)

### DIFF
--- a/src/BloomBrowserUI/bookEdit/bloomField/BloomField.ts
+++ b/src/BloomBrowserUI/bookEdit/bloomField/BloomField.ts
@@ -116,6 +116,9 @@ export default class BloomField {
         //     alert(event.data.dataValue);
         // });
         ckeditor.on("paste", event => {
+            event.data.dataValue = this.reconstituteParagraphsOnPlainTextPaste(
+                event.data
+            );
             event.data.dataValue = this.convertStandardFormatVerseMarkersToSuperscript(
                 event.data.dataValue
             );
@@ -150,18 +153,12 @@ export default class BloomField {
                 );
             }
 
-            const paras = event.data.dataValue.match(/<p>/g);
-            if (!paras) {
-                // Enhance: should we remove the probable <span> and just leave the text?
-                // But it might be carrying some important style info.
-                return;
-            }
             // OK, we are going to unwrap the first <p> and just leave its content.
             // This prevents a typically unwanted extra line break being inserted before the
             // start of the material copied. Without the <p> wrapper, that material just
             // gets inserted into the current paragraph.
             let endMkr = "";
-            if (paras.length === 1) {
+            if (event.data.dataValue.match(/<p>/g)?.length === 1) {
                 // Unwrapping the one and only <p> will leave no break between what used to be that
                 // paragraph and the following text, which should now start a new paragraph.
                 // We insert an empty paragraph to force a break, but that will leave an unwanted
@@ -175,6 +172,7 @@ export default class BloomField {
                 .replace("<p>", "")
                 .replace("</p>", endMkr);
         });
+
         ckeditor.on("afterPaste", event => {
             // clean up possible unwanted paragraph inserted by paste event.
             $(".removeMe").remove();
@@ -208,6 +206,40 @@ export default class BloomField {
         // This makes it easy to find the right editor instance. There may be some ckeditor built-in way, but
         // I wasn't able to find one.
         (<any>bloomEditableDiv).bloomCkEditor = ckeditor;
+    }
+
+    // If the original clipboard had no paragraph markup, but only (plain text) newlines (\n)
+    // (e.g. pasting from Notepad), this method will remove the newlines and substitue HTML paragraph
+    // markup.
+    // The problem is that when pasting from Notepad, 'dataValue' has removed the newlines and squished
+    // all the lines together into one line. Pasting the same thing from MSWord, 'dataValue' contains
+    // HTML paragraph markup. So if there are no paragraph marks and there were (in the original
+    // clipboard data) newlines, we reconstitute the needed paragraph marks from the original newlines
+    // that we have to go fishing inside of the dataTransfer object to find. (BL-9961)
+    static reconstituteParagraphsOnPlainTextPaste(eventData: any): any {
+        if (!eventData.dataValue.match(/<p>/g)) {
+            // If we're inserting plain text from Notepad, we will arrive here because 'dataValue'
+            // doesn't have paragraphs, but we actually do want paragraph markup, if there
+            // are multiple lines in the original clipboard data.
+            // Finding where the original string with its newlines was passed inside the
+            // dataTransfer object was a bit tricky.
+            const textWithReturns: string = eventData.dataTransfer._.data.Text;
+            if (!textWithReturns.includes("\n")) {
+                // Enhance: should we remove the probable <span> and just leave the text?
+                // But it might be carrying some important style info.
+                return eventData.dataValue; // no change
+            }
+            // Split the text on carriage returns and put it back together with each bit in a paragraph.
+            const reconstitutedTextWithParas = textWithReturns
+                .split("\n")
+                .reduce(
+                    (resultSoFar, part) => resultSoFar + "<p>" + part + "</p>",
+                    ""
+                );
+            // Reset dataValue
+            return reconstitutedTextWithParas;
+        }
+        return eventData.dataValue; // no change
     }
 
     // Not private so we can unit test it. It is too difficult to get the actual paste
--- a/src/BloomBrowserUI/bookEdit/js/bloomEditing.ts
+++ b/src/BloomBrowserUI/bookEdit/js/bloomEditing.ts
@@ -307,15 +307,6 @@ function SetBookCopyrightAndLicenseButtonVisibility(container) {
         .css("display", shouldShowButton ? "inline" : "none");
 }
 
-function DecodeHtml(encodedString) {
-    return encodedString
-        .replace(/&amp;/g, "&")
-        .replace(/&lt;/g, "<")
-        .replace(/&gt;/g, ">")
-        .replace(/&#39;/g, "'")
-        .replace(/&#169;/g, "©");
-}
-
 function GetEditor() {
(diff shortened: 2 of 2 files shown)


andrew-polk (changes requested):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5139)* status: all files reviewed, 2 unresolved discussions (waiting on @gmartin7)

___
*[`src/BloomBrowserUI/bookEdit/bloomField/BloomField.ts` line 153 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5139#-N1eyX-r7l16q7PV5xB_:-N1eyX-r7l16q7PV5xBa:b-j4xpsq) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/0bba18c4d6dc5cb36dc8ced0f501c1050dde4a7e/src/BloomBrowserUI/bookEdit/bloomField/BloomField.ts#L153)):*
> ```XML
>             }
> 
>             let paras = event.data.dataValue.match(/<p>/g);
> ```

Probably due to some intermediate state you had while working on this, but you've changed const to let unnecessarily.



_Code quote:_
```XML
let
```
___
*[`src/BloomBrowserUI/bookEdit/bloomField/BloomField.ts` line 175 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5139#-N1eyBgU76_QyqNDUyCW:-N1eyBgV02X9SqXfEHKY:b-h5xj8y) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/0bba18c4d6dc5cb36dc8ced0f501c1050dde4a7e/src/BloomBrowserUI/bookEdit/bloomField/BloomField.ts#L175)):*
> ```XML
>                     );
>                 // Reset dataValue
>                 event.data.dataValue = reconstitutedTextWithParas;
> ```

If we do this, won't we lose any changes made to `event.data.dataValue` above (such as `convertStandardFormatVerseMarkersToSuperscript`, etc.)?

Can/should you do the replacement before the rest of this handler runs?
___
*[`src/BloomBrowserUI/bookEdit/js/bloomEditing.ts` line 317 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5139#-N1eylsx3AoXBdu5YacI:-N1eylsx3AoXBdu5YacJ:b-71kmno) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/96b0d409abe8699847bb057bb7f8e676e9dd1990/src/BloomBrowserUI/bookEdit/js/bloomEditing.ts#L317)):*
> ```XML
>         .replace(/&#39;/g, "'")
>         .replace(/&#169;/g, "©");
> }
> ```

(for next time)
It is generally helpful to add a comment in reviewable when making a change for which the reason is not obvious like this.
(I verified this is not used myself by looking at code.)

andrew-polk (changes requested):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5139)* status: all files reviewed, 4 unresolved discussions (waiting on @gmartin7)

___
*[`src/BloomBrowserUI/bookEdit/bloomField/BloomField.ts` line 211 at r2](https://reviewable.io/reviews/bloombooks/bloomdesktop/5139#-N1j-jrk9V4vTHGJawYK:-N1j-jrk9V4vTHGJawYL:bb4y6pg) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/78a140fecc41fe472f256a31de42b5347f8da017/src/BloomBrowserUI/bookEdit/bloomField/BloomField.ts#L211)):*
> ```XML
>     }
> 
>     // Return value will reset 'event.data.dataValue'
> ```

That is only true for the one place this is currently called, but there could be other callers later.

I think it would be better to write a comment explaining what this method does and returns.
i.e. if no paras but does have line breaks, returns blah blah...



_Code quote:_
```XML
// Return value will reset 'event.data.dataValue'
```
___
*[`src/BloomBrowserUI/bookEdit/bloomField/BloomField.ts` line 212 at r2](https://reviewable.io/reviews/bloombooks/bloomdesktop/5139#-N1j-a_dCoZq4uiL4CMH:-N1j-a_dCoZq4uiL4CMI:bb7phvt) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/78a140fecc41fe472f256a31de42b5347f8da017/src/BloomBrowserUI/bookEdit/bloomField/BloomField.ts#L212)):*
> ```XML
> 
>     // Return value will reset 'event.data.dataValue'
>     static reconstitueParagraphsOnPlainTextPaste(eventData: any): any {
> ```

typo



_Code quote:_
```XML
reconstitue
```
___
*[`src/BloomBrowserUI/bookEdit/bloomField/BloomField.ts` line 215 at r2](https://reviewable.io/reviews/bloombooks/bloomdesktop/5139#-N1j-9rSBEnhxFNC5zfF:-N1j-9rSBEnhxFNC5zfG:b4gjnaz) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/78a140fecc41fe472f256a31de42b5347f8da017/src/BloomBrowserUI/bookEdit/bloomField/BloomField.ts#L215)):*
> ```XML
>         if (!eventData.dataValue.match(/<p>/g)) {
>             // BL-9961 If we're inserting plain text from Notepad, we will arrive here because
>             // 'dataValue' doesn't have paragraphs, but we actually want them.
> ```

I think this comment is now incorrect (or at least misleading?).
We only want paragraphs if there are line breaks, right?



_Code quote:_
```XML
            // BL-9961 If we're inserting plain text from Notepad, we will arrive here because
            // 'dataValue' doesn't have paragraphs, but we actually want them.
```
___
*[`src/BloomBrowserUI/bookEdit/bloomField/BloomField.ts` line 216 at r2](https://reviewable.io/reviews/bloombooks/bloomdesktop/5139#-N1j-JM4Bfg26IZcKNKc:-N1j-JM4Bfg26IZcKNKd:b32838a) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/78a140fecc41fe472f256a31de42b5347f8da017/src/BloomBrowserUI/bookEdit/bloomField/BloomField.ts#L216)):*
> ```XML
>             // BL-9961 If we're inserting plain text from Notepad, we will arrive here because
>             // 'dataValue' doesn't have paragraphs, but we actually want them.
>             // Finding where the original string with its paragraphs was passed inside the
> ```

Can't you just pass in event.data.dataValue to this method now and use that instead?



_Code quote:_
```XML
            // Finding where the original string with its paragraphs was passed inside the
            // dataTransfer object was a bit tricky.
            const textWithReturns: string = eventData.dataTransfer._.data.Text;
```

gmartin7 (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5139)* status: 1 of 2 files reviewed, 3 unresolved discussions (waiting on @andrew-polk)

___
*[`src/BloomBrowserUI/bookEdit/bloomField/BloomField.ts` line 211 at r2](https://reviewable.io/reviews/bloombooks/bloomdesktop/5139#-N1j-jrk9V4vTHGJawYK:-N1jd7HPCgzmhQpSaRLn:b-fsijxj) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/78a140fecc41fe472f256a31de42b5347f8da017/src/BloomBrowserUI/bookEdit/bloomField/BloomField.ts#L211)):*

That is only true for the one place this is currently called, but there could be other callers later.

I think it would be better to write a comment explaining what this method does and returns.
i.e. if no paras but does have line breaks, returns blah blah...
</blockquote></details>

Comment updated.
___
*[`src/BloomBrowserUI/bookEdit/bloomField/BloomField.ts` line 215 at r2](https://reviewable.io/reviews/bloombooks/bloomdesktop/5139#-N1j-9rSBEnhxFNC5zfF:-N1jdkep45IcUXS409IL:baw1xrg) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/78a140fecc41fe472f256a31de42b5347f8da017/src/BloomBrowserUI/bookEdit/bloomField/BloomField.ts#L215)):*

I think this comment is now incorrect (or at least misleading?).
We only want paragraphs if there are line breaks, right?
</blockquote></details>

Updated all the comments. Should be clearer now.
___
*[`src/BloomBrowserUI/bookEdit/bloomField/BloomField.ts` line 216 at r2](https://reviewable.io/reviews/bloombooks/bloomdesktop/5139#-N1j-JM4Bfg26IZcKNKc:-N1j6LlAATtgHpowL7JL:b-txy56c) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/78a140fecc41fe472f256a31de42b5347f8da017/src/BloomBrowserUI/bookEdit/bloomField/BloomField.ts#L216)):*

Can't you just pass in event.data.dataValue to this method now and use that instead?
</blockquote></details>

That's kinda the whole point! The `dataValue` at this point has NO line breaks. Just all the lines squashed together in one line! e.g.: `One.Two.Three`, where the `dataTransfer._.data.Text` has the original `One.\nTwo.\nThree.\n`.

andrew-polk (approved):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5139)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @gmartin7)

___
*[`src/BloomBrowserUI/bookEdit/bloomField/BloomField.ts` line 216 at r2](https://reviewable.io/reviews/bloombooks/bloomdesktop/5139#-N1j-JM4Bfg26IZcKNKc:-N1jhJnQ9O7STB0pRr2Q:b-nctsom) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/78a140fecc41fe472f256a31de42b5347f8da017/src/BloomBrowserUI/bookEdit/bloomField/BloomField.ts#L216)):*

That's kinda the whole point! The `dataValue` at this point has NO line breaks. Just all the lines squashed together in one line! e.g.: `One.Two.Three`, where the `dataTransfer._.data.Text` has the original `One.\nTwo.\nThree.\n`.
</blockquote></details>

Ah, I see. I made the mistake of assuming that it was there in `event.data.dataValue` but had been altered by the time we got to this point in the code originally.

PR #5145, opened 2022-05-11
Implement new IErrorReporter interface from libpalaso 9.0.0-beta0167

sujeffreyl:
This change is [Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5145)

### DIFF
--- a/src/BloomExe/BloomExe.csproj
+++ b/src/BloomExe/BloomExe.csproj
@@ -1586,6 +1586,10 @@ cp -p "$(SolutionDir)lib/dotnet/GeckofxHtmlToPdf.exe.config" "$(OutDir)BloomPdfM
   <Import Project="..\..\packages\LargeAddressAware.1.0.5\build\LargeAddressAware.targets" Condition="'$(OS)'=='Windows_NT' AND Exists('..\..\packages\LargeAddressAware.1.0.5\build\LargeAddressAware.targets')" />
   <Import Project="..\..\packages\NDesk.DBus.0.15.0\build\NDesk.DBus.targets" Condition="Exists('..\..\packages\NDesk.DBus.0.15.0\build\NDesk.DBus.targets')" />
   <Import Project="..\..\packages\Mono.Posix.5.4.0.201\build\net45\Mono.Posix.targets" Condition="'$(OS)'=='Windows_NT' AND Exists('..\..\packages\Mono.Posix.5.4.0.201\build\net45\Mono.Posix.targets')" />
+  <Import Project="..\..\packages\SQLitePCLRaw.lib.e_sqlite3.2.0.3\build\net461\SQLitePCLRaw.lib.e_sqlite3.targets" Condition="Exists('..\..\packages\SQLitePCLRaw.lib.e_sqlite3.2.0.3\build\net461\SQLitePCLRaw.lib.e_sqlite3.targets')" />  
+  <Import Project="..\..\packages\SIL.Media.9.0.0-beta0176\build\SIL.Media.targets" Condition="Exists('..\..\packages\SIL.Media.9.0.0-beta0176\build\SIL.Media.targets')" />
+  <Import Project="..\..\packages\SIL.Windows.Forms.9.0.0-beta0176\build\SIL.Windows.Forms.targets" Condition="Exists('..\..\packages\SIL.Windows.Forms.9.0.0-beta0176\build\SIL.Windows.Forms.targets')" />
+  <Import Project="..\..\packages\SIL.Windows.Forms.Keyboarding.9.0.0-beta0176\build\SIL.Windows.Forms.Keyboarding.targets" Condition="Exists('..\..\packages\SIL.Windows.Forms.Keyboarding.9.0.0-beta0176\build\SIL.Windows.Forms.Keyboarding.targets')" />
   <Target Name="AfterBuild">
     <Copy SourceFiles="@(SourceFiles)" DestinationFolder="$(OutDir)" />
     <Copy SourceFiles="@(LameFiles)" DestinationFolder="../../DistFiles" Condition="'$(OS)'=='Windows_NT'" />
@@ -1651,10 +1655,6 @@ cp -p "$(SolutionDir)lib/dotnet/GeckofxHtmlToPdf.exe.config" "$(OutDir)BloomPdfM
     <Delete Files="@(DeleteForLinux)" TreatErrorsAsWarnings="true" />
     <RemoveDir Directories="@(RemoveForLinux)" />
   </Target>
-  <Import Project="..\..\packages\SQLitePCLRaw.lib.e_sqlite3.2.0.3\build\net461\SQLitePCLRaw.lib.e_sqlite3.targets" Condition="Exists('..\..\packages\SQLitePCLRaw.lib.e_sqlite3.2.0.3\build\net461\SQLitePCLRaw.lib.e_sqlite3.targets')" />
-  <Import Project="..\..\packages\SIL.Media.9.0.0-beta0176\build\SIL.Media.targets" Condition="Exists('..\..\packages\SIL.Media.9.0.0-beta0176\build\SIL.Media.targets')" />
-  <Import Project="..\..\packages\SIL.Windows.Forms.9.0.0-beta0176\build\SIL.Windows.Forms.targets" Condition="Exists('..\..\packages\SIL.Windows.Forms.9.0.0-beta0176\build\SIL.Windows.Forms.targets')" />
-  <Import Project="..\..\packages\SIL.Windows.Forms.Keyboarding.9.0.0-beta0176\build\SIL.Windows.Forms.Keyboarding.targets" Condition="Exists('..\..\packages\SIL.Windows.Forms.Keyboarding.9.0.0-beta0176\build\SIL.Windows.Forms.Keyboarding.targets')" />
   <!-- Note: Visual Studio automatically inserts any imports from installing packages into the end here,
   but the AfterBuild target may not run if .targets files are imported after it.
   I recommend you move the Import statements before the "AfterBuild" target,
--- a/src/BloomExe/BloomExe.csproj
+++ b/src/BloomExe/BloomExe.csproj
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Project ToolsVersion="12.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
-  <Import Project="..\..\packages\SIL.Windows.Forms.Keyboarding.9.0.0-beta0165\build\SIL.Windows.Forms.Keyboarding.props" Condition="Exists('..\..\packages\SIL.Windows.Forms.Keyboarding.9.0.0-beta0165\build\SIL.Windows.Forms.Keyboarding.props')" />
+  <Import Project="..\..\packages\SIL.Windows.Forms.Keyboarding.9.0.0-beta0176\build\SIL.Windows.Forms.Keyboarding.props" Condition="Exists('..\..\packages\SIL.Windows.Forms.Keyboarding.9.0.0-beta0176\build\SIL.Windows.Forms.Keyboarding.props')" />
   <PropertyGroup>
     <ApplicationManifest>app.manifest</ApplicationManifest>
     <ProjectGuid>{304D5612-167C-4725-AF27-B9F2BB788B57}</ProjectGuid>
@@ -234,28 +234,28 @@
       <HintPath>../../packages/SharpFont.3.1.0/lib/net20/SharpFont.dll</HintPath>
     </Reference>
     <Reference Include="SIL.Core, Version=9.0.0.0, Culture=neutral, PublicKeyToken=cab3c8c5232dfcf2, processorArchitecture=MSIL">
-      <HintPath>..\..\packages\SIL.Core.9.0.0-beta0165\lib\net461\SIL.Core.dll</HintPath>
+      <HintPath>..\..\packages\SIL.Core.9.0.0-beta0176\lib\net461\SIL.Core.dll</HintPath>
     </Reference>
     <Reference Include="SIL.Core.Desktop, Version=9.0.0.0, Culture=neutral, PublicKeyToken=cab3c8c5232dfcf2, processorArchitecture=MSIL">
-      <HintPath>..\..\packages\SIL.Core.Desktop.9.0.0-beta0165\lib\net461\SIL.Core.Desktop.dll</HintPath>
+      <HintPath>..\..\packages\SIL.Core.Desktop.9.0.0-beta0176\lib\net461\SIL.Core.Desktop.dll</HintPath>
     </Reference>
     <Reference Include="SIL.Media, Version=9.0.0.0, Culture=neutral, processorArchitecture=MSIL">
-      <HintPath>..\..\packages\SIL.Media.9.0.0-beta0165\lib\net461\SIL.Media.dll</HintPath>
+      <HintPath>..\..\packages\SIL.Media.9.0.0-beta0176\lib\net461\SIL.Media.dll</HintPath>
     </Reference>
     <Reference Include="SIL.Windows.Forms, Version=9.0.0.0, Culture=neutral, PublicKeyToken=cab3c8c5232dfcf2, processorArchitecture=MSIL">
-      <HintPath>..\..\packages\SIL.Windows.Forms.9.0.0-beta0165\lib\net461\SIL.Windows.Forms.dll</HintPath>
+      <HintPath>..\..\packages\SIL.Windows.Forms.9.0.0-beta0176\lib\net461\SIL.Windows.Forms.dll</HintPath>
     </Reference>
     <Reference Include="SIL.Windows.Forms.GeckoBrowserAdapter, Version=9.0.0.0, Culture=neutral, PublicKeyToken=cab3c8c5232dfcf2, processorArchitecture=MSIL">
-      <HintPath>..\..\packages\SIL.Windows.Forms.GeckoBrowserAdapter.9.0.0-beta0165\lib\net461\SIL.Windows.Forms.GeckoBrowserAdapter.dll</HintPath>
+      <HintPath>..\..\packages\SIL.Windows.Forms.GeckoBrowserAdapter.9.0.0-beta0176\lib\net461\SIL.Windows.Forms.GeckoBrowserAdapter.dll</HintPath>
     </Reference>
     <Reference Include="SIL.Windows.Forms.Keyboarding, Version=9.0.0.0, Culture=neutral, PublicKeyToken=cab3c8c5232dfcf2, processorArchitecture=MSIL">
-      <HintPath>..\..\packages\SIL.Windows.Forms.Keyboarding.9.0.0-beta0165\lib\net461\SIL.Windows.Forms.Keyboarding.dll</HintPath>
+      <HintPath>..\..\packages\SIL.Windows.Forms.Keyboarding.9.0.0-beta0176\lib\net461\SIL.Windows.Forms.Keyboarding.dll</HintPath>
     </Reference>
     <Reference Include="SIL.Windows.Forms.WritingSystems, Version=9.0.0.0, Culture=neutral, PublicKeyToken=cab3c8c5232dfcf2, processorArchitecture=MSIL">
-      <HintPath>..\..\packages\SIL.Windows.Forms.WritingSystems.9.0.0-beta0165\lib\net461\SIL.Windows.Forms.WritingSystems.dll</HintPath>
+      <HintPath>..\..\packages\SIL.Windows.Forms.WritingSystems.9.0.0-beta0176\lib\net461\SIL.Windows.Forms.WritingSystems.dll</HintPath>
     </Reference>
     <Reference Include="SIL.WritingSystems, Version=9.0.0.0, Culture=neutral, PublicKeyToken=cab3c8c5232dfcf2, processorArchitecture=MSIL">
-      <HintPath>..\..\packages\SIL.WritingSystems.9.0.0-beta0165\lib\net461\SIL.WritingSystems.dll</HintPath>
+      <HintPath>..\..\packages\SIL.WritingSystems.9.0.0-beta0176\lib\net461\SIL.WritingSystems.dll</HintPath>
     </Reference>
     <Reference Include="SourceMapDotNet, Version=1.0.5478.26629, Culture=neutral, PublicKeyToken=d357635542166cea, processorArchitecture=MSIL">
       <HintPath>..\..\packages\SourceMapDotNet.1.0.5478.26629\lib\SourceMapDotNet.dll</HintPath>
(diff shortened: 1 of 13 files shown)


JohnThomson (changes requested):
I'm disappointed. There may be good reasons, but where I was hoping the changes to libpalaso would make things simpler on the Bloom side, they seem to have gotten more complicated, and quite hard to understand. Perhaps it would be less so seeing the whole code rather than just a set of changes. But it makes me wonder, could we do something just a bit different that would move things towards simplicity rather than complexity? Sorry to be so vaguely negative, but I think it's worth at least some thought to try for more improvement. Possibly some of my more detailed comments will suggest a way forward.

> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5145#-:-N1oER4iBqKRgA2hePqZ:bogiui0)* status: all files reviewed, 6 unresolved discussions (waiting on @sujeffreyl)

___
*[`src/BloomExe/Program.cs` line 100 at r3](https://reviewable.io/reviews/bloombooks/bloomdesktop/5145#-N1o2MP7-6wTT5KQkAZX:-N1o2MP8DU1jGNdICOFj:brxx1li) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/143373ac2fd8aa365b371968f33b82a03088040e/src/BloomExe/Program.cs#L100)):*
> ```Smalltalk
> 		// Note: This mechanism is NOT compatible with .NET Core.
> 		// As of May 2022, this mechanism is not used for any super-essential logic. If we need to move to .NET core,
> 		// you could consider just deleting this variable entirely.
> ```

This puzzled me until I carefully read the doc on ThreadStatic. Maybe comment something like "Only the one thread that executes the class constructor runs the initializer for ThreadStatic variables; thus, on all other threads, it will have its default false value."
https://stackoverflow.com/questions/2374451/how-to-tell-if-a-thread-is-the-main-thread-in-c-sharp suggest several other ways to do this...maybe one of them would work in other .NETs as well?
___
*[`src/BloomExe/ErrorReporter/HtmlErrorReporter.cs` line 260 at r3](https://reviewable.io/reviews/bloombooks/bloomdesktop/5145#-N1o4S-j1Dvx7NwD2cEG:-N1o4S-j1Dvx7NwD2cEH:bmbo4f5) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/143373ac2fd8aa365b371968f33b82a03088040e/src/BloomExe/ErrorReporter/HtmlErrorReporter.cs#L260)):*
> ```Smalltalk
> 		public ErrorResult NotifyUserOfProblem(IRepeatNoticePolicy policy, string alternateButton1Label, ErrorResult resultIfAlternateButtonPressed, string message)
> 		{
> 			Debug.Assert(Program.IsMainThread || IsWaitAllowed, "NotifyUserOfProblem (Advanced) running off the main thread. This thread will wait while the UI pops up. This has not been thoroughly tested to prevent deadlock. If the caller is sure this is safe, the caller should set IsWaitAllowed to true before invoking this method");
> ```

I think IsWaitAllowed should be an optional argument. The Assert should only pass if this particular error invocation is known by the caller to be safe, not just because (possibly) some other error happening at about the same time is safe, or even because some code forgot to turn it off. (But maybe that would require yet another libpalaso interface change?? If so, I think we can live with it.)
Are you sure Debug.Assert() is sufficient? Lots of errors only happen in unexpected cases, so code that makes this mistake might never run in a debug build. Of course, even if it IS run on some other thread, it may work fine, at least most of the time, so maybe we don't want to bother the user. But perhaps we should at least log the problem?

Why is it only this overload that asserts if not called on the main thread?
___
*[`src/BloomExe/ErrorReporter/HtmlErrorReporter.cs` line 478 at r3](https://reviewable.io/reviews/bloombooks/bloomdesktop/5145#-N1o9mhr4i9b-PAPpfaL:-N1o9mhr4i9b-PAPpfaM:b-dw65tx) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/143373ac2fd8aa365b371968f33b82a03088040e/src/BloomExe/ErrorReporter/HtmlErrorReporter.cs#L478)):*
> ```Smalltalk
> 								else
> 								{
> 									// Advanced case: Libpalaso wants to deal with the call result itself. [not recommended]
> ```

This would be for things like "try again" to delete a file, right? As long as this is done on the main thread and the caller knows there is going to be a wait for user response, I don't think we need to recommend against it...what we were worried about was threads expecting to complete quickly, and possibly holding locks, waiting a long time when some error gets reported. But haven't we reached a point where reporting of unexpected exceptions never uses letCallerHandleClicks=true? Because I thought that was a main objective of all this.
___
*[`src/BloomExe/ErrorReporter/HtmlErrorReporter.cs` line 481 at r3](https://reviewable.io/reviews/bloombooks/bloomdesktop/5145#-N1oCgQW82Vtk6gT_PW8:-N1oCgQW82Vtk6gT_PW9:bbmkjh0) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/143373ac2fd8aa365b371968f33b82a03088040e/src/BloomExe/ErrorReporter/HtmlErrorReporter.cs#L481)):*
> ```Smalltalk
> 									// OnShowDetails will be invoked if this method returns {resultIfAlternateButtonPressed}
> 									// FYI, setting to null is OK. It should cause ErrorReport to reset to default handler.
> 									ErrorReport.OnShowDetails = OnSecondaryActionPressed;
> ```

This seems to be what we were doing before, but it's mind-blowing that we get our OnSecondaryActionPressed method called by changing a global static in a library that is meant for configuring how exception details are reported. What makes sure that this gets set back to the normal thing?? Do we use this path, or could we remove it, and do something like throw if a secondary action is configured when letCallerHandleClicks is true? After all, the fact that we want our own thing done when the button is clicked really means we do NOT want the caller to handle the clicks...so why configure things to tell this method that the caller SHOULD handle them, and then turn around and trick the caller into letting us handle them?
Of course there may be a good reason, perhaps arising out of legacy stuff in libpalaso that we haven't found a good way to improve yet...but if so I think we need comments addressing the points I'm raising here.
___
*[`src/BloomExe/ErrorReporter/HtmlErrorReporter.cs` line 494 at r3](https://reviewable.io/reviews/bloombooks/bloomdesktop/5145#-N1o9Ze6-7fFKs_H3mY6:-N1o9Ze7BU0bSx-T-x02:b-7ug3we) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/143373ac2fd8aa365b371968f33b82a03088040e/src/BloomExe/ErrorReporter/HtmlErrorReporter.cs#L494)):*
> ```Smalltalk
> 								}
> 								else
> 								{
> ```

I thought we had so improved libpalaso that it would never need to handle the report button itself?
___
*[`src/BloomExe/ErrorReporter/HtmlErrorReporter.cs` line 532 at r3](https://reviewable.io/reviews/bloombooks/bloomdesktop/5145#-N1o7zahA7LL6se7P3y1:-N1o7zahA7LL6se7P3y2:bu6mwfy) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/143373ac2fd8aa365b371968f33b82a03088040e/src/BloomExe/ErrorReporter/HtmlErrorReporter.cs#L532)):*
> ```Smalltalk
> 			});
> 
> 			// Reset IsWaitAllowed after the SafeInvoke call
> ```

Can we get this into a 'finally' so it can't possibly be skipped?

andrew-polk (approved):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5145)* status: all files reviewed, 6 unresolved discussions (waiting on @JohnThomson)

sujeffreyl (commented):
Ok, I converted the legacy method to call the new method.

This allowed me to delete a lot of supporting infrastructure in both HtmlErrorReporter and especially in the entry point, ErrorReportUtils, that was supporting customizing the legacy path.

I also made the ErrorReportUtils entry point expose a simpler NotifyUserOfProblem(message, [optional exception,] [more optional params]) method, which I believe is much simpler to grok.  IMO, message is the most important parameter, so now that is first. Everything else is optional, and now that can all be optional params. Previously, message was the most important parameter but it was 2nd from the end, so that it could support the params object[] args syntax, but that seemed of little utility to me because the only time you gain even a little is if your IRepeatNoticePolicy doesn't require you to show the message, then you won't have to String.Format(msgFormat, args), but we always use ShowAlwaysPolicy(), so there's no savings because we always have to format the string...  Moving away from the params object[] args syntax makes the method signature much more usable for practical use IMO.

> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5145#-:-N2Ce0hq3lcSyucl2SmR:b3e38oe)* status: 5 of 13 files reviewed, 6 unresolved discussions (waiting on @andrew-polk and @JohnThomson)

___
*[`src/BloomExe/ErrorReporter/HtmlErrorReporter.cs` line 260 at r3](https://reviewable.io/reviews/bloombooks/bloomdesktop/5145#-N1o4S-j1Dvx7NwD2cEG:-N2CfAeU36Ki2GyKWXYs:bnm7l8r) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/143373ac2fd8aa365b371968f33b82a03088040e/src/BloomExe/ErrorReporter/HtmlErrorReporter.cs#L260)):*

1) I didn't do it as an optional argument because it would need to get plumbed through libpalaso. That would require yet another libpalaso interface change, as you mentioned, which I didn't want to do.

2) Yeah, I was also worried about the caller forgetting to turn it off, so I made it so that IsWaitAllowed is automatically turned off by this class, so we don't need to worry about the caller forgetting.

3) I think it is sufficient. So far, I have seen many observations of it running just fine, but 0 concrete examples of deadlock, that's all just a hypothetical concern. I don't feel that's enough evidence to justify something like throwing an exception -> Problem Report Dialog. If it happens in the wild, I'd rather it just run and do its thing and hope for the best (which has good chances of succeeding).

4)  This overload asserts because it eventually sets forceSync to true (and has a theoretical possibility to deadlock), whereas the other overload does not, so the other overload doesn't need to worry about IsWaitAllowed.


</blockquote></details>

With my big simplification/refactoring, IsWaitAllowed is unnecessary now. Yay.
___
*[`src/BloomExe/ErrorReporter/HtmlErrorReporter.cs` line 478 at r3](https://reviewable.io/reviews/bloombooks/bloomdesktop/5145#-N1o9mhr4i9b-PAPpfaL:-N2CfFQX3QpxVNaxYfz5:b-vcwxk2) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/143373ac2fd8aa365b371968f33b82a03088040e/src/BloomExe/ErrorReporter/HtmlErrorReporter.cs#L478)):*

Actually, I have an idea I think I like better. I think I have 8 lines of code that will fully convert a call to the old API method to the new API method.  Then, I can simplify all the internals of HtmlErrorReporter to only deal with the one case (simpler case) and not have to worry about supporting the legacy path directly anymore.
I'm optimistic that that should clean that up a lot for us!
</blockquote></details>

Oops, it was longer than 8 lines, but I was able to delete a lot of old code because we don't need all the infrastructure added to support the legacy path anymore.
___
*[`src/BloomExe/ErrorReporter/HtmlErrorReporter.cs` line 481 at r3](https://reviewable.io/reviews/bloombooks/bloomdesktop/5145#-N1oCgQW82Vtk6gT_PW8:-N2CfUgXB4MKPz5iSh-u:b-sbyidn) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/143373ac2fd8aa365b371968f33b82a03088040e/src/BloomExe/ErrorReporter/HtmlErrorReporter.cs#L481)):*

Yeah, the libpalaso interface was definitely not very pleasant for me...

Q: What makes sure that this gets set back to the normal thing??
A: Line 290 of HtmlErrorReporter sets it to null before we can ShowNotifyDialog.    I think that should work as long as nobody resets the HtmlErrorReporter?
If you want, we could copy it to the finally clause of this branch as well.

Q: Do we use this path?
Yes, if libpalaso or a Bloom developer calls the function on line 258, which is required by the IErrorReporter interface.

Q: , or could we remove it
Not unless we jettison our full implementation of the IErrorReporter interface.

"After all, the fact that we want our own thing done when the button is clicked really means we do NOT want the caller to handle the clicks"
This is not actually a true statement. This is the legacy way of handling extra buttons. It had to go through libpalaso. And if you're coming through this path, you are going through libpalaso, and you want libpalaso to handle the clicks, so you do want (really need) the caller to handle the clicks. (I did say "[not recommended]").

[As mentioned in a previous comment, if we decide to jettison our full implementation of the IErrorReporter interface, then we can probably delete this whole part of the code]
</blockquote></details>

Now we don't need to deal with OnShowDetails anymore. Yay!
___
*[`src/BloomExe/ErrorReporter/HtmlErrorReporter.cs` line 494 at r3](https://reviewable.io/reviews/bloombooks/bloomdesktop/5145#-N1o9Ze6-7fFKs_H3mY6:-N2Cfy7377CRopY17ZlJ:b493ptv) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/143373ac2fd8aa365b371968f33b82a03088040e/src/BloomExe/ErrorReporter/HtmlErrorReporter.cs#L494)):*

No, we added a new interface method, but the original interface method is still there.
</blockquote></details>

Original interface method is now a 2nd-class citizen, allowing us to get rid of this letCallerHandleClicks stuff.

sujeffreyl (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5145)* status: 5 of 13 files reviewed, 6 unresolved discussions (waiting on @andrew-polk and @JohnThomson)

___
*[`src/BloomTests/ErrorReporter/HtmlErrorReporterTests.cs` line 222 at r5](https://reviewable.io/reviews/bloombooks/bloomdesktop/5145#-N2CgL3O7z5Z8T_jNy9p:-N2CgL3O7z5Z8T_jNy9q:bvopm4h) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/d8f4f5519d859babcb3a992cb9441f375f1c41a7/src/BloomTests/ErrorReporter/HtmlErrorReporterTests.cs#L222)):*
> ```Smalltalk
> 		{
> 			var mockFactory = GetDefaultMockReactDialogFactory();
> 			IBloomErrorReporter reporter = new HtmlErrorReporterBuilder()
> ```

FYI, the reason this type is given has to do with the confusing world of default parameters in interfaces/the classes that implement the interface.  (and that the two could potentially specify different defaults, and then you have the really confusing situation of which default wins...)

I decided to go with only giving the defaults on the interface, and not specifying default values on the implementing classes, a la https://michael-mckenna.com/optional-arguments-in-c-sharp-interfaces/
(This means that anything that used the interface can get the default values, yay, but it did require me to cast to the interface in several places in the unit tests)

I could alternatively be talked into going with only specifying it in HtmlErrorReporter's implementation. (This is the only place where I actively make use of the default parameters right now).

I'll admit I'm not 100% sure of the best way to deal with this can of worms.

Personally, I am a fan of default values and think (reportButtonLabel: "Details") is more readable than ("Details", null, null, null)



_Code quote:_
```Smalltalk
IBloomErrorReporter
```

JohnThomson (changes requested):
I like the way you're moving now. Just a couple more things to think about.

> Reviewed 3 of 5 files at r1, 2 of 2 files at r2, 8 of 8 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5145#-:-N2DYTJZDGxL3A-Qzxdg:bwkqtyh)* status: all files reviewed, 3 unresolved discussions (waiting on @sujeffreyl)

___
*[`src/BloomExe/ErrorReporter/ErrorReportUtils.cs` line 15 at r5](https://reviewable.io/reviews/bloombooks/bloomdesktop/5145#-N2DWqk539gBIkmDBWzD:-N2DWqk539gBIkmDBWzE:b4xvz9o) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/d8f4f5519d859babcb3a992cb9441f375f1c41a7/src/BloomExe/ErrorReporter/ErrorReportUtils.cs#L15)):*
> ```Smalltalk
> 		/// Sets all the extra parameters required to customize the buttons/extra buttons when calling NotifyUserOfProblem
> 		/// They can't be directly added to NotifyUserOfProblem because it needs to match the IErrorReporter interface.
> 		/// As a result, we work around that by having class instance variables that you set before invoking NotifyUserOfProblem.
> ```

I'm not following this. All our implementations of IErrorReporter are now implementing IBloomErrorReporter so as to implement this method. So why can't they instead implement an overload of NotifyUserOfProblem (or give it another name if you like) that takes these things as extra arguments? The nice, short version that implements the required method of IErrorReporter can call the new one with the appropriate defaults, and that implementation just uses the extra arguments the way it now uses the static variables. The advantage of doing this is that if we somehow get re-entrant calls to error handling their desired values for these things can't interfere with each other. There may be a disadvantage but I'm not seeing it.
Feel free to push back. I may be asking you to undo a major improvement. I like the simplification you've achieved, and I may well be missing something.
___
*[`src/BloomExe/ErrorReporter/ErrorReportUtils.cs` line 98 at r5](https://reviewable.io/reviews/bloombooks/bloomdesktop/5145#-N2DS1568dr6-fZQGbL4:-N2DS1568dr6-fZQGbL5:bo90r0f) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/d8f4f5519d859babcb3a992cb9441f375f1c41a7/src/BloomExe/ErrorReporter/ErrorReportUtils.cs#L98)):*
> ```Smalltalk
> 			{
> 				// Tests a path through libPalaso directly (goes thru overloads 1, 2, 4)
> 				ErrorReport.NotifyUserOfProblem(fakeProblemMessage);
> ```

Looks like an existing problem, but why is all this test code in a production class? Can't we make some method virtual and add the special test behavior in a test subclass, or something like that?
___
*[`src/BloomExe/ErrorReporter/HtmlErrorReporter.cs` line 97 at r5](https://reviewable.io/reviews/bloombooks/bloomdesktop/5145#-N2DT68y6EM6jkZCNYd5:-N2DT68y6EM6jkZCNYd6:b-9nwjho) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/d8f4f5519d859babcb3a992cb9441f375f1c41a7/src/BloomExe/ErrorReporter/HtmlErrorReporter.cs#L97)):*
> ```Smalltalk
> 		/// <param name="onReportButtonPressed">The action to execute after the report button is pressed. Pass null to use the default behavior.</param>
> 		/// <param name="extraButtonLabel">The localized text for the Secondary Action button.  For example, you might offer a "Retry" button.
> 		/// Pass null for default behavior (disabled). Pass "" to disable.</param>
> ```

So in fact, for this one, there's no difference between null and ""?

sujeffreyl (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5145)* status: all files reviewed, 3 unresolved discussions (waiting on @JohnThomson)

___
*[`src/BloomExe/ErrorReporter/ErrorReportUtils.cs` line 15 at r5](https://reviewable.io/reviews/bloombooks/bloomdesktop/5145#-N2DWqk539gBIkmDBWzD:-N2DfLm90h6MvyERB1KA:b-e7hi9l) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/d8f4f5519d859babcb3a992cb9441f375f1c41a7/src/BloomExe/ErrorReporter/ErrorReportUtils.cs#L15)):*

I'm not following this. All our implementations of IErrorReporter are now implementing IBloomErrorReporter so as to implement this method. So why can't they instead implement an overload of NotifyUserOfProblem (or give it another name if you like) that takes these things as extra arguments? The nice, short version that implements the required method of IErrorReporter can call the new one with the appropriate defaults, and that implementation just uses the extra arguments the way it now uses the static variables. The advantage of doing this is that if we somehow get re-entrant calls to error handling their desired values for these things can't interfere with each other. There may be a disadvantage but I'm not seeing it.
Feel free to push back. I may be asking you to undo a major improvement. I like the simplification you've achieved, and I may well be missing something.
</blockquote></details>

Right, so that's an interesting idea that I tried, but it's not without its pitfalls.

So to get rid of the static variables, we would need to avoid going through libpalaso land at all. That has some pros, but also introduces some headaches.
1) Some boilerplate code from libpalaso's ErrorReport.cs (the beginning and end of Overload #4 of NotifyUserOfProblem) should theoretically be copied over into our implementations. That means two very different places to keep up to date, if things were ever changed. So that makes me a little dissatisfied, but it's not that big of a deal.
2) The boilerplate code relies on some private variables in libpalaso's ErrorReport, so I can't do it out of the box. I'd have to go and muddle around more in libpalaso. But if I changed one variable from private to protected, I'd feel like I'd need to think about what exactly kind of things ErrorReport should expose at a public vs protected level vs not at all (since there are other private vars in that class that annoyed me. I don't think this question was thought about at the time it was written), and that felt like more work than I was willing to bite off. But you know, I could gather up the time and motivation to work on more libpalaso changes.
3) It was a bigger change to the code flow, so it was bigger/scarier, so I discarded my changes when it became clear that there were more obstacles than I originally envisioned.

However, after writing this out, I'm thinking that even though it's a groaner task for me, it's probably worth it to just bite the bullet now instead of setting up another new 2nd-choice paradigm. Thoughts?
___
*[`src/BloomExe/ErrorReporter/ErrorReportUtils.cs` line 98 at r5](https://reviewable.io/reviews/bloombooks/bloomdesktop/5145#-N2DS1568dr6-fZQGbL4:-N2DiSwWE3cNGu6stIhF:b-rrqo92) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/d8f4f5519d859babcb3a992cb9441f375f1c41a7/src/BloomExe/ErrorReporter/ErrorReportUtils.cs#L98)):*

Looks like an existing problem, but why is all this test code in a production class? Can't we make some method virtual and add the special test behavior in a test subclass, or something like that?
</blockquote></details>

This isn't unit test code, but this allows you to run Bloom and open specific books with specific titles to generate specific error reports. We have a .bloompack with them.
I'm thinking... do you think "Exercises a path" would be better than "Tests a path"
___
*[`src/BloomExe/ErrorReporter/HtmlErrorReporter.cs` line 97 at r5](https://reviewable.io/reviews/bloombooks/bloomdesktop/5145#-N2DT68y6EM6jkZCNYd5:-N2Dimh-2eoJAb4zqoeH:bfi26mb) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/d8f4f5519d859babcb3a992cb9441f375f1c41a7/src/BloomExe/ErrorReporter/HtmlErrorReporter.cs#L97)):*

So in fact, for this one, there's no difference between null and ""?
</blockquote></details>

Right, for extraButtonLabel, they're the same.

JohnThomson (approved):
I'm approving this now, but will wait to merge it in case you want to improve things further.

> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5145#-:-N2HFfWvE-SW-4PzroOz:bg4ajlm)* status: all files reviewed, 3 unresolved discussions (waiting on @sujeffreyl)

___
*[`src/BloomExe/ErrorReporter/ErrorReportUtils.cs` line 15 at r5](https://reviewable.io/reviews/bloombooks/bloomdesktop/5145#-N2DWqk539gBIkmDBWzD:-N2HD2bk0Xfd9hBnR5p0:b-8jg5iq) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/d8f4f5519d859babcb3a992cb9441f375f1c41a7/src/BloomExe/ErrorReporter/ErrorReportUtils.cs#L15)):*

Right, so that's an interesting idea that I tried, but it's not without its pitfalls.

So to get rid of the static variables, we would need to avoid going through libpalaso land at all. That has some pros, but also introduces some headaches.
1) Some boilerplate code from libpalaso's ErrorReport.cs (the beginning and end of Overload #4 of NotifyUserOfProblem) should theoretically be copied over into our implementations. That means two very different places to keep up to date, if things were ever changed. So that makes me a little dissatisfied, but it's not that big of a deal.
2) The boilerplate code relies on some private variables in libpalaso's ErrorReport, so I can't do it out of the box. I'd have to go and muddle around more in libpalaso. But if I changed one variable from private to protected, I'd feel like I'd need to think about what exactly kind of things ErrorReport should expose at a public vs protected level vs not at all (since there are other private vars in that class that annoyed me. I don't think this question was thought about at the time it was written), and that felt like more work than I was willing to bite off. But you know, I could gather up the time and motivation to work on more libpalaso changes.
3) It was a bigger change to the code flow, so it was bigger/scarier, so I discarded my changes when it became clear that there were more obstacles than I originally envisioned.

However, after writing this out, I'm thinking that even though it's a groaner task for me, it's probably worth it to just bite the bullet now instead of setting up another new 2nd-choice paradigm. Thoughts?


</blockquote></details>

I had not realized that the extra parameters would have to go through libpalaso. Maybe the comment could make that clearer. In view of that, I'm not opposed to your exploring alternatives more, but also wouldn't mind merging without that fix.
I presume it would be unexpected for the static variables to be not in their default state when SetNotifyUserOfProblemCustomParams is called? That would indicate a reentrant call that might be a problem? Can we take advantage of that to make it less of a problem at all? For example, maybe this would already indicate a failure of error handling, and we could do something that would force a switch to the fallback handler?
___
*[`src/BloomExe/ErrorReporter/ErrorReportUtils.cs` line 98 at r5](https://reviewable.io/reviews/bloombooks/bloomdesktop/5145#-N2DS1568dr6-fZQGbL4:-N2HEt0i7h_yLMqxZ9XH:bi73omk) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/d8f4f5519d859babcb3a992cb9441f375f1c41a7/src/BloomExe/ErrorReporter/ErrorReportUtils.cs#L98)):*

This isn't unit test code, but this allows you to run Bloom and open specific books with specific titles to generate specific error reports. We have a .bloompack with them.
I'm thinking... do you think "Exercises a path" would be better than "Tests a path"
</blockquote></details>

If it isn't already commented somewhere, it would definitely be worth explaining that this isn't for unit tests, but for manual testing to exercise errors by opening books with a specific title.
___
*[`src/BloomExe/ErrorReporter/HtmlErrorReporter.cs` line 97 at r5](https://reviewable.io/reviews/bloombooks/bloomdesktop/5145#-N2DT68y6EM6jkZCNYd5:-N2HFJci6IIq7cyi6fgo:b3ovwdj) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/d8f4f5519d859babcb3a992cb9441f375f1c41a7/src/BloomExe/ErrorReporter/HtmlErrorReporter.cs#L97)):*

Right, for extraButtonLabel, they're the same.
</blockquote></details>

Maybe, "Passing null (the default) or empty string result in no extra button"?

PR #5149, opened 2022-05-12
Improve page range control (BL-11146)

JohnThomson:
Tweaks things so a small page range can be shown in a long book.
Fixes problems with extra page labels being shown.
Makes things work right even if bloomPub has more pages than original.

---
This change is [Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5149)


JohnThomson:
Possibly we should wait for UI review before merging this?

JohnThomson:
Converted to draft as it is proving difficult to get the look and feel exactly right and is taking too much time.

andrew-polk:
Closing as we decided to live with what we have for now. (See https://issues.bloomlibrary.org/youtrack/issue/BL-11146)

PR #5150, opened 2022-05-13
Improvements to "change to a different layout" (BL-11147)

JohnThomson:
- Change all similar pages won't change pages where data would be lost unless this was expected from the different layouts of the current and selected page
- It also won't change pages whose block counts are different from the current page, even if data loss is approved
- HTML widgets are transferred and counted in deciding whether data will be lost

---
This change is [Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5150)

### DIFF
--- a/src/BloomBrowserUI/pageChooser/page-chooser.ts
+++ b/src/BloomBrowserUI/pageChooser/page-chooser.ts
@@ -207,6 +207,13 @@ export class PageChooser {
             ),
             10
         );
+        const selectedTemplateWidgetCount = parseInt(
+            this.getAttributeStringSafely(
+                this._selectedGridItem,
+                "data-widgetCount"
+            ),
+            10
+        );
 
         const page = <HTMLIFrameElement>(
             window.parent.document.getElementById("page")
@@ -232,12 +239,16 @@ export class PageChooser {
         const currentVideoCount = current.getElementsByClassName(
             "bloom-videoContainer"
         ).length;
+        const currentWidgetCount = current.getElementsByClassName(
+            "bloom-widgetContainer"
+        ).length;
 
         return (
             selectedTemplateTranslationGroupCount <
                 currentTranslationGroupCount ||
             selectedTemplatePictureCount < currentPictureCount ||
-            selectedTemplateVideoCount < currentVideoCount
+            selectedTemplateVideoCount < currentVideoCount ||
+            selectedTemplateWidgetCount < currentWidgetCount
         );
     }
 
@@ -295,7 +306,9 @@ export class PageChooser {
                 {
                     pageId: pageId,
                     templateBookPath: templateBookPath,
-                    convertWholeBook: convertWholeBookChecked
+                    convertWholeBook: convertWholeBookChecked,
+                    numberToAdd: 1, // meaningless here, but prevents throwing an exception in C#
+                    allowDataLoss: convertAnywayChecked
                 },
                 PageChooser.closeup
             );
@@ -305,8 +318,9 @@ export class PageChooser {
                 {
                     templateBookPath: templateBookPath,
                     pageId: pageId,
-                    convertWholeBook: false,
-                    numberToAdd: numberToAdd
+                    convertWholeBook: false, // meaningless here, but keeps C# happy
+                    numberToAdd: numberToAdd,
+                    allowDataLoss: convertAnywayChecked // meaningless here, but keeps C# happy
                 },
                 PageChooser.closeup
             );
@@ -566,6 +580,12 @@ export class PageChooser {
                 "data-videoCount",
                 currentPageDiv
                     .getElementsByClassName("bloom-videoContainer")
+                    .length.toString()
+            );
+            currentGridItemHtml.setAttribute(
+                "data-widgetCount",
+                currentPageDiv
+                    .getElementsByClassName("bloom-widgetContainer")
                     .length.toString()
             );
             const helpLink = currentPageDiv.getAttribute("help-link");
--- a/src/BloomExe/Book/HtmlDom.cs
+++ b/src/BloomExe/Book/HtmlDom.cs
@@ -738,6 +738,16 @@ namespace Bloom.Book
 			return false;
 		}
 
+		public static (int textCount, int imageCount, int videoCount, int widgetCount) GetEditableDataCounts(XmlElement page)
+		{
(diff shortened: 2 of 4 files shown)


andrew-polk (changes requested):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5150)* status: all files reviewed, 2 unresolved discussions (waiting on @JohnThomson)

___
*[`src/BloomExe/Book/HtmlDom.cs` line 1113 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5150#-N2CWnvx7nIfb91UljFY:-N2CWnvx7nIfb91UljFZ:bsu22tg) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/a265c84d0a7f72c3455805d58567f46fe41c7a1f/src/BloomExe/Book/HtmlDom.cs#L1113)):*
> ```Smalltalk
> 					newParent.SetAttribute("class",
> 						newParent.Attributes["class"].Value.Replace("bloom-noWidgetSelected", ""));
> 				}
> ```

I see based on the test that this is handled somewhere for `bloom-noVideoSelected`, but I don't see where.
Just curious if the logic could/should be more similar.
___
*[`src/BloomExe/web/controllers/AddOrChangePageApi.cs` line 105 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5150#-N2CUIpsAqRXW9a8OMzR:-N2CUIpsAqRXW9a8OMzS:bwn2nyu) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/a265c84d0a7f72c3455805d58567f46fe41c7a1f/src/BloomExe/web/controllers/AddOrChangePageApi.cs#L105)):*
> ```Smalltalk
> 				var (textCount, imageCount, videoCount, widgetCount) =
> 					HtmlDom.GetEditableDataCounts(page.GetDivNodeForThisPage());
> 				// Even if the pages have the same original template, if one has been edited to have a different number
> ```

I think this is referencing the check above for `page.IdOfFirstAncestor != ancestorPageId`?
I wonder if we even care at all about the original page template anymore?
For example, if I make a 'custom' page with picture on top and text on bottom, don't I expect it to get treated as a `Picture on Top` page?
Of course, then we would have to decide if it matters what order the elements appear on the page...



_Code quote:_
```Smalltalk
Even if the pages have the same original template
```
___
*[`src/BloomExe/web/controllers/AddOrChangePageApi.cs` line 110 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5150#-N2CVW7N6NJUIiFke8Mt:-N2CVW7OD-JAm5iCnDyM:b-jhmhv7) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/a265c84d0a7f72c3455805d58567f46fe41c7a1f/src/BloomExe/web/controllers/AddOrChangePageApi.cs#L110)):*
> ```Smalltalk
> 				if (textCount != oldTextCount || imageCount != oldImageCount || videoCount != oldVideoCount ||
> 				    widgetCount != oldWidgetCount)
> 					continue;
> ```

I'm confused about the relationship between these two `continue` statements and the `allowDataLoss` param. Under what scenario(s) do we not want to do the update if the user said they want to allow data loss?

Is it that the user only allowed data loss for this page **type**?

Ok, then I think that's ok. We are in a method named `ChangeSimilarPages...` after all.

JohnThomson (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5150)* status: 2 of 4 files reviewed, 2 unresolved discussions (waiting on @andrew-polk)

___
*[`src/BloomExe/Book/HtmlDom.cs` line 1113 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5150#-N2CWnvx7nIfb91UljFY:-N2DL9zT5SpVFPtBpg8O:b-yj5ej5) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/a265c84d0a7f72c3455805d58567f46fe41c7a1f/src/BloomExe/Book/HtmlDom.cs#L1113)):*

I see based on the test that this is handled somewhere for `bloom-noVideoSelected`, but I don't see where.
Just curious if the logic could/should be more similar.
</blockquote></details>

I redid it using more similar logic.
___
*[`src/BloomExe/web/controllers/AddOrChangePageApi.cs` line 105 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5150#-N2CUIpsAqRXW9a8OMzR:-N2DOkQ3DLAY7V4IOuME:b-u64ld4) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/a265c84d0a7f72c3455805d58567f46fe41c7a1f/src/BloomExe/web/controllers/AddOrChangePageApi.cs#L105)):*

I think this is referencing the check above for `page.IdOfFirstAncestor != ancestorPageId`?
I wonder if we even care at all about the original page template anymore?
For example, if I make a 'custom' page with picture on top and text on bottom, don't I expect it to get treated as a `Picture on Top` page?
Of course, then we would have to decide if it matters what order the elements appear on the page...
</blockquote></details>

Right. I've added a comment.
___
*[`src/BloomExe/web/controllers/AddOrChangePageApi.cs` line 110 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5150#-N2CVW7N6NJUIiFke8Mt:-N2DNseU6U-J_dVeSb95:bl656tj) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/a265c84d0a7f72c3455805d58567f46fe41c7a1f/src/BloomExe/web/controllers/AddOrChangePageApi.cs#L110)):*

I'm confused about the relationship between these two `continue` statements and the `allowDataLoss` param. Under what scenario(s) do we not want to do the update if the user said they want to allow data loss?

Is it that the user only allowed data loss for this page **type**?

Ok, then I think that's ok. We are in a method named `ChangeSimilarPages...` after all.
</blockquote></details>

Yes, the difficulty is to define 'similar' so that we can make a useful variety of similar changes, but not often surprise the user by deleting data he wants. I've tried to improve the comments.

andrew-polk (approved):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5150)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @JohnThomson)

PR #5154, opened 2022-05-17
BL-11162 Fix toasts for missing branding files

gmartin7:
* Ensure reference for 'branding.css' so branding files
   are handled correctly
* Keep missing and unneeded logo file from reporting

### DIFF
--- a/src/BloomExe/Book/Book.cs
+++ b/src/BloomExe/Book/Book.cs
@@ -1422,6 +1422,8 @@ namespace Bloom.Book
 			// To allow migration, pretend it has no ID if there is not yet a meta.json.
 			bookDOM.RemoveMetaElement("bloomBookId", () => (RobustFile.Exists(BookInfo.MetaDataPath) ? BookInfo.Id : null),
 				val => BookInfo.Id = val);
+			// BL-11162 Repair existing books, if they didn't have this reference.
+			bookDOM.AddStyleSheetIfMissing("branding.css");
 
 			// Title should be replicated in json
 			//if (!string.IsNullOrWhiteSpace(Title)) // check just in case we somehow have more useful info in json.
--- a/src/BloomExe/web/BloomServer.cs
+++ b/src/BloomExe/web/BloomServer.cs
@@ -474,13 +474,17 @@ namespace Bloom.Api
 				return true;
 			}
 
-			if (localPath.StartsWith(OriginalImageMarker) && IsImageTypeThatCanBeDegraded(localPath))
-			{
+			if (localPath.StartsWith(OriginalImageMarker)) {
 				// Path relative to simulated page file, and we want the file contents without modification.
 				// (Note that the simulated page file's own URL starts with this, so it's important to check
 				// for that BEFORE we do this check.)
+				// BL-11162 If we get here with the 'OriginalImageMarker' prefix and it's not an image type
+				// that can be degraded, there's no point in continuing on with the prefix!
 				localPath = localPath.Substring(OriginalImageMarker.Length + 1);
-				return ProcessAnyFileContent(info, localPath);
+				if (IsImageTypeThatCanBeDegraded(localPath))
+				{
+					return ProcessAnyFileContent(info, localPath);
+				}
 			}
 
 			if (localPath.StartsWith("localhost/", StringComparison.InvariantCulture))
@@ -1415,9 +1419,16 @@ namespace Bloom.Api
 				return false;
 
 			var localPath = GetLocalPathWithoutQuery(info);
-			// We don't need even a toast for missing files in the book folder. That's the user's problem and should be adequately
-			// documented by the browser message saying the file is missing.
-			if (currentBookFolderPath != null && localPath.StartsWith(currentBookFolderPath.Replace("\\", "/")))
+			var localFolderTestPath = localPath;
+			// We don't need even a toast for missing files in the book folder. That's the user's problem
+			// and should be adequately documented by the browser message saying the file is missing.
+			// BL-11162 This includes showing up here with "OriginalImages" prefixed to the url for
+			// publishing.
+			if (localFolderTestPath.StartsWith(OriginalImageMarker))
+			{
+				localFolderTestPath = localFolderTestPath.Substring(OriginalImageMarker.Length + 1);
+			}
+			if (currentBookFolderPath != null && localFolderTestPath.StartsWith(currentBookFolderPath.Replace("\\", "/")))
 				return false;
 			// Likewise if it's part of the current book we're publishing. If we didn't give a message about something being
 			// missing while creating the book, it's just confusing to do so when they create a publication preview. See BL-9738
--- a/src/content/templates/bloom-foundation-mixins.pug
+++ b/src/content/templates/bloom-foundation-mixins.pug
@@ -19,6 +19,7 @@ mixin core-stylesheets
 	+stylesheet('editMode.css')
 	+stylesheet('editPaneGlobal.css')
 	+stylesheet('origami.css')
+	+stylesheet('branding.css')
 
 mixin stylesheets(nameOfTemplateSpecificCSS)
 	+core-stylesheets


andrew-polk (changes requested):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5154)* status: all files reviewed, 1 unresolved discussion (waiting on @gmartin7)

___
*[`src/content/templates/bloom-foundation-mixins.pug` line 22 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5154#-N2ICTqyBGrIWb0gtD-5:-N2ICTqz5_Dfaq3T-kvm:bmmgwua) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/a06cd3f6b111d5788436b0df3505dcfe840fa6cc/src/content/templates/bloom-foundation-mixins.pug#L22)):*
> ```Pug
> 	+stylesheet('editPaneGlobal.css')
> 	+stylesheet('origami.css')
> 	+stylesheet('branding.css')
> ```

If I understand correctly, one downside to including this in every html file is that bloom-player will ask the server for the file for every book. That is one more (failed) web request on blorg.

Do we have any other alternatives?

gmartin7 (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5154)* status: all files reviewed, 2 unresolved discussions (waiting on @andrew-polk)

___
*[`src/BloomExe/Book/Book.cs` line 1426 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5154#-N2IdmFzAgjbu1n6afOL:-N2IdmFzAgjbu1n6afOM:b-jcupys) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/a06cd3f6b111d5788436b0df3505dcfe840fa6cc/src/BloomExe/Book/Book.cs#L1426)):*
> ```Smalltalk
> 				val => BookInfo.Id = val);
> 			// BL-11162 Repair existing books, if they didn't have this reference.
> 			bookDOM.AddStyleSheetIfMissing("branding.css");
> ```

This is the one that really makes sure any book has a reference to `branding.css`. Like you said below, it'd be nice to not have this. The real problem is that the `Traditional XMatter` template references a `full-bloom-logo-grey.svg` file that only exists in the various branding folders.
And there's this code in BloomServer that activates the Branding Placeholder:
```
// When developing brandings, it's convenient to just use the very-latest build
// which will be sitting in the output/browser/branding/<branding name>/ folder,
// rather than the one that is already in the book.  It's also possible that the
// branding hasn't even been developed beyond assigning a license key.
if (fileName == "branding.css")
{
	path = _fileLocator.GetBrandingFile(true, "branding.css");
	if (String.IsNullOrEmpty(path))
		return UsePlaceHolderBrandingCss(info);
}
```
___
*[`src/content/templates/bloom-foundation-mixins.pug` line 22 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5154#-N2ICTqyBGrIWb0gtD-5:-N2IddzP0tjdgYErhrdF:b-q64pkt) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/a06cd3f6b111d5788436b0df3505dcfe840fa6cc/src/content/templates/bloom-foundation-mixins.pug#L22)):*

If I understand correctly, one downside to including this in every html file is that bloom-player will ask the server for the file for every book. That is one more (failed) web request on blorg.

Do we have any other alternatives?
</blockquote></details>

See above.

gmartin7 (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5154)* status: all files reviewed, 2 unresolved discussions (waiting on @andrew-polk)

___
*[`src/BloomExe/Book/Book.cs` line 1426 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5154#-N2IdmFzAgjbu1n6afOL:-N2Iwt2C4Op3p4ZiHSxN:b-vaknhp) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/a06cd3f6b111d5788436b0df3505dcfe840fa6cc/src/BloomExe/Book/Book.cs#L1426)):*

This is the one that really makes sure any book has a reference to `branding.css`. Like you said below, it'd be nice to not have this. The real problem is that the `Traditional XMatter` template references a `full-bloom-logo-grey.svg` file that only exists in the various branding folders.
And there's this code in BloomServer that activates the Branding Placeholder:
```
// When developing brandings, it's convenient to just use the very-latest build
// which will be sitting in the output/browser/branding/<branding name>/ folder,
// rather than the one that is already in the book.  It's also possible that the
// branding hasn't even been developed beyond assigning a license key.
if (fileName == "branding.css")
{
	path = _fileLocator.GetBrandingFile(true, "branding.css");
	if (String.IsNullOrEmpty(path))
		return UsePlaceHolderBrandingCss(info);
}
```
</blockquote></details>

Of course, this means if there's no `branding.css` file referenced, you don't get the placeholder.

andrew-polk (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5154)* status: all files reviewed, 2 unresolved discussions (waiting on @andrew-polk and @gmartin7)

___
*[`src/BloomExe/Book/Book.cs` line 1426 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5154#-N2IdmFzAgjbu1n6afOL:-N2J5EETCmm0P2Z2ulPQ:bh2lfqf) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/a06cd3f6b111d5788436b0df3505dcfe840fa6cc/src/BloomExe/Book/Book.cs#L1426)):*

Of course, this means if there's no `branding.css` file referenced, you don't get the placeholder.
</blockquote></details>

If I understand, it was a faulty assumption that the asking for branding.css could be used to trigger this code (since not all books ask for it).
Is there anything else we can use instead? 
If nothing pops out, we can discuss with JH tomorrow.

gmartin7 (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5154)* status: 1 of 3 files reviewed, 2 unresolved discussions (waiting on @andrew-polk)

___
*[`src/BloomExe/Book/Book.cs` line 1426 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5154#-N2IdmFzAgjbu1n6afOL:-N2NG-mp8iPY9jO2k9EH:bvb3o95) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/a06cd3f6b111d5788436b0df3505dcfe840fa6cc/src/BloomExe/Book/Book.cs#L1426)):*

If I understand, it was a faulty assumption that the asking for branding.css could be used to trigger this code (since not all books ask for it).
Is there anything else we can use instead? 
If nothing pops out, we can discuss with JH tomorrow.
</blockquote></details>

So that code I pointed out above is in ProcessCssFile(). I think it is correct for BloomServer to look for `branding.css` IF there is an 'active' branding. I think my mistake was just to add a reference to it in ALL cases. But if a book we're opening doesn't have the reference and there IS an active branding, we should probably add it. How does that sound?

andrew-polk (changes requested):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5154)* status: all files reviewed, 1 unresolved discussion (waiting on @gmartin7)

___
*[`src/BloomExe/Book/Book.cs` line 1426 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5154#-N2IdmFzAgjbu1n6afOL:-N2SQU4l81lO729hnVlk:b25qldo) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/a06cd3f6b111d5788436b0df3505dcfe840fa6cc/src/BloomExe/Book/Book.cs#L1426)):*

So that code I pointed out above is in ProcessCssFile(). I think it is correct for BloomServer to look for `branding.css` IF there is an 'active' branding. I think my mistake was just to add a reference to it in ALL cases. But if a book we're opening doesn't have the reference and there IS an active branding, we should probably add it. How does that sound?
</blockquote></details>

I guess I'm missing something. Somewhere we add branding.css link to the html for a book if it has a real branding. Should we change the logic in that place to add it for this pseudo-branding case?

gmartin7 (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5154)* status: 2 of 4 files reviewed, all discussions resolved (waiting on @andrew-polk)

___
*[`src/BloomExe/Book/Book.cs` line 1426 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5154#-N2IdmFzAgjbu1n6afOL:-N2TGFWW4eX71EZyNt-E:b-c8rl1e) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/a06cd3f6b111d5788436b0df3505dcfe840fa6cc/src/BloomExe/Book/Book.cs#L1426)):*

I guess I'm missing something. Somewhere we add branding.css link to the html for a book if it has a real branding. Should we change the logic in that place to add it for this pseudo-branding case?
</blockquote></details>

OK. That's better. I've taken my code out of book migration (bring up to date) and modified the code that adds `branding.css` to depend on enterprise being active instead of the presence of an actual css file. That activates the "we don't have the files yet" message.

JohnThomson (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5154)* status: 1 of 6 files reviewed, 1 unresolved discussion (waiting on @andrew-polk, @gmartin7, and @JohnThomson)

___
*[`src/BloomExe/Book/Book.cs` line 4164 at r3](https://reviewable.io/reviews/bloombooks/bloomdesktop/5154#-N3UsTl48zHpVUExB8OQ:-N3_Cmba6zQ7VZK15Xxe:b-kfy3nk) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/5fc2cca414a6ac85c131225509e590c9f2ea90c6/src/BloomExe/Book/Book.cs#L4164)):*

This is the part I'm not too happy about. It works, but it seems too hacky.
</blockquote></details>

Can we simplify? The outside back cover always contains some branding or substitute branding, so it should always be published, so maybe we just look for an element with the class that indicates the outside back cover (and adjust the comment in point 3 above.

gmartin7 (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5154)* status: 1 of 6 files reviewed, 1 unresolved discussion (waiting on @andrew-polk and @JohnThomson)

___
*[`src/BloomExe/Book/Book.cs` line 4164 at r3](https://reviewable.io/reviews/bloombooks/bloomdesktop/5154#-N3UsTl48zHpVUExB8OQ:-N3_wAa216_RA453wR_P:b-nk1m5k) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/5fc2cca414a6ac85c131225509e590c9f2ea90c6/src/BloomExe/Book/Book.cs#L4164)):*

Can we simplify? The outside back cover always contains some branding or substitute branding, so it should always be published, so maybe we just look for an element with the class that indicates the outside back cover (and adjust the comment in point 3 above.
</blockquote></details>

Perhaps. But I think I remember seeing an outside back cover and a separate outside-back-cover-branding-bottom-html data-book element on different pages once (due to device xmatter, maybe?)

JohnThomson (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5154)* status: 1 of 6 files reviewed, 1 unresolved discussion (waiting on @andrew-polk, @gmartin7, and @JohnThomson)

___
*[`src/BloomExe/Book/Book.cs` line 4164 at r3](https://reviewable.io/reviews/bloombooks/bloomdesktop/5154#-N3UsTl48zHpVUExB8OQ:-N3_xT075tBl6UkYHSBr:bbtgq6t) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/5fc2cca414a6ac85c131225509e590c9f2ea90c6/src/BloomExe/Book/Book.cs#L4164)):*

Perhaps. But I think I remember seeing an outside back cover and a separate outside-back-cover-branding-bottom-html data-book element on different pages once (due to device xmatter, maybe?)
</blockquote></details>

Then what if we just put the bloom-force-publish class on the xmatter elements that have  @data-book='outside-back-cover-branding-bottom-html'? And write a more general rule that if ANY element on the page has that class we don't skip the page.

gmartin7 (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5154)* status: 1 of 6 files reviewed, all discussions resolved (waiting on @andrew-polk and @JohnThomson)

___
*[`src/BloomExe/Book/Book.cs` line 4164 at r3](https://reviewable.io/reviews/bloombooks/bloomdesktop/5154#-N3UsTl48zHpVUExB8OQ:-N3_zrQ066tspesQDOBJ:bojur50) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/5fc2cca414a6ac85c131225509e590c9f2ea90c6/src/BloomExe/Book/Book.cs#L4164)):*

That could work.
</blockquote></details>

Since we want branding elements to always be published.

gmartin7:
John Thomson created/merged a different PR to do this.

PR #5166, opened 2022-05-23
Fix missing simpleComprehensionQuiz.js (BL-11221)

andrew-polk:
This change is [Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5166)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -158,5 +158,3 @@ src/BloomBrowserUI/bloomWebFonts.css
 src/BloomBrowserUI/templates/template books/Activity/domActivities.css
 
 src/BloomBrowserUI/templates/template books/Activity/choiceActivities.css
-
-src/content/templates/template books/Activity/simpleComprehensionQuiz.js
--- a/src/BloomBrowserUI/gulpfile.js
+++ b/src/BloomBrowserUI/gulpfile.js
@@ -32,9 +32,7 @@ var child_process = require("child_process");
 const version = engines.node;
 if (!semver.satisfies(process.version, version)) {
     console.log(
-        `Required node version ${version} not satisfied with current version ${
-            process.version
-        }.`
+        `Required node version ${version} not satisfied with current version ${process.version}.`
     );
     process.exit(1);
 }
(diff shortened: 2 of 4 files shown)


andrew-polk (commented):
Note, I went with the approach of having ts not transpile into the scr dir at all but only into output directly.
I realized afterward that I probably could have just changed the order of the `content` sub-builds to do the ts->js transpile first and then copy which should also fix the issue, in theory.
But I figure this is cleaner overall.

> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5166#-:-N2nBw495JIMKMyTBD_3:b-4ahola)* status: 0 of 4 files reviewed, 3 unresolved discussions

___
*[`src/BloomBrowserUI/gulpfile.js` line 75 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5166#-N2nBhJ90SaU4_ZVuYmZ:-N2nBhJ90SaU4_ZVuYm_:beo7sav) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/d996886aa74265c62b63abedc3386821e97dfa00/src/BloomBrowserUI/gulpfile.js#L75)):*
> ```JavaScript
>         "!./**/*.bat",
>         "!./**/node_modules/**/*.*",
>         "!./**/tsconfig.json"
> ```

Not directly related to this change...
This prevents tsconfig.json from being installed.
___
*[`src/content/package.json` line 18 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5166#-N2nBnfT-jG1HmHFEWbY:-N2nBnfT-jG1HmHFEWbZ:beo7sav) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/d996886aa74265c62b63abedc3386821e97dfa00/src/content/package.json#L18)):*
> ```JSON
>         "// there is no current way to exclude node_modules, so we can't just do the whole directory.": "https://github.com/jonycheung/deadsimple-less-watch-compiler/issues/72",
>         "watch:less": "npm-run-all --parallel watch:branding:less watch:templates:less watch:layout:less",
>         "build:templates:files": "cpx \"templates/**/!(tsconfig).{png,jpg,svg,css,json,htm,html,txt,js}\" \"../../output/browser/templates\" -v",
> ```

Not directly related to this change...
This prevents tsconfig.json from being installed.
___
*[`src/content/templates/template books/Activity/tsconfig.json` line 54 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5166#-N2nBs01BSevjwHHjyVz:-N2nBs01BSevjwHHjyW-:b-udzhbj) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/d996886aa74265c62b63abedc3386821e97dfa00/src/content/templates/template%20books/Activity/tsconfig.json#L54)):*
> ```jsonc
>         // "sourceMap": true,                                /* Create source map files for emitted JavaScript files. */
>         // "outFile": "./",                                  /* Specify a file that bundles all outputs into one JavaScript file. If `declaration` is true, also designates a file that bundles all .d.ts output. */
>         "outDir": "../../../../../output/browser/templates/template books/Activity" /* Specify an output folder for all emitted files. */,
> ```

This is the only real change in this file.

PR #5170, opened 2022-05-25
Fix tests that expect BBUD to keep data not in data-div

JohnThomson:
A recent change broke these tests, which make the very dubious assumption that BringBookUpToDate will retain data-book data that is in content pages even if it has never been copied to the data-div.

---
This change is [Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5170)


andrew-polk (approved):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5170)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @JohnThomson)

PR #5179, opened 2022-05-31
Fix some problems with video recording (BL-11243)

JohnThomson:
As well as the problem with Play Recording, this fixes an issue that could arise if the user's temp folder has a space in its path.

---
This change is [Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5179)

### DIFF
--- a/src/BloomExe/Publish/Video/PublishAudioVideoAPI.cs
+++ b/src/BloomExe/Publish/Video/PublishAudioVideoAPI.cs
@@ -286,7 +286,7 @@ namespace Bloom.Publish.Video
 			_recordVideoWindow.SetPageReadTime(request.CurrentBook.BookInfo.PublishSettings.AudioVideo.PageTurnDelayDouble.ToString());
 			_recordVideoWindow.SetVideoSettingsFromPreview(request.CurrentBook.BookInfo.PublishSettings.AudioVideo.PlayerSettings);
 			_recordVideoWindow.SetPageRange(request.CurrentBook.BookInfo.PublishSettings.AudioVideo.PageRange);
-			_recordVideoWindow.Closed += (sender, args) =>
+			_recordVideoWindow.FinishedProcessingRecording += (sender, args) =>
 			{
 				if (!_recordVideoWindow.GotFullRecording)
 				{
--- a/src/BloomExe/Publish/Video/RecordVideoWindow.cs
+++ b/src/BloomExe/Publish/Video/RecordVideoWindow.cs
@@ -33,6 +33,7 @@ namespace Bloom.Publish.Video
 		private string _ffmpegPath;
 		private TempFile _htmlFile;
 		private TempFile _initialVideo;
+		private TempFile _capturedVideo;
 		private TempFile _finalVideo;
 		private bool _recording = true;
 		private bool _saveReceived;
@@ -67,6 +68,13 @@ namespace Bloom.Publish.Video
 			var dummy = Handle; // force handle to be created
 			var dummy2 = _content.Handle;
 		}
+
+		// Used for various kinds of cleanup, previously hooked to the Close event.
+		// In normal recording, this is raised when we are done with the post-processing
+		// of a recording (which now happens distinctly after this window is closed).
+		// It also gets raised if the user aborts the recording by closing the
+		// recording window.
+		public event EventHandler FinishedProcessingRecording;
 
 		public static RecordVideoWindow Create(BloomWebSocketServer webSocketServer)
 		{
@@ -280,7 +288,7 @@ namespace Bloom.Publish.Video
 				+ $"-i title=\"{Text}\" " // identifies the window for gdigrab
 				
 				+ videoArgs
-				+ _videoOnlyPath; // the intermediate output file for the recording.
+				+ "\"" +_videoOnlyPath + "\""; // the intermediate output file for the recording.
 			//Debug.WriteLine("ffmpeg capture args: " + args);
 			RunFfmpeg(args);
 			_startTime = DateTime.Now;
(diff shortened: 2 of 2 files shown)


andrew-polk (commented):
Oops. I didn't realize this had been merged when I started reviewing.
I'll publish anyway in case anything I said seems worth making changes for...

> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5179#-:-N3Q1fjUA5He8WTIUUrO:b-te9pqf)* status: all files reviewed, 2 unresolved discussions

___
*[`src/BloomExe/Publish/Video/RecordVideoWindow.cs` line 77 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5179#-N3Q-sDC17FevHanu7gY:-N3Q-sDC17FevHanu7gZ:bg267a4) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/b1e550ec19f2b4e92b50f9a72ce2077d59f3d88a/src/BloomExe/Publish/Video/RecordVideoWindow.cs#L77)):*
> ```Smalltalk
> 		// It also gets raised if the user aborts the recording by closing the
> 		// recording window.
> 		public event EventHandler FinishedProcessingRecording;
> ```

Can we come up with a name which applies better to both scenarios?

`ReadyForCleanupIfNeeded`
`FinishedOrCanceled`
`RecordedOrCanceled`
`RecordingEnded`
___
*[`src/BloomExe/Publish/Video/RecordVideoWindow.cs` line 326 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5179#-N3Q0_f3Ck1EsDOKmNQ_:-N3Q0_f3Ck1EsDOKmNQa:b-qaema2) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/b1e550ec19f2b4e92b50f9a72ce2077d59f3d88a/src/BloomExe/Publish/Video/RecordVideoWindow.cs#L326)):*
> ```Smalltalk
> 			// Make sure that OnClosed() won't dispose of anything we need.
> 			_capturedVideo = _initialVideo; // save so we can dispose eventually
> 			_initialVideo = null; // prevent automatic dispose in OnClosed
> ```

I don't claim to know much about memory management in .net, but should we be concerned about having (what seems to be) two full copies of the video in memory?

Can we achieve the same thing by instead setting a boolean here like `_userCanceled` or `_processingFinished`?

JohnThomson (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5179)* status: all files reviewed, 2 unresolved discussions

___
*[`src/BloomExe/Publish/Video/RecordVideoWindow.cs` line 77 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5179#-N3Q-sDC17FevHanu7gY:-N3QNyaK4VcGh3rzmwNT:b-pcpvgs) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/b1e550ec19f2b4e92b50f9a72ce2077d59f3d88a/src/BloomExe/Publish/Video/RecordVideoWindow.cs#L77)):*

Can we come up with a name which applies better to both scenarios?

`ReadyForCleanupIfNeeded`
`FinishedOrCanceled`
`RecordedOrCanceled`
`RecordingEnded`
</blockquote></details>

Maybe. I don't much like the last two because we want to distinguish being done recording from being done with the post-processing. And we do more than just cleanup in response, such as enabling buttons, so I'm not keen on the first. The second leaves it ambiguous what is finished. Maybe it would be better to have a separate event for Canceled? Or since it's only used in one place we could be really longwinded and call it FinishedProcessingOrCanceled?
___
*[`src/BloomExe/Publish/Video/RecordVideoWindow.cs` line 326 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5179#-N3Q0_f3Ck1EsDOKmNQ_:-N3QNOQJCP57cs7vs2YF:b5awoam) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/b1e550ec19f2b4e92b50f9a72ce2077d59f3d88a/src/BloomExe/Publish/Video/RecordVideoWindow.cs#L326)):*

I don't claim to know much about memory management in .net, but should we be concerned about having (what seems to be) two full copies of the video in memory?

Can we achieve the same thing by instead setting a boolean here like `_userCanceled` or `_processingFinished`?
</blockquote></details>

These are tempfiles. We're just moving a pointer to one small object from one variable to another, to help control when a file (which is not copied) should be deleted.

andrew-polk (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5179)* status: :shipit: complete! all files reviewed, all discussions resolved

___
*[`src/BloomExe/Publish/Video/RecordVideoWindow.cs` line 77 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5179#-N3Q-sDC17FevHanu7gY:-N3QUGXw5SOb2tw-_AEU:btsyy8w) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/b1e550ec19f2b4e92b50f9a72ce2077d59f3d88a/src/BloomExe/Publish/Video/RecordVideoWindow.cs#L77)):*

Maybe. I don't much like the last two because we want to distinguish being done recording from being done with the post-processing. And we do more than just cleanup in response, such as enabling buttons, so I'm not keen on the first. The second leaves it ambiguous what is finished. Maybe it would be better to have a separate event for Canceled? Or since it's only used in one place we could be really longwinded and call it FinishedProcessingOrCanceled?
</blockquote></details>

Let's not worry about it, especially since it is merged already.
___
*[`src/BloomExe/Publish/Video/RecordVideoWindow.cs` line 326 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5179#-N3Q0_f3Ck1EsDOKmNQ_:-N3QUJiH63LADbSvKtsM:b-k73z76) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/b1e550ec19f2b4e92b50f9a72ce2077d59f3d88a/src/BloomExe/Publish/Video/RecordVideoWindow.cs#L326)):*

These are tempfiles. We're just moving a pointer to one small object from one variable to another, to help control when a file (which is not copied) should be deleted.
</blockquote></details>

thanks

PR #5180, opened 2022-05-31
Catch and log error starting FFMPEG (BL-11192)

JohnThomson:
This change is [Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5180)


andrew-polk:
Merging based on JH's intention to do so, based on his comment in the card.

PR #5201, opened 2022-06-06
Don't allow files in the preview pane to be cached (BL-11239)

JohnThomson:
This change is [Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5201)

### DIFF
--- a/src/BloomExe/CollectionTab/CollectionTabView.cs
+++ b/src/BloomExe/CollectionTab/CollectionTabView.cs
@@ -105,15 +105,6 @@ namespace Bloom.CollectionTab
 				return;
 			if (book.IsEditable)
 				_model.UpdateThumbnailAsync(book);
-			// This may help to free up memory. More importantly, we allow files that are not in
-			// the current collection to be cached by the browser. The preview iframe uses urls
-			// like /book-preview/index.htm, which means urls inside it like "src='image.jpg'"
-			// translate to something like /book-preview/image.jpg, which is not book-specific.
-			// If the preview is of a book not in the editable collection, we allow files to be
-			// cached. So if we don't clear the cache when switching books, we could use an image
-			// from one book when displaying another book. And pathologically, it might not
-			// actually be the same image! See BL-11239.
-			Browser.ClearCache();
 			book.ContentsChanged += (sender, args) =>
 			{
 				if (_tabSelection.ActiveTab == WorkspaceTab.collection)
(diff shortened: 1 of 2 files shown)


andrew-polk (approved):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5201)* status: all files reviewed, 1 unresolved discussion (waiting on @JohnThomson)

___
*[`src/BloomExe/CollectionTab/CollectionTabView.cs` line 116 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5201#-N3v4IZj3K5mEvXyttIu:-N3v4IZj3K5mEvXyttIv:b-3yy69e) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/7130ff05bd5c30f19da8db6dca17a58bf3755f0c/src/BloomExe/CollectionTab/CollectionTabView.cs#L116)):*
> ```Smalltalk
> 			// from one book when displaying another book. And pathologically, it might not
> 			// actually be the same image! See BL-11239.
> 			Browser.ClearCache();
> ```

Maybe we don't want to keep this in as attached to this issue since it doesn't solve it. But do we want to consider doing it if it may free up memory? Or was that just pie in the sky thinking that we may get a secondary gain from this?

andrew-polk (approved):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5201)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @JohnThomson)

___
*[`src/BloomExe/CollectionTab/CollectionTabView.cs` line 116 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5201#-N3v4IZj3K5mEvXyttIu:-N3yhwIO23XRroE_HAL5:b-4ehkhd) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/7130ff05bd5c30f19da8db6dca17a58bf3755f0c/src/BloomExe/CollectionTab/CollectionTabView.cs#L116)):*

Maybe we don't want to keep this in as attached to this issue since it doesn't solve it. But do we want to consider doing it if it may free up memory? Or was that just pie in the sky thinking that we may get a secondary gain from this?
</blockquote></details>

Answered in card: 
I have no definite information about why clearing the cache doesn't fix it. It's possible that this API was planned but never actually implemented in GeckoFx. It's possible that it clears some other cache besides the cache of recently downloaded URLs...for example, it might only clear stuff related to pages that have been closed, or a cache of bitmaps but not downloaded objects. All I know is, by actual test, it didn't fix the problem.
The possibility of saving some memory was a very secondary issue in the original PR. It may be worth doing for that reason, assuming it actually does anything.
There are other things we could do. It's our decision that all previews are presented to the browser as /book-preview/index.htm. We could instead use something like /book-preview/guid.htm, with a new guid each time we change selection, or include the book folder name. Whether that would result in any useful caching, I'm not sure. It struck me as a considerably riskier change, with the possibility of missing some bit of code that requires the preview to be at the known URL.

PR #5206, opened 2022-06-07
Publish to video: Suggest filename based on language chooser (BL-11081)

sujeffreyl:
This change is [Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5206)

### DIFF
--- a/src/BloomExe/Book/Book.cs
+++ b/src/BloomExe/Book/Book.cs
@@ -2197,7 +2197,7 @@ namespace Bloom.Book
 			}
 		}
 
-		public BookInfo BookInfo { get; protected set; }
+		public virtual BookInfo BookInfo { get; protected set; }
 
 		public UserPrefs UserPrefs { get; private set; }
 
--- a/src/BloomExe/Book/BookInfo.cs
+++ b/src/BloomExe/Book/BookInfo.cs
@@ -221,7 +221,7 @@ namespace Bloom.Book
 		/// This stores a Json string representing lang:title pairs, e.g.,
 		/// {"en":"my nice title","de":"Mein schönen Titel","es":"мy buen título"}.
 		/// </summary>
-		public string AllTitles
+		public virtual string AllTitles
 		{
 			get { return MetaData.AllTitles; }
 			set { MetaData.AllTitles = value; }
--- a/src/BloomExe/Book/PublishSettings.cs
+++ b/src/BloomExe/Book/PublishSettings.cs
@@ -267,7 +267,7 @@ namespace Bloom.Book
 		/// is intended to allow more relevant controls to be added to BP without changing
 		/// Bloom Editor.
 		/// One downside is that quotes in the PlayerSettings JSON must be escaped in the
-		/// file. We think is is worth this to keep the content of this string opaque to BE.
+		/// file. We think it is worth this to keep the content of this string opaque to BE.
 		/// </summary>
 		[JsonProperty("playerSettings")]
 		public string PlayerSettings { get; set; }
--- a/src/BloomExe/Publish/Video/PublishAudioVideoAPI.cs
+++ b/src/BloomExe/Publish/Video/PublishAudioVideoAPI.cs
@@ -196,6 +196,9 @@ namespace Bloom.Publish.Video
 					// This shouldn't be possible, but just in case, we'll kick off the recording now.
 					RecordVideo(request);
 				}
+
+				_recordVideoWindow?.SetBook(request.CurrentBook);
+
 				// If we bring up the dialog inside this API call, it may time out.
 				Application.Idle += SaveVideoOnIdle;
 				request.PostSucceeded();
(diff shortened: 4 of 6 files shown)


andrew-polk (changes requested):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5206)* status: all files reviewed, 2 unresolved discussions (waiting on @sujeffreyl)

___
*[`src/BloomExe/Publish/Video/RecordVideoWindow.cs` line 910 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5206#-N3zKUsb6Mzzay4E1Kqq:-N3zKUsc41oXXkJVNjU8:b-77dfbd) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/85fe1f3953b343e48ac8391102769d80ce194f44/src/BloomExe/Publish/Video/RecordVideoWindow.cs#L910)):*
> ```Smalltalk
> 				}
> 
> 				// In case of any unexpected errors above, just fallback to the legacy way which just uses the path to the book.
> ```

Would it be worth wrapping the second part in a try/catch as well? (or extending the scope of the existing one?)



_Code quote:_
```Smalltalk
In case of any unexpected errors above
```
___
*[`src/BloomExe/Publish/Video/RecordVideoWindow.cs` line 914 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5206#-N3zKGbd3lujlDgHFhKM:-N3zKGbd3lujlDgHFhKN:be5q6br) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/85fe1f3953b343e48ac8391102769d80ce194f44/src/BloomExe/Publish/Video/RecordVideoWindow.cs#L914)):*
> ```Smalltalk
> 
> 			// Default method
> 			return string.Format($"{Path.GetFileName(_pathToRealBook)}");
> ```

Can be simplified to `return Path.GetFileName(_pathToRealBook);`



_Code quote:_
```Smalltalk
return string.Format($"{Path.GetFileName(_pathToRealBook)}");
```

sujeffreyl (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5206)* status: 5 of 6 files reviewed, 2 unresolved discussions (waiting on @andrew-polk)

___
*[`src/BloomExe/Publish/Video/RecordVideoWindow.cs` line 910 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5206#-N3zKUsb6Mzzay4E1Kqq:-N4-BIOZ2dvrFMZVhaIc:b-w9ks37) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/85fe1f3953b343e48ac8391102769d80ce194f44/src/BloomExe/Publish/Video/RecordVideoWindow.cs#L910)):*

Would it be worth wrapping the second part in a try/catch as well? (or extending the scope of the existing one?)
</blockquote></details>

Hmm yeah, I went back and forth on what to do about this... but I felt like parsing the videoSettings stuff was an area that I felt I could put less guarantees on it actually working consistently, and wasn't really show-stopping if it failed.
However, for the 2nd part, I felt like I really am not expecting or wanting exceptions out of that, and that AllTitles was more reliable/it would be problematic if AllTitles were in an invalid format, and if there was a problem for the 2nd half, that I should be on the hook for that / would want to get an exception for that.

So that's my train of thought. It's definitely not the only way to approach it though. Thoughts?
___
*[`src/BloomExe/Publish/Video/RecordVideoWindow.cs` line 914 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5206#-N3zKGbd3lujlDgHFhKM:-N4-AoD97rBBZPk6y1Bv:b-wqfoo0) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/85fe1f3953b343e48ac8391102769d80ce194f44/src/BloomExe/Publish/Video/RecordVideoWindow.cs#L914)):*

Can be simplified to `return Path.GetFileName(_pathToRealBook);`
</blockquote></details>

Oh yeah, that was copy/paste from the existing code... but I'll update it.

andrew-polk (approved):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5206)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @sujeffreyl)

___
*[`src/BloomExe/Publish/Video/RecordVideoWindow.cs` line 910 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5206#-N3zKUsb6Mzzay4E1Kqq:-N4-KoPp0W3Fbky-xvGy:b-84vczj) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/85fe1f3953b343e48ac8391102769d80ce194f44/src/BloomExe/Publish/Video/RecordVideoWindow.cs#L910)):*

Hmm yeah, I went back and forth on what to do about this... but I felt like parsing the videoSettings stuff was an area that I felt I could put less guarantees on it actually working consistently, and wasn't really show-stopping if it failed.
However, for the 2nd part, I felt like I really am not expecting or wanting exceptions out of that, and that AllTitles was more reliable/it would be problematic if AllTitles were in an invalid format, and if there was a problem for the 2nd half, that I should be on the hook for that / would want to get an exception for that.

So that's my train of thought. It's definitely not the only way to approach it though. Thoughts?
</blockquote></details>

I'll not stop that train of thought.

PR #5209, opened 2022-06-08
Populate stack trace in fake error report dialogs

sujeffreyl:
All this does is give slightly more realistic behavior in the fake error report dialogs used for development/testing

---
This change is [Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5209)

### DIFF
--- a/src/BloomExe/ErrorReporter/ErrorReportUtils.cs
+++ b/src/BloomExe/ErrorReporter/ErrorReportUtils.cs
@@ -97,7 +97,17 @@ namespace Bloom.ErrorReporter
 		private static void CheckForFakeTestErrors(string title)
 		{
 			const string fakeProblemMessage = "Fake problem for development/testing purposes";
-			var fakeException = new ApplicationException("Fake exception for development/testing purposes");
+			Exception fakeException;
+
+			// Throwing/catching the exception populates the stack trace
+			try
+			{
+				throw new ApplicationException("Fake exception for development/testing purposes");
+			}
+			catch (ApplicationException e)
+			{
+				fakeException = e;
+			}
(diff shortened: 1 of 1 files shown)


andrew-polk (approved):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5209)* status: all files reviewed, 1 unresolved discussion (waiting on @sujeffreyl)

___
*[`src/BloomExe/ErrorReporter/ErrorReportUtils.cs` line 105 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5209#-N42ggriBUkovHz3uHsl:-N44Qmc5683rbXE1HjlJ:b-gx568p) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/72c68f10cc4230b8c180491dca9e3ce5925e6dc9/src/BloomExe/ErrorReporter/ErrorReportUtils.cs#L105)):*

Technically, one downside is that all Debug users as well as Release users w/BloomSandbox on will get a slight theoretical perf hit due to throwing and catching an unnecessary exception.
However, these aren't "real" users, so I think it's okay.

If we wanted to avoid that, we could throw/catch this exception a bunch of times within the if statements. My thought was that this was unnecessary.
</blockquote></details>

Other ideas are reflection or extending ApplicationException to set the stacktrace but I agree it doesn't matter enough to bother and complicate things.

andrew-polk (approved):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5209)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @sujeffreyl)

PR #5238, opened 2022-06-16
Prevent previewMode.css from being added after removal (BL-11330)

StephenMcConnel:


andrew-polk (changes requested):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5238)* status: all files reviewed, 1 unresolved discussion (waiting on @StephenMcConnel)

___
*[`src/BloomExe/Book/BookStorage.cs` line 2451 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5238#-N4hoFm0DFDZ3TciBnkf:-N4hoFm0DFDZ3TciBnkg:b-u1z1iz) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/9c74f90cf51f32de6d476aa70ae093ba523b43a0/src/BloomExe/Book/BookStorage.cs#L2451)):*
> ```Smalltalk
> 		// such time as it's clear if it matters.
> 		public readonly static string[] CssFilesToLink =
> 			{ "basePage.css", "previewMode.css", "origami.css", "langVisibility.css", "branding.css" };
> ```

Since this list of css files is used in common code, can you help me understand why removing this doesn't break the book preview or PDF gen (which I think uses preview mode?)?

StephenMcConnel (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5238)* status: all files reviewed, 1 unresolved discussion (waiting on @andrew-polk)

___
*[`src/BloomExe/Book/BookStorage.cs` line 2451 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5238#-N4hoFm0DFDZ3TciBnkf:-N4hy2qz66wOnk0vIf4k:bquppgm) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/9c74f90cf51f32de6d476aa70ae093ba523b43a0/src/BloomExe/Book/BookStorage.cs#L2451)):*

Since this list of css files is used in common code, can you help me understand why removing this doesn't break the book preview or PDF gen (which I think uses preview mode?)?
</blockquote></details>

The only places this is used is in publishing/piecing together a "folio" book and in publishing a BloomPub book.  In both places, it's used to add the standard set of .css files immediately after explicitly removing any links to editMode.css and previewMode.css.  It doesn't make sense to add something back in that we just went to the trouble of removing.
In Book.cs,  GetPreviewHtmlFileForWholeBook() and GetDomForPrinting() have these lines, which insert this stylesheet link where it is needed.



_Code snippet:_
```Smalltalk
var previewDom= GetBookDomWithStyleSheets("previewMode.css", "origami.css");
var printingDom = GetBookDomWithStyleSheets("previewMode.css", "origami.css");
```

andrew-polk (changes requested):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5238)* status: all files reviewed, 1 unresolved discussion (waiting on @StephenMcConnel)

___
*[`src/BloomExe/Book/BookStorage.cs` line 2451 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5238#-N4hoFm0DFDZ3TciBnkf:-N4i4AXKElvIFPgGtgpS:b6yu6db) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/9c74f90cf51f32de6d476aa70ae093ba523b43a0/src/BloomExe/Book/BookStorage.cs#L2451)):*

The only places this is used is in publishing/piecing together a "folio" book and in publishing a BloomPub book.  In both places, it's used to add the standard set of .css files immediately after explicitly removing any links to editMode.css and previewMode.css.  It doesn't make sense to add something back in that we just went to the trouble of removing.
In Book.cs,  GetPreviewHtmlFileForWholeBook() and GetDomForPrinting() have these lines, which insert this stylesheet link where it is needed.
</blockquote></details>

Thanks.

But now I'm confused about something else...
The rule which is causing this in previewMode.less is
```
.bloomPlayer-page.template {...}
```
The `.bloomPlayer-page` class is added in bloom-player itself. So doesn't that mean we, at least at one point, expected this stylesheet to be present in the bloom-player context and intended this behavior?

StephenMcConnel:
This problem was fixed elsewhere.

PR #5239, opened 2022-06-16
[5.3] BL-11329 Fix perf measurement on book selection

gmartin7:


gmartin7 (commented):
Is it important that Perf. Meas. only happen in Team Collections (as per the issue)? This fix does the check regardless of TC status.

> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5239#-:-N4iXB5c2kPK9uy67Fie:bekq8dd)* status: 0 of 1 files reviewed, all discussions resolved

hatton (changes requested):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5239)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @gmartin7 and @hatton)

___
*[`src/BloomExe/Workspace/WorkspaceView.cs` line 292 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5239#-N4iXmIX1t_lvc-ZGR0s:-N4iXmIX1t_lvc-ZGR0t:bjze1hs) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/2af03a6a595362647d68ba23b1b47afa987ca519/src/BloomExe/Workspace/WorkspaceView.cs#L292)):*
> ```Smalltalk
> 				? "No Book Selected"
> 				: _bookSelection.CurrentSelection.BookInfo.QuickTitleUserDisplay;
> 			using (PerformanceMeasurement.Global?.Measure("select book", titleString))
> ```

Yeah.... no.  The time it takes to send a websocket message is not what we want to measure.

gmartin7 (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5239)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @hatton)

___
*[`src/BloomExe/Workspace/WorkspaceView.cs` line 292 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5239#-N4iXmIX1t_lvc-ZGR0s:-N4mV0KcBH01mx38-vJX:bfvkccl) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/2af03a6a595362647d68ba23b1b47afa987ca519/src/BloomExe/Workspace/WorkspaceView.cs#L292)):*

Yeah.... no.  The time it takes to send a websocket message is not what we want to measure.
</blockquote></details>

Moved to CollectionApi.

andrew-polk (changes requested):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5239)* status: all files reviewed, 2 unresolved discussions (waiting on @gmartin7 and @hatton)

___
*[`src/BloomExe/web/controllers/CollectionApi.cs` line 87 at r2](https://reviewable.io/reviews/bloombooks/bloomdesktop/5239#-N4mXTz80TIp4H-RRZDt:-N4mXTz80TIp4H-RRZDu:b-gsow6t) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/24dcbe5a6f3bd142b2c9b2feaf34df19d12a8c75/src/BloomExe/web/controllers/CollectionApi.cs#L87)):*
> ```Smalltalk
> 							{
> 								_collectionModel.SelectBook(book);
> 							}
> ```

Sorry; I'm confused. 
You say we should include `GetUpdatedBookObjectFromBookInfo` in the measurement because it does a non-trivial amount of work, but you just introduced it in this PR.

Did you decide we need to call `GetBookIFromBookInfo` for other reasons than measuring performance? Are you fixing a bug?

Or am I just missing something?

gmartin7 (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5239)* status: all files reviewed, 2 unresolved discussions (waiting on @andrew-polk and @hatton)

___
*[`src/BloomExe/web/controllers/CollectionApi.cs` line 87 at r2](https://reviewable.io/reviews/bloombooks/bloomdesktop/5239#-N4mXTz80TIp4H-RRZDt:-N4mn-Aa-fJ3jo8ni4IO:b-j5mq1q) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/24dcbe5a6f3bd142b2c9b2feaf34df19d12a8c75/src/BloomExe/web/controllers/CollectionApi.cs#L87)):*

Sorry; I'm confused. 
You say we should include `GetUpdatedBookObjectFromBookInfo` in the measurement because it does a non-trivial amount of work, but you just introduced it in this PR.

Did you decide we need to call `GetBookIFromBookInfo` for other reasons than measuring performance? Are you fixing a bug?

Or am I just missing something?
</blockquote></details>

Sorry. The GetBookObjectFromPost() (previous code) had the fullyUpdateBookFiles param set to true. But I needed the BookInfo object to get the titleString to put in the measurement, so I split GetBookObjectFromPost into the 2 parts that make that method up. The new method does the rest of the old method and automatically sets the updateFiles param.

andrew-polk (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5239)* status: all files reviewed, 1 unresolved discussion (waiting on @hatton)

___
*[`src/BloomExe/web/controllers/CollectionApi.cs` line 87 at r2](https://reviewable.io/reviews/bloombooks/bloomdesktop/5239#-N4mXTz80TIp4H-RRZDt:-N4mpnCv2h8N4XZLvNMR:b-y0ykf3) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/24dcbe5a6f3bd142b2c9b2feaf34df19d12a8c75/src/BloomExe/web/controllers/CollectionApi.cs#L87)):*

Sorry. The GetBookObjectFromPost() (previous code) had the fullyUpdateBookFiles param set to true. But I needed the BookInfo object to get the titleString to put in the measurement, so I split GetBookObjectFromPost into the 2 parts that make that method up. The new method does the rest of the old method and automatically sets the updateFiles param.
</blockquote></details>

Thanks, I get it now. I definitely missed the dropping of the `true` parameter for `GetBookInfoFromPost`.

Please discuss with Jeffrey/JT to decide if this should be here or in `SelectBook` since they just added it there in a different PR.

If we are going with this approach, it is probably worth some more comments and/or better naming to make it explicit that we are purposefully not doing the full update before do the full update.

gmartin7 (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5239)* status: all files reviewed, 1 unresolved discussion (waiting on @hatton)

___
*[`src/BloomExe/web/controllers/CollectionApi.cs` line 87 at r2](https://reviewable.io/reviews/bloombooks/bloomdesktop/5239#-N4mXTz80TIp4H-RRZDt:-N4n2Jvx1odEM_831AAd:b-1tl8b1) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/24dcbe5a6f3bd142b2c9b2feaf34df19d12a8c75/src/BloomExe/web/controllers/CollectionApi.cs#L87)):*

Thanks, I get it now. I definitely missed the dropping of the `true` parameter for `GetBookInfoFromPost`.

Please discuss with Jeffrey/JT to decide if this should be here or in `SelectBook` since they just added it there in a different PR.

If we are going with this approach, it is probably worth some more comments and/or better naming to make it explicit that we are purposefully not doing the full update before do the full update.
</blockquote></details>

JT says `Grrrr...we'd probably like to include BBUD, since that's definitely part of the time it takes to get a book selected. We'd probably also like to measure performance every time a book is selected, not just when the request to do so came through the collectionApi. On the whole, I think the former is more important, so we should go with yours.`

gmartin7 (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5239)* status: all files reviewed, 1 unresolved discussion (waiting on @hatton)

___
*[`src/BloomExe/web/controllers/CollectionApi.cs` line 87 at r2](https://reviewable.io/reviews/bloombooks/bloomdesktop/5239#-N4mXTz80TIp4H-RRZDt:-N4n2ipu0h_daSLqAUU4:b-ygaivu) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/24dcbe5a6f3bd142b2c9b2feaf34df19d12a8c75/src/BloomExe/web/controllers/CollectionApi.cs#L87)):*

JT says `Grrrr...we'd probably like to include BBUD, since that's definitely part of the time it takes to get a book selected. We'd probably also like to measure performance every time a book is selected, not just when the request to do so came through the collectionApi. On the whole, I think the former is more important, so we should go with yours.`
</blockquote></details>

Adding some better comments now...

andrew-polk (approved):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5239)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @gmartin7)

PR #5240, opened 2022-06-16
Fix image url retrieval for C/L dialog (BL-11336)

andrew-polk:
This change is [Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5240)

### DIFF
--- a/src/BloomBrowserUI/bookEdit/editViewFrame.ts
+++ b/src/BloomBrowserUI/bookEdit/editViewFrame.ts
@@ -37,6 +37,7 @@ import "errorHandler";
 import { reportError } from "../lib/errorHandler";
 import { IToolboxFrameExports } from "./toolbox/toolboxBootstrap";
 import { showCopyrightAndLicenseInfoOrDialog } from "./copyrightAndLicense/CopyrightAndLicenseDialog";
+export { getImageUrlFromImageButton } from "./js/bloomImages";
 
 //Called by c# using editTabBundle.handleUndo()
 export function handleUndo(): void {
--- a/src/BloomBrowserUI/bookEdit/js/bloomImages.ts
+++ b/src/BloomBrowserUI/bookEdit/js/bloomImages.ts
@@ -130,14 +130,14 @@ function SetupImageContainer(containerDiv: HTMLElement) {
 
     $(containerDiv)
         .mouseenter(function() {
-            const $this = $(this);
-            let img = $this.find("img");
+            let img = getImgFromContainer(this);
             if (img.length === 0)
                 // This case is probably a left over from some previous Bloom where
                 // we were using background images instead of <img>? But it does
                 // no harm so I'm leaving it in.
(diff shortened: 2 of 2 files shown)


JohnThomson (approved):
One suggestion, but it can merge as is if you don't agree.

> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5240#-:-N4nDQmD9w33Yyc4kLEa:b-x3kmg9)* status: all files reviewed, 1 unresolved discussion (waiting on @andrew-polk)

___
*[`src/BloomBrowserUI/bookEdit/js/bloomImages.ts` line 222 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5240#-N4nBtzQ5Ibkyv1Zn3Tp:-N4nBtzREDDceN2kJp0c:b-hb5u2d) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/7759ae375ad5332574440242633cc5b09de7f964/src/BloomBrowserUI/bookEdit/js/bloomImages.ts#L222)):*
> ```XML
>                     // We have to do this strange editTabBundle stuff because at the time of the click, we are in a context where that is all we can access.
>                     `(window.parent || window).editTabBundle.showCopyrightAndLicenseDialog(
>                         (window.parent || window).editTabBundle.getImageUrlFromImageButton(this));`
> ```

I dislike such uses of 'this' where IIRC it does not mean anything obvious (since we're not in a method of a class). Can we declare an arrow function and get the object we want...is it the thing that has the onClick attr, or the (possibly child) thing that was most directly clicked, or doesn't it matter because the button has no children?...from some property of the event passed to the function?

andrew-polk (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5240)* status: all files reviewed, 1 unresolved discussion (waiting on @JohnThomson)

___
*[`src/BloomBrowserUI/bookEdit/js/bloomImages.ts` line 222 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5240#-N4nBtzQ5Ibkyv1Zn3Tp:-N4nah8M5Zbx9V-M6PcQ:b-rfv5hq) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/7759ae375ad5332574440242633cc5b09de7f964/src/BloomBrowserUI/bookEdit/js/bloomImages.ts#L222)):*

I dislike such uses of 'this' where IIRC it does not mean anything obvious (since we're not in a method of a class). Can we declare an arrow function and get the object we want...is it the thing that has the onClick attr, or the (possibly child) thing that was most directly clicked, or doesn't it matter because the button has no children?...from some property of the event passed to the function?
</blockquote></details>

In this case, "this" is the element clicked, i.e. the element with the onclick attribute.
Unfortunately, it looks like that is the best we can do in this context. I don't see a way to use an arrow function or access the event.
I could definitely just be missing it, but I've looked for a while now.

PR #5241, opened 2022-06-16
Improve UI for license restrictions on publishing (BL-11328)

StephenMcConnel:
This change is [Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5241)

### DIFF
--- a/src/BloomBrowserUI/publish/ReaderPublish/MethodChooser.tsx
+++ b/src/BloomBrowserUI/publish/ReaderPublish/MethodChooser.tsx
@@ -32,8 +32,8 @@ export const MethodChooser: React.FunctionComponent = () => {
         "publish/android/method",
         "wifi"
     );
-    const saveButtonEnabled = BloomApi.useWatchString(
-        "true",
+    const isLicenseOK = BloomApi.useWatchBooleanEvent(
+        true,
         "publish-android",
         "publish/licenseOK"
     );
@@ -62,10 +62,7 @@ export const MethodChooser: React.FunctionComponent = () => {
                             )
                         }}
                     />
-                    {getStartButton(
-                        method,
-                        saveButtonEnabled.toLowerCase() === "true"
-                    )}
+                    {getStartButton(method, isLicenseOK)}
                 </div>
                 <div className={"column2"}>
                     <img
--- a/src/BloomBrowserUI/publish/ePUBPublish/ePUBPublishScreen.tsx
+++ b/src/BloomBrowserUI/publish/ePUBPublish/ePUBPublishScreen.tsx
@@ -81,8 +81,8 @@ const EPUBPublishScreenInternal: React.FunctionComponent<{
             setClosePending(true);
         }
     );
-    const saveButtonEnabled = BloomApi.useWatchString(
-        "true",
+    const isLicenseOK = BloomApi.useWatchBooleanEvent(
+        true,
         "publish-epub",
         "publish/licenseOK"
     );
@@ -109,7 +109,7 @@ const EPUBPublishScreenInternal: React.FunctionComponent<{
                 <PublishPanel>
                     <BloomButton
                         className="save-button"
-                        enabled={saveButtonEnabled.toLowerCase() === "true"}
+                        enabled={isLicenseOK}
                         clickApiEndpoint={"publish/epub/save"}
                         hasText={true}
                         l10nKey="PublishTab.Save"
--- a/src/BloomBrowserUI/publish/video/PublishAudioVideo.tsx
+++ b/src/BloomBrowserUI/publish/video/PublishAudioVideo.tsx
@@ -153,6 +153,12 @@ const PublishAudioVideoInternalInternal: React.FunctionComponent<{
         "recording"
     );
 
+    const isLicenseOK = BloomApi.useWatchBooleanEvent(
+        true,
+        "recordVideo",
+        "publish/licenseOK"
+    );
+
     const [debouncedPageTurnDelay] = useDebounce(
         avSettings.pageTurnDelay,
         1000
@@ -378,7 +384,10 @@ const PublishAudioVideoInternalInternal: React.FunctionComponent<{
                                             justify-content: center;
                                         `}
                                     >
-                                        <Button onClick={reset}>
+                                        <Button
+                                            onClick={reset}
+                                            disabled={!isLicenseOK}
+                                        >
                                             <SkipPreviousIcon
                                                 // unfortunately this icon doesn't come in a variant with a built-in circle.
                                                 // To make it match the other two we have to shrink it, make it white,
@@ -410,7 +419,10 @@ const PublishAudioVideoInternalInternal: React.FunctionComponent<{
                                                 size="1.6rem"
                                             ></CircularProgress>
                                         ) : (
-                                            <Button onClick={play}>
+                                            <Button
+                                                onClick={play}
+                                                disabled={!isLicenseOK}
+                                            >
                                                 <PlayIcon
                                                     css={css`
                                                         color: ${kBloomBlue};
@@ -421,7 +433,10 @@ const PublishAudioVideoInternalInternal: React.FunctionComponent<{
                                         )}
                                         <Button
                                             onClick={pause}
-                                            disabled={isPauseButtonDisabled}
+                                            disabled={
+                                                isPauseButtonDisabled ||
+                                                !isLicenseOK
+                                            }
                                         >
                                             <PauseIcon
                                                 css={css`
@@ -501,7 +516,7 @@ const PublishAudioVideoInternalInternal: React.FunctionComponent<{
                                     </ErrorBox>
                                 )}
                                 <BloomButton
-                                    enabled={!recording}
+                                    enabled={!recording && isLicenseOK}
                                     l10nKey="PublishTab.RecordVideo.Record"
                                     l10nComment="'Record' as in 'Record a video recording'"
                                     clickApiEndpoint="publish/av/recordVideo"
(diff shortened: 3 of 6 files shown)


andrew-polk (changes requested):
If I remember from when I created the issue, I think the audio/video preview wasn't loading if license wasn't ok which in turn meant you couldn't change the language. Is that what you see? I don't see anything in this PR to address that. 
Ideally (I think), we would allow the preview to load so the user can change languages. If we can't do that, we shouldn't let the preview show an error.

> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5241#-:-N4ikXUo2vhEBw7wIjyy:buq86vt)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @StephenMcConnel)

andrew-polk (changes requested):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5241)* status: all files reviewed, 1 unresolved discussion (waiting on @StephenMcConnel)

___
*[`src/BloomExe/Publish/Video/PublishAudioVideoAPI.cs` line 236 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5241#-N4mhRze7awHhbzA3M6g:-N4mhRzfE6Npw-SXkPeL:b4umxv3) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/30433007a840cc8242089d91e22e1e5e2de42cbe/src/BloomExe/Publish/Video/PublishAudioVideoAPI.cs#L236)):*
> ```Smalltalk
> 			if (request.CurrentBook?.ActiveLanguages != null)
> 			{
> 				var message = new LicenseChecker().CheckBook(request.CurrentBook, request.CurrentBook.ActiveLanguages.ToArray());
> ```

I might be wrong, but I think this is the wrong list of languages to use.
Isn't this the languages active in the book in edit mode? i.e. won't there only be more than one language here if the book is bi- or trilingual?

BloomPUBs can and do contain all the languages in a book unless turned off.

I think you used the same param for epub in the 5.2 change. It might be right there... I'm not sure 100% sure about that either. I guess it needs to be triggered on the same thing which triggers the message in both cases.



_Code quote:_
```Smalltalk
request.CurrentBook.ActiveLanguages
```

StephenMcConnel (commented):
I think the audio/video preview doesn't load only if none of the languages in the book are valid for the license.  If any of them are valid, I don't know how you would restrict creating a video from the book with a restricted license rather than an allowed license if the preview allows choosing the language.  I suspect this requires more thinking and design, and possibly another card specific to audio/video publishing.  In particular the language choice checkboxes from publish to BloomPub need to also be present in the publish to Audio or Video tool.
I suppose feeding the current language choice out of bloom-player back to C# might work as well, but I have not idea how to pull that off.

> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5241#-:-N51HI701mwY_N1AHLAk:b-pghiau)* status: all files reviewed, 1 unresolved discussion (waiting on @andrew-polk)

___
*[`src/BloomExe/Publish/Video/PublishAudioVideoAPI.cs` line 236 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5241#-N4mhRze7awHhbzA3M6g:-N51Dyr92mfVa_-kj2Ee:b-22pzc0) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/7a50088f77403cb7b07e6cde7164ef74e8c6fe27/src/BloomExe/Publish/Video/PublishAudioVideoAPI.cs#L236)):*

I might be wrong, but I think this is the wrong list of languages to use.
Isn't this the languages active in the book in edit mode? i.e. won't there only be more than one language here if the book is bi- or trilingual?

BloomPUBs can and do contain all the languages in a book unless turned off.

I think you used the same param for epub in the 5.2 change. It might be right there... I'm not sure 100% sure about that either. I guess it needs to be triggered on the same thing which triggers the message in both cases.
</blockquote></details>

I changed this to use the result from checking _publishToAndroidApi.MakeBloompubPreview().
In ePUBs, I think only the current ActiveLanguages  are published (if bilingual books are even allowed), so that should be okay I think.

andrew-polk (changes requested):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5241)* status: all files reviewed, 2 unresolved discussions (waiting on @StephenMcConnel)

___
*[`src/BloomExe/Publish/Android/PublishToAndroidApi.cs` line 612 at r2](https://reviewable.io/reviews/bloombooks/bloomdesktop/5241#-N51j2LY-OnZeOut-Zyg:-N51j2LY-OnZeOut-Zyh:b-nq1235) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/c69766277985d666898cd7252b6557e35d9344f9/src/BloomExe/Publish/Android/PublishToAndroidApi.cs#L612)):*
> ```Smalltalk
> 					LicenseOK = false;
> 					_webSocketServer.SendString(kWebSocketContext, kWebsocketState_LicenseOK, "false");
> 					if (!forVideo)
> ```

I don't think I understand the implications of this. Does not returning mean we will produce the preview anyway?
___
*[`src/BloomExe/Publish/Video/PublishAudioVideoAPI.cs` line 233 at r2](https://reviewable.io/reviews/bloombooks/bloomdesktop/5241#-N51i_hADkmu-mOGzISu:-N51i_hADkmu-mOGzISv:bjbtn7f) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/c69766277985d666898cd7252b6557e35d9344f9/src/BloomExe/Publish/Video/PublishAudioVideoAPI.cs#L233)):*
> ```Smalltalk
> 		{
> 			_publishToAndroidApi.MakeBloompubPreview(request, true);
> 			_webSocketServer.SendString("recordVideo", "publish/licenseOK", _publishToAndroidApi.LicenseOK ? "true" : "false");
> ```

Is this line needed if we are sending the same message on the same socket in `MakeBloomPubForPreview`?

StephenMcConnel (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5241)* status: all files reviewed, 2 unresolved discussions (waiting on @andrew-polk)

___
*[`src/BloomExe/Publish/Android/PublishToAndroidApi.cs` line 612 at r2](https://reviewable.io/reviews/bloombooks/bloomdesktop/5241#-N51j2LY-OnZeOut-Zyg:-N51lopJ4lVo8jigC218:b-wo1la1) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/c69766277985d666898cd7252b6557e35d9344f9/src/BloomExe/Publish/Android/PublishToAndroidApi.cs#L612)):*

I don't think I understand the implications of this. Does not returning mean we will produce the preview anyway?
</blockquote></details>

Yes, not returning means that the preview is produced even with the warning in the logbox.
___
*[`src/BloomExe/Publish/Video/PublishAudioVideoAPI.cs` line 233 at r2](https://reviewable.io/reviews/bloombooks/bloomdesktop/5241#-N51i_hADkmu-mOGzISu:-N51lyHWBEXy7LHwjkFV:bx6nvgg) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/c69766277985d666898cd7252b6557e35d9344f9/src/BloomExe/Publish/Video/PublishAudioVideoAPI.cs#L233)):*

Is this line needed if we are sending the same message on the same socket in `MakeBloomPubForPreview`?
</blockquote></details>

The message is not sent with the same context in the other file.  I suppose we have the information there so that we could use the "recordVideo" context.

andrew-polk (changes requested):
I agree it seems we need further design to determine how that situation should be handled.

> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5241#-:-N51oATY-8W3arSDlNd1:b-o2b2rn)* status: all files reviewed, 1 unresolved discussion (waiting on @StephenMcConnel)

___
*[`src/BloomExe/Publish/Android/PublishToAndroidApi.cs` line 612 at r2](https://reviewable.io/reviews/bloombooks/bloomdesktop/5241#-N51j2LY-OnZeOut-Zyg:-N51nstGB_mKTzOxwXQ5:bvs61bw) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/c69766277985d666898cd7252b6557e35d9344f9/src/BloomExe/Publish/Android/PublishToAndroidApi.cs#L612)):*

Yes, not returning means that the preview is produced even with the warning in the logbox.
</blockquote></details>

As discussed, let's not create the preview since that seems to be what the BloomPub tab is doing currently if it doesn't have a valid license.
___
*[`src/BloomExe/Publish/Video/PublishAudioVideoAPI.cs` line 233 at r2](https://reviewable.io/reviews/bloombooks/bloomdesktop/5241#-N51i_hADkmu-mOGzISu:-N51o16Z5FN92eNDAXZD:b-sfoyt1) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/c69766277985d666898cd7252b6557e35d9344f9/src/BloomExe/Publish/Video/PublishAudioVideoAPI.cs#L233)):*

The message is not sent with the same context in the other file.  I suppose we have the information there so that we could use the "recordVideo" context.
</blockquote></details>

Ok, sorry; for some reason, I had convinced myself they were the same previously.

StephenMcConnel (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5241)* status: all files reviewed, 1 unresolved discussion (waiting on @andrew-polk)

___
*[`src/BloomExe/Publish/Android/PublishToAndroidApi.cs` line 612 at r2](https://reviewable.io/reviews/bloombooks/bloomdesktop/5241#-N51j2LY-OnZeOut-Zyg:-N51qVEH1rrhCvu88pKb:b-896fix) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/c69766277985d666898cd7252b6557e35d9344f9/src/BloomExe/Publish/Android/PublishToAndroidApi.cs#L612)):*

As discussed, let's not create the preview since that seems to be what the BloomPub tab is doing currently if it doesn't have a valid license.
</blockquote></details>

Done.

andrew-polk (approved):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5241)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @StephenMcConnel)

PR #5247, opened 2022-06-18
Fix BL-11319

hatton:
This change is [Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5247)

### DIFF
--- a/src/BloomExe/web/BloomServer.cs
+++ b/src/BloomExe/web/BloomServer.cs
@@ -741,13 +741,21 @@ namespace Bloom.Api
 					}
 				}
 			}
+			// This was REMOVED to fix BL-11319. Problems with it:
+			// 1) it is now testing localPath AFTER we've already moved on to "path"
+			// 2) it is tesing the infor.RawUrl, again, after we've already move on to locaPath and then path
+			// 3) I can't reproduce the original problem of BL-3835 any more, if I remove it.
+			// 4) The unit test that came with the PR now passes without this code. (https://github.com/BloomBooks/BloomDesktop/pull/1221)
+			/*
+			 * 
 			// Use '%25' to detect that the % in a Url encoded character (for example space encoded as %20) was encoded as %25.
 			// In this example we would have %2520 in info.RawUrl and %20 in localPath instead of a space.  Note that if an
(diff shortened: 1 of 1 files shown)


hatton (commented):
The code is *so very wrong* that I don't think it makes sense to make a test. There are infinite numbers of *so very wrong* bits of code we could insert :-)

> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5247#-:-N50QEqJ2GkWPsT5yiqO:b-j4m0r8)* status: all files reviewed, 1 unresolved discussion (waiting on @hatton)

hatton (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5247)* status: all files reviewed, 1 unresolved discussion (waiting on @JohnThomson)

___
*[`src/BloomExe/web/BloomServer.cs` line 767 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5247#-N50Kc0J28F9tzjq9fEV:-N50QWdx9WJ6ikEz5x53:bhxay63) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/88e923fac448e128152e63a765bb59df5b331283/src/BloomExe/web/BloomServer.cs#L767)):*

I'm nervous about simply removing this logic, even though tests still pass. You're probably right, but I'd like to actually understand how the test now passes without it, and also why this book name is a problem for video recording but not for creating BloomPUBs.
How about using something like UrlPathString.GetFullyDecodedPath(), which is smart enough to 'improve' the path by UrlDecoding it only if no file is found at the path given and one IS found at the decoded path? But then, before doing that, it would be good to have a test case that fails without it, and I don't know what that might be.
(Feel free to push back. I'm not at all sure we need to replace the removed code with anything.)
Also: I prefer to comment out blocks of code with a leading // on each line. It's easy to miss that a block is wrapped in /* */.

</blockquote></details>

Yeah, I think I should never have approved this code in the first place. It would *never* make sense to check on `localPath` after we've processed it to `path`.

JohnThomson (approved):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5247)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @hatton)

PR #5253, opened 2022-06-21
Fix BloomPUB preview when book file name has parentheses (BL-11352)

andrew-polk:
This change is [Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5253)

### DIFF
--- a/src/BloomExe/Publish/Android/PublishToAndroidApi.cs
+++ b/src/BloomExe/Publish/Android/PublishToAndroidApi.cs
@@ -594,9 +594,9 @@ namespace Bloom.Publish.Android
 		internal bool LicenseOK;
 
 		/// <summary>
-		/// Generates a .bloompub file from the book
+		/// Generates an unzipped, staged BloomPUB from the book
 		/// </summary>
-		/// <returns>A valid, well-formed URL on localhost that points to the bloompub</returns>
+		/// <returns>A valid, well-formed URL on localhost that points to the staged book's htm file</returns>
 		public string MakeBloomPubForPreview(Book.Book book, BookServer bookServer, WebSocketProgress progress, Color backColor, AndroidPublishSettings settings = null)
 		{
 			progress.Message("PublishTab.Epub.PreparingPreview", "Preparing Preview");	// message shared with Epub publishing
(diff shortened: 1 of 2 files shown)


hatton (changes requested):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5253)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @andrew-polk and @hatton)

___
*[`src/BloomExe/web/BloomServer.cs` line 453 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5253#-N56iWGjEGnIqCxYDedN:-N56iWGjEGnIqCxYDedO:b2ta242) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/be025305e54093657663cec8dfe509bb2270c561/src/BloomExe/web/BloomServer.cs#L453)):*
> ```Smalltalk
> 			{
> 				info.ResponseContentType = "text/html";
> 				info.WriteCompleteOutput($"<meta http-equiv=\"Refresh\" content=\"0; url='{PublishToAndroidApi.PreviewUrl}'\" />");
> ```

I don't understand this change.

hatton (approved):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5253)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @andrew-polk and @hatton)

___
*[`src/BloomExe/web/BloomServer.cs` line 453 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5253#-N56iWGjEGnIqCxYDedN:-N56p_SWE80SrpK08_YR:bfvamr6) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/be025305e54093657663cec8dfe509bb2270c561/src/BloomExe/web/BloomServer.cs#L453)):*

I'm actual fixing an existing problem, but I found it because I went looking for usages of PublishToAndroidApi.PreviewUrl.

Before this change, loading http://localhost:8089/bloom/current-bloompub-url would just give a page displaying the text of the book folder path in the staging directory. With the change in PublishToAndroidApi, all that changed is it would show the file path instead of the folder path.
But since the intent is to actually load the staged book in the browser, it made it do that, albeit through an awkward redirect.
</blockquote></details>

Sorry, I still don't understand how this  `<meta` works. But I don't need to since you seem confident.

PR #5260, opened 2022-06-23
Save "Export to Word" content as separate folder (BL-11335)

StephenMcConnel:
This change is [Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5260)

### DIFF
--- a/src/BloomExe/CollectionTab/CollectionModel.cs
+++ b/src/BloomExe/CollectionTab/CollectionModel.cs
@@ -10,6 +10,7 @@ using System.Xml;
 using Bloom.Api;
 using Bloom.Book;
 using Bloom.Collection;
+using Bloom.MiscUI;
 using Bloom.Properties;
 using Bloom.TeamCollection;
 using Bloom.ToPalaso;
@@ -22,8 +23,8 @@ using SIL.IO;
 using SIL.Progress;
 using SIL.Reporting;
 using SIL.Windows.Forms;
-using SIL.Xml;
 using SIL.Windows.Forms.FileSystem;
+using SIL.Xml;
 
 namespace Bloom.CollectionTab
 {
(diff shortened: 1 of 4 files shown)


JohnThomson (approved):
This is good as far as it goes, but I'm concerned that the behavior is to copy all the files INTO the folder the user selects.
- It will overwrite, without warning, any existing files in the folder with the same names
- The book's private files will be mingled with any existing files and could be difficult to distinguish if this isn't what the user intended
- If the user wants a new folder to keep the book and its files together, he must create and name it manually.
- If the user keeps on using the same folder for repeated export attempts, obsolete files will not be deleted.
- It may be confusing that the behavior is different from that of the similar spreadsheet export code, which creates a NEW folder (using the book name) inside the folder that the user selects. If the folder resulting from that algorithm already exists, Bloom warns the user and deletes it before making the new folder (so no obsolete files).
- But maybe this is what @hatton intended to just minimally repair this feature for 5.3?
- LGTM if JohnH is happy with this behavior. (Note however that the card does say "like we do for spreadsheets," which we haven't quite achieved yet.)

> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5260#-:-N5aStpBCySZ7tYsKzRI:bkp6vww)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @StephenMcConnel)

sujeffreyl (changes requested):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5260)* status: all files reviewed, 1 unresolved discussion (waiting on @StephenMcConnel)

___
*[`src/BloomExe/CollectionTab/CollectionModel.cs` line 517 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5260#-N5aXoQd5sEMoSR0aUTZ:-N5aXoQd5sEMoSR0aUT_:b-2sjogs) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/fcddeb2955430fb39ab17899b0c50836282f393c/src/BloomExe/CollectionTab/CollectionModel.cs#L517)):*
> ```Smalltalk
> 					case ".jpg":
> 					case ".jpeg":
> 					case ".svg":
> ```

In BookCompressor.cs, we have a variable ImageFileExtensions which also includes the following image extensions as valid:
".tif", ".tiff", ".bmp",

Is there any reason why we wouldn't want to include those 3 when exporting to word?

sujeffreyl (changes requested):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5260)* status: all files reviewed, 1 unresolved discussion (waiting on @StephenMcConnel)

___
*[`src/BloomExe/CollectionTab/CollectionModel.cs` line 517 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5260#-N5aXoQd5sEMoSR0aUTZ:-N5aY9Cc9Lp9bU0u5Uec:bx8uwar) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/fcddeb2955430fb39ab17899b0c50836282f393c/src/BloomExe/CollectionTab/CollectionModel.cs#L517)):*

In BookCompressor.cs, we have a variable ImageFileExtensions which also includes the following image extensions as valid:
".tif", ".tiff", ".bmp",

Is there any reason why we wouldn't want to include those 3 when exporting to word?
</blockquote></details>

(And if not, maybe we can directly reference ImageFileExtensions)

sujeffreyl (changes requested):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5260)* status: 2 of 4 files reviewed, 2 unresolved discussions (waiting on @JohnThomson and @StephenMcConnel)

___
*[`src/BloomExe/CollectionTab/CollectionModel.cs` line 517 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5260#-N5aXoQd5sEMoSR0aUTZ:-N5eZE6Q3Q0jU3jvPdPu:b3e7ox9) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/fcddeb2955430fb39ab17899b0c50836282f393c/src/BloomExe/CollectionTab/CollectionModel.cs#L517)):*

Although Bloom allows TIFF and BMP files to be selected, I think they're always converted to PNG files before being stored in the book's folder.
</blockquote></details>

Ok, that does appear to be true, so I'm ok with leaving this as is, if desired. However, I'll still comment that I'd prefer to just reference the same list instead of having 2 lists here, so if you like that too you can change, but if not that's fine too.
___
*[`src/BloomExe/CollectionTab/CollectionModel.cs` line 490 at r2](https://reviewable.io/reviews/bloombooks/bloomdesktop/5260#-N5eYIUr99aaExuvg_Vd:-N5eYIUr99aaExuvg_Ve:bh3cv3i) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/427314bed3d2cec708474d624924a498ec47963d/src/BloomExe/CollectionTab/CollectionModel.cs#L490)):*
> ```Smalltalk
> 		{
> 			var outputFolder = Path.GetDirectoryName(destDocPath);
> 			if (Directory.Exists(outputFolder))
> ```

Shouldn't we first confirm that they wish to overwrite, before blowing it away? (Export to Spreadsheet does that)
___
*[`src/BloomExe/web/controllers/BookCommandsApi.cs` line 297 at r2](https://reviewable.io/reviews/bloombooks/bloomdesktop/5260#-N5eZnBT7-nkP--qbTh9:-N5eZnBT7-nkP--qbThA:bf3nlc6) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/427314bed3d2cec708474d624924a498ec47963d/src/BloomExe/web/controllers/BookCommandsApi.cs#L297)):*
> ```Smalltalk
> 		}
> 
> 		private void HandleExportToWord(Book.Book book, OverwriteOptions overwrite = OverwriteOptions.Ask)
> ```

This parameter doesn't seem to be referenced, as far as I can tell.



_Code quote:_
```Smalltalk
overwrite
```

JohnThomson (approved):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5260)* status: all files reviewed, 2 unresolved discussions (waiting on @sujeffreyl)

PR #5263, opened 2022-06-27
Initial webview2 in Collection & Publish

hatton:
This change is [Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5263)


hatton (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5263)* status: all files reviewed, 14 unresolved discussions (waiting on @hatton, @JohnThomson, and @StephenMcConnel)

___
*[`src/BloomExe/GeckoFxBrowser.cs` line 1481 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5263#-N5aR75K0Z0FJwHn01mD:-N5aygsADw9W3AIItRsY:b-9g77pe) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/136a30c79b961f97af6a66ee873a34ca1640769a/src/BloomExe/GeckoFxBrowser.cs#L1481)):*

A few Linux-specific items were removed, but otherwise looks good as an adjustment to the previous Browser.cs to be the explicit Geckofx implementation.
</blockquote></details>

Apologies, Steve. When I rebased or whatever this code from Dec, I managed to overwrite what had been done in the meantime, and couldn't see it because of file name changes. All fixed up now and broken into 2 commits to make changes to (Browser-->GeckoFxBrowser).
___
*[`src/BloomExe/IBrowser.cs` line 7 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5263#-N5ae8j96DYZL2deDYrh:-N5azJba91h-LBG5Z-q_:b-896fix) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/136a30c79b961f97af6a66ee873a34ca1640769a/src/BloomExe/IBrowser.cs#L7)):*

through-away -> throw-away
But do we want to throw it away? Having done all the work to tease apart what might need doing differently for different web controls?
</blockquote></details>

Done.
___
*[`src/BloomExe/IBrowser.cs` line 7 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5263#-N5aRpaS0Bu8mEj9B3RO:-N5azK0L9cOluyHYSVYS:b-896fix) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/136a30c79b961f97af6a66ee873a34ca1640769a/src/BloomExe/IBrowser.cs#L7)):*

"throw-away"  instead of "through-away"?
</blockquote></details>

Done.
___
*[`src/BloomExe/IBrowser.cs` line 18 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5263#-N5af7sgDlwnWS4aTdjL:-N5azS6oB-MEAYn5EzRH:bcg6v11) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/136a30c79b961f97af6a66ee873a34ca1640769a/src/BloomExe/IBrowser.cs#L18)):*

Maybe this should be an abstract class? Seems like several of these methods...e.g., RaiseBowserReady, OnOpenPageInSystemBrowser...might not require different implementations. As an abstract class it could inherit from UserControl and keep the current class name. Both would reduce the number of changes needed elsewhere.
But if it's really throw-away I don't suppose it matters much.
If we end up keeping this interface/abstract class we should give thought to method naming and commenting.

</blockquote></details>

yeah, we could go that direction.
___
*[`src/BloomExe/IBrowser.cs` line 25 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5263#-N5apT6U1Bw-3iIdyu-Q:-N5azdLp-61y1oR9OCpl:b-8kik9o) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/136a30c79b961f97af6a66ee873a34ca1640769a/src/BloomExe/IBrowser.cs#L25)):*
> Since the new implementation does NOT ensure the handle is created,

I don't know why you say that. It's the same code as before, just now in the GfxBrowser class itself instead of the client. It was lame before, and lame now. What I wanted was for the client to not have to think about it, but I didn't come up with a way to do that. GfxB is what we're trying to retire, so not worth much more thought.
___
*[`src/BloomExe/packages.config` line 47 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5263#-N5aSH8H-mlAL1xV-sEL:-N5b-6jd5tjE5Epi1VYp:b-sw2756) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/136a30c79b961f97af6a66ee873a34ca1640769a/src/BloomExe/packages.config#L47)):*

This line should be added to Linux/packages.config even though the WebView2 won't work in that environment.  It will build on Linux with this package included, and that's the simplest fix to keep Bloom compiling on Linux with this initial change set.
</blockquote></details>

I figure you're going to have to do who knows what when this lands.
___
*[`src/BloomExe/WebView2Browser.cs` line 32 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5263#-N5agtvz2Rg4fMAlIJFn:-N5b-Hei6dkHLq8ID-Gn:b-896fix) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/136a30c79b961f97af6a66ee873a34ca1640769a/src/BloomExe/WebView2Browser.cs#L32)):*

Deserves a comment either to explain or to warn that it's meant to be temporary.
</blockquote></details>

Done.
___
*[`src/BloomExe/WebView2Browser.cs` line 96 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5263#-N5ahdZd6-s2Db_bGkXj:-N5b-KItEOKtOq19MNlx:b-so6mk8) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/136a30c79b961f97af6a66ee873a34ca1640769a/src/BloomExe/WebView2Browser.cs#L96)):*

Wants at least a todo indicating that params like timeLimit and throwOnTimeout are not implemented.
</blockquote></details>

I'd go so far as to suggest that *most* things are implemented at this time. 
___
*[`src/BloomExe/XmlHtmlConverter.cs` line 61 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5263#-N5af2sj2ZahRuvAuQyq:-N5b-XeG2h8v25pPusWM:b6bu8hd) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/136a30c79b961f97af6a66ee873a34ca1640769a/src/BloomExe/XmlHtmlConverter.cs#L61)):*

Could these constant strings be moved to somewhere more appropriate?  I'm not sure what to suggest, however, so it may be okay to leave them in place for the moment.
</blockquote></details>

The only change from this PR is `Browser` --> `GeckoFxBrowser`

hatton (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5263)* status: all files reviewed, 9 unresolved discussions (waiting on @JohnThomson and @StephenMcConnel)

___
*[`src/BloomExe/IBrowser.cs` line 25 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5263#-N5apT6U1Bw-3iIdyu-Q:-N6i4O3o98BM0WFUqlAZ:b-vt6np) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/136a30c79b961f97af6a66ee873a34ca1640769a/src/BloomExe/IBrowser.cs#L25)):*

But it's now in the interface, so a reasonable expectation would be that BOTH implementations will ensure that the handle is created. If that's really the intent, then WebView2Browser should also do the trick to force the handle to be created. If only one implementation needs that to be done, the name should be indicative of the real purpose of calling it...e.g., PrepareToNavigate.
But I suppose it will go away when the new control takes over completely if it doesn't do anything, so maybe it doesn't matter. Up to you.
</blockquote></details>

Yeah I think it's going to go away.
___
*[`src/BloomExe/Publish/HtmlPublishPanel.cs` line 26 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5263#-N5ajO6J75bWTcl0XH2t:-N6i4dde6dBF_oRLzbJ1:b2nih0v) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/136a30c79b961f97af6a66ee873a34ca1640769a/src/BloomExe/Publish/HtmlPublishPanel.cs#L26)):*

Something obviously will need to change to extend our WebView2 class to be used here.
</blockquote></details>

yup
___
*[`src/BloomExe/Publish/HtmlPublishPanel.Designer.cs` line 28 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5263#-N5ajaLZ20EZivAldbP5:-N6i4kM23TvTbWwn3aLo:b-xrlpks) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/136a30c79b961f97af6a66ee873a34ca1640769a/src/BloomExe/Publish/HtmlPublishPanel.Designer.cs#L28)):*

Or could you just put Dispose() into IBrowser? Or doesn't WebView2 need to be disposed?
</blockquote></details>

It seems good to just let the designer code work as normal in disposing of its children. The only problem is that I don't know how to make c# let IBrowser extend UserControl.
___
*[`src/BloomExe/web/ReactControl.cs` line 73 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5263#-N5ajzig9eAIl517osjp:-N6oeZPS3saJAi02Hzbr:b-896fix) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/136a30c79b961f97af6a66ee873a34ca1640769a/src/BloomExe/web/ReactControl.cs#L73)):*

Another advantage of making IBrowser an abstract class: it could inherit from UserControl.
</blockquote></details>

Done.
___
*[`src/BloomExe/web/ReactControl.cs` line 87 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5263#-N5alnUf6HPed7Z8jsGG:-N6ohYwf2Lq42lJh3MKR:b-5p2vfz) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/136a30c79b961f97af6a66ee873a34ca1640769a/src/BloomExe/web/ReactControl.cs#L87)):*

Some sort of ToDo, maybe with a keyword like // Todo: finishWebView2, to remind us to reinstate this at some point?
</blockquote></details>

Good catch. I went ahead and just implemented what was needed to keep this code working with gfx
___
*[`src/BloomExe/web/ReactControl.Designer.cs` line 54 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5263#-N5amNxmET9-qkCb5SIF:-N6oiM1E9YQmfKBXT0MV:b-896fix) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/136a30c79b961f97af6a66ee873a34ca1640769a/src/BloomExe/web/ReactControl.Designer.cs#L54)):*

Something here seems to have converted tabs to spaces. Is that just something the latest VS does in designer mode and we can't help it, or was it intentional, or was it a mistake?
</blockquote></details>

Done.

JohnThomson (approved):
:lgtm:

> Reviewed 7 of 7 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5263#-:-N6rmpMR5obTCbjbUUsU:bnfp4nl)* status: all files reviewed, 4 unresolved discussions (waiting on @StephenMcConnel)

PR #5287, opened 2022-07-08
Make a clearer model for IBloomErrorReporter (BL-11193)

sujeffreyl:


sujeffreyl (commented):
Instead of having Bloom have a separate ErrorReportUtils class to add functionality to libpalaso's ErrorReport class, we can now make a subclass of ErrorReport. (The subclass is called BloomErrorReport). This allows us to avoid setting function parameters in class instance vars in a mult-stage, stateful scenario. Instead, we can now instead pass the function parameters directly through, so it's more direct and less stateful now (and this should relieve JT's concerns about any potential situation where the instance vars might get overwritten by multiple error reports)

> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5287#-:-N6TlWUWD81fFBYU0kfU:bx63ikb)* status: 0 of 16 files reviewed, all discussions resolved

sujeffreyl (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5287)* status: 0 of 16 files reviewed, 1 unresolved discussion

___
*[`src/BloomExe/BloomExe.csproj` line 3 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5287#-N6Tn70HAMfFCNM3_TXI:-N6Tn70HAMfFCNM3_TXJ:bvnqax1) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/fc832df434d36f7c84451aaccbc02c65f77b0c99/src/BloomExe/BloomExe.csproj#L3)):*
> ```XML
> <Project ToolsVersion="12.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
>   <Import Project="..\..\packages\SIL.Windows.Forms.Keyboarding.10.0.0-beta0042\build\SIL.Windows.Forms.Keyboarding.props" Condition="Exists('..\..\packages\SIL.Windows.Forms.Keyboarding.10.0.0-beta0042\build\SIL.Windows.Forms.Keyboarding.props')" />
> ```

Waiting upon https://github.com/sillsdev/libpalaso/pull/1191 to be merged.
Then, will need to go back and edit this to beta0043 instead.

sujeffreyl (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5287)* status: 0 of 16 files reviewed, 2 unresolved discussions

___
*[`src/BloomExe/ErrorReporter/AllowSendReport.cs` line 5 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5287#-N6TnU8ZAh1wQvfrj80m:-N6TnU8ZAh1wQvfrj80n:b-k90io5) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/fc832df434d36f7c84451aaccbc02c65f77b0c99/src/BloomExe/ErrorReporter/AllowSendReport.cs#L5)):*
> ```Smalltalk
> namespace Bloom.ErrorReporter
> {
> 	public enum AllowSendReport
> ```

So, JH does not want to receive reports that just say what the notify said.  So, we don't want the Report button to show up all the time by default...
I explored making it not show up by default unless specifically requested, but this is contrary to the legacy behavior which determined it based on whether the exception was null or non-null. There could be code in libpalaso that makes this assumption, and... so far I've found about 20% of the calls to NotifyUserOfProblem that I've checked rely on this behavior.  (And there are 138 calls total in BloomExe). So I'm reluctant to change the default behavior.

So, what I'm thinking is that the default will be AllowIfException, same as the legacy behavior, but I provide a way to manually override the default behavior if you don't want the report button to show (or do, even if there's no exception).

JohnThomson (changes requested):
I like it! A few minor suggestions.

> Reviewed 15 of 16 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5287#-:-N6Uobat3X-QYwejauQp:bm1pdew)* status: all files reviewed, 9 unresolved discussions (waiting on @andrew-polk and @sujeffreyl)

___
*[`src/BloomExe/ErrorReporter/AllowSendReport.cs` line 10 at r2](https://reviewable.io/reviews/bloombooks/bloomdesktop/5287#-N6UegnZ37GWddqFdkp3:-N6UegnZ37GWddqFdkp4:bp3fdie) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/e041ac7770338d72a99fae15532184e471996b12/src/BloomExe/ErrorReporter/AllowSendReport.cs#L10)):*
> ```Smalltalk
> 		Disallow,
> 		Allow,
> 		AllowIfException
> ```

If the default is AllowIfException, why are they two distinct members of the enumeration?
___
*[`src/BloomExe/ErrorReporter/BloomErrorReport.cs` line 16 at r2](https://reviewable.io/reviews/bloombooks/bloomdesktop/5287#-N6Uffd99gB4Lnv6N7cC:-N6Uffd99gB4Lnv6N7cD:bowjmup) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/e041ac7770338d72a99fae15532184e471996b12/src/BloomExe/ErrorReporter/BloomErrorReport.cs#L16)):*
> ```Smalltalk
> 		/// <param name="policy">Optional: The policy to use to decide whether to show the notification. Defaults to ShowAlwaysPolicy</param>
> 		/// <param name="allowSendReport">Whether or not to show the Report button.</param>
> 		/// <param name="extraButtonLabel">Optional: The label of the Report button. Pass "" to disable. Defaults to disabled.</param>
> ```

This is confusing. Since the default is null, presumably that also disables, though the comment doesn't say so. But then there is the possibility of passing AllowSendReport.Allow along with an empty string or null, which feels contradictory. Maybe the default should just use a default string, and the allowSendReport should control that? Or is that a problem because callers assume an old behavior?
___
*[`src/BloomExe/ErrorReporter/BloomErrorReport.cs` line 34 at r2](https://reviewable.io/reviews/bloombooks/bloomdesktop/5287#-N6UlV1h3rIvV4uin1vb:-N6UlV1h3rIvV4uin1vc:b9lcig4) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/e041ac7770338d72a99fae15532184e471996b12/src/BloomExe/ErrorReporter/BloomErrorReport.cs#L34)):*
> ```Smalltalk
> 						allowSendReport, extraButtonLabel, onExtraButtonPressed);
> 				}
> 				else
> ```

I presume there's no way GetErrorReporter can return null?
___
*[`src/BloomExe/ErrorReporter/BloomErrorReport.cs` line 47 at r2](https://reviewable.io/reviews/bloombooks/bloomdesktop/5287#-N6UkUkH-rvyFkiO06UE:-N6UkUkH-rvyFkiO06UF:bxky62j) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/e041ac7770338d72a99fae15532184e471996b12/src/BloomExe/ErrorReporter/BloomErrorReport.cs#L47)):*
> ```Smalltalk
> 					if (result == ErrorResult.Yes)
> 						HtmlErrorReporter.DefaultOnReportClicked(message, exception);
> 				}
> ```

My first thought looking at the above was that there should be some way to follow the reporting policy while using the base type. But IIRC, that's why we added the improved API.  Might be worth commenting explicitly that the legacy API doesn't support control over the appearance or presence of the Report button.
___
*[`src/BloomExe/ErrorReporter/HtmlErrorReporter.cs` line 128 at r2](https://reviewable.io/reviews/bloombooks/bloomdesktop/5287#-N6UmKHe5HfawPmWVn6b:-N6UmKHe5HfawPmWVn6c:b-otilcs) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/e041ac7770338d72a99fae15532184e471996b12/src/BloomExe/ErrorReporter/HtmlErrorReporter.cs#L128)):*
> ```Smalltalk
> 		/// <param name="allowSendReport">Whether or not to show the Report button.</param>
> 		/// <param name="extraButtonLabel">The label of the extra button. "" means disabled. null means default, which is disabled.</param>
> 		/// <param name="onExtraButtonClicked">The action to take when the extra button is clicked.</param>
> ```

See comment above.
___
*[`src/BloomExe/ErrorReporter/IBloomErrorReporter.cs` line 18 at r2](https://reviewable.io/reviews/bloombooks/bloomdesktop/5287#-N6UmjN97oVcCikyBkHG:-N6UmjN97oVcCikyBkHH:bio604c) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/e041ac7770338d72a99fae15532184e471996b12/src/BloomExe/ErrorReporter/IBloomErrorReporter.cs#L18)):*
> ```Smalltalk
> 		/// <param name="policy">The policy to use to decide whether to show the notification.</param>
> 		/// <param name="allowSendReport">Whether or not to show the Report button.</param>
> 		/// <param name="extraButtonLabel">The label of the extra button. "" means disabled. null means use the default label.</param>
> ```

And again.
___
*[`src/BloomExe/ErrorReporter/SentryErrorReporter.cs` line 35 at r2](https://reviewable.io/reviews/bloombooks/bloomdesktop/5287#-N6Umx3R6e3GMRKZDJFN:-N6Umx3R6e3GMRKZDJFO:b-9hl6w9) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/e041ac7770338d72a99fae15532184e471996b12/src/BloomExe/ErrorReporter/SentryErrorReporter.cs#L35)):*
> ```Smalltalk
> 
> 		public void NotifyUserOfProblem(IRepeatNoticePolicy policy, Exception exception, string message)
> 			=> NotifyUserOfProblemCustom(message, exception, policy);
> ```

Less and less happy about this name as I see more. What's Custom about it? It's like an overload with all the options, but you can't make it an overload because so many are optional that it would be ambiguous? Could we make fewer arguments optional, maybe at the cost of another overload or two, so that one can use the same name? Or how about NotifyUserOfProblemCore or NotifyUserOfProblemImpl[ementation]?

sujeffreyl (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5287)* status: all files reviewed, 9 unresolved discussions (waiting on @andrew-polk, @JohnThomson, and @sujeffreyl)

___
*[`src/BloomExe/ErrorReporter/SentryErrorReporter.cs` line 35 at r2](https://reviewable.io/reviews/bloombooks/bloomdesktop/5287#-N6Umx3R6e3GMRKZDJFN:-N6hcPe19vdeFcMtLQUe:bb7w9eg) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/e041ac7770338d72a99fae15532184e471996b12/src/BloomExe/ErrorReporter/SentryErrorReporter.cs#L35)):*

Less and less happy about this name as I see more. What's Custom about it? It's like an overload with all the options, but you can't make it an overload because so many are optional that it would be ambiguous? Could we make fewer arguments optional, maybe at the cost of another overload or two, so that one can use the same name? Or how about NotifyUserOfProblemCore or NotifyUserOfProblemImpl[ementation]?
</blockquote></details>

Hmm. I guess there are 2 aspects to why it's "custom". 1) It allows customizing the two buttons.  2) it's custom in the sense that we customized  IErrorReporter with IBloomErrorReporter.

I wouldn't want to call it Impl[emenation], because it's 100% valid and reasonable for other parts of the code to call NotifyUserOfProblemCustom directly.

However, I would be ok with just renaming it to NotifyUserOfProblem if you prefer that to NotifyUserOfProblemCustom. I think I can make it an overload without any problem? I'll just add comments around each one as to whether it's required by IErrorReporter or IBloomErrorReporter.

sujeffreyl (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5287)* status: all files reviewed, 9 unresolved discussions (waiting on @andrew-polk, @JohnThomson, and @sujeffreyl)

___
*[`src/BloomExe/ErrorReporter/BloomErrorReport.cs` line 16 at r2](https://reviewable.io/reviews/bloombooks/bloomdesktop/5287#-N6Uffd99gB4Lnv6N7cC:-N6htoNGCrNrRlq2IcVL:b-8h69tl) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/e041ac7770338d72a99fae15532184e471996b12/src/BloomExe/ErrorReporter/BloomErrorReport.cs#L16)):*

This is confusing. Since the default is null, presumably that also disables, though the comment doesn't say so. But then there is the possibility of passing AllowSendReport.Allow along with an empty string or null, which feels contradictory. Maybe the default should just use a default string, and the allowSendReport should control that? Or is that a problem because callers assume an old behavior?
</blockquote></details>

Ooops. Well, one big part of the problem is that I copy/pasted the wrong comment...
___
*[`src/BloomExe/ErrorReporter/IBloomErrorReporter.cs` line 18 at r2](https://reviewable.io/reviews/bloombooks/bloomdesktop/5287#-N6UmjN97oVcCikyBkHG:-N6hxEfBBKNZwIZlI4vu:b-896fix) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/e041ac7770338d72a99fae15532184e471996b12/src/BloomExe/ErrorReporter/IBloomErrorReporter.cs#L18)):*

And again.
</blockquote></details>

Done.

sujeffreyl (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5287)* status: all files reviewed, 9 unresolved discussions (waiting on @andrew-polk, @JohnThomson, and @sujeffreyl)

___
*[`src/BloomExe/ErrorReporter/BloomErrorReport.cs` line 34 at r2](https://reviewable.io/reviews/bloombooks/bloomdesktop/5287#-N6UlV1h3rIvV4uin1vb:-N6hxpnf3UZIhsK38nE5:b2pktti) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/e041ac7770338d72a99fae15532184e471996b12/src/BloomExe/ErrorReporter/BloomErrorReport.cs#L34)):*

I presume there's no way GetErrorReporter can return null?
</blockquote></details>

No, at least not right now. Well, the libpalaso ErrorReport class also makes assumptions throughtout that its _errorReporter member is always non-null, so I think I feel okay making the assumption here too.
(I guess this is where if we could turn on strict null checking that would've helped clarify this situation...)
___
*[`src/BloomExe/ErrorReporter/BloomErrorReport.cs` line 47 at r2](https://reviewable.io/reviews/bloombooks/bloomdesktop/5287#-N6UkUkH-rvyFkiO06UE:-N6hyQU68_BGC_PeeoHI:buaz0d3) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/e041ac7770338d72a99fae15532184e471996b12/src/BloomExe/ErrorReporter/BloomErrorReport.cs#L47)):*

My first thought looking at the above was that there should be some way to follow the reporting policy while using the base type. But IIRC, that's why we added the improved API.  Might be worth commenting explicitly that the legacy API doesn't support control over the appearance or presence of the Report button.
</blockquote></details>

Done (in the sense that i added some comments).

I'm not sure I follow what your thoughts about the "reporting policy". Is that the same as the "policy" parameter? It is passed to the old API.
___
*[`src/BloomExe/ErrorReporter/HtmlErrorReporter.cs` line 128 at r2](https://reviewable.io/reviews/bloombooks/bloomdesktop/5287#-N6UmKHe5HfawPmWVn6b:-N6hxMYS5asJwRWxSQFq:b-896fix) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/e041ac7770338d72a99fae15532184e471996b12/src/BloomExe/ErrorReporter/HtmlErrorReporter.cs#L128)):*

See comment above.
</blockquote></details>

Done.

JohnThomson (changes requested):
> Reviewed 8 of 8 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5287)* status: all files reviewed, 4 unresolved discussions (waiting on @andrew-polk and @sujeffreyl)

___
*[`src/BloomExe/ErrorReporter/BloomErrorReport.cs` line 47 at r2](https://reviewable.io/reviews/bloombooks/bloomdesktop/5287#-N6UkUkH-rvyFkiO06UE:-N6iyu0J7xQRRW-v4WFI:b-8afdlr) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/e041ac7770338d72a99fae15532184e471996b12/src/BloomExe/ErrorReporter/BloomErrorReport.cs#L47)):*

Done (in the sense that i added some comments).

I'm not sure I follow what your thoughts about the "reporting policy". Is that the same as the "policy" parameter? It is passed to the old API.
</blockquote></details>

IIRC, the reporting policy has to do with whether the message is seen at all...some should e.g. only be seen the first time they happen. But the old API does not (I think) have a way to control whether the Report button appears.
___
*[`src/BloomExe/ErrorReporter/SentryErrorReporter.cs` line 35 at r2](https://reviewable.io/reviews/bloombooks/bloomdesktop/5287#-N6Umx3R6e3GMRKZDJFN:-N6j0NVs0pVbrzigwkVg:bg013s3) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/e041ac7770338d72a99fae15532184e471996b12/src/BloomExe/ErrorReporter/SentryErrorReporter.cs#L35)):*

Hmm. So, it works fine to rename it at the IErrorReporter/IBloomErrorReporter level. (Notice the "-er" suffix at the end). However, at the ErrorReport/BloomErrorReport (no "-er" suffix"), it is more problematic... having the same name would indeed clash with one of the existing overloads.
I'm not sure what's wrong with the name Custom? Generally, if you want to do something simple, just calling one of the original NotifyUserOfProblem methods is totally fine/probably better because there's less confusing stuff to think about.
Only when you want to bring up a Notify dialog which is customized away from the defaults do you want to deal with the 3 additional parameters.
Thoughts?

(Note: this is in a partially converted state, where I changed the name in IBloomErrorReporter but not the one in BloomErrorReport)
</blockquote></details>

This is always difficult. "I don't like it" is not an easy reason to justify. And you're right, when this is a public method designed to be called directly, "Implementation" doesn't work. Sorry. But I still don't much like Custom, so I'll have another go.
First, it's not fluent. Asking an object to notify the user of a problem does exactly what it says. But asking it to notify the user of a problem custom sounds like something an anthropology professor would do at a missionary orientation.  (Never point at anyone in Japan...) Unless it's so heavily used that a long name is a nuisance, I'd prefer NotifyUserOfProblemUsingCustomButtons. But then, you don't _have_ to be using custom buttons...those arguments are optional. In fact, you can (as here) pass exactly the same arguments (though, confusingly, in a different order) as to NotifyUserOfProblem.
We want to think of this from the point of view of the client. Clients should never care that this way of reporting errors is a custom enhancement Bloom made to LibPalaso. But the fact that it's useful for custom buttons is worth considering.
When I'm trying to guess how to report an error the way I want, I want to type "ER." and have some options pop up like various overloads of ErrorReport.NotifyUserOfProblem. (It's a nuisance that the system typically brings up ErrorReporter, which doesn't have the method I want, and then I have to go searching, especially since ErrorReporter sounds like the thing that _should_ have a method to report an error. A report is not expected to do things. But it's too late to fix that, I expect.) 
There are two ways intellisense works well. One is that there are different methods, whose names come up as options (NotifyUserOfProblem...NotifyUserOfException...NotifyUserUsingCustomDialog...NotifyUserIfNewProblem....). The other is that there is one method, and when I select it, I'm offered a list of overloads, and hopefully it's obvious which set of arguments corresponds to what I want to do (the current situation with NotifyUserOfProblem). But if you combine them, it gets harder. I have to choose between NotifyUserOfProblem and NotifyUserOfProblemCustom before I can see what arguments each accepts. Do I use the second if I want custom buttons, a custom window title, a custom policy for showing the Report button, a custom way of displaying the stack trace in the exception, support for hot links in the text,...? If we need a new name, we want it to be really clear when you want the new method and when you want an overload of the old one. So maybe if the only situation where it makes sense to use the new one is to customize buttons, we should call it NotifyUserOfProblemUsingCustomButtons, and maybe even make the arguments obligatory, so no one is tempted to use it otherwise. Or we could borrow from WinApi and call it NotifyUserOfProblemExtended, indicating that it should be used when you want unusual, possibly advanced options...leaving open the possibility of adding more arguments and overloads of this name eventually, that are special in different ways.
Well, that's my best attempt to clarify. If you're still not convinced, maybe we should see if someone else can come up with a better idea than either of us. When all's said and done, it's only a name; I can live with NotifyUserOfProblemCustom if I can't convince anyone that something else is better.

sujeffreyl (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5287)* status: all files reviewed, 4 unresolved discussions (waiting on @andrew-polk and @JohnThomson)

___
*[`src/BloomExe/ErrorReporter/BloomErrorReport.cs` line 47 at r2](https://reviewable.io/reviews/bloombooks/bloomdesktop/5287#-N6UkUkH-rvyFkiO06UE:-N6s7TXeDrvM2RSDquyj:buf3w8) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/e041ac7770338d72a99fae15532184e471996b12/src/BloomExe/ErrorReporter/BloomErrorReport.cs#L47)):*

IIRC, the reporting policy has to do with whether the message is seen at all...some should e.g. only be seen the first time they happen. But the old API does not (I think) have a way to control whether the Report button appears.
</blockquote></details>

Yes, the reporting policy is as you say. We already pass it in as the first parameter, so we are already "follow[ing] the reporting policy while using the base type".

For whether the old API can control whether the Report button appears... depends on what you mean. The ErrorReport class/API itself does not have a generalized way, so you'd be correct in that sense. Although, if you're using WinFormsErrorReporter, passing null or empty string for {alternateButton1Label} will disable it, but technically that would be an implementation detail of one of the classes implementing the interface... not true in general...
___
*[`src/BloomExe/ErrorReporter/SentryErrorReporter.cs` line 35 at r2](https://reviewable.io/reviews/bloombooks/bloomdesktop/5287#-N6Umx3R6e3GMRKZDJFN:-N6s98R5A_rQITgROpoj:b-7ioxiq) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/e041ac7770338d72a99fae15532184e471996b12/src/BloomExe/ErrorReporter/SentryErrorReporter.cs#L35)):*

This is always difficult. "I don't like it" is not an easy reason to justify. And you're right, when this is a public method designed to be called directly, "Implementation" doesn't work. Sorry. But I still don't much like Custom, so I'll have another go.
First, it's not fluent. Asking an object to notify the user of a problem does exactly what it says. But asking it to notify the user of a problem custom sounds like something an anthropology professor would do at a missionary orientation.  (Never point at anyone in Japan...) Unless it's so heavily used that a long name is a nuisance, I'd prefer NotifyUserOfProblemUsingCustomButtons. But then, you don't _have_ to be using custom buttons...those arguments are optional. In fact, you can (as here) pass exactly the same arguments (though, confusingly, in a different order) as to NotifyUserOfProblem.
We want to think of this from the point of view of the client. Clients should never care that this way of reporting errors is a custom enhancement Bloom made to LibPalaso. But the fact that it's useful for custom buttons is worth considering.
When I'm trying to guess how to report an error the way I want, I want to type "ER." and have some options pop up like various overloads of ErrorReport.NotifyUserOfProblem. (It's a nuisance that the system typically brings up ErrorReporter, which doesn't have the method I want, and then I have to go searching, especially since ErrorReporter sounds like the thing that _should_ have a method to report an error. A report is not expected to do things. But it's too late to fix that, I expect.) 
There are two ways intellisense works well. One is that there are different methods, whose names come up as options (NotifyUserOfProblem...NotifyUserOfException...NotifyUserUsingCustomDialog...NotifyUserIfNewProblem....). The other is that there is one method, and when I select it, I'm offered a list of overloads, and hopefully it's obvious which set of arguments corresponds to what I want to do (the current situation with NotifyUserOfProblem). But if you combine them, it gets harder. I have to choose between NotifyUserOfProblem and NotifyUserOfProblemCustom before I can see what arguments each accepts. Do I use the second if I want custom buttons, a custom window title, a custom policy for showing the Report button, a custom way of displaying the stack trace in the exception, support for hot links in the text,...? If we need a new name, we want it to be really clear when you want the new method and when you want an overload of the old one. So maybe if the only situation where it makes sense to use the new one is to customize buttons, we should call it NotifyUserOfProblemUsingCustomButtons, and maybe even make the arguments obligatory, so no one is tempted to use it otherwise. Or we could borrow from WinApi and call it NotifyUserOfProblemExtended, indicating that it should be used when you want unusual, possibly advanced options...leaving open the possibility of adding more arguments and overloads of this name eventually, that are special in different ways.
Well, that's my best attempt to clarify. If you're still not convinced, maybe we should see if someone else can come up with a better idea than either of us. When all's said and done, it's only a name; I can live with NotifyUserOfProblemCustom if I can't convince anyone that something else is better.
</blockquote></details>

Yeah, I agree the ErrorReport vs ErrorReporter situation is quite confusing... but yeah, agree that it's too late/too big to address that here in this PR.

Ok, thanks for writing this up and explaining more thoughts/pain points/ideas. This helped me get a spark for some new ideas to address your pain points in a way that I'd be happy with.
I took some inspiration from C#'s XmlWriterSettings class too.
So, the idea of making a Settings parameter helped solve a few problems for me:
1) Previously, I have this sense that at least one of the settings-related parameters for the "Custom" function should be required, but all 3 of them were optional because none of them was required all the time. Having a Settings parameter allows me to make it so you have to specify a settings parameter in order to go into the "Custom" function.
2) This type disambiguates the overloading, allowing us to rename to NotifyUserOfProblem.

sujeffreyl (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5287)* status: 8 of 17 files reviewed, 5 unresolved discussions (waiting on @andrew-polk and @JohnThomson)

___
*[`src/BloomExe/ErrorReporter/BloomErrorReport.cs` line 15 at r4](https://reviewable.io/reviews/bloombooks/bloomdesktop/5287#-N6sHs-E39fJJ6BPEIfs:-N6sHs-E39fJJ6BPEIft:bcbe8jx) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/628ca727c02596a2d4184ebc971aa7a1dde85e27/src/BloomExe/ErrorReporter/BloomErrorReport.cs#L15)):*
> ```Smalltalk
> 		/// <param name="settings">Optional: The customization settings controlling which/how extra buttons show up</param>
> 		/// <param name="policy">Optional: The policy to use to decide whether to show the notification. Defaults to ShowAlwaysPolicy</param>
> 		public static void NotifyUserOfProblem(string message, NotifyUserOfProblemSettings settings, IRepeatNoticePolicy policy = null)
> ```

This adds 2 overloads to the existing ErrorReport... which already had 5.... and now BloomErrorReport has 7 total :(

I could reduce it from 7 down to 6 by having a single new overload (instead of 2 here) with the order (message, settings, exception=null, policy=null) or (settings, message, exception=null, policy=null). That has the pro of reducing the number of overloads, but it has the cons that I like the order of parameters less. (I feel like message is the most necessary/unique parameter and should be first. exception is optional but is often thought of in the same breath as message, so I like it being right after message).

JohnThomson (approved):
LGTM. However, there's a comment about merging something else first, and a request for @andrew-polk to review, and I'm not sure whether any other reviewer comments have not been addressed, so I'm not merging yet. Feel free to do so if those concerns have been addressed.

> Reviewed 9 of 9 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5287#-:-N6wuu0u-F-PwRyMjjuX:br2lj7k)* status: all files reviewed, 3 unresolved discussions (waiting on @andrew-polk and @sujeffreyl)

___
*[`src/BloomExe/ErrorReporter/BloomErrorReport.cs` line 15 at r4](https://reviewable.io/reviews/bloombooks/bloomdesktop/5287#-N6sHs-E39fJJ6BPEIfs:-N6wtK5-1iA_TgJ10oPt:bu5cye2) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/628ca727c02596a2d4184ebc971aa7a1dde85e27/src/BloomExe/ErrorReporter/BloomErrorReport.cs#L15)):*

This adds 2 overloads to the existing ErrorReport... which already had 5.... and now BloomErrorReport has 7 total :(

I could reduce it from 7 down to 6 by having a single new overload (instead of 2 here) with the order (message, settings, exception=null, policy=null) or (settings, message, exception=null, policy=null). That has the pro of reducing the number of overloads, but it has the cons that I like the order of parameters less. (I feel like message is the most necessary/unique parameter and should be first. exception is optional but is often thought of in the same breath as message, so I like it being right after message).
</blockquote></details>

I like the idea of keeping message and exception as the first arguments. If that means we need one more overload, I don't think that's a problem
___
*[`src/BloomExe/ErrorReporter/SentryErrorReporter.cs` line 35 at r2](https://reviewable.io/reviews/bloombooks/bloomdesktop/5287#-N6Umx3R6e3GMRKZDJFN:-N6wuVrt5-ozTx1SImk9:b-aemr3s) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/e041ac7770338d72a99fae15532184e471996b12/src/BloomExe/ErrorReporter/SentryErrorReporter.cs#L35)):*

Yeah, I agree the ErrorReport vs ErrorReporter situation is quite confusing... but yeah, agree that it's too late/too big to address that here in this PR.

Ok, thanks for writing this up and explaining more thoughts/pain points/ideas. This helped me get a spark for some new ideas to address your pain points in a way that I'd be happy with.
I took some inspiration from C#'s XmlWriterSettings class too.
So, the idea of making a Settings parameter helped solve a few problems for me:
1) Previously, I have this sense that at least one of the settings-related parameters for the "Custom" function should be required, but all 3 of them were optional because none of them was required all the time. Having a Settings parameter allows me to make it so you have to specify a settings parameter in order to go into the "Custom" function.
2) This type disambiguates the overloading, allowing us to rename to NotifyUserOfProblem.
</blockquote></details>

I like your new approach. If necessary we can keep enhancing the Settings object without adding yet more overloads to the main message.

andrew-polk:
I don't need to review.

sujeffreyl (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5287)* status: 0 of 17 files reviewed, 3 unresolved discussions (waiting on @andrew-polk and @JohnThomson)

___
*[`src/BloomExe/ErrorReporter/BloomErrorReport.cs` line 15 at r4](https://reviewable.io/reviews/bloombooks/bloomdesktop/5287#-N6sHs-E39fJJ6BPEIfs:-N71N9ZH95uDRcbA6XGg:b-7ihon4) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/628ca727c02596a2d4184ebc971aa7a1dde85e27/src/BloomExe/ErrorReporter/BloomErrorReport.cs#L15)):*

I like the idea of keeping message and exception as the first arguments. If that means we need one more overload, I don't think that's a problem
</blockquote></details>

Ok, sounds good, thanks!
___
*[`src/BloomExe/ErrorReporter/SentryErrorReporter.cs` line 35 at r2](https://reviewable.io/reviews/bloombooks/bloomdesktop/5287#-N6Umx3R6e3GMRKZDJFN:-N71N7KsCWuHL8-_4BNV:b-896fix) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/e041ac7770338d72a99fae15532184e471996b12/src/BloomExe/ErrorReporter/SentryErrorReporter.cs#L35)):*

I like your new approach. If necessary we can keep enhancing the Settings object without adding yet more overloads to the main message.
</blockquote></details>

Done.

sujeffreyl (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5287)* status: 0 of 17 files reviewed, 2 unresolved discussions (waiting on @andrew-polk and @JohnThomson)

___
*[`src/BloomExe/BloomExe.csproj` line 3 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5287#-N6Tn70HAMfFCNM3_TXI:-N7GabGf65cev6Z98vA2:bvx6l6b) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/fc832df434d36f7c84451aaccbc02c65f77b0c99/src/BloomExe/BloomExe.csproj#L3)):*

Waiting upon https://github.com/sillsdev/libpalaso/pull/1191 to be merged.
Then, will need to go back and edit this to beta0043 instead.
</blockquote></details>

The libpalaso PR has been merged and the libpalaso version in Bloom has been upgraded to beta0063 actually.

PR #5300, opened 2022-07-14
Allow tails to be dragged onto overlay images (BL-11184)

JohnThomson:
This change is [Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5300)

### DIFF
--- a/src/BloomBrowserUI/bloomUI.less
+++ b/src/BloomBrowserUI/bloomUI.less
@@ -12,11 +12,25 @@
 // there are probably other sources of ui font size.
 @ui-font-size: 10pt;
 
-@bubbleCanvasZIndex: 2;
 @dragHandleVisibleZIndex: @bubbleCanvasZIndex - 1;
 
-@imageContainerZIndex: 1000;
-@textOverPictureZIndex: @imageContainerZIndex + 1;
+// This name is almost obsolete, we do NOT want bloom-imageContainer to have a z-index,
+// since that makes it a stacking context, and then we can't get child buttons to
+// be above the comicaljs canvas. However, I'm keeping it because it is a basis for
+// several things we want to be above image containers, and also used for videos and
+// widgets.
+@baseZIndexForContainers: 1000;
+
+// needs to be above overlay images so tails can cross into them
+@bubbleCanvasZIndex: @baseZIndexForContainers + 1;
+
+// Above the canvas so they can be clicked
+@imageEditingButtonZIndex: @bubbleCanvasZIndex + 5;
+// And these are on top of those main buttons
+@miniImageEditingButtonZIndex: @imageEditingButtonZIndex + 1;
+
+// Text in bubbles needs to be above the bubble so we can see and edit it.
+@textOverPictureZIndex: @bubbleCanvasZIndex + 1;
 @formatButtonZIndex: @textOverPictureZIndex + 1;
 @dragHandleEventZIndex: @textOverPictureZIndex + 1;
 
--- a/src/BloomBrowserUI/bookEdit/css/editMode.less
+++ b/src/BloomBrowserUI/bookEdit/css/editMode.less
@@ -166,7 +166,7 @@ div.readOnlyInTranslationMode {
 .imageButton {
     width: 87px;
     height: 52px;
-    z-index: 100;
+    z-index: @imageEditingButtonZIndex;
     position: absolute;
     background-repeat: no-repeat;
     background-size: contain;
@@ -320,7 +320,7 @@ button.miniButton {
     background-repeat: no-repeat;
     background-size: contain;
     background-color: transparent;
-    z-index: 120;
+    z-index: @miniImageEditingButtonZIndex;
     &:hover {
         border: 1px inset @ImageButtonBorder;
     }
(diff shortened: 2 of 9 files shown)


jsubloom (changes requested):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5300)* status: all files reviewed, 2 unresolved discussions (waiting on @JohnThomson)

___
*[`src/BloomBrowserUI/bloomUI.less` line 22 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5300#-N8ZbLZ_-0LnIwvM1ks3:-N8ZaxaU7LxLbCRDEhRt:b73b3di) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/f96e628c3ec4afa18f27ebb78e27efd8be4a85ec/src/BloomBrowserUI/bloomUI.less#L22)):*
> ```Less
> // several things we want to be above image containers, and also used for videos and
> // widgets.
> @imageContainerZIndex: 1000;
> ```

Just some brainstorming here. Can we fix up this name now? Like, "aboveImageContainerBaselineZIndex"? idk, something that accurately describes its current usage.
___
*[`src/BloomBrowserUI/bookEdit/js/bubbleManager.ts` line 973 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5300#-N8ZdZT705Aeykhe5CZt:-N8ZdZT705Aeykhe5CZu:bm1l5wh) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/f96e628c3ec4afa18f27ebb78e27efd8be4a85ec/src/BloomBrowserUI/bookEdit/js/bubbleManager.ts#L973)):*
> ```XML
> 
>         if (
>             hoveredBubble &&
> ```

These 3 if conditions share a lot of overlapping conditions. I'm pretty sure that could be combined more into a series of connected if/else statments?
Something like this is what I had in mind.

if (hoveredBubble)
{
    if (activeElement && [...])
    {
      [...]
    }
    else if (!activeElement)
    {
       [...]
    }
}
else
{
    // Cleanup [...]
}

JohnThomson (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5300)* status: 4 of 9 files reviewed, 2 unresolved discussions (waiting on @sujeffreyl)

___
*[`src/BloomBrowserUI/bloomUI.less` line 22 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5300#-N8ZbLZ_-0LnIwvM1ks3:-N8dXYGo0mlJ9TNXel2H:bkxq87d) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/f96e628c3ec4afa18f27ebb78e27efd8be4a85ec/src/BloomBrowserUI/bloomUI.less#L22)):*

Just some brainstorming here. Can we fix up this name now? Like, "aboveImageContainerBaselineZIndex"? idk, something that accurately describes its current usage.
</blockquote></details>

I tried. It's actually used for other containers, not just images, so your name isn't right either.
___
*[`src/BloomBrowserUI/bookEdit/js/bubbleManager.ts` line 973 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5300#-N8ZdZT705Aeykhe5CZt:-N8dXhNm-kgRAv_hiCyM:b68ryl1) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/f96e628c3ec4afa18f27ebb78e27efd8be4a85ec/src/BloomBrowserUI/bookEdit/js/bubbleManager.ts#L973)):*

These 3 if conditions share a lot of overlapping conditions. I'm pretty sure that could be combined more into a series of connected if/else statments?
Something like this is what I had in mind.

if (hoveredBubble)
{
    if (activeElement && [...])
    {
      [...]
    }
    else if (!activeElement)
    {
       [...]
    }
}
else
{
    // Cleanup [...]
}
</blockquote></details>

I tried, and broke it. Note (what I missed) that the first "if" may CHANGE hoveredBubble to undefined, which then causes the second one to apply, so the second one can't just move to an 'else' on testing the original hoveredBubble. I think it is actually easier to understand the way it is. I added a comment.

jsubloom (approved):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5300)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @JohnThomson)

___
*[`src/BloomBrowserUI/bookEdit/js/bubbleManager.ts` line 973 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5300#-N8ZdZT705Aeykhe5CZt:-N8iHe1n3haiNhLoKmuQ:bn6ba9p) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/f96e628c3ec4afa18f27ebb78e27efd8be4a85ec/src/BloomBrowserUI/bookEdit/js/bubbleManager.ts#L973)):*

I tried, and broke it. Note (what I missed) that the first "if" may CHANGE hoveredBubble to undefined, which then causes the second one to apply, so the second one can't just move to an 'else' on testing the original hoveredBubble. I think it is actually easier to understand the way it is. I added a comment.
</blockquote></details>

Ah yes, that's right. Thanks for the explanation!

PR #5301, opened 2022-07-18
Add binding redirects

sujeffreyl:
PDFDroplet currently uses SIL.Core 9.*

---
This change is [Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5301)

### DIFF
--- a/src/BloomExe/app.config
+++ b/src/BloomExe/app.config
@@ -1,4 +1,4 @@
-<?xml version="1.0" encoding="utf-8"?>
+﻿<?xml version="1.0" encoding="utf-8"?>
 <configuration>
     <configSections>
         <sectionGroup name="userSettings" type="System.Configuration.UserSettingsGroup, System, Version=2.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089">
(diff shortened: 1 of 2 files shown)


sujeffreyl (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5301)* status: 0 of 2 files reviewed, 1 unresolved discussion

___
*[`src/BloomExe/app.config` line 168 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5301#-N7IIldCD-hcy-C8jDsV:-N7IIQAJDi8cf_0qWqws:b-g68nap) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/e345cf33ac08fcc421ed6ab68fcdcf14adf62831/src/BloomExe/app.config#L168)):*
> ```
>                 <bindingRedirect oldVersion="0.0.0.0-5.0.0.0" newVersion="5.0.0.0" />
>             </dependentAssembly>
>             <dependentAssembly>
> ```

An alternate approach (which we could still do) is to update PDFDroplet to version 10, but I figured binding redirect can catch it more broadly

sujeffreyl:
Let's just update PDFDroplet instead

PR #5306, opened 2022-07-25
Support internationalization of bubble labels (BL-11385)

JohnThomson:
This change is [Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5306)

### DIFF
--- a/src/BloomBrowserUI/bookEdit/js/BloomHintBubbles.ts
+++ b/src/BloomBrowserUI/bookEdit/js/BloomHintBubbles.ts
@@ -381,8 +381,13 @@ export default class BloomHintBubbles {
                 elementToAttachBubbleTo
             );
         } else {
+            // Most legacy label elements don't have data-i18n and are inserted into our
+            // localization dictionary with their English text as key. Newer ones can have
+            // explicit keys.
+            const l10nId =
+                elementWithBubbleAttributes.attr("data-i18n") || whatToSay;
             whatToSay = theOneLocalizationManager.getLocalizedHint(
-                whatToSay,
+                l10nId,
                 elementToAttachBubbleTo
             );
         }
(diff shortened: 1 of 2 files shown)


hatton (changes requested):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5306)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @hatton and @JohnThomson)

___
*[`src/BloomExe/Book/RuntimeInformationInjector.cs` line 276 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5306#-N8PuUGA3zLZQtfDpd9W:-N8PuUGA3zLZQtfDpd9X:b-w3qsc9) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/f9edcb58f4638d8578af621154d220b531ece2e8/src/BloomExe/Book/RuntimeInformationInjector.cs#L276)):*
> ```Smalltalk
> 				var key = label.GetAttribute("data-i18n");
> 				// Copied from AddTranslationToDictionaryUsingEnglishAsKey. I'm not sure whether we need it here.
> 				var translation = _collectDynamicStrings
> ```

It's not clear how dynamic string collection would be relevant here? Better to not create that dependency/connection if it's not needed.

JohnThomson (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5306)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @hatton)

___
*[`src/BloomExe/Book/RuntimeInformationInjector.cs` line 276 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5306#-N8PuUGA3zLZQtfDpd9W:-N8UQxOv0vM1lubElBKy:b-yse9qa) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/f9edcb58f4638d8578af621154d220b531ece2e8/src/BloomExe/Book/RuntimeInformationInjector.cs#L276)):*

It's not clear how dynamic string collection would be relevant here? Better to not create that dependency/connection if it's not needed.
</blockquote></details>

It took over an hour, but my conclusion is that there is no reason to use GetDynamicString ever. So I have removed it here, at least.

PR #5309, opened 2022-07-28
Add 6x9 inch page size, including full-bleed (BL-11386)

sujeffreyl:


JohnThomson (changes requested):
Looks good. Couple of very minor points worth at least considering.

> Reviewed 11 of 11 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5309#-:-N85umK55FiY2oAiN6uj:b-6mwtkq)* status: all files reviewed, 2 unresolved discussions (waiting on @sujeffreyl)

___
*[`DistFiles/localization/en/Bloom.xlf` line 3084 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5309#-N85sXWr7B3d_NlpWjzU:-N85sXWsBX1u-U1p3xba:bwsu7dr) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/813e9fb9734170137416f561fdc07c55f3b9e72a/DistFiles/localization/en/Bloom.xlf#L3084)):*
> ```XML
>                 <note>ID: LayoutChoices.B5Portrait</note>
>             </trans-unit>
> 			<trans-unit id="LayoutChoices.Cm13Landscape" sil:dynamic="true">
> ```

There seem to be whitespace changes here...spaces to tabs?...which may not be intended, though I'm not sure whether they are harmful.
___
*[`src/BloomExe/Publish/PDF/PdfMaker.cs` line 234 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5309#-N85tcDMA1ocQvS4i8Ry:-N85tcDMA1ocQvS4i8Rz:b-ubsqq4) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/813e9fb9734170137416f561fdc07c55f3b9e72a/src/BloomExe/Publish/PDF/PdfMaker.cs#L234)):*
> ```Smalltalk
> 					// Twice as big as B5 would be B4, so we'll use B4 as a substitue for 9x12".
> 					pageSize = PageSize.B4;	// 9.84in x 13.90 inches
> 					break;
> ```

I gather it doesn't support custom page sizes, so we can't get it right?

sujeffreyl (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5309)* status: 10 of 11 files reviewed, 2 unresolved discussions (waiting on @JohnThomson)

___
*[`DistFiles/localization/en/Bloom.xlf` line 3084 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5309#-N85sXWr7B3d_NlpWjzU:-N89DtrX3qnKreyjUYXh:brgsm8g) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/813e9fb9734170137416f561fdc07c55f3b9e72a/DistFiles/localization/en/Bloom.xlf#L3084)):*

There seem to be whitespace changes here...spaces to tabs?...which may not be intended, though I'm not sure whether they are harmful.
</blockquote></details>

Sorry. I ran Tab to Space on the whole file. Thankfully, these were the only places that had them mixed up.
___
*[`src/BloomExe/Publish/PDF/PdfMaker.cs` line 234 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5309#-N85tcDMA1ocQvS4i8Ry:-N89Qjcb95BxLPlTAgu9:bd9rceh) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/813e9fb9734170137416f561fdc07c55f3b9e72a/src/BloomExe/Publish/PDF/PdfMaker.cs#L234)):*

I gather it doesn't support custom page sizes, so we can't get it right?
</blockquote></details>

Wow! I found we actually can do custom page sizes. i have the code to do it now.

However, not sure which is better? Is it better to have the perfect page size? Or, is it nice to have a more standard size like B4. (I have no clue if 9x12" paper exists... although I can't even find evidence that 6x9" exists either, even though supposedly it's common in India! o_O)

JohnThomson (changes requested):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5309)* status: 10 of 11 files reviewed, 1 unresolved discussion (waiting on @JohnThomson and @sujeffreyl)

___
*[`src/BloomExe/Publish/PDF/PdfMaker.cs` line 234 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5309#-N85tcDMA1ocQvS4i8Ry:-N8AXyxG75SSUCqnQnKB:b9lvomf) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/813e9fb9734170137416f561fdc07c55f3b9e72a/src/BloomExe/Publish/PDF/PdfMaker.cs#L234)):*

Wow! I found we actually can do custom page sizes. i have the code to do it now.

However, not sure which is better? Is it better to have the perfect page size? Or, is it nice to have a more standard size like B4. (I have no clue if 9x12" paper exists... although I can't even find evidence that 6x9" exists either, even though supposedly it's common in India! o_O)
</blockquote></details>

I have frequently printed A5 booklets on US letter paper with reasonable results. The Acrobat print engine seems to adapt the size that the PDF was generated to the paper actually available. How well it does this may depend on how close the page sizes are, and/or on whether the ideal fit involves scaling up or down. Still, I think it's worth a try (since you said elsewhere you already have the code) to generate an accurate page size. If they need a different actual print size and our/Acrobat's print software doesn't handle it well, we can think some more.
Might be good to check with SteveMc. Linux printing is quite different, and IIRC Ghostscript (? or something) did very badly with page sizes it doesn't recognize.

JohnThomson (approved):
> Reviewed 1 of 1 files at r2, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5309)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @sujeffreyl)

PR #5323, opened 2022-08-05
[5.3] BL-11273 MXB tweaks

gmartin7:
* Set default HalfLetterPortrait page size
* fix "Funding Agency" line breaks
* add Device xmatters for Literacy and Scripture
* close up space under license block
* add custom spanish hint bubbles for various
   Scripture fields

- - -
This change is [Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5323)

### DIFF
--- a/src/BloomExe/Book/SizeAndOrientation.cs
+++ b/src/BloomExe/Book/SizeAndOrientation.cs
@@ -81,7 +81,7 @@ namespace Bloom.Book
 			{
 				var fileName = link.GetStringAttribute("href");
 				if (fileName.ToLowerInvariant().Contains("mode") || fileName.ToLowerInvariant().Contains("page") ||
-					fileName.ToLowerInvariant().Contains("matter") || fileName.ToLowerInvariant().Contains("languagedisplay") ||
+					fileName.ToLowerInvariant().Contains("languagedisplay") ||
 					fileName.ToLowerInvariant().Contains("origami") || fileName.ToLowerInvariant().Contains("defaultlangstyles") ||
 					fileName.ToLowerInvariant().Contains("customcollectionstyles") ||
 					// Ignore this obsolete styles file as well.  See https://issues.bloomlibrary.org/youtrack/issue/BL-9128.
--- a/src/content/branding/MXB-Book-Literacy/branding.json
+++ b/src/content/branding/MXB-Book-Literacy/branding.json
@@ -28,7 +28,7 @@
         {
             "data-book": "funding",
             "lang": "es",
-            "content": "Publicado por el <strong>Instituto Lingüístico de Verano, A. C.</strong> Oaxaca de Juárez",
+            "content": "Publicado por el<br /><strong>Instituto Lingüístico de Verano, A. C.</strong><br />Oaxaca de Juárez",
             "condition": "always"
         },
         {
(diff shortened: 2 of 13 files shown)


andrew-polk (changes requested):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5323)* status: all files reviewed, 1 unresolved discussion (waiting on @gmartin7)

___
*[`src/content/templates/xMatter/project-specific/MXBLiteracy-XMatter/MXBLiteracy-XMatter.less` line 4 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5323#-N8jIpIE1of0jmHoi9WT:-N8jIpIF4y9_M5Clqwhb:b-td7dyz) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/282446017074e5c61cfb650493db597438a67c9a/src/content/templates/xMatter/project-specific/MXBLiteracy-XMatter/MXBLiteracy-XMatter.less#L4)):*
> ```Less
> This is a dummy rule used to transmit to javascript what size/orientations this stylesheet can handle.
> This used to "lock down" a template to only certain layouts, now it just seems to cause a new book to
> default to the first layout.
> ```

So is there something else causing Big Book template to lock down?


I think I would be explicit in this comment that we are doing it to get a default (i.e. we don't want it locked down.
Actually, I question if the part about what it used to do it is at all helpful.
And the comment is somewhat inconsistent. It starts by saying this defines what is supported. But really we just want to define a default.

So I think I would rewrite the entire comment with something like,
"Used to make select the default size/layout when using this xmatter."

And do we need to add something to that to describe what happens if both the template and xmatter both try to set it?

(all this also applies to the Scripture xmatter)



_Code quote:_
```Less
This used to "lock down" a template to only certain layouts, now it just seems to cause a new book to
default to the first layout.
```

gmartin7 (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5323)* status: all files reviewed, 1 unresolved discussion (waiting on @andrew-polk)

___
*[`src/content/templates/xMatter/project-specific/MXBLiteracy-XMatter/MXBLiteracy-XMatter.less` line 4 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5323#-N8jIpIE1of0jmHoi9WT:-N8jd3P92W_7XBD8oN0W:btc6p9y) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/282446017074e5c61cfb650493db597438a67c9a/src/content/templates/xMatter/project-specific/MXBLiteracy-XMatter/MXBLiteracy-XMatter.less#L4)):*

So is there something else causing Big Book template to lock down?


I think I would be explicit in this comment that we are doing it to get a default (i.e. we don't want it locked down.
Actually, I question if the part about what it used to do it is at all helpful.
And the comment is somewhat inconsistent. It starts by saying this defines what is supported. But really we just want to define a default.

So I think I would rewrite the entire comment with something like,
"Used to make select the default size/layout when using this xmatter."

And do we need to add something to that to describe what happens if both the template and xmatter both try to set it?

(all this also applies to the Scripture xmatter)

</blockquote></details>

Interesting...
At this point I have:
```
This json comment in a default Bloom template "locks" the book to only these layouts,
whereas in a branding-induced xmatter pack it only acts to make this the default layout for a new book.
```

I'm not 100% sure what the difference is, but when you click in "Sources" to create a Big Book, the sizes get limited to what's in the comment in `template books/Big Book/Big Book.less`, whereas when you create any other book with the specified xmatter, the xmatter.less file just specifies the default size.

andrew-polk (approved):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5323)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @gmartin7)

PR #5333, opened 2022-08-11
Fix changing background color causing error (BL-11403)

sujeffreyl:
Another related issue is that the background color in the preview is often outdated.

- - -
This change is [Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5333)

### DIFF
--- a/src/BloomBrowserUI/publish/ReaderPublish/ThumbnailGroup.tsx
+++ b/src/BloomBrowserUI/publish/ReaderPublish/ThumbnailGroup.tsx
@@ -16,10 +16,10 @@ export const ThumbnailGroup: React.FunctionComponent<{
 const ThumbnailControl: React.FunctionComponent<{
     onChange?: () => void;
 }> = props => {
-    const [bookCoverColor, setBookCoverColor] = BloomApi.useApiStringState(
-        "publish/android/backColor",
-        "white"
-    );
+    const [
+        bookCoverColor,
+        setBookCoverColor
+    ] = BloomApi.useApiStringStatePromise("publish/android/backColor", "white");
     const [canModifyCurrentBook] = BloomApi.useApiBoolean(
         "common/canModifyCurrentBook",
         false
@@ -31,8 +31,8 @@ const ThumbnailControl: React.FunctionComponent<{
             disabled={!canModifyCurrentBook}
             imagePath="/bloom/api/publish/android/thumbnail?color="
             color={bookCoverColor}
-            onColorChanged={colorChoice => {
-                setBookCoverColor(colorChoice);
+            onColorChanged={async colorChoice => {
+                await setBookCoverColor(colorChoice);
                 if (props.onChange) {
                     props.onChange();
                 }
--- a/src/BloomBrowserUI/utils/bloomApi.ts
+++ b/src/BloomBrowserUI/utils/bloomApi.ts
@@ -349,7 +349,72 @@ export class BloomApi {
         urlSuffix: string,
         defaultValue: string,
         conditional?: () => boolean
-    ): [string, (value: string) => void] {
+    ) {
+        const generateSetStateWrapper = (
+            setState: React.Dispatch<React.SetStateAction<string>>
+        ) => {
+            const setStateWrapper = (value: string) => {
+                BloomApi.postString(urlSuffix, value);
+                setState(value);
+            };
+
+            return setStateWrapper;
+        };
+
+        return this.useApiStringStateInternal(
+            urlSuffix,
+            defaultValue,
+            conditional,
+            generateSetStateWrapper
+        );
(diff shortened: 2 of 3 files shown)


sujeffreyl (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5333)* status: 0 of 1 files reviewed, 1 unresolved discussion

___
*[`src/BloomExe/Publish/Android/PublishToAndroidApi.cs` line 167 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5333#-N9DTznpAv55EQ2w-QbT:-N9DTznpAv55EQ2w-QbU:b-7euobb) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/35695da4e4c1560748ff973349a3f697b0838e10/src/BloomExe/Publish/Android/PublishToAndroidApi.cs#L167)):*
> ```Smalltalk
> 			, true);
> 
> 			// FYI, handleOnUiThread needs to be true.
> ```

What about doing requiresSync=true vs handleOnUiThread=true?
    A) requiresSync seems more blocking -> more risky than handleOnUiThread, so it seems preferable to use handleOnUiThread
    B) I tested it out anyway, it didn't actually fix the issue. (kinda surprising to me)

sujeffreyl (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5333)* status: 0 of 1 files reviewed, 2 unresolved discussions

___
*[`src/BloomExe/Publish/Android/PublishToAndroidApi.cs` line 173 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5333#-N9DW5J206ubdM6XKXXh:-N9DW5J206ubdM6XKXXi:b1ud9x1) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/35695da4e4c1560748ff973349a3f697b0838e10/src/BloomExe/Publish/Android/PublishToAndroidApi.cs#L173)):*
> ```Smalltalk
> 			{
> 				MakeBloompubPreview(request, false);
> 			}, true);
> ```

Actually, an alternative way (which is more appealing to me right now) is to try to fix this on the Javascript side.  (Unless... someone else sees a reason why updatepreview should always be on UI Thread even if the order of these two requests wasn't an issue?)
I experimented with this before, but it didn't seem to fix the problem, which surprised me.
However, I now realized that the reason it didn't fix the problem is because I was making the wrong part of the code wait, but the code that was forcing the reset was still not waiting. I fixed it up and voila, I have a working prototype of the fix coming from the JS side of things. I'll work on that some more, and we can discuss it in Revision 2.

sujeffreyl (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5333)* status: 0 of 3 files reviewed, 2 unresolved discussions

___
*[`src/BloomExe/Publish/Android/PublishToAndroidApi.cs` line 173 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5333#-N9DW5J206ubdM6XKXXh:-N9Dkq688KEc5c58IXKd:b-73sj2b) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/35695da4e4c1560748ff973349a3f697b0838e10/src/BloomExe/Publish/Android/PublishToAndroidApi.cs#L173)):*

Actually, an alternative way (which is more appealing to me right now) is to try to fix this on the Javascript side.  (Unless... someone else sees a reason why updatepreview should always be on UI Thread even if the order of these two requests wasn't an issue?)
I experimented with this before, but it didn't seem to fix the problem, which surprised me.
However, I now realized that the reason it didn't fix the problem is because I was making the wrong part of the code wait, but the code that was forcing the reset was still not waiting. I fixed it up and voila, I have a working prototype of the fix coming from the JS side of things. I'll work on that some more, and we can discuss it in Revision 2.
</blockquote></details>

Ok, now fixed on the client side. (Because to me, I feel like the better perspective on the problem is that the client doesn't wait for the value to be updated before refreshing things). Thoughts?

andrew-polk (approved):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5333)* status: all files reviewed, 2 unresolved discussions (waiting on @andrew-polk)

andrew-polk (approved):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5333)* status: all files reviewed, 2 unresolved discussions (waiting on @andrew-polk)

andrew-polk (approved):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5333)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @sujeffreyl)

PR #5337, opened 2022-08-15
Avoid showing Progress dialog for wrong tab (BL-11282)

JohnThomson:


andrew-polk (changes requested):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5337)* status: all files reviewed, 1 unresolved discussion (waiting on @JohnThomson)

___
*[`src/BloomBrowserUI/react_components/Progress/ProgressDialog.tsx` line 279 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5337#-N9XaDGR3_yOHCHWg_C1:-N9XaDGR3_yOHCHWg_C2:bm7vpxr) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/879cce20559ff689573267be1046ccac77cc8297/src/BloomBrowserUI/react_components/Progress/ProgressDialog.tsx#L279)):*
> ```XML
>                                     // (I'm not sure whether we actually need the line above as well; I think it
>                                     // is necessary when using a non-embedded dialog.)
>                                     BloomApi.post("progress/close");
> ```

Sorry; this smells, but I'm not sure I can articulate exactly...
Happy to discuss, but here's my attempt to organize thoughts.

It seems like this moves us away from a concerted effort to have a common architecture for all dialogs (using BloomDialogPlumbing). Does more need to be done in BloomDialogPlumbing and/or EmbeddedProgressDialog to bring them more in alignment?

And another aspect is we have js telling C# to make a websocket call which we consume in js. Even if we needed this separate websocket call, can't we just issue in js?

JohnThomson (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5337)* status: all files reviewed, 1 unresolved discussion (waiting on @andrew-polk)

___
*[`src/BloomBrowserUI/react_components/Progress/ProgressDialog.tsx` line 279 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5337#-N9XaDGR3_yOHCHWg_C1:-N9YFgE_4cd7WetnJBZb:b-sd69k0) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/879cce20559ff689573267be1046ccac77cc8297/src/BloomBrowserUI/react_components/Progress/ProgressDialog.tsx#L279)):*

Sorry; this smells, but I'm not sure I can articulate exactly...
Happy to discuss, but here's my attempt to organize thoughts.

It seems like this moves us away from a concerted effort to have a common architecture for all dialogs (using BloomDialogPlumbing). Does more need to be done in BloomDialogPlumbing and/or EmbeddedProgressDialog to bring them more in alignment?

And another aspect is we have js telling C# to make a websocket call which we consume in js. Even if we needed this separate websocket call, can't we just issue in js?
</blockquote></details>

I agree, it smells a bit. I'm expecting this smell to go away eventually, when Bloom is all one web page. Then there will only be one instance of EmbeddedProgressDialog and it can close itself.
I don't know of any other BloomDialog that gets displayed in more than one root HTML document, so it doesn't seem we need a general solution (to a problem that is going away once we only have one root document).
I don't _think_ one web page can send to all the other client pages over our websocket; a send to the websocket would be an attempt to communicate to the server, but it's not listening on that channel AFAIK. But I don't know much about websocket except the way we use it to communicate from the server to the clients, so I could be wrong.
The other possible solution I came up with is that each instance of EmbeddedProgressDialog should have its own ID that would be part of the websocket protocol. Then the progress dialog in the Collection tab would not respond to messages intended for the one in the Audio/Video publish tab. But that complicates the API in a different way, one that will have to be undone as our UI moves towards a single web page with only one EmbeddedProgressDialog.
(Note that we can't directly communicate from one EmbeddedProgressDialog to another; they are in different root HTML documents, in different instances of GeckoFx.)
Happy to discuss other options...

andrew-polk (changes requested):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5337)* status: all files reviewed, 1 unresolved discussion (waiting on @JohnThomson)

___
*[`src/BloomBrowserUI/react_components/Progress/ProgressDialog.tsx` line 279 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5337#-N9XaDGR3_yOHCHWg_C1:-N9Yf-fz9uoysy434r9Y:be8adrb) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/879cce20559ff689573267be1046ccac77cc8297/src/BloomBrowserUI/react_components/Progress/ProgressDialog.tsx#L279)):*

I agree, it smells a bit. I'm expecting this smell to go away eventually, when Bloom is all one web page. Then there will only be one instance of EmbeddedProgressDialog and it can close itself.
I don't know of any other BloomDialog that gets displayed in more than one root HTML document, so it doesn't seem we need a general solution (to a problem that is going away once we only have one root document).
I don't _think_ one web page can send to all the other client pages over our websocket; a send to the websocket would be an attempt to communicate to the server, but it's not listening on that channel AFAIK. But I don't know much about websocket except the way we use it to communicate from the server to the clients, so I could be wrong.
The other possible solution I came up with is that each instance of EmbeddedProgressDialog should have its own ID that would be part of the websocket protocol. Then the progress dialog in the Collection tab would not respond to messages intended for the one in the Audio/Video publish tab. But that complicates the API in a different way, one that will have to be undone as our UI moves towards a single web page with only one EmbeddedProgressDialog.
(Note that we can't directly communicate from one EmbeddedProgressDialog to another; they are in different root HTML documents, in different instances of GeckoFx.)
Happy to discuss other options...
</blockquote></details>

Regarding js sending on the websocket: yes, of course you're right. I was mistakenly thinking of it as just a message it fires out for anyone listening, but it is more than that.

What do you think of the idea of having the `common/close` api call always send the close-progress message? At least that doesn't introduce another api call. 
But I recognize that isn't an ideal solution either.
Let's talk about it after the meeting tomorrow unless you come up with somethign before then.

JohnThomson (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5337)* status: all files reviewed, 1 unresolved discussion (waiting on @andrew-polk)

___
*[`src/BloomBrowserUI/react_components/Progress/ProgressDialog.tsx` line 279 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5337#-N9XaDGR3_yOHCHWg_C1:-N9YgiUn3Y74QkRNk5uM:b-y6qxbl) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/879cce20559ff689573267be1046ccac77cc8297/src/BloomBrowserUI/react_components/Progress/ProgressDialog.tsx#L279)):*

Regarding js sending on the websocket: yes, of course you're right. I was mistakenly thinking of it as just a message it fires out for anyone listening, but it is more than that.

What do you think of the idea of having the `common/close` api call always send the close-progress message? At least that doesn't introduce another api call. 
But I recognize that isn't an ideal solution either.
Let's talk about it after the meeting tomorrow unless you come up with somethign before then.
</blockquote></details>

It doesn't feel right to me that closing any arbitrary dialog would also close any progress dialog. What if, say, we wanted to ask a question about a particular book while the progress dialog is reporting on progress through a collection of BloomPUBs we are making?

hatton (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5337)* status: all files reviewed, 1 unresolved discussion (waiting on @andrew-polk)

___
*[`src/BloomBrowserUI/react_components/Progress/ProgressDialog.tsx` line 279 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5337#-N9XaDGR3_yOHCHWg_C1:-N9b-Ozi2vY0IWoiVtu4:ba6ayfa) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/879cce20559ff689573267be1046ccac77cc8297/src/BloomBrowserUI/react_components/Progress/ProgressDialog.tsx#L279)):*
>                                     // The above will close this instance; however, as long as Bloom has multiple
                                    // root windows that contain EmbeddedBrowserDialogs, it's possible that
                                    // more than one of them, though not visible, responded when the open progress
                                    // socket event happened. 

I just don't understand what that is saying



_Code quote:_
```XML
                                    // The above will close this instance; however, as long as Bloom has multiple
                                    // root windows that contain EmbeddedBrowserDialogs, it's possible that
                                    // more than one of them, though not visible, responded when the open progress
                                    // socket event happened. This post makes sure that a close event goes out
                                    // on the same channel, in case there's another one needing to be closed.
```

andrew-polk (changes requested):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5337)* status: all files reviewed, 1 unresolved discussion (waiting on @hatton and @JohnThomson)

___
*[`src/BloomBrowserUI/react_components/Progress/ProgressDialog.tsx` line 279 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5337#-N9XaDGR3_yOHCHWg_C1:-N9bHFfE015YjYbXHSlR:b-b0txjw) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/879cce20559ff689573267be1046ccac77cc8297/src/BloomBrowserUI/react_components/Progress/ProgressDialog.tsx#L279)):*

>                                     // The above will close this instance; however, as long as Bloom has multiple
                                    // root windows that contain EmbeddedBrowserDialogs, it's possible that
                                    // more than one of them, though not visible, responded when the open progress
                                    // socket event happened. 

I just don't understand what that is saying
</blockquote></details>

We discussed and decided we could go either way.
But I think JT is planning to add an id to each so it only opens the correct one.

JohnThomson:
We decided to fix this another way.

PR #5338, opened 2022-08-15
Order audio files properly for ePUB concatenation (BL-9016)

StephenMcConnel:
This change is [Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5338)

### DIFF
--- a/src/BloomExe/Publish/Epub/EpubMaker.cs
+++ b/src/BloomExe/Publish/Epub/EpubMaker.cs
@@ -1046,8 +1046,19 @@ namespace Bloom.Publish.Epub
 		/// </summary>
 		private string MergeAudioElements(IEnumerable<XmlElement> elementsWithAudio)
 		{
+			// The elementsWithAudio need to be ordered the same way as in bloom-player
+			// (narrationUtils.ts): by data-audio-order if it exists, or by document order
+			// if data-audio-order does not exist (or has equal values), with all elements
+			// having data-audio-order coming before any elements without.  (BL-9016)
+			var count = elementsWithAudio.Count();
+			var elementArray = elementsWithAudio.ToArray();
+			var keys = new List<(int index, int order)>(count);
+			for (int i = 0; i < count; ++i)
+				keys.Add((i, int.Parse(elementArray[i].GetOptionalStringAttribute("data-audio-order","999"), CultureInfo.InvariantCulture)));
+			Array.Sort(keys.ToArray(), elementArray, new CompareAudioOrder());
+
 			var mergeFiles =
-				elementsWithAudio
+				elementArray
 					.Select(s => AudioProcessor.GetOrCreateCompressedAudio(Storage.FolderPath, s.Attributes["id"]?.Value))
 					.Where(s => !string.IsNullOrEmpty(s));
 			Directory.CreateDirectory(Path.Combine(_contentFolder, kAudioFolder));
@@ -1061,6 +1072,19 @@ namespace Bloom.Publish.Epub
 			Logger.WriteEvent("Failed to merge audio files for page " + _pageIndex + " " + errorMessage);
 			// and we will do it the old way. Works for some readers.
 			return null;
+		}
+
+		private class CompareAudioOrder : System.Collections.IComparer
+		{
+			public int Compare(object x, object y)
+			{
+				var xx = ((int index, int order))x;
+				var yy = ((int index, int order))y;
+				var order = xx.order - yy.order;
+				if (order != 0)
+					return order;
+				return xx.index - yy.index;
+			}
 		}
 
 		private static string GetOverlayName(string pageDocName)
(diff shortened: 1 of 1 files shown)


andrew-polk (changes requested):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5338)* status: all files reviewed, 3 unresolved discussions (waiting on @StephenMcConnel)

___
*[`src/BloomExe/Publish/Epub/EpubMaker.cs` line 1047 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5338#-N9YY_Rm0UtCP9WJfktn:-N9YY_Rm0UtCP9WJfkto:bgobxaj) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/7cd341a3ee2f40015ba6775c8869eb9e66a99f6b/src/BloomExe/Publish/Epub/EpubMaker.cs#L1047)):*
> ```Smalltalk
> 			var countOrdered = elementsWithAudio
> 				.Select(x => !String.IsNullOrEmpty(x.GetStringAttribute("data-order")))
> 				.Sum(x => x?1:0);
> ```

Can we use `.Count()`?



_Code quote:_
```Smalltalk
.Sum(x => x?1:0)
```
___
*[`src/BloomExe/Publish/Epub/EpubMaker.cs` line 1049 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5338#-N9YYgENErYVydSruCld:-N9YYgENErYVydSruCle:byagjxo) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/7cd341a3ee2f40015ba6775c8869eb9e66a99f6b/src/BloomExe/Publish/Epub/EpubMaker.cs#L1049)):*
> ```Smalltalk
> 				.Sum(x => x?1:0);
> 			var count = elementsWithAudio.Count();
> 			if (countOrdered == count)
> ```

It seems like this will only attempt to order if every audio element has a tabIndex.

Is it true that opening the ordering UI is enough to save a tabIndex for every audio element? What if an audio element is added after the last time the ordering UI was open? Does it still get a tabIndex?

I'm asking because the bloom-player side does something a little more complex where it maintains dom order for elements with the same tabindex and puts elements with no tabindex last. And I was wondering if we need to follow that more closely.
___
*[`src/BloomExe/Publish/Epub/EpubMaker.cs` line 2609 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5338#-N9Ya3xV24nI0mNFDEh3:-N9Ya3xV24nI0mNFDEh4:b-23ec3t) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/7cd341a3ee2f40015ba6775c8869eb9e66a99f6b/src/BloomExe/Publish/Epub/EpubMaker.cs#L2609)):*
> ```Smalltalk
> 				if (tabIndex != "0" && classes.Contains("bloom-translationGroup"))
> 				{
> 					var audioElt = elt.SelectSingleNode(".//*[@data-duration]") as XmlElement;
> ```

Relying on the presence of `data-duration` feels a little fragile to me.
Would it be better to set `data-order` on all translation groups with nonzero tabindex?

Also, would it be better to call it `data-audio-order`?

StephenMcConnel (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5338)* status: all files reviewed, 3 unresolved discussions (waiting on @andrew-polk)

___
*[`src/BloomExe/Publish/Epub/EpubMaker.cs` line 1047 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5338#-N9YY_Rm0UtCP9WJfktn:-N9YljDa5j_VeN5i-6sa:b-hjg052) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/7cd341a3ee2f40015ba6775c8869eb9e66a99f6b/src/BloomExe/Publish/Epub/EpubMaker.cs#L1047)):*

Can we use `.Count()`?
</blockquote></details>

OK, I would need to use a .Where() instead of a .Select() to filter instead of map.  Then .Count() works.  But this code no longer exists.  :-)
___
*[`src/BloomExe/Publish/Epub/EpubMaker.cs` line 1049 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5338#-N9YYgENErYVydSruCld:-N9bnaLD1WcoatExAggi:botsg72) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/7cd341a3ee2f40015ba6775c8869eb9e66a99f6b/src/BloomExe/Publish/Epub/EpubMaker.cs#L1049)):*

It seems like this will only attempt to order if every audio element has a tabIndex.

Is it true that opening the ordering UI is enough to save a tabIndex for every audio element? What if an audio element is added after the last time the ordering UI was open? Does it still get a tabIndex?

I'm asking because the bloom-player side does something a little more complex where it maintains dom order for elements with the same tabindex and puts elements with no tabindex last. And I was wondering if we need to follow that more closely.
</blockquote></details>

I've totally revised the sorting code to follow what bloom-player does in narrationUtils.ts.
___
*[`src/BloomExe/Publish/Epub/EpubMaker.cs` line 2609 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5338#-N9Ya3xV24nI0mNFDEh3:-N9bLMJa71yMKRqD6Nka:b4vfok3) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/7cd341a3ee2f40015ba6775c8869eb9e66a99f6b/src/BloomExe/Publish/Epub/EpubMaker.cs#L2609)):*

Relying on the presence of `data-duration` feels a little fragile to me.
Would it be better to set `data-order` on all translation groups with nonzero tabindex?

Also, would it be better to call it `data-audio-order`?
</blockquote></details>

I changed the attribute name to data-audio-order.
I also changed the code to use the same HtmlDom method used to select the audio elements elsewhere.  The data-audio-order attribute needs to be placed on the audio elements for it to be available in the MergeAudioElements() data.

andrew-polk (approved):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5338)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @StephenMcConnel)

___
*[`src/BloomExe/Publish/Epub/EpubMaker.cs` line 1049 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5338#-N9YYgENErYVydSruCld:-N9bpH3S8rluem_N2s3s:bw5hc66) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/7cd341a3ee2f40015ba6775c8869eb9e66a99f6b/src/BloomExe/Publish/Epub/EpubMaker.cs#L1049)):*

I've totally revised the sorting code to follow what bloom-player does in narrationUtils.ts.
</blockquote></details>

Great. Thanks!

PR #5343, opened 2022-08-16
Warn user better about ePUB audio merge problems (BL-11435)

StephenMcConnel:
This change is [Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5343)

### DIFF
--- a/src/BloomExe/Publish/AudioProcessor.cs
+++ b/src/BloomExe/Publish/AudioProcessor.cs
@@ -153,14 +153,18 @@ namespace Bloom.Publish
 				// This may or may not be preferable to flattening all of the files to monaural at the
 				// standard recording rate before concatenating them.  The current approach at least
 				// has deterministic output for all files.
-				var monoFiles = TryEnsureConsistentInputFiles(ffmpeg, mergeFiles);
+				var monoFiles = TryEnsureConsistentInputFiles(ffmpeg, mergeFiles, out IEnumerable<string> mergeFileSpecs);
+				Debug.Assert(mergeFiles.Count() == mergeFileSpecs.Count());
 				try
 				{
-					var argsBuilder = new StringBuilder("-i \"concat:");
+					var argsBuilder = new StringBuilder("-hide_banner -i \"concat:");
 					argsBuilder.Append(string.Join("|", monoFiles));
 					argsBuilder.Append($"\" -c copy \"{combinedAudioPath}\"");
 					var result = CommandLineRunner.Run(ffmpeg, argsBuilder.ToString(), "", 60 * 10, new NullProgress());
-					var output = result.ExitCode != 0 ? result.StandardError : null;
+					// Having a WAV file masquerading as a .mp3 file in the file list doesn't work, but ffmpeg still
+					// returns an exit code of zero when that happens.  We can detect the problem by checking the stderr
+					// output.  See https://issues.bloomlibrary.org/youtrack/issue/BL-11435.
+					var output = CheckFfmpegErrors(result, mergeFiles, mergeFileSpecs);
 					if (output != null)
 					{
 						RobustFile.Delete(combinedAudioPath);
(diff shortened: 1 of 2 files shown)


JohnThomson (changes requested):
Looks basically good. A few minor suggestions.

> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5343#-:-N9fuoIP7WSvBoNGLV-6:b-b0b1at)* status: all files reviewed, 5 unresolved discussions (waiting on @StephenMcConnel)

___
*[`src/BloomExe/Publish/AudioProcessor.cs` line 193 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5343#-N9foaYZBaUufCu0aVrz:-N9foaYZBaUufCu0aVs-:b9sci7z) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/7a356712b80cea8593dedbdd0c486676f404915b/src/BloomExe/Publish/AudioProcessor.cs#L193)):*
> ```Smalltalk
> 				return stderr;
> 			// This error message can occur no more than once in a valid mp3 file.
> 			var matches = Regex.Matches(stderr, "Audio packet of size [0-9]+ \\(starting with [0-9A-F]+\\.\\.\\.\\) is invalid, writing it anyway\\.", RegexOptions.CultureInvariant);
> ```

Is there any concern that FFMPEG might localize these messages if the system language is something other than English?
___
*[`src/BloomExe/Publish/AudioProcessor.cs` line 208 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5343#-N9fnrnODbuhBiA-p_aM:-N9fnrnODbuhBiA-p_aN:b-bxk2zx) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/7a356712b80cea8593dedbdd0c486676f404915b/src/BloomExe/Publish/AudioProcessor.cs#L208)):*
> ```Smalltalk
> 						errorReport.AppendLine(specs[i]);
> 					}
> 					return errorReport.ToString();
> ```

It's not obvious how this amounts to an error report. I eventually figured out that it's making a list of file names, each paired with "unknown" if we can't tell what it is, or the specs of what kind of audio stream it contains, and which will eventually be combined with a message that we couldn't merge the audio files for this page. Maybe a comment to this effect would be helpful?
___
*[`src/BloomExe/Publish/AudioProcessor.cs` line 227 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5343#-N9frw6R69gxbWpTWNXK:-N9frw6R69gxbWpTWNXL:b-xf4l0y) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/7a356712b80cea8593dedbdd0c486676f404915b/src/BloomExe/Publish/AudioProcessor.cs#L227)):*
> ```Smalltalk
> 		/// If there's only one file in the list, just return the input list.
> 		/// If any file fails to convert for any reason, it is returned in the output list.  So the
> 		/// files in the output list may not really be consistent in reality...
> ```

I think it would be helpful to add something like, "mergeFileSpecs is returned with Ffmpeg's report on the contents of each input, or 'unknown' if it is not recognized as MP3."
___
*[`src/BloomExe/Publish/AudioProcessor.cs` line 239 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5343#-N9ftNZd4ffHcO3USl69:-N9ftNZd4ffHcO3USl6A:b-yw8upy) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/7a356712b80cea8593dedbdd0c486676f404915b/src/BloomExe/Publish/AudioProcessor.cs#L239)):*
> ```Smalltalk
> 			if (mergeFiles.Count() < 2)
> 			{
> 				internalFileSpecs.Add("unknown");
> ```

We're not going to use this, are we? But it's the same string that usually means "this is a problem file". Or maybe this is never going to happen at all, since we won't try to merge a single file? Anyway, I'm wondering whether we should "fail fast" better if we returned null for mergeFileSpecs, or even threw immediately if count is less than two. But at least, I think it would be better to return something different, like "not checked".
___
*[`src/BloomExe/Publish/AudioProcessor.cs` line 280 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5343#-N9fuFRw3Su31tb0LSwQ:-N9fuFRw3Su31tb0LSwR:bcsxlx4) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/7a356712b80cea8593dedbdd0c486676f404915b/src/BloomExe/Publish/AudioProcessor.cs#L280)):*
> ```Smalltalk
> 				else
> 				{
> 					internalFileSpecs.Add("unknown");
> ```

Would it be clearer to say something like "unknown...does not appear to be a valid MP3 file"? or "unknown...Bloom does not recognize this file as a valid MP3"?

JohnThomson (approved):
This LGTM, but I did have one suggestion to consider:
I gather that if the user has a wav file pretending to be mp3, they'll get a message like
Could not merge audio for page 2
.../audio/good.mp3
MP#3 44100 Hz...
../audio/badFile.mp3
WAV 44100 Hz.....

That's fairly subtle and technical. What would you think of "Could not merge audio for page 2. Possibly one of the files is not really an MP3.
(And possibly add "Details of what we think the files are follow" or something similar)
@hatton 
(If you and/or John don't like this idea, it's OK to merge as-is.) 

> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5343#-:-N9hAd_w8OTDaC9BklQ3:btf1c3f)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @StephenMcConnel)

hatton (commented):
How about `There was a problem processing one of the audio files on page 2: (list the files or log or whatever)`

> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5343#-:-N9hD5j153Nij5cBGoVv:b-43skvs)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @StephenMcConnel)

StephenMcConnel (commented):
I've adopted @hatton's suggestion.

> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5343#-:-N9hGACoBXDtQ9aTKoTD:bqqoyj3)* status: 1 of 2 files reviewed, all discussions resolved (waiting on @JohnThomson)

JohnThomson (approved):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5343)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @StephenMcConnel)

PR #5361, opened 2022-08-30
WebView2 for PDF preview and Wall Calendar initialization

JohnThomson:
This change is [Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5361)

### DIFF
--- a/src/BloomExe/Edit/Configurator.cs
+++ b/src/BloomExe/Edit/Configurator.cs
@@ -2,10 +2,12 @@
 using System.Collections.Generic;
 using System.Linq;
 using System.IO;
+using System.Text;
 using System.Windows.Forms;
 using System.Xml;
 using Bloom.ToPalaso.Experimental;
 using Bloom.Api;
+using Bloom.Book;
 using Newtonsoft.Json.Linq;
 using SIL.Code;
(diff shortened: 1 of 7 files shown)


hatton (approved):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5361)* status: all files reviewed, 1 unresolved discussion (waiting on @JohnThomson)

___
*[`src/BloomExe/Publish/PDF/PdfViewer.cs` line 51 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5361#-NAkCzpN-0aHy1z3cMus:-NAkCzpN-0aHy1z3cMut:b5m7nwy) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/3981ce408d9b4d3647cdbdb70f5f83a109df4f28/src/BloomExe/Publish/PDF/PdfViewer.cs#L51)):*
> ```Smalltalk
> 				// Review: do we still want to try Acrobat? We're trying to get rid of it,
> 				// so I'm inclined not to at least while WV2 is experimental.
> 				var wv2 = new WebView2Browser();
> ```

No biggie, but in case you have to do a new commit, this only needs one line, not two.

PR #5364, opened 2022-09-01
Image Compression settings in publish-settings.json (BL-11242)

sujeffreyl:


andrew-polk (changes requested):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5364)* status: all files reviewed, 2 unresolved discussions (waiting on @sujeffreyl)

___
*[`src/BloomExe/Book/BookCompressor.cs` line 251 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5364#-NAuTD6xDurmyO1AL8Fj:-NAuTD6yCKeUqA-wncy9:bq8avcs) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/95fc9ecaa97526c7bbc851b76a93c78a7aabf850/src/BloomExe/Book/BookCompressor.cs#L251)):*
> ```Smalltalk
> 					newEntry.Size = modifiedContent.Length;
> 				}
> 				// CompressBookForDevice is always called with non-null imagePublishSettings (that is, requesting to reduce images)
> ```

This comment concerns me.
I *think* the author of it is trying to communicate that it is safe to assume we want to process videos because reduceImages is set and therefore we are in the compress book context. 
Do we need a new boolean to control this?
What if we start setting imagePublishSettings for something else?



_Code quote:_
```Smalltalk
// CompressBookForDevice is always called with non-null imagePublishSettings (that is, requesting to reduce images)
```
___
*[`src/BloomExe/Book/PublishSettings.cs` line 320 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5364#-NAuU-Hj5vLU8sUikq91:-NAuU-Hj5vLU8sUikq92:b-23s86y) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/95fc9ecaa97526c7bbc851b76a93c78a7aabf850/src/BloomExe/Book/PublishSettings.cs#L320)):*
> ```Smalltalk
> 		/// </summary>
> 		[JsonProperty("images")]
> 		public ImagePublishSettings Images { get; set; }
> ```

Should this be ImageSettings?



_Code quote:_
```Smalltalk
Images
```

andrew-polk (changes requested):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5364)* status: all files reviewed, 2 unresolved discussions (waiting on @sujeffreyl)

___
*[`src/BloomExe/Book/BookCompressor.cs` line 251 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5364#-NAuTD6xDurmyO1AL8Fj:-NAylU9u8CuMFZ7bvy3B:bmxp0rl) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/95fc9ecaa97526c7bbc851b76a93c78a7aabf850/src/BloomExe/Book/BookCompressor.cs#L251)):*

Your last question seems to be similar to the scenario in the old code "What if we start setting reduceImages=true for something else [but don't wan't to compress videos]".    I believe that the answer is the same to both questions and that this PR didn't materially change anything regarding this "if".

We could bring back reduceImages, which does simplify the PR in the short-term, but I didn't like in the long-term that the code had 2 parameters for one idea.

However, if you want to introduce a separate "reduceVideos" or "compressVideos" boolean to control the video compression separate from the image compression, I'm totally down for that.  I didn't like how the old code was commandeering the reduceImages parameter to control video compression.
</blockquote></details>

I agree that we never should have let the logic be commandeered.
I agree you haven't materially changed the logic. 
I think now is a good time to clean it up while we are looking at it.
And it *feels* like we are making things less safe since we are moving from a boolean to the existence of an object.

I think part of what is tripping me up is it seems odd, maybe even wrong?, to base whether or not we compress images on the existence of a settings object. It would be more natural to base it on a setting in that object. 
Going back to the idea that one day we will want other settings, we would have to introduce such logic then. Does that make sense?
___
*[`src/BloomExe/Book/PublishSettings.cs` line 320 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5364#-NAuU-Hj5vLU8sUikq91:-NAylI249OGa2sHT49UL:bq3tqpe) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/95fc9ecaa97526c7bbc851b76a93c78a7aabf850/src/BloomExe/Book/PublishSettings.cs#L320)):*

Maybe, maybe not.
1) The spec asks for a json field named "images". It's kinda sounds like a misnomer in isolation, but in the context of being in a file called publish-settings.json, I guess the interpretation should be "the publish settings for images", which is right. And it's parallel to the other fields, which theoretically also have the same problem even though they feel less off.  I initally preferred something like "imageSettings", but if you consider the above points,  you'll realize that "imageSettings" is not really great either...
2) All the other fields are named by just Pascal Casing the json field, even when an extra word might have made it clearer (or redundant, depending on your point of view)
3) There is some confusion if the property name doesn't correspond to the JSON field name.
Therefore, in the end I decided to go for "just Pascal Case the name"

Does it make you feeling any better if in practice, this will get read as "settings.Images"?
</blockquote></details>

The problem is that in code, it looks like this:
`bookInfo.PublishSettings.BloomPub.Images`
which is pretty confusing, in my opinion.
Indeed, even in the json, it is `bloomPUB {images {}}`.
In what context will it display as `settings.Images`?

The other settings feel less off because they are actually defining the thing. TextLangs give  a list of langs, etc. This feel like it should give a list of images. So comparing to the other settings actually makes this feel more wrong, in my opinion. Maybe it is the spec that is wrong. I'll ask about it in the card.



_Code quote (from src/BloomExe/Publish/Android/AndroidPublishSettings.cs):_
```Smalltalk
bookInfo.PublishSettings.BloomPub.Images
```

sujeffreyl (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5364)* status: all files reviewed, 2 unresolved discussions (waiting on @andrew-polk and @sujeffreyl)

___
*[`src/BloomExe/Book/BookCompressor.cs` line 251 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5364#-NAuTD6xDurmyO1AL8Fj:-NAyyPQ7EoePNLVaI_Lh:btzjdz3) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/95fc9ecaa97526c7bbc851b76a93c78a7aabf850/src/BloomExe/Book/BookCompressor.cs#L251)):*

I agree that we never should have let the logic be commandeered.
I agree you haven't materially changed the logic. 
I think now is a good time to clean it up while we are looking at it.
And it *feels* like we are making things less safe since we are moving from a boolean to the existence of an object.

I think part of what is tripping me up is it seems odd, maybe even wrong?, to base whether or not we compress images on the existence of a settings object. It would be more natural to base it on a setting in that object. 
Going back to the idea that one day we will want other settings, we would have to introduce such logic then. Does that make sense?
</blockquote></details>

So.... maybe we should re-introduce reduceImages, and also a reduceVideo?

(We could also just re-introduce reduceImages and rename it to reduceImagesAndVideo or reduceMedia/etc.)

sujeffreyl (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5364)* status: all files reviewed, 2 unresolved discussions (waiting on @andrew-polk)

___
*[`src/BloomExe/Book/PublishSettings.cs` line 320 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5364#-NAuU-Hj5vLU8sUikq91:-NAyyw-b4T_1FFe8O8Yt:b-5nc22u) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/95fc9ecaa97526c7bbc851b76a93c78a7aabf850/src/BloomExe/Book/PublishSettings.cs#L320)):*

The problem is that in code, it looks like this:
`bookInfo.PublishSettings.BloomPub.Images`
which is pretty confusing, in my opinion.
Indeed, even in the json, it is `bloomPUB {images {}}`.
In what context will it display as `settings.Images`?

The other settings feel less off because they are actually defining the thing. TextLangs give  a list of langs, etc. This feel like it should give a list of images. So comparing to the other settings actually makes this feel more wrong, in my opinion. Maybe it is the spec that is wrong. I'll ask about it in the card.
</blockquote></details>

sorry you're right, i was mixing it up with AndroidPublishSettings.

andrew-polk (changes requested):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5364)* status: all files reviewed, 2 unresolved discussions (waiting on @andrew-polk and @sujeffreyl)

___
*[`src/BloomExe/Book/BookCompressor.cs` line 251 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5364#-NAuTD6xDurmyO1AL8Fj:-NAz-Ue-5OlYVnSeiCxB:bj0y05f) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/95fc9ecaa97526c7bbc851b76a93c78a7aabf850/src/BloomExe/Book/BookCompressor.cs#L251)):*

So.... maybe we should re-introduce reduceImages, and also a reduceVideo?

(We could also just re-introduce reduceImages and rename it to reduceImagesAndVideo or reduceMedia/etc.)
</blockquote></details>

Hm. Looks like it is actually an `isForDevice` flag.
This else if is not only processing video, but creating the version.txt file.
I think that one flag should do it. What do you think?

Sorry to push you through all this, but I think it is a good improvement.

sujeffreyl (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5364)* status: 0 of 5 files reviewed, 2 unresolved discussions (waiting on @andrew-polk)

___
*[`src/BloomExe/Book/BookCompressor.cs` line 251 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5364#-NAuTD6xDurmyO1AL8Fj:-NBIhm285wR_m01XAj2-:b-bgugnk) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/95fc9ecaa97526c7bbc851b76a93c78a7aabf850/src/BloomExe/Book/BookCompressor.cs#L251)):*

Hm. Looks like it is actually an `isForDevice` flag.
This else if is not only processing video, but creating the version.txt file.
I think that one flag should do it. What do you think?

Sorry to push you through all this, but I think it is a good improvement.
</blockquote></details>

Add "forDevice" flag, to match "forReaderTools
___
*[`src/BloomExe/Book/PublishSettings.cs` line 320 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5364#-NAuU-Hj5vLU8sUikq91:-NBIhpclE7itlXIUIDAY:bgrlyve) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/95fc9ecaa97526c7bbc851b76a93c78a7aabf850/src/BloomExe/Book/PublishSettings.cs#L320)):*

sorry you're right, i was mixing it up with AndroidPublishSettings.
</blockquote></details>

Change to imageSettings as agreed upon in the card

andrew-polk (approved):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5364)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @sujeffreyl)

PR #5371, opened 2022-09-01
Don't wait forever for a startup milestone

JohnThomson:
May fix problem with splash screen not closing (and get more info).

- - -
This change is [Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5371)

### DIFF
--- a/src/BloomExe/MiscUI/StartupScreenManager.cs
+++ b/src/BloomExe/MiscUI/StartupScreenManager.cs
@@ -5,6 +5,7 @@ using System.Linq;
 using System.Threading;
 using System.Windows.Forms;
 using SIL.IO;
+using SIL.Reporting;
 
 namespace Bloom.MiscUI
 {
@@ -150,8 +151,18 @@ namespace Bloom.MiscUI
 
 			if (_current.WaitForMilestone != null && !_milestones.Contains(_current.WaitForMilestone))
 			{
-				_current = null; // nothing in progress we need to wait for
-				return; // we'll try again on another idle event
+				if (DateTime.Now < _earliestWeShouldCloseTheSplashScreen.Add(new TimeSpan(0, 0, 5)))
+				{
+					_current = null; // nothing in progress we need to wait for
+					return; // we'll try again on another idle event
+				}
+				else
+				{
+					Logger.WriteEvent("Failed to get to startup milestone " + _current.WaitForMilestone +
+						" after five seconds");
+					// And go ahead with the next task, so we're not stuck forever.
+					// (Currently, waiting is just to help things happen in the optimum order for performance.)
+				}
 			}
 
 			if (_current.ShouldHideSplashScreen)


andrew-polk (changes requested):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5371)* status: all files reviewed, 1 unresolved discussion (waiting on @JohnThomson)

___
*[`src/BloomExe/MiscUI/StartupScreenManager.cs` line 165 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5371#-NAyu5Lq3NX9RvkSRzyF:-NAyu5Lq3NX9RvkSRzyG:b-fgb8ii) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/75a5f309d3f65329fdd2d253968d06e9af232386/src/BloomExe/MiscUI/StartupScreenManager.cs#L165)):*
> ```Smalltalk
> 					// And go ahead with the next task, so we're not stuck forever.
> 					// (Currently, waiting is just to help things happen in the optimum order for performance.)
> 				}
> ```

If I understand correctly, if we leave this in, it means the problem scenario is taking Bloom about 5 seconds longer to load than we think it should need.

If that's correct, I would like to treat this change as a temporary one used to glean information for a real fix rather than something we consider leaving in.

Given that, is there more we should be capturing? `_milestones`, maybe? Any timing?

(Also, could this be exposing a real problem that some milestone never actually happens?)

JohnThomson (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5371)* status: all files reviewed, 1 unresolved discussion (waiting on @andrew-polk)

___
*[`src/BloomExe/MiscUI/StartupScreenManager.cs` line 165 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5371#-NAyu5Lq3NX9RvkSRzyF:-NB-MUxu7F-dZQTUCpwu:bpdeo1s) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/75a5f309d3f65329fdd2d253968d06e9af232386/src/BloomExe/MiscUI/StartupScreenManager.cs#L165)):*

If I understand correctly, if we leave this in, it means the problem scenario is taking Bloom about 5 seconds longer to load than we think it should need.

If that's correct, I would like to treat this change as a temporary one used to glean information for a real fix rather than something we consider leaving in.

Given that, is there more we should be capturing? `_milestones`, maybe? Any timing?

(Also, could this be exposing a real problem that some milestone never actually happens?)
</blockquote></details>

Not exactly. The time I'm using as a benchmark is called "earliestWeShouldCloseTheSplashScreen"...IOW, we want it up at least this long to give people time to read it, even if Bloom starts up faster. But I'm pretty sure typically Bloom is slower than that, at least on most hardware.
Also, we're not slowing down startup if all goes well; we're just putting an upper limit on how long we'll wait before trying to select the current book and then closing the splash screen. It would be nice to have a real fix for whatever is preventing the milestone notification we're expecting, which is what the extra logging is about; but I'm inclined to think this is a safeguard worth keeping, since the always-on-top splash screen is a pain to have around for long. Even if we can't find a better fix, the user will seldom see this delay, since the problem only occurs when the collection is empty.
However, I am hoping for confirmation (or refutation) of my hypothesis that we're not getting this milestone message. Then the puzzle remains: why?
So far I haven't thought of anything else to log, but I'm open to suggestions. I don't have any current hypothesis for what might prevent it (only in release builds, only in WV2) so it's hard to know what to instrument.

andrew-polk (approved):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5371)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @JohnThomson)

___
*[`src/BloomExe/MiscUI/StartupScreenManager.cs` line 165 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5371#-NAyu5Lq3NX9RvkSRzyF:-NB-Pj4M0mNdvv1deNFf:b-a83ydc) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/75a5f309d3f65329fdd2d253968d06e9af232386/src/BloomExe/MiscUI/StartupScreenManager.cs#L165)):*

Not exactly. The time I'm using as a benchmark is called "earliestWeShouldCloseTheSplashScreen"...IOW, we want it up at least this long to give people time to read it, even if Bloom starts up faster. But I'm pretty sure typically Bloom is slower than that, at least on most hardware.
Also, we're not slowing down startup if all goes well; we're just putting an upper limit on how long we'll wait before trying to select the current book and then closing the splash screen. It would be nice to have a real fix for whatever is preventing the milestone notification we're expecting, which is what the extra logging is about; but I'm inclined to think this is a safeguard worth keeping, since the always-on-top splash screen is a pain to have around for long. Even if we can't find a better fix, the user will seldom see this delay, since the problem only occurs when the collection is empty.
However, I am hoping for confirmation (or refutation) of my hypothesis that we're not getting this milestone message. Then the puzzle remains: why?
So far I haven't thought of anything else to log, but I'm open to suggestions. I don't have any current hypothesis for what might prevent it (only in release builds, only in WV2) so it's hard to know what to instrument.
</blockquote></details>

I still don't think I understand why it is safe to potentially let the program keep going when we have carefully set up a milestone to wait for and aren't hitting it. 
Maybe I'm making a bad assumption. I think this class is about more than just simply when to hide the splash screen. But if that's all it does, then I guess it must be safe.

But as you are the author of this and the one making the change and you are going to monitor the results of this change, I'm going to merge so you can get the additional info.

JohnThomson (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5371)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @JohnThomson)

___
*[`src/BloomExe/MiscUI/StartupScreenManager.cs` line 165 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5371#-NAyu5Lq3NX9RvkSRzyF:-NB04itj2OHqhO3WMh0P:b-cj1faa) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/75a5f309d3f65329fdd2d253968d06e9af232386/src/BloomExe/MiscUI/StartupScreenManager.cs#L165)):*

I still don't think I understand why it is safe to potentially let the program keep going when we have carefully set up a milestone to wait for and aren't hitting it. 
Maybe I'm making a bad assumption. I think this class is about more than just simply when to hide the splash screen. But if that's all it does, then I guess it must be safe.

But as you are the author of this and the one making the change and you are going to monitor the results of this change, I'm going to merge so you can get the additional info.
</blockquote></details>

The reason for the milestone was not to delay selecting the current element until it would work, but to delay it so that the buttons of the current collection (which are fast) would render before the preview (which can take a long time to come up, especially for larger books). (I don't recall why it works to do it earlier, but I have a clear recollection that this milestone stuff was just an optimization.) So if we don't get the optimization, especially when there are no buttons to draw, it shouldn't matter much.

PR #5375, opened 2022-09-06
Fix Linux build problems related to WebView2 work (20220906)

StephenMcConnel:
This change is [Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5375)

### DIFF
--- a/src/BloomExe/Edit/EditingView.cs
+++ b/src/BloomExe/Edit/EditingView.cs
@@ -514,7 +514,7 @@ namespace Bloom.Edit
 				// here is to set a flag that page selection is still processing and block any
 				// further page selecting until the current page has finished loading.
 				_model.PageSelectionStarted();
-				_browser1.WebBrowser.DocumentCompleted += WebBrowser_DocumentCompleted;
+				(_browser1 as GeckoFxBrowser).WebBrowser.DocumentCompleted += WebBrowser_DocumentCompleted;
 #endif
 			}
 #if MEMORYCHECK
@@ -557,7 +557,7 @@ namespace Bloom.Edit
 		void WebBrowser_DocumentCompleted(object sender, EventArgs e)
 		{
 			_model.PageSelectionFinished();
-			_browser1.WebBrowser.DocumentCompleted -= WebBrowser_DocumentCompleted;
+			(_browser1 as GeckoFxBrowser).WebBrowser.DocumentCompleted -= WebBrowser_DocumentCompleted;
 		}
 #endif
 
--- a/src/BloomExe/GeckoFxBrowser.cs
+++ b/src/BloomExe/GeckoFxBrowser.cs
@@ -1,26 +1,18 @@
 using System;
-using System.ComponentModel;
 using System.Diagnostics;
 using System.Drawing;
 using System.IO;
 using System.Linq;
-using System.Net;
 using System.Reflection;
 using System.Text;
 using System.Windows.Forms;
-using System.Windows.Navigation;
-using System.Xml;
 using Bloom.Book;
 using Bloom.Api;
 using Bloom.Edit;
 using Gecko;
-using Gecko.DOM;
 using Gecko.Events;
-using SIL.IO;
 using SIL.Reporting;
 using SIL.Windows.Forms.Miscellaneous;
-using L10NSharp;
-using SimulatedPageFileSource = Bloom.Api.BloomServer.SimulatedPageFileSource;
 
 namespace Bloom
 {
@@ -30,6 +22,7 @@ namespace Bloom
 		bool _browserIsReadyToNavigate;
 		protected string _url;
 
+		private bool _offScreen = false;
 
 		private PasteCommand _pasteCommand;
 		private CopyCommand _copyCommand;
@@ -187,8 +180,13 @@ namespace Bloom
 			}
 		}
 
-		public GeckoFxBrowser()
+		public GeckoFxBrowser(bool offScreen = false)
 		{
+#if __MonoCS__
+			_offScreen = offScreen;
+			if (_offScreen)
+				_browser = new OffScreenGeckoWebBrowser();	// can't wait for loading
+#endif
 			InitializeComponent();
 			_isolator = NavigationIsolator.GetOrCreateTheOneNavigationIsolator();
 		}
@@ -408,7 +406,8 @@ namespace Bloom
 				return;
 			}
 
-			_browser = new GeckoWebBrowser();
+			if (!_offScreen)
+				_browser = new GeckoWebBrowser();
 
 			_browser.Parent = this;
 			_browser.Dock = DockStyle.Fill;
@@ -533,12 +532,26 @@ namespace Bloom
 				ControlKeyEvent.Raise(keyData);
 			}
 #if __MonoCS__
-			_controlPressed = e.CtrlKey && e.KeyChar == 'm';	// m for menu...
+			_lastKeypressWasCtrlM = e.CtrlKey && e.KeyChar == 'm';	// m for menu...
 #endif
 		}
 
 #if __MonoCS__
-		private bool _controlPressed;
+		private bool _lastKeypressWasCtrlM;
+		/// <summary>
+		/// The overridden method uses code that only works on Windows to determine whether the CTRL-key
+		/// is down during a right click.  This is not exactly equivalent but the best we can do on Linux:
+		/// pressing Ctrl-M enables native menus until the next keypress.  The different behavior affects
+		/// only developers.  Note that OnDomKeyDown and OnDomKeyUp don't seem to work reliably when I
+		/// tested them.
+		/// </summary>
+		override protected bool WantNativeMenu
+		{
+			get
+			{
+				return !_lastKeypressWasCtrlM && ContextMenuProvider == null;
+			}
+		}
 #endif
 
 		private void Paste()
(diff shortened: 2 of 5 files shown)


JohnThomson (changes requested):
Thanks for fixing my mess!

> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5375#-:-NBMu8dE2z6RnGFS8QCg:bhj23cv)* status: all files reviewed, 2 unresolved discussions (waiting on @StephenMcConnel)

___
*[`src/BloomExe/GeckoFxBrowser.cs` line 183 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5375#-NBMlLAJ-9mJXCWyAMxU:-NBMlLAJ-9mJXCWyAMxV:b-fm7bze) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/891cfecca3445643b31b7b63284e0093073d8cb2/src/BloomExe/GeckoFxBrowser.cs#L183)):*
> ```Smalltalk
> 		}
> 
> 		public GeckoFxBrowser(bool offScreen = false)
> ```

I'm not sure I understand what is different about an offScreen browser. For example, is the critical thing that it's not visible, that it never becomes part of a form, or what? If you can clarify that in a comment it might be helpful (here or maybe better on BrowserMaker.MakeBrowser).
But if it means any of the things I think likely, the one that gets created in PublishHelper.BrowserForPageChecks should probably call MakeBrowser(offscreen:true) also.
___
*[`src/BloomExe/GeckoFxBrowser.cs` line 549 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5375#-NBMpC938DTTBeTKDaLY:-NBMpC938DTTBeTKDaLZ:b-k6xsic) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/891cfecca3445643b31b7b63284e0093073d8cb2/src/BloomExe/GeckoFxBrowser.cs#L549)):*
> ```Smalltalk
> 				return !_controlPressed && ContextMenuProvider == null;
> 			}
> 		}
> ```

You're basically moving logic that depends on _controlPressed into GeckoFxBrowser, where it belongs. That's clearly right, but it had me looking closer at this variable, which really means something like "The most recent keypress was ctrl-M". But don't we need WantNativeMenu to be true if the ctrl key is held down during a right click? And that might not involve a keypress at all, certainly not a ctrl-m.
I presume (Control.ModifierKeys & Keys.Control) != Keys.Control (from the overridden method) just doesn't work on Linux, but replacing it with "the most recent keypress was ctrl-M" isn't the same. Is the only way to get the native menu in Linux to press ctrl-M (which does nothing else?) and then right-click? And will right-click then go on giving native menu until the next keypress? That doesn't feel right.
But then, the native menu is mainly for developers, and you are the only Linux developer, so maybe this is just something you put up with?
Seems like it would at least be good to rename _controlPressed something like _lastKeypressWasCtrlM, and comment roughly that the overridden method uses code that only works on Windows to determine whether the CTRL-key was down during a right click, and this is not equivalent but is the best we can do on Linux, producing a different behavior but one that only affects developers.

StephenMcConnel (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5375)* status: all files reviewed, 2 unresolved discussions (waiting on @JohnThomson)

___
*[`src/BloomExe/GeckoFxBrowser.cs` line 183 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5375#-NBMlLAJ-9mJXCWyAMxU:-NBNlf2f7T1-Ozg-Wqeo:brsdtvr) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/891cfecca3445643b31b7b63284e0093073d8cb2/src/BloomExe/GeckoFxBrowser.cs#L183)):*

I'm not sure I understand what is different about an offScreen browser. For example, is the critical thing that it's not visible, that it never becomes part of a form, or what? If you can clarify that in a comment it might be helpful (here or maybe better on BrowserMaker.MakeBrowser).
But if it means any of the things I think likely, the one that gets created in PublishHelper.BrowserForPageChecks should probably call MakeBrowser(offscreen:true) also.
</blockquote></details>

I'm not sure the Offscreen browser is actually needed.  It does two things: 1) doesn't wrap the browser's GTK window in a Winforms window and 2) provides a GetBitmap function.  The latter seems to have been used in the past, but the only references to it now are in a comment and in a Log message string.  As far as I can tell, this code is used in only two situations:  building page thumbnails for a SHRP Teacher's Guide (whatever that is, and which has never worked on Linux anyway) and creating thumbnails for page templates that don't have a .svg or .png thumbnail already provided.
I've revised the log message and removed the comment that talked about GetBitmap since it's obviously obsolete.  In testing, creating a new page in a template "book" invokes this code, and the browser *must* be an OffscreenGeckoFxBrowser or the program crashes.  I'm not sure I fully understand why, but I would advise against changing code anywhere else that is working.  As noted before, this is a Linux only feature, and enforced as such by #if \_\_MonoCS\_\_.
___
*[`src/BloomExe/GeckoFxBrowser.cs` line 549 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5375#-NBMpC938DTTBeTKDaLY:-NBO65Zf8nyesAvFqz8R:b-q9wpvt) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/891cfecca3445643b31b7b63284e0093073d8cb2/src/BloomExe/GeckoFxBrowser.cs#L549)):*

You're basically moving logic that depends on _controlPressed into GeckoFxBrowser, where it belongs. That's clearly right, but it had me looking closer at this variable, which really means something like "The most recent keypress was ctrl-M". But don't we need WantNativeMenu to be true if the ctrl key is held down during a right click? And that might not involve a keypress at all, certainly not a ctrl-m.
I presume (Control.ModifierKeys & Keys.Control) != Keys.Control (from the overridden method) just doesn't work on Linux, but replacing it with "the most recent keypress was ctrl-M" isn't the same. Is the only way to get the native menu in Linux to press ctrl-M (which does nothing else?) and then right-click? And will right-click then go on giving native menu until the next keypress? That doesn't feel right.
But then, the native menu is mainly for developers, and you are the only Linux developer, so maybe this is just something you put up with?
Seems like it would at least be good to rename _controlPressed something like _lastKeypressWasCtrlM, and comment roughly that the overridden method uses code that only works on Windows to determine whether the CTRL-key was down during a right click, and this is not equivalent but is the best we can do on Linux, producing a different behavior but one that only affects developers. 
</blockquote></details>

This gruesome hack was the best I could come up with on Linux.  I renamed the variable and added a comment as suggested.

JohnThomson (changes requested):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5375)* status: all files reviewed, 1 unresolved discussion (waiting on @StephenMcConnel)

___
*[`src/BloomExe/GeckoFxBrowser.cs` line 183 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5375#-NBMlLAJ-9mJXCWyAMxU:-NBP4y372r8zN_O3fVer:bly52td) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/891cfecca3445643b31b7b63284e0093073d8cb2/src/BloomExe/GeckoFxBrowser.cs#L183)):*

I'm not sure the Offscreen browser is actually needed.  It does two things: 1) doesn't wrap the browser's GTK window in a Winforms window and 2) provides a GetBitmap function.  The latter seems to have been used in the past, but the only references to it now are in a comment and in a Log message string.  As far as I can tell, this code is used in only two situations:  building page thumbnails for a SHRP Teacher's Guide (whatever that is, and which has never worked on Linux anyway) and creating thumbnails for page templates that don't have a .svg or .png thumbnail already provided.
I've revised the log message and removed the comment that talked about GetBitmap since it's obviously obsolete.  In testing, creating a new page in a template "book" invokes this code, and the browser *must* be an OffscreenGeckoFxBrowser or the program crashes.  I'm not sure I fully understand why, but I would advise against changing code anywhere else that is working.  As noted before, this is a Linux only feature, and enforced as such by #if \_\_MonoCS\_\_.
</blockquote></details>

So, would it be fair to comment on MakeBrowser something like, "Passing offscreen true means that, on Linux, we will use an OffScreenGeckoWebBrowser internally instead of the usual one. The only known case where it should be true is when the browser will be used for GetPreview (and not actually displayed). It is not necessary when the browser is used to navigate to a document and make queries about it using Javascript, even if the browser will never actually appear on the screen. We can remove this when we retire Gecko."?
I'd suggest experimenting to see whether GetPreview actually needs an offscreenGWB, but you have it implemented and working and the whole GeckoFX thing is going away in our next milestone. So the main thing is it's clear this is just for Gecko and we can drop it when cleaning up for 5.5.

StephenMcConnel (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5375)* status: 4 of 5 files reviewed, 1 unresolved discussion (waiting on @JohnThomson)

___
*[`src/BloomExe/GeckoFxBrowser.cs` line 183 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5375#-NBMlLAJ-9mJXCWyAMxU:-NBPD1N77Q8aA54wciXM:b-896fix) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/891cfecca3445643b31b7b63284e0093073d8cb2/src/BloomExe/GeckoFxBrowser.cs#L183)):*

So, would it be fair to comment on MakeBrowser something like, "Passing offscreen true means that, on Linux, we will use an OffScreenGeckoWebBrowser internally instead of the usual one. The only known case where it should be true is when the browser will be used for GetPreview (and not actually displayed). It is not necessary when the browser is used to navigate to a document and make queries about it using Javascript, even if the browser will never actually appear on the screen. We can remove this when we retire Gecko."?
I'd suggest experimenting to see whether GetPreview actually needs an offscreenGWB, but you have it implemented and working and the whole GeckoFX thing is going away in our next milestone. So the main thing is it's clear this is just for Gecko and we can drop it when cleaning up for 5.5.
</blockquote></details>

Done.

JohnThomson (approved):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5375)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @StephenMcConnel)

PR #5383, opened 2022-09-09
BL-11484 Prevent moving collection directory while Bloom is running

hatton:
This change is [Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5383)

### DIFF
--- a/src/BloomExe/ProjectContext.cs
+++ b/src/BloomExe/ProjectContext.cs
@@ -60,8 +60,8 @@ namespace Bloom
 			SettingsPath = projectSettingsPath;
 
 			// Prevent the collection folder from being moved or renamed while we're running. We can't actually lock a folder,
-			// so we do this by locking the .bloomColelctionFile. We do this in a way that allows others (and our own code)
-			// from reading and wrighting to it, just not deleting/renaming/moving it.  BL-11484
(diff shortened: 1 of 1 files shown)


src/BloomExe/ProjectContext.cs
```diff
@@ -55,6 +58,26 @@ public class ProjectContext : IDisposable
 		public ProjectContext(string projectSettingsPath, IContainer parentContainer, bool justEnoughForHtmlDialog = false)
 		{
 			SettingsPath = projectSettingsPath;
+
+			// Prevent the collection folder from being moved or renamed while we're running. We can't actually lock a folder,
+			// so we do this by locking the .bloomColelctionFile. We do this in a way that allows others (and our own code)
```
andrew-polk, line 63:
```suggestion
			// so we do this by locking the .bloomCollectionFile. We do this in a way that allows others (and our own code)
```

src/BloomExe/ProjectContext.cs
```diff
@@ -55,6 +58,26 @@ public class ProjectContext : IDisposable
...
+			// from reading and wrighting to it, just not deleting/renaming/moving it.  BL-11484
```
andrew-polk, line 64:
```suggestion
			// to read and write to it, just not delete/rename/move it.  BL-11484
```

PR #5386, opened 2022-09-13
Manual refresh of ePUB publish screen instead of automatic (BL-11043)

StephenMcConnel:
This change is [Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5386)

### DIFF
--- a/src/BloomBrowserUI/publish/ePUBPublish/ePUBPublishScreen.tsx
+++ b/src/BloomBrowserUI/publish/ePUBPublish/ePUBPublishScreen.tsx
@@ -12,7 +12,7 @@ import PublishScreenTemplate from "../commonPublish/PublishScreenTemplate";
 import { DeviceAndControls } from "../commonPublish/DeviceAndControls";
 import * as ReactDOM from "react-dom";
 import ThemeProvider from "@material-ui/styles/ThemeProvider";
-import { lightTheme } from "../../bloomMaterialUITheme";
+import { darkTheme, lightTheme } from "../../bloomMaterialUITheme";
 import { StorybookContext } from "../../.storybook/StoryBookContext";
 import {
     useSubscribeToWebSocketForStringMessage,
@@ -48,9 +48,10 @@ export const EPUBPublishScreen = () => {
 
 const EPUBPublishScreenInternal: React.FunctionComponent<{
     onReset: () => void;
-}> = () => {
+}> = props => {
     const inStorybookMode = useContext(StorybookContext);
     const [closePending, setClosePending] = useState(false);
+    const [highlightRefresh, setHighlightRefresh] = useState(false);
     const [progressState, setProgressState] = useState(ProgressState.Working);
     React.useEffect(() => hookupLinkHandler(), []);
     const [bookUrl, setBookUrl] = useState(
@@ -81,6 +82,7 @@ const EPUBPublishScreenInternal: React.FunctionComponent<{
             // add a random component so that react will reload the iframe
             setBookUrl(url + "&random=" + Math.random().toString());
             setClosePending(true);
+            setHighlightRefresh(false);
         }
     );
     const isLicenseOK = BloomApi.useWatchBooleanEvent(
@@ -92,11 +94,16 @@ const EPUBPublishScreenInternal: React.FunctionComponent<{
     const mainPanel = (
         <div className="ePUBPublishScreen">
             <PreviewPanel>
-                <DeviceAndControls
-                    defaultLandscape={landscape}
-                    canRotate={false}
-                    url={bookUrl}
-                />
+                <ThemeProvider theme={darkTheme}>
+                    <DeviceAndControls
+                        defaultLandscape={landscape}
+                        canRotate={false}
+                        url={bookUrl}
+                        showRefresh={true}
+                        highlightRefreshIcon={highlightRefresh}
+                        onRefresh={() => props.onReset()}
+                    />
+                </ThemeProvider>
                 <Typography
                     css={css`
                         color: whitesmoke;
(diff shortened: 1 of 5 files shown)


andrew-polk (changes requested):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5386)* status: all files reviewed, 1 unresolved discussion (waiting on @StephenMcConnel)

___
*[`src/BloomExe/Publish/Epub/PublishEpubApi.cs` line 162 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5386#-NBtEAYc1w84oGLkq0vT:-NBtEAYc1w84oGLkq0vU:b-3e9h03) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/35ae3b6bf23250174a4e143d7e6c794f6c7c9cc6/src/BloomExe/Publish/Epub/PublishEpubApi.cs#L162)):*
> ```Smalltalk
> 					_webSocketServer.SendString(kWebsocketContext, kWebsocketEventId_NeedRefresh, "true");
> 				else
> 					RefreshPreview(request.CurrentBook.BookInfo.PublishSettings.Epub);
> ```

I'm unsure about this. Maybe it is just a matter of naming... but it feels like more.

Is the license check useful in the case where we only send the websocket event and don't actually refresh?

Is it odd that the needRefresh = "true" case here is essentially just having the js tell c# to tell the js to do something?

Assuming those two questions don't propose a change... should the url param name be something more like `notifyUserRatherThanRefresh`? Should it just be a separate api endpoint since we aren't actually updating the preview? 
Maybe something like `checkForValidLicenseAndNotifyForRefresh` (ok, I don't like that one...)

StephenMcConnel (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5386)* status: 0 of 4 files reviewed, 1 unresolved discussion (waiting on @andrew-polk)

___
*[`src/BloomExe/Publish/Epub/PublishEpubApi.cs` line 162 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5386#-NBtEAYc1w84oGLkq0vT:-NBx4wFq9pv5vvib_z4Z:b-fts72w) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/35ae3b6bf23250174a4e143d7e6c794f6c7c9cc6/src/BloomExe/Publish/Epub/PublishEpubApi.cs#L162)):*

I'm unsure about this. Maybe it is just a matter of naming... but it feels like more.

Is the license check useful in the case where we only send the websocket event and don't actually refresh?

Is it odd that the needRefresh = "true" case here is essentially just having the js tell c# to tell the js to do something?

Assuming those two questions don't propose a change... should the url param name be something more like `notifyUserRatherThanRefresh`? Should it just be a separate api endpoint since we aren't actually updating the preview? 
Maybe something like `checkForValidLicenseAndNotifyForRefresh` (ok, I don't like that one...)
</blockquote></details>

I rewrote the change handling to all be in javascript.  I'm not sure it's all that much cleaner, but it doesn't bounce back and forth to CS.

andrew-polk (changes requested):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5386)* status: all files reviewed, 1 unresolved discussion (waiting on @StephenMcConnel)

___
*[`src/BloomBrowserUI/publish/metadata/BookMetadataDialog.tsx` line 61 at r2](https://reviewable.io/reviews/bloombooks/bloomdesktop/5386#-NBxCfEp3k-SZYyD49X9:-NBxCfEp3k-SZYyD49XA:b-cqg25c) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/43d885ce5e5ab98b5ca7d6710f4f7b82a25c214d/src/BloomBrowserUI/publish/metadata/BookMetadataDialog.tsx#L61)):*
> ```XML
> 
>     public static show(signalChangeOnClose?: () => void) {
>         BookMetadataDialog.signalChangeOnClose = signalChangeOnClose;
> ```

Is the only code remaining which doesn't set `signalChangeOnClose` for storybook? If so, how about we set it there, too, but just have an alert which states it is calling the change on close code?
Then this param becomes required and we get rid of the else in handleCloseModal.
___
*[`src/BloomExe/Publish/Epub/PublishEpubApi.cs` line 162 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5386#-NBtEAYc1w84oGLkq0vT:-NBxDBGbCDflnkigcSGa:bprmv9b) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/35ae3b6bf23250174a4e143d7e6c794f6c7c9cc6/src/BloomExe/Publish/Epub/PublishEpubApi.cs#L162)):*

I rewrote the change handling to all be in javascript.  I'm not sure it's all that much cleaner, but it doesn't bounce back and forth to CS.
</blockquote></details>

Seems cleaner to me. Thanks!

StephenMcConnel (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5386)* status: all files reviewed, 1 unresolved discussion (waiting on @andrew-polk)

___
*[`src/BloomBrowserUI/publish/metadata/BookMetadataDialog.tsx` line 61 at r2](https://reviewable.io/reviews/bloombooks/bloomdesktop/5386#-NBxCfEp3k-SZYyD49X9:-NBxQryi2ifG1HIM3TuI:b-c2hkv7) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/43d885ce5e5ab98b5ca7d6710f4f7b82a25c214d/src/BloomBrowserUI/publish/metadata/BookMetadataDialog.tsx#L61)):*

Is the only code remaining which doesn't set `signalChangeOnClose` for storybook? If so, how about we set it there, too, but just have an alert which states it is calling the change on close code?
Then this param becomes required and we get rid of the else in handleCloseModal.
</blockquote></details>

I don't see any way to make the parameter required while allowing the ensuing variable to be set to undefined, or to have a new props passed in from storybook not become required for every use.
But I have added code to storybook to have this handler just throw up an alert with an appropriate message.

andrew-polk (approved):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5386)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @StephenMcConnel)

PR #5390, opened 2022-09-14
Clean up obsolete files/folders when uploading (BL-7616)

StephenMcConnel:
This change is [Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5390)

### DIFF
--- a/src/BloomExe/Book/BookStorage.cs
+++ b/src/BloomExe/Book/BookStorage.cs
@@ -194,13 +194,38 @@ namespace Bloom.Book
 		/// code for those tasks has no business knowing about that status file. OTOH,
 		/// some of those functions don't need some of the audio files; but a generic
 		/// cleanup function like this doesn't know which ones. For now, it just deals
-		/// with TC cleanup.
+		/// with TC cleanup and with any unused copies of the placeHolder.png image file.
 		/// </summary>
 		public static List<string> LocalOnlyFiles(string folderPath)
 		{
 			var accumulator = new List<string>();
 			TeamCollection.TeamCollection.AddTCSpecificFiles(folderPath, accumulator);
+			AddUnusedPlaceholderImages(folderPath, accumulator);
 			return accumulator;
+		}
+
+		/// <summary>
+		/// Add unused copies of placeHolder.png in bookFolderPath to accumulator.
+		/// </summary>
+		private static void AddUnusedPlaceholderImages(string bookFolderPath, List<string> accumulator)
+		{
+			// See https://issues.bloomlibrary.org/youtrack/issue/BL-7616 and also
+			// https://issues.bloomlibrary.org/youtrack/issue/BL-9479 for what has
+			// happened in the past that still can use cleanup in new work.
+			var placeholders = Directory.GetFiles(bookFolderPath, "placeHolder*.png");
+			if (placeholders.Length == 0)
+				return;
+			var htmlPath = FindBookHtmlInFolder(bookFolderPath);
+			if (String.IsNullOrEmpty(htmlPath))
+				return;	// shouldn't happen, but if it does we'll surely flag it elsewhere
+			var htmlContent = RobustFile.ReadAllText(htmlPath);
+			foreach (var filepath in placeholders)
+			{
+				var filename = Path.GetFileName(filepath);
+				if (htmlContent.Contains($" src=\"{filename}\""))
+					continue;	// file is used
+				accumulator.Add(filepath);
+			}
 		}
 
 		public string PathToExistingHtml
@@ -893,13 +918,15 @@ namespace Bloom.Book
 			//Collect up all the image files in our book's directory
 			var imageFiles = new List<string>();
 			var imageExtentions = new HashSet<string>(new []{ ".jpg", ".png", ".svg" });
-			var ignoredFilenameStarts = new HashSet<string>(new [] { "thumbnail", "placeholder", "license", "video-placeholder", "coverImage200", "widget-placeholder" });
+			var ignoredFilenameStarts = new HashSet<string>(new [] { "thumbnail", "license", "video-placeholder", "coverImage200", "widget-placeholder" });
 			foreach (var path in Directory.EnumerateFiles(FolderPath).Where(
 				s => imageExtentions.Contains(Path.GetExtension(s).ToLowerInvariant())))
 			{
 				var filename = Path.GetFileName(path);
 				if (ignoredFilenameStarts.Any(s=>filename.StartsWith(s, StringComparison.InvariantCultureIgnoreCase)))
 					continue;
+				if (filename.ToLowerInvariant() == "placeholder.png")
+					continue;	// delete unused copies, but keep base placeholder image file even if unused at the moment
 				imageFiles.Add(Path.GetFileName(GetNormalizedPathForOS(path)));
 			}
 			//Remove from that list each image actually in use
--- a/src/BloomExe/WebLibraryIntegration/BloomS3Client.cs
+++ b/src/BloomExe/WebLibraryIntegration/BloomS3Client.cs
@@ -245,6 +245,8 @@ namespace Bloom.WebLibraryIntegration
 
 			BookStorage.EnsureSingleHtmFile(destDirName);
 
+			RemoveUnwantedSubdirectories(destDirName);
+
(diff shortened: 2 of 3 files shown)


andrew-polk (changes requested):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5390)* status: all files reviewed, 1 unresolved discussion (waiting on @StephenMcConnel)

___
*[`src/BloomExe/Book/BookStorage.cs` line 206 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5390#-NByB734899fHc7bXzTA:-NByB735Bz5I7mU3uAgX:b-iyvqpf) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/1a8e60e176422167cc217d130f7ab7610dd29858/src/BloomExe/Book/BookStorage.cs#L206)):*
> ```Smalltalk
> 			// https://issues.bloomlibrary.org/youtrack/issue/BL-9479 for what has
> 			// happened in the past that still can use cleanup in new work.
> 			accumulator.AddRange(Directory.GetFiles(folderPath, "placeHolder*.png"));
> ```

Does this mean we never upload any placeHolder files?
That doesn't see right, either. 
It seems like if the book contains a placeholder image at the time of upload, we still want to include it.

StephenMcConnel (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5390)* status: 2 of 3 files reviewed, 1 unresolved discussion (waiting on @andrew-polk)

___
*[`src/BloomExe/Book/BookStorage.cs` line 206 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5390#-NByB734899fHc7bXzTA:-NC1JDFr3kJ0C8l4vhEd:bwzpxrp) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/1a8e60e176422167cc217d130f7ab7610dd29858/src/BloomExe/Book/BookStorage.cs#L206)):*

Does this mean we never upload any placeHolder files?
That doesn't see right, either. 
It seems like if the book contains a placeholder image at the time of upload, we still want to include it.
</blockquote></details>

I've changed the code to upload placeHolder.png files that are actually referenced in the book's html file,
but to delete any other leftover files that are not referenced by the html.

andrew-polk (changes requested):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5390)* status: all files reviewed, 2 unresolved discussions (waiting on @StephenMcConnel)

___
*[`src/BloomExe/Book/BookStorage.cs` line 216 at r2](https://reviewable.io/reviews/bloombooks/bloomdesktop/5390#-NC668cI5JqENhyo8FaS:-NC668cI5JqENhyo8FaT:b42xxov) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/81d2431a88155b329742eb383a684d5a6d621c7e/src/BloomExe/Book/BookStorage.cs#L216)):*
> ```Smalltalk
> 			// happened in the past that still can use cleanup in new work.
> 			var bookName = Path.GetFileName(bookFolderPath);
> 			var htmlPath = Path.Combine(bookFolderPath, bookName+".htm");
> ```

Can we rely on the book file name being the same as the folder name? (I can't remember how the rename stuff works.)
___
*[`src/BloomExe/Book/BookStorage.cs` line 227 at r2](https://reviewable.io/reviews/bloombooks/bloomdesktop/5390#-NC66M5D3ZGJtREtX-x3:-NC66M5D3ZGJtREtX-x4:bf6v2gc) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/81d2431a88155b329742eb383a684d5a6d621c7e/src/BloomExe/Book/BookStorage.cs#L227)):*
> ```Smalltalk
> 					accumulator.Add(filepath);
> 				}
> 			}
> ```

If I understand the context correctly, we are just removing the obsolete placeholder files from the staged book which will be uploaded. Should we instead remove them from the real book, perhaps in BringBookUpToDate?

StephenMcConnel (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5390)* status: 2 of 3 files reviewed, 2 unresolved discussions (waiting on @andrew-polk)

___
*[`src/BloomExe/Book/BookStorage.cs` line 216 at r2](https://reviewable.io/reviews/bloombooks/bloomdesktop/5390#-NC668cI5JqENhyo8FaS:-NCLgR9u922-0dd8Wbm8:b-87mjdx) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/81d2431a88155b329742eb383a684d5a6d621c7e/src/BloomExe/Book/BookStorage.cs#L216)):*

Can we rely on the book file name being the same as the folder name? (I can't remember how the rename stuff works.)
</blockquote></details>

FindBookHtmlInFolder(folder) is probably better. What I had here is actually the first candidate found by that method.
___
*[`src/BloomExe/Book/BookStorage.cs` line 227 at r2](https://reviewable.io/reviews/bloombooks/bloomdesktop/5390#-NC66M5D3ZGJtREtX-x3:-NCLigziBxr-jCNCv7K8:b-70e92e) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/81d2431a88155b329742eb383a684d5a6d621c7e/src/BloomExe/Book/BookStorage.cs#L227)):*

If I understand the context correctly, we are just removing the obsolete placeholder files from the staged book which will be uploaded. Should we instead remove them from the real book, perhaps in BringBookUpToDate?
</blockquote></details>

I've added code to CleanupUnusedImageFiles (called by ExpensiveInitialization) to delete unused copies of placeHolder.png, but not the base image itself.  But I think leaving the code here is safe, and may help in rare cases where a user goes directly to Publish from Collections on an old book.

andrew-polk (changes requested):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5390)* status: all files reviewed, 1 unresolved discussion (waiting on @StephenMcConnel)

___
*[`src/BloomExe/Book/BookStorage.cs` line 227 at r2](https://reviewable.io/reviews/bloombooks/bloomdesktop/5390#-NC66M5D3ZGJtREtX-x3:-NCLq13i1MOB8wTrjeOd:bequ2bv) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/81d2431a88155b329742eb383a684d5a6d621c7e/src/BloomExe/Book/BookStorage.cs#L227)):*

I've added code to CleanupUnusedImageFiles (called by ExpensiveInitialization) to delete unused copies of placeHolder.png, but not the base image itself.  But I think leaving the code here is safe, and may help in rare cases where a user goes directly to Publish from Collections on an old book.
</blockquote></details>

Ok, but I think we can optimize a little.
Looks like we don't need to check if the file exists after calling `FindBookHtmlInFolder` as that will return null if it doesn't.
We could check if there are any placeHolder images in the book folder before reading the html file.
Indeed we can wait to find the html file until after that, too, since we always expect to find one.
(And finding a placeholder file should be the unusual case? Or am I wrong and we usually get here with the base one?)

StephenMcConnel (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5390)* status: 2 of 3 files reviewed, 1 unresolved discussion (waiting on @andrew-polk)

___
*[`src/BloomExe/Book/BookStorage.cs` line 227 at r2](https://reviewable.io/reviews/bloombooks/bloomdesktop/5390#-NC66M5D3ZGJtREtX-x3:-NCM1K9L6AEq2ChEhzBX:b-896fix) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/81d2431a88155b329742eb383a684d5a6d621c7e/src/BloomExe/Book/BookStorage.cs#L227)):*

Ok, but I think we can optimize a little.
Looks like we don't need to check if the file exists after calling `FindBookHtmlInFolder` as that will return null if it doesn't.
We could check if there are any placeHolder images in the book folder before reading the html file.
Indeed we can wait to find the html file until after that, too, since we always expect to find one.
(And finding a placeholder file should be the unusual case? Or am I wrong and we usually get here with the base one?)
</blockquote></details>

Done.

andrew-polk (approved):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5390)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @StephenMcConnel)

PR #5394, opened 2022-09-15
Fix a display glitch when refreshing the ePUB publish screen (BL-11043)

StephenMcConnel:


sujeffreyl (changes requested):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5394)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @StephenMcConnel and @sujeffreyl)

___
*[`src/BloomBrowserUI/publish/ePUBPublish/ePUBSettingsGroup.tsx` line 43 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5394#-NC5jN307efmExwIYaN8:-NC5jN3290M6asBcHHWc:bn1s9mg) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/4e9bc21cd06df26ba5ea727655abc554dfba7c5b/src/BloomBrowserUI/publish/ePUBPublish/ePUBSettingsGroup.tsx#L43)):*
> ```XML
> 
> // If we don't adjust this to the current mode, there's a display glitch when refreshing.
> let defaultEpubMode = "fixed";
> ```

Is there any particular reason we need to define it here instead of as a state/effect inside the React function? I think it could go inside. Was reading https://stackoverflow.com/questions/68644968/is-it-an-anti-pattern-to-use-variables-the-outside-of-react-function-component and I think it's more normal to be there?
Thoughts?

andrew-polk (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5394)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @andrew-polk and @sujeffreyl)

___
*[`src/BloomBrowserUI/publish/ePUBPublish/ePUBSettingsGroup.tsx` line 43 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5394#-NC5jN307efmExwIYaN8:-NCLicGR6MxBejWU8Mid:bzgnvyn) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/4e9bc21cd06df26ba5ea727655abc554dfba7c5b/src/BloomBrowserUI/publish/ePUBPublish/ePUBSettingsGroup.tsx#L43)):*

useRef seems to be function to use for maintaining state, but I can't get it work in our code.  It appears that the refresh button instantiates a new copy of the function, causing a new useRef variable to be created, although it didn't create a new copy of the outer variable.  In this situation, I think the anti-pattern is also the only (or at least simplest) workable pattern.
</blockquote></details>

If I understand, this would indicate a glitch in our server interaction.
Don't we only want to ask the server one time for the initial value and then we know what it is locally thereafter?
I haven't looked into the guts of `useApiStringState` but maybe it needs another param or an alternate version?

Or am I just totally missing what is going on?

My assumptions:
defaultEpubMode is only used to display the the UI until we get the result of the server call.
That should only ever matter on initial load.
So having to modify the default should never be needed.

StephenMcConnel (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5394)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @andrew-polk and @sujeffreyl)

___
*[`src/BloomBrowserUI/publish/ePUBPublish/ePUBSettingsGroup.tsx` line 43 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5394#-NC5jN307efmExwIYaN8:-NCLt3221yWU-cQz8uug:bvwlsvh) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/4e9bc21cd06df26ba5ea727655abc554dfba7c5b/src/BloomBrowserUI/publish/ePUBPublish/ePUBSettingsGroup.tsx#L43)):*

If I understand, this would indicate a glitch in our server interaction.
Don't we only want to ask the server one time for the initial value and then we know what it is locally thereafter?
I haven't looked into the guts of `useApiStringState` but maybe it needs another param or an alternate version?

Or am I just totally missing what is going on?

My assumptions:
defaultEpubMode is only used to display the the UI until we get the result of the server call.
That should only ever matter on initial load.
So having to modify the default should never be needed.

</blockquote></details>

Being a ReactFunction, this code gets called every time anything wants to trigger a render of the epub settings. This happens much more often than one would expect.  Calling the useApiStringState and useApiBoolean methods should ideally need to happen only once, but I don't think that's possible with ReactFunction designs.  Here's what I traced for calls to this ReactFunction:

**initial display**
FROM JS: 1/ DEBUG EPUBSettingsGroup: defaultEpubMode=fixed
FROM JS: 2/ DEBUG EPUBSettingsGroup: defaultEpubMode=fixed
FROM JS: 3/ DEBUG EPUBSettingsGroup: defaultEpubMode=fixed
FROM JS: 4/ DEBUG EPUBSettingsGroup: defaultEpubMode=fixed
FROM JS: 5/ DEBUG EPUBSettingsGroup: defaultEpubMode=fixed
FROM JS: 6/ DEBUG EPUBSettingsGroup: defaultEpubMode=fixed
FROM JS: 7/ DEBUG EPUBSettingsGroup: defaultEpubMode=fixed
FROM JS: 8/ DEBUG EPUBSettingsGroup: defaultEpubMode=fixed
FROM JS: 9/ DEBUG EPUBSettingsGroup: defaultEpubMode=fixed

**open mode dropdown, select flowable**
FROM JS: 10/ DEBUG EPUBSettingsGroup: defaultEpubMode=fixed
FROM JS: 11/ DEBUG EPUBSettingsGroup: defaultEpubMode=fixed
FROM JS: 12/ DEBUG EPUBSettingsGroup: defaultEpubMode=fixed
FROM JS: 13/ DEBUG EPUBSettingsGroup: defaultEpubMode=flowable

**click on Refresh**
FROM JS: 14/ DEBUG EPUBSettingsGroup: defaultEpubMode=flowable
FROM JS: 15/ DEBUG EPUBSettingsGroup: defaultEpubMode=flowable
FROM JS: 16/ DEBUG EPUBSettingsGroup: defaultEpubMode=flowable
FROM JS: 17/ DEBUG EPUBSettingsGroup: defaultEpubMode=flowable
FROM JS: 18/ DEBUG EPUBSettingsGroup: defaultEpubMode=flowable
FROM JS: 19/ DEBUG EPUBSettingsGroup: defaultEpubMode=flowable
FROM JS: 20/ DEBUG EPUBSettingsGroup: defaultEpubMode=flowable
FROM JS: 21/ DEBUG EPUBSettingsGroup: defaultEpubMode=flowable
FROM JS: 22/ DEBUG EPUBSettingsGroup: defaultEpubMode=flowable

andrew-polk (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5394)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @andrew-polk and @sujeffreyl)

___
*[`src/BloomBrowserUI/publish/ePUBPublish/ePUBSettingsGroup.tsx` line 43 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5394#-NC5jN307efmExwIYaN8:-NCMzTeq4vOOub0m0aFy:b8298kd) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/4e9bc21cd06df26ba5ea727655abc554dfba7c5b/src/BloomBrowserUI/publish/ePUBPublish/ePUBSettingsGroup.tsx#L43)):*

Being a ReactFunction, this code gets called every time anything wants to trigger a render of the epub settings. This happens much more often than one would expect.  Calling the useApiStringState and useApiBoolean methods should ideally need to happen only once, but I don't think that's possible with ReactFunction designs.  Here's what I traced for calls to this ReactFunction:

**initial display**
FROM JS: 1/ DEBUG EPUBSettingsGroup: defaultEpubMode=fixed
FROM JS: 2/ DEBUG EPUBSettingsGroup: defaultEpubMode=fixed
FROM JS: 3/ DEBUG EPUBSettingsGroup: defaultEpubMode=fixed
FROM JS: 4/ DEBUG EPUBSettingsGroup: defaultEpubMode=fixed
FROM JS: 5/ DEBUG EPUBSettingsGroup: defaultEpubMode=fixed
FROM JS: 6/ DEBUG EPUBSettingsGroup: defaultEpubMode=fixed
FROM JS: 7/ DEBUG EPUBSettingsGroup: defaultEpubMode=fixed
FROM JS: 8/ DEBUG EPUBSettingsGroup: defaultEpubMode=fixed
FROM JS: 9/ DEBUG EPUBSettingsGroup: defaultEpubMode=fixed

**open mode dropdown, select flowable**
FROM JS: 10/ DEBUG EPUBSettingsGroup: defaultEpubMode=fixed
FROM JS: 11/ DEBUG EPUBSettingsGroup: defaultEpubMode=fixed
FROM JS: 12/ DEBUG EPUBSettingsGroup: defaultEpubMode=fixed
FROM JS: 13/ DEBUG EPUBSettingsGroup: defaultEpubMode=flowable

**click on Refresh**
FROM JS: 14/ DEBUG EPUBSettingsGroup: defaultEpubMode=flowable
FROM JS: 15/ DEBUG EPUBSettingsGroup: defaultEpubMode=flowable
FROM JS: 16/ DEBUG EPUBSettingsGroup: defaultEpubMode=flowable
FROM JS: 17/ DEBUG EPUBSettingsGroup: defaultEpubMode=flowable
FROM JS: 18/ DEBUG EPUBSettingsGroup: defaultEpubMode=flowable
FROM JS: 19/ DEBUG EPUBSettingsGroup: defaultEpubMode=flowable
FROM JS: 20/ DEBUG EPUBSettingsGroup: defaultEpubMode=flowable
FROM JS: 21/ DEBUG EPUBSettingsGroup: defaultEpubMode=flowable
FROM JS: 22/ DEBUG EPUBSettingsGroup: defaultEpubMode=flowable


</blockquote></details>

How many times does the server receive the request?
`useApiStringState` does seem to be coded to only ask once. 

But I guess the problem is, as you stated in the comment above, this component is actually getting completely recreated, not just rerendered. Let's run this by the group to see if anyone has an insight.

andrew-polk (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5394)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @andrew-polk and @sujeffreyl)

___
*[`src/BloomBrowserUI/publish/ePUBPublish/ePUBSettingsGroup.tsx` line 43 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5394#-NC5jN307efmExwIYaN8:-NCN278G2L_Al3fDsIeU:bhtbpdk) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/4e9bc21cd06df26ba5ea727655abc554dfba7c5b/src/BloomBrowserUI/publish/ePUBPublish/ePUBSettingsGroup.tsx#L43)):*

How many times does the server receive the request?
`useApiStringState` does seem to be coded to only ask once. 

But I guess the problem is, as you stated in the comment above, this component is actually getting completely recreated, not just rerendered. Let's run this by the group to see if anyone has an insight.
</blockquote></details>

Hm. Taking a look at the refresh code, why do we rerender the entire screen? Shouldn't we only rerender the preview?

sujeffreyl (changes requested):
> Reviewed 1 of 1 files at r2.
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5394)* status: all files reviewed (commit messages unreviewed), 1 unresolved discussion (waiting on @andrew-polk and @StephenMcConnel)

___
*[`src/BloomBrowserUI/publish/ePUBPublish/ePUBSettingsGroup.tsx` line 43 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5394#-NC5jN307efmExwIYaN8:-NCQReHI2oexvbBGmll3:blhf9zo) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/4e9bc21cd06df26ba5ea727655abc554dfba7c5b/src/BloomBrowserUI/publish/ePUBPublish/ePUBSettingsGroup.tsx#L43)):*

Hm. Taking a look at the refresh code, why do we rerender the entire screen? Shouldn't we only rerender the preview?
</blockquote></details>

Ok, I debugged into this. The component is getting recreated by design. In "ePUBPublishScreen.tsx", the "EPUBPublishScreen" component implements the refresh button by incrementing keyForReset and then SETTING key = {keyForReset}. This "key" is the special parameter that React uses to determine whether two components are the same (in lists, normally I guess). We seem to be taking advantage of this to get a brand new component and start everything off from the beginning.  So yeah, that would definitely explain why the useApiStringState is looking at the initial value again (which suprised me) and explains the rest of Steve's comments.

I tried changing renaming key to something else. not much happens, which I guess makes sense. We use the same component, and since it was just a dummy key, I guess not too much (if anything) needs to re-render.  So, it seems to me that it's definitely by current design that we use 'key' to get a brand new component.

Actually, I question the whole point of this card. It seems to me, according to the code, it is by design of the code. (Whether it's by design of the requirements are murkier...)
The code comments there say "// need to rebuild the book and re-run all of our Bloom API queries.
    // This requires a hard-reset of the whole screen, which we do by
    // incrementing a `key` prop on the core of this screen."

If we're planning on starting everything from scratch, it makes sense that we start off with the default constant, and the when the GET request finishes, we update it with the current value.

(BTW, I don't think what i repro is 100% the same as Colin described? What I see is that when I hit refresh, it briefly changes to Fixed and then changes back to Flowable quickly. Colin's description made it sound like it stays on Fixed until the Rebuild is complete. Steve, what did you see?)


So, with this current card, it sounds like we're asking for "When you hit the refresh button, re-build everything completely from scratch... except for the epub mode dropdown, don't start that one from scratch, but re-use the state."  From the code point of view, it sounds weird to me.

TL;DR: I vote for "Won't Fix - By Design"

sujeffreyl (changes requested):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5394)* status: all files reviewed (commit messages unreviewed), 1 unresolved discussion (waiting on @andrew-polk and @StephenMcConnel)

___
*[`src/BloomBrowserUI/publish/ePUBPublish/ePUBSettingsGroup.tsx` line 43 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5394#-NC5jN307efmExwIYaN8:-NCVTOOD87wMsZNLl4sw:b-ge21xq) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/4e9bc21cd06df26ba5ea727655abc554dfba7c5b/src/BloomBrowserUI/publish/ePUBPublish/ePUBSettingsGroup.tsx#L43)):*

Ok, I debugged into this. The component is getting recreated by design. In "ePUBPublishScreen.tsx", the "EPUBPublishScreen" component implements the refresh button by incrementing keyForReset and then SETTING key = {keyForReset}. This "key" is the special parameter that React uses to determine whether two components are the same (in lists, normally I guess). We seem to be taking advantage of this to get a brand new component and start everything off from the beginning.  So yeah, that would definitely explain why the useApiStringState is looking at the initial value again (which suprised me) and explains the rest of Steve's comments.

I tried changing renaming key to something else. not much happens, which I guess makes sense. We use the same component, and since it was just a dummy key, I guess not too much (if anything) needs to re-render.  So, it seems to me that it's definitely by current design that we use 'key' to get a brand new component.

Actually, I question the whole point of this card. It seems to me, according to the code, it is by design of the code. (Whether it's by design of the requirements are murkier...)
The code comments there say "// need to rebuild the book and re-run all of our Bloom API queries.
    // This requires a hard-reset of the whole screen, which we do by
    // incrementing a `key` prop on the core of this screen."

If we're planning on starting everything from scratch, it makes sense that we start off with the default constant, and the when the GET request finishes, we update it with the current value.

(BTW, I don't think what i repro is 100% the same as Colin described? What I see is that when I hit refresh, it briefly changes to Fixed and then changes back to Flowable quickly. Colin's description made it sound like it stays on Fixed until the Rebuild is complete. Steve, what did you see?)


So, with this current card, it sounds like we're asking for "When you hit the refresh button, re-build everything completely from scratch... except for the epub mode dropdown, don't start that one from scratch, but re-use the state."  From the code point of view, it sounds weird to me.

TL;DR: I vote for "Won't Fix - By Design"
</blockquote></details>

Or we could have the initial state say "Loading..." or blank or something like that

andrew-polk (changes requested):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5394)* status: all files reviewed, 1 unresolved discussion (waiting on @sujeffreyl)

___
*[`src/BloomBrowserUI/publish/ePUBPublish/ePUBSettingsGroup.tsx` line 43 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5394#-NC5jN307efmExwIYaN8:-NCWO04x9BdWCN7tk7UD:bfwuwdr) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/4e9bc21cd06df26ba5ea727655abc554dfba7c5b/src/BloomBrowserUI/publish/ePUBPublish/ePUBSettingsGroup.tsx#L43)):*

Or we could have the initial state say "Loading..." or blank or something like that
</blockquote></details>

Is there any reason we need to refresh more than just the preview window?
Anyway, after seeing the glitch myself, I also saw only a brief flash. I would also be inclined to say it isn't worth fixing. Or having it blank would be fine, if that was easy.

I also wouldn't be greatly upset with what you have here if we comment a bit more. Specifically we would need to say that we are purposefully refreshing the parent which causes new children to be created. And we think we could have it only refresh the preview but it isn't worth refactoring at that level for this minor thing. And probably say something about how this won't work if we ever reuse the component.

StephenMcConnel (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5394)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @andrew-polk and @sujeffreyl)

___
*[`src/BloomBrowserUI/publish/ePUBPublish/ePUBSettingsGroup.tsx` line 43 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5394#-NC5jN307efmExwIYaN8:-NCWwBDMEq62p2S0MuN4:b8egdig) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/4e9bc21cd06df26ba5ea727655abc554dfba7c5b/src/BloomBrowserUI/publish/ePUBPublish/ePUBSettingsGroup.tsx#L43)):*

Is there any reason we need to refresh more than just the preview window?
Anyway, after seeing the glitch myself, I also saw only a brief flash. I would also be inclined to say it isn't worth fixing. Or having it blank would be fine, if that was easy.

I also wouldn't be greatly upset with what you have here if we comment a bit more. Specifically we would need to say that we are purposefully refreshing the parent which causes new children to be created. And we think we could have it only refresh the preview but it isn't worth refactoring at that level for this minor thing. And probably say something about how this won't work if we ever reuse the component.
</blockquote></details>

If the design doesn't work properly, then there's a design bug, not something we can just shrug at.  That being said, I think I've achieved a reasonable solution by passing props up and down a couple of levels.  Refresh causes a blink in the display (which is unavoidable) but always displays the correct value for the epub mode with this latest change.

andrew-polk (approved):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5394)* status: all files reviewed, 2 unresolved discussions (waiting on @StephenMcConnel and @sujeffreyl)

___
*[`src/BloomBrowserUI/publish/ePUBPublish/ePUBPublishScreen.tsx` line 47 at r3](https://reviewable.io/reviews/bloombooks/bloomdesktop/5394#-NCWzQAF7spEy6Fa8m9Y:-NCWzQAF7spEy6Fa8m9Z:bx4ojn9) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/ac4fc1653b0961358d364984ce6be4b87beee9cf/src/BloomBrowserUI/publish/ePUBPublish/ePUBPublishScreen.tsx#L47)):*
> ```XML
>     // not destroyed during refresh.  See comments toward the end of
>     // https://issues.bloomlibrary.org/youtrack/issue/BL-11043.
>     const modeEpub = React.useRef("fixed");
> ```

If you're making changes, I think I would make this defaultEpubModeRef or similar.
But if J approves, go ahead and merge.



_Code quote:_
```XML
modeEpub
```

StephenMcConnel (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5394)* status: 1 of 2 files reviewed, 1 unresolved discussion (waiting on @andrew-polk and @sujeffreyl)

___
*[`src/BloomBrowserUI/publish/ePUBPublish/ePUBPublishScreen.tsx` line 47 at r3](https://reviewable.io/reviews/bloombooks/bloomdesktop/5394#-NCWzQAF7spEy6Fa8m9Y:-NCX6bpvCWiTDFTryH5O:b-896fix) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/ac4fc1653b0961358d364984ce6be4b87beee9cf/src/BloomBrowserUI/publish/ePUBPublish/ePUBPublishScreen.tsx#L47)):*

If you're making changes, I think I would make this defaultEpubModeRef or similar.
But if J approves, go ahead and merge.
</blockquote></details>

Done.

andrew-polk (approved):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5394)* status: all files reviewed, 1 unresolved discussion (waiting on @sujeffreyl)

PR #5402, opened 2022-09-21
On SS import, keep one editable from original as a template (BL-11483)

JohnThomson:
This change is [Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5402)

### DIFF
--- a/src/BloomExe/Book/TranslationGroupManager.cs
+++ b/src/BloomExe/Book/TranslationGroupManager.cs
@@ -521,7 +521,7 @@ namespace Bloom.Book
 			if (prototype == null) //this was an empty translation-group (unusual, but we can cope)
 			{
 				newElementInThisLanguage = groupElement.OwnerDocument.CreateElement("div");
-				newElementInThisLanguage.SetAttribute("class", "bloom-editable");
+				newElementInThisLanguage.SetAttribute("class", "bloom-editable normal-style");
 				newElementInThisLanguage.SetAttribute("contenteditable", "true");
 				if (groupElement.HasAttribute("data-placeholder"))
 				{
--- a/src/BloomExe/Edit/WebThumbNailList.cs
+++ b/src/BloomExe/Edit/WebThumbNailList.cs
@@ -281,7 +281,7 @@ namespace Bloom.Edit
 			_browser.DocumentCompleted += WebBrowser_DocumentCompleted;
 
 			_baseForRelativePaths = pageListDom.BaseForRelativePaths;
-			_browser.Navigate(pageListDom, source:BloomServer.SimulatedPageFileSource.Pagelist);
+			Invoke((Action) (() => _browser.Navigate(pageListDom, source:BloomServer.SimulatedPageFileSource.Pagelist)));
 			return result.ToList();
 		}
 
--- a/src/BloomExe/Spreadsheet/SpreadsheetImporter.cs
+++ b/src/BloomExe/Spreadsheet/SpreadsheetImporter.cs
@@ -551,7 +551,16 @@ namespace Bloom.Spreadsheet
 			var editables = newPage.SelectNodes(".//div[contains(@class, 'bloom-editable') and @lang != 'z']").Cast<XmlElement>().ToArray();
 			foreach (var e in editables)
 			{
-				e.ParentNode.RemoveChild(e);
+				var allInGroup = editables.Where(x => x.ParentNode == e.ParentNode && x != e);
+				if (allInGroup.Any())
+					e.ParentNode.RemoveChild(e);
+				else
+				{
+					// The only thing in the group is an element with a language other than 'z'.
+					// We want to keep this, but only as a template (e.g., to preserve the user style).
+					e.SetAttribute("lang", "z");
+					e.InnerText = "";
+				}
 			}
 
 			var imageContainers = GetImageContainers(newPage);
--- a/src/BloomTests/Book/TranslationGroupManagerTests.cs
+++ b/src/BloomTests/Book/TranslationGroupManagerTests.cs
@@ -108,7 +108,7 @@ namespace BloomTests.Book
 		[Test]
 		public void PrepareElementsOnPage_HasEmptyTranslationGroup_MakesVernacularAndNational()
 		{
-			var contents = @"<html><body><div class='bloom-page bloom-translationGroup normal-style'>
+			var contents = @"<html><body><div class='bloom-page bloom-translationGroup'>
 					</div></body></html>";
 			var dom = new XmlDocument();
 			dom.LoadXml(contents);
--- a/src/BloomTests/Spreadsheet/SpreadsheetImporterTests.cs
+++ b/src/BloomTests/Spreadsheet/SpreadsheetImporterTests.cs
@@ -934,10 +934,82 @@ namespace BloomTests.Spreadsheet
 	}
 
 	/// <summary>
-	/// There are some special cases we can only test when the last page of a book has room
-	/// for more than one image and text.
+	/// There are some special cases we can only test with a DOM in which the last content page
+	/// has neither text nor image elements.
 	/// </summary>
-	public class SpreadsheetImageAndTextImportToBookWithComplexLastPageTests
+	public class SpreadsheetImportWithNoZLanguageTests
+	{
+		private HtmlDom _dom;
+
+		private TemporaryFolder _bookFolder;
+		private List<XmlElement> _contentPages;
+		private string _spreadsheetFolder;
+
+		[OneTimeSetUp]
+		public void OneTimeSetUp()
+		{
+			// We will re-use the images from another test.
+			// Conveniently the images are in a folder called "images" which is what the importer expects.
+			// So we give it the parent directory of that images folder.
+			_spreadsheetFolder =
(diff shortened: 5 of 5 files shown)


hatton (changes requested):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5402)* status: 0 of 5 files reviewed, 1 unresolved discussion (waiting on @hatton and @JohnThomson)

___
*[`src/BloomExe/Book/TranslationGroupManager.cs` line 524 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5402#-NCWmj0fEVvVrd3mjF4q:-NCWmj0fEVvVrd3mjF4r:bqhimc3) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/8b6dc2df6a55aa98836baf0e5cdca8e60a50f4d0/src/BloomExe/Book/TranslationGroupManager.cs#L524)):*
> ```Smalltalk
> 			{
> 				newElementInThisLanguage = groupElement.OwnerDocument.CreateElement("div");
> 				newElementInThisLanguage.SetAttribute("class", "bloom-editable normal-style");
> ```

Normally, bubbles get "Bubble". Is that relevant to this task?

JohnThomson (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5402)* status: 0 of 5 files reviewed, 1 unresolved discussion (waiting on @hatton)

___
*[`src/BloomExe/Book/TranslationGroupManager.cs` line 524 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5402#-NCWmj0fEVvVrd3mjF4q:-NCWqKqeAi9jLD1kHVH-:b-58l4a2) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/8b6dc2df6a55aa98836baf0e5cdca8e60a50f4d0/src/BloomExe/Book/TranslationGroupManager.cs#L524)):*

Normally, bubbles get "Bubble". Is that relevant to this task?
</blockquote></details>

It's relevant to the task, but not to this part of the fix. A different part of the fix will preserve the Bubble style (or any other) on the relevant part of the page that the importer is duplicating as a template (at least until/unless we decide we should export and import the style of a block). This is an extra patch to make sure that if we ever again get in a pathological situation where there is no prototype bloom-editable to copy, the one we create from scratch has _some_ style. (Apparently if it doesn't, our cog is not offered, so there's no way to repair things.)

hatton (approved):
> Reviewed 5 of 5 files at r1.
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5402)* status: all files reviewed, 2 unresolved discussions (waiting on @JohnThomson)

___
*[`src/BloomExe/Book/TranslationGroupManager.cs` line 524 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5402#-NCWmj0fEVvVrd3mjF4q:-NCWs4FGCg9xWIN5X9E1:bd5l6mr) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/8b6dc2df6a55aa98836baf0e5cdca8e60a50f4d0/src/BloomExe/Book/TranslationGroupManager.cs#L524)):*

It's relevant to the task, but not to this part of the fix. A different part of the fix will preserve the Bubble style (or any other) on the relevant part of the page that the importer is duplicating as a template (at least until/unless we decide we should export and import the style of a block). This is an extra patch to make sure that if we ever again get in a pathological situation where there is no prototype bloom-editable to copy, the one we create from scratch has _some_ style. (Apparently if it doesn't, our cog is not offered, so there's no way to repair things.)
</blockquote></details>

thanks!
___
*[`src/BloomExe/Edit/WebThumbNailList.cs` line 283 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5402#-NCWsANl06rXQEdT6y0T:-NCWsANl06rXQEdT6y0U:b-s74llv) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/8b6dc2df6a55aa98836baf0e5cdca8e60a50f4d0/src/BloomExe/Edit/WebThumbNailList.cs#L283)):*
> ```Smalltalk
> 			_browser.DocumentCompleted += WebBrowser_DocumentCompleted;
> 
> 			_baseForRelativePaths = pageListDom.BaseForRelativePaths;
> ```

Would it make sense to roll this into `_browser`? like `_browser.NavigateAsync` and then maybe make`Navigate` private?

JohnThomson (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5402)* status: all files reviewed, 2 unresolved discussions (waiting on @hatton and @JohnThomson)

___
*[`src/BloomExe/Edit/WebThumbNailList.cs` line 283 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5402#-NCWsANl06rXQEdT6y0T:-NCX8VuE8BGuz-RdtomP:b-20nw7y) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/8b6dc2df6a55aa98836baf0e5cdca8e60a50f4d0/src/BloomExe/Edit/WebThumbNailList.cs#L283)):*

Would it make sense to roll this into `_browser`? like `_browser.NavigateAsync` and then maybe make`Navigate` private?
</blockquote></details>

The thought of making Navigate async fills me with dread. That's what WebView2 wants you to do, but it will be viral through a whole lot of our code base: every method that navigates, every method that calls one of those methods, recursively.
We COULD try to make our Navigate thread-safe (which is what this change is about), but I'm a little nervous. It could presumably use itself as the control to invoke on, but that only works (IIRC) if it has a handle (somewhere after it's added to a form and the form is loaded), so the apparent safety might be  misleading., since we tend to want to navigate a browser soon after creating it.

hatton (approved):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5402)* status: all files reviewed, 2 unresolved discussions (waiting on @JohnThomson)

___
*[`src/BloomExe/Edit/WebThumbNailList.cs` line 283 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5402#-NCWsANl06rXQEdT6y0T:-NCXDfQXAb39txnxv0OH:bwt8fe0) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/8b6dc2df6a55aa98836baf0e5cdca8e60a50f4d0/src/BloomExe/Edit/WebThumbNailList.cs#L283)):*

The thought of making Navigate async fills me with dread. That's what WebView2 wants you to do, but it will be viral through a whole lot of our code base: every method that navigates, every method that calls one of those methods, recursively.
We COULD try to make our Navigate thread-safe (which is what this change is about), but I'm a little nervous. It could presumably use itself as the control to invoke on, but that only works (IIRC) if it has a handle (somewhere after it's added to a form and the form is loaded), so the apparent safety might be  misleading., since we tend to want to navigate a browser soon after creating it.
</blockquote></details>

Sorry, I just mean, if the right way to Navigate is to `Invoke((Action) (() => `, why not put that in the class? Why leave it to each caller to figure out? Maybe that doesn't make sense, just askin'

JohnThomson (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5402)* status: all files reviewed, 2 unresolved discussions (waiting on @hatton and @JohnThomson)

___
*[`src/BloomExe/Edit/WebThumbNailList.cs` line 283 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5402#-NCWsANl06rXQEdT6y0T:-NC_54c0EaHStfrWri6W:b-8p7r9q) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/8b6dc2df6a55aa98836baf0e5cdca8e60a50f4d0/src/BloomExe/Edit/WebThumbNailList.cs#L283)):*

Sorry, I just mean, if the right way to Navigate is to `Invoke((Action) (() => `, why not put that in the class? Why leave it to each caller to figure out? Maybe that doesn't make sense, just askin'
</blockquote></details>

Yes, that's what the second paragraph above is addressing. Invoke() is something that can only be done by a Windows control that has a handle. Sometimes that could be the Browser itself. Sometimes that won't work.

hatton (approved):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5402)* status: all files reviewed, 1 unresolved discussion (waiting on @JohnThomson)

___
*[`src/BloomExe/Edit/WebThumbNailList.cs` line 283 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5402#-NCWsANl06rXQEdT6y0T:-NCkOtju6A3M7VZsij65:b-fcdh16) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/8b6dc2df6a55aa98836baf0e5cdca8e60a50f4d0/src/BloomExe/Edit/WebThumbNailList.cs#L283)):*

Yes, that's what the second paragraph above is addressing. Invoke() is something that can only be done by a Windows control that has a handle. Sometimes that could be the Browser itself. Sometimes that won't work.
</blockquote></details>

I don't think it's the case in c# that something can "only be done" by code in location A, not B. So this same code could be in the class rather than in every caller of the class. Not worth continuing to go back and forth though :-)

PR #5423, opened 2022-09-29
Cleanup of ExpensiveInitialization and optimization of BringBookUpToDate (BL-11308)

JohnThomson:
This change is [Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5423)

### DIFF
--- a/src/BloomExe/Book/Book.cs
+++ b/src/BloomExe/Book/Book.cs
@@ -846,8 +846,6 @@ namespace Bloom.Book
 			{
 				return GetErrorDom();
 			}
-			//Console.WriteLine("DEBUG GetPreviewHtmlFileForWholeBook(): calling BringBookUpToDate() for new previewDOM");
-			BringBookUpToDateInternal(previewDom, new NullProgress());
 
 			// this is normally the vernacular, but when we're previewing a shell, well it won't have anything for the vernacular
 			var primaryLanguage = Language1IsoCode;
--- a/src/BloomExe/Book/Book.cs
+++ b/src/BloomExe/Book/Book.cs
@@ -1360,8 +1360,6 @@ namespace Bloom.Book
 			// anything (may fix BL-3166).
 			var licenseMetadata = GetLicenseMetadata();
 
-			Storage.CompleteFullyUpdatingFilesIfNeeded();
-
 			progress.WriteStatus("Updating collection settings...");
 			try
 			{
--- a/src/BloomExe/Book/BookStorage.cs
+++ b/src/BloomExe/Book/BookStorage.cs
@@ -41,8 +41,6 @@ namespace Bloom.Book
 		string Key { get; }
 		string FolderName { get; }
 		string FolderPath { get; }
-
-		void CompleteFullyUpdatingFilesIfNeeded();
 		string PathToExistingHtml { get; }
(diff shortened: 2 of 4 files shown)


andrew-polk (changes requested):
> Reviewed 2 of 2 files at r1, 2 of 2 files at r2, 2 of 2 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5423)* status: all files reviewed, 1 unresolved discussion (waiting on @JohnThomson)

___
*[`-- commits` line 9 at r2](https://reviewable.io/reviews/bloombooks/bloomdesktop/5423#-NENG3CTD-CFaWlghUMz):*
Sorry, I can't find this (a BookStorage created by BBUD).

Every book object has a BookStorage, of course, but can't the BookStorage have been constructed and then we do something when needs the calling of ExpensiveIntialization again?



_Code quote:_
```
BBUD has a BookStorage
```

JohnThomson (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5423)* status: all files reviewed, 1 unresolved discussion (waiting on @andrew-polk)

___
*[`-- commits` line 9 at r2](https://reviewable.io/reviews/bloombooks/bloomdesktop/5423#-NENG3CTD-CFaWlghUMz):*

Sorry, I can't find this (a BookStorage created by BBUD).

Every book object has a BookStorage, of course, but can't the BookStorage have been constructed and then we do something when needs the calling of ExpensiveIntialization again?
</blockquote></details>

Sorry, I didn't express that well. The point is that Book has a BookStorage, and its constructor called ExpensiveInitialization, and there are no longer two versions of it, so that did everything that ExpensiveInitialization ever does, so it's a waste to call it again in BBUD, at least at startup.

Now, that leaves the question of whether it might be necessary at any other time BBUD is called. I don't think so. From the name, it's supposed to be something that only needs doing as a BookStorage is initialized. The fact that BBUD needed to call what should be a private method is part of the anomaly this PR is supposed to correct. If the book has been so utterly mangled that we need to do this again (e.g., spreadsheet import or TeamCollection forget changes), we don't just do BBUD, we do BookStorage.ReloadFromDisk.

As I look at what it actually does, it's about validating the HTML (just one HTML file, syntactically valid, doesn't require a newer Bloom), possibly restoring a backup if something is wrong, bringing some support files up to date with the current version of Bloom. The only thing that might conceivably need redoing later that I can see is adding FolderPath to the _fileLocator, which (if it's needed at all, which I doubt very much) should be done again if the folder name changes. But it isn't being, even now; renaming the book doesn't result in ExpensiveInitialization getting called, either through BBUD or any other way.

Of course, it's conceivable that between selecting a book and, say, switching to publish tab the user has gone into the book folder and renamed the HTML file and messed up its syntax and added another one or similar things. And calling EI again at that point might give such a pathological user a better error message than whatever disaster his misbehavior will otherwise cause. But I don't think the chance of it is worth what it costs to call this twice every time we select a book.

andrew-polk (approved):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5423)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @JohnThomson)

PR #5431, opened 2022-09-30
Fix upload new book then cancel makes zombie in Bloom Library (BL-10811)

sujeffreyl:
This fix will prevent a Parse entry from being created for the new book in Parse, and consequently prevent the book from showing up in BloomLibrary. It doesn't change the books on S3, but that doesn't matter. (Deleting a book only deletes the Parse entry, not S3 objects, so orphaned S3 objects are not a big disaster for us). This fix does not deal with the case that an uploaded book already exists, is re-uploaded and then cancelled. (The book will be left in a partial state)

### DIFF
--- a/src/BloomExe/WebLibraryIntegration/BookUpload.cs
+++ b/src/BloomExe/WebLibraryIntegration/BookUpload.cs
@@ -242,7 +242,7 @@ namespace Bloom.WebLibraryIntegration
 						metaMsg = "(Dry run) Would upload book metadata";	// TODO: localize?
 					progress.WriteStatus(metaMsg);
 					// Do this after uploading the books, since the ThumbnailUrl is generated in the course of the upload.
-					if (!IsDryRun)
+					if (!IsDryRun && !progress.CancelRequested)
 					{
 						var response = ParseClient.SetBookRecord(metadata.WebDataJson);
 						parseId = response.ResponseUri.LocalPath;


sujeffreyl (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5431)* status: 0 of 1 files reviewed, 1 unresolved discussion

___
*[`src/BloomExe/WebLibraryIntegration/BookUpload.cs` line 259 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5431#-NDE40DR22NH_LgHvHOV:-NDE40DR22NH_LgHvHOW:blostwf) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/5df9972056d8c8d43b48d086c558cebc1c8a02d4/src/BloomExe/WebLibraryIntegration/BookUpload.cs#L259)):*
> ```Smalltalk
> 						//   if (!UseSandbox) // don't make it seem like there are more uploads than their really are if this a tester pushing to the sandbox
> 						{
> 							Analytics.Track("UploadBook-Success", new Dictionary<string, string>() { { "url", metadata.BookOrder }, { "title", metadata.Title } });
> ```

Do we want/need analytics for Cancel? We could add an "UploadBook-Cancelled" event if desired.

andrew-polk (approved):
> Dismissed @sujeffreyl from a discussion.
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5431)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @sujeffreyl)

___
*[`src/BloomExe/WebLibraryIntegration/BookUpload.cs` line 259 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5431#-NDE40DR22NH_LgHvHOV:-NDEkzcf0TXf4ciMjdLb:b-wtpjt4) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/5df9972056d8c8d43b48d086c558cebc1c8a02d4/src/BloomExe/WebLibraryIntegration/BookUpload.cs#L259)):*

Do we want/need analytics for Cancel? We could add an "UploadBook-Cancelled" event if desired.
</blockquote></details>

I don't think so.

PR #5435, opened 2022-09-30
Record Video make preview with only publishable langs (BL-11332)

JohnThomson:
This change is [Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5435)

### DIFF
--- a/src/BloomExe/Book/LicenseChecker.cs
+++ b/src/BloomExe/Book/LicenseChecker.cs
@@ -29,6 +29,25 @@ namespace Bloom.Book
         public static string kUnlicenseLanguageMessage = "The copyright holder of this book has not licensed it for publishing in {0}. Please contact the copyright holder to learn more about licensing.";
         public static string kCannotReachLicenseServerMessage = "To publish this book, Bloom must check which languages the copyright owner has permitted, but Bloom is having trouble reaching the server that has this information.";
 
+        /// <summary>
+		/// Return that subset of inputLangs that are allowed to be published for the specified book. Usually all of them.
+		/// </summary>
+		/// <remarks>
+		/// If the book has a restricted license and we cannot contact the server and have not previously cached
+		/// information about licensing this book, the method will return an empty list. Currently this works
+		/// because we will try to build a preview with all the languages, and the check that is done during
+		/// that process will report that it can't confirm which ones are publishable.
+		/// </remarks>
+		public IEnumerable<string> AllowedLanguages(string[] inputLangs, Book book)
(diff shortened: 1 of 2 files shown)


andrew-polk (changes requested):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5435)* status: all files reviewed, 1 unresolved discussion (waiting on @JohnThomson)

___
*[`src/BloomExe/Book/LicenseChecker.cs` line 40 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5435#-NDFaU_y4WwHqc0_Y8J0:-NDFaU_y4WwHqc0_Y8J1:bg6q6xs) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/9f456b2515b7717b96b7b58069e9f4097c853c73/src/BloomExe/Book/LicenseChecker.cs#L40)):*
> ```Smalltalk
> 	        var key = meta.GetStringAttribute("content");
> 	        var problems = GetProblemLanguages(inputLangs, key, out bool didCheck);
> 	        return inputLangs.Except(problems);
> ```

If I understand this correctly, we need to make sure `didCheck` is true. Otherwise, we don't actually know that the languages are allowed.
I think we can repurpose the `kCannotReachLicenseServerMessage` message.

JohnThomson (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5435)* status: all files reviewed, 1 unresolved discussion (waiting on @andrew-polk)

___
*[`src/BloomExe/Book/LicenseChecker.cs` line 40 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5435#-NDFaU_y4WwHqc0_Y8J0:-NDT50YUAxsK-WOzI-qY:b-9yw0yr) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/9f456b2515b7717b96b7b58069e9f4097c853c73/src/BloomExe/Book/LicenseChecker.cs#L40)):*

If I understand this correctly, we need to make sure `didCheck` is true. Otherwise, we don't actually know that the languages are allowed.
I think we can repurpose the `kCannotReachLicenseServerMessage` message.
</blockquote></details>

I added some comments to clarify my thinking

andrew-polk (approved):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5435)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @JohnThomson)

PR #5439, opened 2022-10-03
Ensure new text is handled for talking book tool (BL-11522)

StephenMcConnel:
This change is [Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5439)

### DIFF
--- a/src/BloomBrowserUI/bookEdit/toolbox/talkingBook/recordable.ts
+++ b/src/BloomBrowserUI/bookEdit/toolbox/talkingBook/recordable.ts
@@ -192,7 +192,9 @@ export default class Recordable {
     }
 
     public async shouldUpdateMarkupAsync(): Promise<boolean> {
-        if (this.areAnyRecordingsOutOfDate()) {
+        if (this.checkForMissingAudioMarkup()) {
+            return true; // that was quick...
+        } else if (this.areAnyRecordingsOutOfDate()) {
             // Should be updated
             return true;
         } else {
@@ -210,6 +212,29 @@ export default class Recordable {
                 return true;
             }
         }
+    }
+
+    // check whether our textBox has any text that is not covered by an audio sentence.
+    // Users can type text in when the talking book tool is closed, after all.  See BL-11522.
+    private checkForMissingAudioMarkup() {
+        if (this.textBox.classList.contains(kAudioSentence)) {
+            return false;
+        }
+        let textMissingMarkup = this.textBox.innerText;
+        const sentenceList = this.getAudioSentences();
+        for (let i = 0; i < sentenceList.length; ++i) {
+            const element = sentenceList[i];
+            textMissingMarkup = textMissingMarkup.replace(
+                element.innerText,
+                ""
+            );
+        }
+        // regex based on what the extension method LibSynphony.stringToSentences() uses.
+        const intersentenceSpace = /^[\s\p{PEP}\u00A0]+/;
+        return (
+            textMissingMarkup &&
+            textMissingMarkup.replace(intersentenceSpace, "").length > 0
+        );
     }
 
     // Determines whether the text box is fully, partially, or not recorded all in one go,


JohnThomson (changes requested):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5439)* status: 0 of 1 files reviewed, 3 unresolved discussions (waiting on @JohnThomson and @StephenMcConnel)

___
*[`src/BloomBrowserUI/bookEdit/toolbox/talkingBook/audioRecording.ts` line 2672 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5439#-NDhEwaADdY2GBkIwXfk:-NDhEwaADdY2GBkIwXfl:bj4vpnl) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/a3c37b54bf54c70ca551a40ed44740468fa5aa38/src/BloomBrowserUI/bookEdit/toolbox/talkingBook/audioRecording.ts#L2672)):*
> ```XML
>         const shouldUpdateMarkup = await recordable.shouldUpdateMarkupAsync();
> 
>         const missingMarkup = this.checkForMissingAudioMarkup(textBox);
> ```

Should this logic not be part of shouldUpdateMarkupAsync? Are there other callers that should not consider the possibility of new text outside the existing sentence objects? (perhaps the new method should be called from there, if a separate method is useful; but it may also be more efficient to combine them).
___
*[`src/BloomBrowserUI/bookEdit/toolbox/talkingBook/audioRecording.ts` line 2692 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5439#-NDhEXpL8M3tK5N74DKu:-NDhEXpL8M3tK5N74DKv:b-um6kj3) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/a3c37b54bf54c70ca551a40ed44740468fa5aa38/src/BloomBrowserUI/bookEdit/toolbox/talkingBook/audioRecording.ts#L2692)):*
> ```XML
>     private checkForMissingAudioMarkup(textBox: HTMLElement) {
>         let textMissingMarkup = textBox.innerText;
>         const sentenceList = textBox.getElementsByClassName(kAudioSentence);
> ```

I believe this will not find the textBox itself; but in TextBox recording mode, it is the whole TextBox that has this class, and if it does, there would be no missing markup on anything inside it and the method could immediately return false.
___
*[`src/BloomBrowserUI/bookEdit/toolbox/talkingBook/audioRecording.ts` line 2701 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5439#-NDhG-sq3AXRGNGTiZPL:-NDhG-sr2TE_gI1hz2XJ:bi9598f) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/a3c37b54bf54c70ca551a40ed44740468fa5aa38/src/BloomBrowserUI/bookEdit/toolbox/talkingBook/audioRecording.ts#L2701)):*
> ```XML
>                 );
>         }
>         return textMissingMarkup && textMissingMarkup.trim().length > 0;
> ```

Do trim() and our sentence-breaking algorithm agree about what constitutes white space? If not, this will tend to return true when (e.g.) the sentence breaking algorithm excludes something like a wide Unicode space from the sentences we make.

StephenMcConnel (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5439)* status: 0 of 2 files reviewed, 3 unresolved discussions (waiting on @JohnThomson)

___
*[`src/BloomBrowserUI/bookEdit/toolbox/talkingBook/audioRecording.ts` line 2672 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5439#-NDhEwaADdY2GBkIwXfk:-NDiMJGUBwxqqdAv0SkN:b-896fix) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/a3c37b54bf54c70ca551a40ed44740468fa5aa38/src/BloomBrowserUI/bookEdit/toolbox/talkingBook/audioRecording.ts#L2672)):*

Should this logic not be part of shouldUpdateMarkupAsync? Are there other callers that should not consider the possibility of new text outside the existing sentence objects? (perhaps the new method should be called from there, if a separate method is useful; but it may also be more efficient to combine them).
</blockquote></details>

Done.
___
*[`src/BloomBrowserUI/bookEdit/toolbox/talkingBook/audioRecording.ts` line 2692 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5439#-NDhEXpL8M3tK5N74DKu:-NDiOJAU9lawpGuAAmq_:b-896fix) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/a3c37b54bf54c70ca551a40ed44740468fa5aa38/src/BloomBrowserUI/bookEdit/toolbox/talkingBook/audioRecording.ts#L2692)):*

I believe this will not find the textBox itself; but in TextBox recording mode, it is the whole TextBox that has this class, and if it does, there would be no missing markup on anything inside it and the method could immediately return false.
</blockquote></details>

Done.
___
*[`src/BloomBrowserUI/bookEdit/toolbox/talkingBook/audioRecording.ts` line 2701 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5439#-NDhG-sq3AXRGNGTiZPL:-NDifjKcE-_ptUp2konQ:b-yrn2e7) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/a3c37b54bf54c70ca551a40ed44740468fa5aa38/src/BloomBrowserUI/bookEdit/toolbox/talkingBook/audioRecording.ts#L2701)):*

Do trim() and our sentence-breaking algorithm agree about what constitutes white space? If not, this will tend to return true when (e.g.) the sentence breaking algorithm excludes something like a wide Unicode space from the sentences we make.
</blockquote></details>

I've changed to use a regex based on the "inter-sentence space" used by the Libsynphony sentence splitting method.

JohnThomson (approved):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5439)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @StephenMcConnel)

PR #5451, opened 2022-10-06
Remove "Use ePUB reader's text size" (BL-11499)

JohnThomson:
This change is [Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5451)


hatton (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5451)* status: 0 of 4 files reviewed, 1 unresolved discussion (waiting on @hatton and @JohnThomson)

___
*[`src/BloomExe/Publish/Epub/PublishEpubApi.cs` line 293 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5451#-NDmr0S2DmldjyrlMb3I:-NDmr0S2DmldjyrlMb3J:b1cydf7) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/1a766410bc79ff8c3287f5a3d3ddc856ea9673d5/src/BloomExe/Publish/Epub/PublishEpubApi.cs#L293)):*
> ```Smalltalk
> 			// up to the reader always, while 'fixed' ones must control everything.
> 			// (Previous versions of Bloom had a checkbox and API to control this in flowable mode.)
> 			EpubMaker.RemoveFontSizes = settings.Mode == "flowable";
> ```

I'm little surprised that in c# we're using a string as the symbol, vs. epub (vs typescript string values can be type-checked)?

JohnThomson (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5451)* status: 0 of 4 files reviewed, 1 unresolved discussion (waiting on @hatton)

___
*[`src/BloomExe/Publish/Epub/PublishEpubApi.cs` line 293 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5451#-NDmr0S2DmldjyrlMb3I:-NDn2FwK6wF0ZkiScIm3:b-ujrb35) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/1a766410bc79ff8c3287f5a3d3ddc856ea9673d5/src/BloomExe/Publish/Epub/PublishEpubApi.cs#L293)):*

I'm little surprised that in c# we're using a string as the symbol, vs. epub (vs typescript string values can be type-checked)?
</blockquote></details>

Yes, it's a pain using an enumeration in C# and converting to a constrained list of allowed strings in TS. Maybe it's worth doing. I can if you want. Probably should be a separate PR, though.

PR #5461, opened 2022-10-13
Fix a problem when pasting an image (BL-11558)

StephenMcConnel:
This change is [Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5461)

### DIFF
--- a/src/BloomExe/ImageProcessing/ImageUtils.cs
+++ b/src/BloomExe/ImageProcessing/ImageUtils.cs
@@ -180,6 +180,11 @@ namespace Bloom.ImageProcessing
 					RobustFile.Copy(sourcePath, destinationPath);
 				else
 					imageInfo.Image.Save(destinationPath, ImageFormat.Png); // destinationPath already has .png extension
+				if (_createdTempImageFile != null && RobustFile.Exists(_createdTempImageFile))
+				{
+					RobustFile.Delete(_createdTempImageFile);
+					_createdTempImageFile = null;
+				}
 				return imageFileName;
 			}
 			catch (IOException)
@@ -723,19 +728,7 @@ namespace Bloom.ImageProcessing
 				var isJpegImage = AppearsToBeJpeg(imageInfo);
 				if (String.IsNullOrEmpty(sourcePath) || !RobustFile.Exists(sourcePath))
 				{
-					// This must be from a paste instead of the ImageChooser dialog.
-					sourcePath = Path.GetTempFileName();
-					RobustFile.Delete(sourcePath);
-					if (isJpegImage)
-					{
-						sourcePath = sourcePath + ".jpg";
-						imageInfo.Image.Save(sourcePath, ImageFormat.Jpeg);
-					}
-					else
-					{
-						sourcePath = sourcePath + ".png";
-						imageInfo.Image.Save(sourcePath, ImageFormat.Png);
-					}
+					sourcePath = CreateSourceFileForImage(imageInfo, isJpegImage);
 				}
 				var destPath = TempFileUtils.GetTempFilepathWithExtension(isJpegImage ? ".jpg" : ".png");
 				try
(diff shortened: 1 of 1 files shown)


hatton:
It's difficult to see what changed because of the refactor. Do you want to give us a description? Thx.

StephenMcConnel (commented):
TryResizeImageWithGraphicsMagick() created a temporary file for GraphicsMagick to use when the image came from being pasted, but then promptly deleted the file at the end of the method.
TryResizeImageWithGraphicsMagick() did not try this, so a pasted image still had a null file, causing the problem report to pop up.
I essentially did three things i this changeset:
1) Abstract/refactor a method for creating the temporary file, which is now called from both TryResizeImageWithGraphicsMagick and TryResizeImageWithGraphicsMagick.
2) In that new method, store the path to the temporary file both in a class variable and in the image (where it may get overwritten or it may prevent TryResizeImageWithGraphicsMagick() from deleting it and still be available for TryResizeImageWithGraphicsMagick to use)
3) use the class variable to delete the file if it hasn't already been deleted by the time we know we'd be through with it

> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5461#-:-NEIOjsp81YbmoGxpSsF:b-mah3xm)* status: 0 of 1 files reviewed, all discussions resolved

JohnThomson (approved):
A minor suggestion...up to you whether it's worth the trouble. If you don't think so, feel free to merge.

> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5461#-:-NEc0GVyA1DrmNfWY5c5:b9q0bbp)* status: all files reviewed, 2 unresolved discussions (waiting on @StephenMcConnel)

___
*[`src/BloomExe/ImageProcessing/ImageUtils.cs` line 184 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5461#-NEc-U8c9mpyCxLua9P7:-NEc-U8c9mpyCxLua9P8:bklmlt1) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/cb8828995b00177fd02f2521142e041f48b3981d/src/BloomExe/ImageProcessing/ImageUtils.cs#L184)):*
> ```Smalltalk
> 					imageInfo.Image.Save(destinationPath, ImageFormat.Png); // destinationPath already has .png extension
> 				if (_createdTempImageFile != null && RobustFile.Exists(_createdTempImageFile))
> 					RobustFile.Delete(_createdTempImageFile);
> ```

Maybe worth setting the variable to null here?
___
*[`src/BloomExe/ImageProcessing/ImageUtils.cs` line 798 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5461#-NEc-aypCgUSiyARP8pF:-NEc-aypCgUSiyARP8pG:b9fh94k) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/cb8828995b00177fd02f2521142e041f48b3981d/src/BloomExe/ImageProcessing/ImageUtils.cs#L798)):*
> ```Smalltalk
> 			}
> 			imageInfo.SetCurrentFilePath(sourcePath);
> 			_createdTempImageFile = sourcePath;
> ```

This code would be in trouble with race conditions involving this variable if ever called on multiple threads. I'm not very worried since it's aimed at things the user does one at at time. But we could help ourselves catch this, in case we ever used it another way, by setting _createdTempImageFile to null above when done with it, and here throwing (or at least asserting) that it is null.

StephenMcConnel (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5461)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @JohnThomson)

___
*[`src/BloomExe/ImageProcessing/ImageUtils.cs` line 184 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5461#-NEc-U8c9mpyCxLua9P7:-NEffkq345S_ILkrYaiL:b-896fix) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/cb8828995b00177fd02f2521142e041f48b3981d/src/BloomExe/ImageProcessing/ImageUtils.cs#L184)):*

Maybe worth setting the variable to null here?
</blockquote></details>

Done.
___
*[`src/BloomExe/ImageProcessing/ImageUtils.cs` line 798 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5461#-NEc-aypCgUSiyARP8pF:-NEfghgS9GU_3c2Ylqqi:b-896fix) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/cb8828995b00177fd02f2521142e041f48b3981d/src/BloomExe/ImageProcessing/ImageUtils.cs#L798)):*

This code would be in trouble with race conditions involving this variable if ever called on multiple threads. I'm not very worried since it's aimed at things the user does one at at time. But we could help ourselves catch this, in case we ever used it another way, by setting _createdTempImageFile to null above when done with it, and here throwing (or at least asserting) that it is null.
</blockquote></details>

Done.

JohnThomson (approved):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5461)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @StephenMcConnel)

PR #5475, opened 2022-10-18
Move bubbles more smoothly (BL-11615)

JohnThomson:
This change is [Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5475)

### DIFF
--- a/src/BloomBrowserUI/bookEdit/js/bubbleManager.ts
+++ b/src/BloomBrowserUI/bookEdit/js/bubbleManager.ts
@@ -782,19 +782,31 @@ export class BubbleManager {
 
     // Setup event handlers that allow the bubble to be moved around or resized.
     private setMouseDragHandlers(container: HTMLElement): void {
+        // An earlier version of this code set onmousedown to this.onMouseDown, etc.
+        // We need to use addEventListener so we can capture.
+        // It's unlikely, but I can't rule it out, that a deliberate side effect
+        // was to remove some other onmousedown handler. Just in case, clear the fields.
+        // I don't think setting these has any effect on handlers done with addEventListener,
+        // but just in case, I'm doing this first.
+        container.onmousedown = null;
+        container.onmousemove = null;
+        container.onmouseup = null;
+
         // We use mousemove effects instead of drag due to concerns that drag effects would make the entire image container appear to drag.
         // Instead, with mousemove, we can make only the specific bubble move around
-        container.onmousedown = (event: MouseEvent) => {
-            this.onMouseDown(event, container);
-        };
+        // Grabbing these (particularly the move event) in the capture phase allows us to suppress
+        // effects of ctrl and alt clicks on the text.
+        container.addEventListener("mousedown", this.onMouseDown, {
+            capture: true
+        });
 
-        container.onmousemove = (event: MouseEvent) => {
-            this.onMouseMove(event, container);
-        };
+        container.addEventListener("mousemove", this.onMouseMove, {
+            capture: true
+        });
 
-        container.onmouseup = (event: MouseEvent) => {
-            this.onMouseUp(event, container);
-        };
+        container.addEventListener("mouseup", this.onMouseUp, {
+            capture: true
+        });
 
         container.onkeypress = (event: Event) => {
             // If the user is typing in a bubble, make sure automatic shrinking is off.
@@ -930,11 +942,20 @@ export class BubbleManager {
         }
     }
 
-    private onMouseDown(event: MouseEvent, container: HTMLElement) {
+    // MUST be defined this way, rather than as a member function, so that it can
+    // be passed directly to addEventListener and still get the correct 'this'.
+    private onMouseDown = (event: MouseEvent) => {
+        const container = event.currentTarget as HTMLElement;
         // Let standard clicks on the bloom editable only be processed on the editable
         if (this.isEventForEditableOnly(event)) {
             return;
         }
+        // since that returned false, either ctrl or alt is down, or we clicked outside the
+        // editable. If we're outside the editable, we don't need any default event processing,
+        // and if we're inside and ctrl or alt is down, we want to prevent the events being
+        // processed by the text.
+        event.preventDefault();
+        event.stopPropagation();
 
         // These coordinates need to be relative to the canvas (which is the same as relative to the image container).
         const coordinates = this.getPointRelativeToCanvas(event, container);
@@ -975,6 +996,8 @@ export class BubbleManager {
                 // Move action started
                 this.bubbleToDrag = bubble;
                 this.activeContainer = container;
+                // in case this is somehow left from earlier, we want a fresh start for the new move.
+                this.animationFrame = 0;
                 // mouse is presumably moving inside this image, it should have the buttons.
                 // (It does not get them in the usual way by mouseEnter, because that event is
                 // suppressed by the comicaljs canvas, which is above the image container.)
@@ -1026,9 +1049,12 @@ export class BubbleManager {
                 };
             }
         }
-    }
+    };
 
-    private onMouseMove(event: MouseEvent, container: HTMLElement) {
+    // MUST be defined this way, rather than as a member function, so that it can
+    // be passed directly to addEventListener and still get the correct 'this'.
+    private onMouseMove = (event: MouseEvent) => {
+        const container = event.currentTarget as HTMLElement;
         // Prevent two event handlers from triggering if the text box is currently being resized
         if (this.isResizing(container)) {
             this.bubbleToDrag = undefined;
@@ -1043,7 +1069,7 @@ export class BubbleManager {
         } else {
             this.handleMouseMoveResizeBubble(event, container);
         }
-    }
+    };
 
     // Mouse hover - No move or resize is currently active, but check if there is a bubble under the mouse that COULD be
     // and add or remove the classes we use to indicate this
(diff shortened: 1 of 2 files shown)


jsubloom (changes requested):
Does this change address the box moving to the bottom-right at the very beginning? I looked through the change but am not following how it would address it.

Also, I agree with the concerns JH expressed during the meeting

> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5475#-:-NEkczK75nm51gIbZ-Pp:b-qqsa8r)* status: 0 of 2 files reviewed, 2 unresolved discussions (waiting on @JohnThomson)

___
*[`src/BloomBrowserUI/bookEdit/js/bubbleManager.ts` line 1165 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5475#-NEkd3y41MEsYXpGjlzF:-NEkd3y41MEsYXpGjlzG:btg3deh) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/a70f3941c4b3ae583583747c85b4803586914396/src/BloomBrowserUI/bookEdit/js/bubbleManager.ts#L1165)):*
> ```XML
>             // then certain to get processed. But it seemed to be significantly less effective
>             // at getting frames fully rendered often, and the difference in where the box ends up
>             // is unlikely to be significant...the user will keep dragging until satisfied.
> ```

Hmm... I'm a little hesitant to buy into this, despite the performance benefit from doing it this way. So, say the user drags it quite a ways, but the CPU is busy doing whatever and it's still stuck processing the first frame of the animation by the time the user has dragged the mouse to its final destination. Doesn't that mean that the result will be that the box is rendered at the first frame and will never be rendered at the final frame unless the user jiggles their mouse again? That seems not ideal either.
___
*[`src/BloomBrowserUI/bookEdit/js/bubbleManager.ts` line 1193 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5475#-NEkcPUO1o7Z21W3MW5N:-NEkcPUO1o7Z21W3MW5O:b-td8mtt) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/a70f3941c4b3ae583583747c85b4803586914396/src/BloomBrowserUI/bookEdit/js/bubbleManager.ts#L1193)):*
> ```XML
>         //   to stop the event from reaching the target element (the paragraph or the .bloom-editable div or whatever).
>         //   Unfortunately, while it prevented the event handlers in the subsequent Bubble phase from being fired,
>         //   this funny dual-drag behavior would still happen.  I don't have a solution for that yet.
> ```

Does this change fix the problem? If so, could we delete these comments?

JohnThomson (commented):
I've not been able to reproduce that so have not been able to address it. I have a theory that it was fixed between the original report and when I started on this.

> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5475#-:-NEmJE707PyDW-msvTWT:b-t7hfxl)* status: 0 of 2 files reviewed, 2 unresolved discussions (waiting on @sujeffreyl)

___
*[`src/BloomBrowserUI/bookEdit/js/bubbleManager.ts` line 1165 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5475#-NEkd3y41MEsYXpGjlzF:-NEmHLlNDluAzTtdvkBh:b-emaj04) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/a70f3941c4b3ae583583747c85b4803586914396/src/BloomBrowserUI/bookEdit/js/bubbleManager.ts#L1165)):*

Hmm... I'm a little hesitant to buy into this, despite the performance benefit from doing it this way. So, say the user drags it quite a ways, but the CPU is busy doing whatever and it's still stuck processing the first frame of the animation by the time the user has dragged the mouse to its final destination. Doesn't that mean that the result will be that the box is rendered at the first frame and will never be rendered at the final frame unless the user jiggles their mouse again? That seems not ideal either.
</blockquote></details>

I added a comment about this. I don't think it's possible to lose more than about 1/30 second worth of movement even on a slowish machine.
___
*[`src/BloomBrowserUI/bookEdit/js/bubbleManager.ts` line 1193 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5475#-NEkcPUO1o7Z21W3MW5N:-NEmJ5s04_1bg9Bl0vVp:b-86x1n9) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/a70f3941c4b3ae583583747c85b4803586914396/src/BloomBrowserUI/bookEdit/js/bubbleManager.ts#L1193)):*

Does this change fix the problem? If so, could we delete these comments?
</blockquote></details>

The first round of changes didn't entirely fix this, but now they do, so I removed it.

jsubloom (changes requested):
Not sure if it addresses all the issues, but it seems to address enough to be worthwhile! Just a couple minor comments RE the new changes.

> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5475#-:-NEqHQET3Y-GbUHIjkg0:b72bz1g)* status: 1 of 2 files reviewed, 2 unresolved discussions (waiting on @JohnThomson)

___
*[`src/BloomBrowserUI/bookEdit/js/bubbleManager.ts` line 787 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5475#-NEqElY46EuwlRLGDqPa:-NEqElY5-j-TbHRQpPfN:b-dgi5ei) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/a70f3941c4b3ae583583747c85b4803586914396/src/BloomBrowserUI/bookEdit/js/bubbleManager.ts#L787)):*
> ```XML
>         // We use mousemove effects instead of drag due to concerns that drag effects would make the entire image container appear to drag.
>         // Instead, with mousemove, we can make only the specific bubble move around
>         container.onmousedown = (event: MouseEvent) => {
> ```

I think this is okay, but just wanted to check:
we're not expecting there to be any other existing event listeners that we need to remove, right?
___
*[`src/BloomBrowserUI/bookEdit/js/bubbleManager.ts` line 1165 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5475#-NEkd3y41MEsYXpGjlzF:-NEqH6tr1pLUVaC7kmTA:bjcxh29) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/a70f3941c4b3ae583583747c85b4803586914396/src/BloomBrowserUI/bookEdit/js/bubbleManager.ts#L1165)):*

I added a comment about this. I don't think it's possible to lose more than about 1/30 second worth of movement even on a slowish machine.
</blockquote></details>

Ah, I see. the callback processes the event which is stored via a member variable, so the member variable should be getting updated all the time...
Ok, thanks for the explanation!
___
*[`src/BloomBrowserUI/bookEdit/js/bubbleManager.ts` line 1193 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5475#-NEkcPUO1o7Z21W3MW5N:-NEqGETr4ncLF4Sa-8R8:b-jjihnk) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/a70f3941c4b3ae583583747c85b4803586914396/src/BloomBrowserUI/bookEdit/js/bubbleManager.ts#L1193)):*

The first round of changes didn't entirely fix this, but now they do, so I removed it.
</blockquote></details>

Wonderful! Thanks for fixing that issue!
___
*[`src/BloomBrowserUI/bookEdit/js/bubbleManager.ts` line 788 at r2](https://reviewable.io/reviews/bloombooks/bloomdesktop/5475#-NEqFPcCCFOa9YK0o6rk:-NEqFPcCCFOa9YK0o6rl:bsd7cpe) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/38f54f7bf0e4dd7e10824eba9ee1bc0f043de6fd/src/BloomBrowserUI/bookEdit/js/bubbleManager.ts#L788)):*
> ```XML
>     // function rather than a new one on each callto setDragHandlers, we won't get duplicates if we add more
>     // than once.
>     private onMouseDownBound = this.onMouseDown.bind(this);
> ```

I wonder if this can be simplified via a different trick. What about changing the definition of this.onMouseDown to use arrow functions?
e.g. instead of private onMouseDown([...]) { [...] }, change it to private onMouseDown = (...) => { ...}. I think this syntax would have the "this" keyword be bound properly inside the arrow function without needing to call .bind().
I think this should work, not 100% sure. I'd prefer that syntax though over doing the .bind() calls

JohnThomson (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5475)* status: 0 of 2 files reviewed, 2 unresolved discussions (waiting on @sujeffreyl)

___
*[`src/BloomBrowserUI/bookEdit/js/bubbleManager.ts` line 787 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5475#-NEqElY46EuwlRLGDqPa:-NEqNYI1A_vIJE6Ql1VQ:b-23bag7) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/a70f3941c4b3ae583583747c85b4803586914396/src/BloomBrowserUI/bookEdit/js/bubbleManager.ts#L787)):*

I think this is okay, but just wanted to check:
we're not expecting there to be any other existing event listeners that we need to remove, right?
</blockquote></details>

Good thought. I don't know of any, but added some code just in case.
___
*[`src/BloomBrowserUI/bookEdit/js/bubbleManager.ts` line 788 at r2](https://reviewable.io/reviews/bloombooks/bloomdesktop/5475#-NEqFPcCCFOa9YK0o6rk:-NEqNOpz917TfFJMUlx9:bxeb39p) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/38f54f7bf0e4dd7e10824eba9ee1bc0f043de6fd/src/BloomBrowserUI/bookEdit/js/bubbleManager.ts#L788)):*

I wonder if this can be simplified via a different trick. What about changing the definition of this.onMouseDown to use arrow functions?
e.g. instead of private onMouseDown([...]) { [...] }, change it to private onMouseDown = (...) => { ...}. I think this syntax would have the "this" keyword be bound properly inside the arrow function without needing to call .bind().
I think this should work, not 100% sure. I'd prefer that syntax though over doing the .bind() calls
</blockquote></details>

You're right, that works too. Done.

jsubloom (changes requested):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5475)* status: 1 of 2 files reviewed, 1 unresolved discussion (waiting on @JohnThomson)

___
*[`src/BloomBrowserUI/bookEdit/js/bubbleManager.ts` line 789 at r3](https://reviewable.io/reviews/bloombooks/bloomdesktop/5475#-NEqap7x9yEC3Yrp1vfb:-NEqap7x9yEC3Yrp1vfc:b-aigg2q) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/7d25dcf4f1c364fb1099b82c21ca60700f391738/src/BloomBrowserUI/bookEdit/js/bubbleManager.ts#L789)):*
> ```XML
>         // It's unlikely, but I can't rule it out, that a deliberate side effect
>         // was to remove some other onmousedown handler. Just in case, clear the fields.
>         // I don't think setting these has any effect on handlers done with addEventListener,
> ```

I would guess that you definitely need to do it first because I think it would have an effect on addEventListener handlers, but oh well.

jsubloom (approved):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5475)* status: 1 of 2 files reviewed, 1 unresolved discussion (waiting on @JohnThomson)

jsubloom (commented):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5475#-:-NEqbE6X9GkH_ovEPEBP:bnfp4nl)* status: 1 of 2 files reviewed, 1 unresolved discussion (waiting on @JohnThomson)

jsubloom (approved):
> Reviewed 1 of 2 files at r1.
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5475)* status: all files reviewed, 1 unresolved discussion (waiting on @JohnThomson)

jsubloom (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5475#-:-NEqbLvu-fHogIo0mwqL:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on @JohnThomson)

jsubloom (approved):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5475)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @JohnThomson)

PR #5477, opened 2022-10-19
Select new image/video boxes when dropped from overlay tool (BL-11620)

StephenMcConnel:
This change is [Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5477)

### DIFF
--- a/src/BloomBrowserUI/bookEdit/js/bloomEditing.ts
+++ b/src/BloomBrowserUI/bookEdit/js/bloomEditing.ts
@@ -966,15 +966,36 @@ function focusLastEditableTopBox(): boolean {
         // equivalent to the :visible check we were previously doing in jquery
         s => window.getComputedStyle(s).getPropertyValue("display") != "none"
     );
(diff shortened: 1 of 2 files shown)


src/BloomBrowserUI/bookEdit/js/bloomEditing.ts
```diff
@@ -966,15 +966,36 @@ function focusLastEditableTopBox(): boolean {
...
+    }
+    const visibleVideoBoxesInLastTop = Array.from(
+        lastTop.getElementsByClassName("bloom-videoContainer")
+    ).filter(
+        s => window.getComputedStyle(s).getPropertyValue("display") != "none"
+    );
+    if (visibleVideoBoxesInLastTop.length !== 0) {
+        $(visibleVideoBoxesInLastTop[0] as HTMLElement).focus();
+        return true;
+    }
```
andrew-polk, line 997: Personally, I think it would have been an improvement to extract a function which passes in the class to filter on and call that three times, but not going to hold this up.

PR #5480, opened 2022-10-20
Tweak the fix for dropping images from overlay tool (BL-11620)

StephenMcConnel:
This fixes the side-effect of setting text background to transparent. It also extracts a method from some repeated code.

- - -
This change is [Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5480)

### DIFF
--- a/src/BloomBrowserUI/bookEdit/js/bloomEditing.ts
+++ b/src/BloomBrowserUI/bookEdit/js/bloomEditing.ts
@@ -959,43 +959,32 @@ function focusLastEditableTopBox(): boolean {
     const topBoxes = document.getElementsByClassName("bloom-textOverPicture");
     if (topBoxes.length == 0) return false;
     const lastTop = topBoxes[topBoxes.length - 1];
-    const visibleEditBoxesInLastTop = Array.from(
-        lastTop.getElementsByClassName("bloom-editable")
+    if (focusOnChildIfFound(lastTop, "bloom-editable")) return true;
+    // image and video boxes are also possibilities (BL-11620)
+    if (focusOnChildIfFound(lastTop, "bloom-imageContainer")) return true;
+    if (focusOnChildIfFound(lastTop, "bloom-videoContainer")) return true;
+    return false; // unexpected
+}
+
+function focusOnChildIfFound(lastTop: Element, className: string): boolean {
+    const visibleChildBoxesInLastTop = Array.from(
+        lastTop.getElementsByClassName(className)
     ).filter(
         // this is a crude check for visibility, but according to stack overflow
         // equivalent to the :visible check we were previously doing in jquery
         s => window.getComputedStyle(s).getPropertyValue("display") != "none"
     );
-    if (visibleEditBoxesInLastTop.length !== 0) {
+    if (visibleChildBoxesInLastTop.length !== 0) {
         // This doesn't work reliably if we just use the Element.focus(), so we use the JQuery version that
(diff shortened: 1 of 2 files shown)


JohnThomson (changes requested):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5480)* status: all files reviewed, 1 unresolved discussion (waiting on @StephenMcConnel)

___
*[`src/BloomBrowserUI/bookEdit/js/bubbleManager.ts` line 2124 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5480#-NEqXKaYBpSmf0fxoO54:-NEqXKaYBpSmf0fxoO55:b-xc256k) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/9862d42364bc917dc799ae96c6c426e77a6d59dd/src/BloomBrowserUI/bookEdit/js/bubbleManager.ts#L2124)):*
> ```XML
>         // overlay tool. But we don't need/want the actions of setActiveElement() which overlap
>         // with refreshBubbleEditing(). Text bubbles are handled properly elsewhere.
>         // See https://issues.bloomlibrary.org/youtrack/issue/BL-11620.
> ```

I may be being picky here, but as I reviewed this, my main thought was, I have no idea why this change helps stop bubbles' backgrounds being made transparent. And if I came back to this code in a year, I think I would be asking "WHERE are text bubbles handled?" and "WHY should making text bubbles active be handled differently?" The reference to the issue helps identify what might go wrong if you handle them here as well, but while you have the information in your head, could we capture your understanding of what's going on? Especially the 'why'.

StephenMcConnel (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5480)* status: all files reviewed, 1 unresolved discussion (waiting on @JohnThomson)

___
*[`src/BloomBrowserUI/bookEdit/js/bubbleManager.ts` line 2124 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5480#-NEqXKaYBpSmf0fxoO54:-NEqr-n_D5Ov1zWdwRz6:bi1op3j) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/9862d42364bc917dc799ae96c6c426e77a6d59dd/src/BloomBrowserUI/bookEdit/js/bubbleManager.ts#L2124)):*

I may be being picky here, but as I reviewed this, my main thought was, I have no idea why this change helps stop bubbles' backgrounds being made transparent. And if I came back to this code in a year, I think I would be asking "WHERE are text bubbles handled?" and "WHY should making text bubbles active be handled differently?" The reference to the issue helps identify what might go wrong if you handle them here as well, but while you have the information in your head, could we capture your understanding of what's going on? Especially the 'why'.
</blockquote></details>

I have no idea why the previous change made text bubbles' background transparent, but not touching the handling of text bubbles restored the previous (good) behavior.

Once a text bubble drops, the focusin handler set by addEventsToFocusableElements() is fired and sets everything up properly.  Through tracing, I can see that method being applied to the picture bubble when it drops, but the focusin event doesn't fire.  This is the crux of why this code is needed.  Of course, this code also results in the bubble retaining focus from before the handler is attached, so the fact the handler never fires becomes a moot point.
I've expanded the comment in the code to explain some what I'm seeing, but WHY the program is being so perverse in not calling an event handler, I have no idea.

JohnThomson (approved):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5480)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @StephenMcConnel)

PR #5492, opened 2022-10-25
Tweak harvester font analytics to trim unwanted items (BL-11512)

StephenMcConnel:
This change is [Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5492)

### DIFF
--- a/src/BloomExe/CLI/CreateArtifactsCommand.cs
+++ b/src/BloomExe/CLI/CreateArtifactsCommand.cs
@@ -122,12 +122,6 @@ namespace Bloom.CLI
 			{
 				exitCode |= CreateBloomDigitalArtifacts(parameters.BookPath, parameters.Creator, zippedBloomPubOutputPath, unzippedBloomDigitalOutputPath);
 			}
-			if (exitCode == CreateArtifactsExitCode.Success && BloomPubMaker.BloomPubFontsAndLangsUsed != null && !parameters.NoAnalytics)
-			{
-				// Report what we can about fonts and languages for this book.
-				// (See https://issues.bloomlibrary.org/youtrack/issue/BL-11512.)
-				SendFontAnalyticsCommand.ReportFontAnalytics(_book.ID, "harvester createArtifacts", parameters.Testing);
-			}
 			if (!String.IsNullOrEmpty(zippedBloomSourceOutputPath))
 			{
 				exitCode |= CreateBloomSourceArtifact(parameters.BookPath, parameters.Creator, zippedBloomSourceOutputPath);
@@ -165,6 +159,13 @@ namespace Bloom.CLI
 
 			CreateThumbnailArtifact(parameters);
 
+			if (exitCode == CreateArtifactsExitCode.Success && BloomPubMaker.BloomPubFontsAndLangsUsed != null && !parameters.NoAnalytics)
+			{
+				// Report what we can about fonts and languages for this book.
+				// (See https://issues.bloomlibrary.org/youtrack/issue/BL-11512.)
+				SendFontAnalyticsCommand.ReportFontAnalytics(_book.ID, "harvester createArtifacts", parameters.Testing,
+					!RobustFile.Exists(parameters.EpubOutputPath) || parameters.SkipEpubAnalytics, parameters.SkipPdfAnalytics);
+			}
 			return exitCode;
 		}
 
@@ -381,5 +382,11 @@ namespace Bloom.CLI
 
 		[Option("noAnalytics", Required = false, Default = false, HelpText = "Do not send any analytics")]
 		public bool NoAnalytics { get; set; }
+
+		[Option("skipEpubAnalytics", Required = false, Default = false, HelpText ="Do not send analytics for the Epub")]
+		public bool SkipEpubAnalytics { get; set; }
+
+		[Option("skipPdfAnalytics", Required = false, Default = false, HelpText ="Do not send analytics for the PDF")]
+		public bool SkipPdfAnalytics { get; set; }
 	}
 }
--- a/src/BloomExe/CLI/SendFontAnalyticsCommand.cs
+++ b/src/BloomExe/CLI/SendFontAnalyticsCommand.cs
@@ -7,6 +7,7 @@ using Bloom.Book;
 using BloomTemp;
 using Bloom.Publish.Android;
 using Bloom.ToPalaso;
+using System.Linq;
 
 namespace Bloom.CLI
 {
@@ -23,7 +24,7 @@ namespace Bloom.CLI
 	public class SendFontAnalyticsCommand
 	{
 		static ProjectContext _projectContext;
-		static string _bookID;
+		static Book.Book _book;
 
 		public static int Handle(SendFontAnalyticsParameters options)
 		{
@@ -44,7 +45,8 @@ namespace Bloom.CLI
 						{
 							// Report what we can about fonts and languages for this book.
 							// (See https://issues.bloomlibrary.org/youtrack/issue/BL-11512.)
-							ReportFontAnalytics(_bookID, "harvester sendFontAnalytics", options.Testing);
+							ReportFontAnalytics(_book.ID, "harvester sendFontAnalytics", options.Testing,
+								options.SkipEpubAnalytics, options.SkipPdfAnalytics);
 							return (int)SendFontAnalyticsExitCode.Success;
 						}
 					}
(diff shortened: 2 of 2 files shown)


andrew-polk (changes requested):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5492)* status: all files reviewed, 2 unresolved discussions (waiting on @StephenMcConnel)

___
*[`src/BloomExe/CLI/SendFontAnalyticsCommand.cs` line 113 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5492#-NFFQ-It9iz3JbpP5GCT:-NFFQ-It9iz3JbpP5GCU:b2f0v11) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/b811784193153c80ccb12ef9b5c61dbb0444443a/src/BloomExe/CLI/SendFontAnalyticsCommand.cs#L113)):*
> ```Smalltalk
> 		}
> 
> 		private static float GetBloomVersionFromBook(Book.Book book)
> ```

HtmlDom has a GetGeneratorVersion method.
___
*[`src/BloomExe/CLI/SendFontAnalyticsCommand.cs` line 135 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5492#-NFFQk5t9Y_6yip3098L:-NFFQk5t9Y_6yip3098M:b5xljoi) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/b811784193153c80ccb12ef9b5c61dbb0444443a/src/BloomExe/CLI/SendFontAnalyticsCommand.cs#L135)):*
> ```Smalltalk
> 		private static bool WouldEpubBeValidForPublishing(Book.Book book, float bloomVersion)
> 		{
> 			// This logic is copied from BloomHarvester.
> ```

How do we prevent that?
Does Bloom need to be the source of truth for whether an epub should be published and that information is passed back to Harvester when the epub is created?

Related: why do we determine some things here but pass in whether epub was created as a flag?

Also related: why do we pass whether an epub was created but determine programmatically if a PDF should be?



_Code quote:_
```Smalltalk
// This logic is copied from BloomHarvester.
```

StephenMcConnel (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5492)* status: 0 of 2 files reviewed, 2 unresolved discussions (waiting on @andrew-polk)

___
*[`src/BloomExe/CLI/SendFontAnalyticsCommand.cs` line 113 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5492#-NFFQ-It9iz3JbpP5GCT:-NFFzmUX4YBCjG1y-lEz:bc2kger) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/b811784193153c80ccb12ef9b5c61dbb0444443a/src/BloomExe/CLI/SendFontAnalyticsCommand.cs#L113)):*

HtmlDom has a GetGeneratorVersion method.
</blockquote></details>

Who knew?  :-)
___
*[`src/BloomExe/CLI/SendFontAnalyticsCommand.cs` line 135 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5492#-NFFQk5t9Y_6yip3098L:-NFFywC0AQj9MFAf2vLl:b8ust58) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/b811784193153c80ccb12ef9b5c61dbb0444443a/src/BloomExe/CLI/SendFontAnalyticsCommand.cs#L135)):*

How do we prevent that?
Does Bloom need to be the source of truth for whether an epub should be published and that information is passed back to Harvester when the epub is created?

Related: why do we determine some things here but pass in whether epub was created as a flag?

Also related: why do we pass whether an epub was created but determine programmatically if a PDF should be?
</blockquote></details>

I've changed to using a skipFontAnalytics option on both createArtifacts and sendFontAnalytics to pass on the information available in the harvester code as
to whether an epub artifact would be suitable.

The sendFontAnalytics command does not have access to the generated artifacts while the createArtifacts command (which shares the code for actually sending font analytics) does have access on the disk to know what was created, so it can't easily check whether an epub artifact actually was created.

The harvester never creates PDFs.  They're either created at upload or not.  The logic here follows what we have done for deciding whether PDFs are created and uploaded.

andrew-polk (changes requested):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5492)* status: all files reviewed, 1 unresolved discussion (waiting on @StephenMcConnel)

___
*[`src/BloomExe/CLI/SendFontAnalyticsCommand.cs` line 135 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5492#-NFFQk5t9Y_6yip3098L:-NFGCNzt6fdkjlA-Ofc5:b1nk88o) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/b811784193153c80ccb12ef9b5c61dbb0444443a/src/BloomExe/CLI/SendFontAnalyticsCommand.cs#L135)):*

I've changed to using a skipFontAnalytics option on both createArtifacts and sendFontAnalytics to pass on the information available in the harvester code as
to whether an epub artifact would be suitable.

The sendFontAnalytics command does not have access to the generated artifacts while the createArtifacts command (which shares the code for actually sending font analytics) does have access on the disk to know what was created, so it can't easily check whether an epub artifact actually was created.

The harvester never creates PDFs.  They're either created at upload or not.  The logic here follows what we have done for deciding whether PDFs are created and uploaded.
</blockquote></details>

Thanks.

Perhaps I'm overly paranoid about such things, but bugs we never see because we prevent them are the best kind! : )
Here's the kind of thing I'm thinking about...
What if we change the logic for when we produce PDFs?
I was going to say we should just extract that logic to a common place and call it from both... but, it is actually more complicated than that. Because at that point, we actually have to have logic which says, if it is for this version, determine this way, else that way. So that would point toward just a comment in the upload code which states we must change this if we change that logic.

But all that points to it being mucho mucho better if we can actually determine if a pdf exists.
Since we have the original book downloaded on the machine, can we just search the book folder for a PDF?
Or what creates/updates the `show` field in parse with the information about whether a pdf exists?

StephenMcConnel (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5492)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @andrew-polk)

___
*[`src/BloomExe/CLI/SendFontAnalyticsCommand.cs` line 135 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5492#-NFFQk5t9Y_6yip3098L:-NFK2oaJ4QaBZ5kJMLSz:b-202aie) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/b811784193153c80ccb12ef9b5c61dbb0444443a/src/BloomExe/CLI/SendFontAnalyticsCommand.cs#L135)):*

Thanks.

Perhaps I'm overly paranoid about such things, but bugs we never see because we prevent them are the best kind! : )
Here's the kind of thing I'm thinking about...
What if we change the logic for when we produce PDFs?
I was going to say we should just extract that logic to a common place and call it from both... but, it is actually more complicated than that. Because at that point, we actually have to have logic which says, if it is for this version, determine this way, else that way. So that would point toward just a comment in the upload code which states we must change this if we change that logic.

But all that points to it being mucho mucho better if we can actually determine if a pdf exists.
Since we have the original book downloaded on the machine, can we just search the book folder for a PDF?
Or what creates/updates the `show` field in parse with the information about whether a pdf exists?
</blockquote></details>

The PDF file is not included in the download of the book sources, so we can't just look for the file, but we can check S3 for
whether the PDF file exists.  (I think harvester already does this somewhere.)
I've added command line options for createArtifacts and sendFontAnalytics to optionally skip PDF analytics.

StephenMcConnel (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5492)* status: 0 of 4 files reviewed, 1 unresolved discussion (waiting on @andrew-polk)

___
*[`src/BloomExe/CLI/SendFontAnalyticsCommand.cs` line 135 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5492#-NFFQk5t9Y_6yip3098L:-NFL-u65CsSQn_sd37Q-:bpxz6ae) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/b811784193153c80ccb12ef9b5c61dbb0444443a/src/BloomExe/CLI/SendFontAnalyticsCommand.cs#L135)):*

The PDF file is not included in the download of the book sources, so we can't just look for the file.  Nothing automatically marks the PDF as existing.
I've added an output flag for BloomS3Client.DownloadBooks() to check for whether the PDF file exists, and command line options for createArtifacts and sendFontAnalytics to optionally skip PDF analytics.
</blockquote></details>

I didn't end up adding an output flag for BloomS3Client.DownloadBooks().  Harvester can check directly via S3 for the existence of the PDF file, which it already does in one place.

andrew-polk (approved):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5492)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @StephenMcConnel)

___
*[`src/BloomExe/CLI/SendFontAnalyticsCommand.cs` line 135 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5492#-NFFQk5t9Y_6yip3098L:-NFL3SSF0O8m1bHkFM8y:b-qy2rh) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/b811784193153c80ccb12ef9b5c61dbb0444443a/src/BloomExe/CLI/SendFontAnalyticsCommand.cs#L135)):*

I didn't end up adding an output flag for BloomS3Client.DownloadBooks().  Harvester can check directly via S3 for the existence of the PDF file, which it already does in one place.
</blockquote></details>

We do set the show.pdf property in Harvester.cs, around line 923.
But as I assume that is the place you mention when you say we check for its existence, and therefore the logic will be in one place, this looks good. 
Thanks.

I think this can be merged without the harvester changes being ready, but I'm not 100% sure, so I'll let you do the merge.

PR #5501, opened 2022-10-27
Fix PDF generation on Linux for full bleed books (BL-11685)

StephenMcConnel:
This looks safe for cherry-picking to 5.3.  The bug has been around since at least 5.0.

- - -
This change is [Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5501)


andrew-polk (approved):

andrew-polk:
Agree with cherry-pick.

PR #5502, opened 2022-10-28
Ctrl or Alt key causes image editing buttons to hide (BL-11638)

sujeffreyl:
This change is [Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5502)

### DIFF
--- a/src/BloomBrowserUI/bookEdit/css/editMode.less
+++ b/src/BloomBrowserUI/bookEdit/css/editMode.less
@@ -117,7 +117,8 @@ div.bloom-imageContainer {
 }
 
 // The editing buttons are hidden when this class is added (currently by motion and comic tools)
-.bloom-imageContainer.bloom-hideImageButtons {
+.bloom-imageContainer.bloom-hideImageButtons,
+.bloom-imageContainer.ui-suppressImageButtons {
     > .imageButton,
     > .miniButton {
         display: none;
--- a/src/BloomBrowserUI/bookEdit/js/bloomImages.ts
+++ b/src/BloomBrowserUI/bookEdit/js/bloomImages.ts
@@ -58,6 +58,25 @@ export function SetupImagesInContainer(container) {
                 });
         });
 
+    // Add the event listener to handle control/alt being pressed.
+    // We have a pretty interesting way of listening for these keydown/keyup events.
+    // I initially tried putting the event listeners on the bloom-imageContainer,
+    // but that didn't work reliably.
+    // Keyboard events don't fire for every element, but only certain valid ones.
+    // For the short answer, divs with contenteditable=true meet the criteria.
+    // (Although... I tested what happens with slapping contenteditable=true onto the bloom-imageContainer,
+    // and I don't think it helped)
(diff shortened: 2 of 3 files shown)


andrew-polk (changes requested):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5502)* status: all files reviewed, 1 unresolved discussion (waiting on @sujeffreyl)

___
*[`src/BloomBrowserUI/bookEdit/js/bloomImages.ts` line 294 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5502#-NFUruqA0SW9CBtZtKHv:-NFUruqA0SW9CBtZtKHw:b-89ep45) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/18aca9e9be724b0ab127b9875bc48fbd0cf93098/src/BloomBrowserUI/bookEdit/js/bloomImages.ts#L294)):*
> ```XML
>     const selector =
>         ".imageOverlayButton" +
>         (options.skipProblemIndicator ? "" : ":not(.imgMetadataProblem)");
> ```

It's Friday...
But isn't this logic backwards? Aren't we excluding it if skip is false?

sujeffreyl (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5502)* status: 2 of 3 files reviewed, 1 unresolved discussion (waiting on @andrew-polk)

___
*[`src/BloomBrowserUI/bookEdit/js/bloomImages.ts` line 294 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5502#-NFUruqA0SW9CBtZtKHv:-NFV6vo78jXL7h8UqaiY:bwz1m1w) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/18aca9e9be724b0ab127b9875bc48fbd0cf93098/src/BloomBrowserUI/bookEdit/js/bloomImages.ts#L294)):*

It's Friday...
But isn't this logic backwards? Aren't we excluding it if skip is false?
</blockquote></details>

Interesting catch. In one sense I think you're right. But a weird discovery / can of worms...
It actually doesn't matter one bit. The imgMetadataProblem button doesn't have the class "imageOverlayButton", so removeImageEditingButtons never considers it all. I think it's been this way for at least 4 years.

Was it missing the imageOverlayButton class to it? Well, that doesn't actually make any styling difference. imageOverlayButton is never used for styling... Honestly, it's only referenced a couple times too.

Why do we have this class at all? Actually, I feel like merging "imageOverlayButton" into "imageButton" would probably be ok. Well - almost. I tried it out, it almost works except the styling of the buttons which have only "imageButton" but not "imageOverlayButton" is slightly affected.

I decided to just forget that and just fix the order and write a comment.

andrew-polk (approved):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5502)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @sujeffreyl)

PR #5521, opened 2022-11-03
Extend fix for titles with linebreaks (BL-11717)

StephenMcConnel:
This change is [Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5521)

### DIFF
--- a/src/BloomExe/Book/Book.cs
+++ b/src/BloomExe/Book/Book.cs
@@ -13,6 +13,7 @@ using System.Threading.Tasks;
 using System.Web;
 using System.Windows.Forms;
 using System.Xml;
+using System.Xml.Linq;
 using Bloom.Api;
 using Bloom.Collection;
 using Bloom.Edit;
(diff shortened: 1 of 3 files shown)


sujeffreyl (changes requested):
Thank you for being brave and figuring out how to extract a function out of that! It looks like I was scared off last time by the XElement vs XmlElement thing.

> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5521#-:-NFzUkM6BDO8U5ByKhuH:b-37mhck)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @StephenMcConnel)

___
*[`src/BloomExe/Book/BookData.cs` line 1782 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5521#-NFzT_ZoDEcnPCeivGLg:-NFzT_ZoDEcnPCeivGLh:bsae46j) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/dfbc4e96b4c74a211fd2af69a71303caad18ddf9/src/BloomExe/Book/BookData.cs#L1782)):*
> ```Smalltalk
> 		}
> 
> 		public static string RemoveHtmlMarkup(string input, string replacementForLinebreakSpan)
> ```

I think this function should take in LineBreakSpanConversionMode instead.
(At which point, maybe this function should live in Book.cs instead? Or even a different *Utils class entirely? but that's optional)
Then, BookData's function wouldn't need that awkward try { if (...) doAlgorithm1(); else doAlgorithm2(); } thing. It could all be processed using a call to RemoveHtmlMarkup

StephenMcConnel (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5521)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @sujeffreyl)

___
*[`src/BloomExe/Book/BookData.cs` line 1782 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5521#-NFzT_ZoDEcnPCeivGLg:-NFzX9BB5tfB_nwaJWQ5:bqg2544) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/dfbc4e96b4c74a211fd2af69a71303caad18ddf9/src/BloomExe/Book/BookData.cs#L1782)):*

I think this function should take in LineBreakSpanConversionMode instead.
(At which point, maybe this function should live in Book.cs instead? Or even a different *Utils class entirely? but that's optional)
Then, BookData's function wouldn't need that awkward try { if (...) doAlgorithm1(); else doAlgorithm2(); } thing. It could all be processed using a call to RemoveHtmlMarkup
</blockquote></details>

I passed in a string because in the BookData function, it replaced the span with a simple "\n", while in the Book function, it replaced it with either a space or an Environment.NewLine, which can be either "\n" or "\r\n".  I thought the deterministic behavior for what goes into meta.json regardless of source was worth preserving.
Looking at the possible values of LineBreakConversionMode, and where Book's function is called, I don't really see the need for testing its value and doing something different if it's neither of the two possible valid values.  So I've moved the actual function to Book, extended the LineBreakConversionMode to allow a simple "\n", and changed the function to use the enum as an argument.  I kept the try/catch in the Book function because it seems the right thing to do.

sujeffreyl (commented):
> Reviewed 2 of 3 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5521)* status: 2 of 3 files reviewed, all discussions resolved

sujeffreyl (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5521#-:-NFziC9e-IhT8wKMPiXF:bnfp4nl)* status: 2 of 3 files reviewed, all discussions resolved

sujeffreyl (approved):
> Reviewed 1 of 3 files at r2.
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5521)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @StephenMcConnel)

PR #5522, opened 2022-11-04
Re-do fix for titles with linebreaks (BL-11717)

sujeffreyl:
Addresses the failing unit test BloomTests.Book.BookInfoTests.TitleSetter_FixesTitleWithXml. The \r\n is handled different by XmlDocument vs XElement. (Also, that path was previously not trimmed)

- - -
This change is [Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5522)

### DIFF
--- a/src/BloomExe/Book/Book.cs
+++ b/src/BloomExe/Book/Book.cs
@@ -319,55 +319,64 @@ namespace Bloom.Book
 			// Handle both Windows and Linux line endings in case a file copied between the two
 			// ends up with the wrong one.
 			display = display.Replace("<br />", " ").Replace("\r\n", " ").Replace("\n", " ").Replace("  ", " ");
-			display = RemoveXmlMarkup(display, LineBreakSpanConversionMode.ToSpace).Trim();
+			display = RemoveHtmlMarkup(display, LineBreakSpanConversionMode.ToSpace).Trim();
 			return display;
 		}
 
 		public enum LineBreakSpanConversionMode
 		{
-			ToNewline,
-			ToSpace
+			ToNewline,		// Environment.Newline
+			ToSpace,		// " "
+			ToSimpleNewline	// "\n", used for export to meta.json
 		}
 
-		// might be better named RemoveMarkup; what can be in here is *HTML*, e.g. bold, italics, etc.
-		public static string RemoveXmlMarkup(string input, LineBreakSpanConversionMode lineBreakSpanConversionOptions)
+		// what can be in here is *XHTML*, e.g. bold, italics, etc.
+		public static string RemoveHtmlMarkup(string input, LineBreakSpanConversionMode lineBreakSpanConversionOption)
 		{
+			if (input == null)
+				return null;
 			try
 			{
 				var doc = new XmlDocument();
 				doc.PreserveWhitespace = true;
 				doc.LoadXml("<div>" + input + "</div>");
 
-				// NOTE: There is a similar section of code dealing with bloom-linebreak
-				//       in BookData.cs::TextOfInnerHtml().
-				//       (That operates on XElements though, this deals with XmlElements)
-				if (lineBreakSpanConversionOptions == LineBreakSpanConversionMode.ToSpace || lineBreakSpanConversionOptions == LineBreakSpanConversionMode.ToNewline)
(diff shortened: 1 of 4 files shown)


sujeffreyl (commented):
This might be a little confusing without looking at a bunch of things in Git.

Originally, we had functions in both Book and BookData. They were very similar, but they hadn't been combined in the past mainly because one used XMLDocument and one used XElement.
Round 1 fix: Fixes the BookData function (but not the Book function)
Round 2 fix: Steve attempts to combine the two. The Book function is now basically running the code from the BookData function.
However, XMLDocument and XElement treat \r\n differently, so that is 1 cause of the failing unit test.  (The other cause is that we now apply .Trim() to both, whereas we previously only apply it in the BookData scenario. That's easy to fix though)
Round 3 fix: I do it the other way around... still combined, but using the Book function as the base for both instead of BookData as the base for both. There's still some risk in how the 2 might Parse/Load documents differently... but hopefully it doesn't really matter.  (IDK. BookData mostly uses XElement, but there is one reference to XmlDocument)

The other idea is to give up yet again on fully combining the 2 functions. But I feel like we're as close as we'll ever be to pulling it off.

> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5522#-:-NG2TeW02jgNjwGYQxss:bzbblfw)* status: 0 of 2 files reviewed, all discussions resolved

andrew-polk:
Marking as draft pending discussion in the card.

sujeffreyl (commented):
I merged this work with the less-risky version of it in 5.4. I changed this PR to target master now.

> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5522#-:-NGd1zpJEpOTIEA-NNjF:b-v1iywv)* status: 0 of 4 files reviewed, all discussions resolved

JohnThomson (approved):
> Reviewed 2 of 2 files at r1, 2 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5522)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @sujeffreyl)

JohnThomson (changes requested):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5522)* status: all files reviewed, 1 unresolved discussion (waiting on @sujeffreyl)

___
*[`src/BloomTests/Book/BookTests.cs` line 4633 at r2](https://reviewable.io/reviews/bloombooks/bloomdesktop/5522#-NHBXspj65NCYhVXLqR5:-NHBXspk6a_wJcagVg8b:b-gk28ly) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/6e58a2f9f9996f613a0b30ac573ae4febf082db6/src/BloomTests/Book/BookTests.cs#L4633)):*
> ```Smalltalk
> 			
> 			Assert.That(result, Is.EqualTo("Enter Shift-Enter Last Line "));
> 		}
> ```

Seems like we should have at least one unit test for each of the (now three) enumeration members. It might work to just change the param and output of two of these.

sujeffreyl (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5522)* status: 3 of 4 files reviewed, 1 unresolved discussion (waiting on @JohnThomson)

___
*[`src/BloomTests/Book/BookTests.cs` line 4633 at r2](https://reviewable.io/reviews/bloombooks/bloomdesktop/5522#-NHBXspj65NCYhVXLqR5:-NHPuPkcE9t9BmKgq6QO:b-896fix) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/6e58a2f9f9996f613a0b30ac573ae4febf082db6/src/BloomTests/Book/BookTests.cs#L4633)):*

Seems like we should have at least one unit test for each of the (now three) enumeration members. It might work to just change the param and output of two of these.
</blockquote></details>

Done.

JohnThomson (approved):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5522)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @sujeffreyl)

PR #5528, opened 2022-11-08
Don't highlight correct answer in PDF output (BL-11726)

StephenMcConnel:
The first answer is always the correct answer anyway.  Solving this would be much more complicated.  bloom-player uses the .player-shuffle-buttons to apply javascript to shuffle the order.  PDF output doesn't have any intelligence along those lines.

- - -
This change is [Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5528)

### DIFF
--- a/src/BloomExe/Book/Book.cs
+++ b/src/BloomExe/Book/Book.cs
@@ -1285,6 +1285,7 @@ namespace Bloom.Book
 			}
 			RemoveObsoleteSoundAttributes(bookDOM);
 			RemoveObsoleteImageAttributes(bookDOM);
+			RemoveUnwantedClasses(bookDOM);
 			BringBookInfoUpToDate(oldMetaData);
 			FixErrorsEncounteredByUsers(bookDOM);
 			AddReaderBodyAttributes(bookDOM);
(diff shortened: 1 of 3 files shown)


sujeffreyl (changes requested):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5528)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @StephenMcConnel)

___
*[`src/content/templates/template books/Activity/simple-dom-choice-activities.less` line 263 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5528#-NGMpFx9DB9eU3As1lux:-NGMpFx9DB9eU3As1luy:b-equiu1) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/741d10492f793309ec567eccfffe3de7d63ec24a/src/content/templates/template%20books/Activity/simple-dom-choice-activities.less#L263)):*
> ```Less
>         }
>         // highlight correct answer while editing
>         &[data-activityRole="correct-answer"] {
> ```

Would it work to just add this as a second condition (OR'd) to the rule above by using ","?

Two of the declarations are the same, and the third one is a few lines above, I think it should be applied already too

JohnThomson (commented):
I really think we need some sort of shuffle. See comment in issue.

> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5528#-:-NGMq_Zb7fRd0SS0zDax:b-er6n79)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @StephenMcConnel and @sujeffreyl)

StephenMcConnel:
This pull request has been withdrawn.  Work on this issue will be postponed until 5.5.

PR #5535, opened 2022-11-08
Ensure the tooltip for the active field shows on top (BL-11745)

StephenMcConnel:
This change is [Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5535)

### DIFF
--- a/src/BloomBrowserUI/bookEdit/bloomField/BloomField.ts
+++ b/src/BloomBrowserUI/bookEdit/bloomField/BloomField.ts
@@ -185,6 +185,46 @@ export default class BloomField {
             $(".removeMe").remove();
         });
 
+        // The focus and blur event handlers ensure that the qtip tooltip for the
+        // currently focused editor element is on top of any other qtip tooltips
+        // such as Source Bubbles.
+        // These tooltips are not to be confused with Source Bubbles.  The editor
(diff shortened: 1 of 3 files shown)


StephenMcConnel (commented):
One problem is that BloomSourceBubbles only handles *Source Bubbles* -- it doesn't handle normal tooltip bubbles that merely describe the text box.  A second problem is that both the L1 and L2 title text boxes are associated with the same div.bloom-translationGroup which has the assigned qtip-1 containing the translation choices.  So setting the focus in the L2 title textbox doesn't affect anything as far as the Source Bubble is concerned.  I don't know if you'd call this a design defect or a design limitation.  It's a special corner case that I don't think happens anywhere else.

I've updated the fix to use adding/removing classes as its base, and added cross references between BloomField and BloomSourceBubbles.

> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5535#-:-NGSFnwx824xmU4JRelc:bnlt4x9)* status: 0 of 3 files reviewed, all discussions resolved (waiting on @JohnThomson)

JohnThomson (approved):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5535)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @StephenMcConnel)

PR #5540, opened 2022-11-10
Don't make white transparent on covers unless B&W image (BL-11708)

StephenMcConnel:
This PR is currently on Version5.4.

- - -
This change is [Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5540)


StephenMcConnel (commented):
Checking 100 pixels may be overkill.  While testing, detecting a third color generally occurred within the first 20 pixels.  So the slowest images to check by far were non-indexed black and white images.

> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5540#-:-NGTtbK94jXCJWR9wPvB:b-wm3yxl)* status: 0 of 2 files reviewed, all discussions resolved

hatton (commented):
Can we get a unit test to show the problem?

> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5540#-:-NGWp916BGQpEr7LDYxd:b-dblekb)* status: 0 of 2 files reviewed, all discussions resolved

JohnThomson (commented):
Looks basically good. A few suggestions, which may not be helpful.

> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5540#-:-NGYfcwSAtUHDGDi7WR-:b2yvlhe)* status: 0 of 2 files reviewed, 4 unresolved discussions (waiting on @StephenMcConnel)

___
*[`src/BloomExe/Book/Book.cs` line 4167 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5540#-NGYbe7Y-1dJ6QehQX2W:-NGYbe7Z7Ak3Q8fhQcHi:bbkr891) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/4e843b67d1f2f2b6a16a37e9b2a24ef026998fcc/src/BloomExe/Book/Book.cs#L4167)):*
> ```Smalltalk
> 				if (new FileInfo(imagePath).Length != new FileInfo(coverImagePath).Length)
> 					return false;
> 				using (var image = PalasoImage.FromFileRobustly(imagePath))
> ```

Is it worth first testing IsJpegFile? We can do that without reading the whole image from disk.
___
*[`src/BloomExe/ImageProcessing/ImageUtils.cs` line 88 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5540#-NGYaRj14VjnhDaz28zn:-NGYaRj14VjnhDaz28zo:b-98ilvs) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/4e843b67d1f2f2b6a16a37e9b2a24ef026998fcc/src/BloomExe/ImageProcessing/ImageUtils.cs#L88)):*
> ```Smalltalk
> 				var color2 = new Color();
> 				// Yes, this is as expensive as it looks.  But we only sample 100 pixels
> 				// spread through the picture, stopping as soon as we hit 3 colors.
> ```

Actually the need to read the entire image from disk into memory is likely to be MUCH more expensive than checking 100 pixels. We probably have to do that anyway, but is there any reasonable way to refactor so that we make this check AFTER we've already read it in, so we only do that once?
___
*[`src/BloomExe/ImageProcessing/ImageUtils.cs` line 90 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5540#-NGYd9hs5F6ZVw2KZnfn:-NGYd9hs5F6ZVw2KZnfo:b-jcdqap) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/4e843b67d1f2f2b6a16a37e9b2a24ef026998fcc/src/BloomExe/ImageProcessing/ImageUtils.cs#L90)):*
> ```Smalltalk
> 				// spread through the picture, stopping as soon as we hit 3 colors.
> 				int yDelta = Math.Max(bitmapImage.Height / 10, 6);
> 				int xDelta = Math.Max(bitmapImage.Width / 10, 6);
> ```

Why a max of 6? If the image happens to be a 16x16px icon, we'll only sample 9 points. You need to make sure it doesn't come out zero, of course.
___
*[`src/BloomExe/ImageProcessing/ImageUtils.cs` line 94 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5540#-NGYf2yIA6BuxaYN6piq:-NGYf2yIA6BuxaYN6pir:ba9xe8h) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/4e843b67d1f2f2b6a16a37e9b2a24ef026998fcc/src/BloomExe/ImageProcessing/ImageUtils.cs#L94)):*
> ```Smalltalk
> 				{
> 					for (int x = xDelta / 2; x < bitmapImage.Width; x += xDelta)
> 					{
> ```

What would you think of here subtracting from each coordinate a random number between 0 and half of the appropriate delta? Just so we aren't limiting ourselves to 10 rows and columns.

StephenMcConnel (commented):
I've added several images and written a unit test.

> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5540#-:-NGYkXxJ85CRVbXnAUP5:b-1l7swp)* status: 0 of 20 files reviewed, 4 unresolved discussions (waiting on @JohnThomson)

___
*[`src/BloomExe/Book/Book.cs` line 4167 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5540#-NGYbe7Y-1dJ6QehQX2W:-NGYk8dY2LVLn5Y1DYmz:b-a2nb66) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/4e843b67d1f2f2b6a16a37e9b2a24ef026998fcc/src/BloomExe/Book/Book.cs#L4167)):*

Is it worth first testing IsJpegFile? We can do that without reading the whole image from disk.
</blockquote></details>

The refactoring removed the need for this.
___
*[`src/BloomExe/ImageProcessing/ImageUtils.cs` line 88 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5540#-NGYaRj14VjnhDaz28zn:-NGYnkzV-aW8srLY_OqO:b-896fix) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/4e843b67d1f2f2b6a16a37e9b2a24ef026998fcc/src/BloomExe/ImageProcessing/ImageUtils.cs#L88)):*

Actually the need to read the entire image from disk into memory is likely to be MUCH more expensive than checking 100 pixels. We probably have to do that anyway, but is there any reasonable way to refactor so that we make this check AFTER we've already read it in, so we only do that once?
</blockquote></details>

Done.
___
*[`src/BloomExe/ImageProcessing/ImageUtils.cs` line 90 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5540#-NGYd9hs5F6ZVw2KZnfn:-NGZ0atVCbO-Z_NQJeUJ:b-mlrosg) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/4e843b67d1f2f2b6a16a37e9b2a24ef026998fcc/src/BloomExe/ImageProcessing/ImageUtils.cs#L90)):*

Why a max of 6? If the image happens to be a 16x16px icon, we'll only sample 9 points. You need to make sure it doesn't come out zero, of course.
</blockquote></details>

changed to a minimum value of 2 instead of 6.
___
*[`src/BloomExe/ImageProcessing/ImageUtils.cs` line 94 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5540#-NGYf2yIA6BuxaYN6piq:-NGZ0jj66BF7eSQVGNhx:bd7po91) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/4e843b67d1f2f2b6a16a37e9b2a24ef026998fcc/src/BloomExe/ImageProcessing/ImageUtils.cs#L94)):*

What would you think of here subtracting from each coordinate a random number between 0 and half of the appropriate delta? Just so we aren't limiting ourselves to 10 rows and columns.
</blockquote></details>

I'm not sure of the utility of randomizing the sample, but could test it next week.  It might help in some perverse cases, but makes the check non-repeatable as well.

JohnThomson (changes requested):
> Reviewed 20 of 20 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5540)* status: all files reviewed, 2 unresolved discussions (waiting on @StephenMcConnel)

___
*[`src/BloomExe/ImageProcessing/ImageUtils.cs` line 90 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5540#-NGYd9hs5F6ZVw2KZnfn:-NGclH9r9fXEF80stEnM:b-hjkczo) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/4e843b67d1f2f2b6a16a37e9b2a24ef026998fcc/src/BloomExe/ImageProcessing/ImageUtils.cs#L90)):*

changed to a minimum value of 2 instead of 6.
</blockquote></details>

I'm not seeing that change. Unfortunately, making it may invalidate some of your comments about how many pixels were tried before finding 3 colors.
___
*[`src/BloomExe/ImageProcessing/ImageUtils.cs` line 94 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5540#-NGYf2yIA6BuxaYN6piq:-NGclUkE-mPwT0xNthzS:b-j0ewkg) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/4e843b67d1f2f2b6a16a37e9b2a24ef026998fcc/src/BloomExe/ImageProcessing/ImageUtils.cs#L94)):*

I'm not sure of the utility of randomizing the sample, but could test it next week.  It might help in some perverse cases, but makes the check non-repeatable as well.
</blockquote></details>

You could make it repeatable by seeding the random number generator with some predictable value, like the width of the image, or even a permanently fixed seed. But I'm not sure it's worth it. It would only be a bizarre special case where it would help, and who is to say that there wouldn't be other bizarre cases where it makes things worse? It might be interesting to see what it did to your test cases, though at best it will probably invalidate some of your comments about how many pixels were tried before finding 3 colors.
Actually, as a curiosity, it would be interesting to see how much variation you got in those counts with different seeds. That would give more idea of whether it's really better or just got lucky.

StephenMcConnel (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5540)* status: 16 of 21 files reviewed, 2 unresolved discussions (waiting on @JohnThomson)

___
*[`src/BloomExe/ImageProcessing/ImageUtils.cs` line 90 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5540#-NGYd9hs5F6ZVw2KZnfn:-NGsK1cF1j2pSHd4V7QI:b-iroi56) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/4e843b67d1f2f2b6a16a37e9b2a24ef026998fcc/src/BloomExe/ImageProcessing/ImageUtils.cs#L90)):*

I'm not seeing that change. Unfortunately, making it may invalidate some of your comments about how many pixels were tried before finding 3 colors.
</blockquote></details>

I hadn't yet pushed any changes pending more work today.
___
*[`src/BloomExe/ImageProcessing/ImageUtils.cs` line 94 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5540#-NGYf2yIA6BuxaYN6piq:-NGsIn-OAMNHoxE3INKG:b-tex1ue) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/4e843b67d1f2f2b6a16a37e9b2a24ef026998fcc/src/BloomExe/ImageProcessing/ImageUtils.cs#L94)):*

You could make it repeatable by seeding the random number generator with some predictable value, like the width of the image, or even a permanently fixed seed. But I'm not sure it's worth it. It would only be a bizarre special case where it would help, and who is to say that there wouldn't be other bizarre cases where it makes things worse? It might be interesting to see what it did to your test cases, though at best it will probably invalidate some of your comments about how many pixels were tried before finding 3 colors.
Actually, as a curiosity, it would be interesting to see how much variation you got in those counts with different seeds. That would give more idea of whether it's really better or just got lucky.
</blockquote></details>

I've added a randomization with a fixed seed.  I measured what happened with no randomization, and seeds of 123456, 654321, 314159265, and 271828182.  I settled on using the last of those seeds.  With 18 files and 5 setups, only 1 file for 1 setup gave an incorrect answer and that was for a line drawing with probably more than 95% of the pixels being white or black.  I think most realistic drawings should be handled okay.

JohnThomson (changes requested):
Looks good. Thanks for you patience with my suggestion, which seems to have been more work and less help than I hoped.
My only remaining suggestion is that it looks like you could clean out some unused code and comments.

> Reviewed 5 of 5 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5540#-:-NGwNbzY0idaVRzC1RTg:bipgjre)* status: all files reviewed, 1 unresolved discussion (waiting on @StephenMcConnel)

___
*[`src/BloomExe/ImageProcessing/ImageUtils.cs` line 114 at r3](https://reviewable.io/reviews/bloombooks/bloomdesktop/5540#-NGwNEEk3zBW3wVptk5Q:-NGwNEEk3zBW3wVptk5R:b-10sira) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/c4359a6ace5a2075968d0e966dba1979b8b886f9/src/BloomExe/ImageProcessing/ImageUtils.cs#L114)):*
> ```Smalltalk
> 				var randomYFix = GenerateRandomAdjustments(271828182, yDelta);
> 				var pixelsExamined = 0;
> 				var grays = new HashSet<Color>();
> ```

This doesn't seem to be used anywhere (except some code that is commented out). Is there a reason to keep either the commented or uncommented code that uses it?

JohnThomson (approved):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5540)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @StephenMcConnel)

PR #5544, opened 2022-11-11
Fix a few resize handle problems (BL-11738)

sujeffreyl:
This change is [Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5544)

### DIFF
--- a/src/BloomBrowserUI/bookEdit/js/bloomImages.ts
+++ b/src/BloomBrowserUI/bookEdit/js/bloomImages.ts
@@ -111,23 +111,18 @@ function ctrlAltKeyDownListener(event: KeyboardEvent) {
     if ((event.ctrlKey || event.altKey) && event.target) {
         event.target.addEventListener("keyup", ctrlAltKeyUpListener);
 
-        // Note (paranoia): Add this class conservatively (ensure event.target is good first),
+        // Note (paranoia): Add the ui-suppressImageButtons class conservatively (ensure event.target is good (non-null) first),
         // remove it liberally (regardless of event.target),
         // Since if it gets stuck on, the image editing buttons won't come back (unless SetupImageContainer is re-run)
         document
-            .querySelectorAll(
-                ".bloom-imageContainer:not(.ui-suppressImageButtons)"
-            )
+            .querySelectorAll<HTMLElement>(".bloom-imageContainer")
             .forEach(imageContainer => {
                 imageContainer.classList.add("ui-suppressImageButtons");
 
                 if (event.ctrlKey) {
                     imageContainer.classList.add("ui-ctrlDown");
-                } else if (
-                    event.altKey &&
-                    imageContainer instanceof HTMLElement
-                ) {
-                    theOneBubbleManager.addResizeModeClass(imageContainer, {});
+                } else if (event.altKey) {
+                    theOneBubbleManager.tryApplyResizingUI(imageContainer);
                 }
             });
     }
--- a/src/BloomBrowserUI/bookEdit/js/bloomImages.ts
+++ b/src/BloomBrowserUI/bookEdit/js/bloomImages.ts
@@ -5,7 +5,7 @@ import { BloomApi } from "../../utils/bloomApi";
 
 import theOneLocalizationManager from "../../lib/localizationManager/localizationManager";
 
-import { updateOverlayClass } from "./bubbleManager";
+import { theOneBubbleManager, updateOverlayClass } from "./bubbleManager";
 
 const kPlaybackOrderContainerSelector: string =
     ".bloom-playbackOrderControlsContainer";
@@ -147,6 +147,14 @@ function ctrlAltKeyUpListener(event: KeyboardEvent) {
                 // Remember, for keyup events, you want to check event.key === "Control" instead of event.ctrlKey
                 if (event.key === "Control") {
                     imageContainer.classList.remove("ui-ctrlDown");
+                } else if (
+                    event.key === "Alt" &&
+                    !theOneBubbleManager.isResizing(imageContainer)
+                ) {
+                    // FYI: Check !isResizing() so if you release Alt but still keep the mouse down, resizing will continue.
+                    // Unsure if this needs to be set in stone, but it's for consistency with
+                    // 1) our historical practice, and 2) how Ctrl+drag currently works
+                    theOneBubbleManager.turnOffResizing(imageContainer);
                 }
             });
 
--- a/src/BloomBrowserUI/bookEdit/js/bloomImages.ts
+++ b/src/BloomBrowserUI/bookEdit/js/bloomImages.ts
@@ -123,6 +123,11 @@ function ctrlAltKeyDownListener(event: KeyboardEvent) {
 
                 if (event.ctrlKey) {
                     imageContainer.classList.add("ui-ctrlDown");
+                } else if (
+                    event.altKey &&
+                    imageContainer instanceof HTMLElement
+                ) {
+                    theOneBubbleManager.addResizeModeClass(imageContainer, {});
                 }
             });
     }
--- a/src/BloomBrowserUI/bookEdit/js/bubbleManager.ts
+++ b/src/BloomBrowserUI/bookEdit/js/bubbleManager.ts
@@ -1187,7 +1187,7 @@ export class BubbleManager {
         );
 
         // Now there are several options depending on various conditions. There's some
-        // overlap in the conditions and it is tempting to try to comine into a single compound
+        // overlap in the conditions and it is tempting to try to combine into a single compound
         // "if" statement. But note, this first one may change hoveredBubble to undefined,
         // which then changes which of the following options is chosen. Be careful!
         if (hoveredBubble && hoveredBubble.content !== this.activeElement) {
@@ -1446,6 +1446,10 @@ export class BubbleManager {
             // on the text inside the bubble.
             event.preventDefault();
             event.stopPropagation();
+        }
+
+        if (this.bubbleToResize) {
+            this.clearResizeModeClasses(container);
         }
         this.bubbleToDrag = undefined;
         this.activeContainer = undefined;
--- a/src/BloomBrowserUI/bookEdit/js/bubbleManager.ts
+++ b/src/BloomBrowserUI/bookEdit/js/bubbleManager.ts
@@ -1149,6 +1149,10 @@ export class BubbleManager {
     // MUST be defined this way, rather than as a member function, so that it can
     // be passed directly to addEventListener and still get the correct 'this'.
     private onMouseMove = (event: MouseEvent) => {
+        // Capture the most recent data to use when our animation frame request is satisfied.
+        // or so keyboard events can reference the current mouse position.
+        this.lastMoveEvent = event;
+
         const container = event.currentTarget as HTMLElement;
         // Prevent two event handlers from triggering if the text box is currently being resized
         if (this.isJQueryResizing(container)) {
@@ -1248,15 +1252,68 @@ export class BubbleManager {
                 container.classList.add("grabbable");
             }
         } else {
-            // Resizing case
-            if (isVideo) {
-                targetElement.removeAttribute("controls");
-            }
-            const resizeMode = this.getResizeMode(hoveredBubble.content, event);
-
-            this.cleanupMouseMoveHover(container); // Need to clear both grabbable and *-resizables
-            container.classList.add(`${resizeMode}-resizable`);
+            this.addResizeModeClass(container, {
+                event,
+                hoveredBubble,
+                isVideo
+            });
         }
+    }
+
+    // Adds the appropriate resize handle to the containers
+    // You can provide just the imageContainer, and let the function handle everything else.
+    // If you already know some of these values in {optionalParams}, you can pass them in so the function won't re-compute them.
+    public addResizeModeClass(
+        container: HTMLElement,
+        optionalParams: {
+            event?: MouseEvent | undefined; // Leave undefined if unknown, this function will determine it
+            hoveredBubble?: Bubble | undefined; // Leave undefined if unknown, this function will determine it
+            isVideo?: boolean | undefined; // Leave undefined if unknown, this function will determine it
+        }
+    ) {
+        let { event, hoveredBubble, isVideo } = optionalParams;
+
+        //////////////////////////////////////////////////////////
+        // First, calculate the values for any undefined values //
+        //////////////////////////////////////////////////////////
+        event = event || this.lastMoveEvent;
+        if (!event) {
+            // Give up
+            return;
+        }
+
+        if (!hoveredBubble) {
+            const coordinates = this.getPointRelativeToCanvas(event, container);
+            if (!coordinates) {
+                // Give up
+                return;
+            }
+
+            hoveredBubble = Comical.getBubbleHit(
+                container,
+                coordinates.getUnscaledX(),
+                coordinates.getUnscaledY()
+            );
+            if (!hoveredBubble) {
+                // Give up
+                return;
+            }
+        }
+
+        if (isVideo === undefined) {
+            isVideo = this.isVideoOverPictureElement(hoveredBubble.content);
+        }
+
+        /////////////////////////////////////////////////////////////////////////////////
+        // Now that we've got all our parameters to be non-undefined, do the real logic//
+        /////////////////////////////////////////////////////////////////////////////////
+        if (isVideo && event.target instanceof Element) {
+            event.target.removeAttribute("controls");
+        }
+        const resizeMode = this.getResizeMode(hoveredBubble.content, event);
+
+        this.cleanupMouseMoveHover(container); // Need to clear both grabbable and *-resizables
+        container.classList.add(`${resizeMode}-resizable`);
     }
 
     private animationFrame: number;
@@ -1270,7 +1327,6 @@ export class BubbleManager {
         container: HTMLElement
     ) {
         // Capture the most recent data to use when our animation frame request is satisfied.
-        this.lastMoveEvent = event;
         this.lastMoveContainer = container;
         // We don't want any other effects of mouse move, like selecting text in the box,
         // to happen while we're dragging it around.
--- a/src/BloomBrowserUI/bookEdit/js/bubbleManager.ts
+++ b/src/BloomBrowserUI/bookEdit/js/bubbleManager.ts
@@ -1110,7 +1110,7 @@ export class BubbleManager {
 
                 // Even though Alt+Drag resize is not in effect, we still check using isResizing() to
                 // make sure JQuery Resizing is not in effect before proceeding.
-                if (!this.isResizing(container)) {
+                if (!this.isJQueryResizing(container)) {
                     container.classList.add("grabbing");
                 }
             } else {
@@ -1151,7 +1151,7 @@ export class BubbleManager {
     private onMouseMove = (event: MouseEvent) => {
         const container = event.currentTarget as HTMLElement;
         // Prevent two event handlers from triggering if the text box is currently being resized
-        if (this.isResizing(container)) {
+        if (this.isJQueryResizing(container)) {
             this.bubbleToDrag = undefined;
             this.activeContainer = undefined;
             return;
@@ -1296,8 +1296,8 @@ export class BubbleManager {
         }
         this.animationFrame = requestAnimationFrame(() => {
             if (!this.bubbleToDrag) {
+                // This case could be reached when using the JQuery drag handle.
                 this.animationFrame = 0; // must clear, or move will forever be blocked.
-                console.assert(false, "bubbleToDrag is undefined");
                 return;
             }
 
@@ -1448,18 +1448,25 @@ export class BubbleManager {
             event.stopPropagation();
         }
 
-        if (this.bubbleToResize) {
-            this.clearResizeModeClasses(container);
-        }
         this.bubbleToDrag = undefined;
+        container.classList.remove("grabbing");
+        this.turnOffResizing(container);
+    };
+
+    public turnOffResizing(container: Element) {
+        this.clearResizeModeClasses(container);
         this.activeContainer = undefined;
         this.bubbleToResize = undefined;
         this.bubbleResizeMode = "";
-        container.classList.remove("grabbing");
-    };
+    }
+
+    public isResizing(container: Element) {
+        return this.bubbleToResize || this.isJQueryResizing(container);
+    }
 
     // Returns true if any of the container's children are currently being resized
-    private isResizing(container: HTMLElement) {
+    // Remarks: I believe this only tells you whether a JQ
+    private isJQueryResizing(container: Element) {
         // First check if we have our custom class indicator applied
         if (container.classList.contains("bloom-resizing")) {
             return true;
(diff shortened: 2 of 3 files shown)


sujeffreyl:
Changed base to 5.5 instead of 5.4 since it's just some QOL changes w/no huge negative impact.

JohnThomson (changes requested):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5544)* status: 0 of 2 files reviewed, 9 unresolved discussions (waiting on @sujeffreyl)

___
*[`src/BloomBrowserUI/bookEdit/js/bloomImages.ts` line 114 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5544#-NL2QG-x34ctpCuH3GQ3:-NL2QG-x34ctpCuH3GQ4:b2sy0h0) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/dd4627b1f721430d7850da5d951a5a428009bc78/src/BloomBrowserUI/bookEdit/js/bloomImages.ts#L114)):*
> ```XML
>         event.target.addEventListener("keyup", ctrlAltKeyUpListener);
> 
>         // Note (paranoia): Add this class conservatively (ensure event.target is good first),
> ```

There's at least two classes being manipulated here. Let's improve this comment and name it (ui-suppressImageButtons, I think).
What is meant by "ensure event.target is good"? Just that it is truthy? Because that seems to be all we're checking.
___
*[`src/BloomBrowserUI/bookEdit/js/bloomImages.ts` line 119 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5544#-NL2AjLK0Wr9zebqV_s-:-NL2AjLK0Wr9zebqV_s0:b9vc4wd) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/dd4627b1f721430d7850da5d951a5a428009bc78/src/BloomBrowserUI/bookEdit/js/bloomImages.ts#L119)):*
> ```XML
>         document
>             .querySelectorAll(
>                 ".bloom-imageContainer:not(.ui-suppressImageButtons)"
> ```

 :not(.ui-suppressImageButtons) originally may have just been an optimization: the loop was about adding this class, so we don't need to do that to any that already has it.

But the logic for showing image buttons recently got more complicated, with the addition of image overlays and then click-to-show-image-buttons on an overlay background. Are you sure the new logic (or even the ui-ctrlDown logic) is never needed when this class is already present?
___
*[`src/BloomBrowserUI/bookEdit/js/bloomImages.ts` line 128 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5544#-NL2QtSWCudUw96lrw04:-NL2QtSWCudUw96lrw05:b-5gsto4) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/dd4627b1f721430d7850da5d951a5a428009bc78/src/BloomBrowserUI/bookEdit/js/bloomImages.ts#L128)):*
> ```XML
>                 } else if (
>                     event.altKey &&
>                     imageContainer instanceof HTMLElement
> ```

Is there some way that a result from querySelectorAll searching for a particular class might NOT be an HTMLElement? Or is this just making LINT happy?
___
*[`src/BloomBrowserUI/bookEdit/js/bubbleManager.ts` line 1266 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5544#-NL2EN1w140mEGPRS7bY:-NL2EN1w140mEGPRS7bZ:bj6zps6) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/dd4627b1f721430d7850da5d951a5a428009bc78/src/BloomBrowserUI/bookEdit/js/bubbleManager.ts#L1266)):*
> ```XML
>     // You can provide just the imageContainer, and let the function handle everything else.
>     // If you already know some of these values in {optionalParams}, you can pass them in so the function won't re-compute them.
>     public addResizeModeClass(
> ```

Let's call this something like addResizeModeClassToAnyHovereredBubble or addResizeClassToAnyBubbleBeingResized. The current name is confusing when called without optionalParams, as it gives the idea that resize classes will be added to the imageContainer itself, which would not be right.

Will this do the right thing if container is for an overlay image, and thus doesn't have a canvas?
___
*[`src/BloomBrowserUI/bookEdit/js/bubbleManager.ts` line 1279 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5544#-NL2SgaD7WgJr7njkciZ:-NL2SgaD7WgJr7njkci_:badzr6p) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/dd4627b1f721430d7850da5d951a5a428009bc78/src/BloomBrowserUI/bookEdit/js/bubbleManager.ts#L1279)):*
> ```XML
>         // First, calculate the values for any undefined values //
>         //////////////////////////////////////////////////////////
>         event = event || this.lastMoveEvent;
> ```

Is it a concern that this.lastMoveEvent is the last movement INSIDE an image container; but this function may be called because the ctrl key was pushed while editing a text block? Conceivably, if a bubble overlaps the edge of the image container and the mouse moves out of that container through the bubble, lastMoveEvent may be over a bubble that is nowhere near the current mouse position. (Should we arrange to clear lastMoveEvent when the mouse moves out of the image container?)
___
*[`src/BloomBrowserUI/bookEdit/js/bubbleManager.ts` line 1311 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5544#-NL2OgUQ8OoqGMhIb1iK:-NL2OgUQ8OoqGMhIb1iL:b5aw2u0) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/dd4627b1f721430d7850da5d951a5a428009bc78/src/BloomBrowserUI/bookEdit/js/bubbleManager.ts#L1311)):*
> ```XML
>         /////////////////////////////////////////////////////////////////////////////////
>         if (isVideo && event.target instanceof Element) {
>             event.target.removeAttribute("controls");
> ```

Is it somehow guaranteed that event.target is the video element that has the controls attribute? For example, could the target be a control that is overlaid on the video?

It is somewhat surprising that nothing obviously restores the controls when we're no longer resizing (e.g., there's nothing in turnOffResizing). A comment might be appropriate. I _think_ it's OK because any mouse movement over the video element makes sure the controls are present iff they ought to be.
___
*[`src/BloomBrowserUI/bookEdit/js/bubbleManager.ts` line 1524 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5544#-NL2LJnH25jJrEh5tUEL:-NL2LJnH25jJrEh5tUEM:b-jwvnp6) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/dd4627b1f721430d7850da5d951a5a428009bc78/src/BloomBrowserUI/bookEdit/js/bubbleManager.ts#L1524)):*
> ```XML
> 
>     // Returns true if any of the container's children are currently being resized
>     // Remarks: I believe this only tells you whether a JQ
> ```

I believe it too...let's just say, "...currently being resized using jquery"
___
*[`src/BloomBrowserUI/bookEdit/js/bubbleManager.ts` line 1526 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5544#-NL2LX_-51QHh-C8nKOW:-NL2LX_-51QHh-C8nKOX:b-nd7fs6) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/dd4627b1f721430d7850da5d951a5a428009bc78/src/BloomBrowserUI/bookEdit/js/bubbleManager.ts#L1526)):*
> ```XML
>     // Remarks: I believe this only tells you whether a JQ
>     private isJQueryResizing(container: Element) {
>         // First check if we have our custom class indicator applied
> ```

Let's improve this comment while we're here. "First check the class that we try to always apply when starting a JQuery resize, so that we know one is in progress even before the mouse moves."

I notice that the class is only removed by a mouseup handler on the handle that started the resize. Will that fire even if the mouse isn't inside the handle at the time of the mouseup? If not, could that account for the resize cursor hanging around when we don't want it? Maybe we need a mouseup on the whole doc to remove this? (But if so be careful not to accumulate lots of them.)
___
*[`src/BloomBrowserUI/bookEdit/js/bubbleManager.ts` line 1527 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5544#-NL2JXec3hwiBBuRBmSg:-NL2JXec3hwiBBuRBmSh:by3yxm9) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/dd4627b1f721430d7850da5d951a5a428009bc78/src/BloomBrowserUI/bookEdit/js/bubbleManager.ts#L1527)):*
> ```XML
>     private isJQueryResizing(container: Element) {
>         // First check if we have our custom class indicator applied
>         if (container.classList.contains("bloom-resizing")) {
> ```

Can we fix the name of this class, too, since it seems it's only applied when we're starting a jquery resize? It ought to have ui- (or is it bloom-ui- or bloomui-?) to indicate that it's a class that shouldn't be saved, so maybe something like ui-resizing-with-jquery?

sujeffreyl (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5544)* status: 0 of 2 files reviewed, 9 unresolved discussions (waiting on @JohnThomson)

___
*[`src/BloomBrowserUI/bookEdit/js/bloomImages.ts` line 114 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5544#-NL2QG-x34ctpCuH3GQ3:-NL2q46E34GcIZWBT5er:bl4xvqo) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/dd4627b1f721430d7850da5d951a5a428009bc78/src/BloomBrowserUI/bookEdit/js/bloomImages.ts#L114)):*

There's at least two classes being manipulated here. Let's improve this comment and name it (ui-suppressImageButtons, I think).
What is meant by "ensure event.target is good"? Just that it is truthy? Because that seems to be all we're checking.
</blockquote></details>

Yeah, because if the event.target is null, then the event listener to remove ui-suppressImageButtons wouldn't be registered.
___
*[`src/BloomBrowserUI/bookEdit/js/bloomImages.ts` line 119 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5544#-NL2AjLK0Wr9zebqV_s-:-NL2rMR5Ap6bw50FaRjF:bhwwspl) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/dd4627b1f721430d7850da5d951a5a428009bc78/src/BloomBrowserUI/bookEdit/js/bloomImages.ts#L119)):*

 :not(.ui-suppressImageButtons) originally may have just been an optimization: the loop was about adding this class, so we don't need to do that to any that already has it.

But the logic for showing image buttons recently got more complicated, with the addition of image overlays and then click-to-show-image-buttons on an overlay background. Are you sure the new logic (or even the ui-ctrlDown logic) is never needed when this class is already present?
</blockquote></details>

Uh... interesting question. I think probably nothing bad happens? But it seems like the situation would be simpler to reason about if I just remove the optimization.
___
*[`src/BloomBrowserUI/bookEdit/js/bloomImages.ts` line 128 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5544#-NL2QtSWCudUw96lrw04:-NL2rqq_DzWX2E1It6WY:b-p2oc47) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/dd4627b1f721430d7850da5d951a5a428009bc78/src/BloomBrowserUI/bookEdit/js/bloomImages.ts#L128)):*

Is there some way that a result from querySelectorAll searching for a particular class might NOT be an HTMLElement? Or is this just making LINT happy?
</blockquote></details>

I would say it's more to make tsc happy than the linter happy. (Theoretically it returns Element, but in practice we always expect HtmlElement out of it. instanceof is more typesafe than "as". However, I think I found a less annoying way to make tsc happy now)
___
*[`src/BloomBrowserUI/bookEdit/js/bubbleManager.ts` line 1266 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5544#-NL2EN1w140mEGPRS7bY:-NL6TOR21OV8jkcD0Xfl:byupoxe) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/dd4627b1f721430d7850da5d951a5a428009bc78/src/BloomBrowserUI/bookEdit/js/bubbleManager.ts#L1266)):*

Let's call this something like addResizeModeClassToAnyHovereredBubble or addResizeClassToAnyBubbleBeingResized. The current name is confusing when called without optionalParams, as it gives the idea that resize classes will be added to the imageContainer itself, which would not be right.

Will this do the right thing if container is for an overlay image, and thus doesn't have a canvas?
</blockquote></details>

Changed to addResizeModeClassToHoveredBubble.

If it's an overlay image - yes, it works just fine. Under the hood, this function actually runs once for each image container. The container for the overlay image returns early nicely (due to no canvas -> no coordinates), whereas the main container does the processing.
___
*[`src/BloomBrowserUI/bookEdit/js/bubbleManager.ts` line 1279 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5544#-NL2SgaD7WgJr7njkciZ:-NL6VC-500QynlxU-20c:b-j9fi5) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/dd4627b1f721430d7850da5d951a5a428009bc78/src/BloomBrowserUI/bookEdit/js/bubbleManager.ts#L1279)):*

Is it a concern that this.lastMoveEvent is the last movement INSIDE an image container; but this function may be called because the ctrl key was pushed while editing a text block? Conceivably, if a bubble overlaps the edge of the image container and the mouse moves out of that container through the bubble, lastMoveEvent may be over a bubble that is nowhere near the current mouse position. (Should we arrange to clear lastMoveEvent when the mouse moves out of the image container?)
</blockquote></details>

The Ctrl key being pressed doesn't cause this function to be called.  I can't think of how  Alt button is actually useful during text editing? I think it is expected to bring up the resize handle if it is pressed.

However, maybe here's a more interesting scenario:
1) Put an overlay image on the page.
2) Put a speech bubble over the overlay image
3) Move the mouse through the overlay image into the speech bubble.
4) Click the speech bubble text.
5) Hold down Alt. The resize mode class comes up. (This part I think is no problem and is expected.)
6) Alt+move the mouse to resize.

Observed: The overlay image is resized.

What is Expected:
Well, partially I think it makes more sense for the speech bubble to resize because 1) we just clicked on it, and 2) it is higher.

However, I see that in current alpha, if you Alt+drag to resize in a region that overlaps both a speech bubble and an overlay image, the overlay image will always be the resized even if  1) the speech bubble was higher, and 2) you moved your mouse through the speech-bubble-only region to the both-overlap region while never touching a overlay-image-only image.
(If we think that behavior in Alpha isn't correct, we should make a new card for that)

Based on that my conclusion is that nothing needs to be changed here.
___
*[`src/BloomBrowserUI/bookEdit/js/bubbleManager.ts` line 1311 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5544#-NL2OgUQ8OoqGMhIb1iK:-NL2xHX15m0WC1LpR1pn:betcimv) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/dd4627b1f721430d7850da5d951a5a428009bc78/src/BloomBrowserUI/bookEdit/js/bubbleManager.ts#L1311)):*

Is it somehow guaranteed that event.target is the video element that has the controls attribute? For example, could the target be a control that is overlaid on the video?

It is somewhat surprising that nothing obviously restores the controls when we're no longer resizing (e.g., there's nothing in turnOffResizing). A comment might be appropriate. I _think_ it's OK because any mouse movement over the video element makes sure the controls are present iff they ought to be.
</blockquote></details>

Your concerns may be valid, but I'd call this out of scope - This is pretty much the same as it was before, except instead of a variable targetElement (which was assigned the value event.target) we are using event.target.   I don't want to bite off fixing the video elements code in this PR too.
___
*[`src/BloomBrowserUI/bookEdit/js/bubbleManager.ts` line 1524 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5544#-NL2LJnH25jJrEh5tUEL:-NL6PULm6WOlJtOEYM5Y:b-896fix) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/dd4627b1f721430d7850da5d951a5a428009bc78/src/BloomBrowserUI/bookEdit/js/bubbleManager.ts#L1524)):*

I believe it too...let's just say, "...currently being resized using jquery"
</blockquote></details>

Done.

sujeffreyl (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5544)* status: 0 of 2 files reviewed, 9 unresolved discussions (waiting on @JohnThomson)

___
*[`src/BloomBrowserUI/bookEdit/js/bubbleManager.ts` line 1266 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5544#-NL2EN1w140mEGPRS7bY:-NL6gAC6CAqUdb-3ZTCh:bixjr5s) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/dd4627b1f721430d7850da5d951a5a428009bc78/src/BloomBrowserUI/bookEdit/js/bubbleManager.ts#L1266)):*

Changed to addResizeModeClassToHoveredBubble.

If it's an overlay image - yes, it works just fine. Under the hood, this function actually runs once for each image container. The container for the overlay image returns early nicely (due to no canvas -> no coordinates), whereas the main container does the processing.
</blockquote></details>

Actually, I changed it back to the original name. After reviewing what the function actually does, I think it is correct. It is the class list of the container that is being modified (see "container.classList.add(`${resizeMode}-resizable`);")
The hovered bubble is just used to calculate which direction the resize arrows should be.
___
*[`src/BloomBrowserUI/bookEdit/js/bubbleManager.ts` line 1526 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5544#-NL2LX_-51QHh-C8nKOW:-NL6cugw-oanrBPPv11y:b-z2cvxm) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/dd4627b1f721430d7850da5d951a5a428009bc78/src/BloomBrowserUI/bookEdit/js/bubbleManager.ts#L1526)):*

Let's improve this comment while we're here. "First check the class that we try to always apply when starting a JQuery resize, so that we know one is in progress even before the mouse moves."

I notice that the class is only removed by a mouseup handler on the handle that started the resize. Will that fire even if the mouse isn't inside the handle at the time of the mouseup? If not, could that account for the resize cursor hanging around when we don't want it? Maybe we need a mouseup on the whole doc to remove this? (But if so be careful not to accumulate lots of them.)
</blockquote></details>

Well, the handle basically is the mouse, so I don't know if that's possible.
That said, I did experimentally find that it was occasionally possible that the clearResizingClass function wouldn't fire at the end of a JQuery resize. (I think I have more success with very short, sudden resizes).  However, I could not observe any ill immediate or downstream effect of this.
Also, Suzanne's problems are related to Alt+drag resizing, not JQuery resizing, so that wouldn't be related.
Since I couldn't observe any ill effect, I'm inclined not to worry about why the mouseup event didn't work in these rare cases.
___
*[`src/BloomBrowserUI/bookEdit/js/bubbleManager.ts` line 1527 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5544#-NL2JXec3hwiBBuRBmSg:-NL6Zhgj3ecEc4lSXxIH:b-jycwuo) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/dd4627b1f721430d7850da5d951a5a428009bc78/src/BloomBrowserUI/bookEdit/js/bubbleManager.ts#L1527)):*

Can we fix the name of this class, too, since it seems it's only applied when we're starting a jquery resize? It ought to have ui- (or is it bloom-ui- or bloomui-?) to indicate that it's a class that shouldn't be saved, so maybe something like ui-resizing-with-jquery?
</blockquote></details>

The prefix is ui

sujeffreyl (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5544)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @JohnThomson)

___
*[`src/BloomBrowserUI/bookEdit/js/bubbleManager.ts` line 1266 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5544#-NL2EN1w140mEGPRS7bY:-NLR59Oj1fQWyNWkC_OF:b4m55hh) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/dd4627b1f721430d7850da5d951a5a428009bc78/src/BloomBrowserUI/bookEdit/js/bubbleManager.ts#L1266)):*
> It will actually only add a resize class IF the most recent mouse movement in this image container was over a child bubble.
I added the option to call without arguments because I was thinking... I mean, yeah, I can add it to the call that needs it, but what happens if tomorrow we need to call it in a 3rd place that doesn't have that info? Then I'd have to copy and paste the logic to compute the arguments.

Well, I tried splitting it into 2 functions. One function is more direct and straightforward now. If you need help getting the other params, you can call the try[...] version of it instead which will help calculate and stuff all the parameters in.
___
*[`src/BloomBrowserUI/bookEdit/js/bubbleManager.ts` line 1263 at r2](https://reviewable.io/reviews/bloombooks/bloomdesktop/5544#-NLN4YNH24oXHcNQFET0:-NLR4f4gEVejwdItcoZa:b-82b77g) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/0be9ff650284a78568d9204b42f8eb7d1b5d6357/src/BloomBrowserUI/bookEdit/js/bubbleManager.ts#L1263)):*

Is this accurate? I can't find what the classes like ne-resizable actually do...must be in some jquery library? But I'm guessing they are there to change the shape of the cursor, not to add a child control?
If so, I'd rather see, "Add a class to the container to make the cursor reflect the direction in which a child bubble is being resized"
</blockquote></details>

The classes are in the content folder (bubble.less), did you look there? (A lot of times I forget...) But yeah, it just sets the cursor to "ne-resize"/etc.

sujeffreyl (commented):
I didn't get that impression! :)

> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5544#-:-NLRSynE9YKAjbOSVszW:b-du0bwi)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @JohnThomson)

JohnThomson (approved):
> Reviewed 1 of 3 files at r2, 2 of 2 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5544)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @sujeffreyl)

___
*[`src/BloomBrowserUI/bookEdit/js/bubbleManager.ts` line 1266 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5544#-NL2EN1w140mEGPRS7bY:-NLSlVlV51xprfd5X3I0:b-5jqteh) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/dd4627b1f721430d7850da5d951a5a428009bc78/src/BloomBrowserUI/bookEdit/js/bubbleManager.ts#L1266)):*

> It will actually only add a resize class IF the most recent mouse movement in this image container was over a child bubble.
I added the option to call without arguments because I was thinking... I mean, yeah, I can add it to the call that needs it, but what happens if tomorrow we need to call it in a 3rd place that doesn't have that info? Then I'd have to copy and paste the logic to compute the arguments.

Well, I tried splitting it into 2 functions. One function is more direct and straightforward now. If you need help getting the other params, you can call the try[...] version of it instead which will help calculate and stuff all the parameters in.
</blockquote></details>

I think this is way better. Thanks!

PR #5546, opened 2022-11-16
Finish fixing bulk upload for Linux (BL-11716)

StephenMcConnel:
The previous fix was necessary but not sufficient when running an installed Bloom.

- - -
This change is [Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5546)

### DIFF
--- a/src/BloomExe/Publish/BloomLibrary/BloomLibraryUploadControl.cs
+++ b/src/BloomExe/Publish/BloomLibrary/BloomLibraryUploadControl.cs
@@ -912,6 +912,19 @@ namespace Bloom.Publish.BloomLibrary
 					Arguments = command,
 					WorkingDirectory = Path.GetDirectoryName(Application.ExecutablePath)
 				};
+				// LD_PRELOAD is a Linux environment variable for a shared library that should be loaded before any other library is
+				// loaded by a program that is starting up.  It is rarely needed, but the mozilla code used by Geckofx is one place
+				// where this feature is used, specifically to load a xulrunner patch (libgeckofix.so) that must be in place before
+				// xulrunner can be initialized for GeckoFx60 on Linux.  This must be in place in the environment before launching
+				// any process (such as Bloom, here) that will initialize xulrunner, but may cause problems for other programs so
+				// it is best not to have it in the environment unless we know it is needed.  In particular having LD_PRELOAD set to
+				// load libgeckofix.so is known to cause problems when running some programs (possibly only BloomPdfMaker.exe) using
+				// CommandLineRunner.  To guard against this Program.Main() removes it from the environment, but here we need to
+				// temporarily restore it so it can be inherited by the instance of Bloom we are about to launch. Fortunately, it's
+				// easy to reconstruct.
+				var xulRunner = Environment.GetEnvironmentVariable("XULRUNNER");
+				if (!String.IsNullOrEmpty("xulRunner"))
+					Environment.SetEnvironmentVariable("LD_PRELOAD", $"{xulRunner}/libgeckofix.so");
 			}
 
 			Process.Start(startInfo);
@@ -921,7 +934,8 @@ namespace Bloom.Publish.BloomLibrary
 			_progressBox.WriteMessage("When the upload is complete, there will be a file named 'BloomBulkUploadLog.txt' in your collection folder.");
 			var url = $"{BloomLibraryUrls.BloomLibraryUrlPrefix}/{_model.Book.CollectionSettings.DefaultBookshelf}";
 			_progressBox.WriteMessage("Your books will show up at {0}", url);
-		
+			if (SIL.PlatformUtilities.Platform.IsLinux)	// LD_PRELOAD interferes with CommandLineRunner and GeckoFx60 on Linux
+				Environment.SetEnvironmentVariable("LD_PRELOAD", null);
 		}
 
 		private void _uploadSource_SelectedIndexChanged(object sender, EventArgs e)


JohnThomson (changes requested):
Looks reasonable but needs more explanation, I think, to be maintainable.

> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5546#-:-NH0rYR49k6F8njG0dq-:b-me123t)* status: all files reviewed, 2 unresolved discussions (waiting on @StephenMcConnel)

___
*[`src/BloomExe/Publish/BloomLibrary/BloomLibraryUploadControl.cs` line 915 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5546#-NH0ofFWC6jQ0i1QzQPY:-NH0ofFWC6jQ0i1QzQPZ:b-v52tvf) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/c62541b437c656a1821b5f8099440760736df3fc/src/BloomExe/Publish/BloomLibrary/BloomLibraryUploadControl.cs#L915)):*
> ```Smalltalk
> 					WorkingDirectory = Path.GetDirectoryName(Application.ExecutablePath)
> 				};
> 				// The process that starts will inherit our environment variables, and we need LD_PRELOAD to be set on Linux.
> ```

I think this needs more explanation. I don't know what LD_PRELOAD is or what uses it. Apparently something other than GeckoFx, since the comment below says that command line running and GeckoFx don't mix. But we're setting LD_PRELOAD to a value that is clearly GeckoFx-related. It also wouldn't hurt to explain why we need a "libgeckofix.so", and if it's really not GeckoFx that it fixes, can we find a better name for this file? Maybe we're fixing Gecko itself, the engine at the core of FF and GeckoFx? Think about what someone other than you would want to know, if they come to this in a couple of year's time and we're in the process of replacing GeckoFx with some linux-compatible modern browser component. (You don't necessarily need to answer all of this...perhaps some of the questions just arise out of my ignorance of the answers to others. They are meant to prompt the sort of information I'd like to see in a comment rather than to insist on exactly what I've asked.)
___
*[`src/BloomExe/Publish/BloomLibrary/BloomLibraryUploadControl.cs` line 930 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5546#-NH0p-Qi-8CNE0nAHQgB:-NH0p-Qi-8CNE0nAHQgC:b-bv9hq) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/c62541b437c656a1821b5f8099440760736df3fc/src/BloomExe/Publish/BloomLibrary/BloomLibraryUploadControl.cs#L930)):*
> ```Smalltalk
> 			_progressBox.WriteMessage("Your books will show up at {0}", url);
> 			if (SIL.PlatformUtilities.Platform.IsLinux)
> 				Environment.SetEnvironmentVariable("LD_PRELOAD", null);	// LD_PRELOAD interferes with CommandLineRunner with GeckoFx60
> ```

Needs more explanation. We've just been doing command line stuff, during which we apparently needed LD_PRELOAD. But it will cause a problem if we're BOTH running command lines AND using GeckoFx60? When would we do that? Would this become a problem if the upload code used a GeckoFx window to, say, figure out accurately which fonts to report for analytics?

StephenMcConnel (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5546)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @JohnThomson)

___
*[`src/BloomExe/Publish/BloomLibrary/BloomLibraryUploadControl.cs` line 915 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5546#-NH0ofFWC6jQ0i1QzQPY:-NH18qSK2pUzztMNNAWN:b-pi0lq0) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/c62541b437c656a1821b5f8099440760736df3fc/src/BloomExe/Publish/BloomLibrary/BloomLibraryUploadControl.cs#L915)):*

I think this needs more explanation. I don't know what LD_PRELOAD is or what uses it. Apparently something other than GeckoFx, since the comment below says that command line running and GeckoFx don't mix. But we're setting LD_PRELOAD to a value that is clearly GeckoFx-related. It also wouldn't hurt to explain why we need a "libgeckofix.so", and if it's really not GeckoFx that it fixes, can we find a better name for this file? Maybe we're fixing Gecko itself, the engine at the core of FF and GeckoFx? Think about what someone other than you would want to know, if they come to this in a couple of year's time and we're in the process of replacing GeckoFx with some linux-compatible modern browser component. (You don't necessarily need to answer all of this...perhaps some of the questions just arise out of my ignorance of the answers to others. They are meant to prompt the sort of information I'd like to see in a comment rather than to insist on exactly what I've asked.)
</blockquote></details>

If you never use Linux, or develop for it, you don't need to know about LD_PRELOAD.
libgeckofix.so was the name chosen by Tom Hindle while developing Geckofx45.
However, I beefed up the comments.
The need for clearing LD_PRELOAD dates back 4 years to  getting Geckofx60 to work.  My memory of the details may not be quite accurate, but I think it's plausible enough.
___
*[`src/BloomExe/Publish/BloomLibrary/BloomLibraryUploadControl.cs` line 930 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5546#-NH0p-Qi-8CNE0nAHQgB:-NH1DjLt-URoUMjATn5N:bjcg8o8) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/c62541b437c656a1821b5f8099440760736df3fc/src/BloomExe/Publish/BloomLibrary/BloomLibraryUploadControl.cs#L930)):*

Needs more explanation. We've just been doing command line stuff, during which we apparently needed LD_PRELOAD. But it will cause a problem if we're BOTH running command lines AND using GeckoFx60? When would we do that? Would this become a problem if the upload code used a GeckoFx window to, say, figure out accurately which fonts to report for analytics?
</blockquote></details>

I revised the comment.  But I don't think this needs that much explanation in addition to the (newly enhanced) comment where the variable is set.

JohnThomson (approved):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5546)* status: all files reviewed, 1 unresolved discussion (waiting on @StephenMcConnel)

___
*[`src/BloomExe/Publish/BloomLibrary/BloomLibraryUploadControl.cs` line 915 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5546#-NH0ofFWC6jQ0i1QzQPY:-NH2ggbZ5D9lnPrmcl8I:bi6i1i4) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/c62541b437c656a1821b5f8099440760736df3fc/src/BloomExe/Publish/BloomLibrary/BloomLibraryUploadControl.cs#L915)):*

If you never use Linux, or develop for it, you don't need to know about LD_PRELOAD.
libgeckofix.so was the name chosen by Tom Hindle while developing Geckofx45.
However, I beefed up the comments.
The need for clearing LD_PRELOAD dates back 4 years to  getting Geckofx60 to work.  My memory of the details may not be quite accurate, but I think it's plausible enough.
</blockquote></details>

A critical thing I missed, but probably only because of only looking at diffs, is that here we are launching another copy of Bloom, which will of course need to initialize xulrunner for itself.

So let's see if I've got it right. LD_PRELOAD
- must be set to this value before Bloom's main exe starts so that Bloom can initialize xulrunner (something sets this before our Main() even starts?)
- must NOT be set any time we use CommandLineRunner [at all? or specifically to run BloomPdfMaker.exe? or just, most things don't need it so it's best if we don't do it to any process we start that we don't know needs it?] (so our Main() code removes it once we've initialized xulrunner, and during most of Bloom's execution it's supposed to be unset)
- but here, we're about to launch Bloom itself as a subtask to upload a book. The new instance will initialize xulrunner as always, so we need to restore LD_PRELOAD to its normal value for starting Bloom.

Is it having ANY value in LD_PRELOAD that interferes with CommandLineRunner/BloomPdfMaker? Or specifically having it set to run libgeckofix.so?

Would it be correct (and helpful) to say "...where this feature is used, specifically to load a xulrunner patch (libgeckofix.so) that must be in place before xulrunner can be initialized for GeckoFx 60 on Linux. This must be in place in the environment before launching any process (such as Bloom, here) that will initialize xulrunner, but may cause problems for other programs so it is best not to have it in the environment except when we know it is needed. In particular having LD_PRELOAD set to load libgeckofix.so is known to cause problems when running BloomPdfMaker using CommandLineRunner. To guard against this Main() removes it from the environment, but here we need to temporarily restore it so it can be inherited by the instance of Bloom we are about to launch.

If that's accurate I think I would have found it more helpful, but maybe an experienced Linux developer would find what you have better. If you don't think it's an improvement I'm OK with what you have.

StephenMcConnel (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5546)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @JohnThomson)

___
*[`src/BloomExe/Publish/BloomLibrary/BloomLibraryUploadControl.cs` line 915 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5546#-NH0ofFWC6jQ0i1QzQPY:-NH5YmUy6d654NUgK-IE:b-j578bt) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/c62541b437c656a1821b5f8099440760736df3fc/src/BloomExe/Publish/BloomLibrary/BloomLibraryUploadControl.cs#L915)):*

A critical thing I missed, but probably only because of only looking at diffs, is that here we are launching another copy of Bloom, which will of course need to initialize xulrunner for itself.

So let's see if I've got it right. LD_PRELOAD
- must be set to this value before Bloom's main exe starts so that Bloom can initialize xulrunner (something sets this before our Main() even starts?)
- must NOT be set any time we use CommandLineRunner [at all? or specifically to run BloomPdfMaker.exe? or just, most things don't need it so it's best if we don't do it to any process we start that we don't know needs it?] (so our Main() code removes it once we've initialized xulrunner, and during most of Bloom's execution it's supposed to be unset)
- but here, we're about to launch Bloom itself as a subtask to upload a book. The new instance will initialize xulrunner as always, so we need to restore LD_PRELOAD to its normal value for starting Bloom.

Is it having ANY value in LD_PRELOAD that interferes with CommandLineRunner/BloomPdfMaker? Or specifically having it set to run libgeckofix.so?

Would it be correct (and helpful) to say "...where this feature is used, specifically to load a xulrunner patch (libgeckofix.so) that must be in place before xulrunner can be initialized for GeckoFx 60 on Linux. This must be in place in the environment before launching any process (such as Bloom, here) that will initialize xulrunner, but may cause problems for other programs so it is best not to have it in the environment except when we know it is needed. In particular having LD_PRELOAD set to load libgeckofix.so is known to cause problems when running BloomPdfMaker using CommandLineRunner. To guard against this Main() removes it from the environment, but here we need to temporarily restore it so it can be inherited by the instance of Bloom we are about to launch.

If that's accurate I think I would have found it more helpful, but maybe an experienced Linux developer would find what you have better. If you don't think it's an improvement I'm OK with what you have.
</blockquote></details>

I adopted your wording, more or less.

PR #5551, opened 2022-11-18
Add missing languages to BloomLibrary.AudioLangs (BL-11784)

JohnThomson:
This change is [Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5551)

### DIFF
--- a/src/BloomExe/Publish/BloomLibrary/BloomLibraryPublishModel.cs
+++ b/src/BloomExe/Publish/BloomLibrary/BloomLibraryPublishModel.cs
@@ -296,15 +296,39 @@ namespace Bloom.Publish.BloomLibrary
 			if (bookInfo.PublishSettings.BloomLibrary.AudioLangs == null)
 			{
 				bookInfo.PublishSettings.BloomLibrary.AudioLangs = new Dictionary<string, InclusionSetting>();
-				var allLangCodes = allLanguages.Select(x => x.Key);
-				foreach (var langCode in allLangCodes)
-				{
-					bookInfo.PublishSettings.BloomLibrary.AudioLangs[langCode] = InclusionSetting.IncludeByDefault;
-				}
+			}
+
+			var allLangCodes = allLanguages.Select(x => x.Key);
+
+			// This is tricky, because currently our UI does not support different settings for different languages.
+			// Thus, an older version of this code only did this init the first time, when AudioLangs was null.
+			// However, at least one book (BL-11784) got into a state where AudioLangs is an empty set.
+			// Then there is NO item in the set to be switched on or off, so effectively, it remains off.
+			// (IncludeAudio is true if at least one is Include or IncludeByDefault).
+			// It is therefore necessary, minimally, to make sure the set has at least one language
+			// if allLangCodes has any. But properly, it is intended to  have a value for every language in
+			// allLangCodes...we just temporarily expect that they will all be the same.
+			// It seems more future-proof to add any languages that are missing, while not changing the
+			// setting for any we already have. But then, what setting should we use for any language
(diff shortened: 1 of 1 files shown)


jsubloom (changes requested):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5551)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @JohnThomson)

___
*[`src/BloomExe/Publish/BloomLibrary/BloomLibraryPublishModel.cs` line 326 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5551#-NHQ5S-S-6hvIrhtG3YV:-NHQ5S-T69MGHXS-9ESn:b-hoytrm) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/897124c429d7b6927b9cf4657cc53dc910c779d6/src/BloomExe/Publish/BloomLibrary/BloomLibraryPublishModel.cs#L326)):*
> ```Smalltalk
> 			foreach (var langCode in allLangCodes)
> 			{
> 				if (!bookInfo.PublishSettings.BloomLibrary.AudioLangs.TryGetValue(langCode, out InclusionSetting _))
> ```

It seems like we could just use ContainsKey instead, which would also be (slightly) easier to grok.

jsubloom (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5551)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @JohnThomson)

___
*[`src/BloomExe/Publish/BloomLibrary/BloomLibraryPublishModel.cs` line 317 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5551#-NHQ892k7Mc6L0Z93Gdm:-NHQ892k7Mc6L0Z93Gdn:b-e5s1l0) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/897124c429d7b6927b9cf4657cc53dc910c779d6/src/BloomExe/Publish/BloomLibrary/BloomLibraryPublishModel.cs#L317)):*
> ```Smalltalk
> 			// any time we find a new language, even if it was previously off. So it seems better to copy
> 			// the setting from an arbitrary current language if any; if there are none, then we go with
> 			// the old default. For now, this will keep the old value of the setting. If we enhance the UI
> ```

Ok, i'll buy it.



_Code quote:_
```Smalltalk
			// any time we find a new language, even if it was previously off. So it seems better to copy
			// the setting from an arbitrary current language if any; if there are none, then we go with
			// the old default. For now, this will keep the old value of the setting. If we enhance the UI
```

JohnThomson (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5551)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @sujeffreyl)

___
*[`src/BloomExe/Publish/BloomLibrary/BloomLibraryPublishModel.cs` line 326 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5551#-NHQ5S-S-6hvIrhtG3YV:-NHV7Fq27wT949T5IFUm:b-896fix) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/897124c429d7b6927b9cf4657cc53dc910c779d6/src/BloomExe/Publish/BloomLibrary/BloomLibraryPublishModel.cs#L326)):*

It seems like we could just use ContainsKey instead, which would also be (slightly) easier to grok.
</blockquote></details>

Done.

jsubloom (approved):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5551)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @JohnThomson)

PR #5567, opened 2023-01-05
Fix crashing from rapid zooming with WebView2 (BL-11530)

StephenMcConnel:
This change is [Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5567)

### DIFF
--- a/src/BloomExe/Edit/EditingView.cs
+++ b/src/BloomExe/Edit/EditingView.cs
@@ -1727,6 +1727,9 @@ namespace Bloom.Edit
 		/// Timeout timer for HandleDelayedZoom (handler for _browser1.WebBrowser.DocumentFinished)
 		/// _zoomTimer.Enabled flags that a previous SetZoom request is still being processed, and
 		/// that the current request needs to be delayed.
+		/// For WebView2, this timer is also used to ensure that changing the IFrame src doesn't
+		/// happen while handling the FrameDocumentCompleted event.  Doing so seems to have very unhappy
+		/// results: either freezing or crashing the program with a very low level error in the browser.
 		/// </summary>
 		private Timer _zoomTimer = new Timer();
 
@@ -1735,6 +1738,7 @@ namespace Bloom.Edit
 			// We need to synchronize between user clicks and browser DocumentCompleted events.
 			lock (_zoomTimer)
 			{
+				//Debug.WriteLine($"DEBUG SetZoom({zoom}): _desiredZoomLevel={_desiredZoomLevel}; _zoomTimer.Enabled={_zoomTimer.Enabled}");
 				if (_zoomTimer.Enabled)
 				{
 					// Store the desired zoom level for use later when the previous request has
@@ -1752,6 +1756,8 @@ namespace Bloom.Edit
 				// longest time I measured was 2.961 seconds for ZoomDocumentCompleted to fire.  The
 				// shortest time interval measured was 0.431 seconds.  The average was somewhere around
 				// 0.500-0.600 seconds.
+				// The timer is also used (with the Interval reset to 1 msec) to call SetZoom indirectly
+				// from inside the ZoomDocumentCompleted handler, which is needed for WebView2.
 				_zoomTimer.Interval = 6000;
 				_zoomTimer.Tick += HandleDelayedZoom;
 				_zoomTimer.Start();
@@ -1785,8 +1791,16 @@ namespace Bloom.Edit
 			_browser1.DocumentCompleted -= ZoomDocumentCompleted;
 			Debug.WriteLine("EditingView.ZoomDocumentCompleted() after SetZoom({0}): desired Zoom = {1}, time interval = {2} ms",
 				_previousZoomLevel, _desiredZoomLevel, (DateTime.Now - _previousZoomTime).TotalMilliseconds);
-			// short-circuit the timer.  The only purpose of the timer is a time-out for this event to occur.
-			HandleDelayedZoom(sender, e);
+			// short-circuit the timer.
+			// A minimal delay will do: we can call HandleDelayedZoom directly for Geckofx, but WebView2 just needs the
+			// call to HandleDelayed Zoom to not come from inside this event handler as far as I can tell.
+			// (Note that Geckofx doesn't actually use the native DocumentCompleted event handler while WebView2 does.)
+			lock (_zoomTimer)
+			{
+				_zoomTimer.Stop();
+				_zoomTimer.Interval = 1;
+				_zoomTimer.Start();
+			}
 		}
 
 		public void AdjustPageZoom(int delta)


JohnThomson (changes requested):
I only have one minor suggestion about this as a patch. But looking at this code has made me wonder whether other things besides zoom need (or zoom no longer needs as much of) the sort of protection against overlapping updates that is implemented here.

> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5567#-:-NL75DuOB_k_EZ62Dy_V:bkf0vod)* status: all files reviewed, 2 unresolved discussions (waiting on @StephenMcConnel)

___
*[`src/BloomExe/Edit/EditingView.cs` line 1746 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5567#-NL6tKJWCmSipnb3rsIA:-NL6tKJWCmSipnb3rsIB:bnhu4ev) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/38fcab008387865d0ce993ae00015d5e9d11880f/src/BloomExe/Edit/EditingView.cs#L1746)):*
> ```Smalltalk
> 					// Store the desired zoom level for use later when the previous request has
> 					// finished its UI refresh.
> 					_zoomTimer.Stop();	// restart the wait...
> ```

I don't see why we need to restart it. We haven't done anything that would make the page load that is already in progress take longer, have we?
___
*[`src/BloomExe/Edit/EditingView.cs` line 1775 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5567#-NL6uqx02OP6sK-JWsyl:-NL6uqx02OP6sK-JWsym:butk2sr) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/38fcab008387865d0ce993ae00015d5e9d11880f/src/BloomExe/Edit/EditingView.cs#L1775)):*
> ```Smalltalk
> 				// However, when it was later changed to post a request for reloading the page,
> 				// it became cleaner to just do the reload directly here.
> 				_model.RethinkPageAndReloadIt();
> ```

Seems that the goal of the logic here is that we should not kick off a new attempt to RethinkPageAndReloadIt (in order to change zoom level again) while we are already doing one from a previous Zoom click. But why would that be specific to when the previous page load is because of a change to zoom level? If there's some reason we can't start a new RethinkPageAndReloadIt while one is already in progress, then the logic to prevent it should be in that method, or even at a lower level.
But why shouldn't we start another RethinkPageAndReloadIt immediately? According to comments in NavigationIsolator, it's not a problem to start a new navigation while one is still in progress, so we don't NEED to wait for the browser to finish navigating. Possibly the issue is that we at least need to FinishSavingPage (the first thing RethinkPageAndReloadIt does)? But if that's all we need to wait for, it seems we get a reliable indication when it's done, so maybe we can get rid of the timer altogether? If we get a new zoom while we're in the process of saving the page, we don't even need to save it again...we just need to to navigate to a page with the right final zoom level after we get done saving.
But what if we get a new zoom after we saved the page but before we're done navigating? It's possible we can't reliably save the page content while navigating to a new version of the page. (And I can't think of any case where we'd want to...there can't be changes that need saving in a page we haven't even finished loading, surely.)
Is it possible that what we really need is not any special logic in the Zoom code, but instead, a combination of timer and watching for navigation completed that would allow the SavePage code to know if we are in the process of navigating to a new version of the editable page? And then if we are, SavePage can just return without trying to save? And then Zoom (and anything else that results in reloading the page) can proceed without worrying about navigation in progress?
Of course 6 seconds is too long to just assume we're still navigating if we don't get a document completed. The user could well make a change and switch pages again in less than 6s, and we don't want to miss saving the change. So if documentCompleted is still unreliable in WebView2, we might have to think again.
Maybe we should capture these thoughts in a new card rather than trying to tackle it here. Your change will at least fix the immediate problem.

StephenMcConnel (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5567)* status: all files reviewed, 2 unresolved discussions (waiting on @JohnThomson)

___
*[`src/BloomExe/Edit/EditingView.cs` line 1746 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5567#-NL6tKJWCmSipnb3rsIA:-NLO2pae0fdhHyR7ekvf:b-ik42af) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/38fcab008387865d0ce993ae00015d5e9d11880f/src/BloomExe/Edit/EditingView.cs#L1746)):*

I don't see why we need to restart it. We haven't done anything that would make the page load that is already in progress take longer, have we?
</blockquote></details>

I think you're right.  These lines were probably left from some early trials that didn't actually fix anything.  I've removed the Stop and Start calls here.
___
*[`src/BloomExe/Edit/EditingView.cs` line 1775 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5567#-NL6uqx02OP6sK-JWsyl:-NLO37XhAZZ0YU2UK905:b1qbr3e) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/38fcab008387865d0ce993ae00015d5e9d11880f/src/BloomExe/Edit/EditingView.cs#L1775)):*

Seems that the goal of the logic here is that we should not kick off a new attempt to RethinkPageAndReloadIt (in order to change zoom level again) while we are already doing one from a previous Zoom click. But why would that be specific to when the previous page load is because of a change to zoom level? If there's some reason we can't start a new RethinkPageAndReloadIt while one is already in progress, then the logic to prevent it should be in that method, or even at a lower level.
But why shouldn't we start another RethinkPageAndReloadIt immediately? According to comments in NavigationIsolator, it's not a problem to start a new navigation while one is still in progress, so we don't NEED to wait for the browser to finish navigating. Possibly the issue is that we at least need to FinishSavingPage (the first thing RethinkPageAndReloadIt does)? But if that's all we need to wait for, it seems we get a reliable indication when it's done, so maybe we can get rid of the timer altogether? If we get a new zoom while we're in the process of saving the page, we don't even need to save it again...we just need to to navigate to a page with the right final zoom level after we get done saving.
But what if we get a new zoom after we saved the page but before we're done navigating? It's possible we can't reliably save the page content while navigating to a new version of the page. (And I can't think of any case where we'd want to...there can't be changes that need saving in a page we haven't even finished loading, surely.)
Is it possible that what we really need is not any special logic in the Zoom code, but instead, a combination of timer and watching for navigation completed that would allow the SavePage code to know if we are in the process of navigating to a new version of the editable page? And then if we are, SavePage can just return without trying to save? And then Zoom (and anything else that results in reloading the page) can proceed without worrying about navigation in progress?
Of course 6 seconds is too long to just assume we're still navigating if we don't get a document completed. The user could well make a change and switch pages again in less than 6s, and we don't want to miss saving the change. So if documentCompleted is still unreliable in WebView2, we might have to think again.
Maybe we should capture these thoughts in a new card rather than trying to tackle it here. Your change will at least fix the immediate problem.
</blockquote></details>

I'd rather create a new card to look at the generalized issue involving RethinkPageAndReloadIt.

JohnThomson (approved):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5567)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @StephenMcConnel)

PR #5569, opened 2023-01-06
Switch to server finding images by exact paths (BL-9989)

JohnThomson:
After the change, images that are part of the contents of a book are expected to be in the root folder of the book; images (like UI icons) that are part of Bloom's implementation must be at the specified path relative to output/browser, with a handful of exceptions.
As well as improving performance, this prevents finding unintended files in obscure locations when the file we want is not where we expect.

- - -
This change is [Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5569)


andrew-polk (changes requested):
Is this commit comment supposed to say, "Bloom implements images..."?
I can't make sense of the second part of the first sentence (after the semi-colon).

> Reviewed 24 of 24 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5569#-:-NM0x_W88zlyrqn-ooFo:bnvoyig)* status: all files reviewed, 6 unresolved discussions (waiting on @JohnThomson)

___
*[`src/BloomBrowserUI/bookEdit/toolbox/talkingBook/audioRecording.less` line 134 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5569#-NM0yICs-GvqzgoNhVAH:-NM0yICs-GvqzgoNhVAI:b-6dp4v7) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/b8581ad854a5deccd17f345bd1edf89b7d29ac7e/src/BloomBrowserUI/bookEdit/toolbox/talkingBook/audioRecording.less#L134)):*
> ```Less
> 
>         &.expected {
>             background-image: url(play_expected.svg);
> ```

Shouldn't all of these have the `/bloom/images` prefix?
___
*[`src/BloomBrowserUI/bookEdit/toolbox/talkingBook/talkingBookToolboxTool.pug` line 37 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5569#-NM0ypc90RJXq9iCUUcH:-NM0ypc90RJXq9iCUUcI:b81hbg4) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/b8581ad854a5deccd17f345bd1edf89b7d29ac7e/src/BloomBrowserUI/bookEdit/toolbox/talkingBook/talkingBookToolboxTool.pug#L37)):*
> ```Pug
> 				div.ui-audioInputGroup
> 					div.ui-audioInput
> 						img(id='audio-input-dev' src='/bloom/bookEdit/toolbox/talkingBook/microphone.svg' height='15' width='15' alt='mic')
> ```

It looks like this previously didn't provide a src because audioRecording.ts fills it in later.
Did you intentionally add it here to prevent flicker or some other reason? (to get rid of a warning or something else?)
I could see how this might introduce flicker of the image but perhaps reduce flicker of the layout if it provides the correct vertical spacing from the beginning.
Anyway, it may be worth a comment.
___
*[`src/BloomBrowserUI/publish/ReaderPublish/MethodChooser.tsx` line 31 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5569#-NM11DF4BgMq7sI0D-S_:-NM11DF4BgMq7sI0D-Sa:b-oi0j6b) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/b8581ad854a5deccd17f345bd1edf89b7d29ac7e/src/BloomBrowserUI/publish/ReaderPublish/MethodChooser.tsx#L31)):*
> ```XML
> const wifiImage = "publish-via-wifi.svg";
> const usbImage = "publish-via-usb.svg";
> const fileImage = "publish-to-file.svg";
> ```

Should these be prefixed with `/bloom/publish/ReaderPublish`?
(or you could put the prefix in the jsx below.)
___
*[`src/BloomExe/web/BloomServer.cs` line 576 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5569#-NM15_q77qUKerQh13z7:-NM15_q77qUKerQh13z8:bgri107) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/b8581ad854a5deccd17f345bd1edf89b7d29ac7e/src/BloomExe/web/BloomServer.cs#L576)):*
> ```Smalltalk
> 
> 				if (!RobustFile.Exists(imageFile))
> 				{
> ```

This section -- with the three exceptions -- could use further clarity.
Does the "Some special cases" comment apply to all three? It seems to only apply to the first but is plural.
But also, I don't think you need to say "can't find the source of". The references are in the ckeditor source js and css. So you could just say something like, "this file src comes from the guts of ckeditor, so rather than trying to change that code/css, we just handle it here as a special case."


```
					// These two get copied to the book folder, but while we're doing origami layout
					// I don't want to, in case the user doesn't use them after all.

```
seems incomplete. They get copied later? But we might just be doing origami currently?
Then how about something like, "If the user does add a video or widget, these placeholder .svgs will get copied to the book folder and used from there. But we don't copy to the book folder while the user is still in origami in case the user doesn't actually add the video or widget. So while origami is open, it hits this path and we grab the .svgs from their original locations.
___
*[`src/BloomExe/web/BloomServer.cs` line 589 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5569#-NM13u3yEsFdTEJm6Cn6:-NM13u3yEsFdTEJm6Cn7:b9yoldu) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/94a45f9f832ad3eebd51851733908e440b885094/src/BloomExe/web/BloomServer.cs#L589)):*
> ```Smalltalk
> 				// BL-10625: Cover images still need to be processed to be transparent for book previewing.
> 				if (true == CurrentBook?.ImageFileIsForBookCover(imageFile))
> 					imageFile = _cache.GetPathToResizedImage(imageFile, getThumbnail: false, isForCover: true);
> ```

I'm probably missing something, but I'm seeing a change in logic here. Previously it seems we only did this for preview for cover images. Now it appears we do it for all book preview images.

A comment would probably help our future selves put it together.
___
*[`src/content/templates/template books/Digital Comic Book/placeHolder.png` line 0 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5569#-NM129dy1OXpOnmijhhl:-NM129dy1OXpOnmijhhm:b-vd8bes) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/b8581ad854a5deccd17f345bd1edf89b7d29ac7e/src/content/templates/template%20books/Digital%20Comic%20Book/placeHolder.png#L0)):*
Is every template supposed to have its own placeHolder.png?
Seems like we should be able to just ship one (or at least only one in the templates dir).
If we must maintain one per template, I notice that `Activity` and `Paper Comic Book` don't have it, either. (Also the `Full Screen` template under experiments.)

andrew-polk (changes requested):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5569)* status: all files reviewed, 6 unresolved discussions (waiting on @JohnThomson)

___
*[`src/BloomBrowserUI/bookEdit/toolbox/talkingBook/talkingBookToolboxTool.pug` line 37 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5569#-NM0ypc90RJXq9iCUUcH:-NM18yPPAi3jXQiCPfpi:b-m864k3) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/b8581ad854a5deccd17f345bd1edf89b7d29ac7e/src/BloomBrowserUI/bookEdit/toolbox/talkingBook/talkingBookToolboxTool.pug#L37)):*

It looks like this previously didn't provide a src because audioRecording.ts fills it in later.
Did you intentionally add it here to prevent flicker or some other reason? (to get rid of a warning or something else?)
I could see how this might introduce flicker of the image but perhaps reduce flicker of the layout if it provides the correct vertical spacing from the beginning.
Anyway, it may be worth a comment.
</blockquote></details>

Also, the code in audioRecording.ts which fills in the appropriate src looks like it needs the prefix.

JohnThomson (commented):
Hopefully clarified

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5569#-:-NM4Kuwb6T-Msti_ozHt:b-2z7rza)* status: 23 of 26 files reviewed, 6 unresolved discussions (waiting on @andrew-polk)

___
*[`src/BloomBrowserUI/bookEdit/toolbox/talkingBook/audioRecording.less` line 134 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5569#-NM0yICs-GvqzgoNhVAH:-NM4KzKJ8pPdXroLNjzP:b5a8c64) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/b8581ad854a5deccd17f345bd1edf89b7d29ac7e/src/BloomBrowserUI/bookEdit/toolbox/talkingBook/audioRecording.less#L134)):*

Shouldn't all of these have the `/bloom/images` prefix?
</blockquote></details>

They don't need it because the build process puts the image is in the same folder as audioRecording.css
___
*[`src/BloomBrowserUI/bookEdit/toolbox/talkingBook/talkingBookToolboxTool.pug` line 37 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5569#-NM0ypc90RJXq9iCUUcH:-NM4LjSk4JyuTnmacodN:b-1b7i9j) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/b8581ad854a5deccd17f345bd1edf89b7d29ac7e/src/BloomBrowserUI/bookEdit/toolbox/talkingBook/talkingBookToolboxTool.pug#L37)):*

Also, the code in audioRecording.ts which fills in the appropriate src looks like it needs the prefix.
</blockquote></details>

That code gets the prefix from here. It replaces everything in the original src AFTER the last slash with the appropriate image name. (I think it only worked before by luck...because we used to search EVERYWHERE.)
___
*[`src/BloomBrowserUI/publish/ReaderPublish/MethodChooser.tsx` line 31 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5569#-NM11DF4BgMq7sI0D-S_:-NM4MS4K4b1WkBJXcLbc:b-357mh0) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/b8581ad854a5deccd17f345bd1edf89b7d29ac7e/src/BloomBrowserUI/publish/ReaderPublish/MethodChooser.tsx#L31)):*

Should these be prefixed with `/bloom/publish/ReaderPublish`?
(or you could put the prefix in the jsx below.)
</blockquote></details>

I tried to explain in the comment above the group. Since these images end up in the same folder as the HTML file that uses them, they don't need a prefix; the browser will supply it by copying from the containing file.
___
*[`src/BloomExe/web/BloomServer.cs` line 576 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5569#-NM15_q77qUKerQh13z7:-NM4N8AK5DKBn_509xoG:b-nahmws) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/b8581ad854a5deccd17f345bd1edf89b7d29ac7e/src/BloomExe/web/BloomServer.cs#L576)):*

This section -- with the three exceptions -- could use further clarity.
Does the "Some special cases" comment apply to all three? It seems to only apply to the first but is plural.
But also, I don't think you need to say "can't find the source of". The references are in the ckeditor source js and css. So you could just say something like, "this file src comes from the guts of ckeditor, so rather than trying to change that code/css, we just handle it here as a special case."


```
					// These two get copied to the book folder, but while we're doing origami layout
					// I don't want to, in case the user doesn't use them after all.

```
seems incomplete. They get copied later? But we might just be doing origami currently?
Then how about something like, "If the user does add a video or widget, these placeholder .svgs will get copied to the book folder and used from there. But we don't copy to the book folder while the user is still in origami in case the user doesn't actually add the video or widget. So while origami is open, it hits this path and we grab the .svgs from their original locations.
</blockquote></details>

I meant it literally...I could not find the source. Since you did, I can now improve the comment!
I liked your version of the other part.
___
*[`src/BloomExe/web/BloomServer.cs` line 589 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5569#-NM13u3yEsFdTEJm6Cn6:-NM4u1UU4cjjmhWz-3Ce:b-rs8qwt) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/94a45f9f832ad3eebd51851733908e440b885094/src/BloomExe/web/BloomServer.cs#L589)):*

I'm probably missing something, but I'm seeing a change in logic here. Previously it seems we only did this for preview for cover images. Now it appears we do it for all book preview images.

A comment would probably help our future selves put it together.
</blockquote></details>

I've added one. Yes, it is a logic change, not directly related to the purpose of this PR. Basically, I did it to correct what feels to me like a bug. Isn't the whole point of resizing images to help manage memory for the preview? It isn't going to help much in edit mode, where we're only dealing with one page's worth of images, nor in PDF creation, where we insert a marker to prevent it. What's left but the preview? But maybe I'm missing something and this shouldn't be changed. If you're remembering something I've missed, please persist.
___
*[`src/content/templates/template books/Digital Comic Book/placeHolder.png` line  at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5569#-NM129dy1OXpOnmijhhl:-NM4lDRLDNJxEcbSpzqg:b-7fsh1j) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/b8581ad854a5deccd17f345bd1edf89b7d29ac7e/src/content/templates/template%20books/Digital%20Comic%20Book/placeHolder.png#L)):*

Is every template supposed to have its own placeHolder.png?
Seems like we should be able to just ship one (or at least only one in the templates dir).
If we must maintain one per template, I notice that `Activity` and `Paper Comic Book` don't have it, either. (Also the `Full Screen` template under experiments.)
</blockquote></details>

It's normal for book folders to contain all the images the book needs. One reason is so that the book can render in a browser, and doesn't depend on BloomServer to find components elsewhere. With the changes here placeholder won't be found when we're showing a preview of these templates, either, though it gets found and inserted while we're making a book from them. So the simplest thing is to add it to them all. If we want to save a few 7kbs in the installer, we could make another special case for it. But for now we decided to add the two other missing ones.

andrew-polk (changes requested):
Yes, thanks

> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5569#-:-NM52HXb6QZHsC89poaQ:b-izbj8h)* status: all files reviewed, 2 unresolved discussions (waiting on @JohnThomson)

___
*[`src/BloomBrowserUI/bookEdit/toolbox/talkingBook/audioRecording.less` line 134 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5569#-NM0yICs-GvqzgoNhVAH:-NM52NQF-73tC_wtLyGA:brguoof) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/b8581ad854a5deccd17f345bd1edf89b7d29ac7e/src/BloomBrowserUI/bookEdit/toolbox/talkingBook/audioRecording.less#L134)):*

They don't need it because the build process puts the image is in the same folder as audioRecording.css
</blockquote></details>

It confused me because you added it to one (play_enabled a few lines up) but not the others.
We should probably be consistent. (Or is that file truly handled differently? In that case, we should comment.)
___
*[`src/BloomExe/web/BloomServer.cs` line 589 at r1](https://reviewable.io/reviews/bloombooks/bloomdesktop/5569#-NM13u3yEsFdTEJm6Cn6:-NM51NwdDPRc_QBC2jz9:b-vm3umy) ([raw file](https://github.com/bloombooks/bloomdesktop/blob/94a45f9f832ad3eebd51851733908e440b885094/src/BloomExe/web/BloomServer.cs#L589)):*

I've added one. Yes, it is a logic change, not directly related to the purpose of this PR. Basically, I did it to correct what feels to me like a bug. Isn't the whole point of resizing images to help manage memory for the preview? It isn't going to help much in edit mode, where we're only dealing with one page's worth of images, nor in PDF creation, where we insert a marker to prevent it. What's left but the preview? But maybe I'm missing something and this shouldn't be changed. If you're remembering something I've missed, please persist.
</blockquote></details>

I guess I didn't even know we resized images for memory management.
The two cases I knew of for resizing images are:
- Reducing the size of an image the user gives us because it is simply too big to start with.
- Reducing the size of images in published ebooks.

But if you're right that we are trying to manage memory, then doing it for all images in the preview makes sense.
Is there a possibility this might cause performance issues for a large book with lots of large images?

JohnThomson (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5569)* status: 24 of 28 files reviewed, 2 unresolved discussions (waiting on @andrew-polk)

___
*[`src/BloomBrowserUI/bookEdit/toolbox/talkingBook/audioRecording.less` line 134 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5569#-NM0yICs-GvqzgoNhVAH:-NMA31fy1B4Q-0UeDHVR:brgo5gr) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/b8581ad854a5deccd17f345bd1edf89b7d29ac7e/src/BloomBrowserUI/bookEdit/toolbox/talkingBook/audioRecording.less#L134)):*

It confused me because you added it to one (play_enabled a few lines up) but not the others.
We should probably be consistent. (Or is that file truly handled differently? In that case, we should comment.)
</blockquote></details>

Yes, that one is different. It is using a shared image in bloom/images, whereas the others are using images private to talking book co-located with audioRecording.less.
I added a comment
___
*[`src/BloomExe/web/BloomServer.cs` line 589 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5569#-NM13u3yEsFdTEJm6Cn6:-NMAyp270-8hiuE-8Bwg:bczm8px) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/94a45f9f832ad3eebd51851733908e440b885094/src/BloomExe/web/BloomServer.cs#L589)):*

I guess I didn't even know we resized images for memory management.
The two cases I knew of for resizing images are:
- Reducing the size of an image the user gives us because it is simply too big to start with.
- Reducing the size of images in published ebooks.

But if you're right that we are trying to manage memory, then doing it for all images in the preview makes sense.
Is there a possibility this might cause performance issues for a large book with lots of large images?
</blockquote></details>

You're right. Apparently we used to and stopped. GetPathToResizedImage is now only relevant for cover images and thumbnails, and only the latter get resized. Talk about a bad name! I've changed it.

andrew-polk (approved):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5569)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @JohnThomson)

PR #5584, opened 2023-01-18
Make PDFs using WebView2 (BL-11437)

StephenMcConnel:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5584)


src/BloomExe/Publish/PDF/MakePdfUsingGeckofxHtmlToPdfProgram.cs
```diff
@@ -15,23 +15,12 @@
...
-	/// The program always seems to fail after it finishes while xulrunner is trying
-	/// to clean up its COM objects.  Running the program standalone reveals this on
-	/// Linux by printing a scary SIGSEGV message and stack trace.  On Windows, it
-	/// brings up the "The Program Has Stopped Working" dialog which is even scarier.
-	/// (I have seen the corresponding Linux dialog once after running this program.)
-	/// However, the problem occurs after the desired output file has been safely
-	/// written to the disk, and is nicely hidden away from the user by running it
-	/// via CommandLineRunner.
-	/// </remarks>
 	class MakePdfUsingGeckofxHtmlToPdfProgram
```
andrew-polk, line 24: Please rename this class.

andrew-polk (commented):
This might mean we can get rid of BloomTestsThatAvoidTheSetupFixture based on the comment there.
Worth a try. If it doesn't work now, at least we can get rid of it when Geckofx goes away completely.

andrew-polk:
Btw, this wasn't a full review. Just a cursory one, at least so far...

StephenMcConnel (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5584)* status: 0 of 20 files reviewed, 1 unresolved discussion (waiting on @andrew-polk)

___
*[`src/BloomExe/Publish/PDF/MakePdfUsingGeckofxHtmlToPdfProgram.cs` line 24 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5584#-NM4yuobPMYxvC8M5E1F-r1-24:-NM5et1X1SkAjTAz5w_u:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/4eeb1019b0cd88b3dffea3085fc3d539e696b34f/src/BloomExe/Publish/PDF/MakePdfUsingGeckofxHtmlToPdfProgram.cs#L24)):*

Please rename this class.
</blockquote></details>

Done.

andrew-polk (changes requested):
> Reviewed 17 of 20 files at r1, 3 of 3 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5584)* status: all files reviewed, 6 unresolved discussions (waiting on @StephenMcConnel)

___
*[`build/getDependencies-windows.sh` line 108 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5584#-NM5yKrKAKTEVs9tSEha:-NM5yKrKAKTEVs9tSEhb:be0b87s) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/ea9d2b299415b85b2bf7b901cbba71d3049872ea/build/getDependencies-windows.sh#L108)):*
> ```Shell
> #     paths: {"pdfjs-viewer.zip!**"=>"DistFiles/pdf"}
> #     VCS: https://github.com/mozilla/pdf.js.git [gh-pages]
> # [5] build: GeckofxHtmlToPdf-Win32-continuous (bt463)
> ```

Doing this manually is probably ok...
The "right" way is to update it by running the script, of course. But you would have to wait until after the PR was merged to do that.

Did you not want to update the Linux one now, too?
___
*[`src/BloomExe/Publish/PDF/PdfMaker.cs` line 133 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5584#-NM5yqFLDNqiMie-2bvK:-NM5yqFLDNqiMie-2bvL:b2tvb5g) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/52a0282eb73d93106fe5a41410b397d1b997a880/src/BloomExe/Publish/PDF/PdfMaker.cs#L133)):*
> ```Smalltalk
> 		/// This would be easy to move to a simple command line program if memory use proves to be a problem.
> 		/// </remarks>
> 		private void RemoveBlankTrailingPageIfAny(string inputPdfPath, string outputPdfPath)
> ```

It is, of course, unfortunate to need to add this logic.
Not knowing how much time you've already put into solving this, it is hard to know how much to ask you to do more.
Worth pairing or just more investigation?
___
*[`src/WebView2PdfMaker/ConversionProgress.cs` line 25 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5584#-NM5qY8nBfJOW8GjGuK0:-NM5qY8nBfJOW8GjGuK1:b-lggayh) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/52a0282eb73d93106fe5a41410b397d1b997a880/src/WebView2PdfMaker/ConversionProgress.cs#L25)):*
> ```Smalltalk
> 			_pdfMaker.Finished += new System.EventHandler(OnPdfMaker_Finished);
> 			//_pdfMaker.StatusChanged += new System.EventHandler<PdfMakingStatus>(this.OnPdfMaker_StatusChanged);
> 			_progressBar.Maximum = 100;
> ```

Is there any point in keeping this progressBar? It appears we only set it to finished.
___
*[`src/WebView2PdfMaker/Program.cs` line 60 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5584#-NM5pS3eDaYXW4icCKT2:-NM5pS3eDaYXW4icCKT3:bo4ifw6) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/52a0282eb73d93106fe5a41410b397d1b997a880/src/WebView2PdfMaker/Program.cs#L60)):*
> ```Smalltalk
> 		private static void Application_ThreadException(object sender, ThreadExceptionEventArgs e)
> 		{
> 			Console.WriteLine("GeckofxHtmlToPdf Thread Exception: " + e.Exception.Message);
> ```

old name (GeckofxHtmlToPdf)
same about a dozen lines down



_Code quote:_
```Smalltalk
GeckofxHtmlToPdf
```
___
*[`src/WebView2PdfMaker/WebView2PdfMaker.csproj` line 1 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5584#-NM5vvc92FN2YoB__UrS:-NM5vvc92FN2YoB__UrT:b-nn31xt) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/52a0282eb73d93106fe5a41410b397d1b997a880/src/WebView2PdfMaker/WebView2PdfMaker.csproj#L1)):*
> ```XML
> ```

If I'm not mistaken, this is the old-style .csproj.
Did you start by copying from the old project?
If so, it is probably easier to just leave as is.
I just bring it up in case we want to go ahead and use the new style/format.
___
*[`src/BloomExe/Publish/PDF/MakePdfUsingExternalPdfMakerProgram.cs` line 124 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5584#-NM5pZHwCy8rCz8C_Rho:-NM5pZHwCy8rCz8C_Rhp:b-u79xtj) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/52a0282eb73d93106fe5a41410b397d1b997a880/src/BloomExe/Publish/PDF/MakePdfUsingExternalPdfMakerProgram.cs#L124)):*
> ```Smalltalk
> 			var progress = new NullProgress();
> 			var res = runner.Start(exePath, arguments, Encoding.UTF8, fromDirectory, 3600, progress,
> 				ProcessGeckofxReporting);
> ```

I'm not seeing any benefit in trying to start this process if we know it won't do anything.
So we might as well remove the code, too, right?
And I guess if we're doing that, we should mark the deleted l10n strings as deprecated in the .xlf?

(Or do you think we might find a way to report progress in the future?)

andrew-polk (commented):
Ok, I've done a full review now.
But I think this one might be worth a second set of eyes. Especially someone who knows more about webview2 than me. @JohnThomson 

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5584#-:-NM5zgqC7o8dcqvxN-W9:bppk7f8)* status: all files reviewed, 6 unresolved discussions (waiting on @StephenMcConnel)

StephenMcConnel (commented):
I've removed the BloomTestsThatAvoidTheSetupFixture namespace and the preceding commentary.

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5584#-:-NMA6xow4yuXYIxqH1Jb:b-t29tqj)* status: all files reviewed, 6 unresolved discussions (waiting on @andrew-polk)

___
*[`build/getDependencies-windows.sh` line 108 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5584#-NM5yKrKAKTEVs9tSEha:-NM6IE3XE8BbOoklejSm:bauu5qe) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/ea9d2b299415b85b2bf7b901cbba71d3049872ea/build/getDependencies-windows.sh#L108)):*

Doing this manually is probably ok...
The "right" way is to update it by running the script, of course. But you would have to wait until after the PR was merged to do that.

Did you not want to update the Linux one now, too?
</blockquote></details>

I'm not even trying to build on Linux for 5.5, and have disabled the Jenkins build on master.  Once WebView2 is standard, nothing will work (or probably build) on Linux anyway.
Running the script is, of course, the "right" way to do this, but I didn't want to wait until later to ensure that the old pdf maker program is absolutely unavailable.  And updating the TeamCity dependencies can't be done until after this PR.
___
*[`src/BloomExe/Publish/PDF/PdfMaker.cs` line 133 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5584#-NM5yqFLDNqiMie-2bvK:-NM6NYLGB9FExDJ7wTqz:b-clhadr) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/52a0282eb73d93106fe5a41410b397d1b997a880/src/BloomExe/Publish/PDF/PdfMaker.cs#L133)):*

It is, of course, unfortunate to need to add this logic.
Not knowing how much time you've already put into solving this, it is hard to know how much to ask you to do more.
Worth pairing or just more investigation?
</blockquote></details>

I'd rather look at this more after the initial check-in, and would be happy to pair in the further investigation.  The set of sizes that cause the problem is mystifying to me.
I spent a few hours looking at this problem, probably half trying to figure it out and half trying to come up with this patch.
___
*[`src/WebView2PdfMaker/ConversionProgress.cs` line 25 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5584#-NM5qY8nBfJOW8GjGuK0:-NM6NvP9DvKPeQsWN9eJ:b-8qr0re) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/52a0282eb73d93106fe5a41410b397d1b997a880/src/WebView2PdfMaker/ConversionProgress.cs#L25)):*

Is there any point in keeping this progressBar? It appears we only set it to finished.
</blockquote></details>

I removed the progressBar.  I also renamed the form to "ConversionHost" since it doesn't really report progress, but it does host the component that does the conversion from HTML to PDF.
___
*[`src/WebView2PdfMaker/Program.cs` line 60 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5584#-NM5pS3eDaYXW4icCKT2:-NM6RBSd22-L8aI4rY4a:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/52a0282eb73d93106fe5a41410b397d1b997a880/src/WebView2PdfMaker/Program.cs#L60)):*

old name (GeckofxHtmlToPdf)
same about a dozen lines down

</blockquote></details>

Done.
___
*[`src/WebView2PdfMaker/WebView2PdfMaker.csproj` line 1 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5584#-NM5vvc92FN2YoB__UrS:-NM6RYWv8Y6C36WQeMn-:boaxqwm) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/52a0282eb73d93106fe5a41410b397d1b997a880/src/WebView2PdfMaker/WebView2PdfMaker.csproj#L1)):*

If I'm not mistaken, this is the old-style .csproj.
Did you start by copying from the old project?
If so, it is probably easier to just leave as is.
I just bring it up in case we want to go ahead and use the new style/format.
</blockquote></details>

I think VS may have created it that way because it's still a WinForms project (even though the form is always hidden).  Keeping it a WinForms program means that it's compatible with Bloom for sharing the WebView2 and CommandLineParser assemblies.
___
*[`src/BloomExe/Publish/PDF/MakePdfUsingExternalPdfMakerProgram.cs` line 124 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5584#-NM5pZHwCy8rCz8C_Rho:-NM6J3AZ0HUCkd4JPOxO:b2a3wal) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/52a0282eb73d93106fe5a41410b397d1b997a880/src/BloomExe/Publish/PDF/MakePdfUsingExternalPdfMakerProgram.cs#L124)):*

I'm not seeing any benefit in trying to start this process if we know it won't do anything.
So we might as well remove the code, too, right?
And I guess if we're doing that, we should mark the deleted l10n strings as deprecated in the .xlf?

(Or do you think we might find a way to report progress in the future?)
</blockquote></details>

Progress reporting won't work, but runner.Start() is how the external program is run.  On the other hand, the method for dealing with status change strings from the process doesn't need to do anything.

andrew-polk (changes requested):
> Reviewed 9 of 9 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5584)* status: all files reviewed, 2 unresolved discussions (waiting on @StephenMcConnel)

___
*[`src/BloomExe/Publish/PDF/MakePdfUsingExternalPdfMakerProgram.cs` line 124 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5584#-NM5pZHwCy8rCz8C_Rho:-NMAY74RD_YY8tD64s1P:b2aqhkn) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/52a0282eb73d93106fe5a41410b397d1b997a880/src/BloomExe/Publish/PDF/MakePdfUsingExternalPdfMakerProgram.cs#L124)):*

Progress reporting won't work, but runner.Start() is how the external program is run.  On the other hand, the method for dealing with status change strings from the process doesn't need to do anything.
</blockquote></details>

Thanks for clarifying.

On deprecating the l10n strings:
I just realized that we are not only losing the information about progress through the pages, but also the starting and finished messages. Do we want to reinstate those somehow?

StephenMcConnel (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5584)* status: all files reviewed, 2 unresolved discussions (waiting on @andrew-polk)

___
*[`src/BloomExe/Publish/PDF/MakePdfUsingExternalPdfMakerProgram.cs` line 124 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5584#-NM5pZHwCy8rCz8C_Rho:-NMUUOdT-KUw-hPhz6id:bvd8xnx) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/52a0282eb73d93106fe5a41410b397d1b997a880/src/BloomExe/Publish/PDF/MakePdfUsingExternalPdfMakerProgram.cs#L124)):*

Thanks for clarifying.

On deprecating the l10n strings:
I just realized that we are not only losing the information about progress through the pages, but also the starting and finished messages. Do we want to reinstate those somehow?
</blockquote></details>

I don't think the starting and finished messages add anything without the scroll bar advancing (and replacing the starting message).  The finished message would show only as a flash before the program moves on to the next step of compressing the pdf output.  There's already a message displayed "Making PDF from HTML" which conveys what is happening in that step of the process.  I agree that it's unfortunate that we've lost the progress reporting with switching to WebView2, but that's a limitation of the WebView2 API.

andrew-polk (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5584)* status: all files reviewed, 1 unresolved discussion (waiting on @StephenMcConnel)

___
*[`src/BloomExe/Publish/PDF/PdfMaker.cs` line 133 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5584#-NM5yqFLDNqiMie-2bvK:-NMUZ8Yk52GToA4Up1SP:b6tx8hr) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/52a0282eb73d93106fe5a41410b397d1b997a880/src/BloomExe/Publish/PDF/PdfMaker.cs#L133)):*

I'd rather look at this more after the initial check-in, and would be happy to pair in the further investigation.  The set of sizes that cause the problem is mystifying to me.
I spent a few hours looking at this problem, probably half trying to figure it out and half trying to come up with this patch.
</blockquote></details>

Ok, I've added a checkbox in the issue.
___
*[`src/BloomExe/Publish/PDF/MakePdfUsingExternalPdfMakerProgram.cs` line 124 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5584#-NM5pZHwCy8rCz8C_Rho:-NMUZ-Dk70Bd8ooFUyE9:b8pgrr8) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/52a0282eb73d93106fe5a41410b397d1b997a880/src/BloomExe/Publish/PDF/MakePdfUsingExternalPdfMakerProgram.cs#L124)):*

I don't think the starting and finished messages add anything without the scroll bar advancing (and replacing the starting message).  The finished message would show only as a flash before the program moves on to the next step of compressing the pdf output.  There's already a message displayed "Making PDF from HTML" which conveys what is happening in that step of the process.  I agree that it's unfortunate that we've lost the progress reporting with switching to WebView2, but that's a limitation of the WebView2 API.
</blockquote></details>

Ok, thanks. 
Then if we are not using the l10n strings anymore, I think we need to add deprecation notes in the .xlf file.

StephenMcConnel (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5584)* status: 22 of 23 files reviewed, 1 unresolved discussion (waiting on @andrew-polk)

___
*[`src/BloomExe/Publish/PDF/MakePdfUsingExternalPdfMakerProgram.cs` line 124 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5584#-NM5pZHwCy8rCz8C_Rho:-NMVxb4RBrXl97DihmS3:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/52a0282eb73d93106fe5a41410b397d1b997a880/src/BloomExe/Publish/PDF/MakePdfUsingExternalPdfMakerProgram.cs#L124)):*

Ok, thanks. 
Then if we are not using the l10n strings anymore, I think we need to add deprecation notes in the .xlf file.
</blockquote></details>

Done.

andrew-polk (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5584)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @StephenMcConnel)

PR #5587, opened 2023-01-19
Fix a race condition displaying and renaming a new book (BL-11524)

JohnThomson:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5587)

### DIFF
--- a/src/BloomExe/Book/Book.cs
+++ b/src/BloomExe/Book/Book.cs
@@ -899,8 +899,28 @@ namespace Bloom.Book
 			htmlDom.AddCreationType(LockedDown ? "translation" : "original");
 		}
 
+		// Generally BringBookUpToDate only needs doing once, so keep track of whether we have.
+		// Since we don't currently have any way to know whether a book on disk needs updating,
+		// this is true (as far as we know) until BringBookUpToDate is first called for this session.
+		private bool _needsUpdate = true;
+
+		public void EnsureUpToDate()
+		{
+			if (_needsUpdate)
+				BringBookUpToDate(new NullProgress());
+		}
+
+		/// <summary>
+		/// Make any needed changes to make a book which might have come from an old version of Bloom
+		/// consistent with the current data model. Also makes sure it has the current XMatter
+		/// and a folder name consistent with its title (unless folder name has been overridden).
+		/// Consider using EnsureUpToDate() unless you know for sure that the book object is new
+		/// or that something has changed which requires BBUD to happen again. Usually it only
+		/// needs doing at most once, and it is slow.
+		/// </summary>
 		public void BringBookUpToDate(IProgress progress, bool forCopyOfUpToDateBook = false)
 		{
+			_needsUpdate = false;
 			_pagesCache = null;
 			string oldMetaData = string.Empty;
 			if (RobustFile.Exists(BookInfo.MetaDataPath))
--- a/src/BloomExe/Book/BookSelection.cs
+++ b/src/BloomExe/Book/BookSelection.cs
@@ -42,7 +42,7 @@ namespace Bloom.Book
 			// BringUpToDate, typically only in unit tests.
 			if (book != null && book.BookData != null && book.IsEditable)
 			{
-				book.BringBookUpToDate(new NullProgress());
+				book.EnsureUpToDate();
 			}
 
 			_currentSelection = book;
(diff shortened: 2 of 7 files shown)


hatton (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5587)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @JohnThomson)

___
*[`src/BloomExe/Book/Book.cs` line 902 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5587#-NMEd00M0Ppa8KVmy06Z:-NMEd00M0Ppa8KVmy06_:b-8qkb02) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/49fb0a6edff8134d2fd5092073b6634bd3b6ebdd/src/BloomExe/Book/Book.cs#L902)):*
> ```Smalltalk
> 		}
> 
> 		// This is useful when we need to BBUD immediately, but we are about to do another
> ```

Is there some way to fix the problem without this hack? It's kinda shocking.

hatton (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5587)* status: 0 of 7 files reviewed, 2 unresolved discussions (waiting on @JohnThomson)

___
*[`src/BloomExe/Book/Book.cs` line 921 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5587#-NMFh-Y14bbKfj1Vclz9:-NMFh-Y14bbKfj1VclzA:b-vb1i2b) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/2321b6720a2338077b2e3eda6d69d63cdb0b5090/src/BloomExe/Book/Book.cs#L921)):*
> ```Smalltalk
> 		public void BringBookUpToDate(IProgress progress, bool forCopyOfUpToDateBook = false)
> 		{
> 			_hasBeenBroughtUpToDate = true;
> ```

The polarity here seems backwards. It is common in software to have a `dirty` flag, not a `clean` tag.  I think this should be `needsUpdate` or some such.

Beyond convention, there is a semantic problem. If we do an update, and then change a book setting (future) that requires an update, well it is still true that the book was updated. However, we need it updated again, i.e., `needsUpdate = true`.

JohnThomson (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5587)* status: 0 of 7 files reviewed, 2 unresolved discussions (waiting on @hatton)

___
*[`src/BloomExe/Book/Book.cs` line 902 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5587#-NMEd00M0Ppa8KVmy06Z:-NMFrTE63JiapIk-6Kxf:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/49fb0a6edff8134d2fd5092073b6634bd3b6ebdd/src/BloomExe/Book/Book.cs#L902)):*

Is there some way to fix the problem without this hack? It's kinda shocking.
</blockquote></details>

Done.
___
*[`src/BloomExe/Book/Book.cs` line 921 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5587#-NMFh-Y14bbKfj1Vclz9:-NMFrVm58kSVm8XvT0Nj:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/2321b6720a2338077b2e3eda6d69d63cdb0b5090/src/BloomExe/Book/Book.cs#L921)):*

The polarity here seems backwards. It is common in software to have a `dirty` flag, not a `clean` tag.  I think this should be `needsUpdate` or some such.

Beyond convention, there is a semantic problem. If we do an update, and then change a book setting (future) that requires an update, well it is still true that the book was updated. However, we need it updated again, i.e., `needsUpdate = true`.

</blockquote></details>

Done.

hatton (approved):
> Reviewed 6 of 7 files at r2, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5587)* status: all files reviewed, 2 unresolved discussions (waiting on @JohnThomson)

hatton (approved):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomdesktop/5587)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @JohnThomson)

PR #5597, opened 2023-01-23
More TeamCollection analytics (BL-11867)

JohnThomson:
Joining a TC: reports whether we created a new local collection, opened an existing one normally, or did a first-time merge.
Errors that put things in Lost and Found now generate a TeamCollectionError analytics report.

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5597)

### DIFF
--- a/src/BloomBrowserUI/teamCollection/CollectionHistoryTable.tsx
+++ b/src/BloomBrowserUI/teamCollection/CollectionHistoryTable.tsx
@@ -60,7 +60,8 @@ const kEventTypes = [
     "Uploaded",
     "Force Unlock",
     "Import Spreadsheet",
-    "Sync Problem"
+    "Sync Problem",
+    "Deleted"
 ]; // REVIEW maybe better to do this in c# and just send it over?
 
 export const CollectionHistoryTable: React.FunctionComponent<{
--- a/src/BloomBrowserUI/teamCollection/CollectionHistoryTable.tsx
+++ b/src/BloomBrowserUI/teamCollection/CollectionHistoryTable.tsx
@@ -59,7 +59,8 @@ const kEventTypes = [
     "Renamed",
     "Uploaded",
     "Force Unlock",
-    "Import Spreadsheet"
+    "Import Spreadsheet",
+    "Sync Problem"
 ]; // REVIEW maybe better to do this in c# and just send it over?
 
 export const CollectionHistoryTable: React.FunctionComponent<{
--- a/src/BloomExe/CollectionTab/CollectionModel.cs
+++ b/src/BloomExe/CollectionTab/CollectionModel.cs
@@ -354,6 +354,7 @@ namespace Bloom.CollectionTab
 					}
 				}
 				var bookName = _bookSelection.CurrentSelection.NameBestForUserDisplay;
+				var bookId = _bookSelection.CurrentSelection.ID;
 				var confirmRecycleDescription = L10NSharp.LocalizationManager.GetString("CollectionTab.ConfirmRecycleDescription", "The book '{0}'");
 				if (ConfirmRecycleDialog.JustConfirm(string.Format(confirmRecycleDescription, bookName), false, "Palaso"))
 				{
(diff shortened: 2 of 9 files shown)


andrew-polk (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5597)* status: all files reviewed, 2 unresolved discussions (waiting on @JohnThomson)

___
*[`src/BloomExe/TeamCollection/TeamCollection.cs` line 1559 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5597#-NNcWTpV9Ix5sglT0wse:-NNcWTpV9Ix5sglT0wsf:b-17yqf3) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/fd3e5375fe3a5c151a5d0d111399d30723bd6ee4/src/BloomExe/TeamCollection/TeamCollection.cs#L1559)):*
> ```Smalltalk
> 		/// <summary>
> 		/// This overload reports the problem to the progress box, log, and Analytics. If possible, use the overload which
> 		/// also adds the report to the book's history.
> ```

I don't see that this is used outside the class. So should we just make it private? (And remove the comment or at least the second part of the comment?



_Suggestion:_
```Smalltalk
		/// ... If possible, use the overload which
		/// also adds the report to the book's history.
```
___
*[`src/BloomExe/TeamCollection/TeamCollection.cs` line 1602 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5597#-NNcW3J60jBVjdiHZ0T5:-NNcW3J60jBVjdiHZ0T6:buxrr6e) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/fd3e5375fe3a5c151a5d0d111399d30723bd6ee4/src/BloomExe/TeamCollection/TeamCollection.cs#L1602)):*
> ```Smalltalk
> 		}
> 
> 		private void MakeYouTrackIssueIfWanted(IWebSocketProgress progress, bool alsoMakeYouTrackIssue, string msg)
> ```

I'm probably just missing something obvious, but I don't understand why this isn't
`private void MakeYouTrackIssue(IWebSocketProgress progress, string msg)` 
instead. I.e. why pass in a boolean at all?

JohnThomson (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5597)* status: 8 of 9 files reviewed, 2 unresolved discussions (waiting on @andrew-polk)

___
*[`src/BloomExe/TeamCollection/TeamCollection.cs` line 1559 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5597#-NNcWTpV9Ix5sglT0wse:-NNgnOOsBClwWsm8qj5w:bhcxt4d) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/fd3e5375fe3a5c151a5d0d111399d30723bd6ee4/src/BloomExe/TeamCollection/TeamCollection.cs#L1559)):*

I don't see that this is used outside the class. So should we just make it private? (And remove the comment or at least the second part of the comment?
</blockquote></details>

It is private; this is the default in C#. I've modified the name and comment to try to make my intentions clearer.
___
*[`src/BloomExe/TeamCollection/TeamCollection.cs` line 1602 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5597#-NNcW3J60jBVjdiHZ0T5:-NNgqXTp6W11q0uw8zI2:b6xen9g) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/fd3e5375fe3a5c151a5d0d111399d30723bd6ee4/src/BloomExe/TeamCollection/TeamCollection.cs#L1602)):*

I'm probably just missing something obvious, but I don't understand why this isn't
`private void MakeYouTrackIssue(IWebSocketProgress progress, string msg)` 
instead. I.e. why pass in a boolean at all?
</blockquote></details>

At the time, I was thinking that there's already a test at the start of the method to determine whether the issue should actually be made, and both callers have a similar additional test, so this seemed simpler. It also makes clear that an issue won't actually be created in every case.
But with two weeks' perspective I think I agree with you. Changed.

andrew-polk (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5597)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @JohnThomson)

PR #5603, opened 2023-01-25
Show image info tooltip only when no bubbles selected (BL-11738)

sujeffreyl:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5603)

### DIFF
--- a/src/BloomBrowserUI/bookEdit/js/bloomImages.ts
+++ b/src/BloomBrowserUI/bookEdit/js/bloomImages.ts
@@ -1,5 +1,9 @@
 import "../../lib/jquery.resize"; // makes jquery resize work on all elements
-import { getWithConfig, postJson } from "../../utils/bloomApi";
+import {
+    getWithConfig,
+    getWithConfigAsync,
+    postJson
+} from "../../utils/bloomApi";
 
 // Enhance: this could be turned into a Typescript Module with only two public methods
 
@@ -387,20 +391,51 @@ function getImgFromContainer(imageContainer: HTMLElement) {
         .not(".bloom-ui img"); // e.g. cog.svg
 }
 
-function SetImageTooltip(container: HTMLElement) {
+export function DisableImageTooltip(container: HTMLElement) {
+    if (container) {
+        container.title = "";
+    }
+}
+
+export function EnableImageTooltip(container: HTMLElement) {
+    const dataTitle = container.getAttribute("data-title");
+
+    // If dataTitle is null for some unexpected reason, let's just leave it as is.
+    if (dataTitle !== null) {
+        container.title = dataTitle;
+    }
+}
+
+async function SetImageTooltip(container: HTMLElement) {
+    const title = await DetermineImageTooltipAsync(container);
+
+    // We use data-title to store what the tooltip should be, regardless of whether the tooltip should actually be currently visible
+    // Use the real title attribute to show the tooltip only when desired
+    // (e.g. show when no bubble selected but hide when a bubble is selected)
+    container.setAttribute("data-title", title);
+
+    // Only show the tooltip if no bubble is currently selected.
+    if (theOneBubbleManager.getActiveElement()) {
+        DisableImageTooltip(container);
+    } else {
+        EnableImageTooltip(container);
+    }
+}
+
+async function DetermineImageTooltipAsync(
+    container: HTMLElement
+): Promise<string> {
     const containerJQ = $(container);
     const imgElement = $(container).find("img");
 
     if (!imgElement) {
-        container.title = "";
-        return;
+        return "";
     }
     const url = GetRawImageUrl(imgElement);
     // Don't try to go getting image info for a built in Bloom image (like cogGrey.svg).
     // It'll just throw an exception.
     if (url.startsWith("/bloom/")) {
-        container.title = "";
-        return;
+        return "";
     }
 
     const targetDpiWidth = Math.ceil((300 * containerJQ.width()) / kBrowserDpi);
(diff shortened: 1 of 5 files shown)


andrew-polk (changes requested):
> Reviewed 2 of 3 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5603)* status: 2 of 3 files reviewed, 1 unresolved discussion (waiting on @sujeffreyl)

___
*[`src/BloomBrowserUI/bookEdit/js/bloomImages.ts` line 422 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5603#-NMf8LgGAzwG_WmvyKIn:-NMf8LgGAzwG_WmvyKIo:b-s8uqqt) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/f88336dba1433e5a331b92a52caf0bc72120a5c5/src/BloomBrowserUI/bookEdit/js/bloomImages.ts#L422)):*
> ```XML
>     } else {
>         EnableImageTooltip(container);
>     }
> ```

Isn't a bubble being selected different than the image being selected?
How does the logic work which makes the on-hover buttons (select image, paste, etc.) only appear when the image is "selected"? From playing with it, it seems different than if a bubble is selected. Can we use the same logic for this?

Also, that would (hopefully) disentangle the showing/hiding of the tooltip from bubbleManager. Ideally, bubbleManager should know nothing about it.

sujeffreyl (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5603)* status: 2 of 3 files reviewed, 1 unresolved discussion (waiting on @andrew-polk)

___
*[`src/BloomBrowserUI/bookEdit/js/bloomImages.ts` line 422 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5603#-NMf8LgGAzwG_WmvyKIn:-NMjFJzq2lmaPL-Tso0z:b1nzpk9) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/f88336dba1433e5a331b92a52caf0bc72120a5c5/src/BloomBrowserUI/bookEdit/js/bloomImages.ts#L422)):*

Isn't a bubble being selected different than the image being selected?
How does the logic work which makes the on-hover buttons (select image, paste, etc.) only appear when the image is "selected"? From playing with it, it seems different than if a bubble is selected. Can we use the same logic for this?

Also, that would (hopefully) disentangle the showing/hiding of the tooltip from bubbleManager. Ideally, bubbleManager should know nothing about it.
</blockquote></details>

It does sound a attractive from the UI perspective to make it align with the on-hover buttons.

Unfortunately, since I don't believe it's possible to disable the tooltip via CSS (only via setting title), that means we need to track the Javascript that adds the .bloom-hideImageButtons and .bloom-uiSuppressImageButtons classes.   Some references are in bloomImages (fine), 2 are in bubbleManager (not improving anything) and 2 more are in motionTool (entangling another file).  So is that even any better than the original situation?

Another con: When you load a page, the focus is in a bubble. In my PR, therefore the tooltip will not show up. However, the image editing buttons sometimes (!!!!!!!!) do show up initially, despite a bubble having focus. (This is arguably a problem with the image editing buttons).

Well... I guess it's probably still worth it to follow the on-hover image editing buttons?  Lining up with the on-hover image editing buttons does make a lot of sense from a user model point of view.  Having one model instead of two is also nice.
Thoughts before I do it?

andrew-polk (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5603)* status: 2 of 3 files reviewed, 1 unresolved discussion (waiting on @sujeffreyl)

___
*[`src/BloomBrowserUI/bookEdit/js/bloomImages.ts` line 422 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5603#-NMf8LgGAzwG_WmvyKIn:-NMjRGCd2v-RSfOmGsYj:b-4qgdhr) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/f88336dba1433e5a331b92a52caf0bc72120a5c5/src/BloomBrowserUI/bookEdit/js/bloomImages.ts#L422)):*

It does sound a attractive from the UI perspective to make it align with the on-hover buttons.

Unfortunately, since I don't believe it's possible to disable the tooltip via CSS (only via setting title), that means we need to track the Javascript that adds the .bloom-hideImageButtons and .bloom-uiSuppressImageButtons classes.   Some references are in bloomImages (fine), 2 are in bubbleManager (not improving anything) and 2 more are in motionTool (entangling another file).  So is that even any better than the original situation?

Another con: When you load a page, the focus is in a bubble. In my PR, therefore the tooltip will not show up. However, the image editing buttons sometimes (!!!!!!!!) do show up initially, despite a bubble having focus. (This is arguably a problem with the image editing buttons).

Well... I guess it's probably still worth it to follow the on-hover image editing buttons?  Lining up with the on-hover image editing buttons does make a lot of sense from a user model point of view.  Having one model instead of two is also nice.
Thoughts before I do it?
</blockquote></details>

Yes, I think we need one source of truth for whether an image is selected. It sounds like we could really improve that model, but I'm not sure this tweak is worth diving into that.

I'm not sure I like this idea, but you could, in theory, add an event listener on hover which looks for the class (or a new one) and then sets or removes title from data-title.

sujeffreyl (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5603)* status: 1 of 5 files reviewed, 1 unresolved discussion (waiting on @andrew-polk)

___
*[`src/BloomBrowserUI/bookEdit/js/bloomImages.ts` line 422 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5603#-NMf8LgGAzwG_WmvyKIn:-NMoLYqg9-C-pCz3GlP2:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/f88336dba1433e5a331b92a52caf0bc72120a5c5/src/BloomBrowserUI/bookEdit/js/bloomImages.ts#L422)):*

Yes, I think we need one source of truth for whether an image is selected. It sounds like we could really improve that model, but I'm not sure this tweak is worth diving into that.

I'm not sure I like this idea, but you could, in theory, add an event listener on hover which looks for the class (or a new one) and then sets or removes title from data-title.

</blockquote></details>

Done.

andrew-polk (approved):
Nice improvements. Thanks.

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5603#-:-NMpLH8B6w40mAtUxcGx:bjuqd5k)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @sujeffreyl)

PR #5610, opened 2023-02-02
Publish things using whitelists, not blacklists, to copy files (BL-10869)

JohnThomson:
Changes how we copy and zip the files in a bloom book folder when making various kinds of publications, including BloomPacks.

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5610)


hatton (changes requested):
> Reviewed 17 of 21 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5610)* status: 17 of 21 files reviewed, 18 unresolved discussions (waiting on @JohnThomson)

___
*[`src/BloomExe/Book/BookCompressor.cs` line 18 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5610#-NOMGY9P9XiwLT_iu0FA:-NOMGY9P9XiwLT_iu0FB:b-duzo5a) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/fb66b947d4a694452d328409872d09b330a92969/src/BloomExe/Book/BookCompressor.cs#L18)):*
> ```Smalltalk
> using System.Collections.Generic;
> using Bloom.Collection;
> using Bloom.CollectionTab;
> ```

Seems odd that this class relies on a UI component?
___
*[`src/BloomExe/Book/BookCompressor.cs` line 26 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5610#-NOMH-C3ACJ2KgUjuDqO:-NOMH-C3ACJ2KgUjuDqP:b-6k9695) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/fb66b947d4a694452d328409872d09b330a92969/src/BloomExe/Book/BookCompressor.cs#L26)):*
> ```Smalltalk
> 	public class BookCompressor
> 	{
> 		public const string BloomPubExtensionWithDot = ".bloompub";
> ```

Is there a more natural place for this bit of info?
___
*[`src/BloomExe/Book/BookCompressor.cs` line 30 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5610#-NOMGotY6OM3D4uzoBfa:-NOMGotY6OM3D4uzoBfb:besxa3o) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/fb66b947d4a694452d328409872d09b330a92969/src/BloomExe/Book/BookCompressor.cs#L30)):*
> ```Smalltalk
> 
> 		// these image files may need to be reduced before being stored in the compressed output file
> 		public static readonly string[] ImageFileExtensions = { ".tif", ".tiff", ".png", ".bmp", ".jpg", ".jpeg" };
> ```

Is there a more natural class to hang this off of? 
___
*[`src/BloomExe/Book/BookCompressor.cs` line 81 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5610#-NOMHA7v3oI71GLtQWy2:-NOMHA7v3oI71GLtQWy3:b-ng8pev) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/fb66b947d4a694452d328409872d09b330a92969/src/BloomExe/Book/BookCompressor.cs#L81)):*
> ```Smalltalk
> 				foreach (var bookFolder in Directory.GetDirectories(directoryToCompress))
> 				{
> 					GetBookOverrides(bookFolder, overrides);
> ```

This looks like it returns something that we ignore. Clearer would be `AddBookOverrides(....)`
___
*[`src/BloomExe/Book/BookCompressor.cs` line 121 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5610#-NOMHNdPC_q5mvikLtNy:-NOMHNdPC_q5mvikLtNz:b-busl8v) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/fb66b947d4a694452d328409872d09b330a92969/src/BloomExe/Book/BookCompressor.cs#L121)):*
> ```Smalltalk
> 		}
> 
> 		private static void GetBookOverrides(string bookFolderPath, Dictionary<string, byte[]> overrides)
> ```

It's not clear what an "override" is. Would `AddToFileToAlwaysAccept()` be accurate?
___
*[`src/BloomExe/Book/BookFileFilter.cs` line 21 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5610#-NOM84-d59anCdH1ReUs:-NOM84-d59anCdH1ReUt:bhr2y6j) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/fb66b947d4a694452d328409872d09b330a92969/src/BloomExe/Book/BookFileFilter.cs#L21)):*
> ```Smalltalk
> 
> 	/// <summary>
>     /// BookFileFilter is a configurable filter for determining which books should be copied either into a temporary folder or
> ```

Should this be "determining which **files** should"?
___
*[`src/BloomExe/Book/BookFileFilter.cs` line 52 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5610#-NOM989e6HvVJ1fMoLsW:-NOM989e6HvVJ1fMoLsX:b-be0ipp) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/fb66b947d4a694452d328409872d09b330a92969/src/BloomExe/Book/BookFileFilter.cs#L52)):*
> ```Smalltalk
> 				Debug.Assert(value == true, "We don't support reverting to not-for-edit");
> 				_forEdit = value;
> 				ForInteractive = true;
> ```

Can add a comment? It's not immediately clear how "ForInteractive" relates to editing. I assume it is that when editing, we need *everything*?
___
*[`src/BloomExe/Book/BookFileFilter.cs` line 53 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5610#-NOM8zjO73px-kN-7AuX:-NOM8zjO73px-kN-7AuY:bidm8b5) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/fb66b947d4a694452d328409872d09b330a92969/src/BloomExe/Book/BookFileFilter.cs#L53)):*
> ```Smalltalk
> 				_forEdit = value;
> 				ForInteractive = true;
> 				// I'm not sure what this is for, possibly storing month names etc. in Wall Calendar,
> ```

Yes, that's all it ever got used for.
___
*[`src/BloomExe/Book/BookFileFilter.cs` line 66 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5610#-NOM9SI67f1GovEXh16S:-NOM9SI67f1GovEXh16T:bmkc7p1) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/fb66b947d4a694452d328409872d09b330a92969/src/BloomExe/Book/BookFileFilter.cs#L66)):*
> ```Smalltalk
> 
> 		/// <summary>
> 		/// Copying for a destination where the user can interact with the book.
> ```

But also for editing? (see comment above)
___
*[`src/BloomExe/Book/BookFileFilter.cs` line 108 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5610#-NOMA1Og1bk76eufEBTb:-NOMA1Og1bk76eufEBTc:bqao7s6) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/fb66b947d4a694452d328409872d09b330a92969/src/BloomExe/Book/BookFileFilter.cs#L108)):*
> ```Smalltalk
> 		/// pattern, like passing a BloomBookOrder for upload.
> 		/// </summary>
> 		public void AddException(string path, bool accept = true)
> ```

Two functions, `AlwaysAccept()` and `AlwaysReject()` would be worth it for clarity.
___
*[`src/BloomExe/Book/BookFileFilter.cs` line 111 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5610#-NOMAUMIEnM96bcXLh0v:-NOMAUMIEnM96bcXLh0w:b-19fcgu) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/fb66b947d4a694452d328409872d09b330a92969/src/BloomExe/Book/BookFileFilter.cs#L111)):*
> ```Smalltalk
> 		{
> 			if (!SIL.PlatformUtilities.Platform.IsLinux)
> 				path = path.ToLowerInvariant();
> ```

The 5 instances of this could become `path = normalizePath(path)` with the platform check then encapsulated in that.
___
*[`src/BloomExe/Book/BookFileFilter.cs` line 147 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5610#-NOMBs6BBJouD9fAjCxU:-NOMBs6C57rWkgRK6geg:b-bv3gb) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/fb66b947d4a694452d328409872d09b330a92969/src/BloomExe/Book/BookFileFilter.cs#L147)):*
> ```Smalltalk
> 					// This is mainly used in unit tests, but could be helpful
> 					// in some other situation where we don't already have one.
> 					_dom = new HtmlDom(RobustFile.ReadAllText(_theOneHtmlPath, Encoding.UTF8), true);
> ```

If this fails (e.g. due to Cloud interference), does the user get a good message?
___
*[`src/BloomExe/Book/BookFileFilter.cs` line 326 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5610#-NOMCNo439_HT1RRzNtD:-NOMCNo439_HT1RRzNtE:bed0335) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/fb66b947d4a694452d328409872d09b330a92969/src/BloomExe/Book/BookFileFilter.cs#L326)):*
> ```Smalltalk
> 				var pathFromBookFolder = path.Substring(_bookFolderPrefixLength);
> 				var dest = Path.Combine(destinationFolder, pathFromBookFolder);
> 				Directory.CreateDirectory(Path.GetDirectoryName(dest));
> ```

If this fails for any reason, does the user get a good message?
___
*[`src/BloomExe/Collection/CollectionFilter.cs` line 16 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5610#-NOMCkrk8g6OoFq0bUB8:-NOMCkrk8g6OoFq0bUB9:b-txj0x8) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/fb66b947d4a694452d328409872d09b330a92969/src/BloomExe/Collection/CollectionFilter.cs#L16)):*
> ```Smalltalk
> 	/// entire collection by combining the book filters of all the books in it.
> 	/// </summary>
> 	internal class CollectionFilter: IFilter
> ```

I keep reading this to mean it filters collections. How about `CollectionFileFilter`, in parallel with your `BookFileFilter`?
___
*[`src/BloomExe/Collection/CollectionFilter.cs` line 46 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5610#-NOMD4WIDb3uWHXm5Hba:-NOMD4WIDb3uWHXm5Hbb:b-q8z3hu) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/fb66b947d4a694452d328409872d09b330a92969/src/BloomExe/Collection/CollectionFilter.cs#L46)):*
> ```Smalltalk
> 				_rootFolder = root;
> 			}
> 			Debug.Assert(_rootFolder == root, "CollectionFilter requires all books to be in the same parent collection");
> ```

It would be nice if we had a version of Assert that will through on a tester. Absent that, you might as well throw and Exception.
___
*[`src/BloomExe/CollectionTab/CollectionModel.cs` line 717 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5610#-NOMDem7DJoufVFkpfdm:-NOMDem7DJoufVFkpfdn:b-zfsxwx) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/fb66b947d4a694452d328409872d09b330a92969/src/BloomExe/CollectionTab/CollectionModel.cs#L717)):*
> ```Smalltalk
> 			filter.AddException("thumbnail-70.png", false);
> 			// The Bloom receiving the pack already has this, so we can save a little space.
> 			filter.AddException("placeHolder.png", false);
> ```

Searching my drive for placeHolder*.png suggests we need a regex here :-)  Then you can use that on `thumbnail-` as well.


___
*[`src/BloomExe/Publish/Android/AndroidPublishSettings.cs` line 27 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5610#-NOMHfnPBfEsLhJOEYjf:-NOMHfnPBfEsLhJOEYjg:bb752ax) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/fb66b947d4a694452d328409872d09b330a92969/src/BloomExe/Publish/Android/AndroidPublishSettings.cs#L27)):*
> ```Smalltalk
> 		public HashSet<string> AudioLanguagesToExclude;
> 
> 		public string[] AudioLangugesToInclude
> ```

Typo
___
*[`src/BloomExe/WebLibraryIntegration/BloomS3Client.cs` line 210 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5610#-NOMFElb3-jHwUKETioY:-NOMFElcDnZgObpssvAg:b-goqyw3) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/fb66b947d4a694452d328409872d09b330a92969/src/BloomExe/WebLibraryIntegration/BloomS3Client.cs#L210)):*
> ```Smalltalk
> 		/// except the one we want.
> 		/// </summary>
> 		public void UploadBook(string storageKeyOfBookFolder, string pathToBloomBookDirectory, IProgress progress,
> ```

Probably it's just the "Typescript programmer"  in me talking, but 9 parameters seems really fragile and hard to read? At least in TS, we'd be better off defining an options class, e.g.

```
includeNarrationAudio: bool,
includeMusic: bool,
languagesToInclude,
```

Do you think that would be good? Would such a thing be useable in more than just this function?

Alternatively (I think I like this better), would it be clearer if this method were to take a Filter as a parameter, rather than creating one?

hatton (changes requested):
> Reviewed 1 of 21 files at r1.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5610)* status: 18 of 21 files reviewed, 19 unresolved discussions (waiting on @JohnThomson)

___
*[`src/BloomExe/Publish/Epub/EpubMaker.cs` line 287 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5610#-NOMHzca-3zu5qBzXet0:-NOMHzca-3zu5qBzXet1:b-m8634y) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/fb66b947d4a694452d328409872d09b330a92969/src/BloomExe/Publish/Epub/EpubMaker.cs#L287)):*
> ```Smalltalk
> 				_book = PublishHelper.MakeDeviceXmatterTempBook(_book.FolderPath, _bookServer, tempBookPath,
> 					_book.IsTemplateBook, _omittedPageLabels,
> 					fullCapabilities: false, // no activities in Epubs.
> ```

This seems vague/fragile? It appears to only be used by ePUBs. Should this instead be saying what it doesn't want, explicitly?

hatton (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5610)* status: 18 of 21 files reviewed, 20 unresolved discussions (waiting on @JohnThomson)

___
*[`src/BloomExe/Publish/Android/BloomPubMaker.cs` line 34 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5610#-NOMIXJnCYA6KGs0Ppr7:-NOMIXJnCYA6KGs0Ppr8:b4ctka1) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/fb66b947d4a694452d328409872d09b330a92969/src/BloomExe/Publish/Android/BloomPubMaker.cs#L34)):*
> ```Smalltalk
> 		internal const string kCreatorBloom = "bloom";
> 		internal const string kCreatorHarvester = "harvester";
> 		public static string LastVersionCode { get; private set; }
> ```

While we're in here, let's improve this name. Is it only for WifiPublishing? If so, how about `HashOfMostRecentlyCreatedBook` or something like that?

JohnThomson (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5610)* status: 6 of 23 files reviewed, 19 unresolved discussions (waiting on @hatton)

___
*[`src/BloomExe/Book/BookCompressor.cs` line 18 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5610#-NOMGY9P9XiwLT_iu0FA:-NORT8l19eM1e1qNq0E8:b55omw) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/fb66b947d4a694452d328409872d09b330a92969/src/BloomExe/Book/BookCompressor.cs#L18)):*

Seems odd that this class relies on a UI component?
</blockquote></details>

It doesn't. CollectionTab is a namespace; it uses a static method from CollectionModel, which is in that namespace.
___
*[`src/BloomExe/Book/BookCompressor.cs` line 26 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5610#-NOMH-C3ACJ2KgUjuDqO:-NORUjoW5n1Jx4FMT2ry:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/fb66b947d4a694452d328409872d09b330a92969/src/BloomExe/Book/BookCompressor.cs#L26)):*

Is there a more natural place for this bit of info?
</blockquote></details>

Done.
___
*[`src/BloomExe/Book/BookCompressor.cs` line 30 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5610#-NOMGotY6OM3D4uzoBfa:-NORVRSBEdKjKSEKqy_x:b-gkzkxq) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/fb66b947d4a694452d328409872d09b330a92969/src/BloomExe/Book/BookCompressor.cs#L30)):*

Is there a more natural class to hang this off of? 
</blockquote></details>

It's functioning as list of extensions for kinds of images we may want to shrink. I think this is a good place for it once I renamed it.
___
*[`src/BloomExe/Book/BookCompressor.cs` line 81 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5610#-NOMHA7v3oI71GLtQWy2:-NOW4sKSEpPlfQ42lAAt:bg51q4w) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/fb66b947d4a694452d328409872d09b330a92969/src/BloomExe/Book/BookCompressor.cs#L81)):*

This looks like it returns something that we ignore. Clearer would be `AddBookOverrides(....)`
</blockquote></details>

I did something a little different but I think it achieves the goal.
___
*[`src/BloomExe/Book/BookCompressor.cs` line 121 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5610#-NOMHNdPC_q5mvikLtNy:-NOW4xU_6tDJhJiu4zDo:bg9om4f) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/fb66b947d4a694452d328409872d09b330a92969/src/BloomExe/Book/BookCompressor.cs#L121)):*

It's not clear what an "override" is. Would `AddToFileToAlwaysAccept()` be accurate?
</blockquote></details>

Commented and improved name.
___
*[`src/BloomExe/Book/BookFileFilter.cs` line 21 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5610#-NOM84-d59anCdH1ReUs:-NOW5EvSDr4GRxQi_Y3k:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/fb66b947d4a694452d328409872d09b330a92969/src/BloomExe/Book/BookFileFilter.cs#L21)):*

Should this be "determining which **files** should"?
</blockquote></details>

Done.
___
*[`src/BloomExe/Book/BookFileFilter.cs` line 52 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5610#-NOM989e6HvVJ1fMoLsW:-NOW6lpL0PkdNoA53ZsL:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/fb66b947d4a694452d328409872d09b330a92969/src/BloomExe/Book/BookFileFilter.cs#L52)):*

Can add a comment? It's not immediately clear how "ForInteractive" relates to editing. I assume it is that when editing, we need *everything*?
</blockquote></details>

Done.
___
*[`src/BloomExe/Book/BookFileFilter.cs` line 53 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5610#-NOM8zjO73px-kN-7AuX:-NOW73YF3l-Xh044XIek:b-4iduyw) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/fb66b947d4a694452d328409872d09b330a92969/src/BloomExe/Book/BookFileFilter.cs#L53)):*

Yes, that's all it ever got used for.
</blockquote></details>

Cleaned up
___
*[`src/BloomExe/Book/BookFileFilter.cs` line 66 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5610#-NOM9SI67f1GovEXh16S:-NOW7_EK1pNzM7Hqay6d:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/fb66b947d4a694452d328409872d09b330a92969/src/BloomExe/Book/BookFileFilter.cs#L66)):*

But also for editing? (see comment above)
</blockquote></details>

Done.
___
*[`src/BloomExe/Book/BookFileFilter.cs` line 108 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5610#-NOMA1Og1bk76eufEBTb:-NOW8Ons7CJ9TmhDPu3s:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/fb66b947d4a694452d328409872d09b330a92969/src/BloomExe/Book/BookFileFilter.cs#L108)):*

Two functions, `AlwaysAccept()` and `AlwaysReject()` would be worth it for clarity.
</blockquote></details>

Done.
___
*[`src/BloomExe/Book/BookFileFilter.cs` line 111 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5610#-NOMAUMIEnM96bcXLh0v:-NOW9GYjE2T92kkQHs33:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/fb66b947d4a694452d328409872d09b330a92969/src/BloomExe/Book/BookFileFilter.cs#L111)):*

The 5 instances of this could become `path = normalizePath(path)` with the platform check then encapsulated in that.
</blockquote></details>

Done.
___
*[`src/BloomExe/Book/BookFileFilter.cs` line 147 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5610#-NOMBs6BBJouD9fAjCxU:-NOW9NoBAxLGk15tIm1i:b-q9slea) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/fb66b947d4a694452d328409872d09b330a92969/src/BloomExe/Book/BookFileFilter.cs#L147)):*

If this fails (e.g. due to Cloud interference), does the user get a good message?
</blockquote></details>

I would expect about as good a message as they would have gotten before when trying to copy the file later in the process.
___
*[`src/BloomExe/Book/BookFileFilter.cs` line 326 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5610#-NOMCNo439_HT1RRzNtD:-NOWA8fp8RX0lLXQsQOs:bn1usf7) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/fb66b947d4a694452d328409872d09b330a92969/src/BloomExe/Book/BookFileFilter.cs#L326)):*

If this fails for any reason, does the user get a good message?
</blockquote></details>

It's always called inside the same try...catch context as the code that actually copies the files, so the error handling should be the same as before this change.
___
*[`src/BloomExe/CollectionTab/CollectionModel.cs` line 717 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5610#-NOMDem7DJoufVFkpfdm:-NOWBZ6h9VB0Epid77Zc:bsn24ua) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/fb66b947d4a694452d328409872d09b330a92969/src/BloomExe/CollectionTab/CollectionModel.cs#L717)):*

Searching my drive for placeHolder*.png suggests we need a regex here :-)  Then you can use that on `thumbnail-` as well.



</blockquote></details>

I'm not sure we want to exclude all thumbnails...there might be another size we want. Though I guess the receiving Bloom could re-create any thumbnail it needs? I don't want to do this, but I'm not sure exactly why. Maybe a combination of
- not worth the effort to exclude maybe one more thumbnail?
- excluding individual files is more efficient: one dictionary lookup for each file accounts for all of them, vs each regex must be tested against each file. (Probably not a significant optimization, though.)
- the code I started from excluded these particular sizes and I don't know why it excludes those and only those, but wonder if there was some good reason.
I can do it if you want.

And while we can exclude placeHolder.png since the Bloom using the BloomPack will find its own version, I don't think Bloom is currently smart enough to substitute its own placeHolder.png for any file whose name starts with placeHolder that might be referred to in the HTML. (Images that are not referred to in the HTML will get rejected anyway.)
___
*[`src/BloomExe/Publish/Android/AndroidPublishSettings.cs` line 27 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5610#-NOMHfnPBfEsLhJOEYjf:-NOWEB981-a12rQo1xtH:bb7f2mf) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/fb66b947d4a694452d328409872d09b330a92969/src/BloomExe/Publish/Android/AndroidPublishSettings.cs#L27)):*

Typo
</blockquote></details>

done
___
*[`src/BloomExe/Publish/Android/BloomPubMaker.cs` line 34 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5610#-NOMIXJnCYA6KGs0Ppr7:-NOWEXb6Emd-_samrmUq:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/fb66b947d4a694452d328409872d09b330a92969/src/BloomExe/Publish/Android/BloomPubMaker.cs#L34)):*

While we're in here, let's improve this name. Is it only for WifiPublishing? If so, how about `HashOfMostRecentlyCreatedBook` or something like that?
</blockquote></details>

Done.
___
*[`src/BloomExe/Publish/Epub/EpubMaker.cs` line 287 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5610#-NOMHzca-3zu5qBzXet0:-NOWFE888-T6OAaY7MTg:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/fb66b947d4a694452d328409872d09b330a92969/src/BloomExe/Publish/Epub/EpubMaker.cs#L287)):*

This seems vague/fragile? It appears to only be used by ePUBs. Should this instead be saying what it doesn't want, explicitly?
</blockquote></details>

Done.
___
*[`src/BloomExe/WebLibraryIntegration/BloomS3Client.cs` line 210 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5610#-NOMFElb3-jHwUKETioY:-NOWHdV3C2FGmEUFzF22:bb938qm) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/fb66b947d4a694452d328409872d09b330a92969/src/BloomExe/WebLibraryIntegration/BloomS3Client.cs#L210)):*

Probably it's just the "Typescript programmer"  in me talking, but 9 parameters seems really fragile and hard to read? At least in TS, we'd be better off defining an options class, e.g.

```
includeNarrationAudio: bool,
includeMusic: bool,
languagesToInclude,
```

Do you think that would be good? Would such a thing be useable in more than just this function?

Alternatively (I think I like this better), would it be clearer if this method were to take a Filter as a parameter, rather than creating one?
</blockquote></details>

Bit of a toss-up in my mind. That pattern is less necessary in C# than in Javascript because of C#'s support for overloads and for passing parameters by name, which allows leaving out any that have defaults that don't need overriding. It is sometimes useful when a large set of params needs to be passed through several layers of method that don't actually use them. That's not the case here. 
I tried to see what the effect would be and only got rid of three parameters (while adding the filter)., so the benefit is not all that great. There is only one caller apart from test code (and not likely to be others) so we would not be greatly reducing the chance of confusion.
Actually I think a better change would be to move all the code that filters and manipulates the files to be uploaded into the caller in BookUpload. That's the right class to know what should be uploaded; BloomS3Client should only be about doing the upload. Then we'd have a lot fewer parameters.
But it's a messy refactor (there are tests of the current version of BloomS3Client) and that class really only exists to support BookUpload. I don't think it's worth it, even though I know you are (and I try to be) keen to do anything that improves the organization of our code. Some refactorings are less valuable and more costly than others, and I think this one has one of the poorer cost-benefit ratios. But I can if you want.
___
*[`src/BloomExe/Collection/CollectionFilter.cs` line 16 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5610#-NOMCkrk8g6OoFq0bUB8:-NOWAhWQBE33Mfx63q9t:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/fb66b947d4a694452d328409872d09b330a92969/src/BloomExe/Collection/CollectionFilter.cs#L16)):*

I keep reading this to mean it filters collections. How about `CollectionFileFilter`, in parallel with your `BookFileFilter`?
</blockquote></details>

Done.
___
*[`src/BloomExe/Collection/CollectionFilter.cs` line 46 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5610#-NOMD4WIDb3uWHXm5Hba:-NOWBMiUB0-qoC6B6U70:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/fb66b947d4a694452d328409872d09b330a92969/src/BloomExe/Collection/CollectionFilter.cs#L46)):*

It would be nice if we had a version of Assert that will through on a tester. Absent that, you might as well throw and Exception.
</blockquote></details>

Done.

hatton (changes requested):
> Reviewed 17 of 17 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5610)* status: all files reviewed, 5 unresolved discussions (waiting on @JohnThomson)

___
*[`src/BloomExe/Book/BookFileFilter.cs` line 52 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5610#-NOM989e6HvVJ1fMoLsW:-NOkZgV47u-uz1nmg6oe:bh8cbn9) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/fb66b947d4a694452d328409872d09b330a92969/src/BloomExe/Book/BookFileFilter.cs#L52)):*

Done.
</blockquote></details>

Hmmm... I shouldn't have asked for a comment. I always end up wanting a more transparent name :-)

What if it was named "IncludeAllThingsNeededByBloomPlayer"? This conveys that it doesn't exclude anything.
___
*[`src/BloomExe/Book/BookFileFilter.cs` line 147 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5610#-NOMBs6BBJouD9fAjCxU:-NOkan-g297pclGFkkhV:byy8k3j) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/fb66b947d4a694452d328409872d09b330a92969/src/BloomExe/Book/BookFileFilter.cs#L147)):*

I would expect about as good a message as they would have gotten before when trying to copy the file later in the process.
</blockquote></details>

Is that a Yes?
___
*[`src/BloomExe/Book/BookFileFilter.cs` line 326 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5610#-NOMCNo439_HT1RRzNtD:-NOkarfG8y2XtXtthwac:b-41hka6) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/fb66b947d4a694452d328409872d09b330a92969/src/BloomExe/Book/BookFileFilter.cs#L326)):*

It's always called inside the same try...catch context as the code that actually copies the files, so the error handling should be the same as before this change.
</blockquote></details>

Did the error handling before give the user a good message?
___
*[`src/BloomExe/CollectionTab/CollectionModel.cs` line 717 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5610#-NOMDem7DJoufVFkpfdm:-NOkbLPz0ERj6ZPbvAYa:bx3gjye) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/fb66b947d4a694452d328409872d09b330a92969/src/BloomExe/CollectionTab/CollectionModel.cs#L717)):*

I'm not sure we want to exclude all thumbnails...there might be another size we want. Though I guess the receiving Bloom could re-create any thumbnail it needs? I don't want to do this, but I'm not sure exactly why. Maybe a combination of
- not worth the effort to exclude maybe one more thumbnail?
- excluding individual files is more efficient: one dictionary lookup for each file accounts for all of them, vs each regex must be tested against each file. (Probably not a significant optimization, though.)
- the code I started from excluded these particular sizes and I don't know why it excludes those and only those, but wonder if there was some good reason.
I can do it if you want.

And while we can exclude placeHolder.png since the Bloom using the BloomPack will find its own version, I don't think Bloom is currently smart enough to substitute its own placeHolder.png for any file whose name starts with placeHolder that might be referred to in the HTML. (Images that are not referred to in the HTML will get rejected anyway.)
</blockquote></details>

Yes, let's remove all thumbnail-.*. If you're concerned that the user will have a file named /placeHolder.+/ (with exactly that case, exactly that no-space phrase), don't be.
___
*[`src/BloomExe/WebLibraryIntegration/BloomS3Client.cs` line 210 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5610#-NOMFElb3-jHwUKETioY:-NOkcQnV0lv-CxJMj2ee:b4z4qxn) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/fb66b947d4a694452d328409872d09b330a92969/src/BloomExe/WebLibraryIntegration/BloomS3Client.cs#L210)):*

Bit of a toss-up in my mind. That pattern is less necessary in C# than in Javascript because of C#'s support for overloads and for passing parameters by name, which allows leaving out any that have defaults that don't need overriding. It is sometimes useful when a large set of params needs to be passed through several layers of method that don't actually use them. That's not the case here. 
I tried to see what the effect would be and only got rid of three parameters (while adding the filter)., so the benefit is not all that great. There is only one caller apart from test code (and not likely to be others) so we would not be greatly reducing the chance of confusion.
Actually I think a better change would be to move all the code that filters and manipulates the files to be uploaded into the caller in BookUpload. That's the right class to know what should be uploaded; BloomS3Client should only be about doing the upload. Then we'd have a lot fewer parameters.
But it's a messy refactor (there are tests of the current version of BloomS3Client) and that class really only exists to support BookUpload. I don't think it's worth it, even though I know you are (and I try to be) keen to do anything that improves the organization of our code. Some refactorings are less valuable and more costly than others, and I think this one has one of the poorer cost-benefit ratios. But I can if you want.
</blockquote></details>

If you tried having the client create and pass the filter and it blew up into a mess, OK. Else, please give it a go.

JohnThomson (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5610)* status: 17 of 23 files reviewed, 5 unresolved discussions (waiting on @hatton)

___
*[`src/BloomExe/Book/BookFileFilter.cs` line 52 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5610#-NOM989e6HvVJ1fMoLsW:-NOkvkau4whsSWM1WsVB:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/fb66b947d4a694452d328409872d09b330a92969/src/BloomExe/Book/BookFileFilter.cs#L52)):*

Hmmm... I shouldn't have asked for a comment. I always end up wanting a more transparent name :-)

What if it was named "IncludeAllThingsNeededByBloomPlayer"? This conveys that it doesn't exclude anything.
</blockquote></details>

Done.
___
*[`src/BloomExe/Book/BookFileFilter.cs` line 147 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5610#-NOMBs6BBJouD9fAjCxU:-NOkmNLp7a0bUUBAumlv:b-fqrmii) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/fb66b947d4a694452d328409872d09b330a92969/src/BloomExe/Book/BookFileFilter.cs#L147)):*

Is that a Yes?
</blockquote></details>

More a combination of "I'm not sure" and "it could vary based on where the filter is used". So I looked around some more. Making a BloomPack might be a case where before there was no reason to actually read the HTML file, so this change might introduce a new possibility of a poorly-reported error if the file is corrupt. (If it can't be read at all because of cloud interference, it looks like the exception will propagate all the way 
___
*[`src/BloomExe/Book/BookFileFilter.cs` line 326 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5610#-NOMCNo439_HT1RRzNtD:-NOl5GIqAe_nuv18PMtN:b-8fb2wx) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/fb66b947d4a694452d328409872d09b330a92969/src/BloomExe/Book/BookFileFilter.cs#L326)):*

Did the error handling before give the user a good message?
</blockquote></details>

After about an hour's work I managed to trace one call chain back to a definitive answer: No. BulkBloomPubCreator.PublishAllBooks appears to log exceptions but not report them to the user at all.
There are about 24 call chains I have explored partly but haven't yet finished tracing back far enough to answer. So this will be a BIG project.
___
*[`src/BloomExe/CollectionTab/CollectionModel.cs` line 717 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5610#-NOMDem7DJoufVFkpfdm:-NOkqAIiALEbPSrWiudV:b-7aa3vl) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/fb66b947d4a694452d328409872d09b330a92969/src/BloomExe/CollectionTab/CollectionModel.cs#L717)):*

Yes, let's remove all thumbnail-.*. If you're concerned that the user will have a file named /placeHolder.+/ (with exactly that case, exactly that no-space phrase), don't be.
</blockquote></details>

My thought is that your folder probably has placeHolder122.png because some version of Bloom at some point decided it needed a new name and there is some img element in that book that has placeHolder122.png as its src attribute. (If that wasn't true, I think Bloom would already have deleted it.) So I'm guessing that if we leave it out, someone looking at that page will see a missing image symbol. But it's possible that we're already doing something that will prevent that.
Of course a missing image symbol is maybe no worse than seeing the placeholder. We don't want to see that in publications anyway.
OK, I'll make it remove thumbnails. I suppose I can make it remove placeholders too, and we can see if it causes a problem.
___
*[`src/BloomExe/WebLibraryIntegration/BloomS3Client.cs` line 210 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5610#-NOMFElb3-jHwUKETioY:-NOkvbAI3aAbZN2Hm5b-:bs68ytw) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/fb66b947d4a694452d328409872d09b330a92969/src/BloomExe/WebLibraryIntegration/BloomS3Client.cs#L210)):*

If you tried having the client create and pass the filter and it blew up into a mess, OK. Else, please give it a go.
</blockquote></details>

That's what I was trying to convey. I tried it and it was a mess.

JohnThomson (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5610)* status: 17 of 25 files reviewed, 5 unresolved discussions (waiting on @hatton)

___
*[`src/BloomExe/Book/BookFileFilter.cs` line 326 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5610#-NOMCNo439_HT1RRzNtD:-NOpDKaF3Bs842-yK1ed:b-yq442y) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/fb66b947d4a694452d328409872d09b330a92969/src/BloomExe/Book/BookFileFilter.cs#L326)):*

After about an hour's work I managed to trace one call chain back to a definitive answer: No. BulkBloomPubCreator.PublishAllBooks appears to log exceptions but not report them to the user at all.
There are about 24 call chains I have explored partly but haven't yet finished tracing back far enough to answer. So this will be a BIG project.
</blockquote></details>

I did some real testing with FileMeddler and reported the results in the issue. There's room for improvement, but most of them are reasonable. https://github.com/sillsdev/libpalaso/pull/1246 should help with the worst and most frequent problem (both to stop it happening and improve the reporting).

JohnThomson (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5610)* status: 17 of 25 files reviewed, 5 unresolved discussions (waiting on @hatton)

___
*[`src/BloomExe/Book/BookFileFilter.cs` line 147 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5610#-NOMBs6BBJouD9fAjCxU:-NOpE90-99Z3lmdJjuA7:bb2mncu) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/fb66b947d4a694452d328409872d09b330a92969/src/BloomExe/Book/BookFileFilter.cs#L147)):*

More a combination of "I'm not sure" and "it could vary based on where the filter is used". So I looked around some more. Making a BloomPack might be a case where before there was no reason to actually read the HTML file, so this change might introduce a new possibility of a poorly-reported error if the file is corrupt. (If it can't be read at all because of cloud interference, it looks like the exception will propagate all the way 
</blockquote></details>

Sorry, that comment wasn't finished. Looks as if it will propagate all the way to the default API error handling. I did try making a BloomPack using FileMeddler, but it didn't cause an error.

hatton (changes requested):
> Reviewed 8 of 10 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5610)* status: 25 of 27 files reviewed, 1 unresolved discussion (waiting on @JohnThomson)

___
*[`src/BloomExe/CollectionTab/CollectionModel.cs` line 717 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5610#-NOMDem7DJoufVFkpfdm:-NOyPiYiEdbidsncAzPy:bls3nh4) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/fb66b947d4a694452d328409872d09b330a92969/src/BloomExe/CollectionTab/CollectionModel.cs#L717)):*

My thought is that your folder probably has placeHolder122.png because some version of Bloom at some point decided it needed a new name and there is some img element in that book that has placeHolder122.png as its src attribute. (If that wasn't true, I think Bloom would already have deleted it.) So I'm guessing that if we leave it out, someone looking at that page will see a missing image symbol. But it's possible that we're already doing something that will prevent that.
Of course a missing image symbol is maybe no worse than seeing the placeholder. We don't want to see that in publications anyway.
OK, I'll make it remove thumbnails. I suppose I can make it remove placeholders too, and we can see if it causes a problem.
</blockquote></details>

I see that `BookFileFilter` got regexes for thumbnails and placeholders. I'm not clear how that relates to these two narrow instanced of `filter.AlwaysReject()`? Maybe worth a comment.

JohnThomson (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5610)* status: 25 of 27 files reviewed, 1 unresolved discussion (waiting on @hatton)

___
*[`src/BloomExe/CollectionTab/CollectionModel.cs` line 717 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5610#-NOMDem7DJoufVFkpfdm:-NP3-GVK5ZflRFQaZwLq:b-mjyksi) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/fb66b947d4a694452d328409872d09b330a92969/src/BloomExe/CollectionTab/CollectionModel.cs#L717)):*

I see that `BookFileFilter` got regexes for thumbnails and placeholders. I'm not clear how that relates to these two narrow instanced of `filter.AlwaysReject()`? Maybe worth a comment.
</blockquote></details>

No, you caught a real problem. We don't need a duplicate deletion of placeholder, and thumbnails should ONLY be excluded for Bloompacks (here).

hatton (approved):
> Reviewed 1 of 10 files at r3, 3 of 3 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5610)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @JohnThomson)

PR #5611, opened 2023-02-02
Ensure unit tests always use the WebView2 browser (BL-11898)

StephenMcConnel:
Some entanglement with Geckofx remains, specifically with at least tests using the ElementProxy class, so the references cannot yet be removed from the csproj file.

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5611)


JohnThomson (changes requested):
Very good news if this is all we need to do to get the bulk of the tests running with WebView2.
What's the plan for ElementProxy? A quick look indicates that we aren't using GeckoFx-element ElementProxies; there are four methods that call the relevant constructor outside ElementProxy itself, but none of them is used, so they can be removed; and within ElementProxy, we could very well remove all the code that references parentG and end up with a version that only handles XmlElement. At that point the class has little point, so we could either refactor the callers to use XmlElement directly (not a huge job; there are 15 places we use this outside the class and its tests) or just comment that ElementProxy is currently obsolete but retained in case we might one day use yet another browser that could benefit. (But this is very unlikely; we want to be moving away from direct manipulation of the DOM in C#.)
In either case, it should then not be a cause of our tests needing to reference GeckoFx.

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5611#-:-NNFV3Su-3seAEAYiMa8:b-89ctug)* status: all files reviewed, 1 unresolved discussion (waiting on @StephenMcConnel)

___
*[`src/BloomTests/NavigationIsolatorTests.cs` line 1 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5611#-NNFUV3k0xSFNPmie1Nz:-NNFUV3l3wLvSwsmMiS5:b-qdpzvi) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/6ff5531fbd99e1d70adf4b1a8216a70c610deb1f/src/BloomTests/NavigationIsolatorTests.cs#L1)):*
> ```Smalltalk
> using System;
> ```

Do you have reason to be sure we don't need NavigationIsolator for WebView2? Or at least that we aren't using it? If it's still a part of our code base, why are we no longer testing it?

StephenMcConnel (commented):
Done.  ElementProxy still exists, but I added a comment that we may want to remove it as part of the Geckofx cleanup.

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5611#-:-NNI0Ujp8tg6PHyiHWtX:b-9bt8pr)* status: 7 of 11 files reviewed, 1 unresolved discussion (waiting on @JohnThomson)

___
*[`src/BloomTests/NavigationIsolatorTests.cs` line 1 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5611#-NNFUV3k0xSFNPmie1Nz:-NNHvcO13moQarmuDIow:bxwi97h) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/6ff5531fbd99e1d70adf4b1a8216a70c610deb1f/src/BloomTests/NavigationIsolatorTests.cs#L1)):*

Do you have reason to be sure we don't need NavigationIsolator for WebView2? Or at least that we aren't using it? If it's still a part of our code base, why are we no longer testing it?
</blockquote></details>

NavigationIsolator is a Geckofx workaround.  It goes away at the same time as all of Geckofx.  But if we want tests to use only WebView2, these tests are unnecessary.

JohnThomson (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5611)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @StephenMcConnel)

PR #5612, opened 2023-02-02
BL-11539 AddPage dialog -> React

gmartin7:
* get basic React dialog to appear
* get template urls loaded into React
* moved CSS from .less file to emotion
* get Storybook working
* upgrade files to latest emotion reference
* get stories working reliably
* dialog loads multiple groups
* fix title swapping by recreating dialog container
   each time
* filter out custom page if for change layout
* remove unused property on c# side
* pull out thumb image source function to page group
* get custom template images working
* add blue overlay
* selecting a page highlights it, activates the preview, and
   shows the appropriate image
* adjustments to size/layout of dialog
* fixed colors of title bar and dialog background
* move utility counting functions up to PageChooserDialog
* simplify open/close and submit button handling
* fix Learn More links to open external browser
* added enterprise-only page markers
* tweaked spacing and font-size
* fix initial double-selecting of 1st and default
* delete old files
* add error handling/reporting in case of bad template
   file or no template pages in file
* get double-click handler working

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5612)


gmartin7 (commented):
I added a 2nd commit that limits the dialog dragging to the main edit pane. Otherwise it can be dragged under the page list on the left or just clear off the screen! It's a separate commit in case it's not wanted.

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5612#-:-NNJHLUNBW_5mareZ_Q6:b-4hetzq)* status: 0 of 28 files reviewed, all discussions resolved

gmartin7 (commented):
OK, I just repushed the PR with a new 2nd commit for upgrading to MUI 5. The old 2nd commit (see above) is now the 3rd commit. :)

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5612#-:-NNNGE_CDqgq5SPd2NHA:b-yhbe4q)* status: 0 of 28 files reviewed, all discussions resolved

JohnThomson (changes requested):
I've made a lot of comments, but it's a big piece of work! Thanks for taking it on. Don't regard my suggestions as gospel truth...I may well have missed a good reason for things to be the way they are, so feel free to push back. Even if a suggestion is doable, it may be more work than it is worth.

> Reviewed 21 of 28 files at r1, 9 of 9 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5612#-:-NNJYcUvDwpQ7vWafgGf:b-m9jkps)* status: all files reviewed, 25 unresolved discussions (waiting on @gmartin7)

___
*[`src/BloomBrowserUI/pageChooser/ChooserPageGroup.tsx` line 23 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5612#-NNJQmluBCFC2LHlz1JN:-NNJQmlvA_bkQrmm2VAj:b4ckeio) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/670b681e0763ba40b9ebf4d4147bde14f0be4937/src/BloomBrowserUI/pageChooser/ChooserPageGroup.tsx#L23)):*
> ```XML
>     orientation: string;
>     forChooseLayout: boolean;
>     onThumbSelect?: React.MouseEventHandler<HTMLDivElement>;
> ```

Thumbnail? or Page? Thumb by itself suggests the draggable part of a scroll bar. (Coming back to this: later I suggest that you want a single consistent name, probably TemplatePage, for the things the use can choose in the dialog.)
___
*[`src/BloomBrowserUI/pageChooser/ChooserPageGroup.tsx` line 74 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5612#-NNJSF6dBmYz1UC_7sRF:-NNJSF6dBmYz1UC_7sRG:b6fass6) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/670b681e0763ba40b9ebf4d4147bde14f0be4937/src/BloomBrowserUI/pageChooser/ChooserPageGroup.tsx#L74)):*
> ```XML
>                     result.data,
>                     "text/html"
>                 ).body;
> ```

We usually use JSON to pass complex data across Bloom's API. Why is this passed as HTML? Deserves a comment, at least.
___
*[`src/BloomBrowserUI/pageChooser/ChooserPageGroup.tsx` line 89 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5612#-NNJT4gYBMhLbA_AjTmK:-NNJT4gYBMhLbA_AjTmL:b-ce83qs) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/670b681e0763ba40b9ebf4d4147bde14f0be4937/src/BloomBrowserUI/pageChooser/ChooserPageGroup.tsx#L89)):*
> ```XML
>                 // Don't add a group for books that don't have template pages; just move on.
>                 // (This will always be true for a newly created template.)
>                 if (originalPages.length === 0) {
> ```

Why are we testing originalPages, rather than the filtered list we are actually going to show? Do we really want a heading with no pages if none of them is 'extra'?
(For that matter: do we use the non-extra pages for anything? If not, why not filter them out on the sending side of the API? In fact, we could make the sending side responsible to not send books that don't have at least one addable page at all.
___
*[`src/BloomBrowserUI/pageChooser/ChooserPageGroup.tsx` line 111 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5612#-NNJSfApBi6_WCqTdcZD:-NNJSfApBi6_WCqTdcZE:bceoqpm) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/670b681e0763ba40b9ebf4d4147bde14f0be4937/src/BloomBrowserUI/pageChooser/ChooserPageGroup.tsx#L111)):*
> ```XML
>                         (elem: HTMLElement) =>
>                             elem.id &&
>                             elem.getAttribute("data-page") === "extra"
> ```

Might deserve a comment? Why are the pages the user can add tagged "extra"?
___
*[`src/BloomBrowserUI/pageChooser/ChooserPageGroup.tsx` line 158 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5612#-NNJUIqdDIJcjRWKaww8:-NNJUIqdDIJcjRWKaww9:b-ly6gad) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/670b681e0763ba40b9ebf4d4147bde14f0be4937/src/BloomBrowserUI/pageChooser/ChooserPageGroup.tsx#L158)):*
> ```XML
>         ? pageData.map((currentPageDiv: HTMLElement, index) => {
>               // Process each page and create a gridItem div containing a thumb and data and an overlay.
>               if (currentPageDiv.getAttribute("data-page") === "singleton")
> ```

Didn't we already make sure they all have data-page="extra"?
___
*[`src/BloomBrowserUI/pageChooser/ChooserPageGroup.tsx` line 195 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5612#-NNJUlyyBj3sDJXyIP0S:-NNJUlyyBj3sDJXyIP0T:b-44vdv4) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/670b681e0763ba40b9ebf4d4147bde14f0be4937/src/BloomBrowserUI/pageChooser/ChooserPageGroup.tsx#L195)):*
> ```XML
>               const enterpriseOnlyRules = isEnterprise
>                   ? `:after {
>                         content: "\u25cf";
> ```

Probably deserves to be a constant with a name and possibly comment somewhere.
___
*[`src/BloomBrowserUI/pageChooser/ChooserPageGroup.tsx` line 227 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5612#-NNJWKLME2Cs8bubpuRg:-NNJWKLME2Cs8bubpuRh:b-q2fbdd) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/670b681e0763ba40b9ebf4d4147bde14f0be4937/src/BloomBrowserUI/pageChooser/ChooserPageGroup.tsx#L227)):*
> ```XML
>                           width: 104px;
>                           display: inline-block;
>                           position: sticky; //this makes the blue overlay, which is a child, be correctly positioned even when scrolled
> ```

I'd be very surprised if 'sticky' is what you want. I think you just need position: relative to make child positions be relative to this rather than an ancestor. Sticky does very little without specifying one of the position values (top, left, etc.) beyond 'relative' except perhaps to make its own stacking context. If that's what you want, try isolation:isolate. But try just position:relative first.
___
*[`src/BloomBrowserUI/pageChooser/ChooserPageGroup.tsx` line 222 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5612#-NNNVj6kB3yxS3ZLI1q-:-NNNVj6kB3yxS3ZLI1q0:b-wy6s93) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/bc4b34c0892601bcdc0adb85c11fbb5064b98947/src/BloomBrowserUI/pageChooser/ChooserPageGroup.tsx#L222)):*
> ```XML
>                       data-digital-only={isDigitalOnly(currentPageDiv)}
>                       data-tool-id={tool}
>                       data-help-link={helpLink}
> ```

In line with my later suggestion about how to handle selecting, consider whether it would work to pass this information as props to the element that represents a TemplatePage.
___
*[`src/BloomBrowserUI/pageChooser/ErrorGroup.tsx` line 12 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5612#-NNNSdE--AUkCzKoFk7O:-NNNSdE--AUkCzKoFk7P:b-3joiis) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/bc4b34c0892601bcdc0adb85c11fbb5064b98947/src/BloomBrowserUI/pageChooser/ErrorGroup.tsx#L12)):*
> ```XML
> }
> 
> export const ErrorGroup: React.FunctionComponent<IErrorGroupProps> = props => {
> ```

More specific name? And/Or class comment? I gather this class is used in place of showing a template book's pages when something is wrong?
___
*[`src/BloomBrowserUI/pageChooser/PageChooserDialog.tsx` line 61 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5612#-NNJZNe27d8UFi6F386P:-NNJZNe3-GufCwhCDgQv:bpzlkbh) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/670b681e0763ba40b9ebf4d4147bde14f0be4937/src/BloomBrowserUI/pageChooser/PageChooserDialog.tsx#L61)):*
> ```XML
>         string | undefined
>     >(undefined);
>     const [enterpriseAvailable, setEnterpriseAvailable] = useState(false);
> ```

I think we have a specific function for enterprise, something like useIsEnterpriseEnabled
___
*[`src/BloomBrowserUI/pageChooser/PageChooserDialog.tsx` line 104 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5612#-NNJZuf5AxjxfMHgOWW1:-NNJZuf5AxjxfMHgOWW2:b26xwuu) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/670b681e0763ba40b9ebf4d4147bde14f0be4937/src/BloomBrowserUI/pageChooser/PageChooserDialog.tsx#L104)):*
> ```XML
>             setDefaultPageToSelect(defaultPageId);
>             setTemplateBookUrls(arrayOfBookUrls);
>             setOrientation(initializationObject["orientation"]);
> ```

IIRC, each of these may cause a render. If so, it may be best to set the book URLs last, so that when you actually have content to render, all the other state things are already right.
___
*[`src/BloomBrowserUI/pageChooser/PageChooserDialog.tsx` line 123 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5612#-NNJ_Mz51rd9-1O52bnm:-NNJ_Mz51rd9-1O52bnn:bg3cxzb) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/670b681e0763ba40b9ebf4d4147bde14f0be4937/src/BloomBrowserUI/pageChooser/PageChooserDialog.tsx#L123)):*
> ```XML
>                 "ui-selected"
>             );
>         }
> ```

There might be a good performance reason for the above, but it should be documented. Otherwise, it's not very Reacty. You are changing the DOM directly! The natural React thing to do is to make the presence of absence of this class happen in the render function.
___
*[`src/BloomBrowserUI/pageChooser/PageChooserDialog.tsx` line 130 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5612#-NNJaDU75jpDLV-e6LS8:-NNJaDU82y8IMcAxeHOg:bv6iddn) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/670b681e0763ba40b9ebf4d4147bde14f0be4937/src/BloomBrowserUI/pageChooser/PageChooserDialog.tsx#L130)):*
> ```XML
>         // an item partly scrolled off the bottom. There's no way currently to select an item
>         // that's entirely scrolled off the top, and it doesn't seem worth the complication
>         // to force a partly-visible one at the top to become wholly visible.
> ```

Are you sure? My recollection is that 'nearest' will scroll the minimum distance in either direction to make the thing fully visible (if possible)
___
*[`src/BloomBrowserUI/pageChooser/PageChooserDialog.tsx` line 168 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5612#-NNNTfCsD-T01b5hWr1T:-NNNTfCt7QtTUtiCJ7uL:bvjuxcw) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/bc4b34c0892601bcdc0adb85c11fbb5064b98947/src/BloomBrowserUI/pageChooser/PageChooserDialog.tsx#L168)):*
> ```XML
>             // No default thumb to select. Select the first thumb instead.
>             selectNewGridItem(loadedThumbs[0] as HTMLDivElement);
>         }, 250);
> ```

How do you know 250 will be enough? This problem would go away if the different appearance of the selected item was done by rendering.
___
*[`src/BloomBrowserUI/pageChooser/PageChooserDialog.tsx` line 187 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5612#-NNNUQU682-tpVmHQ7O3:-NNNUQU682-tpVmHQ7O4:bkjtall) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/bc4b34c0892601bcdc0adb85c11fbb5064b98947/src/BloomBrowserUI/pageChooser/PageChooserDialog.tsx#L187)):*
> ```XML
>         // 'clickedDiv' is the selection overlay that sits on top of the thumbnail
>         // Select new thumbnail
>         const newSelection = clickedDiv.parentElement as HTMLDivElement; // What used to be known as .gridItem
> ```

This comment might have been useful during the refactoring, but it's not now. (Which makes me think...gridItem is a very generic name. Why not call it something that brings out what items your grid contains...something like TemplatePage or pageThumbnail?
___
*[`src/BloomBrowserUI/pageChooser/PageChooserDialog.tsx` line 210 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5612#-NNNXE2XAIZ7eQZ8mDqT:-NNNXE2XAIZ7eQZ8mDqU:b-l3mooa) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/bc4b34c0892601bcdc0adb85c11fbb5064b98947/src/BloomBrowserUI/pageChooser/PageChooserDialog.tsx#L210)):*
> ```XML
>         const convertAnywayCheckbox = document.getElementById(
>             "convertAnywayCheckbox"
>         ) as HTMLInputElement;
> ```

Seems like all this code that figures out what you want to do about choosing this page should be common to clicking the button and inside handleAddPageOrChooseLayoutButtonClick
___
*[`src/BloomBrowserUI/pageChooser/PageChooserDialog.tsx` line 214 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5612#-NNNWfEP2oIeflrrUFs4:-NNNWfEP2oIeflrrUFs5:b910ewz) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/bc4b34c0892601bcdc0adb85c11fbb5064b98947/src/BloomBrowserUI/pageChooser/PageChooserDialog.tsx#L214)):*
> ```XML
>             "convertWholeBookCheckbox"
>         ) as HTMLInputElement;
>         const closestGroupAncestor = newSelection.closest(".gridGroup");
> ```

unless "gridGroup" is forced on us by the library, why not give is a meaningful name like 'book' or 'template-book'?
___
*[`src/BloomBrowserUI/pageChooser/PageChooserDialog.tsx` line 232 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5612#-NNNXl1YEU_sjlltsdzC:-NNNXl1YEU_sjlltsdzD:b-1rzdk1) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/bc4b34c0892601bcdc0adb85c11fbb5064b98947/src/BloomBrowserUI/pageChooser/PageChooserDialog.tsx#L232)):*
> ```XML
>     };
> 
>     const getGroupData = (): JSX.Element[] | undefined => {
> ```

Not just getting data, it's getting React elements. Maybe GetPageGroups, but a page group is really a representation of a TemplateBook, so consider renaming it that and calling this GetTempateBooks.
___
*[`src/BloomBrowserUI/pageChooser/PageChooserDialog.tsx` line 272 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5612#-NNNZQ523SGxFizh3Lb_:-NNNZQ523SGxFizh3Lba:bh5undk) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/bc4b34c0892601bcdc0adb85c11fbb5064b98947/src/BloomBrowserUI/pageChooser/PageChooserDialog.tsx#L272)):*
> ```XML
>         );
> 
>         const page = window.parent.document.getElementById(
> ```

I'm not enthusiastic about this rather low-level function having such a lot of high-level knowledge about how our iframes are organized and how pages are structured. Consider whether the four counts you need would make sense as top-level props passed from C# to the PageChooserDialog., or retrieved by it through an API.
___
*[`src/BloomBrowserUI/pageChooser/PageChooserDialog.tsx` line 370 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5612#-NNNdccs2A9XTMyWc_VE:-NNNdccs2A9XTMyWc_VF:b-gpj97e) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/bc4b34c0892601bcdc0adb85c11fbb5064b98947/src/BloomBrowserUI/pageChooser/PageChooserDialog.tsx#L370)):*
> ```XML
>                 padding-bottom: 20px;
>                 width: 875px;
>                 height: 690px;
> ```

The screen sizes we try to support include 800x600 and 1024x586, so this dialog could more than fill the whole screen, and overflow a long way out of the page area. How bad is it when you configure Bloom to shrink the screen size? Can we do anything simple to make it smaller, or responsive? (Maybe that's another card, especially if it proves to be another big project...)
___
*[`src/BloomBrowserUI/pageChooser/PageChooserDialog.tsx` line 390 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5612#-NNNeo4e70miiHTodivg:-NNNeo4e70miiHTodivh:b3doc3s) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/bc4b34c0892601bcdc0adb85c11fbb5064b98947/src/BloomBrowserUI/pageChooser/PageChooserDialog.tsx#L390)):*
> ```XML
>                     border: none;
>                     margin: 0;
>                     height: 600px;
> ```

If the dialog is resizable and the parent div is display;flex, you may be able to make the DialogMiddle the only one that can grow and just give it a min-height. That might help it be more responsive.
___
*[`src/BloomBrowserUI/pageChooser/PageChooserDialog.tsx` line 423 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5612#-NNNfQyaBcG-YXM0a0tv:-NNNfQyaBcG-YXM0a0tw:b-o1glaq) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/bc4b34c0892601bcdc0adb85c11fbb5064b98947/src/BloomBrowserUI/pageChooser/PageChooserDialog.tsx#L423)):*
> ```XML
>                 >
>                     {selectedGridItem && (
>                         <SelectedTemplatePageControls
> ```

Let's try to keep naming as consistent as possible. Are the things you can select pages, templates, TemplatePages, or grid items? Please make a choice and stick to it across all relevant variable and component and class names.
Are you sure we don't want to see any of these controls until something is selected? Of course, you arrange to always select something at startup, so I suppose there's no way to have nothing selected except while it's starting up...so this really just delays showing the controls until we have a selection to show them for. That might be worth commenting.
___
*[`src/BloomBrowserUI/pageChooser/PageChooserDialog.tsx` line 427 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5612#-NNNgHmFDHvRz1jml2Rl:-NNNgHmGC7OXhHsTTii9:b-wb02j6) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/bc4b34c0892601bcdc0adb85c11fbb5064b98947/src/BloomBrowserUI/pageChooser/PageChooserDialog.tsx#L427)):*
> ```XML
>                                 selectedGridItem,
>                                 "data-page-label"
>                             )}
> ```

I wish all these were props or fields in a selectedTemplatePage object or something similar, rather than something we have to go to the DOM for.
___
*[`src/BloomBrowserUI/pageChooser/PageChooserDialog.tsx` line 487 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5612#-NNNhWkM7W8EbCRnDL0T:-NNNhWkM7W8EbCRnDL0U:bo2tbkj) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/bc4b34c0892601bcdc0adb85c11fbb5064b98947/src/BloomBrowserUI/pageChooser/PageChooserDialog.tsx#L487)):*
> ```XML
> function getModalContainer(): HTMLElement {
>     // If our container already exists, remove it and create a new one. Otherwise it maintains the
>     // state of the last time, which may not be what we want (e.g. dialog title).
> ```

That feels wrong. React should render the dialog consistent with the current state., and re-render as the state changes. If it doesn't, you're probably holding state in something React doesn't know about, or some useEffect needs more cases under which it re-runs. (If some data it fetched from C# can become out of date, we have mechanisms for broadcasting and watching for such changes.)
___
*[`src/BloomBrowserUI/react_components/BloomDialog/BloomDialog.tsx` line 294 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5612#-NNNldGk2T5w7BsJfAYi:-NNNldGk2T5w7BsJfAYj:b65dvrx) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/bc4b34c0892601bcdc0adb85c11fbb5064b98947/src/BloomBrowserUI/react_components/BloomDialog/BloomDialog.tsx#L294)):*
> ```XML
> // which returns a component makes a significant difference.
> // When this was a function, typing 3 digits in the hex box would cause
> // a loss of focus which caused it to convert, e.g. FFF to FFFFFF.
> ```

This is not a new comment, but it's not very helpful. DraggablePaper is a function now (props => {...}). Was it at some point a function but not a FunctionComponent? What mistake exactly is this comment helping us not repeat?

gmartin7 (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5612)* status: 20 of 29 files reviewed, 25 unresolved discussions (waiting on @JohnThomson)

___
*[`src/BloomBrowserUI/pageChooser/ChooserPageGroup.tsx` line 23 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5612#-NNJQmluBCFC2LHlz1JN:-NNc7K66-WHuwLB7tQbQ:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/670b681e0763ba40b9ebf4d4147bde14f0be4937/src/BloomBrowserUI/pageChooser/ChooserPageGroup.tsx#L23)):*

Thumbnail? or Page? Thumb by itself suggests the draggable part of a scroll bar. (Coming back to this: later I suggest that you want a single consistent name, probably TemplatePage, for the things the use can choose in the dialog.)
</blockquote></details>

Done.
___
*[`src/BloomBrowserUI/pageChooser/ChooserPageGroup.tsx` line 74 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5612#-NNJSF6dBmYz1UC_7sRF:-NNcELhO3KMp6CV4ra67:bg74b55) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/670b681e0763ba40b9ebf4d4147bde14f0be4937/src/BloomBrowserUI/pageChooser/ChooserPageGroup.tsx#L74)):*

We usually use JSON to pass complex data across Bloom's API. Why is this passed as HTML? Deserves a comment, at least.
</blockquote></details>

Commented. I am quite happy with converting what we did before to good React, but I think changing the basic API goes beyond what I want to do, at least in this PR.
___
*[`src/BloomBrowserUI/pageChooser/ChooserPageGroup.tsx` line 89 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5612#-NNJT4gYBMhLbA_AjTmK:-NNcGLxx9uTL61iHT2ny:b4kfxcj) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/670b681e0763ba40b9ebf4d4147bde14f0be4937/src/BloomBrowserUI/pageChooser/ChooserPageGroup.tsx#L89)):*

Why are we testing originalPages, rather than the filtered list we are actually going to show? Do we really want a heading with no pages if none of them is 'extra'?
(For that matter: do we use the non-extra pages for anything? If not, why not filter them out on the sending side of the API? In fact, we could make the sending side responsible to not send books that don't have at least one addable page at all.
</blockquote></details>

Took me a minute to realize you meant the extra page filter farther down. Good point. I rearranged things a bit. For the API question, it's probably a good idea, but see my previous comment about the API.
___
*[`src/BloomBrowserUI/pageChooser/ChooserPageGroup.tsx` line 111 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5612#-NNJSfApBi6_WCqTdcZD:-NNcIMQ0CYjp_ZgLevUv:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/670b681e0763ba40b9ebf4d4147bde14f0be4937/src/BloomBrowserUI/pageChooser/ChooserPageGroup.tsx#L111)):*

Might deserve a comment? Why are the pages the user can add tagged "extra"?
</blockquote></details>

Done.
___
*[`src/BloomBrowserUI/pageChooser/ChooserPageGroup.tsx` line 158 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5612#-NNJUIqdDIJcjRWKaww8:-NNcIOE09JfvL0fv7BPw:bmnd3zm) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/670b681e0763ba40b9ebf4d4147bde14f0be4937/src/BloomBrowserUI/pageChooser/ChooserPageGroup.tsx#L158)):*

Didn't we already make sure they all have data-page="extra"?
</blockquote></details>

Good point. Removed.
___
*[`src/BloomBrowserUI/pageChooser/ChooserPageGroup.tsx` line 195 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5612#-NNJUlyyBj3sDJXyIP0S:-NNhH8mv4FhlM5r_fSWP:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/670b681e0763ba40b9ebf4d4147bde14f0be4937/src/BloomBrowserUI/pageChooser/ChooserPageGroup.tsx#L195)):*

Probably deserves to be a constant with a name and possibly comment somewhere.
</blockquote></details>

Done.
___
*[`src/BloomBrowserUI/pageChooser/ChooserPageGroup.tsx` line 227 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5612#-NNJWKLME2Cs8bubpuRg:-NNcJPlS5sPt0A8EjI1E:b-vzjfwy) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/670b681e0763ba40b9ebf4d4147bde14f0be4937/src/BloomBrowserUI/pageChooser/ChooserPageGroup.tsx#L227)):*

I'd be very surprised if 'sticky' is what you want. I think you just need position: relative to make child positions be relative to this rather than an ancestor. Sticky does very little without specifying one of the position values (top, left, etc.) beyond 'relative' except perhaps to make its own stacking context. If that's what you want, try isolation:isolate. But try just position:relative first.
</blockquote></details>

You were correct! `position: relative` was sufficient.
___
*[`src/BloomBrowserUI/pageChooser/ChooserPageGroup.tsx` line 222 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5612#-NNNVj6kB3yxS3ZLI1q-:-NNrPHkn12vwtxi5o49h:b-v370w2) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/bc4b34c0892601bcdc0adb85c11fbb5064b98947/src/BloomBrowserUI/pageChooser/ChooserPageGroup.tsx#L222)):*

In line with my later suggestion about how to handle selecting, consider whether it would work to pass this information as props to the element that represents a TemplatePage.
</blockquote></details>

Done. I'm not sure if I actually did what you had in mind, but the whole selection mechanism has been reworked.
___
*[`src/BloomBrowserUI/pageChooser/PageChooserDialog.tsx` line 61 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5612#-NNJZNe27d8UFi6F386P:-NNcLAAVEuBvafSLOghD:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/670b681e0763ba40b9ebf4d4147bde14f0be4937/src/BloomBrowserUI/pageChooser/PageChooserDialog.tsx#L61)):*

I think we have a specific function for enterprise, something like useIsEnterpriseEnabled
</blockquote></details>

Done.
___
*[`src/BloomBrowserUI/pageChooser/PageChooserDialog.tsx` line 104 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5612#-NNJZuf5AxjxfMHgOWW1:-NNcLE8w4lM4PFeoyJdJ:b-51twoq) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/670b681e0763ba40b9ebf4d4147bde14f0be4937/src/BloomBrowserUI/pageChooser/PageChooserDialog.tsx#L104)):*

IIRC, each of these may cause a render. If so, it may be best to set the book URLs last, so that when you actually have content to render, all the other state things are already right.
</blockquote></details>

I commented in the code on what I found in reading online. I did the re-ordering, but multiple state setters in a `useEffect` appear to get batched so there's only one new render. Who knew!?
___
*[`src/BloomBrowserUI/pageChooser/PageChooserDialog.tsx` line 123 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5612#-NNJ_Mz51rd9-1O52bnm:-NNiFJYD1Rkyp1HTwEDN:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/670b681e0763ba40b9ebf4d4147bde14f0be4937/src/BloomBrowserUI/pageChooser/PageChooserDialog.tsx#L123)):*

There might be a good performance reason for the above, but it should be documented. Otherwise, it's not very Reacty. You are changing the DOM directly! The natural React thing to do is to make the presence of absence of this class happen in the render function.
</blockquote></details>

Done.
___
*[`src/BloomBrowserUI/pageChooser/PageChooserDialog.tsx` line 130 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5612#-NNJaDU75jpDLV-e6LS8:-NNcQ4FB5vJCnR5RuZwa:beasd8r) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/670b681e0763ba40b9ebf4d4147bde14f0be4937/src/BloomBrowserUI/pageChooser/PageChooserDialog.tsx#L130)):*

Are you sure? My recollection is that 'nearest' will scroll the minimum distance in either direction to make the thing fully visible (if possible)
</blockquote></details>

I eliminated most of that comment as unnecessary. The old code used `block: "start"` with part of the comment saying we would prefer "nearest", but that it wasn't supported until FF v58 (which no longer applies). It's good that you made me go test it, though, because I had the class in the `.closest()` call wrong! It didn't work at all! Now it works beautifully. [Later: It took awhile to sort it out after refactoring the whole selection mechanism, but the solution involved putting this `scrollIntoView` back in this overall dialog component.]
___
*[`src/BloomBrowserUI/pageChooser/PageChooserDialog.tsx` line 168 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5612#-NNNTfCsD-T01b5hWr1T:-NNiFKnuCsZFLZZ_7FPe:b-21mmu1) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/bc4b34c0892601bcdc0adb85c11fbb5064b98947/src/BloomBrowserUI/pageChooser/PageChooserDialog.tsx#L168)):*

How do you know 250 will be enough? This problem would go away if the different appearance of the selected item was done by rendering.
</blockquote></details>

Done. I'm doing the different selected item method... I too believe this will go away.
___
*[`src/BloomBrowserUI/pageChooser/PageChooserDialog.tsx` line 187 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5612#-NNNUQU682-tpVmHQ7O3:-NNchUhPAQX_uuLG0V4e:b-ms0klo) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/bc4b34c0892601bcdc0adb85c11fbb5064b98947/src/BloomBrowserUI/pageChooser/PageChooserDialog.tsx#L187)):*

This comment might have been useful during the refactoring, but it's not now. (Which makes me think...gridItem is a very generic name. Why not call it something that brings out what items your grid contains...something like TemplatePage or pageThumbnail?
</blockquote></details>

Done. We're actually doing this all differently too.
___
*[`src/BloomBrowserUI/pageChooser/PageChooserDialog.tsx` line 210 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5612#-NNNXE2XAIZ7eQZ8mDqT:-NNrf3vnD6TG-J096ZvF:b3llcxh) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/bc4b34c0892601bcdc0adb85c11fbb5064b98947/src/BloomBrowserUI/pageChooser/PageChooserDialog.tsx#L210)):*

Seems like all this code that figures out what you want to do about choosing this page should be common to clicking the button and inside handleAddPageOrChooseLayoutButtonClick
</blockquote></details>

Yes, it seems so... at first glance.
Unfortunately, the `handleAddPageOrChooseLayoutButtonClick()` method is used exactly 2 places. Once here in the double-click handler in the main dialog component (where it has the selected div, but not the details of checkboxes etc from the Preview component. And once passed into the Preview component where we know all the checkbox details... but don't know the selected div so easily.
___
*[`src/BloomBrowserUI/pageChooser/PageChooserDialog.tsx` line 214 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5612#-NNNWfEP2oIeflrrUFs4:-NNciFy957wkq1IXjvlc:bezsdgu) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/bc4b34c0892601bcdc0adb85c11fbb5064b98947/src/BloomBrowserUI/pageChooser/PageChooserDialog.tsx#L214)):*

unless "gridGroup" is forced on us by the library, why not give is a meaningful name like 'book' or 'template-book'?
</blockquote></details>

Done. Refactoring did away with this.
___
*[`src/BloomBrowserUI/pageChooser/PageChooserDialog.tsx` line 232 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5612#-NNNXl1YEU_sjlltsdzC:-NNciZW03xouoPzIKm3t:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/bc4b34c0892601bcdc0adb85c11fbb5064b98947/src/BloomBrowserUI/pageChooser/PageChooserDialog.tsx#L232)):*

Not just getting data, it's getting React elements. Maybe GetPageGroups, but a page group is really a representation of a TemplateBook, so consider renaming it that and calling this GetTempateBooks.
</blockquote></details>

Done.
___
*[`src/BloomBrowserUI/pageChooser/PageChooserDialog.tsx` line 272 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5612#-NNNZQ523SGxFizh3Lb_:-NNckQje9gbLtvcYuG1e:bmgw52c) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/bc4b34c0892601bcdc0adb85c11fbb5064b98947/src/BloomBrowserUI/pageChooser/PageChooserDialog.tsx#L272)):*

I'm not enthusiastic about this rather low-level function having such a lot of high-level knowledge about how our iframes are organized and how pages are structured. Consider whether the four counts you need would make sense as top-level props passed from C# to the PageChooserDialog., or retrieved by it through an API.
</blockquote></details>

Yeah. I understand your concern. But since we currently launch the dialog via a `RunJavaScript()` command in EditingView... I'd rather make that part of another PR too. This would definitely make sense if we moved more of the template book analysis to C# too, because then the `countX()` functions would move too.
___
*[`src/BloomBrowserUI/pageChooser/PageChooserDialog.tsx` line 370 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5612#-NNNdccs2A9XTMyWc_VE:-NNclMaUD5ZnjCH22Q22:b-4epnwp) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/bc4b34c0892601bcdc0adb85c11fbb5064b98947/src/BloomBrowserUI/pageChooser/PageChooserDialog.tsx#L370)):*

The screen sizes we try to support include 800x600 and 1024x586, so this dialog could more than fill the whole screen, and overflow a long way out of the page area. How bad is it when you configure Bloom to shrink the screen size? Can we do anything simple to make it smaller, or responsive? (Maybe that's another card, especially if it proves to be another big project...)
</blockquote></details>

Another card sounds good to me. :+1:  [Later: I actually did the work for this. The dialog now responds to the size of the edit pane pretty well!]
___
*[`src/BloomBrowserUI/pageChooser/PageChooserDialog.tsx` line 390 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5612#-NNNeo4e70miiHTodivg:-NNcv7Ss9WnFYhzLAnHo:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/bc4b34c0892601bcdc0adb85c11fbb5064b98947/src/BloomBrowserUI/pageChooser/PageChooserDialog.tsx#L390)):*

If the dialog is resizable and the parent div is display;flex, you may be able to make the DialogMiddle the only one that can grow and just give it a min-height. That might help it be more responsive.
</blockquote></details>

Done.
___
*[`src/BloomBrowserUI/pageChooser/PageChooserDialog.tsx` line 423 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5612#-NNNfQyaBcG-YXM0a0tv:-NNcveju-3x3awYDqnpM:b8ep7zs) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/bc4b34c0892601bcdc0adb85c11fbb5064b98947/src/BloomBrowserUI/pageChooser/PageChooserDialog.tsx#L423)):*

Let's try to keep naming as consistent as possible. Are the things you can select pages, templates, TemplatePages, or grid items? Please make a choice and stick to it across all relevant variable and component and class names.
Are you sure we don't want to see any of these controls until something is selected? Of course, you arrange to always select something at startup, so I suppose there's no way to have nothing selected except while it's starting up...so this really just delays showing the controls until we have a selection to show them for. That might be worth commenting.
</blockquote></details>

Done. Commented. `We will always select something. This only delays the preview display until we set a selection.`
I named it `SelectedTemplatePageControls` because it shows the preview and contains several checkboxes, notices, and the main button... i.e. controls relating to the selected template page. Also, this name was already there before this PR.
___
*[`src/BloomBrowserUI/pageChooser/PageChooserDialog.tsx` line 427 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5612#-NNNgHmFDHvRz1jml2Rl:-NNwIAbM0A6gbPEZgtXk:b-yofl0j) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/bc4b34c0892601bcdc0adb85c11fbb5064b98947/src/BloomBrowserUI/pageChooser/PageChooserDialog.tsx#L427)):*

I wish all these were props or fields in a selectedTemplatePage object or something similar, rather than something we have to go to the DOM for.
</blockquote></details>

Done. The new selection mechanism straightens this out a lot.
___
*[`src/BloomBrowserUI/pageChooser/PageChooserDialog.tsx` line 487 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5612#-NNNhWkM7W8EbCRnDL0T:-NNwIQql4Ibygy-f7LQF:b-npss0j) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/bc4b34c0892601bcdc0adb85c11fbb5064b98947/src/BloomBrowserUI/pageChooser/PageChooserDialog.tsx#L487)):*

That feels wrong. React should render the dialog consistent with the current state., and re-render as the state changes. If it doesn't, you're probably holding state in something React doesn't know about, or some useEffect needs more cases under which it re-runs. (If some data it fetched from C# can become out of date, we have mechanisms for broadcasting and watching for such changes.)
</blockquote></details>

The main problem seemed to be in the guts of the `useL10n()` hook. If we didn't remove and recreate the container, changing the `forChooseLayout` prop would always return the previous string from the hook. I wonder if the first usage was being cached by the LocalizationManager, or something. Ideas are welcome.
___
*[`src/BloomBrowserUI/react_components/BloomDialog/BloomDialog.tsx` line 294 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5612#-NNNldGk2T5w7BsJfAYi:-NNhFz-59q5h77xoQKDn:bcuhnzw) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/bc4b34c0892601bcdc0adb85c11fbb5064b98947/src/BloomBrowserUI/react_components/BloomDialog/BloomDialog.tsx#L294)):*

This is not a new comment, but it's not very helpful. DraggablePaper is a function now (props => {...}). Was it at some point a function but not a FunctionComponent? What mistake exactly is this comment helping us not repeat?
</blockquote></details>

Done. Changed the comment. The old code was like this:
```
const PaperComponent = paperProps => {
        return (
            <Draggable
                handle="#draggable-dialog-title"
                cancel={'[class*="MuiDialogContent-root"]'}
            > etc.
```
___
*[`src/BloomBrowserUI/pageChooser/ErrorGroup.tsx` line 12 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5612#-NNNSdE--AUkCzKoFk7O:-NNcJvtq8t2mmaru477d:b-j9wu) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/bc4b34c0892601bcdc0adb85c11fbb5064b98947/src/BloomBrowserUI/pageChooser/ErrorGroup.tsx#L12)):*

More specific name? And/Or class comment? I gather this class is used in place of showing a template book's pages when something is wrong?
</blockquote></details>

`TemplateGroupErrorReplacement` :+1:

JohnThomson (changes requested):
Thanks for doing all that. It's a bit difficult to follow all the changes, but I think it's in pretty good shape now. I just picked up a few minor things.

> Reviewed 9 of 9 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5612#-:-NNxHyAM8fXOPd5bcFIB:b-wv3ie5)* status: all files reviewed, 6 unresolved discussions (waiting on @gmartin7)

___
*[`src/BloomBrowserUI/pageChooser/ChooserPageGroup.tsx` line 40 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5612#-NNx8NjK0lG8VC1P1oVR:-NNx8NjK0lG8VC1P1oVS:byasrr0) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/6e3c981871de8beafad5650e0622fe324a0475b5/src/BloomBrowserUI/pageChooser/ChooserPageGroup.tsx#L40)):*
> ```XML
> 
> // Unicode character used to mark pages that are only available to enterprise users.
> const bulletCharacter = "\u25cf";
> ```

It's not obvious why a 'bullet' character would be associated with Enterprise. Maybe enterpriseCharacter?
___
*[`src/BloomBrowserUI/pageChooser/ChooserPageGroup.tsx` line 156 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5612#-NNx9VLQD4fQvJP4l5b_:-NNx9VLQD4fQvJP4l5ba:b-e7z3i8) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/6e3c981871de8beafad5650e0622fe324a0475b5/src/BloomBrowserUI/pageChooser/ChooserPageGroup.tsx#L156)):*
> ```XML
>                           display: inline-block;
>                           // This makes the blue overlay, which is a child, be correctly positioned
>                           // even when scrolled.
> ```

I'm not sure this comment needs changing, but you might want to review the stuff about position in our CSS course. It's not only scrolling that can throw this off! An absolutely positioned element uses top, left, etc to position itself next to the nearest ancestor that has SOME kind of positioning. The nice thing about position:relative is that it usually does nothing to the position of the element that has it (unless it has its own top: etc). But it means that absolutely positioned children will be positioned relative to this, not some ancestor.
I would have commented more like "the absolutely positioned blue overlay child needs to be placed relative to this, not an ancestor"
___
*[`src/BloomBrowserUI/pageChooser/ChooserPageGroup.tsx` line 257 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5612#-NNxBGta3YQsZFypK2oZ:-NNxBGta3YQsZFypK2o_:bqnkdjt) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/6e3c981871de8beafad5650e0622fe324a0475b5/src/BloomBrowserUI/pageChooser/ChooserPageGroup.tsx#L257)):*
> ```XML
>             )}
>             {errorState && (
>                 <TemplateGroupErrorReplacement
> ```

Is a TemplateGroup something different from a TemplateBook? If not let's use the same name.
___
*[`src/BloomBrowserUI/pageChooser/PageChooserDialog.tsx` line 487 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5612#-NNNhWkM7W8EbCRnDL0T:-NNxGYwM6nfUpc0llEU5:bblk7m) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/bc4b34c0892601bcdc0adb85c11fbb5064b98947/src/BloomBrowserUI/pageChooser/PageChooserDialog.tsx#L487)):*

The main problem seemed to be in the guts of the `useL10n()` hook. If we didn't remove and recreate the container, changing the `forChooseLayout` prop would always return the previous string from the hook. I wonder if the first usage was being cached by the LocalizationManager, or something. Ideas are welcome.
</blockquote></details>

How is it that forChooseLayout is even changing? Isn't that fixed for the life of the dialog?
If we want a different string, surely it should have a different L10nId? And I would consider it a bug if useL10n is not re-rendering when the ID changes.
___
*[`src/BloomBrowserUI/pageChooser/PageChooserDialog.tsx` line 136 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5612#-NNxDrlH6Nx05gmk4CXx:-NNxDrlH6Nx05gmk4CXy:bm85m40) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/6e3c981871de8beafad5650e0622fe324a0475b5/src/BloomBrowserUI/pageChooser/PageChooserDialog.tsx#L136)):*
> ```XML
>                 const imgSrc = img.src.replace(/%2F/g, "/");
>                 // force the image to be reloaded
>                 if (imgSrc.startsWith(src)) {
> ```

Why startsWith here? Won't it reload without help if the src is changed in any way?
___
*[`src/BloomBrowserUI/pageChooser/ErrorGroup.tsx` line 12 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5612#-NNNSdE--AUkCzKoFk7O:-NNxHNuR4th11Txnr7Lc:bgkydna) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/bc4b34c0892601bcdc0adb85c11fbb5064b98947/src/BloomBrowserUI/pageChooser/ErrorGroup.tsx#L12)):*

`TemplateGroupErrorReplacement` :+1: 
</blockquote></details>

As noted above, unless a TemplateGroup is something different from a TemplateBook, I think you should use TemplateBook here too.

gmartin7 (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5612)* status: 22 of 32 files reviewed, 6 unresolved discussions (waiting on @JohnThomson)

___
*[`src/BloomBrowserUI/pageChooser/PageChooserDialog.tsx` line 487 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5612#-NNNhWkM7W8EbCRnDL0T:-NNxZSfJ471rwRPps_eX:bhu1nsc) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/bc4b34c0892601bcdc0adb85c11fbb5064b98947/src/BloomBrowserUI/pageChooser/PageChooserDialog.tsx#L487)):*

How is it that forChooseLayout is even changing? Isn't that fixed for the life of the dialog?
If we want a different string, surely it should have a different L10nId? And I would consider it a bug if useL10n is not re-rendering when the ID changes.
</blockquote></details>

Yeah, I think there's a bug in the `useL10n()` hook. If you don't recreate the whole dialog container, you can open the Add Page dialog (it'll have the correct title), close it and open the Change Layout dialog, it'll still say Add Page and vice versa. I've created BL-11927 to investigate that.
___
*[`src/BloomBrowserUI/pageChooser/PageChooserDialog.tsx` line 136 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5612#-NNxDrlH6Nx05gmk4CXx:-NOBF9s695RIZxlS1qNi:btaabu0) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/6e3c981871de8beafad5650e0622fe324a0475b5/src/BloomBrowserUI/pageChooser/PageChooserDialog.tsx#L136)):*

Why startsWith here? Won't it reload without help if the src is changed in any way?
</blockquote></details>

I added some comments and renamed the `src` variable to `updatedThumbUrl`.
The basic problem here is that `imgSrc` always starts with a param "generateThumbnailIfNecessary=true" postpended, whereas the url that's coming into the method doesn't.
___
*[`src/BloomBrowserUI/pageChooser/ChooserPageGroup.tsx` line 40 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5612#-NNx8NjK0lG8VC1P1oVR:-NOAt38-E_XAHsahCqH7:b-h38hqs) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/6e3c981871de8beafad5650e0622fe324a0475b5/src/BloomBrowserUI/pageChooser/ChooserPageGroup.tsx#L40)):*

It's not obvious why a 'bullet' character would be associated with Enterprise. Maybe enterpriseCharacter?
</blockquote></details>

switched to `enterpriseMarkerChar`
___
*[`src/BloomBrowserUI/pageChooser/ChooserPageGroup.tsx` line 156 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5612#-NNx9VLQD4fQvJP4l5b_:-NOAtmb54jf3KTvmCDnp:b-ix40j6) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/6e3c981871de8beafad5650e0622fe324a0475b5/src/BloomBrowserUI/pageChooser/ChooserPageGroup.tsx#L156)):*

I'm not sure this comment needs changing, but you might want to review the stuff about position in our CSS course. It's not only scrolling that can throw this off! An absolutely positioned element uses top, left, etc to position itself next to the nearest ancestor that has SOME kind of positioning. The nice thing about position:relative is that it usually does nothing to the position of the element that has it (unless it has its own top: etc). But it means that absolutely positioned children will be positioned relative to this, not some ancestor.
I would have commented more like "the absolutely positioned blue overlay child needs to be placed relative to this, not an ancestor"
</blockquote></details>

Done. Thanks for the clarification.
___
*[`src/BloomBrowserUI/pageChooser/ChooserPageGroup.tsx` line 257 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5612#-NNxBGta3YQsZFypK2oZ:-NOAtEsCDoad5gTiubm6:b-9gwtfq) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/6e3c981871de8beafad5650e0622fe324a0475b5/src/BloomBrowserUI/pageChooser/ChooserPageGroup.tsx#L257)):*

Is a TemplateGroup something different from a TemplateBook? If not let's use the same name.
</blockquote></details>

Using `TemplateBookErrorReplacement`.
___
*[`src/BloomBrowserUI/pageChooser/ErrorGroup.tsx` line 12 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5612#-NNNSdE--AUkCzKoFk7O:-NOAsQMCDuT8gtZnv3_H:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/bc4b34c0892601bcdc0adb85c11fbb5064b98947/src/BloomBrowserUI/pageChooser/ErrorGroup.tsx#L12)):*

As noted above, unless a TemplateGroup is something different from a TemplateBook, I think you should use TemplateBook here too.
</blockquote></details>

Done.

JohnThomson (approved):
I'm ready to approve this. I'll leave it open for now in case you think the startsWith is worth improving after my last comment, but I'm also OK if you decide it's not.

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5612#-:-NOGiUM4DYAHMJTz7dOl:b-vnx6tr)* status: 22 of 32 files reviewed, 1 unresolved discussion (waiting on @gmartin7)

gmartin7 (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5612)* status: 22 of 32 files reviewed, 1 unresolved discussion (waiting on @JohnThomson)

___
*[`src/BloomBrowserUI/pageChooser/PageChooserDialog.tsx` line 136 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5612#-NNxDrlH6Nx05gmk4CXx:-NOGjHOJCOMt6FcrtjZa:b-vymgd6) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/6e3c981871de8beafad5650e0622fe324a0475b5/src/BloomBrowserUI/pageChooser/PageChooserDialog.tsx#L136)):*

That makes sense but if the page has pictures called, say, "dogs" and "dogs running" and "dogs playing" they will all update when you only meant to update the first one. That may be unlikely enough not to matter...I think the worst result would be a slight delay, maybe a flicker. The more precise thing would to match anything up to the question mark (if any). But it may not be worth it.
</blockquote></details>

I think maybe we won't bother unless a user brings it up, since our published templates generally won't even have this problem. This is only for templates where there is no `.svg` of the Portrait/Landscape version of the page already stored in the templates directory of the book.

JohnThomson (approved):
> Reviewed 6 of 10 files at r4, 1 of 1 files at r5, 2 of 2 files at r6, 1 of 1 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5612)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @gmartin7)

PR #5616, opened 2023-02-06
Fix wrong tooltip for overlay images (BL-11901)

jsubloom:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5616)


sujeffreyl (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5616)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @JohnThomson)

___
*[`src/BloomBrowserUI/bookEdit/js/bloomImages.ts` line 471 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5616#-NNcCUneAEdZjE30pGMR:-NNcCUnfESgAuY2CzaqT:bmi2m0t) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/e02730006f1ab2d570fef9102a1c4a6daaf68202/src/BloomBrowserUI/bookEdit/js/bloomImages.ts#L471)):*
> ```XML
>             // the tooltip will not appear (until you leave the main image and come back).
>             // The tooltip was actually updated, the browser just doesn't want to re-display it.
>             // I don't think it's worth complicating the code to try to address that minor corner case?
> ```

Agree or disagree?

JohnThomson (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5616)* status: 0 of 3 files reviewed, 4 unresolved discussions (waiting on @jsubloom)

___
*[`src/BloomBrowserUI/bookEdit/js/bloomImages.ts` line 440 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5616#-NNh7WKZDHds0vVTZjlN:-NNh7WKZDHds0vVTZjlO:b-dmhsv9) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/e02730006f1ab2d570fef9102a1c4a6daaf68202/src/BloomBrowserUI/bookEdit/js/bloomImages.ts#L440)):*
> ```XML
>     const patriarch = container.farthest<HTMLElement>(".bloom-imageContainer");
>     if (patriarch) {
>         patriarch.title = "";
> ```

I think it might help, if container !== patriarch, to set container's title to its OWN data-title. (That way, when you disable the tooltip for a child container, you restore the main one.) (Maybe it's not that simple...you might need to apply some of the UpdateImageTooltipVisibility logic to patriarch.)
___
*[`src/BloomBrowserUI/bookEdit/js/bloomImages.ts` line 471 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5616#-NNcCUneAEdZjE30pGMR:-NNh8-u15knIeftvZW4b:b-e2bff4) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/e02730006f1ab2d570fef9102a1c4a6daaf68202/src/BloomBrowserUI/bookEdit/js/bloomImages.ts#L471)):*

Agree or disagree?
</blockquote></details>

I think the problem is what I commented on above: when the pointer leaves the child image container, you are detecting this, but you set the root image container's title to empty rather than its own data-title.
So, I would agree that it's not worth a lot of complication to get the title to reappear without re-entering the container, but if there's an easy fix it might be worth it. At least, verify that in the situation where you're not getting the tooltip, it's not just because the parent in fact doesn't have a title.
___
*[`src/BloomBrowserUI/bookEdit/js/bloomImages.ts` line 478 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5616#-NNh5yfDB5Ed4AbuZFDA:-NNh5yfDB5Ed4AbuZFDB:b-lnjz5l) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/e02730006f1ab2d570fef9102a1c4a6daaf68202/src/BloomBrowserUI/bookEdit/js/bloomImages.ts#L478)):*
> ```XML
> 
>             if (patriarch) {
>                 patriarch.title = dataTitle;
> ```

I think you need to be more careful about doing this. One call chain for UpdateImageTooltipVisibility is from EnableAllImageEditing, which will call it for every image container. If there's more than one that is not suppressing image buttons, it will be somewhat arbitrary which title ends up on the patriarch. I'm not quite sure how to fix this, but at least consider which title should win in EnableAllImageEditing.
___
*[`src/BloomBrowserUI/utils/elementExtensions.ts` line 19 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5616#-NNhAb6I4_WxaK0gZBNJ:-NNhAb6I4_WxaK0gZBNK:b-l9e2em) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/e02730006f1ab2d570fef9102a1c4a6daaf68202/src/BloomBrowserUI/utils/elementExtensions.ts#L19)):*
> ```XML
>  * If no self-or-ancestor elements match the selector, then returns null.
>  */
> Element.prototype.farthest = function<E extends Element = Element>(
> ```

It's cute that JS lets you do this, but I'm not sure it's a good idea, at least not unless the extension method is one we use constantly. If you just make a function farthest (element, selector) and export/import that, it's just about as convenient to use, but can't be confused with the built-in methods of Element.
I'm willing to be persuaded. If you still think it's a good idea, it might be a worthwhile team discussion.

jsubloom (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5616)* status: 0 of 4 files reviewed, 3 unresolved discussions (waiting on @JohnThomson)

___
*[`src/BloomBrowserUI/bookEdit/js/bloomImages.ts` line 440 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5616#-NNh7WKZDHds0vVTZjlN:-NNlWUmU39asxET0RyZi:b-yrxsrc) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/e02730006f1ab2d570fef9102a1c4a6daaf68202/src/BloomBrowserUI/bookEdit/js/bloomImages.ts#L440)):*

I think it might help, if container !== patriarch, to set container's title to its OWN data-title. (That way, when you disable the tooltip for a child container, you restore the main one.) (Maybe it's not that simple...you might need to apply some of the UpdateImageTooltipVisibility logic to patriarch.)
</blockquote></details>

I was confused for a while, but now I think you mean "set PATRIARCH'S title to its OWN data-title".

Interesting idea, but no, turns out it's not an improvement (and it actually breaks another requirement). It's because this function represents when we want to completely disable the tooltip (as in, no tooltip, not for this container nor for any other image container, should show). So, it is intentional and correct to set the main one to "". (It's wrong to show the patriarch's own title in that case)
___
*[`src/BloomBrowserUI/bookEdit/js/bloomImages.ts` line 471 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5616#-NNcCUneAEdZjE30pGMR:-NNlcBNz5ZtCGQ6bUvA9:b3cfqmw) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/e02730006f1ab2d570fef9102a1c4a6daaf68202/src/BloomBrowserUI/bookEdit/js/bloomImages.ts#L471)):*

I think the problem is what I commented on above: when the pointer leaves the child image container, you are detecting this, but you set the root image container's title to empty rather than its own data-title.
So, I would agree that it's not worth a lot of complication to get the title to reappear without re-entering the container, but if there's an easy fix it might be worth it. At least, verify that in the situation where you're not getting the tooltip, it's not just because the parent in fact doesn't have a title.
</blockquote></details>

I had verified that the parent's title was set.
What I hadn't verified... is if the parent's title was still set.
When switching between image containers, the patriarch's title gets set (good) and then it gets accidentally cleared while we cleanup the old image container (oops!)
___
*[`src/BloomBrowserUI/bookEdit/js/bloomImages.ts` line 478 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5616#-NNh5yfDB5Ed4AbuZFDA:-NNlpUwn-rd3mQE3N0IK:bkhd55l) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/e02730006f1ab2d570fef9102a1c4a6daaf68202/src/BloomBrowserUI/bookEdit/js/bloomImages.ts#L478)):*

I think you need to be more careful about doing this. One call chain for UpdateImageTooltipVisibility is from EnableAllImageEditing, which will call it for every image container. If there's more than one that is not suppressing image buttons, it will be somewhat arbitrary which title ends up on the patriarch. I'm not quite sure how to fix this, but at least consider which title should win in EnableAllImageEditing.
</blockquote></details>

Hmm, it currently doesn't really matter, because you have to move your mouse off these in order to get EnableAllImageEditing to trigger. When you put your mouse back on the images, stuff will happen that calls UpdateImageTooltipVisibility so it didn't matter what it got set to.
But, I figured it was cheap/reasonable enough to have the first one in document order win in EnableAllImageEditing. That required reversing the traverse order.

I guess the unfortunate thing about this overall approach (with patriarch.title) is that it introduces more state/side effects to consider.  But that sounds less complex/less risk than introducing more z-indexes to consider.
___
*[`src/BloomBrowserUI/utils/elementExtensions.ts` line 19 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5616#-NNhAb6I4_WxaK0gZBNJ:-NNlR4W-1cb6AjF1TOmL:bu7b7l1) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/e02730006f1ab2d570fef9102a1c4a6daaf68202/src/BloomBrowserUI/utils/elementExtensions.ts#L19)):*

It's cute that JS lets you do this, but I'm not sure it's a good idea, at least not unless the extension method is one we use constantly. If you just make a function farthest (element, selector) and export/import that, it's just about as convenient to use, but can't be confused with the built-in methods of Element.
I'm willing to be persuaded. If you still think it's a good idea, it might be a worthwhile team discussion.
</blockquote></details>

I disagree that it's as convenient to use. One reason is autocomplete reasons and another is that it's less obvious what the effects are when it's function (obj, args...) compared to obj.function(args...).
However, I do think the extensions thing is a little too cute/flimsy, and I don't like that VSCode/Typescript don't automatically know that you need to import elementExtensions when you use farthest... so... maybe I'll get rid of this.

JohnThomson (changes requested):
> Reviewed 1 of 3 files at r1, 3 of 3 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5616)* status: all files reviewed, 2 unresolved discussions (waiting on @jsubloom)

___
*[`src/BloomBrowserUI/bookEdit/js/bloomImages.ts` line 440 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5616#-NNh7WKZDHds0vVTZjlN:-NNmbj1B3-9-NVzrE8Fp:b-mrarad) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/e02730006f1ab2d570fef9102a1c4a6daaf68202/src/BloomBrowserUI/bookEdit/js/bloomImages.ts#L440)):*

I was confused for a while, but now I think you mean "set PATRIARCH'S title to its OWN data-title".

Interesting idea, but no, turns out it's not an improvement (and it actually breaks another requirement). It's because this function represents when we want to completely disable the tooltip (as in, no tooltip, not for this container nor for any other image container, should show). So, it is intentional and correct to set the main one to "". (It's wrong to show the patriarch's own title in that case)
</blockquote></details>

You're right about what I meant, sorry, and also sorry it proved a red herring. But it still seems that something should be detecting that we moved the mouse out of the child image container and restoring the title of the patriarch.
___
*[`src/BloomBrowserUI/bookEdit/js/bloomImages.ts` line 471 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5616#-NNcCUneAEdZjE30pGMR:-NNmcSenENnKVFxlQGkq:b-9plztf) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/e02730006f1ab2d570fef9102a1c4a6daaf68202/src/BloomBrowserUI/bookEdit/js/bloomImages.ts#L471)):*

I had verified that the parent's title was set.
What I hadn't verified... is if the parent's title was still set.
When switching between image containers, the patriarch's title gets set (good) and then it gets accidentally cleared while we cleanup the old image container (oops!)
</blockquote></details>

Looks like you found something that needed fixing, but I haven't spotted the fix. Did something not get checked in?
___
*[`src/BloomBrowserUI/bookEdit/js/bloomImages.ts` line 478 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5616#-NNh5yfDB5Ed4AbuZFDA:-NNmcLJ0DvDEy9nHDqkF:buhkzts) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/e02730006f1ab2d570fef9102a1c4a6daaf68202/src/BloomBrowserUI/bookEdit/js/bloomImages.ts#L478)):*

Hmm, it currently doesn't really matter, because you have to move your mouse off these in order to get EnableAllImageEditing to trigger. When you put your mouse back on the images, stuff will happen that calls UpdateImageTooltipVisibility so it didn't matter what it got set to.
But, I figured it was cheap/reasonable enough to have the first one in document order win in EnableAllImageEditing. That required reversing the traverse order.

I guess the unfortunate thing about this overall approach (with patriarch.title) is that it introduces more state/side effects to consider.  But that sounds less complex/less risk than introducing more z-indexes to consider.
</blockquote></details>

I haven't spotted that change (reversing the traverse order)
___
*[`src/BloomBrowserUI/utils/elementExtensions.ts` line 19 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5616#-NNhAb6I4_WxaK0gZBNJ:-NNmb1mS65Qc74GZml-T:b8fe4u8) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/e02730006f1ab2d570fef9102a1c4a6daaf68202/src/BloomBrowserUI/utils/elementExtensions.ts#L19)):*

I disagree that it's as convenient to use. One reason is autocomplete reasons and another is that it's less obvious what the effects are when it's function (obj, args...) compared to obj.function(args...).
However, I do think the extensions thing is a little too cute/flimsy, and I don't like that VSCode/Typescript don't automatically know that you need to import elementExtensions when you use farthest... so... maybe I'll get rid of this.
</blockquote></details>

Does it autocomplete if you haven't imported it yet? If not, you still have to remember the name to get started.
But I still think we spend more time reading code than writing it. An extension might make sense for something we use constantly, but for something only used once or twice in the whole code base, I think it's more important for it to be clear that it is an extension.
But by all means let's see what others think.

jsubloom (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5616)* status: all files reviewed, 2 unresolved discussions (waiting on @JohnThomson)

___
*[`src/BloomBrowserUI/bookEdit/js/bloomImages.ts` line 478 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5616#-NNh5yfDB5Ed4AbuZFDA:-NNmpmUz5u7CVu_z-Zfe:baqkhdf) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/e02730006f1ab2d570fef9102a1c4a6daaf68202/src/BloomBrowserUI/bookEdit/js/bloomImages.ts#L478)):*

I haven't spotted that change (reversing the traverse order)
</blockquote></details>

Are you sure you're looking at revision 2? it's in bloomImages.ts, look for "reverse()"

JohnThomson (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5616)* status: all files reviewed, 1 unresolved discussion (waiting on @jsubloom)

PR #5626, opened 2023-02-09
Bulk Upload: dry-run as a flag instead of a destination

jsubloom:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5626)

### DIFF
--- a/src/BloomExe/CLI/UploadCommand.cs
+++ b/src/BloomExe/CLI/UploadCommand.cs
@@ -36,15 +36,15 @@ namespace Bloom.CLI
 			options.Dest = options.Dest.ToLowerInvariant();
 			switch (options.Dest)
 			{
-				case UploadDestination.DryRun:
 				case UploadDestination.Development:
 				case UploadDestination.Production:
 					break;
 				default:
-					Console.WriteLine($"Error: if present, upload destination (-d) must be one of {UploadDestination.DryRun}, {UploadDestination.Development}, or {UploadDestination.Production}");
+					Console.WriteLine($"Error: Upload destination (-d) must be one of {UploadDestination.Development} or {UploadDestination.Production}");
 					return 1;
 			}
 			BookUpload.Destination = options.Dest;    // must be set before calling SetupErrorHandling() (or BloomParseClient constructor)
+			BookUpload.IsDryRun = options.DryRun;	  // must be set before calling SetupErrorHandling() (or BloomParseClient constructor)
 
 			// This task will be all the program does. We need to do enough setup so that
 			// the upload code can work, then tear it down.
@@ -61,13 +61,15 @@ namespace Bloom.CLI
 					var uploader = new BulkUploader(singleBookUploader);
 
 
+					if (options.DryRun)
+					{
+						Console.WriteLine($"\nThe following actions would happen if --dry-run were disabled:");
+					}
+
 					// Since Bloom is not a normal console app, when run from a command line, the new command prompt
 					// appears at once. The extra newlines here are attempting to separate this from our output.
 					switch (options.Dest)
 					{
-						case UploadDestination.DryRun:
-							Console.WriteLine($"\nThe following actions would happen if you set destination to '{(BookUpload.UseSandboxByDefault ? UploadDestination.Development : UploadDestination.Production)}'.");
-							break;
 						case UploadDestination.Development:
 							Console.WriteLine("\nThe upload will go to dev.bloomlibrary.org.");
 							break;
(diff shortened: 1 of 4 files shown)


jsubloom (commented):
I didn't realize it when I made my initial fix, but then a thought struck me: Why was DryRun a destination? Isn't it normally an extra flag? And if it were a flag, JH's initial version of the check would've worked perfectly fine without  me making the check more complicated.

So i tried making it a flag.
Old Command:
Bloom.exe upload "[path]" --user [userEmail] --destination dry-run

New Command:
Bloom.exe upload "[path]" --user [userEmail] --destination dev --dry-run

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5626#-:-NNrJuqF1hc1xcQ5qugL:bmng6h3)* status: 0 of 3 files reviewed, all discussions resolved (waiting on @andrew-polk and @StephenMcConnel)

andrew-polk (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5626)* status: all files reviewed, 1 unresolved discussion (waiting on @jsubloom and @StephenMcConnel)

___
*[`src/BloomExe/CLI/UploadCommand.cs` line 122 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5626#-NNrKfG52Qc0H9M4AZLU:-NNrKfG52Qc0H9M4AZLV:b-onz540) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/2925a8338a0fa1572c26ac675c6b0336ce4f2246/src/BloomExe/CLI/UploadCommand.cs#L122)):*
> ```Smalltalk
> 
> 	[Option("dry-run", Default=false, HelpText = "If present, performs a dry-run (will just print out what would happen).", Required = false)]
> 	public bool DryRun { get; set; }
> ```

It scares me a bit to go from dry run being the default to not.
I guess you've evaluated the places we run this to ensure there are no surprises?

jsubloom (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5626)* status: all files reviewed, 1 unresolved discussion (waiting on @andrew-polk and @StephenMcConnel)

___
*[`src/BloomExe/CLI/UploadCommand.cs` line 122 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5626#-NNrKfG52Qc0H9M4AZLU:-NNrLDSOBpBjl-50E_A9:b-mvkg96) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/2925a8338a0fa1572c26ac675c6b0336ce4f2246/src/BloomExe/CLI/UploadCommand.cs#L122)):*

It scares me a bit to go from dry run being the default to not.
I guess you've evaluated the places we run this to ensure there are no surprises?
</blockquote></details>

Yeah... it's an interesting question.
Well, apparently for 2+ years nobody ran it in the default mode!!!  (unless they figured out how to get the old code to run in dry-run mode better than me)  They probably explicitly specified the destination

For the CLI, the default goes to Dev now. That mitigates some of the risk

For the BookUpload class: while in terms of the CLI, the default if you don't specify it is false... if you neglect to tell BookUpload at all, the default is true.
Actually, I probably broke some tests that expect it to be false...

jsubloom (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5626)* status: 3 of 4 files reviewed, 1 unresolved discussion (waiting on @andrew-polk and @StephenMcConnel)

___
*[`src/BloomExe/CLI/UploadCommand.cs` line 122 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5626#-NNrKfG52Qc0H9M4AZLU:-NNrQ0Ys7oZHTMvQzhiW:b-dg6f23) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/2925a8338a0fa1572c26ac675c6b0336ce4f2246/src/BloomExe/CLI/UploadCommand.cs#L122)):*

Yeah... it's an interesting question.
Well, apparently for 2+ years nobody ran it in the default mode!!!  (unless they figured out how to get the old code to run in dry-run mode better than me)  They probably explicitly specified the destination

For the CLI, the default goes to Dev now. That mitigates some of the risk

For the BookUpload class: while in terms of the CLI, the default if you don't specify it is false... if you neglect to tell BookUpload at all, the default is true.
Actually, I probably broke some tests that expect it to be false...
</blockquote></details>

fixed the unit tests

andrew-polk (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5626)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @StephenMcConnel)

PR #5640, opened 2023-02-16
Improvement for translation group focus issue (BL-11922)

andrew-polk:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5640)

### DIFF
--- a/src/BloomBrowserUI/bookEdit/js/bloomEditing.ts
+++ b/src/BloomBrowserUI/bookEdit/js/bloomEditing.ts
@@ -468,7 +468,7 @@ export function SetupElements(container: HTMLElement) {
             // If we get focus on a translation group, move it to the first editable. See BL-11922.
             $(this).focus(e => {
                 $(e.target)
-                    ?.find("div.bloom-editable")
+                    ?.find("div.bloom-editable:visible")
                     ?.first()
                     ?.focus();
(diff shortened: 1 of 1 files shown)


JohnThomson (approved):
I'm going to merge this, because it's an important improvement. However, I have a thought about the solution. You are currently focusing the first visible bloom-editable.  That's definitely better than focusing the TG or a bloom-editable that is not visible (!). However, if there are several matches, it will pick the first bloom-editable in document order. IIRC, we use CSS to reorder bloom-editables within a multilingual TG, so first-in-document-order is not necessarily the one highest up the screen that the user would probably expect.
This seems to be mainly for overlays, which only show one language, so it may not be an issue. But I thought it was worth considering.

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5640#-:-NOQBPeA8b8pggZXHFZr:bmc034z)* status: 0 of 1 files reviewed, all discussions resolved

andrew-polk:
Thanks. Yeah, I did think that it wouldn't be sufficient outside the current scenario.
But you've made me think about it again... we don't just order overlays. Any text box can be ordered.
So there is the possibility of running into the scenario you describe.
I'm playing around with it now...

PR #5643, opened 2023-02-16
[5.4] BL-11915 Improve alignment of activity text

gmartin7:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5643)

### DIFF
--- a/src/content/templates/template books/Activity/simple-dom-choice-activities.less
+++ b/src/content/templates/template books/Activity/simple-dom-choice-activities.less
@@ -43,6 +43,19 @@
                     // this number, 95px, was chosen to use the space available in 16x9 portrait, with these text samples: སྔགས་ཀྱི་ཀློག་བས་ཤེས་རབ་མི
                     // But, fixing this interferes with the user's ability to set line-height in the format dialog.
                     // line-height: 95px !important;
+                    &.bloom-visibility-code-on {
+                        min-height: 1.5em;
+                        display: flex;
+                        justify-content: center;
+                        align-items: center;
+                        p {
+                            // Table is an unusual choice here, but the only options here which keep the
+                            // text centered vertically on the button no matter the line-spacing are:
+                            // table, inline-table, and contents. Contents apparently has detrimental effects
+                            // on accessibility.
+                            display: table;
+                        }
+                    }
                 }
             }
         }
@@ -149,6 +162,19 @@
             flex-grow: 1;
             .bloom-editable {
                 min-height: unset;
+                &.bloom-visibility-code-on {
+                    min-height: 1.5em;
+                    display: flex;
+                    justify-content: center;
+                    align-items: center;
+                    p {
+                        // Table is an unusual choice here, but the only options here which keep the
+                        // text centered vertically on the button no matter the line-spacing are:
+                        // table, inline-table, and contents. Contents apparently has detrimental effects
+                        // on accessibility.
+                        display: table;
+                    }
+                }
             }
         }
 


JohnThomson (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5643)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @gmartin7)

___
*[`src/content/templates/template books/Activity/simple-dom-choice-activities.less` line 51 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5643#-NOWPK_W8EwuD19gajlK:-NOWPK_W8EwuD19gajlL:bdvin5o) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/9bd71bbe3b659295f33238f346e8d9d3a89f5815/src/content/templates/template%20books/Activity/simple-dom-choice-activities.less#L51)):*
> ```Less
>                         justify-content: center;
>                         p {
>                             display: inline-table;
> ```

It's very unusual to make something an inline-table. My reading indicates that it is no different from inline-block (except for generating extra wasted stuff) unless there are table-component children, which is not the case here. If there's a reason for it, it should be commented.
I'm not even clear why you want 'inline'. Paragraphs are already block elements and there's not any surrounding text for them to be inline with, so I would have expected their default block mode to work here (with the parent display:flex making a horizontal row, in the unlikely event that there is more than one paragraph of text in a button).
So I think whatever you're trying to do here needs explanation.
___
*[`src/content/templates/template books/Activity/simple-dom-choice-activities.less` line 52 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5643#-NOWRPvLDQ_lMENkwdZt:-NOWRPvLDQ_lMENkwdZu:bvlb4t) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/9bd71bbe3b659295f33238f346e8d9d3a89f5815/src/content/templates/template%20books/Activity/simple-dom-choice-activities.less#L52)):*
> ```Less
>                         p {
>                             display: inline-table;
>                             align-self: center;
> ```

Did you try just putting align-items: center on the parent?

gmartin7 (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5643)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @JohnThomson)

___
*[`src/content/templates/template books/Activity/simple-dom-choice-activities.less` line 51 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5643#-NOWPK_W8EwuD19gajlK:-NOpAUpbBbIacBaMW8Ee:b-nkksks) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/9bd71bbe3b659295f33238f346e8d9d3a89f5815/src/content/templates/template%20books/Activity/simple-dom-choice-activities.less#L51)):*

It's very unusual to make something an inline-table. My reading indicates that it is no different from inline-block (except for generating extra wasted stuff) unless there are table-component children, which is not the case here. If there's a reason for it, it should be commented.
I'm not even clear why you want 'inline'. Paragraphs are already block elements and there's not any surrounding text for them to be inline with, so I would have expected their default block mode to work here (with the parent display:flex making a horizontal row, in the unlikely event that there is more than one paragraph of text in a button).
So I think whatever you're trying to do here needs explanation.
</blockquote></details>

Done. Added explanation. Switched to plain `display: table`, which also works.
___
*[`src/content/templates/template books/Activity/simple-dom-choice-activities.less` line 52 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5643#-NOWRPvLDQ_lMENkwdZt:-NOpAaCI3eMvnZdOg_hx:beynqei) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/9bd71bbe3b659295f33238f346e8d9d3a89f5815/src/content/templates/template%20books/Activity/simple-dom-choice-activities.less#L52)):*

Did you try just putting align-items: center on the parent?
</blockquote></details>

Done. I thought I had! Maybe I just tried align-contents, which wasn't the same. ;(

JohnThomson (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5643)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @gmartin7)

PR #5652, opened 2023-02-21
Warn about audio in video before starting to show the video (BL-11947)

StephenMcConnel:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5652)

### DIFF
--- a/src/BloomExe/Book/HtmlDom.cs
+++ b/src/BloomExe/Book/HtmlDom.cs
@@ -2310,6 +2310,18 @@ namespace Bloom.Book
 			return element.SafeSelectNodes(".//div[contains(@class,'bloom-videoContainer')]/video/source");
 		}
 
+		public IEnumerable<string> GetAllVideoPaths()
+		{
+			var paths = new List<string>();
+			foreach (var source in SelectChildVideoSourceElements(RawDom.DocumentElement).Cast<XmlElement>())
+			{
+				var src = source.GetAttribute("src");
+				if (!String.IsNullOrEmpty(src))
+					paths.Add(src);
+			}
+			return paths;
+		}
+
 		private static readonly string kAudioSentenceElementsXPath = "descendant-or-self::node()[contains(@class,'audio-sentence') and string-length(@id) > 0]";
 		
 		public static XmlNodeList SelectRecordableDivOrSpans(XmlElement element)
--- a/src/BloomExe/Publish/Video/PublishAudioVideoAPI.cs
+++ b/src/BloomExe/Publish/Video/PublishAudioVideoAPI.cs
@@ -1,9 +1,13 @@
 using System;
+using System.Collections.Generic;
 using System.Diagnostics;
 using System.Linq;
 using System.Runtime.InteropServices;
+using System.Text;
 using System.Windows.Forms;
+using System.Xml;
 using Bloom.Api;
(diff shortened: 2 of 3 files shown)


JohnThomson (changes requested):
It will work as-is, but I think we can simplify a bit.

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5652#-:-NOoSfUV3YKMxZH-PsnL:bkmpgv5)* status: all files reviewed, 1 unresolved discussion (waiting on @StephenMcConnel)

___
*[`src/BloomExe/Publish/Video/PublishAudioVideoAPI.cs` line 348 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5652#-NOoQCou2tSutQ8swWCc:-NOoQCou2tSutQ8swWCd:b-r420vt) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/0aae5a020a4dbacfdf847ab4943e047dc2b7d1be/src/BloomExe/Publish/Video/PublishAudioVideoAPI.cs#L348)):*
> ```Smalltalk
> 			{
> 				var sources = ev.GetElementsByTagName("source");
> 				if (sources.Count > 0)
> ```

Maybe better to use HtmlDom.SelectChildVideoSourceElements? Maybe even create a method there that returns all the video paths? We're not doing very well at this, but the original idea was to try to keep knowledge of how Bloom books are represented in HTML in HtmlDom. Such a method might well filter out any empty ones, and then this method could just merge them using string.Join().
Actually, the caller of this is the only client of AnyVideoHasAudio, which just has to split them up again, so why not just make it accept a list, and then you don't need this at all!

StephenMcConnel (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5652)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @JohnThomson)

___
*[`src/BloomExe/Publish/Video/PublishAudioVideoAPI.cs` line 348 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5652#-NOoQCou2tSutQ8swWCc:-NOp8jQXALuaOpxKCpJA:bb7zy) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/0aae5a020a4dbacfdf847ab4943e047dc2b7d1be/src/BloomExe/Publish/Video/PublishAudioVideoAPI.cs#L348)):*

Maybe better to use HtmlDom.SelectChildVideoSourceElements? Maybe even create a method there that returns all the video paths? We're not doing very well at this, but the original idea was to try to keep knowledge of how Bloom books are represented in HTML in HtmlDom. Such a method might well filter out any empty ones, and then this method could just merge them using string.Join().
Actually, the caller of this is the only client of AnyVideoHasAudio, which just has to split them up again, so why not just make it accept a list, and then you don't need this at all!
</blockquote></details>

Done.  I've simplified things as much as possible, adding HtmlDom.GetAllVideoPaths() and changing RecordVideoWindow.AnyVideoHasAudio().

JohnThomson (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5652)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @StephenMcConnel)

PR #5658, opened 2023-02-23
Tweaks to PDF Print and other publish screens after UI review (BL-11266)

JohnThomson:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5658)

### DIFF
--- a/DistFiles/localization/en/Bloom.xlf
+++ b/DistFiles/localization/en/Bloom.xlf
@@ -4169,9 +4169,9 @@
         <note>ID: PublishTab.PdfMaker.PdfFile</note>
       </trans-unit>
       <trans-unit id="PublishTab.PdfMaker.PdfWithCmykSwopV2">
-        <source xml:lang="en">CMYK color (U.S. Web Coated (SWOP) v2)</source>
+        <source xml:lang="en">U.S. Web Coated (SWOP) v2</source>
         <note>ID: PublishTab.PdfMaker.PdfWithCmykSwopV2</note>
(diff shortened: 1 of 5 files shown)


hatton (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5658)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @JohnThomson)

andrew-polk:
*[`DistFiles/localization/en/Bloom.xlf` line 4174 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5658#-NOz1G5v96cp7rGl2Qle:-NOz1G5v96cp7rGl2Qlf:bzgb8ai) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/a2d752e719559cbc846eee9714f9a0e16aeb89f8/DistFiles/localization/en/Bloom.xlf#L4174)):*
> ```XML
>         <source xml:lang="en">U.S. Web Coated (SWOP) v2</source>
>         <note>ID: PublishTab.PdfMaker.PdfWithCmykSwopV2</note>
>         <note>displayed as a checkable menu option for the Publish tab, PDF options menu. 'SWOP' and 'CMYK' may not be translatable, they are print shop standards.</note>
> ```

The note still refers to CMYK, but the text doesn't.

I'm thinking about how this will affect 5.4 localization.
I think when we push this English string change, it will invalidate the existing translations and if we push translations to 5.4 again (which we will), any translation which was there will be lost.

So to prevent that, I think we either need to 
1. manually reinstate existing translations after they get invalidated
2. create a new translation unit with a new ID

... and I didn't get this comment in before the PR was merged.

Option 3: ignore because exposure is so low.

PR #5677, opened 2023-03-01
Remove reliance on null for all Publish Settings

hatton:
* Make constructors for all classes which create all children
* For various language arrays, instead of checking for null, check for empty array
* In method arguments, stop making provision of settings optional
* When deserializing from json, do that on top of a fully constructed object

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5677)

### DIFF
--- a/src/BloomExe/Book/BookCompressor.cs
+++ b/src/BloomExe/Book/BookCompressor.cs
@@ -48,11 +48,11 @@ namespace Bloom.Book
 				if (RobustFile.Exists(blankImage))
 					coverImagePath = blankImage;
 			}
-			if(coverImagePath != null)
+			if (coverImagePath != null)
 			{
 				var thumbPath = Path.Combine(destinationFolder, "thumbnail.png");
-				
-				RuntimeImageProcessor.GenerateEBookThumbnail(coverImagePath, thumbPath, heightAndWidth, heightAndWidth,System.Drawing.ColorTranslator.FromHtml(book.GetCoverColor()));
+
+				RuntimeImageProcessor.GenerateEBookThumbnail(coverImagePath, thumbPath, heightAndWidth, heightAndWidth, System.Drawing.ColorTranslator.FromHtml(book.GetCoverColor()));
 			}
 		}
 
@@ -176,7 +176,7 @@ namespace Bloom.Book
 			{
 				if (!filter.Filter(filePath))
 					continue;
-				
+
 				FileInfo fi = new FileInfo(filePath);
 
 				var entryName = dirNamePrefix + filePath.Substring(dirNameOffset);  // Makes the name in zip based on the folder
@@ -189,7 +189,7 @@ namespace Bloom.Book
 
 				byte[] modifiedContent = null;
 
-				if (overrides !=null && overrides.TryGetValue(filePath, out modifiedContent))
+				if (overrides != null && overrides.TryGetValue(filePath, out modifiedContent))
 					newEntry.Size = modifiedContent.Length;
 				else
 					newEntry.Size = fi.Length;
@@ -248,11 +248,11 @@ namespace Bloom.Book
 			dom.Load(filePath);
 			foreach (var node in dom.SafeSelectNodes("//SubscriptionCode").Cast<XmlElement>().ToArray())
 			{
-				node.RemoveAll();	// should happen at most once
+				node.RemoveAll();   // should happen at most once
 			}
 			foreach (var node in dom.SafeSelectNodes("//BrandingProjectName").Cast<XmlElement>().ToArray())
 			{
-				node.RemoveAll();	// should happen at most once
+				node.RemoveAll();   // should happen at most once
 				node.AppendChild(dom.CreateTextNode("Default"));
 			}
 			return dom.OuterXml;
@@ -290,7 +290,7 @@ namespace Bloom.Book
 					if (matchSuppress.Groups[3].Value.ToLower() != "true")
 					{
 						text = text.Substring(0, matchSuppress.Groups[3].Index) + "true"
-								+  text.Substring(matchSuppress.Groups[3].Index + matchSuppress.Groups[3].Length);
+								+ text.Substring(matchSuppress.Groups[3].Index + matchSuppress.Groups[3].Length);
 					}
 				}
 				else
@@ -322,11 +322,8 @@ namespace Bloom.Book
 		/// transparent backgrounds.
 		/// </remarks>
 		/// <returns>The bytes of the (possibly) adjusted image.</returns>
-		internal static byte[] GetImageBytesForElectronicPub(string filePath, bool needsTransparentBackground, ImagePublishSettings imagePublishSettings = null)
+		internal static byte[] GetImageBytesForElectronicPub(string filePath, bool needsTransparentBackground, ImagePublishSettings imagePublishSettings)
 		{
-			if (imagePublishSettings == null)
-				imagePublishSettings = ImagePublishSettings.Default;
-
 			var maxWidth = imagePublishSettings.MaxWidth;
 			var maxHeight = imagePublishSettings.MaxHeight;
 
@@ -360,11 +357,11 @@ namespace Bloom.Book
 						// we should keep the original just because it's a smaller file.
 						// - possibly we don't need a 32-bit bitmap? Unfortunately the 1bpp/4bpp/8bpp only tells us
 						// that the image uses two, 16, or 256 distinct colors, not what they are or what precision they have.
-							case PixelFormat.Format1bppIndexed:
-							case PixelFormat.Format4bppIndexed:
-							case PixelFormat.Format8bppIndexed:
+						case PixelFormat.Format1bppIndexed:
+						case PixelFormat.Format4bppIndexed:
+						case PixelFormat.Format8bppIndexed:
 							imagePixelFormat = PixelFormat.Format32bppArgb;
-								break;
+							break;
 					}
 					// OTOH, always using 32-bit format for .png files keeps us from having problems in BloomReader
 					// like BL-5740 (where 24bit format files came out in BR with black backgrounds).
--- a/src/BloomExe/Book/PublishSettings.cs
+++ b/src/BloomExe/Book/PublishSettings.cs
@@ -50,10 +50,14 @@ namespace Bloom.Book
 		}
 		public void LoadNewJson(string json)
 		{
-			JsonSerializerSettings settings = new JsonSerializerSettings();
 			try
 			{
-				JsonConvert.PopulateObject(json, this, settings);
+				JsonConvert.PopulateObject(json, this,
+					// Previously, various things could be null. As part of simplifying the use of PublishSettings,
+					// we now never have nulls; everything gets defaults when it is created.
+					// For backwards capabilty, if the json we are reading has a null for a value,
+					// do not override the default value that we already have loaded.
+					new JsonSerializerSettings() { NullValueHandling=NullValueHandling.Ignore});
 			}
 			catch (Exception e) { throw new ApplicationException("publish-settings of this book may be corrupt", e); }
 		}
--- a/src/BloomExe/Book/PublishSettings.cs
+++ b/src/BloomExe/Book/PublishSettings.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.ComponentModel;
 using System.IO;
 using System.Linq;
 using System.Text;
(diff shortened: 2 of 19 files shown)


JohnThomson (changes requested):
Looks basically good. A few questions and minor suggestions/corrections.

> Reviewed 19 of 19 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5677#-:-NPYEuNH87tTe0oMBsZc:b-ie3hmn)* status: all files reviewed, 8 unresolved discussions (waiting on @hatton)

___
*[`src/BloomExe/Book/PublishSettings.cs` line 187 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5677#-NPXuCK08ihFLpQg8LfF:-NPXuCK08ihFLpQg8LfG:b-z4wnfs) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/0bc8ab94dd84cedf5fff79152239381f3a1012d5/src/BloomExe/Book/PublishSettings.cs#L187)):*
> ```Smalltalk
> 			{
> 				result = FromString(RobustFile.ReadAllText(path, Encoding.UTF8));
> 				result.AudioVideo.FixEmptyValues();
> ```

Can we be sure that a file (especially if written before these changes) will never have any empty values we need to fix? (Will PopulateObject leave fields at their constructor values or make them empty, if not in the file? What will happen if at some point we add a field, so it won't be in any existing json?)
___
*[`src/BloomExe/Book/PublishSettings.cs` line 357 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5677#-NPXv9HX7ff0gVYWD8mt:-NPXv9HX7ff0gVYWD8mu:b-j8j4f7) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/5600360865307a9f4212b2a85deeee1614b7f61e/src/BloomExe/Book/PublishSettings.cs#L357)):*
> ```Smalltalk
> 		{
> 			Mode = "fixed";
> 			RemoveFontSizes = false;
> ```

Redundant, but perhaps useful so it's clear all members are initialized?
___
*[`src/BloomExe/Publish/Android/AndroidPublishSettings.cs` line 135 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5677#-NPXwdS7AFnS0w04txY0:-NPXwdS7AFnS0w04txY1:brprk5q) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/5600360865307a9f4212b2a85deeee1614b7f61e/src/BloomExe/Publish/Android/AndroidPublishSettings.cs#L135)):*
> ```Smalltalk
> 
> 			HashSet<string> audioLanguagesToExclude;
> 			if (bloomPubAudioLangs.Count == 0 && libraryAudioLangs.Count == 0)
> ```

I haven't tracked down exactly how these variables are used, so I'll just ask: are you sure they won't both be zero if the user has explicitly chosen not to publish...or not to record at all... any audio? Because I think we were using null to detect that the user had not yet decided.
___
*[`src/BloomExe/Publish/Android/AndroidPublishSettings.cs` line 163 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5677#-NPXwN4g-DgYsBYZCh4x:-NPXwN4g-DgYsBYZCh4y:blgcygy) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/5600360865307a9f4212b2a85deeee1614b7f61e/src/BloomExe/Publish/Android/AndroidPublishSettings.cs#L163)):*
> ```Smalltalk
> 				// All the paths that use this are making settings for BloomPub, not Video.
> 				Motion = bookInfo.PublishSettings.BloomPub.Motion,
> 				ImagePublishSettings = bookInfo.PublishSettings.BloomPub.ImageSettings
> ```

You're quite sure we no longer need a clone?
___
*[`src/BloomExe/Publish/Android/AndroidPublishSettings.cs` line 171 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5677#-NPYAZ9u4Sx8W-8foDP0:-NPYAZ9vEt6SgBSCit2e:bctk74p) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/5600360865307a9f4212b2a85deeee1614b7f61e/src/BloomExe/Publish/Android/AndroidPublishSettings.cs#L171)):*
> ```Smalltalk
> 			// Normally this is setup by the Publish screen, but if you've never visited the Publish screen for this book,
> 			// then this will be empty. In that case, initialize it here.
> 			if (bookInfo.PublishSettings.BloomPub.TextLangs.Count == 0) // Review Feb 2022: previously it would be null, now count==0. Not exactly the same, since you could say "no text languages"?
> ```

This comment should be resolved and removed. But I don't see that it makes much sense to make a BloomPub with no text languages...unless I suppose it's a picture-only book, or the only language is a sign language? But probably in those cases there would be no text languages for InitializeLanguagesInBook() to find, so it would be harmless to call it again?In any case if someone complains that Bloom is forgetting they told it not to publish any languages, we could revisit and think about another way to store that state?
___
*[`src/BloomExe/Publish/Android/PublishToAndroidApi.cs` line 453 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5677#-NPYC7w86ufCNRnoTh9v:-NPYC7w93HmK86MDDpS1:b-6ijbb4) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/5600360865307a9f4212b2a85deeee1614b7f61e/src/BloomExe/Publish/Android/PublishToAndroidApi.cs#L453)):*
> ```Smalltalk
> 
> 			// Initialize the Talking Book Languages settings
> 			if (bookInfo.PublishSettings.BloomPub.AudioLangs.Count == 0)
> ```

Not a new problem, but as I was looking at this it occurred to me that if the book has a new language since we last did this, it probably needs to be added to AudioLangs. We could plausibly not do a test for zero languages at all, but instead, add a default for any langCode in allLangCodes that doesn't already have a setting.
___
*[`src/BloomExe/web/controllers/BookSettingsApi.cs` line 59 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5677#-NPYEZ018sb0D_2Z5e1q:-NPYEZ018sb0D_2Z5e1r:b-zeh2oy) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/5600360865307a9f4212b2a85deeee1614b7f61e/src/BloomExe/web/controllers/BookSettingsApi.cs#L59)):*
> ```Smalltalk
> 					var c = newSettings.appearance.cover.coverColor;
> 					_bookSelection.CurrentSelection.SetCoverColor(c.ToString());
> 					// review: crazy bit here, that above I'm taking json, parsing it into and object, and grabbing part of it. But then
> ```

and => an
___
*[`src/BloomExe/web/controllers/BookSettingsApi.cs` line 61 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5677#-NPYEdvcBflzuNMDErqa:-NPYEdvd1QLQg4EmuJe2:by31t9e) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/5600360865307a9f4212b2a85deeee1614b7f61e/src/BloomExe/web/controllers/BookSettingsApi.cs#L61)):*
> ```Smalltalk
> 					// review: crazy bit here, that above I'm taking json, parsing it into and object, and grabbing part of it. But then
> 					// here we take it back to json and pass it to this thing that is going to parse it agian. In this case, speed
> 					// is irrelevant. The nice thing is, it retains the identity of PublishSettings in case someone is holing on it.
> ```

holing => holding

hatton (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5677)* status: all files reviewed, 8 unresolved discussions (waiting on @JohnThomson)

___
*[`src/BloomExe/Book/PublishSettings.cs` line 187 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5677#-NPXuCK08ihFLpQg8LfF:-NPy4PYV3y1YapXjyTM1:b-ap61fb) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/0bc8ab94dd84cedf5fff79152239381f3a1012d5/src/BloomExe/Book/PublishSettings.cs#L187)):*

Can we be sure that a file (especially if written before these changes) will never have any empty values we need to fix? (Will PopulateObject leave fields at their constructor values or make them empty, if not in the file? What will happen if at some point we add a field, so it won't be in any existing json?)
</blockquote></details>

This was indeed wrong in the PR. I had a commit which does fix this, but was left behind in the BookSettings branch I'm working on. The settings now prevent the importer from overwriting a default value with a NULL one.
___
*[`src/BloomExe/Book/PublishSettings.cs` line 357 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5677#-NPXv9HX7ff0gVYWD8mt:-NPy4zoKCcBqJsfApv5_:b2nhtan) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/5600360865307a9f4212b2a85deeee1614b7f61e/src/BloomExe/Book/PublishSettings.cs#L357)):*

Redundant, but perhaps useful so it's clear all members are initialized?
</blockquote></details>

Yup
___
*[`src/BloomExe/Publish/Android/AndroidPublishSettings.cs` line 135 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5677#-NPXwdS7AFnS0w04txY0:-NPy581J3Cukbyl5N1-f:b-r2w3xd) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/5600360865307a9f4212b2a85deeee1614b7f61e/src/BloomExe/Publish/Android/AndroidPublishSettings.cs#L135)):*

I haven't tracked down exactly how these variables are used, so I'll just ask: are you sure they won't both be zero if the user has explicitly chosen not to publish...or not to record at all... any audio? Because I think we were using null to detect that the user had not yet decided.
</blockquote></details>

Yeah there could conceivably be a problem; all unit tests pass and my own manual tests seem good, so I'm expecting that with Colin's testing, we'll have confidence.
___
*[`src/BloomExe/Publish/Android/AndroidPublishSettings.cs` line 163 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5677#-NPXwN4g-DgYsBYZCh4x:-NPy5Pxc9ocrYuW-UJzs:b4ug56u) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/5600360865307a9f4212b2a85deeee1614b7f61e/src/BloomExe/Publish/Android/AndroidPublishSettings.cs#L163)):*

You're quite sure we no longer need a clone?
</blockquote></details>

There were no comments explaining why we should want one, and I know of no reason.  I think it's inherent in tackling "This code is too complicated to follow" (rather than ignore it) that we do take on some risk. But the newly simplified code can then be fixed and the previously obfuscated requirement made clear.
___
*[`src/BloomExe/Publish/Android/AndroidPublishSettings.cs` line 171 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5677#-NPYAZ9u4Sx8W-8foDP0:-NPy6frr3NBiLStIZF0_:bc9lhfg) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/5600360865307a9f4212b2a85deeee1614b7f61e/src/BloomExe/Publish/Android/AndroidPublishSettings.cs#L171)):*

This comment should be resolved and removed. But I don't see that it makes much sense to make a BloomPub with no text languages...unless I suppose it's a picture-only book, or the only language is a sign language? But probably in those cases there would be no text languages for InitializeLanguagesInBook() to find, so it would be harmless to call it again?In any case if someone complains that Bloom is forgetting they told it not to publish any languages, we could revisit and think about another way to store that state?
</blockquote></details>

The no-languages choice does get retained. We are not longer using the "NULL" status of an array to indicate user intent. Instead, we're using the trinary state of each language entry.
___
*[`src/BloomExe/Publish/Android/PublishToAndroidApi.cs` line 453 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5677#-NPYC7w86ufCNRnoTh9v:-NPy7LvbCMJ-VQhfKkw3:bo60fvu) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/5600360865307a9f4212b2a85deeee1614b7f61e/src/BloomExe/Publish/Android/PublishToAndroidApi.cs#L453)):*

Not a new problem, but as I was looking at this it occurred to me that if the book has a new language since we last did this, it probably needs to be added to AudioLangs. We could plausibly not do a test for zero languages at all, but instead, add a default for any langCode in allLangCodes that doesn't already have a setting.
</blockquote></details>

Seems to be working correctly.
___
*[`src/BloomExe/web/controllers/BookSettingsApi.cs` line 59 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5677#-NPYEZ018sb0D_2Z5e1q:-NPy7V4PD-gdoEQwE8kr:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/5600360865307a9f4212b2a85deeee1614b7f61e/src/BloomExe/web/controllers/BookSettingsApi.cs#L59)):*

and => an
</blockquote></details>

Done.
___
*[`src/BloomExe/web/controllers/BookSettingsApi.cs` line 61 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5677#-NPYEdvcBflzuNMDErqa:-NPy7a-A6FGVC9TtM3Re:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/5600360865307a9f4212b2a85deeee1614b7f61e/src/BloomExe/web/controllers/BookSettingsApi.cs#L61)):*

holing => holding
</blockquote></details>

Done.

JohnThomson (approved):
> Reviewed 9 of 9 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5677)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @hatton)

PR #5690, opened 2023-03-07
BL-11303 better dialog solution

gmartin7:
* instead of NotifyUserOfProblem, use BloomMessageBox
   which doesn't claim that Bloom had a problem when it
   didn't

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5690)

### DIFF
--- a/src/BloomBrowserUI/utils/BloomMessageBox.tsx
+++ b/src/BloomBrowserUI/utils/BloomMessageBox.tsx
@@ -6,6 +6,7 @@ import { WireUpForWinforms } from "../utils/WireUpWinform";
 import BloomButton from "../react_components/bloomButton";
 import { postString } from "./bloomApi";
 import WarningOutlinedIcon from "@mui/icons-material/WarningOutlined";
+import InfoIcon from "@mui/icons-material/Info";
 import {
     BloomDialog,
     DialogBottomButtons,
@@ -18,6 +19,7 @@ import {
     useSetupBloomDialog
 } from "../react_components/BloomDialog/BloomDialogPlumbing";
 import HtmlHelpLink from "../react_components/htmlHelpLink";
+import { kBloomBlue } from "../bloomMaterialUITheme";
 
 export interface IMessageBoxButton {
     text: string;
@@ -39,7 +41,9 @@ export const BloomMessageBox: React.FunctionComponent<{
     // If defined, creates a single "Learn More" button on the left side.
     // This is intended to look for a file whose source is "BloomBrowserUI/help/{helpButtonFileId}-en.md".
     helpButtonFileId?: string;
-    icon?: "warning" | undefined; // Effectively an enumeration, which we will add to as needed
+    // Effectively an enumeration, which we will add to as needed
+    // Interestingly enough, (MessageBoxIcon.Information).toString() in C# comes out as "Asterisk"!
+    icon?: "warning" | "asterisk" | undefined;
     dialogEnvironment?: IBloomDialogEnvironmentParams;
 
     // If defined (and true) the request came from C# and the response is via an api call.
@@ -96,6 +100,31 @@ export const BloomMessageBox: React.FunctionComponent<{
         undefined
     );
 
+    const icon = (): JSX.Element | undefined => {
+        switch (props.icon) {
+            case "warning":
+                return (
+                    <WarningOutlinedIcon
+                        css={css`
+                            font-size: 3rem !important;
+                            color: orange;
+                        `}
+                    />
+                );
+            case "asterisk":
+                return (
+                    <InfoIcon
+                        css={css`
+                            font-size: 3rem !important;
+                            color: ${kBloomBlue};
+                        `}
+                    />
+                );
+            default:
+                return undefined;
+        }
+    };
+
     return (
         <BloomDialog {...propsForBloomDialog}>
             {/* We need a element title to make things space correctly because BloomDialog expects to have one, even though at the moment we aren't including a title */}
@@ -107,15 +136,8 @@ export const BloomMessageBox: React.FunctionComponent<{
                         display: flex;
                     `}
                 >
+                    {icon()}
                     {/* InnerHTML is used so that we can insert markup like <br> into the message. */}
-                    {props.icon === "warning" && (
-                        <WarningOutlinedIcon
-                            css={css`
-                                font-size: 3rem !important;
-                                color: orange;
-                            `}
-                        />
-                    )}
                     <div
                         css={css`
                             -moz-user-select: text; // Firefox before v69
--- a/src/BloomExe/CollectionTab/CollectionModel.cs
+++ b/src/BloomExe/CollectionTab/CollectionModel.cs
@@ -343,7 +343,7 @@ namespace Bloom.CollectionTab
 					{
 						var msg = LocalizationManager.GetString("TeamCollection.CheckOutForDelete",
 							"Please check out the book before deleting it.");
-						ErrorReport.NotifyUserOfProblem(msg);
+						BloomMessageBox.ShowInfo(msg);
 						return false;
 					}
 
@@ -351,7 +351,7 @@ namespace Bloom.CollectionTab
 					{
 						var msg = LocalizationManager.GetString("TeamCollection.ConnectForDelete",
 							"Please connect to the Team Collection before deleting books that are part of it.");
-						ErrorReport.NotifyUserOfProblem(msg);
+						BloomMessageBox.ShowInfo(msg);
 						return false;
 					}
 				}
--- a/src/BloomExe/MiscUI/BloomMessageBox.cs
+++ b/src/BloomExe/MiscUI/BloomMessageBox.cs
@@ -1,4 +1,5 @@
 ﻿using System.Windows.Forms;
+using L10NSharp;
 using Newtonsoft.Json;
 
 namespace Bloom.MiscUI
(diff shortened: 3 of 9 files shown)


andrew-polk (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5690)* status: all files reviewed, 2 unresolved discussions (waiting on @gmartin7)

___
*[`src/BloomExe/Program.cs` line 322 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5690#-NPxx1J92PdHWpDkzk3k:-NPxx1J92PdHWpDkzk3l:bfo24gc) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/9f9d81fd433f0f8d1348bcd1b7a304ea5d639430/src/BloomExe/Program.cs#L322)):*
> ```Smalltalk
> 										// BloomMessageBox.Show() expects a window as 'owner', but we don't have our own window
> 										// open yet. This will pop up the dialog in the upper left corner. Not ideal, but the
> 										// best we can do in this case.
> ```

Does Shell.GetShellOrOtherOpenForm() work here?
If not, I'm pretty sure there is a way to tell winform dialogs to open in the middle of their screen.
___
*[`src/BloomExe/Workspace/WorkspaceView.cs` line 835 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5690#-NPxxYP3DY7hVinbpz_S:-NPxxYP42vPU8eeZlpk9:b-yvkly6) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/9f9d81fd433f0f8d1348bcd1b7a304ea5d639430/src/BloomExe/Workspace/WorkspaceView.cs#L835)):*
> ```Smalltalk
> 						};
> 						var formToInvokeOn = Application.OpenForms.Cast<Form>().FirstOrDefault(f => f is Shell);
> 						formToInvokeOn.Invoke((Action)(() =>
> ```

Why do we need to invoke here? (I'm pretty sure we don't because we're in the else of a `if (InvokeRequired)`)

I think the form we want as owner is  Shell.GetShellOrOtherOpenForm()

gmartin7 (commented):
Made another method of `BloomMessageBox` called `ShowInfo()` that only takes a message string. Greatly simplifies things...

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5690#-:-NQ7XErf0kDcJIaKkE9q:b-66ti4j)* status: 0 of 8 files reviewed, 2 unresolved discussions (waiting on @andrew-polk)

___
*[`src/BloomExe/Program.cs` line 322 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5690#-NPxx1J92PdHWpDkzk3k:-NPy9W7E7LiB8_8EJRv7:brbqwo2) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/9f9d81fd433f0f8d1348bcd1b7a304ea5d639430/src/BloomExe/Program.cs#L322)):*

Does Shell.GetShellOrOtherOpenForm() work here?
If not, I'm pretty sure there is a way to tell winform dialogs to open in the middle of their screen.
</blockquote></details>

Nope and nope. I tried `Shell.GetShellOrOtherOpenForm()` and it still returned null. I tried telling `ReactDialog.cs` that if ShowDialog's 'owner' param was null, it should do `dlg.CenterFormInScreen()` and that didn't work either. :(
___
*[`src/BloomExe/Workspace/WorkspaceView.cs` line 835 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5690#-NPxxYP3DY7hVinbpz_S:-NPyAwCvCAlIzRHZ329h:b-t4gwqs) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/9f9d81fd433f0f8d1348bcd1b7a304ea5d639430/src/BloomExe/Workspace/WorkspaceView.cs#L835)):*

Why do we need to invoke here? (I'm pretty sure we don't because we're in the else of a `if (InvokeRequired)`)

I think the form we want as owner is  Shell.GetShellOrOtherOpenForm() 
</blockquote></details>

OK, right. I was just copying... :D

gmartin7 (commented):
Added Information as a new and better icon type.

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5690#-:-NQCQuF22asjMpgjgqm8:b-bdpfog)* status: 0 of 9 files reviewed, 2 unresolved discussions (waiting on @andrew-polk)

andrew-polk (changes requested):
> Reviewed 9 of 9 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5690)* status: all files reviewed, 2 unresolved discussions (waiting on @gmartin7)

___
*[`src/BloomBrowserUI/utils/BloomMessageBox.tsx` line 110 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5690#-NQQUIet8Fsh4mffhUs7:-NQQUIet8Fsh4mffhUs8:b-7jn1ou) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/fafaaf204242b73101e20d2681e86158adee487c/src/BloomBrowserUI/utils/BloomMessageBox.tsx#L110)):*
> ```XML
>                         css={css`
>                             font-size: 3rem !important;
>                             color: ${kBloomYellow};
> ```

Does this color change need UI Review?
___
*[`src/BloomExe/Program.cs` line 322 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5690#-NPxx1J92PdHWpDkzk3k:-NQQQDXB6eyo-LccWiw-:bar2etl) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/9f9d81fd433f0f8d1348bcd1b7a304ea5d639430/src/BloomExe/Program.cs#L322)):*

Nope and nope. I tried `Shell.GetShellOrOtherOpenForm()` and it still returned null. I tried telling `ReactDialog.cs` that if ShowDialog's 'owner' param was null, it should do `dlg.CenterFormInScreen()` and that didn't work either. :(
</blockquote></details>

There's Form.CenterToScreen().
For some reason, the [doc](https://learn.microsoft.com/en-us/dotnet/api/system.windows.forms.form.centertoscreen?redirectedfrom=MSDN&view=windowsdesktop-7.0#System_Windows_Forms_Form_CenterToScreen) says not to call this but instead set the form's StartPosition to to CenterScreen.
So I guess try the latter first, then the former if it doesn't work.

Also, should this be ShowInfo now?

andrew-polk (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5690)* status: all files reviewed, 2 unresolved discussions (waiting on @gmartin7)

___
*[`src/BloomExe/Program.cs` line 322 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5690#-NPxx1J92PdHWpDkzk3k:-NQQeBEJA3qiSuetiOan:b-t1ig9b) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/9f9d81fd433f0f8d1348bcd1b7a304ea5d639430/src/BloomExe/Program.cs#L322)):*

There's Form.CenterToScreen().
For some reason, the [doc](https://learn.microsoft.com/en-us/dotnet/api/system.windows.forms.form.centertoscreen?redirectedfrom=MSDN&view=windowsdesktop-7.0#System_Windows_Forms_Form_CenterToScreen) says not to call this but instead set the form's StartPosition to to CenterScreen.
So I guess try the latter first, then the former if it doesn't work.

Also, should this be ShowInfo now?
</blockquote></details>

I got info for centering from https://stackoverflow.com/questions/4601827/how-do-i-center-a-window-on-the-screen-in-c-sharp-winforms

gmartin7 (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5690)* status: all files reviewed, 2 unresolved discussions (waiting on @andrew-polk)

___
*[`src/BloomBrowserUI/utils/BloomMessageBox.tsx` line 110 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5690#-NQQUIet8Fsh4mffhUs7:-NQQq3_T-3pGF25YUuud:b-63en7b) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/fafaaf204242b73101e20d2681e86158adee487c/src/BloomBrowserUI/utils/BloomMessageBox.tsx#L110)):*

Does this color change need UI Review?
</blockquote></details>

Hmm... I guess I thought I'd replaced all the warnings with information so it didn't matter, but there are still a couple of them in our code... I'm guessing JH would say change 'em all to information, but it would be simpler to revert my change here. In my defense, the difference is between #FFA500 and #FEBF00... fairly close.
___
*[`src/BloomExe/Program.cs` line 322 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5690#-NPxx1J92PdHWpDkzk3k:-NQQxuVEAgzV_UPrrwXN:bryz7kn) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/9f9d81fd433f0f8d1348bcd1b7a304ea5d639430/src/BloomExe/Program.cs#L322)):*

I got info for centering from https://stackoverflow.com/questions/4601827/how-do-i-center-a-window-on-the-screen-in-c-sharp-winforms
</blockquote></details>

I hadn't changed this one to `ShowInfo()` because of the screen positioning issue, but now that it's centering on the screen when owner is null, I switched to `ShowInfo()`.
Dunno how come I didn't find that StartPosition option, I'm sure I've used it before. :(

andrew-polk (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5690)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @gmartin7)

PR #5696, opened 2023-03-13
Enable uploadCollisionDlg for new upload screen (BL-11581)

andrew-polk:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5696)

### DIFF
--- a/src/BloomBrowserUI/publish/LibraryPublish/LibraryPublishSteps.tsx
+++ b/src/BloomBrowserUI/publish/LibraryPublish/LibraryPublishSteps.tsx
@@ -35,6 +35,11 @@ import {
 } from "../../react_components/confirmDialog";
 import { BloomSplitButton } from "../../react_components/bloomSplitButton";
 import { WaitBox } from "../../react_components/BloomDialog/commonDialogComponents";
+import {
+    IUploadCollisionDlgProps,
+    showUploadCollisionDialog,
+    UploadCollisionDlg
+} from "./uploadCollisionDlg";
 
 interface IReadonlyBookInfo {
     title: string;
@@ -140,11 +145,27 @@ export const LibraryPublishSteps: React.FunctionComponent = () => {
         }
     }
 
+    const [uploadCollisionInfo, setUploadCollisionInfo] = useState<
+        IUploadCollisionDlgProps
+    >({
+        userEmail: "",
+        newTitle: "",
+        existingTitle: "",
+        existingCreatedDate: "",
+        existingUpdatedDate: "",
+        existingBookUrl: ""
+    });
+
     const [isUploading, setIsUploading] = useState<boolean>(false);
     function uploadOneBook() {
         setIsUploadComplete(false);
         setIsUploading(true);
-        post("libraryPublish/upload");
+        get("libraryPublish/getUploadCollisionInfo", result => {
+            if (result.data.shouldShow) {
+                setUploadCollisionInfo(result.data);
+                showUploadCollisionDialog();
+            } else post("libraryPublish/upload");
+        });
     }
 
     function bulkUploadCollection() {
@@ -544,6 +565,12 @@ export const LibraryPublishSteps: React.FunctionComponent = () => {
                 cancelButtonLabelL10nKey="Common.No"
                 onDialogClose={function(result: DialogResult): void {
                     if (result === DialogResult.Confirm) uploadOneBook();
+                }}
+            />
+            <UploadCollisionDlg
+                {...uploadCollisionInfo}
+                onCancel={() => {
+                    setIsUploading(false);
                 }}
             />
         </React.Fragment>
--- a/src/BloomBrowserUI/publish/LibraryPublish/uploadCollisionDlg.tsx
+++ b/src/BloomBrowserUI/publish/LibraryPublish/uploadCollisionDlg.tsx
@@ -22,7 +22,6 @@ import { TextWithEmbeddedLink } from "../../react_components/link";
 import { post, postJson } from "../../utils/bloomApi";
 import { WireUpForWinforms } from "../../utils/WireUpWinform";
 import { lightTheme } from "../../bloomMaterialUITheme";
-import { ThemeProvider, StyledEngineProvider } from "@mui/material/styles";
 import HelpLink from "../../react_components/helpLink";
 import {
     IBloomDialogEnvironmentParams,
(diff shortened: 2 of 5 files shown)


JohnThomson (approved):
Looks basically good. A few take-or-leave minor suggestions. I wish we didn't have so much new code just to support the obsolete control, but I suppose there's no good alternative until the new system can really replace it. Marking it approved because I don't think any of my comments MUST be acted on.

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5696#-:-NQVOIrt0qyJnmDtF4HH:b-4jq6iu)* status: all files reviewed, 4 unresolved discussions (waiting on @andrew-polk)

___
*[`src/BloomBrowserUI/publish/LibraryPublish/uploadCollisionDlg.tsx` line 48 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5696#-NQVMNRy5uqCrw4Ci-Ou:-NQVMNRy5uqCrw4Ci-Ov:bswjufe) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/8ec21649116dfc9cc2caffd4bf8891fba7f0ace1/src/BloomBrowserUI/publish/LibraryPublish/uploadCollisionDlg.tsx#L48)):*
> ```XML
> let legacyMode: boolean = true;
> export function showUploadCollisionDialog() {
>     // This is a horrible hack we have in place until the old upload screen dies.
> ```

I'm not quite clear what the horrible hack is. I gather that this function is only called when the dialog is opened by the new upload screen, so when it is opened in its own Windows.Forms dialog legacyMode stays false? That might be worth saying in the comment.
___
*[`src/BloomBrowserUI/publish/LibraryPublish/uploadCollisionDlg.tsx` line 225 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5696#-NQVGzf37yt1HsBnWlao:-NQVGzf37yt1HsBnWlap:b-lfciat) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/8ec21649116dfc9cc2caffd4bf8891fba7f0ace1/src/BloomBrowserUI/publish/LibraryPublish/uploadCollisionDlg.tsx#L225)):*
> ```XML
> 
>     function cancel() {
>         if (props.onCancel) props.onCancel();
> ```

I'm not sure this is better, but I believe you can just do props.onCancel?.()
___
*[`src/BloomBrowserUI/publish/LibraryPublish/uploadCollisionDlg.tsx` line 309 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5696#-NQVHM9W3yVVM5rQw9a8:-NQVHM9W3yVVM5rQw9a9:b-cfxrnu) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/8ec21649116dfc9cc2caffd4bf8891fba7f0ace1/src/BloomBrowserUI/publish/LibraryPublish/uploadCollisionDlg.tsx#L309)):*
> ```XML
>                             thumbnailUrl={props.newThumbUrl}
>                             languages={
>                                 props.newLanguages ? props.newLanguages : [""]
> ```

maybe props.newLanguages ?? ""
___
*[`src/BloomExe/web/controllers/LibraryPublishApi.cs` line 306 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5696#-NQVJRy08UN7aFXJel_a:-NQVJRy08UN7aFXJel_b:bisrol3) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/8ec21649116dfc9cc2caffd4bf8891fba7f0ace1/src/BloomExe/web/controllers/LibraryPublishApi.cs#L306)):*
> ```Smalltalk
> 				{
> 					shouldShow = Model.BookIsAlreadyOnServer,
> 					// I can't find any way to do this automatically; pretty frustrating...
> ```

If I'm understanding correctly, this code is creating a dynamic object with basically all the same fields as the one returned by the function, and the problem is to add the shouldShow field without having to know about all the others here?
But is there any caller of GetUploadCollisionDialogProps that doesn't want shouldShow to be true? If not, you could just have it set that as well.  If there is, perhaps it could be a parameter of the method to say whether it should be true or false? But I'm pretty sure you always want it true (except for the otherwise empty object above).
I think another option with this sort of object is that once the original creator sets shouldShow to something, that is one of its defined properties, so a client could change it.

andrew-polk (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5696)* status: all files reviewed, 4 unresolved discussions (waiting on @JohnThomson)

___
*[`src/BloomBrowserUI/publish/LibraryPublish/uploadCollisionDlg.tsx` line 48 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5696#-NQVMNRy5uqCrw4Ci-Ou:-NQW3CpgDa4axv6a4X5r:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/8ec21649116dfc9cc2caffd4bf8891fba7f0ace1/src/BloomBrowserUI/publish/LibraryPublish/uploadCollisionDlg.tsx#L48)):*

I'm not quite clear what the horrible hack is. I gather that this function is only called when the dialog is opened by the new upload screen, so when it is opened in its own Windows.Forms dialog legacyMode stays false? That might be worth saying in the comment.
</blockquote></details>

Done.
___
*[`src/BloomBrowserUI/publish/LibraryPublish/uploadCollisionDlg.tsx` line 225 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5696#-NQVGzf37yt1HsBnWlao:-NQW3iaj6zYIRzbwiyqT:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/8ec21649116dfc9cc2caffd4bf8891fba7f0ace1/src/BloomBrowserUI/publish/LibraryPublish/uploadCollisionDlg.tsx#L225)):*

I'm not sure this is better, but I believe you can just do props.onCancel?.()
</blockquote></details>

Done.
___
*[`src/BloomBrowserUI/publish/LibraryPublish/uploadCollisionDlg.tsx` line 309 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5696#-NQVHM9W3yVVM5rQw9a8:-NQW3uIo41YLYpmSBX-3:bi1hhox) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/8ec21649116dfc9cc2caffd4bf8891fba7f0ace1/src/BloomBrowserUI/publish/LibraryPublish/uploadCollisionDlg.tsx#L309)):*

maybe props.newLanguages ?? ""
</blockquote></details>

Done.
(wasn't from this change : ) )
___
*[`src/BloomExe/web/controllers/LibraryPublishApi.cs` line 306 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5696#-NQVJRy08UN7aFXJel_a:-NQW6ZUq9hGKLPWxV2jr:beok988) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/8ec21649116dfc9cc2caffd4bf8891fba7f0ace1/src/BloomExe/web/controllers/LibraryPublishApi.cs#L306)):*

If I'm understanding correctly, this code is creating a dynamic object with basically all the same fields as the one returned by the function, and the problem is to add the shouldShow field without having to know about all the others here?
But is there any caller of GetUploadCollisionDialogProps that doesn't want shouldShow to be true? If not, you could just have it set that as well.  If there is, perhaps it could be a parameter of the method to say whether it should be true or false? But I'm pretty sure you always want it true (except for the otherwise empty object above).
I think another option with this sort of object is that once the original creator sets shouldShow to something, that is one of its defined properties, so a client could change it.

</blockquote></details>

Yeah, I was trying to avoid setting that prop when the other code path (the old control) doesn't use it. But no sense in complicating things for a control which is going away.
It doesn't hurt anything to set it, apparently.

JohnThomson (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5696)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @andrew-polk)

PR #5700, opened 2023-03-15
Support video and widgets in spreadsheet (BL-11542)

JohnThomson:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5700)

### DIFF
--- a/DistFiles/localization/en/BloomLowPriority.xlf
+++ b/DistFiles/localization/en/BloomLowPriority.xlf
@@ -141,10 +141,22 @@
                 <source xml:lang="en">The folder named {0} already exists and looks like it might be a Bloom book folder!</source>
                 <note>ID: Spreadsheet.OverwriteBook</note>
             </trans-unit>
-            <trans-unit id="Spreadsheet.WriteFailed">
-                <source xml:lang="en">Bloom had problems writing files to that location ({0}). Check that you have permission to write there.</source>
-                <note>ID: Spreadsheet.WriteFailed</note>
+          <!-- 5.5? But we may not want to translate any Spreadsheet stuff yet? Anyway, this is new in 5.5-->
+          <trans-unit id="Spreadsheet.TroubleCopying" translate="no">
+            <source xml:lang="en">Bloom had trouble copying the file {0} to the video folder: {1}</source>
+            <note>ID: Spreadsheet.TroubleCopying</note>
+            <note>{1} is a more detailed, usually technical, error message and should be at the end if possible.</note>
+          </trans-unit>
+            <trans-unit id="Spreadsheet.TroubleCopyingFolder" translate="no">
+                <source xml:lang="en">Bloom had trouble copying the folder {0} to the activities folder: {1}</source>
+                <note>ID: Spreadsheet.TroubleCopyingFolder</note>
+              <note>{1} is a more detailed, usually technical, error message and should be at the end if possible.</note>
             </trans-unit>
+
+          <trans-unit id="Spreadsheet.WriteFailed">
+            <source xml:lang="en">Bloom had problems writing files to that location ({0}). Check that you have permission to write there.</source>
+            <note>ID: Spreadsheet.WriteFailed</note>
+          </trans-unit>
             <trans-unit id="Activity.Editing.MultipleChoice.CorrectAnswer">
                 <source xml:lang="en">Put the correct answer in this button. Bloom will randomize the order later.</source>
             </trans-unit>
--- a/src/BloomBrowserUI/collectionsTab/spreadsheet/SpreadsheetExportDialog.tsx
+++ b/src/BloomBrowserUI/collectionsTab/spreadsheet/SpreadsheetExportDialog.tsx
@@ -76,13 +76,10 @@ const SpreadsheetExportDialog: React.FunctionComponent<{
                 <p></p>
                 <WarningBox>
                     <span>
-                        This feature is still a work in progress. Though you can{" "}
-                        <em>export</em> all books, Bloom cannot <em>import</em>{" "}
-                        books with these features: Quizzes and other Activities.
-                        If you change the text after recording audio and before
-                        exporting, the audio may not import correctly. There may
-                        be other export and import limitations we are not aware
-                        of.
+                        This feature is still a work in progress. If you change
+                        the text after recording audio and before exporting, the
+                        audio may not import correctly. There may be other
+                        export and import limitations we are not aware of.
                     </span>
                 </WarningBox>
                 <p></p>
--- a/src/BloomExe/Book/Book.cs
+++ b/src/BloomExe/Book/Book.cs
@@ -2973,16 +2973,10 @@ namespace Bloom.Book
 		/// </remarks>
 		private void CopyMissingStylesheetFiles(IPage templatePage)
 		{
-			foreach (string sheetName in templatePage.Book.OurHtmlDom.GetTemplateStyleSheets())
-			{
-				var destinationPath = Path.Combine(FolderPath, sheetName);
-				if (!RobustFile.Exists(destinationPath))
-				{
-					var sourcePath = Path.Combine(templatePage.Book.FolderPath, sheetName);
-					if (RobustFile.Exists(sourcePath))
-						RobustFile.Copy(sourcePath, destinationPath);
-				}
-			}
+			var sourceDom = templatePage.Book.OurHtmlDom;
+			var sourceFolder = templatePage.Book.FolderPath;
+			var destFolder = FolderPath;
+			HtmlDom.CopyMissingStylesheetFiles(sourceDom, sourceFolder, destFolder);
 		}
 
 		/// <summary>
@@ -3258,7 +3252,11 @@ namespace Bloom.Book
 		/// </summary>
 		internal XmlElement GetOrCreateUserModifiedStyleElementFromStorage()
 		{
-			var headElement = OurHtmlDom.Head;
+			return GetOrCreateUserModifiedStyleElementFromStorage(OurHtmlDom.Head);
+		}
+
+		public static XmlElement GetOrCreateUserModifiedStyleElementFromStorage(XmlElement headElement)
+		{
 			var userStyleElement = HtmlDom.GetUserModifiedStyleElement(headElement);
 			if (userStyleElement == null)
 				return HtmlDom.AddEmptyUserModifiedStylesNode(headElement);
--- a/src/BloomExe/Book/Book.cs
+++ b/src/BloomExe/Book/Book.cs
@@ -61,6 +61,16 @@ namespace Bloom.Book
 
 		int _audioFilesCopiedForDuplication;
 
+		public const string BasicTextAndImageGuid = "adcd48df-e9ab-4a07-afd4-6a24d0398382";
+		public const string JustPictureGuid = "adcd48df-e9ab-4a07-afd4-6a24d0398385";
+		public const string PictureOnLeftGuid = "7b192144-527c-417c-a2cb-1fb5e78bf38a";
+		public const string JustTextGuid = "a31c38d8-c1cb-4eb9-951b-d2840f6a8bdb";
+		public const string JustVideoGuid = "8bedcdf8-3ad6-4967-b027-6c186436572f";
+		public const string VideoOverTextGuid = "299644f5-addb-476f-a4a5-e3978139b188";
+		public const string PictureAndVideoGuid = "24c90e90-2711-465d-8f20-980d9ffae299";
+		public const string BigTextDiglotGuid = "08422e7b-9406-4d11-8c71-02005b1b8095";
+		public const string WidgetGuid = "3a705ac1-c1f2-45cd-8a7d-011c009cf406"; // default page type for a single widget
+
 		//for moq'ing only; parameterless ctor required by Moq
 		public Book()
 		{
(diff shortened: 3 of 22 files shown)


andrew-polk (changes requested):
> Reviewed 17 of 19 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5700)* status: 17 of 19 files reviewed, 7 unresolved discussions (waiting on @JohnThomson)

___
*[`src/BloomExe/Extensions.cs` line 183 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5700#-NR4KxKk5vjwsDvpv4PE:-NR4KxKk5vjwsDvpv4PF:b-prmx44) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/bc532a75f8e2db3cb194ab7de0a8dca27d205a67/src/BloomExe/Extensions.cs#L183)):*
> ```Smalltalk
> 		/// into items that need it.
> 		/// </summary>
> 		public static IEnumerable<T[]> MapUnevenLists<T>(List<T>[] inputs)
> ```

Is it worth testing this explicitly?
I'm ok with the answer being "no" if it is sufficiently tested through the SS tests and we don't have a reasonable expectation of it being reused.
___
*[`src/BloomExe/Book/HtmlDom.cs` line 1321 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5700#-NR4OUqp3J_4vEMabNIk:-NR4OUqp3J_4vEMabNIl:b-fm43ym) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/bc532a75f8e2db3cb194ab7de0a8dca27d205a67/src/BloomExe/Book/HtmlDom.cs#L1321)):*
> ```Smalltalk
> 		}
> 
> 		internal static string GetUserModifiableStylesUsedOnPage(XmlElement head, XmlElement contentToSearch)
> ```

This is a good cleanup.
Looking at the existing code, it appears we can stop modifying the dom here because the only caller (in Book) has a subsequent call which does basically the same thing. Is that accurate?
___
*[`src/BloomExe/Spreadsheet/SpreadsheetImporter.cs` line 1100 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5700#-NR4Ur0n14AF5B3klvOG:-NR4Ur0n14AF5B3klvOH:b-xmt6c1) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/bc532a75f8e2db3cb194ab7de0a8dca27d205a67/src/BloomExe/Spreadsheet/SpreadsheetImporter.cs#L1100)):*
> ```Smalltalk
> 		private Dictionary<BlockTypes, string> _pagesToInsert;
> 
> 		Dictionary<BlockTypes, string> PageToInsert()
> ```

A bigger picture question...

What happens when user creates Basic Text and Picture page, uses origami to add another image, then exports?
Do we still call it Basic Text and Picture even though we know the content slots don't fit?
If we do, that seems inherently wrong.
___
*[`src/BloomExe/Spreadsheet/SpreadsheetImporter.cs` line 1105 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5700#-NR4UlTi-XbIkFNLxVZH:-NR4UlTi-XbIkFNLxVZI:bvmtuph) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/bc532a75f8e2db3cb194ab7de0a8dca27d205a67/src/BloomExe/Spreadsheet/SpreadsheetImporter.cs#L1105)):*
> ```Smalltalk
> 			{
> 				_pagesToInsert = new Dictionary<BlockTypes, string>();
> 				_pagesToInsert[BlockTypes.Image] = "adcd48df-e9ab-4a07-afd4-6a24d0398385"; // just an image
> ```

Seems like we would benefit to make these guids constants somewhere.
___
*[`src/BloomTests/Spreadsheet/SpreadsheetImporterTests.cs` line 328 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5700#-NR4Tbfp2j0jDxQDigco:-NR4Tbfp2j0jDxQDigcp:bktrkf4) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/bc532a75f8e2db3cb194ab7de0a8dca27d205a67/src/BloomTests/Spreadsheet/SpreadsheetImporterTests.cs#L328)):*
> ```Smalltalk
> 		{
> 			return String.Format(
> 				@"<div class=""bloom-videoContainer bloom-noVideoSelected bloom-leadingElement bloom-selected"" id=""vc{0}"">
> ```

I'm sure it wouldn't affect the test either way, but we don't ever have `bloom-noVideoSelected` with a `source`, do we?



_Code quote:_
```Smalltalk
bloom-noVideoSelected
```
___
*[`src/BloomTests/Spreadsheet/SpreadsheetPageTypeImportTests.cs` line 23 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5700#-NR4R3owBUTYAZ5bk2AW:-NR4R3owBUTYAZ5bk2AX:b-ms42lq) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/bc532a75f8e2db3cb194ab7de0a8dca27d205a67/src/BloomTests/Spreadsheet/SpreadsheetPageTypeImportTests.cs#L23)):*
> ```Smalltalk
> 	/// This also covers (along with SpreadsheetImageAndTextImportTests) various cases of how to
> 	/// import when page type is specified.
> 	/// The following is a list I made at the start of interesting cases. I didn't quite end up implementing all of them,
> ```

Is it worth being explicit about which ones you covered? Or at least the ones you didn't?
Otherwise, I'm not sure I see a lot of value in keeping the comment.



_Code quote:_
```Smalltalk
	/// The following is a list I made at the start of interesting cases. I didn't quite end up implementing all of them,
	/// but I thought it was worth keeping. The ones I did seem to capture all the important paths through the code.
```
___
*[`src/BloomTests/Spreadsheet/SpreadsheetPageTypeImportTests.cs` line 231 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5700#-NR4SXEY004gGcBQZUuC:-NR4SXEY004gGcBQZUuD:bkjqpge) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/bc532a75f8e2db3cb194ab7de0a8dca27d205a67/src/BloomTests/Spreadsheet/SpreadsheetPageTypeImportTests.cs#L231)):*
> ```Smalltalk
> 			_progressSpy = new ProgressSpy();
> 
> 			// Not sure we need this. Todo: try removing it when otherwise done.
> ```

Still want to do this?



_Code quote:_
```Smalltalk
Todo: try removing it when otherwise done.
```

andrew-polk (commented):
I've made a pass through everything but the two biggies... SpreadsheetExporter and SpreadsheetImporter. Maybe best to do those interactive?
Let me know if you want to do it before or after looking at the comments I've already added.

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5700#-:-NR4VQiKDIdFX2jscsRH:bvxwuvi)* status: 17 of 19 files reviewed, 7 unresolved discussions (waiting on @JohnThomson)

JohnThomson (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5700)* status: 17 of 19 files reviewed, 7 unresolved discussions (waiting on @andrew-polk)

___
*[`src/BloomExe/Extensions.cs` line 183 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5700#-NR4KxKk5vjwsDvpv4PE:-NRh_9WN4QSbvfXPK77S:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/bc532a75f8e2db3cb194ab7de0a8dca27d205a67/src/BloomExe/Extensions.cs#L183)):*

Is it worth testing this explicitly?
I'm ok with the answer being "no" if it is sufficiently tested through the SS tests and we don't have a reasonable expectation of it being reused.
</blockquote></details>

Done.
___
*[`src/BloomExe/Book/HtmlDom.cs` line 1321 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5700#-NR4OUqp3J_4vEMabNIk:-NRhaHeK3OKyN0fbKZjL:bb831pg) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/bc532a75f8e2db3cb194ab7de0a8dca27d205a67/src/BloomExe/Book/HtmlDom.cs#L1321)):*

This is a good cleanup.
Looking at the existing code, it appears we can stop modifying the dom here because the only caller (in Book) has a subsequent call which does basically the same thing. Is that accurate?
</blockquote></details>

It's been a long time. I don't recall why I removed the DOM updating, except for the obvious consideration that it's obnoxious for a Get method to have side effects. But indeed both callers do their own saving of the new stylesheet (after merging with what this returns).
___
*[`src/BloomExe/Spreadsheet/SpreadsheetImporter.cs` line 1100 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5700#-NR4Ur0n14AF5B3klvOG:-NRhawSt3Loh45OFFQmk:bk7obgt) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/bc532a75f8e2db3cb194ab7de0a8dca27d205a67/src/BloomExe/Spreadsheet/SpreadsheetImporter.cs#L1100)):*

A bigger picture question...

What happens when user creates Basic Text and Picture page, uses origami to add another image, then exports?
Do we still call it Basic Text and Picture even though we know the content slots don't fit?
If we do, that seems inherently wrong.
</blockquote></details>

The only time the name of the page matters is if we are importing into a document and the spreadsheet says it needs a page with a given name (or of course if the user is making a new template, but then it's up to them to make an appropriate name). The situation you're describing will be OK when re-importing into the same document because the spreadsheet rows that expect a 3-slot Basic Text and Picture will align with the page that has exactly that.
But it might make sense to generate a modified name...I'm just not sure what it should be. I suppose it would be something to just change it to "Basic Text & Picture modified" or we could actually indicate the number of slots..."Basic Text & Picture (1T,2P,0V,0W)" so you'd only get a match for something modified to have the expected number of each slot. Something to discuss with JohnH?
If we want to, feels like a plausible distinct card ("Update page name when changing slots") rather than making this PR even bigger.
___
*[`src/BloomExe/Spreadsheet/SpreadsheetImporter.cs` line 1105 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5700#-NR4UlTi-XbIkFNLxVZH:-NRhzpwB4p0wMDzO8Cqu:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/bc532a75f8e2db3cb194ab7de0a8dca27d205a67/src/BloomExe/Spreadsheet/SpreadsheetImporter.cs#L1105)):*

Seems like we would benefit to make these guids constants somewhere.
</blockquote></details>

Done.
___
*[`src/BloomTests/Spreadsheet/SpreadsheetImporterTests.cs` line 328 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5700#-NR4Tbfp2j0jDxQDigco:-NRi-3pk6ebVWHYaaRGi:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/bc532a75f8e2db3cb194ab7de0a8dca27d205a67/src/BloomTests/Spreadsheet/SpreadsheetImporterTests.cs#L328)):*

I'm sure it wouldn't affect the test either way, but we don't ever have `bloom-noVideoSelected` with a `source`, do we?
</blockquote></details>

Done.
___
*[`src/BloomTests/Spreadsheet/SpreadsheetPageTypeImportTests.cs` line 23 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5700#-NR4R3owBUTYAZ5bk2AW:-NRi-K7F78PsEzYoPcl5:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/bc532a75f8e2db3cb194ab7de0a8dca27d205a67/src/BloomTests/Spreadsheet/SpreadsheetPageTypeImportTests.cs#L23)):*

Is it worth being explicit about which ones you covered? Or at least the ones you didn't?
Otherwise, I'm not sure I see a lot of value in keeping the comment.
</blockquote></details>

Done.
___
*[`src/BloomTests/Spreadsheet/SpreadsheetPageTypeImportTests.cs` line 231 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5700#-NR4SXEY004gGcBQZUuC:-NRi45Qb8HmB6YbdDmkQ:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/bc532a75f8e2db3cb194ab7de0a8dca27d205a67/src/BloomTests/Spreadsheet/SpreadsheetPageTypeImportTests.cs#L231)):*

Still want to do this?
</blockquote></details>

Done.

andrew-polk (approved):
> Reviewed 1 of 19 files at r1, 7 of 7 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5700)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @JohnThomson)

___
*[`src/BloomExe/Spreadsheet/SpreadsheetImporter.cs` line 1100 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5700#-NR4Ur0n14AF5B3klvOG:-NRiOZh5Do9DDB6TyrTj:b-71v3sc) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/bc532a75f8e2db3cb194ab7de0a8dca27d205a67/src/BloomExe/Spreadsheet/SpreadsheetImporter.cs#L1100)):*

The only time the name of the page matters is if we are importing into a document and the spreadsheet says it needs a page with a given name (or of course if the user is making a new template, but then it's up to them to make an appropriate name). The situation you're describing will be OK when re-importing into the same document because the spreadsheet rows that expect a 3-slot Basic Text and Picture will align with the page that has exactly that.
But it might make sense to generate a modified name...I'm just not sure what it should be. I suppose it would be something to just change it to "Basic Text & Picture modified" or we could actually indicate the number of slots..."Basic Text & Picture (1T,2P,0V,0W)" so you'd only get a match for something modified to have the expected number of each slot. Something to discuss with JohnH?
If we want to, feels like a plausible distinct card ("Update page name when changing slots") rather than making this PR even bigger.
</blockquote></details>

Ok, we agreed for now to just point out the scenario to Colin so he can make sure it does something reasonable.

andrew-polk (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5700)* status: all files reviewed, 3 unresolved discussions (waiting on @JohnThomson)

___
*[`src/BloomExe/Spreadsheet/SpreadsheetExporter.cs` line 210 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5700#-NRjI3LB2uAoIs34hNRt:-NRjI3LB2uAoIs34hNRu:b-aswb2k) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/8d160a6b01d2af81723a8cd71127d8d33c2c842d/src/BloomExe/Spreadsheet/SpreadsheetExporter.cs#L210)):*
> ```Smalltalk
> 				{
> 					_progress.MessageWithParams("Spreadsheet.TroubleCopyingFolder", "",
> 						"Bloom had trouble copying the folder {0} to the activities folder: " +
> ```

The space at the end of the string should not be part of the i18n. Sorry I didn't catch this before.
___
*[`src/BloomExe/Spreadsheet/SpreadsheetExporter.cs` line 236 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5700#-NRjI9K75oQtNZIB7Qpi:-NRjI9K75oQtNZIB7Qpj:b-aswb2k) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/8d160a6b01d2af81723a8cd71127d8d33c2c842d/src/BloomExe/Spreadsheet/SpreadsheetExporter.cs#L236)):*
> ```Smalltalk
> 				{
> 					_progress.MessageWithParams("SpreadsheetTroubleCopying", "",
> 						"Bloom had trouble copying the file {0} to the video folder: " +
> ```

The space at the end of the string should not be part of the i18n. Sorry I didn't catch this before.
___
*[`src/BloomExe/Spreadsheet/SpreadsheetImporter.cs` line 512 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5700#-NRjHIT4C1seWbwtYZNo:-NRjHIT4C1seWbwtYZNp:bobs4ev) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/8d160a6b01d2af81723a8cd71127d8d33c2c842d/src/BloomExe/Spreadsheet/SpreadsheetImporter.cs#L512)):*
> ```Smalltalk
> 		/// and enough cycles will cause the name to overflow. So, if the name looks like
> 		/// an output (ends in hyphen-number), just increment the number.</remarks>>
> 		public static string GetNextFileName(string input)
> ```

I don't like introducing a third one of these.
(We already have ImageUtils.GetUnusedFilename() and BookStorage.GetUniqueFileName().)
Maybe ImageUtils.GetUnusedFilename() can be reused (maybe extracted first)?
If not, should we at least move this to a common place and document why it needs to be different than the other two?

JohnThomson (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5700)* status: 12 of 22 files reviewed, 3 unresolved discussions (waiting on @andrew-polk)

___
*[`src/BloomExe/Spreadsheet/SpreadsheetExporter.cs` line 210 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5700#-NRjI3LB2uAoIs34hNRt:-NRjKg_N10oYRegwkJPb:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/8d160a6b01d2af81723a8cd71127d8d33c2c842d/src/BloomExe/Spreadsheet/SpreadsheetExporter.cs#L210)):*

The space at the end of the string should not be part of the i18n. Sorry I didn't catch this before.
</blockquote></details>

Done.
___
*[`src/BloomExe/Spreadsheet/SpreadsheetExporter.cs` line 236 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5700#-NRjI9K75oQtNZIB7Qpi:-NRjKguv6GxcbKhyX3MH:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/8d160a6b01d2af81723a8cd71127d8d33c2c842d/src/BloomExe/Spreadsheet/SpreadsheetExporter.cs#L236)):*

The space at the end of the string should not be part of the i18n. Sorry I didn't catch this before.
</blockquote></details>

Done.
___
*[`src/BloomExe/Spreadsheet/SpreadsheetImporter.cs` line 512 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5700#-NRjHIT4C1seWbwtYZNo:-NRjOD9J3hTDxza0uBSi:b-z951su) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/8d160a6b01d2af81723a8cd71127d8d33c2c842d/src/BloomExe/Spreadsheet/SpreadsheetImporter.cs#L512)):*

I don't like introducing a third one of these.
(We already have ImageUtils.GetUnusedFilename() and BookStorage.GetUniqueFileName().)
Maybe ImageUtils.GetUnusedFilename() can be reused (maybe extracted first)?
If not, should we at least move this to a common place and document why it needs to be different than the other two?
</blockquote></details>

I removed it and used the ImageUtils one.

andrew-polk (approved):
> Reviewed 10 of 10 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5700)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @JohnThomson)

PR #5701, opened 2023-03-15
BL-12006 Fix missing gear

gmartin7:
* If the translation group and bloom editable don't have any x-style class
   (after migrating some old situations), assume 'normal-style', so the style
   editor gear will still show up as it should.


andrew-polk (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5701)* status: all files reviewed, 2 unresolved discussions (waiting on @gmartin7)

___
*[`src/BloomBrowserUI/bookEdit/StyleEditor/StyleEditor.ts` line 1000 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5701#-NQaMxUb2qw-Q2uEdsSs:-NQaMxUb2qw-Q2uEdsSt:bfesbe8) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/b868c2f08ee32614b586ff4d8656108e5cf73c25/src/BloomBrowserUI/bookEdit/StyleEditor/StyleEditor.ts#L1000)):*
I'm unsure of this strategy.

There is at least one page type (Translation Instructions) where we don't provide a gear to change settings. I'm not sure that was actually by design... it may have been an oversight.

But are there other scenarios where we don't want the gear? 
JT says,
> I suppose it's not the case that there are translation groups that deliberately don't have a user-modifiable style? e.g., some branding/xmatter that wants to lock the formatting?
___
*[`src/BloomBrowserUI/bookEdit/StyleEditor/StyleEditor.ts` line 1001 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5701#-NQaMpJP95uSifaOqoKI:-NQaMpJP95uSifaOqoKJ:b-v7fi2a) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/b868c2f08ee32614b586ff4d8656108e5cf73c25/src/BloomBrowserUI/bookEdit/StyleEditor/StyleEditor.ts#L1001)):*
> ```XML
>             // Even if we found a "targetBox" with no style, we don't want to dump out here, otherwise
>             // the user will have no "gear icon" to interact with. So we'll add our default normal style.
>             targetBox.className += " normal-style";
> ```

Can this be `targetBox.classList.add("normal-style")` instead?



_Code quote:_
```XML
targetBox.className += " normal-style";
```

gmartin7 (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5701)* status: all files reviewed, 2 unresolved discussions (waiting on @andrew-polk)

___
*[`src/BloomBrowserUI/bookEdit/StyleEditor/StyleEditor.ts` line 1000 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5701#-NQaMxUb2qw-Q2uEdsSs:-NQgODKfEvSWJP9MM59x:b-vnopse) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/b868c2f08ee32614b586ff4d8656108e5cf73c25/src/BloomBrowserUI/bookEdit/StyleEditor/StyleEditor.ts#L1000)):*

I'm unsure of this strategy.

There is at least one page type (Translation Instructions) where we don't provide a gear to change settings. I'm not sure that was actually by design... it may have been an oversight.

But are there other scenarios where we don't want the gear? 
JT says,
> I suppose it's not the case that there are translation groups that deliberately don't have a user-modifiable style? e.g., some branding/xmatter that wants to lock the formatting?
</blockquote></details>

I see this class on some `bloom-editables`: `bloom-nodefaultstylerule`, but it seems to occur in conjunction with `X-style` classes too. I'm not sure what to do here.
___
*[`src/BloomBrowserUI/bookEdit/StyleEditor/StyleEditor.ts` line 1001 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5701#-NQaMpJP95uSifaOqoKI:-NQfgE_y9yvZ7Ii54B8h:b-jgpg00) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/b868c2f08ee32614b586ff4d8656108e5cf73c25/src/BloomBrowserUI/bookEdit/StyleEditor/StyleEditor.ts#L1001)):*

Can this be `targetBox.classList.add("normal-style")` instead?
</blockquote></details>

Done. Silly me.

gmartin7 (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5701)* status: all files reviewed, 1 unresolved discussion (waiting on @andrew-polk)

___
*[`src/BloomBrowserUI/bookEdit/StyleEditor/StyleEditor.ts` line 1000 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5701#-NQaMxUb2qw-Q2uEdsSs:-NQlJn3h-VvzgAQFzabk:bilo656) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/b868c2f08ee32614b586ff4d8656108e5cf73c25/src/BloomBrowserUI/bookEdit/StyleEditor/StyleEditor.ts#L1000)):*

I see this class on some `bloom-editables`: `bloom-nodefaultstylerule`, but it seems to occur in conjunction with `X-style` classes too. I'm not sure what to do here.
</blockquote></details>

It's quite possible that whatever situation caused this problem may have been fixed by a fix that JT put into `5.4` for SS import. Unfortunately, even by going back to `5.3` prior to the fix and importing their SS I can't duplicate the problem.

PR #5709, opened 2023-03-17
Features panel of new Upload screen (BL-11582)

JohnThomson:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5709)

### DIFF
--- a/DistFiles/localization/en/Bloom.xlf
+++ b/DistFiles/localization/en/Bloom.xlf
@@ -4514,6 +4514,11 @@
         <source xml:lang="en">Additional Requests: </source>
         <note>ID: PublishTab.Upload.AdditionalRequests</note>
       </trans-unit>
+      <trans-unit id="PublishTab.Upload.Advertise" translate="no">
+        <!-- added in 5.5 -->
+        <source xml:lang="en">Advertise these special Features</source>
+        <note>ID: PublishTab.Upload.Advertise</note>
+      </trans-unit>
       <trans-unit id="PublishTab.Upload.AllReserved">
         <source xml:lang="en">All rights reserved (Contact the Copyright holder for any permissions.)</source>
         <note>ID: PublishTab.Upload.AllReserved</note>
--- a/src/BloomBrowserUI/publish/ReaderPublish/PublishFeaturesGroup.tsx
+++ b/src/BloomBrowserUI/publish/ReaderPublish/PublishFeaturesGroup.tsx
@@ -8,28 +8,51 @@ import { useApiBoolean } from "../../utils/bloomApi";
 export const PublishFeaturesGroup: React.FunctionComponent<{
     onChange?: () => void;
 }> = props => {
-    const [motionEnabled] = useApiBoolean(
-        "publish/android/canHaveMotionMode",
+    // const [motionEnabled] = useApiBoolean(
+    //     "publish/android/canHaveMotionMode", // awkward to use a publish.android API here in upload screen, but want same implementation
+    //     false
+    // );
+    const [blindAccessibleEnabled] = useApiBoolean(
+        "libraryPublish/blindAccessibleEnabled",
+        false
+    );
+    const [signLanguageEnabled] = useApiBoolean(
+        "libraryPublish/signLanguageEnabled",
         false
     );
 
     return (
         <SettingsGroup
-            label={useL10n("Features", "PublishTab.Android.Features")}
+            label={useL10n(
+                "Advertise these special Features",
+                "PublishTab.Upload.Advertise"
+            )}
         >
             <FormGroup>
-                {/* <ApiCheckbox label="Talking Book" /> */}
                 <ApiCheckbox
+                    english="Accessible to the Blind"
+                    l10nKey="PublishTab.Upload.AccessibleToBlind"
+                    apiEndpoint="libraryPublish/accessibleToBlind"
+                    disabled={!blindAccessibleEnabled}
+                    onChange={props.onChange}
+                />
+                {/* <ApiCheckbox label="Accessible to the visually impaired" /> */}
+                <ApiCheckbox
+                    english="Sign Language"
+                    l10nKey="PublishTab.Upload.SignLanguage"
+                    apiEndpoint="libraryPublish/signLanguage"
+                    disabled={!signLanguageEnabled}
+                />
+                {/* <ApiCheckbox
                     english="Motion Book"
+                    // Need new APIs here, similar but using different feature set. Or generalize somehow.
                     l10nKey="PublishTab.Android.MotionBookMode"
                     // tslint:disable-next-line:max-line-length
                     l10nComment="Motion Books are Talking Books in which the picture fills the screen, then pans and zooms while you hear the voice recording. This happens only if you turn the book sideways."
                     apiEndpoint="publish/android/motionBookMode"
                     onChange={props.onChange}
                     disabled={!motionEnabled}
-                />
-                {/* <ApiCheckbox label="Sign Language" />
-            <ApiCheckbox label="Image Descriptions" /> */}
+                /> */}
             </FormGroup>
         </SettingsGroup>
     );
--- a/src/BloomBrowserUI/publish/ReaderPublish/PublishLanguagesGroup.tsx
+++ b/src/BloomBrowserUI/publish/ReaderPublish/PublishLanguagesGroup.tsx
@@ -15,15 +15,17 @@ export interface ILanguagePublishInfo {
     includeText: boolean;
     containsAnyAudio: boolean;
     includeAudio: boolean;
+    isL1: boolean;
 }
 
-class LanguagePublishInfo implements ILanguagePublishInfo {
+export class LanguagePublishInfo implements ILanguagePublishInfo {
     public code: string;
     public name: string;
     public complete: boolean;
     public includeText: boolean;
     public containsAnyAudio: boolean;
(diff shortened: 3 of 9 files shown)


andrew-polk (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5709)* status: all files reviewed, 12 unresolved discussions (waiting on @JohnThomson)

___
*[`src/BloomBrowserUI/publish/commonPublish/LanguageGroup.tsx` line 1 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5709#-NQkeb0p-8BcWp86J41l:-NQkeb0t0XI73QkvQ8JI:b-byz8ep) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/1ef408146efa8135413b10db4b9f483c94257286/src/BloomBrowserUI/publish/commonPublish/LanguageGroup.tsx#L1)):*
> ```XML
> import * as React from "react";
> ```

Apparently when JH created this file, he wanted it to be something "common" (based on the folder it is in).
If it really is specific to library publish, I guess we need to move it.
If it isn't, I guess we need to move the API handlers.

As a corollary, `LanguageSelectionSettingsGroup` wants to be in `commonPublish` folder. (Probably a separate commit for history.)
___
*[`src/BloomBrowserUI/publish/commonPublish/LanguageGroup.tsx` line 53 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5709#-NQkfagf8T0amFa5ikbt:-NQkfagj4PdVXQfZrf65:b-wi9zrq) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/1ef408146efa8135413b10db4b9f483c94257286/src/BloomBrowserUI/publish/commonPublish/LanguageGroup.tsx#L53)):*
> ```XML
>     // Copied from similar control in BloomPub screen, this will handle individual controls
>     // for narration languages by giving a different fieldToUpdate. Currently we only use
>     // one value for that.
> ```

Sorry; I think I'm just being dense, but I'm not really making sense of this comment.
I think you're trying to explain why this is different enough from the original to justify making a second instance of it.
But the explanation doesn't seem to match the diff I see (1. different api endpoint, 1. doesn't call props.onChange()).
___
*[`src/BloomBrowserUI/publish/ReaderPublish/PublishFeaturesGroup.tsx` line 8 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5709#-NQkcGsZ5Vei6GGVqz7q:-NQkcGsZ5Vei6GGVqz7r:buon3h) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/1ef408146efa8135413b10db4b9f483c94257286/src/BloomBrowserUI/publish/ReaderPublish/PublishFeaturesGroup.tsx#L8)):*
> ```XML
> import { useApiBoolean } from "../../utils/bloomApi";
> 
> export const PublishFeaturesGroup: React.FunctionComponent<{
> ```

This component is currently in use for bloomPUB, but it seems you've heavily modified it here in a way which makes it seem like you didn't realize that.

If it is going to be desired/possible to share this component between the screens in the future, we'll want to move it from its current `ReaderPublish` folder to the `commonPublish` folder. (probably a separate commit for history sake?)
Also, the API calls probably want to be in a common publish API, not libraryPublish.

I'm not sure what to do at this point where this "share" component is completely different between the two screens.
It is obvious that JH wants "Motion Book" on "Publish to Web", but does he want the others on "Publish BloomPUB"?
___
*[`src/BloomExe/web/controllers/LibraryPublishApi.cs` line 39 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5709#-NQkaZc4ACH1NKw2fQ5g:-NQkaZc4ACH1NKw2fQ5h:byenc10) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/1ef408146efa8135413b10db4b9f483c94257286/src/BloomExe/web/controllers/LibraryPublishApi.cs#L39)):*
> ```Smalltalk
> 
> 		// This should probably be persisted somehow, but for now, it just applies to this session.
> 		private bool _includeBackgroundMusic = true;
> ```

Don't we want to go ahead and do this now?
___
*[`src/BloomExe/web/controllers/LibraryPublishApi.cs` line 94 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5709#-NQkb6-F5YokI2DkByCL:-NQkb6-F5YokI2DkByCM:b-6q835g) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/1ef408146efa8135413b10db4b9f483c94257286/src/BloomExe/web/controllers/LibraryPublishApi.cs#L94)):*
> ```Smalltalk
> 						// I don't know why we only want to set it for one language or why we should choose
> 						// one arbitrarily like this. The LanguagesCheckedToUpload does not have any
> 						// significant order.
> ```

I think while we're here, we should fix this. At least we can't leave it arbitrary which must be wrong.
___
*[`src/BloomExe/web/controllers/LibraryPublishApi.cs` line 111 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5709#-NQkbSqe-usVaGdymqnu:-NQkbSqe-usVaGdymqnv:bz73gaa) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/1ef408146efa8135413b10db4b9f483c94257286/src/BloomExe/web/controllers/LibraryPublishApi.cs#L111)):*
> ```Smalltalk
> 				{
> 					if (val)
> 						Model.SetOnlySignLanguageToPublish(Model.Book.CollectionSettings.SignLanguageTag);
> ```

The old screen had a way to set the SL. If we are dropping that feature, we should at least verify with JH (if you haven't).
___
*[`src/BloomExe/web/controllers/LibraryPublishApi.cs` line 241 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5709#-NQkaKW89KvbjdUk6afh:-NQkaKW89KvbjdUk6afi:bo5dy5s) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/1ef408146efa8135413b10db4b9f483c94257286/src/BloomExe/web/controllers/LibraryPublishApi.cs#L241)):*
> ```Smalltalk
> 			_progress.CancelRequested = false;
> 
> 			//TODO verify user doesn't select sign language feature without setting language
> ```

Need to handle this TODO
___
*[`src/BloomExe/web/controllers/LibraryPublishApi.cs` line 336 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5709#-NQkaN-d6Gc-dvp61Rzf:-NQkaN-d6Gc-dvp61Rzg:bo5dy5s) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/1ef408146efa8135413b10db4b9f483c94257286/src/BloomExe/web/controllers/LibraryPublishApi.cs#L336)):*
> ```Smalltalk
> 		void BackgroundUpload(object _, DoWorkEventArgs e)
> 		{
> 			// TODO get selected languages
> ```

Need to handle this TODO
___
*[`src/BloomExe/web/controllers/LibraryPublishApi.cs` line 348 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5709#-NQkaOqn389_HMCPekOc:-NQkaOqn389_HMCPekOd:bo5dy5s) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/1ef408146efa8135413b10db4b9f483c94257286/src/BloomExe/web/controllers/LibraryPublishApi.cs#L348)):*
> ```Smalltalk
> 			}
> 
> 			//TODO
> ```

Need to handle this TODO
___
*[`src/BloomExe/web/controllers/LibraryPublishApi.cs` line 353 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5709#-NQkaPLL2goqkIN2KBYa:-NQkaPLL2goqkIN2KBYb:bo5dy5s) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/1ef408146efa8135413b10db4b9f483c94257286/src/BloomExe/web/controllers/LibraryPublishApi.cs#L353)):*
> ```Smalltalk
> 			);
> 
> 			//TODO
> ```

Need to handle this TODO
___
*[`src/BloomExe/web/controllers/LibraryPublishApi.cs` line 435 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5709#-NQkaTjg5PUuUQZxtjbZ:-NQkaTjg5PUuUQZxtjb_:bo5dy5s) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/1ef408146efa8135413b10db4b9f483c94257286/src/BloomExe/web/controllers/LibraryPublishApi.cs#L435)):*
> ```Smalltalk
> 			if (Model.BookIsAlreadyOnServer)
> 			{
> 				// TODO
> ```

Need to handle this TODO
___
*[`src/BloomExe/web/controllers/LibraryPublishApi.cs` line 453 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5709#-NQkaGQK4ulmBWQVmDQV:-NQkaGQK4ulmBWQVmDQW:bo5dy5s) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/1ef408146efa8135413b10db4b9f483c94257286/src/BloomExe/web/controllers/LibraryPublishApi.cs#L453)):*
> ```Smalltalk
> 		private IEnumerable<string> GetLanguagesToUpload()
> 		{
> 			// TODO waiting to see how we do the settings before I know if this is correct.
> ```

Need to handle this TODO

PR #5714, opened 2023-03-20
RunJavascriptAsync (BL-11919)

JohnThomson:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5714)

### DIFF
--- a/src/BloomExe/CLI/ChangeLayoutCommand.cs
+++ b/src/BloomExe/CLI/ChangeLayoutCommand.cs
@@ -3,6 +3,7 @@ using System.Diagnostics;
 using System.IO;
 using System.Linq;
 using System.Text;
+using System.Threading.Tasks;
 using System.Windows.Forms;
 using Bloom.Book;
 using Bloom.Collection;
@@ -23,7 +24,7 @@ namespace Bloom.CLI
 	// new layout has places for.
 	class ChangeLayoutCommand
 	{
-		public static int Handle(ChangeLayoutParameters options)
+		public static async Task<int> Handle(ChangeLayoutParameters options)
 		{
 			// This task will be all the program does. We need to do enough setup so that
 			// books can be created, then do our work, then tear things down.
--- a/src/BloomExe/CLI/ChangeLayoutCommand.cs
+++ b/src/BloomExe/CLI/ChangeLayoutCommand.cs
@@ -24,7 +24,7 @@ namespace Bloom.CLI
 	// new layout has places for.
 	class ChangeLayoutCommand
 	{
-		public static async Task<int> Handle(ChangeLayoutParameters options)
+		public static Task<int> Handle(ChangeLayoutParameters options)
 		{
 			// This task will be all the program does. We need to do enough setup so that
 			// books can be created, then do our work, then tear things down.
@@ -38,12 +38,12 @@ namespace Bloom.CLI
 					LocalizationManager.SetUILanguage(Settings.Default.UserInterfaceLanguage, false);
 					ChangeLayoutForAllContentPagesInAllBooks(options.CollectionPath, options.BookPath, options.PageGuid);
 				}
-				return 0;
+				return Task.FromResult(0);
 			}
 			catch (Exception ex)
 			{
 				Console.WriteLine(ex);
-				return 1;
+				return Task.FromResult(1);
 			}
 		}
 
--- a/src/BloomExe/CLI/CreateArtifactsCommand.cs
+++ b/src/BloomExe/CLI/CreateArtifactsCommand.cs
@@ -4,6 +4,7 @@ using System.Diagnostics;
 using System.IO;
 using System.IO.Compression;
 using System.Threading;
+using System.Threading.Tasks;
 using System.Windows.Forms;
 using Bloom.Book;
 using Bloom.Properties;
@@ -69,7 +70,7 @@ namespace Bloom.CLI
 			return errors;
 		}
 
-		public static int Handle(CreateArtifactsParameters options)
+		public static async Task<int> Handle(CreateArtifactsParameters options)
 		{
 			try
 			{
--- a/src/BloomExe/CLI/CreateArtifactsCommand.cs
+++ b/src/BloomExe/CLI/CreateArtifactsCommand.cs
@@ -70,7 +70,7 @@ namespace Bloom.CLI
 			return errors;
 		}
 
-		public static async Task<int> Handle(CreateArtifactsParameters options)
+		public static Task<int> Handle(CreateArtifactsParameters options)
 		{
 			try
 			{
@@ -96,14 +96,14 @@ namespace Bloom.CLI
 
 						// Make the .bloompub and /bloomdigital outputs
 						var exitCode = CreateArtifacts(options);
-						return (int)exitCode;
+						return Task.FromResult((int)exitCode);
 					}
 				}
 			}
 			catch (Exception ex)
 			{
 				Console.WriteLine(ex);
-				return (int)CreateArtifactsExitCode.UnhandledException;
+				return Task.FromResult((int)CreateArtifactsExitCode.UnhandledException);
 			}
 		}
 
--- a/src/BloomExe/CLI/DownloadBookCommand.cs
+++ b/src/BloomExe/CLI/DownloadBookCommand.cs
@@ -14,7 +14,7 @@ namespace Bloom.CLI
 	/// </summary>
 	class DownloadBookCommand
 	{
-		public static async Task<int> HandleSilentDownload(DownloadBookOptions options)
+		public static Task<int> HandleSilentDownload(DownloadBookOptions options)
 		{
 			// This task will be all the program does. We need to do enough setup so that
 			// the download code can work, then tear it down.
@@ -34,12 +34,12 @@ namespace Bloom.CLI
 					downloader.HandleDownloadWithoutProgress(options.Url, options.DestinationPath);
 					Console.WriteLine(("\ndownload complete\n"));
 				}
-				return 0;
+				return Task.FromResult(0);
 			}
 			catch (Exception ex)
 			{
 				Console.WriteLine(ex);
-				return 1;
+				return Task.FromResult(1);
 			}
 		}
 	}
--- a/src/BloomExe/CLI/DownloadBookCommand.cs
+++ b/src/BloomExe/CLI/DownloadBookCommand.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Diagnostics;
+using System.Threading.Tasks;
 using Bloom.Properties;
 using Bloom.WebLibraryIntegration;
 using CommandLine;
@@ -13,7 +14,7 @@ namespace Bloom.CLI
 	/// </summary>
 	class DownloadBookCommand
 	{
-		public static int HandleSilentDownload(DownloadBookOptions options)
+		public static async Task<int> HandleSilentDownload(DownloadBookOptions options)
 		{
 			// This task will be all the program does. We need to do enough setup so that
 			// the download code can work, then tear it down.
--- a/src/BloomExe/CLI/GetUsedFontsCommand.cs
+++ b/src/BloomExe/CLI/GetUsedFontsCommand.cs
@@ -33,7 +33,7 @@ namespace Bloom.CLI
 	/// </summary>
 	class GetUsedFontsCommand
 	{
-		public static async  Task<int> Handle(GetUsedFontsParameters options)
+		public static Task<int> Handle(GetUsedFontsParameters options)
 		{
 			try
 			{
@@ -49,7 +49,7 @@ namespace Bloom.CLI
 						Debug.WriteLine("Could not find " + options.BookPath);
 						Console.Error.WriteLine("Could not find " + options.BookPath);
 					}
-					return (int)GetUsedFontsExitCode.BookPathDirectoryNotFound;
+					return Task.FromResult((int)GetUsedFontsExitCode.BookPathDirectoryNotFound);
 				}
 				Console.WriteLine("Gathering font data.");
 
@@ -89,19 +89,19 @@ namespace Bloom.CLI
 					string message = "Exception: " + e.ToString();
 					Debug.WriteLine(message);
 					Console.Error.WriteLine(message);
-					return (int)GetUsedFontsExitCode.ReportIOError;
+					return Task.FromResult((int)GetUsedFontsExitCode.ReportIOError);
 				}
 
 				Console.WriteLine("Finished gathering font data.");
 				Debug.WriteLine("Finished gathering font data.");
-				return (int)GetUsedFontsExitCode.Success;
+				return Task.FromResult((int)GetUsedFontsExitCode.Success);
 			}
 			catch (Exception e)
 			{
 				string message = "Exception: " + e.ToString();
 				Debug.WriteLine(message);
 				Console.Error.WriteLine(message);
-				return (int)GetUsedFontsExitCode.UnhandledException;
+				return Task.FromResult((int)GetUsedFontsExitCode.UnhandledException);
 			}
 		}
 
--- a/src/BloomExe/CLI/GetUsedFontsCommand.cs
+++ b/src/BloomExe/CLI/GetUsedFontsCommand.cs
@@ -5,6 +5,7 @@ using System.IO;
 using System.Linq;
 using System.Text;
 using System.Text.RegularExpressions;
+using System.Threading.Tasks;
 using Bloom.Book;
 using CommandLine;
 using SIL.IO;
@@ -32,7 +33,7 @@ namespace Bloom.CLI
 	/// </summary>
 	class GetUsedFontsCommand
 	{
-		public static int Handle(GetUsedFontsParameters options)
+		public static async  Task<int> Handle(GetUsedFontsParameters options)
 		{
 			try
 			{
--- a/src/BloomExe/CLI/HydrateBookCommand.cs
+++ b/src/BloomExe/CLI/HydrateBookCommand.cs
@@ -2,6 +2,7 @@
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.IO;
+using System.Threading.Tasks;
 using System.Xml;
 using Bloom.Book;
 using Bloom.Collection;
@@ -22,7 +23,7 @@ namespace Bloom.CLI
 	/// </summary>
 	class HydrateBookCommand
 	{
-		public static int Handle(HydrateParameters options)
+		public static async Task<int> Handle(HydrateParameters options)
 		{
 			if (!Directory.Exists(options.Path))
 			{
--- a/src/BloomExe/CLI/HydrateBookCommand.cs
+++ b/src/BloomExe/CLI/HydrateBookCommand.cs
@@ -23,7 +23,7 @@ namespace Bloom.CLI
 	/// </summary>
 	class HydrateBookCommand
 	{
-		public static async Task<int> Handle(HydrateParameters options)
+		public static Task<int> Handle(HydrateParameters options)
 		{
 			if (!Directory.Exists(options.Path))
 			{
@@ -37,7 +37,7 @@ namespace Bloom.CLI
 					Debug.WriteLine("Could not find " + options.Path);
 					Console.Error.WriteLine("Could not find " + options.Path);
 				}
-				return 1;
+				return Task.FromResult(1);
 			}
 			Console.WriteLine("Starting Hydrating.");
 
@@ -91,7 +91,7 @@ namespace Bloom.CLI
 			book.EnsureUpToDate();
 			Console.WriteLine("Finished Hydrating.");
 			Debug.WriteLine("Finished Hydrating.");
-			return 0;
+			return Task.FromResult(0);
 		}
 	}
 }
--- a/src/BloomExe/CLI/SendFontAnalyticsCommand.cs
+++ b/src/BloomExe/CLI/SendFontAnalyticsCommand.cs
@@ -26,7 +26,7 @@ namespace Bloom.CLI
 		static ProjectContext _projectContext;
 		static Book.Book _book;
 
-		public static async Task<int> Handle(SendFontAnalyticsParameters options)
+		public static Task<int> Handle(SendFontAnalyticsParameters options)
 		{
 			Program.SetUpErrorHandling();
 			try
@@ -47,16 +47,16 @@ namespace Bloom.CLI
 							// (See https://issues.bloomlibrary.org/youtrack/issue/BL-11512.)
 							ReportFontAnalytics(_book.ID, "harvester sendFontAnalytics", options.Testing,
 								options.SkipEpubAnalytics, options.SkipPdfAnalytics);
-							return (int)SendFontAnalyticsExitCode.Success;
+							return Task.FromResult((int)SendFontAnalyticsExitCode.Success);
 						}
 					}
-					return (int)SendFontAnalyticsExitCode.NoFontsFound;
+					return Task.FromResult((int)SendFontAnalyticsExitCode.NoFontsFound);
 				}
 			}
 			catch (Exception ex)
 			{
 				Console.WriteLine(ex);
-				return (int)SendFontAnalyticsExitCode.UnhandledException;
+				return Task.FromResult((int)SendFontAnalyticsExitCode.UnhandledException);
 			}
 		}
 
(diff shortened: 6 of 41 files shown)


andrew-polk (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5714)* status: 0 of 34 files reviewed, 2 unresolved discussions (waiting on @JohnThomson)

___
*[`src/BloomExe/IBrowser.cs` line 448 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5714#-NR0HKzk-4yEo8fqqDVh:-NR0HKzk-4yEo8fqqDVi:b-5wt98d) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/583c737bf5fbe588e7ca3177cff38114b8c1b00e/src/BloomExe/IBrowser.cs#L448)):*
> ```Smalltalk
> 		public abstract string RunJavaScript(string script);
> 
> 		public abstract Task<string> RunJavaScriptAsync(string script);
> ```

Shouldn't this be an `async` method?
(maybe you can't have `abstract async`?)
___
*[`src/BloomExe/WebView2Browser.cs` line 321 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5714#-NR0HpI274nhJt5CWrC7:-NR0HmOu44O5UQaYgPm0:bvaz8iq) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/583c737bf5fbe588e7ca3177cff38114b8c1b00e/src/BloomExe/WebView2Browser.cs#L321)):*
> ```Smalltalk
> 		}
> 
> 		public override Task<string> RunJavaScriptAsync(string script)
> ```

Shouldn't this be an `async` method?

Why do we have this and the following method? (And why does the following method start with a lower-case r?
I think the comment on the following method needs to be updated, right?

andrew-polk (commented):
Just did a cursory look, but noticed a couple things I had questions about.

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5714#-:-NR0IAhcDjW138BEsdPq:bomfsc)* status: 0 of 34 files reviewed, 2 unresolved discussions (waiting on @JohnThomson)

andrew-polk (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5714)* status: 0 of 34 files reviewed, 2 unresolved discussions (waiting on @JohnThomson)

___
*[`src/BloomExe/GeckoFxBrowser.cs` line 830 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5714#-NR3nEllBM6fMc60Jf8w:-NR3nEllBM6fMc60Jf8x:b-qturw4) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/583c737bf5fbe588e7ca3177cff38114b8c1b00e/src/BloomExe/GeckoFxBrowser.cs#L830)):*
> ```Smalltalk
> 		}
> 
> 		public static string RunJavaScriptOn(GeckoWebBrowser geckoWebBrowser, string script)
> ```

I don't suppose anyone could accidentally use this now...
But let's go ahead and make it private.

jsubloom (changes requested):
> Reviewed 5 of 8 files at r1.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5714)* status: 3 of 34 files reviewed, 7 unresolved discussions (waiting on @andrew-polk and @JohnThomson)

___
*[`src/BloomExe/Edit/ConfigurationDialog.cs` line 57 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5714#-NR3rt8fCegeCVF0L4P8:-NR3rt8fCegeCVF0L4P9:b-bnug3y) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/583c737bf5fbe588e7ca3177cff38114b8c1b00e/src/BloomExe/Edit/ConfigurationDialog.cs#L57)):*
> ```Smalltalk
> 		}
> 
> 		async private void  _okButton_Click(object sender, EventArgs e)
> ```

I think "private async" is more idiomatic than "async private". For example, see the official docs: https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/async, which use "private async"
___
*[`src/BloomExe/Publish/PDF/PdfViewer.cs` line 202 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5714#-NR3sp-78VPQut3PAW_9:-NR3sp-78VPQut3PAW_A:bec5b14) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/583c737bf5fbe588e7ca3177cff38114b8c1b00e/src/BloomExe/Publish/PDF/PdfViewer.cs#L202)):*
> ```Smalltalk
> 			{
> 				var wv2 = _pdfViewerControl as WebView2Browser;
> 				wv2.RunJavaScriptAsync("window.print()");
> ```

I'm not so sure about this. Is this intentional? Did you mean to make Print async, or are we intentionally trying to fire-and-forget the async operation?
___
*[`src/BloomExe/Spreadsheet/SpreadsheetImporter.cs` line 169 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5714#-NR3uNbw5Bn0X161OrqU:-NR3uNbw5Bn0X161OrqV:bglxgl3) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/583c737bf5fbe588e7ca3177cff38114b8c1b00e/src/BloomExe/Spreadsheet/SpreadsheetImporter.cs#L169)):*
> ```Smalltalk
> 				_browser.Navigate(rootPage, false);
> 				// This extra check that spreadsheetBundle actually exists might not be necessary.
> 				var task = _browser.RunJavaScriptAsync($"spreadsheetBundle");
> ```

I'm not sure what the point of converting to RunJavascriptAsync was here. It didn't seem to make our life any better.  Can you explain what's going on? 

Actually, what I think should happen overall in this file is:
1) An InitializeAsync() method or similar should be added.
2) InitializeAsync() should take charge of loading up the browser / spreadsheet bundle.
3) GetBrowser() can synchronously return the previously loaded browser. I guess you can optioanlly throw an exception if GetBrowser() is called before InitializeAsync() completes by utilizing a isLoaded member variable or something like that.
4) Any callers to GetBrowser() such as GetSentenceFragments() can synchronously return without needing to await anything.

Does that make sense?
___
*[`src/BloomExe/ToPalaso/ProgressDialogForeground.cs` line 35 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5714#-NR3wlbiEyVE7Q2OvJtm:-NR3wlbj4rbKV4rZ31bF:b-mgpr20) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/583c737bf5fbe588e7ca3177cff38114b8c1b00e/src/BloomExe/ToPalaso/ProgressDialogForeground.cs#L35)):*
> ```Smalltalk
> 		}
> 
> 		public void ShowAndDoWork(Func<IProgress, Task> work)
> ```

Add a comment saying that work can be an async method.
___
*[`src/BloomExe/ToPalaso/ProgressDialogForeground.cs` line 61 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5714#-NR3yYq4024XCeS8DWqB:-NR3yYq59QgH6j0Vp7Z6:b-ntopte) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/583c737bf5fbe588e7ca3177cff38114b8c1b00e/src/BloomExe/ToPalaso/ProgressDialogForeground.cs#L61)):*
> ```Smalltalk
> 		{
> 			Application.Idle -= new EventHandler(StartWorking);
> 			if (_work != null)
> ```

I think this works, but I think it could be more straightforward as follows:

1) StartWorking becomes an async method.  (async Task StartWorkingAsync)..  Not 1,000% sure, but as far as I know, there's no problem with Application.Idle receiving an async event handler, right? 
2) I guess you can still have two overloads of ShowAndDoWork. But I'd have the first overload which takes the Action just wrap the action so that it is of type Func<IProgress, Task> and pass it to the 2nd overload. 
3) Now we just need one member variable instead of two, so get rid of the extraneous one.
4) The old StartWorking code can be simplified to:
Application.Idle -= ...
await _asyncWork(Progress)
Close();

I like that this avoids having two mutually exclusive member variables to store the work, which I wasn't a huge fan of. I think having one member variable is more understandable.

What do you think?

jsubloom (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5714)* status: 3 of 34 files reviewed, 7 unresolved discussions (waiting on @andrew-polk and @JohnThomson)

___
*[`src/BloomExe/Spreadsheet/SpreadsheetImporter.cs` line 169 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5714#-NR3uNbw5Bn0X161OrqU:-NR4-IIv3UA9konA64rK:b-szzigp) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/583c737bf5fbe588e7ca3177cff38114b8c1b00e/src/BloomExe/Spreadsheet/SpreadsheetImporter.cs#L169)):*

I'm not sure what the point of converting to RunJavascriptAsync was here. It didn't seem to make our life any better.  Can you explain what's going on? 

Actually, what I think should happen overall in this file is:
1) An InitializeAsync() method or similar should be added.
2) InitializeAsync() should take charge of loading up the browser / spreadsheet bundle.
3) GetBrowser() can synchronously return the previously loaded browser. I guess you can optioanlly throw an exception if GetBrowser() is called before InitializeAsync() completes by utilizing a isLoaded member variable or something like that.
4) Any callers to GetBrowser() such as GetSentenceFragments() can synchronously return without needing to await anything.

Does that make sense?
</blockquote></details>

or, if it's too burdensome to make initalizeAsync async... I think I'd prefer having the cludgy code that synchronously waits for the async stuff to be done in the initialize method instead?

andrew-polk (commented):
> Reviewed 1 of 7 files at r6.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5714)* status: 2 of 35 files reviewed, 6 unresolved discussions (waiting on @JohnThomson and @jsubloom)

jsubloom (changes requested):
> Reviewed 16 of 27 files at r2, 1 of 1 files at r4, 1 of 1 files at r5, 7 of 7 files at r6, 8 of 8 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5714)* status: 34 of 35 files reviewed, 4 unresolved discussions (waiting on @andrew-polk and @JohnThomson)

___
*[`src/BloomExe/MiscUI/BrowserProgressDialog.cs` line 121 at r7](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5714#-NRKU3hcBE_1DYDA2o8_:-NRKU3hcBE_1DYDA2o8a:b-spae8j) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/c779013103bba71db0e7ddc5b564b16e9f77f219/src/BloomExe/MiscUI/BrowserProgressDialog.cs#L121)):*
> ```Smalltalk
> 		private static Action _doWhenProgressDialogCloses;
> 
> 		public static async Task DoWorkWithProgressDialog(IBloomWebSocketServer socketServer,
> ```

In C# it's usually idiomatic to append "Async" to the method names for async methods. (Although Javascript is less of a fan of that practice). Any reason we shouldn't do that here?

Similar question to many other places.
Wouldn't it make it easier to spot which methods ought to be awaited or not?
___
*[`src/BloomExe/Publish/PublishView.cs` line 774 at r7](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5714#-NRKUFZn4V4AXnkvTU07:-NRKUFZn4V4AXnkvTU08:bq7crq4) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/c779013103bba71db0e7ddc5b564b16e9f77f219/src/BloomExe/Publish/PublishView.cs#L774)):*
> ```Smalltalk
> 		}
> 
> 		private async void OnPrint_Click(object sender, EventArgs e)
> ```

I think the return type should be Task<void> or Task instead.

Optionally - maybe Async should be appended to method name.
___
*[`src/BloomExe/ToPalaso/ProgressDialogForeground.cs` line 38 at r7](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5714#-NRKYLeVC_JS7bGA7X81:-NRKYLeVC_JS7bGA7X82:bi3xpwe) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/c779013103bba71db0e7ddc5b564b16e9f77f219/src/BloomExe/ToPalaso/ProgressDialogForeground.cs#L38)):*
> ```Smalltalk
> 		}
> 
> 		// Typically work is an async function that conceptually returns void,
> ```

The parameter name "work" is now "asyncWork"



_Code quote:_
```Smalltalk
work
```

JohnThomson (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5714)* status: 13 of 41 files reviewed, 4 unresolved discussions (waiting on @andrew-polk and @jsubloom)

___
*[`src/BloomExe/GeckoFxBrowser.cs` line 830 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5714#-NR3nEllBM6fMc60Jf8w:-NRJH5vPC-ckPrIvWHpD:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/583c737bf5fbe588e7ca3177cff38114b8c1b00e/src/BloomExe/GeckoFxBrowser.cs#L830)):*

I don't suppose anyone could accidentally use this now...
But let's go ahead and make it private.
</blockquote></details>

Done.
___
*[`src/BloomExe/IBrowser.cs` line 448 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5714#-NR0HKzk-4yEo8fqqDVh:-NR3Q9lc8svK1busVlo3:b-ob6sdf) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/583c737bf5fbe588e7ca3177cff38114b8c1b00e/src/BloomExe/IBrowser.cs#L448)):*

Shouldn't this be an `async` method?
(maybe you can't have `abstract async`?)
</blockquote></details>

Right, abstract methods can't be async. Seems silly since users of the interface need to know. But the fact that it returns a Task tells the compiler that it can be awaited, and all the async really does is tell the compiler that the body of the method is allowed to use await. But an abstract method has no body!
___
*[`src/BloomExe/WebView2Browser.cs` line 321 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5714#-NR0HpI274nhJt5CWrC7:-NR3Qm5-3wzp7-JkK8Ix:b5j96jc) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/583c737bf5fbe588e7ca3177cff38114b8c1b00e/src/BloomExe/WebView2Browser.cs#L321)):*

Shouldn't this be an `async` method?

Why do we have this and the following method? (And why does the following method start with a lower-case r?
I think the comment on the following method needs to be updated, right?
</blockquote></details>

I think I got away with it because it doesn't actually use await. It just returns the task returned by the async method below. But you are right, there's no reason to have both. I merged them.
___
*[`src/BloomExe/Edit/ConfigurationDialog.cs` line 57 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5714#-NR3rt8fCegeCVF0L4P8:-NRJHHcf7tG9fjwENt6Z:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/583c737bf5fbe588e7ca3177cff38114b8c1b00e/src/BloomExe/Edit/ConfigurationDialog.cs#L57)):*

I think "private async" is more idiomatic than "async private". For example, see the official docs: https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/async, which use "private async"
</blockquote></details>

Done.
___
*[`src/BloomExe/MiscUI/BrowserProgressDialog.cs` line 121 at r7](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5714#-NRKU3hcBE_1DYDA2o8_:-NRYdkCr36beHCepbVC2:bq3j1ie) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/c779013103bba71db0e7ddc5b564b16e9f77f219/src/BloomExe/MiscUI/BrowserProgressDialog.cs#L121)):*

In C# it's usually idiomatic to append "Async" to the method names for async methods. (Although Javascript is less of a fan of that practice). Any reason we shouldn't do that here?

Similar question to many other places.
Wouldn't it make it easier to spot which methods ought to be awaited or not?
</blockquote></details>

I think I have done this systematically through our code base, except that it did not seem useful to rename Test methods that are only ever called by NUnit. (IBrowser.UpdateDisplay is another exception, which I have commented.)
___
*[`src/BloomExe/Publish/PublishView.cs` line 774 at r7](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5714#-NRKUFZn4V4AXnkvTU07:-NRKl7-90n1huax8TQ_n:bu94sdm) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/c779013103bba71db0e7ddc5b564b16e9f77f219/src/BloomExe/Publish/PublishView.cs#L774)):*

I think the return type should be Task<void> or Task instead.

Optionally - maybe Async should be appended to method name.
</blockquote></details>

It can't be. The return type of an event handler is required to be void. If it is actually invoked by the event being raised, the assumption is that it doesn't matter that some of the work ends up getting put back in the queue and something else might happen in the meantime. I think that's true in this case. When it does matter, you're in trouble. I think even if you make another async method and await the result, the code after the await might still get executed in another event loop cycle and other things might happen in between.
(If something else wants to call the event handler and await it, you can of course put the relevant code into another method which the event handler also calls, and have that return Task. But that's not the case here.)
___
*[`src/BloomExe/Publish/PDF/PdfViewer.cs` line 202 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5714#-NR3sp-78VPQut3PAW_9:-NRJCgVC9ksEwZjyu0hV:b-n321o4) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/583c737bf5fbe588e7ca3177cff38114b8c1b00e/src/BloomExe/Publish/PDF/PdfViewer.cs#L202)):*

I'm not so sure about this. Is this intentional? Did you mean to make Print async, or are we intentionally trying to fire-and-forget the async operation?
</blockquote></details>

I don't think any of the things we do after calling this depend on it having completed...it just puts the UI in a state where the user can print. But that could change, and it turned out to be easy to make two more methods async, so I did.
___
*[`src/BloomExe/Spreadsheet/SpreadsheetImporter.cs` line 169 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5714#-NR3uNbw5Bn0X161OrqU:-NRJJ0f59yB7bwDrS7AE:b-qne1aj) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/583c737bf5fbe588e7ca3177cff38114b8c1b00e/src/BloomExe/Spreadsheet/SpreadsheetImporter.cs#L169)):*

or, if it's too burdensome to make initalizeAsync async... I think I'd prefer having the cludgy code that synchronously waits for the async stuff to be done in the initialize method instead?
</blockquote></details>

In general, the point of converting to RunJavascriptAsync is to eventually get rid of the kludgy code in WebView2Browser.RunJavascript that tries to convert the async call to sync without causing a deadlock. I'm not confident that code can be trusted, or that it will continue to work with future versions of WebView2Browser, or that it will continue to work when we eventually abandon Windows.Forms.
We'd have to do a similar kludge to be sure that InitializeAsync has actually finished all its delayed tasks before calling anything that assumes GetBrowser is OK, so that kind of defeats the purpose. (You can't await anything in a constructor.)
But your comment did motivate me to try to get rid of the kludge here, and I did! Making GetBrowser async and using that to handle waiting for Navigate as well as for RunJavascript makes things much cleaner. (It also led to converting one more RunJavaScript to async, as a bonus.)
___
*[`src/BloomExe/ToPalaso/ProgressDialogForeground.cs` line 35 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5714#-NR3wlbiEyVE7Q2OvJtm:-NRJEg_J8ELWf5ZsRi81:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/583c737bf5fbe588e7ca3177cff38114b8c1b00e/src/BloomExe/ToPalaso/ProgressDialogForeground.cs#L35)):*

Add a comment saying that work can be an async method.
</blockquote></details>

Done.
___
*[`src/BloomExe/ToPalaso/ProgressDialogForeground.cs` line 61 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5714#-NR3yYq4024XCeS8DWqB:-NRJGuzWEawBl9H3ZWAV:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/583c737bf5fbe588e7ca3177cff38114b8c1b00e/src/BloomExe/ToPalaso/ProgressDialogForeground.cs#L61)):*

I think this works, but I think it could be more straightforward as follows:

1) StartWorking becomes an async method.  (async Task StartWorkingAsync)..  Not 1,000% sure, but as far as I know, there's no problem with Application.Idle receiving an async event handler, right? 
2) I guess you can still have two overloads of ShowAndDoWork. But I'd have the first overload which takes the Action just wrap the action so that it is of type Func<IProgress, Task> and pass it to the 2nd overload. 
3) Now we just need one member variable instead of two, so get rid of the extraneous one.
4) The old StartWorking code can be simplified to:
Application.Idle -= ...
await _asyncWork(Progress)
Close();

I like that this avoids having two mutually exclusive member variables to store the work, which I wasn't a huge fan of. I think having one member variable is more understandable.

What do you think?
</blockquote></details>

Done.
___
*[`src/BloomExe/ToPalaso/ProgressDialogForeground.cs` line 38 at r7](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5714#-NRKYLeVC_JS7bGA7X81:-NRKmXLmDW11VjKmP7QI:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/c779013103bba71db0e7ddc5b564b16e9f77f219/src/BloomExe/ToPalaso/ProgressDialogForeground.cs#L38)):*

The parameter name "work" is now "asyncWork"
</blockquote></details>

Done.

andrew-polk (commented):
> Reviewed 1 of 8 files at r1.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5714)* status: 14 of 41 files reviewed, 3 unresolved discussions (waiting on @jsubloom)

jsubloom (approved):
:lgtm:

> Reviewed 27 of 27 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5714#-:-NRYoTWj9oJB-qiM_HLg:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @JohnThomson)

___
*[`src/BloomExe/MiscUI/BrowserProgressDialog.cs` line 121 at r7](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5714#-NRKU3hcBE_1DYDA2o8_:-NRYnId5DWQYmJTENjsJ:b-a4plpw) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/c779013103bba71db0e7ddc5b564b16e9f77f219/src/BloomExe/MiscUI/BrowserProgressDialog.cs#L121)):*

I think I have done this systematically through our code base, except that it did not seem useful to rename Test methods that are only ever called by NUnit. (IBrowser.UpdateDisplay is another exception, which I have commented.)
</blockquote></details>

Thanks for all the changes
___
*[`src/BloomExe/Publish/PublishView.cs` line 774 at r7](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5714#-NRKUFZn4V4AXnkvTU07:-NRYnSKZEu23qLTGqAfY:b-eke7k9) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/c779013103bba71db0e7ddc5b564b16e9f77f219/src/BloomExe/Publish/PublishView.cs#L774)):*

It can't be. The return type of an event handler is required to be void. If it is actually invoked by the event being raised, the assumption is that it doesn't matter that some of the work ends up getting put back in the queue and something else might happen in the meantime. I think that's true in this case. When it does matter, you're in trouble. I think even if you make another async method and await the result, the code after the await might still get executed in another event loop cycle and other things might happen in between.
(If something else wants to call the event handler and await it, you can of course put the relevant code into another method which the event handler also calls, and have that return Task. But that's not the case here.)
</blockquote></details>

Gotcha, thanks for explaining that.

PR #5722, opened 2023-03-23
Ensure old books use default settings in harvester (BL-12039)

andrew-polk:


andrew-polk (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5722)* status: 0 of 1 files reviewed, 1 unresolved discussion

___
*[`src/BloomExe/Publish/Android/AndroidPublishSettings.cs` line 157 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5722#-NREgw7a5QoUM9CVVWNX:-NREgw7a5QoUM9CVVWNY:bnq12vg) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/2c2cb0c71d12d2ce901a0ef540f138ddf907e8f0/src/BloomExe/Publish/Android/AndroidPublishSettings.cs#L157)):*
> ```Smalltalk
> 			var book = bookServer.GetBookFromBookInfo(bookInfo);
> 			InitializeLanguageSettingsIfNeeded(book);
> 			return FromBookInfo(book.BookInfo);
> ```

In the spirit of trying to make sure we don't forget to call some initialization method (which is, in some ways what the original change was about)... I was going to move the initialize call into `FromBookInfo` (where it needed to become `FromBook`), but the only other caller of `FromBookInfo` right now is the PublishToAndroidApi, and I didn't feel safe initializing there if empty since, presumably it must already handle something like that.
If you think it best, I could look into it more to try to convince myself it would be safe.

JohnThomson (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5722)* status: 0 of 1 files reviewed, 3 unresolved discussions (waiting on @andrew-polk)

___
*[`src/BloomExe/Publish/Android/AndroidPublishSettings.cs` line 106 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5722#-NREmm_R3mc9PuWvFSbv:-NREmm_R3mc9PuWvFSbw:b-jm7edx) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/a629d64dd2ada21ec2150f79a8c97ef8d089cfa4/src/BloomExe/Publish/Android/AndroidPublishSettings.cs#L106)):*
> ```Smalltalk
> 			if (dictToUse == null)
> 				return new HashSet<string>();
> 			return new HashSet<string>(dictToUse.Where(kvp => kvp.Value.IsIncluded()).Select(kvp => kvp.Key));
> ```

Are you absolutely certain (here and below) that there is no path by which this function gets called but InitializeLanguageSettingsIfNeeded has not?
If so, I would be inclined to verify it by having the code throw an exception if the dictionary is empty.
If not, is it possible to refactor so that it gets called here if it is empty?
___
*[`src/BloomExe/Publish/Android/AndroidPublishSettings.cs` line 160 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5722#-NRElAGR0mtyjpNu8I3Y:-NRElAGR0mtyjpNu8I3Z:b2ffzp7) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/2c2cb0c71d12d2ce901a0ef540f138ddf907e8f0/src/BloomExe/Publish/Android/AndroidPublishSettings.cs#L160)):*
> ```Smalltalk
> 		}
> 
> 		private static void InitializeLanguageSettingsIfNeeded(Book.Book book)
> ```

I'm not clear why you pulled this into its own method, since there is only the one caller. Maybe it could both take and return a BookInfo?
The main reason I'd prefer that is that creating a book from a bookInfo is quite expensive...reading and parsing the file...and 99% of the time the relevant collection won't be empty and the Book is not needed.

JohnThomson (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5722)* status: all files reviewed, 1 unresolved discussion

JohnThomson (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5722)* status: all files reviewed, 1 unresolved discussion

PR #5724, opened 2023-03-23
Use common code for creating/remembering output filenames (BL-11996)

StephenMcConnel:
This changeset does not deal with spreadsheets, which look for folder paths instead of file paths.

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5724)

### DIFF
--- a/src/BloomExe/BloomExe.csproj
+++ b/src/BloomExe/BloomExe.csproj
@@ -416,6 +416,7 @@
     <Compile Include="History\HistoryEvent.cs" />
     <Compile Include="Utils\AsyncUtil.cs" />
     <Compile Include="Publish\PDF\PublishPdfApi.cs" />
+    <Compile Include="Utils\OutputFilenames.cs" />
     <Compile Include="WebView2Browser.cs" Condition="'$(OS)'=='Windows_NT'">
       <SubType>UserControl</SubType>
     </Compile>
--- a/src/BloomExe/Book/Book.cs
+++ b/src/BloomExe/Book/Book.cs
@@ -3806,7 +3806,7 @@ namespace Bloom.Book
 			return _bookData.GetMultiTextVariableOrEmpty(name);
 		}
 
-		internal IBookStorage Storage { get; }
+		public virtual IBookStorage Storage { get; }
 
 		/// <summary>
 		/// This gets called as a result of a UI action. It sets the new topic in our data,
--- a/src/BloomExe/Book/BookInfo.cs
+++ b/src/BloomExe/Book/BookInfo.cs
@@ -96,7 +96,7 @@ namespace Bloom.Book
 			None, OnPage	//Removed Links in Bloom 4.6 on June 28, 2019.
 		}
 
-		public string Id
+		public virtual string Id
 		{
 			get { return MetaData.Id; }
 			set { MetaData.Id = value; }
--- a/src/BloomExe/Book/BookStorage.cs
+++ b/src/BloomExe/Book/BookStorage.cs
@@ -144,7 +144,7 @@ namespace Bloom.Book
 		public string InitialLoadErrors { get; private set; }
 		public bool ErrorAllowsReporting { get; private set; }    // True if we want to display a Report to Bloom Support button
 
-		public BookInfo BookInfo
+		public virtual BookInfo BookInfo
 		{
 			get
 			{
@@ -153,6 +153,12 @@ namespace Bloom.Book
 				return _metaData;
 			}
 			set { _metaData = value; }
+		}
+
+		public BookStorage()
+		{
+			if (!Program.RunningUnitTests)
+				throw new ApplicationException("Parameterless BookStorage constructor is allowed only in unit tests!");
 		}
 
 		public BookStorage(string folderPath, IChangeableFileLocator baseFileLocator,
--- a/src/BloomExe/CollectionTab/CollectionModel.cs
+++ b/src/BloomExe/CollectionTab/CollectionModel.cs
@@ -282,20 +282,15 @@ namespace Bloom.CollectionTab
 
 		private Form MainShell => Application.OpenForms.Cast<Form>().FirstOrDefault(f => f is Shell);
 
-		private string _bloomPackFolder;
-
 		// Not entirely happy that this method which launches a dialog is in the Model.
 		// but with the Collection UI moving to JS I don't see a good alternative
 		public void MakeBloomPack(bool forReaderTools)
 		{
-			var folder = Environment.GetFolderPath(Environment.SpecialFolder.MyDocuments);
-			if (!string.IsNullOrEmpty(_bloomPackFolder) && Directory.Exists(_bloomPackFolder))
-				folder = _bloomPackFolder;
-			var initialPath = Path.Combine(folder, GetSuggestedBloomPackPath());
+			var initialPath = OutputFilenames.GetOutputFilePath(TheOneEditableCollection, ".BloomPack");
 			var destFileName = MiscUtils.GetOutputFilePathOutsideCollectionFolder(initialPath, "BloomPack files|*.BloomPack");
 			if (!string.IsNullOrEmpty(destFileName))
 			{
-				_bloomPackFolder = Path.GetDirectoryName(destFileName);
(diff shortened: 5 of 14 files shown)


JohnThomson (changes requested):
Be nice to handle them too, but it's already a nice step forward.

> Reviewed 14 of 14 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5724#-:-NRKChzo-8KnMPWOiUQr:b-iybjqv)* status: all files reviewed, 4 unresolved discussions (waiting on @StephenMcConnel)

___
*[`src/BloomExe/Utils/OutputFilenames.cs` line 44 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5724#-NRK6M8Y2ZIzPbeNPGeJ:-NRK6M8Y2ZIzPbeNPGeK:b-z9apl8) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/c304ed67f4a1ee3063bfc797e30866db6800a829/src/BloomExe/Utils/OutputFilenames.cs#L44)):*
> ```Smalltalk
> 			if (!String.IsNullOrEmpty(proposedName))
> 				return (Path.Combine(startingFolder, $"{proposedName}{extension}"));
> 			return Path.Combine(startingFolder, $"{Path.GetFileName(book.FolderPath)}{extension}");
> ```

Should we make some use of extraTag here if it is set? Or comment in header that if you're using it you should have a proposedName?
___
*[`src/BloomExe/Utils/OutputFilenames.cs` line 62 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5724#-NRK6jw060l9-X2G9lyK:-NRK6jw060l9-X2G9lyL:b-di20f3) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/c304ed67f4a1ee3063bfc797e30866db6800a829/src/BloomExe/Utils/OutputFilenames.cs#L62)):*
> ```Smalltalk
> 		public static void RememberOutputFilePath(Book.Book book, string extension, string outputFilePath, string extraTag = "")
> 		{
> 			_savedOutputFilePaths[Tuple.Create(book.ID, $"{extraTag}{extension}")] = outputFilePath;
> ```

Since it's important that they stay the same in the two methods, it might be worth capturing the way you make a key from the book, extraTag, and extension in a method.
___
*[`src/BloomTests/Utils/OutputFilenamesTests.cs` line 25 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5724#-NRK96Lf8va1AS-eiaua:-NRK96Lg02vwq7GlENcy:b-t548ee) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/c304ed67f4a1ee3063bfc797e30866db6800a829/src/BloomTests/Utils/OutputFilenamesTests.cs#L25)):*
> ```Smalltalk
> 			_mockBook1 = CreateMockBook("536ff975-1bd7-49ca-bcc8-937f8e4b8c9f", @"C:\Users\steve\Documents\Bloom\Plants");
> 			_mockBook2 = CreateMockBook("536ff975-1bd7-49ca-bcc8-937f8e4b8cA0", @"C:\Users\steve\Documents\Bloom\Animals");
> 			_mockBook3 = CreateMockBook("536ff975-1bd7-49ca-bcc8-937f8e4b8cA1", @"C:\Users\steve\Documents\Bloom\Minerals");
> ```

Isn't it a problem that for most computers C:\Users\steve won't exist and you probably don't have permission to create it? Feels like those paths should use _userDocumentFolder. If not, needs a comment.
Also, I don't know if this will ever make it to Linux, but if it does hard-coded backslashes there will probably mess up the string comparisons in the tests.
___
*[`src/BloomTests/Utils/OutputFilenamesTests.cs` line 92 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5724#-NRKBJ5Q4c1hc_n3s6as:-NRKBJ5Q4c1hc_n3s6at:b-1gzmsd) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/c304ed67f4a1ee3063bfc797e30866db6800a829/src/BloomTests/Utils/OutputFilenamesTests.cs#L92)):*
> ```Smalltalk
> 
> 		[Test]
> 		public void ComplexUsesWork()
> ```

This seems very similar to the preceding test. Can you comment or name to highlight what makes this one 'complex' and different?

andrew-polk (commented):
> Reviewed 6 of 14 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5724)* status: all files reviewed, 4 unresolved discussions (waiting on @StephenMcConnel)

StephenMcConnel (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5724)* status: all files reviewed, 4 unresolved discussions (waiting on @JohnThomson)

___
*[`src/BloomExe/Utils/OutputFilenames.cs` line 44 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5724#-NRK6M8Y2ZIzPbeNPGeJ:-NR_3TBq8wFZOXPvvJ_f:bitkp4h) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/c304ed67f4a1ee3063bfc797e30866db6800a829/src/BloomExe/Utils/OutputFilenames.cs#L44)):*

Should we make some use of extraTag here if it is set? Or comment in header that if you're using it you should have a proposedName?
</blockquote></details>

The latter, since filenames derived from titles in a specific language don't really need/want to be tagged with the language tag.
___
*[`src/BloomExe/Utils/OutputFilenames.cs` line 62 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5724#-NRK6jw060l9-X2G9lyK:-NR_65LqDHRJi1QhGwEt:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/c304ed67f4a1ee3063bfc797e30866db6800a829/src/BloomExe/Utils/OutputFilenames.cs#L62)):*

Since it's important that they stay the same in the two methods, it might be worth capturing the way you make a key from the book, extraTag, and extension in a method.
</blockquote></details>

Done.
___
*[`src/BloomTests/Utils/OutputFilenamesTests.cs` line 25 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5724#-NRK96Lf8va1AS-eiaua:-NR_7qK30fjVx7UjCFns:b-smdn6c) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/c304ed67f4a1ee3063bfc797e30866db6800a829/src/BloomTests/Utils/OutputFilenamesTests.cs#L25)):*

Isn't it a problem that for most computers C:\Users\steve won't exist and you probably don't have permission to create it? Feels like those paths should use _userDocumentFolder. If not, needs a comment.
Also, I don't know if this will ever make it to Linux, but if it does hard-coded backslashes there will probably mess up the string comparisons in the tests.
</blockquote></details>

The code being tested puts paths together, but never checks whether any of the paths actually exist.  This is true of the mocking object as well.  But that might not always be the case from now until Bloom disappears.
And I had thought about the Linux issue, but decided not to care since by the time Linux is reintroduced, there will be lots of places to fix.  But I suppose there's no reason not to do it portably to maintain good habits.
___
*[`src/BloomTests/Utils/OutputFilenamesTests.cs` line 92 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5724#-NRKBJ5Q4c1hc_n3s6as:-NR_A77DAxJcME4EVeTq:b5ogpc3) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/c304ed67f4a1ee3063bfc797e30866db6800a829/src/BloomTests/Utils/OutputFilenamesTests.cs#L92)):*

This seems very similar to the preceding test. Can you comment or name to highlight what makes this one 'complex' and different?
</blockquote></details>

I've added a comment at the beginning of each test method explain the overall basis of the test.
Basically this last test uses all three optional parameters whereas the previous test used only one (the folder) and the first test used none.

JohnThomson (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5724)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @StephenMcConnel)

PR #5733, opened 2023-03-29
Add print booklet settings helps as text instead of images (BL-11953)

StephenMcConnel:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5733)


StephenMcConnel (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5733)* status: all files reviewed, 4 unresolved discussions (waiting on @andrew-polk)

___
*[`src/BloomExe/Publish/PDF/PublishPdfApi.cs` line 110 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5733#-NRj7fV8AdNQgAbD-21q:-NRn6FKD-F-CRgFXYMLm:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/4c11291883643f499bc373f16882abec15527122/src/BloomExe/Publish/PDF/PublishPdfApi.cs#L110)):*

I think this would be significantly cleaner if the switch set a few bools such as `isLandscape`, `isLongEdge`, `isCutAndStack`, etc., and then, after the switch, you set the messages based on those bools.
What do you think?
</blockquote></details>

Done.
___
*[`src/BloomExe/Publish/PDF/PublishPdfApi.cs` line 176 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5733#-NRj5Pf5CUCdFjrphO9v:-NRj9UPNEsg2OBCoyqgx:b-eijq0n) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/4c11291883643f499bc373f16882abec15527122/src/BloomExe/Publish/PDF/PublishPdfApi.cs#L176)):*

Do we need this note? 
If you think it is worth keeping, I think we should we at least remove the i18n.
</blockquote></details>

We shouldn't need this note, but the rules for enabling the booklet buttons have been too loose ever since we started adding more page sizes.
I've removed the localization.  Users shouldn't see this message if we get the enabling set correctly.
___
*[`src/BloomExe/Publish/PDF/PublishPdfApi.cs` line 190 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5733#-NRj6bPy-Y44UsGuupf8:-NRj9Oqd2ZeQxSxM0xRf:b-g02ist) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/4c11291883643f499bc373f16882abec15527122/src/BloomExe/Publish/PDF/PublishPdfApi.cs#L190)):*

If you use dynamic, I think it will save you all the escaping, etc. You won't have to build the json string, just an object.

I don't think you need to conditionally add the note. Just let it be empty string if it is not set (if null is not ok).
</blockquote></details>

I tried using dynamic earlier and failed.  I think I got it working now.

andrew-polk (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5733)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @StephenMcConnel)

___
*[`src/BloomExe/Publish/PDF/PublishPdfApi.cs` line 176 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5733#-NRj5Pf5CUCdFjrphO9v:-NRn9oPa5SgjDnLeGxYR:bppeonf) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/4c11291883643f499bc373f16882abec15527122/src/BloomExe/Publish/PDF/PublishPdfApi.cs#L176)):*

We shouldn't need this note, but the rules for enabling the booklet buttons have been too loose ever since we started adding more page sizes.
I've removed the localization.  Users shouldn't see this message if we get the enabling set correctly.
</blockquote></details>

Thanks. Do we have a card for that?

PR #5741, opened 2023-04-01
Show progress dialog for long-running MakeSelectedBook ops (BL-12033)

jsubloom:
Commit 1: Refactor ProgressDialog to be easier to consume
Since ProgressDialog doesn't expose the event args to its consumers, they might as well take it in via a simple scalar prop which is easier to comprehend than having to useRef functions which generate functions and invoke them at the right time

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5741)

### DIFF
--- a/DistFiles/localization/en/Bloom.xlf
+++ b/DistFiles/localization/en/Bloom.xlf
@@ -1041,6 +1041,12 @@
         <note>ID: Common.Loading</note>
         <note>This is shown when Bloom is slowly loading something, so the user doesn't worry about why they don't see the result immediately.</note>
       </trans-unit>
+      <trans-unit id="Common.LongRunningOperation" sil:dynamic="true">
+        <!-- added in 5.5 -->
+        <source xml:lang="en">This may take a while...</source>
+        <note>ID: Common.LongRunningOperation</note>
+        <note>Used to indicate that an operation may take a long amount of time to complete.</note>
+      </trans-unit>
       <trans-unit id="Common.Next">
         <source xml:lang="en">&amp;Next</source>
         <note>ID: Common.Next</note>
--- a/DistFiles/localization/en/Bloom.xlf
+++ b/DistFiles/localization/en/Bloom.xlf
@@ -1041,7 +1041,7 @@
         <note>ID: Common.Loading</note>
         <note>This is shown when Bloom is slowly loading something, so the user doesn't worry about why they don't see the result immediately.</note>
       </trans-unit>
-      <trans-unit id="Common.LongRunningOperation" sil:dynamic="true">
+      <trans-unit id="Common.LongRunningOperation" sil:dynamic="true" translate="no">
         <!-- added in 5.5 -->
         <source xml:lang="en">This may take a while...</source>
         <note>ID: Common.LongRunningOperation</note>
--- a/src/BloomBrowserUI/collectionsTab/collectionsTabBookPane/CollectionsTabBookPane.tsx
+++ b/src/BloomBrowserUI/collectionsTab/collectionsTabBookPane/CollectionsTabBookPane.tsx
@@ -2,7 +2,7 @@
 import { jsx, css } from "@emotion/react";
 import * as React from "react";
 import { useRef, useState } from "react";
-import { get, getBoolean } from "../../utils/bloomApi";
+import { get, getBoolean, postThatMightNavigate } from "../../utils/bloomApi";
 import { TeamCollectionBookStatusPanel } from "../../teamCollection/TeamCollectionBookStatusPanel";
 import {
     IBookTeamCollectionStatus,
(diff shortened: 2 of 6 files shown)


jsubloom (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5741)* status: 0 of 3 files reviewed, 1 unresolved discussion

___
*[`src/BloomBrowserUI/react_components/BloomDialog/BloomDialogPlumbing.tsx` line 87 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5741#-NRxT0A9-NXJtrMzxd4i:-NRxT0A9-NXJtrMzxd4j:bqnnl47) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/65025a1afc5e86501337654dba235b38a1c91b35/src/BloomBrowserUI/react_components/BloomDialog/BloomDialogPlumbing.tsx#L87)):*
> ```XML
>     const [event, setEvent] = useState<any>({});
> 
>     // UseCallback ensures a stable reference, so that consumers of this hook don't get a new reference each time
> ```

Ok, there are actually a ton of consumers of this, but I think this change should be safe as far as I can reason about it. Instead of getting a new function reference (with the same implementation) each time, the consumers may get the same reference returned.  I can think of reasons why it'd break if it went from stable -> new ref each time, but I can't think of any realistic scenario that would care if it went from new ref each time -> stable.  Like, if the consumers worked with ever-changing references each time and weren't bothered by it, they shouldn't be bothered by having a stable ref each time.

JohnThomson (approved):
I like the simplification of the progress dialog props very much, assuming all the callers of the deleted props have been fixed. Just a couple of minor points to think about, but if all is well I'm OK with merging it.

> Reviewed 2 of 3 files at r1, 4 of 4 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5741#-:-NS7wmnc6Xdz4xrVmcG1:bsb8e6v)* status: all files reviewed, 3 unresolved discussions (waiting on @jsubloom)

___
*[`src/BloomBrowserUI/collectionsTab/collectionsTabBookPane/CollectionsTabBookPane.tsx` line 184 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5741#-NS7iXJX50a6Rf_ZO0mr:-NS7iXJX50a6Rf_ZO0ms:b-y9f2sg) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/8b2fa31106ebd2f840262516fc8d8547254978bc/src/BloomBrowserUI/collectionsTab/collectionsTabBookPane/CollectionsTabBookPane.tsx#L184)):*
> ```XML
>     // An easy way out is to use the pattern of mounting the ProgressDialog only when needed.
>     // That way we'll have a fresh instance each time that hasn't had a bunch of messages pumped to it already.
>     const editOrMakeProgress = progressOpen ? (
> ```

The usual React syntax here is to use && instead of question mark and drop everything from the colon on.
___
*[`src/BloomBrowserUI/react_components/Progress/ProgressDialog.tsx` line 67 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5741#-NS7wc0u9dchZHmCjL1R:-NS7wc0u9dchZHmCjL1S:bch3fjq) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/6e736b197481aa496f23169700be840bb2a65b27/src/BloomBrowserUI/react_components/Progress/ProgressDialog.tsx#L67)):*
> ```XML
>     // <ProgressDialog ... setShowDialog={showFunc => show.current = showFunc}...>
>     setShowDialog?: (show: () => void) => void;
>     setCloseDialog?: (close: () => void) => void;
> ```

It surprises me that there aren't more callers of these functions. Did you look to see if there are any C# callers (e.g., using RunJavascript)?

jsubloom (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5741)* status: 5 of 6 files reviewed, 3 unresolved discussions (waiting on @JohnThomson)

___
*[`src/BloomBrowserUI/collectionsTab/collectionsTabBookPane/CollectionsTabBookPane.tsx` line 184 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5741#-NS7iXJX50a6Rf_ZO0mr:-NS7zDCJ3A5wxziCOYIW:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/8b2fa31106ebd2f840262516fc8d8547254978bc/src/BloomBrowserUI/collectionsTab/collectionsTabBookPane/CollectionsTabBookPane.tsx#L184)):*

The usual React syntax here is to use && instead of question mark and drop everything from the colon on.
</blockquote></details>

Done.
___
*[`src/BloomBrowserUI/react_components/Progress/ProgressDialog.tsx` line 67 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5741#-NS7wc0u9dchZHmCjL1R:-NS7zSVYDCgbtafyO4jL:bfirear) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/6e736b197481aa496f23169700be840bb2a65b27/src/BloomBrowserUI/react_components/Progress/ProgressDialog.tsx#L67)):*

It surprises me that there aren't more callers of these functions. Did you look to see if there are any C# callers (e.g., using RunJavascript)?
</blockquote></details>

Hmm, searching for setShowDialog and setCloseDialog in the C# both returned 0 results. 

I do see there is some Async-named functions there, and I do see that some things send "open-progress" or "close-progress" over the socket server.
But the method removed is about saving a function as pseudo-state in react. It doesn't actually open or close the dialog directly.

PR #5769, opened 2023-04-18
Disable summary field if book is not checked out (BL-12124)

andrew-polk:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5769)

### DIFF
--- a/src/BloomBrowserUI/app/SelectedBookContext.tsx
+++ b/src/BloomBrowserUI/app/SelectedBookContext.tsx
@@ -1,0 +1,15 @@
+import React = require("react");
+import {
+    DefaultSelectedBookInfo,
+    ISelectedBookInfo,
+    useMonitorBookSelection
+} from "./selectedBook";
+
+export const SelectedBookContext = React.createContext<ISelectedBookInfo>(
+    DefaultSelectedBookInfo
+);
+
+export const SelectedBookProvider: React.FunctionComponent = props => {
+    const selectedBookInfo = useMonitorBookSelection();
+    return <SelectedBookContext.Provider value={selectedBookInfo} {...props} />;
+};
--- a/src/BloomBrowserUI/app/selectedBook.ts
+++ b/src/BloomBrowserUI/app/selectedBook.ts
@@ -10,19 +10,21 @@ export interface ISelectedBookInfo {
     isTemplate: boolean;
 }
 
+export const DefaultSelectedBookInfo: ISelectedBookInfo = {
+    id: undefined,
+    saveable: false,
+    collectionKind: "error", // better to see no button at all until we know which it is
+    aboutBookInfoUrl: undefined,
+    isTemplate: false
+};
+
 // Anything that uses this will always have the current book info. The first render will see the default
 // (nothing selected); immediately after, when we get the info from our server, another render will
 // see the then-current information, and any time the selection changes another render will
 // yield the new information, based on this code monitoring a websocket.
 export function useMonitorBookSelection(): ISelectedBookInfo {
     const [selectedBookInfo, setSelectedBookInfo] = useState<ISelectedBookInfo>(
-        {
-            id: undefined,
-            saveable: false,
-            collectionKind: "error", // better to see no button at all until we know which it is
-            aboutBookInfoUrl: undefined,
-            isTemplate: false
-        }
+        DefaultSelectedBookInfo
     );
 
     // At this point, even a change of TC status on a book will cause Bloom to fire
--- a/src/BloomBrowserUI/publish/LibraryPublish/LibraryPublishSteps.tsx
+++ b/src/BloomBrowserUI/publish/LibraryPublish/LibraryPublishSteps.tsx
@@ -43,6 +43,8 @@ import {
     UploadCollisionDlg
 } from "./uploadCollisionDlg";
 import { useGetEnterpriseBookshelves } from "../../collection/useGetEnterpriseBookshelves";
+import { MustBeCheckedOut } from "../../react_components/MustBeCheckedOut";
+import { SelectedBookContext } from "../../app/SelectedBookContext";
 
 interface IReadonlyBookInfo {
     title: string;
@@ -59,6 +61,7 @@ const kWebSocketEventId_uploadCanceled: string = "uploadCanceled";
 const kWebSocketEventId_loginSuccessful: string = "loginSuccessful";
 
 export const LibraryPublishSteps: React.FunctionComponent = () => {
+    const selectedBookContext = React.useContext(SelectedBookContext);
     const [bookshelfHasProblem, setBookshelfHasProblem] = useState(false);
     const {
         project,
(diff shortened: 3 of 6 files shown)


hatton (commented):
> Reviewed 1 of 2 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5769)* status: 1 of 2 files reviewed, 1 unresolved discussion (waiting on @andrew-polk)

___
*[`src/BloomBrowserUI/publish/LibraryPublish/LibraryPublishSteps.tsx` line 372 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5769#-NTLEnTY742fGWYbONXe:-NTLEnTY742fGWYbONXf:b-jmmd5m) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/910984aada97d883a35f8b7aa7fcbfd79ef633ea/src/BloomBrowserUI/publish/LibraryPublish/LibraryPublishSteps.tsx#L372)):*
> ```XML
> 
>     const summaryWithPossibleTooltip = isSummaryDisabled ? (
>         <BloomTooltip
> ```

Is there an opportunity to factor this out to a common component?
```
<MustBeCheckedOut>

</>
```

jsubloom (changes requested):
> Reviewed 2 of 2 files at r1.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5769)* status: all files reviewed, 4 unresolved discussions (waiting on @andrew-polk)

___
*[`src/BloomBrowserUI/publish/LibraryPublish/LibraryPublishSteps.tsx` line 55 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5769#-NTLGGer1_BfqtLvn0AT:-NTLGGer1_BfqtLvn0AU:bsr3the) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/910984aada97d883a35f8b7aa7fcbfd79ef633ea/src/BloomBrowserUI/publish/LibraryPublish/LibraryPublishSteps.tsx#L55)):*
> ```XML
>     licenseRights: string;
>     isTemplate: boolean;
>     isSaveable: boolean;
> ```

I was wondering if the spelling was more isSavable. Apparently both spellings are used. Google Trends currently indicates that "saveable" is more common than "savable" in only one state: Florida. Interesting? :)
___
*[`src/BloomBrowserUI/publish/LibraryPublish/LibraryPublishSteps.tsx` line 325 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5769#-NTLGU8T85Srz0x7BFQ8:-NTLGU8U4c0t7XWobQmJ:bgvaakx) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/910984aada97d883a35f8b7aa7fcbfd79ef633ea/src/BloomBrowserUI/publish/LibraryPublish/LibraryPublishSteps.tsx#L325)):*
> ```XML
>     ]);
> 
>     const isSummaryDisabled = bookInfo === undefined || !bookInfo?.isSaveable;
> ```

If you include the check for bookInfo === undefined, then you can get rid of the ? after bookInfo. I don't think it's necessary, and I think the question marks make it more confusing to read/parse so I prefer to get rid of unnecessary ones.

Alternatively, keep the question mark and delete the first condition.
___
*[`src/BloomBrowserUI/publish/LibraryPublish/LibraryPublishSteps.tsx` line 373 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5769#-NTLGq6B4UbkzxIXDFmp:-NTLGq6B4UbkzxIXDFmq:bmfpev8) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/910984aada97d883a35f8b7aa7fcbfd79ef633ea/src/BloomBrowserUI/publish/LibraryPublish/LibraryPublishSteps.tsx#L373)):*
> ```XML
>     const summaryWithPossibleTooltip = isSummaryDisabled ? (
>         <BloomTooltip
>             id={"book-summary"}
> ```

The id only differs from the other component by hyphen or not.  I assume the id's should indeed be unique. If so, would it be better to be unique in a more meaningful way... e.g. "book summary" vs "book summary tooltip"?
Unless something extenuating sitaution like "accessibility readers are going to read the ID out loud" applies...

jsubloom (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5769)* status: 1 of 4 files reviewed, 3 unresolved discussions (waiting on @andrew-polk and @hatton)

___
*[`src/BloomBrowserUI/publish/LibraryPublish/LibraryPublishSteps.tsx` line 373 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5769#-NTOwuJW3xYnEKnWUNPC:-NTOwuJW3xYnEKnWUNPD:bdc6i2l) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/c7570560571c6c17e730b58c0265e86832e5f509/src/BloomBrowserUI/publish/LibraryPublish/LibraryPublishSteps.tsx#L373)):*
> ```XML
>                         <MustBeCheckedOut
>                             id={"book-summary-not-checked-out-tooltip"}
>                             isCheckedOut={bookInfo?.isSaveable || false}
> ```

Optional: I'd a fan of {bookInfo?.isSaveable === true} , but no need to change this.
___
*[`src/BloomBrowserUI/react_components/MustBeCheckedOut.tsx` line 4 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5769#-NTOvfU4CYKJoTnaHDQL:-NTOvfU4CYKJoTnaHDQM:b-qx0c25) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/c7570560571c6c17e730b58c0265e86832e5f509/src/BloomBrowserUI/react_components/MustBeCheckedOut.tsx#L4)):*
> ```XML
> import { BloomTooltip, IBloomToolTipProps } from "./BloomToolTip";
> 
> export const MustBeCheckedOut: React.FunctionComponent<{
> ```

This will add another problematic case whenever we migrate to React 18, because it uses {children} but doesn't explicitly define it. (It uses the old-style implicit children... which is no longer allowed in React 18)

I think you should wrap it in PropsWithChildren. "React.F...C...<PropsWithChildren<{...} & IBloomToolTipProps>> " Except... I think that is only introduced in react 18? My VSCode won't autocomplete it for me, at least. Sigh.

you could probably manually add "children?: ReactNode" or something like that to the props.  Or "& { children?: ReactNode} " Something like that?
IDK, the type definition line of function component props always makes my head hurt...

Or you could punt it to our future selves, if you can't figure out anything clean here.

andrew-polk (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5769)* status: 0 of 6 files reviewed, 2 unresolved discussions (waiting on @hatton and @jsubloom)

___
*[`src/BloomBrowserUI/publish/LibraryPublish/LibraryPublishSteps.tsx` line 372 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5769#-NTLEnTY742fGWYbONXe:-NTQYUwFApTv5lvrT7Rg:b5r78vp) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/910984aada97d883a35f8b7aa7fcbfd79ef633ea/src/BloomBrowserUI/publish/LibraryPublish/LibraryPublishSteps.tsx#L372)):*

How about a react Context thing for books?
</blockquote></details>

Done.
As we discussed, the other instance, in menu items, still doesn't lend itself well to this.
___
*[`src/BloomBrowserUI/react_components/MustBeCheckedOut.tsx` line 4 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5769#-NTOvfU4CYKJoTnaHDQL:-NTQZ4NrBWQa4tEViw0o:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/c7570560571c6c17e730b58c0265e86832e5f509/src/BloomBrowserUI/react_components/MustBeCheckedOut.tsx#L4)):*

This will add another problematic case whenever we migrate to React 18, because it uses {children} but doesn't explicitly define it. (It uses the old-style implicit children... which is no longer allowed in React 18)

I think you should wrap it in PropsWithChildren. "React.F...C...<PropsWithChildren<{...} & IBloomToolTipProps>> " Except... I think that is only introduced in react 18? My VSCode won't autocomplete it for me, at least. Sigh.

you could probably manually add "children?: ReactNode" or something like that to the props.  Or "& { children?: ReactNode} " Something like that?
IDK, the type definition line of function component props always makes my head hurt...

Or you could punt it to our future selves, if you can't figure out anything clean here.


</blockquote></details>

Done.

jsubloom (changes requested):
> Reviewed 2 of 5 files at r3.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5769)* status: 2 of 6 files reviewed, 2 unresolved discussions (waiting on @andrew-polk and @hatton)

___
*[`src/BloomBrowserUI/app/SelectedBookContext.tsx` line 5 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5769#-NTTuV9TEcY2PmkIVFTP:-NTTuV9TEcY2PmkIVFTQ:b9a4b5o) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/a312c7a5203fde129c41a4bce0e1a17a53d8dd32/src/BloomBrowserUI/app/SelectedBookContext.tsx#L5)):*
> ```XML
> 
> export const SelectedBookContext = React.createContext<ISelectedBookInfo>(
>     {} as ISelectedBookInfo
> ```

I'm not a fan of "as"-ing stuff in Typescript. Misusing "as" means that everywhere else in the code we can't actually trust if the types are true, which makes me feel like we have the worst of both worlds - we both have to make a strongly typed typechecker happy, and we don't even get type safety!

I see if I look up what ISelectedBookInfo is that {} is not actually an ISelectedBookInfo.

So, I assume that right now, since we're working on the Publish tab, we can assume that SelectedBookContext will always be initialized to something valid.

However, since we're making a Context/Provider... that's not generally true in all places that could ask for the context (anywhere, right?). In addition to startup, we also have an actual valid state that a book might not be selected.

Since it's perfectly valid for a book not to be selected, I'm thinking that the type of this should actually be ISelectedBookInfo | null.    (Or ISelectedBookInfo | undefined if you prefer). And it should be initialized to null or undefined.

Consumers should check if it's truthy or not.
In the case of the Publish tab, where it doesn't make sense for it to be non-selectable, I am fine if it uses a "!" or explicitly throws an exception if it is null so that the processing of the falsy case can be as minimal as possible.
___
*[`src/BloomBrowserUI/react_components/MustBeCheckedOut.tsx` line 4 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5769#-NTOvfU4CYKJoTnaHDQL:-NTTuRZ2EFb8_6vu7wMJ:b-6mhyx0) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/c7570560571c6c17e730b58c0265e86832e5f509/src/BloomBrowserUI/react_components/MustBeCheckedOut.tsx#L4)):*

Done.
</blockquote></details>

Great, thanks!

jsubloom (approved):
> Reviewed 1 of 3 files at r2, 2 of 5 files at r3.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5769)* status: all files reviewed, 1 unresolved discussion (waiting on @hatton)

andrew-polk (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5769)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @andrew-polk)

PR #5777, opened 2023-04-21
BL-10638 Reset bookshelf when switching status

gmartin7:
* handles the case of down-grading to Local Community or
   None Enterprise status
* disable bookshelf control if the only option is "none"

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5777)

### DIFF
--- a/src/BloomExe/web/controllers/CollectionSettingsApi.cs
+++ b/src/BloomExe/web/controllers/CollectionSettingsApi.cs
@@ -90,9 +90,11 @@ namespace Bloom.web.controllers
 					if (_enterpriseStatus == EnterpriseStatus.None)
 					{
 						_knownBrandingInSubscriptionCode = true;
+						ResetBookshelf();
 						BrandingChangeHandler("Default", null);
 					} else if (_enterpriseStatus == EnterpriseStatus.Community)
 					{
+						ResetBookshelf();
 						BrandingChangeHandler("Local-Community", null);
 					}
 					else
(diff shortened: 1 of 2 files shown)


src/BloomExe/web/controllers/CollectionSettingsApi.cs
```diff
@@ -266,6 +268,12 @@ public void RegisterWithApiHandler(BloomApiHandler apiHandler)
 			apiHandler.RegisterEndpointHandler(kApiUrlPart + "addCustomPaletteColor", HandleAddCustomColor, false);
 		}
 
+		private void ResetBookshelf()
```
andrew-polk, line 271: I think we also want to use this new method on line 128? (and you can remove the null check there)

gmartin7 (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5777)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @andrew-polk)

___
*[`src/BloomExe/web/controllers/CollectionSettingsApi.cs` line 271 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5777#gh-1174067382:-NT_LX5pCLBcA9hn1zgC:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/28f67e33ffab7afdde0f37b85d48802c173f21a7/src/BloomExe/web/controllers/CollectionSettingsApi.cs#L271)):*

I think we also want to use this new method on line 128? (and you can remove the null check there)
</blockquote></details>

Done.

andrew-polk (approved):
> Reviewed 1 of 2 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5777)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @gmartin7)

PR #5782, opened 2023-04-24
Start splash screen sooner for installs to reassure users (BL-12085)

StephenMcConnel:
Also make developing/debugging a bit more pleasant.

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5782)

### DIFF
--- a/src/BloomExe/Program.cs
+++ b/src/BloomExe/Program.cs
@@ -403,6 +403,14 @@ namespace Bloom
 						}
 						gotUniqueToken = true;
 					}
+					if (IsInstallerLaunch(args))
+					{
+						// Start the splash screen early for installer launches to reassure
+						// user while localization and other one-time setup is happening.  For
+						// installer launches, no more dialogs should be popping up after this.
+						// See https://issues.bloomlibrary.org/youtrack/issue/BL-12085.
+						StartupScreenManager.StartManaging();
+					}
 					OldVersionCheck();
 
 					SetUpErrorHandling();
@@ -497,7 +505,7 @@ namespace Bloom
 						_originalPreload = Environment.GetEnvironmentVariable("LD_PRELOAD");
 						Environment.SetEnvironmentVariable("LD_PRELOAD", null);
 
-						Run();
+						Run(args);
 					}
 				}
 			}
@@ -731,9 +739,10 @@ namespace Bloom
 		}
 
 		[HandleProcessCorruptedStateExceptions]
-		private static void Run()
+		private static void Run(string[] args)
 		{
-			StartupScreenManager.StartManaging();
+			if (!IsInstallerLaunch(args))
+				StartupScreenManager.StartManaging();
 			
 			Settings.Default.Save();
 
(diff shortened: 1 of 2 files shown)


andrew-polk (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5782)* status: all files reviewed, 1 unresolved discussion (waiting on @StephenMcConnel)

___
*[`src/BloomExe/UpdateVersionTable.cs` line 140 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5782#-NTpLWCRCGusb17yalrY:-NTpLWCRCGusb17yalrZ:bqf1xs7) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/ba20ff9ff3a85f13ccfbb7eef2844b6969d7873d/src/BloomExe/UpdateVersionTable.cs#L140)):*
> ```Smalltalk
> 						msg = "Internet connection did not allow check for update.";
> 						NonFatalProblem.Report(ModalIf.Alpha, PassiveIf.All, msg); // hopefully this will just 'toast'
> 					}
> ```

If we have an error during development, why don't we want to find out here?

(But I thought we didn't even allow the check to get this far if we are debugging.)

StephenMcConnel (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5782)* status: all files reviewed, 1 unresolved discussion (waiting on @andrew-polk)

___
*[`src/BloomExe/UpdateVersionTable.cs` line 140 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5782#-NTpLWCRCGusb17yalrY:-NTpPkj6BbBB_ssn-Yr3:biawxze) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/ba20ff9ff3a85f13ccfbb7eef2844b6969d7873d/src/BloomExe/UpdateVersionTable.cs#L140)):*

If we have an error during development, why don't we want to find out here?

(But I thought we didn't even allow the check to get this far if we are debugging.)
</blockquote></details>

This code is checking the update channel.  For a debug build on a developer machine, the "channel" is something like "Debug/Developer" which never has a valid update, and therefore always throws an exception resulting in an annoying error message box being displayed.  I just got tired of seeing it every time I tested Bloom with "--squirrel-whatever"  to simulate an installer run.

andrew-polk (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5782)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @StephenMcConnel)

___
*[`src/BloomExe/UpdateVersionTable.cs` line 140 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5782#-NTpLWCRCGusb17yalrY:-NTpU0soAdx3KUYvMRoj:bvts86u) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/ba20ff9ff3a85f13ccfbb7eef2844b6969d7873d/src/BloomExe/UpdateVersionTable.cs#L140)):*

This code is checking the update channel.  For a debug build on a developer machine, the "channel" is something like "Debug/Developer" which never has a valid update, and therefore always throws an exception resulting in an annoying error message box being displayed.  I just got tired of seeing it every time I tested Bloom with "--squirrel-whatever"  to simulate an installer run.
</blockquote></details>

Ok, there's some circumstance where we pop up a message box saying you can't check when debugging, but I guess that path isn't it.

PR #5785, opened 2023-04-26
Remove as much cruft as possible from PublishView (BL-12126)

StephenMcConnel:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5785)

### DIFF
--- a/build/getDependencies-windows.sh
+++ b/build/getDependencies-windows.sh
@@ -98,28 +98,21 @@ cd -
 #     clean: false
 #     revision: latest.lastSuccessful
 #     paths: {"*.chm"=>"DistFiles"}
-# [4] build: pdf.js (bt401)
-#     project: BuildTasks
-#     URL: https://build.palaso.org/viewType.html?buildTypeId=bt401
-#     clean: false
-#     revision: latest.lastSuccessful
-#     paths: {"pdfjs-viewer.zip!**"=>"DistFiles/pdf"}
-#     VCS: https://github.com/mozilla/pdf.js.git [gh-pages]
-# [5] build: PdfDroplet-Win-master-Continuous (bt54)
+# [4] build: PdfDroplet-Win-master-Continuous (bt54)
 #     project: PdfDroplet
 #     URL: https://build.palaso.org/viewType.html?buildTypeId=bt54
 #     clean: false
 #     revision: latest.lastSuccessful
 #     paths: {"PdfDroplet.exe"=>"lib/dotnet", "PdfSharp.dll"=>"lib/dotnet"}
 #     VCS: https://github.com/sillsdev/pdfDroplet [master]
-# [6] build: TidyManaged-master-win32-continuous (bt349)
+# [5] build: TidyManaged-master-win32-continuous (bt349)
 #     project: TidyManaged
 #     URL: https://build.palaso.org/viewType.html?buildTypeId=bt349
 #     clean: false
 #     revision: latest.lastSuccessful
 #     paths: {"*.*"=>"lib/dotnet"}
 #     VCS: https://github.com/BloomBooks/TidyManaged.git [master]
-# [7] build: Windows master continuous (XliffForHtml_WindowsMasterContinuous)
+# [6] build: Windows master continuous (XliffForHtml_WindowsMasterContinuous)
 #     project: XliffForHtml
 #     URL: https://build.palaso.org/viewType.html?buildTypeId=XliffForHtml_WindowsMasterContinuous
 #     clean: false
@@ -130,7 +123,6 @@ cd -
 # make sure output directories exist
 mkdir -p ../DistFiles
 mkdir -p ../DistFiles/ghostscript
-mkdir -p ../DistFiles/pdf
 mkdir -p ../Downloads
 mkdir -p ../build
 mkdir -p ../lib
@@ -189,7 +181,6 @@ copy_auto https://build.palaso.org/guestAuth/repository/download/Bloom_Squirrel/
 copy_auto https://build.palaso.org/guestAuth/repository/download/Bloom_Squirrel/latest.lastSuccessful/squirrel.windows.props ../lib/dotnet/squirrel.windows.props
 copy_auto https://build.palaso.org/guestAuth/repository/download/Bloom_Squirrel/latest.lastSuccessful/update.com ../lib/dotnet/update.com
 copy_auto https://build.palaso.org/guestAuth/repository/download/Bloom_Help_BloomHelp55/latest.lastSuccessful/Bloom.chm ../DistFiles/Bloom.chm
-copy_auto https://build.palaso.org/guestAuth/repository/download/bt401/latest.lastSuccessful/pdfjs-viewer.zip ../Downloads/pdfjs-viewer.zip
 copy_auto https://build.palaso.org/guestAuth/repository/download/bt54/latest.lastSuccessful/PdfDroplet.exe ../lib/dotnet/PdfDroplet.exe
 copy_auto https://build.palaso.org/guestAuth/repository/download/bt54/latest.lastSuccessful/PdfSharp.dll ../lib/dotnet/PdfSharp.dll
 copy_auto https://build.palaso.org/guestAuth/repository/download/bt349/latest.lastSuccessful/TidyManaged.dll ../lib/dotnet/TidyManaged.dll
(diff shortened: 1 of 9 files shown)


andrew-polk (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5785)* status: all files reviewed, 3 unresolved discussions (waiting on @StephenMcConnel)

___
*[`src/BloomExe/Publish/PublishModel.cs` line 206 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5785#-NTyViyF5jUq4OsgyoB1:-NTyViyF5jUq4OsgyoB2:b-jlgqy0) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/26b706f51f0f806af3dd01102507defbff82dbd7/src/BloomExe/Publish/PublishModel.cs#L206)):*
> ```Smalltalk
> 		{
> 			if (!e.Cancelled && e.Result is Exception)
> 				ReportPdfGenerationError(e.Result as Exception);
> ```

The original PublishView code called `UpdateDisplayMode()` (both for the error and normal case). Do we not need that anymore?

I see that for the PublishPdfApi, it sets PdfGenerationSucceeeded which, in turn, calls UpdateDisplayFeatures (the new name), but the PDF generation here doesn't set that. Maybe the UI doesn't need updating for this case?
___
*[`src/BloomExe/Publish/PublishView.cs` line 400 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5785#-NTyX8Fb8x2RpCmPq35_:-NTyX8FcCRN1JS80gWeY:b-e9zbp0) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/2a2a1c7199e801167fb27e322be8b57eb9cd4bc8/src/BloomExe/Publish/PublishView.cs#L400)):*
> ```Smalltalk
> 		}
> 
> 		private void ShowCantPublishMessage()
> ```

I'm not understanding why we don't need this can't-publish code anymore. This display mode can still get set above. And as far as I know, we haven't reimplemented this UI yet.
___
*[`src/BloomExe/WebLibraryIntegration/BulkUploader.cs` line 343 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5785#-NTyTDeT-aqWEF700eu_:-NTyTDeT-aqWEF700eua:bwdck7y) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/26b706f51f0f806af3dd01102507defbff82dbd7/src/BloomExe/WebLibraryIntegration/BulkUploader.cs#L343)):*
> ```Smalltalk
> 
> 			var publishModel = new PublishModel(bookSelection, new PdfMaker(), currentEditableCollectionSelection, context.Settings, server, _thumbnailer);
> 			var view = new PublishView(publishModel, new SelectedTabChangedEvent(), new LocalizationChangedEvent(), _singleBookUploader, null, null, null, null);
> ```

Looks like you can get rid of this `view` now.

StephenMcConnel (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5785)* status: 8 of 9 files reviewed, 3 unresolved discussions (waiting on @andrew-polk)

___
*[`src/BloomExe/Publish/PublishModel.cs` line 206 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5785#-NTyViyF5jUq4OsgyoB1:-NTydYi1AbsHGuhluG4r:b63hct3) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/26b706f51f0f806af3dd01102507defbff82dbd7/src/BloomExe/Publish/PublishModel.cs#L206)):*

The original PublishView code called `UpdateDisplayMode()` (both for the error and normal case). Do we not need that anymore?

I see that for the PublishPdfApi, it sets PdfGenerationSucceeeded which, in turn, calls UpdateDisplayFeatures (the new name), but the PDF generation here doesn't set that. Maybe the UI doesn't need updating for this case?
</blockquote></details>

The UI needs to be updated only for the PDF generation in PublishPdfApi, which is what the UI uses to generate the PDF.  The method for generating PDFs in PublishModel is used for the web upload (either individual book or bulk upload), and doesn't change the UI (apart from the progress reporting).
___
*[`src/BloomExe/Publish/PublishView.cs` line 400 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5785#-NTyX8Fb8x2RpCmPq35_:-NTye-Rr6ST2hZfbmX1z:bijlbtf) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/2a2a1c7199e801167fb27e322be8b57eb9cd4bc8/src/BloomExe/Publish/PublishView.cs#L400)):*

I'm not understanding why we don't need this can't-publish code anymore. This display mode can still get set above. And as far as I know, we haven't reimplemented this UI yet.
</blockquote></details>

All of the buttons are disabled when we can't publish.  Once the buttons are reimplemented in React/typescript/web technology, we should be able to attach a tooltip that explains what the button does (when enabled) or why it's disabled (when disabled).  The current tooltip technology used in C# doesn't show anything when the buttons are disabled, but that shouldn't be a problem in typescript.
___
*[`src/BloomExe/WebLibraryIntegration/BulkUploader.cs` line 343 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5785#-NTyTDeT-aqWEF700eu_:-NTyedFw-EaxEApxwhAC:bne95z2) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/26b706f51f0f806af3dd01102507defbff82dbd7/src/BloomExe/WebLibraryIntegration/BulkUploader.cs#L343)):*

Looks like you can get rid of this `view` now.
</blockquote></details>

Good catch.  That thought had crossed my mind after I submitted the PR.

andrew-polk (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5785)* status: all files reviewed, 2 unresolved discussions (waiting on @hatton)

___
*[`src/BloomExe/Publish/PublishView.cs` line 400 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5785#-NTyX8Fb8x2RpCmPq35_:-NTyhYPw5JTFiqzGe-Oz:bgkd911) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/2a2a1c7199e801167fb27e322be8b57eb9cd4bc8/src/BloomExe/Publish/PublishView.cs#L400)):*

All of the buttons are disabled when we can't publish.  Once the buttons are reimplemented in React/typescript/web technology, we should be able to attach a tooltip that explains what the button does (when enabled) or why it's disabled (when disabled).  The current tooltip technology used in C# doesn't show anything when the buttons are disabled, but that shouldn't be a problem in typescript.
</blockquote></details>

@hatton 
Are we willing to live with a version which prevents publication by disabling the radio buttons but doesn't give an indication why?
If I understand correctly, this is the (only current) scenario where we don't allow publication:
`return _collectionSettings.HaveEnterpriseFeatures || !bookContainsOverlayElements || bookIsTranslatedFromShell;`
i.e. a derivative comic without enterprise.

In my opinion, we should just leave this code in place until it has a replacement of some kind.

andrew-polk (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5785)* status: all files reviewed, 1 unresolved discussion (waiting on @hatton)

hatton:
@andrew-polk not tracking, but it sounds like you just want some new tooltips on disabled stuff? Since that's what I'm working on, I'm happy to add them.

andrew-polk:
Up to you if a tooltip is sufficient. But unless you're replacing the radio buttons now, it would have to be a WinForms tooltip. This is the UI we are losing in this change:

hatton:
> In my opinion, we should just leave this code in place until it has a replacement of some kind.

Sounds right

StephenMcConnel (commented):
I've restored this bit of the old UI.  Its removal is pending the radio buttons migrated to web land.

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5785#-:-NUT478O0-3Ll01_aiEq:bjiaq9h)* status: all files reviewed, 1 unresolved discussion (waiting on @hatton)

andrew-polk (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5785)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @StephenMcConnel)

PR #5786, opened 2023-04-26
Fix a couple of bugs in serving Andika and other fonts (BL-11389)

StephenMcConnel:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5786)

### DIFF
--- a/src/BloomExe/Book/Book.cs
+++ b/src/BloomExe/Book/Book.cs
@@ -1589,13 +1589,7 @@ namespace Bloom.Book
 			if (WriteFontFaces)
 			{
 				var serve = FontServe.GetInstance();
-				var headerBuilder = new StringBuilder();
-				foreach (var fontInfo in serve.FontsServed)
-				{
-					foreach (var face in fontInfo.faces)
-						headerBuilder.AppendLine(face);
-				}
-				cssBuilder.Insert(0, headerBuilder.ToString());
+				cssBuilder.Insert(0, serve.GetAllFontFaceDeclarations());
 			}
 			try
 			{
--- a/src/BloomExe/FontProcessing/FontServe.cs
+++ b/src/BloomExe/FontProcessing/FontServe.cs
@@ -4,7 +4,6 @@ using System.Diagnostics.CodeAnalysis;
 using System.IO;
 using System.Linq;
 using System.Text;
-using System.Threading.Tasks;
 using Newtonsoft.Json;
 using SIL.IO;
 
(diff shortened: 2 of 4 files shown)


andrew-polk (changes requested):
> Reviewed 4 of 4 files at r1.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5786)* status: all files reviewed (commit messages unreviewed), 1 unresolved discussion (waiting on @StephenMcConnel)

___
*[`src/BloomExe/web/BloomServer.cs` line 421 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5786#-NTykPos2GjYK31bxq3G:-NTykPos2GjYK31bxq3H:b-c6ad1l) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/a045a230d802886d7de000cf84e1d76d4e671839/src/BloomExe/web/BloomServer.cs#L421)):*
> ```Smalltalk
> 					return true;
> 				}
> 				else if (localPath == "book-preview/defaultLangStyles.css")
> ```

This makes me uncomfortable.
We have a method on Book to `CreateOrUpdateDefaultLangStyles()` but the preview doesn't necessarily get that created/updated version? Isn't that asking for trouble beyond this issue?

If we do end up with this special serving of the file for preview, it doesn't feel very future-proof to look the for existence of any `font-face`. One example where that might not go well: we add another font to the ones we serve. So the book has the old version and now won't get the new version.



_Code quote (from src/BloomExe/Book/Book.cs):_
```Smalltalk
CreateOrUpdateDefaultLangStyles()
```

StephenMcConnel (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5786)* status: 3 of 4 files reviewed, 1 unresolved discussion (waiting on @andrew-polk)

___
*[`src/BloomExe/web/BloomServer.cs` line 421 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5786#-NTykPos2GjYK31bxq3G:-NTz4gFa2f--hxgKEkZl:bm1ci55) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/a045a230d802886d7de000cf84e1d76d4e671839/src/BloomExe/web/BloomServer.cs#L421)):*

This makes me uncomfortable.
We have a method on Book to `CreateOrUpdateDefaultLangStyles()` but the preview doesn't necessarily get that created/updated version? Isn't that asking for trouble beyond this issue?

If we do end up with this special serving of the file for preview, it doesn't feel very future-proof to look the for existence of any `font-face`. One example where that might not go well: we add another font to the ones we serve. So the book has the old version and now won't get the new version.
</blockquote></details>

Preview can look at books that haven't been updated, but normally doesn't change things in the book's folder.  Some templates may not even have a defaultLangStyles.css file in their folder, and we can't always just add or modify a file in a template book, especially if Bloom is "installed for all users".
I've changed the check to look for the entire string of @font-face declarations.

andrew-polk (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5786)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @StephenMcConnel)

PR #5798, opened 2023-04-28
BL-11513 Get around WV2 1st ePUB build fail

gmartin7:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5798)

### DIFF
--- a/src/BloomExe/IBrowser.cs
+++ b/src/BloomExe/IBrowser.cs
@@ -147,7 +147,7 @@ namespace Bloom
 				Invoke(new Action<HtmlDom, HtmlDom, bool, BloomServer.SimulatedPageFileSource>(Navigate), htmlDom, htmlEditDom, setAsCurrentPageForDebugging, source);
 				return;
 			}
-
+			// This must already be called before calling Navigate(), but it doesn't really hurt to call it again.
 			EnsureBrowserReadyToNavigate();
 
 			XmlDocument dom = htmlDom.RawDom;
--- a/src/BloomExe/Publish/PublishHelper.cs
+++ b/src/BloomExe/Publish/PublishHelper.cs
@@ -7,7 +7,6 @@ using Bloom.Book;
 using System.Windows.Forms;
 using System.Xml;
 using Bloom.Api;
-using Bloom.Publish.AccessibilityChecker;
 using SIL.Reporting;
 using SIL.Xml;
 using Bloom.Publish.Epub;
@@ -17,7 +16,6 @@ using L10NSharp;
 using SIL.IO;
 using SIL.Progress;
 using Bloom.FontProcessing;
-using System.Text.RegularExpressions;
 
 namespace Bloom.Publish
 {
@@ -802,7 +800,7 @@ namespace Bloom.Publish
 						var msgFmt1 = LocalizationManager.GetString("PublishTab.FontProblem",
 							"This book has a font, \"{0}\", which has the following problem:");
 						var msg3 = LocalizationManager.GetString("PublishTab.FontProblem.Result",
-							"BloomLibrary.org will display the PDF and allow downloads for translation, but cannot offer the “READ” button or downloads for BloomPUB or ePUB.");
+							"BloomLibrary.org will display the PDF and allow downloads for translation, but cannot offer the ï¿½READï¿½ button or downloads for BloomPUB or ePUB.");
 						// progress.WriteError() uses Color.Red, but also exposes a link to "report error" which we don't want here.
 						progress.WriteMessageWithColor("Red", msgFmt1, font);
 						progress.WriteMessageWithColor("Red", " \u2022 {0}", msg2);
--- a/src/BloomExe/WebView2Browser.cs
+++ b/src/BloomExe/WebView2Browser.cs
@@ -216,6 +216,9 @@ namespace Bloom
 			// Should be called on UI thread. Since it is quite typical for this method to create the
 			// window handle and browser, it can't do its own Invoke, which depends on already having a handle.
 			// OTOH, Unit tests are often not run on the UI thread (and would therefore just pop up annoying asserts).
+			// For future reference, if we are navigating to produce a preview, make sure that the api call that
+			// requests the call is syncing on the correct thumbnail/preview sync object, otherwise we can get a
+			// deadlock here while trying to navigate (See BL-11513).
 			Debug.Assert(Program.RunningOnUiThread || Program.RunningUnitTests || Program.RunningInConsoleMode,
 				"Should be running on UI Thread or Unit Tests or Console mode");
 			var done = false;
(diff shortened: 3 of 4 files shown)


JohnThomson (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5798)* status: all files reviewed, 1 unresolved discussion (waiting on @gmartin7)

___
*[`src/BloomExe/Publish/PublishHelper.cs` line 253 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5798#-NUMagsk3a3iZX6XgxNE:-NUMagsk3a3iZX6XgxNF:b-byt7zw) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/cf1a6a5b979fa5c1fa423c02977df6f7daad389b/src/BloomExe/Publish/PublishHelper.cs#L253)):*
> ```Smalltalk
> 					// below.
> 					Cancel();
> 					return;
> ```

See the comment above. Looks to me as though a better fix would be to make PublishHelper.IsDisplayed just treat "not found" by returning true...maybe with some sort of warning. (I'm not clear what Cancel does...but if it cancels the whole epub creation, that's not good, and if it cancels all attempts to check anything for visibility, that also seems undesirable.)

JohnThomson (commented):
Also, do we know why we're not getting the document loaded in a reasonable time? Does the notification never come, or is the timeout too short, or....?

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5798#-:-NUMfRYME1rzuDW3zsNx:b-4k7jym)* status: all files reviewed, 1 unresolved discussion (waiting on @gmartin7)

gmartin7 (commented):
I tried doubling the timeout to 20sec from 10s and it made no appreciable difference. How would I go about figuring out why the document isn't getting loaded?

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5798#-:-NUNF7VnDK_twyK2evy-:bxuk0qu)* status: all files reviewed, 1 unresolved discussion (waiting on @JohnThomson)

___
*[`src/BloomExe/Publish/PublishHelper.cs` line 253 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5798#-NUMagsk3a3iZX6XgxNE:-NUNHxuN354mUi_Il65b:b-zdg77f) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/cf1a6a5b979fa5c1fa423c02977df6f7daad389b/src/BloomExe/Publish/PublishHelper.cs#L253)):*

See the comment above. Looks to me as though a better fix would be to make PublishHelper.IsDisplayed just treat "not found" by returning true...maybe with some sort of warning. (I'm not clear what Cancel does...but if it cancels the whole epub creation, that's not good, and if it cancels all attempts to check anything for visibility, that also seems undesirable.)
</blockquote></details>

Yeah, all `Cancel()` does is set `_latestInstance` to null... whatever that does. I'm guessing it does like you suspect and cancels all attempts to check anything for visibility.
Getting `IsDisplayed` to return either true or false doesn't solve the problem. It can still 'hang'

JohnThomson (changes requested):
One thing to try would be some sort of notification when, internally, the document thinks it is loaded (Javascript alert or Console.log if you can get at the result).
You could also add timing...how long does it take to navigate, when it works? My guess is less than a second...in which case it's a fair guess that something other than the browser taking more than 10s to load a single page document is the problem.
Since this seems to be random, other than a slight preference for "the first time", there's probably a race condition involved, and since even a long timeout doesn't help, a deadlock is quite likely. Both can be hard to track down. The deadlock can sometimes be found by looking at the threads debugger window when it is stuck. Is a relevant thread (especially the UI one and the one making the Epub) waiting for some lock? Does another thread have that lock and is waiting for something else? Is the UI thread waiting for something and some other thread doing an await or invoke? If so, can you see anything going on that might have happened in a different sequence and avoided the deadlock? Or some way to break the deadlock, such as deducing that a lock was not actually needed?

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5798#-:-NUNdnIF-1wJL-un0NBU:b-viaoc4)* status: all files reviewed, 1 unresolved discussion (waiting on @gmartin7)

___
*[`src/BloomExe/Publish/PublishHelper.cs` line 253 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5798#-NUMagsk3a3iZX6XgxNE:-NUNcAqhC0iPLPKS8hW1:b-tgraxm) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/cf1a6a5b979fa5c1fa423c02977df6f7daad389b/src/BloomExe/Publish/PublishHelper.cs#L253)):*

Yeah, all `Cancel()` does is set `_latestInstance` to null... whatever that does. I'm guessing it does like you suspect and cancels all attempts to check anything for visibility.
Getting `IsDisplayed` to return either true or false doesn't solve the problem. It can still 'hang'
</blockquote></details>

You ought to know the consequences of a change you are proposing, not be guessing about it!
So, it's not hanging waiting for navigation, because NavigateAndWaitTillDone is timing out. It's not hanging because it can't find the element in IsDisplayed, because it still happens when you change that. Where IS it hanging? See notes above.

JohnThomson (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5798)* status: all files reviewed, 2 unresolved discussions (waiting on @gmartin7)

___
*[`src/BloomExe/Publish/PublishHelper.cs` line 257 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5798#-NUNkEbH1an7t6470uxw:-NUNkEbH1an7t6470uxx:b-u4v6kd) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/cf1a6a5b979fa5c1fa423c02977df6f7daad389b/src/BloomExe/Publish/PublishHelper.cs#L257)):*
> ```Smalltalk
> 			}
> 			if (this != _latestInstance)
> 				return;
> ```

One consequence of this is that after one failure, nothing else unwanted will be removed, maybe even in later pages? IIRC, the purpose of _latestInstance is to speed up failure in an EpubMaker that is running in the background after we made a change that is going to cause us to generate the Epub again. But now you're making it give up on cleaning out unwanted stuff from an EpubMaker that is still going to produce a result we will use. That feels dangerous. Flowable epubs use simplified CSS files; they might not even _have_ the rules that hide the stuff we're failing to delete. How do you know the result won't be an absolute mess?

JohnThomson (approved):
:lgtm:

> Reviewed 3 of 3 files at r2, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5798#-:-NUSRgk9BiFTNmAqmOYV:bnfp4nl)* status: all files reviewed, 1 unresolved discussion

PR #5808, opened 2023-05-01
Fix glitches in deleting books for disconnected TC (BL-12024)

StephenMcConnel:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5808)

### DIFF
--- a/src/BloomExe/CollectionTab/CollectionModel.cs
+++ b/src/BloomExe/CollectionTab/CollectionModel.cs
@@ -337,10 +337,6 @@ namespace Bloom.CollectionTab
 			{
 				if (IsCurrentBookInCollection())
 				{
-					if (_tcManager.CollectionStatus != TeamCollectionStatus.None && !_tcManager.CheckConnection())
-					{
-						return false;
-					}
 					if (!_bookSelection.CurrentSelection.IsSaveable)
 					{
 						var msg = LocalizationManager.GetString("TeamCollection.CheckOutForDelete",
@@ -348,7 +344,7 @@ namespace Bloom.CollectionTab
 						ErrorReport.NotifyUserOfProblem(msg);
 						return false;
 					}
-					if (_tcManager.CannotDeleteBecauseDisconnected(_bookSelection.CurrentSelection.FolderPath))
+					if (_tcManager.CannotDeleteBecauseDisconnected(_bookSelection.CurrentSelection))
 					{
 						var msg = LocalizationManager.GetString("TeamCollection.ConnectForDelete",
 							"Please connect to the Team Collection before deleting books that are part of it.");
--- a/src/BloomExe/TeamCollection/TeamCollection.cs
+++ b/src/BloomExe/TeamCollection/TeamCollection.cs
@@ -1444,6 +1444,12 @@ namespace Bloom.TeamCollection
 			return statusFile;
 		}
 
+		internal static bool IsBookKnownToTeamCollection(string bookFolderPath)
+		{
+			var statusFile = GetStatusFilePathFromBookFolderPath(bookFolderPath);
+			return RobustFile.Exists(statusFile);
+		}
+
 		internal void WriteLocalStatus(string bookFolderName, BookStatus status, string collectionFolder = null, string collectionId = null)
 		{
 #if DEBUG
--- a/src/BloomExe/TeamCollection/TeamCollectionManager.cs
+++ b/src/BloomExe/TeamCollection/TeamCollectionManager.cs
@@ -138,17 +138,30 @@ namespace Bloom.TeamCollection
 
 		/// <summary>
 		/// This is an additional check on delete AFTER we make sure the book is checked out.
-		/// Even if it is, we can't delete it while disconnected because we don't have a way
-		/// to actually remove it from the TC. Our current Delete mechanism, unlike git etc.,
-		/// does not postpone delete until commit.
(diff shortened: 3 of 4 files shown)


StephenMcConnel (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5808)* status: 2 of 5 files reviewed, 1 unresolved discussion (waiting on @JohnThomson)

___
*[`src/BloomExe/Book/Book.cs` line 338 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5808#-NUSYa0JCkZN3wZbfTyX:-NUSmjDT4ntsFvJyestf:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/6f4603bf1e4636674a4edf832f8efdf190060ba5/src/BloomExe/Book/Book.cs#L338)):*

An alternative way to detect that a book has never been checked in is the absence of a status file in the collection folder. Since I'm pretty sure that's how Bloom decides to show it as checked-out even though it's not in the TC at all, it would be more consistent if this worked the same way.
This may be tricky, though, because IIRC Book doesn't have any easy way to get to the TC(Manager), and if the current TC is a disconnected one, it may not have the logic for detecting status files.
I can't think of a case where the two routines would return different answers (absent a bug or file corruption), so I won't insist if it proves difficult, but I thought it was worth making the suggestion.
</blockquote></details>

Done.

JohnThomson (changes requested):
Logic is good now but I thought of a couple more things. Sorry.

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5808#-:-NUT3TSz34_hyj9BFmvI:byfy04g)* status: all files reviewed, 3 unresolved discussions (waiting on @StephenMcConnel)

___
*[`src/BloomExe/TeamCollection/TeamCollectionManager.cs` line 143 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5808#-NUT2hveCL08OqbS1037:-NUT2hveCL08OqbS1038:b75f9g3) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/f243208c8d4020227147f4af6f7b04a4e55f6cc1/src/BloomExe/TeamCollection/TeamCollectionManager.cs#L143)):*
> ```Smalltalk
> 		/// Even if it is, we can't delete it while disconnected because we don't have a way
> 		/// to actually remove it from the TC. Our current Delete mechanism, unlike git etc.,
> 		/// does not postpone delete until commit.
> ```

Sorry I didn't pick this up earlier, but I think the comment should be updated to reflect the exception if the book only exists locally.
___
*[`src/BloomExe/TeamCollection/TeamCollectionManager.cs` line 153 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5808#-NUT2uyd0bXaD_Pt1CFt:-NUT2uyd0bXaD_Pt1CFu:b-jccvjl) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/f243208c8d4020227147f4af6f7b04a4e55f6cc1/src/BloomExe/TeamCollection/TeamCollectionManager.cs#L153)):*
> ```Smalltalk
> 				return false;
> 			// check whether book is known to team collection and we're disconnected at this point in time (not just at startup)
> 			if (RobustFile.Exists(TeamCollection.GetStatusFilePathFromBookFolderPath(book.FolderPath)) && !CheckConnection())
> ```

It would be nice to capture the significance of the status file not existing...maybe capture the first condition in a method like HasBeenCheckedInAtLeastOnce or OnlyExistsLocally or IsInTeamCollection.

StephenMcConnel (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5808)* status: 3 of 5 files reviewed, 3 unresolved discussions (waiting on @JohnThomson)

___
*[`src/BloomExe/TeamCollection/TeamCollectionManager.cs` line 143 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5808#-NUT2hveCL08OqbS1037:-NUTDsgI5Bs_mL7QDu0Q:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/f243208c8d4020227147f4af6f7b04a4e55f6cc1/src/BloomExe/TeamCollection/TeamCollectionManager.cs#L143)):*

Sorry I didn't pick this up earlier, but I think the comment should be updated to reflect the exception if the book only exists locally.
</blockquote></details>

Done.
___
*[`src/BloomExe/TeamCollection/TeamCollectionManager.cs` line 153 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5808#-NUT2uyd0bXaD_Pt1CFt:-NUTEkpb9-nlou8SO2yc:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/f243208c8d4020227147f4af6f7b04a4e55f6cc1/src/BloomExe/TeamCollection/TeamCollectionManager.cs#L153)):*

It would be nice to capture the significance of the status file not existing...maybe capture the first condition in a method like HasBeenCheckedInAtLeastOnce or OnlyExistsLocally or IsInTeamCollection.
</blockquote></details>

Done.

JohnThomson (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5808)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @StephenMcConnel)

PR #5815, opened 2023-05-03
showCancel defaults to false now

jsubloom:
https://github.com/BloomBooks/BloomDesktop/commit/06d0af1e1aafa4331a3c7dd442865f3fee0c8853 set the default to true, but since showing the cancel button does no good unless the caller implements a handler for the button, I think it makes more sense to default to false instead.

### DIFF
--- a/src/BloomExe/MiscUI/BrowserProgressDialog.cs
+++ b/src/BloomExe/MiscUI/BrowserProgressDialog.cs
@@ -122,7 +122,7 @@ namespace Bloom.MiscUI
 			Func<IWebSocketProgress, BackgroundWorker, Task<bool>> doWhat,
 			string id,
 			string title,
-			bool showCancelButton = true,
+			bool showCancelButton = false,	// true will add a cancel button, but the caller is still responsible for handling the clicks (either here in C# via checking worker.CancellationPending or on the Javascript/React side)
 			Action doWhenMainActionFalse = null,
 			Action doWhenDialogCloses = null,
 			string titleIcon= null)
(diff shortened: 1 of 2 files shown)


jsubloom (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5815)* status: 0 of 2 files reviewed, all discussions resolved

___
*[`src/BloomExe/MiscUI/BrowserProgressDialog.cs` line 121 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5815#-NUX_S1gB8mZ5Kiy85_r:-NUX_S1gB8mZ5Kiy85_s:bypyd0f) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/8f148ed9ed05a3b99b20e0217c01c707b7d7aade/src/BloomExe/MiscUI/BrowserProgressDialog.cs#L121)):*
> ```Smalltalk
> 		private static Action _doWhenProgressDialogCloses;
> 
> 		public static async Task DoWorkWithProgressDialogAsync(IBloomWebSocketServer socketServer,
> ```

FYI, there are 4 callers according to Visual Studio.

1 caller had it explicitly false.
1 caller had it implicitly true (and wants it true)
2 callers were previously implicitly true, but didn't seem to want a cancel button. In https://github.com/BloomBooks/BloomDesktop/commit/f8cc6a9c7acdce4f4e4a0674dc6aa8ed80411f99, this was changed to be explicitly false.

andrew-polk (approved):

PR #5817, opened 2023-05-03
Implement clipboard and undo buttons for WV2 (BL-11642)

JohnThomson:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5817)

### DIFF
--- a/src/BloomBrowserUI/bookEdit/editViewFrame.ts
+++ b/src/BloomBrowserUI/bookEdit/editViewFrame.ts
@@ -167,7 +167,7 @@ export function canUndo(): string {
     if (contentWindow && contentWindow.ckeditorCanUndo()) {
         return "yes";
     }
-    return "fail"; //go ask the browser
+    return "fail"; //can't undo in Javascript, possibly something in C# can?
 }
 
 //noinspection JSUnusedGlobalSymbols
--- a/src/BloomBrowserUI/bookEdit/editablePage.ts
+++ b/src/BloomBrowserUI/bookEdit/editablePage.ts
@@ -15,6 +15,9 @@ export interface IPageFrameExports {
     prepareToSavePage(): void;
     pageUnloading(): void;
     disconnectForGarbageCollection(): void;
+    copySelection(): void;
+    cutSelection(): void;
+    pasteClipboardText(): void;
     makeElement(
         html: string,
         parent?: JQuery,
@@ -43,6 +46,9 @@ import {
     userStylesheetContent,
     pageUnloading,
     disconnectForGarbageCollection,
+    copySelection,
+    cutSelection,
+    pasteClipboardText,
     makeElement,
     SetupElements,
     attachToCkEditor
@@ -53,6 +59,9 @@ export {
     userStylesheetContent,
     pageUnloading,
     disconnectForGarbageCollection,
+    copySelection,
+    cutSelection,
+    pasteClipboardText,
     makeElement,
     SetupElements,
     attachToCkEditor
--- a/src/BloomBrowserUI/bookEdit/js/bloomEditing.ts
+++ b/src/BloomBrowserUI/bookEdit/js/bloomEditing.ts
@@ -1297,6 +1297,71 @@ export const disconnectForGarbageCollection = () => {
     });
 };
 
+// These clipboard functions are implemented in Javascript because WebView2 doesn't seem to have
+// a C# api for doing them. I've made the exported functions synchronous because I'm not sure
+// what complications might come from calling an async function from C#. The implementations
+// mostly use clipboard API functions that are async, so those functions must be async.
+// We don't need to await them because nothing is using the result.
+// The buttons that implement clipboard operations are currently only in Edit mode, so
+// this is a reasonable place for this code. If we support them elsewhere, we'll have to
+// find a way to share the code (and call it when not part of the editTabBundle).
+export const copySelection = () => {
+    copyImpl();
+};
+
+async function copyImpl() {
+    const sel = document.getSelection();
+    if (!sel) return;
+    navigator.clipboard.writeText(sel.toString());
+}
+
+// See comment on copySelection
+export const cutSelection = () => {
+    cutSelectionImpl();
+};
+
+async function cutSelectionImpl() {
+    const sel = document.getSelection();
+    if (!sel) return;
+    await navigator.clipboard.writeText(sel.toString());
+    // Using ckeditor here because it's the only way I've found to integrate clipboard
+    // ops into an Undo stack that we can operate from an external button.
+    // We do a Save before and after to make sure that the cut is distinct from
+    // any other editing and that ckEditor actually has an item in its undo stack
+    // so that the Undo gets activated.
+    (<any>CKEDITOR.currentInstance).undoManager.save(true);
+    const range = CKEDITOR.currentInstance.getSelection().getRanges()[0];
+    range.deleteContents();
+    range.select(); // Select emptied range to place the caret in its place.
+    (<any>CKEDITOR.currentInstance).undoManager.save(true);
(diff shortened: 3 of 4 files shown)


andrew-polk (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5817)* status: all files reviewed, 3 unresolved discussions (waiting on @JohnThomson)

___
*[`src/BloomExe/WebView2Browser.cs` line 380 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5817#-NUYakoj1PwbrpW70pY5:-NUYakoj1PwbrpW70pY6:b-jz8a6d) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/451bba5cb60f28155ff71b015d0639381fee5715/src/BloomExe/WebView2Browser.cs#L380)):*
> ```Smalltalk
> 			{
> 				// Note: this is only used for the Undo button in the toolbar;
> 				// ctrl-z is handled in JavaScript directly.
> ```

Is it not feasible to just implement a Ctrl-z from code?
In other words, use C# code (or even js?) to simulate the user typing Ctrl-z?

Now that I say that... why not do that for ctrl-x, -c, and -v?
___
*[`src/BloomExe/WebView2Browser.cs` line 393 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5817#-NUYbEiC1_HZaqPC8lTu:-NUYbEiC1_HZaqPC8lTv:b-wxikr0) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/451bba5cb60f28155ff71b015d0639381fee5715/src/BloomExe/WebView2Browser.cs#L393)):*
> ```Smalltalk
> 		/// We configure something in Javascript to keep track of this, since GeckoFx-45's CanXSelection properties aren't working,
> 		/// and a workaround involving making a GeckoWindow object and querying its selection led to memory leaks (BL-9757).
> 		/// </summary>
> ```

This comment needs to be updated or deleted.
___
*[`src/BloomExe/WebView2Browser.cs` line 439 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5817#-NUYbm7P87Z1DDEQj5Pq:-NUYbm7P87Z1DDEQj5Pr:b-krpuyc) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/451bba5cb60f28155ff71b015d0639381fee5715/src/BloomExe/WebView2Browser.cs#L439)):*
> ```Smalltalk
> 				if (result == "y")
> 				{
> 					result = RunJavaScript("editTabBundle.canUndo()");
> ```

Can we make use of modern js here to something like `editTabBundle?.canUndo.?()`?

JohnThomson (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5817)* status: 3 of 4 files reviewed, 3 unresolved discussions (waiting on @andrew-polk)

___
*[`src/BloomExe/WebView2Browser.cs` line 380 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5817#-NUYakoj1PwbrpW70pY5:-NUYeD69DYLYVr_Z-P0I:b-7cd71y) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/451bba5cb60f28155ff71b015d0639381fee5715/src/BloomExe/WebView2Browser.cs#L380)):*

Is it not feasible to just implement a Ctrl-z from code?
In other words, use C# code (or even js?) to simulate the user typing Ctrl-z?

Now that I say that... why not do that for ctrl-x, -c, and -v?
</blockquote></details>

Seems like it would be a pain at both ends...have to somehow prevent the default browser handling of the keys, and then find a way for C# to intercept keys that would otherwise go to the browser, and then invoke what the browser otherwise just does. But I suppose there might be a gain in consistency. I wasn't aiming for a large improvement in Undo handling, though, just to reinstate a reasonable approximation of what we had in Gecko.
___
*[`src/BloomExe/WebView2Browser.cs` line 393 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5817#-NUYbEiC1_HZaqPC8lTu:-NUYf4Bl3D3twuVpciNH:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/451bba5cb60f28155ff71b015d0639381fee5715/src/BloomExe/WebView2Browser.cs#L393)):*

This comment needs to be updated or deleted.
</blockquote></details>

Done.
___
*[`src/BloomExe/WebView2Browser.cs` line 439 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5817#-NUYbm7P87Z1DDEQj5Pq:-NUaioNfA7fU2x4U_QII:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/451bba5cb60f28155ff71b015d0639381fee5715/src/BloomExe/WebView2Browser.cs#L439)):*

Can we make use of modern js here to something like `editTabBundle?.canUndo.?()`?
</blockquote></details>

Done.

andrew-polk (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5817)* status: all files reviewed, 1 unresolved discussion (waiting on @JohnThomson)

___
*[`src/BloomExe/WebView2Browser.cs` line 380 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5817#-NUYakoj1PwbrpW70pY5:-NUbrIfb1ZkPLPr_Qm0h:b-nsnw2m) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/451bba5cb60f28155ff71b015d0639381fee5715/src/BloomExe/WebView2Browser.cs#L380)):*

Seems like it would be a pain at both ends...have to somehow prevent the default browser handling of the keys, and then find a way for C# to intercept keys that would otherwise go to the browser, and then invoke what the browser otherwise just does. But I suppose there might be a gain in consistency. I wasn't aiming for a large improvement in Undo handling, though, just to reinstate a reasonable approximation of what we had in Gecko.
</blockquote></details>

I was actually suggesting we just let the browser do what it wants with those keystrokes. Isn't that what happens if the user actually types them?

But I see now that the undo button can affect more than just typing.

But I still question whether we need all the code for copy/cut/paste. What if we just invoked the keystroke and let the browser do it?

JohnThomson (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5817)* status: all files reviewed, 1 unresolved discussion (waiting on @andrew-polk)

___
*[`src/BloomExe/WebView2Browser.cs` line 380 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5817#-NUYakoj1PwbrpW70pY5:-NUcurf-BIfH18vC4gjG:b-hts2fu) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/451bba5cb60f28155ff71b015d0639381fee5715/src/BloomExe/WebView2Browser.cs#L380)):*

I was actually suggesting we just let the browser do what it wants with those keystrokes. Isn't that what happens if the user actually types them?

But I see now that the undo button can affect more than just typing.

But I still question whether we need all the code for copy/cut/paste. What if we just invoked the keystroke and let the browser do it?
</blockquote></details>

It would probably work for the clipboard functions. As you note, we've made some enhancements to Undo.
Something in me resists the idea of emulating keystrokes, but I can't really articulate why. It may just be laziness. But here's at least one issue: are we sure there's no way for a user to configure different keystrokes for clipboard? Does a Chinese or Arabic keyboard even have these keys?

andrew-polk (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5817)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @JohnThomson)

PR #5818, opened 2023-05-03
BL-12163 Title page Publish vs Edit

gmartin7:
* removed an obsolete rule

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5818)

### DIFF
--- a/src/BloomBrowserUI/bookEdit/css/editPaneGlobal.less
+++ b/src/BloomBrowserUI/bookEdit/css/editPaneGlobal.less
@@ -7,8 +7,11 @@ Note: the webkit browsers (wkhtml which bloom uses for pdf, and google chrome)
 Need about 2-5% more space to draw the same text on a ine (can't tell if the font is bigger or the box is smaller)
 Therefore, in the basePage.css, we have 10pt, but here, 11pt.
 So if you change this, change it there. Note, this might not be the best long-term solution
-To handling the mis-match.*/
-    font-size: 10.5pt;
+To handling the mis-match.
+[BL-12163] after switching to WV2, I find the font-size here to be 10.5pt, not 11pt, but still
+too big. Let's see if we don't need this anymore. WV2 seems to have gotten it right!
+*/
+    //font-size: 10.5pt;
     margin: 0;
 }
 


andrew-polk (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5818)* status: all files reviewed, 1 unresolved discussion (waiting on @gmartin7)

___
*[`src/BloomBrowserUI/bookEdit/css/editPaneGlobal.less` line 14 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5818#-NUZ0mzH28iACW5-1mcb:-NUZ0mzH28iACW5-1mcc:b-pyeidm) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/8eb7656314e65163c0fc4453fe899506f837525b/src/BloomBrowserUI/bookEdit/css/editPaneGlobal.less#L14)):*
> ```Less
> too big. Let's see if we don't need this anymore. WV2 seems to have gotten it right!
> */
>     //font-size: 10.5pt;
> ```

I think maybe this has nothing to do with the switch to WV2?

If I understand correctly, it is left over from when edit was in gecko but pdf generation was in webkit?
So, now that they are both the same (both gecko for <=5.4 or both wv2 for >5.5), we don't need it?

So it should be removed from both here and whatever is setting it to 10pt for pdf generation, right?

andrew-polk (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5818)* status: all files reviewed, 1 unresolved discussion (waiting on @gmartin7)

___
*[`src/BloomBrowserUI/bookEdit/css/editPaneGlobal.less` line 14 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5818#-NUZ0mzH28iACW5-1mcb:-NUZ1VkO9NumvGRgMH5y:b-8ra4h9) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/8eb7656314e65163c0fc4453fe899506f837525b/src/BloomBrowserUI/bookEdit/css/editPaneGlobal.less#L14)):*

I think maybe this has nothing to do with the switch to WV2?

If I understand correctly, it is left over from when edit was in gecko but pdf generation was in webkit?
So, now that they are both the same (both gecko for <=5.4 or both wv2 for >5.5), we don't need it?

So it should be removed from both here and whatever is setting it to 10pt for pdf generation, right?
</blockquote></details>

Or if it can't be removed, it at least should be set to the same value in both places?
So does that argue for it being set in basePage.less instead?

gmartin7 (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5818)* status: all files reviewed, 1 unresolved discussion (waiting on @andrew-polk)

___
*[`src/BloomBrowserUI/bookEdit/css/editPaneGlobal.less` line 14 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5818#-NUZ0mzH28iACW5-1mcb:-NUgyWGA49IFmz2YVmyq:bqm3huq) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/8eb7656314e65163c0fc4453fe899506f837525b/src/BloomBrowserUI/bookEdit/css/editPaneGlobal.less#L14)):*

Or if it can't be removed, it at least should be set to the same value in both places?
So does that argue for it being set in basePage.less instead?
</blockquote></details>

`basePage.less` sets font-size to 1em. That seems to be what is 'active' in edit and in pdf generation and in collection tab preview.

gmartin7:
Closing this PR in favor of one on `5.4`.

PR #5819, opened 2023-05-04
Fix when a WinForms ReactDialog is not initially open (BL-12171)

JohnThomson:
In particular, this corrects a recent change to ProgressDialog that prevented the progress showing up and thus prevented TeamCollections from syncing.
c99388

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5819)

### DIFF
--- a/src/BloomExe/TeamCollection/FolderTeamCollection.cs
+++ b/src/BloomExe/TeamCollection/FolderTeamCollection.cs
@@ -873,8 +873,15 @@ namespace Bloom.TeamCollection
 			                          repoFolderPathFromLinkPath == repoFolder;
 			var joiningGuid = CollectionSettings.CollectionIdFromCollectionFolder(tcManager.CurrentCollection
 				.LocalCollectionFolder);
-			var localGuid = CollectionSettings.CollectionIdFromCollectionFolder(localCollectionFolder);
-			var isSameCollection = _joiningSameCollection = joiningGuid == localGuid;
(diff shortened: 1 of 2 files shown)


jsubloom (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5819)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @JohnThomson)

___
*[`src/BloomExe/web/ReactControl.cs` line 137 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5819#-NUbd78k1eTEijjNfu4R:-NUbd78k1eTEijjNfu4S:b-n4jxva) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/5e9873f6ade9ef00b9b6b551ac4c233641038dc9/src/BloomExe/web/ReactControl.cs#L137)):*
> ```Smalltalk
> 			tempFile.Detach(); // the browser control will clean it up
> 
> 			// If we're launching a WinForms dialog to show this content, we want it to be visible.
> ```

I'm skeptical about this part of the change.

I have another idea. Can you try it out and see if it works?
In TeamCollection.cs, find where it opens ReactDialog("progressDialogBundle", [...])
Add open="true" to the props.

JohnThomson:
Decided to use another approach

PR #5827, opened 2023-05-05
[5.4] BL-12163 Fix Title page mismatch

gmartin7:
* Edit vs Publish -> PDF vs Collection tab Preview

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5827)

### DIFF
--- a/src/BloomBrowserUI/bookEdit/css/editPaneGlobal.less
+++ b/src/BloomBrowserUI/bookEdit/css/editPaneGlobal.less
@@ -2,13 +2,6 @@
     BODY is not within the scoped css DIV */
 
 body {
-    /*Most text in the document should be % changes from this value. Collection.css can then change it to make all the text  bigger.
-Note: the webkit browsers (wkhtml which bloom uses for pdf, and google chrome)
-Need about 2-5% more space to draw the same text on a ine (can't tell if the font is bigger or the box is smaller)
-Therefore, in the basePage.css, we have 10pt, but here, 11pt.
-So if you change this, change it there. Note, this might not be the best long-term solution
-To handling the mis-match.*/
-    font-size: 10.5pt;
     margin: 0;
 }
 
--- a/src/content/bookLayout/basePage.less
+++ b/src/content/bookLayout/basePage.less
@@ -73,12 +73,12 @@
 .Browser-Reset();
 
 body {
-    /*Most text in the document should be % changes from this value. Collection.css can then change it to make all the text  bigger.
-Note: the webkit browsers (wkhtml which bloom uses for pdf, and google chrome)
-Need about 2% more space to draw the same text on a ine (can't tell if the font is bigger or the box is smaller)
-Therefore, in the editMode.css, we repeat this, but with a 2% larger size.
-So if you change this, change it there. Note, this might not be the best long-term solution
-To handling the mis-match.*/
+    /*
+    We used to set font-size slightly larger in editPaneGlobal.less, to handle a mismatch between webkit
+    pdf generation and geckofx edit mode. That no longer seems to be necessary and the Browser-Reset() above
+    sets a base font-size of 1em, which seems to be used everywhere. Inexplicably, if we remove this rule
+    font-sizes everywhere jump up about 3px in size, so we'll leave this one here.
+    */
     font-size: 10pt;
 }
 


andrew-polk (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5827)* status: all files reviewed, 1 unresolved discussion (waiting on @gmartin7)

___
*[`src/BloomBrowserUI/bookEdit/css/editPaneGlobal.less` line 16 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5827#-NUh0kiH3d5zbkBXwKaj:-NUh0kiH3d5zbkBXwKak:bvibkwu) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/2641c964f547d1e3d7323ecf1269b62761fc3bd3/src/BloomBrowserUI/bookEdit/css/editPaneGlobal.less#L16)):*
Continuing the conversation from the original PR:

> > So it should be removed from both here and whatever is setting it to 10pt for pdf generation, right?
> > 
> > Or if it can't be removed, it at least should be set to the same value in both places?  
So does that argue for it being set in basePage.less instead?
> 
> `basePage.less` sets font-size to 1em. That seems to be what is 'active' in edit and in pdf generation and in collection tab preview.

I see this in basePage.less:
```
body {
    /*Most text in the document should be % changes from this value. Collection.css can then change it to make all the text  bigger.
Note: the webkit browsers (wkhtml which bloom uses for pdf, and google chrome)
Need about 2% more space to draw the same text on a ine (can't tell if the font is bigger or the box is smaller)
Therefore, in the editMode.css, we repeat this, but with a 2% larger size.
So if you change this, change it there. Note, this might not be the best long-term solution
To handling the mis-match.*/
    font-size: 10pt;
}
```

If we now believe these settings are obsolete, I don't think I see any value in keeping them (though commented) with their comments. Unless you disagree, let's just remove all the clutter.

gmartin7 (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5827)* status: all files reviewed, 1 unresolved discussion (waiting on @andrew-polk)

___
*[`src/BloomBrowserUI/bookEdit/css/editPaneGlobal.less` line 16 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5827#-NUh0kiH3d5zbkBXwKaj:-NUh1PeX0R4hit9hvOQE:b-pupsiy) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/2641c964f547d1e3d7323ecf1269b62761fc3bd3/src/BloomBrowserUI/bookEdit/css/editPaneGlobal.less#L16)):*

Continuing the conversation from the original PR:

> > So it should be removed from both here and whatever is setting it to 10pt for pdf generation, right?
> > 
> > Or if it can't be removed, it at least should be set to the same value in both places?  
So does that argue for it being set in basePage.less instead?
> 
> `basePage.less` sets font-size to 1em. That seems to be what is 'active' in edit and in pdf generation and in collection tab preview.

I see this in basePage.less:
```
body {
    /*Most text in the document should be % changes from this value. Collection.css can then change it to make all the text  bigger.
Note: the webkit browsers (wkhtml which bloom uses for pdf, and google chrome)
Need about 2% more space to draw the same text on a ine (can't tell if the font is bigger or the box is smaller)
Therefore, in the editMode.css, we repeat this, but with a 2% larger size.
So if you change this, change it there. Note, this might not be the best long-term solution
To handling the mis-match.*/
    font-size: 10pt;
}
```

If we now believe these settings are obsolete, I don't think I see any value in keeping them (though commented) with their comments. Unless you disagree, let's just remove all the clutter.
</blockquote></details>

Good point! Since what it active is not the 10pt, but the 1em! Silly me. I'll remove it all.

andrew-polk:
*[`src/BloomBrowserUI/bookEdit/css/editPaneGlobal.less` line 16 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5827#-NUh0kiH3d5zbkBXwKaj:-NUh1ReR8pWAB9-rx4Gr:b-hccts4) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/2641c964f547d1e3d7323ecf1269b62761fc3bd3/src/BloomBrowserUI/bookEdit/css/editPaneGlobal.less#L16)):*

Good point! Since what it active is not the 10pt, but the 1em! Silly me. I'll remove it all.
</blockquote></details>

This is actually a somewhat risky change.
How can we convince ourselves that this isn't going to break something unexpected?

gmartin7 (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5827)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @andrew-polk)

___
*[`src/BloomBrowserUI/bookEdit/css/editPaneGlobal.less` line 16 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5827#-NUh0kiH3d5zbkBXwKaj:-NUhYqcRDmX9GrxMr5Xc:bda0mml) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/2641c964f547d1e3d7323ecf1269b62761fc3bd3/src/BloomBrowserUI/bookEdit/css/editPaneGlobal.less#L16)):*

This is actually a somewhat risky change.
How can we convince ourselves that this isn't going to break something unexpected?
</blockquote></details>

Good question. I tried it without the basePage.less rule and things broke. The editPageGlobal rule only makes the edit vs other systems work the same. Testing, I suppose, is the answer?

andrew-polk (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5827)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @gmartin7)

___
*[`src/BloomBrowserUI/bookEdit/css/editPaneGlobal.less` line 16 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5827#-NUh0kiH3d5zbkBXwKaj:-NUheKlK5YIFTDfjKr5i:b-hq9nyn) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/2641c964f547d1e3d7323ecf1269b62761fc3bd3/src/BloomBrowserUI/bookEdit/css/editPaneGlobal.less#L16)):*

Good question. I tried it without the basePage.less rule and things broke. The editPageGlobal rule only makes the edit vs other systems work the same. Testing, I suppose, is the answer?
</blockquote></details>

The fact that we're only modifying the edit rule and that it should now align with everywhere else, the risk seems much lower than if we were to change basePage.

PR #5834, opened 2023-05-09
On level change, fix the body-level attr that reflects it (BL-12184)

JohnThomson:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5834)

### DIFF
--- a/src/BloomExe/Edit/EditingModel.cs
+++ b/src/BloomExe/Edit/EditingModel.cs
@@ -958,6 +958,19 @@ namespace Bloom.Edit
 			if (_currentlyDisplayedBook != null)
 			{
 				ToolboxView.SaveToolboxSettings(_currentlyDisplayedBook, data);
+				EnsureLevelAttrCorrect();
+			}
+		}
+
+		private void EnsureLevelAttrCorrect()
+		{
+			var currentLevel = _domForCurrentPage.Body.Attributes["data-leveledreaderlevel"]?.Value;
+			var correctLevel = _currentlyDisplayedBook.BookInfo.MetaData.LeveledReaderLevel.ToString();
+			if (correctLevel != currentLevel)
+			{
+				SaveNow();
+				_currentlyDisplayedBook.OurHtmlDom.Body.SetAttribute("data-leveledreaderlevel", correctLevel);
+				RethinkPageAndReloadIt();
 			}
 		}
 


andrew-polk (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5834)* status: all files reviewed, 1 unresolved discussion (waiting on @JohnThomson)

___
*[`src/BloomExe/Edit/EditingModel.cs` line 968 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5834#-NV0iIXS9co_V1XEWLbs:-NV0iIXS9co_V1XEWLbt:b-h5l6kt) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/3d1ebc088064800bffbb87ce274a0806a1bda308/src/BloomExe/Edit/EditingModel.cs#L968)):*
> ```Smalltalk
> 		{
> 			var currentLevel = _domForCurrentPage.Body.Attributes["data-leveledreaderlevel"]?.Value;
> 			var correctLevel = _currentlyDisplayedBook.BookInfo.MetaData.LeveledReaderLevel.ToString();
> ```

Is LeveledReaderLevel guaranteed to be non-null?
Since we are considering pushing this to 5.3, it needs to be as robust as possible.

I would even consider wrapping it in a try/catch except:
1. That feels like overkill and extra complexity.
2. If save or reload fails, we can't ignore the error.

hatton:
If this was NOT a regression, please target 5.4. If there was one, then do we need for 5.3?

JohnThomson (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5834)* status: all files reviewed, 1 unresolved discussion (waiting on @andrew-polk)

___
*[`src/BloomExe/Edit/EditingModel.cs` line 968 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5834#-NV0iIXS9co_V1XEWLbs:-NV1Cc_3CiFjeG9BE4r-:b-lppzam) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/3d1ebc088064800bffbb87ce274a0806a1bda308/src/BloomExe/Edit/EditingModel.cs#L968)):*

Is LeveledReaderLevel guaranteed to be non-null?
Since we are considering pushing this to 5.3, it needs to be as robust as possible.

I would even consider wrapping it in a try/catch except:
1. That feels like overkill and extra complexity.
2. If save or reload fails, we can't ignore the error.
</blockquote></details>

Yes. It's defined to return int, so returning undefined is definitely impossible. And it appears to be very robustly code and unlikely to throw.

JohnThomson (commented):
My experiments (recalling that Grade does not change before level 7):
5.3: updates when you switch to collection tab (therefore not noticeable unless you are on the Cover page when switching)
5.4.108: updates when you switch to another book and come back.
So I'm seeing a regression between 5.3 and 5.4. But Colin reported a regression in 5.3 from 5.2.
So, I cherry-picked to 5.4 and, since it was already reviewed, just pushed it.

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5834#-:-NV1E7t06GuEek5mhWnH:b-pkg1ml)* status: all files reviewed, 1 unresolved discussion (waiting on @andrew-polk)

JohnThomson:
Pushed the change direct to 5.4

PR #5863, opened 2023-05-18
Fix "Just Text" boxes having a border outside of Editing (BL-12187)

jsubloom:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5863)

### DIFF
--- a/src/BloomBrowserUI/bookEdit/css/editMode.less
+++ b/src/BloomBrowserUI/bookEdit/css/editMode.less
@@ -122,7 +122,10 @@ body {
 
 /*+}*/
 div.bloom-imageContainer {
-    .borderOnTopOfElement(@MediumGray);
+    // Allow book templates to define a custom borderColor, otherwise fallback to the default color
+    .borderOnTopOfElement(
+        var(--bloom-editMode-custom-border-color, @MediumGray)
+    );
 
     // This shows the focus border on the root (hasOverlay) image container
     // when it has been clicked on to show its editing buttons (:not(.bloom-hideImageButtons)).
@@ -541,10 +544,12 @@ div.pageLabel[contenteditable="true"] {
     /*[disabled]min-height:34px;*/
 }
 
-/*need a darker border when we have a background color*/
-.coverColor textarea,
+// NOTE: I don't think .coverColor textarea matches anything anymore. It and many of the textarea references may be able to be removed?
+.coverColor textarea,   /*need a darker border when we have a background color*/
 div.bloom-editable {
-    border: thin solid rgba(0, 0, 0, 0.2);
+    // Allow book templates to define a custom borderColor, otherwise fallback to the default color
+    border: thin solid
+        var(--bloom-editMode-custom-border-color, rgba(0, 0, 0, 0.2));
 }
 
 img.hoverUp {
--- a/src/BloomBrowserUI/bookEdit/css/editMode.less
+++ b/src/BloomBrowserUI/bookEdit/css/editMode.less
@@ -124,7 +124,7 @@ body {
 div.bloom-imageContainer {
     // Allow book templates to define a custom borderColor, otherwise fallback to the default color
     .borderOnTopOfElement(
-        var(--bloom-editMode-custom-border-color, @MediumGray)
+        var(--bloom-editMode-custom-image-border-color, @MediumGray)
     );
 
     // This shows the focus border on the root (hasOverlay) image container
@@ -549,7 +549,7 @@ div.pageLabel[contenteditable="true"] {
 div.bloom-editable {
     // Allow book templates to define a custom borderColor, otherwise fallback to the default color
     border: thin solid
-        var(--bloom-editMode-custom-border-color, rgba(0, 0, 0, 0.2));
+        var(--bloom-editMode-custom-text-border-color, rgba(0, 0, 0, 0.2));
 }
 
 img.hoverUp {
--- a/src/content/templates/template books/Digital Comic Book/common-comics.less
+++ b/src/content/templates/template books/Digital Comic Book/common-comics.less
@@ -27,6 +27,8 @@
     // the .borderOnTopOfElement(@color) mixin in editMode.less
     div.bloom-imageContainer:after {
         // you can't see the default border with a dark background
-        border: thin solid white;
+        // This sets a custom CSS property on the element, which other stylesheets
+        // (e.g. editMode.less) can reference
+        --bloom-editMode-custom-border-color: white;
     }
 }
--- a/src/content/templates/template books/Digital Comic Book/common-comics.less
+++ b/src/content/templates/template books/Digital Comic Book/common-comics.less
@@ -22,13 +22,8 @@
 .comic {
     background-color: @dark !important;
 
-    div.bloom-editable,
-    // To understand more about this :after selector, reference
-    // the .borderOnTopOfElement(@color) mixin in editMode.less
-    div.bloom-imageContainer:after {
-        // you can't see the default border with a dark background
-        // This sets a custom CSS property on the element, which other stylesheets
-        // (e.g. editMode.less) can reference
-        --bloom-editMode-custom-border-color: white;
-    }
+    // because of the @dark background, we need a different color in various places where we insert a border
+    // in edit mode so it will contrast.
+    --bloom-editMode-custom-text-border-color: white;
+    --bloom-editMode-custom-image-border-color: white;
 }


jsubloom (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5863)* status: 0 of 2 files reviewed, 1 unresolved discussion

___
*[`src/BloomBrowserUI/bookEdit/css/editMode.less` line 548 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5863#-NVku7Tf36VNJQTGKN25:-NVku7Tf36VNJQTGKN26:bm5oqpt) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/ff0462ce354817df316166f22c3cf9b83741dc17/src/BloomBrowserUI/bookEdit/css/editMode.less#L548)):*
> ```Less
> 
> // NOTE: I don't think .coverColor textarea matches anything anymore. It and many of the textarea references may be able to be removed?
> .coverColor textarea,   /*need a darker border when we have a background color*/
> ```

I didn't test the effects on elements matching this, because as far as I can tell, nothing matches it.  I see that WallCalendar still generates some textareas, but I don't think any are on the cover so I don't think any match .coverColor.

jsubloom (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5863)* status: 0 of 2 files reviewed, 2 unresolved discussions

___
*[`src/BloomBrowserUI/bookEdit/css/editMode.less` line 127 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5863#-NVkw5kDDh4WlZXIy-B2:-NVkw5kDDh4WlZXIy-B3:b-9a9t2d) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/ff0462ce354817df316166f22c3cf9b83741dc17/src/BloomBrowserUI/bookEdit/css/editMode.less#L127)):*
> ```Less
>     // Allow book templates to define a custom borderColor, otherwise fallback to the default color
>     .borderOnTopOfElement(
>         var(--bloom-custom-editMode-border-color, @MediumGray)
> ```

This is to deal with the imageContainer:after rule in common-comics.less
___
*[`src/BloomBrowserUI/bookEdit/css/editMode.less` line 548 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5863#-NVku7Tf36VNJQTGKN25:-NVkvrjO2SPciowndUd1:b-desqcu) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/ff0462ce354817df316166f22c3cf9b83741dc17/src/BloomBrowserUI/bookEdit/css/editMode.less#L548)):*

I didn't test the effects on elements matching this, because as far as I can tell, nothing matches it.  I see that WallCalendar still generates some textareas, but I don't think any are on the cover so I don't think any match .coverColor.
</blockquote></details>

Well, I don't think it should even be any functional change for those elements. They ought not to have any custom color defined, so they should fallback to the rgba(0,0,0,0.2) color.

jsubloom (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5863)* status: 0 of 2 files reviewed, 3 unresolved discussions

___
*[`src/content/templates/template books/Digital Comic Book/common-comics.less` line 32 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5863#-NVky3_R579zDHiHv8wC:-NVky3_S7CHHeT0a_7Mf:b1l5frx) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/55f61b3ef5197ff95439f6194bead56b00b3fe0e/src/content/templates/template%20books/Digital%20Comic%20Book/common-comics.less#L32)):*
> ```Less
>         // This sets a custom CSS property on the element, which other stylesheets
>         // (e.g. editMode.less) can reference
>         --bloom-editMode-custom-border-color: white;
> ```

By the way, the simpler change is just to revert moving a rule that set border:none when an imageContainer is not hovered from editMode.less back to bubble.less.
But I was trying to make it so that the CSS files that are shipped with published books have less edit-time stuff in them.

Well, this file still has a little bit of editMode awareness I guess. But hopefully this is more tolerable than writing rules looking at the ".hoverUp" class directly, which I think would couple it a lot more directly, and much more than desired.

I guess in theory, we could try to strip out any rules that mentioned "--bloom-editMode" at publishing time, if we were sufficiently motivated to get rid of them.

JohnThomson (commented):
Have you looked into how border colors on edit boxes change? IIRC we do special things when they are focused and/or hovered (with exceptions in some overlays)...does any of that need special casing for black backgrounds?

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5863#-:-NWEXVcd6MCByeEPGrzV:bh9lydu)* status: 0 of 2 files reviewed, 3 unresolved discussions (waiting on @jsubloom)

___
*[`src/content/templates/template books/Digital Comic Book/common-comics.less` line 32 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5863#-NVky3_R579zDHiHv8wC:-NWEUpeJ7egt7NoSJYyV:b-j37ymr) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/55f61b3ef5197ff95439f6194bead56b00b3fe0e/src/content/templates/template%20books/Digital%20Comic%20Book/common-comics.less#L32)):*

By the way, the simpler change is just to revert moving a rule that set border:none when an imageContainer is not hovered from editMode.less back to bubble.less.
But I was trying to make it so that the CSS files that are shipped with published books have less edit-time stuff in them.

Well, this file still has a little bit of editMode awareness I guess. But hopefully this is more tolerable than writing rules looking at the ".hoverUp" class directly, which I think would couple it a lot more directly, and much more than desired.

I guess in theory, we could try to strip out any rules that mentioned "--bloom-editMode" at publishing time, if we were sufficiently motivated to get rid of them.
</blockquote></details>

Why are we only doing this for image containers? The default 20% black border around text boxes is going to be hard to see, too. Also, it seems undesirable that the comic style sheet knows so precisely how this variable is supposed to be used. If we need two different ones in comics, maybe we should define two differently-named optional variables? And then stylesheets like comic just unconditionally give them a value.

JohnThomson (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5863)* status: all files reviewed, 3 unresolved discussions (waiting on @jsubloom)

jsubloom (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5863)* status: all files reviewed, 3 unresolved discussions (waiting on @JohnThomson)

___
*[`src/content/templates/template books/Digital Comic Book/common-comics.less` line 32 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5863#-NVky3_R579zDHiHv8wC:-NWEYwcOCvnHRjsU-ZGZ:b8k9ha) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/55f61b3ef5197ff95439f6194bead56b00b3fe0e/src/content/templates/template%20books/Digital%20Comic%20Book/common-comics.less#L32)):*

Why are we only doing this for image containers? The default 20% black border around text boxes is going to be hard to see, too. Also, it seems undesirable that the comic style sheet knows so precisely how this variable is supposed to be used. If we need two different ones in comics, maybe we should define two differently-named optional variables? And then stylesheets like comic just unconditionally give them a value. 
</blockquote></details>

1st half: Hmm, unfortunately the REviewable default preview was really misleadingi n this case... If you expand the context, you will see that it also appliesit to div.bloom-editable.  (I think that addresses your 20% black border comment).

2nd half: TBD

JohnThomson (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5863)* status: all files reviewed, 3 unresolved discussions (waiting on @jsubloom)

___
*[`src/content/templates/template books/Digital Comic Book/common-comics.less` line 32 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5863#-NVky3_R579zDHiHv8wC:-NWEcNCg2fhQ5hSefjwX:b-i8bzeu) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/55f61b3ef5197ff95439f6194bead56b00b3fe0e/src/content/templates/template%20books/Digital%20Comic%20Book/common-comics.less#L32)):*

1st half: Hmm, unfortunately the REviewable default preview was really misleadingi n this case... If you expand the context, you will see that it also appliesit to div.bloom-editable.  (I think that addresses your 20% black border comment).

2nd half: TBD
</blockquote></details>

Oh, that makes more sense. But why shouldn't common-comics.less just (completely unconditionally, outside any rule) say
```
// because comics have a black background, we need a different color in various places where we insert a border
// in edit mode so it will contrast.
 --bloom-editMode-custom-border-color: white;
```
Then it's entirely up to the rules you already have in editMode.less WHERE the color in the variable gets used. Seems less fragile and leaks less knowledge of edit-mode specifics.

The second half was just to say that if it helped, we could have two variables: for example,

```
--bloom-editMode-custom-image-border-color: white;
--bloom-editmode-custom-text-border-color: rgb(255,255,255, 30)
```
(or whatever fainter border actually looks good around a text box on a black background).

jsubloom (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5863)* status: 0 of 2 files reviewed, 2 unresolved discussions (waiting on @JohnThomson)

___
*[`src/content/templates/template books/Digital Comic Book/common-comics.less` line 32 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5863#-NVky3_R579zDHiHv8wC:-NWI1vV4E8T-qR4DALY3:b8esf0q) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/55f61b3ef5197ff95439f6194bead56b00b3fe0e/src/content/templates/template%20books/Digital%20Comic%20Book/common-comics.less#L32)):*

Oh, that makes more sense. But why shouldn't common-comics.less just (completely unconditionally, outside any rule) say
```
// because comics have a black background, we need a different color in various places where we insert a border
// in edit mode so it will contrast.
 --bloom-editMode-custom-border-color: white;
```
Then it's entirely up to the rules you already have in editMode.less WHERE the color in the variable gets used. Seems less fragile and leaks less knowledge of edit-mode specifics.

The second half was just to say that if it helped, we could have two variables: for example,

```
--bloom-editMode-custom-image-border-color: white;
--bloom-editmode-custom-text-border-color: rgb(255,255,255, 30)
```
(or whatever fainter border actually looks good around a text box on a black background).
</blockquote></details>

Ok, thanks for explaining! That's starting to make more sense. 

1) Could we just put it completely unconditionaly, outside of any rules?
Technically no, it's not a LESS variable.  It adds this referenceable custom property to the specific elements that the rule matches.
But it could be added to :root or .comic.  But actually, it can't be :root, because that would lead to behavioral changes (it would change the border colors on non .comic pages, whereas before we only wanted it to apply on .comic pages). So it must be .comic at max, or it will apply to too much.

By the way, what I did  doesn't add more knowledge of the structure of the book to common-comics.less? But I guess you're proposing that less knowledge is better and we might as well make that happen now?  Hmm, not 100% sure, but that does sound plausible to me.

I do think this way introduces slightly more risk than the way I did it.  Let's consider just the editable half of the problem. In revision 1, only elements that match the rule from editMode.less AND elements that match the rule in common-comic.less get their border changed from 20% black to white.  In revision 2, we loosen up the 2nd half of the rule.  (Similarly for the imageContainer half)
Well, I think (via reasoning) that applying it to .comic produces an equivalent set of matching elements (for both the editable half and the imageContainer half of the problem), but I wouldn't stake my life on the accuracy of my assessment.


2) Ok, the variable names seem more descriptive and precise when split into two, so let's do that
(Well, except i'm going to use the same value for them right now, since  the card isn't about re-deciding the color scheme)

jsubloom (commented):
Yes, we do special things when focused and hovered.  This PR touches the code that deals with the hover. The focus color works fine on dark backgrounds too.

The biggest problem for dark backgrounds (this is a different problem though) is that Overlay Tool's Text Color  "Default for style" defaults to black even in Digital Comic Books, which is unreadable by default on the dark background.

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5863#-:-NWIWYJH99Drl9KL5uuH:bnfq2lq)* status: 0 of 2 files reviewed, 2 unresolved discussions (waiting on @JohnThomson)

jsubloom (commented):
(The "Default for style" = black thing is also in 5.4, so that is not a new 5.5-only regression.  I guess I'll make a card for it and put it in Incoming)

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5863#-:-NWIXLxTB-nOtte3-hyH:b5zegth)* status: 0 of 2 files reviewed, 2 unresolved discussions (waiting on @JohnThomson)

jsubloom (commented):
Never mind, I won't make a card.  This only affects "Just Text". In real use cases, shouldn't the user have an image underneath already? We have no idea what background the Just Text will be over. Anyway, the black-on-black should hopefully not be as prevalent / problematic in real use cases as opposed to developer testing.

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5863#-:-NWIZaYdBI2iI4lVceZK:bokf5bo)* status: 0 of 2 files reviewed, 2 unresolved discussions (waiting on @JohnThomson)

JohnThomson (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5863)* status: 0 of 2 files reviewed, all discussions resolved (waiting on @jsubloom)

___
*[`src/content/templates/template books/Digital Comic Book/common-comics.less` line 32 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5863#-NVky3_R579zDHiHv8wC:-NWIbG_h6FA_DvZOIgXX:bl5ut8q) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/55f61b3ef5197ff95439f6194bead56b00b3fe0e/src/content/templates/template%20books/Digital%20Comic%20Book/common-comics.less#L32)):*

Ok, thanks for explaining! That's starting to make more sense. 

1) Could we just put it completely unconditionaly, outside of any rules?
Technically no, it's not a LESS variable.  It adds this referenceable custom property to the specific elements that the rule matches.
But it could be added to :root or .comic.  But actually, it can't be :root, because that would lead to behavioral changes (it would change the border colors on non .comic pages, whereas before we only wanted it to apply on .comic pages). So it must be .comic at max, or it will apply to too much.

By the way, what I did  doesn't add more knowledge of the structure of the book to common-comics.less? But I guess you're proposing that less knowledge is better and we might as well make that happen now?  Hmm, not 100% sure, but that does sound plausible to me.

I do think this way introduces slightly more risk than the way I did it.  Let's consider just the editable half of the problem. In revision 1, only elements that match the rule from editMode.less AND elements that match the rule in common-comic.less get their border changed from 20% black to white.  In revision 2, we loosen up the 2nd half of the rule.  (Similarly for the imageContainer half)
Well, I think (via reasoning) that applying it to .comic produces an equivalent set of matching elements (for both the editable half and the imageContainer half of the problem), but I wouldn't stake my life on the accuracy of my assessment.


2) Ok, the variable names seem more descriptive and precise when split into two, so let's do that
(Well, except i'm going to use the same value for them right now, since  the card isn't about re-deciding the color scheme)
</blockquote></details>

Interesting. You have a good point that a comic book could conceivably have non-comic pages that don't have a black background and don't want this rule. OTOH, by default at least, digital comic books have cover-color set to black, so we presumably DO want this fix to apply to the cover; but the cover page does NOT have the "comic" class, so the current fix won't apply there.

Something is apparently preventing us from changing the cover color for a digital comic, so if we can figure out a way to make the rule apply there it would at least be safe.

However, getting this to happen to the cover is a separate problem from the one you were supposed to solve, so I'm going to approve this and make a note in the card.

PR #5870, opened 2023-05-23
Limit length and other properties of various file names (BL-12228)

JohnThomson:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5870)


andrew-polk:
There are a bunch of commits in here you don't want like Promote master to 5.6.

PR #5873, opened 2023-05-24
Prevent bad bookshelf from showing in bulk bloompub publish (BL-12215)

StephenMcConnel:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5873)

### DIFF
--- a/src/BloomBrowserUI/contentful/UseContentful.tsx
+++ b/src/BloomBrowserUI/contentful/UseContentful.tsx
@@ -69,7 +69,6 @@ export function useContentful(
     if (!results || !results.result || results.queryString !== queryString) {
         return { loading: true, result: [], error: false };
     }
-
     return { loading: false, result: results.result, error: results.error };
 }
 
@@ -112,15 +111,21 @@ export function useContentfulBookShelvesForSubscription(
     };
 }
 
+// If urlKey is defined, get the corresponding label for the collection.
+// If it's undefined, return the provided default value.
 export function useGetLabelForCollection(
     urlKey: string,
     defaultResult: string
 ): string {
-    const { loading, result, error } = useContentful({
-        content_type: "collection",
-        select: "fields.label",
-        include: 2, // depth: we want the bookshelf collection objects as part of this query
-        "fields.urlKey": `${urlKey}`
-    });
+    const { loading, result, error } = useContentful(
+        urlKey
+            ? {
+                  content_type: "collection",
+                  select: "fields.label",
+                  include: 2, // depth: we want the bookshelf collection objects as part of this query
+                  "fields.urlKey": `${urlKey}`
+              }
+            : undefined
+    );
     return result && result.length > 0 ? result[0].fields.label : defaultResult;
 }
--- a/src/BloomBrowserUI/publish/ReaderPublish/BulkBloomPub/BulkBloomPubDialog.tsx
+++ b/src/BloomBrowserUI/publish/ReaderPublish/BulkBloomPub/BulkBloomPubDialog.tsx
@@ -74,6 +74,8 @@ export const InnerBulkBloomPubDialog: React.FunctionComponent<{
     )?.defaultBookshelfUrlKey;
 
     // the server doesn't actually know the label for the bookshelf, just its urlKey. So we have to look that up ourselves.
+    // If bookshelfUrlKey is falsy, then we get back an empty string for the label without actually asking the server for
+    // it, but calling all the hooks along the way.
     const bookshelfLabel = useGetLabelForCollection(bookshelfUrlKey, "");
     React.useEffect(() => {
         if (bookshelfLabel) {
--- a/src/BloomExe/Collection/CollectionSettings.cs
+++ b/src/BloomExe/Collection/CollectionSettings.cs
@@ -468,7 +468,7 @@ namespace Bloom.Collection
 					.Split(new[] { "," }, StringSplitOptions.RemoveEmptyEntries);
 				var defaultTags = ReadString(xml, "DefaultBookTags", "").Split(',');
 				var defaultBookshelfTag = defaultTags.Where(t => t.StartsWith("bookshelf:")).FirstOrDefault();
-				DefaultBookshelf = defaultBookshelfTag == null
+				DefaultBookshelf = (defaultBookshelfTag == null || BrandingProjectKey == "Default" || BrandingProjectKey == "Local-Community")
 					? ""
 					: defaultBookshelfTag.Substring("bookshelf:".Length);
 
--- a/src/BloomExe/Collection/CollectionSettingsDialog.cs
+++ b/src/BloomExe/Collection/CollectionSettingsDialog.cs
@@ -305,10 +305,14 @@ namespace Bloom.Collection
 			}
 
 			_collectionSettings.PageNumberStyle = PendingNumberingStyle; // non-localized key
-			
 
+			var oldBrand = _collectionSettings.BrandingProjectKey;
 			_collectionSettings.BrandingProjectKey = _brand;
 			_collectionSettings.SubscriptionCode = _subscriptionCode;
+			// We don't know which if any of the new branding's bookshelves we should upload to by default,
+			// but it will certainly be wrong to upload to one that belongs to some previous branding.
+			if (oldBrand != _brand)
+				_collectionSettings.DefaultBookshelf = "";
 
 			string xmatterKeyForcedByBranding = _collectionSettings.GetXMatterPackNameSpecifiedByBrandingOrNull();
 			PendingXmatter =


JohnThomson (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5873)* status: all files reviewed, 3 unresolved discussions (waiting on @StephenMcConnel)

___
*[`src/BloomBrowserUI/contentful/UseContentful.tsx` line 32 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5873#-NWEDAe_8PeypjBWsHqf:-NWEDAe_8PeypjBWsHqg:bschazj) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/d2de276fc86972facf2aeebd42a82462a82063f1/src/BloomBrowserUI/contentful/UseContentful.tsx#L32)):*
> ```XML
>     const queryString = JSON.stringify(query);
>     useEffect(() => {
>         if (!query || query["fields.urlKey"] === "") {
> ```

I think a comment is needed here, and possibly an adjustment to the block comment on the method. What is the significance of this field being an empty string? Does that mean something different from it being null or undefined? Why would it be missing other than a bug in the caller? Why (if not otherwise obvious) would that trigger returning without doing the query?
___
*[`src/BloomBrowserUI/contentful/UseContentful.tsx` line 65 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5873#-NWEDqea06PaXLsGRrjO:-NWEDqea06PaXLsGRrjP:bc3g4vq) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/d2de276fc86972facf2aeebd42a82462a82063f1/src/BloomBrowserUI/contentful/UseContentful.tsx#L65)):*
> ```XML
> 
>     if (!query || query["fields.urlKey"] === "") {
>         // now we're past the hooks, we can take our early exit.
> ```

Likewise, why is this result appropriate for an empty fields.urlKey? Do we expect a later render with a query that does have this information?
___
*[`src/BloomExe/Collection/CollectionSettingsDialog.cs` line 313 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5873#-NWEEkEf4xCplP3NMpMW:-NWEEkEf4xCplP3NMpMX:bz2mif9) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/d2de276fc86972facf2aeebd42a82462a82063f1/src/BloomExe/Collection/CollectionSettingsDialog.cs#L313)):*
> ```Smalltalk
> 			_collectionSettings.SubscriptionCode = _subscriptionCode;
> 			if (oldBrand != _brand)
> 				_collectionSettings.DefaultBookshelf = "";
> ```

This doesn't seem as though it would be right if we're switching to a valid branding. Don't we want its idea of the right default bookshelf?

StephenMcConnel (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5873)* status: all files reviewed, 3 unresolved discussions (waiting on @JohnThomson)

___
*[`src/BloomBrowserUI/contentful/UseContentful.tsx` line 32 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5873#-NWEDAe_8PeypjBWsHqf:-NWEWQcq3OCe26ud-_Uz:b-8td55u) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/d2de276fc86972facf2aeebd42a82462a82063f1/src/BloomBrowserUI/contentful/UseContentful.tsx#L32)):*

I think a comment is needed here, and possibly an adjustment to the block comment on the method. What is the significance of this field being an empty string? Does that mean something different from it being null or undefined? Why would it be missing other than a bug in the caller? Why (if not otherwise obvious) would that trigger returning without doing the query?
</blockquote></details>

I've revamped the fix to pass in undefined when the urlKey is falsy.  I've also added a bit of commentary to explain things.
___
*[`src/BloomBrowserUI/contentful/UseContentful.tsx` line 65 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5873#-NWEDqea06PaXLsGRrjO:-NWEdsNi1upMch6Vy7lW:bpo6whq) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/d2de276fc86972facf2aeebd42a82462a82063f1/src/BloomBrowserUI/contentful/UseContentful.tsx#L65)):*

Likewise, why is this result appropriate for an empty fields.urlKey? Do we expect a later render with a query that does have this information?
</blockquote></details>

An empty urlKey indiicates that we don't have anything to fetch.  But I've adjusted the code in the calling method to recognize the situation and pass in query===undefined.  (taking undue advantage of it being an "any" type)
___
*[`src/BloomExe/Collection/CollectionSettingsDialog.cs` line 313 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5873#-NWEEkEf4xCplP3NMpMW:-NWEVe2p4lc6HAEw0x0K:b-mx7x2c) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/d2de276fc86972facf2aeebd42a82462a82063f1/src/BloomExe/Collection/CollectionSettingsDialog.cs#L313)):*

This doesn't seem as though it would be right if we're switching to a valid branding. Don't we want its idea of the right default bookshelf?
</blockquote></details>

If a branding has only one bookshelf, I suppose that somehow we could figure that out, but if it has multiple bookshelves I don't think we can tell which should be the chosen one.  (I think DefaultBookshelf might be better named as CurrentBookshelf, but that's a separate issue.)
We've never tried to automatically set a default bookshelf.  And we've never tried to clear out a stale bookshelf setting, which is what I'm trying to do.  I would argue that the latter is far more important.

JohnThomson (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5873)* status: all files reviewed, 2 unresolved discussions (waiting on @StephenMcConnel)

___
*[`src/BloomBrowserUI/contentful/UseContentful.tsx` line 65 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5873#-NWEDqea06PaXLsGRrjO:-NWEj5GZ7oURu1px0sxD:b-lojz7v) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/d2de276fc86972facf2aeebd42a82462a82063f1/src/BloomBrowserUI/contentful/UseContentful.tsx#L65)):*

An empty urlKey indiicates that we don't have anything to fetch.  But I've adjusted the code in the calling method to recognize the situation and pass in query===undefined.  (taking undue advantage of it being an "any" type)
</blockquote></details>

much better. Thanks.
___
*[`src/BloomExe/Collection/CollectionSettingsDialog.cs` line 313 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5873#-NWEEkEf4xCplP3NMpMW:-NWEjKyP-JDDvLTuENJz:buewg5x) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/d2de276fc86972facf2aeebd42a82462a82063f1/src/BloomExe/Collection/CollectionSettingsDialog.cs#L313)):*

If a branding has only one bookshelf, I suppose that somehow we could figure that out, but if it has multiple bookshelves I don't think we can tell which should be the chosen one.  (I think DefaultBookshelf might be better named as CurrentBookshelf, but that's a separate issue.)
We've never tried to automatically set a default bookshelf.  And we've never tried to clear out a stale bookshelf setting, which is what I'm trying to do.  I would argue that the latter is far more important.
</blockquote></details>

I suppose it's the bookshelf that is currently the default one to put uploaded books on. So it has elements of both meanings.
Maybe comment "We don't know which if any of the new branding's bookshelves we should upload to by default, but it will certainly be wrong to upload to one that belongs to some previous branding"

StephenMcConnel (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5873)* status: all files reviewed, 2 unresolved discussions (waiting on @JohnThomson)

___
*[`src/BloomBrowserUI/contentful/UseContentful.tsx` line 65 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5873#-NWEDqea06PaXLsGRrjO:-NWEpfuvC7rn7SiocCGR:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/d2de276fc86972facf2aeebd42a82462a82063f1/src/BloomBrowserUI/contentful/UseContentful.tsx#L65)):*

much better. Thanks.
</blockquote></details>

Done.
___
*[`src/BloomExe/Collection/CollectionSettingsDialog.cs` line 313 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5873#-NWEEkEf4xCplP3NMpMW:-NWEpywR7rbETeSkoRYK:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/d2de276fc86972facf2aeebd42a82462a82063f1/src/BloomExe/Collection/CollectionSettingsDialog.cs#L313)):*

I suppose it's the bookshelf that is currently the default one to put uploaded books on. So it has elements of both meanings.
Maybe comment "We don't know which if any of the new branding's bookshelves we should upload to by default, but it will certainly be wrong to upload to one that belongs to some previous branding"
</blockquote></details>

Done.

JohnThomson (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5873)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @StephenMcConnel)

PR #5876, opened 2023-05-24
Publish BloomPUB handles cases where Preview is not pressed before publishing (BL-12241)

jsubloom:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5876)

### DIFF
--- a/src/BloomBrowserUI/publish/ReaderPublish/MethodChooser.tsx
+++ b/src/BloomBrowserUI/publish/ReaderPublish/MethodChooser.tsx
@@ -4,7 +4,11 @@ import { jsx, css } from "@emotion/react";
 import * as React from "react";
 import { RadioGroup } from "../../react_components/RadioGroup";
 import BloomButton from "../../react_components/bloomButton";
-import { useApiStringState, useWatchBooleanEvent } from "../../utils/bloomApi";
+import {
+    post,
+    useApiStringState,
+    useWatchBooleanEvent
+} from "../../utils/bloomApi";
 import { isLinux } from "../../utils/isLinux";
 import { useL10n } from "../../react_components/l10nHooks";
 import Typography from "@mui/material/Typography";
@@ -23,7 +27,9 @@ const methodNameToImageFileName = {
 
 // Lets the user choose how they want to "publish" the bloompub, along with a button to start that process.
 // This is a set of radio buttons and image that goes with each choice, plus a button to start off the sharing/saving
-export const MethodChooser: React.FunctionComponent = () => {
+export const MethodChooser: React.FunctionComponent<{
+    onStartPublish: () => void;
+}> = props => {
     const [method, setMethod] = useApiStringState(
         "publish/bloompub/method",
         "file"
@@ -100,7 +106,7 @@ export const MethodChooser: React.FunctionComponent = () => {
                             )
                         }}
                     />
-                    {getStartButton(method, isLicenseOK)}
+                    {getStartButton(method, isLicenseOK, props.onStartPublish)}
                 </div>
                 <div
                     css={css`
@@ -128,7 +134,11 @@ export const MethodChooser: React.FunctionComponent = () => {
     );
 };
 
-function getStartButton(method: string, licenseOK: boolean) {
+function getStartButton(
+    method: string,
+    licenseOK: boolean,
+    onClick: () => void
+) {
     const buttonCss =
         "align-self: flex-end; min-width: 120px; margin-top: 20px;";
     switch (method) {
@@ -140,7 +150,10 @@ function getStartButton(method: string, licenseOK: boolean) {
                     `}
                     l10nKey="PublishTab.Save"
                     l10nComment="Button that tells Bloom to save the book in the current format."
-                    clickApiEndpoint="publish/bloompub/file/save"
+                    onClick={() => {
+                        onClick(); // TODO: This doesn't seem to have any effect yet.
+                        post("publish/bloompub/file/save");
+                    }}
                     enabled={licenseOK}
                     hasText={true}
                     size="large"
@@ -157,6 +170,7 @@ function getStartButton(method: string, licenseOK: boolean) {
                     l10nKey="PublishTab.Android.Usb.Start"
                     l10nComment="Button that tells Bloom to send the book to a device via USB cable."
                     enabled={licenseOK}
+                    // TODO: Should this call onClick too?
                     clickApiEndpoint="publish/bloompub/usb/start"
                     hidden={isLinux()}
                     hasText={true}
@@ -174,7 +188,10 @@ function getStartButton(method: string, licenseOK: boolean) {
                     l10nKey="PublishTab.Android.Wifi.Start"
                     l10nComment="Button that tells Bloom to begin offering this book on the wifi network."
                     enabled={licenseOK}
-                    clickApiEndpoint="publish/bloompub/wifi/start"
+                    onClick={() => {
+                        onClick();
+                        post("publish/bloompub/wifi/start");
+                    }}
                     hasText={true}
                     size="large"
                 >
--- a/src/BloomBrowserUI/publish/ReaderPublish/MethodChooser.tsx
+++ b/src/BloomBrowserUI/publish/ReaderPublish/MethodChooser.tsx
@@ -139,6 +139,11 @@ function getStartButton(
     licenseOK: boolean,
     onClick: () => void
 ) {
+    const onButtonClick = (apiEndpoint: string) => {
+        onClick();
+        post(apiEndpoint);
+    };
+
     const buttonCss =
         "align-self: flex-end; min-width: 120px; margin-top: 20px;";
     switch (method) {
@@ -170,8 +175,9 @@ function getStartButton(
                     l10nKey="PublishTab.Android.Usb.Start"
                     l10nComment="Button that tells Bloom to send the book to a device via USB cable."
                     enabled={licenseOK}
-                    // TODO: Should this call onClick too?
-                    clickApiEndpoint="publish/bloompub/usb/start"
+                    onClick={() => {
+                        onButtonClick("publish/bloompub/usb/start");
+                    }}
                     hidden={isLinux()}
                     hasText={true}
                     size="large"
@@ -189,8 +195,7 @@ function getStartButton(
                     l10nComment="Button that tells Bloom to begin offering this book on the wifi network."
                     enabled={licenseOK}
                     onClick={() => {
-                        onClick();
-                        post("publish/bloompub/wifi/start");
+                        onButtonClick("publish/bloompub/wifi/start");
                     }}
                     hasText={true}
                     size="large"
--- a/src/BloomBrowserUI/publish/ReaderPublish/MethodChooser.tsx
+++ b/src/BloomBrowserUI/publish/ReaderPublish/MethodChooser.tsx
@@ -19,16 +19,28 @@ import { kMutedTextGray } from "../../bloomMaterialUITheme";
 import { kBloomWarning } from "../../utils/colorUtils";
 import HelpLink from "../../react_components/helpLink";
 
-const methodNameToImageFileName = {
+export type ReaderPublishMethods = "file" | "usb" | "wifi";
+
+const methodNameToImageFileName: {
+    [Property in ReaderPublishMethods]: string;
+} = {
     wifi: "publish-via-wifi.svg",
     usb: "publish-via-usb.svg",
     file: "publish-to-file.svg"
 };
 
+/**
+ * Typeguard for ReaderPublishMethods
+ */
+export function isReaderPublishMethods(x: string): x is ReaderPublishMethods {
+    return Object.prototype.hasOwnProperty.call(methodNameToImageFileName, x);
+}
+
 // Lets the user choose how they want to "publish" the bloompub, along with a button to start that process.
 // This is a set of radio buttons and image that goes with each choice, plus a button to start off the sharing/saving
 export const MethodChooser: React.FunctionComponent<{
-    onStartPublish: () => void;
+    // A callback that runs when the start button for that process (e.g. "Save", "Share", etc) is clicked
+    onStartButtonClick: (publishMethod: ReaderPublishMethods) => void;
 }> = props => {
     const [method, setMethod] = useApiStringState(
         "publish/bloompub/method",
@@ -106,7 +118,11 @@ export const MethodChooser: React.FunctionComponent<{
                             )
                         }}
                     />
-                    {getStartButton(method, isLicenseOK, props.onStartPublish)}
+                    {getStartButton(
+                        method,
+                        isLicenseOK,
+                        props.onStartButtonClick
+                    )}
                 </div>
                 <div
                     css={css`
@@ -137,13 +153,15 @@ export const MethodChooser: React.FunctionComponent<{
 function getStartButton(
     method: string,
     licenseOK: boolean,
-    onClick: () => void
+    onStartButtonClick: (publishMethod: ReaderPublishMethods) => void
 ) {
-    const onButtonClick = (apiEndpoint: string) => {
-        onClick();
-        post(apiEndpoint);
-    };
+    if (!isReaderPublishMethods(method)) {
+        throw new Error("Unhandled method choice");
+    }
 
+    const onClick = () => {
+        onStartButtonClick(method);
+    };
     const buttonCss =
         "align-self: flex-end; min-width: 120px; margin-top: 20px;";
     switch (method) {
@@ -155,9 +173,7 @@ function getStartButton(
                     `}
                     l10nKey="PublishTab.Save"
                     l10nComment="Button that tells Bloom to save the book in the current format."
-                    onClick={() => {
-                        onButtonClick("publish/bloompub/file/save");
-                    }}
+                    onClick={onClick}
                     enabled={licenseOK}
                     hasText={true}
                     size="large"
@@ -174,9 +190,7 @@ function getStartButton(
                     l10nKey="PublishTab.Android.Usb.Start"
                     l10nComment="Button that tells Bloom to send the book to a device via USB cable."
                     enabled={licenseOK}
-                    onClick={() => {
-                        onButtonClick("publish/bloompub/usb/start");
-                    }}
+                    onClick={onClick}
                     hidden={isLinux()}
                     hasText={true}
                     size="large"
@@ -193,9 +207,7 @@ function getStartButton(
                     l10nKey="PublishTab.Android.Wifi.Start"
                     l10nComment="Button that tells Bloom to begin offering this book on the wifi network."
                     enabled={licenseOK}
-                    onClick={() => {
-                        onButtonClick("publish/bloompub/wifi/start");
-                    }}
+                    onClick={onClick}
                     hasText={true}
                     size="large"
                 >
--- a/src/BloomBrowserUI/publish/ReaderPublish/MethodChooser.tsx
+++ b/src/BloomBrowserUI/publish/ReaderPublish/MethodChooser.tsx
@@ -156,8 +156,7 @@ function getStartButton(
                     l10nKey="PublishTab.Save"
                     l10nComment="Button that tells Bloom to save the book in the current format."
                     onClick={() => {
-                        onClick(); // TODO: This doesn't seem to have any effect yet.
-                        post("publish/bloompub/file/save");
+                        onButtonClick("publish/bloompub/file/save");
                     }}
                     enabled={licenseOK}
                     hasText={true}
--- a/src/BloomBrowserUI/publish/ReaderPublish/ReaderPublishScreen.tsx
+++ b/src/BloomBrowserUI/publish/ReaderPublish/ReaderPublishScreen.tsx
@@ -105,8 +105,7 @@ const ReaderPublishScreenInternal: React.FunctionComponent<{
     );
 
     const pathToOutputBrowser = inStorybookMode ? "./" : "../../";
-    const usbWorking = useL10n("Publishing", "PublishTab.Common.Publishing");
-    const wifiWorking = useL10n("Publishing", "PublishTab.Common.Publishing");
+    const publishing = useL10n("Publishing", "PublishTab.Common.Publishing");
 
     useSubscribeToWebSocketForEvent(
         "publish-bloompub",
@@ -116,15 +115,23 @@ const ReaderPublishScreenInternal: React.FunctionComponent<{
                 case "stopped":
                     setClosePending(true);
                     break;
+                case "done":
+                    setProgressState(ProgressState.Done);
+                    break;
                 case "UsbStarted":
                     setClosePending(false);
-                    setHeading(usbWorking);
+                    setHeading(publishing);
                     setProgressState(ProgressState.Serving);
                     break;
                 case "ServingOnWifi":
                     setClosePending(false);
-                    setHeading(wifiWorking);
+                    setHeading(publishing);
                     setProgressState(ProgressState.Serving);
+                    break;
+                case "SavingFile":
+                    setClosePending(false);
+                    setHeading(publishing);
+                    setProgressState(ProgressState.Working);
                     break;
                 default:
                     throw new Error(
--- a/src/BloomBrowserUI/publish/ReaderPublish/ReaderPublishScreen.tsx
+++ b/src/BloomBrowserUI/publish/ReaderPublish/ReaderPublishScreen.tsx
@@ -115,9 +115,6 @@ const ReaderPublishScreenInternal: React.FunctionComponent<{
                 case "stopped":
                     setClosePending(true);
                     break;
-                case "done":
-                    setProgressState(ProgressState.Done);
-                    break;
                 case "UsbStarted":
                     setClosePending(false);
                     setHeading(publishing);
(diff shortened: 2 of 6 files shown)


JohnThomson (changes requested):
> Reviewed 1 of 3 files at r1, 1 of 3 files at r2, 2 of 2 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5876)* status: all files reviewed, 2 unresolved discussions (waiting on @jsubloom)

___
*[`src/BloomBrowserUI/publish/commonPublish/PublishProgressDialog.tsx` line 16 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5876#-NWEfurk5iEK-bGZjoV6:-NWEfurk5iEK-bGZjoV7:bxoyyqm) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/ffd98151818cdd056e86742ce1f17e86597ebc34/src/BloomBrowserUI/publish/commonPublish/PublishProgressDialog.tsx#L16)):*
> ```XML
>     heading: string; // up to client to localize
>     webSocketClientContext: string;
>     startApiEndpoint: string | null; // Null means no need to start with any API endpoint.
> ```

This needed a comment in the first place along the lines "an API that the dialog should post to inform the C# process that is waiting to do something that it is ready to receive progress messages. This usually starts whatever task we are displaying the progress of"
Which raises the question: if there isn't some process we're showing progress from, why do we need a progress dialog at all? If there is, how else do we make sure it doesn't start sending messages before the websocket is ready to receive them?
___
*[`src/BloomBrowserUI/publish/ReaderPublish/MethodChooser.tsx` line 144 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5876#-NWEhG3c4cElxVv8O65Y:-NWEhG3c4cElxVv8O65Z:baxlv9t) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/ffd98151818cdd056e86742ce1f17e86597ebc34/src/BloomBrowserUI/publish/ReaderPublish/MethodChooser.tsx#L144)):*
> ```XML
>     const onButtonClick = (apiEndpoint: string) => {
>         onClick();
>         post(apiEndpoint);
> ```

I'm not completely sure, but I think this api wants to be the startApiEndpoint of the ProgressDialog rather than getting posted before we're sure the progress dialog is ready to receive messages.

nabalone:
Should this PR also be switched to target Version5.5?

jsubloom (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5876)* status: all files reviewed, 2 unresolved discussions (waiting on @JohnThomson)

___
*[`src/BloomBrowserUI/publish/ReaderPublish/MethodChooser.tsx` line 144 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5876#-NWEhG3c4cElxVv8O65Y:-NWJ8L1qD-_LskBoHtkZ:b-p5hzwr) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/ffd98151818cdd056e86742ce1f17e86597ebc34/src/BloomBrowserUI/publish/ReaderPublish/MethodChooser.tsx#L144)):*

I'm not completely sure, but I think this api wants to be the startApiEndpoint of the ProgressDialog rather than getting posted before we're sure the progress dialog is ready to receive messages.
</blockquote></details>

That does make a lot of theoretical sense... and also opens up a big can of worms and makes this PR horribly complicated :)

(Because previously, I guess it had an underlying assumption (which was safe) that the websocket was guaranteed to be ready)

Working on it...

jsubloom (commented):
Oops! Yes, good catch. I changed it.

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5876#-:-NWJPdemAr5Gwk3eJkdZ:b-1m065a)* status: 1 of 6 files reviewed, 2 unresolved discussions (waiting on @JohnThomson)

___
*[`src/BloomBrowserUI/publish/commonPublish/PublishProgressDialog.tsx` line 16 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5876#-NWEfurk5iEK-bGZjoV6:-NWJPaaHBSJ9iy2pTj1Q:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/ffd98151818cdd056e86742ce1f17e86597ebc34/src/BloomBrowserUI/publish/commonPublish/PublishProgressDialog.tsx#L16)):*

This needed a comment in the first place along the lines "an API that the dialog should post to inform the C# process that is waiting to do something that it is ready to receive progress messages. This usually starts whatever task we are displaying the progress of"
Which raises the question: if there isn't some process we're showing progress from, why do we need a progress dialog at all? If there is, how else do we make sure it doesn't start sending messages before the websocket is ready to receive them?
</blockquote></details>

Done.

nabalone:
I added a PR stacked onto this one to suggest we remove the stop button during the Update Preview step
https://github.com/jsubloom/BloomDesktop/pull/1

jsubloom (commented):
Merged Noel's into mine. Thanks noel!

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5876#-:-NWOt08V6qdxjtZEDrP9:b8k358q)* status: 0 of 6 files reviewed, 2 unresolved discussions (waiting on @JohnThomson)

JohnThomson (changes requested):
I'm finding this hard to follow, but I can't pin down any major thing I don't like. This lot of comments is just a few niggles. See what you think.

> Reviewed 5 of 5 files at r4.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5876#-:-NWP-kiX9Yg6EaXf4qve:baq5axz)* status: 4 of 6 files reviewed, 6 unresolved discussions (waiting on @jsubloom)

___
*[`src/BloomBrowserUI/publish/commonPublish/PublishProgressDialog.tsx` line 21 at r4](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5876#-NWOsQPm6MeCYhA0sCI5:-NWOsQPm6MeCYhA0sCI6:b-6jrl60) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/50e4eb62b801aa7142e83133d7d5a51e0fff39df/src/BloomBrowserUI/publish/commonPublish/PublishProgressDialog.tsx#L21)):*
> ```XML
>     // Callback executed when startApiEndpoint completes successfully.
>     // For example, if you want it to automatically close upon completion of the start API, you can make that happen here.
>     onStartApiSuccess: (() => void) | undefined;
> ```

I hate to be picky when I've already caused so much extra work, but this sounds like the name of a function to call when we've succeeded in starting the task. And that made me even more unhappy with startApiEndpoint, a name I never liked much.
What would you think of apiForStartingTask and onTaskComplete (or perhaps onTaskSucceeded, but I like Complete better....you can't have completed a task that failed.)
___
*[`src/BloomBrowserUI/publish/commonPublish/PublishProgressDialog.tsx` line 27 at r4](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5876#-NWOwSgEDp9TgjihMRO-:-NWOwSgEDp9TgjihMRO0:bg9eiup) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/50e4eb62b801aa7142e83133d7d5a51e0fff39df/src/BloomBrowserUI/publish/commonPublish/PublishProgressDialog.tsx#L27)):*
> ```XML
>     progressState: ProgressState;
>     setProgressState: (ProgressState) => void;
>     forceResetCounter?: number;
> ```

This name sounds like a function that will forcibly reset some counter. We have fairly widely used 'generation' for a number that something increments to force a restarting of something.
___
*[`src/BloomBrowserUI/publish/commonPublish/PublishProgressDialog.tsx` line 45 at r4](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5876#-NWOw9axCo0b2KT3czNZ:-NWOw9axCo0b2KT3czN_:biy2z9f) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/50e4eb62b801aa7142e83133d7d5a51e0fff39df/src/BloomBrowserUI/publish/commonPublish/PublishProgressDialog.tsx#L45)):*
> ```XML
>         // close it
>         setProgressState(ProgressState.Closed);
>         setClosePending(false);
> ```

Is setting closePending part of closing the dialog, or part of setting it up for next time?
___
*[`src/BloomBrowserUI/publish/ePUBPublish/ePUBPublishScreen.tsx` line 246 at r4](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5876#-NWOx_nV87J-JiY1IpPs:-NWOx_nW4IkmCSaWjKT6:b-w1jd6l) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/50e4eb62b801aa7142e83133d7d5a51e0fff39df/src/BloomBrowserUI/publish/ePUBPublish/ePUBPublishScreen.tsx#L246)):*
> ```XML
> 
>     const onStartApiSuccess = React.useCallback(() => {
>         setClosePending(true);
> ```

I don't think you need this wrapper. setClosePending is the setter from a useState, which is supposed to be the same function on every render. (At least, if I'm wrong about that, then it's probably a problem that you pass setProgressState and setClosePending as props that PublishProgressDialog hopes will not change constantly.)
___
*[`src/BloomBrowserUI/publish/ReaderPublish/MethodChooser.tsx` line 144 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5876#-NWEhG3c4cElxVv8O65Y:-NWP-KVFEHNcs8jl6znj:b-o95cu5) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/ffd98151818cdd056e86742ce1f17e86597ebc34/src/BloomBrowserUI/publish/ReaderPublish/MethodChooser.tsx#L144)):*

That does make a lot of theoretical sense... and also opens up a big can of worms and makes this PR horribly complicated :)

(Because previously, I guess it had an underlying assumption (which was safe) that the websocket was guaranteed to be ready)

Working on it...
</blockquote></details>

Sorry, didn't realize it would cause such complications! I suppose the assumption really was no longer safe? Things do seem to have gotten a lot more complicated, but I can't pin down any one thing that should be simplified.
___
*[`src/BloomBrowserUI/publish/video/PublishAudioVideo.tsx` line 612 at r4](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5876#-NWOz9CY-PIQgSI2E2B5:-NWOz9CY-PIQgSI2E2B6:b-tpah9j) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/50e4eb62b801aa7142e83133d7d5a51e0fff39df/src/BloomBrowserUI/publish/video/PublishAudioVideo.tsx#L612)):*
> ```XML
>     );
> 
>     const applyClosePending = React.useCallback(() => {
> ```

It's not obvious what it means to apply a closePending. Maybe just setClosePendingTrue()?

jsubloom (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5876)* status: 4 of 6 files reviewed, 6 unresolved discussions (waiting on @JohnThomson)

___
*[`src/BloomBrowserUI/publish/commonPublish/PublishProgressDialog.tsx` line 21 at r4](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5876#-NWOsQPm6MeCYhA0sCI5:-NWhKryrAoR4tEQ2X4RG:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/50e4eb62b801aa7142e83133d7d5a51e0fff39df/src/BloomBrowserUI/publish/commonPublish/PublishProgressDialog.tsx#L21)):*

I hate to be picky when I've already caused so much extra work, but this sounds like the name of a function to call when we've succeeded in starting the task. And that made me even more unhappy with startApiEndpoint, a name I never liked much.
What would you think of apiForStartingTask and onTaskComplete (or perhaps onTaskSucceeded, but I like Complete better....you can't have completed a task that failed.)
</blockquote></details>

Done.
___
*[`src/BloomBrowserUI/publish/commonPublish/PublishProgressDialog.tsx` line 27 at r4](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5876#-NWOwSgEDp9TgjihMRO-:-NWhKsjZBIFClewKtZge:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/50e4eb62b801aa7142e83133d7d5a51e0fff39df/src/BloomBrowserUI/publish/commonPublish/PublishProgressDialog.tsx#L27)):*

This name sounds like a function that will forcibly reset some counter. We have fairly widely used 'generation' for a number that something increments to force a restarting of something.
</blockquote></details>

Done.
___
*[`src/BloomBrowserUI/publish/commonPublish/PublishProgressDialog.tsx` line 45 at r4](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5876#-NWOw9axCo0b2KT3czNZ:-NWhKybtD1Mi2OYMKBcT:b-hx0abd) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/50e4eb62b801aa7142e83133d7d5a51e0fff39df/src/BloomBrowserUI/publish/commonPublish/PublishProgressDialog.tsx#L45)):*

Is setting closePending part of closing the dialog, or part of setting it up for next time?
</blockquote></details>

Hmm, interesting question. I think both could be valid perspectives.  However, I thought about it some more and we'd like setClosePending(false) to happen as atomically as possible with setProgressState(ProgressState.Closed), so I think it makes more sense to group them together.  (You don't want setClosePending(false) to run unless the progress state actually makes that transition as well).
___
*[`src/BloomBrowserUI/publish/ePUBPublish/ePUBPublishScreen.tsx` line 246 at r4](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5876#-NWOx_nV87J-JiY1IpPs:-NWhXfD12mMxdqx6Sh3G:b-o3jmpv) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/50e4eb62b801aa7142e83133d7d5a51e0fff39df/src/BloomBrowserUI/publish/ePUBPublish/ePUBPublishScreen.tsx#L246)):*

I don't think you need this wrapper. setClosePending is the setter from a useState, which is supposed to be the same function on every render. (At least, if I'm wrong about that, then it's probably a problem that you pass setProgressState and setClosePending as props that PublishProgressDialog hopes will not change constantly.)
</blockquote></details>

That is normally true, but because it's received via props and not in this component itself, the linter is not satisfied with that.
(I mean, I suppose it's technically correct. The parent component can theoretically pass in a different useState setter as long as it has the same types.)

I guess I can rename it to have a "...Props suffix.   (IDK, I feel like that is not very idiomatic in React, but honestly I wish destructured props had a way to identify more easily that there are props. I think a props prefix or suffix would be great, but for some reason that doesn't seem to be a common practice? Dunno why)
___
*[`src/BloomBrowserUI/publish/ReaderPublish/MethodChooser.tsx` line 144 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5876#-NWEhG3c4cElxVv8O65Y:-NWhYaehA3FXY2FDvUf-:bkbkfd9) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/ffd98151818cdd056e86742ce1f17e86597ebc34/src/BloomBrowserUI/publish/ReaderPublish/MethodChooser.tsx#L144)):*

Sorry, didn't realize it would cause such complications! I suppose the assumption really was no longer safe? Things do seem to have gotten a lot more complicated, but I can't pin down any one thing that should be simplified.
</blockquote></details>

Yeah, I don't think the assumption was valid anymore.
___
*[`src/BloomBrowserUI/publish/video/PublishAudioVideo.tsx` line 612 at r4](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5876#-NWOz9CY-PIQgSI2E2B5:-NWhZHjl7xTi5gK2GBSx:b-7dk18x) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/50e4eb62b801aa7142e83133d7d5a51e0fff39df/src/BloomBrowserUI/publish/video/PublishAudioVideo.tsx#L612)):*

It's not obvious what it means to apply a closePending. Maybe just setClosePendingTrue()?
</blockquote></details>

Sure. I kinda had a hard time with it. I was hesitant about starting with "set" because it might look like a useState setter, but it's actually just a normal function.

JohnThomson (changes requested):
Close now. Couple of minor suggestions, and we also had a conversation in Zulip.

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5876#-:-NWhtNVFDbhffT122YjK:bvxg9ws)* status: 1 of 6 files reviewed, 2 unresolved discussions (waiting on @jsubloom)

___
*[`src/BloomBrowserUI/publish/commonPublish/PublishProgressDialog.tsx` line 45 at r4](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5876#-NWOw9axCo0b2KT3czNZ:-NWhgkhM4pPOnM67SJLk:b-svnc4i) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/50e4eb62b801aa7142e83133d7d5a51e0fff39df/src/BloomBrowserUI/publish/commonPublish/PublishProgressDialog.tsx#L45)):*

Hmm, interesting question. I think both could be valid perspectives.  However, I thought about it some more and we'd like setClosePending(false) to happen as atomically as possible with setProgressState(ProgressState.Closed), so I think it makes more sense to group them together.  (You don't want setClosePending(false) to run unless the progress state actually makes that transition as well).
</blockquote></details>

Maybe, then, we don't want ANY of the resetting of the dialog so we can use it again later to happen unless we succeed in closing it? IOW, would it make sense to start with // close it...setProgressState... and then // set up for next time...the other four lines?
___
*[`src/BloomBrowserUI/publish/ePUBPublish/ePUBPublishScreen.tsx` line 246 at r4](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5876#-NWOx_nV87J-JiY1IpPs:-NWhsQ541Kp8EGo5cmKo:bd6b931) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/50e4eb62b801aa7142e83133d7d5a51e0fff39df/src/BloomBrowserUI/publish/ePUBPublish/ePUBPublishScreen.tsx#L246)):*

That is normally true, but because it's received via props and not in this component itself, the linter is not satisfied with that.
(I mean, I suppose it's technically correct. The parent component can theoretically pass in a different useState setter as long as it has the same types.)

I guess I can rename it to have a "...Props suffix.   (IDK, I feel like that is not very idiomatic in React, but honestly I wish destructured props had a way to identify more easily that there are props. I think a props prefix or suffix would be great, but for some reason that doesn't seem to be a common practice? Dunno why)
</blockquote></details>

OK. But since we renamed the thing that calls it, maybe this can be onTaskComplete too?
___
*[`src/BloomBrowserUI/publish/video/PublishAudioVideo.tsx` line 612 at r4](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5876#-NWOz9CY-PIQgSI2E2B5:-NWhsrDk20OrI1LAQQX5:bnvc53o) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/50e4eb62b801aa7142e83133d7d5a51e0fff39df/src/BloomBrowserUI/publish/video/PublishAudioVideo.tsx#L612)):*

Sure. I kinda had a hard time with it. I was hesitant about starting with "set" because it might look like a useState setter, but it's actually just a normal function.
</blockquote></details>

The history of setX in Javascript (originally paired with getX) goes back well before React. I'd be surprised if what you have causes confusion.

JohnThomson (commented):
> Reviewed 5 of 5 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5876)* status: all files reviewed, 2 unresolved discussions (waiting on @jsubloom)

JohnThomson (changes requested):
> Reviewed 2 of 2 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5876)* status: all files reviewed, 2 unresolved discussions (waiting on @jsubloom)

___
*[`src/BloomBrowserUI/publish/commonPublish/PublishProgressDialog.tsx` line 44 at r7](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5876#-NWlvYhADAuBBb58pnXD:-NWlvYhADAuBBb58pnXE:bfdhu4k) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/52d577e445069641e15a8ab9827d60ca59abf5d3/src/BloomBrowserUI/publish/commonPublish/PublishProgressDialog.tsx#L44)):*
> ```XML
>         // close it
>         setProgressStateProp(ProgressState.Closed);
>         setClosePendingProp(false);
> ```

I still think setClosePending belongs under the other comment. IIRC, once the progress state is set to Closed (assuming the client cooperates), the dialog is guaranteed to be hidden in the next render. We want closePending to be false so that it won't close immediately the next time it is used. If it only matters because of the possibility of showing the dialog again later, it feels to me like part of the "set up for next time."

You could also just drop the two comments. The method name is pretty self-explanatory.

Sorry, I feel like I'm being picky. But the whole closePending thing is confusing enough without making someone think that setting it false is going to contribute to getting the dialog closed.

Or maybe I'm the one who is totally misunderstanding. If so, anything you can add by way of clarifying comment would be appreciated.

jsubloom (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5876)* status: all files reviewed, 2 unresolved discussions (waiting on @JohnThomson)

___
*[`src/BloomBrowserUI/publish/commonPublish/PublishProgressDialog.tsx` line 45 at r4](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5876#-NWOw9axCo0b2KT3czNZ:-NWjK3ER4lh1j5ph0sCG:b-j49jvp) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/50e4eb62b801aa7142e83133d7d5a51e0fff39df/src/BloomBrowserUI/publish/commonPublish/PublishProgressDialog.tsx#L45)):*

Maybe, then, we don't want ANY of the resetting of the dialog so we can use it again later to happen unless we succeed in closing it? IOW, would it make sense to start with // close it...setProgressState... and then // set up for next time...the other four lines?
</blockquote></details>

I can't quite think of what can go wrong between the start and end of this function, but I think what you said does make more sense. Flipped. Thansk!
___
*[`src/BloomBrowserUI/publish/ePUBPublish/ePUBPublishScreen.tsx` line 246 at r4](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5876#-NWOx_nV87J-JiY1IpPs:-NWjKjaX5-MM8aeOv0wP:b-oz8tym) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/50e4eb62b801aa7142e83133d7d5a51e0fff39df/src/BloomBrowserUI/publish/ePUBPublish/ePUBPublishScreen.tsx#L246)):*

OK. But since we renamed the thing that calls it, maybe this can be onTaskComplete too?
</blockquote></details>

Good call.
___
*[`src/BloomBrowserUI/publish/video/PublishAudioVideo.tsx` line 612 at r4](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5876#-NWOz9CY-PIQgSI2E2B5:-NWjJvHs8wcpeQxk3LSI:bigpj19) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/50e4eb62b801aa7142e83133d7d5a51e0fff39df/src/BloomBrowserUI/publish/video/PublishAudioVideo.tsx#L612)):*

The history of setX in Javascript (originally paired with getX) goes back well before React. I'd be surprised if what you have causes confusion.
</blockquote></details>

Ok, sounds good!

jsubloom (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5876)* status: 4 of 6 files reviewed, 1 unresolved discussion (waiting on @JohnThomson)

___
*[`src/BloomBrowserUI/publish/commonPublish/PublishProgressDialog.tsx` line 44 at r7](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5876#-NWlvYhADAuBBb58pnXD:-NWmMWyYDDo7m04VvpI6:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/52d577e445069641e15a8ab9827d60ca59abf5d3/src/BloomBrowserUI/publish/commonPublish/PublishProgressDialog.tsx#L44)):*

I still think setClosePending belongs under the other comment. IIRC, once the progress state is set to Closed (assuming the client cooperates), the dialog is guaranteed to be hidden in the next render. We want closePending to be false so that it won't close immediately the next time it is used. If it only matters because of the possibility of showing the dialog again later, it feels to me like part of the "set up for next time."

You could also just drop the two comments. The method name is pretty self-explanatory.

Sorry, I feel like I'm being picky. But the whole closePending thing is confusing enough without making someone think that setting it false is going to contribute to getting the dialog closed.

Or maybe I'm the one who is totally misunderstanding. If so, anything you can add by way of clarifying comment would be appreciated.
</blockquote></details>

Done.

JohnThomson (approved):
> Reviewed 2 of 2 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5876)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @jsubloom)

PR #5878, opened 2023-05-25
[5.5] BL-12245 Fix Web Socket error

gmartin7:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5878)

### DIFF
--- a/src/BloomExe/web/BloomWebSocketServer.cs
+++ b/src/BloomExe/web/BloomWebSocketServer.cs
@@ -1,18 +1,18 @@
-﻿using System;
+using System;
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.Net.Sockets;
 using System.Windows.Forms;
-using Bloom;
-using Bloom.Api;
 using Bloom.web;
 using Fleck;
 using SIL.Reporting;
+using Logger = SIL.Reporting.Logger;
 
 namespace Bloom.Api
 {
 	/// <summary>
-	/// Runs a web socket server on the given port. Useful for high-frequency messages (like audio levels) and allows the backend to send messages to the client.
+	/// Runs a web socket server on the given port. Useful for high-frequency messages (like audio levels)
+	/// and allows the backend to send messages to the client.
 	///
 	/// About ports... we currently only have a single server that is used to pass any number of messages.
 	/// Then on the client side, we have multiple connections to this same server; each connection is called a "socket",
@@ -33,7 +33,15 @@ namespace Bloom.Api
 		//only a single client, we think more in terms of 1 to 1.  However there's nothing preventing multiple parts of the Bloom client from opening their
 		//own connection (socket).
 		private WebSocketServer _server;
+
+		// It would be nice to use something from .Net's System.Collections.Concurrent namespace, but none of the options
+		// seem to fit our needs (i.e. remove a particular connection when it wants to close). To get around this, we will
+		// lock the thread whenever we're changing or reading the collection.
 		private List<IWebSocketConnection> _allSockets;
+
+		// I (gjm) prefer to use a private object for locking, rather than locking on the list itself.
+		// See https://stackoverflow.com/questions/38653739/locking-on-the-object-that-is-being-synchronized-or-using-a-dedicated-lock-objec
+		private readonly object _connectionLock = new object();
 
 		// The one instance which any client may use.
 		// BloomWebSocketServer wants to be an application singleton, but currently,
@@ -60,11 +68,6 @@ namespace Bloom.Api
 			Instance = this;
 		}
 
-		public bool IsSocketOpen(string name)
-		{
-			return _allSockets.Exists(s => s.ConnectionInfo?.SubProtocol == name);
-		}
-
 		public void Init(string port)
 		{
 			FleckLog.Level = LogLevel.Warn;
@@ -78,7 +81,7 @@ namespace Bloom.Api
 			// It would allow Chrome to work without any special Chrome code on the client side.
 			// But it seems like a pain. Firefox is able to negotiate with Fleck just fine, and
 			// we only use subprotocols for debugging, so rather than list every
-			// subprotocol we use here, for now I just changed our browser-side code to to not send
+			// subprotocol we use here, for now I just changed our browser-side code to not send
 			// the subprotocol unless we're in Firefox.
 
 			try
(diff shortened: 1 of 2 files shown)


JohnThomson (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5878)* status: all files reviewed, 2 unresolved discussions (waiting on @gmartin7)

___
*[`src/BloomExe/web/BloomWebSocketServer.cs` line 94 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5878#-NWJQXOBCYuraHbnbR2B:-NWJQXOBCYuraHbnbR2C:b-bx6de3) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/d82f748438f02132c5e9f7f25c39c37806b86608/src/BloomExe/web/BloomWebSocketServer.cs#L94)):*
> ```Smalltalk
> 
> 						// But that breaks Chrome
> 						lock (this)
> ```

Can we lock on allSockets or a private object made just for locking? It's considered bad practice to lock on a public object.
___
*[`src/BloomExe/web/BloomWebSocketServer.cs` line 183 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5878#-NWJQr5Q9kYRs_muX2vb:-NWJQr5Q9kYRs_muX2vc:b-14nz1j) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/d82f748438f02132c5e9f7f25c39c37806b86608/src/BloomExe/web/BloomWebSocketServer.cs#L183)):*
> ```Smalltalk
> 					{
> 						try
> 						{
> ```

It's very dangerous to be doing this while locked. Almost anything could happen while sending something through a socket...javascript may, for example, respond by posting something. That probably happens on another thread. Then if that thread ends up wanting to add or remove a socket or broadcast a message...deadlock.
We can continue to prevent index-out-of-range problems by just locking long enough to make a thread-private copy of the list with ToArray().
But then the question becomes, could a socket become unavailable between when we check that it is available and when we call Send? Is there some other form of locking that would guard against that? I can't think of one that doesn't expose us to deadlock.
Looking at the Fleck source code, it appears that it will throw ConnectionNotAvailableException if the connection is not available. If we caught that immediately, we could ignore it or treat it like we now do when isAvailable is false. I don't see any reason not to continue the loop and send to any sockets that ARE available. Then again, Fleck isn't locking either, so the state of the socket could change asynchronously between when IT checks IsAvailable and when it tries to do stuff assuming that it is. That might result in other exceptions.
So: summary of suggestions:
- change from lock(this) to lock (_allSockets) [everywhere]
- for this loop, lock just while getting _allSockets.ToArray(), not for the whole loop, and especially don't hold any lock while calling socket.Send() or NonFatalProblem.Report() (or anything else that executes code outside this class).
- we can still check for socket null or unavailable, but also put a try-catch for ConnectionNotAvailableException around socket.Send. Treat cacthing it the same as !IsAvailable...that is, do nothing except in Dev mode...just let the loop continue. Actually, I suspect it's not worth doing anything in dev mode, either...it seems to just mean a browser control is shtting down and can no longer receive.

gmartin7 (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5878)* status: 1 of 2 files reviewed, 2 unresolved discussions (waiting on @JohnThomson)

___
*[`src/BloomExe/web/BloomWebSocketServer.cs` line 94 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5878#-NWJQXOBCYuraHbnbR2B:-NWNzFQFBviPRcBUQTWV:bpzu3jf) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/d82f748438f02132c5e9f7f25c39c37806b86608/src/BloomExe/web/BloomWebSocketServer.cs#L94)):*

Can we lock on allSockets or a private object made just for locking? It's considered bad practice to lock on a public object.
</blockquote></details>

Done. I decided on a separate object, since _allSockets is a List<> instance and I'm not sure about the ins and outs of locking on a generic collection.
___
*[`src/BloomExe/web/BloomWebSocketServer.cs` line 183 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5878#-NWJQr5Q9kYRs_muX2vb:-NWO5-uU-92sVw3cYzjB:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/d82f748438f02132c5e9f7f25c39c37806b86608/src/BloomExe/web/BloomWebSocketServer.cs#L183)):*

It's very dangerous to be doing this while locked. Almost anything could happen while sending something through a socket...javascript may, for example, respond by posting something. That probably happens on another thread. Then if that thread ends up wanting to add or remove a socket or broadcast a message...deadlock.
We can continue to prevent index-out-of-range problems by just locking long enough to make a thread-private copy of the list with ToArray().
But then the question becomes, could a socket become unavailable between when we check that it is available and when we call Send? Is there some other form of locking that would guard against that? I can't think of one that doesn't expose us to deadlock.
Looking at the Fleck source code, it appears that it will throw ConnectionNotAvailableException if the connection is not available. If we caught that immediately, we could ignore it or treat it like we now do when isAvailable is false. I don't see any reason not to continue the loop and send to any sockets that ARE available. Then again, Fleck isn't locking either, so the state of the socket could change asynchronously between when IT checks IsAvailable and when it tries to do stuff assuming that it is. That might result in other exceptions.
So: summary of suggestions:
- change from lock(this) to lock (_allSockets) [everywhere]
- for this loop, lock just while getting _allSockets.ToArray(), not for the whole loop, and especially don't hold any lock while calling socket.Send() or NonFatalProblem.Report() (or anything else that executes code outside this class).
- we can still check for socket null or unavailable, but also put a try-catch for ConnectionNotAvailableException around socket.Send. Treat cacthing it the same as !IsAvailable...that is, do nothing except in Dev mode...just let the loop continue. Actually, I suspect it's not worth doing anything in dev mode, either...it seems to just mean a browser control is shtting down and can no longer receive.
</blockquote></details>

Done.

JohnThomson (changes requested):
Thanks. Nice work. One more minor request.

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5878#-:-NWORjCr4gtFsTZyb5hD:b6w7rp1)* status: all files reviewed, 1 unresolved discussion (waiting on @gmartin7)

___
*[`src/BloomExe/web/BloomWebSocketServer.cs` line 207 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5878#-NWOPoHrDdYgd8dSMEiT:-NWOPoHs76ywuKrJxx89:b-pcovt2) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/bf85c070bb5bc90866a789c09aeb20264cbbaf40/src/BloomExe/web/BloomWebSocketServer.cs#L207)):*
> ```Smalltalk
> 							// In addition now that we aren't locking this whole method, we need to catch the
> 							// ConnectionNotAvailableException that we get if the socket becomes unavailable
> 							// between when we check for its presence and when we try to send to it.
> ```

Since not locking is causing some nuisance, let's capture in a comment why we are not locking it. Something like,
"Don't be tempted to reinstate the locking to fix this. We don't want to hold a lock on either the whole collection of sockets or a particular one while calling Send. The reason is that it's not certain that the Send will return before the javascript that recieves
it starts to do things. That javascript might make calls back to our server on another thread. If that results in trying to broadcast
another socket message while this thread still has the socket server locked, we'll be in a deadlock. (It's actually unlikely that the same thread that sends the message is used to handle it in Javascript, or even gets held up until the message is handled, especially in WebView2 which mostly runs in another process. But let's not risk it."

gmartin7 (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5878)* status: all files reviewed, 1 unresolved discussion (waiting on @JohnThomson)

___
*[`src/BloomExe/web/BloomWebSocketServer.cs` line 207 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5878#-NWOPoHrDdYgd8dSMEiT:-NWOhK5k998b2lXPNwoS:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/bf85c070bb5bc90866a789c09aeb20264cbbaf40/src/BloomExe/web/BloomWebSocketServer.cs#L207)):*

Since not locking is causing some nuisance, let's capture in a comment why we are not locking it. Something like,
"Don't be tempted to reinstate the locking to fix this. We don't want to hold a lock on either the whole collection of sockets or a particular one while calling Send. The reason is that it's not certain that the Send will return before the javascript that recieves
it starts to do things. That javascript might make calls back to our server on another thread. If that results in trying to broadcast
another socket message while this thread still has the socket server locked, we'll be in a deadlock. (It's actually unlikely that the same thread that sends the message is used to handle it in Javascript, or even gets held up until the message is handled, especially in WebView2 which mostly runs in another process. But let's not risk it."
</blockquote></details>

Done.

JohnThomson (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5878)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @gmartin7)

PR #5879, opened 2023-05-25
Fix test for changes to CollectionSettings.DefaultBookshelf (BL-12215)

StephenMcConnel:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5879)


hatton:
I don't know how to evaluate this, since the original test tested for a unique value, and the new version tests for null.

StephenMcConnel (commented):
The program behavior has changed to remove obsolete bookshelf tags.  To keep the same test, we would have to embed a valid branding project name and subscription code in the C# code.  I didn't want to do that even with the test project, so I changed the test to verify that obsolete bookshelves are removed.

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5879#-:-NWJhKocCRaHMTz6UIti:bo7eowh)* status: 0 of 1 files reviewed, all discussions resolved

PR #5880, opened 2023-05-25
BL-12256 fix preview dialog from being done/closeable too early

nabalone:
[5.5 Regression] Publish BloomPUB - When updating preview, Publish Progress Dialog appears done prematurely 

Repro Steps:

    [5.5.3006 BetaInternal](https://bloomlibrary.org/installersBetaInternal/BloomInstaller.5.5.3006.BetaInternal.exe) or similar version
    Select a book
    Publish -> BloomPUB
    Click "Preview"

Observed:

The Publish Progress Dialog pops up. A "Close button" appears immediately. Clicking it will close the Publish Progress Dialog, but it will not cancel the updatePreview operation. (If you debug behind the scenes, you can see that the component has ProgressState = Done even while the updatePreview is in progress.

Expected:
In 5.4, the Publish Progress Dialog looks like it's in a working state [No "Close" button"] while the updatePreview operation is ongoing. After the update preview finishes, it closes automatically (probably only if there are no errors).

[Did not implement] Or (if cheap), John Hatton said it can have some kind of Stop / Cancel functionality while the updatePreview is in progress, but I don't think updatePreview has ever had an idea of being cancellable so I'm doubtful we should bite that off for 5.5.

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5880)

### DIFF
--- a/src/BloomBrowserUI/publish/commonPublish/PublishProgressDialog.tsx
+++ b/src/BloomBrowserUI/publish/commonPublish/PublishProgressDialog.tsx
@@ -67,6 +67,7 @@ export const PublishProgressDialog: React.FunctionComponent<{
             // But the alternative gets complicated too... the weirdness here is that we need to
             // do something (change the state of the dialog) when the postData's promise is satisfied.
             // (That is, when the preview construction is complete).
+            props.setProgressState(ProgressState.Working);
             postData(
                 props.startApiEndpoint,
                 {},
--- a/src/BloomBrowserUI/publish/commonPublish/PublishProgressDialog.tsx
+++ b/src/BloomBrowserUI/publish/commonPublish/PublishProgressDialog.tsx
@@ -58,6 +58,7 @@ export const PublishProgressDialog: React.FunctionComponent<{
     }, [props.closePending]);
 
     React.useEffect(() => {
+        props.setProgressState(ProgressState.Working);
         // we need to be ready to listen to progress messages from the server,
         // before we kick anything off on the server.
         WebSocketManager.notifyReady(props.webSocketClientContext, () => {
@@ -67,7 +68,6 @@ export const PublishProgressDialog: React.FunctionComponent<{
             // But the alternative gets complicated too... the weirdness here is that we need to
             // do something (change the state of the dialog) when the postData's promise is satisfied.
             // (That is, when the preview construction is complete).
-            props.setProgressState(ProgressState.Working);
             postData(
                 props.startApiEndpoint,
                 {},


JohnThomson (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5880)* status: all files reviewed, 1 unresolved discussion (waiting on @nabalone)

___
*[`src/BloomBrowserUI/publish/commonPublish/PublishProgressDialog.tsx` line 70 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5880#-NWJChvo7A8gGpvj03Y0:-NWJChvo7A8gGpvj03Y1:bkj18d9) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/7921a4bdb47cdf2d7e5ac017a32167ca1284f729/src/BloomBrowserUI/publish/commonPublish/PublishProgressDialog.tsx#L70)):*
> ```XML
>             // do something (change the state of the dialog) when the postData's promise is satisfied.
>             // (That is, when the preview construction is complete).
>             props.setProgressState(ProgressState.Working);
> ```

The comment above is all about the postData and its outcomes, so I think this new line should go before it. Indeed, I'm not clear that there's any reason for this to wait until the notifyReady action; I don't think the difference would be very noticeable, but it would probably be fine to do it right at the start of the useEffect. NotifyReady is going to call this quite quickly, and as soon as it sends the post we really will be working. That would also put it away from the "we need to be ready" comment, which it also has nothing to do with.

hatton:
@nabalone this is a test

nabalone (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5880)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @JohnThomson)

___
*[`src/BloomBrowserUI/publish/commonPublish/PublishProgressDialog.tsx` line 70 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5880#-NWJChvo7A8gGpvj03Y0:-NWOm9Mm80kvJbaL_mRw:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/7921a4bdb47cdf2d7e5ac017a32167ca1284f729/src/BloomBrowserUI/publish/commonPublish/PublishProgressDialog.tsx#L70)):*

The comment above is all about the postData and its outcomes, so I think this new line should go before it. Indeed, I'm not clear that there's any reason for this to wait until the notifyReady action; I don't think the difference would be very noticeable, but it would probably be fine to do it right at the start of the useEffect. NotifyReady is going to call this quite quickly, and as soon as it sends the post we really will be working. That would also put it away from the "we need to be ready" comment, which it also has nothing to do with.
</blockquote></details>

Done.

JohnThomson (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5880)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @nabalone)

PR #5882, opened 2023-05-25
Ensure truncated image filenames have only single spaces (BL-12261)

StephenMcConnel:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5882)

### DIFF
--- a/src/BloomExe/Book/BookStorage.cs
+++ b/src/BloomExe/Book/BookStorage.cs
@@ -2521,23 +2521,40 @@ namespace Bloom.Book
 		}
 
 
-		public static string SanitizeNameForFileSystem(string name)
+		public static string SanitizeNameForFileSystem(string name, string defaultName = null)
 		{
 			// We want NFC to prevent Dropbox complaining about encoding conflicts.
 			// May as well do that first as it may result in less truncation.
 			name = name.Normalize(NormalizationForm.FormC);
-			// Then make sure it's not too long.
-			if (name.Length > MaxFilenameLength)
-				name = name.Substring(0, MaxFilenameLength);
 			// Then replace invalid characters with spaces and trim off characters
 			// that shouldn't start or finish a directory name.
 			name = RemoveDangerousCharacters(name);
-			if (name.Length == 0)
+			// Multiple spaces are prone to being collapsed in HTML, particularly if the name ends up in
+			// the content of some element like a data-div one, such as the coverImage src. See BL-9145 and BL-12261.
+			// Even if that doesn't happen, it's hard for humans to tell filenames like "a  b" and "a   b" apart,
+			// especially with variable width fonts. So we will just collapse them while sanitizing the name.
+			while (name.Contains("  "))
+				name = name.Replace("  ", " ");
+			// Then make sure it's not too long.
+			if (name.Length > MaxFilenameLength)
 			{
-				// The localized default book name could itself have dangerous characters.
-				name = RemoveDangerousCharacters(BookStarter.UntitledBookName);
-				if (name.Length == 0)
-					name = "Book";	// This should absolutely never be needed, but let's be paranoid.
+				name = name.Substring(0, MaxFilenameLength);
+				// Remove trailing whitespace and periods at the truncation point.
+				name = Regex.Replace(name, "[\\s.]*$", "", RegexOptions.Compiled);
+			}
+			if (String.IsNullOrWhiteSpace(name))
+			{
+				if (String.IsNullOrWhiteSpace(defaultName))
+				{
+					// The localized default book name could itself have dangerous characters.
+					name = RemoveDangerousCharacters(BookStarter.UntitledBookName);
+					if (name.Length == 0)
+						name = "Book";  // This should absolutely never be needed, but let's be paranoid.
+				}
+				else
+				{
+					name = defaultName;
+				}
 			}
 			return name;
 		}
--- a/src/BloomExe/ImageProcessing/ImageUtils.cs
+++ b/src/BloomExe/ImageProcessing/ImageUtils.cs
@@ -404,12 +404,6 @@ namespace Bloom.ImageProcessing
 				// See https://silbloom.myjetbrains.com/youtrack/issue/BL-2627 ("Weird Image Problem").
 				basename = Path.GetFileNameWithoutExtension(imageInfo.FileName);
 			}
-
-			// Multiple spaces are prone to being collapsed in HTML, particularly if the name ends up in
-			// the content of some element like a data-div one, such as the coverImage src. See BL-9145.
-			// So we will just collapse them before we save the file.
-			while (basename.Contains("  "))
-				basename = basename.Replace("  ", " ");
 			return GetUnusedFilename(bookFolderPath, basename, extension);
 		}
 
@@ -418,9 +412,9 @@ namespace Bloom.ImageProcessing
 		/// start with a period. As well as being unused, the name will be truncated enough to minimize
 		/// the danger of exceeding the maximum path length for Windows.
 		/// </summary>
-		internal static string GetUnusedFilename(string bookFolderPath, string basenameIn, string extension)
+		internal static string GetUnusedFilename(string bookFolderPath, string basenameIn, string extension, string defaultName = "image")
 		{
-			var basename = BookStorage.SanitizeNameForFileSystem(basenameIn);
+			var basename = BookStorage.SanitizeNameForFileSystem(basenameIn, defaultName);
 			// basename may already end in one or more digits. Try to strip off digits, parse and increment.
 			try
 			{
--- a/src/BloomExe/Spreadsheet/SpreadsheetImporter.cs
+++ b/src/BloomExe/Spreadsheet/SpreadsheetImporter.cs
@@ -475,7 +475,7 @@ namespace Bloom.Spreadsheet
 				while (RobustFile.Exists(dest))
 				{
 					videoFile = ImageUtils.GetUnusedFilename(videoDirectory,
-						Path.GetFileNameWithoutExtension(videoFile), Path.GetExtension(videoFile));
+						Path.GetFileNameWithoutExtension(videoFile), Path.GetExtension(videoFile), "video");
 					dest = Path.Combine(videoDirectory, videoFile);
 				}
 				// Enhance: we should probably do something to prevent multiple video elements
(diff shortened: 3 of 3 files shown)


JohnThomson (changes requested):
Sorry, I thought of a few more things that deserve thought while cleaning this up. (Inside every small problem is a large problem trying to get out.)

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5882#-:-NWJMPF2D_mbiC0wHxgx:b-uo2vfe)* status: all files reviewed, 3 unresolved discussions (waiting on @StephenMcConnel)

___
*[`src/BloomExe/Book/BookStorage.cs` line 2533 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5882#-NWJGKa6CZo17FuJBCT-:-NWJGKa6CZo17FuJBCT0:b-y5tpcw) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/c034052506122f69ddc637255785d27f21f027d1/src/BloomExe/Book/BookStorage.cs#L2533)):*
> ```Smalltalk
> 				name = name.Substring(0, MaxFilenameLength);
> 			if (nameIsAlreadyFileName)
> 				return name.Trim();	// we don't need to worry about dangerous/invalid characters from an existing filename.
> ```

I'm not convinced. For now this is only going to Windows (5.5) but we might cherry-pick, and in any case, we want to get back to Linux one day. And different characters are illegal on the two systems. I think we want Bloom file names to be OK on either.
And this is only a very slight optimization of something that doesn't need to be fast. We don't generate file names that often.
___
*[`src/BloomExe/ImageProcessing/ImageUtils.cs` line 420 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5882#-NWJIB5OC8I9ysZX-8p2:-NWJIB5PANtlRsjMP5U1:b5hq81x) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/c034052506122f69ddc637255785d27f21f027d1/src/BloomExe/ImageProcessing/ImageUtils.cs#L420)):*
> ```Smalltalk
> 			var basename = BookStorage.SanitizeNameForFileSystem(basenameIn, true);
> 			if (String.IsNullOrWhiteSpace(basename))
> 				basename = "image";
> ```

This is unlikely to happen, because Sanitize makes sure to answer 'Book' if it would otherwise answer an empty string. Of course, pathologically, it might answer an all-white-space string...probably a mistake, and another reason to follow the suggestion below.
It's rather confusing that Sanitize wants to convert empties to 'Book' and this to 'image'. But neither is any longer limited to being used for Book and/or image file names. If we put all the logic except for avoiding duplicates in the Sanitize routine, we could make the default smart based on the extension, e.g., Book for htm{l}, image for jpg/png/svg/jpeg. video for mp4, sound for mp3, any others you can think of based on callers (BloomServer.GetContentType might suggest ideas, too).
Unfortunately, if we put that into Sanitize, we have to pass it an extension. Possibly, it could be an optional argument defaulting to null, in which case the default is just "something". Ideally, callers where there's an obvious answer would be changed to pass it. There aren't a huge number.
I found at least one caller of Sanitize that passes a path that already includes an extension...in that case, the whole result will almost never be all-white, but the part before the dot might be, which would be bad, so that would be good to fix while you're cleaning up this area.
___
*[`src/BloomExe/ImageProcessing/ImageUtils.cs` line 425 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5882#-NWJFsTdDIHyHg2tuJs-:-NWJFsTdDIHyHg2tuJs0:b-7xko6t) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/c034052506122f69ddc637255785d27f21f027d1/src/BloomExe/ImageProcessing/ImageUtils.cs#L425)):*
> ```Smalltalk
> 			// So we will just collapse them before we save the file.
> 			while (basename.Contains("  "))
> 				basename = basename.Replace("  ", " ");
> ```

I wonder whether this logic should be in SanitizeNameForFileSystem. On the one hand, strictly, we don't need to prevent multiple spaces to make a name file-system-legal. On the other hand, I can't ever imagine a file name with groups of spaces being a good thing. Confusion waiting to happen, even if it doesn't get mangled by HTML. (Also it would be nice to do it before truncating, so we don't make the name any shorter than we think we need to. OTOH, we MUST do it after converting other characters to spaces. Maybe the right order is Normalize, remove dangerous, remove multiple white space, truncate if necessary, trim, do something if empty?

StephenMcConnel (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5882)* status: all files reviewed, 3 unresolved discussions (waiting on @JohnThomson)

___
*[`src/BloomExe/Book/BookStorage.cs` line 2533 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5882#-NWJGKa6CZo17FuJBCT-:-NWJTVXF9KirPRSIZGBP:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/c034052506122f69ddc637255785d27f21f027d1/src/BloomExe/Book/BookStorage.cs#L2533)):*

I'm not convinced. For now this is only going to Windows (5.5) but we might cherry-pick, and in any case, we want to get back to Linux one day. And different characters are illegal on the two systems. I think we want Bloom file names to be OK on either.
And this is only a very slight optimization of something that doesn't need to be fast. We don't generate file names that often.
</blockquote></details>

Done.
___
*[`src/BloomExe/ImageProcessing/ImageUtils.cs` line 420 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5882#-NWJIB5OC8I9ysZX-8p2:-NWJd7ymDFktXpta1dgm:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/c034052506122f69ddc637255785d27f21f027d1/src/BloomExe/ImageProcessing/ImageUtils.cs#L420)):*

This is unlikely to happen, because Sanitize makes sure to answer 'Book' if it would otherwise answer an empty string. Of course, pathologically, it might answer an all-white-space string...probably a mistake, and another reason to follow the suggestion below.
It's rather confusing that Sanitize wants to convert empties to 'Book' and this to 'image'. But neither is any longer limited to being used for Book and/or image file names. If we put all the logic except for avoiding duplicates in the Sanitize routine, we could make the default smart based on the extension, e.g., Book for htm{l}, image for jpg/png/svg/jpeg. video for mp4, sound for mp3, any others you can think of based on callers (BloomServer.GetContentType might suggest ideas, too).
Unfortunately, if we put that into Sanitize, we have to pass it an extension. Possibly, it could be an optional argument defaulting to null, in which case the default is just "something". Ideally, callers where there's an obvious answer would be changed to pass it. There aren't a huge number.
I found at least one caller of Sanitize that passes a path that already includes an extension...in that case, the whole result will almost never be all-white, but the part before the dot might be, which would be bad, so that would be good to fix while you're cleaning up this area.
</blockquote></details>

Done.
___
*[`src/BloomExe/ImageProcessing/ImageUtils.cs` line 425 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5882#-NWJFsTdDIHyHg2tuJs-:-NWJTiJ4B_CIFa_yAdJ1:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/c034052506122f69ddc637255785d27f21f027d1/src/BloomExe/ImageProcessing/ImageUtils.cs#L425)):*

I wonder whether this logic should be in SanitizeNameForFileSystem. On the one hand, strictly, we don't need to prevent multiple spaces to make a name file-system-legal. On the other hand, I can't ever imagine a file name with groups of spaces being a good thing. Confusion waiting to happen, even if it doesn't get mangled by HTML. (Also it would be nice to do it before truncating, so we don't make the name any shorter than we think we need to. OTOH, we MUST do it after converting other characters to spaces. Maybe the right order is Normalize, remove dangerous, remove multiple white space, truncate if necessary, trim, do something if empty?
</blockquote></details>

Done.

JohnThomson (approved):
Nice work. Two more small suggestions, but you've already fixed the problem and improved our code overall, so I'm going to approve it and let you decide whether it's worth doing more.

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5882#-:-NWPALSD7sMce0Hy-1V8:b8rvs4k)* status: all files reviewed, 2 unresolved discussions (waiting on @StephenMcConnel)

___
*[`src/BloomExe/Spreadsheet/SpreadsheetImporter.cs` line 1736 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5882#-NWP9QtW2OnKQ-wdwY62:-NWP9QtW2OnKQ-wdwY63:b7ovr7w) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/706827f713d7834ccfa674a30d74a8c5a2db8477/src/BloomExe/Spreadsheet/SpreadsheetImporter.cs#L1736)):*
> ```Smalltalk
> 				return "0";
> 			}
> 			var id = SanitizeXHtmlId(BookStorage.SanitizeNameForFileSystem(Path.GetFileNameWithoutExtension(audioFile)));
> ```

This would be a good place to pass a default name like "sound" or "audio".
For extra credit: could make defaultName an optional argument to GetUnusedFilename, and make use of this where SpreadsheetImporter uses it to get a name for a video file.

StephenMcConnel (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5882)* status: all files reviewed, 2 unresolved discussions (waiting on @JohnThomson)

___
*[`src/BloomExe/Spreadsheet/SpreadsheetImporter.cs` line 1736 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5882#-NWP9QtW2OnKQ-wdwY62:-NWhNdHTAgwqFpYmPTT0:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/706827f713d7834ccfa674a30d74a8c5a2db8477/src/BloomExe/Spreadsheet/SpreadsheetImporter.cs#L1736)):*

This would be a good place to pass a default name like "sound" or "audio".
For extra credit: could make defaultName an optional argument to GetUnusedFilename, and make use of this where SpreadsheetImporter uses it to get a name for a video file.
</blockquote></details>

Done.

JohnThomson (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5882)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @StephenMcConnel)

PR #5883, opened 2023-05-25
Ensure ffmpeg can read its complex filter for video recording (BL-12235)

StephenMcConnel:


hatton:
I note that there are other calls to ffmpeg elsewhere. Might they also be hiding this bug? I see that at least one does this properly, using the `CommandLineRunner` class. Can that class be modified to use your fix here? 

A good approach might be adding a `CommandLindRunner.RunCultureInvariant()` and then use invariant culture instead of `en-US`? That way you don't have to worry about any apps that rely on the current behavior. Then we can use that everywhere in Bloom.  For 5.4, we can weigh the risk of introducing new bug with the likely squashing of old bugs. For 5.5, it seems like a no-brainer.

StephenMcConnel (commented):
I changed to using InvariantCulture instead of the en-US culture.  As far as I can tell from reading the code, this is the only trouble spot where data was being written in the current culture and read in an invariant/US culture.  More extensive changes may be warranted for 5.5/5.6 to emphasize maintaining an invariant culture when spawning another process, but I hesitate to do more to 5.4 at this point when the need hasn't been manifested beyond this one issue.

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5883#-:-NWiaJxzBcfFO-jIr1Hd:b-t4399s)* status: 0 of 3 files reviewed, all discussions resolved

andrew-polk (approved):

PR #5885, opened 2023-05-29
[5.5] BL-12240 Fix broken thumbnail

gmartin7:
* Harvester thumbs and uploaded thumbs are stored in
   different places.

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5885)

### DIFF
--- a/src/BloomBrowserUI/publish/LibraryPublish/uploadCollisionDlg.tsx
+++ b/src/BloomBrowserUI/publish/LibraryPublish/uploadCollisionDlg.tsx
@@ -43,17 +43,8 @@ export interface IUploadCollisionDlgProps {
     dialogEnvironment?: IBloomDialogEnvironmentParams;
 }
 
-let legacyMode: boolean = true;
-export function showUploadCollisionDialog() {
-    // This is a hack we have in place until the old upload screen dies.
-    // The old screens opens uploadCollisionDlg as a winforms dialog and doesn't take this code path.
-    // So we use `legacyMode` to know whether we are dealing with the old screen or new.
-    legacyMode = false;
-    showUploadCollectionDialogInternal();
-}
-
-let showUploadCollectionDialogInternal: () => void = () => {
-    console.error("showUploadCollectionDialogInternal is not set up yet.");
+export let showUploadCollisionDialog: () => void = () => {
+    console.error("showUploadCollisionDialog is not set up yet.");
 };
 
 export const UploadCollisionDlg: React.FunctionComponent<IUploadCollisionDlgProps> = props => {
@@ -62,7 +53,7 @@ export const UploadCollisionDlg: React.FunctionComponent<IUploadCollisionDlgProp
         closeDialog,
         propsForBloomDialog
     } = useSetupBloomDialog(props.dialogEnvironment);
-    showUploadCollectionDialogInternal = showDialog;
+    showUploadCollisionDialog = showDialog;
 
     enum RadioState {
         Indeterminate,
@@ -225,9 +216,6 @@ export const UploadCollisionDlg: React.FunctionComponent<IUploadCollisionDlgProp
 
     function cancel() {
         props.onCancel?.();
-        if (legacyMode) {
-            post("libraryPublish/cancel_obsolete");
-        }
         closeDialog();
     }
 
@@ -368,21 +356,13 @@ export const UploadCollisionDlg: React.FunctionComponent<IUploadCollisionDlgProp
                     onClick={() => {
                         const isSameBook: boolean =
                             buttonState === RadioState.Same;
-                        if (legacyMode) {
-                            postJson("libraryPublish/upload_obsolete", {
-                                sameOrDifferent: isSameBook
-                                    ? "same"
-                                    : "different"
-                            });
-                        } else {
-                            post(
-                                `libraryPublish/${
-                                    isSameBook
-                                        ? "upload"
-                                        : "uploadAfterChangingBookId"
-                                }`
-                            );
-                        }
+                        post(
+                            `libraryPublish/${
+                                isSameBook
+                                    ? "upload"
+                                    : "uploadAfterChangingBookId"
+                            }`
+                        );
                         closeDialog();
                     }}
                 >
--- a/src/BloomExe/Book/Book.cs
+++ b/src/BloomExe/Book/Book.cs
@@ -19,7 +19,6 @@ using Bloom.Edit;
 using Bloom.FontProcessing;
 using Bloom.History;
 using Bloom.Publish;
-using Bloom.TeamCollection;
 using Bloom.ToPalaso;
 using Bloom.Utils;
 using Bloom.web.controllers;
--- a/src/BloomExe/Publish/BloomLibrary/BloomLibraryPublishModel.cs
+++ b/src/BloomExe/Publish/BloomLibrary/BloomLibraryPublishModel.cs
@@ -588,6 +588,9 @@ namespace Bloom.Publish.BloomLibrary
 
 			}
 		}
+
+		private static readonly string kThumbnailFileName = "thumbnail-256.png";
+
 		private static string GetBloomLibraryThumbnailUrl(dynamic existingBookInfo)
 		{
 			// Code basically copied from bloomlibrary2 Book.ts
@@ -600,7 +603,7 @@ namespace Bloom.Publish.BloomLibrary
 				return harvesterThumbnailUrl;
 			}
 			// Try "legacy" version (the one uploaded with the book)
-			return CreateThumbnailUrlFromBase(baseUrl, updatedTime);
+			return CreateUrlWithCacheBusting(baseUrl + BloomLibraryPublishModel.kThumbnailFileName, updatedTime);
 		}
 
 		// Code modified from bloomlibrary2 Book.ts.
@@ -612,20 +615,6 @@ namespace Bloom.Publish.BloomLibrary
 			if (harvestState != "Done")
 				return null;
 
-			var harvesterBaseUrl = GetHarvesterBaseUrl(baseUrl);
-			return CreateThumbnailUrlFromBase(harvesterBaseUrl, lastUpdate);
-		}
-
-		private static string CreateThumbnailUrlFromBase(string baseUrl, string lastUpdate)
-		{
-			// The bloomlibrary2 code calls Book.getCloudFlareUrl(), but it just passes the parameter through
-			// at this point with a bunch of comments on why we don't do anything there.
-			return baseUrl + "thumbnails/thumbnail-256.png?version=" + lastUpdate;
-		}
-
-		// Code basically copied from bloomlibrary2 Book.ts
-		private static string GetHarvesterBaseUrl(string baseUrl)
-		{
 			const string slash = "%2f";
 			var folderWithoutLastSlash = baseUrl;
 			if (baseUrl.EndsWith(slash))
@@ -635,8 +624,21 @@ namespace Bloom.Publish.BloomLibrary
 
 			var index = folderWithoutLastSlash.LastIndexOf(slash, StringComparison.InvariantCulture);
 			var pathWithoutBookName = folderWithoutLastSlash.Substring(0, index);
-			return pathWithoutBookName.Replace("BloomLibraryBooks-Sandbox", "bloomharvest-sandbox")
-					   .Replace("BloomLibraryBooks", "bloomharvest") + "/";
+			var pathToHarvestedBookFolder = pathWithoutBookName.Replace("BloomLibraryBooks-Sandbox", "bloomharvest-sandbox")
+					   .Replace("BloomLibraryBooks", "bloomharvest");
+
+			// Harvested books are stored in a separate S3 bucket from the original uploads.
+			// Harvester stores harvested thumbnails in a 'thumbnails' subfolder,
+			// whereas the orignally-uploaded book doesn't have a separate subfolder for thumbnails.
+			// BloomLibrary code depends on this location to find the thumbnails it displays.
+			return CreateUrlWithCacheBusting(
+				pathToHarvestedBookFolder + "/thumbnails/" + BloomLibraryPublishModel.kThumbnailFileName,
+				lastUpdate);
+		}
+
+		private static string CreateUrlWithCacheBusting(string url, string lastUpdate)
+		{
+			return url + "?version=" + lastUpdate;
 		}
 
 		private IEnumerable<string> ConvertLanguageCodesToNames(IEnumerable<string> languageCodesToUpload, BookData bookData)


JohnThomson (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5885)* status: all files reviewed, 1 unresolved discussion (waiting on @gmartin7)

___
*[`src/BloomExe/Publish/BloomLibrary/BloomLibraryPublishModel.cs` line 622 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5885#-NWiXmpr0KwjL_KA-ejD:-NWiXmpr0KwjL_KA-ejE:b-pyhxa2) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/fe0deb110ef2c1f5f24f7b537cc74feb27a948ed/src/BloomExe/Publish/BloomLibrary/BloomLibraryPublishModel.cs#L622)):*
> ```Smalltalk
> 		{
> 			// The bloomlibrary2 code calls Book.getCloudFlareUrl(), but it just passes the parameter through
> 			// at this point with a bunch of comments on why we don't do anything there.
> ```

I don't see the relevance of the current comment, but the modified code cries out for one that explains why the harvester should use a different URL for thumbnails than in the ordinary bloomdesktop mode.
I suppose there really IS a good reason? It would otherwise be a better fix to change whatever code requires a different thumbnail URL. The less harvester-only behavior we have, the less the chance that the harvester will do unexpected things that we don't see when just testing BloomPub maker changes.

gmartin7 (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5885)* status: 1 of 3 files reviewed, 1 unresolved discussion (waiting on @JohnThomson)

___
*[`src/BloomExe/Publish/BloomLibrary/BloomLibraryPublishModel.cs` line 622 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5885#-NWiXmpr0KwjL_KA-ejD:-NWnQ0pT3cz9LoMFLPSa:b-m5d3vk) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/fe0deb110ef2c1f5f24f7b537cc74feb27a948ed/src/BloomExe/Publish/BloomLibrary/BloomLibraryPublishModel.cs#L622)):*

I don't see the relevance of the current comment, but the modified code cries out for one that explains why the harvester should use a different URL for thumbnails than in the ordinary bloomdesktop mode.
I suppose there really IS a good reason? It would otherwise be a better fix to change whatever code requires a different thumbnail URL. The less harvester-only behavior we have, the less the chance that the harvester will do unexpected things that we don't see when just testing BloomPub maker changes.
</blockquote></details>

Steve has no memory of this code... Ha!

Changing where the Harvester puts its thumbnails involves changing Blorg code too (Blorg's `Model/Book.ts` has this code at the end of its `getHarvesterProducedThumbnailUrl()` method:
```
        return (
            Book.getCloudFlareUrl(harvesterBaseUrl) +
            `thumbnails/thumbnail-${size}.png?version=` +
            book.updatedAt
        );
```
I don't pretend to know why, other than perhaps Harvester didn't want to worry about overwriting the originally uploaded thumbs. I haven't actually found the Bloom Desktop code that writes thumbs to that directory, which is a bit bothersome.

JohnThomson (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5885)* status: all files reviewed, 1 unresolved discussion (waiting on @gmartin7)

___
*[`src/BloomExe/Publish/BloomLibrary/BloomLibraryPublishModel.cs` line 622 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5885#-NWiXmpr0KwjL_KA-ejD:-NWr7Rkf8pXI0iZi-hOn:b-lvgaqp) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/fe0deb110ef2c1f5f24f7b537cc74feb27a948ed/src/BloomExe/Publish/BloomLibrary/BloomLibraryPublishModel.cs#L622)):*

Steve has no memory of this code... Ha!

Changing where the Harvester puts its thumbnails involves changing Blorg code too (Blorg's `Model/Book.ts` has this code at the end of its `getHarvesterProducedThumbnailUrl()` method:
```
        return (
            Book.getCloudFlareUrl(harvesterBaseUrl) +
            `thumbnails/thumbnail-${size}.png?version=` +
            book.updatedAt
        );
```
I don't pretend to know why, other than perhaps Harvester didn't want to worry about overwriting the originally uploaded thumbs. I haven't actually found the Bloom Desktop code that writes thumbs to that directory, which is a bit bothersome.
</blockquote></details>

This is better, but even if no one remembers why, seems like it ought to be possible to find the code that makes the thumbnails. Ideally, it would use this method (or a modified version of it) so that the knowledge that harvester puts thumbnails in a special place is not distributed. It's also worth capturing in a comment that Blorg code depends on this special location for harvester thumbnails. If the knowledge needs to be in two places in Bloom (e.g., because one is producing a file path and the other a URL), please put comments in both places to tie them together, or see if they can share whatever  part of the algorithm is common in a new method.
The reason MIGHT be as simple as that we want all the harvested information in one place, separate from the uploaded information. I believe different S3 buckets are involved.

gmartin7 (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5885)* status: 2 of 3 files reviewed, 1 unresolved discussion (waiting on @JohnThomson)

___
*[`src/BloomExe/Publish/BloomLibrary/BloomLibraryPublishModel.cs` line 622 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5885#-NWiXmpr0KwjL_KA-ejD:-NWrKei3CNFBRiUqybsS:bmb4lgf) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/fe0deb110ef2c1f5f24f7b537cc74feb27a948ed/src/BloomExe/Publish/BloomLibrary/BloomLibraryPublishModel.cs#L622)):*
> The reason MIGHT be as simple as that we want all the harvested information in one place, separate from the uploaded information. I believe different S3 buckets are involved.

Yeah, this is the sort of thing that Andrew or Jeffrey would remember. :(

The Harvester repo's `Harvester.cs` contains the code I was looking for:
```
	/// <summary>
	/// Uploads the thumbnails to S3
	/// </summary>
	/// <param name="thumbnailInfoPath">This is a path to a TEXT file which contains information about where to find the actual thumbnails. The thumbnail paths should be written one per line in this file.</param>
	/// <param name="s3FolderLocation">The S3 path to upload to</param>
	private void UploadThumbnails(Book book, string thumbnailInfoPath, string s3FolderLocation)
	{
		string folderToUploadTo = $"{s3FolderLocation}/thumbnails";
 		...
```
Nothing about why we need to upload thumbnails to a different subdirectory.
At least I was able to simplify my solution, so only the `GetHarvesterThumbnailUrl()` method "knows" about the subfolder oddity.

jsubloom (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5885)* status: 2 of 3 files reviewed, 2 unresolved discussions (waiting on @gmartin7 and @JohnThomson)

___
*[`src/BloomExe/Publish/BloomLibrary/BloomLibraryPublishModel.cs` line 616 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5885#-NWsBBuQCVauqcZB_5vy:-NWsBBuQCVauqcZB_5vz:b-b4w4qe) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/bb5dade39d30684bb441a6ef0869d9eb535908c2/src/BloomExe/Publish/BloomLibrary/BloomLibraryPublishModel.cs#L616)):*
> ```Smalltalk
> 
> 			// Harvested books are stored in a separate S3 bucket from the original uploads.
> 			// For some historical reason, Harvester repo code stores harvested thumbnails in a thumbnails subfolder of
> ```

Since the reason is still current, I would just say:
"Harvester stores harvested thumbnails in its thumbnails subfolder,
whereas the originally-uploaded book doesn't have a separate subfolder for thumbnails."

I think you're getting too worked about about the different URL structures. There's no reason these things need to have a similar URL structure at all. The fact that they're close just happens to be coincidence.  The BloomLibrary code doesn't structure itself with that assumption either.

It's just:
The first choice is to get the harvested thumbnail at (whatever location that happens to be).
The 2nd choice is to get the thumbnail from the originally uploaded book.

gmartin7 (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5885)* status: 2 of 3 files reviewed, 2 unresolved discussions (waiting on @JohnThomson and @jsubloom)

___
*[`src/BloomExe/Publish/BloomLibrary/BloomLibraryPublishModel.cs` line 616 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5885#-NWsBBuQCVauqcZB_5vy:-NWsDxTZDKNVSkoNHov-:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/bb5dade39d30684bb441a6ef0869d9eb535908c2/src/BloomExe/Publish/BloomLibrary/BloomLibraryPublishModel.cs#L616)):*

Since the reason is still current, I would just say:
"Harvester stores harvested thumbnails in its thumbnails subfolder,
whereas the originally-uploaded book doesn't have a separate subfolder for thumbnails."

I think you're getting too worked about about the different URL structures. There's no reason these things need to have a similar URL structure at all. The fact that they're close just happens to be coincidence.  The BloomLibrary code doesn't structure itself with that assumption either.

It's just:
The first choice is to get the harvested thumbnail at (whatever location that happens to be).
The 2nd choice is to get the thumbnail from the originally uploaded book.
</blockquote></details>

Done.

jsubloom (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5885)* status: 2 of 3 files reviewed, 3 unresolved discussions (waiting on @gmartin7 and @JohnThomson)

___
*[`src/BloomExe/Publish/BloomLibrary/BloomLibraryPublishModel.cs` line 623 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5885#-NWsEJ3Y5sJrVUX6PF08:-NWsEJ3Y5sJrVUX6PF09:b-2ky6z7) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/bb5dade39d30684bb441a6ef0869d9eb535908c2/src/BloomExe/Publish/BloomLibrary/BloomLibraryPublishModel.cs#L623)):*
> ```Smalltalk
> 		}
> 
> 		private static string CreateThumbnailUrlFromBase(string baseUrl, string lastUpdate)
> ```

Wait, I was originally not going to try to make a big deal about making it parallel about Bloom Library, but I realized... this function is at the midst of the confusion.  We try to create a shared function for 2 different scenarios to create the thumbnail URL from a baseUrl, but...
1) they don't have the exact same structure (the symptom being addressed in this PR),
2) This function is written that given the baseUrl, they SHOULD have a similar / shared structure, and actually, this is a wrong / false assumption since they're just 2 independent locations to get a thumbnail from, and the Harvester location is not a book per se whereas the originally uploaded book is (obviously) a book. (Notice how BLORG doesn't have an equivalent of this function)

So actually... maybe we should blow up this function and move its logic into its callers. (At this point it's just a one-liner anyways)

On the other hand, I hate to drag this PR out... @JohnThomson , thoughts?

jsubloom (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5885)* status: 2 of 3 files reviewed, 2 unresolved discussions (waiting on @gmartin7 and @JohnThomson)

___
*[`src/BloomExe/Publish/BloomLibrary/BloomLibraryPublishModel.cs` line 623 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5885#-NWsEJ3Y5sJrVUX6PF08:-NWsHz_73WIQY5bBmssE:bpw6efe) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/bb5dade39d30684bb441a6ef0869d9eb535908c2/src/BloomExe/Publish/BloomLibrary/BloomLibraryPublishModel.cs#L623)):*

Wait, I was originally not going to try to make a big deal about making it parallel about Bloom Library, but I realized... this function is at the midst of the confusion.  We try to create a shared function for 2 different scenarios to create the thumbnail URL from a baseUrl, but...
1) they don't have the exact same structure (the symptom being addressed in this PR),
2) This function is written that given the baseUrl, they SHOULD have a similar / shared structure, and actually, this is a wrong / false assumption since they're just 2 independent locations to get a thumbnail from, and the Harvester location is not a book per se whereas the originally uploaded book is (obviously) a book. (Notice how BLORG doesn't have an equivalent of this function)

So actually... maybe we should blow up this function and move its logic into its callers. (At this point it's just a one-liner anyways)

On the other hand, I hate to drag this PR out... @JohnThomson , thoughts?
</blockquote></details>

Actually, if we want to re-use something, I see the cache-busting ?version={lastUpdate} part as shared and reasonable to re-use.  We could move the "thumbnail-256.png" part up to the caller and say that this function should take a URL instead of baseUrl. (And the method name would have to change). I think that would be a reasonable split of what logic is shared and what's not.

gmartin7 (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5885)* status: 2 of 3 files reviewed, 2 unresolved discussions (waiting on @JohnThomson)

___
*[`src/BloomExe/Publish/BloomLibrary/BloomLibraryPublishModel.cs` line 623 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5885#-NWsEJ3Y5sJrVUX6PF08:-NWsm6EODsM-Ky7r1u1m:b-mxyeud) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/bb5dade39d30684bb441a6ef0869d9eb535908c2/src/BloomExe/Publish/BloomLibrary/BloomLibraryPublishModel.cs#L623)):*

Actually, if we want to re-use something, I see the cache-busting ?version={lastUpdate} part as shared and reasonable to re-use.  We could move the "thumbnail-256.png" part up to the caller and say that this function should take a URL instead of baseUrl. (And the method name would have to change). I think that would be a reasonable split of what logic is shared and what's not. 
</blockquote></details>

Done. I like this much better. Thanks guys.

JohnThomson (approved):
LGTM at this point. @jsubloom any more thoughts?

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5885#-:-NWszXESCtCApRNThm_e:b-c3lbdv)* status: all files reviewed, 1 unresolved discussion (waiting on @jsubloom)

jsubloom (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5885#-:-NWt0LXZ7YdpogvqXQiS:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @gmartin7)

___
*[`src/BloomExe/Publish/BloomLibrary/BloomLibraryPublishModel.cs` line 623 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5885#-NWsEJ3Y5sJrVUX6PF08:-NWt0Hoa2exlYFLpgGKE:b-pe6ewx) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/bb5dade39d30684bb441a6ef0869d9eb535908c2/src/BloomExe/Publish/BloomLibrary/BloomLibraryPublishModel.cs#L623)):*

Done. I like this much better. Thanks guys.
</blockquote></details>

sounds good, thanks so much!

PR #5886, opened 2023-05-29
[5.5] BL-12237 Fix for Endpoint handler error

gmartin7:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5886)

### DIFF
--- a/src/BloomExe/web/RequestInfo.cs
+++ b/src/BloomExe/web/RequestInfo.cs
@@ -125,11 +125,11 @@ namespace Bloom.Api
 			FileStream fs;
 			if(!RobustFile.Exists(path))
 			{
-				//for audio, at least, this is not really an error. We constantly are asking if audio already exists for the current segment
-				//enhance: maybe audio should go through a different path, e.g. "/bloom/audio/somefile.wav"
-				//then this path COULD write and error
-				//Logger.WriteError("Server could not find" + path);
-				_actualContext.Response.StatusCode = 404;
+				// Earlier there was concern that we were coming here to look for .wav file existence, but that task
+				// is now handled in the "/bloom/api/audio" endpoint. So if we get here, we're looking for a different file.
+				// Besides, if we don't set HaveOutput to true (w/WriteError), we'll have other problems.
+				Logger.WriteError("Server could not find" + path, new FileNotFoundException());
+				WriteError(404, "Server could not find " + path);
 				return;
 			}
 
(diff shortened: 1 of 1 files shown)


gmartin7 (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5886)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @JohnThomson)

___
*[`src/BloomExe/web/RequestInfo.cs` line 339 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5886#-NWiUOTuBiIRweEkDvdt:-NWign4c-bdbSND6PXeA:b-owbbbf) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/77d7f668c81f84211c923c7f492af8194854febd/src/BloomExe/web/RequestInfo.cs#L339)):*

Sorry, not happy with this.
(a) it's bizarre that we have to call WriteError(500...) when the response status code is already set to that.
(b) The comment is wrong. 500 happens when the file IS found, but something goes wrong when we try to read it. Also, the 404 error for a missing file is purposely NOT logged, as a comment in ReplyWithFileContent explains.
(c) If we actually don't find the file, the StatusCode is instead set to 404, with the same failure to set HaveOutput, so you only fixed the less likely of two problems that could cause this (though perhaps something else catches it first if we don't find the file at all, so that branch of ReplyWithFileContent is never taken?)

Proposal:
ReplyWithFileContent is expected to set HaveOutput. I think it would be good if all its error paths did so, not just the success path. An easy way to do this would be to call WriteError in the relevant paths of that method, rather than just setting the StatusCode directly. This would probably make our error handling more robust in other ways, too.
</blockquote></details>

OK. The fog in here is beginning to clear... :laughing: 
Yes, I've moved the response to `ReplyWithFileContent()`. The error that JH got (during a Team Collection collaboration), was a `FileNotFoundException` at line 138 (the `RobustFile.OpenRead()` line), which is really bizarre because the previous condition should have caught a missing file.

gmartin7 (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5886)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @JohnThomson)

___
*[`src/BloomExe/web/RequestInfo.cs` line 132 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5886#-NWlycAJ0RUxx1XCngCe:-NWmiMwWEo8lWCGhULl9:bnf7bsr) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/567ce16c949ee7126c5dae358b93749896a0e0af/src/BloomExe/web/RequestInfo.cs#L132)):*

Because of the comments about audio and `.wav` files above, I had not logged an error here. But now looking at the audio api, I see that `/bloom/api/audio/` handles those now. i.e. Done. :) I'm sure now that any other way of exiting the method will pass through the `HaveOutput = true` at the bottom. There are a couple of `try-catch`es, but they don't return or throw again.
</blockquote></details>

Hmm... I see that `AudioRecording.cs` `HandleAudioFileRequest()` eventually gets down to calling this method, but it already checks for the file's existence before doing that.

JohnThomson (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5886)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @gmartin7)

PR #5895, opened 2023-06-01
Write version.txt without BOM to match advertisement (BL-12255)

JohnThomson:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5895)

### DIFF
--- a/src/BloomExe/Publish/BloomPub/BloomPubMaker.cs
+++ b/src/BloomExe/Publish/BloomPub/BloomPubMaker.cs
@@ -208,7 +208,9 @@ BookServer bookServer,
 		{
 			var sha = Book.Book.ComputeHashForAllBookRelatedFiles(pathToFileForSha);
 			var name = "version.txt"; // must match what BloomReader is looking for in NewBookListenerService.IsBookUpToDate()
-			RobustFile.WriteAllText(Path.Combine(folderForSha, name), sha, Encoding.UTF8);
+			// We send the straight string without a BOM in our advertisement, so that needs to be what we write
+			// in the file, otherwise, BR never recognizes that it already has the current version.
+			RobustFile.WriteAllText(Path.Combine(folderForSha, name), sha, new UTF8Encoding(encoderShouldEmitUTF8Identifier: false));
 			HashOfMostRecentlyCreatedBook = sha;
 		}
 
--- a/src/BloomTests/Book/BookFileFilterTests.cs
+++ b/src/BloomTests/Book/BookFileFilterTests.cs
@@ -144,6 +144,12 @@ namespace BloomTests.Book
 		}
 
 		[Test]
+		public void Filter_ForInteractive_PassesVersionDotTxt()
+		{
+			Assert.That(_filterForInteractive.FilterRelative("version.txt"), Is.True);
+		}
+
+		[Test]
 		public void Filter_PassesImagesInRootFolder()
 		{
 			Assert.That(_normalFilter.FilterRelative("green elephants.png"), Is.True);


hatton (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5895)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @JohnThomson)

___
*[`src/BloomExe/Publish/BloomPub/BloomPubMaker.cs` line 212 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5895#-NWqhibj4JwTk51zAHRY:-NWqhibkDEDjD_7vanqe:b-o3tmmg) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/cd980d1c03e9a9e046cb114d54f0637d9e0aa983/src/BloomExe/Publish/BloomPub/BloomPubMaker.cs#L212)):*
> ```Smalltalk
> 			var name = "version.txt"; // must match what BloomReader is looking for in NewBookListenerService.IsBookUpToDate()
> 			// We send the straight string without a BOM in our advertisement, so that needs to be what we write
> 			// in the file, otherwise, BR never recognizes that it already has the current version.
> ```

Is an alternative to add the BOM in the advertisement? I ask because it seems like this way, books that are already on the phone will not be updatable (if I understand what's going on correctly).

JohnThomson (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5895)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @hatton)

___
*[`src/BloomExe/Publish/BloomPub/BloomPubMaker.cs` line 212 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5895#-NWqhibj4JwTk51zAHRY:-NWr-GcSAkTlHENLpqTX:b-q5xrt5) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/cd980d1c03e9a9e046cb114d54f0637d9e0aa983/src/BloomExe/Publish/BloomPub/BloomPubMaker.cs#L212)):*

Is an alternative to add the BOM in the advertisement? I ask because it seems like this way, books that are already on the phone will not be updatable (if I understand what's going on correctly).
</blockquote></details>

Yes, that's another option. As it stands, books that are already on the device may go through one cycle of "found a new version of X. Downloading it." Then they will have the new version.txt without the BOM and will know that it is up-to-date and stop repeating.

I am actually surprised that the default is to write a file with a BOM. I am thinking it is probably new, either because of a change in .NET or just possibly a change in libpalaso (we're using RobustFile). AFAIK, we've never sent the 'advertisement' with a BOM. So I'm guessing that books already on the device are most likely NOT to have a BOM.

If you think it's worth the time, I can investigate what changed further.

The other factor in my thinking is that the BOM is invisible in most contexts. It doesn't show when you open the file in most editors, or when you view a string in the debugger. We don't need it; we know how we encode this file. So having it is wasted space and potentially confusing; I spent a while trying to figure out why newSha.equals(oldSha) returned false when both appeared to be "23rIpYlajRDt6s77FYeeEHfltMlDy9GPbmWUzd6G/TE=". So I think it is architecturally better not to have it.

If we want to definitely prevent any unnecessary downloads, another option is to enhance BloomReader so that it ignores a BOM at the start of Version.txt.

andrew-polk (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5895)* status: all files reviewed, 1 unresolved discussion (waiting on @hatton)

andrew-polk (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5895)* status: all files reviewed, 1 unresolved discussion (waiting on @hatton)

andrew-polk (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5895)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @JohnThomson)

PR #5899, opened 2023-06-01
Upgrade Bloom link in book error should open system browser (BL-12285)

JohnThomson:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5899)

### DIFF
--- a/src/BloomExe/Book/BookStorage.cs
+++ b/src/BloomExe/Book/BookStorage.cs
@@ -395,7 +395,11 @@ namespace Bloom.Book
 					string message =
 						$"<strong>{messageNewVersionNeededHeader}</strong><br/><br/><br/>" +
 						$"{messageCurrentRunningVersion}. {messageFeatureRequiresNewerVersion}<br/><br/>" +
-						$"<a href='{UrlLookup.LookupUrl(UrlType.LibrarySite, null)}/installers'>{messageDownloadLatestVersion}</a>";  // Enhance: is there a market-specific version of Bloom Library? If so, ideal to link to it somehow.
+						// If we just embed the URL, since we show this document in a plain browser without any
+						// of our code loaded (in particular, our linkHandler code in typescript), the browser
+						// control inside bloom will navigate there, which isn't what we want. The easiest
+						// way around this is to have an api which does what we want.
+						$"<a href='/bloom/api/app/showDownloadsPage'>{messageDownloadLatestVersion}</a>";
 
 					return message;
 				}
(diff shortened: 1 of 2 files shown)


StephenMcConnel (approved):
:lgtm:
I was thinking an api was probably the only way to go to avoid opening inside Bloom's internal browser window.

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5899#-:-NWt70Fs9sNt5eV_PPmT:bh054k3)* status: all files reviewed, 1 unresolved discussion (waiting on @JohnThomson)

___
*[`src/BloomExe/Book/BookStorage.cs` line 401 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5899#-NWt7JVvAgDbkrbNElHJ:-NWt7JVvAgDbkrbNElHK:b1czmo5) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/615758df9b210cc51c9cb2aa52ccb0994056ec1c/src/BloomExe/Book/BookStorage.cs#L401)):*
> ```Smalltalk
> 						// of our code loaded (in particular, our linkHandler code in typescript), the browser
> 						// control inside bloom will navigate there, which isn't what we want. The easiest
> 						// way around this is to have an api which does what we want.
> ```

Should we have a reference to the youtrack issue?  I'll leave that up to you.

JohnThomson (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5899)* status: all files reviewed, 1 unresolved discussion (waiting on @StephenMcConnel)

___
*[`src/BloomExe/Book/BookStorage.cs` line 401 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5899#-NWt7JVvAgDbkrbNElHJ:-NWwGvAm1Q34KpkvqPgh:b-8iyo0) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/615758df9b210cc51c9cb2aa52ccb0994056ec1c/src/BloomExe/Book/BookStorage.cs#L401)):*

Should we have a reference to the youtrack issue?  I'll leave that up to you.
</blockquote></details>

I think not. The issue is all about a consequence of opening the embedded control which we don't really understand and which I don't think happens in WebView2, so I don't think it has ongoing relevance to understanding why we're doing what we are.

StephenMcConnel (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5899)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @JohnThomson)

PR #5903, opened 2023-06-05
BL-12286 save epub without preview first fix

nabalone:
Fixes Publish epub throws NullReferenceException unless you first preview
Corrects progress modal behavior
Makes epub not regenerate if settings have not changed
Fixes progress messages that were not displaying in the progress modal

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5903)

### DIFF
--- a/src/BloomBrowserUI/publish/ePUBPublish/ePUBPublishScreen.tsx
+++ b/src/BloomBrowserUI/publish/ePUBPublish/ePUBPublishScreen.tsx
@@ -70,6 +70,7 @@ export const EPUBPublishScreen = () => {
     );
 };
 
+const kUpdatePreviewApi = "publish/epub/updatePreview";
 const EPUBPublishScreenInternal: React.FunctionComponent<{
     onReset: () => void;
     showPreview: boolean;
@@ -81,6 +82,8 @@ const EPUBPublishScreenInternal: React.FunctionComponent<{
     const [highlightRefresh, setHighlightRefresh] = useState(false);
     // Starting in ProgressState.Done hides the progress dialog initially.
     const [progressState, setProgressState] = useState(ProgressState.Done);
+    const [publishStarted, setPublishStarted] = useState(false);
+    const [progressDialogGeneration, setProgressDialogGeneration] = useState(0); // Increment the counter to force the PublishProgressDialog to start over (notably, apply its initial side effects again)
     const [bookUrl, setBookUrl] = useState(
         inStorybookMode
             ? window.location.protocol +
@@ -88,6 +91,10 @@ const EPUBPublishScreenInternal: React.FunctionComponent<{
                   window.location.host +
                   "/templates/Sample Shells/The Moon and the Cap" // Enhance: provide an actual epub in the source tree
             : "" // otherwise, wait for the websocket to deliver a url when the c# has finished creating the epub
+    );
+
+    const [currentTaskApi, setCurrentTaskApi] = useState<string>(
+        kUpdatePreviewApi
     );
 
     const [landscape] = useApiBoolean("publish/epub/landscape", false);
@@ -140,7 +147,13 @@ const EPUBPublishScreenInternal: React.FunctionComponent<{
                             margin-inline-end: 50px !important; // !important needed to override material button base
                         `}
                         enabled={isLicenseOK}
-                        clickApiEndpoint={"publish/epub/save"}
+                        onClick={() => {
+                            setPublishStarted(true);
+                            setCurrentTaskApi("publish/epub/save");
+                            setProgressDialogGeneration(
+                                oldValue => oldValue + 1
+                            );
+                        }}
                         hasText={true}
                         l10nKey="PublishTab.Save"
                     >
@@ -246,6 +259,8 @@ const EPUBPublishScreenInternal: React.FunctionComponent<{
         setClosePending(true);
     }, []);
 
+    const showProgressDialog = props.showPreview || publishStarted;
+
     return (
         <React.Fragment>
             <PublishScreenTemplate
@@ -257,16 +272,17 @@ const EPUBPublishScreenInternal: React.FunctionComponent<{
             >
                 {mainPanel}
             </PublishScreenTemplate>
-            {props.showPreview && (
+            {showProgressDialog && (
                 <PublishProgressDialog
                     heading={creatingHeading}
                     webSocketClientContext="publish-epub"
-                    apiForStartingTask="publish/epub/updatePreview"
+                    apiForStartingTask={currentTaskApi}
                     onTaskComplete={onTaskComplete}
                     progressState={progressState}
                     setProgressState={setProgressState}
                     closePending={closePending}
                     setClosePending={setClosePending}
+                    generation={progressDialogGeneration}
                 />
             )}
             <BookMetadataDialog />
--- a/src/BloomExe/Book/PublishSettings.cs
+++ b/src/BloomExe/Book/PublishSettings.cs
@@ -394,51 +394,12 @@ namespace Bloom.Book
 		}
 
 		
-		#region We do a comparison of settings to determine whether the Epub files need to be regenerated
-		
-		public override bool Equals(object obj)
+		// We do a comparison of settings to determine whether the Epub files need to be regenerated
+		public bool RequiresDifferentPreviewThan(EpubSettings other)
 		{
-			var other = obj as EpubSettings;
-			if (other == null)
-				return false;
-			return Equals(other);
+			return other == null || HowToPublishImageDescriptions != other.HowToPublishImageDescriptions
+				|| RemoveFontSizes != other.RemoveFontSizes || Mode != other.Mode;
 		}
-
-		protected bool Equals(EpubSettings other)
-		{
-			return HowToPublishImageDescriptions == other.HowToPublishImageDescriptions && RemoveFontSizes == other.RemoveFontSizes && Mode == other.Mode;
-		}
-
-		public override int GetHashCode()
-		{
-			unchecked
-			{
-				var hashCode = (int)HowToPublishImageDescriptions;
-				hashCode = (hashCode * 397) ^ RemoveFontSizes.GetHashCode();
-				hashCode = (hashCode * 397) ^ (Mode != null ? Mode.GetHashCode() : 0);
-				return hashCode;
-			}
-		}
-
-		public static bool operator ==(EpubSettings a, EpubSettings b)
-		{
-			if (Object.ReferenceEquals(a, b)) return true; // same object, including both null
-			// If one is null, but not both, return false. The casts are needed to prevent
-			// calling this method again recursively (and infinitely).
-			if (((object)a == null) || ((object)b == null))
-			{
-				return false;
-			}
-
-			return a.Equals(b);
-		}
-
-		public static bool operator !=(EpubSettings a, EpubSettings b)
-		{
-			return !(a == b);
-		}
-
-		#endregion
 
 	}
 
--- a/src/BloomExe/Book/PublishSettings.cs
+++ b/src/BloomExe/Book/PublishSettings.cs
@@ -392,6 +392,31 @@ namespace Bloom.Book
 			var serialized = JsonConvert.SerializeObject(this);
 			return JsonConvert.DeserializeObject<EpubSettings>(serialized);
 		}
+
+		public override bool Equals(object obj)
+		{
+			if (ReferenceEquals(null, obj)) return false;
+			if (ReferenceEquals(this, obj)) return true;
+			if (!(obj is EpubSettings)) return false;
+			var other = (EpubSettings)obj;
+			return HowToPublishImageDescriptions == other.HowToPublishImageDescriptions && RemoveFontSizes == other.RemoveFontSizes && Mode == other.Mode;
+		}
+
+		public static bool operator ==(EpubSettings a, EpubSettings b)
+		{
+			if (ReferenceEquals(null, a)) return false;
+			return a.Equals(b);
+		}
+
+		public static bool operator !=(EpubSettings a, EpubSettings b)
+		{
+			return !(a == b);
+		}
+
+		public override int GetHashCode()
+		{
+			return HowToPublishImageDescriptions.GetHashCode() + Mode.GetHashCode() + (RemoveFontSizes? 1 : 0);
+		}
 	}
 
 	public class BloomLibrarySettings
--- a/src/BloomExe/Book/PublishSettings.cs
+++ b/src/BloomExe/Book/PublishSettings.cs
@@ -393,18 +393,43 @@ namespace Bloom.Book
 			return JsonConvert.DeserializeObject<EpubSettings>(serialized);
 		}
 
+		
+		#region We do a comparison of settings to determine whether the Epub files need to be regenerated
+		
 		public override bool Equals(object obj)
 		{
-			if (ReferenceEquals(null, obj)) return false;
-			if (ReferenceEquals(this, obj)) return true;
-			if (!(obj is EpubSettings)) return false;
-			var other = (EpubSettings)obj;
+			var other = obj as EpubSettings;
+			if (other == null)
+				return false;
+			return Equals(other);
+		}
+
+		protected bool Equals(EpubSettings other)
+		{
 			return HowToPublishImageDescriptions == other.HowToPublishImageDescriptions && RemoveFontSizes == other.RemoveFontSizes && Mode == other.Mode;
+		}
+
+		public override int GetHashCode()
+		{
+			unchecked
+			{
+				var hashCode = (int)HowToPublishImageDescriptions;
+				hashCode = (hashCode * 397) ^ RemoveFontSizes.GetHashCode();
+				hashCode = (hashCode * 397) ^ (Mode != null ? Mode.GetHashCode() : 0);
+				return hashCode;
+			}
 		}
 
 		public static bool operator ==(EpubSettings a, EpubSettings b)
 		{
-			if (ReferenceEquals(null, a)) return false;
+			if (Object.ReferenceEquals(a, b)) return true; // same object, including both null
+			// If one is null, but not both, return false. The casts are needed to prevent
+			// calling this method again recursively (and infinitely).
+			if (((object)a == null) || ((object)b == null))
+			{
+				return false;
+			}
+
 			return a.Equals(b);
 		}
 
@@ -413,10 +438,8 @@ namespace Bloom.Book
 			return !(a == b);
 		}
 
-		public override int GetHashCode()
-		{
-			return HowToPublishImageDescriptions.GetHashCode() + Mode.GetHashCode() + (RemoveFontSizes? 1 : 0);
-		}
+		#endregion
+
 	}
 
 	public class BloomLibrarySettings
--- a/src/BloomExe/Publish/Epub/PublishEpubApi.cs
+++ b/src/BloomExe/Publish/Epub/PublishEpubApi.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
@@ -80,9 +80,6 @@ namespace Bloom.Publish.Epub
 		private string _pendingSaveAsPath;
 
 
-		// This constant must match the ID that is used for the listener set up in the React component ProgressBox
-		private const string kWebsocketEventId_Progress = "progress";
-
 		public EpubMaker EpubMaker { get; private set; }
 		public static Control ControlForInvoke { get; set; }
 
@@ -97,15 +94,9 @@ namespace Bloom.Publish.Epub
 			_standardProgress = new WebSocketProgress(_webSocketServer, kWebsocketContext);
 		}
 
-		// Message is presumed already localized.
-		private void ReportProgress(string message)
-		{
-			_webSocketServer.SendString(kWebsocketContext, kWebsocketEventId_Progress, message);
-		}
-
 		public void RegisterWithApiHandler(BloomApiHandler apiHandler)
 		{
-			apiHandler.RegisterEndpointHandler(kApiUrlPart + "save", HandleEpubSave, true, false);
+			apiHandler.RegisterEndpointHandler(kApiUrlPart + "save", HandleEpubSave, false, false);
 
 			apiHandler.RegisterEndpointHandler(kApiUrlPart + "epubMode", request =>
 			{
@@ -162,6 +153,35 @@ namespace Bloom.Publish.Epub
 
 		private void HandleEpubSave(ApiRequest request)
 		{
+			if (ControlForInvoke != null && ControlForInvoke.InvokeRequired)
+			{
+				ControlForInvoke.Invoke((Action)(() => getSaveAsPath(request)));
+			}
+			else
+			{
+				getSaveAsPath(request);
+			}
+						
+			// If we ARE in the middle of staging the epub...quite possible since this
+			// handler is registered with permission to execute in parallel with other
+			// API handlers, the user just has to click Save before the preview is finished...
+			// then we need not do any more here. A call to SaveAsEpub at the end of the
+			// preview generation process will pick up the pending request in _pendingSaveAsPath
+			// and complete the Save. 
+			// 
+			// Otherwise, RefreshPreview will call SaveAsEpub which will pick up the 
+			// _pendingSaveAsPath and save
+
+			if (!_stagingEpub && _pendingSaveAsPath != null) // this should only happen if an output path was chosen
+			{
+				RefreshPreview(request.CurrentBook.BookInfo.PublishSettings.Epub, false);
+			}
+
+			request.PostSucceeded();
+		}
+			
+		private void getSaveAsPath(ApiRequest request)
+		{
 			var initialPath = OutputFilenames.GetOutputFilePath(_bookSelection.CurrentSelection, ".epub");
 			var destFileName = Utils.MiscUtils.GetOutputFilePathOutsideCollectionFolder(initialPath, "ePUB files|*.epub");
 			if (!string.IsNullOrEmpty(destFileName))
@@ -170,25 +190,12 @@ namespace Bloom.Publish.Epub
 				lock (_epubMakerLock)
 				{
 					_pendingSaveAsPath = destFileName;
-					if (!_stagingEpub)
-					{
-						// we can do it right now. No need to check version etc., because anything
-						// that will change the epub we want to save will immediately trigger a new
-						// preview, and we will be staging it until we have it.
-						SaveAsEpub();
-					}
-					// If we ARE in the middle of staging the epub...quite possible since this
-					// handler is registered with permission to execute in parallel with other
-					// API handlers, the user just has to click Save before the preview is finished...
-					// then we need not do any more here. A call to SaveAsEpub at the end of the
-					// preview generation process will pick up the pending request in _pendingSaveAsPath
-					// and complete the Save.
+
 				}
 
-				ReportProgress(LocalizationManager.GetString("PublishTab.Epub.Done", "Done"));
+				_standardProgress.Message("PublishTab.Epub.Done", "Done");
 				ReportAnalytics("Save ePUB");
 			}
-			request.PostSucceeded();
 		}
 
 		public void AbortMakingEpub()
@@ -203,7 +210,7 @@ namespace Bloom.Publish.Epub
 			}
 		}
 
-		private void RefreshPreview(EpubSettings newSettings)
+		private void RefreshPreview(EpubSettings newSettings, bool forceUpdate = true)
 		{
 			// We have seen some exceptions thrown during refresh that cause a pretty yellow
 			// dialog box pop up informing the user, e.g., that the program couldn't find
@@ -215,7 +222,7 @@ namespace Bloom.Publish.Epub
 			{
 				try
 				{
-					if (UpdatePreview(newSettings, true))
+					if (UpdatePreview(newSettings, forceUpdate))
 						_webSocketServer.SendString(kWebsocketContext, kWebsocketEventId_epubReady, _previewSrc);
 					return;
 				}
@@ -394,7 +401,7 @@ namespace Bloom.Publish.Epub
 				// we will really be changing the src attr in the preview iframe so the display will update.
 				_webSocketServer.SendEvent(kWebsocketContext, kWebsocketEventId_epubReady);
 				_bookVersion = newVersion;
-				ReportProgress(LocalizationManager.GetString("PublishTab.Epub.PreparingPreview", "Preparing Preview"));
+				_progress.Message("PreparingPreview", "Preparing Preview");
 
 				// This three-tries loop is an attempt to recover from a weird state the system sometimes gets into
 				// where a browser won't navigate to a temporary page that the EpubMaker uses. I'm not sure it actually
@@ -416,10 +423,10 @@ namespace Bloom.Publish.Epub
 
 						if (i >= 2)
 							throw;
-						ReportProgress("Something went wrong, trying again");
+						_progress.MessageWithoutLocalizing("Something went wrong, trying again", ProgressKind.Error);
 						continue;
 					}
-
+					
 					break; // normal case, no exception
 				}
 
@@ -439,7 +446,7 @@ namespace Bloom.Publish.Epub
 
 			// Do pending save if the user requested it while the preview was still in progress.
 			SaveAsEpub();
-			ReportProgress(LocalizationManager.GetString("PublishTab.Epub.Done", "Done"));
+			_progress.Message("Done", "Done");
 			return true;
 		}
 	}
--- a/src/BloomExe/Publish/Epub/PublishEpubApi.cs
+++ b/src/BloomExe/Publish/Epub/PublishEpubApi.cs
@@ -37,7 +37,6 @@ namespace Bloom.Publish.Epub
 
 		private string _previewSrc;
 		private string _bookVersion;
-		private bool _stagingEpub;
 
 		// This goes out with our messages and, on the client side (typescript), messages are filtered
 		// down to the context (usually a screen) that requested them.
@@ -106,7 +105,8 @@ namespace Bloom.Publish.Epub
 					var message = new LicenseChecker().CheckBook(request.CurrentBook, request.CurrentBook.ActiveLanguages.ToArray());
 					_webSocketServer.SendString(kWebsocketContext, kWebsocketState_LicenseOK, (message == null) ? "true" : "false");
 				}
-			}, false); 
+			}, false); 	// must not be on the UI thread, because it's a long-running task and the UI thread must be available
+			// to update the progress dialog, respond to user actions, etc.
 
 			apiHandler.RegisterEndpointHandler(kApiUrlPart + "abortPreview", request =>
 			{
@@ -121,6 +121,7 @@ namespace Bloom.Publish.Epub
 
 		private void HandleEpubSave(ApiRequest request)
 		{
+			request.PostSucceeded();
 			string destPath = null;
 			if (ControlForInvoke != null && ControlForInvoke.InvokeRequired)
 			{
@@ -133,15 +134,12 @@ namespace Bloom.Publish.Epub
 
 			if (string.IsNullOrEmpty(destPath))  {
 				// No output path means the user cancelled the file chooser dialog.
-				// Simply posting a success with no other messages will cause the UI to do nothing further.
-				request.PostSucceeded();
 				return;
 			}
-				
+		
 			// Will only update the staged ePUB files if needed
 			RefreshPreview(request.CurrentBook.BookInfo.PublishSettings.Epub, forceUpdate:false);
 			SaveAsEpub(destPath);
-			request.PostSucceeded();
 		}
 			
 		private string getSaveAsPath(ApiRequest request)
@@ -157,7 +155,7 @@ namespace Bloom.Publish.Epub
 
 		public void AbortMakingEpub()
 		{
-			if (EpubMaker != null && _stagingEpub)
+			if (EpubMaker != null)
 			{
 				// typically will cause some OTHER thread that is making the epub to wind up quickly.
 				EpubMaker.AbortRequested = true;
@@ -303,7 +301,6 @@ namespace Bloom.Publish.Epub
 
 			if (EpubMaker != null)
 				EpubMaker.AbortRequested = false;
-			_stagingEpub = true;
 
 			// For some unknown reason, if the accessibility window is showing, some of the browser navigation
 			// that is needed to accurately determine which content is visible simply doesn't happen.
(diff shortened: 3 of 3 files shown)


src/BloomBrowserUI/publish/ePUBPublish/ePUBPublishScreen.tsx
nabalone, line 1: Most of these changes copy https://github.com/BloomBooks/BloomDesktop/pull/5876

src/BloomExe/Book/PublishSettings.cs
```diff
@@ -392,6 +392,31 @@ public EpubSettings Clone()
 			var serialized = JsonConvert.SerializeObject(this);
 			return JsonConvert.DeserializeObject<EpubSettings>(serialized);
 		}
+
+		public override bool Equals(object obj)
+		{
+			if (ReferenceEquals(null, obj)) return false;
```
nabalone, line 398: I was thinking that null EpubSettings should not be equal to anything, is that what we want? I'm also not sure if this line is necessary. Also on line 407

src/BloomExe/Publish/Epub/PublishEpubApi.cs
```diff
@@ -161,6 +152,35 @@ public void RegisterWithApiHandler(BloomApiHandler apiHandler)
 		}
 
 		private void HandleEpubSave(ApiRequest request)
+		{
```
nabalone, line 155: Should we have a `Debug.Assert(ControlForInvoke != null || Program.RunningUnitTests || Program.RunningOnUiThread);` here? If so, what is `Debug` and where does it come from?

src/BloomExe/Publish/Epub/PublishEpubApi.cs
```diff
@@ -161,6 +152,35 @@ public void RegisterWithApiHandler(BloomApiHandler apiHandler)
...
+			// handler is registered with permission to execute in parallel with other
+			// API handlers, the user just has to click Save before the preview is finished...
+			// then we need not do any more here. A call to SaveAsEpub at the end of the
+			// preview generation process will pick up the pending request in _pendingSaveAsPath
+			// and complete the Save. 
+			// 
+			// Otherwise, RefreshPreview will call SaveAsEpub which will pick up the 
+			// _pendingSaveAsPath and save
+
+			if (!_stagingEpub && _pendingSaveAsPath != null) // this should only happen if an output path was chosen
```
nabalone, line 175: Should this block take the `_epubMakerLock`? I think it is not necessary, as worst case we run `RefreshPreview` unnecesarily. But I may not have thought of all the race conditions. `RefreshPreview` is a pretty long block to put in a lock

src/BloomExe/Publish/Epub/PublishEpubApi.cs
```diff
@@ -170,25 +190,12 @@ private void HandleEpubSave(ApiRequest request)
...
-					// handler is registered with permission to execute in parallel with other
-					// API handlers, the user just has to click Save before the preview is finished...
-					// then we need not do any more here. A call to SaveAsEpub at the end of the
-					// preview generation process will pick up the pending request in _pendingSaveAsPath
-					// and complete the Save.
+
 				}
 
-				ReportProgress(LocalizationManager.GetString("PublishTab.Epub.Done", "Done"));
+				_standardProgress.Message("PublishTab.Epub.Done", "Done");
```
nabalone, line 196: Would it be better to do something like 
```
progress = _standardProgress.WithL10NPrefix("PublishTab.Epub.");
progress.Message("Done", "Done");
```
Similarly on lines 404 and 449
```

**Review by andrew-polk — CHANGES_REQUESTED** [review]
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5903)* status: all files reviewed, 6 unresolved discussions (waiting on @JohnThomson and @nabalone)

___
*[`src/BloomExe/Book/PublishSettings.cs` line 398 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5903#gh-1218696681:-NXH9vlk1Ub-Vd6OquYL:bviqm78) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/ff8ada1c276d1bb13c5d0e8161efd1fff9f3a8ea/src/BloomExe/Book/PublishSettings.cs#L398)):*

I was thinking that null EpubSettings should not be equal to anything, is that what we want? I'm also not sure if this line is necessary. Also on line 407
</blockquote></details>

JH just pulled out all these overrides.
I'm in discussion with him about the best way forward.
___
*[`src/BloomExe/Publish/Epub/PublishEpubApi.cs` line 175 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5903#gh-1218702041:-NXH5rw40a3dg24Pyb3h:b-w2keb6) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/ff8ada1c276d1bb13c5d0e8161efd1fff9f3a8ea/src/BloomExe/Publish/Epub/PublishEpubApi.cs#L175)):*

Should this block take the `_epubMakerLock`? I think it is not necessary, as worst case we run `RefreshPreview` unnecesarily. But I may not have thought of all the race conditions. `RefreshPreview` is a pretty long block to put in a lock
</blockquote></details>

This is a mess. (Not your fault, as it preexisted this change...)
We have a method called `RefreshPreview` which is relied upon to do the save... yuck.

I'm going to ask @JohnThomson to review this aspect and see if he can make sense of it since
1. I'm pretty sure he did all this locking, etc. to fix some issues with cancellation/restarting/leaving the page.
2. He is much better at parsing existing code than me.
3. He is also much better at understanding locking than me.
4. I'm lost.

Another existing problem I see:
We call `request.PostSucceeded();` after the user selects a destination file and the file is saved. This could easily time out.
___
*[`src/BloomExe/Publish/Epub/PublishEpubApi.cs` line 196 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5903#gh-1218703015:-NXH2b41EIeTygm_Rc6u:b-o4zdop) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/ff8ada1c276d1bb13c5d0e8161efd1fff9f3a8ea/src/BloomExe/Publish/Epub/PublishEpubApi.cs#L196)):*

Would it be better to do something like 
```
progress = _standardProgress.WithL10NPrefix("PublishTab.Epub.");
progress.Message("Done", "Done");
```
Similarly on lines 404 and 449
```
```

</blockquote></details>

Unless I'm missing something, we want all calls to use `_progress.Message...` and without the l10n prefix (as you did the other places below).
___
*[`src/BloomExe/Publish/Epub/PublishEpubApi.cs` line 197 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5903#-NXH4OEI41k1yc7aY3Ah:-NXH4OEI41k1yc7aY3Ai:bpbq0rz) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/ff8ada1c276d1bb13c5d0e8161efd1fff9f3a8ea/src/BloomExe/Publish/Epub/PublishEpubApi.cs#L197)):*
> ReportAnalytics("Save ePUB");

Is there a better place for this?
Doesn't seem like it belongs in getSaveAsPath. 
Rather it wants to be wherever we are actually saving the file.

**Comment by andrew-polk** [comment]
*[`src/BloomExe/Book/PublishSettings.cs` line 398 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5903#gh-1218696681:-NXHBe0k6fM8ll2ymcx8:bwk8ft5) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/ff8ada1c276d1bb13c5d0e8161efd1fff9f3a8ea/src/BloomExe/Book/PublishSettings.cs#L398)):*

JH just pulled out all these overrides.
I'm in discussion with him about the best way forward.
</blockquote></details>

He doesn't remember. So let's just put back what was there:
(and let's add a #region Equals overrides with a comment stating we are currently doing a comparison of settings to determine if we need to regenerate the staged files -- or whatever is accurate)
```
		public override bool Equals(object obj)
		{
			var other = obj as EpubSettings;
			if (other == null)
				return false;
			return Equals(other);
		}

		protected bool Equals(EpubSettings other)
		{
			return HowToPublishImageDescriptions == other.HowToPublishImageDescriptions && RemoveFontSizes == other.RemoveFontSizes && Mode == other.Mode;
		}

		public override int GetHashCode()
		{
			unchecked
			{
				var hashCode = (int)HowToPublishImageDescriptions;
				hashCode = (hashCode * 397) ^ RemoveFontSizes.GetHashCode();
				hashCode = (hashCode * 397) ^ (Mode != null ? Mode.GetHashCode() : 0);
				return hashCode;
			}
		}

		public static bool operator ==(EpubSettings a, EpubSettings b)
		{
			if (Object.ReferenceEquals(a, b)) return true; // same object, including both null
			// If one is null, but not both, return false. The casts are needed to prevent
			// calling this method again recursively (and infinitely).
			if (((object)a == null) || ((object)b == null))
			{
				return false;
			}

			return a.Equals(b);
		}

		public static bool operator !=(EpubSettings a, EpubSettings b)
		{
			return !(a == b);
		}
```

**Review by nabalone — COMMENTED** [review]
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5903)* status: all files reviewed, 6 unresolved discussions (waiting on @andrew-polk, @JohnThomson, and @nabalone)

___
*[`src/BloomExe/Book/PublishSettings.cs` line 398 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5903#gh-1218696681:-NXMMWguBSjFz3vmPe3O:bebrt1w) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/ff8ada1c276d1bb13c5d0e8161efd1fff9f3a8ea/src/BloomExe/Book/PublishSettings.cs#L398)):*

He doesn't remember. So let's just put back what was there:
(and let's add a #region Equals overrides with a comment stating we are currently doing a comparison of settings to determine if we need to regenerate the staged files -- or whatever is accurate)
```
		public override bool Equals(object obj)
		{
			var other = obj as EpubSettings;
			if (other == null)
				return false;
			return Equals(other);
		}

		protected bool Equals(EpubSettings other)
		{
			return HowToPublishImageDescriptions == other.HowToPublishImageDescriptions && RemoveFontSizes == other.RemoveFontSizes && Mode == other.Mode;
		}

		public override int GetHashCode()
		{
			unchecked
			{
				var hashCode = (int)HowToPublishImageDescriptions;
				hashCode = (hashCode * 397) ^ RemoveFontSizes.GetHashCode();
				hashCode = (hashCode * 397) ^ (Mode != null ? Mode.GetHashCode() : 0);
				return hashCode;
			}
		}

		public static bool operator ==(EpubSettings a, EpubSettings b)
		{
			if (Object.ReferenceEquals(a, b)) return true; // same object, including both null
			// If one is null, but not both, return false. The casts are needed to prevent
			// calling this method again recursively (and infinitely).
			if (((object)a == null) || ((object)b == null))
			{
				return false;
			}

			return a.Equals(b);
		}

		public static bool operator !=(EpubSettings a, EpubSettings b)
		{
			return !(a == b);
		}
```
</blockquote></details>

Will do
___
*[`src/BloomExe/Publish/Epub/PublishEpubApi.cs` line 175 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5903#gh-1218702041:-NXMNHmX045YJVptLevI:bd7ver6) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/ff8ada1c276d1bb13c5d0e8161efd1fff9f3a8ea/src/BloomExe/Publish/Epub/PublishEpubApi.cs#L175)):*

This is a mess. (Not your fault, as it preexisted this change...)
We have a method called `RefreshPreview` which is relied upon to do the save... yuck.

I'm going to ask @JohnThomson to review this aspect and see if he can make sense of it since
1. I'm pretty sure he did all this locking, etc. to fix some issues with cancellation/restarting/leaving the page.
2. He is much better at parsing existing code than me.
3. He is also much better at understanding locking than me.
4. I'm lost.

Another existing problem I see:
We call `request.PostSucceeded();` after the user selects a destination file and the file is saved. This could easily time out.
</blockquote></details>

- Should I simply delete the `request.PostSucceeded` line or do we need to make sure to do that elsewhere?
- Looks like a lot of the locking and `pending_save_as_path` was in case the user clicked Save before the preview had finished, which shouldn't be possible now because the Progress Modal blocks the screen while the preview is generating. What are there other multi-threading race conditions we are worried about? The leaving the page scenario?
- I agree that this file is getting pretty tangled. Reassuring to hear because it took me a long time to parse. I would be happy to take a stab at refactoring it if that's what we want to do
- It does feel a little scary that various preview methods are relied upon to do things not suggested by their names
___
*[`src/BloomExe/Publish/Epub/PublishEpubApi.cs` line 196 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5903#gh-1218703015:-NXMRZbW56uI845x7uwG:b1dh227) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/ff8ada1c276d1bb13c5d0e8161efd1fff9f3a8ea/src/BloomExe/Publish/Epub/PublishEpubApi.cs#L196)):*

Unless I'm missing something, we want all calls to use `_progress.Message...` and without the l10n prefix (as you did the other places below).
</blockquote></details>

When moving the ReportAnalytics per your comment below I deleted this message. Is it helpful to anyone for us to say "Done" right after the user selects a save path?
___
*[`src/BloomExe/Publish/Epub/PublishEpubApi.cs` line 197 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5903#-NXH4OEI41k1yc7aY3Ah:-NXMPouK9fsL3auwwVV2:bebrt1w) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/ff8ada1c276d1bb13c5d0e8161efd1fff9f3a8ea/src/BloomExe/Publish/Epub/PublishEpubApi.cs#L197)):*

> ReportAnalytics("Save ePUB");

Is there a better place for this?
Doesn't seem like it belongs in getSaveAsPath. 
Rather it wants to be wherever we are actually saving the file.
</blockquote></details>

Will do

nabalone (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5903)* status: all files reviewed, 6 unresolved discussions (waiting on @andrew-polk and @JohnThomson)

___
*[`src/BloomExe/Publish/Epub/PublishEpubApi.cs` line 175 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5903#gh-1218702041:-NXMdgYJAuX_FXd1vC7Q:bfkxgff) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/ff8ada1c276d1bb13c5d0e8161efd1fff9f3a8ea/src/BloomExe/Publish/Epub/PublishEpubApi.cs#L175)):*

- Should I simply delete the `request.PostSucceeded` line or do we need to make sure to do that elsewhere?
- Looks like a lot of the locking and `pending_save_as_path` was in case the user clicked Save before the preview had finished, which shouldn't be possible now because the Progress Modal blocks the screen while the preview is generating. What are there other multi-threading race conditions we are worried about? The leaving the page scenario?
- I agree that this file is getting pretty tangled. Reassuring to hear because it took me a long time to parse. I would be happy to take a stab at refactoring it if that's what we want to do
- It does feel a little scary that various preview methods are relied upon to do things not suggested by their names
</blockquote></details>

I see that we do need the request.PostSucceeded. I'm not sure where to put it if we keep the current thread-friendly setup. What would happen if it timed out?

JohnThomson (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5903)* status: 1 of 3 files reviewed, 6 unresolved discussions (waiting on @andrew-polk, @JohnThomson, and @nabalone)

___
*[`src/BloomExe/Publish/Epub/PublishEpubApi.cs` line 175 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5903#gh-1218702041:-NXMffWLAbV6djWRbU84:b-ep3ft8) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/ff8ada1c276d1bb13c5d0e8161efd1fff9f3a8ea/src/BloomExe/Publish/Epub/PublishEpubApi.cs#L175)):*

I see that we do need the request.PostSucceeded. I'm not sure where to put it if we keep the current thread-friendly setup. What would happen if it timed out? 
</blockquote></details>

My best recollection is that we mainly needed locking because at one point, we started regenerating the preview any time any setting was changed that affected it. I believe now we do it when the user clicks the Preview or Save button, and until we implement Cancel, it's going to run to completion. So we probably don't need locking at all.

It's probably best to call PostSucceeded before we even start creating the Preview, unless we're depending on it to know that the Preview is complete. But it's probably best (if we're not already) to communicate that through a websocket message.

Hope that makes sense. I haven't reread all the code at this point, I'm just going on what I remember.

andrew-polk (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5903)* status: all files reviewed, 6 unresolved discussions (waiting on @JohnThomson and @nabalone)

___
*[`src/BloomExe/Publish/Epub/PublishEpubApi.cs` line 155 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5903#gh-1218701036:-NXMj19rCGQDCIIzJbnE:b-mlclnp) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/ff8ada1c276d1bb13c5d0e8161efd1fff9f3a8ea/src/BloomExe/Publish/Epub/PublishEpubApi.cs#L155)):*

Should we have a `Debug.Assert(ControlForInvoke != null || Program.RunningUnitTests || Program.RunningOnUiThread);` here? If so, what is `Debug` and where does it come from?
</blockquote></details>

I think we would also get to this code when creating artifacts for harvester, so you would need a condition for that. Not sure what it buys us.
Do you ask because there is similar code elsewhere which has this?
`Debug.Assert` runs an assertion only for debug builds. So it is kinda like a sanity check for developers which doesn't cause messaging or crashing during production.
___
*[`src/BloomExe/Publish/Epub/PublishEpubApi.cs` line 196 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5903#gh-1218703015:-NXMilpd70cAv8SD5Id0:b-ep13p5) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/ff8ada1c276d1bb13c5d0e8161efd1fff9f3a8ea/src/BloomExe/Publish/Epub/PublishEpubApi.cs#L196)):*

When moving the ReportAnalytics per your comment below I deleted this message. Is it helpful to anyone for us to say "Done" right after the user selects a save path?
</blockquote></details>

Good call. I guess we're still sending this message when the generation of the preview completes?

andrew-polk (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5903)* status: all files reviewed, 6 unresolved discussions (waiting on @JohnThomson and @nabalone)

nabalone (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5903)* status: 2 of 3 files reviewed, 4 unresolved discussions (waiting on @andrew-polk and @JohnThomson)

___
*[`src/BloomExe/Book/PublishSettings.cs` line 398 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5903#gh-1218696681:-NXaig_mAv3WrT9x75PF:bb6un1j) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/ff8ada1c276d1bb13c5d0e8161efd1fff9f3a8ea/src/BloomExe/Book/PublishSettings.cs#L398)):*

Will do
</blockquote></details>

Done
___
*[`src/BloomExe/Publish/Epub/PublishEpubApi.cs` line 155 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5903#gh-1218701036:-NXaili20hob_Xnu_PCd:b-v4y8gi) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/ff8ada1c276d1bb13c5d0e8161efd1fff9f3a8ea/src/BloomExe/Publish/Epub/PublishEpubApi.cs#L155)):*

I think we would also get to this code when creating artifacts for harvester, so you would need a condition for that. Not sure what it buys us.
Do you ask because there is similar code elsewhere which has this?
`Debug.Assert` runs an assertion only for debug builds. So it is kinda like a sanity check for developers which doesn't cause messaging or crashing during production.
</blockquote></details>

Yes, I saw Debug.Assert statement in similar code elsewhere. Couldn't figure out where Debug comes from/what I need to import into the file to make it work.

Do I need to do anything else to handle the harvester case here? CreateEpubArtifact takes in a control argument and sets 			maker.ControlForInvoke = control;
___
*[`src/BloomExe/Publish/Epub/PublishEpubApi.cs` line 175 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5903#gh-1218702041:-NXaoihl1wwgHEMtPtg6:b9mvtg9) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/ff8ada1c276d1bb13c5d0e8161efd1fff9f3a8ea/src/BloomExe/Publish/Epub/PublishEpubApi.cs#L175)):*

My best recollection is that we mainly needed locking because at one point, we started regenerating the preview any time any setting was changed that affected it. I believe now we do it when the user clicks the Preview or Save button, and until we implement Cancel, it's going to run to completion. So we probably don't need locking at all.

It's probably best to call PostSucceeded before we even start creating the Preview, unless we're depending on it to know that the Preview is complete. But it's probably best (if we're not already) to communicate that through a websocket message.

Hope that makes sense. I haven't reread all the code at this point, I'm just going on what I remember.
</blockquote></details>

I have removed the locks. Ebook preview generation ends pretty quickly when you click away from the Epub screen
___
*[`src/BloomExe/Publish/Epub/PublishEpubApi.cs` line 196 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5903#gh-1218703015:-NXaoyrtCLEhGFRvCOp4:bx0dkzw) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/ff8ada1c276d1bb13c5d0e8161efd1fff9f3a8ea/src/BloomExe/Publish/Epub/PublishEpubApi.cs#L196)):*

Good call. I guess we're still sending this message when the generation of the preview completes?
</blockquote></details>

Yes, we still send "Done" after preview generation completes
___
*[`src/BloomExe/Publish/Epub/PublishEpubApi.cs` line 197 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5903#-NXH4OEI41k1yc7aY3Ah:-NXaoswj7OWogQXnuW5x:bb6un1j) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/ff8ada1c276d1bb13c5d0e8161efd1fff9f3a8ea/src/BloomExe/Publish/Epub/PublishEpubApi.cs#L197)):*

Will do
</blockquote></details>

Done

nabalone (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5903)* status: 2 of 3 files reviewed, 3 unresolved discussions (waiting on @andrew-polk and @JohnThomson)

___
*[`src/BloomExe/Publish/Epub/PublishEpubApi.cs` line 135 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5903#-NXkDgm04NTkGodInu-5:-NXkSyD52RJP3c9OXfmJ:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/0986079ff00e7eb2206bf473f5b8c23d9f343988/src/BloomExe/Publish/Epub/PublishEpubApi.cs#L135)):*

Maybe worth expounding a little.
Is this the case where the user cancelled during file selection?
Then the UI will know to simply do nothing because it never receives any websocket messages?

So something like:
> No output path means the user cancelled the file chooser dialog. 
> Simply posting a success with no other messages will cause the UI to do nothing further. 

?
</blockquote></details>

Done.
___
*[`src/BloomExe/Publish/Epub/PublishEpubApi.cs` line 140 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5903#-NXkD-5WCBt9_1C8ypLd:-NXkT01-2EP5Q6LSgIgv:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/0986079ff00e7eb2206bf473f5b8c23d9f343988/src/BloomExe/Publish/Epub/PublishEpubApi.cs#L140)):*

Maybe worth a comment like:
> // Will only update the staged ePUB files if needed

And then you can also help document the behavior by making this line:
> RefreshPreview(request.CurrentBook.BookInfo.PublishSettings.Epub, forceUpdate:false);
</blockquote></details>

Done.

JohnThomson (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5903)* status: all files reviewed, 6 unresolved discussions (waiting on @andrew-polk and @nabalone)

___
*[`src/BloomExe/Book/PublishSettings.cs` line 397 at r4](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5903#-NXkfbQnBv2wDOJCSyuS:-NXkfbQnBv2wDOJCSyuT:btypxl8) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/e2e33bc1d81e91c817e56042176a48103dfbdb03/src/BloomExe/Book/PublishSettings.cs#L397)):*
> ```Smalltalk
> 
> 		
> 		#region We do a comparison of settings to determine whether the Epub files need to be regenerated
> ```

This is a good start at explaining what is going on, but it needs a little more. Typically, Equals is overridden because you want objects to be equal if all of their fields are equal (rather than only if they are the same object). So it's quite unusual to only take some fields into account. The comment above hints that that the only reason to test equality of two sets of settings is to find out whether they differ in a way that would require the preview to be rebuilt; but that should be said more explicitly. Maybe also spell out why a change of e.g. HowToPublishImageDescriptions would NOT require rebuilding the preview.
You could also consider, instead of redefininig equality, making a method that is clearly specific to this purpose, like RequiresDiffererentPreviewThan(PublishSettings other). Unfortunately that requires finding all the places that are using equality testing for this purpose, which may be difficult. Up to you both whether you like the idea and whether it is worth the effort.
___
*[`src/BloomExe/Publish/Epub/PublishEpubApi.cs` line 142 at r4](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5903#-NXkj00o20egXe_l7ZOB:-NXkj00o20egXe_l7ZOC:b-8u908u) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/e2e33bc1d81e91c817e56042176a48103dfbdb03/src/BloomExe/Publish/Epub/PublishEpubApi.cs#L142)):*
> ```Smalltalk
> 				
> 			// Will only update the staged ePUB files if needed
> 			RefreshPreview(request.CurrentBook.BookInfo.PublishSettings.Epub, forceUpdate:false);
> ```

Is it possible for this to fail, so that there is no preview to save? For example, if the book uses  a font we're not allowed to embed, or enterprise features when that is turned off...Should we detect that here?
___
*[`src/BloomExe/Publish/Epub/PublishEpubApi.cs` line 144 at r4](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5903#-NXkjbFR62D6gsxJF-W1:-NXkjbFR62D6gsxJF-W2:b9hfjuz) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/e2e33bc1d81e91c817e56042176a48103dfbdb03/src/BloomExe/Publish/Epub/PublishEpubApi.cs#L144)):*
> ```Smalltalk
> 			RefreshPreview(request.CurrentBook.BookInfo.PublishSettings.Epub, forceUpdate:false);
> 			SaveAsEpub(destPath);
> 			request.PostSucceeded();
> ```

At one point we floated the idea of doing this early, for fear of a timeout while the user is picking a file name, or while we're generating the Epub for a very large book. Did you conclude that was not a good idea? (If you move it before getSaveAsPath(), remember to remove it from the Cancel path as well.)
___
*[`src/BloomExe/Publish/Epub/PublishEpubApi.cs` line 150 at r4](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5903#-NXkhpRKBHPRr1XixVna:-NXkhpRKBHPRr1XixVnb:b6pimos) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/01ca7b16840b79935a80546d4dbf6a9b390a988f/src/BloomExe/Publish/Epub/PublishEpubApi.cs#L150)):*
> ```Smalltalk
> 					_webSocketServer.SendString(kWebsocketContext, kWebsocketState_LicenseOK, (message == null) ? "true" : "false");
> 				}
> 			}, false); // in fact, must NOT be on UI thread
> ```

I don't think this has stopped being true, even though one reason went away...in fact we still throw if it happens. It's also not obvious, so I think the comment is worth keeping...maybe even improving, something like "must not be on the UI thread, because it's a long-running task and the UI thread must be available to update the progress dialog, respond to user actions, etc."
___
*[`src/BloomExe/Publish/Epub/PublishEpubApi.cs` line 306 at r4](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5903#-NXkks4j6JueIPvLVOVH:-NXkks4j6JueIPvLVOVI:b-3c91tq) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/e2e33bc1d81e91c817e56042176a48103dfbdb03/src/BloomExe/Publish/Epub/PublishEpubApi.cs#L306)):*
> ```Smalltalk
> 			if (EpubMaker != null)
> 				EpubMaker.AbortRequested = false;
> 			_stagingEpub = true;
> ```

Do we still need this variable? Seems like it was all about the detection of multiple threads trying to update the same epub at the same time, and if that can no longer happen, we might be able to get rid of it along with the locking.

andrew-polk (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5903)* status: all files reviewed, 6 unresolved discussions (waiting on @JohnThomson and @nabalone)

___
*[`src/BloomExe/Book/PublishSettings.cs` line 397 at r4](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5903#-NXkfbQnBv2wDOJCSyuS:-NXknZAU9sBvswgEUN7-:b-kql7o) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/e2e33bc1d81e91c817e56042176a48103dfbdb03/src/BloomExe/Book/PublishSettings.cs#L397)):*
> Maybe also spell out why a change of e.g. HowToPublishImageDescriptions would NOT require rebuilding the preview.

I don't see that we are excluding that property from the check.

> You could also consider, instead of redefininig equality, making a method that is clearly specific to this purpose, like RequiresDiffererentPreviewThan(PublishSettings other). Unfortunately that requires finding all the places that are using equality testing for this purpose, which may be difficult. Up to you both whether you like the idea and whether it is worth the effort.

I really like this idea. Not overriding equals is much simpler.
By the way, John, she is only putting back the override methods which JH removed during his PublishSettings cleanup. But he (apparently) didn't realize we needed it to prevent extraneous preview creations.

JohnThomson (changes requested):
> Reviewed 1 of 3 files at r1, 1 of 2 files at r2, 1 of 1 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5903)* status: all files reviewed, 6 unresolved discussions (waiting on @andrew-polk and @nabalone)

___
*[`src/BloomExe/Book/PublishSettings.cs` line 397 at r4](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5903#-NXkfbQnBv2wDOJCSyuS:-NXixebN3HY0lgYrFsli:b-o5qh6c) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/e2e33bc1d81e91c817e56042176a48103dfbdb03/src/BloomExe/Book/PublishSettings.cs#L397)):*

> Maybe also spell out why a change of e.g. HowToPublishImageDescriptions would NOT require rebuilding the preview.

I don't see that we are excluding that property from the check.

> You could also consider, instead of redefininig equality, making a method that is clearly specific to this purpose, like RequiresDiffererentPreviewThan(PublishSettings other). Unfortunately that requires finding all the places that are using equality testing for this purpose, which may be difficult. Up to you both whether you like the idea and whether it is worth the effort.

I really like this idea. Not overriding equals is much simpler.
By the way, John, she is only putting back the override methods which JH removed during his PublishSettings cleanup. But he (apparently) didn't realize we needed it to prevent extraneous preview creations.
</blockquote></details>

My bad. I have no idea what I was thinking. Maybe I was using a small window and could only see part of the code? I was in a bit of a hurry.
So, what you have is a perfectly reasonable version of Equals and needs no extra commenting. However, that makes two of us that like the RequiresDiffererentPreviewThan idea, and (since JohnH was removing the Equals override) it's clear that he also found it verbose and confusing...to the point that could not see why it was needed at all. RequiresDiffererentPreviewThan would just need the current definition of Equals, and all the related stuff (overriding operators and GetHashCode etc) can go. The only tricky thing is finding where we are using Equals to mean RequiresDiffererentPreviewThan. Hopefully that's clear from your earlier work or the previous history of this class.

JohnThomson (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5903#-:-NXq3_TR-ih6wxEdo0IS:bnfp4nl)* status: all files reviewed, 6 unresolved discussions (waiting on @andrew-polk and @nabalone)

nabalone (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5903)* status: all files reviewed, 6 unresolved discussions (waiting on @andrew-polk and @JohnThomson)

___
*[`src/BloomExe/Book/PublishSettings.cs` line 397 at r4](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5903#-NXkfbQnBv2wDOJCSyuS:-NXpzj0W6ioDYSp4kZUk:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/e2e33bc1d81e91c817e56042176a48103dfbdb03/src/BloomExe/Book/PublishSettings.cs#L397)):*

My bad. I have no idea what I was thinking. Maybe I was using a small window and could only see part of the code? I was in a bit of a hurry.
So, what you have is a perfectly reasonable version of Equals and needs no extra commenting. However, that makes two of us that like the RequiresDiffererentPreviewThan idea, and (since JohnH was removing the Equals override) it's clear that he also found it verbose and confusing...to the point that could not see why it was needed at all. RequiresDiffererentPreviewThan would just need the current definition of Equals, and all the related stuff (overriding operators and GetHashCode etc) can go. The only tricky thing is finding where we are using Equals to mean RequiresDiffererentPreviewThan. Hopefully that's clear from your earlier work or the previous history of this class.
</blockquote></details>

Done.
___
*[`src/BloomExe/Publish/Epub/PublishEpubApi.cs` line 142 at r4](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5903#-NXkj00o20egXe_l7ZOB:-NXprucI2VkvoUXDYWtB:b-828vt6) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/e2e33bc1d81e91c817e56042176a48103dfbdb03/src/BloomExe/Publish/Epub/PublishEpubApi.cs#L142)):*

Is it possible for this to fail, so that there is no preview to save? For example, if the book uses  a font we're not allowed to embed, or enterprise features when that is turned off...Should we detect that here?
</blockquote></details>

What should we do differently if the RefreshPreview fails? I assume that right now it'll behave the same as if you simply preview (which calls RefreshPreview) and then it fails
___
*[`src/BloomExe/Publish/Epub/PublishEpubApi.cs` line 144 at r4](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5903#-NXkjbFR62D6gsxJF-W1:-NXpsaJN9DnFuKDwhOD7:b-dipmg) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/e2e33bc1d81e91c817e56042176a48103dfbdb03/src/BloomExe/Publish/Epub/PublishEpubApi.cs#L144)):*

At one point we floated the idea of doing this early, for fear of a timeout while the user is picking a file name, or while we're generating the Epub for a very large book. Did you conclude that was not a good idea? (If you move it before getSaveAsPath(), remember to remove it from the Cancel path as well.)
</blockquote></details>

Hmm, I think I did not properly understand the concern the first time. I don't know of any reason not to put it earlier. Done
___
*[`src/BloomExe/Publish/Epub/PublishEpubApi.cs` line 150 at r4](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5903#-NXkhpRKBHPRr1XixVna:-NXpuxh146b-E2h9GxbR:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/01ca7b16840b79935a80546d4dbf6a9b390a988f/src/BloomExe/Publish/Epub/PublishEpubApi.cs#L150)):*

I don't think this has stopped being true, even though one reason went away...in fact we still throw if it happens. It's also not obvious, so I think the comment is worth keeping...maybe even improving, something like "must not be on the UI thread, because it's a long-running task and the UI thread must be available to update the progress dialog, respond to user actions, etc."
</blockquote></details>

Done.
___
*[`src/BloomExe/Publish/Epub/PublishEpubApi.cs` line 306 at r4](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5903#-NXkks4j6JueIPvLVOVH:-NXpuzK_AQl4Oz1sH2oe:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/e2e33bc1d81e91c817e56042176a48103dfbdb03/src/BloomExe/Publish/Epub/PublishEpubApi.cs#L306)):*

Do we still need this variable? Seems like it was all about the detection of multiple threads trying to update the same epub at the same time, and if that can no longer happen, we might be able to get rid of it along with the locking.
</blockquote></details>

Done.

andrew-polk (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5903)* status: all files reviewed, 6 unresolved discussions (waiting on @JohnThomson)

JohnThomson (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5903)* status: all files reviewed, 4 unresolved discussions (waiting on @andrew-polk and @nabalone)

___
*[`src/BloomExe/Publish/Epub/PublishEpubApi.cs` line 142 at r4](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5903#-NXkj00o20egXe_l7ZOB:-NXq5wSOCW95s5SRBjZw:bbzckd8) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/e2e33bc1d81e91c817e56042176a48103dfbdb03/src/BloomExe/Publish/Epub/PublishEpubApi.cs#L142)):*

What should we do differently if the RefreshPreview fails? I assume that right now it'll behave the same as if you simply preview (which calls RefreshPreview) and then it fails
</blockquote></details>

Hopefully it will have logged some error message to the Progress dialog. The only other thing would be for it NOT to attempt to zip up an epub file out of any incomplete set of files that may have ended up in the preview folder (and don't ask the user to pick a file to save to...but I think we already did that?)

nabalone (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5903)* status: all files reviewed, 4 unresolved discussions (waiting on @andrew-polk and @JohnThomson)

___
*[`src/BloomExe/Publish/Epub/PublishEpubApi.cs` line 142 at r4](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5903#-NXkj00o20egXe_l7ZOB:-NXq7Rqa24Zdd7KweaJr:b1qzjqm) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/e2e33bc1d81e91c817e56042176a48103dfbdb03/src/BloomExe/Publish/Epub/PublishEpubApi.cs#L142)):*

Hopefully it will have logged some error message to the Progress dialog. The only other thing would be for it NOT to attempt to zip up an epub file out of any incomplete set of files that may have ended up in the preview folder (and don't ask the user to pick a file to save to...but I think we already did that?)
</blockquote></details>

Now that we aren't locking, we could do the refresh preview before we ask the user to select the file. My gut reaction is that this would be a worse user experience, but what do you think? 

Regardless, I will make it not save if the refresh preview failed

andrew-polk (approved):
> Reviewed 1 of 1 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5903)* status: all files reviewed, 5 unresolved discussions (waiting on @JohnThomson)

PR #5904, opened 2023-06-06
Run external programs with invariant culture when appropriate (BL-12235)

StephenMcConnel:
(The Help viewer should run in the normal user culture, but not programs that do file processing like ffmpeg, ghostscript, and GraphicsMagick.)

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5904)

### DIFF
--- a/src/BloomExe/BloomExe.csproj
+++ b/src/BloomExe/BloomExe.csproj
@@ -404,6 +404,7 @@
     <Compile Include="History\CollectionHistory.cs" />
     <Compile Include="History\HistoryEvent.cs" />
     <Compile Include="Publish\Epub\ReadiumManifest.cs" />
+    <Compile Include="ToPalaso\CommandLineRunnerExtra.cs" />
     <Compile Include="ToPalaso\RobustIO.cs" />
     <Compile Include="Utils\AsyncUtil.cs" />
     <Compile Include="Publish\PDF\PublishPdfApi.cs" />
--- a/src/BloomExe/Book/BookStorage.cs
+++ b/src/BloomExe/Book/BookStorage.cs
@@ -1806,6 +1806,8 @@ namespace Bloom.Book
 					b.AppendLine(msg2);
 					ErrorReport.NotifyUserOfProblem(b.ToString());
 					// Open the file explorer on the book's folder to allow the user to delete (or move?) the unwanted file.
+					// The new process should use the current culture, so we don't need to worry about that.
+					// We don't wait for this to finish, so we don't use the CommandLineRunner methods.
 					Process.Start(new ProcessStartInfo {
 						FileName = FolderPath + Path.DirectorySeparatorChar,	// trailing separator char is important!
 						UseShellExecute = true,
--- a/src/BloomExe/Edit/LameEncoder.cs
+++ b/src/BloomExe/Edit/LameEncoder.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.IO;
+using Bloom.ToPalaso;
 using L10NSharp;
 using SIL.CommandLineProcessing;
 using SIL.IO;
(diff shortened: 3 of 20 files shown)


andrew-polk (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5904)* status: all files reviewed, 1 unresolved discussion (waiting on @StephenMcConnel)

___
*[`src/BloomExe/ImageProcessing/ImageUtils.cs` line 1005 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5904#-NXHaMt19mDuB2uqyZIY:-NXHaMt19mDuB2uqyZIZ:bno98a0) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/2e8bf9f7d583869e42cfa8f2c4d4b5b4c36e6b6c/src/BloomExe/ImageProcessing/ImageUtils.cs#L1005)):*
> ```Smalltalk
> 				if (destPath != safeDestPath)
> 					RobustFile.Copy(safeDestPath, destPath, true);
> 				return proc;
> ```

In each of the cases where we don't use the new wrapping method (or a similar new one for `Start()`), I think we should probably document what we would like to do and why we can't. Or is it obvious in some way I'm missing?

And in each case, if there is a reason to wrap the culture change further out than just the `Start()` call, should we document that in comments, too?

I'm thinking we want to guide devs into the right path, and the more obvious we make it, the better. So next time someone models new code off of each of these, we want them to make the right decision.

StephenMcConnel (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5904)* status: 0 of 21 files reviewed, 1 unresolved discussion (waiting on @andrew-polk)

___
*[`src/BloomExe/ImageProcessing/ImageUtils.cs` line 1005 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5904#-NXHaMt19mDuB2uqyZIY:-NXMpu90EFB0qyycP-n-:b467l80) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/2e8bf9f7d583869e42cfa8f2c4d4b5b4c36e6b6c/src/BloomExe/ImageProcessing/ImageUtils.cs#L1005)):*

In each of the cases where we don't use the new wrapping method (or a similar new one for `Start()`), I think we should probably document what we would like to do and why we can't. Or is it obvious in some way I'm missing?

And in each case, if there is a reason to wrap the culture change further out than just the `Start()` call, should we document that in comments, too?

I'm thinking we want to guide devs into the right path, and the more obvious we make it, the better. So next time someone models new code off of each of these, we want them to make the right decision.
</blockquote></details>

Done.  I've added an extension method for CommandLineRunner.StartWithInvariantCulture and replaced a few explicit calls to Process.Start.

andrew-polk (approved):
> Reviewed 21 of 21 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5904)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @StephenMcConnel)

___
*[`src/BloomExe/ImageProcessing/ImageUtils.cs` line 1005 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5904#-NXHaMt19mDuB2uqyZIY:-NXMwRMC5AWlScflDTIy:b-ucfkn4) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/2e8bf9f7d583869e42cfa8f2c4d4b5b4c36e6b6c/src/BloomExe/ImageProcessing/ImageUtils.cs#L1005)):*

Done.  I've added an extension method for CommandLineRunner.StartWithInvariantCulture and replaced a few explicit calls to Process.Start.
</blockquote></details>

Thanks.

PR #5905, opened 2023-06-06
Add missing audio highlighting CSS rules for padded sentences (BL-12287)

StephenMcConnel:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5905)

### DIFF
--- a/src/BloomBrowserUI/bookEdit/StyleEditor/StyleEditor.ts
+++ b/src/BloomBrowserUI/bookEdit/StyleEditor/StyleEditor.ts
@@ -518,6 +518,10 @@ export default class StyleEditor {
     // What we stick after something like ".normal-style" to make a selector that targets
     // the current audio element (or its paragraph children).
     private sentenceHiliteRuleSelector = " span.ui-audioCurrent";
+    // Spans with padding get split into multiple embedded spans, so we need to target the
+    // inner span.
+    private paddedSentenceHiliteRuleSelector =
+        " span.ui-audioCurrent > span.ui-enableHighlight";
     // note, no leading space here. We want (for example) .normal-style.ui-audioCurrent since this rule
     // targets textbox mode where both classes occur on the same element.
     private paraHiliteRuleSelector = ".ui-audioCurrent p";
@@ -536,6 +540,16 @@ export default class StyleEditor {
         );
         this.updateHiliteStyleRuleBody(
             sentenceRule,
+            hiliteTextColor,
+            hiliteBgColor
+        );
+        const paddedSentenceRule = this.GetRuleForStyle(
+            styleName,
+            this.paddedSentenceHiliteRuleSelector,
+            true
+        );
+        this.updateHiliteStyleRuleBody(
+            paddedSentenceRule,
             hiliteTextColor,
             hiliteBgColor
         );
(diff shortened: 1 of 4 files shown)


andrew-polk (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5905)* status: all files reviewed, 1 unresolved discussion (waiting on @StephenMcConnel)

___
*[`src/BloomExe/Book/HtmlDom.cs` line 3130 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5905#-NXWu3w50MW9Dvekbnel:-NXWu3w50MW9Dvekbnem:b-ipt13y) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/fc2dae6a863223618b594c225fe8c0c5f4ce3089/src/BloomExe/Book/HtmlDom.cs#L3130)):*
> ```Smalltalk
> 				return rawCSSCData;
> 			}
> 			return existingUserStyleNode.InnerXml;
> ```

I'm sure I'm just missing something, but I'm confused why this returns a value.
It seems the node must already be modified, so why does the caller set the node's innerXml to what seems to already be the innerXml?

I thought maybe it had something to do with making the method testable, but I think you can just get the innerXml from the node you passed in and test that, right?

StephenMcConnel (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5905)* status: 0 of 4 files reviewed, 1 unresolved discussion (waiting on @andrew-polk)

___
*[`src/BloomExe/Book/HtmlDom.cs` line 3130 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5905#-NXWu3w50MW9Dvekbnel:-NXlS7-OA29HBK9fuLpr:b-z305bo) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/fc2dae6a863223618b594c225fe8c0c5f4ce3089/src/BloomExe/Book/HtmlDom.cs#L3130)):*

I'm sure I'm just missing something, but I'm confused why this returns a value.
It seems the node must already be modified, so why does the caller set the node's innerXml to what seems to already be the innerXml?

I thought maybe it had something to do with making the method testable, but I think you can just get the innerXml from the node you passed in and test that, right?
</blockquote></details>

The node has not yet been modified, just what is supposed to go into it.
I suppose instead of return the value and modifying (or not) the content of the node, we could just modify the node in this method.
Which I've done, along with renaming some variables along the way.

andrew-polk (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5905)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @StephenMcConnel)

___
*[`src/BloomExe/Book/HtmlDom.cs` line 3130 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5905#-NXWu3w50MW9Dvekbnel:-NXlkZyn4ZAfSuvIyY6o:b-g6ccj7) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/fc2dae6a863223618b594c225fe8c0c5f4ce3089/src/BloomExe/Book/HtmlDom.cs#L3130)):*

The node has not yet been modified, just what is supposed to go into it.
I suppose instead of return the value and modifying (or not) the content of the node, we could just modify the node in this method.
Which I've done, along with renaming some variables along the way.
</blockquote></details>

Thanks; that's what I meant.
At least if I didn't communicate well, you and I were still on the same page. : )

PR #5922, opened 2023-06-12
Clean up extraneous wv2 setup call (BL-12257)

andrew-polk:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5922)

### DIFF
--- a/src/BloomExe/WebView2Browser.cs
+++ b/src/BloomExe/WebView2Browser.cs
@@ -236,32 +236,11 @@ namespace Bloom
 			// can leave implementing this until someone identifies a difference in Gecko vs WV2 behavior
 			// that we think is due to not implementing it.
 		}
-
-		private string _targetUrl;
-		private bool _ensuredCoreWebView2;
-		private bool _finishedUpdateDisplay;
-		private bool _urlMatches;
-
-		// For now I have decided not to make this return a Task and to consistently await it.
-		// The fan-out of methods that would have to be made async is daunting, and code
-		// that cares about the completion of the navigation will already have some
-		// mechanism in place for waiting not just until the call to Navigate after the
-		// await, but until we get an indication that the navigation is complete.
-		// Also, I suspect that EnsureCoreWebView2Async() will almost always be already completed
-		// and no awaiting will really be needed.
-		// Callers should nevertheless be aware that it is not absolutely guaranteed that
-		// Navigation has even started when this method returns.
-		protected override async void UpdateDisplay(string newUrl)
+		
+		protected override void UpdateDisplay(string newUrl)
 		{
-			_targetUrl = newUrl;
-			_ensuredCoreWebView2 = false;
-			_finishedUpdateDisplay = false;
-
-			await _webview.EnsureCoreWebView2Async();
-			_ensuredCoreWebView2 = true;
+			EnsureBrowserReadyToNavigate();
 			_webview.CoreWebView2.Navigate(newUrl);
-			_finishedUpdateDisplay = true;
-			_urlMatches = Url == newUrl;
 		}
 
 		protected override void EnsureBrowserReadyToNavigate()


andrew-polk (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5922)* status: all files reviewed, 1 unresolved discussion (waiting on @JohnThomson)

___
*[`src/BloomExe/WebView2Browser.cs` line 241 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5922#-NXp71Ry5V3NbGgkPKIj:-NXpVRN13303FxiqI7Qp:b-3xernh) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/c26936ca94bdfc03fd21844a61a248309dc2482e/src/BloomExe/WebView2Browser.cs#L241)):*
I struggled with this comment, too.

I think it is still as true as it was originally.
Before the change, the method would wait for `await _webview.EnsureCoreWebView2Async();` whereas now it waits for `EnsureBrowserReadyToNavigate();`. The synchronicity hasn't changed.

I left the comment in based on the doc for `Navigate` which states
> Note that this operation starts a navigation and the corresponding Microsoft.Web.WebView2.Core.CoreWebView2.NavigationStarting event is raised sometime after Navigate runs.

But it is dubious whether the caller cares.

So do you think it is better to have the comment? Or just remove it? Or append the comment from `Navigate()`?

andrew-polk:
*[`src/BloomExe/WebView2Browser.cs` line 241 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5922#-NXp71Ry5V3NbGgkPKIj:-NXuqHSJ5xzYQ0LYrhIq:b-andyzf) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/c26936ca94bdfc03fd21844a61a248309dc2482e/src/BloomExe/WebView2Browser.cs#L241)):*

No, you're not thinking through how await works. With the old code, the actual sequence of events is something like this (no doubt with some optimization)
1. This method is called.
2. the code before the await is executed
3. we start executing EnsureCoreWEbView2Async, and get as far as the first thing in it that is really async, that is, we kick off something that will happen in another thread and inform us later
4. EnsureCoreWEbView2Async returns a promise to this method, which  also returns it.
5. The caller carries on. It had better not assume that Navigate has been called, because it hasn't yet. For example, if it asks the browser "are you still navigating" the answer will be "no", because it hasn't even started.
6. Some time later, whatever we were waiting for in EnsureCoreWEbView2Async reports that it is finished
7. Later still, the UI thread processes the event associated with "EnsureCoreWEbView2Async is ready to continue"
8. The rest of the code in EnsureCoreWEbView2Async executes (well, there may be more cycles produced by more awaits)
9. EnsureCoreWEbView2Async signals that it has finished
10. That event gets processed
11. The rest of the code in UpdatePreview executes, including the call to Navigate.

With your change, 
1.  This method is called
2. EnsureReadyToNavigate is called
3. It sleeps, calls DoEvents, and so forth as necessary until it's actually OK to call Navigate
4. The rest of the method runs, including calling Navigate
5. We return to the caller (after calling Navigate). It's safe to count on Navigate having been called, even without using await.

Now, if the caller AWAITS the old version of UpdatePreview, then the rest of the calling method will in turn be postponed until the second half of UpdatePreview executes. But the warning was aimed at the case where it is NOT awaited, the caller just trusts things will get done eventually. So at step 5 of the old sequence, the caller is executing code after the call to UpdatePreview, but the expected call to Navigate doesn't happen until step 11. That's what I was trying to warn against, and it doesn't happen any more.

It may not have been a helpful (or sufficient) comment to begin with, but now, it is simply not true. When this method returns, Navigation is probably not complete, but at least the call that instructs the browser to navigate has happened.
</blockquote></details>

You're right that I wasn't thinking through it correctly.
And now I must humbly admit that I'm not sure I ever have, now that I understand it more fully. Sigh.
Thanks.

PR #5923, opened 2023-06-12
Clean up some LicenseDialog code (BL-12257)

andrew-polk:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5923)

### DIFF
--- a/src/BloomExe/Registration/LicenseDialog.cs
+++ b/src/BloomExe/Registration/LicenseDialog.cs
@@ -1,7 +1,5 @@
 ﻿using System;
 using System.Drawing;
-using System.Globalization;
-using System.IO;
 using System.Reflection;
 using System.Text;
 using System.Windows.Forms;
@@ -16,7 +14,7 @@ namespace Bloom.Registration
 		/// <summary>
 		/// Initialize a new instance of the <see cref="Bloom.Registration.LicenseDialog"/> class.
 		/// </summary>
-		/// <param name="licenseMdFile">filename of the license in HTML format</param>
+		/// <param name="licenseHtmlFile">filename of the license in HTML format</param>
 		/// <param name="prolog">prolog to the license (optional, already localized)</param>
 		/// <remarks>
 		/// The HTML file content should be suitable for inserting inside a body element.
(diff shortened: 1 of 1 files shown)


andrew-polk (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5923)* status: all files reviewed, 3 unresolved discussions (waiting on @JohnThomson)

___
*[`src/BloomExe/Registration/LicenseDialog.cs` line 61 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5923#-NXpVnh77stFf1TgvOwY:-NXpYd3y6FxS6Ufpax96:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/02595d7543362e6102e5b3a607e5ddf246793de4/src/BloomExe/Registration/LicenseDialog.cs#L61)):*

This is a new function, and the only call to it is commented out. Shouldn't we comment out the function, too? Especially since it isn't unit-tested, which makes me wonder how much confidence we have that it is working as intended.
</blockquote></details>

Done.
___
*[`src/BloomExe/Registration/LicenseDialog.cs` line 71 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5923#-NXpWXdQAnrxHE-RL0sb:-NXpbW0JEq7kUD8_Z3EB:b-2qimym) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/02595d7543362e6102e5b3a607e5ddf246793de4/src/BloomExe/Registration/LicenseDialog.cs#L71)):*

Could we give this a new name? I find it confusing, both reading code and especially debugging, if the values of non-ref arguments get changed.
Maybe language? languageOfLocale? languageCode?
</blockquote></details>

I just got rid of the named step.
___
*[`src/BloomExe/Registration/LicenseDialog.cs` line 82 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5923#-NXp8sSy-2BbicoycerL:-NXpbYHQDibYWaX8cZi6:b-8j2023) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/02595d7543362e6102e5b3a607e5ddf246793de4/src/BloomExe/Registration/LicenseDialog.cs#L82)):*

This appears to be assuming the the defaultLicenseFilePath always has a three-letter-plus-hyphen, which surprises me...isn't our default locale 'en'?
(I presume the default really is fixed...we'll never need to strip off a complex one like en-GB?)
</blockquote></details>

Actually, it is stripping off the extension.
I improved (and subsequently commented out... : P) the code.

JohnThomson (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5923)* status: all files reviewed, 1 unresolved discussion (waiting on @andrew-polk)

___
*[`src/BloomExe/Registration/LicenseDialog.cs` line 82 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5923#-NXp8sSy-2BbicoycerL:-NXuabEy371aOK_xca5c:b-etugxp) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/02595d7543362e6102e5b3a607e5ddf246793de4/src/BloomExe/Registration/LicenseDialog.cs#L82)):*

Actually, it is stripping off the extension.
I improved (and subsequently commented out... : P) the code.
</blockquote></details>

Sounds good. If we do start localizing, I wonder if the default will end up having a -en at the end, or being in an 'en' folder so we have a bit more to do...but guessing about it now is pointless.

PR #5927, opened 2023-06-13
Try to ensure that color is preserved in copy/paste in Bloom (BL-12357)

StephenMcConnel:
This change targets Version5.4 as requested in the issue.  I tried to ensure that it's safe enough.

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5927)

### DIFF
--- a/src/BloomBrowserUI/bookEdit/bloomField/BloomField.ts
+++ b/src/BloomBrowserUI/bookEdit/bloomField/BloomField.ts
@@ -163,6 +163,9 @@ export default class BloomField {
         //     alert(event.data.dataValue);
         // });
         ckeditor.on("paste", event => {
+            event.data.dataValue = this.restoreHtmlMarkupIfNecessary(
+                event.data
+            );
             event.data.dataValue = this.reconstituteParagraphsOnPlainTextPaste(
                 event.data
             );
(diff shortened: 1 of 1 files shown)


andrew-polk (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5927)* status: all files reviewed (commit messages unreviewed), 1 unresolved discussion (waiting on @StephenMcConnel)

___
*[`src/BloomBrowserUI/bookEdit/bloomField/BloomField.ts` line 304 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5927#-NXr5F-n8lemwTSyh_BD:-NXr5F-n8lemwTSyh_BE:bbtec36) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/fa806766d6a47c7e28e0489755a3de836153d683/src/BloomBrowserUI/bookEdit/bloomField/BloomField.ts#L304)):*
> ```XML
> 
>     // Spans can be dropped from the clipboard's default data (eventData.dataValue),
>     // but we need to keep them for possible color markup.  See BL-12357.
> ```

How does this interact with ckeditor's pasteFilter (in config.js)?

I see this comment there:
`BL-4775: Removed span from this, so that you can't paste spans`

StephenMcConnel (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5927)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @andrew-polk)

___
*[`src/BloomBrowserUI/bookEdit/bloomField/BloomField.ts` line 304 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5927#-NXr5F-n8lemwTSyh_BD:-NXueqy97F31T_8SPCf4:b-ggm4qf) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/fa806766d6a47c7e28e0489755a3de836153d683/src/BloomBrowserUI/bookEdit/bloomField/BloomField.ts#L304)):*

How does this interact with ckeditor's pasteFilter (in config.js)?

I see this comment there:
`BL-4775: Removed span from this, so that you can't paste spans`
</blockquote></details>

ah, that's a good question.  The pasteFilter applies only to external sources, like Word or Notepad.  It needs to be updated because my fix only applies to the internal CKEditor sources inside Bloom.
After further study and testing, I've refined the fix, although the basic shape of it remains the same.

andrew-polk (changes requested):
> Reviewed 1 of 2 files at r2.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5927)* status: 1 of 2 files reviewed, 2 unresolved discussions (waiting on @StephenMcConnel)

___
*[`src/BloomBrowserUI/lib/ckeditor/config.js` line 105 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5927#-NXw7BaI71qafBTTyHGG:-NXw7BaI71qafBTTyHGH:btyntf7) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/46ac11556c0dde15ea5bdf024d8d5d8e9761def3/src/BloomBrowserUI/lib/ckeditor/config.js#L105)):*
> ```JavaScript
>     // Code in our paste handler deals with this so it doesn't end up in our documents.
>     config.pasteFilter =
>         "p br em i strong sup u; b{font-weight}; a[!href]; span{color}";
> ```

This gives me pause.
If spans, generally, were causing bugs with talking book tool (a la BL-4775), it seems likely that this still opens that door.
For example, what if Word or another program, simply adds `color` to each span?

andrew-polk (changes requested):
> Reviewed 1 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5927)* status: all files reviewed, 1 unresolved discussion (waiting on @StephenMcConnel)

___
*[`src/BloomBrowserUI/lib/ckeditor/config.js` line 105 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5927#-NXw7BaI71qafBTTyHGG:-NY-9dQx6ffGNTun8zmF:blc2tum) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/46ac11556c0dde15ea5bdf024d8d5d8e9761def3/src/BloomBrowserUI/lib/ckeditor/config.js#L105)):*

That won't cause problems with the talking book tool.  The only attribute allowed through on the spans coming from outside is "style", and the only content allowed in the style attribute is "color".  The id and class attributes, if they exist, are omitted by the filter.  If the span doesn't have a color style, then the contents are allowed through, but not the span element/markup.  Once it's in, it's no different than the span created by Bloom itself to mark colored text.
</blockquote></details>

We discussed this and decided, that at least for 5.4, this is too risky.

andrew-polk (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5927)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @StephenMcConnel)

PR #5930, opened 2023-06-15
BL-12339 round-trip color in spreadsheet

nabalone:
Support color import from and export to spreadsheet

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5930)

### DIFF
--- a/src/BloomExe/Spreadsheet/MarkedUpText.cs
+++ b/src/BloomExe/Spreadsheet/MarkedUpText.cs
@@ -1,6 +1,7 @@
 using System;
 using System.Collections;
 using System.Collections.Generic;
+using System.Drawing;
 using System.Linq;
 using System.Text;
 using System.Xml;
@@ -174,7 +175,8 @@ namespace Bloom.Spreadsheet
 				foreach (XmlNode child in node.ChildNodes)
 				{
 					MarkedUpText markedUpChild = ParseXmlRecursive(child);
-					ApplyFormatting(node.Name, markedUpChild);
+					XmlElement el = (XmlElement)node;
+					ApplyFormatting(node.Name, el.GetAttribute("style"), markedUpChild);
 					markedUpText._runList.AddRange(markedUpChild._runList);
 				}
 			}
@@ -182,11 +184,11 @@ namespace Bloom.Spreadsheet
 
 		}
 
-		private static void ApplyFormatting(string formatName, MarkedUpText markedUpText)
+		private static void ApplyFormatting(string formatName, string styleAttribute, MarkedUpText markedUpText)
 		{
 			foreach (MarkedUpTextRun run in markedUpText._runList)
 			{
-				run.setProperty(formatName);
+				run.setProperty(formatName, styleAttribute);
 			}
 		}
 	}
@@ -209,12 +211,13 @@ namespace Bloom.Spreadsheet
 		public bool Italic { get; set; }
 		public bool Underlined { get; set; }
 		public bool Superscript { get; set; }
-		public bool HasFormatting => Bold | Italic | Underlined | Superscript;
+		public Color Color { get; set; }
 
+		public bool HasFormatting => Bold | Italic | Underlined | Superscript | Color != null;
 
-		public void setProperty(string propertyName)
+		public void setProperty(string propertyName, string styleAttribute)
 		{
-			if (propertyName.Equals("strong") || propertyName.Equals("b")) 
+			if (propertyName.Equals("strong") || propertyName.Equals("b"))
 			{
 				Bold = true;
 			}
@@ -229,6 +232,24 @@ namespace Bloom.Spreadsheet
 			else if (propertyName.Equals("u"))
 			{
 				Underlined = true;
+			}
+			else
+			{
+				// remove spaces from the style attribute
+				styleAttribute = styleAttribute.Replace(" ", "");
+				int colorFoundIndex = styleAttribute.IndexOf("color:");
+				if (colorFoundIndex > -1) {
+					// There is a color specified
+					int colorCodeStartIndex = colorFoundIndex + 6;
+					// color code end index is the next semicolon after the start, or the end of the style string
+					int colorCodeEndIndex = styleAttribute.IndexOf(';', colorCodeStartIndex);
+					if (colorCodeEndIndex == -1)
+					{
+						colorCodeEndIndex = styleAttribute.Length;
+					}
+					string colorString = styleAttribute.Substring(colorCodeStartIndex, colorCodeEndIndex - colorCodeStartIndex);
+					Color = ColorTranslator.FromHtml(colorString);
+				}
 			}
 		}
 
--- a/src/BloomExe/Spreadsheet/MarkedUpText.cs
+++ b/src/BloomExe/Spreadsheet/MarkedUpText.cs
@@ -5,6 +5,7 @@ using System.Drawing;
 using System.Linq;
 using System.Text;
 using System.Xml;
+using SIL.Xml;
 
 namespace Bloom.Spreadsheet
 {
@@ -175,8 +176,7 @@ namespace Bloom.Spreadsheet
 				foreach (XmlNode child in node.ChildNodes)
 				{
 					MarkedUpText markedUpChild = ParseXmlRecursive(child);
-					XmlElement el = (XmlElement)node;
-					ApplyFormatting(node.Name, el.GetAttribute("style"), markedUpChild);
+					ApplyFormatting(node.Name, node.GetOptionalStringAttribute("style", ""), markedUpChild);
 					markedUpText._runList.AddRange(markedUpChild._runList);
 				}
 			}
@@ -237,7 +237,7 @@ namespace Bloom.Spreadsheet
 			{
 				// remove spaces from the style attribute
 				styleAttribute = styleAttribute.Replace(" ", "");
-				int colorFoundIndex = styleAttribute.IndexOf("color:");
+				int colorFoundIndex = styleAttribute.IndexOf("color:#");
 				if (colorFoundIndex > -1) {
 					// There is a color specified
 					int colorCodeStartIndex = colorFoundIndex + 6;
@@ -249,6 +249,15 @@ namespace Bloom.Spreadsheet
 					}
 					string colorString = styleAttribute.Substring(colorCodeStartIndex, colorCodeEndIndex - colorCodeStartIndex);
 					Color = ColorTranslator.FromHtml(colorString);
+				}
+				else if (styleAttribute.IndexOf("color:rgb(") > -1)
+				{
+					//extract r, g, b values
+					int startIndex = styleAttribute.IndexOf("color:rgb(") + 10;
+					int endIndex = styleAttribute.IndexOf(")", startIndex);
+					string rgbString = styleAttribute.Substring(startIndex, endIndex - startIndex);
+					string[] rgbValues = rgbString.Split(',');
+					Color = Color.FromArgb(int.Parse(rgbValues[0]), int.Parse(rgbValues[1]), int.Parse(rgbValues[2]));
 				}
 			}
 		}
--- a/src/BloomExe/Spreadsheet/MarkedUpText.cs
+++ b/src/BloomExe/Spreadsheet/MarkedUpText.cs
@@ -250,15 +250,6 @@ namespace Bloom.Spreadsheet
 					string colorString = styleAttribute.Substring(colorCodeStartIndex, colorCodeEndIndex - colorCodeStartIndex);
 					Color = ColorTranslator.FromHtml(colorString);
 				}
-				else if (styleAttribute.IndexOf("color:rgb(") > -1)
-				{
-					//extract r, g, b values
-					int startIndex = styleAttribute.IndexOf("color:rgb(") + 10;
-					int endIndex = styleAttribute.IndexOf(")", startIndex);
-					string rgbString = styleAttribute.Substring(startIndex, endIndex - startIndex);
-					string[] rgbValues = rgbString.Split(',');
-					Color = Color.FromArgb(int.Parse(rgbValues[0]), int.Parse(rgbValues[1]), int.Parse(rgbValues[2]));
-				}
 			}
 		}
 
--- a/src/BloomExe/Spreadsheet/SpreadsheetIO.cs
+++ b/src/BloomExe/Spreadsheet/SpreadsheetIO.cs
@@ -462,7 +462,6 @@ namespace Bloom.Spreadsheet
 		private static bool IsBlack(Color color)
 		{
 			return color.R == 0 && color.B == 0 && color.G == 0;
-			//return ColorTranslator.ToHtml(color) == "Black" || ColorTranslator.ToHtml(color) == "#000000";
 		}
 
 		// Excel formatting can be at the entire cell level (e.g. the entire cell is marked italic)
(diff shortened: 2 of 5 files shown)


JohnThomson (changes requested):
I think there should be unit tests for at least some of this. For example, I think we have tests that prove things like Bold and Italic round trip through an actual file; we should test for color, too. And do any of the methods you enhanced have their own unit tests?

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5930#-:-NY-NtvWEKYsHUPQcdsk:b-yf7hzf)* status: all files reviewed, 2 unresolved discussions (waiting on @nabalone)

___
*[`src/BloomExe/Spreadsheet/SpreadsheetIO.cs` line 97 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5930#-NY-Cvi4CLFben7dCczi:-NY-Cvi4CLFben7dCczj:ball0is) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/f079627899ca1c48a84bcfb4de98d4c9a4c711c5/src/BloomExe/Spreadsheet/SpreadsheetIO.cs#L97)):*
> ```Smalltalk
> 								foreach (MarkedUpTextRun run in markedUpText.Runs)
> 								{
> 									// Maybe in the future we could skip over all whitespace runs, not just empty ones?
> ```

I don't think so. (a) A bold space just might actually be wider than a plain one, and a superscript space might be narrower (b) An underlined space might actually get underlined? Don't remember for sure (c) one day we may support background color, and it will make a major difference for white space.
___
*[`src/BloomExe/Spreadsheet/SpreadsheetIO.cs` line 466 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5930#-NY-EyhSAW_pwnEZy3YB:-NY-EyhSAW_pwnEZy3YC:b-1c1yok) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/f079627899ca1c48a84bcfb4de98d4c9a4c711c5/src/BloomExe/Spreadsheet/SpreadsheetIO.cs#L466)):*
> ```Smalltalk
> 		{
> 			return color.R == 0 && color.B == 0 && color.G == 0;
> 			//return ColorTranslator.ToHtml(color) == "Black" || ColorTranslator.ToHtml(color) == "#000000";
> ```

Does Excel (or Bloom) support transparency in text color? If so, what answer would we want for text that is 50% transparency black?

nabalone (commented):
I made some other changes based on code I saw elsewhere in the database. Please review my latest commit

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5930#-:-NY0GkLt-D5dWP4j9yf1:b8qko70)* status: 0 of 2 files reviewed, 4 unresolved discussions (waiting on @JohnThomson)

___
*[`src/BloomExe/Spreadsheet/MarkedUpText.cs` line 239 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5930#-NY0GZ1sBdmBotyIB9gE:-NY0GZ1t4WxsuU1Cq8_W:bmk8sgs) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/c3901b001929e88bc8945be87b7d8033d2b02927/src/BloomExe/Spreadsheet/MarkedUpText.cs#L239)):*
> ```Smalltalk
> 			{
> 				// remove spaces from the style attribute
> 				styleAttribute = styleAttribute.Replace(" ", "");
> ```

do we need to worry about any other whitespace characters being in the style string?
___
*[`src/BloomExe/Spreadsheet/SpreadsheetIO.cs` line 97 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5930#-NY-Cvi4CLFben7dCczi:-NY01WWOEGgiTAUaQkmi:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/f079627899ca1c48a84bcfb4de98d4c9a4c711c5/src/BloomExe/Spreadsheet/SpreadsheetIO.cs#L97)):*

I don't think so. (a) A bold space just might actually be wider than a plain one, and a superscript space might be narrower (b) An underlined space might actually get underlined? Don't remember for sure (c) one day we may support background color, and it will make a major difference for white space.
</blockquote></details>

Done.
___
*[`src/BloomExe/Spreadsheet/SpreadsheetIO.cs` line 466 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5930#-NY-EyhSAW_pwnEZy3YB:-NY01XKB4_rMoLDoEojA:buzsx3y) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/f079627899ca1c48a84bcfb4de98d4c9a4c711c5/src/BloomExe/Spreadsheet/SpreadsheetIO.cs#L466)):*

Does Excel (or Bloom) support transparency in text color? If so, what answer would we want for text that is 50% transparency black?
</blockquote></details>

I don't see any way to make adjust text transparency in either Bloom or Excel; both use straightforward hex RGB color choosers. If there is transparency denoted, I'm guessing a user did something hacky and we wouldn't need to support it. 

I thought about whether we could use the Alpha channel to keep track of whether text was black just by default or explicitly marked black by the user, but I think it is unlikely this would work and if it did it would probably lead to more confusing unexpected behavior
___
*[`src/BloomExe/Spreadsheet/SpreadsheetIO.cs` line 466 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5930#-NY09E6l6PDOai1R84Rw:-NY09E6l6PDOai1R84Rx:b-8sn2sy) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/f079627899ca1c48a84bcfb4de98d4c9a4c711c5/src/BloomExe/Spreadsheet/SpreadsheetIO.cs#L466)):*
> ```Smalltalk
> 		{
> 			return color.R == 0 && color.B == 0 && color.G == 0;
> 			//return ColorTranslator.ToHtml(color) == "Black" || ColorTranslator.ToHtml(color) == "#000000";
> ```

Which of these two methods of checking for Black is better?

JohnThomson (changes requested):
I made a few more comments. Still don't see any testing going on.

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5930#-:-NYPQaTdEYG3h2nATIWh:b-8cvxez)* status: all files reviewed, 2 unresolved discussions (waiting on @nabalone)

___
*[`src/BloomExe/Spreadsheet/MarkedUpText.cs` line 239 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5930#-NY0GZ1sBdmBotyIB9gE:-NYPNQrZDu3WmbcEix2y:bd6csnn) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/c3901b001929e88bc8945be87b7d8033d2b02927/src/BloomExe/Spreadsheet/MarkedUpText.cs#L239)):*

do we need to worry about any other whitespace characters being in the style string?
</blockquote></details>

I don't think so. I'm not even sure whether HTML allows newlines in attribute values, but even if it does, these style attributes are typically inserted by code, and we're not going to code it to do that.
___
*[`src/BloomExe/Spreadsheet/MarkedUpText.cs` line 260 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5930#-NYPP7b1ELsHfdLvqkez:-NYPP7b1ELsHfdLvqkf-:b-6upcip) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/c3901b001929e88bc8945be87b7d8033d2b02927/src/BloomExe/Spreadsheet/MarkedUpText.cs#L260)):*
> ```Smalltalk
> 					string rgbString = styleAttribute.Substring(startIndex, endIndex - startIndex);
> 					string[] rgbValues = rgbString.Split(',');
> 					Color = Color.FromArgb(int.Parse(rgbValues[0]), int.Parse(rgbValues[1]), int.Parse(rgbValues[2]));
> ```

We should probably use TryParse() here. I'm not sure quite what to do if we get a string that doesn't parse as an integer, and it's unlikely, but we generally try to give some sort of intelligible problem report rather than just crashing if a file is badly formatted. Or you could wrap the line in a try...catch, since not being able to parse the number is definitely unexpected here.

Since we expect the value of this attribute to come from Bloom code (I think), I'm surprised we need to handle both #colors and rgb ones. If we really do use both (but not any of the other options, like names or hsl or rgba) that's probably worth a comment.
___
*[`src/BloomExe/Spreadsheet/SpreadsheetIO.cs` line 466 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5930#-NY09E6l6PDOai1R84Rw:-NYPO0BR8JIu8kl_7Q-u:b38sxre) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/f079627899ca1c48a84bcfb4de98d4c9a4c711c5/src/BloomExe/Spreadsheet/SpreadsheetIO.cs#L466)):*

Which of these two methods of checking for Black is better?
</blockquote></details>

I prefer the one you have, especially if the result of ToHtml is doubtful enough to require testing two options. For one thing, we don't even have to raise the question of whether "Black" could get localized.

nabalone (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5930)* status: 0 of 5 files reviewed, 3 unresolved discussions (waiting on @JohnThomson)

___
*[`src/BloomExe/Spreadsheet/MarkedUpText.cs` line 260 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5930#-NYPP7b1ELsHfdLvqkez:-NYdsY3l3b_vwoRQvxPu:b-2ni3j8) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/c3901b001929e88bc8945be87b7d8033d2b02927/src/BloomExe/Spreadsheet/MarkedUpText.cs#L260)):*

We should probably use TryParse() here. I'm not sure quite what to do if we get a string that doesn't parse as an integer, and it's unlikely, but we generally try to give some sort of intelligible problem report rather than just crashing if a file is badly formatted. Or you could wrap the line in a try...catch, since not being able to parse the number is definitely unexpected here.

Since we expect the value of this attribute to come from Bloom code (I think), I'm surprised we need to handle both #colors and rgb ones. If we really do use both (but not any of the other options, like names or hsl or rgba) that's probably worth a comment.
</blockquote></details>

I don't think rgb() format colors is necessary at this time. I had put it there because it looks like we use rgb color elsewhere, so I thought that conceivably in the future we might use rgb colors for main text coloration also. But now I've taken it out since it seems unnecessarily confusing
___
*[`src/BloomExe/Spreadsheet/SpreadsheetIO.cs` line 466 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5930#-NY09E6l6PDOai1R84Rw:-NYdsWIn29ROWOqdL4-3:b32w1t) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/f079627899ca1c48a84bcfb4de98d4c9a4c711c5/src/BloomExe/Spreadsheet/SpreadsheetIO.cs#L466)):*

I prefer the one you have, especially if the result of ToHtml is doubtful enough to require testing two options. For one thing, we don't even have to raise the question of whether "Black" could get localized.
</blockquote></details>

Ok
___
*[`src/BloomTests/Spreadsheet/SpreadsheetTests.cs` line 487 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5930#-NYds4yf5qRUxv1B9MhV:-NYds4yf5qRUxv1B9MhW:bipifdn) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/a09c521dc768fd504a180e48d1562a40a53cc1b6/src/BloomTests/Spreadsheet/SpreadsheetTests.cs#L487)):*
> ```Smalltalk
> 
> 		[Test]
> 		public void RichtextNotUsedWhenNotNeeded()
> ```

We now use RichText even if IsRichText=false so we can extract color; non-black text won't necessarily have IsRichText=True

nabalone (commented):
Added

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5930#-:-NYdtHs_62Z1pjgAHBNR:b-8b11j9)* status: 0 of 5 files reviewed, 3 unresolved discussions (waiting on @JohnThomson)

JohnThomson (changes requested):
Nice. Just one more question now.

> Reviewed 5 of 5 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5930#-:-NYrxcsyDIOMD0mg_32U:b-xsn0rj)* status: all files reviewed, 1 unresolved discussion (waiting on @nabalone)

___
*[`src/BloomExe/Spreadsheet/SpreadsheetIO.cs` line 406 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5930#-NYrxC5-0w6vNAPpQKjk:-NYrxC5-0w6vNAPpQKjl:bqrbctn) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/3352a79bc4e9a630cdeae264b02647fcc3d9a87f/src/BloomExe/Spreadsheet/SpreadsheetIO.cs#L406)):*
> ```Smalltalk
> 			var whitespaceSplitters = new string[] { "\n", "\r\n" };
> 
> 			var content = cell.RichText;
> ```

Are you satisfied that this works even when the cell is not rich text? Especially a spreadsheet created by hand rather than exported from Bloom. Might deserve a comment.
___
*[`src/BloomTests/Spreadsheet/SpreadsheetTests.cs` line 487 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5930#-NYds4yf5qRUxv1B9MhV:-NYrwVX40h0guQi4-Gap:b-jwya11) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/a09c521dc768fd504a180e48d1562a40a53cc1b6/src/BloomTests/Spreadsheet/SpreadsheetTests.cs#L487)):*

We now use RichText even if IsRichText=false so we can extract color; non-black text won't necessarily have IsRichText=True
</blockquote></details>

What I'm seeing is that we process it as we used to when IsRichText answered true; but there is still conditional logic around MAKING it rich text on export. And of course an end user might not do whatever it would take to make a cell rich text. See the comment there.

nabalone (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5930)* status: 4 of 5 files reviewed, 1 unresolved discussion (waiting on @JohnThomson)

___
*[`src/BloomExe/Spreadsheet/SpreadsheetIO.cs` line 406 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5930#-NYrxC5-0w6vNAPpQKjk:-NYxVs1KDy-BsxbsryGv:b-l4396t) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/3352a79bc4e9a630cdeae264b02647fcc3d9a87f/src/BloomExe/Spreadsheet/SpreadsheetIO.cs#L406)):*

Are you satisfied that this works even when the cell is not rich text? Especially a spreadsheet created by hand rather than exported from Bloom. Might deserve a comment.
</blockquote></details>

Added a comment
___
*[`src/BloomTests/Spreadsheet/SpreadsheetTests.cs` line 487 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5930#-NYds4yf5qRUxv1B9MhV:-NYxVuHy-ESJLtVFsbJJ:bovnhdu) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/a09c521dc768fd504a180e48d1562a40a53cc1b6/src/BloomTests/Spreadsheet/SpreadsheetTests.cs#L487)):*

What I'm seeing is that we process it as we used to when IsRichText answered true; but there is still conditional logic around MAKING it rich text on export. And of course an end user might not do whatever it would take to make a cell rich text. See the comment there.
</blockquote></details>

Added a comment in SpreadsheetIO.cs

PR #5942, opened 2023-06-19
Handle pages without text in Toolbox.doWhenPageReady (BL-12381)

StephenMcConnel:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5942)

### DIFF
--- a/src/BloomBrowserUI/bookEdit/js/bloomEditing.ts
+++ b/src/BloomBrowserUI/bookEdit/js/bloomEditing.ts
@@ -1077,6 +1077,22 @@ function isTextSelected(): boolean {
 }
 
 let reportedTextSelected = isTextSelected();
+
+// We want to attach ckeditor to the contenteditable="true" class="bloom-content1"
+// also to contenteditable="true" and class="bloom-content2" or class="bloom-content3"
+// as well as Equation-style (Used by ArithmeticTemplate.pug, these are language neutral and don't have a content language)
+// attachToCkEditor will skip any element with class="bloom-userCannotModifyStyles" (which might be on the translationGroup)
+// Update 1 Feb 2022: JohnT: in dealing with BL-10893 it became clear that ckEditor should also be attached to
+// contenteditable things with class bloom-contentNational2, as it had previously been made to attach to bloom-contentNational1,
+// though the comment was not updated. It is NOT clear to me exactly what principle is meant to really govern what elements
+// get ckeditor attached (which, at least, allows them to have styled text). There is presumably some reason not to
+// just attach it to everything that is contenteditable or everything with bloom-visibility-code-on, but unfortunately
+// we did not comment that. Possibly there are special fields (ISBN comes to mind) in xmatter that should not be styled?
+export const ckeditableSelector =
+    ".bloom-content1[contenteditable='true'],.bloom-content2[contenteditable='true']," +
+    ".bloom-content3[contenteditable='true'],.bloom-contentNational1[contenteditable='true']," +
+    ".bloom-contentNational2[contenteditable='true'],.Equation-style[contenteditable='true']";
+
 // ---------------------------------------------------------------------------------
 // called inside document ready function
 // ---------------------------------------------------------------------------------
@@ -1132,23 +1148,10 @@ export function bootstrap() {
         return;
     }
 
-    // attach ckeditor to the contenteditable="true" class="bloom-content1"
-    // also to contenteditable="true" and class="bloom-content2" or class="bloom-content3"
-    // as well as Equation-style (Used by ArithmeticTemplate.pug, these are language neutral and don't have a content language)
-    // attachToCkEditor will skip any element with class="bloom-userCannotModifyStyles" (which might be on the translationGroup)
-    // Update 1 Feb 2022: JohnT: in dealing with BL-10893 it became clear that ckEditor should also be attached to
-    // contenteditable things with class bloom-contentNational2, as it had previously been made to attach to bloom-contentNational1,
-    // though the comment was not updated. It is NOT clear to me exactly what principle is meant to really govern what elements
-    // get ckeditor attached (which, at least, allows them to have styled text). There is presumably some reason not to
-    // just attach it to everything that is contenteditable or everything with bloom-visibility-code-on, but unfortunately
-    // we did not comment that. Possibly there are special fields (ISBN comes to mind) in xmatter that should not be styled?
-    const complicatedFind =
-        ".bloom-content1[contenteditable='true'],.bloom-content2[contenteditable='true']," +
-        ".bloom-content3[contenteditable='true'],.bloom-contentNational1[contenteditable='true']," +
-        ".bloom-contentNational2[contenteditable='true'],.Equation-style[contenteditable='true']";
-
+    // Attach ckeditor to the fields that can have styled editable text.
+    // (See comment above on ckeditableSelector for what fields those are.)
     $("div.bloom-page")
-        .find(complicatedFind)
+        .find(ckeditableSelector)
         .each((index: number, element: Element) => {
             attachToCkEditor(element);
         });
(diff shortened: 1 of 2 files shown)


JohnThomson (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5942)* status: all files reviewed, 2 unresolved discussions (waiting on @StephenMcConnel)

___
*[`src/BloomBrowserUI/bookEdit/toolbox/toolbox.ts` line 553 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5942#-NYP0Gq07MeEkLj3dvff:-NYP0Gq07MeEkLj3dvfg:b-ea72s5) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/27610964c8bc092491fc80b851658cb8de73f494/src/BloomBrowserUI/bookEdit/toolbox/toolbox.ts#L553)):*
> ```XML
>         if (!gotOne) {
>             const page = ToolBox.getPage();
>             if (!page || page.querySelector("div.bloom-editable")) {
> ```

This will definitely help. However, while we're finally fixing this, I wonder whether we should use the "complicatedFind" (see bloomEditing.ts) that is used to find the set of divs that we will actually try to make ckeditors for? Or some other strategy so that we access the results of that find and know whether the page is trying to set up any ckEditors? (If you do make "complicatedFind" public please come up with a better name...something like editableBlockSelector.)
___
*[`src/BloomBrowserUI/bookEdit/toolbox/toolbox.ts` line 557 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5942#-NYP0junBVCeyKddqBrB:-NYP0junBVCeyKddqBrC:b2pov82) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/27610964c8bc092491fc80b851658cb8de73f494/src/BloomBrowserUI/bookEdit/toolbox/toolbox.ts#L557)):*
> ```XML
>                 // page gets set up with ckeditor. I'm pretty sure that page will always be truthy
>                 // at this point, but I'm being paranoid about the possibility that I might be wrong.
>                 // See BL-12381. (If page is not truthy here, the bug has no fix as far as I can tell.)
> ```

Could try a timeout if page is falsy? The "when ckEditor is ready" action is typically a manipulation of something in page, so it's probably harmless to not do it until we can find a page to manipulate. That branch wouldn't need to put anything in removers or make any new event handlers, just call this again after a bit and see whether things eventually get loaded enough that we have a page document.

StephenMcConnel (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5942)* status: all files reviewed, 2 unresolved discussions (waiting on @JohnThomson)

___
*[`src/BloomBrowserUI/bookEdit/toolbox/toolbox.ts` line 553 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5942#-NYP0Gq07MeEkLj3dvff:-NYPYqjt47NeY4X0PwTL:b-6a88da) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/27610964c8bc092491fc80b851658cb8de73f494/src/BloomBrowserUI/bookEdit/toolbox/toolbox.ts#L553)):*

This will definitely help. However, while we're finally fixing this, I wonder whether we should use the "complicatedFind" (see bloomEditing.ts) that is used to find the set of divs that we will actually try to make ckeditors for? Or some other strategy so that we access the results of that find and know whether the page is trying to set up any ckEditors? (If you do make "complicatedFind" public please come up with a better name...something like editableBlockSelector.)
</blockquote></details>

How about "ckeditableSelector"?
___
*[`src/BloomBrowserUI/bookEdit/toolbox/toolbox.ts` line 557 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5942#-NYP0junBVCeyKddqBrB:-NYPRdMsEpmsWQoKWZj-:b-85nl6d) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/27610964c8bc092491fc80b851658cb8de73f494/src/BloomBrowserUI/bookEdit/toolbox/toolbox.ts#L557)):*

Could try a timeout if page is falsy? The "when ckEditor is ready" action is typically a manipulation of something in page, so it's probably harmless to not do it until we can find a page to manipulate. That branch wouldn't need to put anything in removers or make any new event handlers, just call this again after a bit and see whether things eventually get loaded enough that we have a page document.
</blockquote></details>

There's a timeout for !page earlier in the call sequence.   I decided that I could just pass in `page` as a method argument instead of calling `Toolbox.getPage()` again.  This has the virtue of emphasizing that `page` has to have a value at this point in the code as well as avoiding unnecessary computations.

JohnThomson (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5942)* status: all files reviewed, 2 unresolved discussions (waiting on @StephenMcConnel)

___
*[`src/BloomBrowserUI/bookEdit/toolbox/toolbox.ts` line 553 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5942#-NYP0Gq07MeEkLj3dvff:-NYTWKox5A4TQrd38AQn:b-c5ywn9) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/27610964c8bc092491fc80b851658cb8de73f494/src/BloomBrowserUI/bookEdit/toolbox/toolbox.ts#L553)):*

How about "ckeditableSelector"?
</blockquote></details>

That would work (but I don't see that you've done it yet).
___
*[`src/BloomBrowserUI/bookEdit/toolbox/toolbox.ts` line 557 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5942#-NYP0junBVCeyKddqBrB:-NYTVT4i5EKTPE26Sf2l:b-i97qi7) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/27610964c8bc092491fc80b851658cb8de73f494/src/BloomBrowserUI/bookEdit/toolbox/toolbox.ts#L557)):*

There's a timeout for !page earlier in the call sequence.   I decided that I could just pass in `page` as a method argument instead of calling `Toolbox.getPage()` again.  This has the virtue of emphasizing that `page` has to have a value at this point in the code as well as avoiding unnecessary computations.
</blockquote></details>

Fair enough. But perhaps the comment is a bit alarmist? Would it be fair to say, "The action we want to do here is typically a manipulation of the page, so if we can't get a page even after the timeout above, it's not going to matter that we don't do the action."

StephenMcConnel (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5942)* status: 0 of 2 files reviewed, 2 unresolved discussions (waiting on @JohnThomson)

___
*[`src/BloomBrowserUI/bookEdit/toolbox/toolbox.ts` line 553 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5942#-NYP0Gq07MeEkLj3dvff:-NYTXpuh7NDu9VNTR3S2:bm7si7m) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/27610964c8bc092491fc80b851658cb8de73f494/src/BloomBrowserUI/bookEdit/toolbox/toolbox.ts#L553)):*

That would work (but I don't see that you've done it yet).
</blockquote></details>

I don't know why you aren't seeing my change.  I'm sure I had pushed it.  But to make sure, I just rebased the branch on Version5.5 and repushed it just a moment ago..
___
*[`src/BloomBrowserUI/bookEdit/toolbox/toolbox.ts` line 557 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5942#-NYP0junBVCeyKddqBrB:-NYTYB5xA5AgniUFacpO:bwexh16) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/27610964c8bc092491fc80b851658cb8de73f494/src/BloomBrowserUI/bookEdit/toolbox/toolbox.ts#L557)):*

Fair enough. But perhaps the comment is a bit alarmist? Would it be fair to say, "The action we want to do here is typically a manipulation of the page, so if we can't get a page even after the timeout above, it's not going to matter that we don't do the action."
</blockquote></details>

I edited the comment down since the code change ensures that page is set.

JohnThomson (approved):
Nice. Thanks.

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5942#-:-NYTr0qm74jOSo111Ab_:b6eg241)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @StephenMcConnel)

PR #5950, opened 2023-06-21
BL-12379 disable unhelpful react-avatar caching and fix offline avatar default initials

nabalone:
By default react-avatar caches which gravatar urls failed (whether because the user was offline or did not have a gravatar) in local storage and then doesn’t retry the cached failed urls for a week. This PR turns off this undesireable failed url caching. It also fixes the default avatars when the user is offline.

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5950)

### DIFF
--- a/src/BloomBrowserUI/react_components/bloomAvatar.tsx
+++ b/src/BloomBrowserUI/react_components/bloomAvatar.tsx
@@ -2,7 +2,7 @@
 import { jsx, css } from "@emotion/react";
 
 import React = require("react");
-import Avatar from "react-avatar";
+import Avatar, { Cache, ConfigProvider } from "react-avatar";
 import { getMd5 } from "../bookEdit/toolbox/talkingBook/md5Util";
 
 export const BloomAvatar: React.FunctionComponent<{
@@ -19,6 +19,15 @@ export const BloomAvatar: React.FunctionComponent<{
     const borderStyle = props.borderColor
         ? `${borderSizeInt}px solid ${props.borderColor}`
         : undefined;
+
+    // react-avatar does not cache actual avatars. It only caches which gravatar urls failed
+    // (whether because user was offline or doesn't have a gravatar),
+    // and then doesn't retry the failed urls so long as they are valid in cache.
+    // We do want it to retry retrieving gravatars, so keep its cache empty
+    const cache = new Cache({
+        sourceTTL: 0, // retain for 0 milliseconds
+        sourceSize: 0 // retain a maximum of 0 items in cache
+    });
     return (
         <React.Suspense fallback={<React.Fragment />}>
             <div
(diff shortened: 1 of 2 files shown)


src/BloomBrowserUI/react_components/bloomAvatar.tsx
```diff
@@ -19,6 +19,15 @@ export const BloomAvatar: React.FunctionComponent<{
     const borderStyle = props.borderColor
         ? `${borderSizeInt}px solid ${props.borderColor}`
         : undefined;
+
+    // react-avatar does not cache actual avatars. It only caches which gravatar urls failed
+    // (whether because user was offline or doesn't have a gravatar),
+    // and then doesn't retry the failed urls so long as they are valid in cache.
+    // We do want it to retry retrieving gravatars, so keep its cache empty
+    const cache = new Cache({
+        sourceTTL: 0, // retain for 0 milliseconds
```
nabalone, line 28: Does updating bloom clear the local storage? If not:
I did not give the cache a new prefix. This means that it if a user has failed gravatar urls in the cache, it may take up to a week for those cached urls to expire and therefore for this change to be visible. I thought this was better than the alternative, in which react-avatar starts using a cache at a new prefix and never clears whatever was in the default prefix cache.

But for now, to test this PR, you will need to clear local storage

nabalone:
I'm also not sure how to realistically simulate a restart of Bloom when running locally. Looks like I am needing to clean and rebuild in order to clear the browser cache.

andrew-polk (approved):
Clearing the browser cache should happen on every run of 5.5.
At least, that's what the code in WebView2Browser.InitWebView() is claiming to do...

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5950#-:-NYUmfnbA4AnWsjTudVI:b-tg8yzp)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @nabalone)

___
*[`src/BloomBrowserUI/react_components/bloomAvatar.tsx` line 28 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5950#gh-1237580331:-NYUmQ6x8vXxgMR8kgKU:bv69t9h) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/39a2e3b9e952518ed4fbd42593abadb8259412cd/src/BloomBrowserUI/react_components/bloomAvatar.tsx#L28)):*

Does updating bloom clear the local storage? If not:
I did not give the cache a new prefix. This means that it if a user has failed gravatar urls in the cache, it may take up to a week for those cached urls to expire and therefore for this change to be visible. I thought this was better than the alternative, in which react-avatar starts using a cache at a new prefix and never clears whatever was in the default prefix cache.

But for now, to test this PR, you will need to clear local storage

</blockquote></details>

I don't think it does clear local storage, no.
And, actually, I don't know how a tester can clear local storage. So if the tester already has problematic ones stored, we may need a dev to test.
But hopefully a tester can just use a new email which isn't already in the cache.

PR #5952, opened 2023-06-21
[5.5] BL-12333 Second try at unauthorized content pt 2

gmartin7:
* bloomPUB publishing

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5952)

### DIFF
--- a/src/BloomBrowserUI/publish/ReaderPublish/ReaderPublishScreen.tsx
+++ b/src/BloomBrowserUI/publish/ReaderPublish/ReaderPublishScreen.tsx
@@ -157,6 +157,9 @@ const ReaderPublishScreenInternal: React.FunctionComponent<{
         "&roundPageWidthToNearestK=2" + // Fractional pixels can cause a small sliver of the next page or background color to show (See BL-11497)
         "&roundMarginToNearestK=2"; // Fractional pixels can cause a small sliver of the next page or background color to show (See BL-11497)
 
+    const showBlankPreviewScreen =
+        !props.showPreview || bookUrl === "stopPreview";
+
     const mainPanel = (
         <React.Fragment>
             <PublishPanel
@@ -187,7 +190,7 @@ const ReaderPublishScreenInternal: React.FunctionComponent<{
                     defaultLandscape={defaultLandscape}
                     canRotate={canRotate}
                     // The following leaves a blank screen until the Preview button is pressed
-                    url={`${props.showPreview ? previewUrl : ""}`}
+                    url={`${showBlankPreviewScreen ? "" : previewUrl}`}
                     showPreviewButton={true}
                     highlightPreviewButton={highlightPreview}
                     onPreviewButtonClicked={() => {
--- a/src/BloomExe/web/WebSocketProgress.cs
+++ b/src/BloomExe/web/WebSocketProgress.cs
@@ -114,6 +114,15 @@ namespace Bloom.web
 
 		public bool LogAllMessages;
 
+		/// <summary>
+		/// Resets all the flags that track whether we've reported any problems.
+		/// </summary>
+		public void Reset()
+		{
+			HaveProblemsBeenReported = false;
+			HasFatalProblemBeenReported = false;
+		}
+
 		public virtual void MessageWithoutLocalizing(string message, ProgressKind kind=ProgressKind.Progress)
 		{
 			dynamic messageBundle = new DynamicJson();
--- a/src/BloomExe/web/controllers/PublishApi.cs
+++ b/src/BloomExe/web/controllers/PublishApi.cs
@@ -346,6 +346,7 @@ namespace Bloom.web.controllers
 		/// <returns>True if the preview was updated successfully, false otherwise.</returns>
 		internal bool UpdatePreview(ApiRequest request, bool forVideo)
 		{
+			_progress.Reset(); // Otherwise errors get carried over between runs of the preview.
 			InitializeLanguagesInBook(request);
 			_lastSettings = GetSettings();
 			_lastThumbnailBackgroundColor = _thumbnailBackgroundColor;
@@ -370,9 +371,11 @@ namespace Bloom.web.controllers
 			// the page-range control.
 			_lastSettings.RemoveInteractivePages = forVideo;
 			PreviewUrl = MakeBloomPubForPreview(request.CurrentBook, _bookServer, _progress, _thumbnailBackgroundColor, _lastSettings);
-			if (_progress.HaveProblemsBeenReported)
+			if (PreviewUrl == null)
 			{
-				_webSocketServer.SendString(kWebSocketContext, kWebsocketEventId_Preview, "");
+				// Tried sending empty string, but SendString() ignores empty messages. "stopPreview" gets interpreted as a command to stop
+				// the preview spinner.
+				_webSocketServer.SendString(kWebSocketContext, kWebsocketEventId_Preview, "stopPreview");
 				return false;
 			}
 			_webSocketServer.SendString(kWebSocketContext, kWebsocketEventId_Preview, PreviewUrl);
@@ -412,7 +415,8 @@ namespace Bloom.web.controllers
 		/// <summary>
 		/// Generates an unzipped, staged BloomPUB from the book
 		/// </summary>
-		/// <returns>A valid, well-formed URL on localhost that points to the staged book's htm file</returns>
+		/// <returns>A valid, well-formed URL on localhost that points to the staged book's htm file,
+		/// or null if we aren't allowed to publish this book in this language (LicenseChecker).</returns>
 		public string MakeBloomPubForPreview(Book.Book book, BookServer bookServer, WebSocketProgress progress, Color backColor, BloomPubPublishSettings settings = null)
 		{
 			progress.Message("PublishTab.Epub.PreparingPreview", "Preparing Preview");  // message shared with Epub publishing


andrew-polk (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5952)* status: all files reviewed, 1 unresolved discussion (waiting on @gmartin7)

___
*[`src/BloomExe/web/WebSocketProgress.cs` line 145 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5952#-NYUiwoR26SWwTLYvLA1:-NYUiwoR26SWwTLYvLA2:byhpubb) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/2148542bbcbda69c6e74c36380fafc8e3cca3c09/src/BloomExe/web/WebSocketProgress.cs#L145)):*
> ```Smalltalk
> 					HaveProblemsBeenReported = true;
> 					HaveDisablingErrorsBeenReported = true;
> 					break;
> ```

I'm probably just missing it, but what does DisablingError add that Error doesn't already provide?
When would we want an error which doesn't disable the output?

So, I think there's only three scenarios:
1. An error so bad we must close: Fatal.
1. An error which should prevent the action from completing: Error.
1. An error/warning which should NOT prevent the action from completing: Warning.

I think I understand the need for this switch to distinguish between Error and Warning... otherwise we don't know if the action should continue. But shouldn't that be switched between Error/Warning?

gmartin7 (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5952)* status: all files reviewed, 1 unresolved discussion (waiting on @andrew-polk)

___
*[`src/BloomExe/web/WebSocketProgress.cs` line 145 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5952#-NYUiwoR26SWwTLYvLA1:-NYUpCgV6QvznB_VCsCe:b-zelzuz) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/2148542bbcbda69c6e74c36380fafc8e3cca3c09/src/BloomExe/web/WebSocketProgress.cs#L145)):*

I'm probably just missing it, but what does DisablingError add that Error doesn't already provide?
When would we want an error which doesn't disable the output?

So, I think there's only three scenarios:
1. An error so bad we must close: Fatal.
1. An error which should prevent the action from completing: Error.
1. An error/warning which should NOT prevent the action from completing: Warning.

I think I understand the need for this switch to distinguish between Error and Warning... otherwise we don't know if the action should continue. But shouldn't that be switched between Error/Warning?
</blockquote></details>

Yes, but. Also Warnings are yellow and Errors are red. And currently neither of them abort the preview enough that it can't be saved.

gmartin7 (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5952)* status: all files reviewed, 1 unresolved discussion (waiting on @andrew-polk)

___
*[`src/BloomExe/web/WebSocketProgress.cs` line 145 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5952#-NYUiwoR26SWwTLYvLA1:-NYUpxyx4g2ao-cX4gHW:b-blqd4i) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/2148542bbcbda69c6e74c36380fafc8e3cca3c09/src/BloomExe/web/WebSocketProgress.cs#L145)):*

Yes, but. Also Warnings are yellow and Errors are red. And currently neither of them abort the preview enough that it can't be saved.
</blockquote></details>

Illegal/Doubtful fonts are Errors, but Bloom replaces them with Andika and soldiers on.

andrew-polk (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5952)* status: all files reviewed, 1 unresolved discussion (waiting on @gmartin7)

___
*[`src/BloomExe/web/WebSocketProgress.cs` line 145 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5952#-NYUiwoR26SWwTLYvLA1:-NYUshtG-sy84VUL6bBJ:b1ligps) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/2148542bbcbda69c6e74c36380fafc8e3cca3c09/src/BloomExe/web/WebSocketProgress.cs#L145)):*

Illegal/Doubtful fonts are Errors, but Bloom replaces them with Andika and soldiers on.
</blockquote></details>

Are you saying that before and after your change, the only difference between Error and Warning is the color?

It is possible that JH intends to have 4 cases (two of which are distinct only by color) as opposed to the 3 I outlined above, but that seems unlikely.

Let me ask the question another way to make sure I'm understanding:
If we change all the ones which should stop the action to Error and all the ones which shouldn't to Warning -- and we leave in your distinction in this switch such that UpdatePreview can tell the difference between Warning and Error -- is the color change the only bad thing which happens? Or are we too ingrained with this system now?

gmartin7 (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5952)* status: all files reviewed, 1 unresolved discussion (waiting on @andrew-polk)

___
*[`src/BloomExe/web/WebSocketProgress.cs` line 145 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5952#-NYUiwoR26SWwTLYvLA1:-NYUwse74VUDZustHcWd:b-70253c) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/2148542bbcbda69c6e74c36380fafc8e3cca3c09/src/BloomExe/web/WebSocketProgress.cs#L145)):*

Are you saying that before and after your change, the only difference between Error and Warning is the color?

It is possible that JH intends to have 4 cases (two of which are distinct only by color) as opposed to the 3 I outlined above, but that seems unlikely.

Let me ask the question another way to make sure I'm understanding:
If we change all the ones which should stop the action to Error and all the ones which shouldn't to Warning -- and we leave in your distinction in this switch such that UpdatePreview can tell the difference between Warning and Error -- is the color change the only bad thing which happens? Or are we too ingrained with this system now?
</blockquote></details>

First question: I believe so.
Second question: This PR's case is the only one I know of that should stop the action, other than actual Exceptions that generate "Fatal" messages that close Bloom.

andrew-polk (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5952)* status: all files reviewed, 1 unresolved discussion (waiting on @gmartin7)

___
*[`src/BloomExe/web/WebSocketProgress.cs` line 145 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5952#-NYUiwoR26SWwTLYvLA1:-NYVNGP-DTkBxvVx66-t:bxqbscw) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/2148542bbcbda69c6e74c36380fafc8e3cca3c09/src/BloomExe/web/WebSocketProgress.cs#L145)):*

First question: I believe so.
Second question: This PR's case is the only one I know of that should stop the action, other than actual Exceptions that generate "Fatal" messages that close Bloom.
</blockquote></details>

Just a quick note that I'm continuing to look at this.
I don't think we can sort of change the definition of Error to mean we should stop. 
1. It is used all over in other contexts.
2. I'm becoming more convinced that the distinction of these `ProgressKind`s should only affect UI, not behavior.

So mostly I'm looking at how we do what we need to here without relying on ProgressKind at all.

andrew-polk (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5952)* status: all files reviewed, 2 unresolved discussions (waiting on @gmartin7)

___
*[`src/BloomExe/web/WebSocketProgress.cs` line 145 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5952#-NYUiwoR26SWwTLYvLA1:-NYVOs3s2_TVOKBdyZQ1:b1xwjwp) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/2148542bbcbda69c6e74c36380fafc8e3cca3c09/src/BloomExe/web/WebSocketProgress.cs#L145)):*

Just a quick note that I'm continuing to look at this.
I don't think we can sort of change the definition of Error to mean we should stop. 
1. It is used all over in other contexts.
2. I'm becoming more convinced that the distinction of these `ProgressKind`s should only affect UI, not behavior.

So mostly I'm looking at how we do what we need to here without relying on ProgressKind at all.
</blockquote></details>

Ok, I added another comment in context on how I think we can do this.
___
*[`src/BloomExe/web/controllers/PublishApi.cs` line 374 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5952#-NYVOQ5_1O3DM6Ui-eb3:-NYVOQ5_1O3DM6Ui-eb4:bavei2u) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/2148542bbcbda69c6e74c36380fafc8e3cca3c09/src/BloomExe/web/controllers/PublishApi.cs#L374)):*
> ```Smalltalk
> 			_lastSettings.RemoveInteractivePages = forVideo;
> 			PreviewUrl = MakeBloomPubForPreview(request.CurrentBook, _bookServer, _progress, _thumbnailBackgroundColor, _lastSettings);
> 			if (_progress.HaveDisablingErrorsBeenReported)
> ```

Instead of checking the state of what has been sent to progress here, what if we just check if PreviewUrl is non-null?

So,
```
			PreviewUrl = MakeBloomPubForPreview(request.CurrentBook, _bookServer, _progress, _thumbnailBackgroundColor, _lastSettings);
			if (PreviewUrl == null)
			{
				// Tried sending empty string, but SendString() ignores empty messages. "stopPreview" gets interpreted as a command to stop
				// the preview spinner.
				_webSocketServer.SendString(kWebSocketContext, kWebsocketEventId_Preview, "stopPreview");
				return false;
			}
			_webSocketServer.SendString(kWebSocketContext, kWebsocketEventId_Preview, PreviewUrl);
```

Then we can get rid of the extra complexity added to ProgressKind.

gmartin7 (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5952)* status: 1 of 6 files reviewed, 1 unresolved discussion (waiting on @andrew-polk)

___
*[`src/BloomExe/web/controllers/PublishApi.cs` line 374 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5952#-NYVOQ5_1O3DM6Ui-eb3:-NYZ6Mci9jjUHgyUbGlE:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/2148542bbcbda69c6e74c36380fafc8e3cca3c09/src/BloomExe/web/controllers/PublishApi.cs#L374)):*

Instead of checking the state of what has been sent to progress here, what if we just check if PreviewUrl is non-null?

So,
```
			PreviewUrl = MakeBloomPubForPreview(request.CurrentBook, _bookServer, _progress, _thumbnailBackgroundColor, _lastSettings);
			if (PreviewUrl == null)
			{
				// Tried sending empty string, but SendString() ignores empty messages. "stopPreview" gets interpreted as a command to stop
				// the preview spinner.
				_webSocketServer.SendString(kWebSocketContext, kWebsocketEventId_Preview, "stopPreview");
				return false;
			}
			_webSocketServer.SendString(kWebSocketContext, kWebsocketEventId_Preview, PreviewUrl);
```

Then we can get rid of the extra complexity added to ProgressKind.
</blockquote></details>

Done.

andrew-polk (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5952)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @gmartin7)

PR #5956, opened 2023-06-22
Move lock to beginning of function

jsubloom:
not that it really matters much, but the thinking is that if GetBookInfoFromRequestParam throws, in the old code the count of _thumbnailEventsToWaitFor would be wrong.


JohnThomson (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5956)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @jsubloom)

PR #5964, opened 2023-06-27
Cleanup errant spaces from ckeditor (BL-12391)

andrew-polk:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5964)

### DIFF
--- a/src/BloomBrowserUI/bookEdit/editablePage.ts
+++ b/src/BloomBrowserUI/bookEdit/editablePage.ts
@@ -12,7 +12,6 @@ import { theOneBubbleManager, BubbleManager } from "./js/bubbleManager";
 // This allows strong typing to be done for exported functions
 export interface IPageFrameExports {
     pageSelectionChanging(): void;
-    prepareToSavePage(): void;
     pageUnloading(): void;
     disconnectForGarbageCollection(): void;
     copySelection(): void;
--- a/src/BloomBrowserUI/bookEdit/js/bloomEditing.ts
+++ b/src/BloomBrowserUI/bookEdit/js/bloomEditing.ts
@@ -1234,6 +1234,17 @@ export const getBodyContentForSavePage = () => {
     if (document.activeElement instanceof HTMLElement) {
         document.activeElement.blur();
     }
+
+    // Get the cleaned up data (getData()) from ckeditor, rather than just the raw html.
+    // Specifically, we want it to remove the zero-width space characters that it inserts.
+    // It inserts them to aid in preserving the cursor position, but we're saving the page;
+    // we don't need the cursor position preserved.
+    // See BL-12391.
+    for (const property in CKEDITOR.instances) {
+        const instance = CKEDITOR.instances[property];
+        instance.element.setHtml(instance.getData());
+    }
+
     const result = document.body.innerHTML;
     if (bubbleEditingOn) {
         theOneBubbleManager.turnOnBubbleEditing();
--- a/src/BloomBrowserUI/bookEdit/toolbox/readers/decodableReader/decodableReaderToolboxTool.ts
+++ b/src/BloomBrowserUI/bookEdit/toolbox/readers/decodableReader/decodableReaderToolboxTool.ts
@@ -171,6 +171,7 @@ export class DecodableReaderToolboxTool implements ITool {
         getTheOneReaderToolsModel().doMarkup();
     }
     public async updateMarkupAsync() {
+        // If you implement this, you may need to do something like cleanUpCkEditorHtml() in audioRecording.ts.
         throw "not implemented...use updateMarkup";
         return () => undefined;
     }
--- a/src/BloomBrowserUI/bookEdit/toolbox/readers/leveledReader/leveledReaderToolboxTool.ts
+++ b/src/BloomBrowserUI/bookEdit/toolbox/readers/leveledReader/leveledReaderToolboxTool.ts
@@ -68,6 +68,7 @@ export class LeveledReaderToolboxTool implements ITool {
         getTheOneReaderToolsModel().doMarkup();
     }
     public async updateMarkupAsync() {
+        // If you implement this, you may need to do something like cleanUpCkEditorHtml() in audioRecording.ts.
         throw "not implemented...use updateMarkup";
         return () => undefined;
     }
--- a/src/BloomBrowserUI/bookEdit/toolbox/talkingBook/audioRecording.ts
+++ b/src/BloomBrowserUI/bookEdit/toolbox/talkingBook/audioRecording.ts
@@ -3189,6 +3189,8 @@ export default class AudioRecording {
         const currentMarker = copy.find(".bloom-ui-current-audio-marker");
         currentMarker.remove();
 
+        this.cleanUpCkEditorHtml(elt.get(0), copy.get(0));
+
         const markedSentences = copy.find(
             `${kAudioSentenceClassSelector},.${kSegmentClass}`
         );
@@ -3303,6 +3305,35 @@ export default class AudioRecording {
             elt.html(newHtml);
             elt.append(formatButton);
         };
+    }
+
+    // When we switched to webview2, we started getting errant zero-width spaces in the text from ckeditor.
+    // The "right" way to get text from ckeditor boxes is to call getData(). So that's what we're doing here.
+    // It does clean up at least most of the zero-width spaces.
+    // See BL-12391.
+    private cleanUpCkEditorHtml(element: HTMLElement, copy: HTMLElement) {
+        const editableDiv = element.closest(".bloom-editable") as HTMLElement;
+        if (!editableDiv) return;
+
+        const ckeditorOfThisBox = (<any>editableDiv).bloomCkEditor;
+        if (!ckeditorOfThisBox) return;
+
+        if (editableDiv.innerHTML !== ckeditorOfThisBox.getData()) {
+            // Flag the element we are processing so we can find it in the version we make from ckeditor's getData().
+            element.setAttribute("data-element-we-are-processing", "this-one");
+
+            // Create a dummy element just so we can stash the result of getData() and then find our element in it.
+            // getData() is for the whole text box, but we are processing one of the child elements.
+            const newElement = document.createElement("div");
+            newElement.innerHTML = ckeditorOfThisBox.getData(); // have to call getData() again so it contains data-element-we-are-processing
+            const newChild = newElement.querySelector(
+                "[data-element-we-are-processing]"
+            );
+            copy.innerHTML = newChild!.innerHTML;
+
+            // Make sure we remove the flag; we don't want to modify the original element.
+            element.removeAttribute("data-element-we-are-processing");
+        }
     }
 
     // Normalization rules for text which has already been processed by CKEditor
--- a/src/BloomBrowserUI/bookEdit/toolbox/toolbox.ts
+++ b/src/BloomBrowserUI/bookEdit/toolbox/toolbox.ts
@@ -38,6 +38,7 @@ export interface ITool {
     // Because it is async, it is not guaranteed that all the async processing will complete before another keystroke is received.
     // To guard against this, it should make no changes to the document; rather, it returns a function which will,
     // synchronously, make the changes. Toolbox will call this returned function iff no more keystrokes have been received.
+    // Note, new implementations of updateMarkupAsync may need to implement something like cleanUpCkEditorHtml() in audioRecording.ts.
     updateMarkupAsync(): Promise<() => void>;
     isUpdateMarkupAsync(): boolean; // should return true if updateMarkupAsync should be called and awaited instead of updateMarkup.
     newPageReady(); // called when a new page is displayed or tool is activated (called after showTool completes)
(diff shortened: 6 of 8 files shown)


JohnThomson (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5964)* status: all files reviewed, 5 unresolved discussions (waiting on @andrew-polk)

___
*[`src/BloomBrowserUI/bookEdit/js/bloomEditing.ts` line 1239 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5964#-NZ1p0DjE01wt5K_TGNy:-NZ1p0Dk46wuB55Yp1jp:bpmdx83) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/503b3959078f8125e99782acb85c67706a4a5d1b/src/BloomBrowserUI/bookEdit/js/bloomEditing.ts#L1239)):*
> ```XML
> 
>     // Get the cleaned up data (getData()) from ckeditor, rather than just the raw html.
>     // Specifically, we want it to remove the zero-width space characters that it inserts. See BL-12391.
> ```

I think we should capture somewhere (assuming it's true) that removing the ZWSP is not a problem. After all, CkEditor puts it in for a purpose, I think to make it possible to have a selection at a location that WebView2 and similar browsers would not otherwise allow.
Here, we're saving the page, and I don't think we expect the insertion point position to be preserved. Elsewhere, I'm guessing it's OK because we insert a bookmark and CkEditor re-inserts the ZWSP when it moves the IP to the bookmark.
___
*[`src/BloomBrowserUI/bookEdit/toolbox/toolbox.ts` line 1018 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5964#-NZ1qZVS7B4CXaCOnIQ-:-NZ1qZVS7B4CXaCOnIQ0:bob382l) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/503b3959078f8125e99782acb85c67706a4a5d1b/src/BloomBrowserUI/bookEdit/toolbox/toolbox.ts#L1018)):*
> ```XML
> // We want to remove nbsps unless
> // 1. they are at the start or end of the div
> // 2. they are adjacent to another space
> ```

Rather than "another space" I think we should say "an ordinary space" and perhaps note that they are needed adjacent to ordinary spaces to prevent the renderer from collapsing sequences of ordinary spaces into a single space, but not needed adjacent to other white space characters which don't have that special behavior.
___
*[`src/BloomBrowserUI/bookEdit/toolbox/toolbox.ts` line 1033 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5964#-NZ1qDZb5-d75j22xAE-:-NZ1qDZb5-d75j22xAE0:bynp9yg) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/503b3959078f8125e99782acb85c67706a4a5d1b/src/BloomBrowserUI/bookEdit/toolbox/toolbox.ts#L1033)):*
> ```XML
>     // The basic algorithm here is to look through the text to discover the adjacent characters, but
>     // do the actual replacement in the html so as to keep the markup. So we simultaneously loop
>     // through the text and the html, finding each corresponding nbsp.
> ```

Possibly add (here or below): "We also make the replacements in the editableDivText so that, for example, if we change one NBSP to a regular space, that space prevents converting an adjacent NBSP.
___
*[`src/BloomBrowserUI/bookEdit/toolbox/toolbox.ts` line 1039 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5964#-NZ1rCaa7BqMhQheMZO4:-NZ1rCaa7BqMhQheMZO5:bdi0kaf) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/503b3959078f8125e99782acb85c67706a4a5d1b/src/BloomBrowserUI/bookEdit/toolbox/toolbox.ts#L1039)):*
> ```XML
>         if (i === -1) break;
>         j = editableDivText.indexOf("\u00A0", j + 1);
>         if (j === 0 || j === editableDivText.length - 1) continue;
> ```

To guard against a crash in some really pathological case (like &nbsp; in an attribute value) I think we should break if j === -1. Maybe also report it, at least console.error().
___
*[`src/BloomBrowserUI/bookEdit/toolbox/talkingBook/audioRecording.ts` line 3328 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5964#-NZ1ycZHA5FDW_MGBjtW:-NZ1ycZHA5FDW_MGBjtX:b-cy1xfu) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/503b3959078f8125e99782acb85c67706a4a5d1b/src/BloomBrowserUI/bookEdit/toolbox/talkingBook/audioRecording.ts#L3328)):*
> ```XML
>             // getData() is for the whole text box, but we are processing one of the child elements.
>             const newElement = document.createElement("div");
>             newElement.innerHTML = ckeditorOfThisBox.getData();
> ```

Let's comment here. I nearly suggested what looks like an obvious optimization, saving the result of getData() in a variable. But we need to call it again to get a different result that includes data-element-we-are-processing.

andrew-polk (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5964)* status: 4 of 8 files reviewed, 5 unresolved discussions (waiting on @JohnThomson)

___
*[`src/BloomBrowserUI/bookEdit/js/bloomEditing.ts` line 1239 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5964#-NZ1p0DjE01wt5K_TGNy:-NZ3XTr9BT5wUwXPIZk3:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/503b3959078f8125e99782acb85c67706a4a5d1b/src/BloomBrowserUI/bookEdit/js/bloomEditing.ts#L1239)):*

I think we should capture somewhere (assuming it's true) that removing the ZWSP is not a problem. After all, CkEditor puts it in for a purpose, I think to make it possible to have a selection at a location that WebView2 and similar browsers would not otherwise allow.
Here, we're saving the page, and I don't think we expect the insertion point position to be preserved. Elsewhere, I'm guessing it's OK because we insert a bookmark and CkEditor re-inserts the ZWSP when it moves the IP to the bookmark.
</blockquote></details>

Done.
___
*[`src/BloomBrowserUI/bookEdit/toolbox/toolbox.ts` line 1018 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5964#-NZ1qZVS7B4CXaCOnIQ-:-NZ3VzZ4EjaqO3Jw_vus:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/503b3959078f8125e99782acb85c67706a4a5d1b/src/BloomBrowserUI/bookEdit/toolbox/toolbox.ts#L1018)):*

Rather than "another space" I think we should say "an ordinary space" and perhaps note that they are needed adjacent to ordinary spaces to prevent the renderer from collapsing sequences of ordinary spaces into a single space, but not needed adjacent to other white space characters which don't have that special behavior.
</blockquote></details>

Done.
___
*[`src/BloomBrowserUI/bookEdit/toolbox/toolbox.ts` line 1033 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5964#-NZ1qDZb5-d75j22xAE-:-NZ3WpPo-6i9UJqC3L_5:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/503b3959078f8125e99782acb85c67706a4a5d1b/src/BloomBrowserUI/bookEdit/toolbox/toolbox.ts#L1033)):*

Possibly add (here or below): "We also make the replacements in the editableDivText so that, for example, if we change one NBSP to a regular space, that space prevents converting an adjacent NBSP.
</blockquote></details>

Done.
___
*[`src/BloomBrowserUI/bookEdit/toolbox/toolbox.ts` line 1039 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5964#-NZ1rCaa7BqMhQheMZO4:-NZ3W3EB0DQ2aWlbOVz3:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/503b3959078f8125e99782acb85c67706a4a5d1b/src/BloomBrowserUI/bookEdit/toolbox/toolbox.ts#L1039)):*

To guard against a crash in some really pathological case (like &nbsp; in an attribute value) I think we should break if j === -1. Maybe also report it, at least console.error().
</blockquote></details>

Done.
___
*[`src/BloomBrowserUI/bookEdit/toolbox/talkingBook/audioRecording.ts` line 3328 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5964#-NZ1ycZHA5FDW_MGBjtW:-NZ3Y1-J8G_QlkYYRntB:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/503b3959078f8125e99782acb85c67706a4a5d1b/src/BloomBrowserUI/bookEdit/toolbox/talkingBook/audioRecording.ts#L3328)):*

Let's comment here. I nearly suggested what looks like an obvious optimization, saving the result of getData() in a variable. But we need to call it again to get a different result that includes data-element-we-are-processing.
</blockquote></details>

Done.

JohnThomson (changes requested):
Close, but I think I found one definite bug, and made one other suggestion.

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5964#-:-NZBxtE78VlolxutvV2H:b-xa1bk)* status: all files reviewed, 5 unresolved discussions (waiting on @andrew-polk)

___
*[`src/BloomBrowserUI/bookEdit/toolbox/toolbox.ts` line 1028 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5964#-NZBvgSV56Gya7JTk1o8:-NZBvgSV56Gya7JTk1o9:b-72010y) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/661292eed66c5ae1fd7aaf7d3550c986435eb440/src/BloomBrowserUI/bookEdit/toolbox/toolbox.ts#L1028)):*
> ```XML
> 
>     const originalHtml = editableDiv.innerHTML;
>     let editableDivHtml = editableDiv.innerHTML;
> ```

If we keep both of these (see below), it's probably faster to initialize one from the other.
___
*[`src/BloomBrowserUI/bookEdit/toolbox/toolbox.ts` line 1040 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5964#-NZBuILe6zwRTUbNk9O3:-NZBuILe6zwRTUbNk9O4:b-cfy4kq) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/661292eed66c5ae1fd7aaf7d3550c986435eb440/src/BloomBrowserUI/bookEdit/toolbox/toolbox.ts#L1040)):*
> ```XML
>     let j = -1;
>     // Simultaneously loop through the text and the html, finding each corresponding nbsp.
>     // The text shows us what the adjacent characters so we can make a replacement decision, but
> ```

typo should be ...what the adjacent characters ARE...
___
*[`src/BloomBrowserUI/bookEdit/toolbox/toolbox.ts` line 1057 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5964#-NZBukA57HFmAJcDGpkG:-NZBukA67gaCvEycLkRU:b886x46) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/661292eed66c5ae1fd7aaf7d3550c986435eb440/src/BloomBrowserUI/bookEdit/toolbox/toolbox.ts#L1057)):*
> ```XML
>             console.error("editableDivText: " + editableDivText);
>             editableDivHtml = originalHtml;
>             return;
> ```

If we return here, there's no need to restore editableDivHtml; it's a local variable. And then there's no need for originalHtml at all.
But, we need to restore the bookmark! We could either just do that here (without ever setting editableDiv.innerHtml), or we can break instead of returning, in which case we probably should restore editableDivHtml.
___
*[`src/BloomBrowserUI/bookEdit/toolbox/toolbox.ts` line 1083 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5964#-NZBwCVe4VQhT20GtDMj:-NZBwCVe4VQhT20GtDMk:b-hf7sjo) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/661292eed66c5ae1fd7aaf7d3550c986435eb440/src/BloomBrowserUI/bookEdit/toolbox/toolbox.ts#L1083)):*
> ```XML
>     // We actually only expect one; but being safe.
>     const ckeBookmarks = editableDiv.querySelectorAll("[id^='cke_bm_']");
>     ckeBookmarks.forEach(bm => (bm.innerHTML = content));
> ```

I find myself wanting to make this more robust. It would be nice, for example, if it would go on working if we switch to a future CkEditor that puts, say, a zwsp or even a bookmark number as the hidden text.
What would you think of making it return the old content of the first (or last) bookmark, or undefined if there aren't any? Then we pass what we got from the first call to the second one (and can skip doing it at all if undefined).

andrew-polk (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5964)* status: all files reviewed, 4 unresolved discussions (waiting on @JohnThomson)

___
*[`src/BloomBrowserUI/bookEdit/toolbox/toolbox.ts` line 1028 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5964#-NZBvgSV56Gya7JTk1o8:-NZC2eZa7ShU7f0G-3rP:bss4a2b) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/661292eed66c5ae1fd7aaf7d3550c986435eb440/src/BloomBrowserUI/bookEdit/toolbox/toolbox.ts#L1028)):*

If we keep both of these (see below), it's probably faster to initialize one from the other.
</blockquote></details>

Removed.
___
*[`src/BloomBrowserUI/bookEdit/toolbox/toolbox.ts` line 1057 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5964#-NZBukA57HFmAJcDGpkG:-NZC-5C3CrX3aWI3ba-u:bh35oaz) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/661292eed66c5ae1fd7aaf7d3550c986435eb440/src/BloomBrowserUI/bookEdit/toolbox/toolbox.ts#L1057)):*

If we return here, there's no need to restore editableDivHtml; it's a local variable. And then there's no need for originalHtml at all.
But, we need to restore the bookmark! We could either just do that here (without ever setting editableDiv.innerHtml), or we can break instead of returning, in which case we probably should restore editableDivHtml.
</blockquote></details>

Yes, I flubbed that. Thanks.
I've added a test now, too.
___
*[`src/BloomBrowserUI/bookEdit/toolbox/toolbox.ts` line 1083 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5964#-NZBwCVe4VQhT20GtDMj:-NZC2bqX-ZuJKSbbr0EC:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/661292eed66c5ae1fd7aaf7d3550c986435eb440/src/BloomBrowserUI/bookEdit/toolbox/toolbox.ts#L1083)):*

I find myself wanting to make this more robust. It would be nice, for example, if it would go on working if we switch to a future CkEditor that puts, say, a zwsp or even a bookmark number as the hidden text.
What would you think of making it return the old content of the first (or last) bookmark, or undefined if there aren't any? Then we pass what we got from the first call to the second one (and can skip doing it at all if undefined).
</blockquote></details>

Done.

JohnThomson (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5964#-:-NZCh05bAee6Xsqn7Qvi:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @andrew-polk)

PR #5982, opened 2023-07-06
Bl 12439 spreadsheet audio and formatting

nabalone:
Fix: when you import from spreadsheet with both audio and formatting, formatting tags can end up in the text

### DIFF
--- a/src/BloomExe/Spreadsheet/SpreadsheetImporter.cs
+++ b/src/BloomExe/Spreadsheet/SpreadsheetImporter.cs
@@ -1645,7 +1645,11 @@ namespace Bloom.Spreadsheet
 							else
 							{
 								// a white space fragment.
-								var node = para.OwnerDocument.CreateTextNode(fragment);
+								// There may be formatting tags (e.g. italics or colors) which we will ignore since this is just whitespace anyway
+								// And otherwise they will get escaped and end up in the text
+								var tempElement = para.OwnerDocument.CreateElement("span");
+								tempElement.InnerXml = fragment;
+								var node = para.OwnerDocument.CreateTextNode(tempElement.InnerText);
 								para.AppendChild(node);
 							}
 						}


src/BloomExe/Spreadsheet/SpreadsheetImporter.cs
```diff
@@ -1645,7 +1645,11 @@ private async Task AddAudioAsync(XmlElement editable, string lang, ContentRow ro
 							else
 							{
 								// a white space fragment.
-								var node = para.OwnerDocument.CreateTextNode(fragment);
+								// There may be formatting tags (e.g. italics or colors) which we will ignore since this is just whitespace anyway
+								// And otherwise they will get escaped and end up in the text
+								var tempElement = para.OwnerDocument.CreateElement("span");
```
nabalone, line 1650: This is a kind of hacky. Is there a better way to do it?

StephenMcConnel (approved):
:lgtm: with one comment in response to a question.

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5982#-:-N_4d-qF2eLCHGk-F1EE:biyrumm)* status: all files reviewed, 1 unresolved discussion (waiting on @nabalone)

___
*[`src/BloomExe/Spreadsheet/SpreadsheetImporter.cs` line 1650 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5982#gh-1254754415:-N_4aJag-BKhwhQuRWbu:b-tftghd) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/7dffd204385efd14041f7b90bcbf0d848dc1ed8f/src/BloomExe/Spreadsheet/SpreadsheetImporter.cs#L1650)):*

This is a kind of hacky. Is there a better way to do it?
</blockquote></details>

One alternative might be something like this to get rid of all XML tags:
	fragment = Regex.Replace(fragment, @"<[^>]*>", "");
This would be just before the existing code.  I'm not sure whether this is more or less hacky than what you have.

nabalone (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5982)* status: all files reviewed, 1 unresolved discussion (waiting on @StephenMcConnel)

___
*[`src/BloomExe/Spreadsheet/SpreadsheetImporter.cs` line 1650 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5982#gh-1254754415:-N_5KYiyC6U3gsDSKIBe:bimiou9) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/7dffd204385efd14041f7b90bcbf0d848dc1ed8f/src/BloomExe/Spreadsheet/SpreadsheetImporter.cs#L1650)):*

One alternative might be something like this to get rid of all XML tags:
	fragment = Regex.Replace(fragment, @"<[^>]*>", "");
This would be just before the existing code.  I'm not sure whether this is more or less hacky than what you have.
</blockquote></details>

Hmm, thinking about this. I like my version better because the fragment really was basically InnerXml, so I'd rather use the XmlElement library to parse it rather than essentially write new parsing code with the regex

StephenMcConnel (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5982)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @nabalone)

PR #5984, opened 2023-07-06
BL-12436 failed spreadsheet shouldnt open

nabalone:
If you export to spreadsheet and it fails because e.g. you have that spreadsheet still open in excel, you get an error message but that spreadsheet comes to the foreground anyway, so it seems like it succeeded until you navigate back to bloom and see the error message. 
We should not open/foreground the spreadsheet path if export failed.

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5984)

### DIFF
--- a/src/BloomExe/Spreadsheet/SpreadsheetExporter.cs
+++ b/src/BloomExe/Spreadsheet/SpreadsheetExporter.cs
@@ -747,6 +747,13 @@ namespace Bloom.Spreadsheet
 						ProgressKind.Error, _outputFolder);
 				}
 			}
+			catch (Exception e)
+			{
+				progress.MessageWithParams("Spreadsheet.ExportFailed", "{0} is a placeholder for the exception message",
+					"Export failed: {0}", ProgressKind.Error, e.Message);
+
+			}
+
 			outputPath = null;
 			return null; // some error occurred and was caught
 		}
--- a/src/BloomExe/Spreadsheet/SpreadsheetExporter.cs
+++ b/src/BloomExe/Spreadsheet/SpreadsheetExporter.cs
@@ -747,10 +747,10 @@ namespace Bloom.Spreadsheet
 			}
 			catch (Exception e)
 			{
-				var msg = LocalizationManager.GetString("Spreadsheet:ExportFailed", "Export failed: ");
-				Console.WriteLine(msg);
+				Console.WriteLine("Export failed: ");
 				Console.WriteLine(e);
-				progress.Message("Spreadsheet:ExportFailed", "", msg + e.Message, ProgressKind.Error);
+				progress.MessageWithParams("Spreadsheet.ExportFailed", "{0} is a placeholder for the exception message",
+					"Export failed: {0}", ProgressKind.Error, e.Message);
 			}
 
 			outputPath = null;
--- a/src/BloomExe/Spreadsheet/SpreadsheetExporter.cs
+++ b/src/BloomExe/Spreadsheet/SpreadsheetExporter.cs
@@ -75,7 +75,17 @@ namespace Bloom.Spreadsheet
 		{
 			var spreadsheet = ExportToFolder(dom, bookFolderPath, outputFolder, out string outputFilePath,
 				progress);
-			resultCallback(outputFilePath);
+			// ExportToFolder will return null if export fails, in which case we don't want to open/foreground the spreadsheet
+			if (spreadsheet != null)
+			{
+				resultCallback(outputFilePath);
+			}
+			else if (!progress.HaveProblemsBeenReported)
+			{
+				// Export failed but we have no progress problem messages. Make sure there is at least some error message
+				// so we aren't just failing silently
+				progress.MessageWithParams("Spreadsheet.ExportFailed", "", "Export failed: {0}", ProgressKind.Error, "");
+			}
 			return progress.HaveProblemsBeenReported;
 		},"collectionTab", "Exporting Spreadsheet", showCancelButton: false);
 		}
@@ -724,11 +734,19 @@ namespace Bloom.Spreadsheet
 			}
 			catch (Exception e) when (e is IOException || e is SecurityException || e is UnauthorizedAccessException)
 			{
-				progress.MessageWithParams("Spreadsheet.WriteFailed", "",
-					"Bloom had problems writing files to that location ({0}). Check that you have permission to write there.",
-					ProgressKind.Error, _outputFolder);
+				if (e is IOException && (e.HResult & 0x0000FFFF) == 32) //ERROR_SHARING_VIOLATION
+				{
+					progress?.Message("Spreadsheet.SpreadsheetLocked", "",
+						"Bloom could not write to the spreadsheet because another program has it locked. Do you have it open in another program?",
+						ProgressKind.Error);
+				}
+				else
+				{
+					progress.MessageWithParams("Spreadsheet.WriteFailed", "",
+						"Bloom had problems writing files to that location ({0}). Check that you have permission to write there.",
+						ProgressKind.Error, _outputFolder);
+				}
 			}
-
 			outputPath = null;
 			return null; // some error occurred and was caught
 		}
--- a/src/BloomExe/Spreadsheet/SpreadsheetExporter.cs
+++ b/src/BloomExe/Spreadsheet/SpreadsheetExporter.cs
@@ -80,12 +80,6 @@ namespace Bloom.Spreadsheet
 			{
 				resultCallback(outputFilePath);
 			}
-			else if (!progress.HaveProblemsBeenReported)
-			{
-				// Export failed but we have no progress problem messages. Make sure there is at least some error message
-				// so we aren't just failing silently
-				progress.MessageWithParams("Spreadsheet.ExportFailed", "", "Export failed: {0}", ProgressKind.Error, "");
-			}
 			return progress.HaveProblemsBeenReported;
 		},"collectionTab", "Exporting Spreadsheet", showCancelButton: false);
 		}
@@ -736,12 +730,16 @@ namespace Bloom.Spreadsheet
 			{
 				if (e is IOException && (e.HResult & 0x0000FFFF) == 32) //ERROR_SHARING_VIOLATION
 				{
+					Console.WriteLine("Writing Spreadsheet failed. Do you have it open in another program?");
+					Console.WriteLine(e);
 					progress?.Message("Spreadsheet.SpreadsheetLocked", "",
 						"Bloom could not write to the spreadsheet because another program has it locked. Do you have it open in another program?",
 						ProgressKind.Error);
 				}
 				else
 				{
+					Console.WriteLine(String.Format("Bloom had problems writing files to that location ({0}). Check that you have permission to write there.", _outputFolder));
+					Console.WriteLine(e);
 					progress.MessageWithParams("Spreadsheet.WriteFailed", "",
 						"Bloom had problems writing files to that location ({0}). Check that you have permission to write there.",
 						ProgressKind.Error, _outputFolder);
@@ -749,9 +747,10 @@ namespace Bloom.Spreadsheet
 			}
 			catch (Exception e)
 			{
-				progress.MessageWithParams("Spreadsheet.ExportFailed", "{0} is a placeholder for the exception message",
-					"Export failed: {0}", ProgressKind.Error, e.Message);
-
+				var msg = LocalizationManager.GetString("Spreadsheet:ExportFailed", "Export failed: ");
+				Console.WriteLine(msg);
+				Console.WriteLine(e);
+				progress.Message("Spreadsheet:ExportFailed", "", msg + e.Message, ProgressKind.Error);
 			}
 
 			outputPath = null;
--- a/src/BloomExe/Spreadsheet/SpreadsheetIO.cs
+++ b/src/BloomExe/Spreadsheet/SpreadsheetIO.cs
@@ -228,26 +228,9 @@ namespace Bloom.Spreadsheet
 					SetBackgroundColorOfColumn(worksheet, iColumn1Based, InternalSpreadsheet.HiddenColor);
 				}
 
-				try
-				{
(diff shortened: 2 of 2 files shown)


src/BloomExe/Spreadsheet/SpreadsheetExporter.cs
```diff
@@ -75,7 +75,17 @@ public SpreadsheetExporter(ILanguageDisplayNameResolver langDisplayNameResolver)
...
+			// ExportToFolder will return null if export fails, in which case we don't want to open/foreground the spreadsheet
+			if (spreadsheet != null)
+			{
+				resultCallback(outputFilePath);
+			}
+			else if (!progress.HaveProblemsBeenReported)
+			{
+				// Export failed but we have no progress problem messages. Make sure there is at least some error message
+				// so we aren't just failing silently
+				progress.MessageWithParams("Spreadsheet.ExportFailed", "", "Export failed: {0}", ProgressKind.Error, "");
```
nabalone, line 87: From a brief glance it looks possible for us to get here without any error messages having been sent to the progress dialog, so I just put a general export failure message that already exists to make sure there is something

src/BloomExe/Spreadsheet/SpreadsheetIO.cs
```diff
@@ -228,26 +228,9 @@ int embedImage(string imageSrcPath, int rowNum, int colNum)
...
 
-				try
-				{
-					RobustFile.Delete(outputPath);
-					var xlFile = new FileInfo(outputPath);
-					package.SaveAs(xlFile);
-				}
-				catch (IOException ex) when ((ex.HResult & 0x0000FFFF) == 32) //ERROR_SHARING_VIOLATION
-				{
-					Console.WriteLine("Writing Spreadsheet failed. Do you have it open in another program?");
```
nabalone, line 239: Do we still need these console messages for anything?

src/BloomExe/Spreadsheet/SpreadsheetExporter.cs
```diff
@@ -75,7 +75,17 @@ public SpreadsheetExporter(ILanguageDisplayNameResolver langDisplayNameResolver)
 		{
 			var spreadsheet = ExportToFolder(dom, bookFolderPath, outputFolder, out string outputFilePath,
 				progress);
-			resultCallback(outputFilePath);
+			// ExportToFolder will return null if export fails, in which case we don't want to open/foreground the spreadsheet
+			if (spreadsheet != null)
+			{
+				resultCallback(outputFilePath);
+			}
+			else if (!progress.HaveProblemsBeenReported)
```
hatton, line 83: Not clear why you don't want to say it failed if previously we said something specific about what went wrong?

nabalone: I really meant this as more of a stopgap. I'm not even sure what to put after the colon in "Export failed:". I think the existing messages about things that went wrong are pretty clear on their own. Otherwise we will get something like this:
image

StephenMcConnel (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5984)* status: all files reviewed, 3 unresolved discussions (waiting on @hatton and @nabalone)

___
*[`src/BloomExe/Spreadsheet/SpreadsheetExporter.cs` line 87 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5984#gh-1254983545:-N_4i89kErtlfqyUpVf9:b-6kzlmm) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/7a6cd2f1b79b7ab7615ced6527d7006119cf89c9/src/BloomExe/Spreadsheet/SpreadsheetExporter.cs#L87)):*

From a brief glance it looks possible for us to get here without any error messages having been sent to the progress dialog, so I just put a general export failure message that already exists to make sure there is something
</blockquote></details>

I don't think it would be a problem (if the trailing colon doesn't bother you) to display the summary message "Export failed:" after the more explicit error messages.  It would sort of fill the slot of the "Done" message we use sometimes, but for errors.  But John is the expert in UX, not me.
___
*[`src/BloomExe/Spreadsheet/SpreadsheetIO.cs` line 0 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5984#gh-1254984359:-N_4gtTJ5v8eccdPwIH4:b-7krcc7) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/7a6cd2f1b79b7ab7615ced6527d7006119cf89c9/src/BloomExe/Spreadsheet/SpreadsheetIO.cs#L0)):*

*(Reviewable was [unable](https://github.com/Reviewable/Reviewable/issues/14) to map this GitHub inline comment thread to the right spot — sorry!)*

Do we still need these console messages for anything? 
</blockquote></details>

The Console.WriteLine messages aren't needed.  Did you intend to remove the exception catching altogether?  (I don't have a global view of the program in Reviewable or my head...) If they're caught in the calling method, then removing the redundant catching makes sense.

nabalone (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5984)* status: all files reviewed, 3 unresolved discussions (waiting on @hatton and @StephenMcConnel)

___
*[`src/BloomExe/Spreadsheet/SpreadsheetIO.cs` line 0 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5984#gh-1254984359:-N_5VPOzBnDIDOzI93Td:b-4h2pub) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/7a6cd2f1b79b7ab7615ced6527d7006119cf89c9/src/BloomExe/Spreadsheet/SpreadsheetIO.cs#L0)):*

The Console.WriteLine messages aren't needed.  Did you intend to remove the exception catching altogether?  (I don't have a global view of the program in Reviewable or my head...) If they're caught in the calling method, then removing the redundant catching makes sense.
</blockquote></details>

You're right, that I removed the general exception catching but they should get caught. I added catching to the spreadsheet exporter where we catch the IO exceptions

andrew-polk (changes requested):
> Reviewed 2 of 2 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5984)* status: all files reviewed, 3 unresolved discussions (waiting on @hatton, @nabalone, and @StephenMcConnel)

___
*[`src/BloomExe/Spreadsheet/SpreadsheetExporter.cs` line 87 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5984#gh-1254983545:-N_6GKa25Hgtf9_2rcmG:b5lex2p) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/7a6cd2f1b79b7ab7615ced6527d7006119cf89c9/src/BloomExe/Spreadsheet/SpreadsheetExporter.cs#L87)):*

I don't think it would be a problem (if the trailing colon doesn't bother you) to display the summary message "Export failed:" after the more explicit error messages.  It would sort of fill the slot of the "Done" message we use sometimes, but for errors.  But John is the expert in UX, not me.
</blockquote></details>

With the catch (Exception) clause, aren't we guaranteed that progress.HaveProblemsBeenReported will be true if ExportToFolder returns null?
(If not, I think we probably should make that true. Then we don't need this condition.)
___
*[`src/BloomExe/Spreadsheet/SpreadsheetIO.cs` line 0 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5984#gh-1254984359:-N_6HPwu5fjRSG-uDC4e:b-l79fa9) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/7a6cd2f1b79b7ab7615ced6527d7006119cf89c9/src/BloomExe/Spreadsheet/SpreadsheetIO.cs#L0)):*

You're right, that I removed the general exception catching but they should get caught. I'll add catching to the spreadsheet exporter where we catch the IO exceptions
</blockquote></details>

We should probably maintain the Console messages.
I don't think anyone is using (or testing it) at this point, but we do have a CLI command for spreadsheet export (see SpreadsheetExportCommand).

nabalone:
Oops wait I misread the comments, let me correct and repush

nabalone (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5984)* status: 1 of 2 files reviewed, 3 unresolved discussions (waiting on @andrew-polk, @hatton, and @StephenMcConnel)

___
*[`src/BloomExe/Spreadsheet/SpreadsheetExporter.cs` line 87 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5984#gh-1254983545:-N_F14KzDeG6VSFZPuEL:btka7zz) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/7a6cd2f1b79b7ab7615ced6527d7006119cf89c9/src/BloomExe/Spreadsheet/SpreadsheetExporter.cs#L87)):*

With the catch (Exception) clause, aren't we guaranteed that progress.HaveProblemsBeenReported will be true if ExportToFolder returns null?
(If not, I think we probably should make that true. Then we don't need this condition.)
</blockquote></details>

I was wrong to add this message - we end up on this code path when the user chooses to cancel rather than overwrite a file. All other code paths either have a problem reported or should only be used by command line
___
*[`src/BloomExe/Spreadsheet/SpreadsheetIO.cs` line 0 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5984#gh-1254984359:-N_EZWZGC4uchtxHk-3r:bp62em0) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/7a6cd2f1b79b7ab7615ced6527d7006119cf89c9/src/BloomExe/Spreadsheet/SpreadsheetIO.cs#L0)):*

We should probably maintain the Console messages.
I don't think anyone is using (or testing it) at this point, but we do have a CLI command for spreadsheet export (see SpreadsheetExportCommand).
</blockquote></details>

I replaced the console messages

andrew-polk (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5984)* status: all files reviewed, 2 unresolved discussions (waiting on @hatton and @nabalone)

___
*[`src/BloomExe/Spreadsheet/SpreadsheetExporter.cs` line 753 at r4](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5984#-N_F5T3V59gNpcrvOC0j:-N_F5T3V59gNpcrvOC0k:b44evp9) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/dfeac162e218afcf6eb30225870bcb76d607e490/src/BloomExe/Spreadsheet/SpreadsheetExporter.cs#L753)):*
> ```Smalltalk
> 				Console.WriteLine(msg);
> 				Console.WriteLine(e);
> 				progress.Message("Spreadsheet:ExportFailed", "", msg + e.Message, ProgressKind.Error);
> ```

This should be 
```
				var msg = String.Format(LocalizationManager.GetString("Spreadsheet.ExportFailed", "Export failed: {0}"), e.Message);
				Console.WriteLine(msg);
				Console.WriteLine(e);
				progress.MessageWithParams("Spreadsheet.ExportFailed", "{0} is a placeholder for the exception message",
					"Export failed: {0}", ProgressKind.Error, ex.Message);

```
or
```
				var msg = String.Format(LocalizationManager.GetString("Spreadsheet.ExportFailed", "Export failed: {0}"), e.Message);
				Console.WriteLine(msg);
				Console.WriteLine(e);
				progress.MessageWithoutLocalization(msg, ProgressKind.Error);
```

Notice "Spreadsheet.ExportFailed" (period, not colon).
Notice the placeholder in the string ({0}). This must match the string as defined in the .xlf file.

nabalone (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5984)* status: 1 of 2 files reviewed, 2 unresolved discussions (waiting on @andrew-polk and @hatton)

___
*[`src/BloomExe/Spreadsheet/SpreadsheetExporter.cs` line 753 at r4](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5984#-N_F5T3V59gNpcrvOC0j:-N_FK2OC7b_GroKj-K3g:b-273bvs) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/dfeac162e218afcf6eb30225870bcb76d607e490/src/BloomExe/Spreadsheet/SpreadsheetExporter.cs#L753)):*

This should be 
```
				var msg = String.Format(LocalizationManager.GetString("Spreadsheet.ExportFailed", "Export failed: {0}"), e.Message);
				Console.WriteLine(msg);
				Console.WriteLine(e);
				progress.MessageWithParams("Spreadsheet.ExportFailed", "{0} is a placeholder for the exception message",
					"Export failed: {0}", ProgressKind.Error, ex.Message);

```
or
```
				var msg = String.Format(LocalizationManager.GetString("Spreadsheet.ExportFailed", "Export failed: {0}"), e.Message);
				Console.WriteLine(msg);
				Console.WriteLine(e);
				progress.MessageWithoutLocalization(msg, ProgressKind.Error);
```

Notice "Spreadsheet.ExportFailed" (period, not colon).
Notice the placeholder in the string ({0}). This must match the string as defined in the .xlf file.
</blockquote></details>

Oops, sloppily messed this up in my last commit. Fixed and tested

andrew-polk (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5984)* status: all files reviewed, 1 unresolved discussion (waiting on @hatton)

andrew-polk (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5984)* status: all files reviewed, 1 unresolved discussion (waiting on @hatton and @StephenMcConnel)

andrew-polk (approved):
> Dismissed @nabalone from a discussion.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5984)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @nabalone)

PR #5988, opened 2023-07-11
Fix Javascript null reference when renaming a book (BL-12384)

StephenMcConnel:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5988)

### DIFF
--- a/src/BloomBrowserUI/collectionsTab/BookButton.tsx
+++ b/src/BloomBrowserUI/collectionsTab/BookButton.tsx
@@ -265,7 +265,11 @@ export const BookButton: React.FunctionComponent<{
                 // I tried the obvious approach of putting an onBlur in the JSX for the renameDiv,
                 // but it gets activated immediately, and I cannot figure out why.
                 renameDiv.current?.addEventListener("blur", () => {
-                    finishRename(renameDiv.current!.innerText);
+                    // As the element loses focus, renameDiv.current is set to null, so we
+                    // must use p.innerText instead of renameDiv.current.innerText.
+                    // p is still assigned to the element losing focus, so it's safe to use.
+                    // (See BL-12384.)
+                    finishRename(p.innerText);
                 });
                 renameDiv.current?.addEventListener("keypress", e => {
                     if (e.key === "Enter") {


andrew-polk (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5988)* status: all files reviewed (commit messages unreviewed), 1 unresolved discussion (waiting on @StephenMcConnel)

___
*[`src/BloomBrowserUI/collectionsTab/BookButton.tsx` line 269 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5988#-N_50jQF5nc8-RhxEaQH:-N_50jQF5nc8-RhxEaQI:bz4pddv) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/f8c5f77b54842eb0b177d438a2eb3a13a2f2ceca/src/BloomBrowserUI/collectionsTab/BookButton.tsx#L269)):*
> ```XML
>                 renameDiv.current?.addEventListener("blur", () => {
>                     // renameDiv.current!.innerText can produce a null dereference here (BL-12384)
>                     finishRename(p.innerText);
> ```

Presumably this fixes the bug (else you wouldn't have submitted the PR), but I'm not sure I understand why.
If it simply works because we are keeping around a reference to the "current" which is no longer "current", doesn't that introduce a new bug (of setting the name to wrong thing)?

Sorry; I'm probably just missing it.

StephenMcConnel (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5988)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @andrew-polk)

___
*[`src/BloomBrowserUI/collectionsTab/BookButton.tsx` line 269 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5988#-N_50jQF5nc8-RhxEaQH:-N_5uxqY7gNHSDlujn0b:bf7ynth) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/f8c5f77b54842eb0b177d438a2eb3a13a2f2ceca/src/BloomBrowserUI/collectionsTab/BookButton.tsx#L269)):*

Presumably this fixes the bug (else you wouldn't have submitted the PR), but I'm not sure I understand why.
If it simply works because we are keeping around a reference to the "current" which is no longer "current", doesn't that introduce a new bug (of setting the name to wrong thing)?

Sorry; I'm probably just missing it.
</blockquote></details>

I've expanded the comment to explain what I think is going on, which fits with my observations.  The useEffect() which this code is inside of won't be called until 1) another book is selected and 2) the user selects the Rename Book menu item.  The OnBlur operation will certainly take place before the second of these operations if not before the OnFocus of the first operation.  So the value of p has to be stable on the element we actually want to get the book's name from.  Depending on how closures work in javascript, it might be stable even without these factors in its favor.

andrew-polk (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5988)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @StephenMcConnel)

___
*[`src/BloomBrowserUI/collectionsTab/BookButton.tsx` line 269 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/5988#-N_50jQF5nc8-RhxEaQH:-N_5wfvO2-I0VMEhAE3A:b-ucfkn4) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/f8c5f77b54842eb0b177d438a2eb3a13a2f2ceca/src/BloomBrowserUI/collectionsTab/BookButton.tsx#L269)):*

I've expanded the comment to explain what I think is going on, which fits with my observations.  The useEffect() which this code is inside of won't be called until 1) another book is selected and 2) the user selects the Rename Book menu item.  The OnBlur operation will certainly take place before the second of these operations if not before the OnFocus of the first operation.  So the value of p has to be stable on the element we actually want to get the book's name from.  Depending on how closures work in javascript, it might be stable even without these factors in its favor.
</blockquote></details>

Thanks.

PR #6001, opened 2023-07-14
Fix various issues with ckeditor and markup (BL-12391, BL-12460)

andrew-polk:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6001)

### DIFF
--- a/src/BloomBrowserUI/.vscode/settings.json
+++ b/src/BloomBrowserUI/.vscode/settings.json
@@ -24,6 +24,7 @@
         "Affordance",
         "bloomlibrary",
         "contentful",
+        "Divs",
         "localizable",
         "portafino",
         "Snappable",
--- a/src/BloomBrowserUI/bookEdit/js/bloomEditing.ts
+++ b/src/BloomBrowserUI/bookEdit/js/bloomEditing.ts
@@ -49,6 +49,7 @@ import {
     getHexColorsForPalette
 } from "../../react_components/color-picking/bloomPalette";
 import { ckeditableSelector } from "../../utils/shared";
+import { EditableDivUtils } from "./editableDivUtils";
 
 // Allows toolbox code to make an element properly in the context of this iframe.
 export function makeElement(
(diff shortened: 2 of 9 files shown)


StephenMcConnel (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6001)* status: all files reviewed, 1 unresolved discussion (waiting on @andrew-polk)

___
*[`src/BloomBrowserUI/bookEdit/js/editableDivUtils.ts` line 346 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6001#-N_ZXSfZEh4K3r3SMAzC:-N_ZXSfZEh4K3r3SMAzD:bgfszkx) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/b0a1c7fd2fa9cea27337dc2a19dfbc63ff08b3cd/src/BloomBrowserUI/bookEdit/js/editableDivUtils.ts#L346)):*
> ```XML
>                 Array.from(p.childNodes)
>                     .filter(n => n.nodeName === "#text")
>                     .forEach(n => {
> ```

do we know that #text nodes are never empty?  ie, that if we find a #text node it has to be one and only one that contains the nonbreaking space?

andrew-polk (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6001)* status: 7 of 9 files reviewed, 1 unresolved discussion (waiting on @StephenMcConnel)

___
*[`src/BloomBrowserUI/bookEdit/js/editableDivUtils.ts` line 346 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6001#-N_ZXSfZEh4K3r3SMAzC:-N__3DHP1ZMhRZoJDs8M:bvysanb) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/b0a1c7fd2fa9cea27337dc2a19dfbc63ff08b3cd/src/BloomBrowserUI/bookEdit/js/editableDivUtils.ts#L346)):*

do we know that #text nodes are never empty?  ie, that if we find a #text node it has to be one and only one that contains the nonbreaking space?
</blockquote></details>

Thanks. That helped me discover a real bug.

StephenMcConnel (changes requested):
One question/comment about a comment but otherwise looks good I think.  It could be merged as is.

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6001#-:-N__sBmoE3USBlkV_u2f:btqvcf4)* status: all files reviewed, 1 unresolved discussion (waiting on @andrew-polk)

___
*[`src/BloomBrowserUI/bookEdit/toolbox/readers/leveledReader/leveledReaderToolboxTool.ts` line 36 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6001#-N__qpWD94cteZXa40LU:-N__qpWD94cteZXa40LV:b2uunzs) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/9d4335009cfd6b4a6787819f7f5936f7c700f39d/src/BloomBrowserUI/bookEdit/toolbox/readers/leveledReader/leveledReaderToolboxTool.ts#L36)):*
> ```XML
>         // This gets called for every tool, whether or not that tool is open.
>         // decodableReaderToolboxTool.ts' configureElements() calls this.setupReaderKeyAndFocusHandlers(container)
>         // which is also needed for leveled reader.
> ```

So if I understand this comment correctly, the call to this.setupReaderKeyAndFocusHandlers(container) in the decodableReaderToolboxTool method covers the need for it to be called here since it gets called even when decodable reader isn't open but leveled reader is open.  It took me a minute to decipher the comment, so maybe the English could be clearer.  But I'm not sure what to suggest as an improvement.

StephenMcConnel (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6001)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @andrew-polk)

PR #6002, opened 2023-07-17
Fix error report to upload bad book, not previous selection (BL-12470)

StephenMcConnel:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6002)

### DIFF
--- a/src/BloomExe/web/controllers/CollectionApi.cs
+++ b/src/BloomExe/web/controllers/CollectionApi.cs
@@ -79,18 +79,34 @@ namespace Bloom.web.controllers
 						// This first method does minimal processing to come up with the right collection and BookInfo object
 						// without actually loading all the files. We need the title for the Performance Measurement and later
 						// we'll use the BookInfo object to get the fully updated book.
-						var newBookInfo = GetBookInfoFromPost(request);
-						var titleString = newBookInfo.QuickTitleUserDisplay;
-						using (PerformanceMeasurement.Global?.Measure("select book", titleString))
+						BookInfo newBookInfo = null;
+						try
 						{
-							// We could just put the PerformanceMeasurement in the CollectionModel.SelectBook() method,
-							// but this GetUpdatedBookObjectFromBookInfo() actually does a non-trivial amount of work,
-							// because it asks the CollectionModel to update the book files (including BringBookUpToDate).
-							var book = GetUpdatedBookObjectFromBookInfo(newBookInfo);
-							if (book.FolderPath != _bookSelection?.CurrentSelection?.FolderPath)
+							newBookInfo = GetBookInfoFromPost(request);
+							var titleString = newBookInfo.QuickTitleUserDisplay;
+							using (PerformanceMeasurement.Global?.Measure("select book", titleString))
 							{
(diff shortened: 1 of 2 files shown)


andrew-polk (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6002)* status: all files reviewed, 1 unresolved discussion (waiting on @StephenMcConnel)

___
*[`src/BloomExe/web/controllers/CollectionApi.cs` line 103 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6002#-N__bKc24wMz-LitiAHy:-N__bKc24wMz-LitiAHz:b-c78ou5) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/eb3dd03514f48d7a5672b911a21f888a2ac2eb92/src/BloomExe/web/controllers/CollectionApi.cs#L103)):*
> ```Smalltalk
> 							if (newBookInfo?.FolderPath != null)
> 							{
> 								var ex = new Exception("Error selecting book: " + newBookInfo.FolderPath, e);
> ```

Would be better to throw a new ApplicationException?
Or you were just throwing Exception because we caught Exception?

StephenMcConnel (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6002)* status: all files reviewed, 1 unresolved discussion (waiting on @andrew-polk)

___
*[`src/BloomExe/web/controllers/CollectionApi.cs` line 103 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6002#-N__bKc24wMz-LitiAHy:-N__udRj2MqZX77o4lK4:b-wtpzg6) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/eb3dd03514f48d7a5672b911a21f888a2ac2eb92/src/BloomExe/web/controllers/CollectionApi.cs#L103)):*

Would be better to throw a new ApplicationException?
Or you were just throwing Exception because we caught Exception?
</blockquote></details>

Microsoft documentation for ApplicationException states "You should derive custom exceptions from the [Exception](https://learn.microsoft.com/en-us/dotnet/api/system.exception?view=netframework-4.7.2) class rather than the [ApplicationException](https://learn.microsoft.com/en-us/dotnet/api/system.applicationexception?view=netframework-4.7.2) class.  You should not throw an [ApplicationException](https://learn.microsoft.com/en-us/dotnet/api/system.applicationexception?view=netframework-4.7.2) exception in your code, and you should not catch an [ApplicationException](https://learn.microsoft.com/en-us/dotnet/api/system.applicationexception?view=netframework-4.7.2) exception unless you intend to re-throw the original exception."  I'm not sure I understand the reasoning, but it might be better to stick with the plain Exception.
My original version of this code defined a custom exception class based off Exception, but I decided that we don't really need a custom class here since I can added the necessary extra information in the Message and Data fields.

andrew-polk (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6002)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @StephenMcConnel)

PR #6003, opened 2023-07-17
BL-12540 allow pasting of text formatted in small caps

hatton:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6003)

### DIFF
--- a/src/BloomBrowserUI/lib/ckeditor/config.js
+++ b/src/BloomBrowserUI/lib/ckeditor/config.js
@@ -101,6 +101,9 @@ CKEDITOR.editorConfig = function(config) {
     // The {font-weight} annotation says that a <b> element's font-weight style may be pasted.
     // Code in our paste handler deals with this so it doesn't end up in our documents.
     // span{font-variant} allows pasting of text that has been formatted as small caps.
+    //
+    // Shown by experiment: when listing something like span{font-variant}, all other
+    //    attributes and style rules will be removed.
     config.pasteFilter =
         "p br em i strong sup u; b{font-weight}; a[!href]; span{font-variant};";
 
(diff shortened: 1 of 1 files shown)


andrew-polk (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6003)* status: all files reviewed, 1 unresolved discussion (waiting on @hatton)

___
*[`src/BloomBrowserUI/lib/ckeditor/config.js` line 105 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6003#-N__Safn0i5EQjk_T4ht:-N__Safn0i5EQjk_T4hu:bfqbqlw) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/02055da794a5ecc6c597d1d324336fb491e61ff6/src/BloomBrowserUI/lib/ckeditor/config.js#L105)):*
Note the comment above:
> BL-4775: Removed span from this, so that you can't paste spans

which gives me pause for making this change.

By adding `span{font-variant};`, will it strip all the other attributes and styles?
I think we especially need to ensure it strips `id`.

hatton:
I do not see an `id` in the result.

src/BloomBrowserUI/lib/ckeditor/config.js
```diff
@@ -100,7 +100,9 @@ CKEDITOR.editorConfig = function(config) {
     // Therefore for now we're limiting pasting to things that a translator could also do.
     // The {font-weight} annotation says that a <b> element's font-weight style may be pasted.
     // Code in our paste handler deals with this so it doesn't end up in our documents.
-    config.pasteFilter = "p br em i strong sup u; b{font-weight}; a[!href];";
+    // span{font-variant} allows pasting of text that has been formatted as small caps.
```
andrew-polk, line 103:
```suggestion
    // span{font-variant} allows pasting of text that has been formatted as small caps.
    //
    // Shown by experiment: when listing something like span{font-variant}, all other
    //    attributes and style rules will be removed.
```

andrew-polk (changes requested):

PR #6017, opened 2023-07-19
Bl 12406 Don't show error Dialog if mic access is off 

nabalone:
Show an popup message but not the error dialog if we can't access microphone.

This PR depends on Palaso changes: https://github.com/sillsdev/libpalaso/pull/1280

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6017)

### DIFF
--- a/DistFiles/localization/en/BloomLowPriority.xlf
+++ b/DistFiles/localization/en/BloomLowPriority.xlf
@@ -172,6 +172,10 @@
             <trans-unit id="Activity.Editing.MultipleChoice.CorrectAnswer">
                 <source xml:lang="en">Put the correct answer in this button. Bloom will randomize the order later.</source>
             </trans-unit>
+          <trans-unit id="EditTab.Toolbox.TalkingBookTool.MicrophoneAccessProblem">
+            <source xml:lang="en">Bloom was not able to access a microphone.</source>
+            <note>ID: EditTab.Toolbox.TalkingBookTool.MicrophoneAccessProblem</note>
+          </trans-unit>
         </body>
     </file>
 </xliff>
--- a/src/BloomExe/Edit/AudioRecording.cs
+++ b/src/BloomExe/Edit/AudioRecording.cs
@@ -241,16 +241,12 @@ namespace Bloom.Edit
 					Recorder.BeginMonitoring(catchAndReportExceptions: false);
 					_monitoringAudio = true;
 				}
-				catch (NAudio.MmException)
+				catch (Exception e)
 				{
+					Logger.WriteError("Could not begin monitoring microphone", e);
 					var msg = LocalizationManager.GetString("EditTab.Toolbox.TalkingBookTool.MicrophoneAccessProblem",
 						"Bloom was not able to access a microphone.");
 					_webSocketServer.SendString(kWebsocketContext, "micError", msg);
-				}
-				catch (Exception e)
-				{
-					ErrorReport.NotifyUserOfProblem(new ShowOncePerSessionBasedOnExactMessagePolicy(),
-						e, "There was a problem starting up volume monitoring.");
 				}
 			}
 		}
@@ -561,13 +557,16 @@ namespace Bloom.Edit
 			try
 			{
 				Recorder.BeginRecording(PathToTemporaryWav);
-				_webSocketServer.SendString(kWebsocketContext, "recordStartStatus", "success");
+				_webSocketServer.SendString(kWebsocketContext, "recordStartSucceeded", "success");
 			}
-			catch (InvalidOperationException)
+			catch (InvalidOperationException ex)
 			{
 				// Likely a case of BL-7568, which as far as we can figure isn't Bloom's fault.
 				// Show a friendly message in the TalkingBook toolbox.
-				_webSocketServer.SendString(kWebsocketContext, "recordStartStatus", "failure");
+				Logger.WriteError("Could not begin recording", ex);
+				var msg = LocalizationManager.GetString("EditTab.Toolbox.TalkingBook.MicrophoneProblem",
+					"Bloom is having problems connecting to your microphone. Please restart your computer and try again.");
+				_webSocketServer.SendString(kWebsocketContext, "recordStartFailed", msg);
 			}
 		}
 
--- a/src/BloomExe/Edit/AudioRecording.cs
+++ b/src/BloomExe/Edit/AudioRecording.cs
@@ -557,7 +557,6 @@ namespace Bloom.Edit
 			try
 			{
 				Recorder.BeginRecording(PathToTemporaryWav);
-				_webSocketServer.SendString(kWebsocketContext, "recordStartSucceeded", "success");
 			}
 			catch (InvalidOperationException ex)
 			{
@@ -566,7 +565,7 @@ namespace Bloom.Edit
 				Logger.WriteError("Could not begin recording", ex);
 				var msg = LocalizationManager.GetString("EditTab.Toolbox.TalkingBook.MicrophoneProblem",
 					"Bloom is having problems connecting to your microphone. Please restart your computer and try again.");
-				_webSocketServer.SendString(kWebsocketContext, "recordStartFailed", msg);
+				_webSocketServer.SendString(kWebsocketContext, "micError", msg);
 			}
 		}
 
--- a/src/BloomExe/Edit/AudioRecording.cs
+++ b/src/BloomExe/Edit/AudioRecording.cs
@@ -236,8 +236,22 @@ namespace Bloom.Edit
 			}
 			if (RecordingDevice != null)
 			{
-				Recorder.BeginMonitoring();
-				_monitoringAudio = true;
+				try
+				{
+					Recorder.BeginMonitoring(catchAndReportExceptions: false);
+					_monitoringAudio = true;
+				}
+				catch (NAudio.MmException)
+				{
+					var msg = LocalizationManager.GetString("EditTab.Toolbox.TalkingBookTool.MicrophoneAccessProblem",
+						"Bloom was not able to access a microphone.");
+					_webSocketServer.SendString(kWebsocketContext, "micError", msg);
+				}
+				catch (Exception e)
+				{
+					ErrorReport.NotifyUserOfProblem(new ShowOncePerSessionBasedOnExactMessagePolicy(),
+						e, "There was a problem starting up volume monitoring.");
+				}
 			}
 		}
 
--- a/src/BloomExe/Edit/AudioRecording.cs
+++ b/src/BloomExe/Edit/AudioRecording.cs
@@ -246,7 +246,7 @@ namespace Bloom.Edit
 					Logger.WriteError("Could not begin monitoring microphone", e);
 					var msg = LocalizationManager.GetString("EditTab.Toolbox.TalkingBookTool.MicrophoneAccessProblem",
 						"Bloom was not able to access a microphone.");
-					_webSocketServer.SendString(kWebsocketContext, "micError", msg);
+					_webSocketServer.SendString(kWebsocketContext, "monitoringStartError", msg);
 				}
 			}
 		}
@@ -565,7 +565,7 @@ namespace Bloom.Edit
 				Logger.WriteError("Could not begin recording", ex);
 				var msg = LocalizationManager.GetString("EditTab.Toolbox.TalkingBook.MicrophoneProblem",
 					"Bloom is having problems connecting to your microphone. Please restart your computer and try again.");
-				_webSocketServer.SendString(kWebsocketContext, "micError", msg);
+				_webSocketServer.SendString(kWebsocketContext, "recordingStartError", msg);
 			}
 		}
 
--- a/src/BloomBrowserUI/bookEdit/toolbox/talkingBook/audioRecording.less
+++ b/src/BloomBrowserUI/bookEdit/toolbox/talkingBook/audioRecording.less
@@ -7,7 +7,6 @@
 @sectionVerticalGap: 25px;
 @buttonColumnWidth: 40px;
 @splitAnimationTime: 300ms;
-@micProblemColor: @bloom-warning;
 @textOnLightBackground: black;
 @disablingOverlayZindex: 1001; // higher than #audio-devlist
 
@@ -464,18 +463,6 @@ body .cursor-progress {
 .toast-toolbox-bottom {
     left: 32px;
     bottom: 46px;
-}
-
-#mic-problem-message {
-    &.initiallyHidden {
-        display: none;
-    }
-
-    background-color: @micProblemColor;
-    color: @textOnLightBackground;
-    border-radius: 6px;
-    font-size: medium;
-    padding: 6px;
 }
 
 #audio-playbackOrderControl-wrapper {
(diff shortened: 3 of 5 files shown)


DistFiles/localization/en/BloomLowPriority.xlf
```diff
@@ -172,6 +172,10 @@
             <trans-unit id="Activity.Editing.MultipleChoice.CorrectAnswer">
                 <source xml:lang="en">Put the correct answer in this button. Bloom will randomize the order later.</source>
             </trans-unit>
+          <trans-unit id="EditTab.Toolbox.TalkingBookTool.MicrophoneAccessProblem">
+            <source xml:lang="en">Bloom was not able to access a microphone.</source>
```
nabalone, line 176: Do we want to translate this? BloomLowPriority?

src/BloomExe/Edit/AudioRecording.cs
```diff
@@ -236,8 +236,22 @@ public void BeginMonitoring()
...
+				}
+				catch (NAudio.MmException)
+				{
+					var msg = LocalizationManager.GetString("EditTab.Toolbox.TalkingBookTool.MicrophoneAccessProblem",
+						"Bloom was not able to access a microphone.");
+					_webSocketServer.SendString(kWebsocketContext, "micError", msg);
+				}
+				catch (Exception e)
+				{
+					ErrorReport.NotifyUserOfProblem(new ShowOncePerSessionBasedOnExactMessagePolicy(),
```
nabalone, line 252: I did this to retain what the original behavior would be if a different exception is throw. But not sure what we want the behavior to be here

nabalone:
Also just noting that there is an unrelated similar error message that occurs when we fail to start recording 
Screenshot 2023-07-18 192036

nabalone:
Also, this PR won't work without the Palaso changes in https://github.com/sillsdev/libpalaso/pull/1280

andrew-polk (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6017)* status: all files reviewed, 2 unresolved discussions (waiting on @nabalone)

___
*[`DistFiles/localization/en/BloomLowPriority.xlf` line 176 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6017#gh-1268721246:-N_kM0Wt1EOXG3wS4p4O:bbe1j0k) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/55ddf8ed8c93bc916c3a03fdb7cab5a56dc13891/DistFiles/localization/en/BloomLowPriority.xlf#L176)):*

Do we want to translate this? BloomLowPriority?
</blockquote></details>

BloomLowPriority seems good.
___
*[`src/BloomExe/Edit/AudioRecording.cs` line 248 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6017#-N_kNZiq3wFjNfjR2_Zs:-N_kNZiq3wFjNfjR2_Zt:b-mpb8a0) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/55ddf8ed8c93bc916c3a03fdb7cab5a56dc13891/src/BloomExe/Edit/AudioRecording.cs#L248)):*
> ```Smalltalk
> 					var msg = LocalizationManager.GetString("EditTab.Toolbox.TalkingBookTool.MicrophoneAccessProblem",
> 						"Bloom was not able to access a microphone.");
> 					_webSocketServer.SendString(kWebsocketContext, "micError", msg);
> ```

Question about UX in card.
___
*[`src/BloomExe/Edit/AudioRecording.cs` line 252 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6017#gh-1268722530:-N_kNj2h-uK3jM6WFioD:b-pfcpe5) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/55ddf8ed8c93bc916c3a03fdb7cab5a56dc13891/src/BloomExe/Edit/AudioRecording.cs#L252)):*

I did this to retain what the original behavior would be if a different exception is throw. But not sure what we want the behavior to be here
</blockquote></details>

Hm. I would lean toward just showing the same message for all exceptions.
1. consistent ux
2. we don't know that the error has anything to do with volume monitoring

Is there any advantage to trying to maintain the original behavior for the unexpected exception scenario?

nabalone (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6017)* status: 1 of 5 files reviewed, 3 unresolved discussions (waiting on @andrew-polk)

___
*[`src/BloomBrowserUI/bookEdit/toolbox/talkingBook/audioRecording.ts` line 1127 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6017#-N_orzcg2ys2FfGHPY-Y:-N_orzcg2ys2FfGHPY-Z:b-xohlyr) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/5f0e48987ef67e62f888b20b6280262899dabb1c/src/BloomBrowserUI/bookEdit/toolbox/talkingBook/audioRecording.ts#L1127)):*
> ```XML
>     private addAudioRecordStartListener(): void {
>         WebSocketManager.addListener(kWebsocketContext, e => {
>             // if the recording did start, we will get a message with e.id === "recordStartSucceeded"
> ```

We were previously sending a success message to the front end upon success and and doing nothing with it. See comment above this listener method. Do we want to keep that? For debugging? I changed the event ids so that the message could contain the localized error message
___
*[`src/BloomExe/Edit/AudioRecording.cs` line 248 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6017#-N_kNZiq3wFjNfjR2_Zs:-N_ormSxCE8jUABag_6N:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/55ddf8ed8c93bc916c3a03fdb7cab5a56dc13891/src/BloomExe/Edit/AudioRecording.cs#L248)):*

Question about UX in card.
</blockquote></details>

Done.

andrew-polk (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6017)* status: all files reviewed, 1 unresolved discussion (waiting on @nabalone)

___
*[`src/BloomBrowserUI/bookEdit/toolbox/talkingBook/audioRecording.ts` line 1127 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6017#-N_orzcg2ys2FfGHPY-Y:-N_tE5SY79ODIpV7KZ3r:b-2uj3h0) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/5f0e48987ef67e62f888b20b6280262899dabb1c/src/BloomBrowserUI/bookEdit/toolbox/talkingBook/audioRecording.ts#L1127)):*

We were previously sending a success message to the front end upon success and and doing nothing with it. See comment above this listener method. Do we want to keep that? For debugging? I changed the event ids so that the message could contain the localized error message
</blockquote></details>

I don't think I would keep it now that it is a different ID.
Especially because I think we can simplify further now:
I think we can have one listener for both recordStartFailed and micError since you are sending the message.
Setting this.recording = false shouldn't have any negative effect if it is already false (in the micError case).
Disabling the record button seems like the right thing to do in both cases.

But I could be missing something. What do you think?

nabalone (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6017)* status: 3 of 5 files reviewed, 1 unresolved discussion (waiting on @andrew-polk)

___
*[`src/BloomBrowserUI/bookEdit/toolbox/talkingBook/audioRecording.ts` line 1127 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6017#-N_orzcg2ys2FfGHPY-Y:-N_tVdDW3hzuTAlWX60F:b-wn0hs3) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/5f0e48987ef67e62f888b20b6280262899dabb1c/src/BloomBrowserUI/bookEdit/toolbox/talkingBook/audioRecording.ts#L1127)):*

I don't think I would keep it now that it is a different ID.
Especially because I think we can simplify further now:
I think we can have one listener for both recordStartFailed and micError since you are sending the message.
Setting this.recording = false shouldn't have any negative effect if it is already false (in the micError case).
Disabling the record button seems like the right thing to do in both cases.

But I could be missing something. What do you think?
</blockquote></details>

Sounds good, done. 

Also, this reminds me of a question - the reporting user wrote "As I'm recording through my laptop using talkback and in order to stop outside noise I'm turning off my microphone." Do they have multiple sound inputs from which to record? And do we need to make sure it behaves properly with this setup (e.g. if one causes an error but the other is ok)? On my computer, when I plug in an external mic it looks like my OS is handling which sound input is getting used

nabalone (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6017)* status: 3 of 5 files reviewed, 1 unresolved discussion (waiting on @andrew-polk)

___
*[`src/BloomBrowserUI/bookEdit/toolbox/talkingBook/audioRecording.ts` line 1127 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6017#-N_orzcg2ys2FfGHPY-Y:-N_tnlRqDJZ4XUFTlF9P:b-7y9poe) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/5f0e48987ef67e62f888b20b6280262899dabb1c/src/BloomBrowserUI/bookEdit/toolbox/talkingBook/audioRecording.ts#L1127)):*

Sounds good, done. 

Also, this reminds me of a question - the reporting user wrote "As I'm recording through my laptop using talkback and in order to stop outside noise I'm turning off my microphone." Do they have multiple sound inputs from which to record? And do we need to make sure it behaves properly with this setup (e.g. if one causes an error but the other is ok)? On my computer, when I plug in an external mic it looks like my OS is handling which sound input is getting used
</blockquote></details>

Also just noting that now that we are setting Record Status to Disabled when we don't have permission to access the mic, then after the user fixes the permissions they need to close and reopen at least the talking book tool. This seems reasonable to me and better than the alternative, where if we don't set Record Status to Disabled, the user can still hit the record button and generate another error opening the error dialog.

andrew-polk:
*[`src/BloomBrowserUI/bookEdit/toolbox/talkingBook/audioRecording.ts` line 1127 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6017#-N_orzcg2ys2FfGHPY-Y:-N_toLSz6WEDs4Lv3uqP:buxbu87) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/5f0e48987ef67e62f888b20b6280262899dabb1c/src/BloomBrowserUI/bookEdit/toolbox/talkingBook/audioRecording.ts#L1127)):*

Also just noting that now that we are setting Record Status to Disabled when we don't have permission to access the mic, then after the user fixes the permissions they need to close and reopen at least the talking book tool. This seems reasonable to me and better than the alternative, where if we don't set Record Status to Disabled, the user can still hit the record button and generate another error opening the error dialog.
</blockquote></details>

Looks good. Thanks.

Yes; generally, the UI is designed to handle switching microphones. The icon should change, etc. But yes, it would be good to try to verify that an error on one mic does not prevent subsequent recording on another.
It would be fine even to simulate the test via code if needed.

andrew-polk:
*[`src/BloomBrowserUI/bookEdit/toolbox/talkingBook/audioRecording.ts` line 1127 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6017#-N_orzcg2ys2FfGHPY-Y:-N_totTX2clBwsgUena3:b-2our6m) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/5f0e48987ef67e62f888b20b6280262899dabb1c/src/BloomBrowserUI/bookEdit/toolbox/talkingBook/audioRecording.ts#L1127)):*
> This seems reasonable to me and better than the alternative, where if we don't set Record Status to Disabled, the user can still hit the record button and generate another error opening the error dialog.

Agree.

andrew-polk (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6017)* status: all files reviewed, 1 unresolved discussion (waiting on @nabalone)

nabalone (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6017)* status: 3 of 5 files reviewed, 1 unresolved discussion (waiting on @andrew-polk)

___
*[`src/BloomBrowserUI/bookEdit/toolbox/talkingBook/audioRecording.ts` line 1127 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6017#-N_orzcg2ys2FfGHPY-Y:-NaDgIB6EBnnIgT6MT8q:b-vkqc8p) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/5f0e48987ef67e62f888b20b6280262899dabb1c/src/BloomBrowserUI/bookEdit/toolbox/talkingBook/audioRecording.ts#L1127)):*

> This seems reasonable to me and better than the alternative, where if we don't set Record Status to Disabled, the user can still hit the record button and generate another error opening the error dialog.

Agree.
</blockquote></details>

- Even if we were to disable recording status, it looks like a race condition is often re-enabling it after we disable it, so it would be more work to figure out how to consistently disable it
- I have not disabled it also because when you switch microphones, we begin monitoring the first while switching to the second. We don't want this to disable the Record button, in case the first mic is inaccessible and the second mic works
- Since we are not disabling the Record button here, the use can still hit the Record button after we failed to begin monitoring on their microphone, resulting in another error dialog popping up
- Furthermore, failure to begin recording (i.e. on an inaccessible microphone) does disable recording, so if a user hits Record while on an inaccessible microphone, they will then need to close and reopen the talking book tool before they would be able to record on another microphone

andrew-polk (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6017)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @nabalone)

___
*[`src/BloomBrowserUI/bookEdit/toolbox/talkingBook/audioRecording.ts` line 1127 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6017#-N_orzcg2ys2FfGHPY-Y:-NaE-09_Ax-7QfqCfuHp:b66b5b4) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/5f0e48987ef67e62f888b20b6280262899dabb1c/src/BloomBrowserUI/bookEdit/toolbox/talkingBook/audioRecording.ts#L1127)):*

- Even if we were to disable recording status, it looks like a race condition is often re-enabling it after we disable it, so it would be more work to figure out how to consistently disable it
- I have not disabled it also because when you switch microphones, we begin monitoring the first while switching to the second. We don't want this to disable the Record button, in case the first mic is inaccessible and the second mic works
- Since we are not disabling the Record button here, the use can still hit the Record button after we failed to begin monitoring on their microphone, resulting in another error dialog popping up
- Furthermore, failure to begin recording (i.e. on an inaccessible microphone) does disable recording, so if a user hits Record while on an inaccessible microphone, they will then need to close and reopen the talking book tool before they would be able to record on another microphone
</blockquote></details>

Let's live with it for now.

PR #6030, opened 2023-08-01
Fix collection settings for language location in HarvesterMode (BL-12583)

StephenMcConnel:
The previous attempt didn't actually work at all since fresh xmatter erased the needed information when publishing.

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6030)

### DIFF
--- a/src/BloomExe/Book/BookData.cs
+++ b/src/BloomExe/Book/BookData.cs
@@ -1020,6 +1020,45 @@ namespace Bloom.Book
 
 				location = TrimEnd(location, separator);
 
+				if (Program.RunningHarvesterMode && String.IsNullOrEmpty(location))
+				{
+					var nodesWithValue = _dom.RawDom.SafeSelectNodes("//*[@data-library='languageLocation']");
+					if (nodesWithValue != null && nodesWithValue.Count != 0)
+					{
+						var node = nodesWithValue[0];
+						location = node.InnerText.Trim();
+						Console.WriteLine("Language location from data-library node: \"{0}\"", location);
+						// We have the dataset value, but we need to preserve something in the actual
+						// collection settings.  Being in harvester mode, these are not saved to the file.
+						var pieces = location.Split(new[] {separator}, StringSplitOptions.RemoveEmptyEntries);
+						switch (pieces.Length)
+						{
+							case 0:
+								// We have no data, so we'll just leave the empty collection settings alone.
+								break;
+							case 1:
+								collectionSettings.Country = pieces[0];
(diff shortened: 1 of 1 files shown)


andrew-polk (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6030)* status: all files reviewed, 1 unresolved discussion (waiting on @StephenMcConnel)

___
*[`src/BloomExe/Book/BookData.cs` line 1060 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6030#-NanP5VLEEMDqMcjCqLf:-NanP5VLEEMDqMcjCqLg:b-hvukiw) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/69dcf82ad836a9b4c977c70fd3ca3d4bc01f7b1b/src/BloomExe/Book/BookData.cs#L1060)):*
> ```Smalltalk
> 						}
> 					}
> 				}
> ```

I don't think I like this approach, but I'm not sure about the right approach. 
I would like to run it by JH in the morning. He usually sees things more clearly at the forest level.

hatton (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6030)* status: all files reviewed, 1 unresolved discussion (waiting on @andrew-polk)

___
*[`src/BloomExe/Book/BookData.cs` line 1060 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6030#-NanP5VLEEMDqMcjCqLf:-NaqIYHyEfjd6x3ByYTp:b-6i0ye4) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/69dcf82ad836a9b4c977c70fd3ca3d4bc01f7b1b/src/BloomExe/Book/BookData.cs#L1060)):*

I don't think I like this approach, but I'm not sure about the right approach. 
I would like to run it by JH in the morning. He usually sees things more clearly at the forest level.
</blockquote></details>

Probably not what you mean, but this would be easier to read:
```c#
collectionSettings.District = pieces.Length > 0 ? pieces[0] : null; 
collectionSettings.Province = pieces.Length > 1 ? pieces[1] : null; 
collectionSettings.Country = pieces.Length > 2 ? string.Join(separator, pieces.Skip(2)) : null;
```

StephenMcConnel:
We decided that we do need to upload the .bloomCollection file (sanitized and with a different name), and that the fix for books without the uploaded settings file should be in harvester as well.

PR #6032, opened 2023-08-01
Prevent imposing transparency on non-greyscale PNG images (BL-12570)

StephenMcConnel:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6032)

### DIFF
--- a/src/BloomExe/Book/BookCompressor.cs
+++ b/src/BloomExe/Book/BookCompressor.cs
@@ -307,7 +307,7 @@ namespace Bloom.Book
 		/// will often be transmitted over slow network connections.  So rather than merely zipping up
 		/// an image file, we set its dimensions to within our desired limit (currently 600x600px) and
 		/// generate the bytes in the desired format.  If the original image is already small enough, we
-		/// retain its dimensions.  We also make png images have transparent backgrounds if requested so
+		/// retain its dimensions.  We also make two-color or graysceale png images have transparent backgrounds if requested so
 		/// that they will work for cover pages.  If transparency is not needed, the original image file
 		/// bytes are returned if that results in a smaller final image file.
 		/// </summary>
@@ -368,8 +368,9 @@ namespace Bloom.Book
 					// See https://issues.bloomlibrary.org/youtrack/issue/BH-5812.
 					imagePixelFormat = EnsureValidPixelFormat(imagePixelFormat);
 
-					var needTransparencyConversion =
-						!appearsToBeJpeg && needsTransparentBackground && !ImageUtils.HasTransparency(image);
+					// Making white pixels transparent for non-two-tone PNG files is not a good idea.  See BL-12570.
+					var needTransparencyConversion = needsTransparentBackground && !appearsToBeJpeg &&
+						ImageUtils.ShouldMakeBackgroundTransparent(originalImage);
 
 					using (var newImage = new Bitmap(newWidth, newHeight, imagePixelFormat))
 					{
--- a/src/BloomExe/ImageProcessing/ImageUtils.cs
+++ b/src/BloomExe/ImageProcessing/ImageUtils.cs
@@ -68,6 +68,13 @@ namespace Bloom.ImageProcessing
 			return false;
 		}
 
+		private class ColorInfo
+		{
+			public Color color;
+			public bool isGrayish;
+			public bool isNearWhite;
+		}
+
 		/// <summary>
 		/// Check whether we should try to make the background of this image transparent.
 		/// Return true only if this is a two-color image with one of the colors being white.
@@ -81,29 +88,27 @@ namespace Bloom.ImageProcessing
 			// JPEG pictures generally never meet that criteria and cannot be made transparent anyway.
 			if (!AppearsToBePng(imageInfo))
 				return false;
+			var colors = new List<ColorInfo>();
 			if ((imageInfo.Image.PixelFormat & PixelFormat.Indexed) == PixelFormat.Indexed)
 			{
 				var palette = imageInfo.Image.Palette;
 				if (palette != null && palette.Entries != null)
 				{
-					// If only two colors are used, assume black and white line art that needs to have
-					// white made transparent.
-					if (palette.Entries.Length == 2)
+					bool whiteFound = false;
+					foreach (var color in palette.Entries)
 					{
-						var whiteFound = IsNearWhite(palette.Entries[0]) || IsNearWhite(palette.Entries[1]);
-						return whiteFound;
+						if (color.A < 255)
+							return false;   // already have transparent pixels
+						if (!IsThisColorForLineDrawing(color, colors, ref whiteFound))
+							return false;   // have a 3rd distinct non-gray color
 					}
-					else
-					{
-						return false;
-					}
+					return colors.Count == 2 && whiteFound;
 				}
 			}
 			// Harder to check if not indexed...
 			if (imageInfo.Image is Bitmap bitmapImage)
 			{
-				var color1 = new Color();
-				var color2 = new Color();
+				var whiteFound = false;
 				// Yes, this is as expensive as it looks.  But we only sample 100 pixels
 				// spread through the picture, stopping as soon as we hit either a
 				// transparent pixel or a 3rd distinct non-gray color.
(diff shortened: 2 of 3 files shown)


andrew-polk (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6032)* status: all files reviewed, 4 unresolved discussions (waiting on @StephenMcConnel)

___
*[`src/BloomExe/ImageProcessing/ImageUtils.cs` line 104 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6032#-NanLG9Q4FQP-1l5Vri6:-NanLG9Q4FQP-1l5Vri7:b-p3kv5b) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/369c42e2322ff26ff50f20ca52142a8fe3942016/src/BloomExe/ImageProcessing/ImageUtils.cs#L104)):*
> // We allow grey-scale images to be made transparent below, so why not here?

This comment makes sense in terms of the change you are making, but I don't think we want it in the code.
In other words, I think our future selves will see the comment 6 months from now and wonder why it is there.
___
*[`src/BloomExe/ImageProcessing/ImageUtils.cs` line 109 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6032#-NanMeOR53ZTpIRdtLpG:-NanMeOR53ZTpIRdtLpH:b8u4y5i) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/369c42e2322ff26ff50f20ca52142a8fe3942016/src/BloomExe/ImageProcessing/ImageUtils.cs#L109)):*
> 							if (color.A < 255 || (!IsGrayish(color) && !whitish))
>								return false;   // either transparent already or not black and white/greyscale

I don't think this is the algorithm we want.
For example, consider line art which is blue or dark green on white.
This will not add transparency.

(One could argue that we don't add it now based on current code, but since you're making this indexed path more like the other path, we might as well go all the way.)

My experiment just now in Bloom 5.4 shows that blue/white line art does get transparency. (I guess my test png must not have exactly 2 colors...)

If I'm right, we should change the new comment in BookCompressor.
___
*[`src/BloomExe/ImageProcessing/ImageUtils.cs` line 191 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6032#-NanO3DS7KhnQVSK9H3Z:-NanO3DS7KhnQVSK9H3_:b-km219t) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/369c42e2322ff26ff50f20ca52142a8fe3942016/src/BloomExe/ImageProcessing/ImageUtils.cs#L191)):*
> ```Smalltalk
> 				// (Our code wouldn't make anything transparent anyway, or would make everything
> 				// transparent.)
> 				return colorCount == 2 && whiteFound && nonWhiteFound;
> ```

I don't think we want to add this check for non-white.
Consider the case where we happen to miss all the non-white pixels in our 100-pixel sample.
___
*[`src/BloomExe/ImageProcessing/ImageUtils.cs` line 194 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6032#-NanORGo6edKf28uSR7_:-NanORGo6edKf28uSR7a:b-vm305z) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/369c42e2322ff26ff50f20ca52142a8fe3942016/src/BloomExe/ImageProcessing/ImageUtils.cs#L194)):*
> ```Smalltalk
> 			}
> 			// we can't tell, so err on the side of caution.
> 			return false;
> ```

Note to self, after the above discussions are resolved, I need to do a full review again of the changes.

StephenMcConnel (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6032)* status: all files reviewed, 4 unresolved discussions (waiting on @andrew-polk)

___
*[`src/BloomExe/ImageProcessing/ImageUtils.cs` line 104 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6032#-NanLG9Q4FQP-1l5Vri6:-NarKzes5Mzw-M_RZi58:b3c7sdr) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/369c42e2322ff26ff50f20ca52142a8fe3942016/src/BloomExe/ImageProcessing/ImageUtils.cs#L104)):*

> // We allow grey-scale images to be made transparent below, so why not here?

This comment makes sense in terms of the change you are making, but I don't think we want it in the code.
In other words, I think our future selves will see the comment 6 months from now and wonder why it is there.
</blockquote></details>

I removed the comment and rewrote this Indexed palette processing to use the same analysis/method as the bitmap sampling code.
___
*[`src/BloomExe/ImageProcessing/ImageUtils.cs` line 109 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6032#-NanMeOR53ZTpIRdtLpG:-NarOd-FBMq3xcewVQxs:b2sw80f) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/369c42e2322ff26ff50f20ca52142a8fe3942016/src/BloomExe/ImageProcessing/ImageUtils.cs#L109)):*

> 							if (color.A < 255 || (!IsGrayish(color) && !whitish))
>								return false;   // either transparent already or not black and white/greyscale

I don't think this is the algorithm we want.
For example, consider line art which is blue or dark green on white.
This will not add transparency.

(One could argue that we don't add it now based on current code, but since you're making this indexed path more like the other path, we might as well go all the way.)

My experiment just now in Bloom 5.4 shows that blue/white line art does get transparency. (I guess my test png must not have exactly 2 colors...)

If I'm right, we should change the new comment in BookCompressor.
</blockquote></details>

Detecting "line art" per se is very difficult.  Mono-color line art (only two colors, one of them white) can be assumed from having only two colors.  It's a bit of a jump to assume line-art for grayscale, but we've been willing to do that for non-indexed images.  Trying to detect, for example, shades of blue or shades of green (or worse, shades of purple) for indicating line art is not nearly as simple as detecting shades of gray.  I'd like to see your blue/white line art image.  If it does get transparency, I'm willing to bet it does only have two colors.
___
*[`src/BloomExe/ImageProcessing/ImageUtils.cs` line 191 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6032#-NanO3DS7KhnQVSK9H3Z:-NarKvlX-Ls7i-wrYlAy:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/369c42e2322ff26ff50f20ca52142a8fe3942016/src/BloomExe/ImageProcessing/ImageUtils.cs#L191)):*

I don't think we want to add this check for non-white.
Consider the case where we happen to miss all the non-white pixels in our 100-pixel sample.
</blockquote></details>

Done.

andrew-polk (changes requested):
Looks good. Thanks.
Much simpler.

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6032#-:-NaskyKD1YfanPpTEgJe:booys4c)* status: all files reviewed, 1 unresolved discussion (waiting on @StephenMcConnel)

___
*[`src/BloomExe/Book/BookCompressor.cs` line 371 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6032#-NaskgHw4i2kWcxg2EAQ:-NaskgHxCiu2UlqDEqCI:b-7ewtlk) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/0bf633bfcb4fc57b053350b8eca9619f8585ec31/src/BloomExe/Book/BookCompressor.cs#L371)):*
Should this say
> // Making white pixels transparent for non-two-tone PNG files is not a good idea.  See BL-12570.

?

Or maybe just
> // Making white pixels transparent for colorful PNG files is not a good idea.  See BL-12570.

StephenMcConnel (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6032)* status: all files reviewed, 1 unresolved discussion (waiting on @andrew-polk)

___
*[`src/BloomExe/Book/BookCompressor.cs` line 371 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6032#-NaskgHw4i2kWcxg2EAQ:-NavgjNL4lDNX3YnfBXw:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/0bf633bfcb4fc57b053350b8eca9619f8585ec31/src/BloomExe/Book/BookCompressor.cs#L371)):*

Should this say
> // Making white pixels transparent for non-two-tone PNG files is not a good idea.  See BL-12570.

?

Or maybe just
> // Making white pixels transparent for colorful PNG files is not a good idea.  See BL-12570.
</blockquote></details>

Done.

andrew-polk (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6032)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @StephenMcConnel)

PR #6037, opened 2023-08-03
Upload sanitized collection settings for harvester use (BL-12583)

StephenMcConnel:
(also removed some unused code)

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6037)

### DIFF
--- a/src/BloomExe/WebLibraryIntegration/BloomS3Client.cs
+++ b/src/BloomExe/WebLibraryIntegration/BloomS3Client.cs
@@ -209,7 +209,8 @@ namespace Bloom.WebLibraryIntegration
 		/// </summary>
 		public void UploadBook(string storageKeyOfBookFolder, string pathToBloomBookDirectory, IProgress progress,
 			string pdfToInclude, bool includeNarrationAudio, bool includeMusic,
-			string[] languagesToInclude, string metadataLang1Code, string metadataLang2Code, bool isForBulkUpload = false)
+			string[] languagesToInclude, string metadataLang1Code, string metadataLang2Code,
+			string collectionSettingsPath = null, bool isForBulkUpload = false)
 		{
 			BaseUrl = null;
 			BookOrderUrlOfRecentUpload = null;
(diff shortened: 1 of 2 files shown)


StephenMcConnel:
This might be a candidate for cherry-picking to Bloom 5.4 if it looks safe enough.

StephenMcConnel (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6037)* status: 0 of 2 files reviewed, 2 unresolved discussions (waiting on @andrew-polk)

___
*[`src/BloomExe/WebLibraryIntegration/BloomS3Client.cs` line 554 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6037#-NawaeHd1Am9uF3WjvF2:-Nax2Yf-DhhHINOZP2ln:bdp2u2a) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/0c6637c78129f365d8cb4dd89864cc4c1c837c38/src/BloomExe/WebLibraryIntegration/BloomS3Client.cs#L554)):*

If I'm not mistaken, we want non-harvester to avoid downloading anything in the collectionFiles directory.
</blockquote></details>

My initial code used Contains matching against "collectionFiles/".  I can go back to that.
___
*[`src/BloomExe/WebLibraryIntegration/BookUpload.cs` line 479 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6037#-Naw_gb95gd-iCNTgEl2:-Nax209q5RdYPtMWwzw7:bn3pwkj) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/0c6637c78129f365d8cb4dd89864cc4c1c837c38/src/BloomExe/WebLibraryIntegration/BookUpload.cs#L479)):*

It is unfortunate to have to run through this (probably not inexpensive) code now for the Enterprise path.

Does it work to just call `CopyCollectionSettingsToTempFolder` at the beginning of the method (before the early return)?

</blockquote></details>

We could postpone the copying to the actual uploading code which currently just does whitelisting.  That design may be cleaner in some ways.

andrew-polk (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6037)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @StephenMcConnel)

PR #6047, opened 2023-08-10
BL-12553 Update Features before bulk uploading

gmartin7:
* if upload is in Bulk mode, it will now update all features in a book before
   uploading

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6047)

### DIFF
--- a/src/BloomExe/WebLibraryIntegration/BookUpload.cs
+++ b/src/BloomExe/WebLibraryIntegration/BookUpload.cs
@@ -566,6 +566,18 @@ namespace Bloom.WebLibraryIntegration
 					}
 				}
 
+				if (bookParams.IsForBulkUpload)
+				{
+					// Convert the dictionary of language codes and booleans to a list of language codes.
+					var talkingBookLangs = book.BookInfo.PublishSettings.BloomLibrary.AudioLangs.
+						Where(kvp => kvp.Value.IsIncluded()).
+						Select(kvp => kvp.Key);
+					// Update all metadata features, since we are doing bulk upload and the user won't be able to do it
+					// by checking boxes individually. The two 'true' params mean to act like the user checked both Talking Book and
+					// Sign Language boxes. (BL-12553)
+					book.UpdateMetadataFeatures(true, true, talkingBookLangs);
+				}
+
 				if (progress.CancelRequested)
 					return "";
 


andrew-polk (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6047)* status: all files reviewed, 1 unresolved discussion (waiting on @gmartin7)

___
*[`src/BloomExe/WebLibraryIntegration/BookUpload.cs` line 574 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6047#-Nbb3_SLAohzoij2CXlO:-Nbb3_SM81bddijn4O2h:bljni1i) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/5ceaa30c4241f7375925f0a2cac2c7477e24609b/src/BloomExe/WebLibraryIntegration/BookUpload.cs#L574)):*
> ```Smalltalk
> 					// by checking boxes individually. The two 'true' params mean to act like the user checked both Talking Book and
> 					// Sign Language boxes. (BL-12553)
> 					book.UpdateMetadataFeatures(true, true, null);
> ```

I'm unsure about passing null here for allowedLanguages.
I think that means if they haven't gone into the publish screen, they will get all languages marked as having talking book. But don't we only want the ones which will actually get uploaded with narration?

gmartin7 (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6047)* status: all files reviewed, 1 unresolved discussion (waiting on @andrew-polk)

___
*[`src/BloomExe/WebLibraryIntegration/BookUpload.cs` line 574 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6047#-Nbb3_SLAohzoij2CXlO:-Nbu3-Wz2h48uGVdQJ7y:b1o1t2w) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/5ceaa30c4241f7375925f0a2cac2c7477e24609b/src/BloomExe/WebLibraryIntegration/BookUpload.cs#L574)):*

I'm unsure about passing null here for allowedLanguages.
I think that means if they haven't gone into the publish screen, they will get all languages marked as having talking book. But don't we only want the ones which will actually get uploaded with narration?
</blockquote></details>

I don't think so. My reading of the code in `UpdateTalkingBookFeature()` is that passing null means any language with audio will get the Talking Book Feature, whereas `allowedLanguages` in single book upload can restrict it further, so that we don't upload Spanish (e.g.) even though there is `lang="es"` text that has attached audio.

But thanks, it was good to check that.

andrew-polk (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6047)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @gmartin7)

PR #6054, opened 2023-08-11
BL-12451 Redux

hatton:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6054)

### DIFF
--- a/src/BloomExe/WebView2Browser.cs
+++ b/src/BloomExe/WebView2Browser.cs
@@ -1,6 +1,7 @@
 using Bloom.Api;
 using Bloom.Book;
 using Bloom.Edit;
+using Bloom.Properties;
 using Microsoft.Web.WebView2.Core;
 using Microsoft.Web.WebView2.WinForms;
 using Microsoft.Win32;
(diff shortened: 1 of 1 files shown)


gmartin7 (changes requested):
My only gripes are spelling. :)

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6054#-:-NbaTwNoE2CLpVfH5kp_:be00b0i)* status: all files reviewed, 3 unresolved discussions (waiting on @hatton)

___
*[`src/BloomExe/WebView2Browser.cs` line 170 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6054#-NbaUD9v6apA8aKQLn8M:-NbaUD9v6apA8aKQLn8N:b-wzwinx) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/badec70f22342b771e0cd96bd9370616381505b2/src/BloomExe/WebView2Browser.cs#L170)):*
> ```Smalltalk
> 			// the same user data folder. That is super rare, but expensive when it happens and the dev or user doesn't know why.
> 			// Therefore,
> 			// 1) we will only set the UI language of the borwser once per run of Bloom. As a result, we don't get to pass on the UI language to the
> ```

browser
___
*[`src/BloomExe/WebView2Browser.cs` line 189 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6054#-NbaUTq-3HxGyxXR6LLL:-NbaUTq-3HxGyxXR6LLM:b-scd8rk) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/badec70f22342b771e0cd96bd9370616381505b2/src/BloomExe/WebView2Browser.cs#L189)):*
> ```Smalltalk
> 			// In 5.5 time period, John Hatton kept getting into a situation where no version of Bloom would run,
> 			// and even a "Hello World" of WV2 would not run.
> 			// One hypothesis was that this was casused by an update to WV2, as it seemed to coincide, and also the
> ```

caused
___
*[`src/BloomExe/WebView2Browser.cs` line 214 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6054#-NbaTtnH1TouIticmlu1:-NbaTtnH1TouIticmlu2:bvomx5t) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/badec70f22342b771e0cd96bd9370616381505b2/src/BloomExe/WebView2Browser.cs#L214)):*
> ```Smalltalk
> 			// I suspect that some situations may require the user deleting this folder to get things working again.
> 			// Normally, it seems to get deleted automatically when we exit. But if we crash, it may not.
> 			// We are adding the language code becuase otherwise if you run two copies of Bloom with different languages, WV2 will fail to initialize.
> ```

because

andrew-polk (approved):
> Dismissed @gmartin7 from 3 discussions.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6054)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @hatton)

PR #6063, opened 2023-08-18
Fixes and improvements for Team Collection syncing (BL-12590)

andrew-polk:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6063)

### DIFF
--- a/src/BloomBrowserUI/teamCollection/TeamCollectionBookStatusPanel.tsx
+++ b/src/BloomBrowserUI/teamCollection/TeamCollectionBookStatusPanel.tsx
@@ -31,7 +31,7 @@ export type StatusPanelState =
     | "lockedByMe" // book is checked out to me, here...I can edit it
     | "lockedByMeElsewhere" // book is checked out to me, but on another computer, so I can't edit it here
     | "needsReload" // the collection needs to be reloaded before we can do anything with this book
-    | "problem" // The book has a problem, like a conflict between my changes and someone else's
+    | "conflictingChange" // The book has a conflict between my changes and someone else's
     | "hasInvalidRepoData" // the book has a catastrophic problem: the repo version is unreadable.
     | "disconnected" // Can't tell what's going on, because we don't have a good connection to the repo
     | "lockedByMeDisconnected" // We're disconnected, but before that happened the book was checked out to me, here
@@ -65,15 +65,15 @@ export const TeamCollectionBookStatusPanel: React.FunctionComponent<IBookTeamCol
 
     // Calculate panel state
     React.useEffect(() => {
-        if (props.disconnected) {
+        if (props.isDisconnected) {
             setTcPanelState(
                 lockedByMe ? "lockedByMeDisconnected" : "disconnected"
             );
-        } else if (props.hasInvalidRepoData) {
+        } else if (props.invalidRepoDataErrorMsg) {
             setTcPanelState("hasInvalidRepoData");
-        } else if (props.hasAProblem) {
-            setTcPanelState("problem");
-        } else if (props.changedRemotely) {
+        } else if (props.hasConflictingChange) {
+            setTcPanelState("conflictingChange");
+        } else if (props.isChangedRemotely) {
             setTcPanelState("needsReload");
         } else if (props.who) {
             // locked by someone
@@ -90,13 +90,13 @@ export const TeamCollectionBookStatusPanel: React.FunctionComponent<IBookTeamCol
             setTcPanelState("unlocked");
         }
     }, [
-        props.disconnected,
-        props.hasAProblem,
-        props.changedRemotely,
+        props.isDisconnected,
+        props.hasConflictingChange,
+        props.isChangedRemotely,
         props.who,
         lockedByMe,
         props.currentUser,
-        props.hasInvalidRepoData
+        props.invalidRepoDataErrorMsg
     ]);
 
     React.useEffect(() => {
@@ -243,8 +243,7 @@ export const TeamCollectionBookStatusPanel: React.FunctionComponent<IBookTeamCol
         true
     );
 
-    // Also used for problem.
-    const mainTitleNeedsReload = useL10n(
+    const mainTitleRemoteChanged = useL10n(
         "The Team Collection folder received a changed version of the book you have selected.",
         "TeamCollection.NeedsReload",
         "",
@@ -253,7 +252,7 @@ export const TeamCollectionBookStatusPanel: React.FunctionComponent<IBookTeamCol
         true
     );
 
-    const subTitleHasProblem = useL10n(
+    const subTitleConflictingChange = useL10n(
         "The Checkin/Checkout system should normally prevent this, but it has happened. Bloom cannot automatically join the work that came in with the work you were doing; you will need Bloom team support for that. Bloom will move your version of the book to the Team Collection Lost & Found when you Reload.",
         "TeamCollection.ConflictingChangeDetails",
         "",
@@ -321,7 +320,7 @@ export const TeamCollectionBookStatusPanel: React.FunctionComponent<IBookTeamCol
             text: "Forget Changes & Check in Book...",
             l10nKey: "TeamCollection.ForgetChangesMenuItem",
             action: () => setForgetDialogOpen(true),
-            disabled: props.newLocalBook
+            disabled: props.isNewLocalBook
         });
     }
 
@@ -376,7 +375,6 @@ export const TeamCollectionBookStatusPanel: React.FunctionComponent<IBookTeamCol
                 // This is just a fallback, which hopefully will never be seen.
                 return (
                     <StatusPanelCommon
-                        lockState={state}
                         title={props.error}
                         subTitle=""
                         icon={
@@ -411,7 +409,6 @@ export const TeamCollectionBookStatusPanel: React.FunctionComponent<IBookTeamCol
                             ${busy &&
                                 "cursor: progress; .checkout-button{cursor:progress;}"}
                         `}
-                        lockState={state}
                         title={mainTitleUnlocked}
                         subTitle={subTitleUnlocked}
                         icon={
@@ -462,7 +459,6 @@ export const TeamCollectionBookStatusPanel: React.FunctionComponent<IBookTeamCol
                                 margin-top: 10px; // leaves some extra space for the "What changes did you make" overlay
                             }
                         `}
-                        lockState={state}
                         title={
                             checkinProgress === 0
                                 ? mainTitleLockedByMe
@@ -492,6 +488,7 @@ export const TeamCollectionBookStatusPanel: React.FunctionComponent<IBookTeamCol
                                 </ThemeProvider>
                             </StyledEngineProvider>
                         }
+                        useWarningColorForButton={true}
                         menu={menu}
                     >
                         {checkinProgress === 0 ? (
@@ -554,7 +551,6 @@ export const TeamCollectionBookStatusPanel: React.FunctionComponent<IBookTeamCol
             case "lockedByMeElsewhere":
                 return (
                     <StatusPanelCommon
-                        lockState={state}
                         title={mainTitleLockedElsewhere}
                         subTitle={subTitleLockedElsewhere}
                         icon={avatar}
@@ -566,7 +562,6 @@ export const TeamCollectionBookStatusPanel: React.FunctionComponent<IBookTeamCol
             case "locked":
                 return (
                     <StatusPanelCommon
-                        lockState={state}
                         title={mainTitleLocked}
                         subTitle={subTitleLocked}
                         icon={avatar}
@@ -575,12 +570,16 @@ export const TeamCollectionBookStatusPanel: React.FunctionComponent<IBookTeamCol
                         {getLockedInfoChild(lockedInfo)}
                     </StatusPanelCommon>
                 );
-            case "problem":
+            case "conflictingChange":
+            case "needsReload":
                 return (
                     <StatusPanelCommon
-                        lockState={state}
-                        title={mainTitleNeedsReload}
-                        subTitle={subTitleHasProblem}
+                        title={mainTitleRemoteChanged}
+                        subTitle={
+                            state === "conflictingChange"
+                                ? subTitleConflictingChange
+                                : subTitleNeedsReload
+                        }
                         icon={avatar}
                         button={getBloomButton(
                             "Reload",
@@ -589,6 +588,7 @@ export const TeamCollectionBookStatusPanel: React.FunctionComponent<IBookTeamCol
                             undefined,
                             () => post("common/reloadCollection")
                         )}
+                        useWarningColorForButton={state === "needsReload"} //REVIEW: why is it different?
                         menu={menu}
                     >
                         {getLockedInfoChild("")}
@@ -600,33 +600,13 @@ export const TeamCollectionBookStatusPanel: React.FunctionComponent<IBookTeamCol
                         css={css`
                             max-width: 560px;
                         `} // to match StatusPanelCommon
-                        errorMessage={props.hasInvalidRepoData}
+                        errorMessage={props.invalidRepoDataErrorMsg}
                         clickHereArg={props.clickHereArg}
                     />
-                );
-            case "needsReload":
-                return (
-                    <StatusPanelCommon
-                        lockState={state}
-                        title={mainTitleNeedsReload}
-                        subTitle={subTitleNeedsReload}
-                        icon={avatar}
-                        button={getBloomButton(
-                            "Reload",
-                            "TeamCollection.Reload",
-                            "reload-button",
-                            undefined,
-                            () => post("common/reloadCollection")
-                        )}
-                        menu={menu}
-                    >
-                        {getLockedInfoChild("")}
-                    </StatusPanelCommon>
                 );
             case "disconnected":
                 return (
                     <StatusPanelCommon
-                        lockState={state}
                         title={mainTitleDisconnected}
                         subTitle={subTitleDisconnected}
                         icon={
@@ -640,7 +620,6 @@ export const TeamCollectionBookStatusPanel: React.FunctionComponent<IBookTeamCol
             case "lockedByMeDisconnected":
                 return (
                     <StatusPanelCommon
-                        lockState={state}
                         title={mainTitleLockedByMe}
                         subTitle={subTitleDisconnectedCheckedOut}
                         icon={avatar}
--- a/src/BloomBrowserUI/teamCollection/statusPanelCommon.tsx
+++ b/src/BloomBrowserUI/teamCollection/statusPanelCommon.tsx
@@ -21,11 +21,11 @@ declare module "@mui/styles/defaultTheme" {
 }
 
 export interface IStatusPanelProps {
-    lockState: StatusPanelState;
     title: string;
     subTitle: string;
     icon: JSX.Element;
     button?: JSX.Element;
+    useWarningColorForButton?: boolean;
     children?: JSX.Element;
     menu?: JSX.Element; // when book is checked out, About my Avatar... and Forget Changes and Check in Book
     className?: string; // additional class for  root div; enables emotion CSS.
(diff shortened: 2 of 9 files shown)


andrew-polk:
*[`src/BloomExe/TeamCollection/TeamCollection.cs` line 2051 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6063#-Nc9NNne9Dy4nCzT4VEr:-Nc9OOr6EuNa2qFd1CH5:bk45krx) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/03780edec5aba266b1071fb09e2e1fc664557c85/src/BloomExe/TeamCollection/TeamCollection.cs#L2051)):*

This is the questionable bit.
The UI was set up to display the same message about replacing the book, lost & found, etc. whether we had a checkout conflict or a checksum conflict. The user was given a button to reload the collection.
But the back end was never set up to do anything if only the local (not book's current state but a record of what the checksum was when we checked out the book) checksum conflicted with the repo checksum.

So this PR makes the back end match the UI.
We could, alternatively, make the UI match the back end.
</blockquote></details>

@JohnThomson 
It would make sense for you to look at this and re-weigh the two options since you apparently decided it wasn't worth doing this the first time around.

JohnThomson (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6063)* status: 0 of 9 files reviewed, 4 unresolved discussions (waiting on @andrew-polk)

___
*[`src/BloomBrowserUI/teamCollection/TeamCollectionBookStatusPanel.tsx` line 574 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6063#-NcUUuZiAQJxFGpx1Meo:-NcUUuZiAQJxFGpx1Mep:b-9a96e3) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/03780edec5aba266b1071fb09e2e1fc664557c85/src/BloomBrowserUI/teamCollection/TeamCollectionBookStatusPanel.tsx#L574)):*
> ```XML
>                 );
>             case "conflictingChange":
>             case "needsReload":
> ```

I would have found it a difficult decision, I think, whether to merge these two cases. Two out of five params to StatusPanelCommon are different. They do have the same main title, so perhaps that deserves to be decisive. If there's a particular reason to treat them similarly it migtht be worth a comment.
___
*[`src/BloomBrowserUI/teamCollection/TeamCollectionBookStatusPanel.tsx` line 591 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6063#-NcUSP7UAwDC068d30pT:-NcUSP7UAwDC068d30pU:bkx8oqe) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/03780edec5aba266b1071fb09e2e1fc664557c85/src/BloomBrowserUI/teamCollection/TeamCollectionBookStatusPanel.tsx#L591)):*
> ```XML
>                             () => post("common/reloadCollection")
>                         )}
>                         useWarningColorForButton={state === "needsReload"} //REVIEW: why is it different?
> ```

I haven't quite figured out whether "useWarningColorForButtonState" indicates a more serious problem (than an informational message) or less serious (than say an Error). However, a book that has changed remotely while NOT checked out here, and therefore must be reloaded before it can be checked out, is a much less serious problem than one that has unexpectedly changed remotely while it IS checked out here. I would have expected that needsReload would NOT require a warning color, while conflictingChange requires at LEAST a warning color (if we're not using something stronger for errors). 
___
*[`src/BloomExe/TeamCollection/BookStatus.cs` line 30 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6063#-NcUVg8dDt-Y1Vw87_Ks:-NcUVg8dDt-Y1Vw87_Kt:bgagyia) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/03780edec5aba266b1071fb09e2e1fc664557c85/src/BloomExe/TeamCollection/BookStatus.cs#L30)):*
> ```Smalltalk
> 		public string ToSanitizedJson()
> 		{
> 			var sanitizedStatus = WithLockedBy(lockedBy == null ? null : lockedBy.Substring(0, 3) + "...", lockedByFirstName, lockedBySurname);
> ```

I'm missing the point of including the first three letters of what I think is typically an email address. Feels to me a bit too drastic if we're wanting to be able to identify the person while not fully revealing their email even if the names are missing. If there's some even less obvious reason it definitely deserves a comment. Is this targeting UI or an error report that will be published or...?
I also somewhat question the property name. It's not a sanitized version of the whole book status (as the name suggests) but only a sanitized version of who it is locked by. SanitizedLockedBy? LockedByJsonForUI?
___
*[`src/BloomExe/TeamCollection/TeamCollection.cs` line 1195 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6063#-NcUZu1tCF8qmO60G5S-:-NcUZu1tCF8qmO60G5S0:b-houeul) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/03780edec5aba266b1071fb09e2e1fc664557c85/src/BloomExe/TeamCollection/TeamCollection.cs#L1195)):*
> ```Smalltalk
> 				Logger.WriteEvent($"*** TeamCollection.HasLocalChangesThatMustBeClobbered(): conflicting checksum. Current={currentChecksum} localStatus={localStatus.checksum} repoStatus={repoStatus.checksum}");
> 
> 			return isConflictingCheckedOutStatus || isConflictingCheckSum;
> ```

Coming back to this long after, I can think of a couple of other cases that may or may not be worth handling specially, but don't absolutely require clobbering the book, or at least, not copying it to lost-and-found:
If repoStatus.checksum == localStatus.checksum, even though currentChecksum is different, the book has been edited locally, but the checked-in version is the same as what this user started editing. It's a catastrophic problem if repoStatus indicates that someone else has the book checked out, but if repoStatus indicates that it isn't checked out at all, we could reasonably just fix repoStatus. (Currently it looks like we're not distinguishing between "someone else has it checked out" and "the repo doesn't think that it is checked out at all.)
If curentChecksum != repoStatus.checksum but localStatus.checksum == currentChecksum, the book hasn't actually been edited locally. It's just (surprisingly) been changed remotely even though this user has it checked out. I'm not actually sure what the method should return: does it just mean that the local book must be updated from the repo before editing, or that it must also be copied to lost-and-found? The former is true, but the latter is not, since there aren't any local changes to save. I'm not clear whether this involves a new status panel option or whether one of the existing ones will serve.
I've commented more below on what SyncAtStartup does/should do in these cases. I think at least some more fixing is needed on the UI side, possibly at least partly by improving the logic here.
___
*[`src/BloomExe/TeamCollection/TeamCollection.cs` line 2051 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6063#-Nc9NNne9Dy4nCzT4VEr:-NcUeDpS9xx1doFBfC_f:b6wp2zc) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/03780edec5aba266b1071fb09e2e1fc664557c85/src/BloomExe/TeamCollection/TeamCollection.cs#L2051)):*

@JohnThomson 
It would make sense for you to look at this and re-weigh the two options since you apparently decided it wasn't worth doing this the first time around.
</blockquote></details>

So, as things currently stand, if the book is checked out locally but the repo and local checksums don't match, we're going to behave as if it wasn't checked out locally. The first branch below won't be taken, since the new case only applies when localAndRepoChecksumsMatch is false. So the new case is where
- book exists locally and in repo
- book has status locally (and of course in repo)
- local status says it is checked out here
- remote status says it is not checked out here
- local and remote checksums differ
And then we will
1. if it hasn't been changed locally (local and CURRENT checksums match), update content to match repo, and status to reflect local checkout
2. if is HAS been changed locally, push local to Lost and Found, and update local to match repo

I do think I see some problems:
1. case 1 above assumes (as was previously guaranteed on this branch) that the repo agrees that the book is checked out here. If that's true, or if the repo says it's not checked out at all, this is fine (if necessary the code will also update the repo to indicate the book is checked out here).  But I'm not sure what we're saying in the status panel reflects this well...we may be using "this is a catastrophe your changes have been moved to lost and found" when we don't need to (and nothing will be there if they look in Lost and Found)
2. But (still in case 1) if the repo indicates that the book is checked out to someone else, we DEFINITELY should update local status to reflect that instead of updating the repo to indicate the local checkout. (But still if there are no local changes it's not a disaster, so the status panel message may need tweaking.)

So it's not really an either-or: change the UI or change the SyncAtStartup. Since we've actually observed conflicting local and remote checksums when the book is checked out locally, SyncAtStartup definitely needs to handle those cases...and I think what you have is still not right for the case where local status == current status, remote status != local status, local status is checked out here, and remote status is checked out somewhere else. (And of course there should be appropriate test cases.)
But we also need UI changes...and I'm not sure we have all we need yet...since some of the cases where local status differs from remote status are less catastrophic than others.

andrew-polk (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6063)* status: 0 of 9 files reviewed, 4 unresolved discussions (waiting on @JohnThomson)

___
*[`src/BloomBrowserUI/teamCollection/TeamCollectionBookStatusPanel.tsx` line 574 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6063#-NcUUuZiAQJxFGpx1Meo:-Ndr42__BSn7hfyDpEmB:b1srtjd) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/03780edec5aba266b1071fb09e2e1fc664557c85/src/BloomBrowserUI/teamCollection/TeamCollectionBookStatusPanel.tsx#L574)):*

I would have found it a difficult decision, I think, whether to merge these two cases. Two out of five params to StatusPanelCommon are different. They do have the same main title, so perhaps that deserves to be decisive. If there's a particular reason to treat them similarly it migtht be worth a comment.
</blockquote></details>

Yeah, I'm unsure, too. But it feels weird to me that the button color is different, so I guess I was leaning toward them really being the same scenario, just with two different messages.
___
*[`src/BloomBrowserUI/teamCollection/TeamCollectionBookStatusPanel.tsx` line 591 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6063#-NcUSP7UAwDC068d30pT:-Ndr3t-V0iqdV4T7sWB6:btyy30u) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/03780edec5aba266b1071fb09e2e1fc664557c85/src/BloomBrowserUI/teamCollection/TeamCollectionBookStatusPanel.tsx#L591)):*

I haven't quite figured out whether "useWarningColorForButtonState" indicates a more serious problem (than an informational message) or less serious (than say an Error). However, a book that has changed remotely while NOT checked out here, and therefore must be reloaded before it can be checked out, is a much less serious problem than one that has unexpectedly changed remotely while it IS checked out here. I would have expected that needsReload would NOT require a warning color, while conflictingChange requires at LEAST a warning color (if we're not using something stronger for errors).
</blockquote></details>

Yeah, I was keeping the UI the same as it was.
Not sure if that's right, either.
___
*[`src/BloomExe/TeamCollection/BookStatus.cs` line 30 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6063#-NcUVg8dDt-Y1Vw87_Ks:-Ndr4HAo52GD_VrZ_yXf:b-hkclod) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/03780edec5aba266b1071fb09e2e1fc664557c85/src/BloomExe/TeamCollection/BookStatus.cs#L30)):*

I'm missing the point of including the first three letters of what I think is typically an email address. Feels to me a bit too drastic if we're wanting to be able to identify the person while not fully revealing their email even if the names are missing. If there's some even less obvious reason it definitely deserves a comment. Is this targeting UI or an error report that will be published or...?
I also somewhat question the property name. It's not a sanitized version of the whole book status (as the name suggests) but only a sanitized version of who it is locked by. SanitizedLockedBy? LockedByJsonForUI?
</blockquote></details>

This is just put in the log file.
That gets put in our error reports, so we can't include the full email address.
I thought 3 characters was enough, especially since we have first and last name.
___
*[`src/BloomExe/TeamCollection/TeamCollection.cs` line 2051 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6063#-Nc9NNne9Dy4nCzT4VEr:-NePQxH8CAZ6Oh0bE2NG:b-bjmb54) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/03780edec5aba266b1071fb09e2e1fc664557c85/src/BloomExe/TeamCollection/TeamCollection.cs#L2051)):*

So, as things currently stand, if the book is checked out locally but the repo and local checksums don't match, we're going to behave as if it wasn't checked out locally. The first branch below won't be taken, since the new case only applies when localAndRepoChecksumsMatch is false. So the new case is where
- book exists locally and in repo
- book has status locally (and of course in repo)
- local status says it is checked out here
- remote status says it is not checked out here
- local and remote checksums differ
And then we will
1. if it hasn't been changed locally (local and CURRENT checksums match), update content to match repo, and status to reflect local checkout
2. if is HAS been changed locally, push local to Lost and Found, and update local to match repo

I do think I see some problems:
1. case 1 above assumes (as was previously guaranteed on this branch) that the repo agrees that the book is checked out here. If that's true, or if the repo says it's not checked out at all, this is fine (if necessary the code will also update the repo to indicate the book is checked out here).  But I'm not sure what we're saying in the status panel reflects this well...we may be using "this is a catastrophe your changes have been moved to lost and found" when we don't need to (and nothing will be there if they look in Lost and Found)
2. But (still in case 1) if the repo indicates that the book is checked out to someone else, we DEFINITELY should update local status to reflect that instead of updating the repo to indicate the local checkout. (But still if there are no local changes it's not a disaster, so the status panel message may need tweaking.)

So it's not really an either-or: change the UI or change the SyncAtStartup. Since we've actually observed conflicting local and remote checksums when the book is checked out locally, SyncAtStartup definitely needs to handle those cases...and I think what you have is still not right for the case where local status == current status, remote status != local status, local status is checked out here, and remote status is checked out somewhere else. (And of course there should be appropriate test cases.)
But we also need UI changes...and I'm not sure we have all we need yet...since some of the cases where local status differs from remote status are less catastrophic than others.
</blockquote></details>

We discussed this and decided @JohnThomson's comments were based on the incorrect premise that local and remote checkout status disagree. But in fact they do agree. So he says this can be resolved.

andrew-polk (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6063)* status: 0 of 9 files reviewed, 4 unresolved discussions (waiting on @JohnThomson)

___
*[`src/BloomExe/TeamCollection/TeamCollection.cs` line 1195 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6063#-NcUZu1tCF8qmO60G5S-:-NePY3PODsS9qZFi0J8o:b-7ekra7) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/03780edec5aba266b1071fb09e2e1fc664557c85/src/BloomExe/TeamCollection/TeamCollection.cs#L1195)):*

Coming back to this long after, I can think of a couple of other cases that may or may not be worth handling specially, but don't absolutely require clobbering the book, or at least, not copying it to lost-and-found:
If repoStatus.checksum == localStatus.checksum, even though currentChecksum is different, the book has been edited locally, but the checked-in version is the same as what this user started editing. It's a catastrophic problem if repoStatus indicates that someone else has the book checked out, but if repoStatus indicates that it isn't checked out at all, we could reasonably just fix repoStatus. (Currently it looks like we're not distinguishing between "someone else has it checked out" and "the repo doesn't think that it is checked out at all.)
If curentChecksum != repoStatus.checksum but localStatus.checksum == currentChecksum, the book hasn't actually been edited locally. It's just (surprisingly) been changed remotely even though this user has it checked out. I'm not actually sure what the method should return: does it just mean that the local book must be updated from the repo before editing, or that it must also be copied to lost-and-found? The former is true, but the latter is not, since there aren't any local changes to save. I'm not clear whether this involves a new status panel option or whether one of the existing ones will serve.
I've commented more below on what SyncAtStartup does/should do in these cases. I think at least some more fixing is needed on the UI side, possibly at least partly by improving the logic here.
</blockquote></details>

Your scenario 1:
- Repo thinks book is not checked out
- repo and local checksums agree (agree on contents at the time it was checked out)
- there are local changes

Based on experimentation, it appears it already does what you suggest. I.e. it updates the repo with the correct checked out status. And it never seems to present any message about a problem.

Your scenario 2:
- Book hasn't been changed locally. (though repo and local checksums differ)
Based on the method comment, that is a prerequisite for this method to return true. So I think I'll leave that case alone.


So I think we're ready to merge?

JohnThomson (approved):
I'm ready to approve this, with just a couple of minor questions causing me to leave it for you to decide whether to make any more tweaks before merging. But it would definitely be an improvement even as-is.

> Reviewed 7 of 9 files at r1, 2 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6063#-:-NfIw2aW5sQfHUbGb6VR:bywf604)* status: all files reviewed, 2 unresolved discussions (waiting on @andrew-polk)

___
*[`src/BloomBrowserUI/teamCollection/TeamCollectionBookStatusPanel.tsx` line 574 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6063#-NcUUuZiAQJxFGpx1Meo:-NfIucOz9vQrX1ALvp7_:b-x1vxgw) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/03780edec5aba266b1071fb09e2e1fc664557c85/src/BloomBrowserUI/teamCollection/TeamCollectionBookStatusPanel.tsx#L574)):*

Yeah, I'm unsure, too. But it feels weird to me that the button color is different, so I guess I was leaning toward them really being the same scenario, just with two different messages.
</blockquote></details>

If they mean what I think (i.e., needsReload just means a remote change but no local change, while conflictingChange indicates that local changes are about to be moved to lost and found and otherwise discarded, then the latter is a considerably more serious problem. But I'm comfortable with you making the decision whether anything further should be improved. You certainly haven't made anything worse.
___
*[`src/BloomBrowserUI/teamCollection/TeamCollectionBookStatusPanel.tsx` line 591 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6063#-NcUSP7UAwDC068d30pT:-NfIvKF68ngn1RKJ2KHS:b-u7ovgu) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/03780edec5aba266b1071fb09e2e1fc664557c85/src/BloomBrowserUI/teamCollection/TeamCollectionBookStatusPanel.tsx#L591)):*

Yeah, I was keeping the UI the same as it was.
Not sure if that's right, either.
</blockquote></details>

I think you should confirm that useWarningColorForButton true indicates a less serious problem than false. That's probably true...as I think about it again, it's probably talking about a Reload button, which we probably only have at all if we need reload, so the alternative to a warning color is probably something more serious.

StephenMcConnel (approved):
:lgtm: I didn't notice anything beyond what John commented on, and he was willing to go with your judgement.

> Reviewed 7 of 9 files at r1, 2 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6063#-:-NfMRW-Z10jNnmbAYJE7:b-8sujmz)* status: all files reviewed, 2 unresolved discussions (waiting on @andrew-polk)

andrew-polk (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6063)* status: all files reviewed, 2 unresolved discussions (waiting on @JohnThomson)

___
*[`src/BloomBrowserUI/teamCollection/TeamCollectionBookStatusPanel.tsx` line 591 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6063#-NcUSP7UAwDC068d30pT:-NfMsOFzCWayuDvvghmA:b-pd9qrk) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/03780edec5aba266b1071fb09e2e1fc664557c85/src/BloomBrowserUI/teamCollection/TeamCollectionBookStatusPanel.tsx#L591)):*

I think you should confirm that useWarningColorForButton true indicates a less serious problem than false. That's probably true...as I think about it again, it's probably talking about a Reload button, which we probably only have at all if we need reload, so the alternative to a warning color is probably something more serious.
</blockquote></details>

I introduced `useWarningColorForButton` as an alternative to passing in the state to the status component.
The original code in the internal component was 
```
const buttonColor =
        props.lockState === "lockedByMe" || props.lockState === "needsReload"
            ? outerTheme.palette.warning.main
            : outerTheme.palette.primary.main;
```
Based on that, I'm not sure that using the "warning" color was really intended to convey anything either way about there being a problem. i.e. the "check in" button gets the "warning" color, but that is clearly not a problem state.

So rather than ask John to rethink all the states and what colors the buttons should be, I think I'll just leave things as they are.

Thanks for pointing it out.

PR #6074, opened 2023-08-29
Simplify CollectionSettingsDialog and fix error on launch (BL-12409)

andrew-polk:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6074)

### DIFF
--- a/src/BloomBrowserUI/collection/CollectionSettingsDialog.tsx
+++ b/src/BloomBrowserUI/collection/CollectionSettingsDialog.tsx
@@ -6,47 +6,35 @@ import {
     BloomDialog,
     DialogBottomButtons,
     DialogMiddle,
-    DialogTitle,
-    IBloomDialogProps
+    DialogTitle
 } from "../react_components/BloomDialog/BloomDialog";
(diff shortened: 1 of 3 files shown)


andrew-polk (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6074)* status: all files reviewed, 1 unresolved discussion (waiting on @gmartin7)

___
*[`src/BloomExe/web/controllers/CollectionSettingsApi.cs` line 79 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6074#-Nd1bsr713NSLL3av3Zj:-Nd1zR_198SezOdKaR6D:bddx1c4) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/2ed31caef5c94f5a7cb58d197219cd2ab446dbea/src/BloomExe/web/controllers/CollectionSettingsApi.cs#L79)):*

Does this actually do anything, or would it be better to complain to the developer who accidentally does a `post` to "collection/settings"?
</blockquote></details>

This is all just setting up the skeleton for the other cards which will fill it in.

PR #6102, opened 2023-09-13
[5.5] BL-12681 Link to docs about Dropbox

gmartin7:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6102)

### DIFF
--- a/DistFiles/localization/en/Bloom.xlf
+++ b/DistFiles/localization/en/Bloom.xlf
@@ -5712,6 +5712,11 @@ Do you want to go ahead?</note>
         <source xml:lang="en">Bloom will create a new collection on your computer based on this Team Collection.</source>
         <note>ID: TeamCollection.Creating</note>
       </trans-unit>
+      <trans-unit id="TeamCollection.NeedDropboxRunning" sil:dynamic="true">
+        <source xml:lang="en">Dropbox does not appear to be running. Please see [Troubleshooting Dropbox](https://docs.bloomlibrary.org/dropbox-trouble).</source>
+        <note>ID: TeamCollection.NeedDropboxRunning</note>
+        <note>The text inside the [square brackets] will become a link to the URL in parentheses.</note>
+      </trans-unit>
       <trans-unit id="TeamCollection.Intro" sil:dynamic="true">
         <source xml:lang="en">Bloom's Team Collection system helps your team collaborate as you create, translate, and edit books. Read about how it works [here]({0}), or view this [video]({1}).</source>
         <note>ID: TeamCollection.Intro</note>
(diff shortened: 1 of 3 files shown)


andrew-polk (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6102)* status: all files reviewed, 1 unresolved discussion (waiting on @gmartin7)

___
*[`src/BloomExe/TeamCollection/TeamCollectionMessage.cs` line 142 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6102#-NeFVc-B4nHmCSf5Gh41:-NeFVc-B4nHmCSf5Gh42:b-2k474h) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/d83610c103eda84d8f5e04ed2db2cb7898985c19/src/BloomExe/TeamCollection/TeamCollectionMessage.cs#L142)):*
> ```Smalltalk
> 					msg = CreateLinkInMessage(msg, Param0);
> 					return leadIn + msg;
> 				}
> ```

Explore an idea with me...
Suppose that instead we did markdown like
`Dropbox does not appear to be running. Please see [Troubleshooting Dropbox](https://docs.bloomlibrary.org/dropbox-trouble).`

Advantages:
- I *think* that simplifies the special logic here for making the link. A regular expression to detect a markdown link and replace it with an `a` tag should be pretty straightforward and more robust. (Imagine the scenario where we want to include square brackets and a param in the future.)
- Translators can actually point people to the localized version of the url like `https://docs.bloomlibrary.org/fr/dropbox-trouble`.

Disadvantages:
- Translators could botch the url. (But Crowdin helpfully guides people to not do this...)

Thoughts?

gmartin7 (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6102)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @andrew-polk)

___
*[`src/BloomExe/TeamCollection/TeamCollectionMessage.cs` line 142 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6102#-NeFVc-B4nHmCSf5Gh41:-NeJhEI7-HcJ5ZAOAUl8:bt4c2o5) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/d83610c103eda84d8f5e04ed2db2cb7898985c19/src/BloomExe/TeamCollection/TeamCollectionMessage.cs#L142)):*

Explore an idea with me...
Suppose that instead we did markdown like
`Dropbox does not appear to be running. Please see [Troubleshooting Dropbox](https://docs.bloomlibrary.org/dropbox-trouble).`

Advantages:
- I *think* that simplifies the special logic here for making the link. A regular expression to detect a markdown link and replace it with an `a` tag should be pretty straightforward and more robust. (Imagine the scenario where we want to include square brackets and a param in the future.)
- Translators can actually point people to the localized version of the url like `https://docs.bloomlibrary.org/fr/dropbox-trouble`.

Disadvantages:
- Translators could botch the url. (But Crowdin helpfully guides people to not do this...)

Thoughts?
</blockquote></details>

Done. Once I got over my unfortunate aversion to regex, yes, it's much simpler and it's safe to pass everything through even if the message doesn't contain a link.

andrew-polk (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6102)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @gmartin7)

PR #6121, opened 2023-09-27
Require book checkout in TC for publish cover color button (BL-12588)

StephenMcConnel:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6121)


hatton (commented):
> Nothing causes the button or label to look disabled, but the tooltip pops up explaining that it is disabled and why.

Let me do some thinkin' on this.

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6121#-:-NfMQacxEV1FldbR6Dgi:b-asoob5)* status: 0 of 2 files reviewed, all discussions resolved

PR #6130, opened 2023-10-03
[5.4] BL-12712 Ace by Daisy catastrophic failure

gmartin7:
* I apparently don't have AbD installed. I thought I did.
   But now Bloom correctly tells me it can't find it.

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6130)

### DIFF
--- a/src/BloomExe/web/controllers/AccessibilityCheckApi.cs
+++ b/src/BloomExe/web/controllers/AccessibilityCheckApi.cs
@@ -317,20 +317,23 @@ namespace Bloom.web.controllers
 			var whereProgram = Platform.IsWindows ? "where" : "which";
 			var npmFileName = Platform.IsWindows ? "npm.cmd" : "npm";
 			var whereResult = CommandLineRunner.Run(whereProgram, npmFileName, Encoding.ASCII, "", 2, new NullProgress());
-			if (!String.IsNullOrEmpty(whereResult.StandardError))
+			if (!string.IsNullOrEmpty(whereResult.StandardError))
 			{
 				_webSocketProgress.MessageWithoutLocalizing(whereResult.StandardError, ProgressKind.Error);
 			}
-			if (!whereResult.StandardOutput.Contains(npmFileName))
+			if (string.IsNullOrEmpty(whereResult.StandardOutput) || !whereResult.StandardOutput.Contains(npmFileName))
 			{
 				ReportErrorAndFailTheRequest(request, whereResult, "Could not find npm.");
 				return null;
 			}
-
-			var fullNpmPath = whereResult.StandardOutput.Split('\n')[0].Trim();
+			// BL-12712: Adding Volta created a situation where there can be more than one location for 'npm.cmd'.
+			// If you run any npm command in the Volta directory, it will cause Volta to spawn
+			// 1000s of spurious npm processes! So we set the working directory to empty string below.
+			var outputArray = whereResult.StandardOutput.Split(new []{'\n'}, StringSplitOptions.RemoveEmptyEntries);
+			var fullNpmPath = outputArray[0].Trim(); // If the (dev) machine has Volta installed, this will be the Volta npm.
 			// note: things like nvm will mess with where the global node_modules lives. The best way seems to be
 			// to ask npm:
-			var result = CommandLineRunner.Run(npmFileName, "root -g", Encoding.ASCII, Path.GetDirectoryName(fullNpmPath), 10,
+			var result = CommandLineRunner.Run(fullNpmPath, "root -g", Encoding.ASCII, "", 10,
 				new NullProgress());
 
 			const string kCoreError = "Could not get \"npm -g root\" to work. Is Node & npm installed and working?";
@@ -357,7 +360,7 @@ namespace Bloom.web.controllers
 				return null;
 			}
 
-			var nodeModulesDirectory = result.StandardOutput.Trim();
+			var nodeModulesDirectory = result.StandardOutput.Trim(); // Standard Output result has an ending carriage return for some reason.
 
 			if (!Directory.Exists((nodeModulesDirectory)))
 			{


andrew-polk (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6130)* status: all files reviewed, 1 unresolved discussion (waiting on @gmartin7)

___
*[`src/BloomExe/web/controllers/AccessibilityCheckApi.cs` line 331 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6130#-NfrSSI_5WobrmreX9D-:-NfrSSIaDs9iRkZtpwf0:bpw28sa) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/54384e26a3089c93e56eefa166f1108d583eeed3/src/BloomExe/web/controllers/AccessibilityCheckApi.cs#L331)):*
> ```Smalltalk
> 			}
> 			// Adding Volta created a situation where there can be more than one location for npm.cmd.
> 			// If we get more than one result, we need the last non-empty one. (non-Volta)
> ```

I think we need to say more.
Why doesn't the volta one work?
Why is it safe to always pick the last one?
It seems that they would probably be in the order the system would use them, so using the first one would be most natural.

gmartin7 (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6130)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @andrew-polk)

___
*[`src/BloomExe/web/controllers/AccessibilityCheckApi.cs` line 331 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6130#-NfrSSI_5WobrmreX9D-:-Nfvm3472hzuzc56QKZW:bs2xcey) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/54384e26a3089c93e56eefa166f1108d583eeed3/src/BloomExe/web/controllers/AccessibilityCheckApi.cs#L331)):*

I think we need to say more.
Why doesn't the volta one work?
Why is it safe to always pick the last one?
It seems that they would probably be in the order the system would use them, so using the first one would be most natural.
</blockquote></details>

As noted in the comments, the Volta version does the spawning of 1000s of processes. I added back in the Encoding and it seems to be working fine, despite many tries on the `where` command above coming back with a timeout error before today!
Today's testing:
* successfully finds Ace by Daisy and uses it to analyze the ePUB file.
* if I delete Ace by Daisy, Bloom successfully determines it isn't available and says so.

andrew-polk (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6130)* status: all files reviewed, 1 unresolved discussion (waiting on @gmartin7)

___
*[`src/BloomExe/web/controllers/AccessibilityCheckApi.cs` line 331 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6130#-NfrSSI_5WobrmreX9D-:-NfvqITGBMAFcPNNgtIB:bkad2gd) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/54384e26a3089c93e56eefa166f1108d583eeed3/src/BloomExe/web/controllers/AccessibilityCheckApi.cs#L331)):*

As noted in the comments, the Volta version does the spawning of 1000s of processes. I added back in the Encoding and it seems to be working fine, despite many tries on the `where` command above coming back with a timeout error before today!
Today's testing:
* successfully finds Ace by Daisy and uses it to analyze the ePUB file.
* if I delete Ace by Daisy, Bloom successfully determines it isn't available and says so.
</blockquote></details>

If we really need to code for not-the-Volta-one, isn't that different than just taking the last one?
I'm not sure this is correct, either, but wouldn't it be better to get the first one which isn't *volta*?

andrew-polk:
*[`src/BloomExe/web/controllers/AccessibilityCheckApi.cs` line 331 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6130#-NfrSSI_5WobrmreX9D-:-NfvqfBhEeTe3wmNmwSu:bay3y3s) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/54384e26a3089c93e56eefa166f1108d583eeed3/src/BloomExe/web/controllers/AccessibilityCheckApi.cs#L331)):*

If we really need to code for not-the-Volta-one, isn't that different than just taking the last one?
I'm not sure this is correct, either, but wouldn't it be better to get the first one which isn't *volta*?
</blockquote></details>

I meant to say where the path doesn't include "volta"

andrew-polk:
*[`src/BloomExe/web/controllers/AccessibilityCheckApi.cs` line 331 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6130#-NfrSSI_5WobrmreX9D-:-NfvvD2k1RRkT2XRpZ9r:b-tedet9) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/54384e26a3089c93e56eefa166f1108d583eeed3/src/BloomExe/web/controllers/AccessibilityCheckApi.cs#L331)):*

I meant to say where the path doesn't include "volta"
</blockquote></details>

Looking at the code a little more, if the problem is with the "root -g" params, what can we do to move away from that?

As I understand, all we're trying to do here is verify that npm is available?
How about `npm --version`?

andrew-polk:
*[`src/BloomExe/web/controllers/AccessibilityCheckApi.cs` line 331 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6130#-NfrSSI_5WobrmreX9D-:-NfvvoNyALpzVCfUtg-i:bjbzqj5) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/54384e26a3089c93e56eefa166f1108d583eeed3/src/BloomExe/web/controllers/AccessibilityCheckApi.cs#L331)):*

Looking at the code a little more, if the problem is with the "root -g" params, what can we do to move away from that?

As I understand, all we're trying to do here is verify that npm is available?
How about `npm --version`?
</blockquote></details>

Oh, no. I see we actually need to find the global node_modules directory.

andrew-polk:
*[`src/BloomExe/web/controllers/AccessibilityCheckApi.cs` line 331 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6130#-NfrSSI_5WobrmreX9D-:-Nfvwi39DL0pVu4X2Gtl:bre0gdp) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/54384e26a3089c93e56eefa166f1108d583eeed3/src/BloomExe/web/controllers/AccessibilityCheckApi.cs#L331)):*

Oh, no. I see we actually need to find the global node_modules directory.
</blockquote></details>

When I run `"C:\Program Files\Volta\npm.cmd"` root -g from the command line, I don't see a memory leak or spawning of extra processes.
Definitely seems we need to understand this more.

gmartin7 (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6130)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @andrew-polk)

___
*[`src/BloomExe/web/controllers/AccessibilityCheckApi.cs` line 331 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6130#-NfrSSI_5WobrmreX9D-:-NfvyDLzBoz6T63fobqc:b-d3779g) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/54384e26a3089c93e56eefa166f1108d583eeed3/src/BloomExe/web/controllers/AccessibilityCheckApi.cs#L331)):*

When I run `"C:\Program Files\Volta\npm.cmd"` root -g from the command line, I don't see a memory leak or spawning of extra processes.
Definitely seems we need to understand this more.
</blockquote></details>

I do... if you start in the Volta directory. And the Command Line Runner takes a start directory that would default to Volta, if you're using that version of npm.cmd.

andrew-polk:
*[`src/BloomExe/web/controllers/AccessibilityCheckApi.cs` line 331 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6130#-NfrSSI_5WobrmreX9D-:-Nfvzt_vDQUEEhoWhBC0:b-7opfv4) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/54384e26a3089c93e56eefa166f1108d583eeed3/src/BloomExe/web/controllers/AccessibilityCheckApi.cs#L331)):*

I do... if you start in the Volta directory. And the Command Line Runner takes a start directory that would default to Volta, if you're using that version of npm.cmd.
</blockquote></details>

Ah. I see that now, yes. Good catch.

gmartin7 (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6130)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @andrew-polk)

___
*[`src/BloomExe/web/controllers/AccessibilityCheckApi.cs` line 331 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6130#-NfrSSI_5WobrmreX9D-:-Nfw0uTU1tddx2zgfOgd:b435p17) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/54384e26a3089c93e56eefa166f1108d583eeed3/src/BloomExe/web/controllers/AccessibilityCheckApi.cs#L331)):*

I find that even running `npm.cmd --version` causes the same problem **if** run in the volta directory.

Let's weigh the pros and cons of your solution (excluding volta) vs. not running the volta directory:
```
var result = CommandLineRunnerExtra.RunWithInvariantCulture(fullNpmPath, "root -g", Encoding.ASCII, "", 10,
				new NullProgress());
```
Probably worth a team discussion, either in person on on Zulip.
</blockquote></details>

Wouldn't I still need to find the other location, just to get the directory we DO want to run from?

gmartin7 (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6130)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @andrew-polk)

___
*[`src/BloomExe/web/controllers/AccessibilityCheckApi.cs` line 331 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6130#-NfrSSI_5WobrmreX9D-:-NfwHBIr29SjNlNjw4PJ:b-eesk76) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/54384e26a3089c93e56eefa166f1108d583eeed3/src/BloomExe/web/controllers/AccessibilityCheckApi.cs#L331)):*

Wouldn't I still need to find the other location, just to get the directory we DO want to run from?
</blockquote></details>

And it's only a problem on developer machines. Nobody else will have Volta.

andrew-polk (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6130)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @gmartin7)

PR #6133, opened 2023-10-05
[5.5] BL-12198 MXB XMatter tweaks

gmartin7:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6133)

### DIFF
--- a/src/content/templates/xMatter/project-specific/MXBLiteracy-XMatter/MXBCommon-XMatter.less
+++ b/src/content/templates/xMatter/project-specific/MXBLiteracy-XMatter/MXBCommon-XMatter.less
@@ -1,6 +1,9 @@
 @import "../../bloom-xmatter-common.less";
 
 @fontSize: 10pt;
+@creditsPageFontSize: 8pt;
+@trimmerLineHeight: 1.1;
+@creditsLineHeight: 1.2;
 @MarginBetweenMajorItems: 3em;
 @contentMinimumHeight: 1em;
 
@@ -35,7 +38,7 @@
 .titlePage {
     // default values for the Languages of Book section; yeah, I know, huh? Credits-Page-style on the Title page!
     .Credits-Page-style .languagesOfBook {
-        line-height: 1.1;
+        line-height: @trimmerLineHeight;
     }
     // default values for the Original Contributions section
     .OriginalContributions-style.bloom-contentNational1 {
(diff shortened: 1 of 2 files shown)


andrew-polk (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6133)* status: all files reviewed, 1 unresolved discussion (waiting on @gmartin7)

___
*[`src/content/templates/xMatter/project-specific/MXBLiteracy-XMatter/MXBCommon-XMatter.less` line 85 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6133#-Ng5wA0Z-TGBQelfMmnC:-Ng5wA0Z-TGBQelfMmnD:bo4deli) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/aebed6d0f926d2114bf456c4a5cd5b6d303e782a/src/content/templates/xMatter/project-specific/MXBLiteracy-XMatter/MXBCommon-XMatter.less#L85)):*
> ```Less
>     }
>     #originalAcknowledgments {
>         margin-top: @creditsPageFontSize; // space between acknowledgments and the copyright block
> ```

Seems odd to me to set a margin to match a font size, especially one which is `pt`, not `px`, but if you're explicitly trying to make the spacing to match the text, I guess that makes some sense.
Though, won't the text height actually be 1.1em, so if you want to make the space the same as a line of text, it would be better to do that?
Anyway, I guess it is probably worth capturing the thinking in a comment.

gmartin7 (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6133)* status: 1 of 2 files reviewed, 1 unresolved discussion (waiting on @andrew-polk)

___
*[`src/content/templates/xMatter/project-specific/MXBLiteracy-XMatter/MXBCommon-XMatter.less` line 85 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6133#-Ng5wA0Z-TGBQelfMmnC:-NgPnKZ34XPXz6hz5NSJ:b-1ukmq9) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/aebed6d0f926d2114bf456c4a5cd5b6d303e782a/src/content/templates/xMatter/project-specific/MXBLiteracy-XMatter/MXBCommon-XMatter.less#L85)):*

Seems odd to me to set a margin to match a font size, especially one which is `pt`, not `px`, but if you're explicitly trying to make the spacing to match the text, I guess that makes some sense.
Though, won't the text height actually be 1.1em, so if you want to make the space the same as a line of text, it would be better to do that?
Anyway, I guess it is probably worth capturing the thinking in a comment.
</blockquote></details>

Added comments and used `calc()` to make things explicit.

andrew-polk (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6133)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @gmartin7)

PR #6158, opened 2023-10-19
BL-12403 Progress spacing

gmartin7:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6158)

### DIFF
--- a/src/BloomBrowserUI/booksOnBlorg/BooksOnBlorgProgressBar.tsx
+++ b/src/BloomBrowserUI/booksOnBlorg/BooksOnBlorgProgressBar.tsx
@@ -134,6 +134,7 @@ export const BooksOnBlorgProgressBar: React.FunctionComponent = () => {
                             font-size: 12px;
                             color: ${kBloomBlue};
                             margin-top: 2px;
+                            margin-bottom: 2px;
                         `}
                     >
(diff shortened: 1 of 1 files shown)


andrew-polk (changes requested):

gmartin7 (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6158)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @andrew-polk)

___
*[`src/BloomBrowserUI/react_components/Progress/ProgressBar.tsx` line 22 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6158#gh-1365891061:-Nh817aC9Dq6idZzx9u6:b-owists) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/91bdbe2108734bc2a89f25cc03cd861ca41c64ce/src/BloomBrowserUI/react_components/Progress/ProgressBar.tsx#L22)):*

One thing I remember in the css course which I thinks makes a lot of sense is that you don't want to define external margins on a component. Because it may or may not make sense for every usage of the component to have that spacing.

So I think we want to put this gap in the caller.
</blockquote></details>

You are right. Done.

andrew-polk (approved):

PR #6166, opened 2023-10-24
BL-12491 fix old publish tab content flashing when reentering tab

nabalone:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6166)

### DIFF
--- a/src/BloomBrowserUI/publish/PublishTab/PublishTabPane.tsx
+++ b/src/BloomBrowserUI/publish/PublishTab/PublishTabPane.tsx
@@ -117,6 +117,10 @@ export const PublishTabPane: React.FunctionComponent<{}> = () => {
     useSubscribeToWebSocketForEvent("publish", "switchToPublishTab", () => {
         setup();
     });
+    // User is switching out of publish tab, clear the display so the the old stuff doesn't flash when the user comes back on another book
+    useSubscribeToWebSocketForEvent("publish", "switchOutOfPublishTab", () => {
+        setPublishTabReady(false);
+    });
     React.useEffect(() => {
         // While the top bar is still in winforms, the first time the user loads the publish tab, the websocket event may occur before the component is ready
         setup();
--- a/src/BloomExe/Publish/PublishView.cs
+++ b/src/BloomExe/Publish/PublishView.cs
@@ -81,6 +81,7 @@ namespace Bloom.Publish
 			// TODO-WV2: Can we clear the cache for WV2? Do we need to?
 			PublishHelper.Cancel();
 			PublishHelper.InPublishTab = false;
+			_webSocketServer.SendEvent("publish", "switchOutOfPublishTab");
 		}
 
 		private void Activate()


andrew-polk (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6166)* status: all files reviewed, 1 unresolved discussion (waiting on @nabalone)

___
*[`src/BloomExe/Publish/PublishView.cs` line 84 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6166#-NhXhlJt8pw5iXR_meKI:-NhXhlJt8pw5iXR_meKJ:b-9158p) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/a81ee26308a6eb5bf711b404db9162889b69ec3d/src/BloomExe/Publish/PublishView.cs#L84)):*
> ```Smalltalk
> 			PublishHelper.Cancel();
> 			PublishHelper.InPublishTab = false;
> 			_webSocketServer.SendEvent("publish", "switchOutOfPublishTab");
> ```

This will work. 

But I'm trying to think how we would implement this if we had no PublishView (which hopefully we won't soon).

What solution can we do on just the web side?
I haven't thought of anything yet.

nabalone (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6166)* status: all files reviewed, 1 unresolved discussion (waiting on @andrew-polk)

___
*[`src/BloomExe/Publish/PublishView.cs` line 84 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6166#-NhXhlJt8pw5iXR_meKI:-NhXjC--08zCWRDvJejL:bol82hm) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/a81ee26308a6eb5bf711b404db9162889b69ec3d/src/BloomExe/Publish/PublishView.cs#L84)):*

I was thinking that when there is no PublishView, there would be a parent component which is the top bar and controls the switching between tabs, is that correct? And then maybe it would have a state variable or something like that corresponding to which tab we are in and we could just use that instead?
</blockquote></details>

Am I missing something

andrew-polk (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6166)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @nabalone)

___
*[`src/BloomExe/Publish/PublishView.cs` line 84 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6166#-NhXhlJt8pw5iXR_meKI:-NhXjg0N53zuTTfr6Pes:b-uptjpy) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/a81ee26308a6eb5bf711b404db9162889b69ec3d/src/BloomExe/Publish/PublishView.cs#L84)):*

Am I missing something
</blockquote></details>

No, that's right. 
At that point, the whole publish tab will be removed from the dom.
So, this is ok for now.

PR #6177, opened 2023-10-30
[5.6] BL-12811 Fix out-of-date bloomPUB message

gmartin7:
* replaced "so text might not fit" with "which may cause text to scroll"

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6177)

### DIFF
--- a/DistFiles/localization/en/Bloom.xlf
+++ b/DistFiles/localization/en/Bloom.xlf
@@ -3901,7 +3901,7 @@
         <note>Button that tells Bloom to begin offering this book on the wifi network.</note>
       </trans-unit>
       <trans-unit id="PublishTab.Android.WrongLayout.Message">
-        <source xml:lang="en">The layout of this book is currently "{0}". Bloom Reader will display it using "{1}", so text might not fit. To see if anything needs adjusting, go back to the Edit Tab and change the layout to "{1}".</source>
+        <source xml:lang="en">The layout of this book is currently "{0}". Bloom Reader will display it using "{1}", which may cause text to scroll. To see if anything needs adjusting, go back to the Edit Tab and change the layout to "{1}".</source>
         <note>ID: PublishTab.Android.WrongLayout.Message</note>
         <note>{0} and {1} are book layout tags.</note>
       </trans-unit>
--- a/src/BloomExe/Publish/BloomPub/PublishToBloomPubApi.cs
+++ b/src/BloomExe/Publish/BloomPub/PublishToBloomPubApi.cs
@@ -304,7 +304,7 @@ namespace Bloom.Publish.BloomPub
 //					"Note", "A heading shown above some messages.");
 //				progress.MessageWithoutLocalizing(msgFormat, ProgressKind.Note);
 				 var msgFormat = L10NSharp.LocalizationManager.GetString("PublishTab.Android.WrongLayout.Message",
-					"The layout of this book is currently \"{0}\". Bloom Reader will display it using \"{1}\", so text might not fit. To see if anything needs adjusting, go back to the Edit Tab and change the layout to \"{1}\".",
+					"The layout of this book is currently \"{0}\". Bloom Reader will display it using \"{1}\", which may cause text to scroll. To see if anything needs adjusting, go back to the Edit Tab and change the layout to \"{1}\".",
 					"{0} and {1} are book layout tags.");
 				var desiredLayout = desiredLayoutSize + layout.SizeAndOrientation.OrientationName;
 				var msg = String.Format(msgFormat, layout.SizeAndOrientation.ToString(), desiredLayout, Environment.NewLine);


andrew-polk (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6177)* status: all files reviewed, 1 unresolved discussion (waiting on @gmartin7)

___
*[`DistFiles/localization/en/Bloom.xlf` line 3904 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6177#-Ni1T7HpDgWqMOWN-mUL:-Ni1T7HpDgWqMOWN-mUM:b-4nio8p) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/e018250db4292ce5ac2e0e07e19b5300172b0d8d/DistFiles/localization/en/Bloom.xlf#L3904)):*
> ```XML
>       </trans-unit>
>       <trans-unit id="PublishTab.Android.WrongLayout.Message">
>         <source xml:lang="en">The layout of this book is currently "{0}". Bloom Reader will display it using "{1}", which may cause text to scroll. To see if anything needs adjusting, go back to the Edit Tab and change the layout to "{1}".</source>
> ```

So... unfortunately, our process doesn't really allow for simply changing a string in the xlf file.
It will invalidate all translations and the next time we sync, we lose all the existing ones.

So we typically have two options.
1. Do what you've done, then manually update translations in Crowdin.
2. Create a second version of the string (with different ID) and make the old one obsolete.

However, since 5.5 is also obsolete wording and we won't sync back further than that now, I think maybe we just change it there?
That will still lose translations until translators retranslate, but at least it gives them a chance to retranslate and we can sync again?
I don't know. Nothing is ideal, but that seems like the best bad option for this.

So I'm proposing we do what you've done, but on 5.5.
Open to others' thoughts, though.

@StephenMcConnel

gmartin7 (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6177)* status: all files reviewed, 1 unresolved discussion (waiting on @StephenMcConnel)

___
*[`DistFiles/localization/en/Bloom.xlf` line 3904 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6177#-Ni1T7HpDgWqMOWN-mUL:-Ni1_OsN6jxUYMoP9owj:btclehs) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/e018250db4292ce5ac2e0e07e19b5300172b0d8d/DistFiles/localization/en/Bloom.xlf#L3904)):*

So... unfortunately, our process doesn't really allow for simply changing a string in the xlf file.
It will invalidate all translations and the next time we sync, we lose all the existing ones.

So we typically have two options.
1. Do what you've done, then manually update translations in Crowdin.
2. Create a second version of the string (with different ID) and make the old one obsolete.

However, since 5.5 is also obsolete wording and we won't sync back further than that now, I think maybe we just change it there?
That will still lose translations until translators retranslate, but at least it gives them a chance to retranslate and we can sync again?
I don't know. Nothing is ideal, but that seems like the best bad option for this.

So I'm proposing we do what you've done, but on 5.5.
Open to others' thoughts, though.

@StephenMcConnel
</blockquote></details>

Steve said: I don't think I have anything to add to what Andrew said. If it's only changes to xlf files, however, they may as well be done on master because that's the only branch that crowdin responds to. If there are code changes, then those need to be on 5.5, and the xlf change could be there and just wait to be merged into master before crowdin would see it.

StephenMcConnel (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6177)* status: all files reviewed, 1 unresolved discussion (waiting on @andrew-polk)

___
*[`DistFiles/localization/en/Bloom.xlf` line 3904 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6177#-Ni1T7HpDgWqMOWN-mUL:-Ni1_KCe7WiV8lzkBZxH:b-mme7sy) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/e018250db4292ce5ac2e0e07e19b5300172b0d8d/DistFiles/localization/en/Bloom.xlf#L3904)):*

Steve said: I don't think I have anything to add to what Andrew said. If it's only changes to xlf files, however, they may as well be done on master because that's the only branch that crowdin responds to. If there are code changes, then those need to be on 5.5, and the xlf change could be there and just wait to be merged into master before crowdin would see it.
</blockquote></details>

If the PR changed only the XLF file, then it could be done on master.  But since there's a code change as well, it should be done on Version5.5 or Version5.6. 
If the change has already been made on Version5.6, it might be simpler to finish it there and cherry-pick back to Version5.5.  Rebasing the current branch won't work in all likelihood.

andrew-polk:
Ok, I'll merge. Then please cherry-pick to 5.5.

PR #6178, opened 2023-10-30
Don't delete image file when shrunk unless it's temporary (BL-12779)

StephenMcConnel:


andrew-polk:
The fix looks good, but we need to target at least back to 5.6.
However, this is a data-loss bug with a pretty clean solution; I propose we target 5.5.

StephenMcConnel (commented):
I'm not sure retargeting this branch would work cleanly backwards to 5.5.  I'll try, but I may have to recreate the PR.

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6178#-:-Ni1VjOm02rRMmfIrB-8:b-wsmzze)* status: 0 of 1 files reviewed, all discussions resolved

StephenMcConnel (commented):
I've rebased back to Version5.6 (which is what I intended to begin with).  Rebasing to Version5.5 does not work, but cherry-picking back to Version5.5 does.  I suggest that we accept and merge this PR if it looks good, and then cherry-pick it back to Version5.5.

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6178#-:-Ni1YCxL28399ZzNS7rT:b-vpdzyr)* status: 0 of 1 files reviewed, all discussions resolved

PR #6196, opened 2023-11-02
BL-5482 Report case errors in local server requests

gmartin7:
* I actually found one in testing (see PublishTabPane)

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6196)

### DIFF
--- a/src/BloomBrowserUI/publish/PublishTab/PublishTabPane.tsx
+++ b/src/BloomBrowserUI/publish/PublishTab/PublishTabPane.tsx
@@ -254,7 +254,7 @@ export const PublishTabPane: React.FunctionComponent<{}> = () => {
                                                 "PublishTab.bloomPUBButton-tooltip"
                                         }}
                                     >
-                                        <img src="/bloom/publish/PublishTab/bloomPUB.png" />
+                                        <img src="/bloom/publish/PublishTab/BloomPUB.png" />
                                         <Span
                                             l10nKey="PublishTab.bloomPUBButton"
                                             className="sidebar-tab-label"
--- a/src/BloomExe/web/BloomServer.cs
+++ b/src/BloomExe/web/BloomServer.cs
@@ -242,7 +242,7 @@ namespace Bloom.Api
 			{
 				pathToInMemoryHtmlFile = Path.Combine(dom.BaseForRelativePaths, simulatedPageFileName).Replace('\\', '/');
 			}
-			if (RobustFile.Exists(pathToInMemoryHtmlFile))
+			if (RobustFileExistsWithCaseCheck(pathToInMemoryHtmlFile))
 			{
 				// Just in case someone perversely calls a book "currentPage" we will use another name.
 				// (We want one that does NOT conflict with anything really in the folder.)
@@ -411,7 +411,7 @@ namespace Bloom.Api
 					// read in current defaultLangStyles.css content, add @font-face info to it if necessary.
 					var cssLangStyles = "";
 					var cssFilePath = Path.Combine(CurrentBook.FolderPath, "defaultLangStyles.css");
-					if (RobustFile.Exists(cssFilePath))
+					if (RobustFileExistsWithCaseCheck(cssFilePath))
 						cssLangStyles = RobustFile.ReadAllText(cssFilePath);
 					var serve = FontServe.GetInstance();
 					var fontFaceDeclarations = serve.GetAllFontFaceDeclarations();
@@ -560,7 +560,7 @@ namespace Bloom.Api
 				processImage = false;
 				imageFile = imageFile.Substring((OriginalImageMarker + "/").Length);
 
-				if (!RobustFile.Exists(imageFile))
+				if (!RobustFileExistsWithCaseCheck(imageFile))
 				{
 					// We didn't find the file here, and don't want to use the following else if or we could errantly
 					// find it in the browser root. For example, this outer if (imageFile.StartsWith...) was added because
@@ -568,7 +568,7 @@ namespace Bloom.Api
 					return false;
 				}
 			}
-			else if (!RobustFile.Exists(imageFile))
+			else if (!RobustFileExistsWithCaseCheck(imageFile))
 			{
 				// Generally, the path we started with will only work when the HTML file is the root file of a book,
 				// or another file (other than preview) that is simulated to be in the book folder,
@@ -590,7 +590,7 @@ namespace Bloom.Api
 
 				imageFile = Path.Combine(sourceDir, imageFile);
 
-				if (!RobustFile.Exists(imageFile))
+				if (!RobustFileExistsWithCaseCheck(imageFile))
 				{
 					// There are a few special cases where it's not desirable to change the source of the image
 					// in our source code.
@@ -614,7 +614,7 @@ namespace Bloom.Api
 						imageFile = Path.Combine(bloomRoot, "images/widget-placeholder.svg");
 					}
 
-					if (!RobustFile.Exists(imageFile))
+					if (!RobustFileExistsWithCaseCheck(imageFile))
 					{
 						if (sourceDir != CurrentBook?.FolderPath)
 						{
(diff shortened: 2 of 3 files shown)


JohnThomson (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6196)* status: all files reviewed, 5 unresolved discussions (waiting on @gmartin7)

*[a discussion](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6196#-NiL5dZo87NC1LTToiWU:-NiL5dZo87NC1LTToiWV:b-ca63pa) (no related file):*
Is it any use doing this other than in a debug build? Seems highly unlikely that anyone will notice such warnings in an end user's log file. And if we encounter a file case error in a debug build, seems like we should do something much more likely to be noticed, like an Assert.

___
___
*[`src/BloomExe/web/BloomServer.cs` line 1692 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6196#-NiL-eD5Dk5QA73Oft91:-NiL-eD5Dk5QA73Oft92:bcgtma2) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/246b8f231c9dd2225d4e527ac87bcb065994f340/src/BloomExe/web/BloomServer.cs#L1692)):*
> ```Smalltalk
> 			// AppData is for Windows and /tmp/ is for Linux (when we use it again)
> 			if (localPath.EndsWith(".htm") && (localPath.Contains("AppData") || localPath.Contains("/tmp/")))
> 				return true; // probably the Temp folder and most likely a random temporary filename).
> ```

This will typically fail to check any HTM files shipped with Bloom (since it is typically installed in .../AppData/Local/... If that's what you intend, the comment does not indicate it.
___
*[`src/BloomExe/web/BloomServer.cs` line 1699 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6196#-NiL0vTF2sInDAQlbkG9:-NiL0vTF2sInDAQlbkGA:b-vjn6i0) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/246b8f231c9dd2225d4e527ac87bcb065994f340/src/BloomExe/web/BloomServer.cs#L1699)):*
> ```Smalltalk
> 
> 			string directoryPath = Path.GetDirectoryName(fullPath);
> 			if (!Array.Exists(Directory.GetFiles(directoryPath), s => CompareCaseOfTwoPaths(s, localPath)))
> ```

How about using Directory.EnumerateFiles(directoryPath, Path.GetFileName(fullPath))? That should match the file we want but very few others, usually only the exact one, so it seems likely to be faster...though maybe that should be tested.
(Note that if we're not looking for an HTM file and it IS in the local temp directory, it could have a LOT of files to consider.)
Did you consider the various possibilities in https://stackoverflow.com/questions/325931/getting-actual-file-name-with-proper-casing-on-windows-with-net?
Also, it might be helpful to apply Path.GetFullPath() to localPath. For example, it's not a case error if somepath/sub/../x.y is is enumerated as somepath/x.y. Not sure whether we would ever do that, but I'm not sure we don't, either.
___
*[`src/BloomExe/web/BloomServer.cs` line 1713 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6196#-NiL8TGAE8UUE62LC3rN:-NiL8TGAE8UUE62LC3rO:bbfq46g) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/246b8f231c9dd2225d4e527ac87bcb065994f340/src/BloomExe/web/BloomServer.cs#L1713)):*
> ```Smalltalk
> 		}
> 
> 		private static bool CompareCaseOfTwoPaths(string path1, string path2)
> ```

It's actually comparing more than the case. EqualsExceptForDirectorySeps? EqualsIgnoringDirectorySeparators?
___
*[`src/BloomTests/web/BloomServerTests.cs` line 790 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6196#-NiL4sDWEPDt1HvJObIB:-NiL4sDWEPDt1HvJObIC:b-ga0lao) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/246b8f231c9dd2225d4e527ac87bcb065994f340/src/BloomTests/web/BloomServerTests.cs#L790)):*
> ```Smalltalk
> 		[TestCase("customcollectionstyles.css", true, true)]
> 		[TestCase("customCollectionStyles.css", true, false)]
> 		[TestCase("xxyyzz.css", false, false)]
> ```

Let's have some test cases with various directory separators. Also, there should be a test where the case error is in the directory path, and probably one with something like /subdir/../
We probably can't make test cases for this, but did you test (or at least consider) other cases where there's more than one name for something? For example, a folder can be mapped to a drive letter and I have a vague notion that there may be more than one way to give a path to a network drive. Not sure that GetFiles will (or won't) use the same one for its results as you gave it...maybe GetFullName should be used BEFORE you pass the name to GetFiles or EnumerateFiles.

gmartin7 (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6196)* status: 1 of 3 files reviewed, 5 unresolved discussions (waiting on @JohnThomson)

*[a discussion](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6196#-NiL5dZo87NC1LTToiWU:-NiLMzgi6q4HtwTBiz-d:b-h6lbsb) (no related file):*

Is it any use doing this other than in a debug build? Seems highly unlikely that anyone will notice such warnings in an end user's log file. And if we encounter a file case error in a debug build, seems like we should do something much more likely to be noticed, like an Assert.
</blockquote></details>

Good point. I've put the case-checking part in an `#if DEBUG` section. Then I'll do `Debug.Fail()` for a failed case match, unless running tests, in which case I'll do the `Logger` message (easier to test than an Assert).

___
___
*[`src/BloomExe/web/BloomServer.cs` line 1692 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6196#-NiL-eD5Dk5QA73Oft91:-NiLMn-s7gGC_W3s6Eqb:bo1ct6i) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/246b8f231c9dd2225d4e527ac87bcb065994f340/src/BloomExe/web/BloomServer.cs#L1692)):*

This will typically fail to check any HTM files shipped with Bloom (since it is typically installed in .../AppData/Local/... If that's what you intend, the comment does not indicate it.
</blockquote></details>

OTOH, since I'm moving to only doing the case check in Debug mode, we're good. I made a comment to that effect.
___
*[`src/BloomExe/web/BloomServer.cs` line 1699 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6196#-NiL0vTF2sInDAQlbkG9:-NiLBxU-E_JZY6qtDlb2:b-2n5a3r) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/246b8f231c9dd2225d4e527ac87bcb065994f340/src/BloomExe/web/BloomServer.cs#L1699)):*
> (Note that if we're not looking for an HTM file and it IS in the local temp directory, it could have a LOT of > files to consider.)

Specifically what I was trying to avoid with the above `if`. :(

> Also, it might be helpful to apply Path.GetFullPath() to localPath.

Done.

Actually, this is quite promising:

> Did you consider the various possibilities in [https://stackoverflow.com/questions/325931/getting-actual-file-name-with-proper-casing-on-windows-with-net](https://stackoverflow.com/questions/325931/getting-actual-file-name-with-proper-casing-on-windows-with-net)?

I think that works well and doesn't involve weeding through a lot of extra files to find what we want.
___
*[`src/BloomExe/web/BloomServer.cs` line 1713 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6196#-NiL8TGAE8UUE62LC3rN:-NiLLP_30gY4La_XXZFj:b5tpkv4) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/246b8f231c9dd2225d4e527ac87bcb065994f340/src/BloomExe/web/BloomServer.cs#L1713)):*

It's actually comparing more than the case. EqualsExceptForDirectorySeps? EqualsIgnoringDirectorySeparators?
</blockquote></details>

How about `EqualsIgnoringCaseAndDirectorySeps`?
___
*[`src/BloomTests/web/BloomServerTests.cs` line 790 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6196#-NiL4sDWEPDt1HvJObIB:-NiM65u66w0tNUL3wDI5:b-86j5se) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/246b8f231c9dd2225d4e527ac87bcb065994f340/src/BloomTests/web/BloomServerTests.cs#L790)):*
> Let's have some test cases with various directory separators. Also, there should be a test where the case error is in the directory path, and probably one with something like /subdir/../

Done.

> maybe GetFullName should be used BEFORE you pass the name to GetFiles or EnumerateFiles.

Done, but we went with a better method that doesn't enumerate (at least at this level).

JohnThomson (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6196)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @gmartin7)

PR #6204, opened 2023-11-06
Make aeneas more robust by running in the temp folder (BL-12801)

StephenMcConnel:
This presumably avoids file sync programs which should ignore files in the temp folder.

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6204)

### DIFF
--- a/src/BloomExe/web/controllers/AudioSegmentationApi.cs
+++ b/src/BloomExe/web/controllers/AudioSegmentationApi.cs
@@ -3,6 +3,7 @@ using Bloom.Book;
 using Bloom.Edit;
 using Bloom.ToPalaso;
 using Bloom.Utils;
+using BloomTemp;
 using Newtonsoft.Json;
 using SIL.Code;
 using SIL.IO;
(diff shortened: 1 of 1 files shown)


andrew-polk (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6204)* status: all files reviewed, 1 unresolved discussion (waiting on @StephenMcConnel)

___
*[`src/BloomExe/web/controllers/AudioSegmentationApi.cs` line 604 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6204#-Nib8UjI3EWPjpqi1nMS:-Nib8UjI3EWPjpqi1nMT:b7oqj1b) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/aa2cc825cd1a85cdbc79ee40a1b53d211749c587/src/BloomExe/web/controllers/AudioSegmentationApi.cs#L604)):*
> ```Smalltalk
> 				// inability to cope with non-ascii characters in file pathnames passed in on the command line by using bare filenames in the
> 				// command.  See https://issues.bloomlibrary.org/youtrack/issue/BL-6927.  Working in the %TEMP% folder also has the benefit of
> 				// avoiding virus checker interference.  See https://issues.bloomlibrary.org/youtrack/issue/BL-12801.
> ```

Working in the %TEMP% folder also has the benefit of
				// avoiding virus checker interference


Is it true that it less likely to be scanned by AV? 

JH's point in the issue is that file sync programs (like Dropbox) would not mess with it.

andrew-polk (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6204)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @StephenMcConnel)

PR #6206, opened 2023-11-07
BL-12805 Use text-wrap property

gmartin7:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6206)

### DIFF
--- a/src/content/bookLayout/basePage-sharedRules.less
+++ b/src/content/bookLayout/basePage-sharedRules.less
@@ -151,3 +151,15 @@ span.bloom-linebreak {
     display: block;
     text-indent: 0;
 }
+
+// Text wrapping algorithms.
+.bloom-page .bloom-editable.bloom-visibility-code-on {
+    // Very long words should just allow themselves to be broken at the end of a line if they're
+    // longer than a line.  See BL-12330.
+    overflow-wrap: break-word;
+    text-wrap: pretty;
+    &.Title-On-Cover-style,
+    &.Cover-Default-style {
+        text-wrap: balance;
+    }
+}
--- a/src/content/bookLayout/basePage.less
+++ b/src/content/bookLayout/basePage.less
@@ -1036,11 +1036,6 @@ div[data-book*="branding"] {
     display: none;
 }
 
-// Very long words should just allow themselves to be broken at the end of a line if they're
-// longer than a line.  See BL-12330.
-.bloom-page .bloom-editable.bloom-visibility-code-on {
-    word-wrap: break-word;
-}
 //... to be continued. We sandwich most other sheets between this one and the langVisibility.css, which comes last-ish
 
 // now apply any book features


JohnThomson (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6206)* status: all files reviewed, 2 unresolved discussions (waiting on @gmartin7)

___
*[`src/content/bookLayout/basePage.less` line 1041 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6206#-Nijsb4z-kbZflF7p1Z1:-Nijsb4z-kbZflF7p1Z2:bhtjeco) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/b778268fa67b3e5155c8d3cee59f8f766d6007f8/src/content/bookLayout/basePage.less#L1041)):*
> ```Less
> // Very long words should just allow themselves to be broken at the end of a line if they're
> // longer than a line.  See BL-12330.
> // Review: Should this rule be moved to basePage-sharedRules.less so it applies to ePUBs too?
> ```

Seems desirable.  I suggest you try it and see whether Daisy complains. It might be good to use word-wrap as well, in case some epub reader uses a very old browser; then again, this rule wasn't in epubs at all before now, so it can only improve things as far as I can see.
It's somewhat puzzling that the original rule was NOT in shared...
___
*[`src/content/bookLayout/basePage.less` line 1043 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6206#-Nijtdod3l98eSuOimXC:-Nijtdod3l98eSuOimXD:b2p05f3) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/b778268fa67b3e5155c8d3cee59f8f766d6007f8/src/content/bookLayout/basePage.less#L1043)):*
> ```Less
> // Review: Should this rule be moved to basePage-sharedRules.less so it applies to ePUBs too?
> .bloom-page .bloom-editable.bloom-visibility-code-on {
>     overflow-wrap: break-word; // 'overflow-wrap' is now the standard name for 'word-wrap'
> ```

Doesn't seem as if we need the comment here (it explains the PR rather than the current code)

gmartin7 (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6206)* status: 0 of 2 files reviewed, 2 unresolved discussions (waiting on @JohnThomson)

___
*[`src/content/bookLayout/basePage.less` line 1041 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6206#-Nijsb4z-kbZflF7p1Z1:-Nil-dgKBC15d07FuSMx:bhvcphu) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/b778268fa67b3e5155c8d3cee59f8f766d6007f8/src/content/bookLayout/basePage.less#L1041)):*
Moved. Compiled. Made sure Daisy was installed. Ran it on my book. 4 minor/1 moderate complaints about ARIA stuff, 1 serious complaint about epub-lang. None seem related to this.

> It's somewhat puzzling that the original rule was NOT in shared...

I agree.
___
*[`src/content/bookLayout/basePage.less` line 1043 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6206#-Nijtdod3l98eSuOimXC:-Nikn0j-9O7gSOs7hmul:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/b778268fa67b3e5155c8d3cee59f8f766d6007f8/src/content/bookLayout/basePage.less#L1043)):*

Doesn't seem as if we need the comment here (it explains the PR rather than the current code)
</blockquote></details>

Done.

gmartin7 (commented):
As noted in the issue, JH has decided we'll live with the non-WYSIWYG-ness for now.

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6206#-:-NitxUB61nsHnnuvo4Fm:b-sz16ae)* status: 0 of 2 files reviewed, 2 unresolved discussions (waiting on @JohnThomson)

JohnThomson (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6206)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @gmartin7)

PR #6221, opened 2023-11-15
Fix a crash initializing L10NSharp at startup (20231115)

StephenMcConnel:


andrew-polk:
This is apparently new behavior with sdk 8.
According to [this](https://learn.microsoft.com/en-us/dotnet/core/compatibility/sdk/8.0/source-link), we should set "the IncludeSourceRevisionInInformationalVersion property to false in [the] project file."

StephenMcConnel (commented):
I've changed the fix to use the project property in the csproj file.

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6221#-:-NjJPAMnB_yn21rKhJ44:b-41ez8i)* status: 0 of 1 files reviewed, all discussions resolved

andrew-polk (approved):

PR #6224, opened 2023-11-16
Fix pasting (from Word) to remove useless spans (BL-12861)

StephenMcConnel:
This might be simple enough to cherry-pick back to 5.5 if we think it's robust enough.

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6224)

### DIFF
--- a/src/BloomBrowserUI/bookEdit/bloomField/BloomField.ts
+++ b/src/BloomBrowserUI/bookEdit/bloomField/BloomField.ts
@@ -150,6 +150,28 @@ export default class BloomField {
         return input;
     }
 
+    public static removeUselessSpanMarkup(input: string): string {
+        // Paste data from Word can have span elements with no attributes.
+        // These serve no purpose and can cause problems in Bloom audio
+        // highlighting after splitting with aeneas.  We remove them here.
+        // See BL-12861.
+        if (!input.includes("<span>")) return input;
+        const newDiv = document.createElement("div");
+        newDiv.innerHTML = input;
+        const spans = newDiv.getElementsByTagName("span");
+        for (let i = spans.length - 1; i >= 0; i--) {
+            const span = spans[i];
+            if (!span.hasAttributes()) {
+                const spanContent: (string | Node)[] = [];
+                span.childNodes.forEach(child => {
+                    spanContent.push(child);
+                });
+                span.replaceWith(...spanContent);
+            }
+        }
+        return newDiv.innerHTML;
+    }
+
     // This BloomField thing was done before ckeditor; ckeditor kinda expects to the only thing
     // taking responsibility for the field, so that creates problems. Eventually, we could probably
     // re-cast everything in this BloomField class as a plugin or at least callbacks to ckeditor.
@@ -181,6 +203,9 @@ export default class BloomField {
             );
 
             event.data.dataValue = this.fixPasteData(event.data.dataValue);
+            event.data.dataValue = this.removeUselessSpanMarkup(
+                event.data.dataValue
+            );
 
             // We can't just duplicate audio ids without running into trouble later!
             if (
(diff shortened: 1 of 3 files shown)


andrew-polk:
Do we understand what caused the regression?
ckeditor's config.js is supposed to control what we allow to be pasted in.
But this line didn't change between 5.4 and 5.5:
```
config.pasteFilter = 
        "p br em i strong sup u; b{font-weight}; a\[!href\]; span{font-variant};";
```

StephenMcConnel (commented):
My reading of the filter is that font-variant is the only style allowed on span, but it's not required and no attributes are required.

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6224#-:-NjO4nZAAHcH-_Ahn115:b-b1f0w0)* status: 0 of 2 files reviewed, all discussions resolved

StephenMcConnel (commented):
But there is a regression from 5.4 to 5.5 in displaying the audio splits which I don't understand.

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6224#-:-NjO64Yn-8xgB0Ls-6Ce:bhtjmb9)* status: 0 of 2 files reviewed, all discussions resolved

StephenMcConnel (commented):
Looking at the generated HTML with the audio splits, I don't understand why the display ever worked with the pasted text even in 5.4.

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6224#-:-NjO6NBW4yT8SjAhQYFT:b-sy5q6b)* status: 0 of 2 files reviewed, all discussions resolved (waiting on @andrew-polk)

StephenMcConnel (commented):
The actual pasting input is working the same as before, albeit WebView2 passes in much simpler HTML than Geckofx60 did, minus all the Word-specific cruft.  (which happened in 5.4 as well when using WebView2)
There were 24 commits with changes to audioRecording.ts since Version5.4, one of which must have changed a detail about how the split audio is applied for this edge case.  I haven't been able to determine which commit caused the problem.  Even when it was working, the useless span element was causing strange looking HTML that I don't understand how it was working properly.
I rewrote the fixup function to work on nodes/elements instead of relying on regular expressions handling only some valid inputs.

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6224#-:-NjP2aS2EAnbeEEQ00E6:b-dxpues)* status: 0 of 2 files reviewed, all discussions resolved (waiting on @andrew-polk)

JohnThomson (commented):
What we've done here is probably desirable; it cleans useless markup from our content.
However, I think it might also be worth improving the CSS. The idea is to alternate colors for every nth sentence, which really means the nth span that has class bloom-highlightSegment.
There are three rules, currently starting at line 84 of audioRecording.less, with selectors like
span:nth-child(3n + 1)

If I'm understanding nth-child correctly, I think we could change them to
span:nth-child(3n + 1 of .bloom-highlightSegment) 

and they would then ignore everything except things we specifically intended to treat as sentence splits.
This was not supported in Gecko60, but I think we can safely use it now.

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6224#-:-NlAfzEwBlC1Xeb-TWb4:b-ggh4k4)* status: 0 of 2 files reviewed, all discussions resolved (waiting on @andrew-polk)

JohnThomson (approved):
> Reviewed 2 of 2 files at r2, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6224)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @StephenMcConnel)

PR #6226, opened 2023-11-17
Bl-12758 book status to api

nabalone:
Hide Parse behind API: Book Status

Part of the larger project to replace direct calls to Parse with calls to azure fns, so that we reduce our lockin to Parse.

**Depends on https://github.com/BloomBooks/bloom-azure-functions/pull/68**

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6226)

### DIFF
--- a/src/BloomExe/WebLibraryIntegration/BloomLibraryBookApiClient.cs
+++ b/src/BloomExe/WebLibraryIntegration/BloomLibraryBookApiClient.cs
@@ -1,5 +1,6 @@
 using System;
 using System.Collections.Generic;
+using System.Linq;
 using System.Net;
 using System.Text;
 using System.Windows.Forms;
@@ -101,24 +102,6 @@ namespace Bloom.WebLibraryIntegration
 				SIL.Reporting.Logger.WriteEvent("upload-finish failed: " + response.Content);
 				throw new ApplicationException("Unable to finalize book upload on the server.");
 			}
-		}
-
-		/// <summary>
-		/// Get the number of books on bloomlibrary.org that are in the given language.
-		/// </summary>
-		/// <remarks>Query should get all books where the isoCode matches the given languageCode
-		/// and 'rebrand' is not true and 'inCirculation' is not false and 'draft' is not true.</remarks>
-		public string getBookCountByLanguage(string languageTag)
-		{
-			var request = MakeGetRequest("get-book-count-by-language");
-			request.AddQueryParameter("language-tag", languageTag);
-			var response = AzureRestClient.Execute(request);
-			if (response.StatusCode != HttpStatusCode.OK)
-			{
-				SIL.Reporting.Logger.WriteEvent("get-book-count-by-language failed: " + response.Content);
-				throw new ApplicationException("Unable to get book count for language " + languageTag);
-			}
-			return response.Content;
 		}
 
 		protected RestClient AzureRestClient
@@ -259,6 +242,33 @@ namespace Bloom.WebLibraryIntegration
 			return MakeParseRequest(path, Method.POST);
 		}
 
+		/// <summary>
+		/// Get the number of books on bloomlibrary.org that are in the given language.
+		/// </summary>
+		/// <remarks>Query should get all books where the isoCode matches the given languageCode
+		/// and 'rebrand' is not true and 'inCirculation' is not false and 'draft' is not true.</remarks>
+		public int GetBookCountByLanguage(string languageCode)
+		{
+			if (!UrlLookup.CheckGeneralInternetAvailability(false))
+				return -1;
+			var request = MakeGetRequest("get-book-count-by-language");
+			request.AddQueryParameter("language-tag", languageCode);
+			var response = AzureRestClient.Execute(request);
+			if (response.StatusCode != HttpStatusCode.OK)
+			{
+				SIL.Reporting.Logger.WriteEvent("get-book-count-by-language failed: " + response.Content);
+				return -1;
+			}
+			try {
+				return int.Parse(response.Content);
+			}
+			catch (Exception e)
+			{
+				SIL.Reporting.Logger.WriteEvent("get-book-count-by-language failed: " + e.Message);
+				return -1;
+			}
+		}
+
 		// Setting param 'includeLanguageInfo' to true adds a param to the query that causes it to fold in
 		// useful language information instead of only having the arcane langPointers object.
 		public IRestResponse GetBookRecordsByQuery(string query, bool includeLanguageInfo)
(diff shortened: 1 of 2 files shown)


andrew-polk (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6226)* status: all files reviewed, 4 unresolved discussions (waiting on @nabalone)

___
*[`src/BloomExe/WebLibraryIntegration/BloomLibraryBookApiClient.cs` line 247 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6226#-NjUGdpED473UiRfQtJ1:-NjUGdpED473UiRfQtJ2:b-yw8f5) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/d9e854bde4d8cfcf0af129b373cdabc2b6e4e1a7/src/BloomExe/WebLibraryIntegration/BloomLibraryBookApiClient.cs#L247)):*
> ```Smalltalk
> 		{
> 			if (!UrlLookup.CheckGeneralInternetAvailability(false))
> 				return -1;
> ```

I think we need to preserve this internet check at the beginning of GetBookCountByLanguage.
___
*[`src/BloomExe/WebLibraryIntegration/BloomLibraryBookApiClient.cs` line 261 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6226#-NjUFe0k0MjCJZ6Y5VlW:-NjUFe0l-E7LRk98HWPt:btka563) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/36dbdfb417c9c1e9ce601f053df5c81b87ce94eb/src/BloomExe/WebLibraryIntegration/BloomLibraryBookApiClient.cs#L261)):*
> ```Smalltalk
> 			}
> 			return response.Content;
> 		}
> ```

Unless I'm missing something, I think it would be better to keep this method returning an int.
You can parse the string to ensure it is an int, else return -1.

In fact, I think we want any problem in this method to just return -1 (rather than throw).
I could be persuaded differently on this point. In the future, one could foresee different callers needing to handle exceptions differently. But since we only have one caller at this point, I think I would go with simplicity.
___
*[`src/BloomExe/WebLibraryIntegration/BloomLibraryBookApiClient.cs` line 365 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6226#-NjUH2Ec8nSSldN0GIQ3:-NjUH2Ec8nSSldN0GIQ4:bux12ww) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/36dbdfb417c9c1e9ce601f053df5c81b87ce94eb/src/BloomExe/WebLibraryIntegration/BloomLibraryBookApiClient.cs#L365)):*
> ```Smalltalk
> 				
> 			List<string> bookIds = bookInfos.Select(book => book.Id).ToList();
> 			var request = MakeGetRequest("get-books-with-these-ids");
> ```

Naming is hard...
I think I would go with `get-books-with-ids` or `get-books-by-ids`. 

Possibly even `get-books`.
After all, if we wanted to allow for getting books by some other parameter, such as a query, that could all be handled at the parameter level.

Thinking more, I think I prefer `get-books`.
___
*[`src/BloomExe/WebLibraryIntegration/BloomLibraryBookApiClient.cs` line 366 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6226#-NjUHf2T8PeA9x1H7pT9:-NjUHf2T8PeA9x1H7pTA:bvuk9lj) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/36dbdfb417c9c1e9ce601f053df5c81b87ce94eb/src/BloomExe/WebLibraryIntegration/BloomLibraryBookApiClient.cs#L366)):*
> ```Smalltalk
> 			List<string> bookIds = bookInfos.Select(book => book.Id).ToList();
> 			var request = MakeGetRequest("get-books-with-these-ids");
> 			request.AddJsonBody(bookIds);
> ```

You've decided to use the body. I think that's best.
But you're still using GET.
Did you look into this at all? Some stuff I read suggests strongly that this is not really best practice. OTOH, all other things being equal, it would be nice if GET meant get and vice versa.

nabalone:
Due to a git snafoo on my part, this is PR did not contain the latest version of my code. I have fixed that but still need to respond to some of Andrew's comment's

nabalone (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6226)* status: 0 of 2 files reviewed, 4 unresolved discussions (waiting on @andrew-polk)

___
*[`src/BloomExe/WebLibraryIntegration/BloomLibraryBookApiClient.cs` line 247 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6226#-NjUGdpED473UiRfQtJ1:-NjhKRaF3oGgV_R7PuwB:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/d9e854bde4d8cfcf0af129b373cdabc2b6e4e1a7/src/BloomExe/WebLibraryIntegration/BloomLibraryBookApiClient.cs#L247)):*

I think we need to preserve this internet check at the beginning of GetBookCountByLanguage.
</blockquote></details>

Done.
___
*[`src/BloomExe/WebLibraryIntegration/BloomLibraryBookApiClient.cs` line 261 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6226#-NjUFe0k0MjCJZ6Y5VlW:-Njhn9uRA6qZlsz5hn54:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/36dbdfb417c9c1e9ce601f053df5c81b87ce94eb/src/BloomExe/WebLibraryIntegration/BloomLibraryBookApiClient.cs#L261)):*

Unless I'm missing something, I think it would be better to keep this method returning an int.
You can parse the string to ensure it is an int, else return -1.

In fact, I think we want any problem in this method to just return -1 (rather than throw).
I could be persuaded differently on this point. In the future, one could foresee different callers needing to handle exceptions differently. But since we only have one caller at this point, I think I would go with simplicity.
</blockquote></details>

Done.
___
*[`src/BloomExe/WebLibraryIntegration/BloomLibraryBookApiClient.cs` line 365 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6226#-NjUH2Ec8nSSldN0GIQ3:-Njhodmz2QUWJJ6L2Sbc:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/36dbdfb417c9c1e9ce601f053df5c81b87ce94eb/src/BloomExe/WebLibraryIntegration/BloomLibraryBookApiClient.cs#L365)):*

Naming is hard...
I think I would go with `get-books-with-ids` or `get-books-by-ids`. 

Possibly even `get-books`.
After all, if we wanted to allow for getting books by some other parameter, such as a query, that could all be handled at the parameter level.

Thinking more, I think I prefer `get-books`.
</blockquote></details>

Done.
___
*[`src/BloomExe/WebLibraryIntegration/BloomLibraryBookApiClient.cs` line 366 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6226#-NjUHf2T8PeA9x1H7pT9:-NjhoPG26Y6RaRY5Z5_F:bmeqxic) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/36dbdfb417c9c1e9ce601f053df5c81b87ce94eb/src/BloomExe/WebLibraryIntegration/BloomLibraryBookApiClient.cs#L366)):*

You've decided to use the body. I think that's best.
But you're still using GET.
Did you look into this at all? Some stuff I read suggests strongly that this is not really best practice. OTOH, all other things being equal, it would be nice if GET meant get and vice versa.
</blockquote></details>

I had accidentally pushed an old version of the code. Should be changed to POST now. Yeah it won't let me put a body on a GET request

andrew-polk (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6226)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @nabalone)

PR #6228, opened 2023-11-21
Fix timeout errors when calling API (BL-12568)

andrew-polk:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6228)

### DIFF
--- a/src/BloomExe/WebLibraryIntegration/BloomLibraryBookApiClient.cs
+++ b/src/BloomExe/WebLibraryIntegration/BloomLibraryBookApiClient.cs
@@ -37,6 +37,17 @@ namespace Bloom.WebLibraryIntegration
 			_parseApplicationId = keys.ParseApplicationKey;
 		}
 
+		private void LogApiError(string apiEndpoint, IRestResponse response)
+		{
+			SIL.Reporting.Logger.WriteEvent($@"BloomLibraryBookApiClient call to {apiEndpoint} failed
+  StatusCode: {response.StatusCode}
+  Content: {response.Content}
+  ResponseStatus: {response.ResponseStatus}
+  StatusDescription: {response.StatusDescription}
+  ErrorMessage: {response.ErrorMessage}
+  ErrorException: {response.ErrorException}");
+		}
+
 		// This calls an azure function which does the following:
 		// New book:
 		//  - Creates an empty `books` record in parse-server with an uploadPendingTimestamp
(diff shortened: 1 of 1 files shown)


JohnThomson (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6228)* status: all files reviewed, 1 unresolved discussion (waiting on @andrew-polk)

___
*[`src/BloomExe/WebLibraryIntegration/BloomLibraryBookApiClient.cs` line 63 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6228#-Njs2UtO-6URByOb2rTK:-Njs2UtO-6URByOb2rTL:brc096c) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/294ddaca2ce6ced244a40ff87462997eefc1b1b6/src/BloomExe/WebLibraryIntegration/BloomLibraryBookApiClient.cs#L63)):*
> ```Smalltalk
> 			if (response.StatusCode != HttpStatusCode.OK)
> 			{
> 				SIL.Reporting.Logger.WriteEvent($"upload-start failed: {response.StatusCode} - {response.Content}");
> ```

I'm not sure exactly what response.Content is likely to contain. If we won't otherwise get it, we should include anything that will help us identify the book that we failed to upload, at least its ID, but if we have it handy it would be good for the log to include the full pathname to the book. (Ideally some link that would take us straight to the book, but unfortunately we FAILED to upload it so it's not there.)
Is there any thing else (like an error message?) in the response that might help?

nabalone (approved):
If this works I don't see anything wrong with doing it. I'm not used to seeing timeouts in minutes but that's just me

andrew-polk:
*[`src/BloomExe/WebLibraryIntegration/BloomLibraryBookApiClient.cs` line 63 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6228#-Njs2UtO-6URByOb2rTK:-NjsEgDEEVM8UhiUlVTp:b63iegr) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/294ddaca2ce6ced244a40ff87462997eefc1b1b6/src/BloomExe/WebLibraryIntegration/BloomLibraryBookApiClient.cs#L63)):*

I'm not sure exactly what response.Content is likely to contain. If we won't otherwise get it, we should include anything that will help us identify the book that we failed to upload, at least its ID, but if we have it handy it would be good for the log to include the full pathname to the book. (Ideally some link that would take us straight to the book, but unfortunately we FAILED to upload it so it's not there.)
Is there any thing else (like an error message?) in the response that might help?
</blockquote></details>

I've enhanced the error messaging.
We don't have easy access to the book path at this point.
But there are other logs during the upload process which occur before this point which do have the book path.

PR #6232, opened 2023-12-01
Fix BL-12890 Duplicate thumbnails

hatton:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6232)

### DIFF
--- a/src/BloomExe/Collection/BookCollection.cs
+++ b/src/BloomExe/Collection/BookCollection.cs
@@ -306,12 +306,19 @@ namespace Bloom.Collection
 			_bookInfos.Insert(newIndex, info);
 			NotifyCollectionChanged();
 		}
+        private void AddBookInfoInternalNoSideEffects(BookInfo bookInfo)
+        {
+            // Note BL-12890 it's actually the norm, when we add a book, that the new book will already be there
+            // by the time we officially add it.
+            if (_bookInfos.FindIndex(i => i.Id == bookInfo.Id) < 0)
+                _bookInfos.Add(bookInfo);
+        }
 
-		public void AddBookInfo(BookInfo bookInfo)
-		{
-			_bookInfos.Add(bookInfo);
-			NotifyCollectionChanged();
-		}
+        public void AddBookInfo(BookInfo bookInfo)
+        {
+            AddBookInfoInternalNoSideEffects(bookInfo);
+            NotifyCollectionChanged();
+        }
 
 		/// <summary>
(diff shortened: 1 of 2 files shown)


JohnThomson (approved):
As I look at this more, I'm not satisfied.
- My first objection was, in looking at AddBookInfo(), it felt wrong for it to trigger all the likely side effects of NotifyCollectionChanged(); if in fact we didn't do anything to the collection. So i was going to suggest we only call it if we really added something.
- But then I realized: the way the bookInfo gets added to the collection earlier is through InsertBookInfo. And that NEVER calls NotifyCollectionChanged(). So the only way the notification gets out is in the later call to AddBookInfo.
- So then I was going to suggest that we get rid of AddBookInfo and just use InsertBookInfo, or make AddBookInfo call Insert, and fix InsertBookInfo so that, at least if it really inserted a bookInfo rather than just replacing one, it calls NotifyCollectionChanged().
- And something like that may be the right answer. But it means that the notification will go out at some possibly unexpected point during BBUD rather than after it is complete. Might be OK: I THINK the BookInfo gets saved and inserted as one of the last steps of Save, which is the last step of BBUD. But it feels fragile. And there may be other callers of the Insert method that really don't want notifications.
So I think I would suggest:
- Fix AddBookInfo like you have so it won't add a duplicate. It should also not send a notification if it does nothing.
- Rename the current InsertBookInfo to InsertBookInfoWithoutNotifying. Let it return a boolean indicating whether the bookInfo was new.
- Check callers of InsertBookInfo. Are there any where a notification SHOULD go out right away if we're adding a new book? If so, make a new InsertBookInfo that calls the renamed one and also does the notification if appropriate. Ones that really don't want that should call the renamed method.
- Callers like CreateFromSourceBook, where we know a new one is going to get inserted, can unconditionally call NotifyCollectionChanged(); when all is ready for it.
- Yet other callers may need to make their own use of the boolean returned by InsertBookInfoWithoutNotifying and, if it is true, call NotifyCollectionChanged() when appropriate.
Now, that's a lot of work. And I think what you have will fix the bug. You probably don't have time for this suggestion. But for the sake of improving the code base, I think we should do it soon. I'm OK with merging as-is if we plan to do better soon.

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6232#-:-Nkau0RMDv1ZOG_TO4VW:b-xiqo2f)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @hatton)

PR #6236, opened 2023-12-06
Allow for long-running actions (BL-12568)

andrew-polk:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6236)

### DIFF
--- a/src/BloomExe/WebLibraryIntegration/BloomLibraryBookApiClient.cs
+++ b/src/BloomExe/WebLibraryIntegration/BloomLibraryBookApiClient.cs
@@ -2,7 +2,7 @@ using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Net;
-using System.Text;
+using System.Threading;
 using System.Windows.Forms;
 using Bloom.Book;
 using Bloom.Properties;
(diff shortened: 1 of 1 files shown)


andrew-polk:
Draft because it must be merged after the azure function change (PR pending)

JohnThomson (commented):
One minor suggestion.

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6236#-:-NlAR4GF1rBQFhe8N3ZX:b-38b9fk)* status: all files reviewed, 1 unresolved discussion (waiting on @andrew-polk)

___
*[`src/BloomExe/WebLibraryIntegration/BloomLibraryBookApiClient.cs` line 90 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6236#-NlAMk7638vwzkDx0olt:-NlAMk7638vwzkDx0olu:bn9leda) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/38757028244828d54ad0f69135a77c507ccc9c34/src/BloomExe/WebLibraryIntegration/BloomLibraryBookApiClient.cs#L90)):*
> ```Smalltalk
> 				catch
> 				{
> 					handleError?.Invoke(response);
> ```

Could we make handleError take a message as well as a response, so that we could make the log contain not just an (at first sight) success response, but  more like "upload-start: status response did not contain valid data".
Possibly we might want to log the exception, too.

andrew-polk (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6236)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @JohnThomson)

___
*[`src/BloomExe/WebLibraryIntegration/BloomLibraryBookApiClient.cs` line 90 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6236#-NlAMk7638vwzkDx0olt:-Nlidz4x49P5hpPKP8pu:b-9w1d5f) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/38757028244828d54ad0f69135a77c507ccc9c34/src/BloomExe/WebLibraryIntegration/BloomLibraryBookApiClient.cs#L90)):*

Could we make handleError take a message as well as a response, so that we could make the log contain not just an (at first sight) success response, but  more like "upload-start: status response did not contain valid data".
Possibly we might want to log the exception, too.
</blockquote></details>

The logging of the response does start with 
"BloomLibraryBookApiClient call to {apiEndpoint} failed"

But it is a good point about logging specific messages and exceptions.

I switched it to the other way of doing things (letting the base method throw). It has its own ugliness, though.

Open to further improvements.

JohnThomson (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6236)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @andrew-polk)

PR #6237, opened 2023-12-12
BL-12901 Auto-format c# code in VS and VSCode

hatton:


andrew-polk:
I can't say I agree with every change it made, but I guess that is just something we accept as we did on the TS side.

I did question switching indentation from tab to space, but I see that's what we have on the TS side.

andrew-polk (approved):

PR #6238, opened 2023-12-14
Make checking for book records report errors and add logging (BL-12921)

andrew-polk:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6238)

### DIFF
--- a/src/BloomBrowserUI/publish/LibraryPublish/LibraryPublishSteps.tsx
+++ b/src/BloomBrowserUI/publish/LibraryPublish/LibraryPublishSteps.tsx
@@ -221,6 +221,10 @@ export const LibraryPublishSteps: React.FunctionComponent = () => {
         setIsUploadComplete(false);
         setIsUploading(true);
         get("libraryPublish/getUploadCollisionInfo", result => {
+            if (result.data.error) {
+                // The API already sent an error message
+                return;
+            }
             if (result.data.shouldShow) {
                 setUploadCollisionInfo(result.data);
                 showUploadCollisionDialog();
--- a/src/BloomExe/Publish/BloomLibrary/BloomLibraryPublishModel.cs
+++ b/src/BloomExe/Publish/BloomLibrary/BloomLibraryPublishModel.cs
@@ -116,9 +116,14 @@ namespace Bloom.Publish.BloomLibrary
 
 		internal bool IsBookPublicDomain => _license?.Url != null && _license.Url.StartsWith("http://creativecommons.org/publicdomain/zero/");
 
-		internal bool BookIsAlreadyOnServer => LoggedIn && _uploader.IsBookOnServer(Book.FolderPath);
+		internal bool IsBookAlreadyOnServer() {
+			return LoggedIn && _uploader.IsBookOnServer(Book.FolderPath);
+		}
 
-		internal dynamic ConflictingBookInfo => _uploader.GetBookOnServer(Book.FolderPath);
+		internal dynamic GetConflictingBookInfo()
+		{
+			return _uploader.GetBookOnServer(Book.FolderPath);
+		}
 
 		private string Uploader => _uploader.UserId;
 
@@ -525,7 +530,7 @@ namespace Bloom.Publish.BloomLibrary
 				newLanguages = newLangs;
 			}
 
-			var existingBookInfo = ConflictingBookInfo;
+			var existingBookInfo = GetConflictingBookInfo();
 			var updatedDateTime = (DateTime)existingBookInfo.updatedAt;
 			var createdDateTime = (DateTime)existingBookInfo.createdAt;
 			// Find the best title available (BL-11027)
@@ -571,7 +576,7 @@ namespace Bloom.Publish.BloomLibrary
 			// Must match IUploadCollisionDlgProps in uploadCollisionDlg.tsx.
 			return new
 			{
-				shouldShow = BookIsAlreadyOnServer,
+				shouldShow = true,
 				userEmail = LoggedIn ? WebUserId : "",
 				newThumbUrl = newThumbPath,
 				newTitle,
--- a/src/BloomExe/WebLibraryIntegration/BloomParseClient.cs
+++ b/src/BloomExe/WebLibraryIntegration/BloomParseClient.cs
@@ -175,7 +175,7 @@ namespace Bloom.WebLibraryIntegration
 
 		// Setting param 'includeLanguageInfo' to true adds a param to the query that causes it to fold in
 		// useful language information instead of only having the arcane langPointers object.
-		public IRestResponse GetBookRecordsByQuery(string query, bool includeLanguageInfo)
+		private IRestResponse GetBookRecordsByQuery(string query, bool includeLanguageInfo)
 		{
 			var request = MakeGetRequest("classes/books");
 			request.AddParameter("where", query, ParameterType.QueryString);
@@ -186,6 +186,7 @@ namespace Bloom.WebLibraryIntegration
 			return Client.Execute(request);
 		}
 
+		// Will throw an exception if there is any reason we can't make a successful query, including if there is no internet.
 		public dynamic GetSingleBookRecord(string id, bool includeLanguageInfo = false)
 		{
 			var json = GetBookRecords(id, includeLanguageInfo);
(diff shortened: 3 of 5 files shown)


JohnThomson (changes requested):
Mostly good, a couple of possible improvements.

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6238#-:-NldHSQ14dQK52tb487B:bq2a8x2)* status: all files reviewed, 2 unresolved discussions (waiting on @andrew-polk)

___
*[`src/BloomExe/WebLibraryIntegration/BloomParseClient.cs` line 199 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6238#-NldAZjC1VFJ4rYU_GQC:-NldAZjD-ANk_yTg7SvS:b-y9qbts) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/d3d91353d9fb16152e24f828d573b7ca008bca26/src/BloomExe/WebLibraryIntegration/BloomParseClient.cs#L199)):*
> ```Smalltalk
> 				bookRecord = json[0];
> 
> 			return true;
> ```

Shouldn't it return false if json.Count is zero? Maybe that will never happen? It would previously have returned null in this case.
If feels like there are two boolean results here: one indicating "we got a valid response from the server" and one indicating "we found a book with that ID". To me it feels more natural that the latter is the meaning of the return value, but the calling code is treating false as an error condition. In fact, it feels close to an abuse of the TryGet pattern to return true along with a null result.
Would we log it elsewhere if the server request timed out or otherwise failed?
Do we know this book exists on the server, so it's definitely an error if we don't find it?
What would you think of going back to GetSingleBookRecord() returning record or null, but throwing an exception here in this method if we get an unexpected (i.e., null) response?
___
*[`src/BloomExe/web/controllers/LibraryPublishApi.cs` line 336 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6238#-NldH8__52YNz57YwiXr:-NldH8__52YNz57YwiXs:b-cot3tl) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/d3d91353d9fb16152e24f828d573b7ca008bca26/src/BloomExe/web/controllers/LibraryPublishApi.cs#L336)):*
> ```Smalltalk
> 					{
> 						error = true,
> 						shouldShow = false
> ```

I'm not quite sure what this means we should or shouldn't show, but will it prevent the user seeing the progress message? Maybe it needs a clearer name, or at least a comment.

andrew-polk (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6238)* status: 3 of 7 files reviewed, 2 unresolved discussions (waiting on @JohnThomson)

___
*[`src/BloomExe/WebLibraryIntegration/BloomParseClient.cs` line 199 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6238#-NldAZjC1VFJ4rYU_GQC:-Nlegh192c9tJtc2x5Bm:bb135kb) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/d3d91353d9fb16152e24f828d573b7ca008bca26/src/BloomExe/WebLibraryIntegration/BloomParseClient.cs#L199)):*

Shouldn't it return false if json.Count is zero? Maybe that will never happen? It would previously have returned null in this case.
If feels like there are two boolean results here: one indicating "we got a valid response from the server" and one indicating "we found a book with that ID". To me it feels more natural that the latter is the meaning of the return value, but the calling code is treating false as an error condition. In fact, it feels close to an abuse of the TryGet pattern to return true along with a null result.
Would we log it elsewhere if the server request timed out or otherwise failed?
Do we know this book exists on the server, so it's definitely an error if we don't find it?
What would you think of going back to GetSingleBookRecord() returning record or null, but throwing an exception here in this method if we get an unexpected (i.e., null) response?
</blockquote></details>

Reworked completely so that the lowest level throws.
This might not be the ideal code structure, but it is only temporary for 5.6.
___
*[`src/BloomExe/web/controllers/LibraryPublishApi.cs` line 336 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6238#-NldH8__52YNz57YwiXr:-Nlej-SNDF1LcOcSYXoy:blq9c0v) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/d3d91353d9fb16152e24f828d573b7ca008bca26/src/BloomExe/web/controllers/LibraryPublishApi.cs#L336)):*

I'm not quite sure what this means we should or shouldn't show, but will it prevent the user seeing the progress message? Maybe it needs a clearer name, or at least a comment.
</blockquote></details>

Hope this helps.

JohnThomson (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6238)* status: all files reviewed, 2 unresolved discussions (waiting on @andrew-polk)

___
*[`src/BloomExe/WebLibraryIntegration/BloomParseClient.cs` line 211 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6238#-NliNlYF8gHcxMWtnsIo:-NliNlYF8gHcxMWtnsIp:bqkj3vx) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/50303e2ca4d3f062f186384c8d2aff5086481132/src/BloomExe/WebLibraryIntegration/BloomParseClient.cs#L211)):*
> ```Smalltalk
> 
> 		// Query parse for books. Throw if there is any reason we can't make a successful query, including if there is no internet.
> 		public dynamic GetBookRecordsOrThrow(string id, bool includeLanguageInfo, bool includeBooksFromOtherUploaders = false)
> ```

It feels slightly odd to have OrThrow in this name. I can sort of see the point that you want to emphasize that it will throw if it can't actually get the records requested. But in C#, any method will throw if something bad enough goes wrong, so that isn't entirely unexpected. Also, it looks as if it was meant to contrast with the previous method (which someone might then assume will NOT throw if it can't GetSingleBookRecord). But in fact both of them will throw if they fail...in fact the other one calls this!
It's not a showstopper, especially for code that I think is part of what you said is going away soon. But maybe worth thinking about again.
___
*[`src/BloomExe/web/controllers/LibraryPublishApi.cs` line 347 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6238#-NlhzUkXD5QUrudk3Aba:-NlhzUkXD5QUrudk3Abb:bxi8iqi) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/50303e2ca4d3f062f186384c8d2aff5086481132/src/BloomExe/web/controllers/LibraryPublishApi.cs#L347)):*
> ```Smalltalk
> 		{
> 			error = true, // Inform the client there was an error. Don't continue with the upload.
> 			shouldShow = false // Don't show the dialog. (Currently this is ignored if error is true.)
> ```

If it's ignored, and that means we WILL show the dialog, seems like it would convey intent more clearly to pass true.

andrew-polk (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6238)* status: 4 of 7 files reviewed, 2 unresolved discussions (waiting on @JohnThomson)

___
*[`src/BloomExe/WebLibraryIntegration/BloomParseClient.cs` line 211 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6238#-NliNlYF8gHcxMWtnsIo:-NliRGgS30R-9O79fxcA:bu9ax6r) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/50303e2ca4d3f062f186384c8d2aff5086481132/src/BloomExe/WebLibraryIntegration/BloomParseClient.cs#L211)):*

It feels slightly odd to have OrThrow in this name. I can sort of see the point that you want to emphasize that it will throw if it can't actually get the records requested. But in C#, any method will throw if something bad enough goes wrong, so that isn't entirely unexpected. Also, it looks as if it was meant to contrast with the previous method (which someone might then assume will NOT throw if it can't GetSingleBookRecord). But in fact both of them will throw if they fail...in fact the other one calls this!
It's not a showstopper, especially for code that I think is part of what you said is going away soon. But maybe worth thinking about again.
</blockquote></details>

Good points. I've put it back.
___
*[`src/BloomExe/web/controllers/LibraryPublishApi.cs` line 347 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6238#-NlhzUkXD5QUrudk3Aba:-NliRCFY3EyL5FAfqhlv:b-15csfz) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/50303e2ca4d3f062f186384c8d2aff5086481132/src/BloomExe/web/controllers/LibraryPublishApi.cs#L347)):*

If it's ignored, and that means we WILL show the dialog, seems like it would convey intent more clearly to pass true.
</blockquote></details>

See if that comment clarification helps.

JohnThomson (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6238)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @andrew-polk)

PR #6242, opened 2024-01-05
In Debugger mode watch files and refresh when they change

hatton:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6242)

### DIFF
--- a/src/BloomExe/Edit/EditingModel.cs
+++ b/src/BloomExe/Edit/EditingModel.cs
@@ -70,6 +70,16 @@ namespace Bloom.Edit
 
         internal const string PageScalingDivId = "page-scaling-container";
 
+        /// <summary>
+        /// Currently this is only valid in EditingView, since it depends on the Javascript code being
+        /// configured to send appropriate messages to the editView/setIsSelectionRange API.
+        /// </summary>
+        public static bool IsTextSelected;
+
+        private FileSystemWatcher _developerFileWatcher;
+        private DateTime _lastTimeWeReloadedBecauseOfDeveloperChange;
+        private bool _skipNextSaveBecauseDeveloperIsTweakingSupportingFiles;
+
         //public event EventHandler UpdatePageList;
 
         public delegate EditingModel Factory(); //autofac uses this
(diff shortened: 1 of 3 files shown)


JohnThomson (approved):
LGTM. Some very minor suggestions, or you can just merge.

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6242#-:-NnPsVrV5fwb_wOIk_xJ:b-37uk7p)* status: all files reviewed, 3 unresolved discussions (waiting on @hatton)

___
*[`src/BloomExe/Edit/EditingModel.cs` line 1783 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6242#-NnPqaEe8qWKzCNhFNlJ:-NnPqaEe8qWKzCNhFNlK:bfdreci) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/4373365ff1a6a85f022ac7811f188ab1c6ebd6e6/src/BloomExe/Edit/EditingModel.cs#L1783)):*
> ```Smalltalk
>                     if (CurrentBook == null)
>                         return;
>                     // if we've been called already in the past 5 seconds, don't do it again
> ```

2 seconds :-)
___
*[`src/BloomExe/Edit/EditingModel.cs` line 1798 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6242#-NnPqsJWC9kHU5c5IzeD:-NnPqsJWC9kHU5c5IzeE:bg4ys9f) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/4373365ff1a6a85f022ac7811f188ab1c6ebd6e6/src/BloomExe/Edit/EditingModel.cs#L1798)):*
> ```Smalltalk
>                     // Enhance: I suspect all the problems here are related to us changing the page id's each time we load, which I don't understand.
>                     // It may just be a mistake.
>                     if (args.Name.EndsWith(".json"))
> ```

.ToLowerCaseInvariant() ?
___
*[`src/BloomExe/web/controllers/BrandingSettings.cs` line 180 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6242#-NnPru3r8ZZBJfg2FYyz:-NnPru3r8ZZBJfg2FYz-:bvdehh4) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/4373365ff1a6a85f022ac7811f188ab1c6ebd6e6/src/BloomExe/web/controllers/BrandingSettings.cs#L180)):*
> ```Smalltalk
>                     {
>                         NonFatalProblem.Report(
>                             ModalIf.Beta,
> ```

Do we want Alpha, or even do this at all only for developers? At least the message about watch:branding:files will only help devs.

PR #6243, opened 2024-01-05
Major rework of BringBookUpToDate and AppearanceSettings (BL-12853)

JohnThomson:
The original idea was to allow the user to select a non-legacy theme even if the book has incompatible custom CSS. That branched out in to fixing a multitude of problems, both with AppearanceSettings.cs itself, and pre-existing issues which the new system made more serious.
BringBookUpToDate was drastically reorganized and simplified, and is ow only done at all if we can save changes to the folder.
customBookStyles2.css was introduced as the place to put substitute CSS that allows certain known problem customCss to be replaced with a modern theme.
Books can no longer be published if they are in a TC and not checked out.

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6243)


JohnThomson:
Extensive interactive reviews with JohnH and Andrew

PR #6244, opened 2024-01-08
Improve performance of new upload (BL-12568)

andrew-polk:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6244)

### DIFF
--- a/src/BloomExe/WebLibraryIntegration/BloomS3Client.cs
+++ b/src/BloomExe/WebLibraryIntegration/BloomS3Client.cs
@@ -221,6 +221,7 @@ namespace Bloom.WebLibraryIntegration
             string[] audioLanguagesToInclude,
             string metadataLang1Code,
             string metadataLang2Code,
+            bool isNewBook,
             string collectionSettingsPath = null,
             bool isForBulkUpload = false
         )
@@ -263,6 +264,7 @@ namespace Bloom.WebLibraryIntegration
                 PublishHelper.ReportInvalidFonts(stagingDirectory, progress);
 
                 SyncBookFiles(
+                    isNewBook,
                     storageKeyOfBookFolderParent,
                     stagingParentDirectory.FolderPath,
                     progress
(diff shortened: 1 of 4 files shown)


JohnThomson (approved):
One minor and optional suggestion

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6244#-:-Nnf3Pog59dwuHIzFndE:becd87l)* status: all files reviewed, 1 unresolved discussion (waiting on @andrew-polk)

___
*[`src/BloomExe/WebLibraryIntegration/BloomS3Client.cs` line 449 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6244#-Nnf2xZcADGSH8KlVoKA:-Nnf2xZcADGSH8KlVoKB:bbbkgni) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/510cc908ab0dc006ed375c3e4b1f16eb83d7ceaa/src/BloomExe/WebLibraryIntegration/BloomS3Client.cs#L449)):*
> ```Smalltalk
>                     + localFilePath.Substring(stagingDirectory.Length).Replace("\\", "/");
>                 key = key.Replace("//", "/"); // safety net
>                 if (isNewBook || !existingObjectsOnS3.Any(o => o.Key == key))
> ```

Depends on knowing that existingObjectsOnS3 will always be non-null if isNewBook is false. I'd be inclined to either make it a test of the list being null, or just initialize to an empty list so you don't have to check.

andrew-polk (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6244)* status: 3 of 4 files reviewed, 1 unresolved discussion (waiting on @JohnThomson)

___
*[`src/BloomExe/WebLibraryIntegration/BloomS3Client.cs` line 449 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6244#-Nnf2xZcADGSH8KlVoKA:-Nnf3kNk2KiOrrgzhgSa:bdcv6l8) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/510cc908ab0dc006ed375c3e4b1f16eb83d7ceaa/src/BloomExe/WebLibraryIntegration/BloomS3Client.cs#L449)):*

Depends on knowing that existingObjectsOnS3 will always be non-null if isNewBook is false. I'd be inclined to either make it a test of the list being null, or just initialize to an empty list so you don't have to check.
</blockquote></details>

Good point.

PR #6252, opened 2024-01-17
Change upload API - client sends files and hashes (BL-12568)

andrew-polk:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6252)

### DIFF
--- a/src/BloomExe/Book/AppearanceMigrator.cs
+++ b/src/BloomExe/Book/AppearanceMigrator.cs
@@ -78,11 +78,7 @@ namespace Bloom.Book
 
         public static string GetChecksum(string css)
         {
-            using (var md5 = MD5.Create())
-            {
-                var hash = md5.ComputeHash(Encoding.UTF8.GetBytes(css));
-                return BitConverter.ToString(hash).Replace("-", "").ToLowerInvariant();
-            }
+            return Utils.MiscUtils.GetMd5HashOfString(css);
         }
     }
 }
--- a/src/BloomExe/Book/BookInfo.cs
+++ b/src/BloomExe/Book/BookInfo.cs
@@ -1073,7 +1073,7 @@ namespace Bloom.Book
             return Path.ChangeExtension(MetaDataPath(bookFolderPath), BackupExtension);
         }
 
-        public void WriteToFolder(string bookFolderPath)
+        public void WriteToFolder(string bookFolderPath, bool makeBackup = true)
         {
             string tempFilePath;
             var metaDataPath = MetaDataPath(bookFolderPath);
@@ -1091,7 +1091,11 @@ namespace Bloom.Book
 
             RobustFile.WriteAllText(tempFilePath, Json);
             if (RobustFile.Exists(metaDataPath))
-                RobustFile.Replace(tempFilePath, metaDataPath, BackupFilePath(bookFolderPath));
+                RobustFile.Replace(
+                    tempFilePath,
+                    metaDataPath,
+                    makeBackup ? BackupFilePath(bookFolderPath) : null
+                );
             else
                 RobustFile.Move(tempFilePath, metaDataPath);
         }
--- a/src/BloomExe/Utils/MiscUtils.cs
+++ b/src/BloomExe/Utils/MiscUtils.cs
@@ -2,11 +2,11 @@ using System;
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.Drawing;
-using System.Globalization;
 using System.IO;
 using System.Linq;
 using System.Management;
 using System.Security.AccessControl;
+using System.Security.Cryptography;
 using System.Security.Principal;
 using System.Text;
 using System.Text.RegularExpressions;
@@ -14,8 +14,6 @@ using System.Windows.Forms;
 using Bloom.Book;
 using Mono.Unix;
 using NAudio.Wave;
-using SIL.Code;
-using SIL.CommandLineProcessing;
 using SIL.IO;
 using SIL.PlatformUtilities;
 #if __MonoCS__
@@ -669,6 +667,30 @@ namespace Bloom.Utils
                     return true;
             }
             return false;
+        }
+
+        public static string GetMd5HashOfFile(string filePath)
+        {
+            using (var stream = RobustFile.OpenRead(filePath))
+            using (var md5 = MD5.Create())
+            {
+                byte[] hash = md5.ComputeHash(stream);
+                return ConvertHashToString(hash);
+            }
+        }
+
+        public static string GetMd5HashOfString(string str)
+        {
+            using (var md5 = MD5.Create())
+            {
+                byte[] hash = md5.ComputeHash(Encoding.UTF8.GetBytes(str));
+                return ConvertHashToString(hash);
+            }
+        }
+
+        private static string ConvertHashToString(byte[] hash)
+        {
+            return BitConverter.ToString(hash).Replace("-", "").ToLowerInvariant();
         }
     }
 }
--- a/src/BloomExe/WebLibraryIntegration/BloomLibraryBookApiClient.cs
+++ b/src/BloomExe/WebLibraryIntegration/BloomLibraryBookApiClient.cs
@@ -141,14 +141,23 @@ namespace Bloom.WebLibraryIntegration
         // Existing book:
         //  - Verifies the user has permission to update the book (using parse-server session)
         //  - Sets uploadPendingTimestamp on the `books` record in parse-server
-        //  - Copies book files from existing S3 location to a new S3 location based on bookObjectId/timestamp
+        //  - Using the provided file paths and hashes, determines which files need to be copied from the existing
+        //     S3 location and which need to be uploaded by the client
+        //  - Copies book files from existing S3 location to a new S3 location based on bookObjectId/timestamp;
+        //     sets public-read permissions on these files
         // New and existing books:
         //  - Generates temporary credentials for the client to upload the book files to the new S3 location
         public (
             string transactionId,
-            string storageKeyOfBookFolderParentOnS3,
-            AmazonS3Credentials uploadCredentials
-        ) InitiateBookUpload(IProgress progress, string existingBookObjectId = null)
+            AmazonS3Credentials uploadCredentials,
+            string s3PrefixToUploadTo,
+            string[] filesToUpload
+        ) InitiateBookUpload(
+            IProgress progress,
+            List<FilePathAndHash> bookFiles,
+            string bookTitle,
+            string existingBookObjectId = null
+        )
         {
             if (!LoggedIn)
                 throw new ApplicationException("Must be logged in to upload a book");
@@ -158,6 +167,20 @@ namespace Bloom.WebLibraryIntegration
             if (!string.IsNullOrEmpty(existingBookObjectId))
                 request.AddQueryParameter("existing-book-object-id", existingBookObjectId);
 
+            // We give the server a list of files and their hashes so it can be the controller of which files to upload.
+            // body format is
+            // { title: "My title/", files: [{ "path": "abc", "hash": "123" }, ...] }
+            var obj = new Dictionary<string, object>
+            {
+                // Because the title could change between uploads, we can't rely on the existing
+                // path on S3 (which includes the title). We have to provide it ourselves.
+                // However, we don't call it book-title in the API because some day we would like
+                // to get away from that being part of the path.
+                { "title", bookTitle + "/" },
+                { "files", JsonConvert.SerializeObject(bookFiles) },
+            };
+            request.AddJsonBody(obj);
+
             var result = CallLongRunningAction(
                 request,
                 progress,
@@ -166,24 +189,34 @@ namespace Bloom.WebLibraryIntegration
             );
 
             if (progress.CancelRequested)
-                return (null, null, null);
+                return (null, null, null, null);
+
+            var s3PrefixToUploadTo = BloomS3Client.GetPrefixFromBookFileUploadUrl(
+                (string)result.url
+            );
+
+            if (string.IsNullOrEmpty(s3PrefixToUploadTo))
+                throw new ApplicationException("Unable to initiate book upload on the server.");
+
+            if (!s3PrefixToUploadTo.EndsWith("/"))
+                s3PrefixToUploadTo += "/";
 
             return (
                 result["transaction-id"],
-                BloomS3Client.GetStorageKeyOfBookFolderParentFromUrl((string)result.url),
                 new AmazonS3Credentials
                 {
                     AccessKey = result.credentials.AccessKeyId,
                     SecretAccessKey = result.credentials.SecretAccessKey,
                     SessionToken = result.credentials.SessionToken
-                }
+                },
+                s3PrefixToUploadTo,
+                result["files-to-upload"].ToObject<string[]>()
             );
         }
 
         // This calls an azure function which does the following:
         //  - Verifies the user has permission to update the book (using parse-server session)
         //  - Verifies the baseUrl includes the expected S3 location
-        //  - Sets up read permission on the files in the new S3 location
         //  - Updates the `books` record in parse-server with all fields from the client,
         //     including the new baseUrl which points to the new S3 location. Sets uploadPendingTimestamp to null.
         //  - Deletes the book files from the old S3 location
--- a/src/BloomExe/WebLibraryIntegration/BloomS3Client.cs
+++ b/src/BloomExe/WebLibraryIntegration/BloomS3Client.cs
@@ -5,17 +5,12 @@ using System.IO;
 using System.Linq;
 using System.Net;
 using System.Security;
-using System.Security.Cryptography;
 using System.Text;
 using System.Text.RegularExpressions;
 using System.Web;
-using System.Xml;
 using Amazon.S3;
 using Amazon.S3.Model;
 using Amazon.S3.Transfer;
-using Bloom.Book;
-using Bloom.Publish;
-using Bloom.web.controllers;
 using BloomTemp;
 using L10NSharp;
 using SIL.Code;
@@ -163,11 +158,11 @@ namespace Bloom.WebLibraryIntegration
             set { _s3Config.MaxErrorRetry = value; }
         }
 
-        internal List<S3Object> GetMatchingItems(string bucketName, string key)
+        internal List<S3Object> GetMatchingItems(string bucketName, string prefix)
         {
             return GetAmazonS3WithAccessKey(bucketName)
                 .ListAllObjects(
-                    new ListObjectsV2Request() { BucketName = bucketName, Prefix = key }
+                    new ListObjectsV2Request() { BucketName = bucketName, Prefix = prefix }
                 );
         }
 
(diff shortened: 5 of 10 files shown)


andrew-polk:
Draft because it must be coordinated with the azure function side (https://github.com/BloomBooks/bloom-azure-functions/pull/72).

andrew-polk (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6252)* status: 0 of 7 files reviewed, 2 unresolved discussions

___
*[`src/BloomExe/WebLibraryIntegration/BloomS3Client.cs` line 473 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6252#-NoNGRAP-IUS7lVIlkfP:-NoNGRAQ2ARO20Lf4phr:b-q8v28j) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/9e4cd18819af374fe0418855ee311e9bbea919d1/src/BloomExe/WebLibraryIntegration/BloomS3Client.cs#L473)):*
> ```Smalltalk
>                 Path.GetExtension(localFilePath).ToLowerInvariant()
>             );
>         }
> ```

These were already not needed since we implemented the white list (BookFileFilter).

But it was handling one thing kinda by accident. It was still actively preventing meta.bak from being uploaded. See changes in `RemoveUnwantedLanguageData` and `BookInfo.cs` which now handle this.
___
*[`src/BloomExe/WebLibraryIntegration/BookUpload.cs` line 300 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6252#-NoNHepDEH5Ubamul1OM:-NoNHepDEH5Ubamul1ON:bpnakog) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/edaa6773add27c04d505300f784f8a4733fcc514/src/BloomExe/WebLibraryIntegration/BookUpload.cs#L300)):*
> ```Smalltalk
>                         List<FilePathAndHash> existingBookFiles = null;
>                         if (!isNewBook)
>                             existingBookFiles = GetStagedFiles(stagingDirectory);
> ```

I struggled with this.

For new books, we could just follow the pattern of existing books. We would send the list of all book files, and the API would return that same exact list. That seems like a waste of bandwidth.

On the other hand, we like the idea of having the azure function be the controller. And the only way to let that happen is to send this list back and forth.
I think we even discussed the possibility that the function could become smart enough to restart a failed upload. If that failed upload was for a new book, the only way to achieve that would be for the file list to get sent/received even for new books.

Please give any thoughts or suggestions.
___
*[`src/BloomExe/WebLibraryIntegration/BookUpload.cs` line 493 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6252#-NoNHEuI4VxhJcE3o9QI:-NoNHEuJ-tlJklX6WoE0:b-yredv0) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/edaa6773add27c04d505300f784f8a4733fcc514/src/BloomExe/WebLibraryIntegration/BookUpload.cs#L493)):*
> ```Smalltalk
> 
>         // Copy the needed files to the staging directory and make any modifications needed before upload.
>         private string SetUpStaging(
> ```

This method is essentially the old BloomS3Client.UploadBook with the following changes:
1. The parent creates the staging temp folder and passes in the path.
2. At the end, we simply return the staged book dir rather than call SyncBookFiles.
___
*[`src/BloomExe/WebLibraryIntegration/BookUpload.cs` line 543 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6252#-NoNH0641VSOZ6uZ80zf:-NoNH0641VSOZ6uZ80zg:b-g166bv) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/edaa6773add27c04d505300f784f8a4733fcc514/src/BloomExe/WebLibraryIntegration/BookUpload.cs#L543)):*
> ```Smalltalk
>         }
> 
>         private void ProcessVideosInTempDirectory(string destDirName)
> ```

This method was moved from BloomS3Client unchanged.
___
*[`src/BloomExe/WebLibraryIntegration/BookUpload.cs` line 566 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6252#-NoNH3bd0iJLlcyzaEnL:-NoNH3bd0iJLlcyzaEnM:b-g166bv) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/edaa6773add27c04d505300f784f8a4733fcc514/src/BloomExe/WebLibraryIntegration/BookUpload.cs#L566)):*
> ```Smalltalk
>         /// See BL-12583.
>         /// </remarks>
>         private static void CopyCollectionSettingsToTempDirectory(
> ```

This method was moved from BloomS3Client unchanged.
___
*[`src/BloomExe/WebLibraryIntegration/BookUpload.cs` line 586 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6252#-NoNH5HKDtRFyeUmaNdx:-NoNH5HKDtRFyeUmaNdy:bybak91) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/edaa6773add27c04d505300f784f8a4733fcc514/src/BloomExe/WebLibraryIntegration/BookUpload.cs#L586)):*
> ```Smalltalk
>         }
> 
>         private void RemoveUnwantedLanguageData(
> ```

This method was moved from BloomS3Client with one change:
on the last line, I added the `makeBackup:false` parameter to `WriteToFolder`.

hatton (commented):
> Reviewed 5 of 7 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6252)* status: 5 of 7 files reviewed, 4 unresolved discussions (waiting on @andrew-polk)

___
*[`src/BloomExe/WebLibraryIntegration/BloomLibraryBookApiClient.cs` line 171 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6252#-NoODlMO3cVOYz3VCdmB:-NoODlMO3cVOYz3VCdmC:b-rkrbb9) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/edaa6773add27c04d505300f784f8a4733fcc514/src/BloomExe/WebLibraryIntegration/BloomLibraryBookApiClient.cs#L171)):*
> ```Smalltalk
>                 request.AddQueryParameter("existing-book-object-id", existingBookObjectId);
>                 // If uploading an existing book, we give the server a list of files and their hashes.
>                 // Format is `files: [{ "path": "abc", "hash": "123" }, ...]`
> ```

If you want you could update this to include this `files-prefix` but it's not a big deal
___
*[`src/BloomExe/WebLibraryIntegration/BloomS3Client.cs` line 284 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6252#-NoOFI534yUCnIBUQRYK:-NoOFI534yUCnIBUQRYL:b-a940ji) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/edaa6773add27c04d505300f784f8a4733fcc514/src/BloomExe/WebLibraryIntegration/BloomS3Client.cs#L284)):*
> ```Smalltalk
>         }
> 
>         public static string GetEtag(string filePath)
> ```

worth a comment maybe pointing to how we know these are just md5's of the whole file?  [This blog](https://teppen.io/2018/06/23/aws_s3_etags/) makes it seem more complicated

hatton (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6252)* status: 5 of 7 files reviewed, 4 unresolved discussions (waiting on @andrew-polk)

___
*[`src/BloomExe/WebLibraryIntegration/BookUpload.cs` line 300 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6252#-NoNHepDEH5Ubamul1OM:-NoOGIg3EYBdgigOQMk0:b-x9y9qo) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/edaa6773add27c04d505300f784f8a4733fcc514/src/BloomExe/WebLibraryIntegration/BookUpload.cs#L300)):*
> That seems like a waste of bandwidth.

I would think it is so small that whatever is easier to code/maintain/reason about is preferred.

hatton (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6252)* status: 5 of 7 files reviewed, 5 unresolved discussions (waiting on @andrew-polk)

___
*[`src/BloomExe/WebLibraryIntegration/BookUpload.cs` line 278 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6252#-NoOGTxHCxk3xOVhBkFV:-NoOGTxHCxk3xOVhBkFW:b-c0fjmz) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/edaa6773add27c04d505300f784f8a4733fcc514/src/BloomExe/WebLibraryIntegration/BookUpload.cs#L278)):*
> ```Smalltalk
> 
>                     // This currently (unfortunately) enforces a single upload at a time.
>                     // We considered modifying it now, but decided we would discover the problem
> ```

I'm not following this comment: we would discover *what* problem?

JohnThomson (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6252)* status: all files reviewed, 6 unresolved discussions (waiting on @andrew-polk and @hatton)

___
*[`src/BloomExe/WebLibraryIntegration/BookUpload.cs` line 278 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6252#-NoOGTxHCxk3xOVhBkFV:-NoOUFmG3kHgWdGSH7tY:b-oqk6o3) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/edaa6773add27c04d505300f784f8a4733fcc514/src/BloomExe/WebLibraryIntegration/BookUpload.cs#L278)):*

I'm not following this comment: we would discover *what* problem?
</blockquote></details>

That they are all trying to use the same staging directory.
___
*[`src/BloomExe/WebLibraryIntegration/BookUpload.cs` line 300 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6252#-NoNHepDEH5Ubamul1OM:-NoO8LfL3vxyLpMcJ8bG:b-55wcg9) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/edaa6773add27c04d505300f784f8a4733fcc514/src/BloomExe/WebLibraryIntegration/BookUpload.cs#L300)):*

> That seems like a waste of bandwidth.

I would think it is so small that whatever is easier to code/maintain/reason about is preferred.
</blockquote></details>

It could be a useful optimization to not have to compute checksums and send them back and forth for new books. I don't know how often books get re-uploaded, but it would not surprise me if first-time uploads are a majority, perhaps even by a good margin. Any idea how long it takes to compute this sort of checksum for a large book full of files?

If we one day implement restartable, I would expect things to begin with a call rather like the one that currently asks "is this an existing book". It would have a three-way answer, new, existing, or incomplete. Existing or incomplete would both result in sending the list of files and checksums and getting back a list of what to upload. (I think 'incomplete' should only be used for new books. Bloom probably needs to know that it is attempting to upload an existing book. If there has already been an incomplete attempt to upload an existing book, the server function can just quietly take advantage of any files already successfully copied or uploaded to shrink the list of things to upload this time. Elegantly, since we're sending the checksums, we can continue an incomplete upload even if the book has been edited. We just need the server to checksum any files in the staging area.)

OTOH, it would definitely be simpler if this code didn't even have to think about whether the book is new and/or being retried. It just tells the server, I want the book to consist of these files with these checksums. And the server tells it what needs uploading. Quite possibly the client doesn't have to change at all to make it restartable. It would be nice to be able to ship restartability any time, without having to wait for a new client to make it to release.

So I think I'm voting for "always send the list" unless we have evidence that the cost of computing the checksums and sending the extra data is likely to be a significant part of the upload time (say, more than 5%? 10%?)
___
*[`src/BloomTests/WebLibraryIntegration/BloomS3StandardUpDownloadTests.cs` line 89 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6252#-NoOTB-1D6wztSWvTqrj:-NoOTB-1D6wztSWvTqrk:bjh1qjt) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/9e4cd18819af374fe0418855ee311e9bbea919d1/src/BloomTests/WebLibraryIntegration/BloomS3StandardUpDownloadTests.cs#L89)):*
> ```Smalltalk
>             File.WriteAllText(Path.Combine(bookFolder, "one.css"), @"test");
>             File.WriteAllText(Path.Combine(bookFolder, "preview.pdf"), @"test pdf file");
>             File.WriteAllText(Path.Combine(bookFolder, "extra.pdf"), @"unwanted pdf file");
> ```

I'm not clear why we no longer need to test that unwanted stuff is not getting sent. Maybe you're just satisfied with testing of the file filter elsewhere? Or maybe I missed some equivalent new test?

andrew-polk (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6252)* status: all files reviewed, 5 unresolved discussions (waiting on @hatton and @JohnThomson)

___
*[`src/BloomExe/WebLibraryIntegration/BloomS3Client.cs` line 284 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6252#-NoOFI534yUCnIBUQRYK:-NoOY0gjEu_Hw_6cp3UP:b-bpbrrt) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/edaa6773add27c04d505300f784f8a4733fcc514/src/BloomExe/WebLibraryIntegration/BloomS3Client.cs#L284)):*

worth a comment maybe pointing to how we know these are just md5's of the whole file?  [This blog](https://teppen.io/2018/06/23/aws_s3_etags/) makes it seem more complicated
</blockquote></details>

Does that help?
___
*[`src/BloomExe/WebLibraryIntegration/BookUpload.cs` line 278 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6252#-NoOGTxHCxk3xOVhBkFV:-NoORmYv2gaUNKCewwpe:bxqjm44) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/edaa6773add27c04d505300f784f8a4733fcc514/src/BloomExe/WebLibraryIntegration/BookUpload.cs#L278)):*

That they are all trying to use the same staging directory.
</blockquote></details>

I just copied the comment; but I *did* write it originally... : )
The idea was just that if we ever wanted to call this method in parallel for multiple books, it wouldn't work because they would all be staged in the same place.
I tried to clean up the comment.
___
*[`src/BloomExe/WebLibraryIntegration/BookUpload.cs` line 300 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6252#-NoNHepDEH5Ubamul1OM:-NoObVoa5aGJyclQcJNU:bgyrv8v) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/edaa6773add27c04d505300f784f8a4733fcc514/src/BloomExe/WebLibraryIntegration/BookUpload.cs#L300)):*

It could be a useful optimization to not have to compute checksums and send them back and forth for new books. I don't know how often books get re-uploaded, but it would not surprise me if first-time uploads are a majority, perhaps even by a good margin. Any idea how long it takes to compute this sort of checksum for a large book full of files?

If we one day implement restartable, I would expect things to begin with a call rather like the one that currently asks "is this an existing book". It would have a three-way answer, new, existing, or incomplete. Existing or incomplete would both result in sending the list of files and checksums and getting back a list of what to upload. (I think 'incomplete' should only be used for new books. Bloom probably needs to know that it is attempting to upload an existing book. If there has already been an incomplete attempt to upload an existing book, the server function can just quietly take advantage of any files already successfully copied or uploaded to shrink the list of things to upload this time. Elegantly, since we're sending the checksums, we can continue an incomplete upload even if the book has been edited. We just need the server to checksum any files in the staging area.)

OTOH, it would definitely be simpler if this code didn't even have to think about whether the book is new and/or being retried. It just tells the server, I want the book to consist of these files with these checksums. And the server tells it what needs uploading. Quite possibly the client doesn't have to change at all to make it restartable. It would be nice to be able to ship restartability any time, without having to wait for a new client to make it to release.

So I think I'm voting for "always send the list" unless we have evidence that the cost of computing the checksums and sending the extra data is likely to be a significant part of the upload time (say, more than 5%? 10%?)
</blockquote></details>

On my machine, calculating the hashes of 444 files (mostly images, no audio or video) took 1.2 seconds.

So I'm planning to change things to simplify on the singular path of sending all files and hashes to the function so it can be the controller.
___
*[`src/BloomTests/WebLibraryIntegration/BloomS3StandardUpDownloadTests.cs` line 89 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6252#-NoOTB-1D6wztSWvTqrj:-NoOYvEQCPTZdXpSBjuO:b8wkpa4) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/9e4cd18819af374fe0418855ee311e9bbea919d1/src/BloomTests/WebLibraryIntegration/BloomS3StandardUpDownloadTests.cs#L89)):*

I'm not clear why we no longer need to test that unwanted stuff is not getting sent. Maybe you're just satisfied with testing of the file filter elsewhere? Or maybe I missed some equivalent new test?
</blockquote></details>

Yeah, we were already using BookFileFilter to create the staging directory. So that logic is already (in theory) tested there.
I just removed the redundant (except meta.bak which was handled differently) logic to blacklist certain files.

andrew-polk (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6252)* status: 1 of 10 files reviewed, 2 unresolved discussions (waiting on @hatton and @JohnThomson)

___
*[`src/BloomExe/WebLibraryIntegration/BloomLibraryBookApiClient.cs` line 171 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6252#-NoODlMO3cVOYz3VCdmB:-NoYxtJ6Dy9K9L1eT0c0:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/edaa6773add27c04d505300f784f8a4733fcc514/src/BloomExe/WebLibraryIntegration/BloomLibraryBookApiClient.cs#L171)):*

If you want you could update this to include this `files-prefix` but it's not a big deal
</blockquote></details>

Done.

JohnThomson (commented):
LGTM, with one thing I noticed about some comments after the name changes.

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6252#-:-NonaHCNCPE2Hvj0FX6e:b7sohkm)* status: 0 of 11 files reviewed, 1 unresolved discussion (waiting on @andrew-polk and @hatton)

___
*[`src/BloomExe/WebLibraryIntegration/BloomS3Client.cs` line 677 at r4](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6252#-NonZWQUAY9tjuc9lVfC:-NonZWQUAY9tjuc9lVfD:b-7amw2b) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/d3973e6b356d0088a5dbd2b41d03f1c6564e7f30/src/BloomExe/WebLibraryIntegration/BloomS3Client.cs#L677)):*
> ```Smalltalk
>         // baseUrl looks like https://s3.amazonaws.com/BloomLibraryBooks/5b4b4b4b%2f1234567890%2fBook+Title%2f
>         // We want the part after the bucket name, without encoding, like 5b4b4b4b/1234567890/Book Title
>         public static string GetPrefixFromBaseUrl(string baseUrl)
> ```

Something is funny here. Comparing this and the previous method...
---both say they are getting a prefix, but they get different things
---one says it takes a url and the other says it takes a baseUrl, but the comments indicate both are expected to be passed more-or-less the same thing.
It's not obvious why a typically URL has so many components. In real life, 5b4b4b4b would typically be an email address, and 1234567890 is a book ID, right?
Looks like the real difference is encoding...
But it would be nicer still if the names and/or comments indicated more of why the part at the END of the URL is called a Prefix, and what it's good for...
Of course the comments are not new (may even be mine!) but the name change drew attention to the problem :-)

andrew-polk (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6252)* status: 0 of 11 files reviewed, 1 unresolved discussion (waiting on @hatton and @JohnThomson)

___
*[`src/BloomExe/WebLibraryIntegration/BloomS3Client.cs` line 677 at r4](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6252#-NonZWQUAY9tjuc9lVfC:-Norc3tz1-2DYXA1qEop:bsv3v4h) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/d3973e6b356d0088a5dbd2b41d03f1c6564e7f30/src/BloomExe/WebLibraryIntegration/BloomS3Client.cs#L677)):*

Something is funny here. Comparing this and the previous method...
---both say they are getting a prefix, but they get different things
---one says it takes a url and the other says it takes a baseUrl, but the comments indicate both are expected to be passed more-or-less the same thing.
It's not obvious why a typically URL has so many components. In real life, 5b4b4b4b would typically be an email address, and 1234567890 is a book ID, right?
Looks like the real difference is encoding...
But it would be nicer still if the names and/or comments indicated more of why the part at the END of the URL is called a Prefix, and what it's good for...
Of course the comments are not new (may even be mine!) but the name change drew attention to the problem :-)
</blockquote></details>

I agree it is awkward. See if that helps.

JohnThomson (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6252)* status: 0 of 11 files reviewed, 1 unresolved discussion (waiting on @andrew-polk and @hatton)

___
*[`src/BloomExe/WebLibraryIntegration/BloomS3Client.cs` line 677 at r4](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6252#-NonZWQUAY9tjuc9lVfC:-Nors46V1erjl9ewLqCC:b-c2x4uv) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/d3973e6b356d0088a5dbd2b41d03f1c6564e7f30/src/BloomExe/WebLibraryIntegration/BloomS3Client.cs#L677)):*

I agree it is awkward. See if that helps.
</blockquote></details>

Definite improvement. It's clear what the URL in the first method is. I think it would help to say a little more about baseUrl....i.e., that it is the value of one of the book fields that our API yields when we query parse.com for data about a book (if I figured it out correctly from the old code). Up to you if it's worth doing more. It wasn't very hard to deduce what it is from VS.

andrew-polk (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6252)* status: 0 of 11 files reviewed, 1 unresolved discussion (waiting on @hatton and @JohnThomson)

___
*[`src/BloomExe/WebLibraryIntegration/BloomS3Client.cs` line 677 at r4](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6252#-NonZWQUAY9tjuc9lVfC:-NorvlFM0lfUmCwEpG9g:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/d3973e6b356d0088a5dbd2b41d03f1c6564e7f30/src/BloomExe/WebLibraryIntegration/BloomS3Client.cs#L677)):*

Definite improvement. It's clear what the URL in the first method is. I think it would help to say a little more about baseUrl....i.e., that it is the value of one of the book fields that our API yields when we query parse.com for data about a book (if I figured it out correctly from the old code). Up to you if it's worth doing more. It wasn't very hard to deduce what it is from VS.
</blockquote></details>

Done.

JohnThomson (approved):
:lgtm:

> Reviewed 1 of 4 files at r2, 4 of 7 files at r3, 4 of 4 files at r4, 1 of 2 files at r5, 1 of 1 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6252#-:-Np0kDAB2fs2Vj4YJA3U:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on @hatton)

PR #6253, opened 2024-01-17
Don't throw if file doesn't exist in GetLongPath (BL-12986)

andrew-polk:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6253)

### DIFF
--- a/src/BloomExe/Utils/LongPathAware.cs
+++ b/src/BloomExe/Utils/LongPathAware.cs
@@ -55,7 +55,15 @@ namespace Bloom.Utils
 			var result = Convert.ToInt32(GetLongPathName(path, builder, builder.Capacity));
 			if (result == 0)
 			{
-				throw new FileNotFoundException("The file was not found", path);
+				// The original code threw a FileNotFoundException here.
+				// But there are cases where we call this and don't expect the file to exist yet,
+				// such as validating the destination path of a copy is not too long.
+				// If the given path doesn't exist, it certainly isn't a short path which needs
+				// to be converted to a long path, and the best we can do here is simply return
+				// the path we were given. And in the case of a file we are about to create,
+				// returning the given path is always the right thing to do.
+				// See BL-12986.
+				return path;
 			}
 			if (result >= builder.Capacity)
 			{


hatton (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6253)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @andrew-polk)

___
*[`src/BloomExe/Utils/LongPathAware.cs` line 58 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6253#-NoOJab2Cpm_4nvuwiFB:-NoOJab2Cpm_4nvuwiFC:b-dj1gxu) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/d01f9939f5786d7277121bed73cd1db9f6fb710a/src/BloomExe/Utils/LongPathAware.cs#L58)):*
> ```Smalltalk
> 			if (result == 0)
> 			{
> 				// The original code threw a FileNotFoundException here.
> ```

It's not clear to me if this is the right thing to do. @JohnThomson

JohnThomson (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6253)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @hatton and @JohnThomson)

___
*[`src/BloomExe/Utils/LongPathAware.cs` line 58 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6253#-NoOJab2Cpm_4nvuwiFB:-NpW8Zeq9jQBA-2aFC68:b78qw33) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/d01f9939f5786d7277121bed73cd1db9f6fb710a/src/BloomExe/Utils/LongPathAware.cs#L58)):*

It's not clear to me if this is the right thing to do. @JohnThomson
</blockquote></details>

The documentation is infuriatingly unclear, but it seems likely that a file that does not exist can't have a shortened form of its name. Where would it be stored? What could create a short/long name pair except to create a file with a (long) name that is more than 8.3? So if a given name is not the short name of an existing file, it must not be a shortened name at all, and must be the full, long name that we want, even if it has a tilde and is short enough to fit the 8.3 pattern.
It is also not clear why the OS does not follow this logic and return the name unmodified rather than throwing an exception. Maybe there is some pathological case where the name IS a short name of something, but the user does not have permission to know about the file?? But as far as I can tell, what Andrew has is the behavior that makes sense.
I think the comment could be improved. "It seems safer" is pretty vague. Let's say explicitly something along the lines that if a name is not the short name of an existing file, it can't be a short name at all, and must therefore be a full name (presumably of a file we are about to create), so the exception that Windows throws in this situation is unhelpful, and the unchanged path is definitely the answer we want.
(Another thought: the doc I read indicates that this function may expand short names for directories as well as the file name. Would there ever be a case where getting this exception would mean that we should apply the function again to the directory path, and then concatenate the unchanged filename? Then what if the second call throws the same exception? Ugh.)

andrew-polk:
*[`src/BloomExe/Utils/LongPathAware.cs` line 58 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6253#-NoOJab2Cpm_4nvuwiFB:-NpWUIz8AHgpVINdMOxU:b-6f6bwh) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/d01f9939f5786d7277121bed73cd1db9f6fb710a/src/BloomExe/Utils/LongPathAware.cs#L58)):*

The documentation is infuriatingly unclear, but it seems likely that a file that does not exist can't have a shortened form of its name. Where would it be stored? What could create a short/long name pair except to create a file with a (long) name that is more than 8.3? So if a given name is not the short name of an existing file, it must not be a shortened name at all, and must be the full, long name that we want, even if it has a tilde and is short enough to fit the 8.3 pattern.
It is also not clear why the OS does not follow this logic and return the name unmodified rather than throwing an exception. Maybe there is some pathological case where the name IS a short name of something, but the user does not have permission to know about the file?? But as far as I can tell, what Andrew has is the behavior that makes sense.
I think the comment could be improved. "It seems safer" is pretty vague. Let's say explicitly something along the lines that if a name is not the short name of an existing file, it can't be a short name at all, and must therefore be a full name (presumably of a file we are about to create), so the exception that Windows throws in this situation is unhelpful, and the unchanged path is definitely the answer we want.
(Another thought: the doc I read indicates that this function may expand short names for directories as well as the file name. Would there ever be a case where getting this exception would mean that we should apply the function again to the directory path, and then concatenate the unchanged filename? Then what if the second call throws the same exception? Ugh.)
</blockquote></details>

I made an attempt at improving the comment.

PR #6256, opened 2024-01-18
Add data-default-languages-legacy (BL-12975)

JohnThomson:
Adds this attribute for use on fields (currently just .Title-On-Cover-style) that previously used data-default-languages to control which languages should be visible (by adding the class bloom-visibility-code-on). The value in the new attribute, if present, will be used when the book is in legacy mode. Otherwise, the bloom-visibility-code-on class will be removed.
Also removes all reference to bloom-visibility-user-on/off, an idea we introduced with bloom-visibility-code-on but never used.

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6256)

### DIFF
--- a/src/BloomExe/Book/Book.cs
+++ b/src/BloomExe/Book/Book.cs
@@ -580,6 +580,7 @@ namespace Bloom.Book
             TranslationGroupManager.UpdateContentLanguageClasses(
                 dom.RawDom,
                 _bookData,
+                BookInfo.AppearanceSettings.UsingLegacy,
                 Language1Tag,
                 Language2Tag,
                 Language3Tag
@@ -2777,7 +2778,7 @@ namespace Bloom.Book
                     if (ElementIsInXMatter(parent))
                         continue;
                     isXmatterOnly = false;
-                    if (IsLanguageWanted(parent, lang) && !HasContentInLang(parent, lang))
+                    if (IsLanguageWantedInContent(parent, lang) && !HasContentInLang(parent, lang))
                     {
                         result[lang] = false; // not complete
                         break; // no need to check other parents.
@@ -2820,7 +2821,14 @@ namespace Bloom.Book
             return false;
         }
 
-        private bool IsLanguageWanted(XmlElement parent, string lang)
+        /// <summary>
+        /// Determines whether the given language bloom-editable is wanted (should be visible, and therefore typically
+        /// published) in the given parent.
+        /// Currently this is reliable only for elements where visibility is controlled by the old data-default-languages
+        /// attribute rather than the new appearance system. It will need enhancement if we ever need to use it for xmatter,
+        /// or for that matter if the appearance system is used to control visibility of something on content pages.
+        /// </summary>
+        private bool IsLanguageWantedInContent(XmlElement parent, string lang)
         {
             var defaultLangs = parent.GetAttribute("data-default-languages");
             if (String.IsNullOrEmpty(defaultLangs) || defaultLangs == "auto")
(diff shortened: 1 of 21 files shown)


hatton (changes requested):
> Reviewed 21 of 21 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6256)* status: all files reviewed, 2 unresolved discussions (waiting on @JohnThomson)

___
*[`src/BloomExe/Book/Book.cs` line 2848 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6256#-NoTn-3l8pWYEZ0ukbj1:-NoTn-3l8pWYEZ0ukbj2:b-c88t9j) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/7e68bc88aeb644e399d83efd956d4431a42f2231/src/BloomExe/Book/Book.cs#L2848)):*
> ```Smalltalk
>         /// Determines whether the given language bloom-editable is wanted (should be visible, and therefore typically
>         /// published) in the given parent.
>         /// Currently this is reliable only for elements where visibility is controlled by the old data-default-languages
> ```

Maybe reword this like "This only works for....  So only use it for content (not matter)

Then is how about a name change that would suggest that? LIke "IsLanguageUsedInContentPages()" ?
___
*[`src/content/bookLayout/basePage.less` line 538 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6256#-NoTmD9t6O3x2tRAI4SR:-NoTmD9t6O3x2tRAI4SS:bqn7xuq) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/7e68bc88aeb644e399d83efd956d4431a42f2231/src/content/bookLayout/basePage.less#L538)):*
> ```Less
> }
> 
> // Currently we don't have anything that sets this variable, but we needed something to
> ```

Will we know to update this comment if it's not true? Maybe better to say "As of this writing" which may prompt the reader to check it.

JohnThomson (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6256)* status: 19 of 21 files reviewed, 2 unresolved discussions (waiting on @hatton)

___
*[`src/BloomExe/Book/Book.cs` line 2848 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6256#-NoTn-3l8pWYEZ0ukbj1:-NoXbfnJC2BafX0mcfv1:b-lfm40w) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/7e68bc88aeb644e399d83efd956d4431a42f2231/src/BloomExe/Book/Book.cs#L2848)):*

Maybe reword this like "This only works for....  So only use it for content (not matter)

Then is how about a name change that would suggest that? LIke "IsLanguageUsedInContentPages()" ?
</blockquote></details>

Done. (I stuck with 'wanted' because this is only asking about visibility of the fields, not whether it actually has content in that language.)
___
*[`src/content/bookLayout/basePage.less` line 538 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6256#-NoTmD9t6O3x2tRAI4SR:-NoXcWWV755BLFnWs3UX:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/7e68bc88aeb644e399d83efd956d4431a42f2231/src/content/bookLayout/basePage.less#L538)):*

Will we know to update this comment if it's not true? Maybe better to say "As of this writing" which may prompt the reader to check it.
</blockquote></details>

Done.

hatton (commented):
> Reviewed 2 of 2 files at r2.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6256)* status: all files reviewed (commit messages unreviewed), all discussions resolved (waiting on @JohnThomson)

hatton (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6256)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @JohnThomson)

PR #6264, opened 2024-01-26
Get locked settings from xmatter and branding (BL-12980)

JohnThomson:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6264)

### DIFF
--- a/DistFiles/localization/en/BloomMediumPriority.xlf
+++ b/DistFiles/localization/en/BloomMediumPriority.xlf
@@ -324,6 +324,21 @@
         <source xml:lang="en">Problem</source>
         <note>CollectionTab.OnBlorgBadge.Problem</note>
       </trans-unit>
+      <trans-unit id="BookSettings.LockedByBranding" sil:dynamic="true" translate="no">
+        <source xml:lang="en">Locked by {0} Branding</source>
+        <note>BookSettings.LockedByBranding</note>
+        <note>{0} is the name of a Branding pack</note>
+      </trans-unit>
+      <trans-unit id="BookSettings.LockedByLegacy" sil:dynamic="true" translate="no">
+        <source xml:lang="en">Locked by Legacy-5-6 theme</source>
+        <note>BookSettings.LockedByLegacy</note>
+        <note>{0} is the name of a Branding pack</note>
+      </trans-unit>
+      <trans-unit id="BookSettings.LockedByXMatter" sil:dynamic="true" translate="no">
+        <source xml:lang="en">Locked by {0} Front/Back matter</source>
+        <note>BookSettings.LockedByXMatter</note>
+        <note>{0} is the name of an xmatter pack</note>
+      </trans-unit>
       <trans-unit id="BookSettings.Title" sil:dynamic="true">
         <source xml:lang="en">Book Settings</source>
         <note>BookSettings.Title</note>
--- a/src/BloomBrowserUI/.vscode/settings.json
+++ b/src/BloomBrowserUI/.vscode/settings.json
@@ -17,7 +17,10 @@
         "statusBar.background": "#75416E",
         "statusBar.debuggingBackground": "#75416E",
         "statusBar.noFolderBackground": "#75416E",
-        "statussBar.prominentBackground": "#D65649"
+        "statussBar.prominentBackground": "#D65649",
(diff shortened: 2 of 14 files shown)


StephenMcConnel (changes requested):
> Reviewed 14 of 14 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6264)* status: all files reviewed, 1 unresolved discussion (waiting on @JohnThomson)

___
*[`src/BloomExe/Book/AppearanceSettings.cs` line 668 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6264#-NpL-oNL2J6E6kEgcrpQ:-NpL-oNL2J6E6kEgcrpR:b-pqw6q) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/e6894e9b11823994acc150a71bf4efc21ccbc318/src/BloomExe/Book/AppearanceSettings.cs#L668)):*
> ```Smalltalk
>             return;
>         }
> ```

This block could presumably just replace the block checking the css file in the earlier code, that is, move ahead of creating the settings object.  It works here, but wastes whatever effort was used to create the object in harvestermode when it short-circuit returns.
___
*[`src/BloomTests/Book/AppearanceSettingsTests.cs` line 103 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6264#-NpL3bGOB_rfeUFs1V1A:-NpL3bGOB_rfeUFs1V1B:b6k205o) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/e6894e9b11823994acc150a71bf4efc21ccbc318/src/BloomTests/Book/AppearanceSettingsTests.cs#L103)):*
> ```Smalltalk
>             Assert.That(
>                 brandingCss,
>                 Does.Contain("--cover-title-L3-show: doShow-css-will-ignore-this-and-use-default;")
> ```

So for these show variables, "true" in the json file translates to "doShow-css-will-ignore-this-and-use-default" in the css file and "false" translates to "none".  I think this is right according to what I remember of our discussions for variables that affect the CSS display property.

JohnThomson (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6264)* status: 13 of 14 files reviewed, 1 unresolved discussion (waiting on @StephenMcConnel)

___
*[`src/BloomExe/Book/AppearanceSettings.cs` line 668 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6264#-NpL-oNL2J6E6kEgcrpQ:-NpL7yZg0fLggC-qEjy-:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/e6894e9b11823994acc150a71bf4efc21ccbc318/src/BloomExe/Book/AppearanceSettings.cs#L668)):*

This block could presumably just replace the block checking the css file in the earlier code, that is, move ahead of creating the settings object.  It works here, but wastes whatever effort was used to create the object in harvestermode when it short-circuit returns.
</blockquote></details>

Done.
___
*[`src/BloomTests/Book/AppearanceSettingsTests.cs` line 103 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6264#-NpL3bGOB_rfeUFs1V1A:-NpL71nV9O1nvxyb3r2e:b-18xm0) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/e6894e9b11823994acc150a71bf4efc21ccbc318/src/BloomTests/Book/AppearanceSettingsTests.cs#L103)):*

So for these show variables, "true" in the json file translates to "doShow-css-will-ignore-this-and-use-default" in the css file and "false" translates to "none".  I think this is right according to what I remember of our discussions for variables that affect the CSS display property.
</blockquote></details>

I've never understood why JohnH wanted to use "doShow-css-will-ignore-this-and-use-default" rather than 'block'. I suppose it's possible that some such variable will one day be used in a context (e.g., a span) where the default is something other than 'block'

StephenMcConnel (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6264)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @JohnThomson)

PR #6270, opened 2024-01-30
BL-12977 OneDrive error dialog

nabalone:
Catch OneDrive exceptions and give a clear notice with no report button

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6270)

### DIFF
--- a/DistFiles/localization/en/Bloom.xlf
+++ b/DistFiles/localization/en/Bloom.xlf
@@ -5138,6 +5138,10 @@ Do you want to go ahead?</note>
         <note>ID: ReportProblemDialog.WhatDoing</note>
         <note>This is the label for the text field where the user enters what they were doing at the time of the problem.</note>
       </trans-unit>
+      <trans-unit id="ReportProblemDialog.SearchOnline" translate="no">
+        <source xml:lang="en">Search online</source>
+        <note>ID: ReportProblemDialog.SearchOnline</note>
+      </trans-unit>
       <trans-unit id="SamplePrintNotification.IGetIt">
         <source xml:lang="en">I get it. Do not show this again.</source>
         <note>ID: SamplePrintNotification.IGetIt</note>
--- a/DistFiles/localization/en/BloomLowPriority.xlf
+++ b/DistFiles/localization/en/BloomLowPriority.xlf
@@ -185,6 +185,15 @@
             <source xml:lang="en">Bloom was not able to access a microphone.</source>
             <note>ID: EditTab.Toolbox.TalkingBookTool.MicrophoneAccessProblem</note>
           </trans-unit>
+          <trans-unit id="ReportProblemDialog.OneDriveProblem" translate="no">
+            <source xml:lang="en">OneDrive Problem</source>
+            <note>ID: ReportProblemDialog.OneDriveProblem</note>
+            <note>The title of a dialog telling the user about a problem with their Microsoft OneDrive.</note>
+          </trans-unit>
+          <trans-unit id="ReportProblemDialog.OneDriveErrorMessage" translate="no">
+            <source xml:lang="en">There is a problem with your Microsoft OneDrive which is preventing Bloom from accessing files.</source>
+            <note>ID: ReportProblemDialog.OneDriveErrorMessage</note>
+          </trans-unit>
         </body>
     </file>
 </xliff>
--- a/src/BloomBrowserUI/problemDialog/NotifyDialog.tsx
+++ b/src/BloomBrowserUI/problemDialog/NotifyDialog.tsx
@@ -1,4 +1,5 @@
 import * as React from "react";
+import { css } from "@emotion/react";
 import {
     Dialog,
     DialogActions,
@@ -14,17 +15,27 @@ import BloomButton from "../react_components/bloomButton";
 import { makeTheme, kindParams } from "./theme";
 import { useL10n } from "../react_components/l10nHooks";
 import { ProblemKind } from "./ProblemDialog";
+import { hookupLinkHandler } from "../utils/linkHandler";
+import { kBloomBlue, kFormBackground } from "../utils/colorUtils";
 
-export const NotifyDialog: React.FunctionComponent<{
-    reportLabel?: string | null;
-    secondaryLabel?: string | null;
-    message: string | null;
-}> = props => {
-    const theme = makeTheme(ProblemKind.NonFatal);
+export interface INotifyDialogProps {
+    message?: string | null; // The localized message to notify the user about.
+    reportLabel?: string | null; // The localized text that goes on the Report button. Omit or pass "" to disable Report button.
+    secondaryLabel?: string | null; // The localized text that goes on the secondary action button. Omit or pass "" to disable the secondary action button.
+    detailsBoxText?: string | null; // Localized text to go into a grey details box under the message. Omit or pass "" to not show a details box.
+    titleOverride?: string | null; // If present, wil be used in place of the dialog title defined for this level in themes.ts
+    titleL10nKeyOverride?: string | null; // The L10nKey for the titleOverride, if present.
+    themeOverride?: ProblemKind | null; // If present, will be used in place of the dialog theme defined for this level in themes.ts
+}
 
-    const englishTitle = kindParams[ProblemKind.NonFatal].title;
-    const titleKey = kindParams[ProblemKind.NonFatal].l10nKey;
+export const NotifyDialog: React.FC<INotifyDialogProps> = props => {
+    const theme = makeTheme(props.themeOverride || ProblemKind.NonFatal);
+
+    const englishTitle = props.titleOverride ?? kindParams[ProblemKind.NonFatal].title;
+    const titleKey = props.titleL10nKeyOverride ?? kindParams[ProblemKind.NonFatal].l10nKey;
     const localizedDlgTitle = useL10n(englishTitle, titleKey);
+
+    React.useEffect(() => hookupLinkHandler(), []);
 
     const getDialog = () => {
         return (
@@ -36,6 +47,10 @@ export const NotifyDialog: React.FunctionComponent<{
                 maxWidth={"md"}
                 fullScreen={true}
                 onClose={() => post("common/closeReactDialog")}
+                css={css`
+                a {
+                    color: ${kBloomBlue};
+                }`}
             >
                 <DialogTitle className={"dialog-title allowSelect"}>
                     <Typography variant="h6">{localizedDlgTitle}</Typography>
@@ -45,16 +60,29 @@ export const NotifyDialog: React.FunctionComponent<{
                         onClick={() => post("common/closeReactDialog")}
                     /> */}
                 </DialogTitle>
-                <DialogContent className={"dialog-content"}>
-                    {/* InnerHTML is used so that we can insert markup like <br> into the message. */}
-                    <DialogContentText
-                        className="allowSelect"
-                        dangerouslySetInnerHTML={{
-                            __html: props.message || ""
-                        }}
-                    />
-                </DialogContent>
-                {getDialogActionButtons()}
+                    <DialogContent className={"dialog-content"}>
+                        {/* InnerHTML is used so that we can insert markup like <br> into the message. */}
+                        <DialogContentText
+                            className="allowSelect"
+                            dangerouslySetInnerHTML={{
+                                __html: props.message || ""
+                            }}
+                        />
+                        {props.detailsBoxText && 
+                            <DialogContentText
+                                css={css`
+                                    background-color: ${kFormBackground}; 
+                                    padding: 10px; 
+                                    margin-top: 20px; 
+                                    margin-bottom: 20px; 
+                                }`}
+                                dangerouslySetInnerHTML={{
+                                    __html: props.detailsBoxText || ""
+                                }}
+                            />
+                        }
+                    </DialogContent>
+                    {getDialogActionButtons()} 
             </Dialog>
         );
     };
@@ -111,16 +139,21 @@ export const NotifyDialog: React.FunctionComponent<{
     };
 
     const getCloseButton = (): JSX.Element | null => {
+        const buttonLabel = props.themeOverride === ProblemKind.Fatal ? "Quit" : "Close";
+        const l10nKey =
+                props.themeOverride === ProblemKind.Fatal
+                    ? "ReportProblemDialog.Quit" 
+                    : "Common.Close";
         return (
             <BloomButton
                 enabled={true}
-                l10nKey={"Common.Close"}
+                l10nKey={l10nKey}
                 hasText={true}
                 onClick={() => {
                     post("common/closeReactDialog");
                 }}
             >
-                Close
+                {buttonLabel}
             </BloomButton>
         );
     };
--- a/src/BloomBrowserUI/problemDialog/ProblemDialog.tsx
+++ b/src/BloomBrowserUI/problemDialog/ProblemDialog.tsx
@@ -1,6 +1,6 @@
 import * as React from "react";
 import "./ProblemDialog.less";
-import { NotifyDialog } from "./NotifyDialog";
+import { INotifyDialogProps, NotifyDialog } from "./NotifyDialog";
 import { ReportDialog } from "./ReportDialog";
 import { WireUpForWinforms } from "../utils/WireUpWinform";
 
@@ -9,15 +9,10 @@ export enum ProblemKind {
     Notify = "notify",
     User = "user",
     NonFatal = "nonfatal",
-    Fatal = "fatal"
+    Fatal = "fatal",
 }
 
-export const ProblemDialog: React.FunctionComponent<{
-    level: ProblemKind;
-    message: string; // The localized message to notify the user about.
-    reportLabel?: string; // The localized text that goes on the Report button. Omit or pass "" to disable Report button.
-    secondaryLabel?: string; // The localized text that goes on the secondary action button. Omit or pass "" to disable the secondary action button.
-}> = props => {
+export const ProblemDialog: React.FC<{level:ProblemKind} & Partial<INotifyDialogProps>> = props => {
     if (props.level === ProblemKind.Notify) {
         return <NotifyDialog {...props} />;
     } else {
--- a/src/BloomBrowserUI/problemDialog/ReportDialog.tsx
+++ b/src/BloomBrowserUI/problemDialog/ReportDialog.tsx
@@ -105,7 +105,6 @@ export const ReportDialog: React.FunctionComponent<{
                     includeScreenshot
                 },
                 result => {
-                    console.log(JSON.stringify(result.data));
                     const failureResponseString = "failed:";
                     const link = result.data.issueLink;
                     if (link.startsWith(failureResponseString)) {
--- a/src/BloomBrowserUI/problemDialog/theme.ts
+++ b/src/BloomBrowserUI/problemDialog/theme.ts
@@ -29,7 +29,7 @@ export const kindParams = {
         primaryColor: kNonFatalColor,
         title: "Bloom had a problem",
         l10nKey: "ReportProblemDialog.NonFatalTitle"
-    }
+    },
 };
 
 export function makeTheme(kind: ProblemKind): Theme {
--- a/src/BloomBrowserUI/utils/linkHandler.ts
+++ b/src/BloomBrowserUI/utils/linkHandler.ts
@@ -13,7 +13,8 @@ const handler = (e: Event) => {
     if (!href) {
         return;
     }
-    if (href.startsWith("http") || href.startsWith("mailto")) {
+    if (href.startsWith("http") || href.startsWith("mailto") || href.startsWith('file')) {
+        console.log('handling link ', href);
         e.preventDefault();
         e.stopPropagation();
         postString("link", href);
--- a/src/BloomExe/Book/BookInfo.cs
+++ b/src/BloomExe/Book/BookInfo.cs
@@ -745,7 +745,7 @@ namespace Bloom.Book
         /// This method recurses through the folders under 'pathToDirectory' and keeps track of all the unique bookInstanceId
         /// guids. When a duplicate is found, we will call InstallFreshInstanceGuid().
         /// </summary>
-        public static void RepairDuplicateInstanceIds(
+        public static void CheckForDuplicateInstanceIdsAndRepair(
             string pathToDirectory,
             Func<string, bool> okToChangeId
         )
@@ -1006,7 +1006,15 @@ namespace Bloom.Book
         {
             if (RobustFile.Exists(metaDataPath))
             {
-                var id = GetIdFromDamagedMetaDataString(RobustFile.ReadAllText(metaDataPath));
+                string id;
+                try
+                {
+                    id = GetIdFromDamagedMetaDataString(RobustFile.ReadAllText(metaDataPath));
+                }
+                catch (IOException error)
+                {
+                    throw new FileException(metaDataPath, error);
+                }
                 if (id != null)
                     return new BookMetaData() { Id = id };
             }
--- a/src/BloomExe/Collection/CollectionSettings.cs
+++ b/src/BloomExe/Collection/CollectionSettings.cs
@@ -11,6 +11,7 @@ using Bloom.Api;
 using Bloom.Book;
 using Bloom.MiscUI;
 using Bloom.Publish.BloomPub;
+using Bloom.Utils;
 using Bloom.web.controllers;
 using DesktopAnalytics;
 using L10NSharp;
@@ -479,7 +480,7 @@ namespace Bloom.Collection
 
                 LoadDictionary(xml, "Palette", ColorPalettes);
             }
-            catch (Exception)
+            catch (Exception originalError)
             {
                 string settingsContents;
                 try
@@ -495,7 +496,8 @@ namespace Bloom.Collection
                 // We used to notify the user of a problem here.
                 // But now we decided it is better to catch at a higher level, at OpenProjectWindow(), else we have two different
                 // error UI dialogs for the same problem. See BL-9916.
-                throw;
+
+                throw new FileException(SettingsFilePath, originalError);
             }
 
             try
--- a/src/BloomExe/CollectionTab/CollectionTabView.cs
+++ b/src/BloomExe/CollectionTab/CollectionTabView.cs
@@ -278,19 +278,19 @@ namespace Bloom.CollectionTab
                         // However, until we get rid of the old collection tab, it's tricky to move it, so I've just
                         // duplicated it here.
                         // Doing it at this point seems to work fine.
-                        RepairDuplicates();
+                        CheckForDuplicatesAndRepair();
                         Controls.Add(_reactControl);
                     }
                 )
             );
         }
 
-        private void RepairDuplicates()
+        private void CheckForDuplicatesAndRepair()
         {
             var collectionPath = _model.TheOneEditableCollection.PathToDirectory;
             // A book's ID may not be changed if we have a TC and the book is actually in the shared folder.
             // Eventually we may allow it if the book is checked out.
-            BookInfo.RepairDuplicateInstanceIds(
+            BookInfo.CheckForDuplicateInstanceIdsAndRepair(
                 collectionPath,
                 (bookPath) =>
                 {
--- a/src/BloomExe/ErrorReporter/HtmlErrorReporter.cs
+++ b/src/BloomExe/ErrorReporter/HtmlErrorReporter.cs
@@ -4,6 +4,7 @@ using System.Windows.Forms;
 using Bloom.Api;
 using Bloom.MiscUI;
 using Bloom.ToPalaso;
+using Bloom.Utils;
 using Bloom.web.controllers;
 using SIL.IO;
 using SIL.Reporting;
@@ -217,6 +218,14 @@ namespace Bloom.ErrorReporter
 
             try
             {
+                string filePath = FileException.GetFilePathIfPresent(exception);
+                // FileException is a Bloom exception to capture the filepath. We want to report the inner, original exception.
+                Exception originalException = FileException.UnwrapIfFileException(exception);
+                if (OneDriveUtils.CheckForAndHandleOneDriveExceptions(originalException, filePath))
+                {
+                    return;
+                }
+
                 if (policy == null)
                 {
                     policy = new ShowAlwaysPolicy();
@@ -227,7 +236,7 @@ namespace Bloom.ErrorReporter
                     ShowNotifyDialog(
                         ProblemLevel.kNotify,
                         message,
-                        exception,
+                        originalException,
                         reportButtonLabel,
                         onReportButtonClicked,
                         extraButtonLabel,
--- a/src/BloomExe/MiscUI/StartupScreenManager.cs
+++ b/src/BloomExe/MiscUI/StartupScreenManager.cs
@@ -319,9 +319,14 @@ namespace Bloom.MiscUI
 
         public static void CloseSplashScreen()
         {
-            _splashForm?.FadeAndClose(); //it's going to hang around while it fades,
-            _doWhenSplashScreenShouldClose?.Invoke();
-            DoLastOfAllAfterClosingSplashScreen?.Invoke();
+            if (_splashForm == null || !_splashForm.IsDisposed)
+            {
+                // We just want to skip the following steps if _splashForm is there but disposed, as this will cause errors.
+                // They should behave properly if _splashForm is null.
+                _splashForm?.FadeAndClose(); //it's going to hang around while it fades,
+                _doWhenSplashScreenShouldClose?.Invoke();
+                DoLastOfAllAfterClosingSplashScreen?.Invoke();
+            }
 
             _splashForm = null;
             _doWhenSplashScreenShouldClose = null; // Do these actions only once.
--- a/src/BloomExe/Program.cs
+++ b/src/BloomExe/Program.cs
@@ -39,6 +39,7 @@ using SIL.Xml;
 using Microsoft.Web.WebView2.Core;
 using System.Text;
 using Bloom.Utils;
+using Bloom.web.controllers;
 
 namespace Bloom
 {
@@ -1297,6 +1298,22 @@ namespace Bloom
                 _projectContext = null;
             }
 
+            // FileException is a Bloom exception to capture the filepath. We want to report the inner, original exception.
+            Exception originalError = FileException.UnwrapIfFileException(error);
+            string errorFilePath = FileException.GetFilePathIfPresent(error);
+            Logger.WriteError(
+                $"*** Error loading collection {Path.GetFileNameWithoutExtension(projectPath)}, on filepath: {errorFilePath}",
+                originalError
+            );
+
+            // Normally, NotifyUserOfProblem would take an exception and do this special-exception processing for us.
+            // But in this case, we don't pass the exception to NotifyUserOfProblem because we may subsequently end up
+            // calling SendReportWithoutUI. Therefore, we must check for the special exception independently.
+            if (OneDriveUtils.CheckForAndHandleOneDriveExceptions(error))
+            {
+                return;
+            }
+
             ErrorResult reportPressedResult = ErrorResult.Yes;
             // NB: I added the email to this directly because, at least on my machine, the old error report dialog had become unworkable
             // because, presumably, I haven't set things up properly with gmail.
@@ -1322,7 +1339,7 @@ namespace Bloom
                     var additionalPathsToInclude = Directory.GetFiles(dirName);
                     _applicationContainer.ProblemReportApi.SendReportWithoutUI(
                         ProblemLevel.kNonFatal,
-                        error,
+                        originalError,
                         errorMessage,
                         "",
                         additionalPathsToInclude
@@ -1333,7 +1350,7 @@ namespace Bloom
                     // No email... just fallback to the WinFormsErrorReporter, which will allow the user to email us.
                     // Unfortunately, we won't be able to automatically get the .bloomCollection file from that.
                     SIL.Reporting.ErrorReport.ReportNonFatalExceptionWithMessage(
-                        error,
+                        originalError,
                         errorMessage
                     );
                 }
--- a/src/BloomExe/Utils/FileException.cs
+++ b/src/BloomExe/Utils/FileException.cs
@@ -1,0 +1,42 @@
+﻿using System;
+
+namespace Bloom.Utils
+{
+    /// <summary>
+    /// Attach the filepath to file-specific exceptions, e.g. errors accessing that file
+    /// </summary>
+    public class FileException : Exception
+    {
+        public string FilePath;
+
+        // When WinForms catches exceptions thrown from invoked methods, it only gets the innermost exception
+        // So we can't put the original exception in innerException or this FilePath will get lost in such cases
+        // See https://stackoverflow.com/questions/15668334/preserving-exceptions-from-dynamically-invoked-methods
+        // https://stackoverflow.com/questions/347502/why-does-the-inner-exception-reach-the-threadexception-handler-and-not-the-actua
+        public Exception OriginalException;
+
+        public FileException(string filePath, Exception originalException)
+        {
+            FilePath = filePath;
+            OriginalException = originalException;
+        }
+
+        public static string GetFilePathIfPresent(Exception exception)
+        {
+            if (exception is FileException)
+            {
+                return ((FileException)exception).FilePath;
+            }
+            return null;
+        }
+
+        public static Exception UnwrapIfFileException(Exception exception)
+        {
+            if (exception is FileException)
+            {
+                return ((FileException)exception).OriginalException;
+            }
+            return exception;
+        }
+    }
+}
--- a/src/BloomExe/Utils/OneDriveUtils.cs
+++ b/src/BloomExe/Utils/OneDriveUtils.cs
@@ -1,0 +1,181 @@
+using System;
+using System.Linq;
+using System.Windows.Forms;
+using Bloom.Api;
+using Bloom.ErrorReporter;
+using Bloom.web.controllers;
+using SIL.Reporting;
+
+namespace Bloom.Utils
+{
+    public static class OneDriveUtils
+    {
+        private static string[] oneDriveErrorCodes =
+        {
+            "80070184", // This one is not in Microsoft's list. I got it by disconnecting my internet.
+            // The rest are from https://support.microsoft.com/en-au/office/what-do-the-onedrive-error-codes-mean-f7a68338-e540-4ebf-ad5d-56c5633acded#ID0EBBH=Error_codes
+            "80010007",
+            "80040c81",
+            "8004de80",
+            "8004de86",
+            "8004de85",
+            "8004de8a",
+            "8004de90",
+            "8004de96",
+            "8004ded7",
+            "8004dedc",
+            "8004def0",
+            "8004def7",
+            "80070005",
+            "8007016a",
+            "80070194",
+            "80071128",
+            "80071129"
+        };
+
+        public static bool IsOneDriveErrorCode(string errorCode)
+        {
+            return oneDriveErrorCodes.Contains(errorCode.ToLowerInvariant());
+        }
+
+        public static string GetOneDriveErrorDialogMessage()
+        {
+            return L10NSharp.LocalizationManager.GetString(
+                "ReportProblemDialog.OneDriveErrorMessage",
+                "There is a problem with your Microsoft OneDrive which is preventing Bloom from accessing files."
+            );
+        }
+
+        public static string GetOneDriveErrorDialogHeader()
+        {
+            return L10NSharp.LocalizationManager.GetString(
+                "ReportProblemDialog.OneDriveProblem",
+                "OneDrive Problem"
+            );
+        }
+
+        /// <summary>
+        /// We want to show different UI for exceptions caused by the user's OneDrive, which we cannot do anything about
+        /// And should inform the user about. See BL-12977
+        /// Recursively checks inner exceptions.
+        /// </summary>
+        /// <param name="error">The exception to check. We will check its inner exceptions also.</param>
+        /// <param name="filePath">Path of file that we were trying to access, causing this error. To display to the user</param>
+        /// <param name="levelOfProblem">"fatal" or "nonfatal", for UI display</param>
+        public static bool CheckForAndHandleOneDriveExceptions(
+            System.Exception error,
+            string filePath = "",
+            string levelOfProblem = "nonfatal"
+        )
+        {
+            if (error == null)
+                return false;
+
+            string fileExceptionFilePath = FileException.GetFilePathIfPresent(error);
+            if (!string.IsNullOrEmpty(fileExceptionFilePath))
+            {
+                filePath = fileExceptionFilePath;
+            }
+            // FileException is a Bloom exception to capture the filepath. We want to report the inner, original exception.
+            Exception originalError = FileException.UnwrapIfFileException(error);
+
+            string errorCode = originalError.HResult.ToString("X");
+
+            if (!IsOneDriveErrorCode(errorCode))
+            {
+                // This exception is not one of the OneDrive exception codes we check for,
+                // but check if it has an inner exception that is
+                if (originalError?.InnerException != null)
+                {
+                    return CheckForAndHandleOneDriveExceptions(
+                        originalError.InnerException,
+                        filePath
+                    );
+                }
+                else
+                {
+                    // this is not a special error code, return to normal error handling
+                    return false;
+                }
+            }
+
+            Control control = Shell.GetShellOrOtherOpenForm();
+            if (control == null) // still possible if we come from a "Details" button
+                control = FatalExceptionHandler.ControlOnUIThread;
+
+            string logPath = Logger.LogPath.Replace('\\', '/');
+
+            string errorMessageLabel = "Error Message";
+            string errorCodeLabel = "Error Number";
+            string filePathLabel = "File";
+            string logPathLabel = "Bloom Log";
+            void ShowOneDriveExceptionFallbackDialog(
+                Exception error,
+                string errorCode,
+                string filePath,
+                string logPath
+            )
+            {
+                // We decided not to localize this for now.
+                MessageBox.Show(
+                    GetOneDriveErrorDialogMessage()
+                        + Environment.NewLine
+                        + Environment.NewLine
+                        + $"{errorMessageLabel}: {error.Message}{Environment.NewLine}"
+                        + $"{errorCodeLabel}: {errorCode}{Environment.NewLine}"
+                        + $"{filePathLabel}: {filePath}{Environment.NewLine}"
+                        + $"{logPathLabel}: {logPath}",
+                    GetOneDriveErrorDialogHeader(),
+                    MessageBoxButtons.OK,
+                    MessageBoxIcon.Error
+                );
+            }
+
+            if (BloomServer._theOneInstance == null)
+            {
+                // We got an error really early, before we can use HTML dialogs
+                ShowOneDriveExceptionFallbackDialog(originalError, errorCode, filePath, logPath);
+
+                return true;
+            }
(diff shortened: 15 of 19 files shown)


nabalone (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6270)* status: 0 of 15 files reviewed, 5 unresolved discussions

___
*[`src/BloomBrowserUI/problemDialog/ProblemDialog.tsx` line 23 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6270#-NpQwP3VA5btCEtpt3yr:-NpQwP3VA5btCEtpt3ys:bhotm68) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/403c999f44b1b394c175c0b2751d9a801b4e6072/src/BloomBrowserUI/problemDialog/ProblemDialog.tsx#L23)):*
> ```XML
>     secondaryLabel?: string; // The localized text that goes on the secondary action button. Omit or pass "" to disable the secondary action button.
> }> = props => {
>     if ([ProblemKind.Notify, ProblemKind.OneDriveNonFatal, ProblemKind.OneDriveFatal].includes(props.level)) { // TODO? is there a better way to do this?
> ```

Would it be better to do 
```
if (
    props.level === ProblemKind.Notify 
    || props.level === ProblemKind. OneDriveNonFatal
    || props.level === ProblemKind.ONeDriveFatal
)
```
___
*[`src/BloomBrowserUI/utils/linkHandler.ts` line 16 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6270#-NpQwml12Z2Oib74cRRW:-NpQwml21jJeAHNIdLuz:b-blnkv2) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/403c999f44b1b394c175c0b2751d9a801b4e6072/src/BloomBrowserUI/utils/linkHandler.ts#L16)):*
> ```XML
>         return;
>     }
>     if (href.startsWith("http") || href.startsWith("mailto") || href.startsWith('file')) { // TODO? will it cause problems elsewhere to add file here?
> ```

I looked around and didn't see anything that looked like it would break because we changed `file://` handling but wasnt sure
___
*[`src/BloomExe/Book/BookInfo.cs` line 1014 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6270#-NpQwvvh4ayVWQVnqxO8:-NpQwvvh4ayVWQVnqxO9:b13lf72) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/403c999f44b1b394c175c0b2751d9a801b4e6072/src/BloomExe/Book/BookInfo.cs#L1014)):*
> ```Smalltalk
>                     id = GetIdFromDamagedMetaDataString(RobustFile.ReadAllText(metaDataPath));
>                 }
>                 // TODO? should I replace all calls to RobustFile.Read... with wrapped versions?
> ```

i.e. make method(s) which call the `RobustFile.ReadAll()` functions but wrap them in a try/catch and convert IOExceptions to FileExceptions? And then put those everywhere? yagni?
___
*[`src/BloomExe/web/controllers/ProblemReportApi.cs` line 742 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6270#-NpQxJOtDVLs-bIXN8if:-NpQxJOu-hXWVE-t0AnJ:bu1g9cr) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/403c999f44b1b394c175c0b2751d9a801b4e6072/src/BloomExe/web/controllers/ProblemReportApi.cs#L742)):*
> ```Smalltalk
>             // }
> 
>             // TODO? is it worth splitting?
> ```

There's a lot of code shared between `ShowProblemDialog` and `CheckForAndHandleOneDriveExceptions`. I was trying to factor out the repeated dialog related code, but it would mean a helper method called twice that takes in a lot of parameters. Maybe we want to be able to arbitrarily adjust the two dialog behaviors independently of each other anyway?
___
*[`src/BloomExe/web/controllers/ProblemReportApi.cs` line 862 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6270#-NpQxoypAhMbjQihPctQ:-NpQxoypAhMbjQihPctR:b-b47flm) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/403c999f44b1b394c175c0b2751d9a801b4e6072/src/BloomExe/web/controllers/ProblemReportApi.cs#L862)):*
> ```Smalltalk
>         )
>         {
>             // TODO? do we want to localize any of these?
> ```

I am generally not sure how much is worth localizing

andrew-polk (changes requested):
> Reviewed 14 of 15 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6270)* status: 14 of 15 files reviewed, 18 unresolved discussions (waiting on @nabalone)

___
*[`DistFiles/localization/en/Bloom.xlf` line 5140 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6270#-NpVNbqK7A6X8YY9gUuo:-NpVNbqK7A6X8YY9gUup:b-9lvbch) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/403c999f44b1b394c175c0b2751d9a801b4e6072/DistFiles/localization/en/Bloom.xlf#L5140)):*
> ```XML
>         <note>ID: ReportProblemDialog.WhatDoing</note>
>         <note>This is the label for the text field where the user enters what they were doing at the time of the problem.</note>
>       </trans-unit>
> ```

Generally, we are trying not to add things to Bloom.xlf. 
New strings should be added to the -Medium or -Low versions.
(I guess high-priority ones still go in Bloom.xlf??)

These can probably go in -Low.
___
*[`DistFiles/localization/en/Bloom.xlf` line 5146 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6270#-NpVNGxf2FAiV3MUx60j:-NpVNGxf2FAiV3MUx60k:betd1dg) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/403c999f44b1b394c175c0b2751d9a801b4e6072/DistFiles/localization/en/Bloom.xlf#L5146)):*
> ```XML
>       </trans-unit>
>       <trans-unit id="ReportProblemDialog.OneDriveProblem" translate="no">
>         <source xml:lang="en">OneDriveProblem</source>
> ```

Shouldn't this source string be "OneDrive Problem"?
___
*[`src/BloomBrowserUI/problemDialog/NotifyDialog.tsx` line 27 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6270#-NpVYa21-FJfzdwdbLTw:-NpVYa21-FJfzdwdbLTx:b3lael) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/403c999f44b1b394c175c0b2751d9a801b4e6072/src/BloomBrowserUI/problemDialog/NotifyDialog.tsx#L27)):*
> console.log(props);

Let's remove this.
___
*[`src/BloomBrowserUI/problemDialog/NotifyDialog.tsx` line 67 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6270#-NpVZ8qO6bkc_CorDG6i:-NpVZ8qO6bkc_CorDG6j:bbtdnxx) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/403c999f44b1b394c175c0b2751d9a801b4e6072/src/BloomBrowserUI/problemDialog/NotifyDialog.tsx#L67)):*
> #f0f0f0

I think it would be an improvement to use colorUtils' kFormBackground.
___
*[`src/BloomBrowserUI/problemDialog/ProblemDialog.tsx` line 23 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6270#-NpQwP3VA5btCEtpt3yr:-NpVmBF22zdjcayyr_HB:b-v8atho) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/403c999f44b1b394c175c0b2751d9a801b4e6072/src/BloomBrowserUI/problemDialog/ProblemDialog.tsx#L23)):*

Would it be better to do 
```
if (
    props.level === ProblemKind.Notify 
    || props.level === ProblemKind. OneDriveNonFatal
    || props.level === ProblemKind.ONeDriveFatal
)
```
</blockquote></details>

Either syntax is fine.
But as I say below, let's discuss the idea of having new "kinds" for OneDrive.
___
*[`src/BloomBrowserUI/problemDialog/theme.ts` line 41 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6270#-NpVO7NNCxr96WIohDRH:-NpVO7NOEZjf1JNCeQp2:bj73q1e) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/403c999f44b1b394c175c0b2751d9a801b4e6072/src/BloomBrowserUI/problemDialog/theme.ts#L41)):*
> #f44336

Since this is designed to be the same as the fatal color, let's extract that code as a const.
___
*[`src/BloomBrowserUI/utils/linkHandler.ts` line 16 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6270#-NpQwml12Z2Oib74cRRW:-NpVOPBLE8pLOldnDAu_:b-db7qxm) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/403c999f44b1b394c175c0b2751d9a801b4e6072/src/BloomBrowserUI/utils/linkHandler.ts#L16)):*

I looked around and didn't see anything that looked like it would break because we changed `file://` handling but wasnt sure
</blockquote></details>

I'm not sure what else we can do but look around (as you did) and test.
Let's remove the TODO comment.
___
*[`src/BloomExe/Program.cs` line 1301 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6270#-NpVjqfb7meRYd9sCU9u:-NpVjqfb7meRYd9sCU9v:b-la5uks) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/403c999f44b1b394c175c0b2751d9a801b4e6072/src/BloomExe/Program.cs#L1301)):*
> string filePath

Looks like this isn't used.
___
*[`src/BloomExe/Program.cs` line 1306 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6270#-NpVnNOdBuJclYiEFlH6:-NpVnNOe3ctCgxDeAxQW:b-ahvzdm) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/403c999f44b1b394c175c0b2751d9a801b4e6072/src/BloomExe/Program.cs#L1306)):*
> ```Smalltalk
> 
>             // Normally NotifyUserOfProblem would check for special exceptions and handle them appropriately.
>             // But, the code below doesn't use the usual NotifyUserOfProblem path (because we may want to SendReportWithoutUI below)
> ```

Sorry; I'm not processing this comment well.
Is this accurate?
Normally, NotifyUserOfProblem would take an exception and do this special-exception processing for us. But in this case, we don't pass the exception to NotifyUserOfProblem because we may subsequently end up calling SendReportWithoutUI. Therefore, we must check for the special exception independently.
___
*[`src/BloomExe/Program.cs` line 1307 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6270#-NpVmc8UEswVskybN5Dq:-NpVmc8UEswVskybN5Dr:bmz0fz9) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/403c999f44b1b394c175c0b2751d9a801b4e6072/src/BloomExe/Program.cs#L1307)):*
> There

Is this supposed to be "Therefore"?
___
*[`src/BloomExe/Book/BookInfo.cs` line 1014 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6270#-NpQwvvh4ayVWQVnqxO8:-NpVhLVu2DcLc2YWxR4w:b-q8sycb) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/403c999f44b1b394c175c0b2751d9a801b4e6072/src/BloomExe/Book/BookInfo.cs#L1014)):*

i.e. make method(s) which call the `RobustFile.ReadAll()` functions but wrap them in a try/catch and convert IOExceptions to FileExceptions? And then put those everywhere? yagni?
</blockquote></details>

Putting them everywhere is definitely out of scope from what I understood.
If there was more than one place we were specifically wrapping a RobustFile. call, I would probably say it was worth a wrapper. But looks like other place you throw FileException is wrapping a bunch of stuff.
So, I could go either way. If you think there will likely be other places we wrap RobustFile. calls, specifically, you could make a wrapper now. But we could also wait until we do.
___
*[`src/BloomExe/Collection/CollectionSettings.cs` line 500 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6270#-NpVi5Jg-MqvGAVjS6s5:-NpVi5Jg-MqvGAVjS6s6:bf2xg5d) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/403c999f44b1b394c175c0b2751d9a801b4e6072/src/BloomExe/Collection/CollectionSettings.cs#L500)):*
> outerError

Rather than outer, isn't this inner? Or maybe "original" is better? (That matches the constructor signature for FileException.)
___
*[`src/BloomExe/CollectionTab/CollectionTabView.cs` line 288 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6270#-NpVOjrx82ymjFrMNCae:-NpVOjrx82ymjFrMNCaf:buh4ndv) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/403c999f44b1b394c175c0b2751d9a801b4e6072/src/BloomExe/CollectionTab/CollectionTabView.cs#L288)):*
> ```Smalltalk
>         }
> 
>         private void CheckForDuplicates()
> ```

Assuming that it actually does, how about `CheckForDuplicatesAndRepair`? Or similar.

Same for BookInfo's version.
___
*[`src/BloomExe/ErrorReporter/HtmlErrorReporter.cs` line 234 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6270#-NpVif9zDSrY8rUeSSad:-NpVif9zDSrY8rUeSSae:b-kp9des) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/403c999f44b1b394c175c0b2751d9a801b4e6072/src/BloomExe/ErrorReporter/HtmlErrorReporter.cs#L234)):*
> ```Smalltalk
>                     {
>                         System.Threading.Monitor.Exit(_lock);
>                     }
> ```

```
                    if (!wasAlreadyLocked)
                    {
                        System.Threading.Monitor.Exit(_lock);
                    }

```
Unless I'm missing something, this isn't needed. The finally block will do it.
___
*[`src/BloomExe/Utils/FileException.cs` line 24 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6270#-NpVjyjx3opgCy13_7gj:-NpVjyjx3opgCy13_7gk:bfrdiut) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/403c999f44b1b394c175c0b2751d9a801b4e6072/src/BloomExe/Utils/FileException.cs#L24)):*
> ```Smalltalk
>         }
> 
>         public static string FilePathIfPresent(Exception exception)
> ```

Should this be GetFilePathIfPresent?
___
*[`src/BloomExe/Utils/FileException.cs` line 30 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6270#-NpVjG0x6EtaEtje6BSj:-NpVjG0x6EtaEtje6BSk:b-hib5in) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/403c999f44b1b394c175c0b2751d9a801b4e6072/src/BloomExe/Utils/FileException.cs#L30)):*
> ```Smalltalk
>                 return ((FileException)exception).FilePath;
>             }
>             return "";
> ```

Is there a reason this is empty string rather than null?
___
*[`src/BloomExe/web/controllers/ProblemReportApi.cs` line 742 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6270#-NpQxJOtDVLs-bIXN8if:-NpVldQDA5-PIgFQd9Xe:b-ccidy8) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/403c999f44b1b394c175c0b2751d9a801b4e6072/src/BloomExe/web/controllers/ProblemReportApi.cs#L742)):*

There's a lot of code shared between `ShowProblemDialog` and `CheckForAndHandleOneDriveExceptions`. I was trying to factor out the repeated dialog related code, but it would mean a helper method called twice that takes in a lot of parameters. Maybe we want to be able to arbitrarily adjust the two dialog behaviors independently of each other anyway?
</blockquote></details>

Let's discuss. 
It may be part of a bigger discussion which is worth having about having different report types just based on OneDrive or not.
___
*[`src/BloomExe/web/controllers/ProblemReportApi.cs` line 862 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6270#-NpQxoypAhMbjQihPctQ:-NpVlJ4G3B-Vug37m-TR:bpcskqi) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/403c999f44b1b394c175c0b2751d9a801b4e6072/src/BloomExe/web/controllers/ProblemReportApi.cs#L862)):*

I am generally not sure how much is worth localizing
</blockquote></details>

Yeah, always a hard call.
Doesn't seem worth it at this time.
Maybe worth replacing the TODO comment with something like
// We decided not to localize this for now.

andrew-polk (changes requested):
> Reviewed 15 of 16 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6270)* status: 16 of 17 files reviewed, 6 unresolved discussions (waiting on @nabalone)

___
*[`src/BloomBrowserUI/problemDialog/ProblemDialog.tsx` line 20 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6270#-NpWxOcN1R-4ESx5R1yu:-NpWxOcN1R-4ESx5R1yv:bmshbw3) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/76b64926c1bded2d6f4f70f154303c6127176ea9/src/BloomBrowserUI/problemDialog/ProblemDialog.tsx#L20)):*
> ```XML
>     reportLabel?: string; // The localized text that goes on the Report button. Omit or pass "" to disable Report button.
>     secondaryLabel?: string; // The localized text that goes on the secondary action button. Omit or pass "" to disable the secondary action button.
>     notifyOnly?: boolean; // If true, use the NotifyDialog with no report button regardless of the level.
> ```

According to the comment on reportLabel, leaving that off should be enough to prevent the report button.
So if that isn't true, we should either make it true and remove this prop, or update the comment.
___
*[`src/BloomExe/ErrorReporter/HtmlErrorReporter.cs` line 231 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6270#-NpWxJaZ5LaTdSlF-rwS:-NpWxJaZ5LaTdSlF-rwT:b19zeac) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/76b64926c1bded2d6f4f70f154303c6127176ea9/src/BloomExe/ErrorReporter/HtmlErrorReporter.cs#L231)):*
> ```Smalltalk
>                 )
>                 {
>                     return; // TODO make sure we are letting go of the lock
> ```

// TODO

I know you're still working on this. I'm just adding this to make sure it doesn't get lost or left in.
___
*[`src/BloomExe/Utils/FileException.cs` line 24 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6270#-NpVjyjx3opgCy13_7gj:-NpWwGlc3E_hCGTSbMgb:b-f6pdg8) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/403c999f44b1b394c175c0b2751d9a801b4e6072/src/BloomExe/Utils/FileException.cs#L24)):*

Should this be GetFilePathIfPresent?
</blockquote></details>

Oops. Should be capital `Get`
___
*[`src/BloomExe/Utils/OneDriveUtils.cs` line 30 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6270#-NpWwx-a1iPOX0qmM25b:-NpWwx-a1iPOX0qmM25c:bmdtzvn) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/76b64926c1bded2d6f4f70f154303c6127176ea9/src/BloomExe/Utils/OneDriveUtils.cs#L30)):*
> ```Smalltalk
>         };
> 
>         public static bool isOneDriveErrorCode(string errorCode)
> ```

Remember, unlike js/ts, in C#, we capitalize method names.
___
*[`src/BloomExe/Utils/OneDriveUtils.cs` line 40 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6270#-NpWx5SA36XTziH7FeV9:-NpWx5SA36XTziH7FeVA:b19zeac) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/76b64926c1bded2d6f4f70f154303c6127176ea9/src/BloomExe/Utils/OneDriveUtils.cs#L40)):*
> ```Smalltalk
>                     "ReportProblemDialog.OneDriveErrorMessage",
>                     "There is a problem with your Microsoft OneDrive which is preventing Bloom from accessing files."
>                 ) + Environment.NewLine; // TODO I thought I got rid of this...
> ```

// TODO

I know you're still working on this. I'm just adding this to make sure it doesn't get lost or left in.

nabalone (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6270)* status: 12 of 18 files reviewed, 5 unresolved discussions (waiting on @andrew-polk)

___
*[`DistFiles/localization/en/Bloom.xlf` line 5140 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6270#-NpVNbqK7A6X8YY9gUuo:-NpW64RjEcHeGPW4KGNf:b-5kltb1) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/403c999f44b1b394c175c0b2751d9a801b4e6072/DistFiles/localization/en/Bloom.xlf#L5140)):*

Generally, we are trying not to add things to Bloom.xlf. 
New strings should be added to the -Medium or -Low versions.
(I guess high-priority ones still go in Bloom.xlf??)

These can probably go in -Low.
</blockquote></details>

Oh ok, Done. I was just trying to keep them together with other problem dialog text.
___
*[`DistFiles/localization/en/Bloom.xlf` line 5146 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6270#-NpVNGxf2FAiV3MUx60j:-NpW6ZKr5oXeleyXaiba:b-nrfmh0) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/403c999f44b1b394c175c0b2751d9a801b4e6072/DistFiles/localization/en/Bloom.xlf#L5146)):*

Shouldn't this source string be "OneDrive Problem"?
</blockquote></details>

Done. oops
___
*[`src/BloomBrowserUI/problemDialog/NotifyDialog.tsx` line 27 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6270#-NpVYa21-FJfzdwdbLTw:-NpW6d73C8nrmx4qRxQF:b-r2w617) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/403c999f44b1b394c175c0b2751d9a801b4e6072/src/BloomBrowserUI/problemDialog/NotifyDialog.tsx#L27)):*

> console.log(props);

Let's remove this.
</blockquote></details>

Done. Oops 🤦‍♀️I searched my diff for leftover TODOs but forgot to search for console.logs. Sorry about that
___
*[`src/BloomBrowserUI/problemDialog/NotifyDialog.tsx` line 67 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6270#-NpVZ8qO6bkc_CorDG6i:-NpW7G53CC6mdXKBS6EJ:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/403c999f44b1b394c175c0b2751d9a801b4e6072/src/BloomBrowserUI/problemDialog/NotifyDialog.tsx#L67)):*

> #f0f0f0

I think it would be an improvement to use colorUtils' kFormBackground.
</blockquote></details>

Done.
___
*[`src/BloomBrowserUI/problemDialog/ProblemDialog.tsx` line 20 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6270#-NpWxOcN1R-4ESx5R1yu:-Np_w_KY6YMiH19Bcu3p:b5e7lda) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/76b64926c1bded2d6f4f70f154303c6127176ea9/src/BloomBrowserUI/problemDialog/ProblemDialog.tsx#L20)):*

According to the comment on reportLabel, leaving that off should be enough to prevent the report button.
So if that isn't true, we should either make it true and remove this prop, or update the comment.
</blockquote></details>

Ah ok, I tried to clarify
___
*[`src/BloomExe/Program.cs` line 1301 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6270#-NpVjqfb7meRYd9sCU9u:-NpW9BEP6d0JK3k0VJvx:b-12utm) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/403c999f44b1b394c175c0b2751d9a801b4e6072/src/BloomExe/Program.cs#L1301)):*

> string filePath

Looks like this isn't used.
</blockquote></details>

Done. But do we eventually want to add filePaths to the error message or log now that we might have them here
___
*[`src/BloomExe/Program.cs` line 1306 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6270#-NpVnNOdBuJclYiEFlH6:-NpWBfi_7jqfWFCKLmjE:b-35y33k) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/403c999f44b1b394c175c0b2751d9a801b4e6072/src/BloomExe/Program.cs#L1306)):*

Sorry; I'm not processing this comment well.
Is this accurate?
Normally, NotifyUserOfProblem would take an exception and do this special-exception processing for us. But in this case, we don't pass the exception to NotifyUserOfProblem because we may subsequently end up calling SendReportWithoutUI. Therefore, we must check for the special exception independently.
</blockquote></details>

Done. Yes, that is what I was trying to say! much better
___
*[`src/BloomExe/Book/BookInfo.cs` line 1014 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6270#-NpQwvvh4ayVWQVnqxO8:-NpWD-IsEpoTtawz0PuK:b-lvjp34) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/403c999f44b1b394c175c0b2751d9a801b4e6072/src/BloomExe/Book/BookInfo.cs#L1014)):*

Putting them everywhere is definitely out of scope from what I understood.
If there was more than one place we were specifically wrapping a RobustFile. call, I would probably say it was worth a wrapper. But looks like other place you throw FileException is wrapping a bunch of stuff.
So, I could go either way. If you think there will likely be other places we wrap RobustFile. calls, specifically, you could make a wrapper now. But we could also wait until we do.
</blockquote></details>

Ok, I'll leave it is as is for now
___
*[`src/BloomExe/Collection/CollectionSettings.cs` line 500 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6270#-NpVi5Jg-MqvGAVjS6s5:-NpWDQBb9fwwKQI4vCVG:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/403c999f44b1b394c175c0b2751d9a801b4e6072/src/BloomExe/Collection/CollectionSettings.cs#L500)):*

> outerError

Rather than outer, isn't this inner? Or maybe "original" is better? (That matches the constructor signature for FileException.)
</blockquote></details>

Done.
___
*[`src/BloomExe/CollectionTab/CollectionTabView.cs` line 288 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6270#-NpVOjrx82ymjFrMNCae:-NpWDv_87gy2i2pfuZvf:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/403c999f44b1b394c175c0b2751d9a801b4e6072/src/BloomExe/CollectionTab/CollectionTabView.cs#L288)):*

Assuming that it actually does, how about `CheckForDuplicatesAndRepair`? Or similar.

Same for BookInfo's version.
</blockquote></details>

Done.
___
*[`src/BloomExe/ErrorReporter/HtmlErrorReporter.cs` line 234 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6270#-NpVif9zDSrY8rUeSSad:-NpWGVMk2uaCXxZXrgF1:bro5dqi) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/403c999f44b1b394c175c0b2751d9a801b4e6072/src/BloomExe/ErrorReporter/HtmlErrorReporter.cs#L234)):*

```
                    if (!wasAlreadyLocked)
                    {
                        System.Threading.Monitor.Exit(_lock);
                    }

```
Unless I'm missing something, this isn't needed. The finally block will do it.
</blockquote></details>

You're right, looks like I shouldn't need it. But I put it in in response to the lock not getting released somewhere. So hopefully I did eventually fix that problem... 
___
*[`src/BloomExe/Utils/FileException.cs` line 24 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6270#-NpVjyjx3opgCy13_7gj:-NpWH4iw4e8SLItyDZ0x:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/403c999f44b1b394c175c0b2751d9a801b4e6072/src/BloomExe/Utils/FileException.cs#L24)):*

Should this be GetFilePathIfPresent?
</blockquote></details>

Done.
___
*[`src/BloomExe/Utils/FileException.cs` line 30 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6270#-NpVjG0x6EtaEtje6BSj:-NpWJuaN39Xtu4DaBkWG:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/403c999f44b1b394c175c0b2751d9a801b4e6072/src/BloomExe/Utils/FileException.cs#L30)):*

Is there a reason this is empty string rather than null?
</blockquote></details>

Done.
___
*[`src/BloomExe/Utils/OneDriveUtils.cs` line 30 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6270#-NpWwx-a1iPOX0qmM25b:-Np_CTL6CQIjo4-IQ6YZ:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/76b64926c1bded2d6f4f70f154303c6127176ea9/src/BloomExe/Utils/OneDriveUtils.cs#L30)):*

Remember, unlike js/ts, in C#, we capitalize method names.
</blockquote></details>

Done.
___
*[`src/BloomExe/web/controllers/ProblemReportApi.cs` line 862 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6270#-NpQxoypAhMbjQihPctQ:-NpWK3i2DwG4_ve6CjXF:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/403c999f44b1b394c175c0b2751d9a801b4e6072/src/BloomExe/web/controllers/ProblemReportApi.cs#L862)):*

Yeah, always a hard call.
Doesn't seem worth it at this time.
Maybe worth replacing the TODO comment with something like
// We decided not to localize this for now.
</blockquote></details>

Done.

nabalone (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6270)* status: 12 of 18 files reviewed, 6 unresolved discussions (waiting on @andrew-polk)

___
*[`src/BloomExe/ErrorReporter/HtmlErrorReporter.cs` line 231 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6270#-NpWxJaZ5LaTdSlF-rwS:-NpeQYwJ3SLK23MrLXxP:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/76b64926c1bded2d6f4f70f154303c6127176ea9/src/BloomExe/ErrorReporter/HtmlErrorReporter.cs#L231)):*

// TODO

I know you're still working on this. I'm just adding this to make sure it doesn't get lost or left in.
</blockquote></details>

Done.
___
*[`src/BloomExe/Utils/OneDriveUtils.cs` line 40 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6270#-NpWx5SA36XTziH7FeV9:-NpeQUn_BbYGwgxnBg7x:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/76b64926c1bded2d6f4f70f154303c6127176ea9/src/BloomExe/Utils/OneDriveUtils.cs#L40)):*

// TODO

I know you're still working on this. I'm just adding this to make sure it doesn't get lost or left in.
</blockquote></details>

Done.
___
*[`src/BloomExe/web/controllers/ProblemReportApi.cs` line 629 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6270#-NpeQjFS8q3Y2zWq9rZm:-NpeQjFS8q3Y2zWq9rZn:b46nsz6) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/a95f1af02dd07a01d7b7605d949f71c5ad40e54f/src/BloomExe/web/controllers/ProblemReportApi.cs#L629)):*
> ```Smalltalk
>         // ENHANCE: Reduce duplication in HtmlErrorReporter and ProblemReportApi code. Some of the ProblemReportApi code can move to HtmlErrorReporter code.
> 
>         // ENHANCE: I think levelOfProblem would benefit from being required and being an enum.
> ```

I see we have ProblemLevel.cs (https://github.com/BloomBooks/BloomDesktop/blob/master/src/BloomExe/ErrorReporter/ProblemLevel.cs)
It feels bad to still use string literals here, in the defaults for ShowProblemDialog and CheckForAndHandleOneDriveExceptions below. Is there an easy way to improve that now?

nabalone (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6270)* status: 12 of 18 files reviewed, 5 unresolved discussions (waiting on @andrew-polk)

___
*[`src/BloomExe/web/controllers/ProblemReportApi.cs` line 629 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6270#-NpeQjFS8q3Y2zWq9rZm:-Npf2Wei7v-MHx3pCKhS:b-fa9hn1) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/a95f1af02dd07a01d7b7605d949f71c5ad40e54f/src/BloomExe/web/controllers/ProblemReportApi.cs#L629)):*

I see we have ProblemLevel.cs (https://github.com/BloomBooks/BloomDesktop/blob/master/src/BloomExe/ErrorReporter/ProblemLevel.cs)
It feels bad to still use string literals here, in the defaults for ShowProblemDialog and CheckForAndHandleOneDriveExceptions below. Is there an easy way to improve that now?
</blockquote></details>

I guess I'm not sure whether to problemLevel to an enum now and if so how far up/down the callstacks to make that change

andrew-polk (changes requested):
> Reviewed 6 of 6 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6270)* status: all files reviewed, 4 unresolved discussions (waiting on @nabalone)

___
*[`DistFiles/localization/en/Bloom.xlf` line 5140 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6270#-NpVNbqK7A6X8YY9gUuo:-Npu4Dla8qScta1qfKOd:b8ol185) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/403c999f44b1b394c175c0b2751d9a801b4e6072/DistFiles/localization/en/Bloom.xlf#L5140)):*

Oh ok, Done. I was just trying to keep them together with other problem dialog text.
</blockquote></details>

Ah, yeah, I guess that would have made sense, too.
But putting them in low seems right for these. Thanks.
___
*[`src/BloomBrowserUI/problemDialog/ProblemDialog.tsx` line 20 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6270#-NpWxOcN1R-4ESx5R1yu:-Npu7k0T9m-cSVWd48Ue:b-dn0qrc) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/76b64926c1bded2d6f4f70f154303c6127176ea9/src/BloomBrowserUI/problemDialog/ProblemDialog.tsx#L20)):*

Ah ok, I tried to clarify
</blockquote></details>

Sigh. I see now why you introducing a new ProblemKind(s). We had conflated the idea of which dialog to present and which theme to use. It worked previously because there was no overlap but now there is.
It seems like the ideal would be to separate them. So I'm trying to figure out how worth it it is....
One option is to add a NotifyFatal ProblemKind, but that just perpetuates the problem.
Another option is to make two separate props, one for which problem kind, one for which theme. 
A third option is to have a theme override such that you would pass Notify as the kind in this case but pass in a theme override of Fatal. I'm leaning toward option #3. What do you think? Happy to discuss in person...


As a somewhat separate concern... if we have all these props which are actually specific to the NotifyDialog, it seems like maybe we should define INotifyDialogProps in NotifyDialog.tsx and make the type here `level:ProblemKind | INotifyDialogProps` or whatever the syntax is. I can pair with you on this if needed.
___
*[`src/BloomExe/Program.cs` line 1301 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6270#-NpVjqfb7meRYd9sCU9u:-Npu3tIQ8nib0FFH3kz8:b-a6v4ms) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/403c999f44b1b394c175c0b2751d9a801b4e6072/src/BloomExe/Program.cs#L1301)):*

Done. But do we eventually want to add filePaths to the error message or log now that we might have them here
</blockquote></details>

Yes, adding it to the log sounds right. I don't think we want to add it to the error message which already has the path to the collection.
___
*[`src/BloomExe/web/controllers/ProblemReportApi.cs` line 633 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6270#-NpuAYuk8S6b0Qxi9ZG5:-NpuAYul2b8FBpHkwm2z:b-27l0r9) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/a95f1af02dd07a01d7b7605d949f71c5ad40e54f/src/BloomExe/web/controllers/ProblemReportApi.cs#L633)):*
> Show Rpoblem Dialog

Show Problem Dialog
___
*[`src/BloomExe/web/controllers/ProblemReportApi.cs` line 641 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6270#-NpuAjWi5cMPS6_lXACU:-NpuAjWi5cMPS6_lXACV:b-lk48sr) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/a95f1af02dd07a01d7b7605d949f71c5ad40e54f/src/BloomExe/web/controllers/ProblemReportApi.cs#L641)):*
> ```Smalltalk
>         /// <param name="controlForScreenshotting"></param>
>         /// <param name="exception"></param>
>         /// <param name="height"></param>
> ```

Let's remove all these empty param tags.
___
*[`src/BloomExe/web/controllers/ProblemReportApi.cs` line 867 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6270#-NpuIyLc3tEaavaIeydY:-NpuIyLc3tEaavaIeydZ:bi7p1yj) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/a95f1af02dd07a01d7b7605d949f71c5ad40e54f/src/BloomExe/web/controllers/ProblemReportApi.cs#L867)):*
> ```Smalltalk
>         }
> 
>         private static void ShowOneDriveExceptionFallbackDialog(
> ```

Should we move this and the next method into OneDriveUtils?

nabalone:
*[`src/BloomBrowserUI/problemDialog/ProblemDialog.tsx` line 20 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6270#-NpWxOcN1R-4ESx5R1yu:-NpuZpix4FCQOqDyraxo:bw2a6yc) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/76b64926c1bded2d6f4f70f154303c6127176ea9/src/BloomBrowserUI/problemDialog/ProblemDialog.tsx#L20)):*

Sigh. I see now why you introducing a new ProblemKind(s). We had conflated the idea of which dialog to present and which theme to use. It worked previously because there was no overlap but now there is.
It seems like the ideal would be to separate them. So I'm trying to figure out how worth it it is....
One option is to add a NotifyFatal ProblemKind, but that just perpetuates the problem.
Another option is to make two separate props, one for which problem kind, one for which theme. 
A third option is to have a theme override such that you would pass Notify as the kind in this case but pass in a theme override of Fatal. I'm leaning toward option #3. What do you think? Happy to discuss in person...


As a somewhat separate concern... if we have all these props which are actually specific to the NotifyDialog, it seems like maybe we should define INotifyDialogProps in NotifyDialog.tsx and make the type here `level:ProblemKind | INotifyDialogProps` or whatever the syntax is. I can pair with you on this if needed.
</blockquote></details>

Do you mean that the level or the notify dialog props should be a single argument? As in, if the level is "notify", then instead of the string "notify", the "level" should just be an INotifyDialogPropsObject? Is this better than making a second, optional argument?

andrew-polk:
*[`src/BloomBrowserUI/problemDialog/ProblemDialog.tsx` line 20 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6270#-NpWxOcN1R-4ESx5R1yu:-Nq3UvXkBOvHUFrt72NF:b-s3kkou) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/76b64926c1bded2d6f4f70f154303c6127176ea9/src/BloomBrowserUI/problemDialog/ProblemDialog.tsx#L20)):*

Do you mean that the level or the notify dialog props should be a single argument? As in, if the level is "notify", then instead of the string "notify", the "level" should just be an INotifyDialogPropsObject? Is this better than making a second, optional argument?
</blockquote></details>

Sorry, I confused you because I didn't use the right syntax.
I'm suggesting that the type of the props for ProblemDialog be the union of `level:ProblemKind` and `INotifyDialogProps`.

andrew-polk:
*[`src/BloomBrowserUI/problemDialog/ProblemDialog.tsx` line 20 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6270#-NpWxOcN1R-4ESx5R1yu:-Nq3XGIu6wRuzqMpSXvO:b9jcj8z) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/76b64926c1bded2d6f4f70f154303c6127176ea9/src/BloomBrowserUI/problemDialog/ProblemDialog.tsx#L20)):*

Sorry, I confused you because I didn't use the right syntax.
I'm suggesting that the type of the props for ProblemDialog be the union of `level:ProblemKind` and `INotifyDialogProps`.
</blockquote></details>

I played with it a little just now in code. I would require an intersection (&) to do what I'm talking about.
But I'm second-guessing that it is worth it. I was just trying to document via type that the other props are really just for the notify dialog. But it doesn't feel clean as I try to make it work.

nabalone:
*[`src/BloomBrowserUI/problemDialog/ProblemDialog.tsx` line 20 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6270#-NpWxOcN1R-4ESx5R1yu:-Nq3brIt-T5rBxokW2Eh:b9jbvce) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/76b64926c1bded2d6f4f70f154303c6127176ea9/src/BloomBrowserUI/problemDialog/ProblemDialog.tsx#L20)):*

I played with it a little just now in code. I would require an intersection (&) to do what I'm talking about.
But I'm second-guessing that it is worth it. I was just trying to document via type that the other props are really just for the notify dialog. But it doesn't feel clean as I try to make it work.
</blockquote></details>

Oh. Do you not like this?
```
export const ProblemDialog: React.FC<{level:ProblemKind} & Partial<INotifyDialogProps>> = props => {

    if (props.level === ProblemKind.Notify) {

        return <NotifyDialog {...props} />;

    } else {

        return <ReportDialog kind\={props.level} />;

    }

};
```

andrew-polk (changes requested):
> Reviewed 3 of 4 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6270)* status: 17 of 18 files reviewed, 2 unresolved discussions (waiting on @nabalone)

___
*[`src/BloomBrowserUI/problemDialog/ProblemDialog.tsx` line 20 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6270#-NpWxOcN1R-4ESx5R1yu:-Nq3e2F3AEc6bdpwfkYB:b-v640t8) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/76b64926c1bded2d6f4f70f154303c6127176ea9/src/BloomBrowserUI/problemDialog/ProblemDialog.tsx#L20)):*

Oh. Do you not like this?
```
export const ProblemDialog: React.FC<{level:ProblemKind} & Partial<INotifyDialogProps>> = props => {

    if (props.level === ProblemKind.Notify) {

        return <NotifyDialog {...props} />;

    } else {

        return <ReportDialog kind\={props.level} />;

    }

};
```
</blockquote></details>

That works!
You did a better job of figuring it out.
___
*[`src/BloomExe/Program.cs` line 1301 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6270#-NpVjqfb7meRYd9sCU9u:-Nq3dZwc4bfTbhhWIFZi:b-o2plu2) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/403c999f44b1b394c175c0b2751d9a801b4e6072/src/BloomExe/Program.cs#L1301)):*

Yes, adding it to the log sounds right. I don't think we want to add it to the error message which already has the path to the collection.
</blockquote></details>

Sorry, I meant to suggest we go ahead and log it now.

nabalone (commented):
I am testing throwing exceptions and catching them and I am getting a "Invoke or BeginInvoke cannot be called on a control until the window handle has been created." error while trying to display the error dialog. I am worried that this looks like a timing thing and that users could reach it while encountering OneDrive errors. I have not yet figured out how to fix it

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6270#-:-Nq4bySpAi3NOQJuDF6D:b-wb8ugz)* status: 15 of 18 files reviewed, 2 unresolved discussions (waiting on @andrew-polk)

___
*[`src/BloomExe/Program.cs` line 1301 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6270#-NpVjqfb7meRYd9sCU9u:-Nq3wX1R-FETE7rnRxA9:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/403c999f44b1b394c175c0b2751d9a801b4e6072/src/BloomExe/Program.cs#L1301)):*

Sorry, I meant to suggest we go ahead and log it now.
</blockquote></details>

Done.
___
*[`src/BloomExe/web/controllers/ProblemReportApi.cs` line 633 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6270#-NpuAYuk8S6b0Qxi9ZG5:-Npug2QYE5ybflEtUYoc:bn9cco2) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/a95f1af02dd07a01d7b7605d949f71c5ad40e54f/src/BloomExe/web/controllers/ProblemReportApi.cs#L633)):*

> Show Rpoblem Dialog

Show Problem Dialog
</blockquote></details>

Done. (will push )
___
*[`src/BloomExe/web/controllers/ProblemReportApi.cs` line 641 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6270#-NpuAjWi5cMPS6_lXACU:-Npug6xy4fvvu7yl_bcN:b-t1br4g) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/a95f1af02dd07a01d7b7605d949f71c5ad40e54f/src/BloomExe/web/controllers/ProblemReportApi.cs#L641)):*

Let's remove all these empty param tags.
</blockquote></details>

Done. (will push)
___
*[`src/BloomExe/web/controllers/ProblemReportApi.cs` line 867 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6270#-NpuIyLc3tEaavaIeydY:-NpugaZ18j7iHx_bEQFj:bsjx17d) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/a95f1af02dd07a01d7b7605d949f71c5ad40e54f/src/BloomExe/web/controllers/ProblemReportApi.cs#L867)):*

Should we move this and the next method into OneDriveUtils?
</blockquote></details>

I'm not sure it makes any sense outside the context of `CheckForAndHandleOneDriveExceptions`, and I feel like they should stay together since they have a lot of the same text. Can I put it inside `CheckForAndHandleOneDriveExceptions` instead? I also factored out the text they share

nabalone (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6270)* status: 15 of 18 files reviewed, 3 unresolved discussions (waiting on @andrew-polk)

___
*[`src/BloomBrowserUI/problemDialog/NotifyDialog.tsx` line 75 at r5](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6270#-Nq4cxaREafreTxRvFz1:-Nq4cxaS75-L2KaJh9lN:b-m4f507) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/c9f370f9dea6c2025392ffd507036f01c4ddd986/src/BloomBrowserUI/problemDialog/NotifyDialog.tsx#L75)):*
> ```XML
>                                     margin-bottom: 20px; 
>                                     a {
>                                         color: ${kBloomBlue}; // we are passing links in the text
> ```

This feels like a band aid. Wondering if it is worth it try and handle links and styling better

nabalone:
*[`src/BloomExe/web/controllers/ProblemReportApi.cs` line 657 at r6](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6270#-Nq4u0Hg9SDKX9f60j0R:-Nq4u0Hg9SDKX9f60j0S:b-3j3owx) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/91ef21bb0c1f46f7462625738c87a237d8d889bf/src/BloomExe/web/controllers/ProblemReportApi.cs#L657)):*
> ```Smalltalk
>                         // this one, while apparently redundant might be wise to keep since closing the splash screen
>                         // needs to be done on the UI thread.
>                         try
> ```

Here I catch the `Invoke or BeginInvoke cannot be called on a control until the window handle has been created.` exception that was happening, at least this should catch it in OneDrive  error cases

nabalone:
> I am testing throwing exceptions and catching them and I am getting a "Invoke or BeginInvoke cannot be called on a control until the window handle has been created." error while trying to display the error dialog. I am worried that this looks like a timing thing and that users could reach it while encountering OneDrive errors. I have not yet figured out how to fix it

Was happening on an attempt to close the splash screen. I added an additional try/catch to catch it.

andrew-polk (changes requested):
> Reviewed 1 of 3 files at r5, 1 of 2 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6270)* status: 17 of 18 files reviewed, 3 unresolved discussions (waiting on @nabalone)

___
*[`src/BloomExe/web/controllers/ProblemReportApi.cs` line 867 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6270#-NpuIyLc3tEaavaIeydY:-Nq4zGGBCVd4CgLRqgQy:b-xge3ej) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/a95f1af02dd07a01d7b7605d949f71c5ad40e54f/src/BloomExe/web/controllers/ProblemReportApi.cs#L867)):*

I'm not sure it makes any sense outside the context of `CheckForAndHandleOneDriveExceptions`, and I feel like they should stay together since they have a lot of the same text. Can I put it inside `CheckForAndHandleOneDriveExceptions` instead? I also factored out the text they share
</blockquote></details>

Sorry; I actually meant to move both `ShowOneDriveExceptionFallbackDialog` and `CheckForAndHandleOneDriveExceptions` to OneDriveUtils. Is that feasible?

andrew-polk (approved):
> Reviewed 5 of 5 files at r7, 1 of 2 files at r8, 1 of 1 files at r9, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6270)* status: all files reviewed, 2 unresolved discussions

andrew-polk (approved):
> Dismissed @nabalone from 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6270)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @nabalone)

PR #6281, opened 2024-02-01
Tweak XMatter layout for Device16x9Landscape and Chetana (BL-13029)

StephenMcConnel:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6281)

### DIFF
--- a/src/content/branding/Chetana-Trust/branding.less
+++ b/src/content/branding/Chetana-Trust/branding.less
@@ -13,6 +13,16 @@
         height: 110px !important;
     }
 }
+// Device16x9Landscape has limited vertical space, so we specify a smaller minimum height
+// for the branding bottom html, and we reduce the size of the logo image.  See BL-13029.
+// The new min-height is very close to the height generated when the min-height is unset.
+.Device16x9Landscape.titlePage [data-book*="title-page-branding-bottom-html"] {
+    min-height: 133px;
+    flex: unset !important;
+    img {
+        height: 100px !important;
+    }
+}
 
 [data-book*="outside-back-cover-branding-bottom-html"] {
     font-weight: bold;
@@ -29,4 +39,19 @@
 .two-em-shrink {
     height: 2em;
     flex-shrink: 10;
+}
+
+// Device16x9Landscape has limited vertical space so we reduce the minimum
+// height of the originalContributions and funding fields.  See BL-13029.
+.Device16x9Landscape.titlePage {
+    #originalContributions {
+        .bloom-editable {
+            min-height: 1.5em; // one line
+        }
+    }
+    #funding {
+        .bloom-editable {
+            min-height: 1.5em; // one line
+        }
+    }
 }


StephenMcConnel:
We may want to also handle A6Portrait and QuarterLetterPortrait since they have the same problem.

JohnThomson (approved):
If this is just for Chetana I would be inclined not to do the work unless we have reason to think they use those page sizes.
I'm approving this, but nor merging it yet since there are a couple of minor points you might want to clarify.

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6281#-:-NpeGHnbB157YPwyzqCs:b-4x1tpj)* status: all files reviewed, 2 unresolved discussions (waiting on @StephenMcConnel)

___
*[`src/content/branding/Chetana-Trust/branding.less` line 17 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6281#-NpeEz5V0YO5q_cviUSU:-NpeEz5V0YO5q_cviUSV:bvznz3m) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/b5b7d04b5d43be808a72f49dbe10c408dfa395e1/src/content/branding/Chetana-Trust/branding.less#L17)):*
> ```Less
> }
> // Device16x9Landscape has limited vertical space, so we don't specify a
> // minimum height for the brinding bottom html, and we slightly reduce the
> ```

brinding => branding
___
*[`src/content/branding/Chetana-Trust/branding.less` line 45 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6281#-NpeFDUpByCdb97rltzt:-NpeFDUpByCdb97rltzu:b4i7eub) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/b5b7d04b5d43be808a72f49dbe10c408dfa395e1/src/content/branding/Chetana-Trust/branding.less#L45)):*
> ```Less
> // Device16x9Landscape has limited vertical space, so we reduce the space between
> // the "an ARM Library Book" and the image.  See BL-13029.
> .Device16x9Landscape.titlePage .two-em-shrink {
> ```

This is a strange name for a class, and even stranger now that it is no longer always 2em. But it doesn't seem to be new.
Since we're already strongly encouraging this space to shrink if we are short of space, I'm not clear why we want to reduce the default space (I'm assuming that on larger pages it really starts at 2em.)

StephenMcConnel (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6281)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @JohnThomson)

___
*[`src/content/branding/Chetana-Trust/branding.less` line 17 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6281#-NpeEz5V0YO5q_cviUSU:-Npu34u6CbLgAQGmttY8:bhgjd6d) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/b5b7d04b5d43be808a72f49dbe10c408dfa395e1/src/content/branding/Chetana-Trust/branding.less#L17)):*

brinding => branding
</blockquote></details>

Done.  I also changed the min-height from unset to 133px, which closely matches what the height was when min-height was unset.  This simplifies the overall change by not requiring a new rule for Device16x9Landscape and two-em-shrink, and preserves the semantics of the two-em-shrink name.  (which admittedly is not the best name, but the one already in use)
___
*[`src/content/branding/Chetana-Trust/branding.less` line 45 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6281#-NpeFDUpByCdb97rltzt:-Npu2mEy9W3dgAnyzktG:bk096do) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/b5b7d04b5d43be808a72f49dbe10c408dfa395e1/src/content/branding/Chetana-Trust/branding.less#L45)):*

This is a strange name for a class, and even stranger now that it is no longer always 2em. But it doesn't seem to be new.
Since we're already strongly encouraging this space to shrink if we are short of space, I'm not clear why we want to reduce the default space (I'm assuming that on larger pages it really starts at 2em.)
</blockquote></details>

I've changed the fix to not need a separate rule here.

JohnThomson (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6281)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @StephenMcConnel)

PR #6283, opened 2024-02-02
Make style name change to all editables in group at once (BL-13031)

JohnThomson:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6283)

### DIFF
--- a/src/BloomBrowserUI/bookEdit/StyleEditor/StyleEditor.ts
+++ b/src/BloomBrowserUI/bookEdit/StyleEditor/StyleEditor.ts
@@ -200,10 +200,28 @@ export default class StyleEditor {
         return styleName.substr(0, suffixIndex);
     }
 
+    // Make the specified style (which should include the trailing '-style') the style for the target element,
+    // and if it's in a translation group, for all the siblings that are bloom-editable.
     private static SetStyleNameForElement(
         target: HTMLElement,
         newStyle: string
     ) {
+        this.SetStyleNameForLeaf(target, newStyle);
+        // If this is a translation group, we need to set the style on all the siblings
+        // that are bloom-editable
+        const group = target.closest(".bloom-translationGroup");
+        if (!group) return;
+        // This will typically include the target itself, but it's not very costly to do
+        // it twice to that one element.
+        const siblings = Array.from(
+            group.getElementsByClassName("bloom-editable")
+        );
+        siblings.forEach(sibling =>
+            this.SetStyleNameForLeaf(sibling as HTMLElement, newStyle)
+        );
+    }
+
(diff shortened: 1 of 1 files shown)


hatton (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6283)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @JohnThomson)

___
*[`src/BloomBrowserUI/bookEdit/StyleEditor/StyleEditor.ts` line 209 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6283#-NpfZKaW9Smath-gUJ5L:-NpfZKaXBL5Ibnq7mnSc:b-7q7vfs) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/038bde62f477ef75c0d438ffa4b8c3772d6467b3/src/BloomBrowserUI/bookEdit/StyleEditor/StyleEditor.ts#L209)):*
> ```XML
>         newStyle: string
>     ) {
>         this.SetStyleNameForLeaf(target, newStyle);
> ```

Would it be clearer to remove the `filter()` and then move this into the `if(!group) return` ?

JohnThomson (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6283)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @hatton)

___
*[`src/BloomBrowserUI/bookEdit/StyleEditor/StyleEditor.ts` line 209 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6283#-NpfZKaW9Smath-gUJ5L:-Npg3r1M24ZjLOft6dha:bkp0jlk) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/038bde62f477ef75c0d438ffa4b8c3772d6467b3/src/BloomBrowserUI/bookEdit/StyleEditor/StyleEditor.ts#L209)):*

Would it be clearer to remove the `filter()` and then move this into the `if(!group) return` ?
</blockquote></details>

It gets complicated. I'm not 100% sure that there isn't some old case involving text-areas where the target might NOT be bloom-editable, or where the target is the group itself.
I decided to simplify by just leaving out the filter. Then it's absolutely definite that it gets done to the original target, and doing it twice doesn't matter, so we don't really need the complexity of filtering.

hatton (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6283)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @JohnThomson)

PR #6294, opened 2024-02-14
Fix various problems with downloading for edit (BL-13036)

JohnThomson:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6294)

### DIFF
--- a/DistFiles/localization/en/Bloom.xlf
+++ b/DistFiles/localization/en/Bloom.xlf
@@ -1170,6 +1170,11 @@
         <source xml:lang="en">Downloading book</source>
         <note>ID: Download.DownloadingDialogTitle</note>
       </trans-unit>
+      <trans-unit id="Download.FromBloomLibrary" translate="no">
+        <source xml:lang="en">From Bloom Library</source>
+        <note>ID: Download.FromBloomLibrary</note>
+        <note>Will be followed by hyphen, space, and (possibly partial) book title and used as a folder name for a collection containing that one book. Don't include illegal path characters like colon.</note>
+      </trans-unit>
       <trans-unit id="Download.GenericNetworkProblemNotice">
         <source xml:lang="en">There was a problem downloading the book.  You can try again at a different time, or write to us at issues@bloomlibrary.org if you cannot get the download to work from your location.</source>
         <note>ID: Download.GenericNetworkProblemNotice</note>
--- a/src/BloomExe/Program.cs
+++ b/src/BloomExe/Program.cs
@@ -98,6 +98,8 @@ namespace Bloom
 
         public static SynchronizationContext MainContext { get; private set; }
 
+        public static bool RunningSecondInstance { get; private set; }
+
         [STAThread]
         [HandleProcessCorruptedStateExceptions]
         static int Main(string[] args1)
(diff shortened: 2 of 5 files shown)


andrew-polk (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6294)* status: all files reviewed, 2 unresolved discussions (waiting on @JohnThomson)

___
*[`DistFiles/localization/en/Bloom.xlf` line 1176 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6294#-Nqck7CN8GxhQm7oiHyw:-Nqck7CN8GxhQm7oiHyx:bp0pgel) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/a6df4430178162c880cfa3989ddc74a72834fc45/DistFiles/localization/en/Bloom.xlf#L1176)):*
> Don't include illegal path characters like colon.

I'm not sure we want to rely on translators for this safety. Does the code try to handle problems?
___
*[`src/BloomExe/Program.cs` line 558 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6294#-Nqcl2fXErJdnfhoLsKm:-Nqcl2fXErJdnfhoLsKn:b-9m7now) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/a6df4430178162c880cfa3989ddc74a72834fc45/src/BloomExe/Program.cs#L558)):*
> ```Smalltalk
>                                             )
>                                     );
>                                     // go forwards as if we weren't given an explicit collection to open (except we're showing a report about what happened)
> ```

We lost the `path = "";` but still have the comment which seems to go with it.
Were we supposed to keep `path = "";`?

JohnThomson (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6294)* status: 3 of 5 files reviewed, 2 unresolved discussions (waiting on @andrew-polk)

___
*[`DistFiles/localization/en/Bloom.xlf` line 1176 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6294#-Nqck7CN8GxhQm7oiHyw:-NqcwExY1F3-tg5vhhYR:b6kxtjh) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/a6df4430178162c880cfa3989ddc74a72834fc45/DistFiles/localization/en/Bloom.xlf#L1176)):*

> Don't include illegal path characters like colon.

I'm not sure we want to rely on translators for this safety. Does the code try to handle problems?
</blockquote></details>

I now made it do so.
___
*[`src/BloomExe/Program.cs` line 558 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6294#-Nqcl2fXErJdnfhoLsKm:-Nqcu-7y66WdsJerNOvg:bei0o45) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/a6df4430178162c880cfa3989ddc74a72834fc45/src/BloomExe/Program.cs#L558)):*

We lost the `path = "";` but still have the comment which seems to go with it.
Were we supposed to keep `path = "";`?
</blockquote></details>

In the old version, the following line checked for path ending in .bloomCollection and if so tried to open it. We cleared it to prevent that.
Now, we already did that, and the path variable is never used again. We "go forward as if we weren't given a collection" in that we don't add anything to MruProjects. So I think the comment is still useful. I've hopefully clarified it.

andrew-polk (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6294)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @JohnThomson)

PR #6297, opened 2024-02-14
Implement the new API design

andrew-polk:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6297)

### DIFF
--- a/src/BloomExe/Book/BookInfo.cs
+++ b/src/BloomExe/Book/BookInfo.cs
@@ -1115,13 +1115,13 @@ namespace Bloom.Book
         }
 
         /// <summary>
-        /// Get the reduced Json string that we upload to set the database entry for the book on our website.
+        /// Get the reduced Json that we upload to set the database entry for the book on our website.
         /// This leaves out some of the metadata that we use while working on the book.
         /// Note that the full metadata is currently uploaded to S3 as part of the book content;
         /// this reduced subset is just for the website itself.
         /// Note that if you add a property to the upload set here, you must manually add a corresponding field to
-        /// the Book table in Parse.com. This is very important. Currently, the field will auto-add to
-        /// the Parse databases used for unit testing and even (I think) the one for sandbox testing,
+        /// the books table in the database. This is very important. Currently, the field will auto-add to
+        /// the parse-server databases used for unit testing and even (I think) the one for sandbox testing,
         /// but not to the live site; so if you forget to do this uploading will suddenly break.
         /// It is for this reason that we deliberately don't automatically add new fields to the upload set.
         /// Note that it is desirable that the name you give each property in the anonymous object which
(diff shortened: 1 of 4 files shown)


JohnThomson (approved):
Looks good, though I'm not sure I understand all the implications. I asked a couple of minor questions.

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6297#-:-Nqi_j4N8ZEpc-eMeGVy:b-kjqxsk)* status: all files reviewed, 3 unresolved discussions (waiting on @andrew-polk)

___
*[`src/BloomExe/Book/BookInfo.cs` line 1132 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6297#-NqiYZJR4xSBVaz8_TRL:-NqiYZJR4xSBVaz8_TRM:b-t4t9ls) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/5a69b2a261046bf99e8af164f383b51cb467f6e5/src/BloomExe/Book/BookInfo.cs#L1132)):*
> ```Smalltalk
>         /// some fields left out. At least one unit test will fail if the names don't match.
>         /// (Though, I don't think anything besides that test currently attempts to create
>         /// a BookMetaData object from a WebDataJson string.)
> ```

Technically it's no longer a string. I suppose 
___
*[`src/BloomExe/WebLibraryIntegration/BloomLibraryBookApiClient.cs` line 177 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6297#-NqhNuvL8_R89Wm33b48:-NqhNuvM0ercGCHYSA1P:b-c4s05i) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/5a69b2a261046bf99e8af164f383b51cb467f6e5/src/BloomExe/WebLibraryIntegration/BloomLibraryBookApiClient.cs#L177)):*
> ```Smalltalk
>                 { "files", JsonConvert.SerializeObject(bookFiles) },
>                 { "clientName", $"Bloom {ApplicationUpdateSupport.ChannelName}" },
>                 { "clientVersion", Application.ProductVersion }
> ```

Should there also be an API version in here somewhere?
___
*[`src/BloomExe/WebLibraryIntegration/BloomLibraryBookApiClient.cs` line 230 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6297#-Nqi_PQYBYVZEyb7Dsjv:-Nqi_PQYBYVZEyb7Dsjw:bkqyzcs) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/5a69b2a261046bf99e8af164f383b51cb467f6e5/src/BloomExe/WebLibraryIntegration/BloomLibraryBookApiClient.cs#L230)):*
> ```Smalltalk
>             var request = MakePostRequest($"{bookId}:upload-finish");
> 
>             request.AddJsonBody(new { transactionId, metadata });
> ```

You've probably thought about this, but do we lose any security passing an object rather than a string for the server to parse?

andrew-polk (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6297)* status: 2 of 4 files reviewed, 3 unresolved discussions (waiting on @JohnThomson)

___
*[`src/BloomExe/Book/BookInfo.cs` line 1132 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6297#-NqiYZJR4xSBVaz8_TRL:-NqmlkQVESihBT3NIDe-:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/5a69b2a261046bf99e8af164f383b51cb467f6e5/src/BloomExe/Book/BookInfo.cs#L1132)):*

Technically it's no longer a string. I suppose
</blockquote></details>

Done.
___
*[`src/BloomExe/WebLibraryIntegration/BloomLibraryBookApiClient.cs` line 177 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6297#-NqhNuvL8_R89Wm33b48:-NqmhOcq1XAmTC68idG8:bug7v9o) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/5a69b2a261046bf99e8af164f383b51cb467f6e5/src/BloomExe/WebLibraryIntegration/BloomLibraryBookApiClient.cs#L177)):*

Should there also be an API version in here somewhere?
</blockquote></details>

The api version is part of the url: api.bloomlibrary.org/v1/books/...
___
*[`src/BloomExe/WebLibraryIntegration/BloomLibraryBookApiClient.cs` line 230 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6297#-Nqi_PQYBYVZEyb7Dsjv:-Nqmgvxu8FvlMnmRDkfd:b-uw9i6s) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/5a69b2a261046bf99e8af164f383b51cb467f6e5/src/BloomExe/WebLibraryIntegration/BloomLibraryBookApiClient.cs#L230)):*

You've probably thought about this, but do we lose any security passing an object rather than a string for the server to parse?
</blockquote></details>

Say more. Why would a string be more secure?
Are you just thinking that the parsing process itself would catch validation errors? In reality, we weren't doing anything which would have validated the string. We were just converting it directly to an object. There are one or two fields we do validate, but the general idea is we said you're allowed to break your own book.

JohnThomson (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6297)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @andrew-polk)

___
*[`src/BloomExe/WebLibraryIntegration/BloomLibraryBookApiClient.cs` line 230 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6297#-Nqi_PQYBYVZEyb7Dsjv:-NqmxX9mDvvGm0halRl4:b-cpokp5) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/5a69b2a261046bf99e8af164f383b51cb467f6e5/src/BloomExe/WebLibraryIntegration/BloomLibraryBookApiClient.cs#L230)):*

Say more. Why would a string be more secure?
Are you just thinking that the parsing process itself would catch validation errors? In reality, we weren't doing anything which would have validated the string. We were just converting it directly to an object. There are one or two fields we do validate, but the general idea is we said you're allowed to break your own book.
</blockquote></details>

I didn't have any clear idea in mind. It struck me as vaguely similar to how it's pretty safe for user-supplied content to be set as the innerText of some element, but allowing them to provide something that will be set as innerHTML is more dangerous...they can insert anything, even embedded javascript that takes over the whole site.
In this case, the object has to pass across the network, so it's no doubt getting serialized and deserialized anyway, and there would be no way to have code attached to anything...I think this is probably an unnecessary concern.

PR #6305, opened 2024-02-16
Fix small discrepancy in 5.7 aspect ratio snap (BL-13073)

JohnThomson:
Also cleans up edit mode borders

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6305)

### DIFF
--- a/src/BloomBrowserUI/bookEdit/css/editMode.less
+++ b/src/BloomBrowserUI/bookEdit/css/editMode.less
@@ -534,7 +534,7 @@ button.moveButton {
 textarea,
 div.bloom-editable,
 div.pageLabel[contenteditable="true"] {
-    border: thin solid @MediumGray;
+    outline: thin solid @MediumGray;
     /*[disabled]min-height:34px;*/
 }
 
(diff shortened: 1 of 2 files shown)


andrew-polk (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6305)* status: all files reviewed, 1 unresolved discussion (waiting on @JohnThomson)

___
*[`src/BloomBrowserUI/lib/split-pane/split-pane.js` line 782 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6305#-NqntvoW75I_P2zcPDN5:-NqntvoW75I_P2zcPDN6:b-3zvsps) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/cc8a1ab6b68c228ad4cbb4affa75813905abcbef/src/BloomBrowserUI/lib/split-pane/split-pane.js#L782)):*
> ```JavaScript
>             //     // 3px of margin on top pane in split that we need to leave room for
>             //     width += 3;
>             // }
> ```

Worth commenting why this is worth keeping as commented code and why we don't need it anymore?

PR #6310, opened 2024-02-19
Add "Show Page Numbers" to Book Settings dialog (BL-13091)

StephenMcConnel:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6310)

### DIFF
--- a/DistFiles/localization/en/BloomMediumPriority.xlf
+++ b/DistFiles/localization/en/BloomMediumPriority.xlf
@@ -409,6 +409,14 @@
         <source xml:lang="en">Show Topic</source>
         <note>BookSettings.ShowTopic</note>
       </trans-unit>
+      <trans-unit id="BookSettings.PageNumbers" sil:dynamic="true" translate="no">
+        <source xml:lang="en">Page Numbers</source>
+        <note>BookSettings.PageNumbers</note>
+      </trans-unit>
+      <trans-unit id="BookSettings.ShowPageNumbers" sil:dynamic="true" translate="no">
+        <source xml:lang="en">Show Page Numbers</source>
+        <note>BookSettings.ShowPageNumbers</note>
+      </trans-unit>
       <trans-unit id="BookSettings.FrontAndBackMatter" sil:dynamic="true" translate="no">
         <source xml:lang="en">Front &amp; Back Matter</source>
         <note>BookSettings.FrontAndBackMatter</note>
--- a/src/BloomBrowserUI/bookEdit/bookSettings/BookSettingsDialog.tsx
+++ b/src/BloomBrowserUI/bookEdit/bookSettings/BookSettingsDialog.tsx
@@ -156,6 +156,14 @@ export const BookSettingsDialog: React.FunctionComponent<{}> = () => {
         "Front & Back Matter",
         "BookSettings.FrontAndBackMatter"
     );
+    const pageNumbersLabel = useL10n(
+        "Page Numbers",
+        "BookSettings.PageNumbers"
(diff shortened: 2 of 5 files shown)


JohnThomson (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6310)* status: all files reviewed, 1 unresolved discussion (waiting on @StephenMcConnel)

___
*[`src/content/appearanceMigrations/efl-ebook-1/appearance.json` line 9 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6310#-Nr2Fa-E1FlrMgEkzU7k:-Nr2Fa-E1FlrMgEkzU7l:bl25m49) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/6c8cee7f3716c85a7488afbbc943ed2d6b1e6740/src/content/appearanceMigrations/efl-ebook-1/appearance.json#L9)):*
> ```JSON
> {
>   "cssThemeName": "zero-margin-ebook",
>   "pageNumber-show": false,
> ```

I think it's misleading to have the theme appear to set it false. This will have no effect, because it will always be completely controlled by the value derived from appearance.json, even if there is no value set there and C# fills in the default (true).  It might be worth retaining with a comment like "We wish we could find a good way to do this as some sort of default"

StephenMcConnel (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6310)* status: all files reviewed, 1 unresolved discussion (waiting on @JohnThomson)

___
*[`src/content/appearanceMigrations/efl-ebook-1/appearance.json` line 9 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6310#-Nr2Fa-E1FlrMgEkzU7k:-Nr2NvNS6JDVaBP6wlRr:b1jsm9s) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/6c8cee7f3716c85a7488afbbc943ed2d6b1e6740/src/content/appearanceMigrations/efl-ebook-1/appearance.json#L9)):*

I think it's misleading to have the theme appear to set it false. This will have no effect, because it will always be completely controlled by the value derived from appearance.json, even if there is no value set there and C# fills in the default (true).  It might be worth retaining with a comment like "We wish we could find a good way to do this as some sort of default"
</blockquote></details>

This works fine in my testing.  It doesn't lock down the value as false, but it starts the value out as false when the book is migrated.  If the line is not there, then the value is initialized to true as called for in the C# code.

hatton (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6310)* status: all files reviewed, 1 unresolved discussion (waiting on @JohnThomson)

___
*[`src/content/appearanceMigrations/efl-ebook-1/appearance.json` line 9 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6310#-Nr2Fa-E1FlrMgEkzU7k:-Nr2OyIc4oNSrIS0o8UR:b-6rosbi) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/6c8cee7f3716c85a7488afbbc943ed2d6b1e6740/src/content/appearanceMigrations/efl-ebook-1/appearance.json#L9)):*

This works fine in my testing.  It doesn't lock down the value as false, but it starts the value out as false when the book is migrated.  If the line is not there, then the value is initialized to true as called for in the C# code.
</blockquote></details>

Yeah, this isn't a theme, this is a migration

PR #6314, opened 2024-02-20
Allow focus in Format dialog when decodable reader is active (BL-13043)

StephenMcConnel:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6314)

### DIFF
--- a/src/BloomBrowserUI/bookEdit/StyleEditor/StyleEditor.ts
+++ b/src/BloomBrowserUI/bookEdit/StyleEditor/StyleEditor.ts
@@ -2414,4 +2414,23 @@ export default class StyleEditor {
         };
         showSimpleColorPickerDialog(colorPickerDialogProps);
     }
+
+    public static isStyleDialogOpen(): boolean {
+        // The #format-toolbar element is instantiated on the page only when the
+        // dialog is showing.
+        const page = parent.window.document.getElementById("page");
+        const contentWindow = page
+            ? (<HTMLIFrameElement>page).contentWindow
+            : null;
+
+        // possibly unit-testing
+        if (!contentWindow) {
+            const formatDialog = $(".bloom-page").find("#format-toolbar");
+            return formatDialog.length > 0;
+        }
+        const formatDialog = $("body", contentWindow.document).find(
+            "#format-toolbar"
+        );
+        return formatDialog.length > 0;
+    }
 }
--- a/src/BloomBrowserUI/bookEdit/toolbox/readers/decodableReader/decodableReaderToolboxTool.ts
+++ b/src/BloomBrowserUI/bookEdit/toolbox/readers/decodableReader/decodableReaderToolboxTool.ts
@@ -14,6 +14,7 @@ import {
 import { ITool } from "../../toolbox";
 import theOneLocalizationManager from "../../../../lib/localizationManager/localizationManager";
 import { get } from "../../../../utils/bloomApi";
+import StyleEditor from "../../../StyleEditor/StyleEditor";
 
 export class DecodableReaderToolboxTool implements ITool {
     public makeRootElement(): HTMLDivElement {
@@ -74,7 +75,11 @@ export class DecodableReaderToolboxTool implements ITool {
                     event.relatedTarget.matches(".bloom-editable")
                 )
                     createCkEditorBookMarks = false;
-
+                // If the Format (Styles) dialog is showing, then we don't want to create
+                // bookmarks.  The div#format-toolbar is instantiated only when the dialog
+                // is showing.  See BL-13043.
+                else if (StyleEditor.isStyleDialogOpen())
+                    createCkEditorBookMarks = false;
                 getTheOneReaderToolsModel().doMarkup(createCkEditorBookMarks);
             });
 


src/BloomBrowserUI/bookEdit/toolbox/readers/readerToolsModel.ts
```diff
@@ -1055,10 +1055,16 @@ export class ReaderToolsModel {
                 .closest("body")
                 .children('.qtip:not(".uibloomSourceTextsBubble")')
                 .remove();
-
+            // If the Format (Styles) dialog is showing, then we don't want to create
+            // bookmarks.  The div#format-toolbar is instantiated only when the dialog
+            // is showing.  See BL-13043.
+            const formatDialog = $(editableElements[0])
+                .closest("body")
+                .find("#format-toolbar");
```
andrew-polk, line 1063: I'm slightly concerned about this fix because the user can actually type in the book while the style dialog is open.

What are the ramifications of not setting these bookmarks? If it is simply that the cursor jumps, we can probably just live with that.
I just didn't know if it might cause data problems (like introduction of unwanted characters, etc.).

StephenMcConnel (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6314)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @andrew-polk)

___
*[`src/BloomBrowserUI/bookEdit/toolbox/readers/readerToolsModel.ts` line 1063 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6314#gh-1497889553:-NrBcNJe3niblPaJrhry:bskyy60) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/0c88c0ce2f561f40281dc528b05f49b82f24d718/src/BloomBrowserUI/bookEdit/toolbox/readers/readerToolsModel.ts#L1063)):*

I'm slightly concerned about this fix because the user can actually type in the book while the style dialog is open.

What are the ramifications of not setting these bookmarks? If it is simply that the cursor jumps, we can probably just live with that.
I just didn't know if it might cause data problems (like introduction of unwanted characters, etc.).
</blockquote></details>

Clicking anywhere outside the Style Dialog closes it, so I think the design is that the user can't type in the book while the dialog is open.  The bug is that trying to type in the Style Dialog actually types in the book instead.
I've changed the fix to be less intrusive on the general operation of the doModal() function.  It now checks for the existence of the format dialog only in the focusout handler attached to the book's page element.

andrew-polk (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6314)* status: all files reviewed, 2 unresolved discussions (waiting on @StephenMcConnel)

___
*[`src/BloomBrowserUI/bookEdit/toolbox/readers/readerToolsModel.ts` line 968 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6314#-NrCMHjyBW7o9M4ilhsQ:-NrCMHjyBW7o9M4ilhsR:byasfw2) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/9b8dc06654061722e81af7642b5caf4549003bf0/src/BloomBrowserUI/bookEdit/toolbox/readers/readerToolsModel.ts#L968)):*
> ```XML
>         );
>         return formatDialog.length > 0;
>     }
> ```

How easy is it to move this to another file like the one for the style dialog?
Don't do any heroics, but it doesn't seem to belong in readerToolsModel.
___
*[`src/BloomBrowserUI/bookEdit/toolbox/readers/decodableReader/decodableReaderToolboxTool.ts` line 84 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6314#-NrCMA-T4uAingCdVo0G:-NrCMA-U90F2R_M1G7BA:b-eize4u) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/9b8dc06654061722e81af7642b5caf4549003bf0/src/BloomBrowserUI/bookEdit/toolbox/readers/decodableReader/decodableReaderToolboxTool.ts#L84)):*
> ```XML
>                 console.log(
>                     `DEBUG: focusout: createCkEditorBookMarks=${createCkEditorBookMarks}`
>                 );
> ```

We probably don't want to leave this console.log in.

StephenMcConnel (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6314)* status: 0 of 3 files reviewed, 2 unresolved discussions (waiting on @andrew-polk)

___
*[`src/BloomBrowserUI/bookEdit/toolbox/readers/decodableReader/decodableReaderToolboxTool.ts` line 84 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6314#-NrCMA-T4uAingCdVo0G:-NrD1I8i67HVz06mpYZd:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/9b8dc06654061722e81af7642b5caf4549003bf0/src/BloomBrowserUI/bookEdit/toolbox/readers/decodableReader/decodableReaderToolboxTool.ts#L84)):*

We probably don't want to leave this console.log in.
</blockquote></details>

Done.
___
*[`src/BloomBrowserUI/bookEdit/toolbox/readers/readerToolsModel.ts` line 968 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6314#-NrCMHjyBW7o9M4ilhsQ:-NrD7H8p48ELjLV_02bD:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/9b8dc06654061722e81af7642b5caf4549003bf0/src/BloomBrowserUI/bookEdit/toolbox/readers/readerToolsModel.ts#L968)):*

How easy is it to move this to another file like the one for the style dialog?
Don't do any heroics, but it doesn't seem to belong in readerToolsModel.
</blockquote></details>

Done.

andrew-polk (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6314)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @StephenMcConnel)

PR #6326, opened 2024-02-27
BL-13084 Text Padding

hatton:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6326)

### DIFF
--- a/src/BloomBrowserUI/bookEdit/EditViewFrame.pug
+++ b/src/BloomBrowserUI/bookEdit/EditViewFrame.pug
@@ -22,7 +22,7 @@ html
 			html, body, #left, #right { margin: 0; padding: 0; border: 0 none; height:100%}
 			iframe { border: 0 none;} /* apparently Gecko gives this a default 2 px wide. */
 			body, .pure-pusher { overflow:hidden }
-			#left { width:100%;}
+			#left { width:100%; padding-left:10px;/* separate from the page list */}
 			div#pusherContainer { padding-right: 35px;height: 100%;}
 			/* These rules achieve a change in the width of the frame I added just inside the pure-pusher
 			(the main content of the pure-drawer system) when the check-box is checked. The second line
--- a/src/content/appearanceThemes/appearance-theme-default.css
+++ b/src/content/appearanceThemes/appearance-theme-default.css
@@ -56,21 +56,21 @@
     /* background-color: value in .bloom-page */
     --page-background-color: white;
 
-    --page-text-padding-top: 0em;
-    --page-text-padding-bottom: 0em;
-    --page-text-padding-left: 0em;
-    --page-text-padding-right: 0em;
+    --text-padding-top: 0em;
+    --text-padding-bottom: 0em;
+    --text-padding-left: 0em;
+    --text-padding-right: 0em;
 
     /* padding-left: value in .split-pane.vertical-percent
      *  .split-pane-component.position-right .split-pane-component-inner > div */
-    --page-split-vertical-gap: 9mm; /* Note: 9 was the largest I could get with the current margins before a max a5 Portrait with txt-on-left would overflow.*/
+    --page-verticalSplit-width: 3px; /* 3px is matching Bloom 5.6's right margin for vertical split */
     /* padding-top: value in .split-pane.horizontal-percent
      *  .split-pane-component.position-bottom .split-pane-component-inner > div */
-    --page-split-horizontal-gap: 5mm; /* note that em units won't work as this is applied to a splitter that doesn't know the font size */
+    --page-horizontalSplit-height: 0;
     /* gap: value for .bloom-translationGroup */
     /* When Bloom is configured to show each block of text in more than one language, this controls
        the space between the different language sections of each block. */
-    --multilingual-editable-vertical-gap: 10px;
+    --multilingual-editable-vertical-gap: 1em;
     /* border-radius: value in .bloom-imageContainer (picture added by user to front cover or content page) */
     --image-border-radius: 0px; /* 0px is the default, but we want to be explicit about square corners */
     /* added to --page-margin-right for padding-right: value in .bloom-page.side-left */
@@ -212,10 +212,10 @@
     --cover-margin-bottom: 15mm;
 }
 
-.numberedPage .bloom-editable {
-    padding-top: var(--page-text-padding-top);
-    padding-bottom: var(--page-text-padding-bottom);
-    padding-right: var(--page-text-padding-right);
-    padding-left: var(--page-text-padding-left);
+/* notice that we don't do this if the edit is inside an image (i.e. and overlay) */
+.numberedPage :not(.bloom-imageContainer) .bloom-editable {
+    padding-top: var(--text-padding-top);
+    padding-bottom: var(--text-padding-bottom);
+    padding-right: var(--text-padding-right);
+    padding-left: var(--text-padding-left);
 }
-/* TODO: don't apply these headings if we're in an overlay */
--- a/src/content/appearanceThemes/appearance-theme-default.css
+++ b/src/content/appearanceThemes/appearance-theme-default.css
@@ -63,10 +63,10 @@
 
     /* padding-left: value in .split-pane.vertical-percent
      *  .split-pane-component.position-right .split-pane-component-inner > div */
-    --page-split-vertical-gap: 2mm;
+    --page-split-vertical-gap: 9mm; /* Note: 9 was the largest I could get with the current margins before a max a5 Portrait with txt-on-left would overflow.*/
     /* padding-top: value in .split-pane.horizontal-percent
      *  .split-pane-component.position-bottom .split-pane-component-inner > div */
-    --page-split-horizontal-gap: 1mm;
+    --page-split-horizontal-gap: 5mm; /* note that em units won't work as this is applied to a splitter that doesn't know the font size */
     /* gap: value for .bloom-translationGroup */
     /* When Bloom is configured to show each block of text in more than one language, this controls
        the space between the different language sections of each block. */
--- a/src/content/appearanceThemes/appearance-theme-default.css
+++ b/src/content/appearanceThemes/appearance-theme-default.css
@@ -55,6 +55,12 @@
 
     /* background-color: value in .bloom-page */
     --page-background-color: white;
+
+    --page-text-padding-top: 0em;
+    --page-text-padding-bottom: 0em;
+    --page-text-padding-left: 0em;
+    --page-text-padding-right: 0em;
+
     /* padding-left: value in .split-pane.vertical-percent
      *  .split-pane-component.position-right .split-pane-component-inner > div */
     --page-split-vertical-gap: 2mm;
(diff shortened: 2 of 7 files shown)


JohnThomson (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6326)* status: all files reviewed, 4 unresolved discussions (waiting on @hatton)

___
*[`src/content/appearanceThemes/appearance-theme-default.css` line 61 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6326#-NrlMHi72je4E1colY06:-NrlMHi72je4E1colY07:b-118ifi) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/bd1cb3a9a63c67d03413ae3b9ecc835a54cdade8/src/content/appearanceThemes/appearance-theme-default.css#L61)):*
> ```CSS
>     --page-text-padding-bottom: 0em;
>     --page-text-padding-left: 0em;
>     --page-text-padding-right: 0em;
> ```

I'm not clear why there is a 'page-' at the start of these. They control padding for text (specifically bloom-editables on numbered pages). numberedPage-... might make sense, but I would call them just text-padding-left (etc).
___
*[`src/content/appearanceThemes/appearance-theme-default.css` line 223 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6326#-NrlLhu6BbpdAq7TrwEi:-NrlLhu6BbpdAq7TrwEj:b-pfhaus) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/bd1cb3a9a63c67d03413ae3b9ecc835a54cdade8/src/content/appearanceThemes/appearance-theme-default.css#L223)):*
> ```CSS
>     padding-left: var(--page-text-padding-left);
> }
> /* TODO: don't apply these headings if we're in an overlay */
> ```

Do you want to do this before merging this PR? If not, does it need to be a card?
___
*[`src/content/appearanceThemes/appearance-theme-zero-margin-ebook.css` line 60 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6326#-NrlN929Ai363BPQs009:-NrlN929Ai363BPQs00A:b-dqef9h) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/bd1cb3a9a63c67d03413ae3b9ecc835a54cdade8/src/content/appearanceThemes/appearance-theme-zero-margin-ebook.css#L60)):*
> ```CSS
> }
> 
> .Device16x9Portrait.numberedPage {
> ```

It's not obvious why top and bottom padding is appropriate for interactive pages, while left and right padding are not. If it was intentional, it deserves a comment.
Also, it would be nice to move this block near (if not into) the block that controls the other text padding variables.
___
*[`src/content/bookLayout/basePage.less` line 564 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6326#-NrlOWue3d5OG2mWu3Ru:-NrlOWue3d5OG2mWu3Rv:b-32kz6z) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/bd1cb3a9a63c67d03413ae3b9ecc835a54cdade8/src/content/bookLayout/basePage.less#L564)):*
> ```Less
> }
> 
> // See comments below about the horizontal version of this rule.
> ```

This comment is now incomplete, because the horizontal version does NOT put space on either or both the top and bottom split like this one. Also it doesn't explain why we want the TG-only padding. (And it is puzzling. In our earlier discussions, and from comments above, I thought we did NOT want padding on just one side of a TG, because we wanted the new bloom-editable padding to usually be the same left and right and look even.)
(Also...we don't have any built-in pages where a split has a TG on both sides, but if someone made one, would you want a double page-split-vertical-gap?)

JohnThomson (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6326)* status: all files reviewed, 8 unresolved discussions (waiting on @hatton)

___
*[`src/content/appearanceThemes/appearance-theme-default.css` line 215 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6326#-NtcH_Ir2RSuQJOjOZLC:-NtcH_Ir2RSuQJOjOZLD:bix6ewp) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/a4c5591915a5feff4ce2b81b8f05ce417f652936/src/content/appearanceThemes/appearance-theme-default.css#L215)):*
> ```CSS
> }
> 
> /* notice that we don't do this if the edit is inside an image (i.e. and overlay) */
> ```

"an overlay" (not "and")
___
*[`src/content/appearanceThemes/appearance-theme-default.css` line 216 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6326#-NtcJeKeAxY7e0HhQ_xS:-NtcJeKeAxY7e0HhQ_xT:bip1ona) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/a4c5591915a5feff4ce2b81b8f05ce417f652936/src/content/appearanceThemes/appearance-theme-default.css#L216)):*
> ```CSS
> 
> /* notice that we don't do this if the edit is inside an image (i.e. and overlay) */
> .numberedPage :not(.bloom-imageContainer) .bloom-editable {
> ```

If top and bottom apply to a bloom-editable, they will both be added to multilingual-editable-vertical-gap between languages in a multilingual book. Is that what we want? (Of course unlike the gap they will be inside the box we draw around the editable.)
Also: bloom-editables occur inside image containers not only for overlays, but also for image descriptions, IIRC. Do we want to suppress this there, too?
___
*[`src/content/bookLayout/basePage.less` line 554 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6326#-NtcKzSy2C_fR2Qq0tDX:-NtcKzSy2C_fR2Qq0tDY:buqlmed) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/a4c5591915a5feff4ce2b81b8f05ce417f652936/src/content/bookLayout/basePage.less#L554)):*
> ```Less
>     // and we're doing that so that when you snap something to 50%, both sides will be the same size.
>     // Related to BL-13073.
>     padding-right: calc(var(--page-verticalSplit-width) / 2);
> ```

It feels like it ought to be possible to have the value that says "split at 60%" in one place. Then we ought to be able to do a flex-box on the .split-pane.vertical-percent and give it a gap: var(--page-verticalSplit-width). Then set the .split-pane-component.position-left
    > .split-pane-component-inner' width to calc((100% - var(--page-verticalSplit-width)) * var(--split-percentage)) and the other one to something similar. Then each box would really be the right percentage of the non-gap space.
But I don't see how to get the UI-controlled percentage into a variable that applies to just this one spot.
___
*[`src/content/bookLayout/basePage.less` line 560 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6326#-NtcItvBCVQ8QN3zSHTy:-NtcItvBCVQ8QN3zSHTz:b-c2b1l1) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/a4c5591915a5feff4ce2b81b8f05ce417f652936/src/content/bookLayout/basePage.less#L560)):*
> ```Less
>     .split-pane-component.position-right
>     > .split-pane-component-inner {
>     padding-left: calc(var(--page-verticalSplit-width) / 2);
> ```

I'm not clear why we're doing half-on-the-left, half-on-the-right, but the horizontal one uses all-on-the-bottom.

hatton (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6326)* status: all files reviewed, 8 unresolved discussions (waiting on @JohnThomson)

___
*[`src/content/appearanceThemes/appearance-theme-default.css` line 61 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6326#-NrlMHi72je4E1colY06:-NtX_UYBD5k130X1oRle:bu1yydv) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/bd1cb3a9a63c67d03413ae3b9ecc835a54cdade8/src/content/appearanceThemes/appearance-theme-default.css#L61)):*

I'm not clear why there is a 'page-' at the start of these. They control padding for text (specifically bloom-editables on numbered pages). numberedPage-... might make sense, but I would call them just text-padding-left (etc).
</blockquote></details>

Well, that's embarrassing!
___
*[`src/content/appearanceThemes/appearance-theme-default.css` line 223 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6326#-NrlLhu6BbpdAq7TrwEi:-NtXaJMNEdA5qBJvYvls:b60r506) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/bd1cb3a9a63c67d03413ae3b9ecc835a54cdade8/src/content/appearanceThemes/appearance-theme-default.css#L223)):*

Do you want to do this before merging this PR? If not, does it need to be a card?
</blockquote></details>

Yes, done.
___
*[`src/content/appearanceThemes/appearance-theme-zero-margin-ebook.css` line 60 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6326#-NrlN929Ai363BPQs009:-NtXbLTlAWx_yc-Enviw:b-d8mu89) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/bd1cb3a9a63c67d03413ae3b9ecc835a54cdade8/src/content/appearanceThemes/appearance-theme-zero-margin-ebook.css#L60)):*
> It's not obvious why top and bottom padding is appropriate for interactive pages, while left and right padding are not. If it was intentional, it deserves a comment.

Yeah I don't remember what that was about. Tested without and it seems right. Removed for now.

> Also, it would be nice to move this block near (if not into) the block that controls the other text padding variables.

I'm not clear what you're referring to here.
___
*[`src/content/bookLayout/basePage.less` line 564 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6326#-NrlOWue3d5OG2mWu3Ru:-NtXs0jz2WymasVFQgjo:bw3989b) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/bd1cb3a9a63c67d03413ae3b9ecc835a54cdade8/src/content/bookLayout/basePage.less#L564)):*

This comment is now incomplete, because the horizontal version does NOT put space on either or both the top and bottom split like this one. Also it doesn't explain why we want the TG-only padding. (And it is puzzling. In our earlier discussions, and from comments above, I thought we did NOT want padding on just one side of a TG, because we wanted the new bloom-editable padding to usually be the same left and right and look even.)
(Also...we don't have any built-in pages where a split has a TG on both sides, but if someone made one, would you want a double page-split-vertical-gap?)
</blockquote></details>

We removed the bloom-translationGroup restriction, so it's back to how it was now.

hatton (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6326)* status: all files reviewed, 8 unresolved discussions (waiting on @JohnThomson)

___
*[`src/content/bookLayout/basePage.less` line 554 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6326#-NtcKzSy2C_fR2Qq0tDX:-NtqGIW-CHKdP7_ZN2dj:b-mkdu3w) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/a4c5591915a5feff4ce2b81b8f05ce417f652936/src/content/bookLayout/basePage.less#L554)):*

It feels like it ought to be possible to have the value that says "split at 60%" in one place. Then we ought to be able to do a flex-box on the .split-pane.vertical-percent and give it a gap: var(--page-verticalSplit-width). Then set the .split-pane-component.position-left
    > .split-pane-component-inner' width to calc((100% - var(--page-verticalSplit-width)) * var(--split-percentage)) and the other one to something similar. Then each box would really be the right percentage of the non-gap space.
But I don't see how to get the UI-controlled percentage into a variable that applies to just this one spot.
</blockquote></details>

we decided to stick with what we have

PR #6329, opened 2024-02-28
Adjust (localized) origami toggle switch to display nicely (BL-13129)

StephenMcConnel:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6329)

### DIFF
--- a/src/BloomBrowserUI/bookEdit/css/origamiEditing.less
+++ b/src/BloomBrowserUI/bookEdit/css/origamiEditing.less
@@ -139,11 +139,17 @@
  * Layout Mode (origami) switch (from proto.io)
  *
  */
+.origami-toggle-container {
+    position: absolute;
+    left: 0;
+    top: @ToggleVerticalOffset;
+    width: 100%;
+}
 .origami-toggle {
     cursor: pointer;
     position: absolute;
-    // horizontal position is set in code, see setupOrigami()
-    top: @ToggleVerticalOffset;
+    right: 19px;
+    top: 0;
     line-height: 1em;
     color: @DisabledColor;
     div {
--- a/src/BloomBrowserUI/bookEdit/js/origami.ts
+++ b/src/BloomBrowserUI/bookEdit/js/origami.ts
@@ -21,12 +21,10 @@ export function setupOrigami() {
                 .append(createTypeSelectors(isEnterpriseEnabled))
                 .append(createTextBoxIdentifier());
             $("#page-scaling-container").append(origamiControl);
-            // We want to say "right: 19px" in css, but that should be relative to the
-            // page; and it's now a child of the page-scaling-container which is
-            // a parent of the page and can't use the page width.
-            // So set the left based on the two widths and an extra 19px.
-            const origamiWidth = origamiControl.get(0).clientWidth;
-            origamiControl.get(0).style.left = `${width - origamiWidth - 19}px`;
+            // The container width is set to 100% in the CSS, but we need to
+            // limit it to no more than the actual width of the page.
+            const toggleContainer = $(".origami-toggle-container").get(0);
+            toggleContainer.style.maxWidth = width + "px";
         }
         // I'm not clear why the rest of this needs to wait until we have
         // the two results, but none of the controls shows up if we leave it all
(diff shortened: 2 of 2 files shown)


JohnThomson (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6329)* status: all files reviewed, 1 unresolved discussion (waiting on @StephenMcConnel)

*[a discussion](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6329#-NsPllRjDGYXqZrXf7qd:-NsPllRjDGYXqZrXf7qe:b-f67tqd) (no related file):*
I don't think this is the whole answer. It's definitely a step in the right direction to make measurements after localizing, but even in English, as the issue screenshot shows, the parts of the origami control can wrap, which produces undesirable ugliness, if the window is made narrower than the page.
It would help to apply CSS white-space:nowrap. Then, however wide the window is to begin with, you will measure the full width needed. And it will get clipped, instead of wrapping in an ugly way, if the user later makes the window smaller.
Since we have all that empty space to the left, it feels like it would be better still to keep the control visible, even if the page isn't entirely. One way:
- wrap another div around the .origami-toggle.
- give it width: 100% (so it will match the page-scaling-container, unless otherwise constrained)
- have the javascript set its max-width to the width of the page. (So it won't be wider than that, however big the window)
- set the .origami-toggle to be right:19px.
If I'm right, that will keep the origami control as far right as will make it visible, but not beyond the width of the page. And even without the white-space rule, it won't wrap unless it really runs out of space.
Another advantage is, the Javascript only has to know the page width, so you don't have to wait for localization.
I'm not sure whether @hatton would prefer the control to stay in place relative to the page and get clipped with it, or stay visible as the page gets clipped. If we used the extra div AND white-space:nowrap, then in a very narrow window the label would get clipped on the left, but you could still see the actual control, and it would never overlap the page. That would get my vote.

___

StephenMcConnel (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6329)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @hatton and @JohnThomson)

*[a discussion](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6329#-NsPllRjDGYXqZrXf7qd:-NsQGxt07mCN-OGBeA9j:b-804e76) (no related file):*

I don't think this is the whole answer. It's definitely a step in the right direction to make measurements after localizing, but even in English, as the issue screenshot shows, the parts of the origami control can wrap, which produces undesirable ugliness, if the window is made narrower than the page.
It would help to apply CSS white-space:nowrap. Then, however wide the window is to begin with, you will measure the full width needed. And it will get clipped, instead of wrapping in an ugly way, if the user later makes the window smaller.
Since we have all that empty space to the left, it feels like it would be better still to keep the control visible, even if the page isn't entirely. One way:
- wrap another div around the .origami-toggle.
- give it width: 100% (so it will match the page-scaling-container, unless otherwise constrained)
- have the javascript set its max-width to the width of the page. (So it won't be wider than that, however big the window)
- set the .origami-toggle to be right:19px.
If I'm right, that will keep the origami control as far right as will make it visible, but not beyond the width of the page. And even without the white-space rule, it won't wrap unless it really runs out of space.
Another advantage is, the Javascript only has to know the page width, so you don't have to wait for localization.
I'm not sure whether @hatton would prefer the control to stay in place relative to the page and get clipped with it, or stay visible as the page gets clipped. If we used the extra div AND white-space:nowrap, then in a very narrow window the label would get clipped on the left, but you could still see the actual control, and it would never overlap the page. That would get my vote.
</blockquote></details>

I think my original javascript code worked better than you thought, but following your suggested approach gives a better solution that doesn't require waiting for localization to finish.  With my changes (and possibly with my original code), the control now displays fine on windows as small as 800x600, which I think is all we could possibly want.

___

JohnThomson (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6329)* status: all files reviewed, 1 unresolved discussion (waiting on @hatton)

JohnThomson (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6329)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @StephenMcConnel)

PR #6337, opened 2024-03-04
Fix Bible for Children branding comic to show white back page (BL-13138)

StephenMcConnel:
The final page used to have a white background in comics before we merged the final two backmatter pages back together for comics.

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6337)

### DIFF
--- a/src/BloomExe/Book/Book.cs
+++ b/src/BloomExe/Book/Book.cs
@@ -1820,6 +1820,7 @@ namespace Bloom.Book
             BringXmatterHtmlUpToDate(OurHtmlDom);
             RepairBrokenSmallCoverCredits(OurHtmlDom);
             RepairCoverImageDescriptions(OurHtmlDom);
+            DetectAndMarkDarkCoverColor(OurHtmlDom);
 
             progress.WriteStatus("Repair page label localization");
             RepairPageLabelLocalization(OurHtmlDom);
@@ -2283,6 +2284,22 @@ namespace Bloom.Book
                 // back to 4.6 with the single lang="*" entry, then all of the entries got filled in with the same obsolete data,
                 // so all of them have to be removed.
                 dataDiv.RemoveChild(coverImageDescription);
+            }
+        }
+
+        internal static void DetectAndMarkDarkCoverColor(HtmlDom bookDOM)
+        {
+            var coverColor = GetCoverColorFromDom(bookDOM.RawDom);
+            if (coverColor != null)
+            {
+                var coverColorIsDark = ColorUtils.IsDark(coverColor);
+                foreach (var page in bookDOM.SafeSelectNodes("//div[contains(@class,'coverColor')]").Cast<XmlElement>())
+                {
+                    if (coverColorIsDark)
+                        HtmlDom.AddClass(page, "darkCoverColor");
+                    else
+                        HtmlDom.RemoveClass(page, "darkCoverColor");
+                }
             }
         }
 
@@ -3213,6 +3230,7 @@ namespace Bloom.Book
                     continue;
                 var newContent = regex.Replace(content, "$1" + color);
                 stylesheet.InnerXml = newContent;
+                DetectAndMarkDarkCoverColor(OurHtmlDom);
                 return true;
             }
 
--- a/src/BloomExe/Utils/ColorUtils.cs
+++ b/src/BloomExe/Utils/ColorUtils.cs
@@ -1,0 +1,49 @@
+﻿using System;
+using System.Drawing;
+using Bloom.ImageProcessing;
+
+namespace Bloom.Utils
+{
+    public class ColorUtils
+    {
+        /// <summary>
+        /// Check whether the cover color is dark enough to need changing for some logos.
+        /// </summary>
+        /// <remarks>
+        /// See https://www.w3.org/TR/WCAG20-TECHS/G17.html#G17-procedure
+        /// 7 is the recommended minimal contrast for accessibility.  We can't guarantee
+        /// this, but we can guarantee that the contrast is better than it would be if we
+        /// ignored the cover color.
+        /// luminance is a value between 0 and 1, where 0 is black and 1 is white.
+        /// </remarks>
+        public static bool IsDark(string coverColor)
+        {
+            if (ImageUtils.TryCssColorFromString(coverColor, out Color color))
+            {
+                var R = AdjustColorForLuminance(color.R);
+                var G = AdjustColorForLuminance(color.G);
+                var B = AdjustColorForLuminance(color.B);
+                var luminance = 0.2126 * R + 0.7152 * G + 0.0722 * B;
+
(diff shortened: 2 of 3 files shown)


JohnThomson (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6337)* status: all files reviewed, 1 unresolved discussion (waiting on @StephenMcConnel)

___
*[`src/BloomExe/Publish/BloomPub/BloomPubMaker.cs` line 602 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6337#-NsPzzS3EuBG1Wtvoi_b:-NsPzzS3EuBG1Wtvoi_c:b-n05t4f) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/8926cdf4007dc3bdeb67254cc72fbb8e14a44fe6/src/BloomExe/Publish/BloomPub/BloomPubMaker.cs#L602)):*
> ```Smalltalk
>                 }
>             }
>         }
> ```

It seems very unprincipled to take a meta value that was intended to prevent automatically generating a cover color, reinterpret it to mean "this book has a black cover", and then apply it to only one of the four cover pages, just so we can make a rule for one branding.
What would you think of applying some logic like https://stackoverflow.com/questions/9733288/how-to-programmatically-calculate-the-contrast-ratio-between-two-colors to the cover color and use it to add (or not add) to each cover page a class darkBackground? Make it present if white text will contrast better than black.
It would then actually make sense to say that if a Bible-for-children book has a dark back cover, we need to force it white to make their logo readable. Whereas it's completely counter-intuitive to change the cover color because there is an attribute present that apparently tells us to preserve it (!!).
It would also have a chance of being useful for other things. For example, we could have a single principled rule somewhere to say that when this class is present the foreground color should be white, instead of embedding a whole bunch of rules about that in the Digital Comic Book template. But that's beyond the scope of this card.
The tricky bit is when to do this. Any time we change the cover color, of course, it needs updating. And somehow it has to get introduced when migrating books that need it to 5.7. To minimize the cost of that it could be a new 'maintenanceLevel'. So it would be a bit more code, but I think it might be worth it.
This is the sort of thing where @hatton might have an even more elegant idea.
Feel free to push back or think of another idea. But I'd really like to avoid reinterpreting 'preserveCoverColor' as 'cover is dark'. (One day some branding may always want a yellow cover.)

StephenMcConnel (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6337)* status: all files reviewed, 1 unresolved discussion (waiting on @hatton)

___
*[`src/BloomExe/Publish/BloomPub/BloomPubMaker.cs` line 602 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6337#-NsPzzS3EuBG1Wtvoi_b:-NstscUABhmV1d5F0nSf:bduyxhz) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/8926cdf4007dc3bdeb67254cc72fbb8e14a44fe6/src/BloomExe/Publish/BloomPub/BloomPubMaker.cs#L602)):*

It seems very unprincipled to take a meta value that was intended to prevent automatically generating a cover color, reinterpret it to mean "this book has a black cover", and then apply it to only one of the four cover pages, just so we can make a rule for one branding.
What would you think of applying some logic like https://stackoverflow.com/questions/9733288/how-to-programmatically-calculate-the-contrast-ratio-between-two-colors to the cover color and use it to add (or not add) to each cover page a class darkBackground? Make it present if white text will contrast better than black.
It would then actually make sense to say that if a Bible-for-children book has a dark back cover, we need to force it white to make their logo readable. Whereas it's completely counter-intuitive to change the cover color because there is an attribute present that apparently tells us to preserve it (!!).
It would also have a chance of being useful for other things. For example, we could have a single principled rule somewhere to say that when this class is present the foreground color should be white, instead of embedding a whole bunch of rules about that in the Digital Comic Book template. But that's beyond the scope of this card.
The tricky bit is when to do this. Any time we change the cover color, of course, it needs updating. And somehow it has to get introduced when migrating books that need it to 5.7. To minimize the cost of that it could be a new 'maintenanceLevel'. So it would be a bit more code, but I think it might be worth it.
This is the sort of thing where @hatton might have an even more elegant idea.
Feel free to push back or think of another idea. But I'd really like to avoid reinterpreting 'preserveCoverColor' as 'cover is dark'. (One day some branding may always want a yellow cover.)
</blockquote></details>

Done.  I decided that since the darkCoverColor class applies only to xmatter pages, it needs to be done whenever xmatter updates, so it's not really a migration issue per se.  But I don't think it's too expensive.

JohnThomson (approved):
:lgtm: But I'd like Andrew's or John's opinion of the idea before merging.

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6337#-:-NtvoAlyExFvo0vC2-Tf:b-g6swyv)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @StephenMcConnel)

andrew-polk (commented):
The overall approach makes sense... detect we have a dark cover and allow the css to change things based on a class indicating that.

Do we have an idea of the performance cost?
A couple ideas if we thought it might be costly at all:
1. JT had suggested a migration. 
2. We could hard-code a look up for the values we expect, such as this dark one and the default colors we assign (and maybe the defaults from the color picker, too?).

JT in a separate thread brought up his second-guessing his suggestion of the class name .darkCoverColor, wondering if it should have the bloom- prefix but noting that we have other related classes without it now so we would lose some consistency. That's a hard call, but I guess I would lean toward keeping it more like the existing classes (so without the prefix).

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6337#-:-Ntw53YI6soEYRAS6Aln:b-1ui4fo)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @StephenMcConnel)

StephenMcConnel (commented):
I fixed the code to predefine the luminance for white (which is 1.0 after all), and tried measuring how long it tooks to compute the IsDark value.  The answer for method duration always came back as 0 ms and then as 0 ticks when I tried finer granularity.  I don't think this computation is going to be very costly.  :-)
Since we have to check this every time the xmatter gets rebuilt, I don't think a migration per se is appropriate or necessary.

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6337#-:-NuV3xuK0K0Opb-ng5ur:b-5aupaw)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @StephenMcConnel)

andrew-polk (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6337)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @StephenMcConnel)

PR #6361, opened 2024-03-26
Prevent localization exceptions in unit tests (20240326)

StephenMcConnel:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6361)

### DIFF
--- a/src/BloomExe/Program.cs
+++ b/src/BloomExe/Program.cs
@@ -1491,6 +1491,9 @@ namespace Bloom
                 applicationContainer = applicationContainerSource;
             var installedStringFileFolder =
                 FileLocationUtilities.GetDirectoryDistributedWithApplication(true, "localization");
+            var productVersion = Application.ProductVersion;
+            if (Program.RunningUnitTests)
+                productVersion = "1.2.3"; // Prevent invalid product version from being used in unit tests.
             if (installedStringFileFolder == null)
             {
                 // nb do NOT try to localize this...it's a shame, but the problem we're reporting is that the localization data is missing!
@@ -1510,7 +1513,7 @@ namespace Bloom
                     "en",
                     "Bloom",
                     "Bloom",
-                    Application.ProductVersion,
+                    productVersion,
                     fakeLocalDir,
                     "SIL/Bloom",
                     Resources.BloomIcon,
@@ -1531,7 +1534,7 @@ namespace Bloom
                     desiredLanguage,
                     "Bloom",
                     "Bloom",
-                    Application.ProductVersion,
+                    productVersion,
                     installedStringFileFolder,
                     "SIL/Bloom",
                     Resources.BloomIcon,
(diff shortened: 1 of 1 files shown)


andrew-polk (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6361)* status: all files reviewed, 1 unresolved discussion (waiting on @StephenMcConnel)

___
*[`src/BloomExe/Program.cs` line 1496 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6361#-Ntx-Z_z3eOm_NYGvUbQ:-Ntx-Z_z3eOm_NYGvUbR:b-3jarn4) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/0d1d21fb4ed375951f7f406da9522bd9747d22c1/src/BloomExe/Program.cs#L1496)):*
> ```Smalltalk
>             var productVersion = Application.ProductVersion;
>             if (Program.RunningUnitTests)
>                 productVersion = "5.7.0"; // Prevent random product version from being used in unit tests.
> ```

I'm curious how this manifested in a unit test failure.

Specifically, do we need a real version number here? 
Does it have to match the actual version number?

Obviously, we don't want to have to update this every version.

Another thing to note...
JT already did something similar but different for ErrorReport.VersionNumberString in Program.cs for this same new test.

StephenMcConnel (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6361)* status: all files reviewed, 1 unresolved discussion (waiting on @andrew-polk)

___
*[`src/BloomExe/Program.cs` line 1496 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6361#-Ntx-Z_z3eOm_NYGvUbQ:-Ntx0rLPCIkvh1PH0dTo:b-s8iaf5) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/0d1d21fb4ed375951f7f406da9522bd9747d22c1/src/BloomExe/Program.cs#L1496)):*

I'm curious how this manifested in a unit test failure.

Specifically, do we need a real version number here? 
Does it have to match the actual version number?

Obviously, we don't want to have to update this every version.

Another thing to note...
JT already did something similar but different for ErrorReport.VersionNumberString in Program.cs for this same new test.
</blockquote></details>

We don't need the "real" version number, but we need to ensure that we have one that is parsed by the C# Version class constructor.  I was seeing something like "17.9.0-release-24037-02" in the unit test, which I think was coming from Visual Studio.  For L10NSharp instantiating the Version objects, it needs to be of the form "1.2.3.4" (or "3.2.1").  Generated xliff files were created with the strange product version, and persisted the crashing until I deleted them.

andrew-polk (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6361)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @StephenMcConnel)

PR #6374, opened 2024-03-28
BL-12872 remove remaining parse server code

nabalone:
Depends on https://github.com/BloomBooks/bloom-azure-functions/pull/83

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6374)


andrew-polk (changes requested):
Yay for lots of red!

> Reviewed 19 of 19 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6374#-:-Nu4qUIF3PV_Xd0KlZT8:bd827s6)* status: all files reviewed, 5 unresolved discussions (waiting on @nabalone)

___
*[`src/BloomExe/app.config` line 115 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6374#-Nu4jj3VDyn1669nEx-K:-Nu4jj3VDyn1669nEx-L:bk83wyd) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/c85cd05c68d4c4ed2328a70420f1d231474d434f/src/BloomExe/app.config#L115)):*
> ```
>         <value />
>       </setting>
>       <setting name="LastLoginDatabaseObjectId" serializeAs="String">
> ```

Normally, we would need to be careful about changing setting names like this since the user will lose their previous setting. But in this case, I think it is ok.

"objectId" is a parse-specific thing we are trying to move away from.
How about `LastLoginDatabaseUserId` or `LastLoginUserId`?
___
*[`src/BloomExe/WebLibraryIntegration/BloomLibraryBookApiClient.cs` line 125 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6374#-Nu4p8OdCGbqM8Ri2JLK:-Nu4p8OdCGbqM8Ri2JLL:b-9rgamv) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/c85cd05c68d4c4ed2328a70420f1d231474d434f/src/BloomExe/WebLibraryIntegration/BloomLibraryBookApiClient.cs#L125)):*
> ```Smalltalk
>                         errorMessage = messageToShowUserOnFailure;
>                     }
>                     string messageIntendedForUser = responseContentError.messageIntendedForUser;
> ```

This probably would benefit from some comments explaining it.
Something like

As of March 2024, this is not used, but we wanted to have a mechanism by which the API could give messages directly to the user. For example, we might shut down uploads for alpha temporarily. This might be used with or without the ClientOutOfDate error code.
Note, we do not anticipate this message would be localized.
___
*[`src/BloomExe/WebLibraryIntegration/BloomLibraryBookApiClient.cs` line 321 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6374#-Nu4oByZ1mHCi4U9LM6_:-Nu4oByZ1mHCi4U9LM6a:byjaen2) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/c85cd05c68d4c4ed2328a70420f1d231474d434f/src/BloomExe/WebLibraryIntegration/BloomLibraryBookApiClient.cs#L321)):*
> ```Smalltalk
>         public void SetLoginData(
>             string account,
>             string userObjectId,
> ```

How about either databaseUserId or userId?

userId matches the class member.
___
*[`src/BloomExe/web/controllers/LibraryPublishApi.cs` line 189 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6374#-Nu4ofg23TnIQ9dknleN:-Nu4ofg23TnIQ9dknleO:blf6i93) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/7f4143dd5ff7eeb57b0e9f410c14758be43ed2a6/src/BloomExe/web/controllers/LibraryPublishApi.cs#L189)):*
> CheckingVersionEligibility

We can mark this string as obsolete in the .xlf file.
(See other instances of doing so there)
___
*[`src/BloomTests/WebLibraryIntegration/BookUploadAndDownloadTests.cs` line 445 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6374#-Nu4mJ_C4yxblDycs33o:-Nu4mJ_C4yxblDycs33p:byz0y90) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/c85cd05c68d4c4ed2328a70420f1d231474d434f/src/BloomTests/WebLibraryIntegration/BookUploadAndDownloadTests.cs#L445)):*
> ```Smalltalk
>                     new BloomLibraryBookApiClientTestDouble();
>                 bogusAccountClient.TestOnly_LogIn("bogus_account@example.com");
>                 Assert.IsNull(bogusAccountClient.GetSingleBookRecord(bookInstanceId));
> ```

Sorry, I'm not understanding this part of the test. Why doesn't another user get to query for the book record?
As far as I know, book records are public read.

andrew-polk (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6374)* status: all files reviewed, 6 unresolved discussions (waiting on @nabalone)

___
*[`src/BloomTests/WebLibraryIntegration/BloomLibraryBookApiClientTests.cs` line 44 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6374#-Nu4sHwn9ABHhDAbpMfG:-Nu4sHwn9ABHhDAbpMfH:b-nctpp0) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/c85cd05c68d4c4ed2328a70420f1d231474d434f/src/BloomTests/WebLibraryIntegration/BloomLibraryBookApiClientTests.cs#L44)):*
> ```Smalltalk
>         // Fake log in so that the unit tests can upload, since we haven't been able to get the new firebase login to work
>         // except when Bloom is actually running.
>         public void TestOnly_LogIn(string account)
> ```

I'm hesitant to call this LogIn.

What about TestOnly_SetUserAccountInfo?

Same with the caller of this in the other test class.

But I'm questioning if we even need it.

nabalone (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6374)* status: 13 of 20 files reviewed, 6 unresolved discussions (waiting on @andrew-polk)

___
*[`src/BloomExe/app.config` line 115 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6374#-Nu4jj3VDyn1669nEx-K:-NuPidVO8QSq1LL79fPW:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/c85cd05c68d4c4ed2328a70420f1d231474d434f/src/BloomExe/app.config#L115)):*

Normally, we would need to be careful about changing setting names like this since the user will lose their previous setting. But in this case, I think it is ok.

"objectId" is a parse-specific thing we are trying to move away from.
How about `LastLoginDatabaseUserId` or `LastLoginUserId`?
</blockquote></details>

Done.
___
*[`src/BloomExe/WebLibraryIntegration/BloomLibraryBookApiClient.cs` line 125 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6374#-Nu4p8OdCGbqM8Ri2JLK:-NuPjC-l5o8jQZvrKyJO:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/c85cd05c68d4c4ed2328a70420f1d231474d434f/src/BloomExe/WebLibraryIntegration/BloomLibraryBookApiClient.cs#L125)):*

This probably would benefit from some comments explaining it.
Something like

As of March 2024, this is not used, but we wanted to have a mechanism by which the API could give messages directly to the user. For example, we might shut down uploads for alpha temporarily. This might be used with or without the ClientOutOfDate error code.
Note, we do not anticipate this message would be localized.
</blockquote></details>

Done.
___
*[`src/BloomExe/WebLibraryIntegration/BloomLibraryBookApiClient.cs` line 321 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6374#-Nu4oByZ1mHCi4U9LM6_:-NuPjK0g9pyryLp5UqjE:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/c85cd05c68d4c4ed2328a70420f1d231474d434f/src/BloomExe/WebLibraryIntegration/BloomLibraryBookApiClient.cs#L321)):*

How about either databaseUserId or userId?

userId matches the class member.
</blockquote></details>

Done.
___
*[`src/BloomExe/web/controllers/LibraryPublishApi.cs` line 189 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6374#-Nu4ofg23TnIQ9dknleN:-NuPoDLYBtw60hQkO8rd:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/7f4143dd5ff7eeb57b0e9f410c14758be43ed2a6/src/BloomExe/web/controllers/LibraryPublishApi.cs#L189)):*

> CheckingVersionEligibility

We can mark this string as obsolete in the .xlf file.
(See other instances of doing so there)
</blockquote></details>

Done.
___
*[`src/BloomTests/WebLibraryIntegration/BloomLibraryBookApiClientTests.cs` line 44 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6374#-Nu4sHwn9ABHhDAbpMfG:-NuQKVMu-FB6Kd2Fj60p:b5j85o7) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/c85cd05c68d4c4ed2328a70420f1d231474d434f/src/BloomTests/WebLibraryIntegration/BloomLibraryBookApiClientTests.cs#L44)):*

I'm hesitant to call this LogIn.

What about TestOnly_SetUserAccountInfo?

Same with the caller of this in the other test class.

But I'm questioning if we even need it.
</blockquote></details>

Renamed.

We have to set `_authentication_token` to something in order for the BloomLibraryBookApiClient to think we are logged in and so let us upload: https://github.com/BloomBooks/BloomDesktop/blob/960066e1807e177716d3a2b98b9d824fbc6c1c43/src/BloomExe/WebLibraryIntegration/BloomLibraryBookApiClient.cs#L384

We have to set `Account` because `GetSingleBookRecord` filters by account (which maybe we should discuss further in your other comment below)
___
*[`src/BloomTests/WebLibraryIntegration/BookUploadAndDownloadTests.cs` line 445 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6374#-Nu4mJ_C4yxblDycs33o:-NuQHUUT6tR8zJXfdlFu:b-hn3e2s) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/c85cd05c68d4c4ed2328a70420f1d231474d434f/src/BloomTests/WebLibraryIntegration/BookUploadAndDownloadTests.cs#L445)):*

Sorry, I'm not understanding this part of the test. Why doesn't another user get to query for the book record?
As far as I know, book records are public read.
</blockquote></details>

Was trying to replicate the test behavior I am removing here: https://github.com/BloomBooks/BloomDesktop/blob/960066e1807e177716d3a2b98b9d824fbc6c1c43/src/BloomTests/WebLibraryIntegration/BloomLibraryBookApiClientTests.cs#L106C1-L109C60

`GetBookRecords` and thus `GetSingleBookRecord` filters by the currently logged in `Account` by default. Perhaps they could be named more clearly https://github.com/BloomBooks/BloomDesktop/blob/960066e1807e177716d3a2b98b9d824fbc6c1c43/src/BloomExe/WebLibraryIntegration/BloomLibraryBookApiClient.cs#L493

I tried to clarify the comment above.

andrew-polk (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6374)* status: all files reviewed, 2 unresolved discussions (waiting on @nabalone)

___
*[`src/BloomTests/WebLibraryIntegration/BookUploadAndDownloadTests.cs` line 445 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6374#-Nu4mJ_C4yxblDycs33o:-NuUxfAo5MZmVNNjxxcF:b-4aamkw) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/c85cd05c68d4c4ed2328a70420f1d231474d434f/src/BloomTests/WebLibraryIntegration/BookUploadAndDownloadTests.cs#L445)):*

Was trying to replicate the test behavior I am removing here: https://github.com/BloomBooks/BloomDesktop/blob/960066e1807e177716d3a2b98b9d824fbc6c1c43/src/BloomTests/WebLibraryIntegration/BloomLibraryBookApiClientTests.cs#L106C1-L109C60

`GetBookRecords` and thus `GetSingleBookRecord` filters by the currently logged in `Account` by default. Perhaps they could be named more clearly https://github.com/BloomBooks/BloomDesktop/blob/960066e1807e177716d3a2b98b9d824fbc6c1c43/src/BloomExe/WebLibraryIntegration/BloomLibraryBookApiClient.cs#L493

I tried to clarify the comment above.
</blockquote></details>

Ok, got it. Thanks for clarifying.

As far as I can tell, `GetSingleBookRecord` is now only called by unit tests.
If true, it seems like we should move it into the test double class and rename it to something like `TestOnly_GetSingleBookRecordForCurrentUser`.

On the other hand, it looks like there is only one other caller of `GetBookRecords` and that wants all books regardless of uploader.
So if it is simple to clean up the unit tests to just not care about uploader, I think you could change `GetBookRecords` to ignore uploader.

At least for this particular check you are trying to reproduce, I think we can get rid of it.

What do you think? Let me know if you want to discuss.
___
*[`src/BloomTests/WebLibraryIntegration/BookUploadAndDownloadTests.cs` line 123 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6374#-NuUw-t3-vSsmsB-JSdu:-NuUw-t4A5bcZ3KMTkcR:b-z54ifx) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/76278e27aaf06b29d95691d30ea392bd264a898d/src/BloomTests/WebLibraryIntegration/BookUploadAndDownloadTests.cs#L123)):*
> Login

How about SetUpForUnitTestUser or similar?

nabalone (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6374)* status: 16 of 20 files reviewed, 2 unresolved discussions (waiting on @andrew-polk)

___
*[`src/BloomTests/WebLibraryIntegration/BookUploadAndDownloadTests.cs` line 445 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6374#-Nu4mJ_C4yxblDycs33o:-NuVHV9v4I-SSHHDYSwE:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/c85cd05c68d4c4ed2328a70420f1d231474d434f/src/BloomTests/WebLibraryIntegration/BookUploadAndDownloadTests.cs#L445)):*

Ok, got it. Thanks for clarifying.

As far as I can tell, `GetSingleBookRecord` is now only called by unit tests.
If true, it seems like we should move it into the test double class and rename it to something like `TestOnly_GetSingleBookRecordForCurrentUser`.

On the other hand, it looks like there is only one other caller of `GetBookRecords` and that wants all books regardless of uploader.
So if it is simple to clean up the unit tests to just not care about uploader, I think you could change `GetBookRecords` to ignore uploader.

At least for this particular check you are trying to reproduce, I think we can get rid of it.

What do you think? Let me know if you want to discuss.
</blockquote></details>

Done.

andrew-polk (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6374)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @nabalone)

andrew-polk:
Great work!

PR #6376, opened 2024-03-28
Another try at preserving branding (BL-13110)

JohnThomson:
- warn if upload will change branding on Blorg
- allow download-for-edit books to be edited using their original branding
- any time we open settings and the collection used to have a branding but lacks a code, show the old branding and message about it.

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6376)

### DIFF
--- a/DistFiles/localization/en/Bloom.xlf
+++ b/DistFiles/localization/en/Bloom.xlf
@@ -4634,6 +4634,10 @@ Do you want to go ahead?</note>
         <note>ID: PublishTab.UploadCollisionDialog.AlreadyIn</note>
         <note>This is the header for the book that is in bloomlibrary.org already.</note>
       </trans-unit>
+      <trans-unit id="PublishTab.UploadCollisionDialog.ChangeBranding" translate="no">
+        <source xml:lang="en">The branding was "{0}" but is now "{1}". This may change logos and other material. Check this box if this is what you want.</source>
+        <note>ID: PublishTab.UploadCollisionDialog.ChangeBranding</note>
+      </trans-unit>
       <trans-unit id="PublishTab.UploadCollisionDialog.ChangeUploader" translate="no">
         <source xml:lang="en">Change the official uploader to {0}. (Bloom Library will hide part of your email address)</source>
         <note>ID: PublishTab.UploadCollisionDialog.ChangeUploader</note>
@@ -5273,6 +5277,11 @@ Do you want to go ahead?</note>
       <trans-unit id="Settings.Enterprise.LearnMore" sil:dynamic="true">
         <source xml:lang="en">Learn More</source>
         <note>ID: Settings.Enterprise.LearnMore</note>
+      </trans-unit>
+      <trans-unit id="Settings.Enterprise.NeedsCode" sil:dynamic="true">
+        <source xml:lang="en">This project previously used **%0**. To continue to use this you will need to enter a current subscription code. Codes for SIL brandings can be found [here](%1). For other subscriptions, contact your project administrator.</source>
+        <note>ID: Settings.Enterprise.NeedsCode</note>
+        <note>**%0** should be kept. It will become the branding name. Keep something like [here](%1), but replace the word "here" as appropriate. This will become a hyperlink that launches a web page in a browser.</note>
       </trans-unit>
       <trans-unit id="Settings.Enterprise.None" sil:dynamic="true">
         <source xml:lang="en">None. Bloom will not show Enterprise features.</source>
--- a/src/BloomBrowserUI/collection/enterpriseSettings.tsx
+++ b/src/BloomBrowserUI/collection/enterpriseSettings.tsx
@@ -138,12 +138,11 @@ export class EnterpriseSettings extends React.Component<{}, IState> {
                                 l10nParam1={this.codesUrl}
                                 className={"legacyBrandingName"}
                             >
-                                In an older version of Bloom, this project used
-                                **%0**. To continue to use this you will need to
-                                enter a current subscription code. Codes for SIL
-                                brandings can be found [here](%1). For other
-                                subscriptions, contact your project
-                                administrator.
+                                This project previously used **%0**. To continue
+                                to use this you will need to enter a current
+                                subscription code. Codes for SIL brandings can
+                                be found [here](%1). For other subscriptions,
+                                contact your project administrator.
                             </Markdown>
                         )}
                         <div
--- a/src/BloomBrowserUI/collectionsTab/BookButton.tsx
+++ b/src/BloomBrowserUI/collectionsTab/BookButton.tsx
@@ -36,6 +36,7 @@ export const BookButton: React.FunctionComponent<{
     //selected: boolean;
     manager: BookSelectionManager;
     isSpreadsheetFeatureActive: boolean;
+    lockedToOneDownloadedBook: boolean;
 }> = props => {
     // TODO: the c# had Font = bookInfo.IsEditable ? _editableBookFont : _collectionBookFont,
 
@@ -197,7 +198,8 @@ export const BookButton: React.FunctionComponent<{
             {
                 label: "Duplicate Book",
                 l10nId: "CollectionTab.BookMenu.DuplicateBook",
-                command: "collections/duplicateBook"
+                command: "collections/duplicateBook",
+                shouldShow: () => !props.lockedToOneDownloadedBook
             },
             {
                 label: "Show in File Explorer",
--- a/src/BloomBrowserUI/collectionsTab/BooksOfCollection.tsx
+++ b/src/BloomBrowserUI/collectionsTab/BooksOfCollection.tsx
@@ -40,6 +40,7 @@ export const BooksOfCollection: React.FunctionComponent<{
     // If true, the collection will be wrapped in a LazyLoad so that most of its rendering
     // isn't done until it is visible on screen.
     lazyLoadCollection?: boolean;
+    lockedToOneDownloadedBook: boolean;
 }> = props => {
     if (!props.collectionId) {
         window.alert("null collectionId");
@@ -73,13 +74,12 @@ export const BooksOfCollection: React.FunctionComponent<{
     );
 
     useEffect(() => {
-        if (books.length > 0)
-        {
+        if (books.length > 0) {
             // once the books variable has been updated with the book-on-blorg statuses,
             // unset the reloadParameter so we don't keep reloading the book-on-blorg statuses
-            setReloadParameter(""); 
+            setReloadParameter("");
         }
-    }, [books])
+    }, [books]);
 
     //const selectedBookInfo = useMonitorBookSelection();
     const collection: ICollection = useApiData(
@@ -161,6 +161,9 @@ export const BooksOfCollection: React.FunctionComponent<{
                                         manager={props.manager}
                                         isSpreadsheetFeatureActive={
                                             props.isSpreadsheetFeatureActive
+                                        }
+                                        lockedToOneDownloadedBook={
+                                            props.lockedToOneDownloadedBook
                                         }
                                     />
                                 </LazyLoad>
--- a/src/BloomBrowserUI/collectionsTab/CollectionsTabPane.tsx
+++ b/src/BloomBrowserUI/collectionsTab/CollectionsTabPane.tsx
@@ -337,6 +337,8 @@ export const CollectionsTabPane: React.FunctionComponent<{}> = () => {
     const collectionsHeaderKey = "CollectionTab.BookSourceHeading";
     const collectionsHeaderText = "Sources For New Books";
 
+    const lockedToOneDownloadedBook = sourcesCollections.length === 0;
+
     return (
         <div
             css={css`
@@ -468,44 +470,52 @@ export const CollectionsTabPane: React.FunctionComponent<{}> = () => {
                             isSpreadsheetFeatureActive={
                                 isSpreadsheetFeatureActive
                             }
+                            lockedToOneDownloadedBook={
+                                lockedToOneDownloadedBook
+                            }
                         />
                     </div>
 
-                    <Transition in={true} appear={true} timeout={2000}>
-                        {state => (
-                            <div
-                                css={css`
-                                    margin: 10px;
-                                `}
-                                className={`group fade-${state}`}
-                            >
-                                <H1
-                                    l10nKey={collectionsHeaderKey}
+                    {lockedToOneDownloadedBook || (
+                        <Transition in={true} appear={true} timeout={2000}>
+                            {state => (
+                                <div
                                     css={css`
-                                        padding-bottom: 20px;
+                                        margin: 10px;
                                     `}
+                                    className={`group fade-${state}`}
                                 >
-                                    {collectionsHeaderText}
-                                </H1>
+                                    <H1
+                                        l10nKey={collectionsHeaderKey}
+                                        css={css`
+                                            padding-bottom: 20px;
+                                        `}
+                                    >
+                                        {collectionsHeaderText}
+                                    </H1>
 
-                                <ShowAfterDelay
-                                    waitBeforeShow={100} // REview: we really want to wait for an event that indicates the main collection is mostly painted
-                                >
-                                    {collectionComponents}
-                                </ShowAfterDelay>
-                                <Link
-                                    l10nKey="CollectionTab.AddSourceCollection"
-                                    css={css`
-                                        text-transform: uppercase;
-                                        padding-bottom: 10px;
-                                    `}
-                                    onClick={() => addSourceCollection()}
-                                >
-                                    Show another collection...
-                                </Link>
-                            </div>
-                        )}
-                    </Transition>
+                                    <ShowAfterDelay
+                                        waitBeforeShow={100} // REview: we really want to wait for an event that indicates the main collection is mostly painted
+                                    >
+                                        {collectionComponents}
+                                    </ShowAfterDelay>
+                                    <Link
+                                        l10nKey="CollectionTab.AddSourceCollection"
+                                        css={css`
+                                            text-transform: uppercase;
+                                            padding-bottom: 10px;
+                                        `}
+                                        onClick={() => addSourceCollection()}
+                                    >
+                                        Show another collection...
+                                    </Link>
+                                </div>
+                            )}
+                        </Transition>
+                    )
+                    // Enhance:possibly if we're NOT showing the Sources for new Books stuff,
+                    // we could have a message saying why and to pick an Enterprise subscription to fix it.
+                    }
                 </SplitPane>
                 {/* This wrapper is used to... fix up some margin/color stuff I was having trouble with from SplitPane */}
                 <div
@@ -749,6 +759,7 @@ const BooksOfCollectionWithHeading: React.FunctionComponent<{
                 manager={props.manager}
                 lazyLoadCollection={true}
                 isSpreadsheetFeatureActive={props.isSpreadsheetFeatureActive}
+                lockedToOneDownloadedBook={false}
             />
         </div>
     );
--- a/src/BloomBrowserUI/publish/LibraryPublish/LibraryPublishSteps.tsx
+++ b/src/BloomBrowserUI/publish/LibraryPublish/LibraryPublishSteps.tsx
@@ -39,7 +39,6 @@ import { BloomSplitButton } from "../../react_components/bloomSplitButton";
 import { ErrorBox, WaitBox } from "../../react_components/boxes";
 import {
     IUploadCollisionDlgData,
-    IUploadCollisionDlgProps,
     showUploadCollisionDialog,
     UploadCollisionDlg
 } from "./uploadCollisionDlg";
--- a/src/BloomBrowserUI/publish/LibraryPublish/uploadCollisionDlg.tsx
+++ b/src/BloomBrowserUI/publish/LibraryPublish/uploadCollisionDlg.tsx
@@ -51,6 +51,8 @@ export interface IUploadCollisionDlgData {
     existingBookUrl: string;
     existingThumbUrl?: string;
     uploader?: string;
+    oldBranding?: string;
+    newBranding?: string;
     // Note that this may be called some time after the dialog closes (in response to Upload failing,
     // for example, because the user asked us to fix the Book ID but Team Collection said we can't).
     onCancel?: () => void;
@@ -90,6 +92,7 @@ export const UploadCollisionDlg: React.FunctionComponent<IUploadCollisionDlgProp
     );
 
     const [doChangeUploader, setDoChangeUploader] = useState(false);
+    const [doChangeBranding, setDoChangeBranding] = useState(false);
 
     const kAskForHelpColor = "#D65649";
     const kDarkerSecondaryTextColor = "#555555";
@@ -137,6 +140,14 @@ export const UploadCollisionDlg: React.FunctionComponent<IUploadCollisionDlgProp
         "Bloom will fix the ID of your book and upload it as a new book. The old book on Bloom Library will stay the same.",
         "PublishTab.UploadCollisionDialog.Radio.DifferentBooks.Commentary2",
         "This is explanatory commentary on a radio button."
+    );
+
+    const changeBrandingMessage = useL10n(
+        'The branding was "{0}" but is now "{1}". This may change logos and other material. Check this box if this is what you want.',
+        "PublishTab.UploadCollisionDialog.ChangeBranding",
+        "",
+        props.oldBranding,
+        props.newBranding
     );
 
     const sameBookRadioLabel = useL10n(
@@ -249,6 +260,9 @@ export const UploadCollisionDlg: React.FunctionComponent<IUploadCollisionDlgProp
         </div>
     );
 
+    const needChangeBranding =
+        props.oldBranding && props.oldBranding !== props.newBranding;
+
     const differentBooksRadioCommentary = (): JSX.Element => (
         <div
             css={css`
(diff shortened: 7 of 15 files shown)


hatton (changes requested):
> Reviewed 15 of 17 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6376)* status: 15 of 17 files reviewed, 10 unresolved discussions (waiting on @JohnThomson)

___
*[`src/BloomBrowserUI/collection/enterpriseSettings.tsx` line 136 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6376#-NuVnyRpAORSVZ7YuVLp:-NuVnyRpAORSVZ7YuVLq:b-6slp9u) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/ce9b500aa3d93f477eaac2111400e2cd05cca541/src/BloomBrowserUI/collection/enterpriseSettings.tsx#L136)):*
> ```XML
>                         {this.state.controlState === "SubscriptionLegacy" && (
>                             <Markdown
>                                 l10nKey="Settings.Enterprise.NeedsCode"
> ```

Should we change the id so that localizations get the new text?
___
*[`src/BloomBrowserUI/react_components/requiresBloomEnterprise.tsx` line 40 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6376#-NuVoq3cCDhot0NSIvgQ:-NuVoq3cCDhot0NSIvgR:b-c1phy6) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/ce9b500aa3d93f477eaac2111400e2cd05cca541/src/BloomBrowserUI/react_components/requiresBloomEnterprise.tsx#L40)):*
> ```XML
>  * @returns A boolean, which is true if Bloom Enterprise is enabled and false otherwise
>  */
> export function useEnterpriseAvailable(failIfLockedToOneBook?: boolean) {
> ```

Probably if I thought harder it would occur to me eventually, but can you add a comment explaining why this parameter is needed here? It's not obvious why an API asking if BE is available should "fail" if we are locked to one book.
___
*[`src/BloomBrowserUI/react_components/requiresBloomEnterprise.tsx` line 188 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6376#-NuVpFOk39A1TNjQlJlh:-NuVpFOk39A1TNjQlJli:b-tj8oun) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/ce9b500aa3d93f477eaac2111400e2cd05cca541/src/BloomBrowserUI/react_components/requiresBloomEnterprise.tsx#L188)):*
> ```XML
>  * Checks the Bloom Enterprise settings and overlays a RequiresBloomEnterprise notice over the children if enterprise is off.
>  */
> export const RequiresBloomEnterpriseOverlayWrapper: React.FunctionComponent<{
> ```

Pls comment on this param too
___
*[`src/BloomExe/Collection/CollectionSettings.cs` line 393 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6376#-NuVpeu90zZHUe9Pi25S:-NuVpeu90zZHUe9Pi25T:b-wrtdxc) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/ce9b500aa3d93f477eaac2111400e2cd05cca541/src/BloomExe/Collection/CollectionSettings.cs#L393)):*
> ```Smalltalk
>         /// as we would do if creating a settings object from the settings file. (This is useful when
>         /// displaying the Branding dialog, to remind the user which branding they might want to find
>         /// a code for. We also use it to record the original branding for a book downloaded for editing,
> ```

To make this a bit clearer to my future self, let's change "a book downloaded for editing" to "a book downloaded using the 'Download for Editing' button on blorg".
___
*[`src/BloomExe/Collection/CollectionSettings.cs` line 397 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6376#-NuVq6b-6PCvje0I5S6R:-NuVq6b-6PCvje0I5S6S:b-dh9twq) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/ce9b500aa3d93f477eaac2111400e2cd05cca541/src/BloomExe/Collection/CollectionSettings.cs#L397)):*
> ```Smalltalk
>         /// to be used, though we make an exception for that one book.)
>         /// </summary>
>         public static string LoadBranding(string pathToCollectionFile)
> ```

`LoadBranding` implied to me a lot more than this does. How about `ReadBrandingNameFromCollectionFile()`
___
*[`src/BloomExe/Collection/CollectionSettings.cs` line 481 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6376#-NuVqUblC0okQ0o-IRMD:-NuVqUblC0okQ0o-IRME:bo77dwe) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/ce9b500aa3d93f477eaac2111400e2cd05cca541/src/BloomExe/Collection/CollectionSettings.cs#L481)):*
> ```Smalltalk
>                         {
>                             // We make an exception when the collection is a single book downloaded for editing,
>                             // since we want to keep its branding. (The user can't add books to this collection without
> ```

How about
`the user can't add books to this collection without`  --> `The original uploader had a valid subscription when first uploaded, and we want to allow re-uploading for editing even if this user doesn't know that code or the subscription is expired`
___
*[`src/BloomExe/Collection/CollectionSettings.cs` line 483 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6376#-NuVqzSyDSaAycfon_bg:-NuVqzSyDSaAycfon_bh:b2q88pp) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/ce9b500aa3d93f477eaac2111400e2cd05cca541/src/BloomExe/Collection/CollectionSettings.cs#L483)):*
> ```Smalltalk
>                             // since we want to keep its branding. (The user can't add books to this collection without
>                             // first providing a valid branding.)
>                             _overrideBrandingForEditDownload = BrandingProjectKey;
> ```

Shall we make it parallel, like `_brandingProjectKeyOverrideForDownloadForEditing` ?
___
*[`src/BloomExe/Collection/CollectionSettingsDialog.cs` line 402 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6376#-NuVs3PT4vXe6ggK5sNV:-NuVs3PT4vXe6ggK5sNW:b-y1zr6) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/ce9b500aa3d93f477eaac2111400e2cd05cca541/src/BloomExe/Collection/CollectionSettingsDialog.cs#L402)):*
> ```Smalltalk
>                 _collectionSettings.BrandingProjectKey = _brand;
>                 _collectionSettings.SubscriptionCode = _subscriptionCode;
>                 // We've definitely selected some branding, even if it's the default. So we no longer
> ```

Sorry I can't tell why we're deleting something here. Please comment.
___
*[`src/BloomExe/Publish/BloomLibrary/BloomLibraryPublishModel.cs` line 802 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6376#-NuVsdnQCW3mD-N-K3w0:-NuVsdnQCW3mD-N-K3w1:b-sog4f8) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/ce9b500aa3d93f477eaac2111400e2cd05cca541/src/BloomExe/Publish/BloomLibrary/BloomLibraryPublishModel.cs#L802)):*
> ```Smalltalk
>             var existingThumbUrl = GetBloomLibraryThumbnailUrl(existingBookInfo);
> 
>             // We could get it from the data about the download-for-edit book, but why limit this to those books?
> ```

This reads like a // Review comment. If it's just an rhetorical question... let's change it to not be.
___
*[`src/BloomExe/Workspace/WorkspaceView.cs` line 957 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6376#-NuVtRJ-9LrmQJA6axlw:-NuVtRJ-9LrmQJA6axlx:b-1rtan7) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/ce9b500aa3d93f477eaac2111400e2cd05cca541/src/BloomExe/Workspace/WorkspaceView.cs#L957)):*
> ```Smalltalk
>                         var temp = dlg.ShowDialog(this);
>                         _currentlyOpenSettingsDialog = null;
>                         CollectionSettingsApi.EndFixEnterpriseBranding();
> ```

I'm uncomfortable with this kind of stateful API usage, but I don't know enough to suggest something different if you can't think of a different way.

JohnThomson (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6376)* status: 9 of 17 files reviewed, 10 unresolved discussions (waiting on @hatton)

___
*[`src/BloomBrowserUI/collection/enterpriseSettings.tsx` line 136 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6376#-NuVnyRpAORSVZ7YuVLp:-NuZ_oQ12dqCNXtCNUSD:b-eycc5w) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/ce9b500aa3d93f477eaac2111400e2cd05cca541/src/BloomBrowserUI/collection/enterpriseSettings.tsx#L136)):*

Should we change the id so that localizations get the new text?
</blockquote></details>

As far as I can discover, we forgot to add this string to our XLF, so it can't ever have been localized. And if it now does (I deliberately didn't put translate="no") earlier Blooms will also be better of with the new wording, even in English.
___
*[`src/BloomBrowserUI/react_components/requiresBloomEnterprise.tsx` line 40 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6376#-NuVoq3cCDhot0NSIvgQ:-NuZdEa29U1dK4Sjqpeo:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/ce9b500aa3d93f477eaac2111400e2cd05cca541/src/BloomBrowserUI/react_components/requiresBloomEnterprise.tsx#L40)):*

Probably if I thought harder it would occur to me eventually, but can you add a comment explaining why this parameter is needed here? It's not obvious why an API asking if BE is available should "fail" if we are locked to one book.
</blockquote></details>

Done.
___
*[`src/BloomBrowserUI/react_components/requiresBloomEnterprise.tsx` line 188 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6376#-NuVpFOk39A1TNjQlJlh:-NuZeVs473IlY7Rvg7T6:bflzv3n) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/ce9b500aa3d93f477eaac2111400e2cd05cca541/src/BloomBrowserUI/react_components/requiresBloomEnterprise.tsx#L188)):*

Pls comment on this param too
</blockquote></details>

Done. (I also renamed it.)
___
*[`src/BloomExe/Collection/CollectionSettings.cs` line 393 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6376#-NuVpeu90zZHUe9Pi25S:-NuZevx_APyt2lNhPMc5:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/ce9b500aa3d93f477eaac2111400e2cd05cca541/src/BloomExe/Collection/CollectionSettings.cs#L393)):*

To make this a bit clearer to my future self, let's change "a book downloaded for editing" to "a book downloaded using the 'Download for Editing' button on blorg".
</blockquote></details>

Done.
___
*[`src/BloomExe/Collection/CollectionSettings.cs` line 397 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6376#-NuVq6b-6PCvje0I5S6R:-NuZf1c7Cdtob7O5E41r:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/ce9b500aa3d93f477eaac2111400e2cd05cca541/src/BloomExe/Collection/CollectionSettings.cs#L397)):*

`LoadBranding` implied to me a lot more than this does. How about `ReadBrandingNameFromCollectionFile()`
</blockquote></details>

Done.
___
*[`src/BloomExe/Collection/CollectionSettings.cs` line 481 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6376#-NuVqUblC0okQ0o-IRMD:-NuZftNM0MuLND77KJ_6:buaz6rm) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/ce9b500aa3d93f477eaac2111400e2cd05cca541/src/BloomExe/Collection/CollectionSettings.cs#L481)):*

How about
`the user can't add books to this collection without`  --> `The original uploader had a valid subscription when first uploaded, and we want to allow re-uploading for editing even if this user doesn't know that code or the subscription is expired`
</blockquote></details>

Done. (Actually I added that and kept a modified version of mine, since it was attempting to make a slightly different point.)
___
*[`src/BloomExe/Collection/CollectionSettings.cs` line 483 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6376#-NuVqzSyDSaAycfon_bg:-NuZgH2h3rQPgUsZedaf:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/ce9b500aa3d93f477eaac2111400e2cd05cca541/src/BloomExe/Collection/CollectionSettings.cs#L483)):*

Shall we make it parallel, like `_brandingProjectKeyOverrideForDownloadForEditing` ?
</blockquote></details>

Done.
___
*[`src/BloomExe/Collection/CollectionSettingsDialog.cs` line 402 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6376#-NuVs3PT4vXe6ggK5sNV:-NuZhvHe559D8vONoEo4:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/ce9b500aa3d93f477eaac2111400e2cd05cca541/src/BloomExe/Collection/CollectionSettingsDialog.cs#L402)):*

Sorry I can't tell why we're deleting something here. Please comment.
</blockquote></details>

Done.
___
*[`src/BloomExe/Publish/BloomLibrary/BloomLibraryPublishModel.cs` line 802 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6376#-NuVsdnQCW3mD-N-K3w0:-NuZiDZpEM9ZxzsFek4E:b-hnd1br) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/ce9b500aa3d93f477eaac2111400e2cd05cca541/src/BloomExe/Publish/BloomLibrary/BloomLibraryPublishModel.cs#L802)):*

This reads like a // Review comment. If it's just an rhetorical question... let's change it to not be.
</blockquote></details>

I suppose in a way it was. I don't think you've ever confirmed my intuition that the "you are changing the branding" warning should be shown always, not just when re-uploading a "Download for editing" book. Assuming that is right, I don't think the comment is useful at all, so I deleted it.
___
*[`src/BloomExe/Workspace/WorkspaceView.cs` line 957 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6376#-NuVtRJ-9LrmQJA6axlw:-NuZlDr6Bn1ouwI8sQh3:b4xvnef) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/ce9b500aa3d93f477eaac2111400e2cd05cca541/src/BloomExe/Workspace/WorkspaceView.cs#L957)):*

I'm uncomfortable with this kind of stateful API usage, but I don't know enough to suggest something different if you can't think of a different way.
</blockquote></details>

I don't like it much either. Note, it is not new: we already had the ability to bring up the settings dialog in a special state where the enterprise tab is open and the key field is used to show the name of the branding this collection previously used, but lacks a current code for. What I basically did was to make that information show any time that tab is opened and the collection is in a state where it knows about a branding but doesn't have the corresponding code, not just when we first open the collection and find an invalid branding code.
I _think_ all the relevant logic could be moved out of the CollectionSettingsDialog and WorkspaceView into the CollectionSettingsApi, which could get its data from the collectionSettings rather than being told it independently. Then we might need to pass a param to the dialog to tell it when to open at the enterprise tab. If you want to do all that (I estimate a half day; I'm no longer familiar with the code if I ever was), I'd suggest a separate card and PR, especially since you're in a hurry to get this PR into testing.

hatton (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6376)* status: 9 of 17 files reviewed, 11 unresolved discussions (waiting on @JohnThomson)

___
*[`src/BloomBrowserUI/react_components/requiresBloomEnterprise.tsx` line 40 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6376#-NuVoq3cCDhot0NSIvgQ:-Nu_mwPA2mPa0ZCic9dt:b-3ishct) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/ce9b500aa3d93f477eaac2111400e2cd05cca541/src/BloomBrowserUI/react_components/requiresBloomEnterprise.tsx#L40)):*

Done.
</blockquote></details>

Two problems. 1) It is not "failing", it's just returning false. 2) it seems kludgy. How about a different hook & API that tells you if it is the one locked book?
___
*[`src/BloomExe/web/controllers/CollectionSettingsApi.cs` line 81 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6376#-Nu_mdQ79e4RQ136sfN5:-Nu_mdQ79e4RQ136sfN6:biwblbc) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/108f54176d332a8095baa071accd98e28a098d44/src/BloomExe/web/controllers/CollectionSettingsApi.cs#L81)):*
> ```Smalltalk
>         private bool IsEnterpriseEnabled(bool failIfLockedToOneBook)
>         {
>             if (failIfLockedToOneBook && _collectionSettings.LockedToOneDownloadedBook)
> ```

This isn't "failing", it's just return false.

hatton (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6376)* status: 9 of 17 files reviewed, 12 unresolved discussions (waiting on @JohnThomson)

___
*[`src/BloomBrowserUI/react_components/requiresBloomEnterprise.tsx` line 244 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6376#-Nu_nP6IDql6up2-sApt:-Nu_nP6IDql6up2-sApu:bxfz4wu) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/108f54176d332a8095baa071accd98e28a098d44/src/BloomBrowserUI/react_components/requiresBloomEnterprise.tsx#L244)):*
> ```XML
>                         <RequiresBloomEnterpriseNotice
>                             darkTheme={true}
>                             failIfLockedToOneBook={props.showIfLockedToOneBook}
> ```

again, no "fail" happening here

hatton (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6376)* status: 9 of 17 files reviewed, 6 unresolved discussions (waiting on @JohnThomson)

___
*[`src/BloomExe/Publish/BloomLibrary/BloomLibraryPublishModel.cs` line 802 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6376#-NuVsdnQCW3mD-N-K3w0:-Nu_oPTP9h9wsQ4_faos:bfccy20) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/ce9b500aa3d93f477eaac2111400e2cd05cca541/src/BloomExe/Publish/BloomLibrary/BloomLibraryPublishModel.cs#L802)):*
> that the "you are changing the branding" warning should be shown always, not just when re-uploading a "Download for editing" book

OK

hatton (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6376)* status: 9 of 17 files reviewed, 4 unresolved discussions (waiting on @JohnThomson)

___
*[`src/BloomBrowserUI/react_components/requiresBloomEnterprise.tsx` line 40 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6376#-NuVoq3cCDhot0NSIvgQ:-Nujne4D9KMTyXCkjcUw:b-bnc94v) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/ce9b500aa3d93f477eaac2111400e2cd05cca541/src/BloomBrowserUI/react_components/requiresBloomEnterprise.tsx#L40)):*

Two problems. 1) It is not "failing", it's just returning false. 2) it seems kludgy. How about a different hook & API that tells you if it is the one locked book?
</blockquote></details>

We discussed this and decided that the best solution is to just not the TC settings in this context.
___
*[`src/BloomExe/Workspace/WorkspaceView.cs` line 957 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6376#-NuVtRJ-9LrmQJA6axlw:-NujnOUY1CnAbh2SxadX:b-ha1c6y) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/ce9b500aa3d93f477eaac2111400e2cd05cca541/src/BloomExe/Workspace/WorkspaceView.cs#L957)):*

I don't like it much either. Note, it is not new: we already had the ability to bring up the settings dialog in a special state where the enterprise tab is open and the key field is used to show the name of the branding this collection previously used, but lacks a current code for. What I basically did was to make that information show any time that tab is opened and the collection is in a state where it knows about a branding but doesn't have the corresponding code, not just when we first open the collection and find an invalid branding code.
I _think_ all the relevant logic could be moved out of the CollectionSettingsDialog and WorkspaceView into the CollectionSettingsApi, which could get its data from the collectionSettings rather than being told it independently. Then we might need to pass a param to the dialog to tell it when to open at the enterprise tab. If you want to do all that (I estimate a half day; I'm no longer familiar with the code if I ever was), I'd suggest a separate card and PR, especially since you're in a hurry to get this PR into testing.
</blockquote></details>

We discussed this and decided that the key is that all this wants to be within the encapsulation of CollectionsSettings (or at least CollectionsSettingsApi)

JohnThomson (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6376)* status: 9 of 17 files reviewed, 4 unresolved discussions (waiting on @hatton)

___
*[`src/BloomExe/Publish/BloomLibrary/BloomLibraryPublishModel.cs` line 802 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6376#-NuVsdnQCW3mD-N-K3w0:-Nu_uySw1BV0ydAxrwld:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/ce9b500aa3d93f477eaac2111400e2cd05cca541/src/BloomExe/Publish/BloomLibrary/BloomLibraryPublishModel.cs#L802)):*

> that the "you are changing the branding" warning should be shown always, not just when re-uploading a "Download for editing" book

OK
</blockquote></details>

Done.
___
*[`src/BloomExe/Workspace/WorkspaceView.cs` line 957 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6376#-NuVtRJ-9LrmQJA6axlw:-NujvengAhkIcROmbuwO:bfr4h66) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/ce9b500aa3d93f477eaac2111400e2cd05cca541/src/BloomExe/Workspace/WorkspaceView.cs#L957)):*

We discussed this and decided that the key is that all this wants to be within the encapsulation of CollectionsSettings (or at least CollectionsSettingsApi)
</blockquote></details>

I think I've accomplished what you want. There's still some state, but it's encapsulated in the CollectionsSettingsApi. We could possibly move a little more of the logic for launching the dialog into its own class, but it makes at least some sense for the Workspace to keep track of whether this dialog is open, I think.
___
*[`src/BloomExe/web/controllers/CollectionSettingsApi.cs` line 81 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6376#-Nu_mdQ79e4RQ136sfN5:-NujzyW86W3rpwsu38SP:bwqrvd2) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/108f54176d332a8095baa071accd98e28a098d44/src/BloomExe/web/controllers/CollectionSettingsApi.cs#L81)):*

This isn't "failing", it's just return false.
</blockquote></details>

Got rid of it.
___
*[`src/BloomBrowserUI/react_components/requiresBloomEnterprise.tsx` line 40 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6376#-NuVoq3cCDhot0NSIvgQ:-Nu_vFpw48Yo2nLHw7UQ:bwqrvd2) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/ce9b500aa3d93f477eaac2111400e2cd05cca541/src/BloomBrowserUI/react_components/requiresBloomEnterprise.tsx#L40)):*

We discussed this and decided that the best solution is to just not the TC settings in this context.
</blockquote></details>

Got rid of it.
___
*[`src/BloomBrowserUI/react_components/requiresBloomEnterprise.tsx` line 244 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6376#-Nu_nP6IDql6up2-sApt:-Nuk-7wG3UqB3BQ1QXa4:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/108f54176d332a8095baa071accd98e28a098d44/src/BloomBrowserUI/react_components/requiresBloomEnterprise.tsx#L244)):*

again, no "fail" happening here
</blockquote></details>

Done.

hatton (approved):
> Reviewed 3 of 6 files at r2, 4 of 5 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6376)* status: 16 of 17 files reviewed, 1 unresolved discussion (waiting on @JohnThomson)

___
*[`src/BloomExe/Workspace/WorkspaceView.cs` line 957 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6376#-NuVtRJ-9LrmQJA6axlw:-NuzpawX8sn6KaoERnJM:b-jmjr7y) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/ce9b500aa3d93f477eaac2111400e2cd05cca541/src/BloomExe/Workspace/WorkspaceView.cs#L957)):*

I think I've accomplished what you want. There's still some state, but it's encapsulated in the CollectionsSettingsApi. We could possibly move a little more of the logic for launching the dialog into its own class, but it makes at least some sense for the Workspace to keep track of whether this dialog is open, I think.
</blockquote></details>

Feels to me like the whole thing is a failure of encapsulation. I'll let it go now.

hatton (commented):
> Reviewed 1 of 5 files at r3.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6376)* status: all files reviewed, 1 unresolved discussion (waiting on @JohnThomson)

PR #6381, opened 2024-04-02
Make languages for cover title ignore language combo (BL-13198)

JohnThomson:
Make languages for cover title ignore language combo (BL-13198)

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6381)

### DIFF
--- a/DistFiles/localization/en/Bloom.xlf
+++ b/DistFiles/localization/en/Bloom.xlf
@@ -3719,7 +3719,7 @@
         <note>ID: PublishTab.BodyOnly-description</note>
       </trans-unit>
       <trans-unit id="PublishTab.bloomPUBButton">
-        <source xml:lang="en">BloomPUB</source>
+        <source xml:lang="en">eBooks</source>
         <note>ID: PublishTab.bloomPUBButton</note>
         <note>This is the label under the BloomPUB icon on the left side of the Publish screen.</note>
       </trans-unit>
--- a/DistFiles/localization/en/BloomMediumPriority.xlf
+++ b/DistFiles/localization/en/BloomMediumPriority.xlf
@@ -393,6 +393,10 @@
         <source xml:lang="en">What to Show on Cover</source>
         <note>BookSettings.WhatToShowOnCover</note>
       </trans-unit>
+      <trans-unit id="BookSettings.ShowContentLanguage" sil:dynamic="true" translate="no">
+        <source xml:lang="en">Show {0}</source>
+        <note>BookSettings.ShowContentLanguage</note>
+      </trans-unit>
       <trans-unit id="BookSettings.ShowWrittenLanguageTitle" sil:dynamic="true" translate="no">
         <source xml:lang="en">Show Title in {0}</source>
         <note>BookSettings.ShowWrittenLanguageTitle</note>
--- a/DistFiles/localization/en/BloomMediumPriority.xlf
+++ b/DistFiles/localization/en/BloomMediumPriority.xlf
@@ -393,13 +393,9 @@
         <source xml:lang="en">What to Show on Cover</source>
         <note>BookSettings.WhatToShowOnCover</note>
       </trans-unit>
-      <trans-unit id="BookSettings.ShowWrittenLanguage2Title" sil:dynamic="true" translate="no">
-        <source xml:lang="en">Show Written Language 2 Title</source>
-        <note>BookSettings.ShowWrittenLanguage2Title</note>
-      </trans-unit>
-      <trans-unit id="BookSettings.ShowWrittenLanguage3Title" sil:dynamic="true" translate="no">
-        <source xml:lang="en">Show Written Language 3 Title</source>
-        <note>BookSettings.ShowWrittenLanguage3Title</note>
+      <trans-unit id="BookSettings.ShowWrittenLanguageTitle" sil:dynamic="true" translate="no">
+        <source xml:lang="en">Show Title in {0}</source>
+        <note>BookSettings.ShowWrittenLanguageTitle</note>
       </trans-unit>
       <trans-unit id="BookSettings.ShowLanguageName" sil:dynamic="true" translate="no">
         <source xml:lang="en">Show Language Name</source>
--- a/DistFiles/localization/en/BloomMediumPriority.xlf
+++ b/DistFiles/localization/en/BloomMediumPriority.xlf
@@ -369,13 +369,17 @@
         <note>BookSettings.eBook.Image.MaxResolution.Directions</note>
         <note>displayed below the slider that is used to set the image resolution</note>
       </trans-unit>
-      <trans-unit id="BookSettings.AppearanceGroupLabel" sil:dynamic="true" translate="no">
-        <source xml:lang="en">Appearance</source>
-        <note>BookSettings.AppearanceGroupLabel</note>
+      <trans-unit id="BookSettings.CoverGroupLabel" sil:dynamic="true" translate="no">
+        <source xml:lang="en">Cover</source>
+        <note>BookSettings.CoverGroupLabel</note>
       </trans-unit>
-      <trans-unit id="BookSettings.InsidePagesGroupLabel" sil:dynamic="true" translate="no">
-        <source xml:lang="en">Inside Pages</source>
-        <note>BookSettings.InsidePagesGroupLabel</note>
+      <trans-unit id="BookSettings.ContentPagesGroupLabel" sil:dynamic="true" translate="no">
+        <source xml:lang="en">Content Pages</source>
+        <note>BookSettings.ContentPagesGroupLabel</note>
+      </trans-unit>
+      <trans-unit id="BookSettings.NormalTextBoxLangsLabel" sil:dynamic="true" translate="no">
+        <source xml:lang="en">Languages to show in normal text boxes</source>
+        <note>BookSettings.NormalTextBoxLangsLabel</note>
       </trans-unit>
       <trans-unit id="BookSettings.PageThemeLabel" sil:dynamic="true" translate="no">
         <source xml:lang="en">Page Theme</source>
@@ -390,7 +394,7 @@
         <note>BookSettings.CoverBackground</note>
       </trans-unit>
       <trans-unit id="BookSettings.WhatToShowOnCover" sil:dynamic="true" translate="no">
-        <source xml:lang="en">What to Show on Cover</source>
+        <source xml:lang="en">Front Cover</source>
         <note>BookSettings.WhatToShowOnCover</note>
       </trans-unit>
       <trans-unit id="BookSettings.ShowContentLanguage" sil:dynamic="true" translate="no">
@@ -462,7 +466,7 @@
         <note>{0} is a placeholder for a filename. The text inside the [square brackets] will become a link to a website.</note>
       </trans-unit>
       <trans-unit id="BookSettings.ThemeDisablesOptionsNotice" sil:dynamic="true" translate="no">
-        <source xml:lang="en">The selected page theme does not support Appearance settings.</source>
+        <source xml:lang="en">The selected page theme does not support the following settings.</source>
         <note>BookSettings.ThemeDisablesOptionsNotice</note>
       </trans-unit>
       <trans-unit id="AppearanceTheme.default" sil:dynamic="true" translate="no">
--- a/src/BloomBrowserUI/bookEdit/bookSettings/BookSettingsDialog.tsx
+++ b/src/BloomBrowserUI/bookEdit/bookSettings/BookSettingsDialog.tsx
@@ -123,13 +123,14 @@ export const BookSettingsDialog: React.FunctionComponent<{}> = () => {
         overrideInformation?.brandingName
     );
 
-    const appearanceLabel = useL10n(
-        "Appearance",
-        "BookSettings.AppearanceGroupLabel"
+    const coverLabel = useL10n("Cover", "BookSettings.CoverGroupLabel");
+    const contentPagesLabel = useL10n(
+        "Content Pages",
+        "BookSettings.ContentPagesGroupLabel"
     );
-    const insidePagesSubgroupLabel = useL10n(
-        "Inside Pages",
-        "BookSettings.InsidePagesGroupLabel",
+    const languagesToShowNormalSubgroupLabel = useL10n(
+        "Languages to show in normal text boxes",
+        "BookSettings.NormalTextBoxLangsLabel",
         ""
     );
     const themeLabel = useL10n("Page Theme", "BookSettings.PageThemeLabel", "");
@@ -160,7 +161,7 @@ export const BookSettingsDialog: React.FunctionComponent<{}> = () => {
     );
     */
     const whatToShowOnCoverLabel = useL10n(
-        "What to Show on Cover",
+        "Front Cover",
         "BookSettings.WhatToShowOnCover"
     );
 
@@ -186,15 +187,13 @@ export const BookSettingsDialog: React.FunctionComponent<{}> = () => {
         "BookSettings.FrontAndBackMatter.Description"
     );
     const resolutionLabel = useL10n("Resolution", "BookSettings.Resolution");
-    const bloomPubLabel = useL10n("BloomPUB", "PublishTab.bloomPUBButton"); // reuse the same string localized for the Publish tab
+    const bloomPubLabel = useL10n("eBooks", "PublishTab.bloomPUBButton"); // reuse the same string localized for the Publish tab
 
     // This is a helper function to make it easier to pass the override information
     function getAdditionalProps<T>(
-        subPath: string,
-        disable?: boolean
+        subPath: string
     ): {
         path: string;
-        disabled?: boolean;
         overrideValue: T;
         overrideDescription?: string;
     } {
@@ -214,7 +213,6 @@ export const BookSettingsDialog: React.FunctionComponent<{}> = () => {
         // make a an object that can be spread as props in any of the Configr controls
         return {
             path: "appearance." + subPath,
-            disabled: appearanceDisabled || disable,
             overrideValue: override as T,
             // if we're disabling all appearance controls (e.g. because we're in legacy), don't list a second reason for this overload
             overrideDescription: appearanceDisabled ? "" : description
@@ -355,51 +353,7 @@ export const BookSettingsDialog: React.FunctionComponent<{}> = () => {
                             //setSettings(s);
                         }}
                     >
-                        <ConfigrGroup label={appearanceLabel} level={1}>
-                            {appearanceDisabled && (
-                                <NoteBox>
-                                    <Div l10nKey="BookSettings.ThemeDisablesOptionsNotice">
-                                        The selected page theme does not support
-                                        Appearance settings. They have been
-                                        disabled.
-                                    </Div>
-                                </NoteBox>
-                            )}
-                            <ConfigrSubgroup
-                                label={insidePagesSubgroupLabel}
-                                path={`appearance`}
-                            >
-                                <ConfigrSelect
-                                    label={themeLabel}
-                                    disabled={false}
-                                    path={`appearance.cssThemeName`}
-                                    options={appearanceUIOptions.themeNames.map(
-                                        x => {
-                                            return {
-                                                label: x.label,
-                                                value: x.value
-                                            };
-                                        }
-                                    )}
-                                    description={themeDescription}
-                                />
-                                <ConfigrBoolean
-                                    label={showPageNumbersLabel}
-                                    {...getAdditionalProps<boolean>(
-                                        `pageNumber-show`
-                                    )}
-                                />
-                                <FieldVisibilityGroup
-                                    field="mlcontent"
-                                    labelFrame="Show {0}"
-                                    labelFrameL10nKey="BookSettings.ShowContentLanguage"
-                                    settings={settings}
-                                    settingsToReturnLater={
-                                        settingsToReturnLater
-                                    }
-                                    getAdditionalProps={getAdditionalProps}
-                                />
-                            </ConfigrSubgroup>
+                        <ConfigrGroup label={coverLabel} level={1}>
                             {
                                 // This group of three possible messages...sometimes none of them shows, so there are four options...
                                 // is very similar to the one in BookInfoIndicator.tsx. If you change one, you may need to change the other.
@@ -538,6 +492,18 @@ export const BookSettingsDialog: React.FunctionComponent<{}> = () => {
                                 label={whatToShowOnCoverLabel}
                                 path={`appearance`}
                             >
+                                {appearanceDisabled && (
+                                    <NoteBox
+                                        css={css`
+                                            margin-left: 20px;
+                                        `}
+                                    >
+                                        <Div l10nKey="BookSettings.ThemeDisablesOptionsNotice">
+                                            The selected page theme does not
+                                            support the following settings.
+                                        </Div>
+                                    </NoteBox>
+                                )}
                                 <FieldVisibilityGroup
                                     field="cover-title"
                                     labelFrame="Show Title in {0}"
@@ -546,17 +512,21 @@ export const BookSettingsDialog: React.FunctionComponent<{}> = () => {
                                     settingsToReturnLater={
                                         settingsToReturnLater
                                     }
+                                    disabled={appearanceDisabled}
+                                    L1MustBeTurnedOn={true}
                                     getAdditionalProps={getAdditionalProps}
                                 />
 
                                 <ConfigrBoolean
                                     label={showLanguageNameLabel}
+                                    disabled={appearanceDisabled}
                                     {...getAdditionalProps<boolean>(
                                         `cover-languageName-show`
                                     )}
                                 />
                                 <ConfigrBoolean
                                     label={showTopicLabel}
+                                    disabled={appearanceDisabled}
                                     {...getAdditionalProps<boolean>(
                                         `cover-topic-show`
                                     )}
@@ -596,6 +566,62 @@ export const BookSettingsDialog: React.FunctionComponent<{}> = () => {
                                     description={frontAndBackMatterDescription}
                                 />
                             </ConfigrSubgroup> */}
+                        </ConfigrGroup>
+                        <ConfigrGroup label={contentPagesLabel} level={1}>
+                            <ConfigrSubgroup label="" path={`appearance`}>
+                                {/* Wrapping these two in a div prevents Config-R from sticking a divider between them */}
+                                <div>
+                                    <ConfigrSelect
+                                        label={themeLabel}
+                                        disabled={false}
+                                        path={`appearance.cssThemeName`}
+                                        options={appearanceUIOptions.themeNames.map(
+                                            x => {
+                                                return {
+                                                    label: x.label,
+                                                    value: x.value
+                                                };
+                                            }
+                                        )}
+                                        description={themeDescription}
+                                    />
+                                    {appearanceDisabled && (
+                                        <NoteBox
+                                            css={css`
+                                                margin-left: 20px;
+                                            `}
+                                        >
+                                            <Div l10nKey="BookSettings.ThemeDisablesOptionsNotice">
+                                                The selected page theme does not
+                                                support the following settings.
+                                            </Div>
+                                        </NoteBox>
+                                    )}
+                                </div>
+                                <ConfigrBoolean
+                                    label={showPageNumbersLabel}
+                                    disabled={appearanceDisabled}
+                                    {...getAdditionalProps<boolean>(
+                                        `pageNumber-show`
+                                    )}
+                                />
+                            </ConfigrSubgroup>
+                            <ConfigrSubgroup
+                                label={languagesToShowNormalSubgroupLabel}
+                                path={`appearance`}
+                            >
+                                <FieldVisibilityGroup
+                                    field="autoTextBox"
+                                    labelFrame="Show {0}"
+                                    labelFrameL10nKey="BookSettings.ShowContentLanguage"
+                                    settings={settings}
+                                    settingsToReturnLater={
+                                        settingsToReturnLater
+                                    }
+                                    disabled={false}
+                                    getAdditionalProps={getAdditionalProps}
+                                />
+                            </ConfigrSubgroup>
                         </ConfigrGroup>
                         <ConfigrGroup label={bloomPubLabel} level={1}>
                             {/* note that this is used for bloomPUB and ePUB, but we don't have separate settings so we're putting them in bloomPUB and leaving it to c# code to use it for ePUB as well. */}
--- a/src/BloomBrowserUI/bookEdit/bookSettings/BookSettingsDialog.tsx
+++ b/src/BloomBrowserUI/bookEdit/bookSettings/BookSettingsDialog.tsx
@@ -47,6 +47,8 @@ import { Div } from "../../react_components/l10nComponents";
 import { NoteBox, WarningBox } from "../../react_components/boxes";
 import { default as TrashIcon } from "@mui/icons-material/Delete";
 import { PWithLink } from "../../react_components/pWithLink";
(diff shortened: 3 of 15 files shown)


hatton (changes requested):
I'm only seeing the minimum of the UI changes in BL-13198 here.  I'm assuming you want to get this approach verified first?

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6381#-:-NuW9qw68qr6ITtWtGJh:bsssqh6)* status: all files reviewed, 10 unresolved discussions (waiting on @JohnThomson)

___
*[`src/BloomBrowserUI/bookEdit/bookSettings/BookSettingsDialog.tsx` line 305 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6381#-NuW3ubs8zRqPAaPngKO:-NuW3ubs8zRqPAaPngKP:b-w9o5az) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/a45e63a02429025af37d08c25f7860cea8470cca/src/BloomBrowserUI/bookEdit/bookSettings/BookSettingsDialog.tsx#L305)):*
> ```XML
>     }, [settings, settingsToReturnLater]);
> 
>     // This is a bit of a hack to figure out which of the cover title languages are turned on,
> ```

Due to the size and unwieldiness of this and the part where some things are rendered, shall we move this to its own component and file?
___
*[`src/BloomBrowserUI/bookEdit/js/bloomEditing.ts` line 376 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6381#-NuWAqpO6l4iPxHNy4Uh:-NuWAqpO6l4iPxHNy4Ui:bf4u2ju) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/a45e63a02429025af37d08c25f7860cea8470cca/src/BloomBrowserUI/bookEdit/js/bloomEditing.ts#L376)):*
> ```XML
>             // set our font size so that we can use em units when setting padding of the translation group
>             // If visibility is under the control of the appearance system for this field the child we
>             // want has bloom-contentFirst, otherwise, bloom-content1.
> ```

Please add `// At the time of this writing (5.7) this is only used for cover page titles.`
___
*[`src/BloomExe/Book/TranslationGroupManager.cs` line 408 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6381#-NuWB2bI2uWdNaUM_9Cy:-NuWB2bJ-GB0IJuQuOke:boho73s) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/a45e63a02429025af37d08c25f7860cea8470cca/src/BloomExe/Book/TranslationGroupManager.cs#L408)):*
> ```Smalltalk
>             }
> 
>             AddThemeVisibleOrderClass(
> ```

Please add `// At the time of this writing (5.7) this class only affects cover page titles.`
___
*[`src/BloomExe/Book/TranslationGroupManager.cs` line 427 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6381#-NuW86xG9Nwazxqmm7ol:-NuW86xG9Nwazxqmm7om:b-l8sf3q) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/a45e63a02429025af37d08c25f7860cea8470cca/src/BloomExe/Book/TranslationGroupManager.cs#L427)):*
> ```Smalltalk
> 
>         /// <summary>
>         /// If this field's visibility is controlled by the appearance system, add a class indicating its order within the
> ```

"appearance" --> "appearance (introduced in 5.7)"
___
*[`src/BloomExe/Book/TranslationGroupManager.cs` line 428 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6381#-NuW7xVP8ABe4oEaf37u:-NuW7xVP8ABe4oEaf37v:b-sb1zpy) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/a45e63a02429025af37d08c25f7860cea8470cca/src/BloomExe/Book/TranslationGroupManager.cs#L428)):*
> ```Smalltalk
>         /// <summary>
>         /// If this field's visibility is controlled by the appearance system, add a class indicating its order within the
>         /// collection languages that are visible for this field. Unlike in legacy, the order, like the visibility choice,
> ```

"legacy" --> "legacy (pre-5.7 and an optional mode in 5.7)"
___
*[`src/BloomExe/Book/TranslationGroupManager.cs` line 429 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6381#-NuW8MgW-HJcoRJjQ_dN:-NuW8MgW-HJcoRJjQ_dO:b-eu5b35) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/a45e63a02429025af37d08c25f7860cea8470cca/src/BloomExe/Book/TranslationGroupManager.cs#L429)):*
> ```Smalltalk
>         /// If this field's visibility is controlled by the appearance system, add a class indicating its order within the
>         /// collection languages that are visible for this field. Unlike in legacy, the order, like the visibility choice,
>         /// does not depend on what languages are chosen in the language combo to appear in content pages.
> ```

"the language combo" may not always be there, and even in this issue (see mockups) we're echoing in Configr. Let's say 

"does not depend on what languages are chosen ~~in the language combo~~ to appear in content pages."
___
*[`src/BloomExe/Book/TranslationGroupManager.cs` line 444 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6381#-NuW9Ptm6_XKXu6xeNOC:-NuW9Ptm6_XKXu6xeNOD:b-tctpm5) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/a45e63a02429025af37d08c25f7860cea8470cca/src/BloomExe/Book/TranslationGroupManager.cs#L444)):*
> ```Smalltalk
>         {
>             if (legacy)
>                 return; // not in theme mode
> ```

I suppose you decided this is clearer than just not calling this if we are in legacy. Just checking.
___
*[`src/content/templates/xMatter/bloom-xmatter-sharedRules.less` line 25 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6381#-NuW5B3m5dXsxYCj1fpT:-NuW5B3m5dXsxYCj1fpU:bqnabqh) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/a45e63a02429025af37d08c25f7860cea8470cca/src/content/templates/xMatter/bloom-xmatter-sharedRules.less#L25)):*
> ```Less
>             min-height: auto; // elsewhere there is a min-height, that may or may not be a good idea, but for sure we don't want it here.
>             // This group is tricky. Because it's in xmatter, and we don't have distinct xmatter rules for legacy, it needs
>             // to produce the expected behavior for both. The first two rules are the legacy behavior. When the theme applies,
> ```

"when the __legacy__ theme __is applied__"...
___
*[`src/content/templates/xMatter/bloom-xmatter-sharedRules.less` line 47 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6381#-NuW5iOCA6rXQTVOB7zs:-NuW5iOCA6rXQTVOB7zt:b-wt6y4c) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/a45e63a02429025af37d08c25f7860cea8470cca/src/content/templates/xMatter/bloom-xmatter-sharedRules.less#L47)):*
> ```Less
>             }
>             &.bloom-contentNational2 {
>                 font-size: @DefaultCoverTitle3FontSise;
> ```

`Sise` --> `Size`
___
*[`src/content/templates/xMatter/bloom-xmatter-sharedRules.less` line 76 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6381#-NuW52c92TEkSr3_dfJA:-NuW52c92TEkSr3_dfJB:b-os41ps) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/a45e63a02429025af37d08c25f7860cea8470cca/src/content/templates/xMatter/bloom-xmatter-sharedRules.less#L76)):*
> ```Less
>             // (if the user selected collection L2 as book L1). If they do this and also turn on all three languages for
>             // the title, then the third language will get bloom-contentNational1 as well as bloom-contentThird.
>             // In this case, we don't want
> ```

Was there more to this comment?

JohnThomson (commented):
No. The screenshots were described as "Some mockups towards a solution". I did not realize we had decided to go ahead with any UI changes except adding L1 and rewording the labels. That said, it was probably a good idea to verify the main approach first. If you want to do anything to call attention to changes I might miss, it would be welcome.
I'll deal with that and the rest of your comments after working on BL-13110 since you said that was urgent.

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6381#-:-NuZZKCy4ncCpO-7DmFs:b793c9p)* status: all files reviewed, 10 unresolved discussions (waiting on @JohnThomson)

JohnThomson (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6381)* status: all files reviewed, 10 unresolved discussions (waiting on @hatton)

___
*[`src/BloomBrowserUI/bookEdit/bookSettings/BookSettingsDialog.tsx` line 305 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6381#-NuW3ubs8zRqPAaPngKO:-NudvAxmDoi_5XA6E1cd:b-in580o) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/a45e63a02429025af37d08c25f7860cea8470cca/src/BloomBrowserUI/bookEdit/bookSettings/BookSettingsDialog.tsx#L305)):*

Due to the size and unwieldiness of this and the part where some things are rendered, shall we move this to its own component and file?
</blockquote></details>

Done, though it was difficult and somewhat messy.
___
*[`src/BloomBrowserUI/bookEdit/js/bloomEditing.ts` line 376 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6381#-NuWAqpO6l4iPxHNy4Uh:-NudcjNu5-FyYtUn8x_T:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/a45e63a02429025af37d08c25f7860cea8470cca/src/BloomBrowserUI/bookEdit/js/bloomEditing.ts#L376)):*

Please add `// At the time of this writing (5.7) this is only used for cover page titles.`
</blockquote></details>

Done.
___
*[`src/BloomExe/Book/TranslationGroupManager.cs` line 408 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6381#-NuWB2bI2uWdNaUM_9Cy:-Nudd4JxAdact_UWp2XT:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/a45e63a02429025af37d08c25f7860cea8470cca/src/BloomExe/Book/TranslationGroupManager.cs#L408)):*

Please add `// At the time of this writing (5.7) this class only affects cover page titles.`
</blockquote></details>

Done.
___
*[`src/BloomExe/Book/TranslationGroupManager.cs` line 427 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6381#-NuW86xG9Nwazxqmm7ol:-NuddQVr7fMI3tbJhv63:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/a45e63a02429025af37d08c25f7860cea8470cca/src/BloomExe/Book/TranslationGroupManager.cs#L427)):*

"appearance" --> "appearance (introduced in 5.7)"
</blockquote></details>

Done.
___
*[`src/BloomExe/Book/TranslationGroupManager.cs` line 428 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6381#-NuW7xVP8ABe4oEaf37u:-Nude-TMESkw1pruvmh7:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/a45e63a02429025af37d08c25f7860cea8470cca/src/BloomExe/Book/TranslationGroupManager.cs#L428)):*

"legacy" --> "legacy (pre-5.7 and an optional mode in 5.7)"
</blockquote></details>

Done.
___
*[`src/BloomExe/Book/TranslationGroupManager.cs` line 429 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6381#-NuW8MgW-HJcoRJjQ_dN:-NuddaDY6Dzs5SSPJvyg:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/a45e63a02429025af37d08c25f7860cea8470cca/src/BloomExe/Book/TranslationGroupManager.cs#L429)):*

"the language combo" may not always be there, and even in this issue (see mockups) we're echoing in Configr. Let's say 

"does not depend on what languages are chosen ~~in the language combo~~ to appear in content pages."
</blockquote></details>

Done.
___
*[`src/BloomExe/Book/TranslationGroupManager.cs` line 444 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6381#-NuW9Ptm6_XKXu6xeNOC:-Nude4sR5CsprgiSqGlR:b4qpawy) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/a45e63a02429025af37d08c25f7860cea8470cca/src/BloomExe/Book/TranslationGroupManager.cs#L444)):*

I suppose you decided this is clearer than just not calling this if we are in legacy. Just checking.
</blockquote></details>

Bit of a toss-up. There are several cases (besides legacy) where this method doesn't do anything. That is difficult to reflect in its name (and we often don't in similar cases). I wanted to be able to unit-test as much of the new functionality as possible, so I put all the guards that prevent it doing something when it should not in the method.
___
*[`src/content/templates/xMatter/bloom-xmatter-sharedRules.less` line 25 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6381#-NuW5B3m5dXsxYCj1fpT:-NudfiwNCUWDaKpR55Sv:bkti6a5) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/a45e63a02429025af37d08c25f7860cea8470cca/src/content/templates/xMatter/bloom-xmatter-sharedRules.less#L25)):*

"when the __legacy__ theme __is applied__"...
</blockquote></details>

Clearly needed clarification, because I meant the exact opposite! I ended up rewriting the whole comment and fixing a bug, so this was a useful suggestion indeed!
This is a mess mainly because we have specific margin-bottom values for the first and second titles, but not for the third. I'm not even sure if we deliberately decided that we should have @MarginBetweenMinorItems margin below the V and N1 titles, but not below the N2 one, or whether we just forgot to add it for content-National2. Not many xmatters, if any, call for an N2 title. Nor do I have any idea why this particular gap is still a less variable rather than an appearance system CSS variable.
It would make a good deal of sense, I think, for --cover-title-{bottom}margin to be a new Appearance variable and to apply it to all cover titles in the new world, and then we could just put it in all the new bloom-contentFirst/Second/Third rules and get rid of the :not's.
___
*[`src/content/templates/xMatter/bloom-xmatter-sharedRules.less` line 47 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6381#-NuW5iOCA6rXQTVOB7zs:-NudquY4EnhDtPbVPsV-:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/a45e63a02429025af37d08c25f7860cea8470cca/src/content/templates/xMatter/bloom-xmatter-sharedRules.less#L47)):*

`Sise` --> `Size`
</blockquote></details>

Done.
___
*[`src/content/templates/xMatter/bloom-xmatter-sharedRules.less` line 76 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6381#-NuW52c92TEkSr3_dfJA:-Nudr5Cg3e2XhjvP9Yrk:bcs26v0) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/a45e63a02429025af37d08c25f7860cea8470cca/src/content/templates/xMatter/bloom-xmatter-sharedRules.less#L76)):*

Was there more to this comment?
</blockquote></details>

I think it was left over from earlier in my thought processes. Removed.

hatton (commented):
> Reviewed 5 of 11 files at r2.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6381)* status: 7 of 13 files reviewed, 6 unresolved discussions (waiting on @JohnThomson)

___
*[`src/BloomExe/Book/AppearanceSettings.cs` line 944 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6381#-NujojzjEENtD-QZvZMH:-NujojzjEENtD-QZvZMI:b6rrexo) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/7d28a91875fcff36fcb8d58122edca6fd832b047/src/BloomExe/Book/AppearanceSettings.cs#L944)):*
> ```Smalltalk
> 
>     public ExpandoObject ChangeableSettingsForUI
>     {
> ```

Let's follow the rule that we use `get___` to indicate a method that does something non-trivial.  How about `getCopyOfProperties()`?
___
*[`src/BloomExe/web/controllers/BookSettingsApi.cs` line 145 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6381#-Nujp_m4-wC-Djz4S0gq:-Nujp_m4-wC-Djz4S0gr:b-rhn57q) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/7d28a91875fcff36fcb8d58122edca6fd832b047/src/BloomExe/web/controllers/BookSettingsApi.cs#L145)):*
> ```Smalltalk
>                     }
> 
>                     appearance["mlcontent-L1-show"] = bookLangs.Contains(
> ```

Let's stick with our camel-or-dash convention. Can we improve on "ml"? I can guess that it means "multilingual", but is there some parallel word we can take either from the UI or its implementation?

JohnThomson (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6381)* status: 6 of 13 files reviewed, 6 unresolved discussions (waiting on @hatton)

___
*[`src/BloomExe/Book/AppearanceSettings.cs` line 944 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6381#-NujojzjEENtD-QZvZMH:-Nuy-TeP57x_Y2jk3VFi:b-si4746) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/7d28a91875fcff36fcb8d58122edca6fd832b047/src/BloomExe/Book/AppearanceSettings.cs#L944)):*

Let's follow the rule that we use `get___` to indicate a method that does something non-trivial.  How about `getCopyOfProperties()`?
</blockquote></details>

I'm not convinced that returning a shallow copy of an object constitutes "non-trivial". But it's not worth arguing, so done.
___
*[`src/BloomExe/web/controllers/BookSettingsApi.cs` line 145 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6381#-Nujp_m4-wC-Djz4S0gq:-Nuy4ynQ1x6EhU1ChGKT:beuc4qe) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/7d28a91875fcff36fcb8d58122edca6fd832b047/src/BloomExe/web/controllers/BookSettingsApi.cs#L145)):*

Let's stick with our camel-or-dash convention. Can we improve on "ml"? I can guess that it means "multilingual", but is there some parallel word we can take either from the UI or its implementation?


</blockquote></details>

I had a go.

hatton (changes requested):
> Reviewed 2 of 11 files at r2, 1 of 3 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6381)* status: 9 of 13 files reviewed, 4 unresolved discussions (waiting on @JohnThomson)

___
*[`src/BloomBrowserUI/bookEdit/bookSettings/FieldVisibilityGroup.tsx` line 91 at r4](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6381#-Nuzm3hB6wABeq_ZM4MB:-Nuzm3hB6wABeq_ZM4MC:b5x2ar1) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/4411f98a5515fada0bf70aea1324500757781818/src/BloomBrowserUI/bookEdit/bookSettings/FieldVisibilityGroup.tsx#L91)):*
> ```XML
> 
>     return (
>         <div
> ```

Let's remember that the point of config-r is to lower the cost of adding UI for settings and maintain consistency across all uses of it. The cost we pay is that when we come across something that will not be unique, we need to add it to config-r. In this case, we need to add a property to `<ConfigrBoolean>` to differentiate between values like this that are "locked" vs. those that are completely visually disabled.

hatton (commented):
Almost there. We just need to extract the "locked" vs. "disabled" idea to config-r

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6381#-:-NuznVBF4OomYPgpGrXJ:b1imuuy)* status: 9 of 13 files reviewed, 1 unresolved discussion (waiting on @JohnThomson)

JohnThomson (commented):
I'm not sure what options we actually need or what the right visual effects are. I see at least three kinds
- the primary kind of 'disabled': because of some other choice I made, this option is obviously irrelevant. For example, I said not to show page numbers, so settings for the placement of page numbers are not interesting.
- locked by some unobvious policy. For example, I can't turn off cover title L2 because the xmatter I'm using won't allow it.
- prevented to protect me. For example, it's not meaningless to turn off all three cover titles, but it's very unlikely to be what I really want.
For now, let's call them irrelevant, locked, and protected. I'm not sure a three-way distinction is useful, nor that it is sufficient, but it's a start.
My personal opinion is that the current config-R 'disabled' is too faint for any of them. The grey that I made it show for the language checkbox when only one is checked (38% black) feels to me about right for irrelevant. But I would also think it about right for locked and protected in most cases. Preventing turning off the last of a set of options feels like a special case: we don't want to allow it, but it's weird for the option that is the user's current selection from a list of options to be dimmed.
But nothing else I know has more than one kind of disabled. In this particular case, I think I would like to see the checkbox dimmed but not the text. But what property to add to config-R to achieve that I'm not sure. Indeed, I'm not at all sure that achieving it is worth inventing a locked/disabled distinction. (And I wouldn't describe this case as 'locked'...that's feels more like the xmatter case.)
So, I think my top recommendation would be
- change config-R so it always uses 38% black rather than the current much fainter grey for things that are disabled
- Live with the slightly odd effect of disabling the (one) selected title language.
- But you may have a better idea.

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6381#-:-Nv23vLTB2VbcPeWRozS:b37inp8)* status: 9 of 13 files reviewed, 1 unresolved discussion (waiting on @hatton)

hatton (commented):
Let's add a property to the element, alongside the existing `disabled`, called `locked`. It has the effect of disabled except the label is not greyed.

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6381#-:-Nv2CNL23r_AvbwNUOS5:bmaa6ob)* status: 9 of 13 files reviewed, 1 unresolved discussion (waiting on @JohnThomson)

JohnThomson (commented):
Done

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6381#-:-Nv3nRjCChhVkfhLi1h0:bb6un1j)* status: 9 of 15 files reviewed, 1 unresolved discussion (waiting on @hatton)

___
*[`src/BloomBrowserUI/bookEdit/bookSettings/FieldVisibilityGroup.tsx` line 91 at r4](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6381#-Nuzm3hB6wABeq_ZM4MB:-Nv3nU_OEMSCkbDSNrCe:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/4411f98a5515fada0bf70aea1324500757781818/src/BloomBrowserUI/bookEdit/bookSettings/FieldVisibilityGroup.tsx#L91)):*

Let's remember that the point of config-r is to lower the cost of adding UI for settings and maintain consistency across all uses of it. The cost we pay is that when we come across something that will not be unique, we need to add it to config-r. In this case, we need to add a property to `<ConfigrBoolean>` to differentiate between values like this that are "locked" vs. those that are completely visually disabled.
</blockquote></details>

Done.

hatton (changes requested):
> Reviewed 1 of 3 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6381)* status: 10 of 15 files reviewed, 5 unresolved discussions (waiting on @JohnThomson)

___
*[`src/BloomBrowserUI/bookEdit/bookSettings/FieldVisibilityGroup.tsx` line 93 at r5](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6381#-Nv3pDpr9_fZ1ZGWmP2T:-Nv3pDpr9_fZ1ZGWmP2U:b-ylss75) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/8a60ec19f83a5114d3ba468d2a9b46fd4a292495/src/BloomBrowserUI/bookEdit/bookSettings/FieldVisibilityGroup.tsx#L93)):*
> ```XML
>         <div
>             // The defaultConfig-R styles for disabled rows are not very visible.
>             // There seems to be both a color that makes them about 26% opacity black,
> ```

This comment is left over from before the styling moved to config-r, can be removed.
___
*[`src/BloomBrowserUI/bookEdit/bookSettings/FieldVisibilityGroup.tsx` line 108 at r5](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6381#-Nv3paa02bgeFr0MuMM9:-Nv3paa02bgeFr0MuMMA:b-iltvu4) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/8a60ec19f83a5114d3ba468d2a9b46fd4a292495/src/BloomBrowserUI/bookEdit/bookSettings/FieldVisibilityGroup.tsx#L108)):*
> ```XML
>         >
>             <ConfigrBoolean
>                 label={showWrittenLanguage1TitleLabel}
> ```

instead, put a small comment here like
```
// 'locked' will leave the label readable while making the checkbox disabled
```

___
*[`src/BloomBrowserUI/bookEdit/bookSettings/FieldVisibilityGroup.tsx` line 109 at r5](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6381#-Nv3pbNpBdb_-FZ8sdaK:-Nv3pbNpBdb_-FZ8sdaL:bo9v1t9) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/8a60ec19f83a5114d3ba468d2a9b46fd4a292495/src/BloomBrowserUI/bookEdit/bookSettings/FieldVisibilityGroup.tsx#L109)):*
> ```XML
>             <ConfigrBoolean
>                 label={showWrittenLanguage1TitleLabel}
>                 locked={props.disabled || (showL1 && !showL2 && !showL3)}
> ```

Am I remembering in correctly that we decided that the user does not have the option to turn off showing the L1 title?  In that case, I'd expect to see
`locked={true}`
Also, I think `disabled` can just be `disabled={props.disabled}`?
___
*[`src/BloomBrowserUI/bookEdit/bookSettings/FieldVisibilityGroup.tsx` line 117 at r5](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6381#-Nv3o-avCTcN7E_hx60-:-Nv3o-avCTcN7E_hx600:b-4iw76q) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/8a60ec19f83a5114d3ba468d2a9b46fd4a292495/src/BloomBrowserUI/bookEdit/bookSettings/FieldVisibilityGroup.tsx#L117)):*
> ```XML
>                 <ConfigrBoolean
>                     label={showWrittenLanguage2TitleLabel}
>                     locked={props.disabled || (showL2 && !showL1 && !showL3)}
> ```

I don't see why this would ever be locked?  Instead, I would expect
```
disabled={props.disabled}
```
We do choose to show it only if it we haven't shown it again.
___
*[`src/BloomBrowserUI/bookEdit/bookSettings/FieldVisibilityGroup.tsx` line 130 at r5](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6381#-Nv3qDgi-_X5sO_ChQdK:-Nv3qDgi-_X5sO_ChQdL:b-ya81ng) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/8a60ec19f83a5114d3ba468d2a9b46fd4a292495/src/BloomBrowserUI/bookEdit/bookSettings/FieldVisibilityGroup.tsx#L130)):*
> ```XML
>                     <ConfigrBoolean
>                         label={showWrittenLanguage3TitleLabel}
>                         locked={
> ```

Same as above

JohnThomson (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6381)* status: 9 of 15 files reviewed, 5 unresolved discussions (waiting on @hatton)

___
*[`src/BloomBrowserUI/bookEdit/bookSettings/FieldVisibilityGroup.tsx` line 93 at r5](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6381#-Nv3pDpr9_fZ1ZGWmP2T:-Nv3ulH445D98P6hLj_V:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/8a60ec19f83a5114d3ba468d2a9b46fd4a292495/src/BloomBrowserUI/bookEdit/bookSettings/FieldVisibilityGroup.tsx#L93)):*

This comment is left over from before the styling moved to config-r, can be removed.
</blockquote></details>

Done.
___
*[`src/BloomBrowserUI/bookEdit/bookSettings/FieldVisibilityGroup.tsx` line 108 at r5](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6381#-Nv3paa02bgeFr0MuMM9:-Nv3uiuX6uRqx8Dj1Joi:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/8a60ec19f83a5114d3ba468d2a9b46fd4a292495/src/BloomBrowserUI/bookEdit/bookSettings/FieldVisibilityGroup.tsx#L108)):*

instead, put a small comment here like
```
// 'locked' will leave the label readable while making the checkbox disabled
```

</blockquote></details>

Done.
___
*[`src/BloomBrowserUI/bookEdit/bookSettings/FieldVisibilityGroup.tsx` line 109 at r5](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6381#-Nv3pbNpBdb_-FZ8sdaK:-Nv3xOy5CjH7H93PuKbg:b-s8s8sv) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/8a60ec19f83a5114d3ba468d2a9b46fd4a292495/src/BloomBrowserUI/bookEdit/bookSettings/FieldVisibilityGroup.tsx#L109)):*

Am I remembering in correctly that we decided that the user does not have the option to turn off showing the L1 title?  In that case, I'd expect to see
`locked={true}`
Also, I think `disabled` can just be `disabled={props.disabled}`?
</blockquote></details>

We decided to add a control for whether L1 must be on. And since it doesn't have to be for content pages, we need the more complex logic.
___
*[`src/BloomBrowserUI/bookEdit/bookSettings/FieldVisibilityGroup.tsx` line 117 at r5](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6381#-Nv3o-avCTcN7E_hx60-:-Nv3uIr95qs8oXpMG6Id:b4vkcdp) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/8a60ec19f83a5114d3ba468d2a9b46fd4a292495/src/BloomBrowserUI/bookEdit/bookSettings/FieldVisibilityGroup.tsx#L117)):*

I don't see why this would ever be locked?  Instead, I would expect
```
disabled={props.disabled}
```
We do choose to show it only if it we haven't shown it again.
</blockquote></details>

It's certainly relevant when the control is used for content page language choice. And harmless if L1 is always on, since showL1 will always be true.
___
*[`src/BloomBrowserUI/bookEdit/bookSettings/FieldVisibilityGroup.tsx` line 130 at r5](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6381#-Nv3qDgi-_X5sO_ChQdK:-Nv3xKs753sE70UOA1Al:b5mln06) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/8a60ec19f83a5114d3ba468d2a9b46fd4a292495/src/BloomBrowserUI/bookEdit/bookSettings/FieldVisibilityGroup.tsx#L130)):*

Same as above
</blockquote></details>

Same response

PR #6390, opened 2024-04-04
Fix the check for license l10n to ignore Custom "und" (BL-13228)

StephenMcConnel:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6390)

### DIFF
--- a/src/BloomExe/Book/BookCopyrightAndLicense.cs
+++ b/src/BloomExe/Book/BookCopyrightAndLicense.cs
@@ -158,11 +158,17 @@ namespace Bloom.Book
 				langPriorities,
 				out languageUsedForDescription
 			);
-			LocalizationHelper.CheckForMissingLocalization(
-				langPriorities.ToList(),
-				languageUsedForDescription,
-				"Palaso.xlf"
-			);
+			// CustomLicense returns "und" for the description language unless the description is empty.
+			// For an empty description, it returns the localized form of the boilerplate text
+			// "For permission to reuse, contact the copyright holder." with the appropriate language tag.
+			if (!(languageUsedForDescription == "und" && metadata.License is CustomLicense))
+			{
+				LocalizationHelper.CheckForMissingLocalization(
+					langPriorities.ToList(),
+					languageUsedForDescription,
+					"Palaso.xlf"
+				);
+			}
 			dom.SetBookSetting("licenseDescription", languageUsedForDescription, ConvertNewLinesToHtmlBreaks(description));
 
 			// Book may have old licenseNotes, typically in 'en'. This can certainly show up again if licenseNotes in '*' is removed,


JohnThomson (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6390)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @StephenMcConnel)

___
*[`src/BloomExe/Book/BookCopyrightAndLicense.cs` line 161 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6390#-NuefQJQ9yH5YcWJxbKR:-NuefQJQ9yH5YcWJxbKS:b-vb4c45) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/3c6564da07346c90b2181fad873ef04196c4d9b2/src/BloomExe/Book/BookCopyrightAndLicense.cs#L161)):*
> ```Smalltalk
> 				out languageUsedForDescription
> 			);
> 			if (languageUsedForDescription != "und" || !(metadata.License is CustomLicense))
> ```

Feels to me that this should be "and" rather than "or", if languageUsedForDescription is even relevant. We surely don't expect to have a localization for ANY CustomLicense in ANY language, and it's perfectly plausible (and no error) for any custom license to be in a language that isn't one of our localization ones.
I don't know why a non-custom license would ever claim it's languageUsedForDescription was und...it's true that we don't need to check to know that we don't have a localization for und, but maybe if a non-custom license is trying to be und, we might want to know?
I think I would be inclined to just take out the first condition.

JohnThomson (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6390)* status: all files reviewed, 1 unresolved discussion (waiting on @StephenMcConnel)

StephenMcConnel (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6390)* status: all files reviewed, 1 unresolved discussion (waiting on @JohnThomson)

___
*[`src/BloomExe/Book/BookCopyrightAndLicense.cs` line 161 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6390#-NuefQJQ9yH5YcWJxbKR:-NuesfAS3muxOe6ou8-k:b-6rag3n) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/3c6564da07346c90b2181fad873ef04196c4d9b2/src/BloomExe/Book/BookCopyrightAndLicense.cs#L161)):*

Feels to me that this should be "and" rather than "or", if languageUsedForDescription is even relevant. We surely don't expect to have a localization for ANY CustomLicense in ANY language, and it's perfectly plausible (and no error) for any custom license to be in a language that isn't one of our localization ones.
I don't know why a non-custom license would ever claim it's languageUsedForDescription was und...it's true that we don't need to check to know that we don't have a localization for und, but maybe if a non-custom license is trying to be und, we might want to know?
I think I would be inclined to just take out the first condition.
</blockquote></details>

This is a NAND operation if you look more closely at the boolean object.  :-)
(NOT A OR NOT B) == NOT (A AND B)

I checked further and "und" is the official ISO 639-3 code for "undefined language" for language data.   So I suppose Ken did do the right thing here back when.

JohnThomson (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6390)* status: all files reviewed, 1 unresolved discussion (waiting on @StephenMcConnel)

___
*[`src/BloomExe/Book/BookCopyrightAndLicense.cs` line 161 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6390#-NuefQJQ9yH5YcWJxbKR:-Nuisfh_7zRYxeufnYf_:b-f5b77n) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/3c6564da07346c90b2181fad873ef04196c4d9b2/src/BloomExe/Book/BookCopyrightAndLicense.cs#L161)):*

I changed the code to be a more explicit NAND test, and added a comment explaining what is going on and why the check for "und" needs to remain.
</blockquote></details>

Yes, I got the NAND, just had no idea that "und" was returned only for non-empty custom licenses or that the substitute message for an empty one was something we DID want to check.

I found "it" a bit ambiguous ("it returns the localized form...").
How about something like,
"Check that we have an appropriate localization of the license description, unless it is a truly custom license. We do want to check in the case of an empty custom license, which results in boilerplate text; a really custom one (non-empty text given by the author) is recognized by GetDescription returning the languageUsedForDescription as "und")

JohnThomson (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6390#-:-NuivZtd1M6flEt0PjqO:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on @StephenMcConnel)

PR #6397, opened 2024-04-09
BL-13215 fix background colors and flashing button outline

nabalone:
- Fix dark oval around collection name dropdown menu triangle
- editing tab left pane color consistency
- Fix flashing white outline around "Add Page" button when changing page selection

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6397)

### DIFF
--- a/src/BloomBrowserUI/bookEdit/pageThumbnailList/pageControls/pageControls.less
+++ b/src/BloomBrowserUI/bookEdit/pageThumbnailList/pageControls/pageControls.less
@@ -36,7 +36,7 @@
 }
 
 button {
-    background-color: @bloom-panelBackground;
+    background-color: inherit;
     border: none;
     padding: @ButtonPadding;
     &:hover {
--- a/src/BloomBrowserUI/collectionsTab/CollectionsTabPane.tsx
+++ b/src/BloomBrowserUI/collectionsTab/CollectionsTabPane.tsx
@@ -434,6 +434,8 @@ export const CollectionsTabPane: React.FunctionComponent<{}> = () => {
                                 css={css`
                                     // Use && to get enough specificity to beat out .MuiButtonRoot. Alternatively, you can slap !important on inline-style them
                                     && {
+                                        background-color: inherit;
+
                                         // Triangle positioning.
 
                                         // Match historical 15-16 px of padding between the text and left edge of triangle
--- a/src/BloomExe/Book/AppearanceSettings.cs
+++ b/src/BloomExe/Book/AppearanceSettings.cs
@@ -245,6 +245,8 @@ public class AppearanceSettings
     /// </summary>
     public bool ShouldUseAppearanceCss => !UsingLegacy;
 
+    // All the possible files this might return must be included in BookStorage.AutomaticallyAddedCssFilePrefixes
+    // so that we can delete the ones we don't want
     public List<string> AppearanceRelatedCssFiles(bool useLocalCollectionStyles)
     {
         var result = new List<string>();
@@ -258,27 +260,6 @@ public class AppearanceSettings
             result.Add(BookStorage.RelativePathToCollectionStyles(useLocalCollectionStyles));
         result.Add(BasePageCssName);
         return result;
-    }
-
-    /// <summary>
-    /// List all the CSS files that AppearanceRelatedCssFiles might ever return
-    /// (for use in deleting the ones that it does not currently return).
-    /// </summary>
-    public static string[] PossibleAppearanceRelatedCssFiles
-    {
-        get
-        {
-            return new[]
-            {
-                "appearance.css",
-                "customBookStyles.css",
-                "customBookStyles2.css",
-                "customCollectionStyles.css",
-                "../customCollectionStyles.cs",
-                "basePage.css",
-                "basePage-legacy-5-6.css"
-            };
-        }
     }
 
     /// <summary>
--- a/src/BloomExe/Book/Book.cs
+++ b/src/BloomExe/Book/Book.cs
@@ -513,11 +513,11 @@ namespace Bloom.Book
 
             var pageDom = GetHtmlDomWithJustOnePage(page);
             pageDom.RemoveModeStyleSheets();
-            pageDom.AddStyleSheet(this.BookInfo.AppearanceSettings.BasePageCssName);
-            pageDom.AddStyleSheet("editMode.css");
-            pageDom.AddStyleSheet("editPaneGlobal.css");
-            pageDom.AddStyleSheet("");
-            pageDom.SortStyleSheetLinks();
+            pageDom.EnsureStylesheetLinks(
+                this.BookInfo.AppearanceSettings.BasePageCssName,
+                "editMode.css",
+                "editPaneGlobal.css"
+            );
             AddJavaScriptForEditing(pageDom);
             RuntimeInformationInjector.AddUIDictionaryToDom(pageDom, _bookData, BookInfo);
             RuntimeInformationInjector.AddUISettingsToDom(
@@ -651,7 +651,7 @@ namespace Bloom.Book
             //	importNode.AppendChild(child);
             //inputHead.ParentNode.ReplaceChild(importNode, inputHead);
             var result = Storage.MakeDomRelocatable(inputDom);
-            Storage.EnsureHasLinkToStyleSheet(result, "previewMode.css");
+            result.EnsureStylesheetLinks("previewMode.css");
             return result;
         }
 
@@ -663,15 +663,9 @@ namespace Bloom.Book
             }
             var pageDom = GetHtmlDomWithJustOnePage(page);
             pageDom.RemoveModeStyleSheets();
-            foreach (var cssFileName in this.Storage.GetCssFilesToLinkForPreview())
-            {
-                pageDom.AddStyleSheet(cssFileName);
-            }
+            pageDom.EnsureStylesheetLinks(this.Storage.GetCssFilesToLinkForPreview());
             // Note: it would be a fine enhancement here to first check for "branding-{flavor}.css",
             // but we'll leave that until we need it.
-
-            pageDom.SortStyleSheetLinks();
-
             AddPreviewJavascript(pageDom); //review: this is just for thumbnails... should we be having the javascript run?
             return pageDom;
         }
@@ -689,7 +683,6 @@ namespace Bloom.Book
                 return GetErrorDom();
             }
             var pageDom = GetHtmlDomWithJustOnePage(page);
-            pageDom.SortStyleSheetLinks();
             AddPreviewJavascript(pageDom);
             HtmlDom.AddClass(pageDom.Body, "bloom-templateThumbnail");
             return pageDom;
@@ -770,11 +763,7 @@ namespace Bloom.Book
         {
             var dom = Storage.GetRelocatableCopyOfDom();
             dom.RemoveModeStyleSheets();
-            foreach (var cssFileName in cssFileNames)
-            {
-                dom.AddStyleSheet(cssFileName);
-            }
-            dom.SortStyleSheetLinks();
+            dom.EnsureStylesheetLinks(cssFileNames);
 
             return dom;
         }
@@ -945,7 +934,7 @@ namespace Bloom.Book
             // Earlier code also added origami.css, but I don't know any reason to add it now
             // in the unlikely event it isn't already loaded. We want the preview to look just
             // like the HTML would look if we opened the file.
-            previewDom.AddStyleSheet("previewMode.css");
+            previewDom.EnsureStylesheetLinks("previewMode.css");
 
             // Older code (pre-5.7) called UpdateContentLanguageClasses() here with code that chose the
             // collection L1 language if the book has it, otherwise, the book's own L1. That now results
@@ -2293,7 +2282,11 @@ namespace Bloom.Book
             if (coverColor != null)
             {
                 var coverColorIsDark = ColorUtils.IsDark(coverColor);
-                foreach (var page in bookDOM.SafeSelectNodes("//div[contains(@class,'coverColor')]").Cast<XmlElement>())
+                foreach (
+                    var page in bookDOM
+                        .SafeSelectNodes("//div[contains(@class,'coverColor')]")
+                        .Cast<XmlElement>()
+                )
                 {
                     if (coverColorIsDark)
                         HtmlDom.AddClass(page, "darkCoverColor");
@@ -4127,7 +4120,6 @@ namespace Bloom.Book
             if (IsFolio)
             {
                 AddChildBookContentsToFolio(printingDom, currentBookCollection, bookServer);
-                printingDom.SortStyleSheetLinks();
             }
 
             //we do this now becuase the publish ui allows the user to select a different layout for the pdf than what is in the book file
@@ -4218,7 +4210,6 @@ namespace Bloom.Book
                 //add links to the template css needed by the children.
 
                 HtmlDom.AddStylesheetFromAnotherBook(childBook.OurHtmlDom, printingDom);
-                printingDom.SortStyleSheetLinks();
 
                 foreach (
                     XmlElement pageDiv in childBook.OurHtmlDom.RawDom.SafeSelectNodes(
--- a/src/BloomExe/Book/BookStorage.cs
+++ b/src/BloomExe/Book/BookStorage.cs
@@ -81,8 +81,6 @@ namespace Bloom.Book
         void CleanupUnusedVideoFiles();
         void CleanupUnusedActivities();
 
-        void EnsureHasLinkToStyleSheet(HtmlDom dom, string path);
-
         ExpandoObject XmatterAppearanceSettings { get; }
         ExpandoObject BrandingAppearanceSettings { get; }
 
@@ -2968,26 +2966,14 @@ namespace Bloom.Book
         {
             //clear out any old ones
             dom.RemoveNormalStyleSheetsLinks();
-            EnsureHasLinkToStyleSheet(dom, Path.GetFileName(PathToXMatterStylesheet));
-
-            CssFilesThatAreAlwaysWanted.ForEach(x =>
-            {
-                EnsureHasLinkToStyleSheet(dom, x);
-            });
+            //Stylesheets will all get sorted at the end by EnsureStylesheetLinks
+            dom.EnsureStylesheetLinksWithoutSorting(Path.GetFileName(PathToXMatterStylesheet));
+            dom.EnsureStylesheetLinksWithoutSorting(CssFilesThatAreAlwaysWanted);
             var appearanceRelatedCssFiles = BookInfo.AppearanceSettings.AppearanceRelatedCssFiles(
                 LinkToLocalCollectionStyles
             );
-            appearanceRelatedCssFiles.ForEach(x =>
-            {
-                EnsureHasLinkToStyleSheet(dom, x);
-            });
 
-            // If we're not supposed to have these links, make sure we don't.
-            AppearanceSettings.PossibleAppearanceRelatedCssFiles
-                .Except(appearanceRelatedCssFiles)
-                .ForEach(file => EnsureDoesNotHaveLinkToStyleSheet(dom, file));
-
-            dom.SortStyleSheetLinks();
+            dom.EnsureStylesheetLinks(appearanceRelatedCssFiles.ToArray());
         }
 
         public string HandleRetiredXMatterPacks(HtmlDom dom, string nameOfXMatterPack)
@@ -2998,7 +2984,7 @@ namespace Bloom.Book
             {
                 const string xmatterSuffix = "-XMatter.css";
                 EnsureDoesNotHaveLinkToStyleSheet(dom, nameOfXMatterPack + xmatterSuffix);
-                EnsureHasLinkToStyleSheet(dom, nameOfXMatterPack + xmatterSuffix);
+                dom.EnsureStylesheetLinks(nameOfXMatterPack + xmatterSuffix);
                 // Since HtmlDom.GetMetaValue() is always called with the collection's xmatter pack as default,
                 // we can just remove this wrong meta element.
                 dom.RemoveMetaElement("xmatter");
@@ -3014,24 +3000,6 @@ namespace Bloom.Book
                 if (fileName == path)
                     dom.RemoveStyleSheetIfFound(path);
             }
-        }
-
-        public void EnsureHasLinkToStyleSheet(HtmlDom dom, string path)
-        {
-            foreach (XmlElement link in dom.SafeSelectNodes("//link[@rel='stylesheet']"))
-            {
-                var fileName = link.GetStringAttribute("href");
-                if (fileName == path)
-                    return;
-                // We may also have an obsolete link with a Windows-specific path.
-                if (fileName.Replace('\\', '/') == path)
-                {
-                    // We want a link that will work on both platforms, so correct it.
-                    link.SetAttribute("href", path);
-                    return;
-                }
-            }
-            dom.AddStyleSheetIfMissing(path);
         }
 
         public string[] CssFilesThatAreAlwaysWanted
@@ -3064,8 +3032,7 @@ namespace Bloom.Book
         {
             EnsureHasLinksToStylesheets(dom);
             dom.RemoveModeStyleSheets(); // nb must be before we add previewMode, which it removes
-            dom.AddStyleSheetIfMissing("previewMode.css");
-            dom.SortStyleSheetLinks();
+            dom.EnsureStylesheetLinks("previewMode.css");
             dom.RemoveFileProtocolFromStyleSheetLinks();
         }
 
@@ -3824,18 +3791,23 @@ namespace Bloom.Book
             "editTranslationMode.css"
         };
 
-        public static readonly string[] KnownCssFilePrefixesInOrder =
+        // These go before "unknown" stylesheets in the sort order
+        public static readonly string[] OrderedPrefixesOfCssFilesToSortBeforeUnknownStylesheets =
         {
-            // list in the order that you want their <link> to appear
             "basePage", // we leave off ".css" so that this can match version ones, like "basePage-legacy-5-6.css"
             "baseEPUB.css",
             "editMode.css",
             "previewMode.css",
             "origami.css",
-            "UNKNOWN_STYLESHEETS_HERE",
+        };
+
+        // These go after "unknown" stylesheets in the sort order
+        public static readonly string[] OrderedPrefixesOfCssFilesToSortAfterUnknownStylesheets =
+        {
             "branding.css",
             "defaultLangStyles.css",
             "customCollectionStyles.css",
+            "../customCollectionStyles.css",
             "appearance.css",
             // We don't usually have both of these, and I don't have a clear idea why one should come before
             // the other. But the order should be consistent, and if both are there, typically customBookStyles2.css
@@ -3843,6 +3815,15 @@ namespace Bloom.Book
             "customBookStyles2.css",
             "customBookStyles.css"
         };
+
+        // RemoveNormalStyleSheetsLinks uses this list to get rid of old css files before we add new ones.
+        // This list must include all the CSS files that AppearanceRelatedCssFiles might ever return so we can
+        // delete the ones we don't want
+        public static readonly string[] AutomaticallyAddedCssFilePrefixes =
+        // These are split up for the sake of the StyleSheetLinkSorter
+        OrderedPrefixesOfCssFilesToSortBeforeUnknownStylesheets
+            .Concat(OrderedPrefixesOfCssFilesToSortAfterUnknownStylesheets)
+            .ToArray();
 
         /// <summary>
         /// Relative to the book folder, where should we find the customCollectionStyles.css file?
--- a/src/BloomExe/Book/HtmlDom.cs
+++ b/src/BloomExe/Book/HtmlDom.cs
@@ -251,25 +251,53 @@ namespace Bloom.Book
         /// </summary>
         internal bool UseOriginalImages { get; set; }
 
-        public void AddStyleSheet(string path)
+        /// <summary>
+        /// Make sure the DOM has a link to each of the stylesheets passed, and that all stylesheet links are in the right order.
+        /// Return true if any stylesheets were added
+        /// </summary>
+        public bool EnsureStylesheetLinks(params string[] paths)
+        {
+            bool anyAdded = EnsureStylesheetLinksWithoutSorting(paths);
+            SortStyleSheetLinks();
+            return anyAdded;
+        }
+
+        /// <summary>
+        /// Add a reference if we don't already have it for each of the specified stylesheets in the list of paths.
+        /// Return true if any stylesheets were added
+        /// </summary>
+        public bool EnsureStylesheetLinksWithoutSorting(params string[] paths)
+        {
+            var anyAdded = false;
+            foreach (string path in paths)
+            {
+                anyAdded |= EnsureStylesheetLinkWithoutSorting(path);
+            }
+            return anyAdded;
+        }
+
+        private bool EnsureStylesheetLinkWithoutSorting(string path)
         {
             // This version is in libpalaso, and it does weird things that are file:/// oriented, like looking for the file and giving it file path
             // RawDom.AddStyleSheet(path);
 
+
+            // Remember, Linux filenames are case sensitive.
+            var pathToCheck = path;
+            pathToCheck = pathToCheck.Replace('\\', '/');
+
             // We don't need any link to be there twice. If we're already linked to this stylesheet, don't add it again.
-            // Review: the libpalaso code instead removes the old link before adding the new one.
-            // The code here is more efficient, but it means that if the stylesheet is in the wrong order, or even if the old
-            // link looked for it in the wrong folder, we don't fix it.
-            // Generally we sort the links after adding all of them, so order doesn't matter.
-            // I'm fairly sure we're not depending on this code to fix links to moved stylesheets.
-            // Most of our stylesheets live directly in the book folder.
             foreach (XmlElement linkNode in RawDom.SafeSelectNodes("/html/head/link"))
             {
-                var href = linkNode.GetAttribute("href");
-                if (Path.GetFileName(href) == Path.GetFileName(path))
-                {
-                    return;
-                }
+                var filename = linkNode.GetAttribute("href");
+
+                filename = filename.Replace('\\', '/');
+                if (filename == pathToCheck)
+                    return false;
+                Debug.Assert(
+                    filename.ToLowerInvariant() != pathToCheck.ToLowerInvariant(),
+                    "passing a filename with wrong case: " + filename
+                );
             }
 
             var head = XmlUtils.GetOrCreateElement(RawDom, "//html", "head");
@@ -278,6 +306,7 @@ namespace Bloom.Book
             link.SetAttribute("href", path);
             link.SetAttribute("type", "text/css");
             head.AppendChild(link);
+            return true;
         }
 
         public XmlNodeList SafeSelectNodes(string xpath)
@@ -660,32 +689,6 @@ namespace Bloom.Book
         );
 
         /// <summary>
-        /// Add a reference to the specified stylesheet if we don't already have it.
-        /// Return true if added, and if it's not a known stylesheet that is automatically
-        /// present in a page being edited.
-        /// </summary>
-        /// <param name="path"></param>
-        /// <returns></returns>
-        public bool AddStyleSheetIfMissing(string path)
-        {
-            // Remember, Linux filenames are case sensitive.
-            var pathToCheck = path;
-            if (Environment.OSVersion.Platform == PlatformID.Win32NT)
-                pathToCheck = pathToCheck.ToLowerInvariant();
-            foreach (XmlElement link in _dom.SafeSelectNodes("//link[@rel='stylesheet']"))
-            {
-                var fileName = link.GetStringAttribute("href");
-                if (Environment.OSVersion.Platform == PlatformID.Win32NT)
-                    fileName = fileName.ToLowerInvariant();
-                if (fileName == pathToCheck)
-                    return false;
-            }
-            //_dom.AddStyleSheet(path.Replace("file://", "")); this used the libpalaso version which is slow and looks for the file and such
-            AddStyleSheet(path);
-            return !stylesheetsToIgnoreAdding.Contains(path);
-        }
-
-        /// <summary>
         /// We're trying to get names of style sheets that need to be copied from one book to another.
         /// So not the standard ones that every book has, but things like one specific to the particular
         /// template the source was made from, or that got added becuase we used an Activity page that
@@ -805,7 +808,7 @@ namespace Bloom.Book
                     var name = Path.GetFileName(href).ToLowerInvariant();
                     if (
                         name.EndsWith("xmatter.css")
-                        || BookStorage.KnownCssFilePrefixesInOrder.Any(
+                        || BookStorage.AutomaticallyAddedCssFilePrefixes.Any(
                             prefix => name.StartsWith(prefix.ToLowerInvariant())
                         )
                     )
@@ -2942,23 +2945,12 @@ namespace Bloom.Book
             HtmlDom targetBookDom
         )
         {
-            var addedModifiedStyleSheets = new List<string>();
             //This was refactored from book, where there was these notes:
             //     NB: at this point this code can't handle the "userModifiedStyles" from children, it'll ignore them (they would conflict with each other)
             //     NB: at this point custom styles (e.g. larger/smaller font rules) from children will be lost.
-            bool anyAdded = false;
-
-            //At this point, this addedModifiedStyleSheets is just used as a place to track which stylesheets we already have
-            foreach (string sheetName in sourceBookDom.GetTemplateStyleSheets())
-            {
-                if (!addedModifiedStyleSheets.Contains(sheetName))
-                //nb: if two books have stylesheets with the same name, we'll only be grabbing the 1st one.
-                {
-                    addedModifiedStyleSheets.Add(sheetName);
-                    anyAdded |= targetBookDom.AddStyleSheetIfMissing(sheetName);
-                }
-            }
-            return anyAdded;
+            return targetBookDom.EnsureStylesheetLinks(
+                sourceBookDom.GetTemplateStyleSheets().ToArray()
+            );
         }
 
         public static string ConvertHtmlBreaksToNewLines(string html)
--- a/src/BloomExe/Book/StyleSheetLinkSorter.cs
+++ b/src/BloomExe/Book/StyleSheetLinkSorter.cs
@@ -1,5 +1,6 @@
 using System;
 using System.Collections.Generic;
+using System.Linq;
 using System.Xml;
 using SIL.Linq;
 using SIL.Xml;
@@ -12,6 +13,14 @@ namespace Bloom.Book
     /// </summary>
     public class StyleSheetLinkSorter : IComparer<XmlElement>
     {
+        public static readonly string[] KnownCssFilePrefixesInOrder =
+            BookStorage.OrderedPrefixesOfCssFilesToSortBeforeUnknownStylesheets
+                .Append("UNKNOWN_STYLESHEETS_HERE")
+                .Concat(BookStorage.OrderedPrefixesOfCssFilesToSortAfterUnknownStylesheets)
+                .Append("pageControls.css")
+                .Append("pageThumbnailList.css")
+                .ToArray();
+
         const int kDefaultValueForStyleSheetsThatShouldListInTheMiddle = 100;
 
         // This is set up as a dictionary, but it's really just a list of pairs.
@@ -29,7 +38,7 @@ namespace Bloom.Book
             {
                 _values = new Dictionary<string, int>();
                 var weight = 0;
-                BookStorage.KnownCssFilePrefixesInOrder.ForEach(x =>
+                KnownCssFilePrefixesInOrder.ForEach(x =>
                 {
                     if (x == "UNKNOWN_STYLESHEETS_HERE")
                         weight = kDefaultValueForStyleSheetsThatShouldListInTheMiddle + 1000;
--- a/src/BloomExe/Book/XMatterHelper.cs
+++ b/src/BloomExe/Book/XMatterHelper.cs
@@ -344,16 +344,8 @@ namespace Bloom.Book
                 }
             }
 
-            //note: for debugging the template/css purposes, it makes our life easier if, at runtime, the html is pointing the original.
-            //makes it easy to drop into a css editor and fix it up with the content we're looking at.
-            //TODO:But then later, we want to save it so that these are found in the same dir as the book.
-            _bookDom.AddStyleSheet(PathToXMatterStylesheet.ToLocalhost());
-            // Get the xMatter stylesheet link in the proper place rather than at the end.
-            // See https://issues.bloomlibrary.org/youtrack/issue/BL-8845.
-            _bookDom.SortStyleSheetLinks();
+            _bookDom.EnsureStylesheetLinks(Path.GetFileName(PathToXMatterStylesheet));
 
-            //it's important that we append *after* this, so that these values take precedence (the template will just have empty values for this stuff)
-            //REVIEW: I think all stylesheets now get sorted once they are all added: see HtmlDoc.SortStyleSheetLinks()
             XmlNode divBeforeNextFrontMatterPage = _bookDom.RawDom.SelectSingleNode(
                 "//body/div[@id='bloomDataDiv']"
             );
--- a/src/BloomExe/Edit/EditingView.cs
+++ b/src/BloomExe/Edit/EditingView.cs
@@ -528,7 +528,6 @@ namespace Bloom.Edit
                 }
                 _model.CheckForBL2634("navigated to page");
                 SetModalState(false); // ensure _pageListView is enabled (BL-9712).
-                _pageListView.Focus();
             }
 #if MEMORYCHECK
             // Check memory for the benefit of developers.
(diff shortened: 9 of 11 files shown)


nabalone (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6397)* status: 0 of 6 files reviewed, 2 unresolved discussions

___
*[`src/BloomExe/Book/Book.cs` line 2296 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6397#-Nv7OJVT8F8L6i801f9U:-Nv7OJVT8F8L6i801f9V:beqf2o) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/7191a0876c183773cdd509742a07817dfaa74fbb/src/BloomExe/Book/Book.cs#L2296)):*
> ```Smalltalk
>                 var coverColorIsDark = ColorUtils.IsDark(coverColor);
>                 foreach (
>                     var page in bookDOM
> ```

Just whitespace, not sure why it did this. 
___
*[`src/BloomExe/Edit/EditingView.cs` line 531 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6397#-Nv7OQQrBElylRXkgru9:-Nv7OQQtAWtMU92grk9W:b-ppuzus) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/e1509209b90b7ea225afbd4a39f4034fe62d08f6/src/BloomExe/Edit/EditingView.cs#L531)):*
> ```Smalltalk
>                 _model.CheckForBL2634("navigated to page");
>                 SetModalState(false); // ensure _pageListView is enabled (BL-9712).
>                 _pageListView.Focus();
> ```

I'm a little nervous that there will be side effects of removing this line that I didn't notice. But focus goes back to the page you are editing very soon anyway, and it seems complicated to keep this line and yet prevent the browser from putting the :focus-visible automatic outline

JohnThomson (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6397)* status: all files reviewed, 2 unresolved discussions (waiting on @nabalone)

___
*[`src/BloomExe/Edit/EditingView.cs` line 531 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6397#-Nv7OQQrBElylRXkgru9:-Nv7VKPN6rbPH3OIn7ZC:bx9ikks) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/e1509209b90b7ea225afbd4a39f4034fe62d08f6/src/BloomExe/Edit/EditingView.cs#L531)):*

I'm a little nervous that there will be side effects of removing this line that I didn't notice. But focus goes back to the page you are editing very soon anyway, and it seems complicated to keep this line and yet prevent the browser from putting the :focus-visible automatic outline
</blockquote></details>

I would not think we ever want the page list view focused, unless the user does something explicit with the keyboard (like pressing tab key) to manipulate those controls with the keyboard. The natural place for initial focus is somewhere in the page content. Even if there's nothing there to focus, I'm not sure we want focus in the page list.
You might try checking in git in case blame ties this line to a YouTrack issue that would suggest a reason for it or something we should test to make sure removing it doesn't re-introduce a problem. But I wouldn't spend a lot of time on it.

andrew-polk:
*[`src/BloomExe/Edit/EditingView.cs` line 531 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6397#-Nv7OQQrBElylRXkgru9:-Nv7ZbnQ0nYq4qYOmEaH:b-zbr9pg) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/e1509209b90b7ea225afbd4a39f4034fe62d08f6/src/BloomExe/Edit/EditingView.cs#L531)):*

I would not think we ever want the page list view focused, unless the user does something explicit with the keyboard (like pressing tab key) to manipulate those controls with the keyboard. The natural place for initial focus is somewhere in the page content. Even if there's nothing there to focus, I'm not sure we want focus in the page list.
You might try checking in git in case blame ties this line to a YouTrack issue that would suggest a reason for it or something we should test to make sure removing it doesn't re-introduce a problem. But I wouldn't spend a lot of time on it.
</blockquote></details>

I have a vague recollection that Gordon did this to work around some bizarro bug.
I even have the vague recollection that we decided to live with this flashing as a result.
So yes, if you can use history to look at why we did, that would be good.

nabalone:
*[`src/BloomExe/Edit/EditingView.cs` line 531 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6397#-Nv7OQQrBElylRXkgru9:-Nv7cVSE2-oKMMiGYMbZ:b-n0ccy) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/e1509209b90b7ea225afbd4a39f4034fe62d08f6/src/BloomExe/Edit/EditingView.cs#L531)):*

I have a vague recollection that Gordon did this to work around some bizarro bug.
I even have the vague recollection that we decided to live with this flashing as a result.
So yes, if you can use history to look at why we did, that would be good.
</blockquote></details>


Looks like JH added it 12 years ago. I don't see any related information in the commit

nabalone:
*[`src/BloomExe/Edit/EditingView.cs` line 531 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6397#-Nv7OQQrBElylRXkgru9:-Nv7dAI1EGvypTDVYRMT:b-6rj6ic) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/e1509209b90b7ea225afbd4a39f4034fe62d08f6/src/BloomExe/Edit/EditingView.cs#L531)):*


Looks like JH added it 12 years ago. I don't see any related information in the commit
</blockquote></details>

Looks like it came up in https://issues.bloomlibrary.org/youtrack/issue/BL-11539/WV2-add-page-dialogue-reappears-briefly-after-closing, I'm looking at what is going on there

nabalone:
*[`src/BloomExe/Edit/EditingView.cs` line 531 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6397#-Nv7OQQrBElylRXkgru9:-Nv7dYba8uzKVz50oCE1:b-hcln4f) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/e1509209b90b7ea225afbd4a39f4034fe62d08f6/src/BloomExe/Edit/EditingView.cs#L531)):*

Looks like it came up in https://issues.bloomlibrary.org/youtrack/issue/BL-11539/WV2-add-page-dialogue-reappears-briefly-after-closing, I'm looking at what is going on there
</blockquote></details>

Ah, was considered Issues We'll Live With: https://issues.bloomlibrary.org/youtrack/issue/BL-11980/WV2-styling-difference-for-Add-Page-button

andrew-polk:
*[`src/BloomExe/Edit/EditingView.cs` line 531 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6397#-Nv7OQQrBElylRXkgru9:-Nv7hGPu84lWLpODsWog:b8p38r9) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/e1509209b90b7ea225afbd4a39f4034fe62d08f6/src/BloomExe/Edit/EditingView.cs#L531)):*

Ah, was considered Issues We'll Live With: https://issues.bloomlibrary.org/youtrack/issue/BL-11980/WV2-styling-difference-for-Add-Page-button
</blockquote></details>

I think this is the change I was thinking of. So probably not related.
https://reviewable.io/reviews/BloomBooks/BloomDesktop/5641

andrew-polk (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6397)* status: all files reviewed, 2 unresolved discussions (waiting on @JohnThomson and @nabalone)

___
*[`src/BloomExe/Book/BookStorage.cs` line 3847 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6397#-NvDVhE79ZRcI2uUUSe9:-NvDVhE79ZRcI2uUUSeA:b28lwx8) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/7191a0876c183773cdd509742a07817dfaa74fbb/src/BloomExe/Book/BookStorage.cs#L3847)):*
> ```Smalltalk
>             "pageControls.css",
>             "pageThumbnailList.css",
>         };
> ```

I'm pretty sure this list of css files should only include ones which are part of the book.
But you've added a couple which are part of the Bloom UI.

At least, that's what I would expect in BookStorage. But maybe I'm wrong. I see editMode and previewMode there, too. But at least those are about how to show the book. The two new ones you added are not.

Hopefully your other PR is making all this irrelevant, but if we need two lists like this for some reason, 
1. Let's put the one which has UI stuff in it in another class
2. Let's make on derived from the other

nabalone:
*[`src/BloomExe/Book/BookStorage.cs` line 3847 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6397#-NvDVhE79ZRcI2uUUSe9:-NvDoUh1BBeK9dr8KRvn:bl5ov04) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/7191a0876c183773cdd509742a07817dfaa74fbb/src/BloomExe/Book/BookStorage.cs#L3847)):*

I'm pretty sure this list of css files should only include ones which are part of the book.
But you've added a couple which are part of the Bloom UI.

At least, that's what I would expect in BookStorage. But maybe I'm wrong. I see editMode and previewMode there, too. But at least those are about how to show the book. The two new ones you added are not.

Hopefully your other PR is making all this irrelevant, but if we need two lists like this for some reason, 
1. Let's put the one which has UI stuff in it in another class
2. Let's make on derived from the other
</blockquote></details>

Hmm, is this better?

nabalone (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6397)* status: 5 of 7 files reviewed, 3 unresolved discussions (waiting on @andrew-polk and @JohnThomson)

___
*[`src/BloomExe/Book/BookStorage.cs` line 3828 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6397#-NvDpqjY1xTdYQELFOVz:-NvDpqjY1xTdYQELFOW-:bkhhtz5) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/fb677d91f23f245170b62eaf6e7bdb71a19d7ace/src/BloomExe/Book/BookStorage.cs#L3828)):*
> ```Smalltalk
> 
>         // These go before "unknown" stylesheets in the sort order
>         public static readonly string[] AutomaticallyAddedCssPrefixesPart1 =
> ```

Also not sure what to name these

andrew-polk:
*[`src/BloomExe/Book/BookStorage.cs` line 3847 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6397#-NvDVhE79ZRcI2uUUSe9:-NvXQQQAAwmGsJVIdJSR:bmt1yvo) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/7191a0876c183773cdd509742a07817dfaa74fbb/src/BloomExe/Book/BookStorage.cs#L3847)):*

Hmm, is this better?
</blockquote></details>

The comment 
// MakeDomRelocatable adds these and uses this list to get rid of old ones before addding new ones.

Does not seem to be true. In fact, `AutomaticallyAddedCssFilePrefixes` isn't used anywhere. Probably things got messed up when you tried to pull the PRs apart.
So maybe it is best/easiest just to make it all one PR.
Whatever is going to be easiest to move this forward for you.

andrew-polk (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6397)* status: 5 of 7 files reviewed, 2 unresolved discussions (waiting on @JohnThomson and @nabalone)

andrew-polk (changes requested):
@JohnThomson will you please review this?

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6397#-:-NvbZ4oo9VHeP_lpHFGr:b-bifarg)* status: 3 of 12 files reviewed, 2 unresolved discussions (waiting on @JohnThomson and @nabalone)

___
*[`src/BloomExe/Book/BookStorage.cs` line 3847 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6397#-NvDVhE79ZRcI2uUUSe9:-NvbZ--Q0Vqr8tUCh6sY:b-wb7ioq) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/7191a0876c183773cdd509742a07817dfaa74fbb/src/BloomExe/Book/BookStorage.cs#L3847)):*

The comment 
// MakeDomRelocatable adds these and uses this list to get rid of old ones before addding new ones.

Does not seem to be true. In fact, `AutomaticallyAddedCssFilePrefixes` isn't used anywhere. Probably things got messed up when you tried to pull the PRs apart.
So maybe it is best/easiest just to make it all one PR.
Whatever is going to be easiest to move this forward for you.
</blockquote></details>

I'm still missing something, because `AutomaticallyAddedCssFilePrefixes` still doesn't appear to be used.

andrew-polk (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6397)* status: 3 of 12 files reviewed, 3 unresolved discussions (waiting on @JohnThomson)

___
*[`src/BloomExe/Book/BookStorage.cs` line 3847 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6397#-NvDVhE79ZRcI2uUUSe9:-Nvcq8oB7ijtmyIfM1tA:bbaygyo) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/7191a0876c183773cdd509742a07817dfaa74fbb/src/BloomExe/Book/BookStorage.cs#L3847)):*

It's just for RemoveNormalStyleSheetsLinks  in HtmlDom.css (I added a review comment to flag it)
</blockquote></details>

Thanks. Sorry; I don't know why I wasn't finding that before.

JohnThomson (changes requested):
Looks pretty good. I caught a few things and made some suggestions.

> Reviewed 8 of 9 files at r3, 1 of 1 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6397#-:-Nvd3LxC22ZTQWQnnY-w:btqkniw)* status: all files reviewed, 11 unresolved discussions (waiting on @andrew-polk and @nabalone)

___
*[`src/BloomExe/Book/Book.cs` line 655 at r4](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6397#-Nvcttaf5GQ6kvcU_XV4:-Nvcttaf5GQ6kvcU_XV5:bf083z8) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/dc4726b07127d8f2f714598bcf87c3c9ad77dd71/src/BloomExe/Book/Book.cs#L655)):*
> ```Smalltalk
>             var result = Storage.MakeDomRelocatable(inputDom);
>             result.AddStyleSheets("previewMode.css");
>             return result;
> ```

I think this is probably right. But is it a deliberate change that previewMode.css is now only added to the result, not to inputDom?
___
*[`src/BloomExe/Book/BookStorage.cs` line 3828 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6397#-NvDpqjY1xTdYQELFOVz:-Nvcv-PtEoIItg-IR23V:b-jybrhd) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/fb677d91f23f245170b62eaf6e7bdb71a19d7ace/src/BloomExe/Book/BookStorage.cs#L3828)):*

Also not sure what to name these
</blockquote></details>

How about, PrefixesOfCssFilesToSortBeforeUnknownStylesheets? And then PrefixesOfCssFilesToSortAfterUnknownStylesheets? It would be nice to get in there the notion that they are ordered, though. Maybe OrderedPrefixesOfCssFilesToSortBeforeUnknownStylesheets? We only use these in a couple of places, so we can afford to be verbose.
___
*[`src/BloomExe/Book/BookStorage.cs` line 3825 at r4](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6397#-NvcvuiB86yyNX9pG_49:-NvcvuiB86yyNX9pG_4A:bybdoii) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/dc4726b07127d8f2f714598bcf87c3c9ad77dd71/src/BloomExe/Book/BookStorage.cs#L3825)):*
> ```Smalltalk
>             AutomaticallyAddedCssPrefixesPart1.Concat(AutomaticallyAddedCssPrefixesPart2).ToArray();
> 
>         // These are split up for the sake of the StyleSheetLinkSorter
> ```

Looks like this applies to the line above. Usually we put comments before the line they describe.
___
*[`src/BloomExe/Book/HtmlDom.cs` line 269 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6397#-Nvd1QCeBPFNIy5gc24B:-Nvd1QCf8ytowLeVqhY8:bn9cje0) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/7191a0876c183773cdd509742a07817dfaa74fbb/src/BloomExe/Book/HtmlDom.cs#L269)):*
> ```Smalltalk
>             {
>                 var href = linkNode.GetAttribute("href");
>                 if (Path.GetFileName(href) == Path.GetFileName(path))
> ```

On thing we've changed here is losing the GetFileName. I was assuming we don't need it because the stylesheets (except ../customCollectionStyles.css) are always at the root of the book folder, and we always pass the plain file names. But that isn't true down where EpubMaker is passing some full paths, or even ToLocalHost() of some full paths. I'm not keen to put GetFileName() back in, because I want it to handle ../customCollectionStyles.css correctly. But you need to check callers, maybe even worth asserting that anything other than ../customCollectionStyles.css doesn't have path separators.
___
*[`src/BloomExe/Book/HtmlDom.cs` line 2951 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6397#-NvcDdZR4b608JHxdHg9:-Nvd-V_T948zG-kkG47b:bfrtde1) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/e1509209b90b7ea225afbd4a39f4034fe62d08f6/src/BloomExe/Book/HtmlDom.cs#L2951)):*

I don't really understand why this is the way it is. Could a reviewer please check if it looks like my refactor preserves behavior here?
</blockquote></details>

I think what we need is 
anyAdded = targetBookDom.AddStyleSheetsWithoutSorting(sourceBookDom.GetTemplateStyleSheets().ToArray());
targetBookDom.SortStyleSheetLinks();
return anyAdded.

Alternatively, you could make AddStyleSheets() return anyAdded, and then it would all collapse to
return targetBookDom.AddStyleSheets(...);

It raises an interesting question of whether AddStyleSheets() should sort if nothing was added. In some cases (but I'm not sure how many) it would save work if we didn't. OTOH in at least one place we are adding several things and then counting on one call to AddStyleSheets to sort them all. I'm guessing it's fairly unusual that nothing gets added. so probably not a case worth optimizing for.
___
*[`src/BloomExe/Book/HtmlDom.cs` line 254 at r4](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6397#-NvczBQdE11UxGFvjrbJ:-NvczBQdE11UxGFvjrbK:b-285sd2) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/dc4726b07127d8f2f714598bcf87c3c9ad77dd71/src/BloomExe/Book/HtmlDom.cs#L254)):*
> ```Smalltalk
>         internal bool UseOriginalImages { get; set; }
> 
>         public void AddStyleSheets(params string[] paths)
> ```

I think this deserves a comment along the lines "Make sure the DOM has a link to each of the stylesheets passed, and that all stylesheet links are in the right order.
Another idea is to go with the name we had before in BookStorage. EnsureStylesheets or EnsureStylesheetLinks is not as unwieldy as AddStylesheetIfMissing, but it does convey the idea that we will add them only if necessary.
___
*[`src/BloomExe/Book/HtmlDom.cs` line 261 at r4](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6397#-Nvcw52b1WLehvPBnVOY:-Nvcw52b1WLehvPBnVOZ:b-p6vqpv) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/dc4726b07127d8f2f714598bcf87c3c9ad77dd71/src/BloomExe/Book/HtmlDom.cs#L261)):*
> ```Smalltalk
> 
>         /// <summary>
>         /// Add a reference to the specified stylesheet if we don't already have it.
> ```

Needs a little tweaking to describe a list of paths.
___
*[`src/BloomExe/Book/HtmlDom.cs` line 283 at r4](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6397#-Nvcxm3k8v3CTq5lILqh:-Nvcxm3k8v3CTq5lILqi:bzjlmk) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/dc4726b07127d8f2f714598bcf87c3c9ad77dd71/src/BloomExe/Book/HtmlDom.cs#L283)):*
> ```Smalltalk
>             var pathToCheck = path;
>             if (Environment.OSVersion.Platform == PlatformID.Win32NT)
>                 pathToCheck = pathToCheck.ToLowerInvariant();
> ```

I'm having second thoughts about messing with case here. These are not random names the user typed. They are known names of standard stylesheets we use. If we get the case wrong in the name we pass to this method, unless it's already present with the right case, Linux won't find the file. We won't find that out, of course, until we start supporting Linux again.
Then again, if we do get it wrong, and don't check like this, currently on Windows all that will happen is that the old one won't get removed. So we'll have two links to 'the same' file.
Suggestion: remove the case changing code here. Remove it below also. Then down where we compare paths, do this:

filename = filename.Replace('\\', '/');
if (filename == pathToCheck)
   return false;
Debug.Assert(filename.ToLowerInvariant != pathToCheck.ToLowerInvariant, "passing a filename with wrong case: " + filename).

That way, we both don't waste time ignoring case in production, and do get immediate definite developer feedback if we're using the wrong case.
___
*[`src/BloomExe/Publish/Epub/EpubMaker.cs` line 1683 at r4](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6397#-Nvd2a2H7PE9TMclaK5P:-Nvd2a2H7PE9TMclaK5Q:b8r5e71) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/dc4726b07127d8f2f714598bcf87c3c9ad77dd71/src/BloomExe/Publish/Epub/EpubMaker.cs#L1683)):*
> ```Smalltalk
>                 pageDom.AddStyleSheetsWithoutSorting(
>                     baseEpubPath.ToLocalhost(),
>                     brandingPath.ToLocalhost()
> ```

Not sure why we're passing full paths and ToLocalhost() results here, but I think the new version of AddStyleSheets just wants file names...assuming the files are going to end up in the right place. I would expect our EpubMaker tests to have caught this...make sure they pass before you submit! (In fact this change is risky enough that you should run ALL our unit tests.)
___
*[`src/BloomExe/Publish/Epub/EpubMaker.cs` line 1695 at r4](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6397#-Nvd30BYB-Pt6WIpl0ql:-Nvd30BYB-Pt6WIpl0qm:bub44xh) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/dc4726b07127d8f2f714598bcf87c3c9ad77dd71/src/BloomExe/Publish/Epub/EpubMaker.cs#L1695)):*
> ```Smalltalk
>                     .LocateFileWithThrow(@"previewMode.css");
>                 var origamiPath = Storage.GetFileLocator().LocateFileWithThrow(@"origami.css");
>                 pageDom.AddStyleSheetsWithoutSorting(previewModePath, origamiPath);
> ```

Again, the new version won't handle paths.

nabalone:
*[`src/BloomExe/Publish/Epub/EpubMaker.cs` line 1683 at r4](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6397#-Nvd2a2H7PE9TMclaK5P:-NvhxFLW3CcGYj-Uv3rS:b-snue52) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/dc4726b07127d8f2f714598bcf87c3c9ad77dd71/src/BloomExe/Publish/Epub/EpubMaker.cs#L1683)):*

Not sure why we're passing full paths and ToLocalhost() results here, but I think the new version of AddStyleSheets just wants file names...assuming the files are going to end up in the right place. I would expect our EpubMaker tests to have caught this...make sure they pass before you submit! (In fact this change is risky enough that you should run ALL our unit tests.)
</blockquote></details>

Should I keep `Storage.GetFileLocator().LocateFileWithThrow(@"baseEPUB.css");` in there so that it will still throw if the file can't be located?

nabalone (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6397)* status: 5 of 12 files reviewed, 11 unresolved discussions (waiting on @andrew-polk and @JohnThomson)

___
*[`src/BloomExe/Book/Book.cs` line 655 at r4](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6397#-Nvcttaf5GQ6kvcU_XV4:-NvgQD0q3ZztBPgQ5zAX:bmabzfy) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/dc4726b07127d8f2f714598bcf87c3c9ad77dd71/src/BloomExe/Book/Book.cs#L655)):*

I think this is probably right. But is it a deliberate change that previewMode.css is now only added to the result, not to inputDom?
</blockquote></details>

The code on master is:
```
            //inputHead.ParentNode.ReplaceChild(importNode, inputHead);
            var result = Storage.MakeDomRelocatable(inputDom);
            Storage.EnsureHasLinkToStyleSheet(result, "previewMode.css");
            return result;
```

I think I had changed it to add previewMode.css to the inputDom and then changed it back
___
*[`src/BloomExe/Book/BookStorage.cs` line 3828 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6397#-NvDpqjY1xTdYQELFOVz:-NviXzeS3mi9SpAaZtM-:b-7ra9d5) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/fb677d91f23f245170b62eaf6e7bdb71a19d7ace/src/BloomExe/Book/BookStorage.cs#L3828)):*

How about, PrefixesOfCssFilesToSortBeforeUnknownStylesheets? And then PrefixesOfCssFilesToSortAfterUnknownStylesheets? It would be nice to get in there the notion that they are ordered, though. Maybe OrderedPrefixesOfCssFilesToSortBeforeUnknownStylesheets? We only use these in a couple of places, so we can afford to be verbose.
</blockquote></details>

Is it okay that that doesn't specify what css files should be in the list? I'm okay with it though

Also should we still keep `AutomaticallyAddedCssFilePrefixes` with its current name?
___
*[`src/BloomExe/Book/BookStorage.cs` line 3825 at r4](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6397#-NvcvuiB86yyNX9pG_49:-NvgqHNf03CSzMXiPn28:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/dc4726b07127d8f2f714598bcf87c3c9ad77dd71/src/BloomExe/Book/BookStorage.cs#L3825)):*

Looks like this applies to the line above. Usually we put comments before the line they describe.
</blockquote></details>

Done.
___
*[`src/BloomExe/Book/HtmlDom.cs` line 269 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6397#-Nvd1QCeBPFNIy5gc24B:-NviPOix00uJ_s7UScGR:bfqjwoa) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/7191a0876c183773cdd509742a07817dfaa74fbb/src/BloomExe/Book/HtmlDom.cs#L269)):*

On thing we've changed here is losing the GetFileName. I was assuming we don't need it because the stylesheets (except ../customCollectionStyles.css) are always at the root of the book folder, and we always pass the plain file names. But that isn't true down where EpubMaker is passing some full paths, or even ToLocalHost() of some full paths. I'm not keen to put GetFileName() back in, because I want it to handle ../customCollectionStyles.css correctly. But you need to check callers, maybe even worth asserting that anything other than ../customCollectionStyles.css doesn't have path separators.
</blockquote></details>

What is the situation with `css/fonts.css`? 
___
*[`src/BloomExe/Book/HtmlDom.cs` line 2951 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6397#-NvcDdZR4b608JHxdHg9:-NvhQmggCIdy7XIyYBBz:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/e1509209b90b7ea225afbd4a39f4034fe62d08f6/src/BloomExe/Book/HtmlDom.cs#L2951)):*

I think what we need is 
anyAdded = targetBookDom.AddStyleSheetsWithoutSorting(sourceBookDom.GetTemplateStyleSheets().ToArray());
targetBookDom.SortStyleSheetLinks();
return anyAdded.

Alternatively, you could make AddStyleSheets() return anyAdded, and then it would all collapse to
return targetBookDom.AddStyleSheets(...);

It raises an interesting question of whether AddStyleSheets() should sort if nothing was added. In some cases (but I'm not sure how many) it would save work if we didn't. OTOH in at least one place we are adding several things and then counting on one call to AddStyleSheets to sort them all. I'm guessing it's fairly unusual that nothing gets added. so probably not a case worth optimizing for.
</blockquote></details>

Done.
___
*[`src/BloomExe/Book/HtmlDom.cs` line 254 at r4](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6397#-NvczBQdE11UxGFvjrbJ:-NvgrtA0-Aw17CnzdTAU:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/dc4726b07127d8f2f714598bcf87c3c9ad77dd71/src/BloomExe/Book/HtmlDom.cs#L254)):*

I think this deserves a comment along the lines "Make sure the DOM has a link to each of the stylesheets passed, and that all stylesheet links are in the right order.
Another idea is to go with the name we had before in BookStorage. EnsureStylesheets or EnsureStylesheetLinks is not as unwieldy as AddStylesheetIfMissing, but it does convey the idea that we will add them only if necessary.
</blockquote></details>

Done.
___
*[`src/BloomExe/Book/HtmlDom.cs` line 261 at r4](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6397#-Nvcw52b1WLehvPBnVOY:-NvgsCab0ibTUeCE--lg:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/dc4726b07127d8f2f714598bcf87c3c9ad77dd71/src/BloomExe/Book/HtmlDom.cs#L261)):*

Needs a little tweaking to describe a list of paths.
</blockquote></details>

Done.
___
*[`src/BloomExe/Book/HtmlDom.cs` line 283 at r4](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6397#-Nvcxm3k8v3CTq5lILqh:-NvhFA9t-7-NUS_vp3Ax:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/dc4726b07127d8f2f714598bcf87c3c9ad77dd71/src/BloomExe/Book/HtmlDom.cs#L283)):*

I'm having second thoughts about messing with case here. These are not random names the user typed. They are known names of standard stylesheets we use. If we get the case wrong in the name we pass to this method, unless it's already present with the right case, Linux won't find the file. We won't find that out, of course, until we start supporting Linux again.
Then again, if we do get it wrong, and don't check like this, currently on Windows all that will happen is that the old one won't get removed. So we'll have two links to 'the same' file.
Suggestion: remove the case changing code here. Remove it below also. Then down where we compare paths, do this:

filename = filename.Replace('\\', '/');
if (filename == pathToCheck)
   return false;
Debug.Assert(filename.ToLowerInvariant != pathToCheck.ToLowerInvariant, "passing a filename with wrong case: " + filename).

That way, we both don't waste time ignoring case in production, and do get immediate definite developer feedback if we're using the wrong case.
</blockquote></details>

Done.
___
*[`src/BloomExe/Publish/Epub/EpubMaker.cs` line 1695 at r4](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6397#-Nvd30BYB-Pt6WIpl0ql:-NviMDzR0tXJBZRM0p8p:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/dc4726b07127d8f2f714598bcf87c3c9ad77dd71/src/BloomExe/Publish/Epub/EpubMaker.cs#L1695)):*

Again, the new version won't handle paths.
</blockquote></details>

Done.
___
*[`src/BloomExe/Publish/Epub/EpubMaker.cs` line 1767 at r4](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6397#-Nvig6D6BNpQBTWqjxpV:-Nvig6D6BNpQBTWqjxpW:b-s9ejkb) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/dc4726b07127d8f2f714598bcf87c3c9ad77dd71/src/BloomExe/Publish/Epub/EpubMaker.cs#L1767)):*
> ```Smalltalk
>                 link.SetAttribute("href", kCssFolder + "/" + name);
>             }
>             pageDom.AddStyleSheets(kCssFolder + "/" + "fonts.css"); // enhance: could omit if we don't embed any
> ```

Flagging here `css/fonts.css` that is also not a filename but seems like we still want href=css/fonts.css?

nabalone:
*[`src/BloomExe/Book/XMatterHelper.cs` line 350 at r5](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6397#-NvlXD2B375dqZqhRyw4:-NvlXD2B375dqZqhRyw5:bntzdur) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/a272590ac827c6c17a0b15be7709cf3064ab1dd0/src/BloomExe/Book/XMatterHelper.cs#L350)):*
> ```Smalltalk
>             //makes it easy to drop into a css editor and fix it up with the content we're looking at.
>             //TODO:But then later, we want to save it so that these are found in the same dir as the book.
>             _bookDom.EnsureStylesheetLinks(Path.GetFileName(PathToXMatterStylesheet));
> ```

@JohnThomson  is this change okay? If not, how do we want to handle the path? The original code is 
` _bookDom.AddStyleSheet(PathToXMatterStylesheet.ToLocalhost());` (https://github.com/BloomBooks/BloomDesktop/blob/d6e59e7943a203cec590863a6643eb138e54489f/src/BloomExe/Book/XMatterHelper.cs#L346C1-L350C75)

JohnThomson (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6397)* status: 5 of 12 files reviewed, 3 unresolved discussions (waiting on @nabalone)

___
*[`src/BloomExe/Book/BookStorage.cs` line 3828 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6397#-NvDpqjY1xTdYQELFOVz:-NvlqWdx1YH-yd2yD2jZ:b-tqnthy) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/fb677d91f23f245170b62eaf6e7bdb71a19d7ace/src/BloomExe/Book/BookStorage.cs#L3828)):*

Is it okay that that doesn't specify what css files should be in the list? I'm okay with it though

Also should we still keep `AutomaticallyAddedCssFilePrefixes` with its current name?
</blockquote></details>

I think what you have now is good.
___
*[`src/BloomExe/Book/HtmlDom.cs` line 269 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6397#-Nvd1QCeBPFNIy5gc24B:-Nvlr8ob3LWjuHoeTK_u:bwf81ro) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/7191a0876c183773cdd509742a07817dfaa74fbb/src/BloomExe/Book/HtmlDom.cs#L269)):*

What is the situation with `css/fonts.css`?
</blockquote></details>

I'm not sure. I'm afraid you probably need to look at all callers. If they are passing something other than a name with no path separators, make sure the relevant code is still working: that the link we put into the book is actually finding a stylesheet. If you can get the DOM on screen in a browser, the Network tab will tell you whether each file request was successful.
For example, if something is passing css/fonts.css, then hopefully something is making sure that the right fonts.css file is in a css folder under the book folder. If in fact fonts.css is being put in the root folder, then you should change the caller to just pass fonts.css.
___
*[`src/BloomExe/Book/XMatterHelper.cs` line 350 at r5](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6397#-NvlXD2B375dqZqhRyw4:-NvltSDE3cN_K9CFWGkk:b-jvk6nv) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/a272590ac827c6c17a0b15be7709cf3064ab1dd0/src/BloomExe/Book/XMatterHelper.cs#L350)):*

@JohnThomson  is this change okay? If not, how do we want to handle the path? The original code is 
` _bookDom.AddStyleSheet(PathToXMatterStylesheet.ToLocalhost());` (https://github.com/BloomBooks/BloomDesktop/blob/d6e59e7943a203cec590863a6643eb138e54489f/src/BloomExe/Book/XMatterHelper.cs#L346C1-L350C75)
</blockquote></details>

It's a bit tricky. The original code would LOOK for an exiting link with the same filename, and if it found it, leave it and not add the new one. But otherwise (which is probably more common) it would add the new one complete with its path.
I wrote that code quite recently, but it now seems clearly wrong: if the current caller wants a link to, say, ../customCollectionStyles.css, we should not just ignore that because we find one to customCollectionStyles.css.
The current xmatter stylesheet gets copied to the book folder, so a simple link to the filename will work. I'm about 95% sure that there is no longer a debugging advantage to pointing the link at the original (like the old comment says). JohnH's latest change (which might not be in your source yet unless you rebase) causes things to be automatically updated when a new xmatter CSS is created. So I think it is fine that we just point to the local copy. In fact, it will probably make things more robust. I think we should go with the change you've made. But you should probably do a bit of testing that changing xmatter packs has the expected effects.
___
*[`src/BloomExe/Edit/EditingView.cs` line 531 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6397#-Nv7OQQrBElylRXkgru9:-Nvlykje-V_PApZ0t_Y1:b-8wa64a) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/e1509209b90b7ea225afbd4a39f4034fe62d08f6/src/BloomExe/Edit/EditingView.cs#L531)):*

I think this is the change I was thinking of. So probably not related.
https://reviewable.io/reviews/BloomBooks/BloomDesktop/5641
</blockquote></details>

Looks like we've done all we can to look for reasons why focus was needed there. I think we can go ahead and remove it, as you have done.
___
*[`src/BloomExe/Publish/Epub/EpubMaker.cs` line 1683 at r4](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6397#-Nvd2a2H7PE9TMclaK5P:-NvlzPeFCrVYgjIX9Leg:b-m0cny9) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/dc4726b07127d8f2f714598bcf87c3c9ad77dd71/src/BloomExe/Publish/Epub/EpubMaker.cs#L1683)):*

Should I keep `Storage.GetFileLocator().LocateFileWithThrow(@"baseEPUB.css");` in there so that it will still throw if the file can't be located?
</blockquote></details>

I don't really think we need it. We've been reliably generating these files for a long time now.
___
*[`src/BloomExe/Publish/Epub/EpubMaker.cs` line 1767 at r4](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6397#-Nvig6D6BNpQBTWqjxpV:-Nvm-pZZ0_p_A1CWk0Nr:beq0ntr) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/dc4726b07127d8f2f714598bcf87c3c9ad77dd71/src/BloomExe/Publish/Epub/EpubMaker.cs#L1767)):*

Flagging here `css/fonts.css` that is also not a filename but seems like we still want href=css/fonts.css?
</blockquote></details>

Yes, looks like we want that. Check that the sorter has what it needs to sort this in the sequence we want, if it's in that list at all. Also that the file is, indeed, getting kept in this css directory. (Though I find myself asking, why?? Seems like if we have a CSS directory, all our CSS should be there, not just one file?? But let's not complicate this PR with trying to fix yet another historical mistake!)
___
*[`src/BloomExe/Publish/Epub/EpubMaker.cs` line 1688 at r7](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6397#-NvlzaSs6SnHQFFCFMz2:-NvlzaSs6SnHQFFCFMz3:bdh0gw9) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/88beedcb80e10081a2820c2579b61f62f0af5b3a/src/BloomExe/Publish/Epub/EpubMaker.cs#L1688)):*
> ```Smalltalk
>                 // Adding basePage.css is definitely a bad idea, because it's now possible we want basePageLegacy instead.
>                 Storage.GetFileLocator().LocateFileWithThrow(@"previewMode.css");
>                 Storage.GetFileLocator().LocateFileWithThrow(@"origami.css");
> ```

I don't think we need to locate the files here any more. Something, probably BookStorage.UpdateSupportFiles(), should be copying all the CSS we need into the book folder; that code should fail appropriately if they can't be found.
It's just possible that previewMode.css does NOT get copied into the book folder, if we only use it at runtime, in which case we may actually NEED to put some sort of full path to it into the link. Worth checking that it actually gets copied and found (probably you already did).
If we DO need a path here, we need to think about implications for the sorter. If we put previewMode.css into the link with a path, it won't match the entry in the sorter list.

nabalone (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6397)* status: 5 of 12 files reviewed, 2 unresolved discussions (waiting on @JohnThomson)

___
*[`src/BloomExe/Book/HtmlDom.cs` line 269 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6397#-Nvd1QCeBPFNIy5gc24B:-NvmsjXQC2FzRCr69GEK:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/7191a0876c183773cdd509742a07817dfaa74fbb/src/BloomExe/Book/HtmlDom.cs#L269)):*

I'm not sure. I'm afraid you probably need to look at all callers. If they are passing something other than a name with no path separators, make sure the relevant code is still working: that the link we put into the book is actually finding a stylesheet. If you can get the DOM on screen in a browser, the Network tab will tell you whether each file request was successful.
For example, if something is passing css/fonts.css, then hopefully something is making sure that the right fonts.css file is in a css folder under the book folder. If in fact fonts.css is being put in the root folder, then you should change the caller to just pass fonts.css.
</blockquote></details>

Done.
___
*[`src/BloomExe/Book/XMatterHelper.cs` line 350 at r5](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6397#-NvlXD2B375dqZqhRyw4:-Nvn3jUzBolrNY750GU_:bb6un1j) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/a272590ac827c6c17a0b15be7709cf3064ab1dd0/src/BloomExe/Book/XMatterHelper.cs#L350)):*

It's a bit tricky. The original code would LOOK for an exiting link with the same filename, and if it found it, leave it and not add the new one. But otherwise (which is probably more common) it would add the new one complete with its path.
I wrote that code quite recently, but it now seems clearly wrong: if the current caller wants a link to, say, ../customCollectionStyles.css, we should not just ignore that because we find one to customCollectionStyles.css.
The current xmatter stylesheet gets copied to the book folder, so a simple link to the filename will work. I'm about 95% sure that there is no longer a debugging advantage to pointing the link at the original (like the old comment says). JohnH's latest change (which might not be in your source yet unless you rebase) causes things to be automatically updated when a new xmatter CSS is created. So I think it is fine that we just point to the local copy. In fact, it will probably make things more robust. I think we should go with the change you've made. But you should probably do a bit of testing that changing xmatter packs has the expected effects.
</blockquote></details>

Done
___
*[`src/BloomExe/Publish/Epub/EpubMaker.cs` line 1683 at r4](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6397#-Nvd2a2H7PE9TMclaK5P:-NvmWap4EwrB0YGk4T6k:bb6un1j) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/dc4726b07127d8f2f714598bcf87c3c9ad77dd71/src/BloomExe/Publish/Epub/EpubMaker.cs#L1683)):*

I don't really think we need it. We've been reliably generating these files for a long time now.
</blockquote></details>

Done
___
*[`src/BloomExe/Publish/Epub/EpubMaker.cs` line 1767 at r4](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6397#-Nvig6D6BNpQBTWqjxpV:-NvmWEWX1gSmVRcH7qhr:bxkdwo9) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/dc4726b07127d8f2f714598bcf87c3c9ad77dd71/src/BloomExe/Publish/Epub/EpubMaker.cs#L1767)):*

Yes, looks like we want that. Check that the sorter has what it needs to sort this in the sequence we want, if it's in that list at all. Also that the file is, indeed, getting kept in this css directory. (Though I find myself asking, why?? Seems like if we have a CSS directory, all our CSS should be there, not just one file?? But let's not complicate this PR with trying to fix yet another historical mistake!)
</blockquote></details>

It works. Looks like there are other files in the css directory as well, not sure what the situation is
___
*[`src/BloomExe/Publish/Epub/EpubMaker.cs` line 1688 at r7](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6397#-NvlzaSs6SnHQFFCFMz2:-NvmWQxh6HPF0SBhVW_R:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/88beedcb80e10081a2820c2579b61f62f0af5b3a/src/BloomExe/Publish/Epub/EpubMaker.cs#L1688)):*

I don't think we need to locate the files here any more. Something, probably BookStorage.UpdateSupportFiles(), should be copying all the CSS we need into the book folder; that code should fail appropriately if they can't be found.
It's just possible that previewMode.css does NOT get copied into the book folder, if we only use it at runtime, in which case we may actually NEED to put some sort of full path to it into the link. Worth checking that it actually gets copied and found (probably you already did).
If we DO need a path here, we need to think about implications for the sorter. If we put previewMode.css into the link with a path, it won't match the entry in the sorter list.
</blockquote></details>

Done.

nabalone (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6397)* status: 4 of 12 files reviewed, 3 unresolved discussions (waiting on @andrew-polk and @JohnThomson)

___
*[`src/BloomExe/Book/BookStorage.cs` line 2986 at r8](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6397#-Nvn6SamE8PJaQO5GpB2:-Nvn6SamE8PJaQO5GpB3:bbhozt0) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/e1509209b90b7ea225afbd4a39f4034fe62d08f6/src/BloomExe/Book/BookStorage.cs#L2986)):*
> ```Smalltalk
> 
>             // If we're not supposed to have these links, make sure we don't.
>             AppearanceSettings.PossibleAppearanceRelatedCssFiles
> ```

@JohnThomson did you tell me to delete these lines? I think so but don't remember very well and just wanted to make sure it was intentional

JohnThomson (changes requested):
Sorry, picked up one more thing I missed earlier. Looking good overall.

> Reviewed 3 of 7 files at r5, 1 of 1 files at r6, 1 of 1 files at r7, 1 of 1 files at r8, 2 of 2 files at r9, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6397#-:-NvnOWrA16jpmpE-YWyP:bneo7rl)* status: all files reviewed, 1 unresolved discussion (waiting on @nabalone)

___
*[`src/BloomExe/Book/BookStorage.cs` line 2986 at r8](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6397#-Nvn6SamE8PJaQO5GpB2:-NvnMkJN9kY2f55_-afK:b-tllqrc) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/e1509209b90b7ea225afbd4a39f4034fe62d08f6/src/BloomExe/Book/BookStorage.cs#L2986)):*

@JohnThomson did you tell me to delete these lines? I think so but don't remember very well and just wanted to make sure it was intentional
</blockquote></details>

Yes, the whole block removing PossibleAppearanceRelatedCssFiles can go, because we verified that all the files in it area already cleaned out by RemoveNormalStylesheetsLinks (you may want to check that one more time).
___
*[`src/BloomExe/Book/HtmlDom.cs` line 269 at r9](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6397#-NvnNB3R4OX2zTftqHdR:-NvnNB3R4OX2zTftqHdS:bgy0pwf) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/56485a2cea1c84b196aaa0dbd06cb8b6a364768d/src/BloomExe/Book/HtmlDom.cs#L269)):*
> ```Smalltalk
>         /// Return true if any stylesheets were added
>         /// </summary>
>         public bool AddStyleSheetsWithoutSorting(params string[] paths)
> ```

Let's be consistent and call this EnsureStylesheetLinksWithoutSorting.
Also: it's not obvious to me whether it should be Stylesheet or StyleSheet. You might see which we use most often. I think stylesheet is a single word (my spell checker thinks so, anyway) in which case we should not be capitalizing "sheet". Let's at least get all the names we're changing consistent with whichever we decide is right.

nabalone (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6397)* status: 10 of 12 files reviewed, 1 unresolved discussion (waiting on @JohnThomson)

___
*[`src/BloomExe/Book/HtmlDom.cs` line 269 at r9](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6397#-NvnNB3R4OX2zTftqHdR:-Nvnj2-Q06NVcyY2An8a:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/56485a2cea1c84b196aaa0dbd06cb8b6a364768d/src/BloomExe/Book/HtmlDom.cs#L269)):*

Let's be consistent and call this EnsureStylesheetLinksWithoutSorting.
Also: it's not obvious to me whether it should be Stylesheet or StyleSheet. You might see which we use most often. I think stylesheet is a single word (my spell checker thinks so, anyway) in which case we should not be capitalizing "sheet". Let's at least get all the names we're changing consistent with whichever we decide is right.
</blockquote></details>

Done.

PR #6403, opened 2024-04-11
BL-12872 fix error messages regression

nabalone:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6403)


src/BloomExe/WebLibraryIntegration/BloomLibraryBookApiClient.cs
```diff
@@ -16,6 +16,13 @@
 
 namespace Bloom.WebLibraryIntegration
 {
+    // To bubble up the fact that the exception was caused by the client being out of date.
+    public class VersionCannotUploadException : ApplicationException
```
nabalone, line 20: Not sure where to put this

PR #6411, opened 2024-04-12
Show badges for books from Sandbox (dev.bloomLibrary) (BL-13216)

JohnThomson:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6411)

### DIFF
--- a/src/BloomExe/Book/BloomLibraryStatus.cs
+++ b/src/BloomExe/Book/BloomLibraryStatus.cs
@@ -35,7 +35,10 @@ namespace Bloom.Book
             BloomLibraryBookUrl = bloomLibraryBookUrl;
         }
 
-        public static BloomLibraryStatus FromDynamicJson(dynamic bookState)
+        public static BloomLibraryStatus FromDynamicJson(
+            dynamic bookState,
+            bool forceUseProductionData
+        )
         {
             //Debug.WriteLine($"DEBUG draft={bookState.draft}, inCirculation={bookState.inCirculation}, harvestState={bookState.harvestState}, objectId={bookState.objectId}");
             HarvesterState harvesterState = HarvesterState.Failed;
@@ -67,7 +70,10 @@ namespace Bloom.Book
             var inCirculation =
                 (bookState.inCirculation == null)
                 || bookState.inCirculation.ToString().ToLowerInvariant() == "true";
-            var url = BloomLibraryUrls.BloomLibraryDetailPageUrlFromBookId(bookState.id.ToString());
+            var url = BloomLibraryUrls.BloomLibraryDetailPageUrlFromBookId(
+                bookState.id.ToString(),
+                forceUseProductionData: forceUseProductionData
+            );
             return new BloomLibraryStatus(draft, !inCirculation, harvesterState, url);
         }
 
--- a/src/BloomExe/Collection/BookCollection.cs
+++ b/src/BloomExe/Collection/BookCollection.cs
@@ -505,13 +505,30 @@ namespace Bloom.Collection
             // (unless told not to).
             var bloomLibraryApiClient = new BloomLibraryBookApiClient();
             var bloomLibraryStatusesById = bloomLibraryApiClient.GetLibraryStatusForBooks(
-                bookInfos
+                bookInfos,
+                forceUseProductionData: true
             );
+            var sandboxBloomLibraryStatusesById = new Dictionary<string, BloomLibraryStatus>();
+            if (BookUpload.UseSandbox)
+            {
+                sandboxBloomLibraryStatusesById = bloomLibraryApiClient.GetLibraryStatusForBooks(
+                    bookInfos,
+                    forceUseProductionData: false
+                );
+            }
             // Now to store the data into the BookInfos and signal the UI to update.
             foreach (var bookInfo in bookInfos)
             {
                 bool updateThumbnailBadge = false;
-                if (bloomLibraryStatusesById.TryGetValue(bookInfo.Id, out var status))
+                if (
+                    // This gives precedence to the production status, if the book is in both places.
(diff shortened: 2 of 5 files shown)


andrew-polk (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6411)* status: all files reviewed, 3 unresolved discussions (waiting on @JohnThomson)

___
*[`src/BloomExe/Book/BloomLibraryStatus.cs` line 38 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6411#-NvJ8ug150ezISFD5LKB:-NvJ8ug150ezISFD5LKC:bkhia82) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/8ceea27f132b59a4da914edc71dfc212d337ae08/src/BloomExe/Book/BloomLibraryStatus.cs#L38)):*
> ```Smalltalk
>         }
> 
>         public static BloomLibraryStatus FromDynamicJson(dynamic bookState, bool ignoreSandbox)
> ```

Even though I knew what you were trying to do, I still struggled with understanding the name "ignoreSandbox".
I think ignoreSandboxFlag or ignoreSandboxSetting might help, but I wonder if something like alwaysUseProductionData or forceProductionData might be even better?
___
*[`src/BloomExe/Collection/BookCollection.cs` line 528 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6411#-NvJ85a42LfnSEv1kke4:-NvJ85a42LfnSEv1kke5:b-2zt3wz) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/8ceea27f132b59a4da914edc71dfc212d337ae08/src/BloomExe/Collection/BookCollection.cs#L528)):*
> ```Smalltalk
>                     // I think not: it's not a problem to have an extra upload in the sandbox,
>                     // since it won't create any ambiguity about which one to overwrite,
>                     // so the warning is inappropriate.
> ```

I agree we don't want to make it multiple just because there is one in each. 
IMO, you can remove this Review comment and make it more of a documentation of the decision.

If you're making a change, there is a typo in "th" on the first line of the comment.
___
*[`src/BloomExe/web/controllers/CollectionApi.cs` line 737 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6411#-NvJ6Vb68LiER_e6oc9q:-NvJ6Vb68LiER_e6oc9r:b-g54f0e) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/8ceea27f132b59a4da914edc71dfc212d337ae08/src/BloomExe/web/controllers/CollectionApi.cs#L737)):*
> ```Smalltalk
>                         draft = infos[0].BloomLibraryStatus.Draft,
>                         inCirculation = !infos[0].BloomLibraryStatus.NotInCirculation,
>                         harvestState = HarvesterState.Multiple.ToString().ToLowerInvariant()
> ```

I know you didn't do this, but I think setting this to multiple is wrong.
It should only be multiple if we got multiple back from the API call (which we will if the first book has multiple).
So I think we should just set this to the first book's value just like the other values.

In that case, I think this else clause and the else if before it are the same. And that makes sense to me. If we are only getting info based on bookInstanceId from the API, that info will be the same no matter how many times we ask for it.

JohnThomson (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6411)* status: 0 of 5 files reviewed, 3 unresolved discussions (waiting on @andrew-polk)

___
*[`src/BloomExe/Book/BloomLibraryStatus.cs` line 38 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6411#-NvJ8ug150ezISFD5LKB:-NvXQDZeBD5fepzu24L0:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/8ceea27f132b59a4da914edc71dfc212d337ae08/src/BloomExe/Book/BloomLibraryStatus.cs#L38)):*

Even though I knew what you were trying to do, I still struggled with understanding the name "ignoreSandbox".
I think ignoreSandboxFlag or ignoreSandboxSetting might help, but I wonder if something like alwaysUseProductionData or forceProductionData might be even better?
</blockquote></details>

Done.
___
*[`src/BloomExe/Collection/BookCollection.cs` line 528 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6411#-NvJ85a42LfnSEv1kke4:-NvXQp-i-yGNffgbi1b1:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/8ceea27f132b59a4da914edc71dfc212d337ae08/src/BloomExe/Collection/BookCollection.cs#L528)):*

I agree we don't want to make it multiple just because there is one in each. 
IMO, you can remove this Review comment and make it more of a documentation of the decision.

If you're making a change, there is a typo in "th" on the first line of the comment.
</blockquote></details>

Done.
___
*[`src/BloomExe/web/controllers/CollectionApi.cs` line 737 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6411#-NvJ6Vb68LiER_e6oc9q:-NvXRjPrDBGkg0X6RmD2:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/8ceea27f132b59a4da914edc71dfc212d337ae08/src/BloomExe/web/controllers/CollectionApi.cs#L737)):*

I know you didn't do this, but I think setting this to multiple is wrong.
It should only be multiple if we got multiple back from the API call (which we will if the first book has multiple).
So I think we should just set this to the first book's value just like the other values.

In that case, I think this else clause and the else if before it are the same. And that makes sense to me. If we are only getting info based on bookInstanceId from the API, that info will be the same no matter how many times we ask for it.
</blockquote></details>

Done.

andrew-polk:
LGTM but Reviewable seems to be having issues.

PR #6422, opened 2024-04-17
Put various qtips into the right container (BL-13285)

JohnThomson:
In several places we were using the container: option of qtip wrongly.
This would cause qtips to show up in the wrong place when zoomed.
It also, for a reason I don't understand, prevented the proper functioning of a CSS rule for making qtips visible, overriding a rule intended to hide them when focus and the mouse are elsewhere. So this fixes a problem where overflow tooltips were being always hidden, too.

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6422)


andrew-polk (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6422)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @JohnThomson)

PR #6423, opened 2024-04-18
Fix failing unit tests related to front cover fixes (BL-13264)

StephenMcConnel:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6423)

### DIFF
--- a/src/BloomExe/Book/AppearanceSettings.cs
+++ b/src/BloomExe/Book/AppearanceSettings.cs
@@ -99,7 +99,7 @@ public class AppearanceSettings
     // It can also be set when WriteToFolder() updates the appearance files to be consistent with the current settings.
     // This only gets called if Bloom is allowed to make changes to the folder, so it remains false
     // if we are loading a legacy book in a folder we can't write.
-    private bool _areSettingsConsistentWithFiles;
+    internal bool _areSettingsConsistentWithFiles;  // internal for test access
 
     public bool CssFilesChecked { get; private set; }
 
--- a/src/BloomExe/Book/Book.cs
+++ b/src/BloomExe/Book/Book.cs
@@ -1710,11 +1710,6 @@ namespace Bloom.Book
             Storage.MigrateToLevel3PutImgFirst();
             Storage.MigrateToLevel4UseAppearanceSystem();
 
-            // Enhance: there's probably some case where this will get unnecessarily repeated,
-            // but I think it would only be in the very rare case of repeatedly needing to update
-            // a pre-4.9 book when we can't Save().
-            OurHtmlDom.FixDivOrdering();
-
             // Make sure the appearance settings have checked the current state of the css files.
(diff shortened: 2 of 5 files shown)


JohnThomson (changes requested):
Good sleuthing, but one thing I don't like about how you fixed it.

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6423#-:-Nvn3PR2-7z-7G0TmSSD:b-2er4ix)* status: all files reviewed, 1 unresolved discussion (waiting on @StephenMcConnel)

___
*[`src/BloomExe/Book/AppearanceSettings.cs` line 819 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6423#-Nvn2mBf596v5k_V7rhi:-Nvn2mBf596v5k_V7rhj:b-n3vp9q) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/1f5606c3e062b1a119b3c483a65af4ef1c22ce75/src/BloomExe/Book/AppearanceSettings.cs#L819)):*
> ```Smalltalk
>     /// Read in our settings from JSON (typically from appearance.json in the book folder)
>     /// </summary>
>     internal void UpdateFromJson(string json, bool updateConsistentFilesFlag = false)
> ```

I don't like adding this param, which puts back into this function something we decided didn't belong, as far as I can see only because one unit test needs to have the flag set after calling UpdateFromJson.
Could we instead just make _areSettingsConsistentWithFiles internal, with a note that this is for testing, and set it directly in the unit test?

StephenMcConnel (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6423)* status: 3 of 5 files reviewed, 1 unresolved discussion (waiting on @JohnThomson)

___
*[`src/BloomExe/Book/AppearanceSettings.cs` line 819 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6423#-Nvn2mBf596v5k_V7rhi:-Nvn5Y_t3nZI9EX6MUJp:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/1f5606c3e062b1a119b3c483a65af4ef1c22ce75/src/BloomExe/Book/AppearanceSettings.cs#L819)):*

I don't like adding this param, which puts back into this function something we decided didn't belong, as far as I can see only because one unit test needs to have the flag set after calling UpdateFromJson.
Could we instead just make _areSettingsConsistentWithFiles internal, with a note that this is for testing, and set it directly in the unit test?
</blockquote></details>

Done.

JohnThomson (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6423)* status: all files reviewed, 1 unresolved discussion (waiting on @StephenMcConnel)

PR #6425, opened 2024-04-18
Replace outline with padding when no outline is shown (BL-13288)

StephenMcConnel:
The 2px difference in height and width could have led to a very subtle cause of differences in text wrapping between editing and publishing.

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6425)


hatton:
Hmmmm?

StephenMcConnel (commented):
It probably does deserve some thought, but this does fix the problem.

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6425#-:-NvnLR4j0I9i-RHydq3A:b2ytq7u)* status: 0 of 3 files reviewed, all discussions resolved

PR #6426, opened 2024-04-19
Fix BL-13288 where edit mode was 1px off from publish

hatton:
The border of the translationGroup was pushing the editable box over. We can live with out it if we add an `outline` to the editable box.

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6426)

### DIFF
--- a/src/BloomBrowserUI/bookEdit/css/editMode.less
+++ b/src/BloomBrowserUI/bookEdit/css/editMode.less
@@ -1218,11 +1218,16 @@ body :has(.bloom-page:focus-within, .bloom-page:hover) {
     // Show the padding, multilingual gap, etc as colored areas.
     // The ".split-pane-component-inner >"  disables this for overlays, where it's common to have backgrounds like transparent
     .split-pane-component-inner > .bloom-translationGroup {
-        border: solid thin var(--page-structure-color);
-        box-sizing: border-box;
+        // no: a border here will take up space, wrecking WYSIWYG
+        // border: solid thin var(--page-structure-color);
+        // box-sizing: border-box;
+        // Also, you'd think you could use outline instead, but it doesn't show on the side; the
+        // children over it up because they are the full width of this parent.
 
         background-color: var(--page-structure-color) !important;
         .bloom-editable {
+            outline: 1px solid var(--page-structure-color);
+            outline-offset: -1px; // else the white of an un-focussed box can just bleed into the white of the margin
             background-color: var(--marginBox-background-color) !important;
         }
     }


JohnThomson (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6426)* status: all files reviewed, 1 unresolved discussion (waiting on @hatton)

___
*[`src/BloomBrowserUI/bookEdit/css/editMode.less` line 1225 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6426#-Nvqht3e6vau0Mal2Q-w:-Nvqht3e6vau0Mal2Q-x:bsun0wj) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/f0415b4a324b9679601663f8a30969284430bea6/src/BloomBrowserUI/bookEdit/css/editMode.less#L1225)):*
> ```Less
>         // box-sizing: border-box;
>         // Also, you'd think you could use outline instead, but it doesn't show on the side; the
>         // children over it up because they are the full width of this parent.
> ```

over => cover. But I find that explanation hard to believe because by default an outline is OUTSIDE the box it belongs to. Might you instead be running up against a case where overflow is hidden? But that would affect the 100% width children, too. Maybe the issue is that you can't use an outline _with negative offset_ because the children cover it?

Note that outlining the bloom-editables will have a different effect on multilingual books. Does it do what you want there?

hatton (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6426)* status: all files reviewed, 1 unresolved discussion (waiting on @JohnThomson)

___
*[`src/BloomBrowserUI/bookEdit/css/editMode.less` line 1225 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6426#-Nvqht3e6vau0Mal2Q-w:-Nvr8HQ9CjBvuue6l4Zi:bbm3o3g) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/f0415b4a324b9679601663f8a30969284430bea6/src/BloomBrowserUI/bookEdit/css/editMode.less#L1225)):*
The outline approach doesn't work. I can't tell you why.


> Note that outlining the bloom-editables will have a different effect on multilingual books. Does it do what you want there?

Yes, the book I did this on was showing multilingual.

JohnThomson (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6426)* status: all files reviewed, 2 unresolved discussions (waiting on @hatton)

___
*[`src/BloomBrowserUI/bookEdit/css/editMode.less` line 1225 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6426#-Nvqht3e6vau0Mal2Q-w:-NvsUxrd1qtGeT8rhyBe:ba3w45p) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/f0415b4a324b9679601663f8a30969284430bea6/src/BloomBrowserUI/bookEdit/css/editMode.less#L1225)):*

The outline approach doesn't work. I can't tell you why.


> Note that outlining the bloom-editables will have a different effect on multilingual books. Does it do what you want there?

Yes, the book I did this on was showing multilingual.
</blockquote></details>

Well, if you don't know why putting an outline on TG doesn't work, I think you should just say that, rather than giving an explanation that I don't think is right. Or you could say "...but the outline is not visible where the TG is adjacent to the page margin [is that what you mean by "on the side"?]; possibly this is to do with the children covering it up because they are the full width of this parent."

However I did some experimenting and I think the real issue is that a couple of the parents (split-pane for example) have overflow:hidden, and since the default position of an outline is 1px outside the box, it gets clipped by the parent. So you would again need outline-offset:-1px. That makes outline work on a TG when I try it in an inspector. But you discovered that for the bloom-editables, so I'm surprised you didn't try it for the TG, too.

Anyway, if it works better on the bloom-editable, that's fine. But I think the comments need tweaking. I'm pretty sure the overflow:hidden is also one reason the bloom-editable needs outline-offset: -1px, too.
___
*[`src/BloomBrowserUI/bookEdit/css/editMode.less` line 1230 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6426#-NvsYHtiDyh_ZMJCezt2:-NvsYHtiDyh_ZMJCezt3:b-4j4234) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/f0415b4a324b9679601663f8a30969284430bea6/src/BloomBrowserUI/bookEdit/css/editMode.less#L1230)):*
> ```Less
>         .bloom-editable {
>             outline: 1px solid var(--page-structure-color);
>             outline-offset: -1px; // else the white of an un-focussed box can just bleed into the white of the margin
> ```

Also because various parents have overflow:hidden and some of the outline would otherwise get clipped?

hatton (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6426)* status: all files reviewed, 2 unresolved discussions (waiting on @JohnThomson)

___
*[`src/BloomBrowserUI/bookEdit/css/editMode.less` line 1225 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6426#-Nvqht3e6vau0Mal2Q-w:-Nvsc7ba-2483cLrHxhq:b-xm2baz) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/f0415b4a324b9679601663f8a30969284430bea6/src/BloomBrowserUI/bookEdit/css/editMode.less#L1225)):*

Well, if you don't know why putting an outline on TG doesn't work, I think you should just say that, rather than giving an explanation that I don't think is right. Or you could say "...but the outline is not visible where the TG is adjacent to the page margin [is that what you mean by "on the side"?]; possibly this is to do with the children covering it up because they are the full width of this parent."

However I did some experimenting and I think the real issue is that a couple of the parents (split-pane for example) have overflow:hidden, and since the default position of an outline is 1px outside the box, it gets clipped by the parent. So you would again need outline-offset:-1px. That makes outline work on a TG when I try it in an inspector. But you discovered that for the bloom-editables, so I'm surprised you didn't try it for the TG, too.

Anyway, if it works better on the bloom-editable, that's fine. But I think the comments need tweaking. I'm pretty sure the overflow:hidden is also one reason the bloom-editable needs outline-offset: -1px, too.
</blockquote></details>

I stand by my comment. Inset the children, and it will show. I think you're being awfully aggressive here over a comment something we aren't doing.

JohnThomson (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6426)* status: all files reviewed, 2 unresolved discussions (waiting on @hatton)

___
*[`src/BloomBrowserUI/bookEdit/css/editMode.less` line 1225 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6426#-Nvqht3e6vau0Mal2Q-w:-NwAVjJT3JTg_QYpHkS1:bqgooqn) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/f0415b4a324b9679601663f8a30969284430bea6/src/BloomBrowserUI/bookEdit/css/editMode.less#L1225)):*

I stand by my comment. Inset the children, and it will show. I think you're being awfully aggressive here over a comment something we aren't doing.
</blockquote></details>

OK, I finally realized that if you do an outline on the TG that is inset by a pixel, it gets overwritten by the focus box on the editable (because, as John's comment says, the editable is the full width of the TG) and this looks odd. I can live with John's comment.

PR #6454, opened 2024-04-29
A few more things that should be internationalized for 6.0 (BL-13352)

JohnThomson:


andrew-polk:
I am already fixing the book info text in [BL-13341](https://issues.bloomlibrary.org/youtrack/issue/BL-13341/Update-warning-messages-in-book-info-tip) with more complete solution of pulling out shared PWithLinks so they don't get out of sync again.

andrew-polk:
I've gone ahead and pushed the change to drop translate="no" directly.
And the other change is handled in [BL-13341](https://issues.bloomlibrary.org/youtrack/issue/BL-13341/Update-warning-messages-in-book-info-tip).

PR #6455, opened 2024-04-29
Prevent likely kinds of reentrant event handling during Save (BL-13120)

JohnThomson:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6455)

### DIFF
--- a/src/BloomExe/WebView2Browser.cs
+++ b/src/BloomExe/WebView2Browser.cs
@@ -20,6 +20,7 @@ using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
 using System.Windows.Forms;
+using SIL.Windows.Forms.ClearShare;
 
 namespace Bloom
 {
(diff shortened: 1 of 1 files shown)


hatton (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6455)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @JohnThomson)

___
*[`src/BloomExe/WebView2Browser.cs` line 549 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6455#-Nwfwm7K8lx7kGgHzPkI:-Nwfwm7L2m8VZH-mMz5i:b9vccc4) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/9642eec7e62d974dec39de40a1e68bb5b55cb44c/src/BloomExe/WebView2Browser.cs#L549)):*
> ```Smalltalk
>                 // that to indicate the completion of an async task. But it produced deadlock. I think this is enough to make it
>                 // pretty hard for the user to initiate something that will interfere with Save().
>                 return m.Msg == WM_TOUCH
> ```

If we threw an assert here, would that help to validate that this was the problem?

JohnThomson (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6455)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @hatton)

___
*[`src/BloomExe/WebView2Browser.cs` line 549 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6455#-Nwfwm7K8lx7kGgHzPkI:-NwkNawk9DujHeOtMiat:b-nwigsi) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/9642eec7e62d974dec39de40a1e68bb5b55cb44c/src/BloomExe/WebView2Browser.cs#L549)):*

If we threw an assert here, would that help to validate that this was the problem?
</blockquote></details>

I put one in. Not sure hitting it validates that this is the problem, though. I also tried introducing some extra delay so it became easy to hit the assert, and had it return false...and I hit it, but never subsequently got a reentrant call to SaveNow. If anyone encounters this for real while running the Javascript, it would be interesting to bypass the filter (return false) and see it it goes reentrant.

PR #6456, opened 2024-04-30
BL-13379 For Discussion Save via callback

hatton:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6456)


hatton:
See https://issues.bloomlibrary.org/youtrack/issue/BL-13379

PR #6467, opened 2024-05-03
Throw if we try to navigate to a new page while saving (BL-13120)

JohnThomson:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6467)

### DIFF
--- a/src/BloomExe/Edit/EditingModel.cs
+++ b/src/BloomExe/Edit/EditingModel.cs
@@ -59,7 +59,7 @@ namespace Bloom.Edit
         internal event EventHandler PageSelectModelChangesComplete;
 
         // These variables are not thread-safe. Access only on UI thread.
-        private bool _inProcessOfSaving;
+        internal bool InProcessOfSaving;
         private List<Action> _tasksToDoAfterSaving = new List<Action>();
 
         // Perhaps a bit hack-ish, but this causes a full save to be done when our datadiv has been modified
@@ -340,9 +340,9 @@ namespace Bloom.Edit
         internal void DeletePage(IPage page)
         {
             // This can only be called on the UI thread in response to a user button click.
-            // If that ever changed we might need to arrange locking for access to _inProcessOfSaving and _tasksToDoAfterSaving.
+            // If that ever changed we might need to arrange locking for access to InProcessOfSaving and _tasksToDoAfterSaving.
             Debug.Assert(!_view.InvokeRequired);
-            if (_inProcessOfSaving && page == _pageSelection.CurrentSelection)
+            if (InProcessOfSaving && page == _pageSelection.CurrentSelection)
             {
                 // Somehow (BL-431) it's possible that a Save is still in progress when we start executing a delete page.
                 // If this happens, to prevent crashes we need to let the Save complete before we go ahead with the delete.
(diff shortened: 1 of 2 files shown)


andrew-polk (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6467)* status: all files reviewed, 1 unresolved discussion (waiting on @JohnThomson)

___
*[`src/BloomExe/Edit/EditingView.cs` line 471 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6467#-NwzI2HR0EIJwJFy-z4x:-NwzI2HR0EIJwJFy-z4y:b-blnp1w) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/28f70102ea7d7587bd4072a21fd332e74021151e/src/BloomExe/Edit/EditingView.cs#L471)):*
> ```Smalltalk
>                         "UpdateSingleDisplayedPage called while still in the process of saving a page. Changes to the page may not get saved. Please report and connect to BL-13120."
>                     );
>                 }
> ```

Do we want this gated for DEBUG? 
If not, where are we capturing all the things we want to undo once we finally solve this?

JohnThomson (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6467)* status: all files reviewed, 1 unresolved discussion (waiting on @andrew-polk)

___
*[`src/BloomExe/Edit/EditingView.cs` line 471 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6467#-NwzI2HR0EIJwJFy-z4x:-NwzJfal-dRHGXiwvE2q:bwaaeox) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/28f70102ea7d7587bd4072a21fd332e74021151e/src/BloomExe/Edit/EditingView.cs#L471)):*

Do we want this gated for DEBUG? 
If not, where are we capturing all the things we want to undo once we finally solve this?
</blockquote></details>

In our Zulip discussion, JohnH said "that could stay in even when not in debug mode". I don't even see much need to remove it when we fully solve the problem. It's not an expensive check, and if any future situation causes us to navigate while trying to save, I think we'd like to know.
But I suppose we could start inserting comments like // Hack to try to catch BL-13120. In this case, it's in the message, so just searching for BL-13120 would find this.

andrew-polk (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6467)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @JohnThomson)

___
*[`src/BloomExe/Edit/EditingView.cs` line 471 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6467#-NwzI2HR0EIJwJFy-z4x:-NwzQNXu295nYPgs7WWO:bg413lb) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/28f70102ea7d7587bd4072a21fd332e74021151e/src/BloomExe/Edit/EditingView.cs#L471)):*

In our Zulip discussion, JohnH said "that could stay in even when not in debug mode". I don't even see much need to remove it when we fully solve the problem. It's not an expensive check, and if any future situation causes us to navigate while trying to save, I think we'd like to know.
But I suppose we could start inserting comments like // Hack to try to catch BL-13120. In this case, it's in the message, so just searching for BL-13120 would find this.
</blockquote></details>

Discussed further in zulip.

PR #6469, opened 2024-05-06
Fix BL-13387 Prevent double clicking image buttons

hatton:
The minimal fix here is just the one in bloomImages.ts.
However, looking at the rest of the code, it seems we were needlessly doing dangerous things to simulate synchronous behavior when it wasn't needed, so I removed that too.

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6469)

### DIFF
--- a/src/BloomBrowserUI/bookEdit/editablePage.ts
+++ b/src/BloomBrowserUI/bookEdit/editablePage.ts
@@ -50,7 +50,8 @@ import {
     pasteClipboardText,
     makeElement,
     SetupElements,
-    attachToCkEditor
+    attachToCkEditor,
+    changeImage
 } from "./js/bloomEditing";
 export {
     pageSelectionChanging,
@@ -63,7 +64,8 @@ export {
     pasteClipboardText,
     makeElement,
     SetupElements,
-    attachToCkEditor
+    attachToCkEditor,
+    changeImage
 };
 import { origamiCanUndo, origamiUndo } from "./js/origami";
 export { origamiCanUndo, origamiUndo };
--- a/src/BloomBrowserUI/bookEdit/js/bloomEditing.ts
+++ b/src/BloomBrowserUI/bookEdit/js/bloomEditing.ts
@@ -417,7 +417,6 @@ export function SetupElements(container: HTMLElement) {
         target.setAttribute("data-copyright", event.copyright);
         target.setAttribute("data-creator", event.creator);
         target.setAttribute("data-license", event.license);
-        post("edit/taskComplete");
     });
     SetupVideoEditing(container);
     SetupWidgetEditing(container);
--- a/src/BloomBrowserUI/bookEdit/js/bloomEditing.ts
+++ b/src/BloomBrowserUI/bookEdit/js/bloomEditing.ts
@@ -388,6 +388,34 @@ export function SetupThingsSensitiveToStyleChanges(container: HTMLElement) {
         });
 }
 
+export function changeImage(imageInfo: {
+    imageIndex: number;
+    src: string; // must already appropriately URL-encoded.
+    copyright: string;
+    creator: string;
+    license: string;
+}) {
+    const container = Array.from(
+        document.getElementsByClassName("bloom-imageContainer")
+    )[imageInfo.imageIndex];
+
+    // I can't remember why, but what this is doing is saying that if the imageContainer
+    // has an <img> element, we're setting the src on that. But if it does not, we're
+    // setting the background-image on the container itself.
+    const img = container.getElementsByTagName("img")[0];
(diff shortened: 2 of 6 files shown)


hatton:
Reviewers: Note that I removed some apparently needless and dangerous code that was attempting to delay saving the image until the Browser had loaded it. I couldn't figure out what this would be.  So I removed it.

JohnThomson (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6469)* status: all files reviewed, 1 unresolved discussion (waiting on @hatton)

___
*[`src/BloomExe/Edit/PageEditingModel.cs` line 82 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6469#-NxEsKos1KvsA7kto_Ms:-NxEsKos1KvsA7kto_Mt:b-k5asi4) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/cb14f5a0a3e47cd8e63caa182c6b8046d5342bc6/src/BloomExe/Edit/PageEditingModel.cs#L82)):*
> ```Smalltalk
>             messageBundle.creator = imageInfo.Metadata.Creator ?? "";
>             messageBundle.license = imageInfo.Metadata.License?.ToString() ?? "";
>             webSocketServer.SendBundle("edit", "changeImage", messageBundle);
> ```

After this method returns, the calling code in EditingModel saves the page. I think we need some guarantee that sending this bundle has updated the page before we save it, unless we can make a more radical change like initiating the save from the changeImage handler.

hatton (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6469)* status: 2 of 6 files reviewed, 1 unresolved discussion (waiting on @JohnThomson)

___
*[`src/BloomExe/Edit/PageEditingModel.cs` line 82 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6469#-NxEsKos1KvsA7kto_Ms:-NxF9sbg7ECrNHpTpE0u:bf6mk50) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/cb14f5a0a3e47cd8e63caa182c6b8046d5342bc6/src/BloomExe/Edit/PageEditingModel.cs#L82)):*

After this method returns, the calling code in EditingModel saves the page. I think we need some guarantee that sending this bundle has updated the page before we save it, unless we can make a more radical change like initiating the save from the changeImage handler.
</blockquote></details>

Thanks. Changed to use `RunJavascriptAsync()` to remove question around whether the websocket msg would be handled before the subsequent  `RunJavascriptAsync()` of the Save

JohnThomson (approved):
> Reviewed 5 of 5 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6469)* status: all files reviewed, 2 unresolved discussions (waiting on @hatton)

___
*[`src/BloomExe/Edit/PageEditingModel.cs` line 91 at r4](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6469#-NxNt4XS9dE4j3c-j-e6:-NxNt4XS9dE4j3c-j-e7:b3yhm9c) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/c543871de98d1cc23d4b226da51dc695b8f1600b/src/BloomExe/Edit/PageEditingModel.cs#L91)):*
> ```Smalltalk
>             browser.RunJavascriptAsync(
>                 $"editTabBundle.getEditablePageBundleExports().changeImage({s})"
>             );
> ```

I've started putting in the pragma to turn off the warning for running async methods without await, in cases like this where it's been thought through carefully.

hatton (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6469)* status: all files reviewed, 2 unresolved discussions (waiting on @JohnThomson)

___
*[`src/BloomExe/Edit/PageEditingModel.cs` line 91 at r4](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6469#-NxNt4XS9dE4j3c-j-e6:-NxO1Ml0Ewn0L6aa2hhz:bkx5cuf) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/c543871de98d1cc23d4b226da51dc695b8f1600b/src/BloomExe/Edit/PageEditingModel.cs#L91)):*

I've started putting in the pragma to turn off the warning for running async methods without await, in cases like this where it's been thought through carefully.
</blockquote></details>

in my version of the code, it returns`Task`, and I don't get any warning. I think I'll go ahead and merge and we'll hide this warning eventually if my version is just behind somehow.

PR #6471, opened 2024-05-08
Sleep while javascript calls us back on another thread (BL-13120)

JohnThomson:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6471)

### DIFF
--- a/src/BloomBrowserUI/bookEdit/editablePage.ts
+++ b/src/BloomBrowserUI/bookEdit/editablePage.ts
@@ -13,7 +13,6 @@ import { theOneBubbleManager, BubbleManager } from "./js/bubbleManager";
 export interface IPageFrameExports {
     pageSelectionChanging(): void;
     pageUnloading(): void;
-    disconnectForGarbageCollection(): void;
     copySelection(): void;
     cutSelection(): void;
     pasteClipboardText(): void;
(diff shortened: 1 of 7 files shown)


hatton (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6471)* status: all files reviewed, 1 unresolved discussion (waiting on @JohnThomson)

___
*[`src/BloomExe/WebView2Browser.cs` line 579 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6471#-NxO3k6l1mf-WuJvw1Zk:-NxO3k6l1mf-WuJvw1Zl:bvm6ufs) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/1e42f2a3dbe7cfd27e037b00b48d7bacd56989f8/src/BloomExe/WebView2Browser.cs#L579)):*
> ```Smalltalk
>                 );
>                 throw new TimeoutException(
>                     "RunJavascriptThatPostsStringResultSync: Timed out waiting for script to complete"
> ```

If we get a report of this, will we wish we'd put the script into the log or the error. Or maybe there will be only one call per c# method, so the call stack will tell us all we want to know?

JohnThomson (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6471)* status: all files reviewed, 1 unresolved discussion (waiting on @hatton)

___
*[`src/BloomExe/WebView2Browser.cs` line 579 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6471#-NxO3k6l1mf-WuJvw1Zk:-NxO4mWb93o7grUDM202:b-5qltsx) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/1e42f2a3dbe7cfd27e037b00b48d7bacd56989f8/src/BloomExe/WebView2Browser.cs#L579)):*

If we get a report of this, will we wish we'd put the script into the log or the error. Or maybe there will be only one call per c# method, so the call stack will tell us all we want to know?
</blockquote></details>

At this point, there is only one caller in our whole code base. I think we can hold off on adding that.

PR #6479, opened 2024-05-09
Improvements made while investigating page wipeout (BL-13120)

andrew-polk:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6479)

### DIFF
--- a/src/BloomBrowserUI/bookEdit/editViewFrame.ts
+++ b/src/BloomBrowserUI/bookEdit/editViewFrame.ts
@@ -93,6 +93,7 @@ export function switchContentPage(newSource: string) {
     // The normal way to deal with this is to wait and call the function after the new
     // page is loaded. But sometimes (see comment below) the load event doesn't get
     // raised, so we have a fall-back.
+    // Enhance: Should this use DOMContentLoaded instead of load?
     let handlerCalled = false;
     const handler = () => {
         handlerCalled = true;
@@ -108,7 +109,6 @@ export function switchContentPage(newSource: string) {
     // other cases since I have no explanation. Rather than leaving control state
     // permanently wrong, if the event is delayed much longer than expected we just
     // call the handler.
-    // For a similar event not firing problem, see bloomEditing.ts.
     window.setTimeout(() => {
         if (!handlerCalled) {
             handler();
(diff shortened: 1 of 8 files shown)


andrew-polk (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6479)* status: 0 of 8 files reviewed, all discussions resolved

___
*[`src/BloomBrowserUI/bookEdit/js/bloomEditing.ts` line 338 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6479#-NxUid4SDl4g3wnqrn-5:-NxUiCTK6ShRJohPnzg5:bl4mxn2) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/d30989c3bfd7729beb0f516fcd4349c6e506bcc4/src/BloomBrowserUI/bookEdit/js/bloomEditing.ts#L338)):*
> ```XML
> document.addEventListener("DOMContentLoaded", () =>
>     post("editView/pageDomLoaded")
> );
> ```

This simplification has two main implications:
1. We no longer wait for load and then one animation frame. I.e. we only want the dom load, not painting.
2. We no longer guarantee we will call this. The old code would always call after 1 second if not before.
___
*[`src/BloomExe/Edit/EditingView.cs` line 521 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6479#-NxUiMuN4JpCmUI-rrTh:-NxUiMuN4JpCmUI-rrTi:bo3d042) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/d30989c3bfd7729beb0f516fcd4349c6e506bcc4/src/BloomExe/Edit/EditingView.cs#L521)):*
> ```Smalltalk
>                         Logger.WriteEvent("changing page via editTabBundle.switchContentPage()");
>                         var pageUrl = _model.GetUrlForCurrentPage();
>                         _browser1.RunJavascriptFireAndForget(
> ```

This is the most dangerous change here.
We discussed that this could lead to race conditions. But logically, we don't think anything should need to wait for this.
We looked through many of the callers but not all.

I can continue to investigate more if desired.

PR #6485, opened 2024-05-11
BL-13454 Open *.BloomProblemBooks

hatton:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6485)


andrew-polk (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6485)* status: all files reviewed, 3 unresolved discussions (waiting on @hatton)

___
*[`src/BloomExe/Program.cs` line 572 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6485#-NxmvUWt5ITSYRQm-bsW:-NxmvUWt5ITSYRQm-bsX:b-cod79g) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/2fe9464bc6de2b7bf1a42f1922a22a882653df01/src/BloomExe/Program.cs#L572)):*
> ```Smalltalk
>                                     .FirstOrDefault();
>                                 Settings.Default.MruProjects.AddNewPath(collectionPath);
>                             }
> ```

I guess we're just going to let these accumulate in the temp folder?

I guess we're going to live with whatever bad thing happens if we try to open an unzipped collection which *has* been cleaned out of temp?

Worth commenting about either or both of those?
___
*[`src/BloomExe/Collection/CollectionSettings.cs` line 494 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6485#-Nxmt2Xp0PHSAqC9bhA9:-Nxmt2Xp0PHSAqC9bhAA:brs9rc8) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/2fe9464bc6de2b7bf1a42f1922a22a882653df01/src/BloomExe/Collection/CollectionSettings.cs#L494)):*
> ```Smalltalk
>                         // even if the subscription has expired.
>                         // 2) this is a developer looking into a Bloom Problem Report.
>                         var downloadInfoPath = RobustFile.Exists(downloadEditPath)
> ```

We've already set 
`LockedToOneDownloadedBook = RobustFile.Exists(downloadEditPath);` 
above.
So this can use that property instead of checking for the file existence again.
___
*[`src/BloomExe/Collection/CollectionSettings.cs` line 516 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6485#-Nxmtv1a5YxQQpxCY_7Y:-Nxmtv1a5YxQQpxCY_7Z:bggipmo) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/2fe9464bc6de2b7bf1a42f1922a22a882653df01/src/BloomExe/Collection/CollectionSettings.cs#L516)):*
> ```Smalltalk
>                         else
>                         {
>                             BrandingProjectKey = "Default"; // keep the code, but don't use it as active branding.
> ```

I think the comment would make more sense as something like

```
// We've stored the original code in InvalidBranding;
// but we need to not use it as the active branding since it is expired.
BrandingProjectKey = "Default";
```

andrew-polk (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6485)* status: all files reviewed, 2 unresolved discussions (waiting on @hatton)

___
*[`src/BloomExe/Program.cs` line 572 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6485#-NxmvUWt5ITSYRQm-bsW:-Nxx57J62f6XUifvWeny:b-qd8swg) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/2fe9464bc6de2b7bf1a42f1922a22a882653df01/src/BloomExe/Program.cs#L572)):*

Yes, this is all temp things.
Sorry, I'm not getting what you're asking for.
</blockquote></details>

It might be worth documenting the decisions we made:
1. Yes, these temp folders will accumulate on dev machines and never get cleaned up.
2. Yes, if one does get cleaned up, and we try to open it, bad things will happen. (Maybe this isn't actually a problem since we don't list items in mru which don't exist.)
___
*[`src/BloomExe/Collection/CollectionSettings.cs` line 494 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6485#-Nxmt2Xp0PHSAqC9bhA9:-Nxx3D4sDZX4p2UOtMSK:bixc1tz) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/2fe9464bc6de2b7bf1a42f1922a22a882653df01/src/BloomExe/Collection/CollectionSettings.cs#L494)):*

Maybe, but in the case of opening a BloomProblemBook, we are not actually doing that (locking down), so that would be confusing code to me. This is very clear, and what performance are we aiming for in this code?
</blockquote></details>

Unless I'm missing something, this check *is* saying: if we are LockedToOneDownloadedBook, use that path, else see if there is a problem book path. 

But if you think this is clearer, that's fine.

hatton (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6485)* status: all files reviewed, 2 unresolved discussions (waiting on @andrew-polk)

___
*[`src/BloomExe/Program.cs` line 572 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6485#-NxmvUWt5ITSYRQm-bsW:-NxxfYhG1RAkkuqCD52l:b-65c2nj) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/2fe9464bc6de2b7bf1a42f1922a22a882653df01/src/BloomExe/Program.cs#L572)):*

It might be worth documenting the decisions we made:
1. Yes, these temp folders will accumulate on dev machines and never get cleaned up.
2. Yes, if one does get cleaned up, and we try to open it, bad things will happen. (Maybe this isn't actually a problem since we don't list items in mru which don't exist.)
</blockquote></details>

I've added them over in ProblemReportApi.UnpackProblemBook(path) so we can stop wasting time on this PR.
___
*[`src/BloomExe/Collection/CollectionSettings.cs` line 494 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6485#-Nxmt2Xp0PHSAqC9bhA9:-NxxfJiL-3HLMPaXfkb_:b-lyb87q) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/2fe9464bc6de2b7bf1a42f1922a22a882653df01/src/BloomExe/Collection/CollectionSettings.cs#L494)):*

Unless I'm missing something, this check *is* saying: if we are LockedToOneDownloadedBook, use that path, else see if there is a problem book path. 

But if you think this is clearer, that's fine.
</blockquote></details>

You're right. And yes I think this is clear.

andrew-polk (approved):
> Reviewed 1 of 1 files at r3, 1 of 1 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6485)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @hatton)

PR #6490, opened 2024-05-13
BL-13401 don't rename audio to existing filenames

nabalone:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6490)

### DIFF
--- a/src/BloomExe/Publish/Video/RecordVideoWindow.cs
+++ b/src/BloomExe/Publish/Video/RecordVideoWindow.cs
@@ -645,12 +645,21 @@ namespace Bloom.Publish.Video
             var workingDirectory = Path.GetDirectoryName(soundLog[0].src);
             LocalAudioNamesMessedUp = true;
             var renames = new Dictionary<string, string>();
+            int shortNameIndex = 0;
             for (int i = 0; i < soundLog.Length; i++)
             {
                 var item = soundLog[i];
                 if (Path.GetDirectoryName(item.src) == workingDirectory)
                 {
-                    var shortName = GetShortName(i) + Path.GetExtension(item.src);
+                    string shortName = "";
+                    while (
+                        shortName == ""
+                        || RobustFile.Exists(Path.Combine(workingDirectory, shortName))
+                    )
+                    {
+                        shortName = GetShortName(shortNameIndex) + Path.GetExtension(item.src);
+                        shortNameIndex++;
+                    }
                     if (renames.TryGetValue(item.src, out string prevShortName))
                     {
                         // If we already saw (and renamed) this item, just use what we already renamed it to.


andrew-polk (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6490)* status: all files reviewed, 1 unresolved discussion (waiting on @nabalone)

___
*[`src/BloomExe/Publish/Video/RecordVideoWindow.cs` line 658 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6490#-NxoDLs_7m8EVCjgM0tu:-NxoDLs_7m8EVCjgM0tv:bl0f015) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/4eec0cb2753b2b951051987f6c87c61de8f1861c/src/BloomExe/Publish/Video/RecordVideoWindow.cs#L658)):*
> ```Smalltalk
>                         shortName == ""
>                         || RobustFile.Exists(Path.Combine(workingDirectory, shortName))
>                     )
> ```

Given that we've had to optimize this code for the possibility of a few thousand files, I'm a bit hesitant to do an IO check for every file.
How easy it is to create a regex which would find any existing file which *might* be a conflict and prefilter before doing the IO check?

Another option would be to create a list of all the existing files before starting and check against that list. The list could be narrowed by the regex.

But the first idea seems like it would be the least amount of IO. We don't have to do anything unless the file *might* match a name we could use as a short name. Doesn't seem like that would happen very often.

However, I admit I don't always have the best head for performance things. 
Thoughts, @hatton @StephenMcConnel?

nabalone:
*[`src/BloomExe/Publish/Video/RecordVideoWindow.cs` line 658 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6490#-NxoDLs_7m8EVCjgM0tu:-NxrN-MA8QHyMTZzQl-8:b-tcso4c) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/4eec0cb2753b2b951051987f6c87c61de8f1861c/src/BloomExe/Publish/Video/RecordVideoWindow.cs#L658)):*

Given that we've had to optimize this code for the possibility of a few thousand files, I'm a bit hesitant to do an IO check for every file.
How easy it is to create a regex which would find any existing file which *might* be a conflict and prefilter before doing the IO check?

Another option would be to create a list of all the existing files before starting and check against that list. The list could be narrowed by the regex.

But the first idea seems like it would be the least amount of IO. We don't have to do anything unless the file *might* match a name we could use as a short name. Doesn't seem like that would happen very often.

However, I admit I don't always have the best head for performance things. 
Thoughts, @hatton @StephenMcConnel?
</blockquote></details>

If we use `Directory.GetFiles` it looks like we have the option of `*` and `?` wildcards. I was thinking we could calculate approximately how long the shortName will get for this situation, and then do `Directory.GetFiles(workingDirectory, "?.*`, `Directory.GetFiles(workingDirectory, "??.*`, `Directory.GetFiles(workingDirectory, "???.*`, etc. Is that better or worse than just get all files in this working directory? Or is there another way to do this?

nabalone (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6490)* status: all files reviewed, 1 unresolved discussion (waiting on @hatton and @StephenMcConnel)

___
*[`src/BloomExe/Publish/Video/RecordVideoWindow.cs` line 658 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6490#-NxoDLs_7m8EVCjgM0tu:-Nxrtb712ZTMB78b2oK6:b-lf3i0h) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/4eec0cb2753b2b951051987f6c87c61de8f1861c/src/BloomExe/Publish/Video/RecordVideoWindow.cs#L658)):*

If we use `Directory.GetFiles` it looks like we have the option of `*` and `?` wildcards. I was thinking we could calculate approximately how long the shortName will get for this situation, and then do `Directory.GetFiles(workingDirectory, "?.*`, `Directory.GetFiles(workingDirectory, "??.*`, `Directory.GetFiles(workingDirectory, "???.*`, etc. Is that better or worse than just get all files in this working directory? Or is there another way to do this?
</blockquote></details>

https://learn.microsoft.com/en-us/dotnet/api/system.io.directory.getfiles?view=net-8.0

andrew-polk (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6490)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @nabalone)

___
*[`src/BloomExe/Publish/Video/RecordVideoWindow.cs` line 658 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6490#-NxoDLs_7m8EVCjgM0tu:-NxrtguDA7BRyig96XDE:bvru35i) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/4eec0cb2753b2b951051987f6c87c61de8f1861c/src/BloomExe/Publish/Video/RecordVideoWindow.cs#L658)):*

https://learn.microsoft.com/en-us/dotnet/api/system.io.directory.getfiles?view=net-8.0
</blockquote></details>

As John and I discussed this, we decided that file existence is probably pretty fast, and it probably isn't worth adding complexity.

PR #6491, opened 2024-05-13
Fix previews of sample shells, including downloads (BL-13445)

StephenMcConnel:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6491)


andrew-polk (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6491)* status: all files reviewed, 1 unresolved discussion (waiting on @StephenMcConnel)

___
*[`src/BloomExe/Book/Book.cs` line 934 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6491#-NxoLPPPBwNHXL63Z2cQ:-NxoLPPQC32H-yZWbAqm:bc2tur3) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/abc132f8f7e691615b8f212f35cb0feb44e04ce2/src/BloomExe/Book/Book.cs#L934)):*
> ```Smalltalk
>             // are previewing a book we can't EnsureUpToDate().
>             var previewDom = Storage.GetRelocatableCopyOfDom(false);
>             foreach (XmlElement meta in previewDom.Head.GetElementsByTagName("meta"))
> ```

I would have thought the answer was rather to ensure a link to appearance.css. (And you previously added code such that, for previews, if that file doesn't exist, it will return appearance-theme-default.css.)

Of course, for the Sample Shells themselves, you could just upgrade them to actually use the default theme. And that seems even better.

StephenMcConnel (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6491)* status: 1 of 8 files reviewed, 1 unresolved discussion (waiting on @andrew-polk)

___
*[`src/BloomExe/Book/Book.cs` line 934 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6491#-NxoLPPPBwNHXL63Z2cQ:-NxwziL8CPS2e9bwV9g-:bqcy9f5) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/abc132f8f7e691615b8f212f35cb0feb44e04ce2/src/BloomExe/Book/Book.cs#L934)):*

I would have thought the answer was rather to ensure a link to appearance.css. (And you previously added code such that, for previews, if that file doesn't exist, it will return appearance-theme-default.css.)

Of course, for the Sample Shells themselves, you could just upgrade them to actually use the default theme. And that seems even better.
</blockquote></details>

I agree that upgrading the Sample Shells is the right answer, and the HTML files will then contain the link to appearance.css.
However, for older (than 6.0) downloaded books, many of them will have custom CSS that is incompatible with the appearance system.  I think it's safer to use the legacy basePage than to assume that adding appearance.css will make things look okay.

andrew-polk (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6491)* status: all files reviewed, 1 unresolved discussion (waiting on @StephenMcConnel)

___
*[`src/BloomExe/Book/Book.cs` line 934 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6491#-NxoLPPPBwNHXL63Z2cQ:-Nxx9ZGJBU6gpQa4qQEL:b66h6az) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/abc132f8f7e691615b8f212f35cb0feb44e04ce2/src/BloomExe/Book/Book.cs#L934)):*

I agree that upgrading the Sample Shells is the right answer, and the HTML files will then contain the link to appearance.css.
However, for older (than 6.0) downloaded books, many of them will have custom CSS that is incompatible with the appearance system.  I think it's safer to use the legacy basePage than to assume that adding appearance.css will make things look okay.
</blockquote></details>

Hm. This actually makes me realize there is a more meta question here.
The comment a few lines up indicates our intention is just to show the book as it exists on disk with its actual on-disk stylesheets.

I see that for Sample Shells, the folder doesn't actually contain all the relevant stylesheets (like basePage.css), so I do think upgrading them is required as you've done.

However, for source books, it seems that they should be displaying exactly as they would if you opened them from the .htm file directly in a browser. So is that the real bug? That we aren't using the stylesheets in the source books' folders for preview?

andrew-polk (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6491)* status: all files reviewed, 1 unresolved discussion (waiting on @StephenMcConnel)

___
*[`src/BloomExe/Book/Book.cs` line 934 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6491#-NxoLPPPBwNHXL63Z2cQ:-NxxByzA9HpMxIODia9A:bdyao4h) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/abc132f8f7e691615b8f212f35cb0feb44e04ce2/src/BloomExe/Book/Book.cs#L934)):*

Hm. This actually makes me realize there is a more meta question here.
The comment a few lines up indicates our intention is just to show the book as it exists on disk with its actual on-disk stylesheets.

I see that for Sample Shells, the folder doesn't actually contain all the relevant stylesheets (like basePage.css), so I do think upgrading them is required as you've done.

However, for source books, it seems that they should be displaying exactly as they would if you opened them from the .htm file directly in a browser. So is that the real bug? That we aren't using the stylesheets in the source books' folders for preview?
</blockquote></details>

Actually, I haven't been able to reproduce this issue with downloaded books. Is it possible you just have some corrupt books?

StephenMcConnel (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6491)* status: 3 of 29 files reviewed, 1 unresolved discussion (waiting on @andrew-polk)

___
*[`src/BloomExe/Book/Book.cs` line 934 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6491#-NxoLPPPBwNHXL63Z2cQ:-NxyFtEVDayj_TjbIBNT:b-yhqlyd) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/abc132f8f7e691615b8f212f35cb0feb44e04ce2/src/BloomExe/Book/Book.cs#L934)):*

Actually, I haven't been able to reproduce this issue with downloaded books. Is it possible you just have some corrupt books?
</blockquote></details>

The books weren't corrupt, but Bloom was removing langVisibility.css from preview books when it shouldn't.  The basic fix for downloaded books was to prevent any tampering with the stylesheet links (except possibly ordering?).
I also updated the sample shells to 6.0 and added in the missing css files.

andrew-polk (changes requested):
> Reviewed 6 of 26 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6491)* status: 9 of 29 files reviewed, 3 unresolved discussions (waiting on @StephenMcConnel)

___
*[`src/BloomExe/Book/HtmlDom.cs` line 499 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6491#-NxyKKIG0d33zlLwepXv:-NxyKKIG0d33zlLwepXw:b-yvscvx) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/bb2c1f1398fd68527f818eb4e27ff30b749b3aed/src/BloomExe/Book/HtmlDom.cs#L499)):*
> ```Smalltalk
>                 {
>                     Debug.WriteLine($"HtmlDom.SortStyleSheetLinks(): Skipping obsolete css file: {href}");
>                 }
> ```

Doesn't this Debug.WriteLine only apply when `BookStorage.CssFilesThatAreObsolete.Contains(href)` is true?
Currently, it will be printed any time `PreserveExistingStylesheets` is false.

(easiest fix is to just remove it)
___
*[`src/content/templates/Sample Shells/The Moon and the Cap/The Moon and the Cap.htm` line 261 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6491#-NxyJ2Fn5AoDw2NLFVfi:-NxyJ2Fn5AoDw2NLFVfj:bz461ci) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/bb2c1f1398fd68527f818eb4e27ff30b749b3aed/src/content/templates/Sample%20Shells/The%20Moon%20and%20the%20Cap/The%20Moon%20and%20the%20Cap.htm#L261)):*
> ```HTML
>             <div class="copyright Credits-Page-style" data-derived="originalCopyrightAndLicense" lang="*">
>                 This book is an adaptation of the original, <cite data-book="originalTitle">The Moon and the Cap</cite>, Copyright © 2007, Pratham Books. Licensed under CC BY 4.0.
>             </div>
> ```

I think something went wrong with the most recent update of the Sample Books. Both now have "This book is an adaptation..." Copyright moved to OriginalCopyright, etc.
I.e. it appears to have been converted to a derivative of the original books.
___
*[`src/BloomExe/Book/Book.cs` line 934 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6491#-NxoLPPPBwNHXL63Z2cQ:-NxyJNFY34We_xp7NJyF:b-ezpvbt) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/abc132f8f7e691615b8f212f35cb0feb44e04ce2/src/BloomExe/Book/Book.cs#L934)):*

The books weren't corrupt, but Bloom was removing langVisibility.css from preview books when it shouldn't.  The basic fix for downloaded books was to prevent any tampering with the stylesheet links (except possibly ordering?).
I also updated the sample shells to 6.0 and added in the missing css files.
</blockquote></details>

I'm sorry; I may have miscommunicated. I was not asking you to add the support files to the Sample Shells but affirming that what you had already done to them was necessary.
I think we can continue to let the magic work to pull in the correct version of the support files instead of adding them to the repo?

StephenMcConnel (commented):
I've added the fixes for BL-13444 to this pull request since they all involve updating the HTML file for the sample book, which is also involved in the fix for BL-13445.

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6491#-:-Ny1ewWwDkSaIcsv3Ofa:bwnxoln)* status: 5 of 30 files reviewed, 3 unresolved discussions (waiting on @andrew-polk)

___
*[`src/BloomExe/Book/HtmlDom.cs` line 499 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6491#-NxyKKIG0d33zlLwepXv:-Ny1eQvz0MhN-Ouqxo7T:b8mwtas) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/bb2c1f1398fd68527f818eb4e27ff30b749b3aed/src/BloomExe/Book/HtmlDom.cs#L499)):*

Doesn't this Debug.WriteLine only apply when `BookStorage.CssFilesThatAreObsolete.Contains(href)` is true?
Currently, it will be printed any time `PreserveExistingStylesheets` is false.

(easiest fix is to just remove it)
</blockquote></details>

Done.  But I think you were misinterpreting the logic.  The debugging statement isn't really needed any longer.
___
*[`src/content/templates/Sample Shells/The Moon and the Cap/The Moon and the Cap.htm` line 261 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6491#-NxyJ2Fn5AoDw2NLFVfi:-Ny1eghF3skmJkUa2RNI:bbml3og) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/bb2c1f1398fd68527f818eb4e27ff30b749b3aed/src/content/templates/Sample%20Shells/The%20Moon%20and%20the%20Cap/The%20Moon%20and%20the%20Cap.htm#L261)):*

I think something went wrong with the most recent update of the Sample Books. Both now have "This book is an adaptation..." Copyright moved to OriginalCopyright, etc.
I.e. it appears to have been converted to a derivative of the original books.
</blockquote></details>

I think I've gotten this fixed now.
___
*[`src/BloomExe/Book/Book.cs` line 934 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6491#-NxoLPPPBwNHXL63Z2cQ:-Ny1epekB-76GpiedGl7:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/abc132f8f7e691615b8f212f35cb0feb44e04ce2/src/BloomExe/Book/Book.cs#L934)):*

I'm sorry; I may have miscommunicated. I was not asking you to add the support files to the Sample Shells but affirming that what you had already done to them was necessary.
I think we can continue to let the magic work to pull in the correct version of the support files instead of adding them to the repo?
</blockquote></details>

Done.

andrew-polk (approved):
> Reviewed 2 of 26 files at r3, 23 of 23 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6491)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @StephenMcConnel)

PR #6495, opened 2024-05-16
BL-13401 fix dont rename audio file to existing filename

nabalone:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6495)

### DIFF
--- a/src/BloomExe/Publish/Video/RecordVideoWindow.cs
+++ b/src/BloomExe/Publish/Video/RecordVideoWindow.cs
@@ -646,6 +646,8 @@ namespace Bloom.Publish.Video
             LocalAudioNamesMessedUp = true;
             var renames = new Dictionary<string, string>();
             int shortNameIndex = 0;
+            string[] origAudioFileNames = Directory.GetFiles(workingDirectory);
+
             for (int i = 0; i < soundLog.Length; i++)
             {
                 var item = soundLog[i];
@@ -654,7 +656,7 @@ namespace Bloom.Publish.Video
                     string shortName = "";
                     while (
                         shortName == ""
-                        || RobustFile.Exists(Path.Combine(workingDirectory, shortName))
+                        || origAudioFileNames.Contains(Path.Combine(workingDirectory, shortName))
                     )
                     {
                         shortName = GetShortName(shortNameIndex) + Path.GetExtension(item.src);


nabalone (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6495)* status: 0 of 1 files reviewed, 2 unresolved discussions

___
*[`src/BloomExe/Publish/Video/RecordVideoWindow.cs` line 649 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6495#-Nxz1Jq7Aa-bityo2iT2:-Nxz1Jq83TOLsBMnyvrD:b-6qcvg2) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/5da078b2dc7a95f07ae3f4c4ed2bb2f3521e09e8/src/BloomExe/Publish/Video/RecordVideoWindow.cs#L649)):*
> ```Smalltalk
>             var renames = new Dictionary<string, string>();
>             int shortNameIndex = 0;
>             string[] origAudioFileNames = Directory.GetFiles(workingDirectory);
> ```

Would it be more efficient to combine `Directory.GetFiles(workingDirectory, "?.*")`, `Directory.GetFiles(workingDirectory, "??.*")`, `Directory.GetFiles(workingDirectory, "???.*")` etc. (depending on the number of audio files)? 
___
*[`src/BloomExe/Publish/Video/RecordVideoWindow.cs` line 659 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6495#-Nxz20KQ0kcOdWbx_Qax:-Nxz20KR1V701zDR8scA:b1i1bg6) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/5da078b2dc7a95f07ae3f4c4ed2bb2f3521e09e8/src/BloomExe/Publish/Video/RecordVideoWindow.cs#L659)):*
> ```Smalltalk
>                     while (
>                         shortName == ""
>                         || origAudioFileNames.Contains(Path.Combine(workingDirectory, shortName))
> ```

using `RobustFile.Exists` didn't work because of the case where the preexistant shortnamed-file (e.g. "i.mp3") gets renamed to a different name before we try to check if "i.mp3" is an available shortname, in which case RobustFile.Exists`is false but then we will hit trouble later on when we try to rename everything back to their original names

andrew-polk (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6495)* status: all files reviewed, 2 unresolved discussions (waiting on @nabalone)

___
*[`src/BloomExe/Publish/Video/RecordVideoWindow.cs` line 649 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6495#-Nxz1Jq7Aa-bityo2iT2:-Ny14g4q2JUnqPDCmqDc:b3f08zt) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/5da078b2dc7a95f07ae3f4c4ed2bb2f3521e09e8/src/BloomExe/Publish/Video/RecordVideoWindow.cs#L649)):*

Would it be more efficient to combine `Directory.GetFiles(workingDirectory, "?.*")`, `Directory.GetFiles(workingDirectory, "??.*")`, `Directory.GetFiles(workingDirectory, "???.*")` etc. (depending on the number of audio files)?
</blockquote></details>

How about `Directory.EnumerateFiles(workingDirectory, "{1,4}.*")`?

Given that the vast majority of files will be longer than that (guids), it does seem worth prefiltering it down.

We'll want a comment about what we're doing. And then the variable would be something like origAudioFileNameWhichMightCollide.

PR #6501, opened 2024-05-22
BL-13496 Simplify Image Command handling

hatton:


hatton:
*[`src/BloomExe/Edit/EditingModel.cs` line 1537 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6501#-NyW6wtbB3BoAbtpQRbS:-NyWBFjY0nwTWbpJ7CWe:b7t00x3) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/d369ffd3a0e2c8b0019c65684c98c802d36acfa1/src/BloomExe/Edit/EditingModel.cs#L1537)):*

Unless we succeed in not stripping pages, we should keep RefreshDisplayOfCurrentPage() any time we do a SaveNow() and don't move to some other page.
I would be inclined to keep the record of why we don't immediately clean up the old image.
</blockquote></details>

If/When we reinstate saving, it will be done in the new system which will not leave it to every caller to know details like whether the page has to be reloaded because we dared to Save it :-)

JohnThomson (approved):
> Reviewed 9 of 10 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6501)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @hatton)

PR #6507, opened 2024-05-24
Fix BL-13514 Speed up Large Book with SVGs

hatton:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6507)

### DIFF
--- a/src/BloomExe/XmlHtmlConverter.cs
+++ b/src/BloomExe/XmlHtmlConverter.cs
@@ -544,17 +544,14 @@ namespace Bloom
 
         public static string RestoreSvgs(string input, List<string> svgs)
         {
-            var result = input;
+            var builder = new StringBuilder(input);
+
             foreach (var svg in svgs)
             {
-                // We want a ReplaceFirst but .Net doesn't have it.
-                int index = result.IndexOf(SvgPlaceholder);
-                result =
-                    result.Substring(0, index)
-                    + svg
-                    + result.Substring(index + SvgPlaceholder.Length);
+                builder.Replace(SvgPlaceholder, svg);
             }
-            return result;
+
+            return builder.ToString();
         }
     }
 }


StephenMcConnel (changes requested):
The proposed code gives wrong results if there are more than one SVG to restore.  I've proposed a block of code that does work properly and is even faster.

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6507#-:-Nz-EJdy7i0sZJtZKwVB:bq7eepb)* status: all files reviewed, 1 unresolved discussion (waiting on @hatton)

___
*[`src/BloomExe/XmlHtmlConverter.cs` line 551 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6507#-Nz--vMW2EEEh5TvGc_Y:-Nz--vMW2EEEh5TvGc_Z:bq50xgj) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/c53f7f585d833da97ef5085da509e2c499878a1a/src/BloomExe/XmlHtmlConverter.cs#L551)):*
> ```Smalltalk
>             foreach (var svg in svgs)
>             {
>                 builder.Replace(SvgPlaceholder, svg);
> ```

This won't work because builder.Replace replaces ALL of the matches with the replacement string.   (That's what the comment in the original code was trying to say.) The result will be wrong unless svgs.Count == 1 (or svgs.Count == 0 and no placeholders exist).  (or if the svgs are all the same value)
This code achieves an even better speedup.  I've tested it with the given dictionary, checking that it gets the same results as the original code.
```
            if (svgs.Count == 0)
                return input;
            if (svgs.Count == 1)
                return input.Replace(SvgPlaceholder, svgs[0]);
            var splitInput = input.Split(new[] { SvgPlaceholder }, svgs.Count+1, StringSplitOptions.None);
            var builder = new StringBuilder();
            for (int i = 0; i < svgs.Count; i++)
            {
                builder.Append(splitInput[i]);
                builder.Append(svgs[i]);
            }
            builder.Append(splitInput[svgs.Count]);
            return builder.ToString();
```
svgs.Count=0:  My method took 00:00:00; old method took 00:00:00.0339980; your method took 00:00:00.
svgs.Count=2: My method took 00:00:00; old method took 00:00:00.0620134; your method took 00:00:00.0010006 and gave a wrong result
svgs.Count=98: My method took 00:00:00.0150026; old method took 00:00:03.0151914; your method took 00:00:00.7270079 and gave a wrong result

PR #6508, opened 2024-05-28
Allow Portrait orientation for Big Books (BL-13510)

StephenMcConnel:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6508)


hatton:
@StephenMcConnel OK, can you re-target to 5.6, merge, and then flow it through to 6.0? Thanks.

PR #6510, opened 2024-05-28
Improve performance of RemoveSvgs and RestoreSvgs (BL-13515)

StephenMcConnel:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6510)

### DIFF
--- a/src/BloomExe/XmlHtmlConverter.cs
+++ b/src/BloomExe/XmlHtmlConverter.cs
@@ -519,42 +519,32 @@ namespace Bloom
             }
         }
 
+        private static readonly Regex _svgMatch = new Regex("<svg\\s.*?</svg>", RegexOptions.Singleline|RegexOptions.Compiled);
         public static string RemoveSvgs(string input, List<string> svgs)
         {
             // Tidy utterly chokes (exits the whole program without even a green screen) on SVGs.
-            var regexSvg = new Regex("<svg .*?</svg>", RegexOptions.Singleline);
-            // This is probably not the most efficient way to remove and restore them but the common case is that none are found,
+            // This may not the most efficient way to remove and restore them but the common case is that none are found,
             // and this is not too bad for that case.
-            var result = input;
-            for (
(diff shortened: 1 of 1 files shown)


hatton:
*[`src/BloomExe/XmlHtmlConverter.cs` line 527 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6510#-Nz-oCP0E4U7s0LLz1St:-Nz-oCP0E4U7s0LLz1Su:byz97yw) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/fdd810d9398c598fb861a63e1deed26d892cbd46/src/BloomExe/XmlHtmlConverter.cs#L527)):*
> ```Smalltalk
>             // This may not the most efficient way to remove and restore them but the common case is that none are found,
>             // and this is not too bad for that case.
>             return Regex.Replace(input, "<svg .*?</svg>", (Match match) => {
> ```

What do you think about this optimization? https://stackoverflow.com/a/5685544

StephenMcConnel (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6510)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @hatton)

___
*[`src/BloomExe/XmlHtmlConverter.cs` line 527 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6510#-Nz-oCP0E4U7s0LLz1St:-Nz09EzOCC59mf_e8TGb:b-mpzg6n) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/fdd810d9398c598fb861a63e1deed26d892cbd46/src/BloomExe/XmlHtmlConverter.cs#L527)):*

What do you think about this optimization? https://stackoverflow.com/a/5685544
</blockquote></details>

Given that svg elements have other elements inside of them, I don't see any way to avoid the non-greedy regex.   Adding the Compiled flag does speed things up by 25-30%.  I've revised the code to use a static Regex object with the compiled flag added.

PR #6512, opened 2024-05-29
Spin up replacements for dead worker threads

JohnThomson:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6512)

### DIFF
--- a/src/BloomExe/web/BloomServer.cs
+++ b/src/BloomExe/web/BloomServer.cs
@@ -2,6 +2,7 @@
 // Copyright (c) 2014-2018 SIL International
 // This software is licensed under the MIT License (http://opensource.org/licenses/MIT)
 using System;
+using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.Globalization;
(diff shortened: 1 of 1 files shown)


src/BloomExe/web/BloomServer.cs
```diff
@@ -1448,6 +1456,33 @@ private void EnqueueIncomingRequests()
...
+                // nor whether it might have incremented but not decremented one of our counts of threads-in-use.
+                // But it's better than freezing up.
+                var deadThreads = _workers.Where(kvp => !kvp.Value.IsAlive);
+                foreach (var kvp in deadThreads)
+                {
+                    //thread.Join(); Copilot suggested this but I don't think you can join a dead thread???
+                    // Do we want a more drastic report? We don't know what went wrong with the thread, so a report
+                    // is unlikely to be very informative. But it's remotely possible that it damaged some data.
+                    // Could this be related to the wipeout bug?
+                    Debug.WriteLine(
```
hatton, line 1476: Should probably be a Logger.Event() so that we can see if this is happening a lot to someone.

PR #6514, opened 2024-05-30
Allow missing .map files in BloomServer (BL-13526)

StephenMcConnel:
These exist in developer builds, but not in installed builds.  (We don't want to pay the 11MB price of including files that might be useful for debugging, but not that helpful for user machines when expanded to 52MB.)

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6514)

### DIFF
--- a/src/BloomExe/NonFatalProblem.cs
+++ b/src/BloomExe/NonFatalProblem.cs
@@ -146,26 +146,7 @@ namespace Bloom
                         }
                         else
                         {
-                            // We don't want any notification (MessageBox or toast) targetting a ProgressDialog,
-                            // since the dialog seems to disappear quickly and leave us hanging... and not able to show.
-                            // We'll keep the form if it's not a ProgressDialog in order to center our message properly.
-                            if (formForSynchronizing is ProgressDialog)
-                            {
-                                MessageBox.Show(
-                                    fullDetailedMessage,
-                                    string.Empty,
-                                    MessageBoxButtons.OK
-                                );
-                            }
-                            else
-                            {
-                                MessageBox.Show(
-                                    formForSynchronizing,
-                                    fullDetailedMessage,
-                                    string.Empty,
-                                    MessageBoxButtons.OK
-                                );
-                            }
+                            ShowProblemInMessageBox(fullDetailedMessage, formForSynchronizing);
                         }
                     }
                     catch (Exception ex)
@@ -209,6 +190,43 @@ namespace Bloom
                         errorWhileReporting,
                         "Error while reporting non fatal error"
                     );
+            }
+        }
+
+        private static void ShowProblemInMessageBox(string fullDetailedMessage, Form formForSynchronizing)
+        {
+            if (formForSynchronizing == null)
+                return; // could be on wrong thread
+
+            if (formForSynchronizing.InvokeRequired)
+            {
+                formForSynchronizing.BeginInvoke(
+                    new Action(() =>
+                    {
+                        ShowProblemInMessageBox(fullDetailedMessage, formForSynchronizing);
+                    })
+                );
+                return;
+            }
+            // We don't want any notification (MessageBox or toast) targetting a ProgressDialog,
+            // since the dialog seems to disappear quickly and leave us hanging... and not able to show.
+            // We'll keep the form if it's not a ProgressDialog in order to center our message properly.
+            if (formForSynchronizing is ProgressDialog)
+            {
+                MessageBox.Show(
+                    fullDetailedMessage,
+                    string.Empty,
+                    MessageBoxButtons.OK
+                );
+            }
+            else
+            {
+                MessageBox.Show(
+                    formForSynchronizing,
+                    fullDetailedMessage,
+                    string.Empty,
+                    MessageBoxButtons.OK
+                );
             }
         }
 
--- a/src/BloomExe/web/BloomServer.cs
+++ b/src/BloomExe/web/BloomServer.cs
@@ -894,6 +894,7 @@ namespace Bloom.Api
             }
         }
 
+        private int _missingMapFileCount = 0;
         private bool ProcessAnyFileContent(IRequestInfo info, string localPath)
         {
             string modPath = localPath;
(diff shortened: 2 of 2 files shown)


JohnThomson (changes requested):
Wiping out these messages entirely seems a bit strong to me. It's annoying to get a dozen of them every time you open an inspector, but it's also annoying if you want to debug and can't access the original source files.
I'd be inclined, once per run of Bloom, to report a missing map file, along the lines
- real local build (minor version number zero?) : "Bloom was unexpectedly unable to find the map file X"
- Team City build: "Shipping versions of Bloom don't come with map files. For a better debugging experience you may want to copy them to the X directory."

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6514#-:-NzEHflW0NO__4WPz87Z:b1tryvg)* status: all files reviewed, 2 unresolved discussions (waiting on @StephenMcConnel)

___
*[`src/BloomExe/web/BloomServer.cs` line 893 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6514#-NzEK63HDj6zRIPlEV-M:-NzEK63HDj6zRIPlEV-N:ba7eao3) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/f962df0e9e0d2b4da2743e637a692523dada6bad/src/BloomExe/web/BloomServer.cs#L893)):*
> ```Smalltalk
>                 var isMap = localPath.EndsWith(".map");
>                 return FileLocationUtilities.GetFileDistributedWithApplication(
>                     optional: isMap,
> ```

It's surprising that we need a special case for .map here. As things stand, if GFDWA can't find it, it throws an ApplicationException with a particular message, which the caller catches, then tries various other ways to find the file, and if none of them work it eventually calls ReportMissingFile if ShouldReportFailedRequest returns true.
I would have thought the place to handle this was to make ShouldReportFailedRequest return false for .map files, or if you like my suggestion about reporting them once, make ReportMissingFile handle that.
(I also don't see why we don't just always pass optional:true and have the calling function handle a null result, rather than throwing an exception for a not-particularly-exceptional situation and catching it. Nor do I see any clear reason not to just inline the logic of this function, since it only has one caller and its name is not particularly helpful.)
___
*[`src/BloomExe/web/BloomServer.cs` line 912 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6514#-NzEFYLN4E9Lsa4r6m7T:-NzEFYLO-35w0C9Tcdqr:b-smvatj) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/f962df0e9e0d2b4da2743e637a692523dada6bad/src/BloomExe/web/BloomServer.cs#L912)):*
> ```Smalltalk
>                 path = ProcessPath(localPath, modPath);
>                 // Silently accept missing source map files. They are not critical to the functioning of the app.
>                 if (path == null && localPath.EndsWith(".map"))
> ```

I think you want path != null.

StephenMcConnel (commented):
I've retained a few messages, but simpler than you suggested.  I also redid detection of the problem to actually work properly and to fit in with the existing code and comments.
I also fixed the desired error reported to work from the worker threads.

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6514#-:-NzUfVIv-uhkx4sVoTb4:b-qe3dt1)* status: 0 of 2 files reviewed, 2 unresolved discussions (waiting on @JohnThomson)

___
*[`src/BloomExe/web/BloomServer.cs` line 893 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6514#-NzEK63HDj6zRIPlEV-M:-NzToHAM1xxCj_DvlB-n:b-ocuxh2) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/f962df0e9e0d2b4da2743e637a692523dada6bad/src/BloomExe/web/BloomServer.cs#L893)):*

It's surprising that we need a special case for .map here. As things stand, if GFDWA can't find it, it throws an ApplicationException with a particular message, which the caller catches, then tries various other ways to find the file, and if none of them work it eventually calls ReportMissingFile if ShouldReportFailedRequest returns true.
I would have thought the place to handle this was to make ShouldReportFailedRequest return false for .map files, or if you like my suggestion about reporting them once, make ReportMissingFile handle that.
(I also don't see why we don't just always pass optional:true and have the calling function handle a null result, rather than throwing an exception for a not-particularly-exceptional situation and catching it. Nor do I see any clear reason not to just inline the logic of this function, since it only has one caller and its name is not particularly helpful.)
</blockquote></details>

For most files, I think not finding the file would be an exceptional situation.  But I changed the implementation for detection albeit still using an optional argument to force a null value on return instead of throwing an exception.
___
*[`src/BloomExe/web/BloomServer.cs` line 912 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6514#-NzEFYLN4E9Lsa4r6m7T:-NzToPcr6AXqGnZ9ZTRe:batd8nf) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/f962df0e9e0d2b4da2743e637a692523dada6bad/src/BloomExe/web/BloomServer.cs#L912)):*

I think you want path != null.
</blockquote></details>

path == null is what is returned for an optional file that is not found.
I've expanded on reporting a missing optional file at this point.

JohnThomson (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6514)* status: all files reviewed, 1 unresolved discussion (waiting on @StephenMcConnel)

___
*[`src/BloomExe/web/BloomServer.cs` line 912 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6514#-NzEFYLN4E9Lsa4r6m7T:-Nz_EtTq41U41xFiaoHD:b7bm7g9) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/f962df0e9e0d2b4da2743e637a692523dada6bad/src/BloomExe/web/BloomServer.cs#L912)):*

path == null is what is returned for an optional file that is not found.
I've expanded on reporting a missing optional file at this point.
</blockquote></details>

Sorry, my mistake: I missed that you were testing two different variables.

PR #6516, opened 2024-05-31
Keep xmatter IDs through BringXmatterUpToDate (BL-13527)

JohnThomson:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6516)

### DIFF
--- a/src/BloomExe/Book/Book.cs
+++ b/src/BloomExe/Book/Book.cs
@@ -2346,7 +2346,8 @@ namespace Bloom.Book
             // Per BL-3571, if it's using a layout we don't know (e.g., from a newer Bloom) we switch to A5Portrait.
             // Various things, especially publication, don't work with unknown page sizes.
             Layout layout = Layout.FromDomAndChoices(bookDOM, Layout.A5Portrait, fileLocator);
-            XMatterHelper.RemoveExistingXMatter(bookDOM);
+            var oldIds = new List<string>();
+            XMatterHelper.RemoveExistingXMatter(bookDOM, oldIds);
             // this says, if you can't figure out the page size, use the one we got before we removed the xmatter...
             // still requiring it to be a valid layout.
             layout = Layout.FromDomAndChoices(bookDOM, layout, fileLocator);
@@ -2354,7 +2355,8 @@ namespace Bloom.Book
                 _bookData.WritingSystemAliases,
                 layout,
                 BookInfo.UseDeviceXMatter,
-                _bookData.MetadataLanguage1Tag
+                _bookData.MetadataLanguage1Tag,
+                oldIds
             );
 
             var dataBookLangs = bookDOM.GatherDataBookLanguages();
--- a/src/BloomExe/Book/BookStarter.cs
+++ b/src/BloomExe/Book/BookStarter.cs
@@ -239,7 +239,9 @@ namespace Bloom.Book
             // class to figure out which BookData keys to remove.
             ClearUnneededOriginalContentFromDerivative(storage.Dom, bookData);
 
-            XMatterHelper.RemoveExistingXMatter(storage.Dom);
+            // For a new book, we discard any old xmatter Ids, so the new book will have its own page IDs.
+            // One way this is helpful is caching cover images by page ID, so each book has a different cover page ID.
+            XMatterHelper.RemoveExistingXMatter(storage.Dom, new List<string>());
 
             // BL-4586 Some old books ended up with background-image urls containing XML img tags
             // in the HTML-encoded string. This happened because the coverImage data-book element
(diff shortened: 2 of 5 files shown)


hatton (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6516)* status: all files reviewed, 1 unresolved discussion (waiting on @JohnThomson)

___
*[`src/BloomExe/Book/XMatterHelper.cs` line 361 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6516#-NzLBGFuEmNS9zlZtpMX:-NzLBGFuEmNS9zlZtpMY:b-t8eqg6) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/d954b16f4d571ec9508e77a29315b3242035b155/src/BloomExe/Book/XMatterHelper.cs#L361)):*
> ```Smalltalk
>             {
>                 var newPageDiv = _bookDom.RawDom.ImportNode(xmatterPage, true) as XmlElement;
>                 // If we're updating an existing book, we want to keep the IDs (as much as possible; sometimes
> ```

This was prompted by a page with an ID going away unexpectedly as it was replaced. I'm left wondering if we're looking at the back cover and the update of the xmatter were to cause us to have one less page then before, would we have the same problem we're trying to solve? The answer to my Q would probably work well as a comment somewhere around here.

JohnThomson (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6516)* status: 4 of 5 files reviewed, 1 unresolved discussion (waiting on @hatton)

___
*[`src/BloomExe/Book/XMatterHelper.cs` line 361 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6516#-NzLBGFuEmNS9zlZtpMX:-NzTA99a4Xq7LTzGlNLV:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/d954b16f4d571ec9508e77a29315b3242035b155/src/BloomExe/Book/XMatterHelper.cs#L361)):*

This was prompted by a page with an ID going away unexpectedly as it was replaced. I'm left wondering if we're looking at the back cover and the update of the xmatter were to cause us to have one less page then before, would we have the same problem we're trying to solve? The answer to my Q would probably work well as a comment somewhere around here.
</blockquote></details>

Done.

hatton (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6516)* status: 4 of 5 files reviewed, 1 unresolved discussion (waiting on @JohnThomson)

___
*[`src/BloomExe/Book/XMatterHelper.cs` line 361 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6516#-NzLBGFuEmNS9zlZtpMX:-NzTMotH49eoNkmj3Ej8:bvy6swh) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/d954b16f4d571ec9508e77a29315b3242035b155/src/BloomExe/Book/XMatterHelper.cs#L361)):*

Done.
</blockquote></details>

OK, I don't see from the comment how this is achieved? I.e. how is it that it will reset somehow to the first page, when before it would crash?

JohnThomson (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6516)* status: 4 of 5 files reviewed, 1 unresolved discussion (waiting on @hatton)

___
*[`src/BloomExe/Book/XMatterHelper.cs` line 361 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6516#-NzLBGFuEmNS9zlZtpMX:-NzTxiWk6BMnIjOXFd-X:b-tp7xzo) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/d954b16f4d571ec9508e77a29315b3242035b155/src/BloomExe/Book/XMatterHelper.cs#L361)):*

OK, I don't see from the comment how this is achieved? I.e. how is it that it will reset somehow to the first page, when before it would crash?
</blockquote></details>

It's part of https://github.com/BloomBooks/BloomDesktop/pull/6517. Exception handler around like 946. (Maybe I should have combined these, but they were presented as separate cards.)

hatton (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6516)* status: 4 of 5 files reviewed, all discussions resolved (waiting on @JohnThomson)

hatton (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6516)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @JohnThomson)

PR #6517, opened 2024-05-31
Cope better with error state (BL-13529)

JohnThomson:
- If we're trying to quit and not in a state where we can save, just quit.
- If we something goes wrong navigating to a page, try navigating to page 1 (before reporting it).

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6517)

### DIFF
--- a/src/BloomExe/Edit/EditingModel.cs
+++ b/src/BloomExe/Edit/EditingModel.cs
@@ -194,7 +194,8 @@ namespace Bloom.Edit
                     // generally a bad idea, but we may have failed to save some changes. On the other hand,
                     // the only likely reason for this is that the program is in a bad state, probably from a previously
                     // reported error.
-                    args.Delayed = SaveThen(
+                    args.Delayed = true;
+                    SaveThen(
                         () =>
                         {
                             // We are setting skipSaveToDisk true so that we can do it ourselves here BEFORE
@@ -204,6 +205,7 @@ namespace Bloom.Edit
                             args.PostponedWork();
                             return null;
                         },
+                        doIfNotInRightStateToSave: () => args.Delayed = false, // go ahead and quit now
                         skipSaveToDisk: true
                     );
                 }
@@ -220,11 +222,14 @@ namespace Bloom.Edit
                 //shown so the view has never been full constructed, so we're not in a good state to do a refresh
                 if (Visible)
                 {
-                    SaveThen(() =>
-                    {
-                        _view.UpdatePageList(false);
-                        return _pageSelection.CurrentSelection.Id;
-                    });
+                    SaveThen(
+                        () =>
+                        {
+                            _view.UpdatePageList(false);
+                            return _pageSelection.CurrentSelection.Id;
+                        },
+                        () => { } // wrong state, I think there's nothing we can safely do.
+                    );
                 }
             });
             _contentLanguages = new List<ContentLanguage>();
@@ -388,6 +393,23 @@ namespace Bloom.Edit
                         details.PostponedWork?.Invoke();
                         return null; // leaving this tab, show blank page
                     },
+                    () =>
+                    {
+                        // We disable the tab control while we're in SavePending or SavedAndStripped.
+                        // We shouldn't be in NoPage while in the edit tab, but if we somehow are, we take the branch above.
+                        // If we're Editing, we will take the branch above.
+                        // So this is just the case where we're Navigating, either because we clicked on the Edit tab
+                        // and then immediately something else, or clicked another tab during the fraction of a second
+                        // while Bloom is navigating to a new page after doing some command. Abort the navigate, then go ahead.
+                        Guard.AssertThat(
+                            StateMachine.Navigating,
+                            "This branch should only be taken when navigating"
+                        );
+                        StateMachine.ToNoPage();
+                        _oldActiveForm = Form.ActiveForm;
+                        Application.Idle += ReactivateFormOnIdle;
+                        details.PostponedWork?.Invoke();
+                    },
                     skipSaveToDisk: true
                 );
             }
@@ -510,6 +532,7 @@ namespace Bloom.Edit
                     }
                     return newPageId;
                 },
+                () => { }, // wrong state, do nothing
                 forceFullSave: true
             );
         }
@@ -556,6 +579,7 @@ namespace Bloom.Edit
                         _inProcessOfDeleting = false;
                     }
                 },
+                () => { }, // wrong state, do nothing
                 forceFullSave: true
             );
         }
@@ -625,6 +649,7 @@ namespace Bloom.Edit
                     Logger.WriteEvent("InsertTemplatePage");
                     return newPageId;
                 },
+                () => { }, // wrong state, do nothing
                 forceFullSave: true
             );
         }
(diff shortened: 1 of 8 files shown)


hatton (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6517)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @JohnThomson)

___
*[`src/BloomExe/Edit/EditingModel.cs` line 193 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6517#-NzLBnb27c0iF5ySUCIt:-NzLBnb27c0iF5ySUCIu:b1lb1aw) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/dc6e74733756dfd88f1dcf864db94e7d47245571/src/BloomExe/Edit/EditingModel.cs#L193)):*
> ```Smalltalk
>                     // action that continues the shutdown process (after the Save completes).
>                     // If we can't initiate a Save, we'll just let the shutdown proceed (leave Delayed false).
>                     // Review: should we warn the user? Displaying UI while the user is tying to close the program is
> ```

I'd say we should either delay or just lose it.
___
*[`src/BloomExe/Edit/EditingModel.cs` line 200 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6517#-NzLCWF052OQSZkGIRSU:-NzLCWF052OQSZkGIRSV:bd8izt5) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/dc6e74733756dfd88f1dcf864db94e7d47245571/src/BloomExe/Edit/EditingModel.cs#L200)):*
> ```Smalltalk
>                         () =>
>                         {
>                             CurrentBook.Save();
> ```

This is not part of this PR but I find this confusing. You call `SaveThen()` and in the then, you actually `CurrentBook.Save()`.

hatton (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6517)* status: 0 of 1 files reviewed, 3 unresolved discussions (waiting on @JohnThomson)

___
*[`src/BloomExe/Edit/EditingModel.cs` line 1445 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6517#-NzLD2SR1-NYt-xhy_cb:-NzLD2SR1-NYt-xhy_cc:bxy5hy2) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/dc6e74733756dfd88f1dcf864db94e7d47245571/src/BloomExe/Edit/EditingModel.cs#L1445)):*
> ```Smalltalk
>         /// will not be called (even later).
>         /// </summary>
>         public bool SaveThen(
> ```

An idea: instead of an ignorable return values, we could require a `Func<string> doIfSaveFailed`.

JohnThomson (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6517)* status: 0 of 1 files reviewed, 3 unresolved discussions (waiting on @hatton)

___
*[`src/BloomExe/Edit/EditingModel.cs` line 193 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6517#-NzLBnb27c0iF5ySUCIt:-NzT69S56XJ_h8Y2gIxx:bkp8yjg) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/dc6e74733756dfd88f1dcf864db94e7d47245571/src/BloomExe/Edit/EditingModel.cs#L193)):*

I'd say we should either delay or just lose it.
</blockquote></details>

Losing it is the current proposal.
___
*[`src/BloomExe/Edit/EditingModel.cs` line 200 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6517#-NzLCWF052OQSZkGIRSU:-NzT6F4G0CYHEDbNp7RT:b61rx96) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/dc6e74733756dfd88f1dcf864db94e7d47245571/src/BloomExe/Edit/EditingModel.cs#L200)):*

This is not part of this PR but I find this confusing. You call `SaveThen()` and in the then, you actually `CurrentBook.Save()`.
</blockquote></details>

There are two parts to saving: saving the content of the live browser DOM to the C# XML DOM, and saving the C# DOM to disk. The former is done before the 'doAfterSaving' action, the latter is usually done after it (unless saveActionHandlesSaveBook is passed true, as here).
___
*[`src/BloomExe/Edit/EditingModel.cs` line 1445 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6517#-NzLD2SR1-NYt-xhy_cb:-NzT7Bop4BgtBcnPGIXe:bnt4y6c) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/dc6e74733756dfd88f1dcf864db94e7d47245571/src/BloomExe/Edit/EditingModel.cs#L1445)):*

An idea: instead of an ignorable return values, we could require a `Func<string> doIfSaveFailed`.
</blockquote></details>

In most cases, there is nothing to do. But I suppose it might be good to force the programmer to think about it.

hatton (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6517)* status: 0 of 1 files reviewed, 3 unresolved discussions (waiting on @JohnThomson)

___
*[`src/BloomExe/Edit/EditingModel.cs` line 200 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6517#-NzLCWF052OQSZkGIRSU:-NzTNZ-1499a1el_RfrR:b-n92r04) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/dc6e74733756dfd88f1dcf864db94e7d47245571/src/BloomExe/Edit/EditingModel.cs#L200)):*

There are two parts to saving: saving the content of the live browser DOM to the C# XML DOM, and saving the C# DOM to disk. The former is done before the 'doAfterSaving' action, the latter is usually done after it (unless saveActionHandlesSaveBook is passed true, as here).
</blockquote></details>

I think we chatted about "Save" which doesn't go to disk being too misleading. I thought we came up with a different name.
___
*[`src/BloomExe/Edit/EditingModel.cs` line 1445 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6517#-NzLD2SR1-NYt-xhy_cb:-NzTNtzb7ugi-_CBp3qc:b32v57) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/dc6e74733756dfd88f1dcf864db94e7d47245571/src/BloomExe/Edit/EditingModel.cs#L1445)):*

In most cases, there is nothing to do. But I suppose it might be good to force the programmer to think about it.
</blockquote></details>

+1

JohnThomson (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6517)* status: 0 of 1 files reviewed, 3 unresolved discussions (waiting on @hatton)

___
*[`src/BloomExe/Edit/EditingModel.cs` line 200 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6517#-NzLCWF052OQSZkGIRSU:-NzTkZ5V4mg2lRuOHyGO:b-pvqf80) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/dc6e74733756dfd88f1dcf864db94e7d47245571/src/BloomExe/Edit/EditingModel.cs#L200)):*

I think we chatted about "Save" which doesn't go to disk being too misleading. I thought we came up with a different name.
</blockquote></details>

I renamed the relevant parameter of SaveThen and added comments in the two places where the action does the saveToDisk

andrew-polk (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6517)* status: all files reviewed, 4 unresolved discussions (waiting on @hatton and @JohnThomson)

___
*[`src/BloomExe/Edit/EditingModel.cs` line 1445 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6517#-NzLD2SR1-NYt-xhy_cb:-NzTtYf07K2MPZ24jKaA:b-sjs3l8) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/4bd89a8058871f5ed06a3011ae7a789976325646/src/BloomExe/Edit/EditingModel.cs#L1445)):*

+1
</blockquote></details>

I assume the `+1` means lets force the programmer to think about it; but that hasn't been implemented.
___
*[`src/BloomExe/Edit/EditingModel.cs` line 975 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6517#-NzTuGFm3Si9q958c13d:-NzTuGFm3Si9q958c13e:b-fldhep) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/791355043b3085a4832d9747472ae039699c04a7/src/BloomExe/Edit/EditingModel.cs#L975)):*
> (This is untested; it's a fall-back to a fall-back, and I can't think of a way to make it happen.)

You could at least throw inside your outer catch to at least that something reasonable happens.

JohnThomson (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6517)* status: 0 of 8 files reviewed, 4 unresolved discussions (waiting on @andrew-polk and @hatton)

___
*[`src/BloomExe/Edit/EditingModel.cs` line 1445 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6517#-NzLD2SR1-NYt-xhy_cb:-NzU9Z2W89lbwZjhsVJ0:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/4bd89a8058871f5ed06a3011ae7a789976325646/src/BloomExe/Edit/EditingModel.cs#L1445)):*

I assume the `+1` means lets force the programmer to think about it; but that hasn't been implemented.
</blockquote></details>

Done.
___
*[`src/BloomExe/Edit/EditingModel.cs` line 975 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6517#-NzTuGFm3Si9q958c13d:-NzU914kCdnHJwBZhzMl:bsk5uiy) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/791355043b3085a4832d9747472ae039699c04a7/src/BloomExe/Edit/EditingModel.cs#L975)):*

> (This is untested; it's a fall-back to a fall-back, and I can't think of a way to make it happen.)

You could at least throw inside your outer catch to at least that something reasonable happens.
</blockquote></details>

I tried, and it did not work. But I can't think what more to try that would. The only valid way to a good state where
we can switch pages is a successful navigation. If we can't navigate to the front cover, what else do we try? This book may have had it. I updated the comment a bit; depending on what exception might be thrown (hopefully this never happens?) it may do some good.

andrew-polk (changes requested):
> Reviewed 8 of 8 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6517)* status: all files reviewed, 4 unresolved discussions (waiting on @hatton and @JohnThomson)

___
*[`src/BloomExe/Edit/EditingModel.cs` line 401 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6517#-NzUNJ8m5UeRZtc-Hb8b:-NzUNJ8m5UeRZtc-Hb8c:b-x8jgc8) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/f256f53e336d27ae02952289c1c535e2534e0bd7/src/BloomExe/Edit/EditingModel.cs#L401)):*
> So this is just the case where we're Navigating

Is it worth guarding for that?

JohnThomson (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6517)* status: 7 of 8 files reviewed, 4 unresolved discussions (waiting on @andrew-polk and @hatton)

___
*[`src/BloomExe/Edit/EditingModel.cs` line 401 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6517#-NzUNJ8m5UeRZtc-Hb8b:-NzUVV9r0cqLQwzbyGZW:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/f256f53e336d27ae02952289c1c535e2534e0bd7/src/BloomExe/Edit/EditingModel.cs#L401)):*

> So this is just the case where we're Navigating

Is it worth guarding for that?
</blockquote></details>

Done.

andrew-polk (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6517)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @hatton)

PR #6535, opened 2024-06-11
Fix up error handling around save failures in the state machine

andrew-polk:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6535)

### DIFF
--- a/src/BloomExe/Book/Book.cs
+++ b/src/BloomExe/Book/Book.cs
@@ -3938,7 +3938,10 @@ namespace Bloom.Book
                     {
                         // We've seen pages get emptied out, and we don't know why. This is a safety check.
                         // See BL-13078, BL-13120, BL-13123, and BL-13143 for examples.
-                        if (BookStorage.CheckForEmptyMarginBoxOnPage(pageToSaveToDisk))
+                        if (
+                            pageToSaveToDisk != null
+                            && BookStorage.CheckForEmptyMarginBoxOnPage(pageToSaveToDisk)
+                        )
                         {
                             // This has been logged and reported to the user. We don't want to save the empty page.
                             return;
--- a/src/BloomExe/Edit/EditingModel.cs
+++ b/src/BloomExe/Edit/EditingModel.cs
@@ -147,11 +147,14 @@ namespace Bloom.Edit
                     RequestBrowserToSave();
                 },
                 // updateBookWithPageContent
-                (string pageId, string htmlAndUserStyles) =>
-                    UpdateBookDomFromBrowserPageContent(htmlAndUserStyles),
+                (string pageId, string pageContentData) =>
+                    UpdateBookDomFromBrowserPageContent(pageContentData),
                 // saveBook
                 () =>
                 {
+                    if (_modifiedPageElement == null)
+                        return;
+
                     CurrentBook.SavePageToDisk(_modifiedPageElement, _nextSaveMustBeFull);
                     _nextSaveMustBeFull = false;
                     _pageHasUnsavedDataDerivedChange = false;
@@ -225,12 +228,18 @@ namespace Bloom.Edit
                             // We are setting skipSaveToDisk true so that we can do it ourselves here BEFORE
                             // the postponed work, which is going to shut everything down and would prevent
                             // the normal automatic save-to-disk from working.
+                            // If the save failed before this action gets called, Delayed is true, and PostponedWork
+                            // doesn't get done at all. This typically means the collection will not close.
+                            // However, FailureAction should be called in this case which allows closing the collection
+                            // to try again. If we do try again and the same page fails again, the state machine will
+                            // call this action anyway. So, finally PostponedWork will get called and we can close the collection.
                             CurrentBook.Save();
                             args.PostponedWork();
                             return null;
                         },
                         doIfNotInRightStateToSave: () => args.Delayed = false, // go ahead and quit now
-                        skipSaveToDisk: true
+                        skipSaveToDisk: true,
+                        failureAction: args.FailureAction
                     );
                 }
             });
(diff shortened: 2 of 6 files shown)


JohnThomson (approved):
A few comment/naming suggestions, which I consider optional, so I've approved it.

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6535#-:-O-C4R8N9FuwDeLhLQJR:bciy6lf)* status: all files reviewed, 3 unresolved discussions (waiting on @andrew-polk)

___
*[`src/BloomExe/Shell.cs` line 222 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6535#-O-C0tYDAl4Gdk2sgurR:-O-C0tYDAl4Gdk2sgurS:bje1ypn) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/ab7e3a9a02caf2794a0aa25b1e817876c9d24a88/src/BloomExe/Shell.cs#L222)):*
> ```Smalltalk
>                     FailureAction = () =>
>                     {
>                         _startedClosingEvent = false;
> ```

I think this wants a comment along the lines "We didn't want a second attempt at saving if the user clicks the close box while we are still trying to save after the first click on Close. But if the first attempt fails, we don't want to stay in a state where all attempts to close the program are ignored."
___
*[`src/BloomExe/Edit/EditingModel.cs` line 326 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6535#-O-C1jU377Q9icfRN2Tn:-O-C1jU377Q9icfRN2To:b-5uhp0d) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/ab7e3a9a02caf2794a0aa25b1e817876c9d24a88/src/BloomExe/Edit/EditingModel.cs#L326)):*
> ```Smalltalk
>             string userCssContent
>         )
>         {
> ```

I think we want a comment somewhere, and this might be a good place, along the lines "If anything goes badly wrong here, we want to throw rather then just bringing up a dialog. The process of saving the page content to the DOM should either succeed or throw, so that we don't get stuck in an invalid state that locks up the UI."
___
*[`src/BloomExe/Edit/EditingModel.cs` line 1543 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6535#-O-C3hAS87w9HIzCa3Sa:-O-C3hAS87w9HIzCa3Sb:bls5oe4) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/ab7e3a9a02caf2794a0aa25b1e817876c9d24a88/src/BloomExe/Edit/EditingModel.cs#L1543)):*
> ```Smalltalk
>         /// </summary>
>         /// <param name="htmlAndUserStyles"></param>
>         public void ReceivePageContent(string htmlAndUserStyles)
> ```

Wonder if we could name the param better, here and elsewhere, to reflect that it might also contain an error message. pageContentData? pageContentResponse? In one way it's useful to know that it's a combination of html and styles (usually), but really that's private to the (JS) code that sends it and the code the processes it, and the intermediate code that just passes it on doesn't care.

andrew-polk (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6535)* status: all files reviewed, 3 unresolved discussions (waiting on @JohnThomson)

___
*[`src/BloomExe/Shell.cs` line 222 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6535#-O-C0tYDAl4Gdk2sgurR:-O-C87DA4nQitkAwjwsv:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/ab7e3a9a02caf2794a0aa25b1e817876c9d24a88/src/BloomExe/Shell.cs#L222)):*

I think this wants a comment along the lines "We didn't want a second attempt at saving if the user clicks the close box while we are still trying to save after the first click on Close. But if the first attempt fails, we don't want to stay in a state where all attempts to close the program are ignored."
</blockquote></details>

Done.
___
*[`src/BloomExe/Edit/EditingModel.cs` line 326 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6535#-O-C1jU377Q9icfRN2Tn:-O-C7yJcEhGzeNDaEscn:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/ab7e3a9a02caf2794a0aa25b1e817876c9d24a88/src/BloomExe/Edit/EditingModel.cs#L326)):*

I think we want a comment somewhere, and this might be a good place, along the lines "If anything goes badly wrong here, we want to throw rather then just bringing up a dialog. The process of saving the page content to the DOM should either succeed or throw, so that we don't get stuck in an invalid state that locks up the UI."
</blockquote></details>

Done.
___
*[`src/BloomExe/Edit/EditingModel.cs` line 1543 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6535#-O-C3hAS87w9HIzCa3Sa:-O-C7eofDxPYgSiK1yPh:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/ab7e3a9a02caf2794a0aa25b1e817876c9d24a88/src/BloomExe/Edit/EditingModel.cs#L1543)):*

Wonder if we could name the param better, here and elsewhere, to reflect that it might also contain an error message. pageContentData? pageContentResponse? In one way it's useful to know that it's a combination of html and styles (usually), but really that's private to the (JS) code that sends it and the code the processes it, and the intermediate code that just passes it on doesn't care.
</blockquote></details>

Done.

JohnThomson (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6535)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @andrew-polk)

PR #6559, opened 2024-06-20
Wrap XmlNode based classes for thread safety

StephenMcConnel:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6559)


andrew-polk (changes requested):
My review is still in progress. But I'm publishing in case it's helpful to see what I already commented.

> Reviewed 56 of 75 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6559#-:-O-xHTJ3D17f_rA33vtQ:b7u3gop)* status: 56 of 75 files reviewed, 13 unresolved discussions (waiting on @StephenMcConnel)

___
*[`src/BloomExe/Book/Book.cs` line 427 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6559#-O-vN0rH-rC8Jpno6-47:-O-vN0rH-rC8Jpno6-48:b-dxju4u) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/1a3c62fb26620da2b697f0ed435cfb7858a2294f/src/BloomExe/Book/Book.cs#L427)):*
> ```Smalltalk
>                     "//span[contains(concat(' ', normalize-space(@class), ' '), ' bloom-linebreak ')]"
>                 );
>                 var lineBreakElements = lineBreaks.Cast<XmlElement>().ToArray();    // NOT SafeXmlElement
> ```

Let's specify why we can't/shouldn't use SafeXmlElement
___
*[`src/BloomExe/Book/Book.cs` line 1508 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6559#-O-vNvVa2MiOkDIzMD85:-O-vNvVa2MiOkDIzMD86:b-d7tgkn) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/1a3c62fb26620da2b697f0ed435cfb7858a2294f/src/BloomExe/Book/Book.cs#L1508)):*
> ```Smalltalk
>             var templateDoc = XmlHtmlConverter.GetXmlDomFromHtmlFile(bookPath, false);
>             var newPage = SafeXmlElement.FakeWrap(
>                 (XmlElement)templateDoc.SafeSelectNodes("//div[@id='" + updateTo.Guid + "']")[0]);
> ```

note to self: what is this FakeWrap about?
___
*[`src/BloomExe/Book/Book.cs` line 2895 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6559#-O-vOhOq7kEKkells41p:-O-vOhOr7AkMUPyXIbys:b-1qtqct) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/1a3c62fb26620da2b697f0ed435cfb7858a2294f/src/BloomExe/Book/Book.cs#L2895)):*
> ```Smalltalk
>             {
>                 var div = node as SafeXmlElement;
>                 if (div?.GetAttribute("lang") != lang || div.Name == "label")
> ```

I think this caused a subtle change in the logic if `lang == null` but I can't tell if that matters or not.
___
*[`src/BloomExe/Book/Book.cs` line 3396 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6559#-O-vPOjc09QvU1nrzcgs:-O-vPOjc09QvU1nrzcgt:b5xl7ly) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/1a3c62fb26620da2b697f0ed435cfb7858a2294f/src/BloomExe/Book/Book.cs#L3396)):*
> ```Smalltalk
>                 throw new ApplicationException("FindPageDiv(): pageElement is null");
>             }
>             Require.That(pageElement != null, "Page could not be found: " + page.XPathToDiv);
> ```

Isn't this null check doing the same thing as the Require?
___
*[`src/BloomExe/Book/Book.cs` line 4022 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6559#-O-vPl4t2JJ5DuN3Fevq:-O-vPl4t2JJ5DuN3Fevr:b-7kntpm) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/1a3c62fb26620da2b697f0ed435cfb7858a2294f/src/BloomExe/Book/Book.cs#L4022)):*
> ```Smalltalk
>             var body = pageDiv.ParentNode;
>             body.RemoveChild(pageDiv);
>             pages.Remove(pageDiv);
> ```

It isn't clear to me why we want to remove pageDiv from pages.
I guess it must get re-added by one of the methods below?

Probably worth a comment.

Also, is this fixing an existing bug? i.e. how did this come up?
___
*[`src/BloomExe/Book/BookCompressor.cs` line 333 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6559#-O-vQIQzE3mfVMGzMOfI:-O-vQIQzE3mfVMGzMOfJ:b5o7sy1) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/1a3c62fb26620da2b697f0ed435cfb7858a2294f/src/BloomExe/Book/BookCompressor.cs#L333)):*
> ```Smalltalk
>             foreach (
>                 var node in dom.SafeSelectNodes("//BrandingProjectName")
>                     .Cast<XmlElement>() // NOT SafeXmlElement
> ```

Please be explicit about why not.
(here and a few lines above)
___
*[`src/BloomExe/Book/BookStarter.cs` line 312 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6559#-O-x8k1z1o_Fpzrs-5jS:-O-x8k1z1o_Fpzrs-5jT:b-d5q7i2) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/8c2b206ef72daf018bebf6d5f817b4087edc6b74/src/BloomExe/Book/BookStarter.cs#L312)):*
> ```Smalltalk
>                 var sourceDiv =
>                     sourceDom.SelectSingleNode("//div[@id='" + div.GetAttribute("id") + "']")
>                     as XmlElement;  // NOT SafeXmlElement
> ```

Let's say why not.
___
*[`src/BloomExe/Book/BookStorage.cs` line 3973 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6559#-O-x9BcF6LFy_yzp3gMj:-O-x9BcF6LFy_yzp3gMk:b-tmbdd0) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/8c2b206ef72daf018bebf6d5f817b4087edc6b74/src/BloomExe/Book/BookStorage.cs#L3973)):*
> ```Smalltalk
>             // See XMatterHelper.InjectFlyleafIfNeeded().
>             if (HtmlDom.IsFlyleafPage(page))
>                 return false;
> ```

This code was "added" in your PR, likely as the result of a bad merge.
It was removed recently.
___
*[`src/BloomExe/Book/Layout.cs` line 271 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6559#-O-x9crVBwCiW1EwK_bh:-O-x9crVBwCiW1EwK_bi:b-18iwl4) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/8c2b206ef72daf018bebf6d5f817b4087edc6b74/src/BloomExe/Book/Layout.cs#L271)):*
> ```Smalltalk
>                             ((XmlElement)sizeAndOrientation).Name   // NOT SafeXmlElement
>                         );
>                         foreach (XmlElement option in ((XmlElement)sizeAndOrientation).ChildNodes)  // NOT SafeXmlElement
> ```

Let's say why not (x4).

This raises another question I have:
(How) can we ensure that new code properly uses the SafeXml namespace classes?
We would be able to do something clever like you did for another class (RobustX?) to do a text search on commit(?). But if we have all these where we still use the original classes, how are we going to catch the problem locations?

Could we do something like `XmlElement/* not SafeXml */` in addition to a comment explaining why we can't in each instance?
___
*[`src/BloomExe/Book/TranslationGroupManager.cs` line 141 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6559#-O-xBFi80Cx98kNbrryP:-O-xBFi80Cx98kNbrryQ:bsx1xme) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/8c2b206ef72daf018bebf6d5f817b4087edc6b74/src/BloomExe/Book/TranslationGroupManager.cs#L141)):*
> ```Smalltalk
>         private static void GenerateEditableDivsWithPreTranslatedContent(SafeXmlNode elementOrDom)
>         {
>             var ownerDoc = elementOrDom.OwnerDocument;
> ```

I don't understand this change. Did you decide NodeType is never Document?
___
*[`src/BloomExe/Book/TranslationGroupManager.cs` line 523 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6559#-O-xBk4kCmBkt2Nowhtz:-O-xBk4kCmBkt2Nowhu-:b-s7ghhq) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/8c2b206ef72daf018bebf6d5f817b4087edc6b74/src/BloomExe/Book/TranslationGroupManager.cs#L523)):*
> ```Smalltalk
>             var visibleSiblingLangs = (editable.ParentNode).ChildNodes
>                 .Where(ed => ed is SafeXmlElement)
>                 .Cast<SafeXmlElement>()
> ```

This is existing, but there's no reason to Cast here, right? You just did .Where is SafeXmlElement.
___
*[`src/BloomExe/Publish/Epub/ReadiumManifest.cs` line 109 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6559#-O-xDkL41-q4S1TV25n4:-O-xDkL41-q4S1TV25n5:b-4nbaaq) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/8c2b206ef72daf018bebf6d5f817b4087edc6b74/src/BloomExe/Publish/Epub/ReadiumManifest.cs#L109)):*
> ```Smalltalk
>             }
> 
>             var spineElts = _opfDoc.SafeSelectNodes("//spine/itemref").Cast<XmlElement>();  // NOT SafeXmlElement
> ```

why not?
Same about 20 lines down. And 30 lines.
___
*[`src/BloomExe/Spreadsheet/MarkedUpText.cs` line 111 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6559#-O-xEMhrBiyZ76seTGJk:-O-xEMhrBiyZ76seTGJl:b-iq3w19) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/8c2b206ef72daf018bebf6d5f817b4087edc6b74/src/BloomExe/Spreadsheet/MarkedUpText.cs#L111)):*
> ```Smalltalk
> 
>             //There are no paragraph elements, just keep all whitespace
>             if (((XmlElement)root).GetElementsByTagName("p").Count == 0)    // NOT SafeXmlElement
> ```

why not? x4

andrew-polk (changes requested):
> Reviewed 1 of 75 files at r1.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6559)* status: 57 of 75 files reviewed, 15 unresolved discussions (waiting on @StephenMcConnel)

___
*[`src/BloomExe/XmlHtmlConverter.cs` line 45 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6559#-O0A3HEq1lU0UORU_ELI:-O0A3HEq1lU0UORU_ELJ:bb0q3ld) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/8c2b206ef72daf018bebf6d5f817b4087edc6b74/src/BloomExe/XmlHtmlConverter.cs#L45)):*
> ```Smalltalk
>         );
> 
>         public static XmlDocument GetXmlDomFromHtmlFile(
> ```

Why is it safe/preferred/necessary for these two methods to return XmlDocument rather than SafeXmlDocument?

Maybe we better discuss this PR from a high level so I don't keep asking these questions...
___
*[`src/BloomExe/XmlHtmlConverter.cs` line 388 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6559#-O0A1AWw4_mP2KJm_vqg:-O0A1AWw4_mP2KJm_vqh:b94pvs7) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/8c2b206ef72daf018bebf6d5f817b4087edc6b74/src/BloomExe/XmlHtmlConverter.cs#L388)):*
> ```Smalltalk
>         {
>             MakeXmlishTagsSafeForInterpretationAsHtml(new SafeXmlDocument(dom));
>         }
> ```

Here and `SaveDOMAsHtml5` and `ConvertDomToHtml5`, probably deserves comments as to why we still allow an XmlDocument version.

andrew-polk (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6559)* status: 57 of 75 files reviewed, 15 unresolved discussions (waiting on @StephenMcConnel)

___
*[`src/BloomExe/XmlHtmlConverter.cs` line 45 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6559#-O0A3HEq1lU0UORU_ELI:-O0B67bB57vDu_0Nrp3F:b-j4be5w) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/8c2b206ef72daf018bebf6d5f817b4087edc6b74/src/BloomExe/XmlHtmlConverter.cs#L45)):*

Why is it safe/preferred/necessary for these two methods to return XmlDocument rather than SafeXmlDocument?

Maybe we better discuss this PR from a high level so I don't keep asking these questions...
</blockquote></details>

We decided to wait until JT is available to discuss:
- Is it better to just go all the way and convert everything to SafeXml? That makes the code simpler in one sense. You don't have to wonder if this use of Xml* is inherently safe. But there is a potential performance hit. (any other risks?)

andrew-polk (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6559)* status: 57 of 75 files reviewed, 15 unresolved discussions (waiting on @StephenMcConnel)

___
*[`src/BloomExe/XmlHtmlConverter.cs` line 45 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6559#-O0A3HEq1lU0UORU_ELI:-O0EzMFX7aDK_EOAg4vo:blpnwsf) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/8c2b206ef72daf018bebf6d5f817b4087edc6b74/src/BloomExe/XmlHtmlConverter.cs#L45)):*

We decided to wait until JT is available to discuss:
- Is it better to just go all the way and convert everything to SafeXml? That makes the code simpler in one sense. You don't have to wonder if this use of Xml* is inherently safe. But there is a potential performance hit. (any other risks?)
</blockquote></details>

After discussion, Steve is going to look into how much more it would take to make all the production code use SafeXml.

StephenMcConnel (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6559)* status: 38 of 88 files reviewed, 15 unresolved discussions (waiting on @andrew-polk)

___
*[`src/BloomExe/XmlHtmlConverter.cs` line 45 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6559#-O0A3HEq1lU0UORU_ELI:-O0GeQNWEjFJnmc3sPj8:b-3ovtln) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/8c2b206ef72daf018bebf6d5f817b4087edc6b74/src/BloomExe/XmlHtmlConverter.cs#L45)):*

After discussion, Steve is going to look into how much more it would take to make all the production code use SafeXml.
</blockquote></details>

Done. I've replaced XmlDocument with SafeXmlDocument as much as possible.
___
*[`src/BloomExe/XmlHtmlConverter.cs` line 388 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6559#-O0A1AWw4_mP2KJm_vqg:-O0GgED4Bs8vzQGhgtt1:b-k29tad) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/8c2b206ef72daf018bebf6d5f817b4087edc6b74/src/BloomExe/XmlHtmlConverter.cs#L388)):*

Here and `SaveDOMAsHtml5` and `ConvertDomToHtml5`, probably deserves comments as to why we still allow an XmlDocument version.
</blockquote></details>

I've removed these unused methods.
___
*[`src/BloomExe/Book/Book.cs` line 427 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6559#-O-vN0rH-rC8Jpno6-47:-O0GhHoS2E32nQTfKva6:b-n83dpa) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/1a3c62fb26620da2b697f0ed435cfb7858a2294f/src/BloomExe/Book/Book.cs#L427)):*

Let's specify why we can't/shouldn't use SafeXmlElement
</blockquote></details>

Changed to using SafeXmlDocument/SafeXmlElement.
___
*[`src/BloomExe/Book/Book.cs` line 1508 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6559#-O-vNvVa2MiOkDIzMD85:-O0GhTVl2m9CGRLk_VC5:byo9kuj) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/1a3c62fb26620da2b697f0ed435cfb7858a2294f/src/BloomExe/Book/Book.cs#L1508)):*

note to self: what is this FakeWrap about?
</blockquote></details>

got rid of it: Changed to using SafeXmlDocument/SafeXmlElement.
___
*[`src/BloomExe/Book/Book.cs` line 2895 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6559#-O-vOhOq7kEKkells41p:-O0GiM2778-EtqfWI7tr:bmbht2k) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/1a3c62fb26620da2b697f0ed435cfb7858a2294f/src/BloomExe/Book/Book.cs#L2895)):*

I think this caused a subtle change in the logic if `lang == null` but I can't tell if that matters or not.
</blockquote></details>

I don't think it matters.  lang should never be null in a  method named "HasContentInLang"
___
*[`src/BloomExe/Book/Book.cs` line 3396 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6559#-O-vPOjc09QvU1nrzcgs:-O0Gj0kXA4LNxHOvl5_Y:b-t70zs2) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/1a3c62fb26620da2b697f0ed435cfb7858a2294f/src/BloomExe/Book/Book.cs#L3396)):*

Isn't this null check doing the same thing as the Require?
</blockquote></details>

yes.  I'll get rid of one.
___
*[`src/BloomExe/Book/Book.cs` line 4022 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6559#-O-vPl4t2JJ5DuN3Fevq:-O0GjZVp83mMtZcRxNL9:b-fzovlj) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/1a3c62fb26620da2b697f0ed435cfb7858a2294f/src/BloomExe/Book/Book.cs#L4022)):*

It isn't clear to me why we want to remove pageDiv from pages.
I guess it must get re-added by one of the methods below?

Probably worth a comment.

Also, is this fixing an existing bug? i.e. how did this come up?
</blockquote></details>

I've added a comment.  It's needed for the body.InsertAfter to work properly.  When using the XmlNodeList before, this was apparently done under the hood.
___
*[`src/BloomExe/Book/BookCompressor.cs` line 333 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6559#-O-vQIQzE3mfVMGzMOfI:-O0Juu2p-3Q72ma72Lj-:buyp7z7) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/1a3c62fb26620da2b697f0ed435cfb7858a2294f/src/BloomExe/Book/BookCompressor.cs#L333)):*

Please be explicit about why not.
(here and a few lines above)
</blockquote></details>

changed to use SafeXml classes
___
*[`src/BloomExe/Book/BookStarter.cs` line 312 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6559#-O-x8k1z1o_Fpzrs-5jS:-O0JuzXr8EBzYtNAVUtR:buyp7z7) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/8c2b206ef72daf018bebf6d5f817b4087edc6b74/src/BloomExe/Book/BookStarter.cs#L312)):*

Let's say why not.
</blockquote></details>

changed to use SafeXml classes
___
*[`src/BloomExe/Book/BookStorage.cs` line 3973 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6559#-O-x9BcF6LFy_yzp3gMj:-O0FGAzH14gpuBUZk5hg:btesnx1) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/8c2b206ef72daf018bebf6d5f817b4087edc6b74/src/BloomExe/Book/BookStorage.cs#L3973)):*

This code was "added" in your PR, likely as the result of a bad merge.
It was removed recently.
</blockquote></details>

Done.  Removed.
___
*[`src/BloomExe/Book/Layout.cs` line 271 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6559#-O-x9crVBwCiW1EwK_bh:-O0GeAn10h5w3JJBUgjQ:basfod1) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/8c2b206ef72daf018bebf6d5f817b4087edc6b74/src/BloomExe/Book/Layout.cs#L271)):*

Let's say why not (x4).

This raises another question I have:
(How) can we ensure that new code properly uses the SafeXml namespace classes?
We would be able to do something clever like you did for another class (RobustX?) to do a text search on commit(?). But if we have all these where we still use the original classes, how are we going to catch the problem locations?

Could we do something like `XmlElement/* not SafeXml */` in addition to a comment explaining why we can't in each instance?
</blockquote></details>

I've replaced XmlDocument with SafeXmlDocument everywhere it makes any sense.
___
*[`src/BloomExe/Book/TranslationGroupManager.cs` line 141 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6559#-O-xBFi80Cx98kNbrryP:-O0K6dL-0tDLmbC6Gp28:b-gassjp) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/8c2b206ef72daf018bebf6d5f817b4087edc6b74/src/BloomExe/Book/TranslationGroupManager.cs#L141)):*

I don't understand this change. Did you decide NodeType is never Document?
</blockquote></details>

The SafeXmlNode implementation of OwnerDocument returns itself if the node is a SafeXmlDocument.
___
*[`src/BloomExe/Book/TranslationGroupManager.cs` line 523 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6559#-O-xBk4kCmBkt2Nowhtz:-O0JvG7974K9Dp8LJ0o2:b-gcqtah) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/8c2b206ef72daf018bebf6d5f817b4087edc6b74/src/BloomExe/Book/TranslationGroupManager.cs#L523)):*

This is existing, but there's no reason to Cast here, right? You just did .Where is SafeXmlElement.
</blockquote></details>

I don't think the .Where changes/casts the type: it just filters.
___
*[`src/BloomExe/Publish/Epub/ReadiumManifest.cs` line 109 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6559#-O-xDkL41-q4S1TV25n4:-O0K70IXA0xVwsmzDd3O:b5b94h7) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/8c2b206ef72daf018bebf6d5f817b4087edc6b74/src/BloomExe/Publish/Epub/ReadiumManifest.cs#L109)):*

why not?
Same about 20 lines down. And 30 lines.
</blockquote></details>

changed to used SafeXml classes everywhere
___
*[`src/BloomExe/Spreadsheet/MarkedUpText.cs` line 111 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6559#-O-xEMhrBiyZ76seTGJk:-O0K7KOJ4_5z6E7878Dt:b5b94h7) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/8c2b206ef72daf018bebf6d5f817b4087edc6b74/src/BloomExe/Spreadsheet/MarkedUpText.cs#L111)):*

why not? x4
</blockquote></details>

changed to used SafeXml classes everywhere

andrew-polk (changes requested):
> Reviewed 3 of 75 files at r1, 47 of 47 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6559)* status: all files reviewed, 16 unresolved discussions (waiting on @StephenMcConnel)

___
*[`src/BloomExe/Book/HtmlDom.cs` line 49 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6559#-O0kJmLI94NVMRsrNJco:-O0kJmLI94NVMRsrNJcp:bim9su7) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/ebaf58c113c1a4c818dbb3db8f3059a7cbc05727/src/BloomExe/Book/HtmlDom.cs#L49)):*
> ```Smalltalk
>         }
> 
>         public HtmlDom(XmlDocument domToClone)
> ```

Looks like we can get rid of the XmlDocument version now.
___
*[`src/BloomExe/Book/HtmlDom.cs` line 525 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6559#-O0kKMVECRuz1zMMkJJ0:-O0kKMVECRuz1zMMkJJ1:b21jf71) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/ebaf58c113c1a4c818dbb3db8f3059a7cbc05727/src/BloomExe/Book/HtmlDom.cs#L525)):*
> ```Smalltalk
>         }
> 
>         // Review SafeXml
> ```

Is this comment still helpful?
___
*[`src/BloomExe/Book/HtmlDom.cs` line 558 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6559#-O0kKan-8d1HKNfp9jup:-O0kKan-8d1HKNfp9juq:blc14lm) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/ebaf58c113c1a4c818dbb3db8f3059a7cbc05727/src/BloomExe/Book/HtmlDom.cs#L558)):*
> ```Smalltalk
>         public void RemoveMetaElement(string name)
>         {
>             lock (_dom.Lock)
> ```

Is it worth moving this into XmlDocument so all the locking happens in that namespace?
(Same with override below)
___
*[`src/BloomExe/SafeXml/SafeXmlDocument.cs` line 5 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6559#-O0kFPpqBzMBdOWwfSsj:-O0kFPpqBzMBdOWwfSsk:b-hp0krk) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/ebaf58c113c1a4c818dbb3db8f3059a7cbc05727/src/BloomExe/SafeXml/SafeXmlDocument.cs#L5)):*
> ```Smalltalk
> using SIL.Xml;
> 
> namespace Bloom.SafeXml
> ```

Can we write a few barebones tests just to prove that this is thread-safe?
___
*[`src/BloomExe/SafeXml/SafeXmlDocument.cs` line 10 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6559#-O0k276B-EmglRVzcc7m:-O0k276B-EmglRVzcc7n:b-mmfrqd) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/ebaf58c113c1a4c818dbb3db8f3059a7cbc05727/src/BloomExe/SafeXml/SafeXmlDocument.cs#L10)):*
> return and appropriate

typo
___
*[`src/BloomExe/SafeXml/SafeXmlDocument.cs` line 13 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6559#-O0k22c74YJ8TwyGRqH6:-O0k22c74YJ8TwyGRqH7:b-ec40zz) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/ebaf58c113c1a4c818dbb3db8f3059a7cbc05727/src/BloomExe/SafeXml/SafeXmlDocument.cs#L13)):*
>     /// Where relevant, if an object that can be wrapped is an input argument, we will provide overloads
>     /// that take either the wrapped or unwrapped version.

Is this still true?

Hm. I see the constructor. Maybe this general comment should be removed and we should add something more specific on the constructor? Or should this comment be made more specific?
___
*[`src/BloomExe/SafeXml/SafeXmlDocument.cs` line 20 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6559#-O0k3Me7CQBl96zrovg7:-O0k3Me7CQBl96zrovg8:b2dg9oj) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/ebaf58c113c1a4c818dbb3db8f3059a7cbc05727/src/BloomExe/SafeXml/SafeXmlDocument.cs#L20)):*
> ```Smalltalk
>     public class SafeXmlDocument : SafeXmlNode
>     {
>         private XmlDocument Doc => (XmlDocument)_node;
> ```

Was there a reason you decided to call the private member `Doc` as opposed to `_doc`?

And I'm wondering if it is better to call it something like `_wrappedDoc` or `_wrappedDocument`?
___
*[`src/BloomExe/SafeXml/SafeXmlDocument.cs` line 73 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6559#-O0k4gXYDPCb-PMF4B3n:-O0k4gXYDPCb-PMF4B3o:bkek21r) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/ebaf58c113c1a4c818dbb3db8f3059a7cbc05727/src/BloomExe/SafeXml/SafeXmlDocument.cs#L73)):*
> ```Smalltalk
> 
>         /// <summary>
>         /// Review: do we need a locked wrapper for these?
> ```

Sorry; I don't understand the question.

These *are* the locked wrappers, right?
So you're just asking if we need these?
Now that everything is using SafeXml, seems like we do?
___
*[`src/BloomExe/SafeXml/SafeXmlElement.cs` line 17 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6559#-O0kB6VdDtEK8LDxo2FI:-O0kB6VdDtEK8LDxo2FJ:bcw3aha) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/ebaf58c113c1a4c818dbb3db8f3059a7cbc05727/src/BloomExe/SafeXml/SafeXmlElement.cs#L17)):*
> ```Smalltalk
>         // In this class, _node is always an XmlElement, since it is readonly and set by our constructor, which requires that.
>         // This saves us from having to cast it in every method.
>         private XmlElement Element => (XmlElement)_node;
> ```

I think it would be better to call this _element or _wrappedElement.
___
*[`src/BloomExe/SafeXml/SafeXmlNode.cs` line 12 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6559#-O0k7rhPDELVzmdQbuAE:-O0k7rhPDELVzmdQbuAF:b-du0ji6) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/ebaf58c113c1a4c818dbb3db8f3059a7cbc05727/src/BloomExe/SafeXml/SafeXmlNode.cs#L12)):*
> ```Smalltalk
>     public class SafeXmlNode
>     {
>         internal readonly XmlNode _node;
> ```

I think it will be helpful to have all these wrapped elements named consistently as _wrappedX (_wrappedNode).

Working through this file, I got a little tripped up with the fact that _node is a raw XmlNode but _doc is a SafeXmlDocument.
___
*[`src/BloomExe/SafeXml/SafeXmlNode.cs` line 202 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6559#-O0k7ha-6KxsUH9vXamU:-O0k7ha-6KxsUH9vXamV:b-eo77cd) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/ebaf58c113c1a4c818dbb3db8f3059a7cbc05727/src/BloomExe/SafeXml/SafeXmlNode.cs#L202)):*
> Only Elements can write to XmlWriter

Only Documents and Elements can write to XmlWriter?
___
*[`src/BloomTests/FluentAssertXml.cs` line 35 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6559#-O0k-h7N8zdPlIFpd43P:-O0k-h7N8zdPlIFpd43Q:b-eskm4j) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/ebaf58c113c1a4c818dbb3db8f3059a7cbc05727/src/BloomTests/FluentAssertXml.cs#L35)):*
> ```Smalltalk
>         }
> 
>         public static AssertElement Element(XmlElement element)
> ```

Looks like `Element(XmlElement element)` and `Dom(XmlDocument dom)` can be removed.
___
*[`src/BloomTests/ToPalaso/XmlExtensionsTests.cs` line 16 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6559#-O0kF2a_Da-X_winIMu0:-O0kF2a_Da-X_winIMu1:bp7f29) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/ebaf58c113c1a4c818dbb3db8f3059a7cbc05727/src/BloomTests/ToPalaso/XmlExtensionsTests.cs#L16)):*
> ```Smalltalk
>     /// The SafeXmlDocument and SafeXmlElement classes implement some of the same methods,
>     /// either directly or using the XmlExtension method.
>     /// </remarks>
> ```

I think XmlExtensions is no longer used since you copied the implementations into SafeXmlElement.

So should we remove that class can move/convert these tests to SafeXmlElementTests?

andrew-polk (commented):
Great work, gentlemen!
I think I have finished my first round of review now.

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6559#-:-O0kMEBpDKObtS3LI1hG:b4pveeu)* status: all files reviewed, 16 unresolved discussions (waiting on @StephenMcConnel)

andrew-polk (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6559)* status: all files reviewed, 16 unresolved discussions (waiting on @StephenMcConnel)

___
*[`src/BloomExe/Book/Layout.cs` line 271 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6559#-O-x9crVBwCiW1EwK_bh:-O0kMORq2xqaVb-fOVRF:b-riuaid) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/8c2b206ef72daf018bebf6d5f817b4087edc6b74/src/BloomExe/Book/Layout.cs#L271)):*

I've replaced XmlDocument with SafeXmlDocument everywhere it makes any sense.
</blockquote></details>

Thanks.
I'm keeping this conversation open as a reminder that we still would like to find a way to "guarantee" we stick to using SafeXml* going forward.

I'm looking for the other hook we (you?) did previously...

andrew-polk:
*[`src/BloomExe/Book/Layout.cs` line 271 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6559#-O-x9crVBwCiW1EwK_bh:-O0kQ5HfEvPnVD_U0Rel:b-2iankd) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/8c2b206ef72daf018bebf6d5f817b4087edc6b74/src/BloomExe/Book/Layout.cs#L271)):*

Thanks.
I'm keeping this conversation open as a reminder that we still would like to find a way to "guarantee" we stick to using SafeXml* going forward.

I'm looking for the other hook we (you?) did previously...
</blockquote></details>

Found it.
C:\dev\BloomDesktop\build\check-csharp-files.sh

But somehow the running of it got removed in 6.0, so I'm working on fixing that...

StephenMcConnel (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6559)* status: all files reviewed, 16 unresolved discussions (waiting on @andrew-polk)

___
*[`src/BloomExe/Book/HtmlDom.cs` line 49 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6559#-O0kJmLI94NVMRsrNJco:-O0kUQp23cpGCSF7nH2h:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/ebaf58c113c1a4c818dbb3db8f3059a7cbc05727/src/BloomExe/Book/HtmlDom.cs#L49)):*

Looks like we can get rid of the XmlDocument version now.
</blockquote></details>

Done.
___
*[`src/BloomExe/Book/HtmlDom.cs` line 525 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6559#-O0kKMVECRuz1zMMkJJ0:-O0kUgRJATjyQY315N7Q:bss4a2b) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/ebaf58c113c1a4c818dbb3db8f3059a7cbc05727/src/BloomExe/Book/HtmlDom.cs#L525)):*

Is this comment still helpful?
</blockquote></details>

Removed.
___
*[`src/BloomExe/Book/HtmlDom.cs` line 558 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6559#-O0kKan-8d1HKNfp9jup:-O0kauX29p8hM9ucHzVK:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/ebaf58c113c1a4c818dbb3db8f3059a7cbc05727/src/BloomExe/Book/HtmlDom.cs#L558)):*

Is it worth moving this into XmlDocument so all the locking happens in that namespace?
(Same with override below)
</blockquote></details>

Done.
___
*[`src/BloomExe/Book/Layout.cs` line 271 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6559#-O-x9crVBwCiW1EwK_bh:-O0kbScV9FO1-T1QObZ9:b-m7ihl4) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/8c2b206ef72daf018bebf6d5f817b4087edc6b74/src/BloomExe/Book/Layout.cs#L271)):*

Found it.
C:\dev\BloomDesktop\build\check-csharp-files.sh

But somehow the running of it got removed in 6.0, so I'm working on fixing that...
</blockquote></details>

It might be easier to implement a second shell script than to make the current check-csharp-files.sh more complicated.
___
*[`src/BloomExe/Book/TranslationGroupManager.cs` line 141 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6559#-O-xBFi80Cx98kNbrryP:-O0kcUq0EVkQ45RWlUhp:bnoj22u) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/8c2b206ef72daf018bebf6d5f817b4087edc6b74/src/BloomExe/Book/TranslationGroupManager.cs#L141)):*

The SafeXmlNode implementation of OwnerDocument returns itself if the node is a SafeXmlDocument.
</blockquote></details>

So if elementOrDom.SafeXmlNode.NodeType == XmlNodeType.Document, then the right value is already being returned by the SafeXmlDocument implementation of OwnerDocument.
___
*[`src/BloomExe/SafeXml/SafeXmlDocument.cs` line 10 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6559#-O0k276B-EmglRVzcc7m:-O0kdkfh751aWavS10d-:b-p8ls69) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/ebaf58c113c1a4c818dbb3db8f3059a7cbc05727/src/BloomExe/SafeXml/SafeXmlDocument.cs#L10)):*

> return and appropriate

typo
</blockquote></details>

Fixed.  "return an appropriate"
___
*[`src/BloomExe/SafeXml/SafeXmlDocument.cs` line 13 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6559#-O0k22c74YJ8TwyGRqH6:-O0kf4MABe47rPJqxKM8:bcjpggv) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/ebaf58c113c1a4c818dbb3db8f3059a7cbc05727/src/BloomExe/SafeXml/SafeXmlDocument.cs#L13)):*

>     /// Where relevant, if an object that can be wrapped is an input argument, we will provide overloads
>     /// that take either the wrapped or unwrapped version.

Is this still true?

Hm. I see the constructor. Maybe this general comment should be removed and we should add something more specific on the constructor? Or should this comment be made more specific?
</blockquote></details>

I've replaced these lines with a comment about constructor taking the original classes.
___
*[`src/BloomExe/SafeXml/SafeXmlDocument.cs` line 20 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6559#-O0k3Me7CQBl96zrovg7:-O0kfO8sABdlqFTViw79:b67jelw) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/ebaf58c113c1a4c818dbb3db8f3059a7cbc05727/src/BloomExe/SafeXml/SafeXmlDocument.cs#L20)):*

Was there a reason you decided to call the private member `Doc` as opposed to `_doc`?

And I'm wondering if it is better to call it something like `_wrappedDoc` or `_wrappedDocument`?
</blockquote></details>

It's a reference to the _node stored by the base class typecast to XmlDocument, not a separate member variable.  A longer name might be appropriate, but it should be something like WrappedDocument, not _wrappedDocument which implies a separate member variable to my mind.
___
*[`src/BloomExe/SafeXml/SafeXmlDocument.cs` line 73 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6559#-O0k4gXYDPCb-PMF4B3n:-O0kgWWc5hjYXJfvrCE3:b11x3hf) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/ebaf58c113c1a4c818dbb3db8f3059a7cbc05727/src/BloomExe/SafeXml/SafeXmlDocument.cs#L73)):*

Sorry; I don't understand the question.

These *are* the locked wrappers, right?
So you're just asking if we need these?
Now that everything is using SafeXml, seems like we do?
</blockquote></details>

Removed the comment.
___
*[`src/BloomExe/SafeXml/SafeXmlElement.cs` line 17 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6559#-O0kB6VdDtEK8LDxo2FI:-O0kgevl9XdU6H5d4lw0:bpfbrn7) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/ebaf58c113c1a4c818dbb3db8f3059a7cbc05727/src/BloomExe/SafeXml/SafeXmlElement.cs#L17)):*

I think it would be better to call this _element or _wrappedElement.
</blockquote></details>

I disagree vehemently: the leading underscore signals a separate member variable, and this is a typecast of an existing member variable.  WrappedElement or WrappedXmlElement might be a better name, but not something that implies a separate member variable.
___
*[`src/BloomExe/SafeXml/SafeXmlNode.cs` line 12 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6559#-O0k7rhPDELVzmdQbuAE:-O0o6feN8_QqppqKXLOu:b-6pes8u) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/ebaf58c113c1a4c818dbb3db8f3059a7cbc05727/src/BloomExe/SafeXml/SafeXmlNode.cs#L12)):*

I think it will be helpful to have all these wrapped elements named consistently as _wrappedX (_wrappedNode).

Working through this file, I got a little tripped up with the fact that _node is a raw XmlNode but _doc is a SafeXmlDocument.
</blockquote></details>

I've changed the names to _wrappedNode, WrappedElement, and WrappedDocument.
___
*[`src/BloomExe/SafeXml/SafeXmlNode.cs` line 202 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6559#-O0k7ha-6KxsUH9vXamU:-O0o9gLi0IjLGisW9YzK:bfa1n03) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/ebaf58c113c1a4c818dbb3db8f3059a7cbc05727/src/BloomExe/SafeXml/SafeXmlNode.cs#L202)):*

> Only Elements can write to XmlWriter

Only Documents and Elements can write to XmlWriter?
</blockquote></details>

Noted.  I updated the message.
___
*[`src/BloomTests/FluentAssertXml.cs` line 35 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6559#-O0k-h7N8zdPlIFpd43P:-O0oAZVB1DTHP0Sqr-5g:b-f8bo50) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/ebaf58c113c1a4c818dbb3db8f3059a7cbc05727/src/BloomTests/FluentAssertXml.cs#L35)):*

Looks like `Element(XmlElement element)` and `Dom(XmlDocument dom)` can be removed.
</blockquote></details>

Done.  AssertFile File(string path) was also unused.
___
*[`src/BloomTests/ToPalaso/XmlExtensionsTests.cs` line 16 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6559#-O0kF2a_Da-X_winIMu0:-O0oD7kk5V5GuLQFCJWU:bytvq7f) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/ebaf58c113c1a4c818dbb3db8f3059a7cbc05727/src/BloomTests/ToPalaso/XmlExtensionsTests.cs#L16)):*

I think XmlExtensions is no longer used since you copied the implementations into SafeXmlElement.

So should we remove that class can move/convert these tests to SafeXmlElementTests?
</blockquote></details>

I've removed both XmlExtensions.cs and XmlExtensionsTests.cs.

andrew-polk (changes requested):
> Reviewed 13 of 13 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6559)* status: all files reviewed, 4 unresolved discussions (waiting on @StephenMcConnel)

___
*[`src/BloomExe/SafeXml/SafeXmlDocument.cs` line 20 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6559#-O0k3Me7CQBl96zrovg7:-O0oPYYt0CoHrfFyWqVC:b-wv02cs) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/ebaf58c113c1a4c818dbb3db8f3059a7cbc05727/src/BloomExe/SafeXml/SafeXmlDocument.cs#L20)):*

It's a reference to the _node stored by the base class typecast to XmlDocument, not a separate member variable.  A longer name might be appropriate, but it should be something like WrappedDocument, not _wrappedDocument which implies a separate member variable to my mind.
</blockquote></details>

Ok, I was considering asking JT for his thoughts, but I see the initial commit which you worked on together has it this way.
___
*[`src/BloomExe/SafeXml/SafeXmlDocument.cs` line 274 at r4](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6559#-O0oQh0J8fxyUvlqka1J:-O0oQh0J8fxyUvlqka1K:b-l84f5n) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/ce0fa1671bd595c1e44797f47debf72b09b3a151/src/BloomExe/SafeXml/SafeXmlDocument.cs#L274)):*
> ```Smalltalk
>         public bool HasMetaElement(string name)
>         {
>             return SafeSelectNodes("//head/meta[@name='" + name + "']").Length > 0;
> ```

I think this is safe because SafeSelectNodes gets the lock, but should we also get it here to be consistent?
___
*[`src/BloomTests/ToPalaso/XmlExtensionsTests.cs` line 16 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6559#-O0kF2a_Da-X_winIMu0:-O0oOh_V9w4ZnFY_I0-y:b-aogtv4) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/ebaf58c113c1a4c818dbb3db8f3059a7cbc05727/src/BloomTests/ToPalaso/XmlExtensionsTests.cs#L16)):*

I've removed both XmlExtensions.cs and XmlExtensionsTests.cs.
</blockquote></details>

Rather than just removing XmlExtensionsTests, can we move them to SafeXmlElementTests?
Or maybe you're already working on that...

andrew-polk (changes requested):
> Reviewed 19 of 19 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6559)* status: all files reviewed, 3 unresolved discussions (waiting on @StephenMcConnel)

___
*[`src/BloomBrowserUI/package.json` line 209 at r5](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6559#-O0pA2Al5an7VqAYhv7I:-O0pA2Al5an7VqAYhv7J:b-1eds8e) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/4689fcc78df61698a9a8d593dd14487e081e6d11/src/BloomBrowserUI/package.json#L209)):*
> ```JSON
>     "husky": {
>         "hooks": {
>             "pre-commit": "pretty-quick --staged && lint-staged --no-stash && ../../build/check-csharp-files.sh && ../../build/check-csharp-xmlclasses.sh",
> ```

Since we are making a separate .sh file for each set of things we are searching for, should we rename the original one to `check-csharp-robustfile.sh`?

StephenMcConnel (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6559)* status: all files reviewed, 3 unresolved discussions (waiting on @andrew-polk)

___
*[`src/BloomBrowserUI/package.json` line 209 at r5](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6559#-O0pA2Al5an7VqAYhv7I:-O0pHqzf1wWFb75qWCdX:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/4689fcc78df61698a9a8d593dd14487e081e6d11/src/BloomBrowserUI/package.json#L209)):*

Since we are making a separate .sh file for each set of things we are searching for, should we rename the original one to `check-csharp-robustfile.sh`?
</blockquote></details>

Done.
___
*[`src/BloomExe/SafeXml/SafeXmlDocument.cs` line 5 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6559#-O0kFPpqBzMBdOWwfSsj:-O0pIpvm0td_wKfzwJSk:b-5xsghx) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/ebaf58c113c1a4c818dbb3db8f3059a7cbc05727/src/BloomExe/SafeXml/SafeXmlDocument.cs#L5)):*

Can we write a few barebones tests just to prove that this is thread-safe?
</blockquote></details>

Did you see the new src/BloomTests/SafeXml/SafeXmlTests.cs file?  Admittedly only two very simple tests, but definitely demonstrating improved thread safety.
___
*[`src/BloomExe/SafeXml/SafeXmlDocument.cs` line 274 at r4](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6559#-O0oQh0J8fxyUvlqka1J:-O0pIUg217Zhncs-cxh0:b-k09rg8) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/ce0fa1671bd595c1e44797f47debf72b09b3a151/src/BloomExe/SafeXml/SafeXmlDocument.cs#L274)):*

I think this is safe because SafeSelectNodes gets the lock, but should we also get it here to be consistent?
</blockquote></details>

OK.  It doesn't really help, but it doesn't hurt either.
___
*[`src/BloomTests/ToPalaso/XmlExtensionsTests.cs` line 16 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6559#-O0kF2a_Da-X_winIMu0:-O0pG7i3CiYTLrUsQ0KZ:b-8okzyx) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/ebaf58c113c1a4c818dbb3db8f3059a7cbc05727/src/BloomTests/ToPalaso/XmlExtensionsTests.cs#L16)):*

Rather than just removing XmlExtensionsTests, can we move them to SafeXmlElementTests?
Or maybe you're already working on that...
</blockquote></details>

I removed the code that was being tested because it was being used only in the tests.  I figured getting rid of the two files was worthwhile to reduce our code base.  Without the code, there was no longer anything to test.  (What a circular argument...)

StephenMcConnel (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6559)* status: 92 of 96 files reviewed, 3 unresolved discussions (waiting on @andrew-polk)

___
*[`src/BloomTests/ToPalaso/XmlExtensionsTests.cs` line 16 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6559#-O0kF2a_Da-X_winIMu0:-O0pLDFu-ApN7jK1bn4T:b53gr1a) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/ebaf58c113c1a4c818dbb3db8f3059a7cbc05727/src/BloomTests/ToPalaso/XmlExtensionsTests.cs#L16)):*

I removed the code that was being tested because it was being used only in the tests.  I figured getting rid of the two files was worthwhile to reduce our code base.  Without the code, there was no longer anything to test.  (What a circular argument...)
</blockquote></details>

OK, I finally understand: copy/adjust the tests that reflect methods reimplemented by one of the SafeXml classes.  I'm working on it.

StephenMcConnel (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6559)* status: 91 of 96 files reviewed, 3 unresolved discussions (waiting on @andrew-polk)

___
*[`src/BloomTests/ToPalaso/XmlExtensionsTests.cs` line 16 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6559#-O0kF2a_Da-X_winIMu0:-O0pS9IcCbsRlZZXxpld:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/ebaf58c113c1a4c818dbb3db8f3059a7cbc05727/src/BloomTests/ToPalaso/XmlExtensionsTests.cs#L16)):*

OK, I finally understand: copy/adjust the tests that reflect methods reimplemented by one of the SafeXml classes.  I'm working on it.
</blockquote></details>

Done.

andrew-polk (approved):
> Reviewed 5 of 5 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6559)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @StephenMcConnel)

___
*[`src/BloomExe/SafeXml/SafeXmlDocument.cs` line 5 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6559#-O0kFPpqBzMBdOWwfSsj:-O0pYTB0BvglVLntZpov:bem5gdq) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/ebaf58c113c1a4c818dbb3db8f3059a7cbc05727/src/BloomExe/SafeXml/SafeXmlDocument.cs#L5)):*

Did you see the new src/BloomTests/SafeXml/SafeXmlTests.cs file?  Admittedly only two very simple tests, but definitely demonstrating improved thread safety.
</blockquote></details>

Yes, I wrote this before you added those. Thanks.
___
*[`src/BloomTests/ToPalaso/XmlExtensionsTests.cs` line 16 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6559#-O0kF2a_Da-X_winIMu0:-O0pYh1OEQ7ZTz8rBqOu:b-ucfkn4) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/ebaf58c113c1a4c818dbb3db8f3059a7cbc05727/src/BloomTests/ToPalaso/XmlExtensionsTests.cs#L16)):*

Done.
</blockquote></details>

Thanks.

PR #6562, opened 2024-06-24
Remove image ids after the program is done using them (BL-13619)

StephenMcConnel:


andrew-polk:
*[`src/BloomBrowserUI/bookEdit/js/bloomImages.ts` line 239 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6562#-O0B8xz7BYt2vwIQcit4:-O0B8xz87Z7t8HH251Pb:b8swv4h) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/6c521fe0e97256d57f64b9afb9558b430e5cc8a1/src/BloomBrowserUI/bookEdit/js/bloomImages.ts#L239)):*
> ```XML
>             }
> 
>             postJson("editView/" + command + "Image", { imageId, imageSrc });
> ```

Can we avoid all the C# changes by removing the ID in a success and failure callback here?

StephenMcConnel (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6562)* status: 0 of 5 files reviewed, 1 unresolved discussion (waiting on @andrew-polk)

___
*[`src/BloomBrowserUI/bookEdit/js/bloomImages.ts` line 239 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6562#-O0B8xz7BYt2vwIQcit4:-O0BOYlP4gXW0g5o-eQL:blsa23w) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/6c521fe0e97256d57f64b9afb9558b430e5cc8a1/src/BloomBrowserUI/bookEdit/js/bloomImages.ts#L239)):*

Can we avoid all the C# changes by removing the ID in a success and failure callback here?
</blockquote></details>

I don't know how that would work.  The "editView/changeImage" API call causes a dialog to come up, which can take an indeterminate amount of time, and which needs the id to still be in place if/when the user actually chooses an image.  It currently returns a successful result immediately without waiting for the dialog to come up.

StephenMcConnel (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6562)* status: 0 of 5 files reviewed, 1 unresolved discussion (waiting on @andrew-polk)

___
*[`src/BloomBrowserUI/bookEdit/js/bloomImages.ts` line 239 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6562#-O0B8xz7BYt2vwIQcit4:-O0BR19O4udJVAKoyare:b-rpwi1m) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/6c521fe0e97256d57f64b9afb9558b430e5cc8a1/src/BloomBrowserUI/bookEdit/js/bloomImages.ts#L239)):*

I don't know how that would work.  The "editView/changeImage" API call causes a dialog to come up, which can take an indeterminate amount of time, and which needs the id to still be in place if/when the user actually chooses an image.  It currently returns a successful result immediately without waiting for the dialog to come up.
</blockquote></details>

The other commands would probably work with a "finally" block on the postJson, but not the change command.

andrew-polk (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6562)* status: 0 of 5 files reviewed, 1 unresolved discussion (waiting on @StephenMcConnel)

___
*[`src/BloomBrowserUI/bookEdit/js/bloomImages.ts` line 239 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6562#-O0B8xz7BYt2vwIQcit4:-O0BXXx3CpiE194IWuaQ:b-amtg8n) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/6c521fe0e97256d57f64b9afb9558b430e5cc8a1/src/BloomBrowserUI/bookEdit/js/bloomImages.ts#L239)):*

The other commands would probably work with a "finally" block on the postJson, but not the change command.
</blockquote></details>

Yes, you're right.
I think JT and I were both thinking this would be more straightforward. Let's discuss it briefly in the morning.

andrew-polk (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6562)* status: 0 of 5 files reviewed, 1 unresolved discussion (waiting on @StephenMcConnel)

___
*[`src/BloomBrowserUI/bookEdit/js/bloomImages.ts` line 239 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6562#-O0B8xz7BYt2vwIQcit4:-O0EyzB-Dndd0IJwiO6V:bfcuziw) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/6c521fe0e97256d57f64b9afb9558b430e5cc8a1/src/BloomBrowserUI/bookEdit/js/bloomImages.ts#L239)):*

Yes, you're right.
I think JT and I were both thinking this would be more straightforward. Let's discuss it briefly in the morning.
</blockquote></details>

We decided to simplify the copy case to not add the ID.
Then the C# just has to handle the early returns as you've already done.

Let's add a comment here stating such.
I.e. the changeImage function will remove the ID after it uses it. If we don't get there, C# should make the call to removeImageId.

Thanks.

Could you also check those early returns I was looking at earlier in OnPasteImage?

StephenMcConnel (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6562)* status: 0 of 5 files reviewed, 1 unresolved discussion (waiting on @andrew-polk)

___
*[`src/BloomBrowserUI/bookEdit/js/bloomImages.ts` line 239 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6562#-O0B8xz7BYt2vwIQcit4:-O0FBa618wuP-kLBqsYA:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/6c521fe0e97256d57f64b9afb9558b430e5cc8a1/src/BloomBrowserUI/bookEdit/js/bloomImages.ts#L239)):*

We decided to simplify the copy case to not add the ID.
Then the C# just has to handle the early returns as you've already done.

Let's add a comment here stating such.
I.e. the changeImage function will remove the ID after it uses it. If we don't get there, C# should make the call to removeImageId.

Thanks.

Could you also check those early returns I was looking at earlier in OnPasteImage?
</blockquote></details>

Done.

andrew-polk (approved):
> Reviewed 3 of 5 files at r1, 3 of 3 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6562)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @StephenMcConnel)

PR #6602, opened 2024-08-19
Allow cover to be full-bleed image (BL-13271)

andrew-polk:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6602)

### DIFF
--- a/DistFiles/localization/en/BloomMediumPriority.xlf
+++ b/DistFiles/localization/en/BloomMediumPriority.xlf
@@ -5,6 +5,10 @@
       <note>We are in the process of moving strings from Bloom.xlf to BloomMediumPriority.xlf and BloomLowPriority.xlf while trying very hard to preserve translations and approval status.</note>
     </header>
     <body>
+      <trans-unit id="AvailableWithEnterprise" translate="no">
+        <source xml:lang="en">Available with your Enterprise Subscription</source>
+        <note>ID: AvailableWithEnterprise</note>
+      </trans-unit>
       <!-- Drag Activity Tool -->
       <trans-unit id="EditTab.Toolbox.DragActivity.Check">
         <source xml:lang="en">Check</source>
@@ -599,6 +603,14 @@
       <trans-unit id="BookSettings.CoverGroupLabel" sil:dynamic="true">
         <source xml:lang="en">Cover</source>
         <note>BookSettings.CoverGroupLabel</note>
+      </trans-unit>
+      <trans-unit id="BookSettings.CoverIsImage" sil:dynamic="true" translate="no">
+        <source xml:lang="en">Fill the front cover with a single image</source>
+        <note>BookSettings.CoverIsImage</note>
+      </trans-unit>
+      <trans-unit id="BookSettings.CoverIsImage.Description" sil:dynamic="true" translate="no">
+        <source xml:lang="en">Using this option turns on the [Print Bleed](https://docs.bloomlibrary.org) indicators on paper layouts. See [Full Page Cover Images](https://docs.bloomlibrary.org) for information on sizing your image to fit.</source>
+        <note>BookSettings.CoverIsImage.Description</note>
       </trans-unit>
       <trans-unit id="BookSettings.ContentPagesGroupLabel" sil:dynamic="true">
         <source xml:lang="en">Content Pages</source>
--- a/src/BloomBrowserUI/bookEdit/bookSettings/BookSettingsDialog.tsx
+++ b/src/BloomBrowserUI/bookEdit/bookSettings/BookSettingsDialog.tsx
@@ -50,6 +50,10 @@ import { default as TrashIcon } from "@mui/icons-material/Delete";
 import { PWithLink } from "../../react_components/pWithLink";
 import { FieldVisibilityGroup } from "./FieldVisibilityGroup";
 import { StyleAndFontTable } from "./StyleAndFontTable";
+import {
+    BloomEnterpriseIndicatorIconAndText,
+    useEnterpriseAvailable
+} from "../../react_components/requiresBloomEnterprise";
 
 let isOpenAlready = false;
 
@@ -208,6 +212,16 @@ export const BookSettingsDialog: React.FunctionComponent<{
         "BookSettings.TopLevelTextPadding.1emLabel"
     );
 
+    const coverIsImageLabel = useL10n(
+        "Fill the front cover with a single image",
+        "BookSettings.CoverIsImage"
+    );
+    //TODO real links (and change .xlf)
+    const coverIsImageDescription = useL10n(
+        "Using this option turns on the [Print Bleed](https://docs.bloomlibrary.org) indicators on paper layouts. See [Full Page Cover Images](https://docs.bloomlibrary.org) for information on sizing your image to fit.",
+        "BookSettings.CoverIsImage.Description"
+    );
+
     // This is a helper function to make it easier to pass the override information
     function getAdditionalProps<T>(
         subPath: string
@@ -306,6 +320,8 @@ export const BookSettingsDialog: React.FunctionComponent<{
         setMigratedTheme("");
     };
 
+    const enterpriseAvailable = useEnterpriseAvailable();
+
     function saveSettingsAndCloseDialog() {
         if (settingsToReturnLater) {
             // If nothing changed, we don't get any...and don't need to make this call.
@@ -353,7 +369,7 @@ export const BookSettingsDialog: React.FunctionComponent<{
                         height: 600px;
                         // This odd width was chosen to make the customBookStyles alert box format nicely.
                         // See BL-12956. It's not that important, but I don't think anything else is affected
-                        // much by the exact witdh.
+                        // much by the exact width.
                         width: 638px;
                         #groups {
                             margin-right: 10px; // make room for the scrollbar
@@ -398,6 +414,34 @@ export const BookSettingsDialog: React.FunctionComponent<{
                                 label={whatToShowOnCoverLabel}
                                 path={`appearance`}
                             >
+                                <div>
+                                    <ConfigrBoolean
+                                        label={coverIsImageLabel}
+                                        description={coverIsImageDescription}
+                                        {...getAdditionalProps<boolean>(
+                                            `coverIsImage`
+                                        )}
+                                        disabled={
+                                            appearanceDisabled ||
+                                            !enterpriseAvailable
+                                        }
+                                    />
+                                    <div
+                                        css={css`
+                                            display: flex;
+                                            padding-bottom: 5px;
+                                            font-size: 12px;
+                                            font-weight: bold;
+                                        `}
+                                    >
+                                        <BloomEnterpriseIndicatorIconAndText
+                                            css={css`
+                                                margin-left: auto;
+                                            `}
+                                            disabled={appearanceDisabled}
+                                        />
+                                    </div>
+                                </div>
                                 <FieldVisibilityGroup
                                     field="cover-title"
                                     labelFrame="Show Title in {0}"
--- a/src/BloomBrowserUI/react_components/requiresBloomEnterprise.tsx
+++ b/src/BloomBrowserUI/react_components/requiresBloomEnterprise.tsx
@@ -7,7 +7,7 @@ import { get, post } from "../utils/bloomApi";
 import Button from "@mui/material/Button";
 import { kBloomBlue50Transparent, lightTheme } from "../bloomMaterialUITheme";
 import { ThemeProvider, StyledEngineProvider } from "@mui/material/styles";
-import { Div } from "./l10nComponents";
+import { Div, Span } from "./l10nComponents";
 import { useL10n } from "./l10nHooks";
 import { WireUpForWinforms } from "../utils/WireUpWinform";
 import { Dialog, DialogActions, DialogContent } from "@mui/material";
@@ -487,6 +487,48 @@ export const RequiresBloomEnterpriseDialog: React.FunctionComponent<{
                 </DialogBottomButtons>
             </div>
         </BloomDialog>
+    );
+};
+
+export const BloomEnterpriseIndicatorIconAndText: React.FunctionComponent<{
+    disabled?: boolean;
+    className?: string;
+}> = props => {
+    const enterpriseAvailable = useEnterpriseAvailable();
+
+    return (
+        <div
+            onClick={() => {
+                enterpriseAvailable ||
+                    props.disabled ||
+                    openBloomEnterpriseSettings();
+            }}
+            css={css`
+                display: flex;
+                align-items: center;
+                ${enterpriseAvailable || props.disabled || "cursor:pointer"};
+
+                opacity: ${props.disabled ? kBloomDisabledOpacity : 1.0};
+            `}
+            className={props.className}
+        >
+            <img
+                src={badgeUrl}
+                css={css`
+                    height: 1.5em;
+                    padding-right: 0.5em;
+                `}
+            />
+            {enterpriseAvailable ? (
+                <Span l10nKey={"AvailableWithEnterprise"}>
+                    Available with your Enterprise Subscription
+                </Span>
+            ) : (
+                <Span l10nKey={"Common.EnterpriseRequired"}>
+                    Enterprise Required
+                </Span>
+            )}
+        </div>
     );
 };
 
--- a/src/BloomExe/Book/AppearanceSettings.cs
+++ b/src/BloomExe/Book/AppearanceSettings.cs
@@ -25,9 +25,9 @@ public class AppearanceSettings
     public AppearanceSettings()
     {
         _properties = new ExpandoObject();
-        if (_substitutinator == null)
+        if (_appearanceMigrator == null)
         {
-            _substitutinator = new AppearanceMigrator();
+            _appearanceMigrator = new AppearanceMigrator();
         }
 
         // copy in the default values from each definition
@@ -43,7 +43,7 @@ public class AppearanceSettings
     public static string kDoShowValueForDisplay = "doShow-css-will-ignore-this-and-use-default"; // by using an illegal value, we just get a no-op rule, which is what we want
     public static string kHideValueForDisplay = "none";
     public static string kOverrideGroupsArrayKey = "groupsToOverrideFromParent"; // e.g. "coverFields, xmatter"
-    private static AppearanceMigrator _substitutinator;
+    private static AppearanceMigrator _appearanceMigrator;
 
     // A representation of the content of Appearance.json
     internal dynamic _properties;
@@ -108,7 +108,20 @@ public class AppearanceSettings
         // The default here is rarely if ever relevant. Usually a newly created instance will be initialized from a folder, and the default will be overwritten,
         // either to whatever we find in appearance.json, or to "legacy-5-6" if there is no appearance.json.
         new StringPropertyDef("cssThemeName", "cssThemeName", "default"),
+        // BooleanPropertyDef, not CssXDef because this is not a css variable.
+        new BooleanPropertyDef(
+            "coverIsImage",
+            "coverIsImage",
+            defaultValue: false,
+            requiresXmatterUpdate: true,
+            valueRequiredIfLegacyTheme: false
+        ), // If true, cover page is just a full bleed image.
+        // Not implemented yet. When it is, it needs to be tied together with the coverIsImage setting
+        // such that coverIsImage cannot be true unless fullBleed is also true.
+        //new BooleanPropertyDef("fullBleed", "fullBleed", false), // If true, book is full bleed.
+
         new CssStringVariableDef("cover-background-color", "colors"),
+        // User can turn the visibility of these fields on and off in the dialog.
         new CssDisplayVariableDef("cover-title-L1-show", "coverFields", true),
         new CssDisplayVariableDef("cover-title-L2-show", "coverFields", true),
         new CssDisplayVariableDef("cover-title-L3-show", "coverFields", false),
@@ -156,8 +169,68 @@ public class AppearanceSettings
     public string CssThemeName
     {
         get { return _properties.cssThemeName; }
-        set { _properties.cssThemeName = value; }
+        set
+        {
+            _properties.cssThemeName = value;
+            SetRequiredValuesIfLegacyTheme();
+        }
     }
+
+    // Some setting's values are not allowed in legacy mode.
+    // REVIEW:
+    // This concept of forcing a value based on the legacy theme was introduced at the time coverIsImage was added.
+    // And currently (Dec 2024), it is the only property that has a valueRequiredIfLegacyTheme.
+    // But I think the properties which existed before that and which get disabled by setting the theme
+    // to legacy should also be set. e.g. cover-topic-show
+    // Without this, the user can set the theme to non-legacy, change the property to whatever he wants,
+    // then change the theme back to legacy.
+    private void SetRequiredValuesIfLegacyTheme()
+    {
+        if (CssThemeName != "legacy-5-6") // Can't use UsingLegacy here because it includes logic about the syncing of files which we don't want.
+            return;
+
+        foreach (var propertyDefinition in propertyDefinitions)
+        {
+            if (propertyDefinition.ValueRequiredIfLegacyTheme != null)
+            {
+                SetProperty(
+                    new KeyValuePair<string, object>(
+                        propertyDefinition.Name,
+                        propertyDefinition.ValueRequiredIfLegacyTheme
+                    )
+                );
+            }
+        }
+    }
+
+    private void SetProperty(KeyValuePair<string, object> property)
+    {
+        if (
+            !Properties.ContainsKey(property.Key)
+            || !Properties[property.Key].Equals(property.Value)
+        )
+        {
+            var propDef = propertyDefinitions.FirstOrDefault(pd => pd.Name == property.Key);
+            if (propDef?.RequiresXmatterUpdate == true)
+                PendingChangeRequiresXmatterUpdate = true;
+        }
+
+        Properties[property.Key] = property.Value;
+    }
+
+    public bool CoverIsImage
+    {
+        get { return _properties.coverIsImage; }
+    }
+
+    //public bool FullBleed
+    //{
+    //    get { return _properties.fullBleed; }
+    //    set { _properties.fullBleed = value; }
+    //}
+
+    // When this is set to true, we ensure that the xmatter is updated before saving the book.
+    public bool PendingChangeRequiresXmatterUpdate;
 
     /// <summary>
     /// Usually, this is simply the theme name, but if the book doesn't have one (that is, it was made by
@@ -682,9 +755,9 @@ public class AppearanceSettings
                 }
             }
 
-            if (definition is CssPropertyDef)
+            if (definition is CssPropertyDef cssPropertydefinition)
             {
-                var setting = ((PropertyDef)definition).GetCssVariableDeclaration(keyValuePair);
+                var setting = cssPropertydefinition.GetCssVariableDeclaration(keyValuePair);
                 if (!string.IsNullOrEmpty(setting))
                     cssBuilder.AppendLine("\t" + setting);
             }
@@ -801,13 +874,12 @@ public class AppearanceSettings
     {
         // parse the json into an object
         var x = JsonConvert.DeserializeObject<ExpandoObject>(json);
-        //and then for each property, copy into the _properties object
-        // For backwards capabilty, if the json we are reading has a null for a value,
-        // do not override the default value that we already have loaded.
+
+        // and then for each property, copy into the Properties object.
         foreach (var property in (IDictionary<string, object>)x)
-        {
-            Properties[property.Key] = property.Value;
-        }
+            SetProperty(property);
+
+        SetRequiredValuesIfLegacyTheme();
     }
 
     /// <summary>
@@ -971,6 +1043,8 @@ public abstract class PropertyDef
 {
     public string Name;
     public dynamic DefaultValue;
+    public bool RequiresXmatterUpdate;
+    public object ValueRequiredIfLegacyTheme;
 
     public void SetDefault(dynamic prop)
     {
@@ -981,25 +1055,49 @@ public abstract class PropertyDef
     /// The name of the group of properties that can a book can override from a collection, or a page can override from a book.
     /// </summary>
     public string OverrideGroup;
-
-    public abstract string GetCssVariableDeclaration(dynamic property);
 }
 
-public abstract class CssPropertyDef : PropertyDef { }
-
+// StringPropertyDefs and BooleanPropertyDefs get written to appearance.json but not appearance.css
 public class StringPropertyDef : PropertyDef
 {
-    public StringPropertyDef(string name, string overrideGroup, string defaultValue)
+    public StringPropertyDef(
+        string name,
+        string overrideGroup,
+        string defaultValue,
+        bool requiresXmatterUpdate = false,
+        object valueRequiredIfLegacyTheme = null
+    )
     {
         Name = name;
         DefaultValue = defaultValue;
         OverrideGroup = overrideGroup;
+        RequiresXmatterUpdate = requiresXmatterUpdate;
+        ValueRequiredIfLegacyTheme = valueRequiredIfLegacyTheme;
     }
+}
 
-    public override string GetCssVariableDeclaration(dynamic property)
+public class BooleanPropertyDef : PropertyDef
+{
+    public BooleanPropertyDef(
+        string name,
+        string overrideGroup,
+        bool defaultValue,
+        bool requiresXmatterUpdate = false,
+        object valueRequiredIfLegacyTheme = null
+    )
     {
-        return $"--{Name}: {property.Value};";
+        Name = name;
+        OverrideGroup = overrideGroup;
+        DefaultValue = defaultValue;
+        RequiresXmatterUpdate = requiresXmatterUpdate;
+        ValueRequiredIfLegacyTheme = valueRequiredIfLegacyTheme;
     }
+}
+
+// CssPropertyDefs get written to appearance.json and appearance.css
+public abstract class CssPropertyDef : PropertyDef
+{
+    public abstract string GetCssVariableDeclaration(dynamic property);
 }
 
 public class CssStringVariableDef : CssPropertyDef
--- a/src/BloomExe/Book/Book.cs
+++ b/src/BloomExe/Book/Book.cs
@@ -1031,10 +1031,7 @@ namespace Bloom.Book
                 return;
             }
             _pagesCache = null;
-            string oldMetaData = string.Empty;
 
-            // This does its own check for whether it needs to be done, and updates the state
-            // after doing anything it needs to.
             EnsureUpToDateMemory(progress);
 
             Storage.MigrateToMediaLevel1ShrinkLargeImages();
@@ -2447,7 +2444,9 @@ namespace Bloom.Book
                 layout,
                 BookInfo.UseDeviceXMatter,
                 _bookData.MetadataLanguage1Tag,
-                oldIds
+                oldIds,
+                BookInfo.AppearanceSettings.CoverIsImage
+                    && CollectionSettings.HaveEnterpriseFeatures
             );
 
             var dataBookLangs = bookDOM.GatherDataBookLanguages();
@@ -3979,8 +3978,13 @@ namespace Bloom.Book
         }
 
         public bool FullBleed =>
-            BookData.GetVariableOrNull("fullBleed", "*").Xml == "true"
-            && CollectionSettings.HaveEnterpriseFeatures;
+            (
+                // Wants to be
+                // BookInfo.AppearanceSettings.FullBleed
+                // but we haven't put that in the book settings yet.
+                BookData.GetVariableOrNull("fullBleed", "*").Xml == "true"
+                || BookInfo.AppearanceSettings.CoverIsImage
+            ) && CollectionSettings.HaveEnterpriseFeatures;
 
         /// <summary>
         /// Save the page content to the DOM.
@@ -4666,6 +4670,29 @@ namespace Bloom.Book
                 // current book location.
                 PageTemplateSource = Path.GetFileName(FolderPath);
             }
+
+            // This doesn't seem like the best place for this code.
+            // For example, it would probably be better to do immediately when saving book settings dialog.
+            // But this is the only place I could find to plumb things to actually get it to work.
+            // Other attempts to break into the complicated book save cycle were unsuccessful.
+            if (BookInfo.AppearanceSettings.PendingChangeRequiresXmatterUpdate)
+            {
+                // Yes, calling EnsureUpToDateMemory() is unfortunately overkill.
+                // And an unfortunate expansion of the use of this version of the method.
+                // It would be great if there was a way to just update the XMatter;
+                // in theory, that would be BringXmatterHtmlUpToDate().
+                // But just calling that leaves several things undone, including
+                // - calling either UpdateVariablesAndDataDivThroughDOM() or SynchronizeDataItemsThroughoutDOM()
+                //    (such that data-book values are lost)
+                // - calling UpdatePageNumberAndSideClassOfPages()
+                //    (such that side-right/left classes are lost)
+                // If we knew it was just those, we could call them here instead
+                // (we can even move this above the call to UpdateVariablesAndDataDivThroughDOM above).
+                // But there is a whole slew of things EnsureUpToDateMemory() does after calling BringXmatterHtmlUpToDate()
+                // and I wouldn't have any confidence that some of the rest of it is not needed here as well.
+                EnsureUpToDateMemory(new NullProgress());
+            }
+            BookInfo.AppearanceSettings.PendingChangeRequiresXmatterUpdate = false;
 
             try
             {
--- a/src/BloomExe/Book/Page.cs
+++ b/src/BloomExe/Book/Page.cs
@@ -21,7 +21,7 @@ namespace Bloom.Book
         bool IsBackMatter { get; }
         bool IsXMatter { get; }
         bool IsCoverPage { get; }
-        bool IsFrontCoverPage { get; }
+        bool IsOutsideFrontCoverPage { get; }
         string GetCaptionOrPageNumber(ref int pageNumber, out string captionI18nId);
         int GetIndex();
         string IdOfFirstAncestor { get; }
@@ -140,9 +140,12 @@ namespace Bloom.Book
             get { return XMatterHelper.IsCoverPage(_getDivNodeForThisPageMethod(this)); }
         }
 
-        public bool IsFrontCoverPage
+        public bool IsOutsideFrontCoverPage
         {
-            get { return XMatterHelper.IsFrontCoverPage(_getDivNodeForThisPageMethod(this)); }
+            get
+            {
+                return XMatterHelper.IsOutsideFrontCoverPage(_getDivNodeForThisPageMethod(this));
+            }
         }
 
         public string GetCaptionOrPageNumber(ref int pageNumber, out string captionI18nId)
--- a/src/BloomExe/Book/XMatterHelper.cs
+++ b/src/BloomExe/Book/XMatterHelper.cs
@@ -22,6 +22,7 @@ namespace Bloom.Book
     {
         private readonly HtmlDom _bookDom;
         private readonly string _nameOfXMatterPack;
+        private readonly IFileLocator _fileLocator;
 
         /// <summary>
         /// Constructs by finding the file and folder of the xmatter pack, given the its key name e.g. "Factory", "SILIndonesia".
@@ -40,6 +41,7 @@ namespace Bloom.Book
             bool useDeviceVersionIfAvailable = false
         )
         {
+            _fileLocator = fileLocator;
             string directoryPath = null;
             _bookDom = bookDom;
             var bookSpecificXMatterPack = bookDom.GetMetaValue("xmatter", null);
@@ -316,7 +318,8 @@ namespace Bloom.Book
             Layout layout,
             bool orderXmatterForDeviceUse,
             string metadataLangTag,
-            List<string> oldIds = null
+            List<string> oldIds = null,
+            bool coverIsImage = false
         )
         {
             //don't want to pollute shells with this content
@@ -358,7 +361,30 @@ namespace Bloom.Book
                 )
             )
             {
-                var newPageDiv = _bookDom.RawDom.ImportNode(xmatterPage, true) as SafeXmlElement;
+                var newPageSource = xmatterPage;
+
+                // If we are using an image for the front cover, replace the typical front cover with
+                // a special one which has a full-page image container.
+                if (coverIsImage && IsOutsideFrontCoverPage(xmatterPage))
+                {
+                    var directoryPath = GetXMatterDirectory(
+                        "CoverIsImage",
+                        _fileLocator,
+                        null,
+                        true
+                    );
+                    var coverIsImageDom = XmlHtmlConverter.GetXmlDomFromHtmlFile(
+                        directoryPath.CombineForPath("CoverIsImage-XMatter.html"),
+                        false
+                    );
+                    var coverIsImagePage = coverIsImageDom.SelectSingleNode(
+                        "/html/body/div[contains(@data-page,'required')]"
+                    );
+                    newPageSource = coverIsImagePage as SafeXmlElement;
+                }
+
+                var newPageDiv = _bookDom.RawDom.ImportNode(newPageSource, true) as SafeXmlElement;
+
                 // If we're updating an existing book, we want to keep the IDs (as much as possible; sometimes
                 // the number of xmatter pages changes and we have to add IDs). In this case, oldIds is obtained
                 // while running RemoveExistingXMatter.
@@ -498,37 +524,32 @@ namespace Bloom.Book
 
         public static bool IsFrontMatterPage(SafeXmlElement pageDiv)
         {
-            return pageDiv.SelectSingleNode("self::div[contains(@class, 'bloom-frontMatter')]")
-                != null;
+            return pageDiv.HasClass("bloom-frontMatter");
+        }
+
+        public static bool IsOutsideFrontCoverPage(SafeXmlElement pageDiv)
+        {
+            return pageDiv.HasClass("outsideFrontCover");
         }
 
         public static bool IsBackMatterPage(SafeXmlElement pageDiv)
         {
-            return pageDiv.SelectSingleNode("self::div[contains(@class, 'bloom-backMatter')]")
-                != null;
+            return pageDiv.HasClass("bloom-backMatter");
         }
 
         public static bool IsXMatterPage(SafeXmlElement pageDiv)
         {
-            return pageDiv.SelectSingleNode(
-                    "self::div[contains(@class, 'bloom-frontMatter') or contains(@class, 'bloom-backMatter')]"
-                ) != null;
+            return IsFrontMatterPage(pageDiv) || IsBackMatterPage(pageDiv);
         }
 
         public static bool IsCoverPage(SafeXmlElement pageDiv)
         {
-            return pageDiv.SelectSingleNode("self::div[contains(@class, 'cover')]") != null;
-        }
-
-        public static bool IsFrontCoverPage(SafeXmlElement pageDiv)
-        {
-            return pageDiv.SelectSingleNode("self::div[contains(@class, 'outsideFrontCover')]")
-                != null;
+            return pageDiv.HasClass("cover");
         }
 
         public static bool ShouldBeInBackForDeviceUse(SafeXmlElement pageDiv)
         {
-            return pageDiv.SelectSingleNode("self::div[contains(@class, 'frontCover')]") == null;
+            return !pageDiv.HasClass("frontCover");
         }
 
         /// <summary>
--- a/src/BloomExe/Book/XMatterPackFinder.cs
+++ b/src/BloomExe/Book/XMatterPackFinder.cs
@@ -39,7 +39,8 @@ namespace Bloom.Book
             "shrp",
             "sharp",
             "forunittest",
-            "templatestarter"
+            "templatestarter",
+            "coverisimage",
         };
 
         public IEnumerable<XMatterInfo> GetXMattersToOfferInSettings(
--- a/src/BloomExe/Edit/EditingModel.cs
+++ b/src/BloomExe/Edit/EditingModel.cs
@@ -702,9 +702,9 @@ namespace Bloom.Edit
             }
         }
 
-        private bool IsFrontCover
+        private bool IsOutsideFrontCover
         {
-            get { return _pageSelection?.CurrentSelection?.IsFrontCoverPage ?? false; }
+            get { return _pageSelection?.CurrentSelection?.IsOutsideFrontCoverPage ?? false; }
         }
 
         // For now, the same rules govern copying hyperlinks. Working hyperlinks to xmatter pages are difficult,
@@ -712,7 +712,7 @@ namespace Bloom.Edit
         // a different strategy for identifying them. And then, choosing a different xmatter pack might cause a page
         // to cease to exist altogether.
         // The only exception is the front cover page, which uses a magic value instead of the page id.
-        public bool CanCopyHyperlink => CanCopyPage || IsFrontCover;
+        public bool CanCopyHyperlink => CanCopyPage || IsOutsideFrontCover;
 
         public bool CanDeletePage
         {
@@ -1940,7 +1940,7 @@ namespace Bloom.Edit
 
         public void CopyHyperlink(IPage page)
         {
-            var pageId = page.IsFrontCoverPage
+            var pageId = page.IsOutsideFrontCoverPage
                 ? "cover"
                 : page.GetDivNodeForThisPage().GetAttribute("id");
             var bookId = _bookSelection.CurrentSelection.ID;
--- a/src/BloomExe/Edit/PlaceHolderPage.cs
+++ b/src/BloomExe/Edit/PlaceHolderPage.cs
@@ -88,7 +88,7 @@ namespace Bloom.Edit
             get { return false; }
         }
 
-        public bool IsFrontCoverPage
+        public bool IsOutsideFrontCoverPage
         {
             get { return false; }
         }
--- a/src/BloomExe/web/controllers/BookSettingsApi.cs
+++ b/src/BloomExe/web/controllers/BookSettingsApi.cs
@@ -3,7 +3,6 @@ using System.Collections.Generic;
 using System.Diagnostics;
 using System.Dynamic;
 using System.IO;
-using System.Linq;
 using Bloom.Book;
 using Bloom.Edit;
 using Bloom.web.controllers;
--- a/src/content/bookLayout/basePage.less
+++ b/src/content/bookLayout/basePage.less
@@ -1,18 +1,18 @@
 @import "basePage-sharedRules.less";
 @import "bubble.less";
 @import (reference) "../../BloomBrowserUI/bloomUI.less"; // importing by reference only brings in the constants
-// NOTE: more style sheet imports are at the end of this document,
-// ones that should supersede the normal rules
-
 @import "pageBoxesSizing.less";
 @import "marginBox.less";
 @import "pageNumbers.less";
+// NOTE: more style sheet imports come later in this document, ones that should supersede the normal rules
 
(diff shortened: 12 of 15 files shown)


hatton (commented):
> Reviewed 2 of 12 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6602)* status: 2 of 12 files reviewed, 1 unresolved discussion (waiting on @andrew-polk)

___
*[`src/BloomBrowserUI/bookEdit/bookSettings/BookSettingsDialog.tsx` line 202 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6602#-O4klo0MEkjTHpg5FvUi:-O4klo0MEkjTHpg5FvUj:b-hf136l) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/a261bc1555c0aaaedf550940ce202a531bb898d6/src/BloomBrowserUI/bookEdit/bookSettings/BookSettingsDialog.tsx#L202)):*
> ```XML
>     );
>     const coverIsImageDescription = useL10n(
>         "For paper books, this image will be full bleed.",
> ```

What can we say that will help people who don't know that term}  I'm tempted to just not say anything. Best try:
"Using this option turns on the [Print Bleed](https://docs.bloomlibrary.org) indicators on paper layouts".

hatton (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6602)* status: 2 of 12 files reviewed, 1 unresolved discussion (waiting on @andrew-polk)

___
*[`src/BloomBrowserUI/bookEdit/bookSettings/BookSettingsDialog.tsx` line 202 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6602#-O4klo0MEkjTHpg5FvUi:-O4kmjlF3nMtimrNAmYa:b-qzgj5k) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/a261bc1555c0aaaedf550940ce202a531bb898d6/src/BloomBrowserUI/bookEdit/bookSettings/BookSettingsDialog.tsx#L202)):*

What can we say that will help people who don't know that term}  I'm tempted to just not say anything. Best try:
"Using this option turns on the [Print Bleed](https://docs.bloomlibrary.org) indicators on paper layouts".
</blockquote></details>

Yeah that works. Or "Enabling this option". Shall we also add a pointer to the inevitable docs?

"Using this option turns on the [Print Bleed](https://docs.bloomlibrary.org) indicators on paper layouts. See [Full Page Cover Images](https://docs.bloomlibrary.org) for information on sizing your image to fit.".

andrew-polk:
*[`src/BloomBrowserUI/bookEdit/bookSettings/BookSettingsDialog.tsx` line 202 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6602#-O4klo0MEkjTHpg5FvUi:-O4lqNym723S-IV9giw7:b-i9efql) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/a261bc1555c0aaaedf550940ce202a531bb898d6/src/BloomBrowserUI/bookEdit/bookSettings/BookSettingsDialog.tsx#L202)):*

Yeah that works. Or "Enabling this option". Shall we also add a pointer to the inevitable docs?

"Using this option turns on the [Print Bleed](https://docs.bloomlibrary.org) indicators on paper layouts. See [Full Page Cover Images](https://docs.bloomlibrary.org) for information on sizing your image to fit.".
</blockquote></details>

Is "Print Bleed" a link to an external website?
More and more, I'm liking all our links going to our own docs where we can direct people to external sites.

andrew-polk (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6602)* status: 2 of 12 files reviewed, 1 unresolved discussion (waiting on @hatton)

___
*[`src/BloomBrowserUI/bookEdit/bookSettings/BookSettingsDialog.tsx` line 202 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6602#-O4klo0MEkjTHpg5FvUi:-O4m1iJZ3G21zGKzbmGp:bj4gg3u) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/a261bc1555c0aaaedf550940ce202a531bb898d6/src/BloomBrowserUI/bookEdit/bookSettings/BookSettingsDialog.tsx#L202)):*

Is "Print Bleed" a link to an external website?
More and more, I'm liking all our links going to our own docs where we can direct people to external sites.
</blockquote></details>

I've made the change and put a TODO in the code for actual links.

hatton (commented):
> Reviewed 7 of 12 files at r1, 2 of 3 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6602)* status: 11 of 12 files reviewed, 2 unresolved discussions (waiting on @andrew-polk)

___
*[`src/BloomExe/Book/AppearanceSettings.cs` line 112 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6602#-O4pf7_uET3DDvAjHMFu:-O4pf7_uET3DDvAjHMFv:bqdddgi) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/e644cd75ffea5bd8b60aa9faf9410cdba675bc1a/src/BloomExe/Book/AppearanceSettings.cs#L112)):*
> ```Smalltalk
>         new StringPropertyDef("cssThemeName", "cssThemeName", "default"),
>         // Cover page is just a full bleed image.
>         new BooleanPropertyDef("coverIsImage", "coverIsImage", false),
> ```

Please add a comment explaining why this needs to be a `BooleanPropertyDef` and cannot be a `CssPropertyDef`, thx.
___
*[`src/BloomExe/Book/AppearanceSettings.cs` line 1035 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6602#-O4pewrU2n7aRrQBqxLF:-O4pewrU2n7aRrQBqxLG:b-gpvbtl) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/e644cd75ffea5bd8b60aa9faf9410cdba675bc1a/src/BloomExe/Book/AppearanceSettings.cs#L1035)):*
> ```Smalltalk
> }
> 
> public class BooleanPropertyDef : PropertyDef
> ```

Please add a comment explaining how this is different from `CssPropertyDef`

hatton (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6602)* status: 11 of 12 files reviewed, 3 unresolved discussions (waiting on @andrew-polk)

___
*[`src/BloomExe/Book/AppearanceSettings.cs` line 208 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6602#-O4pgbN6CJP5JEhvxeJk:-O4pgbN6CJP5JEhvxeJl:b-gxujaw) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/e644cd75ffea5bd8b60aa9faf9410cdba675bc1a/src/BloomExe/Book/AppearanceSettings.cs#L208)):*
> ```Smalltalk
>     public bool PendingChangeRequiresXmatterUpdate;
> 
>     private static readonly string[] s_propertiesWhichRequireXmatterUpdate = new string[]
> ```

Please add a comment explaining why this one property is different. Every existing property has defaults, many of them don't make sense in legacy, but we didn't have to introduce all this to support them.

hatton (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6602)* status: 11 of 12 files reviewed, 3 unresolved discussions (waiting on @andrew-polk)

___
*[`src/BloomExe/Book/AppearanceSettings.cs` line 215 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6602#-O4phcVr2-gxqGOF9V9j:-O4phcVr2-gxqGOF9V9k:b-gxujaw) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/e644cd75ffea5bd8b60aa9faf9410cdba675bc1a/src/BloomExe/Book/AppearanceSettings.cs#L215)):*
> ```Smalltalk
>         new Dictionary<string, object>
>         {
>             { "coverIsImage", false }
> ```

Please add a comment explaining why this one property is different. Every existing property has defaults, many of them don't make sense in legacy, but we didn't have to introduce all this to support them.

hatton (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6602)* status: 11 of 12 files reviewed, 4 unresolved discussions (waiting on @andrew-polk and @hatton)

___
*[`src/BloomExe/Book/AppearanceSettings.cs` line 208 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6602#-O4phdr99l0YOcRfdEmU:-O4phdr99l0YOcRfdEmV:bz731va) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/e644cd75ffea5bd8b60aa9faf9410cdba675bc1a/src/BloomExe/Book/AppearanceSettings.cs#L208)):*
> ```Smalltalk
>     public bool PendingChangeRequiresXmatterUpdate;
> 
>     private static readonly string[] s_propertiesWhichRequireXmatterUpdate = new string[]
> ```

Instead of having a new list, would it be cleaner to add a new property to exiting definitions (`CssStringVariableDef` or `BooleanPropertyDef`) ?

JohnThomson (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6602)* status: 11 of 12 files reviewed, 12 unresolved discussions (waiting on @andrew-polk and @hatton)

*[a discussion](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6602#-O4pugRa0xYFEZNeL3pE:-O4pugRa0xYFEZNeL3pF:b-w6xdl3) (no related file):*
Did you consider doing this entirely with CSS, so that we don't need to change the xmatter html? I'm not sure it's possible, but I would think pretty much all front cover pages have a marginBox containing a .bloom-imageContainer.data-title. If that's true, would something like this work?
```
.cover-is-image {
  * {display:none}
  .marginBox {
    display: block;
    height: 100%; // etc
    * {display: none}
    .imageContainer.data-title {
      display: block;
      height: 100% // etc
...
```

Apart from simplifying the code if we don't have to reload xmatter, I have a vague feeling that it might be a good thing not to break the expectation that books always have a title on the front cover (even if it's not visible). I don't know of any code that expects that, but it wouldn't greatly surprise me to find some.
Also, if these rules were in basePage.less, the new feature would just automatically not happen in legacy, without any special cases (except perhaps to disable the control).
Another possible benefit is that if the title is still there, even if hidden, a talking book might be able to read it. Don't know whether we want that...we still couldn't highlight it in the proper place inside the cover image...

___
___
*[`src/BloomBrowserUI/bookEdit/bookSettings/BookSettingsDialog.tsx` line 201 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6602#-O4pkif72jsrsDY1Bi1g:-O4pkif72jsrsDY1Bi1h:bdemzvj) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/e644cd75ffea5bd8b60aa9faf9410cdba675bc1a/src/BloomBrowserUI/bookEdit/bookSettings/BookSettingsDialog.tsx#L201)):*
> ```XML
>         "BookSettings.CoverIsImage"
>     );
>     //TODO real links (and change .xlf)
> ```

Is this ToDo still applicable?
___
*[`src/BloomBrowserUI/react_components/requiresBloomEnterprise.tsx` line 524 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6602#-O4plr3f4QTWzZCKUbT7:-O4plr3f4QTWzZCKUbT8:b-6o2l9i) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/e644cd75ffea5bd8b60aa9faf9410cdba675bc1a/src/BloomBrowserUI/react_components/requiresBloomEnterprise.tsx#L524)):*
> ```XML
>             {enterpriseAvailable ? (
>                 <Span l10nKey={"AvailableWithEnterprise"}>
>                     Available with your Enterprise Subscription
> ```

To me this phrase is ambiguous between "You can do this because you have a subscription" and "you could do this if only you had a subscription", with the latter (wrong) interpretation being more likely. @hatton 
___
*[`src/BloomExe/Book/AppearanceSettings.cs` line 184 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6602#-O4psA-T5x0-IX82Rkg0:-O4psA-U2G60Ong8Lqna:b5wsxhx) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/e644cd75ffea5bd8b60aa9faf9410cdba675bc1a/src/BloomExe/Book/AppearanceSettings.cs#L184)):*
> ```Smalltalk
>     }
> 
>     private void EvaluateForXmatterChangeNeeded(KeyValuePair<string, object> property)
> ```

This would be much easier to read if it was clear that the "object" of the KVP is the  value that we are about to change the property to.
You could simply give it two arguments, propertyName and newValue.
Or at least name the argument propertyNameAndNewValue.
Also consider making this responsible for doing the change. It's plausible for a method SetPropertyValue to first evaluate whether the property is actually changing, and if so, figure out whether an xmatter refresh is needed. Both callers immediately proceed to change the value.
___
*[`src/BloomExe/Book/AppearanceSettings.cs` line 212 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6602#-O4pproQE1InmIud52AQ:-O4pproQE1InmIud52AR:b-79r0s9) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/e644cd75ffea5bd8b60aa9faf9410cdba675bc1a/src/BloomExe/Book/AppearanceSettings.cs#L212)):*
> ```Smalltalk
>         "coverIsImage"
>     };
>     private static readonly Dictionary<string, object> s_mapOfPropertyValuesRequiredByLegacyTheme =
> ```

Wants a comment explaining what the keys and values are (I think, property name -> value the property must have in legacy.
___
*[`src/content/bookLayout/basePage.less` line 709 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6602#-O4pnYcyA3P8tMEpgRh6:-O4pnYcyA3P8tMEpgRh7:b-dpf2ud) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/e644cd75ffea5bd8b60aa9faf9410cdba675bc1a/src/content/bookLayout/basePage.less#L709)):*
> ```Less
>     --page-background-color: @dark !important;
> 
>     background-color: @dark !important;
> ```

Why do we even want it? If the image covers the front cover we don't need it for areas outside the image, and if the image has transparent areas, isn't it more likely we want them white than dark?
___
*[`src/content/templates/xMatter/bloom-xmatter-common.less` line 114 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6602#-O4poOLD9uDuxJQR3bBM:-O4poOLD9uDuxJQR3bBN:bxd6ml0) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/e644cd75ffea5bd8b60aa9faf9410cdba675bc1a/src/content/templates/xMatter/bloom-xmatter-common.less#L114)):*
> ```Less
>             margin: 0mm;
>             height: 100%;
>         }
> ```

I'm not clear on the purpose of xmatter-common. Seems like stuff that is common to all x-matter could just be in basePage.css.
In this case, especially, I wonder if that would simplify things because this rule would just not be present in the legacy version of basePage.css, so it automatically wouldn't happen there.
___
*[`src/content/templates/xMatter/CoverIsImage-XMatter/CoverIsImage-XMatter.pug` line 9 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6602#-O4pzVwqA88NFC7vjK_q:-O4pzVwqA88NFC7vjK_r:b-sjrsex) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/e644cd75ffea5bd8b60aa9faf9410cdba675bc1a/src/content/templates/xMatter/CoverIsImage-XMatter/CoverIsImage-XMatter.pug#L9)):*
> ```Pug
> 	body
> 		+page-cover('Front Cover')(data-export='front-matter-cover', data-xmatter-page='frontCover').cover-is-image.no-margin-page.frontCover.outsideFrontCover#b5169df5-6a40-4c52-bd30-4cab45afe0ed
> 			.split-pane-component-inner
> ```

Isn't this an origami thing? I don't see it in the front cover of my current Bloom book.

andrew-polk (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6602)* status: 10 of 12 files reviewed, 12 unresolved discussions (waiting on @hatton and @JohnThomson)

___
*[`src/BloomBrowserUI/bookEdit/bookSettings/BookSettingsDialog.tsx` line 201 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6602#-O4pkif72jsrsDY1Bi1g:-O4q4kE61WRdfRX7sxMS:btr9dc3) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/e644cd75ffea5bd8b60aa9faf9410cdba675bc1a/src/BloomBrowserUI/bookEdit/bookSettings/BookSettingsDialog.tsx#L201)):*

Is this ToDo still applicable?
</blockquote></details>

yes, right now they are just links to the main docs site.
I think the first is supposed to be an external link, and the second will be to a yet-created docs page.
___
*[`src/content/bookLayout/basePage.less` line 709 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6602#-O4pnYcyA3P8tMEpgRh6:-O4q5Qi9EWjR-jxhUoM2:b3d3k2d) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/e644cd75ffea5bd8b60aa9faf9410cdba675bc1a/src/content/bookLayout/basePage.less#L709)):*

Why do we even want it? If the image covers the front cover we don't need it for areas outside the image, and if the image has transparent areas, isn't it more likely we want them white than dark?
</blockquote></details>

Good question.

Regarding the image filling the container:
The way I originally implemented this was with black bars if the image wasn't the same aspect ratio as the page (`object-fit: contain;`). But then I decided to do what Paper Comic pages do and fill the container (`object-fit: cover;`).
When I asked JH, he wasn't sure which was desirable.
@hatton, if you think `object-fit: cover;` is right, we might be able to remove this. But see below.

Regarding transparency:
I admit I hadn't thought of how transparent images would be affected.
If we didn't do this and left the image transparent, they won't get white, either, but rather the cover color.
So, for a transparent image, do we want cover color, dark gray, or white?
FWIW, 
- Current digital and paper comic (when using the respective template pages):
    - cover: black (because that is the cover color)
    - content page: white
___
*[`src/content/templates/xMatter/CoverIsImage-XMatter/CoverIsImage-XMatter.pug` line 9 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6602#-O4pzVwqA88NFC7vjK_q:-O4q8tIQ2VOY6CC1WliY:b-xkfaxs) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/e644cd75ffea5bd8b60aa9faf9410cdba675bc1a/src/content/templates/xMatter/CoverIsImage-XMatter/CoverIsImage-XMatter.pug#L9)):*

Isn't this an origami thing? I don't see it in the front cover of my current Bloom book.
</blockquote></details>

Good catch. I must have started with paper comic page instead of an existing cover.

andrew-polk (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6602)* status: 10 of 12 files reviewed, 12 unresolved discussions (waiting on @hatton and @JohnThomson)

*[a discussion](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6602#-O4pugRa0xYFEZNeL3pE:-O4qBHb1-yzGCeyvU_tQ:beixid4) (no related file):*

Did you consider doing this entirely with CSS, so that we don't need to change the xmatter html? I'm not sure it's possible, but I would think pretty much all front cover pages have a marginBox containing a .bloom-imageContainer.data-title. If that's true, would something like this work?
```
.cover-is-image {
  * {display:none}
  .marginBox {
    display: block;
    height: 100%; // etc
    * {display: none}
    .imageContainer.data-title {
      display: block;
      height: 100% // etc
...
```

Apart from simplifying the code if we don't have to reload xmatter, I have a vague feeling that it might be a good thing not to break the expectation that books always have a title on the front cover (even if it's not visible). I don't know of any code that expects that, but it wouldn't greatly surprise me to find some.
Also, if these rules were in basePage.less, the new feature would just automatically not happen in legacy, without any special cases (except perhaps to disable the control).
Another possible benefit is that if the title is still there, even if hidden, a talking book might be able to read it. Don't know whether we want that...we still couldn't highlight it in the proper place inside the cover image...
</blockquote></details>

I did make an attempt to do this via css rather than changing the DOM.
A few points came up which ultimately drove me the other way:
1. There were little corners to deal with in various places like qtips showing up for text boxes which were there but covered by the image.
    - JH pointed out that we may have lots of those things coming up like toolbox interactions with text boxes. So, my understanding was that it would undesirable for the Talking Book tool to try to interact with these text boxes. That was actually the final straw which pushed me fully to the DOM approach. Of course, if we decided we do want the user to be able record the title and have it play on the cover, that would be a strong argument for the css approach.
2. We can't have a pure css solution, anyway, since the appearance settings currently has no way set a class (and rules can't be turned on/off by the value of a css variable.
    - We talked briefly about how we would introduce a generic way for an appearance setting to turn a class on or off. We would go back to that approach if we go this direction.

I didn't try implementing it exactly the way you have it above. I'll give a shot again as I think it does simplify some of the things I was struggling with in my css approach. It may even be that if the divs are display:none, their qtips don't display (and perhaps even the Talking Book tool ignores them?

___

andrew-polk (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6602)* status: 10 of 12 files reviewed, 12 unresolved discussions (waiting on @hatton and @JohnThomson)

___
*[`src/BloomExe/Book/AppearanceSettings.cs` line 184 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6602#-O4psA-T5x0-IX82Rkg0:-O4qKytq9ZfWWH7PqMZw:b-3bat9e) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/e644cd75ffea5bd8b60aa9faf9410cdba675bc1a/src/BloomExe/Book/AppearanceSettings.cs#L184)):*

This would be much easier to read if it was clear that the "object" of the KVP is the  value that we are about to change the property to.
You could simply give it two arguments, propertyName and newValue.
Or at least name the argument propertyNameAndNewValue.
Also consider making this responsible for doing the change. It's plausible for a method SetPropertyValue to first evaluate whether the property is actually changing, and if so, figure out whether an xmatter refresh is needed. Both callers immediately proceed to change the value.
</blockquote></details>

I've attempted to improve things.
___
*[`src/BloomExe/Book/AppearanceSettings.cs` line 208 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6602#-O4phdr99l0YOcRfdEmU:-O4qL7lr0HUyPX2Jehcm:b1kv63v) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/e644cd75ffea5bd8b60aa9faf9410cdba675bc1a/src/BloomExe/Book/AppearanceSettings.cs#L208)):*

Instead of having a new list, would it be cleaner to add a new property to exiting definitions (`CssStringVariableDef` or `BooleanPropertyDef`) ?
</blockquote></details>

I wanted to do it that way, but at the time we are changing values, we no longer have access to our `PropertyDef`s. Or at least, I couldn't figure out how to access it.
___
*[`src/BloomExe/Book/AppearanceSettings.cs` line 212 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6602#-O4pproQE1InmIud52AQ:-O4qLYRb5UCSBdAMGtV5:bbrg6ta) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/e644cd75ffea5bd8b60aa9faf9410cdba675bc1a/src/BloomExe/Book/AppearanceSettings.cs#L212)):*

Wants a comment explaining what the keys and values are (I think, property name -> value the property must have in legacy.
</blockquote></details>

I think this is clearer now, but maybe it still wants more commenting?
___
*[`src/BloomExe/Book/AppearanceSettings.cs` line 215 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6602#-O4phcVr2-gxqGOF9V9j:-O4qLPCg80vASD4Tq2Bo:bdi0qrk) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/e644cd75ffea5bd8b60aa9faf9410cdba675bc1a/src/BloomExe/Book/AppearanceSettings.cs#L215)):*

Please add a comment explaining why this one property is different. Every existing property has defaults, many of them don't make sense in legacy, but we didn't have to introduce all this to support them.
</blockquote></details>

Actually, I don't think this one is different. I think it was an oversight in the original implementation. I've added a review comment. Let me know what you think.

andrew-polk:
*[`src/BloomExe/Book/AppearanceSettings.cs` line 208 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6602#-O4phdr99l0YOcRfdEmU:-O4qNZDHCdpj0tAFbKdg:b-tuzpe1) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/e644cd75ffea5bd8b60aa9faf9410cdba675bc1a/src/BloomExe/Book/AppearanceSettings.cs#L208)):*

I wanted to do it that way, but at the time we are changing values, we no longer have access to our `PropertyDef`s. Or at least, I couldn't figure out how to access it.
</blockquote></details>

Hm. I see now that I should just be able to get it by matching key to Name, so I'm not sure what my problem was before.
I'll pursue this after I verify we don't want go back to the css approach.

hatton (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6602)* status: 10 of 12 files reviewed, 12 unresolved discussions (waiting on @andrew-polk and @JohnThomson)

___
*[`src/BloomBrowserUI/react_components/requiresBloomEnterprise.tsx` line 524 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6602#-O4plr3f4QTWzZCKUbT7:-O4r7gIv98Pw2fL3DaDe:b-vlb1mx) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/e644cd75ffea5bd8b60aa9faf9410cdba675bc1a/src/BloomBrowserUI/react_components/requiresBloomEnterprise.tsx#L524)):*

To me this phrase is ambiguous between "You can do this because you have a subscription" and "you could do this if only you had a subscription", with the latter (wrong) interpretation being more likely. @hatton
</blockquote></details>

Is it only ambiguous to you because you don't know if you have an Enterprise Subscription? If you had one, and it was enabled, what would be your confusion?
___
*[`src/BloomExe/Book/AppearanceSettings.cs` line 215 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6602#-O4phcVr2-gxqGOF9V9j:-O4r8CsZ7FN8C4km3gcY:b4eciiy) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/e644cd75ffea5bd8b60aa9faf9410cdba675bc1a/src/BloomExe/Book/AppearanceSettings.cs#L215)):*

Actually, I don't think this one is different. I think it was an oversight in the original implementation. I've added a review comment. Let me know what you think.
</blockquote></details>

Sorry, I don't understand your reply or know what a "review comment" is, expect this conversation here.  The heat mut be getting to me!
___
*[`src/content/bookLayout/basePage.less` line 709 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6602#-O4pnYcyA3P8tMEpgRh6:-O4r8U9VDx053PDLTDZv:b-cb01si) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/e644cd75ffea5bd8b60aa9faf9410cdba675bc1a/src/content/bookLayout/basePage.less#L709)):*

Good question.

Regarding the image filling the container:
The way I originally implemented this was with black bars if the image wasn't the same aspect ratio as the page (`object-fit: contain;`). But then I decided to do what Paper Comic pages do and fill the container (`object-fit: cover;`).
When I asked JH, he wasn't sure which was desirable.
@hatton, if you think `object-fit: cover;` is right, we might be able to remove this. But see below.

Regarding transparency:
I admit I hadn't thought of how transparent images would be affected.
If we didn't do this and left the image transparent, they won't get white, either, but rather the cover color.
So, for a transparent image, do we want cover color, dark gray, or white?
FWIW, 
- Current digital and paper comic (when using the respective template pages):
    - cover: black (because that is the cover color)
    - content page: white
</blockquote></details>

I'm not sure I'm following, but if the image doesn't cover the page in some dimension, the printer should not print anything in the extra areas (e.g. white).

andrew-polk (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6602)* status: 10 of 12 files reviewed, 12 unresolved discussions (waiting on @hatton and @JohnThomson)

___
*[`src/BloomExe/Book/AppearanceSettings.cs` line 215 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6602#-O4phcVr2-gxqGOF9V9j:-O4v7iLj1lu6JSAZM1_h:b-pend2a) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/e644cd75ffea5bd8b60aa9faf9410cdba675bc1a/src/BloomExe/Book/AppearanceSettings.cs#L215)):*

Sorry, I don't understand your reply or know what a "review comment" is, expect this conversation here.  The heat mut be getting to me!
</blockquote></details>

Sorry; I moved the code, so that probably confused things.
I added this in the code before this dictionary is defined:
```
            //REVIEW: I think that some of the other properties should also be here.
            // This concept of forcing a value based on the legacy theme was introduced at the time coverIsImage was added.
            // But I think the properties which existed before that and which get disabled by setting the theme
            // to legacy should also be here.
            // e.g. cover-topic-show
            // Without this, the user can set the theme to non-legacy, change the property to whatever he wants,
            // then change the theme back to legacy.
```
___
*[`src/BloomExe/Book/AppearanceSettings.cs` line 1035 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6602#-O4pewrU2n7aRrQBqxLF:-O4v9Fed1LrGoTtu_sax:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/e644cd75ffea5bd8b60aa9faf9410cdba675bc1a/src/BloomExe/Book/AppearanceSettings.cs#L1035)):*

Please add a comment explaining how this is different from `CssPropertyDef`
</blockquote></details>

Done.

andrew-polk (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6602)* status: 10 of 12 files reviewed, 12 unresolved discussions (waiting on @hatton and @JohnThomson)

___
*[`src/content/templates/xMatter/CoverIsImage-XMatter/CoverIsImage-XMatter.pug` line 9 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6602#-O4pzVwqA88NFC7vjK_q:-O4v9TMdD4o7ATMB0tpm:b-ynwl21) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/e644cd75ffea5bd8b60aa9faf9410cdba675bc1a/src/content/templates/xMatter/CoverIsImage-XMatter/CoverIsImage-XMatter.pug#L9)):*

Good catch. I must have started with paper comic page instead of an existing cover.
</blockquote></details>

I removed it.

andrew-polk:
*[`src/content/bookLayout/basePage.less` line 709 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6602#-O4pnYcyA3P8tMEpgRh6:-O4wX0_J0CzFeba5avfg:bz6avza) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/e644cd75ffea5bd8b60aa9faf9410cdba675bc1a/src/content/bookLayout/basePage.less#L709)):*
> if the image doesn't cover the page in some dimension, the printer should not print anything in the extra areas (e.g. white).

Apparently we made a very specific decision at some point in the past that the PDF for a full bleed book should include the cover color, not white as non-full-bleed books do.
https://github.com/BloomBooks/BloomDesktop/commit/aa581d14fabe41c5deff63e383f4a40bd2388c8e

I think that makes sense if you have a normal cover with stuff besides the image. You're sending the cover color you want printed to the printer. (And the user can alway set the cover color to white if desired.)
Are you saying we want to make an exception for this cover-is-image case?

I think its mostly a moot point currently. With the `object-fit: cover;`, the image will always cover the whole page, and only transparency will matter. And I can't imagine a real use case of cover-is-image with a transparent image.

andrew-polk (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6602)* status: 3 of 12 files reviewed, 12 unresolved discussions (waiting on @hatton and @JohnThomson)

___
*[a discussion](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6602#-O4pugRa0xYFEZNeL3pE:-OCAeAbu1jGqTQKIPTtK:bnbppmx) (no related file):*

I did make an attempt to do this via css rather than changing the DOM.
A few points came up which ultimately drove me the other way:
1. There were little corners to deal with in various places like qtips showing up for text boxes which were there but covered by the image.
    - JH pointed out that we may have lots of those things coming up like toolbox interactions with text boxes. So, my understanding was that it would undesirable for the Talking Book tool to try to interact with these text boxes. That was actually the final straw which pushed me fully to the DOM approach. Of course, if we decided we do want the user to be able record the title and have it play on the cover, that would be a strong argument for the css approach.
2. We can't have a pure css solution, anyway, since the appearance settings currently has no way set a class (and rules can't be turned on/off by the value of a css variable.
    - We talked briefly about how we would introduce a generic way for an appearance setting to turn a class on or off. We would go back to that approach if we go this direction.

I didn't try implementing it exactly the way you have it above. I'll give a shot again as I think it does simplify some of the things I was struggling with in my css approach. It may even be that if the divs are display:none, their qtips don't display (and perhaps even the Talking Book tool ignores them?
</blockquote></details>

I went back to trying to do this with css and got stuck enough that JH said it wasn't worth it. We decided to go with the DOM swap approach I had implemented orginally.
___
*[`src/BloomExe/Book/AppearanceSettings.cs` line 112 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6602#-O4pf7_uET3DDvAjHMFu:-OCAMIey90RaF2VeG0E7:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/e644cd75ffea5bd8b60aa9faf9410cdba675bc1a/src/BloomExe/Book/AppearanceSettings.cs#L112)):*

Please add a comment explaining why this needs to be a `BooleanPropertyDef` and cannot be a `CssPropertyDef`, thx.
</blockquote></details>

Done.
___
*[`src/BloomExe/Book/AppearanceSettings.cs` line 208 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6602#-O4phdr99l0YOcRfdEmU:-OCATBFi-6suQoX5aoPh:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/e644cd75ffea5bd8b60aa9faf9410cdba675bc1a/src/BloomExe/Book/AppearanceSettings.cs#L208)):*

Hm. I see now that I should just be able to get it by matching key to Name, so I'm not sure what my problem was before.
I'll pursue this after I verify we don't want go back to the css approach.
</blockquote></details>

Done.
___
*[`src/content/templates/xMatter/bloom-xmatter-common.less` line 114 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6602#-O4poOLD9uDuxJQR3bBM:-OCAdGVEBS5TS1LQYKWr:b6nrg8m) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/e644cd75ffea5bd8b60aa9faf9410cdba675bc1a/src/content/templates/xMatter/bloom-xmatter-common.less#L114)):*

I'm not clear on the purpose of xmatter-common. Seems like stuff that is common to all x-matter could just be in basePage.css.
In this case, especially, I wonder if that would simplify things because this rule would just not be present in the legacy version of basePage.css, so it automatically wouldn't happen there.
</blockquote></details>

I could certainly move this to basePage.less (and I don't think I have any principled reason for it to be in one place vs. the other).
But I don't see how it would simplify things.
I think we would still have to prevent the user from turning it on in legacy mode which is what all the code in AppearanceSettings.cs is for.

JohnThomson (commented):
I picked up a few more things to at least question. Maybe being too picky. It could merge as-is, though one of us will have to figure out how it combines with cover-image-is-background-overlay.

> Reviewed 3 of 12 files at r1, 1 of 1 files at r3, 7 of 8 files at r6, 1 of 1 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6602#-:-OD7fMPN4o_lub3YCZ29:bfy2bel)* status: all files reviewed, 13 unresolved discussions (waiting on @andrew-polk and @hatton)

___
*[`src/BloomExe/Book/AppearanceSettings.cs` line 184 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6602#-O4psA-T5x0-IX82Rkg0:-OD7c9nh3Me5IHYW231j:bp2923e) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/e644cd75ffea5bd8b60aa9faf9410cdba675bc1a/src/BloomExe/Book/AppearanceSettings.cs#L184)):*

I've attempted to improve things.
</blockquote></details>

Definitely improved. See my comment below, though.
___
*[`src/BloomExe/Book/AppearanceSettings.cs` line 212 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6602#-O4pproQE1InmIud52AQ:-OD7E9wiCnRCcIdTj2l6:bsscsq8) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/e644cd75ffea5bd8b60aa9faf9410cdba675bc1a/src/BloomExe/Book/AppearanceSettings.cs#L212)):*

I think this is clearer now, but maybe it still wants more commenting?
</blockquote></details>

I still like JohnH's idea that RequiredValueForLegacyTheme could be a field in PropertyDef, assuming that we don't have properties whose value must be null, which at the moment we don't. Then the bulk of SetRequiredValuesIfLegacyTheme would just be a loop over Properties, calling SetProperty(pd.Name, pd.RequiredValueForLegacyTheme) for each of them. (Of course, that requires making SetProperty take the key and value as separate arguments. And either the caller or callee must handle RequiredValueForLegacyTheme being null.)
___
*[`src/content/bookLayout/basePage.less` line 709 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6602#-O4pnYcyA3P8tMEpgRh6:-OD7Umj_EzAUv2638BoX:b-eg59hv) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/e644cd75ffea5bd8b60aa9faf9410cdba675bc1a/src/content/bookLayout/basePage.less#L709)):*

> if the image doesn't cover the page in some dimension, the printer should not print anything in the extra areas (e.g. white).

Apparently we made a very specific decision at some point in the past that the PDF for a full bleed book should include the cover color, not white as non-full-bleed books do.
https://github.com/BloomBooks/BloomDesktop/commit/aa581d14fabe41c5deff63e383f4a40bd2388c8e

I think that makes sense if you have a normal cover with stuff besides the image. You're sending the cover color you want printed to the printer. (And the user can alway set the cover color to white if desired.)
Are you saying we want to make an exception for this cover-is-image case?

I think its mostly a moot point currently. With the `object-fit: cover;`, the image will always cover the whole page, and only transparency will matter. And I can't imagine a real use case of cover-is-image with a transparent image.
</blockquote></details>

Ouch. My coming PR will mess this up, with the cover image becoming a background-overlay that is not using either content-fit or cover, but absolute size and position attributes set by code to make the size of the overlay simulate content-fit. If we want the "cover" effect it will have to be achieved with a special case in the code that sets those attributes.

And it's going to be a pain anyway, because the size adjusting normally only runs (in JS!) when the page is opened. But somehow the cover image will have to get adjusted to the right size and position every time we update the xmatter, even if the user never edits that page, at least before publication. I guess that's my problem, though.

One benefit is that if we do end up with crop controls on the cover, the user will have full control over what part of the image shows. Maybe that means we don't need a special case to make it simulate object-fit:cover?

Those things aside, I'm not seeing any case for new code to force a dark background. We want a special case for paper comic PDFs to show the cover color. We just possibly want a white background for partly-transparent images. Otherwise we want nothing printed where there isn't image.
___
*[`src/content/templates/xMatter/bloom-xmatter-common.less` line 114 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6602#-O4poOLD9uDuxJQR3bBM:-OD7b_cL6GA06zfEvYHW:b-nioxl2) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/e644cd75ffea5bd8b60aa9faf9410cdba675bc1a/src/content/templates/xMatter/bloom-xmatter-common.less#L114)):*

I could certainly move this to basePage.less (and I don't think I have any principled reason for it to be in one place vs. the other).
But I don't see how it would simplify things.
I think we would still have to prevent the user from turning it on in legacy mode which is what all the code in AppearanceSettings.cs is for.
</blockquote></details>

Since you've thoroughly dealt with not allowing this to happen in legacy, I agree it doesn't matter much. And as long as we have an xmatter-common, that's quite possibly a good place to put rules that apply to all xmatter pages for all variations. I think we can leave it.
___
*[`src/BloomExe/Book/XMatterHelper.cs` line 322 at r7](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6602#-OD7LWYf21VS2vLlosoS:-OD7LWYf21VS2vLlosoT:b-5x8aqv) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/86faf05fcfdf128f08bba1657e2122a1f00cf478/src/BloomExe/Book/XMatterHelper.cs#L322)):*
> ```Smalltalk
>             string metadataLangTag,
>             List<string> oldIds = null,
>             bool useCoverIsImage = false
> ```

This apparently means something like "use the front cover page from the special CoverIsImage template instead of the one from the book's own xmatter." That feels like exposing (and requiring understanding of) an implementation detail. How about something like "forceCoverToBeASingleImage"?
But even that isn't what this is heading towards. When my latest change lands, the cover won't be a simple image; it will be an image container with a single (background) image overlay. Someday we will probably allow other overlays on the cover, including (I hope) one identified as the book title. Not sure how that will be implemented; possibly there will be three options in appearance, something like Normal (Front Cover), Single Image, Single Image with moveable title? Or maybe some trick to identify a particular text overlay as the title?
How about coverImageShouldFillPage? Or maybe we get better consistency with other places by just calling it coverIsImage? Or mazimizeCoverImage? (I like that last one because it is clearer that it won't actually fill the page unless it is the right shape; content-fit still applies. But that's true everywhere we use coverIsImage.)
___
*[`src/BloomExe/Book/XMatterHelper.cs` line 387 at r7](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6602#-OD7SqFI6WGpFbERaLJQ:-OD7SqFJ1oa4r-hRpUg4:b-f19rz9) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/86faf05fcfdf128f08bba1657e2122a1f00cf478/src/BloomExe/Book/XMatterHelper.cs#L387)):*
> ```Smalltalk
>                 }
>                 else
>                     newPageDiv = _bookDom.RawDom.ImportNode(xmatterPage, true) as SafeXmlElement;
> ```

Is it any improvement to start out with
var newPageSource = xmatterPage;
if (...) {
...
newPageSource = SelectSingleNode(...)
}
var newPageDiv = _bookDom.RawDom.ImportNode(newPageSource)...
?
Not shorter, but I think it makes it clearer that what's special about the new path is just finding a different source for the page to copy into the current book.
___
*[`src/BloomExe/Book/XMatterHelper.cs` line 536 at r7](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6602#-OD7Tlr7Bgoo3GUj5ntg:-OD7Tlr7Bgoo3GUj5nth:b7nujy9) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/86faf05fcfdf128f08bba1657e2122a1f00cf478/src/BloomExe/Book/XMatterHelper.cs#L536)):*
> ```Smalltalk
>             return pageDiv.SelectSingleNode("self::div[contains(@class, 'outsideFrontCover')]")
>                 != null;
>         }
> ```

SafeXmlElement has a HasClass method that could be used here (or maybe make this method redundant)
___
*[`src/BloomExe/Book/AppearanceSettings.cs` line 872 at r7](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6602#-OD7cmKj7818ojWjJ-wH:-OD7cmKj7818ojWjJ-wI:b-lwgsi6) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/86faf05fcfdf128f08bba1657e2122a1f00cf478/src/BloomExe/Book/AppearanceSettings.cs#L872)):*
> ```Smalltalk
>         // and then for each property, copy into the Properties object.
>         // The original comment here also said: For backwards capability, if the json we are reading has a null for a value, do not override the default value that we already have loaded.
>         //  But the code didn't seem to implement that when I got here. (AP, Aug 2024)
> ```

As far as I can see, it still doesn't. You handle unknown properties and ones that haven't changed, but don't do anything special if the value is null. Did you intend to change this?
___
*[`src/BloomExe/Book/AppearanceSettings.cs` line 876 at r7](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6602#-OD7I_to5jgNnCpapDGg:-OD7I_to5jgNnCpapDGh:bormc4l) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/86faf05fcfdf128f08bba1657e2122a1f00cf478/src/BloomExe/Book/AppearanceSettings.cs#L876)):*
> ```Smalltalk
>             SetProperty(property);
> 
>         SetRequiredValuesIfLegacyTheme();
> ```

I suppose it's OK that this (or the previous line) might set RequiresXmatterUpdate? Does every caller of this function follow it by saving the book, so that RequiresXmatterUpdate gets properly dealt with? Is it worth commenting on this method that it's important to follow it with a Save so that any required xmatter update happens?
___
*[`src/BloomExe/Book/Book.cs` line 4678 at r7](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6602#-OD7JcuRAC9zV9v6aZkf:-OD7JcuRAC9zV9v6aZkg:b-tgo2bm) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/86faf05fcfdf128f08bba1657e2122a1f00cf478/src/BloomExe/Book/Book.cs#L4678)):*
> ```Smalltalk
> 
>             if (BookInfo.AppearanceSettings.PendingChangeRequiresXmatterUpdate)
>                 EnsureUpToDateMemory(new NullProgress());
> ```

I'm not thrilled that we're doing this whole process just to get Xmatter updated, especially since the name EnsureUpToDateMemory implies that nothing will be done if we think it's already up to date. As it happens, this method doesn't currently check that, but I think only because it's a chunk of the main EnsureUpToDate which does. In fact, I think I left it as a separate method only to better preserve our Git change history; it's unfortunate for such an arbitrary chunk to be further established by giving it another caller.
Could we pull out the chunk of EnsureUpToDateMemory that actually updates xmatter and give it a less conditional-sounding name?

andrew-polk:
*[`src/BloomBrowserUI/bookEdit/bookSettings/BookSettingsDialog.tsx` line 201 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6602#-O4pkif72jsrsDY1Bi1g:-OD82oDS4FQSqZFnocsz:b-5sy5tm) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/e644cd75ffea5bd8b60aa9faf9410cdba675bc1a/src/BloomBrowserUI/bookEdit/bookSettings/BookSettingsDialog.tsx#L201)):*

yes, right now they are just links to the main docs site.
I think the first is supposed to be an external link, and the second will be to a yet-created docs page.
</blockquote></details>

I added a checkbox to the card so this doesn't get lost.

JohnThomson (commented):
> Reviewed 6 of 6 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6602)* status: all files reviewed, 10 unresolved discussions (waiting on @andrew-polk and @hatton)

___
*[`src/BloomExe/Book/AppearanceSettings.cs` line 872 at r7](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6602#-OD7cmKj7818ojWjJ-wH:-OD8_-Qf5xheUiiMFym-:bh605go) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/86faf05fcfdf128f08bba1657e2122a1f00cf478/src/BloomExe/Book/AppearanceSettings.cs#L872)):*

No; just stating that the comment seems to be wrong. If you agree, I can just remove it.
</blockquote></details>

It's not doing it, so either the comment or the code was wrong. We don't seem to have missed it, and backwards compatibility can't be much of an issue for appearance.json, so it's probably safe to just remove. I wish we knew what might have produced JSON that needed it, though.
___
*[`src/BloomExe/Book/AppearanceSettings.cs` line 876 at r7](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6602#-OD7I_to5jgNnCpapDGg:-OD8dPJT2a7eP4FlPQNB:bqhjysc) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/86faf05fcfdf128f08bba1657e2122a1f00cf478/src/BloomExe/Book/AppearanceSettings.cs#L876)):*

I don't think we actually do call Save on all paths.
I'll think about this some more. It is tied in with the comments below about calling EnsureUpToDateMemory.
</blockquote></details>

Since you're still thinking I won't resolve this one yet.
___
*[`src/BloomExe/Book/Book.cs` line 4678 at r7](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6602#-OD7JcuRAC9zV9v6aZkf:-OD8ax7V7DLQb_2uu5pV:bqla9qg) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/86faf05fcfdf128f08bba1657e2122a1f00cf478/src/BloomExe/Book/Book.cs#L4678)):*

Sigh. We do have a BringXmatterHtmlUpToDate already, but apparently if we only call that here, we lose the data-book values. E.g. the title is lost. I guess I need to investigate more why that is...
I agree this is some less than ideal code... maybe we can discuss other ideas.

Note, there is a comment on the only caller to EnsureUpToDateMemory which states it does a check to see if it needs to be done, but I don't think that comment is accurate. I added a review comment in the code there. But if you agree, I can just remove the comment.
</blockquote></details>

Sorry to make you do that, but in the interests of improving the code rather than making it worse...seems that BringXmatterHtmlUpToDate SHOULD NOT destroy any book data. (It might plausibly save any data present in the current xmatter that isn't already in the data div, but removing is dangerous for this and other cases: we don't want to lose data-div data just because one xmatter doesn't use it.) It seems likely that we're using a version of BookData that is a member variable of the book, and (without realizing it) depending on something earlier in EnsureUpToDateMemory to initialize it properly...not sure if that idea will help...
___
*[`src/BloomExe/Book/XMatterHelper.cs` line 322 at r7](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6602#-OD7LWYf21VS2vLlosoS:-OD8cCOg49iqV6fArkHW:bi0jnz) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/86faf05fcfdf128f08bba1657e2122a1f00cf478/src/BloomExe/Book/XMatterHelper.cs#L322)):*

Actually, we don't use `object-fit: contain` in this case; we use `object-fit: cover`.
That happens because the `img` in the new pug gets the `bloom-imageObjectFit-cover` class.
So it truly does fill the page.

The only caller which passes this right now calls it with `BookInfo.AppearanceSettings.CoverIsImage && CollectionSettings.HaveEnterpriseFeatures`, so I've gone with `coverIsImage` for now, but open to more discussion.
</blockquote></details>

Good for now, but if you merge my image unifying PR, something will need more attention to preserve the object-fit:cover behavior.
___
*[`src/BloomExe/Book/XMatterHelper.cs` line 387 at r7](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6602#-OD7SqFI6WGpFbERaLJQ:-OD8cYNWA7pvWDP0rGS2:b4unyxx) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/86faf05fcfdf128f08bba1657e2122a1f00cf478/src/BloomExe/Book/XMatterHelper.cs#L387)):*

Yes, that's better. Thanks.
</blockquote></details>

Not important, but if you're doing another pass, there is no longer any need to declare newPageDiv before it is initialized.
___
*[`src/content/bookLayout/basePage.less` line 709 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6602#-O4pnYcyA3P8tMEpgRh6:-OD8d1qDBIkJIuTV7-BR:b-4yzxku) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/e644cd75ffea5bd8b60aa9faf9410cdba675bc1a/src/content/bookLayout/basePage.less#L709)):*

The special case is because we are treating this page as full bleed.
</blockquote></details>

I'm happy to leave this for you and JohnH to work out, but it's not obvious to me why a full-bleed page needs a dark background, so a comment (or perhaps a cross-reference to a comment elsewhere) might be helpful.
___
*[`src/BloomExe/Book/Book.cs` line 1036 at r8](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6602#-OD8ahPY9272Ig5vRBew:-OD8ahPZAHkKQV1AP3ST:bfnms5t) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/4b4dbe583a708d25ce1fc3a3b61af995047ee9de/src/BloomExe/Book/Book.cs#L1036)):*
> ```Smalltalk
>             string oldMetaData = string.Empty;
> 
>             // REVIEW: This comment doesn't seem to be true in Dec 2024:
> ```

Yes, I think this is an old comment, from some (brief) days when we used to do what updating we could without modifying files. Now we don't do any updating unless we can do it all the way.

andrew-polk (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6602)* status: all files reviewed, 10 unresolved discussions (waiting on @hatton and @JohnThomson)

___
*[`src/BloomExe/Book/AppearanceSettings.cs` line 872 at r7](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6602#-OD7cmKj7818ojWjJ-wH:-ODCScOu5GUPQeTEh5ij:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/86faf05fcfdf128f08bba1657e2122a1f00cf478/src/BloomExe/Book/AppearanceSettings.cs#L872)):*

It's not doing it, so either the comment or the code was wrong. We don't seem to have missed it, and backwards compatibility can't be much of an issue for appearance.json, so it's probably safe to just remove. I wish we knew what might have produced JSON that needed it, though.
</blockquote></details>

Done.
___
*[`src/BloomExe/Book/Book.cs` line 1036 at r8](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6602#-OD8ahPY9272Ig5vRBew:-ODCSpC81spYrh32Ls6e:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/4b4dbe583a708d25ce1fc3a3b61af995047ee9de/src/BloomExe/Book/Book.cs#L1036)):*

Yes, I think this is an old comment, from some (brief) days when we used to do what updating we could without modifying files. Now we don't do any updating unless we can do it all the way.
</blockquote></details>

Done.
___
*[`src/BloomExe/Book/XMatterHelper.cs` line 322 at r7](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6602#-OD7LWYf21VS2vLlosoS:-ODDOA_N-ojQaRDSsQBN:b-fft7ib) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/86faf05fcfdf128f08bba1657e2122a1f00cf478/src/BloomExe/Book/XMatterHelper.cs#L322)):*

Good for now, but if you merge my image unifying PR, something will need more attention to preserve the object-fit:cover behavior.
</blockquote></details>

As we discussed on Zulip, this will need the same treatment as full-bleed image pages such as in paper comics.
___
*[`src/BloomExe/Book/XMatterHelper.cs` line 387 at r7](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6602#-OD7SqFI6WGpFbERaLJQ:-ODCSvi15ET5D6_VkjcU:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/86faf05fcfdf128f08bba1657e2122a1f00cf478/src/BloomExe/Book/XMatterHelper.cs#L387)):*

Not important, but if you're doing another pass, there is no longer any need to declare newPageDiv before it is initialized.
</blockquote></details>

Done.

andrew-polk (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6602)* status: 11 of 15 files reviewed, 10 unresolved discussions (waiting on @hatton and @JohnThomson)

___
*[`src/BloomExe/Book/AppearanceSettings.cs` line 876 at r7](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6602#-OD7I_to5jgNnCpapDGg:-ODhEKx08OL_Xfeky_N0:b-4nbfp1) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/86faf05fcfdf128f08bba1657e2122a1f00cf478/src/BloomExe/Book/AppearanceSettings.cs#L876)):*

Since you're still thinking I won't resolve this one yet.
</blockquote></details>

I did add some comments regarding this whole process in Book.cs.
I'm not sure that comments here are particularly helpful. (i.e. at what level(s) would you put them?)
___
*[`src/BloomExe/Book/Book.cs` line 4678 at r7](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6602#-OD7JcuRAC9zV9v6aZkf:-ODhEZxiDsZ9xVdgxnVx:b3an6sk) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/86faf05fcfdf128f08bba1657e2122a1f00cf478/src/BloomExe/Book/Book.cs#L4678)):*

Sorry to make you do that, but in the interests of improving the code rather than making it worse...seems that BringXmatterHtmlUpToDate SHOULD NOT destroy any book data. (It might plausibly save any data present in the current xmatter that isn't already in the data div, but removing is dangerous for this and other cases: we don't want to lose data-div data just because one xmatter doesn't use it.) It seems likely that we're using a version of BookData that is a member variable of the book, and (without realizing it) depending on something earlier in EnsureUpToDateMemory to initialize it properly...not sure if that idea will help...
</blockquote></details>

I wasn't able to find a better way. I added some comments to that end.
___
*[`src/content/bookLayout/basePage.less` line 709 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6602#-O4pnYcyA3P8tMEpgRh6:-ODhF_cQDLYfSmqZpeIK:bpvuut3) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/e644cd75ffea5bd8b60aa9faf9410cdba675bc1a/src/content/bookLayout/basePage.less#L709)):*

I'm happy to leave this for you and JohnH to work out, but it's not obvious to me why a full-bleed page needs a dark background, so a comment (or perhaps a cross-reference to a comment elsewhere) might be helpful.
</blockquote></details>

I figure JH will change it if it affects the UX in an undesirable way.
For now, I've just commented that I'm doing the same thing as paper comic pages.

JohnThomson (approved):
LGTM, however, one thing you did in response to one of JohnH's comments doesn't seem to have gotten committed. Also, the PR that makes all images be background images is now merged, which means cover and other images that are supposed to be object-fit:cover will need attention (not really because of this PR, but the new case of a cover image that is object-fit:cover will also be broken). BubbleManager.adjustBackgroundImageSize needs to somehow figure out that we want this and simulate that instead of contain.
I think it would be quite reasonable to just merge this, but in that case, you should create a card for the problem and cross-reference it from this card and also BL-14069

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6602#-:-ODrOZBN6yb392SSf7Tj:b-dr0wtg)* status: 11 of 15 files reviewed, 4 unresolved discussions (waiting on @hatton)

___
*[`src/BloomExe/Book/AppearanceSettings.cs` line 1035 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6602#-O4pewrU2n7aRrQBqxLF:-ODrO8q95aKVqF7KBZp4:bsaetu8) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/e644cd75ffea5bd8b60aa9faf9410cdba675bc1a/src/BloomExe/Book/AppearanceSettings.cs#L1035)):*

Done.
</blockquote></details>

This was JohnH's comment, and I don't see the need; but you say it's done and I can't find anything new, so I thought I'd just point that out...

PR #6604, opened 2024-08-20
Page number line-height should not be changed by L1 setting (BL-13699)

StephenMcConnel:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6604)

### DIFF
--- a/src/BloomExe/Collection/CollectionSettings.cs
+++ b/src/BloomExe/Collection/CollectionSettings.cs
@@ -328,13 +328,18 @@ namespace Bloom.Collection
             sb.AppendLine(
                 "/* They may be over-ridden by rules in customCollectionStyles.css, appearance.css or customBookStyles[2].css */"
             );
-            // note: css pseudo elements  cannot have a @lang attribute. So this is needed to show page numbers in scripts not covered by Andika New Basic.
+            // Note: css pseudo elements  cannot have a @lang attribute. So this is needed to show page numbers in scripts not
+            // covered by Andika.  The important information needed is the name of the font to use for the page numbers, the
+            // directionality of the script, and possibly how to break lines in the script.  The line-height for page numbers
+            // is set in basePage.css using an appearance system variable, so we don't need to set it here.  Indeed, setting
+            // it here breaks the page numbers in some cases since defaultLangStyles.css is loaded after basePage.css.  See
+            // BL-13699.  (line-height really matters only for multi-line page numbers, which are rare to say the least.)
             WritingSystem.AddSelectorCssRule(
                 sb,
                 ".numberedPage::after",
                 Language1.FontName,
                 Language1.IsRightToLeft,
-                Language1.LineHeight,
+                -1, // Prevents a line-height rule from being created.
                 Language1.BreaksLinesOnlyAtSpaces,
                 omitDirection
             );


andrew-polk (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6604)* status: all files reviewed, 1 unresolved discussion (waiting on @StephenMcConnel)

___
*[`src/BloomExe/Collection/CollectionSettings.cs` line 337 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6604#-O4pcpL48jF9vInf1gk9:-O4pcpL5AM11ttvQvvKD:b-kwh2fx) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/24c8d4d0167dfb78ef997b314733c74abe4b214b/src/BloomExe/Collection/CollectionSettings.cs#L337)):*
> ```Smalltalk
>                 Language1.FontName,
>                 Language1.IsRightToLeft,
>                 0, // The page number has its own line-height value.  See BL-13699.
> ```

Sorry; I'm not understanding this fix.
Where is the line-height for the page number overriding this? If it is overriding it, why doesn't it override what we are setting here already?
Why is 6.1 working but not 6.0?
Is it ok to merge this into 6.1?
Maybe the answer to some of these want to be in an expanded comment?

If it's easier to discuss than type, I'm available.

StephenMcConnel (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6604)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @andrew-polk)

___
*[`src/BloomExe/Collection/CollectionSettings.cs` line 337 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6604#-O4pcpL48jF9vInf1gk9:-O4q_DPcDTlcm5zonRmH:bku88m1) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/24c8d4d0167dfb78ef997b314733c74abe4b214b/src/BloomExe/Collection/CollectionSettings.cs#L337)):*

Sorry; I'm not understanding this fix.
Where is the line-height for the page number overriding this? If it is overriding it, why doesn't it override what we are setting here already?
Why is 6.1 working but not 6.0?
Is it ok to merge this into 6.1?
Maybe the answer to some of these want to be in an expanded comment?

If it's easier to discuss than type, I'm available.
</blockquote></details>

1) This PR targets 6.0 since it fails the same way 6.1 does.
2) I should probably comment in the code why this is needed.  Basically, defaultLangStyles.css is loaded after basePage.css which sets the correct value for the page number's line-height based on an appearance system variable.  The important setting provided by defaultLangStyles.css is the actual font and possibly the directionality of the language script.

andrew-polk (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6604)* status: all files reviewed, 1 unresolved discussion (waiting on @StephenMcConnel)

___
*[`src/BloomExe/Collection/CollectionSettings.cs` line 337 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6604#-O4pcpL48jF9vInf1gk9:-O4qnH3B4Hb8ZACuxJ0l:b-f2vce) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/24c8d4d0167dfb78ef997b314733c74abe4b214b/src/BloomExe/Collection/CollectionSettings.cs#L337)):*

1) This PR targets 6.0 since it fails the same way 6.1 does.
2) I should probably comment in the code why this is needed.  Basically, defaultLangStyles.css is loaded after basePage.css which sets the correct value for the page number's line-height based on an appearance system variable.  The important setting provided by defaultLangStyles.css is the actual font and possibly the directionality of the language script.
</blockquote></details>

Sorry; I could have sworn you stated somewhere this was broken in 6.0 but working in 6.1. 
That misunderstanding caused other misunderstandings.

I think another source of confusion is that setting it to 0 here doesn't set the line height in the css to 0 as I was thinking. Rather that is a flag that it should not be set at all in the css.

Since you now have the fuller explanation in the larger comment up front, should we change the comment on the `0,` line to be just `// Prevents a line-height rule from being created`?
Or can we at least add that information in a comment somewhere?

andrew-polk:
*[`src/BloomExe/Collection/CollectionSettings.cs` line 337 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6604#-O4pcpL48jF9vInf1gk9:-O4qoYfSENi0x_KDBdl9:bdhzcxk) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/24c8d4d0167dfb78ef997b314733c74abe4b214b/src/BloomExe/Collection/CollectionSettings.cs#L337)):*

Sorry; I could have sworn you stated somewhere this was broken in 6.0 but working in 6.1. 
That misunderstanding caused other misunderstandings.

I think another source of confusion is that setting it to 0 here doesn't set the line height in the css to 0 as I was thinking. Rather that is a flag that it should not be set at all in the css.

Since you now have the fuller explanation in the larger comment up front, should we change the comment on the `0,` line to be just `// Prevents a line-height rule from being created`?
Or can we at least add that information in a comment somewhere?
</blockquote></details>

Or maybe even use -1 instead of 0 along with the comment?

StephenMcConnel (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6604)* status: all files reviewed, 1 unresolved discussion (waiting on @andrew-polk)

___
*[`src/BloomExe/Collection/CollectionSettings.cs` line 337 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6604#-O4pcpL48jF9vInf1gk9:-O4qv--IDF_6wYHLSXLX:b-7h84gi) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/24c8d4d0167dfb78ef997b314733c74abe4b214b/src/BloomExe/Collection/CollectionSettings.cs#L337)):*

Or maybe even use -1 instead of 0 along with the comment?
</blockquote></details>

Done.
I may have claimed it was working properly in 6.1 before testing it properly by closing the book, which actually triggers writing the defaultLangStyles.css file.  But I tried to delete all such comments in hopes that nobody had read them.  :-)

andrew-polk (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6604)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @StephenMcConnel)

PR #6609, opened 2024-08-22
Fix crash when downloading books with >1000 S3 objects (20240822)

StephenMcConnel:
Books with more than 1000 S3 objects would silently fail, crashing with an out of memory error due to an infinite (nonterminating) loop.  The symptom was a harvesterState of "Failed" and an empty harvesterLog array.

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6609)

### DIFF
--- a/src/BloomExe/WebLibraryIntegration/S3Extensions.cs
+++ b/src/BloomExe/WebLibraryIntegration/S3Extensions.cs
@@ -31,7 +31,10 @@ namespace Bloom.WebLibraryIntegration
             {
                 matchingItemsResponse = s3.ListObjectsV2(request);
                 allMatchingItems.AddRange(matchingItemsResponse.S3Objects);
-                request.ContinuationToken = matchingItemsResponse.ContinuationToken; // ContinuationToken indicates where the next request should start
+                // matchingItemsResponse.ContinuationToken indicates where the request that generated the response started
+                // matchingItemsResponse.NextContinuationToken indicates where the next request (if needed) should start
+                // request.ContinuationToken indicates where the request starts the next time it is used
+                request.ContinuationToken = matchingItemsResponse.NextContinuationToken;
             } while (matchingItemsResponse.IsTruncated); // IsTruncated returns true if it's not at the end
 
             return allMatchingItems;
--- a/src/BloomTests/WebLibraryIntegration/BloomS3ClientTests.cs
+++ b/src/BloomTests/WebLibraryIntegration/BloomS3ClientTests.cs
@@ -162,7 +162,7 @@ namespace BloomTests.WebLibraryIntegration
 
                 // Prep the next request (if needed)
                 listMatchingObjectsRequest.ContinuationToken =
-                    matchingFilesResponse.ContinuationToken;
+                    matchingFilesResponse.NextContinuationToken;
             } while (matchingFilesResponse.IsTruncated); // Returns true if haven't reached the end yet
         }
     }


andrew-polk (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6609)* status: all files reviewed, 1 unresolved discussion (waiting on @StephenMcConnel)

___
*[`src/BloomExe/WebLibraryIntegration/S3Extensions.cs` line 37 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6609#-O4vDVb-5lGzDg4Gz8s-:-O4vDVb-5lGzDg4Gz8s0:b-vj4hcx) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/70c3776a3d4fa70a72db2e9f89fdd67ea856a720/src/BloomExe/WebLibraryIntegration/S3Extensions.cs#L37)):*
> ```Smalltalk
>                 // matchingItemsResponse.NextContinuationToken indicates where the next request (if needed) should start
>                 // request.ContinuationToken indicates where the request starts the next time it is used
>                 request.ContinuationToken = matchingItemsResponse.NextContinuationToken;
> ```

Good find.
Looks like we have the same bug in DeleteFromUnitTestBucket, though that probably never manifests. Will you please fix that one while we're looking at this?

Do you think this fixes any bugs in actual user downloads?

StephenMcConnel (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6609)* status: all files reviewed, 1 unresolved discussion (waiting on @andrew-polk)

___
*[`src/BloomExe/WebLibraryIntegration/S3Extensions.cs` line 37 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6609#-O4vDVb-5lGzDg4Gz8s-:-O4vFO4Y8hOsrPX2Z8aC:beqbi3d) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/70c3776a3d4fa70a72db2e9f89fdd67ea856a720/src/BloomExe/WebLibraryIntegration/S3Extensions.cs#L37)):*

Good find.
Looks like we have the same bug in DeleteFromUnitTestBucket, though that probably never manifests. Will you please fix that one while we're looking at this?

Do you think this fixes any bugs in actual user downloads?
</blockquote></details>

Done.  I would hope that none of the unit tests have more than a handful of objects, but we should have correct code nonetheless.

At least one book has this problem.  (With 6209 objects, I half expect an time-out creating the artifacts.)  Possibly two more also.  (All three are marked "Failed" with an empty harvester log.)

StephenMcConnel (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6609)* status: 1 of 2 files reviewed, 1 unresolved discussion (waiting on @andrew-polk)

___
*[`src/BloomExe/WebLibraryIntegration/S3Extensions.cs` line 37 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6609#-O4vDVb-5lGzDg4Gz8s-:-O4vHmPA1tkIC1xThuRr:bd98xg3) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/70c3776a3d4fa70a72db2e9f89fdd67ea856a720/src/BloomExe/WebLibraryIntegration/S3Extensions.cs#L37)):*

Done.  I would hope that none of the unit tests have more than a handful of objects, but we should have correct code nonetheless.

At least one book has this problem.  (With 6209 objects, I half expect an time-out creating the artifacts.)  Possibly two more also.  (All three are marked "Failed" with an empty harvester log.)
</blockquote></details>

I stepped through the loop once I found the problem, replacing the request.ContinuationToken manually to verify the fix.

andrew-polk (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6609)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @StephenMcConnel)

PR #6612, opened 2024-08-23
Move the pages in the Activity book to the new Games book

JohnThomson:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6612)


JohnThomson:
Converting to draft for now because of https://issues.bloomlibrary.org/youtrack/issue/BL-13798/Quiz-page-crashes-trying-to-play-sound-in-BloomPub-Preview, which interferes with testing. I'd also like to find an test a suitable widget for a Captive Widget page.

JohnThomson:
The bulk of the new content is just pages that have moved from the output of Activity.pug to the new Games.html template book (which does not use pug). I did not edit them at all, so a detailed review is probably not needed.

andrew-polk (changes requested):
> Reviewed 27 of 27 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6612)* status: all files reviewed, 2 unresolved discussions (waiting on @JohnThomson)

___
*[`src/content/templates/template books/Activity/Activity.less` line 0 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6612#-O5Ed7eZBRWZISi0zPbW:-O5Ed7eZBRWZISi0zPbX:bfxtegq) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/8929394ec4e72b5f65b8c84821a567aad476e7d6/src/content/templates/template%20books/Activity/Activity.less#L0)):*
I have a vague concern about the backwards compatibility of removing this file which is looked for in existing books.
See, for example, Special.less which says it needs to be maintained as a copy of Activity.less because older books are looking for it.
___
*[`src/content/templates/template books/Games/Games.less` line 3 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6612#-O5EcwAV5awH_1lLbydq:-O5EcwAV5awH_1lLbydr:br5osmz) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/8929394ec4e72b5f65b8c84821a567aad476e7d6/src/content/templates/template%20books/Games/Games.less#L3)):*
> ```Less
> @import (reference) "../../../../BloomBrowserUI/bloomUI.less";
> @import "./simpleComprehensionQuiz.less";
> @import "./simple-dom-choice-activities.less";
> ```

I don't suppose there is any simple way to keep the less files separate for the different types of activities/games?
It is a shame that every book which uses one type has to carry the css for all the other types.

JohnThomson (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6612)* status: all files reviewed, 2 unresolved discussions (waiting on @andrew-polk)

___
*[`src/content/templates/template books/Activity/Activity.less` line  at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6612#-O5Ed7eZBRWZISi0zPbW:-O5EzK0B2EWIbrjCyZ5c:bngapnl) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/8929394ec4e72b5f65b8c84821a567aad476e7d6/src/content/templates/template%20books/Activity/Activity.less#L)):*

I have a vague concern about the backwards compatibility of removing this file which is looked for in existing books.
See, for example, Special.less which says it needs to be maintained as a copy of Activity.less because older books are looking for it.
</blockquote></details>

I can't prove that's not the case. But surely such books would have their own copy? So at worst we're losing the ability to update it with a more recent version. If we find we need to do that, we could presumably reintroduce the file then?
___
*[`src/content/templates/template books/Games/Games.less` line 3 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6612#-O5EcwAV5awH_1lLbydq:-O5EyJ9LEk9SCULYdKvh:bfil1od) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/8929394ec4e72b5f65b8c84821a567aad476e7d6/src/content/templates/template%20books/Games/Games.less#L3)):*

I don't suppose there is any simple way to keep the less files separate for the different types of activities/games?
It is a shame that every book which uses one type has to carry the css for all the other types.
</blockquote></details>

There may be a friendlier attribute to determine it by than the page ID guid. But I tend to think complicating the "which stylesheets do I need to copy" logic is less desirable than carrying around a bit of CSS we don't need.

andrew-polk (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6612)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @JohnThomson)

___
*[`src/content/templates/template books/Activity/Activity.less` line  at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6612#-O5Ed7eZBRWZISi0zPbW:-O5F0fF_5Be--B2_jN5G:b-df05is) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/8929394ec4e72b5f65b8c84821a567aad476e7d6/src/content/templates/template%20books/Activity/Activity.less#L)):*

I can't prove that's not the case. But surely such books would have their own copy? So at worst we're losing the ability to update it with a more recent version. If we find we need to do that, we could presumably reintroduce the file then?
</blockquote></details>

Ok, I'll add a test checkbox in the issue to ensure that there are no obvious problems when bringing a book up to date which has a quiz.

PR #6621, opened 2024-08-28
Fix spacing weirdness on quiz pages when not focused (BL-13807)

StephenMcConnel:
This is really a one-line change.  I don't know why the reformatting happened.   (commit hook?)
The last changed line is the only actual change.

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6621)

### DIFF
--- a/src/content/templates/template books/Activity/simpleComprehensionQuiz.less
+++ b/src/content/templates/template books/Activity/simpleComprehensionQuiz.less
@@ -164,40 +164,24 @@
         input:active ~ .placeToPutVariableCircle {
             opacity: 0.5;
             transform: scale(1);
-            transition:
-                opacity 0.3s,
-                transform 0.2s; //opacity 0.3s, transform 0.2s;
+            transition: opacity 0.3s, transform 0.2s; //opacity 0.3s, transform 0.2s;
         }
     }
 }
 
 // overrides a rule in EditMode.less to push the language label over and make room for cog on right.
-.bloom-editable.QuizHeader-style[contentEditable="true"][data-languageTipContent]:not(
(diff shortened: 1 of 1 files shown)


hatton (commented):
Steve can you tell me what caused the problem and how this fixes it?

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6621#-:-O5Q7l3i5GJOv7uDQTgM:b-ew9s0t)* status: 0 of 1 files reviewed, all discussions resolved

StephenMcConnel (commented):
The content: "" still displays something with a minimum height even though it effectively has zero width.   Being in an :after rule when focus was off, it got added after the real data, causing the height of the containing element to grow when focus was removed.  The none keyword causes the CSS to know that absolutely nothing should be displayed, and thus zero height as well as zero width gets added in the :after rule.  So the :after selector no longer affects the height of the element.

But I'll admit it wasn't my cleverness or knowledge that led me to the fix.  I providentially stumbled over it while tearing my hair out.  :-)

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6621#-:-O5Q8StW34Z3zKPV0dx_:b-93z1q8)* status: 0 of 1 files reviewed, all discussions resolved

hatton (commented):
Thanks Steve

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6621#-:-O5QCJ2b136OOwRJ7BTv:b7c7o01)* status: 0 of 1 files reviewed, all discussions resolved

PR #6627, opened 2024-08-29
Store all language names from spreadsheet import (BL-13551)

StephenMcConnel:
This change does not address any GUI changes for setting or changing this information within Bloom itself.  More information than the language tag and name can be stored but currently would not be used anywhere.

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6627)

### DIFF
--- a/src/BloomExe/Collection/CollectionSettings.cs
+++ b/src/BloomExe/Collection/CollectionSettings.cs
@@ -38,10 +38,10 @@ namespace Bloom.Collection
     {
         private const int kCurrentOneTimeCheckVersionNumber = 1; // bumping this will trigger a new one time check
         public const string kDefaultXmatterName = "Traditional";
-        public List<WritingSystem> LanguagesZeroBased = new List<WritingSystem>();
-        public WritingSystem Language1 => LanguagesZeroBased[0];
-        public WritingSystem Language2 => LanguagesZeroBased[1];
-        public WritingSystem Language3 => LanguagesZeroBased[2];
+        public List<WritingSystem> AllLanguages = new List<WritingSystem>();
+        public WritingSystem Language1 => AllLanguages[0];
+        public WritingSystem Language2 => AllLanguages[1];
+        public WritingSystem Language3 => AllLanguages[2];
 
         // Email addresses of users authorized to change collection settings if this is a TeamCollection.
         public string[] Administrators;
@@ -113,11 +113,11 @@ namespace Bloom.Collection
             //Note: I'm not convinced we actually ever rely on dynamic name lookups anymore?
             //See: https://issues.bloomlibrary.org/youtrack/issue/BL-7832
             Func<string> getTagOfDefaultLanguageForNaming = () => Language2.Tag;
-            LanguagesZeroBased.Add(new WritingSystem(1, getTagOfDefaultLanguageForNaming));
-            LanguagesZeroBased.Add(new WritingSystem(2, getTagOfDefaultLanguageForNaming));
-            LanguagesZeroBased.Add(new WritingSystem(3, getTagOfDefaultLanguageForNaming));
+            AllLanguages.Add(new WritingSystem(getTagOfDefaultLanguageForNaming));
+            AllLanguages.Add(new WritingSystem(getTagOfDefaultLanguageForNaming));
+            AllLanguages.Add(new WritingSystem(getTagOfDefaultLanguageForNaming));
 
-            SignLanguage = new WritingSystem(-1, getTagOfDefaultLanguageForNaming);
+            SignLanguage = new WritingSystem(getTagOfDefaultLanguageForNaming);
 
             BrandingProjectKey = "Default";
             PageNumberStyle = "Decimal";
@@ -141,7 +141,7 @@ namespace Bloom.Collection
         {
             AllowNewBooks = collectionInfo.AllowNewBooks;
 
-            LanguagesZeroBased = new List<WritingSystem>(collectionInfo.LanguagesZeroBased);
+            AllLanguages = new List<WritingSystem>(collectionInfo.AllLanguages);
             Language2.FontName = Language3.FontName = WritingSystem.GetDefaultFontName();
             SignLanguage = collectionInfo.SignLanguage;
 
@@ -255,17 +255,11 @@ namespace Bloom.Collection
         {
             // Use the WritingSystem name based on Ethnologue (or customized) in preference to the
             // IeftLanguageTag name based on older ISO 639 data.  See BL-12992.
-            if (tag == Language1Tag)
-                return Language1.Name;
-            if (tag == Language2Tag)
-                return Language2.Name;
-            if (tag == Language3Tag)
-                return Language3.Name;
-            if (tag == SignLanguageTag)
-                return SignLanguage.Name;
-            var language = LanguagesZeroBased.Find(x => x.Tag == tag);
+            var language = AllLanguages.Find(x => x.Tag == tag);
             if (language != null)
                 return language.Name;
+            if (tag == SignLanguageTag)
+                return SignLanguage.Name;
             // Note: the inLanguage parameter is often ignored by IetfLanguageTag.GetLocalizedLanguageName().
             return IetfLanguageTag.GetLocalizedLanguageName(tag, inLanguage);
         }
@@ -279,12 +273,12 @@ namespace Bloom.Collection
             XElement xml = new XElement("Collection");
             xml.Add(new XAttribute("version", "0.2"));
             xml.Add(new XElement("CollectionId", CollectionId));
-            Language1.SaveToXElement(xml);
-            Language2.SaveToXElement(xml);
-            Language3.SaveToXElement(xml);
-            var languageList = new XElement("Languages");
+            Language1.SaveToXElementLegacy(xml, 1);
+            Language2.SaveToXElementLegacy(xml, 2);
+            Language3.SaveToXElementLegacy(xml, 3);
+            var languagesElement = new XElement("Languages");
             int wsNum = 0;
-            foreach (var langWs in LanguagesZeroBased)
+            foreach (var langWs in AllLanguages)
             {
                 var language = new XElement("Language");
                 switch (++wsNum)
@@ -303,11 +297,11 @@ namespace Bloom.Collection
                         language.Add(new XAttribute("L3", "true"));
                         break;
                 }
-                langWs.SaveToXElement(language, 0);
-                languageList.Add(language);
+                langWs.SaveToXElement(language);
+                languagesElement.Add(language);
             }
-            xml.Add(languageList);
-            SignLanguage.SaveToXElement(xml);
+            xml.Add(languagesElement);
+            SignLanguage.SaveToXElement(xml, isSignLanguage: true);
             xml.Add(new XElement("OneTimeCheckVersionNumber", OneTimeCheckVersionNumber));
             xml.Add(new XElement("IsSourceCollection", IsSourceCollection.ToString()));
             xml.Add(new XElement("XMatterPack", XMatterPackName));
@@ -479,11 +473,11 @@ namespace Bloom.Collection
                 }
                 else
                 {
-                    Language1.ReadFromXml(xml, true, "en");
-                    Language2.ReadFromXml(xml, true, "self");
-                    Language3.ReadFromXml(xml, true, Language2.Tag);
+                    Language1.ReadFromXmlLegacy(xml, "en", 1);
+                    Language2.ReadFromXmlLegacy(xml, "self", 2);
+                    Language3.ReadFromXmlLegacy(xml, Language2.Tag, 3);
                 }
-                SignLanguage.ReadFromXml(xml, true, Language2.Tag);
+                SignLanguage.ReadFromXml(xml, Language2.Tag, isSignLanguage: true);
 
                 XMatterPackName = ReadString(xml, "XMatterPack", "Factory");
 
@@ -654,16 +648,16 @@ namespace Bloom.Collection
             foreach (var language in languageList.Descendants("Language"))
             {
                 if (wsNum == 0)
-                    Language1.ReadFromXml(language, true, "en", 0);
+                    Language1.ReadFromXml(language, "en");
                 else if (wsNum == 1)
-                    Language2.ReadFromXml(language, true, "self", 0);
+                    Language2.ReadFromXml(language, "self");
                 else if (wsNum == 2 && language.Attribute("L3")?.Value == "true")
-                    Language3.ReadFromXml(language, true, Language2.Tag, 0);
+                    Language3.ReadFromXml(language, Language2.Tag);
                 else
                 {
-                    var writingSystem = new WritingSystem(0, DefaultLanguageForNamingLanguages);
-                    writingSystem.ReadFromXml(language, true, Language2.Tag, 0);
-                    LanguagesZeroBased.Add(writingSystem);
+                    var writingSystem = new WritingSystem(DefaultLanguageForNamingLanguages);
+                    writingSystem.ReadFromXml(language, Language2.Tag);
+                    AllLanguages.Add(writingSystem);
                 }
                 ++wsNum;
             }
@@ -676,21 +670,34 @@ namespace Bloom.Collection
             }
         }
 
-        public void UpdateOrCreateCollectionLanguage(string tag, string name)
+        /// <summary>
+        /// Update the collection settings if necessary with the language tag and name.
+        /// But if okToMakeChange is false, don't make the change, just return whether it would be necessary.
+        /// </summary>
+        /// <returns>true if the settings change, false if the tag and name are already set that way</returns>
+        public bool UpdateOrCreateCollectionLanguage(
+            string tag,
+            string name,
+            bool okToMakeChange = true
+        )
         {
-            var language = LanguagesZeroBased.Find(x => x.Tag == tag);
+            var language = AllLanguages.Find(x => x.Tag == tag);
             if (language != null)
             {
-                language.SetName(name, true);
+                if (language.Name == name)
+                    return false;
+                if (okToMakeChange)
+                    language.SetName(name, true);
             }
-            else
+            else if (okToMakeChange)
             {
-                var writingSystem = new WritingSystem(0, DefaultLanguageForNamingLanguages);
+                var writingSystem = new WritingSystem(DefaultLanguageForNamingLanguages);
                 writingSystem.Tag = tag;
                 writingSystem.SetName(name, true);
                 writingSystem.FontName = WritingSystem.GetDefaultFontName();
-                LanguagesZeroBased.Add(writingSystem);
+                AllLanguages.Add(writingSystem);
             }
+            return true;
         }
 
         public bool LockedToOneDownloadedBook;
--- a/src/BloomExe/Collection/CollectionSettings.cs
+++ b/src/BloomExe/Collection/CollectionSettings.cs
@@ -22,7 +22,6 @@ using SIL.Extensions;
 using SIL.IO;
 using SIL.Reporting;
 using SIL.WritingSystems;
-using Directory = System.IO.Directory;
 
 namespace Bloom.Collection
 {
@@ -39,10 +38,10 @@ namespace Bloom.Collection
     {
         private const int kCurrentOneTimeCheckVersionNumber = 1; // bumping this will trigger a new one time check
         public const string kDefaultXmatterName = "Traditional";
-        public WritingSystem Language1;
-        public WritingSystem Language2;
-        public WritingSystem Language3;
-        public WritingSystem[] LanguagesZeroBased;
+        public List<WritingSystem> LanguagesZeroBased = new List<WritingSystem>();
+        public WritingSystem Language1 => LanguagesZeroBased[0];
+        public WritingSystem Language2 => LanguagesZeroBased[1];
+        public WritingSystem Language3 => LanguagesZeroBased[2];
 
         // Email addresses of users authorized to change collection settings if this is a TeamCollection.
         public string[] Administrators;
@@ -114,13 +113,9 @@ namespace Bloom.Collection
             //Note: I'm not convinced we actually ever rely on dynamic name lookups anymore?
             //See: https://issues.bloomlibrary.org/youtrack/issue/BL-7832
             Func<string> getTagOfDefaultLanguageForNaming = () => Language2.Tag;
-            Language1 = new WritingSystem(1, getTagOfDefaultLanguageForNaming);
-            Language2 = new WritingSystem(2, getTagOfDefaultLanguageForNaming);
-            Language3 = new WritingSystem(3, getTagOfDefaultLanguageForNaming);
-            LanguagesZeroBased = new WritingSystem[3];
-            this.LanguagesZeroBased[0] = Language1;
-            this.LanguagesZeroBased[1] = Language2;
-            this.LanguagesZeroBased[2] = Language3;
+            LanguagesZeroBased.Add(new WritingSystem(1, getTagOfDefaultLanguageForNaming));
+            LanguagesZeroBased.Add(new WritingSystem(2, getTagOfDefaultLanguageForNaming));
+            LanguagesZeroBased.Add(new WritingSystem(3, getTagOfDefaultLanguageForNaming));
 
             SignLanguage = new WritingSystem(-1, getTagOfDefaultLanguageForNaming);
 
@@ -145,11 +140,8 @@ namespace Bloom.Collection
             : this(collectionInfo.PathToSettingsFile)
         {
             AllowNewBooks = collectionInfo.AllowNewBooks;
-            Language1.FontName = collectionInfo.Language1.FontName;
-            Language1 = collectionInfo.Language1;
-            Language2 = collectionInfo.Language2;
-            Language3 = collectionInfo.Language3;
 
+            LanguagesZeroBased = new List<WritingSystem>(collectionInfo.LanguagesZeroBased);
             Language2.FontName = Language3.FontName = WritingSystem.GetDefaultFontName();
             SignLanguage = collectionInfo.SignLanguage;
 
@@ -271,6 +263,9 @@ namespace Bloom.Collection
                 return Language3.Name;
             if (tag == SignLanguageTag)
                 return SignLanguage.Name;
+            var language = LanguagesZeroBased.Find(x => x.Tag == tag);
+            if (language != null)
+                return language.Name;
             // Note: the inLanguage parameter is often ignored by IetfLanguageTag.GetLocalizedLanguageName().
             return IetfLanguageTag.GetLocalizedLanguageName(tag, inLanguage);
         }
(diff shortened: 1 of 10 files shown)


JohnThomson (changes requested):
I'm still puzzled about a few things and have some suggested simplifications I want JohnH's input on. But I think I'll give you what I have.

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6627#-:-O63UoEv185X7Jzoiy0C:b955nf5)* status: all files reviewed, 14 unresolved discussions (waiting on @StephenMcConnel)

___
*[`src/BloomExe/Collection/CollectionSettings.cs` line 144 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6627#-O62swt08k9XpvnJ5rFj:-O62swt08k9XpvnJ5rFk:b-2wkr2d) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/cd19097ac9850e94593796ca4bb7dda6ebb36aec/src/BloomExe/Collection/CollectionSettings.cs#L144)):*
> ```Smalltalk
>             AllowNewBooks = collectionInfo.AllowNewBooks;
> 
>             LanguagesZeroBased = new List<WritingSystem>(collectionInfo.LanguagesZeroBased);
> ```

All C# lists are zero-based. Is there some particular reason to call attention to it in this name? I suppose maybe because it's awkward that Languages[0] is what we've always called Language1. Of course we apparently always called it that...
___
*[`src/BloomExe/Collection/CollectionSettings.cs` line 265 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6627#-O63042A1KPdVLZO26um:-O63042A1KPdVLZO26un:b-a3049m) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/cd19097ac9850e94593796ca4bb7dda6ebb36aec/src/BloomExe/Collection/CollectionSettings.cs#L265)):*
> ```Smalltalk
>                 return Language3.Name;
>             if (tag == SignLanguageTag)
>                 return SignLanguage.Name;
> ```

We could omit the previous six lines. Is it significantly more efficient to keep them?
___
*[`src/BloomExe/Collection/CollectionSettings.cs` line 308 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6627#-O62uiGA2F8pOOYAH6aM:-O62tfU5DIr49EGkObNa:b-owr1dc) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/cd19097ac9850e94593796ca4bb7dda6ebb36aec/src/BloomExe/Collection/CollectionSettings.cs#L308)):*
> ```Smalltalk
>                 langWs.SaveToXElement(language, 0);
>                 languageList.Add(language);
>             }
> ```

If we know there are always three, could we just do three lines here like
languageList[0].Add(new XAttribute("L1", "true"))
rather than the switch above? (But mabe SaveToXElement would get messed up...)
___
*[`src/BloomExe/Collection/CollectionSettings.cs` line 484 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6627#-O63M-P00nnoOIUKpANP:-O63M-P00nnoOIUKpANQ:b-auj7va) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/cd19097ac9850e94593796ca4bb7dda6ebb36aec/src/BloomExe/Collection/CollectionSettings.cs#L484)):*
> ```Smalltalk
>                     Language1.ReadFromXml(xml, true, "en");
>                     Language2.ReadFromXml(xml, true, "self");
>                     Language3.ReadFromXml(xml, true, Language2.Tag);
> ```

I suppose it's rare that we don't get a language code in the file...the only time a writing system doesn't have one is when the collection really has no L3? And then we don't write it to the new list at all. And I suppose the old Language3Tag attribute is empty? But if there's ever a case to not default to en, surely this would be it? Conversely, if this argument is always passed true, as I think it is, let's simplify the code and get rid of it.
Actually, I'm not clear how the old code worked if there is no L3. If it wrote out an empty tag, it looks like this code would cause it to read back as "en"!
___
*[`src/BloomExe/Collection/CollectionSettings.cs` line 661 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6627#-O63No3_4c09uCvsa-ql:-O63No3a7oOdX5BfVRJc:b-jqldcy) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/cd19097ac9850e94593796ca4bb7dda6ebb36aec/src/BloomExe/Collection/CollectionSettings.cs#L661)):*
> ```Smalltalk
>                     Language2.ReadFromXml(language, true, "self", 0);
>                 else if (wsNum == 2 && language.Attribute("L3")?.Value == "true")
>                     Language3.ReadFromXml(language, true, Language2.Tag, 0);
> ```

I've started a discussion in Flowdock around the idea that we could either just use en when we don't find a language name in the file (why would that ever happen??), or read the L2 tag out of the XML before the start of the loop and always use that. Then L1/2/3 are not special...we just have to remove the old ones from the list before we start.
Except...L3 is special because we don't write it at all if there isn't one. Yet there might be a third language in the list, if there is no current L3 but we have records for more than two languages.
What about if we just read all the languages we find into the list, then, if languageList[2].Attribute("L3")?.Value is not true, we insert the necessary dummy L3 into the list?
___
*[`src/BloomExe/Collection/CollectionSettingsDialog.cs` line 273 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6627#-O62xFmJ6H8ao6NhRzTE:-O62xFmJ6H8ao6NhRzTF:b-a83j1z) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/cd19097ac9850e94593796ca4bb7dda6ebb36aec/src/BloomExe/Collection/CollectionSettingsDialog.cs#L273)):*
> ```Smalltalk
>                 // Review/Todo: update the rest of PendingLanguage1's properties if a match found in
>                 // existing languages? _collectionSettings.LanguagesZeroBased.Find(x => x.Tag == l.languageTag);
>                 // (and possibly update PendingFontSelections[0] as well)
> ```

This is where the user chose a new L1, right? So we could now initialize the relevant fields in the dialog using info about that language (if any) from LanguagesZerobased? But I gather this is not actually a review or Todo, because we decided NOT to do it for 6.1? So I think it should be an Enhance.
I find "update" a bit ambiguous here, between updating the dialog from the data or the opposite. Maybe "Enhance: if we already have information about the newly chosen language we could copy it into the dialog"
___
*[`src/BloomExe/Collection/CollectionSettingsDialog.cs` line 413 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6627#-O62wQkS91AkQqYDe6XH:-O62wQkS91AkQqYDe6XI:b-kbpt7) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/2afdd326487b9dd5aaa90c118376de5c5821cf13/src/BloomExe/Collection/CollectionSettingsDialog.cs#L413)):*
> ```Smalltalk
>             _collectionSettings.District = _districtText.Text.Trim();
> 
>             var languages = _collectionSettings.LanguagesZeroBased;
> ```

I'm confused here. LanguagesZeroBased appears to be a new variable, yet here we are deleting code that references it.
___
*[`src/BloomExe/Collection/CollectionSettingsDialog.cs` line 562 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6627#-O63-IYNDsQzNfLQy___:-O63-IYNDsQzNfLQy__a:b-3uggbi) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/cd19097ac9850e94593796ca4bb7dda6ebb36aec/src/BloomExe/Collection/CollectionSettingsDialog.cs#L562)):*
> ```Smalltalk
>                 }
>                 else if (
>                     !String.IsNullOrEmpty(PendingLanguage3.Tag)
> ```

Don't think we need this check? Null or empty won't be equal to the tag of anything in the list.
___
*[`src/BloomExe/Collection/WritingSystem.cs` line 135 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6627#-O635J6l9mN2LqrZEuCK:-O635J6l9mN2LqrZEuCL:b-447c9p) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/cd19097ac9850e94593796ca4bb7dda6ebb36aec/src/BloomExe/Collection/WritingSystem.cs#L135)):*
> ```Smalltalk
>                 langNumber = _languageNumberInCollection;
>             var langNumTag = (langNumber > 0) ? langNumber.ToString() : "";
>             var pfx = langNumber < 0 ? "SignLanguage" : "Language" + langNumTag;
> ```

I found this very confusing. If we must use magic numbers, how about a switch on langNumber (without modifying it):
-2: pfx = "Language" + _languageNumberInCollection;
-1: pfx = "SignLanguage";
0: pfx = "";
default: pfx = "Language" + langNumber.ToString();

But also consider whether this could become SaveToXElementInternal and take pfx (and maybe langNumTag?) as an argument.
Then have distinct methods:
SaveToXElement...used for the new approach, calls internal with empty pfx
SaveToXElementLegacy or SaveToXElementForOldBlooms...passes pfx as "Language" + _languageNumberInCollection
and also handles the backwards compatibility notice.
SaveSignLanguageToXElement...passes pfx as "SignLanguage"

And then...there's only one usage of _languageNumberInCollection except for SaveToXElement and ReadFromXml, and the callers of those methods that want legacy behavior all know which of L1/L2/L3/Sign they are trying to read or write. If we made the number an argument to SaveToXElementLegacy, we could get rid of _languageNumberInCollection, except for a very rare usage in coming up with a name for a language that isn't in our database.
___
*[`src/BloomExe/Collection/WritingSystem.cs` line 145 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6627#-O639nJCArPpJ0db3-Xe:-O639nJCArPpJ0db3-Xf:bogam8d) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/cd19097ac9850e94593796ca4bb7dda6ebb36aec/src/BloomExe/Collection/WritingSystem.cs#L145)):*
> ```Smalltalk
>             xml.Add(new XElement(pfx + "IsCustomName", IsCustomName));
>             xml.Add(new XElement(pfx + "Iso639Code", Tag));
>             if (langNumber < 0)
> ```

by this point it will only be negative for sign language, right? If we do the "internal" method approach maybe you could pass an isSignLanguage argument.
___
*[`src/BloomExe/Collection/WritingSystem.cs` line 147 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6627#-O6398hW9nEsS0XUJk0x:-O6398hW9nEsS0XUJk0y:b-bz11q) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/cd19097ac9850e94593796ca4bb7dda6ebb36aec/src/BloomExe/Collection/WritingSystem.cs#L147)):*
> ```Smalltalk
>             if (langNumber < 0)
>                 return;
>             xml.Add(new XElement($"DefaultLanguage{langNumTag}FontName", FontName));
> ```

could this just be $"Default{pfx}FontName"? and similarly in the line below? Not a prefix here, but the point is that "LanguageN" is or is not part of the tag.
___
*[`src/BloomExe/Collection/WritingSystem.cs` line 212 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6627#-O63G0Xq9h2C3COloUH0:-O63G0Xq9h2C3COloUH1:b6cgza0) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/cd19097ac9850e94593796ca4bb7dda6ebb36aec/src/BloomExe/Collection/WritingSystem.cs#L212)):*
> ```Smalltalk
>             bool defaultToEnglishIfMissing,
>             string languageForDefaultNameLookup,
>             int langNumber = -2
> ```

Similar suggestions to SaveToXElement.
___
*[`src/BloomExe/Spreadsheet/SpreadsheetImporter.cs` line 447 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6627#-O63GhGh2bOR0wKsnvFF:-O63GhGh2bOR0wKsnvFG:b-cgahmy) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/cd19097ac9850e94593796ca4bb7dda6ebb36aec/src/BloomExe/Spreadsheet/SpreadsheetImporter.cs#L447)):*
> ```Smalltalk
>                     continue;
>                 // Set the name in the collection settings language list, creating the language if needed.
>                 _collectionSettings.UpdateOrCreateCollectionLanguage(lang, langName);
> ```

This could return a boolean that is true if either the language was added or the name was changed. Then we could |= that with _collectionSettingsAreDirty, and (I think) not need special cases for L1/2/3 (though probably still for sign, since it's not in the list). Plus we don't need to save collection settings if nothing changed, which is really desirable because in a TC you have to be an admin to change Collection settings.
___
*[`src/BloomExe/Spreadsheet/SpreadsheetImporter.cs` line 451 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6627#-O63JwpLAyuH3dPLK6jz:-O63JwpLAyuH3dPLK6k-:b-fqed34) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/cd19097ac9850e94593796ca4bb7dda6ebb36aec/src/BloomExe/Spreadsheet/SpreadsheetImporter.cs#L451)):*
> ```Smalltalk
>             }
>             if (_collectionSettingsAreDirty)
>                 _collectionSettings.Save();
> ```

You'll have to do something to prevent this if we're in a TC and not an admin. There may also be some protocol that CollectionSettings changes in a TC should be immediately auto-committed? Or something to do to make sure the next commit syncs them? I don't remember, but the collection settings dialog should be handling it.

StephenMcConnel (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6627)* status: 0 of 9 files reviewed, 14 unresolved discussions (waiting on @JohnThomson)

___
*[`src/BloomExe/Collection/CollectionSettings.cs` line 144 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6627#-O62swt08k9XpvnJ5rFj:-O6NwaAq-gk3iZNpPT7Q:b-vhh9c6) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/cd19097ac9850e94593796ca4bb7dda6ebb36aec/src/BloomExe/Collection/CollectionSettings.cs#L144)):*

All C# lists are zero-based. Is there some particular reason to call attention to it in this name? I suppose maybe because it's awkward that Languages[0] is what we've always called Language1. Of course we apparently always called it that...
</blockquote></details>

It's always been called that, apparently to remind programmers that L1 maps to Languages[0].  I didn't want to change the name in addition to all the other changes I was making.  But I've now changed the name to AllLanguages.
___
*[`src/BloomExe/Collection/CollectionSettings.cs` line 265 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6627#-O63042A1KPdVLZO26um:-O6NxzcK0Tr1VQK0sFEd:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/cd19097ac9850e94593796ca4bb7dda6ebb36aec/src/BloomExe/Collection/CollectionSettings.cs#L265)):*

We could omit the previous six lines. Is it significantly more efficient to keep them?
</blockquote></details>

Done.
___
*[`src/BloomExe/Collection/CollectionSettings.cs` line 308 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6627#-O62uiGA2F8pOOYAH6aM:-O6NydUe5yb2PiUr77yj:b-aoygdc) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/cd19097ac9850e94593796ca4bb7dda6ebb36aec/src/BloomExe/Collection/CollectionSettings.cs#L308)):*

If we know there are always three, could we just do three lines here like
languageList[0].Add(new XAttribute("L1", "true"))
rather than the switch above? (But mabe SaveToXElement would get messed up...)
</blockquote></details>

The whole problem is that we don't know there are always three because L3 can be unassigned, which is detected by the L3Tag  being empty.  I still think the switch is the most elegant way to handle the original L1, L2, and (possibly empty) L3 being the first three elements of the complete list.
I've renamed languageList to languagesElement to make it clearer that it's an XElement object, not a list or array object.
___
*[`src/BloomExe/Collection/CollectionSettings.cs` line 484 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6627#-O63M-P00nnoOIUKpANP:-O6O1Q7W5okaqSz1wQDr:bp0xb12) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/cd19097ac9850e94593796ca4bb7dda6ebb36aec/src/BloomExe/Collection/CollectionSettings.cs#L484)):*

I suppose it's rare that we don't get a language code in the file...the only time a writing system doesn't have one is when the collection really has no L3? And then we don't write it to the new list at all. And I suppose the old Language3Tag attribute is empty? But if there's ever a case to not default to en, surely this would be it? Conversely, if this argument is always passed true, as I think it is, let's simplify the code and get rid of it.
Actually, I'm not clear how the old code worked if there is no L3. If it wrote out an empty tag, it looks like this code would cause it to read back as "en"!
</blockquote></details>

Writing out an empty tag causes an empty tag to be read.  It's only if you don't write out the tag at all that you get "en" back as the default value.
But removing the redundantly useless parameter makes sense.  Done.
___
*[`src/BloomExe/Collection/CollectionSettings.cs` line 661 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6627#-O63No3_4c09uCvsa-ql:-O6RK6952vzOUKM0HDPY:bhjk7p8) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/cd19097ac9850e94593796ca4bb7dda6ebb36aec/src/BloomExe/Collection/CollectionSettings.cs#L661)):*

I've started a discussion in Flowdock around the idea that we could either just use en when we don't find a language name in the file (why would that ever happen??), or read the L2 tag out of the XML before the start of the loop and always use that. Then L1/2/3 are not special...we just have to remove the old ones from the list before we start.
Except...L3 is special because we don't write it at all if there isn't one. Yet there might be a third language in the list, if there is no current L3 but we have records for more than two languages.
What about if we just read all the languages we find into the list, then, if languageList[2].Attribute("L3")?.Value is not true, we insert the necessary dummy L3 into the list?
</blockquote></details>

In order to preserve writing out the original language elements, the first three elements are pre-created with with the readonly index attributes being assigned properly.  This is done in one place.  I'm not convinced that juggling the creation/insertion of the dummy (empty) Language3 is clearer than the existing code.  We need to check for the existence of the L3 attribute as the elements are read in any case, and the current code makes it clear that it belongs only on the third element, but optionally.
I don't understand your comments about L1/2/3 not being special if we're trying to maintain at least a partial backwards compatibility.  I though we had made the decision to do so.
___
*[`src/BloomExe/Collection/CollectionSettingsDialog.cs` line 273 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6627#-O62xFmJ6H8ao6NhRzTE:-O6RY8kP0HHQGrwWuo8U:b-ipatx3) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/cd19097ac9850e94593796ca4bb7dda6ebb36aec/src/BloomExe/Collection/CollectionSettingsDialog.cs#L273)):*

This is where the user chose a new L1, right? So we could now initialize the relevant fields in the dialog using info about that language (if any) from LanguagesZerobased? But I gather this is not actually a review or Todo, because we decided NOT to do it for 6.1? So I think it should be an Enhance.
I find "update" a bit ambiguous here, between updating the dialog from the data or the opposite. Maybe "Enhance: if we already have information about the newly chosen language we could copy it into the dialog"
</blockquote></details>

Done.  Changed the comment in all three places.
___
*[`src/BloomExe/Collection/CollectionSettingsDialog.cs` line 413 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6627#-O62wQkS91AkQqYDe6XH:-O6RYPTBBfTJKWt43w1_:b-51vzdl) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/2afdd326487b9dd5aaa90c118376de5c5821cf13/src/BloomExe/Collection/CollectionSettingsDialog.cs#L413)):*

I'm confused here. LanguagesZeroBased appears to be a new variable, yet here we are deleting code that references it.
</blockquote></details>

I'm also confused because I don't see what you apparently were seeing.
___
*[`src/BloomExe/Collection/CollectionSettingsDialog.cs` line 562 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6627#-O63-IYNDsQzNfLQy___:-O6RY_8B2hkCSKfaoK-D:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/cd19097ac9850e94593796ca4bb7dda6ebb36aec/src/BloomExe/Collection/CollectionSettingsDialog.cs#L562)):*

Don't think we need this check? Null or empty won't be equal to the tag of anything in the list.
</blockquote></details>

Done.
___
*[`src/BloomExe/Collection/WritingSystem.cs` line 135 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6627#-O635J6l9mN2LqrZEuCK:-O6RlUKQ0tzAzym7qylx:b-8ygw1c) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/cd19097ac9850e94593796ca4bb7dda6ebb36aec/src/BloomExe/Collection/WritingSystem.cs#L135)):*

I found this very confusing. If we must use magic numbers, how about a switch on langNumber (without modifying it):
-2: pfx = "Language" + _languageNumberInCollection;
-1: pfx = "SignLanguage";
0: pfx = "";
default: pfx = "Language" + langNumber.ToString();

But also consider whether this could become SaveToXElementInternal and take pfx (and maybe langNumTag?) as an argument.
Then have distinct methods:
SaveToXElement...used for the new approach, calls internal with empty pfx
SaveToXElementLegacy or SaveToXElementForOldBlooms...passes pfx as "Language" + _languageNumberInCollection
and also handles the backwards compatibility notice.
SaveSignLanguageToXElement...passes pfx as "SignLanguage"

And then...there's only one usage of _languageNumberInCollection except for SaveToXElement and ReadFromXml, and the callers of those methods that want legacy behavior all know which of L1/L2/L3/Sign they are trying to read or write. If we made the number an argument to SaveToXElementLegacy, we could get rid of _languageNumberInCollection, except for a very rare usage in coming up with a name for a language that isn't in our database.
</blockquote></details>

I've split into three methods: SaveToXElement, SaveToXElementLegacy, and SaveToXElementInternal.  The first and last take a boolean isSignLanguage parameter.  The second takes an integer languageNumber parameter.

I removed the _languageNumberInCollection member variable adjusting the constructors not to need that argument and adjusting the  name for totally unknown languages to be just "Unknown-" + tag.
___
*[`src/BloomExe/Collection/WritingSystem.cs` line 145 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6627#-O639nJCArPpJ0db3-Xe:-O6Rhs9j1iCuflLtDPuE:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/cd19097ac9850e94593796ca4bb7dda6ebb36aec/src/BloomExe/Collection/WritingSystem.cs#L145)):*

by this point it will only be negative for sign language, right? If we do the "internal" method approach maybe you could pass an isSignLanguage argument.
</blockquote></details>

Done.
___
*[`src/BloomExe/Collection/WritingSystem.cs` line 147 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6627#-O6398hW9nEsS0XUJk0x:-O6Rk2AtCwb9TLIuAsQw:b-l8v9s4) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/cd19097ac9850e94593796ca4bb7dda6ebb36aec/src/BloomExe/Collection/WritingSystem.cs#L147)):*

could this just be $"Default{pfx}FontName"? and similarly in the line below? Not a prefix here, but the point is that "LanguageN" is or is not part of the tag.
</blockquote></details>

Done.  Although the details differ.
___
*[`src/BloomExe/Collection/WritingSystem.cs` line 212 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6627#-O63G0Xq9h2C3COloUH0:-O6SHjKG6LsTG854imit:bqky3p6) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/cd19097ac9850e94593796ca4bb7dda6ebb36aec/src/BloomExe/Collection/WritingSystem.cs#L212)):*

Similar suggestions to SaveToXElement.
</blockquote></details>

Done.  similar fix to split into three methods.
___
*[`src/BloomExe/Spreadsheet/SpreadsheetImporter.cs` line 447 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6627#-O63GhGh2bOR0wKsnvFF:-O6SMU4c4CzN7gWsk72J:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/cd19097ac9850e94593796ca4bb7dda6ebb36aec/src/BloomExe/Spreadsheet/SpreadsheetImporter.cs#L447)):*

This could return a boolean that is true if either the language was added or the name was changed. Then we could |= that with _collectionSettingsAreDirty, and (I think) not need special cases for L1/2/3 (though probably still for sign, since it's not in the list). Plus we don't need to save collection settings if nothing changed, which is really desirable because in a TC you have to be an admin to change Collection settings.
</blockquote></details>

Done.
___
*[`src/BloomExe/Spreadsheet/SpreadsheetImporter.cs` line 451 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6627#-O63JwpLAyuH3dPLK6jz:-O6Sywjc6HAf0U-YrMu9:bq7j19z) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/cd19097ac9850e94593796ca4bb7dda6ebb36aec/src/BloomExe/Spreadsheet/SpreadsheetImporter.cs#L451)):*

You'll have to do something to prevent this if we're in a TC and not an admin. There may also be some protocol that CollectionSettings changes in a TC should be immediately auto-committed? Or something to do to make sure the next commit syncs them? I don't remember, but the collection settings dialog should be handling it.
</blockquote></details>

Done.  I think what I've put in should work.

StephenMcConnel (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6627)* status: 0 of 10 files reviewed, 9 unresolved discussions (waiting on @JohnThomson)

___
*[`src/BloomExe/Collection/CollectionSettingsDialog.cs` line 562 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6627#-O63-IYNDsQzNfLQy___:-O6XpDjz26YcehnmH2q_:bhj7yt9) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/cd19097ac9850e94593796ca4bb7dda6ebb36aec/src/BloomExe/Collection/CollectionSettingsDialog.cs#L562)):*

Was I wrong? The list CAN contain a WS in position 3 with an empty tag, representing the fact that the book doesn't have an L3. Was this guarding against removing that?
</blockquote></details>

If you expand the context above, the code is looking only at items past the first 3.  So your comment was apropos.
___
*[`src/BloomExe/Spreadsheet/SpreadsheetImporter.cs` line 451 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6627#-O6XAqu12yhDvohqCJdo:-O6Y1fmc51u4zytX5bnQ:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/5c5799b04ea85dcd118abdea4588674d6ed9116e/src/BloomExe/Spreadsheet/SpreadsheetImporter.cs#L451)):*

It feels weird to me that if we don't Save, in normal Bloom operation we will have changed the in-memory settings but the changes will not persist. I'm not sure of the implications...if we publish the book (e.g., to Blorg or BloomPub), will we get language names from memory or disk? (I'm not even sure which we'd want.) And if we're in command line mode, it will just quietly fail to change the collection settings without any indication.
Could we avoid making the changes if we're not allowed to? e.g., figure out whether it is allowed at the start of the method, and pass that flag as a new argument to UpdateOrCreateCollectionLanguage, which would still answer true if it wanted to make a change, but not actually do so, if allowedToModify is false.
And then maybe put a warning-level message in the log that the importer is building if the import called for changing a language name but the TC status prevented it?
</blockquote></details>

Done.
___
*[`src/BloomExe/Spreadsheet/SpreadsheetImporter.cs` line 459 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6627#-O6X7cO82zcAWb-gSfkb:-O6XuQEmDa3037Dptgh-:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/5c5799b04ea85dcd118abdea4588674d6ed9116e/src/BloomExe/Spreadsheet/SpreadsheetImporter.cs#L459)):*

This appears to be roughly duplicating the logic in OkToEditCollectionSettings. Normal Bloom operation should have a TCM, even if there's no TC, so I'm guessing this is for command-line mode. Could we extract the common code (passed a collectionSettings) into a static method of TCM?
</blockquote></details>

Done.

JohnThomson (approved):
> Reviewed 7 of 9 files at r2, 3 of 3 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6627)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @StephenMcConnel)

PR #6628, opened 2024-09-03
BL-13805 Branding control over credits lang

hatton:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6628)

### DIFF
--- a/src/BloomExe/Book/Book.cs
+++ b/src/BloomExe/Book/Book.cs
@@ -2832,15 +2832,18 @@ namespace Bloom.Book
         /// </summary>
         private bool IsLanguageWantedInContent(SafeXmlElement parent, string lang)
         {
-            var defaultLangs = parent.GetAttribute("data-default-languages");
-            if (String.IsNullOrEmpty(defaultLangs) || defaultLangs == "auto")
+            var dataDefaultLanguages = parent.GetAttribute("data-default-languages");
+
+            if (String.IsNullOrEmpty(dataDefaultLanguages) || dataDefaultLanguages == "auto")
                 return true; // assume we want everything
-            var dataDefaultLanguages = defaultLangs.Split(
-                new char[] { ',', ' ' },
-                StringSplitOptions.RemoveEmptyEntries
+
+            var defaultLangs = TranslationGroupManager.ProcessLanguageList(
+                dataDefaultLanguages,
+                _bookData
             );
-            foreach (var defaultLang in dataDefaultLanguages)
+            foreach (var defaultLang in defaultLangs)
             {
+                // enhance: have ProcessLanguageList() just do these conversions so that this becomes a simple comparison between lang and the defaultLang
                 switch (defaultLang)
                 {
                     case "V":
--- a/src/BloomExe/Book/TranslationGroupManager.cs
+++ b/src/BloomExe/Book/TranslationGroupManager.cs
@@ -4,6 +4,7 @@ using System.Diagnostics;
 using System.Linq;
 using System.Xml;
 using Bloom.SafeXml;
+using Bloom.Api;
 using L10NSharp;
 using SIL.Reporting;
 using SIL.WritingSystems;
@@ -277,10 +278,7 @@ namespace Bloom.Book
                 // This controls visibility for fields not yet controlled by the Appearance system
                 // (or in legacy mode).
                 var defLangsAttr = group.GetAttribute("data-default-languages");
-                var dataDefaultLanguages = defLangsAttr.Split(
-                    new char[] { ',', ' ' },
-                    StringSplitOptions.RemoveEmptyEntries
-                );
+                var dataDefaultLanguages = ProcessLanguageList(defLangsAttr, bookData);
 
                 //nb: we don't necessarily care that a div is editable or not
                 UpdateGeneratedClassesOnChildren(
@@ -313,6 +311,45 @@ namespace Bloom.Book
                     dataDefaultLanguages
                 );
             }
+        }
+
+        public static string[] ProcessLanguageList(
+            string languagesAttributeString,
+            BookData bookData
+        )
+        {
+            // Enhance: could probably just do all the "V", "L1", etc. replacements here, rather than in ShouldNormallyShowEditable()
+            // and Book.IsLanguageWantedInContent() and then simplify those functions.
+
+            var placeholdersWithDefaults = new Dictionary<string, string>
+            {
+                // Allow branding to set the language for the cover credits (needed by Angola in 2024)
+                { "CoverCreditsLanguage", "V" },
+            };
+            // As of Bloom 6.0, only front cover credits has one of these placeholders.
+            // Example: +field-prototypeDeclaredExplicity("[CoverCreditsLanguage]")
+            if (languagesAttributeString.Contains("["))
+            {
+                var brandingSettings = BrandingSettings.GetSettingsOrNull(
+                    bookData.CollectionSettings.BrandingProjectKey
+                );
+                foreach (var placeholderWithDefault in placeholdersWithDefaults)
+                {
+                    var replacement = brandingSettings?.GetPresetKeyValueOrDefault(
+                        placeholderWithDefault.Key,
+                        placeholderWithDefault.Value // default if not specified by this branding
+                    );
+                    languagesAttributeString = languagesAttributeString.Replace(
+                        // we are using square brackets in the pug for no particular reason, just looks right
+                        "[" + placeholderWithDefault.Key + "]",
+                        replacement
+                    );
+                }
+            }
+            return languagesAttributeString.Split(
+                new char[] { ',', ' ' },
+                StringSplitOptions.RemoveEmptyEntries
+            );
         }
 
         public static bool IsPageAffectedByLanguageMenu(SafeXmlElement page, bool legacy)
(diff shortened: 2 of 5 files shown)


JohnThomson (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6628)* status: 0 of 5 files reviewed, 3 unresolved discussions (waiting on @hatton)

___
*[`src/BloomExe/Book/TranslationGroupManager.cs` line 328 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6628#-O5xr2YfCxQnMy-bZbrS:-O5xr2YfCxQnMy-bZbrT:b-nhu21i) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/ec96000fe1978162c0c0e756fb024a67b5c7c918/src/BloomExe/Book/TranslationGroupManager.cs#L328)):*
> ```Smalltalk
>                 // Allow branding to set the language for the cover credits (needed by Angola in 2024)
>                 { "CoverCreditsLanguage", "V" },
>             };
> ```

I don't like having the defaults recorded here.
- it seems overly complex
- it will be hard to find (we expect to find these settings in the pug file)
- adding a new branding setting of this type will require code changes, not just pug/branding changes
- it can't have a different default in different xmatters, which I think is a quite plausible thing to want
Conversely, I don't foresee any need for combining a value taken from the data-default-languages attribute with one taken from the branding.
Why can't the default be present in the attribute along with the branding key? So if you find an attribute value like "[CoverCreditsLanguage] V" then you look for CoverCreditsLanguage as a key in the branding. If you find it, the branding value entirely replaces the attribute value. If you don't, you just delete the key and use the rest of the attribute contents ("V") as the value. Nothing in C# code even has to know that CoverCreditsLanguage is a possible branding key, much less what its default should be.
(Even wilder idea: this code looks down into the bloom-editable and notices that there is a data-book="smallCoverCredits". It then looks for a branding key smallCoverCredits-languages. If found, that becomes the languages for that translation group. If not, use the attribute value. With this approach, you don't even have to change the xmatter pug to enable this for a new field in a new branding. OTOH, I don't like giving data-book a new purpose, or having to dig down to the bloom-editable, so overall I think I prefer not to go this way.)
If you don't like either of those ideas, I think at least the variable needs a comment along the lines that is is a map from a key that can occur in data-default-languages to the default set of languages that should be used if the branding doesn't specify one for that field. And placeholder is to vague a name...how about defaultLanguageKeyToDefaultLanguages?
___
*[`src/BloomExe/Book/TranslationGroupManager.cs` line 329 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6628#-O5xTH3X0cxSiRtwbINl:-O5xTH3X0cxSiRtwbINm:b-x3pg4i) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/ec96000fe1978162c0c0e756fb024a67b5c7c918/src/BloomExe/Book/TranslationGroupManager.cs#L329)):*
> ```Smalltalk
>                 { "CoverCreditsLanguage", "V" },
>             };
>             // As of Bloom 6.0, only front cover credits has one of these placeholders.
> ```

This is now targeting 6.1, right?
___
*[`src/BloomExe/web/controllers/BrandingSettings.cs` line 166 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6628#-O5xUl3v-kTwVppFdcK-:-O5xUl3v-kTwVppFdcK0:b-k8apyp) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/ec96000fe1978162c0c0e756fb024a67b5c7c918/src/BloomExe/web/controllers/BrandingSettings.cs#L166)):*
> ```Smalltalk
>                     return CachedBrandingSettings;
>                 }
>             }
> ```

This block should use the lock, too. Otherwise, some other thread could change CachedBrandingSettings between when you check the name and time and when you actually return the settings.

hatton (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6628)* status: 0 of 5 files reviewed, 3 unresolved discussions (waiting on @JohnThomson)

___
*[`src/BloomExe/Book/TranslationGroupManager.cs` line 328 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6628#-O5xr2YfCxQnMy-bZbrS:-O63B_HK7tca9kDMq80I:b-5bx2nx) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/ec96000fe1978162c0c0e756fb024a67b5c7c918/src/BloomExe/Book/TranslationGroupManager.cs#L328)):*

I don't like having the defaults recorded here.
- it seems overly complex
- it will be hard to find (we expect to find these settings in the pug file)
- adding a new branding setting of this type will require code changes, not just pug/branding changes
- it can't have a different default in different xmatters, which I think is a quite plausible thing to want
Conversely, I don't foresee any need for combining a value taken from the data-default-languages attribute with one taken from the branding.
Why can't the default be present in the attribute along with the branding key? So if you find an attribute value like "[CoverCreditsLanguage] V" then you look for CoverCreditsLanguage as a key in the branding. If you find it, the branding value entirely replaces the attribute value. If you don't, you just delete the key and use the rest of the attribute contents ("V") as the value. Nothing in C# code even has to know that CoverCreditsLanguage is a possible branding key, much less what its default should be.
(Even wilder idea: this code looks down into the bloom-editable and notices that there is a data-book="smallCoverCredits". It then looks for a branding key smallCoverCredits-languages. If found, that becomes the languages for that translation group. If not, use the attribute value. With this approach, you don't even have to change the xmatter pug to enable this for a new field in a new branding. OTOH, I don't like giving data-book a new purpose, or having to dig down to the bloom-editable, so overall I think I prefer not to go this way.)
If you don't like either of those ideas, I think at least the variable needs a comment along the lines that is is a map from a key that can occur in data-default-languages to the default set of languages that should be used if the branding doesn't specify one for that field. And placeholder is to vague a name...how about defaultLanguageKeyToDefaultLanguages?
</blockquote></details>

Thanks, I like it better now.
___
*[`src/BloomExe/Book/TranslationGroupManager.cs` line 329 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6628#-O5xTH3X0cxSiRtwbINl:-O63BYPvDEhL7DD4JDOn:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/ec96000fe1978162c0c0e756fb024a67b5c7c918/src/BloomExe/Book/TranslationGroupManager.cs#L329)):*

This is now targeting 6.1, right?
</blockquote></details>

Done.

JohnThomson (commented):
I like this much better. One comment (regarding locking the cache) not addressed yet.

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6628#-:-O66XR6Y3Na_MJms6oIq:bcfwhfb)* status: 0 of 5 files reviewed, 1 unresolved discussion (waiting on @hatton)

___
*[`src/BloomExe/Book/TranslationGroupManager.cs` line 328 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6628#-O5xr2YfCxQnMy-bZbrS:-O66XCS_BNl2UGcfFRbg:bnvumlm) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/ec96000fe1978162c0c0e756fb024a67b5c7c918/src/BloomExe/Book/TranslationGroupManager.cs#L328)):*

Thanks, I like it better now.
</blockquote></details>

Me too! Nice work.

hatton (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6628)* status: 0 of 5 files reviewed, 1 unresolved discussion (waiting on @JohnThomson)

___
*[`src/BloomExe/web/controllers/BrandingSettings.cs` line 166 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6628#-O5xUl3v-kTwVppFdcK-:-O68EW3kAei6FEXdiDfY:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/ec96000fe1978162c0c0e756fb024a67b5c7c918/src/BloomExe/web/controllers/BrandingSettings.cs#L166)):*

This block should use the lock, too. Otherwise, some other thread could change CachedBrandingSettings between when you check the name and time and when you actually return the settings.
</blockquote></details>

Done.

JohnThomson (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6628)* status: 0 of 5 files reviewed, 2 unresolved discussions (waiting on @hatton)

___
*[`src/BloomExe/web/controllers/BrandingSettings.cs` line 158 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6628#-O68ZjbT8kESEsqP4xJg:-O68ZjbT8kESEsqP4xJh:b4b0sy7) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/55fbc785e1fc24edb12b4ef169efe9ad8189e98a/src/BloomExe/web/controllers/BrandingSettings.cs#L158)):*
> ```Smalltalk
>             if (CachedBrandingSettingsName == brandingNameOrFolderPath)
>             {
>                 lock (_cacheLock)
> ```

Sorry, but I think the lock should be outside this. CachedBrandingSettingsName is one of the things that the lock is supposed to guard, so that it can't be changed by another thread between starting this method and determining its result.
(Of course, maybe this is something that can't change without restarting Bloom, so it doesn't really need to be locked. If so, worth commenting.)

JohnThomson (approved):
> Reviewed 2 of 5 files at r1, 2 of 2 files at r2, 1 of 1 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6628)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @hatton)

PR #6629, opened 2024-09-03
BL-13819 Move cover background color to Appearance

hatton:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6629)


JohnThomson (changes requested):
Looks good. Just a few niggles.

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6629#-:-O5yDgqp5fNoQL1rXoV1:bp902ga)* status: all files reviewed, 14 unresolved discussions (waiting on @hatton)

___
*[`DistFiles/localization/en/BloomMediumPriority.xlf` line 611 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6629#-O5xzlGkB5C8HBwFAUCa:-O5xzlGkB5C8HBwFAUCb:b-irjwog) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/dde699c6581a6ccd62d743a2040c301ec8a15206/DistFiles/localization/en/BloomMediumPriority.xlf#L611)):*
> ```XML
>         <note>BookSettings.Theme.Description</note>
>       </trans-unit>
>       <trans-unit id="Common.BackgroundColor" sil:dynamic="true">
> ```

We usually put translate="no" on new settings until the version goes beta.
We usually don't remove old keys, just add a note that they are "obsolete as of Bloom 6.1". IIRC the reason is that removing the old key will cause all old translations to be discarded from CrowdIn before Bloom 6.0 and before are done with them.
___
*[`src/BloomBrowserUI/bookEdit/bookSettings/BookSettingsDialog.tsx` line 415 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6629#-O5y-quf1c6as0a_omC1:-O5y-quf1c6as0a_omC2:bmzyfkn) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/dde699c6581a6ccd62d743a2040c301ec8a15206/src/BloomBrowserUI/bookEdit/bookSettings/BookSettingsDialog.tsx#L415)):*
> ```XML
>                                     label={coverBackgroundColorLabel}
>                                     control={ColorPickerForConfigr}
>                                     disabled={appearanceDisabled}
> ```

Don't we need something here to disable it when the branding locks it?
___
*[`src/BloomExe/Book/AppearanceSettings.cs` line 551 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6629#-O5y0ZrcB0IsMIVixtLV:-O5y0ZrcB0IsMIVixtLW:b-2tj2vw) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/dde699c6581a6ccd62d743a2040c301ec8a15206/src/BloomExe/Book/AppearanceSettings.cs#L551)):*
> ```Smalltalk
>         else
>         {
>             if (bookFolderPath.Contains("template books"))
> ```

Should probably make use of BloomFileLocator.FactoryTemplateBookDirectory
___
*[`src/BloomExe/Book/Book.cs` line 3142 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6629#-O5y4LN40is_hRH4Yl62:-O5y4LN40is_hRH4Yl63:b-vhnrij) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/dde699c6581a6ccd62d743a2040c301ec8a15206/src/BloomExe/Book/Book.cs#L3142)):*
> ```Smalltalk
>         }
> 
>         private void UpdateFallbackAppearanceCoverBackgroundColor(string color)
> ```

This needs a better name or comment because, even though I helped come up with the idea, I had a hard time figuring out in what sense it is setting a fallback. It's not, for example, the cover color that will be used if the user doesn't specify one in the book settings dialog. It's actually creating the rule that will control cover color for most books. The exceptions are those where xmatter or  branding or some other stylesheet overrides it.
We should also capture, at least in a comment, that it's really important for this rule to come ahead of other stylesheets that the book imports, so they CAN override it without using !important or some specificity trick.
Maybe SetOverrideableAppearanceCoverBackgroundColor? Or Initial?
___
*[`src/BloomExe/Book/Book.cs` line 3149 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6629#-O5y8QQT6MSMB8yDVcUT:-O5y8QQT6MSMB8yDVcUU:brgvljd) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/dde699c6581a6ccd62d743a2040c301ec8a15206/src/BloomExe/Book/Book.cs#L3149)):*
> ```Smalltalk
>                 "appearanceCoverBackgroundColor",
>                 $".bloom-page {{ --cover-background-color: {color}; }}",
>                 HtmlDom.GetUserModifiedStyleElement(RawDom.Head)
> ```

Deserves a comment. Why do we want it to follow the user-defined styles element?
Is there any way a book might not have a user-defined stylesheet? We definitely don't want this at the END of the head.
___
*[`src/BloomExe/Book/Book.cs` line 3152 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6629#-O5y6N8J2aitYje3VYSg:-O5y6N8J2aitYje3VYSh:b-oddzet) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/dde699c6581a6ccd62d743a2040c301ec8a15206/src/BloomExe/Book/Book.cs#L3152)):*
> ```Smalltalk
>             );
> 
>             // if the appearance setting for this color is missing, set it now
> ```

Why only if it's missing? We're setting the style rule in the book that will actually make the cover this color (absent some override); why would we want to leave the appearance setting something else?
___
*[`src/BloomExe/Book/Book.cs` line 3154 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6629#-O5y5IeBDr-kmD34ZLSV:-O5y5IeC31h33PMyRHGc:b3ybvdf) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/dde699c6581a6ccd62d743a2040c301ec8a15206/src/BloomExe/Book/Book.cs#L3154)):*
> ```Smalltalk
>             // if the appearance setting for this color is missing, set it now
>             if (
>                 !string.IsNullOrWhiteSpace(color)
> ```

When would we ever call this method with an empty argument?
And if we did, would we really want to set the cover-background-color variable (above) to empty, making the cover black?
___
*[`src/BloomExe/Book/Book.cs` line 3184 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6629#-O5y8lEYD5gPLvLJPfSP:-O5y8lEYD5gPLvLJPfSQ:b1fsz5j) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/dde699c6581a6ccd62d743a2040c301ec8a15206/src/BloomExe/Book/Book.cs#L3184)):*
> ```Smalltalk
>                 "legacyCoverBackgroundColor", // legacy here means pre-6.1
>                 $"DIV.bloom-page.coverColor {{ background-color: {color} !important;}}",
>                 HtmlDom.GetUserModifiedStyleElement(RawDom.Head)
> ```

Again, why in that particular spot?
___
*[`src/BloomExe/Book/Book.cs` line 4225 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6629#-O5y734x9pby2gxrgk_0:-O5y734x9pby2gxrgk_1:b6pfoc) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/dde699c6581a6ccd62d743a2040c301ec8a15206/src/BloomExe/Book/Book.cs#L4225)):*
> ```Smalltalk
>             }
>             if (!FullBleed)
>                 SetBackwardsCompatibleCoverBackgroundColor(printingDom, "white");
> ```

I'm guessing we want a white cover background in the expectation that the real cover background color will come from the paper we print on. (Nice to have a comment explaining that, if correct.)  If that's the purpose, why would we only do it for legacy books?
___
*[`src/BloomExe/Book/Book.cs` line 5671 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6629#-O5y7eZa8c0Hxcop_a0N:-O5y7eZa8c0Hxcop_a0O:b-2mquaa) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/dde699c6581a6ccd62d743a2040c301ec8a15206/src/BloomExe/Book/Book.cs#L5671)):*
> ```Smalltalk
>             );
>             if (color != null)
>                 SetCoverColor(color);
> ```

The comment on this method says that only tests should use it. That or this should be fixed.
___
*[`src/BloomTests/Book/BookTests.cs` line 1367 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6629#-O5yBHooEosAVWQ444q7:-O5yBHooEosAVWQ444q8:badykwn) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/dde699c6581a6ccd62d743a2040c301ec8a15206/src/BloomTests/Book/BookTests.cs#L1367)):*
> ```Smalltalk
>             );
>             // why this should come first is lost to time
>             Assert.Less(indexOfUserStyles, indexOfCoverColorRule);
> ```

It would be good to check that it ends up before the lines that import appearance.css, xmatter...css, and branding.css, since I think we want any of those to be able to override it without messing with specificity etc.
___
*[`src/BloomTests/Book/BookTests.cs` line 2076 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6629#-O5yC4OcEg7gxR2XKf6d:-O5yC4OcEg7gxR2XKf6e:bvy8bhn) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/dde699c6581a6ccd62d743a2040c301ec8a15206/src/BloomTests/Book/BookTests.cs#L2076)):*
> ```Smalltalk
>                     "//div[contains(@class, 'bloom-page') and contains(@class, 'darkCoverColor')]"
>                 )
>             );
> ```

We would usually use AssertThatXmlIn(book.RawDom).HasNoMatchFor(...), though I'm not sure it's much better.
___
*[`src/BloomTests/Book/BookTests.cs` line 2109 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6629#-O5yCV4AAnqu8_5DpSvM:-O5yCV4AAnqu8_5DpSvN:b-2qzdne) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/dde699c6581a6ccd62d743a2040c301ec8a15206/src/BloomTests/Book/BookTests.cs#L2109)):*
> ```Smalltalk
>             // should look like a hex color
>             Assert.IsTrue(result.StartsWith("#"));
>             Assert.IsTrue(result.Length == 7);
> ```

Worth asserting that it's not #000000? Probably not; that happens if we don't set anything, but it probably doesn't get returned by this method.
___
*[`src/content/bookLayout/basePage.less` line 650 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6629#-O5yCmD01R-b_isIZUO1:-O5yCmD01R-b_isIZUO2:b2nh1lg) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/dde699c6581a6ccd62d743a2040c301ec8a15206/src/content/bookLayout/basePage.less#L650)):*
> ```Less
> 
> // the extra specificity here is to win out over the legacy (but still active) cover-color rule in the html itself
> // As of 6.0 this is only used for Angola branding
> ```

6.1

src/BloomTests/Book/BookTests.cs
```diff
@@ -98,42 +96,54 @@ DIV.coverColor  TEXTAREA {
...
             var book = CreateBook();
-            var dom = book.RawDom;
-            book.SetCoverColorInternal(newValue);
-            var coverColorText = dom.SafeSelectNodes("//style[text()]")[0].InnerText;
-            var first = coverColorText.IndexOf(newValue, StringComparison.InvariantCulture);
-            var last = coverColorText.LastIndexOf(newValue, StringComparison.InvariantCulture);
-            Assert.That(first > 0);
-            Assert.That(last > 0 && last != first);
+            book.SetCoverColor(newValue);
+            Assert.That(GetLegacyCoverColorStyleNode(book).InnerText.Contains(newValue));
```
hatton, line 100: This is  a test

DistFiles/localization/en/BloomMediumPriority.xlf
```diff
@@ -608,9 +608,9 @@
         <source xml:lang="en">Page Themes are a bundle of margins, borders, and other page settings. For information about each theme, see [Page Themes Catalog](https://docs.bloomlibrary.org/page-themes-catalog).</source>
         <note>BookSettings.Theme.Description</note>
       </trans-unit>
-      <trans-unit id="BookSettings.CoverBackground" sil:dynamic="true">
-        <source xml:lang="en">Cover Background</source>
-        <note>BookSettings.CoverBackground</note>
+      <trans-unit id="Common.BackgroundColor" sil:dynamic="true">
+        <source xml:lang="en">Background Color</source>
+        <note>Common.CoverBackgroundColor</note>
```
hatton, line 613: This is a test, trying to see in vscode

hatton (commented):
Gotta say, it felt much more like an interrogation than a few little niggles. Could me be, though.

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6629#-:-O685SKvDeaC_-3Bim-O:b-576wb6)* status: all files reviewed, 14 unresolved discussions (waiting on @JohnThomson)

___
*[`DistFiles/localization/en/BloomMediumPriority.xlf` line 611 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6629#-O5xzlGkB5C8HBwFAUCa:-O67xO1H3ClbiHiZGMys:b-qxjzff) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/dde699c6581a6ccd62d743a2040c301ec8a15206/DistFiles/localization/en/BloomMediumPriority.xlf#L611)):*

We usually put translate="no" on new settings until the version goes beta.
We usually don't remove old keys, just add a note that they are "obsolete as of Bloom 6.1". IIRC the reason is that removing the old key will cause all old translations to be discarded from CrowdIn before Bloom 6.0 and before are done with them.
</blockquote></details>

I'm happy for these to be translated now, and using the new key.
___
*[`src/BloomBrowserUI/bookEdit/bookSettings/BookSettingsDialog.tsx` line 415 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6629#-O5y-quf1c6as0a_omC1:-O67y68HA4W-VnHHgvia:b-njcnos) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/dde699c6581a6ccd62d743a2040c301ec8a15206/src/BloomBrowserUI/bookEdit/bookSettings/BookSettingsDialog.tsx#L415)):*

Don't we need something here to disable it when the branding locks it?
</blockquote></details>

`getAdditionProps()` does that
___
*[`src/BloomExe/Book/AppearanceSettings.cs` line 551 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6629#-O5y0ZrcB0IsMIVixtLV:-O67yaxhEfb-vs6vQYIA:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/dde699c6581a6ccd62d743a2040c301ec8a15206/src/BloomExe/Book/AppearanceSettings.cs#L551)):*

Should probably make use of BloomFileLocator.FactoryTemplateBookDirectory
</blockquote></details>

Done.
___
*[`src/BloomExe/Book/Book.cs` line 3142 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6629#-O5y4LN40is_hRH4Yl62:-O67z7zD49VJjTuc5QIT:b7pye9h) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/dde699c6581a6ccd62d743a2040c301ec8a15206/src/BloomExe/Book/Book.cs#L3142)):*

This needs a better name or comment because, even though I helped come up with the idea, I had a hard time figuring out in what sense it is setting a fallback. It's not, for example, the cover color that will be used if the user doesn't specify one in the book settings dialog. It's actually creating the rule that will control cover color for most books. The exceptions are those where xmatter or  branding or some other stylesheet overrides it.
We should also capture, at least in a comment, that it's really important for this rule to come ahead of other stylesheets that the book imports, so they CAN override it without using !important or some specificity trick.
Maybe SetOverrideableAppearanceCoverBackgroundColor? Or Initial?
</blockquote></details>

Hmm, the existing comment explains it already:

```
// This is here because basePage wants to use  --cover-background-color but if this variable is
// undefined, you get black. You'd like CSS to just ignore the rule, but we found no way to do that.
```

I'll use the word "fallback in a comment" and say that other code puts css links later in the file.
___
*[`src/BloomExe/Book/Book.cs` line 3149 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6629#-O5y8QQT6MSMB8yDVcUT:-O67zuQ_2Aj0zglE-9MF:ba9oji8) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/dde699c6581a6ccd62d743a2040c301ec8a15206/src/BloomExe/Book/Book.cs#L3149)):*

Deserves a comment. Why do we want it to follow the user-defined styles element?
Is there any way a book might not have a user-defined stylesheet? We definitely don't want this at the END of the head.
</blockquote></details>

Actually I don't know (or care much). It just preserves the order that was there. I'll add a comment 
___
*[`src/BloomExe/Book/Book.cs` line 3154 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6629#-O5y5IeBDr-kmD34ZLSV:-O682tK1B32AkrMCi1hw:b-zf21hj) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/dde699c6581a6ccd62d743a2040c301ec8a15206/src/BloomExe/Book/Book.cs#L3154)):*

When would we ever call this method with an empty argument?
And if we did, would we really want to set the cover-background-color variable (above) to empty, making the cover black?
</blockquote></details>

Are you really complaining that the code is too defensive?
___
*[`src/BloomExe/Book/Book.cs` line 3184 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6629#-O5y8lEYD5gPLvLJPfSP:-O683JBD9BXBKvvck2r9:b-bttoee) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/dde699c6581a6ccd62d743a2040c301ec8a15206/src/BloomExe/Book/Book.cs#L3184)):*

Again, why in that particular spot?
</blockquote></details>

Added comment
___
*[`src/BloomExe/Book/Book.cs` line 4225 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6629#-O5y734x9pby2gxrgk_0:-O683ORQ8zWSmqvVsenn:b-24ezb7) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/dde699c6581a6ccd62d743a2040c301ec8a15206/src/BloomExe/Book/Book.cs#L4225)):*

I'm guessing we want a white cover background in the expectation that the real cover background color will come from the paper we print on. (Nice to have a comment explaining that, if correct.)  If that's the purpose, why would we only do it for legacy books?
</blockquote></details>

I don't know John. That decision is in the old code. 
___
*[`src/BloomExe/Book/Book.cs` line 5671 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6629#-O5y7eZa8c0Hxcop_a0N:-O684hFt0LhUV41nP92h:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/dde699c6581a6ccd62d743a2040c301ec8a15206/src/BloomExe/Book/Book.cs#L5671)):*

The comment on this method says that only tests should use it. That or this should be fixed.
</blockquote></details>

Done.
___
*[`src/BloomTests/Book/BookTests.cs` line 1367 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6629#-O5yBHooEosAVWQ444q7:-O67nccf4Ja656VUk-Wx:b5kku9v) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/dde699c6581a6ccd62d743a2040c301ec8a15206/src/BloomTests/Book/BookTests.cs#L1367)):*

It would be good to check that it ends up before the lines that import appearance.css, xmatter...css, and branding.css, since I think we want any of those to be able to override it without messing with specificity etc.
</blockquote></details>

Yeah, but that appears to be up to the code that adds styles... i.e. it's not up to this code to relocate after the fact; the code that adds links puts them at the end of the `head`.
___
*[`src/BloomTests/Book/BookTests.cs` line 2076 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6629#-O5yC4OcEg7gxR2XKf6d:-O63PKoW--VmKxyoj_P2:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/dde699c6581a6ccd62d743a2040c301ec8a15206/src/BloomTests/Book/BookTests.cs#L2076)):*

We would usually use AssertThatXmlIn(book.RawDom).HasNoMatchFor(...), though I'm not sure it's much better.
</blockquote></details>

Done.
___
*[`src/BloomTests/Book/BookTests.cs` line 2109 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6629#-O5yCV4AAnqu8_5DpSvM:-O63OoE29tZ6BQQTm8-a:b-rx44q5) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/dde699c6581a6ccd62d743a2040c301ec8a15206/src/BloomTests/Book/BookTests.cs#L2109)):*

Worth asserting that it's not #000000? Probably not; that happens if we don't set anything, but it probably doesn't get returned by this method.
</blockquote></details>

I don't care, at least in the test, if it's white or not. If something decides that should be the default, this doesn't need to fail. I just don't want it to throw an exception.
___
*[`src/content/bookLayout/basePage.less` line 650 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6629#-O5yCmD01R-b_isIZUO1:-O63MEf81c5hPGRcm1OK:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/dde699c6581a6ccd62d743a2040c301ec8a15206/src/content/bookLayout/basePage.less#L650)):*

6.1
</blockquote></details>

Done.

hatton (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6629)* status: 9 of 13 files reviewed, 14 unresolved discussions (waiting on @JohnThomson)

___
*[`src/BloomExe/Book/Book.cs` line 3152 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6629#-O5y6N8J2aitYje3VYSg:-O6RtlwE8dvefSLc2byH:bb6un1j) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/dde699c6581a6ccd62d743a2040c301ec8a15206/src/BloomExe/Book/Book.cs#L3152)):*

Why only if it's missing? We're setting the style rule in the book that will actually make the cover this color (absent some override); why would we want to leave the appearance setting something else?
</blockquote></details>

Done

JohnThomson (changes requested):
Sorry if I was too picky. The only thing I think still needs more thought is whether books using the new cover color scheme should have the cover color set white for printing. I doubt very much that should be a legacy-only behavior.

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6629#-:-O6S_ZjI8cXIawtrGCPL:bdm4izi)* status: 9 of 13 files reviewed, 1 unresolved discussion (waiting on @hatton)

___
*[`src/BloomExe/Book/Book.cs` line 3154 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6629#-O5y5IeBDr-kmD34ZLSV:-O6SIT7B1dV1E9bUnfyS:b-2sl81k) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/dde699c6581a6ccd62d743a2040c301ec8a15206/src/BloomExe/Book/Book.cs#L3154)):*

Are you really complaining that the code is too defensive?
</blockquote></details>

I was confused, I think. Anyway it's gone now.
___
*[`src/BloomExe/Book/Book.cs` line 4225 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6629#-O5y734x9pby2gxrgk_0:-O6SZUfN6V8UqSQKehyA:bwjvs3w) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/dde699c6581a6ccd62d743a2040c301ec8a15206/src/BloomExe/Book/Book.cs#L4225)):*

I don't know John. That decision is in the old code.
</blockquote></details>

My point is, before this change, almost all books (except full bleed) were apparently printed with white cover background color. I'm thinking it very likely that we don't just need to set the backwards compatible cover color to white here, but somehow make the cover white in modern books' printing DOMs, too.

hatton (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6629)* status: 8 of 13 files reviewed, 1 unresolved discussion (waiting on @JohnThomson)

___
*[`src/BloomExe/Book/Book.cs` line 4225 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6629#-O5y734x9pby2gxrgk_0:-O6Sxo82BrCvBgOsvPPB:b-j8uwz6) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/dde699c6581a6ccd62d743a2040c301ec8a15206/src/BloomExe/Book/Book.cs#L4225)):*

My point is, before this change, almost all books (except full bleed) were apparently printed with white cover background color. I'm thinking it very likely that we don't just need to set the backwards compatible cover color to white here, but somehow make the cover white in modern books' printing DOMs, too.
</blockquote></details>

OK, thanks for that.

JohnThomson (changes requested):
> Reviewed 3 of 4 files at r2, 2 of 2 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6629)* status: all files reviewed, 4 unresolved discussions (waiting on @hatton)

___
*[`src/BloomExe/Book/Book.cs` line 4225 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6629#-O5y734x9pby2gxrgk_0:-O6WuiDD2ti4m_LxulYl:b-gtxif7) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/dde699c6581a6ccd62d743a2040c301ec8a15206/src/BloomExe/Book/Book.cs#L4225)):*

OK, thanks for that.
</blockquote></details>

No doubt this works, but it feels ugly to use a trick to make the legacy rule work in appearance-system books. What would you think of appending a rule to Head that forces the new cover color variable to white?
Then again, maybe I've been picky enough and held this PR up long enough. I can live with this if you don't think it's worth changing.
___
*[`src/BloomExe/Book/Book.cs` line 3173 at r4](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6629#-O6WuSjUBywJntyJKIuk:-O6WuSjUBywJntyJKIul:b3mu3me) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/73503584a26ee6e6d0ec9f3ef95cdc337253d46b/src/BloomExe/Book/Book.cs#L3173)):*
> ```Smalltalk
>         // Add the old-style rule for books either using legacy theme or running in older
>         // Bloom Editors that won't read --cover-background-color so that they just get the original color
>         // This rule will get overridden in contexts using appearance
> ```

...except when overrideAllOtherRules is true.
___
*[`src/BloomExe/Book/Book.cs` line 3194 at r4](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6629#-O6Wt3xFCPZAYA1c47tz:-O6Wt3xGCbSTtWgBsT3a:bo12aox) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/73503584a26ee6e6d0ec9f3ef95cdc337253d46b/src/BloomExe/Book/Book.cs#L3194)):*
> ```Smalltalk
>                 $"DIV.bloom-page.coverColor {{ background-color: {color} !important;}}",
>                 // I don't know if the order matters, but it has to go *somewhere* and in pre-6.1 bloom put it here,
>                 // and I see no reason to start putting it somewhere else.
> ```

The exact position may not matter, but it's important that (except when overrideAllOtherRules is true) it comes before we import the stylesheets that SHOULD be able to override it. I think that should at least be clear in the comment. It actually feels somewhat fragile if I'm right that it only works because we currently put the user-defined styles element before the style elements that pull in files, but I'm not sure what would be better.
___
*[`src/BloomExe/Book/HtmlDom.cs` line 348 at r4](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6629#-O6Wpwz38RrsDI1I1tOv:-O6Wpwz38RrsDI1I1tOw:b3fzhp8) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/73503584a26ee6e6d0ec9f3ef95cdc337253d46b/src/BloomExe/Book/HtmlDom.cs#L348)):*
> ```Smalltalk
>             string name,
>             string content,
>             SafeXmlNode nodeThisShouldComeAfter // set to null if it should come at the end
> ```

But it doesn't...it puts it at the beginning.

hatton (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6629)* status: all files reviewed, 4 unresolved discussions (waiting on @JohnThomson)

___
*[`src/BloomExe/Book/Book.cs` line 4225 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6629#-O5y734x9pby2gxrgk_0:-O6XVxrB5uK-feVT3iHm:bca58xo) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/dde699c6581a6ccd62d743a2040c301ec8a15206/src/BloomExe/Book/Book.cs#L4225)):*

No doubt this works, but it feels ugly to use a trick to make the legacy rule work in appearance-system books. What would you think of appending a rule to Head that forces the new cover color variable to white?
Then again, maybe I've been picky enough and held this PR up long enough. I can live with this if you don't think it's worth changing.
</blockquote></details>

Yeah a review with the posture of "could this possibly be better?" can be exhausting from the receiving end.
___
*[`src/BloomExe/Book/Book.cs` line 3173 at r4](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6629#-O6WuSjUBywJntyJKIuk:-O6XaV7g4YhqMW16_D4b:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/73503584a26ee6e6d0ec9f3ef95cdc337253d46b/src/BloomExe/Book/Book.cs#L3173)):*

...except when overrideAllOtherRules is true.
</blockquote></details>

Done.
___
*[`src/BloomExe/Book/Book.cs` line 3194 at r4](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6629#-O6Wt3xFCPZAYA1c47tz:-O6Xac6O7BkHuuK2QQGp:b5mzvdi) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/73503584a26ee6e6d0ec9f3ef95cdc337253d46b/src/BloomExe/Book/Book.cs#L3194)):*

The exact position may not matter, but it's important that (except when overrideAllOtherRules is true) it comes before we import the stylesheets that SHOULD be able to override it. I think that should at least be clear in the comment. It actually feels somewhat fragile if I'm right that it only works because we currently put the user-defined styles element before the style elements that pull in files, but I'm not sure what would be better.
</blockquote></details>

I'm exhausted by this PR an your thoughts on it.
___
*[`src/BloomExe/Book/HtmlDom.cs` line 348 at r4](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6629#-O6Wpwz38RrsDI1I1tOv:-O6XVtzlCAU46ku7LNeA:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/73503584a26ee6e6d0ec9f3ef95cdc337253d46b/src/BloomExe/Book/HtmlDom.cs#L348)):*

But it doesn't...it puts it at the beginning.
</blockquote></details>

Done.

PR #6637, opened 2024-09-05
Update Activity.css to Games.css (BL-13815)

StephenMcConnel:
But maintain backward compatibility so that older Blooms could still edit the book.  (Albeit with probably the same issue that triggered BL-13815.)

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6637)


StephenMcConnel:
We've decided a different approach is needed.

PR #6641, opened 2024-09-07
More small fixes

JohnThomson:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6641)

### DIFF
--- a/src/BloomBrowserUI/bookEdit/css/editMode.less
+++ b/src/BloomBrowserUI/bookEdit/css/editMode.less
@@ -579,6 +579,10 @@ div.pageLabel[contenteditable="true"]:focus {
 // In edit mode we don't want such an automatic outline, because the control frame
 // provides one, and typically they don't even line up, becuase we grow the control
 // frame a little bit to prevent the side buttons overlapping text.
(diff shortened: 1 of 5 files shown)


hatton (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6641)* status: all files reviewed, 1 unresolved discussion (waiting on @JohnThomson)

___
*[`src/BloomBrowserUI/bookEdit/js/OverlayContextControls.tsx` line 601 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6641#-O6DOvVd0UAAcR8JsuwL:-O6DOvVd0UAAcR8JsuwM:brp2hkc) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/887a0d43024de1d61b3990045cbdfd9cd2a83185/src/BloomBrowserUI/bookEdit/js/OverlayContextControls.tsx#L601)):*
> ```XML
>                     // It would be more natural to handle a click. But clicks are a combination of
>                     // mouse down and mouse up, and those have side effects, especially change of focus,
>                     // that we need to prevent. So we handle them ourselves.
> ```

Clever

PR #6648, opened 2024-09-11
Remove menu as well as overlay control on overlay deletion (BL-13852)

StephenMcConnel:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6648)

### DIFF
--- a/src/BloomBrowserUI/bookEdit/js/bubbleManager.ts
+++ b/src/BloomBrowserUI/bookEdit/js/bubbleManager.ts
@@ -1028,10 +1028,19 @@ export class BubbleManager {
     private gotAMoveWhileMouseDown: boolean = false;
 
     // Remove the overlay control frame if it exists (when no overlay is active)
+    // Also remove the menu if it's still open.  See BL-13852.
     removeControlFrame() {
+        // this.activeElement is still set and works for hiding the menu.
+        const eltWithControlOnIt = this.activeElement;
         const controlFrame = document.getElementById("overlay-control-frame");
         if (controlFrame) {
-            controlFrame.remove();
+            if (eltWithControlOnIt) {
+                renderOverlayContextControls(eltWithControlOnIt, false);
+            }
+            // Reschedule so that the rerender can finish before removing the control frame.
+            setTimeout(() => {
+                controlFrame.remove();
+            }, 0);
         }
     }
 
@@ -3971,6 +3980,10 @@ export class BubbleManager {
 
         // Update UI and make sure things get redrawn correctly.
         this.refreshBubbleEditing(containerElement, undefined, false);
+        // We no longer have an active element, but the old active element may be
+        // needed by the removeControlFrame method called by refreshBubbleEditing
+        // to remove a popup menu.
+        this.setActiveElement(undefined);
         // By this point it's really gone, so this will clean up if it had a target.
         this.removeDetachedTargets();
     }


JohnThomson (approved):
I'm approving this, but as I mentioned in Flowdock, I'd like you to at least consider whether it would be better to hide it by calling  renderOverlayContextControls with menuIsOpen false.

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6648#-:-O6WybRR2eoUyCXLc82m:b-izd9tx)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @StephenMcConnel)

StephenMcConnel (commented):
I've reimplemented the fix along the lines suggested.

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6648#-:-O6Xn0PI-F4OuK3hAYg1:bycdwwn)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @StephenMcConnel)

JohnThomson (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6648)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @StephenMcConnel)

___
*[`src/BloomBrowserUI/bookEdit/js/bubbleManager.ts` line 1034 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6648#-O6apkxV9N9Palt8ZaAZ:-O6apkxV9N9Palt8ZaA_:b-x50gy0) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/b3085aaa6d693bb3f241c5833ba28851eec698fb/src/BloomBrowserUI/bookEdit/js/bubbleManager.ts#L1034)):*
> ```XML
>     removeControlFrame() {
>         // this.activeElement is still set and works for hiding the menu.
>         const eltWithControlOnIt = this.activeElement;
> ```

I got to thinking, how sure are we that it is still set when we delete an overlay? So I played around enough to determine that it NEVER gets cleared...it goes on pointing at the old activeElement until the user selects something else.
That feels wrong. I think something should clean it up. Then we could comment explicitly that it shouldn't be cleared before we removeControlFrame.
It would be plausible for refreshBubbleEditing to SetActiveElement(undefined) if bubble is undefined, but it doesn't feel like its responsibility. Better in my view that deleteTOPBox() should clean it up. But then we would have to SetActiveElement(undefined) AFTER calling refreshBubbleEditing, which feels wrong...every other caller of refreshBubbleEditing sets activeElement to the appropriate thing before calling this method. But short of enhancing OverlayTextControls so that we can call renderOverlayContextControls with an undefined first argument, I think it's our best option.
So, I would like to see a call to SetActiveElement(undefined) after refreshBubbleEditing in deleteTOPbox, with a comment explaining why it is after.
___
*[`src/BloomBrowserUI/bookEdit/js/bubbleManager.ts` line 1044 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6648#-O6alzu8-6IQVUxrBRQu:-O6alzu8-6IQVUxrBRQv:br05xaw) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/b3085aaa6d693bb3f241c5833ba28851eec698fb/src/BloomBrowserUI/bookEdit/js/bubbleManager.ts#L1044)):*
> ```XML
>             setTimeout(() => {
>                 controlFrame.remove();
>             }, 50);
> ```

Sorry, but I think we should try zero here. If that's not enough, we should probably document the shortest time that works on your dev machines, and set it around five times that to allow for slower end user machines. Of course that may end up being 50ms!

StephenMcConnel (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6648)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @JohnThomson)

___
*[`src/BloomBrowserUI/bookEdit/js/bubbleManager.ts` line 1034 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6648#-O6apkxV9N9Palt8ZaAZ:-O6b1kPF5l8loAr1vbgO:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/b3085aaa6d693bb3f241c5833ba28851eec698fb/src/BloomBrowserUI/bookEdit/js/bubbleManager.ts#L1034)):*

I got to thinking, how sure are we that it is still set when we delete an overlay? So I played around enough to determine that it NEVER gets cleared...it goes on pointing at the old activeElement until the user selects something else.
That feels wrong. I think something should clean it up. Then we could comment explicitly that it shouldn't be cleared before we removeControlFrame.
It would be plausible for refreshBubbleEditing to SetActiveElement(undefined) if bubble is undefined, but it doesn't feel like its responsibility. Better in my view that deleteTOPBox() should clean it up. But then we would have to SetActiveElement(undefined) AFTER calling refreshBubbleEditing, which feels wrong...every other caller of refreshBubbleEditing sets activeElement to the appropriate thing before calling this method. But short of enhancing OverlayTextControls so that we can call renderOverlayContextControls with an undefined first argument, I think it's our best option.
So, I would like to see a call to SetActiveElement(undefined) after refreshBubbleEditing in deleteTOPbox, with a comment explaining why it is after.
</blockquote></details>

Done.
___
*[`src/BloomBrowserUI/bookEdit/js/bubbleManager.ts` line 1044 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6648#-O6alzu8-6IQVUxrBRQu:-O6b0AEW49HHGlKGfm2d:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/b3085aaa6d693bb3f241c5833ba28851eec698fb/src/BloomBrowserUI/bookEdit/js/bubbleManager.ts#L1044)):*

Sorry, but I think we should try zero here. If that's not enough, we should probably document the shortest time that works on your dev machines, and set it around five times that to allow for slower end user machines. Of course that may end up being 50ms!
</blockquote></details>

Done.

JohnThomson (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6648)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @StephenMcConnel)

PR #6664, opened 2024-09-17
BL-13844 Simple Padding Control

hatton:


JohnThomson (changes requested):
Very nice work but I think  topLevel-text-padding should move out of the "not in the dialog yet" list, and I have a few suggestions for documenting the magic that makes the empty-string value work for Default.

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6664#-:-O70m5gr-7-GkmIfbAvs:bbycfax)* status: all files reviewed, 1 unresolved discussion (waiting on @hatton)

___
*[`src/BloomBrowserUI/bookEdit/bookSettings/BookSettingsDialog.tsx` line 605 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6664#-O70g2Pl9o37WUf1aibB:-O70g2Pm3SuXNvFqPUKS:b-jzsc84) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/9ea8adfbbdb02f641e672b4d784a0da229882c4b/src/BloomBrowserUI/bookEdit/bookSettings/BookSettingsDialog.tsx#L605)):*
> ```XML
>                                         {
>                                             label: textPaddingDefaultLabel,
>                                             value: "" // use whatever the theme provides
> ```

The reason this works is complex enough to take me several tries to figure out. It works because topLevel-text-padding is declared in AppearanceSettings to be a CssStringVariableDef, and the GetCssVariableDeclaration override for this class returns an empty string (no CSS variable declaration at all) if the value is empty. Therefore no rule about this property gets put in the part of Appearance.css that is created by GetCssOwnPropsDeclaration, and the (earlier) rules from the theme win.
It feels like it should be documented somewhere that empty string is a magic value that doesn't set the variable to that value in Appearance.css (as every other value does) but leaves it unspecified, allowing the theme to win. It's also important that the constructor for this property in the AppearanceSettings.propertyDefinitions initializer has no default (unlike all the other properties that are handled in the dialog), which means this setting gets the theme setting by default.

Minimally, in the AppearanceSettings.propertyDefinitions initializer, topLevel-text-padding should be moved out of the list with the comment "these ones are not in the dialog yet" into the list at the start, and I think it would help to comment it "Deliberately has no default, which allows the theme to provide the default".

If you can't think of a better place to add something more, maybe spell out why CssStringVariableDef.GetCssVariableDeclaration returns an empty string  when the value is empty with something along the lines "When the value is empty, we don't generate a rule at all. Doing this rather than generating a rule with an empty value allows empty string to be a "magic" value for a property which will not override whatever the theme sets, since the part of Appearance.css that is generated from JSON using this method won't have a rule to win over earlier ones. See for example the Default value for topLevel-text-padding"

Maybe also worth noting in the comment on GetCssOwnPropsDeclaration that "If a property's value is set to an empty string we don't generate a rule for it, so the value in the current theme or the default theme can win"

hatton (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6664)* status: all files reviewed, 1 unresolved discussion (waiting on @JohnThomson)

___
*[`src/BloomBrowserUI/bookEdit/bookSettings/BookSettingsDialog.tsx` line 605 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6664#-O70g2Pl9o37WUf1aibB:-O70pOBW4UxmtBwVPnLx:b-8nvf4c) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/9ea8adfbbdb02f641e672b4d784a0da229882c4b/src/BloomBrowserUI/bookEdit/bookSettings/BookSettingsDialog.tsx#L605)):*

The reason this works is complex enough to take me several tries to figure out. It works because topLevel-text-padding is declared in AppearanceSettings to be a CssStringVariableDef, and the GetCssVariableDeclaration override for this class returns an empty string (no CSS variable declaration at all) if the value is empty. Therefore no rule about this property gets put in the part of Appearance.css that is created by GetCssOwnPropsDeclaration, and the (earlier) rules from the theme win.
It feels like it should be documented somewhere that empty string is a magic value that doesn't set the variable to that value in Appearance.css (as every other value does) but leaves it unspecified, allowing the theme to win. It's also important that the constructor for this property in the AppearanceSettings.propertyDefinitions initializer has no default (unlike all the other properties that are handled in the dialog), which means this setting gets the theme setting by default.

Minimally, in the AppearanceSettings.propertyDefinitions initializer, topLevel-text-padding should be moved out of the list with the comment "these ones are not in the dialog yet" into the list at the start, and I think it would help to comment it "Deliberately has no default, which allows the theme to provide the default".

If you can't think of a better place to add something more, maybe spell out why CssStringVariableDef.GetCssVariableDeclaration returns an empty string  when the value is empty with something along the lines "When the value is empty, we don't generate a rule at all. Doing this rather than generating a rule with an empty value allows empty string to be a "magic" value for a property which will not override whatever the theme sets, since the part of Appearance.css that is generated from JSON using this method won't have a rule to win over earlier ones. See for example the Default value for topLevel-text-padding"

Maybe also worth noting in the comment on GetCssOwnPropsDeclaration that "If a property's value is set to an empty string we don't generate a rule for it, so the value in the current theme or the default theme can win"
</blockquote></details>

Thanks @JohnThomson. Would you be able to take this one forward?

PR #6673, opened 2024-09-19
Allow pasting link to cover page or to another book (BL-13880)

StephenMcConnel:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6673)

### DIFF
--- a/src/BloomBrowserUI/bookEdit/bloomField/BloomField.ts
+++ b/src/BloomBrowserUI/bookEdit/bloomField/BloomField.ts
@@ -316,24 +316,32 @@ export default class BloomField {
                     if (!result.data) {
                         return; // More sanity checks are in bloomEditing.updateCkEditorButtonStatus
                     }
-                    const anchor = document.createElement("a");
-                    anchor.href = result.data;
-                    try {
-                        document
-                            .getSelection()
-                            ?.getRangeAt(0)
-                            ?.surroundContents(anchor);
-                    } catch (ex) {
-                        const englishErrorMessage =
-                            "Bloom was not able to make a link. Try selecting only simple text.";
-                        console.log(`${englishErrorMessage} ${ex}`);
-                        BloomMessageBoxSupport.CreateAndShowSimpleMessageBox(
-                            "EditTab.HyperlinkPasteFailure",
-                            englishErrorMessage,
-                            "Shows when a hyperlink cannot be pasted due to invalid selection.",
-                            "CantPasteHyperlink"
+                    get("app/selectedBookInfo", bookInfo => {
+                        if (!bookInfo.data) {
+                            return;
+                        }
+                        const anchor = document.createElement("a");
+                        anchor.href = processPastedHyperlink(
+                            result.data,
+                            bookInfo.data.id
                         );
-                    }
+                        try {
+                            document
+                                .getSelection()
+                                ?.getRangeAt(0)
+                                ?.surroundContents(anchor);
+                        } catch (ex) {
+                            const englishErrorMessage =
+                                "Bloom was not able to make a link. Try selecting only simple text.";
+                            console.log(`${englishErrorMessage} ${ex}`);
+                            BloomMessageBoxSupport.CreateAndShowSimpleMessageBox(
+                                "EditTab.HyperlinkPasteFailure",
+                                englishErrorMessage,
+                                "Shows when a hyperlink cannot be pasted due to invalid selection.",
+                                "CantPasteHyperlink"
+                            );
+                        }
+                    });
                 });
                 return true; // probaby means success, but I'm not sure. Typescript says this function has to return a boolean.
             }
@@ -948,4 +956,17 @@ interface FFSelection extends Selection {
 }
 interface JQuery {
     reverse(): JQuery;
+}
+function processPastedHyperlink(dataPaste, dataBookId): string {
+    const url = dataPaste as string;
+    const bookId = dataBookId as string;
+    if (!url || !bookId) {
+        return "";
+    }
+    const currentBookPrefix = `bloomnav://book/${bookId}?page=`;
+    if (url.startsWith(currentBookPrefix)) {
+        return "#" + url.substring(currentBookPrefix.length);
+    } else {
+        return url;
+    }
 }
--- a/src/BloomExe/Book/Page.cs
+++ b/src/BloomExe/Book/Page.cs
@@ -21,6 +21,7 @@ namespace Bloom.Book
         bool IsBackMatter { get; }
         bool IsXMatter { get; }
         bool IsCoverPage { get; }
+        bool IsFrontCoverPage { get; }
         string GetCaptionOrPageNumber(ref int pageNumber, out string captionI18nId);
         int GetIndex();
         string IdOfFirstAncestor { get; }
@@ -137,6 +138,11 @@ namespace Bloom.Book
         public bool IsCoverPage
         {
             get { return XMatterHelper.IsCoverPage(_getDivNodeForThisPageMethod(this)); }
+        }
+
+        public bool IsFrontCoverPage
+        {
+            get { return XMatterHelper.IsFrontCoverPage(_getDivNodeForThisPageMethod(this)); }
         }
 
         public string GetCaptionOrPageNumber(ref int pageNumber, out string captionI18nId)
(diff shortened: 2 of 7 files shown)


andrew-polk (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6673)* status: all files reviewed, 5 unresolved discussions (waiting on @StephenMcConnel)

___
*[`src/BloomExe/Edit/EditingModel.cs` line 714 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6673#-O7VnWUt5HIvMH8rSidL:-O7VnWUu-HKT6wKIURZ0:b-71xq0w) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/7b6793c592f5f3a832e043054d7e30818eedeaa4/src/BloomExe/Edit/EditingModel.cs#L714)):*
> ```Smalltalk
>         // a different strategy for identifying them. And then, choosing a different xmatter pack might cause a page
>         // to cease to exist altogether.
>         // The only exception is the front cover page, which uses a magic value instead of the page id.
> ```

Is it only the outside front cover? Because I think your IsFrontCover methods will return true for the inside front cover.
Instead, XMatterHelper should look for the class 'outsideFrontCover' (and I think the method names should be changed).
___
*[`src/BloomExe/Publish/BloomPub/BloomPubMaker.cs` line 578 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6673#-O7VrD828hD4vhOwr-wx:-O7VrD828hD4vhOwr-wy:bcbk2ck) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/7b6793c592f5f3a832e043054d7e30818eedeaa4/src/BloomExe/Publish/BloomPub/BloomPubMaker.cs#L578)):*
> ```Smalltalk
>             var coverPage = book.RawDom.SelectSingleNode(
>                 "//div[@id and contains(@class, 'frontCover')]"
>             );
> ```

Is it possible to use the XMatterHelper here?
If not, is it possible to have both share the same check? Maybe it isn't worth it. 

I think this should be outsideFrontCover also?
___
*[`src/BloomExe/Publish/BloomPub/BloomPubMaker.cs` line 584 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6673#-O7VsqqW8_fWO0J9EzDa:-O7VsqqW8_fWO0J9EzDb:b-o6yiqg) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/7b6793c592f5f3a832e043054d7e30818eedeaa4/src/BloomExe/Publish/BloomPub/BloomPubMaker.cs#L584)):*
> ```Smalltalk
>             foreach (var anchor in coverLinks.Cast<SafeXmlElement>())
>             {
>                 anchor.SetAttribute("href", "#" + coverPageId);
> ```

I thought we were persisting the book with the bloomnav protocol, but this seems to suggest otherwise. Or I'm just not understand?
___
*[`src/BloomExe/web/controllers/CommonApi.cs` line 114 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6673#-O7VruMq4oQG8jT4KsyR:-O7VruMq4oQG8jT4KsyS:bo46biz) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/7b6793c592f5f3a832e043054d7e30818eedeaa4/src/BloomExe/web/controllers/CommonApi.cs#L114)):*
> ```Smalltalk
>                 },
>                 false
>             );
> ```

Can we use AppApi's `selectedBookInfo` endpoint instead?
___
*[`src/BloomExe/web/controllers/EditingViewApi.cs` line 282 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6673#-O7VsQAx0J-7MinrAUU_:-O7VsQAx0J-7MinrAUUa:b5u0kgx) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/7b6793c592f5f3a832e043054d7e30818eedeaa4/src/BloomExe/web/controllers/EditingViewApi.cs#L282)):*
> ```Smalltalk
>                 {
>                     return pageId == "cover"
>                         || book.GetPages().Any(page => page.Id == pageId && !page.IsXMatter);
> ```

I think it is helpful to maintain the comment about why we can't link to (most) xmatter pages (i.e. the ids change).

StephenMcConnel (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6673)* status: all files reviewed, 5 unresolved discussions (waiting on @andrew-polk)

___
*[`src/BloomExe/Edit/EditingModel.cs` line 714 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6673#-O7VnWUt5HIvMH8rSidL:-O7VzjH--SMwXWlCQoZq:b-buiijp) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/44e9b93f0d0ec02693e19d14a50908a939df1682/src/BloomExe/Edit/EditingModel.cs#L714)):*

Is it only the outside front cover? Because I think your IsFrontCover methods will return true for the inside front cover.
Instead, XMatterHelper should look for the class 'outsideFrontCover' (and I think the method names should be changed).
</blockquote></details>

If you examine a Bloom book, the inside front cover doesn't have the frontCover class.  But it would probably be safer to use the outsideFrontCover class just to perfectly clear.
___
*[`src/BloomExe/Publish/BloomPub/BloomPubMaker.cs` line 578 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6673#-O7VrD828hD4vhOwr-wx:-O7W2iy0BrvAVwr3GuCb:b-e18rky) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/44e9b93f0d0ec02693e19d14a50908a939df1682/src/BloomExe/Publish/BloomPub/BloomPubMaker.cs#L578)):*

Is it possible to use the XMatterHelper here?
If not, is it possible to have both share the same check? Maybe it isn't worth it. 

I think this should be outsideFrontCover also?
</blockquote></details>

I don't think it's worth trying to force this into the existing XMatterHelper code.
But switching to outsideFrontCover for clarity (and possible safety) makes sense.
___
*[`src/BloomExe/Publish/BloomPub/BloomPubMaker.cs` line 584 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6673#-O7VsqqW8_fWO0J9EzDa:-O7W-e4m0B6PecVV99tF:buaoc45) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/44e9b93f0d0ec02693e19d14a50908a939df1682/src/BloomExe/Publish/BloomPub/BloomPubMaker.cs#L584)):*

I thought we were persisting the book with the bloomnav protocol, but this seems to suggest otherwise. Or I'm just not understand?
</blockquote></details>

We persist external links with the bloomnav protocol, but internal links are persisted with the normal HTML shorthand for internal links (a reference starting with the # character).  This preserves the current page jumping behavior (with the new feature of being able to jump to the beginning / front cover of the book) before any container code is written to handle the bloomnav protocol.
___
*[`src/BloomExe/web/controllers/CommonApi.cs` line 114 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6673#-O7VruMq4oQG8jT4KsyR:-O7W3vCOBqq7RUnpWadf:bwgcv0u) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/44e9b93f0d0ec02693e19d14a50908a939df1682/src/BloomExe/web/controllers/CommonApi.cs#L114)):*

Can we use AppApi's `selectedBookInfo` endpoint instead?
</blockquote></details>

I suppose we could, although it's rather overkill for this purpose.
___
*[`src/BloomExe/web/controllers/EditingViewApi.cs` line 282 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6673#-O7VsQAx0J-7MinrAUU_:-O7W69T806Un1vK_annf:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/44e9b93f0d0ec02693e19d14a50908a939df1682/src/BloomExe/web/controllers/EditingViewApi.cs#L282)):*

I think it is helpful to maintain the comment about why we can't link to (most) xmatter pages (i.e. the ids change).
</blockquote></details>

Done.

andrew-polk (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6673)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @StephenMcConnel)

PR #6694, opened 2024-10-08
Fix various TC rename issues and allow multiple local TCs

JohnThomson:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6694)

### DIFF
--- a/src/BloomExe/Book/BookStorage.cs
+++ b/src/BloomExe/Book/BookStorage.cs
@@ -1668,10 +1668,10 @@ namespace Bloom.Book
             return relocatableDom;
         }
 
-        private static bool ShouldWeChangeFolderName(
-            string sanitizedFilename,
+        private static string GetActualPathToSave(
+            string idealFolderName,
             string currentFolderName,
-            string idealFolderName
+            string idealFolderPath
         )
         {
             // As of 16 Dec 2019 we changed our definition of "sanitized" to include some more characters that can
@@ -1680,21 +1680,17 @@ namespace Bloom.Book
             // we go ahead and do the rename.
             // The various cases of sanitized names and folders and existing book names (e.g. Foo1) are tricky
             // to tease apart. We can't just let it go, though, since the new higher level of sanitization is
-            // necessary to keep bloom-player from throwing an error.
+            // necessary to keep bloom-player from throwing an error. However, idealFolderName is
+            // sanitized by the new system, so it can't be equal to currentFolderName if the latter has a problem.
 
-            // 1. If the current folder name needs sanitizing, we definitely can't use it; keep going.
-            if (currentFolderName != SanitizeNameForFileSystem(currentFolderName))
-                return true;
-            // 2. If the name we are already using is the name we want, keep using it (don't change anything)
-            if (currentFolderName == sanitizedFilename)
-                return false;
-            // 3. If our most prefered sanitized filename works (doesn't exist), go ahead and change to it; keep going.
-            if (!Directory.Exists(idealFolderName))
-                return true;
-            // 4. If our current folder name is one of the possible work-arounds (because preferred name is in use),
-            //    then return early and don't change. (We don't need to generate another alternative.)
-            // 5. Otherwise change to some other (sanitized) name, ensuring availability.
-            return !currentFolderName.StartsWith(sanitizedFilename);
+            // 1. If the name we are already using is the name we want, keep using it
+            if (currentFolderName.ToLowerInvariant() == idealFolderName.ToLowerInvariant())
+                return idealFolderPath; // might be a change of case, we will just change it.
+            // 2. If our most preferred sanitized filename works (doesn't exist), go ahead and change to it.
+            if (!Directory.Exists(idealFolderPath))
+                return idealFolderPath;
+            // 3. ideal name is in use, so find a variant that isn't.
+            return GetUniqueFolderPath(idealFolderPath);
         }
 
         public void SetBookName(string name)
@@ -1720,12 +1716,10 @@ namespace Bloom.Book
 
             var currentFilePath = PathToExistingHtml;
             var currentFolderName = Path.GetFileNameWithoutExtension(currentFilePath);
-            var idealFolderName = Path.Combine(Directory.GetParent(FolderPath).FullName, name);
-            if (!ShouldWeChangeFolderName(name, currentFolderName, idealFolderName))
-                return;
-
-            // Figure out what name we're really going to use (might need to add a number suffix).
-            idealFolderName = GetUniqueFolderPath(idealFolderName);
+            var idealFolderPath = Path.Combine(Directory.GetParent(FolderPath).FullName, name);
+            var actualSavePath = GetActualPathToSave(name, currentFolderName, idealFolderPath);
+            if (actualSavePath == Path.GetDirectoryName(currentFilePath))
+                return; // for this path they must be exactly the same, even by case.
 
             // Next, rename the file
             Guard.Against(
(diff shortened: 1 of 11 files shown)


andrew-polk (commented):
I haven't gotten to review this yet, but I notice there are no unit test additions/updates. Is any of this testable?

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6694#-:-O8iMi5o3ev2AzM6CZFf:b-5xcveu)* status: 0 of 8 files reviewed, all discussions resolved

JohnThomson (commented):
I added one for the bit that I think is most testable (and in a class that is already much tested).

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6694#-:-O8ibxMl10W49WQMPrXz:btuj4s5)* status: 0 of 11 files reviewed, all discussions resolved

andrew-polk:
We decided we want to retarget this on master to test it there first before deciding if it should go into 6.1.

andrew-polk (commented):
> Reviewed 1 of 8 files at r1, 7 of 7 files at r2, 10 of 10 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6694)* status: all files reviewed, 4 unresolved discussions (waiting on @JohnThomson)

___
*[`src/BloomExe/Book/BookStorage.cs` line 1671 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6694#-O8wLxVGC2u7isvHh9Cu:-O8wLxVGC2u7isvHh9Cv:b-q4igqv) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/e0583826655b5b3557f0df78ddb7e4303b3d406f/src/BloomExe/Book/BookStorage.cs#L1671)):*
> ```Smalltalk
>         }
> 
>         private static string GetActualPathToSave(
> ```

If I'm seeing this correctly, you fixed this method and then made a change to stop using it.
Is it worth keeping? If so, should we have a caveat comment stating we changed it since it was last used, so it hasn't been tested in its current state?
___
*[`src/BloomExe/Collection/BookCollection.cs` line 562 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6694#-O8wLVSc42nYdueuzLpu:-O8wLVSc42nYdueuzLpv:b-y5g0g6) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/e0583826655b5b3557f0df78ddb7e4303b3d406f/src/BloomExe/Collection/BookCollection.cs#L562)):*
> ```Smalltalk
>         {
>             return GetBookInfos()
>                 .FirstOrDefault(b => b.FolderPath.ToLowerInvariant() == path.ToLowerInvariant());
> ```

Are there future implications here for when we get back to Linux and Mac?
Same for the change in CollectionModel.
___
*[`src/BloomExe/TeamCollection/FolderTeamCollection.cs` line 417 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6694#-O8mCcUEDMWFxbK3Eaja:-O8mCcUEDMWFxbK3Eajb:bw6asyq) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/e0583826655b5b3557f0df78ddb7e4303b3d406f/src/BloomExe/TeamCollection/FolderTeamCollection.cs#L417)):*
> ```Smalltalk
>                 && !string.IsNullOrEmpty(realLocalFolderName)
>                 && realRepoName != realLocalFolderName
>             );
> ```

I was surprised that a method which claims to check if things are the same but for case doesn't do a case-insensitive comparison. But rather it just checks if they are different in any way.
Seems like it would be better to rename the method to match what it is actually doing even if the use case we have right now is for case difference.

Similar for FixPossibleCaseChange which could be something like `EnsureConsistentRepoName`
___
*[`src/BloomExe/TeamCollection/TeamCollection.cs` line 1346 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6694#-O8wKWn--_JmlmBiedct:-O8wKWn--_JmlmBiedcu:bywqb2j) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/e0583826655b5b3557f0df78ddb7e4303b3d406f/src/BloomExe/TeamCollection/TeamCollection.cs#L1346)):*
> ```Smalltalk
>             return localStatus.checksum != GetStatus(bookName).checksum
>                 || (
>                     DoLocalAndRemoteNamesDifferByCase(bookName) && !IsCheckedOutHereBy(localStatus)
> ```

Based on the comment, it seems like the check here really wants to be that it different ONLY by case. But as I stated in the comment on that method, I don't think that is strictly what the method checks.

JohnThomson (commented):
Done

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6694#-:-O8x0Qyr-q2_TIANjk87:bb6un1j)* status: 4 of 11 files reviewed, 4 unresolved discussions (waiting on @andrew-polk)

___
*[`src/BloomExe/Book/BookStorage.cs` line 1671 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6694#-O8wLxVGC2u7isvHh9Cu:-O8x0e1-1Jd30SZ-RjIw:bam3d3m) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/e0583826655b5b3557f0df78ddb7e4303b3d406f/src/BloomExe/Book/BookStorage.cs#L1671)):*

If I'm seeing this correctly, you fixed this method and then made a change to stop using it.
Is it worth keeping? If so, should we have a caveat comment stating we changed it since it was last used, so it hasn't been tested in its current state?
</blockquote></details>

No, it's used in SetBookName. Things started getting very complicated until I realized that ShouldWeChangeFolderName was only used by that one routine, and making it instead responsible for the entire job of coming up with the name we want simplified things greatly. But I'm not surprised that it's hard to see the net effect of the change. It's now possible that we DO want to do some renaming even in the (change case) situation where the destination book DOES exist (just in the wrong case). So this routine considers all factors and comes up with what the name should be, and SetBookName changes things if is different in any way at all.
Maybe there's somewhere more of that could be captured in comments.
___
*[`src/BloomExe/Collection/BookCollection.cs` line 562 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6694#-O8wLVSc42nYdueuzLpu:-O8x1vtXE4oHbR6KdGIk:b-w5g5u0) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/e0583826655b5b3557f0df78ddb7e4303b3d406f/src/BloomExe/Collection/BookCollection.cs#L562)):*

Are there future implications here for when we get back to Linux and Mac?
Same for the change in CollectionModel.
</blockquote></details>

Don't think it will be a problem. I commented.
___
*[`src/BloomExe/TeamCollection/FolderTeamCollection.cs` line 417 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6694#-O8mCcUEDMWFxbK3Eaja:-O8x4eH0599wNAytq-Xz:byrk1f8) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/e0583826655b5b3557f0df78ddb7e4303b3d406f/src/BloomExe/TeamCollection/FolderTeamCollection.cs#L417)):*

I was surprised that a method which claims to check if things are the same but for case doesn't do a case-insensitive comparison. But rather it just checks if they are different in any way.
Seems like it would be better to rename the method to match what it is actually doing even if the use case we have right now is for case difference.

Similar for FixPossibleCaseChange which could be something like `EnsureConsistentRepoName`
</blockquote></details>

They can't differ in any way except by case or the attempt to find a file and folder by that exact name (no * in the filters) would fail.
I went with EnsureConsistentCasingInLocalName, reflecting that (a) case is the only inconsistency it will try to fix, and (b) it's the local name that we're going to change to make it consistent.
___
*[`src/BloomExe/TeamCollection/TeamCollection.cs` line 1346 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6694#-O8wKWn--_JmlmBiedct:-O8x6CDH3GwiEx41ISan:b-w6gywz) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/e0583826655b5b3557f0df78ddb7e4303b3d406f/src/BloomExe/TeamCollection/TeamCollection.cs#L1346)):*

Based on the comment, it seems like the check here really wants to be that it different ONLY by case. But as I stated in the comment on that method, I don't think that is strictly what the method checks.
</blockquote></details>

Again, they wouldn't be the same book at all if the names differed by more than case (unless there's a bigger rename going on , in which case the test above for _remotelyRenamedBooks.Contains(bookName) should apply). I renamed it DoLocalAndRemoteNamesDifferOnlyByCase as well as commenting why that really is what it does. Hope that helps.

andrew-polk (commented):
> Reviewed 7 of 7 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6694)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @JohnThomson)

___
*[`src/BloomExe/Book/BookStorage.cs` line 1671 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6694#-O8wLxVGC2u7isvHh9Cu:-O8xJMZOCvuyrxFRjD1j:b7w9seu) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/e0583826655b5b3557f0df78ddb7e4303b3d406f/src/BloomExe/Book/BookStorage.cs#L1671)):*

No, it's used in SetBookName. Things started getting very complicated until I realized that ShouldWeChangeFolderName was only used by that one routine, and making it instead responsible for the entire job of coming up with the name we want simplified things greatly. But I'm not surprised that it's hard to see the net effect of the change. It's now possible that we DO want to do some renaming even in the (change case) situation where the destination book DOES exist (just in the wrong case). So this routine considers all factors and comes up with what the name should be, and SetBookName changes things if is different in any way at all.
Maybe there's somewhere more of that could be captured in comments.
</blockquote></details>

Sorry; I'm not sure what I was seeing. I must have been searching for the old name.

It seems pretty clear now.
___
*[`src/BloomExe/TeamCollection/FolderTeamCollection.cs` line 417 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6694#-O8mCcUEDMWFxbK3Eaja:-O8xIvmS6yW76GOzrWpw:b-7aitaz) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/e0583826655b5b3557f0df78ddb7e4303b3d406f/src/BloomExe/TeamCollection/FolderTeamCollection.cs#L417)):*

They can't differ in any way except by case or the attempt to find a file and folder by that exact name (no * in the filters) would fail.
I went with EnsureConsistentCasingInLocalName, reflecting that (a) case is the only inconsistency it will try to fix, and (b) it's the local name that we're going to change to make it consistent.
</blockquote></details>

Thanks for clarifying.
___
*[`src/BloomExe/TeamCollection/TeamCollection.cs` line 1346 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6694#-O8wKWn--_JmlmBiedct:-O8xJCvmBaYgqrBLv-0C:b-k614gh) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/e0583826655b5b3557f0df78ddb7e4303b3d406f/src/BloomExe/TeamCollection/TeamCollection.cs#L1346)):*

Again, they wouldn't be the same book at all if the names differed by more than case (unless there's a bigger rename going on , in which case the test above for _remotelyRenamedBooks.Contains(bookName) should apply). I renamed it DoLocalAndRemoteNamesDifferOnlyByCase as well as commenting why that really is what it does. Hope that helps.
</blockquote></details>

Yes, thanks.

PR #6707, opened 2024-10-14
Improvements to video controls, mainly for Game Play mode (BL-13957)

JohnThomson:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6707)

### DIFF
--- a/src/BloomBrowserUI/bookEdit/css/editMode.less
+++ b/src/BloomBrowserUI/bookEdit/css/editMode.less
@@ -1538,3 +1538,17 @@ svg.bloom-videoControl {
         left: calc(50% - @videoIconSize / 2);
     }
 }
+
+// ...except in Game Play mode. This is tricky. It needs to override the rule above, but not
+// override the rule that makes Play visible when we are paused. We could put a :not(.drag-activity-play)
+// on the rule above, but I'm worried about the performance of a rule with a top-level negation.
+// We could use !important here, but we really need at least the .drag-activity-play and
+// the .bloom-videoContainer:not(.paused) or the !important will hide it when paused.
+// (This rule would be more at home in Games.less, but we don't need any other Game-specific
+// rules about the video controls, so I think it's more helpful to keep it with the rules
+// it modifies.)
+.drag-activity-play .bloom-videoContainer:not(.paused):not(.playing):hover {
+    .bloom-videoPlayIcon {
+        display: none;
+    }
+}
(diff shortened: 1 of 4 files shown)


andrew-polk (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6707)* status: all files reviewed, 1 unresolved discussion (waiting on @JohnThomson)

___
*[`src/BloomBrowserUI/bookEdit/js/bloomVideo.ts` line 439 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6707#-O9C3uPt1HFi2-kno0QJ:-O9C3uPt1HFi2-kno0QK:b-mci09e) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/4490c32ce90f6595238b61ce62ca0a259ffaaa14/src/BloomBrowserUI/bookEdit/js/bloomVideo.ts#L439)):*
> ```XML
>     const video = ev.currentTarget as HTMLVideoElement;
> 
>     // If we're not in Play mode, we don't need these behaviors.
> ```

I don't understand this.
In edit mode, we only want to play the video if they click the button, not the video itself?
Or is this another case where that gets handled elsewhere?

JohnThomson (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6707)* status: 3 of 4 files reviewed, 1 unresolved discussion (waiting on @andrew-polk)

___
*[`src/BloomBrowserUI/bookEdit/js/bloomVideo.ts` line 439 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6707#-O9C3uPt1HFi2-kno0QJ:-O9CC0pa9R8seTMFnhiN:b-y68xrp) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/4490c32ce90f6595238b61ce62ca0a259ffaaa14/src/BloomBrowserUI/bookEdit/js/bloomVideo.ts#L439)):*

I don't understand this.
In edit mode, we only want to play the video if they click the button, not the video itself?
Or is this another case where that gets handled elsewhere?
</blockquote></details>

I don't know. The card says that in BP, a touch on the video should play it. In the desktop, hovering a video that is not playing should show a play button. There is some discussion that it should still be easy to drag the overlay, even if the drag starts on the button. It does not say explicitly that hovering in BP should NOT show an icon, or that touching the video in BE should NOT play it. I guessed. Seems the main purpose of clicking on something in design in editor is to move it, so maybe only the explicit Play button should play it? John says my guesses (or something else) are so obviously awful that anyone could see they are wrong, even if they don't know Bloom; I'm not even in the ballpark. I've fixed the things I think are obviously wrong but it's quite likely I've completely missed the most obvious and important ones.
I added a comment.

andrew-polk (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6707)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @JohnThomson)

PR #6708, opened 2024-10-15
Use nicescroll for bloom editing, not just reading (BL-13865)

StephenMcConnel:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6708)


andrew-polk (changes requested):
With this change, I thought there would also be the removing of some code somewhere which turns on the browser's scrollbar just for some specific scenario (Device Landscape)?

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6708#-:-O9H-TNFAUwcmYwkR5MB:bbu0h1m)* status: all files reviewed, 3 unresolved discussions (waiting on @StephenMcConnel)

___
*[`src/BloomBrowserUI/bookEdit/js/bloomEditing.ts` line 1665 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6708#-O9GyvaQ1PQj3llhNWWw:-O9GyvaQ1PQj3llhNWWx:bsi6c63) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/991f74cd54f86631cb459e00c3ebec4370b0338b/src/BloomBrowserUI/bookEdit/js/bloomEditing.ts#L1665)):*
> ```XML
> }
> 
> // This constant string was copied from bloom-player with minimal changes and should be in a common library.
> ```

I think it is worth moving all these nicescroll consts and functions into a separate file.
___
*[`src/BloomBrowserUI/bookEdit/js/bloomEditing.ts` line 1794 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6708#-O9Gzk8P4E7tjcs89G9D:-O9Gzk8P4E7tjcs89G9E:bgx9n53) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/991f74cd54f86631cb459e00c3ebec4370b0338b/src/BloomBrowserUI/bookEdit/js/bloomEditing.ts#L1794)):*
> ```XML
>                                 const group = entry.target.parentElement!
>                                     .parentElement!;
>                                 group.classList.remove(
> ```

In bloom-player, we can change the dom all we want because every time we load the book, we read it from the files again.

But here, I don't think we can do that. Won't this get persisted?
Same for other class changes.
___
*[`src/BloomBrowserUI/bookEdit/js/bloomEditing.ts` line 1821 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6708#-O9H-3sCCmUWl6zs3TxW:-O9H-3sCCmUWl6zs3TxX:b-sn1m10) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/991f74cd54f86631cb459e00c3ebec4370b0338b/src/BloomBrowserUI/bookEdit/js/bloomEditing.ts#L1821)):*
> ```XML
>                                     cursoropacitymax: 0.1,
>                                     cursorborderradius: "12px" // Make the corner more rounded than the 5px default.
>                                 });
> ```

It is good and right that we remove the niceScroll for each element in cleanupNiceScroll(), but it would great if we could somehow guarantee this doesn't get persisted some way. Usually we do that by putting the `bloom-ui` class on something. Would that be possible here somehow?

(or is the dom of the page itself not actually modified in this case?)

StephenMcConnel (commented):
nicescroll overrides such code by setting overflow:hidden on the element.  But the existing CSS/less should probably be removed since it's no longer being used and unused code that used to be used is never a good idea.

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6708#-:-O9LQ_mQ4SCVyiy0rcow:bhxzrxb)* status: 4 of 7 files reviewed, 3 unresolved discussions (waiting on @andrew-polk)

___
*[`src/BloomBrowserUI/bookEdit/js/bloomEditing.ts` line 1665 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6708#-O9GyvaQ1PQj3llhNWWw:-O9L4kVF5F0Y4g0c9_eY:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/991f74cd54f86631cb459e00c3ebec4370b0338b/src/BloomBrowserUI/bookEdit/js/bloomEditing.ts#L1665)):*

I think it is worth moving all these nicescroll consts and functions into a separate file.
</blockquote></details>

Done.
___
*[`src/BloomBrowserUI/bookEdit/js/bloomEditing.ts` line 1794 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6708#-O9Gzk8P4E7tjcs89G9D:-O9LQR1AEQ-pJzarE9xw:br2hfvo) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/991f74cd54f86631cb459e00c3ebec4370b0338b/src/BloomBrowserUI/bookEdit/js/bloomEditing.ts#L1794)):*

In bloom-player, we can change the dom all we want because every time we load the book, we read it from the files again.

But here, I don't think we can do that. Won't this get persisted?
Same for other class changes.
</blockquote></details>

I've added code here and in the cleanup method to handle the class changes to not persist the changes.
___
*[`src/BloomBrowserUI/bookEdit/js/bloomEditing.ts` line 1821 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6708#-O9H-3sCCmUWl6zs3TxW:-O9L5-be81Lh4W-p7OB4:b-lszku0) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/991f74cd54f86631cb459e00c3ebec4370b0338b/src/BloomBrowserUI/bookEdit/js/bloomEditing.ts#L1821)):*

It is good and right that we remove the niceScroll for each element in cleanupNiceScroll(), but it would great if we could somehow guarantee this doesn't get persisted some way. Usually we do that by putting the `bloom-ui` class on something. Would that be possible here somehow?

(or is the dom of the page itself not actually modified in this case?)
</blockquote></details>

The cleanup code prevents the elements inserted by nicescroll from being persisted by removing them from the dom.  It's called when the page changes, both entering and leaving.  Other than restoring the class lists modified by addScrollBarsToPage, this is the only operation needed to ensure nothing gets persisted.

andrew-polk (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6708)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @StephenMcConnel)

PR #6712, opened 2024-10-17
Repair loss of style with resulting loss of format cog (BL-13953)

StephenMcConnel:
The repair takes two forms: fix a broken book as best we can, and prevent breakage when changing the page format.

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6712)

### DIFF
--- a/src/BloomExe/Book/HtmlDom.cs
+++ b/src/BloomExe/Book/HtmlDom.cs
@@ -1377,6 +1377,9 @@ namespace Bloom.Book
             return classAttr.Split(' ').FirstOrDefault(x => x.EndsWith("-style"));
         }
 
+        /// <remarks>
+        /// Child may already have a style, but we want it to have the style used by the template.
+        /// </remarks>
         private static void FixStyle(
             SafeXmlNode child,
             string requiredClass,
@@ -1384,8 +1387,7 @@ namespace Bloom.Book
         )
         {
             if (
-                desiredStyleByLang.Count == 0
-                || !(child is SafeXmlElement) // can be a whitespace node (BL-13977)
+                !(child is SafeXmlElement) // can be a whitespace node (BL-13977)
                 || !child.GetAttribute("class").Contains(requiredClass)
             )
                 return;
@@ -1393,17 +1395,37 @@ namespace Bloom.Book
             var langAttr = child.GetAttribute("lang");
             if (langAttr == null || langAttr == "*" || langAttr == "z")
                 langAttr = defaultLangKey;
-            string defaultStyle;
-            desiredStyleByLang.TryGetValue(defaultLangKey, out defaultStyle); // if unsuccessful, 'defaultStyle' will be ""
             string newStyle;
             if (!desiredStyleByLang.TryGetValue(langAttr, out newStyle))
+            {
+                // If there is no default style, the copied text will not have a style, and the
+                // Styles gear icon will not show up in the text box.  So we need a default style.
+                // See BL-13977.
+                string defaultStyle = null;
+                // child has already been added to the translationGroup copied from the template.
+                var group =
+                    child.SelectSingleNode(
+                        "ancestor::div[contains(@class,'bloom-translationGroup')]"
+                    ) as SafeXmlElement;
+                if (group != null)
+                    defaultStyle = GetStyle(group);
+                if (string.IsNullOrEmpty(defaultStyle))
+                    desiredStyleByLang.TryGetValue(defaultLangKey, out defaultStyle); // if unsuccessful, 'defaultStyle' will be ""
+                if (string.IsNullOrEmpty(defaultStyle))
+                {
+                    defaultStyle = desiredStyleByLang.Values.FirstOrDefault(
+                        x => x != null && x.EndsWith("-style")
+                    );
+                    if (string.IsNullOrEmpty(defaultStyle))
+                        defaultStyle = string.IsNullOrEmpty(childStyle)
+                            ? "normal-style" // Normal is better than nothing.
+                            : childStyle; // Retain original style if template doesn't have one.
+                }
                 newStyle = defaultStyle;
-            string newclass;
-            if (childStyle != null)
-                newclass = child.GetAttribute("class").Replace(childStyle, newStyle);
-            else
-                newclass = child.GetAttribute("class") + " " + newStyle;
-            child.SetAttribute("class", newclass);
+            }
+            if (!string.IsNullOrEmpty(childStyle))
+                (child as SafeXmlElement).RemoveClass(childStyle);
+            (child as SafeXmlElement).AddClass(newStyle);
         }
 
         // relative to the DOM's Head element
--- a/src/BloomExe/Book/TranslationGroupManager.cs
+++ b/src/BloomExe/Book/TranslationGroupManager.cs
@@ -925,8 +925,12 @@ namespace Bloom.Book
                 where x.GetAttribute("lang") == langTag
                 select x;
             if (elementsAlreadyInThisLanguage.Any())
-                //don't mess with this set, it already has a vernacular (this will happen when we're editing a shellbook, not just using it to make a vernacular edition)
-                return elementsAlreadyInThisLanguage.First();
+            {
+                // don't create a new element for this language, it already has one.
+                var element = elementsAlreadyInThisLanguage.First();
+                EnsureStyleClassOnElement(groupElement, editableChildrenOfTheGroup, element);
+                return element;
+            }
 
             if (
                 groupElement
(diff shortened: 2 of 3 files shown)


StephenMcConnel:
This could possibly be a candidate for cherry-picking to 6.0.

andrew-polk (changes requested):
> Reviewed 2 of 3 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6712)* status: 2 of 3 files reviewed, 3 unresolved discussions (waiting on @StephenMcConnel)

___
*[`src/BloomExe/Book/Book.cs` line 1821 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6712#-O9kTKnj5KvQ296Mafx5:-O9kTKnj5KvQ296Mafx6:bvadinm) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/677ecdec037328b08f7fbfe911ce33d6103c4fb0/src/BloomExe/Book/Book.cs#L1821)):*
> ```Smalltalk
>             RepairBrokenSmallCoverCredits(OurHtmlDom);
>             RepairCoverImageDescriptions(OurHtmlDom);
>             RepairMissingStylesAndLangZInTranslationGroups(OurHtmlDom);
> ```

I'm pretty torn over the idea of looking for this situation every time we run EnsureUpToDate.

In theory, this is the kind of thing we should be using maintenance level migrations for so we only do it once. Of course, there could, in theory, be other situations which cause this which we haven't fixed....
But I'm not sure that theoretical case warrants this check for every book.

On the other hand, using a whole maintenance level for this fix seems a bit heavy handed.

Thoughts?

Do you have an opinion, @JohnThomson?
___
*[`src/BloomExe/Book/Book.cs` line 2307 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6712#-O9kOket29Ywid1yjC8N:-O9kOket29Ywid1yjC8O:blg9xgr) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/677ecdec037328b08f7fbfe911ce33d6103c4fb0/src/BloomExe/Book/Book.cs#L2307)):*
> ```Smalltalk
>                             langZExists = true;
>                         var classList = editableDiv.GetAttribute("class").Trim().Split(' ', '\t');
>                         var style = classList.FirstOrDefault(x => x.EndsWith("-style"));
> ```

HtmlDom has a GetStyle method which looks appropriate here.
___
*[`src/BloomExe/Book/Book.cs` line 2313 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6712#-O9kOy4E8uQ9JvPUnHZf:-O9kOy4E8uQ9JvPUnHZg:bivjk7f) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/677ecdec037328b08f7fbfe911ce33d6103c4fb0/src/BloomExe/Book/Book.cs#L2313)):*
> ```Smalltalk
>                                 "class",
>                                 $"{string.Join(" ", classList)} {styleForRepair}"
>                             );
> ```

HtmlDom has an AddClass method

andrew-polk (commented):
> Reviewed 1 of 3 files at r1.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6712)* status: all files reviewed, 3 unresolved discussions (waiting on @StephenMcConnel)

JohnThomson (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6712)* status: all files reviewed, 4 unresolved discussions (waiting on @andrew-polk and @StephenMcConnel)

___
*[`src/BloomExe/Book/Book.cs` line 1821 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6712#-O9kTKnj5KvQ296Mafx5:-O9kdJEO6SZ9nU9vO48M:b-lhwjbx) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/677ecdec037328b08f7fbfe911ce33d6103c4fb0/src/BloomExe/Book/Book.cs#L1821)):*

I'm pretty torn over the idea of looking for this situation every time we run EnsureUpToDate.

In theory, this is the kind of thing we should be using maintenance level migrations for so we only do it once. Of course, there could, in theory, be other situations which cause this which we haven't fixed....
But I'm not sure that theoretical case warrants this check for every book.

On the other hand, using a whole maintenance level for this fix seems a bit heavy handed.

Thoughts?

Do you have an opinion, @JohnThomson?
</blockquote></details>

I think I would be more inclined to fix this in TranslationGroupManager.MakeElementWithLanguageForOneGroup. That could pretty easily repair an existing editable. If it has to make a new one and the prototype doesn't have a style class, it can repair the new one.
I think the cost of doing it once per editable on a page would be negligible (especially since we already found where it needs doing, so we don't even add a SelectNode), whereas the cost for doing it for a whole book might be nontrivial. And, like this solution, it will be in place if we ever again make a mistake that leaves some languages without a style class.
Of course that would not be a great solution if we actually need to fix this across the whole book (e.g., to publish a broken one).
If it's not feasible I think this is worth a maintenance level. Numbers don't cost anything, and making one more number comparison is a whole lot cheaper than searching the whole book.
___
*[`src/BloomExe/Book/Book.cs` line 2320 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6712#-O9kgCmE7CW_G-qMABgL:-O9kgCmE7CW_G-qMABgM:b-cqqy1e) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/677ecdec037328b08f7fbfe911ce33d6103c4fb0/src/BloomExe/Book/Book.cs#L2320)):*
> ```Smalltalk
>                         else
>                         {
>                             styleForRepair = style; // we expect the style to stay the same inside a translationGroup.
> ```

We should hunt for this possibility in a separate loop, unless there is reason to be quite sure that a good block that already has a style will come before any that does not.
Also, check the classes of the TG. Sometimes that has the style, too.

StephenMcConnel (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6712)* status: 0 of 5 files reviewed, 4 unresolved discussions (waiting on @andrew-polk and @JohnThomson)

___
*[`src/BloomExe/Book/Book.cs` line 1821 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6712#-O9kTKnj5KvQ296Mafx5:-O9upe5H8kJ5PpErGBht:btix9ay) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/677ecdec037328b08f7fbfe911ce33d6103c4fb0/src/BloomExe/Book/Book.cs#L1821)):*

I think I would be more inclined to fix this in TranslationGroupManager.MakeElementWithLanguageForOneGroup. That could pretty easily repair an existing editable. If it has to make a new one and the prototype doesn't have a style class, it can repair the new one.
I think the cost of doing it once per editable on a page would be negligible (especially since we already found where it needs doing, so we don't even add a SelectNode), whereas the cost for doing it for a whole book might be nontrivial. And, like this solution, it will be in place if we ever again make a mistake that leaves some languages without a style class.
Of course that would not be a great solution if we actually need to fix this across the whole book (e.g., to publish a broken one).
If it's not feasible I think this is worth a maintenance level. Numbers don't cost anything, and making one more number comparison is a whole lot cheaper than searching the whole book.
</blockquote></details>

I've moved the repair code to TranslationGroupManager, and adjusted the unit tests accordingly.
___
*[`src/BloomExe/Book/Book.cs` line 2307 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6712#-O9kOket29Ywid1yjC8N:-O9upctW0ZeGPsLP4cgq:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/677ecdec037328b08f7fbfe911ce33d6103c4fb0/src/BloomExe/Book/Book.cs#L2307)):*

HtmlDom has a GetStyle method which looks appropriate here.
</blockquote></details>

Done.
___
*[`src/BloomExe/Book/Book.cs` line 2313 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6712#-O9kOy4E8uQ9JvPUnHZf:-O9uqQMt8cdx7gBEg0Cm:b-iz5aox) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/677ecdec037328b08f7fbfe911ce33d6103c4fb0/src/BloomExe/Book/Book.cs#L2313)):*

HtmlDom has an AddClass method
</blockquote></details>

The method has shifted to SafeXmlElement, but I changed to using it.
___
*[`src/BloomExe/Book/Book.cs` line 2320 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6712#-O9kgCmE7CW_G-qMABgL:-O9uuSD21LfqsJ0UB9_g:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/677ecdec037328b08f7fbfe911ce33d6103c4fb0/src/BloomExe/Book/Book.cs#L2320)):*

We should hunt for this possibility in a separate loop, unless there is reason to be quite sure that a good block that already has a style will come before any that does not.
Also, check the classes of the TG. Sometimes that has the style, too.
</blockquote></details>

Done.

JohnThomson (commented):
LGTM. Waiting for Andrew before merging.

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6712#-:-O9ymrTZAmC2cAwQVRZK:bpux9h8)* status: all files reviewed, 4 unresolved discussions (waiting on @andrew-polk)

andrew-polk (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6712)* status: all files reviewed, 1 unresolved discussion (waiting on @JohnThomson)

andrew-polk (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6712)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @StephenMcConnel)

PR #6714, opened 2024-10-18
Books with overlays migrate to legacy (BL-14007)

JohnThomson:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6714)


andrew-polk (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6714)* status: all files reviewed, 5 unresolved discussions (waiting on @JohnThomson)

___
*[`src/BloomExe/Book/AppearanceSettings.cs` line 280 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6714#-O9WCOai3MM4oEN1xGaL:-O9WCOai3MM4oEN1xGaM:b2xdcdc) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/4e004fbdc9970631f83bf1b0fcc354e123529fe2/src/BloomExe/Book/AppearanceSettings.cs#L280)):*
> ```Smalltalk
>         CssThemeName = "default";
>         string substituteAppearance = null;
>         if (Program.RunningHarvesterMode || bookDom.HasOverlays())
> ```

This is probably easier to explain to me in person instead of typing, but I don't understand something.
If the book in question in the card got messed up simply because the book had already been in the appearance system once before being harvested from 5.6, why does adding this condition not essentially do the same thing?

I.e., If you open your book in 6.0 then 5.6 and it has overlays, why does the code now do something different than it did on harvester?
___
*[`src/BloomExe/Book/AppearanceSettings.cs` line 284 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6714#-O9WCEd-EIy5ae3C9cmD:-O9WCEd-EIy5ae3C9cmE:b-hjz3wv) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/4e004fbdc9970631f83bf1b0fcc354e123529fe2/src/BloomExe/Book/AppearanceSettings.cs#L284)):*
>             // This should only apply to temporary copies on their way to becoming artifacts,
>             // so it doesn't interfere with getting books migrated when they are to be edited.

This is no longer true and makes me wonder if this is safe.
___
*[`src/BloomExe/Book/Book.cs` line 5175 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6714#-O9WBbg09IMuaHVbjP5l:-O9WBbg09IMuaHVbjP5m:bj0r371) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/4e004fbdc9970631f83bf1b0fcc354e123529fe2/src/BloomExe/Book/Book.cs#L5175)):*
> ```Smalltalk
>         public bool HasActivities => OurHtmlDom.HasActivityPages();
> 
>         public bool HasOverlayPages => OurHtmlDom.HasComicalOverlays();
> ```

Seems like `HasOverlayPages` should probably be `HasComicalOverlayPages`.
Otherwise there is a mismatch between Book.HasOverlayPages and HtmlDom.HasOverlays.
___
*[`src/BloomExe/Book/HtmlDom.cs` line 2793 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6714#-O9WD5kX5ZfhzUuOY-Gj:-O9WD5kX5ZfhzUuOY-Gk:bgewvi2) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/4e004fbdc9970631f83bf1b0fcc354e123529fe2/src/BloomExe/Book/HtmlDom.cs#L2793)):*
> ```Smalltalk
>             return _dom.SelectSingleNode("//div[contains(@class, 'bloom-textOverPicture')]")
>                 != null;
>         }
> ```

Why wasn't the existing one correct for the case we are concerned with in this card?
We can't need to worry about positioning of elements if there is no svg, right?
___
*[`src/BloomTests/Book/AppearanceSettingsTests.cs` line 347 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6714#-O9WDUXfDpglfGHaHr79:-O9WDUXfDpglfGHaHr7A:b-coprzx) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/4e004fbdc9970631f83bf1b0fcc354e123529fe2/src/BloomTests/Book/AppearanceSettingsTests.cs#L347)):*
> ```Smalltalk
> 				<div class='bloom-imageContainer'>
> 				</div>
> 			</div></body></html>"
> ```

Somehow, you've gotten a mix of indentation types. CSharpier should have prevented that.

andrew-polk (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6714)* status: all files reviewed, 3 unresolved discussions (waiting on @JohnThomson)

___
*[`src/BloomExe/Book/AppearanceSettings.cs` line 280 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6714#-O9WCOai3MM4oEN1xGaL:-O9WMSm3AS_gZ_LnTKY6:b-pqntfg) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/4e004fbdc9970631f83bf1b0fcc354e123529fe2/src/BloomExe/Book/AppearanceSettings.cs#L280)):*

This is probably easier to explain to me in person instead of typing, but I don't understand something.
If the book in question in the card got messed up simply because the book had already been in the appearance system once before being harvested from 5.6, why does adding this condition not essentially do the same thing?

I.e., If you open your book in 6.0 then 5.6 and it has overlays, why does the code now do something different than it did on harvester?
</blockquote></details>

We discussed.
It is true that in a case where the book is already in the appearance system, this will not help.
___
*[`src/BloomExe/Book/AppearanceSettings.cs` line 284 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6714#-O9WCEd-EIy5ae3C9cmD:-O9WN8Ut0dCwy07X-EHP:b6d0pgy) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/4e004fbdc9970631f83bf1b0fcc354e123529fe2/src/BloomExe/Book/AppearanceSettings.cs#L284)):*

>             // This should only apply to temporary copies on their way to becoming artifacts,
>             // so it doesn't interfere with getting books migrated when they are to be edited.

This is no longer true and makes me wonder if this is safe.
</blockquote></details>

Discussed.
Comment should just be updated.
___
*[`src/BloomTests/Book/AppearanceSettingsTests.cs` line 347 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6714#-O9WDUXfDpglfGHaHr79:-O9WO3fG68iSVPi_oYFm:b5x8jlx) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/4e004fbdc9970631f83bf1b0fcc354e123529fe2/src/BloomTests/Book/AppearanceSettingsTests.cs#L347)):*

Somehow, you've gotten a mix of indentation types. CSharpier should have prevented that.
</blockquote></details>

I see now these are @"" strings.
So CSharpier probably ignores them.
I think you can ignore.

andrew-polk (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6714)* status: all files reviewed, 2 unresolved discussions (waiting on @JohnThomson)

___
*[`src/BloomExe/Book/HtmlDom.cs` line 2793 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6714#-O9WD5kX5ZfhzUuOY-Gj:-O9WOGZUEU1SBbWP5g_j:bff7vgv) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/4e004fbdc9970631f83bf1b0fcc354e123529fe2/src/BloomExe/Book/HtmlDom.cs#L2793)):*

Why wasn't the existing one correct for the case we are concerned with in this card?
We can't need to worry about positioning of elements if there is no svg, right?
</blockquote></details>

Discussed.
The text-only ones don't have an svg.

JohnThomson (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6714)* status: all files reviewed, 2 unresolved discussions (waiting on @andrew-polk)

___
*[`src/BloomExe/Book/AppearanceSettings.cs` line 284 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6714#-O9WCEd-EIy5ae3C9cmD:-O9WbcnI4j1xuztsixFN:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/4e004fbdc9970631f83bf1b0fcc354e123529fe2/src/BloomExe/Book/AppearanceSettings.cs#L284)):*

Discussed.
Comment should just be updated.
</blockquote></details>

Done.
___
*[`src/BloomExe/Book/Book.cs` line 5175 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6714#-O9WBbg09IMuaHVbjP5l:-O9Wbnds-lh_sL7m9iBZ:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/4e004fbdc9970631f83bf1b0fcc354e123529fe2/src/BloomExe/Book/Book.cs#L5175)):*

Seems like `HasOverlayPages` should probably be `HasComicalOverlayPages`.
Otherwise there is a mismatch between Book.HasOverlayPages and HtmlDom.HasOverlays.
</blockquote></details>

Done.

andrew-polk (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6714)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @JohnThomson)

PR #6724, opened 2024-10-24
Snap cropping when close to nothing (BL-13952)

JohnThomson:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6724)

### DIFF
--- a/src/BloomBrowserUI/bookEdit/js/bubbleManager.ts
+++ b/src/BloomBrowserUI/bookEdit/js/bubbleManager.ts
@@ -1534,6 +1534,17 @@ export class BubbleManager {
     private initialCropBubbleHeight: number;
     private initialCropBubbleTop: number;
     private initialCropBubbleLeft: number;
+    // If we're dragging a crop control, we generally want to snap when the edege
+    // of the (underlying, uncropped) image is close to the corresponding edge
+    // of the overlay in which it is cropped...that is, no cropping on that edge,
+    // nor have we (this cycle) expanded the image by dragging the crop handle outward.
+    // However, if the drag started in the crop position we disable cropping so small
+    // adjustments can be made. If the pointer moves more than the snap distance,
+    // we resume cropping. (Cropping can also be disabled by holding down the ctrl key).
+    // This variable is true when we are in that state where cropping is disabled
+    // because we've made only a small movement from an uncropped state. It is
+    // independent of the ctrl key state (though irrelevant if it is down).
+    private cropSnapDisabled: boolean = false;
 
     private currentDragSide: string | undefined;
     // For both resize and crop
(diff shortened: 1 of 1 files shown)


andrew-polk (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6724)* status: all files reviewed, 1 unresolved discussion (waiting on @JohnThomson)

___
*[`src/BloomBrowserUI/bookEdit/js/bubbleManager.ts` line 1601 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6724#-OA3u-135iyQyeQxZYCK:-OA3u-135iyQyeQxZYCL:bdrzy82) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/0776e8de7b9f17e560d2cc24489d030801df0aa2/src/BloomBrowserUI/bookEdit/js/bubbleManager.ts#L1601)):*
> ```XML
>                     break;
>                 case "e":
>                     // LHS is where the right of the image is. RHS is where the right of the overlay is.
> ```

Sorry; I don't understand these comments.
I don't see LHS or RHS except in these comments, but it seems like the comment is supposed to be telling me what they mean.

I wonder if it is worth extracting this logic to a function called `shouldCropSnapBeDisabled` and giving it a comment explaining, generally, why we want it disabled vs enabled.
But on second thought, maybe that implies too much (namely that you've encapsulated all such logic in one place) since you have lots more logic below about when it should be disabled/enabled.

JohnThomson (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6724)* status: all files reviewed, 1 unresolved discussion (waiting on @andrew-polk)

___
*[`src/BloomBrowserUI/bookEdit/js/bubbleManager.ts` line 1601 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6724#-OA3u-135iyQyeQxZYCK:-OA4SFTLCKm_Ohqx0a_x:bv5ydjq) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/0776e8de7b9f17e560d2cc24489d030801df0aa2/src/BloomBrowserUI/bookEdit/js/bubbleManager.ts#L1601)):*

Sorry; I don't understand these comments.
I don't see LHS or RHS except in these comments, but it seems like the comment is supposed to be telling me what they mean.

I wonder if it is worth extracting this logic to a function called `shouldCropSnapBeDisabled` and giving it a comment explaining, generally, why we want it disabled vs enabled.
But on second thought, maybe that implies too much (namely that you've encapsulated all such logic in one place) since you have lots more logic below about when it should be disabled/enabled.
</blockquote></details>

I think when I wrote that I was comparing a left-hand-side expression with a right-hand-side. Then Math.abs got involved, and then prettier, and now it's not two sides but five or six lines. I reworded and added some more comments.
I don't think the nature of what John asked for lends itself to handling crop snapping all in one place.

andrew-polk (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6724)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @JohnThomson)

PR #6729, opened 2024-10-29
Refactor FontMetadata for use by harvester font backup (BL-10035)

StephenMcConnel:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6729)

### DIFF
--- a/src/BloomExe/FontProcessing/FontMetadata.cs
+++ b/src/BloomExe/FontProcessing/FontMetadata.cs
@@ -48,7 +48,7 @@ namespace Bloom.FontProcessing
         public const string kInvalid = "invalid"; // bad file format (eg, .ttc)
 
         /// <summary>
-        /// On Window, we can use System.Windows.Media (which provides the GlyphTypeface class) to
+        /// On Windows, we can use System.Windows.Media (which provides the GlyphTypeface class) to
         /// provide all the font metadata information.
         /// On Linux, we have to use Sharpfont from nuget (which provides the Sharpfont.Face class)
         /// for reading the font's embedding flag plus running /usr/bin/otfino for everything else.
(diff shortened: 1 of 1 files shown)


andrew-polk (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6729)* status: all files reviewed, 1 unresolved discussion (waiting on @StephenMcConnel)

___
*[`src/BloomExe/FontProcessing/FontMetadata.cs` line 62 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6729#-OAcvR_W3JriURkABkSe:-OAcvR_W3JriURkABkSf:b-qyxk4g) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/74e5fe61b01071fdd83b51038cc251809f39f5b9/src/BloomExe/FontProcessing/FontMetadata.cs#L62)):*
Is the setting of variants purposefully only in this constructor because the other one doesn't need to do it so you're saving cycles?

Probably worth a comment on the other one to warn any caller that it won't set up the variants. Something like
> /// Note that for performance reasons this constructor does not initialize `variants`.

StephenMcConnel (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6729)* status: all files reviewed, 1 unresolved discussion (waiting on @andrew-polk)

___
*[`src/BloomExe/FontProcessing/FontMetadata.cs` line 62 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6729#-OAcvR_W3JriURkABkSe:-OArxi3f3JuiC_KUZUBY:b-srh5ir) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/74e5fe61b01071fdd83b51038cc251809f39f5b9/src/BloomExe/FontProcessing/FontMetadata.cs#L62)):*

Is the setting of variants purposefully only in this constructor because the other one doesn't need to do it so you're saving cycles?

Probably worth a comment on the other one to warn any caller that it won't set up the variants. Something like
> /// Note that for performance reasons this constructor does not initialize `variants`.
</blockquote></details>

Done.  Although I gave a fuller explanation.

andrew-polk (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6729)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @StephenMcConnel)

PR #6730, opened 2024-10-29
Another fix to prevent overflow calc from scrolling text box (BL-13942)

andrew-polk:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6730)

### DIFF
--- a/src/BloomBrowserUI/bookEdit/OverflowChecker/OverflowChecker.ts
+++ b/src/BloomBrowserUI/bookEdit/OverflowChecker/OverflowChecker.ts
@@ -160,8 +160,10 @@ export default class OverflowChecker {
             0
         );
 
-        // So the scroll height can be calculated without the language label affecting the height/width...
-        element.classList.add("hideLanguageLabel");
+        // Add a class so that the scroll height can be calculated without the language label affecting the height/width.
+        // (Currently, only done so for text-over-picture elements because the language label is OUTSIDE the box,
+        // but in a normal text box it is inside the box.)
+        element.classList.add("hideTOPLanguageLabel");
 
         const overflowY =
             this.contentHeight(element) -
@@ -173,59 +175,64 @@ export default class OverflowChecker {
             element.scrollWidth -
             (element.clientWidth + focusedBorderFudgeFactor);
 
-        element.classList.remove("hideLanguageLabel");
+        element.classList.remove("hideTOPLanguageLabel");
 
         return [overflowX, overflowY];
     }
 
     private static contentHeight(element: HTMLElement): number {
-        let maxContentBottom = 0;
-        const children = Array.from(element.children);
-        // First, we need to make sure that the size and position of the children are not affected
-        // by flex grow or shrink. Possibly at some point we might need to remember the old values
-        // of these styles, but currently we only mess with them using stylesheets, so setting them
-        // back to empty at the end is sufficient.
-        // Conceivably, there could be some flicker from setting and clearing these, but I haven't
-        // noticed it. Typically we use the results of this function to make overlays the right size,
-        // and then grow/shrink don't do anything.
         if (element.childNodes.length === 0) {
             return 0;
         }
-        // Note that there may be no children, even though there are childNodes. We can only set style
-        // for elements, but we need to measure all the nodes (at least for one unit test).
-        children.forEach((e: HTMLElement) => {
-            e.style.flexGrow = "0";
-            e.style.flexShrink = "0";
-        });
-        // The element itself must not shrink, lest its scrollHeight or clientHeight be inaccurate.
-        element.style.flexGrow = "0";
-        element.style.flexShrink = "0";
-        // This is a reliable way to get the information if it is greater than the clientHeight,
-        // and means we don't have to worry about scroll position.
-        let result = element.scrollHeight;
-        if (element.scrollHeight <= element.clientHeight) {
-            // But if scrollHeight is less than or equal to clientHeight, we use our own algorithm.
-            //
-            element.childNodes.forEach((x: HTMLElement) => {
-                if (!(x instanceof HTMLElement)) return; // not an element; I think this is redundant
-                if (window.getComputedStyle(x).position === "absolute") return; // special element like format button
-                // Not sure if offsetTop can ever be negative, now that I prevented flexShrink.
-                // But in case it is, we don't want to shrink the content height. And often there is only one
-                // child so we'll get an accurate figure this way even if there is some scrolling.
-                const xbottom = Math.max(x.offsetTop, 0) + x.offsetHeight;
-                if (xbottom > maxContentBottom) {
-                    maxContentBottom = xbottom;
-                }
+
+        // Use a temporary clone so we don't modify the actual element.
+        // Otherwise, the text box can get scrolled back to the top when the user types. See BL-13942.
+        const elementClone = element.cloneNode(true) as HTMLElement;
+        elementClone.style.visibility = "hidden";
+        elementClone.style.position = "absolute"; // prevent any inadvertent layout effects
+        element.parentElement!.appendChild(elementClone);
+
+        let result = 0;
+        try {
+            let maxContentBottom = 0;
+
+            // First, we need to make sure that the size and position of the children are not affected
+            // by flex grow or shrink. Currently, there is no reason to set these back because we are
+            // operating on a temporary clone.
+
+            // Note that there may be no children, even though there are childNodes. We can only set style
+            // for elements, but we need to measure all the nodes (at least for one unit test).
+            const children = Array.from(elementClone.children);
+            children.forEach((e: HTMLElement) => {
+                e.style.flexGrow = "0";
+                e.style.flexShrink = "0";
             });
-            result = maxContentBottom;
+            // The element itself must not shrink, lest its scrollHeight or clientHeight be inaccurate.
+            elementClone.style.flexGrow = "0";
+            elementClone.style.flexShrink = "0";
+            // This is a reliable way to get the information if it is greater than the clientHeight,
+            // and means we don't have to worry about scroll position.
+            result = elementClone.scrollHeight;
+            if (elementClone.scrollHeight <= elementClone.clientHeight) {
+                // But if scrollHeight is less than or equal to clientHeight, we use our own algorithm.
+                elementClone.childNodes.forEach((x: HTMLElement) => {
+                    if (!(x instanceof HTMLElement)) return; // not an element; I think this is redundant
+                    if (window.getComputedStyle(x).position === "absolute")
+                        return; // special element like format button
+                    // Not sure if offsetTop can ever be negative, now that I prevented flexShrink.
+                    // But in case it is, we don't want to shrink the content height. And often there is only one
+                    // child so we'll get an accurate figure this way even if there is some scrolling.
+                    const xbottom = Math.max(x.offsetTop, 0) + x.offsetHeight;
+                    if (xbottom > maxContentBottom) {
+                        maxContentBottom = xbottom;
+                    }
+                });
+                result = maxContentBottom;
+            }
+        } finally {
+            // Remove the temporary clone we added.
+            element.parentElement!.removeChild(elementClone);
         }
-        // return control to stylesheet
-        children.forEach((e: HTMLElement) => {
-            e.style.flexGrow = "";
-            e.style.flexShrink = "";
-        });
-        element.style.flexGrow = "";
-        element.style.flexShrink = "";
         return result;
     }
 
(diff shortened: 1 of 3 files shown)


JohnThomson (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6730)* status: 0 of 3 files reviewed, 5 unresolved discussions (waiting on @andrew-polk)

___
*[`src/BloomBrowserUI/bookEdit/OverflowChecker/OverflowChecker.ts` line 165 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6730#-OAS_q1o6wZ_PZRmVfNe:-OAS_q1o6wZ_PZRmVfNf:bhigsf6) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/a5f5393ce714d13bbb805bd4edc7d4deb806cc23/src/BloomBrowserUI/bookEdit/OverflowChecker/OverflowChecker.ts#L165)):*
> ```XML
>         // Add a class so that the scroll height can be calculated without the language label affecting the height/width.
>         // (Currently, only done so for text-over-picture elements because the language label is OUTSIDE the box,
>         // but in a normal text box it is inside the box.)
> ```

I find this baffling, so I'm probably missing something. In TOP boxes the language label is now part of the context controls, and I don't think it affects the layout of the TOP box at all. It's not even a child of the TOP, let alone inside it. And why would we only need to hide it to calculate an accurate size when it is OUTSIDE the box (which to me suggests that it would already NOT be part of the calculation of the size of the box)?
Also puzzling: I don't see anything to make it so (that is, I don't see any test here that would prevent us adding it to non-TOP boxes.
___
*[`src/BloomBrowserUI/bookEdit/OverflowChecker/OverflowChecker.ts` line 187 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6730#-OASbcoJAti0VOERiNrL:-OASbcoJAti0VOERiNrM:b-mr4tx8) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/a5f5393ce714d13bbb805bd4edc7d4deb806cc23/src/BloomBrowserUI/bookEdit/OverflowChecker/OverflowChecker.ts#L187)):*
> ```XML
>         // Otherwise, the text box can get scrolled back to the top when the user types. See BL-13942.
>         const elementClone = element.cloneNode(true) as HTMLElement;
>         elementClone.style.visibility = "hidden";
> ```

This worries me. A hidden node takes up space in its parent. Things changing size tend to cause scroll bars to appear and disappear and scroll positions to change, which is what we're trying to prevent. It needs to be a child of the same parent so all relevant CSS rules apply, but could it be position:absolute so it doesn't disturb the layout?

Is it automatically the case that elementClone's scrollPosition is zero? If not, setting that might help get a reliable result.
___
*[`src/BloomBrowserUI/bookEdit/OverflowChecker/OverflowChecker.ts` line 195 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6730#-OASb3pF0eLpI957mg7y:-OASb3pF0eLpI957mg7z:bp8bl9r) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/a5f5393ce714d13bbb805bd4edc7d4deb806cc23/src/BloomBrowserUI/bookEdit/OverflowChecker/OverflowChecker.ts#L195)):*
> ```XML
>             const children = Array.from(elementClone.children);
>             // First, we need to make sure that the size and position of the children are not affected
>             // by flex grow or shrink. Currently, there is no reason to set these back because we
> ```

append "are"
___
*[`src/BloomBrowserUI/bookEdit/OverflowChecker/OverflowChecker.ts` line 199 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6730#-OASbGWk5QiQcDoxbqZb:-OASbGWk5QiQcDoxbqZc:b-pbslji) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/a5f5393ce714d13bbb805bd4edc7d4deb806cc23/src/BloomBrowserUI/bookEdit/OverflowChecker/OverflowChecker.ts#L199)):*
> ```XML
>             if (elementClone.childNodes.length === 0) {
>                 return 0;
>             }
> ```

I don't suppose it makes much difference, but this test and early return could be done before making the clone.
___
*[`src/content/bookLayout/bubble.less` line 149 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6730#-OASdoxECdVisfWh3SFT:-OASdoxECdVisfWh3SFU:b-nlie24) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/a5f5393ce714d13bbb805bd4edc7d4deb806cc23/src/content/bookLayout/bubble.less#L149)):*
> ```Less
>             // places (and is bug prone, if we change the other visibility rules).
>             // We absolutely want these hidden during overflow calculations!
>             .bloom-editable.hideTOPLanguageLabel {
> ```

I'm not convinced this is an improvement. There's certainly nothing about the rule that would prevent it applying to non-TOP boxes, and as noted above, I actually think it's only the NON-TOP boxes where we need this now.

andrew-polk (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6730)* status: 0 of 3 files reviewed, 5 unresolved discussions (waiting on @JohnThomson)

___
*[`src/BloomBrowserUI/bookEdit/OverflowChecker/OverflowChecker.ts` line 165 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6730#-OAS_q1o6wZ_PZRmVfNe:-OAcnXyL-sAzihlzRXH8:b-4opmpc) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/a5f5393ce714d13bbb805bd4edc7d4deb806cc23/src/BloomBrowserUI/bookEdit/OverflowChecker/OverflowChecker.ts#L165)):*
See comment on the .less change below.

I also don't understand why we thought that a label outside the box would affect the calc while one inside wouldn't. But that's clearly what this comment (which I'm simply reinstating) states.
The original comment and code before my previous PR was:
```
        // Adds a class so that the scroll height can be calculated without text-over-picture element controls affecting the height/width
        // (Currently, only done so for text-over-picture elements because the format gear and language tip are OUTSIDE the box,
        // but in a normal text box they are inside the box)
        element.classList.add("disableTOPControls");
```

I'm hesitant to change things in 6.0 I don't fully understand unless I have to.

> Also puzzling: I don't see anything to make it so (that is, I don't see any test here that would prevent us adding it to non-TOP boxes.

You're right that we add the class unconditionally, but the less rule only applies to TOP boxes.
___
*[`src/BloomBrowserUI/bookEdit/OverflowChecker/OverflowChecker.ts` line 187 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6730#-OASbcoJAti0VOERiNrL:-OAcsORJ3Q2tFjABigQC:b-u1bq27) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/a5f5393ce714d13bbb805bd4edc7d4deb806cc23/src/BloomBrowserUI/bookEdit/OverflowChecker/OverflowChecker.ts#L187)):*

This worries me. A hidden node takes up space in its parent. Things changing size tend to cause scroll bars to appear and disappear and scroll positions to change, which is what we're trying to prevent. It needs to be a child of the same parent so all relevant CSS rules apply, but could it be position:absolute so it doesn't disturb the layout?

Is it automatically the case that elementClone's scrollPosition is zero? If not, setting that might help get a reliable result.
</blockquote></details>

Setting it to absolute doesn't seem to have any bad effects.

It does appear that elementClone's scrollTop is always 0 even when elements is > 0.
___
*[`src/BloomBrowserUI/bookEdit/OverflowChecker/OverflowChecker.ts` line 195 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6730#-OASb3pF0eLpI957mg7y:-OAconVM1l-w_-hAxxCF:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/a5f5393ce714d13bbb805bd4edc7d4deb806cc23/src/BloomBrowserUI/bookEdit/OverflowChecker/OverflowChecker.ts#L195)):*

append "are"
</blockquote></details>

Done.
___
*[`src/BloomBrowserUI/bookEdit/OverflowChecker/OverflowChecker.ts` line 199 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6730#-OASbGWk5QiQcDoxbqZb:-OAcoogr56hC3dtzWMxx:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/a5f5393ce714d13bbb805bd4edc7d4deb806cc23/src/BloomBrowserUI/bookEdit/OverflowChecker/OverflowChecker.ts#L199)):*

I don't suppose it makes much difference, but this test and early return could be done before making the clone.
</blockquote></details>

Done.
___
*[`src/content/bookLayout/bubble.less` line 149 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6730#-OASdoxECdVisfWh3SFT:-OAcmq41-oGfKcmuTY-A:boin5o7) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/a5f5393ce714d13bbb805bd4edc7d4deb806cc23/src/content/bookLayout/bubble.less#L149)):*

I'm not convinced this is an improvement. There's certainly nothing about the rule that would prevent it applying to non-TOP boxes, and as noted above, I actually think it's only the NON-TOP boxes where we need this now.
</blockquote></details>

In my last PR, I changed this from `disableTOPControls` to `hideLanguageLabel` because I thought it was applying to all text boxes. But then I realized that because of the hierarchy here, it does only apply to TOP boxes (see 50-60 lines above).

So I added back in the TOP part of the label and restored much of the comment I had removed in the code where we call it.

andrew-polk:
*[`src/BloomBrowserUI/bookEdit/OverflowChecker/OverflowChecker.ts` line 165 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6730#-OAS_q1o6wZ_PZRmVfNe:-OActrup2r3L6b2K3Z_x:bau01uy) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/a5f5393ce714d13bbb805bd4edc7d4deb806cc23/src/BloomBrowserUI/bookEdit/OverflowChecker/OverflowChecker.ts#L165)):*

See comment on the .less change below.

I also don't understand why we thought that a label outside the box would affect the calc while one inside wouldn't. But that's clearly what this comment (which I'm simply reinstating) states.
The original comment and code before my previous PR was:
```
        // Adds a class so that the scroll height can be calculated without text-over-picture element controls affecting the height/width
        // (Currently, only done so for text-over-picture elements because the format gear and language tip are OUTSIDE the box,
        // but in a normal text box they are inside the box)
        element.classList.add("disableTOPControls");
```

I'm hesitant to change things in 6.0 I don't fully understand unless I have to.

> Also puzzling: I don't see anything to make it so (that is, I don't see any test here that would prevent us adding it to non-TOP boxes.

You're right that we add the class unconditionally, but the less rule only applies to TOP boxes.
</blockquote></details>

I'm happy to discuss any of this.

JohnThomson (approved):
> Reviewed 2 of 3 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6730)* status: all files reviewed, 2 unresolved discussions (waiting on @andrew-polk)

___
*[`src/BloomBrowserUI/bookEdit/OverflowChecker/OverflowChecker.ts` line 165 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6730#-OAS_q1o6wZ_PZRmVfNe:-OAdDALCBckiar1jxawa:b-gf7o4y) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/a5f5393ce714d13bbb805bd4edc7d4deb806cc23/src/BloomBrowserUI/bookEdit/OverflowChecker/OverflowChecker.ts#L165)):*

I'm happy to discuss any of this.
</blockquote></details>

I think the label is always absolutely positioned. In the old code, when it was absolutely positioned just above the bottom or just below the top of an ordinary box, comparing its bottom with the bottom of the parent would never cause us to think it overflowing. On the other hand, when it is absolutely positioned below the bottom, including it will ALWAYS make us think the box is overflowing.
Now that it's not a child of this box at all except when it's safely inside, I think we could get rid of this code and the class. But it may be safer to save that for 6.2.
In fact, I forgot this was targeting 6.0. That probably does still have a child label outside the editable. So it should stay for 6.0 for sure.
___
*[`src/content/bookLayout/bubble.less` line 149 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6730#-OASdoxECdVisfWh3SFT:-OAdEi1CE5H6P6BwdXQ6:bove19l) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/a5f5393ce714d13bbb805bd4edc7d4deb806cc23/src/content/bookLayout/bubble.less#L149)):*

In my last PR, I changed this from `disableTOPControls` to `hideLanguageLabel` because I thought it was applying to all text boxes. But then I realized that because of the hierarchy here, it does only apply to TOP boxes (see 50-60 lines above).

So I added back in the TOP part of the label and restored much of the comment I had removed in the code where we call it.
</blockquote></details>

I'm fairly sure we can just remove it. But not until at least 6.1!

PR #6731, opened 2024-10-30
Handle asynchronous paste for bubble height adjustment (BL-14004)

StephenMcConnel:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6731)

### DIFF
--- a/src/BloomBrowserUI/bookEdit/js/OverlayContextControls.tsx
+++ b/src/BloomBrowserUI/bookEdit/js/OverlayContextControls.tsx
@@ -373,7 +373,6 @@ const OverlayContextControls: React.FunctionComponent<{
     const handlePasteClick = () => {
         // We don't actually know there's no image on the clipboard, but it's not relevant for a text box.
         pasteClipboard(false);
-        theOneBubbleManager.updateAutoHeight(); // update the bubble height (BL-14004)
     };
     const editable = props.overlay.getElementsByClassName(
         "bloom-editable bloom-visibility-code-on"
--- a/src/BloomBrowserUI/bookEdit/js/bloomEditing.ts
+++ b/src/BloomBrowserUI/bookEdit/js/bloomEditing.ts
@@ -1459,6 +1459,8 @@ async function pasteImpl(imageAvailable: boolean) {
             editor.insertText(textToPaste);
             manager.unlock();
             manager.save(true);
+            // We need to update the bubble height (BL-14004).
+            bubbleManager.updateAutoHeight();
         }
         // It shouldn't happen that we don't have an editor, but if we don't, we just don't paste.
         // Otherwise, the paste is likely to go somewhere unexpected, wherever a ckeditor last had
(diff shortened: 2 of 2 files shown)


JohnThomson (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6731)* status: all files reviewed, 1 unresolved discussion (waiting on @StephenMcConnel)

___
*[`src/BloomBrowserUI/bookEdit/js/OverlayContextControls.tsx` line 377 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6731#-OAUGzFe4zp33jhESsnB:-OAUGzFf-jRU10yPfU0l:bwgccej) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/20e9ebbc7c0ba4ce2c20504159f22f538ddc950a/src/BloomBrowserUI/bookEdit/js/OverlayContextControls.tsx#L377)):*
> ```XML
>         // We need to update the bubble height (BL-14004).  pasteClipboard is actually asynchronous in
>         // its implementation, so we need to adjust the height in a callback.
>         pasteClipboard(false, () => theOneBubbleManager.updateAutoHeight());
> ```

This appears to be the only caller of pasteClipboard, and if there were others and we're pasting into a bubble, they probably want this adjustment too. Why not just move this call into the pasteImpl method wherever appropriate? (We don't need it when pasting images, for example.) pasteImpl is async, so it can use await wherever necessary.
Apart from future-proofing things, it just feels weird to do something tricky to make an async method use a callback like this. Sort of going back two generations in coding style: async code was first done with callbacks, then with Promises, now everything is moving to async. So there should be a very good reason to go back to callbacks.

StephenMcConnel (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6731)* status: all files reviewed, 1 unresolved discussion (waiting on @JohnThomson)

___
*[`src/BloomBrowserUI/bookEdit/js/OverlayContextControls.tsx` line 377 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6731#-OAUGzFe4zp33jhESsnB:-OAUMQw1AP_nyGWlT3SK:b-qddvs9) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/20e9ebbc7c0ba4ce2c20504159f22f538ddc950a/src/BloomBrowserUI/bookEdit/js/OverlayContextControls.tsx#L377)):*

This appears to be the only caller of pasteClipboard, and if there were others and we're pasting into a bubble, they probably want this adjustment too. Why not just move this call into the pasteImpl method wherever appropriate? (We don't need it when pasting images, for example.) pasteImpl is async, so it can use await wherever necessary.
Apart from future-proofing things, it just feels weird to do something tricky to make an async method use a callback like this. Sort of going back two generations in coding style: async code was first done with callbacks, then with Promises, now everything is moving to async. So there should be a very good reason to go back to callbacks.
</blockquote></details>

OK.  I thought maybe it was too hard to determine whether we're pasting into a bubble, but that doesn't appear to be the case.  And updating the height of the wrong bubble won't help matters, but won't hurt things either.

JohnThomson (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6731)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @StephenMcConnel)

PR #6732, opened 2024-10-30
Clean up PublishSettings from improper language tags (BL-14038)

StephenMcConnel:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6732)

### DIFF
--- a/src/BloomExe/Book/Book.cs
+++ b/src/BloomExe/Book/Book.cs
@@ -1217,7 +1217,54 @@ namespace Bloom.Book
                         }
                         badDiv.ParentNode.RemoveChild(badDiv);
                     }
+                    // Also clean up the publish settings (stored in a separate json file).
+                    // Copy the setting for the bad tag if the good tag does not have a setting.
+                    // (essentially the same approach used in cleaning up the div content above)
+                    RepairPublishLanguageSetting(
+                        BookInfo.PublishSettings.BloomLibrary.TextLangs,
+                        lang,
+                        tag
+                    );
+                    RepairPublishLanguageSetting(
+                        BookInfo.PublishSettings.BloomLibrary.AudioLangs,
+                        lang,
+                        tag
+                    );
+                    RepairPublishLanguageSetting(
+                        BookInfo.PublishSettings.BloomLibrary.SignLangs,
+                        lang,
+                        tag
+                    );
+                    RepairPublishLanguageSetting(
+                        BookInfo.PublishSettings.BloomPub.TextLangs,
+                        lang,
+                        tag
+                    );
+                    RepairPublishLanguageSetting(
+                        BookInfo.PublishSettings.BloomPub.AudioLangs,
+                        lang,
+                        tag
+                    );
+                    RepairPublishLanguageSetting(
+                        BookInfo.PublishSettings.BloomPub.SignLangs,
+                        lang,
+                        tag
+                    );
                 }
+            }
+        }
+
+        private void RepairPublishLanguageSetting(
+            Dictionary<string, InclusionSetting> langSettings,
+            string badTag,
+            string goodTag
+        )
+        {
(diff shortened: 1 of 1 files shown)


andrew-polk (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6732)* status: all files reviewed, 2 unresolved discussions (waiting on @StephenMcConnel)

___
*[`src/BloomExe/Book/Book.cs` line 1220 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6732#-OAUMFMk8dCcA757rRdL:-OAUMFMk8dCcA757rRdM:b47kv61) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/89e03133ce8f69034605d240cb081d9cded1d3cf/src/BloomExe/Book/Book.cs#L1220)):*
> ```Smalltalk
>                         badDiv.ParentNode.RemoveChild(badDiv);
>                     }
>                     // Also clean up the publish settings (stored in a separate json file)
> ```

Why don't we need to do anything with PublishSettings.BloomPUB?

Could you add a comment explaining why it is better to remove rather than replace the setting?
___
*[`src/BloomExe/Book/Book.cs` line 1222 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6732#-OAUMhYH3HXzgCmhostC:-OAUMhYH3HXzgCmhostD:b-ndp7wn) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/89e03133ce8f69034605d240cb081d9cded1d3cf/src/BloomExe/Book/Book.cs#L1222)):*
> ```Smalltalk
>                     // Also clean up the publish settings (stored in a separate json file)
>                     if (BookInfo.PublishSettings.BloomLibrary.TextLangs.ContainsKey(lang))
>                         BookInfo.PublishSettings.BloomLibrary.TextLangs.Remove(lang);
> ```

It looks like Remove can be called unconditionally.

StephenMcConnel (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6732)* status: all files reviewed, 2 unresolved discussions (waiting on @andrew-polk)

___
*[`src/BloomExe/Book/Book.cs` line 1220 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6732#-OAUMFMk8dCcA757rRdL:-OAUXDJB8Bffj_NXw-2N:b27izct) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/89e03133ce8f69034605d240cb081d9cded1d3cf/src/BloomExe/Book/Book.cs#L1220)):*

Why don't we need to do anything with PublishSettings.BloomPUB?

Could you add a comment explaining why it is better to remove rather than replace the setting?
</blockquote></details>

I simply forgot the BloomPUB set of settings.
I added code to copy the setting if it isn't already there for the good tag.
___
*[`src/BloomExe/Book/Book.cs` line 1222 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6732#-OAUMhYH3HXzgCmhostC:-OAUXPA7A6PlyMPNqAMA:b8jle6p) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/89e03133ce8f69034605d240cb081d9cded1d3cf/src/BloomExe/Book/Book.cs#L1222)):*

It looks like Remove can be called unconditionally.
</blockquote></details>

Noted.  But the new code for checking whether values should be transferred instead of simply removed already needs to do the check.

andrew-polk (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6732)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @StephenMcConnel)

PR #6742, opened 2024-11-07
Ensure overflow height is accurate (BL-14053)

StephenMcConnel:


andrew-polk (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6742)* status: all files reviewed (commit messages unreviewed), 1 unresolved discussion (waiting on @StephenMcConnel)

___
*[`src/BloomBrowserUI/bookEdit/OverflowChecker/OverflowChecker.ts` line 196 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6742#-OB7a53k9b7Tb9a-yWWb:-OB7a53k9b7Tb9a-yWWc:bfb0sgk) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/022e80c72c7d5c715d3998e27221515e3c3de9db/src/BloomBrowserUI/bookEdit/OverflowChecker/OverflowChecker.ts#L196)):*
> ```XML
>         if (element.scrollHeight > element.clientHeight) {
>             return element.scrollHeight;
>         }
> ```

The if condition and code about 25 lines lower:
`if (elementClone.scrollHeight <= elementClone.clientHeight) {`
assumes we can have less.
And also that we should be running the rest of the code outside of that instead of returning early if scrollHeight is greater.

So maybe your test case didn't need the flexGrow/Shrink changes?

andrew-polk (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6742)* status: all files reviewed, 1 unresolved discussion (waiting on @JohnThomson)

___
*[`src/BloomBrowserUI/bookEdit/OverflowChecker/OverflowChecker.ts` line 196 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6742#-OB7a53k9b7Tb9a-yWWb:-OBS0EgP0YkWlfViLzqA:b-wh3aoj) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/022e80c72c7d5c715d3998e27221515e3c3de9db/src/BloomBrowserUI/bookEdit/OverflowChecker/OverflowChecker.ts#L196)):*

https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollHeight says "If the element's content can fit without a need for vertical scrollbar, its `scrollHeight` is equal to [`clientHeight`](https://developer.mozilla.org/en-US/docs/Web/API/Element/clientHeight "clientHeight")."

I still think this code works, but I've made a different change which is probably necessary even with this bit of code.  The width of the cloned element does not take into account the padding applied to its parent, which made it wider (and taller) than the original element.  This messed up the height calculations.
</blockquote></details>

@JohnThomson 
I think you understand this much more than I do since this function was recently added by you (though modified by me). 
Will you take a look, please?

JohnThomson (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6742)* status: all files reviewed, 2 unresolved discussions (waiting on @StephenMcConnel)

___
*[`src/BloomBrowserUI/bookEdit/OverflowChecker/OverflowChecker.ts` line 202 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6742#-OBVyxFB2ks6V6tAJ9qo:-OBVyxFB2ks6V6tAJ9qp:b9r4hhv) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/f6c34b9278a0bf071e80ff772e6b71284cc4971e/src/BloomBrowserUI/bookEdit/OverflowChecker/OverflowChecker.ts#L202)):*
> ```XML
>         // the padding of the parent.)  See BL-14053.
>         elementClone.style.width = element.clientWidth + "px";
>         elementClone.style.height = element.clientHeight + "px";
> ```

The comment here is not quite working for me. What layout effects are we trying to prevent? An unintended change to the parent? How does setting a size on the child prevent that? Surely we WANT the hidden element to get laid out so we can measure it? Why do we need to take account of the padding of the parent? We're just trying to find out the height of the content of the element we cloned, in the context of the parent, but not distorted by any flex grow or shrink.
I could imagine that position: absolute causes the clone to ignore the width of the parent so we would need to set that, but I can't imagine why we'd set an explicit height.

StephenMcConnel (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6742)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @andrew-polk and @JohnThomson)

___
*[`src/BloomBrowserUI/bookEdit/OverflowChecker/OverflowChecker.ts` line 202 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6742#-OBVyxFB2ks6V6tAJ9qo:-OBbkRIf9G8BWERJAF0u:bpwu3po) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/f6c34b9278a0bf071e80ff772e6b71284cc4971e/src/BloomBrowserUI/bookEdit/OverflowChecker/OverflowChecker.ts#L202)):*

The comment here is not quite working for me. What layout effects are we trying to prevent? An unintended change to the parent? How does setting a size on the child prevent that? Surely we WANT the hidden element to get laid out so we can measure it? Why do we need to take account of the padding of the parent? We're just trying to find out the height of the content of the element we cloned, in the context of the parent, but not distorted by any flex grow or shrink.
I could imagine that position: absolute causes the clone to ignore the width of the parent so we would need to set that, but I can't imagine why we'd set an explicit height.
</blockquote></details>

I edited the comment extensively and removed setting the height, which is optional anyway.

JohnThomson (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6742#-:-OBf71TW9r7Pcil1VZVj:bnfp4nl)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @andrew-polk and @StephenMcConnel)

andrew-polk (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6742)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @StephenMcConnel)

PR #6744, opened 2024-11-07
Implement bloomnav links in bloompub preview (BL-13882)

StephenMcConnel:
This searches the other books in the same collection and creates a fast (not quite real) bloompub of the book pointed to in the bloomnav href.


StephenMcConnel:
This depends on https://github.com/BloomBooks/bloom-player/pull/337

JohnThomson (changes requested):
Nice work! I picked up a few minor things and raised a few questions, which you may already have thought about.

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6744#-:-OC5w0w0CHh1hBYM5Dh4:bsdkov3)* status: all files reviewed, 10 unresolved discussions (waiting on @StephenMcConnel)

___
*[`src/BloomExe/web/controllers/PublishApi.cs` line 423 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6744#-OC5sDQE67mYfez2M1dN:-OC5sDQE67mYfez2M1dO:b-3e4w86) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/187e799de2f09e6ccfa0f97019a26211096f57d2/src/BloomExe/web/controllers/PublishApi.cs#L423)):*
> ```Smalltalk
>             if (!string.IsNullOrEmpty(href))
>             {
>                 var bookId = href.Replace("bloomnav://book/", "");
> ```

I suppose for a GUID we don't need to worry about URLEncoding? This is a GUID, right, not a path? Might be worth a comment.
___
*[`src/BloomExe/web/controllers/PublishApi.cs` line 424 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6744#-OC5rxY6BkVnm3R-mapf:-OC5rxY6BkVnm3R-mapg:b-hd6kpq) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/187e799de2f09e6ccfa0f97019a26211096f57d2/src/BloomExe/web/controllers/PublishApi.cs#L424)):*
> ```Smalltalk
>             {
>                 var bookId = href.Replace("bloomnav://book/", "");
>                 var idx = bookId.IndexOf("?page=", StringComparison.Ordinal);
> ```

C# also has classes that understand params in URLs.
___
*[`src/BloomExe/web/controllers/PublishApi.cs` line 425 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6744#-OC5se4BEKdZCzQLYgwF:-OC5se4BEKdZCzQLYgwG:b-4bappp) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/187e799de2f09e6ccfa0f97019a26211096f57d2/src/BloomExe/web/controllers/PublishApi.cs#L425)):*
> ```Smalltalk
>                 var bookId = href.Replace("bloomnav://book/", "");
>                 var idx = bookId.IndexOf("?page=", StringComparison.Ordinal);
>                 string pageId = null;
> ```

Is this a decimal number or the page number the user sees? The latter might mean URLEncoding issues.

Oh, I see, it's a guid, right? Possibly worth commenting.
___
*[`src/BloomExe/web/controllers/PublishApi.cs` line 448 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6744#-OC5tE5FBHXVN4PHnU1s:-OC5tE5G2lyzSHfuaXXD:b-4law2u) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/187e799de2f09e6ccfa0f97019a26211096f57d2/src/BloomExe/web/controllers/PublishApi.cs#L448)):*
> ```Smalltalk
>                     var bookInfo = new Book.BookInfo(path, false);
>                     var book = _bookServer.GetBookFromBookInfo(bookInfo);
>                     var url = MakeFastFakeBloomPubForPreviewLink(
> ```

Even though this is somewhat optimized, it worries me that our server will be locked for so long. Also we could be deadlocked if ever the construction of the preview requires a response from our server. Would it be a problem to postpone the rest of this, say to an Idle event handler?
___
*[`src/BloomExe/web/controllers/PublishApi.cs` line 460 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6744#-OC5ug0ADUBUehiW_NsH:-OC5ug0B7ph7OXQ9U8-n:b-4xb6mk) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/187e799de2f09e6ccfa0f97019a26211096f57d2/src/BloomExe/web/controllers/PublishApi.cs#L460)):*
> ```Smalltalk
>                     var pageNumber = pageDiv?.GetAttribute("data-page-number");
>                     if (!string.IsNullOrEmpty(pageNumber))
>                         url += "&start-page=" + pageNumber; // & instead of ? simplifies javascript
> ```

Now we're talking about the attribute value, so again need to consider whether we store non-Roman page numbers in this attribute, or only decimal ones. Might need some URLEncoding? Or maybe not for websocket...anyway worth asking.
___
*[`src/BloomExe/web/controllers/PublishApi.cs` line 809 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6744#-OC5vH6h401qZLeAgEnH:-OC5vH6i3TkUnPKZWiX6:bi1xy4p) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/187e799de2f09e6ccfa0f97019a26211096f57d2/src/BloomExe/web/controllers/PublishApi.cs#L809)):*
> ```Smalltalk
>         /// This does no compressing or other really slow operations.  It does update the
>         /// book to use the Device XMatter which is needed to make it look like a real
>         /// bloompub book.
> ```

If this continues to happen inside a normal BloomServer handler, we should note here that it MUST NOT create any browsers that call back to our server, otherwise we'll get deadlocks. (The server has some special tricks that can be used to get around this if we have to...)
___
*[`src/BloomBrowserUI/utils/bloomNavHandler.ts` line 19 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6744#-OC5r38YErrnWuQD4p-i:-OC5r38YErrnWuQD4p-j:b-t8pygl) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/187e799de2f09e6ccfa0f97019a26211096f57d2/src/BloomBrowserUI/utils/bloomNavHandler.ts#L19)):*
> ```XML
> export function hookupBloomNavHandler() {
>     // Removing it first ensures that we only have one, even if we call this more than once.
>     window.removeEventListener("message", msghandler, { capture: true });
> ```

Not needed in JS
___
*[`src/BloomExe/Publish/Epub/EpubMaker.cs` line 609 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6744#-OC5rRDB1G2oflPfNF41:-OC5rRDB1G2oflPfNF42:b-6o3szj) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/187e799de2f09e6ccfa0f97019a26211096f57d2/src/BloomExe/Publish/Epub/EpubMaker.cs#L609)):*
> ```Smalltalk
>                 var msg = LocalizationManager.GetString(
>                     "PublishTab.Epub.LocalLinksProblem",
>                     "Links to other pages do not work in epubs. They will be changed to plain text that does not look like a link."
> ```

Should we tweak this to say "other pages or books"? Or is it not worth the cost of re-localizing?
___
*[`src/BloomBrowserUI/publish/ReaderPublish/ReaderPublishScreen.tsx` line 153 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6744#-OC5qLy74j9IaArkrFXo:-OC5qLy74j9IaArkrFXp:b-xavn63) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/187e799de2f09e6ccfa0f97019a26211096f57d2/src/BloomBrowserUI/publish/ReaderPublish/ReaderPublishScreen.tsx#L153)):*
> ```XML
>         baseUrl = baseUrl.substring(0, idxStart);
>         startPage = bookUrl.substring(idxStart);
>     }
> ```

I think JS has a URL class that understands how to strip the params from a url and access them by name. If we used that here, it would be less fragile if we later add more params and start-page is not always the first and only one.
___
*[`src/BloomBrowserUI/publish/commonPublish/PublishScreenTemplate.tsx` line 24 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6744#-OC5pfnoC5Rz0YOGOeQl:-OC5pfnoC5Rz0YOGOeQm:b-b8nzfg) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/187e799de2f09e6ccfa0f97019a26211096f57d2/src/BloomBrowserUI/publish/commonPublish/PublishScreenTemplate.tsx#L24)):*
> ```XML
>     bottomBanner?: React.ReactNode;
> }> = props => {
>     // Tells CSharp land to handle our external links correctly (by opening in the system browser)
> ```

That's what hookupLinkHandler does. I think the comment needs to move or be added to.

StephenMcConnel (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6744)* status: all files reviewed, 10 unresolved discussions (waiting on @JohnThomson)

___
*[`src/BloomBrowserUI/publish/commonPublish/PublishScreenTemplate.tsx` line 24 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6744#-OC5pfnoC5Rz0YOGOeQl:-OCADikD6AaiovImDRHK:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/187e799de2f09e6ccfa0f97019a26211096f57d2/src/BloomBrowserUI/publish/commonPublish/PublishScreenTemplate.tsx#L24)):*

That's what hookupLinkHandler does. I think the comment needs to move or be added to.
</blockquote></details>

Done.
___
*[`src/BloomBrowserUI/publish/ReaderPublish/ReaderPublishScreen.tsx` line 153 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6744#-OC5qLy74j9IaArkrFXo:-OCAFSIOEVI0J1e5Uqj_:b-lanejc) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/187e799de2f09e6ccfa0f97019a26211096f57d2/src/BloomBrowserUI/publish/ReaderPublish/ReaderPublishScreen.tsx#L153)):*

I think JS has a URL class that understands how to strip the params from a url and access them by name. If we used that here, it would be less fragile if we later add more params and start-page is not always the first and only one.
</blockquote></details>

Strictly speaking, the search params should start with a ?, not a &.  I've passed the (currently one and only) param in with an & so that I don't need to change the ? to a & before adding it to the **real** URL.  There's no check that this is the only param: other params could be added in the future following this one and work just fine.  I've simplified the search string to "&" instead of "&start-page=" and added a comment explaining the trickery going on here.
___
*[`src/BloomBrowserUI/utils/bloomNavHandler.ts` line 19 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6744#-OC5r38YErrnWuQD4p-i:-OCAIAv60yIaYjfoE8ut:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/187e799de2f09e6ccfa0f97019a26211096f57d2/src/BloomBrowserUI/utils/bloomNavHandler.ts#L19)):*

Not needed in JS
</blockquote></details>

Done.
___
*[`src/BloomExe/Publish/Epub/EpubMaker.cs` line 609 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6744#-OC5rRDB1G2oflPfNF41:-OCAIKp5-xuPMI2DtayO:bu9gcnd) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/187e799de2f09e6ccfa0f97019a26211096f57d2/src/BloomExe/Publish/Epub/EpubMaker.cs#L609)):*

Should we tweak this to say "other pages or books"? Or is it not worth the cost of re-localizing?
</blockquote></details>

I thought about and decided that the localization cost more than it was worth since it's such a similar concept.
___
*[`src/BloomExe/web/controllers/PublishApi.cs` line 423 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6744#-OC5sDQE67mYfez2M1dN:-OCAm2HZ2sA-NwwGf7k9:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/187e799de2f09e6ccfa0f97019a26211096f57d2/src/BloomExe/web/controllers/PublishApi.cs#L423)):*

I suppose for a GUID we don't need to worry about URLEncoding? This is a GUID, right, not a path? Might be worth a comment.
</blockquote></details>

Done.
___
*[`src/BloomExe/web/controllers/PublishApi.cs` line 424 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6744#-OC5rxY6BkVnm3R-mapf:-OCB2bhHCx-iuOTBMhi1:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/187e799de2f09e6ccfa0f97019a26211096f57d2/src/BloomExe/web/controllers/PublishApi.cs#L424)):*

C# also has classes that understand params in URLs.
</blockquote></details>

Done.
___
*[`src/BloomExe/web/controllers/PublishApi.cs` line 425 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6744#-OC5se4BEKdZCzQLYgwF:-OCAmOjG6efD61piUb_-:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/187e799de2f09e6ccfa0f97019a26211096f57d2/src/BloomExe/web/controllers/PublishApi.cs#L425)):*

Is this a decimal number or the page number the user sees? The latter might mean URLEncoding issues.

Oh, I see, it's a guid, right? Possibly worth commenting.
</blockquote></details>

Done.
___
*[`src/BloomExe/web/controllers/PublishApi.cs` line 448 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6744#-OC5tE5FBHXVN4PHnU1s:-OCB2fMDAec1AZA38wvk:bfifpy5) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/187e799de2f09e6ccfa0f97019a26211096f57d2/src/BloomExe/web/controllers/PublishApi.cs#L448)):*

Even though this is somewhat optimized, it worries me that our server will be locked for so long. Also we could be deadlocked if ever the construction of the preview requires a response from our server. Would it be a problem to postpone the rest of this, say to an Idle event handler?
</blockquote></details>

Application.Idle didn't work, so I used a timer to start the processing after returning.
___
*[`src/BloomExe/web/controllers/PublishApi.cs` line 460 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6744#-OC5ug0ADUBUehiW_NsH:-OCEdXc05Sae_9LPFPx2:b-wojgkv) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/187e799de2f09e6ccfa0f97019a26211096f57d2/src/BloomExe/web/controllers/PublishApi.cs#L460)):*

Now we're talking about the attribute value, so again need to consider whether we store non-Roman page numbers in this attribute, or only decimal ones. Might need some URLEncoding? Or maybe not for websocket...anyway worth asking.
</blockquote></details>

We do store non-Roman page numbers in this attribute, but bloom-player is expecting ASCII numbers because it calls parseInt on the string.  I've added https://github.com/BloomBooks/bloom-player/pull/341 to handle non-ASCII page numbers on the bloom-player side.
The websocket string doesn't need encoding, but the URL sent to bloom-player does.
___
*[`src/BloomExe/web/controllers/PublishApi.cs` line 809 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6744#-OC5vH6h401qZLeAgEnH:-OCEWWj51QSxcwg1Tyik:bo8k7rn) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/187e799de2f09e6ccfa0f97019a26211096f57d2/src/BloomExe/web/controllers/PublishApi.cs#L809)):*

If this continues to happen inside a normal BloomServer handler, we should note here that it MUST NOT create any browsers that call back to our server, otherwise we'll get deadlocks. (The server has some special tricks that can be used to get around this if we have to...)
</blockquote></details>

I've used a timer to put the processing on another thread, so this shouldn't pose a deadlock situation.  It already was not on the UI thread to begin with.

JohnThomson (approved):
Looks good. One thing I wasn't sure got done, but if it's OK you can go ahead and merge.

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6744#-:-OCG-7HK9oMeHfQxAz1V:b-mgtkmo)* status: 2 of 8 files reviewed, 1 unresolved discussion (waiting on @StephenMcConnel)

___
*[`src/BloomExe/web/controllers/PublishApi.cs` line 460 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6744#-OC5ug0ADUBUehiW_NsH:-OCFysXrBUnM94feWtr3:baim99z) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/187e799de2f09e6ccfa0f97019a26211096f57d2/src/BloomExe/web/controllers/PublishApi.cs#L460)):*

We do store non-Roman page numbers in this attribute, but bloom-player is expecting ASCII numbers because it calls parseInt on the string.  I've added https://github.com/BloomBooks/bloom-player/pull/341 to handle non-ASCII page numbers on the bloom-player side.
The websocket string doesn't need encoding, but the URL sent to bloom-player does.
</blockquote></details>

I'm having trouble telling whether this is done or not. The page number becomes part of a url which is sent through a web socket, presumably eventually to BloomPlayer. Is it deliberately not encoded here, because WebSocket doesn't need it, and JS somewhere encodes it before passing the URL to BloomPlayer? Or did you mean to do it here and forgot? Or did I miss it? By now you've probably done an end-to-end test involving non-Roman page numbers, so all is probably well, but it seemed worth one more check, since this is an obscure corner case that testing well might miss.
___
*[`src/BloomExe/web/controllers/PublishApi.cs` line 809 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6744#-OC5vH6h401qZLeAgEnH:-OCFxUa2Afg2vopVch0v:b-1imi6m) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/187e799de2f09e6ccfa0f97019a26211096f57d2/src/BloomExe/web/controllers/PublishApi.cs#L809)):*

I've used a timer to put the processing on another thread, so this shouldn't pose a deadlock situation.  It already was not on the UI thread to begin with.
</blockquote></details>

Even when not on the UI thread, we have a lock that stops our server from trying to do more than one thing at a time (with a few exceptions). But starting a task on another thread should be fine.

StephenMcConnel (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6744)* status: 2 of 8 files reviewed, 1 unresolved discussion (waiting on @JohnThomson)

___
*[`src/BloomExe/web/controllers/PublishApi.cs` line 460 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6744#-OC5ug0ADUBUehiW_NsH:-OCK3OElDm21hvNS1LAH:bwxprej) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/187e799de2f09e6ccfa0f97019a26211096f57d2/src/BloomExe/web/controllers/PublishApi.cs#L460)):*

I'm having trouble telling whether this is done or not. The page number becomes part of a url which is sent through a web socket, presumably eventually to BloomPlayer. Is it deliberately not encoded here, because WebSocket doesn't need it, and JS somewhere encodes it before passing the URL to BloomPlayer? Or did you mean to do it here and forgot? Or did I miss it? By now you've probably done an end-to-end test involving non-Roman page numbers, so all is probably well, but it seemed worth one more check, since this is an obscure corner case that testing well might miss.
</blockquote></details>

I verified that encoding is not needed for websocket data transfer, and as part of the changes to the javascript recipient, I made sure that the parameter values (such as page numbers) get encoded when added to the actual URL sent to bloom-player.  I don't know why my changes to ReaderPublishScreen.tsx aren't being shown by Reviewable.

PR #6770, opened 2024-11-27
Part 1 of unifying legacy and new BG image UIs (BL-14069)

JohnThomson:
Primary image containers will always have a BG overlay, even if it's a placeholder. There are also some menu option changes.
Several tasks remain to be done, but this much will allow other work to proceed.

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6770)


JohnThomson:
This is known to be incomplete. Most worrying thing is probably cover images. The code in my current PR is converting cover images, like any others, to background overlays. I was afraid this would result in losing cover images altogether until I redo the xmatter updating code, but it doesn't seem to be that bad...the data-book="cover-image" is getting moved to the img in the BG overlay. From there it gets saved, and I think we then convert things back to the old style when we update the xmatter, install the data-book src from the data-div, and then convert it to an overlay again the next time we open the page (!). So it's not a total disaster, though not a great state of affairs. Still, something is going wrong with CSS in this state, so you might want to try it out before merging. I will make sorting out the cover my next priority.

JohnThomson:
Added a commit which fixes the misalignment of the control frame and the cover image. But cover images are still not getting set to the right size when the image is changed (though it gets fixed when it gets converted to the old approach and back)

hatton (approved):
> Reviewed 12 of 14 files at r1, 2 of 2 files at r2, 2 of 2 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6770)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @JohnThomson)

PR #6773, opened 2024-12-04
Restore automatically showing source bubbles (BL-14109)

StephenMcConnel:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6773)

### DIFF
--- a/src/BloomBrowserUI/bookEdit/js/bloomEditing.ts
+++ b/src/BloomBrowserUI/bookEdit/js/bloomEditing.ts
@@ -910,6 +910,23 @@ export function SetupElements(
                 BloomSourceBubbles.ShowSourceBubbleForElement(elementToFocus);
             } else {
                 // It's OK not to focus anything.
+                if (bubbleDivs.length) {
+                    // HACK for BL-14109: source bubbles have stopped appearing automatically
+                    // because focus is set before the SourceBubble qtips have been created.
+                    // Refocusing here fixes this.
+                    // We want focus set in only a standard text box, not in an overlay
+                    const editables = container.querySelectorAll(
+                        "textarea, div.bloom-editable.bloom-visibility-code-on"
+                    ); // :visible and :not don't work in this context
+                    if (editables.length) {
+                        const editable = Array.from(editables).find(
+                            e => e.closest(".bloom-textOverPicture") === null
+                        );
+                        if (editable) {
+                            $(editable).focus(); // jquery focus() is more reliable than the native focus()
+                        }
+                    }
+                }
             }
         }
     }, bloomQtipUtils.horizontalOverlappingBubblesDelay);


JohnThomson (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6773)* status: 0 of 1 files reviewed, 3 unresolved discussions (waiting on @StephenMcConnel)

___
*[`src/BloomBrowserUI/bookEdit/js/bloomEditing.ts` line 851 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6773#-ODqJOBp6OG6Yv98EvM-:-ODqJOBp6OG6Yv98EvM0:b-xdoc8c) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/6ad2351e35cbd8aebc4423d8cd0df9e36ecdc0d5/src/BloomBrowserUI/bookEdit/js/bloomEditing.ts#L851)):*
> ```XML
>         // HACK for BL-14109: source bubbles have stopped appearing automatically because focus
>         // is set before the SourceBubble qtips have been created.  Refocusing here fixes this.
>         if (!toolboxVisible && bubbleDivs.length) {
> ```

Comment needs to explain why this should only be done if the toolbox is not visible. I sure can't think of a reason.
___
*[`src/BloomBrowserUI/bookEdit/js/bloomEditing.ts` line 853 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6773#-ODqJEZQ-11JrJ1atoZM:-ODqJEZQ-11JrJ1atoZN:bx9jt5p) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/6ad2351e35cbd8aebc4423d8cd0df9e36ecdc0d5/src/BloomBrowserUI/bookEdit/js/bloomEditing.ts#L853)):*
> ```XML
>         if (!toolboxVisible && bubbleDivs.length) {
>             // The focus set here may not actually be what is focused in the end.  Other setup
>             // code (such as for Comical) may have other--or identical--ideas and set focus.
> ```

Are you sure? This is in a (fairly long) timeout...it will beat most things. For that reason I'd like to see the minimal change that will fix BL-14109.
___
*[`src/BloomBrowserUI/bookEdit/js/bloomEditing.ts` line 860 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6773#-ODqJgMDB6FKY2uDKPe3:-ODqJgMDB6FKY2uDKPe4:bnmsryg) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/6ad2351e35cbd8aebc4423d8cd0df9e36ecdc0d5/src/BloomBrowserUI/bookEdit/js/bloomEditing.ts#L860)):*
> ```XML
>                 $(focusableElement).focus();
>             } else {
>                 const editableElement = $(container)
> ```

The long comment about 20 lines down starts out "If we don't have some specific reason to focus on a particular overlay, we don't want to arbitrarily select one." Looks as though this will. For example, if we switch to a page that has no text except overlays, I believe none of them should be initially focused (i.e., typing will not send text to any of them, and none of them has the active-overlay rectangle). (I don't know why JohnH didn't want to focus one, probably because an overlay that is fully focused for text editing can't be dragged, which can be surprising, and shows source bubbles, which can get in the way of other things. Also, I think it would be a problem if an overlay text box is actually focused, but the bubbleManager doesn't know that it is theBubbleWeAreTextEditing. We might have enough focusIn handlers to prevent that, but I'm not sure.)
Is there a case where elementToFocus is an overlay that contains nothing focusable? Maybe when we just added a picture overlay? I think this code will pick some other overlay to focus (and therefore make active) in such a case, which is probably not what we want. I would be very surprised if it is desirable to focus anything that's not inside elementToFocus (unless that is falsy).
There's code very similar to this in the block below that tests isComicEditingOn (which it pretty much always will be...code outside the timeout calls initializeOverPictureEditing(), which more-or-less always sets it...I'm not clear why we even have the test.)
IIRC, we DO want to focus the first thing OTHER than an overlay that can take focus when elementToFocus is falsy. Unless you know for sure that this needs doing in some case where isComicEditingOn is false (in which case, please comment), I think it would be better to have all the focus-handling (at least in this method) in one place. Maybe we only need a change in the else below where it says "It's OK not to focus anything" to handle that case where elementToFocus is falsy and we can find some non-overlay text box to focus?

StephenMcConnel (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6773)* status: 0 of 1 files reviewed, 3 unresolved discussions (waiting on @JohnThomson)

___
*[`src/BloomBrowserUI/bookEdit/js/bloomEditing.ts` line 851 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6773#-ODqJOBp6OG6Yv98EvM-:-ODr-a-B7A08tDt1BmUw:b-xjrpc6) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/6ad2351e35cbd8aebc4423d8cd0df9e36ecdc0d5/src/BloomBrowserUI/bookEdit/js/bloomEditing.ts#L851)):*

Comment needs to explain why this should only be done if the toolbox is not visible. I sure can't think of a reason.
</blockquote></details>

I removed the !toolboxVisible.  I don't remember why I did it, and it seems counterproductive.
___
*[`src/BloomBrowserUI/bookEdit/js/bloomEditing.ts` line 853 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6773#-ODqJEZQ-11JrJ1atoZM:-ODr7aFLBNMox3PBbb2X:bpxmuxi) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/6ad2351e35cbd8aebc4423d8cd0df9e36ecdc0d5/src/BloomBrowserUI/bookEdit/js/bloomEditing.ts#L853)):*

Are you sure? This is in a (fairly long) timeout...it will beat most things. For that reason I'd like to see the minimal change that will fix BL-14109.
</blockquote></details>

This comment was copied from elsewhere and may not apply.  I've changed it.
___
*[`src/BloomBrowserUI/bookEdit/js/bloomEditing.ts` line 860 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6773#-ODqJgMDB6FKY2uDKPe3:-ODr8Nj-CNJFaYuiy11C:bshnc14) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/6ad2351e35cbd8aebc4423d8cd0df9e36ecdc0d5/src/BloomBrowserUI/bookEdit/js/bloomEditing.ts#L860)):*

The long comment about 20 lines down starts out "If we don't have some specific reason to focus on a particular overlay, we don't want to arbitrarily select one." Looks as though this will. For example, if we switch to a page that has no text except overlays, I believe none of them should be initially focused (i.e., typing will not send text to any of them, and none of them has the active-overlay rectangle). (I don't know why JohnH didn't want to focus one, probably because an overlay that is fully focused for text editing can't be dragged, which can be surprising, and shows source bubbles, which can get in the way of other things. Also, I think it would be a problem if an overlay text box is actually focused, but the bubbleManager doesn't know that it is theBubbleWeAreTextEditing. We might have enough focusIn handlers to prevent that, but I'm not sure.)
Is there a case where elementToFocus is an overlay that contains nothing focusable? Maybe when we just added a picture overlay? I think this code will pick some other overlay to focus (and therefore make active) in such a case, which is probably not what we want. I would be very surprised if it is desirable to focus anything that's not inside elementToFocus (unless that is falsy).
There's code very similar to this in the block below that tests isComicEditingOn (which it pretty much always will be...code outside the timeout calls initializeOverPictureEditing(), which more-or-less always sets it...I'm not clear why we even have the test.)
IIRC, we DO want to focus the first thing OTHER than an overlay that can take focus when elementToFocus is falsy. Unless you know for sure that this needs doing in some case where isComicEditingOn is false (in which case, please comment), I think it would be better to have all the focus-handling (at least in this method) in one place. Maybe we only need a change in the else below where it says "It's OK not to focus anything" to handle that case where elementToFocus is falsy and we can find some non-overlay text box to focus?
</blockquote></details>

I adjusted the selector for the find to exclude overlay bubbles, and moved a simplified version of the code to the else clause below as suggested.  it seems to still be working as desired.  (For some reason in my earlier work, I didn't think that was the case.)

JohnThomson (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6773)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @StephenMcConnel)

___
*[`src/BloomBrowserUI/bookEdit/js/bloomEditing.ts` line 923 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6773#-ODrVnP70xkVBhMxxEDt:-ODrVnP70xkVBhMxxEDu:b-v217b0) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/26ab5e8b95b1307e95680e7cc821f3be8ab0e0ab/src/BloomBrowserUI/bookEdit/js/bloomEditing.ts#L923)):*
> ```XML
>                     if (editableElement.length) {
>                         editableElement.focus();
>                     }
> ```

This is close. However,
(a) I don't think editables in overlays HAVE to have Bubble-style; it's the default but the user could change it. The definitive thing is whether they are inside a .bloom-textOverPicture.  I'm not sure how to filter for that in jquery; in regular JS I would use filter and closest
(b) I think there are cases...maybe only in some special template like Calendar or BigBook?...where we have textarea rather than div.bloom-editable.  A simple find could use "textarea, div.bloom-editable" but I'm not sure how to combine that union with the other conditions.
(c) The way jquery works, editableElement will be a list  of all the matches. They can't all be focused. Probably it will just focus the first, but I wanted to make sure you checked that it does whatever we want.
Fwiw, my preference is getting stronger and stronger for using non-jquery code where there's not a strong reason to use jquery. There's still a lot of jquery in this file, but I'd prefer to see it decreasing rather than increasing, other things being equal.

StephenMcConnel (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6773)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @JohnThomson)

___
*[`src/BloomBrowserUI/bookEdit/js/bloomEditing.ts` line 923 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6773#-ODrVnP70xkVBhMxxEDt:-ODrZ22i6P5eSKe69YmV:bkxxmf4) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/26ab5e8b95b1307e95680e7cc821f3be8ab0e0ab/src/BloomBrowserUI/bookEdit/js/bloomEditing.ts#L923)):*

This is close. However,
(a) I don't think editables in overlays HAVE to have Bubble-style; it's the default but the user could change it. The definitive thing is whether they are inside a .bloom-textOverPicture.  I'm not sure how to filter for that in jquery; in regular JS I would use filter and closest
(b) I think there are cases...maybe only in some special template like Calendar or BigBook?...where we have textarea rather than div.bloom-editable.  A simple find could use "textarea, div.bloom-editable" but I'm not sure how to combine that union with the other conditions.
(c) The way jquery works, editableElement will be a list  of all the matches. They can't all be focused. Probably it will just focus the first, but I wanted to make sure you checked that it does whatever we want.
Fwiw, my preference is getting stronger and stronger for using non-jquery code where there's not a strong reason to use jquery. There's still a lot of jquery in this file, but I'd prefer to see it decreasing rather than increasing, other things being equal.
</blockquote></details>

a) I've rewritten the code to use regular javascript to find the appropriate element, using Array.find to filter against .bloom-textOverPicture.
b) I've rewritten the selector code to include textarea.  Using regular javascript meant that :visible and :not() are not valid syntax in selectors.
c) The previous jquery find call was followed by a first call to trim the list to no more than one element, fwiw.
I'm still using jquery for the focus() method call because jquery's focus method just works a lot better.  The regular javascript focus() method often seems to have no effect.  I can't find the reference now, but I read earlier that this is a known behavior.

andrew-polk (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6773)* status: all files reviewed, 1 unresolved discussion (waiting on @JohnThomson)

andrew-polk (approved):
> Dismissed @StephenMcConnel from a discussion.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6773)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @StephenMcConnel)

PR #6774, opened 2024-12-05
Replace SIL.W.F.Registration with simpler storage (BL-14101)

StephenMcConnel:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6774)

### DIFF
--- a/src/BloomExe/Program.cs
+++ b/src/BloomExe/Program.cs
@@ -111,12 +111,9 @@ namespace Bloom
             // other programs for 64K available default temp file names.
             TempFile.NamePrefix = "bloom";
             CheckForCorruptUserConfig();
-            // We want to do this as early as possible, definitely before we create
-            // the TeamCollectionManager, which needs the registered user information.
-            // But since it's difficult to predict what else might want it, it's best
-            // to do it before we create the Autofac context objects that create all
-            // our singletons.
-            RegistrationDialog.UpgradeRegistrationIfNeeded();
+            // Ensure that the registration information is loaded early before Team Collection
+            // needs it.
+            Registration.Registration.Default.EnsureLoaded();
 
             // Initializing localization can pop up a dialog if the system language
             // is not one of our supported languages. The following calls must be done
@@ -1370,7 +1367,7 @@ namespace Bloom
             // User clicked the report button.
             if (result == reportPressedResult)
             {
-                var userEmail = SIL.Windows.Forms.Registration.Registration.Default.Email;
+                var userEmail = Bloom.Registration.Registration.Default.Email;
                 if (!String.IsNullOrWhiteSpace(userEmail))
                 {
                     // Just send the report in for them.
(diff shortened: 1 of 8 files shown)


JohnThomson (changes requested):
One case where we will lose a little information is if the user launches a 6.2 (and so creates a new Registration file), then launches a 6.1 or earlier, which will update LaunchCount in the old settings. But when 6.2 is launched again, the new Registration LaunchCount will not have been updated. Tricky to fix...we'd probably have to remember the old settings launch count and adjust it.
Also, of course, any changes to registration made using 6.1 or earlier after a valid registration has been migrated to 6.2 will get lost.
I doubt this is worth worrying about. Not sure we even report LaunchCount; we may just use it to tell whether Bloom has been launched twice and we should ask them to register.

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6774#-:-OFd50Z0Erdm9m7s9N2l:bpxayed)* status: all files reviewed, 3 unresolved discussions (waiting on @StephenMcConnel)

___
*[`src/BloomExe/Registration/Registration.cs` line 59 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6774#-OFd6fuS3nvY_NXTzEhm:-OFd6fuS3nvY_NXTzEhn:bkv27tv) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/f8be1ba682dc03175cdebfe770651d701b1907f0/src/BloomExe/Registration/Registration.cs#L59)):*
> ```Smalltalk
>             if (string.IsNullOrEmpty(rawData))
>                 return;
>             var settings = DynamicJson.Parse(rawData);
> ```

I would be inclined to save this "settings" in a member variable, make the various properties read from it and update it, and then simply serialize it in Save(). Then if any future version of Bloom adds another setting, old versions will preserve it.
Probably unlikely...I won't insist.
___
*[`src/BloomExe/Registration/RegistrationDialog.cs` line 141 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6774#-OFd3inN6QTrwPS0PE6S:-OFd3inN6QTrwPS0PE6T:b-cgpa8e) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/f8be1ba682dc03175cdebfe770651d701b1907f0/src/BloomExe/Registration/RegistrationDialog.cs#L141)):*
> ```Smalltalk
>             //    Registration.Default.NeedUpgrade = false;
>             //    Registration.Default.Save();
>             //}
> ```

I doubt we've ever modified our registration settings. If we had, and were loading a sufficiently old version, wouldn't we want this? But maybe the point is that we had this code in case we ever modified it, and now we know we never will.
If the method is really obsolete, I'd be inclined to delete it along with all callers.
___
*[`src/BloomExe/Program.cs` line 1756 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6774#-OFd2C3BD77QnsqHmxYw:-OFd2C3BD77QnsqHmxYx:b-1tojiu) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/f8be1ba682dc03175cdebfe770651d701b1907f0/src/BloomExe/Program.cs#L1756)):*
> ```Smalltalk
>                         scope.User = new User
>                         {
>                             Email = Bloom.Registration.Registration.Default.FirstName,
> ```

I think it's a mistake that we now set the email to firstName, and the UserName not at all?

JohnThomson (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6774)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @StephenMcConnel)

PR #6777, opened 2024-12-09
Allow Andika to be used in the UI when not installed (BL-14102)

andrew-polk:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6777)

### DIFF
--- a/src/BloomBrowserUI/bloomWebFonts.less
+++ b/src/BloomBrowserUI/bloomWebFonts.less
@@ -117,6 +117,17 @@
     font-weight: normal;
 }
 
+// There are at least two places in the UI where we expect to use Andika:
+// - the font chooser (it lists font names in their font)
+// - the book titles in the collection tab book lists
+// Without this rule, if the user doesn't have Andika installed
+// (which is likely), they will get a fallback font. See BL-14102.
+@font-face {
+    font-family: "Andika";
+    src: url("/bloom/fonts/Andika-Regular.woff2") format("woff2");
+    font-weight: normal;
+}
+
 // Should match kUiFontStack in bloomMaterialUITheme.ts.
 // And a rule in jquery-ui-1.8.16.custom.css.
 @UIFontStack: Roboto, NotoSans, sans-serif;
--- a/src/BloomExe/web/BloomServer.cs
+++ b/src/BloomExe/web/BloomServer.cs
@@ -843,6 +843,17 @@ namespace Bloom.Api
             {
                 return FontsApi.ProcessHostFontsRequest(info, localPath);
             }
+            if (localPath.Contains("fonts/Andika"))
+            {
+                // Rightly or wrongly, Andika is in a different place in our
+                // repo structure than the other UI fonts because it with the book fonts.
+                // To keep from having to duplicate the Andika font files in both places,
+                // this workaround maps the request for Andika as a UI font as if it were a book font.
+                return FontsApi.ProcessHostFontsRequest(
+                    info,
+                    localPath.Replace("fonts/", "/host/fonts/")
+                );
+            }
 
             switch (localPath)
             {


andrew-polk (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6777)* status: 0 of 2 files reviewed, all discussions resolved

___
*[`src/BloomExe/ProjectContext.cs` line 94 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6777#-ODhrKOCDFePXd8WteYF:-ODhrKOCDFePXd8WteYG:bcvkhfh) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/1c3aaa0516951918b19a2f209105af57175145d9/src/BloomExe/ProjectContext.cs#L94)):*
> ```Smalltalk
>                         GetAfterXMatterFileLocations()
>                     )
>                 );
> ```

Without doing something like this, the BloomFileLocator is first created without many directories available to it, including the DistFiles/fonts folder which is needed to allow the server to find the font on the initial load.
(Previously, it would later comes along with a call on line 287 and create it fully.)

But autofac is still enough of a mystery to me that I'm not 100% this is safe. So any input on that would be great.

andrew-polk (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6777)* status: 1 of 2 files reviewed, all discussions resolved (waiting on @hatton)

___
*[`src/BloomExe/ProjectContext.cs` line 94 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6777#-ODhrKOCDFePXd8WteYF:-ODlHAZg8SfxgOqmgdXA:b-gssqof) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/1c3aaa0516951918b19a2f209105af57175145d9/src/BloomExe/ProjectContext.cs#L94)):*
That's correct.
We create a BloomFileLocator here before the full BloomFileLocator created on line 287.
The timing is such that at startup, when the server is trying to resolve the path to the hosted Andika font, it uses the incomplete BloomFileLocator and doesn't find the font.
So making this first-created BloomFileLocator a full one was my workaround.

> let's do this is in a cleaner way

:+1:

hatton (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6777)* status: 1 of 2 files reviewed, all discussions resolved

___
*[`src/BloomExe/ProjectContext.cs` line 94 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6777#-ODhrKOCDFePXd8WteYF:-ODlJ9NL7LTW0Joxl1Zn:b-j9i79c) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/1c3aaa0516951918b19a2f209105af57175145d9/src/BloomExe/ProjectContext.cs#L94)):*

That's correct.
We create a BloomFileLocator here before the full BloomFileLocator created on line 287.
The timing is such that at startup, when the server is trying to resolve the path to the hosted Andika font, it uses the incomplete BloomFileLocator and doesn't find the font.
So making this first-created BloomFileLocator a full one was my workaround.

> let's do this is in a cleaner way

:+1:
</blockquote></details>

In the case of fonts, we're not really wanting or needing to look in various places, which is what BloomFileLocator is for. Can we instead just look in the distribution fonts?

andrew-polk (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6777)* status: 1 of 3 files reviewed, all discussions resolved (waiting on @hatton)

___
*[`src/BloomExe/ProjectContext.cs` line 94 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6777#-ODhrKOCDFePXd8WteYF:-ODx2GO0AZhtHUS1XPoq:b5n0s43) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/1c3aaa0516951918b19a2f209105af57175145d9/src/BloomExe/ProjectContext.cs#L94)):*

In the case of fonts, we're not really wanting or needing to look in various places, which is what BloomFileLocator is for. Can we instead just look in the distribution fonts?
</blockquote></details>

Good point. Thanks.
This simplifies it enough that we can probably push it to 6.0.

PR #6778, opened 2024-12-10
Don't rely on style settings for element position and size (BL-14106)

StephenMcConnel:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6778)

### DIFF
--- a/src/BloomBrowserUI/bookEdit/js/bubbleManager.ts
+++ b/src/BloomBrowserUI/bookEdit/js/bubbleManager.ts
@@ -1230,8 +1230,8 @@ export class BubbleManager {
         )[0];
         const img = imgC?.getElementsByTagName("img")[0];
         if (!img) return;
-        this.oldImageTop = BubbleManager.pxToNumber(img.style.top);
-        this.oldImageLeft = BubbleManager.pxToNumber(img.style.left);
+        this.oldImageTop = img.offsetTop;
+        this.oldImageLeft = img.offsetLeft;
         this.lastCropControl = undefined;
         this.startMoveCropControlX = this.currentDragControl.offsetLeft;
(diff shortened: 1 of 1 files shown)


JohnThomson (commented):
In many cases you are switching from using the style value to clientX, while in other cases you are making the opposite change. I have a vague recollection that I switched between the two in some cases while I was writing this code, but I seem to have failed to capture the reason. Can you find a good place to document which one is to be preferred in what circumstances? For example, it might be that using clientX forces the browser to do a layout, so using style.X is better if we're sure that's the only thing determining it. Conversely, there may be cases (maybe only in old or legacy books) where style.X might be a percentage and we need clientX to get a value that is reliably in pixels. Or it may be that some elements we're working with have not yet been added to the document and clientX will not be defined. In any case, anything you know should be documented, since apparently it is easy to make a mistake that is not immediately obvious.

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6778#-:-ODrRrT76yvQAYonQiSA:b4vy7eh)* status: 0 of 1 files reviewed, all discussions resolved

StephenMcConnel (commented):
I am not ever changing from clientX to style.X in this code.  The style values can be either in px or % (from old books) and thus are totally useless to get a good value without a great deal of computation.  The critique I was expecting was whether offsetHeight and offsetWidth are better measurements to use than clientWidth and clientHeight.  The former include any scrollbars and borders, and the latter don't.

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6778#-:-ODrqenX1s2EU2t6Mo5O:bpjfaln)* status: 0 of 1 files reviewed, all discussions resolved

PR #6779, opened 2024-12-11
Store color palettes apart from other collection settings (BL-14129)

StephenMcConnel:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6779)

### DIFF
--- a/src/BloomExe/Collection/CollectionSettings.cs
+++ b/src/BloomExe/Collection/CollectionSettings.cs
@@ -324,13 +324,31 @@ namespace Bloom.Collection
                 xml.Add(new XElement("DefaultBookTags", "bookshelf:" + DefaultBookshelf));
             }
             xml.Add(BulkPublishBloomPubSettings.ToXElement());
-            foreach (var key in ColorPalettes.Keys)
-            {
-                var palette = new XElement("Palette", ColorPalettes[key]);
-                palette.Add(new XAttribute("id", key));
-                xml.Add(palette);
-            }
-            SIL.IO.RobustIO.SaveXElement(xml, SettingsFilePath);
+            RobustIO.SaveXElement(xml, SettingsFilePath);
+
+            // Color palette settings are stored in a separate Json file
+            SaveColorPalettesToJsonFile();
+        }
+
+        internal void SaveColorPalettesToJsonFile()
+        {
+            // If we're in a Team Collection, the file checker will try to sync up with any changes
+            // that have occurred remotely.
+            var jsonFilePath = Path.Combine(FolderPath, "colorPalettes.json");
+            SaveColorPalettesToJsonFile(ColorPalettes, jsonFilePath);
+        }
+
+        internal static void SaveColorPalettesToJsonFile(
+            Dictionary<string, string> colorPalettes,
+            string jsonFilePath
+        )
+        {
+            var jsonString = new StringBuilder();
+            jsonString.AppendLine("{");
+            foreach (var key in colorPalettes.Keys)
+                jsonString.AppendLine($"\"{key}\":\"{colorPalettes[key]}\",");
+            jsonString.AppendLine("}");
+            RobustFile.WriteAllText(jsonFilePath, jsonString.ToString());
         }
 
         public string GetCollectionStylesCss(bool omitDirection)
@@ -739,19 +757,51 @@ namespace Bloom.Collection
             }
         }
 
-        internal static void LoadDictionary(
-            XElement document,
-            string tag,
-            Dictionary<string, string> dict
-        )
+        internal void LoadDictionary(XElement document, string tag, Dictionary<string, string> dict)
         {
             dict.Clear();
+
+            // The color palettes are now stored in a separate JSON file, so we try to load them
+            // from the JSON file before looking at the XML data.  (The XML data is not removed
+            // until the whole collection settings are saved.)
+            if (tag == "Palette")
+            {
+                var path = Path.Combine(FolderPath, "colorPalettes.json");
+                if (LoadColorPalettesFromJsonFile(dict, path))
+                    return;
+            }
             var elements = document.Descendants(tag);
             if (elements != null)
             {
                 foreach (XElement element in elements)
                     dict[element.Attribute("id").Value] = element.Value;
             }
+        }
+
+        internal static bool LoadColorPalettesFromJsonFile(
+            Dictionary<string, string> dict,
+            string path
+        )
+        {
+            if (RobustFile.Exists(path))
+            {
+                var jsonString = RobustFile.ReadAllText(path);
+                try
+                {
+                    var json = JObject.Parse(jsonString);
+                    if (json != null)
+                    {
+                        foreach (var property in json.Properties())
+                            dict[property.Name] = property.Value.ToString();
+                    }
+                    return true;
+                }
+                catch (Exception ex)
+                {
+                    Logger.WriteEvent($"Error loading color palettes from {path}: " + ex.Message);
+                }
+            }
+            return false;
         }
 
         public virtual string CollectionName { get; protected set; }
@@ -1026,7 +1076,7 @@ namespace Bloom.Collection
         public bool HaveEnterpriseSubscription =>
             HaveEnterpriseFeatures && BrandingProjectKey != "Local-Community";
 
-        private readonly Dictionary<string, string> ColorPalettes =
+        internal readonly Dictionary<string, string> ColorPalettes =
             new Dictionary<string, string>();
 
         private string _brandingProjectKey;
@@ -1046,6 +1096,8 @@ namespace Bloom.Collection
                     var pieces = savedColor.Split('/');
                     if (pieces.Length > 1)
                         opacity = double.Parse(pieces[1], NumberFormatInfo.InvariantInfo);
+                    if (opacity > 1)
+                        opacity = opacity / 100; // some old values stored range of 0-100 instead of 0-1
                     var colors = pieces[0].Split('-');
                     // Opacity needs to be formatted in an invariant manner to keep the file format stable.
                     var colorElement =
@@ -1092,7 +1144,7 @@ namespace Bloom.Collection
                         return;
                     savedPalette = savedPalette + " " + colorToSave;
                     ColorPalettes[paletteTag] = savedPalette.Trim();
-                    Save();
+                    SaveColorPalettesToJsonFile();
                 }
                 catch (Microsoft.CSharp.RuntimeBinder.RuntimeBinderException)
                 {
--- a/src/BloomExe/TeamCollection/DisconnectedTeamCollection.cs
+++ b/src/BloomExe/TeamCollection/DisconnectedTeamCollection.cs
@@ -166,6 +166,16 @@ namespace Bloom.TeamCollection
             throw new NotImplementedException();
         }
 
+        protected override void SyncColorPaletteFileWithRepo(string destFolder)
+        {
+            throw new NotImplementedException();
+        }
+
+        protected override DateTime GetRepoColorPaletteTime()
+        {
+            throw new NotImplementedException();
+        }
+
         protected override DateTime LastRepoCollectionFileModifyTime { get; }
 
         protected override void CopyLocalFolderToRepo(string folderName)
--- a/src/BloomExe/TeamCollection/FolderTeamCollection.cs
+++ b/src/BloomExe/TeamCollection/FolderTeamCollection.cs
@@ -467,7 +467,11 @@ namespace Bloom.TeamCollection
         }
 
         /// <summary>
-        /// Copy collection level files from the repo to the local directory
+        /// Copy collection level files from the repo to the local directory.
+        /// The local copy of the colorPalettes.json file will be merged with,
+        /// rather than copied from, the repo copy.  The colorPalettes.json file
+        /// in the repo may be updated as a result of this call although that is
+        /// unlikely.
         /// </summary>
         /// <param name="destFolder"></param>
         protected override void CopyRepoCollectionFilesToLocalImpl(string destFolder)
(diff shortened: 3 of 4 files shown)


JohnThomson (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6779)* status: all files reviewed, 4 unresolved discussions (waiting on @StephenMcConnel)

___
*[a discussion](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6779#-OFdKiWX8TanrMPfCoGV:-OFdKiWX8TanrMPfCoGW:b-5rz4lr) (no related file):*
This is an elegant way to preserve the behavior that defined colors are shared across all books and the TC team, while fixing the bug that changing the collection settings when not an admin (by adding a color) results in an obscure error.
But it worries me. Previously, TC collection-level files were only modified by the team admin (and we have no official support for more than one admin). So there was no question who had permission to modify these files. And book repo files could only be modified by someone who had them checked out. So (barring errors) there was only one possible writer for any repo file at any time.
Now, one of our most protected files (the zip of collection settings) can be modified by anyone at any time. The most recent change wins. Two at once will probably lead to DropBox creating a conflict file.
It won't often be a problem. Disciplined teams will probably rarely add colors. There is no way to remove them (IIRC), and you elegantly merge repo colors with local ones so local additions never get lost. (Well, in theory they could at startup, when the repo collection settings are copied over the local ones. But collection settings changes are written to the repo at once, so a loss should only be possible if a color is added while offline.)
I don't know what else to do, so probably we should just give it a try. I'd kind of like to talk it over with @andrew-polk and @hatton, though.
___
*[`src/BloomExe/TeamCollection/TeamCollection.cs` line 758 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6779#-OFdCTfRB_2lcHQ6mCKf:-OFdCTfRB_2lcHQ6mCKg:b-rmxful) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/a57bcf2d2c751844e903af0563624fe77fb8db3c/src/BloomExe/TeamCollection/TeamCollection.cs#L758)):*
> ```Smalltalk
>         /// file, and when we detect a possibly-significant change to a local file while Bloom
>         /// is running. Only at startup are we allowed to copy changes TO the local system,
>         /// since changes to the collection-level files normally require a restart.
> ```

If I'm reading the code right, changes to the palette are an exception to this (can be copied from repo to local even when not atStartup), so the comment needs some adjusting.
___
*[`src/BloomExe/TeamCollection/TeamCollection.cs` line 920 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6779#-OFd9U1j9LKsbOxGhE91:-OFd9U1j9LKsbOxGhE92:buc2kjc) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/a57bcf2d2c751844e903af0563624fe77fb8db3c/src/BloomExe/TeamCollection/TeamCollection.cs#L920)):*
> ```Smalltalk
> 
>         /// <summary>
>         /// Return true if local collection-level files have changed and need to be copied
> ```

No longer returns true, so some update is needed here
___
*[`src/BloomExe/TeamCollection/TeamCollection.cs` line 942 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6779#-OFdRc0x7h7EIQq9armv:-OFdRc0x7h7EIQq9armw:b-f474te) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/a57bcf2d2c751844e903af0563624fe77fb8db3c/src/BloomExe/TeamCollection/TeamCollection.cs#L942)):*
> ```Smalltalk
>                 var colorPaletteTime = new FileInfo(colorPaletteFile).LastWriteTime;
>                 if (colorPaletteTime > savedModTime)
>                     colorPaletteChanged = SyncColorPaletteFileFromRepo();
> ```

It bothers me that a routine that is supposed to just tell you whether anything local updated is changing not only program state, but file contents. Would it be very hard to change it so that this tells you whether there's a real change, but updating the local file if necessary is done in SyncLocalAndRepoCollectionFiles? We might duplicate a bit of logic (and do extra work) in the case where a merge is really needed, but since changing the colors file will result in an immediate update to the repo, that should be very rare.

StephenMcConnel (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6779)* status: 0 of 4 files reviewed, 4 unresolved discussions (waiting on @JohnThomson)

___
*[`src/BloomExe/TeamCollection/TeamCollection.cs` line 758 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6779#-OFdCTfRB_2lcHQ6mCKf:-OG1wHmq73s7IZrroqSD:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/a57bcf2d2c751844e903af0563624fe77fb8db3c/src/BloomExe/TeamCollection/TeamCollection.cs#L758)):*

If I'm reading the code right, changes to the palette are an exception to this (can be copied from repo to local even when not atStartup), so the comment needs some adjusting.
</blockquote></details>

Done.
___
*[`src/BloomExe/TeamCollection/TeamCollection.cs` line 920 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6779#-OFd9U1j9LKsbOxGhE91:-OG1vo9H48VvWHt5p1aj:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/a57bcf2d2c751844e903af0563624fe77fb8db3c/src/BloomExe/TeamCollection/TeamCollection.cs#L920)):*

No longer returns true, so some update is needed here
</blockquote></details>

Done.
___
*[`src/BloomExe/TeamCollection/TeamCollection.cs` line 942 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6779#-OFdRc0x7h7EIQq9armv:-OG1wJGTCgayrxmRc1AF:b-aeu72x) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/a57bcf2d2c751844e903af0563624fe77fb8db3c/src/BloomExe/TeamCollection/TeamCollection.cs#L942)):*

It bothers me that a routine that is supposed to just tell you whether anything local updated is changing not only program state, but file contents. Would it be very hard to change it so that this tells you whether there's a real change, but updating the local file if necessary is done in SyncLocalAndRepoCollectionFiles? We might duplicate a bit of logic (and do extra work) in the case where a merge is really needed, but since changing the colors file will result in an immediate update to the repo, that should be very rare.
</blockquote></details>

The code has been adjusted along these lines.  The code has also been adjusted to store a plain colorPalettes.json file instead of zipping it up with the other collection files.  This should minimize damage in cases of conflicts.

JohnThomson (changes requested):
Looking good now. A few minor points.

> Reviewed 3 of 4 files at r2, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6779#-:-OG5xZL5-Mls0HWgj30r:by7czrm)* status: all files reviewed, 4 unresolved discussions (waiting on @StephenMcConnel)

___
*[`src/BloomExe/TeamCollection/FolderTeamCollection.cs` line 483 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6779#-OG5qmpx-f7gXym2qe0d:-OG5qmpx-f7gXym2qe0e:b-9gyg9s) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/83a87d34fc722fd039935fcc2a44baf977aa2335/src/BloomExe/TeamCollection/FolderTeamCollection.cs#L483)):*
> ```Smalltalk
>             ExtractFolder(destFolder, _repoFolderPath, "Allowed Words");
>             ExtractFolder(destFolder, _repoFolderPath, "Sample Texts");
>             SyncColorPaletteFileWithRepo(destFolder);
> ```

This method may update the repo, too, if there are local changes not yet synced (unlikely, since I think you save them immediately). But that feels awkward in a method whose name suggests only copying the other way. I can't think of a better name, but at least the exception should be noted in the summary.
___
*[`src/BloomExe/TeamCollection/FolderTeamCollection.cs` line 539 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6779#-OG5ry9C5ENkz9VKMPfF:-OG5pIAy-VM-E4HC6Dmx:b-pewahy) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/83a87d34fc722fd039935fcc2a44baf977aa2335/src/BloomExe/TeamCollection/FolderTeamCollection.cs#L539)):*
> ```Smalltalk
>                             _tcManager.Settings.SaveColorPalettesToJsonFile();
>                             RobustFile.Copy(destColorPalettePath, repoColorPalettePath, true);
>                         }
> ```

Is the third case (_tcManager.Settings != null && _tcManager.Settings.FolderPath != destFolder) actually possible? If we're sure it's not maybe we don't need the condition on the else? Or if it's possible but we definitely don't need to save the dirty settings in that case, I think a comment would help. Or if we think it's not possible but want to check for safety, maybe it's worth having an else with at least a Debug.Fail?
Also: both normal paths write the repo file even if if didn't change. I think it's more likely than not that it didn't change (only if there are local changes not yet synced) even when local did (new changes from others), so many of these writes are not needed. And they are dangerous writes: they could revert someone else's recent change, or cause DropBox to create a conflict file, or maybe even crash Bloom: if we're working in a shared-local-folder collection and two users try to write it at the same time we might get an error where one has the file locked and the other is trying to write it.
I think it's possibly worth checking whether the repo file needs to be written, and definitely worth a try...catch to just ignore failure to write the repo file (worst case: the new color I added won't be seen by the rest of the team until something causes another sync).
___
*[`src/BloomExe/TeamCollection/FolderTeamCollection.cs` line 611 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6779#-OG5vkcWBpZHULir-cJd:-OG5vkcWBpZHULir-cJe:bmenzew) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/83a87d34fc722fd039935fcc2a44baf977aa2335/src/BloomExe/TeamCollection/FolderTeamCollection.cs#L611)):*
> ```Smalltalk
>         }
> 
>         private static void SyncColorPaletteFileWithRepo(string destFolder, string repoFolder)
> ```

This appears to be very nearly a duplicate of the non-static method above with the same name. Should the other just call this, supplying the needed _repoFolderPath?

StephenMcConnel (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6779)* status: 2 of 4 files reviewed, 4 unresolved discussions (waiting on @andrew-polk, @hatton, and @JohnThomson)

___
*[a discussion](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6779#-OFdKiWX8TanrMPfCoGV:-OG6-bs4CaNqhTl_3pUE:b-hwftsw) (no related file):*

This is an elegant way to preserve the behavior that defined colors are shared across all books and the TC team, while fixing the bug that changing the collection settings when not an admin (by adding a color) results in an obscure error.
But it worries me. Previously, TC collection-level files were only modified by the team admin (and we have no official support for more than one admin). So there was no question who had permission to modify these files. And book repo files could only be modified by someone who had them checked out. So (barring errors) there was only one possible writer for any repo file at any time.
Now, one of our most protected files (the zip of collection settings) can be modified by anyone at any time. The most recent change wins. Two at once will probably lead to DropBox creating a conflict file.
It won't often be a problem. Disciplined teams will probably rarely add colors. There is no way to remove them (IIRC), and you elegantly merge repo colors with local ones so local additions never get lost. (Well, in theory they could at startup, when the repo collection settings are copied over the local ones. But collection settings changes are written to the repo at once, so a loss should only be possible if a color is added while offline.)
I don't know what else to do, so probably we should just give it a try. I'd kind of like to talk it over with @andrew-polk and @hatton, though.
</blockquote></details>

The color palettes are now stored in the json file independently of the zipped up collection files.  They are not zipped up to minimize risk of conflicts with other users since any user can change the color palettes at any time.
___
*[`src/BloomExe/TeamCollection/FolderTeamCollection.cs` line 483 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6779#-OG5qmpx-f7gXym2qe0d:-OG61r6aE1nDvjMg3ZB0:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/83a87d34fc722fd039935fcc2a44baf977aa2335/src/BloomExe/TeamCollection/FolderTeamCollection.cs#L483)):*

This method may update the repo, too, if there are local changes not yet synced (unlikely, since I think you save them immediately). But that feels awkward in a method whose name suggests only copying the other way. I can't think of a better name, but at least the exception should be noted in the summary.
</blockquote></details>

Done.
___
*[`src/BloomExe/TeamCollection/FolderTeamCollection.cs` line 539 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6779#-OG5ry9C5ENkz9VKMPfF:-OG6XCj42lXCYOPqyr_C:b84f9ya) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/83a87d34fc722fd039935fcc2a44baf977aa2335/src/BloomExe/TeamCollection/FolderTeamCollection.cs#L539)):*

Is the third case (_tcManager.Settings != null && _tcManager.Settings.FolderPath != destFolder) actually possible? If we're sure it's not maybe we don't need the condition on the else? Or if it's possible but we definitely don't need to save the dirty settings in that case, I think a comment would help. Or if we think it's not possible but want to check for safety, maybe it's worth having an else with at least a Debug.Fail?
Also: both normal paths write the repo file even if if didn't change. I think it's more likely than not that it didn't change (only if there are local changes not yet synced) even when local did (new changes from others), so many of these writes are not needed. And they are dangerous writes: they could revert someone else's recent change, or cause DropBox to create a conflict file, or maybe even crash Bloom: if we're working in a shared-local-folder collection and two users try to write it at the same time we might get an error where one has the file locked and the other is trying to write it.
I think it's possibly worth checking whether the repo file needs to be written, and definitely worth a try...catch to just ignore failure to write the repo file (worst case: the new color I added won't be seen by the rest of the team until something causes another sync).
</blockquote></details>

I've fixed the code to write the repo file only if it seems to need to be written.
I added a comment for the "third case" that you mentioned above.  I haven't convinced myself that the situation can never happen, or that if it happens we need to save anything.
___
*[`src/BloomExe/TeamCollection/FolderTeamCollection.cs` line 611 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6779#-OG5vkcWBpZHULir-cJd:-OG6XNFKBcg7X-bVOc8J:bqzqx1r) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/83a87d34fc722fd039935fcc2a44baf977aa2335/src/BloomExe/TeamCollection/FolderTeamCollection.cs#L611)):*

This appears to be very nearly a duplicate of the non-static method above with the same name. Should the other just call this, supplying the needed _repoFolderPath?
</blockquote></details>

More information than just _repoFolderPath is needed, but I've revised the code to avoid what was mostly duplicate code.

JohnThomson (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6779)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @StephenMcConnel)

PR #6781, opened 2024-12-12
Fix scroll position for zoomed "Just text" pages (BL-14112)

StephenMcConnel:
This could be a candidate for cherry-picking to 6.1.  A similar fix is needed in bloom-player.  (The contexts are just different enough to not to be able to share code.)

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6781)

### DIFF
--- a/src/BloomBrowserUI/bookEdit/js/niceScrollBars.ts
+++ b/src/BloomBrowserUI/bookEdit/js/niceScrollBars.ts
@@ -63,19 +63,15 @@ export function addScrollbarsToPage(bloomPage: Element): void {
                         );
                     }
                 }
-                let topAdjust = 0;
-                let leftAdjust = 0;
-                if (scale !== 1) {
-                    const compStyles = window.getComputedStyle(
-                        elt.parentElement!
-                    );
-                    const topPadding =
-                        compStyles.getPropertyValue("padding-top") ?? "0";
-                    const leftPadding =
-                        compStyles.getPropertyValue("padding-left") ?? "0";
-                    topAdjust = parseFloat(topPadding) * (scale - 1);
-                    leftAdjust = parseFloat(leftPadding) * (scale - 1);
-                }
+                // nicescroll doesn't provide a way to adjust the height of the scrollbar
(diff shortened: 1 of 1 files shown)


JohnThomson (changes requested):
I haven't looked closely at the two main functions since you say they are the same as in Bloom Player, where I already reviewed them.

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6781#-:-OG5PltbCNRoQ_FsATbM:bexvq8j)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @StephenMcConnel)

___
*[`src/BloomBrowserUI/bookEdit/js/niceScrollBars.ts` line 68 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6781#-OG5Oayq42yBlxxjNouz:-OG5OayrDn08ZisyDOFf:bgsguhq) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/cfec1dde37af7d97ad1b43d34029a1b183b1e5a6/src/BloomBrowserUI/bookEdit/js/niceScrollBars.ts#L68)):*
> ```XML
>                 // We also need to adjust the height of the rail and the cursor, but nicescroll
>                 // doesn't allow us to adjust the height of the rail.  Adjusting the height of
>                 // the cursor is almost counterproductive if the rail height doesn't change.
> ```

I'm not sure what to make of this comment. Are we doing something to make things right even though nicescroll won't let us adjust the rail? Are we adjusting the cursror only even though it is "almost counterproductive"? What do "cursor" and "rail" even mean? Nicescroll uses "cursor" for "thumb" so I'm guessing you mean that we can adjust the length of the thumb but not the length of the whole scrollbar. If we end up making the thumb bigger than the whole scrollbar, I'd say that would definitely be counterproductive.

JohnThomson (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6781)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @StephenMcConnel)

PR #6784, opened 2025-01-08
Fix possible crash when processing BloomPlayer request (BL-13881)

StephenMcConnel:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6784)


StephenMcConnel:
This fix was submitted independently.

PR #6788, opened 2025-01-14
Fix image copyright button when no copyright/license (BL-14151)

andrew-polk:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6788)

### DIFF
--- a/src/BloomBrowserUI/bookEdit/js/bloomImages.ts
+++ b/src/BloomBrowserUI/bookEdit/js/bloomImages.ts
@@ -77,7 +77,9 @@ export function SetupImagesInContainer(container) {
         });
 }
 
-export function SetupImage(image: JQuery) {
+export function SetupImage(image) {
+    // Caution! image may or may not be a jQuery object.
+
     // Remove any obsolete explicit image size and position left over from earlier versions of Bloom, before we had object-fit:contain.
     // Note: any changes to this should probably also be made to (C#) Book.RemoveObsoleteImageAttributes(), if it still exists.
     if (image.style) {
@@ -109,6 +111,7 @@ export function GetButtonModifier(container) {
     let buttonModifier = "";
     const imageButtonWidth = 87;
     const imageButtonHeight = 52;
+    // container may or may not be a jQuery object already.
     const $container = $(container);
     if ($container.height() < imageButtonHeight * 2) {
         buttonModifier = "smallButtonHeight";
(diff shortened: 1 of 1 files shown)


StephenMcConnel (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6788)* status: all files reviewed, 1 unresolved discussion (waiting on @andrew-polk)

___
*[`src/BloomBrowserUI/bookEdit/js/bloomImages.ts` line 81 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6788#-OGenQeV8Ia9wVXf7bvr:-OGenQeV8Ia9wVXf7bvs:b-vbej7d) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/2d77a637854789d510cced5594da08a8b8104bab/src/BloomBrowserUI/bookEdit/js/bloomImages.ts#L81)):*
> ```XML
> 
> export function SetupImage(image) {
>     // Caution! image may or may not be a jQuery object.
> ```

I assume that image being or not being a jquery object didn't actually cause problems in this particular function?  Is that why you didn't treat it like container in UpdateOverlay?

andrew-polk:
*[`src/BloomBrowserUI/bookEdit/js/bloomImages.ts` line 81 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6788#-OGenQeV8Ia9wVXf7bvr:-OGerlB74ybGFQdKts8K:bg8177p) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/2d77a637854789d510cced5594da08a8b8104bab/src/BloomBrowserUI/bookEdit/js/bloomImages.ts#L81)):*

I assume that image being or not being a jquery object didn't actually cause problems in this particular function?  Is that why you didn't treat it like container in UpdateOverlay?
</blockquote></details>

I was trying to make minimal changes since this is going into 6.1. Ideally we would clean things up to not have this mess of maybe it is, maybe it isn't anywhere.
But yes, as far as I can tell, this isn't causing a current problem.

I discovered the ambiguity in this function because `.closest()` on a jquery object always produces a truthy result (a jquery object). And I thought I had discovered a bug. But apparently that code path is only taken for the case when image is not a jquery object.

StephenMcConnel (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6788)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @andrew-polk)

PR #6571, opened 2024-07-08
Bloom Games project

JohnThomson:
Omits sounds oh-oh, waah, and yay, for which we have not yet found versions we can license.
Also crying boy.jpg and love-fox.gif because I'm not sure we really want to ship those as parts of our template pages.

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6571)

PR #6790, opened 2025-01-14
BL-14167 import recording should reopen to last location

nabalone:
This undoes a change made by JT in https://github.com/BloomBooks/BloomDesktop/pull/6571/files#diff-fda3f19b08588223b7b2ec9253a08eb8a388139f08bbf0023c025233bab01174R105. I talked to him about it and he thought this change would be okay.

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6790)

### DIFF
--- a/src/BloomExe/web/controllers/FileIOApi.cs
+++ b/src/BloomExe/web/controllers/FileIOApi.cs
@@ -102,28 +102,53 @@ namespace Bloom.web.controllers
 
         private string SelectFileUsingDialog(OpenFileRequest requestParameters)
         {
+            // For saving and recalling the last chosen file location
+            // requestParameters.title is something like "Choose Audio File", use as an identifier for recalling this filepath
+            var extraTag = requestParameters.title;
+            var extension = "." + requestParameters.fileTypes[0].extensions[0];
+
             var initialDirectory = string.IsNullOrEmpty(requestParameters.defaultPath)
-                ? System.Environment.GetFolderPath(System.Environment.SpecialFolder.MyDocuments)
-                : Path.GetDirectoryName(requestParameters.defaultPath); // enhance would be better to actually select the file, not just the Dir?
-            using (var dlg = new MiscUI.BloomOpenFileDialog
-            {
-                Title = requestParameters.title,
-                InitialDirectory = initialDirectory,
-                FileName = Path.GetFileName(requestParameters.defaultPath),
-                Filter = string.Join(
-                    "|",
-                    requestParameters.fileTypes.Select(
-                        fileType =>
-                            $"{fileType.name}|{string.Join(";", fileType.extensions.Select(e => "*." + e))}"
+                ? Path.GetDirectoryName(
+                    OutputFilenames.GetOutputFilePath(
+                        CurrentBook,
+                        extension,
+                        extraTag: extraTag,
+                        proposedFolder: System.Environment.GetFolderPath(
+                            System.Environment.SpecialFolder.MyDocuments
+                        )
                     )
-                ),
-            })
+                )
+                : Path.GetDirectoryName(requestParameters.defaultPath); // enhance would be better to actually select the file, not just the Dir?
+
+            using (
+                var dlg = new MiscUI.BloomOpenFileDialog
+                {
+                    Title = requestParameters.title,
+                    InitialDirectory = initialDirectory,
+                    FileName = Path.GetFileName(requestParameters.defaultPath),
+                    Filter = string.Join(
+                        "|",
+                        requestParameters.fileTypes.Select(
+                            fileType =>
+                                $"{fileType.name}|{string.Join(";", fileType.extensions.Select(e => "*." + e))}"
+                        )
+                    ),
+                }
+            )
             {
                 var result = dlg.ShowDialog();
                 if (result == DialogResult.OK)
                 {
                     // We are not trying get a memory or time diff, just a point measure.
                     PerformanceMeasurement.Global.Measure("Choose file", dlg.FileName)?.Dispose();
+
+                    // remember selected filepath for next time
+                    OutputFilenames.RememberOutputFilePath(
+                        CurrentBook,
+                        extension,
+                        dlg.FileName,
+                        extraTag
+                    );
 
                     if (!string.IsNullOrEmpty(requestParameters.destFolder))
                     {
--- a/src/BloomExe/web/controllers/MusicApi.cs
+++ b/src/BloomExe/web/controllers/MusicApi.cs
@@ -6,6 +6,7 @@ using System.Windows.Forms;
 using Bloom.Api;
 using Bloom.Book;
 using Bloom.Publish;
+using Bloom.Utils;
 using L10NSharp;
 using SIL.IO;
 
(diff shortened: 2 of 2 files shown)


src/BloomExe/web/controllers/FileIOApi.cs
```diff
@@ -102,22 +102,21 @@ private void ChooseFile(ApiRequest request)
...
-                        fileType =>
-                            $"{fileType.name}|{string.Join(";", fileType.extensions.Select(e => "*." + e))}"
-                    )
-                ),
-            })
+            using (
+                var dlg = new MiscUI.BloomOpenFileDialog
+                {
+                    Title = requestParameters.title,
+                    InitialDirectory = Path.GetDirectoryName(requestParameters.defaultPath), // enhance would be better to actually select the file, not just the Dir?
```
nabalone, line 109: If InitialDirectory is null, the file chooser opens to wherever it was last opened to. We want that behavior when we don't have an InitialDirectory specified, instead of manually defaulting to MyDocuments

hatton: Can we have both? 
1) remember your last location for this command, at least during a run of Bloom ( this is what the PR does, presumably?)
2) but if this is the first time, then My Documents

I could imagine enhancing (1) to remember where we last were when doing a particular kind of thing. E.g. Where is the last place I imported audio from, vs where is the last place I imported an image from. But that's for another day.

hatton (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6790)* status: all files reviewed (commit messages unreviewed), 1 unresolved discussion (waiting on @nabalone)

src/BloomExe/web/controllers/FileIOApi.cs
andrew-polk: FWIW, we do have a whole system in place like this for remembering output file locations. See OutputFilenames.cs.
It might be as simple as using that as is (though the "Output" part of the name would be wrong).

nabalone: Done

nabalone (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6790)* status: 0 of 2 files reviewed, 2 unresolved discussions (waiting on @andrew-polk and @hatton)

___
*[`src/BloomExe/web/controllers/FileIOApi.cs` line 106 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6790#-OGkufTh0Oq0-_cnCCR6:-OGkufTh0Oq0-_cnCCR7:b-yohsex) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/3f0fbf4b19c45ed5439d874a2c72c1fa001fdadc/src/BloomExe/web/controllers/FileIOApi.cs#L106)):*
> ```Smalltalk
>         {
>             // For saving and recalling the last chosen file location
>             var extraTag = requestParameters.title;
> ```

Identifier for the remembered file location. requestParameters.title is currently unique for each call to this API endpoint: "Choose Audio File", "Choose Timing File" or localized "Choose Sound File". Is it better to use this or add a new optional parameter to the API request?

If there's a remembered path for the book+extension+extraTag it uses that, and if not then it uses the last remembered path for the extension 
___
*[`src/BloomExe/web/controllers/MusicApi.cs` line 98 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6790#-OGkxIMN7JWgGh0DCKVt:-OGkxIMN7JWgGh0DCKVu:b2gl28e) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/3f0fbf4b19c45ed5439d874a2c72c1fa001fdadc/src/BloomExe/web/controllers/MusicApi.cs#L98)):*
> ```Smalltalk
>             );
>             string srcFile = null;
> ```

I happened upon a parallel situation for music selection so made the change here also. Perhaps there are more similar situations out there, I didn't look

andrew-polk (changes requested):
> Reviewed 1 of 2 files at r2, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6790)* status: all files reviewed, 3 unresolved discussions (waiting on @hatton and @nabalone)

___
*[`src/BloomExe/web/controllers/FileIOApi.cs` line 106 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6790#-OGkufTh0Oq0-_cnCCR6:-OGlHlSu2MsUuC723DpK:bmu5vo9) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/3f0fbf4b19c45ed5439d874a2c72c1fa001fdadc/src/BloomExe/web/controllers/FileIOApi.cs#L106)):*

Identifier for the remembered file location. requestParameters.title is currently unique for each call to this API endpoint: "Choose Audio File", "Choose Timing File" or localized "Choose Sound File". Is it better to use this or add a new optional parameter to the API request?

If there's a remembered path for the book+extension+extraTag it uses that, and if not then it uses the last remembered path for the extension
</blockquote></details>

I think title is fine, but it is probably worth trying to capture what it is in comments.
___
*[`src/BloomExe/web/controllers/FileIOApi.cs` line 153 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6790#-OGlGDWi5YMdq6wzb_2k:-OGlGDWi5YMdq6wzb_2l:b-67s1ft) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/66d0f9cc7185e9370c211de62ee562bdb0b77f9e/src/BloomExe/web/controllers/FileIOApi.cs#L153)):*
> ```Smalltalk
>                     if (!string.IsNullOrEmpty(requestParameters.destFolder))
>                     {
>                         // remember selected filename for next time
> ```

Doesn't this comment go with the new call above?

nabalone (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6790)* status: all files reviewed, 2 unresolved discussions (waiting on @andrew-polk and @hatton)

___
*[`src/BloomExe/web/controllers/FileIOApi.cs` line 106 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6790#-OGkufTh0Oq0-_cnCCR6:-OGpvFrsB5kvCjDbZRe-:bb6un1j) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/3f0fbf4b19c45ed5439d874a2c72c1fa001fdadc/src/BloomExe/web/controllers/FileIOApi.cs#L106)):*

I think title is fine, but it is probably worth trying to capture what it is in comments.
</blockquote></details>

Done
___
*[`src/BloomExe/web/controllers/FileIOApi.cs` line 153 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6790#-OGlGDWi5YMdq6wzb_2k:-OGpvEZH0j6PbNrXA4oU:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/66d0f9cc7185e9370c211de62ee562bdb0b77f9e/src/BloomExe/web/controllers/FileIOApi.cs#L153)):*

Doesn't this comment go with the new call above?
</blockquote></details>

Done.

andrew-polk (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6790)* status: all files reviewed, 1 unresolved discussion (waiting on @hatton)

andrew-polk (approved):
> Dismissed @nabalone from a discussion.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6790)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @nabalone)

PR #6794, opened 2025-01-16
Prevent overflow when changing overlay padding (BL-14182)

StephenMcConnel:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6794)

### DIFF
--- a/src/BloomBrowserUI/bookEdit/OverflowChecker/OverflowChecker.ts
+++ b/src/BloomBrowserUI/bookEdit/OverflowChecker/OverflowChecker.ts
@@ -307,7 +307,7 @@ export default class OverflowChecker {
     // Checks for overflow on a bloom-page and adds/removes the proper class
     // N.B. This function is specifically designed to be called from within AddOverflowHandler()
     // but is also called from within StyleEditor (and therefore public)
-    public static MarkOverflowInternal(box) {
+    public static MarkOverflowInternal(box, doNotShrink?: boolean) {
         // There are two types of overflow that we need to check.
         // 1-When we're called by a handler on an element, we need to check that that element
         // doesn't overflow internally (i.e. has too much stuff to fit in itself).
@@ -353,7 +353,9 @@ export default class OverflowChecker {
         const overflowAmounts = OverflowChecker.getSelfOverflowAmounts(box);
         const overflowX = overflowAmounts[0];
         let overflowY = overflowAmounts[1];
-        if (theOneBubbleManager.growOverflowingBox(box, overflowY)) {
+        if (
+            theOneBubbleManager.growOverflowingBox(box, overflowY, doNotShrink)
+        ) {
             overflowY = 0;
         }
         if (preventOverflowY) {
--- a/src/BloomBrowserUI/bookEdit/StyleEditor/StyleEditor.ts
+++ b/src/BloomBrowserUI/bookEdit/StyleEditor/StyleEditor.ts
@@ -1191,7 +1191,7 @@ export default class StyleEditor {
         const rule = this.getStyleRule(true, false);
         if (rule !== null) {
             rule.style.setProperty("padding", padding, "important");
-            this.cleanupAfterStyleChange();
+            this.cleanupAfterStyleChange(true); // do not shrink box that we may have grown
             RenderOverlayRoot(padding, (newPadding: string) => {
                 this.changePadding(newPadding);
             });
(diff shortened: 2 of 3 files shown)


JohnThomson (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6794)* status: all files reviewed, 1 unresolved discussion (waiting on @StephenMcConnel)

___
*[`src/BloomBrowserUI/bookEdit/js/bubbleManager.ts` line 116 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6794#-OGkG4Yo5VqUXsivWpHN:-OGkG4Yo5VqUXsivWpHO:bvn3jmf) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/534a0e2c4876ff24b8597e3a7dc90ee53da8381b/src/BloomBrowserUI/bookEdit/js/bubbleManager.ts#L116)):*
> ```XML
>         if (wrapperBox.classList.contains("bloom-paddingChanged")) {
>             wrapperBox.classList.remove("bloom-paddingChanged");
>             return false; // We don't need to fix it.  See BL-14182.
> ```

It's not just that we don't need to fix it...from your comments this morning I gather that if we do try to fix it here in this situation, we will mess things up. I think you said the following code would shrink the box back down in this situation, causing overflow. Can you explain why the code would do the wrong thing in this case? We'll only avoid running the code here if we've recently done adjustOverlaysForPaddingChange; shouldn't that be using the same algorithm to figure out the size the overlay needs to be? IOW, if the code here sometimes makes the overlay too small, why does it only do so immediately after adjusting padding?
It also bothers me that this will suppress exactly the first call to growOverflowingBox after adjusting padding. Can we be quite sure that whatever would make it do the wrong thing is no longer the case on the following call (e.g., if the user types after changing padding)? Can we be sure that it won't be called twice while the problem situation applies? Can we be sure that it WILL be called once, so that the flag class gets removed and it won't be suppressed when needed later?
At least, I think the fix deserves comments addressing these points, and I'd feel happier with a solution that got to the root of why this function would sometimes do the wrong thing, and fixed that. But I expect you tried and there's some problem that makes it difficult.

StephenMcConnel (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6794)* status: all files reviewed, 1 unresolved discussion (waiting on @JohnThomson)

___
*[`src/BloomBrowserUI/bookEdit/js/bubbleManager.ts` line 116 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6794#-OGkG4Yo5VqUXsivWpHN:-OGkuXmf4R0ls24PhPvg:b-62c2n4) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/534a0e2c4876ff24b8597e3a7dc90ee53da8381b/src/BloomBrowserUI/bookEdit/js/bubbleManager.ts#L116)):*

It's not just that we don't need to fix it...from your comments this morning I gather that if we do try to fix it here in this situation, we will mess things up. I think you said the following code would shrink the box back down in this situation, causing overflow. Can you explain why the code would do the wrong thing in this case? We'll only avoid running the code here if we've recently done adjustOverlaysForPaddingChange; shouldn't that be using the same algorithm to figure out the size the overlay needs to be? IOW, if the code here sometimes makes the overlay too small, why does it only do so immediately after adjusting padding?
It also bothers me that this will suppress exactly the first call to growOverflowingBox after adjusting padding. Can we be quite sure that whatever would make it do the wrong thing is no longer the case on the following call (e.g., if the user types after changing padding)? Can we be sure that it won't be called twice while the problem situation applies? Can we be sure that it WILL be called once, so that the flag class gets removed and it won't be suppressed when needed later?
At least, I think the fix deserves comments addressing these points, and I'd feel happier with a solution that got to the root of why this function would sometimes do the wrong thing, and fixed that. But I expect you tried and there's some problem that makes it difficult.
</blockquote></details>

After studying the flow of control through stack traces, I determined that we can pass an (optional) parameter through a relatively short chain of function calls to achieve the fix.  I tried the other approach because I wasn't sure this was possible.  I was probably too tired after looking at this issue all day to think of looking at the stack traces at the salient points.
I hope this new approach passes muster.

JohnThomson (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6794)* status: all files reviewed, 1 unresolved discussion (waiting on @StephenMcConnel)

___
*[`src/BloomBrowserUI/bookEdit/js/bubbleManager.ts` line 116 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6794#-OGkG4Yo5VqUXsivWpHN:-OGor9rO1Lu0rkdL9Ady:b-80yfpy) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/534a0e2c4876ff24b8597e3a7dc90ee53da8381b/src/BloomBrowserUI/bookEdit/js/bubbleManager.ts#L116)):*

After studying the flow of control through stack traces, I determined that we can pass an (optional) parameter through a relatively short chain of function calls to achieve the fix.  I tried the other approach because I wasn't sure this was possible.  I was probably too tired after looking at this issue all day to think of looking at the stack traces at the salient points.
I hope this new approach passes muster.
</blockquote></details>

Makes me rather embarrassed that this routine called "growOverflowingBox" EVER shrinks the box. It wants a better name. But I can't think of one, and your change makes it more obvious rather than less that it sometimes does the opposite of what it promises.

PR #6811, opened 2025-01-23
BloomDesktop WiFi issue: force advertising to use proper network interface

wmergenthal:


wmergenthal:
I really wanted to get this Pull Request underway. Chris Hubbard helped me do so. I thought that commit squashing was something I needed to do ahead of the PR, but he showed me that the changes (in 21 commits) can be squashed at the last moment with a "squash and merge" action.
If any of this is incompatible with Bloom procedures, please let me know.

andrew-polk:
Sorry this has taken so long to get looked at. 
I've now created a [card](https://issues.bloomlibrary.org/youtrack/issue/BL-14477/Wades-change-for-better-Wifi-transfer-stability) for it, so it should get reviewed now.

andrew-polk:
Actually, I just noticed this is marked as draft. Is it ready for review?

JohnThomson:
Just responding to your description so far, since as Andrew noted, it's not clear this is ready for review, being marked draft. But what makes the IP address "that can reach the internet" the one we want? For one thing, it's possible that in a village situation there is NO IP address that can reach the internet. I could also imagine that a desktop (or even laptop) might have an Ethernet connection to the internet, but also be on a purely local WiFi network (maybe its own hotspot) which can't reach the internet but which is the one the phone can access. If things get complicated in a campus network, one approach is to disconnect from the main network (and internet), and get the two devices talking on a hotspot WiFi that one of them creates. Again that may NOT be able to reach the internet.

wmergenthal:
Sorry for the long delay. I did some further testing and encountered a failure. From there I got into a repeating cycle of finding root cause, making a fix, testing the fix which turned up a different failure, dig into that, etc, etc.
I think I have accounted for the shortcomings now and am not seeing further failures. John, you are right that an IP address "that can reach the internet" may not be the right one. I have since learned that the PC's network stack will use the interface that has the lowest "interface metric." The new changeset includes this.
The new changeset is also on a different branch, so I will close this PR. Tomorrow I expect to open a new PR with the new branch.
Again, sorry for the long delay on this.

PR #6812, opened 2025-01-23
Make thumbnails in collision dialog more uniform (BL-13769)

StephenMcConnel:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6812)

### DIFF
--- a/src/BloomBrowserUI/react_components/bookInfoCard.tsx
+++ b/src/BloomBrowserUI/react_components/bookInfoCard.tsx
@@ -108,8 +108,8 @@ export const BookInfoCard: React.FunctionComponent<IBookInfoCardProps> = props =
                             >
                                 <img
                                     css={css`
-                                        width: 70px;
-                                        min-height: 50px;
+                                        max-width: 70px;
+                                        max-height: 70px;
                                     `}
                                     src={props.thumbnailUrl}
                                 />
--- a/src/BloomExe/ImageProcessing/RuntimeImageProcessor.cs
+++ b/src/BloomExe/ImageProcessing/RuntimeImageProcessor.cs
@@ -267,7 +267,8 @@ namespace Bloom.ImageProcessing
             string pathToProcessedImage,
             int thumbnailWidth,
             int thumbnailHeight,
-            Color backColor
+            Color backColor,
+            bool padImageToRequestedSize = true
         )
         {
             using (var coverImage = PalasoImage.FromFileRobustly(coverImagePath))
@@ -298,6 +299,11 @@ namespace Bloom.ImageProcessing
                 // pad to center the cover image
                 var horizontalPadding = (availableThumbnailWidth - targetImageWidth) / 2;
                 var verticalPadding = (availableThumbnailHeight - targetImageHeight) / 2;
+                if (!padImageToRequestedSize)
+                {
+                    horizontalPadding = 0;
+                    verticalPadding = 0;
+                }
                 var destRect = new Rectangle(
                     kborder + horizontalPadding,
                     kborder + verticalPadding,
@@ -310,7 +316,7 @@ namespace Bloom.ImageProcessing
 
                 Rectangle backgroundAndBorderRect;
                 var appearsToBeJpeg = ImageUtils.AppearsToBeJpeg(coverImage);
-                if (appearsToBeJpeg)
+                if (appearsToBeJpeg || !padImageToRequestedSize)
                 {
                     backgroundAndBorderRect = destRect;
                     backgroundAndBorderRect.Inflate(kborder * 2, kborder * 2);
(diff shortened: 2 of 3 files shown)


JohnThomson (approved):
I'm approving this because the problems are very minor and you may not like either of my name suggestions. But I won't merge till you have a chance to think about them.

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6812#-:-OHPa9_N86-yAnjAZmqi:bygh95n)* status: all files reviewed, 2 unresolved discussions (waiting on @StephenMcConnel)

___
*[`src/BloomExe/ImageProcessing/RuntimeImageProcessor.cs` line 271 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6812#-OHPWnarEeUn95wnIBcD:-OHPWnarEeUn95wnIBcE:bpnpqx2) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/c1b9754b9832f5ecb0e21dec08e8957fbbf4ad96/src/BloomExe/ImageProcessing/RuntimeImageProcessor.cs#L271)):*
> ```Smalltalk
>             int thumbnailHeight,
>             Color backColor,
>             bool preserveAspectRatio = false
> ```

This name threw me at first. I could not imagine a case where we don't want to preserve the aspect ratio of the image itself. Then I figured out that the distinction is whether we produce an image exactly the size requested (with padding as needed to preserve aspect ratio) or one that is at most the size requested, but typically smaller in one dimension so as to preserve the aspect ratio without padding.
I thought of "padToRequestedSize" (though that would have the opposite meaning and default) or "allowSmallerImageInOneDirection".
___
*[`src/BloomExe/ImageProcessing/RuntimeImageProcessor.cs` line 319 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6812#-OHPZZV6EHxyNvvDH02-:-OHPZZV77sm0HKOT-4UN:b-ontooj) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/c1b9754b9832f5ecb0e21dec08e8957fbbf4ad96/src/BloomExe/ImageProcessing/RuntimeImageProcessor.cs#L319)):*
> ```Smalltalk
>                 Rectangle backgroundAndBorderRect;
>                 var appearsToBeJpeg = ImageUtils.AppearsToBeJpeg(coverImage);
>                 if (appearsToBeJpeg || preserveAspectRatio)
> ```

This is a pre-existing problem...but it's not at all obvious why we want an aspect-ratio-sized rectangle if the image "appearsToBeJpeg". Is the idea that even if we're making the thumbnail bigger than the image in one direction, we don't want to fill all the rest of it with the background? Why would that be wanted for jpgs but not pngs? (Of course you may not know...but if you do a comment would be helpful.)

StephenMcConnel (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6812)* status: all files reviewed, 2 unresolved discussions (waiting on @JohnThomson)

___
*[`src/BloomExe/ImageProcessing/RuntimeImageProcessor.cs` line 271 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6812#-OHPWnarEeUn95wnIBcD:-OHdKOAnEXdS93LLLXbe:bvwfgoq) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/c1b9754b9832f5ecb0e21dec08e8957fbbf4ad96/src/BloomExe/ImageProcessing/RuntimeImageProcessor.cs#L271)):*

This name threw me at first. I could not imagine a case where we don't want to preserve the aspect ratio of the image itself. Then I figured out that the distinction is whether we produce an image exactly the size requested (with padding as needed to preserve aspect ratio) or one that is at most the size requested, but typically smaller in one dimension so as to preserve the aspect ratio without padding.
I thought of "padToRequestedSize" (though that would have the opposite meaning and default) or "allowSmallerImageInOneDirection".
</blockquote></details>

Done.  changed the parameter name to padIamgeToRequestedSize with a default of true.
___
*[`src/BloomExe/ImageProcessing/RuntimeImageProcessor.cs` line 319 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6812#-OHPZZV6EHxyNvvDH02-:-OHdKXUqC2h2nmqBBkeU:b-k0f5of) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/c1b9754b9832f5ecb0e21dec08e8957fbbf4ad96/src/BloomExe/ImageProcessing/RuntimeImageProcessor.cs#L319)):*

This is a pre-existing problem...but it's not at all obvious why we want an aspect-ratio-sized rectangle if the image "appearsToBeJpeg". Is the idea that even if we're making the thumbnail bigger than the image in one direction, we don't want to fill all the rest of it with the background? Why would that be wanted for jpgs but not pngs? (Of course you may not know...but if you do a comment would be helpful.)
</blockquote></details>

I don't know the reason why this is doing different things for jpegs, so I tried to preserve the behavior for jpeg images.

JohnThomson (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6812#-:-OHjBzdU6yH2O3Yl4AKK:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @StephenMcConnel)

PR #6816, opened 2025-01-27
Optimize overflow checks, especially in origami drags (BL-14234)

JohnThomson:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6816)


JohnThomson:
Converted to draft until we release 6.1 beta. We want to do some testing in beta internal before releasing this patch to full beta.

PR #6830, opened 2025-01-30
Restore reasonable display for missing images (BL-14241)

StephenMcConnel:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6830)

### DIFF
--- a/src/BloomBrowserUI/bookEdit/css/editMode.less
+++ b/src/BloomBrowserUI/bookEdit/css/editMode.less
@@ -476,6 +476,14 @@ body:not(.origami-drag) .bloom-imageContainer.hoverUp {
         background-image: url("/bloom/bookEdit/img/imageMissingMetaData.svg") !important;
     }
 }
+// In an active background overlay, don't show the edit metadata icon if the image failed
+// to load.  Clicking on it would just show an error message anyway, and the button would
+// obscure part of the alt attribute error message.
+.bloom-backgroundImage:has(img.bloom-imageLoadError) {
+    .editMetadataButton.imgMetadataProblem {
+        display: none !important;
+    }
+}
 
 // we need to inset the buttons a bit when the full bleed borders are showing
 body.bloom-fullBleed {
--- a/src/BloomBrowserUI/bookEdit/js/bloomEditing.ts
+++ b/src/BloomBrowserUI/bookEdit/js/bloomEditing.ts
@@ -4,6 +4,7 @@ import * as $ from "jquery";
 import bloomQtipUtils from "./bloomQtipUtils";
 import {
     cleanupImages,
+    HandleImageError,
     SetupMetadataButton,
     SetupResizableElement,
     SetupImagesInContainer,
@@ -405,6 +406,12 @@ export function changeImage(imageInfo: {
     // has an <img> element, we're setting the src on that. But if it does not, we're
     // setting the background-image on the container itself.
     if (imgOrImageContainer.tagName === "IMG") {
+        // We need to reset the image load error flag for a new image.  We also want to
+        // install an error handler to set the flag if needed.  See BL-14241.
+        (imgOrImageContainer as HTMLImageElement).classList.remove(
+            "bloom-imageLoadError"
+        );
+        (imgOrImageContainer as HTMLImageElement).onerror = HandleImageError;
         (imgOrImageContainer as HTMLImageElement).src = imageInfo.src;
     }
     // else if it has class bloom-imageContainer, we need to set the background-image on the container
--- a/src/BloomBrowserUI/bookEdit/js/bloomImages.ts
+++ b/src/BloomBrowserUI/bookEdit/js/bloomImages.ts
@@ -110,6 +110,33 @@ export function SetupImage(image) {
         image.removeAttribute("width");
         image.removeAttribute("height");
     }
+    var hndlr = image.getAttribute("onerror");
+    if (!hndlr) {
+        // Cover images get a handler assigned in C# code because assigning it here
+        // doesn't work for the initial load on cover pages.  (The image resource load
+        // fails before this bootstrap code is called in document.ready(), but only
+        // for the cover image for some reason.)  The C# code also doesn't preserve
+        // the bloom-imageLoadError class from earlier editing sessons.
+        // The onerror handler can be null if this is not a cover image and the error
+        // handler has not yet been assigned.  In that case, we want to initialize the
+        // error handler and class.  See BL-14241.
+        // Note that the error handler assigned this way will not be persisted as an
+        // attribute in the image element in the HTML.
+        image.classList.remove("bloom-imageLoadError");
+        image.onerror = HandleImageError;
+    }
+}
+
+// This handler behaves exactly the same as the one assigned in C# code to the
+// cover image elementin BookData.cs/UpdateImageFromDataSet().  Any change here
+// should be reflected there as well.  This might be difficult since the C# code
+// adds the very simple "this.classList.add('bloom-imageLoadError');" as the content
+// of the onerror attribute.  Anything much more complicated would probably require
+// more javascript work to bundle up the error handler appropriately for access.
+export function HandleImageError(event: Event) {
+    const target = event.target as HTMLImageElement;
+    target.classList.add("bloom-imageLoadError");
+    // console.error("Image failed to load:", target.src);
 }
 
 export function GetButtonModifier(container) {
@@ -795,6 +822,11 @@ function SetAlternateTextOnImages(element) {
             "This picture, {0}, is missing or was loading too slowly."; // Also update HtmlDom.cs::IsPlaceholderImageAltText
         const nameWithoutQueryString = GetRawImageUrl(element).split("?")[0];
         const decodedName = decodeURI(nameWithoutQueryString);
+        // show English to start with in case localization never returns even a failure
+        $(element).attr(
+            "alt",
+            theOneLocalizationManager.simpleFormat(englishText, [decodedName])
+        );
         theOneLocalizationManager
             .asyncGetText(
                 "EditTab.Image.AltMsg",
--- a/src/BloomBrowserUI/bookEdit/js/bubbleManager.ts
+++ b/src/BloomBrowserUI/bookEdit/js/bubbleManager.ts
@@ -28,7 +28,8 @@ import {
     kImageContainerClass,
     getBackgroundImageFromContainer,
     SetupMetadataButton,
-    UpdateImageTooltipVisibility
+    UpdateImageTooltipVisibility,
+    HandleImageError
 } from "./bloomImages";
 import { adjustTarget } from "../toolbox/dragActivity/dragActivityTool";
 import BloomSourceBubbles from "../sourceBubbles/BloomSourceBubbles";
@@ -4224,6 +4225,8 @@ export class BubbleManager {
             // just revert it to a placeholder
             const img = getImageFromOverlay(textOverPicDiv);
             if (img) {
+                img.classList.remove("bloom-imageLoadError");
+                img.onerror = HandleImageError;
                 img.src = "placeHolder.png";
                 this.updateBubbleForChangedImage(img);
             }
@@ -5196,6 +5199,7 @@ export class BubbleManager {
             newImgContainer.classList.remove(kOverlayClass);
             bgOverlay.appendChild(newImgContainer);
             const newImg = img.cloneNode(false) as HTMLImageElement;
+            newImg.classList.remove("bloom-imageLoadError");
             newImgContainer.appendChild(newImg);
 
             // Set level so Comical will consider the new overlay to be under the existing ones.
@@ -5233,6 +5237,8 @@ export class BubbleManager {
             imageContainer
         ) as HTMLElement; // must exist by now
         // Whether it's a new bgImage or not, copy its src from the old-style img
+        bgImage.classList.remove("bloom-imageLoadError");
+        bgImage.onerror = HandleImageError;
         bgImage.setAttribute("src", img.getAttribute("src") ?? "");
         this.adjustBackgroundImageSize(imageContainer, bgOverlay, true);
         bgOverlay.style.visibility = ""; // now we can show it, if it was new and hidden
@@ -5266,6 +5272,21 @@ export class BubbleManager {
         );
     }
 
+    // Given a bgOverlay element, which is always an overlay having the bloom-backgroundImage
+    // class, and the height and width of the parent imageContainer, this method attempts to
+    // make the bgOverlay the right size and position to fill as much as possible of the parent,
+    // rather like object-fit:contain. It is used in two main scenarios: the user may have
+    // selected a different image, which means we must adjust to suit a different image aspect
+    // ratio. Or, the size of the container may have changed, e.g., using origami. We must also
+    // account for the possibility that the image has been cropped, in which case, we want to
+    // keep the cropped aspect ratio. (Cropping attributes will already have been removed if it
+    // is a new image.)
+    // Things are complicated because it's possible the image has not loaded yet, so we can't
+    // get its natural dimensions to figure an aspect ratio. In this case, the method arranges
+    // to be called again after the image loads or a timeout.
+    // A further complication is that the image may fail to load, so we never get natural
+    // dimensions. In this case, we expand the bgOverlay to the full size of the container so
+    // all the space is available to display the error icon and message.
     private adjustBackgroundImageSizeToFit(
         containerWidth: number,
         containerHeight: number,
@@ -5292,22 +5313,34 @@ export class BubbleManager {
         }
         let imgAspectRatio = bgOverlay.clientWidth / bgOverlay.clientHeight;
         const img = getImageFromOverlay(bgOverlay);
+        let failedImage = false;
+        // We don't ever expect there not to be an img. If it happens, we'll just go
+        // ahead and adjust based on the current shape of the overlay (as set above).
         if (img) {
-            // We don't ever expect there not to be an img. If it happens, we'll just go
-            // ahead and adjust based on the current shape of the overlay.
-            // if we don't have a height and width, or we know the image src changed
-            // and have not yet waited for new dimensions, go ahead and wait.
-            if (img.style.width) {
-                // there is established cropping. Use the cropped size to determine the
-                // aspect ratio.
-                imgAspectRatio =
-                    BubbleManager.pxToNumber(bgOverlay.style.width) /
-                    BubbleManager.pxToNumber(bgOverlay.style.height);
+            // The image may not have loaded yet or may have failed to load.  If either of these
+            // cases is true, then the naturalHeight and naturalWidth will be zero.  If the image
+            // failed to load, a special class is added to the image to indicate this fact (if all
+            // goes well).  However, we may know that this is called in response to a new image, in
+            // which case the class may not have been added yet.
+            // We conclude that the image has truly failed if 1) we don't have natural dimensions set
+            // to something other than zero, 2) we are not waiting for new dimensions, and 3) the
+            // image has the special class indicating that it failed to load.  (The class is supposed
+            // to be removed when we change the src attribute, which leads to a new load attempt.)
+            failedImage =
+                img.naturalHeight === 0 && // not loaded successfully (yet)
+                !useSizeOfNewImage && // not waiting for new dimensions
+                img.classList.contains("bloom-imageLoadError"); // error occurred while trying to load
+            if (failedImage) {
+                // If the image failed to load, just use the container aspect ratio to fill up
+                // the container with the error message (alt attribute string).
+                imgAspectRatio = containerWidth / containerHeight;
             } else if (
                 img.naturalHeight === 0 ||
                 img.naturalWidth === 0 ||
                 useSizeOfNewImage
             ) {
+                // if we don't have a height and width, or we know the image src changed
+                // and have not yet waited for new dimensions, go ahead and wait.
                 // We set up this timeout
                 const handle = (setTimeout(
                     () =>
@@ -5340,6 +5373,12 @@ export class BubbleManager {
                     { once: true }
                 );
                 return; // try again once we have valid image data
+            } else if (img.style.width) {
+                // there is established cropping. Use the cropped size to determine the
+                // aspect ratio.
+                imgAspectRatio =
+                    BubbleManager.pxToNumber(bgOverlay.style.width) /
+                    BubbleManager.pxToNumber(bgOverlay.style.height);
             } else {
                 // not cropped, so we can use the natural dimensions
                 imgAspectRatio = img.naturalWidth / img.naturalHeight;
@@ -5371,6 +5410,27 @@ export class BubbleManager {
                 BubbleManager.pxToNumber(img.style.left) * scale + "px";
             img.style.top =
                 BubbleManager.pxToNumber(img.style.top) * scale + "px";
+        }
+        // Ensure that the missing image message is displayed without being cropped.
+        // See BL-14241.
+        if (failedImage && img && img.style && img.style.width.length > 0) {
+            const imgLeft = BubbleManager.pxToNumber(img.style.left);
+            const imgTop = BubbleManager.pxToNumber(img.style.top);
+            if (imgLeft < 0 || imgTop < 0) {
+                // The failed image was cropped. Remove the cropping to facilitate displaying the error state.
+                img.setAttribute(
+                    "data-style",
+                    `left:${img.style.left}; width:${img.style.width}; top:${img.style.top};`
+                );
+                const imgWidth = BubbleManager.pxToNumber(img.style.width);
+                console.warn(
+                    `Missing image: resetting left from ${imgLeft} to 0, top from ${imgTop} to 0, and width from ${imgWidth} to ${imgWidth +
+                        imgLeft}`
+                );
+                img.style.left = "0px";
+                img.style.top = "0px";
+                img.style.width = imgWidth + imgLeft + "px";
+            }
         }
         this.alignControlFrameWithActiveElement();
     }
--- a/src/BloomExe/Book/BookData.cs
+++ b/src/BloomExe/Book/BookData.cs
@@ -1836,6 +1836,19 @@ namespace Bloom.Book
             var oldImageUrl = HtmlDom.GetImageElementUrl(node);
             var imgOrDivWithBackgroundImage = node;
             HtmlDom.SetImageElementUrl(imgOrDivWithBackgroundImage, newImageUrl);
+            // Attaching the onerror handler in javascript is too late for the cover page
+            // image.  The image resource load fails before the bootstrap code is called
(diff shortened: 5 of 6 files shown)


JohnThomson (changes requested):
I think you're headed in a good direction now, but I still have quite a few suggestions and questions.
Something to test: what happens if you have overlays present above a background image, then delete the image file, then resize the image container? Code will be trying to scale everything based on the size that AdjustBackgroundImage makes the background image. I'm not quite sure what to hope for, but it shouldn't be too horrendous.
Another place we wait for image dimensions and expect to get them eventually is setting the image for an image overlay. Since we've just set the image, presumably we don't have to worry about a missing file, but it could be corrupt. While you're looking at this area, would you make sure something reasonable happens in that case, too?

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6830#-:-OIIEtEP5DzK9O2MgIaL:b-lm3k1u)* status: all files reviewed, 13 unresolved discussions (waiting on @StephenMcConnel)

___
*[`src/BloomBrowserUI/bookEdit/js/bubbleManager.ts` line 5269 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6830#-OIIBWzO7zQg0mhribGZ:-OIIBWzO7zQg0mhribG_:bpzroou) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/9b20ec327915150547743f1f2b957e91d2d71884/src/BloomBrowserUI/bookEdit/js/bubbleManager.ts#L5269)):*
> ```XML
>     }
> 
>     private adjustBackgroundImageSizeToFit(
> ```

I may be missing the point of your changes here. It's probably time to give this method an overall comment (I should have written most of it before), along the lines
"Given a bgOverlay element, which is always an overlay having the bloom-backgroundImage class, and the height and width of the parent imageContainer, this method attempts to make the bgOverlay the right size and position to fill as much as possible of the parent, rather like object-fit:contain. It is used in two main scenarios: the user may have selected a different image, which means we must adjust to suit a different image aspect ratio. Or, the size of the container may have changed, e.g., using origami. We must also account for the possibility that the image has been cropped, in which case, we want to keep the cropped aspect ratio. (Cropping attributes will already have been removed if it is a new image.)
Things are complicated because it's possible the image has not loaded yet, so we can't get its natural dimensions to figure an aspect ratio. In this case, the method arranges to be called again after the image loads or a timeout.
A further complication is that the image may fail to load, so we never get natural dimensions. In this case, we [is this right?] expand the bgOverlay to the full size of the container so all the space is available to display the error icon and message.
___
*[`src/BloomBrowserUI/bookEdit/js/bubbleManager.ts` line 5298 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6830#-OIIA8k9E-gCRdf6dOu5:-OIIA8k9E-gCRdf6dOu6:bpax4q5) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/9b20ec327915150547743f1f2b957e91d2d71884/src/BloomBrowserUI/bookEdit/js/bubbleManager.ts#L5298)):*
> ```XML
>         if (img) {
>             missingImage =
>                 img.naturalHeight === 0 &&
> ```

Not getting the point of this. How can it have a naturalHeight if it failed to load, and why would we want to ignore the failed load if we somehow have a naturalHeight anyway?
___
*[`src/BloomBrowserUI/bookEdit/js/bubbleManager.ts` line 5309 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6830#-OII9MXI90HT7g-HDyR8:-OII9MXJ4gjJkGWs2xVe:b-pbsayp) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/9b20ec327915150547743f1f2b957e91d2d71884/src/BloomBrowserUI/bookEdit/js/bubbleManager.ts#L5309)):*
> ```XML
>                 imgAspectRatio =
>                     BubbleManager.pxToNumber(bgOverlay.style.width) /
>                     BubbleManager.pxToNumber(bgOverlay.style.height);
> ```

I doubt we want to keep the bg overlay shape that was produced by cropping if we no longer have the image and are going to remove the cropping?
___
*[`src/BloomBrowserUI/bookEdit/js/bubbleManager.ts` line 5346 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6830#-OII4SZo7tNhyEdheNWr:-OII4SZo7tNhyEdheNWs:bjdm1ad) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/9b20ec327915150547743f1f2b957e91d2d71884/src/BloomBrowserUI/bookEdit/js/bubbleManager.ts#L5346)):*
> ```XML
>                     { once: true }
>                 );
>                 return; // try again once we have valid image data
> ```

The point of this return is to try this call again later in the expectation that eventually it will load and we will get dimensions. But if bloom-imageLoadError is present, then presumably, we never will. So we had better adjust it to something that will show the error message now, hadn't we? I'm thinking that if we know we're never going to get image dimensions, then we may as well match the overlay dimensions to the container, which I think will happen if we set imgAspectRatio to containerWidth/containerHeight.
___
*[`src/BloomBrowserUI/bookEdit/js/bubbleManager.ts` line 5355 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6830#-OII8B5M0IPs9q9AI8Fu:-OII8B5M0IPs9q9AI8Fv:b-vkltz2) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/9b20ec327915150547743f1f2b957e91d2d71884/src/BloomBrowserUI/bookEdit/js/bubbleManager.ts#L5355)):*
> ```XML
>         const oldWidth = bgOverlay.clientWidth;
>         const containerAspectRatio = containerWidth / containerHeight;
>         if (imgAspectRatio > containerAspectRatio) {
> ```

I'm not sure how we get here if the img is missing (see comment below). But if we do, what imgAspectRatio do we want to use? If we can't match the real image aspect ratio I'd think we should use the container aspect ratio rather than the current default of keeping the old BG overlay aspect ratio?
___
*[`src/BloomBrowserUI/bookEdit/js/bubbleManager.ts` line 5379 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6830#-OII7cB-2r_4idN1gi54:-OII7cB-2r_4idN1gi55:b-gy5so6) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/9b20ec327915150547743f1f2b957e91d2d71884/src/BloomBrowserUI/bookEdit/js/bubbleManager.ts#L5379)):*
> ```XML
>                 BubbleManager.pxToNumber(img.style.top) * scale + "px";
>         }
>         // Ensure that the missing image message is displayed without being cropped.
> ```

This must somehow work, but I don't understand how we even get here if the image is missing. If it's not cropped, image.style.width will be zero, which takes us to the second branch above, where presumably if we can't load it, naturalWidth and naturalHeight are both zero, at which point we set up load and timeout event handlers and return. Load will never happen, and the timeout call will just find that naturalHeight and width are still zero. Or do we get some naturalWidth and height somehow, maybe the shape of the missing image icon?? Anyway, I think a comment is indicated somewhere.
___
*[`src/BloomBrowserUI/bookEdit/js/bubbleManager.ts` line 5382 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6830#-OII8xBe-JtzxLipxm5_:-OII8xBfC22kdYvovCze:b-vhsts4) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/9b20ec327915150547743f1f2b957e91d2d71884/src/BloomBrowserUI/bookEdit/js/bubbleManager.ts#L5382)):*
> ```XML
>         // See BL-14241.
>         if (missingImage && img && img.style && img.style.width.length > 0) {
>             const imgLeft = BubbleManager.pxToNumber(img.style.left);
> ```

comment (I think): the missing image was cropped. Remove the cropping to facilitate displaying the error state.
___
*[`src/BloomBrowserUI/bookEdit/js/bubbleManager.ts` line 5393 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6830#-OII9A60ALfvpG90MouY:-OII9A60ALfvpG90MouZ:b-9sg9tl) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/9b20ec327915150547743f1f2b957e91d2d71884/src/BloomBrowserUI/bookEdit/js/bubbleManager.ts#L5393)):*
> ```XML
>                 //         imgLeft}`
>                 // );
>                 img.style.left = "0px";
> ```

probably clear top too? We don't set height on the img, but top is used for cropping on the top edge.
___
*[`src/BloomBrowserUI/bookEdit/js/bloomImages.ts` line 114 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6830#-OIIDxRU9TExzT7a8KQT:-OIIDxRU9TExzT7a8KQU:b-xvtlao) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/9b20ec327915150547743f1f2b957e91d2d71884/src/BloomBrowserUI/bookEdit/js/bloomImages.ts#L114)):*
> ```XML
>     }
>     var hndlr = image.getAttribute("onerror");
>     if (!hndlr) {
> ```

I'm not sure exactly when this gets called. Would it ever be when hndlr is null not because C# didn't set it, but because HandleImageError cleared it?
___
*[`src/BloomBrowserUI/bookEdit/js/bloomImages.ts` line 117 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6830#-OIIGusF7Spmk_OkFse3:-OIIGusG5ENjN_MNPpQu:b-qg6ea) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/9b20ec327915150547743f1f2b957e91d2d71884/src/BloomBrowserUI/bookEdit/js/bloomImages.ts#L117)):*
> ```XML
>         // Cover images get a handler assigned in C# code, and also don't
>         // preserve the bloom-imageLoadError class.  Assigning the handler
>         // happens too late for the initial load on cover pages.  See BL-14241.
> ```

"too late" sounds bad. Does that mean this fix won't work and we won't see the error properly if the cover image is missing?
___
*[`src/BloomBrowserUI/bookEdit/js/bloomImages.ts` line 120 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6830#-OII6a1B9SVU7rchvG2o:-OII6a1C896zNX99_6dx:b-ei25z7) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/9b20ec327915150547743f1f2b957e91d2d71884/src/BloomBrowserUI/bookEdit/js/bloomImages.ts#L120)):*
> ```XML
>         image.classList.remove("bloom-imageLoadError");
>         image.onerror = HandleImageError;
>     }
> ```

I'm not sure I understand the problem here, but if part of it is contention for the onerror attribute, I think you can just addEventListener("error", HandleImageError), and that will not conflict with whatever the C# puts in onerror. Also, you can add a third argument {once: true} and not have to worry about removing it.
(If you change this here, probably should make the same change where you handle new image above.)
___
*[`src/content/bookLayout/basePage-sharedRules.less` line 60 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6830#-OIIAu3k5rJYUV1wTNwz:-OIIAu3l9BUNkGuV4hYc:bkb451b) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/9b20ec327915150547743f1f2b957e91d2d71884/src/content/bookLayout/basePage-sharedRules.less#L60)):*
> ```Less
>     width: 100%;
>     height: 100%;
> }
> ```

Don't see how this can work. The adjustBackgroundImage code is still setting explicit height and width on the bloom-backgroundImage object, and style attribute values always beat CSS rules. Hopefully we can get those right in that method and not need this?
___
*[`src/content/bookLayout/basePage-sharedRules.less` line 65 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6830#-OIIBCcq4Jirxe0m5_Na:-OIIBCcq4Jirxe0m5_Nb:bhm7lw4) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/9b20ec327915150547743f1f2b957e91d2d71884/src/content/bookLayout/basePage-sharedRules.less#L65)):*
> ```Less
>     #overlay-control-frame {
>     height: 100%;
> }
> ```

I think we shouldn't need this...if we're making the bg overlay the full size of the imageContainer, normal alignment code should make the overlay control frame that size, too.

StephenMcConnel (commented):
One question that remains in my mind is whether or not the .bloom-imageLoadError classes should be persisted in the book's html.  As the code stands, this class remains permanently in the book for each failed image until the user explicitly changes (or deletes) the image.

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6830#-:-OIMU6cM0u1ydCChOdl0:bsuwhx6)* status: all files reviewed, 13 unresolved discussions (waiting on @JohnThomson)

___
*[`src/BloomBrowserUI/bookEdit/js/bloomImages.ts` line 114 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6830#-OIIDxRU9TExzT7a8KQT:-OIIKLzl2TmAt8GaX7JT:b-hc5fqa) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/9b20ec327915150547743f1f2b957e91d2d71884/src/BloomBrowserUI/bookEdit/js/bloomImages.ts#L114)):*

I'm not sure exactly when this gets called. Would it ever be when hndlr is null not because C# didn't set it, but because HandleImageError cleared it?
</blockquote></details>

Two cases: it's not a cover image so C# didn't set it, and it was previously missing possibly because HandleImageError cleared it already but possibly because it had never been set.  You came up with both.  :-)  I've expanded the comment below.
___
*[`src/BloomBrowserUI/bookEdit/js/bloomImages.ts` line 117 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6830#-OIIGusF7Spmk_OkFse3:-OIIKgWT-cgT7Y3MWITz:b-l56est) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/9b20ec327915150547743f1f2b957e91d2d71884/src/BloomBrowserUI/bookEdit/js/bloomImages.ts#L117)):*

"too late" sounds bad. Does that mean this fix won't work and we won't see the error properly if the cover image is missing?
</blockquote></details>

I should clarify that assigning the handler in C# for cover pages/images before sending the document to the browser works fine.  Waiting to assign the handler in javascript fails consistently, but only for cover pages/images.  I don't know why that seems to affect only cover pages/images.
___
*[`src/BloomBrowserUI/bookEdit/js/bloomImages.ts` line 120 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6830#-OII6a1B9SVU7rchvG2o:-OIIQZIF-rLLoLj3p7sm:b-odakqf) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/9b20ec327915150547743f1f2b957e91d2d71884/src/BloomBrowserUI/bookEdit/js/bloomImages.ts#L120)):*

I'm not sure I understand the problem here, but if part of it is contention for the onerror attribute, I think you can just addEventListener("error", HandleImageError), and that will not conflict with whatever the C# puts in onerror. Also, you can add a third argument {once: true} and not have to worry about removing it.
(If you change this here, probably should make the same change where you handle new image above.)
</blockquote></details>

No, contention for onerror is not a problem.  If the HandleImageError method didn't remove itself, it would be functionally identical to the code inserted by C#.  Not deleting itself might be less confusing.  Assigning it this way in javascript doesn't persist in the saved HTML, so there's no real need to remove it.  I've adjusted the javascript method accordingly.
___
*[`src/BloomBrowserUI/bookEdit/js/bubbleManager.ts` line 5269 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6830#-OIIBWzO7zQg0mhribGZ:-OILv3jw0FTl2zpVqUGj:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/9b20ec327915150547743f1f2b957e91d2d71884/src/BloomBrowserUI/bookEdit/js/bubbleManager.ts#L5269)):*

I may be missing the point of your changes here. It's probably time to give this method an overall comment (I should have written most of it before), along the lines
"Given a bgOverlay element, which is always an overlay having the bloom-backgroundImage class, and the height and width of the parent imageContainer, this method attempts to make the bgOverlay the right size and position to fill as much as possible of the parent, rather like object-fit:contain. It is used in two main scenarios: the user may have selected a different image, which means we must adjust to suit a different image aspect ratio. Or, the size of the container may have changed, e.g., using origami. We must also account for the possibility that the image has been cropped, in which case, we want to keep the cropped aspect ratio. (Cropping attributes will already have been removed if it is a new image.)
Things are complicated because it's possible the image has not loaded yet, so we can't get its natural dimensions to figure an aspect ratio. In this case, the method arranges to be called again after the image loads or a timeout.
A further complication is that the image may fail to load, so we never get natural dimensions. In this case, we [is this right?] expand the bgOverlay to the full size of the container so all the space is available to display the error icon and message.
</blockquote></details>

Done.
___
*[`src/BloomBrowserUI/bookEdit/js/bubbleManager.ts` line 5298 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6830#-OIIA8k9E-gCRdf6dOu5:-OILv6bVEymo-HWem0RR:bcfuggw) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/9b20ec327915150547743f1f2b957e91d2d71884/src/BloomBrowserUI/bookEdit/js/bubbleManager.ts#L5298)):*

Not getting the point of this. How can it have a naturalHeight if it failed to load, and why would we want to ignore the failed load if we somehow have a naturalHeight anyway?
</blockquote></details>

naturalHeight is set to zero until/before the image loads.  If an error happens instead, it stays set to 0.  I suppose the check is somewhat edundant and can be removed, but it's conceivable that the image has somehow loaded successfully this time after previously failing and setting the class.
I"ve added a couple of explanatory comments
___
*[`src/BloomBrowserUI/bookEdit/js/bubbleManager.ts` line 5309 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6830#-OII9MXI90HT7g-HDyR8:-OIMBOVG4pnkEvFEHs5j:ba4g7tg) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/9b20ec327915150547743f1f2b957e91d2d71884/src/BloomBrowserUI/bookEdit/js/bubbleManager.ts#L5309)):*

I doubt we want to keep the bg overlay shape that was produced by cropping if we no longer have the image and are going to remove the cropping?
</blockquote></details>

This was the original code, but I can rearrange the order of if/else actions.
___
*[`src/BloomBrowserUI/bookEdit/js/bubbleManager.ts` line 5346 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6830#-OII4SZo7tNhyEdheNWr:-OIMC7Uz0rEODNSf-wLh:b6w1z1h) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/9b20ec327915150547743f1f2b957e91d2d71884/src/BloomBrowserUI/bookEdit/js/bubbleManager.ts#L5346)):*

The point of this return is to try this call again later in the expectation that eventually it will load and we will get dimensions. But if bloom-imageLoadError is present, then presumably, we never will. So we had better adjust it to something that will show the error message now, hadn't we? I'm thinking that if we know we're never going to get image dimensions, then we may as well match the overlay dimensions to the container, which I think will happen if we set imgAspectRatio to containerWidth/containerHeight.
</blockquote></details>

I've fixed the code to at most call the timeout once if the class is set, just in case the class is stale and the image loads successfully (as indicated by setting naturalWidth and naturalHeight to something greater than zero).
___
*[`src/BloomBrowserUI/bookEdit/js/bubbleManager.ts` line 5355 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6830#-OII8B5M0IPs9q9AI8Fu:-OIMCWvwBfdWqhqb8P2B:bwiweo6) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/9b20ec327915150547743f1f2b957e91d2d71884/src/BloomBrowserUI/bookEdit/js/bubbleManager.ts#L5355)):*

I'm not sure how we get here if the img is missing (see comment below). But if we do, what imgAspectRatio do we want to use? If we can't match the real image aspect ratio I'd think we should use the container aspect ratio rather than the current default of keeping the old BG overlay aspect ratio?
</blockquote></details>

I've changed the code to use the container aspect ratio for the image if it fails to load.
___
*[`src/BloomBrowserUI/bookEdit/js/bubbleManager.ts` line 5379 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6830#-OII7cB-2r_4idN1gi54:-OIMChmR4kEEh2-ASLpx:bgdlsni) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/9b20ec327915150547743f1f2b957e91d2d71884/src/BloomBrowserUI/bookEdit/js/bubbleManager.ts#L5379)):*

This must somehow work, but I don't understand how we even get here if the image is missing. If it's not cropped, image.style.width will be zero, which takes us to the second branch above, where presumably if we can't load it, naturalWidth and naturalHeight are both zero, at which point we set up load and timeout event handlers and return. Load will never happen, and the timeout call will just find that naturalHeight and width are still zero. Or do we get some naturalWidth and height somehow, maybe the shape of the missing image icon?? Anyway, I think a comment is indicated somewhere.
</blockquote></details>

I've added code and comments to make things clearer (I think).
___
*[`src/BloomBrowserUI/bookEdit/js/bubbleManager.ts` line 5382 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6830#-OII8xBe-JtzxLipxm5_:-OIMEh4n7SzlUyIXV_lr:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/9b20ec327915150547743f1f2b957e91d2d71884/src/BloomBrowserUI/bookEdit/js/bubbleManager.ts#L5382)):*

comment (I think): the missing image was cropped. Remove the cropping to facilitate displaying the error state.
</blockquote></details>

Done.
___
*[`src/BloomBrowserUI/bookEdit/js/bubbleManager.ts` line 5393 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6830#-OII9A60ALfvpG90MouY:-OIMDOCw9yYIXxe66IdG:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/9b20ec327915150547743f1f2b957e91d2d71884/src/BloomBrowserUI/bookEdit/js/bubbleManager.ts#L5393)):*

probably clear top too? We don't set height on the img, but top is used for cropping on the top edge.
</blockquote></details>

Done.
___
*[`src/content/bookLayout/basePage-sharedRules.less` line 60 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6830#-OIIAu3k5rJYUV1wTNwz:-OIMN5qBAJ54eJM8Pkyc:bjg0fxl) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/9b20ec327915150547743f1f2b957e91d2d71884/src/content/bookLayout/basePage-sharedRules.less#L60)):*

Don't see how this can work. The adjustBackgroundImage code is still setting explicit height and width on the bloom-backgroundImage object, and style attribute values always beat CSS rules. Hopefully we can get those right in that method and not need this?
</blockquote></details>

This was left over from my initial attempt to fix in CSS as much as possible.  It's been removed.
___
*[`src/content/bookLayout/basePage-sharedRules.less` line 65 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6830#-OIIBCcq4Jirxe0m5_Na:-OIMNGVcCU-I2jinKxnZ:bjg0fxl) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/9b20ec327915150547743f1f2b957e91d2d71884/src/content/bookLayout/basePage-sharedRules.less#L65)):*

I think we shouldn't need this...if we're making the bg overlay the full size of the imageContainer, normal alignment code should make the overlay control frame that size, too.
</blockquote></details>

This was left over from my initial attempt to fix in CSS as much as possible.  It's been removed.

JohnThomson (changes requested):
Not my read. Your other comments say that the class is not persisted on the cover, and while it may actually be persisted on content pages, in effect it is not, since your page load code clears it.
I tend to think this is a good thing, though. The missing image might have been restored.

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6830#-:-OISEGUE3cWbaD5W8-tq:b-ayxzaz)* status: all files reviewed, 1 unresolved discussion (waiting on @StephenMcConnel)

___
*[`src/BloomBrowserUI/bookEdit/js/bloomImages.ts` line 123 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6830#-OISG-0B-EJO8MHV1wq4:-OISG-0B-EJO8MHV1wq5:b-7k6scs) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/04e42899cc8d0a744e7450437e16b7d892d403b4/src/BloomBrowserUI/bookEdit/js/bloomImages.ts#L123)):*
> ```XML
>         // error handler and class.  See BL-14241.
>         // Note that the error handler assigned this way will not be persisted as an
>         // attribute in the image element in the HTML.
> ```

From your other responses, I think the error handler added by C# actually does the same thing as HandleImageError; we just have to set it up in C# for cover images because for some reason this code doesn't execute soon enough in that case. I think it would be helpful to capture that here: it would make the comment clearer, and also make clearer that any changes here probably need to be copied there, and also save the reader wondering why we want to do something different about cover image errors.

StephenMcConnel (commented):
Each time the book is edited, the bloom-imageLoadError classes get cleared out on page load, but they are persisted in the book's HTML.  At one time, I had the page save code remove them, but I took that out before the PR.  It seemed like using both a belt and suspenders to address the issue.  :-)

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6830#-:-OISQvBU4ElbExm1-EnE:b-hk8j96)* status: all files reviewed, 1 unresolved discussion (waiting on @JohnThomson)

___
*[`src/BloomBrowserUI/bookEdit/js/bloomImages.ts` line 123 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6830#-OISG-0B-EJO8MHV1wq4:-OISN0307XK8j_4i87al:b-u3st0) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/04e42899cc8d0a744e7450437e16b7d892d403b4/src/BloomBrowserUI/bookEdit/js/bloomImages.ts#L123)):*

From your other responses, I think the error handler added by C# actually does the same thing as HandleImageError; we just have to set it up in C# for cover images because for some reason this code doesn't execute soon enough in that case. I think it would be helpful to capture that here: it would make the comment clearer, and also make clearer that any changes here probably need to be copied there, and also save the reader wondering why we want to do something different about cover image errors.
</blockquote></details>

Done.  I've revised the comment here (and in the C# code), and added a comment on the handler method itself.

JohnThomson (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6830#-:-OIWVvD57O2BSByW0U8D:bnfp4nl)* status: 5 of 7 files reviewed, all discussions resolved (waiting on @StephenMcConnel)

PR #6832, opened 2025-02-03
BL-14233 Make language chooser dialog work before project

nabalone:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6832)

### DIFF
--- a/src/BloomBrowserUI/collection/LanguageChooserDialog.tsx
+++ b/src/BloomBrowserUI/collection/LanguageChooserDialog.tsx
@@ -43,9 +43,11 @@ export function getLanguageData(
 ): ILanguageData {
     return {
         LanguageTag: languageTag || null,
-        DefaultName: defaultDisplayName(selection?.language),
+        DefaultName: selection?.language
+            ? defaultDisplayName(selection.language)
+            : null,
         DesiredName: selection?.customDetails?.displayName || null,
-        Country: defaultRegionForLangTag(languageTag).name
+        Country: languageTag ? defaultRegionForLangTag(languageTag)?.name : null
     };
 }
 
--- a/src/BloomExe/ApplicationContainer.cs
+++ b/src/BloomExe/ApplicationContainer.cs
@@ -7,6 +7,9 @@ using System.Linq;
 using L10NSharp;
 using System.Windows.Forms;
 using Bloom.web.controllers;
+using Bloom.Api;
+using Bloom.ImageProcessing;
+using Bloom.Book;
 
 namespace Bloom
 {
@@ -45,6 +48,19 @@ namespace Bloom
             builder.Register<HtmlThumbNailer>(c => new HtmlThumbNailer()).SingleInstance();
             builder
                 .Register<BookThumbNailer>(c => new BookThumbNailer(c.Resolve<HtmlThumbNailer>()))
+                .SingleInstance();
+
+            var bookRenameEvent = new BookRenamedEvent();
+            builder.Register(c => bookRenameEvent).AsSelf().InstancePerLifetimeScope();
+            builder.Register<BookSelection>(c => new BookSelection()).SingleInstance();
+            builder
+                .Register<BloomServer>(
+                    c =>
+                        new BloomServer(
+                            new RuntimeImageProcessor(bookRenameEvent),
+                            c.Resolve<BookSelection>()
+                        )
+                )
                 .SingleInstance();
 
             _container = builder.Build();
--- a/src/BloomExe/ApplicationContainer.cs
+++ b/src/BloomExe/ApplicationContainer.cs
@@ -63,9 +63,20 @@ namespace Bloom
                 )
                 .SingleInstance();
 
+            //Other classes which are also singletons for the whole application
+            builder
+                .RegisterAssemblyTypes(Assembly.GetExecutingAssembly())
+                .InstancePerLifetimeScope()
+                .Where(
+                    t => new[] { typeof(CommonApi), typeof(NewCollectionWizardApi), }.Contains(t)
+                );
+
             _container = builder.Build();
 
             Application.ApplicationExit += OnApplicationExit;
+            var server = _container.Resolve<BloomServer>();
+            _container.Resolve<CommonApi>().RegisterWithApiHandler(server.ApiHandler);
+            _container.Resolve<NewCollectionWizardApi>().RegisterWithApiHandler(server.ApiHandler);
         }
 
         private void OnApplicationExit(object sender, EventArgs e)
@@ -85,6 +96,8 @@ namespace Bloom
         public BookThumbNailer BookThumbNailer => _container.Resolve<BookThumbNailer>();
 
         internal ProblemReportApi ProblemReportApi => _container.Resolve<ProblemReportApi>();
+
+        public BloomServer BloomServer => _container.Resolve<BloomServer>();
 
         public void Dispose()
         {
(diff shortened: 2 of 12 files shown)


JohnThomson (changes requested):
Looks good. (Most of it we did together). I'd like to add a few more comments (suggested individually), and I have one question about the dialog changes; otherwise, LGTM

> Reviewed 12 of 12 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6832#-:-OICPGgH2QE03sEdPI4U:bwuxpno)* status: all files reviewed, 4 unresolved discussions (waiting on @nabalone)

___
*[`src/BloomExe/Program.cs` line 1419 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6832#-OICMHOo3FiOFmcrWDlK:-OICMHOo3FiOFmcrWDlL:b-uyw9iu) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/57dcade39b0be7fc8be7ec951320071d33f43627/src/BloomExe/Program.cs#L1419)):*
> ```Smalltalk
>                 //		return;
>                 //	}
>                 //}
> ```

Comment, something like "We normally start listening when setting up the project context. However, this dialog might get run at startup when we don't have a chosen project from which to make a ProjectContext."
We could also explore starting listening earlier in startup so there is less need for various things to EnsureListening. However, there may be paths...especially in automated testing...where we would end up doing a lot of work to spin up a server we don't need.
___
*[`src/BloomExe/ApplicationContainer.cs` line 77 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6832#-OICKsrZ6sGAYtQwd4cA:-OICKsr_B_EjockAOFKj:bmruwdz) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/57dcade39b0be7fc8be7ec951320071d33f43627/src/BloomExe/ApplicationContainer.cs#L77)):*
> ```Smalltalk
> 
>             Application.ApplicationExit += OnApplicationExit;
>             var server = _container.Resolve<BloomServer>();
> ```

I think we want a blank line here, then a comment "Register the API Handlers that are global to the application (not dependent on knowing a particular project).
Note: it is is a work in progress to transfer more API handlers from ProjectContext to here. Ideally, nothing in BloomServer, and hence not in any API handler, would know the current project. Any API call whose answer is project-dependent would pass a project identifier. Then all the handlers could all be registered here (and created by the ApplicationContainer). It's likely that a lot more could already be moved, but so far we just did enough for the handful of dialogs that need to work independent of a project.
___
*[`src/BloomBrowserUI/collection/LanguageChooserDialog.tsx` line 48 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6832#-OICOwg87YXrGMlC3lyt:-OICOwg87YXrGMlC3lyu:b8fiouf) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/57dcade39b0be7fc8be7ec951320071d33f43627/src/BloomBrowserUI/collection/LanguageChooserDialog.tsx#L48)):*
> ```XML
>         DefaultName: selection ? defaultDisplayName(selection?.language) : null,
>         DesiredName: selection?.customDetails?.displayName || null,
>         Country: languageTag ? defaultRegionForLangTag(languageTag).name : null
> ```

There are more conceivable nulls here. Is it guaranteed that defaultRegionForLangTag never returns null? Will defaultDisplayName handle a null input, or is selection.language guaranteed never to return one?
___
*[`src/BloomExe/web/BloomApiHandler.cs` line 46 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6832#-OICNh1BE5huOOyiMQ9Q:-OICNh1BE5huOOyiMQ9R:b-np585i) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/57dcade39b0be7fc8be7ec951320071d33f43627/src/BloomExe/web/BloomApiHandler.cs#L46)):*
> ```Smalltalk
>         }
> 
>         public void SetCollectionSettingsDuringInitialization(CollectionSettings collectionSettings)
> ```

Might be worth a comment cross-referencing the one we wrote for the same method in BloomServer.

nabalone (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6832)* status: 8 of 12 files reviewed, 4 unresolved discussions (waiting on @JohnThomson)

___
*[`src/BloomBrowserUI/collection/LanguageChooserDialog.tsx` line 48 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6832#-OICOwg87YXrGMlC3lyt:-OICvNARE1DiN5OkoKVU:bnputky) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/57dcade39b0be7fc8be7ec951320071d33f43627/src/BloomBrowserUI/collection/LanguageChooserDialog.tsx#L48)):*

There are more conceivable nulls here. Is it guaranteed that defaultRegionForLangTag never returns null? Will defaultDisplayName handle a null input, or is selection.language guaranteed never to return one?
</blockquote></details>

Done. Thanks
___
*[`src/BloomExe/ApplicationContainer.cs` line 77 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6832#-OICKsrZ6sGAYtQwd4cA:-OICw3I70A0a7V1ILJxB:bsa0f4h) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/57dcade39b0be7fc8be7ec951320071d33f43627/src/BloomExe/ApplicationContainer.cs#L77)):*

I think we want a blank line here, then a comment "Register the API Handlers that are global to the application (not dependent on knowing a particular project).
Note: it is is a work in progress to transfer more API handlers from ProjectContext to here. Ideally, nothing in BloomServer, and hence not in any API handler, would know the current project. Any API call whose answer is project-dependent would pass a project identifier. Then all the handlers could all be registered here (and created by the ApplicationContainer). It's likely that a lot more could already be moved, but so far we just did enough for the handful of dialogs that need to work independent of a project.
</blockquote></details>

Done. Like this?
___
*[`src/BloomExe/Program.cs` line 1419 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6832#-OICMHOo3FiOFmcrWDlK:-OICxfty-5zSI5-iqeez:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/57dcade39b0be7fc8be7ec951320071d33f43627/src/BloomExe/Program.cs#L1419)):*

Comment, something like "We normally start listening when setting up the project context. However, this dialog might get run at startup when we don't have a chosen project from which to make a ProjectContext."
We could also explore starting listening earlier in startup so there is less need for various things to EnsureListening. However, there may be paths...especially in automated testing...where we would end up doing a lot of work to spin up a server we don't need.
</blockquote></details>

Done.
___
*[`src/BloomExe/web/BloomApiHandler.cs` line 46 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6832#-OICNh1BE5huOOyiMQ9Q:-OICyJJG4Yt8I5xVsfIE:bsa0f4h) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/57dcade39b0be7fc8be7ec951320071d33f43627/src/BloomExe/web/BloomApiHandler.cs#L46)):*

Might be worth a comment cross-referencing the one we wrote for the same method in BloomServer.
</blockquote></details>

Done. Like this?

PR #6840, opened 2025-02-06
Force CommonApi to be a singleton object (BL-14293)

StephenMcConnel:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6840)

### DIFF
--- a/src/BloomExe/ApplicationContainer.cs
+++ b/src/BloomExe/ApplicationContainer.cs
@@ -66,7 +66,7 @@ namespace Bloom
             //Other classes which are also singletons for the whole application
             builder
                 .RegisterAssemblyTypes(Assembly.GetExecutingAssembly())
-                .InstancePerLifetimeScope()
+                .SingleInstance()
                 .Where(
                     t => new[] { typeof(CommonApi), typeof(NewCollectionWizardApi), }.Contains(t)
                 );


JohnThomson (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6840)* status: 0 of 1 files reviewed, 1 unresolved discussion

___
*[`src/BloomExe/ApplicationContainer.cs` line 69 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6840#-OISQU7Y7K5qTFL5wtot:-OISQU7Y7K5qTFL5wtou:bpnsc0c) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/32b69e904417830241aed054f681c7aab6e357cb/src/BloomExe/ApplicationContainer.cs#L69)):*
> ```Smalltalk
>             builder
>                 .RegisterAssemblyTypes(Assembly.GetExecutingAssembly())
>                 .SingleInstance()
> ```

So if our discussion on Zulip is right, may be helpful to add a comment
// Not InstancePerLifetimeScope! Although that would make for a singleton at the application level, if one of these objects is requested from the child scope ProjectContext, it would make an independent instance (possibly every time it is asked for one, since ProjectContext has not been told to only make one). Singleton seems to be a much stronger constraint that forces a single one for this and all child containers, which is what we want for all the application singletons.

JohnThomson (commented):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6840#-:-OISRbDLCP4JU-024HKT:bnfp4nl)* status: all files reviewed, 1 unresolved discussion

PR #6846, opened 2025-02-12
Properly get unique name for file with period in name (BL-14213)

JohnThomson:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6846)

### DIFF
--- a/src/BloomExe/Book/BookStorage.cs
+++ b/src/BloomExe/Book/BookStorage.cs
@@ -3296,16 +3296,23 @@ namespace Bloom.Book
         }
 
         /// <summary>
-        /// if necessary, append a number to make the file name unique within the given folder
+        /// if necessary, append a number to make the file name unique within the given folder.
+        /// name must NOT include an extension. The supplied extension will be added rather than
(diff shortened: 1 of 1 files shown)


andrew-polk (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6846)* status: all files reviewed, 1 unresolved discussion (waiting on @JohnThomson)

___
*[`src/BloomExe/Book/BookStorage.cs` line 3300 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6846#-OIwb_0x7wbZpryqIJmJ:-OIwb_0x7wbZpryqIJmK:b-tu68qp) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/039416f45334c8639e84a2a3c05d4d25c001fc5a/src/BloomExe/Book/BookStorage.cs#L3300)):*
> name must NOT include an extension

I did a quick search, and I'm not convinced this is the case with current callers.
See, e.g. BookStorage.SaveCopyBeforeImportOverwrite()

PR #6850, opened 2025-02-13
Don't try to embed fonts from invisible elements (BL-14267)

StephenMcConnel:
See my comment at the end of https://issues.bloomlibrary.org/youtrack/issue/BL-14267.  But this may be a mistake for a multilingual book where one of the chosen languages uses a font, but it isn't one that is being displayed by default when publishing.

This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6850)


StephenMcConnel:
I've decided not embedding fonts for invisible elements makes it too likely to miss fonts needed for multilingual books.

PR #6851, opened 2025-02-13
Don't try to embed unused fonts (BL-14267)

StephenMcConnel:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6851)

### DIFF
--- a/src/BloomExe/Publish/BloomPub/BloomPubMaker.cs
+++ b/src/BloomExe/Publish/BloomPub/BloomPubMaker.cs
@@ -844,13 +844,11 @@ namespace Bloom.Publish.BloomPub
             IFontFinder fontFileFinder
         )
         {
-            const string defaultFont = "Andika";
-
             // "Andika" already in BR in the standard four faces, don't need to embed or make rule.
-            fontsWanted.RemoveWhere(x => x.fontName == defaultFont);
+            fontsWanted.RemoveWhere(x => x.fontFamily == PublishHelper.DefaultFont);
             // We don't need to embed Andika New Basic, because Andika will handle it.
             fontsWanted.RemoveWhere( // The default Andika font will handle Andika New Basic
-                x => x.fontName == "Andika New Basic"
+                x => x.fontFamily == "Andika New Basic"
             );
 
             PublishHelper.CheckFontsForEmbedding(
@@ -870,9 +868,9 @@ namespace Bloom.Publish.BloomPub
             var sb = new StringBuilder();
             foreach (var font in fontsWanted.OrderBy(x => x.ToString()))
             {
-                if (badFonts.Contains(font.fontName))
+                if (badFonts.Contains(font.fontFamily))
                     continue;
-                var group = fontFileFinder.GetGroupForFont(font.fontName);
+                var group = fontFileFinder.GetGroupForFont(font.fontFamily);
                 if (group != null)
                 {
                     EpubMaker.AddFontFace(sb, font, group);
@@ -889,10 +887,14 @@ namespace Bloom.Publish.BloomPub
             // Repair defaultLangStyles.css and other places in the output book if needed.
             if (badFonts.Any())
             {
-                PublishHelper.FixCssReferencesForBadFonts(book.FolderPath, defaultFont, badFonts);
+                PublishHelper.FixCssReferencesForBadFonts(
+                    book.FolderPath,
+                    PublishHelper.DefaultFont,
+                    badFonts
+                );
                 PublishHelper.FixXmlDomReferencesForBadFonts(
                     book.OurHtmlDom.RawDom,
-                    defaultFont,
+                    PublishHelper.DefaultFont,
                     badFonts
                 );
             }
--- a/src/BloomExe/Publish/Epub/EpubMaker.cs
+++ b/src/BloomExe/Publish/Epub/EpubMaker.cs
@@ -2731,7 +2731,6 @@ namespace Bloom.Publish.Epub
         {
             ISet<string> warningMessages = new HashSet<string>();
             var fontFileFinder = FontFileFinder.GetInstance(Program.RunningUnitTests);
-            const string defaultFont = "Andika";
             PublishHelper.CheckFontsForEmbedding(
                 progress,
                 _fontsUsedInBook,
@@ -2755,10 +2754,10 @@ namespace Bloom.Publish.Epub
             var sb = new StringBuilder();
             foreach (var font in _fontsUsedInBook.OrderBy(x => x.ToString()))
             {
-                if (badFonts.Contains(font.fontName))
+                if (badFonts.Contains(font.fontFamily))
                     continue;
                 var file = fontFileFinder.GetFileForFont(
-                    font.fontName,
+                    font.fontFamily,
                     font.fontStyle,
                     font.fontWeight
                 );
@@ -2772,7 +2771,7 @@ namespace Bloom.Publish.Epub
                     // If we can't embed the font, no reason to refer to it in the css.
                     continue;
                 }
-                var group = fontFileFinder.GetGroupForFont(font.fontName);
+                var group = fontFileFinder.GetGroupForFont(font.fontFamily);
                 if (group != null)
                 {
                     // The fonts.css file is stored in a subfolder as are the font files.  They are in different
@@ -2786,7 +2785,7 @@ namespace Bloom.Publish.Epub
                 && !_fontsUsedInBook
                     .Where(
                         x =>
-                            x.fontName == defaultFont
+                            x.fontFamily == PublishHelper.DefaultFont
                             && x.fontStyle == "normal"
                             && x.fontWeight != "700"
                     )
@@ -2794,10 +2793,10 @@ namespace Bloom.Publish.Epub
             )
                 AddFontFace(
                     sb,
-                    defaultFont,
+                    PublishHelper.DefaultFont,
                     "normal",
                     "normal",
-                    fontFileFinder.GetGroupForFont(defaultFont).Normal,
+                    fontFileFinder.GetGroupForFont(PublishHelper.DefaultFont).Normal,
                     "../" + kFontsFolder + "/"
                 );
             Directory.CreateDirectory(Path.Combine(_contentFolder, kCssFolder));
@@ -2811,10 +2810,10 @@ namespace Bloom.Publish.Epub
             {
                 PublishHelper.FixCssReferencesForBadFonts(
                     Path.Combine(_contentFolder, kCssFolder),
-                    defaultFont,
+                    PublishHelper.DefaultFont,
                     badFonts
                 );
-                FixXhtmlReferencesForBadFonts(_contentFolder, defaultFont, badFonts);
+                FixXhtmlReferencesForBadFonts(_contentFolder, PublishHelper.DefaultFont, badFonts);
             }
 
             return warningMessages;
@@ -2868,7 +2867,7 @@ namespace Bloom.Publish.Epub
                 path = group.Normal;
             AddFontFace(
                 sb,
-                font.fontName,
+                font.fontFamily,
                 weight,
                 font.fontStyle,
                 path,
--- a/src/BloomExe/Publish/PublishHelper.cs
+++ b/src/BloomExe/Publish/PublishHelper.cs
@@ -1,7 +1,6 @@
 using System;
 using System.Collections.Generic;
 using System.Diagnostics;
-using System.Drawing.Imaging;
 using System.Drawing;
 using System.IO;
 using System.Linq;
@@ -15,13 +14,11 @@ using Bloom.Publish.Epub;
 using Bloom.SafeXml;
 using Bloom.web;
 using Bloom.web.controllers;
-using Bloom.Workspace;
 using L10NSharp;
 using Newtonsoft.Json;
 using SIL.IO;
 using SIL.Progress;
 using SIL.Reporting;
-using SIL.Xml;
 
 namespace Bloom.Publish
 {
@@ -184,23 +181,27 @@ namespace Bloom.Publish
 
         public class FontInfo
         {
-            public string fontName;
+            public string fontFamily;
             public string fontStyle;
             public string fontWeight;
 
             public override string ToString()
             {
                 if (string.IsNullOrEmpty(fontStyle) && string.IsNullOrEmpty(fontWeight))
-                    return string.IsNullOrEmpty(fontName) ? "<uninitialized FontInfo>" : fontName;
+                    return string.IsNullOrEmpty(fontFamily)
+                        ? "<uninitialized FontInfo>"
+                        : fontFamily;
                 if (fontStyle == "normal" && fontWeight == "400")
-                    return string.IsNullOrEmpty(fontName) ? "<uninitialized FontInfo>" : fontName;
+                    return string.IsNullOrEmpty(fontFamily)
+                        ? "<uninitialized FontInfo>"
+                        : fontFamily;
                 if (fontStyle == "normal" && fontWeight == "700")
-                    return $"{fontName} Bold";
+                    return $"{fontFamily} Bold";
                 if (fontStyle == "italic" && fontWeight == "400")
-                    return $"{fontName} Italic";
+                    return $"{fontFamily} Italic";
                 if (fontStyle == "italic" && fontWeight == "700")
-                    return $"{fontName} Bold Italic";
-                return $"{fontName} weight=\"{fontWeight}\" style=\"{fontStyle}\"";
+                    return $"{fontFamily} Bold Italic";
+                return $"{fontFamily} weight=\"{fontWeight}\" style=\"{fontStyle}\"";
             }
 
             public override bool Equals(object obj)
@@ -208,14 +209,16 @@ namespace Bloom.Publish
                 var that = obj as FontInfo;
                 if (that == null)
                     return false;
-                return this.fontName == that.fontName
+                return this.fontFamily == that.fontFamily
                     && this.fontStyle == that.fontStyle
                     && this.fontWeight == that.fontWeight;
             }
 
             public override int GetHashCode()
             {
-                return fontName.GetHashCode() ^ fontStyle.GetHashCode() ^ fontWeight.GetHashCode();
+                return fontFamily.GetHashCode()
+                    ^ fontStyle.GetHashCode()
+                    ^ fontWeight.GetHashCode();
             }
 
             public static bool operator ==(FontInfo info1, FontInfo info2)
@@ -233,8 +236,9 @@ namespace Bloom.Publish
             }
         }
 
-        Dictionary<string, string> _mapIdToDisplay = new Dictionary<string, string>();
-        Dictionary<string, FontInfo> _mapIdToFontInfo = new Dictionary<string, FontInfo>();
+        protected Dictionary<string, string> _mapIdToDisplay = new Dictionary<string, string>();
+        protected Dictionary<string, FontInfo> _mapIdToFontInfo =
+            new Dictionary<string, FontInfo>();
 
         private void RemoveUnwantedContentInternal(
             HtmlDom dom,
@@ -438,7 +442,7 @@ namespace Bloom.Publish
                             FontsUsed.Add(
                                 new FontInfo
                                 {
-                                    fontName = font,
+                                    fontFamily = font,
                                     fontStyle = info.fontStyle,
                                     fontWeight = info.fontWeight
                                 }
@@ -449,7 +453,7 @@ namespace Bloom.Publish
                     _mapIdToDisplay[info.id] = info.display;
                     var fontInfo = new FontInfo
                     {
-                        fontName = info.fontFamily,
+                        fontFamily = info.fontFamily,
                         fontStyle = info.fontStyle,
                         fontWeight = info.fontWeight
                     };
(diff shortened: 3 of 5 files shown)


hatton (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6851)* status: 0 of 2 files reviewed, 1 unresolved discussion

___
*[`src/BloomExe/Publish/PublishHelper.cs` line 636 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6851#-OJ3z2U7EMUjU2_7BrgC:-OJ3z2U7EMUjU2_7BrgD:btekhxk) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/fbf34659dea466feb49cc018536b62a84e566ab6/src/BloomExe/Publish/PublishHelper.cs#L636)):*
> ```Smalltalk
>             var defaultFonts =
>                 "Andika, \"Andika New Basic\", \"Andika Basic\", \"Gentium Basic\", \"Gentium Book Basic\", \"Doulos SIL\", sans-serif";
>             if (fontInfo.fontName == defaultFonts)
> ```

I'm not understanding how this line works. Maybe I need more coffee?

StephenMcConnel (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6851)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @hatton)

___
*[`src/BloomExe/Publish/PublishHelper.cs` line 636 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6851#-OJ3z2U7EMUjU2_7BrgC:-OJJWoWt46846iE07idE:brf2fbl) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/fbf34659dea466feb49cc018536b62a84e566ab6/src/BloomExe/Publish/PublishHelper.cs#L636)):*

I'm not understanding how this line works. Maybe I need more coffee?
</blockquote></details>

The fontIInfo.fontName is set by the CSS font-family property, which can be a fallback list of fonts, not just a single font.  The variable name is perhaps unfortunate.

StephenMcConnel (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6851)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @hatton)

___
*[`src/BloomExe/Publish/PublishHelper.cs` line 636 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6851#-OJ3z2U7EMUjU2_7BrgC:-OJK0nQ70eufDpfBZLmw:b-t7xz4r) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/fbf34659dea466feb49cc018536b62a84e566ab6/src/BloomExe/Publish/PublishHelper.cs#L636)):*

The fontIInfo.fontName is set by the CSS font-family property, which can be a fallback list of fonts, not just a single font.  The variable name is perhaps unfortunate.
</blockquote></details>

I've changed the property name from fontName to fontFamily to reflect its CSS origin.  I also added a comment here to explain what is going on.

hatton (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6851)* status: 0 of 5 files reviewed, 2 unresolved discussions (waiting on @StephenMcConnel)

___
*[`src/BloomExe/Publish/PublishHelper.cs` line 636 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6851#-OJ3z2U7EMUjU2_7BrgC:-OJK1nxX8IO2fpWaIFuF:bs1tcss) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/fbf34659dea466feb49cc018536b62a84e566ab6/src/BloomExe/Publish/PublishHelper.cs#L636)):*

I've changed the property name from fontName to fontFamily to reflect its CSS origin.  I also added a comment here to explain what is going on.
</blockquote></details>

So this is relying on an exact (including spaces) copy of some other CSS file to work. Can we do better?

OK 2 improvement ideas:
1) We have an ordered set of font family names. We don't know which one the browser would use. But a fair simplification would be to just take the first one and reason about it. If we did that, what we want to know is, is the first one in the list of fonts that we ship with all of our platforms?

2) If we are going to filter those out here, then maybe the name of the method wants to be `CollectFontsToEmbedAlongWithLanguagesUsingThem` and then it can operate on a `FontsToEmbedWithLanguagseUsingThem` variable.  Yeah those names are awkward but actually tell us what we need to know.
___
*[`src/BloomExe/Publish/PublishHelper.cs` line 447 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6851#-OJK3Td79UfzkJJsS3--:-OJK3Td79UfzkJJsS3-0:bpg8izn) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/fbf34659dea466feb49cc018536b62a84e566ab6/src/BloomExe/Publish/PublishHelper.cs#L447)):*
> ```Smalltalk
>                     var fontInfo = new FontInfo
>                     {
>                         fontName = info.fontFamily,
> ```

This is an seemingly unnecessary complexity, changing from `fontFamily` to `fontName`?

StephenMcConnel (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6851)* status: 0 of 5 files reviewed, 2 unresolved discussions (waiting on @hatton)

___
*[`src/BloomExe/Publish/PublishHelper.cs` line 447 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6851#-OJK3Td79UfzkJJsS3--:-OJKxK5H0uiAZIw10zXI:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/fbf34659dea466feb49cc018536b62a84e566ab6/src/BloomExe/Publish/PublishHelper.cs#L447)):*

This is an seemingly unnecessary complexity, changing from `fontFamily` to `fontName`?
</blockquote></details>

Done.
___
*[`src/BloomExe/Publish/PublishHelper.cs` line 636 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6851#-OJ3z2U7EMUjU2_7BrgC:-OJLLEru3W88zmALwQMy:bggmuqd) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/fbf34659dea466feb49cc018536b62a84e566ab6/src/BloomExe/Publish/PublishHelper.cs#L636)):*

So this is relying on an exact (including spaces) copy of some other CSS file to work. Can we do better?

OK 2 improvement ideas:
1) We have an ordered set of font family names. We don't know which one the browser would use. But a fair simplification would be to just take the first one and reason about it. If we did that, what we want to know is, is the first one in the list of fonts that we ship with all of our platforms?

2) If we are going to filter those out here, then maybe the name of the method wants to be `CollectFontsToEmbedAlongWithLanguagesUsingThem` and then it can operate on a `FontsToEmbedWithLanguagseUsingThem` variable.  Yeah those names are awkward but actually tell us what we need to know.
</blockquote></details>

I've revamped the algorithm here to try to detect when Andika (aka DefaultFont) as the head of a fall-back chain is not really used to display any characters.  When that happens, it's not really used and doesn't need to be added the list of fonts used.

JohnThomson (changes requested):
> Reviewed 1 of 4 files at r2, 4 of 4 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6851)* status: all files reviewed, 6 unresolved discussions (waiting on @hatton)

___
*[a discussion](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6851#-OJQUO30DHgSLOrJbIeM:-OJQUO30DHgSLOrJbIeN:b9ek1em) (no related file):*
A few minor points, but basically looks good.
___
*[`src/BloomExe/Publish/PublishHelper.cs` line 643 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6851#-OJQTmT19bZQoXDkAGlx:-OJQTmT19bZQoXDkAGly:biokbjo) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/5c0a67bd13ff4373d15cecb0ba0a5bf024d99c04/src/BloomExe/Publish/PublishHelper.cs#L643)):*
> ```Smalltalk
>             // with our default font ("Andika") and then adds a bunch of fallback fonts.  We don't
>             // need to record that "Andika" is used in this case, at least not when it isn't used
>             // for actual text.
> ```

If I'm reading the usages of FontsUsed right, we only care about this so that we don't spuriously report Analytics that Andika was used when it really wasn't. In case someone worries that this algorithm might not be perfectly reliable, it might be worth commenting that we're not depending on this for something critical like deciding whether to embed the font.
___
*[`src/BloomExe/Publish/PublishHelper.cs` line 692 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6851#-OJQRFCEEg2dIg75ymZ3:-OJQRFCEEg2dIg75ymZ4:b-tigl02) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/5c0a67bd13ff4373d15cecb0ba0a5bf024d99c04/src/BloomExe/Publish/PublishHelper.cs#L692)):*
> ```Smalltalk
>                     && display == "none"
>                     && elt.InnerText.Trim() == ""
>                 )
> ```

Unless "display == "none"" means something very unexpected, I would think that if we're not displaying it at all, we don't care whether it's empty? We could continue anyway? Or maybe we should delete its innerText without further checking, since it is NOT displayed in the font we care about?
___
*[`src/BloomExe/Publish/PublishHelper.cs` line 696 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6851#-OJQSJV50dGw-ROrW-ia:-OJQSJV50dGw-ROrW-ib:bnvtjlr) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/5c0a67bd13ff4373d15cecb0ba0a5bf024d99c04/src/BloomExe/Publish/PublishHelper.cs#L696)):*
> ```Smalltalk
>                     continue;
>                 }
>                 if (!_mapIdToFontInfo.TryGetValue(id, out var fontInfo))
> ```

I think the idea here is that if our map says that the font actually used by the browser to display this element is not the one we are looking for, we remove it from the text we return. Might be worth a comment.
Is "font actually used" monotonic over an element? I vaguely recall that the browser is pretty smart about substituting fonts if a character is missing. But maybe this is the best we can do?

StephenMcConnel (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6851)* status: all files reviewed, 6 unresolved discussions (waiting on @hatton and @JohnThomson)

___
*[a discussion](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6851#-OJQUO30DHgSLOrJbIeM:-OJV4U2220jj0ipUDN2K:bisdl2) (no related file):*

A few minor points, but basically looks good.
</blockquote></details>

I've responded by beefing up the comments in the code, as well as making a few comments here.
___
*[`src/BloomExe/Publish/PublishHelper.cs` line 643 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6851#-OJQTmT19bZQoXDkAGlx:-OJUxKkDBMJN1rWaj4kR:b-nhd7ht) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/5c0a67bd13ff4373d15cecb0ba0a5bf024d99c04/src/BloomExe/Publish/PublishHelper.cs#L643)):*

If I'm reading the usages of FontsUsed right, we only care about this so that we don't spuriously report Analytics that Andika was used when it really wasn't. In case someone worries that this algorithm might not be perfectly reliable, it might be worth commenting that we're not depending on this for something critical like deciding whether to embed the font.
</blockquote></details>

It's also used for embedding fonts in epubs and bloompubs, although Andika (and Andika New Basic) is later filtered out for bloompubs.
___
*[`src/BloomExe/Publish/PublishHelper.cs` line 692 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6851#-OJQRFCEEg2dIg75ymZ3:-OJUxWYn91hDs1bF56BU:b8gqoy0) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/5c0a67bd13ff4373d15cecb0ba0a5bf024d99c04/src/BloomExe/Publish/PublishHelper.cs#L692)):*

Unless "display == "none"" means something very unexpected, I would think that if we're not displaying it at all, we don't care whether it's empty? We could continue anyway? Or maybe we should delete its innerText without further checking, since it is NOT displayed in the font we care about?
</blockquote></details>

For epubs, we probably don't care about whether any content exists since those don't support multilingual books apart from all languages being displayed all of the time.  For bloompubs, we could be publishing a language that isn't currently displayed and thus would have content but be currently invisible while publishing.
___
*[`src/BloomExe/Publish/PublishHelper.cs` line 696 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6851#-OJQSJV50dGw-ROrW-ia:-OJV1Zk30NN1gDyLY8qe:bu4o558) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/5c0a67bd13ff4373d15cecb0ba0a5bf024d99c04/src/BloomExe/Publish/PublishHelper.cs#L696)):*

I think the idea here is that if our map says that the font actually used by the browser to display this element is not the one we are looking for, we remove it from the text we return. Might be worth a comment.
Is "font actually used" monotonic over an element? I vaguely recall that the browser is pretty smart about substituting fonts if a character is missing. But maybe this is the best we can do?
</blockquote></details>

I don't understand your comment asking for a comment.  :-)
I think this information is the best we can do.  I don't think we have access to whatever fallback is actually used for some random character.

JohnThomson (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6851)* status: all files reviewed, 6 unresolved discussions (waiting on @hatton and @StephenMcConnel)

___
*[a discussion](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6851#-OJQUO30DHgSLOrJbIeM:-OJVOPUA6dILvTsht1nI:by405h3) (no related file):*

I've responded by beefing up the comments in the code, as well as making a few comments here.
</blockquote></details>

I've responded to your comments, and also picked up a couple of new issues. Sorry I missed them before.
___
*[`src/BloomExe/Publish/PublishHelper.cs` line 692 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6851#-OJQRFCEEg2dIg75ymZ3:-OJVGTRjCgFAV66VAgCo:bjense5) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/5c0a67bd13ff4373d15cecb0ba0a5bf024d99c04/src/BloomExe/Publish/PublishHelper.cs#L692)):*

For epubs, we probably don't care about whether any content exists since those don't support multilingual books apart from all languages being displayed all of the time.  For bloompubs, we could be publishing a language that isn't currently displayed and thus would have content but be currently invisible while publishing.
</blockquote></details>

Hmm, I think I was reading this backwards. If this condition is true, we continue, that is, we do NOT remove the content of this node from the result. Now, that makes sense if its InnerText is empty: there's noting significant to remove. But why would we want to refrain from removing empty innertext that is displayed?
___
*[`src/BloomExe/Publish/PublishHelper.cs` line 696 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6851#-OJQSJV50dGw-ROrW-ia:-OJVMUs16LWXwNlE2Kae:b-5a192c) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/5c0a67bd13ff4373d15cecb0ba0a5bf024d99c04/src/BloomExe/Publish/PublishHelper.cs#L696)):*

I don't understand your comment asking for a comment.  :-)
I think this information is the best we can do.  I don't think we have access to whatever fallback is actually used for some random character.
</blockquote></details>

OK, as I look at this more closely, I have another observation. If we continue here, we aren't really deciding to ignore this element (as the comment says); we're deciding to count it as text in the interesting font, because we're NOT going to delete it from the original innerText.
I guess I'm looking for a comment (if this is true, and this line is probably not the place for it) along the lines,
"For each interesting child of the element, if the font family (actually the first font in the family) that the browser says is used to display it isn't the one we are looking for, delete the corresponding text from innerText, so that we're left with just the text that is displayed in the font we're asking about"
___
*[`src/BloomExe/Publish/PublishHelper.cs` line 711 at r4](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6851#-OJVIWQo2D9dlV6NKkd0:-OJVIWQo2D9dlV6NKkd1:b-rht1jy) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/3bcb1a8e82bd55b46365ab359b872de19ed265b1/src/BloomExe/Publish/PublishHelper.cs#L711)):*
> ```Smalltalk
>                 var eltText = elt.InnerText;
>                 if (font1 != font && eltText != "" && innerText.Contains(eltText))
>                     innerText = innerText.Remove(innerText.IndexOf(eltText), eltText.Length);
> ```

The first occurrence of the element text might not be the current one. For example, suppose we have "This <strong>word</strong> is <strong>bold</strong>, and this <em>word</em> is <em>italic</em>", and we ask the routine what text is displayed in the relevant bold font. The answer ought to be "wordbold" but will actually be "boldword" because the first "word" will be removed when we find that the italic "word" is in the wrong font.
I think this doesn't actually matter because you only use the result to test whether it is empty. But to guard against some future use where it might matter, maybe we should make the routine do the test, so instead of GetTrimmedInnerTextDisplayedInFont we would have "HasAnyTrimmedTextInFont" returning a boolean.
(It could also be slightly more efficient because as soon as we find any element that we are NOT going to remove and that isn't white we can return true.)

StephenMcConnel (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6851)* status: 4 of 5 files reviewed, 6 unresolved discussions (waiting on @hatton and @JohnThomson)

___
*[a discussion](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6851#-OJQUO30DHgSLOrJbIeM:-OJZsd262hPtjcu1-Jxs:b-896fix) (no related file):*

I've responded to your comments, and also picked up a couple of new issues. Sorry I missed them before.
</blockquote></details>

Done.
___
*[`src/BloomExe/Publish/PublishHelper.cs` line 692 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6851#-OJQRFCEEg2dIg75ymZ3:-OJZs9Lg-fnJfB7mc720:bugw73e) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/5c0a67bd13ff4373d15cecb0ba0a5bf024d99c04/src/BloomExe/Publish/PublishHelper.cs#L692)):*

Hmm, I think I was reading this backwards. If this condition is true, we continue, that is, we do NOT remove the content of this node from the result. Now, that makes sense if its InnerText is empty: there's noting significant to remove. But why would we want to refrain from removing empty innertext that is displayed?
</blockquote></details>

I've replaced the entire method with a new recursive method that should work reliably unless different fonts are used below the div level which have been tagged with ids.
___
*[`src/BloomExe/Publish/PublishHelper.cs` line 696 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6851#-OJQSJV50dGw-ROrW-ia:-OJZsQwvCiDCAT0xl4lZ:bo9zyuu) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/5c0a67bd13ff4373d15cecb0ba0a5bf024d99c04/src/BloomExe/Publish/PublishHelper.cs#L696)):*

OK, as I look at this more closely, I have another observation. If we continue here, we aren't really deciding to ignore this element (as the comment says); we're deciding to count it as text in the interesting font, because we're NOT going to delete it from the original innerText.
I guess I'm looking for a comment (if this is true, and this line is probably not the place for it) along the lines,
"For each interesting child of the element, if the font family (actually the first font in the family) that the browser says is used to display it isn't the one we are looking for, delete the corresponding text from innerText, so that we're left with just the text that is displayed in the font we're asking about"
</blockquote></details>

I've replaced this code altogether.
___
*[`src/BloomExe/Publish/PublishHelper.cs` line 711 at r4](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6851#-OJVIWQo2D9dlV6NKkd0:-OJZsWXw6n8unF0QEB7x:b-q9gcn0) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/3bcb1a8e82bd55b46365ab359b872de19ed265b1/src/BloomExe/Publish/PublishHelper.cs#L711)):*

The first occurrence of the element text might not be the current one. For example, suppose we have "This <strong>word</strong> is <strong>bold</strong>, and this <em>word</em> is <em>italic</em>", and we ask the routine what text is displayed in the relevant bold font. The answer ought to be "wordbold" but will actually be "boldword" because the first "word" will be removed when we find that the italic "word" is in the wrong font.
I think this doesn't actually matter because you only use the result to test whether it is empty. But to guard against some future use where it might matter, maybe we should make the routine do the test, so instead of GetTrimmedInnerTextDisplayedInFont we would have "HasAnyTrimmedTextInFont" returning a boolean.
(It could also be slightly more efficient because as soon as we find any element that we are NOT going to remove and that isn't white we can return true.)
</blockquote></details>

The new recursive method doesn't suffer from this indeterminancy.  :-)

JohnThomson (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6851)* status: all files reviewed, 3 unresolved discussions (waiting on @hatton and @StephenMcConnel)

___
*[a discussion](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6851#-OJQUO30DHgSLOrJbIeM:-OJ_RWmCCSj569d63Htf:bn1886u) (no related file):*

Done.
</blockquote></details>

Again I should have picked this up before, but these changes deserve unit tests. Particularly the quite complex and fairly self-contained IsRealTextDisplayedInDesiredFont
___
*[`src/BloomExe/Publish/PublishHelper.cs` line 692 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6851#-OJQRFCEEg2dIg75ymZ3:-OJ_RDYQ-wDuC9KXanMP:bszm2fk) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/5c0a67bd13ff4373d15cecb0ba0a5bf024d99c04/src/BloomExe/Publish/PublishHelper.cs#L692)):*

I've replaced the entire method with a new recursive method that should work reliably unless different fonts are used below the div level which have been tagged with ids.
</blockquote></details>

Much better.

JohnThomson (approved):
> Reviewed 2 of 2 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6851)* status: all files reviewed, 3 unresolved discussions (waiting on @hatton and @StephenMcConnel)

PR #6857, opened 2025-02-18
Use the source files provided by bloom-player (BL-14008)

StephenMcConnel:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6857)


JohnThomson (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6857)* status: all files reviewed, 1 unresolved discussion (waiting on @StephenMcConnel)

___
*[a discussion](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6857#-OJVF8ca4kpREYsOKnor:-OJVF8ca4kpREYsOKnos:b45b97s) (no related file):*
I'm curious why we need to copy the bloom-player src files to bookEdit/shared. Usually, to import something from a node_modules package like bloom-player, we just say the equivalent of "import blah from bloom-player" and something figures out that blah is a function that bloom-player exports and pulls in whatever code we need.
There may well be some reason doing things this way is impossible, unreasonably difficult, or unwise...maybe because of the different jquery packages involved? Or whatever makes deleting the line from scrolling.ts necessary?
Anyway, unless I'm just missing something that should be obvious, it might be worth putting something in shared/README to explain WHY things are done this way, before you forget.

StephenMcConnel (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6857)* status: all files reviewed, 1 unresolved discussion (waiting on @JohnThomson)

___
*[a discussion](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6857#-OJVF8ca4kpREYsOKnor:-OJ_CSRtDShXPP5E3mY2:bps291c) (no related file):*

I'm curious why we need to copy the bloom-player src files to bookEdit/shared. Usually, to import something from a node_modules package like bloom-player, we just say the equivalent of "import blah from bloom-player" and something figures out that blah is a function that bloom-player exports and pulls in whatever code we need.
There may well be some reason doing things this way is impossible, unreasonably difficult, or unwise...maybe because of the different jquery packages involved? Or whatever makes deleting the line from scrolling.ts necessary?
Anyway, unless I'm just missing something that should be obvious, it might be worth putting something in shared/README to explain WHY things are done this way, before you forget.
</blockquote></details>

bloom-player doesn't actually export any of these methods.  It's not set up to act like a library in that way.  Even it were, I think it would add about 840K to whatever module imported it.
I've added some commentary to the end of the README.md file.  What I said there may just expose my ignorance, but I still think this is the least cost way to share the code.

JohnThomson (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6857)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @StephenMcConnel)

PR #6860, opened 2025-02-21
Adjust Timings tweaks (BL-12953)

JohnThomson:
This is the new UI described in the middle of the issue.

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6860)


StephenMcConnel (changes requested):
A few questions and comments, but generally looks good as far as I can tell.

> Reviewed 20 of 20 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6860#-:-OJszFBJ9s8KocBIAxV9:b-2zp6gu)* status: all files reviewed, 6 unresolved discussions (waiting on @JohnThomson)

___
*[`src/BloomBrowserUI/bookEdit/toolbox/talkingBook/AdjustTimingsDialog.tsx` line 98 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6860#-OJsjr0yEF7g6TTvQIEw:-OJsjr0z3E77tiM86VN6:b53ojs) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/1d269e028497f298f03012d172d27b8cadb6b4f8/src/BloomBrowserUI/bookEdit/toolbox/talkingBook/AdjustTimingsDialog.tsx#L98)):*
> ```XML
>     const moreMenuOpen = Boolean(moreElForAdvancedMenu);
>     const handleClick = () => {
>         const anchor = document.getElementById("timimgsMenuAnchor");
> ```

This ID has an apparent misspelling.  Of course, if it's consistently misspelled this way, it doesn't matter as much.  :-)
___
*[`src/BloomBrowserUI/bookEdit/toolbox/talkingBook/AdjustTimingsDialog.tsx` line 111 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6860#-OJskcLWCrIhVHdl2uRv:-OJskcLWCrIhVHdl2uRw:b-z2jx3c) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/1d269e028497f298f03012d172d27b8cadb6b4f8/src/BloomBrowserUI/bookEdit/toolbox/talkingBook/AdjustTimingsDialog.tsx#L111)):*
> ```XML
>     // when we pass in segments from some other source, such as Aeneas or a timings file.
>     // So when we get this message, we clear the flag. That doesn't call for another render, so
>     // I made this a useRef rather than a state.
> ```

The comment would be clearer if the final "this" was replaced by "shouldAdjustSegments" since more is happening than just this operation, and setEndTimes is a useState.  (Is the import of this comment even true at this point?)
___
*[`src/BloomBrowserUI/bookEdit/toolbox/talkingBook/AdjustTimingsDialog.tsx` line 200 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6860#-OJsnfVLALDv5MoDGKCH:-OJsnfVLALDv5MoDGKCI:b-6t68wd) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/1d269e028497f298f03012d172d27b8cadb6b4f8/src/BloomBrowserUI/bookEdit/toolbox/talkingBook/AdjustTimingsDialog.tsx#L200)):*
> ```XML
>         ).getComputedStyle(bloomEditable).fontFamily;
>         ff = ff.replace(/^['"]/, "").replace(/['"]$/, "");
>         setFontFamily(ff);
> ```

Is it ever possible for ff to be a fallback list rather than a single font?  (I don't think so, without user custom CSS, but it may be worth asking.)
___
*[`src/BloomBrowserUI/bookEdit/toolbox/talkingBook/AdjustTimingsDialog.tsx` line 495 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6860#-OJst3cBEi1au2LLQ-z-:-OJst3cBEi1au2LLQ-z0:b-zdqlhs) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/1d269e028497f298f03012d172d27b8cadb6b4f8/src/BloomBrowserUI/bookEdit/toolbox/talkingBook/AdjustTimingsDialog.tsx#L495)):*
> ```XML
>     let start = 0;
>     for (let i = 0; i < segmentElements.length; i++) {
>         const end = endTimes ? endTimes[i] : start + 1;
> ```

So if the end data is incomplete relative to the text segments, we assume the segment is 1 second long.  Am I interpreting the code correctly?  A short comment might be in order, something like "if no timing data, default segment length is set to 1 second".
___
*[`src/BloomBrowserUI/react_components/localizableMenuItem.tsx` line 39 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6860#-OJsuQeg4NmzrVyt1x8z:-OJsuQeg4NmzrVyt1x9-:bm3ja7m) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/1d269e028497f298f03012d172d27b8cadb6b4f8/src/BloomBrowserUI/react_components/localizableMenuItem.tsx#L39)):*
> ```XML
>     // To make a sublabel, you may pass an entire node. If that is missing, you may pass a l10nId,
>     // and the corresponding string will be shown. If that also is missing, you may pass the actual
>     // string to show, already localized if necessary.
> ```

The priority order appears to be 1) entire node, 2) actual string to show (already localized), and 3) l10n ID to look up.  This comment implies 1) entire node, 2) l10n ID to look up, and 3) actual string to show.  Admittedly, this is the order of the parameters, but something should be adjusted for consistency.
___
*[`src/BloomBrowserUI/react_components/localizableMenuItem.tsx` line 159 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6860#-OJsvJkZ5po2Jm7mDtpE:-OJsvJkZ5po2Jm7mDtpF:b-krt54v) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/1d269e028497f298f03012d172d27b8cadb6b4f8/src/BloomBrowserUI/react_components/localizableMenuItem.tsx#L159)):*
> ```XML
>     const localizedSublabel = useL10n("", props.subLabelL10nId ?? null);
>     const sublabel =
>         props.sublabel ?? props.generatedSubLabel ?? localizedSublabel;
> ```

Are these in the right order?  (See my previous comment.)

JohnThomson (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6860)* status: 18 of 20 files reviewed, 6 unresolved discussions (waiting on @StephenMcConnel)

___
*[`src/BloomBrowserUI/bookEdit/toolbox/talkingBook/AdjustTimingsDialog.tsx` line 98 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6860#-OJsjr0yEF7g6TTvQIEw:-OJtLE5-E9CPsGPBaDxi:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/1d269e028497f298f03012d172d27b8cadb6b4f8/src/BloomBrowserUI/bookEdit/toolbox/talkingBook/AdjustTimingsDialog.tsx#L98)):*

This ID has an apparent misspelling.  Of course, if it's consistently misspelled this way, it doesn't matter as much.  :-)
</blockquote></details>

Done.
___
*[`src/BloomBrowserUI/bookEdit/toolbox/talkingBook/AdjustTimingsDialog.tsx` line 111 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6860#-OJskcLWCrIhVHdl2uRv:-OJtN3z28xNyhuxX463g:b-jicntv) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/1d269e028497f298f03012d172d27b8cadb6b4f8/src/BloomBrowserUI/bookEdit/toolbox/talkingBook/AdjustTimingsDialog.tsx#L111)):*

The comment would be clearer if the final "this" was replaced by "shouldAdjustSegments" since more is happening than just this operation, and setEndTimes is a useState.  (Is the import of this comment even true at this point?)
</blockquote></details>

Done. (In fact you convinced me there's no point in messing with useRef)
___
*[`src/BloomBrowserUI/bookEdit/toolbox/talkingBook/AdjustTimingsDialog.tsx` line 200 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6860#-OJsnfVLALDv5MoDGKCH:-OJtYKeU5HJb0SPs1Kgd:baakpzo) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/1d269e028497f298f03012d172d27b8cadb6b4f8/src/BloomBrowserUI/bookEdit/toolbox/talkingBook/AdjustTimingsDialog.tsx#L200)):*

Is it ever possible for ff to be a fallback list rather than a single font?  (I don't think so, without user custom CSS, but it may be worth asking.)
</blockquote></details>

I have no idea why I wrote this line. Any font family specification in getComputedStyle should work as a line in a stylesheet element, so why change it?
___
*[`src/BloomBrowserUI/bookEdit/toolbox/talkingBook/AdjustTimingsDialog.tsx` line 495 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6860#-OJst3cBEi1au2LLQ-z-:-OJtbaAH7sAAyfY3Ljxs:b-ecf5o0) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/1d269e028497f298f03012d172d27b8cadb6b4f8/src/BloomBrowserUI/bookEdit/toolbox/talkingBook/AdjustTimingsDialog.tsx#L495)):*

So if the end data is incomplete relative to the text segments, we assume the segment is 1 second long.  Am I interpreting the code correctly?  A short comment might be in order, something like "if no timing data, default segment length is set to 1 second".
</blockquote></details>

Actually the code as written does not handle differing lengths, just the possibility that endTimes is falsy (null or undefined). The calling code does its best to handle problematic endTimes (including the wrong number of them), mostly by reverting to our default timings. So I put in an assert and took out the fallback here.
___
*[`src/BloomBrowserUI/react_components/localizableMenuItem.tsx` line 39 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6860#-OJsuQeg4NmzrVyt1x8z:-OJtcc482waPlIJKz1s3:bcq772m) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/1d269e028497f298f03012d172d27b8cadb6b4f8/src/BloomBrowserUI/react_components/localizableMenuItem.tsx#L39)):*

The priority order appears to be 1) entire node, 2) actual string to show (already localized), and 3) l10n ID to look up.  This comment implies 1) entire node, 2) l10n ID to look up, and 3) actual string to show.  Admittedly, this is the order of the parameters, but something should be adjusted for consistency.
</blockquote></details>

I didn't intend to change the previous behavior when the new option is missing, so I've fixed the comment here.
___
*[`src/BloomBrowserUI/react_components/localizableMenuItem.tsx` line 159 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6860#-OJsvJkZ5po2Jm7mDtpE:-OJtcifU6kYLv1L6tIPR:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/1d269e028497f298f03012d172d27b8cadb6b4f8/src/BloomBrowserUI/react_components/localizableMenuItem.tsx#L159)):*

Are these in the right order?  (See my previous comment.)
</blockquote></details>

Done.

StephenMcConnel (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6860)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @JohnThomson)

PR #6865, opened 2025-02-21
Handle a corner case in reporting Font info (BL-14357)

JohnThomson:
There was a problem where a book has differing default fonts for different languages and use a style (like ImageDescription) that has no definition (or possibly one that specifies no font info).

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6865)


andrew-polk (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6865)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @JohnThomson)

PR #6868, opened 2025-02-24
Fix positioning of GamesPromptDialog tooltip (BL-14347)

JohnThomson:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6868)

### DIFF
--- a/src/BloomBrowserUI/bookEdit/js/editableDivUtils.ts
+++ b/src/BloomBrowserUI/bookEdit/js/editableDivUtils.ts
@@ -193,8 +193,17 @@ export class EditableDivUtils {
     }
 
     // look for an existing transform:scale setting and extract the scale. If not found, use 1.0 as starting point.
-    public static getPageScale(): number {
+    // If target is supplied, we only want a scale based on the page scaler if the target is inside it.
+    public static getPageScale(target?: HTMLElement): number {
         const page = this.getPage();
+        let scaler: HTMLElement | undefined = undefined;
+        const getScaler = () => page.find("div#page-scaling-container").get(0);
+        if (target) {
+            scaler = getScaler();
+            if (!scaler || !scaler.contains(target)) {
+                return 1.0;
+            }
+        }
 
         // With full bleed, we have a transform on the page in addition to the possible scaling using the zoom control.
         // This calculation gets the scale experimentally. Because offsetWidth is an integer,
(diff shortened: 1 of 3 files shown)


JohnThomson (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6868)* status: 0 of 3 files reviewed, 1 unresolved discussion

___
*[`src/BloomBrowserUI/bookEdit/sourceBubbles/BloomSourceBubbles.tsx` line 469 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6868#-OJtIyHy5KFbFZQ3vgYg:-OJtIyHy5KFbFZQ3vgYh:b-z6dvsi) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/2cfafac4272959a911a4026f5a5be24e7c6c7a3b/src/BloomBrowserUI/bookEdit/sourceBubbles/BloomSourceBubbles.tsx#L469)):*
> ```XML
>     // Note, in that case we'd want to observe each .bloom-editable, not the whole page.
>     // (The page typically will NOT change size when a text block does.)
>     public static setupSizeChangedHandling(groups: HTMLElement[]) {
> ```

It's a bit hard to see what really changed here. The important things are:
- we're observing the body, not just the bloom-page (therefore including parts of the GamesPromptDialog)
- we're fixing anything that has data-hasqtip, so this no longer depends on getting a complete list of groups to update or that list staying correct
- we're noticing attributes (e.g., elements of element.style like left and right) as well as other changes
- since all that makes it happen more often and take longer, I added some debouncing.

andrew-polk (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6868)* status: all files reviewed, 2 unresolved discussions

___
*[`src/BloomBrowserUI/bookEdit/sourceBubbles/BloomSourceBubbles.tsx` line 466 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6868#-OJtKoCX8LhNfM8UYi6b:-OJtKoCX8LhNfM8UYi6c:bbvqswt) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/2cfafac4272959a911a4026f5a5be24e7c6c7a3b/src/BloomBrowserUI/bookEdit/sourceBubbles/BloomSourceBubbles.tsx#L466)):*
> ```XML
>     // the page.
>     // I don't think we need worry about disposing of this. It's useful until the page is reloaded.
>     // This could be better done with resizeObserver? But we need it for position, too...FixGecko60.
> ```

Might as well go ahead and remove the "FixGecko60" while we're here.

Otherwise, LGTM. Feel free to merge.

PR #6882, opened 2025-02-27
Prevent focus shift between overlays choosing an image (BL-14123)

StephenMcConnel:


JohnThomson (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6882)* status: all files reviewed, 2 unresolved discussions (waiting on @StephenMcConnel)

___
*[a discussion](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6882#-OKXF6RFAZLI9Y6Oz1QQ:-OKXF6RGA8KQxUCzrvIn:br53tie) (no related file):*
I don't like this, but don't immediately have a better idea. 
- it suppresses focus handling undesirably when a menu command does NOT bring up a dialog
- and if someone complains about that a year from now we may have a hard time tracking this down
- it further complicates difficult-to-deal-with focus-managing code
- how do we know that only one should-be-ignored focus change will happen while the dialog is open? Would it mess things up if any of these dialogs launches another dialog?
Would it work to force the active overlay to be restored to the one that was active when the menu was pulled down? Maybe even to restore focus to what was document.activeElement at the time it opened? There are at least two exceptions (we can't put it back after Delete, and shouldn't after Duplicate) so we might need a new field in the menu item to control it. And even that won't handle dialogs.
Could we put a field in menuItem to indicate whether the item launches a dialog? Then only items that do would set skipNextFocusChange to true. Or perhaps, since I think there's a separate call to setMenuOpen() for each item, it could take another argument that controls whether we do this? And only the dialog-launching ones pass true? I've also suggested this below.
Sorry to be negative without a clearly better idea. But I think at least the idea of passing another (optional) arg to setMenuOpen is worth exploring.
___
*[`src/BloomBrowserUI/bookEdit/js/OverlayContextControls.tsx` line 115 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6882#-OKXJZ2c51gkWm6w4ILs:-OKXJZ2c51gkWm6w4ILt:b-5kej1d) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/f46632de4d5458aee81b49c5550ab5609094a591/src/BloomBrowserUI/bookEdit/js/OverlayContextControls.tsx#L115)):*
> ```XML
>         if (!open) {
>             setTimeout(() => {
>                 BubbleManager.skipNextFocusChange = true;
> ```

I think setMenuOpen is called by each of the menu item click handlers. Maybe it could take an optional argument launchingDialog, and only if this is true should we set this?

StephenMcConnel (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6882)* status: 0 of 2 files reviewed, 2 unresolved discussions (waiting on @JohnThomson)

___
*[a discussion](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6882#-OKXF6RFAZLI9Y6Oz1QQ:-OKYQQDMB3JZTGWpoNBp:b-p6gqlc) (no related file):*

I don't like this, but don't immediately have a better idea. 
- it suppresses focus handling undesirably when a menu command does NOT bring up a dialog
- and if someone complains about that a year from now we may have a hard time tracking this down
- it further complicates difficult-to-deal-with focus-managing code
- how do we know that only one should-be-ignored focus change will happen while the dialog is open? Would it mess things up if any of these dialogs launches another dialog?
Would it work to force the active overlay to be restored to the one that was active when the menu was pulled down? Maybe even to restore focus to what was document.activeElement at the time it opened? There are at least two exceptions (we can't put it back after Delete, and shouldn't after Duplicate) so we might need a new field in the menu item to control it. And even that won't handle dialogs.
Could we put a field in menuItem to indicate whether the item launches a dialog? Then only items that do would set skipNextFocusChange to true. Or perhaps, since I think there's a separate call to setMenuOpen() for each item, it could take another argument that controls whether we do this? And only the dialog-launching ones pass true? I've also suggested this below.
Sorry to be negative without a clearly better idea. But I think at least the idea of passing another (optional) arg to setMenuOpen is worth exploring.
</blockquote></details>

I've added the optional argument and used it in the places I saw where it may be needed.
___
*[`src/BloomBrowserUI/bookEdit/js/OverlayContextControls.tsx` line 115 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6882#-OKXJZ2c51gkWm6w4ILs:-OKYQKbMBbN1s0YBlPr0:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/f46632de4d5458aee81b49c5550ab5609094a591/src/BloomBrowserUI/bookEdit/js/OverlayContextControls.tsx#L115)):*

I think setMenuOpen is called by each of the menu item click handlers. Maybe it could take an optional argument launchingDialog, and only if this is true should we set this?
</blockquote></details>

Done.

JohnThomson (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6882)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @StephenMcConnel)

PR #6884, opened 2025-03-03
Upgrade storybook to 8.6.2, repairing build

JohnThomson:
Andrew and I worked together on much of this, although a lot was also done by copilot without much supervision. I think only the last two commits should be considered unreviewed. Unfortunately, most of the second-last commit is Copilot generating new files by splitting up existing ones; it's hard to compare the results. I haven't even tried myself to review how it did that, but the results seem to work much as in earlier versions.

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6884)


JohnThomson:
Many of the commits bypassed hooks.

PR #6893, opened 2025-03-06
Allow a branding to be replaced by a newer subscription (BL-14381)

StephenMcConnel:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6893)

### DIFF
--- a/src/BloomExe/Collection/CollectionSettingsDialog.cs
+++ b/src/BloomExe/Collection/CollectionSettingsDialog.cs
@@ -425,26 +425,23 @@ namespace Bloom.Collection
             _collectionSettings.PageNumberStyle = PendingNumberingStyle; // non-localized key
 
             var oldBrand = _collectionSettings.BrandingProjectKey;
-            if (oldBrand != _brand || _collectionSettings.IgnoreExpiration)
-            {
-                _collectionSettings.BrandingProjectKey = _brand;
-                _collectionSettings.SubscriptionCode = _subscriptionCode;
-                // An early version of this code allowed a download-for-edit collection to be unlocked once a valid
-                // branding code was provided. We decided not to do that, but here is the code we'd reinstate if we change our minds.
-                // We've definitely selected some branding, even if it's the default. If necessary, we have a valid code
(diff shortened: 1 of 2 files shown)


hatton (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6893)* status: 0 of 1 files reviewed, 1 unresolved discussion

___
*[`src/BloomExe/Collection/CollectionSettingsDialog.cs` line 428 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6893#-OKiNEkd6Nbp8lTttEjj:-OKiNEkd6Nbp8lTttEjk:b-gr9jrh) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/a6478a114fcb4bac75184528b2f1eb1103a570f3/src/BloomExe/Collection/CollectionSettingsDialog.cs#L428)):*
> ```Smalltalk
> 
>             var oldBrand = _collectionSettings.BrandingProjectKey;
>             var oldDate = CollectionSettingsApi.GetExpirationDate(_collectionSettings.SubscriptionCode);
> ```

What I'm not seeing is why we are checking anything before assigning the values.

StephenMcConnel (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6893)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @hatton)

___
*[`src/BloomExe/Collection/CollectionSettingsDialog.cs` line 428 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6893#-OKiNEkd6Nbp8lTttEjj:-OL1BK4Y9nM1gvdjYb4B:bfrc8jb) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/a6478a114fcb4bac75184528b2f1eb1103a570f3/src/BloomExe/Collection/CollectionSettingsDialog.cs#L428)):*

What I'm not seeing is why we are checking anything before assigning the values.
</blockquote></details>

Done.  I removed the check.  I also fixed what looked like an edge-case bug when updating the subscription.

PR #6896, opened 2025-03-11
Improve display of bilingual page in Add Page Dialog (BL-14364)

StephenMcConnel:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6896)

### DIFF
--- a/src/BloomBrowserUI/pageChooser/PageChooserDialog.tsx
+++ b/src/BloomBrowserUI/pageChooser/PageChooserDialog.tsx
@@ -659,6 +659,11 @@ export const PageChooserDialog: React.FunctionComponent<IPageChooserDialogProps>
                             pageIsEnterpriseOnly={selectedTemplatePageDiv.classList.contains(
                                 "enterprise-only"
                             )}
+                            pageIsMarkedBilingual={
+                                selectedTemplatePageDiv.getAttribute(
+                                    "data-ui-mark-bilingual"
+                                ) === "true"
+                            }
                             pageIsDigitalOnly={isDigitalOnly(
                                 selectedTemplatePageDiv
                             )}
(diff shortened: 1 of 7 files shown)


andrew-polk (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6896)* status: all files reviewed, 1 unresolved discussion (waiting on @StephenMcConnel)

___
*[`src/content/templates/template books/standard-page-mixins.pug` line 96 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6896#-OL5yiHV9_jybGoOXQEX:-OL5yiHW0S0M8MXV8lHs:b-j41g3o) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/46ea6231d2ede3985e7323e8fd06ac4235a247a4/src/content/templates/template%20books/standard-page-mixins.pug#L96)):*
> ```Pug
>            +lower(30)
>              +fieldx
>    +page-splittable-choice-custom('Bilingual & Picture in Middle', 'A page with one language above the image and another below it.')#adcd48df-e9ab-4a07-afd4-6a24d039838A&attributes(attributes).mark-bilingual
> ```

If I'm not mistaken, this class will be persisted on the page in the book html.
It feels like that, then, warrants a different name. But I'm having a hard time coming up with the right thing.
`bilingual-page` feels off because the user can make other bilingual pages, and they can even change this to not be bilingual.
`bilingual-template` is similarly wrong and once it gets into the dom, this is no longer a template.
`created-as-bilingual` is accurate, but I'm still not excited about it.

And then there is the question of whether the class name should start with "bloom-". I know we haven't been consistent about this at all....

On another tack, is there a different way to indicate to the page chooser that the page should be marked a certain way other than a class name? Something which doesn't persist in the book?

andrew-polk (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6896)* status: all files reviewed, 1 unresolved discussion (waiting on @StephenMcConnel)

PR #6910, opened 2025-03-17
Hide overlay canvas elements for Image Description Tool (BL-14312)

StephenMcConnel:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6910)


StephenMcConnel:
The actual bug is totally different than what was written up.

PR #6913, opened 2025-03-18
Prevent talking book markup on game pages (BL-14434)

StephenMcConnel:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6913)

### DIFF
--- a/src/BloomBrowserUI/bookEdit/toolbox/talkingBook/audioRecording.ts
+++ b/src/BloomBrowserUI/bookEdit/toolbox/talkingBook/audioRecording.ts
@@ -42,7 +42,7 @@ import {
 } from "../../../utils/bloomApi";
 import * as toastr from "toastr";
 import WebSocketManager from "../../../utils/WebSocketManager";
-import { ToolBox } from "../toolbox";
+import { getActiveToolId, ToolBox } from "../toolbox";
 import * as React from "react";
 import * as ReactDOM from "react-dom";
 import {
@@ -939,6 +939,16 @@ export default class AudioRecording implements IAudioRecorder {
     public async setSoundAndHighlightAsync(
         setHighlightParams: ISetHighlightParams
     ): Promise<void> {
+        // Check that the active tool is not something other than "talkingBook".  A page
+        // can specify a tool it wants to be active initially when it first loads.
+        // The multiple asynchronous calls during page loading can result in the wrong
+        // tool getting the "newPageReady" method called when the new page specifies a
+        // different tool to be activated.  This is the simplest fix that I've found.
+        // See BL-14434.
+        const activeToolId = getActiveToolId();
+        const notInTalkingBook = activeToolId && activeToolId !== "talkingBook";
+        if (notInTalkingBook) return;
+
         // Note: setHighlightToAsync() should be run first so that ui-audioCurrent points to the correct element when setSoundFrom() is run.
         await this.setHighlightToAsync(setHighlightParams);
         this.setSoundFrom(setHighlightParams.newElement);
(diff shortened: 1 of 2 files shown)


andrew-polk (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6913)* status: all files reviewed, 1 unresolved discussion (waiting on @StephenMcConnel)

___
*[`src/BloomBrowserUI/bookEdit/toolbox/talkingBook/audioRecording.ts` line 990 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6913#-OLfRKF90tkoH_0m1PIu:-OLfRKF90tkoH_0m1PIv:bvmpmrt) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/f932b3efc1e6e5f7143dd6e878d3e8467d56493b/src/BloomBrowserUI/bookEdit/toolbox/talkingBook/audioRecording.ts#L990)):*
> ```XML
>         // variable in the outer if statement, gets Bloom into a wierd frozen state. The
>         // innocent looking "visible" is a method that has some necessary side-effects,
>         // which is poor programming practice, but I'm saving that for another day.)
> ```

I think the fix looks ok, but there's a couple things about the comment:

typo "weird"

You reference the variable "notInTalkingBook" but I think you ended up calling it "mayBeInTalkingBook"

I don't understand the comment about visible being a method. As far as I can tell, it is a boolean which is the result of calling isVisible above.

StephenMcConnel (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6913)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @andrew-polk)

___
*[`src/BloomBrowserUI/bookEdit/toolbox/talkingBook/audioRecording.ts` line 990 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6913#-OLfRKF90tkoH_0m1PIu:-OLj1uRfCrkGBkzx_KUh:bcx4vxy) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/f932b3efc1e6e5f7143dd6e878d3e8467d56493b/src/BloomBrowserUI/bookEdit/toolbox/talkingBook/audioRecording.ts#L990)):*

I think the fix looks ok, but there's a couple things about the comment:

typo "weird"

You reference the variable "notInTalkingBook" but I think you ended up calling it "mayBeInTalkingBook"

I don't understand the comment about visible being a method. As far as I can tell, it is a boolean which is the result of calling isVisible above.
</blockquote></details>

The fix should have been in the calling method.  And looking only at the current page prevents audio markup even when the talking book tool is activated.  (which is what I thought of last night)  I don't know why the weird frozen state didn't always happen before with this approach, or why I have trouble spelling weird.  Except maybe for the old spelling rule "I before E except after C and when sounded as A in neighbor and weigh", which obviously has at least one exception.  :-)

andrew-polk (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6913)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @StephenMcConnel)

___
*[`src/BloomBrowserUI/bookEdit/toolbox/talkingBook/audioRecording.ts` line 990 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6913#-OLfRKF90tkoH_0m1PIu:-OLjMdqi21PVn9Qyl5jN:b-fnsq25) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/f932b3efc1e6e5f7143dd6e878d3e8467d56493b/src/BloomBrowserUI/bookEdit/toolbox/talkingBook/audioRecording.ts#L990)):*

The fix should have been in the calling method.  And looking only at the current page prevents audio markup even when the talking book tool is activated.  (which is what I thought of last night)  I don't know why the weird frozen state didn't always happen before with this approach, or why I have trouble spelling weird.  Except maybe for the old spelling rule "I before E except after C and when sounded as A in neighbor and weigh", which obviously has at least one exception.  :-)
</blockquote></details>

You have to append, "and except weird words like, 'weird'"

PR #6918, opened 2025-03-19
Keep text inside canvas overlays when editing image description (BL-14312)

StephenMcConnel:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6918)

### DIFF
--- a/src/BloomBrowserUI/bookEdit/js/CanvasElementManager.ts
+++ b/src/BloomBrowserUI/bookEdit/js/CanvasElementManager.ts
@@ -161,8 +161,10 @@ export class CanvasElementManager {
             return false;
         }
 
-        wrapperBox.style.height = newHeight + "px"; // next line will change to percent
-        CanvasElementManager.convertTextboxPositionToAbsolute(
+        wrapperBox.style.height = newHeight + "px";
+        // The next method call will change from % positioning to px if needed.  Bloom originally
+        // used % values to position canvas elements before we realized that was a bad idea.
+        CanvasElementManager.convertCanvasElementPositionToAbsolute(
             wrapperBox,
             container
         );
@@ -254,11 +256,6 @@ export class CanvasElementManager {
                 wrapperBox.style.top = oldTop - delta + "px";
                 const oldLeft = this.pxToNumber(wrapperBox.style.left);
                 wrapperBox.style.left = oldLeft - delta + "px";
-
-                CanvasElementManager.convertTextboxPositionToAbsolute(
-                    wrapperBox,
-                    this.getTopLevelImageContainerElement(wrapperBox)!
-                );
             }
         });
     }
@@ -5284,25 +5281,40 @@ export class CanvasElementManager {
         }
     }
 
-    // Converts a text box's position to absolute in pixels (using CSS styling)
+    // Converts a canvas element's position to absolute in pixels (using CSS styling)
     // (Used to be a percentage of parent size. See comments on setTextboxPosition.)
-    // textBox: The thing we want to position
-    // container: Optional. The image container the text box is in. If this parameter is not defined, the function will automatically determine it.
-    private static convertTextboxPositionToAbsolute(
-        textBox: Element,
+    // canvasElement: The thing we want to position
+    // container: Optional. The image container the convas element is in. If this parameter is not defined, the function will automatically determine it.
+    private static convertCanvasElementPositionToAbsolute(
+        canvasElement: HTMLElement,
         container?: Element | null | undefined
     ): void {
         let unscaledRelativeLeft: number;
         let unscaledRelativeTop: number;
 
+        const left = canvasElement.style.left;
+        const top = canvasElement.style.top;
+        if (left.endsWith("px") && top.endsWith("px")) {
+            // We're already in absolute pixel position.
+            return;
+        }
+
+        // Note: if the convasElement is scaled by a transform applied to an ancestor
+        // element, then the following calculations will be woefully off.  See BL-14312.
+        // We think all such cases will be caught by the check above for already being
+        // in absolute pixel position.  But this is still something worth considering
+        // if canvas elements show up in strange positions.  (Showing image descriptions
+        // was the original case where we discovered this problem, and led to realizing
+        // that most calls to this method are not really needed.)
+
         if (!container) {
             container = CanvasElementManager.getTopLevelImageContainerElement(
-                textBox
+                canvasElement
             );
         }
 
         if (container) {
-            const positionInfo = textBox.getBoundingClientRect();
+            const positionInfo = canvasElement.getBoundingClientRect();
             const wrapperBoxPos = new Point(
                 positionInfo.left,
                 positionInfo.top,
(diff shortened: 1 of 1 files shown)


StephenMcConnel (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6918)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @JohnThomson)

___
*[a discussion](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6918#-OLnnjTO2JG6HmMHMzFK:-OLpJrR98y14bbEOnGCR:b-tjvoo3) (no related file):*

Brilliant work tracking down what was causing the problem! But the fix feels a bit like a bandaid. A 50% transform causes a problem for this method, but probably any other transform:scale would, too. Also, I wonder if there aren't better options.
First, convertTextboxPositionToAbsolute() shouldn't be doing anything in this or any similar situation, because the position is undoubtedly absolute already (that is, it is positioned and sized by top, left, height, and width in px, not %; we haven't done % for many releases). So if we simply detect that textBox.style.height.endsWith("px") (and maybe width, left, and top too), we could exit early and save time in many other situations as well as this one.
Moving outwards, I'm not convinced that growOverflowingBox needs to call convertTextboxPositionToAbsolute() at all. It happens right after we set a height in px and an obsolete comment indicates that it replaces a previous call expected to change the px to %. Seems like it should have been removed rather than changed to forcing absolute. (Though, conversely, I'm surprised we don't appear to call convertTextboxPositionToAbsolute() somewhere in the process of setting up a page. Maybe this call is what does that, since I think we check overflow every time we open a page). So maybe we could fix it by just deleting that call! But would need to test on an old book that still uses percent positioning; it might be important to convert all positions to absolute while launching the page. (For example, if you go this way try cropping a canvas element image whose size and position start out as percents. I think that code assumes px size and position.)
Then there is the issue that we're not really changing the size of anything, just applying a transform. It feels like that should not need to trigger an overflow check at all. I'm having trouble figuring out why it does; it appears that we're triggering code that's meant to run when we move the splitter in origami. It seems like (a) we're not changing the size of the .split-pane-component-inner, which is what the handler is attached to in AddOverflowHandlers(), so the resize event should not be raised at all, and (b) an overflow check triggered by an origami split change may need to check for overflow on editables not in canvas elements, but it shouldn't change the size of any canvas elements, so they should not even need checking.
I'm not sure which of these would be best to fix...maybe more than one. I think I would start by seeing if we can fix it by removing the call to convertTextboxPositionToAbsolute() in growOverflowingBox(), even if we have to replace it with a new call somewhere in page bootstrapping (that happens before we add the image description transform!). But I'm pretty sure there is some fix that will simplify our code rather than making it more complicated.
BTW, while you're in this area, convertTextboxPositionToAbsolute is misnamed. It should be convertCanvasElementPositionToAbsolute, and the first argument is a canvasElement, not a textBox.
Feel free to push back if none of these ideas works out or they turn out to complicate things or are a lot of work.
</blockquote></details>

Some debugging output from the two consecutive calls to this method, the first at full scale and the second with the transform in place:
```
 convertToAbsolute: already absolute - left=277.641px, top=11.625px
 convertToAbsolute: new left=277.640625, top=11.625

 convertToAbsolute: already absolute - left=277.641px, top=11.625px
 convertToAbsolute: new left=138.8203125, top=112.9375
```
Given these results, I think it's safe to put the check in place for style.left and style.top both ending in "px" to short-circuit this method, so I've done that instead of my more complicated code looking for a transform on a specific type of ancestor element.  Also, one of the two places where this method was called was explicitly setting these left and top with px values just before the call, so I deleted that call.
That wasn't the call in growOverflowingBox(), however, and that method didn't seem to touch left and top.  I'm hesitant to disturb this code more.  Testing indicates that % left and top can arrive here, so it's on the path of getting things fixed.  Perhaps a better path could be devised but I don't want to tackle that in the scope of this issue.

JohnThomson (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6918)* status: all files reviewed, 3 unresolved discussions (waiting on @StephenMcConnel)

___
*[a discussion](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6918#-OLnnjTO2JG6HmMHMzFK:-OLphPtK0--mcTvMfY0u:bmsr3px) (no related file):*

Some debugging output from the two consecutive calls to this method, the first at full scale and the second with the transform in place:
```
 convertToAbsolute: already absolute - left=277.641px, top=11.625px
 convertToAbsolute: new left=277.640625, top=11.625

 convertToAbsolute: already absolute - left=277.641px, top=11.625px
 convertToAbsolute: new left=138.8203125, top=112.9375
```
Given these results, I think it's safe to put the check in place for style.left and style.top both ending in "px" to short-circuit this method, so I've done that instead of my more complicated code looking for a transform on a specific type of ancestor element.  Also, one of the two places where this method was called was explicitly setting these left and top with px values just before the call, so I deleted that call.
That wasn't the call in growOverflowingBox(), however, and that method didn't seem to touch left and top.  I'm hesitant to disturb this code more.  Testing indicates that % left and top can arrive here, so it's on the path of getting things fixed.  Perhaps a better path could be devised but I don't want to tackle that in the scope of this issue.
</blockquote></details>

I think that's OK as long as you're sure they WILL be already-fixed by the time we are checking overflow after opening the toolbox. You don't think that's something that could be a race condition? I'm guessing that CheckForOverflowSoon() gets called as part of page load (usually switching everything to px), and we know it is called as part of turning on the toolbox. Both calls involve a 1s delay and aborting the first request if the second occurs within that second. So my guess would be that if a page with percentage sizes opens with this toolbox active, the first actual calls to convertToAbsolute will see percentage sizes.
But that's wildly rare condition, and the fix will just be to turn the tool off and on again. I think we can live with it.
Couple of minor suggestions is all if nothing here worries you.
___
*[`src/BloomBrowserUI/bookEdit/js/CanvasElementManager.ts` line 5287 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6918#-OLpgix0EJleLYNsJWQ2:-OLpgix0EJleLYNsJWQ3:b-pvg9of) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/c12ab58ca7b67ce3b42bbfa1735e1f62c41c1a82/src/BloomBrowserUI/bookEdit/js/CanvasElementManager.ts#L5287)):*
> ```XML
>     // (Used to be a percentage of parent size. See comments on setTextboxPosition.)
>     // canvasElement: The thing we want to position
>     // container: Optional. The image container the text box is in. If this parameter is not defined, the function will automatically determine it.
> ```

Sorry, you missed one "text box", which could make it confusing.
___
*[`src/BloomBrowserUI/bookEdit/js/CanvasElementManager.ts` line 5298 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6918#-OLpgp660R6cPEcWXRcz:-OLpgp660R6cPEcWXRd-:b-wgsni5) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/c12ab58ca7b67ce3b42bbfa1735e1f62c41c1a82/src/BloomBrowserUI/bookEdit/js/CanvasElementManager.ts#L5298)):*
> ```XML
>         const top = canvasElement.style.top;
>         if (left.endsWith("px") && top.endsWith("px")) {
>             // We're already in absolute pixel position.
> ```

Let's still capture in a comment that it's important NOT to do the following adjustment if we're scaled (as in showing image descriptions).

PR #6931, opened 2025-03-26
Back migration for level 7 canvasElement

JohnThomson:
This should be null-merged into master (don't want any of these changes) and eventually cherry-picked to 6.0

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6931)

### DIFF
--- a/src/BloomExe/Book/BookStorage.cs
+++ b/src/BloomExe/Book/BookStorage.cs
@@ -215,10 +215,11 @@ namespace Bloom.Book
             string folderPath,
             IChangeableFileLocator baseFileLocator,
             BookRenamedEvent bookRenamedEvent,
-            CollectionSettings collectionSettings
+            CollectionSettings collectionSettings,
+            bool isInEditableCollection = false
         )
             : this(
-                new BookInfo(folderPath, false),
+                new BookInfo(folderPath, isInEditableCollection),
                 baseFileLocator,
                 bookRenamedEvent,
                 collectionSettings
@@ -3820,10 +3821,14 @@ namespace Bloom.Book
             if (GetMaintenanceLevel() <= kMaintenanceLevel)
                 return;
             var breakingFeatureRequirements = GetBreakingFeatureRequirements();
-            // Handle the back migrations 6.2 knows about, in the proper reverse order.
-            // This should not be merged to 6.2.
+            // Handle the back migrations this version of Bloom knows about, in the proper reverse order.
+            // This should not be merged to 6.2, though it may be a useful model of how to handle future
+            // back migrations.
             // If you add a new back migration, you must also modify GetHtmlMessageIfFeatureIncompatibility
             // to not complain if the only breaking changes are ones we can back-migrate.
+            // Also, each back migration has the potential to cause problems since the corresponding forward
+            // migration is likely to be run again. Make sure that each migration that might be re-run will
+            // not cause problems if it is run repeatedly, and comment it to reinforce this.
             if (breakingFeatureRequirements.Any(fr => fr.FeatureId == "bloomCanvas"))
             {
                 MigrateBackFromLevel7BloomCanvas();
@@ -3873,12 +3878,12 @@ namespace Bloom.Book
             );
             if (GetMaintenanceLevel() <= 6)
                 return;
-            var legacyCanvasElements = Dom.SafeSelectNodes("//*[contains(@class, 'bloom-canvas')]")
+            var bloomCanvases = Dom.SafeSelectNodes("//*[contains(@class, 'bloom-canvas')]")
                 .Cast<SafeXmlElement>()
                 // the crude xpath above will also match bloom-canvas-element, which we don't want to change
                 .Where(e => e.HasClass("bloom-canvas"))
                 .ToList();
-            foreach (var element in legacyCanvasElements)
+            foreach (var element in bloomCanvases)
             {
                 element.RemoveClass("bloom-canvas");
                 element.AddClass("bloom-imageContainer");
--- a/src/BloomExe/Book/BookStorage.cs
+++ b/src/BloomExe/Book/BookStorage.cs
@@ -535,8 +535,9 @@ namespace Bloom.Book
 
             // For 6.1 and 6.0 only; don't merge into 6.2
             if (
-                breakingFeatureRequirements.Count() == 1
-                && breakingFeatureRequirements.First().FeatureId == "canvasElement"
+                breakingFeatureRequirements.All(
+                    fr => fr.FeatureId == "canvasElement" || fr.FeatureId == "bloomCanvas"
+                )
             )
                 return "";
 
@@ -3819,18 +3820,17 @@ namespace Bloom.Book
             if (GetMaintenanceLevel() <= kMaintenanceLevel)
                 return;
             var breakingFeatureRequirements = GetBreakingFeatureRequirements();
-            // Building a general mechanism here, but this is the only one 6.1 currently
-            // knows how to do. (Note that this method doesn't get called at all if there
-            // are breaking feature requirements we don't know how to deal with.)
-            foreach (var requirement in breakingFeatureRequirements)
+            // Handle the back migrations 6.2 knows about, in the proper reverse order.
+            // This should not be merged to 6.2.
+            // If you add a new back migration, you must also modify GetHtmlMessageIfFeatureIncompatibility
+            // to not complain if the only breaking changes are ones we can back-migrate.
+            if (breakingFeatureRequirements.Any(fr => fr.FeatureId == "bloomCanvas"))
             {
-                // These two lines should be removed merging to master, since not wanted in
-                // 6.2 and beyond. That will leave this method doing nothing until we have some
-                // new reason for a back migration, but I think it's worth keeping.
-                if (requirement.FeatureId == "canvasElement")
-                    MigrateBackFromLevel5CanvasElement();
-                // If you add a new back migration, you must also modify GetHtmlMessageIfFeatureIncompatibility
-                // to not complain if the only breaking changes are ones we can back-migrate.
+                MigrateBackFromLevel7BloomCanvas();
+            }
+            if (breakingFeatureRequirements.Any(fr => fr.FeatureId == "canvasElement"))
+            {
+                MigrateBackFromLevel5CanvasElement();
             }
         }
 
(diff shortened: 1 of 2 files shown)


andrew-polk (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6931)* status: all files reviewed, 5 unresolved discussions (waiting on @JohnThomson)

___
*[`src/BloomExe/Book/BookStorage.cs` line 3823 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6931#-OMO_-Q8AkWM-t1u62_F:-OMO_-Q8AkWM-t1u62_G:bvf10ip) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/5fac22f93e653d411e73adcfbfde248cfcdef8e9/src/BloomExe/Book/BookStorage.cs#L3823)):*
> // Handle the back migrations 6.2 knows about

Shouldn't that say the ones 6.1 and 6.0 knows about? Or better, something like

> // Handle the back migrations this version of Bloom knows about
___
*[`src/BloomExe/Book/BookStorage.cs` line 3834 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6931#-OMOa6ZTEJpyxLru9PTh:-OMOa6ZU2sKpeytoyLm_:bvt9w71) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/5fac22f93e653d411e73adcfbfde248cfcdef8e9/src/BloomExe/Book/BookStorage.cs#L3834)):*
> ```Smalltalk
>             {
>                 MigrateBackFromLevel5CanvasElement();
>             }
> ```

I think we may have a problem with this back migration mechanism.

If I'm thinking clearly, since we back migrate 7 to 6, then 6 to 4, we didn't undo the to-6 migration. So the to-6 migration remains but will happen again because this book is now level 4.

If our to-6 migration is able to be run twice on the same book, fine, but I don't think we wrote migrations that way.
___
*[`src/BloomExe/Book/BookStorage.cs` line 3876 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6931#-OMOavFuC-03vPVQjkKw:-OMOavFuC-03vPVQjkKx:b8a0mlb) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/5fac22f93e653d411e73adcfbfde248cfcdef8e9/src/BloomExe/Book/BookStorage.cs#L3876)):*
> ```Smalltalk
>             if (GetMaintenanceLevel() <= 6)
>                 return;
>             var legacyCanvasElements = Dom.SafeSelectNodes("//*[contains(@class, 'bloom-canvas')]")
> ```

I'm not understanding why you've labeled them as "legacy".

And aren't these bloomCanvases? (not canvas elements?)
___
*[`src/BloomExe/Book/BookStorage.cs` line 3892 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6931#-OMObR3Y4KkSvyMcH8AU:-OMObR3Y4KkSvyMcH8AV:beitjci) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/5fac22f93e653d411e73adcfbfde248cfcdef8e9/src/BloomExe/Book/BookStorage.cs#L3892)):*
> ```Smalltalk
>                     bgImage = element.OwnerDocument.CreateElement("img");
>                     bgImage.SetAttribute("src", "placeHolder.png");
>                     element.InsertBefore(bgImage, element.FirstChild);
> ```

Is this ok if there are no children of element?
___
*[`src/BloomExe/Book/BookStorage.cs` line 3896 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6931#-OMObVkh2ofpvpAyxZwC:-OMObVkh2ofpvpAyxZwD:b-jqoe5v) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/5fac22f93e653d411e73adcfbfde248cfcdef8e9/src/BloomExe/Book/BookStorage.cs#L3896)):*
> ```Smalltalk
>             }
>             Dom.UpdateMetaElement("maintenanceLevel", "6");
>         }
> ```

Seems worth adding a unit test or two.

JohnThomson (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6931)* status: 0 of 2 files reviewed, 5 unresolved discussions (waiting on @andrew-polk)

___
*[`src/BloomExe/Book/BookStorage.cs` line 3823 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6931#-OMO_-Q8AkWM-t1u62_F:-OMTgDWSE3fWuG1ieoZY:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/5fac22f93e653d411e73adcfbfde248cfcdef8e9/src/BloomExe/Book/BookStorage.cs#L3823)):*

> // Handle the back migrations 6.2 knows about

Shouldn't that say the ones 6.1 and 6.0 knows about? Or better, something like

> // Handle the back migrations this version of Bloom knows about
</blockquote></details>

Done.
___
*[`src/BloomExe/Book/BookStorage.cs` line 3834 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6931#-OMOa6ZTEJpyxLru9PTh:-OMTgOHq00CWnEfgb3wR:b-l21o22) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/5fac22f93e653d411e73adcfbfde248cfcdef8e9/src/BloomExe/Book/BookStorage.cs#L3834)):*

I think we may have a problem with this back migration mechanism.

If I'm thinking clearly, since we back migrate 7 to 6, then 6 to 4, we didn't undo the to-6 migration. So the to-6 migration remains but will happen again because this book is now level 4.

If our to-6 migration is able to be run twice on the same book, fine, but I don't think we wrote migrations that way.
</blockquote></details>

The level 6 migration sets the data-activity attribute of simple-checkbox-quiz and the data-tool-id of those and also elements with data-activity simple-dom-choice.
Level 5 migration finds elements with class bloom-textOverPicture and changes it to bloom-canvas-element, and ones with class hasOverlay and changes them to bloom-has-canvas-element, and ones with the attribute data-bubble-id and replaces it with data-draggable-id.

I don't think any of those will hurt (except wasting a little time) if we run them again. Nor will it hurt 6.1 to have the extra attributes on games.
I added to the test list to try going back and forth several times.
We will have to be careful that any future migrations which don't forbid back-migrations are repeatable. But this is temporary...not much of a worry once 6.2 reaches release.
___
*[`src/BloomExe/Book/BookStorage.cs` line 3876 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6931#-OMOavFuC-03vPVQjkKw:-OMguI-72T60XHI3Tq40:b-qos20d) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/5fac22f93e653d411e73adcfbfde248cfcdef8e9/src/BloomExe/Book/BookStorage.cs#L3876)):*

I'm not understanding why you've labeled them as "legacy".

And aren't these bloomCanvases? (not canvas elements?)
</blockquote></details>

Done. (I think this code was copied from similar code that does the forward migration).
___
*[`src/BloomExe/Book/BookStorage.cs` line 3892 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6931#-OMObR3Y4KkSvyMcH8AU:-OMgv04lAPRbpATmtolz:b-l6jnpb) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/5fac22f93e653d411e73adcfbfde248cfcdef8e9/src/BloomExe/Book/BookStorage.cs#L3892)):*

Is this ok if there are no children of element?
</blockquote></details>

Yes. FirstChild returns null, and InsertBefore does an append.
___
*[`src/BloomExe/Book/BookStorage.cs` line 3896 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6931#-OMObVkh2ofpvpAyxZwC:-OMh0Eek50hpR7jGebmv:b-svqzvx) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/5fac22f93e653d411e73adcfbfde248cfcdef8e9/src/BloomExe/Book/BookStorage.cs#L3896)):*

Seems worth adding a unit test or two.
</blockquote></details>

Not sure about that for code we don't plan to keep. But, easier to do than argue.

andrew-polk (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6931)* status: all files reviewed, 1 unresolved discussion (waiting on @JohnThomson)

___
*[`src/BloomExe/Book/BookStorage.cs` line 3834 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6931#-OMOa6ZTEJpyxLru9PTh:-OMh93JOAeRtWT39FipA:b-zeu7ef) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/5fac22f93e653d411e73adcfbfde248cfcdef8e9/src/BloomExe/Book/BookStorage.cs#L3834)):*

The level 6 migration sets the data-activity attribute of simple-checkbox-quiz and the data-tool-id of those and also elements with data-activity simple-dom-choice.
Level 5 migration finds elements with class bloom-textOverPicture and changes it to bloom-canvas-element, and ones with class hasOverlay and changes them to bloom-has-canvas-element, and ones with the attribute data-bubble-id and replaces it with data-draggable-id.

I don't think any of those will hurt (except wasting a little time) if we run them again. Nor will it hurt 6.1 to have the extra attributes on games.
I added to the test list to try going back and forth several times.
We will have to be careful that any future migrations which don't forbid back-migrations are repeatable. But this is temporary...not much of a worry once 6.2 reaches release.
</blockquote></details>

I'm more concerned that when we write migrations, we are writing them with the explicit expectation they only happen once. 

So the mechanism of going back, knowing that we have not undone a migration but will run it again causes me concern.

If we are going to leave things as they are, I think we at least owe it to our future selves to 
1. Document here that we are breaking the assumption that migrations only happen once for any migrations which happen between back migrations. This comment should get merged to 6.2.
2. Document on the to-6 migration that it must be kept safe to run multiple times on the same book.

JohnThomson (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6931)* status: 1 of 2 files reviewed, 1 unresolved discussion (waiting on @andrew-polk)

___
*[`src/BloomExe/Book/BookStorage.cs` line 3834 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6931#-OMOa6ZTEJpyxLru9PTh:-OMlZ4Sk622gGuWQ_C6N:bp7wu2g) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/5fac22f93e653d411e73adcfbfde248cfcdef8e9/src/BloomExe/Book/BookStorage.cs#L3834)):*

I'm more concerned that when we write migrations, we are writing them with the explicit expectation they only happen once. 

So the mechanism of going back, knowing that we have not undone a migration but will run it again causes me concern.

If we are going to leave things as they are, I think we at least owe it to our future selves to 
1. Document here that we are breaking the assumption that migrations only happen once for any migrations which happen between back migrations. This comment should get merged to 6.2.
2. Document on the to-6 migration that it must be kept safe to run multiple times on the same book.
</blockquote></details>

I think I've done what you want. Though I don't see as much need for these comments. We almost never modify an existing migration function, because (for example) if we change what the 6-to-7 migration does, we will have to contend with books that claim to be level 7, but actually are only in the state that the old 6-to-7 migration produced. So if things are good when we write the back-migration, they are likely to stay that way.

andrew-polk (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6931)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @JohnThomson)

PR #6947, opened 2025-04-02
Fix remaining issues with subscription/branding (BL-14379, BL-14504)

andrew-polk:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6947)

### DIFF
--- a/DistFiles/localization/en/Bloom.xlf
+++ b/DistFiles/localization/en/Bloom.xlf
@@ -4735,9 +4735,9 @@ Do you want to go ahead?</note>
         <note>ID: PublishTab.UploadCollisionDialog.AlreadyIn</note>
         <note>This is the header for the book that is in bloomlibrary.org already.</note>
       </trans-unit>
-      <trans-unit id="PublishTab.UploadCollisionDialog.ChangeBranding">
-        <source xml:lang="en">The branding was "{0}" but is now "{1}". This may change logos and other material. Check this box if this is what you want.</source>
-        <note>ID: PublishTab.UploadCollisionDialog.ChangeBranding</note>
+      <trans-unit id="PublishTab.UploadCollisionDialog.ChangeSubscription">
+        <source xml:lang="en">The subscription was "{0}" but is now "{1}". This may change logos and other material. Check this box if this is what you want.</source>
+        <note>ID: PublishTab.UploadCollisionDialog.ChangeSubscription</note>
       </trans-unit>
       <trans-unit id="PublishTab.UploadCollisionDialog.ChangeUploader">
         <source xml:lang="en">Change the official uploader to {0}. (Bloom Library will hide part of your email address)</source>
@@ -5389,17 +5389,17 @@ is mostly status, which shows on the Collection Tab -->
       <trans-unit id="SubscriptionStatus.ExpiringSoonMessage">
         <source xml:lang="en">Your {0} subscription expires on {1}.</source>
         <note>ID: SubscriptionStatus.ExpiringSoonMessage</note>
-        <note>{0} is the branding project key, {1} is the formatted expiration date</note>
+        <note>{0} is the subscription descriptor, {1} is the formatted expiration date</note>
       </trans-unit>
       <trans-unit id="SubscriptionStatus.ExpiredMessage">
         <source xml:lang="en">Your {0} subscription expired on {1}.</source>
         <note>ID: SubscriptionStatus.ExpiredMessage</note>
-        <note>{0} is the branding project key, {1} is the formatted expiration date</note>
+        <note>{0} is the subscription descriptor, {1} is the formatted expiration date</note>
       </trans-unit>
       <trans-unit id="SubscriptionStatus.DefaultMessage">
         <source xml:lang="en">Using subscription: {0}. Expires {1}</source>
         <note>ID: SubscriptionStatus.DefaultMessage</note>
-        <note>{0} is the branding project key, {1} is the formatted expiration date</note>
+        <note>{0} is the subscription descriptor, {1} is the formatted expiration date</note>
       </trans-unit>
 
 
--- a/DistFiles/localization/en/BloomLowPriority.xlf
+++ b/DistFiles/localization/en/BloomLowPriority.xlf
@@ -72,7 +72,7 @@
       <trans-unit id="Settings.Subscription.NeedsCode" sil:dynamic="true">
         <source xml:lang="en">This project previously used **%0**. To continue to use this you will need to enter a current subscription code. </source>
         <note>ID: Settings.Subscription.NeedsCode</note>
-        <note>**%0** should be kept. It will become the branding name. Keep something like [here](%1), but replace the word "here" as appropriate. This will become a hyperlink that launches a web page in a browser.</note>
+        <note>**%0** should be kept. It will become the subscription descriptor.</note>
       </trans-unit>
       <trans-unit id="Settings.Subscription.IntroText">
         <source xml:lang="en">To help cover a portion of the costs associated with providing Bloom, we offer [advanced features](%0) and customizations as a subscription service.</source>
@@ -121,7 +121,7 @@
       <trans-unit id="Settings.Subscription.NeedsCode">
         <source xml:lang="en">In an older version of Bloom, this project used **%0**. To continue to use this you will need to enter a current subscription code.</source>
         <note>ID: Settings.Subscription.NeedsCode</note>
-        <note>%0 will be replaced with the name of the branding. Keep surrounding **...**. [here](%1) will become a hot link; %1 will not be seen, but will be the URL for the destination page. Translate "here" and keep the [](%1) around it.</note>
+        <note>%0 will be replaced with the name of the subscription. Keep surrounding **...**.</note>
       </trans-unit>
       <trans-unit id="Settings.Subscription.NotValid">
         <source xml:lang="en">That code appears to be incorrect.</source>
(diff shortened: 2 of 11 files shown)


andrew-polk (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6947)* status: 0 of 4 files reviewed, all discussions resolved

___
*[`src/BloomExe/Publish/BloomLibrary/BloomLibraryPublishModel.cs` line 844 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6947#-OMrVe5vAGt_DYrUQvCj:-OMrVe5vAGt_DYrUQvCk:bm9d6tb) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/0b8c7155c8b5e3e5c23fb7afe12c6dee1c70a9be/src/BloomExe/Publish/BloomLibrary/BloomLibraryPublishModel.cs#L844)):*
> ```Smalltalk
>                 existingBookUrl,
>                 existingThumbUrl,
>                 newBranding = Book.BookInfo.SubscriptionDescriptor,
> ```

The implication here is that if there is a discrepancy between old and new branding when reuploading a book, it will now show the descriptor rather than the branding. I think that's what we want.
I think the only downside is going from Local-Community to Abc-LC will be flagged. But the warning is that the branding may change in the book, and, indeed, it will. It will add "Abc" to the back cover.
___
*[`src/BloomExe/Book/Book.cs` line 1967 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6947#-OMrW43A8a3eIFZQyTK5:-OMrW43A8a3eIFZQyTK6:b-nzj6x) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/0b8c7155c8b5e3e5c23fb7afe12c6dee1c70a9be/src/BloomExe/Book/Book.cs#L1967)):*
> ```Smalltalk
>                 RobustFile.Delete(oldTagsPath);
>             }
>             BookInfo.SubscriptionDescriptor = CollectionSettings.Subscription.Descriptor;
> ```

By getting the Descriptor here rather than the BrandingKey, we skirt the issue of what BrandingKey gives us in the redacted case altogether.
And I've convinced myself that this is what we want in the two places it matters.
1. brandingProjectName in blorg/parse/analytics
2. the upload collision dialog (see below)

andrew-polk (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6947)* status: 1 of 9 files reviewed, 1 unresolved discussion (waiting on @JohnThomson)

___
*[`src/BloomExe/Book/BookInfo.cs` line 1225 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6947#-ON0YpPs78FJ-he1HbeP:-ON0hzVA-k6GpH1ang2i:brjytvb) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/1b9053cbdfbaec81b1bbb595a66975fa20ec43e6/src/BloomExe/Book/BookInfo.cs#L1225)):*

I think it might be helpful to capture here why we're switching to storing a descriptor in this JsonProperty which claims to be a branding project name. The current comment explains why we don't change the name, but not why we now want to store something different in it. E.g., our new system allows analytics to have more precision than branding project name would give us, since many distinct subscriptions may share branding settings.
</blockquote></details>

I've made an attempt...
___
*[`src/BloomExe/Publish/BloomLibrary/BloomLibraryPublishModel.cs` line 844 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6947#-OMrVe5vAGt_DYrUQvCj:-ON0hjIv-MwfFBHWJO0S:bebw9en) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/bc9f2ba16b95a0dde53e763995924d3f1fdbe7b1/src/BloomExe/Publish/BloomLibrary/BloomLibraryPublishModel.cs#L844)):*

Should we rename newBranding and oldBranding (both here and in the tsx)? Since a SubscriptionDescriptor now encodes a good deal more than a branding key.
Will it work for users to see "'The branding was "Local Community" but is now "ABC-LC". This may change logos and other material. Check this box if this is what you want.',?
To put this another way, do you think it's John's intent that what we internally consider a Descriptor (since it doesn't correspond exactly to a branding folder, though it determines one, along with other things) will continue to be referred to as a branding?
How about, "'The branding was "Local Community" but is now "joe_smith@gmail.com-Pro". This may change logos and other material. Check this box if this is what you want.'?
I suppose we can always improve that later as necessary.
</blockquote></details>

Thanks.
Yeah, John himself admits that understanding where to use subscription vs branding is not always clear.
I think we should go with this for now, and John can tweak things further in the future if needed.

andrew-polk (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6947)* status: 1 of 9 files reviewed, 1 unresolved discussion (waiting on @JohnThomson)

___
*[`src/BloomExe/Publish/BloomLibrary/BloomLibraryPublishModel.cs` line 844 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6947#-OMrVe5vAGt_DYrUQvCj:-ON0iswI4Mh6RPeFGRCt:byyzqqy) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/bc9f2ba16b95a0dde53e763995924d3f1fdbe7b1/src/BloomExe/Publish/BloomLibrary/BloomLibraryPublishModel.cs#L844)):*

Thanks.
Yeah, John himself admits that understanding where to use subscription vs branding is not always clear.
I think we should go with this for now, and John can tweak things further in the future if needed.
</blockquote></details>

It looks like there is only one remaining place "branding" is used in the UI.
That is in the book settings when something is locked due to the branding being used. (I'm not sure if that should still be branding or not, but it does feel more closely tied to branding than subscription...)
So it seems that generally, the idea was to get away from "branding" in the UI wherever possible.

andrew-polk (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6947)* status: 1 of 11 files reviewed, 1 unresolved discussion (waiting on @JohnThomson)

___
*[`src/BloomExe/Collection/CollectionSettings.cs` line 503 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6947#-ON171A26QGthTpgo3W_:-ON171A26QGthTpgo3Wa:b-9akuxy) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/761a8e38fdf62d9ef3da2659ec43d1125eb20c0e/src/BloomExe/Collection/CollectionSettings.cs#L503)):*
> ```Smalltalk
>                     // In other words, when running harvester, subscription code will be something like SIL-LEAD-***-***,
>                     // but we still want to process the book using the original branding.
>                     || Program.RunningHarvesterMode;
> ```

There is a good chance that I could have fixed the original harvester issue with just this one change.
But I think the other changes I made were still an improvement/clarification.

But just pointing out we could actually go back to having brandingProjectName on blorg and analytics be BrandingKey instead of SubscriptionDescriptor if we decided that was more betterer.

andrew-polk:
*[`src/BloomExe/Collection/CollectionSettings.cs` line 503 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6947#-ON171A26QGthTpgo3W_:-ON17XWI0dYkLHIvovlz:bpzmfb7) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/761a8e38fdf62d9ef3da2659ec43d1125eb20c0e/src/BloomExe/Collection/CollectionSettings.cs#L503)):*

There is a good chance that I could have fixed the original harvester issue with just this one change.
But I think the other changes I made were still an improvement/clarification.

But just pointing out we could actually go back to having brandingProjectName on blorg and analytics be BrandingKey instead of SubscriptionDescriptor if we decided that was more betterer.
</blockquote></details>

But once I made the other changes which ensures that Default is used for expired codes, this became a necessary change.

JohnThomson (changes requested):
> Reviewed 10 of 10 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6947)* status: all files reviewed, 1 unresolved discussion (waiting on @andrew-polk)

___
*[`src/BloomExe/Publish/BloomLibrary/BloomLibraryPublishModel.cs` line 827 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6947#-ON1915x5-J3V06kTsZt:-ON1915x5-J3V06kTsZu:b-y5n07) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/761a8e38fdf62d9ef3da2659ec43d1125eb20c0e/src/BloomExe/Publish/BloomLibrary/BloomLibraryPublishModel.cs#L827)):*
> ```Smalltalk
>             var existingThumbUrl = GetBloomLibraryThumbnailUrl(existingBookInfo);
> 
>             var oldSubscriptionDescriptor = existingBookInfo.brandingProjectName?.ToString();
> ```

This time round, I found it surprising that we're using the brandinProjectName property of a bookInfo, since that got renamed. But I'm guessing this is a raw JSON representation of a bookInfo as sent to Blorg. I think an explanatory comment would be good.

andrew-polk:
*[`src/BloomExe/Publish/BloomLibrary/BloomLibraryPublishModel.cs` line 827 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6947#-ON1915x5-J3V06kTsZt:-ON1Pf8E8Mv_oKqYoU7S:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/761a8e38fdf62d9ef3da2659ec43d1125eb20c0e/src/BloomExe/Publish/BloomLibrary/BloomLibraryPublishModel.cs#L827)):*

This time round, I found it surprising that we're using the brandinProjectName property of a bookInfo, since that got renamed. But I'm guessing this is a raw JSON representation of a bookInfo as sent to Blorg. I think an explanatory comment would be good.
</blockquote></details>

Done.

PR #6950, opened 2025-04-02
Allow "Local Community" subscription books to be edited (BL-14513)

StephenMcConnel:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6950)

### DIFF
--- a/src/BloomExe/Book/BookData.cs
+++ b/src/BloomExe/Book/BookData.cs
@@ -2328,28 +2328,22 @@ namespace Bloom.Book
             }
         }
 
-        /// <summary>
-        /// If we have "personalization" in the subscription and the HTML from the DOM,
-        //  fill in the personalization.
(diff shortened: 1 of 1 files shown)


hatton:
This is a fine fix but the thrust of my rewrite was to keep all such logic in Subscription.cs, and nice and unit-tested from there. Can you see if instead you can cause Subscription.cs to emit "" as the personalization for that subscription? I can do it if isn't clear.

StephenMcConnel (commented):
I think the personalization is already set to "", but the test immediately below to throw the exception is "IsNullOrWhitespace".  Is that test too stringent?

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6950#-:-OMvzb6-D1EgumtAey0o:bbegcmr)* status: 0 of 1 files reviewed, all discussions resolved

PR #6951, opened 2025-04-03
Don't limit bloom-show-en-when-blank to drag-activity-start (BL-14489)

JohnThomson:
It could be useful elsewhere, and we get flicker because some code that depends on it runs before drag-activity-start gets set

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6951)


JohnThomson:
Found this was incomplete.

PR #6952, opened 2025-04-03
Fix overflow sizing glitches while loading page (BL-14489)

JohnThomson:
Don't limit bloom-show-en-when-blank to drag-activity-start. It could be useful elsewhere, and we get flicker because some code that depends on it runs before drag-activity-start gets set.
But there are other sources of flicker from checking overflow too soon. And we had two separate sources initiating overflow checks and two different ways of delaying the checks and making them incremental. So I removed the old one that was happening too soon.

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6952)


andrew-polk (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6952)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @JohnThomson)

PR #6964, opened 2025-04-07
BL-14507: fix TBT wrongly opening image descriptions

nabalone:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6964)

### DIFF
--- a/src/BloomBrowserUI/bookEdit/toolbox/talkingBook/talkingBook.ts
+++ b/src/BloomBrowserUI/bookEdit/toolbox/talkingBook/talkingBook.ts
@@ -1,11 +1,9 @@
 import { kBloomCanvasClass } from "../../js/bloomImages";
-import {
-    hideImageDescriptions,
-    showImageDescriptions
-} from "../imageDescription/imageDescriptionUtils";
+import { hideImageDescriptions } from "../imageDescription/imageDescriptionUtils";
 import { beginLoadSynphonySettings } from "../readers/readerTools";
 import { getTheOneToolbox, ITool } from "../toolbox";
 import { ToolBox } from "../toolbox";
+import { getAudioRecorder } from "./audioRecording";
 import * as AudioRecorder from "./audioRecording";
 
 export default class TalkingBookTool implements ITool {
@@ -56,15 +54,6 @@ export default class TalkingBookTool implements ITool {
     // Called when a new page is loaded.
     public async newPageReady(): Promise<void> {
         this.showImageDescriptionsIfAny();
-        const imageDescToolActive = this.isImageDescriptionToolActive();
-        const checkbox = document.getElementById(
-            "audio-showImageDescription-wrapper"
-        );
-        if (checkbox) {
-            checkbox.style.display = imageDescToolActive
-                ? "inline-block"
-                : "none";
-        }
         const pageReadyPromise = AudioRecorder.theOneAudioRecorder.newPageReady(
             TalkingBookTool.deshroudPhraseDelimiters
         );
@@ -153,7 +142,6 @@ export default class TalkingBookTool implements ITool {
 
     // Called whenever the user edits text.
     public async updateMarkupAsync(): Promise<() => void> {
-        this.showImageDescriptionsIfAny();
         return AudioRecorder.theOneAudioRecorder.getUpdateMarkupAction();
     }
 
@@ -177,6 +165,7 @@ export default class TalkingBookTool implements ITool {
             return;
         }
         if (!this.isImageDescriptionToolActive()) {
+            getAudioRecorder()?.setShowingImageDescriptions(false);
             return;
         }
         const bloomCanvases = page.getElementsByClassName(kBloomCanvasClass);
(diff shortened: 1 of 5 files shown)


src/BloomBrowserUI/bookEdit/toolbox/talkingBook/talkingBook.ts
```diff
@@ -175,7 +175,7 @@ export default class TalkingBookTool implements ITool {
         if (!page) {
             return;
         }
-        if (!this.isImageDescriptionToolActive()) {
+        if (!AudioRecorder.theOneAudioRecorder.showingImageDescriptions) {
```
nabalone, line 178: This at least brings the image descriptions display into sync with the slider which is supposed to control them, and stops them from spontaneously turning themselves on and off

JohnThomson (changes requested):
This basically looks good. Only one place I think the code is probably wrong. Other comments are just wanting to be sure you tried various things; if you already did, you can just say so!

If you didn't already, do some testing on a page with both image descriptions and text canvas elements. Make sure the talking book tool sequences through things in some reasonable way both when the image descriptions are shown and when they are not. Testers should be encouraged to beat on this (new) situation, too. Watch especially for text that Next/Prev never takes you to, so you can't record it, and for cases where you have to click the button twice (probably because it advanced to something that's not visible).

> Reviewed 1 of 2 files at r1, 5 of 5 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6964#-:-ONL8yx16S15SUcAHiAr:b-majsy0)* status: all files reviewed, 2 unresolved discussions (waiting on @nabalone)

___
*[`src/BloomBrowserUI/bookEdit/toolbox/talkingBook/talkingBook.ts` line 156 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6964#-ONL85iy0k6wo4WksRIZ:-ONL85iy0k6wo4WksRI_:b-p9phmg) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/64c733acc8f96da180a72d5feff5db66768c95b4/src/BloomBrowserUI/bookEdit/toolbox/talkingBook/talkingBook.ts#L156)):*
> ```XML
>     // Called whenever the user edits text.
>     public async updateMarkupAsync(): Promise<() => void> {
>         this.showImageDescriptionsIfAny();
> ```

I wish I had some clue why this was ever added. We certainly don't want it here...this shouldn't happen every keystroke...but it was probably intended to handle some path to showing the page and the talking book tool that doesn't involve newPageReady. I think we tested all the ways of getting to that state (opening the toolbox, switching tools) that don't involve switching pages, but it would be good to add a note to encourage testers to look for such a way to break things.
___
*[`src/BloomBrowserUI/bookEdit/toolbox/talkingBook/audioRecording.ts` line 682 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6964#-ONL6u5tDgAV416TzpRQ:-ONL6u5tDgAV416TzpRR:b-di2cgc) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/a15ef1e13fb5116ddb2706ca079612e52a767675/src/BloomBrowserUI/bookEdit/toolbox/talkingBook/audioRecording.ts#L682)):*
> ```XML
>                     return false;
>                 }
>             }
> ```

I would think we still need the second half of this if. When image descriptions are hidden, we don't want prev/next to try to highlight one of them. (Or maybe something else prevents that? Anyway, when turned off they are NOT visible, so if we want this to return true in that case it deserves a comment and maybe a better method name.

nabalone (commented):
I have tested with both image descriptions and text canvas elements and the next/prev buttons

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6964#-:-ONLMX-h3zJPj5lpVcSu:bggvt3t)* status: 4 of 5 files reviewed, 2 unresolved discussions (waiting on @JohnThomson)

___
*[`src/BloomBrowserUI/bookEdit/toolbox/talkingBook/audioRecording.ts` line 682 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6964#-ONL6u5tDgAV416TzpRQ:-ONLLKWh4JFh72-IOjyd:b-pvifdr) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/a15ef1e13fb5116ddb2706ca079612e52a767675/src/BloomBrowserUI/bookEdit/toolbox/talkingBook/audioRecording.ts#L682)):*

I would think we still need the second half of this if. When image descriptions are hidden, we don't want prev/next to try to highlight one of them. (Or maybe something else prevents that? Anyway, when turned off they are NOT visible, so if we want this to return true in that case it deserves a comment and maybe a better method name.
</blockquote></details>

When the image descriptions are hidden, we are detecting that with the `EditableDivUtils.isInHiddenLanguageBlock(elem)`check above, which finds that they have `display: none`. Do we still want an additional check, like this?

JohnThomson (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6964)* status: 4 of 5 files reviewed, 1 unresolved discussion (waiting on @nabalone)

PR #6965, opened 2025-04-07
Properly migrate Local-Community to Legacy-LC (BL-14513)

StephenMcConnel:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6965)

### DIFF
--- a/src/BloomExe/Book/BookData.cs
+++ b/src/BloomExe/Book/BookData.cs
@@ -2329,24 +2329,23 @@ namespace Bloom.Book
         }
 
 		/// <summary>
-		/// If we have "personalization" in the subscription and the HTML from the DOM,
-		/// fill in the personalization.
+		/// If we have "{personalization}" in the subscription's HTML template, fill in the
+		/// personalization from the subscription code.  If the personalization is empty,
+		/// throw an exception because it should always exist if it's in the template.
 		/// </summary>
-		/// <remarks>
-		/// Note that "Local-Community" sets personalization to "" which is okay for that situation.
-		/// So there is no point to a sanity check here requiring a personalization.  See BL-14513.
-		/// </remarks>
 		private string MergeInPersonalization(string content)
         {
             // if the CollectionSettings has a Subscription Personalization, replace any instances of
-            // {personalization} with the value
-            if (content.Contains("{personalization}") &&
-                !string.IsNullOrWhiteSpace(this.CollectionSettings.Subscription.Personalization))
+            // {personalization} with that value.
+            if (content.Contains("{personalization}"))
             {
-                return content.Replace(
-                    "{personalization}",
-                    this.CollectionSettings.Subscription.Personalization
-                );
+                var personalization = this.CollectionSettings.Subscription.Personalization;
+                if (string.IsNullOrEmpty(personalization))
+                {
+                    throw new ApplicationException(
+                        "Branding personalization is not set, but the branding template contains {personalization}.");
+                }
+                return content.Replace("{personalization}", personalization);
             }
 
             return content;
--- a/src/BloomExe/Collection/Subscription.cs
+++ b/src/BloomExe/Collection/Subscription.cs
@@ -37,33 +37,55 @@ public class Subscription
         Tier = CalculateTier();
     }
 
-    // Factory Method
-    // We would really like to say look it's only the code, from that we can get everything else.
-    // However there are two cases where that is not enough:
-    // 1. Legacy collections will have no code, but only a branding of "Local-Community".
-    // 2. We are editing a book that was uploaded to BloomLibrary.org, and we want to use the same branding as when it was uploaded but the code
-    // was intentionally blanked so as not to publish it.
+    /// <summary>
+    /// Factory Method
+    /// We would really like to say look it's only the code, from that we can get everything else.
+    /// However there are two cases where that is not enough:
+    /// 1. Local legacy collections have no code, but only a branding of "Local-Community".
+    /// 2. We are editing a book that was uploaded to BloomLibrary.org, and we want to use the same branding
+    ///    as when it was uploaded, but the code was intentionally blanked so as not to publish it.
+    /// </summary>
+    /// <param name="code">content of &lt;SubscriptionCode&gt; from file, or null if empty</param>
+    /// <param name="brandingProjectName">content of &lt;BrandingProjectName&gt; from file, or "Default" if empty</param>
+    /// <param name="editingABlorgBook">true iff we're editing a book downloaded specifically for edit/update</param>
+    /// <returns>a new Subscription object based on the inputs</returns>
+    /// <remarks>
+    /// When downloading books for editing/updating, Bloom Library provides a code of "Local-Community-***-***"
+    /// for the branding of "Local-Community" and does not provide a branding.  (so it goes to "Default")
+    /// </remarks>
     public static Subscription FromCollectionSettingsInfo(
         string code,
-        string descriptor,
+        string brandingProjectName,
         bool editingABlorgBook = false
     )
     {
-        // If the collection has <SubscriptionCode>foobar-***-***</SubscriptionCode> but <BrandingName>Default</BrandingName>,
-        // then we have a conflict (which is why we're phasing out the use of <BrandingName>). In that case, go with what
-        // is given in the code by clearing the descriptor.
-        if (editingABlorgBook && descriptor == "Default" && code.Contains(kRedactedCodeSuffix))
+        // Collections created before Bloom 4.4 use the branding name "Local Community" instead of "Local-Community"
+        if (brandingProjectName == "Local Community")
+            brandingProjectName = "Local-Community";
+        string descriptor = null;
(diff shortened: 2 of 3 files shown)


andrew-polk (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6965)* status: all files reviewed, 1 unresolved discussion (waiting on @StephenMcConnel)

___
*[`src/BloomExe/Collection/Subscription.cs` line 43 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6965#-ONHAypaEFWtyOJ33OEp:-ONHAypbBYM3_KuQe1f1:bwof1af) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/acf23bab267abe4465384a316dc564d1a6a2c5ee/src/BloomExe/Collection/Subscription.cs#L43)):*
> ```Smalltalk
>     // We would really like to say look it's only the code, from that we can get everything else.
>     // However there are two cases where that is not enough:
>     // 1. Legacy collections will have no code, but only a branding of "Local-Community".
> ```

This comment needs updating.

JohnThomson (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6965)* status: all files reviewed, 2 unresolved discussions (waiting on @StephenMcConnel)

___
*[`src/BloomExe/Collection/Subscription.cs` line 62 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6965#-ONHBaFGAxNqlx0sQzxb:-ONHBaFH9PmmAvVxvY_T:b-8p2xb3) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/acf23bab267abe4465384a316dc564d1a6a2c5ee/src/BloomExe/Collection/Subscription.cs#L62)):*
> ```Smalltalk
>         if (
>             (string.IsNullOrWhiteSpace(code) || code.StartsWith("Local-Community"))
>             && (descriptor == "Local-Community" || descriptor == "Local Community")
> ```

Is there some history to this complex conditional that is worth capturing in a comment? Did we release different versions of the legacy Local Community branding?

andrew-polk (changes requested):
> Reviewed 2 of 2 files at r2.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6965)* status: all files reviewed (commit messages unreviewed), 2 unresolved discussions (waiting on @JohnThomson)

___
*[`src/BloomExe/Collection/Subscription.cs` line 49 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6965#-ONLIzIo1hWWpb9fGbyo:-ONLIzIo1hWWpb9fGbyp:b-29l268) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/6702f1e68cef970b57741ef21302c56dc74be8b9/src/BloomExe/Collection/Subscription.cs#L49)):*
> descriptor

If this is always the legacy BrandingProjectName or "Default", I think we should change the name of the parameter.

How about legacyBrandingProjectName?

StephenMcConnel (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6965)* status: 1 of 3 files reviewed, 2 unresolved discussions (waiting on @andrew-polk and @JohnThomson)

___
*[`src/BloomExe/Collection/Subscription.cs` line 49 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6965#-ONLIzIo1hWWpb9fGbyo:-ONMZz22Bc8_kJB1gHrF:bb7h5dn) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/6702f1e68cef970b57741ef21302c56dc74be8b9/src/BloomExe/Collection/Subscription.cs#L49)):*

> descriptor

If this is always the legacy BrandingProjectName or "Default", I think we should change the name of the parameter.

How about legacyBrandingProjectName?
</blockquote></details>

I prefer just brandingProjectName.  Which I had instantiated earlier and then removed (before deciding that we don't need to set the Subscription.Descriptor field explicitly in this method).

andrew-polk (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6965)* status: all files reviewed, 2 unresolved discussions (waiting on @JohnThomson and @StephenMcConnel)

___
*[`src/BloomExe/Collection/Subscription.cs` line 49 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6965#-ONLIzIo1hWWpb9fGbyo:-ONM_jnHCpGiWavXx4xP:bf26595) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/6702f1e68cef970b57741ef21302c56dc74be8b9/src/BloomExe/Collection/Subscription.cs#L49)):*

I prefer just brandingProjectName.  Which I had instantiated earlier and then removed (before deciding that we don't need to set the Subscription.Descriptor field explicitly in this method).
</blockquote></details>

Ugh. Sorry. This subscription stuff is confusing...

I hadn't realized that you were using the same `descriptor` later on in the method to write to.

Now some lines are definitely wrong. For example, setting `brandingProjectName = "Legacy-LC";`. There it needs to be descriptor.

I found when I was trying to think through all this with my own changes that it was best to think of branding only as useful in very limited terms: 
- analytics where it could probably be relabeled as subscriptionDescriptor if not for historical reasons
- the folder name where we load branding files from
- in this case, we are also using it in the case of older books which only had this set and not the subscription code. That's why I was wondering if legacyBrandingProjectName or something which indicates that is only for older books might be helpful.

Maybe we should try to work through this together, but here's a quick analysis which might be wrong...
Should we leave the input to the method as brandingProjectName but only use that for reading? Then create a new variable called descriptor which we set and use through most of the method?

andrew-polk (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6965)* status: all files reviewed, 1 unresolved discussion (waiting on @JohnThomson)

___
*[`src/BloomExe/Collection/Subscription.cs` line 49 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6965#-ONLIzIo1hWWpb9fGbyo:-ONMmyT_0AmDxsrDK48T:b-ucfkn4) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/6702f1e68cef970b57741ef21302c56dc74be8b9/src/BloomExe/Collection/Subscription.cs#L49)):*

I took your final suggestion essentially, then made one more improvement with regard to setting the expiration date for editing a book downloaded from blorg.  I had previously almost done what your suggested, then changed back to a single (originally input) variable.  I'm glad I'm not the only one who has found this method to be  somewhat confusing.  I hope I've cleared up most of the confusion by now.  Since branding names are still being written to the collection settings file, I'm not sure JH has totally decided on whether they serve any purpose or not.  And although I'm not afraid of long variable names,  I prefer not to get beyond 20 or so characters.  :-)
</blockquote></details>

Thanks.

JohnThomson (approved):
LGTM, with two minor take-or-leave suggestions.

> Reviewed 1 of 3 files at r1, 1 of 2 files at r3, 1 of 1 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6965#-:-ONQdRpICJuo6byLCUQE:bm3djg8)* status: all files reviewed, 2 unresolved discussions (waiting on @StephenMcConnel)

___
*[`src/BloomExe/Collection/Subscription.cs` line 41 at r4](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6965#-ONQc7SF35IlaXFMj3Ii:-ONQc7SF35IlaXFMj3Ij:bnypzrj) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/8db9dcfe82e48d346750d616f14d63718f886664/src/BloomExe/Collection/Subscription.cs#L41)):*
> ```Smalltalk
> 
>     /// <summary>
>     // Factory Method
> ```

If you do another version, may as well make this a /// also.
___
*[`src/BloomExe/Collection/Subscription.cs` line 81 at r4](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6965#-ONQcHD86wkHvZ-zp3CF:-ONQcHD86wkHvZ-zp3CG:b-mrgpkd) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/8db9dcfe82e48d346750d616f14d63718f886664/src/BloomExe/Collection/Subscription.cs#L81)):*
> ```Smalltalk
>             }
>         }
>         // Collections created before Bloom 4.4 have the branding "Local Community" instead of "Local-Community"
> ```

There's at least one other place where we check for Local-Community but do NOT check for "Local Community". I convinced myself that it couldn't be a problem, but I think it would be easier to reason about if, right at the start of the method, you just put 

```
if (brandingProjectName == "Local Community")
    brandingProjectName = "Local-Community"
```

 (with the comment above), and then you don't have to worry about it anywhere else.

andrew-polk (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6965)* status: all files reviewed, 2 unresolved discussions (waiting on @JohnThomson)

PR #6976, opened 2025-04-08
Hide unwanted items in Game editing tabs (BL-14532)

StephenMcConnel:
Also try to prevent other items from being deleted or moved except in an appropriate tab.

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6976)

### DIFF
--- a/src/BloomBrowserUI/bookEdit/toolbox/games/DragActivityTabControl.tsx
+++ b/src/BloomBrowserUI/bookEdit/toolbox/games/DragActivityTabControl.tsx
@@ -12,6 +12,7 @@ import { useL10n } from "../../../react_components/l10nHooks";
 import { default as PencilIcon } from "@mui/icons-material/Edit";
 import { showGamePromptDialog } from "../games/GameTool";
 import BloomButton from "../../../react_components/bloomButton";
+import { getCanvasElementManager } from "../overlay/canvasElementUtils";
 
 // This component is responsible for the Game Setup mode tabs in the Game tool.
 // Although the code seems to belong in this folder with the other Game code, it is actually
@@ -24,6 +25,7 @@ export const DragActivityTabControl: React.FunctionComponent<{
 }> = props => {
     const changeHandler = (tab: number) => {
         getToolboxBundleExports()?.setActiveDragActivityTab(tab);
+        getCanvasElementManager()?.setActiveElement(undefined);
     };
     const setupMode = useL10n(
         "Game Setup mode:",
--- a/src/content/templates/template books/Games/Games.less
+++ b/src/content/templates/template books/Games/Games.less
@@ -36,6 +36,41 @@
 .drag-activity-wrong .drag-item-wrong {
     display: block;
 }
+// Hide draggable items in the Correct and Wrong tabs. See BL-14532.
+// The data-target-of elements will show the positions of the targets and
+// indicate the correct answer in the Correct tab.
+.drag-activity-correct,
+.drag-activity-wrong {
+    [data-draggable-id] {
+        display: none;
+    }
+}
+// The drag-activity-play and drag-activity-correct are both present when the user has
+// given a correct answer and clicked for it to be checked.  (The same goes for a wrong
+// answer and drag-activity-wrong.)  Show the user's answer, right or wrong,
+// via the [data-draggable-id] elements which have been dragged to the targets.
+// Unmoved [data-draggable-id] elements are hidden in showing the right answer.
+.drag-activity-play.drag-activity-correct,
+.drag-activity-play.drag-activity-wrong {
+    .bloom-canvas-element.bloom-draggedToTarget[data-draggable-id] {
+        display: block;
+    }
+}
+// Prevent unwanted interaction with the background images and other items that are not
+// supposed to be changed in the Correct and Wrong tabs.  See BL-14532.
+// Less doesn't appear to handle the :is() operator with a list of selectors, so we have
+// to use two rules to get the result we want.
+.drag-activity-correct .bloom-canvas-element:not(.drag-item-correct),
+.drag-activity-wrong .bloom-canvas-element:not(.drag-item-wrong) {
+    pointer-events: none;
+}
+.drag-activity-correct,
+.drag-activity-wrong {
+    [data-target-of] {
+        pointer-events: none;
+    }
+}
+
 // in play mode suppress these focus effects
 .drag-activity-play {
     .bloom-canvas .bloom-canvas-element div:focus,
@@ -212,11 +247,17 @@
     .drag-activity-start
         .bloom-page:not([data-activity="drag-image-to-target"])
         &,
-    div:not(.drag-activity-start) & {
+    .drag-activity-play & {
         > * {
             // targets have children to show a shadow of what is supposed to go there.
             // the contents are identical, this makes them a shadow.
             opacity: 50%;
+        }
+    }
+    .drag-activity-wrong & {
+        // In the wrong tab, we don't want to see the target at all, but just the dashed box if it's visible.
+        > * {
+            opacity: 0%;
         }
     }
 
@@ -350,11 +391,20 @@
         background-color: transparent;
     }
 }
-// And don't show them at all in play mode.
-.drag-activity-play .bloom-page[data-show-targets-during-play="false"] {
-    [data-target-of] {
-        // display:none would prevent the snap code from using them, since they would not have a location.
+// And don't show them at all in play mode or the Wrong tab.
+.drag-activity-play,
+.drag-activity-wrong {
+    .bloom-page[data-show-targets-during-play="false"] [data-target-of] {
+        // display:none would prevent the snap code from using them in play mode, since they would not have a location.
         visibility: hidden;
+    }
+}
+// If we don't show targets during play, we don't want to show the dashed box in the Correct tab.
+.drag-activity-correct {
+    .bloom-page[data-show-targets-during-play="false"] [data-target-of] {
+        border: none;
+        outline: none;
+        box-shadow: none;
     }
 }
 


andrew-polk (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6976)* status: all files reviewed, 2 unresolved discussions (waiting on @StephenMcConnel)

___
*[`src/content/templates/template books/Games/Games.less` line 38 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6976#-ONMhxoYByinb4SRMpLG:-ONMhxoYByinb4SRMpLH:bh28jbr) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/d96e073609f988f2190d8c211bcf6dfa1cc06a31/src/content/templates/template%20books/Games/Games.less#L38)):*
> ```Less
> .drag-activity-wrong .drag-item-wrong {
>     display: block;
> }
> ```

I think I better ask @JohnThomson to review this as I'm feeling quite out of my depth here.

If I'm not mistaken, this file should only include things which need to travel with the book to bloom-player.
Anything which is specific to edit mode should be in editMode.less?

I'm not sure which of these rules fit into which category. But I think at least the context controls one?
___
*[`src/content/templates/template books/Games/Games.less` line 57 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6976#-ONMh2Is6CfRJEFjX7Dz:-ONMh2It9iiNZ6XSMGbg:b-lu68kd) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/d96e073609f988f2190d8c211bcf6dfa1cc06a31/src/content/templates/template%20books/Games/Games.less#L57)):*
> ```Less
> }
> // The following rule has to be extremely specific to override the previous rule when in the Play tab.
> div.drag-activity-play.drag-activity-correct
> ```

I didn't think we ever had drag-activity-play and drag-activity-correct at the same time.
Is this is special case?
Maybe worth a more detailed comment?

JohnThomson (changes requested):
I don't see anything that will cause the draggable correct items to be in their correct places when in the correct tab. (Instead you have hidden them.)

It will look right in the drag images game, since the draggables are now indistinguishable from the copies in the targets. Maybe that's OK, but for the moment at least, the target copies have a different appearance in the drag-letters game. We could fix that with CSS (override the opacity rule for targets in Correct). But then we need another set of overrides to make the target copies visible in Play mode when a Check is correct (or turn the draggables back on, as you are partly doing)...otherwise your answer will usually disappear if you got it right. It feels messy and likely to get messed up as we make other changes.

I'm not sure, but it might be easier to use code and actually move them when entering the Correct tab. Basically just reuse the code that shows the correct answer, along with the code that restores positions after being in Play tab.

> Reviewed 1 of 2 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6976#-:-ONQKVxe2YezjuuId_wk:bhi3ium)* status: all files reviewed, 10 unresolved discussions (waiting on @andrew-polk)

___
*[`src/content/templates/template books/Games/Games.less` line 38 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6976#-ONMhxoYByinb4SRMpLG:-ONQNwJx1RqyF6LQAcIw:b6rqkzw) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/d96e073609f988f2190d8c211bcf6dfa1cc06a31/src/content/templates/template%20books/Games/Games.less#L38)):*

I think I better ask @JohnThomson to review this as I'm feeling quite out of my depth here.

If I'm not mistaken, this file should only include things which need to travel with the book to bloom-player.
Anything which is specific to edit mode should be in editMode.less?

I'm not sure which of these rules fit into which category. But I think at least the context controls one?
</blockquote></details>

It would make sense for anything that is only for the start, correct, or wrong tabs to be in editMode.less. OTOH, it's nice to have all the game-specific CSS together in one place, and it's not very expensive to carry a few extra rules along with the book. Moreover, drag-activity-wrong and drag-activity-correct are actually used at play time as well (to show the right and wrong elements after checking), so distinguishing which of those rules are only applicable to their own tabs would be challenging. I don't think there's currently anything specific to these games in editMode.less, and I lean towards keeping it that way.
___
*[`src/content/templates/template books/Games/Games.less` line 57 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6976#-ONMh2Is6CfRJEFjX7Dz:-ONPqD7PBDTeLvgiXCH3:b7kb6ek) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/d96e073609f988f2190d8c211bcf6dfa1cc06a31/src/content/templates/template%20books/Games/Games.less#L57)):*

I didn't think we ever had drag-activity-play and drag-activity-correct at the same time.
Is this is special case?
Maybe worth a more detailed comment?
</blockquote></details>

Currently we turn both on when the user clicks "check" and the answer is correct. The idea was to turn on the same things we turn on in the correct tab. 
___
*[`src/content/templates/template books/Games/Games.less` line 39 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6976#-ONQFePV12vX9OW2_64J:-ONQFePV12vX9OW2_64K:bmle1j5) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/d96e073609f988f2190d8c211bcf6dfa1cc06a31/src/content/templates/template%20books/Games/Games.less#L39)):*
> ```Less
>     display: block;
> }
> // Hide unwanted clutter in the Correct and Wrong tabs. See BL-14532.
> ```

These rules are complex enough that I think we should have a comment on pretty much each one that says more about why we want that behavior in that situation.
___
*[`src/content/templates/template books/Games/Games.less` line 43 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6976#-ONQDQ7C0f25sMJ6X7nD:-ONQDQ7C0f25sMJ6X7nE:b6dohe4) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/d96e073609f988f2190d8c211bcf6dfa1cc06a31/src/content/templates/template%20books/Games/Games.less#L43)):*
> ```Less
> .drag-activity-wrong {
>     [data-draggable-id],
>     ~ #canvas-element-context-controls {
> ```

This 'following sibling' thing strikes me as fragile. It depends on something that's more-or-less an accident of the way the canvas element controls was implemented. (For example, we may decide we want these controls above the full bleed shadows and promote them to a higher level in the DOM.) I'm also nervous about having the element still actually active while just making it look as though it's not. (Have you hidden the toolbar, too? It's not  a child of context controls.)
I would be inclined instead to make every tab switch call setActiveElement(undefined). That way nothing has the controls when you go to a different tab, and there really is no active canvas element. Then if things that shouldn't become active have pointer-events:none, they can't ever become active. I don't think there's ever a case where it would be useful to keep the same canvas element active while switching tabs.
To really guarantee that you can't edit something that belongs to a different tab, perhaps we should also blur anything that has focus when switching tabs. Otherwise for example you could be editing the instructions, switch to Correct, and go on typing there.
___
*[`src/content/templates/template books/Games/Games.less` line 46 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6976#-ONQG4vJ7gp4yrqbJmrO:-ONQG4vJ7gp4yrqbJmrP:bwne8t3) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/d96e073609f988f2190d8c211bcf6dfa1cc06a31/src/content/templates/template%20books/Games/Games.less#L46)):*
> ```Less
>         display: none;
>     }
> }
> ```

This rule will hide ALL the draggables when in the correct tab (and when the user clicks Check and got it right). John asked us to hide all of them in the wrong tab, but I think only the distractors in the correct tab. (I've also asked him to clarify whether they should all disappear after Checking a wrong answer. I think that's a bad idea.)
It's possible that you're simulating "show them in the correct position" by hiding them and just leaving the copies in the targets visible. But those will be hidden in Play mode. See discussion at the end.
So I think you might want .drag-activity-correct [data-no-target], .drag-activity-wrong [data-draggable-id]. (Only if you go with my suggestion of really moving the correct ones, though.)
___
*[`src/content/templates/template books/Games/Games.less` line 55 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6976#-ONQGnYk4zVwISOOI_dI:-ONQGnYk4zVwISOOI_dJ:b8izxvb) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/d96e073609f988f2190d8c211bcf6dfa1cc06a31/src/content/templates/template%20books/Games/Games.less#L55)):*
> ```Less
>         display: none;
>     }
> }
> ```

I'm not sure what this is getting at. You already hid the draggables. 
So I think the effect of this will be to hide images, videos, and gifs that belong to the start tab and are part of the scene (not draggable in Play). I don't think John asked for that.
___
*[`src/content/templates/template books/Games/Games.less` line 61 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6976#-ONQQdAe7DF8wgCX14Nh:-ONQQdAe7DF8wgCX14Ni:b-1jvzat) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/d96e073609f988f2190d8c211bcf6dfa1cc06a31/src/content/templates/template%20books/Games/Games.less#L61)):*
> ```Less
>     div.bloom-canvas-element[data-draggable-id]:not([data-no-target]):has(img[src]) {
>     display: block;
> }
> ```

I think this is to turn them back on in Play mode, where you usually don't get the target copies. If you keep that approach, you probably also need one to turn draggable letters back on...
___
*[`src/content/templates/template books/Games/Games.less` line 66 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6976#-ONQIbDL1OZmfZjUdOXN:-ONQIbDL1OZmfZjUdOXO:b-97a18i) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/d96e073609f988f2190d8c211bcf6dfa1cc06a31/src/content/templates/template%20books/Games/Games.less#L66)):*
> ```Less
> .drag-activity-correct [data-target-of] .bloom-imageContainer {
>     pointer-events: none;
> }
> ```

This would still allow you to interact with, say, a start-mode video. I think it wants to be more general. How about doing this for .drag-activity-correct .bloom-canvas-element:not(.drag-item-correct), .drag-activity-wrong .bloom-canvas-element:not(.drag-item-wrong)? That is, any item that is not specific to the tab you're in will be inert. (That would also replace the following rule.)
(May need an exception to allow the Retry and Show Correct buttons to work after checking a wrong answer, but I think they are not canvas elements.)
(If you do this, make sure it doesn't suppress any proper interactions with start-mode things in Play, like replaying a sound or video.)
___
*[`src/content/templates/template books/Games/Games.less` line 71 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6976#-ONQKFb868_jGQ_c-Pbc:-ONQKFb868_jGQ_c-Pbd:b-k3x3uj) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/d96e073609f988f2190d8c211bcf6dfa1cc06a31/src/content/templates/template%20books/Games/Games.less#L71)):*
> ```Less
>     .bloom-canvas-element.bloom-backgroundImage,
>     .bloom-canvas-element:has(.bloom-translationGroup) {
>         pointer-events: none;
> ```

This will prevent interacting with a text block that is e.g. correct-mode-only. See the suggestion above.
___
*[`src/BloomBrowserUI/bookEdit/js/CanvasElementContextControls.tsx` line 843 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6976#-ONQE7kU3xoS8qnhX9aN:-ONQE7kU3xoS8qnhX9aO:byj2vjw) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/d96e073609f988f2190d8c211bcf6dfa1cc06a31/src/BloomBrowserUI/bookEdit/js/CanvasElementContextControls.tsx#L843)):*
> ```XML
>             currentDraggableTarget.remove();
>             setCurrentDraggableTarget(undefined);
>             canvasElement.setAttribute("data-no-target", "true");
> ```

I think I would be more inclined to use a class here. I suppose I've cheated a bit using CSS that looks for [data-draggable-id] and [data-target-of], so maybe it's natural to have parallel rules that use [data-no-target], but the other two attributes have actually necessary values, and adding a class as well would be redundant. This thing is only there to affect CSS, and that's what classes were made for.

StephenMcConnel (commented):
I've added images to the issue to give you an idea what shows on the screen with this PR.  The current set of rules is smaller and simpler than what was in the original PR.

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6976#-:-ONRbob6BVrX5j2Nb3yA:bkwfwi2)* status: 1 of 3 files reviewed, 10 unresolved discussions (waiting on @andrew-polk and @JohnThomson)

___
*[`src/BloomBrowserUI/bookEdit/js/CanvasElementContextControls.tsx` line 843 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6976#-ONQE7kU3xoS8qnhX9aN:-ONQkoE6E-gbcBoe2-To:bz2160p) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/d96e073609f988f2190d8c211bcf6dfa1cc06a31/src/BloomBrowserUI/bookEdit/js/CanvasElementContextControls.tsx#L843)):*

I think I would be more inclined to use a class here. I suppose I've cheated a bit using CSS that looks for [data-draggable-id] and [data-target-of], so maybe it's natural to have parallel rules that use [data-no-target], but the other two attributes have actually necessary values, and adding a class as well would be redundant. This thing is only there to affect CSS, and that's what classes were made for.
</blockquote></details>

Done.  I chose the class name "bloom-noDraggableTarget".
___
*[`src/content/templates/template books/Games/Games.less` line 38 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6976#-ONMhxoYByinb4SRMpLG:-ONQmtN72XcOWfJ4Jb9U:bp15sm4) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/d96e073609f988f2190d8c211bcf6dfa1cc06a31/src/content/templates/template%20books/Games/Games.less#L38)):*

It would make sense for anything that is only for the start, correct, or wrong tabs to be in editMode.less. OTOH, it's nice to have all the game-specific CSS together in one place, and it's not very expensive to carry a few extra rules along with the book. Moreover, drag-activity-wrong and drag-activity-correct are actually used at play time as well (to show the right and wrong elements after checking), so distinguishing which of those rules are only applicable to their own tabs would be challenging. I don't think there's currently anything specific to these games in editMode.less, and I lean towards keeping it that way.
</blockquote></details>

I'll change it back from the second iteration of the PR, which moved a couple of rules.
___
*[`src/content/templates/template books/Games/Games.less` line 39 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6976#-ONQFePV12vX9OW2_64J:-ONQz02e3dcGWD382h6u:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/d96e073609f988f2190d8c211bcf6dfa1cc06a31/src/content/templates/template%20books/Games/Games.less#L39)):*

These rules are complex enough that I think we should have a comment on pretty much each one that says more about why we want that behavior in that situation.
</blockquote></details>

Done.
___
*[`src/content/templates/template books/Games/Games.less` line 43 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6976#-ONQDQ7C0f25sMJ6X7nD:-ONR3Asy15znuN3TL-xD:bt31v9h) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/d96e073609f988f2190d8c211bcf6dfa1cc06a31/src/content/templates/template%20books/Games/Games.less#L43)):*

This 'following sibling' thing strikes me as fragile. It depends on something that's more-or-less an accident of the way the canvas element controls was implemented. (For example, we may decide we want these controls above the full bleed shadows and promote them to a higher level in the DOM.) I'm also nervous about having the element still actually active while just making it look as though it's not. (Have you hidden the toolbar, too? It's not  a child of context controls.)
I would be inclined instead to make every tab switch call setActiveElement(undefined). That way nothing has the controls when you go to a different tab, and there really is no active canvas element. Then if things that shouldn't become active have pointer-events:none, they can't ever become active. I don't think there's ever a case where it would be useful to keep the same canvas element active while switching tabs.
To really guarantee that you can't edit something that belongs to a different tab, perhaps we should also blur anything that has focus when switching tabs. Otherwise for example you could be editing the instructions, switch to Correct, and go on typing there.
</blockquote></details>

I've added the call to setActiveElement(undefined).  I think that automatically blurs the element.
___
*[`src/content/templates/template books/Games/Games.less` line 57 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6976#-ONMh2Is6CfRJEFjX7Dz:-ONQoc3nA-mP4YkI7onI:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/d96e073609f988f2190d8c211bcf6dfa1cc06a31/src/content/templates/template%20books/Games/Games.less#L57)):*

Currently we turn both on when the user clicks "check" and the answer is correct. The idea was to turn on the same things we turn on in the correct tab.
</blockquote></details>

Done.
___
*[`src/content/templates/template books/Games/Games.less` line 61 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6976#-ONQQdAe7DF8wgCX14Nh:-ONR56gwEtRwOR9QjiEA:b49z7hi) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/d96e073609f988f2190d8c211bcf6dfa1cc06a31/src/content/templates/template%20books/Games/Games.less#L61)):*

I think this is to turn them back on in Play mode, where you usually don't get the target copies. If you keep that approach, you probably also need one to turn draggable letters back on...
</blockquote></details>

I've changed this rule, and removed any reference to img.
___
*[`src/content/templates/template books/Games/Games.less` line 66 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6976#-ONQIbDL1OZmfZjUdOXN:-ONR8egk1BHcr0XfZ6Ek:bs4jrs4) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/d96e073609f988f2190d8c211bcf6dfa1cc06a31/src/content/templates/template%20books/Games/Games.less#L66)):*

This would still allow you to interact with, say, a start-mode video. I think it wants to be more general. How about doing this for .drag-activity-correct .bloom-canvas-element:not(.drag-item-correct), .drag-activity-wrong .bloom-canvas-element:not(.drag-item-wrong)? That is, any item that is not specific to the tab you're in will be inert. (That would also replace the following rule.)
(May need an exception to allow the Retry and Show Correct buttons to work after checking a wrong answer, but I think they are not canvas elements.)
(If you do this, make sure it doesn't suppress any proper interactions with start-mode things in Play, like replaying a sound or video.)
</blockquote></details>

I used your suggestion, but also added a similar rule for [data-target-of] in those tabs/modes.
___
*[`src/content/templates/template books/Games/Games.less` line 71 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6976#-ONQKFb868_jGQ_c-Pbc:-ONR8eFa0Qd1qT-BLroq:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/d96e073609f988f2190d8c211bcf6dfa1cc06a31/src/content/templates/template%20books/Games/Games.less#L71)):*

This will prevent interacting with a text block that is e.g. correct-mode-only. See the suggestion above.
</blockquote></details>

Done.

andrew-polk (commented):
> Reviewed 3 of 4 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6976)* status: 3 of 4 files reviewed, 8 unresolved discussions (waiting on @JohnThomson and @StephenMcConnel)

StephenMcConnel (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6976)* status: 3 of 4 files reviewed, 8 unresolved discussions (waiting on @andrew-polk and @JohnThomson)

___
*[`src/BloomBrowserUI/bookEdit/js/CanvasElementContextControls.tsx` line 843 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6976#-ONQE7kU3xoS8qnhX9aN:-ONWhTCl9-8CM5k9hRVv:bqb1z9m) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/d96e073609f988f2190d8c211bcf6dfa1cc06a31/src/BloomBrowserUI/bookEdit/js/CanvasElementContextControls.tsx#L843)):*

Done.  I chose the class name "bloom-noDraggableTarget".
</blockquote></details>

I've changed this to use the class added by bloom-player code which was needed to hide unused draggables when the answer is wrong.

StephenMcConnel (commented):
I've edited this file so much that I'm not sure Reviewable can do a reasonable job of presenting revisions.  If you can review the final change against what is in master, that would probably be best.

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6976#-:-ONWnM4Y3FFQk3BFbZBo:bkixgjm)* status: 2 of 4 files reviewed, 8 unresolved discussions (waiting on @andrew-polk and @JohnThomson)

___
*[`src/content/templates/template books/Games/Games.less` line 46 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6976#-ONQG4vJ7gp4yrqbJmrO:-ONWn0CI2mvj5TN7iZH2:b-jd8t19) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/d96e073609f988f2190d8c211bcf6dfa1cc06a31/src/content/templates/template%20books/Games/Games.less#L46)):*

This rule will hide ALL the draggables when in the correct tab (and when the user clicks Check and got it right). John asked us to hide all of them in the wrong tab, but I think only the distractors in the correct tab. (I've also asked him to clarify whether they should all disappear after Checking a wrong answer. I think that's a bad idea.)
It's possible that you're simulating "show them in the correct position" by hiding them and just leaving the copies in the targets visible. But those will be hidden in Play mode. See discussion at the end.
So I think you might want .drag-activity-correct [data-no-target], .drag-activity-wrong [data-draggable-id]. (Only if you go with my suggestion of really moving the correct ones, though.)
</blockquote></details>

I've revised these rules repeatedly in the last day.
___
*[`src/content/templates/template books/Games/Games.less` line 55 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6976#-ONQGnYk4zVwISOOI_dI:-ONWnCoG0hABZxLg23Wk:b5o8spm) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/d96e073609f988f2190d8c211bcf6dfa1cc06a31/src/content/templates/template%20books/Games/Games.less#L55)):*

I'm not sure what this is getting at. You already hid the draggables. 
So I think the effect of this will be to hide images, videos, and gifs that belong to the start tab and are part of the scene (not draggable in Play). I don't think John asked for that.
</blockquote></details>

I don't think any of the current CSS would have this effect.

StephenMcConnel (commented):
This PR now needs an updated bloom-player.  (https://github.com/BloomBooks/bloom-player/pull/369)  If it gets merged without updating the bloom-player version in src/BloomBrowserUI/packages.json, another PR will be needed to get things all working.

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6976#-:-ONWo9ac1Ias919Hyebg:b-56rzyd)* status: 2 of 4 files reviewed, 8 unresolved discussions (waiting on @andrew-polk and @JohnThomson)

andrew-polk (commented):
> Reviewed 1 of 2 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6976)* status: 3 of 4 files reviewed, 8 unresolved discussions (waiting on @JohnThomson)

JohnThomson (approved):
> Reviewed 2 of 4 files at r3, 2 of 2 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6976)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @StephenMcConnel)

PR #6982, opened 2025-04-10
Make letter boxes grow wide enough for letters (BL-14541)

JohnThomson:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6982)


andrew-polk:
Marking as draft as I think it is quite different than the current spec. (Correct me if I'm wrong.)

PR #6984, opened 2025-04-11
Fix expired subscription message (BL-14563)

andrew-polk:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6984)

### DIFF
--- a/DistFiles/localization/en/Bloom.xlf
+++ b/DistFiles/localization/en/Bloom.xlf
@@ -5967,12 +5967,14 @@ Do you want to go ahead?</note>
       <trans-unit id="SubscriptionStatus.ExpiringSoonMessage">
         <source xml:lang="en">Your {0} subscription expires on {1}.</source>
         <note>ID: SubscriptionStatus.ExpiringSoonMessage</note>
-        <note>{0} is the branding project key, {1} is the formatted expiration date</note>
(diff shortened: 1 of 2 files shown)


andrew-polk (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6984)* status: 0 of 2 files reviewed, all discussions resolved (waiting on @JohnThomson)

___
*[`src/BloomBrowserUI/collectionsTab/SubscriptionStatus.tsx` line 77 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6984#-ONaJqFDAb21iA4FzXq7:-ONaXPk8DGmhMo55nak-:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/54240daa8d9a5fc7c99dfee7c3ba24797efcaef1/src/BloomBrowserUI/collectionsTab/SubscriptionStatus.tsx#L77)):*

Leaving it blank works in English, especially with the double-space suppressed, but what if a translation results in something like "Your subscription called "" expired on July 15"?
I don't see how to improve things here, but could we at least add a note to the XLF that the translation should work when {0} is empty?
</blockquote></details>

Done.

PR #6986, opened 2025-04-11
Fix publication image cropping for non-ASCII filepaths (BL-14566)

andrew-polk:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6986)

### DIFF
--- a/src/BloomExe/ImageProcessing/ImageUtils.cs
+++ b/src/BloomExe/ImageProcessing/ImageUtils.cs
@@ -1371,34 +1371,47 @@ namespace Bloom.ImageProcessing
         public static bool TryGetImageSize(string path, out Size size)
         {
             size = new Size(0, 0);
-            try
-            {
-                var arguments = "identify -format %P \"" + path + "\"";
-                var result = CommandLineRunnerExtra.RunWithInvariantCulture(
-                    GetGraphicsMagickPath(),
-                    arguments,
-                    "",
-                    600,
-                    new NullProgress()
-                );
-                if (result.ExitCode == 0)
+            var calculatedSize = new Size(0, 0);
+
+            var result = WithSafeFilePath(
+                path,
+                safePath =>
                 {
-                    var parts = result.StandardOutput.Split('x');
-                    if (parts.Length == 2)
+                    try
                     {
-                        size.Width = int.Parse(parts[0]);
-                        size.Height = int.Parse(parts[1]);
-                        return true;
+                        var arguments = "identify -format %P \"" + safePath + "\"";
+                        var result = CommandLineRunnerExtra.RunWithInvariantCulture(
+                            GetGraphicsMagickPath(),
+                            arguments,
+                            "",
+                            600,
+                            new NullProgress()
+                        );
+                        if (result.ExitCode == 0)
+                        {
+                            var parts = result.StandardOutput.Split('x');
+                            if (parts.Length == 2)
+                            {
+                                calculatedSize.Width = int.Parse(parts[0]);
+                                calculatedSize.Height = int.Parse(parts[1]);
+                                return true;
+                            }
+                        }
+
+                        return false;
+                    }
+                    catch (Exception e)
+                    {
+                        Console.WriteLine(e);
+                        return false;
                     }
                 }
+            );
 
-                return false;
-            }
-            catch (Exception e)
-            {
-                Console.WriteLine(e);
-                return false;
-            }
+            if (result)
+                size = calculatedSize;
+
+            return result;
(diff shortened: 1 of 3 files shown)


andrew-polk:
*[`src/BloomExe/ImageProcessing/ImageUtils.cs` line 1461 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6986#-ONb5TUD5cjWfP5AbwxJ:-ONb5TUD5cjWfP5AbwxK:bnadpkt) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/1f91fc73743d430c5b3272be8ae3c9563764990e/src/BloomExe/ImageProcessing/ImageUtils.cs#L1461)):*
> ```Smalltalk
>                         {
>                             var argsBldr = new StringBuilder();
>                             argsBldr.AppendFormat("convert \"{0}\"", safeSourcePath);
> ```

My intention was to leave the logic in this method as it was but to extract the safe path logic to a shared method.

However, this change from sourcePath to safeSourcePath was made because it represented a bug in the original code. We weren't using safeSourcePath previously.

hatton (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6986)* status: all files reviewed, 2 unresolved discussions (waiting on @andrew-polk)

___
*[`src/BloomExe/ImageProcessing/ImageUtils.cs` line 1497 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6986#-ONerabP-E3fiDIhqRKV:-ONerabP-E3fiDIhqRKW:b-xdv5sr) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/1f91fc73743d430c5b3272be8ae3c9563764990e/src/BloomExe/ImageProcessing/ImageUtils.cs#L1497)):*
> ```Smalltalk
>                                 argsBldr.Append(" -quality 75"); // no lossage in output, use adaptive filter in changing image dimensions
>                             else if (
>                                 destPath.EndsWith(
> ```

I notice we're still getting a lot of formatting churn. That shouldn't be, right? We should have a completely re-formatted code base with `.git-blame-ignore-revs` entries.
___
*[`src/BloomExe/ImageProcessing/ImageUtils.cs` line 1987 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6986#-ONesRCj41FkHizH8cM-:-ONesRCj41FkHizH8cM0:b-1k4ck4) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/1f91fc73743d430c5b3272be8ae3c9563764990e/src/BloomExe/ImageProcessing/ImageUtils.cs#L1987)):*
> ```Smalltalk
>         ///
>         /// Original comment before this was factored out into a separate wrapping method:
>         ///  We have no idea what the input (and output) file names are, and whether they can be safely represented
> ```

I didn't care for the original comment, and I propose you not carry it forward. "we have no idea"...

I think you comment above it plenty?

andrew-polk (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6986)* status: 2 of 3 files reviewed, 2 unresolved discussions (waiting on @hatton)

___
*[`src/BloomExe/ImageProcessing/ImageUtils.cs` line 1497 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6986#-ONerabP-E3fiDIhqRKV:-ONp8GOA9MSVvCVBysSt:b3zrw4j) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/1f91fc73743d430c5b3272be8ae3c9563764990e/src/BloomExe/ImageProcessing/ImageUtils.cs#L1497)):*

I notice we're still getting a lot of formatting churn. That shouldn't be, right? We should have a completely re-formatted code base with `.git-blame-ignore-revs` entries.
</blockquote></details>

I think this change was a result of CSharpier hitting the max line limit as a result of extra indentation and thus reformating.
___
*[`src/BloomExe/ImageProcessing/ImageUtils.cs` line 1987 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/6986#-ONesRCj41FkHizH8cM-:-ONp8PCd-dDKiDXYM49a:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/1f91fc73743d430c5b3272be8ae3c9563764990e/src/BloomExe/ImageProcessing/ImageUtils.cs#L1987)):*

I didn't care for the original comment, and I propose you not carry it forward. "we have no idea"...

I think you comment above it plenty?
</blockquote></details>

Done.

PR #7011, opened 2025-04-23
Provide enough space for the DragActivityTabControl widget (BL-14614)

StephenMcConnel:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7011)

### DIFF
--- a/src/BloomBrowserUI/bookEdit/css/editMode.less
+++ b/src/BloomBrowserUI/bookEdit/css/editMode.less
@@ -1303,6 +1303,14 @@ body.hideAllCKEditors .cke_chrome {
     top: -8px;
     width: 100%;
 }
+// Very narrow pages need more space for the #drag-activity-tab-control
+.above-page-control-container:has(+ .Device16x9Portrait),
+.above-page-control-container:has(+ .QuarterLetterPortrait),
+.above-page-control-container:has(+ .A6Portrait) {
+    #drag-activity-tab-control {
+        top: -22px;
+    }
+}
 
 // we want :blank, but it's not in browsers yet. So we simulate it by a mutation observer
 // that adds a class to the editable when it is empty.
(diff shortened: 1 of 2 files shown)


andrew-polk:
Converting to draft as I think the introduction of the gap for other sizes is a show-stopper.
And I'm deprioritizing this card.

StephenMcConnel (commented):
I figured out how to avoid increasing the gap when it's not needed.

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7011#-:-OOdq3376vzuVhLjiXpA:b-vas0bj)* status: 0 of 2 files reviewed, all discussions resolved

andrew-polk (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7011)* status: all files reviewed, 1 unresolved discussion (waiting on @StephenMcConnel)

___
*[`src/BloomBrowserUI/bookEdit/css/editMode.less` line 1313 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7011#-OOhXJMjEp1bcra8mxV3:-OOhXJMk1pAXwvw-bUnv:b-edfego) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/124f7756da8afc5ee8f464f3079d75917e44ca2e/src/BloomBrowserUI/bookEdit/css/editMode.less#L1313)):*
> ```Less
>         top: -22px;
>     }
> }
> ```

It looks like we can do this dynamically with 
```
.above-page-control-container {
    container-type: inline-size;
    container-name: above-page-control;
}
@container above-page-control (max-width: 450px) {
    #drag-activity-tab-control {
        top: -22px;
    }
}
```

(I don't what the 450 should be)

But I'm not sure this is sufficient.
Probably we need JH to take a look since the wrapped text looks bad to me. Particularly that the button and text next to it wrap differently (aren't lined up).
There's also a problem when zooming in. The controls appear to move up relative to the view and thus start to get cut off on top.

StephenMcConnel:
obsolete by this point.  probably not the right approach anyway.

PR #7017, opened 2025-04-24
Allow choosing from multiple CMYK color profile (.icc) files (BL-14666)

StephenMcConnel:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7017)


andrew-polk (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7017)* status: all files reviewed, 2 unresolved discussions (waiting on @StephenMcConnel)

___
*[`src/BloomExe/Publish/PublishModel.cs` line 687 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7017#-OOdnAHo6JzaPUojro4x:-OOdnAHo6JzaPUojro4y:b-4ht2w5) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/b1bb759b8916a70a6931d9bdf42d6b8faf0f4ad8/src/BloomExe/Publish/PublishModel.cs#L687)):*
> ```Smalltalk
>                 if (!String.IsNullOrEmpty(_currentlyLoadedBook.UserPrefs.ColorProfileForPdf))
>                 {
>                     // PDF for Printshop (CMYK US Web Coated V2)
> ```

this comment is now obsolete
___
*[`src/BloomBrowserUI/publish/PDFPrintPublish/PDFPrintFeaturesGroup.tsx` line 211 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7017#-OOdmC6SCrQEQ4UUDJaq:-OOdmC6T5hsZzXti3O_x:bgx1nj4) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/b1bb759b8916a70a6931d9bdf42d6b8faf0f4ad8/src/BloomBrowserUI/publish/PDFPrintPublish/PDFPrintFeaturesGroup.tsx#L211)):*
> ```XML
>                                         {profile}
>                                     </MenuItem>
>                                 ))}
> ```

I think we want to maintain the internationalization of profiles we ship?
Granted, it probably isn't all that helpful to localize them... but we already do have that in place and don't want to lose it.
Plus, if I understand the change, we used to display "U.S. Web Coated (SWOP) v2" but will now display "USWebCoatedSWOP".

And if you do that internationalization here in ts land (which I think makes the most sense?), I think you might as well move the internationalization of "None" here, too.

andrew-polk (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7017)* status: all files reviewed, 3 unresolved discussions (waiting on @StephenMcConnel)

___
*[`src/BloomExe/Publish/PDF/PublishPdfApi.cs` line 160 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7017#-OOdp0at1zUYMAb84Y78:-OOdp0auCLdmjQXq5bxh:b-da74zc) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/b1bb759b8916a70a6931d9bdf42d6b8faf0f4ad8/src/BloomExe/Publish/PDF/PublishPdfApi.cs#L160)):*
> ```Smalltalk
>                 var none = LocalizationManager.GetString("PublishTab.PdfMaker.None", "None");
>                 result.Append($"[\"{none}\"");
>                 foreach (var file in files)
> ```

Do we need to filter out the AdobeRGB1998 one?

StephenMcConnel (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7017)* status: all files reviewed, 3 unresolved discussions (waiting on @andrew-polk)

___
*[`src/BloomBrowserUI/publish/PDFPrintPublish/PDFPrintFeaturesGroup.tsx` line 211 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7017#-OOdmC6SCrQEQ4UUDJaq:-OOdqWu19Mtjn73MEGKe:bn93htf) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/b1bb759b8916a70a6931d9bdf42d6b8faf0f4ad8/src/BloomBrowserUI/publish/PDFPrintPublish/PDFPrintFeaturesGroup.tsx#L211)):*

I think we want to maintain the internationalization of profiles we ship?
Granted, it probably isn't all that helpful to localize them... but we already do have that in place and don't want to lose it.
Plus, if I understand the change, we used to display "U.S. Web Coated (SWOP) v2" but will now display "USWebCoatedSWOP".

And if you do that internationalization here in ts land (which I think makes the most sense?), I think you might as well move the internationalization of "None" here, too.
</blockquote></details>

I thought we agreed to use filenames going to allowing multiple (open-ended) files.  You can't internationalize file names.  And even if I figured out how to extract the description string from the .icc file, it's not going to be internationalized in any form.
I suppose I could leave the .icc endings on the filenames to make it more clear that they're filenames.
I appreciate that you don't want to throw away any localizations, but "U.S. Web Coated (SWOP) v2" barely is more understandable than (USWebCoatedSWOP).  They're both gobbledygook to users (and to most developers as well).
___
*[`src/BloomExe/Publish/PublishModel.cs` line 687 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7017#-OOdnAHo6JzaPUojro4x:-OOdrbML8PYR-r7LB8pm:b-lstdur) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/b1bb759b8916a70a6931d9bdf42d6b8faf0f4ad8/src/BloomExe/Publish/PublishModel.cs#L687)):*

this comment is now obsolete
</blockquote></details>

I changed the comment to "(CMYK color conversion)"
___
*[`src/BloomExe/Publish/PDF/PublishPdfApi.cs` line 160 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7017#-OOdp0at1zUYMAb84Y78:-OOdrlm76Y5gGzl0IKQ-:b-qls9fu) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/b1bb759b8916a70a6931d9bdf42d6b8faf0f4ad8/src/BloomExe/Publish/PDF/PublishPdfApi.cs#L160)):*

Do we need to filter out the AdobeRGB1998 one?
</blockquote></details>

Why?  I strongly don't want to special case things if we're allowing multiple files here.  That's why I include a "none" string in the data sent to UI code.  That simplifies the creation of the menu.

StephenMcConnel (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7017)* status: 1 of 7 files reviewed, 3 unresolved discussions (waiting on @andrew-polk)

___
*[`src/BloomBrowserUI/publish/PDFPrintPublish/PDFPrintFeaturesGroup.tsx` line 211 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7017#-OOdmC6SCrQEQ4UUDJaq:-OOwrFgg9nhKSrINTvxc:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/b1bb759b8916a70a6931d9bdf42d6b8faf0f4ad8/src/BloomBrowserUI/publish/PDFPrintPublish/PDFPrintFeaturesGroup.tsx#L211)):*

After sleeping on it, I think it should be straightforward to maintain the i18n of the two existing items, and to restrict the l10n work to the .tsx side.  I'll get this done on Monday.
</blockquote></details>

Done.

andrew-polk (changes requested):
JH wants to have the users add .icc files somewhere else like maybe the collection.
My understanding is this has now shifted down in priority since it doesn't solve the user's immediate issue.
So I think we can wait for a more clear spec now.

> Reviewed 1 of 1 files at r2, 6 of 6 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7017#-:-OOxf6Wj5ILDYcSr9jMw:b-nr55ys)* status: all files reviewed, 1 unresolved discussion (waiting on @StephenMcConnel)

___
*[`src/BloomExe/Publish/PDF/PublishPdfApi.cs` line 160 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7017#-OOdp0at1zUYMAb84Y78:-OOxelNSCcwB50pzSySh:b-ss7c1x) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/b1bb759b8916a70a6931d9bdf42d6b8faf0f4ad8/src/BloomExe/Publish/PDF/PublishPdfApi.cs#L160)):*

Why?  I strongly don't want to special case things if we're allowing multiple files here.  That's why I include a "none" string in the data sent to UI code.  That simplifies the creation of the menu.
</blockquote></details>

Maybe I'm misunderstanding.
I thought this would introduce the profile into the list whereas it wasn't there before.
Unless we make a decision otherwise, I think we want the behavior to remain the same as it is now unless the user adds an .icc file.

JohnThomson (changes requested):
Even a very minor new release of Bloom generates a new installed folder with the version number in its name, to which the desired icc files must be added again, at least if we continue to look for them in the distribution folder.
Could we somehow migrate them from one release to another? For example, I think there's a function somewhere in InstallerSupport.cs that is executed when a new version first runs; it could look for icc folders in any other release in the same parent folder and copy them over.
Minimally, I've commented a couple of places where I think we might need to better handle not finding a file that corresponds to the current setting.
But I do think this is a reason to think of some other place for the user to add these. Maybe a CMYK directory in the project folder, or in %localappdata%/SIL/Bloom?

> Reviewed 1 of 6 files at r1, 6 of 6 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7017#-:-OQnzPAJCPP_miDeKMrn:bq4unkn)* status: all files reviewed, 4 unresolved discussions (waiting on @andrew-polk)

___
*[`src/BloomExe/Publish/PDF/PublishPdfApi.cs` line 160 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7017#-OOdp0at1zUYMAb84Y78:-OQo3iIXC6qKoSxWV32g:bzbxej2) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/b1bb759b8916a70a6931d9bdf42d6b8faf0f4ad8/src/BloomExe/Publish/PDF/PublishPdfApi.cs#L160)):*

Maybe I'm misunderstanding.
I thought this would introduce the profile into the list whereas it wasn't there before.
Unless we make a decision otherwise, I think we want the behavior to remain the same as it is now unless the user adds an .icc file.
</blockquote></details>

We don't need to filter out AdobeRGB1998.icc because it's in the RGB folder, not the CYMK one. Unless of course the user copies it over because for some reason his publisher wants it, in which case we'd better show it.
___
*[`src/BloomBrowserUI/publish/PDFPrintPublish/PDFPrintFeaturesGroup.tsx` line 201 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7017#-OQnxvAlDFX6BJXnzYKk:-OQnxvAlDFX6BJXnzYKl:b-afky15) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/b4472dcf881cc5c85fa3c53f6a49e5d4644e00fc/src/BloomBrowserUI/publish/PDFPrintPublish/PDFPrintFeaturesGroup.tsx#L201)):*
> ```XML
>                                 `}
>                                 variant="outlined"
>                                 value={colorProfile}
> ```

What will (and should) happen here if colorProfile does not correspond to any of the menu items?
___
*[`src/BloomExe/Book/UserPrefs.cs` line 96 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7017#-OQnxC2v1MyRZZIV_4rK:-OQnxC2v1MyRZZIV_4rL:b3e5f6n) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/b4472dcf881cc5c85fa3c53f6a49e5d4644e00fc/src/BloomExe/Book/UserPrefs.cs#L96)):*
> ```Smalltalk
>                 Save();
>             }
>         }
> ```

Is there something we can easily do so that an old userprefs with a boolean cmkPdf true will migrate to the appropriate new value?

I don't think this is worth hours, much less days, but it might be worth a half hour or so.
___
*[`src/BloomExe/Publish/PDF/ProcessPdfWithGhostscript.cs` line 257 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7017#-OQnyLjc1BxyHtPec5TS:-OQnyLjd2SP6m8NXlifk:b-det11k) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/b4472dcf881cc5c85fa3c53f6a49e5d4644e00fc/src/BloomExe/Publish/PDF/ProcessPdfWithGhostscript.cs#L257)):*
> ```Smalltalk
>                     var cmykProfile = FileLocationUtilities.GetFileDistributedWithApplication(
>                         $"ColorProfiles/CMYK/{_colorProfile}.icc"
>                     );
> ```

Seems like we should do something friendlier if the file is not found, as it well may not be if the user added it to one release of Bloom and then updated. (I believe this method will crash, probably with a message saying that Bloom has been corrupted by antivirus and can't find one of its own files.)

StephenMcConnel (commented):
I've fixed the code to (also) look in %localappdata%/SIL/Bloom/ColorProfiles/CMYK for user installed profiles.  

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7017#-:-OQoL3i9DhA2AtiNgysN:bngh1yr)* status: all files reviewed, 4 unresolved discussions (waiting on @andrew-polk and @JohnThomson)

___
*[`src/BloomBrowserUI/publish/PDFPrintPublish/PDFPrintFeaturesGroup.tsx` line 201 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7017#-OQnxvAlDFX6BJXnzYKk:-OQoM8B4D0SJPe_hOV8g:bfzylhp) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/b4472dcf881cc5c85fa3c53f6a49e5d4644e00fc/src/BloomBrowserUI/publish/PDFPrintPublish/PDFPrintFeaturesGroup.tsx#L201)):*

What will (and should) happen here if colorProfile does not correspond to any of the menu items?
</blockquote></details>

I don't think that can happen.  I've added a sanity check to the onChange handler.  I also added the default value "none" to the colorProfiles array.
___
*[`src/BloomExe/Book/UserPrefs.cs` line 96 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7017#-OQnxC2v1MyRZZIV_4rK:-OQom_qf6ISLxgQNF5_K:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/b4472dcf881cc5c85fa3c53f6a49e5d4644e00fc/src/BloomExe/Book/UserPrefs.cs#L96)):*

Is there something we can easily do so that an old userprefs with a boolean cmkPdf true will migrate to the appropriate new value?

I don't think this is worth hours, much less days, but it might be worth a half hour or so.
</blockquote></details>

Done.

JohnThomson (changes requested):
> Reviewed 5 of 5 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7017)* status: all files reviewed, 3 unresolved discussions (waiting on @andrew-polk and @StephenMcConnel)

___
*[`src/BloomBrowserUI/publish/PDFPrintPublish/PDFPrintFeaturesGroup.tsx` line 201 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7017#-OQnxvAlDFX6BJXnzYKk:-OQtHrh44dln0LYVPm2c:bfuhago) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/b4472dcf881cc5c85fa3c53f6a49e5d4644e00fc/src/BloomBrowserUI/publish/PDFPrintPublish/PDFPrintFeaturesGroup.tsx#L201)):*

I don't think that can happen.  I've added a sanity check to the onChange handler.  I also added the default value "none" to the colorProfiles array.
</blockquote></details>

I didn't mean in the onChange handler. What if the value saved in colorProfile (when we're loading the control) is the name of an icc that has since been deleted, so we don't create a corresponding menu item, because it's no longer in the colorProfiles list?
___
*[`src/BloomExe/Publish/PDF/ProcessPdfWithGhostscript.cs` line 261 at r4](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7017#-OQtJFoH2-o4eJjvdNr-:-OQtJFoH2-o4eJjvdNr0:b-y0yhga) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/10c3120780bc15d2da2fc40b2bfbe1feeb20d6a6/src/BloomExe/Publish/PDF/ProcessPdfWithGhostscript.cs#L261)):*
> ```Smalltalk
>                         var userFolder = PdfMaker.GetUserColorProfilesFolder();
> 						cmykProfile = Path.Combine(userFolder, _colorProfile + ".icc");
> 					}
> ```

It still might not exist, if we saved the name of a user-added icc file but it later got deleted, or if we saved it on one computer but now we're publishing on a different one. Maybe it's too unlikely to worry about, but I'd be happier if we could put some error message in the log, even if we don't bother to localize it.

JohnThomson (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7017)* status: all files reviewed, 1 unresolved discussion (waiting on @andrew-polk)

PR #7044, opened 2025-05-06
Copy sound file and attributes when copying canvas editables (BL-14715)

StephenMcConnel:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7044)

### DIFF
--- a/src/BloomBrowserUI/bookEdit/js/CanvasElementManager.ts
+++ b/src/BloomBrowserUI/bookEdit/js/CanvasElementManager.ts
@@ -53,6 +53,8 @@ import { getTarget } from "bloom-player";
 import { CanvasGuideProvider } from "./CanvasGuideProvider";
 import { CanvasElementKeyboardProvider } from "./CanvasElementKeyboardProvider";
 import { CanvasSnapProvider } from "./CanvasSnapProvider";
+import { postData, postJson } from "../../utils/bloomApi";
+import AudioRecording from "../toolbox/talkingBook/audioRecording";
 
 export interface ITextColorInfo {
     color: string;
@@ -5252,6 +5254,11 @@ export class CanvasElementManager {
                 sourceDataSound
             );
         }
+        // copy any sound files found in an editable div
+        this.copyAnySoundFileAndAttributesForEditable(
+            sourceElement,
+            patriarchDuplicateElement
+        );
 
         this.setActiveElement(patriarchDuplicateElement);
         this.matchSizeOfSource(sourceElement, patriarchDuplicateElement);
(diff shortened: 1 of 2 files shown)


andrew-polk (changes requested):
I'm slightly concerned about the approach.
It looks like we had made a decision in the past to simply remove all audio "stuff" from the duplicate.
Now we are making a duplicate with correctly-working audio "stuff". That could certainly be considered an improvement. But I wonder if we need to clarify with JH. I'll post something in the card.

Did you weigh the pros/cons of just making the error not happen? i.e. there must be some other piece which is getting partially copied and causing the duplicate to think it has something to play when it doesn't.

I think in an ideal world, if we are keeping the current approach, it would be better to create the clone with a new ID and the existing data-duration rather than remove them at clone time, then go back and add them in later. I'm not sure how difficult that is...

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7044#-:-OPauqWeBSfr9gKti966:b-wzd21)* status: all files reviewed, 2 unresolved discussions (waiting on @StephenMcConnel)

___
*[`src/BloomBrowserUI/bookEdit/js/CanvasElementManager.ts` line 5278 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7044#-OPavtgK3uwkfdVVpcPT:-OPavtgK3uwkfdVVpcPU:b-8qwqnh) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/844809800cd04a62dd115210a3adede21571bbdb/src/BloomBrowserUI/bookEdit/js/CanvasElementManager.ts#L5278)):*
> // REVIEW: is this needed? or wanted?

Probably if we are going with the improvement of actually copying at the parent level, we want the child level, too.
___
*[`src/BloomBrowserUI/bookEdit/js/CanvasElementManager.ts` line 5195 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7044#-OPaqjSRD7j1_ul6PX_X:-OPaqjSS1hlYmAkykyHs:b-qw927o) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/844809800cd04a62dd115210a3adede21571bbdb/src/BloomBrowserUI/bookEdit/js/CanvasElementManager.ts#L5195)):*
> ```XML
>         }
>         return newId;
>     }
> ```

Can you use EditableDivUtils.createUuid()?

andrew-polk (commented):
Decision: make a full copy with audio.

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7044#-:-OPbJuxbA5BYkIADdg-c:bftvfe6)* status: all files reviewed, 2 unresolved discussions (waiting on @StephenMcConnel)

andrew-polk (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7044)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @StephenMcConnel)

PR #7068, opened 2025-05-12
BL-14621 allow moving book from tc

nabalone:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7068)

### DIFF
--- a/src/BloomExe/CollectionTab/CollectionModel.cs
+++ b/src/BloomExe/CollectionTab/CollectionModel.cs
@@ -194,14 +194,42 @@ namespace Bloom.CollectionTab
             var possibleTCFilePath = TeamCollectionManager.GetTcLinkPathFromLcPath(
                 origCollection.PathToDirectory
             );
-            if (RobustFile.Exists(possibleTCFilePath))
+            var origCollectionIsTC = RobustFile.Exists(possibleTCFilePath);
+
+            TeamCollectionManager origTcManager = null;
+            if (origCollectionIsTC)
             {
-                // Original collection is a TC.
-                // To remove a book from a TC, we would have "load up" the collection to check that the book is checked
-                // out, the tc is connected, etc. So for now we don't allow it
-                throw new ApplicationException(
-                    $"{origCollection.Name} is a Team Collection. Cannot move a book that is currently in a Team Collection."
+                // Original collection is a TC. We should only allow the move if the book is already checked out in its
+                // original tc, and the original tc is connected so we can properly delete it there
+
+                var origSettingsFilePath = TeamCollection.TeamCollection.CollectionPath(
+                    origCollection.PathToDirectory
                 );
+
+                // This team collection manager is only used to check and delete the book from the original tc, so it
+                // doesn't need the normal objects, and it's probably safer not to give it objects which might have
+                // connections to the current collection.
+                origTcManager = new TeamCollectionManager(
+                    origSettingsFilePath,
+                    null,
+                    null,
+                    null,
+                    null,
+                    null
+                );
+                if (origTcManager.CannotDeleteBecauseDisconnected(origBook))
+                {
+                    throw new ApplicationException(
+                        $"{origCollection.Name} is a Team Collection which is currently disconnected. Please connect the Team Collection before moving books that are part of it."
+                    );
+                }
+                var tc = origTcManager.CurrentCollection;
+                if (tc == null || !tc.CanSaveChanges(origBook.BookInfo))
+                {
+                    throw new ApplicationException(
+                        $"{origCollection.Name} is a Team Collection. Please open {origCollection.Name} and check out {origBook.BookInfo.Title} before moving it."
+                    );
+                }
             }
 
             var origBookFolderPath = origBook.FolderPath;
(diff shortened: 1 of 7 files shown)


src/BloomExe/CollectionTab/CollectionModel.cs
```diff
@@ -194,21 +197,63 @@ public void moveBookIntoThisCollection(Book.Book origBook, BookCollection origCo
...
+                // book is already checked out in its original tc, and the original tc is connected so we can properly delete it there
+
+                if (_tcManager.CollectionStatus == TeamCollectionStatus.None)
+                {
+                    throw new ApplicationException(
+                        $"{origCollection.Name} is a Team Collection but {TheOneEditableCollection.Name} is not. Cannot move a book from a Team Collection to a non-Team Collection."
+                    );
+                }
+
+                var settingsFilePath = Path.Combine(
```
nabalone, line 213: TODO we have various code with various different logic for finding the settings file throughout the repo (search for "bloomCollection"). Compare them and factor out a method that can be used everywhere? Or at least make sure we are using the correct logic here

src/BloomExe/CollectionTab/CollectionModel.cs
```diff
@@ -194,21 +197,63 @@ public void moveBookIntoThisCollection(Book.Book origBook, BookCollection origCo
...
             var bookFolderName = Path.GetFileName(origBookFolderPath);
             var newCollectionDir = TheOneEditableCollection.PathToDirectory;
             var newBookFolderPath = Path.Combine(newCollectionDir, bookFolderName);
             SIL.IO.RobustIO.MoveDirectory(origBookFolderPath, newBookFolderPath);
+            if (origCollectionIsTC)
+            {
+                _tcManager.CurrentCollection.DeleteBookFromRepo(origBookFolderPath);
+
+                // Get rid of any TC status we copied over
+                BookStorage.RemoveLocalOnlyFiles(newBookFolderPath);
```
nabalone, line 254: Do we indeed want to remove all these files when moving between collections? They mostly look collection-specific

src/BloomExe/CollectionTab/CollectionModel.cs
nabalone: I have now switched to using the tc logic here.

JohnThomson: Looks like there isn't really still a Todo here? Unless it seems worth making a Technical debt card to try to clean up the various ways we access collection settings?

src/BloomExe/CollectionTab/CollectionModel.cs
JohnThomson: I think they're all appropriate to remove. The status file is all about its checkout status in the old TC; the sync time tells when it was last synced with that TC; the link makes it part of the collection it just got moved out of; and impersonate.txt is a trick for pretending it is owned by someone else (for testing, hopefully) in that TC. I don't remember why log.txt is considered a TC-specific file, but since it is, it probably should be deleted.

JohnThomson (changes requested):
Looks good. A couple of things worth thinking about, but if they prove difficult I think we can merge as-is.

> Reviewed 3 of 4 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7068#-:-OQMDZr35UNkJHEoI6AD:bos339a)* status: all files reviewed, 2 unresolved discussions (waiting on @nabalone)

___
*[`src/BloomExe/CollectionTab/CollectionModel.cs` line 214 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7068#-OQMAMIR2pldXWHTdmAx:-OQMAMIR2pldXWHTdmAy:b-tq177o) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/cb4658ed26862cd6ac9ecef258959a69caea7ee3/src/BloomExe/CollectionTab/CollectionModel.cs#L214)):*
> ```Smalltalk
>                     origSettingsFilePath,
>                     _webSocketServer,
>                     _bookRenamedEvent,
> ```

I don't much like passing our bookRenamedEvent to a TC manager for the other TC. Might it cause the other TC to do something inappropriate when a book is renamed in our collection? Might also prevent the other TC getting garbage-collected, if it sets up a connection of some sort to our BRE.
For this situation it probably doesn't actually need a BRE. Can you pass null? (might require making the TC constructor skip something when it's null). Or, just directly construct a separate BRE if that's not difficult.
I'm also wondering if the other TC manager will send something to our web socket when we tell it to delete the book. No doubt you've tried it and it wasn't a problem. But again, it would feel safer if we didn't have to.

src/BloomExe/CollectionTab/CollectionModel.cs
nabalone: Created https://issues.bloomlibrary.org/youtrack/issue/BL-14788/tech-debt-a-consistent-method-for-accessing-the-collectionSettings-file

nabalone (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7068)* status: 1 of 7 files reviewed, 2 unresolved discussions (waiting on @JohnThomson)

___
*[`src/BloomExe/CollectionTab/CollectionModel.cs` line 214 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7068#-OQMAMIR2pldXWHTdmAx:-OQPCPCgB5BNQqNUXNbC:b-5p8gco) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/cb4658ed26862cd6ac9ecef258959a69caea7ee3/src/BloomExe/CollectionTab/CollectionModel.cs#L214)):*

I don't much like passing our bookRenamedEvent to a TC manager for the other TC. Might it cause the other TC to do something inappropriate when a book is renamed in our collection? Might also prevent the other TC getting garbage-collected, if it sets up a connection of some sort to our BRE.
For this situation it probably doesn't actually need a BRE. Can you pass null? (might require making the TC constructor skip something when it's null). Or, just directly construct a separate BRE if that's not difficult.
I'm also wondering if the other TC manager will send something to our web socket when we tell it to delete the book. No doubt you've tried it and it wasn't a problem. But again, it would feel safer if we didn't have to.
</blockquote></details>

I got the BRE out of there. The web socket situation looks more complicated. Simply sticking in a new webSocketServer causes problems. It's not obvious to me how to address this, would it be worth pairing on it?

JohnThomson (approved):
> Reviewed 5 of 6 files at r3, 1 of 1 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7068)* status: all files reviewed, 1 unresolved discussion (waiting on @nabalone)

PR #7069, opened 2025-05-12
Move default game user styles to html template (BL-14732)

andrew-polk:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7069)

### DIFF
--- a/src/BloomBrowserUI/bookEdit/toolbox/games/GameTool.tsx
+++ b/src/BloomBrowserUI/bookEdit/toolbox/games/GameTool.tsx
@@ -1164,7 +1164,7 @@ const DragActivityControls: React.FunctionComponent<{
                                         l10nKey="EditTab.Toolbox.DragActivity.Letter"
                                         makeTarget={true}
                                         addClasses="draggable-text"
-                                        userDefinedStyleName="GameDraggableLetterMediumCenter"
+                                        userDefinedStyleName="GameDragMediumCenter"
                                     />
                                 )}
                                 {showImageDraggable && (
@@ -1205,11 +1205,6 @@ const DragActivityControls: React.FunctionComponent<{
                                     addClasses="draggable-text"
                                     hide={!showWordDraggable}
                                     userDefinedStyleName="GameDragMediumCenter"
-                                    defaultStyleProps={{
-                                        fontSize: "30pt",
-                                        lineHeight: "1",
-                                        textAlign: "center"
-                                    }}
                                 />
 
                                 {/* Slider: rather than reinstating this item, make the "selected item is part of answer" control work.
@@ -1416,11 +1411,6 @@ const GameTextItem: React.FunctionComponent<{
             makeTarget={false}
             addClasses={props.addClasses}
             userDefinedStyleName="GameTextMediumCenter"
-            defaultStyleProps={{
-                fontSize: "30pt",
-                lineHeight: "1",
-                textAlign: "center"
-            }}
         />
     );
 };
--- a/src/BloomExe/Book/Book.cs
+++ b/src/BloomExe/Book/Book.cs
@@ -3784,7 +3784,6 @@ namespace Bloom.Book
         /// different book, we may need to copy user-defined styles from that book to our own.
         /// </summary>
         /// <returns>true if anything added</returns>
-        /// <param name="templatePage"></param>
         private bool AddMissingStylesFromTemplatePage(IPage templatePage)
         {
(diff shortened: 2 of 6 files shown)


JohnThomson (changes requested):
> Reviewed 1 of 5 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7069)* status: 1 of 5 files reviewed, 4 unresolved discussions

___
*[`src/content/templates/template books/Games/Games.less` line 624 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7069#-OQ55IDb4c3N-LVVPOIe:-OQ55IDb4c3N-LVVPOIf:b-tepczm) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/e4223afb627fff1121a6aed01d054d492d7ec0db/src/content/templates/template%20books/Games/Games.less#L624)):*
> ```Less
> // These styles and more are set up in the Games.html template.
> // An attribute on the body (data-copy-all-styles) causes them
> // all to get added when any game page is added to a book.
> ```

Maybe say a little more along the lines that the main definition is in Games.htm but these properties are defined here so that...is it because the HTML definitions can't just use variables like this for some reason?
___
*[`src/content/templates/template books/Games/Games.html` line 14 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7069#-OQ548wc-zf216TMsBpR:-OQ548wc-zf216TMsBpS:b-7dou83) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/e4223afb627fff1121a6aed01d054d492d7ec0db/src/content/templates/template%20books/Games/Games.html#L14)):*
> ```HTML
> 
>         <title>Game Templates</title>
>

andrew-polk (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7069)* status: 1 of 5 files reviewed, 4 unresolved discussions (waiting on @JohnThomson)

___
*[`src/content/templates/template books/Games/Games.html` line 14 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7069#-OQ548wc-zf216TMsBpR:-OQ59upz5lYrPyvC1m-G:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/e4223afb627fff1121a6aed01d054d492d7ec0db/src/content/templates/template%20books/Games/Games.html#L14)):*

get -> gets
</blockquote></details>

Done.
___
*[`src/content/templates/template books/Games/Games.html` line 17 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7069#-OQ567plDYO9-ljFgkia:-OQ59vOm0F5s-Hgg7xIt:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/e4223afb627fff1121a6aed01d054d492d7ec0db/src/content/templates/template%20books/Games/Games.html#L17)):*

...from this template?
May be worth noting that they are then available to edit with the style dialog, so these definitions are not locked down.
</blockquote></details>

Done.
___
*[`src/content/templates/template books/Games/Games.html` line 128 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7069#-OQ54tuU9QlHkXDGbbBe:-OQ5C5Qn--T2D1LLiLf_:b-577jjh) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/e4223afb627fff1121a6aed01d054d492d7ec0db/src/content/templates/template%20books/Games/Games.html#L128)):*

Here or possibly above we should note the reason: that the game toolbox may add elements to the page which use these styles, and there is no easy way to retrieve them from the game template book at that time if they are not already defined.
</blockquote></details>

Done.

But in my mind, this is more about having the style varieties available.
___
*[`src/content/templates/template books/Games/Games.less` line 624 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7069#-OQ55IDb4c3N-LVVPOIe:-OQ5DbNX0vh-gzBV7UUo:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/e4223afb627fff1121a6aed01d054d492d7ec0db/src/content/templates/template%20books/Games/Games.less#L624)):*

Maybe say a little more along the lines that the main definition is in Games.htm but these properties are defined here so that...is it because the HTML definitions can't just use variables like this for some reason?
</blockquote></details>

Done.

PR #7072, opened 2025-05-12
Fix control frame width bug by removing jquery use (BL-14750)

StephenMcConnel:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7072)

### DIFF
--- a/src/BloomBrowserUI/bookEdit/js/CanvasElementManager.ts
+++ b/src/BloomBrowserUI/bookEdit/js/CanvasElementManager.ts
@@ -385,7 +385,7 @@ export class CanvasElementManager {
                     // by pixel. Don't want to do this if we don't have to, because there could be rounding
                     // errors that would move it very slightly.
                     this.setCanvasElementPosition(
-                        $(wrapperBox as HTMLElement),
+                        wrapperBox,
                         wrapperBox.offsetLeft - container.offsetLeft,
                         wrapperBox.offsetTop - container.offsetTop
                     );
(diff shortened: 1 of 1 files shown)


andrew-polk (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7072)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @StephenMcConnel)

JohnThomson (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7072)* status: all files reviewed, 1 unresolved discussion (waiting on @StephenMcConnel)

___
*[`src/BloomBrowserUI/bookEdit/js/CanvasElementManager.ts` line 5916 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7072#-OQ58-SA5ACGRqhN4sk4:-OQ58-SA5ACGRqhN4sk5:b-tyswiz) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/0138679b6335388349dd85931feaaed6bf15536d/src/BloomBrowserUI/bookEdit/js/CanvasElementManager.ts#L5916)):*
> ```XML
>         // Depending on jquery's .css() method can result in not having the width not actually
>         // set even though canvasElement.css("width") returns a value.  (See BL-14750.)  So we
>         // use the actual HTMLElement .style property, and pass in the actual HTML element.
> ```

This comment feels out of place if it's needed at all. It's down where we're testing for style.width ending in px that we MUST not use JQuery, so you could point this out there. But in general we're moving away from JQuery, and I don't think not using it calls for comment.

JohnThomson (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7072)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @StephenMcConnel)

PR #7098, opened 2025-05-21
BL-14768 fix bottom outline of image getting cut off

nabalone:


JohnThomson (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7098)* status: 0 of 2 files reviewed, 2 unresolved discussions

___
*[`src/BloomBrowserUI/utils/elementUtils.ts` line 53 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7098#-OQopeOHAmypP-NmqXtE:-OQopeOIAnj9bsZfLADB:bdrr38l) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/813e36d6ca5227c66be8a5f7ecce3091f2a323fc/src/BloomBrowserUI/utils/elementUtils.ts#L53)):*
> ```XML
> 
> // .getClientHeight and .getClientWidth return integers, so if rounding errors are a problem we can
> // use this to get the exact fractional values
> ```

Let's improve this. For one thing, the things we're replacing are clientHeight and clientWidth; they don't have 'get'. Let's also save our users a lookup and be explicit about what that is: IIRC, the height and width of the element, including padding but excluding borders, margins, overflow, and scrollbars. You might want to check that last...I'm pretty sure we've come up with something that is like the clientX methods in excluding borders, margins, and overflow, but I'm not sure what getBoundingClientRect does about scroll bars. (The doc of clientHeight has some special notes about the HTML element itself and the <body> element in quirks mode...if you can easily determine whether this method will do the same thing it would be good to say, or you could just note that it may not do exactly what clientX does for those cases.)
If you know of any other exceptions where this does not give the same result as clientHeight and clientWidth, except for being more exact, it would be good to document.
___
*[`src/BloomBrowserUI/utils/elementUtils.ts` line 54 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7098#-OQorAuu9NFSDnpQffSe:-OQorAuu9NFSDnpQffSf:b-lrh0fi) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/813e36d6ca5227c66be8a5f7ecce3091f2a323fc/src/BloomBrowserUI/utils/elementUtils.ts#L54)):*
> ```XML
> // .getClientHeight and .getClientWidth return integers, so if rounding errors are a problem we can
> // use this to get the exact fractional values
> export function getExactClientDimensions(
> ```

Do you think 'Size' would be any less clear than 'Dimensions'? We may end up using this a lot, and where possible I prefer a short, simple word. My first thought on reading Dimensions is what more there might be than height and width.

JohnThomson (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7098)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @nabalone)

PR #7099, opened 2025-05-21
Migrate enterprise-only class to specific data-feature value (BL-14511)

StephenMcConnel:


andrew-polk (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7099)* status: 0 of 5 files reviewed, 1 unresolved discussion

___
*[`src/BloomExe/Book/BookStorage.cs` line 113 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7099#-OQt9vP57IGu3PlOC4bd:-OQt9vP57IGu3PlOC4be:brk0fiv) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/b824fb5fa46d07fec36bd5ffd09417c668c99882/src/BloomExe/Book/BookStorage.cs#L113)):*
> ```Smalltalk
>         void MigrateToLevel7BloomCanvas();
> 
>         void MigrateToDataFeatureAttribute();
> ```

Is there a reason we didn't do this with a level migration?

StephenMcConnel (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7099)* status: 0 of 5 files reviewed, 1 unresolved discussion (waiting on @andrew-polk)

___
*[`src/BloomExe/Book/BookStorage.cs` line 113 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7099#-OQt9vP57IGu3PlOC4bd:-OQtSk9qEjjB6u_CFch2:b-6jfai5) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/b824fb5fa46d07fec36bd5ffd09417c668c99882/src/BloomExe/Book/BookStorage.cs#L113)):*

Is there a reason we didn't do this with a level migration?
</blockquote></details>

I thought about doing a level migration, but decided against it because 1) it has be easily reversible going back at least 2 versions, and what level would you reverse it to? and 2) the DOM change is very trivial, modifying two attributes of the page div.  The former reason is what persuaded me.

JohnThomson (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7099)* status: 0 of 5 files reviewed, 2 unresolved discussions (waiting on @andrew-polk)

___
*[`src/BloomExe/Book/BookStorage.cs` line 113 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7099#-OQt9vP57IGu3PlOC4bd:-OQu9eZl96aKXeRWR9-I:boq9co9) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/b824fb5fa46d07fec36bd5ffd09417c668c99882/src/BloomExe/Book/BookStorage.cs#L113)):*

I thought about doing a level migration, but decided against it because 1) it has be easily reversible going back at least 2 versions, and what level would you reverse it to? and 2) the DOM change is very trivial, modifying two attributes of the page div.  The former reason is what persuaded me.
</blockquote></details>

Just finding all the page divs is probably somewhat costly; xpaths are slow. We already have at least two reversable level-based migrations between 6.2 and 6.0; I handled them simply by doing the reverse ones in the reverse order, so that the 6.0 code knows that if it finds a book at level 7, it does X to get it back to 6, and then Y to get it back to 5. I don't see why it couldn't first do Z to get it back from 8 to 7 (or whatever the current number is).
___
*[a discussion](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7099#-OQuAbju5z9-K9Js60Aa:-OQuAbju5z9-K9Js60Ab:b-z4nsag) (no related file):*
We ought to be adding a FeatureRequirement so that versions of Bloom that don't know how to back-migrate books with data-feature know they can't open them at all.

StephenMcConnel (commented):
Note that the reverse migration in 6.0 / 6.1 that I did earlier needs to be redone since it now uses a level migration and a Feature that has a required version.

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7099#-:-ORHX17u8KV4Jsa76AfT:b-k75lky)* status: 0 of 5 files reviewed, 2 unresolved discussions (waiting on @andrew-polk and @JohnThomson)

___
*[a discussion](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7099#-OQuAbju5z9-K9Js60Aa:-ORHWseE98PqRRpY6X8P:b-896fix) (no related file):*

We ought to be adding a FeatureRequirement so that versions of Bloom that don't know how to back-migrate books with data-feature know they can't open them at all.
</blockquote></details>

Done.
___
*[`src/BloomExe/Book/BookStorage.cs` line 113 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7099#-OQt9vP57IGu3PlOC4bd:-ORHVQjJ7MVD1zJR1iz2:bvxqlwi) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/b824fb5fa46d07fec36bd5ffd09417c668c99882/src/BloomExe/Book/BookStorage.cs#L113)):*

Just finding all the page divs is probably somewhat costly; xpaths are slow. We already have at least two reversable level-based migrations between 6.2 and 6.0; I handled them simply by doing the reverse ones in the reverse order, so that the 6.0 code knows that if it finds a book at level 7, it does X to get it back to 6, and then Y to get it back to 5. I don't see why it couldn't first do Z to get it back from 8 to 7 (or whatever the current number is).
</blockquote></details>

Okay, I've added in what's needed for a level migration.

JohnThomson (changes requested):
> Reviewed 1 of 5 files at r1, 1 of 2 files at r2, 3 of 3 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7099)* status: all files reviewed, 4 unresolved discussions (waiting on @andrew-polk and @StephenMcConnel)

___
*[a discussion](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7099#-OQuAbju5z9-K9Js60Aa:-ORIViQk9z85DTVxfPEz:bkt58ni) (no related file):*

Done.
</blockquote></details>

I think we're getting close. Not sure how much to make of not handling overlay, motion, and music, because I'm not clear how much if anything depends on having data-feature on those pages.
___
*[`src/BloomExe/SubscriptionAndFeatures/FeatureRegistry.cs` line 64 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7099#-ORIKFPS7E0P5VjEQY-f:-ORIKFPT47fTcvJ6X8ij:bny3bxo) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/dfec1f098ebdd147deb82a2e1e9f55d76db7aca4/src/BloomExe/SubscriptionAndFeatures/FeatureRegistry.cs#L64)):*
> ```Smalltalk
>                 Feature = FeatureName.Music,
>                 SubscriptionTier = SubscriptionTier.Pro,
>                 ExistsInPageXPath = "self::div[@data-backgroundaudio and string-length(@data-backgroundaudio)!=0]",
> ```

I'm not sure about this. May not yet be used at all, since old music pages don't have enterprise-only. Nor am I sure exactly which pages we WANT to mark with this feature. For example, if page 2 starts some music, it continues through page 3 and 4, and stops on page 5, are pages 3 and 4 music pages? Is page 5? Only pages 2 and 5 have data-backgroundaudio, but the one on page 5 will have length zero, so will (I think correctly) not be marked.
Of course it's only page 2 that needs to have the attribute removed, so maybe that's the only one that needs the feature?
(Note that we're currently failing to implement or persist "No music" (BL-14846))
___
*[`src/BloomExe/Book/BookStorage.cs` line 4089 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7099#-ORIU0ra7lGUhfoPNlf_:-ORIU0rb9mSNUvNDk4oB:b-uo8utl) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/dfec1f098ebdd147deb82a2e1e9f55d76db7aca4/src/BloomExe/Book/BookStorage.cs#L4089)):*
> ```Smalltalk
>             }
>             // Not sure if this can happen or what to prioritize.
>             featureList.RemoveRange(1, featureList.Count - 1);
> ```

It would be logical to at least prioritize the feature with the highest SubscriptionTier, since if (e.g.) it has something that requires Enterprise, we don't want to let it sneak by just because it also has something that only requires Pro.

StephenMcConnel (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7099)* status: all files reviewed, 4 unresolved discussions (waiting on @andrew-polk and @JohnThomson)

___
*[a discussion](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7099#-OQuAbju5z9-K9Js60Aa:-ORIm4z4CU_2dxPoWEOe:b8pcqw3) (no related file):*

I think we're getting close. Not sure how much to make of not handling overlay, motion, and music, because I'm not clear how much if anything depends on having data-feature on those pages.
</blockquote></details>

Since those didn't used to be marked enterprise-only, I wasn't going to worry about them.  It's one of the rough edges of the changeover that I'm sure sure we know how to smooth out yet, or what needs smoothing out.
___
*[`src/BloomExe/Book/BookStorage.cs` line 4089 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7099#-ORIU0ra7lGUhfoPNlf_:-ORIxR8XDX4RYU3NXtDp:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/dfec1f098ebdd147deb82a2e1e9f55d76db7aca4/src/BloomExe/Book/BookStorage.cs#L4089)):*

It would be logical to at least prioritize the feature with the highest SubscriptionTier, since if (e.g.) it has something that requires Enterprise, we don't want to let it sneak by just because it also has something that only requires Pro.
</blockquote></details>

Done.
___
*[`src/BloomExe/SubscriptionAndFeatures/FeatureRegistry.cs` line 64 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7099#-ORIKFPS7E0P5VjEQY-f:-ORIkVWCDmFxA5cowZKv:b6ba5xc) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/dfec1f098ebdd147deb82a2e1e9f55d76db7aca4/src/BloomExe/SubscriptionAndFeatures/FeatureRegistry.cs#L64)):*

I'm not sure about this. May not yet be used at all, since old music pages don't have enterprise-only. Nor am I sure exactly which pages we WANT to mark with this feature. For example, if page 2 starts some music, it continues through page 3 and 4, and stops on page 5, are pages 3 and 4 music pages? Is page 5? Only pages 2 and 5 have data-backgroundaudio, but the one on page 5 will have length zero, so will (I think correctly) not be marked.
Of course it's only page 2 that needs to have the attribute removed, so maybe that's the only one that needs the feature?
(Note that we're currently failing to implement or persist "No music" (BL-14846))
</blockquote></details>

I had noticed the "No music" wasn't being persisted at all. :-)
This is a bit tricky to think about since it's the only multi-page feature we have.  But since only the first page sets up the necessary data, I think what is here is okay.  If we fix continuing music pages with a marker, that could be added as an OR condition to the xpath, but I don't know what that would be.

JohnThomson (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7099#-:-ORNf51r6mCaRIy1G8Se:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on @andrew-polk)

___
*[a discussion](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7099#-OQuAbju5z9-K9Js60Aa:-ORN2_ib0kL3ePDZnM4j:ba5qwcu) (no related file):*

Since those didn't used to be marked enterprise-only, I wasn't going to worry about them.  It's one of the rough edges of the changeover that I'm sure sure we know how to smooth out yet, or what needs smoothing out.
</blockquote></details>

Turns out we don't need data-feature on them because they can be identified by xpath. But then, I think games could be too...rough edges.

PR #7108, opened 2025-05-28
Fix checkForAnyRecording to return boolean (BL-14850)

JohnThomson:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7108)


nabalone:
Left a comment but I think you can just merge it

PR #7123, opened 2025-05-30
Centralize figuring out which pages not to publish (BL-14587)

JohnThomson:
A single function figures out which pages should not be published either because they have a feature which should be removed at the current subscription tier (this also depends on whether the book is a derivative) or because the have a feature which is not supported in the current publishing medium.
Also cleans up some problems with generating deleted page messages in EpubMaker, and possibly generating messages in the wrong order in other places. (We were depending on Hashset to return insertions in the order they were added.)
WIP because work is still needed on tests.

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7123)

### DIFF
--- a/src/BloomExe/Publish/PublishHelper.cs
+++ b/src/BloomExe/Publish/PublishHelper.cs
@@ -278,7 +278,10 @@ namespace Bloom.Publish
             }
 
             RemovePagesByFeatureSystem(book, dom, pageElts, medium, omittedPages);
-            if (!book.CollectionSettings.Subscription.HaveActiveSubscription)
+            if (
+                !book.CollectionSettings.Subscription.HaveActiveSubscription
+                && !book.BookData.BookIsDerivative()
+            )
                 RemoveEnterpriseOnlyAssets(book);
             RemoveClassesAndAttrsToDisableFeatures(
                 pageElts,
@@ -646,10 +649,10 @@ namespace Bloom.Publish
             ISet<string> messages
         )
         {
-            var disabledRemoveByModifyingFeatures = FeatureStatus.GetDisabledFeatures(
+            var disabledRemoveByModifyingFeatures = FeatureStatus.GetFeaturesToDisableUsingMethod(
                 subscription,
                 forDerivative,
-                PublicationRestrictions.DisabledByModifyingDom
+                PreventionMethod.DisabledByModifyingDom
             );
             foreach (var page in pageElts)
             {
@@ -683,8 +686,9 @@ namespace Bloom.Publish
             }
         }
 
-        // Called if we don't have any subscription level at all (basic tier), removes things we
-        // don't need in that case because pages that use them have been removed.
+        // Called if we don't have any subscription level at all (basic tier) and the book
+        // is an original, removes things we don't need in that case because pages that use
+        // them have been removed.
         // Not really sure what we need here; embedding the simple comprehension quiz JS is
         // probably obsolete in all situations, but there may be some old version of Bloom Player
         // that needs it. But for sure we don't need it in a Basic tier publication which
@@ -692,6 +696,9 @@ namespace Bloom.Publish
         // I don't know why the video placeholder ever needs to be published, but we used
         // to remove it for enterprise publications, so it should still be safe to remove
         // it for Basic tier, at least.
+        // The old code did not distinguish derivatives, but we're moving more in the
+        // direction of allowing most things to be published in derivatives, so I decided
+        // to err on the side of safety by NOT deleting these assets when publishing a derivative.
         private static void RemoveEnterpriseOnlyAssets(Book.Book book)
         {
             RobustFile.Delete(Path.Combine(book.FolderPath, kSimpleComprehensionQuizJs));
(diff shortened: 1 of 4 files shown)


StephenMcConnel (changes requested):
Nothing major, just some questions and comments.

> Reviewed 6 of 13 files at r1, 12 of 12 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7123#-:-ORrMMON73JrG3W2pePr:b9eo19t)* status: all files reviewed, 5 unresolved discussions (waiting on @JohnThomson)

___
*[`src/BloomTests/Publish/PublishHelperTests.cs` line 1086 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7123#-ORrLROF0khBxyXcPRam:-ORrLROF0khBxyXcPRan:bnhrnf9) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/38f2af8269773b73ba07b1d08c5f7979991280c2/src/BloomTests/Publish/PublishHelperTests.cs#L1086)):*
> ```Smalltalk
>             );
>             void checkNotRemoved()
>             {
> ```

Is it really worth defining a method that is called only once immediately after the definition?
I suppose if you want to maintain consistency for the new test code that is okay.
But if these internal methods differ only by the number of pages, then a single external method with a page count argument
could be written instead and shared between tests.
___
*[`src/BloomExe/Publish/PublishHelper.cs` line 282 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7123#-ORqyGv38Xs7T-BNdVxn:-ORqyGv38Xs7T-BNdVxo:b-r751jr) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/38f2af8269773b73ba07b1d08c5f7979991280c2/src/BloomExe/Publish/PublishHelper.cs#L282)):*
> ```Smalltalk
>             RemovePagesByFeatureSystem(book, dom, pageElts, medium, omittedPages);
>             if (!book.CollectionSettings.Subscription.HaveActiveSubscription)
>                 RemoveEnterpriseOnlyAssets(book);
> ```

Should RemoveEnterpriseOnlyAssets(book) depend on whether or not the book is a derivative?  (or is that handled internally to the method)
___
*[`src/BloomExe/SubscriptionAndFeatures/FeatureStatus.cs` line 24 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7123#-ORrHt-aCpgqEcIwFdUq:-ORrHt-aCpgqEcIwFdUr:b-99x6i6) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/38f2af8269773b73ba07b1d08c5f7979991280c2/src/BloomExe/SubscriptionAndFeatures/FeatureStatus.cs#L24)):*
> ```Smalltalk
>     }
> 
>     public enum PublicationRestrictions
> ```

This name isn't new, but having a plural name here is misleading unless the enum values can be ORed together.  This leads to variables having a plural name but only a single value.

Adding the comments below is a valuable addition.  :-)  :+1: 
___
*[`src/BloomExe/SubscriptionAndFeatures/FeatureStatus.cs` line 143 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7123#-ORrIbolD4U6fp2PP5iJ:-ORrIbon8a-H7zLoFiWJ:b9zms4v) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/38f2af8269773b73ba07b1d08c5f7979991280c2/src/BloomExe/SubscriptionAndFeatures/FeatureStatus.cs#L143)):*
> ```Smalltalk
>             Subscription subscription,
>             bool forDerivative,
>             PublicationRestrictions restrictions
> ```

As noted before, the plural name restrictions is misleading because it's only a single value.
___
*[`src/BloomExe/SubscriptionAndFeatures/FeatureStatus.cs` line 151 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7123#-ORr9RzkEe1lUlfBjqVU:-ORr9RzkEe1lUlfBjqVV:b-ulrt4b) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/38f2af8269773b73ba07b1d08c5f7979991280c2/src/BloomExe/SubscriptionAndFeatures/FeatureStatus.cs#L151)):*
> ```Smalltalk
>                 return !featureStatus.Enabled
>                     && (
>                         forDerivative
> ```

This expression should perhaps be extracted to a variable for clarity.
```
    var featureStatus = GetFeatureStatus(subscription, feature);
    var hasMatchingRestriction = forDerivative
        ? feature.RestrictionInDerivativeBooks == restrictions
        : feature.RestrictionInOriginalBooks == restrictions;
   return !feature.Enabled && hasMatchingRestriction;
```
The logic follows what is said in the header comment, but the method name does not really communicate what it returns.  Maybe a longer name like `GetDisabledFeaturesWithThePublicationRestrictions`?  (although that name is getting rather unwieldy...)

StephenMcConnel (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7123)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @JohnThomson)

PR #7124, opened 2025-05-30
BL-14819 Playground Template Book

hatton:
A template book called "playground" that gives enterprise access.
✔️ watermark
✔️ can't publish if you've used...
✔️ most features enabled
✔️ new pages  get the watermark
❌ user can publish if we don't detect enterprise things. Less confusing to not allow publishing at all.

Improved the styling of template readmes


JohnThomson (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7124)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @hatton)

PR #7125, opened 2025-06-02
Put recapitalized language tag back together manually (BL-14597)

StephenMcConnel:
Using the SIL.WritingSystem.IetfLanguageTag function could lead to pieces of the input tag being removed because they are implied by default, but that can lead to unexpected problems.  Recapitalization works okay because tags are somewhat case insensitive, but removing subtags is a major change that makes the tag unrecognizable.

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7125)

### DIFF
--- a/src/BloomExe/Book/Book.cs
+++ b/src/BloomExe/Book/Book.cs
@@ -1206,6 +1206,7 @@ namespace Bloom.Book
                         if (goodDiv == null)
                         {
                             badDiv.SetAttribute("lang", tag);
+                            continue;   // we don't want to delete a div that has been fixed.  BL-14597
                         }
                         else if (
                             string.IsNullOrWhiteSpace(goodDiv.InnerText)
--- a/src/BloomExe/Utils/MiscUtils.cs
+++ b/src/BloomExe/Utils/MiscUtils.cs
@@ -777,6 +777,10 @@ namespace Bloom.Utils
         /// ISO 639-2 code), Yyyy is the script code, and ZZ is the country code.  ZZ and Yyyy are both optional.
         /// We need to ensure the pieces are properly capitalized.  (BL-14038)
         /// </summary>
+        /// <remarks>
+        /// This method doesn't try to do any normalization of the input tag apart from capitalization.
+        /// See BL-14597 for an example of a problem that has occurred when we tried to do more.
+        /// </remarks>
         /// <returns>possibly recapitalized language tag</returns>
         public static string NormalizeLanguageTagCapitalization(string tag)
         {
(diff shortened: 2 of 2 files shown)


nabalone (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7125)* status: 0 of 1 files reviewed, 1 unresolved discussion

___
*[`src/BloomExe/Utils/MiscUtils.cs` line 817 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7125#-ORm4s708r9F3DXXeshu:-ORm4s708r9F3DXXeshv:bom8fvh) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/77485e0bf95eac65bd07a42ef8a5ef438b0738ff/src/BloomExe/Utils/MiscUtils.cs#L817)):*
> ```Smalltalk
>                         script.Substring(0, 1).ToUpperInvariant()
>                         + script.Substring(1).ToLowerInvariant();
>                     if (bldr.Length > 0)
> ```

There can't be a language tag with no language. Would TryGetParts ever give us a null or empty language? If so, can we check for that case just once at the beginning and treat it as the failed to parse scenario?

StephenMcConnel (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7125)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @nabalone)

___
*[`src/BloomExe/Utils/MiscUtils.cs` line 817 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7125#-ORm4s708r9F3DXXeshu:-ORquuVO4GPeB48Mi2Py:b-zdhwu3) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/77485e0bf95eac65bd07a42ef8a5ef438b0738ff/src/BloomExe/Utils/MiscUtils.cs#L817)):*

There can't be a language tag with no language. Would TryGetParts ever give us a null or empty language? If so, can we check for that case just once at the beginning and treat it as the failed to parse scenario?
</blockquote></details>

I looked at the code for TryGetParts and it won't return a null language if it succeeds, and I'm pretty sure it won't return an empty string either.  I added some commentary and a Debug.Assert to verify this.

StephenMcConnel (commented):
I think I found where too much was being deleted, and fixed it.  Bloom no longer crashes even with the old code creating a different tag.  Publishing to BloomPub with an empty book results in an empty front cover page and nothing else: not even any other xmatter pages.  Going back to edit the book restores all of the xmatter pages.  But I'm leaving in the changes to NormalizeLanguageTagCapitalization because it doesn't claim to do any other normalization.

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7125#-:-ORrjZuF0lCYbdrwtloy:bwm8rye)* status: all files reviewed, 2 unresolved discussions (waiting on @JohnThomson and @nabalone)

___
*[`src/BloomExe/Utils/MiscUtils.cs` line 807 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7125#-ORrDbnv4OGJt-zv5R8h:-ORriEzBAfAgk6a2Wd74:b-hjon8w) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/f0ef9e9a4d4f0286adf5d23800e6145d33933e83/src/BloomExe/Utils/MiscUtils.cs#L807)):*

And they have to be Ascii, right, so the first 'letter' couldn't be a surrogate pair?
</blockquote></details>

That's right.  Language codes, script codes, and region codes all come from the ASCII character set.  It's possible that all variant and private codes are restricted to the same, although I haven't looked at the standard in a long time.

JohnThomson (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7125#-:-ORrv1f59uEFoJoyQ-6_:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on @nabalone)

PR #7126, opened 2025-06-02
BL-14791 toolbox subscription badges and overlays

nabalone:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7126)

### DIFF
--- a/src/BloomBrowserUI/bookEdit/toolbox/motion/motionTool.tsx
+++ b/src/BloomBrowserUI/bookEdit/toolbox/motion/motionTool.tsx
@@ -18,6 +18,8 @@ import {
     kBloomCanvasClass
 } from "../../js/bloomImages";
 import { css } from "@emotion/react";
+import { kMotionToolId } from "../toolIds";
+import { RequiresSubscriptionOverlayWrapper } from "../../../react_components/requiresSubscription";
 
 // The toolbox is included in the list of tools because of this line of code
 // in tooboxBootstrap.ts:
@@ -35,7 +37,6 @@ export class MotionTool extends ToolboxToolReactAdaptor {
 
     public makeRootElement(): HTMLDivElement {
         const root = document.createElement("div");
-        root.setAttribute("class", "ui-motionBody");
         this.rootControl = (ReactDOM.render(
             <MotionControl
                 onPreviewClick={() => this.toggleMotionPreviewPlaying()}
@@ -284,8 +285,10 @@ export class MotionTool extends ToolboxToolReactAdaptor {
     }
 
     public id(): string {
-        return "motion";
+        return kMotionToolId;
     }
+
+    public featureName? = kMotionToolId;
 
     private getBloomCanvasToAnimate(): HTMLElement | null {
         const page = this.getPage();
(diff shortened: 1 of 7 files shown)


nabalone (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7126)* status: 0 of 8 files reviewed, 5 unresolved discussions

___
*[`src/BloomBrowserUI/bloomUI.less` line 186 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7126#-ORmUpNn-W5xbB2EpA4w:-ORmUpNn-W5xbB2EpA4x:bpb3wvv) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/e6479cba760085478bdfe915b1c2e837f0791992/src/BloomBrowserUI/bloomUI.less#L186)):*
> ```Less
> 
> // Moved from toolbox.less to here so it can be used by templatePagePreview
> [data-feature]::after {
> ```

We don't want the purple dots on the toolbox checkboxes anymore. The purple dots on the page templates are added by different code
___
*[`src/BloomBrowserUI/bookEdit/toolbox/toolbox.less` line 218 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7126#-ORmUzKmDqV-PCI0-jD9:-ORmUzKmDqV-PCI0-jDA:b8qovy4) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/20c20792f2074c96bbe1641c00b9d34cf4fa6e9f/src/BloomBrowserUI/bookEdit/toolbox/toolbox.less#L218)):*
> ```Less
> }
> 
> // Only show the subscription badge for tools with data-feature
> ```

Is there a better way to do this?
___
*[`src/BloomBrowserUI/bookEdit/toolbox/motion/motionTool.tsx` line 291 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7126#-ORmVlrz5CXWyT6byG7I:-ORmVlrz5CXWyT6byG7J:b-bmm1t6) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/20c20792f2074c96bbe1641c00b9d34cf4fa6e9f/src/BloomBrowserUI/bookEdit/toolbox/motion/motionTool.tsx#L291)):*
> ```XML
>     }
> 
>     public static featureName?: string | undefined = kMotionToolId;
> ```

I wanted it to be static so I can reuse it on the SubscriptionOverlayWrapper in MotionControl below, where we don't have a MotionTool instance. That way we have one source of truth for both the subscription badge and the subscription overlay. But is it too silly to have both a static and a non-static version like this? Is there a better way?

Also in MusicTool and OverlayTool
___
*[`src/BloomBrowserUI/bookEdit/toolbox/motion/motionTool.tsx` line 38 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7126#-ORmVJeaCtmcqmvAvfo3:-ORmVJeaCtmcqmvAvfo4:byug2tu) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/e6479cba760085478bdfe915b1c2e837f0791992/src/BloomBrowserUI/bookEdit/toolbox/motion/motionTool.tsx#L38)):*
> ```XML
>     public makeRootElement(): HTMLDivElement {
>         const root = document.createElement("div");
>         root.setAttribute("class", "ui-motionBody");
> ```

This sets the class on a parent element which makes it apply to the subscription overlay, which we don't want. We are already setting ui-motionBody on the appropriate child element elsewhere
___
*[`src/BloomBrowserUI/bookEdit/toolbox/overlay/overlayTool.tsx` line 656 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7126#-ORmXbdt4I2-H8pmCaVb:-ORmXbdt4I2-H8pmCaVc:bk4zhgm) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/20c20792f2074c96bbe1641c00b9d34cf4fa6e9f/src/BloomBrowserUI/bookEdit/toolbox/overlay/overlayTool.tsx#L656)):*
> ```XML
>         <div id="overlayToolControls">
>             <RequiresSubscriptionOverlayWrapper
>                 featureName={OverlayTool.featureName as string}
> ```

Here we could use theOneOverlayTool.featureName and then get rid of the static version, but I did this to be consistent with MusicTool and MotionTool, see comment above

nabalone (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7126)* status: 2 of 8 files reviewed, 9 unresolved discussions (waiting on @JohnThomson)

___
*[`src/BloomBrowserUI/bookEdit/toolbox/toolbox.less` line 156 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7126#-ORqXhOu5hQIwYACdHBD:-ORrd6p057RRvq1qSbLF:b-rqyeyr) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/20c20792f2074c96bbe1641c00b9d34cf4fa6e9f/src/BloomBrowserUI/bookEdit/toolbox/toolbox.less#L156)):*

Puzzling. One of the items can grow, so I think all the space will be taken up, so what does centering accomplish?
</blockquote></details>

align-items works on the cross-axis, in this case to put the items at similar vertical placements

___
*[`src/BloomBrowserUI/bookEdit/toolbox/toolbox.less` line 218 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7126#-ORmUzKmDqV-PCI0-jD9:-ORrnBXt3dXyrq8Khrwy:blnje7o) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/20c20792f2074c96bbe1641c00b9d34cf4fa6e9f/src/BloomBrowserUI/bookEdit/toolbox/toolbox.less#L218)):*

I'm probably missing something, but why did we add a subscription badge at all to any tool that doesn't have the feature? The presence/absence of data-feature is not going to change, so badges on features without it will never be seen, so why have the badge?
Assuming there's a reason for the badge to exist when we don't want to show it, I think this is reasonable.
</blockquote></details>

You're right, I can't think of a reason
___
*[`src/BloomBrowserUI/bookEdit/toolbox/toolbox.ts` line 953 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7126#-ORqZYtN3yX1lkEBHA0Y:-ORrn98H2SE9GQKUiiYT:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/20c20792f2074c96bbe1641c00b9d34cf4fa6e9f/src/BloomBrowserUI/bookEdit/toolbox/toolbox.ts#L953)):*

I think here I would just make the badge span be inside ${rectTool.featureName ? "<span.../span>" : ""}
</blockquote></details>

Done.
___
*[`src/BloomBrowserUI/bookEdit/toolbox/toolbox.ts` line 1349 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7126#-ORqaANF00GqagCrPRcp:-ORrnzwX9TttZ6Q2KXVi:bjg14m6) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/20c20792f2074c96bbe1641c00b9d34cf4fa6e9f/src/BloomBrowserUI/bookEdit/toolbox/toolbox.ts#L1349)):*

Do we need to build the list of promises and await them? I don't think any callers need to wait until all the badges have titles, so can't we just let it complete in its own time?
Of course we'd be glad we did if we ever NEED to await the completion of this function, but I think YAGNI applies. Then again, the code is written, so I don't see any harm in keeping it. Up to you.
</blockquote></details>

Huh, yeah, that was a relic of a previous way I had implemented this. Since it's not affecting anything now, I guess I'll leave it
___
*[`src/BloomBrowserUI/bookEdit/toolbox/motion/motionTool.tsx` line 291 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7126#-ORmVlrz5CXWyT6byG7I:-ORrpQMUC56CgitpVn3X:bhc8tzr) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/20c20792f2074c96bbe1641c00b9d34cf4fa6e9f/src/BloomBrowserUI/bookEdit/toolbox/motion/motionTool.tsx#L291)):*

My first thought it to just use MotionTool.featureName wherever you need it. But probably there's a reason featureName needs to be an instance property, e.g., so some code can find the featureName of any tool object it happens to have. In that case I think what you have is OK.
There might be some advantage in defining the constant independently (kMotionToolFeatureName) in another file so that MotionControl (which I think is in the content page bundle?) doesn't have to import the whole MotionTool (which belongs in the toolbox). But we have so many cross-links that it's probably hopeless to keep this code out of the common bundle.
</blockquote></details>

We decided to use the id

nabalone (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7126)* status: 7 of 8 files reviewed, 1 unresolved discussion (waiting on @JohnThomson)

___
*[`src/BloomBrowserUI/bookEdit/toolbox/toolbox.less` line 218 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7126#-ORmUzKmDqV-PCI0-jD9:-ORwcCJI3iILzeML0XWv:bb6un1j) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/20c20792f2074c96bbe1641c00b9d34cf4fa6e9f/src/BloomBrowserUI/bookEdit/toolbox/toolbox.less#L218)):*

I think you removed all the badges where the aren't wanted, so can you remove these new rules too?
</blockquote></details>

Done
___
*[`src/BloomBrowserUI/bookEdit/toolbox/settings/Settings.pug` line 26 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7126#-ORqeDk8ENwv2apgWe-u:-ORwcHDX5TE0hohVeWXN:bb6un1j) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/20c20792f2074c96bbe1641c00b9d34cf4fa6e9f/src/BloomBrowserUI/bookEdit/toolbox/settings/Settings.pug#L26)):*

Can we just leave this out of tools where we don't want it, instead of including it and then writing CSS to hide it?
</blockquote></details>

Done

JohnThomson (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7126)* status: 7 of 8 files reviewed, 1 unresolved discussion

PR #7130, opened 2025-06-04
Fix a broken unit test (BL-14597)

StephenMcConnel:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7130)

### DIFF
--- a/src/BloomExe/Utils/MiscUtils.cs
+++ b/src/BloomExe/Utils/MiscUtils.cs
@@ -784,6 +784,9 @@ namespace Bloom.Utils
         /// <returns>possibly recapitalized language tag</returns>
         public static string NormalizeLanguageTagCapitalization(string tag)
         {
+            // Degenerate (albeit valid) case that may only ever occur in unit tests.
+            if (tag.StartsWith("x-"))
+                return tag;
             // The IetfLanguageTag parser appears to be case insensitive.
             // We need to ensure the pieces are properly capitalized.
             if (


andrew-polk (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7130)* status: all files reviewed, 1 unresolved discussion (waiting on @StephenMcConnel)

___
*[`src/BloomExe/Utils/MiscUtils.cs` line 787 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7130#-ORvd_dRBnfrvC1VuH2f:-ORvd_dRBnfrvC1VuH2g:bhchw1e) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/b5781fd661dbc357cdbb0cec2df5f8dd459a6c95/src/BloomExe/Utils/MiscUtils.cs#L787)):*
> ```Smalltalk
>         public static string NormalizeLanguageTagCapitalization(string tag)
>         {
>             // Degenerate case that probably only ever occurs in unit tests.
> ```

As I understand it, x- represents a "private-use" tag and should be treated as at least somewhat legitimate.
I don't know if that means we should handle the code differently or just the comment.

andrew-polk (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7130)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @StephenMcConnel)

PR #7147, opened 2025-06-06
BL-14659 disable the Choose Layout menu item for pages with activities

nabalone:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7147)

### DIFF
--- a/src/BloomBrowserUI/bookEdit/js/origami.ts
+++ b/src/BloomBrowserUI/bookEdit/js/origami.ts
@@ -333,6 +333,8 @@ function getSplitPaneComponentInner() {
 function getAbovePageControlContainer(): JQuery {
     // for dragActivities we don't want the origami control, but we still make the
     // wrapper so that the dragActivity can put a different control in it.
+    // Note: We also have to disable the Choose Different layout option in
+    // the right click menu, in PageListView.cs
     if (
         document
             .getElementsByClassName("bloom-page")[0]
--- a/src/BloomExe/Edit/PageListView.cs
+++ b/src/BloomExe/Edit/PageListView.cs
@@ -137,7 +137,14 @@ namespace Bloom.Edit
                         "EditTab.ChooseLayoutButton",
                         "Choose Different Layout"
                     ),
-                    EnableFunction = (page) => page != null && !page.Required,
+                    // We don't want to allow layout changes for game pages (except Widget pages).
+                    // Note: we also have to disable the Change Layout switch, in origami.ts
+                    EnableFunction = (page) =>
+                        page != null
+                        && !page.Required
+                        && !page.GetDivNodeForThisPage()
+                            .GetAttribute("data-tool-id")
+                            .Equals("game"),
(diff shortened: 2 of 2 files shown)


andrew-polk (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7147)* status: all files reviewed, 1 unresolved discussion (waiting on @nabalone)

___
*[`src/BloomExe/Edit/PageListView.cs` line 143 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7147#-OS4wk290zYVxlCbEXyQ:-OS4wk2A7XwXiNS5qOGp:b-bapshw) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/1a65e8a20f694e694f547b8dd10f50e4b47604a7/src/BloomExe/Edit/PageListView.cs#L143)):*
> ```Smalltalk
>                         page != null
>                         && !page.Required
>                         && !page.GetDivNodeForThisPage().HasAttribute("data-activity"),
> ```

Rightly or wrongly, we currently allow change layout in all versions for widgets.
In fact, we allow the origami (Change Layout) control, and in Change Layout mode, the user can add a widget.

If I understand correctly, this will turn off this item for widget pages, too. 
Unless I'm missing something, I think we want the same logic to turn this off as to turn off the "Change Layout" switch.
That's in origami.ts and is currently `getAttribute("data-tool-id") === "game"`.
May not be worth trying to share any code, but at least a comment in both places is probably warranted.

Let me know if I'm missing something.

andrew-polk:
*[`src/BloomExe/Edit/PageListView.cs` line 143 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7147#-OS4wk290zYVxlCbEXyQ:-OS4xtNPCHNUQdQV29PH:bw25zt2) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/1a65e8a20f694e694f547b8dd10f50e4b47604a7/src/BloomExe/Edit/PageListView.cs#L143)):*

Rightly or wrongly, we currently allow change layout in all versions for widgets.
In fact, we allow the origami (Change Layout) control, and in Change Layout mode, the user can add a widget.

If I understand correctly, this will turn off this item for widget pages, too. 
Unless I'm missing something, I think we want the same logic to turn this off as to turn off the "Change Layout" switch.
That's in origami.ts and is currently `getAttribute("data-tool-id") === "game"`.
May not be worth trying to share any code, but at least a comment in both places is probably warranted.

Let me know if I'm missing something.
</blockquote></details>

Hm. I see now that your testing note on the card says it should still work for widget pages, so I guess I *am* missing something.

So I guess we at least need some code comments.

andrew-polk (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7147)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @nabalone)

PR #7161, opened 2025-06-11
BL-14444 improve error message for unsupported full bleed paper sizes

nabalone:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7161)

### DIFF
--- a/src/BloomExe/Publish/PDF/MakePdfUsingExternalPdfMakerProgram.cs
+++ b/src/BloomExe/Publish/PDF/MakePdfUsingExternalPdfMakerProgram.cs
@@ -373,9 +373,11 @@ namespace Bloom.Publish.PDF
                     width = Size6x9PortraitWidth + bleedExtra;
                     break;
                 default:
-                    throw new ArgumentException(
+                    var exception = new ArgumentException(
                         "Full bleed printing of paper sizes other than A5, A4, A3, USComic, and Size6x9 is not yet implemented"
                     );
+                    exception.Data["argument"] = "specs.PaperSizeName";
+                    throw exception;
             }
 
             if (specs.Landscape)
--- a/src/BloomExe/Publish/PublishModel.cs
+++ b/src/BloomExe/Publish/PublishModel.cs
@@ -232,6 +232,15 @@ namespace Bloom.Publish
                 );
                 BloomMessageBox.ShowWarning(msg);
             }
+            else if (
+                error is ArgumentException
+                && error.Data.Contains("argument")
+                && (string)error.Data["argument"] == "specs.PaperSizeName"
+            )
+            {
+                // Full bleed printing for this paper size is not supported. Notify user but don't let them report.
+                BloomMessageBox.ShowWarning(error.Message);
+            }
             else // for others, just give a generic message and include the original exception in the message
             {
                 ErrorReport.NotifyUserOfProblem(


StephenMcConnel (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7161)* status: all files reviewed, 2 unresolved discussions (waiting on @nabalone)

___
*[`src/BloomExe/Publish/PDF/MakePdfUsingExternalPdfMakerProgram.cs` line 379 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7161#-OSUz-5s5AqFYQlFAG2Z:-OSV47fb1J7nIn6Ecb4R:b-ufrd6j) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/1c433d130ffa1f4f473ad8db0d7fc6b258922788/src/BloomExe/Publish/PDF/MakePdfUsingExternalPdfMakerProgram.cs#L379)):*

Two points about the message that gets displayed:
1) I think we would want this message to be localized (either in BloomMediumPriority.xlf or BloomLowPriority.xlf)
2) We don't want the parameter name to be displayed: this is meaningless to users.
Code something like this might work:  (details may not be quite exact...)
</blockquote></details>

You can double check with Andrew about whether this message should be localized.  But I still maintain that displaying the parameter part of the message would be more confusing than helpful for users.

andrew-polk:
*[`src/BloomExe/Publish/PDF/MakePdfUsingExternalPdfMakerProgram.cs` line 379 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7161#-OSUz-5s5AqFYQlFAG2Z:-OSVkJZn8NX8Ua0rxCFx:b-rygvyk) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/1c433d130ffa1f4f473ad8db0d7fc6b258922788/src/BloomExe/Publish/PDF/MakePdfUsingExternalPdfMakerProgram.cs#L379)):*

You can double check with Andrew about whether this message should be localized.  But I still maintain that displaying the parameter part of the message would be more confusing than helpful for users.
</blockquote></details>

Would be low priority if we add it. 
Given that this has not yet been localized, I would be fine leaving it that way. Especially given it is a an error message for an advanced feature.

nabalone (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7161)* status: 0 of 2 files reviewed, 2 unresolved discussions (waiting on @StephenMcConnel)

___
*[`src/BloomExe/Publish/PDF/MakePdfUsingExternalPdfMakerProgram.cs` line 379 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7161#-OSUz-5s5AqFYQlFAG2Z:-OSVyLal2l__OTD8LCyY:b-nkd5dx) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/1c433d130ffa1f4f473ad8db0d7fc6b258922788/src/BloomExe/Publish/PDF/MakePdfUsingExternalPdfMakerProgram.cs#L379)):*

Would be low priority if we add it. 
Given that this has not yet been localized, I would be fine leaving it that way. Especially given it is a an error message for an advanced feature.
</blockquote></details>

Good point, I should have noticed the parameter name was showing

StephenMcConnel (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7161)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @nabalone)

___
*[`src/BloomExe/Publish/PublishModel.cs` line 242 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7161#-OSV2LES8dwn2tVGj4h4:-OSW0reU69twsMDxqWlQ:bb6ywmx) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/1c433d130ffa1f4f473ad8db0d7fc6b258922788/src/BloomExe/Publish/PublishModel.cs#L242)):*

See comment below for suggested change.
</blockquote></details>

LGTM

PR #7167, opened 2025-06-12
Add book template stock images to those not credited (BL-14829)

StephenMcConnel:
Also simplify the display of the Pixabay license to not include the URL.

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7167)


andrew-polk (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7167)* status: all files reviewed, 2 unresolved discussions (waiting on @StephenMcConnel)

___
*[`src/BloomExe/web/controllers/ImageApi.cs` line 53 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7167#-OStxn-32Y4_GZMN0Q-p:-OStxn-32Y4_GZMN0Q-q:b-xne3b6) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/d81265698b0ee957b8177e66f349fd8c97ef4c98/src/BloomExe/web/controllers/ImageApi.cs#L53)):*
> ```Smalltalk
>             }
>             var templateDirectory = BloomFileLocator.GetBrowserDirectory("templates");
>             foreach (var templateDir in Directory.GetDirectories(Path.Combine(templateDirectory, "template books")))
> ```

I'm hesitant about this.
Unless I'm missing something, I see from searching that this will find quite a lot of thumbnail svgs.
Probably, users won't have real images named with those names, but I'm also not sure it is worth the overhead of searching and adding all those to the imageFiles set.
Would it be better/cleaner to just make a list of filenames like we did for the don't-change-metadata change? Maybe we even consider using the same list until it diverges?
___
*[`src/BloomExe/web/controllers/ImageApi.cs` line 134 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7167#-OStyWV87Fpn1TzRK5AG:-OStyWV87Fpn1TzRK5AH:b-jf1er9) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/d81265698b0ee957b8177e66f349fd8c97ef4c98/src/BloomExe/web/controllers/ImageApi.cs#L134)):*
> ```Smalltalk
>                 {
>                     // This common source of images has an overly verbose license. We don't need the URL portion. (BL-14829)
>                     credit = credit.Replace("Pixabay license (https://pixabay.com/service/license-summary/)", "Pixabay license");
> ```

Sorry this wasn't clear in the card, but I'm pretty sure what JH meant was he wanted to remove the url from Games.html.

andrew-polk (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7167)* status: all files reviewed, 2 unresolved discussions (waiting on @StephenMcConnel)

___
*[`src/BloomExe/web/controllers/ImageApi.cs` line 53 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7167#-OStxn-32Y4_GZMN0Q-p:-OSurmJ_BxTvLG7QlTju:bs30j9a) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/d81265698b0ee957b8177e66f349fd8c97ef4c98/src/BloomExe/web/controllers/ImageApi.cs#L53)):*

I think we want the correct and wrong images to give credit since they're more substantial and show the image controls to edit or replace them.  I've changed to using a fixed list of files from the Games folder.  Almost all of the other images found in the main template book folders were placeholder, thumbnail, and license images which would be discarded anyway.
</blockquote></details>

Sorry; I really mangled my understanding of this.
I should have looked at the card more closely. I guess I was confusing the two changes.

... now that I think I have my head more clear...

Since these are all elements we add ourselves and the user can't change, can we use a class name to label images we don't want to paste credits for?
Then we could add another check in `GetWhichImagesAreUsedOnWhichPages` similar to
```
                if (IsImgNotInAPage(img))
                    continue;
                if (IsImgInsideBrandingElement(img))
                    continue;
```
where we look for the class name on the img (or probably better, on the image container?).

If I'm still confused, we should probably discuss.
___
*[`src/BloomExe/web/controllers/ImageApi.cs` line 134 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7167#-OStyWV87Fpn1TzRK5AG:-OSuvVzSA3086AXI18hu:bg9e51a) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/d81265698b0ee957b8177e66f349fd8c97ef4c98/src/BloomExe/web/controllers/ImageApi.cs#L134)):*

The URL is embedded in the image metadata, not just in the HTML attributes.  I still think this fix is needed from my reading of JH's comment in the issue.
</blockquote></details>

Sorry; I forgot the metadata was saved in the images.
However, I'm the one who added it there, so I think the ideal is to modify it there as well as change Games.html.
We do have the issue of existing images/books, but in my opinion, it isn't worth adding this replace code for those.

Open to pushback, though.

StephenMcConnel (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7167)* status: all files reviewed, 2 unresolved discussions (waiting on @andrew-polk)

___
*[`src/BloomExe/web/controllers/ImageApi.cs` line 53 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7167#-OStxn-32Y4_GZMN0Q-p:-OT-2cDg6YbCyN1uGLLB:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/d81265698b0ee957b8177e66f349fd8c97ef4c98/src/BloomExe/web/controllers/ImageApi.cs#L53)):*

Sorry; I really mangled my understanding of this.
I should have looked at the card more closely. I guess I was confusing the two changes.

... now that I think I have my head more clear...

Since these are all elements we add ourselves and the user can't change, can we use a class name to label images we don't want to paste credits for?
Then we could add another check in `GetWhichImagesAreUsedOnWhichPages` similar to
```
                if (IsImgNotInAPage(img))
                    continue;
                if (IsImgInsideBrandingElement(img))
                    continue;
```
where we look for the class name on the img (or probably better, on the image container?).

If I'm still confused, we should probably discuss.
</blockquote></details>

Done.
___
*[`src/BloomExe/web/controllers/ImageApi.cs` line 134 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7167#-OStyWV87Fpn1TzRK5AG:-OSv5CK_CK9VkuZmbpol:b-yfsa08) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/d81265698b0ee957b8177e66f349fd8c97ef4c98/src/BloomExe/web/controllers/ImageApi.cs#L134)):*

Sorry; I forgot the metadata was saved in the images.
However, I'm the one who added it there, so I think the ideal is to modify it there as well as change Games.html.
We do have the issue of existing images/books, but in my opinion, it isn't worth adding this replace code for those.

Open to pushback, though.
</blockquote></details>

Okay, if we can fiddle with the embedded license, I can remove the URL from there instead.  The HTML just says "Custom License" -- it doesn't have the URL or even say Pixabay.

PR #7169, opened 2025-06-12
BL-14875 TBT playback order in games

nabalone:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7169)


andrew-polk:
This was already check picked as part of a bunch of commits.
I ended up taking the shuffle one as well because it seemed safer than trying to do the cherry pick without it.

PR #7178, opened 2025-06-16
Require Pro subscription for Import Audio button (BL-14900)

StephenMcConnel:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7178)

### DIFF
--- a/src/BloomBrowserUI/bookEdit/toolbox/talkingBook/audioRecording.ts
+++ b/src/BloomBrowserUI/bookEdit/toolbox/talkingBook/audioRecording.ts
@@ -69,6 +69,10 @@ import {
     kCanvasElementClass
 } from "../overlay/canvasElementUtils";
 import { RecordingMode } from "./recordingMode";
+import {
+    FeatureStatus,
+    getFeatureStatusAsync
+} from "../../../react_components/featureStatus";
 
 enum Status {
     Disabled, // Can't use button now (e.g., Play when there is no recording)
@@ -188,6 +192,7 @@ export default class AudioRecording implements IAudioRecorder {
     private previousRecordMode: RecordingMode;
     private haveAudio: boolean;
     private inShowPlaybackOrderMode: boolean = false;
+    private wholeTextBoxAudioFeatureStatus: FeatureStatus | undefined;
 
     // Corresponds to the collection default recording mode which we would theoretically get via async call to C# API
     // This is rather annoying because then we have async calls being introduced all over the place.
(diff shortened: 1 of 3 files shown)


andrew-polk (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7178)* status: all files reviewed, 1 unresolved discussion (waiting on @StephenMcConnel)

___
*[`src/BloomExe/SubscriptionAndFeatures/FeatureRegistry.cs` line 22 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7178#-OSv5v8cBGipBxphbyLN:-OSv5v8cBGipBxphbyLO:bg42x2q) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/4e74cbfad16f56317b4603b250f7fd3fea9e2a1c/src/BloomExe/SubscriptionAndFeatures/FeatureRegistry.cs#L22)):*
> ```Smalltalk
> 
>         WholeTextBoxAudio,
>         ImportAudioRecording,
> ```

I was thinking that this would not be a new feature but rather tie into the WholeTextBoxAudio feature.
In fact, I would think we don't need a separate UI affordance but rather guarantee that if the Whole Text Box Audio radio button is disabled then the import button is also disabled.

But I have a question about the logic for Colin in the card which is relevant...

StephenMcConnel (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7178)* status: 2 of 3 files reviewed, 1 unresolved discussion (waiting on @andrew-polk)

___
*[`src/BloomBrowserUI/bookEdit/toolbox/talkingBook/talkingBookAdvancedSection.tsx` line 44 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7178#-OT9HBO00EWEN3S4TvWM:-OT9JqUN938UqsorUcU6:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/123cd4351e20abfd2d1aa0ae7146437e03ff4f0f/src/BloomBrowserUI/bookEdit/toolbox/talkingBook/talkingBookAdvancedSection.tsx#L44)):*

I think it is safer to have an undefined featureStatus mean disabled?

If you agree, I think you can just do 
`wholeTextBoxAudioFeatureStatus?.enabled`
</blockquote></details>

Done.

andrew-polk (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7178)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @StephenMcConnel)

PR #7194, opened 2025-06-18
Don't let Save() remove xmatter attribute changes in publish (BL-14907)

JohnThomson:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7194)

### DIFF
--- a/src/BloomExe/Book/Book.cs
+++ b/src/BloomExe/Book/Book.cs
@@ -4709,7 +4709,7 @@ namespace Bloom.Book
             }
         }
 
-        public void Save()
+        public void Save(bool preventUpdateFromDataDiv = false)
         {
             // If you add something here, consider whether it is needed in SaveForPageChanged().
             // I believe all the things currently here before the actual Save are not needed
@@ -4733,10 +4733,17 @@ namespace Bloom.Book
 
             RemoveObsoleteSoundAttributes(OurHtmlDom);
             RemoveVideoWarnings();
-            // Note that at this point _bookData has already been updated with the edited page's data, if any.
-            // This will take priority over other data it finds in the book, even earlier in the book
-            // than the edited page.
-            _bookData.UpdateVariablesAndDataDivThroughDOM(BookInfo); //will update the title if needed
+            if (!preventUpdateFromDataDiv)
+            {
+                // Note that at this point _bookData has already been updated with the edited page's data, if any.
+                // This will take priority over other data it finds in the book, even earlier in the book
+                // than the edited page.
+                // JohnT June 2025: not sure why we sometimes need to do this here, but it's definitely a problem
+                // when working on a duplicate to be published, because it can restore xmatter attributes that
+                // we carefully removed.
+                _bookData.UpdateVariablesAndDataDivThroughDOM(BookInfo); //will update the title if needed
+            }
+
             if (OkToChangeFileAndFolderName)
             {
                 Storage.UpdateBookFileAndFolderName(CollectionSettings); //which will update the file name if needed
--- a/src/BloomExe/Publish/BloomPub/BloomPubMaker.cs
+++ b/src/BloomExe/Publish/BloomPub/BloomPubMaker.cs
@@ -593,7 +593,9 @@ namespace Bloom.Publish.BloomPub
 
             AddDistributionFile(modifiedBookFolderPath, creator, settings);
 
-            modifiedBook.Save();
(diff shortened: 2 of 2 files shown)


andrew-polk (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7194)* status: all files reviewed, 1 unresolved discussion (waiting on @JohnThomson)

___
*[`src/BloomExe/Publish/BloomPub/BloomPubMaker.cs` line 598 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7194#-OT3z4823KFExH2VOGDO:-OT3z4823KFExH2VOGDP:b-tffeyj) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/eece3991a8f0b43aeb1bc6e1acf891ef602aaec0/src/BloomExe/Publish/BloomPub/BloomPubMaker.cs#L598)):*
> ```Smalltalk
>             // If we allow update from the data div we created when we made the book, we will undo changes
>             // we made intentionally (e.g., to attributes of pages in xmatter, BL-14907).
>             modifiedBook.Save(preventUpdateFromDataDiv: true);
> ```

Did you consider whether we should rather remove the attributes from the data div initially? Or sync the changes from the pages to the data div when we make the initial change?

Those seem more principled, IMO

JohnThomson (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7194)* status: all files reviewed, 1 unresolved discussion (waiting on @andrew-polk)

___
*[`src/BloomExe/Publish/BloomPub/BloomPubMaker.cs` line 598 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7194#-OT3z4823KFExH2VOGDO:-OT7Qa__8L4a9LXYc7jj:bsoqwzr) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/eece3991a8f0b43aeb1bc6e1acf891ef602aaec0/src/BloomExe/Publish/BloomPub/BloomPubMaker.cs#L598)):*

Did you consider whether we should rather remove the attributes from the data div initially? Or sync the changes from the pages to the data div when we make the initial change?

Those seem more principled, IMO
</blockquote></details>

I did. It's a tough call. Both of those approaches require this and any future attribute removal, or anything that changes the DOM in a way that the update from the data set would reverse, to do something special (and unobvious) for xmatter pages to prevent a surprising special case like the one I'm trying to fix. (And it was not easy to figure out how the properly removed attribute was coming back!). So they strike me as options I can see clear reasons to dislike. OTOH, what I'm doing carries a less predictable risk that something fixed in one part of the DOM (perhaps this is even already the case?) needs to be copied to another part. However, that won't just happen, because the DataSet is being loaded up before we make ANY of the changes we want for publishing. It would again require special code to reload the DataSet after making the change, and it's probably about as easy to just change it in more than one place if necessary. I'm also encouraged by the comment on the call to UpdateVariablesAndDataDivThroughDOM that suggests rather cryptically that it's only being done in case the title of the book has changed. That won't be the case in making a publishing DOM. Nor can I think of any other reason UpdateVariablesAndDataDivThroughDOM might be needed. So I'm moderately confident that there's no reason to do it in these Save()s where we're making a publishing DOM. But I don't see any way to be completely sure, so I'm not very happy about this approach, either.

JohnThomson (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7194)* status: all files reviewed, 1 unresolved discussion (waiting on @andrew-polk)

___
*[`src/BloomExe/Publish/BloomPub/BloomPubMaker.cs` line 598 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7194#-OT3z4823KFExH2VOGDO:-OT917zzD0_PncmntQHv:bhyzu87) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/eece3991a8f0b43aeb1bc6e1acf891ef602aaec0/src/BloomExe/Publish/BloomPub/BloomPubMaker.cs#L598)):*

Is there any risk of creating a BloomPUB which has a data-div which isn't synced with the content?
I can't think of anything in particular, though it does feel "wrong".

Otherwise, and after seeing the tradeoffs you've listed, I'm ok as is. Feel free to merge.
</blockquote></details>

I don't think so. But can't be sure it never will be, so I made another attempt. Closing this.

JohnThomson:
Found a better approach

PR #7203, opened 2025-06-19
Tweak playground book behavior in various ways (BL-14891)

StephenMcConnel:
This includes opening up the Publish tab but without the ability to save or print the book.  Preview is fully operational for all features.


StephenMcConnel:
See the comment in the issue for what this pull request is trying to accomplish.

JohnThomson (changes requested):
Improved, but I picked up a couple of things I should have caught earlier. Sorry.

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7203#-:-OTXsbn05CmerLB8GKJX:b-1ay07g)* status: all files reviewed, 3 unresolved discussions (waiting on @StephenMcConnel)

___
*[`src/BloomExe/SubscriptionAndFeatures/FeatureStatus.cs` line 121 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7203#-OTXpoEW3uHYSmtHousC:-OTXpoEW3uHYSmtHousD:b-jkzv4j) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/562ab278203050768634e5692972e63c5bcb73f9/src/BloomExe/SubscriptionAndFeatures/FeatureStatus.cs#L121)):*
> ```Smalltalk
>             {
>                 subToUse = Subscription.CreateTempSubscriptionForTier(SubscriptionTier.Enterprise);
>             }
> ```

I should have picked this up earlier, but can't we leave this whole 'if' out and just let the other overload that we're about to call deal with the special case?
My idea in creating the three overloads was that this one handles having the feature identified by a string by parsing the string to get an enum value; the next one handles an enum value by finding the actual feature data in the registry; and given the actual feature data from the registry, the third overload contains all the logic about actually determining the feature status. So if possible I'd like to keep all actual status-determining logic out of the first two.
___
*[`src/BloomExe/SubscriptionAndFeatures/FeatureStatus.cs` line 143 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7203#-OTXqBQo8iYjGe6QyVtT:-OTXqBQo8iYjGe6QyVtU:bs0xccz) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/562ab278203050768634e5692972e63c5bcb73f9/src/BloomExe/SubscriptionAndFeatures/FeatureStatus.cs#L143)):*
> ```Smalltalk
>             {
>                 subToUse = Subscription.CreateTempSubscriptionForTier(SubscriptionTier.Enterprise);
>             }
> ```

Again, can we let the next overload deal with the special case?
___
*[`src/BloomExe/SubscriptionAndFeatures/FeatureStatus.cs` line 201 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7203#-OTXqhsABp1GuBqkpOZ3:-OTXqhsABp1GuBqkpOZ4:b2rv391) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/562ab278203050768634e5692972e63c5bcb73f9/src/BloomExe/SubscriptionAndFeatures/FeatureStatus.cs#L201)):*
> ```Smalltalk
>                 var featureStatus = GetFeatureStatus(
>                     feature.Feature == FeatureName.TeamCollection ? subscription : subToUse,
>                     feature);
> ```

Why do we need any special cases here? Won't we get the desired result by just passing the book to GetFeatureStatus? I suppose there's a tiny performance gain, but don't think it's enough for the cost of having the logic in two places.

StephenMcConnel (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7203)* status: all files reviewed, 3 unresolved discussions (waiting on @JohnThomson)

___
*[`src/BloomExe/SubscriptionAndFeatures/FeatureStatus.cs` line 121 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7203#-OTXpoEW3uHYSmtHousC:-OTXyrHZ6sMQ5JV5mgW-:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/562ab278203050768634e5692972e63c5bcb73f9/src/BloomExe/SubscriptionAndFeatures/FeatureStatus.cs#L121)):*

I should have picked this up earlier, but can't we leave this whole 'if' out and just let the other overload that we're about to call deal with the special case?
My idea in creating the three overloads was that this one handles having the feature identified by a string by parsing the string to get an enum value; the next one handles an enum value by finding the actual feature data in the registry; and given the actual feature data from the registry, the third overload contains all the logic about actually determining the feature status. So if possible I'd like to keep all actual status-determining logic out of the first two.
</blockquote></details>

Done.
___
*[`src/BloomExe/SubscriptionAndFeatures/FeatureStatus.cs` line 143 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7203#-OTXqBQo8iYjGe6QyVtT:-OTXysV4BKsprIMs7XcQ:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/562ab278203050768634e5692972e63c5bcb73f9/src/BloomExe/SubscriptionAndFeatures/FeatureStatus.cs#L143)):*

Again, can we let the next overload deal with the special case?
</blockquote></details>

Done.
___
*[`src/BloomExe/SubscriptionAndFeatures/FeatureStatus.cs` line 201 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7203#-OTXqhsABp1GuBqkpOZ3:-OTXywr-DVHDPMWGT5mv:bwdx8cg) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/562ab278203050768634e5692972e63c5bcb73f9/src/BloomExe/SubscriptionAndFeatures/FeatureStatus.cs#L201)):*

Why do we need any special cases here? Won't we get the desired result by just passing the book to GetFeatureStatus? I suppose there's a tiny performance gain, but don't think it's enough for the cost of having the logic in two places.
</blockquote></details>

Done.  Added the book to the argument list.

JohnThomson (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7203)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @StephenMcConnel)

PR #7208, opened 2025-06-23
Don't complain about Comic Sans MS from playground watermark (BL-14932)

StephenMcConnel:
Also provide the watermark for the Legacy theme.

### DIFF
--- a/src/BloomExe/Publish/PublishHelper.cs
+++ b/src/BloomExe/Publish/PublishHelper.cs
@@ -1602,6 +1602,7 @@ namespace Bloom.Publish
             return false;
         }
 
+        // TODO: (?) use the same file scanning in this method as StylesAndFontsApi.HandleGetDataRows.
         public static void ReportInvalidFonts(string destDirName, IProgress progress)
         {
             // For ePUB and BloomPub, we display the book to determine exactly which fonts are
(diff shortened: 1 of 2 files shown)


JohnThomson (commented):
This will work and we need the change to base page legacy. However, as we discussed, rather than add the complexity of preventing the complaint about this font, let's just change the font used for the watermark to something that won't trigger this.

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7208#-:-OTXzNNJ0-noHBqQiOwI:bdhzq3g)* status: 0 of 2 files reviewed, all discussions resolved

StephenMcConnel (commented):
I changed the watermark font to Noto Sans, which appears to be standard in Windows 10 and 11, and which has the SIL Font license.

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7208#-:-OTYD8qM9CVC8iOkYWdy:bt5kn6z)* status: 0 of 3 files reviewed, all discussions resolved

StephenMcConnel (commented):
I think it's probably only the CJK variants of Noto Sans that are standard now that I look closer.  Regardless, the sans-serif fallback will give something reasonable based on the fonts available.

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7208#-:-OTYIO66D0XPivbGHzaw:benwy7z)* status: 0 of 3 files reviewed, all discussions resolved

JohnThomson (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7208)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @StephenMcConnel)

PR #7209, opened 2025-06-23
BL-14721 menu item toggling draggability

nabalone:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7209)

### DIFF
--- a/DistFiles/localization/am/BloomMediumPriority.xlf
+++ b/DistFiles/localization/am/BloomMediumPriority.xlf
@@ -48,10 +48,10 @@
         <target xml:lang="am" state="needs-translation">Add items that will show when the learner gets the correct answer. For example, you might add a text that says “Correct!” with a GIF of a dancing chicken.</target>
         <note>ID: EditTab.Toolbox.DragActivity.CorrectInstructions</note>
       </trans-unit>
-      <trans-unit id="EditTab.Toolbox.DragActivity.Draggable">
+      <trans-unit id="EditTab.Toolbox.DragActivity.ObjectsToDrag">
         <source xml:lang="en">Draggable</source>
         <target xml:lang="am" state="needs-translation">Draggable</target>
-        <note>ID: EditTab.Toolbox.DragActivity.Draggable</note>
+        <note>ID: EditTab.Toolbox.DragActivity.ObjectsToDrag</note>
       </trans-unit>
       <trans-unit id="EditTab.Toolbox.DragActivity.DragImageHeading">
         <source xml:lang="en">Drag Images to Targets</source>
--- a/DistFiles/localization/ar/BloomMediumPriority.xlf
+++ b/DistFiles/localization/ar/BloomMediumPriority.xlf
@@ -48,10 +48,10 @@
         <target xml:lang="ar" state="needs-translation">Add items that will show when the learner gets the correct answer. For example, you might add a text that says “Correct!” with a GIF of a dancing chicken.</target>
         <note>ID: EditTab.Toolbox.DragActivity.CorrectInstructions</note>
       </trans-unit>
-      <trans-unit id="EditTab.Toolbox.DragActivity.Draggable">
+      <trans-unit id="EditTab.Toolbox.DragActivity.ObjectsToDrag">
         <source xml:lang="en">Draggable</source>
         <target xml:lang="ar" state="needs-translation">Draggable</target>
-        <note>ID: EditTab.Toolbox.DragActivity.Draggable</note>
+        <note>ID: EditTab.Toolbox.DragActivity.ObjectsToDrag</note>
       </trans-unit>
       <trans-unit id="EditTab.Toolbox.DragActivity.DragImageHeading">
         <source xml:lang="en">Drag Images to Targets</source>
--- a/DistFiles/localization/az/BloomMediumPriority.xlf
+++ b/DistFiles/localization/az/BloomMediumPriority.xlf
@@ -48,10 +48,10 @@
         <target xml:lang="az" state="needs-translation">Add items that will show when the learner gets the correct answer. For example, you might add a text that says “Correct!” with a GIF of a dancing chicken.</target>
         <note>ID: EditTab.Toolbox.DragActivity.CorrectInstructions</note>
       </trans-unit>
-      <trans-unit id="EditTab.Toolbox.DragActivity.Draggable">
+      <trans-unit id="EditTab.Toolbox.DragActivity.ObjectsToDrag">
         <source xml:lang="en">Draggable</source>
         <target xml:lang="az" state="needs-translation">Draggable</target>
-        <note>ID: EditTab.Toolbox.DragActivity.Draggable</note>
+        <note>ID: EditTab.Toolbox.DragActivity.ObjectsToDrag</note>
       </trans-unit>
       <trans-unit id="EditTab.Toolbox.DragActivity.DragImageHeading">
         <source xml:lang="en">Drag Images to Targets</source>
--- a/DistFiles/localization/bn/BloomMediumPriority.xlf
+++ b/DistFiles/localization/bn/BloomMediumPriority.xlf
@@ -48,10 +48,10 @@
         <target xml:lang="bn" state="needs-translation">Add items that will show when the learner gets the correct answer. For example, you might add a text that says “Correct!” with a GIF of a dancing chicken.</target>
         <note>ID: EditTab.Toolbox.DragActivity.CorrectInstructions</note>
       </trans-unit>
-      <trans-unit id="EditTab.Toolbox.DragActivity.Draggable">
+      <trans-unit id="EditTab.Toolbox.DragActivity.ObjectsToDrag">
         <source xml:lang="en">Draggable</source>
         <target xml:lang="bn" state="needs-translation">Draggable</target>
-        <note>ID: EditTab.Toolbox.DragActivity.Draggable</note>
+        <note>ID: EditTab.Toolbox.DragActivity.ObjectsToDrag</note>
       </trans-unit>
       <trans-unit id="EditTab.Toolbox.DragActivity.DragImageHeading">
         <source xml:lang="en">Drag Images to Targets</source>
--- a/DistFiles/localization/en/BloomMediumPriority.xlf
+++ b/DistFiles/localization/en/BloomMediumPriority.xlf
@@ -31,9 +31,17 @@
         <source xml:lang="en">Add items that will show when the learner gets the correct answer. For example, you might add a text that says “Correct!” with a GIF of a dancing chicken.</source>
         <note>ID: EditTab.Toolbox.DragActivity.CorrectInstructions</note>
       </trans-unit>
-      <trans-unit id="EditTab.Toolbox.DragActivity.Draggable">
+      <trans-unit id="EditTab.Toolbox.DragActivity.ObjectsToDrag">
         <source xml:lang="en">Objects to Drag</source>
-        <note>ID: EditTab.Toolbox.DragActivity.Draggable</note>
+        <note>ID: EditTab.Toolbox.DragActivity.ObjectsToDrag</note>
+      </trans-unit>
+      <trans-unit id="EditTab.Toolbox.DragActivity.Draggability">
+        <source xml:lang="en">Draggable</source>
+        <note>ID: EditTab.Toolbox.DragActivity.Draggability</note>
+      </trans-unit>
+      <trans-unit id="EditTab.Toolbox.DragActivity.DraggabilityMore">
+        <source xml:lang="en">During game play, this item can be moved around.</source>
+        <note>ID: EditTab.Toolbox.DragActivity.DraggabilityMore</note>
       </trans-unit>
       <!-- <trans-unit id="EditTab.Toolbox.DragActivity.DragImageHeading">
         <source xml:lang="en">Drag Images to Targets</source>
--- a/DistFiles/localization/es/BloomMediumPriority.xlf
+++ b/DistFiles/localization/es/BloomMediumPriority.xlf
@@ -48,10 +48,10 @@
         <target xml:lang="es-ES" state="translated">Añadir elementos que se mostrarán cuando el estudiante reciba la respuesta correcta. Por ejemplo, puedes añadir un texto que dice “Correcto!” con un GIF de un pollo bailando.</target>
         <note>ID: EditTab.Toolbox.DragActivity.CorrectInstructions</note>
       </trans-unit>
-      <trans-unit id="EditTab.Toolbox.DragActivity.Draggable">
+      <trans-unit id="EditTab.Toolbox.DragActivity.ObjectsToDrag">
         <source xml:lang="en">Draggable</source>
         <target xml:lang="es-ES" state="translated">Arrastrable</target>
-        <note>ID: EditTab.Toolbox.DragActivity.Draggable</note>
+        <note>ID: EditTab.Toolbox.DragActivity.ObjectsToDrag</note>
       </trans-unit>
       <trans-unit id="EditTab.Toolbox.DragActivity.DragImageHeading">
         <source xml:lang="en">Drag Images to Targets</source>
--- a/DistFiles/localization/fr/BloomMediumPriority.xlf
+++ b/DistFiles/localization/fr/BloomMediumPriority.xlf
@@ -48,10 +48,10 @@
         <target xml:lang="fr" state="needs-translation">Add items that will show when the learner gets the correct answer. For example, you might add a text that says “Correct!” with a GIF of a dancing chicken.</target>
         <note>ID: EditTab.Toolbox.DragActivity.CorrectInstructions</note>
       </trans-unit>
-      <trans-unit id="EditTab.Toolbox.DragActivity.Draggable">
+      <trans-unit id="EditTab.Toolbox.DragActivity.ObjectsToDrag">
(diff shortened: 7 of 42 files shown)


JohnThomson (commented):
This could stand, but I think we could simplify the key stuff, and also awaiting a decision on whether to include text items, so I won't merge or officially approve yet.

> Reviewed 2 of 3 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7209#-:-OTY_4-fEEhvIIWvM_R1:butcryp)* status: all files reviewed, 4 unresolved discussions (waiting on @nabalone)

___
*[`DistFiles/localization/en/BloomMediumPriority.xlf` line 34 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7209#-OTU4Ky_9sgoCrfiC-xF:-OTY1SbcAJuugTorsfft:bjltcdu) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/2a15f369d33a7db72311704c54a0efe3f6a038d9/DistFiles/localization/en/BloomMediumPriority.xlf#L34)):*

It seems a little confusing that EditTab.Toolbox.DragActivity.Draggable is a different string than the one that has English text "Draggable", but is it too late to change the existing IDs? Seems like that might have complications?
</blockquote></details>

We haven't shipped any drag games except to FAL. I'm a little surprised they don't still have translate="no". I suppose someone might have translated some, but I don't think it's too late to change an ID that is clearly a poor choice.
___
*[`src/BloomBrowserUI/bookEdit/js/CanvasElementContextControls.tsx` line 610 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7209#-OTTzNJa5qzabkMnmQmp:-OTYPuZ4BDCEEwUevwus:bpezhec) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/2a15f369d33a7db72311704c54a0efe3f6a038d9/src/BloomBrowserUI/bookEdit/js/CanvasElementContextControls.tsx#L610)):*

Might we ever make multiple menu items with the same l10nId so this would be a problem? I'm not sure what would be better...

If we were to put the "Part of the Right Answer" menu item above the "Draggable" menu item (as I had originally mistakenly thought), then "real" keys are necessary here -- otherwise if you click "Draggable" and make the element non-draggable, then the "Part of the Right Answer" menu item goes away, and the menu is messed up such that the click handler can't tell that your click was on the menu and so it thinks you clicked the page and it unselects the element. 

Even though (if I tested correctly) things work okay now with the indices as keys and the menu items in the correct order, it still seems like such weird unpredictable fragility is undesirable?
</blockquote></details>

Since we're not trying to make something totally general-purpose but just something that will work for this menu, if in fact every possible item has a unique l10nId, we can just use key={option.l10nId"}, I think without any addition or fallback.
If we wanted a more general mechanism that would handle options without unique l10nId, I'd be inclined to give the typedef for option its own key property. To save filling them in unnecessarily, we could do key={options.key ?? options.l10nId} and then you only have to fill in key if for any reason using L10nId as key is a problem.
Sorry I didn't think that using indexes was risky in an environment where we change our minds about which ones to include.
___
*[`src/BloomBrowserUI/bookEdit/js/CanvasElementContextControls.tsx` line 1081 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7209#-OTU3Su08u5floQ3hIWE:-OTYS6GK6AqfmoJ5PsCO:b-lyfzeo) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/2a15f369d33a7db72311704c54a0efe3f6a038d9/src/BloomBrowserUI/bookEdit/js/CanvasElementContextControls.tsx#L1081)):*

Do we want to be able to change the draggability of Word / Text? By default, Word (draggable) has a visible background and Text (non-draggable) does not, and this seems like a helpful visible distinction, so I was thinking better not?

Also, is it preferable to use element.getElementsByClassName or element.querySelector? In the code sometimes we use one and sometimes the other when checking existence like this
</blockquote></details>

In the past, I've thought of getElementsByClassName as the natural way to get things that have a particular class, and querySelector as a more complicated way to get things I need to describe in other ways, like having a certain attribute. However, Copilot seems to prefer querySelector(".X") to getElementsByClassName("X")[0] when you only want the first, and I'm coming round to its opinion. It's shorter and arguably clearer and might even be more efficient, especially if getElementsByClassName has to search beyond the first one to find all of them.

I raised the other question in the chat. I lean towards allow them to be switched, but switching the style, too.

nabalone (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7209)* status: 0 of 42 files reviewed, 3 unresolved discussions (waiting on @JohnThomson)

___
*[`DistFiles/localization/en/BloomMediumPriority.xlf` line 34 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7209#-OTU4Ky_9sgoCrfiC-xF:-OTYw5Uu3cP8V37XCSRs:bb6un1j) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/2a15f369d33a7db72311704c54a0efe3f6a038d9/DistFiles/localization/en/BloomMediumPriority.xlf#L34)):*

We haven't shipped any drag games except to FAL. I'm a little surprised they don't still have translate="no". I suppose someone might have translated some, but I don't think it's too late to change an ID that is clearly a poor choice.
</blockquote></details>

Done
___
*[`src/BloomBrowserUI/bookEdit/js/CanvasElementContextControls.tsx` line 610 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7209#-OTTzNJa5qzabkMnmQmp:-OTYx0fEDb68j7iOdvo3:bb6un1j) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/2a15f369d33a7db72311704c54a0efe3f6a038d9/src/BloomBrowserUI/bookEdit/js/CanvasElementContextControls.tsx#L610)):*

Since we're not trying to make something totally general-purpose but just something that will work for this menu, if in fact every possible item has a unique l10nId, we can just use key={option.l10nId"}, I think without any addition or fallback.
If we wanted a more general mechanism that would handle options without unique l10nId, I'd be inclined to give the typedef for option its own key property. To save filling them in unnecessarily, we could do key={options.key ?? options.l10nId} and then you only have to fill in key if for any reason using L10nId as key is a problem.
Sorry I didn't think that using indexes was risky in an environment where we change our minds about which ones to include.
</blockquote></details>

Done
___
*[`src/BloomBrowserUI/bookEdit/js/CanvasElementContextControls.tsx` line 1081 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7209#-OTU3Su08u5floQ3hIWE:-OTnPsMK7yCCI9x-lFfT:bx0yikp) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/2a15f369d33a7db72311704c54a0efe3f6a038d9/src/BloomBrowserUI/bookEdit/js/CanvasElementContextControls.tsx#L1081)):*

In the past, I've thought of getElementsByClassName as the natural way to get things that have a particular class, and querySelector as a more complicated way to get things I need to describe in other ways, like having a certain attribute. However, Copilot seems to prefer querySelector(".X") to getElementsByClassName("X")[0] when you only want the first, and I'm coming round to its opinion. It's shorter and arguably clearer and might even be more efficient, especially if getElementsByClassName has to search beyond the first one to find all of them.

I raised the other question in the chat. I lean towards allow them to be switched, but switching the style, too.
</blockquote></details>

Done. I have moved canToggleDraggability to be inside the render 
___
*[`src/BloomBrowserUI/bookEdit/js/CanvasElementContextControls.tsx` line 213 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7209#-OTnQ26t1pTygOmDTm4e:-OTnQ26uBrbwuePDp6gl:b-q1j9uz) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/e1f1df05e029d1be79ce3cf424b695d6bdcd5c6d/src/BloomBrowserUI/bookEdit/js/CanvasElementContextControls.tsx#L213)):*
> ```XML
>         // Audio currently cannot be made non-draggable
>         !props.canvasElement.querySelector(`[data-icon-type="audio"]`);
> ```

Moved here since we're doing the other menu logic in the render. I also added a bunch of other restrictions of things that would be problematic to make draggable. Please take a look and see if we want to be more aggressive or conservative and if there are other cases I missed
___
*[`src/BloomBrowserUI/bookEdit/toolbox/games/GamePromptDialog.tsx` line 264 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7209#-OTnQqh03R4ZXkSM4R0w:-OTnQqh03R4ZXkSM4R0x:bvriqs4) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/e1f1df05e029d1be79ce3cf424b695d6bdcd5c6d/src/BloomBrowserUI/bookEdit/toolbox/games/GamePromptDialog.tsx#L264)):*
> ```XML
>     // capture where the top left draggable and target are (before we add or remove any).
>     // Also capture various bits of initial state that cancel() might need.
>     originalDraggables = Array.from(
> ```

Not sure what to do here... why is there originalDraggables and also draggables below? I don't think this code is even in use because it only applies to the Drag Letters game, right? Is it worth cleaning up and testing?

nabalone:
Might be helpful to review the two commits separately. Hope I didn't confuse Reviewable too much when I rebased.

JohnThomson (changes requested):
Good work. Just a few more things to consider

> Reviewed 42 of 42 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7209#-:-OU5zua4AcilyssWnbjn:beiwbn8)* status: all files reviewed, 4 unresolved discussions (waiting on @nabalone)

___
*[`src/BloomBrowserUI/bookEdit/js/CanvasElementContextControls.tsx` line 213 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7209#-OTnQ26t1pTygOmDTm4e:-OU5p6WZCKNAWoqCAGwT:b7cabuq) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/e1f1df05e029d1be79ce3cf424b695d6bdcd5c6d/src/BloomBrowserUI/bookEdit/js/CanvasElementContextControls.tsx#L213)):*

Moved here since we're doing the other menu logic in the render. I also added a bunch of other restrictions of things that would be problematic to make draggable. Please take a look and see if we want to be more aggressive or conservative and if there are other cases I missed
</blockquote></details>

Some of these we might relax later (for example, an order words game could in principle also have some other tasks to do...maybe when unscrambled the sentence says "Give the monkey a banana" and you need to do that). You also made me realize that it might be desirable to have a way to move elements between start/correct/wrong. But sometime soon we will have a way to cut and paste entire overlay elements which will be one way to do that. I think what you have is consistent with what is currently possible to do in other ways.
___
*[`src/BloomBrowserUI/bookEdit/toolbox/games/GamePromptDialog.tsx` line 264 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7209#-OTnQqh03R4ZXkSM4R0w:-OU5xVQs30-fghefjo92:bz3oiff) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/e1f1df05e029d1be79ce3cf424b695d6bdcd5c6d/src/BloomBrowserUI/bookEdit/toolbox/games/GamePromptDialog.tsx#L264)):*

Not sure what to do here... why is there originalDraggables and also draggables below? I don't think this code is even in use because it only applies to the Drag Letters game, right? Is it worth cleaning up and testing?
</blockquote></details>

I think draggables gets modified as we go, while originalDraggables maintains the original state of things to restore if the dialog gets canceled. I doubt there is any good reason to calculate them differently, though. I'd be inclined to set originalDraggables and then set draggables (initially) to [...originalDraggables]. We'll test when we re-enable this game, if we haven't totally changed the model by then (e.g., we might make it work like the order words game, where creating and shuffling targets is a play-time action).
___
*[`src/BloomBrowserUI/bookEdit/js/CanvasElementContextControls.tsx` line 1124 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7209#-OU5s2z3474MndR2daGW:-OU5s2z3474MndR2daGX:brnkvga) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/e1f1df05e029d1be79ce3cf424b695d6bdcd5c6d/src/BloomBrowserUI/bookEdit/js/CanvasElementContextControls.tsx#L1124)):*
> ```XML
>     const newClassList = classList.map(modification);
>     element.classList.remove(...classList);
>     element.classList.add(...newClassList);
> ```

You don't need it for the current purpose, but it might be worth filtering out empty strings (which will cause a crash) from newClassList.
___
*[`src/BloomBrowserUI/bookEdit/js/CanvasElementContextControls.tsx` line 1159 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7209#-OU5tYhV0B56YLPav5ev:-OU5tYhV0B56YLPav5ew:btgzlnw) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/e1f1df05e029d1be79ce3cf424b695d6bdcd5c6d/src/BloomBrowserUI/bookEdit/js/CanvasElementContextControls.tsx#L1159)):*
> ```XML
>             ) {
>                 modifyAllDescendantsClassNames(canvasElement, className =>
>                     className.replace(/GameDrag(\S*)-style/, "GameText$1-style")
> ```

I'm tossing up whether \S* should instead be (Small|Medium|Large)(Start|Center). On the one hand, if the user creates a pair of styles like GameDragTiny and GameTextTiny, your code will do exactly what we want. OTOH, if they create just one of those, then converting will switch the element to a completely undefined style. But that's unlikely, and they can go ahead and define it or choose a different one. I'm inclined to let what you have stand, but I thought it was worth thinking about.
___
*[`src/BloomBrowserUI/bookEdit/js/CanvasElementManager.ts` line 428 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7209#-OU5vUXDE3JjOMJFDAAI:-OU5vUXDE3JjOMJFDAAJ:b-jjp4dv) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/e1f1df05e029d1be79ce3cf424b695d6bdcd5c6d/src/BloomBrowserUI/bookEdit/js/CanvasElementManager.ts#L428)):*
> ```XML
>             cleanupNiceScroll();
>             addScrollbarsToPage(editable.closest(".bloom-page") as HTMLElement);
>         }, 200);
> ```

This looks like a piece of another change got pulled in, maybe during rebase? Make sure we don't end up getting it added twice or something like that.

nabalone (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7209)* status: 38 of 42 files reviewed, 3 unresolved discussions (waiting on @JohnThomson)

___
*[`src/BloomBrowserUI/bookEdit/js/CanvasElementManager.ts` line 428 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7209#-OU5vUXDE3JjOMJFDAAI:-OU6y4OaAuWkQYt32Lvs:bktuahv) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/e1f1df05e029d1be79ce3cf424b695d6bdcd5c6d/src/BloomBrowserUI/bookEdit/js/CanvasElementManager.ts#L428)):*

Is it better now?
</blockquote></details>

I don't see it in the diff, I think it's gone?

JohnThomson (changes requested):
> Reviewed 4 of 4 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7209)* status: all files reviewed, 1 unresolved discussion (waiting on @nabalone)

___
*[`src/BloomBrowserUI/bookEdit/js/CanvasElementContextControls.tsx` line 1186 at r4](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7209#-OU6z9lI9WuPk2i1c4g8:-OU6z9lI9WuPk2i1c4g9:b-czip5) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/8c5aa3fd861b8b7a191919569c93dbceddcb2736/src/BloomBrowserUI/bookEdit/js/CanvasElementContextControls.tsx#L1186)):*
> ```XML
>             ) {
>                 modifyAllDescendantsClassNames(canvasElement, className =>
>                     className.replace(/GameText(\S*)-style/, "GameDrag$1-style")
> ```

Sorry, if you're changing one direction, you should change the other, too.

nabalone (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7209)* status: 41 of 42 files reviewed, 1 unresolved discussion (waiting on @JohnThomson)

___
*[`src/BloomBrowserUI/bookEdit/js/CanvasElementContextControls.tsx` line 1186 at r4](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7209#-OU6z9lI9WuPk2i1c4g8:-OUAUcNl-7Hw0m00XGLG:b-vpa8dr) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/8c5aa3fd861b8b7a191919569c93dbceddcb2736/src/BloomBrowserUI/bookEdit/js/CanvasElementContextControls.tsx#L1186)):*

Sorry, if you're changing one direction, you should change the other, too.
</blockquote></details>

Silly me, forgot about that. Done.

PR #7220, opened 2025-06-25
Added a comment indicating how to generate a new sample book for Bloo…

NJStriebel:
…m Reader (BL-14927)

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7220)


andrew-polk:
Thanks for figuring out the best way to do this.
Making this change in the code and generating the book is probably simpler than the initial instructions, actually.

But I'm not seeing much value in adding this comment to BloomDesktop.
I think the update to the README in BloomReader is sufficient.

I'll go ahead and close this, but if you think differently, let me know.

PR #7231, opened 2025-06-26
WiFi: ensure proper network interface is used to send adverts and receive requests

wmergenthal:
When setting up to advertise a book over WiFi, BloomDesktop can sometimes use the wrong local IP address due to using the wrong network interface. I observed this in my home setup last November for the first time: C# sent UDP broadcast adverts out over a virtual interface associated with my VMware installation. Its local IP address lived on a different subnet, which a UDP broadcast cannot reach beyond, and the desired Desktop/Reader connection never occurred.

This changeset addresses the issue by determining the local IP like this:
A) Examine the machine's network interfaces. Note that there could be several interfaces - my Win10 laptop has 11. Focus on the *active* Wi-Fi and Ethernet network interfaces and record their "interface metric" values.
B) Determine which interface the network stack will use for network traffic. Literature says the stack wants to use the interface with the lowest metric value. We want to prefer Wi-Fi over Ethernet. I have not seen a case where these criteria conflict, so the "winning" interface is selected thus:
    - IF a Wi-Fi interface is active, choose that
    - ELSE IF an Ethernet interface is active, choose that
    - ELSE declare 'none' and no adverts will be sent
An interface is considered active if its interface metric is less than the initialized value of (2^31 - 1). On my machine at least, inactive interfaces all do maintain this initial value.
C) Note the winning interface's IP address - it will be the Local IP address.
D) Also note the winning interface's subnet mask and, in conjunction with the Local IP address, calculate the "Directed" broadcast address. This will be the Remote IP address to which Desktop advertising is sent. This replaces 255.255.255.255, which is the "Limited" broadcast address.
E) Use Local IP to instantiate an IPEndPoint, then use that to instantiate a UdpClient.
F) Use Remote IP to instantiate another IPEndPoint, then make UdpClient use it for sending book adverts.

This changeset also addresses a second issue that I encountered: occasionally the book advert got to Reader okay, and Reader replied with a book request, but Desktop never got the request. The Reader side looked good, as far as I could tell. I concluded that Desktop was perhaps listening on the wrong interface. I learned that Desktop's listening UdpClient can be instantiated such that it will listen on *all* interfaces. This is what the changes in BloomReaderUDPListener are for. UdpClient was previously instantiated without specifying an IP address, but now we specify 0.0.0.0 ("IPAddress.Any"). I have not experienced similar failures since.

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7231)

### DIFF
--- a/src/BloomExe/Publish/BloomPub/wifi/WiFiAdvertiser.cs
+++ b/src/BloomExe/Publish/BloomPub/wifi/WiFiAdvertiser.cs
@@ -213,23 +213,38 @@ namespace Bloom.Publish.BloomPub.wifi
             //
             // The network stack is going to use the interface it wants. We can't control what it
             // chooses. But we *can* control the IP address we advertise to remote Androids. We get
-            // that address up front using the same selection criteria as the network stack: find
-            // the network interface having the lowest "interface metric" -- Ethernet or WiFi, it
-            // could be either -- and advertise *that* interface's IP address.
-            // We then use that same IP address as the basis for the UdpClient.
+            // that address up front via an educated guess of which network interface the stack will
+            // choose, and advertise that interface's IP address.
             //
-            // The PC on which this runs likely has both WiFi and Ethernet. Either can work,
-            // but preference is given to WiFi. The reason: although this PC can likely go either
-            // way, the Android device only has WiFi. For the book transfer to work both Desktop
-            // and Reader must be on the same subnet. If Desktop is using Ethernet it may not be
-            // on the same subnet as Reader, especially on larger networks. The chances that both
-            // PC and Android are on the same subnet are greatest if both are using WiFi.
+            // This educated guess starts by checking each network in the system and, if it is up,
+            // noting its "interface metric" -- the main criterion the stack uses most of the time.
+            // There are rare exceptions when additional criteria carry greater weight, resulting
+            // in the stack making a different selection. As far as I can tell, those are rare
+            // enough and have mitigations complex enough that accounting for them could introduce
+            // more risk than reward. For more information these posts are helpful:
+            // https://learn.microsoft.com/en-us/troubleshoot/windows-server/networking/automatic-metric-for-ipv4-routes
+            // https://en.softcomputers.org/change-wireless-network-connection-priority-in-windows/
+            //
+            // The stack normally chooses the network interface having the lowest interface metric.
+            // This could be either Ethernet or WiFi (most PC laptops have both). However, we will
+            // prefer WiFi since (a) the Android only has WiFi, and (b) a successful book transfer
+            // requires both Desktop and Reader to be on the same subnet. If Desktop is using Ethernet
+            // it may not be on the same subnet as Reader, especially on larger networks. The chances
+            // that both PC and Android are on the same subnet are greatest if both are using WiFi.
+            //
+            // Our process boils down to this:
+            //   1. Find the active WiFi interface with the lowest interface metric,
+            //      and use its IP address;
+            //   2. Otherwise, find the active Ethernet interface with the lowest interface metric,
+            //      and use its IP address;
+            //   3. Otherwise, fail: IP address not found.
+            // The winning IP address is then used as the basis for the UdpClient which performs the
+            // actual data transfer.
             //
             // This method generates (a) the local IP address *from* which the book advert is
             // sent, and (b) the "directed broadcast address" *to* which the book advert is sent.
 
-            // Examine the network interfaces and determine which will be used for network traffic.
-            InterfaceInfo ifcResult = GetInterfaceStackWillUse();
+            InterfaceInfo ifcResult = GetInterfaceToAdvertise();
 
             if (ifcResult == null)
             {
@@ -307,6 +322,8 @@ namespace Bloom.Publish.BloomPub.wifi
 
                 // If we do eventually add capability to display the advert as a QR code,
                 // the local IP address will need to be included. Might as well add it now.
+                // It will be visible in a network trace, showing what Desktop considers
+                // its local IP to be.
                 advertisement.senderIP = _currentIpAddress;
 
                 _sendBytes = Encoding.UTF8.GetBytes(advertisement.ToString());
@@ -314,16 +331,16 @@ namespace Bloom.Publish.BloomPub.wifi
             }
         }
 
-        // Survey the network interfaces and determine which one, if any, the network
-        // stack will use for network traffic.
+        // Survey the network interfaces and make an educated guess as to which one,
+        // if any, the network stack will use for network traffic.
         //   - During the assessment the current leading WiFi candidate will be held in
         //     'IfaceWifi', and similarly the current best candidate for Ethernet will
         //     be in 'IfaceEthernet'.
-        //   - After assessment inform calling code of the winner by returning an enum
-        //     indicating which of the candidate structs to draw from: WiFi, Ethernet,
-        //     or neither.
+        //   - After assessment inform calling code of the winner by returning the winning
+        //     candidate struct to draw from -- WiFi or Ethernet -- or null if there is
+        //     no winner.
         //
-        private InterfaceInfo GetInterfaceStackWillUse()
+        private InterfaceInfo GetInterfaceToAdvertise()
         {
             InterfaceInfo IfaceWifi = new();
             InterfaceInfo IfaceEthernet = new();


andrew-polk:
By the way, @wmergenthal. We usually use Reviewable for reviews. I think you'll have a better experience if you use that then the normal github UI.
See the big purple "Reviewable" button at the top.

wmergenthal (commented):
I have addressed all comments. I have also made corresponding changes where appropriate, which have been committed and pushed.

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7231#-:-Ocliigh4gYRA2O576V2:b-1j21f0)* status: 0 of 2 files reviewed, 17 unresolved discussions (waiting on @JohnThomson)

___
*[a discussion](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7231#-Oauj2QI-G81Gnz5zmnc:-Ob9X25g6ehtmMq_h-f0:bltxmu7) (no related file):*

Wade, you've done some great work here that clearly represents a lot of research, and I think it is likely to improve things significantly. I've made quite a lot of comments, but that is not a reflection on the quality of your work, just something pretty much to be expected when you're a new contributor to a project.
It's also possible that some of my comments reflect a fundamental misunderstanding of how some of this works. Feel free to point out that I'm off track, either in a response to the review or (usually better) by improving comments so that the next reader, who will likely be at least as ignorant as I am, will be less likely to have the same problem.

Concerning Debug.Writeline: if these are producing information that is only useful when someone is working on an update to this code, they should probably be commented out once you have it working, just to reduce clutter in the Output window when devs are working on something else.
Conversely, if they capture information that would be useful for remote technical support (e.g., when someone uses Report a Problem to tell us that their phone is not seeing our adds or not getting a book it says it has requested), then Debug.Writeline is no use. (The compiler doesn't even generate code for Debug.WriteLine in a Release build.) You should use EventLog.WriteEntry for this kind of information.)
</blockquote></details>

Thanks John, will do.
And thanks for mentioning Debug.Writeline and its disappearance in release code. There are definitely a few of those that I will want to convert to EventLog.WriteEntry, and will remove the rest.
___
*[`src/BloomExe/Publish/BloomPub/wifi/BloomReaderUDPListener.cs` line 52 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7231#-OaukTJKExrOkV77Muv7:-Oc1YgvhAtKYZgbOEc48:bhohbao) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/9e06a5003ae2ce833333b42ad8c7e9d157e08155/src/BloomExe/Publish/BloomPub/wifi/BloomReaderUDPListener.cs#L52)):*

I'm not sure, but I think this wants a bit of adjustment in view of what you say below about IPAddress.Any. By using that, we don't have to figure out a 'right' IP address, we just prevent the default behavior of letting the system choose a single one that might be wrong.
</blockquote></details>

I agree. I improved the comments.
___
*[`src/BloomExe/Publish/BloomPub/wifi/WiFiAdvertiser.cs` line 38 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7231#-OauMMyqD1q4e82qwMfr:-Ob9p5PQ09GmXKZJa3cd:b-lu8di7) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/9e06a5003ae2ce833333b42ad8c7e9d157e08155/src/BloomExe/Publish/BloomPub/wifi/WiFiAdvertiser.cs#L38)):*

I think this is the client we will receive 'answers' on resulting from our BeginSend. It deserves a comment and I think a better name, like _clientForAddReplies or even _clientForBookRequests. I think it's worth capturing here that this client captures our own IP address on the network interface we have decided to broadcast on as well as the (fixed) port we are using.
</blockquote></details>

No, this client does only the book advert broadcasts. The client that receives replies from Android devices is (well, was) '_listener' in BloomReaderUDPListener.cs.
In trying to minimize changes I didn't rename _listener, but I certainly can. In that spirit, I am renaming both clients more precisely: '_clientForBookAdvertSend' and '_clientForBookRequestReceive'.
___
*[`src/BloomExe/Publish/BloomPub/wifi/WiFiAdvertiser.cs` line 40 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7231#-OauNxfJ8GNWgbpugeEC:-Ob9rUu64QFIEBLxxg_8:bcejsr5) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/9e06a5003ae2ce833333b42ad8c7e9d157e08155/src/BloomExe/Publish/BloomPub/wifi/WiFiAdvertiser.cs#L40)):*

This seems to be obsolete and can be removed. (The only previous use was in BeginSend for the advert, and you now use _remoteEP instead
</blockquote></details>

Agree. Removing _localEP.
___
*[`src/BloomExe/Publish/BloomPub/wifi/WiFiAdvertiser.cs` line 43 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7231#-OauVF5iCnAPqYO-buqm:-Ob9sWz-21JkuxIEwOlH:b-2s7maj) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/9e06a5003ae2ce833333b42ad8c7e9d157e08155/src/BloomExe/Publish/BloomPub/wifi/WiFiAdvertiser.cs#L43)):*

We need better names and commenting here. It took me a LONG time to figure out that _localIp is the address on which we expect to receive replies, while _remoteIp is the address TO which we will broadcast (thus not really the IP address of a specific device, but a magic value that causes the widest broadcast that the router will allow). How about _ipForReceivingReplies (or _ipToListenForBookRequests or even _ipToSendToDeviceForBookRequests) and _ipForSendingBroadcast?
</blockquote></details>

Ok. This is another area where I tried to maintain consistency with existing names, but I agree that improvements are possible.
I will clarify names and comments.
___
*[`src/BloomExe/Publish/BloomPub/wifi/WiFiAdvertiser.cs` line 98 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7231#-Oaufwts88WLyf3ghBnm:-Oc1R-1V5-osclUxBeRL:bxnphmo) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/9e06a5003ae2ce833333b42ad8c7e9d157e08155/src/BloomExe/Publish/BloomPub/wifi/WiFiAdvertiser.cs#L98)):*

In what sense is it a type to Expect? It seems to me to be all about deciding which IP address we will use to send the broadcast. Maybe just CommType? Or get rid of; see below.
</blockquote></details>

Renamed to CommType.
Ah, update -- removed it altogether in response to a later comment below.
___
*[`src/BloomExe/Publish/BloomPub/wifi/WiFiAdvertiser.cs` line 109 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7231#-OauXUERCQcyHv602tz8:-Oc1eAJW0IPYEmlREvQZ:b-siaw3j) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/9e06a5003ae2ce833333b42ad8c7e9d157e08155/src/BloomExe/Publish/BloomPub/wifi/WiFiAdvertiser.cs#L109)):*

If I'm understanding right, this is BOTH the port we will use to send the broadcast AND the one through which we expect replies...which in principle don't have to be the same? (But they do because that's what BR expects.)
Also, the leading underline implies a field which might change during execution; this is a constant which must not change.
How about kPortForBroadcastAndReplies? or kPortForBroadcastAndBookRequests?
(A lot of names I'm suggesting indicate things that have to do with receiving replies to BeginSend, which will be requests for the book. I'm not sure which concept it is most important to capture, but we should pick one or both and be consistent. Maybe it's even worth using something like RepliesRequestingBooks?)
</blockquote></details>

No, _portForBroadcast (5913) is for the outgoing book advert only; Android/BloomReader listens on that port.
BloomReader responds by sending on port 5915 a book request; BloomDesktop listens on that port (BloomReaderUDPListener._portToListen).
Regarding leading underscore -- information I find online indicates that it has no single universal meaning. One of the main usages mentioned in several articles, including one from Microsoft Learn (https://learn.microsoft.com/en-us/dotnet/csharp/fundamentals/coding-style/identifier-names), is often to indicate something with private scope. Projects *could* presumably use it to indicate a static value, if such practice is consistently followed and documented (but BloomDesktop does not).
___
*[`src/BloomExe/Publish/BloomPub/wifi/WiFiAdvertiser.cs` line 134 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7231#-OauaaPNB0ib-lJ82UGe:-Oc1kz186TiKAzd8fiAQ:b-n5g435) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/9e06a5003ae2ce833333b42ad8c7e9d157e08155/src/BloomExe/Publish/BloomPub/wifi/WiFiAdvertiser.cs#L134)):*

If I'm understanding this right, "use for the broadcast" could mean two different things. In the BeginSend, there is an IP address and port set in what is currently called _clientBroadcast, which is our own IP address where we will listen for replies. (Maybe that's what you mean by 'advertise in the broadcast?).  And there is an IP address where we will send the broadcast, which is derived from that.
</blockquote></details>

I revised the sentence for clarity.
___
*[`src/BloomExe/Publish/BloomPub/wifi/WiFiAdvertiser.cs` line 148 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7231#-Oauece48Dn2fqWD12WM:-Oc1mEVL6J5mlzWUsugu:bvgutzh) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/9e06a5003ae2ce833333b42ad8c7e9d157e08155/src/BloomExe/Publish/BloomPub/wifi/WiFiAdvertiser.cs#L148)):*

I'm not sure I'm getting this. I expect the Android to reply to the IP address we figured out as _localIp and put in _clientBroadcast, which is somehow communicated as the sender of the broadcast, though we don't explicitly put it in the message. I think you're saying that we can't control which network interface the reply will come on; rather (something?) will choose the network with the lowest metric.
But if that's so, what if the lowest overall metric is for Ethernet? What causes the reply to come on the lowest WiFi network rather than the lowest overall?
</blockquote></details>

You are correct in that the Android will reply to the _localIp that Desktop advertised from (_localIp is baked into the UDP packet header, and now Desktop DOES explicitly put it also in the packet payload: the 'senderIP' field, a newly added field in this pull request).
Regarding network interface -- the thing not being controlled was not which network interface the Android's reply comes in on, but rather which interface the advert goes out on. The "wrong" interface's "wrong" IP address gets put into the broadcast. The Android thus becomes misinformed about where to respond. Android sends its book request somewhere else, and Desktop never hears it.
___
*[`src/BloomExe/Publish/BloomPub/wifi/WiFiAdvertiser.cs` line 191 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7231#-OauII2NBrj23z1Ax8vQ:-Oc1rmy9AYJ662L3rvZU:beqfqyc) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/9e06a5003ae2ce833333b42ad8c7e9d157e08155/src/BloomExe/Publish/BloomPub/wifi/WiFiAdvertiser.cs#L191)):*

You can declare and init this in one line.
Also, I don't like the name. For a long time, I thought this endpoint encapsulated the IP address and port to which we would send the advertisement broadcast. But if I'm now understanding correctly, it's more involved in receiving replies. That is, it combines THIS device's IP address on the chosen network with the fixed port.
I suppose we can't use IPAddress.Any here to listen to all networks, because this is the IP address that gets sent to the Android so it can find the sender?
</blockquote></details>

I agree, two lines are not needed to declare and init. I combined them into one line.
This endpoint DOES encapsulate the IP address and port to which we send the advert broadcast. I renamed it to make this more explicit.
___
*[`src/BloomExe/Publish/BloomPub/wifi/WiFiAdvertiser.cs` line 217 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7231#-OauIy-94H_7v9UagnVB:-Oc1vGov2ausicha8IIc:b-wahwyr) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/9e06a5003ae2ce833333b42ad8c7e9d157e08155/src/BloomExe/Publish/BloomPub/wifi/WiFiAdvertiser.cs#L217)):*

Since we're not trying to localize here, this is probably more readable using the new $ syntax to inline variable subsctitutions, e.g.,
$"                          _remoteIp = {_remoteEP.Address}:{_remoteEP.Port}"
</blockquote></details>

Agree. Made that change to the 3 lines. Also converted the lines to use EventLog.WriteEntry() since their purpose is to provide context for Tech Support.
___
*[`src/BloomExe/Publish/BloomPub/wifi/WiFiAdvertiser.cs` line 219 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7231#-Oaun2xICT_L5vYs08iQ:-Oc1xnUX6NPIrKdJ1oas:b6bu271) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/9e06a5003ae2ce833333b42ad8c7e9d157e08155/src/BloomExe/Publish/BloomPub/wifi/WiFiAdvertiser.cs#L219)):*

Well, we don't have any idea whether any remote devices are ready. We have the IP addresses we need set up in the relevant objects.
</blockquote></details>

True. Comment has been stricken.
___
*[`src/BloomExe/Publish/BloomPub/wifi/WiFiAdvertiser.cs` line 226 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7231#-OauQtGKAdPXUYyfwjs4:-Oc1y5cW9XJOHQT3MaNa:bbgceuk) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/9e06a5003ae2ce833333b42ad8c7e9d157e08155/src/BloomExe/Publish/BloomPub/wifi/WiFiAdvertiser.cs#L226)):*

What is TEMPORARY?
</blockquote></details>

It's a reminder to myself to delete this debug output line before commit & push.
I missed this one; removed it now.
___
*[`src/BloomExe/Publish/BloomPub/wifi/WiFiAdvertiser.cs` line 272 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7231#-OauYln-2yK9Qw9HIV8V:-OclcbykA6WnMx52i1Sy:b-dzsp6s) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/9e06a5003ae2ce833333b42ad8c7e9d157e08155/src/BloomExe/Publish/BloomPub/wifi/WiFiAdvertiser.cs#L272)):*

Somehow you've lost the point of these two variables and the GetLocalIpAddress function (my bad, failing to comment). My idea was that the IP address we want to use might change; for example, perhaps the user realizes he needs to unplug his Ethernet cable and turn on WiFi. Perhaps he decides to disconnect his computer from the campus WiFi and connect to a hotspot on the phone he wants to transfer to. Perhaps he has turned on and connected to the classroom WiFi router. It would be helpful then if the code that calculates _localIp ran again and noticed that it can now, more helpfully, broadcast on a different network. So the idea of this code is that GetLocalIpAddress figures out what would Right Now be the best address to work with for this device, and if it has changed since the last broadcast (not equal to _cachedIpAddress, which might better be called _previousLocalIp) then we update everything that depends on it.
If that isn't feasible with the new approach, there's no point in having these two variables, which always end up being the same as _localIp.
If it is feasible, there may well be other variables that need to be updated along with _localIp/_currentIpAddress.
</blockquote></details>

Your commenting was fine; I understood the intent to handle a possible change of local IP address. I assumed that once per advertising session would be sufficient, that if the user changed network connections he would also restart advertising. I didn't think that the time spent rediscovering the local IP address, and consequently recalculating the broadcast address, was necessary every advert.
But I never actually measured the time, until now. Turns out it's not that bad (a quarter second on my laptop). I have therefore put IP discovery back in the advertising loop. Testing shows that it can take a minute or more for the network stack to fully switch from Wi-Fi to Ethernet or vice versa, but when it does the local IP in our advert reflects it.
___
*[`src/BloomExe/Publish/BloomPub/wifi/WiFiAdvertiser.cs` line 282 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7231#-Oaunw7F8ixrKn-dt62T:-Oc24-B02BpBT9Clnj2k:bkum3mf) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/9e06a5003ae2ce833333b42ad8c7e9d157e08155/src/BloomExe/Publish/BloomPub/wifi/WiFiAdvertiser.cs#L282)):*

If the android code was updated to look for this, could we then make use of IpAddress.Any to listen everywhere?
</blockquote></details>

Yes, IPAddress.Any is already present in this PR for exactly this reason. See BloomReaderUDPListener::ListenForUDPPackages().
___
*[`src/BloomExe/Publish/BloomPub/wifi/WiFiAdvertiser.cs` line 305 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7231#-OaugsMJAnI1fNqBseuu:-Oc2H4bU79pozEWE_dke:bfp2hc9) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/9e06a5003ae2ce833333b42ad8c7e9d157e08155/src/BloomExe/Publish/BloomPub/wifi/WiFiAdvertiser.cs#L305)):*

Would it be cleaner to
- make IfaceWifi and IfaceEthernet local to this method
- return the one we decide should be used?
I think we can then get rid of CommTypeToExpect altogether, and the caller just gets back the InterfaceInfo that should be used (or null if we can't find any).
If I've missed something that needs to know whether it is WiFi or Ethernet, we could make that another field of InterfaceInfo.
</blockquote></details>

I  think you're right, it would be cleaner. I made the change.
___
*[`src/BloomExe/Publish/BloomPub/wifi/WiFiAdvertiser.cs` line 325 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7231#-OauJx_746lXjkXCVbVG:-Oc28ZI1-PSU6v9vbrth:b-sbtj83) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/9e06a5003ae2ce833333b42ad8c7e9d157e08155/src/BloomExe/Publish/BloomPub/wifi/WiFiAdvertiser.cs#L325)):*

I think you can drop this first test by just making the next line use ipProps?.GetIPv4Properties (which automatically returns null if ipPros is null). Maybe even inline ipProps.
</blockquote></details>

Good idea, made the change.

wmergenthal (commented):
I think I addressed all concerns; hope you agree.

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7231#-:-OeJo3J52mB7IelWHkRX:b-9kuhi6)* status: all files reviewed, 5 unresolved discussions (waiting on @JohnThomson)

___
*[`src/BloomExe/Publish/BloomPub/wifi/BloomReaderUDPListener.cs` line 60 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7231#-Oe33uo03-kCnnTnlBoa:-OeHctgf1WYFqLhH-0kR:b-atb2rr) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/93ef15548e2cabbb3533d8d355f8df33f6e94c31/src/BloomExe/Publish/BloomPub/wifi/BloomReaderUDPListener.cs#L60)):*

Let's make this epForBookRequestReceive
</blockquote></details>

Ok, made the change.
___
*[`src/BloomExe/Publish/BloomPub/wifi/WiFiAdvertiser.cs` line 43 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7231#-OauVF5iCnAPqYO-buqm:-OeIr7X6EKralhRPmGU9:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/9e06a5003ae2ce833333b42ad8c7e9d157e08155/src/BloomExe/Publish/BloomPub/wifi/WiFiAdvertiser.cs#L43)):*

Definitely improved, but as you will see elsewhere I still found some things confusing. Congratulations on having figured out this very complex problem, but I think we can still make it a little clearer how it works for future readers (or our future selves).
</blockquote></details>

Done.
___
*[`src/BloomExe/Publish/BloomPub/wifi/WiFiAdvertiser.cs` line 44 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7231#-Oe2u0ly27bHgRQ13FpY:-OeIqr1N1TWuz6kS-cvi:bfq9nu5) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/93ef15548e2cabbb3533d8d355f8df33f6e94c31/src/BloomExe/Publish/BloomPub/wifi/WiFiAdvertiser.cs#L44)):*

It was very unobvious to me, until (I think) I pieced it together from studying the code and comments here and there, why there are two different IP addresses associated with the broadcast. I think it would be helpful to say something like this (if I have it right), which might involve moving these variables so they come together in the file in this order:
_ipForReceivingReplies is the IP address of this device on the network interface we decide to use for sending the advertisement broadcasts. It is used both to create _bcastSourceEP and included in the advertisement to tell recipients where to send book requests.
_bcastSourceEP is an endpoint made from the _ipForReceivingReplies. It gets baked into  _clientForBookAdvertSend and thus determines the network interface that will be used to send the broadcast. Sending from this IP address makes sure that the IP address we put in the message is the correct address to use for book requests on the same subnet.
_ipForSendingBroadcast is an IP address derived from  _ipForReceivingReplies by GetDirectedBroadcastAddress. It doesn't identify a single device but all the devices on the same subnet.
_bcastDestinationEP is an 'endpoint' made from _ipForSendingBroadcast and passed as the destination parameter to the BeginSend call that does the advertizing. This is what makes the Send a broadcast.

I'm going to suggest some more renames, if I have that right.
_ipForReceivingReplies => _ipForThisDeviceOnChosenInterface
_ipForSendingBroadcast  => _fakeIpForBroadcastDestination
_bcastSourceEP => _epForThisDeviceOnChosenInterface
_bcastDestinationEP => epForBroadcastDestination

Do you agree those would be clearer?
</blockquote></details>

I prefer shorter names, but we should use names that *you* find most beneficial. I'm fine with all of them except the one with 'fake' in the name - it is a valid type of address, no fakery involved. I propose _ipBroadcastForDestination and have used that in this update.
___
*[`src/BloomExe/Publish/BloomPub/wifi/WiFiAdvertiser.cs` line 196 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7231#-Oe2p0wN1x2K3802R6pQ:-OeIt58N6Dv1-LsHnPAR:b-9pre9t) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/93ef15548e2cabbb3533d8d355f8df33f6e94c31/src/BloomExe/Publish/BloomPub/wifi/WiFiAdvertiser.cs#L196)):*

Still not quite sure I'm understanding the complexities of this.  GetInterfaceStackWillUse finds the lowest-metric WiFi interface or if there are none, the lowest-metric Ethernet one. That sounds plausible as the one we'd like to use, but if the network stack will actually send on the interface with the absolutely lowest metric, why would it not use an Ethernet interface rather than the WiFi one we want, if it happens to have a lower metric?
But I _think_ the network stack actually has to use the interface that corresponds to the IP address of the endpoint used to initialize _clientForBookAdvertSend. So we just need to make sure that is the right IP address for the desired interface. But that sounds as though the issue is not so much ambiguity about which which IP address to put in the advertisement, but choosing the right one to put in _clientForBookAdvertSend.
Not sure what needs doing here, but see if the confusion I'm expressing after reading this comment block suggests any way to improve it. (The comments and renamings I suggested above may help.)
</blockquote></details>

The network stack *will* use an Ethernet interface if it has a lower metric than an active Wi-Fi interface. And if the PC and Android are on the same subnet, that won't be a problem.
When I test, to "encourage" the stack to choose well I disconnect the ethernet cable. This makes the ethernet interface inactive, and the Wi-Fi interface activates. Plugging the cable back in deactivates the Wi-Fi interface. I don't think Windows will allow both to be active at the same time.
The network stack is going to do what it wants to do. We can't control it. What we *can* control is the IP address we advertise to remote Androids. The mitigation here is to determine that address up front using similar criteria as the network stack, instead of the Dns.GetHostEntry() process. I updated the block comment in attempts to communicate this more clearly.
___
*[`src/BloomExe/Publish/BloomPub/wifi/WiFiAdvertiser.cs` line 233 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7231#-Oe2rupHEysdyIkjlJPj:-OeJ4gt_3oMbgggngA_4:bo1ehq3) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/93ef15548e2cabbb3533d8d355f8df33f6e94c31/src/BloomExe/Publish/BloomPub/wifi/WiFiAdvertiser.cs#L233)):*

I find this baffling. _bcastSouceEP is the endPoint we use to create _clientForBookAdvertSend which is clearly used to SEND the broadcast. So why is it created using _ipForReceivingReplies rather than _ipForSendingBroadcast? I think I finally figured this out, but it would have been easier with the renames I suggested above.
</blockquote></details>

It should be better now with your suggested renames in use (well, 3 of them anyway).

JohnThomson (approved):
> @JohnThomson reviewed 1 of 1 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7231)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @wmergenthal)

PR #7235, opened 2025-06-27
A different placement of the theme info tooltip

JohnThomson:
I also moved the label into the component, which seemed like a logical place for it anyway.

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7235)

### DIFF
--- a/src/BloomBrowserUI/bookEdit/toolbox/games/GameTool.tsx
+++ b/src/BloomBrowserUI/bookEdit/toolbox/games/GameTool.tsx
@@ -1303,14 +1303,6 @@ const DragActivityControls: React.FunctionComponent<{
                         <GameIntroText
                             gameType={getGameType(activityType, getPage())}
                         />
-                        <Div
-                            css={css`
-                                margin-top: 10px;
-                                margin-bottom: 5px;
-                                margin-right: 10px;
-                            `}
-                            l10nKey="EditTab.Toolbox.Games.Theme"
-                        ></Div>
                         <ThemeChooser pageGeneration={props.pageGeneration} />
                         {anyOptions && (
                             <Div
--- a/src/BloomBrowserUI/bookEdit/toolbox/games/ThemeChooser.tsx
+++ b/src/BloomBrowserUI/bookEdit/toolbox/games/ThemeChooser.tsx
@@ -152,61 +152,25 @@ export const ThemeChooser: React.FunctionComponent<{
                     display: flex;
                 `}
             >
-                <Select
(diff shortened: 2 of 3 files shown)


andrew-polk (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7235)* status: all files reviewed, 1 unresolved discussion (waiting on @JohnThomson)

___
*[`src/BloomBrowserUI/react_components/icons/InfoIconUrl.tsx` line 18 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7235#-OTmhE7kEMeHOl1XnGEe:-OTmhE7kEMeHOl1XnGEf:b-rp2m33) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/37750a50dc3126f1205668ef51ca9af6eac7daad/src/BloomBrowserUI/react_components/icons/InfoIconUrl.tsx#L18)):*
> ```XML
>     l10nKey?: string;
> }> = props => {
>     const localizedTip = useL10n("Learn something", props.l10nKey ?? null);
> ```

This dummy "Learn something" string is confusing.
If we really aren't going to use it regardless of whether l10nKey is set, can we set it to empty string? Or at least add a comment stating that?

andrew-polk (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7235)* status: all files reviewed, 1 unresolved discussion (waiting on @JohnThomson)

___
*[`src/BloomBrowserUI/react_components/icons/InfoIconUrl.tsx` line 18 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7235#-OTmhE7kEMeHOl1XnGEe:-OU1X84x0k4wTA0yX7OL:bn6sfvb) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/37750a50dc3126f1205668ef51ca9af6eac7daad/src/BloomBrowserUI/react_components/icons/InfoIconUrl.tsx#L18)):*

This dummy "Learn something" string is confusing.
If we really aren't going to use it regardless of whether l10nKey is set, can we set it to empty string? Or at least add a comment stating that?
</blockquote></details>

Maybe useL10n2 is better because it doesn't require English?
Also, according the comments on both useL10n and useL10n2, passing null as the ID means it will use the English passed in. So the current code, I think, is wrong. `localizedTip` will be "Learn something" if l10nKey is not defined.
But useL10n2 would solve that too, I think.

PR #7240, opened 2025-07-02
Revamp Bloom logging by removing old and adding new events (BL-13484)

StephenMcConnel:
Part of this change is to add explicit logging from typescript code via a new api.

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7240)

### DIFF
--- a/src/BloomBrowserUI/bookEdit/toolbox/toolbox.ts
+++ b/src/BloomBrowserUI/bookEdit/toolbox/toolbox.ts
@@ -827,6 +827,8 @@ async function activateToolInternalAsync(
     // happen after showTool.
     await newTool.showTool();
 
+    postString("logger/writeEvent", `Toolbox activated: ${newTool.id()}`);
+
     // Note: Allowed to begin some async work too, and we will await its result.
     // (This apparently solves the single flash mentioned in BL-10471.)
     await newTool.newPageReady();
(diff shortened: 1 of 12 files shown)


JohnThomson (changes requested):
> Reviewed 12 of 12 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7240)* status: all files reviewed, 2 unresolved discussions (waiting on @StephenMcConnel)

___
*[`src/BloomBrowserUI/publish/PublishTab/PublishTabPane.tsx` line 141 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7240#-OUCIMrD9ulEeiFlLTgT:-OUCIMrD9ulEeiFlLTgU:b-ib20wj) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/5d7817c74565fc7e577aaf61350e058e17731e39/src/BloomBrowserUI/publish/PublishTab/PublishTabPane.tsx#L141)):*
> ```XML
>             "ePUB",
>             "Audio or Video"
>         ];
> ```

Is there any way we could get a useful name from the selected tool itself, so we don't add this array which we are likely to forget to keep in sync as we add tools? (And then pass it instead of idx).
___
*[`src/BloomExe/WebView2Browser.cs` line 81 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7240#-OUCJEDd9rMoC0zEEUHs:-OUCJEDd9rMoC0zEEUHt:bergsnd) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/5d7817c74565fc7e577aaf61350e058e17731e39/src/BloomExe/WebView2Browser.cs#L81)):*
> ```Smalltalk
>                     );
>                     if (!string.IsNullOrEmpty(_stackTopForLog))
>                         Logger.WriteEvent(_stackTopForLog);
> ```

Is there something useful you can comment about why this particular line of the stack is useful to log? Even just an example of the line you are looking for might help.

PR #7254, opened 2025-07-10
Fix WebView2 folder cleanup to handle reused process ids (BL-14984)

StephenMcConnel:
This can be needed for unit tests in some environments.

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7254)

### DIFF
--- a/src/BloomExe/Program.cs
+++ b/src/BloomExe/Program.cs
@@ -1040,7 +1040,7 @@ namespace Bloom
             {
                 if (FileMeddlerManager.IsMeddling)
                     FileMeddlerManager.Stop();
-                CleanupWebView2UserFolders();
+                WebView2Browser.CleanupWebView2UserFolders();
             }
 
             try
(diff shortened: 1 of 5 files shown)


JohnThomson (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7254)* status: all files reviewed, 3 unresolved discussions (waiting on @StephenMcConnel)

___
*[`src/BloomExe/WebView2Browser.Designer.cs` line 43 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7254#-OUqGpF679tKawdd8uJM:-OUqGpF679tKawdd8uJN:b-9m1swc) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/241e0d3ec1e9ae3084d5994b8439426073c9b522/src/BloomExe/WebView2Browser.Designer.cs#L43)):*
> ```Smalltalk
> 					{
> 						if (Directory.Exists(priorUserFolder))
> 							Program.WebView2UserFoldersToDelete.Add(priorUserFolder);
> ```

Can't we delete it here and now, if the process for which it was created is no longer using it? I suppose not, if you've concluded that it isn't safe to delete the current one when disposing of the browser...
___
*[`src/BloomExe/Program.cs` line 1083 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7254#-OUqH0UN8dLe0SxdpESI:-OUqH0UOB9i-zBPIaW8v:b-kq30zp) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/241e0d3ec1e9ae3084d5994b8439426073c9b522/src/BloomExe/Program.cs#L1083)):*
> ```Smalltalk
>         /// </remarks>
>         public static readonly Dictionary<int, string> WebView2ProcessToUserFolder = new Dictionary<int, string>();
>         public static HashSet<string> WebView2UserFoldersToDelete = new HashSet<string>();
> ```

I was badly thrown off by thinking this was all of them, rather than just extras. Maybe WebView2UserFoldersNoLongerUsed? ObsoleteWebView2UserFolders? But see the comment below.
___
*[`src/BloomExe/Program.cs` line 1084 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7254#-OUqHxGK85No3knv64nT:-OUqHxGK85No3knv64nU:bhdm6qx) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/241e0d3ec1e9ae3084d5994b8439426073c9b522/src/BloomExe/Program.cs#L1084)):*
> ```Smalltalk
>         public static readonly Dictionary<int, string> WebView2ProcessToUserFolder = new Dictionary<int, string>();
>         public static HashSet<string> WebView2UserFoldersToDelete = new HashSet<string>();
>         private static void CleanupWebView2UserFolders()
> ```

It feels to me as if this function and the variables it uses would be more at home in the WebView2Browser class, even though the call needs to be in Program.cs.

PR #7255, opened 2025-07-14
Handle pasting into a controlled <input> (BL-14989)

JohnThomson:


andrew-polk (commented):
Did you try doing something like
(in bloomEditing.ts)

```
document.addEventListener("keydown", e => {
    if (e.key === "v" && e.ctrlKey) {

      if (not in a react dialog) { // or we are in a contenteditable div? Or on the book page?

        // If the user pressed Ctrl+V we want to do exactly what the paste button does.
        // We'll invoke it through C#.
        postJson("editView/paste", {});
        e.preventDefault(); // Prevent the default paste behavior.

      }

    }
});
```

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7255#-:-OVA1InZ2KmYo4sK5AKI:b-hqpizu)* status: 0 of 2 files reviewed, all discussions resolved

JohnThomson (commented):
I don't remember exactly how it's done but we already have something intercepting ctrl-V and sending it through C#. That's part of the cause of this problem: C# now has to be able to emulate what the browser used to do for us when ctrl-V is pressed. The function I'm trying to fix is the one that runs when C# determines that paste ought to insert text at the IP in the browser.

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7255#-:-OVDRt4x-z95j8s_L6iS:bde7yfy)* status: 0 of 2 files reviewed, all discussions resolved (waiting on @andrew-polk)

andrew-polk (commented):
Sorry I wasn't clear.
I'm suggesting we try to NOT intercept and handle ctrl-V for this problem case and just go back to letting the browser handle it.
(I added the second if condition into the existing code above.)

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7255#-:-OVDrpa301HIGlLv-QKu:b-dsza02)* status: 0 of 2 files reviewed, all discussions resolved

JohnThomson (commented):
Sorry, I should have realized that. Something like that would probably work for ctrl-V. In fact, I think we only want to override the default behavior if the clipboard contains an image, so that might provide a sufficient criterion for deciding whether to change the default behavior.
However, we do also have a Paste button. Is it OK if that doesn't work in the copyright dialog?

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7255#-:-OVE7Vq29dPuwdsCvojj:buu3lli)* status: 0 of 2 files reviewed, all discussions resolved (waiting on @andrew-polk)

andrew-polk (commented):
A quick test in 6.1 shows it doesn't work already, so I guess so : )
If we can keep from adding this complexity, especially when it feels risky on the verge of 6.2 beta, seems worth it. But I'm happy to discuss.

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7255#-:-OVEAqPCD-Xt0WbZ5is4:b-v6k7od)* status: 0 of 2 files reviewed, all discussions resolved

JohnThomson (commented):
OK, I'll see what I can do.

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7255#-:-OVETz05DGbZtmgALBis:bhfa1ui)* status: 0 of 2 files reviewed, all discussions resolved (waiting on @andrew-polk)

andrew-polk:
Superceded by #7255

PR #7259, opened 2025-07-17
Improved initial audio alignment guess algorithm (BL-14914)

NJStriebel:
After experimentation, changed algorithm to use a variable number of intervals with a fixed duration rather than a fixed number of intervals with a variable duration.

That fixed duration is set at 0.1s.

Those intervals still extend to cover 30% of the phrases on either side of the split in question.

### DIFF
--- a/src/BloomBrowserUI/bookEdit/toolbox/talkingBook/AdjustPauses.ts
+++ b/src/BloomBrowserUI/bookEdit/toolbox/talkingBook/AdjustPauses.ts
@@ -1,0 +1,104 @@
+import { TimedTextSegment } from "./AdjustTimingsControl";
+
+const BREAKSPOT_DURATION = 0.1;
+const MAX_SPLIT_MOVE_PERCENTAGE = 0.3;
+
+//for a framework created to test and optimize this kind of algorithm, feel free to use and modify anything you find at https://github.com/NJStriebel/Spoken-Sentence-Divider
+
+export function adjustPausesToQuietestNearby(
+    initialSegments: TimedTextSegment[],
+    audioData: number[],
+    duration: number
+) {
+    // We're going to fine-tune the segment breaks that we made based on text length.
+    // The idea is to look 30% of the length of the adjacent segments eother side
+    // of the split, and break that into 15 pieces. We'll look for the quietest spot
+    // in that range and move the split to the middle of that quiet spot.
+    // The 30% and 15 pieces are my first guess. They seem to work pretty well but
+    // we might want to fine-tune them.
+    // We might also want to weight things so that similarly quiet spots are preferred
+    // if closer to the original guess. Or we might want to stretch the quiet spot
+    // as far as we can either way without it getting much louder, and take the middle
+    // of that. We might also bias it somehow towards large quiet spots.
+    // Then again, this might be good enough.
+
+    // Look for leading pause. Here the maxAmplitude * 0.1 and the 0.7 limit
+    // are pretty arbitrary. We're assuming the start of what we want will be
+    // at least a tenth as loud as the loudest sound in the file.
+    // And if the first sound is more than 70% of the way through the file,
+    // it seems likely that some very loud sound has confused things, so better
+    // not to adjust at all. This also limits how close together the initial
+    // splits can get as a result of trying to ignore initial silence.
+    let maxAmplitude = 0;
+    for (const amp of audioData) {
+        maxAmplitude = Math.max(maxAmplitude, Math.abs(amp));
+    }
+    const firstSound = audioData.findIndex(
+        amp => Math.abs(amp) > maxAmplitude * 0.1
+    );
+    //adjust is the fraction of the audio that comes after the leading silence
+    let adjust = 1;
+    if (firstSound > 0 && firstSound < audioData.length * 0.7) {
+        adjust = (audioData.length - firstSound) / audioData.length;
+    }
+    const startOfSound = duration * (1 - adjust);
+    //const realDuration = duration * adjust;
+    const adjustTime = (time: number): number => {
+        // We're going to not count the part of the audioData before firstSound.
+        // For example: suppose the first third of the audio is silence,
+        // and the first segment starts half way through the total duration.
+        // We want it instead to start half way through the non-silent final 2/3.
+        return time * adjust + startOfSound;
+    };
+    let segments = initialSegments.map(s => ({
+        start: adjustTime(s.start),
+        end: adjustTime(s.end),
+        text: s.text
+    }));
+    for (let i = 0; i < segments.length - 1; i++) {
+        const seg = segments[i];
+        const mid = (audioData.length * seg.end) / duration;
+        const slop =
+            ((seg.end - seg.start) / duration) *
+            MAX_SPLIT_MOVE_PERCENTAGE *
+            audioData.length;
+        const start = mid - slop;
+        const nextSeg = segments[i + 1];
+        const nextSlop =
+            ((nextSeg.end - nextSeg.start) / duration) *
+            MAX_SPLIT_MOVE_PERCENTAGE *
+            audioData.length;
+        const end = mid + nextSlop;
+
+        const sampleLength = duration / audioData.length;
+        const slopDuration = (end - start) * sampleLength;
+        const numberOfBreaks = slopDuration / BREAKSPOT_DURATION;
+
+        const breakSpots: number[] = [];
+        const breakSpotLength = (end - start) / numberOfBreaks;
+        for (let j = 0; j < numberOfBreaks; j++) {
+            let max = 0;
+            const startBreakSpot = Math.floor(start + j * breakSpotLength);
+            const endBreakSpot = Math.floor(start + (j + 1) * breakSpotLength);
+            for (let k = startBreakSpot; k < endBreakSpot; k++) {
+                max = Math.max(max, Math.abs(audioData[k]));
+            }
+            breakSpots.push(max);
+        }
+        const minBreakSpot = Math.min(...breakSpots);
+        const breakSpot = breakSpots.indexOf(minBreakSpot);
+        const newEnd =
+            (Math.floor(start + (breakSpot + 0.5) * breakSpotLength) *
+                duration) /
+            audioData.length;
+        segments[i].end = newEnd;
+        segments[i + 1].start = newEnd;
+    }
+    // It sometimes looks better if we leave the silence out of the first segment,
+    // but if we guess wrong, there's no way to get the first split before the start of
+    // the first segment. Moreover, until we implement a way to trim the start of the audio,
+    // the first segment actually will include it, so this is also more realistic.
+    segments[0].start = 0;
+
+    return segments;
+}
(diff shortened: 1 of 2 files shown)


NJStriebel (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7259)* status: 1 of 2 files reviewed, 11 unresolved discussions (waiting on @JohnThomson)

___
*[`src/BloomBrowserUI/bookEdit/toolbox/talkingBook/AdjustPauses.ts` line 4 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7259#-OVP6Stp9msXoR4gyC8D:-OVgp4Kp4YQ_95xHXS1I:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/b460e53557ebd85ff878284df939a3e8c86c7aec/src/BloomBrowserUI/bookEdit/toolbox/talkingBook/AdjustPauses.ts#L4)):*

Technically, it's a fraction, not a percentage, since you aren't going to divide it by 100.
Both deserve comments, I think.
Most of our constants use camelCase with a leading k, like kBloomBlueTextBackground. So maybe kIntervalDuration (see below about breakspot) and kMaxSplitMoveFraction?
</blockquote></details>

Done.
___
*[`src/BloomBrowserUI/bookEdit/toolbox/talkingBook/AdjustPauses.ts` line 9 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7259#-OVP1Duh0TG2B2pVmGfi:-OVgq2WJ47W87MIouiwk:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/b460e53557ebd85ff878284df939a3e8c86c7aec/src/BloomBrowserUI/bookEdit/toolbox/talkingBook/AdjustPauses.ts#L9)):*

All these arguments should have comment. It's also good to give a type and comment on what the method returns. For example, without reading the whole method carefully, I don't know whether it is going to modify initialSegments or return a new segment array, or...
</blockquote></details>

Done.
___
*[`src/BloomBrowserUI/bookEdit/toolbox/talkingBook/AdjustPauses.ts` line 18 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7259#-OVP-j3wAXKsPSIikVKc:-OVgq3o-01eXJMxTYdDB:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/b460e53557ebd85ff878284df939a3e8c86c7aec/src/BloomBrowserUI/bookEdit/toolbox/talkingBook/AdjustPauses.ts#L18)):*

The comments above should be updated.
</blockquote></details>

Done.
___
*[`src/BloomBrowserUI/bookEdit/toolbox/talkingBook/AdjustPauses.ts` line 23 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7259#-OVPBXHi-hGRf_CgiGbE:-OVgq4Xx4VTgMTw7sGup:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/b460e53557ebd85ff878284df939a3e8c86c7aec/src/BloomBrowserUI/bookEdit/toolbox/talkingBook/AdjustPauses.ts#L23)):*

The whole block of comments above here is a description of what the whole function you've extracted does. We usually put such comments before the function declaration.
</blockquote></details>

Done.
___
*[`src/BloomBrowserUI/bookEdit/toolbox/talkingBook/AdjustPauses.ts` line 60 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7259#-OVP3eE1-FXiGVmqK7oR:-OVgr32i4dMDyVk85w32:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/b460e53557ebd85ff878284df939a3e8c86c7aec/src/BloomBrowserUI/bookEdit/toolbox/talkingBook/AdjustPauses.ts#L60)):*

I'm going to suggest several new names, even though I came up with the old ones. Reading it a year later, these are not as clear as I'd like. Mid is an index into audioData where our original estimate thought the segment we're adjusting should end. Since the slops are of differing lengths, it's not really the middle of anything! How about originalIndexOfEndOfSegment?
Coming back to this code I find it a struggle to keep track of which numbers are durations in seconds and which represent a range of indices in audioData or an absolute index into it. If you can think of something consistent to do with naming to make it clearer, that would be great. I think we're at least consistent that everything that's a duration is in seconds, but we're not yet consistent in having 'duration' in the names of all such quantities. Maybe we can have duration in the name of everything that's seconds and either index or length in the name of anything that's a number of values in audioData?
Feel free to push back if you think the new names are worse or you come up with something even better.
</blockquote></details>

Done.
___
*[`src/BloomBrowserUI/bookEdit/toolbox/talkingBook/AdjustPauses.ts` line 65 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7259#-OVP4N9800ubivD3xTLM:-OVgr0ka97RL7HR6Tlrd:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/b460e53557ebd85ff878284df939a3e8c86c7aec/src/BloomBrowserUI/bookEdit/toolbox/talkingBook/AdjustPauses.ts#L65)):*

startIndexToSearch?
</blockquote></details>

Done.
___
*[`src/BloomBrowserUI/bookEdit/toolbox/talkingBook/AdjustPauses.ts` line 71 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7259#-OVP4Us48CPiORBZZIR0:-OVgqXGK-0Ejy0thxZ4W:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/b460e53557ebd85ff878284df939a3e8c86c7aec/src/BloomBrowserUI/bookEdit/toolbox/talkingBook/AdjustPauses.ts#L71)):*

endIndexToSearch?
</blockquote></details>

Done.
___
*[`src/BloomBrowserUI/bookEdit/toolbox/talkingBook/AdjustPauses.ts` line 73 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7259#-OVP52Xt0xCmZmP6F1d7:-OVgqJQWB-xsM6Y7rBcK:b-dthc3p) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/b460e53557ebd85ff878284df939a3e8c86c7aec/src/BloomBrowserUI/bookEdit/toolbox/talkingBook/AdjustPauses.ts#L73)):*

I see why I called this sampleLength...it's the time in seconds that is represented by one value or sample in audioData...but I don't think I'm calling them samples anywhere else. Anything better come to your mind? singleValueDuration?
</blockquote></details>

Done. I changed every reference to a single element of the audioData array as a sample. Do you think that's clear enough?
___
*[`src/BloomBrowserUI/bookEdit/toolbox/talkingBook/AdjustPauses.ts` line 75 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7259#-OVP8AjwEqwY0PtT-RK4:-OVgqaH79rG2pIlwEwM2:b-kwnpag) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/b460e53557ebd85ff878284df939a3e8c86c7aec/src/BloomBrowserUI/bookEdit/toolbox/talkingBook/AdjustPauses.ts#L75)):*

It worries me that this is not an integer. The way the loop below is written, if numberOfBreaks is 10.1, j will still be less than 10.1 when it is 10, and we will actually push 11 values into breakSpots, which I think means the last one will extend beyond our calculated slop, since the width of the breakpoints assumes there will somehow be 10.1 of them. It would be pathological if that was beyond the end of audioData...but I think I'd feel safer using ceiling() here. (or floor...but ceiling can't produce zero... )

There's also something of a naming problem here. Are we making breaks or spots or what? Really, they are intervals, one of which we will pick as the best one to put the split in, based on minimizing a score, which currently is the max volume of the interval. How about
numberOfBreaks  -> numberOfIntervals
breakSpots -> intervalScores
breakSpotLength -> intervalLength
startBreakSpot -> startInterval
endBreakSpot -> endInterval
and anything I missed that should change to be consistent.
</blockquote></details>

Done. I don't want to refer to them as just intervals, because we have intervals and segments, which could be interpreted as referring to the same thing. I always refer to those 0.1s splits as "sub intervals." Is that ok, or can you think of a clearer way to make that distinction? Maybe sub-segments?
___
*[`src/BloomBrowserUI/bookEdit/toolbox/talkingBook/AdjustPauses.ts` line 88 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7259#-OVPDfUK8FOkT62sb8oh:-OVgqx7S-J6-Zk21ikyy:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/b460e53557ebd85ff878284df939a3e8c86c7aec/src/BloomBrowserUI/bookEdit/toolbox/talkingBook/AdjustPauses.ts#L88)):*

minScore
</blockquote></details>

Done.
___
*[`src/BloomBrowserUI/bookEdit/toolbox/talkingBook/AdjustPauses.ts` line 89 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7259#-OVPDino9S72PWsxzEkM:-OVgqy6REBOPxYQ85cWo:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/b460e53557ebd85ff878284df939a3e8c86c7aec/src/BloomBrowserUI/bookEdit/toolbox/talkingBook/AdjustPauses.ts#L89)):*

indexOfBestInterval
</blockquote></details>

Done.

NJStriebel (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7259)* status: 1 of 2 files reviewed, 3 unresolved discussions (waiting on @JohnThomson)

___
*[`src/BloomBrowserUI/bookEdit/toolbox/talkingBook/AdjustPauses.ts` line 8 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7259#-OVhGDiA4ie8Lorf8Dyd:-OVhL2aU83lbkm1-1_p6:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/db6fd11ec962a995cc226d30ba154917920c4a56/src/BloomBrowserUI/bookEdit/toolbox/talkingBook/AdjustPauses.ts#L8)):*

I think our code varies about whether to put comments like this at the start of the file or with the function they describe. Especially when the whole file is about one function like this, I like they way you have it. It might be slightly clearer, though, to say "adjustPausesToQuietestNearby" rather than "this method".
</blockquote></details>

Done.
___
*[`src/BloomBrowserUI/bookEdit/toolbox/talkingBook/AdjustPauses.ts` line 10 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7259#-OVhGMfVAifFPfzMFfWk:-OVhL3CYEeaRF3e2cFnL:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/db6fd11ec962a995cc226d30ba154917920c4a56/src/BloomBrowserUI/bookEdit/toolbox/talkingBook/AdjustPauses.ts#L10)):*

We often mark comments like this, that describe possible improvements rather than how the current code works, with a leading "Enhance:"
</blockquote></details>

Done.

PR #7261, opened 2025-07-21
Simplify simpleComprehensionQuiz.ts and its handling (BL-14565)

StephenMcConnel:
This file now only has code for editing quiz pages, and the javascript generated from it is stored in the top level output/browser folder instead of the template folder.

### DIFF
--- a/src/BloomExe/Book/Book.cs
+++ b/src/BloomExe/Book/Book.cs
@@ -3645,16 +3645,7 @@ namespace Bloom.Book
             foreach (SafeXmlElement scriptElt in newPageDiv.SafeSelectNodes(".//script[@src]"))
             {
                 var fileName = scriptElt.GetAttribute("src");
-
-                // TODO BL-14565: this is probably out of date; any such file should be part of Bloom Player shared code.
-
-                // Bloom Desktop accesses simpleComprehensionQuiz.js from the output/browser folder.
-                // Bloom Reader uses the copy of that file which comes with bloom-player.
-                // See https://issues.bloomlibrary.org/youtrack/issue/BL-8480.
-                if (
-                    string.IsNullOrEmpty(fileName)
-                    || fileName == PublishHelper.kSimpleComprehensionQuizJs
-                )
+                if (string.IsNullOrEmpty(fileName))
                     continue;
                 var destinationPath = Path.Combine(FolderPath, fileName);
                 // In other similar operations above we don't overwrite an existing file (e.g., images, css).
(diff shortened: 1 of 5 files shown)


JohnThomson (approved):
LGTM if we really still need the book to reference simpleComprehensionQuiz.js (presumably to make Bloom Player work?)
If Bloom Player is handling this some other way, since the remaining code is only needed in edit mode, seems like we could just get it included in the editing bundle by referencing it somewhere. Would that simplify things even more?

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7261#-:-OVnAOyK96qqFx8W7y-R:bdyiu86)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @StephenMcConnel)

StephenMcConnel (commented):
The issue is to get the initialization code to run when the page is loaded for editing.  It has nothing to do with Bloom Player, and hasn't for a long time.  The reference in the script element both loads and effectively runs the script file.  At the end of the script is the code that actually calls the editing initialization code (which is all that the file now contains).
When Bloom Player loads the page, the script file is not found, and thus nothing happens.  I suppose we could remove this particular script element when creating a bloomPub, but it's not strictly needed since missing script files are gracefully ignored.
If it weren't generated from a pub file, I could stick an HTML comment in to emphasize that the script is to initialize editing state.

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7261#-:-OVnKf3o90xX1JVy-JyW:bw1enf9)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @StephenMcConnel)

JohnThomson (approved):
OK, so Bloom Player doesn't need the book to contain the <script> element. So what I'm suggesting is that we remove it...at the level of the line that calls for it in SimpleComprehensionQuiz.pug. Then, we add to SetupElements (in bloomEditing.ts) the call to initChoiceWidgets(). Either there or in initChoiceWidgets we can check whether the page has class simple-comprehension-quiz and stop if it does not. Importing it and calling it there should automatically pull the code we need into the editing bundle and get it called when the page loads. And all the special cases for this file can go away...it's just another part of the source in bookEdit/js.

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7261#-:-OVnZ8IH8ilVq4LgMBGr:bfnxyot)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @StephenMcConnel)

StephenMcConnel (commented):
Done.

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7261#-:-OVoEUc7DOJlnFdivDev:b-896fix)* status: 0 of 9 files reviewed, all discussions resolved (waiting on @JohnThomson)

PR #7272, opened 2025-07-23
Remove some gratuitous exceptions being thrown (BL-15013)

StephenMcConnel:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7272)

### DIFF
--- a/src/BloomExe/ErrorReporter/ErrorReportUtils.cs
+++ b/src/BloomExe/ErrorReporter/ErrorReportUtils.cs
@@ -61,17 +61,6 @@ namespace Bloom.ErrorReporter
         private static void CheckForFakeTestErrors(string title)
         {
             const string fakeProblemMessage = "Fake problem for development/testing purposes";
-            Exception fakeException;
-
-            // Throwing/catching the exception populates the stack trace
-            try
-            {
-                throw new ApplicationException("Fake exception for development/testing purposes");
-            }
-            catch (ApplicationException e)
-            {
-                fakeException = e;
-            }
 
             if (title == "Error NotifyUser NoReport")
             {
@@ -102,11 +91,13 @@ namespace Bloom.ErrorReporter
             else if (title == "Error NotifyUser Report NoRetry")
             {
                 // Exercises another path through libPalaso directly (goes thru overloads 3, 4)
+                var fakeException = GenerateFakeException();
                 ErrorReport.NotifyUserOfProblem(fakeException, fakeProblemMessage);
             }
             else if (title == "Error NotifyUser Report NoRetry 2")
             {
                 // Exercises a path where you go through BloomErrorReport
+                var fakeException = GenerateFakeException();
                 BloomErrorReport.NotifyUserOfProblem(
                     fakeProblemMessage,
                     fakeException,
@@ -120,6 +111,7 @@ namespace Bloom.ErrorReporter
                     "ErrorReportDialog.Retry",
                     "Retry"
                 );
+                var fakeException = GenerateFakeException();
                 BloomErrorReport.NotifyUserOfProblem(
                     fakeProblemMessage,
                     fakeException,
@@ -148,10 +140,12 @@ namespace Bloom.ErrorReporter
             }
             else if (title == "Error ReportNonFatalException")
             {
+                var fakeException = GenerateFakeException();
                 ErrorReport.ReportNonFatalException(fakeException);
             }
             else if (title == "Error ReportNonFatalExceptionWithMessage")
             {
+                var fakeException = GenerateFakeException();
                 ErrorReport.ReportNonFatalExceptionWithMessage(fakeException, fakeProblemMessage);
             }
             else if (title == "Error ReportNonFatalExceptionWithMessage Scrollbar")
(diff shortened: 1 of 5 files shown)


JohnThomson (changes requested):
Nice work. Just a couple of suggestions. (But I'm really surprised these are the only ones you found. My recollection is that if I turn on "stop for caught exceptions" for all possible types I get heaps of them. Maybe you just meant that these are the only ones easily fixed, or the only ones in our own code.)

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7272#-:-OVxK_Ip0ct6P2yj4UZe:b-95iaqt)* status: all files reviewed, 2 unresolved discussions (waiting on @StephenMcConnel)

___
*[`src/BloomExe/web/BloomServer.cs` line 964 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7272#-OVxJ7kZ7_a6mKR9v0zE:-OVxJ7kZ7_a6mKR9v0zF:b-2mjqk9) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/6149c96386057954824b858c90513971b9a0e22c/src/BloomExe/web/BloomServer.cs#L964)):*
> ```Smalltalk
>                     if (String.IsNullOrEmpty(path))
>                         path = localPath;
>                 }
> ```

It's a bit of a toss-up, since ProcessPath is a vague name and this is the only caller. But I wonder if this code would be better inside the method? We could call it something like TryToFindAFullPathTo (or LookForAFullPathTo, since methods starting with Try have expectations in C#).
It's also somewhat tempting to inline it. Then we don't need a name at all.
___
*[`src/BloomExe/web/BloomServer.cs` line 972 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7272#-OVxHamxEioXT6h0_YmN:-OVxHamyDM_83reXAVI7:b-kjvlqz) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/6149c96386057954824b858c90513971b9a0e22c/src/BloomExe/web/BloomServer.cs#L972)):*
> ```Smalltalk
>                 // Otherwise ignore. Assume this means that this class/method cannot serve that request,
>                 // but something else may.
>                 if (e.Message.StartsWith("Could not locate the required file"))
> ```

Seems like the situations that would result in a message starting with this are now prevented by passing "optional = true" to GetFileDistributedWithApplication, so we don't need this 'if' or what is now a duplicate of the new code above.
I think the insight captured in the first comment above probably belongs with the new use of BloomFileLocator.sTheMostRecentBloomFileLocator?.LocateFile.
That will leave an empty catch, explained by the last two lines of comment above, if you think there are other ApplicationExceptions that might happen in ProcessPath that we want to ignore. I find myself wondering whether we really want to do that. Guessing that this was really about handling "Could not locate the required file" and little if any thought went into the possibility of any other ApplicationExceptions. I'd be a inclined to drop the whole try...catch and see if testing actually reveals any other exceptions we want to ignore.

StephenMcConnel (commented):
I'm not seeing a lot of exceptions myself, which surprised me a bit.  Maybe I'm not setting things up right, or maybe I'm not exercising enough of the program.  Several of the ones that Noel noted down seem to be data-dependent and probably come from outside our codebase.
The primary exception I get is the System.Net.WebException: 'The operation has timed out' when checking the internet connection in UrlLookup.TestInternetConnection().  I'm not sure what to do about that.

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7272#-:-OVyPzh8-oSwOm2TG4j0:b-2xn8r5)* status: all files reviewed, 2 unresolved discussions (waiting on @JohnThomson)

___
*[`src/BloomExe/web/BloomServer.cs` line 964 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7272#-OVxJ7kZ7_a6mKR9v0zE:-OVyQVKR-xFbky8vFJor:btlhuhk) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/6149c96386057954824b858c90513971b9a0e22c/src/BloomExe/web/BloomServer.cs#L964)):*

It's a bit of a toss-up, since ProcessPath is a vague name and this is the only caller. But I wonder if this code would be better inside the method? We could call it something like TryToFindAFullPathTo (or LookForAFullPathTo, since methods starting with Try have expectations in C#).
It's also somewhat tempting to inline it. Then we don't need a name at all.
</blockquote></details>

As far as I can tell from the comments, this code was originally inlined, but was split out to avoid getting another 1000 line function.  I changed the name to "LookForAFullPathToFile".
___
*[`src/BloomExe/web/BloomServer.cs` line 972 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7272#-OVxHamxEioXT6h0_YmN:-OVyRu7y2xmd-4Z_uVOk:b-c7i5d1) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/6149c96386057954824b858c90513971b9a0e22c/src/BloomExe/web/BloomServer.cs#L972)):*

Seems like the situations that would result in a message starting with this are now prevented by passing "optional = true" to GetFileDistributedWithApplication, so we don't need this 'if' or what is now a duplicate of the new code above.
I think the insight captured in the first comment above probably belongs with the new use of BloomFileLocator.sTheMostRecentBloomFileLocator?.LocateFile.
That will leave an empty catch, explained by the last two lines of comment above, if you think there are other ApplicationExceptions that might happen in ProcessPath that we want to ignore. I find myself wondering whether we really want to do that. Guessing that this was really about handling "Could not locate the required file" and little if any thought went into the possibility of any other ApplicationExceptions. I'd be a inclined to drop the whole try...catch and see if testing actually reveals any other exceptions we want to ignore.
</blockquote></details>

I had thought about removing this redundancy since we're not returning/checking null instead of throwing and catching.  I'll remove the try/catch machinery.

PR #7280, opened 2025-07-28
Report more when there are sync problems (BL-15029)

JohnThomson:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7280)


JohnThomson:
Too messed up to save

PR #7292, opened 2025-07-30
Preserve bookshelf setting for expired subscriptions better (BL-15056)

StephenMcConnel:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7292)

### DIFF
--- a/src/BloomExe/Collection/CollectionSettingsDialog.cs
+++ b/src/BloomExe/Collection/CollectionSettingsDialog.cs
@@ -59,12 +59,6 @@ namespace Bloom.Collection
         // Ugly I know, but we need to be able to access these by an index number sometimes.
         internal WritingSystem[] PendingLanguages = new WritingSystem[3];
 
-        /// <summary>
-        /// Client that is launching dialog sets this to try to preserve the bookshelf setting when a
-        /// user updates an expired subscription, but doesn't touch the bookshelf.  (BL-15056)
-        /// </summary>
-        public string ExpiredBookshelf;
-
         public CollectionSettingsDialog(
             CollectionSettings collectionSettings,
             QueueRenameOfCollection queueRenameOfCollection,
@@ -440,6 +434,7 @@ namespace Bloom.Collection
                     _pendingSubscription = null; // not allowed; dialog already explained this
                 else
                 {
+                    _originalSubscription = _collectionSettings.Subscription;
                     _collectionSettings.Subscription = _pendingSubscription;
 
                     if (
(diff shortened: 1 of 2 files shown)


JohnThomson (changes requested):
Sorry, I'm just not getting this fix. Maybe I haven't studied the context enough. Anyway, I'm asking the questions that come to mind.  Maybe you can think of some comment to add that will turn the lights on for me and future editors? Or maybe I just need to study more how this whole dialog and these variables work.

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7292#-:-OWS8tf30c621w0BNcCq:bpebj76)* status: 0 of 1 files reviewed, 3 unresolved discussions

___
*[`src/BloomExe/Collection/CollectionSettingsDialog.cs` line 496 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7292#-OWS5oCI7iScpcYJoSAb:-OWS5oCI7iScpcYJoSAc:b-qfhl3u) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/1fb8ad9aa829490bf3f7ab5e18a22eb465e46624/src/BloomExe/Collection/CollectionSettingsDialog.cs#L496)):*
> ```Smalltalk
>             if (String.IsNullOrEmpty(PendingDefaultBookshelf) &&
>                 !string.IsNullOrEmpty(ExpiredBookshelf) &&
>                 (_pendingSubscription == null ||
> ```

I'm not quite sure how these _pending things work; I gather they are the values showing in the dialog but not yet verified and saved. If that's right, then doesn't _pendingSubscription == null mean no subscription and therefore no bookshelf is allowed? IOW, I'm thinking this line should be more like _pendingSubscription != null &&
___
*[`src/BloomExe/Collection/CollectionSettingsDialog.cs` line 504 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7292#-OWS6XgL5u8emPb-jYOE:-OWS6XgL5u8emPb-jYOF:b-tmf9m) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/1fb8ad9aa829490bf3f7ab5e18a22eb465e46624/src/BloomExe/Collection/CollectionSettingsDialog.cs#L504)):*
> ```Smalltalk
>             _collectionSettings.Save();
>             // This instance of CollectionSettings is reused for the next dialog, so don't show
>             // the old, expired bookshelf value next time.
> ```

This comment makes me expect ExpiredBookshelf to be cleared.
But, it feels wrong that the would always lose an ExpiredBookshelf if they close the dialog. Maybe they needed to quit Bloom and go search for an up-to-date enterprise code. Have they lost their chance to reinstate the expired bookshelf?
Sorry, I'm seeing this mostly out of context, I may be way off track.
___
*[`src/BloomExe/Collection/CollectionSettingsDialog.cs` line 506 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7292#-OWS802y4I-O6_INKn8M:-OWS802y4I-O6_INKn8N:bgk0mzn) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/1fb8ad9aa829490bf3f7ab5e18a22eb465e46624/src/BloomExe/Collection/CollectionSettingsDialog.cs#L506)):*
> ```Smalltalk
>             // the old, expired bookshelf value next time.
>             if (_collectionSettings.DefaultBookshelf != PendingDefaultBookshelf)
>                _collectionSettings.DefaultBookshelf = PendingDefaultBookshelf;
> ```

Is this some private collectionSettings? If it's the main one that everyone uses, it seems very strange to leave it in a state different from the one we saved.

JohnThomson (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7292)* status: all files reviewed, 1 unresolved discussion (waiting on @StephenMcConnel)

___
*[`src/BloomExe/SubscriptionAndFeatures/Subscription.cs` line 334 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7292#-OWXDMKr2lJLmmM9YQnn:-OW_qLHU5xU4pZ8GBW1v:b8l4z7g) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/5d0fd8a79711c3499b33f9ba2b43dd25b9bc5ec6/src/BloomExe/SubscriptionAndFeatures/Subscription.cs#L334)):*

The problem I'm trying to solve (which may not be solvable) is to match the bookshelf tag against the beginning of the code before the date and checksum even if the date and checksum are only partially or incorrectly entered.  Maybe I should restore the original Subscription code and either ignore the problem or try to fix it in the dialog code.
</blockquote></details>

That sounds like good logic, but the comment seems to say the opposite. Shouldn't it be something like, "If the last two parts are numbers but not the right length, assume they are in incorrectly entered date and checksum values, and should NOT be part of the descriptor"?
Or maybe your point is that if the last TWO parts are numbers, we remove them, but if only the last ONE part is a number, we don't remove it?

StephenMcConnel (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7292)* status: all files reviewed, 1 unresolved discussion (waiting on @JohnThomson)

___
*[`src/BloomExe/SubscriptionAndFeatures/Subscription.cs` line 334 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7292#-OWXDMKr2lJLmmM9YQnn:-OWpMv7h9Grsl6ZUsANr:b819u6b) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/5d0fd8a79711c3499b33f9ba2b43dd25b9bc5ec6/src/BloomExe/SubscriptionAndFeatures/Subscription.cs#L334)):*

That sounds like good logic, but the comment seems to say the opposite. Shouldn't it be something like, "If the last two parts are numbers but not the right length, assume they are in incorrectly entered date and checksum values, and should NOT be part of the descriptor"?
Or maybe your point is that if the last TWO parts are numbers, we remove them, but if only the last ONE part is a number, we don't remove it?
</blockquote></details>

This change was reverted.  Subscription.cs should be unchanged.

JohnThomson (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7292)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @StephenMcConnel)

PR #7301, opened 2025-08-04
More fixes related to cropping cover images (BL-15063)

JohnThomson:
- crop when making an image for the Upload dialog
- include metadata in cropped images
- Fix an unrelated problem parsing HTML pixel sizes

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7301)

### DIFF
--- a/src/BloomExe/ImageProcessing/ImageUtils.cs
+++ b/src/BloomExe/ImageProcessing/ImageUtils.cs
@@ -27,6 +27,7 @@ using Bloom.ToPalaso;
 using SIL.CommandLineProcessing;
 using SIL.Windows.Forms.ClearShare;
 using Bloom.ErrorReporter;
+using FFMpegCore.Builders.MetaData;
 using L10NSharp;
 
 namespace Bloom.ImageProcessing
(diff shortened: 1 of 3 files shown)


JohnThomson (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7301)* status: all files reviewed, 2 unresolved discussions (waiting on @andrew-polk)

___
*[`src/BloomExe/web/controllers/PublishApi.cs` line 316 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7301#-OWqqVgJ5Xl5YcBcAG4-:-OWr2gEL-kHXe84UxK7V:brkz6kl) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/5f528f5790d5b663166bfccf430f56ffe811d01b/src/BloomExe/web/controllers/PublishApi.cs#L316)):*

There are a couple things I don't understand about this.

1. It seems to be placing a file in `Path.GetTempPath()`. But the comment says it is in the book folder.
2. Why do we need it after the api call? The current code disposes of the image after sending it to the client. I don't see why we can't also dispose if the image made to create that temporary image.
</blockquote></details>

I was not thinking clearly.
1. I was confusing where it looks for the source file with where it puts the cropped image
2. I still don't know how (or whether) the end-result thumbnail file gets cleaned up, but you are certainly right that once we've made it we no longer need the cropped one.
Fixed.

PR #7303, opened 2025-08-05
Normalize getting collection settings filename (BL-14788)

StephenMcConnel:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7303)

### DIFF
--- a/src/BloomExe/Book/Book.cs
+++ b/src/BloomExe/Book/Book.cs
@@ -5072,7 +5072,7 @@ namespace Bloom.Book
                         var name = Path.GetFileName(path);
                         if (
                             name == "customCollectionStyles.css"
-                            || name.EndsWith(".bloomCollection", StringComparison.Ordinal)
+                            || name.EndsWith(CollectionSettings.kFileExtension, StringComparison.Ordinal)
                         )
                         {
                             //AppendDebugInfo(debugBldr, path);
--- a/src/BloomExe/CLI/GetUsedFontsCommand.cs
+++ b/src/BloomExe/CLI/GetUsedFontsCommand.cs
@@ -58,7 +58,7 @@ namespace Bloom.CLI
                 // Some of this might be useful if we end up needing to instantiate the book to figure out what
                 // is REALLY needed (as opposed to just mentioned in a style sheet).
                 //var collectionFolder = Path.GetDirectoryName(options.BookPath);
-                //var projectSettingsPath = Directory.GetFiles(collectionFolder, "*.bloomCollection").FirstOrDefault();
+                //CollectionSettings.TryGetSettingsFilePath(collectionFolder, out var projectSettingsPath);
                 //var collectionSettings = new CollectionSettings(projectSettingsPath);
 
                 //XMatterPackFinder xmatterFinder = new XMatterPackFinder(new[] {BloomFileLocator.GetInstalledXMatterDirectory()});
--- a/src/BloomExe/Collection/CollectionSettings.cs
+++ b/src/BloomExe/Collection/CollectionSettings.cs
@@ -114,6 +114,64 @@ namespace Bloom.Collection
                 { "Tibetan", "༠༡༢༣༤༥༦༧༨༩" }, // from bo-CN
             };
 
+        public const string kFileExtension = ".bloomCollection";
+        public const string kWildSearchPattern = $"*{kFileExtension}";   // used in directory searches
+
+        /// <summary>
+        /// Generate the file name for the collection settings file given the collection folder name.
+        /// </summary>
+        /// <remarks>
+        /// collectionFolderName is usually the name of the folder.  It can also be the full path to the folder.
+        /// </remarks>
+        public static string GetFileName(string collectionFolderName)
+        {
+            // Avoiding use of Path.ChangeExtension as it's just possible the collectionName could have a period.
+            return $"{Path.GetFileName(collectionFolderName)}{kFileExtension}";
+        }
+
+        /// <summary>
+        /// Generate the path to the collection settings file given the collection folder.
+        /// </summary>
+        public static string GetSettingsFilePath(string collectionFolder)
+        {
+            return Path.Combine(collectionFolder, CollectionSettings.GetFileName(collectionFolder));
+        }
+
+        public static string GetFileDialogFilterString()
+        {
+            return LocalizationManager.GetString(
+                "OpenCreateNewCollectionsDialog.Bloom Collections",
+                "Bloom Collections",
+                "This shows in the file-open dialog that you use to open a different bloom collection"
+            ) + $"|{kWildSearchPattern}";
+        }
+
+        /// <summary>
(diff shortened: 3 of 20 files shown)


JohnThomson (changes requested):
> Reviewed 20 of 20 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7303)* status: all files reviewed, 6 unresolved discussions (waiting on @StephenMcConnel)

___
*[`src/BloomExe/Collection/CollectionSettings.cs` line 118 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7303#-OX03qlY7cFj4IQSIAQv:-OX03qlY7cFj4IQSIAQw:b5qb6lj) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/d0a1f1146d2f2607ba838ef105debc85db1adf77/src/BloomExe/Collection/CollectionSettings.cs#L118)):*
> ```Smalltalk
> 
>         public const string kFileExtension = ".bloomCollection";
>         public const string kWildSearchPattern = "*.bloomCollection";   // used in directory searches
> ```

"*" + kFileExtension?
___
*[`src/BloomExe/Collection/CollectionSettings.cs` line 147 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7303#-OX04Mcd-HIKmwwy8FLb:-OX04Mcd-HIKmwwy8FLc:b-qozukf) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/d0a1f1146d2f2607ba838ef105debc85db1adf77/src/BloomExe/Collection/CollectionSettings.cs#L147)):*
> ```Smalltalk
>                 "Bloom Collections",
>                 "This shows in the file-open dialog that you use to open a different bloom collection"
>             ) + @"|*.bloomLibrary;*.bloomCollection";
> ```

Seems this should use kWildSearchPattern.
Maybe comment what *.bloomLibrary is about? Is this historical?
All the other methods don't consider the possibility that the path might end in this...won't something fail if we actually open a collection whose settings are .bloomLibrary?
Seems like we should either decide it's obsolete and stop searching for them, or have a further fallback that searches for one if we don't find a .bloomCollection. Since it's very easy for a user who has that old a collection to repair, I vote for removing it.
___
*[`src/BloomExe/Collection/CollectionSettings.cs` line 162 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7303#-OX04cF23_U2BhBHtnKD:-OX04cF23_U2BhBHtnKE:b-6z75ed) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/d0a1f1146d2f2607ba838ef105debc85db1adf77/src/BloomExe/Collection/CollectionSettings.cs#L162)):*
> ```Smalltalk
>         {
>             settingsFilePath = Directory.EnumerateFiles(collectionFolder, CollectionSettings.kWildSearchPattern).FirstOrDefault();
>             return settingsFilePath != null && RobustFile.Exists(settingsFilePath);
> ```

How could it not exist, since we selected based on enumerating files that do?
Should we first see whether the result of GetSettingsFilePath() exists, and if so return that? (a) will usually succeed, and probably be faster; (b) if for some reason there are two, probably the correctly named is the one we want?
___
*[`src/BloomExe/Collection/CollectionSettings.cs` line 454 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7303#-OX05fCv4sEv1PTRnHMj:-OX05fCv4sEv1PTRnHMk:b3udgr2) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/d0a1f1146d2f2607ba838ef105debc85db1adf77/src/BloomExe/Collection/CollectionSettings.cs#L454)):*
> ```Smalltalk
>                     // When we're joining a TC, we extract settings in to a temp folder whose name does not
>                     // match the settings file.
>                     if (TryGetSettingsFilePath(collectionFolder, out var realSettingsFile))
> ```

It's not exactly the point of this card, but now we have THREE algorithms for finding the collection settings path
(a) append folder name to folder path
(b) enumerate .bloomCollection files and take the first
(c) Try (a) and if it fails try (b)
I'd be in favor of cutting it down to one public method that uses (c),especially since we have GetPathForNewSettings() to cover what I think is the only legitimate use of (a). (One exception might be uses in testing where we don't need to actually create a file. If that's an issue, maybe we could have an internal GetSettingsPathForTests()).
___
*[`src/BloomExe/Collection/CollectionSettings.cs` line 951 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7303#-OX076_Z0iMyECfANbp1:-OX076_Z0iMyECfANbp2:bpmvrdn) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/d0a1f1146d2f2607ba838ef105debc85db1adf77/src/BloomExe/Collection/CollectionSettings.cs#L951)):*
> ```Smalltalk
>             return parentFolderPath.CombineForPath(
>                 newCollectionName,
>                 CollectionSettings.GetFileName(newCollectionName)
> ```

Doesn't this need to append kFileExtension?
___
*[`src/BloomExe/Collection/CollectionSettings.cs` line 991 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7303#-OX07QG07i5uMWVtlaiv:-OX07QG07i5uMWVtlaiw:brws03p) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/d0a1f1146d2f2607ba838ef105debc85db1adf77/src/BloomExe/Collection/CollectionSettings.cs#L991)):*
> ```Smalltalk
>             {
>                 //we now make a default name based on the name of the directory
>                 string destinationPath = CollectionSettings.GetSettingsFilePath(toDirectory);
> ```

Should probably use GetPathForNewSettings since we're trying to move it to the standard location.

StephenMcConnel (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7303)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @StephenMcConnel)

___
*[`src/BloomExe/Collection/CollectionSettings.cs` line 951 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7303#-OX076_Z0iMyECfANbp1:-OX5Flcy7oHNCvFI6ZZW:b-ibfzmu) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/d0a1f1146d2f2607ba838ef105debc85db1adf77/src/BloomExe/Collection/CollectionSettings.cs#L951)):*

Yes, I think while we're trying to improve this area of our code, I'd like to improve a couple of method names. GetFileName is too similar to Path.GetFileName (and doesn't do the same). And GetSettingsFilePath is too tempting to use when one ought to be using TryGetSettingsFilePath.

What would you think of GetStandardSettingsFileName (for GetFileName) and GetStandardSettingsFilePath (for GetSettingsFilePath)? Or would 'default' be better than 'standard'?
</blockquote></details>

Done.  Standard is fine.

PR #7306, opened 2025-08-05
Add a problem font report for the createArtifacts command (BL-15003)

StephenMcConnel:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7306)

### DIFF
--- a/src/BloomExe/CLI/CreateArtifactsCommand.cs
+++ b/src/BloomExe/CLI/CreateArtifactsCommand.cs
@@ -8,6 +8,7 @@ using System.Threading.Tasks;
 using System.Windows.Forms;
 using Bloom.Book;
 using Bloom.Properties;
+using Bloom.Publish;
 using Bloom.Publish.BloomPub;
 using Bloom.Publish.Epub;
 using Bloom.web;
@@ -28,7 +29,8 @@ namespace Bloom.CLI
         UnhandledException = 1,
         BookHtmlNotFound = 2,
         EpubException = 4,
-        LegacyBookCannotHarvest = 8
+        LegacyBookCannotHarvest = 8,
+        FontProblems = 16, // This is used to indicate that there were font problems that prevent the upload of the book.
     }
 
     class CreateArtifactsCommand
@@ -66,6 +68,11 @@ namespace Bloom.CLI
             {
                 errors.Add(CreateArtifactsExitCode.LegacyBookCannotHarvest.ToString());
                 exitCode &= ~(int)CreateArtifactsExitCode.LegacyBookCannotHarvest;
+            }
+            if ((exitCode & (int)CreateArtifactsExitCode.FontProblems) != 0)
+            {
+                errors.Add(CreateArtifactsExitCode.FontProblems.ToString());
+                exitCode &= ~(int)CreateArtifactsExitCode.FontProblems;
             }
 
             // Check if:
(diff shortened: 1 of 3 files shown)


JohnThomson (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7306)* status: 0 of 3 files reviewed, 2 unresolved discussions

___
*[`src/BloomExe/CLI/CreateArtifactsCommand.cs` line 542 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7306#-OX0FHQR5NfwrPgFPBbC:-OX0FHQR5NfwrPgFPBbD:b-2sj5dp) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/ad33d4f4b6ea47348e0468ab1f34afa46b50a850/src/BloomExe/CLI/CreateArtifactsCommand.cs#L542)):*
> ```Smalltalk
>         [Option(
>             "problemFontsPath",
>             HelpText = "Output destination path for a text file which contains information about problematic fonts that prevent artifact upload",
> ```

I don't think they prevent Upload. They are just displayed as warnings. They are likely to prevent the harvester from creating artifacts, though even that is not guaranteed: the local computer might be missing a requested font (e.g., used in a downloaded book) which the Harvester has and which is fine.
___
*[`src/BloomExe/Publish/PublishHelper.cs` line 1598 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7306#-OX0GLhB1uUScArvNhY9:-OX0GLhB1uUScArvNhYA:b-cow4dz) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/ad33d4f4b6ea47348e0468ab1f34afa46b50a850/src/BloomExe/Publish/PublishHelper.cs#L1598)):*
> ```Smalltalk
>                 var fontNames = new List<string>
>                     {
>                         DefaultFont // Ensure that Andika's metadata is always available.
> ```

I vaguely recall that another Font, something like ABeeCee, is also standard now?

JohnThomson (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7306)* status: all files reviewed, 2 unresolved discussions (waiting on @StephenMcConnel)

JohnThomson (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7306)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @StephenMcConnel)

___
*[`src/BloomExe/CLI/CreateArtifactsCommand.cs` line 542 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7306#-OX0FHQR5NfwrPgFPBbC:-OX4Naw38plq2R6rgM82:bygarxs) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/ad33d4f4b6ea47348e0468ab1f34afa46b50a850/src/BloomExe/CLI/CreateArtifactsCommand.cs#L542)):*

Font problems have prevented upload of bloompub and epub artifacts (including the bloomdigital online reading artifact).  At this point, it appears that unlicensed fonts don't stop uploads, but they will as soon as both the BloomDesktop and bloom-harvester changes are merged and deployed.  Missing fonts are those that are missing on the machine where the command is run, ie, the harvester machine when harvesting occurs.
</blockquote></details>

Ah, I see. I think I missed that this is the CreateArtifactsCommand. I was thinking of Bloom Editor uploading a book to Blorg. I don't think any of these font problems prevents you from uploading the book. But of course when the harvester encounters any of them, they prevent uploading those artifacts. And technically, especially now we're getting the font info in the process of creating the BloomPub, it does only prevent uploading, not prevent them from being created at all.

PR #7308, opened 2025-08-06
Fix cover image transparency for publishing bloompub and epub (BL-15105)

StephenMcConnel:


andrew-polk (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7308)* status: all files reviewed, 3 unresolved discussions (waiting on @StephenMcConnel)

___
*[`src/BloomExe/Publish/BloomPub/BloomPubMaker.cs` line 282 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7308#-OX0HH26D2AYWspM4g1I:-OX0HH271Goo9anEWMLW:bey2pia) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/26cdbc0afc916c965b24a4ec43946450ee20677f/src/BloomExe/Publish/BloomPub/BloomPubMaker.cs#L282)):*
> ```Smalltalk
>                 var div in xmlDom
>                     .SafeSelectNodes(
>                         "//div[contains(concat(' ',@class,' '),' coverColor ')]//div[contains(@class,'bloom-background-image-in-style-attr')][@data-book='coverImage']"
> ```

If I understand correctly, this previously operated on all page with cover color but now only operates on the cover image.
What are the chances we need it to operate on some non-cover-image xmatter image?
___
*[`src/BloomExe/Publish/BloomPub/BloomPubMaker.cs` line 288 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7308#-OX0H_I2Aqa4WBONM2e-:-OX0H_I2Aqa4WBONM2e0:b-m073fx) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/26cdbc0afc916c965b24a4ec43946450ee20677f/src/BloomExe/Publish/BloomPub/BloomPubMaker.cs#L288)):*
> ```Smalltalk
>             {
>                 var style = div.GetAttribute("style");
>                 if (!String.IsNullOrEmpty(style) && style.Contains("background-image:url"))
> ```

I think I better let @JohnThomson look at this. It looks like this could never have worked since we were looking for class names in a style attribute?
___
*[`src/BloomExe/Publish/Epub/EpubMaker.cs` line 2187 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7308#-OX0I8Xe8S2IzmntFiIW:-OX0I8XfDZPlP_eiYyiA:b-rpoq0r) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/26cdbc0afc916c965b24a4ec43946450ee20677f/src/BloomExe/Publish/Epub/EpubMaker.cs#L2187)):*
> ```Smalltalk
>                     var isCoverImage =
>                         img.GetAttribute("data-book") == "coverImage" &&
>                         img.SafeSelectNodes("ancestor::div[contains(concat(' ',@class,' '),' coverColor ')]")
> ```

See comment above about cover image.

PR #7314, opened 2025-08-13
Fix some problems with sentence splitting for audio recording (BL-15111)

JohnThomson:


JohnThomson (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7314)* status: all files reviewed, 2 unresolved discussions (waiting on @andrew-polk)

___
*[`src/BloomBrowserUI/bookEdit/toolbox/talkingBook/audioRecording.ts` line 4104 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7314#-OXixr5_767jWy9Aqfg0:-OXjEahz-wP0erGF68W2:bi0x39s) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/6939c2c63596d38e119a26c8c525f1961bde9198/src/BloomBrowserUI/bookEdit/toolbox/talkingBook/audioRecording.ts#L4104)):*

Probably worth a couple unit tests?
</blockquote></details>

I added one. See what you think.
___
*[`src/BloomBrowserUI/bookEdit/toolbox/talkingBook/audioRecording.ts` line 4114 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7314#-OXiwJfg2F1VltbgRCxL:-OXj6zZs3dTpMW4LOU3z:b-12u304) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/6939c2c63596d38e119a26c8c525f1961bde9198/src/BloomBrowserUI/bookEdit/toolbox/talkingBook/audioRecording.ts#L4114)):*

This part above (with your review comment) does seem very safe and makes sense as an enhancement to stringToSentences.
</blockquote></details>

The thing that makes it less so is that stringToSentences is handling strings, which sometimes are HTML. This patch is operating on a known HTMLElement. To make stringToSentences work the same way, we'd have to parse the HTML and then convert it back (risks there, it might have dangerous characters). Or, we need to change each appropriate caller of stringToSentences to use this function.

andrew-polk (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7314)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @JohnThomson)

___
*[`src/BloomBrowserUI/bookEdit/toolbox/talkingBook/audioRecording.ts` line 4114 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7314#-OXiwJfg2F1VltbgRCxL:-OXjSeWS4VgIC8R1CeWw:bjog559) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/6939c2c63596d38e119a26c8c525f1961bde9198/src/BloomBrowserUI/bookEdit/toolbox/talkingBook/audioRecording.ts#L4114)):*

The thing that makes it less so is that stringToSentences is handling strings, which sometimes are HTML. This patch is operating on a known HTMLElement. To make stringToSentences work the same way, we'd have to parse the HTML and then convert it back (risks there, it might have dangerous characters). Or, we need to change each appropriate caller of stringToSentences to use this function.
</blockquote></details>

Ok, yeah, let's keep it simple.

PR #7315, opened 2025-08-14
Fix reallyCropImage to handle duplicate names and spaces (BL-15121)

JohnThomson:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7315)

### DIFF
--- a/src/BloomExe/Publish/PublishHelper.cs
+++ b/src/BloomExe/Publish/PublishHelper.cs
@@ -1170,28 +1170,94 @@ namespace Bloom.Publish
         )
         {
             var images = bookDom.SafeSelectNodes("//img").Cast<SafeXmlElement>().ToArray();
+            var uncropped = new HashSet<string>();
+            // key is a combination of src, style, and canvas element style height and width that determines
+            // the name of an image file that should be used for this combination, which is the value.
+            var cropped = new Dictionary<string, string>();
             foreach (var img in images)
             {
-                ReallyCropImage(img, imageSourceFolder, imageDestFolder);
+                var src = img.GetAttribute("src");
+                var style = img.GetAttribute("style");
+                if (!SignifiesCropping(style))
+                {
+                    uncropped.Add(src);
+                }
+            }
+
+            foreach (var img in images)
+            {
+                var src = img.GetAttribute("src");
+                var style = img.GetAttribute("style");
+                var imgContainer = img.ParentNode as SafeXmlElement;
+                var canvasElement = imgContainer?.ParentNode as SafeXmlElement;
+                var canvasElementStyle = canvasElement?.GetAttribute("style");
+                if (!SignifiesCropping(style) || canvasElementStyle == null)
+                    continue;
+                var canvasElementWidth = GetNumberFromPx("width", canvasElementStyle);
+                var canvasElementHeight = GetNumberFromPx("height", canvasElementStyle);
+
+                var key = $"{src}|{style}|{canvasElementWidth}|{canvasElementHeight}";
+                if (cropped.TryGetValue(key, out string fileName))
+                {
+                    // This is a duplicate. We can use the same cropped image file.
+                    img.SetAttribute("src", fileName);
+                    continue;
+                }
+
+                var croppedFileName = ReallyCropImage(
+                    img,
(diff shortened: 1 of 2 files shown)


JohnThomson (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7315)* status: all files reviewed, 4 unresolved discussions (waiting on @andrew-polk)

___
*[`src/BloomExe/Publish/PublishHelper.cs` line 1166 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7315#-OXecoMbA3JGBzCy_MZL:-OXhzrd820gYBTg_hfld:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/1fde1e6e2a18f062b74f05a55c139ffd9d402305/src/BloomExe/Publish/PublishHelper.cs#L1166)):*

All this code really wants a home somewhere else like in ImageUtils or something. (I know that isn't new with this change, but each bit of added complexity in the code makes it more obvious.)

But since we are hoping to pull this back at least to 6.2, we don't want to do that now.
Is it worth a follow-up task to do it in master as a separate commit?
</blockquote></details>

Done.
___
*[`src/BloomExe/Publish/PublishHelper.cs` line 1187 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7315#-OXebrIU37QdpHUJ-3nR:-OXi5xUR5b872vC8KFrh:bcc8pcx) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/1fde1e6e2a18f062b74f05a55c139ffd9d402305/src/BloomExe/Publish/PublishHelper.cs#L1187)):*

I think this should iterate through [images - uncropped].
Most images will already be in uncropped at this point.

Then you can also get rid of the `!SignifiesCropping(style)` check below.
</blockquote></details>

Sorry, they are not parallel. I've renamed it and added a comment to try to make it clearer.
___
*[`src/BloomExe/Publish/PublishHelper.cs` line 1247 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7315#-OXeenB45qLdjtZ6G5Ao:-OXi6NUoC5x8GokpSMEc:bb8eqye) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/1fde1e6e2a18f062b74f05a55c139ffd9d402305/src/BloomExe/Publish/PublishHelper.cs#L1247)):*

I guess MakeCroppedImage is statistically guaranteed to produce a unique file name?

Did you consider how much complexity it would add to make `my_original_file_name.jpg`, `my_original_file_name_cropped1.jpg`, `my_original_file_name_cropped2.jpg`, etc.? Probably isn't worth it since this only goes into the published, zipped files.
</blockquote></details>

I believe the chance that a newly generated guid is not unique is considered so minute it is fair to ignore it.
I spent 20 minutes or so trying to get it to use a modified version of the original name, and I'm actually not sure of the right encoding function to make the new src actually locate an arbitrary file name. If you are sure, I have some stashed code I can try again...if it seems worth it. I suppose we might be glad to know image A is a cropped version of image B sometime when we're debugging a BloomPub, but I'm not sure the benefit is worth the effort.
___
*[`src/BloomTests/Publish/PublishHelperTests.cs` line 1324 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7315#-OXegLI273PtomjrZzqE:-OXiGqHF9bIciHbKwvmp:bb9or86) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/1fde1e6e2a18f062b74f05a55c139ffd9d402305/src/BloomTests/Publish/PublishHelperTests.cs#L1324)):*

Maybe worth a comment stating we are purposefully testing a space?
Is it worth including any other characters like `+`?
</blockquote></details>

I added the comment. We're using an otherwise-tested function to do the unencoding, so I think just a name with a space is enough to test that we're using that function in all necessary places.

andrew-polk (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7315)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @JohnThomson)

PR #7325, opened 2025-08-18
Fix various problems with cropping and impairment (BL-15138)

JohnThomson:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7325)

### DIFF
--- a/src/BloomBrowserUI/bookEdit/css/editMode.less
+++ b/src/BloomBrowserUI/bookEdit/css/editMode.less
@@ -1112,7 +1112,7 @@ body.simulateCataracts {
 // the yellow background, not the image.
 body.simulateColorBlindness {
     .bloom-canvas {
-        img {
+        img:not([src="placeHolder.png"]) {
             visibility: hidden;
         }
     }
--- a/src/BloomBrowserUI/bookEdit/css/editMode.less
+++ b/src/BloomBrowserUI/bookEdit/css/editMode.less
@@ -1110,6 +1110,11 @@ body.simulateCataracts {
 // More critically, when simulating BOTH color-blindness AND cataracts,
 // the Canvas gets merged with what is behind it. We want that to be just
 // the yellow background, not the image.
+// Placeholder images (especially background ones) have more complicated
+// visibility rules, so we don't want this one to apply to them.
+// (We decided we don't need the complication of doing color-blindness
+// simulations for placeholders, which don't appear at all in finished
+// publications.)
 body.simulateColorBlindness {
     .bloom-canvas {
         img:not([src="placeHolder.png"]) {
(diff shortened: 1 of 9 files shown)


JohnThomson (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7325)* status: all files reviewed, 3 unresolved discussions (waiting on @andrew-polk)

___
*[`src/BloomBrowserUI/bookEdit/css/editMode.less` line 1115 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7325#-OXzMSoo2Ce0tfBcTGyQ:-OY1vToF9TP8C1jRl8mz:b-lb6gy9) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/8f7c893f0d7e8dc029f4e16e412350a9badad460/src/BloomBrowserUI/bookEdit/css/editMode.less#L1115)):*

I guess I'm not understanding how something works. This change seems backwards from what I would expect.
Seems like this change means placeHolder will NOT be hidden. But I think we always want it hidden.
</blockquote></details>

I've added a comment. Placeholder visibility is complicated. My idea is that their visibility should not be affected by whether we are or are not doing color-blindness simulation.
___
*[`src/BloomBrowserUI/bookEdit/js/CanvasElementManager.ts` line 1505 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7325#-OXzNOK2DXUZnSln-Nmx:-OY1x9T5-huBQis0t4hD:b-k6vdxe) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/8f7c893f0d7e8dc029f4e16e412350a9badad460/src/BloomBrowserUI/bookEdit/js/CanvasElementManager.ts#L1505)):*

I'm a bit unclear about this comment.
Is it a review comment?
Are you still undecided?
</blockquote></details>

Sorry. Yes, I intended it mainly as a review comment. I initially liked the idea of a single place to put everything that needs to happen when an overlay image changes. Then I found that almost every place I want to call it needs only a subset of those things to happen. I wonder a bit if MORE of the places that call notifyToolOfChangedImage should actually call this. There's a method updateCanvasElementForChangedImage which has a suspiciously similar name. But I'm not seeing anything that seems like a definite improvement.
___
*[`src/BloomBrowserUI/bookEdit/toolbox/impairmentVisualizer/impairmentVisualizer.tsx` line 132 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7325#-OXzOCneCnc7cSfUk_2b:-OY2051n2lgvC-qBTNCl:b-ef967k) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/8f7c893f0d7e8dc029f4e16e412350a9badad460/src/BloomBrowserUI/bookEdit/toolbox/impairmentVisualizer/impairmentVisualizer.tsx#L132)):*

I guess you chose to force the caller to specifically send undefined because you wanted the dev to specifically opt in to the unoptimized route if needed?

Either way, I think it probably deserves a comment on this function explaining why img is preferred if available.
</blockquote></details>

I've added the comment. Not sure whether it would have been better to just make the argument optional. There aren't a lot of callers.

PR #7340, opened 2025-08-26
Update csharpier to 1.1.2 (BL-15176)

StephenMcConnel:


andrew-polk (changes requested):
> @andrew-polk reviewed 285 of 285 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7340)* status: all files reviewed, 1 unresolved discussion (waiting on @StephenMcConnel)

___
*[`src/BloomExe/app.config` line 1 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7340#-OYc2iE945VRuLFPX6Ip:-OYc2iE945VRuLFPX6Iq:b-x5o96f) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/278a3c1b5660dc865431b21659f467864aeee043/src/BloomExe/app.config#L1)):*
> ```
> ```

I don't think we want to format the app.config files or the .csproj files since VS often modifies those itself.
Though I could see the argument either way.... but only if we have a pre-commit hook.
In the end, I think I would still lean toward leaving tool-managed files alone.
Let's discuss if you disagree.

StephenMcConnel (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7340)* status: all files reviewed, 1 unresolved discussion (waiting on @andrew-polk)

___
*[`src/BloomExe/app.config` line 1 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7340#-OYc2iE945VRuLFPX6Ip:-OYcCDQQ1dZD46WxCEPJ:bbzeapi) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/278a3c1b5660dc865431b21659f467864aeee043/src/BloomExe/app.config#L1)):*

I don't think we want to format the app.config files or the .csproj files since VS often modifies those itself.
Though I could see the argument either way.... but only if we have a pre-commit hook.
In the end, I think I would still lean toward leaving tool-managed files alone.
Let's discuss if you disagree.
</blockquote></details>

I see your point.  I didn't find an easy way to avoid changing those files if you apply csharpier to a directory as described in the updated README.md file.  But maybe there's something that can be done.  I'll probably have to submit an entirely changed pair of commits, but so be it.

andrew-polk (approved):
> @andrew-polk reviewed 8 of 8 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7340)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @StephenMcConnel)

PR #7342, opened 2025-08-27
BL-14973 .net 8 upgrade

nabalone:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7342)

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -1,0 +1,29 @@
+{
+    // Use IntelliSense to learn about possible attributes.
+    // Hover to view descriptions of existing attributes.
+    // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
+    "version": "0.2.0",
+    "configurations": [
+        {
+            // Use IntelliSense to find out which attributes exist for C# debugging
+            // Use hover for the description of the existing attributes
+            // For further information visit https://github.com/dotnet/vscode-csharp/blob/main/debugger-launchjson.md
+            "name": ".NET Core Launch (console)",
+            "type": "coreclr",
+            "request": "launch",
+            "preLaunchTask": "build",
+            // If you have changed target frameworks, make sure to update the program path.
+            "program": "${workspaceFolder}/output/Debug/Bloom.dll",
+            "args": [],
+            "cwd": "${workspaceFolder}/src/BloomExe",
+            // For more information about the 'console' field, see https://aka.ms/VSCode-CS-LaunchJson-Console
+            "console": "internalConsole",
+            "stopAtEntry": false
+        },
+        {
+            "name": ".NET Core Attach",
+            "type": "coreclr",
+            "request": "attach"
+        }
+    ]
+}
--- a/.vscode/tasks.json
+++ b/.vscode/tasks.json
@@ -1,0 +1,41 @@
+{
+    "version": "2.0.0",
+    "tasks": [
+        {
+            "label": "build",
+            "command": "dotnet",
+            "type": "process",
+            "args": [
+                "build",
+                "${workspaceFolder}/Bloom.sln",
+                "/property:GenerateFullPaths=true",
+                "/consoleloggerparameters:NoSummary;ForceNoAlign"
+            ],
+            "problemMatcher": "$msCompile"
+        },
+        {
+            "label": "publish",
+            "command": "dotnet",
+            "type": "process",
+            "args": [
+                "publish",
+                "${workspaceFolder}/Bloom.sln",
+                "/property:GenerateFullPaths=true",
+                "/consoleloggerparameters:NoSummary;ForceNoAlign"
+            ],
+            "problemMatcher": "$msCompile"
+        },
+        {
+            "label": "watch",
+            "command": "dotnet",
+            "type": "process",
+            "args": [
+                "watch",
+                "run",
+                "--project",
+                "${workspaceFolder}/Bloom.sln"
+            ],
+            "problemMatcher": "$msCompile"
+        }
+    ]
+}
--- a/DistFiles/localization/en/Bloom.xlf
+++ b/DistFiles/localization/en/Bloom.xlf
@@ -2039,6 +2039,7 @@
       <trans-unit id="EditTab.TextBoxProperties.BordersAndBackground" sil:dynamic="true">
         <source xml:lang="en">Borders &amp; Background</source>
         <note>ID: EditTab.TextBoxProperties.BordersAndBackground</note>
+        <note>&amp; is an abbreviation symbol for "and"</note>
       </trans-unit>
       <trans-unit id="EditTab.TextBoxProperties.CollectionLanguage2" sil:dynamic="true">
         <source xml:lang="en">Collection Language 2</source>
--- a/ReadMe.md
+++ b/ReadMe.md
@@ -31,6 +31,10 @@ yarn
 
 #### Build the .NET part
 
+##### In VS Code:
+Make sure you have the C# Devkit extensions installed. Run and Debug through the gui or F5.
+
+##### In Visual Studio:
 1. Open Bloom.sln in Visual Studio
 2. Build the "WebView2PdfMaker" project
 3. Run the "BloomExe" project
--- a/build/getDependencies-windows.sh
+++ b/build/getDependencies-windows.sh
@@ -102,16 +102,9 @@ cd -
 #     URL: https://build.palaso.org/buildConfiguration/bt54?mode=builds
 #     clean: false
 #     revision: latest.lastSuccessful
-#     paths: {"PdfDroplet.exe"=>"lib/dotnet", "PdfSharp.dll"=>"lib/dotnet"}
+#     paths: {"PdfDroplet.dll"=>"lib/dotnet", "PdfSharp.dll"=>"lib/dotnet"}
 #     VCS: https://github.com/sillsdev/pdfDroplet [master]
-# [4] build: TidyManaged-master-win32-continuous (bt349)
-#     project: TidyManaged
-#     URL: https://build.palaso.org/buildConfiguration/bt349?mode=builds
-#     clean: false
-#     revision: latest.lastSuccessful
-#     paths: {"*.*"=>"lib/dotnet"}
-#     VCS: https://github.com/BloomBooks/TidyManaged.git [master]
-# [5] build: Windows master continuous (XliffForHtml_WindowsMasterContinuous)
+# [4] build: Windows master continuous (XliffForHtml_WindowsMasterContinuous)
 #     project: XliffForHtml
 #     URL: https://build.palaso.org/buildConfiguration/XliffForHtml_WindowsMasterContinuous?mode=builds
 #     clean: false
@@ -143,10 +136,8 @@ copy_auto https://build.palaso.org/guestAuth/repository/download/Bloom_PortableD
 copy_auto https://build.palaso.org/guestAuth/repository/download/Bloom_PortableDevicesFromPodcastUtitlies/latest.lastSuccessful/Interop.PortableDeviceApiLib.dll ../lib/dotnet/Interop.PortableDeviceApiLib.dll
 copy_auto https://build.palaso.org/guestAuth/repository/download/Bloom_PortableDevicesFromPodcastUtitlies/latest.lastSuccessful/Interop.PortableDeviceTypesLib.dll ../lib/dotnet/Interop.PortableDeviceTypesLib.dll
 copy_auto https://build.palaso.org/guestAuth/repository/download/Bloom_Help_BloomHelp63/latest.lastSuccessful/Bloom.chm ../DistFiles/Bloom.chm
-copy_auto https://build.palaso.org/guestAuth/repository/download/bt54/latest.lastSuccessful/PdfDroplet.exe ../lib/dotnet/PdfDroplet.exe
+copy_auto https://build.palaso.org/guestAuth/repository/download/bt54/latest.lastSuccessful/PdfDroplet.dll ../lib/dotnet/PdfDroplet.dll
 copy_auto https://build.palaso.org/guestAuth/repository/download/bt54/latest.lastSuccessful/PdfSharp.dll ../lib/dotnet/PdfSharp.dll
-copy_auto https://build.palaso.org/guestAuth/repository/download/bt349/latest.lastSuccessful/TidyManaged.dll ../lib/dotnet/TidyManaged.dll
-copy_auto https://build.palaso.org/guestAuth/repository/download/bt349/latest.lastSuccessful/TidyManaged.dll.config ../lib/dotnet/TidyManaged.dll.config
 copy_auto https://build.palaso.org/guestAuth/repository/download/bt349/latest.lastSuccessful/libtidy.dll ../lib/dotnet/libtidy.dll
 copy_auto https://build.palaso.org/guestAuth/repository/download/XliffForHtml_WindowsMasterContinuous/latest.lastSuccessful/HtmlXliff.exe ../lib/dotnet/HtmlXliff.exe
 copy_auto https://build.palaso.org/guestAuth/repository/download/XliffForHtml_WindowsMasterContinuous/latest.lastSuccessful/HtmlXliff.pdb ../lib/dotnet/HtmlXliff.pdb
--- a/debian/copyright
+++ b/debian/copyright
@@ -131,10 +131,6 @@ Files: /usr/lib/bloom-desktop-alpha/PdfSharp.dll
 Copyright: 2005-2014 empira Software GmbH, Troisdorf (Germany)
 License: MIT
 
-Files: /usr/lib/bloom-desktop-alpha/TidyManaged.dll*
-Copyright: 2009 Mark Beaton
-License: MIT
-
 Files: /usr/lib/bloom-desktop-alpha/DistFiles/pdf/*
 Copyright: 2012 Mozilla Foundation
 License: Apache-2.0
--- a/src/BloomBrowserUI/bookEdit/TextBoxProperties/TextBoxProperties.pug
+++ b/src/BloomBrowserUI/bookEdit/TextBoxProperties/TextBoxProperties.pug
@@ -36,7 +36,7 @@
 					#align-bottom.icon16x16.propButton
 						img(src="/bloom/bookEdit/img/align-bottom.png")
 		.tab-page
-			h2.tab(data-i18n="EditTab.TextBoxProperties.BordersAndBackground") Borders & Background
+			h2.tab(data-i18n="EditTab.TextBoxProperties.BordersAndBackground") Borders &amp; Background
 			.mainBlock.leftBlock
 				#borderstyle-group
 					.group-label(data-i18n="EditTab.TextBoxProperties.BorderStyle") Border Style
--- a/src/BloomBrowserUI/react_components/aboutDialog.tsx
+++ b/src/BloomBrowserUI/react_components/aboutDialog.tsx
@@ -480,15 +480,6 @@ export const AboutDialog: React.FunctionComponent<{
                         </li>
                         <li>
                             <a
-                                href="https://github.com/markbeaton/TidyManaged"
-                                target="_blank"
-                                rel="noreferrer"
-                            >
-                                TidyManaged
-                            </a>
-                        </li>
-                        <li>
-                            <a
                                 href="https://github.com/CodeSeven/toastr"
                                 target="_blank"
                                 rel="noreferrer"
--- a/src/BloomExe/BloomExe.csproj
+++ b/src/BloomExe/BloomExe.csproj
@@ -1,7 +1,9 @@
-<Project Sdk="Microsoft.NET.Sdk">
+<Project Sdk="Microsoft.NET.Sdk.WindowsDesktop">
 	<PropertyGroup>
 		<ApplicationManifest>app.manifest</ApplicationManifest>
 		<ProjectGuid>{304D5612-167C-4725-AF27-B9F2BB788B57}</ProjectGuid>
+		<UseWindowsForms>true</UseWindowsForms>
+		<UseWPF>false</UseWPF>
 		<SolutionDir Condition="$(SolutionDir) == '' Or $(SolutionDir) == '*Undefined*'">..\..\</SolutionDir>
 		<OutputPath>..\..\output\$(Configuration)\</OutputPath>
 		<Platform Condition=" '$(Platform)' == '' ">x86</Platform>
@@ -12,7 +14,7 @@
 		<Version>6.3.0.0</Version>
 		<IncludeSourceRevisionInInformationalVersion>false</IncludeSourceRevisionInInformationalVersion>
 		<BuildDependsOn>$(BuildDependsOn);AfterBuildMigrated</BuildDependsOn>
-		<TargetFramework>net472</TargetFramework>
+		<TargetFramework>net8.0-windows</TargetFramework>
 		<LangVersion>11.0</LangVersion>
 		<PostBuildEvent>
 copy /Y "$(SolutionDir)\lib\msvcr120.dll" $(OutDir)
@@ -21,7 +23,7 @@ copy /Y "$(SolutionDir)\lib\dotnet\meddle.exe" $(OutDir)
 xcopy "$(SolutionDir)lib\RedistributableDlls" $(OutDir) /S /R /Y
 xcopy "$(SolutionDir)lib\gm" "$(OutDir)gm\" /S /R /Y
 copy /Y "$(SolutionDir)\lib\dotnet\libtidy.dll" $(OutDir)
-    </PostBuildEvent>
+		</PostBuildEvent>
 		<GenerateAssemblyInfo>true</GenerateAssemblyInfo>
 		<AppendTargetFrameworkToOutputPath>false</AppendTargetFrameworkToOutputPath>
 	</PropertyGroup>
@@ -134,6 +136,10 @@ copy /Y "$(SolutionDir)\lib\dotnet\libtidy.dll" $(OutDir)
 	<PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Release|AnyCPU'">
 		<Prefer32Bit>True</Prefer32Bit>
 	</PropertyGroup>
+	<PropertyGroup>
+		<!-- We should delete this when we get rid of WebThumbNailList.Designer.cs. I believe we are only using binary formatter serialization in working with the ImageStream there. -->
+		<EnableUnsafeBinaryFormatterSerialization>true</EnableUnsafeBinaryFormatterSerialization>
+	</PropertyGroup>
 	<ItemGroup>
 		<SourceFiles Include="..\..\lib\dotnet\*.config" />
 		<LameFiles Include="..\..\lib\lame\lame*.*" />
@@ -187,6 +193,7 @@ copy /Y "$(SolutionDir)\lib\dotnet\libtidy.dll" $(OutDir)
 		<PackageReference Include="EPPlus" Version="6.2.4" />
 		<PackageReference Include="Fleck" Version="1.1.0" />
 		<PackageReference Include="Glob" Version="1.1.8" />
+		<PackageReference Include="HtmlAgilityPack" Version="1.12.1" />
 		<PackageReference Include="L10NSharp" Version="8.0.0" />
 		<PackageReference Include="LargeAddressAware" Version="1.0.5" />
 		<PackageReference Include="Markdig.Signed" Version="0.37.0" />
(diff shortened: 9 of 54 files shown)


nabalone (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7342)* status: 30 of 50 files reviewed, 17 unresolved discussions (waiting on @andrew-polk)

___
*[`src/BloomExe/ClassDiagram1.cd` line 225 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7342#-OYgZ06yDjjXUjZhO7Kn:-OYgZ06zEzUHGeuMn4v_:bug20j2) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/cfae803e8bf0a0c6f72d928f82c3079825ffbce1/src/BloomExe/ClassDiagram1.cd#L225)):*
> ```
> 	<Lollipop Position="0.2" Collapsed="true" />
>   </Class>
>   <Class Name="Bloom.Shortcut" Collapsed="true">
> ```

I renamed ResolveShortcut to Shortcut throughout because it is a class that has a resolve method but also other code (interfaces) that I am now reusing elsewhere
___
*[`src/BloomExe/Program.cs` line 2052 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7342#-OYgb8Pj2LoSW2nstc_1:-OYgb8Pj2LoSW2nstc_2:b-51armw) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/cfae803e8bf0a0c6f72d928f82c3079825ffbce1/src/BloomExe/Program.cs#L2052)):*
> ```Smalltalk
>                     if (
>                         name != null
>                         && name.StartsWith("testhost", StringComparison.OrdinalIgnoreCase)
> ```

There are other things you can check for too, but this was enough locally
___
*[`src/BloomExe/Shortcut.cs` line 50 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7342#-OYgbj-7D3O6CW3JMMuJ:-OYgbj-7D3O6CW3JMMuK:b2mx01) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/cfae803e8bf0a0c6f72d928f82c3079825ffbce1/src/BloomExe/Shortcut.cs#L50)):*
> ```Smalltalk
> 
>         [Flags()]
>         public enum SLR_FLAGS
> ```

now used in ShortcutMaker.cs. Also below
___
*[`src/BloomExe/Publish/Video/RecordVideoWindow.cs` line 1348 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7342#-OYgechW7T5yl0dI0hHy:-OYgechXCVjwaVvU2Frq:b-6zzuid) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/cfae803e8bf0a0c6f72d928f82c3079825ffbce1/src/BloomExe/Publish/Video/RecordVideoWindow.cs#L1348)):*
> ```Smalltalk
>             _webSocketServer.SendString("recordVideo", "recording", "false");
>             base.OnClosed(e);
>             if (_recording && _ffmpegProcess != null && !_ffmpegExited)
> ```

In the happy path, we hit this code after other code has already written the "q" and stopped ffmpeg. But the _ffmpegProcess is still non-null. But the new behavior in .net 8 is that if we now try to write a 'q' again, we get a pipe closing error (that can kill Bloom).

I'm now thinking about whether it would be better to set _ffmpegProcess to null here: https://github.com/BloomBooks/BloomDesktop/blob/6fc41e200b8c67630a16de1a5da365087e9dbe1c/src/BloomExe/Publish/Video/RecordVideoWindow.cs#L469 We would need to make sure we don't need to get anything out of the exited process object afterwards though
___
*[`src/BloomExe/Collection/ShortcutMaker.cs` line 21 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7342#-OYgdbofEd3TZSDZMtxW:-OYgdbofEd3TZSDZMtxX:b-bxhrv3) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/cfae803e8bf0a0c6f72d928f82c3079825ffbce1/src/BloomExe/Collection/ShortcutMaker.cs#L21)):*
> ```Smalltalk
>             int shortPathLength
>         );
> ```

The changes in this file were mostly made by copilot
___
*[`src/BloomExe/RobustFileIO.cs` line 8 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7342#-OYgbdm_7n7yYahHaP4k:-OYgbdm_7n7yYahHaP4l:byf4otu) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/6fc41e200b8c67630a16de1a5da365087e9dbe1c/src/BloomExe/RobustFileIO.cs#L8)):*
> ```Smalltalk
> using TidyManaged;
> #if !__MonoCS__
> using NAudio.Wave;
> ```

was unused
___
*[`src/BloomExe/RobustFileIO.cs` line 25 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7342#-OYgbZgF3UC3iEogZ4U7:-OYgbZgF3UC3iEogZ4U8:bcduuc) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/6fc41e200b8c67630a16de1a5da365087e9dbe1c/src/BloomExe/RobustFileIO.cs#L25)):*
> ```Smalltalk
>     public class RobustFileIO
>     {
>         public static Document DocumentFromFile(string filePath)
> ```

This is tidymanaged. I have dleeted the only usage
___
*[`src/BloomExe/Publish/Video/PublishAudioVideoAPI.cs` line 55 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7342#-OYge4NQ7ZgvMDBhhbh4:-OYge4NQ7ZgvMDBhhbh5:b6vankw) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/cfae803e8bf0a0c6f72d928f82c3079825ffbce1/src/BloomExe/Publish/Video/PublishAudioVideoAPI.cs#L55)):*
> ```Smalltalk
>                 {
>                     var soundLog = request.RequiredPostJson();
>                     await _recordVideoWindow?.StopRecordingAsync(soundLog);
> ```

If the user closes the recording window, which triggers logic to stop the recording, but then  while that's happening the video ends triggering this StopRecording call, we can get into a state where _recordVideoWindow is null when we hit this line so it throws an error
___
*[`src/BloomExe/Edit/EditingModel.cs` line 347 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7342#-OYgduWTCwl-kPJduPrh:-OYgduWUEYwpdWFQk5p0:b-uc3sls) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/6fc41e200b8c67630a16de1a5da365087e9dbe1c/src/BloomExe/Edit/EditingModel.cs#L347)):*
> ```Smalltalk
> 
>             SaveCustomizedCssRules(dom, userCssContent);
>             XmlHtmlConverter.ThrowIfHtmlHasErrors(dom.OuterXml);
> ```

I think it was JT who looked at this with me and thought it was okay to delete (not easy to replicate without TidyManaged)

JohnThomson (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7342)* status: 48 of 50 files reviewed, 34 unresolved discussions (waiting on @andrew-polk and @nabalone)

___
*[`src/BloomExe/XmlHtmlConverter.cs` line 18 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7342#-OYhNWwM29f48tRUeIcl:-OYpzzaJCmVTsdRBzlvZ:b7woumj) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/64e6167661c72d3e9f16c1b2b3e99fb4f615e4b8/src/BloomExe/XmlHtmlConverter.cs#L18)):*

I'm unsure about this name.
I wasn't expecting it to return a string but rather an object (like XmlDocument).

But I'm also not coming up with any great names.
`Get/CreateValidHtmlFromContent`
`Get/CreateValidHtmlDocumentAsString`
`Get/CreateValidHtmlString`
`Get/CreateHtmlString`
</blockquote></details>

Few thoughts:
- it's awkward that the optional headContent argument comes second, when the head comes before the body. Of course it has to if it's optional, but you could also just do two overloads, a single-argument one that takes a bodyContent, and a two-argument one that takes headContent, bodyContent.
- why do we want this to be a single method with an optional argument? What about something like WrapBodyContentAsHtml and WrapHeadAndBodyContentAsHtml? Or if you want the same name, how about WrapAsHtml?
- it needs to be clear, at least in comments, whether the arguments and result are XML or HTML (or maybe it works either way, with the output being whatever the input is?)
- it doesn't quite feel "at home" to me in this class, since it has nothing obvious to do with converting between XML and HTML. Would it fit better somewhere like XmlUtils? Or even in a test class, if it's only used for testing. Then again, it seems to be only used in this class an in making test data to pass to this class, so maybe this is reasonable home. Up to you.
___
*[`src/BloomExe/Edit/EditingModel.cs` line 347 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7342#-OYgduWTCwl-kPJduPrh:-OYqF6aVET05hoEJWg18:bhaxzwv) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/6fc41e200b8c67630a16de1a5da365087e9dbe1c/src/BloomExe/Edit/EditingModel.cs#L347)):*

Maybe I'm missing something, but it seems like this is a safeguard in case we ever do get into a bad situation. Don't we want to keep this? And especially now since we're changing the creator?

Is the idea that HtmlAgilityPack can't give us a bad dom?
Still seems like it is worth keeping the check in place. We have our own code between HAP and this.
</blockquote></details>

I think we mainly wanted this because Tidy cleaned things up so aggressively that we would tend to miss errors in the html. Even with Tidy, it would probably have been more efficient to catch any errors during the original parse in GetXmlDomFromHtml(). Now, we're starting with HTML that is generated from an element in the WebView2 Dom, so it's very likely to be good HTML, unless we're somehow doing a bad patch to it before returning it to C#. We've already parsed it as HTML and converted it to XML using the new package, which I think is much more likely than Tidy to throw if the HTML is bad, and then parsed the XML into a Dom, which does a good deal of validation on the output. I'm not sure what possible errors could be caught by now converting it back to a string and then parsing it again, but I don't recall that the old check ever actually caught anything. If HtmlAgilityPack has any settings that make it stricter about validating HTML, maybe we should turn them on during the call above to GetXmlDomFromHtml(), but otherwise, I'm not sure what we would want to look for here or how we would do it, and I don't think we're losing much by just dropping it.
___
*[`src/BloomExe/Publish/Video/RecordVideoWindow.cs` line 1348 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7342#-OYgechW7T5yl0dI0hHy:-OYqMgZy3OJo051FqoeN:bqnjw9a) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/cfae803e8bf0a0c6f72d928f82c3079825ffbce1/src/BloomExe/Publish/Video/RecordVideoWindow.cs#L1348)):*

In the happy path, we hit this code after other code has already written the "q" and stopped ffmpeg. But the _ffmpegProcess is still non-null. But the new behavior in .net 8 is that if we now try to write a 'q' again, we get a pipe closing error (that can kill Bloom).

I'm now thinking about whether it would be better to set _ffmpegProcess to null here: https://github.com/BloomBooks/BloomDesktop/blob/6fc41e200b8c67630a16de1a5da365087e9dbe1c/src/BloomExe/Publish/Video/RecordVideoWindow.cs#L469 We would need to make sure we don't need to get anything out of the exited process object afterwards though
</blockquote></details>

I don't recall whether we set _ffmpegExited right when we send the 'q' or after we get some confirmation that it actually exited. If the latter, maybe we should have a new flag, something like "_attemptedToStopFfmpeg" that is set true right before we send the 'q'? That seems simpler than examining all our logic to see if we might need the process object after we asked it to quit.
___
*[`src/BloomTests/Book/BookTestsBase.cs` line 281 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7342#-OYqUuDh-iHT2I0CheDc:-OYqUuDh-iHT2I0CheDd:bj6dcsp) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/64e6167661c72d3e9f16c1b2b3e99fb4f615e4b8/src/BloomTests/Book/BookTestsBase.cs#L281)):*
> ```Smalltalk
>         protected virtual Bloom.Book.Book CreateBookWithPhysicalFile(
>             string bodyContent,
>             string headContent = "",
> ```

This is a different default from CreateDocumentWithContent and will produce an empty head, I believe. Is that intentional?
___
*[`src/BloomTests/XmlHtmlConverterTests.cs` line 479 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7342#-OYqQoJc0UYO4nvVjtYz:-OYqQoJc0UYO4nvVjtZ-:bwrp34) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/64e6167661c72d3e9f16c1b2b3e99fb4f615e4b8/src/BloomTests/XmlHtmlConverterTests.cs#L479)):*
> ```Smalltalk
>             var dom = XmlHtmlConverter.GetXmlDomFromHtml(html);
>             var xml = dom.DocumentElement.GetElementsByTagName("style")[0].InnerXml;
>             // Trim because it may mess with leading or trailing white space in ways we don't care about.
> ```

The comment seems wrong. Trim has a more important function. The test is about making sure that GetElementsByTagName removes trailing blank lines, and Trim makes sure that the expected result doesn't have trailing blank lines. Also, I don't see anything here that's forgiving about leading or trailing white space we don't care about. xml must absolutely have none for it to pass.
___
*[`src/BloomTests/XmlHtmlConverterTests.cs` line 539 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7342#-OYqRoZa8nYbALeqWQOd:-OYqRoZa8nYbALeqWQOe:bsn5iqh) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/64e6167661c72d3e9f16c1b2b3e99fb4f615e4b8/src/BloomTests/XmlHtmlConverterTests.cs#L539)):*
> ```Smalltalk
>                 Does.Contain("data-foo=\"bar\" data-baz=\"qux\""),
>                 "The remaining attributes are present."
>             );
> ```

What if the two values for data-foo are different? Do we care which it keeps? Or should that produce an error?
___
*[`src/BloomExe/XmlHtmlConverter.cs` line 79 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7342#-OYq2YiJ4uQ1ooluKpM3:-OYq2YiJ4uQ1ooluKpM4:b5gaaul) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/64e6167661c72d3e9f16c1b2b3e99fb4f615e4b8/src/BloomExe/XmlHtmlConverter.cs#L79)):*
> ```Smalltalk
>                     RegexOptions.Singleline | RegexOptions.IgnoreCase
>                 )
>             )
> ```

- Could the initial comment span more than one line? Like

<html>....
___
*[`src/BloomExe/XmlHtmlConverter.cs` line 112 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7342#-OYq4VtV86M193B8SjOD:-OYq4VtW8dc1oLE79REr:b-h169x9) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/64e6167661c72d3e9f16c1b2b3e99fb4f615e4b8/src/BloomExe/XmlHtmlConverter.cs#L112)):*
> ```Smalltalk
>             doc.LoadHtml(content);
> 
>             string newContents = doc.DocumentNode.OuterHtml;
> ```

This is now XML, right, because of OptionOutputAsXml? Given that OuterHtml suggests the opposite, might be worth using a name like xmlContents or newXml instead of newContents.
___
*[`src/BloomExe/XmlHtmlConverter.cs` line 114 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7342#-OYq5f3F8qRsSNfln0Pu:-OYq5f3F8qRsSNfln0Pv:bkuzwa4) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/64e6167661c72d3e9f16c1b2b3e99fb4f615e4b8/src/BloomExe/XmlHtmlConverter.cs#L114)):*
> ```Smalltalk
>             string newContents = doc.DocumentNode.OuterHtml;
>             var firstChild =
>                 doc.DocumentNode.ChildNodes.Count > 0 ? doc.DocumentNode.ChildNodes[0] : null;
> ```

FirstOrDefault?
___
*[`src/BloomExe/XmlHtmlConverter.cs` line 153 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7342#-OYq6TVW3mDIbWMbxYny:-OYq6TVW3mDIbWMbxYnz:b580np0) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/64e6167661c72d3e9f16c1b2b3e99fb4f615e4b8/src/BloomExe/XmlHtmlConverter.cs#L153)):*
> ```Smalltalk
>                 // not in the data returned by editor.getData().  Since assigning to div.innerHTML doesn't
>                 // affect what gets written to the file, this hack was implemented instead.
>                 newContents = Regex.Replace(newContents, @"(<br></br>|<br ?/>)[\r\n]*</p>", "</p>");
> ```

Not a new problem, but removing breaks at the end of paragraphs does not feel like a responsibility of this class, unless HtmlAgilityPack is the source of the unwanted breaks. Feels more like something HtmlDom should know, or possibly the method that retrieves HTML from the editable page and puts it back into the main document. Such a fix might be out of scope for this change, but if you agree it would at least be worth commenting that this is something Bloom-specific that might be better elsewhere.
___
*[`src/BloomExe/Program.cs` line 2043 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7342#-OYpxIObDpydpwAl5AN9:-OYpxIObDpydpwAl5ANA:b-a8yt37) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/64e6167661c72d3e9f16c1b2b3e99fb4f615e4b8/src/BloomExe/Program.cs#L2043)):*
> ```Smalltalk
>         {
>             get
>             {
> ```

Two thoughts
1. Can we rely on GetEntryAssembly() returning Bloom*.exe when we're NOT running tests? Would that be cleaner?
2. If we need to run this check over all assemblies, is that a large enough number to make it worth caching the result? I'm not sure how much this is called, either in production or testing, but it's not going to change, right?

nabalone:
- [ ] Update with new libpalaso version after https://github.com/sillsdev/libpalaso/pull/1457

andrew-polk (commented):
Ok, finished first round.

> @andrew-polk reviewed 14 of 16 files at r3, 2 of 2 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7342#-:-OZBFobL2dwfh_rfjgeX:b-4p2mfr)* status: all files reviewed, 8 unresolved discussions (waiting on @nabalone)

___
*[`.vscode/launch.json` line 11 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7342#-OYhLEpn2VXt1SD5kqtA:-OZB7WJzE81AGL7gi7Bh:b-loijxe) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/64e6167661c72d3e9f16c1b2b3e99fb4f615e4b8/.vscode/launch.json#L11)):*

VSCode generated this file. The gui for running locally only works when it is present. How should I document it?
</blockquote></details>

Actually, rather than say anything about this file, let's just enhance the "Build the .NET part" of the existing Readme to give alternate instructions for running in VS Code.
Or rather, give the VS Code instructions and make the VS ones the alternate.
___
*[`src/BloomExe/BloomExe.csproj` line 17 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7342#-OYgfgCc5pfXZJ-ED2vG:-OZBD_w-2CzUG5U_j5pA:b3gfhvb) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/cfae803e8bf0a0c6f72d928f82c3079825ffbce1/src/BloomExe/BloomExe.csproj#L17)):*

Hmm, that never occurred to me. https://learn.microsoft.com/en-us/dotnet/core/install/windows#supported-windows-versions says it supports "Windows 10 (22H2+)" so only the latest version of Windows 10 and no windows 8, but it also says "Windows 10 versions end-of-service dates are segmented by edition. Only Home, Pro, Pro Education, and Pro for Workstations editions are considered in the following table. Check the [Windows lifecycle fact sheet](https://support.microsoft.com/help/13853/windows-lifecycle-fact-sheet) for specific details."
</blockquote></details>

The information is definitely not straightforward, but I think it does mean we leave most of Win10 behind.
So we'll probably need to document this as the last version to support Win10 <22H2.
Let's add a todo in the card.
___
*[`src/BloomExe/Edit/EditingModel.cs` line 347 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7342#-OYgduWTCwl-kPJduPrh:-OZB9Weg-DsrKVJKMtw3:byqfwvd) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/6fc41e200b8c67630a16de1a5da365087e9dbe1c/src/BloomExe/Edit/EditingModel.cs#L347)):*

Up to Andrew. I'm good with taking it out.
</blockquote></details>

I defer to John. Let's just leave it out.
___
*[`src/BloomExe/ToPalaso/ProcessExtra.cs` line 44 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7342#-OYghROmB2Aphsi5TO8B:-OZBFjGr7dZzFU2_oDF9:bg6whm8) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/cfae803e8bf0a0c6f72d928f82c3079825ffbce1/src/BloomExe/ToPalaso/ProcessExtra.cs#L44)):*

SIL.Program.Process.SafeStart is basically identical to the first portion of this method (except for the `var windowMap = GetAllWindows();` line.) 

Could there have been a reason we weren't using SafeStart in the first place? Otherwise

TODO after fixing SIL.Program.Process.SafeStart in libpalasso I can refactor these to use that code instead of repeating it
</blockquote></details>

I don't know of a reason, but that doesn't mean there isn't one. : )
___
*[`src/BloomTests/Book/BookStorageTests.cs` line 252 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7342#-OYhSTre62oA9JwFi-YB:-OZBAKRNDaoGB2uH9tG-:buxisvl) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/64e6167661c72d3e9f16c1b2b3e99fb4f615e4b8/src/BloomTests/Book/BookStorageTests.cs#L252)):*

There are still a bunch of calls where we pass in a premade hardcoded html string, and for a least some I think it would sacrifice readability to break it into head and body content. I could make a second method that calls  GetInitialStorageWithCustomHtml(XmlHtmlConverter.CreateDocumentWithContent()). What should I call it though?
</blockquote></details>

Let's just leave it as is. Thanks.
___
*[`src/BloomExe/WebLibraryIntegration/S3Extensions.cs` line 33 at r4](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7342#-OZB9s0P4yflQvipoVOg:-OZB9s0P4yflQvipoVOh:b-6mrs10) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/e1312d10438a10fe454b535a77a57a0339ce6492/src/BloomExe/WebLibraryIntegration/S3Extensions.cs#L33)):*
> ```Smalltalk
>             {
>                 // Use the async method for .NET 8 compatibility
>                 // (I don't think this method is getting called at all)
> ```

The comment about not being used is not incorrect since you reinstated its use.

nabalone (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7342)* status: all files reviewed, 6 unresolved discussions (waiting on @andrew-polk and @JohnThomson)

___
*[`.vscode/launch.json` line 11 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7342#-OYhLEpn2VXt1SD5kqtA:-OZGS5lIDctMAXDPRkfm:b-a46eu6) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/64e6167661c72d3e9f16c1b2b3e99fb4f615e4b8/.vscode/launch.json#L11)):*

Actually, rather than say anything about this file, let's just enhance the "Build the .NET part" of the existing Readme to give alternate instructions for running in VS Code.
Or rather, give the VS Code instructions and make the VS ones the alternate.
</blockquote></details>

There seem to be major downsides to running in VS Code unless someone can figure out what I have not. I can't figure out how to Click from the tests I am running to their source code. And I don't see how to make the debugger break/ignore specific types of exceptions. 

I'm not sure what to put. I think everything just worked through vs code magic for me. Hopefully people can add if I missed something.
___
*[`src/BloomExe/BloomExe.csproj` line 17 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7342#-OYgfgCc5pfXZJ-ED2vG:-OZGUkuo-28kwlFQ2hCi:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/cfae803e8bf0a0c6f72d928f82c3079825ffbce1/src/BloomExe/BloomExe.csproj#L17)):*

The information is definitely not straightforward, but I think it does mean we leave most of Win10 behind.
So we'll probably need to document this as the last version to support Win10 <22H2.
Let's add a todo in the card.
</blockquote></details>

Done.
___
*[`src/BloomExe/XmlHtmlConverter.cs` line 18 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7342#-OYhNWwM29f48tRUeIcl:-OZFhaeZ7Mp6KSqEX2p2:b-k6fw91) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/64e6167661c72d3e9f16c1b2b3e99fb4f615e4b8/src/BloomExe/XmlHtmlConverter.cs#L18)):*

I'm OK with CreateValidHtmlString, though given that it isn't doing any error-correcting or validating and it's unlikely we'd want to create an invalid one, I think I prefer the shorter CreateHtmlString.
Given that we already have a bunch of tests using the (string bodyContents, string headContents = "") approach, and that most of the calls to this are in tests, I think its probably best to stick with what you have.
I'm not seeing a better place for the method. It doesn't feel general-purpose enough for XmlUtils, given that XmlUtils is shared library code, yet I think that's where I'd probably look for such a thing if the arguments and output are XHTML. If they are plain HTML (still needs to be a comment about that!), this file is about the only place that our C# deals with HTML, so this is probably the best place for it, and even if they are XHTML, I can't think of a better one, so let's leave it here.
</blockquote></details>

Renamed to CreateHtmlString and added a summary comment
___
*[`src/BloomExe/XmlHtmlConverter.cs` line 79 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7342#-OYq2YiJ4uQ1ooluKpM3:-OZGYA4p6do0gDNXL-D_:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/64e6167661c72d3e9f16c1b2b3e99fb4f615e4b8/src/BloomExe/XmlHtmlConverter.cs#L79)):*

Actually, I think your old code would have worked, since the Singleline option makes "." match everything including newlines. 
It's a badly-named option, though. I thought it meant that the regex would not be allowed to match anything longer than one line of the input, which left me surprised that it worked even with your fix.
SingleLine now does nothing, since your regex does not have any periods.
I'd suggest going one way or the other, and either way, say explicitly that multi-line comments are handled. For example, if you keep this regex, you could take out the SingleLine option, and change the comment to something like
// Remove comments (e.g. ) (even multi-line ones, thanks to [\s\S]),...
</blockquote></details>

Done.
___
*[`src/BloomExe/Publish/Video/RecordVideoWindow.cs` line 1348 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7342#-OYgechW7T5yl0dI0hHy:-OYr6_Jp5ksXYKG1CcCF:bb6un1j) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/cfae803e8bf0a0c6f72d928f82c3079825ffbce1/src/BloomExe/Publish/Video/RecordVideoWindow.cs#L1348)):*

I don't recall whether we set _ffmpegExited right when we send the 'q' or after we get some confirmation that it actually exited. If the latter, maybe we should have a new flag, something like "_attemptedToStopFfmpeg" that is set true right before we send the 'q'? That seems simpler than examining all our logic to see if we might need the process object after we asked it to quit.
</blockquote></details>

Done
___
*[`src/BloomExe/WebLibraryIntegration/S3Extensions.cs` line 33 at r4](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7342#-OZB9s0P4yflQvipoVOg:-OZBJfYy5EmnvvvtkGeG:b-4vtqog) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/e1312d10438a10fe454b535a77a57a0339ce6492/src/BloomExe/WebLibraryIntegration/S3Extensions.cs#L33)):*

The comment about not being used is not incorrect since you reinstated its use.
</blockquote></details>

Done. I don't think we really need the comment about .net 8 compatibility right?
___
*[`src/BloomTests/XmlHtmlConverterTests.cs` line 539 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7342#-OYqRoZa8nYbALeqWQOd:-OZFgRIaDshiu3Hc-fd5:b-smu12a) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/64e6167661c72d3e9f16c1b2b3e99fb4f615e4b8/src/BloomTests/XmlHtmlConverterTests.cs#L539)):*

Doc says that it's not possible in a JS DOM to have two of the same attribute. That would suggest it's a bug in our code that manipulates the HTML after we grab the outerHTML of the page? So I would definitely say it's a bad thing that we're depending on this behavior. Or maybe it's just some of our test data that has duplicate attributes?
At this point, then, we know that we're depending on the converter to fix cases of duplicate attributes, but we don't know whether we depend on it to keep a particular one if the values are different.
I'd be inclined to change the first data-foo's value to "ignored duplicate" and check that there is no such attribute in the output. Then comment along the lines "Some unit tests, at least, depend on the converter fixing duplicate attributes. We don't know whether it matters which one of a pair of duplicates it keeps, but the current code keeps the second, so this test is written to make sure we at least notice if this changes."
</blockquote></details>

Making and translating some books, I have not been able to catch Bloom adding two of the same attribute. Maybe it was just in unit tests.

In my one test case,  TidyManaged also kept the second one so behavior is same

nabalone (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7342)* status: all files reviewed, 3 unresolved discussions (waiting on @JohnThomson)

___
*[`src/BloomExe/XmlHtmlConverter.cs` line 24 at r5](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7342#-OZL6w1c65VTr-qw3fQl:-OZLZxGh7mYTHfQUJyF8:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/6ec8dab247a394d93b3fd22b880db82bf4cda07a/src/BloomExe/XmlHtmlConverter.cs#L24)):*

Params and returns are added by default in VS, but Andrew and I prefer to remove them unless it's useful to put content in them.
</blockquote></details>

Done.
___
*[`src/BloomExe/Publish/Video/RecordVideoWindow.cs` line 1358 at r5](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7342#-OZL8LeO2dO2VbAzqW0Z:-OZL_-az9jLYYcP7nQyL:bmbx3zk) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/6ec8dab247a394d93b3fd22b880db82bf4cda07a/src/BloomExe/Publish/Video/RecordVideoWindow.cs#L1358)):*

I'd like to capture these five lines in a method and call it in both places we need to send a 'q' to the process. You could include the Debug.Writeline from the other occurrence if you want. May need a little thought whether the check for _recording would be a problem in the other location.
It's not so much that 5 lines is a lot of code to duplicate as that I'd like to capture in one place the necessary checks and side effects that are needed when we want to make ffmpeg stop, so we won't miss one if we ever need to do something more, or ever need to stop it on another code path.
</blockquote></details>

Looks like we quit ffmpeg before we set _recording to false so should be fine. I'm not sure what to name this new method though
___
*[`src/BloomExe/ToPalaso/ProcessExtra.cs` line 44 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7342#-OYghROmB2Aphsi5TO8B:-OZL_L294Dm9PRhnchlX:bs7wumx) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/cfae803e8bf0a0c6f72d928f82c3079825ffbce1/src/BloomExe/ToPalaso/ProcessExtra.cs#L44)):*

I don't know of a reason, but that doesn't mean there isn't one. : )
</blockquote></details>

Done. 
___
*[`src/BloomTests/XmlHtmlConverterTests.cs` line 515 at r5](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7342#-OZLBXmx9G2kvmvd7PTr:-OZL_UJc9WqmpzfQggmg:b-q6i6w9) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/6ec8dab247a394d93b3fd22b880db82bf4cda07a/src/BloomTests/XmlHtmlConverterTests.cs#L515)):*

I'm not sure how consistent we are about this, or whether all the backwards ones in this file are new. But my intuition, and things I'm seeing online, is that the "message" in this sort of assert, since it is seen in the output only when the test fails, is meant to describe what is wrong about the failure. So for example this one should be "The second of the duplicate attributes is NOT present). (I've raised this in Flowdock, so if we reach the opposite conclusion ignore this.)
</blockquote></details>

Done. In the Zulip conversation we agreed that the "should" approach is okay

JohnThomson (commented):
> @JohnThomson reviewed 4 of 4 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7342)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @nabalone)

JohnThomson (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7342#-:-OZM-1fj8_KidX9K9pMy:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @nabalone)

PR #7349, opened 2025-09-05
Reorganize UpdateSupport (BL-15145)

JohnThomson:
Not sure I actually fixed any bugs, even BL-15145, but it was getting hard to reason about all the states and paths in the update code. I cleaned it up so the possible states are explicit, there is basically only one sequence of steps through the upgrade process, and Check For Updates will always report where you are in the process. Added some exception handling that might help.

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7349)

### DIFF
--- a/src/BloomExe/ApplicationUpdateSupport.cs
+++ b/src/BloomExe/ApplicationUpdateSupport.cs
@@ -84,18 +84,37 @@ namespace Bloom
             Action restartBloom
         )
         {
+            // In Bloom 6.3, we updated to DotNet 8. So at this point, there's no reason to check OS versions;
+            // This Velopack-based update code only runs in 6.3, and 6.3 (at least by the time we release a beta)
+            // only runs on an OS that is at least 10; in fact, it has to be quite a recent 10. But that check
+            // needs to be in the old Squirrel code in 6.2, to prevent updating to 6.3 at all. If we're in 6.3,
+            // we don't currently need to check OS version.
+            // I'm keeping the old code to remember the last state of things for version checking. Sometime, 6.3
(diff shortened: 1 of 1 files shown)


andrew-polk (changes requested):
> @andrew-polk reviewed 3 of 3 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7349)* status: all files reviewed, 1 unresolved discussion (waiting on @JohnThomson)

___
*[`src/BloomExe/ApplicationUpdateSupport.cs` line 95 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7349#-OZkVajMDdTv43iX9dvY:-OZkVajN-Wmuj0SaWtli:b64nnhc) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/2a72313619adaa3c86eb7814306409439a9bf3c3/src/BloomExe/ApplicationUpdateSupport.cs#L95)):*
We discussed that we should have a better message in this case. (Probably 6.2?)
Something like
> Bloom is not able to automatically upgrade to the latest version due to your operating system version.

But actually, no one can be running this code unless they have 10.
We probably want to reinstate something like this in 6.2 which checks for the specific version of Win10 which supports .NET8.
But for this PR, we said we should just comment it out.

andrew-polk (approved):
> @andrew-polk reviewed 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7349)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @JohnThomson)

PR #7362, opened 2025-09-08
Don't crash changing UI lang if creating initial collection (BL-15230)

StephenMcConnel:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7362)

### DIFF
--- a/src/BloomExe/Workspace/WorkspaceView.cs
+++ b/src/BloomExe/Workspace/WorkspaceView.cs
@@ -3,6 +3,7 @@ using System.Collections.Generic;
 using System.ComponentModel;
 using System.Diagnostics;
 using System.Drawing;
+using System.Globalization;
 using System.IO;
 using System.Linq;
 using System.Threading;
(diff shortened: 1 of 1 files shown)


StephenMcConnel (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7362)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @andrew-polk)

___
*[`src/BloomExe/Workspace/WorkspaceView.cs` line 826 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7362#-OZfaheqDdiNqWEYLzIY:-OZjjIH9BW3nqad1fOfs:bjzywcj) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/275cd0cfa8dde7f9659fba219dc9bebe7605ed6f/src/BloomExe/Workspace/WorkspaceView.cs#L826)):*

I tried to just instantiate a BloomWebSocketServer at this point, and the program crashed due to other things not yet being set up.  I'll try a bit harder to see if I can come up with something.
</blockquote></details>

I just hadn't thought through initialization of the BloomWebSocketServer.  The revised code seems to work well, at least if you change the UI language on the initial dialog and don't wait until you see the language chooser.

PR #7369, opened 2025-09-09
BL-14300 move text measurer to avoid scrollbar flicker

nabalone:


nabalone:
Is this okay, to put the text measurer in the top right corner, just with a negative z index?

PR #7379, opened 2025-09-12
Enhance Unauthorized access report if caused by hidden folder (BL-15197)

JohnThomson:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7379)

### DIFF
--- a/src/BloomExe/Book/BookStorage.cs
+++ b/src/BloomExe/Book/BookStorage.cs
@@ -2582,6 +2582,8 @@ namespace Bloom.Book
         /// </summary>
         internal void ShowAccessDeniedErrorHtml(UnauthorizedAccessException error)
         {
+            if (!(error is BloomUnauthorizedAccessException))
+                error = BloomUnauthorizedAccessException.CreateFromException(error);
             var message = GetAccessDeniedHtml(error, writeLog: true);
 
             ErrorMessagesHtml = message;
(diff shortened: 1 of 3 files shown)


JohnThomson (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7379)* status: all files reviewed, 2 unresolved discussions (waiting on @andrew-polk)

___
*[`src/BloomExe/Book/BookStorage.cs` line 504 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7379#-O_-Mrh909y0YPjqrToN:-O_D7Qox2vVIkp0okBzS:b-a330g7) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/40f05d755d2814d9ae68160616541c36f2791076/src/BloomExe/Book/BookStorage.cs#L504)):*

All these formatting changes are undoing the changes Steve made when he upgraded CSharpier. So you must still be running the old one.
</blockquote></details>

I don't know what happened. I seemed to be running 1.1.2, which is what dotnet tool restore installs, but git was already showing that the changes had been reversed at some point. Running from the command line got it right. Restarting VS seems to have fixed things, at least after I removed a global install.
___
*[`src/BloomExe/Book/BookStorage.cs` line 2658 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7379#-O_-NKkr5nkunViTaFdI:-O_DRpfNC_X6kbeqkaza:bebww3n) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/40f05d755d2814d9ae68160616541c36f2791076/src/BloomExe/Book/BookStorage.cs#L2658)):*

Seems worth extracting this to some kind of IO utility.
</blockquote></details>

After looking around a bit, it seemed natural to improve BloomUnauthorizedAccessException

PR #7383, opened 2025-09-16
Removes a lot of files we don't need (BL-15245, BL-15266)

JohnThomson:
A bunch of 32-bit DLLS some of which may have been needed in older versions of Windows and/or with dependencies we've retired.
A bunch of runtimes for platforms we aren't trying to support.
(Not sure if having the ARM runtimes is helpful. I think we would need a native ARM version of Bloom.exe, at least.)

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7383)


andrew-polk (commented):
> @andrew-polk reviewed 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7383)* status: all files reviewed, 1 unresolved discussion (waiting on @JohnThomson)

___
*[`src/BloomExe/BloomExe.csproj` line 23 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7383#-O_JT0Hl9eEqu1q3mbmW:-O_T0QjvDrUPicBzwZz3:bycvox9) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/fc8e5a974a70b33dfef47b06c4bb33bd2ed0e2fc/src/BloomExe/BloomExe.csproj#L23)):*

I added a comment...did you want a separate card?
</blockquote></details>

No, the comment is fine. I'll make the TC change as soon as this is merged.

PR #7386, opened 2025-09-18
Show the real book titles in the FAL book grid page (BL-15252)

StephenMcConnel:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7386)

### DIFF
--- a/src/BloomBrowserUI/bookEdit/bookLinkSetup/BookLinkTypes.ts
+++ b/src/BloomBrowserUI/bookEdit/bookLinkSetup/BookLinkTypes.ts
@@ -1,6 +1,7 @@
 export interface BookInfoForLinks {
     id: string;
     title: string;
+    realTitle?: string;
     thumbnail?: string;
     pageLength?: number;
 }
--- a/src/BloomBrowserUI/bookEdit/bookLinkSetup/LinkCard.tsx
+++ b/src/BloomBrowserUI/bookEdit/bookLinkSetup/LinkCard.tsx
@@ -3,6 +3,7 @@ import { Card, CardContent, Typography, IconButton } from "@mui/material";
 import DeleteIcon from "@mui/icons-material/Delete";
 import { Link } from "./BookLinkTypes";
 import { css } from "@emotion/react";
+import { BloomTooltip } from "../../react_components/BloomToolTip";
 
 interface BookCardProps {
     link: Link;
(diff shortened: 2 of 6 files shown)


andrew-polk (changes requested):
> @andrew-polk reviewed 5 of 5 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7386)* status: all files reviewed, 2 unresolved discussions (waiting on @StephenMcConnel)

___
*[`src/BloomExe/web/controllers/CollectionApi.cs` line 618 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7386#-O_THsyk3IXIzLbVRmCI:-O_THsylEJLAsCjIBkSd:b-qykqeu) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/8eb545c7b3729dc0f9d454b556fd88c6e9f48e2d/src/BloomExe/web/controllers/CollectionApi.cs#L618)):*
> ```Smalltalk
>                         else if (!string.IsNullOrEmpty(info.Title))
>                             title = info.Title;
>                     }
> ```

I know we can't call BookInfo.GetBestTitleForUserDisplay because that isn't a static, but I wonder if we need some more of the logic there.
In fact, it looks like we can make a static version of it which passes in AllTitles, Title, and IsInEditableCollection.
___
*[`src/BloomBrowserUI/bookEdit/bookLinkSetup/LinkCard.tsx` line 79 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7386#-O_TIjtxCzerxOsPkeHY:-O_TIjty-M0bsFE9UIW3:b-gz4oq8) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/8eb545c7b3729dc0f9d454b556fd88c6e9f48e2d/src/BloomBrowserUI/bookEdit/bookLinkSetup/LinkCard.tsx#L79)):*
> ```XML
>                 </IconButton>
>             )}
>             <BloomTooltip tip={link.book.realTitle || link.book.title}>
> ```

Seems like we don't want to fall back to link.book.title since that is what is already displayed.
Or were you thinking it was better to always have a tooltip for consistency? I guess I can see that.

andrew-polk (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7386)* status: all files reviewed, 1 unresolved discussion (waiting on @StephenMcConnel)

___
*[`src/BloomExe/web/controllers/CollectionApi.cs` line 618 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7386#-O_THsyk3IXIzLbVRmCI:-O_TfUkKCfqNegpNtV5_:b-pczygm) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/8eb545c7b3729dc0f9d454b556fd88c6e9f48e2d/src/BloomExe/web/controllers/CollectionApi.cs#L618)):*

We have an actual BookInfo object so we don't need a static method.  But BookInfo.GetBestTitleForUserDisplay explicitly uses the FolderName if it's set, so we really don't want to use it.  The current code is effectively the same for selecting a title from the json, but somewhat simpler.
</blockquote></details>

Sorry, I got mixed up. 

I still think we probably want the newline and trim logic. So maybe we should extract that part and call it from here?

andrew-polk:
*[`src/BloomExe/web/controllers/CollectionApi.cs` line 618 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7386#-O_THsyk3IXIzLbVRmCI:-O_TtD7e5yDyDU9puj_9:b-jr8wxp) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/8eb545c7b3729dc0f9d454b556fd88c6e9f48e2d/src/BloomExe/web/controllers/CollectionApi.cs#L618)):*

Sorry, I got mixed up. 

I still think we probably want the newline and trim logic. So maybe we should extract that part and call it from here?
</blockquote></details>

Actually, I'm not sure about the newlines...
I could see benefit in leaving them as they are.

StephenMcConnel (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7386)* status: all files reviewed, 1 unresolved discussion (waiting on @andrew-polk)

___
*[`src/BloomExe/web/controllers/CollectionApi.cs` line 618 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7386#-O_THsyk3IXIzLbVRmCI:-O_TrABv8Ig0cxmjKBcu:b-qg55vi) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/8eb545c7b3729dc0f9d454b556fd88c6e9f48e2d/src/BloomExe/web/controllers/CollectionApi.cs#L618)):*

Sorry, I got mixed up. 

I still think we probably want the newline and trim logic. So maybe we should extract that part and call it from here?
</blockquote></details>

I decided it was probably simplest to add an optional argument to BookInfo.GetBestTitleForUserDisplay.

PR #7407, opened 2025-09-25
Add help link for file permissions to hard crash report (BL-15282)

StephenMcConnel:
This can happen if the .bloomCollection file is readonly or hidden.

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7407)


andrew-polk:
This looks like an improvement, but I was also thinking it was worth a look to see if we can handle the issue more gracefully.

andrew-polk:
Could very well be that it isn't worth it. I guess I'm actually not sure what we can do. So unless you have any better ideas, we can merge this.

PR #7414, opened 2025-09-30
Replace QuizHeader- and Prompt-style with GameHeader-style (BL-14794)

JohnThomson:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7414)

### DIFF
--- a/src/BloomBrowserUI/bookEdit/css/editMode.less
+++ b/src/BloomBrowserUI/bookEdit/css/editMode.less
@@ -799,7 +799,7 @@ div.bloom-templateMode {
 // We don't want a border on the box itself because that takes up space; a box that
 // once starts overflowing might be stuck that way because of the extra space needed
 // for the border. So we use a pseudo-element to draw the line.
-.childOverflowingThis:after {
+.childOverflowingThis:not(.bloom-ignoreOverflow):after {
     content: "";
     position: absolute;
     top: 0;
--- a/src/BloomExe/Book/Book.cs
+++ b/src/BloomExe/Book/Book.cs
@@ -1787,6 +1787,7 @@ namespace Bloom.Book
             Storage.MigrateToLevel6LegacyActivities();
             Storage.MigrateToLevel7BloomCanvas();
             Storage.MigrateToLevel8RemoveEnterpriseOnly();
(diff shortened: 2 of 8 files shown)


StephenMcConnel (commented):
Generally looks okay, but one question about a comment that doesn't make sense to me.

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7414#-:-OaaCskv4ZPongr7TLcE:b-8k0j9u)* status: all files reviewed, 1 unresolved discussion (waiting on @JohnThomson)

JohnThomson (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7414)* status: all files reviewed, 1 unresolved discussion (waiting on @StephenMcConnel)

___
*[`src/BloomExe/Book/BookStorage.cs` line 4092 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7414#-OaaC0Ox6xQUqR7Uk8XD:-OazHbyy2XgELQYpLkIw:b-lmnve1) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/4dd8ed2bd6eb9a4af1614a305fba11b70a598870/src/BloomExe/Book/BookStorage.cs#L4092)):*

I don't understand this last comment/question about removing QuizHeader-style and Prompt-style definitions.
It looks later on like you did remove the QuizHeader-style and Prompt-style definitions.
</blockquote></details>

I have removed the definitions from the pug files. I don't remove existing definitions of those styles during the migration process here.
The code here replaces anything that uses QuizHeader-style and Prompt-style in class attributes with GameHeader-style.
However, typically older books also have, in the HTML file as part of a user-defined-style element in the header, definitions of what those classes used to do. I've chosen not to try to remove those pieces of the user-defined-styles element in the migration process. So, for example, I think the format dialog will still show Quiz Header and Prompt as possible styles. (If the user really doesn't like how GameHeader looks, he could change them back, and since this migration only runs once the change would stick.)
We'd save a tiny bit of space by removing the style definitions. And maybe it is confusing rather than helpful to have them still available as styles. So I thought it worth asking the question.

PR #7422, opened 2025-10-01
Detect path too long for collection files on startup (BL-15304)

StephenMcConnel:
Also improve the message when detecting an overly long path.

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7422)

### DIFF
--- a/DistFiles/localization/en/Bloom.xlf
+++ b/DistFiles/localization/en/Bloom.xlf
@@ -2854,7 +2854,7 @@
         <note>ID: ErrorReport.Pre10WindowsNotSupported</note>
       </trans-unit>
       <trans-unit id="Errors.PathTooLong">
-        <source xml:lang="en">Please move your collection closer to the root of your hard drive and try again. A file Bloom was working with had a path that was too long. This is usually caused by your collection being too deeply nested in many folders on your hard drive.</source>
+        <source xml:lang="en">Please give your collection a shorter name or move your collection closer to the root of your hard drive and try again. A file Bloom was working with had a path that was too long. This is usually caused by one of two things: 1) the collection has a very long name, or 2) the collection is too deeply nested inside many folders on your hard drive.</source>
         <note>ID: Errors.PathTooLong</note>
       </trans-unit>
       <trans-unit id="Errors.BookProblem">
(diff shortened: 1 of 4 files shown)


andrew-polk (changes requested):
> @andrew-polk reviewed 3 of 3 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7422)* status: all files reviewed, 1 unresolved discussion (waiting on @StephenMcConnel)

___
*[`src/BloomExe/web/controllers/ProblemReportApi.cs` line 1013 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7422#-Oab7KNL44XlPYWxDNOa:-Oab7KNL44XlPYWxDNOb:b-cbri7m) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/7820fe81a8b9e04716c27bb10cc63c18bd045f99/src/BloomExe/web/controllers/ProblemReportApi.cs#L1013)):*
> ```Smalltalk
>                     if (path?.Length >= 260)
>                         heading = LongPathAware.GetGenericPathTooLongMessage();
>                 }
> ```

This logic feels out of place in ProblemReportApi.
I think we should at least make it a method on LongPathAware we can call.

Or perhaps there is a better place to check the exception higher up and convert it into our own exception which already has the message we want in the exception.Message? Perhaps wrapping the original?

StephenMcConnel (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7422)* status: 1 of 4 files reviewed, 1 unresolved discussion (waiting on @andrew-polk)

___
*[`src/BloomExe/web/controllers/ProblemReportApi.cs` line 1013 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7422#-Oab7KNL44XlPYWxDNOa:-OauNwRr6Ug4AVcmymwo:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/7820fe81a8b9e04716c27bb10cc63c18bd045f99/src/BloomExe/web/controllers/ProblemReportApi.cs#L1013)):*

This logic feels out of place in ProblemReportApi.
I think we should at least make it a method on LongPathAware we can call.

Or perhaps there is a better place to check the exception higher up and convert it into our own exception which already has the message we want in the exception.Message? Perhaps wrapping the original?
</blockquote></details>

Done.

PR #7866, opened 2026-04-22
Ensure background cover image covers entire page (BL-16173)

StephenMcConnel:
---

<a href="https://app.devin.ai/review/bloombooks/bloomdesktop/pull/7866" target="_blank">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-open-in-devin-review-dark.svg?v=1">
    Open in Devin Review
  </picture>
</a>

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7866)

### DIFF
--- a/src/BloomExe/Publish/BloomPub/BloomPubMaker.cs
+++ b/src/BloomExe/Publish/BloomPub/BloomPubMaker.cs
@@ -891,9 +891,7 @@ namespace Bloom.Publish.BloomPub
         /// - delete the img element
         /// (See oldImg and newImg in unit test CompressBookForDevice_ImgInImgContainer_ConvertedToBackground for an example).
         /// </summary>
-        /// <param name="wholeBookHtml"></param>
-        /// <returns></returns>
-        private static void ConvertImagesToBackground(SafeXmlDocument dom)
+        internal static void ConvertImagesToBackground(SafeXmlDocument dom)
         {
             foreach (
                 var imgContainer in dom.SafeSelectNodes(
@@ -929,6 +927,38 @@ namespace Bloom.Publish.BloomPub
                 // the img? But we'd still need duplicate rules.
                 if ((img.GetAttribute("class") ?? "").Contains("bloom-imageObjectFit-cover"))
                     classesToAdd += " bloom-imageObjectFit-cover";
+                else
+                {
+                    // See BL-16173 for this additional special case which may be even nastier.
+                    // If the image is hosted in a bloom-canvas instead of a bloom-imageContainer,
+                    // double-check that it is also in a bloom-page with coverColor and bloom-customLayout.
+                    // If so, we need to add the same class to the imgContainer div to ensure that the
+                    // image will cover the whole canvas.
+                    var classes = imgContainer.GetClasses();
+                    if (
+                        classes.Contains("bloom-canvas")
+                        && !classes.Contains("bloom-imageContainer")
+                    )
+                    {
+                        for (
+                            var parent = imgContainer.ParentElement;
+                            parent != null;
+                            parent = parent.ParentElement
+                        )
+                        {
+                            var parentClasses = parent.GetClasses();
+                            if (parentClasses.Contains("bloom-page"))
+                            {
+                                if (
+                                    parentClasses.Contains("coverColor")
+                                    && parentClasses.Contains("bloom-customLayout")
+                                )
+                                    classesToAdd += " bloom-imageObjectFit-cover";
+                                break; // no reason to go any higher than the page
+                            }
+                        }
+                    }
+                }
 
                 imgContainer.SetAttribute(
                     "class",
--- a/src/BloomTests/Publish/BloomPub/BloomPubMakerTests.cs
+++ b/src/BloomTests/Publish/BloomPub/BloomPubMakerTests.cs
@@ -2144,6 +2144,92 @@ namespace BloomTests.Publish.BloomPub
             }
         }
 
+        private static SafeXmlDocument MakeDom(string bodyInnerXml)
+        {
+            var doc = SafeXmlDocument.Create();
+            doc.LoadXml($"<html><body>{bodyInnerXml}</body></html>");
+            return doc;
+        }
+
+        [Test]
+        public void ConvertImagesToBackground_BasicImgInBloomCanvas_ConvertedToBackgroundStyle()
+        {
+            var dom = MakeDom("<div class='bloom-canvas'><img src='image.png'/></div>");
+            BloomPubMaker.ConvertImagesToBackground(dom);
+            var div = dom.SafeSelectNodes("//div[@class]").Cast<SafeXmlElement>().First();
+            Assert.That(div.GetAttribute("style"), Is.EqualTo("background-image:url('image.png')"));
+            Assert.That(
+                div.GetAttribute("class"),
+                Does.Contain("bloom-background-image-in-style-attr")
+            );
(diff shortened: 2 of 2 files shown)


JohnThomson (changes requested):
This is an elegant way to fix things when the image is supposed to fill the container and we didn't get the cropping perfectly right. But we still have lots of images around which don't fill their containers (both cropped and uncropped). I think this will change them all drastically. If I'm right, we can't just do this.
Conceivably we could add a new class at publish time which does this when we think we have cropped an image to exactly fill the container. That would be OK for bloompubs and epubs. But for publishing to the web, we need to end up in a state where the book can continue to be edited. The next editor might crop the image differently or resize the page, and may not want the image locked to cover mode. In any case, I'm pretty sure that editing converts it back from bloom-background-image-in-style-attr to our normal image structure, so in that situation I would expect your change to be ineffective. And "Download for edit" is one of the cases where the bug is reported.
I'm not sure what the answer is. I was vaguely hoping the problem was just a rounding error in the cropping code, but maybe there isn't a perfect answer? The real image size is always going to be rounded to pixels, and the container size may involve fractional pixels; if the two sizes aren't exactly the same, maybe we will always or usually get a pixel along one side with contain?
Maybe we need to round the size of the container to whole pixels? But that size is probably just 100% in at least some cases; publishing removes the canvas element.

> @JohnThomson reviewed 2 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7866#-:-OqvL34t1v7ZZpnyZGAJ:bm0dsid)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on StephenMcConnel).

StephenMcConnel (commented):
I've changed the code to add a class when I think it's needed.

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7866#-:-OrElIg71OPMkuptlPG5:brkpie2)* status: 0 of 3 files reviewed, all discussions resolved (waiting on JohnThomson).

JohnThomson (changes requested):
> @JohnThomson reviewed 4 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7866)* status: all files reviewed, 1 unresolved discussion (waiting on StephenMcConnel).

___
*[`src/BloomExe/Publish/BloomPub/BloomPubMaker.cs` line 944 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7866#-OrKdQqNCly11K3TiaUN:-OrKdQqNCly11K3TiaUO:b6hj61e) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/1844dfa91262843b3cd4a6e635cb50740aebb553/src/BloomExe/Publish/BloomPub/BloomPubMaker.cs#L944)):*
> ```Smalltalk
>                         }
>                     }
>                 }
> ```

I still think we may need to be more picky about when we do this. Have you tried how it works with navigation buttons (image or image-with-caption) set to "fit with margin" or "fit to margin"? Also, if you switch a front cover to custom before you set an image, and then choose one, it doesn't necessarily fill the page. (Maybe this is a bug? I expected 'reset image' or choosing a new image to put things back to imitating object-fit:contain, but they don't.)
I don't really see an alternative to adding this class, because when we crop the image to size, it gets rounded to pixels and sometimes won't fit perfectly, so object-fit:contain is always going to have a high chance of showing a pixel somewhere, since it prefers that to hiding a pixel. But maybe the check could be an option in ReallyCropImage and conditional on whether the image actually fills its container? Then again, maybe you can't tell that in C#. 
Another concern is that an aspect of the original complaint is that when the book is downloaded for editing we can also get this extra pixel. This solution is not going to help there (or happen...BloomPubMaker is not involved). Of course, when we upload the book in the expectation of further editing, we crop but don't convert to a background image. So in that case we could theoretically remove most of the hidden part of the image, but keep a pixel or so more than we need and continue to crop, to ensure that the background is covered. Is that a good way to go?
Or we could decide that upload should not really-crop-images. That would be a decision for JohnH. Concerns would include uploading a bigger file than we need to and possibly privacy...the author might conceivably really not want to share something he has cropped.
Sorry to be picky...I'm afraid this is a quite difficult problem and I'm not sure what will fix it reliably without breaking something else.
(I'm also surprised that this fixes epubs. Are they using some BloomPubMaker code?)

StephenMcConnel (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7866)* status: 2 of 4 files reviewed, 1 unresolved discussion (waiting on JohnThomson).

___
*[`src/BloomExe/Publish/BloomPub/BloomPubMaker.cs` line 944 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7866#-OrKdQqNCly11K3TiaUN:-OrLLjDz95kMKajqCZfZ:bhlw9ga) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/1844dfa91262843b3cd4a6e635cb50740aebb553/src/BloomExe/Publish/BloomPub/BloomPubMaker.cs#L944)):*

I still think we may need to be more picky about when we do this. Have you tried how it works with navigation buttons (image or image-with-caption) set to "fit with margin" or "fit to margin"? Also, if you switch a front cover to custom before you set an image, and then choose one, it doesn't necessarily fill the page. (Maybe this is a bug? I expected 'reset image' or choosing a new image to put things back to imitating object-fit:contain, but they don't.)
I don't really see an alternative to adding this class, because when we crop the image to size, it gets rounded to pixels and sometimes won't fit perfectly, so object-fit:contain is always going to have a high chance of showing a pixel somewhere, since it prefers that to hiding a pixel. But maybe the check could be an option in ReallyCropImage and conditional on whether the image actually fills its container? Then again, maybe you can't tell that in C#. 
Another concern is that an aspect of the original complaint is that when the book is downloaded for editing we can also get this extra pixel. This solution is not going to help there (or happen...BloomPubMaker is not involved). Of course, when we upload the book in the expectation of further editing, we crop but don't convert to a background image. So in that case we could theoretically remove most of the hidden part of the image, but keep a pixel or so more than we need and continue to crop, to ensure that the background is covered. Is that a good way to go?
Or we could decide that upload should not really-crop-images. That would be a decision for JohnH. Concerns would include uploading a bigger file than we need to and possibly privacy...the author might conceivably really not want to share something he has cropped.
Sorry to be picky...I'm afraid this is a quite difficult problem and I'm not sure what will fix it reliably without breaking something else.
(I'm also surprised that this fixes epubs. Are they using some BloomPubMaker code?)
</blockquote></details>

I've refined the filtering for when the bloom-imageObjectFit-cover class is added.  I tested the cover with adding both a simple image and the navigation button images.  As far as I can tell, everthing is working okay.
I'm not sure this code is affecting epubs, but I know epub and bloompub do share code in some non-obvious ways.
I don't know if there's anything actually needed for uploading books to bloomlibrary.org.  If there is, I don't know how to address it.  If we're cropping images for uploading, I'm not sure that is the wisest course of action.

JohnThomson (approved):
> @JohnThomson reviewed 2 files and all commit messages, and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7866)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on StephenMcConnel).

PR #7874, opened 2026-04-27
Fix BL-16226 Error checking in when old repo Book.bloom is missing

andrew-polk:
Treat a missing old repo-side .bloom file as a no-op during rename cleanup on check-in, and clear stale oldName metadata after the cleanup path runs.

---

<a href="https://app.devin.ai/review/bloombooks/bloomdesktop/pull/7874" target="_blank">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-open-in-devin-review-dark.svg?v=1">
    Open in Devin Review
  </picture>
</a>

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7874)

### DIFF
--- a/src/BloomExe/TeamCollection/FolderTeamCollection.cs
+++ b/src/BloomExe/TeamCollection/FolderTeamCollection.cs
@@ -995,6 +995,9 @@ namespace Bloom.TeamCollection
             var oldLocalPath = Path.Combine(Path.GetDirectoryName(newBookFolderPath), oldName);
             var pathToOldBookFileInRepo = GetPathToBookFileInRepo(oldLocalPath);
             var pathToNewBookFileInRepo = GetPathToBookFileInRepo(newBookFolderPath);
+            // If the old repo file is already gone, check-in should continue by writing the new one. See BL-16226.
+            if (!RobustFile.Exists(pathToOldBookFileInRepo))
+                return;
             // There is probably some pathological case where pathToNewBookFileInRepo already exists,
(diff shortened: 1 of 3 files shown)


andrew-polk:
I asked JT to analyze this change, especially if this is the behavior we want. Here's his analysis:

I suppose it depends a bit on what the scenario is. How could this happen?

- While I was renaming the book, I (or someone) deleted the old file. Maybe I did it myself, thinking it a mistake for it to be still there after I renamed it.
- While I had the book checked out for rename, the admin undid my checkout and deleted the book
- While I had the book checked out for rename, the admin undid my checkout and renamed the book (same as I'm attempting? Different?)
- I got short of disk space and changed the whole TC folder to be stored only offline

It seems clearly the right behavior for hand-deletion. At this point, by checking it in, I'm indicating that my desire is for this book by this name to be in the TC, and we're making it so.
I would hope our TC code would detect the canceled-checkout cases. Maybe worth testing. If the admin has already renamed it and my checkin succeeds, we potentially have the very bad situation of two check-in books with the same instance ID.
The last case is even more likely to have that result, since the old and new versions probably both wind up in Dropbox.
I can't think of anything more promising to do.

JohnThomson (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7874)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on andrew-polk).

PR #7898, opened 2026-05-07
Log source of enabling/disabling workspace tabs (BL-16290)

StephenMcConnel:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7898)

### DIFF
--- a/src/BloomExe/Program.cs
+++ b/src/BloomExe/Program.cs
@@ -1420,7 +1420,14 @@ namespace Bloom
 
             Sldr.Cleanup();
             Logger.WriteMinorEvent("shutting down logger, about to dispose project context");
+            // Force the log file to include the minor events.  I don't know why this isn't the default. (BL-16290)
+            var logText = Logger.LogText;
+            var logPath = Logger.LogPath;
             Logger.ShutDown();
+            if (string.IsNullOrWhiteSpace(logPath))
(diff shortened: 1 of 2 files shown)


src/BloomExe/Program.cs
```diff
@@ -1420,7 +1420,14 @@ private static void Run(string[] args)
 
             Sldr.Cleanup();
             Logger.WriteMinorEvent("shutting down logger, about to dispose project context");
+            // Force the log file to include the minor events.  I don't know why this isn't the default. (BL-16290)
```
hatton, line 1423: Because the minor events are just there in case of a crash, writing all of the minor events an to a long-lived log means the log will be larger. Maybe that's okay, but now you know why.

StephenMcConnel: Because of limiting the minor events to the latest 5000 bytes or so written from the program, it doesn't make the log file all that much larger.

StephenMcConnel (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7898)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on hatton).

andrew-polk (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7898)* status: all files reviewed, 1 unresolved discussion (waiting on hatton).

PR #7906, opened 2026-05-12
Remove width limit on Settings button (BL-16307)

JohnThomson:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7906)

### DIFF
--- a/src/BloomBrowserUI/react_components/TopBar/CollectionTopBarControls/CollectionTopBarControls.tsx
+++ b/src/BloomBrowserUI/react_components/TopBar/CollectionTopBarControls/CollectionTopBarControls.tsx
@@ -67,13 +67,11 @@ export const CollectionTopBarControls: React.FunctionComponent = () => {
                     backgroundColor={mainButtonBackground}
                     textColor={mainButtonTextColor}
                     cssOverrides={css`
-                        width: 88px;
                         white-space: normal;
                         line-height: 1.15;
 
                         span {
                             display: inline-block;
-                            max-width: 64px;
                             text-align: center;
                         }
                     `}


andrew-polk (changes requested):
> @andrew-polk reviewed 1 file and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7906)* status: all files reviewed, 1 unresolved discussion (waiting on JohnThomson).

___
*[`src/BloomBrowserUI/react_components/TopBar/CollectionTopBarControls/CollectionTopBarControls.tsx` line 75 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7906#-OsRuWWi54yjwQio6Zdx:-OsRuWWi54yjwQio6Zdy:b-9gt1qb) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/616e3b06eeb3709bd1312ec8bd38774893bd4fb6/src/BloomBrowserUI/react_components/TopBar/CollectionTopBarControls/CollectionTopBarControls.tsx#L75)):*
> ```XML
>                         span {
>                             display: inline-block;
>                             text-align: center;
> ```

I think these changes will mean we never wrap? That doesn't seem ideal. 
Is there any reasonable/feasible way to make it expand its width as needed to accommodate no more than three lines?

JohnThomson (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7906)* status: all files reviewed, 1 unresolved discussion (waiting on andrew-polk).

___
*[`src/BloomBrowserUI/react_components/TopBar/CollectionTopBarControls/CollectionTopBarControls.tsx` line 75 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7906#-OsRuWWi54yjwQio6Zdx:-OsRyZKQ7ujAkUMoHtrP:bxdoyh0) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/616e3b06eeb3709bd1312ec8bd38774893bd4fb6/src/BloomBrowserUI/react_components/TopBar/CollectionTopBarControls/CollectionTopBarControls.tsx#L75)):*

I think these changes will mean we never wrap? That doesn't seem ideal. 
Is there any reasonable/feasible way to make it expand its width as needed to accommodate no more than three lines?
</blockquote></details>

It's inside four layers of boxes that all have display:flex, and has the default shrink value of "1". Seems like that should give it a reasonable chance of starting to wrap if space gets tight. I'm inclined to leave this (YAGNI) until/unless we have enough clutter in the Collection tab toolbar to need it. I think it looks neater (and matches all the other buttons in this row) with a single line label if it fits.

PR #7918, opened 2026-05-15
Make right/wrong elements consistent across game templates (BL-16329)

andrew-polk:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7918)

### DIFF
--- a/src/content/templates/template books/Games/Games.html
+++ b/src/content/templates/template books/Games/Games.html
@@ -371,11 +371,41 @@
                             </div>
                         </div>
                         <div
+                            class="bloom-canvas-element drag-item-wrong"
+                            style="
+                                height: 260px;
+                                left: 116px;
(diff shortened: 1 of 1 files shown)


JohnThomson (approved):
I don't see any real way to review this. The changed positions and added elements are helpful, or they are not. Presumably the HTML parses enough for a smoke test. I think we may as well merge and let testers and users evaluate the result.

> @JohnThomson reviewed 1 file and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7918#-:-Osfkkja7cLxNeyN4_tN:bnm9gbn)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on andrew-polk).

PR #7927, opened 2026-05-21
Handle line art where we have a transparent marginBox (BL-16336)

JohnThomson:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7927)


JohnThomson:
Handling anything but complete transparency is too complicated

PR #7947, opened 2026-06-05
Delay page saves for async canvas image resize (BL-16302)

andrew-polk:
When adjustContainerAspectRatio waits for a newly changed image to load, it can update canvas geometry after changeImage() returns. That left a window where requestPageContent could save stale width, height, and position data.

Register a requestPageContent delay around that image-load retry, keep the existing paste-image wrapper for the separate async feature-status check, add a focused regression test, and update the C# comment that describes the async save behavior.

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7947)


JohnThomson (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7947)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on andrew-polk).

PR #7950, opened 2026-06-09
Fix BL-16416 Delete image option disabled after using "Become Background" (custom cover)

andrew-polk:
https://issues.bloomlibrary.org/youtrack/issue/BL-16416

On a custom front cover, after using "Become Background" the resulting
background image is designated the cover image (data-book="coverImage").
Cover images carry a persisted onerror that sets bloom-imageLoadError, and
their resource load can fail spuriously before bootstrap (BL-14241). Nothing
clears that class on a later successful load. BL-16159 made hasRealImage()
return false whenever bloom-imageLoadError is present, so the image looked
"unreal" and the Delete option was disabled even though the image displays
fine. This is cover-specific (interior images have no cover onerror) and
intermittent (the early-load failure is a timing race).

Fix:
- hasRealImage() now treats a non-placeholder image that actually rendered
  (complete && naturalWidth > 0) as real, ignoring a stale
  bloom-imageLoadError class. A genuinely broken image still has
  naturalWidth 0, preserving BL-16159's intent.
- The "Become Background" handler now clears bloom-imageLoadError and rewires
  onerror before changing the background image's src, matching the canonical
  switchBackgroundToCanvasElement path.

Co-Authored-By: Claude Opus 4.8 (1M context) <noreply@anthropic.com>

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7950)

### DIFF
--- a/src/BloomBrowserUI/bookEdit/js/CanvasElementContextControls.tsx
+++ b/src/BloomBrowserUI/bookEdit/js/CanvasElementContextControls.tsx
@@ -21,6 +21,7 @@ import { showCopyrightAndLicenseDialog } from "../workspaceRoot";
 import {
     doImageCommand,
     getImageUrlFromImageContainer,
+    HandleImageError,
     kImageContainerClass,
     isPlaceHolderImage,
 } from "./bloomImages";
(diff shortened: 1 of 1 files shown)


JohnThomson (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7950)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on andrew-polk).

PR #7952, opened 2026-06-09
Make page thumbnail list update incrementally when adding pages (BL-16418)

andrew-polk:
Adding any page from the Add Page dialog repainted the entire page
thumbnail list, and occasionally the list never updated at all until
something else forced a refresh. Several causes, all addressed here:

- OnInsertPage did a hard location.reload of the page-list iframe on
  every insert. This rebooted the whole React list and raced with the
  deferred pageListChangedEvent notifications; a websocket message
  arriving mid-reload was silently dropped, leaving the list stale.
  Removed; the refresh now flows solely through UpdatePageList.

- GetTemplateStyleSheets treated editPaneGlobal.css as a template
  stylesheet. Since RemoveModeStyleSheets strips that link on every
  save, AddStylesheetFromAnotherBook re-added it on every insert from
  a template that links it (e.g. Basic Book), so stylesChanged was
  perpetually true and every insert took the full page-list document
  swap instead of the cheap websocket refresh. It is now ignored,
  along with origami.css, which storage manages independently.

- The full-rebuild branch of UpdateItemsInternal sent no websocket
  notification, so a caller that reached it without also navigating
  the iframe never triggered any repaint. It now always sends
  pageListNeedsRefresh.

- The page list registered its websocket listener only after an async
  localization call resolved, leaving a load-time window in which
  messages from C# were lost; WebSocketManager does not queue them.
  The listener is now registered immediately, and the list re-fetches
  whenever its socket (re)opens so any message missed while
  disconnected self-heals.

Adding a page whose stylesheets are already in the book now updates
just the affected thumbnails; the full document swap remains only for
inserts that genuinely bring new stylesheets.

Co-Authored-By: Claude Fable 5 <noreply@anthropic.com>

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7952)

### DIFF
--- a/src/BloomBrowserUI/bookEdit/pageThumbnailList/pageThumbnailList.tsx
+++ b/src/BloomBrowserUI/bookEdit/pageThumbnailList/pageThumbnailList.tsx
@@ -533,10 +533,12 @@ const PageList: React.FunctionComponent<{ initialPageLayout: string }> = (
 
     // All the code in this useEffect is one-time initialization.
     useEffect(() => {
-        let localizedNotification = "";
+        // English fallback shown in the "saving" toast until the localized text arrives.
+        // The websocket listener must be registered immediately (see below), so it cannot
+        // wait for localization; in the worst case an early save shows this fallback.
+        let localizedNotification = "Saving...";
 
-        // This function will be hooked up (after we set localizedNotification properly)
-        // to be called when C# sends messages through the web socket.
+        // This function will be called when C# sends messages through the web socket.
         // We need a named function because it looks cleaner.
         webSocketListenerFunction = (event) => {
             switch (event.id) {
@@ -581,17 +583,36 @@ const PageList: React.FunctionComponent<{ initialPageLayout: string }> = (
                     // below that uses resetValue for something we don't care about.
                     setResetValue((oldResetValue) => oldResetValue + 1);
                     break;
+                case `websocket/open/${kWebsocketContext}`:
+                    // The socket just (re)opened. Any messages C# sent while we had no
+                    // connection (e.g. while this iframe was loading, or after a network
+                    // hiccup before the 2-second reconnect loop restored the socket) were
+                    // lost; the server does not queue them. Re-fetch the page list so we
+                    // are sure to be in sync. This bump also drives the very first load:
+                    // the effect below skips its initial run (reloadValue === 0) and waits
+                    // for this, so we fetch the page list exactly once, when the socket is
+                    // ready, instead of fetching immediately and then again on open.
+                    setReloadValue((oldReloadValue) => oldReloadValue + 1);
+                    break;
             }
         };
+
+        // Register the listener right away. This used to happen only inside the .done()
+        // of the localization request below, which widened the window during load in which
+        // messages from C# (e.g. pageListNeedsRefresh after adding a page) were silently
+        // dropped. Registering now shrinks that window but does not eliminate it: anything
+        // C# sends before the socket finishes opening is still lost (the server does not
+        // queue messages). The websocket/open handler above is what actually closes the
+        // remaining gap, by re-fetching the page list once the socket is connected.
+        WebSocketManager.addListener(
+            kWebsocketContext,
+            webSocketListenerFunction,
+        );
 
         theOneLocalizationManager
             .asyncGetText("EditTab.SavingNotification", "Saving...", "")
             .done((savingNotification) => {
                 localizedNotification = savingNotification;
-                WebSocketManager.addListener(
-                    kWebsocketContext,
-                    webSocketListenerFunction,
-                );
             });
         theOneLocalizationManager
             .asyncGetText("EditTab.PageList.Heading", "Pages", "")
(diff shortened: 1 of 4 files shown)


JohnThomson (changes requested):
Very nice work. A few minor things to think about.

> @JohnThomson reviewed 4 files and all commit messages, and made 4 comments.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7952#-:-OusJu7M8Zu-TGwqQXx5:b-uc40ms)* status: all files reviewed, 3 unresolved discussions (waiting on andrew-polk).

___
*[`src/BloomBrowserUI/bookEdit/pageThumbnailList/pageThumbnailList.tsx` line 602 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7952#-OusHzud8ql2JNivV-xM:-OusHzud8ql2JNivV-xN:b-fujojn) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/def519c8ab888cb446c4a05eff040b74dda4c6f1/src/BloomBrowserUI/bookEdit/pageThumbnailList/pageThumbnailList.tsx#L602)):*
> ```XML
>         // of the localization request below, which left a window during load in which
>         // messages from C# (e.g. pageListNeedsRefresh after adding a page) were silently
>         // dropped, so the list never updated until something else forced a refresh.
> ```

But there's still a window, right? That's why you trigger a reload of the page list when the websocket actually opens?
It's a perfectly good change...I don't see any reason not to open the socket ASAP...but the comment suggests that this solves a problem which I think it only reduces.
___
*[`src/BloomBrowserUI/bookEdit/pageThumbnailList/pageThumbnailList.tsx` line 627 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7952#-OusIdMsBIFLH9wRnw4R:-OusIdMsBIFLH9wRnw4S:b-loivm5) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/def519c8ab888cb446c4a05eff040b74dda4c6f1/src/BloomBrowserUI/bookEdit/pageThumbnailList/pageThumbnailList.tsx#L627)):*
> ```XML
>     // of pages changes (e.g., adding a page or re-ordering them).
>     useEffect(() => {
>         get("pageList/pages", (response) => {
> ```

I'm not sure if it's worth it, but if I'm understanding right, the very first get() that happens is wasted: it will always be repeated when the websocket is open. So could we just return early if reloadValue is zero?
___
*[`src/BloomExe/Edit/EditingModel.cs` line 631 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7952#-OusGx6u816m7DqIsvmT:-OusGx6u816m7DqIsvmU:b-gesbvb) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/def519c8ab888cb446c4a05eff040b74dda4c6f1/src/BloomExe/Edit/EditingModel.cs#L631)):*
> ```Smalltalk
>                     // insert and raced with those deferred notifications: a websocket message
>                     // arriving while the iframe was mid-reload was silently dropped, leaving the
>                     // list permanently stale.
> ```

Navigating the iframe to a new URL is happening less often than the old location.reload, but won't it also cause a complete repaint and a period when socket notifications are ignored? Do we still have a race condition when a new page really does introduce a new stylesheet?

andrew-polk (commented):
> @andrew-polk made 3 comments.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7952)* status: all files reviewed, 3 unresolved discussions (waiting on JohnThomson).

___
*[`src/BloomBrowserUI/bookEdit/pageThumbnailList/pageThumbnailList.tsx` line 602 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7952#-OusHzud8ql2JNivV-xM:-Ouscdsq6_JjPLS0Kexw:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/def519c8ab888cb446c4a05eff040b74dda4c6f1/src/BloomBrowserUI/bookEdit/pageThumbnailList/pageThumbnailList.tsx#L602)):*

But there's still a window, right? That's why you trigger a reload of the page list when the websocket actually opens?
It's a perfectly good change...I don't see any reason not to open the socket ASAP...but the comment suggests that this solves a problem which I think it only reduces.
</blockquote></details>

Done.
___
*[`src/BloomBrowserUI/bookEdit/pageThumbnailList/pageThumbnailList.tsx` line 627 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7952#-OusIdMsBIFLH9wRnw4R:-Ousd6vq6xaVnOswqnnt:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/def519c8ab888cb446c4a05eff040b74dda4c6f1/src/BloomBrowserUI/bookEdit/pageThumbnailList/pageThumbnailList.tsx#L627)):*

I'm not sure if it's worth it, but if I'm understanding right, the very first get() that happens is wasted: it will always be repeated when the websocket is open. So could we just return early if reloadValue is zero?
</blockquote></details>

Done.
___
*[`src/BloomExe/Edit/EditingModel.cs` line 631 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7952#-OusGx6u816m7DqIsvmT:-OusditM9prfcfHmrdwu:bm9ro1j) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/def519c8ab888cb446c4a05eff040b74dda4c6f1/src/BloomExe/Edit/EditingModel.cs#L631)):*

Navigating the iframe to a new URL is happening less often than the old location.reload, but won't it also cause a complete repaint and a period when socket notifications are ignored? Do we still have a race condition when a new page really does introduce a new stylesheet?
</blockquote></details>

Comment added for clarity.

JohnThomson (approved):
Very nice!

> @JohnThomson reviewed 2 files and all commit messages, made 1 comment, and resolved 3 discussions.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7952#-:-Ousr55I8go3p4zkw-ps:b-j7qk9t)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on andrew-polk).

PR #7992, opened 2026-06-23
Fix BL-16459 Copying Text Failed

andrew-polk:
https://issues.bloomlibrary.org/youtrack/issue/BL-16459

The common/clipboardText endpoint copies/pastes text via the Windows clipboard. When another
program is momentarily holding the clipboard, Windows throws
System.Runtime.InteropServices.ExternalException ("Requested Clipboard operation did not
succeed"). The handler reported this with ModalIf.All, popping a scary "report this problem"
dialog the user could do nothing about but retry.

Rather than try to infer the cause from the exception type, this stops treating any clipboard
failure as modal-worthy: whatever went wrong, all the user can do is try again. So both the copy
(SetText) and paste (GetText) paths now report with ModalIf.None/PassiveIf.All, which shows a
toast instead of a dialog. showSendReport stays at its default of true, so the toast still
carries a "Report" link if the user does want to tell us about it, and Sentry still hears about
it either way. (An earlier revision of this PR caught ExternalException specifically and kept the
modal for everything else; John's review asked us not to guess at the cause, so the special case
is gone and the pre-existing generic catch does the toast.)

While here, localize the two messages, which were previously hardcoded English ("Error pasting
text"/"Error copying text"): new low-priority strings EditTab.CopyTextFailed and
EditTab.PasteTextFailed.

One consequence worth noting: with the modal threshold now ModalIf.None, NonFatalProblem no
longer calls Analytics.ReportException for these, so clipboard failures stop counting toward the
analytics exception total. The log and Sentry still receive them.

[Devin review](https://devinreview.com/BloomBooks/BloomDesktop/pull/7992)

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7992)

### DIFF
--- a/DistFiles/localization/en/BloomLowPriority.xlf
+++ b/DistFiles/localization/en/BloomLowPriority.xlf
@@ -353,6 +353,16 @@
         <note>ID: CollectionTab.BookMenu.MoveToCurrentCollection</note>
         <note>{0} will be replaced with the name of the collection that the user currently has open for editing.</note>
       </trans-unit>
+      <trans-unit id="EditTab.CopyTextFailed" translate="no" sil:dynamic="true">
+        <source xml:lang="en">Bloom was not able to copy that.</source>
+        <note>ID: EditTab.CopyTextFailed</note>
+        <note>Shown as a toast when Bloom fails to put text on the computer's clipboard, usually because some other program is momentarily holding the clipboard.</note>
+      </trans-unit>
+      <trans-unit id="EditTab.PasteTextFailed" translate="no" sil:dynamic="true">
+        <source xml:lang="en">Bloom was not able to paste.</source>
+        <note>ID: EditTab.PasteTextFailed</note>
+        <note>Shown as a toast when Bloom fails to read text from the computer's clipboard, usually because some other program is momentarily holding the clipboard.</note>
+      </trans-unit>
     </body>
   </file>
 </xliff>
--- a/src/BloomExe/web/controllers/CommonApi.cs
+++ b/src/BloomExe/web/controllers/CommonApi.cs
@@ -130,10 +130,17 @@ namespace Bloom.web.controllers
                                     // Need to make sure to handle exceptions.
                                     // If the worker thread dies with an unhandled exception,
                                     // it causes the whole program to immediately crash without opportunity for error reporting
+                                    // A clipboard failure is not worth a modal problem-report dialog: whatever the
+                                    // cause, all the user can do is try again (BL-16459). So we just toast, which
+                                    // still offers a "Report" link and still tells Sentry.
                                     NonFatalProblem.Report(
-                                        ModalIf.All,
-                                        PassiveIf.None,
-                                        "Error pasting text",
+                                        ModalIf.None,
+                                        PassiveIf.All,
+                                        LocalizationManager.GetDynamicString(
+                                            "BloomLowPriority",
+                                            "EditTab.PasteTextFailed",
+                                            "Bloom was not able to paste."
+                                        ),
                                         exception: e
                                     );
                                 }
@@ -161,10 +168,17 @@ namespace Bloom.web.controllers
                                         // Need to make sure to handle exceptions.
                                         // If the worker thread dies with an unhandled exception,
                                         // it causes the whole program to immediately crash without opportunity for error reporting
+                                        // A clipboard failure is not worth a modal problem-report dialog: whatever the
+                                        // cause, all the user can do is try again (BL-16459). So we just toast, which
+                                        // still offers a "Report" link and still tells Sentry.
                                         NonFatalProblem.Report(
-                                            ModalIf.All,
-                                            PassiveIf.None,
-                                            "Error copying text",
+                                            ModalIf.None,
+                                            PassiveIf.All,
+                                            LocalizationManager.GetDynamicString(
+                                                "BloomLowPriority",
+                                                "EditTab.CopyTextFailed",
+                                                "Bloom was not able to copy that."
+                                            ),
                                             exception: e
                                         );
                                     }


hatton (changes requested):
> @hatton made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7992)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on andrew-polk).

___
*[`src/BloomExe/web/controllers/CommonApi.cs` line 122 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7992#-OvpRDrv98I92LwcbUVB:-OvpRDrv98I92LwcbUVC:bk5wz2r) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/190d83564a2399cee478efd1912962ba4da63ef4/src/BloomExe/web/controllers/CommonApi.cs#L122)):*
> ```Smalltalk
>                                     result = PortableClipboard.GetText();
>                                 }
>                                 catch (ExternalException e)
> ```

If I'm reading this right, we are assuming that the problem was that the clipboard is busy. Surely there are other problems possible?
___
*[`src/BloomExe/web/controllers/CommonApi.cs` line 293 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7992#-OvpRIla3kZ0hufTGayH:-OvpRIla3kZ0hufTGayI:b-pddabe) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/190d83564a2399cee478efd1912962ba4da63ef4/src/BloomExe/web/controllers/CommonApi.cs#L293)):*
> ```Smalltalk
>         /// dialog we just show a gentle toast (and still log/send to Sentry so we can track it).
>         /// </summary>
>         private static void ReportClipboardBusy(ExternalException e)
> ```

This should just be a toast. And if you click on it then I think you can get an error report and then they could send the error report. 

"Bloom was not able to copy that" / "Bloom was not able to paste".

andrew-polk (commented):
> @andrew-polk+AI made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7992)* status: 0 of 2 files reviewed, 3 unresolved discussions (waiting on andrew-polk and hatton).

___
*[`src/BloomExe/web/controllers/CommonApi.cs` line 122 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7992#-OvpRDrv98I92LwcbUVB:-OyL9FHs9PW0D8b2Psxl:b1v4knu) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/190d83564a2399cee478efd1912962ba4da63ef4/src/BloomExe/web/controllers/CommonApi.cs#L122)):*

If I'm reading this right, we are assuming that the problem was that the clipboard is busy. Surely there are other problems possible?
</blockquote></details>

[Claude Opus 5] Good point, and you're right. Rather than defend the guess, I removed it: there is no longer a special `catch (ExternalException)` at all. The pre-existing `catch (Exception)` in both the copy and paste paths now does the gentle toast, on the reasoning that no clipboard failure is worth a modal problem-report dialog -- whatever the cause, the only thing the user can do is try again. So we no longer infer "the clipboard was busy" from the exception type, and the message no longer claims some other program is interfering. The `ReportClipboardBusy` helper is gone too, which makes the change smaller than it was in r1.
___
*[`src/BloomExe/web/controllers/CommonApi.cs` line 293 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7992#-OvpRIla3kZ0hufTGayH:-OyL9TCw3Hp3D5X9_uRW:b45jgkc) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/190d83564a2399cee478efd1912962ba4da63ef4/src/BloomExe/web/controllers/CommonApi.cs#L293)):*

This should just be a toast. And if you click on it then I think you can get an error report and then they could send the error report. 

"Bloom was not able to copy that" / "Bloom was not able to paste".
</blockquote></details>

[Claude Opus 5] Done, both parts.

Toast with a report path: dropped ``showSendReport: false``, so it keeps its default of true. You remembered right -- ``NonFatalProblem.ShowToast`` attaches a clickable "Report" action whenever ``showSendReport`` is true, and with ``ModalIf.None``/``PassiveIf.All`` there is no dialog and no yellow screen, so it is just a toast the user can click to send a report if they want to. Sentry still gets it either way.

Wording: used yours, as two new strings rather than the borrowed image-copy one -- ``EditTab.CopyTextFailed`` ("Bloom was not able to copy that.") and ``EditTab.PasteTextFailed`` ("Bloom was not able to paste."), both in ``BloomLowPriority.xlf``. Side benefit: these two messages were previously hardcoded English ("Error copying text"/"Error pasting text"), so they are localized now for the first time.

One consequence worth flagging: because the modal threshold is now ``ModalIf.None``, ``NonFatalProblem`` no longer calls ``Analytics.ReportException`` for these, so clipboard failures will stop showing up in the analytics exception count. The log and Sentry still receive them, which I assumed is the tracking we actually care about -- say the word if you would rather keep the analytics hit.

The branch is also rebased onto current master and force-pushed, so the line numbers here have moved.

src/BloomExe/web/controllers/CommonApi.cs
```diff
@@ -130,10 +130,17 @@ public void RegisterWithApiHandler(BloomApiHandler apiHandler)
...
+                                        ModalIf.None,
+                                        PassiveIf.All,
+                                        LocalizationManager.GetDynamicString(
+                                            "BloomLowPriority",
+                                            "EditTab.PasteTextFailed",
+                                            "Bloom was not able to paste."
+                                        ),
                                         exception: e
                                     );
                                 }
```
andrew-polk, line 146: [Devin] **Investigate**: Catch downgrades all exceptions, not just the transient clipboard-busy one described

The PR description states the intent is to catch `ExternalException` separately and downgrade only that to a non-modal toast, while "any other unexpected exception keeps the existing modal report." The implemented code instead catches generic `Exception` in both the GET (`CommonApi.cs:128`) and POST (`CommonApi.cs:166`) paths and unconditionally downgrades to `ModalIf.None`/`PassiveIf.All`. This means genuinely unexpected, non-transient clipboard errors (e.g. a real regression) will now only surface as a passive toast rather than a modal report dialog, reducing visibility. It still reports to Sentry via `ReportSentryOnly` (skipSentryReport defaults false), but `Analytics.ReportException` is skipped since `modalThreshold == ModalIf.None` (`NonFatalProblem.cs:105-108`). Note the prior code also caught generic `Exception`, so this is a consistent simplification rather than a transformation bug, but reviewers may want to confirm the broader downgrade is intended given the description's narrower stated scope.

andrew-polk: [Claude Opus 5] Not an issue — the broad downgrade is deliberate, and Devin caught a genuinely stale PR description, which I have now corrected.

What happened: the first revision of this PR did exactly what the old description said (special-case `ExternalException`, keep the modal for everything else). John's review asked us not to infer the cause from the exception type — "Surely there are other problems possible?" — so the special case was removed and the pre-existing generic `catch (Exception)` now does the toast. The description had not caught up; it now describes what the code actually does.

On the substance: yes, an unexpected clipboard error now shows a toast rather than a modal dialog, and that is the intent. Whatever the underlying cause, the user's only recourse for a failed copy or paste is to try again, so a modal problem-report dialog buys them nothing. They are not left without a path to tell us, either — the toast keeps its clickable "Report" link (`showSendReport` is at its default of `true`), and Sentry gets the exception regardless.

Good spot on `Analytics.ReportException` being skipped when the modal threshold is `ModalIf.None` — that is a real side effect of the change. It is now called out in the PR description, and I have raised it with the developer in case we would rather keep the analytics count for these.

andrew-polk:
[Claude Opus 5] Consulted Devin on 2026-07-24 22:15 UTC, up to commit `51a754d89e4e1485813774768dbad4e8bd007731`.

It found no bugs and one thing worth investigating — that the PR description still described the older, narrower approach to catching clipboard errors. That was a fair catch: the description was stale, so I rewrote it to match the code, and the flag is [answered and resolved in its own thread](https://github.com/BloomBooks/BloomDesktop/pull/7992#discussion_r3648454655).

Two further informational items were carried over from the previous commit's review; both are already addressed by this revision (the borrowed image-copy string is gone in favour of two dedicated ones, and the empty-string paste result on failure is pre-existing behaviour this PR does not change). Greptile re-reviewed this commit and passed; CI is green apart from the Reviewable gate, which is waiting on John's own discussions.

DistFiles/localization/en/BloomLowPriority.xlf
```diff
@@ -353,6 +353,16 @@
...
+      <trans-unit id="EditTab.CopyTextFailed" translate="no" sil:dynamic="true">
+        <source xml:lang="en">Bloom was not able to copy that.</source>
+        <note>ID: EditTab.CopyTextFailed</note>
+        <note>Shown as a toast when Bloom fails to put text on the computer's clipboard, usually because some other program is momentarily holding the clipboard.</note>
+      </trans-unit>
+      <trans-unit id="EditTab.PasteTextFailed" translate="no" sil:dynamic="true">
+        <source xml:lang="en">Bloom was not able to paste.</source>
+        <note>ID: EditTab.PasteTextFailed</note>
+        <note>Shown as a toast when Bloom fails to read text from the computer's clipboard, usually because some other program is momentarily holding the clipboard.</note>
+      </trans-unit>
```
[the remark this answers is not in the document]
andrew-polk: [Claude Opus 5] This is our deliberate house convention for brand-new strings, so no change here — and the factual premise isn't quite right either.

`translate="no"` on a *new* entry is what this repo asks for. From `AGENTS.md`: "Mark new XLF entries `translate="no"` unless instructed otherwise", and the repo's own `.github/skills/xlf-strings/SKILL.md` repeats it: "Always mark new entries `translate="no"` unless instructed otherwise." The reasoning is that we don't want to spend translators' effort on wording that may still change before it ships; the attribute gets lifted once the string has settled. In the meantime the English source text in the code is what users see, which is why the fallback string is spelled out at both call sites.

On the claim that these are the only such entries: `translate="no"` appears on 137 entries across the English localization files (36 in `Bloom.xlf`, 101 in `BloomMediumPriority.xlf`), and 4 of those also carry `sil:dynamic="true"` — for example `PublishTab.Apps-tooltip` in `BloomMediumPriority.xlf`. It's true that these are the first two in `BloomLowPriority.xlf`, but that's just because this file is small and newer, not a departure from how the project works.

Worth noting for future reviews of this repo: new XLF entries carrying `translate="no"` is expected, not a defect.

src/BloomExe/web/controllers/CommonApi.cs
andrew-polk: [Claude Opus 5] Developer decision, closing the loop on the analytics question I raised above: leave it as is.

Clipboard copy/paste failures will no longer be counted by the analytics exception reporting, and that is accepted deliberately rather than overlooked. Sentry and the log still receive every one of these, and Sentry is where we actually look when we want to know whether something is happening to real users — so the tracking that matters is intact, and adding an explicit analytics call back in isn't worth the extra code.

DistFiles/localization/en/BloomLowPriority.xlf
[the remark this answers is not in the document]
andrew-polk: [Claude Opus 5] Developer decision on this one, to close it out properly: leaving `translate="no"` in place.

So to confirm for future reviews of this repo: the two new strings will show in English until the attribute is lifted, and that is the intended state for newly added strings here, per the convention in AGENTS.md. This is not something to flag as a defect on this repo when the entry is new.

andrew-polk (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7992)* status: 0 of 2 files reviewed, 2 unresolved discussions (waiting on andrew-polk and hatton).

JohnThomson (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7992)* status: all files reviewed, 2 unresolved discussions (waiting on hatton).

PR #7999, opened 2026-06-24
Save the current book fully before reporting problem (BL-16382)

JohnThomson:
[Devin review](https://devinreview.com/BloomBooks/BloomDesktop/pull/7999)

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7999)

### DIFF
--- a/src/BloomExe/Edit/EditingModel.cs
+++ b/src/BloomExe/Edit/EditingModel.cs
@@ -1549,36 +1549,34 @@ namespace Bloom.Edit
         private bool _nextSaveMustBeFull; // review: store in state machine?
 
         /// <summary>
-        /// Request the needed data to do a save, then when the contents of the current page have been saved,
-        /// do the given action. The result of the action is a page ID which we will then navigate to, or null
-        /// to show a blank screen if we are leaving the edit tab.
-        /// If we are not in the right state to save, doAfterSaving() will not be called at all,
-        /// but instead doIfNotInRightStateToSave() will be called. Usually the latter does nothing,
-        /// but we are deliberately not making it optional to make sure it gets thought about.
-        /// Usually, the book will be saved to disk after executing the action, but before navigating to
-        /// the new page. Sometimes the action needs to save the book itself, in which case it can prevent
-        /// a further Save() by setting skipSaveToDisk to true. (Or just possibly, we might not want the Save
-        /// to go all the way to disk, especially if we aren't going to switch pages.)
-        /// Returns true if we're in a valid state to save, false if we're not. In the latter case, doAfterSaving
-        /// will not be called (even later).
+        /// Request the needed data to do a save, then when the in-memory DOM has been updated from the browser,
+        /// call doBeforeSaveToDisk. Its return value is a page ID to navigate to afterward (or null to show a
+        /// blank screen when leaving the edit tab). Unless skipSaveToDisk is true, the book is then saved to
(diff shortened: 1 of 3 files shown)


andrew-polk (changes requested):
> @andrew-polk reviewed 3 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7999)* status: all files reviewed, 1 unresolved discussion (waiting on JohnThomson).

___
*[`src/BloomExe/Edit/EditingModel.cs` line 1570 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7999#-Ow0BsLGCR5hrbhhZYJw:-Ow0BsLGCR5hrbhhZYJx:bmy3jot) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/7c99c4d9e7067af6953dc0ab0c3d1130a94e50b5/src/BloomExe/Edit/EditingModel.cs#L1570)):*
> ```Smalltalk
>             bool skipSaveToDisk = false,
>             Action failureAction = null,
>             Action doAfterSaveToDisk = null
> ```

I wonder if a better name would reflect
1. We call it even if we didn't save to disk (skipSave true)
2. We call it even if save to disk throws

What about something like 
- `doFinally`
- `doAfterSaveAttempt` (doesn't really match with skipping save to disk)
- `doAfterSaveToDiskAttempted` (doesn't really match with skipping save to disk)
- `doBeforeNavigate`

Or should we really not be calling it if the save fails?

Regardless, we should document the behavior in the comment if save fails.

JohnThomson (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7999)* status: all files reviewed, 1 unresolved discussion (waiting on andrew-polk).

___
*[`src/BloomExe/Edit/EditingModel.cs` line 1570 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7999#-Ow0BsLGCR5hrbhhZYJw:-Ow4TuCuAHgpYq10T2UF:b-tyn589) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/7c99c4d9e7067af6953dc0ab0c3d1130a94e50b5/src/BloomExe/Edit/EditingModel.cs#L1570)):*

I wonder if a better name would reflect
1. We call it even if we didn't save to disk (skipSave true)
2. We call it even if save to disk throws

What about something like 
- `doFinally`
- `doAfterSaveAttempt` (doesn't really match with skipping save to disk)
- `doAfterSaveToDiskAttempted` (doesn't really match with skipping save to disk)
- `doBeforeNavigate`

Or should we really not be calling it if the save fails?

Regardless, we should document the behavior in the comment if save fails.
</blockquote></details>

Don't like any of those much. The distinction I want to emphasize compared to doBeforeSaveToDisk is that changes are saved before this runs.
And for the current purpose, I'm not sure we really want it to run if some exception is thrown while saving. The user will probably be trying to report that crash.
I decided to keep the name but make it true: it only gets called if we really did save (skipSaveToDisk false and succeeded in doing so).

andrew-polk (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/7999)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on JohnThomson).

PR #8019, opened 2026-06-30
Collapse nested character markup to single level (BL-16387)

StephenMcConnel:
For example, <em><em>...</em></em> goes to <em>...</em>.

[Devin review](https://devinreview.com/BloomBooks/BloomDesktop/pull/8019)

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8019)

### DIFF
--- a/src/BloomExe/Book/Book.cs
+++ b/src/BloomExe/Book/Book.cs
@@ -2628,7 +2628,13 @@ namespace Bloom.Book
 
         /// <summary>
         /// Convert old &lt;b&gt; and &lt;i&gt; to &lt;strong&gt; and &lt;em&gt; respectively.
-        /// Also remove instances like &lt;/b&gt;&lt;b&gt; altogether since such markup is redundant.
+        /// Remove instances like &lt;/b&gt;&lt;b&gt; altogether since such markup is redundant.
+        /// Cleanup instances like &lt;strong&gt;&lt;strong&gt;...&lt;/strong&gt;&lt;/strong&gt; to just &lt;strong&gt;...&lt;/strong&gt;.
+        /// Remove empty character markup like &lt;strong&gt;&lt;/strong&gt;, and seemingly empty markup that contains
+        /// only zero-width characters.
+        /// Doubled (nested) markup is handled only to one level, but that should be enough since Bloom can't
+        /// introduce such markup on its own.  Handling general nested markup would require more than regular
+        /// expressions to handle properly.
         /// </summary>
         public static void UpdateCharacterStyleMarkup(HtmlDom bookDOM)
         {
@@ -2667,10 +2673,33 @@ namespace Bloom.Book
                     "<strong>$1</strong>",
                     RegexOptions.CultureInvariant | RegexOptions.IgnoreCase
                 );
+                if (inner.IndexOf("<b>", StringComparison.OrdinalIgnoreCase) >= 0) // handle one level of nesting
+                    inner = Regex.Replace(
+                        inner,
+                        @"<b>(.*?)</b>",
+                        "<strong>$1</strong>",
+                        RegexOptions.CultureInvariant | RegexOptions.IgnoreCase
+                    );
                 inner = Regex.Replace(
                     inner,
                     @"<i>(.*?)</i>",
                     "<em>$1</em>",
+                    RegexOptions.CultureInvariant | RegexOptions.IgnoreCase
+                );
+                if (inner.IndexOf("<i>", StringComparison.OrdinalIgnoreCase) >= 0) // handle one level of nesting
+                    inner = Regex.Replace(
+                        inner,
+                        @"<i>(.*?)</i>",
+                        "<em>$1</em>",
+                        RegexOptions.CultureInvariant | RegexOptions.IgnoreCase
+                    );
+                // Replace doubled (nested) markup with single markup.  This shouldn't happen, but it
+                // has been seen in the wild, possibly as a result of pasting text. (BL-16378)
+                // Only one level of nesting is handled, but that should (almost always) be enough.
+                inner = Regex.Replace(
+                    inner,
+                    @"<(strong|em|sup|u)><\1>(.*?)</\1></\1>",
+                    "<$1>$2</$1>",
                     RegexOptions.CultureInvariant | RegexOptions.IgnoreCase
                 );
                 // Remove empty (or essentially empty) character markup tags.  (BL-16387)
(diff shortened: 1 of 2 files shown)


JohnThomson (approved):
:lgtm:, but a couple of ideas to consider. If you don't think it's worth it you can merge as-is.

> @JohnThomson reviewed 2 files and all commit messages, and made 3 comments.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8019#-:-OwUaveb0IasR-7_pHal:bm4xz7i)* status: all files reviewed, 2 unresolved discussions (waiting on StephenMcConnel).

___
*[`src/BloomExe/Book/Book.cs` line 2631 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8019#-OwU_iSy-pFXVSCy4r3X:-OwU_iSy-pFXVSCy4r3Y:bswmtuk) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/6c5d4096264fbd9deea81063a604585badda11e1/src/BloomExe/Book/Book.cs#L2631)):*
> ```Smalltalk
>         /// <summary>
>         /// Convert old &lt;b&gt; and &lt;i&gt; to &lt;strong&gt; and &lt;em&gt; respectively.
>         /// Also remove instances like &lt;/b&gt;&lt;b&gt; altogether since such markup is redundant.
> ```

Comment is rather specific, so needs update. 
___
*[`src/BloomExe/Book/Book.cs` line 2695 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8019#-OwUaC4zDOrsV29Opf2Y:-OwUaC4zDOrsV29Opf2Z:b-h8ekpy) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/6c5d4096264fbd9deea81063a604585badda11e1/src/BloomExe/Book/Book.cs#L2695)):*
> ```Smalltalk
>                 inner = Regex.Replace(
>                     inner,
>                     @"<(strong|em|sup|u)>(<\1>.*?</\1>)</\1>",
> ```

This also only handles things that are doubled right at the boundary. Maybe that's enough, but I'm wondering if we should handle the case that there is text inside the outer element but outside the inner one.

StephenMcConnel (commented):
> @StephenMcConnel made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8019)* status: all files reviewed, 2 unresolved discussions (waiting on JohnThomson).

___
*[`src/BloomExe/Book/Book.cs` line 2631 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8019#-OwU_iSy-pFXVSCy4r3X:-OwUmtbO8FM2x7Ws7jzM:b5coax1) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/6c5d4096264fbd9deea81063a604585badda11e1/src/BloomExe/Book/Book.cs#L2631)):*

Comment is rather specific, so needs update.
</blockquote></details>

Done.  I added to the comment.
___
*[`src/BloomExe/Book/Book.cs` line 2695 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8019#-OwUaC4zDOrsV29Opf2Y:-OwUn62m2ZMH1pXvSj9Y:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/6c5d4096264fbd9deea81063a604585badda11e1/src/BloomExe/Book/Book.cs#L2695)):*

This also only handles things that are doubled right at the boundary. Maybe that's enough, but I'm wondering if we should handle the case that there is text inside the outer element but outside the inner one.
</blockquote></details>

Done.

JohnThomson (changes requested):
> @JohnThomson reviewed 2 files and all commit messages, made 1 comment, and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8019)* status: all files reviewed, 1 unresolved discussion (waiting on StephenMcConnel).

___
*[`src/BloomExe/Book/Book.cs` line 2695 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8019#-OwUaC4zDOrsV29Opf2Y:-Ox2MN_L6y0qNuG32Lth:bl9ng69) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/6c5d4096264fbd9deea81063a604585badda11e1/src/BloomExe/Book/Book.cs#L2695)):*

Done.
</blockquote></details>

I may be misunderstanding. It looks to me as if both the previous and current versions will convert <b><b>abc</b></b> to <b>abc</b>. My point is that neither of them will convert <b>abc<b>def</b>ghi</b> to <b>abcdefghi</b>. And it seems at least plausible that doing so would be a good idea. The inner b doesn't make anything more bold, and may cause unexpected behavior when changing the boldness of things.
I'm open to reasoning that what I'm suggesting is not a good idea, or unnecessary or too difficult.

StephenMcConnel (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8019)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on JohnThomson).

___
*[`src/BloomExe/Book/Book.cs` line 2695 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8019#-OwUaC4zDOrsV29Opf2Y:-Ox2VtzZAtUbZNTr4hfG:b-abg0w2) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/6c5d4096264fbd9deea81063a604585badda11e1/src/BloomExe/Book/Book.cs#L2695)):*

I may be misunderstanding. It looks to me as if both the previous and current versions will convert <b><b>abc</b></b> to <b>abc</b>. My point is that neither of them will convert <b>abc<b>def</b>ghi</b> to <b>abcdefghi</b>. And it seems at least plausible that doing so would be a good idea. The inner b doesn't make anything more bold, and may cause unexpected behavior when changing the boldness of things.
I'm open to reasoning that what I'm suggesting is not a good idea, or unnecessary or too difficult.
</blockquote></details>

Your attempt to illustrate your point didn't come through: you have to escape html markers like &amp;lt;b&amp;gt;.

I've updated the regex to handle what you want. It now collapses a same-tag pair even when there's text on either side of the inner tag, so &lt;strong&gt;abc&lt;strong&gt;def&lt;/strong&gt;ghi&lt;/strong&gt; goes to &lt;strong&gt;abcdefghi&lt;/strong&gt; (and likewise for  &lt;em&gt;/&lt;u&gt;).
After thinking about it, I dropped &lt;sup&gt; from this collapsing, since nested superscript is semantically meaningful (a superscript of a superscript renders higher and smaller), so collapsing these would change how the text displays.

JohnThomson (approved):
> @JohnThomson reviewed 2 files and all commit messages, and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8019)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on StephenMcConnel).

PR #8021, opened 2026-06-30
More cleanup for some weird data states (BL-16454)

JohnThomson:
[Devin review](https://devinreview.com/BloomBooks/BloomDesktop/pull/8021)

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8021)


JohnThomson:
Decided the benefit of patching one broken very old book is not worth the added computation and code complexity.

PR #8029, opened 2026-07-01
Import .bloomSource files into a collection, with duplicate-by-id handling (BL-16502)

hatton:
[Claude Opus 4.8] Draft opened by an automated preflight run.

## What
Adds a collection-screen command to import one or more `.bloomSource` (and legacy `.bloom`) files into the current editable collection — either as editable books or as new derivatives. When an imported book's `bookInstanceId` already matches a book already in the collection, the user is prompted to **replace** the existing book or **add an independent numbered copy** (which gets a fresh id).

Duplicate detection is keyed on **`bookInstanceId`**, not title/folder name — significant because a shared id causes real bugs and controls re-upload to the Bloom library.

## Notable pieces
- `CollectionModel`: import / extract / validate / duplicate-resolution pipeline; a pre-flight `AnyBloomSourceIsAlreadyInCollection` check decides whether to show the duplicate dialog.
- `CollectionApi`: `chooseBloomSourceFilesToImport` / `anyChosenBloomSourceAlreadyInCollection` / `importBloomSource` endpoints.
- `CollectionsTabPane` + `RadioChoiceDialog`: edit-vs-derivative and replace-vs-addCopy dialogs.
- `BloomOpenFileDialog`: multi-select for the import picker.

## Tests
`CollectionModelTests` covers valid import, duplicate detection keyed on `bookInstanceId` (same-title/different-id is NOT a duplicate; same-id/different-title IS), replace/add-copy/cancel resolution, and malformed-file guards. Also fixed `FakeCollectionModel` (a null `BookSelection` was turning every enumerated book into an `ErrorBookInfo`, silently defeating id-based duplicate detection in tests).

## Preflight review fixes included
- Stage extraction on the collection's volume so the move into the collection can't fail cross-volume (system temp is often on a different drive).
- In the Replace case, recycle the existing book only after the imported book is safely in place, so a mid-import failure can't lose the original.

Ref: https://issues.bloomlibrary.org/issue/BL-16502

[Devin review](https://devinreview.com/BloomBooks/BloomDesktop/pull/8029)

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8029)


src/BloomExe/CollectionTab/CollectionModel.cs
```diff
@@ -189,6 +196,536 @@ public void DuplicateBook(Book.Book book)
...
+                        topLevelDirs.Add(parts[0]);
+                }
+
+                // A single book always has files at the root (meta.json, the book htm). A Bloom Pack
+                // instead wraps everything in one collection folder, so it has no top-level files.
+                if (!hasTopLevelFile && topLevelDirs.Count == 1)
+                    throw new BloomSourceImportException(
+                        "This looks like a Bloom Pack (a whole collection), not a single book. "
+                            + "To install a Bloom Pack, double-click it or use \"Open or Create Another Collection\"."
+                    );
```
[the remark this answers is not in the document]
hatton: [Claude Opus 4.8] Good catch — fixed in cf2214512. Once the imported book is moved into place the import has succeeded, so the post-import cleanup (deselect / recycle the replaced book / drop it from the collection) is now wrapped in its own try/catch. On failure it reports a NonFatalProblem ("the book was imported, but Bloom could not remove the existing copy it replaced") and still returns the destination folder — so the import is reported accurately and the newly-imported book is reloaded into view instead of staying invisible.

src/BloomExe/CollectionTab/CollectionModel.cs
```diff
@@ -189,6 +196,560 @@ public void DuplicateBook(Book.Book book)
...
+                    BookStorage.SanitizeNameForFileSystem(baseName),
+                    instanceId,
+                    folderSeparator
+                );
+                destFolder = Path.Combine(editable.PathToDirectory, destName);
+                var renamedHtmlPath = Path.Combine(tempFolder, destName + ".htm");
+                if (!string.Equals(htmlPath, renamedHtmlPath, StringComparison.OrdinalIgnoreCase))
+                    RobustFile.Move(htmlPath, renamedHtmlPath);
+                SIL.IO.RobustIO.MoveDirectory(tempFolder, destFolder);
+                importSucceeded = true;
```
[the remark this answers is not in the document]
hatton: [Claude Opus 4.8] Fixed in 9e9a96db4. `ImportBloomSourceFiles` now threads a `HashSet<string>` of the bookInstanceIds imported so far in the batch into `ImportBloomSourceFileToCollectionFolder`. When a later file's id isn't in the (once-per-batch) reloaded collection but *is* in that set, we route it through the same `makeIndependentCopy` path as Add-a-copy: fresh Guid id + " - Copy-" folder — so the two can't land sharing an id.

On the pre-flight `AnyBloomSourceIsAlreadyInCollection` not showing the duplicate dialog for this case: that's intentional. "Replace" / "Add as a copy" is a choice about a book already in the collection; a book that only exists elsewhere in the same import batch isn't in the collection, so prompting to replace it would be meaningless. We resolve the within-batch collision automatically instead. New test: `ImportBloomSource_SameIdTwiceInOneBatch_SecondBecomesIndependentCopy`.

src/BloomBrowserUI/collectionsTab/RadioChoiceDialog.tsx
```diff
@@ -0,0 +1,94 @@
...
+    open: boolean;
+    title: string;
+    message?: string;
+    options: IRadioChoice[];
+    onClose: (value?: string) => void;
+}> = (props) => {
+    const [choice, setChoice] = useState<string>();
+
+    // Start with nothing selected each time the dialog opens, so OK is disabled until the user picks.
+    useEffect(() => {
```
hatton, line 42: [Devin] **Bug**: Dialog state reset uses useEffect instead of the key-prop pattern required by repo rules

The radio selection is cleared via a `useEffect` that calls `setChoice(undefined)` when `props.open` changes, instead of the key-prop pattern to reset component state.

Per the repo useEffect rule (src/BloomBrowserUI/AGENTS.md -> .github/skills/react-useeffect): prefer alternatives (e.g. remount via a changing `key`), or add a comment justifying *why* useEffect is necessary. The current comment explains what it does, not why.

src/BloomExe/CollectionTab/CollectionModel.cs
```diff
@@ -189,6 +196,652 @@ public void DuplicateBook(Book.Book book)
...
+            foreach (var path in paths)
+            {
+                try
+                {
+                    progress?.MessageWithoutLocalizing($"Importing {Path.GetFileName(path)}...");
+                    if (makeDerivatives)
+                    {
+                        // The derivative path builds its Book in memory and adds it to the collection
+                        // itself, so it needs no reload.
+                        var book = MakeDerivativeFromBloomSourceFile(path);
```
hatton: [Devin] **Investigate**: Batch edit-import reloads collection once at the end, but derivative path does not reload

The edit path accumulates imported folders and calls `ReloadEditableCollection()` once at the end. The derivative path instead calls `AddBookInfo()` in-memory with no reload, and only the last book gets a selection event. This asymmetry is intentional (derivatives build Book objects in memory), but a batch of derivatives never triggers a collection reload event; if any UI component relies on that event rather than the in-memory addition, intermediate derivative books might not update.

src/BloomExe/CollectionTab/CollectionModel.cs
```diff
@@ -189,6 +196,652 @@ public void DuplicateBook(Book.Book book)
...
+
+            // Extract onto the same volume as the collection. The book is later moved into the
+            // collection folder with a directory move, which cannot cross volumes; the system
+            // temp folder is frequently on a different drive than the user's collection (e.g. %TEMP%
+            // on C: but the collection on D: or a removable/network drive), which would make every
+            // edit-mode import fail. Staging in the collection's parent guarantees a same-volume move.
+            // The name is dot-prefixed so it is ignored by the collection's book scan if it ever ends
+            // up inside a scanned folder.
+            var collectionDir = TheOneEditableCollection.PathToDirectory;
+            var stagingParent = Directory.GetParent(collectionDir)?.FullName ?? Path.GetTempPath();
```
hatton: [Devin] **Investigate**: Staging folder falls back to %TEMP% when the collection is at a filesystem root

`ExtractAndPrepareBloomSourceToTemp` deliberately stages in the collection directory's parent (same volume) because `RobustIO.MoveDirectory` cannot cross volumes. But if the collection is at a filesystem root (e.g. `D:\`), `Directory.GetParent()` returns null and the code falls back to `Path.GetTempPath()` -- which may be on a different volume, making the later move fail. Unlikely but possible on removable drives with books stored at the root.

src/BloomBrowserUI/collectionsTab/CollectionsTabPane.tsx
```diff
@@ -197,6 +199,101 @@ export const CollectionsTabPane: React.FunctionComponent = () => {
...
+        "CollectionTab.ImportBloomSource.Replace",
+    );
+    const importReplaceDescription = useL10n(
+        "The current copies will be overwritten.",
+        "CollectionTab.ImportBloomSource.ReplaceDescription",
+    );
+    const importReplaceNeedsCheckoutDescription = useL10n(
+        "All books must be checked out",
+        "CollectionTab.ImportBloomSource.ReplaceNeedsCheckout",
+    );
```
hatton, line 253: [Devin] **Investigate**: Duplicate detection forces edit-mode, preventing derivative imports for mixed batches

When any chosen book is already in the collection, the flow shows only replace-vs-add-copy and always imports in edit mode; the derivative option is never offered. So selecting 5 files where only 1 is a duplicate imports all 5 for editing -- the 4 non-duplicates cannot be imported as derivatives. The code comments say this is intentional; flagging to confirm it is the desired UX.

hatton:
[Claude Opus 4.8] Consulted Devin on 2026-07-06 22:52 UTC up to commit `f63a1b683c`. Findings mirrored as inline review threads: 1 Bug (useEffect/key-prop) + 3 Investigate (derivative-reload asymmetry, staging %TEMP% at drive root, mixed-batch edit-mode). 1 Resolved bug (localization catalog) and ~14 Informational items were not posted (already fixed / low signal).

src/BloomBrowserUI/collectionsTab/RadioChoiceDialog.tsx
hatton: [Claude Opus 4.8] Fixed in 8273d077a — added a comment justifying why a useEffect is used here rather than the key-prop reset (the dialog is permanently mounted and toggled via `open`, so there is no remount to hang a key on), per the repo useEffect rule.

src/BloomExe/CollectionTab/CollectionModel.cs
hatton: [Claude Opus 4.8] Reviewed and kept as-is by design: the derivative path adds each book in memory and the batch selects the last one; there is no UI component that depends on the reload event rather than the in-memory add. (The separate history-event gap this is near was fixed — every derivative now records its Created event immediately.)

src/BloomExe/CollectionTab/CollectionModel.cs
hatton: [Claude Opus 4.8] Fixed in 8273d077a — when the collection is at a filesystem root (no parent dir), staging now falls back to the collection folder itself (dot-prefixed, scan-ignored) rather than %TEMP%, keeping the move same-volume.

src/BloomBrowserUI/collectionsTab/CollectionsTabPane.tsx
hatton: [Claude Opus 4.8] Confirmed intentional and kept as-is: this is documented in CollectionsTabPane.tsx (we deliberately don't offer edit-vs-derivative when a chosen book is already in the collection; 'Easy to reverse if that turns out to be wrong').

hatton:
[Claude Opus 4.8] Consulted Devin on 2026-07-07 02:34 UTC up to commit `8273d077a` (re-review after implementing overwrite-in-place Replace + preflight fixes). Re-review clean — no new actionable findings: the useEffect Bug and the two Investigate flags were already mirrored and resolved (fixed or kept by design), and Devin marks the Replace-path rollback concern Resolved. Remaining items are Informational only.

src/BloomBrowserUI/collectionsTab/CollectionsTabPane.tsx
```diff
@@ -286,6 +371,33 @@ export const CollectionsTabPane: React.FunctionComponent = () => {
...
+                handleClose();
+                // Choose the file(s) first; the reply tells us both the chosen paths and whether any
+                // chosen book is already in the collection, so we can show the one dialog that fits.
+                //
+                // Note that when a chosen book is already in the collection we deliberately do NOT
+                // offer the edit-vs-derivative choice: the duplicate dialog always imports in edit
+                // mode, so you can't make a derivative of a book that's already in your collection. We
+                // couldn't think of a plausible scenario where you have a book and a true derivative
+                // of it in the same collection, so this path isn't worth the extra complexity. Easy
+                // to reverse if that turns out to be wrong.
```
hatton, line 379: [Claude Opus 4.8 relaying Devin] **Investigate**: Duplicate detection forces edit-mode, silently preventing derivative imports for mixed batches

When any book in the chosen batch is already in the collection, chooseImportDialog (CollectionsTabPane.tsx:248-253) shows only Dialog 2 (replace vs. add-copy) and always sends mode=edit. The user never gets the derivative option. So if a user selects 5 .bloomSource files and only 1 is a duplicate, all 5 are imported for editing — the 4 non-duplicates cannot be imported as derivatives. The comments say this is intentional ("The user never sees both"), but it's a significant UX limitation worth confirming is desired, especially for users who routinely work with derivatives.

src/BloomExe/web/controllers/CollectionApi.cs
```diff
@@ -311,6 +311,63 @@ public void RegisterWithApiHandler(BloomApiHandler apiHandler)
...
+                    request.ReplyWithJson(
+                        new
+                        {
+                            files,
+                            anyDuplicates = _collectionModel.AnyBloomSourceIsAlreadyInCollection(
+                                files
+                            ),
+                            canReplace = _collectionModel.AllBloomSourceDuplicatesAreReplaceable(
+                                files
+                            ),
```
hatton, line 335: [Claude Opus 4.8 relaying Devin] **Investigate**: File-picker endpoint does zip I/O on the UI thread for duplicate pre-flight

The chooseBloomSourceFilesToImport/ endpoint is registered with handleOnUiThread: true (needed for the file dialog). After the dialog closes, it immediately calls AnyBloomSourceIsAlreadyInCollection and AllBloomSourceDuplicatesAreReplaceable, both of which open each chosen .bloomSource as a zip and read meta.json. For a large batch this zip I/O happens on the UI thread and could cause a noticeable freeze. Consider moving the duplicate pre-flight off the UI thread (a second async endpoint, or background the zip reads after the dialog closes).

src/BloomExe/CollectionTab/CollectionModel.BloomSourceImport.cs
hatton, line 1: [Claude Opus 4.8 relaying Devin] **Investigate**: Batch edit-import reloads the collection once at the end, but the derivative path does not reload

The edit path accumulates all imported folders and calls ReloadEditableCollection() once (good). The derivative path (MakeDerivativeFromBloomSourceFile) instead calls TheOneEditableCollection.AddBookInfo() to add the book in-memory without a full reload. This asymmetry is intentional, but a batch of derivatives never triggers a collection reload event — if any UI component relies on the reload event rather than the in-memory addition, it might not update. SelectBookOnUiThread updates the last book, but intermediate books in a batch get no individual selection event. Worth confirming all relevant UI updates on the derivative batch path.

hatton:
[Claude Opus 4.8] Consulted Devin on 2026-07-07 (re-review of the stateless-API refactor + master merge, analyzed at commit `83f7cd2`). Outcome:

- **1 Bug** (XLF note order — ID note must precede context note): **fixed** in commit `969e537`.
- **2 Bugs** already resolved in earlier commits (dialog key-prop reset; missing progress-dialog l10n entry) — no action.
- **3 Investigate flags** mirrored as inline review threads for the developer to weigh: UI-thread zip I/O in the duplicate pre-flight, derivative-batch path not firing a collection reload, and mixed-batch imports forcing edit-mode (no derivative option).
- Remaining items are Informational — not posted.

Findings up to `969e537`.

src/BloomExe/CollectionTab/CollectionModel.BloomSourceImport.cs
hatton: doesn't sound important

src/BloomExe/web/controllers/CollectionApi.cs
hatton: doesn't sound important

src/BloomBrowserUI/collectionsTab/CollectionsTabPane.tsx
hatton: intentional

StephenMcConnel (changes requested):
One minor comment on a comment in a test file.  Otherwise looks good.

> @StephenMcConnel reviewed 15 files and all commit messages, and made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8029#-:-Ox2V0Y7DGtDQs5kSVXu:b-uw3u8f)* status: all files reviewed, 8 unresolved discussions (waiting on hatton).

___
*[`src/BloomTests/CollectionTab/CollectionModelTests.cs` line 127 at r6](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8029#-Ox2THMG-MIU6d52hISi:-Ox2THMH0g_-bAVv7TWV:b-5i1utb) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/b8968eb85ec80b9ec63492e378d898388d77183b/src/BloomTests/CollectionTab/CollectionModelTests.cs#L127)):*
> ```Smalltalk
> 
>         // Build a .bloomSource file the same way Bloom's export does: a zip of a book folder's
>         // contents (flat at the root) plus a .bloomCollection settings file at the root.
> ```

The comments are bit misleading -- they caused me to think, we're zipping up files here instead of using the existing method to do so??  I would say something like
// Create a .bloomSource file for unit testing from a minimal set of dummy files.

hatton (commented):
> @hatton made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8029)* status: 14 of 15 files reviewed, 8 unresolved discussions (waiting on hatton and StephenMcConnel).

___
*[`src/BloomTests/CollectionTab/CollectionModelTests.cs` line 127 at r6](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8029#-Ox2THMG-MIU6d52hISi:-Ox6tqjFDgHJaExyutiL:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/b8968eb85ec80b9ec63492e378d898388d77183b/src/BloomTests/CollectionTab/CollectionModelTests.cs#L127)):*

The comments are bit misleading -- they caused me to think, we're zipping up files here instead of using the existing method to do so??  I would say something like
// Create a .bloomSource file for unit testing from a minimal set of dummy files.
</blockquote></details>

Done.

StephenMcConnel (approved):
:lgtm:

> @StephenMcConnel reviewed 1 file and all commit messages, made 1 comment, and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8029#-:-Ox6x4Lq5E6FCN5GD2tv:bnfp4nl)* status: all files reviewed, 7 unresolved discussions (waiting on hatton).

StephenMcConnel (approved):
> @StephenMcConnel resolved 7 discussions.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8029)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on hatton).

StephenMcConnel (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8029#-:-Ox6xeOc92ULtShLkd9J:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on hatton).

PR #8033, opened 2026-07-03
BL-16523 AI Image Editor: whole-book AI image editing + per-user OpenRouter key

hatton:
[Claude Opus 4.8]

Adds the **AI Image Editor** feature to Bloom (experimental, behind the `AiImageEditing` feature flag):

- New C# API `AiImageEditorApi` that launches an in-app editor iframe, shares the whole-book image list, persists editor state/history under a per-book `.ai-image-editor` folder, and commits image replacements book-wide.
- Per-user OpenRouter API key storage (`OpenRouterCredentialStore`, DPAPI-protected) so keys never travel with a book.
- The editor is served by BloomServer from `output/browser/aiImageEditor/`, built from the local `bloom-ai-image-tools` checkout by the `go.sh` launcher (`aiEditorBuild.mjs`), with `go.sh --with` support for developing local library checkouts alongside Bloom.
- Canvas context control + collection Advanced Settings toggle to enable the feature.

Draft opened by the **preflight** workflow. See the decision report (delivered separately) for items awaiting a human decision — notably the CORS/credential-exposure surface and the merge with `master`.

Ref: (no YouTrack id in branch name)

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8033)

[Devin review](https://devinreview.com/BloomBooks/BloomDesktop/pull/8033)


src/BloomExe/web/controllers/AiImageEditorApi.cs
```diff
@@ -0,0 +1,923 @@
...
+            if (_sessionToken == null || session != _sessionToken)
+            {
+                request.Failed(HttpStatusCode.Unauthorized, "Invalid or expired session");
+                return false;
+            }
+
+            return true;
+        }
+
+        private void HandleOAuthCallback(ApiRequest request)
```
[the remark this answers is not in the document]
hatton: [Claude Opus 4.8] Confirmed accurate. This is a deliberate product decision, not an oversight: the only threat here is a process or web page **on the user's own machine** reading their own OpenRouter key while Bloom is running with a book open. We don't consider that low-value, local-only surface worth the added complexity / editor-protocol change to defend against, so the launch payload and the server's permissive CORS are staying as they are. Resolving as won't-fix.

src/BloomExe/web/controllers/AiImageEditorApi.cs
```diff
@@ -0,0 +1,923 @@
...
+            var book = _bookSelection.CurrentSelection;
+            if (book == null)
+            {
+                request.Failed("No book selected");
+                return;
+            }
+
+            // Tear down any previous session.
+            EndSession();
+
```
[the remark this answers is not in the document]
hatton: [Claude Opus 4.8] Acknowledged and accurate. In practice the request/response round-trip orders these accesses (the editor has the session token before it makes any session-gated call), and this overlaps the launch/session model touched by the accepted key-exposure decision above. We're accepting it rather than adding volatile/locking piecemeal now; worth revisiting if that session model is reworked. Resolving as won't-fix for this PR.

src/BloomExe/web/controllers/AiImageEditorApi.cs
```diff
@@ -0,0 +1,923 @@
...
+            if (creator.IndexOf(EditedWithAiCredit, StringComparison.OrdinalIgnoreCase) >= 0)
+                return;
+            element.SetAttribute(
+                "data-creator",
+                string.IsNullOrWhiteSpace(creator)
+                    ? EditedWithAiCredit
+                    : creator + ", " + EditedWithAiCredit
+            );
+        }
+    }
```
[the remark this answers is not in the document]
hatton: [Claude Opus 4.8] Fixed in c22c67f88: the result file is now resolved by id across the allowed image extensions (new `TryFindHistoryResultFile`) instead of assuming `.png`, so a .jpg/.webp result that EnumerateHistoryImages lists commits correctly. (Thread already resolved.)

src/BloomExe/web/controllers/AiImageEditorApi.cs
```diff
@@ -0,0 +1,973 @@
...
+            }
+
+            // Write the new image under a fresh name (preserving the source extension) so
+            // we never clobber a file shared by another slot or serve stale cached bytes.
+            var extension = Path.GetExtension(sourceBytesPath);
+            if (string.IsNullOrEmpty(extension))
+                extension = ".png";
+            var newFileName = "ai-" + Guid.NewGuid().ToString("N") + extension;
+            RobustFile.Copy(sourceBytesPath, Path.Combine(book.FolderPath, newFileName), true);
+            newSrc = newFileName;
```
hatton, line 919: [Claude Opus 4.8 relaying Devin]

[Devin] **Investigate**: Orphaned AI-prefixed image files accumulate in book folder

Each commit in TryApplyReplacement creates a new file with a GUID-based name (ai-<guid>.<ext>) and never deletes the old image file. Over multiple edit sessions, the book folder will accumulate orphaned ai-*.png files. Bloom's existing image cleanup (if any) may not know to clean these up. The .ai-image-editor/history/ folder also grows without bound. This is a storage concern for users who frequently use the AI editor on the same book.

src/BloomBrowserUI/vite.config.mts
```diff
@@ -619,6 +633,13 @@ export default defineConfig(async ({ command }) => {
                                   src: "node_modules/bloom-player/dist/*",
                                   dest: "./bloom-player/dist/",
                               },
+                              // Copy the AI Image Editor's prebuilt app (dist-app/) so Bloom
+                              // serves it at /bloom/aiImageEditor/. Mirrors the dev-time
+                              // staging in go.mjs/aiEditorBuild.mjs.
+                              {
+                                  src: "node_modules/bloom-ai-image-tools/dist-app/*",
+                                  dest: "./aiImageEditor/",
+                              },
```
hatton, line 642: [Claude Opus 4.8 relaying Devin]

[Devin] **Investigate**: Vite static copy glob for AI editor dist-app may need adjustment when package is published

The glob pattern node_modules/bloom-ai-image-tools/dist-app/* (in the structured:false block) — a typical Vite app build produces an assets/ subdirectory alongside index.html. Whether * copies subdirectories recursively depends on vite-plugin-static-copy / fast-glob directory handling. The existing bloom-player copy uses the same pattern and works, suggesting directories are handled. Verify when bloom-ai-image-tools is actually published and added as a dependency: if its dist-app/ has an assets/ subdirectory, confirm output/browser/aiImageEditor/ is complete.

hatton:
[Claude Opus 4.8] Consulted Devin on 2026-07-09 up to commit c22c67f88. Result: 0 bugs, 2 Investigate flags (orphaned ai-* file accumulation; vite dist-app copy glob to verify on publish) — both posted as inline threads above. 7 informational items skipped.

hatton:
[Claude Opus 4.8] Consulted Devin on 2026-07-10 up to commit `118aa43e5`. Result: 0 bugs, 3 Investigate flags. Two are the previously-mirrored, still-open threads (orphaned ai-image files; Vite static-copy glob). The third is new — "OAuth callback reflects the error query param without HTML-encoding" (AiImageEditorApi.cs:420) — and is already mitigated: ReplyWithText forces Content-Type: text/plain, so the reflected value can't execute. No new thread opened. All Informational items skipped.

src/BloomExe/web/controllers/AiImageEditorApi.cs
hatton: [Claude Opus 4.8 relaying Devin] ✅ Fixed in commit `87637f860`, and Devin's re-review of that commit now marks this finding **Resolved**. On commit, off-page image replacements whose superseded file is one we generated (`ai-image*`) and that nothing else references are now deleted (`DeleteSupersededAiImageFiles` / `CollectReferencedImageFileNames` in AiImageEditorApi.cs). Resolving.

hatton:
[Claude Opus 4.8] Consulted Devin on 2026-07-10 21:05Z up to commit `87637f860`. Result: 1 Bug (CORS preflight on the editor file endpoint — dev/LINKED-mode only; being handled with the developer per their standing preference on localhost CORS), 2 Investigate (OAuth callback reflection — already mitigated via text/plain response; Vite static-copy glob — deferred), and the prior 'Orphaned AI-prefixed image files' Investigate is now **Resolved** (fixed in 87637f860; its thread has been resolved). Greptile also re-reviewed this commit with no new findings.

DistFiles/localization/en/BloomMediumPriority.xlf
```diff
@@ -39,6 +39,21 @@
...
+        <source xml:lang="en">Edit Images with AI</source>
+        <note>ID: CollectionSettingsDialog.AdvancedTab.Experimental.EditWithAI</note>
+        <note>This checkbox enables the experimental "Edit with AI" image editor.</note>
+      </trans-unit>
+      <trans-unit id="Feature.AiImageEditing" translate="no">
+        <source xml:lang="en">Edit Images with AI</source>
+        <note>ID: Feature.AiImageEditing</note>
+        <note>Name of the AI image editing feature, used in subscription messages.</note>
+      </trans-unit>
+      <trans-unit id="EditTab.Image.EditWithAI">
```
hatton, line 52: [Claude Opus 4.8 relaying Devin] Devin flags this as missing the repo's default `translate="no"` marker for new strings. This is **intentional**, not an oversight: the AI Image Editor is being prepared for release, and the maintainer explicitly decided (preflight decision F3) to make this user-facing context-menu item ("Edit with AI…") localizable like its sibling image menu items (e.g. `EditTab.Image.ChooseImage`). The xlf-strings convention is "mark new entries translate=no *unless instructed otherwise*" — this is the otherwise. The two sibling AI strings that are still experimental-only (`CollectionSettingsDialog.AdvancedTab.Experimental.EditWithAI`, `Feature.AiImageEditing`) deliberately keep their `translate="no"`.

hatton:
[Claude Opus 4.8] Consulted Devin on 2026-07-10 21:56Z up to commit `5b5c881f5`. Result: the prior CORS-preflight **Bug is now Resolved** (fixed in this commit) and the orphaned-ai-image Investigate remains Resolved. One new Bug — 'new xlf string missing translate=no' on the `EditTab.Image.EditWithAI` entry — is an intentional, maintainer-approved deviation (preflight decision F3: make the menu item localizable); documented inline on the entry. Remaining Investigate flags: OAuth reflection (already mitigated via text/plain; comment added) and the Vite static-copy glob (deferred). All other items Informational. Greptile re-reviewed with no new findings; CI green.

hatton:
[Claude Opus 4.8] Consulted Devin on 2026-07-15 ~12:25Z up to commit `0ad16c6b0`. Re-review clean — **no new findings from this commit** (host-side test additions + behavior-preserving refactors). Pre-existing findings unchanged and already assessed in prior runs: XLF `translate=no` (intentional per F3 — user-facing menu string should be translated; thread left open), Vite copy glob (future note when the editor package is published; thread open), OAuth callback reflection (documented non-issue — text/plain response, no reflected-XSS; F4), orphan ai-image cleanup (fixed; thread resolved). The one finding touching new code — `aiEditorSlotMatching.ts` "greedy first-match assumes stable document order" — is **Informational**; that ordering assumption is intentional and documented in-code, and is covered by aiEditorSlotMatching.test.ts. Greptile: clean on this commit. CI: green.

DistFiles/localization/en/BloomMediumPriority.xlf
hatton: [Claude Opus 4.8] Decision: leaving as is. `EditTab.Image.EditWithAI` is a user-facing menu label ("Edit with AI…") that *should* be translated, so omitting the repo-default `translate=no` marker is intentional and correct here (the marker means "do not translate"). Confirmed with John. Resolving.

src/BloomBrowserUI/vite.config.mts
hatton: [Claude Opus 4.8] Decision: leaving as is. The current static-copy glob correctly stages the editor's dist-app for both dev and build today; adjusting it is only a future task for when `bloom-ai-image-tools` is published as a real dependency, not a defect now. Confirmed with John. Resolving.

src/BloomExe/web/controllers/AiImageEditorApi.cs
```diff
@@ -0,0 +1,1170 @@
...
+            // editor, and the /oauth-result poll that hands it back is session-gated.
+            var error = request.GetParamOrNull("error");
+            var code = request.GetParamOrNull("code");
+
+            _pendingOAuthError = error;
+            _pendingOAuthCode = code;
+
+            if (!string.IsNullOrEmpty(error))
+            {
+                // We echo OpenRouter's `error` string back to the browser. This is safe
```
hatton: [Devin] **Investigate**: OAuth callback reflects query parameter in response without HTML-encoding

The OAuth callback echoes OpenRouter's `error` query-string value into its HTTP response without HTML-encoding it (AiImageEditorApi.cs:420-422).

src/BloomBrowserUI/bookEdit/toolbox/canvas/aiEditorSlotMatching.ts
```diff
@@ -0,0 +1,60 @@
...
+export function matchReplacementsToElements<TReplacement, TElement>(
+    replacements: TReplacement[],
+    ordinalOf: (replacement: TReplacement) => number,
+    wantedFilenameOf: (replacement: TReplacement) => string,
+    candidates: TElement[],
+    candidateFilenameOf: (element: TElement) => string,
+): Array<IReplacementMatch<TReplacement, TElement>> {
+    const used = new Set<TElement>();
+    const matches: Array<IReplacementMatch<TReplacement, TElement>> = [];
+    [...replacements]
```
hatton, line 44: [Devin] **Informational**: Slot matching uses greedy first-match which assumes stable document order

matchReplacementsToElements pairs each replacement to a live image element by filename in ascending ordinal order using first-match, assuming the live page enumerates image elements in the same document order as the saved page's holders (aiEditorSlotMatching.ts:44-58).

src/BloomExe/web/controllers/AiImageEditorApi.cs
```diff
@@ -0,0 +1,1170 @@
...
+    {
+        private readonly BookSelection _bookSelection;
+
+        /// <summary>Set by the EditingView constructor — used at commit time to detect the
+        /// currently-open page (which the live browser owns) so we don't edit it from here.</summary>
+        public EditingView View { get; set; }
+
+        // Minted at launch; torn down on the next launch (see EndSession). The editor's
+        // `cancel`/`commit` are handled in the overlay JS, which removes the overlay.
+        private string _sessionToken;
```
hatton, line 82: [Devin] **Informational**: Session state fields are read/written across threads without synchronization

The session token and book-folder fields are written on the UI thread at launch and read on worker threads that handle file/commit requests, without locking (AiImageEditorApi.cs:84-93).

src/BloomExe/web/controllers/AiImageEditorApi.cs
hatton: [Claude Opus 4.8] Decision: leaving as is. The response is sent as `text/plain` (ReplyWithText), so a browser renders it literally instead of parsing markup — there is no reflected-XSS vector — and this is Bloom's own localhost server. The rationale is documented in an in-code comment at this spot (F4). Resolving.

src/BloomBrowserUI/bookEdit/toolbox/canvas/aiEditorSlotMatching.ts
hatton: [Claude Opus 4.8] Decision: leaving as is. The ordering holds by construction: the host emits each slot's ordinal in the saved page's image-holder document order, which is the same order the live page enumerates, so first-match lands each replacement on the right slot. It is documented in the function's header comment and pinned by aiEditorSlotMatching.test.ts (including the two-identical-filename-slots case). Resolving.

src/BloomExe/web/controllers/AiImageEditorApi.cs
hatton: [Claude Opus 4.8] Decision: leaving as is. Access is effectively safe for this single-user localhost server: the fields are set once at launch (UI thread) and only read afterward; the worst case from a stale read is a request that fails closed with a 401 (invalid/expired session), never data corruption. Not worth adding locking. Resolving.

src/BloomBrowserUI/vite.config.mts
hatton: [Claude Opus 4.8] Correction to my earlier reply. That reply was right that no code change is needed, but its reasoning was wrong on two counts, and it wrongly framed this as future work.

- **This PR *is* the "when published" moment.** `bloom-ai-image-tools` is now a real dependency here (`src/BloomBrowserUI/package.json`: `github:BloomBooks/bloom-ai-image-tools#dist-v0.1.0`), and its installed `dist-app/` does have an `assets/` subdirectory that `index.html` references (`/bloom/aiImageEditor/assets/...`). So Devin's concern is current, not hypothetical.
- **My "bloom-player uses the same pattern and works" premise was false** — bloom-player's `dist` is flat (no subdirectories), so it never exercised directory recursion. And the correct-looking `output/browser/aiImageEditor/` tree on disk was staged by the go.sh launcher's recursive `fs.cpSync` (note the `.staged-package-version` marker), not by the `viteStaticCopy` build path Devin flagged.

**Actual verification (from `vite-plugin-static-copy` source):** the glob runs with `onlyFiles:false, expandDirectories:false`, so `dist-app/*` matches `index.html` and `assets` (the directory as a single unit). With `structured:false` and both matches at top level (empty parent dir), each copies to `aiImageEditor/<name>`, and the copy is fs-extra's recursive `fs.copy`. Result: `aiImageEditor/index.html` + `aiImageEditor/assets/**`, exactly what `index.html` references. `structured:false` does **not** flatten `assets/` (which *would* break the asset URLs), because `expandDirectories:false` keeps the directory as one match.

Verified by static analysis + the installed tree, not by running a build (repo rule forbids `pnpm build` here — it would wreck a running watch). Outcome unchanged: glob is correct as written. Leaving resolved.

hatton: it is a real dependency. We need to look into this.

hatton (commented):
> @hatton reviewed 39 files and made 4 comments.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8033)* status: 39 of 44 files reviewed, 9 unresolved discussions.

___
*[`src/BloomBrowserUI/bookEdit/toolbox/canvas/canvasControlRegistry.ts` line 510 at r4](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8033#-OxbEMFJD-LkhvAyupF4:-OxbEMFJD-LkhvAyupF5:b-6m0stn) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/0ad16c6b022c1bc481b8b799f6c423487336c31c/src/BloomBrowserUI/bookEdit/toolbox/canvas/canvasControlRegistry.ts#L510)):*
> ```XML
>         },
>     },
>     // "Edit with AI…" — the entry point for the AI Image Editor integration.
> ```

Opus: extract all this to its own file.
___
*[`src/BloomBrowserUI/scripts/aiEditorBuild.mjs` line 2 at r4](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8033#-OxbJMJ0A4nmBUjvhXIl:-OxbJMJ0A4nmBUjvhXIm:bmzg6iz) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/0ad16c6b022c1bc481b8b799f6c423487336c31c/src/BloomBrowserUI/scripts/aiEditorBuild.mjs#L2)):*
> ```JavaScript
> /* eslint-env node */
> /* global clearTimeout, process, setTimeout */
> ```

Add a file up here about why this file exists. To my mind this is a temporary thing that we use while we are still programming the interface between Bloom and the Bloom AI image editor. Once that solidifies we should probably delete this and associated stuff that just makes it easier for us to work on both at the same time.
___
*[`src/BloomExe/Properties/Settings.settings` line 133 at r4](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8033#-OxbyWrz24fjlT2kydjI:-OxbyWs-3FBeSybJSqC_:bb84h0q) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/0ad16c6b022c1bc481b8b799f6c423487336c31c/src/BloomExe/Properties/Settings.settings#L133)):*
> ```
>     <Setting Name="OpenRouterAuthMethod" Provider="SIL.Settings.CrossPlatformSettingsProvider" Type="System.String" Scope="User">
>       <Value Profile="(Default)" />
>     </Setting>
> ```

are all three actually used now? I think we're only supporting the api method. In which case we should remove these others (and from Settings.Designer.cs)

src/BloomBrowserUI/vite.config.mts
hatton: [Claude Opus 4.8] Looked into it empirically (couldn't run a full `pnpm build` here -- repo rule against it in this worktree). Read the installed `vite-plugin-static-copy@3.1.3` source: it globs with `expandDirectories:false`, so `dist-app/*` matches `index.html` and the `assets` dir as single units, and with `structured:false` copies each to `aiImageEditor/<name>` via fs-extra's recursive `fs.copy`. I then simulated exactly those two steps against the real installed dist-app: `dist-app/*` -> [assets, index.html], both copied, `assets/` landed 76 files, and all 5 asset URLs that index.html references resolved under the output tree. So the build produces `output/browser/aiImageEditor/{index.html, assets/**}` correctly -- no code change needed.

hatton (commented):
> @hatton+JHAI made 4 comments and resolved 4 discussions.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8033)* status: 29 of 47 files reviewed, 5 unresolved discussions (waiting on hatton).

___
*[`src/BloomBrowserUI/bookEdit/toolbox/canvas/canvasControlRegistry.ts` line 510 at r4](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8033#-OxbEMFJD-LkhvAyupF4:-Oxccp_k1xvdF6iY7mZZ:b59rsk) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/0ad16c6b022c1bc481b8b799f6c423487336c31c/src/BloomBrowserUI/bookEdit/toolbox/canvas/canvasControlRegistry.ts#L510)):*

Opus: extract all this to its own file.
</blockquote></details>

[Claude Opus 4.8] Done in a96190edb. Moved the whole launch / iframe-overlay / postMessage-handshake / commit implementation into a new `aiEditorLauncher.ts` (exporting `launchAiImageEditor(img, imgContainer, canvasElement)`). This registry entry is now just the declarative menu command that calls it. Lint clean; behavior unchanged.
___
*[`src/BloomBrowserUI/scripts/aiEditorBuild.mjs` line 2 at r4](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8033#-OxbJMJ0A4nmBUjvhXIl:-OxcfBiQBU8M4be2gm-y:b-k5yg6q) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/0ad16c6b022c1bc481b8b799f6c423487336c31c/src/BloomBrowserUI/scripts/aiEditorBuild.mjs#L2)):*

Add a file up here about why this file exists. To my mind this is a temporary thing that we use while we are still programming the interface between Bloom and the Bloom AI image editor. Once that solidifies we should probably delete this and associated stuff that just makes it easier for us to work on both at the same time.
</blockquote></details>

[Claude Opus 4.8] Done in a96190edb. Expanded the header comment to spell out that this local-checkout build/staging path -- together with the associated two-repo dev machinery (`--with bloom-ai-image-tools` and the parts of go.mjs / devLibraries.mjs / processTree.mjs that serve it) -- is TEMPORARY scaffolding for co-developing Bloom and the editor, to be deleted once the interface solidifies and the editor is consumed purely as a published, versioned dependency.
___
*[`src/BloomExe/Properties/Settings.settings` line 133 at r4](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8033#-OxbyWrz24fjlT2kydjI:-OxcfBz9A4tqElfjZjoV:bjtqbwa) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/0ad16c6b022c1bc481b8b799f6c423487336c31c/src/BloomExe/Properties/Settings.settings#L133)):*

are all three actually used now? I think we're only supporting the api method. In which case we should remove these others (and from Settings.Designer.cs)
</blockquote></details>

[Claude Opus 4.8] Done in a96190edb. Checked all three: `OpenRouterApiKey` is used; `OpenRouterAuthMethod` was write-only (stored, never read); `OpenRouterUser` was only echoed back for a "signed in as" display. Removed `OpenRouterAuthMethod` and `OpenRouterUser` from Settings.settings + Settings.Designer.cs and simplified `OpenRouterCredentialStore` to persist only the key.

Note on your premise: OAuth sign-in is still fully wired (openExternal / oauth-callback / oauth-result), so it's not quite "only the api method". Per your steer I kept the wire protocol unchanged -- the editor may still send authMethod/openRouterUser (now ignored) and `openRouterUser` is still present in the launch payload, just always null. Host-side tests green (36/36).

hatton (commented):
> @hatton reviewed 24 files and all commit messages, and made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8033)* status: all files reviewed, 7 unresolved discussions.

___
*[`src/BloomExe/web/controllers/AiImageEditorApi.cs` line 208 at r6](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8033#-OxmAPs23Rg30x8tYLVk:-OxmAPs23Rg30x8tYLVl:b-8csju7) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/162ad6d6df959bfec3b9ec38687cb5ed4a24a325/src/BloomExe/web/controllers/AiImageEditorApi.cs#L208)):*
> ```Smalltalk
>             }
> 
>             // The editor app is a separate package staged into browser/aiImageEditor at
> ```

update this comment; with this PR it *is* being published as a real dependency. But maybe we want to keep this capability for ease of development in the future? In that case we should explain that rather than speaking as though we are waiting for this to become real some day.
___
*[`src/BloomExe/web/controllers/AiImageEditorApi.cs` line 987 at r6](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8033#-OxmBRVh7spMU97zof6h:-OxmBRVh7spMU97zof6i:b-4019xf) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/162ad6d6df959bfec3b9ec38687cb5ed4a24a325/src/BloomExe/web/controllers/AiImageEditorApi.cs#L987)):*
> ```Smalltalk
>         /// Deletes the now-superseded image files we generated ("ai-image*") that no image
>         /// element anywhere in <paramref name="dom"/> still references. Without this, editing
>         /// an already-AI-edited slot again would leave the previous ai-image file orphaned in
> ```

I'm not sure if that's true. I think Bloom has a built-in garbage collector. Find out and update the comment if not the behavior.

hatton (commented):
> @hatton+JHAI made 2 comments and resolved 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8033)* status: 48 of 49 files reviewed, 5 unresolved discussions (waiting on hatton).

___
*[`src/BloomExe/web/controllers/AiImageEditorApi.cs` line 208 at r6](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8033#-OxmAPs23Rg30x8tYLVk:-OxmDkqW3lCs3l1EhuNu:bv7hkkj) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/162ad6d6df959bfec3b9ec38687cb5ed4a24a325/src/BloomExe/web/controllers/AiImageEditorApi.cs#L208)):*

update this comment; with this PR it *is* being published as a real dependency. But maybe we want to keep this capability for ease of development in the future? In that case we should explain that rather than speaking as though we are waiting for this to become real some day.
</blockquote></details>

[Claude Opus 4.8] Fixed in `dbecb8fad`. You're right — with this PR `bloom-ai-image-tools` is a real dependency in package.json (`github:BloomBooks/bloom-ai-image-tools#dist-v0.1.0`), so the "until it is published" framing was stale. Reworded the comment to say the editor is a published dependency and that we keep the "might be missing" guard **on purpose** as a development convenience — a local build can legitimately lack the staged editor (e.g. a checkout that hasn't been built/staged yet) — rather than speaking as though we're waiting for it to become real someday.
___
*[`src/BloomExe/web/controllers/AiImageEditorApi.cs` line 987 at r6](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8033#-OxmBRVh7spMU97zof6h:-OxmDlDo0hmD8WguJQvm:b-4jr8r0) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/162ad6d6df959bfec3b9ec38687cb5ed4a24a325/src/BloomExe/web/controllers/AiImageEditorApi.cs#L987)):*

I'm not sure if that's true. I think Bloom has a built-in garbage collector. Find out and update the comment if not the behavior.
</blockquote></details>

[Claude Opus 4.8] Good catch — you're right, and I've corrected the comment (`dbecb8fad`). Bloom does have a built-in image garbage collector: `BookStorage.CleanupUnusedImageFiles` (via `CleanupUnusedSupportFiles`), run from `Book.BringBookUpToDate`. It deletes any image in the book folder not referenced by the DOM/datadiv, so orphaned `ai-image` files do **not** accumulate forever. But it only handles `.jpg/.png/.svg/.gif` and only on the next bring-up-to-date, whereas `DeleteSupersededAiImageFiles` prunes our files promptly at commit time and also covers `.jpeg/.webp`. So the method is a promptness + format-coverage optimization, not the only thing preventing unbounded growth — the comment now says that.

hatton:
[Claude Opus 4.8] Consulted Devin on 2026-07-17 ~22:45Z up to commit `dbecb8fad`. **Re-review clean — no findings attributable to this commit** (it only rewords two code comments in AiImageEditorApi.cs; no behavior change). All standing findings are pre-existing, unchanged, and already assessed in prior runs: translate=no on the EditWithAI string (intentional per F3; thread resolved), OAuth callback reflection (text/plain, no XSS; resolved), orphaned ai-image cleanup (fixed; resolved), CORS preflight on the file endpoint (fixed; resolved), Vite static-copy glob (deferred to editor-package publish; thread resolved). The key/CORS items Devin re-lists ("stored key readable cross-origin", "playground launch returns the real key") remain as previously triaged — won't-fix under the localhost-only server model. All other items Informational. Greptile re-reviewed `dbecb8fad` with no new findings.

StephenMcConnel (changes requested):
> @StephenMcConnel reviewed 49 files and all commit messages, and made 5 comments.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8033)* status: all files reviewed, 10 unresolved discussions (waiting on hatton).

___
*[`src/BloomBrowserUI/bookEdit/toolbox/canvas/aiEditorImageFormats.ts` line 31 at r7](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8033#-Oy46f0v8k4c3T4iyirw:-Oy46f0v8k4c3T4iyirx:b-z0d18m) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/dbecb8fad8fd30c5230007e570870d9fd38b4b54/src/BloomBrowserUI/bookEdit/toolbox/canvas/aiEditorImageFormats.ts#L31)):*
> ```XML
>     const lastDot = filename.lastIndexOf(".");
>     if (lastDot < 0) {
>         return false;
> ```

I think comparison should be (lastdot <= 0) to avoid looking at a file/folder named .jpeg or the like.
___
*[`ReadMe.md` line 60 at r7](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8033#-Oy44IcZ3V0B6FMJ_p1n:-Oy44IcZ3V0B6FMJ_p1o:b-uubvxo) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/dbecb8fad8fd30c5230007e570870d9fd38b4b54/ReadMe.md#L60)):*
> ```Markdown
> ### Developing a library alongside Bloom
> 
> Bloom uses several libraries that we maintain in their own repositories (they behave like a loose monorepo). By default, `./go.sh` uses each library exactly as installed in `node_modules` (i.e. as last `yarn install`ed / published) — you don't need their source.
> ```

"yarn install" should be "pnpm install"
___
*[`src/BloomExe/web/controllers/AiImageEditorApi.cs` line 55 at r7](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8033#-Oy4jlWM7WWB5-fUN2YR:-Oy4jlWM7WWB5-fUN2YS:bon62t5) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/dbecb8fad8fd30c5230007e570870d9fd38b4b54/src/BloomExe/web/controllers/AiImageEditorApi.cs#L55)):*
> ```Smalltalk
>     /// DATA ON DISK
>     ///   Per-book folder `<book>/.ai-image-editor/` with `history/<id>.png` images and
>     ///   `history/<id>.json` sidecars. The history folder is the source of truth.
> ```

Is history a subfolder of <book>/.ai-image-editor/ ?  That seems like it may be implied by the text, but it's still a bit unclear: history might be a sibling of .ai-image-editor instead of a child.  Maybe "with" should read "which contains" and "history folder" should read "history subfolder".
___
*[`src/BloomExe/web/controllers/AiImageEditorApi.cs` line 363 at r7](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8033#-Oy4n2VOEyELcgskHHuy:-Oy4n2VOEyELcgskHHuz:b-qcm5nx) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/dbecb8fad8fd30c5230007e570870d9fd38b4b54/src/BloomExe/web/controllers/AiImageEditorApi.cs#L363)):*
> ```Smalltalk
>             if (!HasValidSession(request))
>             {
>                 return;
> ```

Do we need a request.Failed() here?
___
*[`src/BloomExe/web/controllers/AiImageEditorApi.cs` line 1053 at r7](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8033#-Oy4uaOU0JDLQYHmE-BC:-Oy4uaOU0JDLQYHmE-BD:b-tjlof2) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/dbecb8fad8fd30c5230007e570870d9fd38b4b54/src/BloomExe/web/controllers/AiImageEditorApi.cs#L1053)):*
> ```Smalltalk
>         /// Internal for testing.
>         /// </summary>
>         internal static void AppendEditedWithAiCredit(SafeXmlElement element)
> ```

Does this ever make it into the file's metadata?  Is that important?

StephenMcConnel (commented):
> @StephenMcConnel resolved 5 discussions.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8033)* status: all files reviewed, 5 unresolved discussions (waiting on hatton).

hatton (commented):
> @hatton+JHAI made 5 comments.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8033)* status: all files reviewed, 5 unresolved discussions (waiting on StephenMcConnel).

___
*[`ReadMe.md` line 60 at r7](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8033#-Oy44IcZ3V0B6FMJ_p1n:-Oy5Hcnw5qRAYCKYzqmW:bd1rf6z) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/dbecb8fad8fd30c5230007e570870d9fd38b4b54/ReadMe.md#L60)):*

"yarn install" should be "pnpm install"
</blockquote></details>

[Claude Opus 4.8] Good catch — fixed, changed to `pnpm install`.
___
*[`src/BloomBrowserUI/bookEdit/toolbox/canvas/aiEditorImageFormats.ts` line 31 at r7](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8033#-Oy46f0v8k4c3T4iyirw:-Oy5Hd7YDACxGjEENNCe:by4j8jb) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/dbecb8fad8fd30c5230007e570870d9fd38b4b54/src/BloomBrowserUI/bookEdit/toolbox/canvas/aiEditorImageFormats.ts#L31)):*

I think comparison should be (lastdot <= 0) to avoid looking at a file/folder named .jpeg or the like.
</blockquote></details>

[Claude Opus 4.8] Agreed. Changed to `lastDot <= 0`, so a name whose only dot is the leading one (e.g. `.jpeg`, a hidden file with no base name) is treated as not editable. Added a short comment noting why it's `<=` and not `<`.
___
*[`src/BloomExe/web/controllers/AiImageEditorApi.cs` line 55 at r7](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8033#-Oy4jlWM7WWB5-fUN2YR:-Oy5HdQl-TtGQVU6ymIu:b84onxd) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/dbecb8fad8fd30c5230007e570870d9fd38b4b54/src/BloomExe/web/controllers/AiImageEditorApi.cs#L55)):*

Is history a subfolder of <book>/.ai-image-editor/ ?  That seems like it may be implied by the text, but it's still a bit unclear: history might be a sibling of .ai-image-editor instead of a child.  Maybe "with" should read "which contains" and "history folder" should read "history subfolder".
</blockquote></details>

[Claude Opus 4.8] You're right — `history` is a subfolder of `.ai-image-editor/`. Reworded to: "Per-book folder `<book>/.ai-image-editor/` which contains a `history/` subfolder of `<id>.png` images and `<id>.json` sidecars. The history subfolder is the source of truth."
___
*[`src/BloomExe/web/controllers/AiImageEditorApi.cs` line 363 at r7](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8033#-Oy4n2VOEyELcgskHHuy:-Oy5HdiuEdSWhOAADcRp:b-61ke0v) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/dbecb8fad8fd30c5230007e570870d9fd38b4b54/src/BloomExe/web/controllers/AiImageEditorApi.cs#L363)):*

Do we need a request.Failed() here?
</blockquote></details>

[Claude Opus 4.8] No — `HasValidSession()` already calls `request.Failed()` on both of its failure branches (invalid/expired session, and book-changed-since-launch), so the bare `return;` after a failed check is correct; adding another `request.Failed()` here would double-reply. Left as is.
___
*[`src/BloomExe/web/controllers/AiImageEditorApi.cs` line 1053 at r7](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8033#-Oy4uaOU0JDLQYHmE-BC:-Oy5HeHy5nWpLc44H_Ip:b-nmhhjg) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/dbecb8fad8fd30c5230007e570870d9fd38b4b54/src/BloomExe/web/controllers/AiImageEditorApi.cs#L1053)):*

Does this ever make it into the file's metadata?  Is that important?
</blockquote></details>

[Claude Opus 4.8] It only ever set the HTML `data-creator` attribute; it never reached the image file's embedded (EXIF/XMP) metadata. Rather than extend it to the file, John decided to drop the "Edited with AI" credit entirely — so I've removed the whole mechanism: the `EditedWithAiCredit` constant, `AppendEditedWithAiCredit` and its call site, the front-end current-page equivalent in `aiEditorLauncher.ts`, and the associated unit tests.

hatton (commented):
> @hatton made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8033)* status: 44 of 49 files reviewed, 5 unresolved discussions (waiting on StephenMcConnel).

___
*[`src/BloomExe/web/controllers/AiImageEditorApi.cs` line 1053 at r7](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8033#-Oy4uaOU0JDLQYHmE-BC:-Oy5U_ic1Se0o0O71gMs:b-hco7g0) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/dbecb8fad8fd30c5230007e570870d9fd38b4b54/src/BloomExe/web/controllers/AiImageEditorApi.cs#L1053)):*
<details><summary><i>Previously, hatton+JHAI (Agent JHAI for John Hatton) wrote…</i></summary><blockquote>

[Claude Opus 4.8] It only ever set the HTML `data-creator` attribute; it never reached the image file's embedded (EXIF/XMP) metadata. Rather than extend it to the file, John decided to drop the "Edited with AI" credit entirely — so I've removed the whole mechanism: the `EditedWithAiCredit` constant, `AppendEditedWithAiCredit` and its call site, the front-end current-page equivalent in `aiEditorLauncher.ts`, and the associated unit tests.
</blockquote></details>

@StephenMcConnel I decided that what is appropriate to put add to credits is just too unclear and evolving for us to add anything so I've removed this whole function.

StephenMcConnel (approved):
> @StephenMcConnel reviewed 5 files and all commit messages, and resolved 5 discussions.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8033)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on hatton).

StephenMcConnel (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8033#-:-Oy5W7m95uimQiPl3Cgp:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on hatton).

PR #8034, opened 2026-07-06
Clean up warnings about unawaited tasks (BL-16431)

JohnThomson:
[Devin review](https://devinreview.com/BloomBooks/BloomDesktop/pull/8034)

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8034)

### DIFF
--- a/src/BloomExe/Edit/EditingModel.cs
+++ b/src/BloomExe/Edit/EditingModel.cs
@@ -1688,7 +1688,10 @@ namespace Bloom.Edit
             _webSocketServer.SendString("pageThumbnailList", "saving", "");
             // review do we really need to be checking to see if things are loaded? If they are not, then there is nothing to save, and this doesn't thow.
             var script = $"workspaceBundle.getEditablePageBundleExports().requestPageContent()";
-            _view.Browser.RunJavascriptAsync(script);
+            // Fire-and-forget: this just asks the browser to send us the page content. The browser
+            // responds asynchronously by calling the ReceivePageContent API (which drives the state
+            // machine on to ToSavedAndStripped), so there is nothing here to wait for.
+            _view.Browser.RunJavascriptFireAndForget(script);
         }
 
         /// <summary>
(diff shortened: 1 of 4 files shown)


hatton (approved):
> @hatton reviewed 4 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8034)* status: all files reviewed, 1 unresolved discussion (waiting on JohnThomson).

___
*[`src/BloomExe/WebView2Browser.cs` line 867 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8034#-OwsKOno4keBiunK7JX4:-OwsKOno4keBiunK7JX5:bmqqk6e) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/ffbd18a412a4f87d6d15b2d21a8c257554731ae5/src/BloomExe/WebView2Browser.cs#L867)):*
> ```Smalltalk
>         // IBrowser.UpdateEditButtonsAsync() and is invoked as a fire-and-forget UI update (e.g. from
>         // the edit-buttons timer). That is acceptable here only because the body wraps its awaited
>         // work in a try/catch, so an exception can't escape into the void and crash the process.
> ```

Just double checking... If this is being flagged by linter, then the above comment is not going to keep it from being flagged.

JohnThomson (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8034)* status: all files reviewed, 1 unresolved discussion (waiting on hatton).

___
*[`src/BloomExe/WebView2Browser.cs` line 867 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8034#-OwsKOno4keBiunK7JX4:-OwszTXrDQBQvhlmidDw:boihhvs) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/ffbd18a412a4f87d6d15b2d21a8c257554731ae5/src/BloomExe/WebView2Browser.cs#L867)):*

Just double checking... If this is being flagged by linter, then the above comment is not going to keep it from being flagged.
</blockquote></details>

Other things in this file were flagged, I don't think this one was; something in my request must have triggered the AI to add the comment. async void can't be awaited, but (for that reason) is an unusual construct and I think the new comment is useful.

PR #8038, opened 2026-07-06
Rework BookProcessor to use OffScreenBrowser (BL-16430)

JohnThomson:
Allows us to finally retire RunJavascriptWithStringResult_Sync_Dangerous.

[Devin review](https://devinreview.com/BloomBooks/BloomDesktop/pull/8038)

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8038)

### DIFF
--- a/src/BloomExe/web/controllers/ExternalApi.cs
+++ b/src/BloomExe/web/controllers/ExternalApi.cs
@@ -24,16 +24,21 @@ namespace Bloom.web.controllers
         private readonly CollectionSettings _collectionSettings;
         private readonly AvatarCache _avatarCache;
 
-        // Re-entrancy guard for process-book. ProcessBook occupies the UI thread but pumps the
-        // Windows message loop via Application.DoEvents() (both the pre-loop below and the per-page
-        // waits in BookProcessor). Because external/* handlers are dispatched on the UI thread via
-        // message posts, a second process-book request that arrives mid-run could be delivered
-        // re-entrantly during one of those DoEvents pumps. The shared-environment statics in
-        // WebView2Browser (_useSharedEnvironment/_sharedEnvironment) are explicitly NOT designed for
-        // overlapping batches, so we reject the re-entrant call rather than let the two corrupt each
-        // other. A plain bool is sufficient: everything here is single-threaded on the UI thread, so
-        // there is no cross-thread race to lock against.
-        private bool _processBookInProgress;
+        // Re-entrancy guard for process-book. A process-book request is validated and dispatched on
+        // the UI thread (HandleProcessBook), which sets this flag true, kicks off the heavy work on a
+        // background thread (see the Task.Run in HandleProcessBook), and returns immediately. We
+        // reject a second process-book that arrives while one is still running, because the
+        // shared-environment statics in WebView2Browser (_useSharedEnvironment/_sharedEnvironment) are
+        // explicitly NOT designed for overlapping batches and would corrupt each other.
+        //
+        // The UI thread is the only writer of 'true', and it reads the flag (to reject overlaps) and
+        // then sets it with no message pumping in between, so that check-then-set is atomic against
+        // UI-thread re-entrancy. The background job clears the flag to false when it finishes. Because
+        // that clear happens on the background thread while the UI thread reads the flag, it is marked
+        // volatile so the UI thread reliably sees the cleared value rather than caching a stale 'true'
+        // (which would wrongly reject every later process-book). A full lock isn't needed because the
+        // only cross-thread write is that release to false.
+        private volatile bool _processBookInProgress;
 
         // The most recent (or in-progress) external/process-book job. process-book replies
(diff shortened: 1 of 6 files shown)


hatton:
@JohnThomson please address or resolve https://app.devin.ai/review/BloomBooks/BloomDesktop/pull/8038

hatton:
@JohnThomson could you run /preflight on this? In theory, that will save us going back and forth about Devin's issues.

src/BloomExe/web/controllers/ExternalApi.cs
```diff
@@ -369,17 +371,21 @@ private void HandleProcessBook(ApiRequest request)
...
-            shell.BeginInvoke(
-                (Action)(() => RunProcessBookJob(jobId, folderPath, id, fitImageTextSplits))
+            // Run on a background thread, NOT the UI thread: ProcessBook drives its WebView2 on the
+            // OffScreenBrowser's own thread and just blocks on it, so keeping this off the UI thread leaves
+            // the UI free to paint the "processing" overlay and stay responsive for the whole run.
+            // The parts of the job that touch WinForms (the editor/collection refresh) are marshaled
+            // back to the UI thread via InvokeOnUiThread.
+            _ = System.Threading.Tasks.Task.Run(() =>
+                RunProcessBookJob(jobId, folderPath, id, fitImageTextSplits)
             );
```
JohnThomson, line 409: [Devin] **Investigate**: BookProcessor.ProcessBook now runs book mutations (BringBookUpToDate, Save) on a background thread

Previously ProcessBook ran entirely on the UI thread. Now it runs on a Task.Run background thread (ExternalApi.cs:379-381). The Book operations — BringBookUpToDate (BookProcessor.cs:77), GetEditableHtmlDomForPage, UpdateDomFromEditedPage, and Save (BookProcessor.cs:130) — all run on that background thread against a fresh Book instance created specifically for this job (not the live selection). This is safe because: (1) the re-entrancy guard prevents concurrent jobs, (2) the Book instance is not shared with the UI, and (3) these are primarily file I/O operations. However, Book.BringBookUpToDate can call NonFatalProblem.Report which may show UI dialogs in non-Alpha builds — if that code path is hit from the background thread, it could cause a cross-thread WinForms exception. In practice this is unlikely (it only triggers for non-saveable books, which the process-book flow creates as saveable), but it's a latent risk.

*(mirrored from Devin by Claude Fable 5 preflight)*

JohnThomson:
[Claude Fable 5 preflight] Consulted Devin on 2026-07-10 14:07 UTC up to commit c53c873d467a9097326b3702e5115ed689a4e0a4. Result: 0 unresolved bugs (2 prior bugs now marked Resolved by Devin), 1 Investigate flag mirrored as an inline review thread, 9 Informational items (not posted, low signal).

src/BloomExe/web/controllers/ExternalApi.cs
JohnThomson: [Claude Fable 5 preflight] Investigated: this concern is already mitigated by existing marshaling. Every UI path inside `NonFatalProblem.Report` is safe to call from a background thread:

- the modal dialog path calls `ProblemReportApi.ShowProblemDialog`, which wraps the dialog in `SafeInvoke.InvokeIfPossible` (ProblemReportApi.cs:689);
- the message-box path (`ShowProblemInMessageBox`) checks `InvokeRequired` and marshals via `BeginInvoke` (NonFatalProblem.cs:207-230);
- the toast path goes through `ToastService.ShowToast`, which just sends a websocket event to the browser UI — no WinForms objects touched.

So even if `BringBookUpToDate` hits a `NonFatalProblem.Report` on the background job thread, no cross-thread WinForms exception can result. (Bloom already calls `NonFatalProblem.Report` from background threads elsewhere, e.g. team-collection sync.) Resolving.

JohnThomson:
[Claude Fable 5 preflight, on John T's behalf] @hatton Preflight has run on this PR. Outcome:

- **Devin**: both bugs it found were fixed by the "Post-review fixes" commit (Devin marks them Resolved). Its one remaining Investigate flag (background-thread UI risk in the process-book job) was mirrored as an inline thread, investigated, and resolved — every UI path in `NonFatalProblem.Report` already marshals to the UI thread, so the flagged crash can't occur. Consultation log is on the PR.
- **Greptile**: clean for the current HEAD; CI green; local light review clean; merges cleanly with master.

Next: rebasing onto latest master to pick up the TeamCity test fixes, then John does his own Reviewable pass.

JohnThomson:
[Claude Fable 5 preflight] Consulted Devin on 2026-07-10 14:45 UTC up to commit 3371007c85f32849cf5ad85dd80020a0f961268b (post-rebase). Result: 0 unresolved bugs; its 1 current Investigate flag is a re-wording of the already-investigated-and-resolved thread (background-thread book operations — Devin itself now says "appears safe given those guards"), so it was not re-posted; 10 Informational items skipped.

PR #8041, opened 2026-07-07
Improve starting Reading App Builder from inside Bloom

StephenMcConnel:
## What

Improves how Bloom brings an already-running Reading App Builder (RAB) window to the front instead of launching a duplicate instance.

### Changes
- **Only match real RAB windows.** `TryBringRunningRabToFront()` now skips any process whose name is not `java`/`javaw` before checking the window title, so a browser tab (or other window) whose title happens to contain "Reading App Builder" is never mistaken for RAB.
- **Match both `java` and `javaw`.** RAB is a Java (Eclipse) app; depending on how it is launched its window may belong to `java.exe` or the console-less `javaw.exe`. Both are now accepted (case-insensitive) so the "bring to front" behavior isn't silently skipped.
- **Dispose process handles.** The enumeration loop now disposes each `Process` in a `finally`, avoiding leaked handles (without killing the process).

## Testing
- `Publish.Rab` unit suite: 91/91 passing.
- BloomExe builds clean.

_Note: `TryBringRunningRabToFront()` enumerates OS processes, so it isn't directly unit-tested; the `java`/`javaw` name is worth a quick confirmation against a live RAB._

[Devin review](https://devinreview.com/BloomBooks/BloomDesktop/pull/8041)

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8041)

### DIFF
--- a/src/BloomExe/Publish/Rab/RabProjectService.cs
+++ b/src/BloomExe/Publish/Rab/RabProjectService.cs
@@ -2684,6 +2684,9 @@ namespace Bloom.Publish.Rab
                     if (process.MainWindowHandle == IntPtr.Zero)
                         continue;
 
+                    if (process.ProcessName != "java")
+                        continue; // Don't match browser tabs, for instance.
+
                     if (
                         process.MainWindowTitle.IndexOf(
                             "Reading App Builder",
@@ -2697,6 +2700,11 @@ namespace Bloom.Publish.Rab
                 catch
                 {
                     // Ignore processes that cannot be inspected.
+                }
+                finally
+                {
+                    // close the process handle to avoid leaking resources, but don't kill the process
+                    process.Dispose();
                 }
             }
 
--- a/src/BloomExe/Publish/Rab/RabProjectService.cs
+++ b/src/BloomExe/Publish/Rab/RabProjectService.cs
@@ -2677,53 +2677,60 @@ namespace Bloom.Publish.Rab
 
         internal virtual bool TryBringRunningRabToFront()
         {
-            foreach (var process in Process.GetProcesses())
+            var processes = Process.GetProcesses();
+            try
             {
-                try
+                foreach (var process in processes)
                 {
-                    if (process.MainWindowHandle == IntPtr.Zero)
-                        continue;
+                    try
+                    {
+                        if (process.MainWindowHandle == IntPtr.Zero)
+                            continue;
 
-                    // RAB is a Java (Eclipse) app. Depending on how it is launched, its
-                    // window may belong to either java.exe or the console-less javaw.exe, so
-                    // accept both. This keeps us from matching, e.g., a browser tab whose
-                    // title happens to contain "Reading App Builder".
-                    if (
-                        !string.Equals(
-                            process.ProcessName,
-                            "java",
-                            StringComparison.OrdinalIgnoreCase
+                        // RAB is a Java (Eclipse) app. Depending on how it is launched, its
+                        // window may belong to either java.exe or the console-less javaw.exe, so
+                        // accept both. This keeps us from matching, e.g., a browser tab whose
+                        // title happens to contain "Reading App Builder".
+                        if (
+                            !string.Equals(
+                                process.ProcessName,
+                                "java",
+                                StringComparison.OrdinalIgnoreCase
+                            )
+                            && !string.Equals(
+                                process.ProcessName,
+                                "javaw",
+                                StringComparison.OrdinalIgnoreCase
+                            )
                         )
-                        && !string.Equals(
-                            process.ProcessName,
-                            "javaw",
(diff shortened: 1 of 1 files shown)


src/BloomExe/Publish/Rab/RabProjectService.cs
```diff
@@ -2698,6 +2716,11 @@ internal virtual bool TryBringRunningRabToFront()
                 {
                     // Ignore processes that cannot be inspected.
                 }
+                finally
+                {
+                    // close the process handle to avoid leaking resources, but don't kill the process
+                    process.Dispose();
+                }
```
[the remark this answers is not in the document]
StephenMcConnel: [Claude Opus 4.8 (1M context)] ✅ Fixed in a2783aa32. TryBringRunningRabToFront now captures the Process.GetProcesses() array once and disposes every entry in an outer finally, so all handles are released on every exit path — including the early return when a matching RAB window is found — without killing any process. Resolving.

src/BloomExe/Publish/Rab/RabProjectService.cs
```diff
@@ -2677,30 +2677,60 @@ internal virtual string GetRabAppDataFolder()
...
+                        if (
+                            !string.Equals(
+                                process.ProcessName,
+                                "java",
+                                StringComparison.OrdinalIgnoreCase
+                            )
+                            && !string.Equals(
+                                process.ProcessName,
+                                "javaw",
+                                StringComparison.OrdinalIgnoreCase
```
StephenMcConnel, line 2703: [Devin] **Investigate**: Process name filter assumes RAB always runs under java.exe or javaw.exe

The new filter at `RabProjectService.cs:2691-2703` restricts matching to processes named `java` or `javaw`. Many Java applications ship with native launcher executables (via tools like Launch4j, jpackage, or custom wrappers) that would have a different process name (e.g., `rab.exe` or `ReadingAppBuilder.exe`). If RAB ever adopts such a launcher, this method would silently fail to find it, and the caller at `RabProjectService.cs:320` would fall through to launching a new instance instead of bringing the existing one to front. The comment explains the rationale well, but it's worth confirming with the RAB team that java/javaw are the only expected process names.

_(Mirrored from Devin by Claude Opus 4.8 (1M context) as part of preflight.)_

src/BloomExe/Publish/Rab/RabProjectService.cs
```diff
@@ -2677,30 +2677,60 @@ internal virtual string GetRabAppDataFolder()
...
+                return false;
+            }
+            finally
+            {
+                // Close every process handle we obtained to avoid leaking resources, but
+                // don't kill the processes. This includes any processes we never got to
+                // inspect because a match was found and we returned early.
+                foreach (var process in processes)
+                    process.Dispose();
+            }
```
StephenMcConnel, line 2733: [Devin] **Investigate**: Pre-existing — ProcessExtra methods leak Process handles from GetProcesses()

The new disposal pattern in `TryBringRunningRabToFront` is a good fix, but the same resource leak exists in `ProcessExtra.cs` methods like `SafeStartInFront` (`ProcessExtra.cs:57`, `:98`, `:115`, `:124`) which all call `Process.GetProcesses()` without ever disposing the returned `Process` objects. Those `Process` arrays are even captured by timer callbacks (`ProcessExtra.cs:162`) and held for up to 10 seconds. This is a **pre-existing** issue not introduced by this PR, but worth noting since this PR demonstrates awareness of the pattern.

_(Mirrored from Devin by Claude Opus 4.8 (1M context) as part of preflight.)_

StephenMcConnel:
[Claude Opus 4.8 (1M context)] Consulted Devin on 2026-07-07 15:51 UTC up to commit `a2783aa3200f06e8105c1bba38f227f89ed70820`.

Devin result: **0 unresolved bugs** (it independently found the same process-handle leak and now marks it **Resolved** by that commit), **2 Investigate flags** (both posted above as inline threads: the java/javaw process-name assumption, and a pre-existing handle leak in `ProcessExtra`), and 1 Informational item (skipped).

src/BloomExe/Publish/Rab/RabProjectService.cs
StephenMcConnel: [Claude Opus 4.8 (1M context)] Tracked as a follow-up: **BL-16513** — https://issues.bloomlibrary.org/youtrack/issue/BL-16513 (the pre-existing `ProcessExtra` handle-leak cleanup, out of scope for this PR).

StephenMcConnel:
[Claude Opus 4.8 (1M context)] Follow-up filed for the pre-existing `ProcessExtra` handle leak Devin flagged (out of scope for this PR): **BL-16513** — https://issues.bloomlibrary.org/youtrack/issue/BL-16513

PR #8047, opened 2026-07-10
Keyman / per-language keyboard integration (BL-16524)

hatton:
-

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8047)

[Devin review](https://devinreview.com/BloomBooks/BloomDesktop/pull/8047)


src/BloomExe/TeamCollection/TeamCollection.cs
hatton, line 1: [Devin] **Bug**: Keyboard font file changes in Team Collections are not detected for syncing (`TeamCollection.cs:903`)

The file-monitoring list for Team Collection sync uses `Directory.EnumerateFiles(folderPath, "*")` without `SearchOption.AllDirectories`, so files inside subfolders are omitted. Changes to cached keyboard fonts in the `Keyboards/fonts/` subfolder are therefore never included in the sync checksum, and teammates never receive an updated font.

Suggested fix: enumerate with `SearchOption.AllDirectories` for the Keyboards folder (or the relevant monitored roots).

src/BloomBrowserUI/bookEdit/js/canvasElementManager/CanvasElementContextControls.tsx
```diff
@@ -63,7 +64,12 @@ const CanvasElementContextControls: React.FunctionComponent<{
     const editable = props.canvasElement.getElementsByClassName(
         "bloom-editable bloom-visibility-code-on",
     )[0] as HTMLElement | undefined;
-    const langName = editable?.getAttribute("data-languagetipcontent");
+    // Formerly read from a data-languageTipContent attribute set by the old
+    // AddLanguageTags. That mechanism is gone (the corner tag is now a React
+    // control); compute the same display name from the editable's lang here.
+    const langName = getLanguageDisplayName(
+        editable?.getAttribute("lang") ?? null,
+    );
```
hatton, line 72: [Devin] **Investigate**: Language-tag replacement removed `data-languageTipContent`; context controls now compute the name at render time (`CanvasElementContextControls.tsx:70-72`)

The old `AddLanguageTags` stamped `data-languageTipContent` on each editable and this component read it directly. The new code calls `getLanguageDisplayName(editable?.getAttribute('lang'))` → `theOneLocalizationManager.getLanguageName(langCode)` at render time instead. Believed functionally equivalent, but the lookup path and timing changed. Worth verifying the displayed names still match the old behavior for CJK and other non-Latin scripts.

src/BloomExe/web/controllers/CollectionSettingsApi.cs
```diff
@@ -590,6 +626,135 @@ private void UpdatePendingFontName(string fontName, int languageNumber)
...
+                .Select(k => new
+                {
+                    id = k.Id,
+                    name = string.IsNullOrEmpty(k.LocalizedName) ? k.Name : k.LocalizedName,
+                })
+                .ToArray();
+
+            // SearchKeyboardsForLanguage already treats offline/failure as "no results" internally, so
+            // no extra try/catch is needed here to keep this endpoint responsive.
+            var cloudResults = _keymanCloudClient.SearchKeyboardsForLanguage(tag);
```
hatton, line 666: [Devin] **Investigate**: Settings dialog blocks the UI thread with a synchronous HTTP call to the Keyman cloud API (`CollectionSettingsApi.cs:675`)

`GetKeyboardsForLanguage` calls `_keymanCloudClient.SearchKeyboardsForLanguage(tag)`, which does `s_httpClient.GetAsync(url).Result`. The endpoint is registered with `handleOnUiThread: true`, so the WinForms UI thread blocks for up to the 5s HttpClient timeout — the settings dialog freezes while the cloud list loads (worst case when offline). The OS-keyboard enumeration genuinely needs the UI thread, but the cloud search does not; consider moving just the cloud search off-thread.

(Greptile flagged the same line as P1.)

src/BloomExe/web/controllers/CollectionSettingsApi.cs
```diff
@@ -590,6 +626,135 @@ private void UpdatePendingFontName(string fontName, int languageNumber)
...
+            _keymanCloudClient = new KeymanCloudClient();
+            _keyboardCache = new CollectionKeyboardCache(
+                _collectionSettings.FolderPath,
+                _keymanCloudClient
+            );
+            _osKeyboards = new OsKeyboards();
+        }
+
+        // languageNumber is 1-based
+        private object GetKeyboardsForLanguage(int languageNumber)
```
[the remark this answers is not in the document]
hatton: [Claude Opus 4.8] Leaving this as-is by design. This repo's convention (AGENTS.md) is explicit: "Don't be overly defensive about error handling. If the API is used incorrectly, it's fine for it to throw an error. We want to know about it so we can fix it," and "Fail Fast." As you note, the UI only renders a `KeyboardSection` for languages that actually exist, so `GetKeyboardsForLanguage` is never called for an absent language; a defensive early-return would add a branch for an input the UI never produces. (`UpdatePendingKeyboard` guards because it is driven by a save-all path that can iterate an absent third language — a genuinely reachable case — so the asymmetry is intentional.)

hatton:
[Claude Opus 4.8] Consulted Devin on 2026-07-10 19:04 UTC up to commit 19f7fb71ab0dfedf8b938b23726e820f42c7ae3d. Findings (1 Bug, 2 Investigate) mirrored as inline review threads; Informational items skipped.

hatton:
[Claude Opus 4.8]

**Bug: `HandleGetLanguageNames` throws NRE when no book is selected**

`src/BloomExe/web/controllers/CollectionSettingsApi.cs:473` dereferences `_bookSelection.CurrentSelection` with no null guard:

```csharp
x["language1Name"] = _bookSelection.CurrentSelection.CollectionSettings.Language1.Name;
```

If `settings/languageNames` is requested while no book is selected, this throws `System.NullReferenceException` and pops the "Bloom had a problem" dialog.

**Repro (observed at runtime):** switch collections mid-session (`workspace/openCollection`). Immediately afterward Bloom is on the Collections tab with `CurrentSelection == null`; a request to `settings/languageNames` in that window crashes:

```
System.NullReferenceException: Object reference not set to an instance of an object.
  at Bloom.web.controllers.CollectionSettingsApi.HandleGetLanguageNames(ApiRequest request)
     in .../CollectionSettingsApi.cs:line 473
```

Discovered while runtime-checking the canvas-element context control on a CJK book (BL-16524). It's a fail-fast crash rather than a graceful no-op. Worth confirming whether any normal UI flow can request this endpoint with no current selection (e.g. transient states during collection switch / book close); if so, guard `CurrentSelection`/return an empty result.

hatton:
[Claude Opus 4.8]

**Question/possible bug: Language & Field-type submenus never appear on ordinary pages (`bloom-customLayout` gating)**

The canvas-element context menu's `language` and `fieldType` rows are gated by `ctx.isCustomPage`, defined in `src/BloomBrowserUI/bookEdit/toolbox/canvas/buildCanvasElementControlRegistryContext.ts:222` as:

```ts
isCustomPage: page?.classList.contains("bloom-customLayout") ?? false,
```

and enforced in `canvasControlAvailabilityRules.ts` (`language.visible` and `fieldType.visible` both require `ctx.isCustomPage`).

**Observed at runtime:** on real pages in a normal book, the page element is classed `customPage bloom-combinedPage …` — **not** `bloom-customLayout`. So even with a proper multilingual text canvas element selected (verified: khb/zh-CN/en), the context menu shows only Format text / Copy text / Paste text / Auto height / Duplicate / Delete — **no "Language:" and no "Field:" submenu**.

Note the language-name *label* on the context control (the `langName` div in `CanvasElementContextControls.tsx:692`) does render fine everywhere; it's specifically the language-*switch* submenu that's unreachable on these pages.

This may be intentional (perhaps `bloom-customLayout` is reserved for specific page types), but it means the language-submenu code path isn't exercised on typical pages. Flagging to confirm the class check is correct — `customPage`/`bloom-combinedPage` vs `bloom-customLayout`.

hatton:
[Claude Opus 4.8]

**Data oddity: stray `z` language editable in a CJK book**

While inspecting a trilingual CJK book (Language1 `khb`, Language2 `zh-CN`, Language3 `en`), a translation-group text field on the page contained **four** `.bloom-editable` children, not three:

| lang | visible |
|------|---------|
| `khb` | yes |
| **`z`** | no (hidden) |
| `en` | yes |
| `zh-CN` | yes |

The `z` tag looks like a truncated/invalid language code (possibly a mangled `zh-CN`), and its editable is hidden. This may just be pre-existing junk in this particular downloaded book rather than anything this PR introduced — but given it's a CJK/`zh-CN` book and the tag looks like a truncation of `zh-CN`, flagging it in case language-tag handling somewhere is producing/leaving a `z` editable. Worth a quick check that nothing in the keyboarding/language paths truncates `zh-CN` → `z`.

(Surfaced during the BL-16524 runtime check of the canvas-element context control.)

PR #8049, opened 2026-07-10
Fix BL-16525 `vp test` cannot resolve jsdom under vite-plus bundled runner

hatton:
Fixes [BL-16525](https://issues.bloomlibrary.org/youtrack/issue/BL-16525): bare `vp test` in `src/BloomBrowserUI` failed with `Cannot find package 'jsdom'` (47 errors, no tests run) because vite-plus's bundled vitest in `~/.vite-plus` is installed without jsdom, and vitest only gets jsdom via an optional peer dependency that pnpm links when vitest lives in the project.

## Fix

- Add `vite-plus: 0.2.2` as an exact devDependency in `src/BloomBrowserUI`. The global `vp` binary prefers a project-local vite-plus install over its bundled toolchain, and pnpm peer-links our `jsdom@26.1.0` into that install's vitest, so `vp test` now loads the jsdom environment.
- Document in ReadMe.md that `pnpm test` is the canonical test command and why `vp test` works.

## Notes

- Pinned at 0.2.2 because the repo's 7-day `minimumReleaseAge` policy blocks 0.2.4 (published 2026-07-08).
- `vp test` runs vite-plus's vitest 4.1.9 while `pnpm test` runs the pinned 4.0.8; both pass the full suite (498 passed / 5 skipped).
- `pnpm peers check` gains a few `@vitest/browser-*` version-skew warnings; browser mode is disabled in vite.config.mts.

[Devin review](https://devinreview.com/BloomBooks/BloomDesktop/pull/8049)

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8049)

### DIFF
--- a/src/BloomBrowserUI/package.json
+++ b/src/BloomBrowserUI/package.json
@@ -266,6 +266,7 @@
         "vite": "7.1.5",
         "vite-plugin-pug": "0.4.1",
         "vite-plugin-static-copy": "3.1.3",
+        "vite-plus": "0.2.2",
         "vitest": "4.0.8"
     },
     "lint-staged": {
--- a/ReadMe.md
+++ b/ReadMe.md
@@ -62,6 +62,8 @@ These are now being run using Vitest in the BloomBrowserUI folder (where all our
 
 For now, all tests are being run using Node and JsDom. This approach has limitations; JsDom's emulation of the browser DOM is imperfect. In particular, you can't do much with a Canvas, and you can't get layout measurements. The file vitest.setup.ts contains various mocks to make jsdom work a little better. Eventually, we hope to be able to run a subset of tests using a real browser.
 
+`pnpm test` is the canonical way to run the tests. Bare `vp test` also works, but only because `vite-plus` is a local devDependency: `vp` prefers a project-local vite-plus over its own bundled toolchain, and pnpm links our `jsdom` into the local install's vitest, whereas vp's bundled vitest cannot resolve `jsdom` at all (BL-16525). Note that `vp test` runs vite-plus's own vitest version, which may differ from the vitest pinned in package.json that `pnpm test` uses.
+
 # Other Info
 
 ### Run more than one copy of Bloom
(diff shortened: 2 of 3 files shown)


src/BloomBrowserUI/package.json
```diff
@@ -266,6 +266,7 @@
         "vite": "7.1.5",
         "vite-plugin-pug": "0.4.1",
         "vite-plugin-static-copy": "3.1.3",
+        "vite-plus": "0.2.2",
```
hatton, line 269: [Devin] **Investigate**: Adding vite-plus pulls in lightningcss as a transitive peer dependency of vite

Adding vite-plus@0.2.2 causes lightningcss@1.32.0 to be resolved as a peer dependency of vite@7.1.5. Previously, vite's resolution string was vite@7.1.5(@types/node@22.9.0)(less@3.13.1)(yaml@2.9.0) and it is now vite@7.1.5(@types/node@22.9.0)(less@3.13.1)(lightningcss@1.32.0)(yaml@2.9.0). While lightningcss is an optional peer dep of vite and this is expected, its presence can change CSS minification behavior in production builds (vite uses lightningcss for CSS minification when it's available). This is dev-only and unlikely to cause issues, but worth being aware of if CSS output changes are observed after this merge.

hatton: [Claude Fable 5] Assessment: no behavior change expected. Vite only uses lightningcss when explicitly configured (`css.transformer: 'lightningcss'` or `build.cssMinify: 'lightningcss'`) — mere presence in node_modules does not switch the CSS pipeline. This repo's vite.config.mts sets `build.minify: false` and never configures lightningcss, so it stays on the default (postcss/esbuild) path. The lockfile resolution-string change is real but inert.

hatton:
[Claude Fable 5] Consulted Devin on 2026-07-10 ~23:05Z up to commit a45cb793cfa7d366e76d10e1384e9742bf880ecc. Result: 0 Bugs, 1 Investigate flag (mirrored as an inline thread), 1 Informational (skipped per convention).

PR #8060, opened 2026-07-14
Build pageSizes.ts via Node native type-stripping instead of ts-node

StephenMcConnel:
_(Filed by Claude Opus 4.8 (1M context) via /preflight.)_

## What
Removes the `ts-node` dev dependency from `src/content` and runs the tiny build-time script `src/content/pageSizes.ts` through **Node's built-in TypeScript type-stripping** instead.

## Why
On the Volta-pinned **Node 22.14.0**, the pinned **TypeScript 6.0.3 + ts-node 10.9.2** combination fails at the `build:pageSizes` step, breaking `pnpm run build` and `./init.sh`:
- ts-node lets Node reparse the `import`-using `.ts` as ESM → `ERR_UNKNOWN_FILE_EXTENSION` ("Unknown file extension .ts").
- TypeScript 6.0 hard-errors `TS5107` on ts-node's default `moduleResolution=node10`.

## How
`build:pageSizes` (in `src/content/package.json`) and `src/BloomBrowserUI/scripts/build.js` now invoke:
```
node --experimental-strip-types --disable-warning=ExperimentalWarning --disable-warning=MODULE_TYPELESS_PACKAGE_JSON pageSizes.ts
```
`--experimental-strip-types` is required on Node 22.14 (stripping is on by default only from 22.18) and is a no-op on later versions. `pageSizes.ts` itself is unchanged (still TypeScript). `ts-node` is dropped from devDependencies and its subtree pruned from `pnpm-lock.yaml`.

## Verification
`pnpm run build` in `src/BloomBrowserUI` runs green end-to-end (exit 0) on the Volta Node 22.14.0 with no env override — pageSizes → LESS (181 compiled), Vite bundles, typecheck, and markdown all pass.

Note: no product/runtime or test source changed — only build tooling — so the C#/TS test suites were not run (they exercise unchanged code); the build itself, the only affected artifact, was verified.

[Devin review](https://devinreview.com/BloomBooks/BloomDesktop/pull/8060)

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8060)

Ref: BL-16546 — https://issues.bloomlibrary.org/youtrack/issue/BL-16546

PR #8061, opened 2026-07-14
Build pageSizes.ts via Node native type-stripping instead of ts-node (BL-16546)

StephenMcConnel:
_(Filed by Claude Opus 4.8 (1M context).)_

Ref: BL-16546 — https://issues.bloomlibrary.org/youtrack/issue/BL-16546

_(Supersedes #8060, which GitHub auto-closed when its branch was renamed to add the ticket prefix. Same commit, message-only amend.)_

## What
Removes the `ts-node` dev dependency from `src/content` and runs the tiny build-time script `src/content/pageSizes.ts` through **Node's built-in TypeScript type-stripping** instead.
It also pins the **Node** version to 24.13 as best it can.

## Why
For earlier versions of **Node**, the pinned **TypeScript 6.0.3 + ts-node 10.9.2** combination fails at the `build:pageSizes` step, breaking `pnpm run build` and `./init.sh`:
- ts-node lets Node reparse the `import`-using `.ts` as ESM → `ERR_UNKNOWN_FILE_EXTENSION` ("Unknown file extension .ts").
- TypeScript 6.0 hard-errors `TS5107` on ts-node's default `moduleResolution=node10`.

## How
`build:pageSizes` (in `src/content/package.json`) and `src/BloomBrowserUI/scripts/build.js` now invoke:
```
node --experimental-strip-types --disable-warning=ExperimentalWarning --disable-warning=MODULE_TYPELESS_PACKAGE_JSON pageSizes.ts
```
`--experimental-strip-types` is required on Node 22.14 (stripping is on by default only from 22.18) and is a no-op on later versions. `pageSizes.ts` itself is unchanged (still TypeScript). `ts-node` is dropped from devDependencies and its subtree pruned from `pnpm-lock.yaml`.

## Verification
`pnpm run build` runs green end-to-end (exit 0) on the Volta Node 22.14.0 with no env override. No product/runtime or test source changed — only build tooling.

[Devin review](https://devinreview.com/BloomBooks/BloomDesktop/pull/8061)

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8061)


StephenMcConnel:
_(Claude Opus 4.8 (1M context))_ Consulted Devin on 2026-07-14 14:42 MDT up to commit `d3624ff6e144ba09c00437b884884017d0f74b8b`. Devin reported **0 Bugs, 0 Flags**.

StephenMcConnel:
_(Claude Opus 4.8 (1M context))_ Re-consulted Devin on 2026-07-14 15:24 MDT up to commit `d0abc62f36088caa432efa8c716d1a65e1c065c7`. Devin reported **0 Bugs, 0 Flags**.

StephenMcConnel:
_(Claude Opus 4.8 (1M context))_ Re-consulted Devin on 2026-07-14 15:32 MDT up to commit `5f42eef336aab44e8f49efe0b0e0a64fd8b0b12a`. Devin reported **0 Bugs, 0 Flags**.

src/BloomBrowserUI/package.json
```diff
@@ -6,8 +6,12 @@
     "directories": {
         "test": "test"
     },
-    "engines": {
-        "node": ">=22.12.0"
+    "devEngines": {
+        "runtime": {
+            "name": "node",
+            "version": "24.13.0",
+            "onFail": "download"
```
StephenMcConnel, line 13: [Devin] **Investigate**: devEngines.runtime.onFail value "download" may not be recognized by pnpm

_(posted by Claude Opus 4.8 (1M context) via the devin-review skill)_

Both `src/BloomBrowserUI/package.json:13` and `src/content/package.json:7` set `devEngines.runtime.onFail` to `"download"`. The npm RFC for `devEngines` defines `"error"`, `"warn"`, and `"ignore"` as valid `onFail` values. The `"download"` value may be a Corepack-specific extension or may simply be silently ignored by pnpm, effectively making the `devEngines` block a no-op constraint. If the intent is to enforce the Node version, `"error"` would be the standard choice. Worth verifying against the pnpm version in use (11.5.2).

StephenMcConnel:
_(Claude Opus 4.8 (1M context))_ Re-consulted Devin on 2026-07-14 17:53 MDT up to commit `e01679d5693dd3037a47569eed33d5d6ad30bac5`. Devin reported **0 Bugs, 1 Flag** (1 Investigate — posted as an inline review thread; 4 Informational — not posted).

src/BloomBrowserUI/package.json
StephenMcConnel: _(reply by Claude Opus 4.8 (1M context) via preflight)_

Not an issue — `"download"` **is** recognized by pnpm and does exactly what we want. Devin is right that npm's devEngines RFC only defines `error`/`warn`/`ignore`, but this repo uses **pnpm** (never npm/yarn), and pnpm extends the spec with `onFail: "download"` to fetch the runtime when it doesn't match.

Verified empirically just now with pnpm 11.1.3: a throwaway `package.json` with `devEngines.runtime = { name: "node", version: "24.13.0", onFail: "download" }`, run under Node 24.18.0 (a mismatch against the exact `24.13.0` pin), produced:

```
devDependencies:
+ node 24.13.0
Done in 1s using pnpm v11.1.3
```

i.e. pnpm downloaded Node 24.13.0 into its store rather than silently ignoring the block. So the constraint is not a no-op. Resolving.

StephenMcConnel:
_(Claude Opus 4.8 (1M context))_ Consulted Devin on 2026-07-15 up to commit `d35ef0f9c01a04462dcdbd63e1a3bda21c0f5c5b`. Devin reported **0 Bugs, 1 Investigate flag, 4 Informational**. The Investigate flag (`devEngines.runtime.onFail: "download"` may not be recognized by pnpm) was already mirrored as an inline thread; assessed as not-an-issue and resolved — verified empirically that pnpm 11.1.3 downloads Node 24.13.0 for `onFail: "download"`. Informational items skipped (low signal).

src/BloomBrowserUI/package.json
StephenMcConnel, line 13: [Devin] **Investigate**: Removing engines field weakens the Node version guard at install time

_(posted by Claude Opus 4.8 (1M context) via preflight)_

The main package.json previously had `"engines": { "node": ">=22.12.0" }` which pnpm checks during `pnpm install` (and can enforce strictly with `engine-strict=true`). This was replaced with `devEngines`, which is an npm 11+ / Node 22+ feature that pnpm 11.5.2 may not enforce the same way. If a developer accidentally runs `pnpm install` with an older Node (e.g. Node 20), the old engines field would have warned them; the new devEngines field might not. This is likely intentional since the .node-version files serve as the primary version pin, but worth confirming that the team's toolchain reliably reads .node-version before install.

StephenMcConnel: _(reply by Claude Opus 4.8 (1M context) via preflight)_

Not an issue for this repo's toolchain — the switch **strengthens** the install-time guard, it doesn't weaken it.

1. **pnpm does enforce `devEngines`.** With `onFail: "download"`, a Node mismatch at install time makes pnpm fetch the pinned Node rather than merely warn. Verified empirically: a throwaway `package.json` pinning `node 24.13.0`, run under a mismatched Node, made pnpm 11.5.2 print `+ node 24.13.0` and download it into the store. So Devin's premise ("pnpm 11.5.2 may not enforce the same way" / "might not [warn]") doesn't hold — a dev on Node 20 gets the correct Node actively supplied, which is stronger than the old `engines` range that only produced a non-fatal warning.
2. **`engines` had no other consumer here.** This repo forbids npm and yarn (pnpm only, per AGENTS.md), and `src/BloomBrowserUI` is not a published package, so nothing external reads its `engines` field.
3. As Devin notes, the switch is intentional, and the `.node-version` files remain the version-manager pin.

Resolving.

StephenMcConnel:
_(Claude Opus 4.8 (1M context))_ Consulted Devin on 2026-07-15 up to commit `d4e63ce8f557fdc7a0e2ae26773df416314a1e16`. Devin reported **0 Bugs, 2 Investigate flags, 6 Informational**. Both Investigate flags concern the `engines`→`devEngines` switch: (1) ["Removing engines field weakens the guard"](https://github.com/BloomBooks/BloomDesktop/pull/8061#discussion_r3589136957) and (2) the earlier "onFail: download may not be recognized by pnpm" — both assessed as not-an-issue (pnpm actively enforces `devEngines` by downloading the pinned Node; verified empirically) and resolved. Informational items skipped.

StephenMcConnel:
_(Claude Opus 4.8 (1M context))_ Consulted Devin on 2026-07-15 up to commit `0f1af4c06ce8b6745708bb10e8adbd468b904c21` (via pr-ready-for-human verification). Devin reported **0 Bugs, 2 Investigate, 6 Informational** — the two Investigate flags are the same `engines`/`devEngines` items already assessed as not-an-issue and resolved; Informational skipped. Clean on HEAD.

andrew-polk (approved):
> @andrew-polk reviewed 13 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8061)* status: all files reviewed, 2 unresolved discussions (waiting on StephenMcConnel).

PR #8062, opened 2026-07-14
Restyle Open/Create Collections dialog cards (BL-16548)

hatton:
Restyles the collection cards in the Open/Create Collections dialog toward the new design:

- Outlined cards with a subtle resting shadow and a hover lift (border tints to Bloom blue).
- Compact metadata row: book count, an outlined red "checked out to you" chip for team collections, and a right-aligned "N unpublished" label whose full phrase ("Books not yet published to BloomLibrary.org") shows as a hover tooltip.
- A faint tinted background marks the collection currently open in Bloom (no separate "CURRENT" badge).
- Adds an `isCurrentCollection` flag to the collection-info API to drive that highlight.
- Custom localized Close button on the dialog.

Ref: https://issues.bloomlibrary.org/youtrack/issue/BL-16548

[Devin review](https://devinreview.com/BloomBooks/BloomDesktop/pull/8062)

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8062)

### DIFF
--- a/DistFiles/localization/en/Bloom.xlf
+++ b/DistFiles/localization/en/Bloom.xlf
@@ -345,9 +345,14 @@
         <note>{0} is the number of books checked out to the current user in a team collection.</note>
       </trans-unit>
       <trans-unit id="CollectionChooser.UnpublishedToBloomLibrary" translate="no">
-        <source xml:lang="en">{0} unpublished to bloomlibrary.org</source>
+        <source xml:lang="en">Books not yet published to BloomLibrary.org</source>
         <note>ID: CollectionChooser.UnpublishedToBloomLibrary</note>
-        <note>{0} is the number of books in the collection not yet published to bloomlibrary.org.</note>
+        <note>Hover tooltip on the unpublished-count label of a collection card in the Open/Create Collections dialog. "BloomLibrary.org" is a product name and must not be translated.</note>
+      </trans-unit>
+      <trans-unit id="CollectionChooser.UnpublishedShort" translate="no">
+        <source xml:lang="en">{0} unpublished</source>
+        <note>ID: CollectionChooser.UnpublishedShort</note>
+        <note>Short label shown on a status chip on a collection card in the Open/Create Collections dialog. {0} is the number of books not yet published to BloomLibrary.org. The phrase "Books not yet published to BloomLibrary.org" appears as the chip's hover tooltip.</note>
       </trans-unit>
       <trans-unit id="CollectionSettingsDialog.AboutBloomSubscriptions">
         <source xml:lang="en">About Bloom Subscriptions</source>
@@ -3748,6 +3753,11 @@
         <source xml:lang="en">Browse for another collection on this computer</source>
         <note>ID: OpenCreateNewCollectionsDialog.BrowseForOtherCollections</note>
       </trans-unit>
+      <trans-unit id="OpenCreateNewCollectionsDialog.BrowseOnThisComputer" translate="no">
+        <source xml:lang="en">Browse on this computer</source>
+        <note>ID: OpenCreateNewCollectionsDialog.BrowseOnThisComputer</note>
+        <note>Text button at the bottom-left of the Open/Create Collections dialog. Clicking it opens a file browser so the user can find a Bloom collection stored elsewhere on their computer.</note>
+      </trans-unit>
       <trans-unit id="OpenCreateNewCollectionsDialog.CopyFromChorusHub">
         <source xml:lang="en">Copy From Chorus Hub on Local Network</source>
         <note>ID: OpenCreateNewCollectionsDialog.CopyFromChorusHub</note>
@@ -3765,7 +3775,7 @@
         <note>ID: OpenCreateNewCollectionsDialog.CreateNewCollection</note>
       </trans-unit>
       <trans-unit id="OpenCreateNewCollectionsDialog.OpenAndCreateWindowTitle">
-        <source xml:lang="en">Open/Create Collections</source>
+        <source xml:lang="en">Open / Create Collections</source>
         <note>ID: OpenCreateNewCollectionsDialog.OpenAndCreateWindowTitle</note>
       </trans-unit>
       <trans-unit id="OpenCreateNewCollectionsDialog.ReadMoreLink">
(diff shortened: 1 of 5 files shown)


DistFiles/localization/en/Bloom.xlf
```diff
@@ -3765,7 +3775,7 @@
         <note>ID: OpenCreateNewCollectionsDialog.CreateNewCollection</note>
       </trans-unit>
       <trans-unit id="OpenCreateNewCollectionsDialog.OpenAndCreateWindowTitle">
-        <source xml:lang="en">Open/Create Collections</source>
+        <source xml:lang="en">Open / Create Collections</source>
```
hatton, line 3778: [Devin] **Bug**: Translated dialog title string is edited in-place, which will silently discard existing translations in dozens of languages

The source text of a fully-translated localization entry is changed in-place (`OpenCreateNewCollectionsDialog.OpenAndCreateWindowTitle` at `DistFiles/localization/en/Bloom.xlf:3778`) instead of marking the old entry obsolete and creating a new one, so existing translations (Bengali, Tamil, etc.) will be overwritten on the next Crowdin sync.

Impact: Users in translated locales will lose their translated dialog title and see the English fallback.

This is an XLF-skill rule violation. Fix options: (1) revert the source back to "Open/Create Collections" to preserve all translations (the spacing tweak is minor), or (2) keep "Open / Create Collections" via the obsolete-old + new-ID-entry pattern, accepting that all locales must re-translate this title.

hatton:
[Claude Opus 4.8] I ran Devin's automated review on this PR (up through commit 19701f9b).

It raised three things, and I've put each on its own thread with what we decided:
- A possible translation-loss bug in the dialog-title spacing change — this turned out to be a false alarm for the way our Crowdin setup actually works, so we kept the change. (resolved)
- The dialog's new custom Close button — checked; keyboard (Esc) and screen-reader behavior are still covered. (resolved)
- The cards no longer having a fixed height — intentional; the new single-line cards are already uniform height. (resolved)

Greptile and CI both came back clean.

(This note is just so we can tell later that Devin has already reviewed everything up through this commit, and avoid re-running it needlessly.)

DistFiles/localization/en/Bloom.xlf
hatton: [Claude Opus 4.8] Resolving — this is a false positive for this repo's Crowdin configuration.

Per `DistFiles/localization/README.md` ("Effects of English xliff file changes") together with the current `crowdin.yml`, translations are keyed by the trans-unit **id**, not the English `<source>`. The doc is explicit that with our current `update_option: update_without_changes`, *"Editing the source element content does not affect the translation."* It is changing the **id** that removes a target translation (README line 39) — so the "mark obsolete + new id" workaround would actually be the destructive move here, not the in-place source edit.

The change here is an internal spacing tweak (`Open/Create` -> `Open / Create`), no id change and no leading/trailing-space or html-markup corner case, so existing translations (fr "Ouvrir/Créer des collections", etc.) are preserved. Leaving the source edit as-is.

src/BloomBrowserUI/collection/CollectionChooserDialog.tsx
```diff
@@ -67,53 +116,59 @@ export const CollectionChooserDialog: React.FunctionComponent<IProps> = (
...
-                            margin-left: auto;
-                            margin-right: 8px;
-                            display: flex;
-                            align-items: center;
-                        `}
-                    >
+                    <div css={titleRightGroupStyle}>
                         <UiLanguageMenu />
+                        <IconButton
+                            aria-label={closeText}
```
hatton, line 122: [Devin] Devin flagged this and thought it was worth a look:

This dialog now uses its own custom Close (X) button instead of the one the dialog framework normally provides. Devin's worry was that going custom might quietly drop the built-in behavior that comes with the standard button — for example, pressing **Esc** to close, and having the X button announced properly to screen-reader users.

src/BloomBrowserUI/collection/CollectionCard.tsx
```diff
@@ -126,47 +218,56 @@ export const CollectionCard: React.FunctionComponent<ICollectionInfo> = (
...
-                                    overflow: hidden;
-                                    text-overflow: ellipsis;
-                                    flex-grow: 1;
-                                `}
-                            >
+                            <Typography variant="body1" css={cardTitleStyle}>
                                 {props.title}
                             </Typography>
                             {props.isTeamCollection && (
                                 <TeamCollectionIcon
```
hatton, line 240: [Devin] Devin flagged this and thought it was worth a look:

The old collection cards reserved a fixed height (room for up to three lines of text) so every card was the same size. The redesign removes that fixed height. Devin wanted us to confirm the cards don't end up looking uneven or ragged now.

src/BloomBrowserUI/collection/CollectionChooserDialog.tsx
hatton: [Claude Opus 4.8] We looked into this and it's fine, no change needed.

Pressing Esc still closes the dialog: that's handled separately by BloomDialog itself (through its onCancel handler), which this dialog wires up. It never relied on the built-in close button, so replacing that button doesn't affect it. And the custom X button has a translated "Close" label attached for screen readers. So both keyboard and screen-reader users are still covered.

src/BloomBrowserUI/collection/CollectionCard.tsx
hatton: [Claude Opus 4.8] This one is intentional, no change needed.

Each redesigned card now shows just a single line of info (the book count plus any status label), and the collection name is limited to one line. So every card comes out the same height anyway — we get the same tidy, uniform grid the old fixed height was there to guarantee, just in a more compact form.

andrew-polk (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8062)* status: all files reviewed, 3 unresolved discussions (waiting on hatton).

andrew-polk (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8062)* status: all files reviewed, 3 unresolved discussions (waiting on hatton).

PR #8063, opened 2026-07-14
Grow canvas text elements to fit content during process-book (BL-16537)

JohnThomson:
### Problem
Canvas text overlays authored by BloomBridge (from EPUB/PDF) are sized to the source text geometry, which is often too short for how Bloom actually renders the text (line-height, padding, re-wrapping). The result is a scrollbar-clipped text box in BloomPUB Viewer that only corrects itself once a human opens the page in the Edit tab and saves. See BL-16537 (Blocker).

### Root cause
Bloom already auto-grows such boxes to fit, but on page load that adjustment is deferred via `AdjustSizeOrMarkOverflowSoon` on a 1000ms `setTimeout` that is **not** registered as a `requestPageContent` delay. The off-screen `process-book` capture (`captureContentForExternalProcessing`) only waits for `activeDelays`, so it snapshots the page before the timer fires and bakes in the too-short height.

### Fix
`captureContentForExternalProcessing` now grows each text canvas element to fit its content synchronously inside `finish()` — after the `activeDelays` wait, so measurements reflect the settled image sizing and fonts — before extracting the page. It calls `adjustSizeOfContainingCanvasElementToMatchContent` directly (not the full `OverflowChecker.AdjustSizeOrMarkOverflow`), so it only resizes the box and does not add page-level overflow markup or qtip debris to the captured DOM.

Only affects the off-screen `process-book` path; live editing is unchanged.

### Testing
Verified interactively: a BloomBridge EPUB conversion (which invokes `process-book`) now produces canvas text boxes at the grown height with no scrollbar, without ever opening the page in the Edit tab.

Ref: https://issues.bloomlibrary.org/youtrack/issue/BL-16537

[Devin review](https://devinreview.com/BloomBooks/BloomDesktop/pull/8063)

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8063)

### DIFF
--- a/src/BloomBrowserUI/bookEdit/js/bloomEditing.ts
+++ b/src/BloomBrowserUI/bookEdit/js/bloomEditing.ts
@@ -1467,7 +1467,9 @@ export function getBodyContentForSavePage() {
 }
 
 // Grow each text canvas element (bloom-canvas-element) to fit its content, matching what the editor
-// does automatically when a page is opened.
+// does automatically when a page is opened. (Despite the "grow" in the name, this fits to content in
+// BOTH directions -- it will also shrink a box that was authored taller than its text -- which is
+// exactly what the editor's auto-grow does; "grow" just names the case that motivated the fix.)
 //
 // On page load the editor schedules this same auto-height adjustment for every editable
 // (OverflowChecker.AddOverflowHandlers -> AdjustSizeOrMarkOverflowSoon), but that runs on a 1000ms
@@ -1497,6 +1499,14 @@ function growTextCanvasElementsToFitContent(): void {
         for (const editable of canvasEditables) {
             const [, overflowY] =
                 OverflowChecker.getSelfOverflowAmounts(editable);
+            // Calling this off-screen is DOM-safe. adjustSizeOfContainingCanvasElementToMatchContent
+            // ends by calling adjustTarget()/alignControlFrameWithActiveElement() (CanvasElementManager
+            // ~514-515), which in the live editor can add drag-game target arrows or the selection
+            // frame. Neither happens here: adjustTarget runs only AFTER the bloom-noAutoHeight
+            // early-return, so drag-game text (which is bloom-noAutoHeight) never reaches it, and a
+            // normal canvas box has no drag target to build one for; alignControlFrame no-ops when
+            // there is no active element (there isn't, off-screen). This is the same resize path the
+            // live editor already runs on auto-grow, and capture strips editing debris afterward.
             theOneCanvasElementManager.adjustSizeOfContainingCanvasElementToMatchContent(
                 editable,
                 overflowY,
--- a/src/BloomBrowserUI/bookEdit/js/bloomEditing.ts
+++ b/src/BloomBrowserUI/bookEdit/js/bloomEditing.ts
@@ -1484,17 +1484,26 @@ export function getBodyContentForSavePage() {
 // overflow markup or qtip debris to the captured DOM. Callers run this on the fully settled layout
 // (after the activeDelays wait), so text measurements reflect the final image sizing and fonts.
 function growTextCanvasElementsToFitContent(): void {
-    const canvasEditables = Array.from(
-        document.querySelectorAll<HTMLElement>(
-            `${kCanvasElementSelector} .bloom-editable.bloom-visibility-code-on`,
-        ),
-    );
-    for (const editable of canvasEditables) {
-        const [, overflowY] = OverflowChecker.getSelfOverflowAmounts(editable);
-        theOneCanvasElementManager.adjustSizeOfContainingCanvasElementToMatchContent(
-            editable,
-            overflowY,
+    // Never throws out: this runs inside finish()'s try/catch, and a failure to grow must not
+    // block capturing/saving the page (same contract as the fitImageTextSplits block). If it
+    // threw, the whole page capture would be abandoned -- strictly worse than falling back to the
+    // authored height. So we self-guard and degrade to "capture without the resize" on any error.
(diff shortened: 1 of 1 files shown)


src/BloomBrowserUI/bookEdit/js/bloomEditing.ts
```diff
@@ -1466,14 +1466,57 @@ export function getBodyContentForSavePage() {
...
+    try {
+        const canvasEditables = Array.from(
+            document.querySelectorAll<HTMLElement>(
+                `${kCanvasElementSelector} .bloom-editable.bloom-visibility-code-on`,
+            ),
+        );
+        for (const editable of canvasEditables) {
+            const [, overflowY] =
+                OverflowChecker.getSelfOverflowAmounts(editable);
+            theOneCanvasElementManager.adjustSizeOfContainingCanvasElementToMatchContent(
```
JohnThomson, line 1508: [Devin] **Investigate**: Resize path invokes editing UI helpers (adjustTarget, alignControlFrame) that add DOM elements

When `adjustSizeOfContainingCanvasElementToMatchContent` resizes a canvas element, it also calls `this.adjustTarget(canvasElement)` and `this.alignControlFrameWithActiveElement()` (CanvasElementManager.ts:514-515). `adjustTarget` can create/adjust SVG target arrows for drag games, and `alignControlFrameWithActiveElement` adjusts the selection UI frame. In the off-screen capture context there is typically no active element and no drag-game targets, so these should be no-ops. `extractAndStripPageContentForSave` also runs `removeEditingDebris`/`turnOffCanvasElementEditing` afterward, which cleans up stray UI. Still, if a drag-game page were processed off-screen, `adjustTarget` could in principle create a target-arrow SVG that persists into the captured HTML.

JohnThomson: (Claude Opus 4.8) Investigated — this is a non-issue for this path:

1. `adjustTarget` (CanvasElementManager.ts:514) runs only *after* the `bloom-noAutoHeight` early-return (lines 423-426). Drag-game draggable text elements carry `bloom-noAutoHeight`, so the resize returns before ever reaching `adjustTarget` — no game target arrow is created here.
2. For a normal (non-game) canvas text element, `adjustDraggableTarget` finds no `data-draggable-id`, so there is no target and no arrow.
3. This is the *same* resize path the live editor already runs on canvas auto-grow (via the load-time overflow timer), followed by the same `removeEditingDebris`/`turnOffCanvasElementEditing` cleanup — no stray UI has leaked from it into saved HTML.
4. As of 8ea2355, `growTextCanvasElementsToFitContent` is wrapped in its own try/catch, so even an unexpected throw degrades gracefully rather than affecting capture.

Leaving the thread open for the human reviewer's awareness.

JohnThomson:
(Claude Opus 4.8) Consulted Devin on 2026-07-14 17:23 CDT up to commit 8ea2355c0419ef4712b285b94f64cf131ea2adbb. Result: 0 bugs, 1 Investigate flag (mirrored as an inline thread with assessment), 2 informational (skipped).

hatton (changes requested):
> @hatton made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8063)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on JohnThomson).

___
*[`src/BloomBrowserUI/bookEdit/js/bloomEditing.ts` line 1488 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8063#-OxXllUy3MgfuoBTC0GT:-OxXllUy3MgfuoBTC0GU:b-k4j3jr) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/bc071d0cbe4ce8c59e26e2d5d0cafcb45750be34/src/BloomBrowserUI/bookEdit/js/bloomEditing.ts#L1488)):*
> ```XML
> // overflow markup or qtip debris to the captured DOM. Callers run this on the fully settled layout
> // (after the activeDelays wait), so text measurements reflect the final image sizing and fonts.
> function growTextCanvasElementsToFitContent(): void {
> ```

Dear Opus: So why not just simply rename it to `resizeCanvasElementsToFitContent()` instead of commenting around it?

andrew-polk (commented):
> @andrew-polk reviewed 1 file and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8063)* status: all files reviewed, 2 unresolved discussions (waiting on JohnThomson).

JohnThomson (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8063)* status: 0 of 1 files reviewed, 2 unresolved discussions.

___
*[`src/BloomBrowserUI/bookEdit/js/bloomEditing.ts` line 1488 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8063#-OxXllUy3MgfuoBTC0GT:-OxXtFEB6HGqETRgh2eG:bkbylej) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/bc071d0cbe4ce8c59e26e2d5d0cafcb45750be34/src/BloomBrowserUI/bookEdit/js/bloomEditing.ts#L1488)):*

Dear Opus: So why not just simply rename it to `resizeCanvasElementsToFitContent()` instead of commenting around it?
</blockquote></details>

[Claude Opus 4.8] Done — renamed to `resizeCanvasElementsToFitContent()` and dropped the explanatory comment, in edc5832. Agreed the name should carry it rather than a comment. Thanks!

PR #8064, opened 2026-07-14
AI Source Bubbles: improvements and fixes (BL-16549)

hatton:
[Claude Opus 4.8]

AI Source Bubbles / AI translation for Bloom: shows machine translations of a book's source text as extra source bubbles when editing, driven by 0..n configured engines (DeepL, Google, SIL Alpha2).

This branch also includes recent fixes:

- **Alpha2 "pending" status**: the poll loop now treats the `pending` (queued) `translation_status` as in-progress rather than a hard failure, which was aborting otherwise-successful translations.
- **Translation progress dialog**: on completion it now always waits, showing an **OK** button; Cancel is only present while work is in progress.
- **"Remove AI Source Translations"** book menu command (Collection tab, gated by save permission) that strips all AI source translations from a book and confirms via a Bloom message box.
- **Resilient supported-languages listing**: one engine failing to list its languages (e.g. a DeepL key without the `languages:read` scope) no longer blanks the whole target-language dropdown or blocks translation. When *every* engine fails, the error is shown in red and wraps instead of scrolling the settings dialog.
- **Fix**: Google credentials were being written onto every engine's wire record, which wiped the other engines' validation whenever Google credentials were edited.

Ref: https://issues.bloomlibrary.org/youtrack/issue/BL-16549

[Devin review](https://devinreview.com/BloomBooks/BloomDesktop/pull/8064)

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8064)


src/BloomBrowserUI/collection/AiTranslationSettingsGroup.tsx
```diff
@@ -0,0 +1,868 @@
...
+    const readyProviderIds = getReadyProviderIds(props.settings);
+    const usesEngineManagedTargetLanguages = readyProviderIds.length > 0;
+
+    const engineDisplayNames: Record<AiTranslationProviderId, string> = {
+        deepl: props.deepLEnabledLabel,
+        google: props.googleEnabledLabel,
+        alpha2: props.alpha2EnabledLabel,
+    };
+
+    const AiTranslationTargetLanguageControl: React.FunctionComponent<{
```
hatton: [Devin] **Investigate**: Target-language chooser is defined inside the settings hook, causing it to remount

`AiTranslationTargetLanguageControl` is a component defined inside `useAiTranslationSettingsGroup`, so it gets a new function identity on every render. React then unmounts/remounts the control (and its open/focus state) instead of updating it in place — so the target-language dropdown can visibly close or jump when the user edits an API key or any other setting field. Greptile flagged the same issue. Suggested fix: hoist the component to module scope and pass its inputs (loadSupportedLanguages, usesEngineManagedTargetLanguages, supportedTargetLanguages, etc.) as props.

src/BloomBrowserUI/bookEdit/sourceBubbles/BloomSourceBubbles.tsx
```diff
@@ -29,6 +38,61 @@ export default class BloomSourceBubbles {
         return $.trim($(obj).text()).length === 0;
     }
 
+    private static maybeRememberSourceBubbleLanguage(langTag: string): void {
+        postString("editView/sourceTextTab", langTag);
```
hatton, line 48: [Devin] **Investigate**: Clicking an AI source tab may be remembered as your source language

`maybeRememberSourceBubbleLanguage` posts the clicked tab's language tag to `editView/sourceTextTab`, which Bloom stores as the preferred source language for next time. AI tabs use synthetic tags like `es-x-ai-deepl`, so clicking one could persist that synthetic tag as the remembered source language, which then feeds back into later source selection. Worth a product decision on whether AI tags should be excluded here.

src/BloomExe/Collection/CollectionSettings.cs
```diff
@@ -400,6 +440,32 @@ public void Save()
...
+                engineElement.Add(
+                    new XAttribute("validationSucceeded", engine.LastValidationSucceeded)
+                );
+                engineElement.Add(
+                    new XAttribute(
+                        "validatedFingerprint",
+                        engine.ValidatedConfigurationFingerprint ?? ""
+                    )
+                );
+                engineElement.Add(new XElement("ApiKey", engine.ApiKey ?? ""));
```
hatton, line 465: [Devin] **Investigate**: API keys and Google private key stored in plain text in the collection file

The DeepL/Google/Alpha2 API keys and the Google private key are written as plain XML (`ApiKey`, `ServiceAccountEmail`, `PrivateKey`) into the `.bloomCollection` file, so anyone with the file can read them and they travel if the collection is shared. This matches how Bloom already stores other settings (pre-existing design), but noting it so the team can decide whether AI credentials warrant different handling.

hatton:
[Claude Opus 4.8] Consulted Devin on 2026-07-14 up to commit `e5711826a`.

**Bugs (4)**
- Web security disabled for the whole app browser (`--disable-web-security`) — **fixed** in `e5711826a` (Devin now marks it Resolved).
- 3× "missing justification comment on a useEffect" (`AiTranslationSettingsGroup.tsx` L416/L432/L616), tagged "Repo rule" — I couldn't find a hard rule mandating this, so treated as an optional style suggestion; left as-is (surfaced in the developer's decision report).

**Investigate (3)** — mirrored as inline threads:
- Target-language control defined inside the hook → remounts (Greptile flagged this too).
- Clicking an AI source tab may be remembered as the source language.
- API keys stored plaintext in the collection file (pre-existing design).

**Informational (4)** — not posted (low signal).

CI: `pr-automation` + PR Review Tracker green (this repo runs no build/test CI). Greptile: passed. CodeRabbit: no activity.

src/BloomExe/AiTranslation/Alpha2TranslationProvider.cs
```diff
@@ -0,0 +1,710 @@
...
+                {
+                    await TryDeleteBestEffortAsync(
+                        httpClient,
+                        apiKey,
+                        outputCollectionId.Value,
+                        ct
+                    );
+                }
+                throw;
+            }
```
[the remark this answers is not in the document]
hatton: [Claude Opus 4.8] Good catch — confirmed and fixed in commit 2e866cf18. The cleanup deletes in the catch block now use `CancellationToken.None` instead of the request's token, so when the user cancels (token already tripped) the source and output Alpha2 text collections are still deleted rather than being orphaned. Resolving.

src/BloomBrowserUI/collection/AiTranslationSettingsGroup.tsx
hatton: [Claude Opus 4.8] Fixed. The target-language control now lives at module scope (`AiTranslationTargetLanguageControl`), receiving its inputs through a React context rather than being redefined inside the hook, so it keeps a stable component identity and no longer remounts (dropdown/focus state survives edits to other fields). Devin's latest re-review marks this Resolved. Resolving.

src/BloomBrowserUI/bookEdit/sourceBubbles/BloomSourceBubbles.tsx
hatton: [Claude Opus 4.8] This is a deliberate product decision, now documented in a code comment at `BloomSourceBubbles.maybeRememberSourceBubbleLanguage` (BL-16549): remembering whichever tab the user activated — including an AI tab — is not clearly wrong, since a user who chose to view the AI translation may well want it to persist. Left as-is for now, to revisit if it proves undesirable. Resolving.

src/BloomBrowserUI/collection/LanguageChooserDialog.tsx
```diff
@@ -125,8 +128,12 @@ export const LanguageChooserDialog: React.FunctionComponent<{
     return (
         <BloomDialog
             {...propsForBloomDialog}
+            fullScreen={true}
```
hatton, line 131: [Devin] **Investigate**: LanguageChooserDialog is now fullscreen for all callers

Adding `fullScreen={true}` and `disableDragging={true}` here changes the dialog's presentation for **every** caller of `LanguageChooserDialog`, not just the new AI-translation target-language picker. In particular the existing Collection Settings "Change Language" chooser (the WinForms-hosted path) now opens fullscreen and non-draggable too. If that dialog was previously a smaller, draggable window, this is a visible UX change to an existing feature. The new inner flex-container wrapper suggests this was to fix a layout issue for the AI picker, but the flag is applied globally. Worth confirming whether making the existing chooser fullscreen everywhere is intended.

src/BloomBrowserUI/collection/AdvancedSettingsPanel.tsx
```diff
@@ -120,17 +240,77 @@ export const AdvancedSettingsPanel: React.FunctionComponent = () => {
...
 
+    // Memoized so the object reference is stable across renders that don't actually change
+    // the AI settings; the group's validation/language-fetch effects key off this reference.
+    // We key the memo on the *serialized* AI settings rather than the whole `settings` object:
+    // extractAiTranslationFlatSettings returns a fresh object every call, so keying on `settings`
+    // handed out a new reference whenever ANY unrelated setting changed. That silently cancelled an
+    // in-flight engine validation and left it stuck showing "Testing translation..." forever (the
+    // per-engine validation effect early-returns when its probe key is unchanged, so it never
+    // restarted or cleared the pending flag). Keying on the serialized value fixes that.
+    const extractedAiTranslationFlatSettings = settings
```
hatton, line 271: [Devin] **Bug**: Engine validation gets stuck in "Testing..." when an unrelated setting changes mid-test

While a per-engine translation test is running (the row shows "Testing translation..."), changing any unrelated Advanced-tab setting handed the AI settings group a brand-new `aiTranslationFlatSettings` reference (the memo keyed on the whole `settings` object, and `extractAiTranslationFlatSettings` returns a fresh object each call). That cancelled the in-flight probe — whose `finally` then skipped clearing the pending flag because it saw `cancelled` — and the effect's re-run hit the probe-key early-return without restarting, leaving the row stuck on "Testing translation..." indefinitely.

hatton: [Claude Opus 4.8] Fixed in commit 0c5ce29ed. The memo now keys on the serialized AI settings, so its reference only changes when the AI settings actually change; an unrelated edit no longer cancels an in-flight validation, and the row clears normally. Resolving.

src/BloomBrowserUI/collection/AiTranslationSettingsGroup.tsx
```diff
@@ -0,0 +1,1439 @@
...
+            languageConfigKey === lastSupportedLanguagesConfigKeyRef.current &&
+            supportedTargetLanguages.length > 0
+        ) {
+            return;
+        }
+
+        setIsLoadingSupportedLanguages(true);
+        setSupportedLanguagesMessage("");
+        try {
+            const response = await postJsonAsync(
```
hatton, line 1094: [Devin] **Investigate**: Supported-language fetches fail silently

If the `settings/aiTranslationSupportedLanguages` request (and the Alpha2 source-language fetch) fails at the network level, the client ends up with an empty language list and no visible error — the failure isn't surfaced to the user. This is largely by design (a missing list must not block the dialog, and per-engine failures are reported by the server's aggregated message), but a genuine connectivity failure is currently indistinguishable from "no languages", so flagging for a decision on whether to surface it.

src/BloomExe/web/controllers/CollectionSettingsApi.cs
```diff
@@ -360,6 +689,98 @@ private void StoreAdvancedSettingsData(ApiRequest request, CollectionSettingsDia
...
+                            InvalidateEngineValidation(engine);
+                            aiTranslationConfigurationChanged = true;
+                        }
+                    }
+                }
+            }
+
+            if (aiTranslationConfigurationChanged)
+            {
+                dialog.ChangeThatRequiresRestart();
```
hatton, line 781: [Devin] **Investigate**: Any AI-engine config change requires a Bloom restart

Any change to an AI engine's enabled state, credentials, target/source language flags the Collection Settings dialog as requiring a restart (`ChangeThatRequiresRestart`). This is consistent with how Bloom already treats most collection-settings changes, and the source-bubble generation reads this config at load, so a restart is defensible — but flagging in case a lighter-weight apply is wanted for, e.g., just editing an API key.

hatton:
[Claude Opus 4.8] Consulted Devin on 2026-07-15 (~22:35 UTC) up to commit `0c5ce29ed`. Result: 0 open bugs. Fixed this run and resolved on their threads — the target-language control remount (also Greptile P2) and the "Testing…" validation-stuck bug. Documented decision + resolved — AI source tab remembered as source language. Left open for developer decision — plaintext API-key storage, LanguageChooserDialog now fullscreen for all callers, silent supported-language fetch failures, and restart-required on any AI config change. CI green; Greptile's P1 (Alpha2 cleanup leak on cancel) fixed and resolved.

PR #8066, opened 2026-07-15
Fix BL-16536 "Culture is not supported" error

JohnThomson:
Switching to a keyboard whose input language reports a BCP-47 tag that .NET does not recognize as a culture (e.g. the Keyman IPA keyboard, which reports `und-Latn`) crashed Bloom. When the input language changes, Windows sends `WM_INPUTLANGCHANGE` to the focused window (the WebView2 control while the user is typing), and WinForms' handler builds a `CultureInfo` from the keyboard's tag, throwing `CultureNotFoundException`. Because that runs inside `WndProc` during message dispatch, it was a fatal, unhandled exception.

Fix: introduce `BloomWebView2`, a subclass of the WinForms `WebView2` control that overrides `WndProc` to run the input-language-change messages (`WM_INPUTLANGCHANGE` / `WM_INPUTLANGCHANGEREQUEST`) inside a try/catch. On `CultureNotFoundException` it logs a minor event and calls `DefWndProc` (the default processing WinForms' own handler ends with) instead of crashing. Bloom does all text editing in WebView2, which tracks its own input language independently of WinForms, so ignoring WinForms' inability to represent the keyboard is safe.

Because every browser in Bloom (the edit view in Shell, every ReactDialog, print preview, thumbnails, etc.) hosts a single WebView2 control via `WebView2Browser`, changing that control fixes the crash in all contexts.

Ref: https://issues.bloomlibrary.org/youtrack/issue/BL-16536

[Devin review](https://devinreview.com/BloomBooks/BloomDesktop/pull/8066)

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8066)

### DIFF
--- a/src/BloomExe/BloomWebView2.cs
+++ b/src/BloomExe/BloomWebView2.cs
@@ -1,0 +1,55 @@
+using System.Globalization;
+using System.Windows.Forms;
+using Microsoft.Web.WebView2.WinForms;
+using SIL.Reporting;
+
+namespace Bloom
+{
+    /// <summary>
+    /// A WebView2 control that guards against a WinForms/.NET crash (BL-16536) that happens when the
+    /// user switches to a keyboard whose input language reports a BCP-47 tag that .NET does not
+    /// recognize as a culture. For example, the Keyman IPA keyboard reports "und-Latn" (undetermined
+    /// language, Latin script). When the input language changes, Windows sends WM_INPUTLANGCHANGE to
+    /// the focused window (which, while the user is typing, is this WebView2 control). WinForms' handler
+    /// for that message builds a CultureInfo from the keyboard's tag and throws
+    /// CultureNotFoundException. Because that runs inside WndProc during message dispatch, it is
+    /// otherwise a fatal, unhandled exception on the UI thread.
+    ///
+    /// Bloom does all its text editing inside WebView2, which tracks its own input language
+    /// independently of WinForms, so we can safely ignore WinForms' inability to represent the
+    /// keyboard and just perform the default window processing.
+    /// </summary>
+    public class BloomWebView2 : WebView2
+    {
+        private const int WM_INPUTLANGCHANGEREQUEST = 0x0050;
+        private const int WM_INPUTLANGCHANGE = 0x0051;
+
+        /// <summary>
+        /// Intercepts the input-language-change messages so that an unrecognized keyboard culture
+        /// cannot crash Bloom. See the class comment for the full explanation (BL-16536).
+        /// </summary>
+        protected override void WndProc(ref Message m)
+        {
+            if (m.Msg == WM_INPUTLANGCHANGE || m.Msg == WM_INPUTLANGCHANGEREQUEST)
+            {
+                try
+                {
+                    base.WndProc(ref m);
+                }
+                catch (CultureNotFoundException e)
+                {
+                    Logger.WriteMinorEvent(
+                        "Ignoring unsupported input-language culture from keyboard (BL-16536): "
+                            + e.Message
+                    );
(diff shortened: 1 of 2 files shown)


src/BloomExe/BloomWebView2.cs
```diff
@@ -0,0 +1,55 @@
...
+    /// the focused window (which, while the user is typing, is this WebView2 control). WinForms' handler
+    /// for that message builds a CultureInfo from the keyboard's tag and throws
+    /// CultureNotFoundException. Because that runs inside WndProc during message dispatch, it is
+    /// otherwise a fatal, unhandled exception on the UI thread.
+    ///
+    /// Bloom does all its text editing inside WebView2, which tracks its own input language
+    /// independently of WinForms, so we can safely ignore WinForms' inability to represent the
+    /// keyboard and just perform the default window processing.
+    /// </summary>
+    public class BloomWebView2 : WebView2
```
JohnThomson, line 22: [Devin] **Investigate**: WebView2PdfMaker uses raw WebView2 without the crash guard

Devin (AI reviewer) noted that the separate WebView2PdfMaker project (`src/WebView2PdfMaker/WebView2PdfComponent.cs:121`) still instantiates a raw `new WebView2()` rather than the new `BloomWebView2` guard, and asked us to confirm the assumption that it can't hit the same `CultureNotFoundException` crash.

JohnThomson: (Claude Opus 4.8) Assessed as **not an issue** — resolving.

`BloomPdfMaker.exe` (the WebView2PdfMaker project) runs as a **separate external process**, launched via `MakePdfUsingExternalPdfMakerProgram` — not in-process on Bloom's UI thread. Its `ConversionHost` form is created **Minimized** with `ShowInTaskbar = false` (see `src/WebView2PdfMaker/ConversionHost.cs:19-20`): a hidden, non-interactive PDF converter with no text-entry surface. `WM_INPUTLANGCHANGE` is delivered to the window that has interactive keyboard focus while the user switches input languages; a minimized, focus-less background converter never occupies that role, so the crash path can't realistically be reached there.

Devin's own note raised this as conditional on "if the PdfMaker runs in-process on the UI thread" — it does not. Additionally, `BloomWebView2` lives in the `BloomExe` project, which WebView2PdfMaker does not reference, so guarding it would mean duplicating the code into a separate project for no practical benefit. Leaving WebView2PdfMaker as-is.

JohnThomson:
(Claude Opus 4.8) **Consulted Devin** on 2026-07-15 up to commit `6d889941870e916a9bef41edd32ec6d28d0a2a1f`.

- **0 Bugs.**
- **1 Investigate flag** — "WebView2PdfMaker uses raw WebView2 without the crash guard". Assessed as not an issue (separate minimized headless PDF process, no interactive keyboard focus); replied and resolved: https://github.com/BloomBooks/BloomDesktop/pull/8066#discussion_r3588817765
- Informational flags: none.

Greptile: reviewed clean (no findings). CI: passing.

src/BloomExe/BloomWebView2.cs
JohnThomson: (Claude Opus 4.8) Developer confirmed: **leave as is** — WebView2PdfMaker stays unguarded. Thread remains resolved.

andrew-polk (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8066)* status: all files reviewed, 1 unresolved discussion (waiting on JohnThomson).

andrew-polk (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8066)* status: all files reviewed, 1 unresolved discussion (waiting on JohnThomson).

PR #8069, opened 2026-07-15
Fix PNG file resizing for new image gallery dialog (BL-16424)

StephenMcConnel:
## What

Restores image resizing for the new image gallery dialog.

`PageEditingModel.ChangePicture` was passing `resizeFileIfNeeded: false` to `ImageUtils.ProcessAndSaveImageIntoFolder`, on the assumption that resizing would happen later in a JavaScript callback. That assumption holds for the legacy edit-page path (paste / image toolbox), but **not** for the new image gallery dialog (`ImageGalleryApi`), which has no such callback. As a result, images chosen from the gallery were never downscaled.

This removes the `resizeFileIfNeeded: false` argument so the parameter takes its `true` default, restoring resizing for all callers.

## Why this is safe

The expensive "shrink-attempted-twice on indexed-color PNGs" slowdown that originally motivated `resizeFileIfNeeded: false` (commit `9cba7a96f`) is already prevented by the `IsIndexedColorPngFile` guard added to `TryResizeImageWithGraphicsMagick` in that same commit. That guard remains, so indexed PNGs still short-circuit out of the resize attempt cheaply. Re-enabling resizing on the C# side therefore does not reintroduce the slowdown, and resizing an already-small image is a no-op.

## Testing

`PageEditingModelTests` and `ImageUtilsTests` pass (73/73).

Ref: https://issues.bloomlibrary.org/issue/BL-16424

[Devin review](https://devinreview.com/BloomBooks/BloomDesktop/pull/8069)

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8069)

### DIFF
--- a/src/BloomExe/Edit/PageEditingModel.cs
+++ b/src/BloomExe/Edit/PageEditingModel.cs
@@ -42,8 +42,7 @@ namespace Bloom.Edit
             var imageFileName = ImageUtils.ProcessAndSaveImageIntoFolder(
                 imageInfo,
                 bookFolderPath,
-                isSameFile,
-                resizeFileIfNeeded: false // resizing will happen later in a callback from javascript
+                isSameFile
             );
             try
             {
--- a/src/BloomExe/ImageProcessing/ImageUtils.cs
+++ b/src/BloomExe/ImageProcessing/ImageUtils.cs
@@ -932,9 +932,13 @@ namespace Bloom.ImageProcessing
         /// Create a display-ready processed version of an image in <paramref name="destDir"/>:
         /// resize if too large, convert format if beneficial (JPEG→PNG for transparent line art,
         /// PNG→JPEG for photographic content), and optionally make backgrounds transparent.
-        /// This contains the processing that was formerly done at import time by
-        /// <see cref="ProcessAndSaveImageIntoFolder"/>; it is now applied at display time so that
-        /// the book folder always stores something as close as we think reasonable to the original/unmodified file.
+        /// This may duplicate some processing that is done at import time by
+        /// <see cref="ProcessAndSaveImageIntoFolder"/>, but this method is also called when
+        /// publishing to possibly adjust the images even further.
+        /// After import processing, the book folder always stores something as close as we think
+        /// reasonable to the original/unmodified file and that is suitable for good quality print
+        /// publication.  The result of this method is not stored in the book folder, but is used
+        /// for display in the Bloom UI or for ebook publication.
         /// </summary>
         /// <param name="sourcePath">Path to the image file in the book folder.</param>
         /// <param name="destDir">Directory in which to write the processed copy.</param>
@@ -946,6 +950,12 @@ namespace Bloom.ImageProcessing
         /// Path of the processed image inside <paramref name="destDir"/>, or <c>null</c> if the
         /// original can be served as-is or if processing failed.
         /// </returns>
+        /// <remarks>
+        /// Care must be taken that any processing here that duplicates what is done at import time does not
+        /// slow down the UI.  For example, if the image is already a web format and small enough, we should
+        /// use it as-is.  Methods that may process the image with GraphicsMagick should avoid processing if
+        /// the result would be the same as the original. (BL-16424)
+        /// </remarks>
         public static string AdjustImageForDisplay(
             string sourcePath,
             string destDir,


src/BloomExe/Edit/PageEditingModel.cs
```diff
@@ -42,8 +42,7 @@ public static ImageInfoForJavascript ChangePicture(
             var imageFileName = ImageUtils.ProcessAndSaveImageIntoFolder(
                 imageInfo,
                 bookFolderPath,
-                isSameFile,
-                resizeFileIfNeeded: false // resizing will happen later in a callback from javascript
+                isSameFile
             );
```
StephenMcConnel, line 46: [Devin] **Investigate**: Behavioral change affects both old toolbox and new gallery image paths

Removing `resizeFileIfNeeded: false` re-enables import-time resizing for **all** callers of `PageEditingModel.ChangePicture`, not just the new image gallery dialog. The legacy edit-page path (`EditingModel.ChangePicture`, `EditingModel.cs:1799` — paste / image toolbox) flows through here too. The removed comment ("resizing will happen later in a callback from javascript") reflects the architecture documented at `ImageUtils.cs:936-937`: resizing was deliberately moved from import time to display time (`AdjustImageForDisplay`) so that "the book folder always stores something as close as we think reasonable to the original/unmodified file."

So for the legacy path this changes behavior: the book folder will now store a resized copy (capped to MaxLength=3840 / MaxBreadth=2800) instead of the original-size file, and the display-time pass becomes a redundant no-op. Only the gallery path was the reported bug.

**Assessment ([Claude Opus 4.8]):** The flag is correct and worth a decision. It is not a crash/data-loss bug — the display-time pass is idempotent and the expensive indexed-PNG double-shrink is already guarded (`IsIndexedColorPngFile`, added in `9cba7a96f`). But it does regress the legacy path away from the documented "store the original, resize at display" architecture. The alternatives are (a) accept import-time resizing for both paths and update the now-inaccurate comment at `ImageUtils.cs:936-937`, or (b) make the fix surgical — resize only for the gallery caller and leave the legacy path storing originals. Leaving this thread open pending the author's decision.

StephenMcConnel:
[Claude Opus 4.8] Consulted Devin on 2026-07-15 20:53 UTC up to commit 14934e1cf08fb72def2f79e7fb9963bf7ea677b3.

Result: **0 Bugs, 1 Investigate flag.** The flag ("Behavioral change affects both old toolbox and new gallery image paths") is posted as an inline review thread on `PageEditingModel.cs` and left **open** pending the author's decision — it raises a legitimate architectural question (re-enabling import-time resizing also changes the legacy toolbox/paste path, contradicting the display-time-resize intent documented at `ImageUtils.cs:936-937`). No Informational items posted (low signal).

src/BloomExe/Edit/PageEditingModel.cs
StephenMcConnel: [Claude Opus 4.8] **Decision (developer): keep the broad fix; the documentation was corrected.**

We deliberately keep import-time resizing enabled for all `ChangePicture` callers (both the new gallery and the legacy paste/toolbox path) — this restores the intended behavior. The behavior is safe: the display-time pass is idempotent for an already-sized image, and the expensive indexed-color-PNG case is guarded by `IsIndexedColorPngFile`.

The one real defect this flag surfaced — the `AdjustImageForDisplay` doc comment claiming the book folder "always stores something as close as we think reasonable to the original/unmodified file" — has been fixed in `0f618d7b7`. The comment now states the stored file is close to the original *and suitable for good-quality print* (i.e. downscaled to reasonable maximums), and that this method's display/publish result is not stored back into the book folder. A `<remarks>` block was also added warning that display-time processing must avoid duplicating import work in a way that slows the UI (BL-16424).

Resolving.

StephenMcConnel:
[Claude Opus 4.8] Consulted Devin on 2026-07-15 21:41 UTC up to commit 0f618d7b75fee2ca6a3d3001f2d77a5c0341957f (re-review after the doc-comment update).

Result: **0 Bugs, 1 Investigate, 1 Informational.**

- The Investigate flag ("Behavioral change: images now resized at import time instead of deferred to JavaScript") is the **same concern** as the earlier flag, which was decided and [resolved here](https://github.com/BloomBooks/BloomDesktop/pull/8069#discussion_r3590673631): keep import-time resizing for both paths, with the doc comment corrected accordingly. Devin itself now notes the updated comment (lines 938-941) states the stored file is "suitable for good quality print publication" — i.e. the print-quality tradeoff is deliberate and documented. Not re-posting it as a separate thread to avoid reopening a settled decision.
- The Informational item ("Unused pageBackgroundColor parameter in ChangePicture") is pre-existing and unrelated to this one-line change — out of scope (Informational, low signal).

CI and Greptile are green on this commit; it differs from the previously-reviewed commit only in comments.

JohnThomson (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8069)* status: all files reviewed, 1 unresolved discussion (waiting on StephenMcConnel).

PR #8073, opened 2026-07-16
Bound reader words at zero-width/directional chars in the highlighter (BL-16490)

StephenMcConnel:
## Problem

Zero-width spaces still break the Decodable Reader tool despite the earlier BL-16490 fix (which stripped orphaned CKEditor filling chars). The root cause is a mismatch between two functions in `synphony_lib.ts`:

- **The analyzer** (`getWordsFromHtmlString`) splits story text into words *on* U+200B ZERO WIDTH SPACE and the LTR/RTL/directional format characters.
- **The highlighter** (`wrap_words_extra`) bounded words only with `\s` / `\p{Z}` / `\p{P}` / `&nbsp;`. U+200B and friends are Unicode category **Cf**, not **Zs**, so `\p{Z}` never matched them.

Result: a decodable word touching or split by one of these invisible characters was counted by the analyzer yet left unmatched by the highlighter, so it displayed unmarked or as not-decodable. Blanket-stripping the characters can't fully fix this — it only fires on boxes with a live CKEditor instance, and would be wrong for scripts that use ZWSP as a legitimate word separator.

## Fix

- Extract the splitter set into a shared `zeroWidthAndDirectionalSplitters` constant (built with `String.fromCharCode` so this source file stays free of the invisible characters it describes).
- Use it both in `getWordsFromHtmlString`'s split regex (behavior unchanged — same code points as before) and in `wrap_words_extra`'s `beforeWord`/`afterWord` boundary character classes, so the two can no longer disagree about where a word ends.
- ZWNJ/ZWJ (U+200C/U+200D) remain excluded; they are legitimate within words in some scripts (BL-13428).

## Tests

Added regression tests covering both sides of the mismatch: the analyzer split (splits on ZWSP, not ZWJ) and the highlighter matching for words adjacent to and split by a ZWSP.

Ref: https://issues.bloomlibrary.org/youtrack/issue/BL-16490

[Devin review](https://devinreview.com/BloomBooks/BloomDesktop/pull/8073)

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8073)

### DIFF
--- a/src/BloomBrowserUI/bookEdit/toolbox/readers/libSynphony/jquery.text-markupSpec.js
+++ b/src/BloomBrowserUI/bookEdit/toolbox/readers/libSynphony/jquery.text-markupSpec.js
@@ -279,4 +279,74 @@ describe("jquery.text-markup", function () {
             '<p> <span class="word-not-found" data-segment="word">This</span><em> <span class="word-not-found" data-segment="word">is</span> </em>a test.</p>',
         );
     });
+
+    // The following three tests lock in that the analyzer (getWordsFromHtmlString) and the
+    // highlighter (wrap_words_extra) agree about ZERO WIDTH SPACE (U+200B) being a word
+    // boundary. They previously disagreed: the analyzer split words on U+200B while the
+    // highlighter's \p{Z}/\p{P} boundaries did not match it (U+200B is Unicode category Cf,
+    // not Zs), so a decodable word touching a ZWSP was left unmarked or mis-marked. See
+    // BL-16490. We build the invisible characters with String.fromCharCode so this source
+    // file stays free of them.
+    it("getWordsFromHtmlString splits on ZERO WIDTH SPACE but not ZERO WIDTH JOINER (BL-16490)", function () {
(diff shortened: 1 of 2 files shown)


StephenMcConnel:
[Claude Opus 4.8] Consulted Devin on 2026-07-16 15:49 MDT up to commit `facaf4535f1a482a135f08fd004e036020972c4c`.

Result: **0 Bugs, 0 Flags.** Devin raised 2 Informational notes only, both confirming the change is sound — that `getWordsFromHtmlString`'s split behavior is provably unchanged by the refactor to the shared constant, and that `wrap_words_extra`'s `afterWord` lookahead is non-consuming so the boundary characters are preserved for the following word. Informational notes are not action items, so nothing was posted as a review thread.

JohnThomson (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8073)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on StephenMcConnel).

PR #8078, opened 2026-07-19
Fix BookOnBlorgBadge refetching on every render (BL-16572)

hatton:
[Claude Opus 4.8]

Fixes the top-frequency Sentry error `System.ApplicationException: Unexpected promise failure: Network Error` (Sentry [BLOOM-DESKTOP-FFG](https://bloom-app.sentry.io/issues/7333896293/): 7,158 events / 58 users in 90 days, on Bloom@6.4.104.0).

## Root cause
`BookOnBlorgBadge`'s effect had no dependency array (`useEffect(() => { updateBadge(); });`), so it re-ran after *every* render. With many badges visible in a collection, any re-render (a websocket broadcast, a selection change) refired every badge's network request. That unbounded redundant refetching amplified any transient network hiccup into bursts of failing axios GETs, which `get()` — lacking an `errorCallback` — turned into unhandled promise rejections reported to Sentry.

## Fix
- Use the repo's `useMountEffect` helper so the badge fetches once on mount. The existing websocket subscription (`bookCollection`/`updateBookBadge`) already refreshes it on server-pushed changes, and each badge is mounted with `key={book.id}`, so book identity is stable per mount.
- Pass an `errorCallback` to `get()` so a transient network failure leaves the badge in its previous state silently instead of throwing to Sentry.

Ref: https://issues.bloomlibrary.org/youtrack/issue/BL-16572

[Devin review](https://devinreview.com/BloomBooks/BloomDesktop/pull/8078)

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8078)

### DIFF
--- a/src/BloomBrowserUI/react_components/BookOnBlorgBadge.tsx
+++ b/src/BloomBrowserUI/react_components/BookOnBlorgBadge.tsx
@@ -2,7 +2,8 @@ import { css } from "@emotion/react";
 
 import * as React from "react";
 import { get } from "../utils/bloomApi";
-import { useEffect, useState } from "react";
+import { useState } from "react";
+import { useMountEffect } from "../utils/useMountEffect";
 import { BloomTooltip } from "./BloomToolTip";
 import { Link } from "../react_components/link";
 import { useSubscribeToWebSocketForStringMessage } from "../utils/WebSocketManager";
@@ -49,6 +50,12 @@ export const BookOnBlorgBadge: React.FunctionComponent<{
                     setBadge(BadgeType.None);
                 }
             },
+            () => {
+                // A transient network failure (e.g. offline, server not ready) should not
+                // throw an unhandled promise rejection to Sentry (BLOOM-DESKTOP-FFG). Silently
+                // leave the badge in its previous state; the websocket subscription above will
+                // refresh it when the server next reports a change.
+            },
         );
     };
 
@@ -62,9 +69,12 @@ export const BookOnBlorgBadge: React.FunctionComponent<{
         },
     );
 
-    useEffect(() => {
-        updateBadge();
-    });
+    // Fetch the badge once when the component mounts. Previously this was a bare
+    // useEffect with no dependency array, so it re-ran on every render; with many badges
+    // visible in a collection, any re-render (websocket broadcast, selection change) refired
+    // every badge's network request, amplifying transient network hiccups into bursts of
+    // errors (BLOOM-DESKTOP-FFG). The websocket subscription above handles later updates.
+    useMountEffect(updateBadge);
 
     return (
         <div


src/BloomBrowserUI/react_components/BookOnBlorgBadge.tsx
```diff
@@ -62,9 +69,12 @@ export const BookOnBlorgBadge: React.FunctionComponent<{
...
 
-    useEffect(() => {
-        updateBadge();
-    });
+    // Fetch the badge once when the component mounts. Previously this was a bare
+    // useEffect with no dependency array, so it re-ran on every render; with many badges
+    // visible in a collection, any re-render (websocket broadcast, selection change) refired
+    // every badge's network request, amplifying transient network hiccups into bursts of
+    // errors (BLOOM-DESKTOP-FFG). The websocket subscription above handles later updates.
+    useMountEffect(updateBadge);
```
hatton, line 77: [Devin] **Investigate**: Badge refresh now relies solely on mount + websocket, not every render

The previous bare `useEffect` refreshed each badge on every render, which masked cases where the badge needed to update without a websocket push. With `useMountEffect`, the badge now fetches only on mount and on `bookCollection`/`updateBookBadge` websocket messages. This is safe as long as (a) each `BookButton` is keyed by `book.id` (so a book identity change forces a remount and re-fetch), and (b) the server reliably pushes `updateBookBadge` for every relevant change (upload, harvest-state transition, draft toggle, in-circulation change). If any server-side change path fails to emit that websocket message, the badge could show stale state until remount, whereas before it self-corrected on the next render. Worth confirming the server emits `updateBookBadge` for all those transitions.

hatton: [Claude Opus 4.8] Assessed as not a regression; resolving.

Both conditions hold:

- **(a) keyed by book.id** — confirmed: `BooksOfCollection.tsx` mounts each `BookButton` with `key={book.id}`, so a book-identity change forces a remount and a fresh mount fetch.
- **(b) server emits `updateBookBadge` on the relevant transitions** — confirmed in `BookExe/Collection/BookCollection.cs` (~line 636): the status-refresh loop calls `_webSocketServer.SendString("bookCollection", "updateBookBadge", bookInfo.Id)` whenever a book's `BloomLibraryStatus` changes. Every badge state (Published / Draft / Harvesting / Problem / not-in-circulation) is derived from that same `BloomLibraryStatus`, so any transition that could change the badge is exactly what triggers the push.

Also worth noting: the old every-render refetch was never a reliable freshness guard. It only "self-corrected on the next render," which itself needs something to trigger a re-render — largely the same websocket broadcasts and selection changes that already drive the badge. What it reliably did was fire a redundant network request for every visible badge on every re-render, which is precisely what turned transient network hiccups into the 7,158 Sentry errors (BLOOM-DESKTOP-FFG) this PR fixes. Mount + the existing websocket subscription is the intended and sufficient update path.

hatton:
[Claude Opus 4.8] Consulted Devin on 2026-07-18 19:05 up to commit cadac546531ac605f490a27729f32da033570ea4. Findings: 0 bugs, 1 Investigate flag ("Badge refresh now relies solely on mount + websocket, not every render") — posted as an inline thread, assessed as not a regression (server emits `updateBookBadge` on the relevant status transitions; badges are keyed by book.id), and resolved. 0 informational items.

JohnThomson (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8078)* status: all files reviewed, 1 unresolved discussion (waiting on hatton).

JohnThomson (approved):
> @JohnThomson resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8078)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on hatton).

PR #8083, opened 2026-07-20
Add bloom-player rendering coverage to visual regression tests (BL-16512)

JohnThomson:
Extends the visual-regression suite (`src/BloomVisualRegressionTests`) to also check how books render in **bloom-player**, not just in the edit/preview.

A book can look different in bloom-player even when the preview is unchanged — bloom-player renders the *staged* BloomPUB, which uses device xmatter and can differ in page count and order. So this adds a parallel bloom-player capture-and-compare pass with its own per-page reference images.

**Per book state (branding × theme), the test now also:**
- enters the publish tab and stages the book as a BloomPUB (the real `Publish:BloomPub` staging path),
- renders the staged book in bloom-player, enumerating pages from the player itself (swiper slides, since it lazy-renders), and
- screenshots each page (just the `.bloom-page` element, no player chrome) and compares it to a committed player reference.

**New DEBUG-only e2e endpoints** (compiled and registered only in Debug builds):
- `e2e/makeBloomPubPreview` — stage the selected book and return the staged htm URL (reuses `PublishApi.MakeBloomPubForPreview`; requires the publish tab, entered via the existing `workspace/selectTab`).
- `e2e/isCollectionReady` — lets the test wait for the collection tab to finish (re)loading before selecting a book, avoiding a readiness race.

**Robustness:** capture pages are parked at `about:blank` before each book is rewritten (so a stale page doesn't request files mid-rewrite); bloom-player's niceScroll scrollbar is hidden during capture (its thumb rendering was non-deterministic); and a collection guard fails fast if Bloom is already on the wrong collection.

Adds 72 player reference images (2 books × 6 states × their pages). Verified locally: two consecutive full runs (regenerate, then compare) both green with all 84 references stable; and an injected player-only defect was caught by the player pass while the preview pass stayed green.

Ref: https://issues.bloomlibrary.org/youtrack/issue/BL-16512

---
*Draft opened via preflight by Claude (Opus 4.8).*

[Devin review](https://devinreview.com/BloomBooks/BloomDesktop/pull/8083)

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8083)


JohnThomson:
[Claude/Opus 4.8] @greptile-apps please review — most of the 130 files are committed PNG reference images and Bloom-generated test fixtures; the actual code change is just a few files (the DEBUG-only e2e endpoints in `src/BloomExe/web/controllers/` and `src/BloomVisualRegressionTests/index.spec.ts`).

src/BloomExe/XmlHtmlConverter.cs
```diff
@@ -327,6 +327,14 @@ static string ConvertXhtmlToHtml5(string input)
             doc.LoadHtml(xml);
             string html = doc.DocumentNode.OuterHtml;
 
+            // The commented-CDATA markers that protect <style> content (/*<![CDATA[*/ ... /*]]>*/)
+            // pass through the Html Agility Pack here, which can leave behind empty CSS comments
+            // (/**/) inside the style block. Repeated xmatter/branding updates accumulate more of
+            // them, so a book's HTML keeps changing on every save even when nothing meaningful did.
+            // These empty comments are meaningless, and are never part of the real CDATA markers
+            // (which do not contain the substring "/**/"), so strip them out here.
+            html = Regex.Replace(html, @"[\t ]*/\*\*/(\r?\n)?", "");
```
JohnThomson: [Devin] **Bug**: Saving a book silently deletes the text "/**/" wherever it appears

Every book HTML conversion strips the text `/**/` from the entire document (this `Regex.Replace` in `ConvertXhtmlToHtml5`), not just inside `<style>` blocks. So if a book ever legitimately contains the literal characters `/**/` in page text or embedded content — e.g. a book about CSS — that text (plus its leading indentation and trailing newline) is silently removed on save.

Context: this strip was added (earlier commit on this branch) to clean up the empty `/**/` comment remnants that the commented-CDATA round-trip leaves behind on every branding/xmatter switch. The concern Devin raises is that the replacement is global rather than scoped to the CDATA remnants / `<style>` blocks it targets.

src/BloomExe/web/controllers/PublishApi.cs
```diff
@@ -754,6 +754,44 @@ BloomPubPublishSettings settings
...
+        /// <paramref name="book"/> exactly as the interactive Publish:BloomPub preview does
+        /// (full PrepareBookForBloomReader + image compression + save into PlaceForStagingBook)
+        /// and return the localhost URL of the staged book's .htm file, ready to be loaded in
+        /// bloom-player. bloom-player is designed to render the files staging produces (the same
+        /// set an unzipped .bloompub contains), so this is the only way to get a faithful player
+        /// rendering of a page. Unlike MakeBloompubPreview, this deliberately does not require the
+        /// publish tab to be active: that guard only exists to abort previews when a user rapidly
+        /// navigates away, which does not apply to a test driving staging directly.
+        /// </summary>
+        public string StageBookForBloomPubPreviewForTest(Book.Book book)
```
JohnThomson, line 767: [Devin] **Investigate**: New test-only staging path bypasses `_allLanguages` init but relies on `FromBookInfo`

`StageBookForBloomPubPreviewForTest` calls only the static `InitializeLanguagesInBook(book)` (→ `BloomLibraryPublishModel.InitializeLanguages`) and derives settings via `BloomPubPublishSettings.FromBookInfo(book.BookInfo)`, without setting the instance `_allLanguages` / `_bookForLanguagesToPublish` that the production `UpdatePreview` path sets. Worth confirming nothing downstream of `MakeBloomPubForPreview` reads `_allLanguages` when driven this way.

JohnThomson: [Claude/Opus 4.8] Not an issue — verified. `MakeBloomPubForPreview` works only from the passed `settings` (whose `LanguagesToInclude` comes from `FromBookInfo`) plus the `book`/`bookServer`; it never reads the instance `_allLanguages` / `_bookForLanguagesToPublish`. Those fields are used only by the language-selection UI path (`InitializeLanguagesInBook(request)`) and the `forVideo` branch of `UpdatePreview`, neither of which this test path hits. Confirmed empirically too: the visual-regression suite stages each book and the resulting bloom-player pages render the book's text, so the languages are correctly included. Resolving.

JohnThomson:
[Claude/Opus 4.8] Consulted Devin on 2026-07-20 up to commit `8032ebc951`.

- **Bug** — `/**/` is stripped globally from book HTML on save (`XmlHtmlConverter.cs:336`): posted as an open thread for the developer's decision (it's an existing design tradeoff with a rare user-data edge case).
- **Investigate** — test-only staging path and `_allLanguages` (`PublishApi.cs:768`): assessed **not an issue** and resolved — `MakeBloomPubForPreview` uses only the passed `settings`; confirmed by the passing visual-regression suite.
- 3 Informational items: skipped (low signal).

CI green (`pr-automation`, `track / set-waiting`).

JohnThomson:
[Claude/Opus 4.8] @greptile-apps please review — the only change since your last review is wrapping `StageBookForBloomPubPreviewForTest` in `#if DEBUG` (the guard you suggested); the 130-file count is again mostly committed PNG reference images.

JohnThomson:
[Claude/Opus 4.8] Consulted Devin again on 2026-07-20 up to commit `365da237b0` (after the `#if DEBUG` guard fix).

Re-review: **no new actionable findings.** The two prior findings are unchanged and already handled —
- Bug `/**/` strip (`XmlHtmlConverter.cs:336`): still open for the developer's decision (thread above).
- Investigate on the test-only staging path (`PublishApi.cs:768`): remains resolved as not-an-issue.

New Informational items only (skipped as low-signal), including one **confirming the DEBUG-only endpoints are correctly guarded**. CI green.

src/BloomExe/XmlHtmlConverter.cs
JohnThomson: [Claude/Opus 4.8] Fixed in `351a482637`: scoped the strip to `<style>` block contents, so an empty `/**/` is still cleaned up there but a legitimate `/**/` in a book's own text is never touched. Resolving.

JohnThomson:
[Claude/Opus 4.8] @greptile-apps please review — since the last review: scoped the `/**/` strip to `<style>` blocks (the Devin/data-loss fix) and made the visual-regression suite discover/launch Bloom on any port. (130-file count is still mostly PNG references.)

src/BloomExe/XmlHtmlConverter.cs
```diff
@@ -327,6 +327,25 @@ static string ConvertXhtmlToHtml5(string input)
...
+            // The commented-CDATA markers that protect <style> content (/*<![CDATA[*/ ... /*]]>*/)
+            // pass through the Html Agility Pack here, which can leave behind empty CSS comments
+            // (/**/) inside the style block. Repeated xmatter/branding updates accumulate more of
+            // them, so a book's HTML keeps changing on every save even when nothing meaningful did.
+            // These empty comments are meaningless, and are never part of the real CDATA markers
+            // (which do not contain the substring "/**/"), so strip them out. Scope the strip to the
+            // contents of <style> elements only: that is the only place these artifacts appear, and
+            // it means a legitimate "/**/" occurring in the book's own text (e.g. a book about CSS)
+            // is never touched.
+            html = Regex.Replace(
```
JohnThomson, line 339: [Devin] **Bug**: Stripping `/**/` inside a `<style>` block could still alter real CSS

Now that the empty-comment strip is scoped to `<style>` contents, it still removes *any* `/**/` inside a style block — so a `/**/` that is meaningful in CSS (e.g. inside a string value like `content:"/**/"`, or a legacy property hack like `color/**/:red`) would be stripped along with the empty CDATA-round-trip artifacts, changing the styling.

JohnThomson: [Claude/Opus 4.8] Assessed — not a practical issue for Bloom, so leaving the scoped strip as is:

- The only genuine-corruption case is `/**/` inside a CSS **string value** (e.g. `content:"/**/"` → `content:""`). The cited property hack `color/**/:red` is functionally identical after stripping (it becomes `color:red`; the empty comment was a no-op token separator), so that one isn't corruption.
- Bloom generates and controls these `<style>` blocks (appearance/xmatter/lang styles and user styles) for a WebView2/Chromium runtime; they don't use `/**/` inside string values or as browser hacks. The only `/**/` that actually occur here are the empty artifacts the commented-CDATA round-trip leaves behind — exactly what this strips.
- Tightening the regex to "standalone empty comments only" would risk no longer matching those real artifacts (regressing the accumulate-on-every-save fix this exists for), for no practical gain given the above.

Resolving; noting it in the preflight report as an FYI for awareness.

JohnThomson:
[Claude/Opus 4.8] Consulted Devin on 2026-07-20 up to commit `cc6b76680f`.

- The earlier `/**/` Bug and the staging Investigate are now **• Resolved** (the scope-to-`<style>` fix and the `#if DEBUG` guard).
- One **new** Bug on the fix — stripping `/**/` inside `<style>` could touch a meaningful `/**/` (e.g. `content:"/**/"`). Assessed **not a practical issue for Bloom** and resolved (reasoning on that thread); flagged as an FYI in the preflight report.
- Remaining items are Informational (skipped). CI green.

andrew-polk (commented):
> @andrew-polk reviewed 130 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8083)* status: all files reviewed, 3 unresolved discussions (waiting on JohnThomson).

andrew-polk (changes requested):
The code changes generally look ok.
I think the readme needs updating.

I have some questions in the card.

It doesn't seem to run for me when I check out this branch.

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8083#-:-Oy5epKc8eqUXcsD37cE:b-k7g5xg)* status: all files reviewed, 3 unresolved discussions (waiting on JohnThomson).

src/BloomVisualRegressionTests/collections/basic/16x9 Landscape/origami.css
```diff
@@ -298,3 +337,5 @@ https://raw.github.com/shagstrom/split-pane/master/LICENSE
   padding-left: 10px;
   display: inline-flex;
 }
+/*# sourceMappingURL=origami.css.map */
```
JohnThomson: [Devin] **Investigate**: origami.css fixture accumulates duplicate sourceMappingURL lines

The committed origami.css fixtures (16x9 Landscape and A5 Portrait) end with two identical `/*# sourceMappingURL=origami.css.map */` lines. Bloom appends a sourceMappingURL comment each time it regenerates the file, so the tracked fixture had drifted from repeated in-repo runs. Unlike branding.css / appearance.css / defaultLangStyles.css (which this branch moved to .gitignore as regenerated outputs), origami.css is tracked.

JohnThomson: [Claude Opus 4.8] Assessment: origami.css must stay tracked — it is the page-layout content the book needs to render, not a pure regenerated output, so gitignoring it (like branding/appearance/defaultLangStyles.css) would remove the layout the tests depend on. The duplicate `sourceMappingURL` line is a harmless pre-existing artifact from past in-repo runs, and the temp-copy refactor in this PR stops any further drift: the suite now copies the collection to a throwaway temp folder and drives Bloom there, so runs no longer mutate the committed fixtures at all. The only open question is whether to tidy the one duplicate line in the two committed origami.css files now — surfaced as a decision item in the preflight report. Leaving this thread open pending that call.

JohnThomson:
[Claude Opus 4.8] Consulted Devin on 2026-07-23 09:16 CDT up to commit `5e60e69e2a`.

- **No bugs.**
- 1 Investigate posted: origami.css fixture accumulates a duplicate `sourceMappingURL` line (assessment on the thread; left open as a developer fixture-hygiene decision, see the preflight report).
- 1 Investigate already **Resolved** by Devin (test-only staging path / `_allLanguages` — no action; the code initializes languages via FromBookInfo).
- ~10 Informational items skipped (Devin confirmed the DEBUG-only e2e endpoints are correctly guarded at both registration sites; the rest were test-harness heuristics/nits).

JohnThomson:
[Claude Opus 4.8] @greptile-apps please review the code changes at HEAD `5e60e69e2a`. Since your last review the branch gained an e2e test-mode (`--e2e` flag + dialog/assertion suppression) and a temp-copy test harness (index.spec.ts), and origin/master was merged in. Most of the 135 files are committed PNG reference images and generated CSS fixtures — the reviewable code is FatalExceptionHandler.cs, NonFatalProblem.cs, Program.cs, E2eTestingApi.cs, PublishApi.cs, ProjectContext.cs, XmlHtmlConverter.cs, and src/BloomVisualRegressionTests/*.ts/json.

src/BloomVisualRegressionTests/collections/basic/16x9 Landscape/origami.css
JohnThomson: [Claude Opus 4.8] Decision (developer): remove origami.css from tracking. Investigated per your note — you were right: origami.css is a shipping support file, not a fixture. BookStorage copies the current output/browser/bookEdit/css/origami.css into every book folder on BringBookUpToDate (supportFilesToUpdate / CssFilesThatAreAlwaysWanted, overwrite=true), so the committed copy never affected rendering. Confirmed empirically: the committed fixture is byte-identical to the shipping origami.css (modulo the duplicate sourceMappingURL), and with the files removed the suite's book-preview screenshots all still pass (Bloom re-supplies origami.css into the temp book folder). So the two origami.css fixtures are now git-removed and added to .gitignore alongside branding/appearance/defaultLangStyles.css — which also lets these tests catch any future unexpected change in what Bloom copies from the distribution. Resolving.

JohnThomson:
[Claude Opus 4.8] @andrew-polk — following up on your review; the branch has moved a lot since 07-21, so here's how each point stands now:

- **"readme needs updating"** — done. `readme.md` was rewritten to match the current design: the suite launches its own Bloom (no manual open), renders the committed collection by default (with a `BLOOM_VR_WORKING_TREE=1` override for deliberate book changes), and documents test failures / diff images.
- **"It doesn't seem to run for me when I check out this branch"** — this was the big one, and it's fixed. The suite now launches a **dedicated Bloom on a throwaway temp copy** (no manual setup, and it never touches your working tree), enters an **e2e mode** (`--e2e`) that suppresses error dialogs so a Bloom-side problem fails the test instead of hanging the run, and makes the bloom-player capture **deterministic** (waits on `document.fonts.ready`, image decode, and a stabilized page count). It runs green 12/12 for me. Please re-pull and try `pnpm test` in `src/BloomVisualRegressionTests`.
- **"questions in the card"** (TeamCity / submodule) — answered on the YouTrack card: the plan is a nightly slow-tests TeamCity task and/or a by-hand run before a beta; a submodule would have been nice but the bytes are already in the repo history.

Two related changes worth noting: the e2e test APIs are now gated at **runtime (`--e2e`)** instead of `#if DEBUG`, so they compile into Release for CI but stay off for normal users; and `origami.css` (a Bloom-regenerated support file) was removed from tracking so runs no longer dirty the repo.

When you get a chance, a re-review would be appreciated — several of these landed after your review.

JohnThomson:
[Claude Opus 4.8] Consulted Devin on 2026-07-23 14:25 CDT up to commit `840a9f4842`.

**No new actionable findings.**

- 1 Bug (`XmlHtmlConverter.cs:339` — an empty-comment sequence inside a `<style>` block could in principle alter CSS). This is the same concern already assessed and **resolved** on this PR (thread https://github.com/BloomBooks/BloomDesktop/pull/8083#discussion_r3615602979): the strip is scoped to Bloom-controlled `<style>` blocks, which never use a meaningful `/**/` (no CSS hacks / string values), so it is not a practical issue. Assessment unchanged.
- Prior findings Devin now marks **Resolved**: the global `/**/` strip (`:336`), the test-only staging `_allLanguages` path, the origami.css fixture duplicate, and the untracked-CSS-regenerated-by-Bloom concern.
- ~18 Informational items skipped (test-harness heuristics + e2e-mode design notes).

CI: `pr-automation` + `track/set-waiting` green. Greptile again declined the PR (too many files, 136 > 100) — the reviewable code was covered by Devin and the local review.

andrew-polk (approved):
> @andrew-polk reviewed 47 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8083)* status: all files reviewed, 4 unresolved discussions (waiting on JohnThomson).

PR #8085, opened 2026-07-20
Strip CKEditor markup before decodable-reader checks (BL-16490)

StephenMcConnel:
Strip leftover CKEditor markup before the decodable-reader check runs.

Adds a `removeCkEditorMarkup()` jQuery plugin, invoked at the start of `checkDecodableReader()`. It:

- unwraps CKEditor's inline highlight spans (spans whose `style` is a bare `background-color: rgb(...);`), keeping their text content, and
- removes CKEditor's hidden `cke_*` helper spans.

Without this, editor-inserted markup skewed the decodable-reader word markup (BL-16490).

Also declares the new method in `jquery.text-markup.d.ts` and adds unit tests for both the unwrap and hidden-span-removal paths.

Ref: https://issues.bloomlibrary.org/issue/BL-16490

[Devin review](https://devinreview.com/BloomBooks/BloomDesktop/pull/8085)

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8085)

### DIFF
--- a/src/BloomBrowserUI/bookEdit/toolbox/readers/libSynphony/jquery.text-markup.ts
+++ b/src/BloomBrowserUI/bookEdit/toolbox/readers/libSynphony/jquery.text-markup.ts
@@ -237,6 +237,7 @@ import { ReaderToolsModel } from "../readerToolsModel";
 
         // remove previous synphony markup
         this.removeSynphonyMarkup();
+        this.removeCkEditorMarkup();
 
         // get all text
         this.each(function () {
@@ -380,6 +381,28 @@ import { ReaderToolsModel } from "../readerToolsModel";
             $("body", page.contentWindow.document)
                 .find("div." + cssTooMuchStuffOnPage)
                 .removeClass(cssTooMuchStuffOnPage);
+    };
+
+    $.fn.removeCkEditorMarkup = function () {
+        this.each(function () {
+            // remove CKEditor-specific markup inserted when replacing marked text
+            $(this)
+                .find("span[style]")
+                .filter((_index, element) => {
+                    const style = (element as HTMLElement).getAttribute(
+                        "style",
+                    );
+                    return !!style && /^background-color: rgb\([0-9, ]*\);$/.test(style);
+                })
+                .contents()
+                .unwrap();
+            // remove spans belonging to CKEditor that are hidden.
+            // (Use *= not ~=: ~= matches a whole whitespace-separated token, so it can
+            // never match the multi-word value "display: none;". *= matches the substring.)
+            $(this)
+                .find("span[id^=cke_][style*='display: none;']")
+                .remove();
+        });
     };
 
     $.extend({
--- a/src/BloomBrowserUI/bookEdit/toolbox/readers/libSynphony/jquery.text-markup.ts
+++ b/src/BloomBrowserUI/bookEdit/toolbox/readers/libSynphony/jquery.text-markup.ts
@@ -392,7 +392,12 @@ import { ReaderToolsModel } from "../readerToolsModel";
                     const style = (element as HTMLElement).getAttribute(
                         "style",
                     );
-                    return !!style && /^background-color: rgb\([0-9, ]*\);$/.test(style);
+                    // rgba? and the optional "." handle highlight colors that carry an
+                    // alpha component (CKEditor serializes those as rgba(r, g, b, a)).
+                    return (
+                        !!style &&
+                        /^background-color: rgba?\([0-9, .]*\);$/.test(style)
+                    );
                 })
                 .contents()
                 .unwrap();
--- a/src/BloomBrowserUI/bookEdit/toolbox/readers/libSynphony/jquery.text-markup.ts
+++ b/src/BloomBrowserUI/bookEdit/toolbox/readers/libSynphony/jquery.text-markup.ts
@@ -392,18 +392,16 @@ import { ReaderToolsModel } from "../readerToolsModel";
                     const style = (element as HTMLElement).getAttribute(
                         "style",
                     );
-                    // rgba? and the optional "." handle highlight colors that carry an
-                    // alpha component (CKEditor serializes those as rgba(r, g, b, a)).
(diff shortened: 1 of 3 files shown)


src/BloomBrowserUI/bookEdit/toolbox/readers/libSynphony/jquery.text-markup.ts
```diff
@@ -382,6 +383,28 @@ import { ReaderToolsModel } from "../readerToolsModel";
...
+    $.fn.removeCkEditorMarkup = function () {
+        this.each(function () {
+            // remove CKEditor-specific markup inserted when replacing marked text
+            $(this)
+                .find("span[style]")
+                .filter((_index, element) => {
+                    const style = (element as HTMLElement).getAttribute(
+                        "style",
+                    );
+                    return !!style && /^background-color: rgb\([0-9, ]*\);$/.test(style);
```
StephenMcConnel: [Devin] **Investigate**: Highlight-span regex relies on browser style normalization

The unwrap filter here matches a span's style against `/^background-color: rgb\([0-9, ]*\);$/`, which insists on exactly one space after `background-color:` and after each comma, and a trailing `;`. That only works because when CKEditor sets the highlight color through the DOM, the browser hands back a normalized `background-color: rgb(255, 255, 155);` string from `getAttribute("style")`.

Worth confirming against real CKEditor output that every leftover highlight span actually serializes this way — i.e. CKEditor never leaves an `rgba(...)`, a hex color, extra whitespace, or an additional declaration in that style attribute. If it can, some leftover highlight markup would slip past the strip and still skew the decodable-reader markup.

StephenMcConnel:
[Claude Opus 4.8] Consulted Devin on 2026-07-20 16:05 MDT up to commit `364d5767c30a0ce47ef56604846b4cfd1c48785c`.

- **0 Bugs.**
- **1 Investigate flag** — "Highlight-span regex relies on browser style normalization" — posted as an inline thread (#discussion_r3617877233), left open for developer confirmation of CKEditor's actual style output.
- **1 Informational flag** (not mirrored, low signal) — notes that `removeCkEditorMarkup()` is only called from `checkDecodableReader`, not `checkLeveledReader`. Scoping the fix to the decodable reader matches the PR/ticket intent.

src/BloomBrowserUI/bookEdit/toolbox/readers/libSynphony/jquery.text-markup.ts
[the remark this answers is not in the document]
StephenMcConnel: [Claude Opus 4.8] Good catch — fixed in `30a421eea6e1e791fe72de8f069fa80f207751fb`. Widened the pattern to `/^background-color: rgba?\([0-9, .]*\);$/` so alpha-carrying highlight colors are stripped too, and added a unit test for the `rgba()` case. Resolving.

StephenMcConnel: [Claude Opus 4.8] Partly addressed in `30a421eea6e1e791fe72de8f069fa80f207751fb`: the regex now also accepts `rgba(...)` (the one concrete case Greptile raised as well), so alpha highlight colors are stripped. What I can't verify from here is the rest of what this flag asks — whether Bloom's CKEditor ever leaves a hex color, extra whitespace (e.g. a double space), or an additional style declaration on these highlight spans, any of which would still slip past the anchored pattern. Leaving this thread open for you to confirm against actual CKEditor output, since you know the editor config.

StephenMcConnel:
[Claude Opus 4.8] Consulted Devin on 2026-07-20 16:15 MDT up to commit `30a421eea6e1e791fe72de8f069fa80f207751fb` (re-review after the rgba fix).

- **0 Bugs.**
- **1 Investigate flag** — "unwrap depends on exact bare background-color serialization" — this is the same concern already tracked by the open thread on `jquery.text-markup.ts:395` (#discussion_r3617877233); not duplicated. Left open for developer confirmation of CKEditor's real style output.
- **2 Informational flags** (not mirrored, low signal): (a) an entirely empty highlight span would not be unwrapped — unlikely CKEditor output and it contributes no analyzable text; (b) the leveled-reader-scope note again.

Greptile re-reviewed the same commit and posted no new findings (its earlier rgba P2 is fixed and resolved). CI green.

src/BloomBrowserUI/bookEdit/toolbox/readers/libSynphony/jquery.text-markup.ts
StephenMcConnel: [Claude Opus 4.8] Resolved by the developer in `1f4a87d6963d386d38930c79b798d3f357626177`: the matcher was generalized to `/^background-color: [^;]*;$/`, so it now recognizes the leftover highlight span by the fact that its inline style is *only* a background-color — regardless of how the color is written (rgb, rgba, hex, or a named color). That removes the exact-serialization dependency this flag was about. The `^...$` anchor still limits it to spans that carry nothing but that one property, which is the CKEditor highlight signature. Resolving.

StephenMcConnel:
[Claude Opus 4.8] Consulted Devin on 2026-07-20 16:25 MDT up to commit `1f4a87d6963d386d38930c79b798d3f357626177` (re-review after generalizing the highlight-span matcher).

**Re-review clean — no Bugs, no Investigate flags.** The earlier Investigate flag about exact style serialization is gone (now that the matcher accepts any `background-color` value); its thread is resolved.

3 Informational observations only (not mirrored): (1) the matcher intentionally requires a single-property `background-color` style; (2) an entirely-empty highlight span wouldn't be unwrapped (unlikely; contributes no analyzable text); (3) the strip runs for the decodable reader but not the leveled reader.

JohnThomson (approved):
> @JohnThomson reviewed 3 files and all commit messages, and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8085)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on StephenMcConnel).

PR #8086, opened 2026-07-20
Add in-app Game Theme Editor (BL-16323)

hatton:
## In-app Game Theme Editor

Adds a self-contained editor for Bloom **game themes** (named sets of CSS custom properties that style drag-activity game pages), reachable from the Games tool's theme chooser in the Edit-tab toolbox.

### What's here
- **`src/gameThemeEditor/`** — a self-contained React/TypeScript editor project (color pickers, contrast checker, draggable/resizable floating panel). It depends on nothing in BloomBrowserUI; the only coupling is the `IGameThemeEditorHost` contract plus `mount()`/`unmount()`.
- **Host bridge** — `bookEdit/toolbox/games/gameThemeEditorHost.ts` mounts the editor over the live page for real-time recoloring; `ThemeChooser.tsx` gains **New…**, **Customize…**, and an edit (pencil) button that open it.
- **C# `GameThemeEditorApi`** — reads/writes theme variables (collection + developer-only factory source).
- **Theme contrast fixes** — adjustments in `gamesThemes.less` (BL-16323).

### This session's follow-up work (on top of a fresh master merge)
- Brought the branch up to date with master (yarn→pnpm migration + new tsgo type-check gate).
- **Build resolution:** because the editor lives in a sibling directory, master's stricter pipeline couldn't resolve its bare deps. Added tsconfig `paths` (type check) and a small, build-only, editor-scoped Vite plugin (production build) — the build-time counterpart to the dev server's existing `optimizeDeps` handling.
- **Localization:** added the three missing theme-chooser strings to `BloomLowPriority.xlf` (`NewTheme`, `CustomizeTheme`, `EditThemeColors`) and routed the edit-button tooltip through `useL10n`.

### Notes for reviewers
- The editor's own internal strings (color picker, contrast checker, etc.) are intentionally hardcoded English for now — it's an experimental, self-contained feature; localizing them would mean threading strings through the host contract.

Ref: https://issues.bloomlibrary.org/youtrack/issue/BL-16323

[Devin review](https://devinreview.com/BloomBooks/BloomDesktop/pull/8086)

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8086)


DistFiles/localization/en/BloomLowPriority.xlf
```diff
@@ -353,6 +353,21 @@
         <note>ID: CollectionTab.BookMenu.MoveToCurrentCollection</note>
         <note>{0} will be replaced with the name of the collection that the user currently has open for editing.</note>
       </trans-unit>
+      <trans-unit id="EditTab.Toolbox.Games.NewTheme" translate="no">
```
hatton, line 356: [Devin] **Bug**: New theme-menu labels will never appear in other languages

The three new theme-chooser labels ("New…", "Customize…", "Edit theme colors") are marked `translate="no"`, so translators never receive them and non-English users always see English — even though the sibling labels in this same menu ("Theme", the theme names) are translated.

Note for the reviewer: this repo's convention (CLAUDE.md + the `xlf-strings` skill) is to add new XLF entries as `translate="no"` by default and flip them to translatable when ready. Greptile flagged the same thing. Left open for the developer to decide whether to make these translatable now (see preflight decision report).

src/BloomExe/web/controllers/GameThemeEditorApi.cs
```diff
@@ -0,0 +1,277 @@
...
+                request.Failed(HttpStatusCode.InternalServerError, e.Message);
+                return;
+            }
+            request.PostSucceeded();
+        }
+
+        /// <summary>Upsert the theme's rule into the file at <paramref name="path"/>, and—when the
+        /// theme was renamed—remove the rule for its previous slug so a rename does not leave a
+        /// duplicate behind.</summary>
+        private static void WriteThemeToFile(string path, ThemeDto theme)
```
hatton, line 182: [Devin] **Investigate**: Theme rename removes the old rule only from the file it's saving into

When a theme is renamed, the old CSS rule is removed only from the stylesheet the renamed theme is being written to. If the original rule lived in a *different* location (e.g. defined at collection level but now saved to the book, or vice-versa), the old rule is never removed and can linger as an orphaned/duplicate theme rule.

Pre-existing editor logic; left open for developer assessment (see preflight decision report).

src/BloomBrowserUI/vite.config.mts
```diff
@@ -852,6 +900,12 @@ export default defineConfig(async ({ command }) => {
             include: [
                 "jquery", // Always pre-bundle jQuery
                 "comicaljs", // Pre-bundle comicaljs (webpack UMD bundle needs processing)
+                // The game theme editor project lives outside BloomBrowserUI (../gameThemeEditor),
+                // so Vite can't auto-resolve its bare deps from BloomBrowserUI/node_modules by
+                // walking up from those files. Pre-bundling react-rnd here makes the bare id
+                // resolvable from the editor's source regardless of its location.
+                "react-rnd",
+                "@emotion/cache",
```
hatton, line 949: [Devin] **Investigate**: Dev server may not resolve the editor's react/react-dom/@emotion/react bare imports

`optimizeDeps.include` pre-bundles only `react-rnd` and `@emotion/cache`, not react / react-dom / @emotion/react, so the editor's sibling source might not resolve those in the dev server.

hatton: [Claude Opus 4.8] Addressed in `8624b7e6e`. The `resolve-game-theme-editor-deps` Vite plugin now runs in **all** modes — production build, Vitest, and the dev server (the earlier `apply` restriction was removed). For any import originating in the editor directory it re-resolves the bare id (react, react-dom, @emotion/react, @emotion/cache, react-rnd) against BloomBrowserUI's node_modules via `this.resolve`, so the dev server no longer depends on `optimizeDeps.include` for these; react et al. stay deduped. Resolving.

hatton:
[Claude Opus 4.8] Consulted Devin on 2026-07-20 (22:36 UTC) up to commit `8624b7e6e`.

**1 Bug** — "New theme-menu labels will never appear in other languages" (`translate="no"`): posted as a thread, left open — it's a developer decision (repo convention vs. matching the translatable sibling strings). See preflight decision report.

**2 Investigate flags:**
- Theme rename removes the old rule only from the file it saves into — posted, left open for developer assessment.
- Dev server may not resolve the editor's bare imports — posted and ✅ resolved: the resolver Vite plugin now runs in all modes (dev included), so this is addressed.

**7 Informational flags** (not posted — low signal): pencil button clickable while editor open; ColorPicker outside-click effect re-subscribes each render; contrast probe/cache never invalidates; C# theme-block rewrite relies on the flat-declaration invariant (verified to hold); white-and-orange-on-blue header background semantics changed (intentional contrast fix); stale mount-location comments; unguarded localStorage write.

CI (`pr-automation`, `track`) green; Greptile green (its one P1 is the same `translate="no"` issue).

src/content/templates/template books/Games/gamesThemes.less
```diff
@@ -65,36 +65,19 @@
 }
 
 .bloom-page.game-theme-white-and-orange-on-blue {
```
hatton, line 67: [Devin] **Investigate**: white-and-orange theme drops the control-button-bg contrast override

Rewriting `.bloom-page.game-theme-white-and-orange-on-blue` removed two colour overrides that had been added *for this very ticket*, along with the comments explaining them:

- `--game-control-button-bg-color: #ffce8f;`
- `--game-checkbox-outline-color: white;`

The same `--game-control-button-bg-color` override (`#ffdccf`) was also dropped from `.game-theme-coral-reef`.

**Why this matters.** With the override gone, the value falls back through the defaults in this file — `--game-control-button-bg-color: var(--game-primary-bg-color)` (line 43) and `--game-primary-bg-color: var(--game-primary-color)` (line 20) — so it lands on the theme's `#ffb453`. The comment that was deleted alongside it recorded that this orange is **2.81:1 against the blue page, just under the 3:1 minimum for UI components**, which is exactly why the brighter `#ffce8f` (3.4:1) was introduced. `--game-control-button-bg-color` is used as a real background (`Games.less:586`), so this is visible, not cosmetic. The deleted checkbox-outline comment likewise recorded a move from a nearly-invisible grey (1.32:1) to white (4.95:1).

I verified that neither derivation was changed by this branch, so the fallback really does resolve to the lower-contrast colour.

This looks like an unintended side effect of round-tripping the factory themes through the new editor: the output is alphabetised, comments are stripped, and `dropRedundantColors` removes values it considers derivable — but these two were deliberately *not* the derived value.

Flagging rather than fixing, because re-tuning these themes may well have been intentional. Raised for a decision in the preflight report.

src/BloomBrowserUI/bookEdit/toolbox/games/ThemeChooser.tsx
```diff
@@ -44,8 +63,122 @@ export const ThemeChooser: React.FunctionComponent<{
...
     // that game pages always do).
     const [currentTheme, setCurrentTheme] = useState("");
+    // While the floating theme editor is open it owns theme changes, so we disable the
+    // dropdown to avoid switching themes out from under it. The editor can be closed from
+    // its own controls, so we track its open state via the host's subscription.
+    const [editorOpen, setEditorOpen] = useState(isGameThemeEditorOpen());
+    // Bumped when the editor closes, so we re-scan the available themes: a save may have added
+    // or renamed one (a rename removes the old name's rule from the page).
+    const [themesRefreshKey, setThemesRefreshKey] = useState(0);
+    useMountEffect(() => {
```
hatton, line 73: [Devin] **Bug**: Mount-only effects use raw empty-dependency useEffect instead of the required helper

The two setup effects added here — the editor open-state subscription and the developer-capability fetch — were written as bare `useEffect(..., [])`. `src/BloomBrowserUI/AGENTS.md` asks for the `useMountEffect` helper (`utils/useMountEffect.ts`) for mount-only effects, so that the one justified `exhaustive-deps` suppression lives in a single place instead of being repeated at each call site.

No runtime difference — this is a convention deviation.

hatton: [Claude Opus 5] Fixed in `d979ffc`. Both effects now use `useMountEffect`; the third effect in this file has real dependencies and was left as-is. Resolving.

src/gameThemeEditor/src/index.tsx
```diff
@@ -0,0 +1,52 @@
+// Public entry point of the self-contained Game Theme Editor project.
+// A host mounts the editor into a container element and passes a concrete
+// IGameThemeEditorHost; the editor knows nothing else about the host application.
+
+import { createRoot, Root } from "react-dom/client";
+import createCache from "@emotion/cache";
```
hatton: [Devin] **Investigate**: @emotion/cache may not be resolvable from BloomBrowserUI root under pnpm's isolated node_modules

`src/gameThemeEditor/src/index.tsx:6` imports `@emotion/cache`, but that package was not a direct dependency of `src/BloomBrowserUI` — only `@emotion/core`, `@emotion/react` and `@emotion/styled` were. It was reaching the build as a transitive dependency of `@emotion/react`. pnpm's default isolated layout only guarantees *direct* dependencies are linked at the project root, so this was resolving by accident of the current layout rather than because anything declared it.

hatton: [Claude Opus 5] Good catch — confirmed it was undeclared. Fixed in `d979ffc` by adding `@emotion/cache` as a direct dependency of BloomBrowserUI, pinned to `11.10.5`, the version already resolved in the tree, so no other package moves. Resolving.

DistFiles/localization/en/BloomLowPriority.xlf
hatton: [Claude Opus 5] Devin has withdrawn this one on re-review of `d979ffc`, marking it resolved with the reason: *"the XLF skill explicitly requires new entries to use translate=\"no\" unless instructed otherwise, so the prior report was incorrect."*

That matches this repo's convention — `.github/skills/xlf-strings/SKILL.md` says to add new entries as `translate="no"` and flip them when the wording has settled — so the code as written is correct, not defective.

The separate question of whether the wording is now final enough to make these three translatable is in this run's preflight decision report for John. Tracking that on the Greptile thread, which raised the same point; resolving this one as withdrawn.

DistFiles/localization/en/BloomLowPriority.xlf
```diff
@@ -353,6 +353,21 @@
...
+        <note>ID: EditTab.Toolbox.Games.CustomizeTheme</note>
+        <note>An item in the game "Theme" dropdown in the Games tool of the Edit tab's toolbox. Choosing it creates a new game theme that starts as a copy of the theme currently applied, then opens it in the theme editor. The trailing ellipsis (…) signals that an editor will open.</note>
+      </trans-unit>
+      <trans-unit id="EditTab.Toolbox.Games.EditThemeColors" translate="no">
+        <source xml:lang="en">Edit theme colors</source>
+        <note>ID: EditTab.Toolbox.Games.EditThemeColors</note>
+        <note>Tooltip (hover text) on the pencil/edit button next to the game "Theme" dropdown in the Games tool of the Edit tab's toolbox. Clicking the button opens the game theme editor to change the current theme's colors.</note>
+      </trans-unit>
     </body>
   </file>
```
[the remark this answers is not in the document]
hatton: [Claude Opus 5] Half agreed, and leaving this open deliberately.

Where this is mistaken: `translate="no"` on a *new* entry is this repo's documented convention, not an oversight. `.github/skills/xlf-strings/SKILL.md` says "Always mark new entries `translate=\"no\"` unless instructed otherwise", the intent being that a string is withheld from translators until its wording has settled, so nobody pays to translate text that then changes. Devin raised the same point independently and has since withdrawn it for exactly this reason.

Where this is right: the observable consequence you describe is real — until the attribute is removed, non-English users see "New…", "Customize…" and "Edit theme colors" in English next to menu text that *is* translated.

So the real question is whether the wording is final enough to flip now, which is the developer's call rather than something to decide from convention alone. It is in this run's preflight decision report; this thread stays open until that decision comes back, at which point the outcome gets recorded here.

hatton:
[Claude Opus 5] Consulted Devin on 2026-07-27 (17:15 UTC) up to commit `5aee1b2ab`, and re-triggered for `d979ffc77` after the fixes below.

That review raised **2 bugs, 4 investigate flags and 11 informational items**. Outcomes:

- **Mount-only effects should use `useMountEffect`** — correct; fixed in `d979ffc`. [thread](https://github.com/BloomBooks/BloomDesktop/pull/8086#discussion_r3661621092)
- **`@emotion/cache` was never declared as a dependency** — correct; it was resolving only as a transitive of `@emotion/react`. Declared directly in `d979ffc`. [thread](https://github.com/BloomBooks/BloomDesktop/pull/8086#discussion_r3661622313)
- **white-and-orange theme lost its contrast overrides** — confirmed, and it also affects coral-reef. Left open for a decision. [thread](https://github.com/BloomBooks/BloomDesktop/pull/8086#discussion_r3661620079)
- **Theme rename can orphan the old rule** — confirmed real; left open for a decision. [thread](https://github.com/BloomBooks/BloomDesktop/pull/8086#discussion_r3618010630)
- **New menu labels marked `translate="no"`** — Devin withdrew this itself on re-review, since the repo convention requires it for new entries. [thread](https://github.com/BloomBooks/BloomDesktop/pull/8086#discussion_r3618010567)
- **Dev server may not resolve the editor's bare imports** — stale: the `env.command === "build"` restriction it cites was already removed in `8624b7e`, so no action.

The 11 informational items were reviewed and none warranted a change. CI is green.

src/gameThemeEditor/src/index.tsx
```diff
@@ -0,0 +1,69 @@
...
+ * The container may live in a DIFFERENT document than the code calling mount() — in Bloom
+ * the host runs in the toolbox iframe but mounts the panel into the editable-page iframe so
+ * it can recolor the live page. So we give Emotion a cache anchored to the container's own
+ * document head; otherwise the css-prop styles would be injected into the caller's document
+ * and never reach the panel.
+ */
+export function mount(
+    container: HTMLElement,
+    host: IGameThemeEditorHost,
+): void {
```
hatton, line 34: [Devin] **Bug**: Leftover styling accumulates each time the theme editor is opened and closed

`mount()` created a fresh Emotion cache on every call, anchored to the page document's `<head>`. Emotion injects its generated `<style>` elements into that head rather than into the editor's container, so when the host removes the container on close, the style elements stay. Since the host recreates the container on each open, every open/close cycle left another full set of orphaned `<style>` nodes in the page.

Impact: a user who opens and closes the theme editor repeatedly over a long editing session slowly bloats the page's DOM.

hatton: [Claude Opus 5] Confirmed and fixed in `fae350e`. The cache is now stored alongside the React root and reused for re-renders rather than recreated, and `unmount()` calls `cache.sheet.flush()` — Emotion's own API for removing the elements it inserted. Verified with an isolated production Vite build as well as typecheck and the full Vitest suite, since this affects real bundling. Resolving.

src/gameThemeEditor/src/GameThemeEditorPanel.tsx
```diff
@@ -0,0 +1,504 @@
...
+    const newThemeName = host.getNewThemeName();
+    const isNewTheme = newThemeName !== null;
+    // Factory themes (shipped in source) can't be renamed and can only be saved back to source.
+    const isFactory = isNewTheme ? false : host.isFactoryTheme();
+    // When renaming an existing custom theme, the slug to replace (empty for new themes/factory).
+    const renameFromSlug = isNewTheme ? "" : startingSlug;
+    // Default the Save button to wherever the theme already lives: a collection theme defaults to
+    // "Save to Collection"; a factory theme to source; otherwise (book/new) to "Save to Book".
+    const themeSource = isNewTheme ? "none" : host.getThemeSource();
+    let defaultTarget: "source" | "collection" | "book";
```
hatton, line 75: [Devin] **Bug**: Save-target selection logic uses a hard-to-read stacked conditional against project style rules

The default save target was chosen with a nested ternary. The root `AGENTS.md` rules that out directly: *"Avoid stacking/nesting ternary (`? :`) operators (e.g. `a ? x : b ? y : z`). They're too hard for humans to read. Use an if/else-if chain (or a switch) instead."*

No runtime impact — a readability rule the repo states explicitly.

hatton: [Claude Opus 5] Agreed — fixed in `fae350e`, now an if/else-if chain. No behavior change. Resolving.

src/gameThemeEditor/src/GameThemeEditorPanel.tsx
```diff
@@ -0,0 +1,504 @@
...
+    };
+
+    const save = async (target: "book" | "collection" | "source") => {
+        const theme = buildTheme();
+        if (!theme.slug) {
+            setStatus("Please enter a theme name.");
+            return;
+        }
+        // Don't clobber a different existing theme by renaming/creating onto its name.
+        if (theme.slug !== startingSlug && host.themeExists(theme.slug)) {
```
hatton: [Devin] **Investigate**: Save clobber-check can be bypassed for a new theme whose name matches its base

`save()` guards against overwriting another theme with `if (theme.slug !== startingSlug && host.themeExists(theme.slug))`. For a **New**/**Customize** theme, `startingSlug` is the *base* theme's slug, because the page was switched to it when the editor opened. So if the user names the new theme identically to the base — leaving the name as the factory theme's name, say — the first half of the condition is false, the existence check is skipped, and Save writes that slug to the book or collection with no warning, silently creating a custom override of the factory theme.

Devin rates this low likelihood, and I agree: new themes are pre-seeded with an "Untitled Theme N" name, so a user has to deliberately retype the base theme's name to reach it. Noting it because the outcome is silent — the user gets an override they may not realise they created.

hatton: [Claude Opus 5] Devin promoted this to a **Bug** on re-review of `fae350e`, and on a closer look it deserved it — so this is now fixed in `3a7e8fb` rather than left for a decision.

The part I under-weighted first time round: when the base is a *custom* book or collection theme rather than a factory one, saving over it genuinely replaces that theme's colours. That is silent data loss, not just an unexpected override.

The fix compares against `renameFromSlug` instead of `startingSlug`, which turns out to be exactly the semantics the guard's own comment describes:

- new or customized theme → `renameFromSlug` is `""`, so any existing name now trips the warning;
- editing a theme in place → it equals the theme's own slug, so saving over yourself is still allowed;
- renaming → it is the old slug, so the new name is still checked against everything else.

The guard already had the right intent; it was just comparing against the theme the page happened to be showing, which for New/Customize *is* the base. Resolving.

src/BloomBrowserUI/bookEdit/toolbox/games/gameThemeEditorHost.ts
```diff
@@ -0,0 +1,512 @@
...
+export const subscribeGameThemeEditorOpen = (
+    listener: () => void,
+): (() => void) => {
+    openListeners.add(listener);
+    return () => {
+        openListeners.delete(listener);
+    };
+};
+
+// The document the floating panel mounts into. We use the TOP-LEVEL workspace document
```
hatton, line 420: [Devin] **Investigate**: Editor React root is mounted into the top document but code runs in the toolbox realm

The editor bundle executes in the toolbox iframe's JS realm, yet `openEditor` mounts the React root into the top document. Emotion is anchored to the container's `ownerDocument` head and portals use the button's `ownerDocument`, so styling and portals stay consistent — but `contrastUtils.ts` uses the module-global `document` (the toolbox document) for its colour-resolution probe, and that probe span is created, cached, and never removed. Cross-realm React 18 event delegation into the top document is also unusual.

hatton: [Claude Opus 5] Assessed as not an issue, on two counts.

**The probe span is not a leak.** `probeEl` in `contrastUtils.ts` is a module-level cached singleton — `getProbe()` creates it only `if (!probeEl)`. So it is exactly one hidden 0×0 absolutely-positioned span for the lifetime of the module, reused for every colour resolution, not one per open or per lookup. (Contrast this with the Emotion caches, which genuinely did accumulate per open/close — that one was real and is fixed in `fae350e`.)

**The cross-realm mount is deliberate and documented.** The header comment of `gameThemeEditorHost.ts` explains it: the editor is driven from the toolbox because the page-frame bundle is not served by the Vite dev server on this branch, and the panel is mounted into the page document precisely so it can float over the live game and recolour it in real time. Devin's own analysis confirms the two things that would actually break — Emotion styling and portals — are correctly anchored to the container's `ownerDocument`. And as Devin notes, the colour maths itself is document-independent.

No change needed. Resolving.

src/gameThemeEditor/src/GameThemeEditorPanel.tsx
```diff
@@ -0,0 +1,510 @@
...
+    // save write a new slug without deleting the original.
+    const newThemeName = host.getNewThemeName();
+    const isNewTheme = newThemeName !== null;
+    // Factory themes (shipped in source) can't be renamed and can only be saved back to source.
+    const isFactory = isNewTheme ? false : host.isFactoryTheme();
+    // When renaming an existing custom theme, the slug to replace (empty for new themes/factory).
+    const renameFromSlug = isNewTheme ? "" : startingSlug;
+    // Default the Save button to wherever the theme already lives: a collection theme defaults to
+    // "Save to Collection"; a factory theme to source; otherwise (book/new) to "Save to Book".
+    const themeSource = isNewTheme ? "none" : host.getThemeSource();
```
hatton, line 74: [Devin] **Investigate**: Collection theme has no primary Save button while Save-to-Collection is disabled

`defaultTarget` can still evaluate to `"collection"`, but the **Save to Collection** button is commented out as "temporarily disabled". For a theme whose source is the collection, that leaves no highlighted default action at all: **Save to Book** renders with the non-primary `outline` variant, because `defaultTarget === "book"` is false.

Two consequences worth confirming are intended for now:

1. **No primary action.** A user editing a collection theme sees only outline buttons, so nothing indicates the expected action.
2. **Silent duplication.** Saving to the book writes the theme there while the collection copy stays on disk, so the same slug ends up defined in both places. (Related to, but distinct from, the rename-orphan thread above.)

I confirmed the button is commented out and that `defaultTarget` is still allowed to be `"collection"`, so both follow. Not a crash, and not fixed here — the right resolution depends on why collection-save is disabled, which is a call for the developer. Raised in the preflight decision report.

hatton:
[Claude Opus 5] Consulted Devin again on 2026-07-27 (18:03 UTC) up to commit `3a7e8fb37` — **all bugs now clear**.

Devin marks all 5 bugs it raised across this run as fixed, and reports **0 unfixed bugs** at this commit. Over the three reviews it raised 5 bugs and 7 investigate flags; 15 informational items were reviewed and none warranted a change.

Fixed and resolved this run: the `useMountEffect` convention, the undeclared `@emotion/cache`, Emotion styles leaking on every editor open/close, a stacked ternary, and the save guard that let a new theme silently overwrite the theme it was based on.

Assessed as not an issue and resolved: the cross-realm React root (the probe span is a cached singleton, and the mount target is deliberate and documented). Two flags were stale — they cite a `vite.config.mts` restriction removed in `8624b7e` and an `@emotion/cache` declaration added in `d979ffc`.

**Three threads are deliberately left open** for the developer, all in the preflight decision report:
- [white-and-orange and coral-reef lost their contrast overrides](https://github.com/BloomBooks/BloomDesktop/pull/8086#discussion_r3661620079)
- [rename can orphan the old rule in the other stylesheet](https://github.com/BloomBooks/BloomDesktop/pull/8086#discussion_r3618010630)
- [no primary Save button while Save-to-Collection is disabled](https://github.com/BloomBooks/BloomDesktop/pull/8086#discussion_r3661779555)

plus [Greptile's P1](https://github.com/BloomBooks/BloomDesktop/pull/8086#discussion_r3617913198) on `translate="no"`, pending the decision on whether to make those labels translatable now.

CI is green. The bot fix/re-review loop reached this skill's 4-cycle cap, so the one new flag at this commit was posted rather than fixed.

hatton:
[Claude Opus 5] **Superseded by #8118 — closed by a branch rename, not abandoned.**

This work was sharing BL-16323 with the game-theme colour tweaks that already shipped separately in #7985. It now has its own card, **BL-16618**, and the branch was renamed `game-theme-editor` → `BL-16618-game-theme-editor` to match. GitHub closes a PR when its head branch is renamed, so the work continues in **#8118** on the same commits.

The review history here is still the record: the Devin consultation log and the mirrored finding threads live on this PR. #8118 links back here and lists the four findings that were still open, so nothing is lost — but new review activity should go to #8118.

One of those four (the factory-theme contrast regression in `gamesThemes.less`) is resolved in #8118: that file has been reverted to master, restoring the #7985 contrast fixes.

PR #8090, opened 2026-07-21
Fix BL-16586 talking book text checksum doesn't remove every vertical line character

andrew-polk:
To decide whether a recording still matches its text, Talking Book compares the text with the temporary `|` phrase markers removed. The code that removed them used `message.replace("|", "")`, which only removes the **first** `|`. So when a piece of text contains two or more `|`, a leftover `|` can make Bloom think the words changed when they didn't.

**Fix:** strip *every* `|` with `message.replace(/\|/g, "")`, so the comparison is always correct.

**User impact:** none observed. In "record by sentence" mode the `|` acts as a sentence break, so text is split into separate audio chunks and no chunk holds more than one `|` (the bug can't arise). In "record by whole text box" mode the box is one chunk and can contain two `|`, so the faulty comparison does happen — but its only effect is to re-do the audio markup on the box, which preserves the recording, its id, and the phrase timings. So this is a latent correctness bug, not audio loss.

**Tests:** added `talkingBookChecksumSpec.ts` (multiple markers, a single marker, and the md5-of-stripped-text case). The multi-marker tests fail on the old code and pass on the new. Full talkingBook suite (233 tests) still green.

**Note:** the card also flagged the misspelled `"undefind"` return value as a typo. It's deliberate — an intentionally unique, greppable sentinel, purposely different from the `"undefined"` that `isMissingRecordingChecksum` treats as "no recording yet", and that branch is test-only. Left unchanged.

Ref: https://issues.bloomlibrary.org/youtrack/issue/BL-16586

[Devin review](https://devinreview.com/BloomBooks/BloomDesktop/pull/8090)

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8090)

### DIFF
--- a/src/BloomBrowserUI/bookEdit/toolbox/talkingBook/talkingBookChecksum.ts
+++ b/src/BloomBrowserUI/bookEdit/toolbox/talkingBook/talkingBookChecksum.ts
@@ -21,6 +21,8 @@ export function getChecksum(message: string): string {
     // However, we don't want the sentence markup to be updated because the checksums differ (since a character was deleted).
     //
     // Thus, our checksum function needs to ignore the vertical line character when computing the checksum.
-    const adjustedMessage = message.replace("|", "");
+    // Use a global replace: a single audio-sentence can contain more than one delimiter
+    // (e.g. "one | two | three."), and every one of them must be stripped (BL-16586).
+    const adjustedMessage = message.replace(/\|/g, "");
     return getMd5(adjustedMessage);
 }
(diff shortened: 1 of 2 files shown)


andrew-polk:
Claude Opus 4.8: Consulted Devin on 2026-07-21 11:40 USMST up to commit fa8697a41ef1dfdf130bb5e25e1e517521dc4400. Re-review clean — no Bugs, no Investigate flags, no Informational items.

StephenMcConnel (approved):
:lgtm:

> @StephenMcConnel reviewed 2 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8090#-:-Oy9m4Wk7Tt_L-TOVPnU:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on andrew-polk).

PR #8098, opened 2026-07-23
Add nightly build-and-test GitHub Actions workflow

andrew-polk:
Adds `.github/workflows/nightly.yml`: a scheduled (04:00 UTC daily) + on-demand
(`workflow_dispatch`) **health-check** build of everything on `master`.

**What it does**
- Builds the front-end (`build-prod`) and the C# solution, and runs *both* test suites
  (vitest via `test:ci` + NUnit via `TestOnly`).
- Renders results as a check run + job-summary report (EnricoMi publish-unit-test-result-action)
  and uploads the raw junit/nunit XML as an artifact (14-day retention).
- Pulls the newest floating front-end deps (bloom-player alpha, latest comicaljs /
  language-chooser / image-gallery) rather than the lockfile-pinned versions — matching the
  TeamCity master build.

**What it deliberately does not do**
- No installer, no signing, no publishing. It exists to catch breakage the PR checks miss
  (tests excluded from PR runs, dependency/runner drift) on a predictable cadence.

**Build-flow note:** `RestoreBuildTasks` is run in its own msbuild invocation with
`RestartBuild=false` so it restores `SIL.BuildTasks` (needed for the `UsingTask` declarations to
load) without also compiling the whole app — that compile then happens exactly once in the
`Build` step.

No YouTrack ticket associated (infrastructure/tooling change).

[Devin review](https://devinreview.com/BloomBooks/BloomDesktop/pull/8098)

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8098)

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -1,0 +1,169 @@
+# Nightly build + full test run of everything on master.
+#
+# This is a health check, not a release: it builds the front-end and the C# solution and
+# runs both test suites (vitest + NUnit), but produces no installer, does no signing, and
+# publishes nothing. It exists to catch breakage that the PR checks miss — e.g. tests
+# excluded from PR runs, or rot from dependency/runner drift — on a predictable cadence.
+#
+# Schedule: 04:00 UTC daily. GitHub cron is always UTC (== GMT, no DST), so this is a
+# literal 4am GMT. Note GitHub may delay scheduled runs during peak load; exact timing is
+# best-effort. Can also be run on demand via the Actions "Run workflow" button.
+
+name: Nightly Build and Test
+
+on:
+    schedule:
+        - cron: "0 4 * * *"
+    workflow_dispatch:
+
+# Don't stack nightlies: if a manual run overlaps the scheduled one, let the first finish.
+concurrency:
+    group: nightly
+    cancel-in-progress: false
+
+jobs:
+    build-and-test:
+        runs-on: windows-latest
+        timeout-minutes: 120
+
+        # checks: write lets the test-report step create its check run.
+        permissions:
+            contents: read
+            checks: write
+
+        env:
+            # VersionNumbers requires a 4-part BUILD_NUMBER; nightlies never publish, so a
+            # throwaway value that just identifies the run is fine.
+            BUILD_NUMBER: 0.0.${{ github.run_number }}.0
+
+        steps:
+            - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
+
+            # ----- Toolchains -----
+
+            - uses: actions/setup-dotnet@26b0ec14cb23fa6904739307f278c14f94c95bf1 # v5.4.0
+              with:
+                  dotnet-version: 8.0.x
+
+            - uses: microsoft/setup-msbuild@30375c66a4eea26614e0d39710365f22f8b0af57 # v3
+
+            - uses: pnpm/action-setup@0ebf47130e4866e96fce0953f49152a61190b271 # v6.0.9
+              with:
+                  package_json_file: src/BloomBrowserUI/package.json
+
+            - uses: actions/setup-node@48b55a011bda9f5d6aeb4c2d9c7362e8dae4041e # v6.4.0
+              with:
+                  # Track the repo's canonical Node version (.node-version, currently
+                  # 24.13.0 — also what package.json devEngines pins) rather than a
+                  # hardcoded major, so the nightly always tests the version developers
+                  # actually use and never drifts from it.
+                  node-version-file: .node-version
+                  cache: pnpm
+                  cache-dependency-path: |
+                      src/BloomBrowserUI/pnpm-lock.yaml
+                      src/content/pnpm-lock.yaml
+
+            # ----- Dependencies (mirrors init.sh) -----
+
+            - name: Get binary dependencies (TeamCity artifacts, chm, ffmpeg, etc.)
+              shell: bash
+              run: ./build/getDependencies-windows.sh
+
+            - name: pnpm install
+              shell: bash
+              run: |
+                  # Install the two workspaces in parallel, but wait on each PID
+                  # individually: a bare `wait` (no operands) always exits 0, so it would
+                  # mask a failed install (e.g. an out-of-date lockfile) and let the step
+                  # pass green. `wait <pid>` propagates that child's exit status instead,
+                  # so under the runner's `set -e` a failed install fails the step.
+                  (cd src/content && pnpm install --frozen-lockfile) &
+                  pid_content=$!
+                  (cd src/BloomBrowserUI && pnpm install --frozen-lockfile) &
+                  pid_ui=$!
+                  wait "$pid_content"
+                  wait "$pid_ui"
+
+            # Master/alpha builds always take the newest floating front-end deps rather
+            # than the lockfile-pinned versions (bloom-player from its alpha channel,
+            # latest comicaljs / language-chooser / image-gallery), matching the TeamCity
+            # master build.
+            - name: Update floating front-end dependencies
+              working-directory: src/BloomBrowserUI
+              shell: bash
+              run: pnpm update bloom-player@alpha comicaljs@latest "@ethnolib/language-chooser-react-mui@latest" bloom-image-gallery
+
+            # ----- Front-end build + tests -----
+
+            - name: Build browser UI (production)
+              id: build_frontend
+              working-directory: src/BloomBrowserUI
+              run: pnpm run build-prod
+
+            # The junit file feeds the "Publish test results" step at the end of the job.
+            - name: Run browser UI tests
+              working-directory: src/BloomBrowserUI
+              run: pnpm run test:ci -- --reporter=default --reporter=junit --outputFile.junit=../../output/Tests/vitest-junit.xml
+
+            # ----- C# build + tests -----
+            # We want the C# suite to run even when the front-end TESTS fail, so a broken
+            # night still reports C# breakage too — not just half the picture. But the C#
+            # build needs the front-end BUILD output, so the whole C# group is gated on
+            # build_frontend having succeeded. Each C# step additionally requires the
+            # previous one to have succeeded, which preserves fail-fast within the group (a
+            # failed restore shouldn't kick off a doomed build, etc.). !cancelled() is what
+            # lets these run despite the earlier front-end-test failure — it drops the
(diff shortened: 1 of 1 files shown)


.github/workflows/nightly.yml
```diff
@@ -0,0 +1,138 @@
...
+              shell: pwsh
+              run: msbuild Bloom.proj /t:Build /m /p:BUILD_NUMBER=$env:BUILD_NUMBER
+
+            - name: Run C# tests
+              working-directory: build
+              shell: pwsh
+              run: |
+                  msbuild Bloom.proj /t:TestOnly `
+                    /p:BUILD_NUMBER=$env:BUILD_NUMBER `
+                    /p:excludedCategories=SkipOnTeamCity
```
andrew-polk, line 144: [Devin] **Bug**: A front-end failure stops the C# suite from ever running in the nightly health check

The workflow's steps run in sequence, and GitHub Actions aborts a job at the first failing step unless a later step opts out (`if: ${{ always() }}` / `if: ${{ !cancelled() }}`). So if **Build browser UI (production)** or **Run browser UI tests** (lines 86–93) fails, the three C# steps below — *Restore MSBuild build tasks*, *Build*, *Run C# tests* (lines 101–117) — are skipped entirely, since none of them run unconditionally.

**Why it matters:** on any night the front-end is broken, C# regressions go completely undetected until the front-end is green again — which partly undermines the workflow's stated goal of running *both* suites every night as a health check.

This is a design tradeoff rather than a clear-cut bug, so I'm leaving it for the developer to decide (see the preflight decision report). A reasonable fix is to add `if: ${{ !cancelled() }}` to the C# steps (all three, so the build-tasks restore still happens) so C# work runs even after a front-end failure — but note the C# `Build` step may itself depend on the front-end build output, so a broken `build-prod` might just yield a second, distinct C# build failure. Alternatives: run the C# suite as a separate parallel job, or keep the current fail-fast behavior.

— posted by Claude (Opus 4.8) via preflight

andrew-polk:
Consulted Devin on 2026-07-23 up to commit `c8abd849726839a4157057f168155f7f80420165`.

- **1 Bug** — "A front-end failure stops the C# suite from ever running" → posted as an open review thread (#discussion_r3640780714) for a developer decision (design tradeoff; see preflight report).
- **0 Investigate** flags.
- **4 Informational** flags (not mirrored — low signal): three independently *confirm* correctness (NUnit result path matches `Bloom.proj`'s non-TeamCity output; the vitest `--reporter=junit --outputFile.junit` CLI override is effective; the two-phase `RestoreBuildTasks`/`Build` avoids a double compile). The fourth notes the `always()` comment wording is imprecise (harmless) — carried into the preflight report as an optional cleanup.

Greptile: complete, clean (workflow "internally consistent with the repository's build configuration", no inline findings). CodeRabbit: config present but inactive on this repo. CI checks (pr-automation, track): passing.

— posted by Claude (Opus 4.8) via preflight

.github/workflows/nightly.yml
andrew-polk: [Devin] ✅ Fixed in `5c779ffdbc`. The three C# steps now run even when the front-end **tests** fail: they're gated on the front-end **build** step (`build_frontend`) succeeding via `if: ${{ !cancelled() && steps.build_frontend.outcome == 'success' }}`, with each C# step also requiring the previous one's success so fail-fast is preserved within the group. Per the developer's decision, the C# build genuinely needs the front-end build output, so a failed `build-prod` still correctly skips the C# suite. Resolving.

— posted by Claude (Opus 4.8) via preflight

.github/workflows/nightly.yml
```diff
@@ -0,0 +1,162 @@
...
+
+            - uses: microsoft/setup-msbuild@30375c66a4eea26614e0d39710365f22f8b0af57 # v3
+
+            - uses: pnpm/action-setup@0ebf47130e4866e96fce0953f49152a61190b271 # v6.0.9
+              with:
+                  package_json_file: src/BloomBrowserUI/package.json
+
+            - uses: actions/setup-node@48b55a011bda9f5d6aeb4c2d9c7362e8dae4041e # v6.4.0
+              with:
+                  node-version: 22
```
andrew-polk: [Devin] **Investigate**: Node version pinned to 22 conflicts with the repo's 24.13.0

This step pins `node-version: 22`, but the repo otherwise standardizes on **Node 24.13.0**: both `src/BloomBrowserUI/package.json` and `src/content/package.json` declare `devEngines.runtime.version: 24.13.0` (with `onFail: download`), and `.node-version` is `24.13.0`. So the nightly builds and tests against a different Node major version than developers actually use — which partly works against the workflow's goal of catching runner/dependency drift.

It will likely still function (Node 22.x satisfies vite 7 / vitest 4), and pnpm's `devEngines` `onFail: download` may even fetch 24.13.0 to run the `pnpm run` scripts regardless of what `setup-node` installed — so the pin may be partly moot. But the mismatch is worth resolving deliberately.

Left open for a developer decision (see the preflight report): bump to `24` / `24.13.0` to match dev + `.node-version`, or keep 22 if intentional.

— posted by Claude (Opus 4.8) via preflight

.github/workflows/nightly.yml
```diff
@@ -0,0 +1,162 @@
...
+                  # Install the two workspaces in parallel, but wait on each PID
+                  # individually: a bare `wait` (no operands) always exits 0, so it would
+                  # mask a failed install (e.g. an out-of-date lockfile) and let the step
+                  # pass green. `wait <pid>` propagates that child's exit status instead,
+                  # so under the runner's `set -e` a failed install fails the step.
+                  (cd src/content && pnpm install --frozen-lockfile) &
+                  pid_content=$!
+                  (cd src/BloomBrowserUI && pnpm install --frozen-lockfile) &
+                  pid_ui=$!
+                  wait "$pid_content"
```
andrew-polk, line 84: [Devin] **Bug**: Dependency install failures were silently ignored

The `pnpm install` step originally launched both workspace installs in the background and joined them with a bare `wait`. A bare `wait` (no operands) always exits 0, so a failed install (e.g. an out-of-date frozen lockfile) would pass green — hiding exactly the dependency breakage this nightly exists to catch. Good catch.

— posted by Claude (Opus 4.8) via preflight

andrew-polk: [Devin] ✅ Fixed in `b874b39b`: each background install's PID is now captured and `wait "$pid_content"` / `wait "$pid_ui"` waits on them individually, so a non-zero child status propagates and (under the runner's `set -e`) fails the step. Parallelism preserved. Resolving.

— posted by Claude (Opus 4.8) via preflight

andrew-polk:
Consulted Devin on 2026-07-23 up to commit `b874b39b56ec2c7e3dfe8b336bbd7ae9f5dbbcbe` (re-review after two fixes).

- **1 open Investigate** — "Node version pinned to 22 vs the repo's 24.13.0" → posted as an open thread (#discussion_r3640948480) for a developer decision (see preflight report).
- **2 Bugs re-listed, both already fixed on this commit — stale re-flags:**
  - "Dependency install failures silently ignored (bare `wait`)" — a *real* catch; fixed in `b874b39b` (wait on each PID individually). Documented + resolved (#discussion_r3641173496).
  - "Front-end failure stops the C# suite" — fixed in `5c779ffdbc` (C# steps gated on the front-end build via `!cancelled()`); already documented + resolved in #discussion_r3640780714. Devin re-listed it with the pre-fix line numbers.
- **5 Informational** (not mirrored): three independently *verify* the fixes/design are correct — the C# step gating against `Bloom.proj`, the parallel-install failure propagation, and the two-phase no-double-compile. The `always()`-comment-imprecise note is stale (reworded in `5c779ffdbc`).

Greptile: complete, clean on every commit (no inline findings). CodeRabbit: inactive on this repo. CI (pr-automation, track): passing.

.github/workflows/nightly.yml
andrew-polk: [Devin] ✅ Fixed in `94a2593f` per the developer's decision: setup-node now uses `node-version-file: .node-version` instead of the hardcoded `22`, so the nightly tracks the repo's canonical Node version (currently 24.13.0, matching `package.json` devEngines) and can't drift from it in future. Resolving.

— posted by Claude (Opus 4.8) via preflight

.github/workflows/nightly.yml
```diff
@@ -0,0 +1,169 @@
...
+            - name: Upload test results
+              if: ${{ always() }}
+              uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
+              with:
+                  name: nightly-test-results
+                  # Forward slashes: upload-artifact treats the path as a glob, where a
+                  # backslash is an escape character (not a separator) even on Windows, so
+                  # back-slashed paths can silently match nothing. Matches the publish step.
+                  path: |
+                      output/Tests/Release/x64/TestResults.xml
```
andrew-polk, line 167: [Devin] **Investigate**: Artifact upload used Windows backslash paths while the publish step used forward slashes

The "Upload test results" step listed `output\Tests\...` (backslashes), whereas `actions/upload-artifact` treats `path` as a glob where a backslash is an escape character — not a separator — even on Windows. So the back-slashed paths could silently match nothing and upload an empty artifact. Good catch.

— posted by Claude (Opus 4.8) via preflight

andrew-polk: [Devin] ✅ Fixed in `f5be0559`: switched both upload paths to forward slashes (matching the publish step and the paths the build writes). Resolving.

— posted by Claude (Opus 4.8) via preflight

andrew-polk:
Consulted Devin on 2026-07-23 up to commit `f5be055916308df5eef2eadd02a7c06a224f382e` (final re-review).

**Re-review clean — no new findings; all real findings fixed & resolved.**

Every finding on this commit's review is a stale re-flag of an item already fixed and resolved on this PR:
- 2 Bugs (masked install failure; C# skipped on front-end failure) — re-listed with pre-fix line numbers; both fixed and resolved earlier (`b874b39`, `5c779ff`).
- 2 Investigate (backslash artifact paths; Node 22 vs 24.13.0) — both fixed and resolved this run (`f5be0559`, `94a2593`); re-listed at old line numbers.

7 Informational — none actionable; several independently *verify* the fixes and design are correct (C# gating semantics, parallel-install failure propagation, no double compile, vitest path, and that the floating-dep `minimumReleaseAge` gate is correctly bypassed via `minimumReleaseAgeExclude` + the git dep).

All Devin review threads on this PR are resolved. Greptile: clean on every commit. CodeRabbit: inactive on this repo. CI (pr-automation, track): passing.

PR #8091, opened 2026-07-21
Fix three Talking Book audio-highlighting regressions (BL-15300)

JohnThomson:
Fixes three regressions in the reworked Talking Book audio-highlighting (Edit tab, CSS Custom Highlight API) reported against BL-15300.

**Problem 1 — text highlight lost after paging through a picture book.** In picture books the recordable text lives inside a canvas element, so highlighting it makes that element the `CanvasElementManager`'s active element — a reference not cleared on page navigation. `setCurrentAudioElementToDefaultAsync` treated any truthy active canvas element as "don't highlight text" and returned early, so once the reference went stale (detached after a page turn) no page highlighted again. Fix: only defer to the active canvas element when it's actually part of the current page (`pageDocBody.contains(...)`).

**Problem 2 — no yellow highlight after clearing a split recording.** `clearRecordingAsync` cleared the blue split `::highlight` paint but never re-registered the yellow current highlight (the old DOM-class model did this automatically). Fix: refresh the highlights at the end of `clearRecordingAsync`.

**Problem 3 — whole box highlighted instead of first sentence on mode switch.** `setRecordingModeAsync` decided whether to convert whole-textbox markup into per-sentence spans based on `previousRecordMode`, a field on the session-long singleton that is never reset when the page/box changes. A stale value sent the switch down the "don't convert" branch, so the box kept its whole-box `audio-sentence` and was painted whole. Fix: decide from `this.recordingMode` captured at entry (re-derived per page) and remove the stale field.

Each fix has a regression test (verified fix-sensitive) in `audioRecordingSpec.ts`. Full talking-book vitest suites pass (233 tests).

Ref: https://issues.bloomlibrary.org/youtrack/issue/BL-15300

[Devin review](https://devinreview.com/BloomBooks/BloomDesktop/pull/8091)

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8091)

PR #8100, opened 2026-07-23
More Talking Book highlighting fixes (BL-15300)

JohnThomson:
Follow-up to #8091. Fixes several more Talking Book audio-highlighting issues in the Edit tab (which paints the highlight with the CSS Custom Highlight API / `::highlight`), all found while testing the reworked highlighting (BL-15300).

**1 — Highlight missing after changing pages (`78f2b16`).** The recorder tracks the current highlight's DOM nodes in `highlightedElement`; when the page frame reloads a beat after the current element is selected, that field is left pointing at a same-id node in the *previous, detached* document, whose window's `CSS.highlights` is null, so `refreshHighlights` silently did nothing and no highlight appeared. `ensureHighlight`'s new `reestablishCurrentHighlightIfNeeded()` detects the staleness (membership in the current page document — `isConnected` can't catch it), re-points at the live node by id, and re-registers.

**2 — Whole box highlighted instead of first sentence on mode switch (`442d0e2`).** `setRecordingModeAsync` chose whether to convert whole-textbox markup to per-sentence spans from a session-long `previousRecordMode` field that isn't reset when the page/box changes; a stale value skipped the conversion. Now it decides from the current box's mode captured at entry. Also: the markup action no longer rewrites `innerHTML` when the result is unchanged (idempotent — `newPageReady` deliberately fires twice), and `clearRecordingAsync` re-shows the yellow highlight after clearing a split recording.

**3 — Highlight stuck after closing the toolbox (`646264e8`).** The `ensureHighlight` loop could re-add the highlight that `handleToolHiding` had just cleared. `ensureHighlight`/`reestablish` now bail when `!isShowing` (the reliable "tool active" signal — `getActiveToolId()` can still be "talkingBook" right after close).

**4 — Page-load flicker (`82bbc63`).** The page painted raw text first, then re-rendered with the highlight/tips after CKEditor finished — a visible "renders twice" flicker. `editMode.less` now keeps `.bloom-page` hidden until `bloomEditing.ts`'s `scheduleRevealPageWhenReady()` adds `bloom-page-ready` once all CKEditors are ready (2 s fallback so it can never stick hidden). Tradeoff: a brief (~0.8 s) blank during load instead of raw content, then the finished page. This one is a separate, independently-revertable commit.

Each fix has a regression test (verified fix-sensitive). A separate pre-existing regression found during testing (Talking Book highlighting dead after visiting a game page) is filed as BL-16602, not addressed here.

Ref: https://issues.bloomlibrary.org/youtrack/issue/BL-15300

[Devin review](https://devinreview.com/BloomBooks/BloomDesktop/pull/8100)

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8100)


src/BloomBrowserUI/bookEdit/js/bloomEditing.ts
JohnThomson, line 1: [Devin] **Bug**: Page can stay permanently blank if setup fails during load.

Devin flagged (and the local review independently agreed) that `scheduleRevealPageWhenReady()` — which arms the fallback timer that guarantees the page is revealed from behind editMode.less's hide-until-ready rule — ran *after* the page-setup work (`Cleanup`/`SetupElements`/`OneTimeSetup`) in `bootstrap()`. If any of that setup threw, the fallback would never be armed and the page would stay permanently hidden — worse than the pre-change degraded-but-visible page.

JohnThomson: [Claude Opus 4.8] Fixed in 83162e15e: moved `scheduleRevealPageWhenReady()` to the very start of `bootstrap()`, so the fallback reveal timer is armed before any setup that could throw. Resolving.

JohnThomson, line 1: [Devin] **Bug**: Page can stay blank for two seconds when a text editor finishes preparing too early.

The reveal readiness check tested `editorInstance.instanceReady`, which CKEditor never sets as a property (`instanceReady` is the event name). An editor that had already finished preparing when the deferred check ran was treated as still-pending, and the `instanceReady` listener added for it would never fire — so the page stayed hidden until the 2-second fallback timer: a visible ~2s blank in that timing case.

JohnThomson: [Claude Opus 4.8] Fixed in 32cc8601a: the fast-path now checks `editorInstance.status === "ready"` (CKEditor's actual readiness property), OR-ed with `instanceReady` defensively. Resolving.

src/BloomBrowserUI/bookEdit/css/editMode.less
JohnThomson, line 1: [Devin] **Investigate**: Off-screen edit page rendered as blank if screenshotted within the reveal window.

editMode.less now hides `.bloom-page` (visibility:hidden) until `bloom-page-ready` is added. Devin verified the off-screen page-capture path (`captureContentForExternalProcessing` / `__bloomEditablePageReady`, editablePage.ts:433-466) reads DOM/HTML content, **not** a rendered screenshot, so it is unaffected. The open question: if any *other* path takes an actual visual screenshot of the edit iframe (e.g. edit-view page-list thumbnails) while editMode.less is applied and before CKEditor readiness / the 2s fallback fires, it would now capture a blank page. Devin could not confirm whether the thumbnail-generation path uses editMode.less — worth a check by someone with that context. (Left open for the developer's decision.)

JohnThomson:
[Claude Opus 4.8] Consulted Devin on 2026-07-23 20:14 UTC up to commit 32cc8601a4. Across the review cycle Devin found 2 Bugs in the new page-reveal logic (both now fixed and resolved: fallback armed too late; wrong CKEditor readiness property) and 1 Investigate flag (possible blank thumbnail if the edit page-list screenshots during the reveal window — left open for the developer). Latest re-review: 0 Bugs. Informational items reviewed, not mirrored.

src/BloomBrowserUI/bookEdit/css/editMode.less
JohnThomson: [Claude Opus 4.8] Decision: leave as is. The developer confirmed this isn't a concern — the edit page-list thumbnails are not generated by screenshotting the live edit page during its load window, and the thumbnail DOM doesn't load editMode.css, so the `visibility:hidden` hide-until-ready rule never applies to them. Combined with Devin's verification that the main off-screen capture path reads HTML rather than a rendered screenshot, there's no path by which the hide rule produces a blank capture. No code change; resolving.

andrew-polk (changes requested):
> @andrew-polk reviewed 5 files and all commit messages, and made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8100)* status: all files reviewed, 5 unresolved discussions (waiting on JohnThomson).

___
*[`src/BloomBrowserUI/bookEdit/js/bloomEditing.ts` line 1199 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8100#-OyK81HQCCz6Ym6eb8Up:-OyK81HQCCz6Ym6eb8Uq:b28arfm) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/6fa71b8476106defb1e2cb9e8b211e6b7f24cbb3/src/BloomBrowserUI/bookEdit/js/bloomEditing.ts#L1199)):*
> ```XML
>                 pending--;
>                 if (pending === 0) {
>                     revealPage();
> ```

This seems like a fairly drastic change in page load behavior.
I guess you've tested quite a bit and decided it didn't degrade UX appreciably? Did you try throttling somehow?
___
*[`src/BloomBrowserUI/bookEdit/toolbox/talkingBook/audioRecording.ts` line 3784 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8100#-OyK6u7y0gh9NEHCIqfs:-OyK6u7y0gh9NEHCIqft:b-x8ghvk) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/6fa71b8476106defb1e2cb9e8b211e6b7f24cbb3/src/BloomBrowserUI/bookEdit/toolbox/talkingBook/audioRecording.ts#L3784)):*
> ```XML
>             contentWithoutTransientChildren
>                 .find(".bloom-ui-current-audio-marker")
>                 .remove();
> ```

Maybe I'm just not understanding something, but this feels pretty fragile to me. What happens when we add another transient element?

JohnThomson (commented):
> @JohnThomson+AGNT made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8100)* status: 1 of 16 files reviewed, 5 unresolved discussions (waiting on andrew-polk, font-face, and JohnThomson).

___
*[`src/BloomBrowserUI/bookEdit/js/bloomEditing.ts` line 1199 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8100#-OyK81HQCCz6Ym6eb8Up:-OyZJ3akEWc3w4ZeSW1l:bx6g8ao) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/6fa71b8476106defb1e2cb9e8b211e6b7f24cbb3/src/BloomBrowserUI/bookEdit/js/bloomEditing.ts#L1199)):*

This seems like a fairly drastic change in page load behavior.
I guess you've tested quite a bit and decided it didn't degrade UX appreciably? Did you try throttling somehow?
</blockquote></details>

[Claude Fable 5] You were right to question it, and we've backed it out entirely (commit 4ea49598f). Instrumenting the page load (CDP screencast + DOM-mutation/font-event capture) showed the flicker wasn't inherent to progressive rendering — it had specific causes we could fix at the source: duplicate @font-face declarations arriving mid-load re-blanked already-rendered text (fixed by single-sourcing them, 93afdc569); the audio highlight was cleared, then an HTTP check awaited, then re-set, painting a highlight-less gap on every setup round (fixed by making the swap atomic, 0a93e32c7); and DOM churn (CKEditor's attach-time paragraph replacement, our own always-re-append language reorder) silently killed the highlight's Ranges (fixed by a repair-before-paint MutationObserver and by skipping the no-op reorder). With those fixed, hiding the page until CKEditor was ready bought nothing and cost up to a second of blank page, so the delayed render is gone — pages paint progressively exactly as before this branch.
___
*[`src/BloomBrowserUI/bookEdit/toolbox/talkingBook/audioRecording.ts` line 3784 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8100#-OyK6u7y0gh9NEHCIqfs:-OyZJ4DA23v0sYKnPlB-:b-y7yl5l) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/6fa71b8476106defb1e2cb9e8b211e6b7f24cbb3/src/BloomBrowserUI/bookEdit/toolbox/talkingBook/audioRecording.ts#L3784)):*

Maybe I'm just not understanding something, but this feels pretty fragile to me. What happens when we add another transient element?
</blockquote></details>

[Claude Fable 5] Agreed — that was fragile, and it's fixed in commit 0a93e32c7. Both this comparison and the leaf-processing guard higher up now remove/skip anything carrying the generic `bloom-ui` class (the class that exists precisely to mark transient UI elements) instead of naming `#formatButton` and the audio marker individually. We verified both existing elements do carry the class (`StyleEditor.ts` creates the format button with `class="bloom-ui"`, and the current-audio marker is created with `bloom-ui-current-audio-marker bloom-ui`), so any future transient element that follows the convention is handled automatically.

src/BloomBrowserUI/bloomWebFonts.less
JohnThomson, line 1: [Devin] **Investigate**: UI @font-face declarations removed from bloomWebFonts.less; coverage now depends on three root stylesheets plus the runtime injector

Devin verified the edit-page frame is covered by editMode.css and that bloomDialog.css losing its @font-face copy is harmless (it was a duplicate contributing to the flicker). The residual risk it flags: any statically-linked (non-React) document that relied on a bloomUI.less-derived stylesheet for UI fonts but is not one of the three approved roots (editMode.less, toolbox.less, readerSetup.less). It notes the new uiFontFacesSingleSource.test.ts enforces single-sourcing but not coverage.

src/BloomBrowserUI/bloomMaterialUITheme.ts
```diff
@@ -1,5 +1,8 @@
 import { createTheme, Theme } from "@mui/material/styles";
```
JohnThomson, line 1: [Devin] **Investigate**: Removing font-face loading from the widely-imported MUI theme relies on all React mounts going through reactRender

bloomMaterialUITheme.ts no longer imports the font-face rules. Coverage for UI fonts now depends on the three document-root stylesheets and on ensureUiFontFacesInstalled() being invoked from getOrCreateRoot in reactRender.tsx for every React mount. Devin notes a few direct createRoot/.render sites outside reactRender.tsx (e.g. react_components/markdown.tsx, CanvasElementContextControls.tsx) that render into documents already carrying the faces, and calls this the assumption to verify if any UI text renders in a fallback font.

src/BloomBrowserUI/bookEdit/css/editMode.less
JohnThomson, line 1: [Devin] **Investigate**: Microphone icon marker may briefly stay hidden after page reveal

Devin analyzed the (earlier) hide-page-until-ready rule in editMode.less: while the page computed as hidden, updateIconMarker could hide the current-audio microphone icon, and nothing re-ran it at reveal time, so the icon could stay hidden briefly until a repeated newPageReady pass self-healed it.

src/BloomBrowserUI/bloomWebFonts.less
JohnThomson: [Claude Fable 5] Assessed — not an issue. We audited every non-React document: toolbox.pug and ReaderSetup.pug link toolbox.css / readerSetup.css (two of the three roots); the page-list, StyleEditor, and TextBoxProperties markup live inside documents that carry the faces (editMode.css); leveledRTInfo uses its own hand-written css that never derived from bloomUI.less, so it never had these @font-face rules to lose; the remaining .pug files are unit-test fixtures. Every other document is React-rooted and funnels through renderRoot(), where the guarded injector installs the faces before first render. A hypothetical missed document degrades to a fallback UI font (cosmetic, immediately visible in use), and the first React mount in it self-heals. Resolving.

src/BloomBrowserUI/bloomMaterialUITheme.ts
JohnThomson: [Claude Fable 5] Assessed — this is the designed behavior, and the risky edge Devin describes is exactly why the injector is guarded. Direct createRoot sites render into documents that already have the faces via their statically-linked css, and any future React mount anywhere goes through renderRoot(), which installs the faces idempotently before the first commit (and is a no-op where a static stylesheet already declares them — installing late duplicates was the original text-flicker bug, BL-15300). The uiFontFacesSingleSource test pins the delivery points so a new bundle cannot silently re-introduce duplicates. Resolving.

src/BloomBrowserUI/bookEdit/css/editMode.less
JohnThomson: [Claude Fable 5] Moot at this HEAD: the hide-page-until-ready mechanism this analyzes was removed entirely in 4ea49598f (the underlying flickers are now fixed at their sources, so the page paints progressively again and there is no hidden window for the icon to get stuck in). The icon is also now re-shown by the same repair path that re-registers the highlight after DOM changes. Resolving.

JohnThomson:
[Claude Fable 5] Consulted Devin on 2026-07-27 16:20 UTC up to commit a6d65e797d. Devin's re-review found 3 Bugs, all of which it marks resolved at this commit (the two page-reveal bugs are gone with the reveal architecture itself; the function-syntax style point was fixed in a6d65e797d), and 3 Investigate flags, each posted as an inline thread with our assessment and resolved (font-face coverage audit; renderRoot funnel assumption; icon-after-reveal — moot since the reveal was removed). 8 informational items reviewed, not mirrored. Greptile and CI are green for this commit.

src/BloomBrowserUI/bookEdit/toolbox/talkingBook/audioRecording.ts
JohnThomson, line 1: [Devin] **Investigate**: Early-return path in setHighlightToAsync can un-suppress a no-audio highlight

The unchanged-element early return calls refreshAudioTextHighlights(newElement) with no suppressCurrentHighlight argument. This PR adds currentHighlightSuppressed tracking, which that call resets to false, so the highlight is re-shown even if the same element previously had its highlight intentionally suppressed because it has no audio. Pre-existing shape, but the new suppression-tracking makes the inconsistency more visible: repair code trusts currentHighlightSuppressed, yet this path flips it. Worth confirming no caller re-invokes with the same element expecting the suppression to persist.

JohnThomson: [Claude Fable 5] Assessed — not an issue: suppression is a per-call state, not a sticky one, by design. Every call to setHighlightToAsync expresses the state the caller wants now; for identical arguments the early-return path produces exactly the same result as the full path (a non-suppressed set shows the highlight either way), so nothing becomes inconsistent. No caller sets a no-audio suppression and then re-invokes for the same element expecting it to persist — suppression happens in play flows, and when playback ends the highlight is deliberately re-established anyway. The suppression memory exists so the repair machinery (which acts between calls, not on behalf of one) will not resurrect a highlight the most recent call chose to hide — and this path correctly updates that memory to match the most recent call. Resolving.

JohnThomson:
[Claude Fable 5] Consulted Devin on 2026-07-27 17:08 UTC up to commit 2b3118fc92 (docs-only papercut-log commit). 0 unresolved bugs. One new Investigate flag (early-return path vs. no-audio highlight suppression) was mirrored, assessed as not-an-issue (suppression is per-call by design), and resolved; the three flags from the previous consultation are unchanged and remain resolved. 9 informational items reviewed, not mirrored.

andrew-polk (approved):
> @andrew-polk reviewed 16 files and all commit messages, and resolved 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8100)* status: all files reviewed, 7 unresolved discussions (waiting on font-face and JohnThomson).

PR #8101, opened 2026-07-23
Preserve image credits when editing with the AI Image Editor (BL-16603)

hatton:
Editing a book image with the AI Image Editor lost the image's credits (copyright / creator / license).

**Why:** a result comes back as a fresh `ai-image*.png` with no embedded (EXIF/XMP) metadata. The credits live on the DOM (`data-copyright` / `data-creator` / `data-license`), but Bloom rebuilds those attributes *from the file's embedded metadata* via `ImageUpdater.UpdateImgMetadataAttributesToMatchImage` (run on book load and data-div sync), so they were overwritten with empty strings on the next sync.

**Fix — the AI image editor decides, Bloom embeds.** Credits now travel in both directions over the host protocol:

- **Outbound:** each `bookImages` entry in the launch payload carries that image's current credits, read from its file metadata (`GetCreditsForImageFile`).
- **Inbound:** every commit replacement returns an explicit `credits` value, and `EmbedCreditsInNewImageFile` writes exactly that into the new file (before the current-page early-return, so it covers both the off-page book-DOM path and the live current-page path).

Bloom never guesses. When the AI image editor sends no credits — `null`, or an object whose every field is empty — that is its answer, and the generated file is left with none. It has to work that way: a user can open an existing illustration, generate something entirely new while in the tool, and click OK, and that new image is not entitled to the old illustration's credits. The editor is the only side that knows which happened. It already implements this contract (`bloom-ai-image-tools` c2cc9b9, shipped in the pinned `dist-v0.1.1`): a result inherits credits only when the tool actually edited the target image, and reference-only generations, text-to-image, and in-editor uploads never inherit them.

Returning without touching the file also means a user-uploaded result keeps its own embedded metadata, which is the only credit information anyone has for it. (That closes the Greptile P1 on this PR.)

Also in here: the current-page save fix (a replacement on the open page is now persisted, so a re-commit's `oldSrc` still matches the live page), the dev/alpha gate on the editor's Local Dummy model, and a `bloom-automation` recipe for driving the editor over CDP.

**Terminology:** inside Bloom, "editor" is ambiguous, because Bloom's own edit tab is an editor too. Comments and private member names in `AiImageEditorApi`, its tests, and `aiEditorLauncher` now say "AI image editor" throughout. Wire and environment names (`editorUrl`, `BLOOM_AI_EDITOR_URL`, the `aiImageEditor` routes, the `.ai-image-editor` folder) are unchanged.

Found while processing review comments on the merged PR #8033.

Ref: https://issues.bloomlibrary.org/youtrack/issue/BL-16603

[Devin review](https://devinreview.com/BloomBooks/BloomDesktop/pull/8101)

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8101)


src/BloomExe/web/controllers/AiImageEditorApi.cs
```diff
@@ -1039,5 +1049,77 @@ internal static HashSet<string> CollectReferencedImageFileNames(HtmlDom dom)
...
+                var credits = new Metadata
+                {
+                    Creator = source.Creator,
+                    License = source.License,
+                    CopyrightNotice = source.CopyrightNotice,
+                    AttributionUrl = source.AttributionUrl,
+                    CollectionName = source.CollectionName,
+                    CollectionUri = source.CollectionUri,
+                };
+                credits.WriteIntellectualPropertyOnly(newPath);
```
hatton: [Claude Opus 4.8] [Devin] **Investigate**: Metadata write may silently fail for webp results (gated, non-regression)

`CarryCreditsToNewImageFile` clears tags and calls `WriteIntellectualPropertyOnly` on the new file. The AI editor's editable formats include **webp** (`kAiEditableImageExtensions` / `AllowedImageExtensions`), and `RobustFileIO.MetadataFromFile` only special-cases png/jpg — webp goes straight to TagLib, which may not support writing metadata. If it throws, the catch here logs and returns, so a **webp** AI result would still come out with no credits.

Assessment: not a regression — before this change *every* format lost its credits on the next resync; this fixes png/jpg and degrades gracefully (log, no crash, no corruption) for webp. Whether it's worth also preserving webp credits is a product call — raised for the developer.

src/BloomExe/web/controllers/AiImageEditorApi.cs
```diff
@@ -918,6 +920,14 @@ out sourceBytesPath
             RobustFile.Copy(sourceBytesPath, Path.Combine(book.FolderPath, newFileName), true);
             newSrc = newFileName;
 
+            // A generated result carries no intellectual-property metadata of its own, so
+            // carry the replaced image's credits into the new file. Reused existing images
+            // (the sourceUrl branch) keep their own file metadata, so only do this for a
+            // freshly generated/uploaded result. See CarryCreditsToNewImageFile for why this
+            // has to be written into the FILE and not just the DOM attributes.
+            if (!string.IsNullOrEmpty(replacement.resultId))
+                CarryCreditsToNewImageFile(book.FolderPath, oldSrc, newFileName);
```
hatton: [Claude Opus 4.8] [Devin] **Investigate**: Current-page path relies on front-end changeImage preserving the file's new metadata

For the current (open) page, this writes the credits into the file and then returns early, leaving the front-end (`aiEditorLauncher.ts` -> `changeImageByElement`) to apply the replacement. The flag asks whether that front-end flow could drop the metadata just written into the file.

Assessment: no. `changeImageByElement` -> `changeImageInfo` only sets the element's `src` and `data-*` attributes in the DOM; it does not re-encode or rewrite the image file, so the bytes (and the metadata we wrote) stay on disk untouched. And even in the paths where Bloom does re-process an image, it does so via `ImageUtils.TrimMetadataInImage`, which explicitly re-writes the intellectual-property metadata — it preserves exactly the fields we set. Either way the file keeps its credits, and Bloom's later resync (`UpdateImgMetadataAttributesToMatchImage`) reads them back correctly.

hatton:
[Claude Opus 4.8] Consulted Devin on 2026-07-23 21:32 UTC up to commit `9cb92e5f77d563b6e15765a93acbfcc1bb04558d`.

Result: **0 bugs, 2 Investigate flags** (no Informational). Both on the new `CarryCreditsToNewImageFile`:
- webp results may still lose credits (gated, non-regression) — left **open** for a developer decision.
- current-page path relies on front-end changeImage preserving file metadata — assessed as a non-issue (changeImage doesn't rewrite the file; re-processing preserves IP metadata) and **resolved**.

CI/other bots tracked separately in this preflight run.

src/BloomExe/web/controllers/AiImageEditorApi.cs
hatton: [Claude Opus 4.8] Assessed — not blocking, and not a regression. Before this change AI-edited images *always* lost their credits; now the common raster formats are handled — I verified end-to-end (dummy-model edit) that a committed PNG result keeps its embedded creator/copyright/license. The write is wrapped so any TagLib limitation for a given format is logged, not fatal. webp IP-metadata support in TagLib is a pre-existing, orthogonal limitation, so this doesn't regress anything. Resolving.

src/BloomExe/web/controllers/AiImageEditorApi.cs
```diff
@@ -918,6 +965,21 @@ out sourceBytesPath
...
+            // Reused existing images (the sourceUrl branch) keep their own file metadata, so we
+            // only do this for a freshly generated/uploaded result. See CarryCreditsToNewImageFile
+            // for why this has to be written into the FILE and not just the DOM attributes.
+            if (!string.IsNullOrEmpty(replacement.resultId))
+                CarryCreditsToNewImageFile(
+                    book.FolderPath,
+                    oldSrc,
+                    newFileName,
+                    replacement.credits
+                );
```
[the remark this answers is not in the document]
hatton: [Claude Opus 4.8] Real edge, thanks. For an *uploaded* result with no editor-supplied credits, the null-fallback copies the replaced slot's metadata over the upload. The right behavior depends on what the editor sends as credits for an uploaded image (a cross-repo, credit-provenance decision), so rather than change credit behavior unilaterally I've routed this to the preflight decision report for the developer to decide.

hatton:
[Claude Opus 4.8] Consulted Devin on 2026-07-24 01:1x UTC up to commit `0f17c3dcb5`.

**1 Bug** — "docs describe the wrong build gate": this was correct on the prior commit and is **fixed in `0f17c3dcb5`** (the docs + driver comment now say `IsDevOrAlpha` / dev+alpha). Devin's re-review still quotes the superseded text, so it's a false positive against the current HEAD.

**3 Investigate flags** (all assessed, none blocking):
- *webp metadata may silently fail* — replied & resolved: not a regression, logged-not-fatal, PNG/JPG verified end-to-end.
- *current-page relies on changeImage preserving file metadata* — verified end-to-end via a dummy-model edit (committed file kept its embedded credits); thread resolved.
- *empty editor credits don't fall back to the replaced file* — intentional per design (editor owns the credit decision) and covered by `CarryCreditsToNewImageFile_EmptyEditorCredits_LeavesGeneratedFileClean`.

**6 Informational** — not mirrored (low signal).

Greptile (`Greptile Review` check ✓) raised one P1 ("uploaded credits still overwritten") now routed to the developer. CI checks green.

src/BloomBrowserUI/bookEdit/toolbox/canvas/aiEditorLauncher.ts
```diff
@@ -355,6 +394,21 @@ export const launchAiImageEditor = (
                                 if (finalOk) {
```
hatton: [Claude Opus 5 from Hatton's machine during preflight] [Devin] **Bug**: AI image editor window can become unclosable when a commit only partly succeeds

When a commit partly succeeds — say one picture on another page couldn't be replaced, while a picture on the page you're looking at was — this `finally` block does two things: it leaves the AI editor panel on screen (so you can see the error), and it asks Bloom to save and re-think the current page.

Those two don't mix. The editor panel is attached to Bloom's top-level window, but all the code that runs it — the ✕ button's handler, and the listener that receives messages from the editor — belongs to the page iframe. Saving and re-thinking the page reloads that iframe, which throws away the code while leaving the panel itself on screen. The result is a full-screen panel that no longer responds to its own close button, so the user has to restart Bloom.

Before this branch no save was issued here, so a partial failure left a panel that still worked.

This has been routed to the developer as a decision, because the fix involves a real trade-off: closing the panel before the reload is the smallest change but throws away the error message the user was meant to read, whereas keeping the panel usable across the reload means moving its ownership out of the page iframe, which is a bigger structural change.

src/BloomExe/web/controllers/AiImageEditorApi.cs
```diff
@@ -1039,5 +1157,221 @@ internal static HashSet<string> CollectReferencedImageFileNames(HtmlDom dom)
...
+            };
+        }
+
+        /// <summary>
+        /// Collapses a ClearShare license to the single wire string the AI image editor
+        /// carries: the license URL for a Creative Commons license (its canonical form),
+        /// otherwise the free-text rights statement for a custom license. Null when there is
+        /// no license.
+        /// </summary>
+        private static string LicenseToWireString(LicenseInfo license)
```
hatton: [Claude Opus 5 from Hatton's machine during preflight] [Devin] **Bug**: Extra license notes on a picture are dropped when it is edited with the AI Image Editor

Bloom lets a picture carry both a license *type* (say CC-BY) and free-text **license notes** — wording the user typed about how the picture may be used. When we hand a picture's credits to the AI editor, `LicenseToWireString` collapses the whole license down to a single string, and for any license that has a URL it sends only the URL. The notes never travel. On the way back, `BuildLicense` rebuilds a bare CC license from that URL, so the new picture's file is written without them.

What the user sees: they typed notes about how an illustration may be used, ran an AI edit on it, and the notes are gone.

There is a second variant. A picture whose license is "contact the copyright holder" (a `NullLicense`) plus notes has no URL, so the notes text itself becomes the wire string. Coming back, that text isn't a creativecommons.org URL, so it is rebuilt as a *custom* license — quietly changing the license type.

This has been routed to the developer as a decision, because the fix is a change to the wire format Bloom and the editor share: the rights statement needs its own field (and probably a license token alongside the URL), which means a matching change in the `bloom-ai-image-tools` repo. That is a cross-repo contract call, not something to change unilaterally from this side.

.github/skills/bloom-automation/driveAiImageEditor.mjs
```diff
@@ -0,0 +1,236 @@
...
+    const f = contentFrame(page);
+    return f.evaluate(() =>
+        Array.from(
+            document.querySelectorAll('img, [style*="background-image"]'),
+        ).map((el) => {
+            const isImg = el.tagName === "IMG";
+            const bg = getComputedStyle(el).backgroundImage;
+            // An <img> carries its url in src; a background-image element has it inside a
+            // css url(...) wrapper we have to unwrap.
+            let src;
```
hatton, line 94: [Claude Opus 5 from Hatton's machine during preflight] [Devin] **Bug**: New automation script uses stacked ternary operators, which the repo style rules forbid

The image-dump helper in this test-automation script chose its source url with two chained conditionals (`isImg ? ... : bg && bg !== "none" ? ... : ""`). `AGENTS.md` (Code Style) asks explicitly for an if/else-if chain instead: "Avoid stacking/nesting ternary (`? :`) operators … They're too hard for humans to read."

hatton: [Claude Opus 5 from Hatton's machine during preflight] Fixed in `69dc139` — it's now a plain if/else-if chain, with a comment saying why the two cases differ (an `` keeps its url in `src`, a background-image element wraps it in `url(...)`). Resolving.

src/BloomExe/web/controllers/AiImageEditorApi.cs
```diff
@@ -1039,5 +1157,221 @@ internal static HashSet<string> CollectReferencedImageFileNames(HtmlDom dom)
...
+        )
+        {
+            if (string.IsNullOrEmpty(relativePath))
+                return null;
+            // A book's image src can still be percent-encoded, so decode our way to the real
+            // file the way the rest of Bloom does (BL-3901) instead of a plain Path.Combine.
+            // Getting this wrong doesn't just miss a file: we'd report credits:null, the AI
+            // image editor would believe the image has none, and a result derived from it would
+            // legitimately get none — losing exactly the credits this code exists to carry.
+            var path = UrlPathString.GetFullyDecodedPath(bookFolderPath, ref relativePath);
```
hatton, line 1189: [Claude Opus 5 from Hatton's machine during preflight] [Devin] **Investigate**: Outbound credits lookup doesn't use Bloom's double-encoding-tolerant path resolution

`GetCreditsForImageFile` found the image with a plain `Path.Combine`, but a book's image `src` can still be percent-encoded — which is why Bloom's own `ImageUpdater.UpdateImgMetadataAttributesToMatchImage` goes through `UrlPathString.GetFullyDecodedPath` (BL-3901).

For a picture whose filename contains anything that gets encoded (a space, an accent), the lookup missed, so the launch payload told the AI editor `credits: null`. The editor then believed that picture had no credits, and any result derived from it correctly got none — losing exactly the credits this branch exists to carry, just narrowed to those filenames.

hatton: [Claude Opus 5 from Hatton's machine during preflight] Good catch, and a real one — fixed in `69dc139`. It now decodes the same way the rest of Bloom does, so this side and `ImageUpdater` agree on which file a given `src` means. Added a test that pins it: a picture written to disk as `my pic.png`, looked up as `my%20pic.png`, must still come back with its creator and copyright. The test asserts up front that the encoded form does *not* exist on disk, so it can't pass by accident. Resolving.

src/BloomExe/web/controllers/AiImageEditorApi.cs
```diff
@@ -918,17 +1011,42 @@ out sourceBytesPath
...
+            // describe the OLD file, so they have to be re-derived from the new one; carrying
+            // them forward would leave the page claiming credits the new file doesn't have
+            // (BL-16603).
             if (isCurrentPage)
             {
                 // Leave the live (current) page to the front-end: it will call Bloom's
-                // changeImage() with newSrc so the canvas + normal save flow handle it.
+                // changeImage() with newSrc and these attributes so the canvas + normal save
+                // flow handle it.
+                creditAttributes = ReadCreditAttributes(book.FolderPath, newFileName);
```
hatton, line 1044: [Claude Opus 5 from Hatton's machine during preflight] [Devin] **Investigate**: Live-page DOM credit attributes still come from the replaced image, not the credits now embedded in the new file

The flag reads: for a picture on the page currently open, Bloom writes the chosen credits into the new image file, but the front-end then set the element's `data-creator` / `data-copyright` / `data-license` from the *old* element's attributes — so the saved page markup disagrees with the file until a later pass re-derives it. It notes the same asymmetry off-page, where the src is repointed but the credit attributes are left alone, and asks whether anything could over-claim the previous illustration's copyright in the meantime.

hatton: [Claude Opus 5 from Hatton's machine during preflight] Exactly right, and this is the bug the newest commit on this branch fixes — Devin's own bug list for this same commit now reports it as resolved ("On the page being edited, a newly generated picture keeps the previous picture's credits on screen"). The flag's line references point at the pre-fix code.

Both halves of the asymmetry are closed now:

- **current page** — the commit reply carries `copyright`/`creator`/`license` read back off the new file (`ReadCreditAttributes`), and the front-end applies those instead of reading the element it is replacing.
- **off page** — after the src is repointed, we call Bloom's own `ImageUpdater.UpdateImgMetadataAttributesToMatchImage`, so the attributes are re-derived from the new file right there. That runs *before* the data-div harvest and save, so a cover/xmatter slot's fresh attributes get carried into the data-div rather than reverted.

On the "could anything over-claim the old copyright" question: there is no window any more — the attributes are written from the new file during the same commit, not left for a later pass. A test pins the two sides together (`ReadCreditAttributes_MatchesWhatBloomsOwnUpdaterWouldWrite`) so they can't drift apart, and I confirmed it in a running Bloom both ways: an edit of a credit-less picture now shows the "missing information" indicator, and an edit of a credited one keeps its credits in the DOM, the file, and the next launch payload alike. Resolving.

src/BloomExe/web/controllers/AiImageEditorApi.cs
```diff
@@ -1039,5 +1157,221 @@ internal static HashSet<string> CollectReferencedImageFileNames(HtmlDom dom)
...
+        ///
+        /// Internal for testing.
+        /// </summary>
+        internal static void EmbedCreditsInNewImageFile(
+            string bookFolderPath,
+            string newFileName,
+            ImageCredits credits
+        )
+        {
+            if (credits == null || CreditsAreEmpty(credits))
```
hatton, line 1295: [Claude Opus 5 from Hatton's machine during preflight] [Devin] **Investigate**: Empty editor credits intentionally do not fall back to the replaced file

When the AI editor sends a `credits` object whose every field is empty, we write nothing into the new picture's file — we deliberately do *not* fall back to the credits of the picture being replaced. Devin notes this is intentional per the design (the editor owns the credit decision; an empty object means "this result has no credits") and is covered by a test, and asks reviewers to confirm the editor really does echo back the credits it was handed at launch — because a bug on that side would silently drop credits.

hatton: [Claude Opus 5 from Hatton's machine during preflight] Confirmed against a running Bloom, which is the part that needed checking. I ran a dummy-model edit on a picture that has real credits (Pratham Books / Angie and Upesh / CC-BY) and the editor did echo them back: the committed result's file came out with those credits embedded, the page showed them, and the next launch payload read them back off the new file. So the editor is not sending an empty object where it should be sending the source's credits.

The other direction checks out too: an edit whose source genuinely had no credits produces a result with none, and the page then shows the "missing information" indicator instead of silently claiming the old picture's copyright — which is the bug this branch exists to fix.

One stale detail in the flag: `CarryCreditsToNewImageFile` and the test it names were renamed earlier on this branch and are now `EmbedCreditsInNewImageFile` and `EmbedCreditsInNewImageFile_EmptyCredits_LeavesFileClean`. The behavior it describes is unchanged. Resolving.

hatton:
[Claude Opus 5 from Hatton's machine during preflight] Consulted Devin on 2026-07-30 23:2x UTC up to commit `69dc1398da`.

Devin ran twice this preflight — once on `120bb21f67` and again after the fixes, on `69dc1398da`. On the final commit it reports 6 bugs (4 of which it now considers fixed), 6 Investigate flags, and 11 Informational items (not mirrored — low signal).

Two things it raised are fixed in `69dc1398da`, each with its own thread:

- A picture whose filename needs percent-encoding had its credits reported as absent at launch, so an AI edit of it correctly produced a result with no credits — the very loss this branch prevents, narrowed to those filenames. Now decoded the way the rest of Bloom does, with a test.
- A stacked ternary in the new test-automation helper, which `AGENTS.md` asks us not to write.

Three Investigate flags were assessed as non-issues and their threads replied to and resolved: the live-page credit attributes (fixed by this branch — Devin's own bug list agrees), the deliberate no-fallback for empty editor credits (confirmed against a running Bloom that the editor really does echo the source's credits back), and the encoding one above. Two more — the webp metadata limitation and whether the current-page `changeImage` flow preserves the file's metadata — were already assessed and documented on resolved threads earlier in this PR, so they weren't duplicated.

**Two findings are left open for the developer**, both genuine and both needing a call I shouldn't make alone:

- [The editor panel can become unclosable after a partly-failed commit](https://github.com/BloomBooks/BloomDesktop/pull/8101#discussion_r3686856702) — a real stuck state introduced by this branch, but the fix is a trade-off between losing the error message and moving the panel's ownership out of the page iframe. Devin also raised this as a separate Investigate flag; the thread covers both.
- [License notes are dropped when a picture is AI-edited](https://github.com/BloomBooks/BloomDesktop/pull/8101#discussion_r3686858204) — fixing it changes the wire format shared with `bloom-ai-image-tools`, so it's a cross-repo contract decision.

CI is green (`pr-automation`, `track`). Greptile stayed quiet on both commits.

src/BloomBrowserUI/bookEdit/toolbox/canvas/aiEditorLauncher.ts
hatton: [Claude Opus 5 following a prompt from Hatton] Fixed in `3d0ce96584`.

This was routed to the developer as a decision because of the trade-off you named. The answer was the third option you didn't list, and it's the one that keeps both properties: **hold the save until the panel comes down**, rather than closing the panel early or relocating its ownership.

The `finally` block no longer saves. It records that a current-page swap is pending, and the save is issued from `cleanup()` instead. On a fully successful commit that is immediately — the panel closes, then the save fires, exactly the old behaviour. On a partial failure the panel stays up with the error readable, and the save waits for the user to dismiss it. Either way the iframe reload happens only after nothing depends on the iframe's code any more.

Your analysis of the cause was right and worth restating: the overlay div lives in the top-level window, but its ✕ handler, its message listener, and `cleanup()` itself are all page-iframe code, so reloading the iframe leaves an overlay nothing can close.

`aiEditorLauncher.test.ts` is new and pins all three paths — partial failure holds the save until ✕ is clicked, full success closes and saves, and a commit that changed nothing on this page never saves at all (a needless page reload would discard unsaved text edits).

src/BloomExe/web/controllers/AiImageEditorApi.cs
hatton: [Claude Opus 5 following a prompt from Hatton] Fixed in `3d0ce96584`, in both repos.

The cause was that the credits we handed the editor flattened the license into one string, which for a Creative Commons license is just the URL. ClearShare lets an image carry a CC license *and* free-text license notes spelling out extra restrictions, so the notes had nowhere to ride, never came back, and the edited copy lost them.

The wire now carries `licenseUrl` and `licenseRightsStatement` as separate fields, which is what an image file actually stores — `cc:license` plus `dc:rights`.

`BuildLicense` deliberately mirrors ClearShare's own `LicenseUtils.FromXmp`, the code that reads a license back out of a file: a CC URL becomes a CC license; failing that, a rights statement becomes a `CustomLicense`; nothing at all becomes a `NullLicense`. Matching `FromXmp` is what makes the round trip faithful, and it's also why no license-*type* token rides on the wire — a token could only express licenses a file cannot store. An "ask the copyright holder" license that also has notes reads back from any image file as a custom license, so that isn't something this code can or should preserve. A test pins that reasoning and will fail if ClearShare ever starts distinguishing more license types in the file.

The `bloom-ai-image-tools` side is types and fixtures only (commit `2aa74aa`, on that repo's master): the editor carries the credits object verbatim and serializes it whole, so the new fields already survive the published `dist-v0.1.1` build and no dependency bump was needed here.

andrew-polk (approved):
> @andrew-polk reviewed 13 files and all commit messages, and resolved 8 discussions.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8101)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on hatton).

PR #8103, opened 2026-07-23
Add visual-regression tests to the nightly workflow

andrew-polk:
Follow-up to #8098 (the nightly workflow, now merged): adds the
`src/BloomVisualRegressionTests` suite as a step in `nightly.yml`.

**What it does**
- After the front-end and C# builds, launches the Release `Bloom.exe`
  (`output/Release/x64`) over HTTP and compares rendered screenshots against the
  committed reference images (pixelmatch). Setup installs the suite's own pnpm deps
  and the Playwright chromium it uses to capture bloom-player pages.

**Gating — both builds must succeed first**
- The step runs only when **both** the front-end build (`build_frontend`) and the C#
  build (`build_csharp`) succeeded: Bloom loads its UI from the built browser assets the
  C# build copies in, and the suite launches the built `Bloom.exe`.
- Uses `!cancelled()` so it still runs (and reports visual breakage) even when the unit
  test steps above failed — consistent with the workflow's health-check philosophy.

**On failure**
- The generated `*-current.png` / `*-diff.png` images are uploaded as a
  `visual-regression-diffs` artifact for inspection (the `*-reference.png` baselines are
  already committed).

The suite's `pnpm-lock.yaml` is added to the setup-node cache key.

No YouTrack ticket (nightly/CI tooling).

[Devin review](https://devinreview.com/BloomBooks/BloomDesktop/pull/8103)

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8103)

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -177,10 +177,39 @@ jobs:
               shell: bash
               run: pnpm test
 
-            # On failure, keep the current + diff images so a human can see what changed. The
-            # committed *-reference.png baselines are already in the repo; *-current.png and
-            # *-diff.png are generated (gitignored). if-no-files-found: ignore avoids a warning
-            # when the failure was not a pixel mismatch (e.g. Bloom failed to launch).
+            # When the visual-regression step fails, the most useful clue is usually Bloom's own
+            # log (it launches a real Bloom.exe and talks to it over HTTP). Bloom writes to
+            # %TEMP%\SIL\Bloom\Log.txt (and, depending on how far it got, %LOCALAPPDATA%\SIL\Bloom);
+            # a crash on startup often leaves a .NET Watson/CLR entry in the Windows Application
+            # event log. Gather whatever exists into a fixed folder so the next step can upload it,
+            # rather than losing the one thing that explains a "Bloom never came up" failure.
+            # Best-effort throughout: never let missing logs turn this diagnostic step red.
+            - name: Collect Bloom diagnostics
+              if: ${{ !cancelled() && steps.visual_regression.outcome == 'failure' }}
+              shell: bash
+              run: |
+                  set +e
+                  dest="bloom-diagnostics"
+                  mkdir -p "$dest"
+                  for src in "$TEMP/SIL/Bloom" "$LOCALAPPDATA/SIL/Bloom" "$LOCALAPPDATA/Temp/SIL/Bloom"; do
+                      if [ -d "$src" ]; then
+                          echo "Found Bloom logs in: $src"
+                          # Flatten the source root into the name so two Log.txt don't collide.
+                          tag=$(echo "$src" | tr '/:\\ ' '____')
+                          cp -r "$src" "$dest/$tag" 2>/dev/null
+                      fi
+                  done
+                  # .NET runtime crashes (e.g. a missing WebView2 runtime) surface here even when
+                  # Bloom wrote no log of its own.
+                  powershell -NoProfile -Command "Get-WinEvent -LogName Application -MaxEvents 40 -ErrorAction SilentlyContinue | Where-Object { \$_.ProviderName -match 'Application Error|.NET Runtime|Windows Error Reporting' } | Format-List TimeCreated, ProviderName, Message" > "$dest/windows-application-events.txt" 2>&1
+                  echo "----- collected diagnostics -----"
+                  ls -R "$dest" || true
+
+            # On failure, keep the current + diff images so a human can see what changed, plus the
+            # Bloom logs / crash events gathered above. The committed *-reference.png baselines are
+            # already in the repo; *-current.png and *-diff.png are generated (gitignored).
+            # if-no-files-found: ignore avoids a warning when the failure was not a pixel mismatch
+            # (e.g. Bloom failed to launch, so no screenshots were produced).
             - name: Upload visual regression diffs
               if: ${{ !cancelled() && steps.visual_regression.outcome == 'failure' }}
               uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
@@ -189,6 +218,7 @@ jobs:
                   path: |
                       src/BloomVisualRegressionTests/collections/**/screenshots/*-current.png
                       src/BloomVisualRegressionTests/collections/**/screenshots/*-diff.png
+                      bloom-diagnostics/**
                   retention-days: 14
                   if-no-files-found: ignore
 
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -62,6 +62,7 @@ jobs:
                   cache-dependency-path: |
                       src/BloomBrowserUI/pnpm-lock.yaml
                       src/content/pnpm-lock.yaml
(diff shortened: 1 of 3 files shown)


.github/workflows/nightly.yml
```diff
@@ -143,6 +144,46 @@ jobs:
...
+                  pnpm install --frozen-lockfile
+                  # The suite drives a headless chromium (Playwright) to capture bloom-player pages.
+                  pnpm exec playwright install chromium
+
+            - name: Run visual regression tests
+              id: visual_regression
+              if: ${{ !cancelled() && steps.build_frontend.outcome == 'success' && steps.build_csharp.outcome == 'success' }}
+              working-directory: src/BloomVisualRegressionTests
+              shell: bash
+              run: pnpm test
```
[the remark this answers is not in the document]
andrew-polk: [Greptile reply] ✅ Fixed in `9fadb58`. Good catch. The run step now chains on the setup step: `if: ${{ !cancelled() && steps.setup_visual_regression.outcome == 'success' }}` (the setup step got an `id`), mirroring the C# step chain. A failed `pnpm install` / `playwright install` now cleanly skips the run step instead of triggering a confusing secondary failure, while the failed setup step still turns the job red on its own.

— posted by Claude (Opus 4.8) via preflight

.github/workflows/nightly.yml
```diff
@@ -143,6 +144,50 @@ jobs:
...
+                  pnpm install --frozen-lockfile
+                  # The suite drives a headless chromium (Playwright) to capture bloom-player pages.
+                  pnpm exec playwright install chromium
+
+            - name: Run visual regression tests
+              id: visual_regression
+              if: ${{ !cancelled() && steps.setup_visual_regression.outcome == 'success' }}
+              working-directory: src/BloomVisualRegressionTests
+              shell: bash
+              run: pnpm test
```
andrew-polk, line 178: [Devin] **Investigate**: The nightly floats bloom-player@alpha, which can drift these reference images

The pre-existing "Update floating front-end dependencies" step pulls `bloom-player@alpha` (and other `@latest` deps) every night. This visual-regression suite compares renders — which include bloom-player output — against static, committed `*-reference.png` baselines. So any legitimate visual change in a new bloom-player alpha will make the visual-regression step fail nightly until someone regenerates the reference images.

That may well be by design (the nightly's whole purpose is to surface drift), but it's worth confirming the team expects to periodically re-baseline the references rather than read every alpha-driven diff as a Bloom regression. Left open for a developer decision (see the preflight report).

— posted by Claude (Opus 4.8) via preflight

.github/workflows/nightly.yml
```diff
@@ -143,6 +144,50 @@ jobs:
...
+            # committed reference images (pixelmatch). It needs BOTH builds: the C# build (the
+            # Bloom.exe it launches) AND the front-end build (Bloom loads its UI from the built
+            # browser assets that the C# build copies into its output). So both must have
+            # succeeded — build_csharp already implies build_frontend (the C# chain is gated on
+            # it), but we require both explicitly for clarity. !cancelled() lets these run even
+            # when the unit-test steps above failed, so the nightly still reports visual breakage.
+            # The run step then chains on this setup step's success (like the C# chain), so a
+            # failed install/browser-download cleanly skips the run rather than triggering a
+            # confusing secondary failure — the failed setup step still turns the job red.
+            - name: Set up visual regression tests
```
andrew-polk, line 163: [Devin] **Investigate**: The 120-minute job timeout may be tight now that the visual-regression suite is added

`timeout-minutes: 120` was not raised in this PR, but the added suite launches a real Bloom.exe and captures many screenshots across books × brandings × themes (each test has a 120s timeout). On top of the front-end + C# builds and both existing test suites, total runtime could approach or exceed 120 minutes on a loaded runner and be killed as a spurious timeout.

No hard data on the suite's duration yet, so this is a judgment call: bump the timeout preemptively for headroom, or watch the first few nightly runs and raise it if needed. Left open for a developer decision (see the preflight report).

— posted by Claude (Opus 4.8) via preflight

andrew-polk:
Consulted Devin on 2026-07-23 up to commit `9fadb58151a04809be19ab0285cde11eaa6795fa`.

- **0 Bugs.**
- **2 Investigate** — both posted as open threads for a developer decision (see preflight report):
  - Nightly floats `bloom-player@alpha`, so the committed VR reference images can drift (#discussion_r3642050416).
  - The 120-min job timeout may be tight now the VR suite is added (#discussion_r3642050495).
- **1 Informational** — "VR run step doesn't depend on its own setup step" — already fixed in this same commit (`9fadb58`), which gated the run step on `setup_visual_regression`; not mirrored.

Greptile: its earlier P2 ("test bypasses failed setup") was fixed in `9fadb58` and its thread resolved; the re-review is clean (no new inline findings). CodeRabbit: inactive on this repo. CI (pr-automation, track): passing.

— posted by Claude (Opus 4.8) via preflight

.github/workflows/nightly.yml
```diff
@@ -152,8 +201,11 @@ jobs:
               uses: EnricoMi/publish-unit-test-result-action/windows@d0a4676d0e0b938bc201470d88276b7c74c712b3 # v2.24.0
               with:
                   check_name: Nightly test results
-                  junit_files: output/Tests/vitest-junit.xml
-                  nunit_files: output/Tests/Release/x64/TestResults.xml
+                  # Unified files input (junit_files/nunit_files are deprecated); the action
+                  # auto-detects each file's format (JUnit for vitest, NUnit3 for the C# suite).
+                  files: |
```
andrew-polk, line 243: [Devin] **Investigate**: unified `files` input verified against the `/windows` action variant

Confirmed: the `files` input exists on `EnricoMi/publish-unit-test-result-action/windows` at the pinned SHA `d0a4676` (v2.24.0) — it's declared in that variant's `action.yml` (line 48), alongside the now-deprecated `junit_files`/`nunit_files`. The deprecation warning we're responding to ("Use 'files' option instead") was itself emitted by this `/windows` variant, so `files` is the supported replacement, and it auto-detects each file's XML dialect (JUnit for the vitest report, NUnit3 for the C# report). Resolving.

— posted by Claude (Opus 4.8) via preflight

andrew-polk:
Consulted Devin on 2026-07-23 up to commit `4da81fc0b152fb85573147ed2e2bbb704a42454f` (after the test-report fix).

- **0 Bugs.**
- **3 Investigate:**
  - "Verify unified `files` input against the `/windows` action variant" — verified (`files` is declared in that variant's action.yml at the pinned SHA; the deprecation we're fixing came from that variant). Documented + resolved (#discussion above on the `files:` line).
  - "Nightly floats bloom-player@alpha → reference-image drift" — unchanged open decision (#discussion_r3642050416).
  - "120-min job timeout may be tight" — unchanged open decision (#discussion_r3642050495).
- **4 Informational** (not mirrored): three verify the change is correct — the `--outputFile=` form is right, the VR artifact glob matches only generated images, and the Bloom.exe discovery order works for a clean Release checkout; the fourth (run-step-vs-setup) was already fixed in `9fadb58`.

Greptile: re-review clean (its only inline finding, the earlier P2, remains fixed + resolved). CodeRabbit: inactive. CI (pr-automation, track): passing.

— posted by Claude (Opus 4.8) via preflight

PR #8104, opened 2026-07-24
Fix Format dialog controls silently dying after a style-switch error (BL-16589)

JohnThomson:
Fixes the reported "word spacing control in the Format dialog stops affecting the document" problem (and a related Highlighting-tab color bug found along the way).

**Root cause:** `UpdateControlsToReflectAppliedStyle` (runs when the user picks a different style in the Format dialog's Style dropdown) set `ignoreControlChanges = true`, refreshed all the controls, and only reset the flag at the end — with no `try/finally`. If anything threw during the refresh, the flag stayed stuck `true`, and from then on every control change handler (`changeWordSpace`, `changeSize`, …) silently returned without touching the document, for the rest of the page's life. Closing and reopening the dialog didn't help, because `runFormatDialog` never reset the flag.

**Changes**

- Wrap the control refresh in `try/finally` so `ignoreControlChanges` is always restored (the original exception still propagates).
- Reset the flag whenever the Format dialog opens, so a freshly opened dialog can never start dead.
- Fix an argument-order bug in the same refresh: `changeHiliteProps(current.color, current.hiliteBgColor)` passed the ordinary text color as the *highlight* text color, so after a style switch the Highlighting tab displayed a bogus highlight text color — and a subsequent background-color change would write it into the book. Now called as `(current.hiliteTextColor, current.hiliteBgColor, current.color)`.
- Two new unit tests: one proves the controls stay alive after a mid-refresh exception (failed before the fix), one proves the highlight controls are re-rendered with the style's real highlight colors (failed before the fix).

Ref: https://issues.bloomlibrary.org/youtrack/issue/BL-16589

[Devin review](https://devinreview.com/BloomBooks/BloomDesktop/pull/8104)

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8104)

### DIFF
--- a/src/BloomBrowserUI/bookEdit/StyleEditor/StyleEditor.ts
+++ b/src/BloomBrowserUI/bookEdit/StyleEditor/StyleEditor.ts
@@ -2210,37 +2210,48 @@ export default class StyleEditor {
 
     public UpdateControlsToReflectAppliedStyle(oldFontName: string) {
         const current = this.getFormatValues();
+        // While we push the new style's values into the controls, their change handlers
+        // must not treat those updates as user edits. The finally is essential: if any
+        // step here throws, leaving ignoreControlChanges stuck true would silently
+        // disable every control in the dialog from then on (the document would stop
+        // responding to them) for the rest of the page's life.
         this.ignoreControlChanges = true;
-
-        // IF the new style changed fonts, we need to reset the font control
-        if (oldFontName !== current.fontName) {
-            get("fonts/metadata", (result) => {
-                const fontMetadata: IFontMetaData[] = result.data;
-                this.updateFontControl(fontMetadata, current.fontName);
-            });
+        try {
+            // IF the new style changed fonts, we need to reset the font control
+            if (oldFontName !== current.fontName) {
+                get("fonts/metadata", (result) => {
+                    const fontMetadata: IFontMetaData[] = result.data;
+                    this.updateFontControl(fontMetadata, current.fontName);
+                });
+            }
+            this.setValueAndUpdateSelect2Control("size-select", current.ptSize);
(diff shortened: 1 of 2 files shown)


JohnThomson:
[Claude Fable 5] Consulted Devin on 2026-07-24 up to commit f929648f132d31c4fb1a2283722f7d9093d2afe2. Result: 0 bugs, 0 Investigate flags, 1 Informational note (the async font-metadata callback in UpdateControlsToReflectAppliedStyle runs after the guard flag is reset — pre-existing behavior, unchanged by this PR, and Devin itself notes it is not a regression). Nothing to mirror as review threads.

hatton:
@JohnThomson this file (what was there, and now continued by this PR) seems so... defensive, so... repair-y to me. As opposed to just having invariants that are true each time the dialog is open. But that may not really be fair, but can you make a judgment on that? Thanks.

JohnThomson (commented):
I think it feels that way because we're now so used to React. There ought to be controls with state that gets initialized from the original state of the DOM and pushes changed state back to the DOM at the same time it changes the controls. This is some of our earliest and ugliest non-React code. For non-React, this seems like a fairly reasonable attempt to keep everything consistent: when we change style, every control potentially changes state, and while we do that, we want to disable all the change handlers that take effect when the user changes individual controls. I don't see an obvious way to make it cleaner that would be less work than doing the full React port. Maybe it's time to do that...

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8104#-:-OyeiqyIDJIsg5amHVi_:b-888i7x)* status: 0 of 2 files reviewed, all discussions resolved (waiting on JohnThomson).

PR #8105, opened 2026-07-24
Restore helpful error message when pasting an unusable image (BL-16594)

JohnThomson:
## Problem

Pasting an unusable image (e.g. a corrupt image file copied in Windows Explorer) showed the unhelpful generic message "Before you can paste an image, copy one onto your 'clipboard', from another program" instead of the more helpful "Bloom failed to interpret the clipboard contents as an image. Possibly it was a damaged file, or too large. Try copying something else."

## Root cause

The C# side already produced the helpful message (verified that `PalasoImage.FromFileRobustly` throws for a corrupt file with the exact libpalaso package Bloom uses), but `RequestInfo.WriteError` sent it only in the HTTP status description with an **empty response body**. The front end reads `error.response.data`, found nothing, and always fell back to the generic message. This regressed when paste errors moved from a WinForms MessageBox to the browser (cee437d356).

## Changes

- `RequestInfo.WriteError` now also writes the message into the response body as UTF-8 (the status description is ASCII-sanitized and not exposed by axios). `PretendRequestInfo` mirrors this.
- When the clipboard yields nothing loadable at all, `EditingView` now distinguishes contents that were probably *meant* to be an image — a copied file with an image extension, or text that looks like a path/URL to an image file (including Explorer's quoted "Copy as path" output) — and shows the "failed to interpret" message for those; genuinely non-image-like contents (plain text) keep the generic message.
- End-to-end test through a real BloomServer asserting the message survives into the 400 response body (fails without the fix), plus unit tests for the image-intent classification.

No new localizable strings (existing XLF IDs reused); no front-end changes needed.

Ref: https://issues.bloomlibrary.org/youtrack/issue/BL-16594

**Verified manually** by the developer: copying the corrupt image file in Windows Explorer and pasting via the image toolbar in a Bloom running this branch now shows the "Bloom failed to interpret the clipboard contents as an image…" message.

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8105)

[Devin review](https://devinreview.com/BloomBooks/BloomDesktop/pull/8105)

### DIFF
--- a/src/BloomExe/Edit/EditingView.cs
+++ b/src/BloomExe/Edit/EditingView.cs
@@ -1,5 +1,6 @@
 //#define MEMORYCHECK
 using System;
+using System.Collections.Generic;
 using System.Diagnostics;
 using System.Drawing;
 using System.Drawing.Imaging;
@@ -695,6 +696,20 @@ namespace Bloom.Edit
 
                     if (clipboardImage == null)
                     {
+                        // The clipboard held nothing we could even attempt to load as an image.
+                        // But if what's there looks like it was *meant* to be one (e.g. a copied
+                        // image file whose path no longer exists, or text that looks like a path
+                        // or URL to an image file), the "failed to interpret" message is more
+                        // helpful than telling the user to go copy an image.
+                        if (ClipboardContentsSuggestImage())
+                        {
+                            throw new InvalidOperationException(
+                                LocalizationManager.GetString(
+                                    "EditTab.NoValidImageFoundOnClipboard",
+                                    "Bloom failed to interpret the clipboard contents as an image. Possibly it was a damaged file, or too large. Try copying something else."
+                                )
+                            );
+                        }
                         throw new InvalidOperationException(
                             LocalizationManager.GetString(
                                 "EditTab.NoImageFoundOnClipboard",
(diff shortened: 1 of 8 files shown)


src/BloomExe/web/RequestInfo.cs
```diff
@@ -443,7 +443,17 @@ public void WriteError(int errorCode, string errorDescription)
...
+            // sanitized to ASCII (mangling localized text) and many HTTP clients don't expose
+            // it at all; the body is what client code (e.g. axios error.response.data) actually
+            // reads, and it can carry the full UTF-8 message.
+            // A HEAD response must not have a body (see WriteOutput and ReplyWithFileContent);
+            // http.sys throws ProtocolViolationException if we try, which would leave the
+            // client hanging without a response.
+            if (_actualContext.Request.HttpMethod == "HEAD")
+                _actualContext.Response.Close();
+            else
+                _actualContext.Response.Close(Encoding.UTF8.GetBytes(errorDescription), false);
```
JohnThomson, line 456: [Devin] **Investigate**: WriteError now sends a body for all error responses, including 404s and .json requests

This change makes `WriteError` write the message into the response body for every error, not just paste-image errors. Two downstream implications worth noting: (1) `WriteError(int)` now returns a "File not found" body for all 404s (e.g. missing image/file requests from the browser), where previously the body was empty. This is benign for browsers but is a behavior change for all callers of the shared method. (2) For requests where `LocalPathWithoutQuery` ends in `.json`, the ContentType is set to `application/json` (per the BL-7900 comment) yet the body is now plain text, which is not valid JSON. Axios's default `transformResponse` silently catches parse failures and leaves the data as a string, so this is unlikely to break existing consumers, but any client that strictly parses the error body as JSON would now see a non-JSON payload. Worth a quick confirmation that no `.json` endpoint relies on an empty/parseable error body.

_(mirrored from Devin's review by Claude Fable 5)_

JohnThomson: [Claude Fable 5] Assessed — not an issue, this behavior change is the point of the PR. The empty error body was the bug: the browser reads the body (axios `error.response.data`) and fell back to a generic message whenever it was empty. On the two specifics: (1) a "File not found" body on 404s is harmless — no client could have depended on parsing the previously-empty body; (2) for `.json` requests, a strict JSON parser would have failed on the old empty body just as it would on plain text now, and axios (what our front end uses) leaves unparseable bodies as strings. HEAD responses deliberately skip the body (fixed in 1d11e2d56c) since http.sys forbids it. Resolving.

src/BloomExe/Edit/EditingView.cs
```diff
@@ -823,6 +838,74 @@ private static PalasoImage GetImageFromClipboard()
...
+            return HasProbableImageExtension(clipboardText);
+        }
+
+        private static bool HasProbableImageExtension(string path)
+        {
+            if (string.IsNullOrWhiteSpace(path))
+                return false;
+            // Explorer's "Copy as path" wraps the path in quotes; ignore them, along with
+            // stray whitespace, so such a path is still recognized as image-intended.
+            return _probableImageExtensions.Contains(Path.GetExtension(path.Trim().Trim('"')));
```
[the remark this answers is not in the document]
JohnThomson: [Claude Fable 5] Fixed in 8ed526846c: for http(s) URLs the query string/fragment is now stripped before the extension check, so `.../bird.png?width=800` is classified as image-intended. Stripping is URL-only since '#' is a legal character in Windows file names. Added test cases for both.

JohnThomson:
[Claude Fable 5] Consulted Devin on 2026-07-24 13:33 up to commit 8ed526846c. Three review rounds ran (one per pushed commit). Final round: 0 bugs, 1 Investigate flag — mirrored as an inline thread, assessed as intentional (the empty error body was the bug this PR fixes), and resolved — plus 4 informational notes (coarse text heuristic, platform/threading observations already guarded in code; not mirrored, low signal). Greptile's one P1 (image URLs with query strings) was fixed in 8ed526846c. CI green.

JohnThomson:
[Claude Fable 5] Consulted Devin again after merging origin/master, up to commit 9874a975a1 (final HEAD): 0 bugs, no new findings — the single Investigate flag is the already-assessed-and-resolved WriteError one. Greptile and CI also green for this commit. Bots quiet.

JohnThomson:
[Claude Fable 5] Consulted Devin on 2026-07-24 15:07 up to commit b83d6c55d2 (history rewritten to add BL-16542 to commit titles; tree identical to the previously reviewed code). Result: 0 bugs, no new findings — the one Investigate flag is the already-assessed-and-resolved WriteError body-change note. Greptile and CI green. Bots quiet.

JohnThomson:
[Claude Fable 5] Note: the final force-push (9adda43946) only corrected the ticket number in commit titles (BL-16542 → BL-16594); the tree is identical to b83d6c55d2, which Devin reviewed clean. Per the developer, not waiting for Devin's re-review of the retitled commit.

StephenMcConnel (approved):
> @StephenMcConnel reviewed 8 files and all commit messages, and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8105)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on JohnThomson).

PR #8102, opened 2026-07-23
Paste replaces the selected background image instead of adding an overlay (BL-16542)

StephenMcConnel:
When the cover page is in **Standard Layout** and the image (the `bloom-canvas` background image) is the selected element, pressing **Ctrl+V** — or the top-bar **Paste** button — pasted an *overlay on top of* the image instead of *replacing* it.

## Cause

`CanvasElementClipboard.finishPasteImageFromClipboard` (the Ctrl+V / top-bar-Paste path, via the `kMakeNewCanvasElement` sentinel) excluded the background image from its "replace the selected image" branch, and additionally only replaced a selected image when it still held a placeholder. So pasting onto a selected, already-filled image (the Standard Layout cover's background) fell through to the "create a new overlay" branch.

## Fix

In `finishPasteImageFromClipboard`, when an image canvas element is the active element, replace its image **in place** — whether it is the background image or an overlay, and whether or not it currently holds a placeholder. This matches the behavior of the image context-menu's **Paste** command and the user's expectation that pasting onto a selected image replaces it.

- Background image → `adjustBackgroundImageSize`; overlay → `adjustContainerAspectRatio` + `adjustTarget` (unchanged from the prior overlay path).
- Non-image selected elements (text/video) and the empty-canvas → set-background case are unchanged.

## Testing

- Added an e2e regression test to `bookEdit/canvas-e2e-tests/specs/08-clipboard-and-paste.spec.ts` (prefers an existing background image; the assertion fails if the fix is reverted, and skips gracefully when the host clipboard can't deliver an image).
- Verified in the running app (CDP): with the Standard Layout cover's background image selected, Ctrl+V replaced the image in place (canvas-element count stayed 1) rather than adding an overlay.

Ref: https://issues.bloomlibrary.org/youtrack/issue/BL-16542

— Claude Opus 4.8 (1M context)

[Devin review](https://devinreview.com/BloomBooks/BloomDesktop/pull/8102)

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8102)

PR #8106, opened 2026-07-24
Refresh the image copyright/metadata button on the Ctrl+V paste path (BL-16605)

andrew-polk:
Ctrl+V / top-bar **Paste** did not refresh a background image's copyright/metadata button after replacing the image, so the button kept reflecting the previous (placeholder) state until the page was reloaded.

## Cause

`changeImage()` in `bloomEditing.ts` normally routes through `CanvasElementManager.updateCanvasElementForChangedImage()`, which rebuilds the button via `SetupMetadataButton()`. But it returns early when `imageId === kMakeNewCanvasElement` — the clipboard path — handing off to `CanvasElementClipboard.finishPasteImageFromClipboard()`, which set the image and fixed the geometry itself and never refreshed the button.

## Fix

Extracted `CanvasElementClipboard.replaceImageInCanvasElement()`, shared by both branches that paste into an image already on the page, and had it call `SetupMetadataButton()` for background images. Only background images ever get a metadata button (see `SetupMetadataButton`), so the overlay branch deliberately does not call it.

Apart from the added `SetupMetadataButton()` call, behavior is unchanged: each branch still makes exactly the same geometry/notification calls as before.

Two deliberate deviations from the approach suggested on the card:

- `SetupMetadataButton` is **imported directly** from `../bloomImages` rather than added to `ICanvasElementClipboardHost`. `CanvasElementClipboard` already imports from that module, and sibling `CanvasElementBackgroundImageManager` imports it the same way, so there was no reason for interface churn.
- **No call on the overlay path**, since it would be a no-op.

## Note for #8102 / BL-16542 (still open)

That PR widens the second branch so it can replace a *background* image too, and its version also omits the metadata refresh. The helper added here is deliberately the shape that branch converges to, so the merge will conflict in this region and the resolution should collapse #8102's branch into `replaceImageInCanvasElement()` rather than losing the fix.

## Testing

New `CanvasElementClipboard.test.ts` (3 cases): background paste rebuilds the button; the button is rebuilt only *after* the new `src`/`data-copyright` are in place; the overlay path does not touch it.

Both a unit-level and an in-app negative control confirm the tests and the fix actually track the bug:

- Unit: commenting out the new call made the two background tests fail while the overlay test still passed.
- In the running app, same clipboard image and same code path, differing only by that one line:

| | fix off | fix on |
|---|---|---|
| before | `placeHolder.png`, no button | `placeHolder.png`, no button |
| after Ctrl+V | `image1.png`, **no button** | `image2.png`, **button present** |

Ref: https://issues.bloomlibrary.org/youtrack/issue/BL-16605

[Devin review](https://devinreview.com/BloomBooks/BloomDesktop/pull/8106)

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8106)

### DIFF
--- a/src/BloomBrowserUI/bookEdit/js/canvasElementManager/CanvasElementClipboard.test.ts
+++ b/src/BloomBrowserUI/bookEdit/js/canvasElementManager/CanvasElementClipboard.test.ts
@@ -126,7 +126,11 @@ function makeClipboard(host: ICanvasElementClipboardHost) {
 describe("CanvasElementClipboard paste refreshes the metadata button (BL-16605)", () => {
     beforeEach(() => {
         document.body.innerHTML = "";
-        vi.mocked(SetupMetadataButton).mockClear();
+        // mockReset (not mockClear) so an implementation installed by one test cannot leak
+        // into the next. Safe here because the mock factory gives SetupMetadataButton no
+        // implementation of its own.
+        vi.mocked(SetupMetadataButton).mockReset();
+        // changeImageInfo must keep the implementation from its factory, so only clear calls.
         vi.mocked(changeImageInfo).mockClear();
     });
 
@@ -172,7 +176,6 @@ describe("CanvasElementClipboard paste refreshes the metadata button (BL-16605)"
 
         expect(srcWhenButtonBuilt).toBe("pasted.png");
         expect(copyrightWhenButtonBuilt).toBe(pastedImageInfo.copyright);
-        vi.mocked(SetupMetadataButton).mockReset();
     });
 
     test("pasting into a selected overlay does not touch the metadata button", () => {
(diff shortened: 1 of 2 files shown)


andrew-polk:
[Claude Opus 5] Consulted Devin on 2026-07-24 20:02 UTC up to commit `88eaf6338fd37b99dedddc9ecd89b291bf3c1610`.

Devin's review completed for that commit and raised **no bugs and nothing to investigate**. It did return one *informational* observation, which was a confirmation rather than a concern: it independently checked that the extracted `replaceImageInCanvasElement` helper is equivalent to the two original inline branches apart from the intended `SetupMetadataButton` call — same call ordering, background call site verified to be the background element and the overlay call site verified not to be — and concluded "no behavioral regression identified". Informational items aren't mirrored as review threads, so nothing was posted.

Greptile also reviewed this commit and had no findings, and all CI checks are green.

StephenMcConnel (approved):
> @StephenMcConnel reviewed 2 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8106)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on andrew-polk).

PR #8108, opened 2026-07-24
Added automatic rtl setting based on script

josiahwall:
This pr makes it so that the right to left setting for a collection's writing system/language is set automatically based on the script that was chosen for that writing system/language. This also includes automatically setting the checked value of the "This script is written right to left like Arabic, Hebrew, and Urdu." checkbox based on the selected script whenever one makes a new collection.

[Devin review](https://devinreview.com/BloomBooks/BloomDesktop/pull/8108)

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8108)

### DIFF
--- a/src/BloomBrowserUI/collection/LanguageChooserDialog.tsx
+++ b/src/BloomBrowserUI/collection/LanguageChooserDialog.tsx
@@ -33,6 +33,7 @@ export interface ILanguageData {
     LanguageTag: string | null;
     DefaultName: string | null;
     DesiredName: string | null;
+    IsRtl: boolean | null;
     Country?: string | null;
 }
 
@@ -46,11 +47,13 @@ export function getLanguageData(
         (selection?.language
             ? defaultDisplayName(selection.language) || null
             : null);
+    const isRtl = selection?.script?.isRtl;
     // Ensure values are null rather than undefined. Otherwise, the property won't be serialized at all.
     return {
         LanguageTag: languageTag || null,
         DefaultName: defaultName,
         DesiredName: selection?.customDetails?.customDisplayName || defaultName,
+        IsRtl: isRtl !== undefined ? isRtl : null,
         Country: languageTag
             ? defaultRegionForLangTag(languageTag, selection?.language)?.name ||
               null
--- a/src/BloomBrowserUI/package.json
+++ b/src/BloomBrowserUI/package.json
@@ -128,7 +128,7 @@
         "@emotion/core": "11.0.0",
         "@emotion/react": "11.14.0",
         "@emotion/styled": "11.14.1",
-        "@ethnolib/language-chooser-react-mui": "0.3.2",
+        "@ethnolib/language-chooser-react-mui": "0.4.0",
         "@fortawesome/fontawesome-svg-core": "6.4.0",
         "@fortawesome/free-solid-svg-icons": "6.4.0",
         "@mui/icons-material": "5.11.0",
--- a/src/BloomBrowserUI/pnpm-lock.yaml
+++ b/src/BloomBrowserUI/pnpm-lock.yaml
@@ -34,8 +34,8 @@ importers:
                 specifier: 11.14.1
                 version: 11.14.1(@emotion/react@11.14.0(@types/react@18.3.31)(react@18.3.1))(@types/react@18.3.31)(react@18.3.1)
             "@ethnolib/language-chooser-react-mui":
-                specifier: 0.3.2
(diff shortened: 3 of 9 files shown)


andrew-polk (changes requested):
> @andrew-polk reviewed 9 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8108)* status: all files reviewed, 1 unresolved discussion (waiting on josiahwall).

___
*[`src/BloomBrowserUI/collection/LanguageChooserDialog.tsx` line 56 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8108#-OyZK-S7Clxq1Qe-hCCb:-OyZK-S7Clxq1Qe-hCCc:bmjrddy) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/426502ea2c864f87d5909042e9c96dcc1ce9a529/src/BloomBrowserUI/collection/LanguageChooserDialog.tsx#L56)):*
> ```XML
>         DefaultName: defaultName,
>         DesiredName: selection?.customDetails?.customDisplayName || defaultName,
>         IsRtl: isRtl !== undefined ? isRtl : false,
> ```

Sorry I didn't think about this before, but I think we better handle the case where we are unsure about the setting of right to left. In other words, `IsRtl` should probably be an optional/nullable Boolean so that Ethnolib is able to say yes, no, or unknown. In the unknown case we wouldn't change anything on the Bloom side. 

You would have to plumb that 3-state boolean all the way through to the back end.

Whether we change the ethnolib side to actually distinguish can be a separate question. I'm thinking there are definitely theoretical cases, but I don't know if there are real users cases where we are unsure.

But if I'm not thinking clearly, please push back.

josiahwall (commented):
> @josiahwall made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8108)* status: 5 of 9 files reviewed, 1 unresolved discussion (waiting on andrew-polk).

___
*[`src/BloomBrowserUI/collection/LanguageChooserDialog.tsx` line 56 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8108#-OyZK-S7Clxq1Qe-hCCb:-OyZW4iT1WQZf9KRTqzD:b-n0j7lt) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/426502ea2c864f87d5909042e9c96dcc1ce9a529/src/BloomBrowserUI/collection/LanguageChooserDialog.tsx#L56)):*

Sorry I didn't think about this before, but I think we better handle the case where we are unsure about the setting of right to left. In other words, `IsRtl` should probably be an optional/nullable Boolean so that Ethnolib is able to say yes, no, or unknown. In the unknown case we wouldn't change anything on the Bloom side. 

You would have to plumb that 3-state boolean all the way through to the back end.

Whether we change the ethnolib side to actually distinguish can be a separate question. I'm thinking there are definitely theoretical cases, but I don't know if there are real users cases where we are unsure.

But if I'm not thinking clearly, please push back.
</blockquote></details>

Done. After I made the changes to allow for the unknown case, I ran the changes through Devin and got this flag "RTL only overridden when script directionality is known; stale value may persist". It's telling me that if a user were to select an RTL script and then change the language to a script with an unknown isRtl value, then the RTL value for the writing system would remain true. Is that fine? Right now, it seems like, on the Ethnolib side, no script keeps their isRtl attribute undefined, so right now the flag is not a potential issue.

andrew-polk (approved):
> @andrew-polk reviewed 4 files and all commit messages, made 1 comment, and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8108)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on josiahwall).

___
*[`src/BloomBrowserUI/collection/LanguageChooserDialog.tsx` line 56 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8108#-OyZK-S7Clxq1Qe-hCCb:-OyZZyO56cHujfJaJjhc:bkfg3qd) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/426502ea2c864f87d5909042e9c96dcc1ce9a529/src/BloomBrowserUI/collection/LanguageChooserDialog.tsx#L56)):*
<details><summary><i>Previously, josiahwall wrote…</i></summary><blockquote>

Done. After I made the changes to allow for the unknown case, I ran the changes through Devin and got this flag "RTL only overridden when script directionality is known; stale value may persist". It's telling me that if a user were to select an RTL script and then change the language to a script with an unknown isRtl value, then the RTL value for the writing system would remain true. Is that fine? Right now, it seems like, on the Ethnolib side, no script keeps their isRtl attribute undefined, so right now the flag is not a potential issue.
</blockquote></details>

Yes, I think that's the behavior we want. Basically, no matter what the setting in Bloom is, if we get undefined from ethnolib, we don't want to make changes on the Bloom side. Thanks.

PR #8109, opened 2026-07-27
Clean up the toolbox infrastructure now that all tools are React (BL-16608)

andrew-polk:
## Problem

All eleven toolbox tools have been React for some time, but the Edit tab's toolbox *frame* still ran the legacy jQuery-UI accordion in parallel: `toolbox.ts` built every tool's panel into a hidden `#toolbox` div, and the visible React toolbox (`ToolboxRoot.tsx`) adopted those DOM nodes back out through a bridge — `window.toolboxReactAdapter`, CustomEvents, a MutationObserver, and a 250 ms polling loop. The double bookkeeping was dead weight and a steady source of subtle bugs: the Talking Book busy cursor targeted the hidden copy and never appeared, and an id-spelling mismatch silently broke image-description detection (the BL-8515 protection).

## What the PR does

- **One React root renders the toolbox.** A small state store (`toolboxState.ts`) is the single source of truth for which tools are offered, which section is expanded, and which tool is running; each tool is an ordinary child component whose lifecycle a React effect runs (`useToolLifecycle.ts`). The accordion, the hidden div, and the whole bridge are deleted — about 2,100 lines net; `toolbox.ts` goes from 1,948 to under 900 lines.
- **`toolIds.ts` owns the canonical tool ids** plus the shims for the persisted spellings, so the saved book format is deliberately byte-for-byte unchanged (`active\tcanvasCheck\t1` still lands as `{"name": "canvas", "enabled": true}`, `current\tcanvasTool` as `currentTool: canvasTool`).
- **The per-keystroke markup engine** (reader-tool highlighting, paste cleanup, CKEditor bookmark coordination) moved verbatim from `toolbox.ts` to `pageEditingMarkup.ts` — it was page-editing code, not toolbox code.
- **Tools deliver their own stylesheets**, and the Playwright harness/uitests were rewritten to drive the real tools.
- **Fixed along the way:** the never-appearing Talking Book busy cursor; the image-description id mismatch; a Decodable Reader crash when a book has no saved toolbox settings; and the toolbox continuing to drive a tool whose collapsed section stopped applying to the page.

## Verification

Typecheck, eslint (0 errors), full Vitest suite (824 passed / 5 skipped), full C# suite (3,446 passed), the 9 toolbox Playwright uitests, and a production bundle build all pass at HEAD. Live smoke testing in a running Bloom covered tool switching, enabling/disabling tools from More…, reader-tool highlighting while typing, and undo/redo restoring highlights (the one behavior the code documents as untestable in jsdom).

Ref: https://issues.bloomlibrary.org/youtrack/issue/BL-16608

---
[Devin review](https://app.devin.ai/review/BloomBooks/BloomDesktop/pull/8109)

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8109)

[Devin review](https://devinreview.com/BloomBooks/BloomDesktop/pull/8109)


src/BloomBrowserUI/bookEdit/toolbox/ToolboxRoot.tsx
```diff
@@ -353,434 +138,103 @@ const LiveToolBodyHost: React.FunctionComponent<{ element: HTMLDivElement }> = (
...
+    );
 
-            return changed ? nextSections : previousSections;
+    const makeToolActive = React.useCallback((toolId: string) => {
+        setExpandedSectionId(toolId);
+        activeToolChangedCallbacks.current.forEach((callback) => {
+            callback(toolId);
         });
-    }, [sections]);
+    }, []);
```
andrew-polk: [Devin] **Investigate**: Active-tool callbacks now receive canonical ids instead of suffixed names

In the old `ToolboxRoot`, the active-tool notification path converted the section id with `toToolboxToolId(...)` before invoking `activeToolChangedCallbacks`, so listeners registered via `onActiveToolChanged` received suffixed names like `canvasTool`. The new `makeToolActive` (and the `onChange` handler) invokes callbacks with the raw canonical `section.id` (e.g. `canvas`). This is a behavioral contract change for whatever `toolbox.ts` registers via `onActiveToolChanged`. `toolbox.ts` is not part of this diff, so I could not verify it was updated to expect canonical ids. If any listener still assumes the suffixed spelling (e.g. to look up/persist the current tool), clicking a section header would set the wrong current tool.

src/BloomBrowserUI/bookEdit/toolbox/ToolboxRoot.tsx
```diff
@@ -79,93 +64,37 @@ const toolboxHeaderIconStyles = css`
...
+// toolbox knows about: toolbox.ts only asks us for tools it found in the master list.
 const makeSectionFromToolId = (toolId: string): ToolboxSection => {
-    const labelInfo = getToolboxLabelInfo(toolId);
+    const tool = getMasterToolList().find(
+        (candidate) => candidate.id() === toolId,
+    )!;
+    const labelInfo = getToolLabelInfo(toolId);
+    const toolBodyElement = tool.makeRootElement();
+    // Some tool stylesheets still select their body by this attribute, using the
+    // historical "Tool"-suffixed name.
```
andrew-polk: [Devin] **Investigate**: addTool assumes the tool is in the master list (non-null assertion)

`makeSectionFromToolId` uses `getMasterToolList().find(c => c.id() === toolId)!`. If `addTool` is ever called with an id not in the master list (e.g. a stale/renamed persisted id from meta.json, or a canonicalization mismatch between `toolbox.ts` and `toolIds.ts`), `tool` is `undefined` and the next line `tool.makeRootElement()` throws, which would surface as a broken toolbox rather than a gracefully-skipped tool. The old code path filtered enabled ids against the master list before building sections. Whether this is safe depends entirely on `toolbox.ts` (not in the diff) only ever passing canonical ids that exist in the master list.

src/BloomExe/Edit/ToolboxView.cs
```diff
@@ -81,9 +92,6 @@ public static IEnumerable<string> GetToolboxServerDirectories()
             yield return BloomFileLocator.GetBrowserDirectory(
                 "bookEdit/toolbox/readers/readerSetup"
             );
-            yield return BloomFileLocator.GetBrowserDirectory("bookEdit/toolbox/signLanguage");
-            yield return BloomFileLocator.GetBrowserDirectory("bookEdit/toolbox/imageDescription");
-            yield return BloomFileLocator.GetBrowserDirectory("bookEdit/toolbox/canvas");
         }
```
andrew-polk, line 102: [Devin] **Investigate**: C# GetToolboxServerDirectories pruning relies on removed tools' CSS being bundled, not served by bare name

`GetToolboxServerDirectories` removes the `signLanguage`, `imageDescription`, and `canvas` folders from the factory search path, with a new comment asserting the search path only matters for bare-filename stylesheet lookups (`BloomServer.ProcessCssFile`) and that assets fetched by full path don't consult it. This holds only because those three tools' `.less` files are imported into their `.tsx` (bundled by Vite) and are not referenced as separate `<link>`/bare-name `.css` (unlike `talkingBook`/`motion`/`music`, which remain in the list and are `<link>`ed in toolbox.pug). If any code path still requests e.g. `canvasTool.css`/`signLanguage.css`/`imageDescription.css` by bare filename, it would now fail to resolve.

src/BloomBrowserUI/bookEdit/toolbox/ToolboxRoot.tsx
andrew-polk: Claude Fable 5: Not an issue — verified in the code Devin could not see (`toolbox.ts` was not in its diff view).

The one listener registered through `onActiveToolChanged` is `switchTool`, and `switchTool` is written to take a **canonical** id: it looks the tool up with `masterToolList.find(tool => tool.id() === requestedToolId)` (an exact match against canonical ids) and converts to the historical spelling only when persisting, via `"current\t" + toPersistedToolName(requestedToolId)`. So the conversion did not disappear in this PR; it moved from ToolboxRoot to the one place that needs it.

Confirmed end-to-end on disk during live testing: clicking the Canvas section header posted `current<TAB>canvasTool` and the book's meta.json got `currentTool: canvasTool`; navigating to a game page produced `currentTool: gameTool`. Both are the suffixed spellings Bloom has always written, so the "wrong current tool" failure mode does not occur.

src/BloomBrowserUI/bookEdit/toolbox/ToolboxRoot.tsx
andrew-polk: Claude Fable 5: Not an issue — the guard Devin was looking for does exist, in `toolbox.ts` (which its diff view did not include).

`addTool` has exactly one caller, `beginAddTool` in toolbox.ts. That function first resolves the tool and bails out if it cannot:

```ts
const tool = getITool(toolId);
if (!tool) {
    console.error(`Tool ${toolId} not found, assuming that was from a different version of Bloom.`);
    return;
}
...
adapter?.addTool(tool.id());
```

So the argument is never a caller-supplied string — it is `tool.id()` read off an `ITool` that was just found in the master list. A stale or renamed id from meta.json (exactly the scenario Devin describes) hits that early return and is logged and skipped, which is the graceful behavior it was hoping for. The `!` therefore cannot be `undefined`.

src/BloomExe/Edit/ToolboxView.cs
andrew-polk: Claude Fable 5: Not an issue — checked the exact condition Devin names.

Nothing anywhere in `src/` asks for those three stylesheets by name. Grepping every `.pug`, `.ts`, `.tsx`, `.cs`, `.html`, and `.less` file for `canvasTool.css`, `signLanguage.css`, `imageDescription.css` (and plain `canvas.css`) returns no hits at all, so there is no bare-filename request left to resolve.

And the reason there is none: each of those tools imports its stylesheet from its component, so Vite bundles it —

- `canvas/CanvasToolControls.tsx`: `import "./canvasTool.less";`
- `imageDescription/imageDescription.tsx`: `import "./imageDescription.less";`
- `signLanguage/signLanguageTool.tsx`: `import "./signLanguage.less";`

That is precisely what distinguishes them from `talkingBook`/`motion`/`music`, whose `.css` files *are* `<link>`ed from toolbox.pug and whose directories this PR deliberately keeps in the list. Their remaining assets are `.svg` icons fetched by full path, which `ProcessImageFileRequest` resolves against the browser root without consulting the search path.

src/BloomBrowserUI/bookEdit/toolbox/ToolboxRoot.tsx
```diff
@@ -353,434 +139,103 @@ const LiveToolBodyHost: React.FunctionComponent<{ element: HTMLDivElement }> = (
...
-        window.toolboxReactAdapter = {
-            isEnabled: () => true,
-            setActiveToolByToolId: (toolId: string) => {
-                const normalizedToolId = normalizeToolId(toolId);
-                setExpandedSectionId(normalizedToolId);
-                const callbackToolId = toToolboxToolId(normalizedToolId);
-                activeToolChangedCallbacks.current.forEach((callback) => {
-                    callback(callbackToolId);
-                });
             },
```
andrew-polk: [Devin] **Investigate**: Removing the expanded section on tool removal does not notify toolbox.ts

When `removeTool` removes the currently-expanded section (e.g. the Game tool when navigating away from a `data-tool-id="game"` page), it updates `expandedSectionId` directly via `setExpandedSectionId` rather than through `makeToolActive`, so `activeToolChangedCallbacks` are NOT fired. Consequently `toolbox.ts` is never told the active tool changed, and its module-level `currentTool` continues to point at the removed tool (which is no longer `isToolInitialized`). This matches the pre-PR `onToolRemoved` behavior (which also did not fire callbacks), and the PR description says game add/remove was verified in live testing, so it is likely benign in practice; noted for follow-up since the state divergence is subtle.

andrew-polk: Claude Fable 5: Confirmed as described, but it is **pre-existing behavior, not a change in this PR** — so deliberately left alone here.

Devin's own hedge is right. Master's `onToolRemoved` did the same thing:

```ts
setExpandedSectionId((previousExpandedSectionId) =>
    previousExpandedSectionId === removedToolId
        ? (firstSection ? firstSection.id : undefined)
        : previousExpandedSectionId,
);
```

— a direct state update, with no callback fired. And on the `toolbox.ts` side, master's removal branch of `showOrHideTool` only removed the DOM nodes, dispatched the `toolbox-tool-removed` event, and called `resizeToolbox()`; it never touched `currentTool` either. So the divergence Devin describes (`currentTool` still pointing at a tool the toolbox no longer offers) is exactly as old as the React toolbox, and this PR neither introduces nor widens it.

Live testing of the game-page flow (navigate onto a `data-tool-id="game"` page, off it, and back) behaved correctly with no console errors, which is consistent with it being benign in practice.

Since this PR is a cleanup that is meant to preserve behavior, changing the removal path to fire the active-tool callbacks would be a behavior change beyond its scope. I am flagging it to the developer as a candidate follow-up card rather than fixing it here. Resolving this thread so the record shows the outcome.

andrew-polk:
Claude Fable 5: Consulted Devin on 2026-07-27 05:31 UTC up to commit `8df32913ae89ccce8476485253261802b460bc28`.

Devin found **no bugs**. It raised 4 Investigate flags and 4 Informational observations.

Each Investigate flag now has its own resolved thread above with the reasoning. In short, three of them were questions Devin could not answer because `toolbox.ts` was outside its diff view, and the answers are all in that file: the active-tool callback contract is canonical-by-design (`switchTool` converts to the persisted spelling itself), `addTool`'s non-null assertion is safe because its only caller passes `tool.id()` from an already-resolved tool, and no code requests the three pruned tools' stylesheets by bare filename (their `.less` is bundled via `import`). The fourth — that removing the expanded section doesn't fire the active-tool callbacks — is real but is byte-for-byte the behavior on master, so it's noted as a possible follow-up rather than changed in a cleanup PR.

The Informational items were assessed and need no action; both of the substantive ones (the reader-tools readiness guard being realm-dependent, and tool bodies being built eagerly at startup) are equivalent to what master already did.

CI is green and Greptile's review of this commit is clean.

src/BloomBrowserUI/bookEdit/toolbox/ToolboxRoot.tsx
```diff
@@ -80,732 +74,250 @@ const toolboxHeaderIconStyles = css`
...
-                if (
-                    canRehydrateFromLegacySource &&
-                    hasDisconnectedLiveToolBodyElement &&
-                    hydratedToolIds.current.has(section.id)
-                ) {
-                    hydratedToolIds.current.delete(section.id);
+            `}
+            disableGutters
+            expanded={props.isExpanded}
+            onChange={(_event, expanded) => {
```
andrew-polk, line 158: [Devin] **Bug**: Clicking the open tool's title bar now leaves the sidebar with nothing open

(`src/BloomBrowserUI/bookEdit/toolbox/ToolboxRoot.tsx:159`)

Clicking the header of the tool that is already open now closes it (`clearActiveTool()` at `src/BloomBrowserUI/bookEdit/toolbox/ToolboxRoot.tsx:163`) instead of being ignored, so the sidebar can end up with every tool collapsed while the tool it was showing goes on working invisibly.

Impact: The user can click a tool's title bar and be left with an empty sidebar (and no visible sign of which tool is still acting on the page) until they click another title bar.

**Why the previous code deliberately ignored the collapse, and what changed**

The removed handler explicitly refused to honor a collapse: "Clicking the open tool's header can't close it: the toolbox always has an active tool, and the effect that syncs us with the legacy toolbox re-expands whatever legacy still thinks is current. Honoring the collapse therefore only produced a flash... (BL-16533)". That re-expanding sync effect i…

src/BloomBrowserUI/bookEdit/toolbox/toolboxState.ts
```diff
@@ -0,0 +1,354 @@
...
+ * listeners, i.e. to toolbox.ts).
+ */
+export function withdrawTool(toolId: string): void {
+    const remainingToolIds = theState.offeredToolIds.filter(
+        (id) => id !== toolId,
+    );
+    if (remainingToolIds.length === theState.offeredToolIds.length) {
+        return;
+    }
+    if (theState.activeToolId !== toolId) {
```
andrew-polk, line 167: [Devin] **Bug**: Collapsed Game tool remains current

(`src/BloomBrowserUI/bookEdit/toolbox/toolboxState.ts:167`)

When a collapsed conditional tool is withdrawn, `withdrawTool` removes it without replacing the still-current tool. `currentToolId` remains Game, so leaving its page never activates another panel or lifecycle.

src/BloomBrowserUI/bookEdit/toolbox/ToolboxRoot.tsx
```diff
@@ -80,732 +74,250 @@ const toolboxHeaderIconStyles = css`
...
-            }
-        };
-    }, [props.element, clearLegacyAccordionSizing]);
+// One tool's section: its header, its panel, and its lifecycle.
+//
+// The lifecycle hook lives here, in the component that *contains* the panel, rather than
+// beside it: React runs a child's effects before its parent's, so by the time we tell the
+// tool to show itself its panel has mounted, as it always had under the old imperative
+// order.
+const ToolSection: React.FunctionComponent<{
```
andrew-polk, line 100: [Devin] **Bug**: New toolbox section component destructures its props, against the project's stated React style rule

(`src/BloomBrowserUI/bookEdit/toolbox/ToolboxRoot.tsx:101`)

The new toolbox section component takes its props apart in the parameter list (`({ section, isExpanded, isRunning, pageGeneration })` at `src/BloomBrowserUI/bookEdit/toolbox/ToolboxRoot.tsx:109`) instead of receiving a single props object, which the repository's coding rules forbid.

Impact: The code deviates from the house style the project requires for all React components, making it inconsistent with every other component.


**Where the rule comes from and how the rest of the PR follows it**

`AGENTS.md` (repo root) states under Code Style: "do not destructure props", and `src/BloomBrowserUI/AGENTS.md` repeats it: "Do not destructure props. `props.foo` is easier to understand."

Every other component added or touched in this PR follows the rule — e.g. `GamePanel` uses `props.tool` (`src/BloomBrowserUI/bookEdit/toolbox/games/GameTool.tsx:1707-1719`), `ToolboxCheckbox` uses `props.toolI…

src/BloomBrowserUI/bookEdit/toolbox/readers/decodableReader/decodableReaderTool.tsx
```diff
@@ -50,72 +52,72 @@ export class DecodableReaderTool extends ToolboxToolReactAdaptor {
...
-                    Promise.resolve(work()).then(
-                        () => restoreDone.resolve(),
-                        () => restoreDone.resolve(),
-                    );
-                } catch {
-                    restoreDone.resolve();
-                }
-            };
+    /** Restores the stage (and sort order) this book was last using; see ITool. */
+    public async beginRestoreSettings(
```
andrew-polk, line 56: [Devin] **Bug**: Decodable Reader tool can crash on activation for a book with no saved settings

(`src/BloomBrowserUI/bookEdit/toolbox/readers/decodableReader/decodableReaderTool.tsx:56`)

The Decodable Reader tool reads its saved settings (`settings["decodableReaderState"]` at `src/BloomBrowserUI/bookEdit/toolbox/readers/decodableReader/decodableReaderTool.tsx:60`) without first checking that the settings object exists, so when a book has no saved toolbox settings the tool errors out before it can open.

Impact: For a book whose saved settings come back empty/absent, opening the Decodable Reader can silently fail to activate the tool.

**Why the settings object can be null and why the sibling tool guards against it**

Both reader tools receive the same `savedSettings` object from `toolbox.ts` (`activateTool`/`applyToolboxStateToUpdatedPage` call `beginRestoreSettings(savedSettings)`; `savedSettings` is reassigned from the server response at `src/BloomBrowserUI/bookEdit/toolbox/toolbox.ts:487` and `:503`). The Leveled Reader's `beginRestoreSettings` in `src/BloomBrowserUI/…

src/BloomBrowserUI/bookEdit/toolbox/useToolLifecycle.ts
```diff
@@ -0,0 +1,122 @@
...
+    // The whole sequence is one async function because its steps are ordered by their
+    // promises: a tool is not ready to be shown until beginRestoreSettings() resolves, and
+    // newPageReady() must not run until showTool() has (talkingBookTool's is async, and
+    // BL-10471 was a flash caused by not waiting).
+    //
+    // pageGeneration is a dependency because a new page needs all three again, not just
+    // newPageReady(): some tools do their page-dependent setup in showTool(), and the tool's
+    // saved state has to be re-read in case the page belongs to a different book. That is
+    // what the toolbox has always done when a page was replaced.
+    React.useEffect(() => {
```
andrew-polk, line 68: [Devin] **Investigate**: Effect justification needs review

(`src/BloomBrowserUI/bookEdit/toolbox/useToolLifecycle.ts:68`)

Repository guidance requires explicit justification for each `useEffect`. These comments explain sequencing and cleanup, but not why effects are necessary.

src/BloomBrowserUI/bookEdit/toolbox/ToolboxRoot.tsx
andrew-polk, line 1: [Devin] **Investigate**: Tools now rely on ToolboxRoot's single ThemeProvider

(`src/BloomBrowserUI/bookEdit/toolbox/ToolboxRoot.tsx:334`)

Removing the per-tool `<ThemeProvider theme={toolboxTheme}>` wrappers (canvas, games, motion, music, impairment visualizer, talking book, both reader tools, ReaderToolSwitch) is safe for the panels themselves: ToolboxRoot wraps all sections in MUI's `ThemeProvider` with `toolboxTheme` (`src/BloomBrowserUI/bookEdit/toolbox/ToolboxRoot.tsx:330`), and MUI's provider also populates Emotion's theme context, so components that previously got the theme from `@emotion/react`'s provider still do. Worth double-checking during testing, though, is any tool UI that is mounted through a *separate* React root rather than as a child of ToolboxRoot (e.g. dialogs rendered into the workspace root document): those no longer inherit `toolboxTheme` from the tool that opened them.

src/BloomBrowserUI/bookEdit/toolbox/toolbox.ts
```diff
@@ -1142,234 +744,98 @@ export function removeToolboxMarkup() {
...
+ *
+ * All this does is persist the choice and record which tool is now the running one.
+ * Running it — restoring its settings, showing it, telling it the page is ready, and
+ * detaching and hiding whatever it replaced — follows from that state; see
+ * useToolLifecycle.ts.
+ *
+ * Note: do not name this parameter newToolId; that is the module-level variable this
+ * function clears at the end, and shadowing it silently breaks getActiveToolId().
+ */
+function switchTool(requestedToolId: string): void {
```
andrew-polk, line 760: [Devin] **Investigate**: The "More..." section can become the book's remembered current tool

(`src/BloomBrowserUI/bookEdit/toolbox/toolbox.ts:760`)

Expanding the "More..." section calls `setActiveTool("settings")`, which notifies `switchTool()` (`src/BloomBrowserUI/bookEdit/toolbox/toolbox.ts:760-779`) and so persists `current\tsettingsTool` in the book and records `settings` as `currentToolId`. This contradicts the comments that say the More section "can never be the current tool of a book" (`toolIds.ts:18-19`, `toolbox.ts:797-798`) and `getFirstOfferedToolId()`'s exclusion of it. I checked the pre-PR code and it behaved the same way (the legacy header carried `data-toolId="settingsTool"`, and the old `switchTool` matched it via `startsWith`), so this is pre-existing rather than introduced here — but the new comments assert an invariant the code does not enforce, which is worth either enforcing in `setActiveTool`/`switchTool` or correcting in the comments.

src/BloomBrowserUI/bookEdit/toolbox/readers/readerToolsModel.ts
```diff
@@ -1532,14 +1533,14 @@ export class ReaderToolsModel {
         return dataWords;
     }
 
+    /**
```
andrew-polk, line 1536: [Devin] **Investigate**: Reader stage/level persistence now becomes live much earlier in startup

(`src/BloomBrowserUI/bookEdit/toolbox/readers/readerToolsModel.ts:1536`)

The old guard (`typeof $("#toolbox").accordion !== "function"`, plus `isNaN(accordion("option","active"))`) was only satisfied after every tool had been loaded and the accordion initialized. The new guard is `isToolboxUiMounted()`, which becomes true as soon as `ToolboxRoot` mounts — in `toolboxBootstrap.ts:116-119` that is immediately before `initialize()` runs, i.e. before the enabled-tools fetch and before `restoreToolboxSettings()`. So `saveState()` can now fire during startup, when stage/sort/level may still be at their defaults, where previously it would have been suppressed. Worth confirming that nothing calls `saveState()` between mount and the first `restoreState()`/`beginRestoreSettings()` (the concern the doc comment itself raises: "saving would overwrite the book's real settings with defaults").

src/BloomBrowserUI/bookEdit/toolbox/games/GameTool.tsx
```diff
@@ -1737,67 +1757,67 @@ export class GameTool extends ToolboxToolReactAdaptor {
...
-    public beginRestoreSettings(_settings: string): JQueryPromise<void> {
-        // Nothing to do, so return an already-resolved promise.
-        const result = $.Deferred<void>();
-        result.resolve();
-        return result;
-    }
-
     private lastPageId = "";
 
     public newPageReady() {
```
andrew-polk, line 1816: [Devin] **Investigate**: The Game tool's panel now refreshes only indirectly, via the tab-control round trip

(`src/BloomBrowserUI/bookEdit/toolbox/games/GameTool.tsx:1816`)

`newPageReady()` used to end with `this.renderRoot()`, which bumped `pageGeneration` and re-rendered `DragActivityControls` unconditionally. It now relies on `setActiveDragActivityTab()` (`src/BloomBrowserUI/bookEdit/toolbox/games/GameTool.tsx:1964-1999`) reaching `GameTool.setActiveTab()` → `notifyPanel()`. That call is made on both branches of `newPageReady`, but `setActiveDragActivityTab` bails out and retries on a 100ms timer when the page frame is not ready, and it goes through `getToolboxBundleExports()`, so the panel refresh is now conditional on that cross-bundle lookup succeeding. Worth exercising "switch to a game page while the Game tool is open" to confirm the panel state (tab, theme chooser, sound options) still refreshes.

src/BloomExe/Edit/ToolboxView.cs
```diff
@@ -72,18 +76,29 @@ public static void SetupToolboxForCollection(CollectionSettings settings)
...
+        /// A new tool's folder does NOT normally belong here. This list does not control what
+        /// the server will serve: everything the toolbox fetches uses a full path
+        /// (e.g. "/bloom/bookEdit/toolbox/motion/motion.svg"), which BloomServer resolves
+        /// against the browser root without consulting this search path. Several toolbox
+        /// folders that serve assets (games, impairmentVisualizer, readers) have never been
+        /// listed here and work fine. What the search path is for is lookups by bare
+        /// file name, which for these folders means only their stylesheets
+        /// (BloomServer.ProcessCssFile).
+        /// </summary>
         public static IEnumerable<string> GetToolboxServerDirectories()
```
andrew-polk, line 91: [Devin] **Investigate**: Toolbox folders dropped from factory search path

(`src/BloomExe/Edit/ToolboxView.cs:91`)

[GetToolboxServerDirectories](src/BloomExe/Edit/ToolboxView.cs:91) no longer lists motion, music, signLanguage, imageDescription, or canvas. Searches found no bare-filename CSS references to these tools; icons use full `/bloom/...` URLs and exist, and each `.less` is bundled via a TSX import. Removal appears safe; worth confirming nothing fetches these folders' assets by bare name.

src/BloomBrowserUI/bookEdit/toolbox/ToolboxRoot.tsx
andrew-polk, line 158: [Devin] **Investigate**: Clicking the open section header now really collapses it, reversing the BL-16533 workaround

(`src/BloomBrowserUI/bookEdit/toolbox/ToolboxRoot.tsx:159`)

The old `onChange` deliberately ignored `willBeExpanded === false` because "the toolbox always has an active tool" and honoring the collapse produced a flash (BL-16533). The new handler calls `clearActiveTool()` instead, and nothing re-expands, so the user can now end up with every section collapsed and no tool panel visible while `currentToolId` (and therefore the running tool, its markup and its keystroke handling) stays as it was. `toolboxState.clearActiveTool()` documents this as deliberate, so it may be the intended new behaviour, but it is a visible UX change from 6.3 that the PR description does not mention; worth confirming it was a decision rather than a side effect of dropping the sync effect.

src/BloomBrowserUI/bookEdit/toolbox/toolboxState.ts
```diff
@@ -0,0 +1,354 @@
...
+        ]),
+    });
+}
+
+/**
+ * Stops offering this tool's section, if it has one. If it was the active section, the
+ * first remaining section becomes active (and that is reported to the active-tool
+ * listeners, i.e. to toolbox.ts).
+ */
+export function withdrawTool(toolId: string): void {
```
andrew-polk, line 160: [Devin] **Investigate**: Withdrawing a tool only picks a replacement when that tool is the expanded one, not when it is merely the current one

(`src/BloomBrowserUI/bookEdit/toolbox/toolboxState.ts:160`)

`withdrawTool()` compares against `activeToolId`. Since the two can now diverge (collapsing a section clears `activeToolId` but leaves `currentToolId`), leaving a game page while the Game section is collapsed withdraws the section without reporting a replacement, so `currentToolId` keeps naming a tool the toolbox no longer offers. Its React section unmounts, so its lifecycle cleanup does run, but `getCurrentTool()` in `src/BloomBrowserUI/bookEdit/toolbox/toolbox.ts:422-428` still returns it, and `pageEditingMarkup.ts` will then call `updateMarkup()` on a tool that is neither shown nor attached. Reachable only through the collapse path discussed separately; keying the replacement logic on `currentToolId` as well would close it.

src/BloomBrowserUI/bookEdit/toolbox/toolboxState.ts
andrew-polk: Claude Fable 5: Fixed in 563f227c4a: withdrawTool now notices when the withdrawn tool is current-but-collapsed and switches toolbox.ts to a surviving tool, so nothing keeps driving markup for an unmounted panel. (The local review found the same gap independently.)

src/BloomBrowserUI/bookEdit/toolbox/toolboxState.ts
andrew-polk: Claude Fable 5: Fixed in 563f227c4a — same fix as the 'Collapsed Game tool remains current' bug: the early-return branch of withdrawTool now keys on currentToolId too.

src/BloomBrowserUI/bookEdit/toolbox/ToolboxRoot.tsx
andrew-polk: Claude Fable 5: Fixed in 563f227c4a: ToolSection now takes (props) per the house style.

src/BloomBrowserUI/bookEdit/toolbox/readers/decodableReader/decodableReaderTool.tsx
andrew-polk: Claude Fable 5: Fixed in 563f227c4a: the Decodable Reader now guards settings?.["decodableReaderState"] exactly like the Leveled Reader's Sentry-motivated guard, with a comment saying why.

src/BloomBrowserUI/bookEdit/toolbox/toolbox.ts
andrew-polk: Claude Fable 5: Agreed with your own analysis that the behavior is pre-existing (the old jQuery header carried data-toolId="settingsTool" and went through the same switchTool path). Rather than change long-standing persistence behavior in a cleanup PR, 563f227c4a corrects the two comments so they no longer claim an invariant the code doesn't enforce.

src/BloomBrowserUI/bookEdit/toolbox/useToolLifecycle.ts
andrew-polk: Claude Fable 5: Not acting on this: useToolLifecycle.ts's comments do carry the justification — the effects exist because a tool must attach/detach exactly when React mounts/unmounts its panel and when isRunning/pageGeneration change, which is stated in the file's header and per-effect comments alongside the sequencing rationale.

src/BloomBrowserUI/bookEdit/toolbox/ToolboxRoot.tsx
andrew-polk: Claude Fable 5: Verified: MUI's ThemeProvider also populates Emotion's theme context, so panel components read the same theme as before (the local review confirmed this too). Tool UI mounted through separate React roots (workspace-level dialogs) never inherited the toolbox theme from these wrappers — they render into a different document. Added a note to the card's manual test ideas to eyeball tool dialogs for theme regressions.

src/BloomBrowserUI/bookEdit/toolbox/readers/readerToolsModel.ts
andrew-polk: Claude Fable 5: Traced the window between ToolboxRoot mounting and beginRestoreSettings: every restore path calls setStageNumber/setLevelNumber/setSort with skipSave, and setMarkupType (the one unconditional saveState caller) only runs from showTool/newPageReady, which the lifecycle orders after restore. So a default-overwriting save in that window would need a user interaction in the sub-second before restore completes, which the old accordion guard did not prevent either once it initialized.

src/BloomBrowserUI/bookEdit/toolbox/games/GameTool.tsx
andrew-polk: Claude Fable 5: Exercised live: switching to a game page with the Game tool open refreshes the panel (tab state and controls) through setActiveDragActivityTab; the 100ms retry only defers until the page frame exists, and getToolboxBundleExports resolves in the same realm that toolboxBootstrap populated. Added game-page switching to the card's manual test ideas as well.

src/BloomExe/Edit/ToolboxView.cs
andrew-polk: Claude Fable 5: Agreed with your own analysis that the removal is safe — the local review additionally confirmed audioRecording.css (which stays listed) is the only toolbox stylesheet fetched by bare name from another document, and its mask-image URLs were absolutized in this PR.

src/BloomBrowserUI/bookEdit/toolbox/ToolboxRoot.tsx
andrew-polk: Claude Fable 5: Same subject as the open 'Clicking the open tool's title bar' thread above — the collapse is deliberate (clearActiveTool documents it), but whether the UX change from BL-16533-era behavior is wanted is with the developer on that thread; keeping the decision in one place and resolving this duplicate.

andrew-polk:
Claude Fable 5: Consulted Devin on 2026-09-12 01:00 UTC up to commit `563f227c4a`. Round 1 (at 22d3e20ee1): 4 bugs + 12 investigate flags + 16 informational; 3 bugs and 1 flag fixed in 563f227c4a, the rest assessed on their threads. Round 2 (at 563f227c4a) re-reported the identical set (full re-review, no incremental diff); screened per the stale-repeat rule — the reviewed commit demonstrably contains the fixes — so no new findings. One thread deliberately left open for a developer decision: the header-click-collapses behavior.

src/BloomBrowserUI/bookEdit/toolbox/ToolboxRoot.tsx
andrew-polk: Claude Fable 5: Decision from the developer (2026-09-11): keep 6.5's behavior — this PR is not the place to change it. Fixed in 1c816f09a5: clicking the open tool's header once again does nothing, exactly as in 6.5/6.3 (originally BL-16533). The state store still supports a section-less state (the withdraw paths produce it, and the withdraw fix from the earlier round stays); the UI just offers no gesture for it.

andrew-polk:
Claude Fable 5: Consulted Devin on 2026-09-12 04:40 UTC up to commit `1c816f09a5`. Round 3 re-reported the identical finding set a third time (no incremental diff on this large PR); screened as stale — the reviewed commit contains the fixes, and the header-click finding is moot since 1c816f09a5 restores 6.5 behavior per the developer's decision (recorded on its thread). No new findings; every thread has a documented outcome.

PR #8110, opened 2026-07-27
Stop publishing hanging forever on the page-checks browser (BL-16612)

andrew-polk:
Found while investigating repeated Nightly Build and Test failures where all 12 visual-regression cases timed out (~24 minutes wasted). The hang is in **publishing**, not in the tests. Bloom's log ended at the same point in every failing run:

```
Failed to navigate fully to RemoveUnwantedContentInternal DOM
```

and then went silent for 23 minutes. Three consecutive nightlies ran the *same* commit `e2a5d45e62` with results fail / pass / fail, so it is a flake rather than a regression.

## Two defects, which have to be fixed together

**1. `OffScreenBrowser.RunAndBlock` waited with no timeout.** When the page-checks navigation failed, `PublishHelper` carried on (per the BL-7892 comment) and called `RunJavascript` on a browser whose page had never loaded — and that never returned. The blocked thread is a BloomServer worker, so the request was never answered and publishing stayed wedged for the life of the process. Every blocking call is now bounded and throws `OffScreenBrowserTimeoutException` instead of waiting forever. The unbounded `_ready.Wait()` in the constructor is bounded for the same reason.

**2. Carrying on after a failed navigation silently produced a bad book.** With no element information, `IsDisplayed` answers "displayed" for everything and `FontsUsed` comes back empty, so a BloomPUB keeps content that should have been stripped and embeds none of its fonts. Only epub publishing noticed, because `IsDisplayed` throws for it. So merely adding the timeout from (1) would have converted the hang into quiet corruption — which is why both land together. Page checks now retry once on a fresh renderer and, failing that, fail the publish (the existing handler reports it) rather than emit a book we know is wrong.

## Why the navigation stalls is still unproven

The leading hypothesis is **BloomServer worker starvation**: the thread blocking on the browser is itself a worker, and the browser's navigation needs a worker to serve the page it is loading. `_countBlockedThreads` is not computed automatically — callers must register — and `OffScreenBrowser` never did, so a worker blocked there was invisible to the logic in `QueueRequest` that adds a worker when all of them are blocked. `RunAndBlock` now registers, which should break that cycle if it is the cause.

Deliberately *not* raised: the 10s navigation budget. Per-test durations from the passing run show the entire first case (tab switches, book select, branding, theme, preview screenshot, full BloomPUB staging, several player captures) taking **18.4s**, so that navigation normally finishes in a small fraction of 10s. The failures blew through 10s and then never completed — bimodal, i.e. a stall rather than a near-miss, so a longer wait is not the fix. Added `BloomServer.GetWorkerPoolDiagnostics()` and log it, with the elapsed time, whenever page checks fail, so the next occurrence tells us whether the pool was starved.

## Test-side

The visual-regression suite now bounds each request to Bloom and names the call that hung, instead of an opaque "test timed out" pointing at the test function — which was all the original failure gave us to go on. Once Bloom stops answering, the remaining cases fail immediately rather than each burning 120s. Ordinary image-diff failures still run every case, so we keep every diff.

## Testing

- New `OffScreenBrowserTests.RunJavascript_WhenTheBrowserDoesNotComeBackInTime_ThrowsRatherThanBlockingForever` proves the backstop fires (it gives up at 500ms rather than waiting out a 5s script) and `GetWorkerPoolDiagnostics_WithAServerRunning_ReportsWorkerCounts` covers the new diagnostics. Both pass, with the existing test as a sanity check.
- Verified separately that `node-fetch` honors `AbortSignal.timeout` — and that it reports `AbortError`, **not** `TimeoutError`, which is why the new code checks `signal.aborted` rather than the error name.
- `BloomTests.Publish` (which drives `RemoveUnwantedContent`) cannot run in a worktree without a built `output/browser`: it aborts on a `Debug.Fail` for a missing `favicon.ico`. Confirmed identical on pristine code (163 vs 164 failures, same abort point), so this PR's CI run is the real check on those suites.

[Devin review](https://devinreview.com/BloomBooks/BloomDesktop/pull/8110)

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8110)

### DIFF
--- a/src/BloomExe/Publish/OffScreenBrowser.cs
+++ b/src/BloomExe/Publish/OffScreenBrowser.cs
@@ -38,10 +38,9 @@ namespace Bloom.Publish
     /// Blocking here cannot deadlock on the message loop, since the thread that blocks is never the thread
     /// that must pump. But it is NOT immune to deadlock in general: navigation asks BloomServer for the page,
     /// so a caller that blocks while holding the last free BloomServer worker would be waiting for a page
-    /// that nobody is left to serve. Two things address that (BL-16612): every blocking call is bounded by a
-    /// backstop timeout and throws <see cref="OffScreenBrowserTimeoutException"/> rather than waiting
-    /// forever, and we register the block with BloomServer so its own starvation logic can see this thread
-    /// at all (see RunAndBlock).
+    /// that nobody is left to serve. Two things guard against that (BL-16612): we register the block with
+    /// BloomServer so it can add a worker (see RunAndBlock), and every blocking call is bounded by a backstop
+    /// timeout and throws <see cref="OffScreenBrowserTimeoutException"/> rather than waiting forever.
     ///
     /// The browser is a real <see cref="WebView2Browser"/> (not a bare WebView2 control) so navigation goes
     /// through Bloom's BloomServer/in-memory-file plumbing and resolves CSS, fonts, and relative paths.
@@ -344,11 +343,16 @@ namespace Bloom.Publish
             // Tell BloomServer that this thread is about to block, in case it is one of its workers.
             // BloomServer does NOT work that out for itself (see RegisterThreadBlocking): a worker that
             // blocks without registering is invisible to the logic in QueueRequest that adds a worker when
-            // every existing one is blocked. Thread stacks from the 2026-08-03 nightly hang show exactly
-            // that blind spot: six workers, five of them registered as blocked on the API lock and the
-            // sixth -- this one -- blocked here without registering, so the top-up test asked 5 >= 6 and
-            // said no while the true state was 6 of 6 unavailable (BL-16612).
+            // every existing one is blocked. That matters a great deal here, because what we are waiting
+            // for is a browser navigating to a page that BloomServer itself has to serve. If every worker
+            // is blocked and none is left to serve that page, the navigation can never complete and we
+            // would be waiting for something that cannot happen (BL-16612).
+            //
+            // NoteWorkerPoolHeadroom is pure measurement: it records how close the pool came to having
+            // nothing left to serve requests, which is what makes the starvation theory testable on a
+            // healthy run instead of only after a failure.
             var server = BloomServer._theOneInstance;
+            server?.NoteWorkerPoolHeadroom();
             server?.RegisterThreadBlocking();
             try
             {
@@ -376,7 +380,8 @@ namespace Bloom.Publish
                     // the browser, so the browser is now in an unknown state: the caller must discard it
                     // (StartFreshBrowser) or stop using this instance.
                     throw new OffScreenBrowserTimeoutException(
-                        $"The off-screen browser did not respond within {effectiveTimeoutMs}ms."
+                        $"The off-screen browser did not respond within {effectiveTimeoutMs}ms. "
+                            + BloomServer.GetWorkerPoolDiagnostics()
                     );
                 }
                 return tcs.Task.GetAwaiter().GetResult();
--- a/src/BloomExe/Publish/OffScreenBrowser.cs
+++ b/src/BloomExe/Publish/OffScreenBrowser.cs
@@ -352,9 +352,14 @@ namespace Bloom.Publish
             // the request that would unblock us can already be sitting in the queue by the time we get here.
             var server = BloomServer._theOneInstance;
             server?.NoteWorkerPoolHeadroom();
-            server?.RegisterThreadBlockingAndEnsureAFreeWorker();
+            // Register INSIDE the try, and unregister only if we got that far, so the pairing is guaranteed
+            // by structure rather than by an argument about what can throw. A leaked count would make the
+            // server permanently believe a worker is blocked.
+            var registered = false;
             try
             {
+                server?.RegisterThreadBlockingAndEnsureAFreeWorker();
+                registered = true;
                 bool completed;
                 try
                 {
@@ -387,7 +392,8 @@ namespace Bloom.Publish
             }
             finally
             {
-                server?.RegisterThreadUnblocked();
+                if (registered)
+                    server?.RegisterThreadUnblocked();
             }
         }
 
--- a/src/BloomExe/Publish/OffScreenBrowser.cs
+++ b/src/BloomExe/Publish/OffScreenBrowser.cs
@@ -38,9 +38,10 @@ namespace Bloom.Publish
     /// Blocking here cannot deadlock on the message loop, since the thread that blocks is never the thread
     /// that must pump. But it is NOT immune to deadlock in general: navigation asks BloomServer for the page,
     /// so a caller that blocks while holding the last free BloomServer worker would be waiting for a page
-    /// that nobody is left to serve. What keeps that survivable (BL-16612) is that every blocking call is
-    /// bounded by a backstop timeout and throws <see cref="OffScreenBrowserTimeoutException"/> rather than
-    /// waiting forever.
+    /// that nobody is left to serve. Two things address that (BL-16612): every blocking call is bounded by a
+    /// backstop timeout and throws <see cref="OffScreenBrowserTimeoutException"/> rather than waiting
+    /// forever, and we register the block with BloomServer so its own starvation logic can see this thread
+    /// at all (see RunAndBlock).
     ///
     /// The browser is a real <see cref="WebView2Browser"/> (not a bare WebView2 control) so navigation goes
     /// through Bloom's BloomServer/in-memory-file plumbing and resolves CSS, fonts, and relative paths.
@@ -340,34 +341,50 @@ namespace Bloom.Publish
                 null
             );
 
-            bool completed;
+            // Tell BloomServer that this thread is about to block, in case it is one of its workers.
+            // BloomServer does NOT work that out for itself (see RegisterThreadBlocking): a worker that
+            // blocks without registering is invisible to the logic in QueueRequest that adds a worker when
+            // every existing one is blocked. Thread stacks from the 2026-08-03 nightly hang show exactly
+            // that blind spot: six workers, five of them registered as blocked on the API lock and the
+            // sixth -- this one -- blocked here without registering, so the top-up test asked 5 >= 6 and
+            // said no while the true state was 6 of 6 unavailable (BL-16612).
+            var server = BloomServer._theOneInstance;
+            server?.RegisterThreadBlocking();
             try
             {
-                completed = tcs.Task.Wait(effectiveTimeoutMs);
+                bool completed;
+                try
+                {
+                    completed = tcs.Task.Wait(effectiveTimeoutMs);
+                }
+                catch (AggregateException)
+                {
+                    // The work threw. Fall through to GetResult() below, which rethrows the original
+                    // exception rather than the AggregateException that Wait wraps it in.
+                    completed = true;
+                }
+                if (!completed)
+                {
+                    // Nobody will ever await the task we are abandoning, so observe its exception here;
+                    // otherwise a later failure on it resurfaces as an unobserved TaskException. We already
+                    // have the diagnosis we need, so there is nothing else to do with it.
+                    _ = tcs.Task.ContinueWith(
+                        t => _ = t.Exception,
+                        TaskContinuationOptions.OnlyOnFaulted
+                    );
+                    // The posted work is still queued or running on the dedicated thread and may yet touch
+                    // the browser, so the browser is now in an unknown state: the caller must discard it
+                    // (StartFreshBrowser) or stop using this instance.
+                    throw new OffScreenBrowserTimeoutException(
+                        $"The off-screen browser did not respond within {effectiveTimeoutMs}ms."
+                    );
+                }
+                return tcs.Task.GetAwaiter().GetResult();
             }
-            catch (AggregateException)
+            finally
             {
-                // The work threw. Fall through to GetResult() below, which rethrows the original
-                // exception rather than the AggregateException that Wait wraps it in.
-                completed = true;
+                server?.RegisterThreadUnblocked();
             }
-            if (!completed)
-            {
-                // Nobody will ever await the task we are abandoning, so observe its exception here;
-                // otherwise a later failure on it resurfaces as an unobserved TaskException. We already
-                // have the diagnosis we need, so there is nothing else to do with it.
-                _ = tcs.Task.ContinueWith(
-                    t => _ = t.Exception,
-                    TaskContinuationOptions.OnlyOnFaulted
-                );
-                // The posted work is still queued or running on the dedicated thread and may yet touch
-                // the browser, so the browser is now in an unknown state: the caller must discard it
-                // (StartFreshBrowser) or stop using this instance.
-                throw new OffScreenBrowserTimeoutException(
-                    $"The off-screen browser did not respond within {effectiveTimeoutMs}ms."
-                );
-            }
-            return tcs.Task.GetAwaiter().GetResult();
         }
 
         /// <summary>
--- a/src/BloomExe/Publish/OffScreenBrowser.cs
+++ b/src/BloomExe/Publish/OffScreenBrowser.cs
@@ -31,10 +31,16 @@ namespace Bloom.Publish
     /// message loop, and THAT thread — not the caller — services the browser's callbacks. So a caller can
     /// simply block for a result. It never has to pump the MAIN UI message loop the way
     /// Browser.RunJavascriptWithStringResult_Sync_Dangerous does (Application.DoEvents), which lets unrelated
-    /// user commands run in the middle of the call stack — the reentrancy blamed for BL-12614 / BL-13120. And
-    /// it never deadlocks, because the thread that blocks is never the thread that must pump. Async
-    /// continuations for the browser's operations run on the owning thread, so the WebView2 is only ever
-    /// touched by that thread.
+    /// user commands run in the middle of the call stack — the reentrancy blamed for BL-12614 / BL-13120.
+    /// Async continuations for the browser's operations run on the owning thread, so the WebView2 is only
+    /// ever touched by that thread.
+    ///
+    /// Blocking here cannot deadlock on the message loop, since the thread that blocks is never the thread
+    /// that must pump. But it is NOT immune to deadlock in general: navigation asks BloomServer for the page,
+    /// so a caller that blocks while holding the last free BloomServer worker would be waiting for a page
+    /// that nobody is left to serve. What keeps that survivable (BL-16612) is that every blocking call is
+    /// bounded by a backstop timeout and throws <see cref="OffScreenBrowserTimeoutException"/> rather than
+    /// waiting forever.
     ///
     /// The browser is a real <see cref="WebView2Browser"/> (not a bare WebView2 control) so navigation goes
     /// through Bloom's BloomServer/in-memory-file plumbing and resolves CSS, fonts, and relative paths.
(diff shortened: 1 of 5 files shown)


src/BloomExe/Publish/OffScreenBrowser.cs
```diff
@@ -286,7 +339,55 @@ private T RunAndBlock<T>(Func<Task<T>> asyncFunc)
...
+            // Tell BloomServer that this thread is about to block, in case it is one of its workers.
+            // BloomServer does NOT work that out for itself (see RegisterThreadBlocking): a worker that
+            // blocks without registering is invisible to the logic in QueueRequest that adds a worker when
+            // every existing one is blocked. That matters a great deal here, because what we are waiting
+            // for is a browser navigating to a page that BloomServer itself has to serve. If every worker
+            // is blocked and none is left to serve that page, the navigation can never complete and we
+            // would be waiting for something that cannot happen (BL-16612). Registering lets BloomServer
+            // spin up the worker that breaks the cycle — and we use the ...AndEnsureAFreeWorker variant
+            // because the plain one only takes effect when the NEXT request arrives, which may be never:
+            // the request that would unblock us can already be sitting in the queue by the time we get here.
```
andrew-polk, line 352: [Devin] **Investigate**: Registration only helps if a NEW request arrives to trigger SpinUpAWorker

Devin flagged that registering the blocked thread might not actually break the deadlock it is meant to break.

`RunAndBlock` calls `RegisterThreadBlocking()` to bump `_countBlockedThreads`, but the escape hatch that acts on that count lives in `BloomServer.QueueRequest` (`src/BloomExe/web/BloomServer.cs:1932-1938`) — and it only runs when a **new** request arrives. The request that would unblock us is the page the off-screen browser is loading, and that may already be sitting in the queue by the time we register. If every worker is blocked and nothing new comes in behind it, no worker is ever added and the cycle stands.

In other words, registering alone only makes the starvation *visible*; it does not necessarily fix it — which matters here because breaking that cycle is the whole point.

andrew-polk: (Claude Opus 5) Good catch — this was a real gap, and it's fixed in `04ccfe0f11`.

Added `BloomServer.RegisterThreadBlockingAndEnsureAFreeWorker()`, which does the "is anyone left to serve requests?" check at the moment the blocked count goes up, and adds a worker there and then, instead of waiting for a new request that may never come. `RunAndBlock` now calls that. The plain `RegisterThreadBlocking` is left alone so the existing callers (modal dialogs, UI-thread invokes) keep their current behavior.

Resolving.

src/BloomVisualRegressionTests/index.spec.ts
```diff
@@ -237,56 +237,147 @@ describe("All books", () => {
...
+    }
+
+    // The common case: we only care whether Bloom accepted the request.
+    async function bloomFetchOk(
+        what: string,
+        url: string,
+        init?: Parameters<typeof fetch>[1],
+        timeoutMs = kRequestTimeoutMs,
+    ) {
+        return bloomRequest(what, url, init, timeoutMs, async (r) => r.ok);
```
andrew-polk, line 307: [Devin] **Investigate**: Abort timer stays armed after unconsumed-body responses; late abort may emit an error on the response body

Devin's concern: `bloomFetchOk` looks at `response.ok` and never reads the body, so node-fetch might keep its abort listener attached. The `AbortSignal.timeout` timer would then still fire ~45s later — during some *later* test case — and node-fetch's abort handler could emit an `error` on the unconsumed `response.body`. An unhandled Node stream `error` becomes an uncaught exception, which would crash the whole suite. Devin noted it could not verify node-fetch's finalize semantics because `node_modules` wasn't available to it, and called out that this is new behavior versus the old code (which passed no signal at all).

Worth taking seriously, since it would turn a diagnostic improvement into a new way for the suite to die.

src/BloomExe/Publish/PublishHelper.cs
```diff
@@ -77,6 +77,90 @@ private OffScreenBrowser GetOrCreatePageChecksBrowser()
...
+                Debug.WriteLine(message);
+                Logger.WriteEvent(message);
+
+                if (attempt < kPageChecksAttempts)
+                    browser.StartFreshBrowser();
+            }
+            throw new ApplicationException(
+                "Bloom could not determine which parts of this book are visible, so it stopped rather than "
+                    + "publish a book with the wrong content or missing fonts. See the log for details."
+            );
```
andrew-polk, line 185: [Devin] **Investigate**: Behavior change: failed page-checks navigation now fails BloomPUB publishing instead of producing a book

Devin's point: this PR deliberately turns a previously-silent degradation into a hard failure for BloomPUB (before, only epub noticed, via `IsDisplayed` throwing). It asks that we confirm the BloomPUB entry points surface the new `ApplicationException` as a friendly error rather than an unhandled crash, since this path was not previously reachable for BloomPUB.

That is the right question to ask of this change, so I traced every path that can now throw.

src/BloomVisualRegressionTests/index.spec.ts
andrew-polk: (Claude Opus 5) Checked this empirically rather than reasoning about it — I do have `node_modules`, so I ran the exact shape you describe against node-fetch 3.3.2: request completes fast, body never read, then idle well past the timeout.

```
request completed fast: ok=true, aborted=false
after waiting past the timeout: signal.aborted=true
PASS: no unhandled error from the late abort on an unconsumed body
```

The timer does fire late and the signal does become aborted, exactly as you predicted — but node-fetch emits nothing on the unconsumed body, with no `uncaughtException` and no `unhandledRejection` (I installed handlers for both, and used a 64KB body so it could not be trivially pre-buffered). So the failure mode doesn't materialize on this version.

Not an issue as things stand. Resolving.

src/BloomExe/Publish/PublishHelper.cs
andrew-polk: (Claude Opus 5) Good question to press on. I traced every path that can now reach this throw. There are exactly two callers of `RemoveUnwantedContent` — `BloomPubMaker.cs:651` (BloomPUB/BloomReader) and `EpubMaker.cs:1733` (epub, which already threw here, so it is unchanged) — and `BloomPubMaker`'s creation entry points are reached from:

| path | what happens now |
|---|---|
| Publish tab preview (`PublishApi.MakeBloompubPreview`) | explicit try/catch → `request.Failed(...)` + `NonFatalProblem.Report` — reported to the user |
| Save to file / Send to device (`PublishToBloomPubApi:388,415`) | no local catch |
| Bulk create (`BulkBloomPubCreator:121`) | no local catch |
| RAB project (`RabProjectService:1257`) | no local catch |
| CLI (`CreateArtifactsCommand`, `SendFontAnalyticsCommand`) | no local catch — CLI/harvester, fails the command |

**No crash risk:** the paths without a local catch still land in `BloomServer.RequestProcessorLoop`'s broad `catch (Exception error)`, which logs the message, URL and stack and fails that request. The worker thread survives and Bloom stays up.

**But the messaging is thin on those paths.** The user gets a failed request and a log entry rather than a friendly explanation — they may just see the operation stop. The exception text is already written in plain English for that reason ("Bloom could not determine which parts of this book are visible, so it stopped rather than publish a book with the wrong content or missing fonts"), but only the preview path actually shows it.

Improving that means touching user-facing error messaging at four call sites, which is past what I should decide alone, so I am leaving this thread open and putting it to the developer. Not a blocker for the fix itself.

andrew-polk:
(Claude Opus 5) Consulted Devin on 2026-07-27 up to commit `04ccfe0f11`.

Devin ran twice — once on `918f86e830` and again on the fix commit `04ccfe0f11`. No Bugs either time.

- **Registration only helps if a NEW request arrives** — a real gap, and the important one: registering the blocked thread only made starvation *visible*, and `QueueRequest` acts on it only when the next request arrives, which may never come. Fixed in `04ccfe0f11` via `RegisterThreadBlockingAndEnsureAFreeWorker()`. ([thread](https://github.com/BloomBooks/BloomDesktop/pull/8110#discussion_r3659431029), resolved.) Devin re-raised it on the second pass while still quoting the old `RegisterThreadBlocking()` call, so that repeat is stale.
- **Late abort on an unconsumed response body** — tested against node-fetch 3.3.2 rather than assumed; no unhandled error. ([thread](https://github.com/BloomBooks/BloomDesktop/pull/8110#discussion_r3659506678), resolved.)
- **Failed page checks now fail BloomPUB publishing** — traced every entry point: no crash risk, but the non-preview paths only log rather than explain. Left **open** for the developer. ([thread](https://github.com/BloomBooks/BloomDesktop/pull/8110#discussion_r3659506843).)

Four informational items from the first pass were acted on anyway (abandoned-task exception, superseded-publish timeout, body reads brought inside the request timeout, and only-timeouts marking Bloom wedged). The second pass's seven needed no action.

CI is green (`pr-automation`, `track`). CodeRabbit is disabled for this repo by `.coderabbit.yml` (`auto_review.enabled: false`).

andrew-polk:
(Claude Opus 5) **Re-split the branch's history** (force-push) so the fix and the instrumentation are separate commits, at Andrew's request — it was hard to tell how much of the diff existed only for diagnosis.

- `92150a65b7` — **the fix**: the bounded waits, retry-then-fail page checks, and the free-worker guarantee.
- `e313457b15` — **the instrumentation**: `GetWorkerPoolDiagnostics()` and its two log/message call sites, its test, and the visual-regression suite's bounded/labelled requests. No product behavior; revertable on its own.

Only ~33 lines of shipping product code are instrumentation (`BloomServer` +21, `PublishHelper` 9, `OffScreenBrowser` 3); the remaining 162 are test-harness and a test.

**The resulting tree is byte-identical to the pre-split `04ccfe0f11`** — same tree hash `10fe6b4feb…`. Nothing in the code changed, only how it is divided into commits, so the three review threads above are all still anchored and current (GitHub did not mark them outdated). Commit 1 was verified to build and pass its test on its own, so the branch bisects cleanly.

src/BloomExe/web/BloomServer.cs
```diff
@@ -2433,9 +2433,144 @@ public void RegisterThreadUnblocked()
...
+        /// </summary>
+        /// <remarks>
+        /// RegisterThreadBlocking on its own only makes starvation *visible*; the escape hatch in QueueRequest
+        /// acts on it when the NEXT request arrives. That is too late for a thread that blocks waiting on
+        /// something a queued request has to produce — the page an off-screen browser is loading, say — because
+        /// the request that would unblock it may already be sitting in the queue, with nothing new coming in
+        /// behind it to trigger the check. Then everyone waits forever. So callers who block on our own server
+        /// use this instead, and we do the check at the moment the count goes up (BL-16612).
+        /// </remarks>
+        public void RegisterThreadBlockingAndEnsureAFreeWorker()
```
andrew-polk, line 2479: [Devin] **Bug**: New server helper method ends up with no documentation while an unrelated field gets its description

The XML doc written for `RegisterThreadBlockingAndEnsureAFreeWorker` ended up above two unrelated private counters instead of above the method, so the public method reads as undocumented and the description is attached to the wrong thing.

andrew-polk: (Claude Opus 5) Correct, and my fault: this is an artifact of splitting the branch into purpose-separated commits, where the doc block was left behind when the method moved. Reattached in `dc98a17e0b`. Resolving.

src/BloomExe/Publish/PublishHelper.cs
```diff
@@ -77,6 +81,105 @@ private OffScreenBrowser GetOrCreatePageChecksBrowser()
...
+                // BL-16612, so leave this in even once the retry usually saves us.
+                var message =
+                    $"Page checks failed on attempt {attempt} of {kPageChecksAttempts} after "
+                    + $"{timer.ElapsedMilliseconds}ms: {whatWentWrong}. "
+                    + BloomServer.GetWorkerPoolDiagnostics();
+                Debug.WriteLine(message);
+                Logger.WriteEvent(message);
+
+                if (attempt < kPageChecksAttempts)
+                {
```
andrew-polk, line 160: [Devin] **Bug**: Recovery attempt after an unresponsive publishing browser can itself hang and abort with an unhelpful message

After the page-checks browser stops responding, the retry asks *that same unresponsive browser* for a clean renderer (`browser.StartFreshBrowser()`), outside any error handling. So it can hit its own backstop, throw, skip the retry entirely, and surface a technical timeout instead of the explanation written for users -- in exactly the hang this retry was added for.

andrew-polk: (Claude Opus 5) Right, and the worst part is it fails precisely in the scenario the retry exists for. Fixed in `dc98a17e0b`: the recovery attempt is now caught and logged, so the loop reaches its own verdict and the user gets the intended explanation. Resolving.

src/BloomExe/Publish/OffScreenBrowser.cs
```diff
@@ -286,7 +339,62 @@ private T RunAndBlock<T>(Func<Task<T>> asyncFunc)
...
+            var server = BloomServer._theOneInstance;
+            server?.NoteWorkerPoolHeadroom();
+            // Register INSIDE the try, and unregister only if we got that far, so the pairing is guaranteed
+            // by structure rather than by an argument about what can throw. A leaked count would make the
```
andrew-polk, line 356: [Devin] **Bug**: Blocked-thread counter can leak if adding a worker fails, wrongly triggering extra worker creation

The register call sat outside the try/finally that later reports the thread unblocked, so a failure partway through would never bring the count back down -- and an inflated count makes the guard fire on essentially every later block.

andrew-polk: (Claude Opus 5) Agreed, and worth noting we converged on this independently -- I had found the same leak reviewing the guard before handing it back, and made the method unable to throw once it has registered (`5994550c6d`).

Your framing is the better one though: that fix rests on an *argument* about what can throw, where the call site should make the pairing structural. So `dc98a17e0b` also moves the register inside the try and unregisters only if it got that far. Both belts now: the method cannot throw after registering, and the pairing holds regardless. Resolving.

src/BloomVisualRegressionTests/index.spec.ts
```diff
@@ -243,56 +243,150 @@ describe("All books", () => {
...
     }
 
+    // How long we allow any single request to Bloom. Generously more than the few seconds these take
+    // when healthy (a whole case is normally 12-25s), but comfortably under vitest's 120s testTimeout,
+    // so that when Bloom stops answering we report WHICH call hung instead of an opaque "test timed
+    // out" pointing at the test function, which is all BL-16612 gave us to go on.
+    const kRequestTimeoutMs = 45000;
+    // Staging a whole BloomPUB is the slowest thing we ask Bloom for, so it gets more room.
+    const kStagingTimeoutMs = 60000;
+    // A trivial read used in a polling loop; it should answer at once. Still generous, because a timeout
```
andrew-polk, line 253: [Devin] **Investigate**: kQuickReadTimeoutMs of 10s in the collection-ready poll could false-trigger the wedged short-circuit

A 10-second cap on the `isCollectionReady` poll is not just one case's timeout: exceeding it marks Bloom wedged and fails every remaining case. Switching to the collection tab reloads its webview, which on a loaded machine can legitimately take a while.

andrew-polk: (Claude Opus 5) Fair, and the asymmetry is the point: the cost of being too generous is waiting, while the cost of being too tight is a spurious abort of the whole suite. Raised to 30s in `dc98a17e0b`. Resolving.

src/BloomExe/Publish/OffScreenBrowser.cs
```diff
@@ -62,6 +68,25 @@ public sealed class OffScreenBrowser : IDisposable
...
         private const int kInitTimeoutMs = 20000;
 
+        // Longest we will block a caller waiting for work we posted to the dedicated thread. This is a
+        // BACKSTOP, not a normal budget: the operations that can legitimately take a while carry their own
+        // timeout (Navigate, and the browser creation inside StartFreshBrowser) and pass it plus a margin,
+        // so the only way to hit this is a WebView2 — or the message loop that services it — that has
+        // stopped making progress altogether. We used to block forever in that case, which wedged the
+        // calling thread permanently; when the caller was a BloomServer worker that wedged publishing with
+        // it, and the process never recovered (BL-16612).
+        private const int kDefaultBlockTimeoutMs = 30000;
```
andrew-polk, line 78: [Devin] **Investigate**: RunJavascript page-checks backstop is a fixed 30s regardless of book size

The backstop that bounds a script run is a constant 30 seconds, but the page-checks script walks every element carrying an id across the whole DOM. On a large enough book that work is not obviously bounded by 30s, and exceeding it now *fails the publish* rather than degrading it.

src/BloomExe/Publish/PublishHelper.cs
```diff
@@ -77,6 +81,105 @@ private OffScreenBrowser GetOrCreatePageChecksBrowser()
...
+                    {
+                        browser.StartFreshBrowser();
+                    }
+                    catch (OffScreenBrowserTimeoutException e)
+                    {
+                        Logger.WriteEvent(
+                            "Could not start a fresh page-checks browser either: " + e.Message
+                        );
+                    }
+                }
```
andrew-polk, line 175: [Devin] **Investigate**: New publish-blocking failure message is not localized

The new message that stops a publish -- "Bloom could not determine which parts of this book are visible..." -- is a hard-coded English literal, while neighbouring publish errors go through `LocalizationManager.GetString`. This repo's instructions also say localizable strings should automatically follow the `xlf-strings` skill.

(Claude Opus 5) Correct, and a fair catch against a standing project rule. I have not fixed it here because that skill explicitly requires asking which priority file a new string belongs in, and the wording of a user-facing error deserves your eye — so it is in the preflight decision report with a recommendation. Leaving this thread open until that comes back.

andrew-polk:
(Claude Opus 5) Consulted Devin on 2026-08-04 up to commit `dc98a17e0b`.

Three passes this run. **Every finding was real** — no false positives — and two of the three bugs were artifacts of my own commit-splitting surgery rather than the design:

- **The guard's documentation had come adrift**, left above two unrelated private counters when the method moved between commits, so the public method read as undocumented. Reattached.
- **The recovery attempt could escape uncaught.** On a failed page check we ask the browser for a clean renderer — but it is the *same unresponsive browser*, so `StartFreshBrowser` can hit its own backstop and throw. Sitting outside the try, that skipped the retry entirely and surfaced a technical timeout instead of the explanation written for users, in exactly the hang the retry exists for.
- **The blocked-count pairing rested on an argument rather than structure.** I had independently found the same leak and made the guard unable to throw once it registers; Devin's framing was better, so the register now happens *inside* the try with a conditional unregister. Both belts.
- Also raised the suite's quick-read timeout 10s → 30s: a timeout there does not fail one case, it marks Bloom wedged and fails every remaining one.

**Three flags are deliberately left open** for Andrew, all judgement calls rather than defects:

- the new publish-blocking message is not localized (this repo's rule says such strings follow the `xlf-strings` skill, which requires asking which priority file);
- the page-checks script backstop is a fixed 30s regardless of book size, and exceeding it now *fails* the publish;
- the pre-existing one about that refusal only reaching the user on the Publish-tab path.

Devin's final pass re-lists the three fixed bugs, but they are stale — I verified each fix is present at `dc98a17e0b` (doc attached to the method, `StartFreshBrowser` caught, `if (registered)` guarding the unregister). Its results are cumulative per PR rather than per-commit.

CI green. CodeRabbit is disabled for this repo by `.coderabbit.yml`.

src/BloomExe/Publish/PublishHelper.cs
andrew-polk: (Claude Opus 5) Andrew's decision: **leave it in English and record why**, which is what the `xlf-strings` guidance asks for when a string is deliberately not localized.

Done in `aed9b6e902`. The reasoning is now in the code: reaching this throw means the off-screen browser failed twice in a row, so it is a fault condition rather than a normal outcome, and such strings stay in English. Keeping the wording identical to the log entry just above it is also what lets a user's description match what we read in their log.

Resolving.

src/BloomExe/Publish/OffScreenBrowser.cs
andrew-polk: (Claude Opus 5) Andrew's response: **needs discussion, and 30s is certainly too low.** So no change yet -- leaving this open rather than picking a number.

Recording the shape of the discussion for whoever picks this up: the value is doing two different jobs. A *work* budget should scale with the book, since the script walks every element carrying an id; a *liveness* detector should be generous and fixed. It is currently being used as the latter while sized like the former.

The useful asymmetry is that bounding the wait *at all* is what fixed the disaster case. Now that it can never wait forever, the exact number only trades a spurious refusal against a slower failure -- and a spurious refusal is far worse. That argues for raising it substantially rather than tuning it finely.

Our only timing evidence is from the visual-regression books, which are tiny, so it says nothing about the large-book case. Logging how long page checks take on SUCCESS (we currently only log it on failure) would give a real distribution from the nightly and from user logs, and let the number be set from data.

src/BloomExe/Publish/PublishHelper.cs
andrew-polk: (Claude Opus 5) Andrew's response: **wants to discuss whether we want this change at all** -- is it belt and suspenders now that the server adds a worker, and it looks high risk.

Both fair. Two honest qualifications, so the discussion has the full picture:

1. It is not purely redundant. The guard addresses *starvation-caused* navigation failures; this refusal can also fire when the WebView2 renderer genuinely dies or wedges for reasons the pool cannot help with, and via the script backstop in the other open thread. So the path is not dead code -- just much rarer than before.
2. Against that: if it ever does fire on Save-to-file or Send-to-device, the user currently sees the operation simply stop, with the explanation going only to the log. That is a poor experience and the kind of thing that generates "it did nothing" support reports.

There is also a middle option I did not offer in the report, which changes the risk calculus: instead of wiring the message into four call sites, report once at the throw site via `NonFatalProblem.Report`, which the preview path already uses. One edit instead of four, covering every path. The wrinkle is that the preview path would then report twice, since its handler reports any exception generically -- so it would need a look, but it is a much smaller change than four separate progress-channel edits.

Leaving open pending that discussion.

PR #8111, opened 2026-07-27
Show a preview for images too large for the browser to display (BL-16597)

JohnThomson:
Chromium — and therefore WebView2 — refuses to decode an image whose pixel count times 4 bytes/pixel overflows a signed 32-bit int (about 536.9 megapixels). The `` fires `error` a moment after the bytes arrive and nothing paints, which is why the image chooser showed an empty preview pane for a 30000x23756 scan. Decoding at a reduced size doesn't help: `createImageBitmap()` with `resizeWidth` and the WebCodecs `ImageDecoder` with `desiredWidth` both fail identically, because the limit is tested against the image's natural size before any scaling is applied. And well below that hard ceiling, handing the renderer a hundred-megapixel image still costs seconds of decode and gigabytes of RAM to fill a preview pane a few hundred pixels tall.

So when a picked file is over 40 megapixels, `imageGallery/localFilePreview` now serves a downscaled JPEG instead of the original:

- WIC's `DecodePixelWidth`/`DecodePixelHeight` does the scaling as part of the decode, so a 713 megapixel JPEG costs about 2.5 seconds and under 100MB rather than the ~2.6GB a full decode would need.
- Any EXIF rotation is baked into the stand-in, since it is re-encoded without an EXIF block.
- The stand-in is generated once per picked file and reused, because the gallery asks for the same URL as both the thumbnail and the large preview. It is deleted when another file is picked, and when the project's lifetime scope is disposed.
- Anything WIC can't open (SVG, and other formats the browser handles fine anyway) falls back to serving the original.

`imageGallery/pickLocalImageFile` now also reports the original file's pixel dimensions and byte count, so the gallery describes the file the user actually chose rather than the downscaled stand-in it is being shown.

Tests cover reading dimensions (including the not-an-image fallback), the below-threshold "serve the original" case, and the downscaling itself.

Ref: https://issues.bloomlibrary.org/youtrack/issue/BL-16597

[Devin review](https://devinreview.com/BloomBooks/BloomDesktop/pull/8111)

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8111)

### DIFF
--- a/src/BloomBrowserUI/react_components/image-gallery/ImageGalleryDialog.tsx
+++ b/src/BloomBrowserUI/react_components/image-gallery/ImageGalleryDialog.tsx
@@ -101,16 +101,24 @@ const ImageGalleryDialog: React.FunctionComponent<{
             {},
         );
         if (!response) return undefined;
-        const { filePath, previewUrl } = response.data as {
+        const { filePath, previewUrl, width, height, size } = response.data as {
             filePath: string;
             previewUrl: string;
+            width: number;
+            height: number;
+            size: number;
         };
         if (!filePath) return undefined;
         return {
             thumbnailUrl: previewUrl,
             reasonableSizeUrl: previewUrl,
             localPath: filePath,
-            size: 0,
+            // C# reports the original file's dimensions and byte count. These matter because
+            // previewUrl may serve a downscaled stand-in for an image too large for the
+            // browser to display, and we want to report what the user actually chose.
+            width: width || undefined,
+            height: height || undefined,
+            size: size ?? 0,
             type: "image",
         };
     };
--- a/src/BloomExe/web/controllers/ImageGalleryApi.cs
+++ b/src/BloomExe/web/controllers/ImageGalleryApi.cs
@@ -7,6 +7,7 @@ using System.Net.Http;
 using System.Text.RegularExpressions;
 using System.Threading.Tasks;
 using System.Windows.Forms;
+using System.Windows.Media.Imaging;
 using Bloom.Api;
 using Bloom.Book;
 using Bloom.Edit;
@@ -15,6 +16,7 @@ using Bloom.MiscUI;
 using Bloom.Utils;
 using SIL.Core.ClearShare;
 using SIL.IO;
+using SIL.Reporting;
 using SIL.Windows.Forms.ClearShare;
 using SIL.Windows.Forms.ImageToolbox;
 
@@ -24,7 +26,7 @@ namespace Bloom.web.controllers
     /// Api for the image gallery (image chooser dialog) — local collections, file picker,
     /// remote search results, and saving chosen images into the current book.
     /// </summary>
-    public class ImageGalleryApi
+    public class ImageGalleryApi : IDisposable
     {
         public EditingView View { get; set; }
 
@@ -36,6 +38,42 @@ namespace Bloom.web.controllers
         /// only serves this exact file, preventing arbitrary local-file access via the endpoint.
         /// </summary>
         private string _lastPickedLocalImagePath;
+
+        /// <summary>
+        /// A temporary downscaled JPEG standing in for _lastPickedLocalImagePath when the
+        /// original is too big for the browser to display (see MakeBrowserSafePreview).
+        /// Null when the original is served as-is. Deleted when another file is picked.
+        /// </summary>
+        private string _lastPickedLocalImagePreviewPath;
+
+        /// <summary>
+        /// Guards _lastPickedLocalImagePreviewPath. The gallery uses the preview URL as both the
+        /// thumbnail and the large preview, so two requests for it typically arrive at once;
+        /// without this they would each spend seconds building a preview and one would leak.
+        /// </summary>
+        private readonly object _previewLock = new object();
+
+        /// <summary>
+        /// Chromium — and therefore WebView2 — flatly refuses to decode an image whose pixel
+        /// count times 4 bytes/pixel overflows a signed 32-bit int, i.e. anything larger than
+        /// about 536.9 megapixels. The &lt;img&gt; fires "error" a fraction of a second after
+        /// the bytes arrive and nothing paints, which is why the image chooser showed an empty
+        /// preview for a 30000x23756 scan (BL-16597). Decoding at a reduced size does not help:
+        /// createImageBitmap() with resizeWidth and the WebCodecs ImageDecoder with desiredWidth
+        /// both fail identically, because the limit is tested against the image's natural size
+        /// before any scaling is applied.
+        ///
+        /// Well below that hard ceiling, handing the renderer a hundred-megapixel image still
+        /// costs it seconds of decode time and gigabytes of RAM, all to fill a preview pane a
+        /// few hundred pixels tall. So past this threshold we substitute a downscaled JPEG.
+        /// </summary>
+        internal const long kMaxPreviewPixels = 40L * 1000 * 1000;
+
+        /// <summary>
+        /// Longest side, in pixels, of a generated preview. The gallery's preview pane caps the
+        /// image at 420px tall, so this leaves plenty of room for high-DPI screens.
+        /// </summary>
+        internal const int kPreviewMaxDimension = 1600;
 
         /// <summary>
         /// The root folder where SIL image collections (including Art of Reading) are installed.
@@ -367,12 +405,254 @@ namespace Bloom.web.controllers
                 )
             );
 
+            // Drop the previous pick's downscaled stand-in *before* authorizing the new path,
+            // so there is no instant in which a preview request for the new file would find,
+            // and serve, the previous file's stand-in.
+            DeleteLastPickedImagePreview();
             _lastPickedLocalImagePath = selectedPath;
+
             var previewUrl = string.IsNullOrEmpty(selectedPath)
                 ? ""
                 : "/bloom/api/imageGallery/localFilePreview?path="
                     + Uri.EscapeDataString(selectedPath);
-            request.ReplyWithJson(new { filePath = selectedPath, previewUrl });
+
+            // Report the *original* file's dimensions and byte count. The gallery displays
+            // these, and without them it would fall back to measuring whatever the preview
+            // turns out to be — which for a huge image is the downscaled stand-in, not the
+            // file the user actually chose.
+            var (width, height) = GetImageDimensions(selectedPath);
+            long size = 0;
+            if (!string.IsNullOrEmpty(selectedPath) && RobustFile.Exists(selectedPath))
+            {
+                try
+                {
+                    size = new FileInfo(selectedPath).Length;
+                }
+                catch (Exception e)
+                {
+                    Logger.WriteMinorEvent(
+                        $"ImageGalleryApi could not read the size of {selectedPath}: {e.Message}"
+                    );
+                }
+            }
+
+            request.ReplyWithJson(
+                new
+                {
+                    filePath = selectedPath,
+                    previewUrl,
+                    width,
+                    height,
+                    size,
+                }
+            );
+        }
+
+        /// <summary>
+        /// Reads an image's pixel dimensions without decoding its pixels. Returns (0,0) if the
+        /// dimensions can't be determined (e.g. an SVG, or a format WIC doesn't know).
+        /// </summary>
+        internal static (int width, int height) GetImageDimensions(string path)
+        {
+            if (string.IsNullOrEmpty(path) || !RobustFile.Exists(path))
+                return (0, 0);
+            try
+            {
+                var header = ReadImageHeader(path);
+                // Report the dimensions as they will actually be seen, i.e. after the viewer
+                // applies any EXIF rotation.
+                return IsQuarterTurned(header.exifOrientation)
+                    ? (header.height, header.width)
+                    : (header.width, header.height);
+            }
+            catch (Exception e)
+            {
+                Logger.WriteMinorEvent(
+                    $"ImageGalleryApi could not read the dimensions of {path}: {e.Message}"
+                );
+                return (0, 0);
+            }
+        }
+
+        /// <summary>
+        /// Reads an image's dimensions and EXIF orientation without decoding its pixels. WIC
+        /// parses the header up front but reads everything else lazily, so all the values we
+        /// want must be pulled out while the stream is still open — hence one method rather
+        /// than handing a BitmapFrame back to the caller.
+        /// </summary>
+        private static (int width, int height, int exifOrientation) ReadImageHeader(string path)
+        {
+            using var stream = RobustFile.OpenRead(path);
+            var decoder = BitmapDecoder.Create(
+                stream,
+                BitmapCreateOptions.IgnoreColorProfile | BitmapCreateOptions.DelayCreation,
+                BitmapCacheOption.None
+            );
+            var frame = decoder.Frames[0];
+            var exifOrientation = 1;
+            try
+            {
+                // 274 (0x112) is the EXIF "Orientation" tag.
+                if (
+                    frame.Metadata is BitmapMetadata metadata
+                    && metadata.GetQuery("/app1/ifd/{ushort=274}") is ushort orientation
+                    && orientation >= 1
+                    && orientation <= 8
+                )
+                    exifOrientation = orientation;
+            }
+            catch (Exception)
+            {
+                // Plenty of images carry no EXIF block at all; the default is what we want.
+            }
+            return (frame.PixelWidth, frame.PixelHeight, exifOrientation);
+        }
+
+        /// <summary>Whether the given EXIF orientation swaps width and height.</summary>
+        private static bool IsQuarterTurned(int exifOrientation) => exifOrientation >= 5;
+
+        /// <summary>
+        /// The rotation/flip an EXIF orientation calls for, or null when none is needed.
+        /// </summary>
+        private static System.Windows.Media.Transform GetOrientationTransform(int exifOrientation)
+        {
+            switch (exifOrientation)
+            {
+                case 2:
+                    return new System.Windows.Media.ScaleTransform(-1, 1);
+                case 3:
+                    return new System.Windows.Media.RotateTransform(180);
+                case 4:
+                    return new System.Windows.Media.ScaleTransform(1, -1);
+                case 5:
+                    var transpose = new System.Windows.Media.TransformGroup();
+                    transpose.Children.Add(new System.Windows.Media.RotateTransform(90));
+                    transpose.Children.Add(new System.Windows.Media.ScaleTransform(-1, 1));
+                    return transpose;
+                case 6:
+                    return new System.Windows.Media.RotateTransform(90);
+                case 7:
+                    var transverse = new System.Windows.Media.TransformGroup();
+                    transverse.Children.Add(new System.Windows.Media.RotateTransform(270));
+                    transverse.Children.Add(new System.Windows.Media.ScaleTransform(-1, 1));
+                    return transverse;
+                case 8:
+                    return new System.Windows.Media.RotateTransform(270);
+                default:
+                    return null;
+            }
+        }
+
+        /// <summary>
+        /// If <paramref name="originalPath"/> is too large for the browser to display (see
+        /// kMaxPreviewPixels), writes a downscaled JPEG to a temp file and returns its path;
+        /// otherwise returns null, meaning "serve the original".
+        ///
+        /// WIC's DecodePixelWidth/Height does the scaling as part of the decode, so a 713
+        /// megapixel JPEG costs about 2.5 seconds and under 100MB rather than the ~2.6GB a
+        /// full decode would need.
+        /// </summary>
+        internal static string MakeBrowserSafePreview(string originalPath)
+        {
+            try
+            {
+                var header = ReadImageHeader(originalPath);
+                long pixels = (long)header.width * header.height;
+                if (pixels <= kMaxPreviewPixels)
+                    return null;
+
+                // Constrain the longer side; WIC preserves the aspect ratio when only one of
+                // DecodePixelWidth/DecodePixelHeight is set, and does the scaling as part of
+                // the decode rather than decoding full size and shrinking afterwards.
+                // A stream rather than a UriSource, because Uri would treat a "#" in a
+                // perfectly legal filename as the start of a fragment. CacheOption.OnLoad
+                // makes EndInit() do the decoding, so the stream can be closed right after.
+                var scaled = new BitmapImage();
+                using (var stream = RobustFile.OpenRead(originalPath))
+                {
+                    scaled.BeginInit();
+                    scaled.StreamSource = stream;
+                    scaled.CacheOption = BitmapCacheOption.OnLoad;
+                    scaled.CreateOptions = BitmapCreateOptions.IgnoreColorProfile;
+                    if (header.width >= header.height)
+                        scaled.DecodePixelWidth = kPreviewMaxDimension;
+                    else
+                        scaled.DecodePixelHeight = kPreviewMaxDimension;
+                    scaled.EndInit();
+                }
+                scaled.Freeze();
+
+                // Bake in any EXIF rotation. We re-encode without an EXIF block, so a viewer
+                // has no orientation tag to apply and the pixels must already be upright.
+                BitmapSource upright = scaled;
+                var transform = GetOrientationTransform(header.exifOrientation);
+                if (transform != null)
+                {
+                    var rotated = new TransformedBitmap(scaled, transform);
+                    rotated.Freeze();
+                    upright = rotated;
+                }
+
+                // A plain temp path rather than SIL's TempFile, whose Dispose would delete the
+                // file we are about to start serving. DeleteLastPickedImagePreview cleans up.
+                var previewPath = Path.Combine(
+                    Path.GetTempPath(),
+                    "BloomImagePreview-" + Guid.NewGuid() + ".jpg"
+                );
+                var encoder = new JpegBitmapEncoder { QualityLevel = 85 };
+                encoder.Frames.Add(BitmapFrame.Create(upright));
+                using (var output = RobustFile.Create(previewPath))
+                    encoder.Save(output);
+
+                Logger.WriteMinorEvent(
+                    $"ImageGalleryApi made a {upright.PixelWidth}x{upright.PixelHeight} preview"
+                        + $" for {Path.GetFileName(originalPath)}"
+                        + $" ({header.width}x{header.height}, too large for the browser)"
+                );
+                return previewPath;
+            }
+            catch (Exception e)
+            {
+                // Includes SVG and anything else WIC can't open. Falling back to the original
+                // is right: those are the formats the browser handles fine anyway.
+                Logger.WriteMinorEvent(
+                    $"ImageGalleryApi could not make a downscaled preview for {originalPath}: {e.Message}"
+                );
+                return null;
+            }
+        }
+
+        /// <summary>Deletes the cached downscaled preview, if there is one.</summary>
+        private void DeleteLastPickedImagePreview()
+        {
+            lock (_previewLock)
+            {
+                if (string.IsNullOrEmpty(_lastPickedLocalImagePreviewPath))
+                    return;
+                try
+                {
+                    RobustFile.Delete(_lastPickedLocalImagePreviewPath);
+                }
+                catch (Exception e)
+                {
+                    // Nothing to do about it; it's a temp file, and it is named
+                    // recognisably enough to be swept up later.
+                    Logger.WriteMinorEvent(
+                        $"ImageGalleryApi could not delete {_lastPickedLocalImagePreviewPath}: {e.Message}"
+                    );
+                }
+                _lastPickedLocalImagePreviewPath = null;
+            }
+        }
+
+        /// <summary>
+        /// Picking a new file removes the previous file's stand-in, so at most one is ever
+        /// left over — the one for the last file picked. This is where that one goes; Autofac
+        /// disposes us with the project's lifetime scope.
+        /// </summary>
+        public void Dispose()
+        {
+            DeleteLastPickedImagePreview();
         }
 
         /// <summary>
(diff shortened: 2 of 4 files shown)


src/BloomExe/web/controllers/ImageGalleryApi.cs
```diff
@@ -397,7 +710,30 @@ private void HandleLocalFilePreview(ApiRequest request)
...
-            request.ReplyWithImage(fullPath);
+            // Images past a certain size don't render in the browser at all, so substitute a
+            // downscaled copy (BL-16597). What we worked out about a file is remembered and
+            // reused, since the gallery asks for this URL as both the thumbnail and the large
+            // preview — including the "this one is fine as it is" answer, which costs a header
+            // read to reach.
+            string pathToServe;
+            lock (_previewLock)
+            {
+                var answerIsAboutThisFile =
```
JohnThomson, line 823: [Devin] **Investigate**: Cached preview is not keyed to the authorized file; narrow race could serve wrong image

Devin's point: the check that says "you are only allowed to preview the file you just picked" happens *outside* the lock, while the cached downscaled stand-in is handed out *inside* it — and the cache itself carried nothing saying which file it was made from. So the guarantee that you get a stand-in for the right image rested on the order of two statements in `HandlePickLocalImageFile`, not on anything visible at the point of use.

Devin also judged it unreachable in practice, and agreed with that assessment: picking a file means working a modal Windows file dialog, and the gallery only asks for the preview URL after the pick has replied, so a request authorized for file A cannot realistically still be waiting on the lock when a whole new pick of file B completes. Its suggestion was to store the originating path alongside the cached preview and regenerate on mismatch.

Full analysis:

> In `HandleLocalFilePreview` the authorization check against `_lastPickedLocalImagePath` happens outside `_previewLock`, while the cached stand-in `_lastPickedLocalImagePreviewPath` is served under the lock without verifying it was generated for the currently-authorized file. In normal use this is safe: `DeleteLastPickedImagePreview()` runs synchronously in `HandlePickLocalImageFile` before the reply is sent, and the front-end only requests the preview URL after that reply, so any request for the new file arrives after the old stand-in is cleared, and a stale request for the old file would 403. A wrong-image serve would require a preview request to pass the auth check for file A, then a full new file pick (B) to complete its delete + `_lastPickedLocalImagePath` reassignment inside the few-microsecond gap before A's request acquires the lock — practically unreachable because picking a file requires interacting with a modal dialog. Flagging for awareness since the cache has no explicit file key; if the code is ever refactored so the auth check moves or picks can occur programmatically, this becomes a real staleness bug. A cheap hardening would be to store the originating path alongside the cached preview and re-generate if it doesn't match.

JohnThomson: (Claude Opus 5) Fixed in `e1237d202d8093647f5a26a997f34796ddc4c55d` — took the suggested hardening. The cache now remembers which file it was made from and regenerates on mismatch, so `HandleLocalFilePreview` establishes for itself that the stand-in belongs to the file being asked for, instead of relying on the ordering in `HandlePickLocalImageFile`.

That key also lets it remember the answer "this one is fine as it is", so an image small enough to serve unchanged is no longer re-examined on every request — which closes Devin's separate informational note about the serve-the-original path repeating its header read.

src/BloomExe/web/controllers/ImageGalleryApi.cs
```diff
@@ -367,12 +415,277 @@ private void HandlePickLocalImageFile(ApiRequest request)
...
+        /// megapixel JPEG costs about 2.5 seconds and under 100MB rather than the ~2.6GB a
+        /// full decode would need.
+        /// </summary>
+        internal static string MakeBrowserSafePreview(string originalPath)
+        {
+            try
+            {
+                var header = ReadImageHeader(originalPath);
+                long pixels = (long)header.width * header.height;
+                if (pixels <= kMaxPreviewPixels)
```
JohnThomson: [Devin] **Investigate**: Small TIFFs still won't render in the preview pane

Devin's point, and it checks out: the file chooser's filter accepts `.tif`/`.tiff` (`Images|*.jpg;*.jpeg;*.png;*.bmp;*.gif;*.tif;*.tiff;*.svg`), but no browser renders TIFF in an ``. This PR substitutes a JPEG only when an image is over the *size* threshold, so a big TIFF now previews fine while a modest one is still served verbatim and shows the same empty pane this PR exists to fix. Behaviour ends up inconsistent by file size.

Worth being clear about what is and isn't new here: TIFFs have never previewed in this chooser, so this is a pre-existing gap rather than a regression. What the PR changes is that it now *partially* covers it, by accident of size.

Flagged to the developer as a scope decision — whether making TIFFs preview belongs in BL-16597 or on its own card. Left open pending that decision.

Full analysis:

> `MakeBrowserSafePreview` only downscales (and re-encodes to JPEG) images over `kMaxPreviewPixels`; below that it serves the original bytes as-is, served at `:733`. The chooser filter accepts `.tif`/`.tiff`, but Chromium/WebView2 does not render TIFF in an ``, so a modest-sized TIFF is served verbatim (content type image/tiff via `ReplyWithImage`) and shows an empty preview — the very symptom this PR fixes for huge images. Large TIFFs now display because they get converted to JPEG, so behavior is inconsistent by size. This is a pre-existing limitation the PR only partially addresses; worth confirming whether small TIFF previews are expected to work.

src/BloomExe/web/controllers/ImageGalleryApi.cs
```diff
@@ -367,12 +415,277 @@ private void HandlePickLocalImageFile(ApiRequest request)
...
+        /// <summary>
+        /// If <paramref name="originalPath"/> is too large for the browser to display (see
+        /// kMaxPreviewPixels), writes a downscaled JPEG to a temp file and returns its path;
+        /// otherwise returns null, meaning "serve the original".
+        ///
+        /// WIC's DecodePixelWidth/Height does the scaling as part of the decode, so a 713
+        /// megapixel JPEG costs about 2.5 seconds and under 100MB rather than the ~2.6GB a
+        /// full decode would need.
+        /// </summary>
+        internal static string MakeBrowserSafePreview(string originalPath)
```
JohnThomson, line 682: [Devin] **Investigate**: WPF/WIC imaging used on server thread for the first time in BloomExe

Devin noticed that this is the only place in Bloom that uses WPF's WIC-backed imaging (`BitmapDecoder` / `BitmapImage` / `JpegBitmapEncoder`) — everything else goes through System.Drawing or the SIL image tools — and that the API server calls it from a thread-pool thread rather than the UI thread. It confirmed the code does the thing that makes that safe (freezing each bitmap before it leaves the method), and asked for a runtime sanity check that decode and encode really do work on the server's threads.

Full analysis:

> This is the only place in BloomExe that uses `System.Windows.Media.Imaging` (BitmapDecoder/BitmapImage/JpegBitmapEncoder); the codebase otherwise uses System.Drawing/SIL tools. WIC-backed WPF imaging is generally safe on background/MTA threads provided the bitmaps are frozen, which the code does (`scaled.Freeze()`, `rotated.Freeze()`). `UseWPF=true` in the csproj provides the assemblies. Given it's new territory, worth a runtime sanity check that decode/encode works on the API server threads across environments.

JohnThomson: (Claude Opus 5) Checked, and it works — covered by a test rather than a one-off run, in `407e275740ee386d43d4ac416a835ea108f44e84`.

Worth doing properly, because the obvious check would have missed the point: NUnit runs this project's tests on an STA thread (`BloomTests.csproj` sets the Apartment attribute), which is *not* the apartment the API server calls from, so the existing tests proved nothing about the real caller. `MakeBrowserSafePreview_WorksOnAnMtaBackgroundThread` runs the downscale on an explicitly MTA thread and checks the resulting JPEG. It passes, and it will keep being checked if the freezing is ever disturbed.

src/BloomExe/web/controllers/ImageGalleryApi.cs
```diff
@@ -367,12 +415,277 @@ private void HandlePickLocalImageFile(ApiRequest request)
...
+                    upright = rotated;
+                }
+
+                // A plain temp path rather than SIL's TempFile, whose Dispose would delete the
+                // file we are about to start serving. DeleteLastPickedImagePreview cleans up.
+                var previewPath = Path.Combine(
+                    Path.GetTempPath(),
+                    "BloomImagePreview-" + Guid.NewGuid() + ".jpg"
+                );
+                var encoder = new JpegBitmapEncoder { QualityLevel = 85 };
```
JohnThomson, line 733: [Devin] **Bug**: Preview of a very large transparent image shows a black background instead of the see-through areas

The stand-in is always written as JPEG, and JPEG has no transparency. So if someone picks a huge image whose background is see-through — a big line-art scan on transparency, say — the preview shows those areas as a solid black block instead. It only affects images over the 40-megapixel threshold (smaller ones are served unchanged and look right), and only ones that actually have transparency, so it is a narrow case; but within it the preview actively misleads about what is being chosen, which is the same kind of problem this PR set out to fix.

Real, and left open for the developer to decide on: preflight had already used up its budget of fix-and-re-review rounds by the time this surfaced, so it is written up in the preflight report rather than fixed here. The two obvious options are to composite onto white before encoding (one extra step, keeps every stand-in a small JPEG) or to write PNG when the source has an alpha channel (exact, but a 1600px PNG is a good deal bigger than the JPEG, and the temp file's extension has to follow so the right content type is served).

Full description:

> A very large image with see-through areas is re-saved for preview in a format that cannot keep transparency (`new JpegBitmapEncoder`), so its transparent regions turn solid (typically black) in the chooser preview. Impact: When someone picks a huge image that has transparent areas, the preview looks wrong — the transparent parts appear as a black block rather than transparent — which can mislead them about what they are choosing.
>
> Mechanism: Only images over the 40-megapixel threshold reach this path. For such an image, the decoded bitmap is re-encoded with `JpegBitmapEncoder`. JPEG has no alpha channel, so WPF discards transparency and the stored RGB (commonly 0,0,0) shows through, producing black where the original was transparent. A large transparent PNG would therefore preview very differently from how the browser renders the original. Compositing over white (or using a format that preserves alpha) would avoid the discrepancy. This is edge-case (huge + transparent), hence non-severe.

src/BloomBrowserUI/react_components/image-gallery/ImageGalleryDialog.tsx
```diff
@@ -101,16 +101,24 @@ const ImageGalleryDialog: React.FunctionComponent<{
...
             thumbnailUrl: previewUrl,
             reasonableSizeUrl: previewUrl,
             localPath: filePath,
-            size: 0,
+            // C# reports the original file's dimensions and byte count. These matter because
+            // previewUrl may serve a downscaled stand-in for an image too large for the
+            // browser to display, and we want to report what the user actually chose.
+            width: width || undefined,
+            height: height || undefined,
+            size: size ?? 0,
```
JohnThomson, line 121: [Devin] **Investigate**: Reporting the real (large) file size to the gallery may interact with any size-based gating

Full analysis:

> Previously `onPickLocalFile` returned `size: 0`; it now returns the real byte count. If the external `bloom-image-gallery` component applies any maximum-file-size validation/gating keyed on `size`, feeding it the true (potentially very large) size could now cause it to warn about or reject exactly the large images this PR intends to support. Worth confirming the gallery does not gate on `size`.

JohnThomson: (Claude Opus 5) Checked the gallery package — not an issue. `size` is only ever displayed: `ImageDetails.tsx` puts it through a formatter and shows it as e.g. "12M". There is no maximum-size check anywhere in `bloom-image-gallery` (no size limit, no warning, no rejection), so a large number cannot gate anything.

Reporting the real size is in fact what makes that display correct. When `size` was 0, `ImageDetails` fell back to a HEAD request against `reasonableSizeUrl` and showed *that* file's length — which for a huge image is now the downscaled stand-in, so the user would have been told their 300MB scan was a 200KB file. Same reasoning for the new `width`/`height`: without them the component measures the `` it rendered, i.e. the 1600px stand-in, instead of the real 30000x23756.

JohnThomson:
(Claude Opus 5) Consulted Devin on 2026-07-27, up to commit `407e275740ee386d43d4ac416a835ea108f44e84` — four reviews in all, one per pushed commit.

What it raised, and where each landed:

- **Transparent huge images preview as a black block** — real; left open above for John to decide how to encode the stand-in. The only finding still needing a person.
- **Small TIFFs still don't preview** — real, and pre-existing (TIFFs have never previewed here); left open above as a scope question, since making them work is new behaviour rather than part of BL-16597.
- **Preview cache wasn't keyed to the file it was made from** — fixed in `e1237d202d`; thread resolved.
- **WPF/WIC imaging on the server thread** — checked and now covered by a test that runs it on an MTA thread, `407e275740`; thread resolved.
- **Reporting the real file size might trip size gating in the gallery** — checked, no such gating exists; thread resolved.
- **EXIF orientation only read from JPEGs, not TIFFs** — fixed in `86057f3d1a`. Devin separately verified the orientation transforms (including transpose/transverse) as correct.
- Plus assorted informational notes; the ones worth acting on became the fixes above.

CI (`pr-automation`, `track / set-waiting`) is green. Greptile has not posted on this PR — nor on #8110, opened twenty minutes earlier by someone else — while every earlier PR today got a review within about three minutes, so it looks like Greptile is down rather than slow.

src/BloomExe/web/controllers/ImageGalleryApi.cs
JohnThomson: (Claude Opus 5) Fixed in `72c295973135e7c4e2f8d097d26a3a021170c9eb` — John chose compositing onto white.

The stand-in is now laid over a white background before it is encoded, so see-through areas come out white rather than showing whatever sits in the colour channels underneath (normally black). White is what the chooser and the page behind it both use, so the preview matches what the person will get.

Done unconditionally rather than only for images that carry transparency: deciding that reliably would mean enumerating every pixel format that can hold an alpha channel *and* checking indexed palettes for a transparent entry, whereas this is one pass over an image already capped at 1600px — milliseconds against the seconds the decode itself costs. It is pixel arithmetic rather than the WPF rendering stack, so it stays safe on the API server's threads.

The test for it was confirmed to fail without the fix (the corner pixel came out 0,0,0), so it is checking what it claims to.

src/BloomExe/web/controllers/ImageGalleryApi.cs
JohnThomson: (Claude Opus 5) Fixed in `72c295973135e7c4e2f8d097d26a3a021170c9eb` — John chose to fix it here rather than spin off a card.

What settled it: Bloom already applies the rule "convert what browsers can't draw so they can draw it" when *importing*. Checked by running the real import path on a TIFF — `PalasoImage` loads it, `ProcessAndSaveImageIntoFolder` classifies it as a non-web format, and `scan.tif` lands in the book folder as a valid `scan.png`; the comment there says exactly that. `BookFileFilter` likewise lists `.tif`/`.tiff` among the file types a book may contain. So a person could already put a TIFF in a book — they just had to choose it blind. The preview pane was the one place not applying the rule.

So the test for "does this need a stand-in?" has widened from size alone to size-or-format. Matched on extension rather than on the bytes, deliberately: `ReplyWithImage` derives the response's content type from the extension, so the extension is what the browser actually goes on. An image already smaller than the preview cap is no longer enlarged either — a stand-in made only because of format has no reason to grow. That also removes the oddity where whether a TIFF previewed depended on how many pixels it had.

src/BloomExe/web/controllers/ImageGalleryApi.cs
```diff
@@ -367,12 +416,370 @@ private void HandlePickLocalImageFile(ApiRequest request)
...
+                            scaled.DecodePixelHeight = kPreviewMaxDimension;
+                    }
+                    scaled.EndInit();
+                }
+                scaled.Freeze();
+
+                // Bake in any EXIF rotation. We re-encode without an EXIF block, so a viewer
+                // has no orientation tag to apply and the pixels must already be upright.
+                BitmapSource upright = scaled;
+                var transform = GetOrientationTransform(header.exifOrientation);
```
JohnThomson, line 719: [Devin] **Investigate**: EXIF orientation is applied manually; relies on WPF not auto-rotating

Two points, both fair. First, rotating the stand-in by hand is only correct if WPF's decoder doesn't already do it — otherwise a sideways photo would come out rotated twice. Devin's own reading was that WPF's imaging stack historically does not auto-orient, so the code is right, but nothing established that. Second, and more concretely: every image fixture in the tests was orientation 1, so none of the rotated paths were exercised at all — including the two `TransformGroup` cases (transpose and transverse), which are the fiddliest part of the change.

Full analysis:

> `MakeBrowserSafePreview` reads the EXIF orientation and bakes in a rotation via `GetOrientationTransform`/`TransformedBitmap`. This is only correct if WPF's `BitmapImage`/WIC decoder does NOT itself apply EXIF orientation. WPF's imaging stack historically does not auto-orient (unlike Windows Photo Viewer), so the manual application is correct and won't double-rotate. Worth noting: the unit tests only exercise orientation-1 (unrotated) TIFF/PNG fixtures, so none of the rotated paths (orientations 2-8, including the `TransformGroup` transpose/transverse cases) are covered by tests. A regression in the rotation math would not be caught.

JohnThomson: (Claude Opus 5) Covered in `d784d7f483ffab85d52afce85c85a582326d02a9`, and it settles both halves with one test.

`GetImageDimensions_AllowsForExifRotation` runs all eight orientations against a JPEG whose EXIF tag is set to each in turn — the fixture is built by splicing a hand-written APP1 segment in after the start-of-image marker, since System.Drawing can't construct a PropertyItem and pulling WPF into the test project just to write a fixture costs more than twenty bytes of header. A stored 400x200 must report 400x200 for orientations 1-4 and 200x400 for 5-8.

That also answers the auto-rotation question empirically rather than historically: if WPF were applying the orientation itself, the decoder would already be handing back swapped dimensions and our swap would undo it, so cases 5-8 would come back 400x200 and fail. They pass, so WPF is not auto-rotating and the manual application is not doubling up.

Note on what is and isn't covered: this pins down reading the tag and the width/height swap. The geometry of the transforms themselves — that orientation 5 really is a transpose and 7 a transverse — still rests on two independent algebraic derivations agreeing (yours and mine) rather than on a test, because asserting it needs WPF types in the test project. Worth knowing if that table is ever edited.

JohnThomson:
(Claude Opus 5) Consulted Devin again on 2026-07-27, up to commit `d784d7f483ffab85d52afce85c85a582326d02a9` — six reviews now, one per pushed commit.

Since the last log, John made the calls on the two findings that were waiting, and both are now fixed with their threads resolved:

- **Transparent huge images previewing black** — fixed in `72c2959731` by compositing onto white before the JPEG encode. The test for it was confirmed to fail without the fix.
- **TIFFs not previewing** — fixed in the same commit. The deciding argument was that Bloom already converts non-web formats to PNG on *import* (verified by running the import path: `scan.tif` lands in the book folder as `scan.png`), so the preview pane was the one place not applying a rule Bloom already has.
- **EXIF orientation applied by hand / rotated paths untested** — new this round, and now covered in `d784d7f483`. All eight orientations are tested against a JPEG with a hand-built EXIF header. That also settles Devin's "only correct if WPF doesn't auto-rotate" caveat empirically: if WPF were auto-rotating, the swap would be applied twice and cases 5-8 would fail.

One thing to be aware of when reading Devin's page directly: its last three jobs each **re-list findings that were already fixed**, quoting line numbers from before the fix — the transparency bug is still reported at `ImageGalleryApi.cs:630` (`new JpegBitmapEncoder`), which at this commit is line 725 and is wrapped in `FlattenOntoWhite`. It appears to accumulate findings across the PR's life rather than re-evaluating each against the current file. Every one of them has a resolved thread here recording the fix.

CI green. Greptile still silent on this PR — as on #8110 — so it has been recorded as timed out rather than clean.

src/BloomExe/web/controllers/ImageGalleryApi.cs
JohnThomson: (Claude Opus 5) Closing the loop on the residual gap noted above: John decided to leave it as is.

To be explicit about what that means, since this thread is otherwise recorded as fixed — the part Devin asked about *is* covered. Reading the orientation tag and swapping width and height are tested across all eight values, and WPF not auto-rotating is now established by those tests rather than assumed. What stays untested is only the geometry of the transform table itself: that orientation 5 is a transpose and 7 a transverse, rather than the two being swapped for each other.

The reasoning for leaving it: those two cases are the only ones in doubt, two independent derivations already agree on them, and getting a rotated photo wrong is the kind of thing anyone notices the instant they look at the preview. Testing it properly would mean making WPF types available in the test project, which is a build-configuration change better made on its own branch than folded into this one.

Worth re-reading if that table is ever edited.

src/BloomExe/web/controllers/ImageGalleryApi.cs
```diff
@@ -59,10 +108,18 @@ public void RegisterWithApiHandler(BloomApiHandler apiHandler)
...
+            // that whole time. Nothing here needs that lock: the only state shared between
+            // requests is the preview cache, and _previewLock already guards it (and still
+            // coalesces the gallery's near-simultaneous thumbnail and large-preview requests
+            // into a single build).
             apiHandler.RegisterEndpointHandler(
                 "imageGallery/localFilePreview",
                 HandleLocalFilePreview,
-                false
+                false,
+                requiresSync: false
```
JohnThomson, line 122: [Devin] **Investigate**: Preview generation runs while holding the server's global API lock

Devin's point, and it is correct. `RegisterEndpointHandler`'s fourth parameter, `requiresSync`, defaults to `true`, and this endpoint was registered with only three arguments — so it ran inside `BloomApiHandler`'s main `SyncObj`, the lock that stops most API requests being handled at the same time as each other.

What that meant in practice: this handler decodes and re-encodes the picked image inline, which takes about two and a half seconds for the 713-megapixel scan the tests use. For that whole time every other synchronized API request queues behind it, so choosing a very large image in the image chooser could briefly freeze unrelated parts of Bloom's UI. It is not a correctness problem and it cannot lose anyone's work, but it is a visible stall.

The endpoint is also not on the short list that `BloomApiHandler` routes to the separate `ThumbnailsAndPreviewsSyncObj`, so it really was taking the main lock rather than the previews one.

_Posted by Claude Opus 5 (1M context) on John's behalf, mirroring Devin's review of `202bc79b`._

JohnThomson: Fixed in `336da95e`: the endpoint is now registered `requiresSync: false`, with a comment saying why.

Nothing here needed the global lock. The only state shared between requests is the preview cache, and `_previewLock` already guards every read and write of it — including still coalescing the gallery's two near-simultaneous requests for the same file (it asks for this URL as both the thumbnail and the large preview) into a single build. Opting out is also the move the `i18n` endpoints already make, for the same reason and with a similar comment.

One consequence recorded rather than fixed: serving the file already happened outside `_previewLock`, and `pickLocalImageFile` still takes the global lock, so those two can now overlap where before they could not. If a new file were picked in the window between this handler choosing a path and streaming it, the temp stand-in could be deleted mid-serve and the preview would fail to load. Reaching that needs a modal file dialog to be opened and confirmed inside a file-serve window, and the worst outcome is a broken preview image, so it was judged not worth holding the lock across the whole response to prevent.

John reviewed this finding and chose this option over routing the endpoint to `ThumbnailsAndPreviewsSyncObj` or leaving it alone.

_— Claude Opus 5 (1M context)_

src/BloomBrowserUI/react_components/image-gallery/ImageGalleryDialog.tsx
```diff
@@ -101,16 +101,24 @@ const ImageGalleryDialog: React.FunctionComponent<{
...
         if (!filePath) return undefined;
         return {
             thumbnailUrl: previewUrl,
             reasonableSizeUrl: previewUrl,
             localPath: filePath,
-            size: 0,
+            // C# reports the original file's dimensions and byte count. These matter because
+            // previewUrl may serve a downscaled stand-in for an image too large for the
+            // browser to display, and we want to report what the user actually chose.
+            width: width || undefined,
```
JohnThomson, line 119: [Devin] **Investigate**: Reported dimensions are now the original's while the served pixels are the stand-in's

Devin's point: this now hands the gallery the original file's width and height while `previewUrl` may serve a ~1600px JPEG stand-in, so any code that used those numbers to *size* something would lay out a pane for a 30000×23756 image while the actual bitmap is 1600 wide. It flagged that it could not check this itself, because `bloom-image-gallery` isn't vendored in this repo.

**Not an issue — checked against the installed package.** `bloom-image-gallery` uses `IImage.width`/`height` in exactly two places, neither of them layout:

1. Deciding whether an image is too big for the browser to decode at all (`ImageDetails.tsx`, `width * height > 536870911`) — a comparison, which is precisely what wants the *original's* numbers.
2. Displaying the dimensions as text under the preview ("30000 x 23756").

Every `width=`/`height=` layout attribute in the gallery is a hardcoded constant — 24px for provider logos, 140px for search-result thumbnails — never derived from the image's reported size. The details pane's `` is sized purely by CSS (`max-height: 420px; max-width: 100%; object-fit: contain`), so it fits whatever bitmap arrives.

So reporting the original's dimensions is not merely harmless here, it is what the gallery wants: it makes the displayed size describe the picture the user actually chose rather than the downscaled stand-in, and it is what lets the gallery tell "too large to preview" apart from "this link is dead". Same conclusion as the neighbouring flag about `size`, which is likewise only ever displayed.

_Posted by Claude Opus 5 (1M context) on John's behalf, mirroring Devin's review of `336da95e`._

JohnThomson:
(Claude Opus 5 (1M context)) Consulted Devin on 2026-08-04, across the rebase — two jobs, up to commit `336da95eefbba7c42635bfafbe1475296bdd9853`.

The branch was rebased onto current `master` (it was 173 commits behind and had stopped being mergeable), so every commit has a new SHA. Devin re-reviewed twice: once on the rebased branch and once after the fix below.

Two findings were acted on, each with its own thread:

- **Preview generation held Bloom's global API lock.** Correct, and now fixed in `336da95e` — the endpoint is registered `requiresSync: false`, since its own `_previewLock` is what actually guards the shared preview cache. [Thread](https://github.com/BloomBooks/BloomDesktop/pull/8111#discussion_r3715373006)
- **Reported dimensions are the original's while the served pixels are the stand-in's.** Not an issue — Devin noted it couldn't check the gallery package; it is installed here, and it uses those numbers only for the too-large test and for display text, never for layout. [Thread](https://github.com/BloomBooks/BloomDesktop/pull/8111#discussion_r3715420068)

The rest of what it reported describes code this branch has already changed, and each already has a resolved thread from the 27 July run: the transparency bug (the JPEG encode is wrapped in `FlattenOntoWhite`), small TIFFs not rendering (any non-renderable extension gets a stand-in regardless of size), EXIF orientation being read only from the JPEG APP1 block (the TIFF query is tried too), and the `size` field possibly being gated on (it is only displayed). Notably the lock finding above was itself re-reported on the job that reviewed the commit fixing it, so Devin appears to accumulate findings over a PR's life rather than re-checking each against the current file — worth knowing before reading its page directly.

Also green at this commit: CI (`pr-automation`), the full C# suite (3014 passed, 0 failed, 13 skipped — the first entirely green full run on this branch, now that master's #8107 fixed the nine environmental failures), vitest (581 passed), typecheck and lint. No other review bot is configured on this repo.

andrew-polk:
I think I'm lost on the premise of this change...
The card says the image is pasted in from edit mode. And edit mode displays it. Then the image chooser is opened and doesn't display it. But the PR states the the problem is in wv2 displaying it. 
Why can't the image chooser display the image the same way edit mode does?

JohnThomson (commented):
[Claude Opus 5 (1M context)] Good question — the confusion dissolves once you notice that steps 2 and 3 of the card are looking at **two different files**.

**Step 2, pasting into Edit, imports the image.** Import resizes anything larger than 3840x2800 and saves the smaller copy into the book folder — the card predicts the result itself: "should shrink to about 3535x2800". So what edit mode displays is a ~10 megapixel copy, which WebView2 draws without complaint. That resize is also where the ~8 seconds in step 2 goes.

**Step 3, choosing that image in the chooser, points at the original file** still sitting in the Van Gogh folder: 30000x23756, about 713 megapixels. Nothing has touched it. WebView2 gives up somewhere above 537 megapixels (int32 max divided by 4 bytes per pixel), firing `error` instead of `load`, so the pane stays empty.

So both statements are true and not in tension: WebView2 genuinely cannot draw the original, and edit mode never asks it to — it only ever sees the imported copy. The image chooser is the one place in Bloom that displays a file *before* it has been imported, which is why this is the only place the problem appears.

On the duplication question, which is the fair one to ask: the server does already have image processing that resizes (`RuntimeImageProcessor` / `ImageUtils.AdjustImageForDisplay`, down to the same 3840x2800). Two things stop it covering this case:

1. It only runs for files that are already in the book folder — `processImage = !isSvg && sourceDir == CurrentBook?.FolderPath` in `BloomServer`. The chooser is previewing a file that is still just sitting on the user's disk, so it goes out untouched.
2. It opens the image with `PalasoImage.FromFileRobustly` before it can measure or resize it, which is a full decode — the 8 seconds you see on import, and about 2.6GB for this file. Doing that in the chooser would mean paying it just to *look* at a candidate the user may not even pick.

The new code's one distinctive move is asking WIC to scale during the decode (`DecodePixelWidth`), so the huge file never becomes a full bitmap in memory at all. So this isn't a second copy of the resize logic — it is the case that logic can't reach, minus the expensive part of it.

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8111#-:-OzDwvV5-NYdXMKFhCfl:bm5ren9)* status: 0 of 4 files reviewed, 8 unresolved discussions (waiting on JohnThomson).

andrew-polk (approved):
> @andrew-polk reviewed 4 files and all commit messages, and resolved 8 discussions.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8111)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on JohnThomson).

andrew-polk (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8111)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on JohnThomson).

PR #8112, opened 2026-07-27
Fix BL-16602: Talking Book highlighting dead after visiting a game page

JohnThomson:
## The bug

A game page requires the Game tool, so arriving on one auto-activates it and leaving one removes it again. When the removed tool was the active one, the React toolbox picked a replacement section to expand but **never told the legacy toolbox code**, which still owns each tool's `showTool()`/`hideTool()` lifecycle.

So legacy went on thinking the Game tool was current: `getActiveToolId()` kept returning `"game"` and the Talking Book tool never got `showTool()` again, leaving `AudioRecording.isShowing` false. Every highlight path is gated on `doesCurrentToolPlayAudio()`, so nothing was ever highlighted — and closing and reopening the Talking Book tool didn't help, because the desync outlived it. Only restarting Bloom cleared it.

This is a regression from 6.3, where the jQuery accordion did the equivalent for us: its `refresh()` noticed the active panel had been removed and activated another one, which fired the `activate` event that ran `switchTool()`. The React accordion that replaced it in 6.4 had no such path.

## The fix

Funnel every change of the expanded section through one `setActiveSection()` helper that both sets the React state and notifies the legacy listeners, and use it on the tool-removed path.

That also covers the same class of bug for any other way the active tool can be taken away — for instance unchecking the currently-active tool in "More...", which left the same dead state.

## Tests

`src/BloomBrowserUI/bookEdit/toolbox/ToolboxRoot.spec.tsx` (new) walks the add/activate/remove sequence through the `window.toolboxReactAdapter` surface and asserts the replacement tool is reported to legacy, plus that removing a *non*-active tool leaves the active tool alone. Verified it fails without the fix (`expected [ 'gameTool' ] to deeply equal [ 'gameTool', 'talkingBookTool' ]`) and passes with it.

## Note for the reviewer

Which tool becomes active after the Game tool is removed is "the first remaining section" — it does not remember that you were on Talking Book before you visited the game page. That matches 6.3 (jQuery picked the panel before the removed one, also not necessarily the one you came from), so this restores the old behavior rather than improving on it. Making it restore the pre-game tool would be a separate, larger change.

Ref: https://issues.bloomlibrary.org/youtrack/issue/BL-16602

[Devin review](https://devinreview.com/BloomBooks/BloomDesktop/pull/8112)

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8112)

### DIFF
--- a/src/BloomBrowserUI/bookEdit/toolbox/ToolboxRoot.tsx
+++ b/src/BloomBrowserUI/bookEdit/toolbox/ToolboxRoot.tsx
@@ -403,6 +403,29 @@ export const ToolboxRoot: React.FunctionComponent = () => {
     >([]);
     const hydratedToolIds = React.useRef<Set<string>>(new Set());
 
+    // Expand the given section and tell the legacy toolbox code about it.
+    // The legacy code keeps its own idea of which tool is current and drives each tool's
+    // showTool()/hideTool() lifecycle from it, so every path that changes which section is
+    // expanded has to go through here. A path that quietly changed only the React state left
+    // the two out of sync, so the tool the user could see was never activated: that is how
+    // visiting a game page killed Talking Book highlighting (BL-16602).
+    // Passing undefined means "nothing is expanded"; we deliberately don't notify legacy in
+    // that case, since it has no representation for "no current tool" and re-expanding the
+    // same section will notify with the same tool anyway.
+    const setActiveSection = React.useCallback(
+        (sectionId: string | undefined) => {
+            setExpandedSectionId(sectionId);
+            if (!sectionId) {
+                return;
+            }
+            const toolId = toToolboxToolId(sectionId);
+            activeToolChangedCallbacks.current.forEach((callback) => {
+                callback(toolId);
+            });
+        },
+        [],
+    );
+
     // Resolve body content for a section.
     // Prefer a live legacy element to avoid duplicate tool roots; otherwise
     // load static legacy HTML as a temporary fallback.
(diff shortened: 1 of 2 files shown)


src/BloomBrowserUI/bookEdit/toolbox/ToolboxRoot.tsx
```diff
@@ -680,24 +703,26 @@ export const ToolboxRoot: React.FunctionComponent = () => {
...
+
+            // The active tool has just been taken away from us (e.g. leaving a game page removes
+            // the Game tool, which that page required). Something else has to become active, and
+            // it must go through setActiveSection so that the legacy code actually activates it.
+            // (Before the React toolbox, the jQuery accordion's refresh did the equivalent, firing
+            // its activate event when the active panel disappeared.)
+            const replacementSection = sections.find(
+                (section) => section.id !== removedToolId,
+            );
+            setActiveSection(replacementSection?.id);
```
JohnThomson, line 725: [Devin] **Investigate**: Replacement tool may be the 'More...' settings pseudo-section

When the active tool is removed, `replacementSection = sections.find(section => section.id !== removedToolId)` (`src/BloomBrowserUI/bookEdit/toolbox/ToolboxRoot.tsx:722-724`) picks the first remaining section, which could be the always-present `settings` ("More...") section if no other real tool exists. That would notify legacy with `settingsTool` via `setActiveSection`. This matches the old behavior (`nextSections[0]` could also be settings), so it is not a regression, but if `settings` is not a legacy tool that responds to `showTool()`, activating it could be a no-op or worse. Worth a sanity check that this edge (removing the only real tool while only "More..." remains) behaves acceptably, though it is out of scope for this fix.

JohnThomson: [Claude Opus 5 (1M context)] Not an issue — the state Devin describes can't be reached, so there's nothing to sanity-check here.

For "More..." to be the tool we fall back to, every real tool would have to be gone from the toolbox at once. But the Talking Book tool can never leave: it's in `alwaysOnToolIds`, and the "More..." panel deliberately offers no checkbox for it (`SettingsToolControls.tsx` lists only Canvas, Decodable Reader, Image Description, Impairment Visualizer, Leveled Reader, Motion, Music and Sign Language as things you can turn off). The only other way a tool gets removed is the required-tool path, and the only tool that uses it is the Game tool.

So there is always at least the Talking Book section left, and because the sections are sorted alphabetically with "More..." forced to the end, `sections.find(...)` reaches Talking Book (or some earlier real tool) before it could ever reach `settings`.

Devin is right that this isn't a regression either way — the old `nextSections[0]` had exactly the same theoretical reach. Resolving.

JohnThomson:
[Claude Opus 5 (1M context)] Consulted Devin on 2026-07-27 14:35 up to commit `597fa34ff6166b27519973007f6ccd9a6ddce059`.

It found no bugs. It raised one Investigate flag — about whether "More..." could end up being the tool we fall back to when the active tool is removed — which I checked and answered on [its thread](https://github.com/BloomBooks/BloomDesktop/pull/8112#discussion_r3660134374); that state can't actually be reached, so the thread is resolved. It also left one informational note observing that the removal path is only correct because `sections` and `expandedSectionId` were added to the effect's dependency list, which is exactly what the code comment there says; nothing to do.

CI is green (`pr-automation`, `PR Review Tracker`). CodeRabbit is installed but has `auto_review` turned off in this repo's `.coderabbit.yml`, so it does not review here.

StephenMcConnel (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8112)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on JohnThomson).

PR #8113, opened 2026-07-27
Get the Playwright component-test harness running again

andrew-polk:
The Playwright "component tester" harness under `src/BloomBrowserUI/react_components/component-tester` could not run at all — every test died at import time. This branch gets the whole suite running again: **144 passed, 0 failed, 25 skipped**, stable across warm and cold-dev-server runs.

The breakage was pre-existing on master and independent of any feature work.

## Three independent blockers

**1. A stale React 17 pin.** The component-tester workspace pinned `react`/`react-dom` at 17.0.2 while the app is on 18.3.1, so Vite could not resolve `react-dom/client` — which `utils/reactRender.tsx` has imported since the React 18 migration. Because `component-harness.tsx` (the harness *entry point*) imports `renderRoot` from that module, this killed all 24 vite-served component tests, not just ones whose components happened to use it.

The pin was a missed spot in the migration, not a deliberate constraint: commit 6d1e449 ("@ upgrade to react 18") edited `component-harness.tsx` inside that very folder while leaving its `package.json` at 17, and nothing documents a reason. Bumped to 18.3.1 / `@types/react` 18.3.31 / `@types/react-dom` 18.3.7 to match the app.

**2. `playwright.config.ts` collected the wrong specs.** With `testDir: ".."` and `testMatch: "**/*.uitest.*"` it also picked up the `bloom-exe*.uitest.ts` specs, which belong to `playwright.bloom-exe.config.ts` and import the repo-level `playwright/test`. That loaded a second copy of Playwright and aborted the entire run with "Requiring @playwright/test second time" before a single test executed. This only manifests once `src/BloomBrowserUI/node_modules` exists — i.e. in the normal developer state — so the React fix alone would not have unblocked anything. Added a `testIgnore`.

**3. The harness's fake jQuery broke select2.** `component-harness.tsx` replaced `window.$`/`window.jQuery` with a stand-in object carrying only `Deferred`. Under Vite, select2 loads as an optimized ESM dep where `module` is undefined, so its UMD wrapper falls through to the browser-globals branch and does `factory(jQuery)` against `window.jQuery`. An object with no `.fn` makes it throw at import time, taking down every component whose graph reaches `StyleEditor` — everything under the toolbox. Replaced the stub with the real jQuery, which provides `Deferred` natively, so nothing was lost.

## Toolbox harness

`ToolboxRoot` only renders a section for a tool present in `getMasterToolList()`, and tools land there as a side effect of importing `toolboxBootstrap.ts`. The harness cannot import that module — it also does `$(document).ready(renderToolboxRoot)` and assigns `window.toolboxBundle`, which would double-render the root and clobber the stub one test installs — so `ToolboxRootTestHarness.tsx` now performs the same registrations itself. That duplicated list can drift; the proper fix (extract a side-effect-free `registerAllToolboxTools()`) is logged in `PAPERCUTS.md` and is probably best folded into the toolbox React refactor rather than done twice.

## Test reliability

Raised the Playwright `expect` timeout from 1000ms to 5000ms (matching `playwright.bloom-exe.config.ts`), with the overall test timeout raised 15s→30s in step so a genuinely broken assertion still reports as an assertion failure rather than a whole-test timeout. This is a retry budget, not a fixed wait, so passing runs cost nothing. At 1000ms roughly one test per full-suite run failed with "element(s) not found" — a different one each time, each passing in isolation.

Also made the registration tests' error-state assertion retry: `markedInvalid` did a one-shot `getAttribute("aria-invalid")`, which was safe under React 17's synchronous `ReactDOM.render` but races React 18's asynchronous `createRoot` commit. Replaced with `expectMarkedInvalid(expected)` across 31 call sites and removed the now-unused racy getter.

## Test assertions corrected

- `bookgridsetup-basic`: matched `/book in app/i`, but commit c7f4795 renamed the heading to "Books in App (%0)" to match `Bloom.xlf` without updating the test.
- `toolbox-root-react`: a precondition used a bare `getByText("Decodable Reader Tool")`, which also matches the tool's checkbox label in the "More..." panel; scoped it to the accordion headers via the file's existing `getToolHeader` helper.
- `toolbox-root-react`: split "header shows icons and subscription badges without chevrons". The valid half (tool names, no chevrons) is now a passing test; the rest is `test.fixme` with an explanation, because it asserts on `.subscription-badge` (emitted only by the legacy jQuery toolbox) and `.toolbox-react-header-icon` (a class that exists nowhere in the app), via computed `background-image` — which `src/BloomBrowserUI/AGENTS.md` tells us not to do. Re-enabling it needs a product decision about whether the React header renders badges and icons at all, so it was not forced green.

## Not in CI

These suites are still not run by CI — `nightly.yml` covers vitest, C# and visual-regression, and does not even install this workspace, which is why the harness rotted unnoticed. It is now green and stable enough to wire in; see `PAPERCUTS.md` for the suggested nightly job.

## Scope

No C# and no shipped front-end source is touched. `ToolboxRootTestHarness.tsx` is referenced only by the component-tester's dynamic `import.meta.glob`, so it never enters the app bundle.

[Devin review](https://devinreview.com/BloomBooks/BloomDesktop/pull/8113)

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8113)

### DIFF
--- a/PAPERCUTS.md
+++ b/PAPERCUTS.md
@@ -20,20 +20,49 @@ House rules:
 ---
 
 
-## 2026-07-27 — ToolboxRoot's Playwright component tests have been dead since React 18
+## 2026-07-27 — The component-tester uitest suites still aren't in CI
 
-- **Cut:** All 7 tests in `react_components/ToolboxRootTestHarness/component-tests/
-  toolbox-root-react.uitest.ts` fail on master before rendering anything: the component-tester
-  workspace pins `react-dom` 17.0.2, so Vite can't resolve `react-dom/client`, which
-  `utils/reactRender.tsx` has imported since the app moved to React 18. Nobody noticed because
-  these tests aren't in CI. That harness was the natural home for a regression test of the
-  toolbox's activation bridge (BL-16602); the test had to go into a vitest spec instead.
-- **Idea:** Bump `react`/`react-dom`/`@types/*` in
-  `src/BloomBrowserUI/react_components/component-tester/package.json` to 18 to match the app,
-  re-verify the 7 tests, and consider whether these should run in CI — a harness nothing runs
-  will rot again.
-- **Context:** BloomDesktop, found during `/preflight` of PR #8112 (BL-16602). Note the
-  component-tester README already says these aren't run by CI or any script.
+- **Cut:** Nothing runs `react_components/component-tester`'s Playwright suites — `nightly.yml`
+  runs vitest, C# and visual-regression only, and doesn't even `pnpm install` that workspace. That
+  is why the whole harness sat broken (React 17 pin + a second-Playwright-copy config bug) without
+  anyone noticing. It is now green at 144 passed / 25 skipped, so it will silently rot again.
+- **Idea:** Add a nightly job mirroring the visual-regression one (`pnpm install --frozen-lockfile`
+  in the component-tester dir, `pnpm exec playwright install chromium`, then
+  `pnpm exec playwright test --config playwright.config.ts`) and publish its junit output. The
+  bloom-exe config needs a running Bloom.exe, so only the component config is CI-able for now.
+- **Context:** BloomDesktop, branch `fix-component-tester`. The component-tester README still
+  says these are "not currently run as part of CI or other script" — update it if this lands.
+
+
+## 2026-07-27 — Changing harness imports + `reuseExistingServer` = 20 bogus test failures
+
+- **Cut:** Adding one bare import (`jquery`) to `component-harness.tsx` invalidated Vite's dep
+  optimizer, but `playwright.config.ts` sets `reuseExistingServer: true`, so the already-running dev
+  server served a half-rebuilt `node_modules/.vite/deps` chunk. Every BookGridSetup test (21 of
+  them) failed with `styled_default is not a function` — which looks exactly like a real MUI/emotion
+  regression and cost a chunk of debugging. Killing the server and deleting `node_modules/.vite`
+  fixed it with no code change.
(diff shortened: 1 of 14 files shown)


andrew-polk:
_(Claude Opus 5 (1M context))_

Consulted Devin on 2026-07-27 up to commit `6d90819f69b5dfa0a26556158803f47b4867cefc`.

**Re-review clean — bots quiet.** No Bugs and no Investigate flags. Devin raised 4 informational items, none of which needed a change:

- It confirmed the duplicated tool-registration list in `ToolboxRootTestHarness` matches `toolboxBootstrap.ts` today, and that the `getMasterToolList().length > 0` guard correctly survives a React-Refresh re-execution. The drift risk is already written up in `PAPERCUTS.md`.
- It noted that asserting the *absence* of a field's error state can't catch an error that appears later. True, but that is no weaker than the one-shot check it replaced, so nothing regressed.
- It confirmed the new `testIgnore` excludes only the `bloom-exe` specs and `node_modules`.
- It re-listed the unused `getMarkedInvalid` helper, which the very commit it analysed had already deleted — a stale item, no action.

An earlier pass on `0e211063d5` also came back with 0 Bugs / 0 Investigate; its one actionable informational item (that dead `getMarkedInvalid` helper, which was already unused on master and also read `aria-invalid` off the wrong element) is what `6d90819f69` removed.

CI is green (`pr-automation`, `track / set-waiting`). CodeRabbit does not run on this repo — `.coderabbit.yml` sets `auto_review.enabled: false`.

PR #8115, opened 2026-07-27
BL-16370: QR badge caption redesign, {bloom-badge-default} token, Local-Community + WorldVision badge layout

hatton:
## What this does

Three related pieces of the "Made with Bloom" QR badge work.

### 1. Badge caption redesign (commit 1 of the branch)

The badge caption changes from the single sentence
`Get more books in the {language} language on BloomLibrary.org.` to a two-line layout:
a caption line with the language name in **bold**, then the BloomLibrary.org address on its
own line in Bloom red.

- `BookStorage.UpdateQrCode` now takes the caption pattern and language name separately and
  builds the caption as two divs (`.bloom-blorg-caption` with the language in `<b>`, and
  `.bloom-blorg-url`). The block is rebuilt on every run, so existing books migrate on their
  next save.
- New localization id `QrCodeBadgeText2` = `More books in {0} language:`; the old
  `QrCodeBadgeText` has been deleted outright rather than kept with an obsolete note:
  it was added inside the 6.4 cycle, never shipped in a stable release, and existed only
  in the `en` file, so there were no translations to preserve. `QrCodeBadgeText2` is
  translatable, and its translator notes spell out that the caption must not contain a web
  address, because Bloom prints one on the line below.

### 2. `{bloom-badge-default}` token — badge markup lives in one place

`branding.json` files now write the token `{bloom-badge-default}` instead of pasting the
verbose `.bloom-branding-wrapper` badge HTML. `BookData.MergeBrandingSettings` expands it from
`BookStorage.BrandingBadgeHtmlByToken`, which sits next to `UpdateQrCode` because that method
rewrites exactly this structure at runtime. The token can appear anywhere in the content
string, so a branding can combine the badge with other markup. Documented in
`src/content/branding/AGENTS.md`.

- Converted **Default**, **WorldVision**, and **World-Vision-International** to the token.
- **Local-Community** gains the QR badge in the outside-back-cover *top* slot, with
  `branding.less` reworked so the badge sits at the top and the Bloom Local block stays
  anchored to the bottom (`min-height: 0` on the middle translation group so short/landscape
  back covers don't push the bottom block off the page).
- WorldVision / World-Vision-International: dropped the top-slot `img { width: 20% }` rule,
  which would otherwise override the badge wrapper's own logo (4.7cm) and QR (2.8cm) sizes.
- Copied `made-with-bloom-badge.svg` and `made-with-bloom-badge-text.svg` into each branding
  folder that uses the token (branding images must be local — see the Warning in
  `branding/AGENTS.md`).

### 3. CSS generic font families are left alone

The Local-Community badge caption asks for `serif`, a CSS generic keyword that the reader's
browser resolves itself. Bloom was treating it as a missing font: warning about it on publish
*and* rewriting it to Andika in the book's CSS. `PublishHelper.CheckFontsForEmbedding` now
skips CSS generic families (new `PublishHelper.IsGenericCssFontFamily`), so they are neither
reported nor substituted.

### Also on this branch: the `branding-report` dev tool

`.github/skills/branding-report` is a dev-only survey tool that drives a running Bloom to
capture one book across branding × page × layout × xmatter and build a filterable viewer.
It is supported by a **DEBUG-only** `collection/branding` POST endpoint in
`CollectionSettingsApi` (Release still throws) and small `Book.cs` hooks. Not shipped behavior.

## Tests

- `MergeBrandingSettings_WithBadgeToken_ExpandsInPlaceAndComposesWithOtherMarkup`
- `EmbedFonts_GenericCssFontFamily_LeftAloneAndNotReported`
- caption-builder tests in `BookStorageTests`

Ref: https://issues.bloomlibrary.org/youtrack/issue/BL-16370

[Devin review](https://devinreview.com/BloomBooks/BloomDesktop/pull/8115)

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8115)


src/BloomExe/Book/BookStorage.cs
```diff
@@ -2924,11 +2957,56 @@ string badgeLabel
...
+
+            var captionDiv = doc.CreateElement("div");
+            captionDiv.SetAttribute("class", "bloom-blorg-caption");
+            AppendCaptionWithBoldLanguage(captionDiv, badgeCaptionPattern, languageName);
+            label.AppendChild(captionDiv);
+
+            var urlDiv = doc.CreateElement("div");
+            urlDiv.SetAttribute("class", "bloom-blorg-url");
+            urlDiv.AppendChild(doc.CreateTextNode(displayUrl));
+            label.AppendChild(urlDiv);
```
hatton, line 2979: [Claude Opus 5] [Devin] **Investigate**: Users with an old-style custom caption will show the URL twice

If someone typed their own badge caption into Collection Settings (Advanced tab) — very likely copying the old default wording, "Get more books in the {0} language on BloomLibrary.org." — that text is stored verbatim and is still used as-is. With this PR the address also gets printed on its own line underneath the caption, so those users will see "BloomLibrary.org" twice on the back-cover badge: once inside their own sentence, and once again on the new URL line.

`CollectionSettings.BadgeQrCodeLabelLocalized` returns the stored custom caption unchanged when it is non-empty (CollectionSettings.cs:441-459), and `AdjustHtmlForHavingQrCode` unconditionally appends the `.bloom-blorg-url` line (BookStorage.cs:2976-2979). New/default collections are fine — the new default caption ends in a colon and deliberately has no address in it.

Related (Devin flagged this separately as informational, same area): the caption builder now substitutes only the **first** `{0}` in the pattern, where the old code used `string.Format` and replaced every occurrence. Only reachable with a hand-authored caption containing more than one `{0}`.

Leaving this thread open — deciding what (if anything) to do about existing custom captions is a user-facing behavior call for the developer, not something to change unilaterally.

DistFiles/localization/en/Bloom.xlf
```diff
@@ -541,6 +541,12 @@
         <source xml:lang="en">Get more books in the {0} language on BloomLibrary.org.</source>
         <note>ID: CollectionSettingsDialog.AdvancedTab.QrCodeBadgeText</note>
         <note>{0} will be replaced by a language name</note>
+        <note>obsolete as of 6.4; replaced by CollectionSettingsDialog.AdvancedTab.QrCodeBadgeText2 (the BloomLibrary.org URL is now shown on its own line beneath the caption).</note>
+      </trans-unit>
+      <trans-unit id="CollectionSettingsDialog.AdvancedTab.QrCodeBadgeText2">
+        <source xml:lang="en">More books in {0} language:</source>
+        <note>ID: CollectionSettingsDialog.AdvancedTab.QrCodeBadgeText2</note>
+        <note>The default caption on the "Made with Bloom" QR-code badge on a book's back cover. {0} is replaced by a language name (shown in bold). It ends with a colon because the BloomLibrary.org web address appears on the next line below it.</note>
```
hatton: [Claude Opus 5] [Devin] **Investigate**: New QrCodeBadgeText2 localization entry is not marked translate="no"

Not an issue — the omission is deliberate. `AGENTS.md` says to mark new XLF entries `translate="no"` "unless instructed otherwise", and this is one of those cases: `QrCodeBadgeText2` is text that readers see printed on the back cover of a book, and it directly replaces `QrCodeBadgeText`, which has always been translatable (and already has translations in the wild). Marking the replacement untranslatable would silently downgrade the badge to English-only for every non-English collection.

The surrounding user-facing caption strings in this file are likewise not marked `translate="no"`, and the branch commit message states the string is intentionally kept translatable. Resolving.

hatton:
[Claude Opus 5] Consulted Devin on 2026-07-27 up to commit `23f307488a245bb24668610bead17cb5c7281a35`.

No bugs. Two things it asked us to look at, plus three informational observations:

- **Custom badge captions can now show the address twice** — left open for a decision: [thread](https://github.com/BloomBooks/BloomDesktop/pull/8115#discussion_r3661538929).
- **The new XLF entry isn't marked `translate="no"`** — not an issue, that's deliberate for a string readers see on the back cover; explained and resolved: [thread](https://github.com/BloomBooks/BloomDesktop/pull/8115#discussion_r3661538991).
- Three informational notes (only the first `{0}` in a caption is substituted; the Local-Community `.bloom-translationGroup` rule matches every group in the back-cover margin box; and the badge-token expansion order relative to personalization is deliberate). Not mirrored as threads — low signal, and all three were checked.

CI is green, and the full C# and front-end test suites pass locally.

src/BloomExe/Book/BookStorage.cs
hatton: [Claude Opus 5] Decided: no code change, and here is why.

The QR-badge caption feature itself is new in 6.4 — the "Show QR Codes" switch and the Caption box (`ShowBlorgLanguageQrCode` / `BadgeQrCodeLabel`) were both added on 2026-03-11, in this same development cycle. So there is no released collection out there carrying a hand-written caption that this redesign could disrupt; the duplicated-address scenario would need data that does not exist yet.

What that leaves is making sure nobody *creates* the problem going forward — mainly translators, who are the ones supplying the default caption for every language. Handled in bcf8f54 (XLF notes only, no behavior change): the notes on `QrCodeBadgeText2` now say the string must stay short, that `{0}` appears verbatim and is rendered in bold, that the translation must **not** contain a web address because Bloom prints "BloomLibrary.org/language:xx" on its own line right beneath it, and they include the rendered two-line example. The obsolete `QrCodeBadgeText` now leads with DO NOT TRANSLATE, and the Caption box's own description carries a note explaining that Bloom appends the address itself.

The related informational point in this thread — that only the first `{0}` in a caption is filled in now, where the old code filled in every one — is also being left as is. It needs a hand-written caption containing two placeholders, which no shipped branding has.

Resolving.

hatton:
[Claude Opus 5] Consulted Devin again on 2026-07-28 up to commit `bcf8f5403448e32894c1d6fed27a98cdde79ceea`.

Re-review clean — nothing new to act on. No bugs. The same two items it raised before came back, and both already have threads on this PR with a documented outcome, now resolved: the [duplicated web address for custom captions](https://github.com/BloomBooks/BloomDesktop/pull/8115#discussion_r3661538929) (decided: no code change — the feature is new in 6.4, so there is no released data to migrate; the translator notes were tightened instead) and the [`translate="no"` question on the new XLF entry](https://github.com/BloomBooks/BloomDesktop/pull/8115#discussion_r3661538991) (deliberate — it is text readers see on a book's back cover).

Four informational notes, none needing action. One is new since last time: it points out that the World Vision back cover now gains a QR code where the top slot previously held a plain logo. That is the intended change, described in the PR body.

CI green.

src/content/branding/branding-base.less
```diff
@@ -111,6 +111,13 @@ Rules that format the branding go in branding-base.less, or in the branding.less
...
         // look strange with a much narrower one.
         max-width: 7cm;
+
+        // The BloomLibrary.org address sits on its own line beneath the caption
+        // text, in Bloom red to read as a link.
+        .bloom-blorg-url {
+            color: @bloom-red;
+            font-weight: bold;
+        }
     }
```
hatton, line 121: [Claude Opus 5] [Devin] **Investigate**: Caption text and URL now inside the `<a>` anchor

Devin's concern: the caption block sits inside the badge's link, so the caption line and the bold language name could pick up the browser's default link appearance — blue and underlined — since only `.bloom-blorg-url` is given an explicit colour.

Not an issue, for two reasons.

Bloom already switches link decoration off inside every branding slot. `basePage.less` (and its legacy twin `basePage-legacy-5-6.less`) carry:

```less
div[data-book*="branding"] {
    a:link,
    a:visited { color: unset; text-decoration: unset; }
}
```

The badge always lives in `outside-back-cover-branding-top-html` or `outside-back-cover-branding-bottom-html`, both of which match `data-book*="branding"`, so no underline and no blue. The `.bloom-blorg-url` rule then wins over the inherited colour for the address line specifically, which is exactly the intended two-tone look.

Second, this is not something the change introduced. The caption element `.bloom-lang-on-blorg` was already being appended inside the anchor before this branch — it is there in `origin/Version6.4`. What changed is only the structure *within* that element (two lines instead of one) and the address line gaining its red.

The tester notes on the card do ask for the badge to be checked in a PDF, in BloomPUB Viewer and in an ePUB reader, so if any reader somehow ignores that rule it will be caught by eye. Resolving.

hatton:
[Claude Opus 5] Consulted Devin on 2026-07-28 up to commit `2dbc671fae6c3fe6229fbba4b32b13d97518a14e`.

No bugs. One genuinely new item, now closed:

- **Caption text and URL now inside the `<a>` anchor** — Devin wondered whether the caption could show up blue and underlined. It can't: `basePage.less` already turns link colour and underline off inside every `data-book*="branding"` slot, which is where the badge always lives, and the caption element was inside the anchor before this branch anyway. Explained and resolved: [thread](https://github.com/BloomBooks/BloomDesktop/pull/8115#discussion_r3667310826).

The other two "investigate" items are the same two settled earlier and already resolved. Five informational notes, none needing action — including one confirming the generic-font skip correctly avoids the bad-font substitution path, and one noting the intended new QR code on the World Vision back cover.

CI green. Every review thread on this PR is resolved.

DistFiles/localization/en/Bloom.xlf
```diff
@@ -533,14 +533,18 @@
...
+        <source xml:lang="en">Bloom shows the BloomLibrary.org address on its own line beneath this caption, so you don't need to include it here. If your caption contains "{0}", Bloom will fill this in with the name of the language.</source>
         <note>ID: CollectionSettingsDialog.AdvancedTab.Caption.Description</note>
         <note>{0} must not be translated but appear verbatim.</note>
-      </trans-unit>
-      <trans-unit id="CollectionSettingsDialog.AdvancedTab.QrCodeBadgeText">
-        <source xml:lang="en">Get more books in the {0} language on BloomLibrary.org.</source>
-        <note>ID: CollectionSettingsDialog.AdvancedTab.QrCodeBadgeText</note>
-        <note>{0} will be replaced by a language name</note>
+        <note>This appears under the "Caption" box in Collection Settings > Advanced, explaining what the user may type there to replace the default QR-badge caption (CollectionSettingsDialog.AdvancedTab.QrCodeBadgeText2). "{0}" is shown to the user literally, as something they can type; it is not filled in here.</note>
+      </trans-unit>
```
hatton, line 540: [Claude Opus 5] [Devin] **Bug**: Old QR badge caption translation entry was deleted instead of kept and marked obsolete

Devin is right that `AGENTS.md` says to retain a superseded entry with an "obsolete as of <version>" note rather than delete it, and right that the PR body still claimed we had done that. The PR body has now been corrected.

But the stated impact — "existing translation work for the old caption string is discarded" — does not hold for this particular string, because there was never any translation work to discard.

`CollectionSettingsDialog.AdvancedTab.QrCodeBadgeText` was added on 2026-03-11 in commit 793aa32, inside the 6.4 development cycle. It has never appeared in a stable release. It also only ever existed in `DistFiles/localization/en/Bloom.xlf` — a search across all ~50 language directories under `DistFiles/localization` finds it in none of them. So deleting it drops nothing; it removes a string that no translator has ever been asked about.

The obsolete-note convention exists to protect translations that are already out in the world. With none in existence, leaving a dead entry behind only asks every translator to work out that they should skip it. John's call, made deliberately: delete it. Resolving.

DistFiles/localization/en/Bloom.xlf
```diff
@@ -533,14 +533,18 @@
         <note>The string labeled as Caption here is used in the QR Code badge to let the user know what the badge is used for.</note>
       </trans-unit>
       <trans-unit id="CollectionSettingsDialog.AdvancedTab.Caption.Description">
-        <source xml:lang="en">If your caption contains "{0}", Bloom will fill this in with the name of the language.</source>
+        <source xml:lang="en">Bloom shows the BloomLibrary.org address on its own line beneath this caption, so you don't need to include it here. If your caption contains "{0}", Bloom will fill this in with the name of the language.</source>
```
hatton, line 536: [Claude Opus 5] [Devin] **Bug**: Existing caption help-text translation was changed in place, orphaning prior translations

Same answer, and Devin quotes the rule correctly — but the rule's own condition is met here, so this is not a violation.

`AGENTS.md` says not to change an existing entry's content "unless it is new (marked `translate="no"`) or **you are sure that the change will not cause problems with existing translations**." We are sure, and it is checkable rather than a matter of opinion: `CollectionSettingsDialog.AdvancedTab.Caption.Description` was added on 2026-03-16 in commit a4a629b (BL-15987), within the 6.4 cycle, and it exists **only** in `DistFiles/localization/en/Bloom.xlf`. None of the ~50 language directories under `DistFiles/localization` contains it. There are no translations to orphan, so the stated impact — users in translated languages seeing stale text — cannot occur.

The reason for changing it in place rather than minting a new id: the sentence is what a user reads while typing their own badge caption, and its old wording was the reason someone would write "BloomLibrary.org" into the caption and then see the address printed twice. A new id would leave the misleading English live under the old id until every caller moved over. Since we are pre-release for 6.4 and free of translation debt on this string, fixing the words in place is the honest fix.

Both halves of the change stay in step: the XLF `<source>` and the English fallback in `AdvancedSettingsPanel.tsx` were updated together, so the fallback can't drift from the resource. Resolving.

hatton:
[Claude Opus 5] Consulted Devin on 2026-07-28 up to commit `5e0357015256de060d3164a2dbd52048a683936c`.

Two bugs, both about the localization changes, both answered and resolved:

- **[Old caption entry deleted rather than marked obsolete](https://github.com/BloomBooks/BloomDesktop/pull/8115#discussion_r3667416912)** — the rule is quoted correctly, but the impact it describes cannot happen: `QrCodeBadgeText` was added on 2026-03-11 inside the 6.4 cycle, never shipped in a stable release, and existed only in the `en` file. It appears in none of the ~50 language directories, so there were no translations to discard. Deleting it beats asking every translator to work out that they should skip it.
- **[Caption help text changed in place](https://github.com/BloomBooks/BloomDesktop/pull/8115#discussion_r3667417015)** — the rule's own exception applies ("or you are sure that the change will not cause problems with existing translations"), and it is checkable rather than a judgement call: `Caption.Description` was added 2026-03-16 and likewise exists only in `en`. Nothing to orphan.

Devin did catch one real thing inside the first item: the PR description still claimed the old string was "marked obsolete as of 6.4". That was stale, and the description has been corrected.

Three "investigate" items are the same ones settled earlier, all still resolved. Seven informational notes, none needing action. CI green; every thread on this PR is resolved.

StephenMcConnel (approved):
> @StephenMcConnel reviewed 27 files and all commit messages, and resolved 5 discussions.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8115)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on hatton).

PR #8116, opened 2026-07-27
Preserve CKEditor bookmark spans so the cursor doesn't jump in reader tools (BL-16490)

StephenMcConnel:
## Problem

When the Decodable (or Leveled) Reader tool is active, pausing while typing in a text box made the cursor jump to the **beginning** of the box.

## Cause

Pausing typing triggers a 500 ms-debounced re-markup pass that rewrites the editable box's `innerHTML`, which destroys the live caret. The caret is saved/restored via hidden CKEditor "bookmark" spans (`<span id="cke_bm_..." style="display: none;">`) inserted before the rewrite and re-selected (via `selectBookmarks`) after. But `removeCkEditorMarkup()` was deleting exactly those bookmark spans (`span[id^=cke_][style*='display: none;']`) during the markup pass — so afterward `selectBookmarks()` could not find them and the selection collapsed to the start of the box.

## Fix

Stop removing the hidden `cke_` bookmark spans in `removeCkEditorMarkup()`. It still unwraps CKEditor's background-color highlight spans, which was its other job. With the bookmarks left intact, the caret is restored to where the user was typing.

Updated the corresponding unit test (`removeCkEditorMarkup ... hidden cke_ spans`) to assert the bookmark spans are now **preserved**.

Ref: https://issues.bloomlibrary.org/youtrack/issue/BL-16490

[Devin review](https://devinreview.com/BloomBooks/BloomDesktop/pull/8116)

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8116)

### DIFF
--- a/src/BloomBrowserUI/bookEdit/toolbox/readers/libSynphony/jquery.text-markupSpec.js
+++ b/src/BloomBrowserUI/bookEdit/toolbox/readers/libSynphony/jquery.text-markupSpec.js
@@ -266,13 +266,16 @@ describe("jquery.text-markup", function () {
         expect($("#text_entry1").html()).toBe("<p>pre new text post</p>");
     });
 
-    it("removeCkEditorMarkup removes hidden cke_ spans", function () {
+    it("removeCkEditorMarkup preserves hidden cke_ spans (BL-16490)", function () {
+        // The hidden cke_ spans are CKEditor bookmarks marking the cursor position.
+        // We must leave them intact so the cursor can be restored after the text is
+        // marked up; removing them here made the cursor jump to the start of the box.
         const input =
             '<p>pre <span id="cke_1" style="display: none;">hidden</span> post</p>';
 
         $("#text_entry1").html(input).removeCkEditorMarkup();
 
-        expect($("#text_entry1").html()).toBe("<p>pre  post</p>");
+        expect($("#text_entry1").html()).toBe(input);
     });
 
     it("checkWrapWordsExtraIgnoresEmptyItems", function () {
--- a/src/BloomBrowserUI/bookEdit/toolbox/readers/libSynphony/jquery.text-markup.ts
+++ b/src/BloomBrowserUI/bookEdit/toolbox/readers/libSynphony/jquery.text-markup.ts
@@ -394,17 +394,12 @@ import { ReaderToolsModel } from "../readerToolsModel";
                     );
                     // CKeditor copies the highlight style as a barebones inline style containing
                     // only the background-color property.
-                    return (
-                        !!style &&
-                        /^background-color: [^;]*;$/.test(style)
-                    );
+                    return !!style && /^background-color: [^;]*;$/.test(style);
                 })
                 .contents()
                 .unwrap();
-            // remove spans belonging to CKEditor that are hidden.
-            $(this)
-                .find("span[id^=cke_][style*='display: none;']")
-                .remove();
+            // leave CKEditor-specific hidden spans that serve as bookmarks intact so that
+            // the cursor can be restored after marking up the text.  (BL-16490)
         });
     };
 


src/BloomBrowserUI/bookEdit/toolbox/readers/libSynphony/jquery.text-markupSpec.js
```diff
@@ -266,13 +266,16 @@ describe("jquery.text-markup", function () {
...
+        // The hidden cke_ spans are CKEditor bookmarks marking the cursor position.
+        // We must leave them intact so the cursor can be restored after the text is
+        // marked up; removing them here made the cursor jump to the start of the box.
         const input =
             '<p>pre <span id="cke_1" style="display: none;">hidden</span> post</p>';
 
         $("#text_entry1").html(input).removeCkEditorMarkup();
 
-        expect($("#text_entry1").html()).toBe("<p>pre  post</p>");
+        expect($("#text_entry1").html()).toBe(input);
```
StephenMcConnel, line 278: [Devin] **Investigate**: Old selector matched cke_ prefix broadly, not just cke_bm_ bookmarks

The deleted code removed any hidden span whose id starts with `cke_` (`span[id^=cke_]`), which is broader than the actual bookmark ids `cke_bm_` that the downstream cleanup (`restoreSelectionFromCkEditorBookmarks`, which only targets `span[id^='cke_bm_']`) will remove. The test uses `id="cke_1"` (not `cke_bm_`). If CKEditor ever inserts a hidden span with a `cke_`-but-not-`cke_bm_` id into editable content, it would now be preserved rather than stripped. The PR description asserts these hidden `cke_` spans only serve as bookmarks, so this is likely fine, but worth confirming no other hidden `cke_` spans end up in the editable content.

StephenMcConnel: (Claude Opus 4.8) Not an issue — assessed and leaving the code as-is.

Inside a `.bloom-editable`'s content, the only hidden (`display:none`) `cke_`-prefixed spans CKEditor ever inserts are the bookmark spans (`cke_bm_*`) it uses to mark the caret/selection. The other `cke_*` elements (`cke_top`, `cke_contents`, `cke_button`, the toolbar, etc.) are editor *chrome* that lives outside the contenteditable body, so they never reach `removeCkEditorMarkup()` — which only ever operates on the editable's own contents. So broadening from "remove all hidden `cke_` spans" to "remove none" only actually affects bookmark spans, which is exactly the intent of the fix.

The companion informational flag confirms those bookmark spans are cleaned up downstream (by a successful `selectBookmarks()`, or the fallback in `restoreSelectionFromCkEditorBookmarks`'s catch), so nothing leaks into saved content. Adding a speculative narrower `cke_`-but-not-`cke_bm_` removal would also run against this repo's "don't be overly defensive" guidance, since there's no case where such a span occurs here. Resolving.

StephenMcConnel:
(Claude Opus 4.8) Consulted Devin on 2026-07-27 up to commit `aff640a34330ca4c0a3bdffc572a40b02fcaed45`.

- **0 Bugs.**
- **1 Investigate flag** — "Old selector matched `cke_` prefix broadly, not just `cke_bm_` bookmarks." Assessed as not an issue and resolved (see [thread](https://github.com/BloomBooks/BloomDesktop/pull/8116#discussion_r3661572262)): the only hidden `cke_` spans inside editable content are the `cke_bm_` bookmarks, which are cleaned up downstream.
- **1 Informational** (not mirrored) — confirmed the fix is safe: the preserved bookmark spans don't leak into saved content because a successful `selectBookmarks()` (or its catch-block fallback) removes them after markup.

CI and the pre-commit gate are green.

PR #8117, opened 2026-07-28
MXB branding: add page sizes, QR back-cover badge, per-size tuning

hatton:
## Page sizes (BL request)

Adds Full/Half/Quarter Letter (portrait & landscape) and Device 16x9 (portrait & landscape) to
the `STARTLAYOUTS` of the MXB Literacy and Scripture xmatter packs, replacing the single
vertical half-letter option.

## Front-matter fit

The MXB packs pack a lot onto the title/credits pages, so this adds per-size tuning in the
shared `MXBCommon-XMatter.less`. The adjustments are **spacing-only** — reclaimed inter-section
margins, tightened leading on auxiliary blocks, top-flow on Letter — font sizes are unchanged
except on the two smallest credits pages (Quarter Letter / A6 / Device 16x9 Landscape), where
the content cannot fit at the normal size. Covers the title-page contributions/branding-logo fit
on the short sizes, the prepublication printing-history block, and the credits flex squeeze that
triggered Bloom's red overflow indicator on the quarter sizes.

Also adds `bloom-padForOverflow` to the MXB title-page fields, matching what the shared xmatter
mixins already do for the cover/title titles.

## Back-cover badge

Switches all three MXB brandings (Literacy, Literacy-Prepub, Scripture) from the old
"Made with Bloom.png" image to the QR-capable `.bloom-branding-wrapper` markup with
`made-with-bloom-badge-text.svg`, and drops the pack-level `img` width override that fought the
wrapper. Adds the Spanish translation of the existing QR badge label
(`CollectionSettingsDialog.AdvancedTab.QrCodeBadgeText`).

## Agent skills

Adds `custom-branding-text` (page sizes, the spacing-vs-font rules, the live overflow-measuring
loop, the real-Bloom capture pipeline, PDF assembly) and `bloom-branding`; documents the CDP
API/host-header gotchas in `bloom-automation`.

---

Base is `Version6.4`.

[Devin review](https://devinreview.com/BloomBooks/BloomDesktop/pull/8117)

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8117)


src/content/templates/xMatter/project-specific/MXBLiteracy-XMatter/MXBLiteracy-XMatter.less
```diff
@@ -4,11 +4,36 @@ whereas in a branding-induced xmatter pack it only acts to make this the default
 STARTLAYOUTS
 {
   "layouts": [
```
hatton, line 8: [Devin] **Investigate**: New STARTLAYOUTS restrict the dropdown despite the comment saying otherwise

The `STARTLAYOUTS` block comment in `MXBLiteracy-XMatter.less:2-3` (and the mirrored Scripture file) claims that in a branding-induced xmatter pack this only sets the default layout, but the new `custom-branding-text` SKILL explicitly notes the opposite: "Despite a code comment claiming a branding xmatter only sets a default, in practice this block restricts the dropdown." So replacing the single HalfLetterPortrait entry with the expanded set is what actually makes the new sizes selectable — the removed `HalfLetterPortrait`-only list would otherwise have been the only option. Worth confirming the full intended size set is present (Letter/HalfLetter/QuarterLetter portrait+landscape + Device16x9 portrait+landscape); note A6, which several new CSS rules target, is intentionally NOT in the layout list.

<sub>Mirrored from the Devin review by [Claude Opus 5 (1M context)] — Devin does not post to GitHub itself.</sub>

hatton: [Claude Opus 5 (1M context)] Verified the factual half of this during preflight:

- The `STARTLAYOUTS` JSON in both packs parses, and both list exactly the 8 sizes the change set out to add: `LetterPortrait`, `LetterLandscape`, `HalfLetterPortrait`, `HalfLetterLandscape`, `QuarterLetterPortrait`, `QuarterLetterLandscape`, `Device16x9Portrait`, `Device16x9Landscape`. All 8 are real Bloom page sizes (each is defined in `src/content/bookLayout/basePage-legacy-5-6.less`), so each one will actually render rather than falling back.
- Bloom reads this block out of the *compiled* CSS (`SizeAndOrientation.cs:161-183`), so I checked it after a real `lessc` run, not just in the `.less` source — it survives compilation intact and parses.
- The A6 point is correct but harmless: the `A6Portrait`/`A6Landscape` selectors were already in this file before this change and A6 has never been in the MXB layout list, so those rules are inert for MXB books. Nothing regressed.

Left open for the developer: whether to correct the file's opening comment, which still claims a branding xmatter pack's layout list "only acts to make this the default layout" when in practice it restricts the size dropdown. That comment is pre-existing boilerplate shared by many xmatter packs, so fixing it in these two files only would make the packs inconsistent — that's a judgment call rather than something to change unilaterally here.

hatton:
[Claude Opus 5 (1M context)] Consulted Devin on 2026-07-28 up to commit `0928507c46b3267ce51ddb83de28aff643dcdc44`.

Devin found **no bugs**. It raised one Investigate flag and six informational observations.

- **Investigate — "New STARTLAYOUTS restrict the dropdown despite the comment saying otherwise"**: mirrored as its own thread on `MXBLiteracy-XMatter.less` (https://github.com/BloomBooks/BloomDesktop/pull/8117#discussion_r3666634823). The factual part checks out and is answered there; one judgment call is left open for the developer, so that thread stays unresolved for now.
- **Informational (not mirrored as threads)**: two of them — the leaked temp profile directory and the dead `PAPER_SIZES` export in the branding-report harness — were real and are fixed in `0928507c46`; Devin's latest pass still lists them because it describes the branch's whole diff, not just the newest commit. The other four are observations rather than problems: the back-cover `img` width override was removed consistently from all three packs; the `"//"` pseudo-comment keys in `branding.json` follow the existing `Default/branding.json` convention and the C# deserializer ignores unmapped keys; the PDF renderer deciding success by whether the file exists rather than by exit code is deliberate and commented; and the overlapping `#contributions` margin rules are covered below.
- The one thing worth a developer's eye from the informational set: for Quarter Letter Landscape, `#contributions { margin-bottom: 0.5em }` is overridden a few rules later by `margin-bottom: 0`, so that first declaration never takes effect even though its comment explains why 0.5em was chosen. Harmless as rendered (0 is tighter, so the branding logo still fits), but misleading to a future editor. Raised in the preflight decision report rather than changed here, because altering which value wins would change the spacing that was already reviewed against live renders.

Other reviewers: CI checks pass (`pr-automation`, `track / set-waiting`). CodeRabbit is switched off for this repo (`auto_review.enabled: false` in `.coderabbit.yml`, and it has posted nothing on recent PRs), so it is not a participant here.

src/content/templates/xMatter/project-specific/MXBLiteracy-XMatter/MXBLiteracy-XMatter.less
hatton: [Claude Opus 5 (1M context)] Decided: the comment gets corrected, in these two MXB packs. Done in `1079f38083`.

It now says plainly that the layout list restricts the page-size dropdown even in a branding-induced xmatter pack — which is the observed behaviour — rather than claiming it only sets the default layout for a new book, and notes that a size has to be listed here to be selectable at all. The wider boilerplate in the other xmatter packs is left alone deliberately; correcting it everywhere is its own change, not something to fold into a branding PR.

No behaviour change: the layout list itself is untouched, and both packs still list the same 8 sizes verified earlier on this thread.

DistFiles/localization/es/Bloom.xlf
```diff
@@ -398,6 +398,12 @@ Luego suelte la tecla que apretó para mostrar esta lista.</target>
         <target xml:lang="es-ES" state="final">Importación/exportación de hojas de cálculo</target>
         <note>ID: CollectionSettingsDialog.AdvancedTab.Experimental.SpreadsheetImportExport</note>
       </trans-unit>
+      <trans-unit id="CollectionSettingsDialog.AdvancedTab.QrCodeBadgeText">
```
hatton, line 401: [Devin] **Investigate**: Spanish translation added directly to es/Bloom.xlf

This PR adds a new `<trans-unit>` translation directly to `DistFiles/localization/es/Bloom.xlf` (`DistFiles/localization/es/Bloom.xlf:401-406`). `AGENTS.md` states "When adding a new string, do not add it to all of the various language files. Just the one in the 'en' subdirectory." Since this is a translation of a string that already exists in `DistFiles/localization/en/Bloom.xlf:540-544` (the English source and notes match exactly), it is not strictly "adding a new string," so this is likely fine. However, if translations for this repo are normally managed through an external pipeline (e.g. Crowdin), manually adding an `es-ES` translation could conflict with that workflow. Worth a quick confirmation that hand-editing non-en xlf files is acceptable here.

<sub>Mirrored from the Devin review by [Claude Opus 5 (1M context)] — Devin does not post to GitHub itself.</sub>

hatton: [Claude Opus 5 (1M context)] Devin's hunch is right, and the repo documents the answer — so this one needs a developer decision rather than a dismissal. Leaving the thread open.

What I checked:

- The **English side is fine**: `CollectionSettingsDialog.AdvancedTab.QrCodeBadgeText` already exists in `en/Bloom.xlf:540-544` (it arrived in `793aa3252a`, not in this PR), and the source text and both notes in the new Spanish entry match it exactly. It is not marked `translate="no"`, so Crowdin does expose it to translators normally.
- The **Spanish side is what's in question**. `DistFiles/localization/README.md` says the Crowdin↔GitHub flow is one-way — Crowdin pulls the English sources and opens PRs for the translated files — and then: "Changes to translated Xliff files made outside of Crowdin ... must be uploaded manually to Crowdin for translators to see those translations. If this is needed, it should be done using the SILCrowdinBot account", and "Once the translation process is started on Crowdin for a given language, translation changes made outside of Crowdin are discouraged because it complicates merging changes made on Crowdin and it negates most of the value of using Crowdin to begin with."
- Spanish is very much a language already in translation (1308 entries in `es/Bloom.xlf` are `state="final"`), and this string currently exists in no other language file — the hand-added Spanish target is the only translation of it anywhere.

So the practical consequences: Spanish translators won't see this translation unless someone uploads it to Crowdin as SILCrowdinBot, and it's an out-of-band edit to a file Crowdin also writes. Neither is a build or runtime problem — the string will load and the badge will read correctly in Spanish — which is presumably why it was added this way, to get the MXB Spanish books their badge text now rather than after a translation round.

Raised in the preflight decision report for the developer to choose between keeping it plus uploading it via SILCrowdinBot, or dropping it and letting Crowdin produce it. Not reverting a deliberate translation on my own.

hatton:
[Claude Opus 5 (1M context)] Consulted Devin on 2026-07-28 up to commit `1079f380836f0747088987f95ea831bd099c829e`.

Still **no bugs**. Two Investigate flags and eight informational observations on this pass.

- **"New STARTLAYOUTS restrict the dropdown despite the comment saying otherwise"** — already mirrored earlier in this PR; the developer decided to correct the comment, that's done in `1079f38083`, and the thread is closed with the decision recorded.
- **"Spanish translation added directly to es/Bloom.xlf"** — new this pass, and it holds up: `DistFiles/localization/README.md` says translation changes made outside Crowdin are discouraged for a language already under translation and have to be uploaded via the SILCrowdinBot account to reach translators. Mirrored as its own thread and left **open** for the developer to decide between keeping it plus uploading it, or dropping it and letting Crowdin produce it. Nothing about it breaks the build or the runtime.
- **Informational**: all eight are ones already assessed on earlier passes. Note that Devin still lists the leaked PDF profile directory, the dead `PAPER_SIZES` export, and the overlapping `#contributions` margin rules even though all three are fixed (in `0928507c46` and `1079f38083`) — its analysis describes the branch's whole diff against the base, not just the newest commit.

CI is green at this commit (`pr-automation`, `track / set-waiting`), and CodeRabbit remains switched off for this repo.

hatton:
[Claude Opus 5 following a prompt from Hatton]

Superseded — this PR has been split in two, both rebased onto `master` (it was based on `Version6.4`):

- #8120 — `BL-13936-MXB-page-sizes` — only the Mexico (MXB) branding work, tied to [BL-13936](https://issues.bloomlibrary.org/youtrack/issue/BL-13936).
- #8121 — `branding-agent-skills` — the agent skills and the `branding-report` PDF fix written along the way. No YouTrack card; nothing here ships in Bloom.

The two file sets are disjoint and their union is exactly this PR's diff. Closing in favor of those.

PR #8122, opened 2026-07-28
Stop Bloom's UI font reaching book content, and give lang="*" text a real font (BL-16624)

andrew-polk:
Fixes the preview/thumbnail/printing font regression from `93afdc5696` at its source, closes the underlying gap that made it possible, and carries the fix through to ePUB. Supersedes #8125 and #8124 (see the bottom).

## The leak

`Book.AddPreviewJavascript()` injects `bookPreviewBundle.js` into three book documents. `bookPreview.ts` ended with `WireUpForWinforms(CollectionsTabBookPane)`, present only for the retired "legacy" collections tab, and that one line pulled React, MUI and `bloomUI.css` in with it — so `body { font-family: Roboto, NotoSans, sans-serif }` was being applied to rendered books. Harmless while the `@font-face` rules rode along in the same bundle; once `93afdc5696` moved them out, those documents named Roboto with nothing declaring it. Confirmed with `CSS.getPlatformFontsForNode`: the `.equals` element rasterized **Arial** before, **Roboto** after.

Deleting the dead line takes the preview's loader from ~60 chunk imports to 3. `CollectionsTabBookPane` is untouched and still used by `CollectionsTabPane.tsx`.

## The underlying gap

Only the arithmetic page showed it, because its editables are `lang="*"` — language-independent by design (BL-5616) — so no per-language rule matches them and they alone inherited the body font. `defaultLangStyles.css` now emits a rule for that content using L1's font.

**Scoped to content pages**, deliberately: `XMatterHelper` sets `lang=<metadata language>` on every xmatter page div precisely so its `lang="*"` fields (the ISBN, the branding-html blocks) inherit that font (BL-8545). An unscoped rule would beat that inheritance and visibly change brandings that put prose in those blocks. Verified in a running Bloom — the scoped selector matches **8/8** `lang="*"` elements on content pages and **0/12** on xmatter.

Font only (no direction, line-height or word-break), and L1's font rather than its own setting because `GetUsedFontsCommand` finds fonts via a regex whose language group never matches `*`, so a font used only there could go unembedded by the harvester.

This also fixes the Edit tab without shielding book content from `editMode.css`'s UI font: a directly-matching rule beats an inherited one.

## ePUB

ePUB export deletes every `lang="*"` attribute, because `*` is the wildcard from HTTP content negotiation and BCP 47 rejects it — which left this text as the one kind in an ePUB with no font of its own. Export now writes the resolved font into the element's `style` first. Content pages only, for the same BL-8545 reason, and it reads `CollectionSettings.Language1` — the same source the css rule uses — because `BookData.Language1` is documented as the hook for book-specific languages and the two are designed to diverge.

**Rewriting the attribute to a real language tag was tried and rejected.** It looks semantically free, since deleting the attribute already left the element inheriting that tag — but it makes language-neutral text *declare* itself to be in L1, so a book published with the vernacular turned off still contained divs tagged with it. `UserSpecifiedNoVernacular_VernacularRemoved` catches exactly that. Inheriting a language and declaring one are not equivalent to code that scans for declarations. An inline font makes no such claim, and still yields to the user's own font choice, which StyleEditor writes as `!important`.

## Test harness

The suite captures the `body` element rather than the whole page. A full-page capture's height includes the last `.bloom-page`'s escaping 15px bottom margin; when that changes, every case fails with "Image sizes do not match", which throws inside pixelmatch **before** any diff image is written.

The `@font-face` guard test previously enforced only that the declarations are never *duplicated*. Nothing enforced the other half — that every Bloom-managed document actually **gets** them — which is exactly how this bug arrived. The delivery points are now pinned, matched by module import so a namespace, dynamic or aliased import can't slip a new one past.

## Verification

Full C# suite **2970 passed**, 12 skipped, with one pre-existing environmental failure (`RabRealBuildTests` needs Reading App Builder installed). Front-end **571 passed**. Every new test was verified to fail without its half of the change.

**The visual regression suite is red on this branch (10 of 12) and on master (11 of 12) alike.** I built and ran pristine `origin/master` to confirm: same failures, same pixel counts (6307 / 91808 on 16x9, ~7225 / 92824 on A5), every differing pixel in the back-cover badge region — BL-16370's "Made with Bloom" QR badge and Local-Community layout landed without regenerated references. The one difference is the **288-pixel** arithmetic diff, present on master and fixed here. Those ten baselines are deliberately left for whoever owns BL-16370, rather than folding another change's rendering into this PR.

Ref: https://issues.bloomlibrary.org/youtrack/issue/BL-16624

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8122)

[Devin review](https://devinreview.com/BloomBooks/BloomDesktop/pull/8122)


andrew-polk:
[Claude Opus 5] Alternative minimal fix opened as #8125 — it links `bloomUIFontFaces.css` into the preview/thumbnail/printing DOMs instead of removing the leak, and changes no reference images. Only one of the two should land; see #8125's description for the trade-off.

andrew-polk:
[Claude Opus 5] Consulted Devin on 2026-07-28 16:23 USMST up to commit `e9b6ea4315e96b975d0b0fbcd2c679505d72c1aa`.

Result: clean — no Bugs and no Investigate flags. Two Informational items, both confirmations rather than problems: "Removed WinForms wiring is safe; React usage untouched" and "argosScreenshot element option for body capture".

CI is green (pr-automation, track/set-waiting). Full C# suite 2942 passed / 12 skipped, with one pre-existing environmental failure (`RabRealBuildTests` — Reading App Builder's buildapp.bat is not installed on this machine). Front-end suite 568 passed. Visual regression suite 12/12.

src/BloomExe/Collection/CollectionSettings.cs
```diff
@@ -496,6 +496,36 @@ public string GetCollectionStylesCss(bool omitDirection)
...
+            // script, and has nothing to say about digits and math symbols.
+            WritingSystem.AddSelectorCssRule(
+                sb,
+                ".bloom-page:not(.bloom-frontMatter):not(.bloom-backMatter) [lang='*']",
+                Language1.FontName,
+                Language1.IsRightToLeft,
+                -1, // suppresses line-height
+                false, // breakOnlyAtSpaces
+                true // omitDirection
+            );
```
andrew-polk, line 530: [Devin] **Investigate**: ePUB known-gap: lang='*' attributes stripped, so the new rule cannot apply

As noted in the PR description, `EpubMaker.RemoveSpuriousLinks` strips `lang="*"` attributes ("Validator doesn't like '*' as value of lang attributes"), so the new content-page-scoped `[lang='*']` font rule will not match in ePUBs and that text keeps its inherited font there. This is acknowledged as a known gap rather than fixed here; flagging in case ePUB rendering of arithmetic / language-independent text matters for this ticket.

---

[Claude Opus 5] Confirmed and deliberately left open for the developer to decide. Our own review found the same thing independently. Closing it would mean either teaching EpubMaker to keep (or translate) `lang="*"`, which touches ePUB validation, or emitting an equivalent ePUB-side rule keyed on something that survives the strip. Both are user-facing changes to published output and out of scope for this PR, so the choice is going to the developer in the preflight decision report rather than being made here.

andrew-polk:
[Claude Opus 5] Consulted Devin on 2026-07-28 18:26 USMST up to commit `649f08effe96cf40efb1676c8e3a2d45243dda62`.

No Bugs. One Investigate flag — the ePUB gap, mirrored as its own thread above and deliberately **left open** for a developer decision; it is already documented as a known gap in the PR description.

Five Informational items, all confirmations rather than problems: that removing the `WireUpForWinforms` line was dead code and the component is still used elsewhere; that the scoped rule correctly omits direction; that the `body`-element screenshot option is valid; and one genuinely useful observation — that because the new rule's selector no longer starts with `[lang='`, the preservation scan never matches it, so the `"*"` guard in `CreateOrUpdateDefaultLangStyles` now only protects against a legacy bare rule rather than the one we write. That matches our own reading; the guard is kept as defence for books that already received an unscoped rule from an interim build.

CI green (pr-automation, track/set-waiting). CodeRabbit does not run on this repo (`.coderabbit.yml` sets `auto_review.enabled: false`). Visual regression 12/12; full C# suite 2962 passed with one pre-existing environmental failure (`RabRealBuildTests`); front-end suite 568 passed.

src/BloomExe/Collection/CollectionSettings.cs
andrew-polk: [Claude Opus 5] Now fixed rather than accepted, in `f7696580fb`. `RemoveSpuriousLinks` still deletes the `lang="*"` attribute — validators do reject it — but it now leaves `bloom-languageIndependent` behind, and the generated rule matches either form. So the ePUB gets the same font as the preview, print and BloomPUB, and the markup still says truthfully that the text is in no language (rewriting the attribute to L1 would have claimed otherwise, and misled screen readers and hyphenation). Resolving.

src/BloomExe/Publish/Epub/EpubMaker.cs
andrew-polk, line 1: [Devin] **Investigate**: ePUB font rule relies on page scoping surviving export

`RemoveSpuriousLinks` stamps `bloom-languageIndependent` onto former `lang='*'` elements so the defaultLangStyles rule can reach them. That rule is scoped to `.bloom-page:not(.bloom-frontMatter):not(.bloom-backMatter)`. The test asserted only that the class is applied, not that the font actually resolves in the exported ePUB — i.e. that the `.bloom-page` / xmatter classes and defaultLangStyles.css are still present and scoped as expected after export.

---

[Claude Opus 5] Fair, and fixed in `3424d3d3fc`. The test now walks the whole chain rather than stopping at the class: the `lang` attribute is gone, the class is on the equation, the equation is still inside a `.bloom-page` that is neither front nor back matter (so the scoping survives export), and `defaultLangStyles.css` is present in the ePUB carrying a selector for the class. Any one of those four failing would have left the class as inert decoration, which was exactly your point. Resolving.

andrew-polk:
[Claude Opus 5] Consulted Devin on 2026-07-28 19:37 USMST up to commit `3424d3d3fc581431f0628a39cc04617f0284fad1`.

No Bugs. Two Investigate flags, both now **fixed and their threads resolved** rather than accepted:

- *ePUB known-gap: lang='*' attributes stripped* — no longer a gap. Export leaves `bloom-languageIndependent` behind and the rule matches either form.
- *ePUB font rule relies on page scoping surviving export* — a fair catch on the test, not the code: it proved the class was applied but not that anything styled it. The test now walks the whole chain.

Eight Informational items, all confirmations or observations we agree with — including that the `"*"` guard in `CreateOrUpdateDefaultLangStyles` now only protects legacy files (the scoped selector no longer starts with `[lang='`), that the rule is deliberately invisible to harvester font detection, and a line-endings nit in the generated selector that is fixed in `3424d3d3fc`.

CI green. CodeRabbit does not run on this repo (`auto_review.enabled: false`). Full C# suite 2963 passed with one pre-existing environmental failure (`RabRealBuildTests`); front-end 569 passed; visual regression 12/12.

src/BloomExe/Publish/Epub/EpubMaker.cs
andrew-polk, line 1: [Devin] **Investigate**: Font source for inline ePUB rule (BookData.Language1) vs defaultLangStyles (CollectionSettings.Language1)

The inline ePUB font is read from `Book.BookData.Language1`, while the `[lang='*']` rule it stands in for is built from `CollectionSettings.Language1`. These return the same value today, but they are not the same accessor.

---

[Claude Opus 5] Good catch, and fixed in `05696ac108`. `BookData.GetLanguage` documents itself as "the crucial method for book-specific language setting" that currently just "defers to the collection settings" — so the two are *designed* to diverge, and the ePUB would have quietly disagreed with every other surface the day book-specific languages land. Both now read `CollectionSettings.Language1`, so that becomes one decision to revisit rather than a latent mismatch. Resolving.

src/BloomVisualRegressionTests/index.spec.ts
andrew-polk, line 1: [Devin] **Investigate**: Switching capture target to body may require regenerating all preview reference images

The visual-regression suite now screenshots the `body` element rather than the whole page. That changes the captured box, so the committed reference images may all need regenerating.

---

[Claude Opus 5] Checked rather than assumed. Only the six A5 Portrait references needed regenerating; the six 16x9 Landscape ones came out **byte-identical** to what was already committed and are untouched by this PR. The reason is that the two boxes only differ by the last `.bloom-page`'s escaping 15px bottom margin, which is exactly what we no longer want to capture, and 16x9's content ends flush. The 72 bloom-player references are unaffected — the change is only to the preview capture. Resolving.

andrew-polk:
[Claude Opus 5] Consulted Devin on 2026-07-29 15:02 USMST up to commit `05696ac108af60167beb64afffab50e245feeeb5`.

No Bugs. Four Investigate flags, **all four now fixed and their threads resolved** rather than accepted:

- *Font source: BookData.Language1 vs CollectionSettings.Language1* — a genuinely good catch. Same value today, but `BookData.GetLanguage` is documented as the hook for book-specific languages that currently just defers to the collection; they are designed to diverge. Both now read `CollectionSettings.Language1` (`05696ac108`).
- *Switching capture target to body may require regenerating all references* — checked: only the six A5 references needed it; the six 16x9 ones came out byte-identical and are untouched.
- *ePUB relies on page scoping surviving export* — obsolete: there is no longer a rule to reach in the ePUB, the font is written inline.
- *ePUB known-gap: lang='*' stripped* — fixed, not accepted.

Ten Informational items, all confirmations or observations we agree with.

CI green. CodeRabbit does not run on this repo (`auto_review.enabled: false`). Full C# suite **2970 passed**, 12 skipped, with one pre-existing environmental failure (`RabRealBuildTests` — Reading App Builder not installed). Front-end **569 passed**.

**Visual regression is red on this branch (10 of 12), and it is red on master too — 11 of 12.** Verified by building and running pristine `origin/master`: the same failures with the same pixel counts (6307 / 91808 on 16x9, ~7225 / 92824 on A5), all confined to the back-cover badge region. They come from BL-16370's "Made with Bloom" QR badge and Local-Community layout work, whose reference images were not updated. The one difference is that master also shows the **288-pixel** arithmetic diff that this PR fixes, and this branch does not. Deliberately not regenerating those ten baselines here — that would absorb someone else's rendering change into this PR.

src/BloomExe/Publish/Epub/EpubMaker.cs
```diff
@@ -3693,9 +3693,50 @@ private void RemoveSpuriousLinks(HtmlDom pageDom)
...
+                {
+                    var existing = elt.GetAttribute("style");
+                    var fontRule = $"font-family: '{font}'";
+                    elt.SetAttribute(
+                        "style",
+                        string.IsNullOrEmpty(existing)
+                            ? fontRule
+                            : existing.TrimEnd().TrimEnd(';') + "; " + fontRule
+                    );
+                }
```
andrew-polk, line 3739: [Devin] **Bug**: Language-neutral text in an ePUB can be told to use a font that was never packaged with the book

The font name is written directly onto the element (`elt.SetAttribute("style", ...)` at `src/BloomExe/Publish/Epub/EpubMaker.cs:3733`) in a place the later "unusable font" clean-up never looks, so text that was given a font the book is not allowed to package keeps naming it and falls back to whatever the reading device happens to have.

Impact: In the rare book where the language-neutral text (equations and similar) uses a font that cannot be embedded, that text still renders in an arbitrary device font instead of Bloom's substitute font.

<details><summary>Why the bad-font substitution misses the new inline style</summary>

After all pages are written, `EpubMaker` (`src/BloomExe/Publish/Epub/EpubMaker.cs:2866-2875`) replaces references to non-embeddable fonts: `PublishHelper.FixCssReferencesForBadFonts` rewrites `font-family: X;` inside `defaultLangStyles.css`/`customCollectionStyles.css` (`src/BloomExe/Publish/PublishHelper.cs:1902-1925`), and `FixXhtmlReferencesForBadFonts` → `PublishHelper.FixXmlDomReferencesForBadFonts` (`src/BloomExe/Publish/PublishHelper.cs:1947-1984`) rewrites only the `userModifiedStyles` `<style>` element in each xhtml file. Neither touches `style` attributes on elements. The new code writes `font-family: '<font>'` into exactly such an attribute, using the browser-measured font (which is the user's own choice, precisely the case most likely to be an unlicensed/unavailable font). So after substitution the rest of the book points at the default font while this element still points at the bad one, which is not in `fonts.css`.

A fix would be to skip (or substitute) the inline font when the font ends up in `badFonts`, e.g. by extending `FixXmlDomReferencesForBadFonts` to also scan `//*[@style]` for `font-family`.

</details>

andrew-polk: [Claude Opus 5] Fixed in `b31539ab7f`, and Devin is right that this was reachable in practice — the measured font is the user's own choice, so it is the likeliest of all our font references to be one we cannot package.

`FixXmlDomReferencesForBadFonts` now also scans elements' `style` attributes, so the inline declaration takes part in the same substitution as every other font reference. The matching there is deliberately looser than the stylesheet version: an inline declaration has no `!important` and often no trailing semicolon, and the font name may be quoted or bare. BloomPub gets the repair too, since it calls the same method.

Two tests in `PublishHelperStaticTests`: one that a bad font inline is swapped for Andika while a different font, an element with no font, and the rest of the style attribute are all left alone; one that with no bad font present the method reports no change, since callers only save the file when it says it changed something. Verified the first fails without the fix.

src/BloomExe/Publish/PublishHelper.cs
andrew-polk, line 1: [Devin] **Bug**: Publishing a book crashes Bloom because the font-cleanup step calls itself forever

The font substitution step for inline styles called itself as its very first action (`FixInlineStyleReferencesForBadFonts`) instead of doing its work, so every publish immediately recursed until the process died.

Impact: Publishing to ePUB or BloomPub crashes Bloom with an unrecoverable stack overflow every time.

<details><summary>Unconditional self-recursion in the new inline-style font fixer</summary>

`FixXmlDomReferencesForBadFonts` now unconditionally calls the new private helper `FixInlineStyleReferencesForBadFonts`. That helper's first statement re-invoked itself with the identical arguments — a copy/paste of the caller's initialization line — so there was no base case and no termination condition. Result: `StackOverflowException`, which .NET cannot catch, terminating the process.

`FixXmlDomReferencesForBadFonts` is public and used by both the ePUB and BloomPub publishing paths, so this fired regardless of whether any bad font or inline style was actually present.

The fix is to delete the recursive call; `fixedSomething` should simply start as `false`.

</details>

andrew-polk: [Claude Opus 5] Fixed in `041d580b8e` (the previous commit `b31539ab7f` was amended away, so this never reaches master).

Devin's diagnosis is exactly right, including the cause: restoring the file after a non-vacuity check, I replaced `var fixedSomething = false;` by string match, and that line existed in *both* the caller and the new helper — so the helper got the caller's delegating call and became infinitely recursive.

The full C# suite caught it independently at the same time: the run aborted with `Test host process crashed : Stack overflow` and the frame repeated 4602 times. Worth noting how it presented, because it is a trap — with the run aborted, `dotnet test` still printed a green-looking `Passed! Failed: 0, Passed: 1124` line. Only the total (1129 against the usual 2986) and the `Test Run Aborted.` line at the very end gave it away.

Now `fixedSomething` starts as `false` in the helper, as Devin suggests. Full suite at `041d580b8e`: 2973 passed, 12 skipped, 2986 total, no abort; the one failure is the pre-existing `RealReadingAppBuilderBuild` environmental test that also fails on master.

andrew-polk, line 1: [Claude Opus 5 relaying Devin] **Bug**: Unusable font can be left in place, or a font name mangled, when cleaning inline font declarations

The unembeddable-font substitution in inline style attributes matched the font name without any end boundary, so an unquoted longer name that merely starts with the bad name was partly rewritten, producing a broken declaration.

Impact: Text in a published book can end up with a nonsense font instruction and render in an unexpected font.

<details><summary>How the unanchored pattern misfires on names that share a prefix</summary>

Bloom really does ship font families whose names are prefixes of others ("Andika" / "Andika New Basic"). If "Andika" is in `badFonts` and an element's style attribute says `style="font-family: Andika New Basic"` (unquoted), the regex matched just `font-family: Andika` and the result was `font-family: 'Andika' New Basic`, an invalid declaration. The pre-existing CSS-file variants avoid this because they require a trailing `;`.

Conversely, adding a plain `;`/end anchor would miss fallback lists (`font-family: Andika, serif`), so the boundary should allow `;` or `,` or end of string.

Note this method now runs for BloomPub too, i.e. against every style attribute in the book, not just the one ePUB export writes.

</details>

andrew-polk: [Claude Opus 5] Fixed in `fc42d94b50`, and the suggested boundary is exactly the one used: a bare name must now end at a semicolon, a comma, or the end of the attribute, while a quoted name is already bounded by its quotes. The comma case matters for the reason given — a bad font may head a fallback list, and the list has to survive.

New test `FixXmlDomReferencesForBadFonts_BadNameIsPrefixOfAnother_LeavesTheLongerNameAlone` uses precisely the Andika / Andika New Basic pair: it asserts the longer family is untouched, the exact name is replaced, and `Andika, serif` becomes `'Substitute', serif`. Verified it fails with the old pattern, producing the `font-family: 'Substitute' New Basic` corruption described here.

src/BloomExe/Collection/CollectionSettings.cs
andrew-polk, line 1: [Claude Opus 5 relaying Devin] **Bug**: Custom book and collection styles can no longer change the font of language-independent text

The new font rule for language-independent text is written with a much stronger selector than the per-language rules (`.bloom-page:not(.bloom-frontMatter):not(.bloom-backMatter) [lang='*']`), so a book's or collection's own custom style for that text is no longer honoured.

Impact: Books that set a font for equation-style or similar language-neutral text through their custom styles will silently render it in the collection's main-language font instead.

<details><summary>Specificity beats the documented override path</summary>

The generated file's own header states that these rules "may be over-ridden by rules in customCollectionStyles.css, appearance.css or customBookStyles[2].css". The per-language rules honour that: `[lang='xyz']` has specificity 0,1,0, tying with a typical `.Equation-style { font-family: X; }` in customBookStyles.css, which wins on load order.

The new rule has specificity 0,4,0 (three classes plus an attribute selector), so it beats any such custom rule regardless of order. Only declarations marked `!important` (which is what StyleEditor writes) can still override it. Before this change, `lang="*"` text inherited its font, so a custom `.Equation-style` font-family did apply.

A lower-specificity way to exclude xmatter (or explicitly resetting the font for xmatter instead of scoping the positive rule) would keep the documented override behaviour.

</details>

---

**Status: left open deliberately, awaiting the developer's decision.**

Confirmed independently by a second reviewer in the same run, and verified against the code: `BookStorage.OrderedPrefixesOfCssFilesToSortAfterUnknownStylesheets` does load `defaultLangStyles.css` before `customCollectionStyles.css` / `customBookStyles.css`, so at equal specificity the custom file wins — and this rule is not at equal specificity.

Not fixed here because each remedy carries its own user-visible trade-off, and picking between them is a design call rather than a clear-cut correction:

1. **Leave it.** L1's font wins over hand-authored custom CSS for language-independent text. Affects only books whose custom CSS sets a font on a style that language-independent text uses; StyleEditor's own choices are unaffected, as they carry `!important`.
2. **`:where(...)` around the page scoping**, dropping the rule to 0,1,0 so it behaves exactly like its per-language siblings. Clean, but `:where()` is unsupported on older WebViews, and an unsupported selector invalidates the *whole* rule — so on an old Bloom Reader the text would silently lose its font again, reintroducing the original bug on the platform we can least observe.
3. **Bare `[lang='*']` plus a separate xmatter reset** (`.bloom-frontMatter [lang='*'] { font-family: inherit }`). Restores the override path for content pages with no new selector syntax, but the reset itself is 0,2,0 and so would newly outrank custom CSS inside xmatter.

src/BloomExe/Publish/PublishHelper.cs
andrew-polk, line 1: [Claude Opus 5 relaying Devin] **Investigate**: Bad-font substitution now sweeps every style attribute in BloomPub output too

`FixXmlDomReferencesForBadFonts` is shared by ePUB and BloomPub. The new inline-style pass therefore also rewrites `font-family` declarations in pre-existing `style` attributes anywhere in a BloomPub book (canvas elements, origami panes, etc.), not just the one ePUB export writes. That is probably desirable, but it is a behaviour change beyond the stated ePUB scope and worth confirming that no inline `font-family` in BloomPub content is intentionally left naming a non-embedded font.

andrew-polk: [Claude Opus 5] Correctly spotted, and intended. Keeping it shared is the point rather than an oversight: `badFonts` means "this book does not carry this font", and that is equally true whichever package the reader opens. A BloomPub that names a font it does not ship gets the same arbitrary device fallback an ePUB would, so exempting BloomPub would preserve the very bug this is fixing, on the more widely used format.

On whether anything relies on the old behaviour: the substitution only ever touches a font already in `badFonts`, i.e. one publishing has just decided it cannot embed and has already stripped from `fonts.css` and rewritten out of every stylesheet and the `userModifiedStyles` block. An inline declaration left naming it would not have rendered in that font either — it would have fallen back to whatever the device had. So there is no case where the old behaviour delivered the author's intent and the new one does not; the change replaces an unpredictable fallback with Bloom's chosen default.

The blast radius is also smaller than it looks: the pass is a no-op unless `badFonts` is non-empty (`EmbedFonts` only calls it then), and within that, only for style attributes that actually contain `font-family`.

src/BloomTests/Publish/PublishHelperStaticTests.cs
andrew-polk, line 1: [Claude Opus 5 relaying Devin] **Investigate**: PR claims a fully green C# suite, but the new inline-style test cannot pass

`PublishHelperStaticTests` exercises `FixXmlDomReferencesForBadFonts`, which (at commit `b31539ab7f`) always reached the self-recursing helper. That means the reported "2970 passed" run cannot include this code, so the inline-style substitution feature is entirely unverified. After removing the recursion, please re-run at least `PublishHelperStaticTests` and the ePUB export tests.

andrew-polk: [Claude Opus 5] Fair challenge, and the reasoning is right, though the specific claim it targets never reached the PR: the "2970 passed" figure was from the commit *before* the recursion existed, and the commit that introduced it (`b31539ab7f`) has been amended away.

What actually happened is worth recording, because the failure mode is a trap. The recursion did not make the tests fail — it crashed the test host, and `dotnet test` still printed `Passed! Failed: 0, Passed: 1124`. Only the total (1129 against the usual 2986) and a `Test Run Aborted.` line at the very end showed anything was wrong. A green-looking summary line is not evidence of a green suite unless the total matches.

Both suggested suites have since been re-run against the fixed code, plus everything else. Full C# suite at `fc42d94b50`: results in the preflight report linked on the card; `PublishHelperStaticTests` (3 tests) and the ePUB `LanguageIndependentText` tests (4) all pass, and each new test was individually verified to fail when its half of the change is disabled.

StephenMcConnel (approved):
> @StephenMcConnel reviewed 18 files and all commit messages, and resolved 10 discussions.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8122)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on andrew-polk).

PR #8123, opened 2026-07-28
Highlight decodable/leveled reader violations without touching the DOM (BL-16558)

JohnThomson:
Fixes BL-16558: quickly replacing text in a Decodable Reader could leave inaccurate — and permanent — highlighting in the book.

## Cause

The decodable and leveled reader tools marked violations (a word that is not decodable at this stage, a sentence too long for this level, a word too long) by rewriting the `innerHTML` of the contenteditable to wrap the offending text in a styled span. CKEditor responds to that by copying the span's computed `background-color` into a bare inline style of its own, which then gets saved into the book.

## Fix

Paint the violations with `::highlight()` pseudo-elements over live `Range`s instead — the technique the Talking Book tool already uses (BL-15300). The DOM is never modified, so there is nothing for CKEditor to copy. As with the Talking Book highlight, the paint is a thick offset underline rather than a background color: drawn behind the glyphs, it covers far less of the previous line when lines are tightly spaced.

Because the analysis no longer has to *produce* HTML, it no longer has to work *in* HTML. `mapVisibleText()` snapshots an element's visible text plus where each character came from in the DOM; the word/sentence analysis runs on that plain string; and the results are highlighted by character offset. That removes the HTML reassembly, the leading-`</span>` nesting hack, the format-button stash/restore, and `removeAllHtmlMarkupFromString`'s role in markup.

## Commits

Deliberately sequenced so the mechanical parts are easy to verify:

1. **Move** the audio highlight manager to shared `bookEdit/js/textHighlightManager.ts` — pure rename, no logic change.
2. **Split** the generic highlight core out of the audio-specific half (`audioHighlightManager.ts`) — behavior-preserving; Talking Book tests pass untouched.
3. **The fix** — reader tools highlight without touching the DOM.
4. Unit tests for the shared primitives.
5. Only show a reader hover tip when the mouse is really over the text.
6. Stop removing CKEditor's selection bookmarks (converges with 323b1b6674 on master).

## Also in here

- Marking of "possible words" is gone. It has had neither a color nor a tooltip for years (both were commented out), so it only ever added noise to the DOM.
- A long word is now marked **everywhere** it appears on the page. The long-word list was always cumulative, but the old code wrapped each leaf as it went, so a word first seen in the second paragraph was left unmarked in the first.
- The hover tips move from per-span qtips to a hit test against the painted ranges. This restores two tips that were in fact broken: `markDecodableStatus` called `qtip()` on a raw DOM element rather than a jQuery wrapper, so the "Sight Word" and "not decodable" tips threw instead of appearing. No new localized strings.
- `removeSynphonyMarkup`/`removeCkEditorMarkup` are retained: existing books contain leftover spans, and CKEditor's copies of their background color, that still need cleaning out. That is what repairs already-damaged books.

## Not done here, deliberately

- `doMarkup` still does the CKEditor `getData()`/bookmark dance around the markup. It no longer has an `innerHTML` rewrite to protect a selection across, but it also strips CKEditor filling chars that would otherwise split a word, so removing it is a separate change.
- The reader tools' own undo/redo (which snapshots `innerHTML`) is untouched; it exists only because markup used to mutate the DOM. Removing it is its own behavioral risk, to be looked at after testing.

Ref: https://issues.bloomlibrary.org/youtrack/issue/BL-16558

[Devin review](https://devinreview.com/BloomBooks/BloomDesktop/pull/8123)

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8123)

### DIFF
--- a/src/BloomBrowserUI/bookEdit/toolbox/readers/readerHighlights.ts
+++ b/src/BloomBrowserUI/bookEdit/toolbox/readers/readerHighlights.ts
@@ -22,7 +22,7 @@ export const kWordTooLongHighlight = "bloom-reader-word-too-long";
 export const kWordNotDecodableHighlight = "bloom-reader-word-not-decodable";
 export const kSightWordHighlight = "bloom-reader-sight-word";
 
-// The class we put on our hover tip. bloom-ui means Bloom strips it when saving the page.
+// The class we put on our hover tip. Accompanied by bloom-ui so the element is never saved.
 const kTipClass = "bloom-reader-highlight-tip";
 
 interface ReaderHighlightLayer {
--- a/src/BloomBrowserUI/bookEdit/toolbox/readers/readerHighlights.ts
+++ b/src/BloomBrowserUI/bookEdit/toolbox/readers/readerHighlights.ts
@@ -178,7 +178,12 @@ class ReaderHighlightManager {
             );
         });
 
-        this.ensureHoverHandler(contextNode.ownerDocument ?? undefined);
+        // Only listen for hovers while there is actually something to explain.
+        if (this.paintedRanges.size > 0) {
+            this.ensureHoverHandler(contextNode.ownerDocument ?? undefined);
+        } else {
+            this.detachHoverHandler();
+        }
     }
 
     // Remove all the reader highlights, e.g. when the user turns the tool off.
@@ -186,7 +191,7 @@ class ReaderHighlightManager {
         this.rangesInProgress = new Map<string, Range[]>();
         this.paintedRanges = new Map<string, Range[]>();
         this.highlights.clearAllManagedHighlights(contextNode);
-        this.hideTip();
+        this.detachHoverHandler();
     }
 
     private ensureHoverHandler(pageDocument: Document | undefined): void {
@@ -194,10 +199,31 @@ class ReaderHighlightManager {
             return;
         }
         // A new page means a new document; the old one is being discarded along with its handler.
+        this.detachHoverHandler();
         this.hoverDocument = pageDocument;
-        this.tipElement = undefined;
         pageDocument.addEventListener("mousemove", this.handleMouseMove);
         pageDocument.addEventListener("mouseleave", this.handleMouseLeave);
+    }
+
+    // Stop listening, and take our tip out of the page. There is nothing to hover over once the
+    // highlights are gone, and leaving the handler attached would mean doing work on every mouse
+    // move for the rest of the page's life. endPass() attaches it again when there is.
+    private detachHoverHandler(): void {
+        this.hideTip();
+        this.tipElement?.remove();
+        this.tipElement = undefined;
+        if (!this.hoverDocument) {
+            return;
+        }
+        this.hoverDocument.removeEventListener(
+            "mousemove",
+            this.handleMouseMove,
+        );
+        this.hoverDocument.removeEventListener(
+            "mouseleave",
+            this.handleMouseLeave,
+        );
+        this.hoverDocument = undefined;
     }
 
     // Since there is no element to hang a tooltip on, we hit-test the mouse against the ranges we
--- a/src/BloomBrowserUI/bookEdit/toolbox/readers/readerHighlights.ts
+++ b/src/BloomBrowserUI/bookEdit/toolbox/readers/readerHighlights.ts
@@ -72,12 +72,15 @@ const allReaderHighlightNames = readerHighlightLayers.map(
 // zero-width phrase markers. This replaces what removeAllHtmlMarkupFromString() used to strip
 // out of the HTML string.
 function shouldSkipElementForReaderText(element: Element): boolean {
+    // Covers Bloom's transient in-page UI generally, including the format cog (StyleEditor
+    // creates it as <div id="formatButton" class="bloom-ui">).
     if (element.classList.contains("bloom-ui")) {
         return true;
     }
-    if (element.id === "formatButton") {
-        return true;
-    }
+    // The Talking Book tool's phrase markers need their own case: unlike the UI above they are
+    // real saved content (invisible and zero-width), and they carry no other class - Bloom's
+    // spreadsheet export/import writes them as class="bloom-audio-split-marker" exactly, and
+    // MarkedUpText.cs matches that attribute for equality, so bloom-ui cannot be added to them.
     if (element.classList.contains("bloom-audio-split-marker")) {
         return true;
     }
--- a/src/BloomBrowserUI/bookEdit/toolbox/readers/readerHighlights.ts
+++ b/src/BloomBrowserUI/bookEdit/toolbox/readers/readerHighlights.ts
@@ -223,9 +223,22 @@ class ReaderHighlightManager {
         this.hideTip();
     };
 
+    // The layer whose tip should show for this mouse position, if any. Only layers that actually
+    // have a tip are considered: a too-long word sits inside a too-long sentence, and the word
+    // layer has nothing to say, so letting it match first would hide the sentence's explanation.
     private layerAtPoint(event: MouseEvent): ReaderHighlightLayer | undefined {
         const pageDocument = this.hoverDocument;
         if (!pageDocument) {
+            return undefined;
+        }
+        // Cheapest possible exit first. The listeners stay attached for the life of the page
+        // document, so with the tool turned off (or on a page with no violations) this runs on
+        // every mouse move and must not reach getCaretPosition(), which forces a layout.
+        const layersWithTips = readerHighlightLayers.filter(
+            (layer) =>
+                layer.tipL10nId && this.paintedRanges.get(layer.name)?.length,
+        );
+        if (layersWithTips.length === 0) {
             return undefined;
         }
         const target = event.target as Element | null;
@@ -242,9 +255,15 @@ class ReaderHighlightManager {
             return undefined;
         }
 
-        return readerHighlightLayers.find((layer) =>
+        return layersWithTips.find((layer) =>
             (this.paintedRanges.get(layer.name) ?? []).some(
                 (range) =>
+                    // Ranges whose content has been replaced since we painted are not under the
+                    // mouse, and asking isPointInRange about one whose root differs from the
+                    // caret's throws WrongDocumentError, so screen them out first.
+                    range.startContainer.isConnected &&
+                    range.startContainer.ownerDocument ===
+                        caret.node.ownerDocument &&
                     // The caret test is the cheap one, and narrows us to a single candidate
                     // range; then we confirm the mouse really is over the text it covers.
                     range.isPointInRange(caret.node, caret.offset) &&
--- a/src/BloomBrowserUI/bookEdit/toolbox/readers/readerHighlights.ts
+++ b/src/BloomBrowserUI/bookEdit/toolbox/readers/readerHighlights.ts
@@ -243,8 +243,12 @@ class ReaderHighlightManager {
         }
 
         return readerHighlightLayers.find((layer) =>
-            (this.paintedRanges.get(layer.name) ?? []).some((range) =>
-                range.isPointInRange(caret.node, caret.offset),
+            (this.paintedRanges.get(layer.name) ?? []).some(
+                (range) =>
+                    // The caret test is the cheap one, and narrows us to a single candidate
+                    // range; then we confirm the mouse really is over the text it covers.
+                    range.isPointInRange(caret.node, caret.offset) &&
+                    isPointOverRange(range, event.clientX, event.clientY),
             ),
         );
     }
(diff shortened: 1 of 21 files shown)


src/BloomBrowserUI/bookEdit/toolbox/readers/readerHighlights.ts
```diff
@@ -0,0 +1,345 @@
...
+            return undefined;
+        }
+
+        return layersWithTips.find((layer) =>
+            (this.paintedRanges.get(layer.name) ?? []).some(
+                (range) =>
+                    // Ranges whose content has been replaced since we painted are not under the
+                    // mouse, and asking isPointInRange about one whose root differs from the
+                    // caret's throws WrongDocumentError, so screen them out first.
+                    range.startContainer.isConnected &&
```
JohnThomson, line 293: [Devin] **Investigate**: Hover hit-test may call isPointInRange on stale/detached Ranges

`layerAtPoint()` hit-tests the mouse against `this.paintedRanges` using `range.isPointInRange(caret.node, caret.offset)`. The mousemove/mouseleave listeners are attached once per page document in `ensureHoverHandler()` and are never removed (even `clearAll()` only empties `paintedRanges` and hides the tip). If the DOM under a painted Range is edited/replaced between reader-markup passes (e.g. CKEditor operations, or an edit that removes a highlighted node) so that the Range's `startContainer` ends up in a different root than the live caret node, `isPointInRange` throws a WrongDocumentError. In practice `doMarkup` repaints frequently on keyup, so the stale window is small, but a mousemove landing in that window would raise an uncaught exception in the handler (no tip, console noise). Worth confirming the ranges are always fresh, or wrapping the hit test defensively.

JohnThomson: (Claude Opus 5) Fixed in `8c7425c5d8`. Devin was right that this could throw.

The hit test now screens each range out before asking `isPointInRange` about it: a range whose start container is no longer connected, or belongs to a different document than the caret node, can't be under the mouse anyway, and asking about the latter is exactly what raises `WrongDocumentError`.

The same commit also fixes the wasted work Devin noticed on the way to this: because the listeners live for the life of the page document, the "nothing is painted" exit needed to come *before* `caretPositionFromPoint()` (which forces a layout), not after. With the reader tool switched off, a mouse move over a text box now costs a map lookup instead of a layout.

src/BloomBrowserUI/bookEdit/toolbox/readers/readerToolsModel.ts
JohnThomson, line 1: [Devin] **Investigate**: Reader highlights are painted, then bookmark spans are removed after markup

(`src/BloomBrowserUI/bookEdit/toolbox/readers/readerToolsModel.ts:762-768` — outside the diff, so this is a file-level thread.)

In `readerToolsModel.doMarkup()`, `restoreSelectionFromCkEditorBookmarks()` runs AFTER `checkLeveledReader`/`checkDecodableReader` have already registered live Ranges as `::highlight()` layers. That call invokes CKEditor's `selectBookmarks`, which removes the hidden `cke_bm_*` bookmark spans (see `editableDivUtils.ts:503-533`). The freshly-painted reader Ranges reference text nodes (bookmarks are skipped by `mapReaderText`), and removing an element node between two text nodes does not merge/normalize those text nodes, so the Ranges should stay valid. I concluded this is not a bug, but it is worth confirming manually that CKEditor's `selectBookmarks` does not call `normalize()` on the editable (which would merge adjacent text nodes and collapse a Range whose end container was merged away). The PR notes the highlight persistence here was only verified by hand, not by unit tests.

src/BloomBrowserUI/bookEdit/toolbox/readers/readerHighlights.ts
JohnThomson: (Claude Opus 5) Devin's re-review on `79504f801` repeats this flag with the same wording and the pre-fix line numbers (`:245-253`; the code is now at `:229-270`), so I read it as unchanged text rather than a new observation. Its two concrete points are both addressed now:

- the throw — screened out in `8c7425c5d8`, as above;
- the listeners outliving the highlights — fixed in `79504f801`: they are attached by a pass that paints something and removed by a pass that finds nothing, or by `clearAll()` when the tool is switched off. With no reader tool active there is now no handler of ours on the page at all.

Leaving this resolved.

JohnThomson:
(Claude Opus 5) Consulted Devin on 2026-07-28 19:05 UTC up to commit `79504f801903fb5ce3dcac7ea72ae6e77d031cce`.

Three reviews across this preflight (one per pushed commit). No Bugs at any point. Two Investigate flags, both now threads on this PR:

- **Hover hit-test may call isPointInRange on stale/detached Ranges** — real; fixed and resolved.
- **Reader highlights are painted, then bookmark spans are removed after markup** — Devin's own conclusion is that this is not a bug, but it asks for manual confirmation that CKEditor's `selectBookmarks` doesn't normalize the editable. Left open for that check.

6 informational items, not mirrored (low signal): mostly observations about the plain-text switch that the branch addresses deliberately — separator handling in `mapVisibleText`, the case-insensitive word matching in `find_words_extra`, and the offset invariant, which now has an explicit guard and a test.

src/BloomBrowserUI/bookEdit/toolbox/readers/readerToolsModel.ts
JohnThomson: (Claude Opus 5) Confirmed by hand — resolving.

John tested this in a running Bloom: typing in a decodable reader and pausing repeatedly, the problem marks stay put. They do not blink off after a pause and come back on the next one, which is what we would see if CKEditor's `selectBookmarks` were normalizing the editable and invalidating the Ranges' anchors.

So the reasoning in the flag holds: removing the hidden bookmark span from between two pieces of text leaves the surrounding text nodes alone, and the highlights anchored to them survive. No code change.

JohnThomson: (Claude Opus 5) Devin re-raised this on `b082977a93` under a second title, "Reader highlight ranges painted before CKEditor bookmark removal", anchored in `toolbox.ts` rather than `readerToolsModel.ts`. It is the same concern seen from the other end of the call chain, it again concludes NOT a bug for the same reason (the painted Ranges' endpoints are in real text nodes, and removing an element between them does not merge them without an explicit `normalize()`), and it again asks for empirical confirmation — which is what John's hand testing above provides.

Keeping the record here rather than opening a third thread for the same finding. Still resolved.

JohnThomson:
(Claude Opus 5) Consulted Devin on 2026-07-29 00:10 UTC up to commit `b082977a936ef4c217ea392b5b6c61884ee1b9a9`.

**No bugs, and no new findings.** Three Investigate flags, all of them re-raises of the two already dispositioned on this PR:

- *Reader highlights are painted, then bookmark spans are removed after markup* and *Reader highlight ranges painted before CKEditor bookmark removal* — the same concern from two ends of the call chain. Devin concludes NOT a bug both times and asks for empirical confirmation; John confirmed by hand that the marks do not blink off across typing pauses. Documented and resolved on [its thread](https://github.com/BloomBooks/BloomDesktop/pull/8123#discussion_r3670088137).
- *Hover hit-test may call isPointInRange on stale/detached Ranges* — fixed in `8c7425c5d8` and `79504f801`, resolved on [its thread](https://github.com/BloomBooks/BloomDesktop/pull/8123#discussion_r3669808720).

12 informational items, not mirrored (low signal).

This run also covered the four commits made after the previous consultation: the papercut note, the `formatButton` dead-clause removal plus its new test, the `caretRangeFromPoint` simplification, and the canvas comment — plus the merge of `master`.

JohnThomson (commented):
> @JohnThomson+AGNT resolved 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8123)* status: 0 of 21 files reviewed, all discussions resolved (waiting on JohnThomson).

StephenMcConnel (changes requested):
A couple of minor questions is all.

> @StephenMcConnel partially reviewed 21 files and all commit messages, and made 3 comments.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8123#-:-OyjU5Pt4cEQqyWWeGh6:b-vstbyg)* status: all files reviewed, 2 unresolved discussions (waiting on JohnThomson).

___
*[`src/BloomBrowserUI/bookEdit/toolbox/readers/readerHighlights.ts` line 26 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8123#-Oyix6FcCYQzUu-vyvJU:-Oyix6FcCYQzUu-vyvJV:b-u7nyfm) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/b082977a936ef4c217ea392b5b6c61884ee1b9a9/src/BloomBrowserUI/bookEdit/toolbox/readers/readerHighlights.ts#L26)):*
> ```XML
> 
> // The class we put on our hover tip. bloom-ui means Bloom strips it when saving the page.
> const kTipClass = "bloom-reader-highlight-tip";
> ```

The comment about bloom-ui doesn't make sense in context.
___
*[`src/BloomBrowserUI/bookEdit/toolbox/readers/readerHighlights.ts` line 56 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8123#-Oyiy3buBDmoND7nG7at:-Oyiy3buBDmoND7nG7au:b-uvvwb3) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/b082977a936ef4c217ea392b5b6c61884ee1b9a9/src/BloomBrowserUI/bookEdit/toolbox/readers/readerHighlights.ts#L56)):*
> ```XML
>     {
>         name: kWordTooLongHighlight,
>         priority: 2,
> ```

Does this not rate a tip?  Maybe it's never had/needed one?

JohnThomson (commented):
> @JohnThomson made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8123)* status: 15 of 21 files reviewed, 2 unresolved discussions (waiting on StephenMcConnel).

___
*[`src/BloomBrowserUI/bookEdit/toolbox/readers/readerHighlights.ts` line 26 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8123#-Oyix6FcCYQzUu-vyvJU:-Oz7ra4e7vRSiEEawm7_:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/b082977a936ef4c217ea392b5b6c61884ee1b9a9/src/BloomBrowserUI/bookEdit/toolbox/readers/readerHighlights.ts#L26)):*

The comment about bloom-ui doesn't make sense in context.
</blockquote></details>

Done.
___
*[`src/BloomBrowserUI/bookEdit/toolbox/readers/readerHighlights.ts` line 56 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8123#-Oyiy3buBDmoND7nG7at:-Oz8XYHZ8Jykc4CD4pxH:b8ccflb) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/b082977a936ef4c217ea392b5b6c61884ee1b9a9/src/BloomBrowserUI/bookEdit/toolbox/readers/readerHighlights.ts#L56)):*

Does this not rate a tip?  Maybe it's never had/needed one?
</blockquote></details>

As far as Claude and I can see, it's never had one. Maybe the idea was that a too-long word is a pretty obvious problem?

StephenMcConnel (changes requested):
> @StephenMcConnel reviewed 6 files and all commit messages, made 1 comment, and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8123)* status: all files reviewed, 1 unresolved discussion (waiting on JohnThomson).

___
*[`src/BloomBrowserUI/bookEdit/toolbox/readers/readerHighlights.ts` line 26 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8123#-Oyix6FcCYQzUu-vyvJU:-Oz8s_6D-KiBGnq3UFJF:b-sbz8ny) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/b082977a936ef4c217ea392b5b6c61884ee1b9a9/src/BloomBrowserUI/bookEdit/toolbox/readers/readerHighlights.ts#L26)):*

Done.
</blockquote></details>

I don't see any change?

JohnThomson (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8123)* status: 20 of 21 files reviewed, 1 unresolved discussion (waiting on StephenMcConnel).

___
*[`src/BloomBrowserUI/bookEdit/toolbox/readers/readerHighlights.ts` line 26 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8123#-Oyix6FcCYQzUu-vyvJU:-OzCA2Gq2mxGdvIL3DUP:b-3bjblv) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/b082977a936ef4c217ea392b5b6c61884ee1b9a9/src/BloomBrowserUI/bookEdit/toolbox/readers/readerHighlights.ts#L26)):*

I don't see any change?
</blockquote></details>

Apparently it somehow got lost in the rebase. Redone.

StephenMcConnel (approved):
> @StephenMcConnel reviewed 1 file and all commit messages, and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8123)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on JohnThomson).

PR #8124, opened 2026-07-28
Deliver the UI @font-face rules to the preview documents 93afdc5696 missed

JohnThomson:
Follow-up to 93afdc5696, which made `bloomUIFontFaces.css` the single source of the UI `@font-face` rules. That change's audit covered the three `.less` includers and every React root, but not documents that merely **load a bundle** — and before it, `bloomMaterialUITheme.ts` imported `bloomWebFonts.less`, so every bundle injected the declarations into whatever document loaded it.

## What was broken

The book preview documents that `Book.AddPreviewJavascript()` sets up — the collection tab's whole-book preview, single-page previews, and the template-page thumbnails behind Add Page — ended up with no UI `@font-face` at all. `previewMode.less` imports `bloomUI.less` by `(reference)`, so `previewMode.css` never carried them itself, and no React root mounts there, so `renderRoot()` never ran. The UI font stack fell back to whatever the system had, which changes text metrics and shifts the layout.

## The fix

- `bookPreview.ts` installs the faces at module scope. Linking `bloomUIFontFaces.css` from `Book.cs` instead was rejected: `previewMode.css` is copied into published BloomPUB/ePUB artifacts, where `/bloom/fonts/...` URLs don't resolve.
- `ensureUiFontFacesInstalled()` now selects **per family** instead of all-or-nothing. This was necessary, not incidental: Bloom prepends the served book fonts (Andika, ABeeZee, at `/host/fonts/`) to every book's `defaultLangStyles.css`, so a preview document already declares an Andika face while lacking Roboto/NotoSans/SymChar. The old check looked only for Roboto, so a naive fix would have injected the whole file and added a **second, late Andika face** — precisely the duplicate-arrives-after-render condition that blanks text for several frames. The filtering is factored out as `selectMissingFontFaceRules()` and unit tested against the real stylesheet, because a `?raw` css import yields an empty string under vitest's default `css:false`.
- The guard test enforced only "no duplicate declarations". Nothing enforced the other half of the invariant — that every Bloom-managed document actually gets them — which is how this was missed. It now also pins the set of files importing `utils/uiFontFaces`, matching on the module import rather than the function name so a namespace, dynamic, or aliased import can't slip a fifth delivery point past it.

## Audited and clear

All 28 Vite entry points and every document that links a stylesheet: the edit page, toolbox, and reader-setup documents are covered statically; the WinForms-hosted React documents and the workspace/app document are covered by `renderRoot()`. The page thumbnail list needed no change — its "Pages" heading is `Segoe UI` and its only `@UIFontStack` consumer is the runtime toast, both after `renderRoot()`. `spreadsheetFunctions.html` loads a bundle with no React root but is a headless utility page with no rendered text. Published ePUB/BloomPUB output correctly uses book fonts, not ours.

## Not verified

Whether restoring these faces fixes the specific layout shift originally reported on the arithmetic `numberRow`. Those `.Equation` editables are `font-family: Aileron, Arial` and Aileron is not among Bloom's served fonts, so the shift is most likely a container whose UI font changed — but that was not confirmed at the glyph level.

## Testing

- `pnpm typecheck`, `pnpm lint` (0 errors)
- front-end suite: 575 tests pass
- C# suite via `build/agent-dotnet.sh`: 2942 pass, 0 fail
- `build/agent-vite.sh` confirms the production bundles compile and `bookPreviewBundle.js` pulls in the chunk carrying the font CSS

No YouTrack card: 93afdc5696's `BL-15300` reference points at an unrelated Talking Book issue, and no card exists for this font work.

[Devin review](https://devinreview.com/BloomBooks/BloomDesktop/pull/8124)

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8124)

### DIFF
--- a/src/BloomBrowserUI/bloomUIFontFaces.css
+++ b/src/BloomBrowserUI/bloomUIFontFaces.css
@@ -9,15 +9,30 @@
  * the text goes invisible for several frames while it re-loads ("FOIT"). That was the cause
  * of a visible text flicker on every page change in edit mode.
  *
+ * Equally, EVERY Bloom-managed document needs them: a document that lacks them silently falls
+ * back to whatever fonts the system happens to have, which changes text metrics and can shift
+ * the layout (that is how BL-14102's missing Andika shows up).
+ *
  * So: do NOT @import this file (or re-declare these fonts) in .less that gets compiled into
  * bundles or component styles. It is delivered in exactly two ways:
  *   1. A plain `@import (less)` in the root stylesheet of each document that contains
  *      server-rendered text (editMode.less, toolbox.less, readerSetup.less), all of which are
- *      linked in their document's <head> at parse time.
- *   2. ensureUiFontFacesInstalled() (utils/uiFontFaces.ts), which injects this file's text --
- *      only if the document doesn't already have these faces -- and is called from
- *      reactRender.tsx for every React root, covering all the WinForms-hosted React documents.
- * A unit test (uiFontFacesSingleSource.test.ts) enforces this.
+ *      linked in their document's <head> at parse time, so they are in effect before that text
+ *      first paints.
+ *   2. ensureUiFontFacesInstalled() (utils/uiFontFaces.ts), which injects the rules for the
+ *      families the document is missing. It is called from reactRender.tsx for every React
+ *      root, covering all the WinForms-hosted React documents, and from bookPreview.ts, whose
+ *      documents (the collection tab preview, single-page previews and template-page
+ *      thumbnails) have neither a stylesheet from (1) nor a React root.
+ *
+ * Note that (2) filters per family rather than injecting all or nothing, because a document may
+ * arrive already declaring some of these families and not others: Bloom prepends the served
+ * book fonts (Andika, ABeeZee, from /host/fonts/) to each book's defaultLangStyles.css, so any
+ * document that links a book's stylesheets already has an Andika face. Adding a second one late
+ * would cause exactly the blanking described above.
+ *
+ * A unit test (uiFontFacesSingleSource.test.ts) enforces both the single source and the two
+ * lists of approved delivery points.
  *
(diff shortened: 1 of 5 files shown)


JohnThomson:
_(Claude Opus 5 (1M context))_

Consulted Devin on 2026-07-28 20:05 UTC up to commit `446e2441dff4ddb17c4c1421fa9481aa46542fbf`.

**Re-review clean — bots quiet.** No bugs and nothing flagged for investigation, so there are no findings mirrored as review threads.

Devin was consulted twice, once per commit:

- **`a5788c5b94`** — 0 bugs, 2 informational notes. One was worth acting on: it pointed out that for the template-page thumbnails, `Book.GetThumbnailXmlDocumentForPage()` adds the preview bundle script *before* `EnsureStylesheetLinks()`, and `SortStyleSheetLinks()` re-appends every stylesheet to the end of `<head>` — so on that path the stylesheets come after the script, and it read as though the browser might not know about the book's Andika font yet when our code runs. It can: bundle scripts are `type="module"`, which are deferred and run only after the whole page is parsed. Devin reached the same "not a bug" conclusion by a different route, but the previous comment invited the misreading, so `446e2441d` writes the real reason down. The other note was an observation about font-family name matching, no action.
- **`446e2441d`** — 0 bugs, 3 informational notes, all confirming the reasoning holds (including that the new `.spec.ts` is picked up by vitest's include pattern). No action.

CI is green (`pr-automation`, `PR Review Tracker`). CodeRabbit is present in this repo but has `auto_review` disabled in `.coderabbit.yml`, so it did not review.

JohnThomson:
Decided not to make this change. A closer look at the code indicated that these fonts are specifically included to support Bloom's UI, NOT to support rendering document content. It was a bug that they changed the document default font used to render document content. This PR would have reinstated the bug (which the previous PR accidentally fixed, though doing so revealed other problems).

PR #8125, opened 2026-07-28
Give the preview/print DOMs the UI @font-face rules they still reference (BL-15300)

andrew-polk:
**This is the minimal-fix alternative to #8122. Only one of the two should land.**

`Book.AddPreviewJavascript()` injects `bookPreviewBundle.js` into three book documents — the whole-book preview, page thumbnails, and the printing DOM. That bundle transitively pulls in `bloomUI.css`, whose `body { font-family: Roboto, NotoSans, sans-serif }` rule applies Bloom's UI font stack to the book. The matching `@font-face` declarations used to ride along in the same bundle (via `bloomMaterialUITheme.ts` importing `bloomWebFonts.less`), but BL-15300 moved them out to `bloomUIFontFaces.css`, which reaches documents only through a statically-linked root stylesheet or `ensureUiFontFacesInstalled()` from `renderRoot()`. A book preview gets neither.

So those documents were left naming Roboto with no Roboto declared, and any text falling through to the `body` rule dropped to a system font. Only the arithmetic page shows it, because its editables are `lang="*"`, which no `defaultLangStyles` rule matches; all other text has an explicit per-language font with its own FontServe `@font-face`. The visual regression suite has been failing all six A5 Portrait cases since BL-15300 landed.

This change links `bloomUIFontFaces.css` into those documents so the font resolves again — statically rather than injected from JS, because their text is server-rendered and a late-arriving `@font-face` is the very flicker BL-15300 was fixing.

**Trade-off vs #8122.** This restores the previous rendering exactly: no reference images change. It does *not* remove the underlying leak — `bloomUI.css` is still being applied to book content, which is what #8122 addresses at the source. #8122 is the deeper fix but changes how the arithmetic page renders (to the book's own Andika stack) and so updates six reference images.

Known, pre-existing and not addressed by either PR: the edit tab renders that same text in Roboto (its own `editMode.css` legitimately carries the UI body font), and the published BloomPUB renders it in Arial (the staged book's `fonts.css` is empty and its `defaultLangStyles.css` declares no `@font-face`). The root issue is that `lang="*"` content has no font specified anywhere.

Visual regression suite: 12/12, no reference images changed. Full C# and front-end suites green.

Ref: https://issues.bloomlibrary.org/youtrack/issue/BL-16624

Alternatives under consideration for the same bug — only one should land:
- #8124 (John Thomson) — install the faces from `bookPreview.ts`, selecting per font family
- #8125 — link `bloomUIFontFaces.css` from `Book.cs`
- #8122 — remove the leak so book documents never name the UI font at all

[Devin review](https://devinreview.com/BloomBooks/BloomDesktop/pull/8125)

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8125)

### DIFF
--- a/src/BloomBrowserUI/bloomUIFontFaces.css
+++ b/src/BloomBrowserUI/bloomUIFontFaces.css
@@ -1,6 +1,7 @@
 /*
- * The @font-face declarations for the fonts Bloom's own UI uses (the content of books gets
- * its fonts from defaultLangStyles.css and friends, not from here).
+ * The @font-face declarations for the fonts Bloom's own UI uses. Book content is supposed to get
+ * its fonts from defaultLangStyles.css and friends rather than from here, and mostly does -- but
+ * see delivery route 3 below for one place where a book document genuinely needs these.
  *
  * IMPORTANT (BL-15300): these declarations must appear EXACTLY ONCE per document, in CSS that
  * is present before the document first paints. When a stylesheet containing a duplicate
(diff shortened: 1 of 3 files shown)


andrew-polk:
[Claude Opus 5] Consulted Devin on 2026-07-28 16:22 USMST up to commit `1f834032ff32e866a1df8b42f8e627b19e70cbdc`.

Result: clean — no Bugs and no Investigate flags. One Informational item, which was a confirmation rather than a problem: it checked that this change keeps the BL-15300 "@font-face declarations appear exactly once per document, before first paint" invariant, and concluded the static link is the sole source for these documents (nothing else statically links a copy, `bloomUI.css` carries no faces, and `EnsureStylesheetLinkWithoutSorting` is idempotent so the cached preview DOM can't get a second link).

CI is green (pr-automation, track/set-waiting). Full C# suite 2943 passed / 12 skipped, with one pre-existing environmental failure (`RabRealBuildTests` — Reading App Builder's buildapp.bat is not installed on this machine). Front-end suite 568 passed. Visual regression suite 12/12 with no reference images changed.

andrew-polk:
[Claude Opus 5] Closing in favour of #8122.

This approach has a defect that #8124 surfaced: `ensureUiFontFacesInstalled()`'s guard is all-or-nothing on Roboto, but Bloom prepends the served book fonts to every book's `defaultLangStyles.css`, so these documents already declare Andika. Measured on this branch, the preview ends up with **five** Andika `@font-face` declarations — one from the `bloomUIFontFaces.css` link this PR adds (`/bloom/fonts/Andika-Regular.woff2`) and four already served (`./host/fonts/Andika-*.woff2`). That is the duplicate the BL-15300 invariant exists to prevent. Because the link is static it should not flicker, but document order then decides whether book text in Andika uses the book's served copy or Bloom's UI copy.

The plan is now #8122 (remove the leak at its source so book documents never name the UI font), plus the guard-test hardening from #8124, plus a fix for the underlying `lang="*"` gap. See BL-16624.

PR #8128, opened 2026-07-29
Remove required tools for Decodable/Leveled Readers (BL-16615)

StephenMcConnel:
## Problem

Books made from the Leveled Reader (LR) or Decodable Reader (DR) templates were quasi-permanently stuck to their original reader type (BL-16615). Repro: make a book from the LR template, mark it "not Leveled", switch to the DR tool and mark it "Decodable" — then on every page change the toolbox snapped back to Leveled Reader.

## Root cause

Each reader template carried `<meta name="pageToolId" content="leveledReader">` (or `decodableReader`). Via the Add Page path that value was stamped onto every page as `data-tool-id`. On each page display, `toolbox.ts` `adjustToolListForPage` read that `data-tool-id` and called `setCurrentTool(...)`, which also persists back to `BookInfo.CurrentTool` — so the manual switch could never stick.

## Fix

Remove the `pageToolId` meta from the Decodable Reader and Leveled Reader templates, so newly-created reader books no longer stamp the stale `data-tool-id` onto their pages. The reader tool still auto-opens on first edit because each template's `meta.json` still declares `currentTool` / enabled `tools` / `toolboxIsOpen`, which drives tool activation independently of `pageToolId`.

## Scope / limitation

Forward-only: books **already created** from these templates keep their baked-in `data-tool-id` and remain stuck. This matches the issue's own "we can probably live with this" framing.

Ref: https://issues.bloomlibrary.org/youtrack/issue/BL-16615

---
<sub>PR prepared with Claude Opus 4.8 (1M context) via /preflight.</sub>

[Devin review](https://devinreview.com/BloomBooks/BloomDesktop/pull/8128)

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8128)


StephenMcConnel:
[Claude Opus 4.8 (1M context)] Consulted Devin on 2026-07-29 15:11 MDT up to commit c64e3ff31c3308dd5f8c6b334d1321a380e902fd. Result: **clean** — 0 Bugs, 0 Investigate flags. 1 Informational item (a confirmation that the fix is safe and complete: the reader tool still auto-opens via meta.json `currentTool`, independent of the removed `pageToolId`). No findings required action.

src/BloomBrowserUI/bookEdit/toolbox/toolbox.ts
```diff
@@ -140,7 +140,20 @@ export class ToolBox {
...
+        // attribute here would force the reader tool open and keep the book "stuck" to its
+        // original type, preventing the user from switching to (and staying on) another
+        // tool. So we ignore those values and leave the last tool shown (stored in the
+        // book's metadata) as the current tool. (BL-16615)
+        if (
+            requiredToolId === "leveledReader" ||
+            requiredToolId === "decodableReader"
+        ) {
+            requiredToolId = null;
+        }
```
StephenMcConnel, line 156: [Devin] **Investigate**: Hardcoded tool-id denylist duplicates what requiresToolId() already expresses

Devin flags that this fix names two specific tool ids ("leveledReader", "decodableReader") to ignore, whereas the general contract for "a page's `data-tool-id` may pin the toolbox to a particular tool" is already expressed by `ITool.requiresToolId()` — which today only the Game tool returns `true` for.

An equivalent, self-maintaining alternative would be to honor `requiredToolId` **only** when the matching entry in `masterToolList` returns `requiresToolId() === true`, instead of a hardcoded denylist. That would also cover other legacy `data-tool-id` values baked into existing books — notably `signLanguage`, which appears on pages in existing content/test data — which today would still force/persist the sign-language tool on every page display via the `setCurrentTool(requiredToolId)` fall-through, i.e. the same "stuck book" symptom BL-16615 describes.

This is a scope/design question rather than a defect: the current change deliberately targets only the two reader tools that BL-16615 and the card discussion (John Hatton's comment) called out. Broadening it to the general `requiresToolId()` rule would also change behavior for `signLanguage`, which is beyond the requested scope. Leaving this thread open for the developer to decide.

StephenMcConnel:
[Claude Opus 4.8 (1M context)] Consulted Devin on 2026-07-30 up to commit 3d72f0361. Result: 0 bugs, 1 Investigate flag, 1 Informational flag. The Investigate flag (hardcoded reader denylist vs. generalizing to requiresToolId(), which would also cover signLanguage) is a scope/design question posted as an open review thread for the developer to decide (#discussion_r3685243989). The Informational flag confirmed removing the pageToolId meta from the templates is safe (auto-open still driven by meta.json currentTool). CI and other checks green.

src/BloomBrowserUI/bookEdit/toolbox/toolbox.ts
StephenMcConnel: [Claude Opus 4.8 (1M context)] Decided: leave as is — keep the narrow, reader-only guard. This change is deliberately scoped to exactly what BL-16615 and John Hatton's card comment asked for: ignore the `leveledReader`/`decodableReader` page tags so reader books stop snapping back to their original tool. Generalizing to the `requiresToolId()` rule would also change behavior for `signLanguage` pages, which is beyond this card's scope and has no test coverage here; if we want that broader behavior it's better handled as its own card. Resolving.

andrew-polk (approved):
> @andrew-polk reviewed 3 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8128)* status: all files reviewed, 1 unresolved discussion (waiting on StephenMcConnel).

andrew-polk (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8128)* status: all files reviewed, 1 unresolved discussion (waiting on StephenMcConnel).

PR #8129, opened 2026-07-30
BL-16627: improve BloomBridge process-book fix-up: auto-fit origami splits (both directions, nested stacks) and shrink oversized images

hatton:
Auto-fit of image/text origami splits gets two related improvements. Both are about the same
governing rules: never leave text overflowing, never waste space to no one's benefit, and always
measure the real browser layout rather than estimating font sizes.

### 1. Let auto-fit shrink the image pane, not just grow it (`7bea4c7`)

`fitImageOverTextSplitOnPage()` computed the right split and then discarded it whenever applying it
would make the image pane smaller. On a page whose illustration is wider than 4:3 the image pane is
bigger than the picture needs, the surplus is dead space, and the text below gets clipped — even
though the page has room to spare. Replaced the one-way ratchet with a small deadband so the divider
moves whenever it is worth moving in either direction.

### 2. Handle nested text/image/text stacks (`e455270`)

The function bailed on any nested split (`// nested split: too complex, skip`), which excluded one of
the most common page shapes there is: **text above, an illustration, text below**.

Nothing ever chose those dividers on purpose. An importer emits the shape as a right-nested chain of
horizontal splits, and with no explicit percentages Bloom's 50/50-per-split CSS default cascades into
an effective **50/25/25** — so the top text block gets half the page for its one line, the
illustration is squeezed into a quarter, and the last (usually longest) block is clipped.

**What makes a top/bottom stack tractable** is that every pane in it has the same *width*, so a text
block's required height doesn't depend on where any divider sits. Each text pane can therefore be
measured on its own — give it a size, ask `OverflowChecker` whether it overflows, binary-search the
boundary — and the answers add up. The illustration takes what the text doesn't need, capped at the
size where it already fills the width; the surplus past that goes back to the text panes in
proportion to what each needed.

Left/right stacks are deliberately excluded: there the panes have different widths, so they aren't
independent and the one-at-a-time measurement doesn't hold.

Still declined, as before: a split inside a *first* pane, mixed axes, a second pane holding a split
alongside other content, a pane that isn't cleanly just-a-picture or just-text, an illustration
carrying overlays, more than one illustration, and genuinely over-full pages (text that overflows
even with nearly the whole page, or blocks that together need more than the page has).

## Measured result

`FUV_Taale_Taale_Fulbe` (Fulfulde, Nigeria; A5Portrait, marginBox 469x703), where **9 of the 11
content pages** are this shape. Every one sat at 351.5/175.75/175.75 px, and 9 pages had text
overflowing. Afterwards **none** do.

| page | before | after |
| --- | --- | --- |
| 5 | 352 / 176 / 176 (clipped) | 103 / 334 / 266 |
| 8 | 352 / 176 / 176 (clipped) | 154 / 325 / 224 |
| 9 | 352 (clipped) / 176 / 176 | 490 / 152 / 62 |
| 11 | 352 / 176 / 176 (clipped) | 238 / 101 / 364 |
| 14 | 352 / 176 / 176 (clipped) | 87 / 397 / 219 |

Page 15 is text/text with no illustration, so it is out of scope and correctly left untouched (it is
genuinely over-full).

## Testing

`autoFitImageOverTextSplitsSpec.ts` covers shape recognition and the nested percentage arithmetic —
including that unset splits read as 50/25/25 rather than thirds, which is the bug captured as an
assertion.

The measuring itself needs real geometry, which jsdom does not provide (`OverflowSpec` is skipped
there for the same reason). That half was verified by loading the real saved book, with its real
Bloom CSS, in Chromium and running `fitImageOverTextSplits()` against it — that is where the numbers
above come from.

Ref: https://issues.bloomlibrary.org/youtrack/issue/BL-16627

[Devin review](https://devinreview.com/BloomBooks/BloomDesktop/pull/8129)

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8129)

### DIFF
--- a/src/BloomBrowserUI/bookEdit/js/autoFitImageOverTextSplits.ts
+++ b/src/BloomBrowserUI/bookEdit/js/autoFitImageOverTextSplits.ts
@@ -144,6 +144,13 @@ function fitImageOverTextSplitOnPage(page: HTMLElement): boolean {
         void splitPane.offsetHeight;
     };
 
+    // Undo the probing exactly, for the paths that end up declining the page — see
+    // captureSplitStyles. Writing back `originalTextPercent` instead would look equivalent but isn't:
+    // on a split with no explicit percentage (the common case, defaulting to 50%) it would leave an
+    // explicit "50%" behind on a page we decided not to touch.
+    const savedStyles = captureSplitStyles([splitConfig]);
+    const restore = () => restoreSplitStyles(savedStyles, splitPane);
+
     const textOverflows = (): boolean => textGroupOverflows(secondTextGroup);
 
     // Upper bound for the search: a text pane this tall is assumed to fit any text we'd auto-fit.
@@ -153,7 +160,7 @@ function fitImageOverTextSplitOnPage(page: HTMLElement): boolean {
     // growing the image (it's just over-full). Leave it exactly as we found it.
     setTextPercent(hiBound);
     if (textOverflows()) {
-        setTextPercent(originalTextPercent);
+        restore();
         return false;
     }
 
@@ -200,7 +207,7 @@ function fitImageOverTextSplitOnPage(page: HTMLElement): boolean {
     // Nothing worth doing (the authored split is already about right): put it back exactly as we
     // found it rather than rewriting the style with a rounding-noise difference.
     if (Math.abs(finalTextPercent - originalTextPercent) < kFitMinMovePercent) {
-        setTextPercent(originalTextPercent);
+        restore();
         return false;
     }
     setTextPercent(finalTextPercent);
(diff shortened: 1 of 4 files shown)


hatton:
[Claude Opus 5 from Hatton's machine during preflight]

**Consulted Devin** on 2026-07-29, up to commit `8833fb4f6b2ba1c6d3d3d823476d3bb4f866264d`.

Devin ran on each of the three pushed commits and I waited for the job matching that exact sha
each time. The final pass is **clean: 0 Bugs, 0 Investigate flags.** Informational items only
throughout (3, then 6, then 7). Those aren't mirrored as review threads, but I read all of them,
and two were worth acting on:

- *"Two-pane decline path writes an explicit percent onto previously-unset splits"* — fixed in
  `496767bf2b`. Measuring rewrites the split styles repeatedly, so a page we then decline had
  already been mutated; the old write-back stamped an explicit `50%` onto splits that had no style
  at all. Both paths now restore the captured style attributes, so a page we choose not to touch
  stays out of the saved HTML's diff.
- *"Overlay-on-image rejection is masked by the translationGroup check"* — the check isn't dead, so
  no code change; but Devin's reading is exactly the misreading that would lead someone to delete
  it, so `8833fb4f6b` records why it's needed (the earlier test excludes text bubbles; this one
  excludes image/video overlays).

The rest were confirmations rather than concerns — notably that the leftover-redistribution and
over-full guards preserve the no-overflow invariant, and that the unit-mixing in the image-fit cap
matches the reference `getImagePercent()` in `split-pane.ts` and only ever skews the whitespace
cap, never the no-overflow guarantee.

Also green at this commit: CI (`pr-automation`, `track / set-waiting`), and the full BloomBrowserUI
test suite (586 passing, 5 skipped). CodeRabbit is configured but switched off in this repo
(`.coderabbit.yml` has `auto_review.enabled: false`), so it is not a pending reviewer.

Full preflight report, including the before/after page geometry and a few decisions for the
developer: https://raw.githack.com/BloomBooks/dev-process-artifacts/main/deciders/BloomDesktop-BL-16627-shrink-image-pane.html

hatton:
[Claude Fable 5 following a prompt from Hatton] Superseded by #8135: the branch was renamed from BL-16627-shrink-image-pane to BL-16627-process-book to match the card's broadened scope (the work now covers more of the process-book fix-up than the image-pane split), and a branch rename closes the PR that pointed at the old name. Same commits plus the new image-shrink step continue there.

PR #8130, opened 2026-07-30
Size spreadsheet image rows from the embedded thumbnail (BL-16529)

StephenMcConnel:
## Problem

Follow-on to the BL-16529 DPI-width fix. After that fix, spreadsheet-export **image rows came out too tall** for small images.

The exporter sized each image row from the *requested* thumbnail target (`finalWidth = defaultImageWidth`, scaled up by the monitor DPI factor, and `finalHeight` derived from it). But `ImageUtils.ResizeImageIfNecessary` never enlarges an image, so a source smaller than that target is embedded at its original, smaller size — while the row stayed sized for the larger target, leaving dead vertical space.

## Fix

Two parts, both anchored to the **actually-embedded thumbnail** rather than the requested target:

- **Size the row from the embedded thumbnail height**, not the target (which the resizer never enlarges up to for a small source).
- **Divide that height back down by the DPI scale factor when `scaleFactor > 100`.** The spreadsheet viewer scales columns *and rows* up for a high-DPI display but draws the embedded image at its absolute pixel size, so the row (whose height is in points) must be scaled down so the *displayed* row matches the image — the vertical counterpart of the width scaling that fixed the original too-narrow bug. At 100% the `else` branch sizes the row straight from the thumbnail height (without it, `finalHeight` would keep the oversized target, e.g. 195px for the 118×154 image, and 100% displays would still be too tall).

## Verification

Confirmed by exporting and reading the raw `.xlsx` on a **125%-scaled** machine (nominal = the row height stored in the file; displayed ≈ nominal × 1.25, which is what the viewer draws):

| Source image | Embedded image | Row height before | Row height after (nominal → displayed) |
|---|---|---|---|
| 118×154px (small) | 154px | 199–248px | 127px → ~159px ✓ (matches image) |
| same @300dpi | 154px | 199–248px | 127px → ~159px ✓ (source DPI irrelevant) |
| 199px → down-scaled to 187 | 187px | 191px | 152px → ~190px ✓ (matches image) |

Regression test `Export_SmallImage_RowNotSizedFromOversizedTarget` asserts (scale-independently) that the row is sized from the embedded thumbnail, not the oversized target. Full Spreadsheet suite green (385/385).

> Note: because the exporter reads the primary monitor's scale factor at runtime, the test's un-enlarge/100% branch is only exercised as a true regression catch on a 100%-scaled machine; on a >100% machine the divide branch runs instead. Both branches are covered by the arithmetic in the fix; the test guards the shared upper-bound invariant at any scale.

Ref: https://issues.bloomlibrary.org/youtrack/issue/BL-16529

[Devin review](https://devinreview.com/BloomBooks/BloomDesktop/pull/8130)

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8130)

### DIFF
--- a/src/BloomExe/Spreadsheet/SpreadsheetIO.cs
+++ b/src/BloomExe/Spreadsheet/SpreadsheetIO.cs
@@ -229,6 +229,12 @@ namespace Bloom.Spreadsheet
                                 )
                             )
                             {
+                                // ResizeImageIfNecessary never enlarges an image, so when the source
+                                // is smaller than the (DPI-scaled) target the thumbnail is embedded at
+                                // its original, smaller size. Size the row from the thumbnail we actually
+                                // embedded, not from the requested finalHeight; otherwise small images
+                                // get a row that is much too tall. See BL-16529.
+                                finalHeight = thumbnail.Height;
                                 using (var ms = new MemoryStream())
                                 {
                                     thumbnail.Save(ms, ImageFormat.Jpeg);
--- a/src/BloomExe/Spreadsheet/SpreadsheetIO.cs
+++ b/src/BloomExe/Spreadsheet/SpreadsheetIO.cs
@@ -229,12 +229,16 @@ namespace Bloom.Spreadsheet
                                 )
                             )
                             {
-                                // ResizeImageIfNecessary never enlarges an image, so when the source
-                                // is smaller than the (DPI-scaled) target the thumbnail is embedded at
-                                // its original, smaller size. Size the row from the thumbnail we actually
-                                // embedded, not from the requested finalHeight; otherwise small images
-                                // get a row that is much too tall. See BL-16529.
-                                finalHeight = thumbnail.Height;
+                                // Size the row from the thumbnail we actually embedded (which, since
+                                // ResizeImageIfNecessary never enlarges, may be smaller than the target
+                                // computed above), not from that target. The row's height is in points,
+                                // which the spreadsheet viewer does NOT scale for a high-DPI display even
+                                // though it scales the columns; so at >100% we divide the height back down
+                                // by the same factor so the displayed row matches the image. See BL-16529.
+                                if (scaleFactor > 100)
+                                    finalHeight = thumbnail.Height * 100 / scaleFactor;
+                                else
+                                    finalHeight = thumbnail.Height;
                                 using (var ms = new MemoryStream())
                                 {
                                     thumbnail.Save(ms, ImageFormat.Jpeg);
--- a/src/BloomTests/Spreadsheet/SpreadsheetImagesTests.cs
+++ b/src/BloomTests/Spreadsheet/SpreadsheetImagesTests.cs
@@ -679,13 +679,19 @@ namespace BloomTests.Spreadsheet
 
         /// <summary>
         /// Regression test for BL-16529 (image cells too high). ResizeImageIfNecessary never
-        /// enlarges an image, so a source smaller than the (possibly DPI-scaled) thumbnail target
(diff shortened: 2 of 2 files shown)


StephenMcConnel:
[Claude Opus 4.8] Consulted Devin on 2026-07-30 14:58 UTC up to commit `b65559131d9f033a1335c97e2b2e454b962e2938`. Result: **clean** — 0 Bugs, 0 Investigate flags, 3 Informational observations (integer truncation in the row-height math already covered by the +3 padding; a note that `scaleFactor` now only affects the down-scale target; and that the new test relies on the shared test image staying small, which its own `img.Width < 150` sanity assert already guards). None are action items. CI and other bots: see checks.

src/BloomExe/Spreadsheet/SpreadsheetIO.cs
```diff
@@ -229,6 +229,16 @@ bool NameIsTaken(string candidate) =>
...
                             )
                             {
+                                // Size the row from the thumbnail we actually embedded (which, since
+                                // ResizeImageIfNecessary never enlarges, may be smaller than the target
+                                // computed above), not from that target. The row's height is in points,
+                                // which the spreadsheet viewer does NOT scale for a high-DPI display even
+                                // though it scales the columns; so at >100% we divide the height back down
+                                // by the same factor so the displayed row matches the image. See BL-16529.
+                                if (scaleFactor > 100)
+                                    finalHeight = thumbnail.Height * 100 / scaleFactor;
```
StephenMcConnel, line 239: [Devin] **Investigate**: PR description's verification table no longer matches the final code

Devin flagged that the PR description still described the earlier `finalHeight = thumbnail.Height` (cap-only) fix — its verification table (e.g. the down-scaled image's row "unchanged at 191px") and the referenced test name (`Export_SmallImage_RowNotTallerThanEmbeddedImage`) were stale relative to commit e5838a1c0, which divides the embedded thumbnail height by the DPI scale factor at >100%. Reviewers should be able to verify the DPI-divide behavior, not the earlier capping behavior.

StephenMcConnel: [Claude Opus 4.8] Fixed by updating the PR description (for commit e5838a1c0): the Fix section now describes the DPI-divide behavior and the else-branch, the verification table shows nominal → displayed row heights on the 125% machine (118×154 image: 127px → ~159px; down-scaled 187px image: 152px → ~190px), and the test name is corrected to `Export_SmallImage_RowNotSizedFromOversizedTarget`. Resolving.

StephenMcConnel:
[Claude Opus 4.8] Consulted Devin on 2026-07-30 17:55 UTC up to commit `e5838a1c0c07b3b77279dcce6850d1041b0e96f7`. Result: 0 bugs, 1 Investigate — "PR description stale vs the DPI-divide code" — fixed by updating the description (thread resolved); 5 informational, none actionable (integer truncation covered by padding; throw-path keeps the target height for a failed-image row; the test's un-enlarge branch is only a live regression catch at 100% scale; test relies on the shared image staying small; scaleFactor now affects only the down-scale target). CI and other bots quiet.

andrew-polk (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8130)* status: all files reviewed, 1 unresolved discussion (waiting on StephenMcConnel).

andrew-polk (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8130)* status: all files reviewed, 1 unresolved discussion (waiting on StephenMcConnel).

PR #8131, opened 2026-07-30
Leveled Reader: restore orange for over-limit stats (BL-16408)

andrew-polk:
The React conversion of the Leveled Reader tool (BL-16408) hardcoded `maxNum={0}` for two statistics rows, which permanently disabled their over-the-limit test, so they always rendered lightgreen instead of orange:

- **This Book / longest sentence** (`actualMaxWordsPerSentence`)
- **Word Lengths / max in book** (`actualMaxGlyphsPerWord`)

Before the conversion, `readerToolsModel.updateActualCount` compared those two against `maxWordsPerSentenceOnThisPage()` and `maxGlyphsPerWord()` respectively and applied the `tooLarge` (orange) class. The old pug template's max cells for these rows (`maxWordsPerSentenceInBook`, `maxGlyphsPerWordInBook`) were never populated by `updateLevelLimits`, so the Max column looked empty for them — which is presumably why the conversion concluded the statistics were unlimited.

This passes the real limits again, and adds a `hideMaxNum` prop so the Max cell stays blank on those two rows, matching 6.4's appearance exactly. Each of them shares its limit with the row directly above it, which does display it.

Also: averages are now compared at full precision instead of comparing the string we display. An average of 9.04 against a max of 9 displays as "9.0" but is over the limit, and the pre-conversion code counted it as such.

### Tests

New `LeveledReaderToolControls.spec.tsx` covers the max-to-color mapping for over-limit, within-limit, no-limit-for-this-level, and rounded-average rows. Verified it fails on the pre-fix code: reinstating either `maxNum={0}` or the old rounded-string comparison turns the relevant assertions red. The conversion had removed the tests that covered level maximums (`sets level max values on init`, `updates max values when level changes`).

Rendering a localized component in a test needed `processSimpleMarkdown` added to the shared `localizationManager` mock in `vitest.setup.ts`.

Ref: https://issues.bloomlibrary.org/youtrack/issue/BL-16408

Also fixes BL-16628, which is the same bug reported separately with a 6.4-vs-6.5 side-by-side.

[Devin review](https://devinreview.com/BloomBooks/BloomDesktop/pull/8131)

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8131)

### DIFF
--- a/src/BloomBrowserUI/bookEdit/toolbox/readers/leveledReader/LeveledReaderToolControls.spec.tsx
+++ b/src/BloomBrowserUI/bookEdit/toolbox/readers/leveledReader/LeveledReaderToolControls.spec.tsx
@@ -11,6 +11,10 @@ import { LeveledReaderStats } from "./LeveledReaderToolControls";
 const kMaxWordsPerSentence = 10;
 const kMaxGlyphsPerWord = 5;
 const kMaxAverageWordsPerSentence = 9;
+// A second level, used to check that the limits are re-read when the level changes.
+// Its glyph limit is high enough that the statistic that is too large for level 1
+// is acceptable here.
+const kLevel2MaxGlyphsPerWord = 20;
 
 const bookStats = {
     levelNumber: 1,
@@ -91,6 +95,10 @@ describe("LeveledReaderStats", () => {
                     maxWordsPerSentence: kMaxWordsPerSentence,
                     maxGlyphsPerWord: kMaxGlyphsPerWord,
                     maxAverageWordsPerSentence: kMaxAverageWordsPerSentence,
+                    thingsToRemember: [""],
+                },
+                {
+                    maxGlyphsPerWord: kLevel2MaxGlyphsPerWord,
                     thingsToRemember: [""],
                 },
             ],
@@ -180,5 +188,26 @@ describe("LeveledReaderStats", () => {
         expect(
             getRow(container, "ThisPage", "PerPage").maxCell.textContent,
         ).toBe("");
+    });
+
+    it("re-reads the level's limits when the level changes", () => {
+        // The conversion dropped the tests that covered this, and the whole fix
+        // depends on the limits being current whenever we render.
+        const atLevel1 = getRow(container, "WordLengths", "MaxInBook");
+        expect(atLevel1.actualCell.classList).toContain("tooLarge");
+
+        getTheOneReaderToolsModel().setLevelNumber(2, true);
+        act(() => {
+            renderRoot(<LeveledReaderStats bookStats={bookStats} />, container);
+        });
+
+        // 6 is within level 2's much larger glyph limit, so it is no longer flagged.
+        const atLevel2 = getRow(container, "WordLengths", "MaxInBook");
+        expect(atLevel2.actualCell.textContent).toBe("6");
+        expect(atLevel2.actualCell.classList).toContain("acceptable");
+        // "this page" displays the new level's limit.
+        expect(
+            getRow(container, "WordLengths", "ThisPageLC").maxCell.textContent,
+        ).toBe(`${kLevel2MaxGlyphsPerWord}`);
     });
 });
--- a/src/BloomBrowserUI/bookEdit/toolbox/readers/leveledReader/LeveledReaderToolControls.spec.tsx
+++ b/src/BloomBrowserUI/bookEdit/toolbox/readers/leveledReader/LeveledReaderToolControls.spec.tsx
@@ -6,69 +6,74 @@ import { getTheOneReaderToolsModel } from "../readerToolsModel";
 import ReadersSynphonyWrapper from "../ReadersSynphonyWrapper";
 import { LeveledReaderStats } from "./LeveledReaderToolControls";
 
-// The level we test against. Deliberately leaves most limits unset (0 = "this
-// level does not limit that statistic") so that we also cover the "no limit"
-// case, and so that the two statistics under test have limits that the book
-// stats below exceed.
+// The level we test against. It deliberately leaves most limits unset (0 means
+// "this level does not limit that statistic") so we also cover the no-limit case.
 const kMaxWordsPerSentence = 10;
 const kMaxGlyphsPerWord = 5;
 const kMaxAverageWordsPerSentence = 9;
 
-// Each value is unique so a cell can be located by its text.
 const bookStats = {
     levelNumber: 1,
     pageCount: 1,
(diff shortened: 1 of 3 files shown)


src/BloomBrowserUI/bookEdit/toolbox/readers/leveledReader/LeveledReaderToolControls.spec.tsx
andrew-polk, line 1: [Devin] **Bug**: New test determines pass/fail by parsing generated stylesheet colors, which the project forbids

The new test decided whether a statistic is flagged by digging the text color out of the generated stylesheets (`getColorOf` in `LeveledReaderToolControls.spec.tsx`), which `src/BloomBrowserUI/AGENTS.md` explicitly rules out in favour of markers such as css classes:

> Don't check for styles in tests as a way to know the status of something. That is fragile. If necessary have components add css classes or whatever that tests can check.

**Impact:** the check can start failing for reasons unrelated to the behavior it is meant to protect (a change in how Emotion groups, nests, minifies or orders rules, or an added `color` in the same rule), wasting developer time.

Devin's suggestion was to have `StatsRow` add a class such as `tooLarge`/`acceptable` — the names the pre-React code used — and assert on that instead.

src/BloomBrowserUI/bookEdit/toolbox/readers/leveledReader/LeveledReaderToolControls.tsx
andrew-polk, line 1: [Devin] **Investigate**: Third row lost its over-limit check in the same conversion but was not restored

Devin's analysis: pre-conversion (at commit `92a9d674b`), `updateActualCount` compared `actualSentenceCount` ("Sentence Counts / total in book") against `maxWordsPerSentenceOnThisPage()` and could colour it orange. The React version compares it against `model.maxSentencesPerBook()`, which is never populated by the leveled reader setup UI (`maxSentencesPerBook` exists only as a default 0 in `ReaderSettings.ts`, with no writer anywhere), so this row can never go orange — potentially the same class of regression this PR fixes for the two other rows.

Devin also noted that the old comparison was itself probably a copy/paste bug (a sentence *count* compared against a words-per-sentence limit), so restoring it verbatim may not be right.

src/BloomBrowserUI/bookEdit/toolbox/readers/leveledReader/LeveledReaderToolControls.spec.tsx
andrew-polk: [Claude Opus 5 (1M context) during preflight] Fixed in `411cca15d7`. Good catch — this is a documented project rule and the test was breaking it.

`StatsRow` now puts `tooLarge` or `acceptable` on the actual-value cell, driven by the same `isOverMax` call that picks the colour, and the test asserts on that class instead of reading any stylesheet. (Those are the pre-React names, which, as the old code's comment warned, are deliberately not substrings of each other.) Nothing about what the user sees changed.

While in there I also took the two related informational flags: the test now locates a row by its section header and row label rather than by the number it displays, so it no longer depends on every fixture value being unique across the panel — which also removed the special case for the one value that collided with a level maximum — and the fixture values are integers now, matching the shape of the real statistics.

src/BloomBrowserUI/bookEdit/toolbox/readers/leveledReader/LeveledReaderToolControls.tsx
andrew-polk: [Claude Opus 5 (1M context) during preflight] Not an issue — the flag is reading a baseline that predates the fix.

`92a9d674b` is from April 2021. Later that year, commit `31e9f29703` ("Fix check for max number of sentences in a book (BL-9885)") deliberately changed that comparison from `maxWordsPerSentenceOnThisPage()` to `maxSentencesPerBook()` — precisely because comparing a sentence count against a words-per-sentence limit was the copy/paste bug Devin suspects here. `origin/Version6.4`, i.e. the code that actually shipped as 6.4, uses `maxSentencesPerBook()`, and so does the React version. So this row behaves exactly as it did in 6.4 and there is nothing to restore.

The residual observation is fair but pre-existing and outside this fix: nothing ever writes `maxSentencesPerBook`, so "total in book" can never go orange in 6.4 either. The code already carries a comment saying as much. Raising that with the developer separately rather than expanding this PR's scope.

andrew-polk:
[Claude Opus 5 (1M context) during preflight] Consulted Devin on 2026-07-30 (three times, once per commit), most recently up to commit `e38ccc662af4f58c53c186c51ab84177c4d0cebe`.

It raised one Bug — the new test was deciding pass/fail by parsing the generated stylesheet, which this project's test guidance rules out. That is fixed in `411cca1` and Devin now reports it resolved; the thread above records it.

It also raised one Investigate flag, suggesting a third statistics row had lost its over-limit check in the same React conversion. That one is a false alarm — it compared against an April 2021 commit, and the comparison it wants restored was deliberately changed later by BL-9885. The thread above has the details.

Ten informational items across the three passes; none needed mirroring, but three led to real changes (the fixture no longer uses a non-integer for an integer statistic, the test no longer depends on every fixture number being unique, and there is now coverage for re-reading the limits when the level changes). One more — that a statistic of NaN would be treated as within-limit where 6.4 treated it as a violation — I assessed and left alone: it needs a book with no pages to reach, and such a value would render as "NaN" on screen regardless of its colour.

CI is green and no other review bot is active on this repo (`.coderabbit.yml` disables auto-review).

src/BloomBrowserUI/bookEdit/toolbox/readers/leveledReader/LeveledReaderToolControls.tsx
andrew-polk: [Claude Opus 5 (1M context)] Following up on the residual point above: the developer's decision is to keep it out of this PR and track it separately, so it is now filed as [BL-16636](https://issues.bloomlibrary.org/youtrack/issue/BL-16636) — "total in book" can never turn orange because nothing ever sets `maxSentencesPerBook`. It needs a product call (add the missing limit to Set Up Levels, or stop showing a Max column for that row), which is why it isn't a code change here.

StephenMcConnel (approved):
> @StephenMcConnel reviewed 3 files and all commit messages, and resolved 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8131)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on andrew-polk).

PR #8132, opened 2026-07-30
Convert Decodable Reader Setup Dialog to React

josiahwall:
This pr is for converting the decodable reader's setup dialog to React; the leveled reader's dialog still needs to be converted.

[Devin review](https://devinreview.com/BloomBooks/BloomDesktop/pull/8132)

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8132)

### DIFF
--- a/DistFiles/localization/en/Bloom.xlf
+++ b/DistFiles/localization/en/Bloom.xlf
@@ -5418,6 +5418,11 @@ Do you want to go ahead?</note>
         <source xml:lang="en">Previous and New Letters</source>
         <note>ID: ReaderSetup.SelectedLetters</note>
       </trans-unit>
+      <trans-unit id="ReaderSetup.NoSampleTextsYet" translate="no">
+        <source xml:lang="en">No sample texts yet. Add text files to the Sample Texts folder.</source>
+        <note>ID: ReaderSetup.NoSampleTextsYet</note>
+        <note>Shown in place of the file list on the Sample Words tab of the Decodable Reader setup dialog, when the collection's Sample Texts folder is empty. "Sample Texts folder" is the name of a folder in the collection and is also the label of the link just above this message.</note>
+      </trans-unit>
       <trans-unit id="ReaderSetup.SeparateWordsWithSpaces" translate="no">
         <source xml:lang="en">Separate words with spaces.</source>
         <note>ID: ReaderSetup.SeparateWordsWithSpaces</note>
--- a/DistFiles/localization/en/Bloom.xlf
+++ b/DistFiles/localization/en/Bloom.xlf
@@ -5418,6 +5418,11 @@ Do you want to go ahead?</note>
         <source xml:lang="en">Previous and New Letters</source>
         <note>ID: ReaderSetup.SelectedLetters</note>
       </trans-unit>
+      <trans-unit id="ReaderSetup.SeparateWordsWithSpaces" translate="no">
+        <source xml:lang="en">Separate words with spaces.</source>
+        <note>ID: ReaderSetup.SeparateWordsWithSpaces</note>
+        <note>Hint shown directly under the "New Sight Words" text box on the Decodable Stages tab of the Decodable Reader setup dialog. Tells the user how to delimit the sight words they type into that box.</note>
+      </trans-unit>
       <trans-unit id="ReaderSetup.SentenceHeader" sil:dynamic="true">
         <source xml:lang="en">Sentence</source>
         <note>ID: ReaderSetup.SentenceHeader</note>
--- a/DistFiles/localization/en/Bloom.xlf
+++ b/DistFiles/localization/en/Bloom.xlf
@@ -5345,6 +5345,16 @@ Do you want to go ahead?</note>
         <source xml:lang="en">Previous and New Letters</source>
         <note>ID: ReaderSetup.SelectedLetters</note>
       </trans-unit>
+      <trans-unit id="ReaderSetup.NoSampleTextsYet" translate="no">
+        <source xml:lang="en">No sample texts yet. Add text files to the Sample Texts folder.</source>
+        <note>ID: ReaderSetup.NoSampleTextsYet</note>
+        <note>Shown in place of the file list on the Sample Words tab of the Decodable Reader setup dialog, when the collection's Sample Texts folder is empty. "Sample Texts folder" is the name of a folder in the collection and is also the label of the link just above this message.</note>
+      </trans-unit>
+      <trans-unit id="ReaderSetup.SeparateWordsWithSpaces" translate="no">
+        <source xml:lang="en">Separate words with spaces.</source>
+        <note>ID: ReaderSetup.SeparateWordsWithSpaces</note>
+        <note>Hint shown directly under the "New Sight Words" text box on the Decodable Stages tab of the Decodable Reader setup dialog. Tells the user how to delimit the sight words they type into that box.</note>
+      </trans-unit>
       <trans-unit id="ReaderSetup.SentenceHeader" sil:dynamic="true">
         <source xml:lang="en">Sentence</source>
         <note>ID: ReaderSetup.SentenceHeader</note>
--- a/src/BloomBrowserUI/bookEdit/toolbox/readers/libSynphony/synphony_lib.ts
+++ b/src/BloomBrowserUI/bookEdit/toolbox/readers/libSynphony/synphony_lib.ts
@@ -29,6 +29,29 @@ import $ from "jquery";
 export function ResetLanguageDataInstance() {
     theOneLanguageDataInstance = new LanguageData();
 }
+
+/**
+ * Clears just the grapheme side of the one LanguageData, leaving the loaded vocabulary alone.
+ * Callers that may have changed the alphabet need this: addGrapheme only ever *adds*, so
+ * nothing otherwise removes a letter combination the user deleted, and stale entries here go
+ * on being counted as single letters by ReaderToolsModel.getWordLength. Clears exactly the
+ * fields ReadersSynphonyWrapper.loadSettings repopulates, so call it immediately before that.
+ *
+ * Known limitations of keeping the vocabulary, reviewed and accepted 2026-08 (BL-16607). Both
+ * come from the words outliving the reset, and both correct themselves on the next full refresh:
+ *   - The words were each split into graphemes when they were added and hold on to that split, so
+ *     after the alphabet changes they still carry the *old* segmentation. Narrower than the bug
+ *     this fixes -- word lengths are right again either way -- and re-splitting the whole
+ *     vocabulary here was judged not worth the cost.
+ *   - Because loadSettings re-adds the typed sample words afterwards, and addWord counts a repeat
+ *     as extra frequency, saving repeatedly in allowed-words mode inflates those counts. That
+ *     only affects frequency ordering, and allowed-words mode does not define stages from sample
+ *     words at all.
+ */
+export function ResetLanguageDataGraphemes() {
+    theOneLanguageDataInstance.GPCS = [];
+    theOneLanguageDataInstance.LanguageSortOrder = [];
+}
 var alwaysMatch: any[] = [];
 
 /**
--- a/src/BloomBrowserUI/bookEdit/toolbox/readers/libSynphony/synphony_lib.ts
+++ b/src/BloomBrowserUI/bookEdit/toolbox/readers/libSynphony/synphony_lib.ts
@@ -37,11 +37,16 @@ export function ResetLanguageDataInstance() {
  * on being counted as single letters by ReaderToolsModel.getWordLength. Clears exactly the
  * fields ReadersSynphonyWrapper.loadSettings repopulates, so call it immediately before that.
  *
- * Known limitation, reviewed and accepted 2026-08 (BL-16607): the words we keep were each split
- * into graphemes when they were added, and they hold on to that split. So after the alphabet
- * changes they still carry the *old* segmentation until something reloads them. That is narrower
- * than the bug this fixes (word lengths are right again either way), and it corrects itself on
- * the next full refresh; re-splitting the whole vocabulary here was judged not worth the cost.
+ * Known limitations of keeping the vocabulary, reviewed and accepted 2026-08 (BL-16607). Both
+ * come from the words outliving the reset, and both correct themselves on the next full refresh:
+ *   - The words were each split into graphemes when they were added and hold on to that split, so
+ *     after the alphabet changes they still carry the *old* segmentation. Narrower than the bug
+ *     this fixes -- word lengths are right again either way -- and re-splitting the whole
+ *     vocabulary here was judged not worth the cost.
+ *   - Because loadSettings re-adds the typed sample words afterwards, and addWord counts a repeat
+ *     as extra frequency, saving repeatedly in allowed-words mode inflates those counts. That
+ *     only affects frequency ordering, and allowed-words mode does not define stages from sample
+ *     words at all.
  */
 export function ResetLanguageDataGraphemes() {
     theOneLanguageDataInstance.GPCS = [];
--- a/src/BloomBrowserUI/bookEdit/toolbox/readers/libSynphony/synphony_lib.ts
+++ b/src/BloomBrowserUI/bookEdit/toolbox/readers/libSynphony/synphony_lib.ts
@@ -29,6 +29,18 @@ import $ from "jquery";
 export function ResetLanguageDataInstance() {
     theOneLanguageDataInstance = new LanguageData();
 }
+
+/**
+ * Clears just the grapheme side of the one LanguageData, leaving the loaded vocabulary alone.
+ * Callers that may have changed the alphabet need this: addGrapheme only ever *adds*, so
+ * nothing otherwise removes a letter combination the user deleted, and stale entries here go
+ * on being counted as single letters by ReaderToolsModel.getWordLength. Clears exactly the
+ * fields ReadersSynphonyWrapper.loadSettings repopulates, so call it immediately before that.
+ */
+export function ResetLanguageDataGraphemes() {
+    theOneLanguageDataInstance.GPCS = [];
+    theOneLanguageDataInstance.LanguageSortOrder = [];
+}
 var alwaysMatch: any[] = [];
 
 /**
--- a/src/BloomBrowserUI/bookEdit/toolbox/readers/libSynphony/synphony_lib.ts
+++ b/src/BloomBrowserUI/bookEdit/toolbox/readers/libSynphony/synphony_lib.ts
@@ -36,6 +36,12 @@ export function ResetLanguageDataInstance() {
  * nothing otherwise removes a letter combination the user deleted, and stale entries here go
  * on being counted as single letters by ReaderToolsModel.getWordLength. Clears exactly the
  * fields ReadersSynphonyWrapper.loadSettings repopulates, so call it immediately before that.
+ *
+ * Known limitation, reviewed and accepted 2026-08 (BL-16607): the words we keep were each split
+ * into graphemes when they were added, and they hold on to that split. So after the alphabet
+ * changes they still carry the *old* segmentation until something reloads them. That is narrower
+ * than the bug this fixes (word lengths are right again either way), and it corrects itself on
+ * the next full refresh; re-splitting the whole vocabulary here was judged not worth the cost.
  */
 export function ResetLanguageDataGraphemes() {
     theOneLanguageDataInstance.GPCS = [];
--- a/src/BloomBrowserUI/bookEdit/toolbox/readers/readerSetup/DecodableReaderSetupDialog.tsx
+++ b/src/BloomBrowserUI/bookEdit/toolbox/readers/readerSetup/DecodableReaderSetupDialog.tsx
@@ -71,12 +71,19 @@ const DecodableReaderSetupDialogLauncher: React.FunctionComponent<{
             );
         }
         const settingsToSave = prepareSettingsForSave(settings);
+        // Compare like with like: what we save has been normalized, so the "what did it
+        // look like before?" baseline has to be normalized the same way. Otherwise stored
+        // settings that happen to contain a comma or a doubled space look changed every
+        // time, and we needlessly re-read every sample text file.
+        const initialForComparison = prepareSettingsForSave(
+            initialSettings.current,
+        );
         void toolboxBundle
             .beginSaveChangedSettings(
                 settingsToSave,
-                initialSettings.current.moreWords,
-                initialSettings.current.letters,
-                initialSettings.current.useAllowedWords,
+                initialForComparison.moreWords,
+                initialForComparison.letters,
+                initialForComparison.useAllowedWords,
             )
             .then(close);
     };
--- a/src/BloomBrowserUI/bookEdit/toolbox/readers/readerSetup/DecodableReaderSetupDialog.tsx
+++ b/src/BloomBrowserUI/bookEdit/toolbox/readers/readerSetup/DecodableReaderSetupDialog.tsx
@@ -25,8 +25,8 @@ import { ReaderStage } from "../ReaderSettings";
 import { getTheOneReaderToolsModel } from "../readerToolsModel";
 import { DecodableStagesSetup } from "./DecodableStagesSetup";
 import {
-    cleanSpaceDelimitedList,
     cloneReaderSettings,
+    prepareSettingsForSave,
 } from "./decodableStagesUtils";
 import { kBloomBlue } from "../../../../utils/colorUtils";
 import { lightTheme } from "../../../../bloomMaterialUITheme";
@@ -70,14 +70,7 @@ const DecodableReaderSetupDialogLauncher: React.FunctionComponent<{
                 "The reader settings must load before they can be saved.",
             );
         }
-        const settingsToSave = cloneReaderSettings(settings);
-        settingsToSave.letters = cleanSpaceDelimitedList(
-            settingsToSave.letters,
-        );
-        for (const stage of settingsToSave.stages) {
-            stage.letters = cleanSpaceDelimitedList(stage.letters);
-            stage.sightWords = cleanSpaceDelimitedList(stage.sightWords);
-        }
+        const settingsToSave = prepareSettingsForSave(settings);
         void toolboxBundle
             .beginSaveChangedSettings(
                 settingsToSave,
--- a/src/BloomBrowserUI/bookEdit/toolbox/readers/readerSetup/DecodableReaderSetupDialog.tsx
+++ b/src/BloomBrowserUI/bookEdit/toolbox/readers/readerSetup/DecodableReaderSetupDialog.tsx
@@ -20,6 +20,7 @@ import {
     DialogOkButton,
 } from "../../../../react_components/BloomDialog/commonDialogComponents";
 import { useL10n } from "../../../../react_components/l10nHooks";
+import { useMountEffect } from "../../../../utils/useMountEffect";
 import type { ReaderSettings } from "../ReaderSettings";
 import { ReaderStage } from "../ReaderSettings";
 import { getTheOneReaderToolsModel } from "../readerToolsModel";
@@ -54,6 +55,14 @@ const DecodableReaderSetupDialogLauncher: React.FunctionComponent<{
         return copy;
     });
     const initialSettings = useRef(settings);
+
+    // Lock the edit view from here rather than from showDecodableReaderSetupDialog, so that the
+    // same component owns both turning the lock on and (in close, below) turning it off. If the
+    // settings check above throws, this never runs, and the user is not left staring at an edit
+    // view they cannot touch with no dialog to dismiss.
+    useMountEffect(() => {
+        postBoolean("editView/setModalState", true);
+    });
 
     const close = () => {
         setOpen(false);
@@ -168,7 +177,7 @@ const DecodableReaderSetupDialogLauncher: React.FunctionComponent<{
 /** Shows the React-hosted decodable reader setup dialog. */
 export const showDecodableReaderSetupDialog = (): void => {
     get("readers/io/readerSettingsEditForbidden", (result) => {
-        postBoolean("editView/setModalState", true);
+        // The edit view's modal lock is set by the dialog component itself, on mount.
         getWorkspaceBundleExports().ShowEditViewDialog(
             <DecodableReaderSetupDialogLauncher
                 editingForbiddenMessage={result.data || undefined}
--- a/src/BloomBrowserUI/bookEdit/toolbox/readers/readerSetup/DecodableReaderSetupDialog.tsx
+++ b/src/BloomBrowserUI/bookEdit/toolbox/readers/readerSetup/DecodableReaderSetupDialog.tsx
@@ -1,0 +1,183 @@
+import { css } from "@emotion/react";
+import { ThemeProvider } from "@mui/material/styles";
+import { useRef, useState } from "react";
+import * as React from "react";
+import {
+    getToolboxBundleExports,
+    getWorkspaceBundleExports,
+} from "../../../js/workspaceFrames";
+import { get, postBoolean } from "../../../../utils/bloomApi";
+import {
+    BloomDialog,
+    DialogBottomLeftButtons,
+    DialogBottomButtons,
+    DialogMiddle,
+    DialogTitle,
+} from "../../../../react_components/BloomDialog/BloomDialog";
+import {
+    DialogCancelButton,
+    DialogHelpButton,
+    DialogOkButton,
+} from "../../../../react_components/BloomDialog/commonDialogComponents";
+import { useL10n } from "../../../../react_components/l10nHooks";
+import type { ReaderSettings } from "../ReaderSettings";
+import { ReaderStage } from "../ReaderSettings";
+import { getTheOneReaderToolsModel } from "../readerToolsModel";
+import { DecodableStagesSetup } from "./DecodableStagesSetup";
+import {
+    cleanSpaceDelimitedList,
+    cloneReaderSettings,
+} from "./decodableStagesUtils";
+import { kBloomBlue } from "../../../../utils/colorUtils";
+import { lightTheme } from "../../../../bloomMaterialUITheme";
+
+let closeDialog: () => void = () => {};
+
+const DecodableReaderSetupDialogLauncher: React.FunctionComponent<{
+    editingForbiddenMessage?: string;
+}> = (props) => {
+    const model = getTheOneReaderToolsModel();
+
+    const [open, setOpen] = useState(true);
+    const [settings, setSettings] = useState<ReaderSettings>(() => {
+        const source = model.synphony?.source;
+        if (!source) {
+            throw new Error(
+                "The reader settings must load before opening setup.",
+            );
+        }
+
+        const copy = cloneReaderSettings(source);
+        if (copy.stages.length === 0) {
+            copy.stages.push(new ReaderStage("1"));
+        }
+        return copy;
+    });
+    const initialSettings = useRef(settings);
+
+    const close = () => {
+        setOpen(false);
+        closeDialog = () => {};
+        postBoolean("editView/setModalState", false);
+    };
+
+    closeDialog = close;
+
+    const save = () => {
+        const toolboxBundle = getToolboxBundleExports();
+        if (!toolboxBundle) {
+            throw new Error(
+                "The reader settings must load before they can be saved.",
+            );
+        }
+        const settingsToSave = cloneReaderSettings(settings);
+        settingsToSave.letters = cleanSpaceDelimitedList(
+            settingsToSave.letters,
+        );
+        for (const stage of settingsToSave.stages) {
+            stage.letters = cleanSpaceDelimitedList(stage.letters);
+            stage.sightWords = cleanSpaceDelimitedList(stage.sightWords);
+        }
+        void toolboxBundle
+            .beginSaveChangedSettings(
+                settingsToSave,
+                initialSettings.current.moreWords,
+                initialSettings.current.letters,
+                initialSettings.current.useAllowedWords,
+            )
+            .then(close);
+    };
+
+    const title = useL10n(
+        "Set up Decodable Reader Tool",
+        "ReaderSetup.SetUpDecodableReaderTool",
+    );
+
+    return (
+        <ThemeProvider theme={lightTheme}>
+            <BloomDialog
+                open={open}
+                onClose={close}
+                onCancel={close}
+                maxWidth={false}
+                css={css`
+                    .MuiPaper-root {
+                        border-radius: 8px;
+                        box-shadow: 0 12px 32px rgb(0 0 0 / 24%);
+                    }
+                    .MuiButton-containedPrimary {
+                        background-color: ${kBloomBlue};
+                    }
+                    .MuiButton-outlinedPrimary {
+                        border-color: ${kBloomBlue};
+                        color: ${kBloomBlue};
+                    }
+                `}
+            >
+                <DialogTitle title={title} />
+                <DialogMiddle
+                    css={css`
+                        height: min(600px, calc(100vh - 190px));
+                        width: min(1000px, calc(100vw - 64px));
+                        margin-left: -24px;
+                        margin-right: -24px;
+                        padding: 20px 24px;
+                        box-sizing: border-box;
+                        background-color: #f4f5f5;
+                        border-top: 1px solid #e5e5e5;
+                    `}
+                >
+                    {props.editingForbiddenMessage ? (
+                        <div>{props.editingForbiddenMessage}</div>
+                    ) : (
+                        <DecodableStagesSetup
+                            settings={settings}
+                            setSettings={setSettings}
+                            fontName={model.fontName}
+                            maxAllowedWords={model.maxAllowedWords}
+                        />
+                    )}
+                </DialogMiddle>
+                <DialogBottomButtons
+                    css={css`
+                        margin-left: -24px;
+                        margin-right: -24px;
+                        margin-bottom: -10px;
+                        padding: 14px 24px;
+                        width: calc(100% + 48px);
+                        box-sizing: border-box;
+                        background-color: white;
+                        border-top: 1px solid #e5e5e5;
+                    `}
+                >
+                    {!props.editingForbiddenMessage && (
+                        <DialogBottomLeftButtons>
+                            <DialogHelpButton helpId="Tasks/Edit_tasks/Decodable_Reader_Tool/Decodable_Stages_tab.htm" />
+                        </DialogBottomLeftButtons>
+                    )}
+                    {!props.editingForbiddenMessage && (
+                        <DialogOkButton default={true} onClick={save} />
+                    )}
+                    <DialogCancelButton />
+                </DialogBottomButtons>
+            </BloomDialog>
+        </ThemeProvider>
+    );
+};
+
+/** Shows the React-hosted decodable reader setup dialog. */
+export const showDecodableReaderSetupDialog = (): void => {
+    get("readers/io/readerSettingsEditForbidden", (result) => {
+        postBoolean("editView/setModalState", true);
+        getWorkspaceBundleExports().ShowEditViewDialog(
+            <DecodableReaderSetupDialogLauncher
+                editingForbiddenMessage={result.data || undefined}
+            />,
+        );
+    });
+};
+
+/** Closes the React-hosted decodable reader setup dialog. */
+export const closeDecodableReaderSetupDialog = (): void => {
+    closeDialog();
+};
--- a/src/BloomBrowserUI/bookEdit/toolbox/readers/readerSetup/DecodableReaderSetupDialog.tsx
+++ b/src/BloomBrowserUI/bookEdit/toolbox/readers/readerSetup/DecodableReaderSetupDialog.tsx
@@ -1,6 +1,6 @@
 import { css } from "@emotion/react";
 import { ThemeProvider } from "@mui/material/styles";
-import { useRef, useState } from "react";
+import { useEffect, useRef, useState } from "react";
 import * as React from "react";
 import {
     getToolboxBundleExports,
@@ -64,13 +64,29 @@ const DecodableReaderSetupDialogLauncher: React.FunctionComponent<{
         postBoolean("editView/setModalState", true);
     });
 
+    const hasClosed = useRef(false);
     const close = () => {
+        // Closing twice must not post the unlock twice. Closing does not unmount this
+        // component, so the handle below stays live afterwards; without this guard a later
+        // closeSetupDialog() would re-unlock the edit view even though something else may by
+        // then be relying on it staying locked.
+        if (hasClosed.current) {
+            return;
+        }
+        hasClosed.current = true;
         setOpen(false);
-        closeDialog = () => {};
         postBoolean("editView/setModalState", false);
     };
 
-    closeDialog = close;
+    // Publish the close handle from an effect, not from the render body: assigning module-level
+    // state while rendering runs again on every re-render (and twice under StrictMode). No
+    // dependency list, so the handle always points at the current close; cleared on unmount.
+    useEffect(() => {
+        closeDialog = close;
+        return () => {
+            closeDialog = () => {};
+        };
+    });
 
     const save = () => {
         const toolboxBundle = getToolboxBundleExports();
--- a/src/BloomBrowserUI/bookEdit/toolbox/readers/readerSetup/DecodableReaderSetupDialog.tsx
+++ b/src/BloomBrowserUI/bookEdit/toolbox/readers/readerSetup/DecodableReaderSetupDialog.tsx
@@ -40,6 +40,8 @@ const DecodableReaderSetupDialogLauncher: React.FunctionComponent<{
     const model = getTheOneReaderToolsModel();
 
     const [open, setOpen] = useState(true);
+    // 2 is the Decodable Stages tab, which is the one this dialog opens on.
+    const [curTab, setCurTab] = useState(2);
     const [settings, setSettings] = useState<ReaderSettings>(() => {
         const source = model.synphony?.source;
         if (!source) {
@@ -113,6 +115,14 @@ const DecodableReaderSetupDialogLauncher: React.FunctionComponent<{
             .then(close);
     };
 
+    let helpId =
+        "Tasks/Edit_tasks/Decodable_Reader_Tool/Decodable_Stages_tab.htm";
+    if (curTab === 0) {
+        helpId = "Tasks/Edit_tasks/Decodable_Reader_Tool/Letters_tab.htm";
+    } else if (curTab === 1) {
+        helpId = "Tasks/Edit_tasks/Decodable_Reader_Tool/Words_tab.htm";
+    }
+
     const title = useL10n(
         "Set up Decodable Reader Tool",
         "ReaderSetup.SetUpDecodableReaderTool",
@@ -160,6 +170,8 @@ const DecodableReaderSetupDialogLauncher: React.FunctionComponent<{
                             setSettings={setSettings}
                             fontName={model.fontName}
                             maxAllowedWords={model.maxAllowedWords}
+                            curTab={curTab}
+                            setCurTab={setCurTab}
                         />
                     )}
                 </DialogMiddle>
@@ -177,7 +189,7 @@ const DecodableReaderSetupDialogLauncher: React.FunctionComponent<{
                 >
                     {!props.editingForbiddenMessage && (
                         <DialogBottomLeftButtons>
-                            <DialogHelpButton helpId="Tasks/Edit_tasks/Decodable_Reader_Tool/Decodable_Stages_tab.htm" />
+                            <DialogHelpButton helpId={helpId} />
                         </DialogBottomLeftButtons>
                     )}
                     {!props.editingForbiddenMessage && (
--- a/src/BloomBrowserUI/bookEdit/toolbox/readers/readerSetup/DecodableReaderSetupDialog.tsx
+++ b/src/BloomBrowserUI/bookEdit/toolbox/readers/readerSetup/DecodableReaderSetupDialog.tsx
@@ -135,6 +135,7 @@ const DecodableReaderSetupDialogLauncher: React.FunctionComponent<{
                 onClose={close}
                 onCancel={close}
                 maxWidth={false}
+                data-testid="decodable-reader-setup-dialog"
                 css={css`
                     .MuiPaper-root {
                         border-radius: 8px;
--- a/src/BloomBrowserUI/bookEdit/toolbox/readers/readerSetup/DecodableReaderSetupDialog.tsx
+++ b/src/BloomBrowserUI/bookEdit/toolbox/readers/readerSetup/DecodableReaderSetupDialog.tsx
@@ -1,6 +1,6 @@
 import { css } from "@emotion/react";
 import { ThemeProvider } from "@mui/material/styles";
-import { useRef, useState } from "react";
+import { useEffect, useRef, useState } from "react";
 import * as React from "react";
 import {
     getToolboxBundleExports,
@@ -20,13 +20,14 @@ import {
     DialogOkButton,
 } from "../../../../react_components/BloomDialog/commonDialogComponents";
 import { useL10n } from "../../../../react_components/l10nHooks";
+import { useMountEffect } from "../../../../utils/useMountEffect";
 import type { ReaderSettings } from "../ReaderSettings";
 import { ReaderStage } from "../ReaderSettings";
 import { getTheOneReaderToolsModel } from "../readerToolsModel";
 import { DecodableStagesSetup } from "./DecodableStagesSetup";
 import {
-    cleanSpaceDelimitedList,
     cloneReaderSettings,
+    prepareSettingsForSave,
 } from "./decodableStagesUtils";
 import { kBloomBlue } from "../../../../utils/colorUtils";
 import { lightTheme } from "../../../../bloomMaterialUITheme";
@@ -39,6 +40,8 @@ const DecodableReaderSetupDialogLauncher: React.FunctionComponent<{
     const model = getTheOneReaderToolsModel();
 
     const [open, setOpen] = useState(true);
+    // 2 is the Decodable Stages tab, which is the one this dialog opens on.
+    const [curTab, setCurTab] = useState(2);
     const [settings, setSettings] = useState<ReaderSettings>(() => {
         const source = model.synphony?.source;
         if (!source) {
@@ -55,13 +58,37 @@ const DecodableReaderSetupDialogLauncher: React.FunctionComponent<{
     });
     const initialSettings = useRef(settings);
 
+    // Lock the edit view from here rather than from showDecodableReaderSetupDialog, so that the
+    // same component owns both turning the lock on and (in close, below) turning it off. If the
+    // settings check above throws, this never runs, and the user is not left staring at an edit
+    // view they cannot touch with no dialog to dismiss.
+    useMountEffect(() => {
+        postBoolean("editView/setModalState", true);
+    });
+
+    const hasClosed = useRef(false);
     const close = () => {
+        // Closing twice must not post the unlock twice. Closing does not unmount this
+        // component, so the handle below stays live afterwards; without this guard a later
+        // closeSetupDialog() would re-unlock the edit view even though something else may by
+        // then be relying on it staying locked.
+        if (hasClosed.current) {
+            return;
+        }
+        hasClosed.current = true;
         setOpen(false);
-        closeDialog = () => {};
         postBoolean("editView/setModalState", false);
     };
 
-    closeDialog = close;
+    // Publish the close handle from an effect, not from the render body: assigning module-level
+    // state while rendering runs again on every re-render (and twice under StrictMode). No
+    // dependency list, so the handle always points at the current close; cleared on unmount.
+    useEffect(() => {
+        closeDialog = close;
+        return () => {
+            closeDialog = () => {};
+        };
+    });
 
     const save = () => {
         const toolboxBundle = getToolboxBundleExports();
@@ -70,23 +97,31 @@ const DecodableReaderSetupDialogLauncher: React.FunctionComponent<{
                 "The reader settings must load before they can be saved.",
             );
         }
-        const settingsToSave = cloneReaderSettings(settings);
-        settingsToSave.letters = cleanSpaceDelimitedList(
-            settingsToSave.letters,
+        const settingsToSave = prepareSettingsForSave(settings);
+        // Compare like with like: what we save has been normalized, so the "what did it
+        // look like before?" baseline has to be normalized the same way. Otherwise stored
+        // settings that happen to contain a comma or a doubled space look changed every
+        // time, and we needlessly re-read every sample text file.
+        const initialForComparison = prepareSettingsForSave(
+            initialSettings.current,
         );
-        for (const stage of settingsToSave.stages) {
-            stage.letters = cleanSpaceDelimitedList(stage.letters);
-            stage.sightWords = cleanSpaceDelimitedList(stage.sightWords);
-        }
         void toolboxBundle
             .beginSaveChangedSettings(
                 settingsToSave,
-                initialSettings.current.moreWords,
-                initialSettings.current.letters,
-                initialSettings.current.useAllowedWords,
+                initialForComparison.moreWords,
+                initialForComparison.letters,
+                initialForComparison.useAllowedWords,
             )
             .then(close);
     };
+
+    let helpId =
+        "Tasks/Edit_tasks/Decodable_Reader_Tool/Decodable_Stages_tab.htm";
+    if (curTab === 0) {
+        helpId = "Tasks/Edit_tasks/Decodable_Reader_Tool/Letters_tab.htm";
+    } else if (curTab === 1) {
+        helpId = "Tasks/Edit_tasks/Decodable_Reader_Tool/Words_tab.htm";
+    }
 
     const title = useL10n(
         "Set up Decodable Reader Tool",
@@ -135,6 +170,8 @@ const DecodableReaderSetupDialogLauncher: React.FunctionComponent<{
                             setSettings={setSettings}
                             fontName={model.fontName}
                             maxAllowedWords={model.maxAllowedWords}
+                            curTab={curTab}
+                            setCurTab={setCurTab}
                         />
                     )}
                 </DialogMiddle>
@@ -152,7 +189,7 @@ const DecodableReaderSetupDialogLauncher: React.FunctionComponent<{
                 >
                     {!props.editingForbiddenMessage && (
                         <DialogBottomLeftButtons>
-                            <DialogHelpButton helpId="Tasks/Edit_tasks/Decodable_Reader_Tool/Decodable_Stages_tab.htm" />
+                            <DialogHelpButton helpId={helpId} />
                         </DialogBottomLeftButtons>
                     )}
                     {!props.editingForbiddenMessage && (
@@ -167,8 +204,34 @@ const DecodableReaderSetupDialogLauncher: React.FunctionComponent<{
 
 /** Shows the React-hosted decodable reader setup dialog. */
 export const showDecodableReaderSetupDialog = (): void => {
+    // The dialog cannot be built without the collection's reader settings: the component
+    // reads them in its very first useState and throws if they are not there. The tool that
+    // owns this button normally loads them when it is activated, but we have no guarantee
+    // that has happened (or has survived -- the model lives on the top window and is replaced
+    // when the frame that made it reloads) by the time the button is clicked. Without this,
+    // clicking "Set Up Stages" appears to do nothing at all, which is the same complaint that
+    // BL-16732 fixed for the legacy dialog in showSetupDialog. Loading is cheap when the
+    // settings are already in the model in hand.
+    //
+    // The setTimeout builds the dialog outside the promise chain, for the reason given in that
+    // same fix: a jQuery promise turns anything thrown inside a callback into a rejection
+    // nobody is listening for, so a failure in here would be swallowed and would once again
+    // look like a button that does nothing. Outside the chain it reaches our error reporting.
+    const toolboxBundle = getToolboxBundleExports();
+    if (!toolboxBundle) {
+        throw new Error(
+            "The toolbox frame must be present to open reader setup.",
+        );
+    }
+    toolboxBundle
+        .beginLoadSynphonySettings()
+        .always(() => window.setTimeout(beginShowSetupDialog, 0));
+};
+
+/** Builds and shows the dialog, once the reader settings are known to be loaded. */
+const beginShowSetupDialog = (): void => {
     get("readers/io/readerSettingsEditForbidden", (result) => {
-        postBoolean("editView/setModalState", true);
+        // The edit view's modal lock is set by the dialog component itself, on mount.
         getWorkspaceBundleExports().ShowEditViewDialog(
             <DecodableReaderSetupDialogLauncher
                 editingForbiddenMessage={result.data || undefined}
--- a/src/BloomBrowserUI/bookEdit/toolbox/readers/readerSetup/DecodableStagesSetup.tsx
+++ b/src/BloomBrowserUI/bookEdit/toolbox/readers/readerSetup/DecodableStagesSetup.tsx
@@ -17,7 +17,7 @@ import { useL10n } from "../../../../react_components/l10nHooks";
 import { Div, Span } from "../../../../react_components/l10nComponents";
 import BloomButton from "../../../../react_components/bloomButton";
 import { ReaderSettings, ReaderStage } from "../ReaderSettings";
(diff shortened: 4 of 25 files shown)


src/BloomBrowserUI/bookEdit/toolbox/readers/readerTools.ts
```diff
@@ -389,16 +389,16 @@ function refreshSettingsExceptSampleWords(newSettings) {
...
+ * Refreshes the reader setup and, for sample-word mode, rebuilds its language data from the sample files.
  * Returns a promise which is resolved when all the sample words files are loaded and the model is ready to use.
  */
 function beginRefreshEverything(settings: ReaderSettings): JQueryPromise<void> {
-    // reset the file and word list
-    ResetLanguageDataInstance();
-    getTheOneReaderToolsModel().allWords = {};
+    // Allowed-word-list mode does not use sample-word data. Keep it available for
+    // a setup-dialog preview if the user temporarily switches back to stages.
+    if (!settings.useAllowedWords) {
```
andrew-polk: **Devin (Bug): letters you delete keep affecting word analysis in allowed-words mode.**

This guard means that when a collection defines its stages by allowed-word lists, the shared Synphony language data is no longer rebuilt when settings change. `ReadersSynphonyWrapper.loadSettings` only ever *adds* graphemes, so a letter combination the user removes from the alphabet stays registered for the rest of the session. `ReaderToolsModel.getWordLength` builds its multi-letter grapheme list from that same data, so a deleted digraph still counts as a single letter and leveled-reader word-length checks can be wrong until Bloom restarts.

The guard is deliberate - it is what lets the setup dialog still preview matching words if the user flips back to stages mode without saving. So this is a real trade-off rather than a slip, which is why I have not touched it. The likely fix is to keep rebuilding the grapheme/GPCS side while separately preserving the loaded sample words the preview needs.

_Left open pending the developer's decision._

src/BloomBrowserUI/bookEdit/toolbox/readers/readerSetup/decodableStagesUtils.ts
```diff
@@ -0,0 +1,114 @@
...
+
+/** Returns whether a word contains only taught or language-defined allowed graphemes. */
+export const hasOnlyKnownGraphemes = (
+    word: string,
+    allGpcs: string[],
+    knownGpcs: string[],
+): boolean => {
+    const allowedGpcs = new Set(knownGpcs.map((gpc) => gpc.toLowerCase()));
+    for (const value of [
+        theOneLanguageDataInstance["AlwaysMatch"],
```
andrew-polk: **Devin (Bug): these special-symbol fields are always empty here, so some typed words vanish from the preview.**

Confirmed. `LanguageData` in synphony_lib.ts does not declare `AlwaysMatch`, `SyllableBreak`, `StressSymbol` or `MorphemeBreak` at all - they exist only if a collection's imported Synphony data supplies them. On top of that, this dialog runs in the workspace bundle, which has its own separate module instance of `synphony_lib`; the populated language data lives in the toolbox iframe, which is exactly why matching words are fetched via `getToolboxBundleExports().getDecodableStageMatchingWords`. So in the dialog's copy these four fields are always `undefined` and this loop does nothing.

User-visible effect, only for collections that define one of those symbols: a word the user *typed* containing a syllable-break or stress mark is filtered out of the matching-words preview, while a word containing the same mark that came from a sample text is kept - two halves of one list disagreeing. Narrow, but real. The fix is presumably to read these values from the toolbox bundle, as the matching words already are.

_Left open pending the developer's decision._

andrew-polk (commented):
**Claude Opus 5 (during `/preflight` on Andrew's machine) - Devin consultation log**

Devin reviewed this PR at `3e0c2978`. Its passes raised 9 bugs and 22 flags; most were either already fixed on the branch or have been fixed during this preflight (see the two commits above). What follows are the findings that are still open and need a human decision, one thread each.

No other review bot ran: this repo's `.coderabbit.yml` sets `auto_review.enabled: false`, and there is no Greptile review on this PR.

**Fixed during preflight:** typed sample words saved with line breaks (Devin's only *severe* finding - the words silently stopped counting as decodable); the hard-coded English hint under the sight-words box; the missing accessible names on the three text boxes; the dropped doc comment on the word-list cleaner; fail-fast when the toolbox bundle is missing; the props type aliases. Plus two things Devin caught in my own first preflight commit: sample words losing the legacy de-duplication, and `save()` comparing cleaned values against uncleaned ones.

**Reported from a stale commit, already fixed before preflight:** stage removal not deleting its word-list file, the stacked ternaries, the legacy 3-argument save path, and the dead LESS / `reactHost` block. I verified each against the current code rather than trusting the line numbers.

src/BloomBrowserUI/bookEdit/toolbox/readers/readerSetup/DecodableStagesSetup.tsx
```diff
@@ -0,0 +1,1086 @@
...
+        // The reader model invokes this only after it has finished loading sample-text words.
+        toolbox.addWordListChangedListener(listenerName, () => {
+            setWordListVersion((version) => version + 1);
+        });
+
+        return () => toolbox.removeWordListChangedListener(listenerName);
+    });
+
+    useMountEffect(() => {
+        const configuredLetters = new Set(
```
andrew-polk, line 616: **Devin (Investigate): stage letters are pruned every time the Stages tab is re-mounted.**

Only the active tab is rendered, so switching to Letters and back re-runs this mount effect, and any letter not currently in the alphabet box is silently dropped from every stage. The hazard is a transient edit: if the user is part-way through retyping a letter in the Letters tab and returns to Stages, that letter is removed from the stages with no undo.

Reconciling stages against the alphabet is clearly intended; the question is whether it should happen on every remount, or only on OK / an explicit alphabet change.

_Left open pending the developer's decision._

src/BloomBrowserUI/bookEdit/toolbox/readers/readerSetup/DecodableStagesSetup.tsx
```diff
@@ -0,0 +1,1086 @@
...
+                                css={css`
+                                    margin-bottom: 8px;
+                                    ${commonHeaderStyles}
+                                `}
+                            >
+                                1) Type Words Here
+                            </Div>
+                            <ReaderDialogTextarea
+                                updateSettings={updateMoreWords}
+                                value={props.settings.moreWords}
```
andrew-polk, line 440: **Devin (Investigate): sample words are no longer shown one per line.**

Confirmed by hand in a running Bloom: type two words on separate lines, press OK, reopen the dialog, and they come back on one wrapped line. The legacy dialog deliberately converted the stored space-delimited list back to one-per-line for display (`data.moreWords.replace(/ /g, "\n")` in readerSetup.io.ts).

The storage half of this is now fixed, so the words are saved correctly as separate words either way; this is purely a presentation decision for the redesigned dialog. My suggestion would be to restore one-per-line, since that is the habit existing users have, but it is a UI call rather than a bug.

_Left open pending the developer's decision._

src/BloomBrowserUI/bookEdit/toolbox/readers/readerSetup/DecodableReaderSetupDialog.tsx
```diff
@@ -0,0 +1,183 @@
...
+    closeDialog = close;
+
+    const save = () => {
+        const toolboxBundle = getToolboxBundleExports();
+        if (!toolboxBundle) {
+            throw new Error(
+                "The reader settings must load before they can be saved.",
+            );
+        }
+        const settingsToSave = prepareSettingsForSave(settings);
```
andrew-polk, line 100: **Devin (Investigate): the saved file now contains whatever the live settings object contained.**

The legacy dialog rebuilt a fresh `ReaderSettings` from the DOM, so only the handful of fields it knew about were written back. This dialog clones `model.synphony.source` and posts the whole clone. Two consequences worth an explicit yes/no: any extra runtime fields sitting on stage objects would now be persisted, and the old "do not save a stage with no data" rule is gone, so an empty stage created in the dialog is saved (and an empty stage 1 is deliberately materialized when a collection has none).

Both look intentional for the new always-a-row UI, but they change the shape of the persisted settings file, so worth confirming deliberately.

_Left open pending the developer's decision._

src/BloomBrowserUI/bookEdit/toolbox/readers/readerSetup/decodableStagesUtils.ts
```diff
@@ -0,0 +1,114 @@
...
+    settingsToSave.moreWords = cleanSampleWordList(settingsToSave.moreWords);
+    for (const stage of settingsToSave.stages) {
+        stage.letters = cleanSpaceDelimitedList(stage.letters);
+        stage.sightWords = cleanSpaceDelimitedList(stage.sightWords);
+    }
+    return settingsToSave;
+};
+
+/** Splits a word into configured graphemes using Synphony's legacy matching order. */
+const getGpcForm = (word: string, allGpcs: string[]): string[] => {
```
andrew-polk, line 93: **Devin (Investigate): this re-implements grapheme splitting that Synphony already does - mitigated with tests, one question left.**

Synphony has `LanguageData.getGpcForm`, which does the same longest-suffix-first split with the same surrogate-pair handling. Having two copies matters because a word typed into the dialog is judged decodable by *this* code, while a word from a sample text is judged by Synphony's - so if they drift, the same word can get two different answers within one list.

I added pinning tests in `decodableStagesUtilsSpec.ts` that run both implementations over the same words and assert they agree, including a guard so the suite cannot pass vacuously. Comparing them line by line I found one real difference: this version lower-cases the grapheme list and Synphony's does not, which makes this one slightly *more* forgiving for an alphabet entered with capitals. Harmless today, but a genuine divergence, so flagging it rather than silently changing either side.

_Left open pending the developer's decision._

src/BloomBrowserUI/bookEdit/toolbox/readers/readerTools.ts
andrew-polk: **Fixed** in 6e3e086 — the developer chose to fix this now rather than defer it.

The reset is now split in two. A new `ResetLanguageDataGraphemes` clears only the letter side of the shared Synphony data, so the alphabet is rebuilt on every save even in allowed-word-list mode, while the loaded sample words — the thing the original guard was protecting — are left in place.

Verified in a running Bloom: switching a collection to allowed-word lists and back restores all ten matching words with no restart, so the preview data really does survive. Three unit tests in `language_dataSpec.js` pin the reset itself: the graphemes are cleared, the vocabulary is not, and the instance is still usable afterwards.

src/BloomBrowserUI/bookEdit/toolbox/readers/readerSetup/decodableStagesUtils.ts
andrew-polk: **Fixed** in 6e3e086.

These values now come from the toolbox frame through a new `getSynphonyAlwaysMatchSymbols` export, exactly the way matching words already do, and `hasOnlyKnownGraphemes` takes them as an argument instead of reaching for a module global — which also makes them testable.

Verified over CDP that the new export is reachable across the bundle boundary from the dialog's frame (it returns an empty list for a collection whose language data defines none of these symbols, which is the expected answer here). I could not produce the user-visible symptom locally, because none of the collections on this machine define a syllable-break or stress symbol; the allow-a-symbol and reject-a-symbol paths are covered by unit tests.

src/BloomBrowserUI/bookEdit/toolbox/readers/readerSetup/DecodableStagesSetup.tsx
andrew-polk: **Leaving as is**, decided by the developer, who did reproduce it:

> I did reproduce this, but it also seems to reproduce in the existing dialog. At this point, it doesn't feel worth the extra complication it would take to figure out how to really make it work correctly.

So it is not a regression from the dialog being converted, and the cost of making the reconciliation precise is not worth it right now. Recorded as a comment above the effect so the next person reading it knows the behavior was noticed and deliberately accepted, rather than overlooked.

src/BloomBrowserUI/bookEdit/toolbox/readers/readerSetup/DecodableStagesSetup.tsx
andrew-polk: **Leaving as is**, by the developer's decision: the single-line display is adopted deliberately as part of the redesign, rather than restoring the old one-per-line presentation.

Worth being explicit that this is only about presentation — the storage half was fixed earlier in this preflight, so words typed one per line are saved and reloaded as separate words either way. QA has been told to expect the single-line display so they do not report it as a bug.

src/BloomBrowserUI/bookEdit/toolbox/readers/readerSetup/DecodableReaderSetupDialog.tsx
andrew-polk: **Still open — the developer wants to discuss this one**, so leaving the thread unresolved.

One part of it can be settled with evidence, though. I checked the reader settings files of all 25 collections on this machine, including one just written by this new dialog. Every stage in every file has exactly the keys `name`, `letters`, `sightWords`, `allowedWordsFile` — the file the new dialog wrote is key-for-key identical to the ones the legacy dialog wrote. The reason is that the dialog clones `model.synphony.source`, whose `stages` are the raw objects parsed from the settings file; the `words` and `allowedWords` arrays live on the separate `ReaderStage` instances in `synphony.stages`, which are never cloned. So the "extra runtime fields could leak into the file" half of this finding does not happen in practice.

That leaves one real difference: the legacy dialog dropped a stage that had no letters, no sight words and no word-list file, and renumbered the rest. The new dialog saves it.

src/BloomBrowserUI/bookEdit/toolbox/readers/readerSetup/decodableStagesUtils.ts
andrew-polk: **Fixed** in 6e3e086 — the duplicate is gone, and my earlier estimate of the difficulty was wrong.

I had said this would be moderate work because typed words are not in the language data yet, so there would be nothing to segment them against. That was mistaken: `LanguageData.getGpcForm` takes the grapheme list as an *argument* and reads no state off the instance it hangs from. The alphabet currently in the dialog is exactly what that argument is for, so the dialog can call Synphony's own routine directly. It was a few lines, and it removed about 24 lines of duplicated logic.

A word typed into the dialog and a word from a sample text are now segmented by the same code, so they cannot disagree. That also retires the lower-casing difference noted above, since there is no longer a second implementation to differ. The tests that used to compare the two implementations would now be comparing one thing with itself, so they assert the decodability outcomes outright instead, plus one check that Synphony really does split "chat" as ch + a + t.

andrew-polk (commented):
**Claude Opus 5 (during `/preflight`, round 2) — Devin consultation log**

Devin re-reviewed at `91864be` and raised **no new bugs against this commit**. Every bug it still lists carries the SHA of a superseded commit; I checked each against the current code.

The developer's decisions from the first round are now implemented and each of those threads has a reply recording the outcome. One thread is deliberately left open for discussion (the persisted-settings shape), and Devin's later passes surfaced three further things worth a decision — one thread each below.

Round 2 also fixed one bug **in my own round-1 work**, which Devin caught: `getSynphonyAlwaysMatchSymbols` filtered on `typeof symbol === "string"` and so silently dropped an array-valued `AlwaysMatch` — the very symptom the function was added to remove. Fixed in `91864be` with four tests; the array case fails against the previous implementation.

Gate at `91864be`: typecheck pass, lint 0 errors, **612 TypeScript tests pass** (up from 574 at the start of preflight — 38 added across this run), CI green. The C# suite was run earlier this preflight (2981 passed; the one failure needs Reading App Builder's build tools, which are not installed on this machine); nothing since has touched C# or the XLF files.

src/BloomBrowserUI/bookEdit/toolbox/readers/libSynphony/synphony_lib.ts
```diff
@@ -29,6 +29,18 @@ import $ from "jquery";
...
 }
+
+/**
+ * Clears just the grapheme side of the one LanguageData, leaving the loaded vocabulary alone.
+ * Callers that may have changed the alphabet need this: addGrapheme only ever *adds*, so
+ * nothing otherwise removes a letter combination the user deleted, and stale entries here go
+ * on being counted as single letters by ReaderToolsModel.getWordLength. Clears exactly the
+ * fields ReadersSynphonyWrapper.loadSettings repopulates, so call it immediately before that.
+ */
+export function ResetLanguageDataGraphemes() {
```
andrew-polk, line 51: **Devin (Investigate): words kept in allowed-words mode still carry segmentation from the old alphabet.**

A fair point about the fix immediately above this. Clearing the grapheme list means `getWordLength` stops counting a deleted letter combination as one letter, which was the reported bug. But the sample words we deliberately keep were segmented when they were loaded, and each stores that segmentation. So after the alphabet changes in allowed-word-list mode, those stored segmentations still reflect the old alphabet until something reloads the words.

What that means in practice: it is strictly better than before (the wrong word-length counts are gone), and it only shows up if a user changes the alphabet while in allowed-words mode and then switches back to stages mode *without* saving, where a sample-text word might be judged against its old segmentation. Fully fixing it means re-segmenting the kept vocabulary, or giving up the preview by doing the full reset again.

I have not done either, because that is a wider change than the decision I was asked to implement and it re-opens the trade-off the developer just settled.

_Left open pending the developer's decision._

src/BloomBrowserUI/bookEdit/toolbox/readers/readerTools.ts
```diff
@@ -472,6 +481,71 @@ export function addWordListChangedListener(
...
+    );
+    return symbols.filter(
+        (symbol): symbol is string =>
+            typeof symbol === "string" && symbol !== "",
+    );
+}
+
+/** Gets the loaded sample words decodable with the given graphemes. */
+export function getDecodableStageMatchingWords(knownGpcs: string[]): string[] {
+    return getTheOneReaderToolsModel().selectWordsFromSynphony(
```
andrew-polk, line 528: **Devin (Investigate): this asks Synphony a different question than the legacy dialog did.**

Confirmed by reading both. The first argument here is `true`, which routes to `selectGPCWordNamesWithArrayCompare`. The legacy dialog passed `false` (readerTools.ts, the `Words` message handler), and the toolbox's own `getStageWords` also passes `false` — both of which route to `selectGPCWordsFromCache`. So the dialog's matching-words list is now produced by a different libSynphony function than the one the rest of the reader code uses.

Evidence that they agree in practice: with a 13-stage French collection I compared the dialog's list against the toolbox's own "Sample words in this stage" panel for the same stage and got exactly the same five words (`bird, dog, frog, in, log`), and a word needing a later stage's letter correctly appeared only at that later stage. So this is not producing a visible discrepancy here.

I have not changed it — it is pre-existing on this branch rather than something preflight introduced, and confirming the two functions are equivalent in general is a bigger piece of work than reading them side by side.

_Left open pending the developer's decision._

src/BloomBrowserUI/bookEdit/toolbox/readers/readerSetup/DecodableStagesSetup.tsx
```diff
@@ -0,0 +1,1099 @@
...
+    cloneReaderSettings,
+    hasOnlyKnownGraphemes,
+} from "./decodableStagesUtils";
+import { Link } from "../../../../react_components/link";
+import { MuiRadio } from "../../../../react_components/muiRadio";
+import { BloomTooltip } from "../../../../react_components/BloomToolTip";
+import { ReaderDialogTextarea } from "./ReaderDialogTextarea";
+
+const kMutedText = "#707477";
+const kSampleTextFileExtensions = ["txt", "js", "json"];
```
andrew-polk: **Devin (Bug, non-severe): a sample text file of an unexpected type just vanishes from the list.**

The dialog filters the Sample Texts folder listing to this hard-coded set of extensions, so a file with any other extension is simply not shown — no message, no hint that it was ignored. Two things worth deciding:

- The model already has its own notion of which files are readable (`getReadableFileExtensions`, used by `ReaderToolsModel.beginSetTextsList`). Having a second, separate list here means the dialog's listing can drift out of step with the files Bloom actually reads.
- The set is also case-sensitive as written, so a `.TXT` file would be filtered out.

This is pre-existing on the branch rather than something preflight introduced, and which extensions *should* appear is a product question, so I have not changed it.

_Left open pending the developer's decision._

src/BloomBrowserUI/bookEdit/toolbox/readers/readerSetup/DecodableReaderSetupDialog.tsx
andrew-polk: **Settled** in 8a7f5c6, and the evidence narrowed it to one question.

I inspected the reader settings files of all 25 collections on this machine, including one this dialog had just written. Every stage in every file has exactly `name`, `letters`, `sightWords`, `allowedWordsFile` — key-for-key identical to the files the legacy dialog wrote. The `words`/`allowedWords` arrays live on separate `ReaderStage` instances that are never cloned, so the "extra runtime fields could leak into the file" concern does not happen in practice.

That left only the drop-empty-stages rule, and the developer chose to restore it: `prepareSettingsForSave` now omits a stage with no letters, no sight words and no word-list file, and renumbers the survivors. Verified in a running Bloom — added an empty 14th stage, pressed OK, and the file still has 13 stages named 1..13.

Worth recording: this means the empty Stage 1 the dialog shows a collection with no stages is not saved until something is entered into it. That is also what the old dialog did.

src/BloomBrowserUI/bookEdit/toolbox/readers/libSynphony/synphony_lib.ts
andrew-polk: **Accepted as a known limitation**, by the developer's decision, and recorded in a comment right here in 8a7f5c6.

The reasoning: the reported bug is fixed (word lengths are correct again either way), what remains is narrower, and it corrects itself on the next full refresh. Re-splitting the whole kept vocabulary would be a wider change in shared reader code and would re-open the trade-off that was just settled.

Devin subsequently raised a second consequence of the same "words outlive the reset" design — `loadSettings` re-adds the typed sample words afterwards, and `addWord` counts a repeat as extra frequency, so saving repeatedly in allowed-words mode inflates those counts. That predates the grapheme-side split and only affects frequency ordering, in a mode that does not define stages from sample words at all. It is now documented in the same comment rather than fixed.

src/BloomBrowserUI/bookEdit/toolbox/readers/readerTools.ts
andrew-polk: **Investigated, and there is nothing to fix** — noted in a comment in 8a7f5c6 so the next reader does not re-raise it.

The two entry points only *look* different. `selectGPCWordNamesWithArrayCompare` and `selectGPCWordsFromCache` both call libSynphony's `selectGPCWordsWithArrayCompare` with identical arguments; the first then plucks `Name` off the results, and the second memoizes them through `theOneWordCache`. So the selection logic is literally the same routine, which is why the dialog's list and the toolbox's panel produced the same five words in testing.

The only practical difference is caching: the dialog's path is uncached, so it always reflects the current data (and, as Devin separately notes, does more work per settings change). No behavior change made.

andrew-polk:
**Claude Opus 5 (during `/preflight`, round 3) — consultation log**

Devin re-reviewed at `8a7f5c6` and raised **no new bugs**. Every bug it still lists carries the SHA of a superseded commit. CI is green and the front-end suite is **615 passing**.

**Round-3 decisions, all now implemented:**
- **Restored the legacy drop-empty-stages rule** on save, with renumbering. Verified in a running Bloom: added an empty 14th stage, pressed OK, and the file still has 13 stages named 1..13.
- **Accepted** the kept-vocabulary limitations (stale grapheme segmentation, and inflated word-frequency counts on repeated saves in allowed-words mode) and documented both where the split happens.
- **Investigated** the two Synphony entry points and found nothing to fix — they call the same routine with the same arguments; noted so it is not re-raised.

**Two things worth flagging for someone other than this PR.**

*1. Master's own leveled-reader tests were failing, and this branch had to unbreak them to go green.* They fail the same way on `origin/master`, independently of this PR:
- `LeveledReaderStats` was declared without `export`, though master's own `LeveledReaderToolControls.spec.tsx` imports it by name — all 7 tests in the file died on render. Added the export.
- The spec still expected an empty Max cell where there is no limit to display. BL-16585 (`e6b66aad`, 3 Aug) changed that cell to render an em dash so the column stays aligned, and did not update the spec, which predates it (`e38ccc66`, 30 Jul). Updated the two expectations and the now-inaccurate test name.

**Someone from BL-16585 should sanity-check that second one** — it assumes the em dash is the intended display, which is what the component does today. If blank was intended, the component is what needs changing, not the spec.

*2. Master's shared `ReaderSetupButton` collided with this branch* (BL-16585 extracted the Set Up Stages/Levels button while this branch was rewiring it). Resolved by taking master's component and styling wholesale and giving it an optional `onClick`: the Decodable Reader passes the action that opens the converted React dialog, and the Leveled Reader keeps falling back to the legacy href. Verified both live — the decodable button opens the React dialog with all 13 stages, and Set Up Levels still opens the old jQuery "Set up Leveled Reader Tool" dialog.

One thread is still open, on the sample-text extension filter, pending discussion with the developer.

andrew-polk:
**Claude Opus 5 (during `/preflight`, after rebase) — consultation log**

Rebased onto master and force-pushed (`--force-with-lease`); the branch is now **9 linear commits** with no merge commits. Devin re-reviewed and found **no new bugs** at `6e3e779`; CI is green and the front-end suite is **618 passing**.

**Master's test fix landed, and mine dropped out.** `246991f8b0` did exactly what my earlier commit did — added the missing `export` on `LeveledReaderStats` and introduced `kNoMaxShown = "—"`, even renaming the same test. My commit became redundant and was skipped during the replay. That also settles the open question from last round about whether the em dash was intended: the BL-16585 author independently reached the same conclusion, so nothing is left to decide there.

**The shared `ReaderSetupButton` reconciliation had to be redone**, because it had lived only inside a merge commit's conflict resolution and rebase drops merges. It now sits in the conversion commit itself. Re-verified live afterwards: the decodable button opens the React dialog with all 13 stages, and Set Up Levels still opens the legacy jQuery "Set up Leveled Reader Tool" dialog.

**Two further bugs, found by Devin on the rebased branch and fixed:**

1. *The edit view could be left locked with no way out* (`025fc05`). Opening the dialog locked the edit view before the dialog rendered, while only the dialog itself unlocks. A failed render therefore left the user unable to touch the book until Bloom restarted — and it was reachable, since clicking Set Up Stages before the reader settings finish loading makes the component throw. The lock now comes from a mount effect, so a failed render locks nothing.

2. *Closing could later unlock the edit view spuriously* (`6e3e779`). `close()` tried to disarm its own handle, but the handle was re-armed by the very re-render that closing causes, so a later close request re-ran the unlock. Closing is now idempotent and the handle is published from an effect rather than during render — which also removes the assign-module-state-while-rendering pattern Devin flagged in its first review.

Both verified in a running Bloom by counting the actual `setModalState` requests: one on open, a second on Cancel, and still only two after a further close request, where previously a third would have fired.

**One item remains open** — the thread on the sample-text extension filter. Worth noting what that finding is and is not: the extension list and its case-sensitivity are *identical* to the pre-PR dialog, so neither is a regression. What genuinely changed is that the old dialog listed unreadable files with an explanation ("Cannot read this format" / "File needs .TXT extension") plus a help link, and showed a message for an empty folder; the new one filters them out silently. Those strings still exist and are translated, but are now referenced only from the legacy pug. Awaiting the developer's call.

andrew-polk:
**Claude Opus 5 (during `/preflight`) — last Devin item, assessed and not actioned**

Devin's final pass at `942c43ad` raised one thing: that some new helpers use `function` declarations where the project asks for arrow functions.

Assessed as not worth a change, for two reasons:

- AGENTS.md scopes that rule to React — "Always use arrow functions and function components in **React**" — and the two files involved are plain TypeScript modules, not components.
- Each file is already internally consistent, which seems the more useful kind of consistency here. `decodableStagesUtils.ts` is arrow-style throughout except `cleanSpaceDelimitedList`, which is a `function` because it was moved verbatim from `readerSetup.io.ts` — Devin itself asked earlier in this review that the move leave the code (and its doc comment) untouched. `readerToolsModel.ts` is a class module whose existing top-level exports are `function` declarations, and the two helpers I added there match them.

Recording it here rather than silently dropping it. Happy to convert them if the team reads the rule more broadly than I have.

Everything else from Devin is fixed; this was the only item left, and CI plus the 624-test front-end suite are green.

src/BloomBrowserUI/bookEdit/toolbox/readers/readerSetup/DecodableStagesSetup.tsx
andrew-polk: **Fixed**, by the developer's decision: restore the explanation rather than keep filtering these files out silently.

Files Bloom cannot read are listed again with the reason beside them — "Cannot read this format", or "File needs .TXT extension" when there is no extension at all — and the "Help exporting and converting files to use as sample texts" link reappears whenever such a file is present. All three strings already existed and were already translated; they were simply orphaned to the legacy pug. They use the same colour the legacy stylesheet did (`@WarningColor`, i.e. `kBloomRed`).

Two things worth recording about this finding, since neither turned out to be what it looked like:

- **The extension list and its case-sensitivity were not regressions.** The old dialog used the same `txt`/`js`/`json` list, and compared case-sensitively too, so `.TXT` was rejected before this PR as well. The developer asked for that to be fixed regardless, so extension matching is now case-insensitive — and to make "everywhere" true rather than coincidental, the list has one home (`kReadableSampleTextFileExtensions`), which the model filters through and which the dialog now asks the toolbox about instead of keeping the private copy that prompted your duplication note.
- **The empty-folder message never existed.** `#please-add-texts` was an empty label with no l10n key that nothing ever filled in, so the legacy "please add texts" path rendered nothing. It is a genuinely new string (`ReaderSetup.NoSampleTextsYet`), not a restoration.

Verified in a running Bloom with four files in the folder: `sample text.txt` and `UPPER.TXT` listed clean, `notes.docx` labelled "Cannot read this format", a file named `noextension` labelled "File needs .TXT extension", and the help link present; emptying the folder then showed the new message and hid the link.

A follow-up bug of my own in that work — a dot in a *folder* name being read as a file type, which mislabelled extension-less files — was caught by Devin and fixed in `6efc5fab`, with a test that fails against the previous version.

andrew-polk (commented):
> @andrew-polk resolved 9 discussions.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8132)* status: 0 of 27 files reviewed, all discussions resolved (waiting on josiahwall).

src/BloomBrowserUI/bookEdit/toolbox/readers/readerSetup/DecodableStagesSetup.tsx
```diff
@@ -0,0 +1,1168 @@
...
+
+        const fileIsUsedByAnotherStage = props.settings.stages.some(
+            (oneStage, index) =>
+                index !== props.curStageIndex &&
+                oneStage.allowedWordsFile === fileName,
+        );
+        if (!fileIsUsedByAnotherStage) {
+            axios.delete(`${getBloomApiPrefix()}readers/io/allowedWordsList`, {
+                params: { fileName },
+            });
```
andrew-polk: **Devin (Bug): pressing Cancel still deletes the word list file.**

Clicking the trash icon next to a stage's allowed-words file deletes the file from the collection straight away, before anything is saved. If the user then presses Cancel, the settings on disk are untouched and still name that file — but the file is gone, so the stage now points at something that no longer exists.

Worth noting this is how the legacy dialog behaved too, so it is not a regression. It is raised here because it sits awkwardly next to the next finding: the two together decide what the file lifecycle should be.

_Left open for the developer's decision._

src/BloomBrowserUI/bookEdit/toolbox/readers/readerSetup/ReaderDialogPhaseSection.tsx
```diff
@@ -0,0 +1,390 @@
...
+            new ReaderStage((updatedSettings.stages.length + 1).toString()),
+        );
+        props.setSettings(updatedSettings);
+        props.setSelectedStageIndex(updatedSettings.stages.length - 1);
+        setStageIds([...stageIds, `decodable-stage-${nextStageId.current++}`]);
+    };
+
+    const removeSelectedStage = () => {
+        const removedFile =
+            props.settings.stages[props.selectedStageIndex]?.allowedWordsFile;
```
andrew-polk, line 188: **Devin (Bug): removing a stage no longer deletes its word list file.**

The legacy dialog's remove-stage path called `checkAndDeleteAllowedWordsFile`, which deleted the word list file once no remaining stage referenced it. The React dialog's `removeSelectedStage` just drops the stage, so unused word list files now accumulate in the collection's Word Lists folder.

This is the other half of the finding above, and the two pull in opposite directions: the trash icon deletes immediately (so Cancel can strand a reference), while removing a whole stage now deletes nothing. Making stage-removal delete immediately would widen the Cancel problem; deferring both to Save would change behaviour the legacy dialog has always had.

_Left open for the developer's decision._

src/BloomBrowserUI/bookEdit/toolbox/readers/readerSetup/DecodableStagesSetup.tsx
```diff
@@ -0,0 +1,1168 @@
...
+                }
+            });
+        };
+
+        // This subscription synchronizes React state with Bloom's non-React Sample Texts folder watcher.
+        refreshSampleTextFiles();
+        toolbox.addSampleTextFilesChangedListener(
+            listenerName,
+            refreshSampleTextFiles,
+        );
```
andrew-polk: **Devin (Bug): the dialog keeps listening for window focus after it is closed.**

The Sample Words tab registers a `focus` listener on the workspace window when it mounts, and removes it in the effect's cleanup. That cleanup never runs, because the dialog's React tree is never unmounted — `ShowEditViewDialog` detaches the old `modal-dialog-react-root` element and creates a new one, and detaching a container does not unmount the React root inside it.

What that means in use: every time the user opens this dialog and visits the Sample Words tab, one more listener is left behind. Each later click back into Bloom then re-queries the Sample Texts folder once per abandoned dialog session, and the old dialog (with its copy of the settings) can never be garbage collected.

Note this is not specific to this dialog — the same non-unmounting applies to everything shown via `ShowEditViewDialog`. This one just happens to add a window-level listener, which is what makes it visible. That is why it is being raised rather than quietly fixed here.

_Left open for the developer's decision._

src/BloomBrowserUI/bookEdit/toolbox/readers/readerSetup/DecodableReaderSetupDialog.tsx
```diff
@@ -0,0 +1,247 @@
...
+        // Closing twice must not post the unlock twice. Closing does not unmount this
+        // component, so the handle below stays live afterwards; without this guard a later
+        // closeSetupDialog() would re-unlock the edit view even though something else may by
+        // then be relying on it staying locked.
+        if (hasClosed.current) {
+            return;
+        }
+        hasClosed.current = true;
+        setOpen(false);
+        postBoolean("editView/setModalState", false);
```
andrew-polk, line 80: **Devin (Bug): closed dialogs keep their subscriptions alive.**

Closing sets `open` to false, which hides the dialog, but the component stays mounted (see the companion finding on the focus listener for why). Its sample-file and word-list subscriptions therefore outlive every completed dialog session.

The toolbox-side listeners are keyed by fixed names, so those are overwritten by the newest dialog rather than accumulating; it is the window-level `focus` listener that multiplies. Same root cause as the companion finding, recorded separately so each can be resolved on its own.

_Left open for the developer's decision._

src/BloomBrowserUI/bookEdit/toolbox/readers/readerTools.ts
andrew-polk, line 1: **Devin (Bug): saving from the legacy dialog now always reloads every sample text file.**

Devin's reasoning: `beginSaveChangedSettings` gained a fourth parameter, `previousUseAllowedWords`; the legacy caller in `readerSetup.io.ts` still passes only three arguments, so the new comparison is never equal and a full rebuild of all sample-word data runs on every legacy save.

DistFiles/localization/en/Bloom.xlf
```diff
@@ -5345,6 +5345,16 @@ Do you want to go ahead?</note>
...
       </trans-unit>
+      <trans-unit id="ReaderSetup.NoSampleTextsYet" translate="no">
+        <source xml:lang="en">No sample texts yet. Add text files to the Sample Texts folder.</source>
+        <note>ID: ReaderSetup.NoSampleTextsYet</note>
+        <note>Shown in place of the file list on the Sample Words tab of the Decodable Reader setup dialog, when the collection's Sample Texts folder is empty. "Sample Texts folder" is the name of a folder in the collection and is also the label of the link just above this message.</note>
+      </trans-unit>
+      <trans-unit id="ReaderSetup.SeparateWordsWithSpaces" translate="no">
+        <source xml:lang="en">Separate words with spaces.</source>
+        <note>ID: ReaderSetup.SeparateWordsWithSpaces</note>
+        <note>Hint shown directly under the "New Sight Words" text box on the Decodable Stages tab of the Decodable Reader setup dialog. Tells the user how to delimit the sight words they type into that box.</note>
```
andrew-polk, line 5356: **Devin (Investigate): the two new strings are marked `translate="no"`.**

Devin's concern: both new entries in `Bloom.xlf` carry `translate="no"`, and it asks for confirmation that they enter the Crowdin workflow so translated interfaces do not keep showing English.

src/BloomBrowserUI/bookEdit/toolbox/readers/readerTools.ts
andrew-polk: [Claude Opus 5 from Andrew Polk's machine during preflight]

**Not an issue.** `beginSaveChangedSettings` guards exactly this case:

```ts
const refreshAllBasedOnAllowedWords: number | boolean =
    previousUseAllowedWords !== undefined
        ? settings.useAllowedWords !== previousUseAllowedWords
        : settings.useAllowedWords;
```

When the legacy caller omits the fourth argument, `previousUseAllowedWords` is `undefined` and the expression falls back to `settings.useAllowedWords` — which is precisely the pre-PR condition. So the legacy dialog keeps the behaviour it always had, and the claim that the comparison is "never equal" and therefore always true does not hold: the comparison is not reached at all on that path.

Resolving.

DistFiles/localization/en/Bloom.xlf
andrew-polk: [Claude Opus 5 from Andrew Polk's machine during preflight]

**Not an issue — this is the documented house rule.** `.github/skills/xlf-strings/SKILL.md` says: "Always mark new entries `translate="no"` unless instructed otherwise." New English strings are added that way deliberately and are picked up by the Crowdin workflow separately; the attribute is how this repo marks an entry as new, and it is also what lets its ID still be changed safely. Both new entries follow that rule, as do the new entries master added alongside them.

Resolving.

andrew-polk:
[Claude Opus 5 from Andrew Polk's machine during preflight]

**Consulted Devin on 2026-09-17 up to commit `54fe49ca8a`.**

Devin's review of this commit listed 22 bugs and 26 investigate flags. Its own `incremental` field shows why that number is misleading: the round before this one re-reviewed the whole diff from scratch (`change_too_large`), so most of the list is findings this PR fixed or decided in earlier rounds. Each was screened against the code at this commit, and 16 of the 22 are already fixed here — the dead Set Up Stages button, the edit-view lock, the line-break-mangled sample words, the special-symbol lookup, the deleted-letter analysis, the mislabelled file extensions, the tab-specific help page, the localized sight-words hint, and the rest. They are not re-posted.

What is left, each with its own thread:

- **Fixed this round** — the Sample Words help link was a plain anchor to `/bloom/api/help`, which in this dialog (rendered in the workspace document, not the old throw-away iframe) navigated the whole edit view away from the book. It now asks Bloom to open help.
- **Also fixed this round**, found by the local review rather than Devin — a replaced ReaderToolsModel got its settings back but no Sample Texts watcher, so the Sample Words tab threw on mount. Both load paths now go through `ensureSampleTextsDirectoryWatcher`, with a regression test.
- **Two not-an-issue**, replied to and resolved: the legacy save path is explicitly guarded for the missing fourth argument, and `translate="no"` on new strings is this repo's documented rule.
- **Four left open for the developer**: the allowed-words file lifecycle (two threads that pull against each other), and the dialog never being unmounted (two threads, one root cause, affecting every dialog shown via `ShowEditViewDialog`).

CI is green, and this repo's other configured review bot is disabled in `.coderabbit.yml`. Local gates at this commit: front-end 902 passing, C# 3521 passing, e2e journey test 3/3.

andrew-polk (commented):
> @andrew-polk reviewed 1 file and resolved 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8132)* status: 1 of 28 files reviewed, 4 unresolved discussions (waiting on josiahwall).

src/BloomBrowserUI/bookEdit/toolbox/readers/readerSetup/DecodableStagesSetup.tsx
andrew-polk: [Claude Opus 5 from Andrew Polk's machine during preflight]

**Decided by the developer: leave the behaviour as it is, and write down how to improve it.**

Deleting the file the moment the user clicks the trash icon is what the legacy dialog did, so this is not a regression, and the developer chose to keep it rather than take on the extra complexity now. The cost is real and is the part worth remembering: pressing Cancel afterwards leaves the saved settings naming a file that is already gone.

A comment at `removeAllowedWordsFile` now records that, and spells out what improving it would take — collect the files the user removed rather than deleting them, then act on that list in the save path, so Cancel leaves both the settings and the files untouched. `removeSelectedStage` points at the same note, since it deletes on the same terms.

Resolving.

src/BloomBrowserUI/bookEdit/toolbox/readers/readerSetup/ReaderDialogPhaseSection.tsx
andrew-polk: [Claude Opus 5 from Andrew Polk's machine during preflight]

**Not an issue — the code already does this.**

`removeSelectedStage` does delete the word list file when no remaining stage references it, using the same check the trash icon uses:

```ts
// Delete the word-list file if no remaining stage references it.
if (removedFile) {
    const stillUsed = updatedSettings.stages.some(
        (oneStage) => oneStage.allowedWordsFile === removedFile,
    );
    if (!stillUsed) { axios.delete(...) }
}
```

That has been there since the original conversion commit (`77cf52eae4`), so the legacy cleanup was never lost and word list files do not accumulate.

Resolving.

src/BloomBrowserUI/bookEdit/toolbox/readers/readerSetup/DecodableStagesSetup.tsx
andrew-polk: [Claude Opus 5 from Andrew Polk's machine during preflight]

**Not an issue — measured, and the listener is removed.**

The premise is that the Sample Words tab's cleanup never runs because the dialog's React tree is never unmounted. Those are two different things. The React *root container* is indeed abandoned on each showing, but the dialog's *children* unmount normally: closing sets `open` to false, and `BloomDialog` does not pass `keepMounted`, so MUI unmounts them and their effect cleanups run.

Measured in a running Bloom rather than argued: instrumenting `window.addEventListener`/`removeEventListener`, then opening this dialog, visiting the Sample Words tab and closing it, three times over, gives **adds=3, removes=3, net=0**. No listener accumulates and nothing re-queries the Sample Texts folder afterwards.

The abandoned root is real, though — an empty shell per showing, never freed for the life of the edit view. Because that belongs to the shared host rather than to this dialog, it is now documented in `ShowEditViewDialog` (`workspaceRoot.ts`), including this measurement, so the next reader does not have to re-derive which half is true.

Resolving.

src/BloomBrowserUI/bookEdit/toolbox/readers/readerSetup/DecodableReaderSetupDialog.tsx
andrew-polk: [Claude Opus 5 from Andrew Polk's machine during preflight]

**Not an issue, for the same reason as the companion thread.**

The dialog's subscriptions do not outlive a session: closing sets `open` to false, MUI unmounts the children because `BloomDialog` passes no `keepMounted`, and the effect cleanups run. The window `focus` listener was measured at a net zero across three open/close cycles.

What does survive each showing is the abandoned React root itself — an empty shell, not its subscriptions. That limitation is now recorded in `ShowEditViewDialog`, where it belongs, since it applies to every edit-view dialog rather than to this one.

Resolving.

andrew-polk (commented):
> @andrew-polk resolved 3 discussions.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8132)* status: 1 of 28 files reviewed, 1 unresolved discussion (waiting on josiahwall).

andrew-polk (commented):
> @andrew-polk reviewed 28 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8132)* status: all files reviewed, 1 unresolved discussion (waiting on josiahwall).

andrew-polk (commented):
> @andrew-polk reviewed 10 files and all commit messages, and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8132)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on josiahwall).

JohnThomson (changes requested):
> @JohnThomson reviewed all commit messages and made 13 comments.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8132)* status: all files reviewed, 13 unresolved discussions (waiting on josiahwall).

___
*[`src/BloomBrowserUI/bookEdit/toolbox/readers/ReaderSetupButton.tsx` line 26 at r16](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8132#-P28vhoI-JYGoa8fKPLB:-P28vhoI-JYGoa8fKPLC:b2xk797) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/2ecae8ac788df6c56690794fb562b7e14d802cd2/src/BloomBrowserUI/bookEdit/toolbox/readers/ReaderSetupButton.tsx#L26)):*
> ```XML
>     } else {
>         legacyDialogHref =
>             "javascript:window.toolboxBundle.showSetupDialog('stages');";
> ```

This else is stale (unreachable) code, right? Since onClick will now always be defined for decodable stages.
___
*[`src/BloomBrowserUI/bookEdit/toolbox/readers/readerTools.ts` line 309 at r16](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8132#-P28waclDaXyNa3U19n-:-P28wacm40q3tGPlPn9d:b-ixqh83) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/2ecae8ac788df6c56690794fb562b7e14d802cd2/src/BloomBrowserUI/bookEdit/toolbox/readers/readerTools.ts#L309)):*
> ```XML
>  * folder.
>  *
>  * Both paths that load settings into a model call this, because only one of them used to. When
> ```

The reason for calling on both paths is hardly that "only one of them used to". More like, "because we need to watch for settings changes whichever reader tool the book uses".
But I'm surprised that we do. I thought sample texts were only applicable to decodable. Maybe we instead want to check in relevant places whether we have a watcher or not?
___
*[`src/BloomBrowserUI/bookEdit/toolbox/readers/readerTools.ts` line 447 at r16](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8132#-P295Dj14aswruH0QpdW:-P295Dj2C1HV-kJ2_8Iw:b-r2ep8l) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/2ecae8ac788df6c56690794fb562b7e14d802cd2/src/BloomBrowserUI/bookEdit/toolbox/readers/readerTools.ts#L447)):*
> ```XML
>     previousUseAllowedWords?: number,
> ): Promise<void> {
>     // Using axios directly because our api at this point calls for returning the promise.
> ```

comment belongs to the axios call below
___
*[`src/BloomBrowserUI/bookEdit/toolbox/readers/readerTools.ts` line 453 at r16](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8132#-P295UBr3zrtk7HhDE4M:-P295UBr3zrtk7HhDE4N:b-ph2glj) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/2ecae8ac788df6c56690794fb562b7e14d802cd2/src/BloomBrowserUI/bookEdit/toolbox/readers/readerTools.ts#L453)):*
> ```XML
>             ? settings.useAllowedWords !== previousUseAllowedWords
>             : settings.useAllowedWords;
> ```

This is hard to follow, especially in a review where I can't quickly jump to the other code involved. For one thing, why is previousUseAllowedWords a number? Apparently because ReaderSettings.useAllowedWords is, though its only values are 0 and 1 and all users interpret these as false/true. Unfortunately I think it's written to a file that way, so fixing it is nontrivial, but is this a good time to improve the variable even if changing the file is too hard?
So it feels like this refreshAllBasedOnAllowedWords really just means "useAllowedWords changed", which is how it is used, in parallel with the other settings and previous values, to call beginRefreshEverything if anything changed.
The exception is that we apparently might not know a previous value for useAllowedWords (why not just make sure we do?? I can only find one caller, which always passes it), and in that case we consider it it to have changed if it is now true, but not otherwise. I think that deserves at least a comment. If we don't know the previous value, why assume it could not have been previously true and we need to refresh everything because it's now false?
(Of course as the old comment said, changing back and forth on this is something only testers are likely to do.)
___
*[`src/BloomBrowserUI/bookEdit/toolbox/readers/readerTools.ts` line 485 at r16](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8132#-P299Idv8XwYkUbAOJPu:-P299Idv8XwYkUbAOJPv:b-kj8gsy) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/2ecae8ac788df6c56690794fb562b7e14d802cd2/src/BloomBrowserUI/bookEdit/toolbox/readers/readerTools.ts#L485)):*
> ```XML
> }
> 
> /** Removes a listener previously added for word-list changes. */
> ```

That's to say, it removes the record of it from a special member of the model object. We're not doing a removeEventListener() here, which the method name suggests. Not sure what should change here, but it smells.
___
*[`src/BloomBrowserUI/bookEdit/toolbox/readers/readerTools.ts` line 514 at r16](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8132#-P29ARMN40GaQZE7Rn_z:-P29ARMN40GaQZE7Rna-:bpa7hhi) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/2ecae8ac788df6c56690794fb562b7e14d802cd2/src/BloomBrowserUI/bookEdit/toolbox/readers/readerTools.ts#L514)):*
> ```XML
>     );
>     return symbols.filter(
>         (symbol): symbol is string =>
> ```

What's going on here? "symbol is string" seems to be in the place I'd expect a return type if we were not leaving it implicit, but the return type of the function should be simply boolean.
(Also: since we appear to be filtering anything that isn't a string, could we improve type safety by declaring that this method returns an array of string rather than one of unknowns?)
___
*[`src/BloomBrowserUI/bookEdit/toolbox/readers/readerTools.ts` line 530 at r16](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8132#-P29CbBu5JsQusrsuthW:-P29CbBu5JsQusrsuthX:b-4fy6qk) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/2ecae8ac788df6c56690794fb562b7e14d802cd2/src/BloomBrowserUI/bookEdit/toolbox/readers/readerTools.ts#L530)):*
> ```XML
>     return paths.map((path) => ({
>         path,
>         readable: isReadableSampleTextFile(path),
> ```

See comment on the function. Maybe validType? We haven't verified that we can actually read it.
___
*[`src/BloomBrowserUI/bookEdit/toolbox/readers/readerTools.ts` line 531 at r16](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8132#-P29CPZz3CJ5FAs7A7nx:-P29CP_-1iGxoPXDtsL2:bghjb6f) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/2ecae8ac788df6c56690794fb562b7e14d802cd2/src/BloomBrowserUI/bookEdit/toolbox/readers/readerTools.ts#L531)):*
> ```XML
>         path,
>         readable: isReadableSampleTextFile(path),
>         hasExtension: getFileExtension(path) !== undefined,
> ```

hasExtension is redundant: any file without one will be readable:false.
___
*[`src/BloomBrowserUI/bookEdit/toolbox/readers/readerToolsModel.ts` line 83 at r16](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8132#-P29BfX3CEAz7-dBXF68:-P29BfX3CEAz7-dBXF69:b-kybmkc) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/2ecae8ac788df6c56690794fb562b7e14d802cd2/src/BloomBrowserUI/bookEdit/toolbox/readers/readerToolsModel.ts#L83)):*
> ```XML
>  * named .TXT counts just as .txt does.
>  */
> export function isReadableSampleTextFile(path: string): boolean {
> ```

How about something like "isValidSampleTextFileType"? The current name suggests that we have verified that it exists and we have permission to access it and maybe even that it parses correctly.
___
*[`src/BloomBrowserUI/bookEdit/toolbox/readers/readerSetup/DecodableReaderSetupDialog.tsx` line 81 at r16](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8132#-P29GDwb0syBhWtlWQez:-P29GDwb0syBhWtlWQf-:b-8rjjfz) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/2ecae8ac788df6c56690794fb562b7e14d802cd2/src/BloomBrowserUI/bookEdit/toolbox/readers/readerSetup/DecodableReaderSetupDialog.tsx#L81)):*
> ```XML
>         setOpen(false);
>         postBoolean("editView/setModalState", false);
>     };
> ```

Does this want to be a useMemo or similar? We don't need a new copy of the function on every render...
___
*[`src/BloomBrowserUI/bookEdit/toolbox/readers/readerSetup/DecodableReaderSetupDialog.tsx` line 195 at r16](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8132#-P29HSEvEH7JOgZyZO79:-P29HSEvEH7JOgZyZO7A:btq5i3x) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/2ecae8ac788df6c56690794fb562b7e14d802cd2/src/BloomBrowserUI/bookEdit/toolbox/readers/readerSetup/DecodableReaderSetupDialog.tsx#L195)):*
> ```XML
>                             <DialogHelpButton helpId={helpId} />
>                         </DialogBottomLeftButtons>
>                     )}
> ```

No relevant help if we can't edit?
___
*[`src/BloomBrowserUI/bookEdit/toolbox/readers/readerSetup/decodableStagesUtils.ts` line 103 at r16](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8132#-P29K2kk7C2kpXnN9iWF:-P29K2kl0r092ZVkDwOV:b-e5ju5u) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/2ecae8ac788df6c56690794fb562b7e14d802cd2/src/BloomBrowserUI/bookEdit/toolbox/readers/readerSetup/decodableStagesUtils.ts#L103)):*
> ```XML
>         sortedGpcs,
>     );
> };
> ```

This function (and maybe others in this file) feel like things LibSynphony already knows. Maybe this is a legitimate step towards moving that knowledge from js to ts, but it seemed worth at least asking if we could reuse the existing logic.
___
*[`src/BloomBrowserUI/bookEdit/toolbox/readers/readerSetup/readerSetup.ui.ts` line 44 at r16](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8132#-P29OEMJ37wV9TA9P2eI:-P29OEMJ37wV9TA9P2eJ:bz0tt9q) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/2ecae8ac788df6c56690794fb562b7e14d802cd2/src/BloomBrowserUI/bookEdit/toolbox/readers/readerSetup/readerSetup.ui.ts#L44)):*
> ```XML
>                 let foundNotSupported: boolean = false;
>                 files.forEach((element, index, array) => {
>                     const filenameComponents: string[] = element.split(".");
> ```

Somewhere in this PR we have other code that knows about getting an extension from a path (better than this...handles dot in folder name) and which extensions are valid.

andrew-polk (commented):
> @andrew-polk+AI made 13 comments.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8132)* status: 29 of 36 files reviewed, 13 unresolved discussions (waiting on andrew-polk, JohnThomson, and josiahwall).

___
*[`src/BloomBrowserUI/bookEdit/toolbox/readers/ReaderSetupButton.tsx` line 26 at r16](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8132#-P28vhoI-JYGoa8fKPLB:-P2A96p04dtumZ2Gim83:b-wbvi0j) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/2ecae8ac788df6c56690794fb562b7e14d802cd2/src/BloomBrowserUI/bookEdit/toolbox/readers/ReaderSetupButton.tsx#L26)):*

This else is stale (unreachable) code, right? Since onClick will now always be defined for decodable stages.
</blockquote></details>

[Claude Opus 5.5 from Andrew Polk's machine during reviewable-replies] Yes, removed. In fact nothing calls `showSetupDialog('stages')` any more, so the legacy dialog's whole stages path is dead. We're deliberately removing only this `else` for now. The Leveled Reader conversion is coming soon and will retire much more of that legacy code at once; this PR has been vetted and tested, and like the servants in the parable we'd rather not pull up the wheat with the tares. We'll sort them at the harvest.
___
*[`src/BloomBrowserUI/bookEdit/toolbox/readers/readerTools.ts` line 309 at r16](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8132#-P28waclDaXyNa3U19n-:-P2A97DYDU3mLmx7zJD-:bu00a5h) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/2ecae8ac788df6c56690794fb562b7e14d802cd2/src/BloomBrowserUI/bookEdit/toolbox/readers/readerTools.ts#L309)):*

The reason for calling on both paths is hardly that "only one of them used to". More like, "because we need to watch for settings changes whichever reader tool the book uses".
But I'm surprised that we do. I thought sample texts were only applicable to decodable. Maybe we instead want to check in relevant places whether we have a watcher or not?
</blockquote></details>

[Claude Opus 5.5 from Andrew Polk's machine during reviewable-replies] Agreed, reworded to give the real reason: whichever path loaded the settings, the model needs the watcher afterwards. Creating it for both reader tools isn't new, though. Master creates it in `initializeSynphony`, which runs for both; this PR only added the reload path. While there I also fixed something you didn't flag: the new function had separated `initializeSynphony` from its doc comment.
___
*[`src/BloomBrowserUI/bookEdit/toolbox/readers/readerTools.ts` line 447 at r16](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8132#-P295Dj14aswruH0QpdW:-P2A97es5MYqqsK9aMct:bvk1ex3) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/2ecae8ac788df6c56690794fb562b7e14d802cd2/src/BloomBrowserUI/bookEdit/toolbox/readers/readerTools.ts#L447)):*

comment belongs to the axios call below
</blockquote></details>

[Claude Opus 5.5 from Andrew Polk's machine during reviewable-replies] Moved back onto the axios call.
___
*[`src/BloomBrowserUI/bookEdit/toolbox/readers/readerTools.ts` line 453 at r16](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8132#-P295UBr3zrtk7HhDE4M:-P2A9857CZEjxR4tw7dt:b-c0kss3) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/2ecae8ac788df6c56690794fb562b7e14d802cd2/src/BloomBrowserUI/bookEdit/toolbox/readers/readerTools.ts#L453)):*

This is hard to follow, especially in a review where I can't quickly jump to the other code involved. For one thing, why is previousUseAllowedWords a number? Apparently because ReaderSettings.useAllowedWords is, though its only values are 0 and 1 and all users interpret these as false/true. Unfortunately I think it's written to a file that way, so fixing it is nontrivial, but is this a good time to improve the variable even if changing the file is too hard?
So it feels like this refreshAllBasedOnAllowedWords really just means "useAllowedWords changed", which is how it is used, in parallel with the other settings and previous values, to call beginRefreshEverything if anything changed.
The exception is that we apparently might not know a previous value for useAllowedWords (why not just make sure we do?? I can only find one caller, which always passes it), and in that case we consider it it to have changed if it is now true, but not otherwise. I think that deserves at least a comment. If we don't know the previous value, why assume it could not have been previously true and we need to refresh everything because it's now false?
(Of course as the old comment said, changing back and forth on this is something only testers are likely to do.)
</blockquote></details>

[Claude Opus 5.5 from Andrew Polk's machine during reviewable-replies] It's now a boolean, `useAllowedWordsChanged`. `ReaderSettings.useAllowedWords` stays a number, because that's how the settings file stores it. There are actually two callers. The legacy jQuery dialog (still used for leveled readers) reaches this through `toolboxBundle`, across frames, which is easy to miss in a search, and it doesn't pass the previous value. That's why the parameter is optional; the fallback keeps master's old rule for that dialog, and the code now says so. It goes away when the leveled dialog is converted.
___
*[`src/BloomBrowserUI/bookEdit/toolbox/readers/readerTools.ts` line 485 at r16](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8132#-P299Idv8XwYkUbAOJPu:-P2A98UQ5_nLvTJIDTf7:b-qjit6u) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/2ecae8ac788df6c56690794fb562b7e14d802cd2/src/BloomBrowserUI/bookEdit/toolbox/readers/readerTools.ts#L485)):*

That's to say, it removes the record of it from a special member of the model object. We're not doing a removeEventListener() here, which the method name suggests. Not sure what should change here, but it smells.
</blockquote></details>

[Claude Opus 5.5 from Andrew Polk's machine during reviewable-replies] Fair. It mirrors the existing `addWordListChangedListener`, which also keeps the model's own table of callbacks rather than using DOM listeners. I left the pair as it is and reworded the doc to say that.
___
*[`src/BloomBrowserUI/bookEdit/toolbox/readers/readerTools.ts` line 514 at r16](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8132#-P29ARMN40GaQZE7Rn_z:-P2A98r7-Hcnk1KXZ6d3:b-7kj42w) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/2ecae8ac788df6c56690794fb562b7e14d802cd2/src/BloomBrowserUI/bookEdit/toolbox/readers/readerTools.ts#L514)):*

What's going on here? "symbol is string" seems to be in the place I'd expect a return type if we were not leaving it implicit, but the return type of the function should be simply boolean.
(Also: since we appear to be filtering anything that isn't a string, could we improve type safety by declaring that this method returns an array of string rather than one of unknowns?)
</blockquote></details>

[Claude Opus 5.5 from Andrew Polk's machine during reviewable-replies] That's a TypeScript type predicate: it's what lets `.filter` narrow `unknown[]` to `string[]`. The function already returns `string[]`; the working array is `unknown[]` only because Synphony's language data is untyped. I added a one-line comment.
___
*[`src/BloomBrowserUI/bookEdit/toolbox/readers/readerTools.ts` line 530 at r16](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8132#-P29CbBu5JsQusrsuthW:-P2A99Ew-7kXIVDZ7sgF:b-ejwkav) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/2ecae8ac788df6c56690794fb562b7e14d802cd2/src/BloomBrowserUI/bookEdit/toolbox/readers/readerTools.ts#L530)):*

See comment on the function. Maybe validType? We haven't verified that we can actually read it.
</blockquote></details>

[Claude Opus 5.5 from Andrew Polk's machine during reviewable-replies] Renamed to `validType`.
___
*[`src/BloomBrowserUI/bookEdit/toolbox/readers/readerTools.ts` line 531 at r16](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8132#-P29CPZz3CJ5FAs7A7nx:-P2A99cH3QwQQ2LFfou-:bvj2s5k) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/2ecae8ac788df6c56690794fb562b7e14d802cd2/src/BloomBrowserUI/bookEdit/toolbox/readers/readerTools.ts#L531)):*

hasExtension is redundant: any file without one will be readable:false.
</blockquote></details>

[Claude Opus 5.5 from Andrew Polk's machine during reviewable-replies] Not quite. It picks between the two messages for a file we won't read: "Cannot read this format" (e.g. .docx) versus "File needs .TXT extension" (no extension at all). `validType` alone can't tell those apart.
___
*[`src/BloomBrowserUI/bookEdit/toolbox/readers/readerToolsModel.ts` line 83 at r16](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8132#-P29BfX3CEAz7-dBXF68:-P2A9A-qEZr9ZiSwXWj6:bqauh8t) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/2ecae8ac788df6c56690794fb562b7e14d802cd2/src/BloomBrowserUI/bookEdit/toolbox/readers/readerToolsModel.ts#L83)):*

How about something like "isValidSampleTextFileType"? The current name suggests that we have verified that it exists and we have permission to access it and maybe even that it parses correctly.
</blockquote></details>

[Claude Opus 5.5 from Andrew Polk's machine during reviewable-replies] Renamed to `isValidSampleTextFileType`, and its doc now says it checks only the extension.
___
*[`src/BloomBrowserUI/bookEdit/toolbox/readers/readerSetup/DecodableReaderSetupDialog.tsx` line 81 at r16](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8132#-P29GDwb0syBhWtlWQez:-P2A9AO7BiEAyaJj4zk7:b-z7t85h) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/2ecae8ac788df6c56690794fb562b7e14d802cd2/src/BloomBrowserUI/bookEdit/toolbox/readers/readerSetup/DecodableReaderSetupDialog.tsx#L81)):*

Does this want to be a useMemo or similar? We don't need a new copy of the function on every render...
</blockquote></details>

[Claude Opus 5.5 from Andrew Polk's machine during reviewable-replies] Done, with `useCallback`. The allocation doesn't matter much, but a stable `close` lets the effect that publishes the close handle run once instead of after every render.
___
*[`src/BloomBrowserUI/bookEdit/toolbox/readers/readerSetup/DecodableReaderSetupDialog.tsx` line 195 at r16](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8132#-P29HSEvEH7JOgZyZO79:-P2A9An707Hez7WPQzed:b2vfpjy) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/2ecae8ac788df6c56690794fb562b7e14d802cd2/src/BloomBrowserUI/bookEdit/toolbox/readers/readerSetup/DecodableReaderSetupDialog.tsx#L195)):*

No relevant help if we can't edit?
</blockquote></details>

[Claude Opus 5.5 from Andrew Polk's machine during reviewable-replies] The legacy dialog hid it too (it added Help and OK in the same "not forbidden" branch), so this was parity. But agreed, it's still useful, so it now always shows.
___
*[`src/BloomBrowserUI/bookEdit/toolbox/readers/readerSetup/decodableStagesUtils.ts` line 103 at r16](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8132#-P29K2kk7C2kpXnN9iWF:-P2A9BB1-b29zCT0vPG0:b82zx36) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/2ecae8ac788df6c56690794fb562b7e14d802cd2/src/BloomBrowserUI/bookEdit/toolbox/readers/readerSetup/decodableStagesUtils.ts#L103)):*

This function (and maybe others in this file) feel like things LibSynphony already knows. Maybe this is a legitimate step towards moving that knowledge from js to ts, but it seemed worth at least asking if we could reuse the existing logic.
</blockquote></details>

[Claude Opus 5.5 from Andrew Polk's machine during reviewable-replies] It does: `getGpcForm` delegates to `theOneLanguageDataInstance.getGpcForm`, and its comment explains why. The only local logic is sorting the graphemes longest-first, which Synphony expects the caller to do.
___
*[`src/BloomBrowserUI/bookEdit/toolbox/readers/readerSetup/readerSetup.ui.ts` line 44 at r16](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8132#-P29OEMJ37wV9TA9P2eI:-P2A9BbxDaFZZXocma7q:bymbjen) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/2ecae8ac788df6c56690794fb562b7e14d802cd2/src/BloomBrowserUI/bookEdit/toolbox/readers/readerSetup/readerSetup.ui.ts#L44)):*

Somewhere in this PR we have other code that knows about getting an extension from a path (better than this...handles dot in folder name) and which extensions are valid.
</blockquote></details>

[Claude Opus 5.5 from Andrew Polk's machine during reviewable-replies] Agreed, that's the naive version we replaced in the new code with `getFileExtension`. It's the legacy dialog's sample-texts code, and with stages now going to the React dialog it's almost certainly unreachable. We'll remove it with the rest of the legacy code when the leveled dialog is converted, rather than patch it now (same wheat-and-tares reasoning as the ReaderSetupButton thread).

src/BloomBrowserUI/bookEdit/toolbox/readers/readerSetup/DecodableReaderSetupDialog.tsx
```diff
@@ -0,0 +1,246 @@
...
+    }, []);
+
+    // Publish the close handle from an effect, not from the render body: assigning module-level
+    // state while rendering runs again on every re-render (and twice under StrictMode). Cleared
+    // on unmount.
+    useMountEffect(() => {
```
andrew-polk, line 87: [Claude Opus 5.5 from Andrew Polk's machine during preflight]

**Devin (Investigate): an effect that only runs on mount should use the project's `useMountEffect` helper.**

Once `close` became stable, the effect that publishes the dialog's close handle was really a run-on-mount effect written as `useEffect(..., [close])`. The Browser UI guidance (`src/BloomBrowserUI/AGENTS.md`) asks for `useMountEffect` in that case, so the intent is stated plainly.

andrew-polk: [Claude Opus 5.5 from Andrew Polk's machine during preflight]

Fixed: it now uses `useMountEffect`. Behaviour is unchanged, since `close` never changes, so the handle was already published only once.

andrew-polk:
[Claude Opus 5.5 from Andrew Polk's machine during preflight]

**Consulted Devin up to `06c8ba902d`** (the branch rebased onto master, with John's review addressed).

The rebase made Devin re-review the whole PR from scratch, so it repeated the findings from earlier rounds. Each of those already has a resolved thread here recording what was done. One flag was new since the last round, [use `useMountEffect` for a run-on-mount effect](https://github.com/BloomBooks/BloomDesktop/pull/8132#discussion_r4077239004). It was already fixed at this commit, so I replied and resolved it. CI (`pr-automation`) passes.

src/BloomBrowserUI/bookEdit/toolbox/readers/readerSetup/DecodableReaderSetupDialog.tsx
andrew-polk: [Claude Opus 5.5 from Andrew Polk's machine during pr-ready-for-human] Fixed and resolved on GitHub; this just clears it here too.

andrew-polk (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8132)* status: 29 of 36 files reviewed, 14 unresolved discussions (waiting on andrew-polk, JohnThomson, and josiahwall).

andrew-polk (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8132)* status: 29 of 36 files reviewed, 13 unresolved discussions (waiting on JohnThomson and josiahwall).

JohnThomson (approved):
> @JohnThomson reviewed 7 files and all commit messages, and resolved 13 discussions.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8132)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on josiahwall).

PR #8133, opened 2026-07-30
Provide explicit Cancel for Publish>Apps and make it modal while running (BL-16350)

StephenMcConnel:
While a Reading App Builder prepare/build/try-on-phone action is running (its **Cancel** button showing), the Publish → Apps operation is now **modal**: the user's only option is the explicit Cancel button, and Bloom prevents navigating away rather than auto-cancelling. This implements the design agreed on the card (the earlier auto-cancel-on-navigate approach, PR #8119, was discarded).

### Modal navigation
- **Workspace tabs** (Collections / Edit / Publish) are disabled from C# for the duration of an action — `RabPublishApi` injects `PublishView` and calls `WorkspaceView.SetTabsEnabled(false/true)` around each action (same mechanism `LibraryPublishApi` uses during an upload).
- **Publish-tool switcher** (BloomPUB / ePUB / Apps / …) is blocked in React — `PublishTabPane` vetoes `onSelect` and disables the other tools while the Apps screen reports busy via a new `onBusyChange` callback.

### Responsive cancellation
- Check for cancellation **between building each BloomPUB** in the export loop, so a Cancel during a multi-book export stops after the current book instead of grinding through all of them.
- Thread a `CancellationToken` through the RAB **installer download** HTTP calls so a stalled download aborts promptly instead of waiting for the 30-minute `HttpClient` timeout.

### No debris left behind on interrupt
- **Download:** writes to a temp `.part` file and moves it into place only on success; a cancelled/failed download leaves no partial installer at the real name (which `FindRabSetupInstallerPath` would otherwise find and try to run).
- **Build:** a cancelled/failed build deletes intermediate `.apk` files under `BuildRoot` so a partial/unsigned APK can't be mistaken for a finished app by `FindLatestApkPath`; the previously built signed APK in `SafeApkRoot` is preserved.
- **Install:** the update-incompatible recovery (uninstall + reinstall) is made atomic via a scoped `_protectCurrentProcessFromCancellation` flag, so a cancel can't leave the phone with the old app removed and nothing reinstalled.

### Tests
New unit tests cover cancel-before-retry and cancel-during-replace in install, failed-build APK cleanup, and the atomic-download cancel/success cases.

Ref: https://issues.bloomlibrary.org/youtrack/issue/BL-16350

[Devin review](https://devinreview.com/BloomBooks/BloomDesktop/pull/8133)

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8133)


src/BloomExe/Publish/Rab/RabPublishApi.cs
StephenMcConnel, line 1: [Devin] **Bug**: A stalled Reading App Builder download that times out is reported as if the user cancelled

Devin flagged that the prepare/build/install handlers caught `OperationCanceledException` and reported it as a user cancellation. An `HttpClient` timeout during the RAB installer download throws `TaskCanceledException` (which derives from `OperationCanceledException`), so a download that stalled and timed out would have been shown to the user as "cancelled" and never logged as the failure it is. (`src/BloomExe/Publish/Rab/RabPublishApi.cs`)

src/BloomExe/Publish/Rab/RabProjectService.cs
StephenMcConnel, line 1: [Devin] **Bug**: Cancel could be silently ignored, leaving the user stuck watching a long app build finish

Devin flagged that `RunProcess`/`RunProcessCapturingOutput` only checked for cancellation *after* a subprocess exited, never before launching one, and registered the process only after `process.Start()`. So a Cancel clicked during the in-process work between two RAB runs let the next multi-minute Gradle build launch and run to completion before the cancel was noticed — and with navigation locked, the user was stuck watching. (`src/BloomExe/Publish/Rab/RabProjectService.cs`)

src/BloomExe/Publish/Rab/RabPublishApi.cs
StephenMcConnel, line 1: [Devin] **Bug**: Cancel does nothing while the Reading App Builder installer is running, and the user can't leave the screen

On first run, if RAB isn't installed, Prepare downloads and runs the installer. `InstallRabFromSetup` runs the setup via `StartShellProcess` + `WaitForExit` and does not register that process for cancellation, so `RequestCancellation` has nothing to kill and no cancel check runs until the next step. Because this change also disables the workspace tabs and the publish-tool switcher for the whole Prepare, a user who cancels during the install is left with an unresponsive Cancel and no way to navigate away until the installer finishes. (`src/BloomExe/Publish/Rab/RabProjectService.cs` — InstallRabFromSetup; lock at `RabPublishApi.cs`)

Sent to the developer as a decision (see the preflight report): what should Cancel do during a third-party installer run — kill it (a killed install is detected as not-installed and re-run), or leave the tabs unlocked during the install/download prerequisite phase?

src/BloomBrowserUI/publish/Apps/AppPublisherScreen.tsx
StephenMcConnel, line 1: [Devin] **Bug**: Busy state is pushed to the parent screen from a `useEffect`, which the repo's React guidance lists as a "don't"

The Apps screen notifies its host that an action is running from inside a `React.useEffect` watching `screenState.busyAction`. `src/BloomBrowserUI/AGENTS.md` → `.github/skills/react-useeffect/SKILL.md` lists "Notify parent of changes" as a DON'T for effects (DO: call in an event handler). (`src/BloomBrowserUI/publish/Apps/AppPublisherScreen.tsx`)

Sent to the developer as a decision (see the preflight report): `busyAction` here is derived from several asynchronous sources (the websocket action-complete event, status-poll recovery, tab-activation restore), not a single user event, so an effect that mirrors it up is arguably the right tool — but the fuller refactor (pass the callback into the hook, notify at each state change, keep an unmount-cleanup) is available if strict compliance is preferred.

src/BloomExe/Publish/Rab/RabProjectService.cs
StephenMcConnel, line 1: [Devin] **Investigate**: A cancelled build could leave a truncated APK at the deliverable name in SafeApkRoot

The interrupted-build cleanup deletes intermediate `.apk` files under `BuildRoot`, but deliberately does not touch `SafeApkRoot` (to preserve the previous good signed APK on a failed rebuild). RAB writes its finished APK to `SafeApkRoot` under a name derived from the project — the same name each build. If RAB/Gradle is killed while writing/signing into `SafeApkRoot`, a truncated file could be left at the real APK name, overwriting the previous good APK and then picked up by `FindLatestApkPath`; if the build signature didn't change, "Try on phone" would still be enabled for it. (`src/BloomExe/Publish/Rab/RabProjectService.cs`)

Whether this is reachable depends on whether RAB writes its output atomically (temp + move). Sent to the developer (see the preflight report) since it needs knowledge of RAB's output behavior.

src/BloomExe/Publish/Rab/RabPublishApi.cs
StephenMcConnel: [Devin] ✅ Fixed in `bf4a6e0c7`. The prepare/build/install catch clauses are now guarded with `when (_rabProjectService.IsCancellationRequested)`. Every user-initiated OperationCanceledException in this code is raised only while `_cancelRequested` is set, so an HttpClient timeout (with `_cancelRequested` false) now falls through to the failure handler, which logs and surfaces the real error. Resolving.

src/BloomExe/Publish/Rab/RabProjectService.cs
StephenMcConnel: [Devin] ✅ Fixed in `815a2c457`. Both runners now throw OperationCanceledException *before* starting a process when a cancel is already pending, and kill the process if a cancel landed in the window between `Start()` and registering it — closing both windows described. (Both checks are skipped only during the protected uninstall+reinstall recovery, which must complete.) Resolving.

StephenMcConnel:
[Claude Opus 4.8 (1M context)] Consulted Devin during preflight, latest up to commit fa3915bb2. Final re-review is clean: 0 unresolved bugs, 0 new bugs; the remaining investigate flags are lagging re-flags of items already fixed/decided below (their threads are resolved).

All findings raised across the run were folded into the fix loop and are documented as resolved (or decided) review threads:

- Cancel silently ignored before launching the next long subprocess — fixed (815a2c45).
- Download timeout reported as a user cancel — fixed (bf4a6e0c).
- Cancel dead + user locked out during the first-run installer — installer made cancellable (c4688ffab).
- Busy-state useEffect vs. the React guidance — kept, comment justifies it (c4688ffab).
- Late cancel discarding a finished app — cleanup keeps a complete APK, deletes only truncated (d983d83e5).
- Tab lock unlocked mid-upload — reference-counted the shared lock by reason (555196881).
- Tab-lock bookkeeping not thread-safe — guarded the disabler set with a lock (47ffbc498).
- BuildRoot cleanup vs FindLatestApkPath — search narrowed to the deliverable roots (108bf76e3).
- Late cancel left Build "needed" — a complete APK produced before an abnormal exit is now recorded (108bf76e3).
- Race-kill window reported as failure — now reported as a cancellation (108bf76e3).
- Half-installed state trusted — interrupted install forces a reinstall; first via a marker, then refined to an in-session flag so GetStatus never hides a real install (108bf76e3, 9e91fc7f1).
- Device detection not cancellable — adb runs through the cancellation-aware path (108bf76e3).
- Interrupted build could record a stale ApkRoot APK as current — snapshot/cleanup/lookup unified on one deliverable-root set (fa3915bb2).

Decided "leave as is": Cancel during an elevated first-run install is deferred until the installer finishes (narrow, recoverable). Verified separately: RAB always writes the deliverable to SafeApkRoot (apk.output is set unconditionally), so narrowing FindLatestApkPath is safe.

CI green (pr-automation, track/set-waiting); CodeRabbit auto_review disabled (not run); no other bots.

src/BloomExe/Publish/Rab/RabPublishApi.cs
StephenMcConnel: [Claude Opus 4.8 (1M context)] Fixed in c4688ffab. The developer chose to make the first-run installer step cancellable rather than leave it as a lockout.

`InstallRabFromSetup` now registers the silent installer as the current process (so pressing Cancel kills it — `Kill(true)` takes the whole process tree, covering the second-phase process a silent InnoSetup installer spawns) and checks for cancellation both before launching and after it exits, mirroring `RunProcess`. A killed install is recoverable: it's detected as not-installed and re-run on the next Prepare, and it's reported to the user as a cancel rather than an installer "failed" error. The protected update-recovery path stays exempt so a recovery install still runs to completion.

src/BloomBrowserUI/publish/Apps/AppPublisherScreen.tsx
StephenMcConnel: [Claude Opus 4.8 (1M context)] Reviewed and keeping the effect, with a rewritten comment justifying it (c4688ffab).

The repo's `react-useeffect` guidance lists "notify the parent from a `useEffect`" as a "don't" *when there is a single event handler that caused the change* — do it there instead. Here there isn't one. `busyAction` is set/cleared from several asynchronous sources inside `useAppBuilderPublisherScreen`: the `actionComplete` websocket event, the status-poll recovery that reconciles with the backend after a blank/reload, and the action-start call. No single handler observes every transition, so a render keyed on the resulting value is the only place that sees them all.

What the effect does is synchronize an external system — the publish-tab host that makes the operation modal (it blocks the other publish tools while C# blocks the main workspace tabs) — to that derived state, which is a legitimate use of an effect. The comment now spells this out, satisfying AGENTS.md's requirement to justify a warranted effect.

src/BloomExe/Publish/Rab/RabProjectService.cs
StephenMcConnel: [Claude Opus 4.8 (1M context)] Investigated, and fixed defensively in c4688ffab.

It IS reachable on Bloom's side: `GetStatus` and `Install` (Try on phone) decide whether a finished app exists by scanning the directory via `FindLatestApkPath`, not from saved state, so a truncated APK left at the deliverable name would be treated as a finished app (`Install` would then push a corrupt APK to the phone and adb would reject it). And Bloom can't confirm the write is atomic: Reading App Builder is the external program that writes the APK (`apk.output` is just a CLI argument to it — there is no gradle template in this repo), so closing this as "not reachable" on the assumption RAB writes atomically wasn't defensible.

So the fix doesn't depend on RAB's behavior: on the abnormal (cancel/fail) build path we now delete only the APK(s) *this attempt* created or rewrote — determined from a snapshot of SafeApkRoot write-times taken before the build — while preserving a previous good APK the attempt never touched (so a failed rebuild still falls back to the last good build). Added `BuildAsync_WhenKilledWhileWritingDeliverableApk_DeletesTruncatedApkButKeepsUntouchedApk` covering both halves.

src/BloomExe/Publish/Rab/RabProjectService.cs
```diff
@@ -734,32 +836,59 @@ private void Build()
...
+                // scan the directories via FindLatestApkPath) could mistake for a finished, signed
+                // app: (1) partial/unsigned intermediates Gradle wrote under BuildRoot, and (2) a
+                // deliverable in SafeApkRoot that RAB was writing when it was killed — Bloom cannot
+                // see whether RAB writes that final APK atomically, so a kill mid-write could leave a
+                // truncated file at the deliverable name (BL-16350). Both run only on the abnormal
+                // path, so a successful build is untouched; and SafeApkRoot cleanup deletes only APKs
+                // this attempt created or rewrote, so a previous good APK still survives a failed
+                // rebuild.
+                DeleteIntermediateBuildApks(paths);
+                DeleteApksWrittenDuringBuild(paths, preBuildApkWriteTimes);
```
StephenMcConnel, line 900: [Devin] **Bug**: Cancelling near the end of an app build could wipe out the previously working app

Reported by Devin against the first version of the SafeApkRoot cleanup on this branch. `RunProcess` throws `OperationCanceledException` after `WaitForExit()` whenever a cancel was requested — even when Reading App Builder had already exited cleanly (code 0) having written a complete, signed APK. The abnormal-path cleanup then deleted that finished APK (its write time had changed), and because a same-named rebuild overwrites the previous APK in place, a user who clicked Cancel just as the build finished could be left with no usable app at all and forced into a full rebuild before "Try on phone" worked again.

StephenMcConnel: [Claude Opus 4.8 (1M context)] Good catch — fixed in d983d83e5.

`DeleteApksWrittenDuringBuild` now deletes an APK this attempt touched only if it is *also* no longer a complete, readable APK. `IsCompleteApk` opens the file as a zip and reads its entry list: a write cut off mid-stream is missing its central directory and fails to open (deleted — the truncation this cleanup exists for), while a fully-written APK opens fine and is kept, even when the cancel arrived just after RAB finished. So a late cancel no longer discards a usable build (nor the previous one it replaced in place). Added `BuildAsync_WhenAbnormalAfterApkFinished_KeepsTheCompletedApk`; the truncated- and Gradle-failure cases still pass.

src/BloomExe/Publish/Rab/RabProjectService.cs
```diff
@@ -734,32 +836,59 @@ private void Build()
...
+                // APK this attempt left behind that a later status check or "Try on phone" (both
+                // scan the directories via FindLatestApkPath) could mistake for a finished, signed
+                // app: (1) partial/unsigned intermediates Gradle wrote under BuildRoot, and (2) a
+                // deliverable in SafeApkRoot that RAB was writing when it was killed — Bloom cannot
+                // see whether RAB writes that final APK atomically, so a kill mid-write could leave a
+                // truncated file at the deliverable name (BL-16350). Both run only on the abnormal
+                // path, so a successful build is untouched; and SafeApkRoot cleanup deletes only APKs
+                // this attempt created or rewrote, so a previous good APK still survives a failed
+                // rebuild.
+                DeleteIntermediateBuildApks(paths);
```
StephenMcConnel, line 899: [Devin] **Investigate**: Cancelled build could leave a truncated APK at the deliverable name in SafeApkRoot

RAB is told to write its finished APK to SafeApkRoot (`apk.output={SafeApkRoot}`), and the filename is derived from the project name, so a rebuild targets the same file as the previous good build. If the RAB/Gradle tree is killed while writing/signing into SafeApkRoot, a truncated file could be left at the real APK name, overwriting the previous good one and being picked up by `FindLatestApkPath` — and if the rebuild's inputs were unchanged, `BuildNeeded` stays false and `apkExists` stays true, so "Try on phone" would still be enabled for that truncated APK.

src/BloomExe/Publish/Rab/RabPublishApi.cs
```diff
@@ -145,6 +173,7 @@ public void RegisterWithApiHandler(BloomApiHandler apiHandler)
                             // until the client is notified, preventing a status-poll in the gap
                             // from incorrectly clearing the client's busyAction via recovery logic.
                             _rabProjectService.ClearAction();
+                            SetWorkspaceTabsEnabled(true);
```
StephenMcConnel, line 179: [Devin] **Bug**: Finishing an Apps action can unlock the main tabs while a web upload is still in progress

`WorkspaceView.SetTabsEnabled` is a single shared boolean, not a reference count. A web-library upload sets it false for its duration and true in its `finally`; this PR adds a second user of the same lock (each RAB prepare/build/install sets it false, then true in its `finally`). So the sequence "start an upload → switch to the Apps tool → run a RAB action → the action ends" runs `SetWorkspaceTabsEnabled(true)` in the RAB `finally` while the upload still needs the tabs locked, silently letting the user navigate away mid-upload. The reverse ordering also breaks the modality this PR adds (the upload's `finally` re-enables the tabs in the middle of a RAB action). A reference count, or a shared "who holds the lock" set, would let the two features compose.

src/BloomExe/Publish/Rab/RabProjectService.cs
```diff
@@ -2632,7 +2945,43 @@ internal virtual void InstallRabFromSetup(string installerPath)
...
                 );
 
-            process.WaitForExit();
+            // Register the installer so RequestCancellation can kill it (Kill(true) takes the whole
+            // process tree, covering the second-phase process a silent InnoSetup installer spawns).
+            // Skip registration when protected so a recovery install still runs to completion.
+            if (!_protectCurrentProcessFromCancellation)
+            {
+                lock (_currentProcessLock)
+                    _currentProcess = process;
```
StephenMcConnel, line 3031: [Devin] **Investigate**: Cancel during the first-run installer may still block until the install finishes, if the installer is elevated

Item 1's fix registers the installer process so `RequestCancellation` can `Kill(true)` it. But `GetDefaultRabInstallDir()` returns a path under Program Files, so the InnoSetup installer requests UAC elevation and does its real work in a second, elevated process. Unelevated Bloom generally cannot kill an elevated process, so `Kill` throws and is swallowed, `WaitForExit()` then blocks until the installer actually finishes, and only then is `OperationCanceledException` thrown. Net effect: during an elevated install, Cancel may appear to do nothing for the duration while the user is locked out of navigation — the exact lockout Item 1 set out to remove, only narrowed to the elevated-install window. Needs verifying against the real installer's elevation/process model; possible mitigations include a per-user install location (no elevation), launching elevated up front, or at least logging "cancelling — will take effect once the installer finishes".

src/BloomExe/Publish/Rab/RabProjectService.cs
StephenMcConnel: [Claude Opus 4.8 (1M context)] Already addressed in c4688ffab / d983d83e5 — resolving.

The cleanup is no longer BuildRoot-only. On the abnormal (cancel/fail) build path, `DeleteApksWrittenDuringBuild` now also runs over SafeApkRoot, and `IsCompleteApk` deletes exactly the truncated case described here: an APK this attempt created or rewrote whose zip central directory is missing/unreadable (a write cut off mid-stream). A fully-written APK is kept. So a killed rebuild can no longer leave a truncated file at the deliverable name that `FindLatestApkPath` would surface as a finished app. Covered by `BuildAsync_WhenKilledWhileWritingDeliverableApk_...` (truncated → deleted) and `BuildAsync_WhenAbnormalAfterApkFinished_...` (complete → kept).

src/BloomExe/Publish/Rab/RabPublishApi.cs
StephenMcConnel: [Claude Opus 4.8 (1M context)] Real, and it crosses the line for an autonomous preflight fix, so I'm routing it to the developer rather than fixing it in this run — see the preflight decision report. Leaving this thread open until that decision comes back.

src/BloomExe/Publish/Rab/RabProjectService.cs
StephenMcConnel: [Claude Opus 4.8 (1M context)] Real, and it crosses the line for an autonomous preflight fix, so I'm routing it to the developer rather than fixing it in this run — see the preflight decision report. Leaving this thread open until that decision comes back.

src/BloomExe/Publish/Rab/RabPublishApi.cs
StephenMcConnel: [Claude Opus 4.8 (1M context)] Fixed in 555196881 — the developer chose to reference-count the lock.

`WorkspaceView.SetTabsEnabled` is now keyed by reason: `SetTabsEnabled(bool, reason)` adds/removes the caller's reason from a set, and the tabs re-enable only once the set is empty. All five callers pass a distinct reason (Apps action, library upload, Edit modal, Edit save), so a library upload and an Apps action no longer clobber each other's lock, and either ordering keeps the tabs locked until *both* release. Being idempotent per reason, it's also safe for the Edit callers that reassert the same state (nested modals / save-state transitions) rather than pairing acquire/release. PublishView's "clear a stuck Edit lock" safety net is now `ReleaseEditTabLocks()`, which clears only the Edit reasons, so it can't clobber an in-progress upload. BloomExe builds clean and the full C# suite is green.

src/BloomExe/Publish/Rab/RabProjectService.cs
StephenMcConnel: [Claude Opus 4.8 (1M context)] Developer decision: leave as is — resolving.

We looked at this and decided not to change it for now: during a *first-run, elevated* Reading App Builder install, Cancel is deferred — it takes effect once the installer finishes rather than killing the elevated process. It's a narrow window (only the very first install on a machine, and only while the elevated setup runs), and the outcome is recoverable (the install completes or is re-run). The pre-launch and post-exit cancel checks and the process-tree kill from Item 1 still apply to the non-elevated cases. Recording the decision here per the review; no code change.

src/BloomExe/Workspace/WorkspaceView.cs
```diff
@@ -1750,11 +1770,33 @@ private void ReportAndLogProblem()
...
+        /// </summary>
+        public void SetTabsEnabled(bool enable, string reason)
+        {
+            if (string.IsNullOrEmpty(reason))
+                throw new ArgumentException("A tab-lock reason is required.", nameof(reason));
+            // Guard the set and the derived flag: this runs on BloomServer request threads and
+            // background tasks as well as the UI thread. Snapshot the holders inside the lock too,
+            // so the log line below can't enumerate the set while another thread mutates it.
+            string holders;
+            lock (_tabDisablersLock)
```
StephenMcConnel: [Devin] **Bug**: Tab-lock bookkeeping was updated from background threads without synchronization

The reference-counted tab lock added for this PR keeps its reasons in a `HashSet<string>` (`_tabDisablers`), but `SetTabsEnabled` is called not only from the UI thread: the RAB prepare/build/install endpoint handlers run on BloomServer request threads and re-enable in a `Task.Run` `finally` on a thread-pool thread, and the library-upload and Edit-modal endpoints are off the UI thread too. The old code wrote a single bool (atomic), so cross-thread calls were harmless; concurrent `Add`/`Remove` on a plain `HashSet` can drop an entry or corrupt the set, after which the wrong enabled state is broadcast (tabs stuck disabled, or clickable while an action runs).

src/BloomExe/Publish/Rab/RabProjectService.cs
```diff
@@ -924,17 +1053,41 @@ private void Install()
             {
                 if (IsUpdateIncompatibleInstallFailure(installResult.Output))
                 {
+                    // The first attempt failed only because a differently-signed copy is already
+                    // installed, and that copy is still intact. If the user asked to cancel during
+                    // that attempt, stop here — before the destructive uninstall — so their existing
+                    // app is left untouched.
+                    ThrowIfCancellationRequested();
```
StephenMcConnel, line 1115: [Devin] **Investigate**: Cancel isn't wired into the "checking for a connected phone" step

Install's first step, `GetSingleConnectedDevice`, runs `adb devices -l` through `RabAdbHelper`, which starts and waits on its own process without registering it in `_currentProcess`. So a Cancel clicked during "Checking for a connected Android device…" won't kill that process; it only takes effect at the next cancellation-aware subprocess call. `adb devices` is usually fast, but it implicitly starts the adb server, which can take several seconds or hang on a bad USB/driver state — exactly when a user reaches for Cancel, and now the workspace tabs are locked while they wait. Same class as the download-stall cancellation this PR already handles.

src/BloomExe/Workspace/WorkspaceView.cs
StephenMcConnel: [Claude Opus 4.8 (1M context)] Fixed in 47ffbc498. All access to _tabDisablers and the derived _tabsEnabled flag is now guarded by a lock (in both SetTabsEnabled and ReleaseEditTabLocks), and the holder list for the diagnostic log is snapshotted inside the lock so it can't enumerate the set mid-mutation. SendTopBarState and logging run outside the lock. BloomExe builds clean and the full C# suite is green.

src/BloomExe/Publish/Rab/RabProjectService.cs
StephenMcConnel: [Claude Opus 4.8 (1M context)] Real, and in the spirit of this PR's stall-cancellation work — but I'm routing it to the developer rather than fixing it in this final cycle. On inspection the fix means changing a shared static adb helper whose stderr/exit-code handling differs from the cancellation-aware RunProcessCapturingOutput, and the method is stubbed in the test harness (so neither path is unit-covered) — enough of a design/risk call that it shouldn't be done blind at the end of the run. Captured in the preflight decision report with a recommendation to route device detection through the cancellation-aware path (or register its process). Leaving this thread open until that decision comes back.

src/BloomExe/Publish/Rab/RabProjectService.cs
```diff
@@ -3682,6 +4031,145 @@ internal virtual string FindAppDefPath(RabWorkspacePaths paths)
...
 
+        /// <summary>
+        /// Deletes any intermediate .apk files Gradle may have written under the build folder. These
+        /// are never the deliverable — a finished build signs the app and writes it to SafeApkRoot —
+        /// so removing them after an interrupted or failed build keeps a partial or unsigned
+        /// intermediate from later being picked up by <see cref="FindLatestApkPath"/> as if it were a
+        /// finished app. SafeApkRoot is intentionally left alone so a failed rebuild still keeps the
+        /// previous good APK.
+        /// </summary>
+        private static void DeleteIntermediateBuildApks(RabWorkspacePaths paths)
```
StephenMcConnel, line 4123: [Devin] **Bug**: DeleteIntermediateBuildApks can delete a deliverable APK found under BuildRoot

`DeleteIntermediateBuildApks` deletes *every* `*.apk` under `BuildRoot` on a cancelled/failed build, on the stated assumption that BuildRoot only ever holds intermediates and the deliverable lives in SafeApkRoot. But `FindLatestApkPath` searches `BuildRoot` and `RabRoot` (recursively, which contains `build/`) in addition to the deliverable roots and returns the newest `*.apk` there — so the code's own contract says an APK under BuildRoot *can* be treated as the current app. If a deliverable ever lands under BuildRoot/RabRoot, the next cancelled/failed build silently destroys it. Its sibling `DeleteApksWrittenDuringBuild` guards SafeApkRoot with a write-time snapshot + completeness check; BuildRoot cleanup has no such guard. (The unit test only exercises the test double, which by construction writes to SafeApkRoot.)

src/BloomExe/Publish/Rab/RabProjectService.cs
StephenMcConnel, line 900: [Devin] **Investigate**: A late cancel after a successful build leaves "Build" still showing as needed

When a cancel is requested but RAB already exited 0 with a complete APK, the cleanup correctly keeps the APK — but the `OperationCanceledException` unwinds out of `Build()` before its tail runs, so `state.LastBuiltInputSignature`/`LastBuiltApkPath` are never saved. Result: `apkExists` is true and "Try on phone" would install the freshly built app, yet GetStatus still reports Build as needed because the saved signature doesn't match. Cosmetic/state inconsistency rather than data loss.

src/BloomExe/Publish/Rab/RabProjectService.cs
```diff
@@ -2211,12 +2371,50 @@ internal virtual void RunProcess(
...
+                    lock (_currentProcessLock)
+                        _currentProcess = process;
+                }
+                // Close the race between Start() above and registering _currentProcess: a cancel that
+                // arrived in that window found no process to kill, so kill it now.
+                if (!_protectCurrentProcessFromCancellation && _cancelRequested)
+                {
+                    try
+                    {
+                        process.Kill(true);
```
StephenMcConnel, line 2452: [Devin] **Investigate**: A cancel in the race-kill window may be reported as a failure, not a cancellation

The race-closing `process.Kill(true)` runs before `BeginOutputReadLine()`/`BeginErrorReadLine()`. If the cancel lands in that narrow window, the process may already be torn down when the async readers start, and `BeginOutputReadLine` on an exited process can throw `InvalidOperationException` — which is not `OperationCanceledException`, so RabPublishApi's `when (IsCancellationRequested)` filter won't match and the user's cancel is reported via ReportFailure as an error. Same pattern in RunProcess and RunProcessCapturingOutput.

src/BloomExe/Publish/Rab/RabProjectService.cs
```diff
@@ -2632,7 +2945,43 @@ internal virtual void InstallRabFromSetup(string installerPath)
...
+                lock (_currentProcessLock)
+                    _currentProcess = process;
+            }
+            // Close the race between StartShellProcess above and registering _currentProcess: a
+            // cancel that arrived in that window found no process to kill, so kill it now.
+            if (!_protectCurrentProcessFromCancellation && _cancelRequested)
+            {
+                try
+                {
+                    process.Kill(true);
```
StephenMcConnel, line 3039: [Devin] **Investigate**: Cancelling the installer mid-run could leave a half-installed state treated as installed

Killing the installer is reported as a clean cancel on the grounds that "a killed install is recoverable — the next Prepare re-runs it." That holds only if `IsRabInstalledForPrepare` reports a half-installed tree as not-installed. It checks for `rab.bat` and `runtime/bin/keytool.exe`; an installer killed after writing those but before finishing would pass the check, so the next Prepare proceeds against a broken installation instead of re-running setup. Unlike the download and build paths, the install path has no cleanup/rollback of a partial install.

src/BloomExe/Publish/Rab/RabProjectService.cs
StephenMcConnel: [Claude Opus 4.8 (1M context)] Real, and noted. This is part of a cluster of non-severe edge cases in the cancellation/cleanup logic that Devin surfaced across successive re-reviews; the preflight fix loop has reached its 4-cycle cap, so rather than keep iterating I'm routing this batch to the developer in the decision report with recommendations. Leaving this thread open until that decision comes back.

src/BloomExe/Publish/Rab/RabProjectService.cs
StephenMcConnel: [Claude Opus 4.8 (1M context)] Real, and noted. This is part of a cluster of non-severe edge cases in the cancellation/cleanup logic that Devin surfaced across successive re-reviews; the preflight fix loop has reached its 4-cycle cap, so rather than keep iterating I'm routing this batch to the developer in the decision report with recommendations. Leaving this thread open until that decision comes back.

src/BloomExe/Publish/Rab/RabProjectService.cs
StephenMcConnel: [Claude Opus 4.8 (1M context)] Real, and noted. This is part of a cluster of non-severe edge cases in the cancellation/cleanup logic that Devin surfaced across successive re-reviews; the preflight fix loop has reached its 4-cycle cap, so rather than keep iterating I'm routing this batch to the developer in the decision report with recommendations. Leaving this thread open until that decision comes back.

src/BloomExe/Publish/Rab/RabProjectService.cs
StephenMcConnel: [Claude Opus 4.8 (1M context)] Real, and noted. This is part of a cluster of non-severe edge cases in the cancellation/cleanup logic that Devin surfaced across successive re-reviews; the preflight fix loop has reached its 4-cycle cap, so rather than keep iterating I'm routing this batch to the developer in the decision report with recommendations. Leaving this thread open until that decision comes back.

src/BloomExe/Publish/Rab/RabProjectService.cs
StephenMcConnel: [Claude Opus 4.8 (1M context)] Fixed in 108bf76e3. FindLatestApkPath now searches only the deliverable roots (ApkRoot, then SafeApkRoot), not BuildRoot/RabRoot — so an unsigned Gradle intermediate is never returned as "the app", and DeleteIntermediateBuildApks (BuildRoot only) can no longer appear to delete a deliverable. ApkRoot stays first so it still wins the tie-break when a copy exists in both roots with the same timestamp. New test: FindLatestApkPath_ReturnsDeliverable_IgnoringApksUnderBuildRoot.

src/BloomExe/Publish/Rab/RabProjectService.cs
StephenMcConnel: [Claude Opus 4.8 (1M context)] Fixed in 108bf76e3. Build() now records a complete APK on the abnormal path too (RecordBuiltApk via FindCompleteApkWrittenThisBuild), so a late cancel that kept a finished app no longer leaves GetStatus reporting a build as needed. The guard ("written this attempt") ensures a surviving previous-build APK is not mis-recorded against the current inputs. New test: BuildAsync_WhenAbnormalAfterApkFinished_RecordsItSoBuildIsNotStillNeeded.

src/BloomExe/Publish/Rab/RabProjectService.cs
StephenMcConnel: [Claude Opus 4.8 (1M context)] Fixed in 108bf76e3. RunProcess and RunProcessCapturingOutput now catch the InvalidOperationException that the race-kill can cause when the async output readers attach to an already-exited process, and — when a cancel was requested — rethrow it as OperationCanceledException, so it's reported as a cancellation rather than a failure.

src/BloomExe/Publish/Rab/RabProjectService.cs
StephenMcConnel: [Claude Opus 4.8 (1M context)] Fixed in 108bf76e3, then refined in 9e91fc7f1. An interrupted Bloom-run install now forces a reinstall over the possibly half-installed tree. The first version used a persisted marker, but (good second catch by the re-review) that gated GetStatus and could hide a manually-completed install; the refinement uses an in-session in-memory flag instead — it re-forces the install within the session (the cancel-then-retry case), while GetStatus keeps reading the real files, so a genuinely-installed RAB is never reported missing and there's no stuck state.

src/BloomExe/Publish/Rab/RabProjectService.cs
StephenMcConnel: [Claude Opus 4.8 (1M context)] Fixed in 108bf76e3. GetSingleConnectedDevice now runs "adb devices -l" through the cancellation-aware RunProcessCapturingOutput (which registers the process for the cancel kill) and parses via the new RabAdbHelper.SelectSingleConnectedDevice, so a Cancel during "checking for a connected phone" can kill a stalled adb (e.g. while it starts the adb server) instead of only taking effect at the next step.

src/BloomExe/Publish/Rab/RabProjectService.cs
```diff
@@ -605,7 +713,11 @@ private bool EnsureRabInstalledForPrepare()
         {
             ReportProgressStage("checking-installer", 0);
 
-            if (IsRabInstalledForPrepare())
+            // If a Bloom-run install this session was interrupted, don't trust a possibly
+            // half-installed tree even if rab.bat + keytool happen to be present — reinstall (the
+            // installer is idempotent and completes/repairs). This gates only the install decision,
+            // not GetStatus, so a genuinely-installed RAB is never reported as missing (BL-16350).
+            if (!_rabInstallInterrupted && IsRabInstalledForPrepare())
```
StephenMcConnel, line 720: [Devin] **Investigate**: A persisted install-incomplete marker could hide a later manual install

An intermediate version of the half-installed-state fix used a marker file that IsRabInstalledForPrepare consulted; because that also feeds GetStatus, a cancelled/failed install made the whole Apps screen report RAB as not installed, and the marker could outlive the staged installer and hide a RAB install the user completed manually, with no way to clear it from the UI.

src/BloomExe/Publish/Rab/RabProjectService.cs
```diff
@@ -734,32 +846,68 @@ private void Build()
             );
             Directory.CreateDirectory(paths.SafeApkRoot);
 
+            // Snapshot the deliverable APKs before the build so the abnormal-path cleanup below can
+            // tell a previous good APK (untouched by this attempt) from one this attempt created or
+            // rewrote. See DeleteApksWrittenDuringBuild for why this matters.
+            var preBuildApkWriteTimes = SnapshotDeliverableApkWriteTimes(paths);
```
StephenMcConnel, line 852: [Devin] **Bug**: After an interrupted build, Bloom could mark an old app as freshly built

The pre-build snapshot recorded only SafeApkRoot, but FindLatestApkPath (used by the abnormal-path "did this build produce a complete APK?" check) also searches ApkRoot. A complete APK already in ApkRoot was therefore absent from the snapshot, looked "written this build", and on a cancelled/failed build got recorded as the current result — so GetStatus reported the app as up to date and would offer a stale APK.

src/BloomExe/Publish/Rab/RabProjectService.cs
StephenMcConnel: [Claude Opus 4.8 (1M context)] Fixed in 9e91fc7f1. Replaced the persisted marker with an in-session in-memory flag (_rabInstallInterrupted): it still forces a reinstall after an interrupted install within the session (the cancel-then-retry case), but GetStatus keeps reading the real install files, so a genuinely-installed RAB — including one installed outside Bloom — is never reported missing, and there is no persisted state that could get stuck.

src/BloomExe/Publish/Rab/RabProjectService.cs
StephenMcConnel: [Claude Opus 4.8 (1M context)] Fixed in fa3915bb2. Introduced GetExistingDeliverableApkRoots as the single source of truth for the deliverable roots (ApkRoot, then SafeApkRoot) and made FindLatestApkPath, the pre-build snapshot (now SnapshotDeliverableApkWriteTimes), and DeleteApksWrittenDuringBuild all use it. An APK can no longer be in a root the snapshot missed, so a pre-existing ApkRoot APK is correctly seen as pre-existing (not "written this build") and a failed build no longer records it as current. Test: BuildAsync_WhenFailedWithAnOlderApkInApkRoot_DoesNotRecordItAsFreshlyBuilt.

StephenMcConnel (commented):
> @StephenMcConnel+CLAU resolved 17 discussions.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8133)* status: 0 of 12 files reviewed, all discussions resolved (waiting on StephenMcConnel).

andrew-polk:
*[`src/BloomExe/Workspace/WorkspaceView.cs` line 72 at r6](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8133#-Oz8xGEK-_iMiJRx6H-c:-Oz8xGEK-_iMiJRx6H-d:b-kejf5p) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/fa3915bb29e8642d8163a79dad3a28f0656ac646/src/BloomExe/Workspace/WorkspaceView.cs#L72)):*
> ```Smalltalk
>         public const string EditSaveTabLock = "edit-save-transition";
>         public const string LibraryUploadTabLock = "library-upload";
>         public const string AppBuilderActionTabLock = "app-builder-action";
> ```

This new machinery feels like the kind of thing the AIs like to add to prevent things which can never happen. Do we need to add this complexity?

StephenMcConnel (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8133)* status: 0 of 12 files reviewed, 1 unresolved discussion (waiting on andrew-polk and StephenMcConnel).

___
*[`src/BloomExe/Workspace/WorkspaceView.cs` line 72 at r6](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8133#-Oz8xGEK-_iMiJRx6H-c:-OzD7C230pS3ykagg1aq:b-k9t99a) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/fa3915bb29e8642d8163a79dad3a28f0656ac646/src/BloomExe/Workspace/WorkspaceView.cs#L72)):*

This new machinery feels like the kind of thing the AIs like to add to prevent things which can never happen. Do we need to add this complexity?
</blockquote></details>

This is primarily to prevent starting a library upload, switching to the Apps tool, and starting a build (or other lengthy) operation, and having the library upload finish and turn off the modal behavior for the major tabs before the build finishes.  This may seem far-fetched, and maybe a better fix would be to keep from switching publish tools while a library upload is in process.

src/BloomExe/Publish/Rab/RabPublishApi.cs
StephenMcConnel: [Claude Opus 5 (1M context)] Reverted in 24572e255 — the reference-counted lock is out, and this finding is now tracked separately as BL-16654.

On review the developer judged the fix disproportionate to the card: ~40 lines plus a new keyed-lock concept threaded through five callers, to close an edge case that needs a specific interleaving (start an upload, switch to the Apps tool, run an action), none of which is what BL-16350 is about. So `WorkspaceView.SetTabsEnabled(bool)` is back to exactly what it was before this PR, and a RAB action simply becomes one more caller of it, just as the library upload already was. `PublishView.Activate` is back to `SetTabsEnabled(true)`, and `ReleaseEditTabLocks()`, the reason constants and the disabler set are all gone.

The problem you describe is not dismissed, just scoped out: making the tab lock compose across its several independent users (Apps action, library upload, Edit modal, Edit save) is now https://issues.bloomlibrary.org/youtrack/issue/BL-16654. `RabPublishApi.SetWorkspaceTabsEnabled` carries a comment noting that `SetTabsEnabled` is a single shared flag, what goes wrong if an upload and an Apps action overlap, and that BL-16654 owns the fix — so it can't get lost. Full C# suite green after the revert (2991 passed, 0 failed).

src/BloomExe/Workspace/WorkspaceView.cs
StephenMcConnel: [Claude Opus 5 (1M context)] Moot as of 24572e255 — the `HashSet` this was about no longer exists.

The reference-counted lock it guarded has been reverted (see the tab-lock thread on `RabPublishApi.cs`, and BL-16654 which now owns that work). `SetTabsEnabled` is back to writing a single `bool`, which is atomic, so the cross-thread calls you correctly identified — BloomServer request threads, the `Task.Run` `finally` blocks, the library-upload and Edit-modal endpoints — are harmless again and no lock is needed.

Worth carrying forward: whatever replaces the boolean under BL-16654 will need exactly the synchronization you flagged here. Full C# suite green after the revert (2991 passed, 0 failed).

src/BloomExe/Publish/Rab/RabProjectService.cs
```diff
@@ -734,32 +846,68 @@ private void Build()
...
+                DeleteIntermediateBuildApks(paths);
+                DeleteApksWrittenDuringBuild(paths, preBuildApkWriteTimes);
+                // If RAB finished a complete, signed APK before a late cancel (or a failure reported a
+                // moment later) arrived, DeleteApksWrittenDuringBuild kept it — so record it as the
+                // built app too. Otherwise apkExists would be true while GetStatus still reported a
+                // build as needed, because the signature is only saved on the normal path below
+                // (BL-16350; Devin). FindCompleteApkWrittenThisBuild guards on "written this attempt"
+                // so a surviving previous-build APK is not mis-recorded against this build's inputs.
+                var finishedApk = FindCompleteApkWrittenThisBuild(paths, preBuildApkWriteTimes);
+                if (finishedApk != null)
```
StephenMcConnel, line 908: [Devin] **Investigate**: Recording a build signature on the abnormal path can mark a cancelled build as "current"

On the cancel/failure path, if `FindCompleteApkWrittenThisBuild` finds a complete APK written during this attempt, `RecordBuiltApk` saves `LastBuiltInputSignature` so `GetStatus().BuildNeeded` becomes false. This assumes RAB writes the deliverable to `apk.output` as its truly final step; if RAB ever performs post-copy work (e.g. re-signing/aligning in place, or writing additional artifacts) after that copy, a cancel in that window would leave Bloom telling the user the app is up to date when it is not. Worth confirming against RAB's actual output ordering.

StephenMcConnel: [Claude Opus 5 (1M context)] Assessed: this is the deliberate side of a trade-off the developer already decided, not an unnoticed hole — so recording the APK stays. Reasoning, so a reviewer can overturn it if they disagree:

This exact behavior was added to answer your earlier flag *"A late cancel after a successful build discards the build signature, so Build still shows as needed"*. The two flags pull in opposite directions: if we don't record, a user who cancels a moment too late has a perfectly good app on disk but Bloom keeps telling them to build again; if we do record, a cancel in a hypothetical post-copy window could mark a not-quite-finished app as current. The developer chose the second, because the first case is the one users actually hit.

The recording is guarded, not blanket. `FindCompleteApkWrittenThisBuild` only returns a path when the APK (a) sits at a deliverable root, (b) opens as a readable archive with entries — a mid-write truncation fails this and gets deleted by `DeleteApksWrittenDuringBuild` instead — and (c) was written or rewritten during *this* attempt, so a surviving previous-build APK is never re-recorded against this build's inputs. Tests cover all three directions: `BuildAsync_WhenAbnormalAfterApkFinished_RecordsItSoBuildIsNotStillNeeded`, `BuildAsync_WhenKilledWhileWritingDeliverableApk_DeletesTruncatedApkButKeepsUntouchedApk`, and `BuildAsync_WhenFailedWithAnOlderApkInApkRoot_DoesNotRecordItAsFreshlyBuilt`.

On "worth confirming against RAB's actual output ordering": we can't see inside RAB, and that limit is already written down in the code comment on this very block ("Bloom cannot see whether RAB writes that final APK atomically"). The residual risk is therefore narrow and cheap: cancel in the last instant of a build, and the worst case is an app that looks current but may need one more Build. The user can always press Build again, and nothing is silently lost. Resolving.

StephenMcConnel:
[Claude Opus 5 (1M context)] Consulted Devin on 2026-08-05 10:40 MDT up to commit `f5cf81d1b03fb27327068db3f655c49b8ae74744` (the boolean-restore commit plus a fresh merge of master).

Devin now considers **all 9 Bugs it has raised on this PR resolved**, including the two about the tab lock — it notes explicitly that "the current revision has returned to a simple atomic boolean, so the HashSet concurrency concern no longer applies." Those two threads already carry replies explaining the revert and pointing at BL-16654, which now owns making the tab lock compose across its several users.

One new Investigate flag this round — that recording a finished APK on the cancel path can mark a cancelled build as "current". Assessed and answered on its own thread (#discussion_r3722359694): it is the deliberate side of a trade-off the developer already made, it is guarded three ways, and it is covered by tests. The other 7 Investigate flags are the same ones already mirrored and answered on this PR earlier. 27 informational items, not mirrored (low signal).

CI is green and no other review bot has anything outstanding.

StephenMcConnel (commented):
> @StephenMcConnel+CLAU made 1 comment and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8133)* status: 0 of 12 files reviewed, 1 unresolved discussion (waiting on andrew-polk and StephenMcConnel).

___
*[`src/BloomExe/Workspace/WorkspaceView.cs` line 72 at r6](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8133#-Oz8xGEK-_iMiJRx6H-c:-OzHnoHz9kAI3HdfiRfQ:b-821fib) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/fa3915bb29e8642d8163a79dad3a28f0656ac646/src/BloomExe/Workspace/WorkspaceView.cs#L72)):*

This is primarily to prevent starting a library upload, switching to the Apps tool, and starting a build (or other lengthy) operation, and having the library upload finish and turn off the modal behavior for the major tabs before the build finishes.  This may seem far-fetched, and maybe a better fix would be to keep from switching publish tools while a library upload is in process.
</blockquote></details>

[Claude Opus 5 (1M context)] Andrew's instinct won: this machinery is gone as of 24572e255. `SetTabsEnabled` is back to the plain boolean it was before this PR, and the Apps action is simply one more caller of it, exactly like the library upload already was — no reason constants, no disabler set, no lock, and `PublishView.Activate` back to `SetTabsEnabled(true)`. The PR is 63 lines lighter and the full C# suite is still green.

The upload-then-Apps overlap is not being ignored, just moved out of this card: it is now BL-16654 (https://issues.bloomlibrary.org/youtrack/issue/BL-16654). Steve's own suggestion above — blocking the publish-tool switcher while a library upload is running — is a much cheaper answer than a keyed lock, and that issue is where it belongs. `RabPublishApi.SetWorkspaceTabsEnabled` now carries a short comment saying `SetTabsEnabled` is a single shared flag, what goes wrong if an upload and an Apps action overlap, and that BL-16654 owns the fix, so the next person to touch it has the context.

src/BloomExe/Publish/Rab/RabProjectService.cs
StephenMcConnel: [Claude Opus 5 (1M context)] Developer decision: leave as is. Steve reviewed this in the preflight report and confirmed the behavior stands — recording a finished, complete APK on the cancel path is the intended trade-off against the "a late cancel leaves Build still showing as needed" flag, and the three guards plus their tests are considered sufficient. No code change, and the code comment on this block already states the limit we cannot see past (whether RAB writes that final APK atomically).

andrew-polk (approved):
> @andrew-polk reviewed 12 files and all commit messages, and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8133)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on StephenMcConnel).

andrew-polk (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8133)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on StephenMcConnel).

PR #8135, opened 2026-08-01
BL-16627: improve BloomBridge process-book fix-up: auto-fit origami splits (both directions, nested stacks) and shrink oversized images

hatton:
Improvements to the `external/process-book` fix-up that BloomBridge drives (replaces #8129, which closed when the branch was renamed). It solves four problems that made imported books arrive with clipped text and wasted space.

1. Auto-fit could only grow the image pane, so a page with a wide illustration kept its oversized picture box while the text below stayed clipped; the divider now moves in either direction.
2. Auto-fit skipped nested splits entirely, which excluded the most common scanned-book shape of all — text above, an illustration, text below — leaving those pages at an effective 50/25/25 that clips the longest text block. It now fits any top-to-bottom stack of picture and text panes, including stacks with several illustrations.
3. Oversized images that BloomBridge wrote straight into the book folder (bypassing Bloom's normal add-image shrink) are now downsized at the start of the pass.
4. Pages auto-fit declines to touch are now restored byte-for-byte, so they no longer show up as noise in the saved HTML's diff.

Ref: [BL-16627](https://issues.bloomlibrary.org/youtrack/issue/BL-16627)

---
[Devin review](https://app.devin.ai/review/BloomBooks/BloomDesktop/pull/8135)

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8135)

[Devin review](https://devinreview.com/BloomBooks/BloomDesktop/pull/8135)


src/BloomBrowserUI/bookEdit/js/autoFitImageOverTextSplits.ts
```diff
@@ -162,8 +196,10 @@ function fitImageOverTextSplitOnPage(page: HTMLElement): boolean {
     }
     const minTextPercent = hi;
 
-    // Cap how far the image can grow: once it fills the page width, growing the image pane further
-    // just adds whitespace around the image, so don't shrink the text below that point.
+    // The point past which the image gains nothing: once it fills the page width (or height, for a
```
hatton, line 199: [Devin] **Investigate**: Image-fit cap in the two-pane path is still measured after the binary search, unlike the stack path

The new stack fitter deliberately measures each illustration's fill point up front (`src/BloomBrowserUI/bookEdit/js/autoFitImageOverTextSplits.ts:526-532`) with the comment that probing squeezes the panes, so measuring afterward would read a picture's chrome out of a collapsed pane. The two-pane path still calls `computeImageFitFirstPanePercent` here, i.e. after the binary search has left the split at the last probe position, so `extra = firstComponent.offsetHeight - firstCanvas.offsetHeight` is read from whatever pane size the search happened to end on. I did not flag this as a bug because the background canvas element is sized in explicit px by `adjustBackgroundImageSizeToFit` (see `CanvasElementBackgroundImageManager.ts:291-319`), so the aspect ratio is stable, and `extra` is normally just constant padding. Still, now that this value can also SHRINK the image (it is used as a text floor), it would be more robust — and consistent with the stack path — to measure it before probing.

src/BloomBrowserUI/bookEdit/js/autoFitImageOverTextSplits.ts
```diff
@@ -176,21 +212,438 @@ function fitImageOverTextSplitOnPage(page: HTMLElement): boolean {
...
+    const textGroup = inner.querySelector(
+        ".bloom-translationGroup",
+    ) as HTMLElement | null;
+    if (canvas && !textGroup) {
+        // Overlays (canvas elements other than the background image) put the page out of scope for
+        // the same reason as in the two-pane case: our math reasons only about the background
+        // image's aspect ratio, and overlays don't scale with it predictably. Not redundant with the
+        // `!textGroup` test above, though it looks it: that one already excludes TEXT bubbles (a
+        // bubble contains a translationGroup), so this is what excludes image/video/other overlays.
+        if (
```
hatton, line 315: [Devin] **Investigate**: A pane holding more than one translation group is only measured on its first one

`classifySlot` records a single `textGroup` (`inner.querySelector(".bloom-translationGroup")`) and `textGroupOverflows` only inspects that group's editables. If a pane ever contained two translation groups (or a translation group plus another text block), the binary search would size the pane to fit only the first, and the second could stay clipped after the fit — the exact failure the feature is meant to prevent. The two-pane path has always had this shape (`secondTextGroup` is also a single `querySelector`), so it is pre-existing, but the stack fitter widens the set of pages it applies to. Origami panes normally hold exactly one group, so this is an edge case rather than an observed failure; a cheap hardening would be to classify only panes with exactly one translation group, or to measure all groups in the pane.

src/BloomExe/Book/BookProcessor.cs
```diff
@@ -74,6 +83,30 @@ public static int ProcessBook(Book book, bool fitImageTextSplits = false)
...
+            // (not gated by mediaMaintenanceLevel), and it must come BEFORE BringBookUpToDate: (a) on
+            // a book old enough that the migration WOULD run, it normally finds nothing left to
+            // shrink, so its modal progress dialog is not created on this background thread (where a
+            // WinForms dialog is illegal, BL-16646) -- "normally" because a GraphicsMagick failure can
+            // leave an image oversized, in which case that narrow old-book case can still hit the
+            // dialog; (b) the off-screen per-page fix-up then measures and lays out against the
+            // final, already-shrunk images.
+            if (ImageUtils.NeedToShrinkImages(book.FolderPath))
+            {
+                Log("shrinking oversized images in the book folder");
```
hatton, line 100: [Devin] **Investigate**: Residual risk that the migration's modal dialog can still appear on the background thread

The ordering argument (shrink first so `MigrateToMediaLevel1ShrinkLargeImages` finds nothing to do) depends on the pre-shrink actually clearing every oversized image. `ImageUtils.ResizeImageFileWithOptionalTransparency` returns false without doing anything when GraphicsMagick is absent (`src/BloomExe/ImageProcessing/ImageUtils.cs:1608-1610`), and `NeedToShrinkImages` swallows per-file exceptions. In that case `BookStorage.MigrateToMediaLevel1ShrinkLargeImages` (`src/BloomExe/Book/BookStorage.cs:4057-4090`) will still see `NeedToShrinkImages == true` for a level-0 book and construct `ProgressDialogBackground` on the process-book background thread — the exact BL-16646 failure the ordering is meant to prevent. Only reachable for old books on a machine without GraphicsMagick, so I did not report it, but it is worth knowing the guarantee is best-effort rather than absolute.

src/BloomBrowserUI/bookEdit/js/autoFitImageOverTextSplits.ts
hatton: [Claude Fable 5 from Hatton's machine during preflight] Not acting on this: in the two-pane path the measurement order doesn't affect the result. `.split-pane-component-inner` is always 100% height with only padding between it and the canvas, so the `extra` term (`firstComponent.offsetHeight - firstCanvas.offsetHeight`) is constant across probe positions — measuring before or after the binary search reads the same value. The stack path measures up front because there a probe genuinely squeezes panes whose chrome is not constant. Devin itself rates this a robustness/consistency point rather than a bug. Resolving.

src/BloomBrowserUI/bookEdit/js/autoFitImageOverTextSplits.ts
hatton: [Claude Fable 5 from Hatton's machine during preflight] Fixed in d5f97adc40: both the stack path (`classifySlot`) and the two-pane path now decline any pane holding more than one translation group, following the code's existing decline-rather-than-guess rule, with a spec test covering the new decline. Such pages are left exactly as authored. Resolving.

src/BloomExe/Book/BookProcessor.cs
hatton: [Claude Fable 5 from Hatton's machine during preflight] Agreed, and this is now documented at exactly this spot: the comment (corrected in fd1d84d0be) says the migration "normally" finds nothing left to shrink and names the GraphicsMagick-failure case as the residual way the dialog could still appear. The window is narrow — a mediaMaintenanceLevel-0 book on a machine where GraphicsMagick is missing or fails — and predates this PR's shrink step (which only narrows it). Accepting that residual risk rather than adding code. Resolving.

src/content/templates/xMatter/bloom-xmatter-mixins.pug
```diff
@@ -275,7 +275,7 @@ mixin factory-insideFrontCover
 
 mixin factoryStandard-insideBackCover
 	// Inside Back Cover
-	+page-xmatter('Inside Back Cover').cover.coverColor.insideBackCover.bloom-backMatter(data-export='back-matter-inside-back-cover', data-xmatter-page='insideBackCover')&attributes(attributes)#502BE62F-A4D0-4225-A598-1A203FA73239
+	+page-xmatter('Inside Back Cover').cover.coverColor.insideBackCover.bloom-backMatter(data-export='back-matter-inside-back-cover', data-xmatter-page='insideBackCover', data-custom-layout-id='customInsideBackCover')&attributes(attributes)#502BE62F-A4D0-4225-A598-1A203FA73239
```
hatton: [Devin] **Investigate**: New xmatter data-custom-layout-id has no matching branding CSS, unlike its outside-back-cover twin

`data-custom-layout-id='customInsideBackCover'` is added to two inside-back-cover xmatter pages. The existing analogue, `customOutsideBackCover`, has a corresponding rule in `src/content/branding/branding-base.less:133` (`.outsideBackCover.bloom-customLayout[data-custom-layout-id="customOutsideBackCover"]`); there is no equivalent rule for the new id. The attribute is still functional on its own — `XMatterHelper.CollectCustomLayoutIds`/`RestoreCustomLayoutClasses` (`src/BloomExe/Book/XMatterHelper.cs:540-570`) use it to preserve the `bloom-customLayout` class across xmatter updates — but if the intent was parity with the outside back cover, the branding LESS side looks unfinished. This change is also unrelated to the PR's stated scope (auto-fit + image shrink), so it may be an accidental inclusion.

src/BloomBrowserUI/bookEdit/js/autoFitImageOverTextSplits.ts
```diff
@@ -176,21 +212,440 @@ function fitImageOverTextSplitOnPage(page: HTMLElement): boolean {
...
+            config.divider,
+            config.secondComponent,
+        ])
+            saved.push([el, el.getAttribute("style")]);
+    return saved;
+}
+
+// Put back the style attributes captured by captureSplitStyles exactly as they were, removing the
+// attribute entirely where there wasn't one, so a declined page ends up byte-for-byte as authored.
+export function restoreSplitStyles(
```
hatton, line 421: [Devin] **Bug**: New exported helper for undoing layout changes has no documenting comment

The newly exported function that puts a page's original layout styles back (`restoreSplitStyles` at `src/BloomBrowserUI/bookEdit/js/autoFitImageOverTextSplits.ts:419`) is added with no comment of its own, which the repository's coding rules require for public methods ("All public methods should have a comment. So should most private ones!").

hatton: [Claude Fable 5 from Hatton's machine during preflight] Fixed in a8b417c70e: added the doc comment. Resolving.

src/BloomExe/Book/BookProcessor.cs
```diff
@@ -74,6 +83,30 @@ public static int ProcessBook(Book book, bool fitImageTextSplits = false)
...
+                var shrinkTimer = Stopwatch.StartNew();
+                ImageUtils.FixSizeAndTransparencyOfImagesInFolder(
+                    book.FolderPath,
+                    new List<string>(),
```
hatton, line 104: [Devin] **Investigate**: Image shrink path calls Application.DoEvents() on the process-book background thread

`ImageUtils.FixSizeAndTransparencyOfImagesInFolder` -> `ResizeImageFileWithOptionalTransparency` (`src/BloomExe/ImageProcessing/ImageUtils.cs:1650`) does `if (progress != null) Application.DoEvents();` after each successful resize. Because a non-null `NullProgress` is passed here, that branch is taken, and `ProcessBook` is documented (and invoked at `src/BloomExe/web/controllers/ExternalApi.cs:588`) as running off the UI thread. `Application.DoEvents()` only pumps the *calling* thread's message queue, so on a worker thread with no windows it is effectively a no-op — which is why I did not flag it — but it is worth confirming that the process-book worker thread never owns any window/STA COM apartment before relying on that. Same call chain also invokes `L10NSharp.LocalizationManager.GetString` off the UI thread.

hatton: [Claude Fable 5 from Hatton's machine during preflight] Not an issue: the shrink runs before the off-screen browser is created, so at that point the process-book worker thread owns no windows at all — `Application.DoEvents()` pumps only the calling thread's queue, so it is a no-op there. Even later, the WebView2 lives on OffScreenBrowser's own dedicated thread, never on this worker. The `GetString` calls off the UI thread are likewise safe (L10NSharp lookups don't touch UI). Devin itself rates this as confirm-only rather than a bug. Resolving.

hatton:
[Claude Fable 5 from Hatton's machine during preflight] Consulted Devin on 2026-08-03 up to commit a8b417c70e43d96c9b17f1c611c78b16f4757cf8 (three rounds, one per push). Across the rounds it raised one bug and five Investigate flags:

- Missing doc comment on `restoreSplitStyles` — fixed in a8b417c70e (thread: https://github.com/BloomBooks/BloomDesktop/pull/8135#discussion_r3704787218)
- A pane holding two translation groups was only measured on its first — fixed in d5f97adc40 (https://github.com/BloomBooks/BloomDesktop/pull/8135#discussion_r3704513866)
- Two-pane image-fit cap measured after the binary search — not an issue, reasoning on the thread (https://github.com/BloomBooks/BloomDesktop/pull/8135#discussion_r3704511275)
- Residual BL-16646 dialog risk when GraphicsMagick fails — accepted and documented in the code comment (https://github.com/BloomBooks/BloomDesktop/pull/8135#discussion_r3704514981)
- `Application.DoEvents()` on the background thread — not an issue, reasoning on the thread (https://github.com/BloomBooks/BloomDesktop/pull/8135#discussion_r3704867260)
- New `customInsideBackCover` layout id has no matching branding CSS — **left open for the developer** (https://github.com/BloomBooks/BloomDesktop/pull/8135#discussion_r3704784496)

Informational items (13 in the final round) were reviewed and not mirrored. CI is green and no other review bots are active on this PR.

src/content/templates/xMatter/bloom-xmatter-mixins.pug
hatton: [Claude Fable 5 from Hatton's machine during preflight] Decision from the developer: this inside-back-cover change is moving forward in an upstream branch instead, so it has been removed from this PR (reverted in 0d05e81010). Resolving.

hatton (commented):
> @hatton partially reviewed 6 files.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8135)* status: 4 of 7 files reviewed, 6 unresolved discussions.

StephenMcConnel (approved):
> @StephenMcConnel reviewed 7 files and all commit messages, and resolved 6 discussions.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8135)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on hatton).

PR #8138, opened 2026-08-03
Remove two sources of nightly CI failure (auto-update modal, urlPrefix mock ordering)

andrew-polk:
Two independent fixes for things making the nightly red, found while evaluating the 2026-08-03 nightly. Neither one touches the actual visual-regression hang — that is BL-16612, whose fix is on `BL-16612-page-checks-hang` (#8110). These just remove the noise around it.

### 1. Don't show the auto-update dialog in automated runs

On a fresh CI runner `Settings.Default.AutoUpdateDialogShown` is 0, so `WorkspaceView` opened the "Auto Update" `ReactDialog` modally as a startup action. With nobody to dismiss it, it sat on the UI thread in its own nested message loop for the whole run — all three stack captures from the 08-03 nightly show the main thread parked in `Form.ShowDialog` under `ShowAutoUpdateDialogIfNeeded`.

`Program.RunningE2eTests` already exists to suppress exactly this (its own comment: "a dialog nobody can dismiss and hanging the whole run") and covers modal error dialogs and `Debug.Assert` boxes. This nag dialog was simply never included.

This was **not** the cause of that night's hang — the visual-regression step passed on other nights with this dialog up — just an independent hazard.

### 2. Install the talking-book `urlPrefix` mock before the recorder exists

`talkingBookSpec > showTool(checksum=missing, audio=missing, scenario=PreTextBox) => UPDATE` failed **every** nightly since 2026-07-29, comparing `http://localhost:3000/...` against the expected `http://localhost:63315/...`, while passing locally both in isolation and under a full-suite run with the same worker count.

`63315` is not a real port; it is a literal inside a mock of `AudioRecording.urlPrefix` that `setupForAudioRecordingTests` installed on `theOneAudioRecorder` *after* awaiting `initializeTalkingBookToolAsync`. Two things made that fragile:

- `theOneAudioRecorder` does not exist until that call creates it (it is an `export let` populated by `getAudioRecorder`), so the mock could not have gone on the instance any earlier.
- The call itself already sets the player's `src`, so that first `src` came from the **real** `urlPrefix` — a relative URL, which jsdom resolves against its own `http://localhost:3000/` base.

`setCurrentAudioId` only refreshes the player when the audio id **changes**, so a test whose id was already current never overwrote that stale pre-mock value, and compared `:3000` against the mocked `:63315`. Which ids were already current depended on what earlier work happened to leave behind — hence local pass, CI fail.

Fix: mock the **prototype**, before initializing, so even the init-time write goes through the mock and no stale value exists to inherit.

One deliberate consequence: the mock now applies to every `AudioRecording` instance rather than only the singleton. That matches the mock's intent ("in tests, `urlPrefix` returns this absolute URL") and the suite confirms it.

### Verification

- `pnpm typecheck` — passed
- `pnpm lint` — 0 errors (774 pre-existing repo-wide warnings; none in the changed files)
- `build/agent-dotnet.sh build BloomExe` — 0 errors
- Front-end suite — 578 passed / 5 skipped, unchanged from the pre-fix baseline
- C# suite — see checks

The full-suite numbers being *unchanged* is the honest reading: it confirms no regression, not that the CI failure is fixed. That test never reproduced locally, so fix 2 is justified by mechanism, and confirming it needs a nightly (or `workflow_dispatch`) run on this branch. Fix 1 likewise can only be observed in CI.

No tracker card — this branch carries no `BL-` id (small CI cleanups). Related context: BL-16612.

[Devin review](https://devinreview.com/BloomBooks/BloomDesktop/pull/8138)

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8138)

### DIFF
--- a/src/BloomBrowserUI/bookEdit/toolbox/talkingBook/audioRecordingSpec.ts
+++ b/src/BloomBrowserUI/bookEdit/toolbox/talkingBook/audioRecordingSpec.ts
@@ -2941,16 +2941,25 @@ export async function setupForAudioRecordingTests() {
         },
     };
 
-    await initializeTalkingBookToolAsync();
-
     // Mock urlPrefix to return the correct base URL for tests
     // The real implementation constructs from iframe.src, but in tests iframe.src is "about:blank"
     // Real format: "/bloom/api/audio/wavFile?id=" + bookFolderUrl + "audio/"
     // In tests, we simulate the full path as if bookFolderUrl was "http://localhost:63315/bloom/C%23Injection/"
+    //
+    // This must go on the PROTOTYPE and BEFORE initializeTalkingBookToolAsync, not on
+    // theOneAudioRecorder afterwards. theOneAudioRecorder does not exist until that call creates
+    // it, and the call itself already sets the player's src. So a mock installed afterwards left
+    // that first src built by the REAL urlPrefix -- a relative URL, which jsdom resolves against
+    // its own base of http://localhost:3000/. Because setCurrentAudioId only refreshes the player
+    // when the audio id CHANGES, a test whose id was already current never overwrote that stale
+    // value, and then compared localhost:3000 against the mocked localhost:63315. It depended on
+    // which ids earlier work happened to leave behind, so it passed locally and failed in CI.
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
-    vi.spyOn(theOneAudioRecorder as any, "urlPrefix").mockReturnValue(
+    vi.spyOn(AudioRecording.prototype as any, "urlPrefix").mockReturnValue(
         "http://localhost:63315/bloom/api/audio/wavFile?id=audio/",
     );
+
+    await initializeTalkingBookToolAsync();
 }
 
 export function StripPlayerSrcNoCacheSuffix(url: string): string {
--- a/src/BloomExe/Workspace/WorkspaceView.cs
+++ b/src/BloomExe/Workspace/WorkspaceView.cs
@@ -1557,6 +1557,11 @@ window.showWorkspaceInitializationFailure = function(message) {
         {
             if (Platform.IsLinux)
                 return;
+            // An automated run has nobody to dismiss a modal dialog. This one is shown as a startup
+            // action, so it would sit on the UI thread in its own message loop for the whole run --
+            // exactly the "dialog nobody can dismiss" that Program.RunningE2eTests exists to avoid.
+            if (Program.RunningE2eTests)
+                return;
             // If Bloom is newly installed or we only had old versions before, this should be 0.
             var isShown = Settings.Default.AutoUpdateDialogShown;
             if (isShown < kCurrentAutoUpdateVersion)


src/BloomBrowserUI/bookEdit/toolbox/talkingBook/audioRecordingSpec.ts
```diff
@@ -2941,16 +2941,25 @@ export async function setupForAudioRecordingTests() {
...
+    // it, and the call itself already sets the player's src. So a mock installed afterwards left
+    // that first src built by the REAL urlPrefix -- a relative URL, which jsdom resolves against
+    // its own base of http://localhost:3000/. Because setCurrentAudioId only refreshes the player
+    // when the audio id CHANGES, a test whose id was already current never overwrote that stale
+    // value, and then compared localhost:3000 against the mocked localhost:63315. It depended on
+    // which ids earlier work happened to leave behind, so it passed locally and failed in CI.
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
-    vi.spyOn(theOneAudioRecorder as any, "urlPrefix").mockReturnValue(
+    vi.spyOn(AudioRecording.prototype as any, "urlPrefix").mockReturnValue(
         "http://localhost:63315/bloom/api/audio/wavFile?id=audio/",
```
andrew-polk, line 2959: [Devin] **Investigate**: Prototype spy is removed mid-file by an existing restoreAllMocks call

The urlPrefix mock is installed once per file (beforeAll -> `setupForAudioRecordingTests`), but `src/BloomBrowserUI/bookEdit/toolbox/talkingBook/audioRecordingSpec.ts:1632` calls `vi.restoreAllMocks()` inside `runClearRecordingDeleteTest`, which un-installs it for every test that runs afterwards. This is pre-existing (the old instance-level spy was restored just the same), and it does not regress with this change, but it means the "the mock is always in place" invariant this fix relies on is only true until that test runs. If the CI flake persists, this is the next thing to look at: consider re-installing the urlPrefix mock in a `beforeEach`, or replacing the blanket `restoreAllMocks()` there with targeted restores.

andrew-polk: (Claude Opus 5) Assessed — accurate observation, but it does not affect this fix or any current test. Not acting on it.

The failing test is `talkingBookSpec.ts > showTool(... scenario=PreTextBox)`, and that file contains **zero** `restoreAllMocks`/`resetAllMocks` calls. Vitest isolates per test file (`isolate` is unset, so the default `true` applies, and the pool is `threads`), so the `restoreAllMocks()` at `audioRecordingSpec.ts:1632` cannot reach `talkingBookSpec.ts`. For the failure this PR fixes, the mock genuinely is always in place.

Within `audioRecordingSpec.ts` itself the restore does remove the prototype spy for later tests, exactly as Devin says — but nothing after line 1632 in that file asserts an audio URL (the only remaining occurrences are the setup helper's own comments and the mock literal). So there is no live consequence today, and as Devin notes it is pre-existing: the old instance-level spy was restored just the same.

Leaving the blanket `restoreAllMocks()` alone deliberately. Making it targeted, or reinstalling the mock in a `beforeEach`, would change mock lifetime for every test in a 3000-line spec file to harden an invariant nothing currently depends on — more risk than the latent weakness it removes. Devin's note is the right thing to check first if the flake somehow survives this change, and it is recorded here for exactly that.

andrew-polk:
(Claude Opus 5) Consulted Devin on 2026-08-03 (during preflight) up to commit `334ad92e73b693ff0c4cb04d5fce2439eb940b0e`.

Result: **0 bugs**, 1 Investigate flag, 2 Informational items.

- The Investigate flag — the `vi.restoreAllMocks()` at `audioRecordingSpec.ts:1632` un-installing the prototype spy for the rest of that file — is mirrored as its own thread, assessed, and resolved there: https://github.com/BloomBooks/BloomDesktop/pull/8138#discussion_r3706309001. Short version: it cannot reach `talkingBookSpec.ts`, where the failing test lives (vitest isolates per file), and nothing after that line in `audioRecordingSpec.ts` asserts an audio URL — so it is a pre-existing latent weakness, not a defect in this change.
- The 2 Informational items are not mirrored (low signal). Both agree with the PR description: the prototype mock now also covers ad-hoc `AudioRecording` instances (deliberate — no assertion depends on the real prefix), and the e2e early return skips the code that would record the dialog as shown (intended — suppress, not consume).

src/BloomBrowserUI/bookEdit/toolbox/talkingBook/audioRecordingSpec.ts
andrew-polk: (Claude Opus 5) **Developer decision: leave as is.**

Andrew reviewed this during preflight and chose not to act on it. No further rationale was recorded beyond the assessment above — that the blanket `restoreAllMocks()` cannot reach `talkingBookSpec.ts` where the failing test lives, that nothing after line 1632 in this file asserts an audio URL, and that the condition is pre-existing rather than introduced here.

So the blanket reset at `audioRecordingSpec.ts:1632` stays. If the nightly flake somehow survives this PR, this thread is the first place to look.

PR #8139, opened 2026-08-03
Fix toolbox not fully grayed out in Show Playback Order mode (BL-16630)

andrew-polk:
Ref: https://issues.bloomlibrary.org/youtrack/issue/BL-16630

## The bug

With **Show playback order buttons** on (Talking Book Tool → Advanced), the "Talking Book Tool" header goes dull except for a bright teal stripe at its right-hand end, and the other tool headers are dimmed too. Reported in 6.4 and 6.5; fine in 6.3.

## Root cause

The Talking Book tool's `#disablingOverlay` has always been a `position: fixed` element covering the whole toolbox viewport at `z-index: 1001` — accordion headers included. What kept the headers looking normal was this rule in `toolbox.less`:

```less
#toolbox h3 {
    ...
    z-index: 1005; // higher than disablingOverlay in Talking Book tool
}
```

When the toolbox root became React/MUI in 6.4, the headers turned into `AccordionSummary` elements and that rule stopped matching them, so they started being painted over by the overlay. The legacy `#toolbox` is now `display: none`, making the old rule dead code.

The bright stripe is the overlay's long-standing `width: calc(100% - 20px)` becoming visible for the first time: the 20px it leaves uncovered started landing on the teal active header instead of on panel-colored tool body. The gap was there in 6.3 too — it just never fell on anything colorful.

Pixel-sampling the two screenshots on the report confirms it. The overlay composites as 0.7 × `#2E2E2E` over what's beneath, and both shots were taken with the mode on:

| element | 6.3 | 6.4 | if under the overlay |
| --- | --- | --- | --- |
| "Talking Book Tool" header | `#1D94A4` | **`#294C51`** | `#294C51` |
| header text | `#FFFFFF` | **`#6C6B6B`** | `#6C6C6C` |
| "Canvas Tool" header | `#404040` | **`#333333`** | `#333333` |
| playback-order switch | `#F3AA18` | `#F3AA18` | ≈ `#6E5A2C` |
| "Help" text | `#F2F2F2` | `#FFFFFF` | ≈ `#6C6C6C` |

The header rows sit on the composited value in 6.4 and not in 6.3 — that's the regression. The switch and Help are undimmed in *both*, showing their own `z-index` opt-outs still work while the headers' no longer did.

## The fix

`ToolboxRoot.tsx` — put the headers back above the overlay with `position: relative; z-index: 1005`, and give the header a `background-color`. The background matters because a *collapsed* header paints none of its own: the colour you see is the `Accordion` root's, and the root can't be lifted along with the header, since a `z-index` there would create a stacking context around the `AccordionDetails` too and hoist the whole tool body out from under the overlay. Without it, lifting the header moved its text but left its background dimmed — which kept the stripe on every collapsed header. It uses the same colour the root does, so nothing changes visually.

`audioRecording.less` — narrows the strip the overlay leaves uncovered on the right from `20px` to `15px`, and names it. That strip is deliberate: it keeps the toolbox scrollbar undimmed and draggable while the tool is disabled, and on a short window that scrollbar can be the only way to reach the switch that turns the mode off — which is why closing the gap altogether (`width: 100%`) was tried during this work and rejected. `15px` is the scrollbar's measured width, and it is also `@toolboxContentPadding`, so when there's no scrollbar the uncovered strip now falls on the tool's right-hand padding instead of clipping ~5px off the ends of the full-width controls, where it used to leave them slightly undimmed and still clickable.

## Deliberately not changed

- **The `✕` isn't dimmed while the bar beside it is.** Different documents: the toolbox is an `<iframe id="toolbox">`, and the `✕` is the `pure-toggle-label` beside the drawer in the parent edit-tab page (`EditTabPane.tsx`). A fixed overlay inside the iframe can't reach out of it. Predates 6.4 — the same pattern is measurable in the 6.3 screenshot.

## Testing

CSS layering, so no unit-test coverage applies; no automated tests were run (skipped at the developer's request). Verified by hand in the running app. With **Show playback order buttons** on:

1. the "Talking Book Tool" header is one even teal colour edge to edge — no bright stripe at the right end;
2. the other tool headers look normal, not dimmed, and are still clickable;
3. turning the mode off restores the tool.

Some resolved review threads below discuss a `pointer-events` approach that was tried and reverted; they don't describe the current diff.

---
[Devin review](https://devinreview.com/BloomBooks/BloomDesktop/pull/8139)

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8139)

### DIFF
--- a/src/BloomBrowserUI/bookEdit/toolbox/ToolboxRoot.tsx
+++ b/src/BloomBrowserUI/bookEdit/toolbox/ToolboxRoot.tsx
@@ -76,6 +76,17 @@ const legacyToolSubPathByToolId: Record<string, string> = {
     settingsTool: "settings/Settings.html",
 };
 
+// While Show Playback Order mode is on, the Talking Book tool covers the whole toolbox with a
+// translucent overlay at this z-index.
+// cf @disablingOverlayZindex in bookEdit/toolbox/talkingBook/audioRecording.less
+const kDisablingOverlayZIndex = 1001;
+
+// The tool headers have to clear that overlay, and also the handful of controls inside the tool
+// that opt out of it at one above it (the Help link, and the Show Playback Order switch in
+// talkingBookAdvancedSection.tsx). The offset lands on 1005, which is what the pre-6.4
+// "#toolbox h3" rule in toolbox.less used, so this stays a like-for-like restoration.
+const kToolboxHeaderZIndex = kDisablingOverlayZIndex + 4;
+
 const toolboxHeaderIconStyles = css`
     width: 16px;
     height: 16px;
@@ -816,6 +827,23 @@ export const ToolboxRoot: React.FunctionComponent = () => {
                                     min-height: 32px;
                                     padding-left: 5px;
                                     padding-right: 12px;
+                                    // Keep the headers above the Talking Book tool's disabling
+                                    // overlay, so they neither look grayed out nor stop
+                                    // responding in Show Playback Order mode. Until 6.4 the
+                                    // "#toolbox h3" rule in toolbox.less did this; it stopped
+                                    // matching once the headers became MUI AccordionSummaries,
+                                    // which is BL-16630. Only works while no ancestor creates a
+                                    // stacking context -- a transform, filter, opacity or
+                                    // z-index on the Accordion, the Collapse or the tool-body
+                                    // host would trap it.
+                                    position: relative;
+                                    z-index: ${kToolboxHeaderZIndex};
+                                    // The header has to paint its own background for that to
+                                    // help. A collapsed header would otherwise be transparent
+                                    // and show the Accordion root's background, which stays
+                                    // under the overlay and so keeps being dimmed. Same colour
+                                    // the root uses, so nothing changes visually.
+                                    background-color: ${kBloomUnselectedTabBackground};
 
                                     & .MuiAccordionSummary-content {
                                         margin: 8px 0;
--- a/src/BloomBrowserUI/bookEdit/toolbox/talkingBook/audioRecording.less
+++ b/src/BloomBrowserUI/bookEdit/toolbox/talkingBook/audioRecording.less
@@ -8,7 +8,14 @@
 @buttonColumnWidth: 40px;
 @splitAnimationTime: 300ms;
 @textOnLightBackground: black;
-@disablingOverlayZindex: 1001; // higher than #audio-devlist
+// higher than #audio-devlist; cf kDisablingOverlayZIndex in bookEdit/toolbox/ToolboxRoot.tsx
+@disablingOverlayZindex: 1001;
+// Keeps #disablingOverlay clear of the toolbox scrollbar, so the scrollbar stays undimmed
+// and draggable while the tool is disabled. This is the scrollbar's actual width, which is
+// also @toolboxContentPadding -- so when there is no scrollbar, the strip the overlay
+// leaves uncovered falls on the tool's right-hand padding instead of clipping the ends of
+// the full-width controls, which used to leave them slightly undimmed and still clickable.
+@disablingOverlayScrollbarAllowance: 15px;
 
 // See BL-7442: When low line height causes lines to overlap, the highlighting overlaps and
 // can over up the bottom of the characters in the line above. For Chrome/Safari we have
@@ -496,6 +503,9 @@ body .cursor-progress {
     display: none;
 }
 
+// Dims and blocks clicks on the toolbox while in Show Playback Order mode.
+// Anything that must remain usable (the tool headers, the Help link, the Show Playback
+// Order switch itself) gives itself a z-index above @disablingOverlayZindex.
 #disablingOverlay {
     z-index: @disablingOverlayZindex;
     opacity: 0.7;
@@ -503,5 +513,5 @@ body .cursor-progress {
     top: inherit;
     background-color: @toolboxBackgroundColor;
     height: 100%;
-    width: calc(100% - 20px);
+    width: calc(100% - @disablingOverlayScrollbarAllowance);
 }


src/BloomBrowserUI/bookEdit/toolbox/talkingBook/audioRecording.less
```diff
@@ -496,12 +496,18 @@ body .cursor-progress {
     display: none;
 }
 
+// Dims and blocks clicks on the whole toolbox while in Show Playback Order mode.
+// Anything that must remain usable (the tool headers, the Help link, the Show Playback
+// Order switch itself) gives itself a z-index above @disablingOverlayZindex.
```
andrew-polk: [Devin] **Investigate**: New comment claims the Help link stays above the overlay, but its z-index has no effect

The added comment says "Anything that must remain usable (the tool headers, the Help link, the Show Playback Order switch itself) gives itself a z-index above @disablingOverlayZindex." That is true for the playback-order switch (`#audio-playbackOrderControl-wrapper` gets an inline `position: relative` from the pug mixin in `src/BloomBrowserUI/bookEdit/toolbox/talkingBook/talkingBookToolboxTool.pug:9`), but `.help-link` (`src/BloomBrowserUI/bookEdit/toolbox/talkingBook/audioRecording.less:348-349`) is a statically positioned `<a>`, so its `z-index` is ignored and it remains under the overlay (dimmed and unclickable). Not introduced by this PR, but the new comment documents behavior that doesn't actually hold; consider adding `position: relative` to `.help-link` or wording the comment to match reality.

src/BloomBrowserUI/bookEdit/toolbox/ToolboxRoot.tsx
```diff
@@ -816,6 +816,15 @@ export const ToolboxRoot: React.FunctionComponent = () => {
...
                                     padding-right: 12px;
+                                    // The Talking Book tool covers the whole toolbox with a
+                                    // translucent #disablingOverlay (z-index 1001) while in
+                                    // Show Playback Order mode. The tool headers must stay above
+                                    // it, both so they don't look grayed out and so the user can
+                                    // still switch tools (BL-16630). Until 6.4 this was done by
+                                    // the "#toolbox h3" rule in toolbox.less, which no longer
+                                    // applies now that the headers are MUI AccordionSummaries.
+                                    position: relative;
+                                    z-index: 1005;
```
andrew-polk: [Devin] **Investigate**: Headers above the overlay stay clickable, including the Talking Book header itself

Lifting every `AccordionSummary` to z-index 1005 makes all headers clickable while Show Playback Order mode is active, including the Talking Book header, so the user can collapse the Talking Book tool while the mode is on. That should be benign because the overlay lives inside the Talking Book body and MUI's Collapse hides collapsed content (`visibility: hidden`), which also hides the fixed-position overlay descendant — but it is worth confirming by hand that leaving/re-entering the tool while in playback-order mode restores normal state (the mode flag lives in `audioRecording.ts`).

src/BloomBrowserUI/bookEdit/toolbox/talkingBook/audioRecording.less
andrew-polk: [Claude Opus 5] Not an issue — the Help link genuinely does stay above the overlay, and the comment is accurate.

`z-index` is not ignored here. `.help-link` is a **flex item**: its parent `.ui-audioBody` is `display: flex; flex-direction: column` (`audioRecording.less:108-110`), and per the Flexbox spec a flex item with a non-`auto` `z-index` creates a stacking context and paints accordingly **even when `position` is `static`**. The `margin-top: auto` right next to the `z-index` — the thing that pushes Help to the bottom of the panel — only works on a flex item, so it's evidence of the same fact.

The screenshot attached to BL-16630 confirms it directly. That shot was taken with Show Playback Order mode **on**, so the overlay was showing. In it:

- the "Help" text measures `#FFFFFF` — undimmed;
- white text that *is* under the overlay composites to `0.3 x 255 + 0.7 x 46 = ~#6C6C6C`, and the dimmed "Talking Book Tool" header text in that same image measures `#6C6B6B` — exactly that value.

So Help was never dimmed or unclickable, and adding `position: relative` to it would be a no-op. Resolving.

src/BloomBrowserUI/bookEdit/toolbox/ToolboxRoot.tsx
andrew-polk: [Claude Opus 5] Confirmed the part you asked about, and it's handled by design — but there's a small pre-existing residual, so I'm leaving this thread open for the developer rather than resolving it.

**Switching tools while the mode is on is explicitly handled.** `TalkingBookModel.detachFromPage()` calls `removeRecordingSetup()` (`audioRecording.ts:582`), which checks `inShowPlaybackOrderMode` and calls `removePlaybackOrderUi(page, true)`. That hides the overlay and strips the on-page number buttons, while deliberately leaving the switch checked so the mode comes back the next time the tool is opened. There's a comment there saying exactly that. So leaving and re-entering the tool restores normal state.

**This PR doesn't newly enable the scenario.** The very defect being fixed here is the escape hatch: the undimmed 20px strip at the right end of the header is *inside* the header and is clickable today, so a user can already collapse or switch away from the Talking Book tool while in playback-order mode in 6.4. This change makes the header uniformly clickable instead of only via a 20px secret.

**The residual, which is pre-existing and applies to every tool.** Collapsing an accordion to *nothing* (clicking the open header rather than opening a different tool) doesn't fire `ToolboxRoot`'s `activeToolChangedCallbacks` — `onChange` sets `expandedSectionId` to `undefined` and notifies nobody — so legacy code never hears "deactivated". In playback-order mode that means the on-page number buttons stay up until you re-open the tool or pick another one. The overlay itself does disappear, as you predicted: `visibility` is inherited and applies to `position: fixed` descendants too. This gap arrived with the 6.4 React toolbox generally (jQuery UI's accordion defaulted to `collapsible: false`, so 6.3 could not collapse-to-nothing at all), and closing it means changing shared toolbox behavior — out of scope for a regression fix on a release branch.

Flagging that residual for the developer as a possible follow-up card.

andrew-polk:
[Claude Opus 5 from Andrew Polk's machine during preflight] Consulted Devin on 2026-08-03 10:44 USMST up to commit `624bdbc7f35f93f1689011aae6178ec278345334`.

Devin found no Bugs and 2 Investigate flags, no Informational items. Both flags are mirrored above as their own threads:

- **Help link's z-index has no effect** — not an issue; refuted with the numbers from the bug report's own screenshot, thread resolved.
- **Headers stay clickable, including the Talking Book header** — the part Devin asked about is handled by design, but it surfaced a small pre-existing gap in the React toolbox. Thread left **open** for Andrew to decide whether it deserves a follow-up card.

CI is green (`pr-automation`, `track / set-waiting`). No other review bots are configured on this repo.

src/BloomBrowserUI/bookEdit/toolbox/ToolboxRoot.tsx
andrew-polk: [Claude Opus 5] Developer decision: leave as is — and the premise doesn't hold up in practice.

Andrew tried it in the running app and reports that the residual I described doesn't actually happen: collapsing the Talking Book panel while Show Playback Order mode is on does not leave the numbered buttons stranded on the page. So the concern behind this flag is moot, and there's nothing to change and no follow-up card to file.

The part you asked to have confirmed by hand is confirmed: leaving and re-entering the tool while in playback-order mode restores normal state.

Resolving.

src/BloomBrowserUI/bookEdit/toolbox/talkingBook/audioRecording.less
```diff
@@ -496,12 +496,39 @@ body .cursor-progress {
...
     height: 100%;
-    width: calc(100% - 20px);
+    width: 100%;
+    // Dim only, never intercept. Because this is a full-width fixed sheet, swallowing
+    // clicks here would also swallow the toolbox scrollbar, and on a short window the
+    // user could then be unable to scroll down to the one switch that turns this mode
+    // off. The click-blocking is done by the tool body below instead, which leaves the
+    // scrollbar (owned by an ancestor) alone.
+    pointer-events: none;
+}
```
andrew-polk: [Devin] **Investigate**: Overlay no longer blocks clicks anywhere outside the Talking Book tool body

Adding `pointer-events: none` to `#disablingOverlay` and moving the click-blocking onto `#disablingOverlay:not(.hiddenOverlay) ~ .ui-audioBody` narrows the disabled area from "the whole toolbox viewport" (the overlay is `position: fixed` and now `width: 100%`) to just the Talking Book tool body. Everything else the overlay merely dims — most notably the toolbox's `✕` close button in the dark band above the headers, and any toolbox chrome outside the accordions — is now clickable while looking grayed out. Tool headers becoming clickable is the stated intent; the close button being simultaneously dimmed and live is an inconsistency worth a quick manual check (`src/BloomBrowserUI/bookEdit/toolbox/talkingBook/audioRecording.less:499-519`). Functionally it should be safe: closing the toolbox or switching tools routes through `removeRecordingSetup()` → `removePlaybackOrderUi(page, true)` → `setDisableEverythingMode(false)` (`src/BloomBrowserUI/bookEdit/toolbox/talkingBook/audioRecording.ts:585-591`, `2273`), so the page-side playback-order UI is cleaned up.

src/BloomBrowserUI/bookEdit/toolbox/talkingBook/audioRecording.less
```diff
@@ -496,12 +496,39 @@ body .cursor-progress {
...
+// to a descendant.
+#disablingOverlay:not(.hiddenOverlay) ~ .ui-audioBody {
+    pointer-events: none;
+
+    // ...except the controls we deliberately leave live while the rest is disabled.
+    // (The Show Playback Order switch does the same thing for itself where it is
+    // defined, in talkingBookAdvancedSection.tsx.)
+    .help-link {
+        pointer-events: auto;
+    }
```
andrew-polk: [Devin] **Investigate**: Pre-existing: user can be stuck in Show Playback Order mode if the Advanced section is collapsed

When returning to a page with the mode still checked, `handleNewPageReady` re-shows the playback-order UI and the overlay (`src/BloomBrowserUI/bookEdit/toolbox/talkingBook/audioRecording.ts:2674-2679`). The only control that opts back in to clicks is the switch itself; the `TriangleCollapse` header for the Advanced section does not. If that section is collapsed at that moment, the switch is not on screen and the collapse header cannot be clicked, leaving no way to exit the mode from the toolbox. This is pre-existing (the old overlay intercepted those clicks too) and not introduced here, but the PR's pointer-events rework is the natural place to fix it by also exempting the Advanced section header.

andrew-polk: [Claude Opus 5] Good catch — this is real and reachable, and it's fixed in `f0d3f3b`.

Verified the whole chain before acting: `TriangleCollapse` is used here with `initiallyOpen={false}`, and `inShowPlaybackOrderMode` survives a page or tool change because `removePlaybackOrderUi()` is called with `leaveChecked=true` and never clears the flag. So `handleNewPageReady()` re-shows the overlay on return with the switch sitting inside a collapsed section whose header was inheriting `pointer-events: none`. No way out of the mode from the toolbox at all.

You're right that it isn't a regression from this branch — the old click-swallowing overlay covered that header too — but it is exactly the failure mode the previous commit was written to prevent, so it belongs here rather than in a follow-up.

The fix routes through `TriangleCollapse`'s existing `extraButtonCss` hook, which targets only the clickable triangle/label. The section's *contents* keep inheriting `pointer-events: none`, so Insert Segment Marker, Recording Mode, and Import Recording stay properly disabled while the header becomes expandable. Resolving.

src/BloomBrowserUI/bookEdit/toolbox/talkingBook/audioRecording.less
andrew-polk: [Claude Opus 5] Accurate description, and the cleanup path you traced is correct — I'm leaving this open as an FYI for the developer rather than resolving it, since it's a small user-visible behavior change and he's testing the build now.

Assessment: the ✕ going from inert to live is, on balance, an improvement. Under the old sheet a user in Show Playback Order mode could not close the toolbox at all, which is its own small trap. And as you note, closing routes through `removeRecordingSetup()` → `removePlaybackOrderUi(page, true)`, so the page-side UI is cleaned up properly — I confirmed that path.

What's left is purely cosmetic: the dark band above the headers is dimmed but its button is live. Worth knowing that band was dimmed in 6.3 too — sampling the reporter's 6.3 screenshot gives `#282828` there against `#1A1A1A` outside the sheet's reach — so the appearance is unchanged from before the regression; only the clickability changed. Narrowing the *dimming* to the tool body as well would be the tidier end state, but it would alter how the panel has looked for several releases, which is more than this regression fix should take on.

src/BloomBrowserUI/bookEdit/toolbox/ToolboxRoot.tsx
andrew-polk: [Claude Opus 5] Adding the mechanism here for the record, because Devin re-raised this concern on a later commit under a different title ("Making all tool headers clickable lets the user collapse the Talking Book section while playback-order mode is still on") with a sharper reading of the code — and it turns out both Devin and Andrew were right about different things.

Devin's code path is accurate as far as it goes: collapsing an expanded section calls `onChange` with `expanded === false`, which only does `setExpandedSectionId(undefined)` and skips the `activeToolChangedCallbacks` notification, so `removeRecordingSetup()` never runs.

What it missed is that the collapse **self-reverts**. `ToolboxRoot.tsx:644-653` is an effect keyed on `expandedSectionId`: whenever that becomes falsy it reads the legacy current tool id — still `talkingBook`, precisely because no `activeToolChangedCallbacks` fired to change it — and immediately calls `setExpandedSectionId(legacyCurrentToolId)` again. So the panel snaps straight back open and the stranded state is unreachable.

That is exactly what Andrew observed when he tried it: nothing appears to happen, and the numbered buttons are never left orphaned. The skipped notification that looks like the bug is the same thing that makes the section re-expand.

Leaving this resolved; no code change needed.

andrew-polk:
[Claude Opus 5 from Andrew Polk's machine during preflight] Consulted Devin on 2026-08-03 11:24 USMST up to commit `f0d3f3b7cee0841bb19789ac14909b3500d59e15`.

Two further rounds since the log above, one per pushed commit. Both came back with **0 Bugs**.

**Round 2** (`3c3c26c`, the pointer-events rework) — 3 Investigate, 3 Informational. It found the one thing that mattered: with the tool body switching pointer events off, the collapsed **Advanced** section's header went dead too, and since that section starts collapsed and `inShowPlaybackOrderMode` survives a page change, a user could return to a page with no way at all to reach the switch that turns the mode off. Fixed in `f0d3f3b`; thread resolved. It also noted the toolbox `✕` is now dimmed-but-clickable — that thread is **open** for Andrew, with a recommendation to leave it.

**Round 3** (`f0d3f3b`) — 3 Investigate, 4 Informational, nothing new to act on:

- The "stuck if Advanced is collapsed" flag reappeared with byte-identical text, but it is stale: the same job's own informational finding verifies that `pointer-events: auto` on "the TriangleCollapse `Button`" correctly re-enables its descendants — an opt-in that only exists in this commit. Confirmed present at this sha; nothing to do.
- The headers-clickable concern was re-raised under a new title with a sharper code reading. Worth it: chasing it down showed the collapse **self-reverts** (`ToolboxRoot.tsx:644-653` re-expands the section the moment `expandedSectionId` goes falsy), so the state it describes is unreachable — which is why Andrew saw nothing happen when he tested it. Written up on the original thread.
- One informational item is Devin verifying the new rules' DOM and inheritance assumptions and agreeing they hold. Another repeats its earlier claim that the Help link's `z-index` is inert; that remains refuted by the reporter's own screenshot (Help renders `#FFFFFF` there while text genuinely under the sheet renders `#6C6B6B`), and it cites `#audio-playbackOrderControl-wrapper`, which is dead CSS — its pug mixin is never invoked. The last one, that the header `z-index` only holds while no ancestor creates a stacking context, was worth keeping: it is now written down as a comment beside the rule in `f0d3f3b`.

CI green on every commit (`pr-automation`, `track / set-waiting`). No other review bots are configured on this repo. Automated tests were skipped for this branch at Andrew's request.

src/BloomBrowserUI/bookEdit/toolbox/talkingBook/audioRecording.less
andrew-polk: [Claude Opus 5] Correction — this finding's premise is false, and so was my earlier reply accepting it. Moot now anyway: the change it describes has been reverted in `544859c`.

**The overlay never blocked the ✕, in any version.** The toolbox is an `<iframe id="toolbox">`, and the ✕ is the `pure-toggle-label` that sits *beside* the drawer in the parent edit-tab page (`src/BloomBrowserUI/app/EditTabPane.tsx:161-170`). A `position: fixed` element inside the iframe is bounded by the iframe's own viewport — it cannot dim or intercept anything in the parent document. So "the close button being simultaneously dimmed and live" was not introduced by `pointer-events: none`; the button was always live.

What *is* dimmed is the bar to the ✕'s left, which is the top of the iframe and genuinely under the overlay. That dimmed-bar-next-to-undimmed-✕ asymmetry is inherent to the iframe boundary and predates 6.4 — the same pattern is measurable in the 6.3 screenshot attached to BL-16630 (`#282828` for the bar, undimmed at the ✕).

I should have checked the document boundary before endorsing this. Andrew spotted the asymmetry in the build and asking "why?" is what turned it up.

Resolving: the `pointer-events` split has been reverted, so there is nothing left here to weigh.

src/BloomBrowserUI/bookEdit/toolbox/talkingBook/audioRecording.less
andrew-polk: [Claude Opus 5] Following up: the fix for this has been reverted in `544859c`, because the scenario turns out not to be reachable and the rule that made it matter is gone.

Two things:

1. **The Advanced section cannot collapse on its own.** `TalkingBookAdvancedSection` is mounted with `ReactDOM.render` into a stable container (`audioRecording.ts:4571-4599`), so repeated renders *reconcile* rather than remount — `useState(props.initiallyOpen)` keeps whatever value it has. The container node itself survives too, since `LiveToolBodyHost` moves the legacy tool root rather than recreating it. So the section is only ever collapsed because the user collapsed it, not on returning to a page. Andrew called this correctly from experience.

2. **The rule that made a collapsed section a trap is gone.** The header was only unclickable because the previous commit put `pointer-events: none` on the tool body. That whole mechanism has been reverted — the overlay is back to keeping the scrollbar clear by geometry instead — so the Advanced header is clickable again for the ordinary reason that nothing is disabling it.

Which leaves the original observation as correct-but-inert: under the old click-swallowing overlay a *user-collapsed* Advanced section did put the switch out of reach. That is unchanged from 6.3 and every release before it, and it is not what BL-16630 is about.

Staying resolved.

src/BloomBrowserUI/bookEdit/toolbox/talkingBook/audioRecording.less
```diff
@@ -496,12 +500,23 @@ body .cursor-progress {
     display: none;
 }
 
+// Dims and blocks clicks on the toolbox while in Show Playback Order mode.
+// Anything that must remain usable (the tool headers, the Help link, the Show Playback
+// Order switch itself) gives itself a z-index above @disablingOverlayZindex.
```
andrew-polk, line 508: [Devin] **Investigate**: New LESS comment claims the Help link and Show Playback Order switch opt out of the overlay via z-index, but both appear to be statically positioned

The comment above `#disablingOverlay` says "Anything that must remain usable (the tool headers, the Help link, the Show Playback Order switch itself) gives itself a z-index above `@disablingOverlayZindex`." `z-index` has no effect on a statically-positioned element, and neither `.help-link` (`audioRecording.less:348-349`) nor the `BloomTooltip` wrapper carrying `z-index: 1002` in `talkingBookAdvancedSection.tsx` sets `position`. If both are static, their `z-index` declarations are inert and the comment documents a mechanism that isn't in force — the controls would be under the overlay rather than above it.

andrew-polk: [Claude Opus 5] Not an issue — the comment is accurate. This is the third time this has come up across the review rounds, so here is the mechanism for both controls plus the measurement, to put it to rest.

**`z-index` does apply to a static element when that element is a flex item.** Per the Flexbox spec, flex items paint like inline blocks and a non-`auto` `z-index` creates a stacking context *regardless of `position`*. Both controls are flex items:

- **Help link** — `.help-link` is a child of `.ui-audioBody`, which is `display: flex; flex-direction: column` (`audioRecording.less:112-114`). The `margin-top: auto` sitting right beside its `z-index` — the thing that pushes Help to the bottom of the panel — only works on a flex item, so it corroborates from the same two lines.
- **Show Playback Order switch** — its `BloomTooltip` wrapper is a child of `.MuiCollapse-wrapperInner`, which `TriangleCollapse` explicitly sets to `display: flex; flex-direction: column` (`TriangleCollapse.tsx:93-100`).

**And the measurement, from the reporter's own screenshots.** Both were taken with Show Playback Order mode *on*, so the overlay was showing. The overlay composites as 0.7 × `#2E2E2E` over what's beneath, so anything genuinely under it is visibly muted:

| element | 6.3 (no bug) | 6.4 (bug present) | if it were under the overlay |
| --- | --- | --- | --- |
| Show Playback Order switch | `#F3AA18` | `#F3AA18` | ≈ `#6E5A2C` |
| "Help" text | `#F2F2F2` | `#FFFFFF` | ≈ `#6C6C6C` |
| "Talking Book Tool" header text | `#FFFFFF` | `#6C6B6B` | ≈ `#6C6C6C` |

The switch measures *identically* in both versions, and Help is undimmed in both — while the header text in 6.4 sits at exactly the composited value, which is the bug this PR fixes. So the two opt-outs demonstrably work and the header's demonstrably didn't.

One incidental correction: earlier rounds cited `#audio-playbackOrderControl-wrapper` as the switch's element. That selector is dead CSS — the pug mixin that would have produced it (`checkboxWithHiddenClickHandler`) is defined but never invoked, since the Advanced section is React now. The live element is the `BloomTooltip` wrapper described above.

Resolving.

andrew-polk:
[Claude Opus 5 from Andrew Polk's machine during preflight] Consulted Devin on 2026-08-03 11:48 USMST up to commit `544859c42f93dcb26cc021b8b70125c50c3f05ad`.

Fourth and final round, on the commit that reverts the overlay-widening detour. **0 Bugs.** 4 Investigate, 5 Informational — but only one Investigate item describes code that still exists at this sha:

- **New:** it challenged the LESS comment's claim that the Help link and the Show Playback Order switch opt out of the overlay via `z-index`, on the grounds that both are statically positioned. Refuted and resolved: `z-index` does apply to a static element when it is a **flex item**, and both are (`.ui-audioBody` and `.MuiCollapse-wrapperInner` are each `display: flex; flex-direction: column`). Measured from the reporter's screenshots, the switch reads `#F3AA18` in *both* 6.3 and 6.4 and Help is undimmed in both, while the header text in 6.4 sits at exactly the composited `#6C6B6B` — so the two opt-outs work and the header's didn't. That thread carries the full table.
- **Stale, not re-posted:** three Investigate items and two Informational items describe the `pointer-events` split and the Advanced-header exemption, both of which this commit removed. Devin appears to carry findings forward across a PR's history rather than scoping strictly to the current diff, so its counts need reading against the current tree rather than taken at face value.

One Informational item is worth passing on even though it is pre-existing: with the overlay stopping 20px short and the tool body padded by 15px, a **5px sliver at the right edge of the full-width controls** (Insert Segment Marker, Import Recording) is left uncovered and therefore still clickable while the tool is disabled. That has been true since long before 6.4. Narrowing the allowance to 15px would close it exactly, since it would then coincide with the padding — but whether a real WebView2 scrollbar fits in 15px wants checking in the running app first, so it is going to Andrew as a decision rather than being changed on a guess.

All Devin threads on this PR are now resolved. CI green on every commit. No other review bots are configured on this repo. Automated tests were skipped for this branch at Andrew's request.

src/BloomBrowserUI/bookEdit/toolbox/talkingBook/audioRecording.less
```diff
@@ -9,6 +9,12 @@
 @splitAnimationTime: 300ms;
 @textOnLightBackground: black;
 @disablingOverlayZindex: 1001; // higher than #audio-devlist
+// Keeps #disablingOverlay clear of the toolbox scrollbar, so the scrollbar stays undimmed
+// and draggable while the tool is disabled. This is the scrollbar's actual width, which is
+// also @toolboxContentPadding -- so when there is no scrollbar, the strip the overlay
+// leaves uncovered falls on the tool's right-hand padding instead of clipping the ends of
+// the full-width controls, which used to leave them slightly undimmed and still clickable.
+@disablingOverlayScrollbarAllowance: 15px;
```
andrew-polk, line 18: [Devin] **Bug**: Uncovered strip beside the disabled tool shrinks by 5 pixels, contrary to the stated no-behavior-change

The uncovered strip left beside the dimming overlay is narrowed from 20px to 15px (`@disablingOverlayScrollbarAllowance: 15px`), even though the change was described as naming-only, so whenever the toolbox scrollbar is wider than 15 pixels part of it becomes dimmed and unusable while the tool is disabled.

Impact: on setups where the scrollbar is wider than 15 pixels (e.g. increased UI scaling), users may not be able to scroll the toolbox to reach the switch that turns Show Playback Order mode off.

The old rule was `width: calc(100% - 20px)`. The new rule is `width: calc(100% - @disablingOverlayScrollbarAllowance)` with the constant set to `15px`. That is a real 5px behavioral change, not a rename: the overlay now extends 5px further right. The comment justifies 15px as "the scrollbar's actual width, which is also `@toolboxContentPadding`", which holds only for a default 15px Chromium scrollbar at 100% scaling. The PR description explicitly states this file has "no behavior change" and that the 5px sliver mismatch is deliberately not changed, so either the constant should stay at 20px or the description/comment should be corrected.

andrew-polk: [Claude Opus 5] Half right, and the half that's right is a good catch — thank you. Splitting the two claims:

**The documentation was stale. Fixed.** The PR description did still say this file had "no behavior change" and did still list the 5px sliver as deliberately unchanged. Both were true of an earlier revision and I updated the commit message without updating the description to match. It now describes the 20px → 15px narrowing as the intentional change it is, and the sliver bullet is gone because the sliver is what the change closes. Precisely the sort of drift worth flagging.

**The 5px is intended, and it is the developer's measured call.** Andrew has the build in front of him and asked for exactly this: match the allowance to the real scrollbar width, and in doing so close the sliver. It does both jobs at once, because 15px is simultaneously the scrollbar's width and `@toolboxContentPadding` — so the uncovered strip now lands on the tool body's right-hand padding rather than clipping the ends of Insert Segment Marker and Import Recording, which is what left them slightly undimmed and still clickable while the tool was supposed to be disabled. Compiling the LESS confirms `calc(100% - 15px)` against a body `padding: 15px`.

**The impact claim doesn't hold, though.** Increased Windows UI scaling doesn't widen the scrollbar in CSS pixels — Chromium scales the whole layout, so a 15px scrollbar stays 15 CSS px at 125% or 150%. A wider bar would need a non-default Windows scrollbar setting or a future WebView2 change. And even then the outcome would be 5px of a 15px bar dimmed, with the other 10px still draggable and mouse-wheel scrolling unaffected — not "unable to scroll".

Not taking the suggested edit, which would revert to 20px and reintroduce the sliver. Resolving.

andrew-polk:
[Claude Opus 5 from Andrew Polk's machine during preflight] Consulted Devin on 2026-08-03 13:29 USMST up to commit `1b8247c7b5a2bdcd55350bd149e63d0108ec3da6`.

Final round, after Andrew re-tested and the fix was completed. **1 Bug, 5 Investigate, 9 Informational** — one item genuinely new and worth acting on, the rest carried forward from superseded revisions.

- **Bug: the uncovered strip narrowed 20px -> 15px "contrary to the stated no-behavior-change".** Half right, and the right half mattered: the **PR description** still claimed this file had no behavior change and still listed the 5px sliver as deliberately unchanged. Both were true of an earlier revision; I had updated the commit message and not the description. Corrected. The 5px itself is intentional — Andrew asked for the allowance to match the measured scrollbar width, which simultaneously closes the sliver, since 15px is also `@toolboxContentPadding`. Devin's impact reasoning does not hold: Windows UI scaling does not widen a scrollbar in CSS pixels, and even a wider bar would leave 10px of 15px draggable with the wheel unaffected. Suggested edit (revert to 20px) not taken; thread resolved with the reasoning.
- The five Investigate items are all previously-handled or describe reverted code — the `pointer-events` split, the Advanced-header exemption, the headers-clickable question Andrew settled from experience, and the flex-item `z-index` claim refuted with measurements. Each already has a resolved thread. One is new only in framing: it notes the header `z-index` restores pre-6.4 clickability, which is the intent.

Devin carries findings forward across a PR's history rather than scoping to the current diff, so across the six rounds its counts needed reading against the current tree. Its two real contributions were the collapsed-Advanced trap (on a revision since reverted) and this documentation drift.

All threads on this PR are resolved. CI green. No other review bots configured. Automated tests skipped for this branch at Andrew's request; Andrew verified the behavior in the running app.

StephenMcConnel (changes requested):
> @StephenMcConnel reviewed 3 files and all commit messages, made 1 comment, and resolved 6 discussions.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8139)* status: all files reviewed, 1 unresolved discussion (waiting on andrew-polk).

___
*[`src/BloomBrowserUI/bookEdit/toolbox/ToolboxRoot.tsx` line 829 at r4](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8139#-Oz8rA1F-BSYvuZ2Yiew:-Oz8rA1F-BSYvuZ2Yiex:b-bymkij) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/1b8247c7b5a2bdcd55350bd149e63d0108ec3da6/src/BloomBrowserUI/bookEdit/toolbox/ToolboxRoot.tsx#L829)):*
> ```XML
>                                     // the Collapse or the tool-body host would trap it.
>                                     position: relative;
>                                     z-index: 1005;
> ```

Would the 1005 be better expressed as a variable, possibly in relation to @disablingOverlayZindex ?

andrew-polk (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8139)* status: 1 of 3 files reviewed, 1 unresolved discussion (waiting on StephenMcConnel).

___
*[`src/BloomBrowserUI/bookEdit/toolbox/ToolboxRoot.tsx` line 829 at r4](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8139#-Oz8rA1F-BSYvuZ2Yiew:-Oz8wwDt6-c_j3Okvr02:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/1b8247c7b5a2bdcd55350bd149e63d0108ec3da6/src/BloomBrowserUI/bookEdit/toolbox/ToolboxRoot.tsx#L829)):*

Would the 1005 be better expressed as a variable, possibly in relation to @disablingOverlayZindex ?
</blockquote></details>

Done.

StephenMcConnel (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8139)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on andrew-polk).

PR #8140, opened 2026-08-03
Report clipboard failures the user's own copy and paste can hit (BL-16459)

JohnThomson:
The earlier fix for BL-16459 made the `common/clipboardText` endpoint toast instead of popping a problem-report dialog. That covered a handful of buttons, but not the ways a user actually copies and pastes: a plain **Ctrl+C, Ctrl+X or Ctrl+V**, the canvas element's **Copy Text** command, and the **Cut/Copy toolbar buttons** all failed in complete silence. Worse:

- **Ctrl+X deleted the selected text it had failed to copy**, losing it outright.
- A locked clipboard was reported to the user as *"Bloom did not find an image on your clipboard. Copy one first"* — advice about a clipboard that may be perfectly full.

## Two measured facts shaped the fix

Both are counter-intuitive enough that they are recorded in `BloomClipboard`'s comments, so they don't get "corrected" back out.

**1. Chromium never tells Javascript that a clipboard operation failed.** With another process holding the clipboard:

| call | clipboard free | clipboard locked |
|---|---|---|
| `navigator.clipboard.writeText()` | resolves | **resolves** |
| `document.execCommand("copy")` | `true` | **`true`** |
| `navigator.clipboard.readText()` | resolves with text | **resolves with `""`** |

So adding `.catch()` to the front-end calls would accomplish nothing — the work has to go through C#.

**2. A .NET clipboard *read* inside Bloom does not fail either.** OLE answers it from a data object cached by that process's own earlier read, and Bloom is always reading (`UpdateEditButtonsAsync` polls the clipboard on a timer). Measured on .NET 8 with the clipboard locked:

| | process that recently read the clipboard | fresh process |
|---|---|---|
| `ContainsText()` | **returned True** | threw `ExternalException` |
| `GetText()` | **returned `""`** | threw `ExternalException` |
| `SetText()` | threw | threw |

Only writes fail honestly. That asymmetry is exactly why the Copy button reported failures and Ctrl+C did not, and why a failed paste looked to the user like an empty clipboard.

## The change

- **`BloomClipboard`** (new) owns clipboard text access: write failures detected by catching, anything read-shaped by asking the OS directly via `OpenClipboard`. Marshals to the UI thread, so API handlers can call it. Reports via a toast (with Report link + Sentry), deduped so one keystroke logs once.
- **`hostClipboard.ts`** (new) routes the front end's clipboard work through C# instead of `navigator.clipboard`. Nothing is lost — those calls were plain-text only.
- **Keyboard shortcuts are checked from `keydown`**, because CKEditor `preventDefault`s copy and cut inside a `.bloom-editable` and swallows paste before it reaches the document. There is deliberately *no* "skip defaultPrevented events" check; a regression test guards that, since adding one silently disables the whole feature.
- **Ctrl+X is left to the browser**, and only checked afterwards. An earlier revision of this branch took the cut over so it could copy first and not delete on failure — but Bloom's own cut can only put *plain* text on the clipboard, so that lost formatting, links and inline pictures on **every** cut rather than only failing ones. So a failed cut still loses the text (Chromium's long-standing behaviour); what is new is that the user is told, and can press Ctrl+Z. Fixing it properly needs the copy path to carry several clipboard formats at once — see the open discussion. AltGr (reported as Ctrl+Alt) is excluded so international layouts can still type.
- Brought the other unguarded C# clipboard writes in: reader stats, BloomPub text-to-clipboard, and copy-page-HTML (an `async void` that a clipboard exception would have crashed the program from).

## Verification

Against a **genuinely locked clipboard** in a running Bloom (verified with a self-testing locker, after discovering that `OpenClipboard(NULL)` reports success while granting no exclusivity at all):

- Ctrl+C → "Bloom was not able to copy that."
- Ctrl+V → "Bloom was not able to paste."
- Ctrl+X → toast (the text is removed by the browser as always; Ctrl+Z restores it)
- Copy/Cut/Paste buttons → toast
- the false "no image on your clipboard" message → gone

597 front-end tests pass (19 new, covering the endpoints, the cut-abort contract, and the defaultPrevented regression). `BloomExe`/`BloomTests` build clean; typecheck, eslint, prettier, csharpier clean.

Ref: https://issues.bloomlibrary.org/youtrack/issue/BL-16459

[Devin review](https://devinreview.com/BloomBooks/BloomDesktop/pull/8140)

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8140)

### DIFF
--- a/src/BloomBrowserUI/bookEdit/bloomField/hyperlinks.ts
+++ b/src/BloomBrowserUI/bookEdit/bloomField/hyperlinks.ts
@@ -33,7 +33,10 @@ export function tryProcessHyperlink(
 export function getHyperlinkFromClipboard(
     callback: (url: string) => void,
 ): void {
-    get("common/clipboardText", (result) => {
+    // The availability-check read: this runs to decide whether the paste-hyperlink button should
+    // be enabled -- including during page initialization -- not because the user asked to paste,
+    // so C# must not report a momentarily unreadable clipboard as a failed paste.
+    get("common/clipboardText?checkingAvailability=true", (result) => {
         if (!result.data) {
             callback("");
         }
--- a/src/BloomBrowserUI/bookEdit/bloomField/hyperlinks.ts
+++ b/src/BloomBrowserUI/bookEdit/bloomField/hyperlinks.ts
@@ -1,4 +1,5 @@
 import { get } from "../../utils/bloomApi";
+import { readClipboardTextForAvailabilityCheck } from "../js/hostClipboard";
 
 // returns empty string if the URL is not valid.
 // returns the url as is if it is external
@@ -33,18 +34,23 @@ export function tryProcessHyperlink(
 export function getHyperlinkFromClipboard(
     callback: (url: string) => void,
 ): void {
-    // The availability-check read: this runs to decide whether the paste-hyperlink button should
-    // be enabled -- including during page initialization -- not because the user asked to paste,
-    // so C# must not report a momentarily unreadable clipboard as a failed paste.
-    get("common/clipboardText?checkingAvailability=true", (result) => {
-        if (!result.data) {
+    // The availability-check read, via hostClipboard: this runs to decide whether the
+    // paste-hyperlink button should be enabled -- including during page initialization -- not
+    // because the user asked to paste, so C# must not report a momentarily unreadable clipboard as
+    // a failed paste. Going through that helper also means the reply arrives as the text it really
+    // is: read directly, axios would turn a clipboard holding 42 into a number and the URL
+    // handling below would throw on it while a page was loading.
+    readClipboardTextForAvailabilityCheck().then((clipboardText) => {
+        if (!clipboardText) {
             callback("");
+            return;
         }
         get("app/selectedBookInfo", (bookInfo) => {
             if (!bookInfo.data) {
                 callback("");
+                return;
             }
-            callback(tryProcessHyperlink(result.data, bookInfo.data.id));
+            callback(tryProcessHyperlink(clipboardText, bookInfo.data.id));
         });
     });
 }
--- a/src/BloomBrowserUI/bookEdit/js/bloomEditing.ts
+++ b/src/BloomBrowserUI/bookEdit/js/bloomEditing.ts
@@ -1803,31 +1803,26 @@ document.addEventListener("keydown", (e: KeyboardEvent) => {
         return;
     }
 
-    if (isKey("x", "KeyX")) {
-        // Do the cut ourselves rather than letting Chromium do it. Chromium deletes the selected
-        // text and then silently discards the failed clipboard write, so a cut from a locked
-        // clipboard loses the text outright (BL-16459). cutSelection copies through C# first and
-        // leaves the text alone if that fails -- the same thing the Cut button does. Only take
-        // over where that implementation applies; elsewhere, let the browser cut and just check.
-        const editor = CKEDITOR.currentInstance;
-        // Only plain Ctrl+X. Ctrl+Shift+X is not a cut, and preventDefaulting it would invent one.
-        if (
-            !e.shiftKey &&
-            document.activeElement?.closest(".bloom-editable") &&
-            editor
-        ) {
-            e.preventDefault();
-            cutSelection();
-        } else {
+    if (isKey("x", "KeyX") || isKey("c", "KeyC")) {
+        // Both are left to Chromium, which is the only one of us that can put the full HTML
+        // flavor -- formatting, links, inline pictures -- on the clipboard. We just check
+        // afterwards and report a failure.
+        //
+        // Ctrl+X deserves more than that, because Chromium deletes the selection and then
+        // discards a failed clipboard write, losing the text outright (BL-16459). An earlier
+        // version of this handler therefore did the cut itself through C#, the way the Cut
+        // button does. That protected the text but put only *plain* text on the clipboard, so
+        // cutting a formatted phrase or an inline picture silently dropped everything but the
+        // words -- a worse bug, and one that struck every cut rather than only failing ones.
+        // Fixing it properly means teaching the copy path to carry several clipboard formats at
+        // once; until then the honest thing is to let the browser cut and tell the user when it
+        // didn't work. See the PR discussion.
+        //
+        // Nothing selected means nothing was copied, so there is nothing to report -- checking
+        // anyway would announce the failure of an operation that never happened.
+        if (document.getSelection()?.toString()) {
             verifyBrowserCopyAfterDefault();
         }
-        return;
-    }
-
-    if (isKey("c", "KeyC")) {
-        // Left to Chromium: its copy carries the HTML flavor, which matters when pasting into
-        // other programs, and unlike cut a failure loses nothing. We just check afterwards.
-        verifyBrowserCopyAfterDefault();
     }
 });
 
--- a/src/BloomBrowserUI/bookEdit/js/bloomEditing.ts
+++ b/src/BloomBrowserUI/bookEdit/js/bloomEditing.ts
@@ -1688,7 +1688,14 @@ async function copyImpl() {
 
 // See comment on copySelection
 export const cutSelection = () => {
-    cutSelectionImpl();
+    // Wrapped because the cut now waits for a round trip to C# before it deletes anything, and
+    // src/BloomBrowserUI/AGENTS.md requires asynchronous changes to the editable page DOM to hold
+    // off a concurrent page save; otherwise a save landing in that gap could capture the page
+    // half-cut. The paste path above does the same.
+    wrapWithRequestPageContentDelay(
+        () => cutSelectionImpl(),
+        "cutSelectionImpl",
+    );
 };
 
 async function cutSelectionImpl() {
@@ -1777,7 +1784,9 @@ document.addEventListener("paste", pasteHandler);
 document.addEventListener("keydown", (e: KeyboardEvent) => {
     // Not AltGr, which Windows reports as Ctrl+Alt: on many keyboard layouts AltGr+letter is how
     // the user types a character, and it must not be mistaken for a clipboard shortcut (least of
-    // all Ctrl+X, where we preventDefault).
+    // all Ctrl+X, where we preventDefault). This also means Ctrl+Alt+V no longer reaches
+    // pasteHandler as it used to; that is not a paste shortcut anyone documents, and letting a
+    // character key double as one is the bigger problem.
     if (!e.ctrlKey || e.altKey) return;
     const key = e.key?.toLowerCase();
     const isKey = (letter: string, code: string) =>
--- a/src/BloomBrowserUI/bookEdit/js/bloomEditing.ts
+++ b/src/BloomBrowserUI/bookEdit/js/bloomEditing.ts
@@ -1701,9 +1701,15 @@ export const cutSelection = () => {
 async function cutSelectionImpl() {
     const sel = document.getSelection();
     if (!sel) return;
-    // Don't delete what we failed to copy: that would lose the text outright. The user has
-    // been told the copy failed and can try the cut again.
-    if (!(await copyTextToClipboard(sel.toString()))) return;
+    const selectedText = sel.toString();
+    // Don't delete text we failed to copy: that would lose it outright. The user has been told the
+    // copy failed and can try the cut again.
+    //
+    // Only when there was text, though. A selection with nothing textual in it -- a picture on its
+    // own, say -- gives us nothing to put on the clipboard, and a plain-text copy never could have
+    // carried it anyway; treating that as a failed copy would make the Cut button do nothing at
+    // all, which is not what it used to do.
+    if (selectedText && !(await copyTextToClipboard(selectedText))) return;
     // Using ckeditor here because it's the only way I've found to integrate clipboard
     // ops into an Undo stack that we can operate from an external button.
     // We do a Save before and after to make sure that the cut is distinct from
--- a/src/BloomBrowserUI/bookEdit/js/bloomEditing.ts
+++ b/src/BloomBrowserUI/bookEdit/js/bloomEditing.ts
@@ -18,6 +18,13 @@ import {
     removeTransientVideoTimestampParams,
     SetupVideoEditing,
 } from "./bloomVideo";
+import {
+    copyTextToClipboard,
+    listenForBrowserClipboardOperations,
+    readTextFromClipboard,
+    verifyBrowserCopyAfterDefault,
+    verifyBrowserPasteAfterDefault,
+} from "./hostClipboard";
 import { SetupWidgetEditing } from "./bloomWidgets";
 import { setupOrigami, cleanupOrigami } from "./origami";
 import theOneLocalizationManager from "../../lib/localizationManager/localizationManager";
@@ -1643,11 +1650,12 @@ export function topBarButtonClick(button: { command: string }) {
     }
 }
 
-// These clipboard functions are implemented in Javascript because WebView2 doesn't seem to have
+// These clipboard functions are driven from Javascript because WebView2 doesn't seem to have
 // a C# api for doing them. I've made the exported functions synchronous because I'm not sure
 // what complications might come from calling an async function from C#. The implementations
-// mostly use clipboard API functions that are async, so those functions must be async.
-// We don't need to await them because nothing is using the result.
+// are async because they hand the actual clipboard work to C# over the API (see
+// hostClipboard.ts for why they must, rather than using navigator.clipboard), so the exported
+// functions do not await them; nothing is using the result.
 // The buttons that implement clipboard operations are currently only in Edit mode, so
 // this is a reasonable place for this code. If we support them elsewhere, we'll have to
 // find a way to share the code (and call it when not part of the workspaceBundle).
@@ -1670,12 +1678,12 @@ async function copyImpl() {
         // whitespace. But there's a greater chance they **don't** want unintended
         // trailing line breaks. This ".trimEnd()" works around an issue where multiple
         // copy/pastes can result in an extra line break being added. See BL-14051.
-        navigator.clipboard.writeText(
+        await copyTextToClipboard(
             activeCanvasElementEditable.innerText.trimEnd(),
         );
         return;
     }
-    navigator.clipboard.writeText(sel.toString());
+    await copyTextToClipboard(sel.toString());
 }
 
 // See comment on copySelection
@@ -1686,7 +1694,9 @@ export const cutSelection = () => {
 async function cutSelectionImpl() {
     const sel = document.getSelection();
     if (!sel) return;
-    await navigator.clipboard.writeText(sel.toString());
+    // Don't delete what we failed to copy: that would lose the text outright. The user has
+    // been told the copy failed and can try the cut again.
+    if (!(await copyTextToClipboard(sel.toString()))) return;
     // Using ckeditor here because it's the only way I've found to integrate clipboard
     // ops into an Undo stack that we can operate from an external button.
     // We do a Save before and after to make sure that the cut is distinct from
@@ -1761,13 +1771,63 @@ document.addEventListener("paste", pasteHandler);
 // end up pasting text into contenteditable elements twice, presumably
 // once because of our handler here and once because the other handler
 // is not looking for exactly this event and still gets called.
+// Clipboard keyboard shortcuts. These are handled from keydown rather than from the copy/cut/
+// paste events because CKEditor gets to those first inside a bloom-editable: it preventDefaults
+// copy and cut, and swallows paste before it reaches the document. keydown always arrives.
 document.addEventListener("keydown", (e: KeyboardEvent) => {
+    // Not AltGr, which Windows reports as Ctrl+Alt: on many keyboard layouts AltGr+letter is how
+    // the user types a character, and it must not be mistaken for a clipboard shortcut (least of
+    // all Ctrl+X, where we preventDefault).
+    if (!e.ctrlKey || e.altKey) return;
     const key = e.key?.toLowerCase();
-    const isCtrlV = e.ctrlKey && (key === "v" || e.code === "KeyV");
-    if (isCtrlV) {
+    const isKey = (letter: string, code: string) =>
+        key === letter || e.code === code;
+
+    if (isKey("v", "KeyV")) {
         pasteHandler(e);
+        // If pasteHandler took over (it preventDefaults when it is going to handle an image
+        // paste itself) it reports its own failures through the API. Otherwise Chromium is doing
+        // the paste, and only C# can tell us whether the clipboard was actually readable.
+        if (!e.defaultPrevented) {
+            verifyBrowserPasteAfterDefault();
+        }
+        return;
+    }
+
+    if (isKey("x", "KeyX")) {
+        // Do the cut ourselves rather than letting Chromium do it. Chromium deletes the selected
+        // text and then silently discards the failed clipboard write, so a cut from a locked
+        // clipboard loses the text outright (BL-16459). cutSelection copies through C# first and
+        // leaves the text alone if that fails -- the same thing the Cut button does. Only take
+        // over where that implementation applies; elsewhere, let the browser cut and just check.
+        const editor = CKEDITOR.currentInstance;
+        // Only plain Ctrl+X. Ctrl+Shift+X is not a cut, and preventDefaulting it would invent one.
+        if (
+            !e.shiftKey &&
+            document.activeElement?.closest(".bloom-editable") &&
+            editor
+        ) {
+            e.preventDefault();
+            cutSelection();
+        } else {
+            verifyBrowserCopyAfterDefault();
+        }
+        return;
+    }
+
+    if (isKey("c", "KeyC")) {
+        // Left to Chromium: its copy carries the HTML flavor, which matters when pasting into
+        // other programs, and unlike cut a failure loses nothing. We just check afterwards.
+        verifyBrowserCopyAfterDefault();
     }
 });
+
+// Ctrl+C, Ctrl+X and Ctrl+V in a text box are handled entirely inside Chromium, which never
+// tells us whether the clipboard worked. (For paste, note that pasteHandler above deliberately
+// bows out when focus is in a bloom-editable, so C# is not involved at all.) This gets each one
+// checked on the C# side so the user finds out, instead of pasting stale content -- or nothing
+// -- and concluding that their copy never happened (BL-16459).
+listenForBrowserClipboardOperations(document);
 
 async function pasteImpl(imageAvailable: boolean) {
     const canvasElementManager = theOneCanvasElementManager;
@@ -1785,7 +1845,7 @@ async function pasteImpl(imageAvailable: boolean) {
         "bloom-editable bloom-visibility-code-on",
     )[0] as HTMLElement;
 
-    const textToPaste = await navigator.clipboard.readText();
+    const textToPaste = await readTextFromClipboard();
     if (!textToPaste) {
         return;
     }
--- a/src/BloomBrowserUI/bookEdit/js/hostClipboard.ts
+++ b/src/BloomBrowserUI/bookEdit/js/hostClipboard.ts
@@ -15,19 +15,32 @@
 import { AxiosResponse } from "axios";
 import { getAsync, postJson, postJsonAsync } from "../../utils/bloomApi";
 
-// Put text on the clipboard. Resolves true if it actually got there; if not, C# has
-// already told the user, and callers that are about to destroy the original (cut) should
-// leave it alone.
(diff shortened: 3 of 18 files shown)


src/BloomExe/Utils/BloomClipboard.cs
```diff
@@ -0,0 +1,266 @@
...
+            text = "";
+
+            // Ask the OS before believing the read: inside Bloom, reading a locked clipboard
+            // returns "" rather than throwing (see the class comment), so a failed paste would
+            // otherwise be indistinguishable from an empty clipboard and reported to nobody.
+            if (!TryOpenClipboardBriefly(out var lastError))
+            {
+                RunOnUiThread(() => ReportPasteFailure(new Win32Exception(lastError)));
+                return false;
+            }
```
JohnThomson, line 96: [Devin] **Bug**: Users get a false "unable to paste" error just from opening a menu or a page

A clipboard failure message is shown whenever any part of Bloom merely asks whether the clipboard holds text, not only when the user actually pastes. `common/clipboardText` GET is not only the paste path: `CanvasElementContextControls.tsx` queries it whenever a canvas element's menu opens (to decide whether "Paste Text" should be enabled), `hyperlinks.ts` queries it for the paste-hyperlink button, and the note on the endpoint registration in `CommonApi.cs` records that this GET also fires during page initialization.

Impact: a clipboard held for a moment by another program produces an unexplained "Bloom was not able to paste." toast — and a Sentry report — at a moment when the user never tried to paste. Before this PR those checks were silent, because a .NET clipboard read inside Bloom does not throw; the new OS-level probe makes them fail loudly.

src/BloomBrowserUI/bookEdit/js/bloomEditing.ts
```diff
@@ -1686,7 +1694,9 @@ export const cutSelection = () => {
 async function cutSelectionImpl() {
     const sel = document.getSelection();
     if (!sel) return;
-    await navigator.clipboard.writeText(sel.toString());
+    // Don't delete what we failed to copy: that would lose the text outright. The user has
+    // been told the copy failed and can try the cut again.
+    if (!(await copyTextToClipboard(sel.toString()))) return;
```
JohnThomson: [Devin] **Bug**: Cutting text can let a page be saved mid-operation, because the cut is not registered as pending work

The cut waits for a round trip to the desktop side before it removes the selected text (`await copyTextToClipboard(...)`) without registering that work as in progress, so a save that happens during that gap can capture the page in a half-finished state.

Impact: a cut that overlaps a page save can be saved inconsistently — the text still present, or removed after the content was already captured.

`src/BloomBrowserUI/AGENTS.md` states: "When code makes changes to the editable page dom using asynchronous operations, it should use wrapWithRequestPageContentDelay to make sure any requests for page content wait until the async tasks complete." The paste path already does this (`pasteClipboard` wraps `pasteImpl`); this PR made the cut asynchronous in the same way — and reachable from the new Ctrl+X handler as well as the Cut button — without the same wrapper.

src/BloomBrowserUI/bookEdit/js/bloomEditing.ts
```diff
@@ -1761,14 +1771,64 @@ document.addEventListener("paste", pasteHandler);
...
+        // leaves the text alone if that fails -- the same thing the Cut button does. Only take
+        // over where that implementation applies; elsewhere, let the browser cut and just check.
+        const editor = CKEDITOR.currentInstance;
+        // Only plain Ctrl+X. Ctrl+Shift+X is not a cut, and preventDefaulting it would invent one.
+        if (
+            !e.shiftKey &&
+            document.activeElement?.closest(".bloom-editable") &&
+            editor
+        ) {
+            e.preventDefault();
```
JohnThomson: [Devin] **Investigate**: Ctrl+X now puts plain text only on the clipboard, losing the HTML flavor a native cut carried

By preventing the default and routing the cut through `cutSelection` (which posts `sel.toString()` to `common/clipboardText`), Ctrl+X inside a `.bloom-editable` now writes **plain text**, whereas Chromium's own cut also carried the HTML flavor. Ctrl+C is deliberately left to Chromium for exactly that reason, so cut and copy now behave differently: cutting a formatted phrase and pasting it back into Bloom — or into Word — drops the formatting, while copying the same phrase keeps it.

This was a knowing trade: routing the cut through C# is what lets us know the copy failed and therefore *not* delete the text, which is the data-loss bug this PR set out to fix. But the fidelity cost is real and user-visible, so it is going to the developer as a decision rather than being settled here.

src/BloomExe/Utils/BloomClipboard.cs
JohnThomson: [Claude Opus 5] Fixed in `331129dcb2`. Good catch, and squarely a regression from the previous commit.

The GET now takes a `checkingAvailability=true` flag, and `BloomClipboard.TryGetText` takes a matching `reportFailure` parameter. The three callers that are only asking whether there is something to paste — the canvas element's menu, the paste-hyperlink button, and the page-initialization query — pass it and get silence; a real paste still reports. So a clipboard held for a moment by another program can no longer produce an "unable to paste" message at someone who never tried to paste.

src/BloomBrowserUI/bookEdit/js/bloomEditing.ts
JohnThomson: [Claude Opus 5] Fixed in `331129dcb2`. `cutSelection` now wraps `cutSelectionImpl` in `wrapWithRequestPageContentDelay` with its own delay id, exactly as `pasteClipboard` wraps `pasteImpl`, so a page save waits for the cut to finish instead of capturing it half-done. Thanks for citing the rule in `src/BloomBrowserUI/AGENTS.md` — the previous commit made the cut asynchronous and missed it.

src/BloomBrowserUI/bookEdit/js/hostClipboard.ts
JohnThomson, line 1: [Devin] **Bug**: Pasting text that looks like a number does nothing

Clipboard text fetched from C# is discarded whenever it isn't a plain string, so copying something like a number and pasting it into a book silently inserts nothing.

The endpoint replies with `ReplyWithText` (`text/plain`), but `getAsync` was a bare `axios.get`, and axios's default `transformResponse` unconditionally attempts `JSON.parse` on the body. So a clipboard containing `42`, `true`, `null` or a JSON object arrived as a number/boolean/object and was thrown away by the `typeof response.data === "string"` check.

Impact: a user who copies a number and pastes gets nothing, with no explanation. It also made such a clipboard look empty to the availability check, so "Paste Text" would appear disabled with something perfectly pasteable on the clipboard.

JohnThomson: [Claude Opus 5] Fixed in `ac7c480a`. Real bug, and a good one — thank you.

Both clipboard reads now pass `transformResponse: [body => body]` so the reply arrives exactly as sent, and `getAsync` takes an optional axios config for the purpose. There is a regression test over both reads asserting the transform is requested and hands the body back untouched, so the JSON-parse trap can't quietly come back.

JohnThomson:
_(posted by Claude Opus 5 on John's behalf)_

**Consulted Devin** on 2026-08-03, three times as the branch changed, most recently up to commit `ac7c480a648495e85a0d1656e45cf0f19d6f51b1`.

It earned its keep: it found three real bugs, all of them introduced by this PR, each now fixed with its own resolved thread.

- [False "unable to paste" from merely opening a menu](https://github.com/BloomBooks/BloomDesktop/pull/8140#discussion_r3706586933) — fixed in `331129dc`
- [A cut that let a page save catch the page half-done](https://github.com/BloomBooks/BloomDesktop/pull/8140#discussion_r3706588819) — fixed in `331129dc`
- [Pasting number-like text did nothing](https://github.com/BloomBooks/BloomDesktop/pull/8140#discussion_r3706666038) — fixed in `ac7c480a`

One [Investigate flag is deliberately left open](https://github.com/BloomBooks/BloomDesktop/pull/8140#discussion_r3706588941): routing Ctrl+X through Bloom's own cut means it now puts plain text on the clipboard, losing the HTML flavour Chromium's cut carried. That is a user-visible trade-off rather than a defect, so it is going to the developer as a decision.

Informational items (6 on the first pass, 9 on the last) were assessed and not mirrored; the ones worth a human's attention are in the preflight report linked on the YouTrack card.

CI is green, and CodeRabbit is switched off in this repo (`.coderabbit.yml` sets `auto_review.enabled: false`), so no other bot has anything pending.

src/BloomBrowserUI/bookEdit/js/bloomEditing.ts
JohnThomson, line 1: [Devin] **Bug**: Cutting a selection that contains anything but plain text now deletes the formatting and pictures without putting them on the clipboard

When the user pressed Ctrl+X inside a text box, the built-in cut was cancelled and replaced by one that copies only the plain text while still deleting everything that was selected, so anything that is not plain text disappeared and could not be pasted back.

Impact: a user who cuts a passage containing bold/italic text, a link, or an inline picture got only unformatted words on the clipboard, and the rest was gone from the page. Previously Ctrl+X was left to Chromium, which puts the full HTML flavour (and images) on the clipboard and then removes the selection.

JohnThomson, line 1: [Devin] **Bug**: Pressing the copy shortcut with nothing selected can pop up a message saying the copy failed

Every press of the copy shortcut asked the host to check the clipboard even when nothing was selected and therefore nothing was copied, so a clipboard briefly held by another program produced a "Bloom was not able to copy that" message about an operation that never happened — and filed a Sentry report for it.

JohnThomson: [Claude Opus 5] Fixed in `b1d0c1d372`, by giving Ctrl+X back to Chromium. You're right that this was the worse bug: it struck *every* cut of formatted content, not only ones where the clipboard failed, so it traded a failure-path problem for a normal-path one.

That does leave the original complaint half-addressed, which is worth stating plainly: a cut that fails still deletes the text — Chromium's behaviour, unchanged since before this branch — but the user is now told about it instead of finding out later. Fixing it properly means teaching the copy path to put several clipboard formats on at once, which is real work and a call for the developer; it's in the preflight report as a decision, and the handler comment records why the take-over was removed so nobody reinstates it by halves.

JohnThomson: [Claude Opus 5] Fixed in `b1d0c1d372`. The check now runs only when there is actually a selection, so a keystroke that copied nothing no longer announces a failure or files a Sentry report.

Worth noting your own observation held: copies from a plain input field, where the page selection reads as empty, are still covered — the browser's own copy event fires there and triggers the same check.

src/BloomExe/Utils/BloomClipboard.cs
JohnThomson, line 1: [Devin] **Investigate**: The verification probe itself takes the clipboard, which may race with Chromium's still-pending write

`TryOpenClipboardBriefly` really does `OpenClipboard`/`CloseClipboard`, i.e. it momentarily owns the clipboard. The comment argued this was harmless because it "runs after the browser's operation has completed", but the front end only defers by `window.setTimeout(..., 0)` plus one localhost API round trip — and Chromium's actual clipboard write happens asynchronously in the browser process, not in the renderer task we deferred past.

If our probe wins that race, Chromium's write fails — silently, since Chromium discards clipboard errors — and **the very failure the feature exists to report could be caused by the check itself.**

A fair hit: the comment overclaimed, and it has been corrected to say the window exists and is unmeasured. Leaving this open as a decision, because closing it properly means either delaying the probe (which weakens the check) or finding a way to ask the OS without taking the clipboard.

JohnThomson, line 1: [Devin] **Investigate**: A read that the OS refuses now returns nothing even when the cached OLE data object would have served the text

`TryGetText` now gates the read on `TryOpenClipboardBriefly`. As the class comment documents, inside Bloom a .NET read of a locked clipboard usually *succeeds*, from OLE's cached data object. So the new behaviour is: whenever another program holds the clipboard for more than ~200ms, a paste that previously inserted the (typically still-correct) cached text now inserts nothing and reports a failure instead.

That is a deliberate trade of availability for honesty — the cached read is exactly what made the original bug invisible — but it applies to every caller of this endpoint, including the link editor's paste button and the subscription-code paste in Collection Settings, whose paste buttons will now do nothing in that situation.

Left open: whether that trade is right for those flows is a product call, not something to settle from the code.

JohnThomson, line 1: [Devin] **Bug**: The new clipboard health check can itself break the copy or paste it is checking on

The check asks the operating system for the clipboard and immediately releases it, which is the only truthful way to ask — but for those microseconds Bloom itself owns the clipboard. The browser defers the check only by a zero-length timeout plus one localhost round trip, while Chromium's actual write happens asynchronously in its own process. If the check wins that race, Chromium's write fails silently, so **the check can cause the very failure it then reports.**

Left open deliberately: closing it means either delaying the check (which weakens it) or finding a way to ask without taking the clipboard, and this run has reached its fix-cycle limit. The code comment now records the race honestly rather than arguing it away, and it is decision 2 in the preflight report.

src/BloomBrowserUI/bookEdit/js/bloomEditing.ts
JohnThomson, line 1: [Devin] **Investigate**: PR description contradicts the shipped Ctrl+X behaviour

The description said "Ctrl+X now goes through the same path as the Cut button — copy first, and leave the text alone if that fails", and listed "Ctrl+X → toast, **and the text is preserved**" under Verification. The code deliberately does the opposite after the take-over was reverted: Ctrl+X is left to Chromium so the rich clipboard flavours survive, and a locked clipboard still loses the cut text.

src/BloomExe/Utils/BloomClipboard.cs
JohnThomson, line 1: [Devin] **Investigate**: P/Invoke into user32 makes the clipboard path Windows-only

`OpenClipboard`/`CloseClipboard` are imported from `user32.dll` unconditionally, and the probe runs on every clipboard read. On a non-Windows run that would throw `DllNotFoundException` out of the API handler rather than degrading gracefully. There are still `#if !__MonoCS__` blocks elsewhere in the tree, so if Linux is nominally supported this needs a platform guard.

JohnThomson, line 1: [Devin] **Investigate**: One remaining unguarded clipboard write was not converted

The PR converts the reader-stats, BloomPub, copy-page-HTML and CommonApi writes to `BloomClipboard.TrySetText`, but `PortableClipboard.SetText(path)` in `EditingView.cs` (copying an image path to the clipboard) is still unguarded and will throw on a locked clipboard. It is the one write left out of the otherwise complete migration.

src/BloomBrowserUI/bookEdit/js/bloomEditing.ts
JohnThomson: [Claude Opus 5] Fixed — thank you, this was the most useful of the four. The description now says Ctrl+X is left to the browser, that a failed cut still loses the text, and why the take-over was reverted. The Verification list no longer claims the text is preserved.

Getting this wrong is worse than a code slip: the description is what a reviewer and the ticket believe, so it would have signed off behaviour the branch does not have.

src/BloomExe/Utils/BloomClipboard.cs
JohnThomson: [Claude Opus 5] Checked, and not an issue for this build. `BloomExe.csproj` declares `<RuntimeIdentifiers>win-x64;win-arm64</RuntimeIdentifiers>`, and the only browser implementation left is WebView2, which is Windows-only — so there is no non-Windows run for the P/Invoke to fail on. The surviving `#if !__MonoCS__` blocks are vestigial rather than a live Linux target.

Worth flagging that your own analysis reached the same conclusion; recording the evidence here so the next person doesn't have to re-derive it. If Linux ever comes back, this probe is one of the places that needs a platform guard.

JohnThomson: [Claude Opus 5] Accurate but not a crash risk, and deliberately out of scope. That write sits inside `CopyImageToClipboard`, which already catches `ExternalException` and shows the image-specific message ("Bloom had problems using your computer's clipboard. Some other program may be interfering"), with a general catch behind it — so a locked clipboard is handled there, just differently from the text paths.

Leaving it deliberately: that dialog predates this work, its wording explains the cause better than the toast does, and unifying them would retire two localized strings. It is in the preflight report as a decision for the developer rather than something to change on the way past.

JohnThomson:
_(posted by Claude Opus 5 on John's behalf)_

**Consulted Devin** again, up to commit `8bf3de6b16f6fd260618dc9fe3a2b5c494829976` — the fourth pass on this branch, and the last for this run: preflight caps its fix-and-re-review cycles at four, so anything outstanding now goes to the developer rather than into another round.

Across the four passes Devin found **five bugs, every one introduced by this branch**, and all five it now marks resolved:

| finding | fixed in |
| --- | --- |
| [False "unable to paste" from opening a menu](https://github.com/BloomBooks/BloomDesktop/pull/8140#discussion_r3706586933) | `331129dc` |
| [A cut that let a page save catch the page half-done](https://github.com/BloomBooks/BloomDesktop/pull/8140#discussion_r3706588819) | `331129dc` |
| [Pasting number-like text did nothing](https://github.com/BloomBooks/BloomDesktop/pull/8140#discussion_r3706666038) | `ac7c480a` |
| [The Ctrl+X take-over destroyed formatting and pictures on every cut](https://github.com/BloomBooks/BloomDesktop/pull/8140#discussion_r3706757417) | `b1d0c1d3` |
| [The copy shortcut reported failures when nothing was selected](https://github.com/BloomBooks/BloomDesktop/pull/8140#discussion_r3706757672) | `b1d0c1d3` |

Settled on their own threads this pass: the [PR description contradicted the shipped Ctrl+X behaviour](https://github.com/BloomBooks/BloomDesktop/pull/8140#discussion_r3706833146) (fixed — it had promised behaviour the revert removed), the [Windows-only P/Invoke](https://github.com/BloomBooks/BloomDesktop/pull/8140#discussion_r3706834720) (not an issue: this project builds `win-x64;win-arm64` only), and [the one unconverted clipboard write](https://github.com/BloomBooks/BloomDesktop/pull/8140#discussion_r3706834884) (already handled by the image-copy dialog; left deliberately).

**Three threads are deliberately left open** for the developer, all judgement calls rather than defects:

- [The health check can itself break the copy it is checking](https://github.com/BloomBooks/BloomDesktop/pull/8140#discussion_r3706833010) — the probe momentarily owns the clipboard and may race with Chromium's write.
- [A busy clipboard now pastes nothing](https://github.com/BloomBooks/BloomDesktop/pull/8140#discussion_r3706766595) where it used to paste cached text.
- [Ctrl+X and the plain-text clipboard](https://github.com/BloomBooks/BloomDesktop/pull/8140#discussion_r3706588941) — now about the Cut *button*, since the keyboard take-over was reverted.

CI is green, and CodeRabbit is switched off in this repo (`auto_review.enabled: false`).

src/BloomBrowserUI/bookEdit/js/bloomEditing.ts
JohnThomson: [Claude Opus 5] **Developer decision: leave as is for now.** John's words: "We plan to replace CkEditor entirely; I think that will be the place to attempt a better fix."

So Ctrl+X stays with the browser (formatting preserved, a failed cut still loses the text and reports it), and the multi-format clipboard work is deferred to that replacement, where Bloom would be able to supply the clipboard payload and learn whether the write succeeded.

Noted on [BL-6681 "Remove ckeditor?"](https://issues.bloomlibrary.org/youtrack/issue/BL-6681) so the requirement isn't lost when that work is designed, with a link back to BL-16459.

src/BloomExe/Utils/BloomClipboard.cs
JohnThomson: [Claude Opus 5] **Developer decision: leave as is** — and your suspicion was right, so this is now backed by a measurement rather than an argument.

John asked whether the retrying that clipboard operations normally do wouldn't protect against a hold as short as ours. It does. A helper process opened and closed the clipboard continuously — **7.4 million times in ten seconds** — while Chromium wrote to it twenty times, and **all twenty writes landed on the clipboard**. Whoever wants the clipboard retries for it, so a hold measured in microseconds is simply ridden over; and the real check takes the clipboard once per copy, not millions of times.

So the race exists on paper and does not bite. The comment in `BloomClipboard` now records the measurement instead of speculating — my original comment waved the risk away for the wrong reason, and the intermediate version overstated it. Fixed in `062e7bde`. Resolving.

JohnThomson: [Claude Opus 5] **Developer decision: leave as is.** Reporting the failure and pasting nothing is preferred over silently pasting possibly-stale cached text — the silent version is what hid this bug in the first place, and a message the user can act on costs them one retry.

That applies to the link editor's paste button and the subscription-code paste in Collection Settings as well as ordinary pasting; the QA notes on the card cover them. Resolving.

src/BloomBrowserUI/bookEdit/bloomField/hyperlinks.ts
JohnThomson, line 1: [Devin] **Bug**: Number-like or JSON-like clipboard content can break the paste-hyperlink check

The clipboard reply was still read through the plain request helper, which converts a reply like `42` or `true` into a non-text value, so the follow-on URL handling throws — while a page is loading, since that is when this check runs.

Impact: if the clipboard happens to hold something that looks like a number or JSON, opening a page can raise an internal error and the paste-hyperlink button ends up in the wrong state.

The PR already documented this hazard and worked around it for the two new read helpers; this older call site was missed.

src/BloomExe/Utils/BloomClipboard.cs
JohnThomson, line 1: [Devin] **Investigate**: Pasted text now comes from the .NET clipboard, so it may carry CRLF where it used to be LF

`pasteImpl` used to call `navigator.clipboard.readText()`; it now goes through C#. Text copied from other Windows programs is stored on the clipboard as CRLF, and Chromium's clipboard read normalizes line endings for the web API, whereas the WinForms read returns the raw string. That text is fed straight into the editor's `insertText`, where a stray `\r` may be treated as an extra break or emitted literally — so multi-line pastes could start looking different.

src/BloomBrowserUI/bookEdit/bloomField/hyperlinks.ts
JohnThomson: [Claude Opus 5] Fixed in `f5d24744`. Fair hit — I fixed exactly this hazard for the two new read helpers and left the one older call site on the raw helper, so the PR documented the trap and then fell into it.

The paste-hyperlink check now goes through `readClipboardTextForAvailabilityCheck` like the other callers, which asks axios for the reply untouched. While there I added the missing `return`s after the empty-string callbacks; without them the code ran on and called back a second time.

src/BloomExe/Utils/BloomClipboard.cs
JohnThomson: [Claude Opus 5] Fixed in `f5d24744` rather than left for a decision — restoring the previous behaviour is unambiguous here, and a silent change to how multi-line pastes look is not something to ship unnoticed.

`TryGetText` now normalizes CRLF and lone CR to LF, which is what `navigator.clipboard.readText()` was giving the front end before. `BloomClipboardTests` covers it, including that a CRLF pair doesn't become two breaks — the obvious way to get this wrong.

src/BloomBrowserUI/bookEdit/js/bloomEditing.ts
JohnThomson, line 1: [Devin] **Bug**: Cut button does nothing when the selected material contains no plain text

`cutSelectionImpl` aborts the deletion when the copy comes back false — which is what protects text from a failed clipboard write — but "there was nothing to copy" also comes back false. So selecting something with no text in it, a picture on its own for instance, and pressing Cut did nothing at all. Before this branch the empty copy resolved and the deletion went ahead.

JohnThomson: [Claude Opus 5] Fixed in `7ccf9142`. The abort now applies only when there actually was text to copy; with no text we fall through to the deletion, as before this branch. A plain-text copy could never have carried a picture anyway, so nothing that used to be kept is lost.

This is the third time the same seam has bitten: `false` from the copy means both "the clipboard refused it" and "there was nothing to put there". Worth watching if anyone extends that contract.

src/BloomExe/Utils/BloomClipboard.cs
JohnThomson, line 1: [Devin] **Bug**: Pasting can briefly freeze the program when another program is holding the clipboard

The clipboard availability probe waits 50 milliseconds after its very last failed try before giving up, and the Paste toolbar command runs this probe on the thread that draws the window, so the whole program stops responding for the full waiting period.

`TryOpenClipboardBriefly` loops 4 times; on each failure it sleeps 50 ms, *including* after attempt 4 whose result is then discarded, so a failing probe always cost 200 ms of sleeping instead of 150 ms. And `_pasteCommand.Implementer` calls into it synchronously on the UI thread.

Impact: when the clipboard is busy, clicking Paste freezes Bloom's window for a fifth of a second before the message appears.

JohnThomson: [Claude Opus 5] **Half fixed in `82f2bfd0`, and the other half is deliberately left for the developer.**

Fixed: the pointless sleep after the final attempt. That was simply a bug in my loop — nothing left to wait for — so a failing probe now costs 150 ms rather than 200 ms.

Not fixed: the check is still synchronous on the UI thread, so a busy clipboard still hitches the window for the remaining 150 ms. Making it asynchronous is a design change to how the paste command works, not a tidy-up, and this is the **ninth** bug this review has found on this branch — every one introduced by it. Eight were mine to fix; this last one is where I stop patching and hand the contract to a human, because the pattern says my judgement on this seam has needed an outside check every time.

Recorded in the preflight report on the card as the one open item.

JohnThomson:
_(posted by Claude Opus 5 on John's behalf)_

**Closing without merging.** The decision is not that the code is broken — it passed everything by the end — but that the shape of the change isn't worth it, and the one symptom that prompted BL-16459 most sharply (a failed Ctrl+X deleting text) is better fixed when CKEditor is replaced.

### Why, in one paragraph

Making clipboard failures visible turns out to require threading a success/failure notion through a lot of places, because neither layer that performs the work will tell you it failed: Chromium silently discards clipboard errors, and a .NET clipboard *read* inside Bloom quietly succeeds from a cached copy while another program holds the clipboard. Nine bugs were found and fixed **within this branch during review** — three of them from a single overloaded return value, where `false` from the copy meant both "the clipboard refused this" and "there was nothing to copy". That rate of self-inflicted defects on one seam is the real signal: the design was error-prone even where each individual fix was correct.

### Worth keeping, if anyone returns to this

The measurements were the expensive part, and they are all in the branch's comments and history:

- **Chromium never reports a clipboard failure to Javascript.** With the clipboard held elsewhere, `navigator.clipboard.writeText()` still resolves, `execCommand("copy")` still returns `true`, and `readText()` resolves with `""`. Adding `.catch()` anywhere in the front end accomplishes nothing.
- **A .NET clipboard read inside Bloom does not fail either.** Measured side by side against a freshly started process: in a process that has recently read the clipboard — which Bloom always has, since the edit-buttons timer polls it — `ContainsText()` returned `True` and `GetText()` returned `""`, while both threw in the fresh process. Only *writes* fail honestly. That asymmetry is exactly why the Copy button reported failures and Ctrl+C did not.
- **The truthful check is raw `OpenClipboard`/`CloseClipboard`.** And a momentary grab like that cannot break another program's write: a helper process took the clipboard 7,366,280 times in ten seconds while Chromium wrote twenty times, and all twenty writes landed.
- **CKEditor prevents default on copy and cut inside a `.bloom-editable`, and swallows paste**, so keyboard shortcuts have to be handled from `keydown`, not from the clipboard events.
- **Protecting Ctrl+X by doing the cut ourselves is a trap** unless the copy path can carry several clipboard formats: Bloom's own cut is plain-text only, so it silently dropped formatting and inline pictures on *every* cut. Windows' HTML clipboard format needs a byte-offset header that .NET does not write for you — that is the real work behind a proper fix.

The branch `BL-16459-clipboard-failure-reporting` is left in place (not deleted) so the diff and these notes stay reachable, and this is referenced from [BL-6681 "Remove ckeditor?"](https://issues.bloomlibrary.org/youtrack/issue/BL-6681), where a future editor replacement should ask: can Bloom supply the clipboard payload (rich *and* plain) and learn whether the write succeeded? If yes, this whole class of problem disappears rather than being reported.

Full write-up, including the seven decisions taken along the way: https://raw.githack.com/BloomBooks/dev-process-artifacts/main/deciders/bloomdesktop-BL-16459-clipboard-failure-reporting.html

Thanks to Devin, which found all nine and was right every time.

PR #8141, opened 2026-08-03
Capture Bloom's stacks at the moment page checks fail (BL-16612)

andrew-polk:
CI-only. No product code — one script, nothing that can affect Bloom's behavior for users or the behavior of the run it watches.

## Why the existing captures can't settle the question

The watchdog from #8114 worked: the [2026-08-03 hang](https://github.com/BloomBooks/BloomDesktop/actions/runs/30791565819) produced three stack captures proving Bloom is **hard-deadlocked** in `OffScreenBrowser.RunAndBlock` — 0.72s of CPU across 24 minutes, working set paging 258MB → 16MB. That much is now certain.

But its captures come from 8 minutes in, and by then the evidence can't distinguish anything. The wedged thread holds the API lock (`SyncObj`), so every later request — and the suite polls constantly — must queue behind it and register as blocked. **A fully-consumed worker pool is the inevitable end state of any hang that holds that lock, whatever made the navigation fail.** So "all 6 workers were unavailable" is explained by the hang rather than being evidence about its cause. It cannot separate worker starvation from a WebView2/renderer stall.

## The moment that does distinguish them

When the navigation gives up — about **26 seconds** after Bloom launches, long before the old 8-minute threshold:

| what the stacks show at that instant | conclusion |
|---|---|
| workers all occupied | starvation **caused** it |
| workers parked idle in `WaitHandle.WaitAny` | starvation did **not**; the pool filling up later is just aftermath |

So the watchdog now also captures **on a trigger**: it watches Bloom's log and captures the instant `Failed to navigate fully` appears, then again a few seconds later (identical stacks prove it is stuck rather than crawling). Polling drops to 5s while Bloom is young so the window isn't missed, and the log's size is recorded as it grows — which both confirms the log is readable and marks when Bloom went quiet.

## Why not just sample stacks continuously

`dotnet-stack` starts an EventPipe session and walks every thread; it isn't free. Doing that every few seconds through startup could itself slow Bloom enough to cause the 10-second navigation timeout we're trying to explain — manufacturing the failure under investigation. The only frequent sampling here is `Get-Process`, a counter read.

Still `dotnet-stack` rather than `dotnet-dump`, for the reason given in #8114: this repo is public and artifacts are widely downloadable, so we must not publish a memory dump of a process holding CI secrets.

## Testing

Exercised locally against a real .NET process, driving both paths: the trigger path fires from a log append and produces two captures of genuine per-thread stacks; the hung path still works after the counter rename.

That testing earned its keep — it caught a bug that would have silently produced **nothing** in CI. PowerShell variable names are case-insensitive, so the counter `$triggerCaptures` *was* the parameter `$TriggerCaptures`; initialising it to 0 turned the limit into 0 and made the capture condition `0 -lt 0`, false forever. The counters are now `$triggerShotsTaken` / `$hungShotsTaken` so they cannot alias the parameters.

At roughly 1 hang in 3 scheduled runs, this should have an answer within a few nights.

[Devin review](https://devinreview.com/BloomBooks/BloomDesktop/pull/8141)

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8141)

### DIFF
--- a/build/ci/watch-for-hung-bloom.ps1
+++ b/build/ci/watch-for-hung-bloom.ps1
@@ -1,46 +1,79 @@
 <#
 .SYNOPSIS
-    Background watchdog for the nightly's visual-regression step: captures managed thread stacks
-    from a Bloom.exe that has stopped making progress.
+    Background watchdog for the nightly's visual-regression step: captures managed thread stacks from a
+    Bloom.exe at the two moments that explain the intermittent publishing hang (BL-16612).
 
 .DESCRIPTION
-    The nightly visual-regression suite intermittently hangs (BL-16612). When it does, Bloom stops
-    writing to its log and every test case then times out, so the run tells us only that something
-    stopped — never what. The one thing that would identify the cause is what each thread was doing
-    while it was stuck, and we cannot get that after the fact: the suite's afterAll kills Bloom with
-    `taskkill /T /F` before any later workflow step could look at it.
+    The nightly visual-regression suite intermittently hangs. When it does, Bloom stops writing to its
+    log and every test case then times out, so the run tells us only that something stopped -- never
+    what. The one thing that would identify it is what each thread was doing, and we cannot get that
+    after the fact: the suite's afterAll kills Bloom with `taskkill /T /F` before any later workflow
+    step could look at it. So this runs alongside the tests.
 
-    So this runs alongside the tests instead. It polls for a Bloom.exe that has been alive far longer
-    than a healthy run needs, and captures `dotnet-stack report` for it while it is still wedged. On
-    the observed failures Bloom sat stuck for roughly 24 minutes, so there is a wide window to catch.
+    It captures at two distinct moments, and the distinction is the whole point:
+
+    1. ON TRIGGER -- the instant Bloom's log says the page-checks navigation gave up
+       ("Failed to navigate fully..."). This is the moment that decides the open question. If the
+       server's worker threads are all occupied right then, worker starvation caused the failure; if
+       they are sitting idle in WaitHandle.WaitAny, starvation did not, and the navigation failed for
+       some other reason (WebView2/renderer) with the pool filling up only afterwards.
+
+    2. WHEN HUNG -- a Bloom alive far longer than a healthy run needs (~5 min healthy versus 20+ when
+       hung). This is the end state. It is worth having, but on its own it cannot answer the question
+       above: the wedged thread holds the API lock, so every later request piles up behind it and a
+       fully-consumed pool is the inevitable consequence of ANY such hang, whatever caused it. That is
+       why moment 1 exists.
+
+    Why trigger off the log rather than just sampling stacks continuously: a `dotnet-stack report`
+    starts an EventPipe session and walks every thread, which is not free. Doing that every few
+    seconds through Bloom's startup could itself slow things enough to cause the 10-second navigation
+    timeout we are trying to explain -- manufacturing the very failure under investigation. So the only
+    frequent sampling here is `Get-Process` (a cheap counter read), and stack walks happen just twice
+    around the trigger and up to three times once already hung.
 
     Deliberately uses dotnet-stack (text call stacks) and NOT dotnet-dump: this repo is public and
-    workflow artifacts are widely downloadable, so we must not publish a memory dump of a process
-    that holds CI secrets. Text stacks carry no heap contents.
+    workflow artifacts are widely downloadable, so we must not publish a memory dump of a process that
+    holds CI secrets. Text stacks carry no heap contents.
 
-    Purely observational — it reads process state and never signals, kills, or otherwise touches
-    Bloom, so it cannot change the outcome of the run it is watching. Every failure is swallowed:
-    a broken watchdog must never turn a nightly red.
+    Purely observational -- it reads process state and a log file, and never signals, kills, or
+    otherwise touches Bloom, so it cannot change the outcome of the run it is watching. Every failure
+    is swallowed: a broken watchdog must never turn a nightly red.
+
+.PARAMETER TriggerPattern
+    The log line meaning "the page-checks navigation just gave up". Matched as a regex. Includes the DOM
+    name on purpose: Bloom logs "Failed to navigate fully" from two different places (page checks in
+    RemoveUnwantedContentInternal, and the separate font scan in ReportInvalidFontsAsync), and matching
+    the shorter string would let the fonts path spend both captures on a moment we are not asking about.
 
 .PARAMETER HungAfterSeconds
-    How long a Bloom.exe must have been alive before we treat it as stuck. Healthy visual-regression
+    How long a Bloom.exe must have been alive before we treat it as hung. Healthy visual-regression
     runs launch Bloom, do their work, and kill it inside about 5 minutes; the hangs keep it alive for
     over 20. The default of 8 minutes sits clearly between the two.
 
-.PARAMETER PollSeconds
-    How often to sample. Sampling is cheap (a process listing), and the samples themselves are useful:
-    a starved-but-alive server and a spinning renderer look quite different in CPU over time.
+.PARAMETER EarlyPollSeconds
+    Sampling interval while Bloom is young, i.e. while the trigger is still plausibly ahead of us. The
+    observed failure came ~26 seconds after launch, so 30-second polling could miss the window
+    entirely. Cheap: a process listing plus a log read, no stack walk.
 
-.PARAMETER Captures
-    How many stack captures to take, spaced CaptureGapSeconds apart. More than one matters: identical
-    stacks minutes apart prove it is genuinely stuck rather than crawling.
+.PARAMETER ProcessName
+    Overridable so the script can be exercised locally against any .NET process.
+
+.PARAMETER LogPath
+    Overridable for the same reason. When empty, the usual Bloom log locations are probed.
 #>
 param(
+    [string] $TriggerPattern = "Failed to navigate fully to RemoveUnwantedContentInternal",
+    [int] $TriggerCaptures = 2,
+    [int] $TriggerGapSeconds = 15,
     [int] $HungAfterSeconds = 480,
-    [int] $PollSeconds = 30,
     [int] $Captures = 3,
     [int] $CaptureGapSeconds = 240,
+    [int] $EarlyPollSeconds = 5,
+    [int] $EarlyWindowSeconds = 150,
+    [int] $PollSeconds = 30,
     [int] $GiveUpAfterMinutes = 75,
+    [string] $ProcessName = "Bloom",
+    [string] $LogPath = "",
     [string] $OutputDir = "bloom-watchdog"
 )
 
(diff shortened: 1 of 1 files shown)


build/ci/watch-for-hung-bloom.ps1
```diff
@@ -57,57 +87,183 @@ function Write-Sample([string] $text) {
...
+        # Decide once per poll, not once per process, so the decision cannot depend on which process
+        # we happen to be looking at.
+        $triggerMatched = ($newLogText -match $TriggerPattern)
+        if ($triggerMatched -and -not $triggerSeen) {
+            $triggerSeen = $true
+            $nextTriggerCaptureAt = Get-Date
+            Write-Sample "TRIGGER matched in log: /$TriggerPattern/ -- this is the moment that decides causation"
+        }
+        # $interval is reset to the slow default at the top of every poll, and the early-window rule
+        # below stops applying once the trigger has fired -- so without this the follow-up capture would
```
andrew-polk, line 218: [Devin] **Bug**: Second follow-up snapshot is taken far later than the intended few seconds

Devin found that the pair of trigger snapshots would not actually be a few seconds apart. `$interval` resets to the slow default at the top of every poll, and the early-window rule that otherwise keeps it brisk stops applying once the trigger has fired -- so nothing held the loop short between the two captures, and in CI they would have landed about 35 seconds apart instead of the configured 15.

That matters because the entire point of taking two is to compare closely-spaced snapshots and see whether the stacks are identical, i.e. whether it is genuinely stuck rather than crawling.

build/ci/watch-for-hung-bloom.ps1
```diff
@@ -57,57 +87,183 @@ function Write-Sample([string] $text) {
...
+        try {
+            $reader = New-Object System.IO.StreamReader($stream)
+            try { return $reader.ReadToEnd() } finally { $reader.Dispose() }
+        } finally { $stream.Dispose() }
+    } catch {
+        return ""
+    }
+}
+
+# Where Bloom writes its log. The visual-regression suite launches Bloom with no special logging
```
andrew-polk, line 115: [Devin] **Bug**: New helper function added without the required explanatory comment

`Find-BloomLog` was added with no comment, unlike `Read-LogText` and `Capture-Stacks` beside it, against this repo's rule in AGENTS.md ("All public methods should have a comment. So should most private ones!"). A reader gets no explanation of why those particular log locations are probed or what an empty result means.

build/ci/watch-for-hung-bloom.ps1
```diff
@@ -57,57 +87,183 @@ function Write-Sample([string] $text) {
...
+        if ($logFile -ne "") {
+            if (-not $loggedLogPath) {
+                Write-Sample "watching Bloom log: $logFile"
+                $loggedLogPath = $true
+            }
+            $logText = Read-LogText $logFile
+            if ($logText.Length -ne $lastLogSize) {
+                Write-Sample "  Bloom log now $($logText.Length) bytes"
+                $lastLogSize = $logText.Length
+            }
```
andrew-polk, line 191: [Devin] **Investigate**: Trigger matches anywhere in the log, including stale content from an earlier Bloom launch

Devin noted the trigger searched the whole log text every poll rather than only newly-appended text. If a `Failed to navigate fully` line is already present -- from an earlier step in the job, or an earlier Bloom launch, which is entirely possible since the pre-fix code logs that line and carries on -- the condition stays permanently true and both captures get spent on whichever process happens to be seen next, not the one that just failed.

build/ci/watch-for-hung-bloom.ps1
andrew-polk: (Claude Opus 5) Correct, and a good catch. Fixed in `ca953432ab`: the loop now stays on the brisk interval while a trigger capture is still owed.

Worth recording why my own testing missed this -- I had run with `-PollSeconds 4`, which masked the reset entirely. Retested with a production-like `-PollSeconds 30` and `-TriggerGapSeconds 5`, where the two captures now land 6.7s apart:

```
14:14:28.09  CAPTURE [trigger 1] ...
14:14:34.78  CAPTURE [trigger 2] ...
```

Resolving.

build/ci/watch-for-hung-bloom.ps1
andrew-polk: (Claude Opus 5) Fair -- fixed in `ca953432ab`. `Find-BloomLog` now explains why those particular locations are probed (the order the nightly's own diagnostics step uses) and that an empty return means Bloom simply has not started writing yet, which the caller treats as "no trigger", never as an error.

Resolving.

build/ci/watch-for-hung-bloom.ps1
andrew-polk: (Claude Opus 5) This was the sharpest of the three, and it is a real hazard rather than a hypothetical: the pre-fix code logs `Failed to navigate fully` and carries on, so a run where the navigation stumbles but recovers leaves exactly that line behind. A later genuine failure would then have both captures already spent, aimed at the wrong moment.

Fixed in `ca953432ab` by making the trigger edge-sensitive: the log's length is baselined when the watch starts and only text appended after that is searched, with a rebaseline if the file shrinks (Bloom recreating it rather than appending). Verified both directions -- a log pre-seeded with the trigger line yields zero captures:

```
ignoring the 99 bytes already in the log; only new lines can trigger
watchdog finished ...; trigger captures: 0
```

while a line appended after the watch starts still fires it. Resolving.

build/ci/watch-for-hung-bloom.ps1
```diff
@@ -57,57 +90,193 @@ function Write-Sample([string] $text) {
...
+# old lines fire the trigger, spending the run's only captures on the wrong moment.
+function Read-LogText([string] $path) {
+    try {
+        if (-not (Test-Path $path)) { return $null }
+        $stream = [System.IO.File]::Open(
+            $path,
+            [System.IO.FileMode]::Open,
+            [System.IO.FileAccess]::Read,
+            [System.IO.FileShare]::ReadWrite)
+        try {
```
andrew-polk, line 106: [Devin] **Bug**: A momentary failure to read Bloom's log makes old log lines look brand new, so the diagnostic snapshots can be spent on the wrong moment

`Read-LogText` returned `""` both for "genuinely empty" and for "could not read it this instant", and the caller could not tell those apart. Since `Find-BloomLog` only returns paths that exist, a transient read failure looked like a zero-byte log -- which tripped the shrink check, reset the stale-line baseline to 0, and let the entire pre-existing log count as newly appended. Old lines could then fire the trigger and spend the run's only two captures on the wrong moment.

build/ci/watch-for-hung-bloom.ps1
```diff
@@ -1,46 +1,79 @@
...
+    entirely. Cheap: a process listing plus a log read, no stack walk.
+
+.PARAMETER ProcessName
+    Overridable so the script can be exercised locally against any .NET process.
 
-.PARAMETER Captures
-    How many stack captures to take, spaced CaptureGapSeconds apart. More than one matters: identical
-    stacks minutes apart prove it is genuinely stuck rather than crawling.
+.PARAMETER LogPath
+    Overridable for the same reason. When empty, the usual Bloom log locations are probed.
```
andrew-polk, line 62: [Devin] **Investigate**: The log message the trigger matches exists in two code paths

`Failed to navigate fully` is logged from two places -- page checks in `RemoveUnwantedContentInternal` (PublishHelper.cs:442) and the separate font scan in `ReportInvalidFontsAsync` (PublishHelper.cs:2106). The watchdog could not distinguish them, so if the fonts path logged first both captures would be spent on a moment we are not asking about.

build/ci/watch-for-hung-bloom.ps1
andrew-polk: (Claude Opus 5) Right, and this one was a defect in the fix from the previous commit rather than in the original -- good catch. Fixed in `dda445628a`: `Read-LogText` now returns `$null` for an unreadable log, distinctly from `""` for an empty one, and a poll that cannot read it leaves both the baseline and the trigger untouched rather than concluding anything from an absence we know to be unreliable.

Resolving.

build/ci/watch-for-hung-bloom.ps1
andrew-polk: (Claude Opus 5) Confirmed against master before acting on it -- both call sites are there:

```
441:  Debug.WriteLine("Failed to navigate fully to RemoveUnwantedContentInternal DOM");
442:  Logger.WriteEvent("Failed to navigate fully to RemoveUnwantedContentInternal DOM");
2104: Debug.WriteLine("Failed to navigate fully to ReportInvalidFontsAsync DOM");
2106:   "Failed to navigate fully to ReportInvalidFontsAsync DOM"
```

Fixed in `dda445628a` by including the DOM name in the default pattern, so it targets page checks only. Verified: a log containing only the fonts-path line yields zero captures; the page-checks line then yields two.

Resolving.

build/ci/watch-for-hung-bloom.ps1
```diff
@@ -57,57 +90,197 @@ function Write-Sample([string] $text) {
...
-                # thread — so it still answers while the app's own threads are deadlocked, which is
-                # exactly the case we are here for.
-                & dotnet-stack report --process-id $bloom.Id *> $stackFile
-                Write-Sample "capture $capturesTaken finished (exit=$LASTEXITCODE, $((Get-Item $stackFile -ErrorAction SilentlyContinue).Length) bytes)"
-            } catch {
-                Write-Sample "capture $capturesTaken FAILED: $($_.Exception.Message)"
+            # --- moment 1: the navigation just gave up ---
+            # Again the latch, not the per-poll match: an unreadable log or a recreated one must not
+            # cancel the confirming second snapshot, which is the whole comparison this exists to make.
+            if ($triggerSeen -and $triggerShotsTaken -lt $TriggerCaptures `
```
andrew-polk, line 261: [Devin] **Bug**: The confirming second stack snapshot can be skipped entirely

The follow-up snapshot was gated on the alert line still being visible in the freshly-read log every poll (`$triggerMatched`) rather than on the already-latched fact that the alert fired (`$triggerSeen`). So a momentary unreadable log, or one Bloom recreated after the failure, cancels it -- and the run ends with a single snapshot, which cannot show whether the state was momentary or truly stuck. That comparison is the whole reason two are taken.

andrew-polk: (Claude Opus 5) Correct, including the recommended fix -- both the capture condition and the brisk-interval line now key off the latched `$triggerSeen`, while the latch itself still keys off the per-poll match (that is the edge being detected). Fixed in `c1b0d4eff6`.

Verified against exactly your scenario: trigger, first capture, then recreate the log with no alert line in it.

```
14:29:33  TRIGGER matched ...
14:29:33  CAPTURE [trigger 1] ...
   (log recreated without the alert line)
14:29:40  CAPTURE [trigger 2] ...
```

Resolving.

andrew-polk:
(Claude Opus 5) Consulted Devin on 2026-08-03 up to commit `c1b0d4eff6`.

Devin ran four times across this branch and found **four bugs and two Investigate flags — every one of them real, and all now fixed and resolved**. This was an unusually productive review for a 200-line CI script, so it is worth recording what it caught:

- **The follow-up snapshot landed ~35s late instead of the configured 15s.** `$interval` resets each poll and the early-window rule stops applying once the trigger fires, so nothing kept the loop brisk between the pair. My own testing had missed it because I ran with `-PollSeconds 4`, masking the reset.
- **The trigger matched stale log content**, including a `Failed to navigate fully` left by an earlier Bloom launch — which the pre-fix code makes likely, since it logs that line and carries on. Now edge-sensitive.
- **A momentary read failure collapsed that baseline**, because `Read-LogText` returned `""` for both "empty" and "unreadable". Now returns `$null` for unreadable, and such a poll changes nothing.
- **The trigger matched a second, unrelated code path** — `ReportInvalidFontsAsync` also logs `Failed to navigate fully` (PublishHelper.cs:2106). The pattern now includes the DOM name.
- **The confirming second snapshot could be cancelled entirely**, because the capture was gated on the per-poll match rather than the latched fact that the alert fired. Both sites now use the latch.
- Plus `Find-BloomLog` missing its comment, against this repo's commenting rule.

Each fix was verified by driving the script against a real .NET process, including the specific failure scenarios: a log pre-seeded with the alert yields zero captures; a fonts-path line yields zero; a log recreated after the first capture still yields the second.

Devin's final pass re-lists all six titles, but they are stale — it still quotes `$triggerMatched` at line 257 where the code now reads `$triggerSeen`. Its results appear cumulative per PR rather than per-commit.

CI green (`pr-automation`, `track`). CodeRabbit is disabled for this repo by `.coderabbit.yml`.

PR #8144, opened 2026-08-03
Way to turn off indenting & don't insert <p> before headers (BL-16649)

hatton:
Two changes for BL-16649, one per problem on the card.

### Problem 1: you don't always want indenting

When a paragraph style has a first-line indent and the author manually splits a paragraph
across two pages, the continuation paragraph at the top of the second page gets an indent it
should not have.

Right-clicking a paragraph of an ordinary text box in the edit view now opens a small context
menu. Plain right-click there was previously unused (the C# side strips the native WebView2
items), ctrl+right-click still reaches the developer menu, and text inside a canvas element
keeps the canvas element's own menu.

The menu's one item, **No Indent**, toggles a `bloom-noIndent` class on that one paragraph. It
shows a check when on, and is disabled when the paragraph has no indent for it to remove.

The class is cancelled by a rule in `basePage-sharedRules.less`. It has to out-specify two
things: the `.foo-style > p {text-indent: … !important}` rule the Format dialog writes into
userModifiedStyles (0-1-1), and any hand-written rule in `customBookStyles.css` /
`customCollectionStyles.css` — which we cannot win on source order, because
`StyleSheetLinkSorter` deliberately sorts `basePage` *before* those files. So the selector is
`.bloom-page .bloom-editable p.bloom-noIndent` (0-3-1) rather than the shorter 0-2-1 form, which
a custom `.bloom-page .foo-style p` (equally specific, loaded later) would have beaten. The rule
is copied into `basePage-legacy-5-6.less`, which does not import the shared rules;
`baseEPUB.less` does import them, so it works in published ePubs too.

Note that the rule deliberately leaves `margin-left` alone. For a hanging-indent style (which
is `text-indent: -20pt` plus `margin-left: 20pt`) that makes a marked paragraph line up with
the body lines of a hanging paragraph, which is what a continuation should do; for an ordinary
first-line indent it sits flush.

Also here:

- `LocalizableMenuItem` gains a `checked` prop, which puts the check at the *end* of the row so
  the leading icon slot stays free for the command's own icon.
- Pressing Enter inside a marked paragraph no longer gives the new paragraph the class:
  ckeditor builds the second half by cloning the paragraph it split, so `BloomField` now strips
  the class from whichever paragraphs the split created.

### Problem 2: you don't always want to start with a `<p>`

The earlier commit on this branch: don't auto-insert a paragraph above a heading.

### Testing

`noIndentSpec.ts` covers the click-targeting (including that clicks inside canvas elements are
ignored) and the enabled/checked/toggle logic, including the hanging-indent and
already-marked cases.

Ref: https://issues.bloomlibrary.org/youtrack/issue/BL-16649

[Devin review](https://devinreview.com/BloomBooks/BloomDesktop/pull/8144)

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8144)

---
[Devin review](https://app.devin.ai/review/BloomBooks/BloomDesktop/pull/8144)

### DIFF
--- a/src/BloomBrowserUI/bookEdit/bloomField/BloomField.ts
+++ b/src/BloomBrowserUI/bookEdit/bloomField/BloomField.ts
@@ -9,6 +9,7 @@ import { EditableDivUtils } from "../js/editableDivUtils";
 import $ from "jquery";
 import { showLinkTargetChooserDialog } from "../../react_components/LinkTargetChooser/LinkTargetChooserDialogLauncher";
 import { getLocalization } from "../../react_components/l10n";
+import { kNoIndentClass } from "../textContextMenu/noIndent";
 
 // This class is actually just a group of static functions with a single public method. It does whatever we need to to make Firefox's contenteditable
 // element have the behavior we need.
@@ -258,11 +259,36 @@ export default class BloomField {
                 }
             });
         }
+        // BL-16649: "No Indent" (see the text context menu) marks one
+        // paragraph as the continuation of a paragraph on the previous page. Pressing Enter
+        // in such a paragraph makes a genuinely new paragraph, which should indent normally
+        // -- but ckeditor builds it by shallow-cloning the paragraph it split, so it would
+        // inherit the class. We note which paragraphs existed just before the Enter, then
+        // take the class off any that the split created. Only paragraphs are considered:
+        // the class is only ever put on a <p>, never on a heading (see kBlockElementSelector).
+        let paragraphsBeforeEnter: HTMLParagraphElement[] | undefined;
         ckeditor.on("key", (event) => {
             if (event.data.keyCode === CKEDITOR.SHIFT + 13) {
                 BloomField.InsertLineBreak();
                 event.cancel();
+            } else if (event.data.keyCode === 13) {
+                paragraphsBeforeEnter = Array.from(
+                    bloomEditableDiv.getElementsByTagName("p"),
+                );
             }
+        });
+        ckeditor.on("afterCommandExec", (event) => {
+            if (event.data.name !== "enter") return;
+            const paragraphsBefore = paragraphsBeforeEnter;
+            paragraphsBeforeEnter = undefined;
+            if (!paragraphsBefore) return;
+            Array.from(bloomEditableDiv.getElementsByTagName("p")).forEach(
+                (paragraph) => {
+                    if (!paragraphsBefore.includes(paragraph)) {
+                        paragraph.classList.remove(kNoIndentClass);
+                    }
+                },
+            );
         });
         ckeditor.on("selectionChange", () => {
             this.EnsureCaretNotInsideLineBreakSpan();
--- a/src/BloomBrowserUI/bookEdit/bloomField/BloomField.ts
+++ b/src/BloomBrowserUI/bookEdit/bloomField/BloomField.ts
@@ -28,6 +28,14 @@ import { getLocalization } from "../../react_components/l10n";
 // Next, we have to keep you from accidentally losing the image placeholder when you do ctrl+a DEL. We prevent this deletion
 // for any element marked with a 'bloom-preventRemoval' class.
 
+// The block-level elements we tolerate as direct children of a bloom-editable. Bloom's own
+// editing UI only ever makes paragraphs, but content converted from other formats (e.g. by
+// BloomBridge) can legitimately arrive with real heading elements, and a box that leads with
+// one must not have an empty paragraph pushed in above it. Several places below ask "which
+// block am I in?" or "does this box have any blocks yet?"; they all consult this rather than
+// looking for a <p> specifically, so a heading counts as a block everywhere or nowhere.
+const kBlockElementSelector = "p,h1,h2,h3,h4,h5,h6";
+
 export default class BloomField {
     public static ManageField(bloomEditableDiv: HTMLElement) {
         BloomField.PreventRemovalOfSomeElements(bloomEditableDiv);
@@ -657,7 +665,7 @@ export default class BloomField {
                     //see if it is one those. Anything marked with bloom-preventRemoval is probably not something we want to
                     //be merging with.
                     const previousElement = $(sel.anchorNode)
-                        .closest("P")
+                        .closest(kBlockElementSelector)
                         .prev();
                     if (
                         previousElement.length > 0 &&
@@ -693,10 +701,24 @@ export default class BloomField {
         });
     }
 
+    // True if the element is one of the block elements we allow as a direct child of a
+    // bloom-editable, i.e. a paragraph or a heading.
+    private static IsBlockElement(element: Element | undefined): boolean {
+        return (
+            !!element &&
+            kBlockElementSelector
+                .split(",")
+                .includes(element.tagName.toLowerCase())
+        );
+    }
+
     private static EnsureStartsWithParagraphElement(field: HTMLElement) {
+        // A heading counts: a box that starts with <h1> already starts with a block, and
+        // prepending an empty paragraph above it would give the reader a blank first line
+        // (and, once the page is saved, keep it).
         if (
             $(field).children().length > 0 &&
-            $(field).children().first().prop("tagName").toLowerCase() === "p"
+            BloomField.IsBlockElement($(field).children().first()[0])
         ) {
             return;
         }
@@ -705,6 +727,10 @@ export default class BloomField {
 
     private static EnsureEndsWithParagraphElement(field: HTMLElement) {
         //Enhance: move any errant paragraphs to after the bloom-canvas
+        // Deliberately still requires a paragraph rather than accepting any block: this is
+        // only used for the embedded-image case below, where the trailing paragraph is there
+        // to give the user somewhere to type text that wraps around the image. A heading is
+        // not that.
         if (
             $(field).children().length > 0 &&
             $(field).children().last().prop("tagName").toLowerCase() === "p"
@@ -773,10 +799,13 @@ export default class BloomField {
         position: CursorPosition,
     ) {
         const range = document.createRange();
+        // Any block will do, and it has to be any block: a box holding only a heading has no
+        // paragraph, and selectNodeContents(undefined) throws.
+        const blocks = $(field).find(kBlockElementSelector);
         if (position === CursorPosition.start) {
-            range.selectNodeContents($(field).find("p").first()[0]);
+            range.selectNodeContents(blocks.first()[0]);
         } else {
-            range.selectNodeContents($(field).find("p").last()[0]);
+            range.selectNodeContents(blocks.last()[0]);
         }
         range.collapse(position === CursorPosition.start); //true puts it at the start
         const sel = window.getSelection();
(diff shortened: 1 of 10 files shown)


hatton (commented):
> @hatton reviewed 10 files.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8144)* status: all files reviewed (commit messages unreviewed), all discussions resolved (waiting on hatton).

src/content/bookLayout/basePage-sharedRules.less
```diff
@@ -190,6 +190,26 @@ span.bloom-linebreak {
...
+// hand-written rules in customBookStyles/customCollectionStyles, which we cannot win on
+// source order: StyleSheetLinkSorter puts basePage *before* those, so among equally
+// specific !important rules theirs would win. Hence the deliberately long selector
+// (0-3-1) rather than just `.bloom-page p.bloom-noIndent` (0-2-1), which a custom
+// `.bloom-page .mystyle-style p` (also 0-2-1, but loaded later) would have beaten.
+// The marked paragraph is always inside a .bloom-editable -- that is the only place the
+// command is offered -- so requiring it here costs us no matches.
+//
+// NB: there is a copy of this rule in basePage-legacy-5-6.less, which does not import this file.
+.bloom-page .bloom-editable p.bloom-noIndent {
```
hatton, line 209: [Devin] **Investigate**: The new rule can lose to book/template stylesheets more specific than the Format dialog's rules

The specificity argument in the original commit message only covered the rule the Format dialog writes (`.mystyle-style > p`, 0-1-1), which the rule did beat. It would not have beaten a hand-written rule in `customBookStyles.css` / `customCollectionStyles.css` such as `.bloom-page .mystyle-style p {text-indent: 20pt !important}`, which is *equally* specific (0-2-1) and, because `StyleSheetLinkSorter` deliberately sorts `basePage` **before** those files, wins on source order. In a book with custom CSS like that, "No Indent" would have appeared enabled and done nothing visible.

hatton: [Claude Opus 5 (1M context) from Hatton's machine during preflight] Fixed in `e2ccc89739`: confirmed the concern and made the rule require the `.bloom-editable` ancestor as well, so it is now 0-3-1 and beats the equally-specific-but-later custom rule. The command is only ever offered on a paragraph inside a `.bloom-editable`, so the longer selector loses no matches. Worth acting on rather than waving off, because the converted books this feature was built for are exactly the ones likely to carry hand-written CSS. Resolving.

hatton (commented):
> @hatton reviewed 2 files.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8144)* status: all files reviewed (commit messages unreviewed), 1 unresolved discussion.

src/BloomBrowserUI/bookEdit/bloomField/BloomField.ts
```diff
@@ -693,10 +727,24 @@ export default class BloomField {
...
+    }
+
     private static EnsureStartsWithParagraphElement(field: HTMLElement) {
+        // A heading counts: a box that starts with <h1> already starts with a block, and
+        // prepending an empty paragraph above it would give the reader a blank first line
+        // (and, once the page is saved, keep it).
         if (
             $(field).children().length > 0 &&
-            $(field).children().first().prop("tagName").toLowerCase() === "p"
+            BloomField.IsBlockElement($(field).children().first()[0])
```
hatton, line 747: [Devin] **Investigate**: Heading-only text boxes now have no `<p>`, which some downstream code assumes exists

By accepting a heading as a block, `EnsureStartsWithParagraphElement` no longer guarantees that every bloom-editable contains at least one `<p>`. `MoveCursorToEdgeOfField` was updated for this, but `StyleEditor` still assumes a paragraph exists: it does `box.find("p")` and reads `margin-bottom` / `text-indent` off it to pick the Format dialog's initial paragraph-spacing and indent values. With an empty jQuery set those come back undefined and parse to NaN. The pre-existing comment there even says "Trying to do so will be a problem if there are no paragraph boxes in the div. I don't think this should happen." — this branch is what makes it happen.

src/BloomBrowserUI/bookEdit/textContextMenu/noIndent.ts
```diff
@@ -0,0 +1,52 @@
...
+    const indent =
+        paragraph.ownerDocument.defaultView?.getComputedStyle(
+            paragraph,
+        ).textIndent;
+    // A computed text-indent is an absolute length ("0px", "26.6667px") or a percentage.
+    return !!indent && parseFloat(indent) !== 0;
+}
+
+/** Turn "No Indent" on or off for this one paragraph. */
+export function toggleNoIndent(paragraph: HTMLElement): void {
```
hatton, line 48: [Devin] **Investigate**: Toggling the class does not mark the page dirty or participate in undo

`toggleNoIndent` only mutates the DOM and relies on the page being posted back later. Two consequences to weigh: the toggle is not part of ckeditor's undo stack, so ctrl+Z will not reverse it; and if Bloom ever skipped saving a page it considered unmodified, the toggle could be lost.

src/BloomBrowserUI/bookEdit/textContextMenu/TextContextMenu.tsx
```diff
@@ -0,0 +1,119 @@
...
+    const handleNoIndentClick = () => {
+        toggleNoIndent(props.paragraph);
+        props.setOpen(false);
+    };
+
+    return (
+        <ThemeProvider theme={lightTheme}>
+            <Menu
+                open={props.open}
+                anchorReference="anchorPosition"
```
hatton, line 41: [Devin] **Investigate**: MUI Modal side effects on the page document while the menu is open

The Menu renders into the page iframe's document, so MUI's ModalManager applies its usual side effects there while open: `overflow: hidden` plus scrollbar padding on `body`, and `aria-hidden="true"` on sibling body children, which includes `div.bloom-page`. They are reverted on close, but a page save triggered while the menu was open could capture the `aria-hidden`. `disableScrollLock` would remove the body half. Devin notes this is the same pattern `CanvasElementContextControls` already uses, so it is not new behavior for this PR.

src/BloomBrowserUI/bookEdit/textContextMenu/noIndent.ts
hatton: [Claude Opus 5 (1M context) from Hatton's machine during preflight] Half of this is refuted, half is real, so leaving the thread open for the developer.

**Not saved: refuted.** I traced the save path. `EditingStateMachine` calls `RequestBrowserToSave` at its save points with no "is it dirty?" condition, and `UpdateBookDomFromBrowserPageContent` then applies whatever the browser posted unconditionally. The dirty-ish logic that does exist (`NeedToDoFullSave`) only chooses between a full-book save and a page-only save — it never skips the page markup. So the class does get written to the book.

**Undo: real.** ctrl+Z will not reverse a No Indent toggle. It is the same limitation the canvas element menu commands already live with, and clicking the item again does undo it, so it may well be acceptable — but that is the developer's call, so this is in the preflight decision report rather than being resolved here.

src/BloomBrowserUI/bookEdit/textContextMenu/TextContextMenu.tsx
hatton: [Claude Opus 5 (1M context) from Hatton's machine during preflight] Left open as a small optional hardening rather than resolved.

Devin is right that this is pre-existing behavior — `CanvasElementContextControls` opens a MUI Menu in the same document the same way — so this PR does not introduce it. The save-while-open path is also hard to hit in practice: the MUI backdrop swallows the first click, so clicking a page thumbnail closes the menu instead of triggering a save. Adding `disableScrollLock` is a one-line change that removes the body-mutation half; whether it is worth touching is in the decision report.

hatton:
[Claude Opus 5 (1M context) from Hatton's machine during preflight] Consulted Devin on 2026-08-03 up to commit `4531bdf512`.

Four review passes ran (one per commit). Across all of them Devin reported **no Bugs**. Of its flagged items:

- **Fixed during the run:** the `No Indent` CSS rule could have lost to a hand-written rule in `customBookStyles.css` / `customCollectionStyles.css`. That was correct and is now fixed in `e2ccc89739` — see the resolved thread on `basePage-sharedRules.less`. Devin restated the same flag on later passes with its original pre-fix wording, so treat those repeats as stale.
- **Also fixed:** a code comment that described the page-save mechanism inaccurately (`cbc1259f71`).
- **Open for the developer** (three threads, each with our assessment as a reply): heading-only text boxes leaving `StyleEditor` without a `<p>` to read paragraph defaults from; the toggle not participating in ckeditor undo (its "might not be saved" half is refuted in the thread); and MUI Modal side effects on the page document, which pre-date this PR.
- **14 informational items** on the final pass — not mirrored, per usual practice. Two of them were worth acting on and were folded into the fixes above.

CI is green (`pr-automation`, `track`). CodeRabbit is configured on this repo but with `auto_review: enabled: false`, so it does not post; no Greptile configuration or activity. `code-review/reviewable` stays pending by design until a human reviews.

src/BloomBrowserUI/bookEdit/bloomField/BloomField.ts
hatton: [Claude Opus 5 (1M context) from Hatton's machine during preflight] Decided: leave as is. Hatton reviewed this in the preflight report and chose not to harden StyleEditor here. The symptom is a blank paragraph-spacing setting in the Format dialog for a text box that holds only a heading — nothing is corrupted, nothing bad is written, and picking a value works normally. Heading-only boxes only arise in converted content. Resolving.

src/BloomBrowserUI/bookEdit/textContextMenu/noIndent.ts
hatton: [Claude Opus 5 (1M context) from Hatton's machine during preflight] Decided: leave as is. Hatton reviewed this in the preflight report and accepted the undo limitation. Choosing the menu item again reverses the toggle, and the canvas element menu commands have long behaved the same way, so wiring this one command into ckeditor's undo machinery is not worth the complexity. (The other half of this flag — that the toggle might not be saved — was already refuted above: page content is applied unconditionally.) Resolving.

src/BloomBrowserUI/bookEdit/textContextMenu/TextContextMenu.tsx
hatton: [Claude Opus 5 (1M context) from Hatton's machine during preflight] Decided: leave as is. Hatton reviewed this in the preflight report and chose not to add `disableScrollLock`. This behavior pre-dates the PR — `CanvasElementContextControls` opens a MUI Menu in the same document the same way — and changing it only here would diverge from those menus for no user-visible gain. Resolving.

StephenMcConnel (approved):
> @StephenMcConnel reviewed 10 files and all commit messages, and resolved 4 discussions.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8144)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on hatton).

PR #8146, opened 2026-08-04
Run AI-editor images through Bloom's import processing (BL-16645)

StephenMcConnel:
When the AI Image Editor committed an image back into the book, `AiImageEditorApi` did a plain `RobustFile.Copy` into the book folder, so none of the processing a user-added image gets was applied. AI services can return very large PNGs, and an unprocessed one bloats the book folder for every sync, upload and backup.

This routes the commit through `ImageUtils.ProcessAndSaveImageIntoFolder` (the same path the image-gallery and paste flows already use), via a new `ImportImageIntoBookFolder` helper.

Two things the helper has to protect:

- **The `ai-image` file name.** `DeleteSupersededAiImageFiles` reclaims the files we generate by that prefix, but `ProcessAndSaveImageIntoFolder` names its output after the *source* file, which for us is an opaque history id. So we let it write and then rename onto the name we reserved. `isSameFile: false` guarantees it wrote a brand-new file, so the rename can never disturb a file another slot shares. The reserved name takes its extension from what was actually produced, since processing can turn a `.jpeg` into `.jpg`.
- **The image itself.** Anything we can't process is copied verbatim rather than lost: `.webp` (PalasoImage decodes through GDI+, which has no WebP codec), a placeholder (`ProcessAndSaveImageIntoFolder` deliberately writes nothing for one), and images that fail to process at all.

Adds tests for the naming invariant, which was previously untested — nothing in the suite went through this path, so a silent rename would not have been caught.

### Known gaps (deliberate, flagged for review)

- A `.webp` result is still copied verbatim, so an oversized `.webp` is not downscaled. Bloom has no WebP decoder.
- A *reused* existing book image now also goes through processing. In practice that's a straight copy (it was already import-processed on its own way in), but if one were oversized enough to be resized, its embedded credits could be lost — the reuse path deliberately does not re-embed credits.

Ref: https://issues.bloomlibrary.org/youtrack/issue/BL-16645

[Devin review](https://devinreview.com/BloomBooks/BloomDesktop/pull/8146)

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8146)

### DIFF
--- a/src/BloomExe/web/controllers/AiImageEditorApi.cs
+++ b/src/BloomExe/web/controllers/AiImageEditorApi.cs
@@ -16,6 +16,7 @@ using SIL.IO;
 using SIL.Progress;
 using SIL.Reporting;
 using SIL.Windows.Forms.ClearShare;
+using SIL.Windows.Forms.ImageToolbox;
 
 namespace Bloom.web.controllers
 {
@@ -111,6 +112,20 @@ namespace Bloom.web.controllers
             ".jpg",
             ".jpeg",
             ".webp",
+        };
+
+        // The subset of AllowedImageExtensions that Bloom's import processing can actually
+        // open. PalasoImage decodes through GDI+, which has no WebP codec, so handing it a
+        // .webp throws every time — those are copied verbatim instead (see
+        // ImportImageIntoBookFolder). Kept as its own list rather than "everything but webp"
+        // so that adding a format above is a deliberate decision here too.
+        private static readonly HashSet<string> ProcessableImageExtensions = new HashSet<string>(
+            StringComparer.OrdinalIgnoreCase
+        )
+        {
+            ".png",
+            ".jpg",
+            ".jpeg",
         };
 
         // Regex alternation of the allowed extensions without the leading dot
@@ -890,8 +905,9 @@ namespace Bloom.web.controllers
 
         /// <summary>
         /// Applies one replacement to the slot named by incomingId ("{pageId}:{ordinal}"),
-        /// copying the new image bytes (from history or a reused book file) into the book
-        /// folder under a fresh name. An off-page slot is edited here in the storage DOM
+        /// bringing the new image bytes (from history or a reused book file) into the book
+        /// folder under a fresh name via <see cref="ImportImageIntoBookFolder"/>, which also
+        /// applies Bloom's normal import processing. An off-page slot is edited here in the storage DOM
         /// (caller saves); a current-page slot is left untouched — we return oldSrc/newSrc and
         /// the new file's <paramref name="creditAttributes"/> (via the out params) for the
         /// front-end to apply to the live DOM. Returns false with <paramref name="error"/> set
@@ -1009,15 +1025,9 @@ namespace Bloom.web.controllers
                 return false;
             }
 
-            // Write the new image under a fresh name (preserving the source extension) so
-            // we never clobber a file shared by another slot or serve stale cached bytes.
-            // GetUnusedFilename guarantees fresh while keeping the name human-readable
-            // (filenames surface in image metadata and problem reports).
-            var extension = Path.GetExtension(sourceBytesPath);
-            if (string.IsNullOrEmpty(extension))
-                extension = ".png";
-            var newFileName = ImageUtils.GetUnusedFilename(book.FolderPath, "ai-image", extension);
-            RobustFile.Copy(sourceBytesPath, Path.Combine(book.FolderPath, newFileName), true);
+            // Bring the bytes into the book folder, import-processed and under a fresh
+            // "ai-image*" name.
+            var newFileName = ImportImageIntoBookFolder(sourceBytesPath, book.FolderPath);
             newSrc = newFileName;
 
             // A generated result arrives with no intellectual-property metadata of its own, so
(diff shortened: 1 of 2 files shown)


src/BloomExe/web/controllers/AiImageEditorApi.cs
```diff
@@ -1009,15 +1025,9 @@ out sourceBytesPath
...
-            // (filenames surface in image metadata and problem reports).
-            var extension = Path.GetExtension(sourceBytesPath);
-            if (string.IsNullOrEmpty(extension))
-                extension = ".png";
-            var newFileName = ImageUtils.GetUnusedFilename(book.FolderPath, "ai-image", extension);
-            RobustFile.Copy(sourceBytesPath, Path.Combine(book.FolderPath, newFileName), true);
+            // Bring the bytes into the book folder, import-processed and under a fresh
+            // "ai-image*" name.
+            var newFileName = ImportImageIntoBookFolder(sourceBytesPath, book.FolderPath);
             newSrc = newFileName;
```
StephenMcConnel, line 1041: [Devin] **Investigate**: Reused existing book images now get re-processed, which can strip embedded credits

The `sourceUrl` (reuse-an-existing-book-image) branch now also flows through `ProcessAndSaveImageIntoFolder`. For an already-import-processed image that is a straight file copy, so metadata survives. But if the reused file is oversized, `TryResizeImageWithGraphicsMagick` produces a new file and the copy is of the resized temp; since `EmbedCreditsInNewImageFile` only runs for `resultId` replacements (`src/BloomExe/web/controllers/AiImageEditorApi.cs:1040-1041`), any credits embedded in the original file could be lost for that slot. The PR description flags this as a deliberate known gap; noting it here so a reviewer can decide whether `resizeFileIfNeeded:false` for the reuse branch would be preferable.

src/BloomTests/web/controllers/AiImageEditorApiTests.cs
```diff
@@ -161,6 +161,143 @@ public void DeleteSupersededAiImageFiles_EmptyCandidates_DoesNothing()
...
+
+            using (var after = Image.FromFile(Path.Combine(_bookFolder.Path, newName)))
+            {
+                Assert.That(
+                    after.Width,
+                    Is.LessThan(5000),
+                    "an oversized image should have been downscaled on the way in"
+                );
+            }
+        }
```
StephenMcConnel, line 253: [Devin] **Investigate**: Downscale test depends on GraphicsMagick being available on the test machine

`ImportImageIntoBookFolder_OversizedImage_IsDownscaled` asserts the width shrinks, which inside `ProcessAndSaveImageIntoFolder` depends on `TryResizeImageWithGraphicsMagick` succeeding (`src/BloomExe/ImageProcessing/ImageUtils.cs:812-823`); if GraphicsMagick is not found the resize is skipped and the original is copied through unchanged, so the test would fail for environmental rather than code reasons. Worth confirming the CI image ships GraphicsMagick the way the other image tests assume.

StephenMcConnel: [Claude Opus 5 (1M context) during preflight] Not an issue — GraphicsMagick is not an environment assumption here, it is part of the build output.

`GetGraphicsMagickPath()` looks for `gm/gm.exe` next to the code base folder, and the build copies it there: it is present at `output/agent/<key>/bin/gm/gm.exe` in the very test-output tree this suite runs from. So any environment that can build `BloomTests` has already produced the binary the test needs — there is no separate provisioning step that could be missing on a build agent.

Verified by running the test: it passes and does observe the downscale.

StephenMcConnel:
[Claude Opus 5 (1M context) from McConnel's machine during preflight] Consulted Devin on 2026-08-04 up to commit `a3d67b040fb4eba6b8925e7951bb976583f96166`.

It found no bugs. Of five flags, two were worth raising and are now threads on this PR:

- Reusing an existing book image could lose its embedded credits if that image is big enough to be resized on the way in — left open, it needs a call on whether the reuse branch should skip resizing.
- The new downscale test looked like it might depend on GraphicsMagick being installed on the build machine — replied and resolved: `gm.exe` is copied into the test output by the build itself, so it is always there.

The other three flags were informational and are not mirrored. CI (`pr-automation`) passed; CodeRabbit's auto-review is turned off for this repo in `.coderabbit.yml`, so it does not run.

src/BloomExe/web/controllers/AiImageEditorApi.cs
StephenMcConnel: [Claude Opus 5 (1M context) during preflight] Fixed in `c719f8e73`, on the developer's decision.

The reuse branch now passes `resizeIfNeeded: false`, so a reused book image is never resized on this path — which removes the only route by which its embedded credits could be lost. The reasoning for preferring this over re-embedding the credits: a reused image was already import-processed on its own way in, so shrinking it a second time gains nothing, and skipping the resize leaves the file byte-identical rather than relying on us to restore metadata after rewriting it.

Only a freshly generated or uploaded result — the one that can actually arrive enormous — is resized now.

Covered by two new tests: a reused image comes through at its original size, and keeps its creator and copyright.

StephenMcConnel:
[Claude Opus 5 (1M context) from McConnel's machine during preflight] Consulted Devin on 2026-08-04 up to commit `8b66b2eb95823222cce3fa9c312b7b33aa5d3919`.

Re-review after the fix commit: no bugs, seven flags, two of them worth raising — but both are the same two findings as the first pass, and both already have a documented, resolved thread on this PR. Nothing new, so nothing re-posted.

Worth noting that Devin re-raised the reused-image credits flag against a commit that already fixes it: it still describes the reuse path as going through the resize step, but as of `c719f8e73` that path passes `resizeIfNeeded: false` and is never resized — which is the very change Devin's own write-up suggested. Treating that one as a stale false positive rather than reopening the thread.

The other five flags were informational and are not mirrored. CI (`pr-automation`) passed.

PR #8147, opened 2026-08-04
Fix "By Sentence" left disabled on a text box with no recording (BL-16632)

andrew-polk:
The Talking Book tool's Recording Mode radio buttons are driven by values derived from whether the
current text box has a recording (`hasAudio`, `haveACurrentTextboxModeRecording`,
`hasRecordableDivs`). Only a full `updateDisplay()` recomputed those, while the buttons (Record,
Play, Clear …) were refreshed every time the current box changed — so the radio buttons went on
describing the box you were on before.

After recording By Whole Text Box and then clicking a box with no recording — or typing into a
newly added page — "By Sentence" stayed disabled, its tooltip telling you to use Clear first, while
the Clear button itself was correctly disabled because the current box has no audio.

**The fix.** The button-state refresh now ends by calling a new, deliberately narrow
`updateAudioStateInDisplay()`, which publishes just those three audio-derived values and changes
nothing else. It is narrow in two ways that matter:

- it does **not** publish the mode fields (`recordingMode`, `inShowPlaybackOrderMode`,
  `showingImageDescriptions`) — those belong to the code that changes those modes, and
  republishing them from a button refresh fights it;
- it never *changes* anything: no highlight is chosen (which would recurse back in via
  `getCurrentTextBox`), and `this.recordingMode` is left alone, so `updateDisplay`'s
  subscription downgrade can't move the mode out from under button statuses just computed.

Two related holes found in review are fixed too:

- `disableInteraction()` published the derived values without refreshing `haveAudio` first, so the
  Advanced section could still report the audio of the box we had just left. Every path that gets
  there has nothing recordable selected, so it now says so.
- Deleting the last of the text on a page disabled every button but left the Recording Mode radios
  and Insert Segment Marker enabled, still describing text that was gone.

**Tests.** Five regression tests in `audioRecordingSpec.ts`, each confirmed to fail without its
fix: the highlight moving to an unrecorded box; typing into a page that started out empty;
deleting the last of the text; clicking something that cannot be recorded; and (on the branch
below) a stale server answer arriving after a page change.

**Not in this PR:** the related race — a `checkForAnyRecording` answer about the *previous* box
arriving after the page changed — which is why the card's own repro steps ("record By Whole Text
Box, add a new page, type") only sometimes show the problem. That is a separate, pre-existing
defect; the guard for it (each refresh takes a generation number; only the most recent may apply
what it hears back) is held on the branch `BL-16632-stale-button-state-answers` pending a decision
about whether it ships with this regression fix.

Not verified live in Bloom: reproducing by hand needs an actual microphone recording. QA test ideas
are on the card.

Ref: https://issues.bloomlibrary.org/youtrack/issue/BL-16632

---
[Devin review](https://devinreview.com/BloomBooks/BloomDesktop/pull/8147)

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8147)

### DIFF
--- a/src/BloomBrowserUI/bookEdit/toolbox/talkingBook/audioRecording.ts
+++ b/src/BloomBrowserUI/bookEdit/toolbox/talkingBook/audioRecording.ts
@@ -4126,6 +4126,19 @@ export default class AudioRecording implements IAudioRecorder {
         );
 
         this.updateButtonStateHelper(expectedVerb, response);
+
+        // The helper just recomputed this.haveAudio from the server, for whatever element is
+        // current now. The Advanced section's Recording Mode radio buttons are driven by values
+        // derived from it (uiState.hasAudio and uiState.haveACurrentTextboxModeRecording), and
+        // only updateDisplay() recomputes those. Several paths that change which box is current
+        // (clicking in another text box, Next/Back, typing new text) refresh the buttons through
+        // here without otherwise calling updateDisplay, which left the radio buttons describing
+        // the *previous* box. That is BL-16632: after recording By Whole Text Box and then
+        // clicking a box with no recording, "By Sentence" stayed disabled, saying to use Clear
+        // first, even though Clear itself was (correctly) disabled because that box has no audio.
+        // Pass false because refreshing what the UI shows must not have the side effect of
+        // choosing a new current element.
+        this.updateDisplay(false);
     }
 
     private updateButtonStateHelper(
--- a/src/BloomBrowserUI/bookEdit/toolbox/talkingBook/audioRecording.ts
+++ b/src/BloomBrowserUI/bookEdit/toolbox/talkingBook/audioRecording.ts
@@ -4129,19 +4129,13 @@ export default class AudioRecording implements IAudioRecorder {
 
         // The helper just recomputed this.haveAudio from the server, for whatever element is
         // current now. The Advanced section's Recording Mode radio buttons are driven by values
-        // derived from it (uiState.hasAudio and uiState.haveACurrentTextboxModeRecording), and
-        // only updateDisplay() recomputes those. Several paths that change which box is current
-        // (clicking in another text box, Next/Back, typing new text) refresh the buttons through
-        // here without otherwise calling updateDisplay, which left the radio buttons describing
-        // the *previous* box. That is BL-16632: after recording By Whole Text Box and then
-        // clicking a box with no recording, "By Sentence" stayed disabled, saying to use Clear
-        // first, even though Clear itself was (correctly) disabled because that box has no audio.
-        // Both arguments are false because this refresh must only *report* state, never change
-        // it: choosing a new current element (maySetHighlight) would recurse back into here via
-        // getCurrentTextBox, and switching the recording mode (maySwitchToSentenceMode) would
-        // change it out from under the button statuses we just computed for the mode we read at
-        // the top of this method.
-        this.updateDisplay(false, false);
+        // derived from it, so they have to be recomputed here too. Several paths that change which
+        // box is current (clicking in another text box, Next/Back, typing new text) refresh the
+        // buttons through here and nothing else, which left the radio buttons describing the
+        // *previous* box. That is BL-16632: after recording By Whole Text Box and then clicking a
+        // box with no recording, "By Sentence" stayed disabled, saying to use Clear first, even
+        // though Clear itself was (correctly) disabled because that box has no audio.
+        this.updateAudioStateInDisplay();
     }
 
     private updateButtonStateHelper(
@@ -4500,6 +4494,13 @@ export default class AudioRecording implements IAudioRecorder {
         // 1- When calling changeStateAndSetExpected() with expectedVerb = ""
         // 2- While doing auto segmenting
         // 3- An unrecoverable error has occurred
+        // Every path that gets here has nothing recordable selected: the page has no editable
+        // text, the click landed on something that can't be recorded, or we could not work out
+        // what is current. So the current selection has no audio, by definition. Saying so keeps
+        // the Advanced section honest -- otherwise it goes on reporting the audio of the box we
+        // were on before, which is the BL-16632 mismatch ("By Sentence" disabled telling you to
+        // press Clear, while Clear is disabled because there is nothing to clear).
+        this.haveAudio = false;
         this.setStatus("record", Status.Disabled);
         this.setStatus("play", Status.Disabled);
         this.setStatus("split", Status.Disabled);
@@ -4509,11 +4510,10 @@ export default class AudioRecording implements IAudioRecorder {
         this.setStatus("listen", Status.Disabled);
 
         // The Advanced section's controls (the Recording Mode radio buttons, Insert Segment
-        // Marker) are driven by derived values in uiState that only updateDisplay recomputes, so
-        // they have to be refreshed here too. Otherwise deleting all the text on a page disables
-        // every button but leaves those controls enabled, still describing the text that used to
-        // be there (BL-16632). Report only: never move the highlight or switch the mode from here.
-        this.updateDisplay(false, false);
+        // Marker) are driven by derived values that have to be refreshed here too. Otherwise
+        // deleting all the text on a page disables every button but leaves those controls enabled,
+        // still describing the text that used to be there (BL-16632).
+        this.updateAudioStateInDisplay();
     }
 
     private showBusy(): void {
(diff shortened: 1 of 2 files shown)


src/BloomBrowserUI/bookEdit/toolbox/talkingBook/audioRecording.ts
```diff
@@ -4126,6 +4126,19 @@ export default class AudioRecording implements IAudioRecorder {
...
+        // derived from it (uiState.hasAudio and uiState.haveACurrentTextboxModeRecording), and
+        // only updateDisplay() recomputes those. Several paths that change which box is current
+        // (clicking in another text box, Next/Back, typing new text) refresh the buttons through
+        // here without otherwise calling updateDisplay, which left the radio buttons describing
+        // the *previous* box. That is BL-16632: after recording By Whole Text Box and then
+        // clicking a box with no recording, "By Sentence" stayed disabled, saying to use Clear
+        // first, even though Clear itself was (correctly) disabled because that box has no audio.
+        // Pass false because refreshing what the UI shows must not have the side effect of
+        // choosing a new current element.
+        this.updateDisplay(false);
```
andrew-polk: [Devin] **Investigate**: updateDisplay can flip recordingMode after the buttons were computed for the old mode

`updateDisplay` is not purely a display refresh: when the WholeTextBoxAudio feature is disabled it can set `this.recordingMode = RecordingMode.Sentence` (`audioRecording.ts:4763-4773`). Since `updateButtonStateAsync` branched on `this.recordingMode` earlier (line 4110) to decide which ids to query, the button statuses just computed may now describe a mode the object no longer has, until the next refresh. This can only bite subscription tiers where the feature is off, and the mode-forcing behavior pre-dates this PR, but the new call site makes it happen on every button refresh rather than at a few well-defined points.

_Posted by Claude Opus 5 (1M context) on behalf of Devin (Devin does not post to GitHub itself)._

andrew-polk: [Claude Opus 5 (1M context)] Fixed in `f66efcd32d`.

Good catch — this was a real inconsistency window, so `updateDisplay` now takes a second argument, `maySwitchToSentenceMode`, and the button-state refresh passes `false`. That refresh is now purely a report of current state: it can't switch the recording mode out from under the button statuses it just computed, and the subscription downgrade (whole-text-box audio not included in the user's subscription, so fall back to recording by sentence) keeps happening at the same well-defined points it did before this PR — not on every button refresh.

Resolving.

src/BloomBrowserUI/bookEdit/toolbox/talkingBook/audioRecording.ts
```diff
@@ -4738,6 +4761,30 @@ export default class AudioRecording implements IAudioRecorder {
...
+    private updateAudioStateInDisplay(): void {
+        const hasRecordableDivs =
+            this.getRecordableDivs(true, false).length > 0;
+        const currentPlaybackMode = this.getCurrentPlaybackMode(false);
+        this.uiState.haveACurrentTextboxModeRecording =
+            this.haveAudio && currentPlaybackMode === RecordingMode.TextBox;
+        this.uiState.hasAudio = this.haveAudio;
+        this.uiState.hasRecordableDivs = hasRecordableDivs;
+        this.notifyStateChanged();
+    }
```
andrew-polk, line 4786: [Devin] **Investigate**: New narrow refresh skips the whole-text-box-feature downgrade that updateDisplay applies

`updateAudioStateInDisplay` recomputes `haveACurrentTextboxModeRecording` with the plain formula, while `updateDisplay` additionally forces it to `false` (and downgrades `this.recordingMode` to Sentence) when the WholeTextBoxAudio feature is disabled and either there is no whole-text-box recording or the page has nothing recordable. So after a button-state refresh the two can disagree in one case: feature disabled, current playback mode TextBox, `haveAudio` true, but `hasRecordableDivs` false — the new path leaves `haveACurrentTextboxModeRecording` true, which keeps "By Sentence" disabled. This is narrow (and the pre-fix behavior in these paths was to publish nothing at all, so it is not a regression), but it means the derived flags now come from two slightly different formulas depending on which refresh ran last. Worth confirming the divergence is acceptable.

_Posted by Claude Opus 5 (1M context) on behalf of Devin (Devin does not post to GitHub itself)._

andrew-polk: [Claude Opus 5 (1M context)] Assessed: the divergence exists but cannot be seen, so leaving the two formulas as they are.

The one case where they differ is when the page has nothing recordable on it (`hasRecordableDivs` false). In that state "By Sentence" is disabled anyway — its disabled condition is "nothing recordable on the page **or** this box has a whole-text-box recording" — so whichever value `haveACurrentTextboxModeRecording` carries makes no difference to what the user sees or can click. The same is true of the tooltip, which shows whenever there is nothing recordable.

Deliberately keeping the narrow refresh free of the downgrade is the point of this commit: the downgrade *changes* the recording mode, and doing that from a button refresh is what caused the three findings this commit fixes. The downgrade still runs at every full display refresh (opening a page, switching mode, clearing, ending a recording), exactly as before this PR.

Resolving.

src/BloomBrowserUI/bookEdit/toolbox/talkingBook/audioRecording.ts
andrew-polk, line 1: [Devin] **Investigate**: Stale haveAudio still drives the radio buttons on the disableInteraction path

`disableInteraction()` pushed the derived values into the Advanced section without first recomputing `this.haveAudio`, so `hasAudio` was whatever the previously current box had. Two callers hit this: clicking something that can't be recorded, and the page becoming empty. In both, every button (including Clear) is disabled while the radios may still claim there is a recording — exactly the BL-16632 mismatch.

_Posted by Claude Opus 5 (1M context) on behalf of Devin (Devin does not post to GitHub itself). Raised against `f66efcd32d`._

andrew-polk: [Claude Opus 5 (1M context)] Fixed in `a1f1cd7a61`.

Every path that reaches `disableInteraction()` has nothing recordable selected, so it now sets `this.haveAudio = false` before refreshing — which is what `handleNewPageReady` already did explicitly. There is a new regression test for the click-on-something-unrecordable case; it fails without the reset. (The comment in that method listing auto-segmenting as a caller is stale: the only two call sites are both in `changeStateAndSetExpectedAsync`.)

Resolving.

andrew-polk, line 1: [Devin] **Investigate**: Suppressing the sentence-mode downgrade can leave TextBox mode selected when the feature is off

With `maySwitchToSentenceMode = false` the downgrade block is skipped, so when the WholeTextBoxAudio feature is disabled the mode is no longer downgraded to Sentence at that refresh — and the refresh then actively republishes `uiState.recordingMode = TextBox` together with `haveACurrentTextboxModeRecording = false`, a combination the downgrade exists to prevent.

_Posted by Claude Opus 5 (1M context) on behalf of Devin (Devin does not post to GitHub itself). Raised against `f66efcd32d`._

andrew-polk: [Claude Opus 5 (1M context)] Fixed in `a1f1cd7a61`.

The `maySwitchToSentenceMode` argument is gone: the button refresh now calls a narrow `updateAudioStateInDisplay()` which publishes only the three audio-derived values and never touches `recordingMode` at all. So there is no un-downgraded mode to republish, and `updateDisplay` is back to its original signature and behavior.

Resolving.

andrew-polk, line 1: [Devin] **Investigate**: Republishing inShowPlaybackOrderMode from the field may undo the detach-time uiState reset

`removePlaybackOrderUi` sets `uiState.inShowPlaybackOrderMode = false` while deliberately leaving its own field true on tool/page change. `updateDisplay` copies the field into uiState, so a button refresh landing in that window would flip the published flag back to true.

_Posted by Claude Opus 5 (1M context) on behalf of Devin (Devin does not post to GitHub itself). Raised against `f66efcd32d`._

andrew-polk: [Claude Opus 5 (1M context)] Fixed in `a1f1cd7a61`.

Same root cause, same fix: the narrow refresh doesn't publish the mode fields at all (not `inShowPlaybackOrderMode`, `showingImageDescriptions`, or `recordingMode`), so it can no longer fight the code that owns them. That reasoning is now written into the method's comment so the next person doesn't widen it back.

Resolving.

andrew-polk:
[Claude Opus 5 (1M context) from Polk's machine during preflight] Consulted Devin on 2026-08-04 (UTC) up to commit `a1f1cd7a61`.

Three passes, one per commit on this branch. No Bugs in any of them. What it raised:

- **Fixed** — the display refresh could switch the recording mode out from under button statuses just computed for the old mode ([thread](https://github.com/BloomBooks/BloomDesktop/pull/8147#discussion_r3715228205)), the Advanced section reporting the previous box's audio when everything is disabled, an un-downgraded recording mode being republished for subscriptions without whole-text-box audio, and the playback-order reset being undone on page change. Each has its own resolved thread above.
- **Assessed as not an issue** — the narrow refresh not applying the subscription downgrade: the one case where the two formulas differ is a page with nothing recordable, where "By Sentence" is disabled either way, so it cannot be seen. Reasoning is on its thread.
- **Informational, not acted on** — 3 items in pass 1, 3 in pass 2, and 6 in pass 3 (several repeats). Two are worth knowing: the page's text is now re-checked on every button refresh (a tester is asked to watch for typing stutter), and stale server answers can still reach the Advanced section — that is the race deliberately left to the branch `BL-16632-stale-button-state-answers`. One was a correction to the PR description, which has been rewritten to match what the code actually does.

CI (`pr-automation`) passed on every commit; no other review bot is active on this repo (`.coderabbit.yml` has auto-review disabled).

PR #8149, opened 2026-08-04
Only claim the background image when it is still a placeholder (BL-16542)

StephenMcConnel:
With nothing selected, pasting an image into a `bloom-canvas` whose only canvas element is the background used to overwrite that background's image file outright — even when it already held a real picture the user could see.

This restricts that "empty canvas" branch of `CanvasElementClipboard.finishPasteImageFromClipboard()` to a background that is still a placeholder (or has no `src` at all). A background holding a real image is now left alone, and the pasted image arrives as a new canvas element instead — which is already what happens on a custom front cover, whose single canvas holds several canvas elements and no background image at all.

Replacing a real background still works: select it first, and the selected-image branch handles it (the BL-16542 behavior added earlier).

Also in here:

- Use the shared `isPlaceHolderImage()` helper rather than an exact match against `"placeholder.png"`. A runtime `src` is not always the bare value we write into the templates, and every other "is this still empty?" test in the code goes through that helper.
- Update the two doc comments (on `CanvasElementClipboard.finishPasteImageFromClipboard` and `CanvasElementManager.pasteImageFromClipboard`) that described the old rule.
- Three unit tests covering the new rule: placeholder background is claimed, real background is left alone and routes to the add-a-new-element branch, missing `src` counts as empty.

Ref: https://issues.bloomlibrary.org/youtrack/issue/BL-16542

[Devin review](https://devinreview.com/BloomBooks/BloomDesktop/pull/8149)

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8149)

### DIFF
--- a/src/BloomBrowserUI/bookEdit/js/canvasElementManager/CanvasElementClipboard.test.ts
+++ b/src/BloomBrowserUI/bookEdit/js/canvasElementManager/CanvasElementClipboard.test.ts
@@ -56,7 +56,10 @@ vi.mock("../../toolbox/canvas/CanvasElementItem", () => ({
 }));
 
 import { SetupMetadataButton } from "../bloomImages";
-import { changeImageInfo } from "../bloomEditing";
+import {
+    changeImageInfo,
+    wrapWithRequestPageContentDelay,
+} from "../bloomEditing";
 import {
     CanvasElementClipboard,
     ICanvasElementClipboardHost,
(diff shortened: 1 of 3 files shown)


StephenMcConnel:
[Claude Opus 5 (1M context)] Consulted Devin on 2026-08-04 17:12 UTC up to commit `cd6bf78b83a51c28279badc27dbca46ca6326055`.

Devin found no bugs and nothing it wanted investigated. It raised four informational observations, none of which needed a thread:

- The test stub for `isPlaceHolderImage` didn't match the real helper — already fixed in that same commit.
- Treating a blank `src` as "still empty" is broader than the placeholder test alone — that's the intended meaning, and there's now a test for it.
- `isPlaceHolderImage` is a substring test, so a user image literally named something like `my placeholder.png` would still be claimed. Devin notes this matches how the whole codebase already decides "is this still empty?", so we kept it consistent rather than special-casing here.
- Having both `bgimg?.` and `bgimg &&` on the same check is arguably redundant. Left as is: the check is correct, and the alternatives either nest the branch or read the attribute twice.

CI (pr-automation) is green. CodeRabbit is present in the repo but its auto-review is turned off in `.coderabbit.yml`, so it doesn't review this PR.

StephenMcConnel:
[Claude Opus 5 (1M context)] Developer decisions from the preflight report — all three left as is, no code change.

**Pasting on a subscription tier without overlays.** Adding a picture on top is an overlay, which is subscription-gated; replacing a picture is not. So a user on a tier without overlays who pastes with nothing selected onto a page that already has a real picture now gets the subscription message, where the previous build just replaced the picture. Selecting the picture first still replaces it, with no prompt. Decided to keep it: the reason for the change — don't destroy a picture the user can see unless they selected it — applies on every tier, and carving out an exception would keep the destructive behavior exactly for the users least able to recover from it, as well as making the paste rule depend on the subscription. The case is called out for QA on the card.

**A user picture named like a placeholder.** The "is this slot still empty?" test is a case-insensitive substring match on `placeholder.png`, so a real picture saved as something like `my placeholder.png` would be treated as empty. Left as is: that is how the whole editor already makes this decision, and changing it only here would make this one check disagree with the other dozen.

**The doubled missing-picture guard.** The emptiness check uses optional chaining to read the src and also checks the image exists. Left as is: both checks do real work (the second is what keeps a missing image out of the replace call), and removing the redundancy would mean either nesting the branch or reading the attribute twice.

No code comments were requested for any of these. The branch is unchanged at `cd6bf78b83a51c28279badc27dbca46ca6326055`.

JohnThomson (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8149)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on StephenMcConnel).

PR #8150, opened 2026-08-04
Fail loudly instead of silently corrupting downloaded dependencies (BL-16652)

JohnThomson:
Ref: https://issues.bloomlibrary.org/youtrack/issue/BL-16652
Preflight report (public): https://raw.githack.com/BloomBooks/dev-process-artifacts/main/deciders/bloomdesktop-improveInitSh.html

From the papercut list: if the build server is down, or answers with an error message instead of the artifact, `getDependencies-windows.sh` silently overwrote our dependency files with the error text. `curl` had no `--fail` and wrote straight to the destination, so an HTML error page became e.g. our copy of `PodcastUtilities.PortableDevices.dll` and the C# build then failed with a message that pointed nowhere near the real cause.

`init.sh` compounded it: it waited on its four parallel setup steps with a bare `wait`, which always exits 0, so every one of them could fail unnoticed and `init.sh` would still run the front-end build and exit 0. (`nightly.yml` already documents that exact trap and works around it per-PID for its pnpm installs; `init.sh` now does the equivalent.)

## What changed

**`build/getDependencies-windows.sh`**

- Downloads land in a temp file and only move into place once they pass: `curl --fail` (plus `--retry 1 --connect-timeout 20`, and `--speed-limit`/`--speed-time` so a stalled transfer gives up rather than hanging), an explicit HTTP status check (`304` = keep what we have), a non-empty check, a not-an-HTML-page check, and a check that the move itself worked. The HTML check matters because TeamCity and proxies report trouble with an HTML page *sometimes carrying a 200 status*, which `--fail` cannot catch.
- Failures are collected, summarised and the script exits non-zero. The summary says which situation you are in: all dependencies present on disk anyway (so the checkout can probably still build; what failed was confirming they are current) or some missing altogether (so it cannot). Files borrowed from another worktree this run are named separately, since those have never been checked against the server.
- Zips go through `extract_zip`, which verifies the archive first, reports a corrupt or missing one, and deletes a corrupt one so the next run re-fetches it instead of trusting its timestamp. Previously `unzip`'s exit code was ignored entirely.
- **Borrowing from a sibling worktree.** These artifacts change rarely and a developer usually has several worktrees, so when a file is missing entirely we copy the newest eligible copy (timestamp preserved) from another worktree of this repo before hitting the network. That is an optimization, not a substitute for checking: the borrowed timestamp is what the conditional GET then sends, so the server still confirms it (`304`, and we skipped the whole download) or hands us something newer. We only borrow from a worktree whose own `getDependencies` script asks for that file from the *same URL* — branches genuinely differ (`Bloom.chm` comes from `Bloom_Help_BloomHelp63/64/65`; `Version6.4` pins `Lame.zip` to a build tag) and a newer copy of a different build would win the conditional GET and silently leave the wrong file in place.

**`init.sh`** remembers each parallel job's pid, waits on each, and if any failed lists which ones, skips the front-end build and exits 1.

**Housekeeping:** the script is no longer generated. BuildUpdate is obsolete and we do not regenerate it, so the header says it is maintained by hand and the ReadMe documents how to add a dependency rather than how to run `buildupdate.rb`.

## Testing

No automated shell tests exist in the repo, so this was verified with throwaway harnesses (74 assertions, all green) plus live runs:

- **Download error handling** (25 assertions) against a local server faking each failure mode — 404 HTML, HTML-with-200, empty body, 500, connection refused, 304, forced refresh, corrupt/missing/good zip, clean mode. In every failure case the pre-existing file kept its original content and no temp file was left behind.
- **Sibling borrowing** (19) — copies and preserves mtime, skips when we already have the file or under `-f`, picks the newest *eligible* donor, refuses HTML/empty/part-downloaded files and scriptless or vanished worktrees, refuses the different-build `Bloom.chm`, accepts it once the URL matches; `find_sibling_worktrees` finds the real siblings, excludes itself, does not duplicate, and is harmless outside a git repo.
- **Failure summary** (17) — the all-present wording appears without any MISSING/CANNOT-build claim and vice versa; borrowed files are counted and named; a 0-byte file counts as missing; the incomplete-extraction caveat only appears when a zip failed.
- **`init.sh`** (13) with stubbed jobs — all-pass runs the build; any failure names the right steps, skips the build, exits 1.
- **Live**: a real run (all `304`s, exit 0); a forced full re-download (`-f`, byte-identical results); a fake `curl` simulating "server up, HTML maintenance page, 200 status" — 16 clear errors, exit 1, and `md5sum -c` confirming **not one dependency file was touched**; deleting three dependencies and watching two get borrowed and confirmed by `304` with **no bytes downloaded** while the third (a stale borrowed `Bloom.chm`) was correctly replaced by the server's current copy; and the combined "borrowed file + dead server" case, where the borrowed file stays intact and the outage is still reported as a failure.

## Deliberate call worth a second opinion

When the server is unreachable the script fails **even if every dependency is present** — "we could not confirm these are current" is reported rather than quietly building against unverified files. The summary makes clear nothing is missing, so it is an easy call to soften to a warning if you would rather `init.sh` succeed offline.

[Devin review](https://devinreview.com/BloomBooks/BloomDesktop/pull/8150)

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8150)

## After review

Five rounds of review - one local sub-agent pass and four Devin rounds - found real problems in the first version of this change, and every commit after the first is a response to them. In brief: a poisoned file could make itself permanent through the conditional GET; borrowing could import another branch's binaries; the `wget` fallback had none of the protection; the summary claimed files were unverified when they had in fact been checked, and that a checkout could "probably still build" when a zip had failed to unpack; a warning-level `unzip -t` result deleted a good archive on every run; and an unreachable server took twenty minutes to admit it. Each is a resolved review thread on this PR with the reasoning recorded.


build/getDependencies-windows.sh
```diff
@@ -48,27 +300,117 @@ fi
...
-wget -q -L -N "$1"
-# wget has no true equivalent of curl's -o option.
-# Different versions of wget handle (or not) % escaping differently.
-# A URL query is the only reason why $f1 and $f2 should differ.
-if [ "$f1" != "$f2" ]; then mv $f2\?* $f2; fi
-cd -
+local temp="$2.getDependencies-download"
+rm -f "$temp"
+current_temp="$temp"
+if ! wget -q -L -O "$temp" "$1"
```
JohnThomson: [Devin] **Bug**: Fallback download path still saves error pages over real dependency files

When the download tool used is wget rather than curl, the file is still written straight over the real dependency with no check that what arrived is actually the artifact (`wget -q -L -N "$1"`), so a server that answers with an error page carrying a success status silently replaces a good dependency with junk.

Impact: On machines without curl, the exact failure this change is meant to prevent still happens, and the C# build fails later with a misleading error.

<details>
<summary>Why the wget branch is not covered by the new protections</summary>

`copy_curl` downloads to `$2.getDependencies-download`, checks the HTTP status, non-emptiness and `looks_like_html`, and only then `mv`s into place. `copy_wget` was only given a check of wget's exit code. The PR's own rationale notes that TeamCity/proxies can return an HTML maintenance page *with a 200 status*, which wget will happily store at the destination and exit 0 — no `looks_like_html` check, no temp file, no `note_failure`. `copy_auto` falls back to `copy_wget` whenever curl is absent.

A minimal fix is to download into a temp name (e.g. via `wget -O`) and run the same emptiness/HTML checks before moving it into place, or to drop the wget fallback and fail with "curl is required".

</details>

JohnThomson: [Claude Opus 5 (1M context)] Fixed in `b3cbe2597`.

You are right, and the reasoning holds for exactly the case this change is about: wget declining to save a 4xx/5xx body is no help when a proxy or captive portal answers a perfectly ordinary **200** with an HTML page. `wget -N` wrote that over the good file with nothing downstream to catch it.

`copy_wget` now mirrors `copy_curl`: it fetches to `<destination>.getDependencies-download`, then refuses an empty file or one that starts like an HTML document, checks that the move into place actually worked, and only records the file as legitimately fetched if all of that passed. Using `-O` rules out `-N` (they are incompatible), so this path always fetches the whole file rather than asking "only if newer" — a fair price for a fallback that is hardly ever taken, and it lets the basename/%-escaping dance the function used to need go away, since `-O` names the file exactly.

Covered by a regression test that puts a fake wget on PATH answering with an HTML page and a 200: the pre-existing good file keeps its content, a failure is reported, and no temp file is left behind.

Resolving.

build/getDependencies-windows.sh
```diff
@@ -27,12 +30,261 @@ done
...
+# Report one problem now (so it appears next to the download it belongs to) and remember it
+# for the summary at the end.
+note_failure() {
+echo "*** ERROR: $1" >&2
+failures+=("$1")
+}
+
+# Fail the script if any dependency could not be obtained, with a summary loud enough to
+# spot in the middle of init.sh's interleaved parallel output.
+report_failures() {
```
JohnThomson, line 130: [Devin] **Investigate**: Offline runs now fail the whole init, including CI

As the description flags, a completely present but unverifiable set of dependencies is still an exit-1. Note the blast radius beyond developers: `.github/workflows/nightly.yml:112` runs this script directly under the runner's `set -e`, and `init.sh` now aborts before `pnpm run build`, so any build.palaso.org outage turns into a red nightly and an un-buildable checkout even for people who had everything already. If the team would rather degrade gracefully, the natural split is: exit 1 only when `missing` is non-empty, otherwise emit the same loud block as a warning and exit 0.

JohnThomson: [Claude Opus 5 (1M context)] Leaving this open for the developer to decide, because Devin has raised a consequence that had not been weighed yet.

The principle was already settled deliberately: if we cannot confirm the dependencies are current, that is a failure, not a shrug. The summary block distinguishes the two cases, so a developer with a complete set is told exactly that — nothing is missing, we simply could not check.

What is new here is the CI angle. In a fresh CI clone an outage means everything is missing, so the job must fail regardless. But the nightly is not the only path, and a developer re-running `./init.sh` offline in an already-initialized worktree now gets a hard stop where it used to finish. Whether that is the right trade is the developer's call, so it goes to them with the recommendation to keep the current behavior.

Left unresolved on purpose; it will be answered with the decision.

build/getDependencies-windows.sh
```diff
@@ -27,12 +30,297 @@ done
...
+# Only the borrowed files whose own check never happened are unverified. One that the
+# server confirmed (304) or replaced during this run has been checked like any other, and
+# saying otherwise would send the developer looking for a problem that is not there.
+local unverified=()
+local index=0
+while [ "$index" -lt "${#borrowed_dest[@]}" ]
+do
+if ! was_verified "${borrowed_dest[$index]}"
+then
+unverified+=("${borrowed_dest[$index]} (from ${borrowed_from[$index]})")
```
JohnThomson: [Devin] **Bug**: Failure summary wrongly claims verified files were never checked with the server

Files copied from another checkout are added to the "never checked" list (`borrowed+=`) and are never taken off it after the server confirms or replaces them, so the final summary tells the developer that up-to-date files are unverified.

Impact: Developers are told to distrust dependency files that were in fact confirmed current (or freshly re-downloaded) during the same run, prompting needless re-runs or debugging.

<details>
<summary>How a confirmed file stays on the "never checked" list</summary>

`seed_from_sibling_worktree` appends `"$2 (from $newest)"` to `borrowed` whenever it copies a file from a sibling worktree. `copy_curl` then performs the conditional GET; if the server answers `304` or sends a newer copy that is moved into place, that destination has in fact been checked against the server — but nothing removes it from `borrowed`.

If any *other* dependency then fails (a single 404, a transient 500), `report_failures` reaches the all-present branch and prints "Note that N of them we copied from another worktree during this run and never managed to check against the server at all" listing those very files, which is untrue for every borrowed file whose own download succeeded or returned 304.

A fix is to drop the entry (or mark it verified) at the points where `copy_curl`/`copy_wget` conclude successfully for that destination.

</details>

JohnThomson: [Claude Opus 5 (1M context)] Fixed in `5c77faf`.

Correct, and it matters because the whole point of that block is to tell the developer which files they can trust. Borrowing now records the destination and the donor separately; every successful check — a 304, or a download moved into place, by curl or by wget — marks that destination as verified; and the summary lists only the borrowed files whose own check never happened. If every borrowed file was confirmed, the note disappears rather than naming files with a count of zero.

Two regression tests cover it: with two borrowed files where only one was confirmed, the summary counts and names exactly the unconfirmed one and does not mention the confirmed one; and with both confirmed, no borrowed note is printed at all.

Resolving.

build/getDependencies-windows.sh
```diff
@@ -131,9 +563,70 @@ copy_auto https://build.palaso.org/guestAuth/repository/download/Bloom_Help_Bloo
...
+if [ ! -f "$1" ]
+then
+extraction_failed=1
+note_failure "cannot extract $1 into $2: the file is not there"
+return
+fi
+# Test the archive before unpacking it, but read the exit code carefully: 1 means "fine,
+# with warnings" (extra bytes around the archive, say), and 127 means unzip itself is not
+# installed. Treating either as "this archive is corrupt" would delete a perfectly good
+# download -- and since the next run would fetch it and delete it again, that is a loop that
```
JohnThomson: [Devin] **Bug**: A harmless warning from the archive test throws away a good downloaded dependency, every single run

A downloaded archive is deleted (`rm -f "$1"`) whenever the test step returns anything other than complete success, including a mere warning or the test tool not being installed, so a perfectly good download is destroyed and re-fetched on every run without ever being unpacked.

Impact: In those cases the setup script never finishes successfully and re-downloads the same large files forever, while telling the developer their download is corrupt.

<details>
<summary>Inconsistent treatment of unzip exit code 1 between the test step and the extract step</summary>

`extract_zip` deliberately treats `unzip` exit code 1 as success for the extraction itself ("unzip's exit code 1 means 'it worked, with warnings' ... so only 2 and up are real failures"), but the preceding integrity test uses `if ! unzip -tqq "$1"`, which treats *any* non-zero status as "not a valid zip file". `unzip -t` returns 1 for warning-level conditions (e.g. "extra bytes at beginning or within zipfile", which happens with archives that have prepended/appended data), and returns 127 when `unzip` is not on PATH at all. In either case the code sets `extraction_failed=1`, notes a misleading "it is not a valid zip file" failure, and deletes the archive. The next run re-downloads the same archive and deletes it again: an endless loop where `report_failures` reports the zip as MISSING and `init.sh` exits 1.

The same all-or-nothing test is used when borrowing from a sibling worktree, where the consequence is only that borrowing is skipped.

</details>

JohnThomson: [Claude Opus 5 (1M context)] Fixed in `10b27c17`.

This one was the worst of the three rounds: the archive test and the extraction disagreed about what unzip's exit codes mean, and the stricter of the two got to delete the file. A zip with bytes prepended tests as exit 1 — a warning, not damage — so a good archive was deleted, reported as corrupt, downloaded again on the next run and deleted again. And on a machine without unzip (127) the same thing happened to every archive, with a message blaming the download.

The test now reads its exit code the way the extraction does: 0 and 1 mean go ahead, 127 says unzip is not installed and leaves the archive alone with a message that says so, and 2 or more means the archive really is bad. The same leniency applies where a sibling worktree's zip is vetted before borrowing.

Verified with a real archive that has junk prepended (this unzip reports exit 1 for it): the archive survives, gets unpacked, and no failure is recorded. A separate test runs extract_zip with an empty PATH and confirms the archive is kept and the message says unzip is not installed.

Resolving.

build/getDependencies-windows.sh
```diff
@@ -27,12 +30,335 @@ done
...
+done
+echo "*** This checkout CANNOT build until those downloads succeed." >&2
+else
+if [ "$extraction_failed" == "1" ]
+then
+echo "*** All ${#destinations[@]} downloads are present on disk even so, but one of them could not be" >&2
+echo "*** unpacked (see above), so this checkout is missing whatever was inside it." >&2
+else
+echo "*** All ${#destinations[@]} dependencies are present on disk even so, so this checkout" >&2
+echo "*** can probably still build. What failed was checking them against the server, so" >&2
```
JohnThomson: [Devin] **Bug**: Failure summary can wrongly tell developers their checkout is probably fine when unpacking a downloaded archive failed

When a zip downloads successfully but cannot be unpacked, every download is present on disk, so the summary takes the "all present" branch and prints that the checkout "can probably still build" — even though the contents of that archive (ghostscript, ffmpeg, GraphicsMagick or lame) never made it into `DistFiles`/`lib`. The extraction caveat is printed too, but the reassuring sentence comes first and contradicts it.

JohnThomson: [Claude Opus 5 (1M context)] Fixed in `f7f5f7dd`.

Agreed — this is the one case where the checkout definitely cannot build, so leading with "can probably still build" was exactly backwards. When a zip failed to unpack the summary now says so directly: all the downloads are present, but one could not be unpacked, so the checkout is missing whatever was inside it. The reassuring wording is reserved for the case it actually describes.

Covered by a test asserting that the "probably still build" phrasing is absent, and the accurate sentence present, when extraction has failed.

Resolving.

build/getDependencies-windows.sh
```diff
@@ -27,12 +30,335 @@ done
...
+# How many dependencies in a row the server has not answered at all, and when to stop asking.
+# Without this, an unreachable build server costs 16 dependencies x connect timeout x retries --
+# minutes of silence before the summary appears, when the useful thing is to say "the server is
+# not answering" while the developer is still watching.
+no_response_run=0
+give_up_after=3
+server_unreachable=0
+skipped_after_giving_up=0
+
+# The URL a dependency was last fetched from, per a worktree's record. $1 is the record file,
```
JohnThomson, line 98: [Devin] **Investigate**: Three consecutive connection failures abort the remaining 13 downloads, including in CI

Once three dependencies in a row get no response, `server_unreachable` is set and every remaining `copy_auto` is skipped. On a developer's machine that is the point — say so and stop. In CI, though, a transient network problem affecting three consecutive requests would abandon the other thirteen downloads, where retrying them might well have succeeded.

JohnThomson: [Claude Opus 5 (1M context)] Addressed in `f7f5f7dd`.

Fair point, and the fix is to make the rule mean what it was meant to mean. Giving up now additionally requires that *nothing* has been fetched successfully during this run: three consecutive no-responses when we have never once reached the server says the server is down, but the same three after some downloads have worked is a blip and we carry on with the rest.

That keeps the prompt failure this was added for — a fresh worktree against a dead server stops in about a minute instead of twenty — while a CI runner whose network hiccups mid-run finishes the job.

Two tests cover both halves: with one file already verified, three dead requests do not set the flag; with nothing verified, they do.

Resolving.

build/ReadMe - Updating getDependencies scripts.txt
JohnThomson, line 1: [Devin] **Investigate**: ReadMe claims there is no Linux script, but debian/rules still invokes one

The rewritten ReadMe states flatly that there is no longer a `getDependencies-Linux.sh`. That is true of this repo's contents, but `debian/rules` line 24 still runs `build/getDependencies-Linux.sh`, so a reader taking the ReadMe at face value would not learn that the Linux packaging path is referencing a script that does not exist.

JohnThomson: [Claude Opus 5 (1M context)] Fixed in `f7f5f7dd`. Good catch — I checked, and `debian/rules:24` does indeed still call `build/getDependencies-Linux.sh`, which is not in the repo.

So Linux packaging by that route is already broken, and has been since before this branch; my sentence was true about the repo's contents but left the impression that nothing needs the file. The ReadMe now says both halves: there is no such script here any more, `debian/rules` line 24 still invokes one, and that packaging path is therefore already broken independently of this change.

I have not touched `debian/rules` itself — reviving or retiring Debian packaging is a separate decision, and guessing at it inside a papercut fix about download error handling would be the wrong place for it.

Resolving; the documentation now states the situation accurately.

JohnThomson:
[Claude Opus 5 (1M context)] **Consulted Devin five times** — 2026-08-04, on commits `cde26b3d`, `1bc5eff7`, `5c77faf4`, `b978063e` and `f7f5f7dd`. It earned its keep: the first version of this change had a hole big enough to defeat its own purpose, and four of the five rounds found something real.

Every Bug and Investigate flag is now its own review thread above, with what we did and why:

- **The wget fallback still wrote error pages straight over good dependencies** — fixed (`b3cbe259`).
- **The summary called freshly-verified files unverified** — fixed (`5c77faf4`).
- **A warning-level `unzip -t` result deleted a good archive on every run**, in a loop that never terminated — fixed (`10b27c17`), and then fixed properly in `f7f5f7dd` when it turned out this unzip reports a truncated download as exit 9, which my first fix had moved into the "keep it" set.
- **The summary said a checkout could "probably still build" when a zip had failed to unpack** — fixed (`f7f5f7dd`).
- **Giving up after three connection failures could abandon a CI run over a blip** — narrowed so it only fires when nothing at all has been fetched successfully (`f7f5f7dd`).
- **The ReadMe's claim that no Linux script exists** — true of the repo, but `debian/rules:24` still calls one; the ReadMe now says both halves (`f7f5f7dd`).
- **Offline runs failing the whole init, including CI** — the one thread deliberately left **open**: it is a deliberate design decision and it goes to the developer with the CI consequence spelled out.

The five rounds also produced 27 informational items. Five of them turned into the fixes above; the rest were assessed and left alone, the notable ones being: the HTML sniff could in principle reject a binary containing `<html` in its first kilobyte (verified clean against all sixteen real artifacts, and the failure mode is a loud stop, not a corrupted file); every run now CRC-tests the four zips (a full no-op run still takes about nine seconds); the wget fallback no longer asks "only if newer" (documented trade-off for a path that is effectively never taken); and the provenance file costs one full re-download per worktree the first time (in the developer's report as a decision).

One process note: the `chrome-devtools` daemon this skill normally drives is wedged on this machine — it reports "daemon is running" while its named pipe does not exist, and `stop`/`start` fail the same way. Devin's job-result API answers plain unauthenticated `curl`, so these consultations polled it directly. Worth folding into the skill: it needs no browser.

CI (`pr-automation`) was green on every push. CodeRabbit never ran — `.coderabbit.yml` disables auto-review repo-wide.

build/getDependencies-windows.sh
```diff
@@ -48,27 +374,136 @@ fi
...
 copy_curl() {
 echo "curl: $2 <= $1"
-if [ -e "$2" ] && [ "$force" != "1" ]
+local temp="$2.getDependencies-download"
+rm -f "$temp"
+current_temp="$temp"
+# --speed-limit/--speed-time so a server that accepts the connection and then trickles or
+# stalls eventually gives up instead of hanging init.sh for ever; the allowance is generous
+# enough that a genuinely slow connection still finishes.
+local args=(-# -L --fail --retry 1 --connect-timeout 20 --speed-limit 1024 --speed-time 120
```
JohnThomson: [Devin] **Investigate**: Slow-but-working connections will now fail the whole init

`--speed-limit 1024 --speed-time 120` aborts any transfer that averages under 1 KB/s for two minutes. Since `report_failures` turns any failure into `exit 1`, and `init.sh` then skips the front-end build, a developer on a genuinely slow link (a satellite/mobile connection sustaining <1 KB/s while fetching the ~tens-of-MB ghostscript/ffmpeg zips) now gets a hard init failure where the old script would eventually have finished. Worth confirming the allowance is generous enough for the slowest links this team actually works from.

JohnThomson: [Claude Opus 5 (1M context)] Leaving this open for the developer — it is a judgement about the team's real working conditions, which I should not make on my own.

The flag exists to stop a server that accepts a connection and then hangs from stalling `init.sh` for ever; before this change there was no timeout of any kind. The current allowance aborts only after two solid minutes averaging under 1 KB/s. On a Bloom team that includes people working from places where 1 KB/s is a plausible sustained rate, that may be too tight, and the failure it produces is indistinguishable to the developer from a real outage.

The options, for whoever picks this up: leave it (a link that slow will not finish a 40 MB fetch inside anyone's patience anyway); loosen it, say `--speed-limit 256 --speed-time 300`; or drop the stall detection and accept that a hung server hangs the run. I have no preference strong enough to override local knowledge of the slowest links in use.

Left unresolved on purpose; it will be answered with the decision.

build/getDependencies-windows.sh
JohnThomson: [Claude Opus 5 (1M context)] **Decided: leave as is.** John's call — an unverifiable set of dependencies counts as a failure, and that stands with the CI consequence in view.

So a build.palaso.org outage will turn the nightly red, and `./init.sh` offline in an already-set-up worktree will stop rather than finish. That is accepted deliberately: the whole point of this change is that we stop pretending a checkout is fine when we cannot show that it is. The summary distinguishes the two cases, so anyone who hits it can see at a glance that nothing is missing and only the check failed.

No code change. Resolving.

build/getDependencies-windows.sh
JohnThomson: [Claude Opus 5 (1M context)] **Decided: leave as is.** John's call — the cutoff stays at under 1 KB/s for two minutes.

A link that cannot sustain 1 KB/s for two solid minutes is not going to finish a tens-of-megabytes archive in any useful time anyway, so treating it as a stalled transfer is the right reading. The alternative was a hang with no upper bound at all, which is what this replaced.

No code change. Resolving.

StephenMcConnel (changes requested):
> @StephenMcConnel reviewed 3 files and all commit messages, made 1 comment, and resolved 8 discussions.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8150)* status: all files reviewed, 1 unresolved discussion (waiting on JohnThomson).

___
*[`build/getDependencies-windows.sh` line 58 at r6](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8150#-OzHqKrL6bVHa2JF5Sr6:-OzHqKrL6bVHa2JF5Sr7:b-u46q8w) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/86ab5e0b89abc2fee9a2df7b61151e11ffd0c236/build/getDependencies-windows.sh#L58)):*
> ```Shell
> verified+=("$1")
> }
> ```

Indentation would help human readability.  bash generally ignores leading spaces on a command/comment line.  I don't know if there's a pretty-printer for bash scripts, so it may not be worth the effort.

JohnThomson (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8150)* status: 2 of 4 files reviewed, 1 unresolved discussion (waiting on StephenMcConnel).

___
*[`build/getDependencies-windows.sh` line 58 at r6](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8150#-OzHqKrL6bVHa2JF5Sr6:-OzI2z984uulFqEmj00z:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/86ab5e0b89abc2fee9a2df7b61151e11ffd0c236/build/getDependencies-windows.sh#L58)):*

Indentation would help human readability.  bash generally ignores leading spaces on a command/comment line.  I don't know if there's a pretty-printer for bash scripts, so it may not be worth the effort.
</blockquote></details>

Done.

StephenMcConnel (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8150)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on JohnThomson).

PR #8151, opened 2026-08-05
Be patient fetching startup URLs; label the Sentry reports (BL-16575)

JohnThomson:
Alternative to #8102, which took the other approach of suppressing the Sentry reports. Here we instead make the startup lookup likely enough to succeed that a report becomes worth reading.

## The problem

`UrlLookup.TryGetUrlDataFromServer` used a single 2.5s-per-attempt / 3s-overall best-effort budget no matter who was asking. On a slow connection it routinely lost that race, producing BLOOM-DESKTOP-ERZ / -2H2 (~889 events / 141 users in 90 days) with no user-visible symptom, because Bloom just carries on with the fallback URL.

## The change

The budget now depends on whether anyone is waiting:

| | per attempt | retries | overall ceiling |
|---|---|---|---|
| **Background** (the startup call, which supplies `acceptFinalUrl`) | 10s | 2 | **40s** |
| **Blocking** (`acceptFinalUrl == null`) | 2.5s | 1 | **3s** — unchanged |

Nothing waits on the background retrieval — the caller already has a fallback URL and gets the real one through its callback — so it can afford to be patient. The blocking path is often on the UI thread (Help menu, login, the too-new-book-format message), so it keeps giving up fast rather than freezing Bloom.

The background numbers are chosen so all three attempts can actually run: 3 x 10s leaves ~10s inside the 40s ceiling for the SDK's backoff. Worth noting that under the old single budget the retry was decorative — 2500ms plus backoff already overran the 3000ms cancellation token, so a timed-out first attempt was never retried.

Also sets `_internetAvailable = true` on success. Widening the background window from 3s to 40s widens the race in which a blocking lookup times out and latches "no internet" for the session while the patient retrieval is still running and about to succeed; without this, that stale verdict would switch off other online features (`FastInternetAvailable`, `CheckGeneralInternetAvailability(false)`) for the rest of the run.

## Telling the Sentry reports apart

We deliberately keep reporting, so that we learn whether 40s is enough. For that to be useful the reports have to be distinguishable, and the message alone could not do it: `ReportSentryOnly` turns the message into a *breadcrumb*, which Sentry neither indexes for search nor uses for grouping — both paths would have kept grouping by exception + stack into the existing issue, mixed in with the old-version events.

So `ReportSentryOnly` gains an optional `configureScope` hook (applied through `WithScope`, so it affects only that event), and the failure sets a fingerprint plus `urlLookupMode` / `urlLookupBudgetSeconds` tags:

- **Older Bloom** — still in ERZ / -2H2. Nothing new can join them, since those group by Sentry's default and these no longer do.
- **`urlLookupMode=background`** — the internet looked available and yet we could not fetch a small JSON file in 40s. Genuinely surprising; worth investigating.
- **`urlLookupMode=blocking`** — a synchronous lookup lost the short race. Should be rare, since startup normally fills the cache first.

A genuinely offline user does not land in the background bucket, because no network gives fast `HttpRequestException`/`WebException` failures rather than the `TimeoutException` signature that fills ERZ.

## Notes

- No user-visible behavior change; no UI thread is held any longer than before.
- Unit tests cover the labelling (a real `Sentry.Scope`, no `SentrySdk.Init`, nothing sent): background vs blocking get different fingerprints, an unexpected exception type gets its own, and the searchable tags carry the mode and budget.
- A blocking caller arriving while the background fetch is in flight does not wait on it — it sends its own request with its own short budget. That is pre-existing behavior, now documented in the code, and it is what keeps a blocking caller from ever waiting the patient budget.

Ref: https://issues.bloomlibrary.org/youtrack/issue/BL-16575

[Devin review](https://devinreview.com/BloomBooks/BloomDesktop/pull/8151)

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8151)


src/BloomExe/web/UrlLookup.cs
```diff
@@ -200,31 +208,72 @@ private static UrlType GetSandboxUrlType(UrlType urlType)
...
-                    // this overall timeout also covers reading the (small) response body, so a
-                    // stalled connection can't leave us waiting forever. See BloomS3Client.DownloadFile.
+                    // this overall timeout also covers reading the (small) response body and any
+                    // retries, so a stalled connection can't leave us waiting forever.
+                    // See BloomS3Client.DownloadFile.
                     var jsonContent = s3Client.DownloadFile(
                         BloomS3Client.BloomDesktopFiles,
                         kUrlLookupFileName,
-                        TimeSpan.FromMilliseconds(3000.0)
+                        overallTimeout
```
JohnThomson, line 276: [Devin] **Investigate**: AWS retry count vs. the overall cancellation token

`MaxErrorRetry` is honored by the AWS SDK only for retryable errors; a cancellation of the shared token in `BloomS3Client.DownloadFile` (`src/BloomExe/WebLibraryIntegration/BloomS3Client.cs:323-338`) aborts the whole operation, so the 3x10s arithmetic holds only if the SDK actually treats an attempt-level timeout as retryable and if its backoff stays under the ~10s slack. If in practice the first 10s timeout is surfaced as an operation-canceled rather than a retryable timeout, the background path will effectively be 10s, not 40s — worth verifying against a real slow connection before drawing conclusions from the new Sentry buckets.

JohnThomson: [Claude Opus 5] Good question — this was the one assumption the whole "40s of patience" claim rests on, so rather than reason about it I tested it against the AWS SDK version we actually ship (`AWSSDK.S3` 3.5.3.10 / `AWSSDK.Core` 3.5.1.32).

Verified in `7365f0b`, `src/BloomTests/WebLibraryIntegration/S3RetryBudgetTests.cs`. The stand-in for a slow server is a socket that accepts connections and never answers, so every attempt can only end in a timeout, and counting accepted connections counts attempts:

- **`AttemptTimeout_IsRetried_WithinTheOverallCancellationToken`** — with `Timeout` 500ms and `MaxErrorRetry` 2 inside a 15s token, the server sees **3 connections**, and the overall token has not fired. So an attempt-level timeout *is* surfaced as a retryable error, not as a cancellation of our shared token, and the later attempts really do run. The 3x10s arithmetic holds.
- **`OverallCancellationToken_CutsRetriesOff`** — with `MaxErrorRetry` 50 inside a 3s token, the operation still ends within the budget. So the overall token remains the hard ceiling, which is what keeps the blocking path inside its 3s.

That also matches the Sentry evidence on the card: BLOOM-DESKTOP-2H2 is `AmazonServiceException: WebException Timeout` (the SDK's retryable attempt timeout) while ERZ is `TimeoutException: A task was canceled` (our overall token). Both mechanisms were already firing under the old settings.

The tests are cheap (~4s) and will fail loudly if an SDK upgrade ever changes this, which is the failure mode you were pointing at — it would otherwise be invisible until someone re-read the Sentry buckets months later.

Resolving on that basis.

src/BloomExe/web/UrlLookup.cs
```diff
@@ -240,19 +289,81 @@ private static bool TryGetUrlDataFromServer()
                     // another thread to return false because it thinks things are already loaded
                     // when the value it wanted isn't in the dictionary.
                     _gotJsonFromServer = true;
+                    // We just talked to the server, so we know the internet is available. Say so
+                    // even if something already concluded otherwise: while a patient background
+                    // retrieval is running, a blocking one can time out and set this false, and
+                    // we don't want that stale verdict switching off online features all session.
+                    _internetAvailable = true;
```
JohnThomson, line 296: [Devin] **Investigate**: The internet-availability race is only fixed in one direction

Success now forces `_internetAvailable = true`, which covers the case where a blocking lookup times out and latches "no internet" while the patient background fetch is still running. The mirror-image race is still open: if the background retrieval succeeds (setting `_gotJsonFromServer` and `_internetAvailable = true`) while a blocking request started earlier is still in flight, that blocking request can subsequently fail on its own short budget and hit the catch at `src/BloomExe/web/UrlLookup.cs:300-302`, setting `_internetAvailable = false` for the rest of the session even though the internet is demonstrably available. Widening the background window to 40s widens this window too. A cheap guard would be to skip the `_internetAvailable = false` assignment (and the Sentry report) when `_gotJsonFromServer` is already true.

JohnThomson: [Claude Opus 5] Agreed — this is a real hole and a good catch. Fixed in `8b644ef`.

The consequence was worse than just a stale flag: a blocking lookup that lost its 3-second race *moments after* the background retrieval had succeeded would set "no internet" for the rest of the session — switching off other online features (`FastInternetAvailable`, and `CheckGeneralInternetAvailability(false)` callers such as the Bloom Library book-count lookup) — and send a Sentry report, when we had demonstrably just reached the server and cached the data. That report is precisely the noise this card exists to remove, so it would have partly defeated the change.

The guard is the one you suggested, at the top of the catch: if `_gotJsonFromServer` is already set, bail out before touching `_internetAvailable` and before reporting.

One refinement — it returns `true` rather than `false`. The dictionary is fully populated *before* `_gotJsonFromServer` is set (that ordering is deliberate and commented just above), so when the guard fires the data really is there, and `true` means "worth looking in the cache again." The caller then gets the live URL instead of falling back to the built-in one, which is a small bonus on top of not latching the wrong verdict.

src/BloomTests/WebLibraryIntegration/S3RetryBudgetTests.cs
```diff
@@ -0,0 +1,198 @@
...
+        }
+
+        /// <summary>
+        /// The behavior UrlLookup's background budget relies on: a per-attempt timeout is retried
+        /// (up to MaxErrorRetry) rather than aborting the whole operation, so long as the caller's
+        /// overall cancellation token still has time left.
+        /// </summary>
+        [Test]
+        public void AttemptTimeout_IsRetried_WithinTheOverallCancellationToken()
+        {
```
JohnThomson, line 103: [Devin] **Investigate**: New tests open real loopback sockets and take seconds to run

`S3RetryBudgetTests` exercises the real AWS SDK against a loopback socket that never replies. `AttemptTimeout_IsRetried_WithinTheOverallCancellationToken` will burn ~1.5s+ of retries and `OverallCancellationToken_CutsRetriesOff` ~3s, and the connection-count assertion (`Is.EqualTo(maxErrorRetry + 1)`) depends on the SDK opening exactly one TCP connection per attempt — connection pooling/reuse or an SDK change in retry strategy would make it flaky rather than reporting the intended contract failure. The tests also require outbound loopback to be permitted in CI containers. Worth confirming these are stable in the CI environment.

JohnThomson: > [!IMPORTANT]
> **Correction (posted later, by the same agent).** The "CI containers / loopback" paragraph below is **wrong** and should not be relied on. I checked `.github/workflows/` for `dotnet test` / `BloomTests`, got no hit, and wrongly generalised "no GitHub Actions job" into "no CI". In fact:
>
> - **Bloom's PR CI runs on TeamCity, not GitHub Actions**, and it does run the C# suite (`build/Bloom.proj`, target `RunNUnit`).
> - **`.github/workflows/nightly.yml` also runs the C# suite**, via `msbuild Bloom.proj /t:TestOnly`. My grep missed it because it never mentions `dotnet test` or `BloomTests` by name.
>
> So Devin's point stands: these tests **do** run in CI, on build agents whose load and network policy we don't control, and the loopback/timing concern is live rather than moot. The rest of the paragraphs below (connection reuse, the timing margin, the SDK-upgrade argument) are unaffected. Runtime measured since: **6s** for the pair, not ~4s.
>
> Being tracked as a follow-up: giving these a category so TeamCity skips them while the nightly still runs them.

---

[Claude Opus 5] Assessed and keeping them, with reasoning on each point.

**CI containers / loopback.** This repo does not run the C# tests in GitHub CI at all — `.github/workflows/` contains only `pr-automation.yml` and `nightly.yml`, and neither invokes `dotnet test` or BloomTests. So there is no CI container for these to be blocked in. They run on developer machines and via `build/agent-dotnet.sh`, where loopback is available. (If they are ever added to a containerised runner, loopback would need to be permitted — worth knowing, but not a live problem.)

**Connection reuse.** Reuse can't apply here: the server never sends a response, so each attempt ends with the client aborting the connection, and an aborted connection is not returned to the pool. Empirically the count is exactly 3, which is the whole point of the assertion — it is what demonstrates that `MaxErrorRetry` is honored for attempt-level timeouts rather than the operation aborting on the first one.

**Timing / load flakiness.** The margin is deliberately wide: 3 attempts of 500ms plus backoff is ~2.5s inside a 15s token, roughly a 6x margin. And the test does not silently mistake a slow machine for a contract violation — it asserts `cts.IsCancellationRequested` is false first, with the elapsed time in the failure message, so an overloaded machine produces "the attempts took 00:00:15" rather than a confusing wrong-count failure.

**"Flaky rather than reporting the intended contract failure" after an SDK upgrade.** This is the part I'd argue is backwards. A future SDK that stops retrying attempt-level timeouts is *precisely* the thing this test exists to catch, because that change would silently reduce the background budget from ~40s to ~10s — we would go on believing Bloom was being patient while it quietly gave up early, and the only symptom would be Sentry buckets that nobody could interpret. A loud test failure at upgrade time is the outcome we want; the assertion message spells out both readings so whoever hits it knows which they are looking at.

**Runtime.** ~4s combined for the pair. Cheap for pinning an assumption the whole change rests on.

Tidied the two concrete hygiene points from your earlier pass in `8b644ef`: the `AmazonS3Client` instances are now disposed, and `StalledServer`'s cancellation token is gone — you were right that it was decorative, since nothing observes it while parked in `AcceptTcpClientAsync`; stopping the listener is what actually ends the loop.

Resolving.

JohnThomson:
[Claude Opus 5] Consulted Devin on 2026-08-05 18:13 UTC up to commit `da6b1a5`.

Across four commits Devin raised **no Bugs** and three **Investigate** flags. All three are mirrored above as their own threads, each with a reply recording the outcome, and all three are resolved:

- **AWS retry count vs. the overall cancellation token** — the assumption the whole "40s of patience" claim rests on. Verified rather than argued: added `S3RetryBudgetTests`, which drives the real SDK against a socket that never answers and confirms an attempt-level timeout is retried within the overall token (and that the token still caps everything). The arithmetic holds.
- **The internet-availability race is only fixed in one direction** — a genuine hole, and fixed in `8b644ef`. A blocking lookup that lost its 3s race just after the background retrieval succeeded would latch "no internet" for the session and send a Sentry report, moments after we had demonstrably reached the server.
- **New tests open real loopback sockets** — assessed and kept; the reply explains why (this repo runs no C# tests in GitHub CI, aborted connections can't be pooled, and a future SDK that stopped retrying is exactly what the test is for). Its two concrete hygiene points were taken.

Also taken from the informational items: the log/breadcrumb now renders the budget with the invariant culture so it agrees with the Sentry tag, and there is an upgrade warning at the `WithScope` call, because Sentry 4.x deprecates it and the labelling would otherwise stop applying silently.

Note for anyone reading the Devin page directly: its findings list is **cumulative** — the latest run still lists flags from earlier commits, including ones since fixed. Judge by the threads here, not by the count there.

CI (`pr-automation`) is green. Full C# suite at this commit: 3027 passed, 1 failed — `XMatterHelperTests.InsideBackCover_PaperAndDeviceXMatter_ShareOneCustomLayoutId`, which fails identically on clean `master` and is unrelated to this change.

JohnThomson:
[Claude Opus 5] Developer decisions on the preflight report — recorded here so they don't only live in the report. No code changed; HEAD is still `da6b1a5`.

- **Background budget stays at 40s.** The point of keeping the Sentry reports (rather than suppressing them, as the closed #8082 did) is that the new `urlLookupMode` / `urlLookupBudgetSeconds` tags make "was 40s enough?" a measurable question. If the `background` bucket stays quiet, it was; if not, we'll be tuning against data instead of guessing again.
- **A blocking caller keeps firing its own request** rather than waiting on the background one already in flight. Pre-existing behaviour, and it is what guarantees a blocking caller is never made to wait the patient budget — the freeze this change exists to avoid. The cost is one extra anonymous GET of a small JSON file.
- **A parse failure still sets the "no internet" flag.** Pre-existing, and not worth a branch for a path that has never fired. It is now at least diagnosable: the fingerprint includes the exception type, so a parse failure lands in its own Sentry bucket instead of hiding among the timeouts.
- **The blocking path keeps its (nearly unreachable) single retry.** 2.5s per attempt inside a 3s ceiling means it only helps when the first attempt fails fast — a refused connection or a DNS failure. That is the deliberate trade: responsiveness over success rate on a path where Bloom is frozen. Now documented in the code.
- **The pre-existing `master` test failure gets no card.** `XMatterHelperTests.InsideBackCover_PaperAndDeviceXMatter_ShareOneCustomLayoutId` fails identically on clean `master` (`87567c87`) and is unrelated to this PR; the developer's view is that a failing C# test will get immediate attention on its own.

Staying in draft for the developer's own review; `pr-ready-for-human` comes after that.

src/BloomTests/WebLibraryIntegration/S3RetryBudgetTests.cs
andrew-polk: It looks like you already thought about this and decided it was worth adding/keeping these tests.
But I should point out that the first reason given why we should is incorrect.
Namely, this is wrong:
> This repo does not run the C# tests in GitHub CI at all — `.github/workflows/` contains only `pr-automation.yml` and `nightly.yml`, and neither invokes `dotnet test` or BloomTests. So there is no CI container for these to be blocked in.

andrew-polk (commented):
> @andrew-polk reviewed 4 files and all commit messages, made 1 comment, and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8151)* status: all files reviewed, 2 unresolved discussions (waiting on JohnThomson).

src/BloomTests/WebLibraryIntegration/S3RetryBudgetTests.cs
JohnThomson: [Claude Opus 5] You're right, thanks for catching it — and it turned out to matter more than a footnote, so it changed the outcome rather than just the record.

**What I got wrong.** I grepped `.github/workflows/` for `dotnet test` / `BloomTests`, found nothing, and generalised "no GitHub Actions job" into "no CI". Two errors in one: Bloom's PR CI is **TeamCity**, and on top of that **`nightly.yml` does run the C# suite** — via `msbuild Bloom.proj /t:TestOnly`, which is exactly why the grep missed it. I've marked the original paragraph as wrong in place, above.

**What changed because of it.** Since these *do* run in CI, your correction reinstates the concern I had dismissed: they spend real seconds on real socket timeouts and assert an exact attempt count, which depends on agent load and on loopback being unrestricted — not something worth gating a PR on, and ~6s the gate would pay on every push forever. So in `5e4b763` they are now nightly-only.

Worth flagging for anyone doing this next, because the obvious marker is a trap: **`[Category("SkipOnTeamCity")]` would have run them in neither place**, since the nightly passes that same name in `excludedCategories`. So there's now a `Nightly` category, excluded only when TeamCity is the one running us:

```xml
<ExtraExcludeCategories Condition="'$(teamcity_version)' != ''">SkipOnTeamCity,Nightly,$(ExtraExcludeCategories)</ExtraExcludeCategories>
```

Verified by filter: `TestCategory!=Nightly` (what TeamCity will do) matches the fixture not at all; `TestCategory=Nightly` (what the nightly will do) runs both and passes.

The rest of that reply — connection reuse, the timing margin, and why an SDK that stopped retrying is precisely what the test is for — is unaffected.

andrew-polk (changes requested):
> @andrew-polk reviewed 2 files and all commit messages, made 1 comment, and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8151)* status: all files reviewed, 2 unresolved discussions (waiting on JohnThomson).

___
*[`build/Bloom.proj` line 235 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8151#-OzMwJbsAuTsCuSpWrdg:-OzMwJbt1T4QBRnGN6xP:b3x9cr7) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/3d54fe25c09ceee0118faca0ad9210531a996aeb/build/Bloom.proj#L235)):*
> ```XML
> 	       so Nightly tests do still run there, daily. -->
>       <CategoriesToExclude>$(excludedCategories)</CategoriesToExclude>
>       <CategoriesToExclude Condition="'$(teamcity_version)' != ''">SkipOnTeamCity;Nightly;$(excludedCategories)</CategoriesToExclude>
> ```

This works today, but kind of by accident.
It is true today that our nightly runs on GHA (not teamcity) and that other runs are on teamcity, but that should not be semantic. In fact, we will probably switch the main runs to GHA soonish.
Probably the "SkipOnTeamCity" category should already be "SkipInCI" or similar.

I think the right way to to do what we want here is to add the Nightly category as you have. Consider renaming to NightlyOnly.
Then, we need to add that exclusion to TeamCity's config rather than adding it to this file. In other words, I don't think we want to change this file at all.

Let me know if that doesn't make sense.

[The other thing we should figure out is how to make "NightlyOnly" and "Integration" categories not run by default for devs. That is separate from this.]

JohnThomson (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8151)* status: all files reviewed, 2 unresolved discussions (waiting on andrew-polk).

___
*[`build/Bloom.proj` line 235 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8151#-OzMwJbsAuTsCuSpWrdg:-OzN9bIz-q5NG9eo2zsW:bt1adi9) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/3d54fe25c09ceee0118faca0ad9210531a996aeb/build/Bloom.proj#L235)):*

This works today, but kind of by accident.
It is true today that our nightly runs on GHA (not teamcity) and that other runs are on teamcity, but that should not be semantic. In fact, we will probably switch the main runs to GHA soonish.
Probably the "SkipOnTeamCity" category should already be "SkipInCI" or similar.

I think the right way to to do what we want here is to add the Nightly category as you have. Consider renaming to NightlyOnly.
Then, we need to add that exclusion to TeamCity's config rather than adding it to this file. In other words, I don't think we want to change this file at all.

Let me know if that doesn't make sense.

[The other thing we should figure out is how to make "NightlyOnly" and "Integration" categories not run by default for devs. That is separate from this.]
</blockquote></details>

Reverted the Bloom.proj change

andrew-polk (approved):
> @andrew-polk reviewed 2 files and all commit messages, and resolved 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8151)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on JohnThomson).

PR #8152, opened 2026-08-05
Run the C# suite with dotnet test, so CI and developers run the same thing (BL-16656)

andrew-polk:
CI and developers did not run the C# suite the same way, so a test could be green everywhere locally and red the moment it merged.

`build/Bloom.proj` ran the suite through the **NUnit3 MSBuild task** from SIL.BuildTasks, which wraps `nunit3-console`. That runner launches a per-framework agent process whose `runtimeconfig.json` asks the host only for `Microsoft.NETCore.App`, never `Microsoft.WindowsDesktop.App`. Managed WPF assemblies still resolve (NUnit finds them via our `deps.json`), but the runtime's **native** search path is derived from what the process asked for, so any call into WPF's native code fails with `Unable to load DLL 'wpfgfx_cor3.dll'`. `dotnet test` starts the host from `BloomTests.runtimeconfig.json`, which *does* ask for WindowsDesktop, so the same tests pass.

That is exactly how BL-16597's `ImageGalleryApiTests` passed on every developer machine and produced 15 failures on TeamCity the moment they merged.

## What changed

- `RunNUnit` now runs `dotnet test` against the already-built assembly (not the csproj: several tests locate content by walking up from the assembly's own folder, e.g. `BloomFileLocatorTests.kRelativePathToBrowserFolder`).
- Results stay **NUnit3 xml at the same path** via `NunitXml.TestLogger`, so `nightly.yml` is untouched and there is **no TeamCity server-side change** — results go over `##teamcity[importData type='nunit']`.
- The `Exec` deliberately does not abort the target: it captures the exit code, emits `importData`, and only then fails. Otherwise a failing run would skip the import and TeamCity would lose exactly the results worth looking at.
- Category exclusion simplified to semicolons (what MSBuild already splits an `Include` on), dropping the comma conversion and the `Unescape` it forced.
- Drops the WPF `DllImportResolver` added to `SetUpFixture.cs` on master, which existed only to work around the console runner. It goes in this commit so reverting the switch restores it too.
- SIL.BuildTasks is left alone — the `NUnit3` task wraps one specific runner and Bloom was its only caller here.

## Verification

- **The real target, end to end**: `msbuild Bloom.proj /t:TestOnly` → 3027 tests, 0 failures, 12 skipped, 5m10s, exit 0, valid NUnit3 xml written to the path `nightly.yml` publishes.
- Filter construction checked for every caller shape: TeamCity with `Integration`, the nightly with `SkipOnTeamCity`, TeamCity with nothing passed, nothing passed at all, and the comma guard firing.
- vstest confirmed to honour the two-clause form: `BloomS3ClientTests` runs 31 tests without the `Integration` exclusion and 30 with it.
- `ImageGalleryApiTests` 23/23 with the resolver removed.

## Still to prove on CI

TeamCity reports **3687** C# test occurrences where only **3027** run locally. Sampling the TeamCity results confirms the difference is C#, not front-end vitest — most likely `TestCaseSource` data that only exists on a build agent. Their behaviour under VSTest is unconfirmed, so this wants one TeamCity run and one dispatched C#-only nightly before merge.

Ref: https://issues.bloomlibrary.org/youtrack/issue/BL-16656

[Devin review](https://devinreview.com/BloomBooks/BloomDesktop/pull/8152)

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8152)

### DIFF
--- a/build/Bloom.proj
+++ b/build/Bloom.proj
@@ -59,7 +59,6 @@
   <UsingTask TaskName="MakeWixForDirTree" AssemblyFile="$(BuildTasksDll)" Condition="Exists('$(BuildTasksDll)')" />
   <UsingTask TaskName="Split" AssemblyFile="$(BuildTasksDll)" Condition="Exists('$(BuildTasksDll)')" />
   <UsingTask TaskName="FileUpdate" AssemblyFile="$(BuildTasksDll)" Condition="Exists('$(BuildTasksDll)')" />
-  <UsingTask TaskName="NUnit3" AssemblyFile="$(BuildTasksDll)" Condition="Exists('$(BuildTasksDll)')" />
   <UsingTask TaskName="S3BuildPublisher" AssemblyFile="$(BuildTasksAwsDll)" Condition="Exists('$(BuildTasksAwsDll)')" />
   <Import Project="$(MSBuildCommunityTasksPath)/MSBuild.Community.Tasks.Targets"/>
   <UsingTask TaskName="WebDownload" AssemblyFile="$(MSBuildCommunityTasksPath)/MSBuild.Community.Tasks.dll"/>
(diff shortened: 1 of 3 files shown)


build/Bloom.proj
```diff
@@ -193,21 +192,64 @@
...
+    <Error Condition="$(CategoriesToExclude.Contains(','))"
+		   Text="excludedCategories is semicolon separated, but got '$(CategoriesToExclude)'"/>
+    <ItemGroup>
+      <ExcludedCategory Include="$(CategoriesToExclude)"/>
+    </ItemGroup>
+    <PropertyGroup>
+      <TestFilterArgument Condition="'@(ExcludedCategory)' != ''">--filter "@(ExcludedCategory->'TestCategory!=%(Identity)', '&amp;')"</TestFilterArgument>
+    </PropertyGroup>
+    <Message Importance="high" Text="Running C# tests: @(TestAssemblies) $(TestFilterArgument)"/>
+    <Exec Command='dotnet test "@(TestAssemblies)" --nologo $(TestFilterArgument) --logger:"nunit;LogFilePath=$(TestResultsXmlFile)"'
```
andrew-polk, line 253: [Devin] **Bug**: A crashed test run can report the previous run's results as if they were this run's

The results file is handed to TeamCity via `##teamcity[importData ...]` at a fixed path, but nothing deleted that path before the run. `NunitXml.TestLogger` only writes the file once a run completes, so a run that dies before that — crashed test host, a filter `dotnet test` rejects, no `dotnet` on PATH — would leave the *previous* run's file sitting there to be imported.

The build would still go red on the exit code, so this could never hide a failure. But the report a human then opens would be a different run's, and plausibly all green — which is worse than no report at all. TeamCity checkouts persist between builds, which is what makes this reachable there; fresh GHA runners are not affected.

build/Bloom.proj
```diff
@@ -193,21 +192,64 @@
...
+		  WorkingDirectory="$(RootDir)/output/Tests/$(Configuration)/x64"
+		  ContinueOnError="true">
```
andrew-polk, line 255: [Devin] **Investigate**: Tests run against a prebuilt DLL, so TestOnly can silently exercise stale code

`dotnet test "@(TestAssemblies)"` targets the already-built `BloomTests.dll`, which behaves like `--no-build`. So `TestOnly` invoked on its own runs whatever DLL happens to be sitting in the output folder, and `AGENTS.md` warns about exactly that hazard ("A stale DLL can cause tests to pass or fail against an old version of the code").

Devin notes this is not a regression, and asks whether the team is happy that the guidance now sits alongside how the build target itself runs the suite.

build/Bloom.proj
```diff
@@ -193,21 +192,64 @@
...
+	     reachable by setting the property in the environment: on the command line MSBuild reads a
+	     comma as its own separator between properties, so /p:excludedCategories=A,B never gets
+	     this far. Worth catching anyway, since the failure it prevents is a silent one. -->
+    <Error Condition="$(CategoriesToExclude.Contains(','))"
+		   Text="excludedCategories is semicolon separated, but got '$(CategoriesToExclude)'"/>
+    <ItemGroup>
+      <ExcludedCategory Include="$(CategoriesToExclude)"/>
+    </ItemGroup>
+    <PropertyGroup>
+      <TestFilterArgument Condition="'@(ExcludedCategory)' != ''">--filter "@(ExcludedCategory->'TestCategory!=%(Identity)', '&amp;')"</TestFilterArgument>
```
andrew-polk, line 243: [Devin] **Investigate**: vstest filter semantics for [Explicit] tests differ from the console runner

The NUnit3 VSTest adapter has historically run `[Explicit]` tests when a test-case filter is supplied, because it cannot tell an explicitly-selected test from a filtered-in one. Since `--filter "TestCategory!=..."` is now always present on TeamCity, an `[Explicit]` test could start executing on CI where `nunit3-console` skipped it — and Devin suggests this as a possible contributor to the unexplained TeamCity/local test-count gap.

build/Bloom.proj
andrew-polk: [Claude Opus 5 during preflight] Good catch, and it was real. Fixed in `b8ff1d6686`: the target now deletes the results file before running, so a run that dies before writing one leaves nothing to import and a missing report reads as missing.

Verified rather than assumed: planted a stale results file, then ran a configuration whose test assembly does not exist so `dotnet test` dies before writing anything. The stale file is gone afterwards, where before the fix it survived to be imported.

build/Bloom.proj
andrew-polk: [Claude Opus 5 during preflight] Not an issue, and Devin says as much itself: this is not a regression. The old NUnit3 task also ran the prebuilt assembly, so `TestOnly` behaves exactly as it did before this PR.

The split is deliberate. `Test` depends on `Build`, so the normal entry point always builds first; `TestOnly` exists precisely so CI can build once and test separately, which is how both systems already invoke it. The `AGENTS.md` warning is aimed at an agent running `dotnet test --no-build` by hand against a tree it just edited, not at a build target whose caller has just built.

Running the csproj instead is not an option here: several tests locate content by walking up from the assembly's own folder (`BloomFileLocatorTests.kRelativePathToBrowserFolder`), so rebuilding into the default bin folder would change what those paths resolve to.

build/Bloom.proj
andrew-polk: [Claude Opus 5 during preflight] False positive. The only `[Test, Explicit]` in the suite is `ImageUtilsTests.DiagnoseLineArtColorBuckets`, and it sits inside an `#if false` block, so it is not compiled into the assembly at all and cannot run under any runner.

Confirmed empirically as well: it does not appear anywhere in the results of the filtered run. So this cannot be contributing to the TeamCity/local count gap — that remains open and is called out in the PR description as needing a real CI run.

build/Bloom.proj
```diff
@@ -193,21 +192,71 @@
...
+	both TeamCity's importData and the nightly's publish action already understand) and hand
+	TeamCity the file. The Exec must not abort the target on failure, or a failing run would
+	skip the import and TeamCity would lose exactly the results worth looking at; so we capture
+	the exit code, import, and only then fail.
+  -->
   <Target Name="RunNUnit">
     <PropertyGroup>
-      <TestResultsXmlFile Condition="$(isOnTeamCity) == false">$(RootDir)/output/Tests/$(Configuration)/x64/TestResults.xml</TestResultsXmlFile>
-      <ExtraExcludeCategories Condition="'$(teamcity_version)' != ''">SkipOnTeamCity,$(ExtraExcludeCategories)</ExtraExcludeCategories>
+      <TestResultsXmlFile>$(RootDir)/output/Tests/$(Configuration)/x64/TestResults.xml</TestResultsXmlFile>
```
andrew-polk, line 221: [Devin] **Investigate**: TestResultsXmlFile now always set (previously TeamCity-only-empty)

Previously `TestResultsXmlFile` was set *only when not on TeamCity*, because on TeamCity the NUnit3 task reported results through service messages rather than through a file. It is now always set, always written, and always imported — the intended mechanism change in this PR.

Devin confirms no stale reference to the old behaviour remains in the file. What it asks a reviewer to confirm is external to this repo: that **the TeamCity build configuration itself no longer depends on the console runner's own service-message reporting**, which disappears with the runner, and instead consumes the imported file.

Leaving this thread open — the TeamCity build config lives on the server, not in this repo, so it is not something this branch can verify.

andrew-polk:
[Claude Opus 5 during preflight] Consulted Devin on 2026-08-05 17:55 UTC up to commit `b8ff1d6686a7bce68a2a32d48f2d923b153f3e04`.

Two jobs ran. The first, on `e2ff8c538a`, raised 1 Bug and 2 Investigate flags (plus 2 informational, not mirrored):

- **Bug — stale results could be imported after a dead run.** Real, and in this PR's new code. Fixed in `b8ff1d6686` and verified by planting a stale results file and running a configuration whose test assembly does not exist. [Thread](https://github.com/BloomBooks/BloomDesktop/pull/8152#discussion_r3722764999)
- **Investigate — tests run against a prebuilt DLL.** Not an issue: unchanged from the previous NUnit3 task, and `Test` still depends on `Build`. [Thread](https://github.com/BloomBooks/BloomDesktop/pull/8152#discussion_r3722765143)
- **Investigate — `[Explicit]` tests under a vstest filter.** False positive: the suite's only `[Test, Explicit]` sits inside an `#if false`, so it is not compiled. [Thread](https://github.com/BloomBooks/BloomDesktop/pull/8152#discussion_r3722765291)

The re-review on `b8ff1d6686` (4 informational) repeated those three and added one new flag. Note the repeated Bug still cites `Bloom.proj:251` for the importData line, which the fix moved to 258 — it re-emitted its original text rather than re-analysing the fixed code, and the fix is verified above.

- **Investigate — `TestResultsXmlFile` is now always set** (it used to be empty on TeamCity, which reported via service messages instead). Left **open**: it asks whether the TeamCity build configuration still relies on the runner's service-message reporting, and that config lives on the server rather than in this repo. [Thread](https://github.com/BloomBooks/BloomDesktop/pull/8152#discussion_r3722806115)

CI (`pr-automation`) is green on this commit. CodeRabbit is present but disabled in `.coderabbit.yml` (`auto_review.enabled: false`), so it is not a reviewer on this repo.

build/Bloom.proj
andrew-polk: [Claude Opus 5] Developer decision: leave as is.

We are not changing anything here and not pre-emptively touching the TeamCity build configuration. The mechanism this PR relies on is self-contained — the target writes NUnit3 xml and hands TeamCity the file via `importData` — and that is verified locally end to end. If the build configuration turns out to still expect the old service-message reporting, it will show up as an empty or wrong Tests tab on the first TeamCity run of this branch, which is a visible and easily corrected outcome rather than a silent one.

Resolving on that basis.

PR #8154, opened 2026-08-05
Shrink large images, reporting wherever the caller already reports (BL-16646)

StephenMcConnel:
`MigrateToMediaLevel1ShrinkLargeImages` shrinks images larger than 3840x2800 the first time an old book (pre-4.9, still at `mediaMaintenanceLevel` 0) is brought up to date, and showed its own "Updating Image Files" progress dialog while it worked. Several of the operations that bring a book up to date run on a **background thread**, and WinForms only allows a `Form` to be created on the UI thread — so on an old book with an oversized image, those paths could hang or leave a stray window.

The fix decides where to report based on **whether the caller already has somewhere real to report, and whether we could legally put up a dialog at all**:

- **The caller has a real progress** — Do Updates of All Books, `.bloomSource` and spreadsheet import, the publish previews, and the single-book "Update Book" command. Its messages go there, so they land in the window the user is already watching and we add nothing of our own.
- **All the caller has is a `NullProgress`, and we are on the UI thread** — most importantly clicking a book to select it. There is nowhere to report, and doing the work inline would freeze Bloom, so we put up our own dialog, which is legal here and which runs the work on a background worker so the app stays responsive.
- **We could not show a dialog even if we wanted to** — off the UI thread (WinForms forbids it), headless (no window exists), or under test. The caller's progress is used, or a `NullProgress` if it passed none.

`MigrateToMediaLevel1ShrinkLargeImages` gained an optional `IProgress` parameter for this, passed at `Book.cs:1110` and `Book.cs:1898`. Only tests rely on the default — both production call sites pass something non-null, since `Book.EnsureUpToDate` substitutes a `NullProgress` for a null argument.

### How a failed shrink is handled

This took several passes to settle, so here is the whole of it in one place.

Originally a failure inside the dialog's background worker vanished — `ProgressDialogBackground` never reads `RunWorkerCompletedEventArgs.Error` — and we set `mediaMaintenanceLevel` to 1 anyway, permanently, since the level is never revisited. That book's images stayed oversized forever with nothing reported.

Now, in **both** branches:

- The failure is caught and reported **passively** via `NonFatalProblem.Report(ModalIf.None, PassiveIf.All, …)`: a toast, plus the whole exception to the log and Sentry. It is passive on purpose — the book still works, its pictures are merely left large, and we will try again — so it is not worth interrupting the user for.
- The failure is also written to the caller's progress when it has one, so an operation already showing a progress box does not appear to have finished cleanly. This deliberately uses `WriteWarning`, **not** `WriteError`: `ProgressDialogForeground` pops a modal "There was a problem performing that operation" whenever its progress records an error, which would defeat the point of reporting passively.
- The user-visible text is localized (`ImageUtils.ShrinkingImagesFailed`, in `BloomLowPriority.xlf` as a rarely-seen failure message). The book's folder path goes in the details that reach the log and Sentry, not in the sentence the user reads.
- `mediaMaintenanceLevel` is bumped **only on success**, so the shrink is retried the next time that book is brought up to date. Retrying is safe: `FixSizeAndTransparencyOfImagesInFolder` only resizes a file whose current dimensions differ from the desired ones, so images shrunk on an earlier attempt are skipped rather than re-encoded.
- **The retry does not happen twice in one pass.** `Book.EnsureUpToDate` calls this migration twice — once by way of `EnsureUpToDateMemory` and once directly afterwards — and the second call was a no-op only because the first had always bumped the level. A flag on the `BookStorage` remembers the failed attempt, so one pass makes one attempt.

The consequence worth a reviewer's eye: a bad image now costs **that one book's shrink** rather than the rest of the operation. Previously an exception from the direct branch abandoned the remaining twelve migrations and the `Save`, and for Do Updates of All Books silently ended the whole batch at the first bad image.

Also fixed along the way: `Shell.GetShellOrOtherOpenForm()` returns null in the headless CLI, and the `InvokeRequired` test would have thrown a `NullReferenceException` there — on exactly the old-book-with-a-huge-image case this ticket is about.

Supporting change: `WebProgressAdapter` now accepts the `IWebSocketProgress` interface rather than the concrete `WebSocketProgress`, so `MakeDerivativeFromBloomSourceFile` can forward the progress it is handed.

### Notes for the reviewer

The branch has been through several designs; the history shows marshal-the-dialog, delete-the-dialog-entirely, and two shapes of hybrid. Some resolved Devin threads on this PR discuss code that is no longer here, and each is annotated to say so.

One consequence worth a look: when the caller has a real progress and happens to be on the UI thread, the shrink runs **inline** on that thread rather than on a background worker. Today the only such caller is "Update Book", which runs under `ProgressDialogForeground` and so already does the whole of `BringBookUpToDate` on the UI thread — its progress pumps `Application.DoEvents` on every message, so the dialog keeps painting and nothing new freezes. It does mean a future UI-thread caller passing a live progress that does *not* pump would block the UI for the whole shrink. The invariant is written down in a comment at the branch.

Known limitations, recorded rather than fixed here, and now tracked as [BL-16663](https://issues.bloomlibrary.org/youtrack/issue/BL-16663):

- `FixSizeAndTransparencyOfImagesInFolder` has no per-file try/catch, so one unreadable image aborts its loop at the same point on every attempt and the images after it are never shrunk. The retry above therefore cannot get past such a file.
- Its most likely failures do not throw at all — a missing or failing GraphicsMagick just returns `false`, and one of the two call sites ignores that — so those books are marked migrated with oversized images. The success flag above only catches failures that throw.

Both fixes change a method with other callers, so they need their own change and test pass.

Ref: https://issues.bloomlibrary.org/youtrack/issue/BL-16646

[Devin review](https://devinreview.com/BloomBooks/BloomDesktop/pull/8154)

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8154)


src/BloomExe/Book/BookStorage.cs
```diff
@@ -4084,24 +4084,53 @@ public void MigrateToMediaLevel1ShrinkLargeImages()
...
+                        // This shouldn't happen in normal operation, but just in case...
+                        ImageUtils.FixSizeAndTransparencyOfImagesInFolder(
+                            FolderPath,
+                            new List<string>(),
+                            new NullProgress()
                         );
                     }
+                    else if (shell.InvokeRequired)
+                    {
+                        shell.Invoke(UpdateImagesWithProgressDialog);
```
StephenMcConnel: [Devin] **Investigate**: Blocking Invoke from a background worker risks deadlock if the UI thread ever waits on that worker

`shell.Invoke(...)` blocks the calling background thread until the UI thread runs the delegate, and the delegate itself shows a modal dialog, so the block lasts for the whole shrink operation. This is safe only where the UI thread is pumping messages and not synchronously waiting on the worker. Worth confirming for each background path named in the description (Do Updates of All Books, BloomPub/ePUB publish preview, .bloomSource/spreadsheet import): any place where the UI thread does a blocking wait (e.g. `Thread.Join`, `Task.Wait`, `WaitOne`) while the worker is doing `BringBookUpToDate` would deadlock rather than merely showing a nested dialog.

StephenMcConnel: [Claude Opus 5 (1M context) during preflight] Checked this — not an issue on any of the paths, so no deadlock. We walked each background path Devin names and looked for a UI thread that blocks without pumping messages:

- **Do Updates of All Books** (`CollectionTab/CollectionModel.cs:1083`) — the UI thread is sitting inside the outer progress dialog's `ShowDialog()`. A modal message loop pumps, so the `Invoke` gets serviced. What you get here is the nested-dialog behavior, not a hang (see the note in the PR description).
- **`.bloomSource` import and the BloomPub/ePUB publish preview** — both go through `MiscUI/BrowserProgressDialog.cs` (`dlg.ShowDialog(owner)`), again a pumping modal loop.
- **`external/process-book`** (`web/controllers/ExternalApi.cs`) — deliberately keeps the UI thread free: it replies to the request first and dispatches the job on `Task.Run`, and it already round-trips through `InvokeOnUiThread` before `BookProcessor.ProcessBook`, so the shell is demonstrably responsive at that moment.
- **The single-book "Update Book" command** (`CollectionModel.BringBookUpToDate`) uses `ProgressDialogForeground`, which runs the work *on* the UI thread — so `InvokeRequired` is false there and this branch is never taken.

We also checked for lock inversion, since that would deadlock even against a pumping loop: no lock is held at either call site. `Book.cs:1898` runs after the `lock (_updateLock)` block has been released, and `Book.cs:1110` holds nothing.

No `Thread.Join`, `Task.Wait`, or `WaitOne` on the UI thread was found on any of these paths. Resolving.

StephenMcConnel:
[Claude Opus 5 (1M context) during preflight] Consulted Devin on 2026-08-05 up to commit `462f97734351b1f447e1b934af86bc8a129a708d`.

It found no bugs. It raised one thing worth investigating — whether the blocking `Invoke` could deadlock rather than just stack a dialog — which we checked path by path and resolved as not an issue; see [that thread](https://github.com/BloomBooks/BloomDesktop/pull/8154#discussion_r3723872372) for the reasoning.

Two further items were informational and not mirrored as threads, but both bear on the design question raised in the PR description, so they are worth a reviewer's eye: that `InvokeRequired` reports false for a form whose window handle isn't created yet (a known rough edge of this idiom, shared with the existing call sites in `AudioRecording` and `OneDriveUtils`), and that the headless/bulk-upload path now shrinks images with a `NullProgress`, so a slow run produces no console diagnostics — something the card's `IProgress` alternative would have preserved.

CI (`pr-automation`) is green.

src/BloomExe/Book/BookStorage.cs
StephenMcConnel: [Claude Opus 5 (1M context) during preflight] Following up so this thread is not misleading to a later reader: the code it is about no longer exists. The branch has since been reworked to the approach BL-16646's card actually asked for — the `ProgressDialogBackground` is deleted and the migration reports through the caller's `IProgress` instead. There is no `shell.Invoke` and no dialog on this path any more, so the deadlock question is moot rather than merely answered.

src/BloomExe/Book/Book.cs
```diff
@@ -1107,7 +1107,7 @@ public void EnsureUpToDate(IProgress progress = null, bool forCopyOfUpToDateBook
             EnsureUpToDateMemory(progress);
             UpdateSupportFiles();
 
-            Storage.MigrateToMediaLevel1ShrinkLargeImages();
+            Storage.MigrateToMediaLevel1ShrinkLargeImages(progress);
```
StephenMcConnel, line 1110: [Devin] **Investigate**: Most callers pass NullProgress, so an old book with huge images now freezes the UI silently

Removing the modal dialog means the shrink runs synchronously on whatever thread calls `EnsureUpToDate`. Many UI-thread call sites pass `new NullProgress()` or nothing at all (e.g. `src/BloomExe/Book/BookSelection.cs:35`, `src/BloomExe/Edit/EditingModel.cs:848`, `src/BloomExe/web/controllers/BookCommandsApi.cs:477`), so on a pre-4.9 book with oversized images Bloom will appear hung for the duration with no feedback where it previously showed "Updating Image Files". `DoUpdatesOfAllBooks` and the .bloomSource import do get real progress channels, so those are fine. If the silent-freeze case matters, an option is to pass the collection/edit progress channel (or a status-bar progress) at those call sites rather than `NullProgress`.

StephenMcConnel: [Claude Opus 5 (1M context) during preflight] Confirmed, and it is sharper than "no feedback" — on the book-selection path the shrink now runs **on the UI thread**, so Bloom genuinely freezes rather than merely going quiet.

The evidence: the `selectBook` endpoint is registered with `handleOnUiThread: true` (`src/BloomExe/web/controllers/CollectionApi.cs:239`), and it reaches `BookSelection.SelectBook` → `book.EnsureUpToDate()` with no progress argument (`src/BloomExe/Book/BookSelection.cs:35`), which `EnsureUpToDate` turns into a `NullProgress` at `Book.cs:1068`. So the shrink executes inline on the UI thread with nothing to report to.

Worth being explicit about what was lost: the old `ProgressDialogBackground` was not only feedback. `ShowAndDoWork` ran the work on a `BackgroundWorker` while the UI thread pumped messages inside `ShowDialog()`, so on this path the old code kept Bloom responsive *and* showed a progress bar. The new code does neither. On the paths that own a real progress channel — Do Updates of All Books, `.bloomSource` import, publish preview — none of this applies and the behavior is fine or better.

I am not deciding this one: choosing where these call sites should get a progress channel (or whether the freeze is acceptable given how rare a pre-4.9 book with an oversized image is) is a product call. Escalated to the developer, so I am leaving this thread open until it comes back.

src/BloomExe/Book/BookStorage.cs
```diff
@@ -4073,30 +4078,11 @@ public void MigrateToMediaLevel1ShrinkLargeImages()
                 // This update can be very slow, so encourage the user that something is happening.
```
StephenMcConnel, line 4099: [Devin] **Investigate**: PR description does not match the implementation

The description says the dialog is kept and marshaled onto the UI thread via `Shell.GetShellOrOtherOpenForm()`/`Invoke` in a new `UpdateImagesWithProgressDialog` helper, but the head code takes the other approach the description calls the "card's suggested fix": the WinForms dialog is deleted and the caller's `IProgress` is threaded in (`src/BloomExe/Book/BookStorage.cs:4050`, `src/BloomExe/Book/Book.cs:1110`, `src/BloomExe/Book/Book.cs:1898`). No `Shell.GetShellOrOtherOpenForm` call exists for this path. Reviewers relying on the description (including the "nested modal dialog under Do Updates of All Books" caveat) will be evaluating the wrong design; the description should be updated before merge.

StephenMcConnel: [Claude Opus 5 (1M context) during preflight] This was correct when Devin first raised it, and it is already fixed — the description was rewritten earlier in this preflight run, before this commit was pushed.

The description now leads with "Rather than teach the dialog which thread it is on, this removes it," and describes the `IProgress` threading as the implemented design rather than as a road not taken. The "nested modal dialog under Do Updates of All Books" caveat Devin mentions is gone, because that consequence belonged to the old approach. The PR title was updated to match as well.

What is probably still triggering the flag is one sentence deliberately kept in the reviewer note: "This replaces an earlier approach on this branch that kept the dialog and marshalled it onto the UI thread with `Shell.GetShellOrOtherOpenForm()` and `Invoke`." That is there on purpose, to explain to a human why an earlier resolved thread on this PR discusses a blocking `Invoke` that no longer exists anywhere in the code. It describes history, not the head implementation. Resolving.

StephenMcConnel:
[Claude Opus 5 (1M context) during preflight] Consulted Devin on 2026-08-05 up to commit `bb09b0d61735fc33b96f930b65ae7b79a9ad9358`.

This was a re-review after the branch was reworked to delete the progress dialog rather than marshal it onto the UI thread. No bugs. Three things needed investigating and four were informational.

The one that matters is [the loss of feedback on the `NullProgress` call sites](https://github.com/BloomBooks/BloomDesktop/pull/8154#discussion_r3724383011). I verified it and it is real — on book selection the shrink now runs on the UI thread with nothing to report to, so Bloom freezes where the old dialog kept it responsive and showed a bar. That thread is deliberately left **open**: deciding where those call sites should get a progress channel is a product call, so it has gone to the developer.

Of the rest, [the stale PR description](https://github.com/BloomBooks/BloomDesktop/pull/8154#discussion_r3724386594) was a fair catch that had already been fixed earlier in this run, and is resolved. Devin also re-raised the blocking-`Invoke` deadlock question against code the rework deleted; that is [the existing thread from the previous run](https://github.com/BloomBooks/BloomDesktop/pull/8154#discussion_r3723872372), already resolved and annotated to say the code is gone, so it was not posted again.

The four informational items were not mirrored: a missing null guard on the new `IProgress` parameter (verified unreachable — `Book.EnsureUpToDate` substitutes a `NullProgress` at `Book.cs:1068`, and every other caller passes one explicitly), migration messages now surfacing in the import dialog (intended), a doc comment still describing the removed dialog (fixed in this very commit, so Devin's copy is stale), and the `InvokeRequired`-before-handle-creation edge (moot, that code is deleted).

CI (`pr-automation`) is green.

src/BloomExe/Book/Book.cs
StephenMcConnel: [Claude Opus 5 (1M context) during preflight] Status update, since a new design has landed on the branch and this thread is still the right place for the question.

`7b14a6d5a` attempts to address this by reinstating the "Updating Image Files" dialog for callers that have nothing better, keeping the direct call for those holding a `WebProgressAdapter`. **It does not yet fix the freeze**, and the branch selection is inverted.

The branch is chosen on `shell.InvokeRequired`, which is `true` when the caller is *not* on the UI thread. So:

- The **synchronous, no-feedback** path is taken when `!shell.InvokeRequired` — that is, precisely when we *are* on the UI thread. Book selection (`collections/selected-book` is registered `handleOnUiThread: true`) still runs the whole shrink inline on the UI thread with a `NullProgress`. The freeze this thread is about is unchanged.
- The **dialog** path is taken when `shell.InvokeRequired` is true — precisely when we are *not* on the UI thread, and there is no `Invoke` anywhere. So `ProgressDialogBackground` is now constructed and shown on a background thread, which is the original BL-16646 complaint. Worth noting honestly: a probe on .NET 8 showed this does not actually throw, so it is not a guaranteed crash — but it is an ownerless top-level window created on a foreign thread, which is a good way to produce the "Bloom appears hung" report this ticket describes.

Swapping the two branches would fix both at once: show the dialog when `!shell.InvokeRequired` (on the UI thread, where a dialog is legal and where the feedback is missing), and go inline with the caller's `progress` when `shell.InvokeRequired` (on a background thread, which is where every caller that owns real progress already lives). That also removes the second-dialog nesting under Do Updates of All Books, since that caller would then report into the progress dialog it already has open.

Still leaving this open — the swap changes user-visible behavior on several paths, so it is the developer's call rather than mine. It is the lead item in this run's preflight report.

src/BloomExe/Book/BookStorage.cs
```diff
@@ -4073,13 +4078,21 @@ public void MigrateToMediaLevel1ShrinkLargeImages()
...
                 // NO images should have transparency removed.  See https://issues.bloomlibrary.org/youtrack/issue/BL-8846.
 
-                if (Program.RunningUnitTests)
+                var shell = Shell.GetShellOrOtherOpenForm();
+                // shell is null when no window is open at all -- the headless CLI (bulk upload,
+                // hydrate). There is nothing to show a dialog on, and nothing to ask about thread
+                // affinity, so just do the work against whatever progress we were handed.
+                if (
```
StephenMcConnel, line 4122: [Devin] **Bug**: Image-shrinking progress window is opened on the wrong thread, so background book updates can hang instead of showing progress

The choice of when to open the image-shrinking progress window is reversed (`!shell.InvokeRequired` at `src/BloomExe/Book/BookStorage.cs:4089`), so it is opened exactly on the background operations where it is unsafe and skipped during ordinary foreground work where it is wanted.

Impact: Operations like updating all books or building a publish preview can freeze or fail on old books with oversized images, while a user who simply opens such a book sees no progress at all.

<details>
<summary>Why the InvokeRequired test is backwards</summary>

`Form.InvokeRequired` is true when the current thread is NOT the UI thread. The new condition takes the "just do the work silently against the caller's progress" branch when `!shell.InvokeRequired`, i.e. when we ARE on the UI thread — the one situation where `ProgressDialogBackground` (a WinForms `Form` shown with `ShowDialog`, see `src/BloomExe/ToPalaso/ProgressDialogBackground.cs:22-38`) is legal and useful. Conversely, when `shell.InvokeRequired` is true (Do Updates of All Books, publish preview, `.bloomSource`/spreadsheet import, bulk upload — all background threads), control falls to the `else` branch, which constructs and modally shows the form on that background thread. That is precisely the situation BL-16646 reports as hanging/throwing.

The fix is to invert the thread test: use the caller's `IProgress` when `shell.InvokeRequired` (background), and show the dialog only when we are on the UI thread.

</details>

StephenMcConnel: [Claude Opus 5 (1M context) during preflight] Confirmed independently, twice, before Devin filed this — and the prescribed fix matches.

A local review agent verified the WinForms semantics empirically with a throwaway net8.0-windows probe rather than from memory: `InvokeRequired` is `False` on the control's creating thread and `True` on a `BackgroundWorker` thread. It is also the established idiom in this repo — `web/controllers/ApiRequest.cs:267` and `CollectionTab/CollectionModel.BloomSourceImport.cs:244` both use `InvokeRequired` as the trigger to `Invoke` *onto* the UI thread.

Caller-by-caller, at this commit:

| Path | progress | thread | branch taken |
| --- | --- | --- | --- |
| Do Updates of All Books | `MultiProgress` of the dialog it already opened | background worker | **dialog** — a second one, nested |
| Spreadsheet import | `NullProgress` | background worker | **dialog** — a stray WinForms window over the React one |
| `.bloomSource` derivative import | `WebProgressAdapter` | background worker | direct — correct |
| Book selection (`handleOnUiThread: true`) | `NullProgress` | **UI thread** | direct — frozen, silent |
| Bring Book Up To Date | `MultiProgress` of `ProgressDialogForeground` | UI thread | direct — correct |
| Bulk upload / hydrate CLI | `NullProgress` | no window | direct, via the null-shell guard added in this commit |

One correction to the ticket's framing, in fairness to the code: the probe found that .NET 8 does **not** throw on `new Form()` + `ShowDialog()` from an MTA thread-pool thread — the window appears and its modal loop pumps on that thread. So the `else` branch is not a guaranteed crash. What it is: an ownerless top-level window created on a foreign thread, not modal to the shell and with no owner to fix its z-order, registered process-wide in `Application.OpenForms` while living on a thread that stops pumping the instant it closes, holding a thread-pool thread for the whole multi-minute shrink. A window stuck behind the shell while the shell cannot proceed is exactly what a user reports as "Bloom hangs".

Two further consequences of the dialog branch worth recording here, since they disappear if the test is inverted:

1. **A failed shrink is recorded as successful.** `ProgressDialogBackground.ShowAndDoWork` wires `RunWorkerCompleted += (sender, e) => Close();` and never reads `e.Error` (`ToPalaso/ProgressDialogBackground.cs:30`), so an exception inside the worker vanishes; control returns normally and `Dom.UpdateMetaElement("mediaMaintenanceLevel", "1")` still runs. The book is then permanently marked migrated and never retried, with nothing shown to the user or the log. The direct branch propagates instead — so the same failure has opposite outcomes depending only on which thread the caller was on.
2. **`progress is WebProgressAdapter` is not a reliable test for "this caller already shows progress".** Do Updates of All Books has a visible WinForms progress dialog but passes a `MultiProgress`, and spreadsheet import has a visible browser dialog but passes `NullProgress`. Both therefore get a second dialog.

Not resolving this: inverting the test changes what the user sees on several paths, so it is the developer's call, not mine. It is the lead item in this run's preflight report, with the swap as the recommended option.

src/BloomExe/Book/BookStorage.cs
```diff
@@ -4058,10 +4062,11 @@ public void MigrateToMediaLevel1ShrinkLargeImages()
             {
                 // If the book contains overlarge images, we want to fix those before editing because this can lead
                 // to thumbnails not being created properly and other bad behavior.  This is a one-time fix that can
-                // permanently change the images in the original book folder.  If any images must be shrunk, then a
-                // progress dialog pops up because that can be a very slow process.  If nothing needs to be done,
-                // nothing will appear on the screen, and it usually takes a small fraction of a second to determine
-                // this.
+                // permanently change the images in the original book folder.  Shrinking can be very slow, so we
```
StephenMcConnel, line 4079: [Devin] **Investigate**: Doc comment and BookProcessor comment describe a dialog that still exists

The block comment and doc comment on `MigrateToMediaLevel1ShrinkLargeImages`, and the comment in `BookProcessor`, state that progress is reported through the caller's `IProgress` rather than by putting up a dialog, and that BL-16646 removed the dialog. This commit reinstates the dialog for some callers, so those comments now describe the previous design rather than the code they sit next to.

StephenMcConnel: [Claude Opus 5 (1M context) during preflight] Correct, and I wrote two of the three offending comments in the previous run, when they were accurate. The `BookProcessor` one is the most consequential: I had reduced its justification to the layout reason alone, on the grounds that BL-16646 had removed the dialog. With a dialog back on the background threads, the thread-safety reason for that pre-shrink is arguably live again, so that comment now understates why the ordering matters.

Deliberately not fixing these yet. They are downstream of the inverted thread test on this PR, and rewriting them to describe the current behavior would amount to documenting the bug as though it were intended. Whichever way that decision goes, all three comments need one pass to match — that is recorded in the preflight report alongside it, so it does not get lost.

src/BloomExe/Book/BookStorage.cs
```diff
@@ -4047,7 +4047,11 @@ public void MigrateMaintenanceLevels()
         /// memory. However, it is still helpful for performance and reducing published file sizes.
         /// Does nothing if mediaMaintenanceLevel indicates it has already been done.
         /// </summary>
-        public void MigrateToMediaLevel1ShrinkLargeImages()
+        /// <param name="progress">Where to report the (potentially very slow) shrinking, so the
+        /// user can see why we are busy. Callers pass whatever progress they already own; headless
+        /// callers such as the bulk-upload CLI pass a NullProgress and so report nothing. Must not
+        /// be null: ImageUtils.FixSizeAndTransparencyOfImagesInFolder dereferences it.</param>
+        public void MigrateToMediaLevel1ShrinkLargeImages(IProgress progress)
```
StephenMcConnel: [Devin] **Investigate**: Dialog branch's lambda parameter shadows the new method parameter

In the dialog branch, `dlg.ShowAndDoWork((progress, args) => ImageUtils.FixSizeAndTransparencyOfImagesInFolder(..., progress))` declares a lambda parameter named `progress`, which shadows the method's own new `IProgress progress` parameter. Inside the lambda, `progress` refers to the dialog's own progress object, so the caller's progress is not used on this path even though the code reads as though it is.

StephenMcConnel: [Claude Opus 5 (1M context) during preflight] Confirmed, including that it compiles — I checked, because shadowing an enclosing parameter is illegal for ordinary locals. For lambda parameters it is permitted, and it builds clean under this project's language version, so this is a readability-and-intent problem rather than a compile error.

The substantive part is the one Devin names: on the dialog path the caller's `progress` is genuinely dead, and the dialog's own `MultiProgress` is what `ImageUtils` writes to. That is what the dialog needs in order to render, so it is not wrong in itself — but it is how the stated intent of this design ("callers that own real progress report into it") quietly stops applying to every caller that lands in this branch. Do Updates of All Books is the clearest case: its existing progress log receives none of the "Preparing image: {0}" messages.

Leaving this open, tied to the inverted-thread-test decision on this PR. If the dialog branch survives that decision, the lambda parameter wants a distinct name so the shadowing is not silently load-bearing; if the branches are swapped, the question mostly evaporates.

src/BloomExe/Book/BookStorage.cs
```diff
@@ -4073,13 +4078,21 @@ public void MigrateToMediaLevel1ShrinkLargeImages()
                 // This update can be very slow, so encourage the user that something is happening.
```
StephenMcConnel, line 4099: [Devin] **Investigate**: PR description does not match the implementation

The description describes the design in which the WinForms dialog is deleted and the caller's `IProgress` is threaded through, but the head code reinstates the dialog for some callers. A reviewer relying on the description would be evaluating a design the code no longer implements.

StephenMcConnel: [Claude Opus 5 (1M context) during preflight] Fair, and fixed earlier in this run — the description had indeed fallen a design behind for the second time on this branch.

It now describes the hybrid that is actually implemented, and leads with an explicit "not ready for review" warning about the inverted thread test, so nobody starts a serious review of the current state by mistake. The title was updated to match as well.

StephenMcConnel:
[Claude Opus 5 (1M context) during preflight] Consulted Devin on 2026-08-05 up to commit `7b14a6d5ab1b833ea75fa79c39505be9049da313`.

This was a re-review after a third design landed — a hybrid that reinstates the "Updating Image Files" dialog for callers with no browser progress. Devin found **one bug** and raised five things to investigate.

The bug is the important one: [the thread test that picks between the dialog and the direct call is inverted](https://github.com/BloomBooks/BloomDesktop/pull/8154#discussion_r3724637297), so the dialog is created on background threads and the silent path runs on the UI thread — the opposite of what either symptom needs. I had reached the same conclusion before Devin filed it, and a local review agent verified the WinForms semantics with an actual probe rather than from memory; that thread carries the caller-by-caller breakdown and two knock-on consequences (a failed shrink still being recorded as migrated, and `progress is WebProgressAdapter` not reliably meaning "this caller already shows progress"). Left **open** — inverting the test changes what users see on several paths, so it is the developer's call.

Also left open, both tied to that same decision: [the comments that still describe the previous design](https://github.com/BloomBooks/BloomDesktop/pull/8154#discussion_r3724642335) (I wrote two of them last run, when they were true), and [the lambda parameter that shadows the new `progress` parameter](https://github.com/BloomBooks/BloomDesktop/pull/8154#discussion_r3724644662) so the caller's progress is dead on the dialog path.

[The stale PR description](https://github.com/BloomBooks/BloomDesktop/pull/8154#discussion_r3724646426) was a fair catch and is fixed and resolved — it had fallen a design behind for the second time.

Two flags were not posted again because they duplicate existing threads. Devin's "silent shrink for `NullProgress` callers on the UI thread" is the same finding as [the still-open freeze thread](https://github.com/BloomBooks/BloomDesktop/pull/8154#discussion_r3724383011) from the previous run, which has been updated with the current status rather than forked into a second thread. And the blocking-`Invoke` deadlock flag continues to be raised against code that no longer exists anywhere on the branch; [that thread](https://github.com/BloomBooks/BloomDesktop/pull/8154#discussion_r3723872372) is already resolved and annotated to say so.

Five informational items were assessed and not mirrored. Two are worth a reviewer's eye and are carried into the report: that the `WebProgressAdapter` test is a fragile proxy for "has visible progress", and that `InvokeRequired` reads `false` for a form whose window handle is not yet created — which matters more now that `InvokeRequired` is what selects the branch. The others: the null-guard concern on the new `IProgress` parameter (verified unreachable — `Book.cs:1068` substitutes a `NullProgress`), migration messages surfacing in the import dialog (intended), and a doc-comment observation superseded by the thread above.

CI (`pr-automation`) is green, and the full C# suite passes at this commit: 3022 passed, 0 failed, 13 skipped.

src/BloomExe/Book/BookStorage.cs
StephenMcConnel: [Claude Opus 5 (1M context) during preflight] Fixed in `47ab0d5dc`. The developer chose the swap.

Off the UI thread we now hand our messages to the progress the caller already owns, on the grounds that anything arriving here from a background thread is already running under some progress of its own. On the UI thread there is nothing to borrow and inline work would freeze Bloom, so we put up our own dialog, which runs the work on a background worker.

What that changes, per the table above: book selection now gets the dialog and stays responsive instead of freezing silently; Do Updates of All Books and spreadsheet import report into the progress they already had open instead of getting a second window; the `.bloomSource` derivative import and the CLI are unaffected.

A `shell.Invoke` that briefly wrapped the dialog is gone rather than inverted along with the test — the dialog branch is now only reachable when `InvokeRequired` is false, so we are already on the shell's thread and there is nothing to marshal. Resolving.

src/BloomExe/Book/Book.cs
StephenMcConnel: [Claude Opus 5 (1M context) during preflight] Fixed in `47ab0d5dc`, and this is the thread that should close on it — the freeze you flagged here is gone.

Book selection is on the UI thread, which is now exactly the case that gets the "Updating Image Files" dialog. `ProgressDialogBackground` runs the shrink on a background worker while the UI thread pumps messages inside `ShowDialog()`, so clicking an old book with an oversized image no longer freezes Bloom and no longer shows nothing: the user gets a window with a progress bar and the per-image "Preparing image: {0}" messages.

Devin was right about this from the start, and it took three designs to land: removing the dialog also removed the background worker it was quietly providing, which is the part that was easy to miss. Resolving.

src/BloomExe/Book/BookStorage.cs
StephenMcConnel: [Claude Opus 5 (1M context) during preflight] Fixed in `47ab0d5dc`. All three now describe the design that is actually in the code.

The block comment in `MigrateToMediaLevel1ShrinkLargeImages` sets out both paths and, more usefully, *why* the thread decides which one — that a WinForms Form may only be created on the UI thread, which is the whole of BL-16646 in one sentence. The doc comment on the `progress` parameter now says plainly that it is used off the UI thread and unused on it, so nobody has to infer that from the branch.

The `BookProcessor` one I had reduced to the layout reason alone last run, on the grounds the dialog was gone. It now says the migration only creates that dialog when already on the UI thread and reports through the caller's `IProgress` otherwise — so it could not create one on BloomBridge's background thread anyway, and the layout reason genuinely is the only one left. Resolving.

src/BloomExe/Book/BookStorage.cs
StephenMcConnel: [Claude Opus 5 (1M context) during preflight] Fixed in `47ab0d5dc`. The dialog body moved into a `ShrinkImagesBehindProgressDialog` helper whose lambda parameter is named `dialogProgress`, so nothing shadows the method's `progress` any more and it is visible at a glance that the dialog's own progress is what `ImageUtils` writes to on that path.

That rename also made room for the related fix in the same commit: the lambda now wraps the work in a try/catch and the exception is rethrown on the calling thread with `ExceptionDispatchInfo`. `ProgressDialogBackground` never reads `RunWorkerCompletedEventArgs.Error`, so a failed shrink used to vanish and the book was still recorded as migrated — permanently, since the level is never revisited. Resolving.

src/BloomExe/Book/BookStorage.cs
```diff
@@ -4073,35 +4088,73 @@ public void MigrateToMediaLevel1ShrinkLargeImages()
...
+        /// on the UI thread: WinForms does not allow creating a Form anywhere else.
+        /// </summary>
+        /// <remarks>
+        /// ProgressDialogBackground never reads RunWorkerCompletedEventArgs.Error, so an exception
+        /// thrown by the work would otherwise disappear and we would carry on and record the book as
+        /// migrated when its images were not in fact shrunk -- permanently, since the level is never
+        /// revisited. So capture it and rethrow it here, which both reports the failure and leaves
+        /// mediaMaintenanceLevel alone, exactly as the direct (no-dialog) path does.
+        /// </remarks>
+        private void ShrinkImagesBehindProgressDialog()
```
StephenMcConnel: [Devin] **Investigate**: Image-shrink failures now abort book update instead of being silently swallowed

Previously an exception inside the progress dialog's worker was swallowed by BackgroundWorker and execution continued to `Dom.UpdateMetaElement("mediaMaintenanceLevel", "1")` (BookStorage.cs:4115), so a GraphicsMagick/TagLib failure left the book marked migrated. Now the exception is captured and rethrown (`ExceptionDispatchInfo` at src/BloomExe/Book/BookStorage.cs:4154-4155), so it propagates out of `EnsureUpToDate`/`BringBookUpToDate` — e.g. out of book selection on the UI thread. That is deliberate per the remarks (it keeps the level un-bumped so the migration is retried), but it converts a previously invisible failure into a user-visible error/crash path on one bad image file. Worth confirming that callers of `EnsureUpToDate` handle a throwing migration gracefully.

StephenMcConnel: [Claude Opus 5 (1M context) during preflight] Correct on the facts, and the change is intended — the developer explicitly chose "only bump `mediaMaintenanceLevel` when the work actually succeeded" from this run's decision report. But the specific thing Devin asks to confirm is worth confirming, so:

**The selection path already has a designed error path for this.** `collections/selected-book` wraps `GetUpdatedBookObjectFromBookInfo` and `SelectBook` in a try/catch (`web/controllers/CollectionApi.cs:198`) that unwraps to the interesting exception, attaches the book's folder and title to `Exception.Data`, and rethrows — which is precisely the plumbing Bloom's error reporting wants. So a bad image file surfaces as a reported error naming the book, not as an unhandled crash.

Two further reasons this is the right trade rather than a new hazard:

- **It is not new behavior for most callers.** The direct (no-dialog) path has always propagated, and that path covers every background caller plus the CLI. Before this commit the *same* failure had opposite outcomes depending only on which thread the caller happened to be on — swallowed behind the dialog, propagated everywhere else. Now both agree.
- **The alternative is worse and silent.** Swallowing left the book recorded as migrated at a level that is never revisited, so its images stayed oversized permanently with nothing in the log. A one-off `GraphicsMagick` failure being visible is recoverable; a book quietly written off is not.

Resolving. Flagging it in the run summary too, so the developer can revisit if they would rather a failed shrink be reported without aborting the rest of the book update.

StephenMcConnel:
[Claude Opus 5 (1M context) during preflight] Consulted Devin on 2026-08-05 up to commit `47ab0d5dc7309f4a589ce85ad0569ecc3ab44586`.

This is the re-review after the developer's decisions were implemented: the thread test was swapped, a failed shrink no longer records the book as migrated, the lambda parameter was renamed, and the comments were brought in line.

Devin's own output confirms the main fix landed — one of its informational items reads "The 'inverted thread test' described in the PR body is not present in this head commit." Its bug list still carries the wrong-thread bug and five Investigate flags from earlier commits on this branch; every one of those already has a thread here that was replied to and resolved this run, so they were not posted again:

- [the inverted thread test](https://github.com/BloomBooks/BloomDesktop/pull/8154#discussion_r3724637297) — fixed and resolved
- [the UI freeze on `NullProgress` callers](https://github.com/BloomBooks/BloomDesktop/pull/8154#discussion_r3724383011) — fixed and resolved; book selection is now the case that gets the dialog, and the dialog runs the work on a background worker, so it neither freezes nor goes silent
- [the stale comments](https://github.com/BloomBooks/BloomDesktop/pull/8154#discussion_r3724642335) and [the shadowed lambda parameter](https://github.com/BloomBooks/BloomDesktop/pull/8154#discussion_r3724644662) — both fixed and resolved
- the stale PR description — resolved earlier; the description has since been rewritten again for the settled design
- the blocking-`Invoke` deadlock flag — still raised against code that is not on the branch; that thread remains resolved and annotated

One genuinely **new** flag, and a fair one: [image-shrink failures now abort the book update instead of being swallowed](https://github.com/BloomBooks/BloomDesktop/pull/8154#discussion_r3724777844). That is the intended consequence of the developer's decision not to mark a book migrated when the shrink failed, and the selection path already has a try/catch that enriches and rethrows into Bloom's error reporting — so it surfaces as a reported error naming the book rather than an unhandled crash. Replied with the reasoning and resolved.

One informational item was left as accepted rather than fixed, and is written up in the PR description as a known limitation: `progress is WebProgressAdapter` is a fragile proxy for "this caller already shows progress". In practice the callers it misclassifies are all on background threads and take the caller's-progress path regardless.

CI (`pr-automation`) is green. On the full C# suite, one run reported 9 failures and two subsequent runs at the identical tree were fully green (3022 passed, 0 failed, 13 skipped); the failing run's detail was lost to an output filter on my side, so those 9 are recorded as non-reproducible rather than diagnosed. This repo has a known history of intermittent C# failures.

StephenMcConnel:
[Claude Opus 5 (1M context) during preflight] Consulted Devin on 2026-08-06 up to commit `e8ccafa0a0a6faae13c933a7116ada1c0fa259eb`.

No new bugs, and no new review threads were needed: Devin's bug and Investigate lists are entirely carried-over findings from earlier commits on this branch, and every one of them already has a thread here that was replied to and resolved. Its own informational items confirm as much — one notes that the "inverted thread test" described in an earlier PR body is not present in this head commit.

Two informational items were new and worth recording.

**The one that matters.** [On the UI thread with a real progress, the shrink now runs inline rather than on a background worker.](https://github.com/BloomBooks/BloomDesktop/pull/8154) That is a direct consequence of this commit, which stopped choosing purely on the thread and started asking whether the caller has anywhere real to report. Devin audited the callers and found the only UI-thread caller supplying a non-null progress is `CollectionModel.BringBookUpToDate` ("Update Book"), which uses `ProgressDialogForeground` and already runs the whole of `BringBookUpToDate` on the UI thread — so nothing new freezes. A local review agent reached the same conclusion independently, adding that the `MultiProgress` there includes `ApplicationDoEventsProgress`, which pumps on every message, so the dialog keeps painting. Both flag the same forward-looking caveat: a future UI-thread caller passing a live progress that does not pump would block the UI for the whole shrink. Recorded in the PR description as a note for the reviewer rather than mirrored as a thread, since it is informational and not a defect today.

**The PR description was stale again** — it still described the `progress is WebProgressAdapter` test as a known limitation, and this commit retired that test in favour of the general "is there anywhere real to report?" question. Rewritten, along with the title. The existing description-mismatch thread stays resolved rather than being reopened for the same class of problem a third time.

Worth noting what this commit fixes, since it was a regression of my own making: the swap two commits ago moved the single-book "Update Book" command onto the dialog path, so it was stacking an "Updating Image Files" dialog inside the `ProgressDialogForeground` it already had on screen and writing nothing to the foreground dialog during the shrink. The commit before the swap did not do that. Deciding on "is there anywhere real to report" rather than on the thread alone puts that caller back to reporting into its own dialog, and leaves book selection — the case this ticket is actually about — on the dialog path, where it stays responsive.

CI (`pr-automation`) is green, and the full C# suite passes at this commit: 3034 passed, 0 failed, 13 skipped.

src/BloomExe/Book/BookStorage.cs
```diff
@@ -4040,28 +4040,54 @@ public void MigrateMaintenanceLevels()
...
+        /// NullProgress (or nothing) do we put up our own dialog instead and leave this unused --
+        /// see the branch below. Passing nothing is equivalent to passing a NullProgress; in
+        /// practice only tests do, since Book.EnsureUpToDate substitutes one for a null.</param>
+        public void MigrateToMediaLevel1ShrinkLargeImages(IProgress progress = null)
         {
             var levelString = Dom.GetMetaValue("mediaMaintenanceLevel", "0");
             if (!int.TryParse(levelString, out int level))
                 level = 0;
-            if (level >= 1)
+            if (level >= 1 || _mediaLevel1ShrinkFailed)
```
StephenMcConnel, line 4072: [Devin] **Bug**: A failed image shrink is retried immediately in the same book update, redoing the slow work and showing the progress window twice

When shrinking a book's oversized images fails, the book is left marked as not-yet-shrunk and the shrink is started all over again a moment later in the same update pass, so the user waits through the whole slow operation, and the "Updating Image Files" window, a second time for the same failure.

**Impact:** opening or updating a book whose images cannot be shrunk takes twice as long and flashes the same progress window twice, with no explanation shown to the user.

<details>
<summary>Why the migration runs twice per update, and why the new failure handling made that visible</summary>

`Book.EnsureUpToDate` invokes the migration twice: once inside `EnsureUpToDateMemory` (`Book.cs:1898`) and again directly at `Book.cs:1110`. Previously the first call always set `mediaMaintenanceLevel` to 1 — even if the shrink threw, because `ProgressDialogBackground` swallowed the exception — so the second call returned immediately at the level check.

Now the helper returns false on an exception and the level is only bumped when the work succeeded. On failure the level is still 0 and `ImageUtils.NeedToShrinkImages` still returns true, so the second call repeats the entire scan and shrink and puts up the dialog again — and, since the underlying cause (e.g. a TagLib failure on one image) is deterministic, fails again with only a log entry.
</details>

src/BloomExe/Book/BookStorage.cs
```diff
@@ -4073,33 +4099,100 @@ public void MigrateToMediaLevel1ShrinkLargeImages()
...
+                    Program.RunningUnitTests
+                    || haveSomewhereToReport
+                    || shell == null
+                    || shell.InvokeRequired
+                )
                 {
-                    // TeamCity enforces not showing modal dialogs during unit tests on Windows 10.
+                    if (progress == null)
+                        progress = new NullProgress();
                     ImageUtils.FixSizeAndTransparencyOfImagesInFolder(
```
StephenMcConnel: [Devin] **Investigate**: An exception during the inline shrink aborts the whole update — and, for Do Updates of All Books, the rest of the batch — with nothing shown

The branch that shrinks synchronously with the caller's progress (this one) has no try/catch, so an exception — a TagLib failure on one image, exactly the case the dialog branch was hardened against — propagates out of `MigrateToMediaLevel1ShrinkLargeImages` and abandons the remainder of `BringBookUpToDate`: the later migrations and the `Save` that follows.

What the user sees depends on the caller:

- **Update Book** runs under `ProgressDialogForeground`, so the exception surfaces as an unhandled UI-thread exception and the user gets Bloom's problem-report dialog. Visible, if alarming.
- **Do Updates of All Books** runs inside a `ProgressDialogBackground` worker, which never inspects `RunWorkerCompletedEventArgs.Error` (`ProgressDialogBackground.cs:30`). The dialog simply closes, the remaining books in the batch are never processed, and **the user sees nothing at all**.

Devin did not file this as a bug, on the grounds that any other exception inside `BringBookUpToDate` already behaves this way. Recording it here because the asymmetry with the now-carefully-guarded dialog branch is worth a deliberate decision rather than an accident: as of this commit, the same failure on the same image is caught and logged on one branch and thrown on the other.

src/BloomExe/Book/BookStorage.cs
```diff
@@ -4073,33 +4099,100 @@ public void MigrateToMediaLevel1ShrinkLargeImages()
...
+                        }
+                    }
+                );
+            }
+            if (errorInWorker != null)
+            {
+                // Log the whole exception, not just its Message: the likeliest failure is a
+                // TagLib error on one particular image, and the message alone often does not
+                // say which file, nor carry the inner exception.
+                Logger.WriteError("Shrinking images failed in " + FolderPath, errorInWorker);
```
StephenMcConnel: [Devin] **Investigate**: Failure handling in the dialog branch differs from the PR description

The PR description said a failure in the dialog's background worker is captured and **rethrown** on the calling thread via `ExceptionDispatchInfo`. That is no longer what the code does: the exception is caught, logged, and reported back as a `false` return, and the caller carries on with the rest of the update.

src/BloomExe/Book/BookStorage.cs
```diff
@@ -4073,33 +4099,100 @@ public void MigrateToMediaLevel1ShrinkLargeImages()
...
+                // Note the invariant this puts on such a caller: taking the caller's progress also
+                // means doing the shrinking synchronously on the caller's thread, so a caller that
+                // is on the UI thread needs a progress that pumps messages, or Bloom will be frozen
+                // for the whole (potentially minutes-long) shrink. Today the only UI-thread caller
+                // with a real progress is CollectionModel.BringBookUpToDate ("Update Book"), which
+                // is safe on both counts: ProgressDialogForeground runs all of BringBookUpToDate on
+                // the UI thread anyway, and its MultiProgress includes an ApplicationDoEventsProgress
+                // that pumps on every message. A future UI-thread caller passing a progress that does
+                // not pump would need the dialog branch below instead.
+                var haveSomewhereToReport = progress != null && !(progress is NullProgress);
```
StephenMcConnel, line 4116: [Devin] **Investigate**: Silent shrink for NullProgress callers on the UI thread

Devin's reading: a caller on the UI thread that passes `NullProgress` — book selection, `PublishModel`, `BookCommandsApi`, `bloom hydrate` — takes the direct-call branch and shows the user nothing while a potentially minutes-long shrink blocks the UI thread; so the reinstated dialog never covers the case it was meant for.

src/BloomExe/Book/BookStorage.cs
StephenMcConnel: [Claude Opus 5 (1M context) during preflight] Confirmed and fixed in `f81e1b815`.

I checked the doubled call directly rather than taking it on trust: `EnsureUpToDate` calls `EnsureUpToDateMemory(progress)` at `Book.cs:1107`, which reaches the migration at `Book.cs:1898`, and then calls the migration again itself at `Book.cs:1110`. So the analysis is exactly right — the second call was a no-op only because the first had already bumped the level, and the new failure handling removed that.

The fix remembers the failed attempt in a field on the `BookStorage`, checked alongside the level check, so one pass makes one attempt. `mediaMaintenanceLevel` still stays at 0, so the shrink is still retried the next time that book is loaded and brought up to date — which is the behavior the previous commit was after. Retrying is cheap for the images that did succeed: `FixSizeAndTransparencyOfImagesInFolder` only resizes a file whose current dimensions differ from the desired ones, so already-shrunk images are skipped rather than re-encoded.

Good catch — this was a real user-visible consequence, and it was introduced by the commit immediately before yours.

src/BloomExe/Book/BookStorage.cs
StephenMcConnel: [Claude Opus 5 (1M context) during preflight] Correct, and fixed — the PR description has been rewritten to match the code.

It now says the failure is caught, logged in full via `Logger.WriteError` (so the stack trace and inner exception survive, which `.Message` alone dropped), and reported back as `false`, with `mediaMaintenanceLevel` bumped only on success. It also records the two consequences that follow: the retry on the next update, and the fact that the direct branch still throws — that last one is open for the developer to decide, on its own thread.

src/BloomExe/Book/BookStorage.cs
StephenMcConnel: [Claude Opus 5 (1M context) during preflight] This is not right about the code as it now stands — it has the branches the other way round.

The test is `Program.RunningUnitTests || haveSomewhereToReport || shell == null || shell.InvokeRequired`. For a UI-thread caller holding only a `NullProgress`: `haveSomewhereToReport` is false (a `NullProgress` reports nowhere, so it counts as having none), `shell` is not null, and `InvokeRequired` is false because we are on the shell's own thread. Every disjunct is false, so control reaches the `else` — the dialog branch. Book selection therefore *does* get the "Updating Image Files" dialog, and the work runs on that dialog's background worker rather than blocking the UI thread.

The flag was accurate about an earlier design on this branch, which had the thread test the other way round; it was fixed before this commit. Resolving on that basis.

StephenMcConnel:
[Claude Opus 5 (1M context) during preflight] Consulted Devin on 2026-08-06 up to commit `f81e1b815e1c7cdeaa0f93926b1c0843808e6217`.

This was the re-review after the developer changed a failed shrink from rethrowing to logging and carrying on. Devin earned its keep this round: it found **one real bug** in that change and it is now fixed.

- **A failed shrink was retried immediately, in the same update pass** — [thread](https://github.com/BloomBooks/BloomDesktop/pull/8154#discussion_r3731074210). `Book.EnsureUpToDate` calls this migration twice, and the second call was harmless only while the first always bumped `mediaMaintenanceLevel`. Once a failure left the level at 0, the second call redid the whole slow shrink and put the dialog up a second time before failing again. Fixed in `f81e1b815`; Devin's re-review marks it resolved.
- **The direct branch still throws where the dialog branch now logs** — [thread](https://github.com/BloomBooks/BloomDesktop/pull/8154#discussion_r3731075877), left **open** for the developer. It is the one thing on this PR still awaiting a human decision.
- **The PR description no longer matched the code** — [thread](https://github.com/BloomBooks/BloomDesktop/pull/8154#discussion_r3731077428). Correct; the description has been rewritten.
- **"Silent shrink for `NullProgress` callers on the UI thread"** — [thread](https://github.com/BloomBooks/BloomDesktop/pull/8154#discussion_r3731077523). Not right about the current code; it has the branches the other way round. Explained and resolved.

Its remaining four Investigate flags and both listed bugs are carried-over findings from earlier commits on this branch — every one already has a thread here that was replied to and resolved, and Devin now marks both bugs resolved. Eleven further items were informational; none needed action.

The full C# suite is green at this commit: 3040 passed, 0 failed, 13 skipped.

src/BloomExe/Book/BookStorage.cs
```diff
@@ -4073,33 +4099,136 @@ public void MigrateToMediaLevel1ShrinkLargeImages()
...
+                    {
+                        ImageUtils.FixSizeAndTransparencyOfImagesInFolder(
+                            FolderPath,
+                            new List<string>(),
+                            progress
+                        );
+                    }
+                    catch (Exception e)
+                    {
+                        ReportShrinkFailure(e, progress);
```
StephenMcConnel, line 4141: [Devin] **Investigate**: The two failure branches still behave differently, contrary to what the PR notes say

The PR description's "Notes for the reviewer" said the direct (non-dialog) branch has no try/catch and lets exceptions propagate out of `BringBookUpToDate`. The code now catches there, reports passively, and continues with the remaining migrations and the `Save`.

That is arguably the better behavior, but it is a real semantic change for callers such as Do Updates of All Books and the `.bloomSource` / spreadsheet imports, which would previously have aborted the update of that book; now the book is saved and the failure appears as a warning line plus a toast. Worth confirming the description is simply stale rather than the code being unintended.

StephenMcConnel: [Claude Opus 5 (1M context) during preflight] Right on both counts: the description was stale, and the code change is intended. The description has been rewritten.

The catch on the direct branch is the deliberate resolution of the open question from the previous round — the two branches genuinely did diverge then, one catching and one throwing, and the developer chose to make them agree by catching in both. So the branches now behave the same, and the description says so.

You have named the consequence correctly, and it is the point rather than a side effect: a bad image now costs that one book's shrink instead of the rest of the operation. The case that motivated it is Do Updates of All Books, which runs on a `ProgressDialogBackground` worker whose `RunWorkerCompleted` never inspects `e.Error` — so before this, one unreadable image ended the whole batch with the dialog simply closing and nothing shown. The remaining books are now processed.

On the failure being "only" a warning plus a toast: that is deliberate too, and it is more than it sounds. `NonFatalProblem` also writes the entire exception to the log and reports to Sentry. The user-facing part is passive because the book still works — its pictures are just left large — and `mediaMaintenanceLevel` stays at 0, so the shrink is retried next time the book is brought up to date. The warning goes to the caller's progress via `WriteWarning` rather than `WriteError` on purpose: `ProgressDialogForeground` shows a modal "There was a problem performing that operation" whenever its progress records an error, which would undo the passive choice.

Resolving, since the only actual defect here — the stale description — is fixed.

src/BloomExe/Book/BookStorage.cs
StephenMcConnel: [Claude Opus 5 (1M context) during preflight] Decided and fixed: the direct branch now catches too, so the two branches agree.

The developer took the option this thread recommended. The failure is caught on both branches and reported the same way — passively, via `NonFatalProblem` — so one bad image costs that book's shrink rather than the rest of the update. For Do Updates of All Books, which was the case that mattered here, the remaining books in the batch are now processed instead of the run ending silently when `ProgressDialogBackground` closed without ever reading `e.Error`.

Two things were added on top while implementing it: the failure is also written to the caller's progress (as a *warning*, not an error — `ProgressDialogForeground` shows a modal problem dialog whenever its progress records an error, which would have defeated reporting this passively), and the user-visible text is localized rather than an interpolated string with the book's folder path in it.

Resolving.

StephenMcConnel:
[Claude Opus 5 (1M context) during preflight] Consulted Devin on 2026-08-06 up to commit `4e0324264af2b388e00b6035511294392e0ae54b`.

This was the re-review after the developer settled the open question from the last round: a failed shrink is now caught on **both** branches and reported passively, so a bad image costs that one book's shrink rather than the rest of the update.

- **One new flag, and it was correct** — [thread](https://github.com/BloomBooks/BloomDesktop/pull/8154#discussion_r3731500948): the PR description still described the old behavior, where the direct branch threw. Devin also asked whether the semantic change was intended. It is, and it is the point rather than a side effect — before this, one unreadable image ended a whole Do-Updates-of-All-Books run with nothing shown. Description rewritten; thread replied to and resolved.
- **The previously open thread is now closed too** — [thread](https://github.com/BloomBooks/BloomDesktop/pull/8154#discussion_r3731075877). It asked for exactly this fix, the developer chose it, and the outcome is recorded there.

Its other seven Investigate flags and both listed bugs are carried-over findings from earlier commits on this branch — each already has a thread here that was replied to and resolved, and Devin marks both bugs resolved. Twelve further items were informational; none needed action.

**There are now no open review threads on this PR.**

The full C# suite is green at this commit: 3031 passed, 0 failed, 13 skipped. That count is lower than the 3040 reported for the previous commit because master has since made `dotnet test` exclude the Integration and Nightly categories by default; it matches the figure in that commit's own message.

JohnThomson (approved):
> @JohnThomson reviewed 7 files and all commit messages, and resolved 13 discussions.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8154)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on StephenMcConnel).

JohnThomson (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8154)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on StephenMcConnel).

PR #8155, opened 2026-08-05
Make reporting a blocked server worker always top up the pool (BL-16612)

andrew-polk:
Fixes the BL-16612 deadlock by repairing BloomServer's blocked-worker bookkeeping, and does nothing else.

## The problem

BloomServer counts workers that report themselves blocked and adds one when they are all blocked, so that somebody is still able to take the request which would let the others finish — that request is often itself sitting in the queue.

Stacks captured at the moment of failure on CI showed four workers and **none idle**: three parked on the api lock a publish held for minutes, and the publish itself blocked in the off-screen browser waiting for a page only a worker could serve. The safeguard evaluated `3 >= 4`, concluded a worker was free, and publishing waited forever — taking every later publish in the session with it, until Bloom was restarted.

## Four things were wrong with that mechanism

1. **The check ran only as a request ARRIVED** (in `QueueRequest`). That cannot be sufficient on its own: when the request that would break the deadlock is already queued, no new one need ever arrive. It now also runs at the other moment the pool can run out — when a worker becomes blocked — because the check moved *into* the reporting method itself. `QueueRequest` calls the same helper, so the rule lives in exactly one place.

2. **`OffScreenBrowser.RunAndBlock` never reported its block at all**, though publishing runs as an api request and blocks a worker there for as long as the browser takes. That is why the count read 3 rather than 4.

3. **A reported block could leak permanently.** Both halves decided whether to count by inspecting the *current thread's name*, so the pair only balanced when both ran on the same worker. In `ApiRequest.InvokeWithErrorHandling` they do not: the second half runs after an `await`, and a worker carries no synchronization context, so the continuation resumes on a thread-pool thread and the decrement was silently skipped — permanently inflating the count, which (given fix 1) would then grow the pool on every later block. The same method leaked again whenever anything other than `ObjectDisposedException` came out of `Invoke`, since that path skipped both halves.

   Reporting a block now returns an `IDisposable` scope that closes over whether it counted. Disposing it undoes exactly that block, from any thread, on every exit path. Disposal is idempotent, so a nested `using` cannot decrement twice and leave the server believing workers are free when they are not. `RegisterThreadUnblocked` is **deleted** rather than kept alongside — a second way to end a block is the same trap as a second way to report one. `IBloomServer` changes with it, so **BloomHarvester needs the same one-line update if it calls these.**

4. **The shortage was judged against workers that may already be dead.** `_workers` keeps an entry for a thread that has died until the listener prunes it, so the count could claim a worker was available when nothing of it remained but the entry. Both the fast path and the under-lock check now count live workers — it has to be both, because the fast path exists only to avoid taking the queue lock, so while it over-counted it returned early and made the accurate check unreachable.

## Where to look, and what to be sceptical about

- **The hot path.** The check now runs on every api request that waits for a lock, and it walks the workers (four to ten, one `IsAlive` each) rather than reading a count. Still far cheaper than joining the queue behind the listener and the workers, which is the only thing that path avoids — but it is not free, and that was a deliberate trade.
- **It must not throw.** A caller increments the count and only receives the scope that undoes it after `EnsureAWorkerCanStillTakeWork` returns, so anything escaping there would leak the count for the life of the process. Hence the catch-all, including around its own logging — a documented, deliberate exception to `AGENTS.md`'s fail-fast guidance, with the reasoning in a `REVIEWED DECISION` comment.
- **Detection still happens at report time.** Whether a block counts depends on the reporting thread being a worker. Verified for the publish path both ways: the two `updatePreview` endpoints register *synchronous* handlers, so nothing yields between the worker entering the handler and the browser call; and the captured stacks show a `Server Worker Thread` inside `RunAndBlock`. It is not structurally guaranteed for a future caller, though — see the review threads.
- **Lock safety.** `EnsureAWorkerCanStillTakeWork` takes `lock (_queue)`. Request handlers run outside that lock, and no BloomServer-internal code reports a block, so it cannot self-deadlock; in `QueueRequest` the lock is already held, which is fine since `Monitor` is reentrant. Devin checked for a lock-order inversion independently and found none.

## What this deliberately does NOT do

No time limit on how long publishing waits for the browser, no retry, and no refusal to publish. Those were considered at length and dropped: the starvation is fixed at its source, and a bound sized anywhere near real working time would convert a correct-but-slow book into a silently wrong one. Apart from no longer deadlocking, **there is no user-visible change** — no new message, no new failure mode.

Consequently the long-standing "silently degraded BloomPUB" path (page checks yielding no data ⇒ nothing stripped, no fonts embedded, success reported) is **untouched**, and still needs its own card.

## Tests

Five, in `BloomServerWorkerPoolTests`, driving a real listening server. Two were confirmed to fail without their fix rather than passing for the wrong reason:

- a worker is added when the last free one blocks — and *not* while one is still free;
- **disposing the scope on another thread still undoes the block** (fails if the thread-identity check is restored);
- disposing twice only undoes it once;
- **a dead worker's entry does not suppress the top-up** (fails if the live count is applied in only one of the two places — which is how that gap was found);
- blocks reported by non-worker threads are ignored entirely.

Full C# suite green apart from `RabRealBuildTests`, which needs Reading App Builder's `buildapp.bat` on the PATH and fails identically on unmodified `master`.

## Known gaps, recorded rather than fixed

In code comments and on the review threads, so they are not rediscovered as bugs: a worker added late during shutdown is only harmless while `Dispose` signals `_stop` without disposing the wait handles — and reaching that signal is itself conditional on `_listener != null`, which `CloseListener()` can have already cleared; and workers are never retired, so the pool can end a session larger than it started.

---
[Devin review](https://devinreview.com/BloomBooks/BloomDesktop/pull/8155)

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8155)

### DIFF
--- a/src/BloomExe/Publish/OffScreenBrowser.cs
+++ b/src/BloomExe/Publish/OffScreenBrowser.cs
@@ -286,7 +286,25 @@ namespace Bloom.Publish
                 },
                 null
             );
-            return tcs.Task.GetAwaiter().GetResult();
+
+            // We are about to block, and publishing runs as an api request, so the thread we block is
+            // usually a BloomServer worker. Tell the server, so that it keeps a worker free to serve the
+            // requests this browser is about to make -- including the very page we navigate to, which
+            // BloomServer serves from memory.
+            //
+            // Without this the server could not tell that its whole pool was blocked. Reporting a block is
+            // opt-in, and neither of the two existing reports covers us. BloomApiHandler's covers ACQUIRING
+            // the api lock, not the long stretch while publishing holds it. ApiRequest's covers marshalling
+            // a handler onto the UI thread, which the publish endpoints deliberately opt out of (they
+            // register with handleOnUiThread: false), so it never runs on this path -- and that opt-out is
+            // also why the thread we block here is a server worker rather than the UI thread, which is what
+            // makes reporting it worth doing at all. So every other worker could be waiting for that lock,
+            // and this one waiting here, with the server believing a worker was still free. That is
+            // BL-16612.
+            using (BloomServer._theOneInstance?.ReportThreadBlocking())
+            {
+                return tcs.Task.GetAwaiter().GetResult();
+            }
         }
 
         /// <summary>
--- a/src/BloomExe/web/BloomServer.cs
+++ b/src/BloomExe/web/BloomServer.cs
@@ -44,8 +44,11 @@ namespace Bloom.Api
     // when it doesn't want to spin up a real one.
     public interface IBloomServer
     {
-        void RegisterThreadBlocking();
-        void RegisterThreadUnblocked();
+        /// <summary>
+        /// See BloomServer.ReportThreadBlocking. Dispose the result when the blocking work is done;
+        /// there is deliberately no separate "unblocked" call to forget or to get wrong.
+        /// </summary>
+        IDisposable ReportThreadBlocking();
 
         // ENHANCE: Add other methods as needed
     }
@@ -136,6 +139,15 @@ namespace Bloom.Api
         /// Pool of threads that pull a request from the _queue and processes it.
         /// This is a ConcurrentDictionary (ManagedThreadId to thread) just so we can add and remove
         /// things from it without worrying about locking (or deadlocking).
+        ///
+        /// Two properties of this collection that other code relies on, so take care before changing
+        /// either. Additions are made only under lock (_queue) (see SpinUpAWorker), which is what lets a
+        /// caller re-check a count and act on it without another thread adding a worker underneath it. And
+        /// an entry is only ever REMOVED for a thread that has already died (see the pruning in
+        /// EnqueueIncomingRequests) -- nothing removes a live worker. That second property is what makes it
+        /// safe for EnsureAWorkerCanStillTakeWork to count live workers WITHOUT the lock: a concurrent
+        /// removal can only take away something that was not going to be counted as live anyway, so it
+        /// cannot inflate the answer.
         /// </summary>
         private readonly ConcurrentDictionary<int, Thread> _workers = new();
 
@@ -151,7 +163,7 @@ namespace Bloom.Api
 
         /// <summary>
         /// Keeps track of the number of worker threads that are blocked
-        /// Note: This is NOT automatically computed. Other code should call RegisterThreadAboutToBlock() and RegisterThreadUnblocked()
+        /// Note: This is NOT automatically computed. Other code should call ReportThreadBlocking() and dispose the scope it returns
         ///        whenever it causes a thread which is or potentially is a server worker thread to block.
         /// Note: This is different than _busyThreads, because a thread may be busy but not blocked.
         /// </summary>
@@ -1928,14 +1940,10 @@ namespace Bloom.Api
                 // Deal with a situation where all the workers are blocked,
                 // but there is a request in the queue that would unblock the current workers
                 // but that request can't run because it's stuck in queue
-                // and none of the existing worker threads are able to make progress anymore
-                if (_countBlockedThreads >= _workers.Count)
-                {
-                    // The worker should be spun up such that it can receive _ready.Set()
-                    SpinUpAWorker();
-
-                    // Note: Currently these workers are never stopped, so as not to complicate the code any further
-                }
+                // and none of the existing worker threads are able to make progress anymore.
+                // Any worker added here is added before the _ready.Set() below, so it receives it.
+                // (Monitor is reentrant, so it is fine that we already hold this lock.)
+                EnsureAWorkerCanStillTakeWork();
 
                 _ready.Set();
             }
@@ -2397,44 +2405,222 @@ namespace Bloom.Api
         }
 
         /// <summary>
-        /// Registers that the current thread is about to block.
-        /// This function should be called immediately before any server thread blocks
-        /// (e.g. waits for a lock, wait for a modal dialog to close, etc.)
-        /// Must be paired with RegisterThreadUnblocked() when done.
+        /// Reports that the calling thread is about to block -- waiting for a lock, for a modal dialog to
+        /// close, for an off-screen browser, and so on. Call it immediately before the blocking work and
+        /// dispose the result as soon as that work is done, normally with a `using` block.
         ///
-        /// This can be called by any code that at least sometimes (if not always)
-        /// is called by a BloomServer worker thread. The caller need not guarantee that
-        /// the current thread is a server thread. This method will check for that.
+        /// Any code that is *sometimes* run by a server worker may call this; it need not know whether the
+        /// current thread is one. Blocks reported by other threads are ignored, since they are not using up
+        /// a worker.
+        ///
+        /// Why this returns a scope instead of having a matching "unblocked" method: whether a block counts
+        /// depends on whether the caller is one of our workers, and that answer has to be REMEMBERED rather
+        /// than worked out again at the end. Blocking work that contains an await can resume on a different
+        /// thread -- a worker carries no synchronization context, so continuations land on the thread pool --
+        /// and asking "am I a worker?" there would answer no and silently skip the decrement, inflating the
+        /// count for the life of the process. The scope closes over the answer, so it cannot drift, and it
+        /// releases on every exit path including an exception. Both bugs were real: see BL-16612.
         /// </summary>
-        public void RegisterThreadBlocking()
+        public IDisposable ReportThreadBlocking()
         {
-            // Check if the current thread looks like a Server Worker
-            // If not, we can just ignore this request.
-            // Notably, ProblemReportApi can be invoked by both server and non-server code
-            if (IsWorkerThread(Thread.CurrentThread))
+            // Notably, ProblemReportApi can be invoked by both server and non-server code.
+            if (!IsWorkerThread(Thread.CurrentThread))
+                return NotBlockingAWorker;
+
+            Interlocked.Increment(ref _countBlockedThreads);
+            // Must not throw; if it did, the caller would never receive the scope that undoes the
+            // increment above. See the guarantee inside it.
+            EnsureAWorkerCanStillTakeWork();
+            return new BlockedWorkerScope(this);
+        }
+
+        // Shared, stateless scope handed to callers whose thread is not one of our workers, so those
+        // callers still get something disposable and need no special case.
+        private static readonly IDisposable NotBlockingAWorker = new DoNothingScope();
+
+        private sealed class DoNothingScope : IDisposable
+        {
+            public void Dispose() { }
+        }
+
+        /// <summary>
+        /// Undoes exactly one counted block, once. Holding the server (rather than re-deriving anything
+        /// from the current thread) is the whole point -- see ReportThreadBlocking.
+        /// </summary>
+        private sealed class BlockedWorkerScope : IDisposable
+        {
+            private BloomServer _server;
+
+            public BlockedWorkerScope(BloomServer server)
             {
-                // Note: So far only BloomApiHandler and problem report dialog have been analyzed to call this when needed.
-                Interlocked.Increment(ref _countBlockedThreads);
+                _server = server;
+            }
+
+            public void Dispose()
+            {
+                // Taking the reference away atomically makes a second Dispose -- e.g. a `using` inside a
+                // method whose caller also disposes -- harmless instead of double-decrementing.
+                var server = Interlocked.Exchange(ref _server, null);
+                if (server != null)
+                    Interlocked.Decrement(ref server._countBlockedThreads);
             }
         }
 
         /// <summary>
-        /// Registers that the current thread is no longer blocked.
-        /// Should be called as a pair with RegisterThreadBlocking(), after any blocking work returns.
+        /// If every worker is now blocked, add one, so that some worker is still able to take a request off
+        /// the queue. This matters because the request that would let the blocked workers finish is often
+        /// itself sitting in the queue: in BL-16612 a publish held an api lock while every other worker
+        /// waited for that same lock, so no worker was left to serve the page the publish was waiting for,
+        /// and the whole server deadlocked.
+        ///
+        /// This is called from ReportThreadBlocking rather than being offered as a separate method for
+        /// callers to remember, so that every caller which correctly reports that it is blocking gets the
+        /// top-up automatically. There is deliberately no way to register a block WITHOUT it -- reporting
+        /// the block is the only thing a caller has to get right.
+        ///
+        /// QueueRequest makes the same check as a request ARRIVES, which is not sufficient by itself: if
+        /// the request that would break the deadlock is already in the queue, no new request need ever
+        /// arrive to trigger the check. Checking here covers the other moment the pool can run out -- when
+        /// a worker becomes blocked.
+        ///
+        /// Deliberately has no shutdown guard, unlike QueueRequest, which gives up once the listener has
+        /// closed. A worker can therefore add one more worker while Dispose is joining threads. That is safe
+        /// only because Dispose signals _stop without disposing it or _ready, so the new worker wakes
+        /// immediately and exits -- see the note in Dispose, which must stay true for this to remain safe.
         /// </summary>
-        public void RegisterThreadUnblocked()
+        private void EnsureAWorkerCanStillTakeWork()
         {
-            // Check if the current thread looks like a Server Worker
-            // If not, we can just ignore this request.
-            // Notably, ProblemReportApi can be invoked by both server and non-server code
-            if (IsWorkerThread(Thread.CurrentThread))
+            // NOTHING in this method may throw, logging and the fast-path reads alike. The caller increments
+            // the blocked count and only receives the scope that undoes it AFTER this returns, so an
+            // exception escaping here would leak that count for the life of the process -- which would then
+            // make every later block add yet another worker. Adding a worker is a safety net; neither
+            // failing to add one nor failing to log it may turn into a failed request. The fast path is
+            // inside the try so that guarantee is structural, rather than resting on an argument about what
+            // ConcurrentDictionary.Count can do.
+            //
+            // REVIEWED DECISION (BL-16612): catching everything here, including around the logging in the
+            // catch below, is a deliberate exception to this repo's "fail fast, don't be defensive"
+            // guidance in AGENTS.md. Do not "clean this up" into a narrower catch without reading the rest
+            // of this comment, because the obvious objection to it has already been raised and answered.
+            //
+            // The immediate reason: the caller increments the blocked count and only receives the scope
+            // that undoes it after this returns, so an exception escaping here would leave that count
+            // permanently high -- the very condition this method exists to relieve.
+            //
+            // That is not the only way to arrange things, and a cheaper alternative than we first thought
+            // does exist: hand the caller its scope BEFORE calling this, dispose it and rethrow if this
+            // throws, and then this method would be free to fail fast. That touches only
+            // ReportThreadBlocking, not every call site. It was considered and declined, and the reason is
+            // a judgement rather than a constraint: topping up the pool is an opportunistic safety net, and
+            // a safety net failing is not itself a reason to fail the user's request that happened to
+            // trigger it. Publishing a book should not die because the server could not create a spare
+            // thread it may well not need. If you disagree with that trade, the restructure above is the
+            // way to change it -- but change it deliberately, not by narrowing this catch and reintroducing
+            // the leak.
+            try
             {
-                Interlocked.Decrement(ref _countBlockedThreads);
+                // Fast path that avoids _queue's lock, which the listener (enqueuing) and the workers
+                // (dequeuing) are already contending for; this runs on every api request that waits for a
+                // lock. Walking the workers to count the live ones is not free, but it is far cheaper than
+                // joining the queue behind those two.
+                //
+                // Why this counts LIVE workers even out here, where a cheap approximation would normally be
+                // fine: the only thing this test can do is SKIP the more careful check below, so the two
+                // directions of error are not symmetric.
+                //   Too LOW (say we race a SpinUpAWorker that has not added its thread yet): we decline to
+                //     return, take the lock, and get the better answer. Self-correcting -- and the reason
+                //     reading this unsynchronized is acceptable at all.
+                //   Too HIGH: we return here and the check below never runs. The raw entry count is
+                //     SYSTEMATICALLY too high, because a dead thread keeps its entry until the listener
+                //     prunes it, so using it here left the careful check unreachable.
+                // A concurrent removal cannot push us into that dangerous direction, because entries are
+                // only removed for threads that are already dead; see the note on _workers.
+                //
+                // Be clear about what a live count does NOT buy, though. It is a fact about the instant it
+                // was taken, and a worker can die immediately afterwards. The re-check under the lock is no
+                // better in that respect: the lock covers changes to _workers, not thread liveness. So
+                // neither reading is authoritative about how many workers are still alive by the time we
+                // act on it. What counting live workers removes is the systematic over-count from lingering
+                // dead entries -- not that race.
+                //
+                // Staleness in the blocked count is harmless: Interlocked.Increment gives the increments a
+                // total order, so whichever thread performs the last one reads a count including every
+                // earlier block. The worker that exhausts the pool therefore always sees the shortage, even
+                // if the ones before it did not.
+                if (Volatile.Read(ref _countBlockedThreads) < LiveWorkerCount())
+                    return;
+
+                var addedWorker = false;
+                // SpinUpAWorker requires this lock, since it modifies _workers. Re-checking here means we
+                // decide against a count taken after any concurrent add, rather than the one above -- but
+                // per the note above, not against a count guaranteed still true when we act on it.
+                lock (_queue)
+                {
+                    if (_countBlockedThreads >= LiveWorkerCount())
+                    {
+                        // REVIEWED DECISION (BL-16612): workers are never retired -- that predates this
+                        // code -- and we accepted that the pool can therefore end a session larger than it
+                        // started. During a long publish, several requests can queue behind the same lock
+                        // and each one can add a worker, so growth is bounded by how many are actually
+                        // waiting, not unbounded. We judged that a fair price for not deadlocking, but it
+                        // is why thread counts in a diagnostic may look higher than you expect. Retiring
+                        // idle workers would be the real fix and is a much larger change.
+                        SpinUpAWorker();
+                        addedWorker = true;
+                    }
+                }
+                // Logged after releasing our own lock (QueueRequest's caller may still hold it), since
+                // logging can be slow. This is the event to look for in a log when investigating a freeze,
+                // so it is worth a line even though it is not an error.
+                if (addedWorker)
+                    Logger.WriteEvent(
+                        "BloomServer: every worker was blocked, so added one to keep requests moving "
+                            + $"({_workers.Count} workers, {_countBlockedThreads} blocked)."
+                    );
+            }
+            catch (Exception e)
+            {
+                // This last attempt to record what went wrong could itself fail, and nothing may escape
+                // (see above), so it is deliberately allowed to fail silently.
+                try
+                {
+                    Logger.WriteEvent(
+                        "BloomServer: trouble adding a worker while one was blocking: " + e.Message
+                    );
+                }
+                catch { }
             }
         }
 
+        /// <summary>
+        /// How many workers are actually alive and so could still take a request off the queue. Bloom has
+        /// seen worker threads die (see the pruning in EnqueueIncomingRequests), and a dead one leaves its
+        /// entry in _workers until that pruning runs, so _workers.Count can overstate the pool.
+        ///
+        /// Counts the dictionary itself rather than its Values, which on a ConcurrentDictionary materialises
+        /// a snapshot list -- worth avoiding on something EnsureAWorkerCanStillTakeWork calls for every api
+        /// request that waits for a lock. Callers may hold lock (_queue) or not; see that method for why
+        /// counting without it is safe.
+        /// </summary>
+        private int LiveWorkerCount() => _workers.Count(kvp => kvp.Value?.IsAlive == true);
+
+        /// <summary>
+        /// The number of worker threads in the pool, alive or not. For tests, which need to observe that a
+        /// worker was added when they made every existing worker report itself blocked.
+        /// </summary>
+        internal int WorkerCount => _workers.Count;
+
+        /// <summary>
+        /// How many workers currently report themselves blocked. For tests, which need to prove that a
+        /// reported block is undone even when the scope is disposed on a different thread than reported it.
+        /// </summary>
+        internal int BlockedWorkerCount => _countBlockedThreads;
+
+        // Ordinal deliberately: the default overloads of both StartsWith and IndexOf(string) are
+        // culture-sensitive, which is the wrong kind of comparison for a thread name we generated
+        // ourselves -- and slower.
         private bool IsWorkerThread(Thread thread) =>
-            thread?.Name?.IndexOf(WorkerThreadNamePrefix) == 0;
+            thread?.Name?.StartsWith(WorkerThreadNamePrefix, StringComparison.Ordinal) == true;
 
         private string GetHtmlForRootOfBloomUI()
         {
@@ -2551,6 +2737,23 @@ namespace Bloom.Api
                         // tell _listenerThread and the worker threads they should stop
                         _stop.Set();
 
+                        // Note (BL-16612): a worker part-way through a request can still report a block while
+                        // we are joining threads below, and EnsureAWorkerCanStillTakeWork has no shutdown
+                        // guard, so that can add a worker after the join has passed it. Two things make that
+                        // survivable, and BOTH have to stay true:
+                        //   1. We only SIGNAL _stop here and never dispose it or _ready, so a late worker's
+                        //      WaitAny returns at once and it exits. Dispose those handles and it would die
+                        //      instead of an ObjectDisposedException on a background thread, taking the
+                        //      process with it.
+                        //   2. We actually reach this line. Note that it is inside `if (_listener != null)`,
+                        //      so if CloseListener() ran first -- it is public, nulls _listener, and its own
+                        //      comment mentions the shutdown timer -- _stop is never signalled at all.
+                        // Caveat worth knowing about (2): workers are FOREGROUND threads, so a worker still
+                        // waiting on _ready/_stop keeps the process alive. Pre-existing workers have always
+                        // been exposed to that; what the top-up adds is the chance of creating a NEW one
+                        // during that window. Devin raised it; it is recorded rather than fixed here because
+                        // the fix (guard the top-up on shutdown, or make added workers background threads)
+                        // changes behaviour and was left for the developer to decide.
                         var secondsToWait = 2.0;
                         // wait for _listenerThread to stop
                         if (_listenerThread.ThreadState != ThreadState.Unstarted)
--- a/src/BloomTests/web/BloomServerWorkerPoolTests.cs
+++ b/src/BloomTests/web/BloomServerWorkerPoolTests.cs
@@ -1,0 +1,291 @@
+using System;
+using System.Collections.Concurrent;
+using System.Collections.Generic;
+using System.Reflection;
+using System.Runtime.ExceptionServices;
+using System.Threading;
+using System.Threading.Tasks;
+using Bloom.Api;
+using Bloom.Book;
+using NUnit.Framework;
+
+namespace BloomTests.web
+{
+    /// <summary>
+    /// Tests for BloomServer's worker pool keeping a worker able to take work while others are blocked.
+    ///
+    /// Deliberately a separate fixture from BloomServerTests, which needs a collection, a localization
+    /// manager, and a BloomFileLocator; none of that is relevant here, and a real listening server is
+    /// (unlike in those tests) exactly what we need, since the workers only exist once it is listening.
+    /// </summary>
+    [TestFixture]
+    public class BloomServerWorkerPoolTests
+    {
+        /// <summary>
+        /// The whole point of reporting a block (BL-16612): once every worker is blocked, no worker is left
+        /// to take a request off the queue -- including the request that would let the blocked workers
+        /// finish. So the server has to add one.
+        /// </summary>
+        [Test]
+        public void ReportThreadBlocking_WhenTheLastFreeWorkerBlocks_AddsAWorker()
+        {
+            using (var server = new BloomServer(new BookSelection()))
+            {
+                server.EnsureListening();
+                var workersAtStart = server.WorkerCount;
+                Assert.That(
+                    workersAtStart,
+                    Is.GreaterThan(1),
+                    "SANITY: the pool should start out with several workers"
+                );
+
+                // The server only counts blocks reported by its own worker threads, so we have to report
+                // from a thread that looks like one.
+                RunOnAPretendWorkerThread(() =>
+                {
+                    var scopes = new List<IDisposable>();
+                    try
+                    {
+                        // One short of all of them: a worker is still free, so leave the pool alone.
+                        for (var i = 0; i < workersAtStart - 1; i++)
+                            scopes.Add(server.ReportThreadBlocking());
+                        Assert.That(
+                            server.WorkerCount,
+                            Is.EqualTo(workersAtStart),
+                            "should not have added a worker while one was still free"
+                        );
+
+                        // Now the last free worker blocks too.
+                        scopes.Add(server.ReportThreadBlocking());
+                        Assert.That(
+                            server.WorkerCount,
+                            Is.EqualTo(workersAtStart + 1),
+                            "should have added a worker once every worker was blocked"
+                        );
+                    }
+                    finally
+                    {
+                        foreach (var scope in scopes)
+                            scope.Dispose();
+                    }
+                });
+            }
+        }
+
+        /// <summary>
+        /// Blocks reported by threads that are not server workers must be ignored: they are not consuming a
+        /// worker, so adding one would be pointless. (ProblemReportApi, for instance, reports a block from
+        /// both server and non-server code.)
+        /// </summary>
+        [Test]
+        public void ReportThreadBlocking_FromAThreadThatIsNotAWorker_DoesNotAddAWorker()
+        {
+            using (var server = new BloomServer(new BookSelection()))
+            {
+                server.EnsureListening();
+                var workersAtStart = server.WorkerCount;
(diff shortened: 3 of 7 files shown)


src/BloomExe/web/BloomServer.cs
```diff
@@ -2411,10 +2407,87 @@ public void RegisterThreadBlocking()
...
+                        // Note: Currently these workers are never stopped, so as not to complicate the code
+                        // any further.
+                        SpinUpAWorker();
+                        addedWorker = true;
+                    }
+                }
+                // Logged after releasing our own lock (QueueRequest's caller may still hold it), since
+                // logging can be slow. This is the event to look for in a log when investigating a freeze,
+                // so it is worth a line even though it is not an error.
+                if (addedWorker)
```
andrew-polk, line 2575: [Devin] **Bug**: Logging a newly added server helper outside the protected section can permanently mis-count blocked helpers

Devin's finding, in plain terms: this method promises in its own comment that nothing inside it can throw, because a caller bumps the blocked-worker count and only *then* enters the `try` whose `finally` brings it back down. Two `Logger.WriteEvent` calls sat outside that protected region -- the one in the `catch`, and the success log after it.

If either ever failed -- a log-file IO error, or an `ObjectDisposedException` racing `Logger.ShutDown` as Bloom exits -- the exception would escape `RegisterThreadBlocking` *after* the count went up but before any caller could bring it down. The count would then stay high for the rest of the process's life, and because this branch makes every later reported block re-check that count, the server would keep adding worker threads it never retires.

Devin rates it non-severe, but it is worth taking seriously here: it is the exact failure this method exists to prevent, and before this branch `RegisterThreadBlocking` could not throw at all, so the branch introduced the path.

src/BloomExe/Publish/OffScreenBrowser.cs
```diff
@@ -286,7 +286,36 @@ private T RunAndBlock<T>(Func<Task<T>> asyncFunc)
                 },
                 null
             );
-            return tcs.Task.GetAwaiter().GetResult();
+
+            // We are about to block, and publishing runs as an api request, so the thread we block is
```
andrew-polk, line 290: [Devin] **Investigate**: Comment understates which code paths already report a block

Devin's finding: the comment said BloomApiHandler was "the only other report on this path", which reads as though it were the only other one anywhere. `ApiRequest.InvokeWithErrorHandling` also reports a block, around marshalling a handler onto the UI thread.

The substantive worry Devin raises from that: api endpoints that ask to be run on the UI thread have their handler run *there*, not on a server worker. If publishing were one of those, the thread that reaches `RunAndBlock` would be the UI thread, `RegisterThreadBlocking` would do nothing (it only counts server workers), and the second half of this fix would never engage at all.

src/BloomExe/web/BloomServer.cs
```diff
@@ -2411,10 +2407,87 @@ public void RegisterThreadBlocking()
             // Check if the current thread looks like a Server Worker
             // If not, we can just ignore this request.
             // Notably, ProblemReportApi can be invoked by both server and non-server code
-            if (IsWorkerThread(Thread.CurrentThread))
+            if (!IsWorkerThread(Thread.CurrentThread))
```
andrew-polk, line 2427: [Devin] **Investigate**: Blocked-count leak in the UI-thread marshalling path can now grow the pool on every later block

Devin's finding, and our own local review found the same thing independently, which is why it is worth acting on rather than filing away:

Reporting a block and reporting an unblock both decide whether to count by looking at the name of the thread they happen to be running on. So the pair only balances when both halves run on the same worker thread. In `ApiRequest.InvokeWithErrorHandling` (`src/BloomExe/web/controllers/ApiRequest.cs:342-374`) the *register* happens on the worker, but the matching *unregister* runs after `await (Task)formForSynchronizing.Invoke(...)`. A server worker thread has no `SynchronizationContext`, so unless that Task is already finished by the time it is awaited, the continuation resumes on a thread-pool thread whose name is not `Server Worker Thread ...` -- and the decrement is silently skipped. The count is permanently one too high, once per occurrence.

Our review traced a concrete route: `libraryPublish/uploadAfterChangingBookId` (registered to run on the UI thread) -> `HandleUploadAfterChangingBookId` -> `await HandleUpload` -> `await UploadBookAsync()`. It also noted a second, smaller path: any exception out of `Invoke` other than `ObjectDisposedException` skips both halves entirely.

**This is pre-existing and this PR does not cause it.** But it makes it matter more. Before, an inflated count only added a worker when a *new request arrived*; now every reported block re-checks the count, so a steadily leaking count means a steadily growing set of worker threads that are never retired.

The real lesson is about the contract, not this one call site: a register/unregister pair must not straddle an `await`, and it should not depend on which thread happens to be running when the second half fires. Fixing that properly means the pair stops asking "am I on a worker thread?" twice and instead remembers what it did the first time -- which touches every call site and the `IBloomServer` interface. That is bigger than this branch's deliberate scope, so it is going to the developer rather than being folded in here.

src/BloomExe/web/BloomServer.cs
andrew-polk: (Claude Opus 5, during preflight) Fixed in `55ffdd1059b9ed1facd4e7174843839fc3009866`. Devin is right, and this was worth catching.

All the logging now sits inside the protected region, and the `catch`'s own logging is wrapped in its own `try`/`catch` -- because that call can fail for the same reasons the first one could, and it runs on the path where something has *already* gone wrong. So there is no longer any route by which this method can throw at a caller that has bumped the count but not yet armed its `finally`.

One note for anyone reading the thread: Devin re-ran against this commit and reported the finding again, but its description still cites the pre-fix line numbers and quotes the old comment ("Nothing in here may throw..."), which the fix replaced. The current code at `BloomServer.cs:2456-2479` has both log calls inside the `try`, so this is Devin repeating its earlier analysis rather than a second, still-live defect. Resolving on that basis.

src/BloomExe/Publish/OffScreenBrowser.cs
andrew-polk: (Claude Opus 5, during preflight) Both halves addressed in `55ffdd1059b9ed1facd4e7174843839fc3009866`.

**On the substantive worry -- does the fix actually engage?** Yes, and it is worth stating the evidence rather than asserting it. Both endpoints that drive page checks register with `handleOnUiThread: false` (`PublishToBloomPubApi.cs:104` and `PublishEpubApi.cs:125`), so `ApiRequest` takes its `else` branch and runs the handler directly on the server worker -- `InvokeWithErrorHandling` is never on this path, and neither is its block report. Our local review reached the same conclusion independently. The decisive evidence is empirical: the stacks captured from the failing CI run showed a thread named `Server Worker Thread ...` sitting inside `RunAndBlock`. So the thread we now report really is a worker, and reporting it is what turns the old `3 >= 4` (which declined to add a worker) into `4 >= 4`.

**On the comment itself** -- fair, and it now names both existing reports rather than implying BloomApiHandler is the only one: BloomApiHandler's covers *acquiring* the api lock, and ApiRequest's covers marshalling onto the UI thread, which the publish endpoints opt out of. It also records why that opt-out is what makes the blocked thread a worker rather than the UI thread, which is the very thing that makes reporting it worthwhile -- the connection Devin is pointing at.

Resolving. Note that a *different* problem in `ApiRequest`'s report -- its register/unregister pair straddling an `await` -- is real, and is tracked in its own thread, left open for the developer.

andrew-polk:
(Claude Opus 5, during preflight) Consulted Devin on 2026-08-05 14:13 USMST up to commit `55ffdd1059b9ed1facd4e7174843839fc3009866`.

It ran twice: once on the first pushed commit, and again after the fixes below. Three things came out of it, each in its own thread:

- **A real bug, now fixed** — two log calls sat outside the region that must not throw, which could have permanently inflated the blocked-worker count. Fixed in `55ffdd10`; thread resolved. (Devin repeated this finding on the re-review, but quoting the pre-fix code, so that repeat is stale.)
- **A comment that understated which paths already report a block** — reworded, and its underlying question (does the fix engage at all, or does publishing run on the UI thread?) verified against both the endpoint registrations and the stacks captured from the failing CI run. Thread resolved.
- **A pre-existing blocked-count leak in the UI-thread marshalling path** — real, found independently by our local review as well. Not caused by this PR, but this PR makes it matter more. Fixing it properly changes the register/unregister contract across every call site, so it is **left open for the developer** rather than folded in here.

Also 8 informational observations, not mirrored. Two are worth knowing: the top-up has no shutdown guard (benign today only because Dispose never disposes the wait handles), and the deliberate swallow-everything error handling is a conscious deviation from the repo's fail-fast guidance. Both are in the preflight report.

CI (`pr-automation`) passed. CodeRabbit is configured on this repo with `auto_review` disabled, so it was not summoned.

src/BloomExe/web/BloomServer.cs
```diff
@@ -2397,42 +2396,177 @@ public static string GetContentType(string extension)
...
+                // -- but a good deal cheaper than joining the queue behind those two, and QueueRequest
+                // already read Count on every request. Both reads are safe unsynchronized, and a stale one
+                // costs us nothing: the same check runs again every time another worker blocks or another
+                // request arrives.
+                //
+                // It is also correct rather than merely likely: Interlocked.Increment gives the increments a
+                // total order, so whichever thread performs the last one reads a count that includes every
+                // earlier block. The worker that exhausts the pool therefore always sees the shortage, even
+                // if the ones before it read stale values.
+                if (Volatile.Read(ref _countBlockedThreads) < _workers.Count)
```
andrew-polk: [Devin] **Investigate**: Dead worker threads still count toward the pool size in the shortage check

The shortage test is `blocked >= _workers.Count`. But `_workers` keeps an entry for a thread that has died until the listener loop notices and prunes it, so `Count` can overstate how many workers could actually pick up a request. The test may therefore conclude that a worker is free when all that is left of it is a dead entry -- and decline to add one.

This is pre-existing (the same expression lived in `QueueRequest`), but this PR makes that test **the** safeguard against the deadlock, so an inflated denominator now weakens the thing we are relying on. Worth noting that Bloom's own code says worker threads do sometimes die: the pruning loop exists precisely because of it.

Mitigating factors, for whoever picks this up: the pruning runs on the listener thread, which is not blocked during the deadlock and keeps accepting connections, so entries are usually cleaned up promptly; and a missed check is retried on the next reported block or the next arriving request rather than being final. So this looks like a delay in the safeguard rather than a permanent failure of it. Devin's suggested improvement is to compare against the count of live workers, which at this pool size costs almost nothing.

Left open for the developer: it changes the semantics of the central check, and the preflight fix-and-re-review cycle had reached its cap.

src/BloomExe/web/BloomServer.cs
```diff
@@ -2551,6 +2685,23 @@ protected virtual void Dispose(bool fDisposing)
                         // tell _listenerThread and the worker threads they should stop
                         _stop.Set();
 
+                        // Note (BL-16612): a worker part-way through a request can still report a block while
```
andrew-polk, line 2740: [Devin] **Investigate**: New worker created after shutdown depends on `_stop` being signalled

This one found a hole in a comment this PR had just added. The note claimed a worker created late during shutdown is harmless "because we merely SIGNAL `_stop` and leave both handles alive", so the new worker wakes at once and exits.

That signal is itself conditional. `_stop.Set()` sits inside `if (_listener != null)`, and `CloseListener()` is public, nulls `_listener`, and by its own comment can have run first from the shutdown timer. In that case `_stop` is never signalled at all.

It matters because worker threads are **foreground** threads, so one still waiting on `_ready`/`_stop` keeps the process alive -- Bloom would not exit. Pre-existing workers have always been exposed to that; what the top-up adds is the possibility of creating a *new* such thread during that window.

Verified against the code rather than taken on trust: `SpinUpAWorker` does not set `IsBackground`, and `CloseListener` does null `_listener`.

**Fixed the misleading comment in `0ab354bcc9`** -- it now states both conditions the safety argument actually depends on, since leaving a comment that asserts a guarantee the code does not provide is worse than the gap itself. The behavioural fix (guard the top-up on shutdown, or make added workers background threads) is **left open for the developer**, as it changes behaviour.

src/BloomTests/web/BloomServerWorkerPoolTests.cs
```diff
@@ -0,0 +1,215 @@
...
+    {
+        /// <summary>
+        /// The whole point of reporting a block (BL-16612): once every worker is blocked, no worker is left
+        /// to take a request off the queue -- including the request that would let the blocked workers
+        /// finish. So the server has to add one.
+        /// </summary>
+        [Test]
+        public void ReportThreadBlocking_WhenTheLastFreeWorkerBlocks_AddsAWorker()
+        {
+            using (var server = new BloomServer(new BookSelection()))
```
andrew-polk, line 32: [Devin] **Investigate**: Test fixture constructs BloomServer instances that overwrite the process-wide singleton

`BloomServer`'s constructor assigns `_theOneInstance = this` and nothing clears it on Dispose, so each test in this fixture leaves that static pointing at a disposed server. A later test in the same NUnit process that reaches code using it could in principle be affected.

src/BloomExe/web/BloomServer.cs
andrew-polk: (Claude Opus 5, during preflight) **Decision: fix it properly. Done in `6607a4f28a`.**

Andrew chose to fix the contract rather than file it, and to do it on this branch. The pair no longer depends on the calling thread, and no longer depends on the caller remembering anything either.

`RegisterThreadBlocking`/`RegisterThreadUnblocked` are gone. In their place, `ReportThreadBlocking()` returns an `IDisposable` scope that closes over whether the block was actually counted. Disposing it undoes exactly that block, from any thread, on every exit path including an exception. So both leaks in this method are closed: the cross-thread one Devin identified, and the one where any exception out of `Invoke` other than `ObjectDisposedException` skipped both halves entirely. All five call sites now use `using`.

`RegisterThreadUnblocked` was deleted rather than kept alongside the new method. A second way to end a block is the same class of trap as the two-method top-up this branch already removed -- there is now exactly one way to do it, and the compiler places it.

A regression test pins the actual bug rather than the implementation: it disposes the scope on a thread-pool thread (which is where an await's continuation lands) and asserts the count returns to zero. I verified it **fails** if the thread-identity check is put back, so it is not passing for the wrong reason. There is also a double-dispose test, since a `using` nested inside a caller that also disposes would otherwise decrement twice and leave the server believing workers were free when they were not.

Note for later readers: Devin has repeated this finding on each subsequent re-review, but its text still refers to `RegisterThreadBlocking`/`RegisterThreadUnblocked`, which no longer exist anywhere in the tree. Those repeats are stale. Resolving.

src/BloomTests/web/BloomServerWorkerPoolTests.cs
andrew-polk: (Claude Opus 5, during preflight) Not worth acting on, and resolving on that basis.

This is the established pattern in this repo rather than something this fixture introduced: BloomServerTests, OffScreenBrowserTests and TeamCollectionApiTests all construct real BloomServer instances the same way, so the static has always been left pointing at the most recently constructed (often disposed) server.

Checked empirically rather than assumed: the full C# suite is green with these four tests added -- 3026 passed, 12 skipped, and the single failure is an unrelated missing Reading App Builder toolchain on this machine. Any later fixture that depended on that static would already have been affected by the existing fixtures.

Clearing the static on Dispose would be a genuine improvement to test isolation, but it changes production code to suit tests and would alter behaviour for anything that reads _theOneInstance after a server is disposed. That is not a call worth making inside this PR.

andrew-polk:
(Claude Opus 5, during preflight) Consulted Devin on 2026-08-05 15:25 USMST up to commit `0ab354bcc9b5d6fc745415383af7ab37e3b401fc`.

Fourth and final run for this preflight, after the developer's decisions were implemented. **The fix-and-re-review cycle has reached its cap of 4**, so the two findings below are left for the developer rather than starting a fifth round.

**Closed since the last log**
- The blocked-count leak in the UI-thread marshalling path — the developer chose to fix the contract properly. Done in `6607a4f28a`; thread resolved with the reasoning.
- Test fixture overwriting the process-wide singleton — assessed as the established pattern in this repo (three existing fixtures do the same) and confirmed harmless against a green suite. Replied and resolved.

**Left open for the developer**
- *Dead worker threads still count toward the pool size.* Pre-existing, but this PR makes that comparison the central safeguard, so an inflated denominator weakens it. Looks like a delay rather than a permanent failure, since pruning runs on the unblocked listener thread and the check is retried.
- *A worker created late in shutdown depends on `_stop` being signalled*, and that signal is conditional on `_listener != null`. Workers are foreground threads, so one left waiting would stop Bloom exiting. The misleading comment this PR had added was corrected in `0ab354bcc9`; the behavioural fix is the developer's call.

**Stale repeats, for the record:** Devin has re-reported the logging bug and two earlier Investigate items on every subsequent run, but their text still cites `RegisterThreadBlocking`/`RegisterThreadUnblocked` and the old comment wording — none of which exist in the tree any more. Those threads stay resolved.

Also 8 informational items, not mirrored. Full C# suite at this HEAD: 3026 passed, 12 skipped, 1 failed (an unrelated missing Reading App Builder toolchain on this machine). CI (`pr-automation`) passed. CodeRabbit remains disabled by repo config.

src/BloomExe/web/BloomServer.cs
andrew-polk: (Claude Opus 5, during preflight) **Decision: count only live workers. Done — but not where the decision said, and that turned out to matter.**

Andrew chose to compare against live workers in the under-lock check. I implemented exactly that first, and the test written for it **failed**.

The reason is worth recording. The under-lock check never ran. The fast path in front of it -- the cheap test that exists purely to avoid taking the queue lock -- was still comparing against the raw entry count, so with a dead entry inflating that count it concluded "a worker is free", returned early, and the accurate check below was unreachable. A cheap-but-over-counting gate in front of a correct check simply makes the correct check dead code.

So the live count is now used in **both** places. That is a deliberate departure from the letter of the decision, in order to achieve its intent; flagging it rather than quietly widening the change. The cost is that the hot path now walks the worker collection (four to ten entries, one `IsAlive` each) instead of reading a count. That is still far cheaper than joining the queue behind the listener and the workers, which is the only thing the fast path exists to avoid.

The test plants a dead thread's entry via reflection -- there is no honest way to make a real worker die on cue -- then blocks every *live* worker and asserts a worker is still added. Judged on raw entries that looks like a worker is free and nothing is added; judged on live workers the pool is exhausted and one must be. It fails with the fix in only one of the two places, which is how the gap above was found in the first place.

Resolving.

src/BloomExe/web/BloomServer.cs
andrew-polk: (Claude Opus 5, during preflight) **Decision: leave as is.** Andrew's call, recorded here so the thread ends with an outcome rather than silence.

To be explicit about what is being accepted: if `CloseListener()` has already run when `Dispose` reaches the stop signal, that signal is skipped, and any worker still waiting -- long-standing or newly added by the top-up -- can keep Bloom's process alive, because workers are foreground threads. Nothing in this PR causes that; the top-up only adds the possibility of a *new* such thread inside that window.

What is in place instead of a behavioural fix: the note at `Dispose` now states both conditions the safety argument actually depends on, including that reaching the signal at all is conditional. It also names the two candidate fixes (guard the top-up on shutdown, or make added workers background threads) so whoever revisits it does not have to re-derive them. That correction went in as `0ab354bcc9`, because a comment asserting a guarantee the code does not provide is worse than the gap it describes.

No separate code comment was requested for this decision, and the `Dispose` note already carries the reasoning, so nothing further was added. Resolving.

andrew-polk:
(Claude Opus 5, during preflight) Consulted Devin on 2026-08-05 15:51 USMST up to commit `119def1f149939fa078327b8aeee5f0f91b28070`. Fifth and final run.

**Nothing new.** Every finding on this pass is a repeat of one already handled, and each cites code that is no longer in the tree:

- the logging bug and two earlier Investigate items still quote `RegisterThreadBlocking`/`RegisterThreadUnblocked` and the old comment wording, none of which exist;
- the dead-worker finding still quotes `_countBlockedThreads >= _workers.Count`, which is exactly the expression this commit replaced — both shortage checks now call `LiveWorkerCount()`;
- the shutdown and test-singleton items are the ones the developer decided on, already recorded on their threads.

I checked each against the code rather than trusting the repeat. No new threads posted; the six existing ones stay resolved.

**All six threads now carry a documented outcome:** three fixed, one fixed per the developer's decision, one accepted as-is with the reasoning recorded, one assessed as not worth acting on.

Full C# suite at this HEAD: **3027 passed, 12 skipped, 1 failed** — the failure is `RabRealBuildTests`, which needs Reading App Builder's `buildapp.bat` on the PATH and is not installed on this machine; the test reports that itself. CI (`pr-automation`) passed. CodeRabbit remains disabled by repo config.

One note on the last fix, since it departed from the letter of the decision: the live-worker count was applied to the fast path as well as the under-lock check. Applying it only under the lock does not work, because the fast path's over-count made the accurate check unreachable — the new test caught that. Detail is on that thread.

src/BloomExe/Publish/OffScreenBrowser.cs
```diff
@@ -286,7 +286,25 @@ private T RunAndBlock<T>(Func<Task<T>> asyncFunc)
...
+            // Without this the server could not tell that its whole pool was blocked. Reporting a block is
+            // opt-in, and neither of the two existing reports covers us. BloomApiHandler's covers ACQUIRING
+            // the api lock, not the long stretch while publishing holds it. ApiRequest's covers marshalling
+            // a handler onto the UI thread, which the publish endpoints deliberately opt out of (they
+            // register with handleOnUiThread: false), so it never runs on this path -- and that opt-out is
+            // also why the thread we block here is a server worker rather than the UI thread, which is what
+            // makes reporting it worth doing at all. So every other worker could be waiting for that lock,
+            // and this one waiting here, with the server believing a worker was still free. That is
+            // BL-16612.
+            using (BloomServer._theOneInstance?.ReportThreadBlocking())
```
andrew-polk, line 304: [Devin] **Investigate**: Worker-thread detection happens at report time and may miss post-await continuations

Devin's point: whether a block counts is decided by inspecting the name of the thread doing the reporting. On the publish path a worker is parked in `MakeReplyAsync(info).GetAwaiter().GetResult()` (`BloomServer.cs:2140`), but if any `await` in the endpoint handler chain resumed on a thread-pool thread before `RunAndBlock` was reached, the report here would be a no-op even though a worker really is tied up -- and the top-up this PR exists to trigger would not happen. It asks that this be verified against the actual publish chain rather than assumed.

andrew-polk: (Claude Opus 5, during pr-ready-for-human) Verified, and it holds. Two independent ways:

**From the code.** Both endpoints that drive page checks register *synchronous* lambdas, not async ones -- `publish/bloompub/updatePreview` (`PublishToBloomPubApi.cs:104`) and `publish/epub/updatePreview` (`PublishEpubApi.cs:125`). A synchronous handler runs inline to completion, so there is no suspension point between the worker entering the handler and `RunAndBlock` being called, and therefore no opportunity for a continuation to move to another thread. The worker stays put in `GetAwaiter().GetResult()` the whole way down.

**From the failure itself.** The stacks captured at the moment of the BL-16612 hang on CI showed a thread named `Server Worker Thread ...` sitting *inside* `RunAndBlock`. That is the scenario this PR is about, and it is direct evidence rather than inference: the thread that blocks there is a worker, so the report counts and the top-up fires.

Devin is right about the general shape, though, and it is worth stating rather than dismissing: the effectiveness of a report does depend on it being made *on* the worker thread, and nothing structurally guarantees that for a future caller. If someone later makes one of these handlers async with a real suspension point before the browser call, the report would quietly stop counting. That is a pre-existing property of how a block is detected -- the old code decided the same way at the same moment -- so this PR neither introduces nor worsens it. What this PR did fix is the *other* half, where the decision was made twice and could differ between the two.

Not an issue for this change. Resolving.

src/BloomTests/web/BloomServerWorkerPoolTests.cs
```diff
@@ -0,0 +1,291 @@
...
```
andrew-polk, line 32: [Devin] **Investigate**: Tests spin up real listening servers on fixed ports

`EnsureListening()` binds ports starting at 8089 and, if it cannot open any of its candidates, calls `ProgramExit.Exit()` rather than throwing -- which would terminate the test process instead of failing a test. On a developer machine already running Bloom, or on CI running suites in parallel, that makes the fixture dependent on port availability. Each test also leaves several foreground worker threads alive until its server is disposed.

andrew-polk: (Claude Opus 5, during pr-ready-for-human) Fair, and worth knowing, but not something to change here.

This is the established pattern rather than something this fixture introduces: `OffScreenBrowserTests` already stands up a real listening server via `EnsureListening()` for exactly the same reason -- the worker pool only exists once the server is listening, so there is nothing to test without one. `BloomServerTests` and `TeamCollectionApiTests` also construct real servers.

On the specific risks: the port search tries a range rather than a single port, and the developer's own Bloom is expected to be running while tests run (the repo's `AGENTS.md` documents that as the normal state and provides the isolated build wrapper for it). The full suite was run repeatedly during this work with Bloom's usual ports in play and these tests passed every time. The foreground threads are disposed with the server at the end of each test, inside a `using`.

The underlying sharp edge -- `EnsureListening` calling `ProgramExit.Exit()` instead of throwing, so a port exhaustion kills the run rather than failing a test -- is a genuine problem with that method's contract, not with these tests, and would be worth its own ticket if anyone hits it. Changing production shutdown behaviour to suit a test is not something to slip into this PR.

Resolving.

JohnThomson (changes requested):
> @JohnThomson reviewed 7 files and all commit messages, made 3 comments, and resolved 8 discussions.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8155)* status: all files reviewed, 3 unresolved discussions (waiting on andrew-polk).

___
*[`src/BloomExe/web/BloomServer.cs` line 2512 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8155#-OzMvvwWA-B5TBdgqcrl:-OzMvvwWA-B5TBdgqcrm:ber62ks) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/b689aa21bf928af8ed1c6f6868bb7256b795b115/src/BloomExe/web/BloomServer.cs#L2512)):*
> ```Smalltalk
>                 // the listener prunes it -- we return early and the accurate check never runs at all. A
>                 // cheap-but-wrong gate in front of a correct check just makes the correct check
>                 // unreachable.
> ```

The paragraph above puzzles me. I would expect that it's OK for the cheap gate to give a wrong answer, provided that the result of being wrong is that we DO take the lock and compute the accurate answer. But the text above indicates that a wrong answer will PREVENT us ever reaching the reliable (but slow?) code.
Let's see. We're going to return (not create a new worker) if we have more usable threads than ones that are blocked. The danger here is that LiveWorkerCount() accesses _workers, and we might be racing another thread that has already decided to change _workers by calling SpinUpAWorker(). But it hasn't actually modified _workers yet, so LiveWorkerCount() will return a number lower than it would get by waiting for the lock. And possibly, that one-too-low count is not greater than _countBlockedThreads, so we will NOT return early. Instead, we will wait for the lock, which we won't get until the other thread finishes adding its new thread...and then LiveWorkerCount() will be right and we won't needlessly add yet another.
So for this mode of wrongness, the code is right but the comment above is wrong: the cheap check outside the lock is always wrong in the direction of a low value of LiveWorkerCount(), but the result of that wrongness is that we DO go on and make the accurate count.
The other failure mode hinted at above is that a thread dies while we are counting, causing LiveWorkerCount() to return a count that is lower than the true answer at the instant it returns. That could indeed cause us to skip the "accurate" check inside the lock. But if I'm understanding correctly, we can't help that; the 'accurate' count is subject to the same risk. The lock only protects _workers; threads don't hold it the whole time they are working. So it's always possible that after we do this check and decide we have enough live threads, the one and only live, unblocked thread dies. I think that could even be a problem if it happened after we got the count.
We could mitigate this risk by having one 'spare' thread: spin one up if we have less than two live, unblocked threads instead of if we have less than one. Of course, they could both die. But it would reduce what is probably already a very small risk. Not sure if it's worth it.
Another note: it's safe to do this preliminary check unlocked mainly because nothing  ever REMOVES a live thread from _workers, which might cause LiveWorkerCount() to give a wrongly HIGH answer when called unlocked, in which case this early return could cause us to skip creating an extra thread we need. This might be worth noting in a comment or two, perhaps on the function and also on the _workers variable and/or the lock variable.
___
*[`src/BloomExe/web/BloomServer.cs` line 2567 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8155#-OzMtw3E6oUIITOsoqT0:-OzMtw3E6oUIITOsoqT1:b-vktpph) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/b689aa21bf928af8ed1c6f6868bb7256b795b115/src/BloomExe/web/BloomServer.cs#L2567)):*
> ```Smalltalk
>         /// entry in _workers until that pruning runs, so _workers.Count can overstate the pool.
>         ///
>         /// Walked rather than LINQ-counted because this runs while holding the queue lock.
> ```

That doesn't help me. What is wrong with using LINQ while holding a lock?
___
*[`src/BloomExe/web/BloomServer.cs` line 2593 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8155#-OzN2XKY4qjDCsnl0Wrg:-OzN2XKY4qjDCsnl0Wrh:bddvcvw) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/b689aa21bf928af8ed1c6f6868bb7256b795b115/src/BloomExe/web/BloomServer.cs#L2593)):*
> ```Smalltalk
> 
>         private bool IsWorkerThread(Thread thread) =>
>             thread?.Name?.IndexOf(WorkerThreadNamePrefix) == 0;
> ```

I don't think this PR changes how often we have to call this function, so this is arguably out of scope. But while we're in this code...surely it would be more efficient to use StartsWith than to search the whole name for a match we're only interested in at index 0?

src/BloomExe/web/BloomServer.cs
```diff
@@ -2397,44 +2405,211 @@ public static string GetContentType(string extension)
...
+                // Logged after releasing our own lock (QueueRequest's caller may still hold it), since
+                // logging can be slow. This is the event to look for in a log when investigating a freeze,
+                // so it is worth a line even though it is not an error.
+                if (addedWorker)
+                    Logger.WriteEvent(
+                        "BloomServer: every worker was blocked, so added one to keep requests moving "
+                            + $"({_workers.Count} workers, {_countBlockedThreads} blocked)."
+                    );
+            }
+            catch (Exception e)
```
andrew-polk, line 2581: [Devin] **Bug**: Server quietly hides failures when topping up its pool of request handlers

Every failure while adding a spare worker is swallowed by the blanket `catch (Exception e)` here, so a genuine inability to create a worker is invisible apart from a log line. Devin's stated impact: if the safety net that keeps Bloom responsive stops working, nothing tells the user or the developers, and Bloom can simply freeze again with no error. It cites `AGENTS.md` -- *"Fail Fast. Don't write code that silently works around failed dependencies"* and *"Don't be overly defensive about error handling... it's fine for it to throw an error. We want to know about it so we can fix it."*

Devin raised this as informational on an earlier commit, and the developer's recorded decision then was to keep the catch-all, on the grounds that an exception escaping here would leak the blocked count that `ReportThreadBlocking` increments *before* calling this -- the very condition the method exists to relieve. That reasoning is in the `REVIEWED DECISION` comment above.

**What is new, and why this is worth re-opening rather than dismissing as a repeat:** Devin now proposes a concrete alternative that preserves fail-fast *without* the leak --

> create the scope (so the decrement is guaranteed via `using`/`try`) before calling the top-up, and let real exceptions propagate

That is a materially cheaper option than the one the decision was weighed against. The earlier framing -- repeated in the code comment -- was that the alternative meant "moving the increment inside the caller's own protected region", which "would mean every caller had to get that right". Devin's version touches only `ReportThreadBlocking`: construct the `BlockedWorkerScope` first, call the top-up inside a `try` that disposes the scope and rethrows, and then `EnsureAWorkerCanStillTakeWork` no longer needs to promise it cannot throw. No call site changes at all.

So the cost side of the recorded decision is now wrong, and the comment overstates it. Two things follow, both for the developer rather than for me to decide:

1. Whether to adopt the restructure. It is a few lines, and it would let the catch-all (and its nested empty `catch`) go away entirely.
2. Whether we actually *want* a failed `SpinUpAWorker` to fail the request that triggered it. That is the real question underneath, and it is a judgement call rather than a correctness one: `AGENTS.md` says we want to know; the counter-argument is that adding a worker is a safety net and a failure to add one is not itself a reason to fail a user's request. Devin's suggestion makes fail-fast *possible*, but choosing it is still a choice.

Left open deliberately. It reverses a decision the developer made explicitly, so it is not mine to fold in -- but the ground it was decided on has shifted, which is why the thread is open rather than closed as a duplicate.

andrew-polk (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8155)* status: 6 of 7 files reviewed, 4 unresolved discussions (waiting on JohnThomson).

___
*[`src/BloomExe/web/BloomServer.cs` line 2512 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8155#-OzMvvwWA-B5TBdgqcrl:-OzNTM5q1bhIDUbofQsj:b-r69yx2) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/b689aa21bf928af8ed1c6f6868bb7256b795b115/src/BloomExe/web/BloomServer.cs#L2512)):*

The paragraph above puzzles me. I would expect that it's OK for the cheap gate to give a wrong answer, provided that the result of being wrong is that we DO take the lock and compute the accurate answer. But the text above indicates that a wrong answer will PREVENT us ever reaching the reliable (but slow?) code.
Let's see. We're going to return (not create a new worker) if we have more usable threads than ones that are blocked. The danger here is that LiveWorkerCount() accesses _workers, and we might be racing another thread that has already decided to change _workers by calling SpinUpAWorker(). But it hasn't actually modified _workers yet, so LiveWorkerCount() will return a number lower than it would get by waiting for the lock. And possibly, that one-too-low count is not greater than _countBlockedThreads, so we will NOT return early. Instead, we will wait for the lock, which we won't get until the other thread finishes adding its new thread...and then LiveWorkerCount() will be right and we won't needlessly add yet another.
So for this mode of wrongness, the code is right but the comment above is wrong: the cheap check outside the lock is always wrong in the direction of a low value of LiveWorkerCount(), but the result of that wrongness is that we DO go on and make the accurate count.
The other failure mode hinted at above is that a thread dies while we are counting, causing LiveWorkerCount() to return a count that is lower than the true answer at the instant it returns. That could indeed cause us to skip the "accurate" check inside the lock. But if I'm understanding correctly, we can't help that; the 'accurate' count is subject to the same risk. The lock only protects _workers; threads don't hold it the whole time they are working. So it's always possible that after we do this check and decide we have enough live threads, the one and only live, unblocked thread dies. I think that could even be a problem if it happened after we got the count.
We could mitigate this risk by having one 'spare' thread: spin one up if we have less than two live, unblocked threads instead of if we have less than one. Of course, they could both die. But it would reduce what is probably already a very small risk. Not sure if it's worth it.
Another note: it's safe to do this preliminary check unlocked mainly because nothing  ever REMOVES a live thread from _workers, which might cause LiveWorkerCount() to give a wrongly HIGH answer when called unlocked, in which case this early return could cause us to skip creating an extra thread we need. This might be worth noting in a comment or two, perhaps on the function and also on the _workers variable and/or the lock variable.
</blockquote></details>

[Claude Opus 5] You're right, and the comment was the problem rather than the code. Rewritten in 434cce5.

Your expectation is the correct one: a wrong answer from the cheap check is fine *provided* being wrong means we go on and take the lock. What the comment failed to say is that the two directions are not symmetric, and that only one of them was actually happening.

Counting too **low** — your half-finished `SpinUpAWorker` example — behaves exactly as you expect: `blocked >= live` is more likely to hold, so we decline to return early, wait for the lock, and get the better answer. Self-correcting, and that is really *why* reading it unlocked is acceptable at all.

Counting too **high** is the dangerous direction, because the only thing this test can do is skip the check below. And the raw entry count was *systematically* too high, since a dead thread keeps its entry until the listener prunes it. So it wasn't a rare race — it was a reliable over-count that made the careful check unreachable. That is what the test I added catches: it plants a dead entry, blocks every live worker, and asserts a worker is still added. It fails if the live count is applied in only one of the two places.

On your second point I've conceded in the code, because you're right and my comment was overclaiming: the check under the lock is **not** "accurate" in the sense that word implies. The lock covers changes to `_workers`, not thread liveness, so a worker can die immediately after either reading and neither is authoritative by the time we act on it. The comment now says that explicitly, and says that what counting live workers removes is the systematic over-count from lingering dead entries — not that race.

On the spare thread: we discussed it and decided against, though it was a close call. The reasoning was that Bloom already recovers from dead workers by another route (the listener notices them and replaces them), so the window needs a worker to die in the gap between our count and our decision, with no other worker free; against that, a permanent margin means one more thread for the rest of every session in a pool that is never retired, which is a direction we'd already accepted reluctantly. Happy to revisit if you think that trade is wrong — you've thought about this code longer than I have.

Your last point was the most useful one and I've taken it: the reason the unlocked read is safe is that entries are only ever removed for threads that are **already dead**, so a concurrent removal can't inflate a live count. Nothing removes a live worker. That's now documented on `_workers` itself, along with the other property callers depend on (additions only happen under the lock), since as you say it wasn't written down anywhere.
___
*[`src/BloomExe/web/BloomServer.cs` line 2567 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8155#-OzMtw3E6oUIITOsoqT0:-OzNTMZ-2w4pDdcOT0wa:by1xyoj) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/b689aa21bf928af8ed1c6f6868bb7256b795b115/src/BloomExe/web/BloomServer.cs#L2567)):*

That doesn't help me. What is wrong with using LINQ while holding a lock?
</blockquote></details>

[Claude Opus 5] Fair hit — nothing is wrong with LINQ while holding a lock, so that was no reason at all. Fixed in 434cce5.

It was worse than an unhelpful comment, in two ways you couldn't have seen from the diff:

The premise had gone stale. When I first wrote that method it was only called from inside the lock; by the time you read it, it was also being called from the fast path above, outside the lock. So the sentence wasn't just unilluminating, it was false.

And the hand-rolled loop didn't even achieve the thing that *would* have been a real reason. If the point were avoiding allocation on a hot path, iterating `_workers.Values` doesn't help — on a `ConcurrentDictionary`, `Values` materialises a snapshot `List` on every call. So I paid for a snapshot and wrote a manual loop to avoid an allocation I was making anyway.

It is now a one-line LINQ count over the dictionary itself, which enumerates lazily:

```csharp
private int LiveWorkerCount() => _workers.Count(kvp => kvp.Value?.IsAlive == true);
```

with the actual reason stated (avoiding the `Values` snapshot, on something called for every api request that waits for a lock) and a note that callers may or may not hold the lock.
___
*[`src/BloomExe/web/BloomServer.cs` line 2593 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8155#-OzN2XKY4qjDCsnl0Wrg:-OzNTMyg4Skxrx2C9Mwk:brtoeke) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/b689aa21bf928af8ed1c6f6868bb7256b795b115/src/BloomExe/web/BloomServer.cs#L2593)):*

I don't think this PR changes how often we have to call this function, so this is arguably out of scope. But while we're in this code...surely it would be more efficient to use StartsWith than to search the whole name for a match we're only interested in at index 0?
</blockquote></details>

[Claude Opus 5] Agreed and done in 434cce5, and there's a better argument for it than efficiency.

`String.IndexOf(string)` uses the **current culture** by default, so that line was doing a linguistic search — on a thread name we generated ourselves from a constant prefix. Culture-sensitive comparison is simply the wrong tool for that, quite apart from scanning the whole name for a match we only care about at index 0. It's now:

```csharp
thread?.Name?.StartsWith(WorkerThreadNamePrefix, StringComparison.Ordinal) == true
```

Ordinal explicitly, because `StartsWith`'s default overload has the same culture-sensitivity trap as `IndexOf`'s.

On scope: you're right that this PR doesn't change how often it's called, but I'd argue it's fair game now for a different reason — this predicate decides whether a reported block counts, and that count is what the PR makes responsible for deciding when to add a worker. It went from bookkeeping to load-bearing, so it's worth having it be exactly right. The worker-pool tests cover it, since they report blocks from a thread named with that prefix and would fail if the match broke.

src/BloomExe/web/BloomServer.cs
andrew-polk: (Claude Opus 5) **Decision: keep the catch-all. Declined on judgement, not on cost — and the comment now says so.** Recorded in `e2517f5717`.

Devin is right that this deviates from `AGENTS.md`, and right that a cheaper alternative exists than the one this was originally weighed against. That matters enough to spell out, because the recorded reasoning was wrong on the cost:

The earlier justification claimed the alternative meant "moving the increment inside the caller's own protected region", so "every caller had to get that right". Devin's version does not: hand the caller its scope *before* calling the top-up, dispose and rethrow if it throws, and the top-up is then free to fail fast. That touches only `ReportThreadBlocking` — no call sites. So the cost objection this decision rested on was overstated, and the code comment saying so has been rewritten rather than left to mislead the next reader.

With the cost objection gone, the choice comes down to the question underneath: **should a failure to create a spare worker fail the request that triggered it?** The developer's answer is no, and the reasoning is now in the comment: topping up the pool is an *opportunistic safety net*. A net failing is not itself a reason to fail the user's request that happened to trip it — publishing a book should not die because the server could not create a spare thread it may well never have needed. `AGENTS.md`'s fail-fast guidance is about not silently working around broken dependencies; here the "dependency" is a redundancy we added ourselves, and the request in hand can still be served without it.

What did change: the comment no longer justifies itself with a cost that isn't real, and it now names the restructure explicitly as *the* way to reverse this if someone later disagrees with the trade — so the next person can change it deliberately rather than by narrowing the catch and quietly reintroducing the leak.

Resolving on that basis.

andrew-polk (commented):
> @andrew-polk partially reviewed 2 files and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8155)* status: 7 of 8 files reviewed, 3 unresolved discussions (waiting on JohnThomson).

JohnThomson (approved):
> @JohnThomson reviewed 1 file and all commit messages, and resolved 3 discussions.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8155)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on andrew-polk).

PR #8156, opened 2026-08-05
Don't require eSpeak to apply a timings file (BL-16640)

JohnThomson:
## The problem

Bloom's Talking Book "Adjust Timings" dialog has two menu items that both hit the same C# endpoint, `AudioSegmentationApi.GetAeneasTimings`:

- **Use Aeneas to guess timings** — runs forced alignment (Python + Aeneas + eSpeak + FFMPEG).
- **Apply timings file...** — just reads start/end times out of a tab-separated file the user picked.

The second one was still shelling out to `espeak -v <lang> -q "hello world"` (via `GetBestSupportedLanguage`) to choose a TTS language code. eSpeak only arrives as an Aeneas dependency, so most users don't have it; for them the probe returned null and Bloom reported

```
eSpeak error:
'espeak' is not recognized as an internal or external command
```

and applied nothing. That also accounts for the card's second symptom — edits to the timings file "not being applied" — because we aborted before applying anything.

## The fix

There's nothing on this path that needs eSpeak; the question it was answering doesn't arise. The TTS language code exists purely to tell Aeneas' engine which voice to synthesize with (and, on fallback, which orthography conversion to apply) — meaningless when we're just reading numbers out of a file.

- Gate the eSpeak language probe on `!usingManualTimings`. `langCode` then stays the requested code, so the `langCode != requestedLangCode` orthography branch falls through naturally.
- Move `WriteAllLines(textFragmentsFilename, ...)` into the Aeneas-only branch. Only Aeneas reads that file; the manual path was writing it into the book's `audio` folder and immediately deleting it.
- Name the condition `usingManualTimings` once instead of re-testing the path string.

Net effect: **"Apply Timings File..." now runs no external process at all.**

## Test

`GetAeneasTimings_ManualTimingsFile_RunsNoExternalCommands` subclasses the API with an override that fails if *any* external command is attempted — so it guards against python/aeneas/ffmpeg creeping back in too, not just eSpeak. That required making the five-arg `DoesCommandCauseError` `virtual` (it was already `protected`; the other overloads funnel through it).

Verified it's a real guard: with the fix reverted it fails with `Should not have run an external command: espeak -v qaa -q "hello world"` — the exact reported failure.

## Also fixed: network-share collections

`GetAeneasTimings` additionally refuses to proceed at all if the collection path doesn't start with a drive letter — typically a collection on a network share. That's another Python/Aeneas limitation (BL-9959) that has no business applying to manual timings, so a second commit moves it under the same `usingManualTimings` guard. Without it, a user working off a network share still couldn't apply a timings file even after the eSpeak fix.

The test for it asserts the manual path gets *past* the guard and applies the timings. It uses the Win32 extended-length `\\?\` form of a real local folder rather than a genuine `\\server\share` path, so it exercises that branch without depending on network name resolution — which would make it slow and flaky on a build agent with an unusual DNS/WINS setup.

(An earlier revision of this description said this was deliberately left out of scope. That's no longer true — thanks to Devin for catching the contradiction.)

Ref: https://issues.bloomlibrary.org/youtrack/issue/BL-16640

[Devin review](https://devinreview.com/BloomBooks/BloomDesktop/pull/8156)

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8156)

### DIFF
--- a/src/BloomExe/web/controllers/AudioSegmentationApi.cs
+++ b/src/BloomExe/web/controllers/AudioSegmentationApi.cs
@@ -229,8 +229,9 @@ namespace Bloom.web.controllers
             );
         }
 
-        // Returns true if the command returned with an error
-        protected bool DoesCommandCauseError(
+        // Returns true if the command returned with an error.
+        // Virtual so that tests can verify that a code path runs no external commands at all.
+        protected virtual bool DoesCommandCauseError(
             string commandString,
             string workingDirectory,
             out string standardOutput,
@@ -363,10 +364,17 @@ namespace Bloom.web.controllers
                 warningMessage = "",
             };
 
+            // "Apply Timings File..." just reads the times out of a file the user already has,
+            // so none of the forced-alignment machinery (Python, Aeneas, eSpeak, FFMPEG) is involved.
+            var usingManualTimings = !string.IsNullOrEmpty(requestParameters.manualTimingsPath);
+
             // The client was supposed to validate this already, but double-check in case something strange happened.
             string directoryName = GetAudioDirectory();
 
-            if (!Platform.IsLinux)
+            // Aeneas runs under Python, which can't cope with a UNC path, so we refuse up front rather
+            // than fail obscurely later. That's a limitation of the forced-alignment tooling, though, so
+            // it doesn't apply when we're only reading times out of a file the user chose.
+            if (!usingManualTimings && !Platform.IsLinux)
             {
                 if (directoryName.StartsWith("\\"))
                 {
@@ -406,10 +414,7 @@ namespace Bloom.web.controllers
             // contain a hot link here. That code is in Typescript.
             // If we're just trying to read a manual timings file, we don't need to check the Aeneas installation.
             string message;
-            if (
-                string.IsNullOrEmpty(requestParameters.manualTimingsPath)
-                && !AreAutoSegmentDependenciesMet(out message)
-            )
+            if (!usingManualTimings && !AreAutoSegmentDependenciesMet(out message))
             {
                 var localizedFormatString = L10NSharp.LocalizationManager.GetString(
                     "EditTab.Toolbox.TalkingBook.MissingDependency",
@@ -422,24 +427,32 @@ namespace Bloom.web.controllers
                 return null;
             }
 
-            // When using TTS overrides, there's no Aeneas error message that tells us if the language is unsupported.
-            // Therefore, we explicitly test if the language is supported by the dependency (eSpeak) before getting started.
-            string stdOut;
-            string stdErr;
-            string langCode = GetBestSupportedLanguage(requestedLangCode, out stdOut, out stdErr);
-
-            if (string.IsNullOrEmpty(langCode))
+            // The language actually spoken to Aeneas' TTS engine. It only differs from the requested
+            // one if eSpeak can't handle the requested one and we fall back to another language
+            // (possibly along with an orthography conversion). Applying a manual timings file involves
+            // no TTS at all, so we leave it as requested and never ask eSpeak anything.
+            string langCode = requestedLangCode;
+            if (!usingManualTimings)
             {
-                // FYI: The error message is expected to be in stdError with an empty stdOut, but I included both just in case.
-                Logger.WriteEvent("AudioSegmentationApi.GetAeneasTimings(): eSpeak error.");
-                ErrorReport.ReportNonFatalMessageWithStackTrace(
-                    $"eSpeak error: {stdOut}\n{stdErr}"
+                // When using TTS overrides, there's no Aeneas error message that tells us if the language is unsupported.
+                // Therefore, we explicitly test if the language is supported by the dependency (eSpeak) before getting started.
+                string stdOut;
+                string stdErr;
+                langCode = GetBestSupportedLanguage(requestedLangCode, out stdOut, out stdErr);
+
+                if (string.IsNullOrEmpty(langCode))
+                {
+                    // FYI: The error message is expected to be in stdError with an empty stdOut, but I included both just in case.
+                    Logger.WriteEvent("AudioSegmentationApi.GetAeneasTimings(): eSpeak error.");
+                    ErrorReport.ReportNonFatalMessageWithStackTrace(
+                        $"eSpeak error: {stdOut}\n{stdErr}"
+                    );
+                    return null;
+                }
+                Logger.WriteMinorEvent(
+                    $"AudioSegmentationApi.GetAeneasTimings(): Attempting to segment with langCode={langCode}"
                 );
-                return null;
             }
-            Logger.WriteMinorEvent(
-                $"AudioSegmentationApi.GetAeneasTimings(): Attempting to segment with langCode={langCode}"
-            );
 
             string textFragmentsFilename = Path.Combine(
                 directoryName,
@@ -478,9 +491,7 @@ namespace Bloom.web.controllers
             List<Tuple<string, string>> timingStartEndRangeList = null;
             try
             {
-                RobustFile.WriteAllLines(textFragmentsFilename, fragmentList);
-
-                if (!string.IsNullOrEmpty(requestParameters.manualTimingsPath))
+                if (usingManualTimings)
                 {
                     response.timingsFilePath = Path.Combine(
                         directoryName,
@@ -503,6 +514,8 @@ namespace Bloom.web.controllers
                 }
                 else
                 {
+                    // Aeneas reads the text it is aligning from this file.
+                    RobustFile.WriteAllLines(textFragmentsFilename, fragmentList);
                     response.timingsFilePath = Path.Combine(
                         directoryName,
                         $"{requestParameters.audioFilenameBase}_timings.txt"
--- a/src/BloomTests/web/controllers/AudioSegmentationApiTests.cs
+++ b/src/BloomTests/web/controllers/AudioSegmentationApiTests.cs
@@ -3,7 +3,10 @@ using System.Collections.Generic;
 using System.IO;
 using System.Linq;
 using System.Text;
+using Bloom.Book;
 using Bloom.web.controllers;
+using BloomTemp;
+using Moq;
 using NUnit.Framework;
 using Assert = NUnit.Framework.Assert;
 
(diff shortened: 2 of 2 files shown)


JohnThomson:
(Claude Opus 5) Consulted Devin on 2026-08-05 21:07 UTC up to commit `d1f5ad0a90bb93cec485f603e9fded1112d3be94`.

Devin's review of this commit is complete: **no Bugs and no Investigate flags**. It raised 2 Informational items, neither of which is an action item — both are confirmations of the reasoning behind the change:

- It notes the manual path no longer writes `<base>_fragments.txt`, while the cleanup block still calls `Delete` on it. Deleting a file that isn't there is a no-op (and the call is inside a try/catch), so this is harmless; the only effect is that a stale fragments file left behind by a crashed Aeneas run would also get cleaned up here. Left as-is rather than churn the diff for a cosmetic guard.
- It independently checked that skipping the orthography conversion on the manual path is behavior-neutral, reaching the same conclusion as the local review: on that path the fragment list is only used for `Count()`, and the conversion is a 1:1 per-item mapping that preserves the count.

src/BloomExe/web/controllers/AudioSegmentationApi.cs
```diff
@@ -363,10 +364,17 @@ internal AutoSegmentResponse GetAeneasTimings(AutoSegmentRequest requestParamete
...
+            var usingManualTimings = !string.IsNullOrEmpty(requestParameters.manualTimingsPath);
+
             // The client was supposed to validate this already, but double-check in case something strange happened.
             string directoryName = GetAudioDirectory();
 
-            if (!Platform.IsLinux)
+            // Aeneas runs under Python, which can't cope with a UNC path, so we refuse up front rather
+            // than fail obscurely later. That's a limitation of the forced-alignment tooling, though, so
+            // it doesn't apply when we're only reading times out of a file the user chose.
+            if (!usingManualTimings && !Platform.IsLinux)
```
JohnThomson, line 377: [Devin] **Investigate**: PR description says the drive-letter guard was left alone, but the diff changes it

The description's "Not included" section stated the UNC/drive-letter refusal was deliberately left in place for manual timings, yet the code now skips it when `usingManualTimings` (and a new test asserts the new behavior). The change looks intentional and harmless (the manual path only reads a file), but the description should be updated so reviewers/QA know the network-share behavior also changed.

src/BloomTests/web/controllers/AudioSegmentationApiTests.cs
```diff
@@ -269,5 +272,144 @@ public void TestParseTimingFileTXT_WithEmptyInput_ReturnsEmptyList()
...
+        /// letter, so Bloom refuses forced alignment outright for one (BL-9959) -- typically a
+        /// collection on a network share. That is a limitation of the forced-alignment tooling, so it
+        /// must not block "Apply Timings File...", which only reads numbers out of a file.
+        /// </summary>
+        [Test]
+        [Platform(
+            Exclude = "Linux",
+            Reason = "The drive-letter guard being tested only runs on Windows"
+        )]
+        public void GetAeneasTimings_ManualTimingsOnNonDriveLetterPath_AppliesThemAnyway()
```
JohnThomson, line 334: [Devin] **Investigate**: Network-path test depends on Windows name-resolution timing

`GetAeneasTimings_ManualTimingsOnNetworkPath_NotBlockedByDriveLetterGuard` relies on `\nosuchserver\share\somebook` failing name resolution and on `GetFileNameToSegment` returning empty rather than throwing for a nonexistent UNC directory. On CI machines with unusual DNS/WINS setups this could be slow (the test itself notes ~1s) or could surface a different exception path, making the test flaky. Worth confirming on the build agents.

src/BloomExe/web/controllers/AudioSegmentationApi.cs
JohnThomson: (Claude Opus 5) Fixed in `98895a2` — well, in the description rather than the code. Good catch: the diff was right and the description was stale.

To be explicit about what changed for reviewers and QA: applying a timings file is now also allowed when the collection lives somewhere that isn't under a drive letter (a network share). That guard exists because Aeneas runs under Python, which can't handle such paths — the same class of Aeneas-only restriction as the eSpeak probe this PR removes — so it has no business blocking a code path that just reads numbers out of a file. The PR description now has an "Also fixed: network-share collections" section instead of the "Not included" one.

src/BloomTests/web/controllers/AudioSegmentationApiTests.cs
JohnThomson: (Claude Opus 5) Fixed in `98895a2`. Fair point — the test was waiting on Windows failing to resolve a made-up server name, which is both slow and not something to rely on across build agents.

The guard actually trips on any path that starts with a backslash instead of a drive letter, so the test never needed a real network path to exercise it. It now uses the Win32 extended-length form of a genuine local temp folder, which removes the network dependency completely (the file's tests went from 2s to 240ms) and lets the test assert something stronger: instead of "it failed, but for a different reason than the guard", the manual path now runs all the way through and applies the timings. Re-verified that it still fails if the production guard change is reverted.

JohnThomson:
(Claude Opus 5) Consulted Devin on 2026-08-05 21:45 UTC up to commit `98895a2b261c72c70e371bd578734066abb72b12`.

**No Bugs.** 6 Informational-level items, 4 of them new since the last round, none needing action:

- It independently traced the callers of `GetAeneasTimings` to confirm that skipping the drive-letter guard is safe — nothing downstream of it shells out with the collection path (the old ffmpeg-based `ExtractAudioSegments` is behind `#if OBSOLETEAPI`). Its conclusion: a network-share collection genuinely can apply a timings file now.
- It notes that the test's extended-length local path exercises the guard through the Win32 `\\?\` namespace rather than a real share, so if it ever misbehaves on a build agent that substitution is the first thing to suspect — while calling the trade-off reasonable and documented. Agreed on both counts.
- The other two are the same fragments-file-delete and orthography-conversion confirmations from the earlier round.

Devin also re-raised its two earlier Investigate flags, but both are **stale** — they describe the state before `98895a2`. The PR description no longer has the "Not included" section it objected to (it now has "Also fixed: network-share collections"), and the test it called flaky no longer exists: `nosuchserver` and the old test name are both gone from HEAD. Both already have replied-to, resolved threads on this PR, so they are not re-posted here.

src/BloomExe/web/controllers/AudioSegmentationApi.cs
```diff
@@ -478,9 +491,7 @@ internal AutoSegmentResponse GetAeneasTimings(AutoSegmentRequest requestParamete
             List<Tuple<string, string>> timingStartEndRangeList = null;
             try
             {
-                RobustFile.WriteAllLines(textFragmentsFilename, fragmentList);
-
-                if (!string.IsNullOrEmpty(requestParameters.manualTimingsPath))
+                if (usingManualTimings)
                 {
                     response.timingsFilePath = Path.Combine(
```
JohnThomson, line 496: [Devin] **Investigate**: Manual timings still resolves the response path via Path.Combine with an absolute path

`response.timingsFilePath = Path.Combine(directoryName, requestParameters.manualTimingsPath)` relies on `Path.Combine` discarding `directoryName` when the second argument is rooted. This is pre-existing, but now that manual timings can come from a collection on a network share (the drive-letter guard was lifted), it is worth confirming the front end always sends an absolute path; a relative one would silently be resolved inside the book's audio folder, inconsistent with the `ReadAllLines(requestParameters.manualTimingsPath)` on the next line which resolves relative to the process working directory.

JohnThomson: (Claude Opus 5) Assessed, not changed — this cannot bite today, and the field it concerns is dead.

Two things make it safe. First, the path is always absolute: the only way to reach this code is the "Apply timings file..." menu item, which gets its path from the OS file picker (`fileIO/chooseFile` in `applyTimingsFileAsync`) and passes that straight through, so `Path.Combine` always just returns the second argument and the two lines agree. Second, `timingsFilePath` on the response is never read — the TypeScript side of this call takes only `allEndTimesString` out of the reply (`processAutoSegmentResponse`). So even if the two lines did disagree, nothing would observe it.

You are right that the inconsistency is real in principle, and right that it is pre-existing. I have left it alone rather than tidy it, on two grounds: the repo prefers failing loudly over defending against inputs we control (both ends of this API are ours), and adding a change to unread code would mean another full review cycle for no behavioural gain. Worth revisiting if this response field ever acquires a consumer.

JohnThomson:
(Claude Opus 5) Consulted Devin on 2026-08-05 23:45 UTC up to commit `1f60f68c99d712930979106bb6031869f8e2e734`.

**No Bugs.** One new Investigate flag, now posted, answered and resolved: it spotted that the manual-timings branch builds the response path with `Path.Combine(directoryName, manualTimingsPath)` while the very next line reads the file from `manualTimingsPath` directly, which would disagree if the path were ever relative. Real in principle, but it cannot happen here — the path always comes from the OS file picker, and that response field is never read by the front end. Left as-is with the reasoning recorded on the thread.

The other two Investigate flags it raised are the same stale pair as last round, describing the state before `98895a2`; both already have answered, resolved threads and are not re-posted.

Four Informational items, none actionable — the fragments-file delete and orthography-conversion confirmations from earlier rounds, plus its own verification that the drive-letter skip is safe and a note that the extended-length path in the test exercises the guard through a slightly different namespace than a real share would.

For the record, this round also fixed a problem Devin did not find and neither did the local review: the drive-letter test added in `36cf084` was itself breaking 47 unrelated spreadsheet-import tests, because it called `BookSelection.SelectBook()`, which writes and saves `Settings.Default.CurrentBookPath`. Found by bisection against master and fixed in `1f60f68`; the full C# suite is now 3024 passed / 0 failed, which is master's count plus the two tests this PR adds.

andrew-polk (approved):
> @andrew-polk reviewed 2 files and all commit messages, and resolved 3 discussions.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8156)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on JohnThomson).

PR #8157, opened 2026-08-05
Subscription badge click no longer makes the toolbox jump (BL-16533)

andrew-polk:
Clicking a tool's subscription badge in the Edit tab toolbox made the closed tools and "More..." briefly jump upward. The badge sits inside the tool header's MUI `AccordionSummary`, whose root toggles the tool open/closed on click, and the badge's handler never stopped the click from bubbling. So clicking the active tool's badge also collapsed that tool; `ToolboxRoot`'s legacy-sync effect then saw nothing expanded, read the legacy toolbox's (unchanged) current tool and re-expanded it. Hence the brief jump. Without a subscription the dialog covered the flash, which is why only subscribers reported it. Clicking a *closed* tool's badge had the mirror problem: it expanded that tool.

Three commits:

1. **The fix** — the badge's `onClick` stops propagation before anything else, so with a subscription clicking the badge does nothing at all, and without one it shows only the dialog.
2. **The same flash by another route** — clicking the header of the tool that is already open produced the identical collapse-then-re-expand. The toolbox always has an active tool, so there was never a state to land in; the Accordion's `onChange` now ignores the collapse instead of undoing it after the fact.
3. **An operator-precedence fix** in `showRequiresSubscriptionDialogInAnyView`, noticed on the way past: `a?.b ?? document !== document` parsed as `a?.b ?? (document !== document)` because `!==` binds tighter than `??`, so the test asked "is there a page iframe body?" instead of "is it in another document?". No behavior change for any current caller — they are either in the toolbox iframe (both readings say yes) or in a top-level publish screen with no `#page` in `parent.document` (both say no, so the `else if`/`else` branches those screens rely on still run). It would only have misrouted a future caller running inside the page iframe itself.

**Testing:** `pnpm typecheck` passed; `pnpm lint` reports 0 errors (only the repo's pre-existing warnings); `pnpm vitest run bookEdit/toolbox/ToolboxRoot.spec.tsx` — 2 passed; full front-end Vitest suite run in preflight. No new unit test: at the developer's direction, these are one-line event-propagation changes with no logic to regress and are quick to verify by hand.

**To verify by hand** (subscription active, Music/Canvas/Motion open): clicking the badge should do nothing and not move anything; clicking the open tool's header should also do nothing, with no jump; without a subscription the dialog should still appear from both the tool header and the "More..." list; and the badges on the PDF print-shop-ready and bulk BloomPub publish screens should still open their dialog — that is the path commit 3 touches.

Ref: https://issues.bloomlibrary.org/youtrack/issue/BL-16533

[Devin review](https://devinreview.com/BloomBooks/BloomDesktop/pull/8157)

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8157)

### DIFF
--- a/src/BloomBrowserUI/bookEdit/toolbox/ToolboxRoot.tsx
+++ b/src/BloomBrowserUI/bookEdit/toolbox/ToolboxRoot.tsx
@@ -891,10 +891,23 @@ export const ToolboxRoot: React.FunctionComponent = () => {
                             `}
                             disableGutters
                             expanded={expandedSectionId === section.id}
-                            onChange={(_event, expanded) => {
-                                setActiveSection(
-                                    expanded ? section.id : undefined,
-                                );
+                            // MUI reports the state the accordion is heading TO, not the state
+                            // it is in: it calls onChange(event, !expanded). So clicking a
+                            // closed tool's header arrives here as true, and clicking the open
+                            // one's arrives as false. (MUI's own name for the parameter is just
+                            // "expanded", which reads like "this one is the active tool" and is
+                            // the opposite of what it means -- hence the name used here.)
+                            onChange={(_event, willBeExpanded) => {
+                                if (!willBeExpanded) {
+                                    // Clicking the open tool's header can't close it: the
+                                    // toolbox always has an active tool, and the effect that
+                                    // syncs us with the legacy toolbox re-expands whatever
+                                    // legacy still thinks is current. Honoring the collapse
+                                    // therefore only produced a flash, in which the tools
+                                    // below jumped up and back down. (BL-16533)
+                                    return;
+                                }
+                                setActiveSection(section.id);
                             }}
                         >
                             <AccordionSummary
--- a/src/BloomBrowserUI/react_components/requiresSubscription.tsx
+++ b/src/BloomBrowserUI/react_components/requiresSubscription.tsx
@@ -156,7 +156,13 @@ export const SubscriptionBadgeWithTooltipAndDialog: React.FunctionComponent<{
                     cursor: ${featureStatus?.enabled ? "default" : "pointer"};
                 `}
                 src={badgeUrl}
-                onClick={() => {
+                onClick={(event) => {
+                    // Clicking the badge must do nothing beyond (possibly) showing the
+                    // dialog. In the toolbox, the badge sits inside the accordion header,
+                    // which toggles the tool open/closed when clicked, so letting the click
+                    // bubble collapsed the active tool (and the toolbox then immediately
+                    // re-expanded it), making everything below it jump. (BL-16533)
+                    event.stopPropagation();
                     if (!featureStatus?.enabled) {
                         showRequiresSubscriptionDialogInAnyView(
                             props.featureName,
@@ -457,7 +463,13 @@ export function showRequiresSubscriptionDialogInEditView(featureName: string) {
 }
 
 export function showRequiresSubscriptionDialogInAnyView(featureName: string) {
-    if (getPageIframeBody()?.ownerDocument ?? document !== document) {
+    // The parentheses matter: !== binds tighter than ??, so without them this asked
+    // "is there a page iframe body?" instead of "is it in some other document than
+    // mine?". Harmless for today's callers -- they are either in the toolbox iframe
+    // (where both readings say yes) or in a top-level window like a publish screen
+    // (where there is no page iframe, so both say no) -- but it would have quietly
+    // sent a future caller inside the page iframe itself down the wrong branch.
+    if ((getPageIframeBody()?.ownerDocument ?? document) !== document) {
         // We're in edit mode, but executing from the toolbox. We want the dialog to
         // show in the edit view, where there is room for it, so we have to
         // use a function in that iframe's code to show it. (It's not enough


andrew-polk:
[Claude Opus 5] Consulted Devin on 2026-08-05 up to commit `4f364c847a7d427043d629c313e4cbbba119a8c1`.

Clean: no Bugs and no Investigate flags. It raised 3 Informational observations, none of which needs action — two of them (that the operator-precedence change is behavior-preserving for both current call sites, and that swallowing the badge click has no effect in the "More..." settings list) independently reach the same conclusions the commit messages do. The third notes that ignoring the collapse means the open tool can't be closed from the keyboard either, and asks whether any product expectation allows all tools to be collapsed at once; pre-6.5 the jQuery accordion set no `collapsible` option, so its default of `collapsible: false` meant the active panel couldn't be closed then either. That is flagged for the developer to confirm.

CI (`pr-automation`) passed. The full front-end Vitest suite and typecheck/lint were run locally.

andrew-polk:
[Claude Opus 5] Developer decision on the one Devin item that was left for a human: **leave as is.**

Devin's third Informational observation asked whether any product expectation allows all tools in the Edit toolbox to be collapsed at once, given that this branch makes the toolbox ignore an attempt to close the tool that is already open. Andrew confirmed the current behavior is what we want. It also matches Bloom 6.4 and earlier: the jQuery accordion never set `collapsible`, and jQuery UI's default of `collapsible: false` meant the open panel could not be closed there either. Making "no tool open" a genuine state would be a real design change (the legacy toolbox has no representation for "no current tool"), not a tidy-up, so it is not being done here.

The other two Informational items needed no decision — they independently confirmed the reasoning already in the commit messages. Devin's review of this HEAD found no Bugs and no Investigate flags, so no review threads were opened and none are outstanding.

JohnThomson (changes requested):
> @JohnThomson reviewed 2 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8157)* status: all files reviewed, 1 unresolved discussion (waiting on andrew-polk).

___
*[`src/BloomBrowserUI/bookEdit/toolbox/ToolboxRoot.tsx` line 895 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8157#-OzMUaZQDnz1GYlCkZqz:-OzMUaZQDnz1GYlCkZr-:b6u01sg) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/4f364c847a7d427043d629c313e4cbbba119a8c1/src/BloomBrowserUI/bookEdit/toolbox/ToolboxRoot.tsx#L895)):*
> ```XML
>                             expanded={expandedSectionId === section.id}
>                             onChange={(_event, expanded) => {
>                                 if (!expanded) {
> ```

I don't see how testing could have missed this, but maybe something is badly named? I would expect that 'expanded' means the same as "this is the active tool", in which case, surely we want to do nothing when the one clicked IS expanded rather than when it is not.

andrew-polk (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8157)* status: 1 of 2 files reviewed, 1 unresolved discussion (waiting on JohnThomson).

___
*[`src/BloomBrowserUI/bookEdit/toolbox/ToolboxRoot.tsx` line 895 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8157#-OzMUaZQDnz1GYlCkZqz:-OzMqDmU3VE2Pnco2fln:b-j0ili0) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/4f364c847a7d427043d629c313e4cbbba119a8c1/src/BloomBrowserUI/bookEdit/toolbox/ToolboxRoot.tsx#L895)):*

I don't see how testing could have missed this, but maybe something is badly named? I would expect that 'expanded' means the same as "this is the active tool", in which case, surely we want to do nothing when the one clicked IS expanded rather than when it is not.
</blockquote></details>

[Claude Opus 5] Good catch on the naming — that is exactly the trap here, and the logic is right.

MUI's `Accordion` reports the state the accordion is heading **to**, not the state it is in: `Accordion.js` calls `onChange(event, !expanded)`, where its own `expanded` is the current state. So clicking a closed tool's header arrives here as `true` (it is about to open) and we call `setActiveSection`; clicking the already-open tool's header arrives as `false` (it is about to collapse) and we return early, which is the flash this card is about.

Two things confirm that reading. Under the other one, clicking any closed tool's header would return early and tools could never be switched at all — and the line this replaced, `setActiveSection(expanded ? section.id : undefined)`, relied on the identical semantics, so 6.5 alphas would never have switched tools either.

But your instinct was right that something is badly named: MUI's own parameter name is just `expanded`, which reads as "this one is the active tool" and is the opposite of what it means. I have renamed it to `willBeExpanded` and documented the semantics above the handler, so the next reader does not have to work this out — 18940eda07.

JohnThomson (approved):
> @JohnThomson reviewed 1 file and all commit messages, and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8157)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on andrew-polk).

PR #8158, opened 2026-08-05
Skip the front-end pre-commit checks when no front-end file is staged

andrew-polk:
The front-end pre-commit hook ran `pretty-quick`, `lint-staged` and a whole-project
typecheck unconditionally. That meant a worktree which had never had `pnpm install` run in
it could not commit **at all** — the hook died on `pretty-quick: command not found` even for
a commit touching only a GitHub workflow or a markdown file.

That is a real scenario rather than a hypothetical: worktrees created for CI / tooling /
docs work never need the front-end dependencies, and installing them purely to satisfy a
hook that has nothing to look at is pure overhead. The escape hatch developers reach for,
`--no-verify`, skips the C# checks too — so the status quo quietly pushes people toward
running *fewer* checks than they should.

## What changed

- The front-end block now runs only when the commit stages a file those tools could act on.
  The match is on **extension**, not merely the `src/BloomBrowserUI/` directory, so a staged
  image, a prettier-ignored `README`, or the hook script itself doesn't trigger a
  whole-project typecheck.
- When front-end files **are** staged and `node_modules` is missing, the hook fails loudly
  and names the fix, rather than skipping. Silently skipping in that case is exactly how an
  unchecked front-end commit would slip through.
- The C# checks are untouched and still run on every commit; they need only dotnet. They
  dedent one level because the front-end block now returns to the repo root itself.
- `.githooks/README.md` gains a short section on what the vite-plus hook actually runs —
  `package.json` already points at that README "for how hooks are installed and what else
  they run", and it didn't cover it. It also draws the distinction the README's
  never-skip-silently theme otherwise makes this change look like it violates: the
  dispatcher's loud failure is for "a hook system was configured but nothing ran"; this skip
  is for "these tools have nothing staged to look at".

## Verification

Both paths were exercised for real, not just by inspection:

- Both commits on this branch were made with the new hook active and **no** `--no-verify`:
  the front-end block skipped, and all four C# checks ran and passed (csharpier included).
- Staging a throwaway `.ts` file made the hook fail with the intended "run `pnpm install`"
  message, confirming the guard is not simply always-skipping. The probe file was removed.
- The script parses under `dash` as well as `bash`, checked against the committed (LF) blob,
  so it is not relying on bash-only syntax.

No tracker card: this is tooling work, which `AGENTS.md` notes is a normal case for going
without one.

[Devin review](https://devinreview.com/BloomBooks/BloomDesktop/pull/8158)

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8158)

### DIFF
--- a/src/BloomBrowserUI/.vite-hooks/pre-commit
+++ b/src/BloomBrowserUI/.vite-hooks/pre-commit
@@ -12,20 +12,178 @@
 # vite/rolldown/oxc toolchain instead of this project's pinned eslint/prettier/tsgo,
 # and would not match how we lint and type-check everywhere else. Instead we invoke
 # the project's own tools directly.
+#
+# The two groups of checks below have deliberately different reach, which is the thing
+# to understand before editing either:
+#
+#   formatting      repo-wide, and runs on EVERY commit
+#   lint+typecheck  front-end only, and runs only when a front-end source file is staged
+#
+# What they have in common: NOTHING here is skippable. A check that cannot run blocks the
+# commit. The `pnpm dlx` fallback exists so that a worktree which never ran `pnpm install`
+# can still commit workflow, docs and C# changes -- it removes the need for the install, it
+# does not make the checks optional. (Only formatting has such a fallback: eslint and tsgo
+# need the project's whole dependency graph, not one tool, so there is nothing to fall back
+# to for them.)
 
 # The dispatcher runs us from the repo root; put the front-end's node_modules/.bin
 # on PATH (as an absolute path, before we cd) so its bins resolve.
-export PATH="$(pwd)/src/BloomBrowserUI/node_modules/.bin:$PATH"
+repo_root=$(pwd)
+export PATH="$repo_root/src/BloomBrowserUI/node_modules/.bin:$PATH"
 
-# JavaScript/TypeScript checks (run from the front-end project)
+# ---------------------------------------------------------------------------
+# Formatting -- every commit, whatever it touches
+# ---------------------------------------------------------------------------
+# `pretty-quick --staged` is NOT scoped to the directory it runs in: it resolves the
+# git root and formats every staged file prettier understands, anywhere in the tree,
+# filtered by .prettierignore. That is why it runs unconditionally. Gating it on
+# "was a front-end file staged" silently stopped formatting ~350 tracked files --
+# .github/workflows/*.yml, scripts/*.mjs, DistFiles/**/*.json, src/content, and so on.
+# The repo-root .prettierignore is the tell: everything it lists lives outside the
+# front-end folder, which would be pointless if prettier only ever looked inside.
+#
+# We still cd into src/BloomBrowserUI to run it, because that is what applies BOTH
+# ignore files -- the repo-root one and the front-end one that keeps prettier off
+# vendored third-party code (lib, Readium, modified_libraries, bookEdit/iconfont,
+# font-awesome). Running it from the repo root would reformat all of that.
+
+# Read a pinned version out of the front-end package.json, so the dlx fallback below
+# can never drift from the prettier the project actually uses -- a different prettier
+# would reformat files differently and produce churn. Prints nothing if not found;
+# note this must NOT exit on failure, because a bare `exit` inside $( ) would only
+# leave the subshell and the hook would sail on.
+pinned_version() {
+    sed -n 's/^[[:space:]]*"'"$1"'"[[:space:]]*:[[:space:]]*"\([^"]*\)".*/\1/p' \
+        "$repo_root/src/BloomBrowserUI/package.json" | head -1
+}
+
+# A version is usable by dlx as long as it is something npm can resolve -- an exact pin,
+# but equally "^3.6.2" or ">=3". What it cannot resolve is a pnpm indirection such as
+# "catalog:", "workspace:*" or "link:../x", all of which are recognisable by their colon.
+# Note this deliberately does NOT reject anything else: an earlier revision demanded a
+# leading digit, which would have blocked EVERY commit the day someone wrote a caret range
+# -- an absurd price for a formatting concern.
+version_usable() {
+    case "$1" in
+        "" | *:*) return 1 ;;
+        *) return 0 ;;
+    esac
+}
+
+# Local install first: it is lockfile-verified, costs no network, and so still works
+# offline. `pnpm dlx` is the fallback that lets a worktree which has never run
+# `pnpm install` -- one used only for workflows, docs or C# -- still commit properly
+# formatted code.
+#
+# The versions are read lazily, inside this branch, precisely so that an unreadable
+# package.json can never affect an installed worktree, which does not need them at all.
+#
+# Because we run from src/BloomBrowserUI (see below for why -- the .prettierignore files),
+# dlx also picks up that directory's pnpm-workspace.yaml, and with it minimumReleaseAge.
+# That is a feature: the same 7-day quarantine that protects a normal install protects this
+# path too. It does mean that for a week after someone bumps prettier, the fallback cannot
+# resolve the new version and uninstalled worktrees are blocked until they install or the
+# week elapses. That is the correct end of the trade -- refusing beats reaching for a
+# package too new to have been vetted -- and the failure message below names it.
+run_pretty_quick() {
+    if [ -x "$repo_root/src/BloomBrowserUI/node_modules/.bin/pretty-quick" ]; then
+        pretty-quick "$@"
+        return $?
+    fi
+
+    prettier_version=$(pinned_version prettier)
+    pretty_quick_version=$(pinned_version pretty-quick)
+    if ! version_usable "$prettier_version" || ! version_usable "$pretty_quick_version"; then
+        echo "pre-commit: cannot resolve the prettier/pretty-quick versions in" >&2
+        echo "            src/BloomBrowserUI/package.json ('$prettier_version' /" >&2
+        echo "            '$pretty_quick_version'). If they moved to a pnpm catalog or" >&2
+        echo "            similar indirection, this hook needs teaching to resolve it." >&2
+        return 1
+    fi
+
+    # Say so before doing it. On a machine whose pnpm store is cold this fetches from the
+    # registry, and an unexplained multi-second pause during `git commit` reads as a hang.
+    echo "pre-commit: src/BloomBrowserUI/node_modules not found; running prettier" >&2
+    echo "            $prettier_version via 'pnpm dlx' (first use may download)." >&2
+    pnpm dlx --package="prettier@$prettier_version" \
+             --package="pretty-quick@$pretty_quick_version" pretty-quick "$@"
+}
+
+# No path here lets a commit through unchecked. The point of the dlx fallback is to remove
+# the need for `pnpm install`, NOT to make the checks optional: if neither the local tool nor
+# the fallback can run -- no pnpm, no network, a registry outage, an unresolvable version --
+# this is a setup problem, and a setup problem stops the commit rather than quietly shipping
+# unformatted code. Each branch above has already said which one it hit.
 cd src/BloomBrowserUI || exit 1
-pretty-quick --staged || exit 1
-# --no-stash keeps linting fast and is safe because lint-staged doesn't modify files here.
-lint-staged --no-stash || exit 1
-node scripts/typecheck.js || exit 1
+if ! run_pretty_quick --staged; then
+    echo "pre-commit: staged files could not be formatted, so the commit is blocked." >&2
+    echo "            Install the front-end dependencies ('pnpm install' in" >&2
+    echo "            src/BloomBrowserUI, or './init.sh'). That is also the fix if" >&2
+    echo "            prettier was bumped within the last 7 days: pnpm's minimumReleaseAge" >&2
+    echo "            quarantine deliberately keeps the dlx fallback off a version that" >&2
+    echo "            new. Otherwise 'pnpm dlx' just needs pnpm on PATH and, the first" >&2
+    echo "            time, network access." >&2
+    exit 1
+fi
+cd "$repo_root" || exit 1
+
+# ---------------------------------------------------------------------------
+# Lint and typecheck -- only when front-end sources are staged
+# ---------------------------------------------------------------------------
+# Unlike formatting, these two genuinely are confined to src/BloomBrowserUI: eslint is
+# configured for its .ts/.tsx, and the tsconfig has no "include", so the typecheck's
+# program is that directory. A commit staging none of those files gives them nothing to
+# do, and skipping them is what lets such a commit proceed without `pnpm install`.
+#
+# The extensions are the ones those two tools actually consume: TypeScript, plus JS
+# because the tsconfig sets "allowJs": true. Not .json in general (resolveJsonModule is
+# commented out, so JSON files are not in the program), and not .css/.less/.html/.yml,
+# which neither tool reads -- matching those would only force needless whole-project
+# typechecks. .pug is out for the same reason: no prettier pug plugin is installed and pug
+# is not in the program.
+#
+# Four config files are named explicitly, because they change what the typecheck MEANS
+# without being sources themselves: tsconfig.json (the compiler options and the shape of
+# the program), and package.json / pnpm-lock.yaml / pnpm-workspace.yaml (which versions of
+# the typings everything resolves against). Bumping a @types dependency and breaking the
+# build is an easy thing to do and an annoying one to discover in CI instead of here.
+#
+# Deletions count (hence D alongside ACM): removing a .ts is the change most likely to
+# break another module's imports, so it is exactly when the typecheck is worth running.
+#
+# --no-renames is what makes that cover a file moved OUT of the front-end, which is the
+# same event wearing a disguise. With rename detection on, --name-only prints only the
+# destination, so moving src/BloomBrowserUI/foo.ts to src/content/foo.ts would produce no
+# path under src/BloomBrowserUI and skip the typecheck. Off, the move is reported as its
+# two halves -- a delete of the old path and an add of the new -- and the delete matches.
+#
+# core.quotePath=false plus -z stops git C-quoting a non-ASCII path ("caf\303\251.ts"),
+# whose added surrounding quotes would defeat both anchors of the regex below and so
+# silently skip the checks for that file.
+staged_frontend=$(git -c core.quotePath=false diff --cached --no-renames --name-only -z --diff-filter=ACMD |
+    tr '\0' '\n' |
+    grep -E '^src/BloomBrowserUI/(.*\.(ts|tsx|mts|cts|js|jsx|mjs|cjs)|tsconfig\.json|package\.json|pnpm-lock\.yaml|pnpm-workspace\.yaml)$')
+
+if [ -n "$staged_frontend" ]; then
+    # Front-end sources are staged, so these must run. There is no dlx fallback here:
+    # eslint and tsgo need the project's whole dependency graph, not one tool, so a
+    # missing install has to fail loudly -- silently skipping is how an unchecked
+    # front-end commit would get through.
+    if [ ! -x "$repo_root/src/BloomBrowserUI/node_modules/.bin/lint-staged" ]; then
+        echo "pre-commit: front-end sources are staged, but src/BloomBrowserUI/node_modules" >&2
+        echo "            is missing, so lint and typecheck cannot run. Run 'pnpm install'" >&2
+        echo "            in src/BloomBrowserUI (or './init.sh' for the whole worktree)." >&2
+        exit 1
+    fi
+
+    cd src/BloomBrowserUI || exit 1
+    # --no-stash keeps linting fast and is safe because lint-staged doesn't modify files here.
+    lint-staged --no-stash || exit 1
+    node scripts/typecheck.js || exit 1
+    cd "$repo_root" || exit 1
+fi
 
 # C# checks (run from the repo root)
-cd ../.. || exit 1
 build/check-csharp-robustfile.sh || exit 1
 build/check-csharp-xmlclasses.sh || exit 1
 build/check-csharp-ApplicationExit.sh || exit 1
--- a/.githooks/README.md
+++ b/.githooks/README.md
@@ -36,6 +36,50 @@ ships:
 2. else if a husky hook exists in `.git/hooks/pre-commit` → run that;
 3. else → **fail loudly** with instructions, instead of skipping silently.
 
+## What the vite-plus hook runs
+
+Routing aside, `src/BloomBrowserUI/.vite-hooks/pre-commit` runs three groups of checks,
+and the thing worth knowing before editing it is that they have deliberately different
+reach:
+
+- **Formatting** — `pretty-quick`, on **every commit, whatever it touches**. Despite living
+  in the front-end folder, it is not scoped to it: `pretty-quick --staged` resolves the git
+  root and formats every staged file prettier understands anywhere in the tree, filtered by
+  `.prettierignore`. That is why the repo root has its own `.prettierignore`. The hook still
+  runs it from `src/BloomBrowserUI`, because that is what applies *both* ignore files —
+  including the front-end one that keeps prettier off vendored third-party code.
+- **Lint and typecheck** — `lint-staged` (eslint) and a whole-project typecheck, **only when
+  the commit stages a front-end source file** (`.ts`/`.tsx`/`.mts`/`.cts` and, since the
+  tsconfig sets `allowJs`, `.js`/`.jsx`/`.mjs`/`.cjs` under `src/BloomBrowserUI/`), **or one
+  of the four config files that change what the typecheck means** — `tsconfig.json`, and
+  `package.json` / `pnpm-lock.yaml` / `pnpm-workspace.yaml`, which decide the typings
+  everything resolves against. These two tools genuinely are confined to that directory, so a
+  commit of only workflows, docs or C# gives them nothing to do.
+- **C#** — the `build/check-csharp-*.sh` scripts and `build/run-csharpier.sh`. Every commit;
+  they need only dotnet, never the front-end dependencies.
+
+**Committing without `pnpm install`.** A worktree used only for workflow, docs or C# work can
+commit without ever installing the front-end dependencies. Formatting still happens there: if
+`node_modules` is absent the hook falls back to `pnpm dlx` at the exact prettier and
+pretty-quick versions pinned in `package.json` (read from the file, so the fallback cannot
+drift from what the project uses). An installed worktree always prefers its own
+lockfile-verified binaries, so the normal case needs no network and works offline.
+
+**Nothing here is skippable.** That fallback removes the need for the install; it does not make
+the checks optional. A check that cannot run **blocks the commit**, and says which fix it
+wants:
+
+- Front-end sources staged with no `node_modules` → blocked. There is no fallback for this
+  case and there cannot easily be one: eslint and tsgo need the project's whole dependency
+  graph, not a single tool.
+- Formatting unable to run at all — no `node_modules` *and* no working `pnpm dlx` (no pnpm on
+  PATH, no network on the first use) → blocked. That is a setup problem, and a setup problem
+  should be fixed rather than worked around.
+
+So the only skip in the whole design is the deliberate one: lint and typecheck do not run when
+the commit stages nothing they could look at. That is not the silent skip this dispatcher
+exists to prevent — that one is "a hook system was configured but nothing ran".
+
 ## How to enable it (per clone)
 
 `core.hooksPath` is git config, not a tracked file, so each clone sets it once:


src/BloomBrowserUI/.vite-hooks/pre-commit
```diff
@@ -17,15 +17,51 @@
...
+# two halves -- a delete of the old path and an add of the new -- and the delete matches.
+#
+# core.quotePath=false plus -z stops git C-quoting a non-ASCII path ("caf\303\251.ts"),
+# whose added surrounding quotes would defeat both anchors of the regex below and so
+# silently skip the checks for that file.
+staged_frontend=$(git -c core.quotePath=false diff --cached --no-renames --name-only -z --diff-filter=ACMD |
+    tr '\0' '\n' |
+    grep -E '^src/BloomBrowserUI/.*\.(ts|tsx|mts|cts|js|jsx|mjs|cjs|json|css|less|scss|html|htm|yaml|yml)$')
+
+if [ -n "$staged_frontend" ]; then
```
andrew-polk, line 167: [Devin] **Bug**: Automatic code formatting no longer runs for files outside the browser UI folder

Devin's point, in plain terms: `pretty-quick --staged` was never scoped to the front-end folder. It finds the **git root** and formats every staged file prettier supports anywhere in the repo, filtered only by `.prettierignore`. The old hook ran it on every commit, so staging a workflow YAML, a root JSON file, or `src/content/pageSizes.ts` got that file auto-formatted. With this PR's guard, a commit that touches only files outside `src/BloomBrowserUI/` skips the whole block, so prettier never runs on them and they go in unformatted.

I verified this independently rather than taking it on faith — a scratch repo mirroring this layout, with the real `pretty-quick` installed: running it from `src/BloomBrowserUI` with only `.github/workflows/x.yml` staged reports `⛔️ Check failed: .github/workflows/x.yml`. The repo-root `.prettierignore` corroborates it: its entries (`csharpier.json`, `*.xlf`, `src/BloomVisualRegressionTests/collections/`) are all *outside* the front-end folder and would be pointless if prettier only ever looked inside it. Roughly 350 tracked files are in scope.

**This one is real and is deliberately left open for the developer**, because the fix is a genuine trade-off rather than a typo:

The lint and typecheck steps *are* legitimately front-end-only (confirmed: `tsconfig.json` has no `include`, and no front-end source imports escape the directory). Formatting is the only step with repo-wide reach. So the guard could be split — run prettier whenever any staged file is prettier-formattable anywhere in the repo, and gate lint + typecheck on `src/BloomBrowserUI/` paths.

The catch is that this partly undoes the PR's purpose. The commit that motivated it staged `.github/workflows/release-installer.yml` plus a markdown file. Markdown is prettier-ignored, but the YAML is not — so under a correct split guard, that commit needs prettier, which needs `node_modules`, and the hook blocks again in a worktree that never ran `pnpm install`. The deps-free commit would still work for C#-only, shell-only, and markdown-only changes, but not for YAML/JSON/mjs.

Options are laid out for the developer in the preflight report; this thread stays open until that decision comes back.

<sub>Posted by Claude Opus 5 (1M context) via the `devin-review` skill.</sub>

src/BloomBrowserUI/.vite-hooks/pre-commit
```diff
@@ -17,15 +17,51 @@
...
+# break another module's imports, so it is exactly when the typecheck is worth running.
+# The tools re-derive their own file lists, so a deletion-only commit simply gives
+# prettier and eslint nothing to do while the typecheck still runs.
+#
+# --no-renames is what makes that cover a file moved OUT of the front-end, which is the
+# same event wearing a disguise. With rename detection on, --name-only prints only the
+# destination, so moving src/BloomBrowserUI/foo.ts to src/content/foo.ts would produce no
+# path under src/BloomBrowserUI and skip the typecheck. Off, the move is reported as its
+# two halves -- a delete of the old path and an add of the new -- and the delete matches.
+#
```
andrew-polk, line 159: [Devin] **Investigate**: Renames out of the front-end folder skip the typecheck

Moving `src/BloomBrowserUI/foo.ts` to `src/content/foo.ts` is, as far as every module that imports it is concerned, the same event as deleting it. But with git's rename detection on, `--name-only` prints only the *destination* path, so nothing matched the `^src/BloomBrowserUI/` anchor and the whole-project typecheck was skipped — for exactly the change most likely to need it.

<sub>Posted by Claude Opus 5 (1M context) via the `devin-review` skill.</sub>

src/BloomBrowserUI/.vite-hooks/pre-commit
```diff
@@ -17,15 +17,51 @@
...
-lint-staged --no-stash || exit 1
-node scripts/typecheck.js || exit 1
+# Which staged paths the front-end tools could actually act on: prettier-formattable or
+# lint/typecheck-able sources under the front-end project. Matching on extension rather
+# than just the directory keeps a staged image, README (prettier-ignored at the repo root)
+# or this hook itself from triggering a full typecheck. A commit that stages none of them
+# -- a workflow, a doc, C# only -- therefore does not need node_modules to exist at all,
+# which is what lets a worktree used only for such work commit without `pnpm install`.
+#
+# Deletions count (hence D alongside ACM): removing a .ts is the change most likely to
```
andrew-polk, line 151: [Devin] **Investigate**: Deleting a front-end source file no longer triggers the typecheck

`--diff-filter=ACMR` leaves out deletions, so a commit that only deletes a `.ts`/`.tsx` under `src/BloomBrowserUI/` skipped the whole-project typecheck. Deleting a file is one of the likeliest ways to break another module's imports, so the type error would only have surfaced later in CI.

<sub>Posted by Claude Opus 5 (1M context) via the `devin-review` skill.</sub>

src/BloomBrowserUI/.vite-hooks/pre-commit
```diff
@@ -17,15 +17,51 @@
...
+# The tools re-derive their own file lists, so a deletion-only commit simply gives
+# prettier and eslint nothing to do while the typecheck still runs.
+#
+# --no-renames is what makes that cover a file moved OUT of the front-end, which is the
+# same event wearing a disguise. With rename detection on, --name-only prints only the
+# destination, so moving src/BloomBrowserUI/foo.ts to src/content/foo.ts would produce no
+# path under src/BloomBrowserUI and skip the typecheck. Off, the move is reported as its
+# two halves -- a delete of the old path and an add of the new -- and the delete matches.
+#
+# core.quotePath=false plus -z stops git C-quoting a non-ASCII path ("caf\303\251.ts"),
```
andrew-polk, line 160: [Devin] **Investigate**: Extension list omits some prettier-relevant front-end extensions

The guard matched `ts|tsx|js|jsx|mjs|cjs|json|css|less|scss|html|htm|yaml|yml`, but the front-end also carries `.mts`/`.cts` and `.pug` files. Devin's concern is that a staged file of those types would slip past both the formatter and the typecheck.

<sub>Posted by Claude Opus 5 (1M context) via the `devin-review` skill.</sub>

src/BloomBrowserUI/.vite-hooks/pre-commit
andrew-polk: ✅ **Fixed** in \`21db7196b4\` — this one was a real hole and is now closed.

The filter now passes \`--no-renames\`, which makes git report a move as its two halves (a delete of the old path plus an add of the new) instead of a single rename entry. The delete then matches the \`src/BloomBrowserUI/\` anchor and the typecheck runs. \`R\` drops out of the diff-filter along with it, since rename entries no longer occur.

Verified against real git rather than by reading: staging an actual \`git mv\` of a front-end \`.ts\` out of the directory produced **no** match under the old filter and the old path under the new one.

<sub>Posted by Claude Opus 5 (1M context) via the \`devin-review\` skill.</sub>

src/BloomBrowserUI/.vite-hooks/pre-commit
andrew-polk: ✅ **Fixed** in \`6608beedae\`, before this flag was re-raised — the filter is now \`--diff-filter=ACMD\` (\`D\` added; \`R\` later dropped when \`--no-renames\` went in).

Worth noting the flag as written is stale: it quotes \`--diff-filter=ACMR\`, which stopped being the code at that commit. Devin re-emitted the original analysis verbatim against a HEAD that no longer matched it.

The tools re-derive their own file lists, so a deletion-only commit simply gives prettier and eslint nothing to do while the typecheck still runs. Verified by staging a real deletion: no match under the old filter, the deleted path under the new one.

<sub>Posted by Claude Opus 5 (1M context) via the \`devin-review\` skill.</sub>

src/BloomBrowserUI/.vite-hooks/pre-commit
andrew-polk: **Half fixed, half deliberate.**

✅ \`.mts\`/\`.cts\` **added** in \`6608beedae\`. This part was right and mattered: \`src/BloomBrowserUI/vite.config.mts\` is a real tracked file, and since \`tsconfig.json\` has no \`include\` and excludes only \`vite.config.ts\` (not \`.mts\`), it *is* in the typecheck program. Staging it alone would have skipped both prettier and the typecheck. (As with the sibling flag, the text above is stale — it was re-emitted verbatim after the fix landed.)

❌ \`.pug\` **deliberately left out.** No prettier pug plugin is installed here (nothing matching \`prettier-plugin-*\` is in \`package.json\`), so prettier cannot format pug at all; and pug is not in the typecheck program. Matching it would buy nothing and would force a needless whole-project typecheck on every template edit.

\`.md\` is correctly absent — it is prettier-ignored in both the repo-root and front-end ignore files, as the flag itself notes.

<sub>Posted by Claude Opus 5 (1M context) via the \`devin-review\` skill.</sub>

andrew-polk:
**Consulted Devin** on 2026-08-05 up to commit `21db7196b4`.

Three rounds ran, because each round of fixes created a new HEAD and re-triggered the review.

- **1 Bug — open, needs a decision.** Automatic formatting no longer runs for staged files outside the browser-UI folder. Real, independently verified, and left open deliberately: [thread](https://github.com/BloomBooks/BloomDesktop/pull/8158#discussion_r3724630464).
- **3 Investigate flags — all handled and resolved.** Renames out of the folder skipping the typecheck (fixed, `21db7196b4`); deletions skipping the typecheck (fixed, `6608beedae`); missing `.mts`/`.cts` (fixed, `6608beedae`) with `.pug` deliberately excluded.
- **4 Informational items** — not mirrored as threads. Of note, one observes that a filename containing a literal newline would defeat the `-z`/`tr` handling; that is pathological and no such path is tracked. Another notes the pipeline's exit status is unused, which is harmless here since the hook does not run under `set -e`.

A caveat for whoever reads this next: on rounds 2 and 3 Devin re-emitted **stale** analyses verbatim — still quoting `--diff-filter=ACMR` and "omits `.mts`/`.cts`" after both had been fixed and pushed — even though the job's `head_sha` matched the new HEAD. Each was checked against the actual pushed file before being refuted, rather than assumed fixed.

CI: the only check on this PR is `pr-automation` (this repo runs no build/test workflow on PRs).

<sub>Posted by Claude Opus 5 (1M context) via the `devin-review` skill.</sub>

src/BloomBrowserUI/.vite-hooks/pre-commit
andrew-polk: ✅ **Fixed** in `17cec64d6c` — this was a real regression and the developer chose to fix it rather than accept it.

Formatting has moved back out of the guard and now runs on **every** commit, exactly as it did before this branch. Repo-wide coverage is restored, and the ~350 files outside `src/BloomBrowserUI/` are formatted again. The guard remains, but now covers only lint and typecheck, which genuinely are confined to the front-end folder; its extension list narrowed to what those two actually consume (TypeScript, plus JS because the tsconfig sets `allowJs`).

That alone would have re-blocked the case this branch exists for, since the motivating commit staged a workflow YAML and prettier does want that file. So when `node_modules` is absent, formatting now runs via `pnpm dlx` at the exact `prettier` and `pretty-quick` versions **read out of `package.json`**, so the fallback cannot drift from the version the project pins. An installed worktree still prefers its own lockfile-verified binaries, keeping the normal path off the network and working offline.

Two details worth recording, since both were ways to get this wrong:

- We fall back to **pretty-quick**, not bare `prettier`. pretty-quick is what re-stages the files it fixed, including the partially-staged case that a naive `prettier --write` + `git add` would botch by staging someone's unrelated unstaged work.
- It is still run from `src/BloomBrowserUI`, because that is what applies **both** `.prettierignore` files. The front-end one is what keeps prettier off vendored third-party code in `lib`, `Readium`, `modified_libraries` and `bookEdit/iconfont`; running from the repo root would have reformatted all of it.

Verified in a worktree with no `node_modules`: a deliberately misformatted `.github/*.yml` was reformatted through the dlx fallback and the **committed** blob was the formatted one, proving re-staging works on that path too.

<sub>Posted by Claude Opus 5 (1M context) via the \`devin-review\` skill.</sub>

src/BloomBrowserUI/.vite-hooks/pre-commit
```diff
@@ -12,20 +12,150 @@
...
+    if [ -x "$repo_root/src/BloomBrowserUI/node_modules/.bin/pretty-quick" ]; then
+        pretty-quick "$@"
+    else
+        pnpm dlx --package="prettier@$prettier_version" \
+                 --package="pretty-quick@$pretty_quick_version" pretty-quick "$@"
+    fi
+}
 
-# JavaScript/TypeScript checks (run from the front-end project)
 cd src/BloomBrowserUI || exit 1
```
andrew-polk, line 117: [Devin] **Bug**: Commits can go through unformatted when the formatting tool cannot run

If `src/BloomBrowserUI/node_modules` is missing *and* the `pnpm dlx` fallback cannot run — no pnpm on PATH, no network, a registry outage — the hook prints a warning and lets the commit through unformatted, rather than blocking.

<sub>Posted by Claude Opus 5 (1M context) via the `devin-review` skill.</sub>

src/BloomBrowserUI/.vite-hooks/pre-commit
```diff
@@ -12,20 +12,150 @@
...
+# commented out, so JSON files are not in the program), and not .css/.less/.html/.yml,
+# which neither tool reads -- matching those would only force needless whole-project
+# typechecks. .pug is out for the same reason: no prettier pug plugin is installed and pug
+# is not in the program.
+#
+# Four config files are named explicitly, because they change what the typecheck MEANS
+# without being sources themselves: tsconfig.json (the compiler options and the shape of
+# the program), and package.json / pnpm-lock.yaml / pnpm-workspace.yaml (which versions of
+# the typings everything resolves against). Bumping a @types dependency and breaking the
+# build is an easy thing to do and an annoying one to discover in CI instead of here.
```
andrew-polk, line 149: [Devin] **Investigate**: Typecheck no longer runs for tsconfig/package.json/dependency changes

Narrowing the guard to source extensions dropped `.json`, which took `tsconfig.json` with it. That file changes the compiler options and the shape of the typecheck's program without being a source file, so editing it alone would have skipped the very check it alters.

<sub>Posted by Claude Opus 5 (1M context) via the `devin-review` skill.</sub>

src/BloomBrowserUI/.vite-hooks/pre-commit
andrew-polk: **Deliberate, and the developer's explicit call** — not an oversight, so recording the reasoning here rather than \"fixing\" it.

Blocking in that situation is precisely the problem this PR exists to remove. A worktree created for CI/tooling/docs work would once again be unable to commit a markdown or workflow change, and the escape hatch people reach for, \`--no-verify\`, skips the **C# checks too** — so a hard failure here pushes developers toward running strictly fewer checks than the warning does.

The trade is deliberately asymmetric, and the hook draws the line by consequence:

- **Unformatted code is cosmetic.** It costs a noisy diff later, and any subsequent commit on an installed machine fixes it automatically.
- **Unlinted, untypechecked code can be broken.** So that case still fails hard — staging front-end sources with no \`node_modules\` blocks, with a message naming the fix.

The warning is loud and on stderr, and this path is reachable only in a worktree that never ran \`pnpm install\` **and** where dlx cannot run. On a normally installed machine the local binary is preferred, so none of this is on the usual path.

One hardening did come out of thinking about this, in \`bb45f45899\`: the version scrape now requires a leading digit. Previously a pnpm \`catalog:\` indirection would have read back as a non-empty string, failed inside dlx, and dropped into *this* warn path — quietly not formatting, which is the one outcome the path must not hide. That now fails loudly instead.

<sub>Posted by Claude Opus 5 (1M context) via the \`devin-review\` skill.</sub>

src/BloomBrowserUI/.vite-hooks/pre-commit
andrew-polk: ✅ **Fixed** in \`bb45f45899\` — a good catch, and a genuine regression I introduced when narrowing the extension list.

Four config files are now named explicitly in the guard, because they change what the typecheck *means* without being sources themselves:

- \`tsconfig.json\` — the compiler options and the shape of the program;
- \`package.json\`, \`pnpm-lock.yaml\`, \`pnpm-workspace.yaml\` — which versions of the typings everything resolves against. Bumping a \`@types\` dependency and breaking the build is an easy thing to do and an annoying one to discover in CI instead of at commit time.

\`.json\` in general stays out on purpose: \`resolveJsonModule\` is commented out in the tsconfig, so ordinary JSON files are genuinely not in the program, and matching them all would force needless whole-project typechecks.

Verified against the real regex — the four config paths and a \`.tsx\` match; \`src/BloomBrowserUI/some/other/data.json\`, a \`.less\` file, and the extensionless hook script do not.

<sub>Posted by Claude Opus 5 (1M context) via the \`devin-review\` skill.</sub>

src/BloomBrowserUI/.vite-hooks/pre-commit
```diff
@@ -12,20 +12,160 @@
...
+}
+
+# A version is usable by dlx as long as it is something npm can resolve -- an exact pin,
+# but equally "^3.6.2" or ">=3". What it cannot resolve is a pnpm indirection such as
+# "catalog:", "workspace:*" or "link:../x", all of which are recognisable by their colon.
+# Note this deliberately does NOT reject anything else: an earlier revision demanded a
+# leading digit, which would have blocked EVERY commit the day someone wrote a caret range
+# -- an absurd price for a formatting concern.
+version_usable() {
+    case "$1" in
```
andrew-polk, line 67: [Devin] **Bug**: Every commit is blocked if the formatter version numbers are written as ranges

A real one, and mine: the version sanity check I added in `bb45f45899` demanded a **leading digit**. The day anyone wrote an ordinary caret range like `^3.6.2` — which `pnpm dlx` resolves perfectly well — the hook would have refused **every commit in the repo**. That is a wildly disproportionate price for a formatting concern, and a worse failure than the pnpm-catalog case the check was guarding against.

✅ **Fixed** in `6c47ca00b9`. Three changes:

- Reject only what dlx genuinely cannot resolve — pnpm's indirections, recognisable by their colon (`catalog:`, `workspace:*`, `link:../x`). Ranges and exact pins pass through. Verified across `3.6.2`, `^3.6.2`, `~3.6`, `>=3`, `catalog:`, `workspace:*`, `link:../x` and the empty string.
- An unresolvable version now **warns and skips formatting**, on the same path as no-pnpm and no-network, instead of aborting the commit. Lint and typecheck — the checks that can actually catch a broken build — still fail hard.
- The versions are read **lazily**, inside the fallback branch, so an unreadable `package.json` cannot affect an installed worktree that never needed them.

<sub>Posted by Claude Opus 5 (1M context) via the `devin-review` skill.</sub>

andrew-polk:
**Consulted Devin** on 2026-08-05 up to commit `6c47ca00b9`.

Six rounds in total, since each round of fixes created a new HEAD and re-triggered the review. Everything Devin raised is now either fixed or documented, and **all seven review threads are resolved**.

Fixed as a result of these rounds: formatting restored to repo-wide (`17cec64d6c`); typecheck now also triggered by `tsconfig.json` / `package.json` / lockfile changes (`bb45f45899`); a version check of mine that would have blocked *every* commit on a caret range (`6c47ca00b9`); plus earlier rounds' renames, deletions and `.mts`/`.cts` gaps.

Documented as deliberate rather than fixed: the formatting fallback [warns rather than blocks](https://github.com/BloomBooks/BloomDesktop/pull/8158#discussion_r3724762937) when neither a local install nor `pnpm dlx` can run. Blocking there would restore the exact problem this PR removes, and unformatted code is cosmetic — whereas lint and typecheck, which can catch a genuinely broken build, still fail hard.

⚠️ **A caveat for whoever reads this next.** Devin's later rounds **accumulate stale findings**: the final round still lists "Every commit is blocked if the formatter version numbers are written as ranges" (fixed in that very HEAD) and "Automatic code formatting no longer runs for files outside the browser UI folder" (fixed three commits earlier), despite the job's `head_sha` matching. Every finding here was checked against the actually-pushed file before being called stale — none was dismissed on the assumption. Do not read that final list as a live defect list.

CI (`pr-automation`, the only check this repo runs on PRs) passes at this commit.

<sub>Posted by Claude Opus 5 (1M context) via the `devin-review` skill.</sub>

src/BloomBrowserUI/.vite-hooks/pre-commit
andrew-polk: ↩️ **Reversed — Devin was right, and the developer has overruled my earlier call.** Fixed in `369a1d12d8`.

I previously closed this as a deliberate trade-off, arguing that unformatted code is cosmetic and that blocking would restore the problem the branch removes. The developer's direction: *\"it can't ever let things through. if we aren't set up, fail. the goal of this was to find a way to make it not need \`pnpm i\` to be set up.\"*

That draws a distinction I had blurred. The dlx fallback exists to remove the need for `pnpm install` — it is not licence to make the checks optional. A worktree where neither the local tool nor the fallback can run is not one with a cosmetic problem; it is one that is **not set up**, and the fix is to set it up rather than ship past it.

So every failure path now blocks, and names which of the two fixes it wants: install the front-end dependencies, or make `pnpm dlx` able to run. The unresolvable-version case blocks on the same path rather than degrading.

Worth noting this costs nothing in the case the branch exists for: a worktree that never ran `pnpm install` still commits workflow, docs and C# changes, because dlx handles the formatting. Blocking now happens only when dlx *itself* cannot run.

Verified with a stub `pnpm` that exits non-zero, in a worktree with no `node_modules`: the hook exits 1 and no commit is created.

<sub>Posted by Claude Opus 5 (1M context) via the \`devin-review\` skill.</sub>

src/BloomBrowserUI/.vite-hooks/pre-commit
```diff
@@ -12,20 +12,178 @@
...
+    pretty_quick_version=$(pinned_version pretty-quick)
+    if ! version_usable "$prettier_version" || ! version_usable "$pretty_quick_version"; then
+        echo "pre-commit: cannot resolve the prettier/pretty-quick versions in" >&2
+        echo "            src/BloomBrowserUI/package.json ('$prettier_version' /" >&2
+        echo "            '$pretty_quick_version'). If they moved to a pnpm catalog or" >&2
+        echo "            similar indirection, this hook needs teaching to resolve it." >&2
+        return 1
+    fi
+
+    # Say so before doing it. On a machine whose pnpm store is cold this fetches from the
```
andrew-polk, line 104: [Devin] **Investigate**: Network-downloaded formatter fallback may diverge from installed toolchain expectations

Two observations about the `pnpm dlx` fallback: that it runs silently, so a first commit in a fresh worktree pauses with no explanation; and that `pnpm dlx` resolves through the config it finds in the cwd, which includes `src/BloomBrowserUI/pnpm-workspace.yaml`'s `minimumReleaseAge: 10080`.

<sub>Posted by Claude Opus 5 (1M context) via the `devin-review` skill.</sub>

andrew-polk: ✅ **Both addressed** in \`74cc9dca13\` — a good catch on the second one especially, which corrected a claim I had made earlier in this PR.

**Silent download — fixed.** The fallback now announces itself before fetching, so the pause is explained rather than looking like a hang.

**minimumReleaseAge — confirmed, and it is a feature, not a defect.** I had previously stated that the 7-day quarantine gave this path no protection, on the reasoning that dlx run from the repo root would not read the front-end's \`pnpm-workspace.yaml\`. That was wrong: the hook deliberately runs from \`src/BloomBrowserUI\` so that **both** \`.prettierignore\` files apply, and dlx therefore picks up that directory's config too. Verified directly — \`pnpm config get minimumReleaseAge\` reports \`10080\` there and \`undefined\` at the repo root.

So the fallback inherits the same quarantine that protects a normal install, which is a **better** supply-chain position than this branch started with. The cost is the one you identify: for a week after someone bumps prettier, the fallback cannot resolve the new version, and since this hook now fails closed, uninstalled worktrees are blocked until they install or the week elapses. That is the correct end of the trade — refusing beats reaching for a package too new to have been vetted — and the block message now names that case explicitly so nobody has to rediscover it.

On the \`AGENTS.md\` \"fail fast, don't silently work around failed dependencies\" point: that is now moot. The warn-and-continue path this flag was weighing has been removed entirely — every failure blocks.

<sub>Posted by Claude Opus 5 (1M context) via the \`devin-review\` skill.</sub>

src/BloomBrowserUI/.vite-hooks/pre-commit
```diff
@@ -12,20 +12,178 @@
...
+# filtered by .prettierignore. That is why it runs unconditionally. Gating it on
+# "was a front-end file staged" silently stopped formatting ~350 tracked files --
+# .github/workflows/*.yml, scripts/*.mjs, DistFiles/**/*.json, src/content, and so on.
+# The repo-root .prettierignore is the tell: everything it lists lives outside the
+# front-end folder, which would be pointless if prettier only ever looked inside.
+#
+# We still cd into src/BloomBrowserUI to run it, because that is what applies BOTH
+# ignore files -- the repo-root one and the front-end one that keeps prettier off
+# vendored third-party code (lib, Readium, modified_libraries, bookEdit/iconfont,
+# font-awesome). Running it from the repo root would reformat all of that.
```
andrew-polk, line 48: [Devin] **Investigate**: Claim that running from `src/BloomBrowserUI` applies both `.prettierignore` files is worth verifying

The comments and commit messages here lean on the claim that running pretty-quick from `src/BloomBrowserUI` applies **both** the repo-root `.prettierignore` and the front-end one — the latter being what keeps prettier off vendored third-party code. Fair challenge: that had been asserted from reading pretty-quick's behaviour, not demonstrated.

<sub>Posted by Claude Opus 5 (1M context) via the `devin-review` skill.</sub>

andrew-polk: ✅ **Verified — the claim holds, and the experiment is worth recording** because it also demonstrates the failure that would follow from getting this wrong.

Method: stage two deliberately misformatted files, one at \`src/BloomBrowserUI/lib/…js\` (\`lib\` is listed in the **front-end** \`.prettierignore\` and *not* in the root one — it is vendored third-party code) and one at \`src/content/…js\` (outside the front-end folder, ignored by neither). Then run the real pinned pretty-quick both ways.

| run from | files it acts on |
|---|---|
| \`src/BloomBrowserUI\` — what the hook does | **1**: \`src/content/…\` only. The vendored \`lib/\` file is skipped. |
| the repo root | **2**: both, including the vendored \`lib/\` file. |

Two things follow, and they are the whole reason the \`cd\` is there:

1. Running from the front-end directory does **not** cost repo-wide reach — it still picked up \`src/content\`, well outside that folder. This is the same property the PR relies on throughout.
2. Running from the repo root **would** reformat vendored third-party code. So the \`cd\` is load-bearing, not incidental tidiness, and this thread is the evidence for anyone who later wonders whether it can be simplified away.

<sub>Posted by Claude Opus 5 (1M context) via the \`devin-review\` skill.</sub>

andrew-polk:
**Consulted Devin** on 2026-08-06 up to commit `74cc9dca13`. Second gauntlet run, after the developer reversed the warn-vs-block decision.

Two new findings this run, both real, both addressed:

- **The dlx fallback ran silently and inherits pnpm's release-age quarantine.** Fixed in `74cc9dca13`, and it corrected a claim I had made earlier: I had said the 7-day `minimumReleaseAge` gave this path no protection. It does — the hook runs from `src/BloomBrowserUI`, so dlx reads that directory's config (`pnpm config get minimumReleaseAge` → `10080` there, `undefined` at the repo root). Better supply-chain position than the branch started with. [Thread](https://github.com/BloomBooks/BloomDesktop/pull/8158#discussion_r3725030045)
- **A challenge to verify the two-`.prettierignore` claim** — fair, since it had been asserted rather than demonstrated. Now demonstrated, and the experiment shows running from the repo root *would* reformat vendored third-party code. [Thread](https://github.com/BloomBooks/BloomDesktop/pull/8158#discussion_r3725041558)

Also this run: the previously-documented "formatting degrades to a warning" finding is now **obsolete** — the developer's call was that nothing may pass unchecked, so every failure path blocks. The reversal is recorded on its original thread rather than left as a silent flip.

**All review threads are resolved.** CI (`pr-automation`) passes at this commit; the PR is mergeable and remains a draft.

⚠️ Same caveat as the first run, now more pronounced: Devin's rounds **accumulate stale findings**. This final round still lists "Every commit is blocked if the formatter version numbers are written as ranges" and "Formatting now degrades to a warning", both of which the current HEAD contradicts. Each was checked against the actually-pushed file before being called stale.

<sub>Posted by Claude Opus 5 (1M context) via the `devin-review` skill.</sub>

PR #8159, opened 2026-08-05
Add tests for the folder-wide image shrink path, and stop one bad image aborting it (BL-16647)

JohnThomson:
Adds the missing test coverage for the folder-wide image shrinking that protects Bloom from oversized images — and, off the back of what the tests turned up, one production fix.

Nothing in `BloomTests` exercised `ImageUtils.NeedToShrinkImages` or `ImageUtils.FixSizeAndTransparencyOfImagesInFolder`. `BookStorageTests` calls `MigrateToMediaLevel1ShrinkLargeImages()` in several places, but always as a chained no-op: no test put an oversized image on disk, so `NeedToShrinkImages` returned false and the real GraphicsMagick resize never ran. (`ImageUtilsTests` covers `ResizeImageIfNecessary`/`DrawResizedImage` — thumbnail drawing, a different path.)

This matters more since BL-16627, where `BookProcessor.ProcessBook` — the `external/process-book` fix-up BloomBridge drives — runs the shrink on every book with oversized images, in addition to the legacy book-open migration.

### What the new `ImageFolderShrinkTests` covers

Images are generated programmatically rather than committed as multi-megapixel binaries.

**`NeedToShrinkImages`**
- false when everything fits, including an image exactly at the 3840x2800 maximum (boundary case)
- true for an oversized PNG, and for an oversized portrait `.jpeg` (the second JPEG extension the scan has to catch)
- false when the only oversized image is `placeholder.png` (the exemption)
- false, rather than a throw, for a folder of non-images (`.htm`, `.json`, and a `.png` that is actually text)

**`FixSizeAndTransparencyOfImagesInFolder`**
- shrinks an oversized landscape PNG, portrait JPG and landscape `.jpeg` so they fit within 3840x2800 with the aspect ratio preserved, while leaving an already-small PNG, an already-small JPG and an oversized `placeholder.png` byte-for-byte untouched
- rewrites nothing at all when nothing is oversized
- preserves intellectual-property metadata (creator, copyright, CC license) across the resize — GraphicsMagick loses TagLib metadata, so the code has to copy it back
- keeps going when a file in the folder cannot be read as an image at all (see the fix below)

### The production fix

`NeedToShrinkImages` tolerates a file it cannot read as an image — it catches, and moves on. `FixSizeAndTransparencyOfImagesInFolder` did not: it called `RobustFileIO.CreateTaglibFile` unprotected for every PNG and every JPEG. So a book folder holding an oversized image **plus one mislabeled file** passed the scan and then threw out of the middle of the fix-up. Every image the loop had not yet reached stayed oversized, and the caller went down with it — the book-open migration, or `BookProcessor.ProcessBook`, which BloomBridge drives over every book.

That input is not hypothetical. `RobustFileIO.MetadataFromFile` says so itself: *"Books sometimes use image files that are mislabeled. JPEG files sometimes have been given .png extensions… TagLib crashes trying to read the metadata in such cases"* — and it works around that by copying to a correct extension first. `CreateTaglibFile` has no such workaround.

The fix reads metadata through a new `TryCreateTaglibFileForImage`, which returns null instead of throwing, so just that one file is skipped — the same thing the scan already does. The failure is logged rather than silently swallowed. The covering test uses a folder with an oversized PNG plus a JPEG-named-`.png`, a PNG-named-`.jpg` and a file that is not an image at all: the oversized image still gets shrunk, and the three bad files are left byte-for-byte alone.

Found by Devin's review of this PR, then reproduced with a throwaway test before being acted on.

### Notes

- These exercise the real GraphicsMagick executable, as the format-conversion tests in the same folder already do. They pass through `build/agent-dotnet.sh test`.
- The metadata test was mutation-checked: temporarily disabling the `CopyTags`/`SaveTaglibFile` copy-back in `ImageUtils` makes it fail, so it is not vacuous.
- The BL-11441 "output would be larger" and BL-16424 indexed-PNG guards mentioned on the card live in `TryResizeImageWithGraphicsMagick`, which the folder path never calls (it goes through `ResizeImageFileWithOptionalTransparency`). They are reachable only from the import/display path, so they are deliberately out of scope here.

Ref: https://issues.bloomlibrary.org/youtrack/issue/BL-16647

[Devin review](https://devinreview.com/BloomBooks/BloomDesktop/pull/8159)

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8159)

### DIFF
--- a/src/BloomExe/ImageProcessing/ImageUtils.cs
+++ b/src/BloomExe/ImageProcessing/ImageUtils.cs
@@ -1489,7 +1489,12 @@ namespace Bloom.ImageProcessing
                 // Very large PNG files can cause "out of memory" errors here, while making thumbnails,
                 // and when creating ePUBs or BloomPub books.  So, we check for sizes bigger than our
                 // maximum and reduce the image here if needed.
-                var tagFile = RobustFileIO.CreateTaglibFile(path);
+                var tagFile = TryCreateTaglibFileForImage(path);
+                if (tagFile == null)
+                {
+                    ++completed;
+                    continue;
+                }
                 if (tagFile.Properties != null && tagFile.Properties.Description.Contains("PNG"))
                 {
                     var size = GetDesiredImageSize(
@@ -1554,7 +1559,12 @@ namespace Bloom.ImageProcessing
                 // Very large JPG files can cause "out of memory" errors while making thumbnails and
                 // when creating ePUBs or BloomPub books.  So, we check for sizes bigger than our
                 // maximum and reduce the image here if needed.
-                var tagFile = RobustFileIO.CreateTaglibFile(path);
+                var tagFile = TryCreateTaglibFileForImage(path);
+                if (tagFile == null)
+                {
+                    ++completed;
+                    continue;
+                }
                 if (tagFile.Properties != null && tagFile.Properties.Description.Contains("JFIF"))
                 {
                     var size = GetDesiredImageSize(
@@ -1577,6 +1587,36 @@ namespace Bloom.ImageProcessing
                     }
                 }
                 ++completed;
+            }
+        }
+
+        /// <summary>
+        /// Read a book image's metadata, or return null if the file cannot be read as the kind of
+        /// image its extension claims it is.
+        /// </summary>
+        /// <remarks>
+        /// Book folders really do contain mislabeled image files: see the comment on
+        /// RobustFileIO.MetadataFromFile, which works around JPEG files that have been given .png
+        /// extensions.  TagLib throws on those, and on any file that is not an image at all.
+        /// NeedToShrinkImages already shrugs such a file off and carries on; before BL-16647 the
+        /// fix-up did not, so one bad file could throw out of the middle of
+        /// FixSizeAndTransparencyOfImagesInFolder, leaving the rest of the folder's oversized
+        /// images unshrunk and aborting whatever asked for the fix-up (the book-open migration, or
+        /// BookProcessor.ProcessBook).  Skipping just the unreadable file is what the caller wants:
+        /// we cannot resize an image we cannot read, and it is not worth failing the whole book over.
+        /// </remarks>
+        private static TagLib.File TryCreateTaglibFileForImage(string path)
+        {
+            try
+            {
+                return RobustFileIO.CreateTaglibFile(path);
+            }
+            catch (Exception e)
+            {
+                Logger.WriteEvent(
+                    $"ImageUtils could not read {path} as an image, so it was left alone: {e.Message}"
+                );
+                return null;
             }
         }
 
--- a/src/BloomTests/ImageProcessing/ImageFolderShrinkTests.cs
+++ b/src/BloomTests/ImageProcessing/ImageFolderShrinkTests.cs
@@ -1,0 +1,452 @@
+using System;
+using System.Collections.Generic;
+using System.Drawing;
+using System.Drawing.Imaging;
+using System.IO;
+using Bloom.ImageProcessing;
+using NUnit.Framework;
+using SIL.Core.ClearShare;
+using SIL.Progress;
+using SIL.TestUtilities;
+using SIL.Windows.Forms.ClearShare;
+
+namespace BloomTests.ImageProcessing
+{
+    /// <summary>
+    /// Tests for the folder-wide image shrinking that protects Bloom from oversized images:
+    /// ImageUtils.NeedToShrinkImages (the cheap TagLib header scan) and
+    /// ImageUtils.FixSizeAndTransparencyOfImagesInFolder (the in-place GraphicsMagick resize).
+    /// This path runs both from the old-book migration (BookStorage.MigrateToMediaLevel1ShrinkLargeImages)
+    /// and, since BL-16627, from BookProcessor.ProcessBook on every book with oversized images.
+    /// </summary>
+    /// <remarks>
+    /// These tests exercise the real GraphicsMagick executable that ships in the "gm" folder
+    /// beside Bloom, just as other tests in this folder (e.g. the format-conversion tests) do.
+    /// The images are generated programmatically rather than committed as binaries, both to keep
+    /// the repository small and because a multi-megapixel image is what makes the test meaningful.
+    /// </remarks>
+    [TestFixture]
+    public class ImageFolderShrinkTests
+    {
+        // Comfortably over ImageUtils.MaxLength x MaxBreadth (3840x2800) in both dimensions, but
+        // not so large that generating and resizing the images makes the tests slow.
+        private const int kOversizedLongSide = 3900;
+        private const int kOversizedShortSide = 2850;
+
+        #region NeedToShrinkImages
+
+        [Test]
+        public void NeedToShrinkImages_AllImagesWithinLimits_ReturnsFalse()
+        {
+            using (var folder = new TemporaryFolder("NeedToShrinkImages_AllImagesWithinLimits"))
+            {
+                CreateImageFile(folder.Combine("small.png"), 400, 300, ImageFormat.Png);
+                CreateImageFile(folder.Combine("small.jpg"), 400, 300, ImageFormat.Jpeg);
+                // Exactly at the maximum is not too big.
+                CreateImageFile(
+                    folder.Combine("exactly-max.png"),
+                    ImageUtils.MaxLength,
+                    ImageUtils.MaxBreadth,
+                    ImageFormat.Png
+                );
+
+                Assert.That(ImageUtils.NeedToShrinkImages(folder.Path), Is.False);
+            }
+        }
+
+        [Test]
+        public void NeedToShrinkImages_OversizedPng_ReturnsTrue()
+        {
+            using (var folder = new TemporaryFolder("NeedToShrinkImages_OversizedPng"))
+            {
+                CreateImageFile(folder.Combine("small.jpg"), 400, 300, ImageFormat.Jpeg);
+                CreateImageFile(
+                    folder.Combine("big.png"),
+                    kOversizedLongSide,
+                    kOversizedShortSide,
+                    ImageFormat.Png
+                );
+
+                Assert.That(ImageUtils.NeedToShrinkImages(folder.Path), Is.True);
+            }
+        }
+
+        [Test]
+        public void NeedToShrinkImages_OversizedJpeg_ReturnsTrue()
+        {
+            using (var folder = new TemporaryFolder("NeedToShrinkImages_OversizedJpeg"))
+            {
+                CreateImageFile(folder.Combine("small.png"), 400, 300, ImageFormat.Png);
+                // Portrait orientation, and the ".jpeg" spelling, which the scan also has to catch.
+                CreateImageFile(
+                    folder.Combine("big.jpeg"),
+                    kOversizedShortSide,
+                    kOversizedLongSide,
+                    ImageFormat.Jpeg
+                );
+
+                Assert.That(ImageUtils.NeedToShrinkImages(folder.Path), Is.True);
+            }
+        }
+
+        [Test]
+        public void NeedToShrinkImages_OnlyOversizedImageIsPlaceholder_ReturnsFalse()
+        {
+            using (var folder = new TemporaryFolder("NeedToShrinkImages_OnlyPlaceholderIsBig"))
+            {
+                CreateImageFile(folder.Combine("small.png"), 400, 300, ImageFormat.Png);
+                CreateImageFile(
+                    folder.Combine("placeholder.png"),
+                    kOversizedLongSide,
+                    kOversizedShortSide,
+                    ImageFormat.Png
+                );
+
+                Assert.That(
+                    ImageUtils.NeedToShrinkImages(folder.Path),
+                    Is.False,
+                    "The placeholder image is exempt, so nothing here needs shrinking."
+                );
+            }
+        }
+
+        [Test]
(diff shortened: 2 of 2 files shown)


src/BloomTests/ImageProcessing/ImageFolderShrinkTests.cs
```diff
@@ -0,0 +1,461 @@
...
+                    "Test setup problem: nothing in the folder was seen as oversized."
+                );
+                var originalLandscapePngSize = GetImageDimensions(landscapePngPath);
+                var originalPortraitJpgSize = GetImageDimensions(portraitJpgPath);
+                var originalLandscapeJpegSize = GetImageDimensions(landscapeJpegPath);
+                var smallPngBytes = File.ReadAllBytes(smallPngPath);
+                var smallJpgBytes = File.ReadAllBytes(smallJpgPath);
+                var placeholderBytes = File.ReadAllBytes(placeholderPath);
+
+                ImageUtils.FixSizeAndTransparencyOfImagesInFolder(
```
JohnThomson, line 184: [Devin] **Investigate**: Tests depend on the real GraphicsMagick executable and GDI+ image encoding

These tests call `ImageUtils.FixSizeAndTransparencyOfImagesInFolder`, which goes through `ResizeImageFileWithOptionalTransparency` and shells out to the bundled GraphicsMagick (`src/BloomExe/ImageProcessing/ImageUtils.cs:1571`). If GraphicsMagick is missing or fails, that method returns false and the file is left oversized, which would surface here as a confusing dimension-assertion failure rather than an explicit "GraphicsMagick unavailable" message. Other tests in this folder that depend on the same executable carry a `Category("SkipOnTeamCity")` guard (`src/BloomTests/ImageProcessing/ImageUtilsTests.cs:155`) for Linux-agent noise; worth confirming with CI whether these new tests need similar treatment.

JohnThomson: [Claude Opus 5 (1M context)] Half acted on, half not an issue.

**The diagnostics point is fair, and is now fixed** in `ccca1fcbe`. If GraphicsMagick is missing or fails, the image is simply left at its original size, and the old failure was a bare "unexpected width" that a person would have to decode. The helper now checks first that the file changed size at all, and says so: *"…was left at its original size. The usual cause is that GraphicsMagick (the 'gm' folder beside the test assembly) is missing or failed to run."*

**The CI-guard suggestion is based on a misreading, so no guard is being added.** Two reasons:

1. That `SkipOnTeamCity` at `ImageUtilsTests.cs:155` is not a GraphicsMagick guard. It sits inside an `#if __MonoCS__` block, so it only applies to Linux/Mono builds, and its comment says it is there because TeamCity turns a low-level warning message into an error on one transparency test. It says nothing about GraphicsMagick being unavailable.

2. GraphicsMagick is already a hard, unguarded dependency of tests that live in this very fixture. `ShouldChangeFormatToJpeg_Photo_True` asserts that `TryChangeFormatToJpegIfHelpful` returns **true** and writes a `.jpg`; that method returns **false** without ever creating a file when `gm` is absent (`ImageUtils.cs:2255`). So if GraphicsMagick were missing on a CI agent, that test — which is not skipped anywhere — would already be red. The new tests add no CI exposure that the suite does not already have.

Resolving.

src/BloomTests/ImageProcessing/ImageFolderShrinkTests.cs
```diff
@@ -0,0 +1,461 @@
...
+                Assert.That(
+                    ImageUtils.NeedToShrinkImages(folder.Path),
+                    Is.False,
+                    "The placeholder image is exempt, so nothing here needs shrinking."
+                );
+            }
+        }
+
+        [Test]
+        public void NeedToShrinkImages_NonImageFilesInFolder_ReturnsFalse()
```
JohnThomson, line 114: [Devin] **Investigate**: Non-image file coverage exists only for the scan, not for the fix-up

`NeedToShrinkImages_NonImageFilesInFolder` relies on the per-file try/catch inside `NeedToShrinkImages` (`src/BloomExe/ImageProcessing/ImageUtils.cs:1375-1379`), which swallows TagLib failures on a bogus `.png`. `FixSizeAndTransparencyOfImagesInFolder` has no such guard: it calls `RobustFileIO.CreateTaglibFile(path)` unprotected for both PNGs (`src/BloomExe/ImageProcessing/ImageUtils.cs:1491`) and JPEGs (`src/BloomExe/ImageProcessing/ImageUtils.cs:1556`). A book folder containing an oversized image plus a mis-named/corrupt `.png` would therefore throw out of the fix-up, aborting the migration or `BookProcessor.ProcessBook`. Since this PR is about covering that path, a matching non-image case for the fix-up would document (or expose) that asymmetry.

JohnThomson:
[Claude Opus 5 (1M context)] Consulted Devin on 2026-08-05 23:35 UTC, up to commit `ccca1fcbe8`.

No bugs. Two things it asked us to look into, each with its own thread:

- **Tests lean on the real GraphicsMagick** — half fair. If `gm` were missing the failure used to read as a bare "unexpected width", so `ccca1fcbe8` now says so plainly. Its suggestion to add a `SkipOnTeamCity` guard is based on a misreading, and no guard was added — the reasoning is in [that thread](https://github.com/BloomBooks/BloomDesktop/pull/8159#discussion_r3724735222), which is resolved.
- **The fix-up has no guard against a corrupt image file** — a real point, and about production code rather than these tests: `NeedToShrinkImages` shrugs off a `.png` that isn't really a PNG, but `FixSizeAndTransparencyOfImagesInFolder` would throw on one. [That thread](https://github.com/BloomBooks/BloomDesktop/pull/8159#discussion_r3724755089) is deliberately left **open** — it is with the developer to decide, since fixing it means changing behavior beyond what this test-only PR set out to do.

Three informational items were not mirrored (low signal): the expected size being computed with the same helper under test, the aspect-ratio tolerance being generous, and a note that `NullProgress` must supply a non-null progress indicator.

CI (`pr-automation`) is green. CodeRabbit does not auto-review this repo (`auto_review.enabled: false` in `.coderabbit.yml`), so it was not waited on.

src/BloomTests/ImageProcessing/ImageFolderShrinkTests.cs
JohnThomson: [Claude Opus 5 (1M context)] Good catch — this was real, and it is now fixed in `42d69337`.

Before dismissing or acting on it I reproduced it, because the interesting question was whether the input is realistic. It is. A folder holding an oversized picture plus a genuine JPEG named `mislabeled.png` passed the scan and then threw `TagLib.CorruptFileException` out of the middle of the fix-up, so every image the loop had not yet reached stayed oversized and the caller went down with it. That input is one Bloom is documented to meet: `RobustFileIO.MetadataFromFile` says in as many words that books contain JPEG files given `.png` extensions and that TagLib crashes on them — it copies the file to a correct extension to work around it, but `CreateTaglibFile`, which the fix-up uses, does not.

The fix reads the metadata through a new `TryCreateTaglibFileForImage`, which returns null instead of throwing, and skips just that one file — exactly what `NeedToShrinkImages` already does. The failure is logged rather than silently swallowed. Both the PNG and the JPEG loop are covered.

Your framing was that a matching non-image case "would document (or expose) that asymmetry". It exposed it. There is now a test for it: a folder with an oversized PNG plus a JPEG-named-`.png`, a PNG-named-`.jpg` and a file that is not an image at all — the oversized image gets shrunk and the three bad files are left byte-for-byte alone.

Resolving.

src/BloomExe/ImageProcessing/ImageUtils.cs
```diff
@@ -1580,6 +1590,36 @@ IProgress progress
...
+        /// RobustFileIO.MetadataFromFile, which works around JPEG files that have been given .png
+        /// extensions.  TagLib throws on those, and on any file that is not an image at all.
+        /// NeedToShrinkImages already shrugs such a file off and carries on; before BL-16647 the
+        /// fix-up did not, so one bad file could throw out of the middle of
+        /// FixSizeAndTransparencyOfImagesInFolder, leaving the rest of the folder's oversized
+        /// images unshrunk and aborting whatever asked for the fix-up (the book-open migration, or
+        /// BookProcessor.ProcessBook).  Skipping just the unreadable file is what the caller wants:
+        /// we cannot resize an image we cannot read, and it is not worth failing the whole book over.
+        /// </remarks>
+        private static TagLib.File TryCreateTaglibFileForImage(string path)
```
JohnThomson, line 1608: [Devin] **Investigate**: PR description says "No production code changed" but ImageUtils gained error swallowing

The PR body states "No production code changed", yet `src/BloomExe/ImageProcessing/ImageUtils.cs` now routes both loops of `FixSizeAndTransparencyOfImagesInFolder` through the new `TryCreateTaglibFileForImage`, which catches every exception and returns null. That is a genuine behavior change (previously an unreadable image aborted the whole folder fix-up and propagated to the caller). It is a sensible fix for BL-16647, but the description should be updated so reviewers/release notes don't miss it.

JohnThomson: [Claude Opus 5 (1M context)] Correct, and my mistake — the description was written when this really was test-only and I did not revisit it after adding the fix. Fixed: the PR title now says the branch also stops one bad image aborting the shrink, and the body has a section describing the behavior change, why the mislabeled-file input is one Bloom actually meets, and the test that covers it.

Resolving.

src/BloomTests/ImageProcessing/ImageFolderShrinkTests.cs
```diff
@@ -0,0 +1,525 @@
...
+                originalSize.Width,
+                originalSize.Height
+            );
+            Assert.That(
+                expectedSize,
+                Is.Not.EqualTo(originalSize),
+                $"Test setup problem: {label} was not oversized to begin with."
+            );
+
+            var newSize = GetImageDimensions(path);
```
JohnThomson, line 438: [Devin] **Investigate**: New tests depend on the real GraphicsMagick binary being present

`AssertShrunkToFitWithAspectRatioPreserved` will fail (with a helpful message) if `gm` is missing, and `FixSizeAndTransparencyOfImagesInFolder_...ShrinksOversized...` additionally asserts `NeedToShrinkImages` becomes false afterwards. If any CI lane runs BloomTests without the `gm` folder staged beside the test assembly, these four tests become hard failures rather than being skipped. Worth confirming the test agents always have GraphicsMagick, or adding an `Assert.Ignore`/category guard as the existing format-conversion tests do.

JohnThomson: [Claude Opus 5 (1M context)] This is the same concern as [the earlier thread on this PR](https://github.com/BloomBooks/BloomDesktop/pull/8159#discussion_r3724735222), re-raised after the fix commit, and the answer has not changed.

GraphicsMagick is **already** a hard, unguarded dependency of a test in this very fixture. `ShouldChangeFormatToJpeg_Photo_True` asserts that `TryChangeFormatToJpegIfHelpful` returns **true** and writes a `.jpg`; without `gm` that method returns **false** and never creates the file (`ImageUtils.cs:2255`). So a CI lane missing GraphicsMagick would already be red today, on a test nobody has felt the need to guard. These tests add no exposure the suite does not already carry.

On the suggestion to copy the existing guard: that `Category("SkipOnTeamCity")` at `ImageUtilsTests.cs:155` is not a GraphicsMagick guard. It sits inside an `#if __MonoCS__` block, so it applies only to Linux/Mono, and its own comment says it exists because TeamCity turns a low-level warning into an error on one transparency test.

The half of the original flag that *was* fair — that a missing `gm` used to surface as a baffling "unexpected width" — was fixed in `ccca1fcb`, which is the helpful message this flag now notes.

Resolving.

JohnThomson:
[Claude Opus 5 (1M context)] Consulted Devin again on 2026-08-06, up to commit `42d69337` — the commit that fixes the crash it found last round.

No bugs. Three things to look into, each on its own thread, all now resolved:

- **The PR description still said "No production code changed"** — quite right, and my oversight; it was written when this branch really was test-only. The title and body now describe the fix. [Thread](https://github.com/BloomBooks/BloomDesktop/pull/8159#discussion_r3729098279)
- **The tests need the real GraphicsMagick** — the same concern as last round, re-raised, and the answer is unchanged: a test already in this fixture fails without `gm` today, unguarded, so these add no new exposure. [Thread](https://github.com/BloomBooks/BloomDesktop/pull/8159#discussion_r3729100883)
- **"The fix-up has no guard against a corrupt image file"** — re-raised word for word from the previous review, but it describes code this commit changed: it cites the two unprotected `CreateTaglibFile` calls that `42d69337` replaced, and asks for fix-up coverage that `42d69337` adds. Nothing to do; the outcome is recorded on [the original thread](https://github.com/BloomBooks/BloomDesktop/pull/8159#discussion_r3724755089).

Seven informational items were not mirrored. Two of them are worth a reader knowing, since they are about the new fix rather than the tests: catching every exception also swallows a transient file-lock failure and would leave that image oversized (deliberate — it matches what the scan next door already does, and it is logged), and the TagLib file handles in those loops are still never disposed (pre-existing, untouched here).

CI (`pr-automation`) is green. CodeRabbit does not auto-review this repo.

andrew-polk (approved):
> @andrew-polk reviewed 2 files and all commit messages, and resolved 4 discussions.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8159)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on JohnThomson).

PR #8160, opened 2026-08-05
Keep the Leveled Reader control block visible while the stats scroll (BL-16585)

StephenMcConnel:
The Decodable Reader tool keeps its stage nav pinned at the top and its "Book is Decodable" / "Set Up Stages" control block pinned at the bottom, letting only the word lists scroll. The Leveled Reader had no internal scrolling region, so a tall stats panel made the whole tool overflow: the accordion panel scrolled and the "Book is Leveled" toggle and "Set Up Levels" button scrolled out of sight.

This gives the Leveled Reader the same three-band layout:

| | pinned top | scrolls | pinned bottom |
|---|---|---|---|
| Decodable | Stage *n* of *m* | grapheme + word grids | Generate Report / toggle / Set Up Stages |
| Leveled | Level *n* of *m* | banner, legend, stats, "For this Level", "Keep in mind" | Copy Book Stats / toggle / Set Up Levels |

The scrolling region carries a `min-height` floor (as the Decodable Reader's sections do) rather than `min-height: 0`. Without a floor it is the only shrinkable item in the column, so in a short panel it collapsed to zero height, leaving the stats both invisible and unscrollable. With the floor, a panel too short even for that overflows and scrolls as a whole — what it did before this change, and what the Decodable Reader does today.

CSS only; no behavior or API changes.

### Verified in a running Bloom

Measured over CDP in the Edit tab with a leveled book, at several forced panel heights:

- Normal height: the stats region scrolls internally; nav and control block stay put.
- Squeezed to 220 / 170 / 140 px: the stats region floors instead of collapsing, the panel scrolls as a whole, and every control (including "Set Up Levels") remains reachable — the same numbers the Decodable Reader produces under the identical squeeze.

Ref: https://issues.bloomlibrary.org/youtrack/issue/BL-16585

[Devin review](https://devinreview.com/BloomBooks/BloomDesktop/pull/8160)

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8160)

### DIFF
--- a/src/BloomBrowserUI/bookEdit/toolbox/readers/leveledReader/LeveledReaderToolControls.tsx
+++ b/src/BloomBrowserUI/bookEdit/toolbox/readers/leveledReader/LeveledReaderToolControls.tsx
@@ -631,53 +631,87 @@ export const LeveledReaderToolControls: FunctionComponent = () => {
                 >
                     {showTool && (
                         <>
+                            {/* "Level x of y" and its arrows stay pinned at the top,
+                                outside the scrolling region below, just as the
+                                Decodable Reader's stage nav does. (BL-16585) */}
                             <ReaderToolNav
                                 isForLeveled={true}
                                 changeFunction={changeLevel}
                             />
-                            {/* The over-level warning banner and the within/over
-                                legend only make sense when at least one measure is
-                                actually over the current level's limit. */}
-                            {isBookOverLevel(model, bookStats) && (
-                                <>
-                                    <OverLevelBanner
-                                        levelNumber={model.levelNumber}
+                            {/* The stats and lists scroll here, inside the panel,
+                                rather than letting the whole panel scroll. That
+                                keeps the nav above and the control block below (the
+                                "Book is Leveled" toggle and Set Up Levels) always
+                                visible, the way the Decodable Reader's word lists
+                                scroll between its fixed nav and control block. */}
+                            <div
+                                css={css`
+                                    display: flex;
+                                    flex-direction: column;
+                                    flex: 1 1 auto;
+                                    // A floor, not min-height:0, for the same reason
+                                    // the Decodable Reader's sections have one: with
+                                    // nothing to stop it, this region is the only
+                                    // shrinkable thing in the column, so in a short
+                                    // panel it would shrink to 0 and the stats would
+                                    // be invisible AND unscrollable. Kept short so it
+                                    // rarely forces the panel to scroll; when the
+                                    // panel is too short even for this, the panel
+                                    // scrolls as a whole (as it did before) and
+                                    // everything stays reachable.
+                                    min-height: 120px;
+                                    overflow-y: auto;
+                                    overflow-x: hidden;
+                                    // Reserve space for the vertical scrollbar so
+                                    // its appearance (when the stats grow tall)
+                                    // doesn't shrink/reflow the content, matching
+                                    // the Decodable Reader's word grids.
+                                    scrollbar-gutter: stable;
+                                `}
+                            >
+                                {/* The over-level warning banner and the within/over
+                                    legend only make sense when at least one measure
+                                    is actually over the current level's limit. */}
+                                {isBookOverLevel(model, bookStats) && (
+                                    <>
+                                        <OverLevelBanner
+                                            levelNumber={model.levelNumber}
+                                        />
+                                        <StatsLegend />
+                                    </>
+                                )}
+                                <LeveledReaderStats bookStats={bookStats} />
+                                {levelReminders.length > 0 && (
+                                    <LeveledReaderList
+                                        l10nKeySuffix="FoThisLevel"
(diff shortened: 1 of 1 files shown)


src/BloomBrowserUI/bookEdit/toolbox/readers/leveledReader/LeveledReaderToolControls.tsx
```diff
@@ -631,53 +639,87 @@ export const LeveledReaderToolControls: FunctionComponent = () => {
...
+                                    // the Decodable Reader's sections have one: with
+                                    // nothing to stop it, this region is the only
+                                    // shrinkable thing in the column, so in a short
+                                    // panel it would shrink to 0 and the stats would
+                                    // be invisible AND unscrollable. Kept short so it
+                                    // rarely forces the panel to scroll; when the
+                                    // panel is too short even for this, the panel
+                                    // scrolls as a whole (as it did before) and
+                                    // everything stays reachable.
+                                    min-height: 120px;
```
StephenMcConnel, line 670: [Devin] **Investigate**: Fixed 120px floor is a magic number that may not fit one stats row plus header

The scrolling region's `min-height: 120px` is a hard-coded floor. If a future toolbox font-size or a localized stats header grows, 120px may not correspond to any meaningful amount of visible content; conversely, in a moderately short panel it forces the whole panel to scroll a little earlier than strictly necessary. Worth confirming this value against the Decodable Reader's equivalent floors so the two tools stay consistent if either is tuned later.

*(Devin finding for commit `2b5d35a`, mirrored onto the PR by Claude Opus 5 during preflight — Devin doesn't post to GitHub itself.)*

StephenMcConnel: [Claude Opus 5, during preflight] Assessed as not an issue — leaving the 120px as it is.

The floor's job isn't to guarantee any particular amount of visible content; it's to stop this region collapsing to nothing. Without a floor it's the only shrinkable item in the column, so a short panel squeezed it to **0px** and the stats became both invisible and unscrollable — that's the bug the floor exists to prevent, and it's why the value being a little too big or too small is harmless: too small just means less shows before the panel scrolls, too large means the panel starts scrolling slightly sooner. Neither can make anything unreachable, because a panel too short even for the floor scrolls as a whole.

On consistency with the Decodable Reader — that's exactly where the number came from. Its two sections use the same technique with hard-coded floors of 100px and 200px (`DecodableReaderToolControls.tsx`), so 120px sits between them and the two tools behave the same way today.

Measured in a running Bloom at forced panel heights of 400 / 220 / 170 / 140 px: the stats region never collapses, and "Set Up Levels" stays reachable at every one of them — the same numbers the Decodable Reader produces under an identical squeeze.

StephenMcConnel:
[Claude Opus 5, during preflight] Consulted Devin on 2026-08-05 17:34 MDT, up to commit `2b5d35a`.

No bugs. One **Investigate** flag — that the 120px floor on the scrolling region is a magic number — which is now its own thread on the diff, assessed as not an issue and resolved there with the reasoning.

Five informational items, not mirrored as threads. Two of them simply confirmed the layout reasoning (that dropping `margin-top: auto` is safe now that the scroll region absorbs the slack, and that the nav can't be squeezed). One noted that the whole-panel fallback scrolling was relying on `overflow-x: hidden` *implicitly* computing `overflow-y` to `auto` — that was worth acting on, and commit `2b5d35a` states the `overflow-y: auto` explicitly so a later edit can't silently remove the fallback. The remaining two (the scrollbar sitting inside the panel's 12px padding, and the earlier run's duplicate of the margin note) need no action: the Decodable Reader's scrolling grids sit inside that same padding, which is the parity this PR is aiming for.

CI (`pr-automation`) passed. CodeRabbit has `auto_review.enabled: false` on this repo, so it does not review PRs here.

src/BloomBrowserUI/bookEdit/toolbox/readers/leveledReader/LeveledReaderToolControls.tsx
StephenMcConnel: [Claude Opus 5 from McConnel's machine during preflight] Developer decision: leave the 120px floor as it is.

Steve reviewed this in the preflight report and confirmed the assessment above — the floor stays a plain pixel value, matching the `min-height: 100px` / `min-height: 200px` already used by the Decodable Reader's sections that this change is deliberately mirroring. Closing the loop so the record shows a human signed off on it rather than only the agent.

andrew-polk (approved):
> @andrew-polk reviewed 1 file and all commit messages, and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8160)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on StephenMcConnel).

PR #8161, opened 2026-08-06
Add e2e survey endpoints and move the branding viewer to its own repo

hatton:
Host side of the contract for **https://github.com/BloomBooks/branding-viewer** — a standalone tool that surveys how a book renders across branding × page × layout × xmatter.

## Why

The branding survey tool used to live in this repo at `.github/skills/branding-report`, and it had two hard ties to a development environment:

1. It built its branding list by reading `src/content/branding` out of a checkout.
2. It drove a `POST settings/branding` handler that existed only inside `#if DEBUG`.

So it could only ever run on a developer's machine. The goal is for a **tester with nothing but a CI-built Bloom** to run it. These two endpoints remove both ties.

## What

Both are registered only when Bloom runs with `--e2e`, in `E2eTestingApi`:

```
POST /bloom/api/e2e/setState      {branding, layout, xmatter}   omitted/null = unchanged
GET  /bloom/api/e2e/surveyOptions -> {bloomVersion, brandings[], layouts[],
                                      xmatterOfferings[], current{}, book{}}
```

`setState` sets all three axes in one call, so a survey costs one book-update per cell instead of three.

`surveyOptions` is what lets the tool live in its own repo: the running Bloom becomes the source of truth for what *that build* ships. `bloomVersion` also gives the tool a capability check, so an older Bloom gets a clear explanation instead of an opaque error.

## Notes for review

- **The `--e2e` gate is a runtime check, not `#if DEBUG`.** That was already true (this file's comment explains CI runs the e2e suite against Release), and the viewer depends on it, because a tester runs an installed Release build. Worth a look that you're happy extending that surface.
- **`setState` catches its own exceptions** and reports via `request.Failed()`, unlike its neighbours. A survey walks hundreds of cells and some combinations are genuinely invalid (a branding whose xmatter the collection doesn't offer, a layout the book's stylesheets don't support). One bad cell must not take Bloom down. This mirrors what the BL-16370 prototype handler did.
- **`setState` reuses `MakeSubscriptionForBranding`**, which already handles the Local-Community personalization trap (a bare `Local-Community` descriptor makes `MergeInPersonalization` throw). The BL-16370 prototype hard-coded Enterprise tier and got that wrong.
- **`surveyOptions` returns an empty `layouts` list until a book is selected**, because Bloom derives layout choices from the selected book's stylesheets. Documented in the XML comment; the tool asks again after it selects a book.

## Verified

Built, then launched `Bloom.exe <collection> --e2e --automation` on a throwaway copy of the visual-regression collection and drove the relocated tool against it:

- `surveyOptions` returned Bloom 6.5.0.0, 73 brandings, the 5 offered xmatter packs, and the current state.
- `setState` + capture produced 3/3 back-cover screenshots for WorldVision, World-Vision-International and Default at Device16x9Landscape, with real QR codes, then restored the baseline.
- The tool ran from a clone with **no** BloomDesktop checkout, and as a compiled single-file exe.

This also incidentally exercises DI: the new `XMatterPackFinder` constructor parameter resolves through Autofac at runtime.

## Not done here

No automated test covers the new endpoints. `E2eTestingApi` has none today, and the endpoints exist to be driven by an out-of-repo tool. Happy to add coverage if you'd rather.

[Devin review](https://devinreview.com/BloomBooks/BloomDesktop/pull/8161)

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8161)


hatton:
[Claude Opus 5 from Hatton's machine following a prompt from Hatton] Added a second commit: retires the in-repo copy of the tool at `.github/skills/branding-report/` (leaving a SKILL.md that says where the code went and what stayed), and adds `.claude/skills/branding-viewer/SKILL.md` as the how-to-run quick path.

The companion repo is now live at https://github.com/BloomBooks/branding-viewer — initial import pushed, CI compiles the standalone exe and smoke-tests it.

Note for whoever reviews: `./go.sh` does not pass `--e2e`, so a Bloom started the usual way will not serve these endpoints. That is called out in the skill.

hatton:
[Claude Opus 5 following a prompt from Hatton] Closing: built on the wrong premise. This gated the endpoints behind `--e2e` and put them in `E2eTestingApi`, but (a) a branding survey is not end-to-end testing, so the name misleads, (b) requiring a launch flag pushes the burden onto testers, who won't remember it, and (c) `--e2e` and `E2eTestingApi` exist only on master, which forced two implementations.

Re-landing against `Version6.4` instead, where the branding work actually lives, as always-on endpoints under the existing `external/` namespace (`ExternalApi` is already registered unconditionally and already documented as "API functions which are called from outside of Bloom"). Because that depends only on things 6.4 already has, one implementation serves both branches and forward-merges normally.

The two non-endpoint commits here (moving the tool out to https://github.com/BloomBooks/branding-viewer, and folding its how-to into the `bloom-branding` skill) will be carried into the new PR.

PR #8165, opened 2026-08-06
Keep the slow test categories out of developers' and the harvester's runs

andrew-polk:
## Why

A developer or agent running `dotnet test src/BloomTests/BloomTests.csproj` got the whole suite — including the `WebLibraryIntegration` tests that upload to and download from the live unit-test S3 bucket, and (since BL-16575) the `Nightly` tests that spend real seconds waiting on socket timeouts. Only the build servers excluded those, by passing `excludedCategories`, so the exclusion existed nowhere a developer would benefit from it.

## What changed

**Developers** now get a default from a new `src/BloomTests/BloomTests.runsettings`, wired in by the project's `RunSettingsFilePath`, leaving out `Integration` and `Nightly`.

The condition on that property is load-bearing. VSTest **ANDs** a runsettings `TestCaseFilter` with a command-line filter rather than replacing it, so setting it unconditionally would make `--filter TestCategory=Integration` resolve to `(TestCategory!=Integration)&(TestCategory=Integration)` — matching nothing, leaving those tests unrunnable. Keying it off `VSTestTestCaseFilter` (the property `dotnet test`'s filter argument arrives as) hands VSTest no runsettings when the caller brought a filter of their own, so asking for a category by name works.

**harvester-artifacts** wants a fast gate rather than full coverage, so it now excludes `Integration` and `Nightly` too.

**nightly** is unchanged, deliberately: it is the only runner that pays for both, and a comment there now says not to "align" its list with the others.

**Bloom.proj** needed a fix to accept more than one category, because MSBuild reads a raw semicolon in a `/p:` switch as a separator between property assignments:

- Quoting the value (`/p:"excludedCategories=A;B"`) works where the quotes reach `MSBuild.exe` — which is why TeamCity's build configuration can use that form — but PowerShell strips them, so it would hard-fail in a workflow.
- The workflows therefore pass `%3B`-escaped separators, which no shell touches, and `CategoriesToExclude` is now unescaped before being split into items.

Without that unescaping the whole list arrived as a **single** category and produced `--filter "TestCategory!=A;B"`, which matches nothing and so excludes nothing: a green build that skipped none of what it was told to skip — the same silent failure the neighbouring comment warns about for commas. Single-category callers are unaffected.

## Verification

Against the real `RunNUnit` target with real `MSBuild.exe`, each caller in its own invocation style:

| Caller | Emitted filter |
| --- | --- |
| TeamCity (quoted form, direct spawn) | `TestCategory!=SkipOnTeamCity&TestCategory!=Integration&TestCategory!=Nightly` |
| harvester (`%3B` form, from PowerShell) | `TestCategory!=Integration&TestCategory!=Nightly&TestCategory!=SkipOnTeamCity` |
| nightly (single category) | `TestCategory!=SkipOnTeamCity` |

On the developer side: the default run leaves out all 9 `Integration` tests and both `Nightly` tests while still running the non-excluded tests in the same fixtures, and `--filter TestCategory=Nightly` / `--filter TestCategory=Integration` each still select and run theirs.

## Note

Also corrects the note on `S3RetryBudgetTests`, which said nothing in this repository acts on the `Nightly` category. Two things now do.

No tracker card — this is tooling work done without one.

---
🤖 Generated with [Claude Code](https://claude.com/claude-code) (Claude Opus 5, 1M context)

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8165)

### DIFF
--- a/.github/workflows/harvester-artifacts.yml
+++ b/.github/workflows/harvester-artifacts.yml
@@ -172,6 +172,16 @@ jobs:
 
             # TestOnly runs the suite with `dotnet test` (see Bloom.proj's RunNUnit) and writes
             # NUnit3 xml to output/Tests/Release/x64/TestResults.xml.
+            #
+            # This workflow's job is to produce harvester artifacts, so it wants a fast, reliable
+            # gate rather than full coverage: Integration (talks to the live S3 bucket) and Nightly
+            # (reserved for the long tests only the nightly should pay for) are both left out. The
+            # nightly is where those run.
+            #
+            # The %3B are escaped semicolons, and they are load-bearing: a raw semicolon here would
+            # be eaten by pwsh as a statement separator and then split by MSBuild into separate
+            # /p: assignments, either of which silently reduces the list to "Integration". See the
+            # comment on CategoriesToExclude in Bloom.proj.
             - name: Run C# tests
               id: csharp_tests
               working-directory: build
@@ -179,7 +189,7 @@ jobs:
               run: |
                   msbuild Bloom.proj /t:TestOnly `
                     /p:BUILD_NUMBER=$env:BUILD_NUMBER `
-                    /p:excludedCategories=SkipOnTeamCity
+                    /p:excludedCategories=Integration%3BNightly%3BSkipOnTeamCity
 
             # ----- Test reports, one per suite -----
             # Each suite gets its OWN invocation of the publish action, and therefore its own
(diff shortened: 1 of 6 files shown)


andrew-polk:
Claude Opus 5 (1M context): closing this — it was opened from the `master_work3` worktree branch by mistake. The same commit now lives on `skip-slow-test-categories-by-default`, and a fresh PR replaces this one. The `master_work3` remote branch is being deleted.

PR #8166, opened 2026-08-06
Keep the slow test categories out of developers' and the harvester's runs

andrew-polk:
## Why

A developer or agent running `dotnet test src/BloomTests/BloomTests.csproj` got the whole suite — including the `WebLibraryIntegration` tests that upload to and download from the live unit-test S3 bucket, and (since BL-16575) the `Nightly` tests that spend real seconds waiting on socket timeouts. Only the build servers excluded those, by passing `excludedCategories`, so the exclusion existed nowhere a developer would benefit from it.

## What changed

**Developers** now get a default from a new `src/BloomTests/BloomTests.runsettings`, wired in by the project's `RunSettingsFilePath`, leaving out `Integration` and `Nightly`.

The condition on that property is load-bearing. VSTest **ANDs** a runsettings `TestCaseFilter` with a command-line filter rather than replacing it, so setting it unconditionally would make `--filter TestCategory=Integration` resolve to `(TestCategory!=Integration)&(TestCategory=Integration)` — matching nothing, leaving those tests unrunnable. Keying it off `VSTestTestCaseFilter` (the property `dotnet test`'s filter argument arrives as) hands VSTest no runsettings when the caller brought a filter of their own, so asking for a category by name works.

**harvester-artifacts** wants a fast gate rather than full coverage, so it now excludes `Integration` and `Nightly` too.

**nightly** is unchanged, deliberately: it is the only runner that pays for both, and a comment there now says not to "align" its list with the others.

**Bloom.proj** needed a fix to accept more than one category, because MSBuild reads a raw semicolon in a `/p:` switch as a separator between property assignments:

- Quoting the value (`/p:"excludedCategories=A;B"`) works where the quotes reach `MSBuild.exe` — which is why TeamCity's build configuration can use that form — but PowerShell strips them, so it would hard-fail in a workflow.
- The workflows therefore pass `%3B`-escaped separators, which no shell touches, and `CategoriesToExclude` is now unescaped before being split into items.

Without that unescaping the whole list arrived as a **single** category and produced `--filter "TestCategory!=A;B"`, which matches nothing and so excludes nothing: a green build that skipped none of what it was told to skip — the same silent failure the neighbouring comment warns about for commas. Single-category callers are unaffected.

## Verification

Against the real `RunNUnit` target with real `MSBuild.exe`, each caller in its own invocation style:

| Caller | Emitted filter |
| --- | --- |
| TeamCity (quoted form, direct spawn) | `TestCategory!=SkipOnTeamCity&TestCategory!=Integration&TestCategory!=Nightly` |
| harvester (`%3B` form, from PowerShell) | `TestCategory!=Integration&TestCategory!=Nightly&TestCategory!=SkipOnTeamCity` |
| nightly (single category) | `TestCategory!=SkipOnTeamCity` |

On the developer side: the default run leaves out all 9 `Integration` tests and both `Nightly` tests while still running the non-excluded tests in the same fixtures, and `--filter TestCategory=Nightly` / `--filter TestCategory=Integration` each still select and run theirs.

## Note

Also corrects the note on `S3RetryBudgetTests`, which said nothing in this repository acts on the `Nightly` category. Two things now do.

No tracker card — this is tooling work done without one.

---
🤖 Generated with [Claude Code](https://claude.com/claude-code) (Claude Opus 5, 1M context)

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8166)

---
[Devin review](https://app.devin.ai/review/BloomBooks/BloomDesktop/pull/8166)

### DIFF
--- a/.github/workflows/harvester-artifacts.yml
+++ b/.github/workflows/harvester-artifacts.yml
@@ -172,6 +172,19 @@ jobs:
 
             # TestOnly runs the suite with `dotnet test` (see Bloom.proj's RunNUnit) and writes
             # NUnit3 xml to output/Tests/Release/x64/TestResults.xml.
+            #
+            # This workflow's job is to produce harvester artifacts, so it wants a fast, reliable
+            # gate rather than full coverage: Integration (talks to the live S3 bucket) and Nightly
+            # (reserved for the long tests only the nightly should pay for) are both left out. The
+            # nightly is where those run.
+            #
+            # The %3B are escaped semicolons, and they are load-bearing: a raw semicolon would
+            # break this step. pwsh reads it as a statement separator, so only "Integration" would
+            # reach msbuild and the rest would be run as a command (which fails the step, since
+            # Actions sets $ErrorActionPreference = 'stop' for pwsh); and a semicolon that did
+            # arrive intact would be split by MSBuild into separate /p: assignments, failing with
+            # "MSB1006: Property is not valid". Both are loud, but neither runs any tests. See the
+            # comment on CategoriesToExclude in Bloom.proj for the two forms that do work.
             - name: Run C# tests
               id: csharp_tests
               working-directory: build
@@ -179,7 +192,7 @@ jobs:
               run: |
                   msbuild Bloom.proj /t:TestOnly `
                     /p:BUILD_NUMBER=$env:BUILD_NUMBER `
-                    /p:excludedCategories=SkipOnTeamCity
+                    /p:excludedCategories=Integration%3BNightly%3BSkipOnTeamCity
 
             # ----- Test reports, one per suite -----
             # Each suite gets its OWN invocation of the publish action, and therefore its own
--- a/src/BloomTests/BloomTests.csproj
+++ b/src/BloomTests/BloomTests.csproj
@@ -20,6 +20,24 @@
 		<SatelliteResourceLanguages>en</SatelliteResourceLanguages>
 		<IsTestProject>true</IsTestProject>
 		<Platforms>x64;ARM64</Platforms>
+		<!-- Gives `dotnet test <this project>` and the IDE test runners a default filter that
+		     leaves out the Integration and Nightly categories; see BloomTests.runsettings for why.
+
+		     The condition is what makes opting back in possible. VSTest ANDs a runsettings
+		     TestCaseFilter with any filter given on the command line (it does not replace it),
+		     so an unconditional setting here would make "TestCategory=Integration" resolve to
+		     "(TestCategory!=Integration)&(TestCategory=Integration)" and match nothing at all,
+		     making the Integration tests unrunnable. dotnet test surfaces its filter
+		     argument as the VSTestTestCaseFilter property, so when the caller asks for specific
+		     tests we hand VSTest no runsettings and let their filter stand alone.
+
+		     Consequence worth knowing: this is all-or-nothing, so ANY filter drops the whole
+		     default, not just the part you were overriding. A run narrowed by name (a filter of
+		     "FullyQualifiedName~BloomS3ClientTests", say) is therefore back to including that
+		     fixture's Integration tests, live S3 bucket and all. Add TestCategory!=Integration to
+		     such a filter yourself if you want it. (Leading hyphens are omitted throughout this
+		     comment on purpose: a double hyphen would make the XML unparseable.) -->
+		<RunSettingsFilePath Condition="'$(VSTestTestCaseFilter)' == ''">$(MSBuildProjectDirectory)\BloomTests.runsettings</RunSettingsFilePath>
 	</PropertyGroup>
 	<ItemGroup>
 		<AssemblyAttribute Include="System.Runtime.InteropServices.ComVisible">
--- a/src/BloomTests/BloomTests.runsettings
+++ b/src/BloomTests/BloomTests.runsettings
@@ -1,0 +1,41 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!--
+	Default settings for `dotnet test` (and Visual Studio's Test Explorer) on this project.
+	Wired in by BloomTests.csproj's RunSettingsFilePath, so a developer or agent gets this
+	without having to remember a flag; see the comment there for how opting back in works.
+
+	Careful: an XML comment cannot contain a double hyphen, so command-line switches are
+	written here without their leading hyphens. A stray "- -filter" in a comment makes VSTest
+	reject the whole file with "Settings file provided does not conform to required format".
+-->
+<RunSettings>
+	<RunConfiguration>
+		<!--
+			What a developer's plain `dotnet test` leaves out:
+
+			Integration: the WebLibraryIntegration tests upload to and download from the live
+			unit-test S3 bucket, so they are slow, need internet access, and need credentials for
+			that bucket.
+
+			Nightly: too slow or too environment-dependent to belong in the run a developer does
+			before every commit. S3RetryBudgetTests carries it (2 tests that wait on real socket
+			timeouts); tagging another test is all it takes to keep that one out of local runs too.
+
+			The nightly workflow is what runs all of the above, so anything you tag still gets
+			run by something. Ask for a category when you want it, by passing
+			`filter TestCategory=Integration` (with the usual leading hyphens), which suppresses
+			this file entirely.
+
+			Untested: whether clicking Run on an excluded test in Visual Studio's Test Explorer runs
+			it or silently skips it, since VSTest may combine this filter with the IDE's selection.
+			We no longer develop in Visual Studio, so nobody has checked.
+
+			The builds do not read this file at all: build/Bloom.proj's RunNUnit target, which is
+			what TeamCity and the GitHub Actions workflows invoke, runs dotnet test against the
+			already-built BloomTests.dll rather than against this project (deliberately; see the
+			comment above that target), so no MSBuild property of ours is ever evaluated there.
+			Those builds keep controlling their own categories through /p:excludedCategories.
+		-->
+		<TestCaseFilter>TestCategory!=Integration&amp;TestCategory!=Nightly</TestCaseFilter>
+	</RunConfiguration>
+</RunSettings>
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -196,6 +196,11 @@ jobs:
               shell: pwsh
               run: msbuild Bloom.proj /t:Build /m /p:BUILD_NUMBER=$env:BUILD_NUMBER
 
+            # The nightly is the run that pays for the slow tests, so it excludes as little as
+            # possible: only SkipOnTeamCity. It deliberately keeps Integration (the tests that
+            # talk to the live S3 bucket) and Nightly, both of which the harvester workflow and
+            # developers' local runs leave out. Don't "align" this list with theirs
+            # — if the nightly skips these too, nothing runs them at all.
             - name: Run C# tests
               id: csharp_tests
               if: ${{ !cancelled() && steps.build_csharp.outcome == 'success' && env.RUN_CSHARP_TESTS == 'true' }}
--- a/PAPERCUTS.md
+++ b/PAPERCUTS.md
@@ -150,6 +150,20 @@ House rules:
   environmental failures that used to accompany them. Not evidence against this cut: intermittent
   is intermittent. Recorded because it removes the nine-failure baseline the original wording
   leaned on.
+- **2026-08-06 — seen again, and the shared state this Idea line asks for is now identified:**
+  a different pair (`BookWithUnknownLayout_GetsUpdatedToA5Portrait`,
+  `CompressBookForDevice_MakesThumbnailFromCoverPicture`) failed on one full run and passed on the
+  next at the identical commit, and the exceptions name the culprit directly: an
+  `UnauthorizedAccessException` on `%TEMP%\BookTests\test\colorPalettes.json` and a
+  `DirectoryNotFoundException` on `%TEMP%\BookTests\book\basePage.css`. That path is a **fixed,
+  machine-wide** folder, so it is shared by every worktree — another worktree was running the suite
+  at the same time, and the two runs deleted each other's files mid-test. So this is not only "under
+  load": `build/agent-dotnet.sh` isolates *build* output per terminal, but the tests themselves still
+  share one scratch directory, which means two agents in two worktrees can never trust a red result.
+  Fix is narrower than the original Idea suggests: give the test scratch root a per-process suffix
+  (e.g. include the pid or the `BLOOM_AGENT_BUILD_DIR` key in the `BookTests` folder name).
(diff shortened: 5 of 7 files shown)


src/BloomTests/BloomTests.runsettings
```diff
@@ -0,0 +1,37 @@
...
+			`filter TestCategory=Integration` (with the usual leading hyphens), which suppresses
+			this file entirely.
+
+			The builds do not read this file at all: build/Bloom.proj's RunNUnit target, which is
+			what TeamCity and the GitHub Actions workflows invoke, runs dotnet test against the
+			already-built BloomTests.dll rather than against this project (deliberately; see the
+			comment above that target), so no MSBuild property of ours is ever evaluated there.
+			Those builds keep controlling their own categories through /p:excludedCategories.
+		-->
+		<TestCaseFilter>TestCategory!=Integration&amp;TestCategory!=Nightly</TestCaseFilter>
```
andrew-polk, line 39: [Devin] **Investigate**: Test Explorer discovery may hide the excluded categories

Claude Opus 5 (1M context), mirroring a Devin finding.

Devin's concern, in plain terms: wiring `RunSettingsFilePath` into the csproj means Visual Studio's Test Explorer reads this runsettings too, not just `dotnet test` on the command line. If a developer clicks Run on one of the Integration or Nightly tests in the IDE, VSTest may AND this file's `TestCaseFilter` with the IDE's own "run just this test" selection — in which case the test they explicitly asked for silently doesn't run, and the documented way to override (passing a filter) isn't available from a click.

Devin's words:

> Wiring `RunSettingsFilePath` in also affects Visual Studio's Test Explorer, whose runs go through the same runsettings. Depending on the VSTest version, a `RunConfiguration/TestCaseFilter` can suppress explicitly selected tests, so a developer clicking Run on an Integration or Nightly test in the IDE may see it silently not run with no obvious way to override (the documented opt-in is a CLI `--filter`). Worth confirming in the IDE before merging.

**Where this stands.** I could not settle this from a terminal — it needs someone to click Run on `S3RetryBudgetTests` (or a `BloomS3ClientTests` Integration case) in Test Explorer and see whether it runs or is silently skipped. The command-line behavior is verified in both directions; only the IDE path is open. Devin classed this Informational-adjacent (`needs_investigation`), and it is a real trade-off of the runsettings approach rather than a defect in the code, so it is going to the developer as a decision rather than being resolved here.

If it turns out the IDE does suppress explicit selections, the alternative is to drop the runsettings and set `VSTestTestCaseFilter` in the csproj instead: that keeps the command-line default identical but does not touch IDE runs at all — the cost being that Test Explorer then shows no default exclusion, so a click could start a live-S3 upload.

src/BloomTests/BloomTests.csproj
```diff
@@ -20,6 +20,24 @@
...
 		<IsTestProject>true</IsTestProject>
 		<Platforms>x64;ARM64</Platforms>
+		<!-- Gives `dotnet test <this project>` and the IDE test runners a default filter that
+		     leaves out the Integration and Nightly categories; see BloomTests.runsettings for why.
+
+		     The condition is what makes opting back in possible. VSTest ANDs a runsettings
+		     TestCaseFilter with any filter given on the command line (it does not replace it),
+		     so an unconditional setting here would make "TestCategory=Integration" resolve to
+		     "(TestCategory!=Integration)&(TestCategory=Integration)" and match nothing at all,
+		     making the Integration tests unrunnable. dotnet test surfaces its filter
```
andrew-polk, line 30: [Devin] **Investigate**: Command-line filter may take precedence over the runsettings filter rather than AND-ing

Claude Opus 5 (1M context), mirroring a Devin finding.

Devin questions the premise this whole condition rests on. It reads Microsoft's runsettings documentation as saying a command-line filter *replaces* a `RunConfiguration/TestCaseFilter` (runsettings value ignored), which would make the `VSTestTestCaseFilter` condition merely belt-and-braces rather than load-bearing — and would make the "any filter drops the default" note inaccurate.

Devin's words:

> The comment asserts VSTest ANDs a runsettings `TestCaseFilter` with a command-line `--filter`. Microsoft's runsettings documentation states the opposite for `RunConfiguration.TestCaseFilter` (the command-line `/TestCaseFilter` takes precedence, with the runsettings value ignored). If that is what current VSTest does, the `VSTestTestCaseFilter` condition is merely belt-and-braces rather than load-bearing — harmless either way, but the stated rationale (and the "all-or-nothing" consequence described at `src/BloomTests/BloomTests.csproj:34-39`) would then be inaccurate for readers who rely on it.

andrew-polk: Claude Opus 5 (1M context): not an issue — the docs Devin is reading are out of date, and the AND behavior is directly observable in this repo.

**VSTest printed the AND itself.** Running the opt-in on this project with the runsettings attached unconditionally, VSTest reported the effective filter as:

```
No test matches the given testcase filter
`(TestCategory!=Integration)&(TestCategory=Integration&FullyQualifiedName~BloomS3ClientTests)`
```

That is the runsettings filter conjoined with the command-line one, not replaced by it — and it is why the unconditional version made the Integration tests unrunnable rather than opt-in-able. That failure is exactly what led to the condition.

**And VSTest's source does it explicitly.** `TestCaseFilterArgumentProcessor.Initialize` composes the two rather than choosing between them:

```csharp
_commandLineOptions.TestCaseFilterValue = $"({defaultFilter})&({argument})";
```

with the runsettings value having been copied into `CommandLineOptions.TestCaseFilterValue` earlier, at `ArgumentProcessorPriority.RunSettings` (which runs before `Normal`).

So the condition is load-bearing, not belt-and-braces, and the "any filter drops the whole default" consequence documented in the csproj is accurate — it is the direct result of this AND. Both facts are also verified end-to-end on this branch: with a filter given, VSTest reports the command-line filter alone (`FullyQualifiedName~ZZZNoSuchTestName`, no `TestCategory!=` conjunct), and `--filter TestCategory=Nightly` runs the 2 Nightly tests that a default run leaves out.

Resolving, since the premise holds.

andrew-polk:
Claude Opus 5 (1M context): **Consulted Devin on 2026-08-06 11:36 USMST up to commit `2e009716e7c4620147b1c5097786e568b68473ed`.**

Two rounds, both complete for the commit they ran against.

**Round 1** (`6907bef2`) — no bugs, 1 Investigate flag, 3 Informational. All three Informational items were fair and I fixed them: a runsettings comment that still claimed nothing carried the `Nightly` category (stale after this branch was rebased onto the master that tagged `S3RetryBudgetTests`), a missing note that *any* command-line filter drops the whole default and not just the category part, and a comma guard that inspected the escaped value so a `%2C` could slip past it. That last one is now a real guard: `/p:excludedCategories=Integration%2CNightly` fails loudly instead of quietly excluding nothing.

**Round 2** (`2e009716`) — no bugs, 2 Investigate flags, 5 Informational. Three of the Informational items are verbatim carry-forwards of round 1's, still quoting the pre-fix text, so they are already addressed. Of the two genuinely new ones: the observation that `Integration` tests now run only in the nightly is true and intended (TeamCity's PR build already excluded them), and the suggestion to assert the split item count is a reasonable hardening idea that goes to the developer rather than being done unasked.

The two Investigate flags each have their own thread:

- **Command-line filter may take precedence rather than AND-ing** — refuted and resolved. VSTest prints the conjunction itself, and its source composes `({defaultFilter})&({argument})`; the docs Devin cited are out of date.
- **Test Explorer discovery may hide the excluded categories** — left **open**. It needs someone to click Run on an excluded test in Visual Studio, which I cannot do from a terminal, so it is in the developer's decision report.

Other reviewers: this repo runs no GitHub test checks on PRs (TeamCity is the PR gate and reports outside GitHub), and CodeRabbit is present but configured `auto_review.enabled: false`, so it does not review here. Note also that `pr-automation.yml` did **not** fire for this PR, so Devin was triggered manually and the review link added by hand.

src/BloomTests/BloomTests.runsettings
andrew-polk: Claude Opus 5 (1M context): recording the developer's decision — **leaving this as is, no change to the mechanism.**

Andrew's answer: "We don't use Visual Studio anymore. So no changes needed. You could add a one-sentence comment stating we aren't sure if this breaks that."

So the concern is real but no longer reaches anyone: the team develops outside Visual Studio, which makes Test Explorer's behaviour with this runsettings moot in practice. Rather than change the mechanism or leave the question dangling for a future reader, the uncertainty is now written down next to the filter itself, in `src/BloomTests/BloomTests.runsettings`:

> Untested: whether clicking Run on an excluded test in Visual Studio's Test Explorer runs it or silently skips it, since VSTest may combine this filter with the IDE's selection. We no longer develop in Visual Studio, so nobody has checked.

That way anyone who does open the solution in Visual Studio later finds the caveat before being surprised by it. Resolving.

.github/workflows/harvester-artifacts.yml
```diff
@@ -172,14 +172,27 @@ jobs:
...
+            # comment on CategoriesToExclude in Bloom.proj for the two forms that do work.
             - name: Run C# tests
               id: csharp_tests
               working-directory: build
               shell: pwsh
               run: |
                   msbuild Bloom.proj /t:TestOnly `
                     /p:BUILD_NUMBER=$env:BUILD_NUMBER `
-                    /p:excludedCategories=SkipOnTeamCity
+                    /p:excludedCategories=Integration%3BNightly%3BSkipOnTeamCity
```
andrew-polk, line 195: [Devin] **Investigate**: Nightly still runs the Integration tests, which need live S3 credentials

Claude Opus 5 (1M context), mirroring a Devin finding.

Devin's concern: now that developers and the harvester both skip the `Integration` category, the nightly is the only runner left that exercises the tests which talk to the real S3 bucket — so if that environment has no credentials for it, those tests would be failing there and effectively nothing would be verifying S3 at all.

Devin's words:

> With the harvester workflow adding `Integration` to its exclusions, the only automated runner left that exercises the live-S3 `Integration` tests on GitHub is the nightly (TeamCity's PR build also excludes them).

andrew-polk: Claude Opus 5 (1M context): checked rather than assumed — **not an issue: the nightly does run them, and they pass.**

Evidence from the most recent nightly (run `31123627983`, commit `83aae362`), taking the `TestResults.xml` it uploads as an artifact rather than reasoning from the summary counts:

```
DownloadBook_DoesNotExist_Throws                 Passed
BloomS3StandardUpDownloadTests                   Passed
UploadDownloadStandardBook_FilesAreInExpectedDirectory  Passed
SmokeTest_DownloadKnownBookFromProductionSite    Passed
ProblemBookUploaderTests                         Passed
UploadSmokeTest                                  Passed
```

That run reported `3026 passed / 13 skipped / 0 failed`. So the live-S3 tests are genuinely exercised there, with working credentials — the nightly environment has what they need.

Worth stating plainly what did and did not change here, since the flag reads as though this PR reduced S3 coverage: TeamCity's PR build already excluded `Integration` before this branch, and the nightly already ran it. What this PR changes is that **developers** stop running it by accident, and the harvester stops running it deliberately. The nightly was always where these ran, and still is — that was the explicit intent ("exclude Integration for harvester, keep it for nightly").

The real consequence is timing, not coverage: an S3-facing break now surfaces in the nightly rather than on the artifact build. That trade-off was put to the developer as its own decision and accepted. Resolving.

src/BloomTests/BloomTests.csproj
andrew-polk, line 30: [Devin] **Investigate**: Any command-line filter now silently drops the Integration/Nightly default

Claude Opus 5 (1M context), mirroring a Devin finding.

Devin's concern: because the runsettings is attached only when no filter was given, *any* filter removes the whole default rather than just the part being overridden. So a developer narrowing a run by fixture name — the pattern `AGENTS.md` itself recommends — is back to running that fixture's live-S3 `Integration` tests.

Devin's words:

> Because the runsettings is attached only when `VSTestTestCaseFilter` is empty, ANY filter — including a name filter like the one AGENTS.md recommends — drops the whole runsettings, not just the category exclusion. So a developer filtering by fixture name (e.g. `FullyQualifiedName~BloomS3ClientTests`) will still run the live-S3 Integration tests, which need internet access and bucket credentials.

andrew-polk: Claude Opus 5 (1M context): correct, and already acted on — Devin raised this as Informational on the previous round and it was documented in `66cdeb02`, one commit before this one. Raising it again on the same code is a duplicate rather than a new finding.

It is a genuine consequence and cannot be designed away: VSTest ANDs a runsettings filter with a command-line one, so the only way to make `--filter TestCategory=Integration` able to *select* those tests is to withhold the runsettings whenever a filter is present. That makes it all-or-nothing by construction. The honest response is to tell the reader, which the csproj comment now does:

> Consequence worth knowing: this is all-or-nothing, so ANY filter drops the whole default, not just the part you were overriding. A run narrowed by name (a filter of "FullyQualifiedName~BloomS3ClientTests", say) is therefore back to including that fixture's Integration tests, live S3 bucket and all. Add TestCategory!=Integration to such a filter yourself if you want it.

Note the practical exposure is smaller than it sounds: a name filter only re-admits `Integration` tests that match that name, and on a developer machine those tests pass against the unit-test bucket. Resolving, since the behaviour is intended and now written down where someone narrowing a run will see it.

andrew-polk:
Claude Opus 5 (1M context): **Consulted Devin on 2026-08-06 12:04 USMST up to commit `019c8b974675229f9d80e6714b9365e60b140470`.** Third round; complete for that commit.

No bugs. 4 Investigate flags and 4 Informational. Two of the Investigate flags were ones already posted and settled in earlier rounds (Test Explorer, filter precedence), so they were not re-posted. The two live ones each got a thread and are resolved:

- **Nightly still runs the Integration tests, which need live S3 credentials** — checked rather than assumed. The most recent nightly's own `TestResults.xml` artifact shows those tests ran and passed there (`3026 passed / 13 skipped / 0 failed`), so the credentials exist in that environment. Coverage did not shrink either: TeamCity's PR build already excluded `Integration` before this branch, and the nightly already ran it.
- **Any command-line filter now drops the whole default** — correct, and already documented one commit earlier in `66cdeb02` after Devin raised it as Informational last round. It is inherent to the AND-vs-replace workaround, so the response is to tell the reader, which the csproj comment now does.

Of the Informational items, two are stale carry-forwards of things fixed in earlier rounds. The other two: an empty `excludedCategories` is harmless (no items, so no filter is emitted at all, which is the pre-existing behaviour), and the new malformed-category guard being unable to fire while the unescaping above it works is the point of an assertion rather than dead code — with the unescape deliberately bypassed it does fire, which is how it was verified.

Local checks at this commit: C# `3031 passed / 12 skipped / 1 failed`, the one failure being the pre-existing `RabRealBuildTests` case that needs Reading App Builder installed and fails on `master` too; front-end typecheck, lint and 664 vitest tests green.

PR #8167, opened 2026-08-06
Let external tools preview any branding, layout and xmatter

hatton:
BloomDesktop side of **https://github.com/BloomBooks/branding-viewer**, a standalone tool that screenshots a book across branding × page × layout × xmatter so branding work can be reviewed as a whole instead of one collection at a time.

```
GET  /bloom/api/external/branding-options   the brandings, layouts and xmatter packs THIS build
                                            ships, plus its version and current state
POST /bloom/api/external/select-branding    {branding, layout, xmatter}; omitted = unchanged
```

## Why this branch and not master

The branding work is happening here, and a tester runs a beta, so endpoints that existed only on master would reach nobody. It also turns out this is the branch that *can* carry it: `--e2e` and `E2eTestingApi` are part of the 879 commits master has and this branch does not, so an earlier attempt built on those (#8161, closed) would have needed two implementations. `BrandingPreviewApi` depends only on `CollectionSettings`, `BookSelection` and `XMatterPackFinder`, all of which exist here, so the same file forward-merges to master unchanged.

## No flag, deliberately

Registered unconditionally, in the `external/` url space alongside `ExternalApi` — which is already unconditional and already documented as "API functions which are called from outside of Bloom".

An earlier version gated this behind `--e2e`. That was wrong twice over: it put a launch argument in front of the people least equipped to reconstruct one, and it tied a branding feature to a class named for end-to-end testing.

**What it costs.** Branding normally flows from a checksum-validated subscription code and `select-branding` bypasses that. Because bringing a book up to date stamps branding files into the book folder and saves, a book can end up wearing a branding nobody holds a subscription for. This is called out in the class comment rather than left implicit. Cutting the other way: every branding pack already ships inside every Bloom, and a determined person can edit a book's `branding.css` by hand, so this is convenience rather than a lock.

## Three things worth a reviewer's attention

- **`select-branding` catches its own exceptions** and reports via `request.Failed()`, unlike its neighbours. A survey walks hundreds of combinations and some are genuinely invalid (a branding whose xmatter this collection does not offer, a layout the book's stylesheets do not support). One bad combination must not take Bloom down or wedge the caller's run.
- **`branding-options` returns an empty `layouts` list until a book is selected**, because Bloom derives layout choices from the selected book's stylesheets. Documented in the XML comment; the caller asks again after selecting.
- **`Program.cs` change, outside the strict scope of "add an API".** The DEBUG-only "Attach debugger now" prompt fired for *any* argument, so launching a Debug build on a chosen collection blocked forever on an invisible modal and no script or external tool could drive it. It now skips when the only non-switch argument is a `.bloomCollection` path, which is not the URL-interpretation case that prompt exists for. Master sidesteps this via a `--e2e` check that does not exist here. Happy to split this into its own PR if you'd rather.

## Verified

Built here, launched `Bloom.exe <collection> --automation` with **no other flag**, on a throwaway copy of `src/BloomVisualRegressionTests/collections/basic`:

- `branding-options` → `6.4.0.0`, 73 brandings, the 5 offered xmatter packs, current state.
- The viewer captured WorldVision, Local-Community and Default back covers at Device16x9Landscape, 3/3, then restored the baseline.
- It ran alongside another worktree's Bloom without interfering, and from a clone with no BloomDesktop checkout.

## Not included

No automated test covers the endpoints; `ExternalApi` has none today either, and these exist to be driven from outside the repo. The `#if DEBUG` branding override in `CollectionSettingsApi` is left alone, since this branch's visual-regression suite drives it; it becomes redundant and can go in a follow-up.

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8167)

### DIFF
--- a/src/BloomExe/Program.cs
+++ b/src/BloomExe/Program.cs
@@ -356,7 +356,7 @@ namespace Bloom
                 {
                     if (IsLocalizationHarvestingLaunch(args))
                         LocalizationManager.IgnoreExistingEnglishTranslationFiles = true;
-                    else
+                    else if (!IsPlainCollectionLaunch(args))
                         // This allows us to debug things like  interpreting a URL.
                         MessageBox.Show("Attach debugger now");
                 }
@@ -1301,6 +1301,23 @@ namespace Bloom
             return args.Length == 1
                 && args[0].StartsWith("--ha")
                 && "--harvest-for-localization".StartsWith(args[0]);
+        }
+
+        /// <summary>
+        /// True if the arguments are just "open this collection" (a .bloomCollection path, possibly
+        /// with switches), rather than something like a URL to interpret.
+        ///
+        /// This exists so the DEBUG-only "Attach debugger now" prompt does not fire for it. That
+        /// prompt is there to let a developer attach before Bloom interprets a URL, but it triggers
+        /// on ANY argument, so simply launching a Debug build on a chosen collection blocks forever
+        /// on a modal dialog. That makes it impossible to drive Bloom from a script or an external
+        /// tool (see BrandingPreviewApi) without a human clicking OK.
+        /// </summary>
+        private static bool IsPlainCollectionLaunch(string[] args)
+        {
+            var nonSwitches = args.Where(a => !a.StartsWith("-")).ToArray();
+            return nonSwitches.Length == 1
+                && nonSwitches[0].EndsWith(".bloomCollection", StringComparison.OrdinalIgnoreCase);
         }
 
         // I think this does something like the Wix element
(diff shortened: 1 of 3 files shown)


hatton:
[Claude Opus 5 following a prompt from Hatton] Second commit changes the `Program.cs` part of this PR.

Rather than the narrower guard the first commit had, the DEBUG-only `MessageBox.Show("Attach debugger now")` is simply **commented out**, with a note saying why and that uncommenting two lines restores it. That matches how this file already treats the same call (there is a commented-out copy of it earlier).

The reason it has to go: it fires for *any* argument, not just the URL case it exists for, so a Debug build launched on a collection path blocks forever on a modal dialog — alive and responding, no window, no stdout, no server, which reads as a hang. Nothing scripted can get past it, including this branch's own visual-regression suite, which does `execFile(Bloom.exe, [collectionPath])` at `index.spec.ts:186`.

DEBUG-only, so no shipped build is affected.

PR #8168, opened 2026-08-06
Wait for the visual-regression book preview to be ready, instead of sleeping

andrew-polk:
## What

The visual-regression suite's **book-preview** capture was preceded by a blind
`await new Promise(r => setTimeout(r, 1000))`. That raced the two things which make the preview
render differently from run to run: the web fonts (text metrics and line breaking change the
instant the real face arrives) and the images.

Nightly run 31077057359 lost that race on its very first case, `16x9 Landscape branding:Default`:
every image was missing (Bloom drew its *"This image, image1.png, is missing or was loading too
slowly"* alt text instead) and the title fell back to a serif. All the failure told us was
*"167,813 pixels differ"*. The same commit passed on re-run, so it was a race, not a regression.

Both capture paths — the book preview and the bloom-player pages — now wait for
`document.fonts.ready`, wait for every picture on the thing being screenshotted to finish loading,
let layout settle for two frames, and then **refuse to screenshot** if any picture or font ended in
an error state, reloading up to `MAX_CAPTURE_ATTEMPTS = 3` first.

## The decision: settle *and* assert

Settling alone makes the capture **deterministic** — it happens once the browser has finished
trying — but not **correct**: an image or font that ended in an *error* state is settled too.
`document.fonts.ready` alone would have fixed the font half of that nightly failure while the
missing-image half still produced a wrong screenshot and the same opaque pixel count.

So the shared helper also collects every picture that finished with nothing in it and every
`FontFace` left in the `"error"` state, and the caller reloads rather than photograph a page that is
missing its illustrations.

**This is a deliberate behaviour change to the suite, called out rather than slipped in:** a book
whose image genuinely never arrives no longer surfaces as a pixel diff but as
`image did not load: image1.png`. The retry heals the transient case *without ever accepting a wrong
render* — if the asset really is missing, the run still fails, just legibly. A book with a
permanently missing image can no longer pass, which is the intended contract.

One trap worth recording: Bloom puts that *"is missing or was loading too slowly"* string in the
`alt` attribute of **every** image, loaded or not (`bloomImages.ts::SetAlternateTextOnImages`), so
searching the DOM for it proves nothing. `naturalWidth` is the signal.

## The player half, and what extracting it turned up

The first commit did the preview only. Applying the same contract to the bloom-player captures
(second commit, at review request) exposed something worth knowing: **`waitForActivePageReady()`
was awaiting an empty list.** It awaited every `` on the active slide — but bloom-player has
none. It paints a page's illustration as a CSS `background-image` on a div; real `` elements
are what the *preview* uses. Verified directly: the active slide reports zero `` and one
background-image URL, pointing at the staged, re-encoded picture (`image1.png` is staged as
`image1_r.jpg`).

A background image exposes no load state to the DOM at all, so the only way to wait for one — or
notice it failed — is to request the same URL ourselves and watch that; the browser has already
fetched it, so it resolves from the cache. We do that for every element that actually occupies
space, skipping hidden ones so we cannot invent a failure that never affected the render.

## Evidence

Each direction was proved, not assumed:

- **The preview assertion fires.** Truncating `16x9 Landscape/image1.png` to zero bytes produced
  `The book preview never rendered correctly (3 attempts) … image did not load: image1.png`
  instead of a pixel diff. Reverted after.
- **The player assertion fires.** Aborting the staged illustration's request on the player page
  alone used to produce a **156,821-pixel diff**; it now fails with
  `background image did not load: …/image1_r.jpg` after three attempts.
- **It reproduced and healed the original race.** On a cold Bloom the suite logged
  `book-preview attempt 1 was not usable (image did not load: image1.png; image did not load:
  lang-qr-code.png; font did not load: Andika; font did not load: Andika; font did not load: Andika
  New Basic); reloading.` — the QR code and all three Andika faces missing on the first load, which
  is exactly the nightly failure. The reload fixed it.
- **No false positives and no cost** from the new background-image scan: consecutive full-suite
  runs green at ~138 s.
- **Not just a longer sleep.** The fixed second is gone, not lengthened: ~156 s per run before this
  PR, ~138 s after.
- No reference images were regenerated or accepted.

## A separate, pre-existing flake found on the way (NOT addressed here)

Twice during verification, a bloom-player case failed with a large illustration rendered with ~1px
haloing on every edge — a different source raster, not a layout or font difference. I ruled out the
two obvious causes:

- **Not captured too early.** Probing the pre-change code, all **72** player captures in a run were
  byte-identical at t0, t+500 ms and t+2000 ms after the readiness wait returned. The raster is
  stable within a run; it differs *between* runs.
- **Not GraphicsMagick failing.** Bloom's image-conversion errors during BloomPUB staging show no
  correlation (the failing run had 0; a passing run had 1).

The likeliest remaining lead is the staging step itself: the player renders a **re-encoded and
resized** copy of each picture (`image1.png` → `image1_r.jpg`), so if that conversion is not
byte-deterministic between runs, the player renders a slightly different bitmap. That is being
investigated separately, on its own branch; it is not what this PR fixes, and the new assertion
would not catch it (the picture loads fine — it is just drawn differently).

## Notes

- No YouTrack card — that was a deliberate decision for this change.

---
[Devin review](https://app.devin.ai/review/BloomBooks/BloomDesktop/pull/8168)

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8168)

### DIFF
--- a/src/BloomVisualRegressionTests/index.spec.ts
+++ b/src/BloomVisualRegressionTests/index.spec.ts
@@ -48,6 +48,124 @@ let bloomExit: { code: number | null; signal: NodeJS.Signals | null } | null =
 // instanceInfo and kill that too.
 let bloomServingPid: number | null = null;
 
+// How many times we will load a page looking for a render that is complete enough to screenshot.
+// More than one because the first request can catch Bloom still busy with the book; bounded
+// because a book whose image really is missing must fail, not retry forever.
+const MAX_CAPTURE_ATTEMPTS = 3;
+
+// The active bloom-player slide: the one page element the player captures screenshot.
+const ACTIVE_PLAYER_PAGE = ".swiper-slide-active .bloom-page";
+
+// Wait until what we are about to screenshot has actually finished rendering — not merely
+// appeared — and report anything that came out wrong. The two things that otherwise render
+// differently from run to run are the web fonts (both Bloom and bloom-player load Andika
+// asynchronously, and text metrics and line breaking change the instant the real face arrives)
+// and the images; a fixed timeout raced both. So wait for document.fonts.ready, for every image
+// to finish, and then for one more layout frame.
+//
+// Waiting makes the capture DETERMINISTIC (it happens once the browser has finished trying) but
+// not necessarily CORRECT: an image or font that ended in an ERROR state is settled too. So we
+// also report those, rather than screenshot a page that is missing its pictures or has fallen
+// back to a substitute font. A finished <img> with a src but no intrinsic width did not load;
+// Bloom then renders its alt text ("This image, X, is missing or was loading too slowly."), which
+// is exactly the wrong render we must never accept. (Note that alt text is on EVERY image, loaded
+// or not, so its presence proves nothing — only naturalWidth does.)
+//
+// `rootSelector` scopes which images matter: the whole document for the book preview, or just the
+// active slide for a bloom-player capture. Fonts are always document-wide. Returns a list of
+// problems; empty means the page is good to capture.
+//
+// This is handed to page.evaluate() and therefore runs INSIDE the browser, so it must be entirely
+// self-contained: it cannot call anything declared outside itself.
+async function settleAndReportProblems(
+    rootSelector: string | null,
+): Promise<string[]> {
+    const g = globalThis as any;
+    const doc = g.document;
+    await doc.fonts.ready;
+    const root = rootSelector ? doc.querySelector(rootSelector) : doc;
+    const problems: string[] = [];
+    if (rootSelector && !root) {
+        problems.push(`${rootSelector} was not present`);
+        return problems;
+    }
+
+    // Pictures reach the screen two different ways, and we have to handle both. The book preview
+    // uses real <img> elements. bloom-player does NOT: it paints a page's illustration as a CSS
+    // background-image on a div, so the active slide contains no <img> at all and an image-only
+    // check would silently pass on an empty list.
+    const imgs: any[] = Array.from(root.querySelectorAll("img"));
+    await Promise.all(
+        imgs.map((img: any) =>
+            img.complete
+                ? Promise.resolve()
+                : new Promise((resolve) => {
+                      img.addEventListener("load", resolve, { once: true });
+                      img.addEventListener("error", resolve, { once: true });
+                  }),
+        ),
+    );
+
+    // A background image exposes no load state to the DOM, so the only way to wait for one — or to
+    // notice that it failed — is to ask for the same URL ourselves and watch that. The browser has
+    // already requested it, so this normally resolves straight from the cache. Only look at
+    // elements that actually occupy space: a hidden element's background is never fetched for the
+    // render, so probing it could invent a failure that does not affect the screenshot.
+    const backgroundUrls = new Set<string>();
+    for (const el of Array.from(root.querySelectorAll("*")) as any[]) {
+        const box = el.getBoundingClientRect();
+        if (box.width === 0 || box.height === 0) continue;
+        const background = g.getComputedStyle(el).backgroundImage;
+        if (!background || background === "none") continue;
+        for (const match of background.matchAll(/url\((["']?)(.*?)\1\)/g)) {
+            const url = match[2];
+            if (url && !url.startsWith("data:")) backgroundUrls.add(url);
+        }
+    }
+    const failedBackgrounds: string[] = [];
+    await Promise.all(
+        Array.from(backgroundUrls).map(
+            (url) =>
+                new Promise((resolve) => {
+                    const probe = new g.Image();
+                    probe.addEventListener("load", resolve, { once: true });
+                    probe.addEventListener(
+                        "error",
+                        () => {
+                            failedBackgrounds.push(url);
+                            resolve(null);
+                        },
+                        { once: true },
+                    );
+                    probe.src = url;
+                }),
+        ),
+    );
+
+    // Fonts/images are in; let layout settle for two frames before we capture.
+    await new Promise((resolve) =>
+        g.requestAnimationFrame(() => g.requestAnimationFrame(resolve)),
+    );
+
+    for (const img of imgs) {
+        const src = img.getAttribute("src");
+        if (src && img.naturalWidth === 0)
+            problems.push(`image did not load: ${src}`);
+    }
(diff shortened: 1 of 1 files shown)


src/BloomVisualRegressionTests/index.spec.ts
```diff
@@ -204,25 +208,94 @@ describe("All books", () => {
...
+            }
+            for (const font of Array.from(doc.fonts) as any[]) {
+                // Include the weight/style: a family is several FontFaces, and knowing which one
+                // failed is the difference between "the font server is down" and "we asked for a
+                // bold that does not exist".
+                if (font.status === "error")
+                    problems.push(
+                        `font did not load: ${font.family} ${font.style} ${font.weight}`,
+                    );
+            }
```
andrew-polk: [Devin] **Investigate**: Font 'error' assertion covers all declared faces in the document

The check iterates the whole `document.fonts` set, not just the faces the book actually uses. A FontFace only reaches `status === "error"` after an attempted load fails, so unused declarations stay `"unloaded"` and are ignored. However, on a machine/CI image where a book's requested font genuinely isn't available from Bloom's font server, this now turns what used to be a (possibly accepted) fallback-font render into a hard failure after 3 attempts. That is consistent with the PR's stated "never accept a wrong render" contract, but it is worth confirming CI has the same fonts the reference images were generated with.

andrew-polk: [Claude Opus 5 (1M context)] Not an issue — CI cannot have different fonts here, by construction rather than by luck.

The preview's fonts don't come from the machine at all. Bloom serves them to itself: `BloomServer.cs` intercepts `book-preview/defaultLangStyles.css` and prepends `FontServe.GetAllFontFaceDeclarations()`, whose faces are the `.woff2` files shipped in the repo at `DistFiles/fonts/` (`Andika-Regular/Bold/Italic/BoldItalic`, `ABeeZee-Regular/Italic`). They are all `url()` sources — none use `local()`, so nothing is ever satisfied from an installed system font. The test collection asks for `<DefaultLanguage1FontName>Andika`, which is one of those shipped files.

So the font a CI run gets is the same file, out of the same checkout, served by the same Bloom binary under test as the one that produced the reference images. A CI image lacking a font can't produce the fallback render this flag is about; the only way these faces fail is the transient one this PR is fixing (which is exactly what the reproduction showed — all three Andika faces in `error` on a cold Bloom's first load, healed by the reload).

Resolving.

andrew-polk:
[Claude Opus 5 (1M context)] Consulted Devin on 2026-08-06 up to commit `963f4a7659e1cf1bd95525869db192a8d014ba89`.

It found **no bugs**. It raised **1 Investigate** flag and 3 informational items.

The Investigate flag asked whether CI has the same fonts the reference images were made with — it does, necessarily: Bloom serves the preview's fonts to itself from `DistFiles/fonts/` in the checkout, so there is no machine-installed font involved. Assessed and resolved in [its own thread](https://github.com/BloomBooks/BloomDesktop/pull/8168#discussion_r3731472269).

The 3 informational items (no back-off between retries; the `naturalWidth` check assuming intrinsic image dimensions; the in-page image waits having no timeout of their own) are all noted rather than mirrored. Each was already a conscious choice, and the second and third Devin itself checked and found no false positive or hang reachable today.

src/BloomVisualRegressionTests/index.spec.ts
```diff
@@ -369,42 +505,52 @@ describe("All books", () => {
...
+                throw new Error(
+                    `bloom-player page ${n} never rendered correctly (${MAX_CAPTURE_ATTEMPTS} attempts). ` +
+                        `The last attempt had these problems:\n  ${problems.join("\n  ")}`,
+                );
+            console.log(
+                chalk.yellow(
+                    `bloom-player page ${n} attempt ${attempt} was not usable (${problems.join("; ")}); reloading.`,
+                ),
+            );
+        }
```
andrew-polk, line 553: [Devin] **Investigate**: Player capture path now hard-fails on missing images/fonts, contrary to the PR description

The PR description says the assertion is deliberately NOT folded into the bloom-player path ("folding the assertion into the player path would be a behaviour change beyond this change's scope"), but `loadPlayerPageAndWaitUntilReady` does exactly that: it evaluates the same `settleAndReportProblems` on the active slide and throws after `MAX_CAPTURE_ATTEMPTS` if any image/background/font problem is reported. That is a real behaviour tightening for the player captures (previously they were only settled, never asserted). The in-code comment acknowledges it; the PR description does not. Worth confirming that this is what the author intends to land, especially given the description also mentions a pre-existing player raster flake.

andrew-polk: [Claude Opus 5 (1M context)] Correct catch, and the fix is to the description — the code is what we want.

The player-side assertion is intentional. The first commit deliberately scoped the change to the preview, and the description was written then; Andrew reviewed it and asked to make both sides consistent, so the second commit applies the same contract to the player captures. I didn't update the description to match. Done now.

Worth adding, since it changes how much of a tightening this really is: extracting the shared helper showed that `waitForActivePageReady()` had been awaiting an **empty list**. It awaited every `` on the active slide, but bloom-player has none — it paints the illustration as a CSS `background-image` on a div. So the player captures were not merely 'settled but not asserted'; for their main picture they were not being waited on at all. The new code probes the background-image URL, which is what makes both the wait and the assertion real there for the first time. Aborting that request used to yield a 156,821-pixel diff and now names the file.

On the raster flake also mentioned in the description: the new assertion does not touch it, and I've said so explicitly there — the picture loads fine, it is just drawn slightly differently. It is being investigated separately.

Resolving.

andrew-polk:
[Claude Opus 5 (1M context)] Consulted Devin on 2026-08-06 up to commit `411bcb94b85af09b5940b088e412e4fdccc0643e` (re-review after the player-side commit).

Still **no bugs**. 2 Investigate flags and 4 informational items.

One flag was new and correct: the PR description still claimed the assertion was deliberately kept out of the bloom-player path, which stopped being true when that was added at review request. The description has been rewritten to match what actually landed — [thread](https://github.com/BloomBooks/BloomDesktop/pull/8168#discussion_r3731780332).

The other flag was the font one from the previous round, re-raised verbatim; it is already [answered and resolved](https://github.com/BloomBooks/BloomDesktop/pull/8168#discussion_r3731472269), so it was not posted a second time.

PR #8169, opened 2026-08-06
Keep page-thumbnail image srcs URL-encoded (BL-16658)

JohnThomson:
Fixes the 6.5 regression where an image whose file name contains characters like `#` or `%` shows as broken in edit mode's page-list (thumbnail) view.

### What was wrong

`PageListApi.MarkImageNodesForThumbnail` builds the thumbnail src correctly as `<url-encoded path> + "?thumbnail=1"`, and then throws the encoding away:

```csharp
var filename = imageElementUrl.PathOnly.UrlEncoded;   // "This%20Image%20!%40%23%24%25%5e%26()2.jpg"  OK
var url = filename + "?thumbnail=1";                  // exactly what we want                        OK
HtmlDom.SetImageElementUrl(
    imgNode,
    UrlPathString.CreateFromUnencodedString(url),     // sees %XX and decodes the whole thing
    urlEncode: false);                                // then writes the decoded form out
```

`CreateFromUnencodedString` decodes any string containing `%XX`, so the raw file name landed in the attribute. The browser then read the `#` as the start of a fragment and requested a truncated path. The log attached to BL-16658 shows exactly that:

```
**BloomServer: File Missing: C:/.../Image with Strange Characters/This Image !@
```

The regression came in with 5d34c530f9 ("Optimize page thumbnail loading", BL-16336), which added `urlEncode: false`. Before that, `urlEncode: true` re-encoded on the way out and so restored the escapes — at the cost of also encoding the `?` as `%3f`, which is what BL-16336 was fixing.

### The fix

Pass `strictlyTreatAsEncoded: true`, so the string is neither decoded on the way in nor re-encoded on the way out. Both flags are needed, and each alone breaks the other case.

### Tests

New `PageListApiTests` covers a plain name, a name with special characters, a name that already has a query, an `/api/` url, and the `background-image` variant. Verified fix-sensitive: reverting the one-line change fails three of the five with exactly the reported symptom.

`MarkImageNodesForThumbnail` becomes `internal` so the tests can reach it; `BloomTests` already has `InternalsVisibleTo`.

Ref: https://issues.bloomlibrary.org/youtrack/issue/BL-16658

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8169)

---
[Devin review](https://app.devin.ai/review/BloomBooks/BloomDesktop/pull/8169)

### DIFF
--- a/src/BloomExe/web/PageListApi.cs
+++ b/src/BloomExe/web/PageListApi.cs
@@ -372,7 +372,12 @@ namespace Bloom.web
                     if (imageElementUrl.NotEncoded.Contains("/api/"))
                         continue;
 
-                    var filename = imageElementUrl.PathOnly.UrlEncoded;
+                    // UrlEncodedForHttpPath rather than UrlEncoded, because the latter escapes the
+                    // path separators themselves ('/' as %2f, ':' as %3a). For an ordinary book
+                    // image the two are identical -- a Windows file name can contain neither
+                    // character -- but for a src that really is a path (say "/bloom/branding/x.png",
+                    // which the "/api/" test above doesn't catch) only this form stays requestable.
+                    var filename = imageElementUrl.PathOnly.UrlEncodedForHttpPath;
                     if (!string.IsNullOrWhiteSpace(filename))
                     {
                         var url = filename + "?thumbnail=1";
--- a/src/BloomExe/web/PageListApi.cs
+++ b/src/BloomExe/web/PageListApi.cs
@@ -358,7 +358,7 @@ namespace Bloom.web
 
         // As a further form of optimization, mark img elements as being thumbnails. The server
         // produces miniatures that take up less memory.
-        private static void MarkImageNodesForThumbnail(SafeXmlElement pageElementForThumbnail)
+        internal static void MarkImageNodesForThumbnail(SafeXmlElement pageElementForThumbnail)
         {
             var imgNodes = HtmlDom.SelectChildImgAndBackgroundImageElements(
                 pageElementForThumbnail
@@ -384,11 +384,20 @@ namespace Bloom.web
                             // gets interpreted as part of the filename.
                             url = filename + query.NotEncoded + "&thumbnail=1";
                         }
-                        // filename is already URL-encoded; we must NOT re-encode the whole string or ? becomes
-                        // %3f and lands in the path, making the thumbnail param invisible to GetQueryParameters().
+                        // At this point url is exactly what we want in the src attribute: an
+                        // already-URL-encoded path followed by a literal (unencoded) query. So we
+                        // must neither encode nor decode it on the way out (BL-16658).
+                        // - urlEncode:false, because re-encoding the whole string would turn the ?
+                        //   into %3f and bury the thumbnail param in the path, invisible to
+                        //   GetQueryParameters().
+                        // - strictlyTreatAsEncoded:true, because otherwise CreateFromUnencodedString
+                        //   sees the %XX escapes in the path and helpfully decodes them, and we would
+                        //   then write out a raw filename like "This Image !@#$%^&()2.jpg?thumbnail=1",
+                        //   where the browser treats the # as the start of a fragment and never
+                        //   requests the real file.
                         HtmlDom.SetImageElementUrl(
                             imgNode,
-                            UrlPathString.CreateFromUnencodedString(url),
+                            UrlPathString.CreateFromUnencodedString(url, true),
                             urlEncode: false
                         );
                     }
--- a/src/BloomBrowserUI/bookEdit/js/bloomImages.ts
+++ b/src/BloomBrowserUI/bookEdit/js/bloomImages.ts
@@ -1182,6 +1182,25 @@ export function SetupMetadataButton(parent: HTMLElement) {
     }
 }
 
+// Turn the src of an image into the file name we want to show a human in the alt text.
+// Strips any query string, then undoes the URL encoding.
+//
+// decodeURIComponent, not decodeURI: decodeURI deliberately leaves the characters that are
+// reserved in a URL (@ # $ & + , / : ; = ?) still escaped, so a file name like
+// "This Image !@#$%^&().jpg" came out as the half-decoded "This Image !%40%23%24%^%26().jpg".
+// It is a file name here, not a URL, so every escape should be undone.
+export function getDisplayNameFromImageUrl(rawImageUrl: string): string {
+    const nameWithoutQueryString = rawImageUrl.split("?")[0];
+    try {
+        return decodeURIComponent(nameWithoutQueryString);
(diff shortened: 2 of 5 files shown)


src/BloomExe/web/PageListApi.cs
```diff
@@ -372,7 +372,12 @@ private static void MarkImageNodesForThumbnail(SafeXmlElement pageElementForThum
                     if (imageElementUrl.NotEncoded.Contains("/api/"))
                         continue;
 
-                    var filename = imageElementUrl.PathOnly.UrlEncoded;
+                    // UrlEncodedForHttpPath rather than UrlEncoded, because the latter escapes the
```
JohnThomson, line 375: [Devin] **Investigate**: Slash/colon in an image src now stays percent-encoded in the thumbnail src

`filename` comes from `imageElementUrl.PathOnly.UrlEncoded`, and `UrlPathString.UrlEncoded` (`src/BloomExe/UrlPathString.cs:63-78`) percent-encodes `/` as `%2f` and `:` as `%3a` (that's why the class also offers `UrlEncodedForHttpPath`). Previously `CreateFromUnencodedString(url)` decoded the whole string, so those separators were restored before being written out with `urlEncode:false`; with `strictlyTreatAsEncoded:true` they are now preserved as `%2f`/`%3a`. For ordinary book images (bare file names in the book folder) this is a no-op, and true `/api/` urls are skipped a few lines above, so I did not flag it as a bug. But any page image whose src contains a path separator or scheme (e.g. a `/bloom/...` or `data:` src that isn't caught by the `/api/` filter) would now be requested with the separators escaped. If such srcs are possible, `UrlEncodedForHttpPath` would be the safer source for `filename`.

JohnThomson: [Claude Opus 5 from Thomson's machine during preflight] Fixed in `cfe19be4d0`.

Good catch — this was the one real behavioral difference the fix introduced, and it's now closed rather than argued away. `filename` is taken from `UrlEncodedForHttpPath` instead of `UrlEncoded`, so the path separators stay usable.

Worth recording what I checked, since it decides whether this was a latent bug or just hardening. I looked for any `` or `background-image:url()` in shipped page content that contains a path separator: the only ones in the tree are the 45 branding `summary.htm` files, and those load into the settings UI, not into a book page — everything else that matched was a node_modules README. Live branding images on a page use `/api/branding/...` and are already skipped by the `Contains("/api/")` test a few lines above. So no shipped content hits this today; it's hardening for a src that *could* be a real path, not a second user-visible bug.

It costs nothing to be safe here: a Windows file name can contain neither `/` nor `:`, so for a real book image the two properties produce identical output. Added `MarkImageNodesForThumbnail_RootedPathSrc_KeepsSeparatorsUnescaped` to hold the line.

JohnThomson:
[Claude Opus 5 from Thomson's machine during preflight] Consulted Devin on 2026-08-06 up to commit `cfe19be4d01d38bcab70ac74c3c4e454bf6213e7`.

Two rounds, because the first round's finding was worth acting on.

**Round 1** (`5d456013a2`) — no bugs, one thing to investigate, one informational note. The investigate item is the one thread on this PR: taking the path from `UrlEncoded` meant `/` and `:` stayed percent-encoded, which was harmless before but not after we stopped the string being decoded again. Fixed in `cfe19be4d0` by using `UrlEncodedForHttpPath`; the thread has the reasoning and is resolved.

**Round 2** (`cfe19be4d0`) — no bugs. Devin re-raised the same investigate item, but its text still describes the code reading from `UrlEncoded`, which this commit changed; the remedy it recommends is the one already applied, so it's stale rather than a new point, and its thread is already resolved. Three informational notes, none needing action:

- A file name containing `:` would now be emitted literally and a browser could read it as a URL scheme. Real, but a Windows file name cannot contain `:` and Bloom is effectively Windows-only; the immediately preceding behavior emitted it literally too, so this is not a regression.
- Confirmation that the downstream consumers of the thumbnail src were checked.
- The query portion round-trip is lossy — pre-existing, unchanged by this PR.

CI: the repo's `pr-automation` workflow never ran for this PR (no check runs registered against the head sha; there's a queue backlog). Its only two jobs are kicking off the Devin review and adding the review link to the description — both done by hand here, so nothing was lost. CodeRabbit is installed but the repo sets `auto_review.enabled: false`, so it does not review unless asked.

JohnThomson:
[Claude Opus 5 from Thomson's machine during preflight] Consulted Devin on 2026-08-06 up to commit `dfa351ad0abde85f85497c2eeae57c62a9cb7bec`.

Round 3, covering the two follow-up fixes the developer approved from the preflight report (the alt-text file name, and the guard on the debug-only filename-case check). **No bugs.**

One item still shows as needing investigation — the `%2f`/`%3a` one — but its text continues to describe `filename` being read from `UrlEncoded`, which `cfe19be4d0` already changed to `UrlEncodedForHttpPath`. It is the same stale finding as round 2, and [its thread](https://github.com/BloomBooks/BloomDesktop/pull/8169#discussion_r3731717235) is resolved with the fix recorded, so it is not re-posted.

Five informational notes. Two are about the new commit and both were assessed rather than waved through:

- *Alt text now differs further from the C# placeholder-alt-text matcher.* Accurate as an observation, but it changes no behavior. `HtmlDom.IsPlaceholderImageAltText` only ever matched on the parts that don't depend on the file name — the English substring test, and the localized format string split around `{0}` — and its exact-match-on-raw-src branch already failed before this change, since `decodeURI` also turned `%20` into a space. Widening a gap that nothing relies on costs nothing.
- *The debug-only case-check hardening is unrelated to the stated fix.* True, and deliberate: it is here because the developer asked for it after it caused about one run in three of `HandleDoubleEncodedUrls` to fail, which read as a regression from this very PR and cost real time to disprove. The note also points out that `GetFileSystemInfos(di.Name)` treats the name as a search pattern, so `?` or `*` would misbehave — that is pre-existing and cannot occur in a Windows file name.

The other three are pre-existing or confirmatory: the lossy query round-trip, the `:`-as-URL-scheme edge case on non-Windows filesystems, and a confirmation that the downstream consumers of the thumbnail src were checked.

Suites at this commit: C# `3037 passed / 13 skipped / 0 failed`, vitest `673 passed / 5 skipped`.

JohnThomson (commented):
> @JohnThomson+AGNT resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8169)* status: 0 of 5 files reviewed, all discussions resolved.

andrew-polk (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8169)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on JohnThomson).

PR #8170, opened 2026-08-06
Nest spreadsheet test temp folders so no test book sits in %TEMP% (BL-16661)

JohnThomson:
`SpreadsheetImporter.GetPageForLabel()` looks for page templates in "the collection the book is in", which it computes as the book folder's **parent**. That is right in production — a book's parent folder really is its collection — but the spreadsheet tests built their book directly in `%TEMP%`, so the parent was the temp directory itself and Bloom walked *every subfolder of it* as a candidate book.

On a working machine that is thousands of folders belonging to other tests and other programs. If any of them was deleted between the enumeration and the inspection, `FindBookHtmlInFolder` threw and took out the whole fixture in `OneTimeSetUp` — naming an unrelated folder on the way out, and hitting a different set of tests each run.

## The fix

New `SpreadsheetTestFolders` hands each fixture a folder of its own, named after the fixture, under a single root:

```
%TEMP%/BloomSpreadsheetTests/<FixtureName>/Book
                                          /Spreadsheet
```

Each fixture nests its book / spreadsheet / other folders inside that, so the folder the importer scans as "the collection" only ever holds that one fixture's folders. Teardown disposes just the root, which removes the rest. A side benefit: these tests now leave one entry in `%TEMP%` instead of three dozen.

The root is made with `Directory.CreateDirectory` rather than the `TemporaryFolder(name)` constructor, because that constructor first deletes any existing folder of the same name — which would throw away a sibling fixture's folders.

## Three latent bugs this surfaced

Naming each folder after its own fixture (`GetType().Name`) fixed some collisions that were already there:

- `SpreadsheetAudioTestsBase` created folders named `SpreadsheetImagesTests` / `SpreadsheetImagesTests_Book` — the same names `SpreadsheetImagesTests` uses, so whichever ran second deleted the other's folder. That base is shared by five fixtures.
- Three license fixtures all used `"SpreadsheetImportRemovingLicenseTests"`, and `SpreadsheetImportDeleteExtraPagesTest` used the name belonging to `SpreadsheetImageAndTextImportToBookWithComplexLastPageTests`.
- Both audio-import fixtures leaked `_spreadsheetFolder` (never disposed) and shared the name `"other audio folder"`.

## Test

Adds one guard test for the invariant the fix rests on: a book folder nested in one of these folders has, as its parent, a folder that is not `%TEMP%` and that contains nothing but our own folders.

All 391 `BloomTests.Spreadsheet` tests pass.

## Both of the card's suggested fixes are here

The card offered two fixes and said "either or both". This PR now does both.

**1. Keep test books out of `%TEMP%`** (the root-cause fix, described above).

**2. Make the scan itself tolerate a folder that vanishes.** `PageTemplatesApi.GetBooksInCollectionDirectories` now skips a book folder that no longer exists rather than letting `FindBookHtmlInFolder` throw and abort the whole scan. This one protects the shipping product, not just the tests: when a user opens the Add Page dialog, Bloom lists the folders in every source collection and inspects each one, and a folder can be deleted in that instant by a sync client, an antivirus tool, or the user. It does not weaken the `bl-291` warning that tells a user their book folder has gone missing — that check is on the save path, not the collection scan.

Its regression test reproduces the race deterministically rather than hoping to hit it: the scan is lazy, so the test takes one book from the enumerator, deletes the folder of whichever book it has *not* yet been handed (decided at run time, so the test does not depend on file-system enumeration order), and keeps enumerating. Without the guard it fails with the exact exception from the bug report — `In FindBookHtmlInFolder('...'), the folder does not exist. (ref bl-291)`.

---
[Devin review](https://app.devin.ai/review/BloomBooks/BloomDesktop/pull/8170)

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8170)

### DIFF
--- a/src/BloomExe/web/controllers/PageTemplatesApi.cs
+++ b/src/BloomExe/web/controllers/PageTemplatesApi.cs
@@ -139,6 +139,13 @@ namespace Bloom.web.controllers
             return collections
                 .Distinct() //seems to be needed in case a shortcut points to a folder that's already in the list.
                 .SelectMany(ProjectContext.SafeGetDirectories) // get all the (book) folders in those collections
+                // A folder can go away between being listed above and being looked at below --
+                // a sync client (Dropbox etc.), an antivirus tool, or the user can delete one
+                // while we are part way through a collection. FindBookHtmlInFolder throws for a
+                // folder that does not exist (that check is there for the save path, where a
+                // missing book folder really is the user's problem), and one such folder would
+                // otherwise abort the whole scan. Skipping it is all we want here. (BL-16661)
+                .Where(Directory.Exists)
                 .Select(BookStorage.FindBookHtmlInFolder); // and get the book from each
         }
 
--- a/src/BloomTests/web/controllers/PageTemplatesApiTests.cs
+++ b/src/BloomTests/web/controllers/PageTemplatesApiTests.cs
@@ -1,3 +1,4 @@
+using System.Collections.Generic;
 using System.IO;
 using Bloom.web.controllers;
 using BloomTemp;
@@ -153,6 +154,71 @@ namespace BloomTests.web.controllers
                     sourceBookPaths
                 );
                 Assert.AreEqual(3, result.Count, "Should list each unique path, not name.");
+            }
+        }
+
+        /// <summary>
+        /// A book folder can be deleted while we are part way through scanning its collection --
+        /// by a sync client, an antivirus tool, or the user. That must not abort the whole scan.
+        /// See BL-16661, where one folder disappearing took out a whole test fixture.
+        /// </summary>
+        [Test]
+        public void GetBooksInCollectionDirectories_FolderDeletedPartWayThrough_SkipsItAndKeepsGoing()
+        {
+            using (var collection = new TemporaryFolder("FolderDeletedPartWayThrough"))
+            {
+                // Two books. "a book" is read first; "b book" is the one we will delete after the
+                // scan has started but before it reaches that folder.
+                var stays = new TemplateBookTestFolder(collection.FolderPath, "a book");
+                File.WriteAllText(stays.HtmlPath, "<html></html>");
+                var vanishes = new TemplateBookTestFolder(collection.FolderPath, "b book");
+                File.WriteAllText(vanishes.HtmlPath, "<html></html>");
+
+                // Sanity check: both books are really there, and findable, before we start.
+                Assert.That(File.Exists(stays.HtmlPath), Is.True);
+                Assert.That(File.Exists(vanishes.HtmlPath), Is.True);
+                Assert.That(
+                    Directory.GetDirectories(collection.FolderPath).Length,
+                    Is.EqualTo(2),
+                    "Setup sanity check: the collection should contain exactly the two book folders."
+                );
+
+                var found = new List<string>();
+                // The result is lazy, so pulling one item at a time lets us delete the second
+                // book folder mid-scan -- exactly the race this guards against.
+                using (
+                    var scan = PageTemplatesApi
+                        .GetBooksInCollectionDirectories(new[] { collection.FolderPath })
+                        .GetEnumerator()
+                )
+                {
+                    Assert.That(scan.MoveNext(), Is.True, "Should have found the first book.");
+                    found.Add(scan.Current);
+
+                    SIL.IO.RobustIO.DeleteDirectoryAndContents(
+                        Path.GetDirectoryName(vanishes.HtmlPath)
+                    );
+                    Assert.That(
+                        Directory.Exists(Path.GetDirectoryName(vanishes.HtmlPath)),
+                        Is.False,
+                        "Sanity check: the second book folder should now be gone."
+                    );
+
+                    // Before the fix this threw, killing the whole scan.
+                    while (scan.MoveNext())
+                        found.Add(scan.Current);
+                }
+
+                Assert.That(
+                    found,
+                    Does.Contain(stays.HtmlPath),
+                    "The book that was still there should have been found."
+                );
+                Assert.That(
+                    found.Count,
+                    Is.EqualTo(1),
+                    "The deleted folder should have been skipped, not reported as a book."
+                );
             }
         }
     }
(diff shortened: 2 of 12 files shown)


JohnThomson:
[Claude Opus 5 (1M context)] Consulted Devin on 2026-08-06 16:13 CDT up to commit `26bba0d6ad4eac7bab314bddcf84d5f5cf5f9ee4`.

Devin's review of this commit finished clean: no bugs and nothing flagged for investigation. It did leave three informational observations, none of which need a change — and two of them are useful confirmations of the assumptions this PR rests on:

- Fixture folder names are the only thing stopping one fixture from deleting another's folder. Devin checked every call site: the names are distinct, `SpreadsheetAudioTestsBase.Setup` really is called only once per subclass, the one static caller uses a name of its own, and NUnit fixture-level parallelism is not enabled in this assembly.
- Nesting the book under the fixture folder means the importer now looks at that fixture's *other* folders (`ss`, `Dest`, `other audio folder`, …) as candidate books. Harmless, because Bloom only looks for a top-level `.htm`/`.html` in each and none of them has one — the widget `drag.html` is in a subdirectory. Worth remembering if a future test drops an HTML file at the top of one of those folders.
- The `%TEMP%/BloomSpreadsheetTests` root is deliberately never deleted (deleting it would take out other fixtures' folders). The side effect is that if a fixture is renamed or removed, its old folder sits there until someone clears it by hand.

src/BloomTests/web/controllers/PageTemplatesApiTests.cs
```diff
@@ -155,5 +156,70 @@ public void GetBookTemplatePaths_TwoTemplatesWithSameName_ListsBoth()
...
+                );
+
+                var found = new List<string>();
+                // The result is lazy, so pulling one item at a time lets us delete the second
+                // book folder mid-scan -- exactly the race this guards against.
+                using (
+                    var scan = PageTemplatesApi
+                        .GetBooksInCollectionDirectories(new[] { collection.FolderPath })
+                        .GetEnumerator()
+                )
```
JohnThomson, line 193: [Devin] **Investigate**: New scan test depends on directory enumeration order

The new test assumed the lazy scan would hand it `a book` first, and that `b book` would still be pending when it deleted that folder. The order comes from `Directory.GetDirectories`, which is up to the file system — alphabetical on NTFS, but arbitrary on ext4. On a file system that returned `b book` first, the test would have deleted a folder the scan had already consumed, found two books, and failed the final count assertion even though the production code behaved correctly.

JohnThomson: [Claude Opus 5 (1M context)] Good catch — fixed in `bf3f174deb2ef3d472d92fe2fbb1b4aa0802bb66`. A flaky test in a branch about flaky tests would have been a poor advertisement.

The test no longer decides in advance which book disappears. It takes the first book the scan hands it, works out at run time which of the two it has *not* yet been given, and deletes that one. The final assertion is now "the deleted book is not among the results" plus a count, rather than naming which book should survive — so it behaves the same whatever order the file system returns folders in.

Re-verified both ways: it passes with the guard, and without it still fails with the exception from the bug report, `In FindBookHtmlInFolder('...'), the folder does not exist. (ref bl-291)`.

src/BloomExe/web/controllers/PageTemplatesApi.cs
```diff
@@ -139,6 +139,13 @@ IEnumerable<string> collections
             return collections
                 .Distinct() //seems to be needed in case a shortcut points to a folder that's already in the list.
                 .SelectMany(ProjectContext.SafeGetDirectories) // get all the (book) folders in those collections
+                // A folder can go away between being listed above and being looked at below --
```
JohnThomson, line 142: [Devin] **Investigate**: Production guard is included even though the description says it is not

The PR description's "Scope note" said option 2 (making the scan tolerate a folder that vanishes mid-enumeration, "the narrow version would be a `.Where(Directory.Exists)` in `PageTemplatesApi.GetBooksInCollectionDirectories`") was *not* included, but the diff adds exactly that, plus a new regression test. The change itself is sound -- the chain stays lazy, `SafeGetDirectories` yields directory paths so `Directory.Exists` is the right predicate, and `FindBookHtmlInFolder` only throws (and shows a user dialog via `ErrorReport.NotifyUserOfProblem`) when the folder is missing. Worth reconciling the description with the diff so reviewers know the production file changed.

JohnThomson: [Claude Opus 5 (1M context)] Correct, and the description was the thing that was wrong -- thanks for catching it. The scope note was written when this PR really did stop at option 1; the developer then reviewed the preflight report and asked for option 2 as well, and I added it without going back to update the description. Exactly the mismatch a reviewer would trip over, since it says "no production code changed" over a diff that changes production code.

The description now has a "Both of the card's suggested fixes are here" section describing the guard, why it is worth having for the shipping product rather than only for the tests, and why it does not weaken the `bl-291` warning on the save path.

Your read of the change is right on every point, including that the chain stays lazy -- the test depends on that, since it deletes a book folder part way through enumerating.

JohnThomson:
[Claude Opus 5 (1M context)] Consulted Devin on 2026-08-06 17:30 CDT up to commit `bf3f174deb2ef3d472d92fe2fbb1b4aa0802bb66`.

Two rounds, because the first round's findings led to a fix commit. Across both, Devin raised no bugs and two things worth acting on, each with its own resolved thread above:

- The new regression test depended on the order the file system returns folders in — alphabetical on Windows, arbitrary on Linux — so it could have been flaky on Linux. Fixed: it now deletes whichever book the scan has not yet handed it.
- The PR description still said the production guard was *not* included, after the developer asked for it and I added it. Description rewritten to match the diff.

Devin also left four informational notes, which are observations rather than action items and are not mirrored as threads. One is worth repeating here because it is a genuine limitation rather than a nitpick: **the `Directory.Exists` check narrows the race but cannot close it** — a folder can still vanish in the instant between that check and `FindBookHtmlInFolder` opening it. That is inherent to a file system that other programs are also writing to; the guard turns a near-certainty on a busy machine into a very small window, and the remaining window would need a caught exception rather than a pre-check to close entirely.

CI: this repo runs no build or test checks on pull requests, so there was no CI signal. Greptile has stayed queued without reviewing.

andrew-polk (approved):
> @andrew-polk reviewed 12 files and all commit messages, and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8170)* status: all files reviewed, 1 unresolved discussion.

PR #8171, opened 2026-08-06
Add diagnostics for missing files requested by bare name (BL-16577)

JohnThomson:
## What this is

Sentry issue [2699459502](https://bloom-app.sentry.io/issues/2699459502/) ("Cannot Find File") is
dominated by requests for the bare name `Checkbox.js`. The card supposed no such file exists in
Bloom. It does: it's a Vite chunk in `output/browser`. A bundle we inject into a page at the server
root (`Book.AddJavascriptFile`) imports its sibling chunks by bare name, so they arrive at the
server with no directory — and they should always be found under `BrowserRoot`.

**This PR does not fix that bug.** I could not establish why the lookup fails on the affected
machines: the third lookup, `GetFileDistributedWithApplication`, is anchored to the application
folder, and the absolute browser root is already a factory search path, so on a valid install the
file is found regardless of anything here. What this adds is the instrumentation needed to find
out, plus a real weakness the investigation turned up.

## Changes

**Diagnostics.** When a request that carries *no directory* misses, the Sentry detail message now
also reports `BrowserRoot`, whether the expected file is actually present, and the process's
current working directory. That distinguishes "the file is genuinely absent from that install"
from "we looked in the wrong place". Requests that carry a directory are unaffected.

**The current-directory weakness.** `BloomFileLocator.BrowserRoot` is a *relative* path, so testing
for a file with it resolves against the process's current working directory. Bloom doesn't control
that directory:

- Windows File Explorer sets it to the folder of a double-clicked file — Bloom's own comment in
  `Program.cs` says so (BL-11004);
- a shortcut's "Start in" value overrides it — this already bit us as BL-16230, which is why
  `Program.NormalizeWorkingDirectory()` exists;
- any Open/Save dialog that doesn't set `RestoreDirectory` moves it **mid-session**, permanently —
  and `BloomOpenFileDialog` never set it.

Two lookups depended on it: `BloomServer.LookForAFullPathToFile`'s browser-root preference (added
in `57893d0f1f` for BL-15894) and `BloomFileLocator.GetSearchPaths`. Both now use a new
`AbsoluteBrowserRoot`, and the file dialogs no longer move the working directory.

**Test double.** `PretendRequestInfo.GetQueryParameters()` always returned an empty collection,
which made any request the server routes on a query parameter — such as the `assetv` one it adds to
every JS request — impossible to test at all. It now parses the query.

## Risk to look at

The `GetSearchPaths` change is the one with teeth: that entry previously resolved to nothing
whenever the working directory wasn't the application folder, so making it work **changes lookup
precedence** — the browser root now genuinely wins over the factory paths, which is what the
comment there always intended. The precedence tests in `BloomFileLocatorTests` cover this and pass.

## Testing

`BloomServerTests` + `BloomFileLocatorTests` + `BookStarterTests`: **125 passed, 0 failed**, stable
across repeated runs, before and after merging `master`.

Full-suite counts are not usable as a signal at the moment — four runs of one unchanged tree gave
135, 1, 32 and 62 failures. That is other worktrees sharing `%TEMP%` plus tests that mutate
process-wide state, not this branch. (My first attempt at a regression test here changed the
process's current directory and took out a different unrelated test on each run; it was rewritten
to assert the resolved path is rooted instead.)

Ref: https://issues.bloomlibrary.org/youtrack/issue/BL-16577

[Devin review](https://devinreview.com/BloomBooks/BloomDesktop/pull/8171)

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8171)

### DIFF
--- a/src/BloomExe/MiscUI/BloomOpenFileDialog.cs
+++ b/src/BloomExe/MiscUI/BloomOpenFileDialog.cs
@@ -15,6 +15,13 @@ namespace Bloom.MiscUI
             Multiselect = false;
             CheckFileExists = true;
             CheckPathExists = true;
+            // The Windows file dialog otherwise leaves the process's current working directory
+            // wherever the user last browsed, for the rest of the session. Bloom looks for some of
+            // the files it ships with relative to that directory, so letting a dialog move it can
+            // break those lookups long after the dialog is closed. See BL-16577 and BL-16230.
+            // (This only restores the process directory; the dialog still opens where the shell or
+            // our InitialDirectory says it should.)
+            RestoreDirectory = true;
             _dialog.FileOk += (sender, args) =>
             {
                 // Truly enforce the filter. See BL-12929 and BL-13552.
--- a/src/BloomExe/BloomFileLocator.cs
+++ b/src/BloomExe/BloomFileLocator.cs
@@ -30,6 +30,12 @@ namespace Bloom
         private readonly IEnumerable<string> _userInstalledSearchPaths;
         private readonly IEnumerable<string> _afterXMatterSearchPaths;
 
+        /// <summary>
+        /// Where our built web assets live, *relative* to the application (or solution) folder.
+        /// Only pass this to something that resolves it against that folder, such as
+        /// FileLocationUtilities.GetFileDistributedWithApplication() or GetBrowserFile(). If you
+        /// want to test for a file yourself, use <see cref="AbsoluteBrowserRoot"/> instead.
+        /// </summary>
         public static string BrowserRoot
         {
             get
@@ -41,6 +47,21 @@ namespace Bloom
                     : "browser";
             }
         }
+
+        /// <summary>
+        /// The full path of the folder that <see cref="BrowserRoot"/> names. Use this anywhere we
+        /// test for, or hand around, the location of a file that ships with Bloom.
+        ///
+        /// BrowserRoot on its own is a relative path, so asking whether something exists at it
+        /// silently resolves against the process's current working directory. Bloom does not
+        /// control that directory: Windows File Explorer sets it to the folder of a file you
+        /// double-click (see the comment on Program.NormalizeWorkingDirectory), a shortcut's
+        /// "Start in" value can override it (BL-16230), and any Open/Save dialog that does not set
+        /// RestoreDirectory moves it mid-session. Files that ship with Bloom must not be findable
+        /// only when that directory happens to be right. See BL-16577.
+        /// </summary>
+        public static string AbsoluteBrowserRoot =>
+            Path.Combine(FileLocationUtilities.DirectoryOfApplicationOrSolution, BrowserRoot);
 
         public static BloomFileLocator sTheMostRecentBloomFileLocator;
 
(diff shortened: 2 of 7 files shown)


src/BloomExe/MiscUI/BloomOpenFileDialog.cs
```diff
@@ -15,6 +15,13 @@ public BloomOpenFileDialog()
             Multiselect = false;
             CheckFileExists = true;
             CheckPathExists = true;
+            // The Windows file dialog otherwise leaves the process's current working directory
+            // wherever the user last browsed, for the rest of the session. Bloom looks for some of
+            // the files it ships with relative to that directory, so letting a dialog move it can
+            // break those lookups long after the dialog is closed. See BL-16577 and BL-16230.
+            // (This only restores the process directory; the dialog still opens where the shell or
+            // our InitialDirectory says it should.)
+            RestoreDirectory = true;
```
JohnThomson, line 24: [Devin] **Investigate**: Turning on RestoreDirectory globally for the open dialog

Devin's point: setting this in the constructor changes behavior for **every** caller of
`BloomOpenFileDialog`, not just the ones this PR was thinking about. Its own analysis mostly
clears it — no caller sets `RestoreDirectory` explicitly, so nothing is being overridden, and on
Vista+ the "open where I last browsed" behavior comes from the shell's per-application MRU rather
than from the process working directory. But it asked for a sanity check on any dialog that does
**not** set `InitialDirectory`, since that is the case where the working directory could
conceivably still matter.

I checked all five call sites. Four set `InitialDirectory` and are unaffected:

- `WorkspaceView.cs:1310`, `CollectionChooserApi.cs:433`, `CollectionApi.cs:510` — all
  `DefaultParentDirectoryForCollections`
- `SpreadsheetApi.cs:116` — `GetSpreadsheetFolderFor(book, false)` (and this one already set
  `RestoreDirectory = true` before this PR)

One does not: **`CollectionModel.BloomSourceImport.cs:62`** — the "import a BloomSource file"
dialog sets `Title` and `Filter` but no `InitialDirectory`.

So the only user-visible question left is: after this change, does that one dialog still open
where you would expect? Expected answer is yes (the shell MRU remembers the last folder per
application, which is what actually drives it), but that is the one thing worth eyeballing rather
than reasoning about, so I am leaving this thread open for a human rather than resolving it myself
— it is user-facing behavior.

JohnThomson:
[Claude Opus 5 (1M context)] Consulted Devin on 2026-08-06 up to commit `33e01b80d1ea273d1322890dd69d354ae7dd63f0`.

It found no bugs. It raised one thing worth a human look — whether turning on `RestoreDirectory`
for every file dialog changes where the one dialog that doesn't set its own starting folder opens
— which is [its own thread](https://github.com/BloomBooks/BloomDesktop/pull/8171#discussion_r3732106720).

Five further items were informational only, and I've folded the two that matter into the PR
description already: that making the browser-root search path absolute changes lookup *precedence*
(which is the intended effect, and the thing to review hardest), and that `LocateFile` now returns
rooted paths where a browser-root hit used to come back relative. The others noted that the new
test writes a scratch file into the shared `output/browser` folder (GUID-named, deleted in a
`finally`), that `PretendRequestInfo.LocalPathWithoutQuery` still keeps the query string unlike the
real one, and that the new diagnostics text is appended for any directory-less miss rather than
only JS ones.

CI reported no checks on this branch, and no other review bots have commented.

src/BloomExe/MiscUI/BloomOpenFileDialog.cs
JohnThomson: [Claude Opus 5 (1M context)] Checked by hand and it behaves correctly: the BloomSource import dialog remembers the folder it was last used to import from, so turning on `RestoreDirectory` has not changed where it opens. That was the only user-facing question here, so resolving.

andrew-polk (approved):
> @andrew-polk reviewed 7 files and all commit messages, and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8171)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on JohnThomson).

andrew-polk (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8171)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on JohnThomson).

PR #8172, opened 2026-08-06
Give each test run its own temp directory (BL-16664)

JohnThomson:
Our C# tests name their scratch folders after themselves — about **180** calls of the form `new TemporaryFolder("SomeFixtureTests")` — and those names resolve to **machine-global** paths. That was harmless when one person ran the suite at a time. It is not harmless now that agents work in several worktrees at once, because `TemporaryFolder`'s constructor **deletes** any existing folder of that name before creating it:

```csharp
public TemporaryFolder(string name)
{
    _path = Path.Combine(Path.GetTempPath(), name);
    if (Directory.Exists(_path))
        DeleteFolderThatMayBeInUseAndIfNotFailSilently(_path);   // <-- someone else's live folder
    Directory.CreateDirectory(_path);
}
```

So one run's setup silently deletes another run's in-flight folder, and the victim then fails somewhere unrelated, naming a folder it never heard of.

## The fix

Rather than rename 180 call sites — which would still leave Bloom's own production code writing to the shared temp directory while under test — redirect the **whole process's** temp directory before any fixture runs, with an assembly-wide `[SetUpFixture]`:

```
%TEMP%\BloomTests\<key>-p<pid>\
```

Every existing call site keeps its familiar name; the name is now scoped to this run. `Path.GetTempPath()` is defined in terms of `TMP`/`TEMP`, so this also covers the dozen tests that build temp paths by hand, and production code under test.

The `<key>` half comes from `BLOOM_AGENT_BUILD_DIR` when `build/agent-dotnet.sh` set it, so a temp folder can be matched to the build tree beside it; the pid is what guarantees uniqueness.

It lives in the **global namespace** deliberately: an NUnit `[SetUpFixture]` applies to its own namespace and those beneath it, and not quite all our fixtures are under `BloomTests` (there is one in `Bloom.Api`).

**A failing run keeps its folder** so the evidence survives — these bugs are precisely about temp folders appearing and disappearing, so deleting the scene of the crime would be perverse. Passing runs delete theirs; anything older than a day is cleared by a later run.

## Effect on the suite

Measured on one developer machine, whose `%TEMP%` holds ~25,000 entries:

| Tree | Total run | Passed | Failed | Aborted? |
| --- | --- | --- | --- | --- |
| pristine `master` | 1678 | 1603 | **65** | yes |
| this branch | **3046** | **3032** | **1** | **no** |

The suite ran to completion, and two consecutive runs gave identical results — so this is deterministic, not luck. The remaining failure, `XMatterHelperTests.InsideBackCover_PaperAndDeviceXMatter_ShareOneCustomLayoutId`, also fails on pristine master and is unrelated to temp; note that `AGENTS.md` calls a failing xmatter-locating test a real wrapper/`Directory.Build.props` regression, so it is worth chasing separately.

## Tests

Two, asserting the invariant directly: that `Path.GetTempPath()` is this run's own folder and not the machine's, and that an ordinary fixed-name `TemporaryFolder` lands inside it.

Ref: https://issues.bloomlibrary.org/youtrack/issue/BL-16664

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8172)

---
[Devin review](https://app.devin.ai/review/BloomBooks/BloomDesktop/pull/8172)

[Devin review](https://devinreview.com/BloomBooks/BloomDesktop/pull/8172)

### DIFF
--- a/src/BloomTests/TestTempDirectory.cs
+++ b/src/BloomTests/TestTempDirectory.cs
@@ -49,6 +49,10 @@ public class TestTempDirectory
     /// <summary>The folder this run's temporary files live in.</summary>
     internal static string RunFolder => _runFolder;
 
+    /// <summary>
+    /// Points this process's temp directory at a folder of our own, before any fixture runs, and
+    /// takes the opportunity to clear out folders left by runs that died. NUnit calls this once.
+    /// </summary>
     [OneTimeSetUp]
     public void RedirectTempToAFolderOfOurOwn()
     {
@@ -65,6 +69,11 @@ public class TestTempDirectory
         RemoveFoldersLeftByRunsThatDiedBeforeCleaningUp(container);
     }
 
+    /// <summary>
+    /// Deletes this run's temp folder — and with it everything the run put in temp, since every
+    /// temp path the process computed descends from it. Kept instead of deleted when tests
+    /// failed, so their files can be examined. NUnit calls this once, at the end of the run.
+    /// </summary>
     [OneTimeTearDown]
     public void RemoveOurTempFolderUnlessSomethingFailed()
     {
--- a/src/BloomTests/TestTempDirectory.cs
+++ b/src/BloomTests/TestTempDirectory.cs
@@ -77,6 +77,13 @@ public class TestTempDirectory
     [OneTimeTearDown]
     public void RemoveOurTempFolderUnlessSomethingFailed()
     {
+        // Point temp back at the machine's own folder before we go. Anything that runs after
+        // this -- NUnit's own shutdown, a background thread that outlives the tests -- would
+        // otherwise compute temp paths inside a directory we are about to delete, and fail
+        // confusingly at the very end of an otherwise good run.
+        Environment.SetEnvironmentVariable("TMP", MachineTempFolder);
+        Environment.SetEnvironmentVariable("TEMP", MachineTempFolder);
+
         // When tests failed, leave the folder alone. What a failing test wrote is often the
         // evidence you need, and this suite's nastiest bugs have been about temp folders
         // appearing and disappearing — deleting the scene of the crime would be perverse.
--- a/src/BloomTests/TestTempDirectory.cs
+++ b/src/BloomTests/TestTempDirectory.cs
@@ -59,7 +59,7 @@ public class TestTempDirectory
         MachineTempFolder = Path.GetTempPath();
         var container = Path.Combine(MachineTempFolder, kContainerName);
         _runFolder = Path.Combine(container, KeyForThisRun());
-        Directory.CreateDirectory(_runFolder);
+        StartFolderEmpty(_runFolder);
 
         // Path.GetTempPath() is defined in terms of these, so from this point on every temp path
         // the process computes — ours and Bloom's own — lands inside _runFolder.
@@ -163,7 +163,27 @@ public class TestTempDirectory
     }
 
     /// <summary>
-    /// A name for this run's folder: unique, so no two runs can ever share one, but still
+    /// Makes sure the run folder exists and is empty.
+    ///
+    /// Emptying matters because the name can repeat. A run that crashes leaves its folder behind
+    /// (and a failing run leaves it deliberately), those folders survive for a day, and Windows
+    /// recycles process ids — so this exact path may already exist, holding the remains of a run
+    /// that is long gone. Inheriting those would produce exactly the confusing, stale-file
+    /// failures this class exists to prevent.
+    ///
+    /// Deleting it is safe: a process id is unique among *running* processes, and this one is
+    /// ours, so nothing alive can be using the folder.
+    /// </summary>
+    internal static void StartFolderEmpty(string folder)
+    {
+        if (Directory.Exists(folder))
+            TemporaryFolder.DeleteFolderThatMayBeInUseAndIfNotFailSilently(folder);
+        Directory.CreateDirectory(folder);
+    }
+
+    /// <summary>
+    /// A name for this run's folder: unique among live runs, but not unique forever, since it is
+    /// built from a process id and those get recycled — see <see cref="StartFolderEmpty"/>. Still
     /// recognizable, so you can tell which terminal a leftover folder came from.
     /// </summary>
     private static string KeyForThisRun()
--- a/src/BloomTests/TestTempDirectory.cs
+++ b/src/BloomTests/TestTempDirectory.cs
@@ -4,231 +4,232 @@ using BloomTemp;
 using NUnit.Framework;
 using NUnit.Framework.Interfaces;
 
-/// <summary>
-/// Gives this test process a temporary directory of its own, so that two test runs on the same
-/// machine cannot tread on each other's scratch folders.
-///
-/// Our tests name their temp folders after themselves — there are around 180 calls of the form
-/// `new TemporaryFolder("SomeFixtureName")` — and those names resolve to machine-global paths.
-/// That was harmless when one person ran the suite at a time. It is not harmless now that agents
-/// work in several worktrees at once, because TemporaryFolder's constructor *deletes* any
-/// existing folder of the name before creating it (see TemporaryFolder in
-/// src/BloomExe/TempFiles.cs). So one run's setup would quietly delete another run's in-flight
-/// folder, and the victim would fail somewhere unrelated, naming a folder it had never heard of.
-/// See BL-16664, and BL-16661 for the same shape of failure from a different cause.
-///
-/// Rather than rename 180 call sites — which would still leave Bloom's own production code
-/// writing to the shared temp directory while under test — we move the whole process's idea of
-/// where "temp" is. Every existing call site then keeps its familiar name, but the name is scoped
-/// to this run.
-/// </summary>
-/// <remarks>
-/// This deliberately sits in the global namespace. An NUnit [SetUpFixture] applies to the
-/// namespace it is declared in and the namespaces beneath it; nearly all our tests are under
-/// BloomTests, but not quite all (there is a fixture in the Bloom.Api namespace), and the
-/// redirect has to be in place before *any* fixture runs. The global namespace covers the whole
-/// assembly and is ordered ahead of the BloomTests one.
-/// </remarks>
-[SetUpFixture]
-public class TestTempDirectory
+namespace BloomTests
 {
-    /// <summary>Everything this assembly writes to temp goes under here, a folder per run.</summary>
-    private const string kContainerName = "BloomTests";
-
-    /// <summary>How long a leftover run folder must sit untouched before another run clears it.</summary>
-    private static readonly TimeSpan kStaleAfter = TimeSpan.FromDays(1);
-
-    private static string _runFolder;
-
     /// <summary>
-    /// The machine-wide temp directory, as it was before we redirected. Kept so the tests for
-    /// this class can check that we really did move somewhere else.
+    /// Gives this test process a temporary directory of its own, so that two test runs on the same
+    /// machine cannot tread on each other's scratch folders.
+    ///
+    /// Our tests name their temp folders after themselves — there are around 180 calls of the form
+    /// `new TemporaryFolder("SomeFixtureName")` — and those names resolve to machine-global paths.
+    /// That was harmless when one person ran the suite at a time. It is not harmless now that agents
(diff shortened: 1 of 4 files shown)


src/BloomTests/TestTempDirectory.cs
```diff
@@ -0,0 +1,137 @@
...
+    /// The machine-wide temp directory, as it was before we redirected. Kept so the tests for
+    /// this class can check that we really did move somewhere else.
+    /// </summary>
+    internal static string MachineTempFolder { get; private set; }
+
+    /// <summary>The folder this run's temporary files live in.</summary>
+    internal static string RunFolder => _runFolder;
+
+    [OneTimeSetUp]
+    public void RedirectTempToAFolderOfOurOwn()
```
JohnThomson: [Devin] **Bug**: New public setup and teardown methods are missing the required documentation comments

The two public NUnit hooks that redirect and clean up the per-run temp folder — `RedirectTempToAFolderOfOurOwn` and `RemoveOurTempFolderUnlessSomethingFailed` — had no documentation comment. `AGENTS.md` ("# Commenting") says "All public methods should have a comment. So should most private ones!" The private helpers here were documented but the two entry points were not, and neither were the two test methods.

JohnThomson: [Claude Opus 5 (1M context)] Fixed in `dbf656b51b02dabae60a24ec9fa55214da9a7e43`: both public hooks now carry a summary saying what NUnit calls them for — the setup one that it redirects temp before any fixture runs and clears folders left by dead runs, the teardown one that deleting the run folder removes everything the run put in temp, and that it is kept rather than deleted when tests failed. The two test methods got summaries too, since Devin correctly noted they were undocumented as well.

src/BloomTests/TestTempDirectory.cs
```diff
@@ -0,0 +1,146 @@
...
+    /// failed, so their files can be examined. NUnit calls this once, at the end of the run.
+    /// </summary>
+    [OneTimeTearDown]
+    public void RemoveOurTempFolderUnlessSomethingFailed()
+    {
+        // When tests failed, leave the folder alone. What a failing test wrote is often the
+        // evidence you need, and this suite's nastiest bugs have been about temp folders
+        // appearing and disappearing — deleting the scene of the crime would be perverse.
+        // Whatever we leave behind is cleared by a later run once it goes stale.
+        if (TestContext.CurrentContext.Result.Outcome.Status == TestStatus.Failed)
```
JohnThomson: [Devin] **Investigate**: Relies on aggregated result being visible in a SetUpFixture's OneTimeTearDown

The keep-on-failure behaviour depends on `TestContext.CurrentContext.Result.Outcome.Status` reporting `Failed` inside the *SetUpFixture's* `OneTimeTearDown` — the outer namespace suite result. NUnit does aggregate child results into the containing suite, but the exact context available to a global `[SetUpFixture]` teardown is a subtle corner of NUnit. If it ever reported Passed for a run with failures, the folder would be deleted and the evidence lost silently.

JohnThomson: [Claude Opus 5 (1M context)] Fair thing to want checked rather than assumed, so I ran it: with the container holding 5 folders, a run of a single **failing** test left 6 behind; a run of a single **passing** test went 6 -> 6, i.e. it created its folder and removed it again. So the outcome really is visible there, and the branch behaves as intended.

Checking it did turn up a related problem, now fixed in `b0ac8a7c94cefcc5acea0280d4b0427a0873f32e`: your suggested way of confirming it -- looking for the "left in ..." message -- does not work, because `dotnet test` hides NUnit's progress channel at its default verbosity. The note I had added to AGENTS.md told developers the path "is printed at the end of the run", which would have sent them looking in the output, seeing nothing, and concluding the folder had not been kept. It now tells them to look in `%TEMP%\BloomTests\` instead.

src/BloomTests/TestTempDirectory.cs
```diff
@@ -0,0 +1,146 @@
...
+        {
+            TestContext.Progress.WriteLine(
+                $"Tests failed, so their temporary files have been left in {_runFolder}"
+            );
+            return;
+        }
+
+        // Failing silently is deliberate: a file some test left open must not turn a green run
+        // red at the very last moment.
+        TemporaryFolder.DeleteFolderThatMayBeInUseAndIfNotFailSilently(_runFolder);
```
JohnThomson: [Devin] **Investigate**: Deleting the process temp folder at teardown leaves TMP/TEMP pointing at a missing directory

`RemoveOurTempFolderUnlessSomethingFailed` deletes the run folder but leaves `TMP` and `TEMP` still pointing at it. Anything that computes a temp path after that point -- NUnit's own shutdown, or a background thread that outlives the tests -- would be handed a directory that no longer exists, and would fail at the very end of an otherwise good run.

JohnThomson: [Claude Opus 5 (1M context)] Agreed, and fixed in `4ba357019edc6ddec694192d31b293740e14f3e7`. Teardown now points `TMP`/`TEMP` back at the machine's own temp folder before it does anything else, whether or not it then deletes ours -- so nothing that runs after the tests can be handed a path inside a directory we removed.

Worth noting this matters more here than it might look, because the failure it would cause arrives after every test has already reported, which is exactly the kind of end-of-run noise this branch exists to get rid of.

JohnThomson:
[Claude Opus 5 (1M context)] Consulted Devin on 2026-08-06 17:42 CDT up to commit `4ba357019edc6ddec694192d31b293740e14f3e7`.

Four rounds, because each round's findings produced a fix commit, which starts a new review. Three things came out of it, each with its own resolved thread above:

- **The two public NUnit hooks had no documentation comments**, which the repo's guidance requires. Added.
- **Nothing had actually verified the keep-on-failure behaviour**, which depends on a corner of NUnit — whether a run's aggregated result is visible inside a global `[SetUpFixture]`'s teardown. Checked it directly: a run of one failing test leaves its folder behind, a run of one passing test creates and removes its own. That check also exposed a wrong sentence in the AGENTS.md note I had added, which is now corrected.
- **Teardown deleted the run folder but left `TMP`/`TEMP` pointing at it**, so anything running after the tests would have been handed a path inside a directory that no longer existed. Now restored before the delete.

Two notes for anyone reading the thread list. Devin appears to **accumulate** findings across rounds rather than recomputing them against the current diff: the final round re-reported all three of the above against commit `4ba35701`, even though every one of them is fixed *in that very commit*. They were not posted again, since each already has a documented, resolved thread. Devin also left four informational observations each round, which are not mirrored.

CI: this repo runs no build or test checks on pull requests, so there was no CI signal. Greptile has stayed queued throughout without reviewing.

src/BloomTests/TestTempDirectory.cs
```diff
@@ -0,0 +1,214 @@
...
+    /// Points this process's temp directory at a folder of our own, before any fixture runs, and
+    /// takes the opportunity to clear out folders left by runs that died. NUnit calls this once.
+    /// </summary>
+    [OneTimeSetUp]
+    public void RedirectTempToAFolderOfOurOwn()
+    {
+        MachineTempFolder = Path.GetTempPath();
+        var container = Path.Combine(MachineTempFolder, kContainerName);
+        _runFolder = Path.Combine(container, KeyForThisRun());
+        Directory.CreateDirectory(_runFolder);
```
JohnThomson: [Devin] **Bug**: Leftover scratch folder from a crashed earlier run can be reused instead of started fresh

The per-run scratch folder was only created, never emptied, so when the folder name repeats -- the name is an optional label plus the process number, which the operating system recycles -- a new run inherits whatever a crashed earlier run left there. Folders are kept when a run fails or crashes and are only removed once a day old, so a repeat within that window is quite possible.

Impact: a test run can start with stale files in its scratch area, producing exactly the confusing failures the isolation was meant to eliminate.

JohnThomson: [Claude Opus 5 (1M context)] A real hole, and the sharpest finding of this review -- thank you. Fixed in `e9188cba75c9ffa7ff9aeb66ffc08b85b9e291eb`.

The folder is now emptied before use, not merely created. Deleting it first is safe for the specific reason that makes the name repeat in the first place: a process id is unique among *running* processes, and this one is ours, so nothing alive can be holding the folder we are about to clear.

It is worth spelling out why this mattered more than it looks. The whole point of the branch is that a run should never see files it did not put there. Two of the retention rules I had just added -- keep the folder when tests fail, and sweep only after a day -- were what made the collision window wide enough to matter, so the feature was quietly undermining itself.

Two tests: one seeds a folder named the way a recycled run would be, with a stale file and a stale subfolder, and asserts nothing survives; the other covers the ordinary case where the folder does not exist yet.

JohnThomson:
[Claude Opus 5 (1M context)] Consulted Devin on 2026-08-07 09:32 CDT up to commit `e9188cba75c9ffa7ff9aeb66ffc08b85b9e291eb`.

Second preflight, after the developer asked for a warning when the run's temp folder cannot be
deleted. Devin found one further bug, now fixed and its thread resolved above:

- **A run could inherit a dead run's scratch folder.** The folder was created but never emptied,
  and its name is a label plus the process id — which Windows recycles — while folders from failing
  or crashed runs are deliberately kept for a day. So the two retention rules added in this branch
  were themselves widening the window for a run to start on top of stale files, which is the exact
  confusion the branch exists to remove. It now starts empty.

Its final round re-reported all four findings from this branch's history against `e9188cba`,
including that one — which is fixed *in that very commit*. As noted in the earlier log, Devin
accumulates findings across rounds rather than recomputing them, so nothing was posted twice; each
already has a documented, resolved thread.

Full suite at this commit: 3067 run, 3053 passed, 1 failed — `XMatterHelperTests.InsideBackCover_PaperAndDeviceXMatter_ShareOneCustomLayoutId`, which fails on pristine master too — and no abort.

One correction for the record: during this round `MakePdf_BookNameIsNonAscii_OutputsPdf` failed twice and I initially took it for a regression from this branch. It is not: it passed 5 of 5 subsequent runs in isolation and in two clean full runs. Both failures fell during a full-suite run that took 11 minutes rather than the usual 4-5, i.e. under heavy contention, and that test shells out to WebView2.

CI: this repo runs no build or test checks on pull requests. Greptile has stayed queued throughout without reviewing.

andrew-polk (changes requested):
> @andrew-polk reviewed 3 files and all commit messages, made 1 comment, and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8172)* status: all files reviewed, 4 unresolved discussions (waiting on JohnThomson).

___
*[`src/BloomTests/TestTempDirectory.cs` line 28 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8172#-OzRxqSC74jiX9AmdwPR:-OzRxqSC74jiX9AmdwPS:blbiw27) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/e9188cba75c9ffa7ff9aeb66ffc08b85b9e291eb/src/BloomTests/TestTempDirectory.cs#L28)):*
> there is a fixture in the Bloom.Api namespace

Seems like we should fix that.

Maybe doesn't change this other than the comment, of course.

JohnThomson (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8172)* status: 2 of 4 files reviewed, 4 unresolved discussions (waiting on andrew-polk).

___
*[`src/BloomTests/TestTempDirectory.cs` line 28 at r2](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8172#-OzRxqSC74jiX9AmdwPR:-OzSmVcw3mv68EkdFbHt:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/e9188cba75c9ffa7ff9aeb66ffc08b85b9e291eb/src/BloomTests/TestTempDirectory.cs#L28)):*

> there is a fixture in the Bloom.Api namespace

Seems like we should fix that.

Maybe doesn't change this other than the comment, of course.
</blockquote></details>

Done.

andrew-polk (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8172)* status: all files reviewed, 3 unresolved discussions.

PR #8174, opened 2026-08-06
Make the publish upload modal on the publish-tool switcher too (BL-16654)

andrew-polk:
Uploading a book to BloomLibrary already greyed out the main workspace tabs, but left the publish-tool buttons (PDF & Print, Web, BloomPUB, …) live. One click tore down the Web screen, which took its Cancel button, its progress log and its eventual "Your Book on BloomLibrary.org" link with it while the upload carried on invisibly; coming back offered an enabled Upload button, so a second concurrent upload of the same book was reachable.

Rather than give the Publish tab a second, weaker notion of "busy", this reports the existing lock to the browser: `WorkspaceView.GetTabInfo` now sends `navigationLocked` alongside the tab states, and `PublishTabPane` vetoes tool switching and greys out the other tools while it is set. Because that is the same flag C# already toggles in a `try`/`finally` around the upload, the publish tools unlock at the moment the upload really finishes or is cancelled — not when the browser guesses it has. The front end's own `isUploading` is cleared as soon as Cancel is clicked, and by any error line in the progress log, both of which would have unlocked too early.

The Apps tool gets this for free and now shares one path, so its `appsBusy` state and the `onBusyChange` callback threaded down through `AppPublisherScreen` are gone. As a side effect the lock now survives that screen remounting mid-action.

`SetTabsEnabled` stays a single shared flag rather than becoming a count. That was a deliberate call: with the tool switcher locked as well, the user can no longer start a second publish operation while one is running, so the overlap the old `RabPublishApi` comment worried about is unreachable. That comment is updated to say so. (Verified there is no programmatic navigation *to* the Publish tab — the only `ChangeTab(WorkspaceTab.publish)` call site is a user click on the top bar, which is itself refused while the tabs are disabled — so `PublishView.Activate()`'s unconditional re-enable cannot fire mid-operation.)

Deliberately not in scope, per decisions on the card: no tooltip/new localized string on the disabled buttons (matches how Apps behaves today), no re-entrancy guard on the upload endpoint, and no change to ePUB/BloomPUB/PDF preview generation (those abort cleanly when you leave, rather than needing to block).

Not verified in a running Bloom: the greying-out has not been watched during a real in-flight upload.

Ref: https://issues.bloomlibrary.org/youtrack/issue/BL-16654

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8174)

---
[Devin review](https://app.devin.ai/review/BloomBooks/BloomDesktop/pull/8174)

[Devin review](https://devinreview.com/BloomBooks/BloomDesktop/pull/8174)

### DIFF
--- a/src/BloomBrowserUI/publish/LibraryPublish/LibraryPublishSteps.tsx
+++ b/src/BloomBrowserUI/publish/LibraryPublish/LibraryPublishSteps.tsx
@@ -62,7 +62,11 @@ interface IReadonlyBookInfo {
 const kWebSocketEventId_uploadSuccessful: string = "uploadSuccessful";
 const kWebSocketEventId_uploadCanceled: string = "uploadCanceled";
 
-export const LibraryPublishSteps: React.FunctionComponent = () => {
+export const LibraryPublishSteps: React.FunctionComponent<{
+    // Called with true once the user commits to an upload and false when it is over. See the
+    // effect below for why the publish-tab host may only OR this into its lock.
+    onUploadingChange?: (uploading: boolean) => void;
+}> = (props) => {
     const selectedBookContext = React.useContext(SelectedBookContext);
     const [bookshelfHasProblem, setBookshelfHasProblem] = useState(false);
     const {
@@ -218,28 +222,62 @@ export const LibraryPublishSteps: React.FunctionComponent = () => {
     const [conflictIndex, setConflictIndex] = useState<number>(0);
 
     const [isUploading, setIsUploading] = useState<boolean>(false);
+    // Tell the publish-tab host as soon as the user commits to an upload, so it can lock the
+    // publish-tool switcher for the whole operation and not just the part C# knows about. C#
+    // takes its lock inside UploadBookAsync, but by then we have already made two API round
+    // trips (the subscription check and the "existing copy on server" query), during which the
+    // screen shows Cancel and a progress log while the tools were still live (BL-16654).
+    // The host OR-s this with C#'s lock rather than replacing it. That direction matters: this
+    // flag is not trustworthy as an *unlock* signal — clicking Cancel clears it while C# keeps
+    // working, and so does any error line in the progress log — but as an extra *lock* term it
+    // can only ever lock more than C# would, never less, so the unreliability is harmless here
+    // and C# stays the authority on when things reopen.
+    //
+    // Why an effect rather than doing this in the handler that starts the upload: isUploading has
+    // no single originating handler. It is set in uploadOneBook and cleared from four unrelated
+    // places — the Cancel button, the uploadSuccessful websocket, an error line in the progress
+    // box, and the collision dialog's cancel — so a render keyed on the resulting value is the
+    // only place that observes every transition. What we are doing is synchronizing an external
+    // system (the publish-tab host) to this state, which is what effects are for.
+    useEffect(() => {
+        props.onUploadingChange?.(isUploading);
+        // Never leave the host locked if this screen goes away mid-upload.
+        return () => props.onUploadingChange?.(false);
+        // eslint-disable-next-line react-hooks/exhaustive-deps
+    }, [isUploading]);
     function uploadOneBook() {
         setIsUploadComplete(false);
         setIsUploading(true);
-        get("libraryPublish/checkSubscriptionMatch", (result) => {
-            if (result.data.error) {
-                // The API already sent an error message
-                return;
-            }
-            get(
-                "libraryPublish/getUploadCollisionInfo?index=" + conflictIndex,
-                (result) => {
-                    if (result.data.error) {
-                        // The API already sent an error message
-                        return;
-                    }
-                    if (result.data.shouldShow) {
-                        setUploadCollisionInfo(result.data);
-                        showUploadCollisionDialog();
-                    } else post("libraryPublish/upload");
-                },
-            );
-        });
+        // If either pre-upload request dies at the transport level we get no reply and no progress
+        // message, so nothing else would ever clear isUploading. That used to leave only a stale
+        // Cancel button, but now it would also keep the other publish tools disabled, so clear it
+        // here. (An error *reported by* the API still arrives as a progress message and is handled
+        // by handleUploadError.)
+        get(
+            "libraryPublish/checkSubscriptionMatch",
+            (result) => {
+                if (result.data.error) {
+                    // The API already sent an error message
+                    return;
+                }
+                get(
+                    "libraryPublish/getUploadCollisionInfo?index=" +
+                        conflictIndex,
+                    (result) => {
+                        if (result.data.error) {
+                            // The API already sent an error message
+                            return;
+                        }
+                        if (result.data.shouldShow) {
+                            setUploadCollisionInfo(result.data);
+                            showUploadCollisionDialog();
+                        } else post("libraryPublish/upload");
+                    },
+                    handleUploadError,
+                );
+            },
+            handleUploadError,
+        );
     }
 
     const changeConflictIndex = (index: number) => {
--- a/src/BloomBrowserUI/publish/PublishTab/PublishTabPane.tsx
+++ b/src/BloomBrowserUI/publish/PublishTab/PublishTabPane.tsx
@@ -30,6 +30,7 @@ import {
 import { AboutDialogLauncher } from "../../react_components/aboutDialog";
 import { RegistrationDialogEventLauncher } from "../../react_components/registration/registrationDialogLauncher";
 import { RequiresSubscriptionOverlayWrapper } from "../../react_components/requiresSubscription";
+import { useWorkspaceTabInfo } from "../../react_components/TopBar/TopBar";
 
 export const CheckoutNeededScreen: React.FunctionComponent<{
     titleForDisplay: string;
@@ -108,9 +109,31 @@ export const PublishTabPane: React.FunctionComponent = () => {
     const [tabIndex, setTabIndex] = React.useState(
         kWaitForUserToChooseTabIndex,
     );
-    // True while the Apps tool has a Reading App Builder action running (its Cancel button is
-    // showing). While busy, switching to another publish tool is blocked so the operation is modal.
-    const [appsBusy, setAppsBusy] = React.useState(false);
+    // True while a long-running publish operation has made itself modal by locking navigation:
+    // a BloomLibrary upload, or one of the Apps tool's Reading App Builder actions. C# owns this
+    // flag (it is the same one that greys out the main workspace tabs), so the publish tools
+    // unlock at exactly the moment the operation really finishes or is cancelled — not when the
+    // browser guesses it has. See BL-16654.
+    const navigationLocked = useWorkspaceTabInfo().navigationLocked;
+    // The Web tool tells us directly when the user has committed to an upload, because C# does
+    // not take its lock until a couple of API round trips later, leaving a window where the
+    // screen already shows Cancel but the tools were still clickable (BL-16654).
+    const [uploadUnderway, setUploadUnderway] = React.useState(false);
+    // OR, never AND. C#'s flag is the authority on when an operation has really finished, and
+    // uploadUnderway is deliberately only ever an *additional* reason to lock: it is unreliable
+    // as an unlock signal (it clears the moment Cancel is pressed, and on any error line in the
+    // progress log) but adding it can only lock more than C# alone would, never less.
+    //
+    // Both are then gated on a tool actually showing. The lock exists to stop the user walking
+    // away from an operation in progress, and none can be in progress while the sentinel "no tool
+    // chosen yet" panel is up. That gate matters because the C# flag is shared with other
+    // subsystems — e.g. the Copyright and License dialog, reachable from this tab's own "Missing
+    // Copyright" link, posts editView/setModalState, which locks. Without it, a lock still set
+    // while tabIndex is the sentinel would grey out every tool at once and leave the user no way
+    // to choose one at all.
+    const publishToolsLocked =
+        (navigationLocked || uploadUnderway) &&
+        tabIndex !== kWaitForUserToChooseTabIndex;
     const appBuilderFeatureStatus = useGetFeatureStatus("AppBuilder");
     const setup = () => {
         setTabIndex(kWaitForUserToChooseTabIndex);
@@ -242,11 +265,11 @@ export const PublishTabPane: React.FunctionComponent = () => {
                             labelBackgroundColor={kPanelBackground}
                             selectedIndex={tabIndex}
                             onSelect={(newIndex) => {
-                                // While a Reading App Builder action is running (its Cancel button
-                                // is showing), the Apps operation is modal: veto switching to another
-                                // publish tool until it finishes or is cancelled. The main workspace
-                                // tabs are locked from C# (RabPublishApi) to match.
-                                if (appsBusy) {
+                                // While an upload or an Apps action is running (its Cancel button is
+                                // showing), the operation is modal: veto switching to another publish
+                                // tool until it finishes or is cancelled. The main workspace tabs are
+                                // locked from C# by the same flag.
+                                if (publishToolsLocked) {
                                     return false;
                                 }
                                 post("publish/switchingPublishMode");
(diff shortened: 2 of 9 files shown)


src/BloomBrowserUI/publish/PublishTab/PublishTabPane.tsx
```diff
@@ -108,9 +109,21 @@ export const PublishTabPane: React.FunctionComponent = () => {
...
     );
-    // True while the Apps tool has a Reading App Builder action running (its Cancel button is
-    // showing). While busy, switching to another publish tool is blocked so the operation is modal.
-    const [appsBusy, setAppsBusy] = React.useState(false);
+    // True while a long-running publish operation has made itself modal by locking navigation:
+    // a BloomLibrary upload, or one of the Apps tool's Reading App Builder actions. C# owns this
+    // flag (it is the same one that greys out the main workspace tabs), so the publish tools
+    // unlock at exactly the moment the operation really finishes or is cancelled — not when the
+    // browser guesses it has. See BL-16654.
+    const navigationLocked = useWorkspaceTabInfo().navigationLocked;
```
andrew-polk, line 117: [Devin] **Bug**: Publish tools can become clickable again in the middle of an upload

*(Posted by Claude Opus 5 from Andrew Polk's machine during preflight — Devin doesn't post its own findings to GitHub, so this mirrors it here as a resolvable thread.)*

Devin's report: the lock behind `navigationLocked` is a single shared boolean, and modal-dialog bookkeeping also writes it. Its scenario: while a book is uploading, open and close the Topic chooser from the Web tool's settings panel; `TopicChooserDialog` posts `editView/setModalState` on open and `false` on close, `EditingView.SetModalState` sees its depth counter return to zero and calls `SetTabsEnabled(true)`, which clears the upload's lock. Every publish tool and every main tab would become clickable again mid-upload, letting the user walk away from the upload this change exists to protect. It named `ConfirmDialog` in `LibraryPublishSteps.tsx` as a second route, and noted this would contradict the new comment in `RabPublishApi.cs`.

andrew-polk: *(Claude Opus 5 from Andrew Polk's machine during preflight)*

**Not an issue — both routes are gated to the Edit tab.** I checked each one:

- **Topic chooser.** `TopicChooserDialog` only posts `editView/setModalState` when its mode is `Mode.Edit` (`TopicChooserDialog.tsx:64`), and the Publish tab's settings panel opens it with `showTopicChooserDialog(Mode.Publish)` (`PublishTopic.tsx:62`). So opening and closing it from the Web tool posts nothing and cannot touch the lock.
- **The template-warning `ConfirmDialog`.** Same gate (`confirmDialog.tsx:43`), and the Publish usage passes no `dialogEnvironment` at all, so its mode is `undefined`. It also closes before the upload starts, so even if it did post, it would land before the lock is taken.

The wider point is fair, though, and worth recording rather than dismissing: the flag really is writable by dialog bookkeeping, and one dialog — Copyright and License — posts it with **no** mode gate and is reachable from the Web screen. The only reason that isn't a bug is that its entry point is the "Missing Copyright" link, and missing copyright makes `isReadyForUpload()` false, so no upload can be in flight when it's reachable. That is a real invariant, not a comfortable one, so it's written up for the developer alongside this run rather than left implicit.

The `RabPublishApi.cs` comment is about two *publish* operations overlapping, which is what the tool-strip lock now prevents; it doesn't claim dialog bookkeeping can't write the flag. Resolving.

src/BloomBrowserUI/publish/PublishTab/PublishTabPane.tsx
```diff
@@ -331,9 +344,14 @@ export const PublishTabPane: React.FunctionComponent = () => {
...
                                         key={index}
-                                        // Grey out the other publish tools while an Apps action is
-                                        // running, so it's clear the operation is modal.
-                                        disabled={appsBusy && tab.id !== "apps"}
+                                        // Grey out the other publish tools while a modal operation
+                                        // is running, so it's clear why they don't respond. The tool
+                                        // the operation belongs to stays looking normal, the same way
+                                        // C# leaves the active workspace tab looking active.
+                                        disabled={
+                                            publishToolsLocked &&
```
andrew-polk, line 362: [Devin] **Investigate**: All publish tools grey out if the shared lock is set while no tool is selected

*(Posted by Claude Opus 5 from Andrew Polk's machine during preflight — mirroring Devin's finding here as a resolvable thread.)*

Devin's report: `tabIndex` starts at, and is reset by `setup()` to, the sentinel `kWaitForUserToChooseTabIndex` = 6, which is outside the 0..5 range of real tools. So if the lock is set while `tabIndex` is that sentinel, *every* visible tool is disabled and `onSelect` vetoes as well — the user has no way to pick a publish tool at all until C# re-enables. Devin judged this unreachable in the normal flow (entering the Publish tab calls `SetTabsEnabled(true)` just before firing the event that resets `tabIndex`), but argued it is reachable in principle because the lock is a shared flag other subsystems also write, and the publish-tool switcher has no equivalent to the "safety net" in `PublishView.Activate()`.

andrew-polk: *(Claude Opus 5 from Andrew Polk's machine during preflight)*

**Fixed in 6a97f74e13.** Agreed, and the reachability argument turned out to be stronger than Devin knew: chasing it down found that the Copyright and License dialog — reachable from this very tab, via the Web screen's "Missing Copyright → Click to fix" link — posts `editView/setModalState` with **no** edit-tab-only gate on either the browser side or in `HandleSetModalState`. So a dialog opened from the Publish tab genuinely does drive this lock. (The specific mechanism Devin proposed, a late Edit-tab page-save transition, is not the one that reaches it.)

The fix gates the lock on a tool actually being selected:

```tsx
const publishToolsLocked =
    navigationLocked && tabIndex !== kWaitForUserToChooseTabIndex;
```

That is the semantically right condition, not just a guard: the lock exists to stop the user abandoning an operation in progress, and no operation can be in progress while the "no tool chosen yet" panel is up. During a real upload `tabIndex` is the Web tool, so the behavior this PR is about is unaffected.

Note for anyone reading Devin's page: it re-emitted this flag verbatim for the commit that fixes it, still anchored to the pre-fix line range and still stating the switcher "has no equivalent safety net" — so treat that repeat as stale, not as the fix having failed. Resolving.

andrew-polk:
*(Claude Opus 5 from Andrew Polk's machine during preflight)*

Consulted Devin on 2026-08-06 up to commit `6a97f74e133125cb4cd9c2ed7183d798b5552193`.

Devin was run twice this session, once per commit. Against `a414056cca` it raised one Investigate flag — that the publish tools could all grey out at once when no tool is yet selected. That was real, and `6a97f74e13` fixes it; the thread has the details.

The re-review of `6a97f74e13` raised one Bug, that dialog bookkeeping could release the upload's lock mid-upload. Assessed as a false positive: both dialogs it named only post `editView/setModalState` when their mode is `Mode.Edit`, and the Publish tab doesn't use that mode. Reasoning is on its thread. It also re-emitted the already-fixed Investigate flag verbatim against the pre-fix line numbers, so that repeat is stale rather than a second finding.

Six informational items in total across the two runs, not mirrored here (low signal). CodeRabbit is switched off in this repo by `.coderabbit.yml`, and there is no PR build/test CI, so the local suites stood in: TypeScript 667 passed, C# 3031 passed with one opt-in Reading-App-Builder test failing purely for lack of a local toolchain.

src/BloomBrowserUI/react_components/TopBar/TopBar.tsx
```diff
@@ -57,21 +67,28 @@ export function getActiveWorkspaceTab(tabStates: TabStates): WorkspaceTabId {
...
         publish: "hidden",
     },
+    navigationLocked: false,
 };
 
-export const TopBar: React.FunctionComponent = () => {
-    const state = useWatchApiObject<{ tabStates: TabStates }>(
+// Subscribes to what C# says about workspace navigation, kept in one place because several
+// screens in different browser controls need the same answer.
+export function useWorkspaceTabInfo(): IWorkspaceTabInfo {
```
andrew-polk, line 81: [Devin] **Bug**: New shared helper is written as an arrow constant instead of the required function syntax

*(Posted by Claude Opus 5 from Andrew Polk's machine during preflight — mirroring Devin's finding as a resolvable thread.)*

Devin's report: the new shared helper for reading workspace navigation state was declared as an arrow constant (`export const useWorkspaceTabInfo = ...`) even though it is a plain hook, not a React component. `src/BloomBrowserUI/AGENTS.md` asks for TypeScript `function` syntax for such functions, and the neighbouring non-component helper `getActiveWorkspaceTab` in the same file already follows that rule.

src/BloomBrowserUI/publish/LibraryPublish/LibraryPublishSteps.tsx
```diff
@@ -218,28 +222,62 @@ export const LibraryPublishSteps: React.FunctionComponent = () => {
...
+    // system (the publish-tab host) to this state, which is what effects are for.
+    useEffect(() => {
+        props.onUploadingChange?.(isUploading);
+        // Never leave the host locked if this screen goes away mid-upload.
+        return () => props.onUploadingChange?.(false);
+        // eslint-disable-next-line react-hooks/exhaustive-deps
+    }, [isUploading]);
     function uploadOneBook() {
         setIsUploadComplete(false);
         setIsUploading(true);
```
andrew-polk, line 250: [Devin] **Investigate**: The switcher lock can stick if a pre-upload API call never answers

*(Posted by Claude Opus 5 from Andrew Polk's machine during preflight.)*

Devin's report: `uploadOneBook` sets `isUploading` true before two GETs, and the recovery paths rely on the server having sent an error line to the progress websocket so that `ProgressBox`'s `onGotErrorMessage` clears it. If either GET instead fails at the transport level there is no callback and no progress message, so `isUploading` stays true indefinitely. Before this PR that only left a stale Cancel button; with the publish-tool strip following the same signal it would also leave the other tools disabled.

src/BloomBrowserUI/publish/LibraryPublish/LibraryPublishSteps.tsx
```diff
@@ -218,28 +222,62 @@ export const LibraryPublishSteps: React.FunctionComponent = () => {
...
+    // and C# stays the authority on when things reopen.
+    //
+    // Why an effect rather than doing this in the handler that starts the upload: isUploading has
+    // no single originating handler. It is set in uploadOneBook and cleared from four unrelated
+    // places — the Cancel button, the uploadSuccessful websocket, an error line in the progress
+    // box, and the collision dialog's cancel — so a render keyed on the resulting value is the
+    // only place that observes every transition. What we are doing is synchronizing an external
+    // system (the publish-tab host) to this state, which is what effects are for.
+    useEffect(() => {
+        props.onUploadingChange?.(isUploading);
```
andrew-polk, line 243: [Devin] **Investigate**: A stuck isUploading now disables every other publish tool, not just a stale Cancel button

*(Posted by Claude Opus 5 from Andrew Polk's machine during preflight.)*

Devin's report: having credited the new transport-level error callbacks for covering the two pre-upload GETs, it asked about one remaining route — the upload-collision dialog. Once that dialog is showing, `isUploading` stays true until either the dialog's `onCancel` runs or the upload actually starts. Its concern was `libraryPublish/uploadAfterChangingBookId`, which can answer `request.Failed("Can't fix ID because in TC")` without emitting an error progress line. Its closing ask: confirm that *every* exit from the collision dialog clears `isUploading`.

src/BloomBrowserUI/react_components/TopBar/TopBar.tsx
andrew-polk: *(Claude Opus 5 from Andrew Polk's machine during preflight)*

**Fixed — and Devin was right that I'd broken a documented rule.** The root `AGENTS.md` points at `src/BloomBrowserUI/AGENTS.md` for TypeScript style and I hadn't opened it. It says: *"For functions, prefer typescript 'function' syntax over const foo = () ==> functions."* The hook is now `export function useWorkspaceTabInfo(): IWorkspaceTabInfo { ... }`, matching `getActiveWorkspaceTab` a few lines above.

Reading that file also surfaced a second rule I'd missed — that a `useEffect` must carry a comment justifying why it is necessary — so the new effect in `LibraryPublishSteps` now explains why an effect is the right tool there (`isUploading` has no single originating handler; it is cleared from four unrelated places, so a render keyed on the resulting value is the only place that observes every transition).

Note the fix is *in the commit Devin analysed* (`40ce746790`) — it re-reported the finding against the same commit that resolves it, as it also did with two earlier findings. Resolving.

src/BloomBrowserUI/publish/LibraryPublish/LibraryPublishSteps.tsx
andrew-polk: *(Claude Opus 5 from Andrew Polk's machine during preflight)*

**Fixed in the same commit Devin analysed (`40ce746790`).** Both pre-upload requests now pass an error callback, so a transport-level failure clears the flag instead of leaving it set forever:

```ts
get("libraryPublish/checkSubscriptionMatch", (result) => { ... }, handleUploadError);
```

An error *reported by* the API still arrives as a progress message and is handled by `handleUploadError` through `ProgressBox`'s `onGotErrorMessage`, as before — this only adds the case where no reply arrives at all. Devin's own re-analysis of this commit confirms the two GETs are now covered. Resolving.

src/BloomBrowserUI/publish/LibraryPublish/LibraryPublishSteps.tsx
andrew-polk: *(Claude Opus 5 from Andrew Polk's machine during preflight)*

**Confirmed, and it's already covered — not an issue.** Devin asked for exactly the right thing, so here is the confirmation it wanted. Every exit from the upload-collision dialog clears `isUploading`, because they all route through the dialog's `onCancel`, which `LibraryPublishSteps` wires to `setIsUploading(false)`:

- **The Upload button, including the `uploadAfterChangingBookId` failure Devin names.** That `post` already passes a failure callback that calls `props.onCancel?.()` — `uploadCollisionDlg.tsx` even comments that this path exists specifically because changing the ID can fail for a book checked in to a Team Collection.
- **The Cancel button** calls `props.onCancel?.()` directly.
- **Dismissing the dialog** goes through the same `onCancel`.

So the "stuck" state Devin was worried about isn't reachable from the collision dialog. Resolving.

andrew-polk:
*(Claude Opus 5 from Andrew Polk's machine during preflight)*

Consulted Devin on 2026-08-06 up to commit `40ce746790bdbb569108d9877c0220a2e525127a`.

Two more rounds since the note above, and the branch has been squashed to a single commit (so earlier commit shas on this PR no longer exist; the review threads survived and are all resolved).

Round 3 (`ea033c412f`) raised two real findings, both fixed in `40ce746790`: the new `useWorkspaceTabInfo` hook used arrow-constant syntax where `src/BloomBrowserUI/AGENTS.md` asks for `function` syntax — a documented rule I had missed — and the publish-tool lock could stick if one of the two pre-upload API calls died at the transport level, which now clears via an error callback.

Round 4 (`40ce746790`) raised one genuinely new question, whether every exit from the upload-collision dialog clears `isUploading`. Confirmed that they all do, via the dialog's `onCancel`, including the `uploadAfterChangingBookId` failure path it specifically asked about. Not an issue; reasoning is on its thread.

Worth knowing when reading Devin's page for this PR: it re-emits findings against the very commit that fixes them. Both round-3 findings, and the two findings from earlier rounds, still appear in the round-4 results even though each was already fixed or refuted in the commit analysed. Every finding across all four rounds now has a documented outcome on a resolved thread here — treat those threads, not Devin's page, as the current state.

One finding has now been reported four times and refuted three: that dialog bookkeeping can unlock the tools mid-upload. It remains a false positive (`TopicChooserDialog` and `ConfirmDialog` both gate the `editView/setModalState` post on `Mode.Edit`, and the Publish tab doesn't use that mode), and since this round it is doubly moot — the Web screen's own upload signal is OR-ed into the lock, so even a cleared C# flag would keep the tools locked during an upload.

JohnThomson (approved):
> @JohnThomson reviewed 9 files and all commit messages, and resolved 5 discussions.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8174)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on andrew-polk).

PR #8175, opened 2026-08-07
Instrument urls built from a JavaScript undefined (BL-16666)

JohnThomson:
## What this is

Sentry issue [2699459502](https://bloom-app.sentry.io/issues/2699459502/) has years of "Cannot Find
File" reports for paths like `C:/Users/<user>/AppData/Local/Temp/undefined`.

BL-16577 framed these as C# building a temp path from a name that was `undefined`. **That framing is
wrong**, and a guard there would catch nothing. What actually happens is that some front-end code
does the equivalent of `element.src = aVariableThatIsUndefined`: the DOM turns the value into the
*text* `"undefined"` and asks the server for a file by that name. The server can only report that a
file called `undefined` is missing — never who asked — which is why finding BL-16447, the one
instance we have fixed, took so much guessing.

**This does not fix any particular instance.** It makes the next one name itself.

Two other things fell out of the investigation and are worth knowing:

- The temp folder is incidental. A bare `"undefined"` src resolves against the *document's* base url,
  and `ShouldReportFailedRequest` suppresses reports for anything under the current book folder — so
  these bugs are invisible when they happen on a book page. Everything reaching Sentry came from
  documents Bloom writes into temp and navigates to. **We have been seeing a filtered slice.**
- React `src={undefined}` is safe (React omits the attribute). Only the stringifying forms bite:
  `el.src = x`, `setAttribute("src", x)`, `new Audio(x)`, and template literals.

## Layer A — server side, with the referrer

`Referer` now flows through `IHttpListenerRequest` → `IRequestInfo` → the report, so a bogus request
names the page that issued it.

`undefined`/`null`/`NaN` are matched as a **whole path segment**, case-sensitively, so a real
`undefined.png`, a folder called `undefinedThings`, or a file called `Undefined` are untouched.

**This will increase reported volume.** These reports are no longer suppressed under the current book
folder — that suppression is exactly what was hiding most instances. That is the intent, and the
extra reports should be actionable, but whoever watches Sentry should expect it.

## Layer B — front-end interception, for the stack

New `lib/undefinedUrlDetector.ts`, installed from `lib/errorHandler.ts` — the module every bundle's
root already imports, so coverage is automatic. It intercepts `src` assignment,
`setAttribute("src")`, `fetch` and `XMLHttpRequest.open`, and reports with a JavaScript stack, which
names the offending line.

It is deliberately **passive**: it reports and lets the assignment proceed. The stack is what we
need and we get it either way, so there is no reason to change what the app does. (The plan called
for throwing in Debug/Alpha; I left that out for the same reason. Easy to add.)

Two things worth knowing if you change this later:

1. **The check takes the value, not a string.** At `img.src = x` the value arrives as the real
   `undefined` — the DOM stringifies it *inside* the native setter. My first version checked strings
   only and therefore caught nothing at the assignment site; the tests caught that.
2. **`fetch`/XHR had to be wrapped too.** BL-16447 came in through WaveSurfer *fetching* the url
   rather than putting it on an element, so element interception alone would have missed the one
   case we actually know about.

It also reports each distinct problem once, capped at 10 per session — each report is an http post,
and these bugs live in components that re-render.

## Worth a careful look

- **The merge with master had one conflict**, in `ReportMissingFile`: PR #8171 (now landed) added
  `GetBareNameDiagnostics` to the same format call this adds `extraDiagnostics` to. I kept **both** —
  they answer different questions, and a bare `undefined` request legitimately raises both.
- The `fetch` wrapper calls through on `window`, not `this`. A bare `fetch(url)` in a module gives
  `this === undefined`, and the browser rejects fetch invoked on anything but the window — that would
  have broken every fetch in Bloom.

## Testing

- C#: 78 pass across `BloomServerTests`, `RequestInfoTests`, `BloomFileLocatorTests` (13 new).
- Front end: 13 new tests in `undefinedUrlDetectorSpec`; full vitest suite green.
- typecheck, lint clean; production bundle compiles via `build/agent-vite.sh` (worth doing — this
  touches a module every bundle imports).

Ref: https://issues.bloomlibrary.org/youtrack/issue/BL-16666 (split from
https://issues.bloomlibrary.org/youtrack/issue/BL-16577)

[Devin review](https://devinreview.com/BloomBooks/BloomDesktop/pull/8175)

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8175)

### DIFF
--- a/src/BloomBrowserUI/lib/errorHandler.ts
+++ b/src/BloomBrowserUI/lib/errorHandler.ts
@@ -1,5 +1,6 @@
 import * as StackTrace from "stacktrace-js";
 import Axios from "axios";
+import { installUndefinedUrlDetector } from "./undefinedUrlDetector";
 
 // This file implements custom Bloom global error handling.
 // It should be imported by the root module in each bundle.
@@ -156,6 +157,11 @@ if (typeof window !== "undefined") {
         });
         return true; // suppress normal handling.
     };
+
+    // Watch for urls built out of a JavaScript value that wasn't ready. Installed here because
+    // every bundle's root module imports this file, which is exactly the coverage we want: the
+    // bug can be in any of our pages, and the server can't tell us which one. See BL-16577.
+    installUndefinedUrlDetector(reportError);
 }
 
 // Saving this as it MIGHT be useful if we decide to have another go at catching
--- a/src/BloomBrowserUI/lib/errorHandler.ts
+++ b/src/BloomBrowserUI/lib/errorHandler.ts
@@ -160,8 +160,23 @@ if (typeof window !== "undefined") {
 
     // Watch for urls built out of a JavaScript value that wasn't ready. Installed here because
     // every bundle's root module imports this file, which is exactly the coverage we want: the
-    // bug can be in any of our pages, and the server can't tell us which one. See BL-16577.
-    installUndefinedUrlDetector(reportError);
+    // bug can be in any of our pages, and the server can't tell us which one. See BL-16666.
+    installUndefinedUrlDetector((message, error) => {
+        // Source-map the stack the same way we do for unhandled errors above. Without this the
+        // report points into a bundle, which would defeat the purpose: the only reason to catch
+        // these in the browser at all is to name the line of *our* source that built the url.
+        // Unlike window.onerror we don't send a preliminary unmapped report first - nothing is
+        // about to crash and take the mapping with it, and these are diagnostics we deliberately
+        // keep quiet, so one report per problem is enough.
+        StackTrace.fromError(error)
+            .then((stackframes) =>
+                reportError(
+                    message,
+                    stackframes.map((sf) => sf.toString()).join("\n"),
+                ),
+            )
+            .catch(() => reportError(message, error.stack));
+    });
 }
 
 // Saving this as it MIGHT be useful if we decide to have another go at catching
--- a/src/BloomBrowserUI/lib/undefinedUrlDetector.ts
+++ b/src/BloomBrowserUI/lib/undefinedUrlDetector.ts
@@ -69,11 +69,27 @@ let installed = false;
 
 // Each report is an http post to the server, and the bug that triggers it is typically a render
 // that repeats - the Adjust Timings dialog in BL-16447 would have fired on every open. We only
-// need to learn about each site once, so report a given message once and stop altogether after a
+// need to learn about each site once, so report a given problem once and stop altogether after a
 // handful. Losing the repeat count costs us nothing: the point is to identify the code, and the
 // server-side count in Sentry already tells us how often it happens.
+//
+// "A given problem" has to mean the call site, not the message. Every component that sets an
+// image src to undefined produces the *same* message, so keying on that would report whichever
+// one happened to run first and hide every other one for the rest of the session - which is
+// exactly the kind of blind spot this whole feature exists to remove.
 const maxReports = 10;
 const alreadyReported = new Set<string>();
+
+/**
+ * What makes two reports "the same problem": the message plus the whole stack. We use the entire
+ * stack rather than trying to pick out "the caller" — any rule for finding that frame has to guess
+ * which frames are ours, and a wrong guess lands on a frame two different call sites share, which
+ * silently reinstates the blind spot. The stack is stable for repeats from one site, and if two
+ * routes reach the same bug, hearing about both is useful rather than noise.
+ */
+function reportKey(message: string, error: Error): string {
+    return message + "|" + (error.stack ?? "");
+}
 
 /** Exported only so tests can start from a clean slate. */
 export function resetReportingForTests(): void {
@@ -93,11 +109,13 @@ export function installUndefinedUrlDetector(report: Reporter): void {
     const check = (url: unknown, whereItWasSet: string) => {
         if (!isJavascriptValueUrl(url)) return;
         const message = describeBadUrl(url, whereItWasSet);
-        if (alreadyReported.has(message) || alreadyReported.size >= maxReports)
-            return;
-        alreadyReported.add(message);
         // The stack of *this* call is the whole point: it names the line that built the url.
-        report(message, new Error(message));
+        const error = new Error(message);
+        const key = reportKey(message, error);
+        if (alreadyReported.has(key) || alreadyReported.size >= maxReports)
+            return;
+        alreadyReported.add(key);
+        report(message, error);
     };
 
     guardSrcProperty(window.HTMLImageElement, "an image's src", check);
--- a/src/BloomBrowserUI/lib/undefinedUrlDetector.ts
+++ b/src/BloomBrowserUI/lib/undefinedUrlDetector.ts
@@ -58,7 +58,12 @@ export function describeBadUrl(url: unknown, whereItWasSet: string): string {
     );
 }
 
-type Reporter = (message: string, stack: string | undefined) => void;
+/**
+ * Takes the Error rather than its stack text, so the caller can source-map it. A raw stack points
+ * into a bundle, which is close to useless for a feature whose whole job is to name a line of our
+ * source. errorHandler does that mapping, since it already owns it for window.onerror.
+ */
+type Reporter = (message: string, error: Error) => void;
 
 let installed = false;
 
@@ -92,7 +97,7 @@ export function installUndefinedUrlDetector(report: Reporter): void {
             return;
         alreadyReported.add(message);
         // The stack of *this* call is the whole point: it names the line that built the url.
-        report(message, new Error().stack);
+        report(message, new Error(message));
     };
 
     guardSrcProperty(window.HTMLImageElement, "an image's src", check);
--- a/src/BloomBrowserUI/lib/undefinedUrlDetector.ts
+++ b/src/BloomBrowserUI/lib/undefinedUrlDetector.ts
@@ -1,0 +1,187 @@
+// Catches the moment a piece of our front-end code puts a JavaScript value it doesn't actually
+// have into a url.
+//
+// `element.src = someUndefinedVariable` doesn't fail: the DOM turns the value into the *text*
+// "undefined" and asks the server for a file by that name. The server can only report that some
+// file called "undefined" is missing, which tells us nothing about who asked for it - so these
+// have sat in Sentry for years without ever being traceable to a line of code (BL-16577). We
+// already fixed one instance the hard way (BL-16447, where the Adjust Timings dialog rendered
+// before its audio url was ready); this exists so the next one names itself.
+//
+// The interception below is deliberately passive: it reports and then lets the assignment
+// proceed exactly as before. What we need from it is the stack, and we get that either way, so
+// there is no reason to change what the app does.
+
+/// The exact strings JavaScript produces when these values are turned into text. We match them
+/// case-sensitively and only as a whole url or a whole path segment, so a real file that happens
+/// to be called "Undefined.png", or a folder called "undefinedThings", is left alone.
+const javascriptValueStrings = ["undefined", "null", "NaN"];
+
+/**
+ * True if this will become a url that is, or contains a path segment that is, the text form of a
+ * JavaScript value. Nothing legitimately loads such a url, so a true here is always one of our bugs.
+ *
+ * Takes the value rather than a string on purpose. At an assignment like `img.src = x` the value
+ * arrives here as the real `undefined`; it is the DOM that turns it into the text "undefined" a
+ * moment later. By the time anyone could see a string it is already too late to know who did it.
+ */
+export function isJavascriptValueUrl(url: unknown): boolean {
+    // These are exactly the values that stringify to the text we are looking for.
+    if (url === undefined || url === null) return true;
+    if (typeof url === "number") return Number.isNaN(url);
+    // Anything else that isn't a string (a URL object, say) stringifies to something real.
+    if (typeof url !== "string" || url.length === 0) return false;
+
+    // Ignore any query string or fragment; only the path can name a file.
+    const path = url.split(/[?#]/)[0];
+    return path
+        .split(/[/\\]/)
+        .some((segment) => javascriptValueStrings.includes(segment));
+}
+
+/**
+ * What we send to the server. Kept separate from the interception so it can be tested, and so the
+ * message reads the same wherever it was caught.
+ */
+export function describeBadUrl(url: unknown, whereItWasSet: string): string {
+    return (
+        `A url was built from a JavaScript value that wasn't ready: ${whereItWasSet} was set to "${String(url)}". ` +
+        `Bloom will now ask the server for a file by that name and fail. See BL-16577.`
+    );
+}
+
+type Reporter = (message: string, stack: string | undefined) => void;
+
+let installed = false;
+
+// Each report is an http post to the server, and the bug that triggers it is typically a render
+// that repeats - the Adjust Timings dialog in BL-16447 would have fired on every open. We only
+// need to learn about each site once, so report a given message once and stop altogether after a
+// handful. Losing the repeat count costs us nothing: the point is to identify the code, and the
+// server-side count in Sentry already tells us how often it happens.
+const maxReports = 10;
+const alreadyReported = new Set<string>();
+
+/** Exported only so tests can start from a clean slate. */
+export function resetReportingForTests(): void {
+    alreadyReported.clear();
+}
+
+/**
+ * Installs the interception. Safe to call more than once; only the first call does anything, so a
+ * bundle that pulls in the bootstrap twice doesn't end up double-reporting.
+ *
+ * @param report how to get the problem back to us - errorHandler's reportError in real use.
+ */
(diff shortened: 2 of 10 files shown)


src/BloomBrowserUI/lib/undefinedUrlDetectorSpec.ts
JohnThomson, line 1: [Devin] **Bug**: A new unit test looks for the wrong issue number and always fails

Devin found that the spec asserted the warning text contains `BL-16577`, while `describeBadUrl` produces `BL-16666` — so that test could never pass, and the PR's claim that the front-end suite was green was wrong.

Correct, and my mistake. When I repointed the card references from BL-16577 to the new card I re-ran the C# suite but not the front-end one, which is exactly how it got through. Fixed in `284dde3ca`: the assertion now looks for BL-16666, and `pnpm vitest run lib/undefinedUrlDetectorSpec.ts` passes 13/13.

src/BloomExe/web/BloomServer.cs
JohnThomson, line 1: [Devin] **Investigate**: Removing the book-folder suppression surfaces toasts to ordinary users, not just Sentry entries

Devin's point: the new early `return true` means these requests always reach `ReportMissingFile`, whose last branch calls `NonFatalProblem.Report(ModalIf.Beta, PassiveIf.All, ...)` — a toast on **every** channel and a modal on alpha/beta. The PR described the consequence as more Sentry volume and said nothing about a user-visible notification appearing on pages that previously failed silently.

This is a good catch and I have acted on it. There is nothing a user can do about our own bug, so interrupting them is pure harm. Fixed in `284dde3ca`: a request whose path contains a JavaScript value now reports with `ModalIf.None`/`PassiveIf.None` — still logged, still sent to Sentry, invisible to the user. It also gets its own short message ("Url built from a JavaScript value") so it forms its own Sentry issue rather than being buried in the general "Cannot Find File" one, which makes triage easier.

src/BloomBrowserUI/lib/undefinedUrlDetector.ts
JohnThomson, line 1: [Devin] **Investigate**: fetch interception is likely a no-op under jsdom, so the two fetch tests may not exercise it

Devin suspected `guardFetch` bails out under vitest's jsdom environment (no `window.fetch`), so the wrapper is never installed and the fetch tests pass vacuously.

Checked, and that isn't what happens here. The test asserts `report` was called exactly once **with a message containing `a fetch()`** — and that string is produced only inside `guardFetch`. If the wrapper had not been installed, or had not run, the assertion would fail rather than pass vacuously. It passes, so `window.fetch` does exist in this environment and the interception is genuinely exercised. (Devin's closing note is right that production coverage would be unaffected either way, since in a real browser the global and `window.fetch` are the same object.)

src/BloomExe/web/BloomServer.cs
JohnThomson, line 1: [Devin] **Bug**: Extra troubleshooting detail never reaches the report for bad urls

Devin's point: `ReportMissingFile` returns early for exactly the JavaScript-value case, so by the time control reaches the image and general branches, `extraDiagnostics` is always empty. That made the "both kinds of detail" comment false — the bare-name detail never reached a report about a JavaScript-value url — and left the general branch with a format argument that could never be filled.

Correct, and a real consequence of how I resolved the merge with PR #8171 (which had added `GetBareNameDiagnostics` to the same call). Fixed in `c2438fb07`: the early branch now emits both kinds of detail, since a bare `undefined` genuinely has no directory *and* is a JavaScript value; the later branches ask only for the bare-name detail, which is the only one they can ever have; and the image branch no longer takes an argument it never had.

src/BloomBrowserUI/lib/undefinedUrlDetector.ts
JohnThomson, line 1: [Devin] **Investigate**: Prototype patching only covers the window the bundle runs in, not iframe documents

Devin's point: the detector patches the prototypes of the window `errorHandler` is loaded in. Bloom's edit screen puts the book page and toolbox in iframes, and elements created inside a frame are instances of *that* frame's constructors — so parent-frame code reaching into a child frame's document is not intercepted.

That is accurate and worth having on the record. It is not a defect so much as the shape of the technique: each frame that runs one of our bundles installs its own copy, because every bundle's root module imports `errorHandler` (the edit page's own bundle and the toolbox bootstrap both do), so the frames that matter are covered. What is genuinely not covered is parent-frame code creating elements in a child's document, which bookEdit does do in places.

Rather than chase full coverage — which would mean hooking every frame we create and is a lot of machinery for a diagnostic — I have written the limitation into the module's header comment in `c2438fb07`, so that a future reader does not take silence here as proof that nothing went wrong.

src/BloomExe/web/BloomServer.cs
JohnThomson, line 1: [Devin] **Investigate**: The two paths that decide 'is this a JavaScript-value url' use two separately-derived strings

Devin's point: `ShouldReportFailedRequest` decided from `GetLocalPathWithoutQuery(info)`, while the branch that keeps the report silent decided from the `localPath` that `ProcessAnyFileContent` happened to be holding. Same string today, but computed independently — and if they ever diverged, the request would fall through to the `ModalIf.Beta`/`PassiveIf.All` branch, which is exactly the user-visible toast the quiet branch exists to prevent.

Agreed, and worth fixing rather than noting: the failure mode is silent and user-visible. Fixed in `79858a523` — both now ask `IsJavascriptValueRequest(info)`, deriving the decision once from the request itself, and `ReportMissingFile` takes the request rather than a pre-computed diagnostics string (passing the string is what made the divergence possible in the first place).

src/BloomBrowserUI/lib/undefinedUrlDetector.ts
JohnThomson, line 1: [Devin] **Bug**: Several new internal helpers are missing the explanatory comments the project requires

Devin noted that `guardSetAttribute`, `urlOfFetchInput` and `guardXmlHttpRequest` had no doc comment, while their neighbours did — against AGENTS.md's "All public methods should have a comment. So should most private ones!"

Fair, and easy. Added in `79858a523`.

src/BloomBrowserUI/lib/errorHandler.ts
JohnThomson, line 1: [Devin] **Investigate**: Reported stacks are raw bundle stacks, not source-mapped like the rest of errorHandler

Devin's point: `window.onerror` runs its stacks through stacktrace-js so a report names a line in our source, but the new detector was handing over a raw `new Error().stack`, which points into a bundle.

This was the most valuable thing in the review. The entire purpose of the front-end layer is to name the line that built the bad url, and an unmapped bundle stack largely defeats that — the feature would have looked like it worked while producing reports nobody could act on.

Fixed in `b8c9cb74c`: the detector now hands over the `Error` itself rather than its stack text, and errorHandler (which already owns the mapping) source-maps it before reporting, falling back to the raw stack if the mapping fails. Unlike `window.onerror` it does not send a preliminary unmapped report first — nothing is about to crash and take the mapping with it, and these are deliberately quiet diagnostics, so one report per problem is enough.

src/BloomBrowserUI/lib/undefinedUrlDetector.ts
JohnThomson, line 1: [Devin] **Bug**: New front-end helper declared with arrow-function syntax against the project's TypeScript style rule

Not an issue — the rule doesn't say what Devin reads it as saying, and the surrounding code shows it.

`src/BloomBrowserUI/AGENTS.md` line 19 reads "Always use arrow functions and function components in React". That is about React components, not every function in a `.ts` file. The decisive evidence is the file this module is installed from: `lib/errorHandler.ts` declares `export function reportError`, `export function reportPreliminaryError` and `export function shouldIgnoreUnhandledError` — plain function declarations, in the same folder, following the same rule. Converting this one module to arrow functions would make it the odd one out rather than the conforming one.

src/BloomExe/web/BloomServer.cs
JohnThomson, line 1: [Devin] **Investigate**: Server-side reporting has no per-session cap, unlike the front-end layer

Devin's point: the front-end detector reports each distinct problem once and stops after ten, but the server side has no such cap, so a repeating bug produces a report on every request.

Real asymmetry, but I have deliberately not matched it, and this is the one item I am leaving for John rather than deciding myself, since it is a judgement about how much Sentry volume we want.

The asymmetry has a reason: the front-end cap exists because each report is an http post from the browser, which is a cost we control and pay per occurrence. On the server the report goes to Sentry, which already groups and rate-limits identical events, and the per-event count is *information* — it tells us how often a given bad url is actually being requested, which the front-end cap deliberately throws away. Capping both would lose that. It does mean this PR increases Sentry volume, which the PR description calls out.

Left open pending John's decision; see the preflight report.

src/BloomBrowserUI/lib/undefinedUrlDetector.ts
JohnThomson, line 1: [Devin] **Bug**: Only the first bad-url report of each kind is ever sent, so later problems stay invisible

Right, and it mattered more than it looks. The once-only rule was keyed on the *message* — but every component that sets an image src to undefined produces the same message, so we would have reported whichever component happened to run first and hidden every other one for the rest of the session. That is precisely the blind spot this whole feature exists to remove.

Fixed in `4607a0bb6`: the key is now the message plus the stack. Deliberately the *whole* stack rather than an attempt to pick out "the caller" — any rule for finding that frame has to guess which frames are ours, and a wrong guess lands on a frame two different call sites share, silently reinstating the same blind spot. I wrote the clever version first and a new test caught it doing exactly that.

There is a real consequence worth stating: this leaves the hard cap, not the collapse, as the actual flood protection, since identical repeats only collapse when the stack is identical (a re-render gives that; a caller at a different line does not). The reworked test now says so rather than implying a guarantee we don't have.

src/BloomExe/web/BloomServer.cs
JohnThomson, line 1: [Devin] **Investigate**: "null" as a path segment could match legitimate third-party urls

Devin's concern: a real url could contain a `null` path segment and be misread as our bug.

Low risk here, and I have left it as is. This check only ever sees requests that reached Bloom's own local server *and already failed to find a file* — we are not inspecting arbitrary third-party traffic, and a genuine file or folder named exactly `null` (lowercase, no extension) inside a Bloom book or our own shipped files would be a remarkable thing to find. The cost of a false positive is also small: a quiet, user-invisible Sentry entry, not a broken request or a dialog. Worth remembering if we ever see a puzzling one, which is why it is on the record here.

src/BloomBrowserUI/lib/errorHandler.ts
JohnThomson, line 1: [Devin] **Bug**: New diagnostic reports pop up as JavaScript-error notifications for alpha users

Devin is right on the facts and this one is going to John rather than being decided here.

The front-end layer reports through `reportError`, which posts to `common/error`. That handler (`CommonApi.ReportJavascriptError`) calls `NonFatalProblem.Report(ModalIf.None, PassiveIf.Alpha, "A JavaScript error occurred", ...)` — so on the alpha channel every detection shows a toast, labelled as a JavaScript error. That partly undoes the care taken on the server side, where this class was deliberately made silent.

It is a genuine judgement call rather than an obvious defect, which is why I have not just picked one:

- Alpha users are our own testers, and a toast is arguably how they find out something is wrong — that is what the channel is for.
- But the label is misleading (nothing has actually gone wrong for the user), and up to ten toasts a session while we are hunting a bug is a poor experience for the people doing the hunting.

Fixing it properly means a separate quiet endpoint alongside `common/error`, which is more than a preflight should decide on its own. Left open; see the preflight report.

JohnThomson: [Claude Opus 5 (1M context)] Decided by John: **keep the notice on alpha** — our testers seeing it is the point.

So no change here. Alpha users will get a small "A JavaScript error occurred" notice for each detection (up to the ten-per-session cap), and that is intentional: alpha is the channel where we want a human to notice there is something to report. The server-side half stays silent as built, so ordinary users are unaffected either way.

src/BloomExe/web/BloomServer.cs
JohnThomson: [Claude Opus 5 (1M context)] Decided by John: **leave the server side uncapped** — the counts are useful.

So no change here. The asymmetry with the front-end cap is deliberate: Sentry already groups identical events, and the per-event count tells us how often a given bad url is actually being requested, which is exactly the frequency data we are collecting these reports to get. A cap would discard it.

andrew-polk (commented):
> @andrew-polk reviewed 10 files and all commit messages, made 1 comment, and resolved 13 discussions.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8175)* status: all files reviewed, 1 unresolved discussion (waiting on JohnThomson).

___
*[`src/BloomExe/web/BloomServer.cs` line 1584 at r5](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8175#-P29Hc6WBVOwcVcpH11D:-P29Hc6WBVOwcVcpH11E:b1l3mp6) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/4607a0bb67fa13e3537175b7e299934467b96acd/src/BloomExe/web/BloomServer.cs#L1584)):*
> ```Smalltalk
>                 info.RawUrl
>             );
>         }
> ```

You and I don't always agree on the value of extracting stuff, so I'm just posing a question more than a request...
Have you considered trying to pull this error handling stuff mostly out of BloomServer to a separate file?

JohnThomson (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8175)* status: all files reviewed, 1 unresolved discussion (waiting on andrew-polk).

___
*[`src/BloomExe/web/BloomServer.cs` line 1584 at r5](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8175#-P29Hc6WBVOwcVcpH11D:-P2De5VX7n2-w4YWUn-T:bz8pwrg) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/4607a0bb67fa13e3537175b7e299934467b96acd/src/BloomExe/web/BloomServer.cs#L1584)):*

You and I don't always agree on the value of extracting stuff, so I'm just posing a question more than a request...
Have you considered trying to pull this error handling stuff mostly out of BloomServer to a separate file?
</blockquote></details>

I didn't really consider it, no. The little bit added here doesn't seem to warrant it. Server code is handling errors all over the place, so it seems like it would be very hard to draw a line; if we did, probably a lot more existing code should move, too. I would think it we want to do that it would deserve its own card.

src/BloomBrowserUI/lib/errorHandler.ts
JohnThomson: [Devin] **Investigate**: Bundle coverage depends on transitive imports

Devin's point: several of our Vite entry points don't import `errorHandler` directly, so whether a given page gets the undefined-url detector depends on it being imported indirectly. Devin suggests checking each generated bundle.

JohnThomson: [Claude Opus 5.5] Not an issue for this PR. The detector is installed as a side effect of loading `errorHandler`, so an indirect import installs it just as well as a direct one. The pages it covers are exactly the pages that already get Bloom's global JavaScript error reporting (`window.onerror`, in the same file), which rests on the same convention: "It should be imported by the root module in each bundle." A bundle that somehow misses `errorHandler` is already missing that error reporting too, which is a separate, existing gap. This change doesn't make it any wider. Resolving.

src/BloomBrowserUI/lib/undefinedUrlDetectorSpec.ts
JohnThomson: [Devin] **Investigate**: Focused tests were unavailable

Devin could not run the focused browser and server tests because its environment has no `pnpm` or `dotnet`, so it asks that they be validated elsewhere.

JohnThomson: [Claude Opus 5.5] They were run locally at this HEAD: the detector spec (14 tests) and the full vitest suite pass, and the C# suite covers the server side (`BloomServerTests`, `RequestInfoTests`). Nothing to change in the code. Resolving.

JohnThomson:
[Claude Opus 5.5] Consulted Devin on 2026-09-23 up to commit adb39495d508081606d862fba3c5899b3c46074b (after rebasing onto master). Devin did a full re-review, and nearly everything it raised repeats a finding that already has a resolved thread here. It raised two new Investigate flags, both answered and resolved on their threads: bundle coverage through indirect imports, and its own sandbox being unable to run the tests. One new informational note, a code comment naming a variable that no longer exists, was fixed in adb39495d5.

andrew-polk (approved):
> @andrew-polk reviewed 5 files and all commit messages, and resolved 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8175)* status: all files reviewed, 1 unresolved discussion.

andrew-polk (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8175)* status: all files reviewed, 1 unresolved discussion.

PR #8176, opened 2026-08-07
Remove a deleted canvas element's hint bubbles with it (BL-16648)

JohnThomson:
Fixes one of the 6.4 findings reported on BL-16648 — **not** the whole card:

> Deleting a "topic" caption box leaves the "choose topic" link showing on the page. (Probably common to custom pages, not just new one.)

The other two findings in that comment (spreadsheet export/import of custom-layout pages; the right-click menu losing items when a text box is set to Topic/Languages) are untouched here.

## What was wrong

The "Choose topic" control is a jQuery **qtip** hint bubble. `BloomHintBubbles` gives one to every `[data-hint]` element, and the topic field's hint carries `data-functiononhintclick="showTopicChooser"`, which turns the bubble's text into the link that opens the topic chooser.

qtip renders that bubble into `#page-scaling-container`, *not* inside the element it annotates, and it only tears itself down when the target is removed **through jQuery** (it binds `remove.qtip-<id>` and relies on the `$.cleanData` patch). Comical removes a canvas element with a native `container.removeChild(...)`, so neither hook fires and the bubble is orphaned — still visible, still clickable — until the page is reloaded.

## The fix

`deleteCanvasElement` now calls `BloomSourceBubbles.removeSourceBubbles(textOverPicDiv)` immediately before `Comical.deleteBubbleFromFamily` takes the element out, i.e. while the divs the bubbles are attached to still exist (which is how that helper finds them).

Notes on the placement:

- It is **scoped to the element being deleted**, so a page with more than one topic block loses only the corresponding control.
- It sits **after** the background-image early return, which does *not* remove anything from the DOM (it swaps in `placeHolder.png`), so those bubbles are correctly left alone.
- `deleteCanvasElement` is the single funnel for every delete path — context menu, keyboard Delete, and the background-image conversion in `canvasControlRegistry` — so all of them are covered.
- It also cleans up source/hint bubbles on ordinary text canvas elements, which had the same leak (just less visible, because those bubbles are focus-only rather than always-shown).

## Testing

New vitest test `CanvasElementDelete.test.ts`: two canvas elements each holding a topic field with a real qtip; delete one and assert its bubble is gone and the other's survives. Confirmed it fails without the fix.

Also verified interactively in a running Bloom on a real custom cover page: duplicated the topic canvas element to get two "Choose topic" bubbles, deleted one, and watched only that one disappear — immediately, with no page reload. With the fix reverted, the orphan bubble was still there a second later.

Ref: https://issues.bloomlibrary.org/youtrack/issue/BL-16648

[Devin review](https://devinreview.com/BloomBooks/BloomDesktop/pull/8176)

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8176)

### DIFF
--- a/src/BloomBrowserUI/bookEdit/js/CanvasElementDelete.test.ts
+++ b/src/BloomBrowserUI/bookEdit/js/CanvasElementDelete.test.ts
@@ -1,0 +1,97 @@
+import { describe, expect, test, vi } from "vitest";
+import $ from "jquery";
+import "../../lib/jquery.qtip.js"; // puts qtip into $.fn
+
+// Comical wants paper.js and a real <canvas>, which jsdom doesn't give us. deleteCanvasElement
+// only needs it to take the element out of the DOM, which is what the real
+// deleteBubbleFromFamily does at the end of its work.
+vi.mock("comicaljs", () => ({
+    Bubble: class {},
+    Comical: {
+        setSelectorForBubblesWhichTailMidpointMayOverlap: () => {},
+        activateElement: () => {},
+        update: () => {},
+        deleteBubbleFromFamily: (
+            element: HTMLElement,
+            container: HTMLElement,
+        ) => container.removeChild(element),
+    },
+}));
+
+// This import deliberately comes after the vi.mock call above, so that the module graph
+// it pulls in gets the stubbed comicaljs.
+import { CanvasElementManager } from "./CanvasElementManager";
+
+// Attach a qtip the way BloomHintBubbles.makeHintBubbleCore does for a
+// data-derived="topic" field: always showing, and rendered into the page scaling
+// container rather than inside the element it annotates.
+function makeTopicHintBubble(topicField: HTMLElement): void {
+    $(topicField).qtip({
+        content: "Choose topic",
+        position: { container: $("div#page-scaling-container") },
+        show: { ready: true },
+        hide: { event: false },
+        style: { classes: "topic-chooser-hint-bubble" },
+    });
+}
+
+function setUpTwoTopicCanvasElements(): HTMLElement[] {
+    document.body.innerHTML = `
+        <div id="page-scaling-container">
+            <div class="bloom-page">
+                <div class="bloom-canvas">
+                    <div class="bloom-canvas-element" id="ce1">
+                        <div class="coverBottomBookTopic" data-derived="topic"
+                             data-hint="Choose topic">Fiction</div>
+                    </div>
+                    <div class="bloom-canvas-element" id="ce2">
+                        <div class="coverBottomBookTopic" data-derived="topic"
+                             data-hint="Choose topic">Fiction</div>
+                    </div>
+                </div>
+            </div>
+        </div>`;
+    const canvasElements = Array.from(
+        document.querySelectorAll<HTMLElement>(".bloom-canvas-element"),
+    );
+    canvasElements.forEach((canvasElement) =>
+        makeTopicHintBubble(
+            canvasElement.querySelector("[data-derived]") as HTMLElement,
+        ),
+    );
+    return canvasElements;
+}
+
+describe("CanvasElementManager.deleteCanvasElement", () => {
+    test("removes the deleted element's topic chooser bubble but not another one's", async () => {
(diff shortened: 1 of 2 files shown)


src/BloomBrowserUI/bookEdit/js/canvasElementManager/CanvasElementDelete.test.ts
```diff
@@ -0,0 +1,97 @@
...
+        ),
+    );
+    return canvasElements;
+}
+
+describe("CanvasElementManager.deleteCanvasElement", () => {
+    test("removes the deleted element's topic chooser bubble but not another one's", async () => {
+        const [canvasElement1, canvasElement2] = setUpTwoTopicCanvasElements();
+        // qtip renders after its default show delay, so let that happen.
+        await new Promise((resolve) => setTimeout(resolve, 200));
```
JohnThomson: [Devin] **Investigate**: Test relies on a 200ms sleep for qtip rendering

The test waits with `setTimeout(..., 200)` for qtip's default show delay. The repo's front-end guidance discourages timeouts in tests as slow/fragile (stated for the Playwright UI tests, so it may not strictly bind vitest). A deterministic alternative would be to configure the qtip with `show: {ready: true, delay: 0}` / `hide` options that render synchronously, or to poll for the rendered `div.qtip` rather than sleep a fixed time; on a loaded CI machine 200ms could occasionally be insufficient and the test would fail in the setup sanity check.

JohnThomson: [Claude Opus 5 (1M context)] Fixed in `413c6ebc835b0249e6fe5a419b3be064cb63c540`: the test now polls for the rendered bubbles (5s deadline) instead of sleeping a fixed 200ms, so a loaded machine can't make it fail in setup. It also fails with a message saying how many bubbles actually rendered, rather than a bare assertion.

JohnThomson:
[Claude Opus 5 (1M context)] Consulted Devin on 2026-08-07 17:18 UTC up to commit `413c6ebc835b0249e6fe5a419b3be064cb63c540`.

First pass (`f6db24e`): no bugs, one Investigate flag — the new test slept a fixed 200ms waiting for the hint bubbles to appear, which could be too short on a busy machine and would then fail in setup rather than tell us anything. Fixed in `413c6eb` (it now polls for them); thread replied to and resolved.

Re-review of `413c6eb`: clean — no bugs and nothing flagged for investigation. Five informational observations, none of which need action: the cleanup destroys every tooltip inside the element being deleted (correct — the whole subtree is going away); the background-image delete path deliberately keeps its bubbles (it doesn't remove the element, it swaps in a placeholder); the test polls with setTimeout; the test stubs out the editing refresh; and the test's Comical stub matches today's real behaviour but wouldn't notice if Comical later removed elements a different way.

CI (pr-automation) passed on both commits. CodeRabbit does not run here — `.coderabbit.yml` has auto-review turned off.

andrew-polk (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8176)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on JohnThomson).

andrew-polk:
Approved, but shouldn't this target 6.4?

JohnThomson (commented):
Yes.

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8176#-:-OzSeN969fZmSlAo_YM8:bb77ufg)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on JohnThomson).

andrew-polk (approved):
> @andrew-polk partially reviewed 3 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8176)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on JohnThomson).

andrew-polk (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8176)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on JohnThomson).

PR #8177, opened 2026-08-07
Take OffScreenBrowser's thread out of WinForms' application lifetime (BL-16668)

andrew-polk:
`Bloom.exe createArtifacts` was dying with `ObjectDisposedException` on the epub step, so the
harvester failed **every** book with epubs enabled and uploaded no artifacts at all — not even
the .bloompub it had already built successfully. Reported downstream as BH-7836.

### What was happening

`Application.Run` registers the loop it starts as an **application** message loop. `OffScreenBrowser`
(introduced in `48d8e2588c` for BL-12614 / BL-13120) runs one on its own dedicated STA thread. The
CLI path never calls `Application.Run` itself — it waits with `Thread.Sleep` + `Application.DoEvents`
— so that browser thread's loop is the *only* application message loop in the process. Therefore:

1. `createArtifacts` builds the .bloompub first; `BloomPubMaker` does `using (var helper = new PublishHelper())`.
2. Disposing it calls `PublishHelper.ReleaseBrowser()` → `OffScreenBrowser.Dispose()` → `ApplicationContext.ExitThread()`.
3. The process's only application message loop ends, so WinForms raises `Application.ApplicationExit`.
4. `ApplicationContainer.OnApplicationExit` disposes the container — the **parent** of `ProjectContext._scope`.
5. The epub step, which runs next, evaluates `s_projectContext.BookServer` → `ObjectDisposedException`.

In the GUI this never fired, because the main thread's loop is still running. It only bites a
non-GUI entry point.

### The fix

Rather than teach one more listener to distrust the event, stop telling the lie: pump the browser
thread with a bare `GetMessage`/`TranslateMessage`/`DispatchMessage` loop and end it with
`PostQuitMessage` from the teardown we already post onto that thread. That dispatches everything
the thread actually needs — the `WindowsFormsSynchronizationContext`'s marshaling control and the
WebView2's windows and async completions — while staying invisible to WinForms'
application-lifetime bookkeeping.

This removes the whole bug class: no non-GUI entry point (harvester, bulk upload, a future CLI
verb) can now be told the application is exiting merely because a worker finished with a browser.
`ReleaseBrowser`'s resource-saving intent is untouched — the browser and its thread are still torn
down per batch of page checks. Since `Application.Run` also disposed the thread's windows as its
loop ended, the pump now disposes the synchronization context itself, so we don't leave a `Control`
to be finalized on a dead thread.

### Regression test

Every existing `CreateArtifactsCommandTests` case requested a single artifact, which is exactly why
CI stayed green through this. The harvester always passes `--bloomdOutputPath`/
`--bloomDigitalOutputPath` **and** `--epubOutputPath` in one run, and only that combination exposes
the bug — with one artifact, nothing runs after the premature disposal. The new test asks for both
and fails on unmodified `master` with the production stack trace, frame for frame:

```
Expected: Success   But was: EpubException
System.ObjectDisposedException: Instances cannot be resolved and nested lifetimes cannot be created
from this LifetimeScope as it (or one of its parent scopes) has already been disposed.
   at Bloom.ProjectContext.get_BookServer() ... ProjectContext.cs:840
   at Bloom.CLI.CreateArtifactsCommand.CreateEpubArtifact ... CreateArtifactsCommand.cs:401
```

The test commit is deliberately byte-identical to the one on `BL-16668-harvester-exit-guard`, which
carries the minimal alternative fix (guarding `OnApplicationExit` on `Program.RunningHarvesterMode`)
for comparison.

Ref: https://issues.bloomlibrary.org/youtrack/issue/BL-16668

[Devin review](https://devinreview.com/BloomBooks/BloomDesktop/pull/8177)

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8177)

### DIFF
--- a/src/BloomExe/Publish/OffScreenBrowser.cs
+++ b/src/BloomExe/Publish/OffScreenBrowser.cs
@@ -112,10 +112,13 @@ namespace Bloom.Publish
                 }
                 finally
                 {
-                    // Application.Run disposed this thread's windows as its loop ended; since we no longer
-                    // use it, dispose the synchronization context (and so its hidden marshaling control)
-                    // ourselves. We are on the owning thread, the only one allowed to do that, and WM_QUIT
-                    // is only retrieved once the queue is otherwise empty, so all posted work has run.
+                    // As its loop ended, Application.Run's teardown destroyed this thread's window
+                    // HANDLES — it did not Dispose the controls themselves; the browser has always been
+                    // disposed explicitly, from Dispose() below. Since we no longer call it, dispose the
+                    // synchronization context — and with it the hidden marshaling control — ourselves,
+                    // so we don't leave a Control to be finalized on a dead thread. We are on the owning
+                    // thread, the only one allowed to do that, and WM_QUIT is only retrieved once the
+                    // queue is otherwise empty, so everything posted has already run.
                     ctx.Dispose();
                 }
             }
@@ -144,6 +147,11 @@ namespace Bloom.Publish
         /// windows and their async completions — while remaining invisible to WinForms' application-lifetime
         /// bookkeeping. So no non-GUI entry point (the harvester, bulk upload, any future CLI verb) can be
         /// told the application is exiting just because we finished with a browser.
+        ///
+        /// Leaving that bookkeeping cuts both ways, deliberately: WinForms no longer knows about this loop,
+        /// so Application.Exit() no longer ends it either. If the GUI shuts down while a browser is still
+        /// alive, this thread keeps pumping until the process goes away — harmless, because the thread is
+        /// IsBackground (it cannot hold the process open) and the CoreWebView2 process dies with its host.
         /// </remarks>
         private static void RunPrivateMessageLoop()
         {
--- a/src/BloomExe/Publish/OffScreenBrowser.cs
+++ b/src/BloomExe/Publish/OffScreenBrowser.cs
@@ -1,5 +1,7 @@
 using System;
+using System.ComponentModel;
 using System.Diagnostics;
+using System.Runtime.InteropServices;
 using System.Threading;
 using System.Threading.Tasks;
 using System.Windows.Forms;
@@ -27,8 +29,8 @@ namespace Bloom.Publish
     /// deliberately not static/shared between instances, so instances on different threads never contend).
     /// A caller that only needs one browser simply never calls StartFreshBrowser.
     ///
-    /// How it stays safe: the browser is owned by a private, dedicated STA thread with its own Windows Forms
-    /// message loop, and THAT thread — not the caller — services the browser's callbacks. So a caller can
+    /// How it stays safe: the browser is owned by a private, dedicated STA thread with its own message
+    /// loop, and THAT thread — not the caller — services the browser's callbacks. So a caller can
     /// simply block for a result. It never has to pump the MAIN UI message loop the way
     /// Browser.RunJavascriptWithStringResult_Sync_Dangerous does (Application.DoEvents), which lets unrelated
     /// user commands run in the middle of the call stack — the reentrancy blamed for BL-12614 / BL-13120. And
@@ -48,9 +50,6 @@ namespace Bloom.Publish
         // every inner browser this instance creates. Captured from the first browser and reused for each
         // fresh browser (see StartFreshBrowser), so we don't pay environment creation each time.
         private CoreWebView2Environment _environment;
-
-        // The message loop we run on the dedicated thread; ExitThread() on it ends the loop at Dispose.
-        private ApplicationContext _appContext;
 
         // The dedicated thread's Windows Forms synchronization context. Posting to it marshals work onto that
         // thread; awaits inside that work resume there too.
@@ -90,14 +89,15 @@ namespace Bloom.Publish
         /// </summary>
         public int BrowserThreadId => _thread.ManagedThreadId;
 
-        // Runs on the dedicated thread: establishes a message loop, creates the browser, then pumps messages
-        // until Dispose() ends the loop via the ApplicationContext.
+        // Runs on the dedicated thread: establishes a synchronization context, creates the browser, then
+        // pumps messages until Dispose() posts WM_QUIT.
         private void ThreadMain()
         {
             try
             {
-                // Give this thread a Windows Forms message loop + synchronization context, so the browser's
-                // async continuations (and cross-thread Posts from callers) run here.
+                // Give this thread a Windows Forms synchronization context, so the browser's async
+                // continuations (and cross-thread Posts from callers) run here. Posting through it works by
+                // sending window messages to a hidden marshaling control, which our pump below dispatches.
                 var ctx = new WindowsFormsSynchronizationContext();
                 SynchronizationContext.SetSynchronizationContext(ctx);
                 _ctx = ctx;
@@ -106,8 +106,18 @@ namespace Bloom.Publish
                 // CoreWebView2 initialization completes via that loop), then signal the constructor.
                 ctx.Post(_ => InitializeAsync(), null);
 
-                _appContext = new ApplicationContext();
-                Application.Run(_appContext); // pump until the context's loop is ended in Dispose
+                try
+                {
+                    RunPrivateMessageLoop();
+                }
+                finally
+                {
+                    // Application.Run disposed this thread's windows as its loop ended; since we no longer
+                    // use it, dispose the synchronization context (and so its hidden marshaling control)
+                    // ourselves. We are on the owning thread, the only one allowed to do that, and WM_QUIT
+                    // is only retrieved once the queue is otherwise empty, so all posted work has run.
+                    ctx.Dispose();
+                }
             }
             catch (Exception e)
             {
@@ -115,6 +125,75 @@ namespace Bloom.Publish
                 _ready.Set();
             }
         }
+
+        /// <summary>
+        /// Pumps this thread's messages until WM_QUIT, using the raw Win32 loop rather than
+        /// <see cref="Application.Run(ApplicationContext)"/>.
+        /// </summary>
+        /// <remarks>
+        /// Why not Application.Run: it registers this loop with WinForms as an *application* message loop.
+        /// Ending it therefore makes WinForms conclude that the application is exiting (when, as in a CLI
+        /// process, it is the process's only such loop) and raise Application.ApplicationExit. That is a lie
+        /// — a worker's private browser thread finishing is not the application quitting — and things
+        /// listening for a real shutdown act on it: ApplicationContainer disposed itself, killing the DI
+        /// scope the harvester's still-running createArtifacts depended on (BL-16668). We create and tear
+        /// down one of these browsers per batch of page checks, so that fired mid-run.
+        ///
+        /// A bare GetMessage/TranslateMessage/DispatchMessage loop dispatches everything this thread
+        /// actually needs — the WindowsFormsSynchronizationContext's marshaling control, the WebView2's
+        /// windows and their async completions — while remaining invisible to WinForms' application-lifetime
+        /// bookkeeping. So no non-GUI entry point (the harvester, bulk upload, any future CLI verb) can be
+        /// told the application is exiting just because we finished with a browser.
+        /// </remarks>
+        private static void RunPrivateMessageLoop()
+        {
+            // GetMessage returns 0 for WM_QUIT and -1 on error; anything else means we have a message.
+            int result;
+            while ((result = GetMessage(out var msg, IntPtr.Zero, 0, 0)) != 0)
+            {
+                if (result == -1)
+                    throw new Win32Exception(
+                        Marshal.GetLastWin32Error(),
+                        "GetMessage failed in the OffScreenBrowser message loop"
+                    );
+                TranslateMessage(ref msg);
+                DispatchMessage(ref msg);
+            }
+        }
+
+        #region Win32 message pump
+
+        [StructLayout(LayoutKind.Sequential)]
+        private struct NativeMessage
+        {
+            public IntPtr HWnd;
+            public uint Message;
+            public IntPtr WParam;
+            public IntPtr LParam;
+            public uint Time;
+            public int PointX;
+            public int PointY;
+        }
+
+        [DllImport("user32.dll", EntryPoint = "GetMessageW", SetLastError = true)]
+        private static extern int GetMessage(
+            out NativeMessage message,
+            IntPtr hWnd,
+            uint filterMin,
+            uint filterMax
+        );
+
+        [DllImport("user32.dll")]
+        private static extern bool TranslateMessage(ref NativeMessage message);
+
+        [DllImport("user32.dll", EntryPoint = "DispatchMessageW")]
+        private static extern IntPtr DispatchMessage(ref NativeMessage message);
+
+        // Queues WM_QUIT for the CALLING thread, which is why Dispose posts the call onto our thread.
+        [DllImport("user32.dll")]
+        private static extern void PostQuitMessage(int exitCode);
+
+        #endregion
 
         private async void InitializeAsync()
         {
@@ -308,7 +387,9 @@ namespace Bloom.Publish
                         }
                         finally
                         {
-                            _appContext?.ExitThread();
+                            // We are on the browser's own thread here, so this queues WM_QUIT for the loop
+                            // in RunPrivateMessageLoop.
+                            PostQuitMessage(0);
                         }
                     },
                     null
--- a/src/BloomExe/Publish/OffScreenBrowser.cs
+++ b/src/BloomExe/Publish/OffScreenBrowser.cs
@@ -119,6 +119,16 @@ namespace Bloom.Publish
                     // so we don't leave a Control to be finalized on a dead thread. We are on the owning
                     // thread, the only one allowed to do that, and WM_QUIT is only retrieved once the
                     // queue is otherwise empty, so everything posted has already run.
+                    //
+                    // What this does NOT clean up, deliberately: WinForms' per-thread ThreadContext.
+                    // Installing the synchronization context creates one, Application.Run's teardown
+                    // used to remove it, and disposing the context here does not. So each browser leaves
+                    // one behind, held by a static table. We decided (BL-16668) to accept that rather
+                    // than chase it: the table is keyed by native thread id, so entries are overwritten
+                    // as ids are reused instead of accumulating without bound, and the per-instance cost
+                    // is a small object plus a thread handle. The only way to really remove it would be
+                    // to keep one browser thread for the process lifetime, which would undo
+                    // ReleaseBrowser's deliberate "don't leave an idle WebView2 running" intent.
                     ctx.Dispose();
                 }
             }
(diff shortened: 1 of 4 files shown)


src/BloomExe/Publish/OffScreenBrowser.cs
```diff
@@ -116,6 +129,80 @@ private void ThreadMain()
...
+        /// alive, this thread keeps pumping until the process goes away — harmless, because the thread is
+        /// IsBackground (it cannot hold the process open) and the CoreWebView2 process dies with its host.
+        /// </remarks>
+        private static void RunPrivateMessageLoop()
+        {
+            // GetMessage returns 0 for WM_QUIT and -1 on error; anything else means we have a message.
+            int result;
+            while ((result = GetMessage(out var msg, IntPtr.Zero, 0, 0)) != 0)
+            {
+                if (result == -1)
```
andrew-polk, line 172: [Devin] **Investigate**: Application.MessageLoop is now false on the browser thread, changing the fatal-error exit path

*Mirrored from Devin's review by Claude Opus 5 (1M context) from Andrew Polk's machine during preflight.*

`Application.MessageLoop` is a per-thread flag meaning "a WinForms message loop is running on this thread". `Application.Run` used to set it on the off-screen browser's thread; a raw `GetMessage` pump does not. That flag is what Bloom's fatal-error handler uses to decide *how* to die: `FatalExceptionHandler.HandleTopLevelError` takes `ProgramExit.Exit()` (Bloom's normal, graceful shutdown) when it is true, and `Environment.Exit(1)` (an immediate kill, no shutdown work) when it is false.

So if an exception ever escapes a callback running on the browser's thread — the `_browser?.Dispose()` we post from `Dispose()`, or a WebView2 event handler — WinForms hands it to `Application.ThreadException`, and **in the desktop app** Bloom would now hard-exit where it previously shut down gracefully. For a user that is the difference between Bloom closing itself down and Bloom vanishing.

Verified while triaging: the harvester/CLI path this PR exists to fix is **not** affected. `FatalExceptionHandler.UseFallback` short-circuits before that branch, and it is only ever set back to `false` by `WorkspaceView`, which the CLI never constructs — so in `createArtifacts` the branch is unreachable either way.

This has been raised for a decision rather than changed here, because how Bloom exits on a fatal error is a user-visible behavior call, not a cleanup detail. Options considered: leave it (this is already the crash path, and the trigger is narrow); catch and log around the posted `_browser.Dispose()` so the most reachable case never becomes a top-level thread exception; or stop `FatalExceptionHandler` depending on `Application.MessageLoop` at all, which touches Bloom's global error path and is well beyond this fix.

src/BloomTests/CLI/CreateArtifactsCommandTests.cs
```diff
@@ -238,6 +238,99 @@ public void CreateArtifacts_LegacyBookWithInvalidXmatter_ReportsLegacyBookCannot
...
+        // bloomdigital spins up (and then tears down) PublishHelper's off-screen browser, whose
+        // dedicated thread runs the only WinForms message loop a CLI process has. Ending that loop
+        // made WinForms raise Application.ApplicationExit, which disposed the ApplicationContainer --
+        // the parent scope of our still-in-use ProjectContext -- so the epub step, which runs
+        // afterwards, died with ObjectDisposedException resolving ProjectContext.BookServer and
+        // createArtifacts returned EpubException. A test that requests a single artifact cannot
+        // catch that, because nothing runs after the premature disposal.
+        [Test]
+        public void CreateArtifacts_BloomDigitalAndEpubRequestedTogether_BothCreated()
+        {
```
andrew-polk, line 253: [Devin] **Investigate**: New regression test depends on a real WebView2 and full epub generation

*Mirrored from Devin's review by Claude Opus 5 (1M context) from Andrew Polk's machine during preflight.*

The new test drives the whole `createArtifacts` pipeline — a real off-screen WebView2 and a real epub build — which makes it much heavier than this file's other cases, most of which fail fast on an invalid xmatter. Devin asked whether it will be reliable on CI agents, where the WebView2 runtime and session state can differ, and noted that the epub suites share one browser via `PublishHelper.ExternalPageChecksBrowserForTests` while this test deliberately does not.

andrew-polk: *Claude Opus 5 (1M context) from Andrew Polk's machine during preflight.*

Assessed as not a new risk, so resolving.

The WebView2 dependency is not new to this fixture. `CreateArtifacts_LegacyBookWithInvalidXmatter_HarvesterAllowedToConvert_ConvertsToDefaultTheme`, already in this file, asks for a bloomdigital, which goes through `BloomPubMaker` → `PublishHelper` → a real off-screen WebView2. So the fixture has been booting a real browser on CI agents for some time; this test adds the epub step on top of a dependency that was already there and already passing. Neither this fixture nor the epub suites are marked `SkipOnTeamCity`.

On not sharing a browser: that is the point of the test rather than an oversight, and Devin agrees. What broke in BL-16668 was precisely the *teardown* of a per-batch browser — `PublishHelper.ReleaseBrowser()` ending the process's only WinForms message loop. Handing the test a long-lived shared browser via `ExternalPageChecksBrowserForTests` would skip that teardown and the test would pass on the unfixed code, which is exactly what we need it not to do. It does fail on unfixed `master`, with the production stack trace frame for frame.

On cost: it runs in about 2–4 seconds locally, and the full suite (3062 tests) is unchanged at roughly 4 minutes. If it ever does turn flaky on an agent the fix is to quarantine it by category, not to share the browser.

src/BloomExe/Publish/OffScreenBrowser.cs
```diff
@@ -106,8 +106,21 @@ private void ThreadMain()
                 // CoreWebView2 initialization completes via that loop), then signal the constructor.
                 ctx.Post(_ => InitializeAsync(), null);
 
-                _appContext = new ApplicationContext();
-                Application.Run(_appContext); // pump until the context's loop is ended in Dispose
+                try
+                {
+                    RunPrivateMessageLoop();
+                }
```
andrew-polk, line 112: [Devin] **Bug**: A crash on the off-screen browser's thread now kills Bloom abruptly instead of shutting it down cleanly

*Mirrored from Devin's review by Claude Opus 5 (1M context) from Andrew Polk's machine during preflight. This sharpens the [Investigate flag above](https://github.com/BloomBooks/BloomDesktop/pull/8177#discussion_r3737967331) — same mechanism, but Devin traced it to a consequence a user would actually feel, so it is worth its own thread.*

The mechanism: `Application.MessageLoop` is per-thread and true only while a WinForms-registered loop (`Application.Run`) is active on *that* thread. `FatalExceptionHandler.HandleTopLevelError` uses it to choose how Bloom dies — `ProgramExit.Exit()` when true, a bare `Environment.Exit(1)` when false. The private pump makes it false, so an exception escaping a callback on this thread (a WebView2 event handler, or the `_browser.Dispose()` we post) now takes the hard-kill branch.

**Why that is worse than just "less graceful", which is what I had assumed when I raised the flag above.** `Environment.Exit` does not run `finally` blocks, and Bloom releases its single-instance token in `Main`'s `finally` (`Program.cs:751`). So this path exits without releasing that token — the very "zombie stops another instance starting" condition `ProgramExit` exists to prevent, and whose own comment says the listener-close and token-release "MUST be done so that Bloom can be started again without problems". A user whose Bloom crashed this way could find it will not start again.

Verified while triaging:

- The harvester/CLI path this PR exists to fix is **not** affected. `FatalExceptionHandler.UseFallback` short-circuits before that branch, and it is only ever set back to `false` by `WorkspaceView`, which the CLI never constructs.
- One correction to the report above: the "MUST be done" cleanup it cites at `ProgramExit.cs:52-68` is in the 20-second *watchdog* thread, not in `Exit()`'s main path. `Exit()` starts that watchdog and calls `Application.Exit()`; normal shutdown then releases the token via `Main`'s `finally`. The conclusion is unchanged, and arguably reinforced — the old path had both a graceful route *and* a watchdog backstop, and `Environment.Exit(1)` has neither.
- How much a stale token really costs depends on the token implementation: an OS mutex is released when the process dies regardless, a lock file is not. Worth a look before deciding.

Not fixed here. The clean fix is Devin's suggestion — decide on whether the *application's main* loop is running (a flag set around `Program`'s `Application.Run`) rather than on whether the current thread has one, since "is there a GUI to shut down?" is the question that branch is really asking. That changes Bloom's global fatal-error path for every thread, which is beyond a targeted bug fix, so it has gone to the developer as a decision.

src/BloomExe/Publish/OffScreenBrowser.cs
```diff
@@ -106,8 +106,21 @@ private void ThreadMain()
...
+                {
+                    // As its loop ended, Application.Run's teardown destroyed this thread's window
+                    // HANDLES — it did not Dispose the controls themselves; the browser has always been
+                    // disposed explicitly, from Dispose() below. Since we no longer call it, dispose the
+                    // synchronization context — and with it the hidden marshaling control — ourselves,
+                    // so we don't leave a Control to be finalized on a dead thread. We are on the owning
+                    // thread, the only one allowed to do that, and WM_QUIT is only retrieved once the
+                    // queue is otherwise empty, so everything posted has already run.
+                    ctx.Dispose();
+                }
```
andrew-polk, line 133: [Devin] **Investigate**: WinForms ThreadContext for this thread is now never torn down

*Mirrored from Devin's review by Claude Opus 5 (1M context) from Andrew Polk's machine during preflight.*

`Application.Run`'s teardown used to run WinForms' per-thread `ThreadContext` exit path, which removes that context from WinForms' static table and releases the duplicated thread handle it holds. The private pump never ends it. The context still gets *created* — implicitly, by `new WindowsFormsSynchronizationContext()` — and disposing the synchronization context only disposes the shared marshaling control, not the context. Because the static table holds a strong reference, no finalizer reclaims it.

So each off-screen browser (one per batch of page checks) leaves a `ThreadContext` object and its OS thread handle behind. Devin notes the table is keyed by native thread id, so entries get overwritten as ids are reused and it should not grow without bound, and the per-instance cost is small — but it is a real change from the previous teardown and worth confirming against the WinForms version in use.

andrew-polk:
*Claude Opus 5 (1M context) from Andrew Polk's machine during preflight.*

Consulted Devin on 2026-08-07, twice: first up to commit `41b404ba6cc02bc8ac0d61ee5072c6e7f9623118`, then again up to `aa4c570295b00792ab8d7ca50e9aafdeb4d40f48` after the comment fixes.

The second pass raised **1 Bug and 3 Investigate flags**, each now its own thread:

- **Bug — a crash on the browser thread now hard-kills Bloom** ([thread](https://github.com/BloomBooks/BloomDesktop/pull/8177#discussion_r3738008860)). Real, and worse than the first pass suggested: the hard-kill skips the `finally` that releases Bloom's single-instance token, so Bloom might not restart. Left **open** — the clean fix changes Bloom's global fatal-error path, so it is the developer's call.
- **Investigate — `Application.MessageLoop` is now false on that thread** ([thread](https://github.com/BloomBooks/BloomDesktop/pull/8177#discussion_r3737967331)). The same mechanism as the Bug, raised in the first pass; left **open** with it.
- **Investigate — the WinForms ThreadContext for that thread is never torn down** ([thread](https://github.com/BloomBooks/BloomDesktop/pull/8177#discussion_r3738008972)). A small per-instance leak. Left **open** for a decision.
- **Investigate — the new regression test needs a real WebView2** ([thread](https://github.com/BloomBooks/BloomDesktop/pull/8177#discussion_r3737967444)). Assessed as not a new risk and **resolved**: this fixture already boots a real browser for its bloomdigital case, and not sharing a browser is the point — a shared one would skip the teardown that BL-16668 broke, and the test would pass on the unfixed code.

Also 3 Informational items, not mirrored. Two of them — a comment that overstated what `Application.Run` cleaned up, and nothing tearing the thread down on GUI exit — were **already fixed** in `aa4c570`, in response to the first pass; the re-review re-emitted the comment one against text that no longer exists, so treat that one as stale. The third (a `GetMessage` failure after startup would be swallowed) is unreachable today and went to the developer alongside the others.

CI: the only GitHub check is `pr-automation`, which passed — note it only triggers this Devin review and adds the review link to the description; nothing server-side builds or tests this branch. CodeRabbit is switched off for this repo in `.coderabbit.yml`. The full C# suite was run locally at this commit: 3049 passed, 12 skipped, 1 failed, and that one failure is an opt-in Reading App Builder test failing its own environment precondition on this machine.

src/BloomExe/Publish/OffScreenBrowser.cs
andrew-polk: *Claude Opus 5 (1M context) from Andrew Polk's machine during preflight.*

**Fixed in `645bbdf2795e831477f9b741c234542f93c0970b`.** Andrew chose the principled option rather than the narrow one.

`FatalExceptionHandler` no longer asks `Application.MessageLoop`. It asks the question it actually means — "is there a running GUI application to shut down?" — via a new `Program.MainMessageLoopIsRunning`, set only while the `Application.Run()` in `RunBloom` is pumping.

Worth noting what that reframing revealed: this was **not** really a regression introduced by the private pump. `Application.MessageLoop` is per-thread, so a fatal error surfacing on *any* worker thread already took the `Environment.Exit(1)` path, and already skipped the `finally` in `Main` that releases the single-instance token. The off-screen browser's thread happened to be immune only because it called `Application.Run`. Moving it to a private pump removed that accidental immunity and made a pre-existing flaw visible. The fix is correct for every thread, not just this one.

Thanks for the catch — the impact analysis (tracing it through to the token, not stopping at "less graceful shutdown") is what turned this from something I was ready to leave alone into something worth fixing.

src/BloomExe/Publish/OffScreenBrowser.cs
andrew-polk: *Claude Opus 5 (1M context) from Andrew Polk's machine during preflight.*

**Fixed in `645bbdf2795e831477f9b741c234542f93c0970b`** — resolving, as this is the same mechanism as the Bug thread below, where the reasoning is recorded in full.

Short version: rather than restore `Application.MessageLoop` on this thread, we removed the dependency on it. `FatalExceptionHandler` now consults `Program.MainMessageLoopIsRunning` — true only while Bloom's main `Application.Run()` is pumping — so the crash-exit path no longer depends on which thread the exception arrived on.

src/BloomExe/Publish/OffScreenBrowser.cs
andrew-polk: *Claude Opus 5 (1M context) from Andrew Polk's machine during preflight.*

**Decided: leave as is** — Andrew's call, and the reasoning is now recorded as a code comment next to `ctx.Dispose()` in `645bbdf2795e831477f9b741c234542f93c0970b` so the next person to read this method finds it.

Accepting it on your own analysis: the static table is keyed by native thread id, so entries get overwritten as ids are reused rather than accumulating without bound, and the per-instance cost is one small object plus a thread handle. That makes it a bounded, self-limiting cost rather than a leak that ends in trouble.

The only change that would genuinely remove it is keeping a single browser thread alive for the process lifetime — and that would undo `ReleaseBrowser`'s deliberate intent not to leave an idle WebView2 (and its browser process) running between publishes. Trading a real, continuous resource cost for a bounded, self-limiting one is not a good deal, so we are not taking it.

src/BloomExe/Publish/OffScreenBrowser.cs
andrew-polk: *Claude Opus 5 (1M context) from Andrew Polk's machine during preflight.*

Devin's third pass (at `645bbdf`) marks this **resolved**: *"FatalExceptionHandler now bases the graceful-exit decision on Program.MainMessageLoopIsRunning rather than the current thread's Application.MessageLoop, so a browser-thread exception while the GUI loop runs takes ProgramExit.Exit()."* Confirmed fixed from both directions, then.

**One correction to my reply above, which overstated things.** I wrote that the pump change "exposed a pre-existing flaw rather than creating it — any other worker thread was always at risk". That is too strong, and Devin's follow-up review rightly pushed back on the framing.

More precisely: asking a *per-thread* question (`Application.MessageLoop`) to decide a *process-wide* one was always the wrong question — that part is genuinely pre-existing. But reaching the bad outcome needs a thread that pumps WinForms messages *without* `Application.Run`, and before this PR nothing in Bloom did that. So the reachable failure really was introduced here; it was not lying in wait for arbitrary worker threads. The fix is still the right one, and still correct for any future thread in that position — but it is a fix for something this PR introduced, not a pre-existing bug this PR merely uncovered. The commit message for `645bbdf` carries the same overstatement; this is the correction.

(Devin also filed an informational note that the practical effect of the swap is narrow. Agreed that it is GUI-only — the branch is unreachable while `UseFallback` is true, which is every CLI path. Its reasoning that the case "no longer exists now that OffScreenBrowser doesn't use Application.Run" is the one part I would push back on: a worker thread with *no* WinForms loop is exactly where the per-thread check misreads, so removing `Application.Run` from that thread creates the case rather than removing it. Devin's own resolution above says the same thing.)

andrew-polk:
*Claude Opus 5 (1M context) from Andrew Polk's machine during preflight.*

Consulted Devin a third time on 2026-08-07, up to commit `645bbdf2795e831477f9b741c234542f93c0970b`.

**Re-review clean — no new findings, and the bug is confirmed fixed.** Devin marks its earlier bug *resolved*, citing the switch to `Program.MainMessageLoopIsRunning`. Every Investigate flag in this pass is one we had already mirrored and closed, so nothing new was posted.

Thread status — all four now resolved with a documented outcome:

- **Bug, a crash on the browser thread hard-kills Bloom** — fixed in `645bbdf`, and Devin agrees ([thread](https://github.com/BloomBooks/BloomDesktop/pull/8177#discussion_r3738008860)). That thread also carries a correction to a claim I had made about the bug being pre-existing; it was not, and the record now says so.
- **`Application.MessageLoop` false on the browser thread** — same fix ([thread](https://github.com/BloomBooks/BloomDesktop/pull/8177#discussion_r3737967331)).
- **Retained WinForms ThreadContext** — Andrew decided to accept it; reasoning recorded both on the [thread](https://github.com/BloomBooks/BloomDesktop/pull/8177#discussion_r3738008972) and as a code comment beside `ctx.Dispose()`.
- **Regression test needs a real WebView2** — assessed as not a new risk and [resolved](https://github.com/BloomBooks/BloomDesktop/pull/8177#discussion_r3737967444) earlier.

Five Informational items this pass, none mirrored. One is new — a note that the practical effect of the `Application.MessageLoop` swap is narrow, which is fair in that the branch is GUI-only (unreachable while `UseFallback` is true, i.e. on every CLI path); the part of its reasoning I disagree with is answered on the bug thread. Two describe things already fixed in earlier commits of this branch, and one of those quotes text that no longer exists, so treat it as stale. The remaining two — a post-dispose call hanging rather than throwing, and a `GetMessage` failure after startup being swallowed — went to Andrew as a decision and he chose to leave them: neither is reachable from any current caller.

Full C# suite at this commit, run locally: 3049 passed, 12 skipped, 1 failed — the one failure is an opt-in Reading App Builder test failing its own environment precondition on this machine, unrelated to this branch. Note that nothing server-side builds or tests this branch (the `pr-automation` check only triggers this Devin review), and CodeRabbit is switched off for this repo, so that local run is the only test evidence. The fatal-error path changed in `645bbdf` has no automated coverage — there is no harness for a path that deliberately terminates the process — so its evidence is the compile plus the absence of regression elsewhere.

PR #8178, opened 2026-08-07
Don't listen for ApplicationExit when there is no GUI application (BL-16668)

andrew-polk:
`Bloom.exe createArtifacts` was dying with `ObjectDisposedException` on the epub step, so the
harvester failed **every** book with epubs enabled and uploaded no artifacts at all — not even
the .bloompub it had already built successfully. Reported downstream as BH-7836.

### What was happening

A command-line verb never calls `Application.Run`, so the only WinForms message loop in the process
belongs to a worker: the dedicated thread of the off-screen browser `PublishHelper` uses for its
page checks, created and disposed once per batch. So:

1. `createArtifacts` builds the .bloompub first; `BloomPubMaker` does `using (var helper = new PublishHelper())`.
2. Disposing it calls `PublishHelper.ReleaseBrowser()` → `OffScreenBrowser.Dispose()`, ending that loop.
3. It was the process's only message loop, so WinForms raises `Application.ApplicationExit`.
4. `ApplicationContainer.OnApplicationExit` disposes the container — the **parent** of `ProjectContext._scope`.
5. The epub step, which runs next, evaluates `s_projectContext.BookServer` → `ObjectDisposedException`.

In the GUI this never fires, because the main thread's loop is still running. It only ever bit a
non-GUI entry point.

`d41fa15f9e` had already met this mechanism and made `ProjectContext.Dispose` tolerate the container
being disposed first — but the premature disposal itself remained. It was harmless there only
because it happened after all the requested work.

### The fix

Rather than act on a signal we know to be wrong, don't subscribe to it. `Program.RunningInConsoleMode`
is set by `Program.Main` for **every** command-line verb (`hydrate`, `upload`, `download`, `getfonts`,
`changeLayout`, `createArtifacts`, `spreadsheetExport`/`Import`, `sendFontAnalytics`) before it
dispatches to any of them, so gating the subscription on it means the container never hears a claim
of application exit that no GUI is behind.

Two notes on why it is this flag and this location:

- **`RunningInConsoleMode`, not `RunningHarvesterMode`.** An earlier draft used the harvester flag.
  This one covers every verb — the others never set the harvester flag and would have stayed
  exposed — and it has no ordering hazard: it is true before any container exists, whereas
  `CreateArtifactsCommand` sets `RunningHarvesterMode` *after* constructing the container, so a
  guard on that flag works only because the event happens to fire later.
- **Gating the subscription, not the handler.** The question is settled before we would subscribe,
  so there is no reason to sign up and then ignore the answer.

Verified that the console-mode branch of `Main` is self-contained — it parses, dispatches, pumps
with `Application.DoEvents()` and returns, never reaching `RunBloom`'s `Application.Run` — so
`RunningInConsoleMode` being true really does mean there is no GUI application to listen for, and no
GUI path is affected.

### Regression test

Every pre-existing `CreateArtifactsCommandTests` case requested a single artifact, which is exactly
why CI stayed green through this. The harvester always passes `--bloomdOutputPath`/
`--bloomDigitalOutputPath` **and** `--epubOutputPath` in one run, and only that combination exposes
the bug — with one artifact, nothing runs after the premature disposal. The new test asks for both
and fails on unmodified `master` with the production stack trace, frame for frame:

```
Expected: Success   But was: EpubException
System.ObjectDisposedException: Instances cannot be resolved and nested lifetimes cannot be created
from this LifetimeScope as it (or one of its parent scopes) has already been disposed.
   at Bloom.ProjectContext.get_BookServer() ... ProjectContext.cs:840
   at Bloom.CLI.CreateArtifactsCommand.CreateEpubArtifact ... CreateArtifactsCommand.cs:401
```

The tests call `HandleInternal` directly and so bypass `Main`, which sets that flag for real runs, so
they have to stand in for it — hence the new `SetUp`, mirroring what the fixture already did for
`RunningHarvesterMode` in `TearDown`. The repro was re-verified with `ApplicationContainer.cs` at
exact `master` state and that `SetUp` in place, confirming it is the fix passing the test rather than
the harness.

### Relationship to #8177

#8177 is the more thorough alternative: it stops `OffScreenBrowser` using `Application.Run` at all,
so the false `ApplicationExit` is never raised and no listener has to know to distrust it. That is
the better end state, but it is a bigger change (a private Win32 message pump) and it turned out to
have a consequence of its own — with no WinForms loop on that thread, `Application.MessageLoop` goes
false, which sent Bloom's fatal-error handler down a hard-kill path that skips releasing the
single-instance token. Fixable, and fixed there, but it widened the change.

This PR is the deliberate small-scope choice for now. **The two should not both land**: with the
pump change in place this guard is dead code.

Ref: https://issues.bloomlibrary.org/youtrack/issue/BL-16668

[Devin review](https://devinreview.com/BloomBooks/BloomDesktop/pull/8178)

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8178)

### DIFF
--- a/src/BloomExe/ApplicationContainer.cs
+++ b/src/BloomExe/ApplicationContainer.cs
@@ -81,7 +81,25 @@ namespace Bloom
 
             _container = builder.Build();
 
-            Application.ApplicationExit += OnApplicationExit;
+            // Only listen for the application exiting when there IS an application in the GUI sense.
+            // A command-line verb never calls Application.Run, so the only WinForms message loop in the
+            // process belongs to some worker -- currently the dedicated thread of the off-screen browser
+            // PublishHelper uses for page checks, created and disposed once per batch. When that loop
+            // ends, WinForms decides the application is exiting and raises ApplicationExit, mid-run. Acting
+            // on that disposed this container, the parent scope of the still-in-use ProjectContext, so the
+            // next artifact step died with ObjectDisposedException (BL-16668). Not subscribing is safe
+            // because in that flow the container's lifetime is already bounded by the `using` blocks in the
+            // CLI command handlers, and the process exits as soon as those unwind.
+            //
+            // One knock-on worth knowing: OnApplicationExit is the only caller of
+            // Program.FinishLocalizationHarvesting(), so a command-line verb no longer runs it. That is
+            // #if DEBUG code which does nothing unless LocalizationManager.IgnoreExistingEnglishTranslationFiles
+            // is set, so release CLI runs are unaffected -- but a DEBUG localization-harvesting run driven
+            // through a CLI verb would no longer merge the English translation files. If we ever want that,
+            // call it from the CLI path explicitly rather than by leaning on a shutdown event that is not
+            // really telling us the application is shutting down.
+            if (!Program.RunningInConsoleMode)
+                Application.ApplicationExit += OnApplicationExit;
 
             // Register the API Handlers that are global to the application (not dependent on knowing a particular project).
             // Note: it is is a work in progress to transfer more API handlers from ProjectContext to here.
@@ -98,6 +116,10 @@ namespace Bloom
             server.ApiHandler.RecordApplicationLevelHandlers();
         }
 
+        /// <summary>
+        /// The application is really shutting down, so tear the container down. Only ever subscribed
+        /// when Bloom is running as a GUI application -- see the constructor for why.
+        /// </summary>
         private void OnApplicationExit(object sender, EventArgs e)
         {
             Application.ApplicationExit -= OnApplicationExit;
--- a/src/BloomTests/CLI/CreateArtifactsCommandTests.cs
+++ b/src/BloomTests/CLI/CreateArtifactsCommandTests.cs
@@ -11,11 +11,23 @@ namespace BloomTests.CLI
     [TestFixture]
     public class CreateArtifactsCommandTests
     {
+        [SetUp]
+        public void SetUp()
+        {
+            // Program.Main sets this for a real command-line run, before it dispatches to the verb. These
+            // tests call HandleInternal directly and so bypass Main, which means they have to stand in for
+            // it — exactly as TearDown below already does for RunningHarvesterMode. ApplicationContainer
+            // reads this flag to know there is no GUI application whose exit it should listen for; without
+            // it set, the container tears itself down mid-run and the epub step fails (BL-16668).
+            Program.RunningInConsoleMode = true;
+        }
+
         [TearDown]
         public void TearDown()
         {
             // Without this, subsequent tests will fail because they think the harvester is still running.
             Program.RunningHarvesterMode = false;
+            Program.RunningInConsoleMode = false;
         }
 
         [Test]
@@ -235,6 +247,99 @@ namespace BloomTests.CLI
                 );
 
                 Assert.That(result, Is.EqualTo(CreateArtifactsExitCode.LegacyBookCannotHarvest));
+            }
+        }
+
+        // The harvester never asks for just one artifact: it passes --bloomdOutputPath,
+        // --bloomDigitalOutputPath AND --epubOutputPath in a single createArtifacts run. That
+        // combination is what BL-16668 broke, and why every other test here missed it. Making the
+        // bloomdigital spins up (and then tears down) PublishHelper's off-screen browser, whose
+        // dedicated thread runs the only WinForms message loop a CLI process has. Ending that loop
+        // made WinForms raise Application.ApplicationExit, which disposed the ApplicationContainer --
+        // the parent scope of our still-in-use ProjectContext -- so the epub step, which runs
+        // afterwards, died with ObjectDisposedException resolving ProjectContext.BookServer and
+        // createArtifacts returned EpubException. A test that requests a single artifact cannot
+        // catch that, because nothing runs after the premature disposal.
+        [Test]
+        public void CreateArtifacts_BloomDigitalAndEpubRequestedTogether_BothCreated()
+        {
+            using (
+                var testFolder = new TemporaryFolder(
+                    "CreateArtifacts_BloomDigitalAndEpubRequestedTogether_BothCreated"
+                )
+            )
+            {
+                var collectionFolderPath = testFolder.Combine("collection");
+
+                var bookFolderPath = Path.Combine(collectionFolderPath, "book");
+                System.IO.Directory.CreateDirectory(bookFolderPath);
+                var collectionFilePath = Path.Combine(
+                    collectionFolderPath,
+                    "collection.bloomCollection"
+                );
+                var settings = new CollectionSettings(collectionFilePath);
+                settings.Save();
+                var metaData = new BookMetaData();
+                metaData.WriteToFolder(bookFolderPath);
+                var bookPath = System.IO.Path.Combine(bookFolderPath, "book.htm");
+                System.IO.File.WriteAllText(
+                    bookPath,
+                    @"<html>
+                    <body>
+						<div class='bloom-page'>
+							<div class='marginBox'>
+								<div class='bloom-translationGroup normal-style'>
+									<div class='bloom-editable normal-style bloom-content1 bloom-contentNational1 bloom-visibility-code-on' lang='en'>
+                                        Hello
+									</div>
+								</div>
+							</div>
+						</div>
+					</body>
+				</html>"
+                );
+
+                var bloomDigitalOutputPath = Path.Combine(testFolder.FolderPath, "bloomdigital");
+                var epubOutputPath = Path.Combine(testFolder.FolderPath, "epub", "book.epub");
+                // Sanity check: neither artifact exists yet, so finding them later really does mean
+                // this run made them.
+                Assert.That(
+                    Directory.Exists(bloomDigitalOutputPath),
+                    Is.False,
+                    "test setup: bloomdigital output should not exist before the run"
+                );
+                Assert.That(
+                    File.Exists(epubOutputPath),
+                    Is.False,
+                    "test setup: epub output should not exist before the run"
+                );
+
+                var result = CreateArtifactsCommand.HandleInternal(
+                    new CreateArtifactsParameters()
+                    {
+                        BookPath = bookFolderPath,
+                        CollectionPath = collectionFilePath,
+                        BloomDigitalOutputPath = bloomDigitalOutputPath,
+                        EpubOutputPath = epubOutputPath,
+                        NoAnalytics = true,
+                    }
+                );
+
+                Assert.That(
+                    result,
+                    Is.EqualTo(CreateArtifactsExitCode.Success),
+                    "createArtifacts should succeed when both a bloomdigital and an epub are requested"
+                );
+                Assert.That(
+                    File.Exists(Path.Combine(bloomDigitalOutputPath, "index.htm")),
+                    Is.True,
+                    "the bloomdigital artifact should have been created"
+                );
+                Assert.That(
+                    File.Exists(epubOutputPath),
+                    Is.True,
+                    "the epub artifact should have been created"
+                );
             }
         }
 
--- a/src/BloomExe/CLI/CreateArtifactsCommand.cs
+++ b/src/BloomExe/CLI/CreateArtifactsCommand.cs
@@ -195,7 +195,15 @@ namespace Bloom.CLI
                     catch (Exception ex)
                     {
                         Console.WriteLine(ex.ToString());
-                        exitCode = CreateArtifactsExitCode.EpubException;
+                        // |=, not =: these are [Flags] values and every other step here accumulates.
+                        // A plain assignment discarded whatever had already been recorded -- most
+                        // notably BookHtmlNotFound from the bloomdigital step just above -- so the
+                        // harvester was told only "EpubException" and lost the fact that the book's
+                        // HTML was also wrong. That is the combination the harvester actually hits,
+                        // since it always asks for both artifacts. Safe to accumulate from this
+                        // worker: the main thread is blocked on countdownEvent below and does not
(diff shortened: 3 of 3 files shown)


src/BloomExe/ApplicationContainer.cs
```diff
@@ -81,7 +81,25 @@ public ApplicationContainer()
...
+            //
+            // One knock-on worth knowing: OnApplicationExit is the only caller of
+            // Program.FinishLocalizationHarvesting(), so a command-line verb no longer runs it. That is
+            // #if DEBUG code which does nothing unless LocalizationManager.IgnoreExistingEnglishTranslationFiles
+            // is set, so release CLI runs are unaffected -- but a DEBUG localization-harvesting run driven
+            // through a CLI verb would no longer merge the English translation files. If we ever want that,
+            // call it from the CLI path explicitly rather than by leaning on a shutdown event that is not
+            // really telling us the application is shutting down.
+            if (!Program.RunningInConsoleMode)
+                Application.ApplicationExit += OnApplicationExit;
```
andrew-polk, line 102: [Devin] **Investigate**: Unit-test runs still subscribe to the bogus ApplicationExit

*Mirrored from Devin's review by Claude Opus 5 (1M context) from Andrew Polk's machine during preflight.*

The guard exempts only `Program.RunningInConsoleMode`. A unit test that constructs an `ApplicationContainer` indirectly — via `DownloadBookCommand`, `UploadCommand`, `ChangeLayoutCommand` or `SendFontAnalyticsCommand`, all of which do `new ApplicationContainer()` — would still subscribe, so the same false exit signal could dispose a container mid-test. Devin suggests also exempting `Program.RunningUnitTests`, since neither mode has a GUI application whose exit is meaningful.

**Assessed and deliberately not changed.** Checked first whether it is reachable: no test in the repo invokes any of those four commands, and no test constructs an `ApplicationContainer` directly (`grep -rn "new ApplicationContainer" src/BloomTests/` is empty). The only test path that builds one is `CreateArtifactsCommand.HandleInternal`, which is this fixture, and it is covered.

The reason for not widening it anyway is test fidelity. `RunningInConsoleMode` is the predicate production actually depends on. If the guard also honoured `RunningUnitTests`, this fixture would pass through the test-only exemption, and a future mistake in the production wiring — the flag set too late, or on the wrong path — would no longer fail anything. That is a real cost, paid to protect tests that do not exist.

Concretely: today's arrangement means the regression test fails if `RunningInConsoleMode` is not correctly true at container-construction time, which is exactly the property worth guarding. Adding `|| Program.RunningUnitTests` is a one-line change and a reasonable call if someone later adds a test for one of those verbs — better then, alongside a test that needs it, than now on speculation. Flagged to Andrew as an FYI so it is a recorded choice rather than an oversight.

src/BloomTests/CLI/CreateArtifactsCommandTests.cs
```diff
@@ -238,6 +250,99 @@ public void CreateArtifacts_LegacyBookWithInvalidXmatter_ReportsLegacyBookCannot
...
+        // bloomdigital spins up (and then tears down) PublishHelper's off-screen browser, whose
+        // dedicated thread runs the only WinForms message loop a CLI process has. Ending that loop
+        // made WinForms raise Application.ApplicationExit, which disposed the ApplicationContainer --
+        // the parent scope of our still-in-use ProjectContext -- so the epub step, which runs
+        // afterwards, died with ObjectDisposedException resolving ProjectContext.BookServer and
+        // createArtifacts returned EpubException. A test that requests a single artifact cannot
+        // catch that, because nothing runs after the premature disposal.
+        [Test]
+        public void CreateArtifacts_BloomDigitalAndEpubRequestedTogether_BothCreated()
+        {
```
andrew-polk, line 265: [Devin] **Investigate**: New regression test does a real epub + bloomdigital build

*Mirrored from Devin's review by Claude Opus 5 (1M context) from Andrew Polk's machine during preflight.*

The test drives the full publish path — the off-screen WebView2 plus epub creation — and is not categorised `SkipOnTeamCity`, making it heavier and more environment-dependent than the rest of this file. Devin asks for confirmation that CI agents have WebView2 and that the runtime is acceptable.

**Assessed as not a new risk, so resolving.**

The WebView2 dependency is not new to this fixture. `CreateArtifacts_LegacyBookWithInvalidXmatter_HarvesterAllowedToConvert_ConvertsToDefaultTheme` already asks for a bloomdigital, which runs `BloomPubMaker` → `PublishHelper` → a real off-screen browser. Devin's own description grants this ("existing tests in the fixture only produce a bloomdigital"), and producing a bloomdigital *is* the browser path. Neither this fixture nor the epub suites carry `SkipOnTeamCity`, so agents have been running a real WebView2 here for some time; this test adds the epub step on top of a dependency that was already present and passing.

On runtime: the full suite at this commit is **3052 passed, 12 skipped, 0 failed in 5m16s**, and this test contributes 2–4 seconds of that. Not a meaningful change in suite cost.

Not sharing a browser is deliberate. What broke in BL-16668 was the *teardown* of a per-batch browser, so handing the test a long-lived shared one via `ExternalPageChecksBrowserForTests` would skip the very thing under test and the test would pass on unfixed code. It has been verified to fail with `ApplicationContainer.cs` at exact `master` state. If it ever does turn flaky on an agent, the right response is to quarantine it by category, not to share the browser.

src/BloomTests/CLI/CreateArtifactsCommandTests.cs
```diff
@@ -238,6 +250,99 @@ public void CreateArtifacts_LegacyBookWithInvalidXmatter_ReportsLegacyBookCannot
...
+                        BloomDigitalOutputPath = bloomDigitalOutputPath,
+                        EpubOutputPath = epubOutputPath,
+                        NoAnalytics = true,
+                    }
+                );
+
+                Assert.That(
+                    result,
+                    Is.EqualTo(CreateArtifactsExitCode.Success),
+                    "createArtifacts should succeed when both a bloomdigital and an epub are requested"
```
andrew-polk, line 331: [Devin] **Investigate**: New test depends on the epub worker's exit code being observed across threads

*Mirrored from Devin's review by Claude Opus 5 (1M context) from Andrew Polk's machine during preflight. Devin filed this as informational; raising it as a thread because verifying it turned up a real adjacent bug that deserves a recorded decision rather than a footnote.*

Devin's point: the epub is produced on a thread-pool thread, which on failure assigns `exitCode`, and the main thread reads it after a `CountdownEvent`/`IsSet` spin. This new test is the first whose pass/fail depends on that cross-thread value, so a flake here would point at that code rather than at the fix. Fair, and worth knowing. In practice the visibility is almost certainly fine — `CountdownEvent.Signal`/`IsSet` carry barriers, and the wait loop also does `Thread.Sleep` and `Application.DoEvents` — but nothing in the code states the intent.

**The part worth acting on is the other half of that sentence.** At `CreateArtifactsCommand.cs:198` the worker does:

```csharp
exitCode = CreateArtifactsExitCode.EpubException;   // plain assignment
```

Every other step in `CreateArtifacts` accumulates with `|=`. These are `[Flags]` values, so this one **overwrites** anything already recorded. If the bloomdigital step set `BookHtmlNotFound` (2), or an earlier step set `UnhandledException` (1) or `LegacyBookCannotHarvest` (8), and the epub then throws, the process exits with just `4`. `GetErrorsFromExitCode` then reports only `EpubException`, and the harvester's log loses the fact that the book's HTML was also wrong — precisely the combination this card is about, since the harvester always asks for both artifacts.

It looks like a one-character fix (`|=`), and it is pre-existing rather than introduced here. Not changed in this PR for two reasons: it alters the exit codes the harvester consumes and records, which is a behavior contract rather than a cleanup detail; and this PR was deliberately chosen for being narrow. Raised with Andrew for a decision — most likely its own small card.

Nothing here affects whether this PR's fix is correct: the test asserts `Success`, and on the fixed code no step sets any flag at all.

src/BloomExe/ApplicationContainer.cs
andrew-polk: *Claude Opus 5 (1M context) from Andrew Polk's machine during preflight.*

Devin restated this at the final HEAD (`878735fc`), worded as "unit tests outside this fixture remain exposed to the same premature-disposal mechanism". Same finding, so it stays here rather than getting a second thread. Still unreachable — no test invokes those four commands and none constructs an `ApplicationContainer` directly — and still with Andrew, since the reason not to widen the guard is a judgement about test fidelity rather than a technical blocker.

andrew-polk:
*Claude Opus 5 (1M context) from Andrew Polk's machine during preflight.*

Consulted Devin on 2026-08-07, up to commit `878735fcfb1331f987d8d6440925ea7a73fb2f9a`. **No bugs.** Six flags, two of them Investigate.

- **Unit tests outside this fixture could still hit the same mechanism** ([thread](https://github.com/BloomBooks/BloomDesktop/pull/8178#discussion_r3738320252)) — raised in both passes. Verified unreachable today: no test invokes `DownloadBookCommand`, `UploadCommand`, `ChangeLayoutCommand` or `SendFontAnalyticsCommand`, and none builds an `ApplicationContainer` directly. Left **open** for Andrew, because also exempting `RunningUnitTests` would let the regression test pass through a test-only exemption and stop proving the production predicate is wired right — a trade worth his call rather than mine.
- **The regression test does a real epub + bloomdigital build** ([thread](https://github.com/BloomBooks/BloomDesktop/pull/8178#discussion_r3738320367)) — assessed as not a new risk and **resolved**: this fixture already booted a real off-screen browser for its bloomdigital case, and not sharing a browser is the point, since a shared one would skip the teardown that BL-16668 broke.
- **The epub worker's exit code is read across threads** ([thread](https://github.com/BloomBooks/BloomDesktop/pull/8178#discussion_r3738419833)) — filed by Devin as informational, promoted to a thread because checking it turned up a real pre-existing bug next door: `CreateArtifactsCommand.cs:198` assigns `exitCode` where every other step uses `|=`, so an epub failure **erases** flags such as `BookHtmlNotFound`. Left **open** for Andrew; it changes exit codes the harvester consumes, so it is not a cleanup to slip into a narrow fix.

Three further informational items, not mirrored. One — `FinishLocalizationHarvesting` no longer running for console verbs — was already addressed in `878735f`, which documents it next to the guard. One notes that console-mode disposal now rests entirely on the `using` blocks in each verb; Devin checked all five verbs and confirmed they all have one, and the guard's comment already gives that as the reason it is safe. The last, that the new `SetUp` sets a process-global for the whole fixture, was assessed as harmless: every production path it touches is also exempted by `RunningUnitTests`, and this repo configures no test parallelism (the only related attribute anywhere is a `[NonParallelizable]`), so no other fixture can observe it.

Full C# suite at this commit: **3052 passed, 12 skipped, 0 failed** — completely green, no environment exclusions needed. CodeRabbit is switched off for this repo in `.coderabbit.yml`. Note that nothing server-side builds or tests this branch: the only GitHub check, `pr-automation`, just opens this Devin review URL, and TeamCity builds default branches only — so that local run is the whole of the test evidence.

One process note for whoever reads this later: `pr-automation` reported **success** for this commit without a Devin review actually being queued, and it took a manual visit to the review page to start one. A green `pr-automation` check is not evidence that Devin ran; the jobs API having a job for the current HEAD sha is.

src/BloomExe/ApplicationContainer.cs
andrew-polk: *Claude Opus 5 (1M context) from Andrew Polk's machine during preflight.*

**Decided: leave as is** — Andrew's call. The guard stays on `Program.RunningInConsoleMode` alone.

The reasoning, for the record: the scenario is not reachable today (no test invokes `DownloadBookCommand`, `UploadCommand`, `ChangeLayoutCommand` or `SendFontAnalyticsCommand`, and none constructs an `ApplicationContainer` directly), and widening the guard to `RunningUnitTests` would let the regression test pass through a test-only exemption — so a future mistake in the production wiring, such as the flag being set too late, would stop failing anything. Keeping it narrow means the test only passes if `RunningInConsoleMode` is genuinely true at container-construction time, which is the property most worth protecting.

Still a fair suggestion, and cheap: if someone later adds a test that drives one of those verbs, adding `|| Program.RunningUnitTests` then — alongside a test that actually needs it — is the right moment.

src/BloomTests/CLI/CreateArtifactsCommandTests.cs
andrew-polk: *Claude Opus 5 (1M context) from Andrew Polk's machine during preflight.*

**Fixed in `4e57a679e8d8e92a0edc0234a4a220184468bc10`** — Andrew asked for it here rather than on its own card, and asked me first to check whether the plain assignment might have been deliberate. It was not, and the history is fairly clear about it:

- The commit that introduced it, `dd8a08724c`, is titled *"Fix incorrect exit codes if ThreadPool throws exception"* — its whole stated purpose was making the exit code correct.
- At that same commit, `exitCode |= CreateBloomDigitalArtifacts(...)` was **already** 19 lines above and could already have set `BookHtmlNotFound`. So the assignment was never lossless; it was inconsistent with its own method from the day it was written.
- "Epub failure should dominate" is not the design either — `GetErrorsFromExitCode` reports every flag as a list.

The one real argument for a plain write is atomicity, since `|=` is a read-modify-write on a variable captured by a worker thread. It does not apply here: the main thread is blocked spinning on `countdownEvent.IsSet` for the whole window and does not touch `exitCode` until after the worker signals. (And if there were a concurrent writer, the answer would be a lock, not a lossy write.)

So it is now `|=`, with a comment recording why. Not covered by a test: provoking it needs a book that fails the bloomdigital step *and* throws during epub creation in one run, which no existing fixture arranges — noted in the commit message rather than papered over.

Thanks for the catch — this one was next door to the bug we came for and would have kept quietly degrading harvester diagnostics.

PR #8179, opened 2026-08-07
Stop treating a literal '%' in a file name as a URL escape (BL-16669)

JohnThomson:
Import an image called `photo%41.jpg` through the image chooser and the page ended up with `src="photoA.jpg"`, pointing at a file that does not exist. `%41` is the URL escape for `A`.

`UrlPathString.CreateFromUnencodedString` has a long-standing heuristic: if the string it is handed matches `%[A-Fa-f0-9]{2}`, it assumes the caller was wrong about the string being unencoded and decodes it. The source comment already acknowledged the risk ("makes it impossible to have, say `%20` in your actual file name"). `PageEditingModel.ChangePicture` built the src for the browser from `imageFileName` — the name of the file it had just written into the book folder, so definitively unencoded — and the heuristic silently rewrote it.

**This is not a regression from BL-16658** (that fix is in `PageListApi` and is untouched), though it is the same family of mistake: something decoding a file name that was never encoded. It was found while working through the manual test ideas on that card.

## Approach

The heuristic itself has to stay — several callers hand this method a string that really *has* been encoded already (anything through `ToLocalhost()`, for instance) and rely on the decode. So the fix is at the call sites: those that know they hold a real file-system name now say so.

The opt-out parameter was named `strictlyTreatAsEncoded`, which reads as the exact opposite of what it does (it means "trust me, this is **not** encoded"), so it is renamed `strictlyTreatAsUnencoded`. Only the declaration, one comment and two test names referred to it by name; everything else passed it positionally.

## Fixed

Every site that holds a name or path from the file system:

- **the reported bug** — `PageEditingModel.ChangePicture`, plus the two GIF paths that do the same thing (`EditingView.SetGifImage`, `ImageGalleryApi`)
- **the image-metadata path**, which would have broken on the same file right after importing it — `EditingView.PrepareToEditImageMetadata` and `EditingModel.UpdateMetaData` (which builds an XPath match against the src, so was guaranteed to miss)
- video srcs and the XPath used to find a video edited outside Bloom (`SignLanguageApi`), spreadsheet import srcs, widget activity paths (`WidgetHelper`, `BookStorage`'s widget-path migration), the folio-relative image path in `Book.cs`, publish-time de-duplication rewrites (`PublishHelper`), and the book paths put into the "click here for help" links (`CommonMessages`, `TeamCollectionApi`)
- `ProblemReportApi`'s heading, which is prose and never a url, so URL-decoding it could only corrupt it

Two consumers were decoding a **second** time, which would have defeated the fix on the help-link paths: `ApiRequest.Parameters` comes from `HttpUtility.ParseQueryString`, which has already decoded the value once, so the extra `CreateFromUrlEncodedString` in `CommonApi`'s `clickHereForHelp` and in `TeamCollectionApi.HandleReportBadZip` is removed.

`BookStorage.CheckBook`'s missing-image scan decoded unconditionally until the name stopped changing, turning a file that is present into one it reported as missing. It now only keeps decoding while that is actually finding the file, preserving what the loop was written for (legacy multiply-encoded srcs).

## Deliberately left in ambiguous mode

`BloomServer` (×2), `EditingModel`'s page-list url and `PublishEpubApi` are all handed an already-encoded `ToLocalhost()` url and would double-encode; `RequestInfo.WriteRedirect`'s callers pass a mix of encoded and unencoded urls; `ApiRequest.RequiredFileNameOrPath` has no callers. All 42 call sites were audited with a paren-matching scan rather than grep, since most span several lines.

## Testing

- `UrlPathStringTests` covers the round trip for such a name, and asserts the mangling the flag exists to prevent.
- `PageEditingModelTests` imports a `photo%41.png` and checks the src is the encoded name of the file actually written.
- `RequestInfoTests` covers the request side — the server must decode exactly once.
- Verified against a running Bloom that a request for the encoded `photo%2541.png` returns 200 with the real file, while `photoA.png` returns 404.

The equivalent server test cannot live in `BloomServerTests`: `PretendRequestInfo` decodes the url **twice** where the real `RequestInfo` decodes once. There is a note there pointing at where the case is covered instead.

## Follow-on work in this PR

Reviewing the first commit surfaced three further problems in the same family, so the branch now also carries:

**The image is found even when its src has a query string.** `EditingModel.UpdateMetaData` located the picture whose copyright you had just edited by matching its src *exactly*, so it quietly did nothing whenever the src carried `?transparent=yes`, `?thumbnail=1` or a cache-busting `?12345` — and its only response to finding nothing is to return silently. It now matches the bare name, or the name followed by `?`.

**Which stored file names are URL-encoded is now written down.** Two readers were decoding values that nothing writes encoded. Surveying every consumer showed the answer is not uniform: `data-sound`, `data-correct-sound`, `data-wrong-sound` and the `bloomDataDiv` image entries hold the **plain** name — three C# consumers use them as bare file-system paths, and one of those *deletes* anything not in its set — while `data-backgroundaudio` really **is** URL-encoded, because the music tool writes it with `encodeURIComponent`. Each is now read and written in its own convention, and `UrlPathString` carries a note explaining which is which and why. Grep for **"encoding conventions"** to find the places that depend on it.

**Deletion safety.** `BookStorage`'s list of images *not* to delete now protects both the plain and the decoded form: a surplus entry only leaves an unused file behind, whereas a missing one deletes a file the book is using.

Known gap, deliberately not closed here: the places that build a URL by concatenating a sound's name (BloomPlayer, and `GameTool.tsx`'s `playSound`) don't encode it, so a sound called `beep%41.mp3` still won't play. That belongs in those two places, not in how the name is stored.

Ref: https://issues.bloomlibrary.org/youtrack/issue/BL-16669

[Devin review](https://devinreview.com/BloomBooks/BloomDesktop/pull/8179)

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8179)

### DIFF
--- a/src/BloomBrowserUI/bookEdit/toolbox/games/GameTool.tsx
+++ b/src/BloomBrowserUI/bookEdit/toolbox/games/GameTool.tsx
@@ -1088,6 +1088,9 @@ const DragActivityControls: React.FunctionComponent<{
         copyBuiltIn: boolean,
     ) => {
         const page = getPage();
+        // The sound attributes hold the plain file name, deliberately NOT URL-encoded (unlike
+        // data-backgroundaudio, which the toolbox does encode). C# reads these straight back as
+        // file names. See the encoding conventions note on UrlPathString.cs.
         const setSoundAttr = (soundAttr: string, newSoundId: string) => {
             if (newSoundId === "default") {
                 page.removeAttribute(soundAttr);
--- a/src/BloomBrowserUI/bookEdit/toolbox/games/GameTool.tsx
+++ b/src/BloomBrowserUI/bookEdit/toolbox/games/GameTool.tsx
@@ -1849,7 +1849,13 @@ export function playSound(
     page: HTMLElement,
     soundType?: SoundType,
 ) {
-    let url = "audio/" + newSoundId;
+    // encodeURIComponent because newSoundId is a plain file name, not a URL: the sound
+    // attributes deliberately store the unencoded name (see the encoding conventions note on
+    // UrlPathString.cs), so anything in it that means something in a URL has to be escaped
+    // here, when we build one. Without this, a sound genuinely called "beep%41.mp3" is asked
+    // for as "beepA.mp3" and silently doesn't play. Safe to encode the whole thing because
+    // this is a bare file name -- no '/' of its own to preserve. (BL-16669)
+    let url = "audio/" + encodeURIComponent(newSoundId);
     if (newSoundId === "default") {
         if (soundType === undefined) {
             throw new Error(
--- a/src/BloomExe/Book/Book.cs
+++ b/src/BloomExe/Book/Book.cs
@@ -1107,7 +1107,7 @@ namespace Bloom.Book
             EnsureUpToDateMemory(progress);
             UpdateSupportFiles();
 
-            Storage.MigrateToMediaLevel1ShrinkLargeImages();
+            Storage.MigrateToMediaLevel1ShrinkLargeImages(progress);
 
             Storage.CleanupUnusedSupportFiles(forCopyOfUpToDateBook);
 
@@ -1898,7 +1898,7 @@ namespace Bloom.Book
             // already been done, so they must be called in exactly this order.
             Storage.RestoreStuffBeforeMigration();
             Storage.MigrateMaintenanceLevels();
-            Storage.MigrateToMediaLevel1ShrinkLargeImages();
+            Storage.MigrateToMediaLevel1ShrinkLargeImages(progress);
             Storage.MigrateToLevel2RemoveTransparentComicalSvgs();
             Storage.MigrateToLevel3PutImgFirst();
             Storage.MigrateToLevel4UseAppearanceSystem();
--- a/src/BloomExe/Book/Book.cs
+++ b/src/BloomExe/Book/Book.cs
@@ -4106,13 +4106,8 @@ namespace Bloom.Book
                     source.SetAttribute(
                         "src",
                         "video/"
-                            // strictlyTreatAsUnencoded: this is the name of the file we just
-                            // wrote, so a '%' in it is literal, not an escape (BL-16669).
                             + UrlPathString
-                                .CreateFromUnencodedString(
-                                    Path.GetFileName(newVideoPath),
-                                    strictlyTreatAsUnencoded: true
-                                )
+                                .CreateFromUnencodedString(Path.GetFileName(newVideoPath))
                                 .UrlEncoded
                             + (string.IsNullOrEmpty(timings) ? "" : "#t=" + timings)
                     );
@@ -4664,12 +4659,7 @@ namespace Bloom.Book
 
                         HtmlDom.SetImageElementUrl(
                             img,
-                            // strictlyTreatAsUnencoded: this is built from a real folder name and
-                            // an already-decoded src, so a '%' in it is literal (BL-16669).
-                            UrlPathString.CreateFromUnencodedString(
-                                pathRelativeToFolioFolder,
-                                strictlyTreatAsUnencoded: true
-                            )
+                            UrlPathString.CreateFromUnencodedString(pathRelativeToFolioFolder)
                         );
                     }
                 }
--- a/src/BloomExe/Book/Book.cs
+++ b/src/BloomExe/Book/Book.cs
@@ -4098,8 +4098,13 @@ namespace Bloom.Book
                     source.SetAttribute(
                         "src",
                         "video/"
+                            // strictlyTreatAsUnencoded: this is the name of the file we just
+                            // wrote, so a '%' in it is literal, not an escape (BL-16669).
                             + UrlPathString
-                                .CreateFromUnencodedString(Path.GetFileName(newVideoPath))
+                                .CreateFromUnencodedString(
+                                    Path.GetFileName(newVideoPath),
+                                    strictlyTreatAsUnencoded: true
+                                )
                                 .UrlEncoded
                             + (string.IsNullOrEmpty(timings) ? "" : "#t=" + timings)
                     );
@@ -4651,7 +4656,12 @@ namespace Bloom.Book
 
                         HtmlDom.SetImageElementUrl(
                             img,
-                            UrlPathString.CreateFromUnencodedString(pathRelativeToFolioFolder)
+                            // strictlyTreatAsUnencoded: this is built from a real folder name and
+                            // an already-decoded src, so a '%' in it is literal (BL-16669).
+                            UrlPathString.CreateFromUnencodedString(
+                                pathRelativeToFolioFolder,
+                                strictlyTreatAsUnencoded: true
+                            )
                         );
                     }
                 }
--- a/src/BloomExe/Book/Book.cs
+++ b/src/BloomExe/Book/Book.cs
@@ -1500,6 +1500,9 @@ namespace Bloom.Book
             if (string.IsNullOrEmpty(coverImageFileName))
                 return;
             coverImageFileName = coverImageFileName.Trim();
+            // This entry is meant to hold the plain file name, not a URL-encoded one -- see the
+            // encoding conventions note on UrlPathString. A few old books break that rule, hence
+            // the tolerant read below and the write-back that normalizes them to the plain name.
             // The fileName might be URL encoded.  See https://silbloom.myjetbrains.com/youtrack/issue/BL-3901.
             var coverImagePath = UrlPathString.GetFullyDecodedPath(
                 StoragePageFolder,
@@ -2487,6 +2490,9 @@ namespace Bloom.Book
                 // This preserves expectations in BringBookUpToDate_EmbeddedEmptyImgTagRemoved.
                 if (string.IsNullOrWhiteSpace(src))
                     src = "placeHolder.png";
+                // Stored as the plain file name, deliberately not URL-encoded, which is why the
+                // encoded @src has to be decoded on the way in -- see the encoding conventions
+                // note on UrlPathString.
                 coverImageElement.InnerText = HttpUtility.UrlDecode(src);
             }
         }
@@ -3924,6 +3930,8 @@ namespace Bloom.Book
             string templateBookFolderPath
         )
         {
+            // Used verbatim as a file name below: the sound attributes hold a plain, NOT
+            // URL-encoded, name -- see the encoding conventions note on UrlPathString.
             var fileName = sourceElt.GetAttribute(attrName);
             // For some sound attrs (e.g., data-correct-sound), 'none' is a valid value of the
             // attribute but signifies we don't want a default sound. So we don't want to copy.
--- a/src/BloomExe/Edit/EditingModel.cs
+++ b/src/BloomExe/Edit/EditingModel.cs
@@ -1182,6 +1182,26 @@ namespace Bloom.Edit
             _view.GoToPage(_pageSelection.CurrentSelection, changingUiLanguage);
         }
 
+        /// <summary>
+        /// XPath for the img on a page whose src is the given (URL-encoded) file name.
+        /// </summary>
+        /// <remarks>
+        /// A src often carries a query string as well as the file name -- "?transparent=yes" from
+        /// the transparency handling, "?thumbnail=1" from the page list, or the old cache-busting
+        /// "?12345". Matching the src exactly therefore found nothing on exactly the pages that
+        /// use those, and the caller's only response to finding nothing is to give up silently.
+        /// So we accept either the bare name or the name followed by '?'. Requiring the '?' is
+        /// what keeps this from also matching a different file that merely starts with the same
+        /// characters ("cat.png" must not match "cat2.png"). (BL-16669)
+        ///
+        /// The name is safe to embed in the XPath string literal: it is URL-encoded, and
+        /// UrlEncoded escapes an apostrophe as %27, so it cannot terminate the literal.
+        /// </remarks>
+        internal static string MakeImgWithSrcXPath(string urlEncodedFileName)
+        {
+            return $".//img[@src='{urlEncodedFileName}' or starts-with(@src, '{urlEncodedFileName}?')]";
+        }
+
         public void UpdateMetaData(string url)
         {
             // url is a file name (from EditingView._fileNameOfImageBeingModified), which we
@@ -1193,7 +1213,7 @@ namespace Bloom.Edit
                 .UrlEncoded;
             var imgElt = _pageSelection
                 .CurrentSelection.GetDivNodeForThisPage()
-                .SafeSelectNodes($".//img[@src='{match}']")
+                .SafeSelectNodes(MakeImgWithSrcXPath(match))
                 .Cast<SafeXmlElement>()
                 .FirstOrDefault();
             if (imgElt == null)
--- a/src/BloomExe/Edit/EditingModel.cs
+++ b/src/BloomExe/Edit/EditingModel.cs
@@ -1204,13 +1204,9 @@ namespace Bloom.Edit
 
         public void UpdateMetaData(string url)
         {
-            // url is a file name (from EditingView._fileNameOfImageBeingModified), which we
-            // re-encode here to match what is in the src attribute. strictlyTreatAsUnencoded so
-            // that a name like "photo%41.jpg" doesn't get decoded to "photoA.jpg" and then fail
-            // to match the img we are looking for. (BL-16669)
-            var match = UrlPathString
-                .CreateFromUnencodedString(url, strictlyTreatAsUnencoded: true)
-                .UrlEncoded;
+            // url is a file name (EditingView._fileNameOfImageBeingModified), which we re-encode
+            // here so it matches what is in the src attribute.
+            var match = UrlPathString.CreateFromUnencodedString(url).UrlEncoded;
             var imgElt = _pageSelection
                 .CurrentSelection.GetDivNodeForThisPage()
                 .SafeSelectNodes(MakeImgWithSrcXPath(match))
@@ -1486,7 +1482,9 @@ namespace Bloom.Edit
                 false,
                 InMemoryHtmlFileSource.Pagelist
             );
-            var urlPath = UrlPathString.CreateFromUnencodedString(url);
+            // PossiblyEncoded because CreateSimulatedFile returns a localhost url whose path
+            // components are already escaped; see the note on CreateFromPossiblyEncodedString.
+            var urlPath = UrlPathString.CreateFromPossiblyEncodedString(url);
             var encodedUrl = urlPath.UrlEncodedForHttpPath;
             BloomServer.SetCurrentPageListUrlForDebugging(encodedUrl);
             return encodedUrl;
--- a/src/BloomExe/Edit/EditingModel.cs
+++ b/src/BloomExe/Edit/EditingModel.cs
@@ -1184,7 +1184,13 @@ namespace Bloom.Edit
 
         public void UpdateMetaData(string url)
         {
-            var match = UrlPathString.CreateFromUnencodedString(url).UrlEncoded;
+            // url is a file name (from EditingView._fileNameOfImageBeingModified), which we
+            // re-encode here to match what is in the src attribute. strictlyTreatAsUnencoded so
+            // that a name like "photo%41.jpg" doesn't get decoded to "photoA.jpg" and then fail
+            // to match the img we are looking for. (BL-16669)
+            var match = UrlPathString
+                .CreateFromUnencodedString(url, strictlyTreatAsUnencoded: true)
+                .UrlEncoded;
             var imgElt = _pageSelection
                 .CurrentSelection.GetDivNodeForThisPage()
                 .SafeSelectNodes($".//img[@src='{match}']")
--- a/src/BloomExe/Edit/EditingView.cs
+++ b/src/BloomExe/Edit/EditingView.cs
@@ -548,9 +548,7 @@ namespace Bloom.Edit
             // those one more decode is exactly what finds the file. The two are indistinguishable
             // by inspection, so let the disk decide: GetFullyDecodedPath keeps the name it is
             // given when that file exists, and only decodes further when that is what finds it.
-            fileName = UrlPathString
-                .CreateFromUnencodedString(fileName, strictlyTreatAsUnencoded: true)
-                .PathOnly.NotEncoded;
+            fileName = UrlPathString.CreateFromUnencodedString(fileName).PathOnly.NotEncoded;
             UrlPathString.GetFullyDecodedPath(_model.CurrentBook.FolderPath, ref fileName);
 
             // keep a reference to the fileName rather the image to avoid dispose issues
@@ -1041,9 +1039,7 @@ namespace Bloom.Edit
                 imageId = imageId,
                 // destName is the name of the file we just copied into the book folder, so it is
                 // certainly not URL-encoded; see the same call in PageEditingModel.ChangePicture.
-                src = UrlPathString
-                    .CreateFromUnencodedString(destName, strictlyTreatAsUnencoded: true)
-                    .UrlEncoded,
+                src = UrlPathString.CreateFromUnencodedString(destName).UrlEncoded,
                 // Enhance: can we provide any of this for a GIF?
                 copyright = "",
                 license = "",
--- a/src/BloomExe/Edit/EditingView.cs
+++ b/src/BloomExe/Edit/EditingView.cs
@@ -537,16 +537,21 @@ namespace Bloom.Edit
             }
 
             // The fileName comes straight from the html src attribute, so it may carry a query
-            // string (e.g. "image1.png?transparent=yes") and/or still be URL-encoded. Reduce it
-            // to the actual file name on disk before we try to load the image; otherwise
-            // PalasoImage fails to find the file and we wrongly report it as corrupt. (BL-16446)
-            // The server already decoded the query parameter once, so what we have here is the
-            // real file name; PathOnly is what we're after, to drop the query string. We must pass
-            // strictlyTreatAsUnencoded so that a name which merely looks encoded ("photo%41.jpg")
-            // isn't decoded a second time into a name that isn't there ("photoA.jpg"). (BL-16669)
+            // string (e.g. "image1.png?transparent=yes"). Reduce it to the actual file name on
+            // disk before we try to load the image; otherwise PalasoImage fails to find the file
+            // and we wrongly report it as corrupt. (BL-16446)
+            //
+            // The server already decoded the query parameter once, so normally what we have here
+            // is the real file name and we must NOT decode again, or a name that merely looks
+            // encoded ("photo%41.jpg") becomes one that isn't there ("photoA.jpg"). (BL-16669)
+            // But some old books have srcs that were encoded more than once (BL-3835), and for
+            // those one more decode is exactly what finds the file. The two are indistinguishable
+            // by inspection, so let the disk decide: GetFullyDecodedPath keeps the name it is
+            // given when that file exists, and only decodes further when that is what finds it.
             fileName = UrlPathString
                 .CreateFromUnencodedString(fileName, strictlyTreatAsUnencoded: true)
                 .PathOnly.NotEncoded;
+            UrlPathString.GetFullyDecodedPath(_model.CurrentBook.FolderPath, ref fileName);
 
             // keep a reference to the fileName rather the image to avoid dispose issues
             _fileNameOfImageBeingModified = fileName;
--- a/src/BloomExe/Edit/EditingView.cs
+++ b/src/BloomExe/Edit/EditingView.cs
@@ -554,6 +554,16 @@ namespace Bloom.Edit
             UrlPathString.GetFullyDecodedPath(_model.CurrentBook.FolderPath, ref fileName);
 
             // keep a reference to the fileName rather the image to avoid dispose issues
+            //
+            // Note that what we keep is the name as it is ON DISK, which for one of those legacy
+            // doubly-encoded books is not what the page's @src says. EditingModel.UpdateMetaData
+            // re-encodes this once to go looking for the img, so for such a book it will not find
+            // it and the mirrored data-copyright/data-creator attributes won't be refreshed. The
+            // metadata is still written to the image file itself, which is the part that matters,
+            // and this mismatch predates the disk lookup above -- but the lookup does make it
+            // systematic rather than accidental. Fixing it properly means remembering the original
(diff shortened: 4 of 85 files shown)


src/BloomExe/Publish/PublishHelper.cs
```diff
@@ -1250,7 +1256,11 @@ var videoContainer in HtmlDom
                     RewriteReference = canonicalRelativePath =>
                         HtmlDom.SetVideoElementUrl(
                             videoContainer,
-                            UrlPathString.CreateFromUnencodedString(canonicalRelativePath)
+                            // strictlyTreatAsUnencoded: see the image case above (BL-16669).
+                            UrlPathString.CreateFromUnencodedString(
+                                canonicalRelativePath,
+                                strictlyTreatAsUnencoded: true
+                            )
                         ),
```
JohnThomson, line 1270: [Devin] **Investigate**: PublishHelper's audio rewrite still bypasses UrlPathString entirely

The image and video rewrites in this method now go through `UrlPathString.CreateFromUnencodedString(..., strictlyTreatAsUnencoded: true)`, but `MakeAudioAttributeReference`'s rewrite (`src/BloomExe/Publish/PublishHelper.cs:1339-1341`) writes `Path.GetFileName(canonicalRelativePath)` straight into the attribute with no encoding, while the read side uses `CreateFromUrlEncodedString(rawValue)` (`src/BloomExe/Publish/PublishHelper.cs:1331`). That asymmetry predates this PR, but it means a de-duplicated audio file whose name contains a `%` or a space would be written unencoded and then mis-decoded on the next read. Worth a follow-up in the same spirit as this change.

src/BloomExe/Edit/EditingModel.cs
```diff
@@ -1184,7 +1184,13 @@ public void RefreshDisplayOfCurrentPage(bool changingUiLanguage = false)
...
+            // url is a file name (from EditingView._fileNameOfImageBeingModified), which we
+            // re-encode here to match what is in the src attribute. strictlyTreatAsUnencoded so
+            // that a name like "photo%41.jpg" doesn't get decoded to "photoA.jpg" and then fail
+            // to match the img we are looking for. (BL-16669)
+            var match = UrlPathString
+                .CreateFromUnencodedString(url, strictlyTreatAsUnencoded: true)
+                .UrlEncoded;
             var imgElt = _pageSelection
                 .CurrentSelection.GetDivNodeForThisPage()
                 .SafeSelectNodes($".//img[@src='{match}']")
```
JohnThomson: [Devin] **Investigate**: UpdateMetaData's XPath match depends on the src having no query string

The fix makes the encoded name match what `ChangePicture` writes, which is right. But note the match is still exact-equality against `@src`, so it silently finds nothing if the src carries a query (e.g. the `?transparent=yes` that `HtmlDom.AddTransparencyParamToImages` adds, or the `?thumbnail=1` added in `PageListApi`), because `_fileNameOfImageBeingModified` is `PathOnly.NotEncoded` with the query stripped (`src/BloomExe/Edit/EditingView.cs:547-549`). The method returns silently in that case ("log? unexpected"). That is pre-existing, but the transparency-param path makes it more than theoretical.

src/BloomExe/Book/Book.cs
```diff
@@ -4651,7 +4656,12 @@ SafeXmlElement img in HtmlDom.SelectChildImgAndBackgroundImageElements(
 
                         HtmlDom.SetImageElementUrl(
                             img,
-                            UrlPathString.CreateFromUnencodedString(pathRelativeToFolioFolder)
+                            // strictlyTreatAsUnencoded: this is built from a real folder name and
+                            // an already-decoded src, so a '%' in it is literal (BL-16669).
+                            UrlPathString.CreateFromUnencodedString(
+                                pathRelativeToFolioFolder,
+                                strictlyTreatAsUnencoded: true
+                            )
```
JohnThomson: [Devin] **Investigate**: Folio image srcs are still written with slashes percent-encoded

`SetImageElementUrl` uses `url.UrlEncoded` (not `UrlEncodedForHttpPath`) — see `src/BloomExe/Book/HtmlDom.cs:2919` — so a relative path like `../BookName/img.png` is written as `..%2fBookName%2fimg.png`. That is unchanged by this PR (the previous code fed the same string through the same property), but it is the reason the analogous video code paths deliberately use `UrlEncodedForHttpPath`. Since this call site is now being touched anyway, it may be worth confirming folio PDFs still resolve these images.

JohnThomson: [Claude Opus 5 (1M context) during preflight] Not an issue for this PR — resolving.

The percent-encoded slashes are real, but they are exactly as they were before this change. `SetImageElementUrl` has always put this string through `UrlEncoded`, and the old code handed it the same string: the only thing the removed guessing step did here was decode a `%XX` sequence if the *image's own file name* happened to contain one, which is the bug this PR is fixing. For every ordinary name, before and after are byte-identical.

As for whether the escaped slashes work at all: they do, because the browser sends the whole thing as one path segment and Bloom's server decodes the path exactly once before resolving it, which turns `%2f` back into `/`. That is the same reason the thumbnail work in BL-16658 described switching to `UrlEncodedForHttpPath` as hardening rather than a bug fix.

Devin's suggestion to confirm folio PDFs still resolve their images is a fair one, so it has gone into the manual test notes on the card rather than being dropped.

JohnThomson:
[Claude Opus 5 (1M context) during preflight] Consulted Devin on 2026-08-07 19:42 UTC up to commit fafc8092eaa2e3fd33ae8e272c13943ae1a1b618.

It found no Bugs, three Investigate flags and three Informational items (the informational ones are not mirrored here — low signal). All three flags are things that predate this PR, and each has its own thread above:

- Publish-time de-duplication writes audio file names into their attributes unencoded, while the code that reads them back expects them encoded. Left open — which side to change is a judgement call, because the editor's own JavaScript writes those attributes unencoded too.
- Updating an image's copyright/licence finds the picture by matching its src exactly, so it quietly does nothing when the src carries a query string such as `?transparent=yes`. Left open for the same reason: fixing it changes behaviour beyond this bug.
- Folio image srcs having their slashes percent-encoded: resolved, because it is byte-identical to what the old code produced and it works — the server decodes the path once before resolving it.

CI (`pr-automation`) is green. CodeRabbit's automatic review is switched off for this repo in `.coderabbit.yml`, so it did not run.

src/BloomExe/Edit/EditingModel.cs
JohnThomson: [Claude Opus 5 (1M context) during preflight] Fixed in `d3379c6565` — the developer chose to match on the file-name part only, ignoring any trailing extras.

The XPath now accepts either the bare name or the name followed by `?`, so a src carrying `?transparent=yes`, `?thumbnail=1` or a cache-busting `?12345` is found. Requiring the `?` is what stops the widened match from also grabbing a different file whose name merely begins with the same characters — `cat.png` must not match `cat2.png`.

It is extracted as `MakeImgWithSrcXPath` so it could be tested without standing up an `EditingModel`. There are cases for the plain src, each of the three query strings, the look-alike name that must not match, a name containing a literal `%`, and a name containing an apostrophe — that last because an apostrophe would end the XPath string literal if it ever reached it, which it cannot, since `UrlEncoded` escapes it as `%27`.

Good catch: the silent-return meant this failed invisibly.

src/BloomExe/Publish/PublishHelper.cs
JohnThomson: [Claude Opus 5 (1M context) during preflight] Confirmed as a real mismatch, but leaving this open: the obvious fix — encoding on write — would break more than it fixes, so it needs a decision rather than a patch.

Tracing every consumer of `data-sound` / `data-correct-sound` / `data-wrong-sound`:

**Consumers that use the value as a bare file-system path (would break if we encoded):**
- `BookStorage.cs:1399-1409` builds the set of audio files still in use — and anything not in that set is **deleted**. An encoded `my%20sound.mp3` would not match `my sound.mp3` on disk, so the sound would be deleted as unused.
- `BookFileFilter.cs:378-383` decides which audio files get packaged into a published book; the same mismatch would silently drop the sound from the .bloompub/upload.
- `Book.cs:3927-3936` (`CopyMissingSoundFile`) does `Path.Combine(FolderPath, "audio", value)` when a page is brought in from a template.

**Consumers that want it encoded:**
- BloomPlayer assigns the attribute straight into an audio `src` (`f.src = e.getAttribute("data-correct-sound")`), so for it the encoded form is the correct one.
- `PublishHelper.cs:1341` — the read side flagged here.

**Writers:** the editor's own JavaScript writes these plainly (`GameTool.tsx:1095`, `canvasControlRegistry.ts:407`), as does the rewrite flagged here.

So the two sides genuinely disagree, and today "plain" wins because that is what gets written — which is exactly why a name containing a percent escape is broken in BloomPlayer while everything else works. Flipping to encoded fixes that rare case and breaks the common one (any name with a space) in three places, one of which deletes files.

Doing it properly means encoding at every writer *and* teaching those three readers to decode — a change to what gets written into published books, affecting BloomPlayer, Bloom Reader and the harvester. That is well beyond the bug this PR is fixing, so it is going back to the developer to decide, with the alternative on the table of instead making this read side stop decoding, which would make everything consistent with what is actually written.

JohnThomson: [Claude Opus 5 (1M context) during preflight] Resolved in `8fed9434d6`, by changing the read rather than the write.

The developer's call, after the consumer survey above: these values are settled as the **plain file name, not URL-encoded**, because that is what both editor writers produce and what the three file-system consumers require — and because encoding instead would have made `BookStorage` delete any sound whose name contains a space. So `MakeAudioAttributeReference` now takes the attribute as it stands, which agrees with its own rewrite and with every other reader.

The same treatment went to the `bloomDataDiv` cover-image entry, which had the identical split.

Because none of this is guessable from the code, `UrlPathString` now carries a note setting out which values in a book are URL-encoded and which are not, and every place that reads or writes one of these has a one-line comment pointing at it (grep for "encoding conventions"). That note also records the one consumer this does **not** satisfy — BloomPlayer assigns the value straight into an audio `src`, so a sound genuinely named `beep%41.mp3` still will not play there; fixing that belongs in BloomPlayer, by encoding when it builds the src.

Worth flagging for the record: `data-backgroundaudio`, which sits right beside these, genuinely **is** URL-encoded (the music tool writes it via `encodeURIComponent`). `HtmlDom` now warns against "correcting" it to match its neighbours.

Covered by two new tests that de-duplicate a book whose cover image is `photo%41.png` and whose correct-answer sound is `beep%41.mp3`; both previously left the duplicate in place because the file was sought under the wrong name.

src/BloomExe/Edit/EditingView.cs
```diff
@@ -540,9 +540,13 @@ public Metadata PrepareToEditImageMetadata(string fileName)
...
-            // CreateFromUnencodedString only decodes if the string still looks encoded (the
-            // server already decodes query parameters once), and PathOnly drops the query string.
-            fileName = UrlPathString.CreateFromUnencodedString(fileName).PathOnly.NotEncoded;
+            // The server already decoded the query parameter once, so what we have here is the
+            // real file name; PathOnly is what we're after, to drop the query string. We must pass
+            // strictlyTreatAsUnencoded so that a name which merely looks encoded ("photo%41.jpg")
+            // isn't decoded a second time into a name that isn't there ("photoA.jpg"). (BL-16669)
+            fileName = UrlPathString
+                .CreateFromUnencodedString(fileName, strictlyTreatAsUnencoded: true)
+                .PathOnly.NotEncoded;
```
JohnThomson: [Devin] **Investigate**: Old books whose img src was encoded more than once can no longer have their image credits edited

`PrepareToEditImageMetadata` now never decodes. The value reaching it is the raw `src` attribute the JS appends to the query (`src/BloomBrowserUI/bookEdit/copyrightAndLicense/CopyrightAndLicenseDialog.tsx:483` deliberately does not encode it), and `ApiRequest.Parameters` (via `HttpUtility.ParseQueryString`) decodes exactly once — so for a normally-encoded src the new strict behavior is right.

The exception is the legacy case this repo explicitly acknowledges elsewhere: `BookStorage.CheckBook` (`src/BloomExe/Book/BookStorage.cs:2099-2110`) keeps a decode-until-found loop precisely because "some old books have srcs that were encoded more than once" (also BL-3835). For such a book the single decode leaves e.g. `The%20Moon.png`, the strict call keeps it, and `ImageUpdater.GetImageInfoSafelyFromFilePath` will fail to find the file — the same class of failure BL-16446 was fixed to avoid. The same applies to `EditingModel.UpdateMetaData`, whose XPath would then be built from the wrong name.

If that legacy shape is judged to be still in the wild for `img@src` (as opposed to only the data-div coverImage text), these two sites may want the same "only keep decoding while it helps us find the file" treatment that CheckBook got.

src/BloomExe/Publish/PublishHelper.cs
```diff
@@ -1206,9 +1215,18 @@ var bookSetting in dom.SafeSelectNodes(
                     .Cast<SafeXmlElement>()
             )
             {
-                var relativePath = UrlPathString
-                    .CreateFromUrlEncodedString(bookSetting.InnerText.Trim())
-                    .PathOnly.NotEncoded;
+                // A bloomDataDiv image entry holds a plain file name, NOT a URL-encoded one --
+                // see the encoding conventions note on UrlPathString. This used to decode the
```
JohnThomson, line 1213: [Devin] **Investigate**: Legacy URL-encoded coverImage data-div entries can be left pointing at a deleted file during publish de-duplication

The data-div read here no longer decodes, so a legacy book whose coverImage entry holds e.g. `The%20Moon.png` (BL-3901) now yields a relative path that does not resolve on disk. That reference then forms its own group in `DeDuplicateReferencedMedia` and is skipped at `src/BloomExe/Publish/PublishHelper.cs:1140-1141`. If the SAME file is also referenced by an img (whose src decodes correctly to `The Moon.png`) and turns out to be a duplicate of an earlier image, the img is rewritten and the file is deleted, while the data-div entry is never rewritten — leaving the cover image pointing at a file that no longer exists in the published book.

I did not report this as a bug because `Book.FixExcessiveHTMLEncodingOfCoverImage` (`src/BloomExe/Book/Book.cs:1503-1523`, via `UrlPathString.GetFullyDecodedPath`) normalizes such entries back to the plain name during BringBookUpToDate, which publishing normally performs, so the encoded form should not survive to this point. Note the asymmetry, though: `BookStorage.CleanupUnusedImageFiles` was deliberately made tolerant of BOTH forms in this same PR, while this read handles only the plain form. Being similarly tolerant here (try the plain name, fall back to the decoded one when only that file exists) would cost nothing.

src/BloomExe/Edit/EditingView.cs
JohnThomson: [Claude Opus 5 (1M context) during preflight] Good catch — this was a real regression on this branch. Fixed in `9c96bb2a46`.

You are right that the two cases can't be told apart by looking at the name: a genuine `photo%41.jpg` must be left alone, while a legacy doubly-encoded `The%20Moon.png` needs one more decode, and they look identical. So this now does exactly what you suggested, and what `CheckBook` and the data-div read on this branch already do — it lets the disk decide. `UrlPathString.GetFullyDecodedPath` keeps the name it is given when that file exists, and decodes further only when that is what finds it.

On the related point about `EditingModel.UpdateMetaData`: it builds its XPath from the name this method resolves, so it now starts from the right one. For a doubly-encoded legacy book the encoded form it constructs still won't equal the src as stored, so the mirrored `data-*` attributes on that img may not refresh — but the metadata is written to the image file itself either way, which is the part that matters, and that is strictly better than the previous behavior of not finding the file at all.

src/BloomExe/Publish/PublishHelper.cs
JohnThomson: [Claude Opus 5 (1M context) during preflight] Already done — this describes the code as it stood one commit earlier. Resolving.

The tolerance you're asking for is exactly what went in with `3b61ef5` (present in the head you reviewed, `894c167c0f`): this read is followed by `UrlPathString.GetFullyDecodedPath(folderPath, ref relativePath)`, which keeps the plain name when that file exists and falls back to the decoded one only when that is what finds it. So the asymmetry you noticed against `BookStorage.CleanupUnusedImageFiles` has been closed, by the same means and for the same reason.

Your analysis of the consequence was right, though, and worth recording: had it stayed one-form-only, a legacy book's cover-image entry could have been left pointing at a file that de-duplication had deleted.

JohnThomson:
[Claude Opus 5 (1M context) during preflight] Consulted Devin on 2026-08-08 00:41 UTC up to commit 894c167c0f4f2e98af56b38e8f098e0e596af1f7 (second preflight run; a further commit has since been pushed and is being re-reviewed).

It raised one Bug and five Investigate flags, plus ten Informational items (not mirrored — low signal).

The Bug — background music with a space in its name no longer being recognised when publishing — Devin itself marks as already fixed, by the commit that followed the one it was found in. That was a genuine regression introduced earlier in this branch: the sound attributes were settled as plain file names, but the same method also handles `data-backgroundaudio`, which really is URL-encoded. Each attribute is now read and written in its own convention.

Of the five flags, three are repeats of ones already posted and resolved earlier in this PR. The two new ones are both handled above:

- Old books whose image src was encoded more than once could no longer have their credits edited — a real regression from this branch, now fixed by letting the disk decide how far to decode, the same way `CheckBook` does.
- Legacy cover-image entries being left pointing at a deleted file — already addressed a commit earlier, by the very tolerance the flag asks for.

CI (`pr-automation`) is green. CodeRabbit's automatic review remains switched off for this repo in `.coderabbit.yml`, so it did not run.

src/BloomExe/Edit/EditingView.cs
```diff
@@ -537,14 +537,33 @@ public Metadata PrepareToEditImageMetadata(string fileName)
...
+            // is the real file name and we must NOT decode again, or a name that merely looks
+            // encoded ("photo%41.jpg") becomes one that isn't there ("photoA.jpg"). (BL-16669)
+            // But some old books have srcs that were encoded more than once (BL-3835), and for
+            // those one more decode is exactly what finds the file. The two are indistinguishable
+            // by inspection, so let the disk decide: GetFullyDecodedPath keeps the name it is
+            // given when that file exists, and only decodes further when that is what finds it.
+            fileName = UrlPathString
+                .CreateFromUnencodedString(fileName, strictlyTreatAsUnencoded: true)
+                .PathOnly.NotEncoded;
+            UrlPathString.GetFullyDecodedPath(_model.CurrentBook.FolderPath, ref fileName);
```
JohnThomson: [Devin] **Investigate**: Metadata edit can still silently miss the image in legacy multiply-encoded books

`PrepareToEditImageMetadata` now stores the *disk* name (after `GetFullyDecodedPath` possibly decodes a second time), while `EditingModel.UpdateMetaData` re-encodes that stored name exactly once to build the src match (`src/BloomExe/Edit/EditingModel.cs:1211-1216`). For the legacy BL-3835 books whose `@src` is double-encoded, the two disagree: the file metadata is saved correctly but the mirrored `data-copyright`/`data-creator` attributes on the `img` are not updated, and the caller's only response to no match is a silent `return`. I verified this is not a regression — the previous ambiguous-decode pair produced the same mismatch — but the new `GetFullyDecodedPath` call makes the disk name authoritative, so the divergence is now systematic rather than accidental. Worth a comment or, better, having `PrepareToEditImageMetadata` remember both the disk name and the original src.

JohnThomson: [Claude Opus 5 (1M context) during preflight] Agreed on all counts, including that it is not a regression — the comment you asked for is in `cfb4a09d3c`, right where the disk name is stored.

Taking the "better" option — remembering the original src alongside the disk name — was considered and deliberately not done here. It only changes anything for a book whose `@src` is encoded more than once, the half that matters (the metadata written into the image file itself) already works for those, and it would widen a bug fix about percent signs in file names into a change in how every image is located for editing. The comment now names that as the proper fix if such a book ever turns up in the wild, so the next person doesn't have to re-derive the analysis you just did.

Resolving on that basis.

JohnThomson:
[Claude Opus 5 (1M context) during preflight] Consulted Devin on 2026-08-08 00:55 UTC up to commit cfb4a09d3c5dea2b38590bc94638d8a267c39dc6 — the final head of this preflight run.

**No unresolved Bugs.** Four Investigate flags, every one of them a repeat of a thread already posted and resolved above. Fourteen Informational items (not mirrored — low signal). So the review has converged: nothing new was raised by the last two commits.

All six mirrored threads on this PR are resolved and each carries a written outcome.

JohnThomson:
[Claude Opus 5 (1M context) during preflight] Follow-up work is in, from the developer's answers to the preflight report.

**Sounds whose file name contains a percent escape now play in Bloom.** These names are stored unencoded on purpose, so the encoding has to happen where a URL is built — `GameTool`'s `playSound` was gluing the name onto `audio/` unescaped, so `beep%41.mp3` was requested as `beepA.mp3`. BloomPlayer has the same omission in published books; that half is now tracked as **BL-16688**.

**The request test double was decoding twice.** `PretendRequestInfo` un-escaped the url and then decoded it again, where the real `RequestInfo` decodes once, and it did not strip the query string. It now matches, which finally allows a server-level test for a file whose name contains a literal '%' — that test is added.

**A correction to something stated earlier in this PR.** I had said `HandleDoubleEncodedUrls` passed only because of the test double's extra decode. That was wrong: with the double decode removed it still passes, because `ProcessAnyFileContent` decodes a second time and keeps the result only when that is what finds the file. So the server already decides by what is on disk rather than by guessing — the same approach this branch adopted elsewhere — and it is what lets a doubly-encoded legacy url and a real `photo%41.png` both work. Both tests now say so and point at each other.

Note for whoever picks this up: Devin re-ran on every push through the earlier rounds and its last complete pass (commit `cfb4a09d`) had **no unresolved bugs**, with all six of its threads resolved. The two commits since then could not be read back from Devin because the browser this session drives it with stopped responding. CI triggered the review as usual, so the results are on its review page — they simply were not folded in here.

src/BloomTests/PretendRequestInfo.cs
```diff
@@ -38,11 +38,17 @@ public PretendRequestInfo(
...
+
+            // Decode EXACTLY ONCE, and drop the query string first, because that is what the real
+            // RequestInfo.LocalPathWithoutQuery does. This used to decode twice -- once here via
+            // UnescapeFileNameForHttp and again via UrlDecode -- which made the tests a poor model
+            // of the server: a doubly-encoded url resolved here but not in production, and a file
+            // whose name genuinely contains "%41" resolved in production but not here. (BL-16669)
+            var queryStart = path.IndexOf("?", StringComparison.Ordinal);
+            var urlToDecode = queryStart == -1 ? path : path.Substring(0, queryStart);
             var pathWithoutLiteralPlusSigns = urlToDecode.Replace("+", "%2B");
             LocalPathWithoutQuery = System.Web.HttpUtility.UrlDecode(pathWithoutLiteralPlusSigns);
```
JohnThomson, line 58: [Devin] **Investigate**: Test double now decodes once, which may shift behavior of other server tests

`PretendRequestInfo` previously applied `UnescapeFileNameForHttp` and then `UrlDecode`, i.e. decoded twice, and did not strip the query before decoding. Making it decode exactly once (and split off the query first) makes it a better model of the real `RequestInfo`, but it changes the input every existing `BloomServerTests`/other test that constructs a url with escapes sees. In particular the double-encoded case in `HandleDoubleEncodedUrls` now depends on `ProcessAnyFileContent`'s conditional second decode rather than on the test double, so that test is now exercising production behavior it previously masked — good, but a reviewer should confirm the full server test suite is green.

JohnThomson: [Claude Opus 5 (1M context) during preflight] Confirmed green, and your reading of `HandleDoubleEncodedUrls` matches what I found.

**The full C# suite passes: 3076 passed, 0 failed, 13 skipped**, against this head with `origin/master` merged in. The server fixtures specifically — `BloomServerTests`, `ImageServerTests`, `RequestInfoTests` — are 61 passed, 0 failed. So nothing depended on the old double decode.

On `HandleDoubleEncodedUrls`: it is worth being blunt about this, because I had it wrong earlier in this PR and said so on the record. I had claimed that test passed *only* because the test double decoded twice, and that it was therefore asserting behavior production had dropped. Removing the double decode disproved that — it still passes, because `ProcessAnyFileContent` decodes a second time and keeps the result only when that is what finds the file. So the server was already deciding by what is on disk rather than by guessing, which is the same approach this branch adopted elsewhere, and it is precisely what lets a doubly-encoded legacy url and a file genuinely named `photo%41.png` both resolve.

That guard is now named in both tests, each pointing at the other, so the next person doesn't have to rediscover why two apparently contradictory cases can both be true.

src/BloomBrowserUI/bookEdit/toolbox/games/GameTool.tsx
```diff
@@ -1846,7 +1849,13 @@ export function playSound(
...
     soundType?: SoundType,
 ) {
-    let url = "audio/" + newSoundId;
+    // encodeURIComponent because newSoundId is a plain file name, not a URL: the sound
+    // attributes deliberately store the unencoded name (see the encoding conventions note on
+    // UrlPathString.cs), so anything in it that means something in a URL has to be escaped
+    // here, when we build one. Without this, a sound genuinely called "beep%41.mp3" is asked
+    // for as "beepA.mp3" and silently doesn't play. Safe to encode the whole thing because
+    // this is a bare file name -- no '/' of its own to preserve. (BL-16669)
+    let url = "audio/" + encodeURIComponent(newSoundId);
```
JohnThomson, line 1858: [Devin] **Investigate**: Sound URL is now encoded in the toolbox but not in the published player

`playSound` now escapes the stored name when building the audio URL, which fixes playback in the editor. The same attributes are read by Bloom Player when the book is published, and that side still concatenates the name unescaped, so a sound named with a literal '%' escape sequence will still fail there. The PR notes this as BL-16688; flagging so the follow-up isn't lost, since the two halves now behave differently for the same book.

JohnThomson: [Claude Opus 5 (1M context) during preflight] Correct, and deliberately so — tracked as [BL-16688](https://issues.bloomlibrary.org/youtrack/issue/BL-16688) (Bug, Bloom 6.5, Ready For Work, linked to BL-16669). Resolving on that basis.

Worth stating the asymmetry plainly, since you're right that the same book now behaves differently in the two places: the fix has to be on the URL-building side rather than the storing side. These names are stored unencoded on purpose — three places in C# use them directly as file-system paths, and one of those deletes any audio file not in the set it builds, so encoding them at rest would lose any sound whose name merely contains a space. That reasoning is written up on the `UrlPathString` class; the note now names Bloom Player as the remaining unfixed consumer and points at BL-16688, so the two halves can't drift apart silently.

The card carries the mechanism traced through the shipped bundle and repro steps, and is explicit that I did not verify it end-to-end in Bloom Player myself.

JohnThomson:
[Claude Opus 5 (1M context) during preflight] Consulted Devin on 2026-08-11 20:41 UTC up to commit 4a0b972d65c2ac213d8e7eabfea469da24d8065b — the final head of this run. This supersedes the caveat in my previous comment: the browser this session drives Devin through has been restored, so the commits it could not read then have now been reviewed.

**No unresolved Bugs.** Six Investigate flags and sixteen Informational items (not mirrored — low signal). Five of the six are repeats of threads already resolved above. The one new flag — that game sounds are now encoded in the editor but still not in Bloom Player — is correct and deliberate, tracked as BL-16688.

**All eight mirrored threads on this PR are resolved, each with a written outcome.** None are left open.

A local review of the commits Devin had not yet seen ran alongside this and found no behavioural defect, but did catch three comments that asserted things the code no longer did — including one I had introduced two commits earlier and one pre-existing error in `HandleDoubleEncodedUrls`. Those are fixed in `4a0b972d65`. Since the point of those comments is to save the next person from re-deriving which stored names are URL-encoded, a comment that lies is worth treating as a defect.

Full suite at this head: **C# 3076 passed / 0 failed**, **Vitest 674 passed / 0 failed**, typecheck and lint clean.

andrew-polk:
I don't really like the shape of this change. It feels wrong.
1. We wanted the UrlPathString mechanism to "just work". It seems like essentially breaks that.
2. This feels like a very heavy change to make based on something an agent found in code but which has never been experienced by a user, even a tester.

I'm open to discussion, but this doesn't feel worth it at all.

JohnThomson (commented):
On point 2, it's very easy to reproduce, and Colin did report it as part of the follow-up to BL-16658. I then asked the AI to make a separate issue; I don't remember exactly why, but it must have felt like a distinct problem. However, they are very close: both issues where we're not doing url encoding accurately.
Your point 1 was helpful, and I changed the PR so that improves this instead of making it worse. The common cases now indeed "just work": CreateFromUnencoded() does exactly that and NEVER does any decoding. For the handful of cases where we have a string that might have some encoding, I made a new method.
It's still a big change to solve a pretty rare problem, but at least it's now moving in the direction of making our code clearer as well as more correct.

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8179#-:-OzvUul81q7MDjqJ1MFA:b-wty2r)* status: 0 of 34 files reviewed, 8 unresolved discussions.

andrew-polk:
> On point 2, it's very easy to reproduce, and Colin did report it as part of the follow-up to BL-16658. I then asked the AI to make a separate issue; I don't remember exactly why, but it must have felt like a distinct problem. However, they are very close: both issues where we're not doing url encoding accurately.

Ok, thanks for clarifying that.

> Your point 1 was helpful, and I changed the PR so that improves this instead of making it worse. The common cases now indeed "just work": CreateFromUnencoded() does exactly that and NEVER does any decoding. For the handful of cases where we have a string that might have some encoding, I made a new method.  
It's still a big change to solve a pretty rare problem, but at least it's now moving in the direction of making our code clearer as well as more correct.

Great!

andrew-polk (commented):
> @andrew-polk reviewed 34 files and all commit messages, and resolved 8 discussions.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8179)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on JohnThomson).

andrew-polk (approved):
LGTM

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8179#-:-OzwONN46RbayI6VouGh:bb6ywmx)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on JohnThomson).

PR #8180, opened 2026-08-07
Keep the cursor in place when showing hidden characters (BL-16616)

andrew-polk:
Ctrl+Shift+Space (show hidden characters) made the cursor jump to the start of the text box, because `showInvisibles`/`hideInvisibles` rewrite the whole bloom-editable's innerHTML (wrapping each invisible character in a symbol span on keydown, unwrapping on keyup), and replacing innerHTML destroys the selection.

Save the caret as a character offset before each rewrite and restore it afterwards, using the existing `EditableDivUtils.getElementSelectionIndex` / `makeSelectionIn` pair — the same pattern `readerToolsModel` uses around its own innerHTML rewrites. Offsets stay valid because each invisible character is swapped one-for-one with a single symbol character. On blur, the selection has already left the editable, so the index is -1 and we don't steal the selection back.

Verified live in Bloom over CDP: with the fix the caret stays put through show/hide (both with and without invisible characters present); reverting to the pre-fix code reproduces the reported jump-to-start exactly.

Ref: https://issues.bloomlibrary.org/youtrack/issue/BL-16616

[Devin review](https://devinreview.com/BloomBooks/BloomDesktop/pull/8180)

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8180)

### DIFF
--- a/src/BloomBrowserUI/bookEdit/js/showInvisibles.ts
+++ b/src/BloomBrowserUI/bookEdit/js/showInvisibles.ts
@@ -1,4 +1,5 @@
 import bloomQtipUtils from "./bloomQtipUtils";
+import { EditableDivUtils } from "./editableDivUtils";
 import $ from "jquery";
 
 // For testing and debugging, functionality to replace invisible characters with symbols.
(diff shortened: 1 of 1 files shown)


andrew-polk:
[Claude Fable 5] Consulted Devin on 2026-08-07 ~20:20 UTC up to commit 20ad15c3c51de99335a33d367821bb09447e95d2. Result: no bugs, no Investigate flags. 3 Informational observations (selection collapse to caret on restore, the one-symbol-per-invisible-character offset invariant, and the blur-path guard) — all describe intended properties of the fix; none need action. Review: https://devinreview.com/BloomBooks/BloomDesktop/pull/8180

JohnThomson (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8180)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on andrew-polk).

PR #8181, opened 2026-08-10
Don't trigger long-press from shortcut keystrokes (BL-16616)

andrew-polk:
Follow-up to #8180 (keep the cursor in place when showing hidden characters). QA reported that Ctrl+Shift+Space now pops up the long-press special-characters panel.

The long-press keydown handler never checked modifier keys: holding Ctrl+Shift+Space repeats the Space keydown, which armed the long-press timer, and the timer offers alternates for whatever character precedes the caret. Before the caret fix, the destroyed selection made that lookup fail — the cursor-jump bug was the only thing keeping the panel hidden (the hole was already reachable by holding Ctrl+Space, for example).

Fix: bail out of the long-press keydown handler when Ctrl or the Windows/Cmd key is down, since such a keystroke is a shortcut and never types a character. Ctrl+Alt is exempt because Windows reports AltGr as Ctrl+Alt, and AltGr keystrokes genuinely type characters users may long-press.

Verified in a running Bloom over CDP: holding Ctrl+Shift+Space in a text box no longer shows the long-press panel and the caret stays put; holding a plain letter key still shows the panel.

Ref: https://issues.bloomlibrary.org/youtrack/issue/BL-16616

[Devin review](https://devinreview.com/BloomBooks/BloomDesktop/pull/8181)

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8181)

### DIFF
--- a/src/BloomBrowserUI/lib/long-press/jquery.longpress.js
+++ b/src/BloomBrowserUI/lib/long-press/jquery.longpress.js
@@ -196,6 +196,19 @@ import "./jquery.mousewheel.js";
     }
 
     function onKeyDown(e) {
+        // A keystroke with Ctrl or the Windows/Cmd key held down is a shortcut, not
+        // typing, so it never enters a character for us to offer alternates for.
+        // Without this check, holding a shortcut like Ctrl+Shift+Space (show hidden
+        // characters) popped up alternates for whatever character happened to precede
+        // the caret (BL-16616). Ctrl+Alt is exempt because Windows reports AltGr that
+        // way, and AltGr keystrokes do type characters.
+        // The exemption does mean a real Ctrl+Alt shortcut held down (e.g. Ctrl+Alt+2,
+        // heading 2) can still open the panel, as it always could. We decided to accept
+        // that rather than try to distinguish true AltGr via
+        // e.getModifierState("AltGraph"), which risks breaking long-press for genuine
+        // AltGr typists if some keyboard layout or IME path fails to report it.
+        if ((e.ctrlKey && !e.altKey) || e.metaKey) return;
+
         // See comment for BL-5215 in toolbox.ts
         window.top[isLongPressEvaluating] = true;
 


src/BloomBrowserUI/lib/long-press/jquery.longpress.js
```diff
@@ -196,6 +196,14 @@ import "./jquery.mousewheel.js";
     }
 
     function onKeyDown(e) {
+        // A keystroke with Ctrl or the Windows/Cmd key held down is a shortcut, not
+        // typing, so it never enters a character for us to offer alternates for.
+        // Without this check, holding a shortcut like Ctrl+Shift+Space (show hidden
+        // characters) popped up alternates for whatever character happened to precede
+        // the caret (BL-16616). Ctrl+Alt is exempt because Windows reports AltGr that
+        // way, and AltGr keystrokes do type characters.
+        if ((e.ctrlKey && !e.altKey) || e.metaKey) return;
```
andrew-polk, line 210: [Devin] **Investigate**: Ctrl+Alt exemption still leaves real Ctrl+Alt shortcuts able to arm long-press

Because Windows reports AltGr as Ctrl+Alt, the guard deliberately lets Ctrl+Alt through. That means genuine Ctrl+Alt+<key> shortcuts (which do not type a character) still fall through to the rest of the handler and can arm the long-press timer, reproducing the original BL-16616 symptom for that modifier combination. If any Ctrl+Alt shortcut exists in Bloom's editing surface, it will still pop the alternates panel when held. Worth confirming whether that combination is used anywhere in the edit page; distinguishing true AltGr from Ctrl+Alt usually requires checking `e.getModifierState("AltGraph")`.

andrew-polk: [Claude Fable 5] Assessed — leaving as is. This is accurate but describes pre-existing behavior, not something this PR introduced or worsened: Ctrl+Alt combos never destroyed the selection, so holding one (e.g. Ctrl+Alt+2, the heading shortcut) could arm long-press even before BL-16616's caret fix. It only shows the panel if the user *holds* the combo down, which isn't how formatting shortcuts get used.

Tightening the guard with `e.getModifierState("AltGraph")` would trade that unlikely cosmetic edge case for a real risk: if the AltGraph modifier state isn't reported for some layout/IME path (including Windows' left-Ctrl+Alt-as-AltGr substitute), long-press breaks for genuine AltGr typists — a core feature for Bloom's African and European keyboard users. The permissive exemption is the safer side of that trade.

andrew-polk:
[Claude Fable 5] Consulted Devin on 2026-08-10 ~17:50 UTC up to commit `f6ad62b86b9bbfb48353e859b01d9da347f1b811`. Result: 0 bugs, 1 Investigate flag (posted as an inline thread, assessed as pre-existing behavior not worth the AltGr risk, resolved), 2 Informational observations (both confirm the guard's early return is harmless — matching our own review). CI (pr-automation) green.

src/BloomBrowserUI/lib/long-press/jquery.longpress.js
andrew-polk: [Claude Fable 5] Andrew confirmed: leave as is. The decision is now recorded as a code comment next to the guard (commit `bd46578cd5`), explaining that we accept held Ctrl+Alt shortcuts opening the panel rather than risk breaking AltGr typing with a `getModifierState("AltGraph")` check.

src/BloomBrowserUI/lib/long-press/jquery.longpress.js
```diff
@@ -196,6 +196,19 @@ import "./jquery.mousewheel.js";
...
+        // Without this check, holding a shortcut like Ctrl+Shift+Space (show hidden
+        // characters) popped up alternates for whatever character happened to precede
+        // the caret (BL-16616). Ctrl+Alt is exempt because Windows reports AltGr that
+        // way, and AltGr keystrokes do type characters.
+        // The exemption does mean a real Ctrl+Alt shortcut held down (e.g. Ctrl+Alt+2,
+        // heading 2) can still open the panel, as it always could. We decided to accept
+        // that rather than try to distinguish true AltGr via
+        // e.getModifierState("AltGraph"), which risks breaking long-press for genuine
+        // AltGr typists if some keyboard layout or IME path fails to report it.
+        if ((e.ctrlKey && !e.altKey) || e.metaKey) return;
```
andrew-polk, line 210: [Devin] **Investigate**: Alt-only shortcuts still arm long-press

The guard covers Ctrl (without Alt) and Meta, but a plain Alt+key keystroke (no Ctrl) still falls through and can arm the long-press timer, the same class of problem BL-16616 describes. Alt-only combinations are common menu accelerators; if any reach the editable box while held, the alternates panel can still appear. Worth confirming whether Alt-only shortcuts ever reach this handler in Bloom's editing frame.

andrew-polk: [Claude Fable 5] Leaving as is per Andrew: same pre-existing family as the Ctrl+Alt case above — bare Alt+key combos fell through to long-press long before this PR, and no code change is wanted here.

JohnThomson (approved):
> @JohnThomson reviewed 1 file and all commit messages, and resolved 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8181)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on andrew-polk).

PR #8184, opened 2026-08-10
Fix Cancel Upload sometimes doing nothing (BL-16340)

andrew-polk:
Cancel's whole job is to set one boolean, `_progress.CancelRequested`, which the upload polls between stages. The publish screen greys out UPLOAD BOOK the instant the user clicks Cancel, and the only thing that ever brings it back is the `uploadCanceled` websocket event — which C# sends only if that flag is still true when the upload finishes. So anything that stops the flag being set in time leaves the upload running to completion **and** the button dead for good, which is exactly what was reported.

Three ways that could happen, all fixed here:

1. **The cancel request had to queue behind the very work it was trying to interrupt.** `libraryPublish/cancel` was registered to run on the UI thread and (by default) to take the global api sync lock — both of which a long upload can occupy for minutes (offscreen-browser thumbnailing on the UI thread, then `BloomPdfMaker.exe` pulling the book and its resources back through Bloom's own server). If the POST was stuck until after the upload's last cancellation check, the flag arrived too late to stop anything. It is now registered `(handleOnUiThread: false, requiresSync: false)`, like the other cancel endpoints (`progress/cancel`, `signLanguage/cancelImportVideo`), which exist to set a flag and nothing more.

2. **A cancel during the pre-upload handshake was thrown away.** The client shows Cancel from the moment the user commits, but two api round trips happen before `libraryPublish/upload` is posted, and `HandleUpload` opened by clearing `CancelRequested`. The flag is now cleared in `HandleCheckSubscriptionMatch` — the first call of every upload attempt — and `HandleUpload` declines to start if it is already set. `HandleCancel` reports the cancellation itself when no upload is running, so the user gets the same feedback in that window.

3. **The UI had no way back regardless of what C# decided.** `isCanceling` is now also cleared on `uploadSuccessful` and on an error in the progress box, so the user can never be left with no working button.

Also:

- A cancel that arrives after the server has committed the upload no longer reports "Upload was cancelled". If a book id came back, the book really is on BloomLibrary, and saying otherwise is a falsehood the user might act on.
- `WebProgressAdapter.CancelRequested` is backed by a `volatile` field; it is written and read from different threads with no lock between them.
- Bulk upload shares `_progress` but has no Cancel of its own, so it now clears the flag too; otherwise a leftover cancel could silently stop it before it started.

### Tests

Added `EndpointHandlerTests.RequestNotRequiringSync_IsServedWhileAnotherRequestHoldsTheSyncLock`, which proves the property fix 1 depends on. Verified it is not falsely passing: flipping the interrupting endpoint to `requiresSync: true` makes it fail.

### Note on verification

The bug itself is a race — the reporter could not reproduce it either — so it was **not** reproduced deterministically here. The fixes are reasoned from the code paths rather than confirmed against a failing repro. Fix 3 is the one that categorically rules out the reported symptom; fixes 1 and 2 address the causes that the code paths support.

Ref: https://issues.bloomlibrary.org/youtrack/issue/BL-16340

[Devin review](https://devinreview.com/BloomBooks/BloomDesktop/pull/8184)

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8184)

### DIFF
--- a/src/BloomBrowserUI/publish/LibraryPublish/LibraryPublishSteps.tsx
+++ b/src/BloomBrowserUI/publish/LibraryPublish/LibraryPublishSteps.tsx
@@ -294,9 +294,14 @@ export const LibraryPublishSteps: React.FunctionComponent<{
         );
     };
 
+    // True from the moment the user clicks Cancel until C# reports what became of the upload.
+    // While it is true, UPLOAD BOOK is greyed out, so every path that ends an upload has to
+    // clear it. Missing one leaves the user with no working button at all and no way back
+    // short of leaving the screen, which is how BL-16340 presented.
     const [isCanceling, setIsCanceling] = useState<boolean>(false);
     const handleUploadError = React.useCallback(() => {
         setIsUploading(false);
+        setIsCanceling(false);
     }, []);
 
     useSubscribeToWebSocketForEvent(
@@ -339,6 +344,10 @@ export const LibraryPublishSteps: React.FunctionComponent<{
         kWebSocketEventId_uploadSuccessful,
         (results) => {
             setIsUploading(false);
+            // The user may have asked to cancel and C# finished anyway -- either the cancel
+            // arrived after the upload was committed, or it was delayed getting to C# at all.
+            // Whatever the reason, the upload is over, so the Cancel state has to end with it.
+            setIsCanceling(false);
             setBookUrl(results.url);
             setIsUploadComplete(true);
         },
--- a/src/BloomBrowserUI/publish/LibraryPublish/LibraryPublishSteps.tsx
+++ b/src/BloomBrowserUI/publish/LibraryPublish/LibraryPublishSteps.tsx
@@ -248,6 +248,8 @@ export const LibraryPublishSteps: React.FunctionComponent<{
     function uploadOneBook() {
         setIsUploadComplete(false);
         setIsUploading(true);
+        cancelRequestedRef.current = false;
+        setHasAttemptBegun(false);
         // If either pre-upload request dies at the transport level we get no reply and no progress
         // message, so nothing else would ever clear isUploading. That used to leave only a stale
         // Cancel button, but now it would also keep the other publish tools disabled, so clear it
@@ -260,6 +262,10 @@ export const LibraryPublishSteps: React.FunctionComponent<{
                     // The API already sent an error message
                     return;
                 }
+                // C# has now marked the attempt as started, so a Cancel from here on cannot be
+                // wiped out by that same request arriving late. Only now do we let the user
+                // press Cancel. See the note on hasAttemptBegun.
+                setHasAttemptBegun(true);
                 get(
                     "libraryPublish/getUploadCollisionInfo?index=" +
                         conflictIndex,
@@ -268,6 +274,12 @@ export const LibraryPublishSteps: React.FunctionComponent<{
                             // The API already sent an error message
                             return;
                         }
+                        // The user cancelled while we were asking the server about an existing
+                        // copy. That reply is no longer wanted: showing the collision dialog now
+                        // would pop a dialog up over an attempt the user has abandoned, and
+                        // posting the upload would only be declined. Either way we say nothing
+                        // more -- C# has already reported the cancellation. (BL-16340)
+                        if (cancelRequestedRef.current) return;
                         if (result.data.shouldShow) {
                             setUploadCollisionInfo(result.data);
                             showUploadCollisionDialog();
@@ -306,8 +318,22 @@ export const LibraryPublishSteps: React.FunctionComponent<{
     // a cancel-pending upload was still running -- and pressing it would then clear the pending
     // cancel server-side and start a second upload alongside the first.
     const [isCanceling, setIsCanceling] = useState<boolean>(false);
+
+    // True once C# has acknowledged the start of this attempt (the subscription check has come
+    // back). Until then Cancel is shown but disabled, because a cancel sent before C# knows an
+    // attempt exists races with the request that begins it -- and if the cancel loses that race
+    // it is wiped out and the book uploads anyway. Waiting for the acknowledgement removes the
+    // race by construction rather than narrowing it. (BL-16340)
+    const [hasAttemptBegun, setHasAttemptBegun] = useState<boolean>(false);
+
+    // Whether the user has cancelled the attempt that is currently starting up. This is a ref,
+    // not state, because the pre-upload callbacks below close over their render's values and
+    // would not see a state update made after they were created.
+    const cancelRequestedRef = useRef(false);
+
     const handleUploadError = React.useCallback(() => {
         setIsUploading(false);
+        setHasAttemptBegun(false);
     }, []);
 
     useSubscribeToWebSocketForEvent(
@@ -315,6 +341,7 @@ export const LibraryPublishSteps: React.FunctionComponent<{
         kWebSocketEventId_uploadCanceled,
         () => {
             setIsCanceling(false);
+            setHasAttemptBegun(false);
         },
     );
 
@@ -354,6 +381,7 @@ export const LibraryPublishSteps: React.FunctionComponent<{
             // arrived after the upload was committed, or it was delayed getting to C# at all.
             // Whatever the reason, the upload is over, so the Cancel state has to end with it.
             setIsCanceling(false);
+            setHasAttemptBegun(false);
             setBookUrl(results.url);
             setIsUploadComplete(true);
         },
@@ -644,9 +672,10 @@ export const LibraryPublishSteps: React.FunctionComponent<{
                             )}
                             {isUploading ? (
                                 <BloomButton
-                                    enabled={!isCanceling}
+                                    enabled={!isCanceling && hasAttemptBegun}
                                     l10nKey={"Common.Cancel"}
                                     onClick={() => {
+                                        cancelRequestedRef.current = true;
                                         setIsCanceling(true);
                                         setIsUploading(false);
                                         post("libraryPublish/cancel");
@@ -762,6 +791,7 @@ export const LibraryPublishSteps: React.FunctionComponent<{
                 {...uploadCollisionInfo}
                 onCancel={() => {
                     setIsUploading(false);
+                    setHasAttemptBegun(false);
                 }}
                 conflictIndex={conflictIndex}
                 setConflictIndex={changeConflictIndex}
--- a/src/BloomBrowserUI/publish/LibraryPublish/LibraryPublishSteps.tsx
+++ b/src/BloomBrowserUI/publish/LibraryPublish/LibraryPublishSteps.tsx
@@ -297,11 +297,17 @@ export const LibraryPublishSteps: React.FunctionComponent<{
     // True from the moment the user clicks Cancel until C# reports what became of the upload.
     // While it is true, UPLOAD BOOK is greyed out, so every path that ends an upload has to
     // clear it. Missing one leaves the user with no working button at all and no way back
-    // short of leaving the screen, which is how BL-16340 presented.
+    // short of leaving the screen, which is how BL-16340 presented. The two things that clear
+    // it are the uploadCanceled event and uploadSuccessful, and C# now guarantees one of them
+    // on every ending of a cancelled upload.
+    //
+    // Deliberately NOT cleared here on an error line: this fires for ANY Error-kind progress
+    // message, not just a terminal one, so clearing it here would re-enable UPLOAD BOOK while
+    // a cancel-pending upload was still running -- and pressing it would then clear the pending
+    // cancel server-side and start a second upload alongside the first.
     const [isCanceling, setIsCanceling] = useState<boolean>(false);
     const handleUploadError = React.useCallback(() => {
         setIsUploading(false);
-        setIsCanceling(false);
     }, []);
 
     useSubscribeToWebSocketForEvent(
--- a/src/BloomExe/web/controllers/LibraryPublishApi.cs
+++ b/src/BloomExe/web/controllers/LibraryPublishApi.cs
@@ -289,12 +289,14 @@ namespace Bloom.web.controllers
             }
             if (alreadyCancelled)
             {
-                // The user cancelled during the pre-upload round trips, yet the client has still
-                // asked us to upload. That is a real path, not a stray request: the "this book is
-                // already on BloomLibrary" dialog can still be on screen after a cancel, and each
-                // of its buttons posts here. Say so rather than declining in silence, which would
-                // leave the user pressing a button that does nothing at all. (BL-16340)
-                ReportUploadCanceled();
+                // The user cancelled and this upload request is what is left of the attempt --
+                // most often one the client had already posted, which then queued behind the UI
+                // thread while the lock-free cancel overtook it. HandleCancel has therefore
+                // already said "Upload was cancelled"; saying it again would put two identical
+                // red lines in the log for a single click. Send the event only: it costs nothing
+                // (the client just clears the same state again) and guarantees the screen is
+                // released even if this request is the last thing to happen. (BL-16340)
+                ReportUploadCanceled(withMessage: false);
                 request.PostSucceeded();
                 return;
             }
@@ -611,10 +613,11 @@ namespace Bloom.web.controllers
             // "already on BloomLibrary" dialog, which can still be on screen after the user
             // cancelled; without this check we would give the book a brand-new instance id on
             // disk and then decline to upload it, leaving its identity permanently changed for
-            // an upload that never happened. (BL-16340)
+            // an upload that never happened. Event only, for the same reason as in HandleUpload:
+            // the cancellation has already been announced. (BL-16340)
             if (_progress.CancelRequested)
             {
-                ReportUploadCanceled();
+                ReportUploadCanceled(withMessage: false);
                 request.PostSucceeded();
                 return;
             }
--- a/src/BloomExe/web/controllers/LibraryPublishApi.cs
+++ b/src/BloomExe/web/controllers/LibraryPublishApi.cs
@@ -178,23 +178,27 @@ namespace Bloom.web.controllers
         /// </summary>
         private void HandleCancel(ApiRequest request)
         {
-            // If the user clicked Cancel before the upload itself got under way, no upload is
-            // running to notice the flag and report on it. That window is real and not small:
-            // the client shows Cancel from the moment the user commits, but two API round trips
-            // (the subscription check and the "existing copy on server" query) happen before
-            // libraryPublish/upload even arrives. Nothing else would ever tell the client the
-            // cancel took effect -- and the client only ever leaves its Cancel state on that
-            // report -- so make the report here. HandleUpload takes the same lock and declines
-            // to start. (BL-16340)
+            // Whether we announce the cancellation ourselves depends on how far the attempt has
+            // got. Only in StartingUp is there nobody else to do it: the client shows Cancel from
+            // the moment the user commits, but two API round trips (the subscription check and
+            // the "existing copy on server" query) happen before libraryPublish/upload arrives,
+            // so during those there is no upload running to notice the flag -- and the screen
+            // only ever leaves its Cancel state on such a report. While Uploading, UploadBookAsync
+            // reports whatever becomes of it. When Idle we must stay SILENT: the attempt is
+            // already over, and announcing a cancellation then would contradict the outcome the
+            // user was just given -- telling someone their upload was cancelled seconds after
+            // being told it succeeded is worse than saying nothing. (BL-16340)
             bool nobodyElseWillReportIt;
             lock (_uploadStateLock)
             {
                 _progress.CancelRequested = true;
-                nobodyElseWillReportIt = !_uploadInProgress;
+                nobodyElseWillReportIt = _attemptState == UploadAttemptState.StartingUp;
+                if (nobodyElseWillReportIt)
+                    _attemptState = UploadAttemptState.Idle; // we are about to report it
             }
 
             // Outside the lock: this sends a websocket message, and the lock exists only to make
-            // the flag pair above indivisible.
+            // the state read/write above indivisible.
             if (nobodyElseWillReportIt)
                 ReportUploadCanceled();
 
@@ -227,15 +231,31 @@ namespace Bloom.web.controllers
 
         private bool _changeUploader = false;
 
-        // True from the moment we take charge of an upload until we have reported its outcome.
-        // HandleCancel uses it to work out whether anyone else is going to report the cancel.
-        private bool _uploadInProgress;
+        /// <summary>
+        /// How far the current upload attempt has got, as C# sees it. It exists so HandleCancel
+        /// can tell whether anyone else is going to report a cancellation -- and, just as
+        /// importantly, whether the attempt is already over and it should say nothing at all.
+        /// </summary>
+        private enum UploadAttemptState
+        {
+            /// No attempt is under way, or the last one's outcome has already been reported.
+            Idle,
 
-        // Makes "is a cancel already pending?" and "an upload is now running" one indivisible
-        // step, shared with HandleCancel. Without it the two could interleave such that
-        // HandleCancel reported the cancel AND the upload started anyway, so the cancel got
-        // reported twice. Only ever held across these two field reads/writes -- never across
-        // the upload, or a websocket send.
+            /// The client has committed to an upload and is making the pre-upload round trips,
+            /// but libraryPublish/upload has not arrived yet.
+            StartingUp,
+
+            /// An upload is running; UploadBookAsync will report whatever becomes of it.
+            Uploading,
+        }
+
+        private UploadAttemptState _attemptState = UploadAttemptState.Idle;
+
+        // Guards _attemptState together with _progress.CancelRequested, so that reading one and
+        // writing the other is a single indivisible step. Without it the two could interleave
+        // such that HandleCancel reported the cancel AND the upload started anyway, so the
+        // cancel got reported twice. Only ever held across those field reads/writes -- never
+        // across the upload, the HTTP response, or a websocket send.
         private readonly object _uploadStateLock = new object();
 
         private async Task HandleUpload(ApiRequest request, bool changeUploader)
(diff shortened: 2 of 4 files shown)


src/BloomBrowserUI/publish/LibraryPublish/LibraryPublishSteps.tsx
```diff
@@ -294,6 +294,17 @@ export const LibraryPublishSteps: React.FunctionComponent<{
...
+    // short of leaving the screen, which is how BL-16340 presented. The two things that clear
+    // it are the uploadCanceled event and uploadSuccessful, and C# now guarantees one of them
+    // on every ending of a cancelled upload.
+    //
+    // Deliberately NOT cleared here on an error line: this fires for ANY Error-kind progress
+    // message, not just a terminal one, so clearing it here would re-enable UPLOAD BOOK while
+    // a cancel-pending upload was still running -- and pressing it would then clear the pending
+    // cancel server-side and start a second upload alongside the first.
     const [isCanceling, setIsCanceling] = useState<boolean>(false);
     const handleUploadError = React.useCallback(() => {
```
andrew-polk, line 334: [Devin] **Investigate**: Any error line now also re-enables UPLOAD BOOK while a cancel-pending upload may still be running

`handleUploadError` is wired to the progress box's `onGotErrorMessage`, so it fires for *any* Error-kind progress line, not only for terminal failures. Combined with the `setIsCanceling(false)` that the first commit added there, an error line emitted by an upload that is still running after the user pressed Cancel would re-enable UPLOAD BOOK. If the user then pressed it, `uploadOneBook` calls `libraryPublish/checkSubscriptionMatch`, which clears `_progress.CancelRequested` — defeating the pending cancel — and since `libraryPublish/upload` is registered `requiresSync: false`, a second `UploadBookAsync` could run concurrently with the first.

Devin noted it could not find a non-terminal `WriteError` in the upload path, so the window looked theoretical — but it was the safety property the old code got from leaving `isCanceling` set.

src/BloomExe/web/controllers/LibraryPublishApi.cs
```diff
@@ -181,14 +227,41 @@ private async Task HandleUploadWithNewUploader(ApiRequest request)
...
+            // throw that cancel away and upload the book anyway (BL-16340). It is cleared in
+            // HandleCheckSubscriptionMatch, which the client calls at the start of every attempt.
+            bool alreadyCancelled;
+            lock (_uploadStateLock)
+            {
+                alreadyCancelled = _progress.CancelRequested;
+                if (!alreadyCancelled)
+                    _uploadInProgress = true;
+            }
+            if (alreadyCancelled)
```
andrew-polk, line 290: [Devin] **Investigate**: Collision dialog can still appear after a cancel, and its Upload button then does nothing visible

The Cancel button is shown from the moment `uploadOneBook` runs — i.e. during the `checkSubscriptionMatch` / `getUploadCollisionInfo` round trips. Nothing in those callbacks checks whether a cancel happened in the meantime, so a cancel issued while "Checking for existing copy on server..." is in flight can be followed by the "this book is already on BloomLibrary" dialog popping up anyway.

If the user then presses Upload in that dialog (`uploadCollisionDlg.tsx` posts `libraryPublish/upload` / `uploadWithNewUploader` / `uploadAfterChangingBookId`), the new early-return here declines silently — the book is not uploaded, and the only feedback is the earlier "Upload was cancelled" line.

Before this PR the flag was cleared in `HandleUpload`, so that click would have uploaded. Worth deciding whether the client should suppress the dialog once a cancel has been requested, or whether pressing Upload in it should count as starting a fresh attempt.

This one is being referred to the developer rather than decided here, since either answer is defensible and it changes what the user sees.

src/BloomBrowserUI/publish/LibraryPublish/LibraryPublishSteps.tsx
andrew-polk: [Claude Opus 5 (1M context) from Andrew's machine during preflight] ✅ Fixed in `8fdbeb6443`.

Agreed — this was a real narrowing of the safety property, and worth closing even though the window looked theoretical.

`setIsCanceling(false)` is gone from `handleUploadError`. It was only ever needed for one case: a cancelled upload ending in the `"quiet"` result, where C# sent no event at all and the screen escaped its Cancel state solely because the pre-upload checker happened to emit an error line. Rather than keep depending on that, `UploadBookAsync` now sends `uploadCanceled` on **every** cancelled ending, including `"quiet"` (without a second message, since one has already been given).

So the Cancel state is now cleared only by `uploadCanceled` or `uploadSuccessful` — both of which C# guarantees — and a mere error line can no longer re-enable UPLOAD BOOK underneath a running upload.

src/BloomExe/web/controllers/LibraryPublishApi.cs
```diff
@@ -163,12 +171,54 @@ private void HandleUseSandbox(ApiRequest request)
...
+            // only ever leaves its Cancel state on such a report. While Uploading, UploadBookAsync
+            // reports whatever becomes of it. When Idle we must stay SILENT: the attempt is
+            // already over, and announcing a cancellation then would contradict the outcome the
+            // user was just given -- telling someone their upload was cancelled seconds after
+            // being told it succeeded is worse than saying nothing. (BL-16340)
+            bool nobodyElseWillReportIt;
+            lock (_uploadStateLock)
+            {
+                _progress.CancelRequested = true;
+                nobodyElseWillReportIt = _attemptState == UploadAttemptState.StartingUp;
```
andrew-polk: [Devin] **Bug**: A cancel clicked just as an upload finishes wrongly tells the user the upload was cancelled

When no upload was running, the cancel request always announced "Upload was cancelled" — even when the reason no upload was running is that it had *just succeeded*. `HandleCancel` decided purely from an "is an upload in progress" flag, and that flag was cleared in `HandleUpload`'s `finally`, which runs **after** `UploadBookAsync` has already sent `uploadSuccessful`. A cancel landing in that window printed a red "Upload was cancelled" line and sent the `uploadCanceled` event seconds after the user had been told the book was on BloomLibrary.

Impact: someone who clicks Cancel a moment too late is told the opposite of what happened, and might reasonably conclude the book is missing and upload it again. It is the same falsehood the PR already guards against inside `UploadBookAsync` — that guard just did not cover this path.

src/BloomExe/web/controllers/LibraryPublishApi.cs
```diff
@@ -436,6 +595,18 @@ private void HandleCheckSubscriptionMatch(ApiRequest request)
 
         private async Task HandleUploadAfterChangingBookId(ApiRequest request)
         {
+            // Check for a pending cancel BEFORE changing anything. This is reached from the
```
andrew-polk, line 612: [Devin] **Bug**: After cancelling, the "book already on server" dialog still appears and its Upload button does nothing — and one of its buttons changed the book's identity anyway

An upload request arriving after the user cancelled was dropped silently — no message, nothing on screen — so pressing Upload in the "existing copy" dialog appeared to do nothing at all.

Worse, one route mutated the book first. `HandleUploadAfterChangingBookId` called `Model.ChangeBookInstanceId` **before** delegating to `HandleUpload`, which then declined because a cancel was pending. The book got a brand-new instance guid on disk and nothing was uploaded — a permanent change to the book's identity in exchange for an upload that never happened.

Sequence: the client shows Cancel from the moment the user commits, while `checkSubscriptionMatch` and `getUploadCollisionInfo` are still in flight. A cancel during that window is reported, but the pending collision-info reply still arrives and the client opens the collision dialog regardless — it has no notion the attempt was abandoned. Every button in that dialog posts an upload command.

src/BloomExe/web/controllers/LibraryPublishApi.cs
andrew-polk: [Claude Opus 5 (1M context) from Andrew's machine during preflight] ✅ Fixed in `e17c6e5b54`.

Agreed, and thank you — this was a real regression and exactly the falsehood the PR set out to avoid.

The two booleans are replaced by an explicit three-state `UploadAttemptState`:

- **StartingUp** — the client has committed but `libraryPublish/upload` hasn't arrived. This is the only state in which `HandleCancel` announces the cancellation itself, because nothing else is running that could notice it.
- **Uploading** — `UploadBookAsync` will report whatever becomes of it.
- **Idle** — no attempt, or the last one's outcome has already been reported. A cancel arriving now says **nothing at all**, which is what closes this bug.

Follow-up in `2fe9684652` closes the related window you flagged separately (a cancel landing after `UploadBookAsync` had decided the outcome, where a failure ending emits only an error line): `UploadBookAsync` now returns whether it sent an outcome *event*, and `HandleUpload`'s `finally` sends `uploadCanceled` if a cancel is pending and none went out. So every cancelled attempt now ends with an outcome event, whichever route it took.

src/BloomExe/web/controllers/LibraryPublishApi.cs
andrew-polk: [Claude Opus 5 (1M context) from Andrew's machine during preflight] ✅ Partly fixed in `e17c6e5b54`; one piece deliberately left for the developer.

**Fixed — the book is no longer mutated for an upload that won't happen.** `HandleUploadAfterChangingBookId` now checks for a pending cancel *before* touching anything, so it can no longer hand the book a new instance guid and then decline to upload it. This was the serious half: the other symptoms are confusing, but this one left a permanent change on disk.

**Fixed — the silent decline.** That path now reports "Upload was cancelled" instead of returning without a word, so the dialog's buttons no longer look simply dead.

**Left for the developer — whether the dialog should appear at all after a cancel.** Suppressing it changes what the user sees, and there is a real choice between "a cancel abandons the attempt, so nothing more should pop up" and "pressing Upload in the dialog starts a fresh attempt". That is on the preflight decision report rather than being decided here, so this thread stays open until it comes back.

src/BloomExe/web/controllers/LibraryPublishApi.cs
```diff
@@ -181,26 +231,101 @@ private async Task HandleUploadWithNewUploader(ApiRequest request)
...
+                    _attemptState = UploadAttemptState.Idle;
+
+                    // A cancel can land after UploadBookAsync has already decided how this
+                    // ended. If the upload then finished in a way that emits only an error line
+                    // -- a failure or an exception -- nothing would ever release the screen from
+                    // its Cancel state, and UPLOAD BOOK would stay greyed out for good: the
+                    // original BL-16340 symptom, just through a narrower window. So make the end
+                    // of every attempt reconcile. Sending the event twice is harmless (the
+                    // client just clears the same state again); not sending it is not.
+                    mustStillReleaseTheScreen = _progress.CancelRequested && !toldClientTheOutcome;
```
andrew-polk, line 328: [Devin] **Investigate**: Residual window where a cancelled upload reports neither success nor cancellation

If the cancel POST landed *after* `UploadBookAsync` evaluated its "was this cancelled?" guard but *before* the attempt finished unwinding, `HandleCancel` stayed silent (it could see an upload in progress) while the upload took a failure branch that emits only an error line. Since the client deliberately does not act on error lines when deciding whether UPLOAD BOOK may be pressed again, the button stayed greyed out — the original BL-16340 symptom, through a much smaller window. The same applied if `UploadBookAsync` itself threw.

andrew-polk: [Claude Opus 5 (1M context) from Andrew's machine during preflight] ✅ Fixed in `2fe9684652`.

Good catch — this one survived the previous round's fix, and the analysis is exactly right: the state machine added in `e17c6e5b54` keeps `HandleCancel` silent during `Uploading`, which is correct, but nothing then covered a failure ending that emits only an error line.

`UploadBookAsync` now returns whether it sent the client an outcome **event** (`uploadSuccessful` or `uploadCanceled`) rather than only a progress message, and `HandleUpload`'s `finally` reconciles: if a cancel is pending and no event went out, it sends `uploadCanceled` (without a message — the failure has already been reported; the point is releasing the screen, not saying it twice).

The invariant that establishes: **every attempt the user cancelled now ends with an outcome event, whatever route it took.** Sending one twice is harmless, since the client just clears the same state again; sending none was the bug.

src/BloomExe/web/controllers/LibraryPublishApi.cs
```diff
@@ -415,6 +564,16 @@ private void HandleGetUploadCollisionInfo(ApiRequest request)
...
         private void HandleCheckSubscriptionMatch(ApiRequest request)
         {
+            // This is the first thing the client asks for once the user commits to an upload, so
+            // as far as C# is concerned it is where a new attempt begins -- and therefore where
+            // any cancellation left over from an earlier attempt is cleared. Deliberately not in
+            // HandleUpload, which arrives too late to clear it safely; see the note there.
+            lock (_uploadStateLock)
+            {
+                _progress.CancelRequested = false;
+                _attemptState = UploadAttemptState.StartingUp;
```
andrew-polk, line 588: [Devin] **Investigate**: Clearing the cancel flag in checkSubscriptionMatch depends on the UI never allowing a second attempt during a live upload

Moving the `CancelRequested = false` reset into `HandleCheckSubscriptionMatch` makes correctness depend on the client never issuing that request while a previous upload is still running in C#. Today that holds because UPLOAD BOOK is hidden while `isUploading` and greyed while `isCanceling`. But those are plain component state: if the publish screen is unmounted and remounted (tab switch/reload) while a cancel-pending upload is still running server-side, the state resets, the user can press UPLOAD BOOK, and this handler would clear the pending cancel and start a second upload alongside the first. `SetParentControlsState(false)` disables the workspace tabs during the upload, which mitigates but does not obviously eliminate this.

andrew-polk: [Claude Opus 5 (1M context) from Andrew's machine during preflight] Assessed as pre-existing and not made worse by this PR — resolving, with the reasoning recorded here.

The remount scenario is real, but the hazard it leads to predates this branch. If the publish screen is remounted while an upload is still running server-side, the component state resets and the user can press UPLOAD BOOK — and *that* starts a second concurrent `UploadBookAsync` regardless of the cancel flag, because `libraryPublish/upload` has always been `requiresSync: false` and nothing has ever refused a second attempt. Two concurrent uploads was already the outcome before this PR.

What this PR changes is only *where* the pending cancel gets cleared: it used to be cleared at the top of `HandleUpload`, so the same remount-and-click sequence cleared it just as effectively. So the cancel-clearing is neither newly introduced nor newly reachable here.

Guarding it properly means refusing a second attempt while one is running — a real improvement, but a separate concern from BL-16340 and one that deserves its own decision about what the UI should do instead (queue? refuse? tell the user?). Out of scope for this fix rather than dismissed; worth a card if it's judged worth closing.

src/BloomExe/web/controllers/LibraryPublishApi.cs
andrew-polk: [Claude Opus 5 (1M context) from Andrew's machine during preflight] Superseded — resolving this one to keep a single live thread.

Devin's re-review raised this same issue again, at Bug severity and with an additional consequence I had not appreciated: the `uploadAfterChangingBookId` route mutates the book's identity on disk *before* the cancel check, so pressing that button after a cancel left the book permanently altered with nothing uploaded.

That part is fixed in `e17c6e5b54`, and the outstanding question — whether the dialog should appear at all once a cancel has been issued — is tracked on the newer thread so the discussion stays in one place: https://github.com/BloomBooks/BloomDesktop/pull/8184#discussion_r3753594558

src/BloomExe/web/controllers/LibraryPublishApi.cs
```diff
@@ -181,26 +231,101 @@ private async Task HandleUploadWithNewUploader(ApiRequest request)
...
+                    _attemptState = UploadAttemptState.Uploading;
+            }
+            if (alreadyCancelled)
+            {
+                // The user cancelled during the pre-upload round trips, yet the client has still
+                // asked us to upload. That is a real path, not a stray request: the "this book is
+                // already on BloomLibrary" dialog can still be on screen after a cancel, and each
+                // of its buttons posts here. Say so rather than declining in silence, which would
+                // leave the user pressing a button that does nothing at all. (BL-16340)
+                ReportUploadCanceled();
```
andrew-polk: [Devin] **Bug**: Cancelling an upload before it starts shows the "Upload was cancelled" notice twice

In exactly the situation this PR targets — cancelling in the moments before the upload begins — the progress log ends up with two identical red "Upload was cancelled" lines for a single click, which reads as though something went wrong twice.

`HandleCancel` announces the cancellation while the attempt is in `StartingUp`. The client's in-flight `getUploadCollisionInfo` GET is still outstanding; when it returns with `shouldShow` false the client posts `libraryPublish/upload` anyway, and the early-return here calls `ReportUploadCanceled()` a second time with `withMessage` defaulting to true. (The duplicate *event* is harmless — the client just clears the same state again — it is only the duplicate message that shows.)

Introduced by `e17c6e5b54`, which added the message on this path deliberately so that pressing Upload in the collision dialog would not be silent. Those two goals conflict, which is why this is going to the developer together with the collision-dialog decision rather than being patched in isolation: the fix is either `withMessage: false` here, or tracking that the cancel has already been announced for this attempt.

**Not a regression against master** — before this PR a cancel in this window was silently discarded and the book uploaded regardless.

src/BloomExe/web/controllers/LibraryPublishApi.cs
```diff
@@ -415,6 +564,16 @@ private void HandleGetUploadCollisionInfo(ApiRequest request)
 
         private void HandleCheckSubscriptionMatch(ApiRequest request)
         {
+            // This is the first thing the client asks for once the user commits to an upload, so
+            // as far as C# is concerned it is where a new attempt begins -- and therefore where
+            // any cancellation left over from an earlier attempt is cleared. Deliberately not in
+            // HandleUpload, which arrives too late to clear it safely; see the note there.
+            lock (_uploadStateLock)
+            {
+                _progress.CancelRequested = false;
```
andrew-polk, line 587: [Devin] **Bug**: Pressing Cancel in the very first instant can still be ignored and the book uploaded anyway

A cancel that reaches the server *before* this start-of-attempt reset runs is thrown away by the reset, so the book uploads despite the user cancelling.

The client shows the Cancel button before `checkSubscriptionMatch` has been answered. That endpoint is registered `handleOnUiThread: true, requiresSync: true`, so it can queue behind the UI thread or the api sync lock — while `libraryPublish/cancel` is now deliberately `(false, false)` and runs immediately on a worker thread. So the cancel really can overtake it. `HandleCancel` then sets the flag but, seeing `_attemptState == Idle` from the previous attempt, deliberately says nothing; this reset then clears the flag and the upload proceeds.

**Not a regression against master, and much narrower than what it replaced.** Before this PR the reset lived at the top of `HandleUpload`, so a cancel anywhere in the *entire* pre-upload handshake was discarded — a window of two full round trips. What remains is only a cancel that beats this one request's handling.

Closing it properly means making the attempt's start unambiguous — e.g. the client not offering Cancel until the attempt has actually begun server-side, or an explicit begin-attempt call — which changes when the Cancel button appears. That is a design choice, so it is going to the developer rather than being decided here.

src/BloomExe/web/controllers/LibraryPublishApi.cs
andrew-polk: [Devin] **Investigate**: A cancel that arrives when no attempt is running leaves the screen greyed out

`HandleCancel` stays silent when `_attemptState == Idle`, so nothing clears the client's Cancel state in that case. The rendering guard makes it hard to hit — the Cancel button only exists while an upload is showing — but there is a window: an upload that ends between the user's mouse-down and the POST reaching C#. In the success case the client's own `setIsCanceling(false)` may run *before* the click sets it true, leaving UPLOAD BOOK disabled with nothing left to re-enable it.

Much narrower than the original BL-16340 window, but the "silent when Idle" choice does trade a rare stuck button for avoiding a contradictory message. A likely fix is to send the *event* without the message when Idle — releasing the screen while claiming nothing about what happened — which is cheap, but it interacts with the two open decisions above, so it is going to the developer with them rather than being applied on its own.

andrew-polk:
[Claude Opus 5 (1M context) from Andrew's machine during preflight] Consulted Devin on 2026-08-10, up to commit `2fe96846528ee13012d93b0168c850e03d710df5`.

Devin ran four times across this branch's commits and its findings drove three of the five commits here. Each finding has its own thread; this is just the tally.

**Acted on and resolved**

- An error line could re-enable UPLOAD BOOK while a cancel-pending upload was still running — fixed in `8fdbeb6443`.
- A cancel clicked just as an upload finished announced the opposite of what happened — fixed in `e17c6e5b54`.
- The collision dialog's "change the book id" button changed the book's identity on disk and then uploaded nothing — fixed in `e17c6e5b54`.
- A cancelled upload that ended in failure released nothing, leaving the button dead — fixed in `2fe9684652`.
- A cancel-flag reset depending on the UI never allowing a second attempt — assessed as pre-existing and not made worse; reasoning on its thread.

**Open, waiting on the developer** (three threads)

- Whether the "already on BloomLibrary" dialog should appear at all after a cancel.
- Two identical "Upload was cancelled" lines in the pre-upload window.
- A cancel that overtakes the start-of-attempt reset, and the related "silent when Idle" trade-off.

These three are interrelated — they are all about what the user should see in the moments before an upload really begins — so they are going back as one decision rather than being patched piecemeal. None is a regression against master.

The bot gauntlet stopped here at its four-cycle cap: each Devin pass was finding issues in code written in the previous pass, which is the point to get a human's eyes on it.

Also seen and not mirrored: nine informational items (low signal). Two of them were acted on anyway — the "quiet" ending that sent no event, and the duplicate-report race now closed by a lock — and the rest were assessed, including confirming that moving the cancel handler off the UI thread is safe because the upload already does the same work from a background thread.

CI (`pr-automation`) passed on every pushed commit. CodeRabbit is configured on this repo but has `auto_review` disabled, so it did not run.

src/BloomExe/web/controllers/LibraryPublishApi.cs
andrew-polk: [Claude Opus 5 (1M context) from Andrew's machine during preflight] ✅ Fixed in `cb14caa890` — the developer chose to suppress the dialog.

The client now records that a cancel was requested for the current attempt, and drops the "is there an existing copy?" reply if one arrives afterwards. So the dialog no longer appears over an attempt the user has abandoned, and no upload is posted for `HandleUpload` to decline.

That also removes the silent-button symptom entirely rather than papering over it: there is no longer a dead button to press, because the dialog isn't there.

(The damaging half — `uploadAfterChangingBookId` giving the book a new instance id and then uploading nothing — was already fixed in `e17c6e5b54`.)

src/BloomExe/web/controllers/LibraryPublishApi.cs
andrew-polk: [Claude Opus 5 (1M context) from Andrew's machine during preflight] ✅ Fixed in `cb14caa890`.

The developer chose to fix this together with the two related findings rather than patch each one, since they pull against each other.

The client now waits for C# to acknowledge the attempt — the subscription check coming back — before it lets Cancel be pressed. Cancel is visible immediately but disabled until then. A cancel therefore cannot be sent before C# knows an attempt exists, so it can no longer be wiped out by that same request arriving late. The race is removed by construction rather than narrowed.

src/BloomExe/web/controllers/LibraryPublishApi.cs
andrew-polk: [Claude Opus 5 (1M context) from Andrew's machine during preflight] ✅ Fixed in `cb14caa890`.

Fixed at the source rather than by suppressing the second message. Now that the client drops the collision-info reply when the user has cancelled, it no longer posts an upload for `HandleUpload` to decline — so only one report is ever made and there is no second line to suppress.

src/BloomExe/web/controllers/LibraryPublishApi.cs
andrew-polk: [Claude Opus 5 (1M context) from Andrew's machine during preflight] ✅ Fixed in `cb14caa890`, along the lines you suggested.

A cancel arriving when C# has nothing running now sends the release event **without** the message. That frees the screen if it was still showing Cancel, while not claiming an upload was cancelled when it may in fact have just succeeded.

This is safe to do now in a way it wouldn't have been before: with the attempt's start made explicit, reaching the idle case no longer means "a real cancel got lost", so releasing the screen there cannot mask a genuine cancellation.

src/BloomExe/web/controllers/LibraryPublishApi.cs
andrew-polk: [Claude Opus 5 (1M context) from Andrew's machine during preflight] Correction, and now actually fixed in `4ac067f509`.

My previous reply here claimed suppressing the collision dialog had removed this at the source. That was wrong — it only covered the common path. Devin still reported it against `2fe96846`, and re-checking the code it was right:

if the collision check comes back clean, the client posts the upload **immediately**, and the user can click Cancel after that. `libraryPublish/upload` waits on the UI thread while `libraryPublish/cancel` is deliberately lock-free, so the cancel overtakes the queued upload — `HandleCancel` announces the cancellation, then the upload request arrives and announces it again. Two identical red lines for one click, exactly as described.

Both declining paths (`HandleUpload` and `HandleUploadAfterChangingBookId`) now send the release event **without** a second message. The event still goes out, so the screen is released even if one of these is the last thing to happen; only the duplicate wording is gone.

andrew-polk:
[Claude Opus 5 (1M context) from Andrew's machine during preflight] Consulted Devin on 2026-08-10, up to commit `4ac067f509`.

**Re-review clean — every finding is now resolved, and Devin confirms it.** All 4 bugs and all 5 Investigate flags raised across six passes are marked resolved by Devin at this commit; 12 informational items were seen and not mirrored (two of them acted on anyway).

The developer answered the earlier decision report, and both open questions were implemented:

- **The collision dialog after a cancel** — suppressed. The client drops the "is there an existing copy?" reply if the user cancelled while it was in flight, so no dialog appears over an abandoned attempt and no upload is posted to be declined.
- **The first moments of an upload** — fixed as one change rather than three patches. The client now waits for C# to acknowledge the attempt before Cancel becomes pressable (visible but disabled until then), which removes the start-of-attempt race by construction rather than narrowing it.

One correction worth recording: I claimed on the duplicate-message thread that suppressing the dialog had removed that problem at the source. It had not — if the collision check comes back clean the client posts the upload immediately, and a later cancel overtakes it because `libraryPublish/cancel` is lock-free while `libraryPublish/upload` waits on the UI thread. Devin re-reported it, was right, and `4ac067f509` fixes it properly: both declining paths send the release event without a second message.

CI (`pr-automation`) passed on every pushed commit. CodeRabbit is configured on this repo but has `auto_review` disabled, so it did not run.

PR #8185, opened 2026-08-10
Lock workspace tabs while a newly selected book loads (BL-15971)

andrew-polk:
Selecting a different book in the Collections tab did not lock the workspace UI while the new book was loaded (which can take several seconds for a heavy book, mostly in `BringBookUpToDate`). During that window the Edit tab and the "Edit this book" button stayed clickable, and entering the Edit tab could briefly show the previously edited book.

Three changes:

- **CollectionApi**: the `selected-book` and `selectAndEditBook` handlers now call `WorkspaceView.SetTabsEnabled(false)` while loading/selecting the book (re-enabled in `finally`) — the same mechanism used by Edit-tab saves, BloomLibrary uploads, and RAB builds (BL-16654).
- **CollectionsTabBookPane**: the "Edit this book"/"Make a book" button now honors the existing `navigationLocked` flag, so it greys out in step with the main tabs.
- **EditingView.OnVisibleChanged**: when entering the Edit tab after the book changed, call `ClearOutDisplay()` (which also blanks the main content browser) instead of only clearing the page list, so the old book's page is no longer visible while the new book navigates.

Ref: https://issues.bloomlibrary.org/youtrack/issue/BL-15971

[Devin review](https://devinreview.com/BloomBooks/BloomDesktop/pull/8185)

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8185)

### DIFF
--- a/src/BloomExe/web/controllers/CollectionApi.cs
+++ b/src/BloomExe/web/controllers/CollectionApi.cs
@@ -171,6 +171,11 @@ namespace Bloom.web.controllers
                             // without actually loading all the files. We need the title for the Performance Measurement and later
                             // we'll use the BookInfo object to get the fully updated book.
                             BookInfo newBookInfo = null;
+                            // Like other long-running operations (Edit-tab saves, uploads), lock the
+                            // workspace tabs while we load the newly selected book. Bringing a heavy
+                            // book up to date can take several seconds, and clicking Edit during that
+                            // window could show the previous book (BL-15971).
+                            _collectionTabView.WorkspaceView?.SetTabsEnabled(false);
                             try
                             {
                                 newBookInfo = GetBookInfoFromPost(request);
@@ -231,6 +236,10 @@ namespace Bloom.web.controllers
                                 }
                                 throw;
                             }
+                            finally
+                            {
+                                _collectionTabView.WorkspaceView?.SetTabsEnabled(true);
+                            }
 
                             request.PostSucceeded();
                             break;
@@ -243,18 +252,28 @@ namespace Bloom.web.controllers
                 kApiUrlPart + "selectAndEditBook",
                 request =>
                 {
-                    var book = GetBookObjectFromPost(request);
-                    if (book.FolderPath != _bookSelection?.CurrentSelection?.FolderPath)
+                    // Lock the workspace tabs while we load the book, as in "selected-book"
+                    // above (BL-15971).
+                    _collectionTabView.WorkspaceView?.SetTabsEnabled(false);
+                    try
                     {
-                        _collectionModel.SelectBook(book);
+                        var book = GetBookObjectFromPost(request);
+                        if (book.FolderPath != _bookSelection?.CurrentSelection?.FolderPath)
+                        {
+                            _collectionModel.SelectBook(book);
+                        }
+                        if (
+                            book.IsSaveable
+                            && GetCollectionOfRequest(request).Type
+                                == BookCollection.CollectionType.TheOneEditableCollection
+                        )
+                        {
+                            _editBookCommand.Raise(_bookSelection.CurrentSelection);
+                        }
                     }
-                    if (
-                        book.IsSaveable
-                        && GetCollectionOfRequest(request).Type
-                            == BookCollection.CollectionType.TheOneEditableCollection
-                    )
+                    finally
                     {
-                        _editBookCommand.Raise(_bookSelection.CurrentSelection);
+                        _collectionTabView.WorkspaceView?.SetTabsEnabled(true);
                     }
 
                     request.PostSucceeded();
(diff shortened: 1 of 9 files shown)


andrew-polk:
[Claude Fable 5] Consulted Devin on 2026-08-10 ~12:35 PT up to commit 63ac774dd305928de22cae0ddd728768ff2fd0f0. Result: clean — 0 bugs, 0 Investigate flags. 4 Informational observations (shared tab-lock re-enable semantics, lock scope vs. async edit-tab navigation, JS ordering of the blank-then-navigate calls, TopBar import side effect) — assessed, none actionable; details on the Devin review page. CI (pr-automation) green.

andrew-polk:
[Claude Fable 5] Consulted Devin on 2026-08-10 ~14:55 PT up to commit f9882869e095bfdc0bca81a2bd9510295cac5a41. Result: clean — 0 bugs, 0 Investigate flags. 6 Informational observations, all assessed, none actionable: four repeats from the previous review (the 'TopBar side effect' one is now stale — that import was moved to the new side-effect-free workspaceTabInfo.ts in this very commit, per the developer's decision), plus two new notes on ClearOutDisplay (a brief dev-server-reload-only window where the page URL points at about:blank, and a pageUnloading call that already fired on the other book-change path). Details on the Devin review page. CI (pr-automation) green.

src/BloomExe/web/controllers/CollectionApi.cs
```diff
@@ -171,6 +171,11 @@ public void RegisterWithApiHandler(BloomApiHandler apiHandler)
                             // without actually loading all the files. We need the title for the Performance Measurement and later
                             // we'll use the BookInfo object to get the fully updated book.
                             BookInfo newBookInfo = null;
+                            // Like other long-running operations (Edit-tab saves, uploads), lock the
+                            // workspace tabs while we load the newly selected book. Bringing a heavy
+                            // book up to date can take several seconds, and clicking Edit during that
+                            // window could show the previous book (BL-15971).
+                            _collectionTabView.WorkspaceView?.SetTabsEnabled(false);
```
andrew-polk, line 178: [Devin] **Investigate**: Lock is taken on the UI thread that the book load then blocks — verify the browser actually receives the disabled state in time

Unlike the existing users of `SetTabsEnabled` (BloomLibrary upload, RAB builds), which run their long work off the UI thread, the `selected-book` handler is registered with `handleOnUiThread: true` and the expensive `GetUpdatedBookObjectFromBookInfo`/`BringBookUpToDate` work happens while the UI thread is blocked. `SetTabsEnabled` pushes the new state via `WorkspaceView.SendTopBarState` → `_webSocketServer.SendBundle("workspace", "tabs", ...)`. This only produces the intended lock if `useWatchApiObject` (workspaceTabInfo.ts) consumes the payload carried in the websocket message. If instead it responds to the event by re-`get`ting `workspace/tabs` through the API, that request cannot be served while the UI thread is blocked in this very handler, so the greyed-out tabs/button would only appear (briefly, then immediately clear) after the load finishes — i.e. the fix would be a no-op for exactly the slow-book case it targets. Worth confirming by inspecting `useWatchApiObject` and by observing a real slow selection.

andrew-polk: [Claude Fable 5] Not an issue — verified the delivery path. `useWorkspaceTabInfo` is built on `useWatchApiObject` (bloomApi.ts), whose websocket subscription is `useSubscribeToWebSocketForObject`: it takes the new state **directly from the websocket message payload** (`setVal(message)`), with no API round-trip per event. The `get("workspace/tabs")` call only happens at component mount and on websocket reconnect. So the disabled state is pushed to the browser by `SendBundle` at the moment `SetTabsEnabled(false)` runs, before the UI thread blocks in `BringBookUpToDate`, and the tabs/button grey out for the whole load — exactly the slow-book case this fix targets. (The re-GET-per-event behavior described is the sibling hook `useWatchApiData`, which this code does not use.) Resolving.

andrew-polk:
[Claude Fable 5] Consulted Devin on 2026-08-10 ~15:20 PT up to commit 25d7f1a00dc88fe100ec7185a26565c7d0553eda (the rebased head; tree identical to the previously reviewed f988286). Result: 0 bugs; 1 Investigate flag (does the tab-disable actually reach the browser while the UI thread is blocked?) — mirrored as an inline thread, verified not an issue (the hook consumes the websocket payload directly, no per-event API round-trip), and resolved: https://github.com/BloomBooks/BloomDesktop/pull/8185#discussion_r3753791497. 7 Informational observations, same set as prior rounds, none actionable. CI green.

andrew-polk:
Converting to draft as JT and I discuss if this is the right approach.

PR #8186, opened 2026-08-10
Fix BL-16209 Bloom won't open a collection with illegal chars in language display name

andrew-polk:
A collection whose language display name contained a `"` or a `\` could not be opened at all: Bloom showed a blank window with no message, permanently, because the name is saved in the `.bloomCollection` file.

`settings/languageData` built its JSON reply by string interpolation:

```csharp
$"{{\"languageName\":\"{languageName}\",\"languageCode\":\"{langTag}\"}}"
```

so the reporter's name produced invalid JSON:

```
{"languageName":"Unsafe \~!@#$%^&amp;\*()_-+={}\|\\:;"'\<\>.?/","languageCode":"qaa-BA-x-Unsafeam"}
```

axios can't parse that, leaves the body as a raw string, `langData.data.languageName` is `undefined`, and `BooksOnBlorgProgressBar.tsx:36` throws on `languageName.length`. With no error boundary above it, React unmounts the whole tree — hence the blank screen and the total silence (the JS error *is* reported to `common/error`, but that only raises a toast, and the toast host lives in the tree that just died).

Not an exotic input: apostrophes are common in real language names, and one double quote is enough.

## Changes

- `CollectionSettingsApi`: serialize the reply with `JsonConvert` via a new testable `MakeLanguageDataJson`. Also fixes the second consumer of that endpoint, the "Make Reader Template Bloom Pack" dialog, which was rendering `undefined` for the collection language.
- `ReadersApi.SaveSynphonyLanguageData`: same bug class — it spliced the raw display name into the `LangName` field of the Synphony language-data JSON. Now uses `JsonConvert.ToString` for quoting/escaping.
- Both keep coercing a null name to `""`, as the old hand-built strings did, so a null can't reintroduce the same crash by another route.
- New `CollectionSettingsApiTests` (3 tests): the JSON-breaking name, a null name, an ordinary name. The first fails against the old code.

## Verification

Reproduced in the real app with the reporter's "Please Delete ME" collection from the card: confirmed the blank screen and captured the `Cannot read properties of undefined (reading 'length')` throw at `BooksOnBlorgProgressBar.tsx:36` over CDP, then after the fix confirmed the collection tab renders fully, with the progress-bar label reading `0 Unsafe \~!@... books on BloomLibrary.org`.

Ref: https://issues.bloomlibrary.org/youtrack/issue/BL-16209

[Devin review](https://devinreview.com/BloomBooks/BloomDesktop/pull/8186)

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8186)

### DIFF
--- a/src/BloomExe/web/ReadersApi.cs
+++ b/src/BloomExe/web/ReadersApi.cs
@@ -415,6 +415,12 @@ namespace Bloom.Api
                     e
                 );
                 request.Failed("Saving synphonyLanguageData failed to get Json");
+                // The return matters (BL-16209 review). Without it we fell through with langdata
+                // still "", wrote that empty string over the collection's ReaderToolsWords-<lang>.json
+                // — discarding the user's word list and grapheme data — and then reported the same
+                // request as succeeded on top of having failed it. Doing nothing here is exactly what
+                // the note above intends by "just not update the file".
+                return;
             }
 
             SaveSynphonyLanguageData(langdata);
@@ -487,6 +493,20 @@ namespace Bloom.Api
             return $"contains(concat(' ', @class, ' '), ' {className} ')";
         }
 
+        /// <remarks>
+        /// Reviewed during BL-16209 (which was caused by a display name with a double quote in it
+        /// breaking hand-built JSON elsewhere). This escaper handles the backslash and the double
+        /// quote — and note it must escape the backslash FIRST, or it would re-escape the
+        /// backslash the quote rule just added — plus carriage return and newline. It does not
+        /// escape the other JSON control characters, e.g. a literal tab or anything else below
+        /// U+0020, which RFC 8259 also requires to be escaped.
+        ///
+        /// We decided deliberately to leave that as it is rather than switch to JsonConvert: the
+        /// text through here is book page content, so a stray tab is far less likely than the
+        /// quote mark that caused BL-16209, and changing it would alter the exact bytes of other
+        /// reader-tool JSON for no known symptom. Revisit if page text ever turns out to carry
+        /// control characters through this path.
+        /// </remarks>
         private static string EscapeJsonValue(string value)
         {
             // As http://stackoverflow.com/questions/42068/how-do-i-handle-newlines-in-json attempts to explain,
@@ -741,17 +761,23 @@ namespace Bloom.Api
         /// records of "when things changed" and so forth.</Note>
         private void SaveSynphonyLanguageData(string jsonString)
         {
-            // insert LangName and LangID if missing
+            // insert LangName and LangID if missing.
+            // JsonConvert.ToString() supplies the enclosing quotes and escapes anything in the
+            // value that JSON would otherwise choke on; a display name is arbitrary user text and
+            // may contain a double quote or backslash (BL-16209). A language tag is far more
+            // constrained (it also has to survive being put in a file name just below), but it
+            // goes through the same escaping so this method has no hand-quoted values left.
             if (jsonString.Contains("\"LangName\":\"\""))
                 jsonString = jsonString.Replace(
                     "\"LangName\":\"\"",
-                    "\"LangName\":\"" + CurrentBook.BookData.Language1.Name + "\""
+                    "\"LangName\":"
+                        + JsonConvert.ToString(CurrentBook.BookData.Language1.Name ?? "")
                 );
 
             if (jsonString.Contains("\"LangID\":\"\""))
                 jsonString = jsonString.Replace(
                     "\"LangID\":\"\"",
-                    "\"LangID\":\"" + CurrentBook.BookData.Language1.Tag + "\""
+                    "\"LangID\":" + JsonConvert.ToString(CurrentBook.BookData.Language1.Tag ?? "")
                 );
 
             var fileName = String.Format(
@@ -768,26 +794,28 @@ namespace Bloom.Api
         }
 
         /// <summary>
-        /// Efficiently determines if we should write the language data file by comparing
-        /// file size and content bytes to avoid unnecessary writes.
+        /// Determines whether the language data file actually needs writing, by comparing what
+        /// is already on disk with what we are about to write. See the Note on
+        /// SaveSynphonyLanguageData for why avoiding a pointless write matters.
         /// </summary>
-        private bool ShouldWriteLanguageDataFile(string fileName, string newContent)
+        /// <remarks>
+        /// This compares the decoded text, not raw bytes. It used to compare
+        /// Encoding.UTF8.GetBytes(newContent) against the file, but the file is written with
+        /// Encoding.UTF8, which emits a byte-order mark that GetBytes() does not. The lengths
+        /// therefore always differed by those 3 bytes, so this always returned true and the
+        /// optimization never once skipped a write (found by Devin's review of BL-16209).
+        /// Comparing text also means a file saved without a BOM by some older version still
+        /// compares equal, so we don't rewrite it just to add one.
+        /// </remarks>
+        internal static bool ShouldWriteLanguageDataFile(string fileName, string newContent)
         {
             try
             {
                 if (!RobustFile.Exists(fileName))
                     return true;
 
-                var newContentBytes = Encoding.UTF8.GetBytes(newContent);
-                var fileInfo = new FileInfo(fileName);
-
-                // Quick check: if file size is different, content must be different
-                if (fileInfo.Length != newContentBytes.Length)
-                    return true;
-
-                // File sizes match, so compare the actual byte content
-                var existingContentBytes = RobustFile.ReadAllBytes(fileName);
-                return !newContentBytes.SequenceEqual(existingContentBytes);
+                // ReadAllText detects and strips a byte-order mark, so this is BOM-insensitive.
+                return RobustFile.ReadAllText(fileName, Encoding.UTF8) != newContent;
             }
             catch
             {
--- a/src/BloomExe/web/controllers/CollectionSettingsApi.cs
+++ b/src/BloomExe/web/controllers/CollectionSettingsApi.cs
@@ -509,9 +509,23 @@ namespace Bloom.web.controllers
                     _collectionSettings.Language1Tag
                 );
             }
-            var jsonString =
-                $"{{\"languageName\":\"{languageName}\",\"languageCode\":\"{langTag}\"}}";
-            request.ReplyWithJson(jsonString);
+            request.ReplyWithJson(MakeLanguageDataJson(languageName, langTag));
+        }
+
+        /// <summary>
+        /// Builds the JSON that the languageData endpoint returns. A display name is arbitrary
+        /// user text: it can perfectly well contain a double quote or a backslash (BL-16209), so
+        /// it has to be serialized rather than pasted into a hand-built JSON string. Getting that
+        /// wrong produced invalid JSON, which made the whole collection tab fail to render.
+        /// The callers of this endpoint treat languageName as a string (one of them asks it for
+        /// its .length), so keep coercing a null name to empty the way the old hand-built string
+        /// did rather than sending a JSON null.
+        /// </summary>
+        internal static string MakeLanguageDataJson(string languageName, string languageTag)
+        {
+            return JsonConvert.SerializeObject(
+                new { languageName = languageName ?? "", languageCode = languageTag ?? "" }
+            );
         }
 
         // Used by BookSettingsDialog and others
--- a/src/BloomTests/web/controllers/CollectionSettingsApiTests.cs
+++ b/src/BloomTests/web/controllers/CollectionSettingsApiTests.cs
@@ -1,0 +1,67 @@
+using Bloom.web.controllers;
+using Newtonsoft.Json;
+using NUnit.Framework;
+
+namespace BloomTests.web.controllers
+{
+    [TestFixture]
+    public class CollectionSettingsApiTests
+    {
+        // A language display name is arbitrary user text. Apostrophes are common in real language
+        // names, and BL-16209 showed that a user can put anything at all in there, including the
+        // characters that break a hand-built JSON string.
+        private const string kNastyDisplayName = "Unsafe ~!@#$%^&*()_-+={}[]\\|:;\"'<>.?/";
+
+        /// <summary>
+        /// BL-16209: the languageData reply used to be assembled by string interpolation, so a
+        /// display name containing a double quote or a backslash produced invalid JSON. The
+        /// collection tab then read undefined for languageName, threw, and left the user with a
+        /// blank screen they could not get past.
+        /// </summary>
+        [Test]
+        public void MakeLanguageDataJson_NameHasJsonBreakingCharacters_ProducesValidParseableJson()
+        {
+            // Sanity check the test data: this only proves anything if the name really does
+            // contain the characters that JSON has to escape.
+            Assert.That(
+                kNastyDisplayName,
+                Does.Contain("\"").And.Contain("\\"),
+                "Test data should contain a double quote and a backslash"
+            );
+
+            var json = CollectionSettingsApi.MakeLanguageDataJson(
+                kNastyDisplayName,
+                "qaa-BA-x-Unsafeam"
+            );
+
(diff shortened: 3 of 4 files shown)


andrew-polk:
[Claude Opus 5 (1M context) from Andrew Polk's machine during preflight]

Consulted Devin on 2026-08-10 up to commit `2462597c0340c037bbe1dbe602ce8eec56ddad66`.

**No bugs and nothing flagged for investigation.** Six informational observations, so nothing was
mirrored as a review thread. Two of them were worth acting on or knowing about:

- Devin noticed that the earlier commit escaped the language *name* in the Decodable Reader's data
  file but left the language *tag* two lines below hand-quoted. Fair point — same bug class, same
  method — so that is now fixed in `2462597c`, which is why this is the second Devin run on this PR.
- It also observed that the language tag still goes straight into the name of the file that data is
  written to, where JSON escaping is irrelevant. That is true and pre-existing, out of scope for
  this card, and raised with the developer as an FYI rather than changed here.

One flag ("LangID is still spliced in without escaping") is stale — it describes the state before
`2462597c` and no longer matches the code.

CI (`pr-automation`) passed. CodeRabbit does not review this repo (`auto_review` is disabled in
`.coderabbit.yml`), and TeamCity does not build feature branches, so there is nothing else pending.

src/BloomExe/web/ReadersApi.cs
```diff
@@ -768,26 +774,28 @@ private void SaveSynphonyLanguageData(string jsonString)
...
+        /// <remarks>
+        /// This compares the decoded text, not raw bytes. It used to compare
+        /// Encoding.UTF8.GetBytes(newContent) against the file, but the file is written with
+        /// Encoding.UTF8, which emits a byte-order mark that GetBytes() does not. The lengths
+        /// therefore always differed by those 3 bytes, so this always returned true and the
+        /// optimization never once skipped a write (found by Devin's review of BL-16209).
+        /// Comparing text also means a file saved without a BOM by some older version still
+        /// compares equal, so we don't rewrite it just to add one.
+        /// </remarks>
+        internal static bool ShouldWriteLanguageDataFile(string fileName, string newContent)
```
andrew-polk, line 810: [Devin] **Investigate**: Pre-existing: the write-avoidance byte comparison likely never matches because of the UTF-8 BOM

`RobustFile.WriteAllText(fileName, jsonString, Encoding.UTF8)` writes a byte-order mark, but `ShouldWriteLanguageDataFile` compares the file's bytes against `Encoding.UTF8.GetBytes(newContent)`, which has no BOM. The lengths therefore always differ by 3 and the method always returns true, so the "only write if changed" optimization documented in the `<Note>` on `SaveSynphonyLanguageData` never actually skips a write (and TC syncing is still triggered). Pre-existing, not introduced here, but adjacent to the changed code.

*(Mirrored from Devin's review of `896e9f69`, which is not posted to GitHub automatically.)*

andrew-polk: [Claude Opus 5 (1M context) from Andrew Polk's machine during preflight] Fixed in `6c91ee7d`: Devin is right, and I confirmed it in the code — the file is written with `Encoding.UTF8` (which emits a BOM) while the comparison used `Encoding.UTF8.GetBytes(newContent)` (which does not), so the lengths always differed by exactly 3 and the check could never return false. The write-avoidance the `<Note>` on `SaveSynphonyLanguageData` describes — and with it the avoidance of a Team Collection sync — never worked once.

It now compares the decoded text instead of raw bytes, which fixes the BOM mismatch and also stops us rewriting a file an older version left without a BOM just to add one. The method is `internal static` so it can be tested directly, and there are four new `ReadersApiTests` cases (BOM, no BOM, changed content, missing file); the BOM one fails against the old byte comparison.

Pre-existing rather than introduced by this branch, but it lives in the method this branch already changes, and it explains Devin's earlier observation that these files get rewritten once — in fact they were being rewritten every time.

andrew-polk, line 810: [Devin] **Investigate**: Behavior change: language-data file writes are now actually skipped

This is not just a refactor of the comparison. Under the old byte comparison the on-disk file (written with `Encoding.UTF8`, which emits a BOM) was always 3 bytes longer than `Encoding.UTF8.GetBytes(newContent)`, so the method always returned true and the file was rewritten on every call. With the text comparison the skip path becomes live for the first time, meaning the Synphony language-data file will now genuinely stop being touched (and, in a Team Collection, stop triggering the collection-level sync described in the Note on `SaveSynphonyLanguageData`). That is the stated intent, but it is a first-time-active code path worth exercising once in a real Team Collection: if anything downstream was implicitly relying on the file's timestamp changing on every book open, it would only surface now.

*(Mirrored from Devin's review of `7ada2427`, which is not posted to GitHub automatically.)*

src/BloomExe/web/ReadersApi.cs
```diff
@@ -741,17 +741,23 @@ private void MakeLetterAndWordList(string jsonSettings, string allWords)
         /// records of "when things changed" and so forth.</Note>
         private void SaveSynphonyLanguageData(string jsonString)
         {
-            // insert LangName and LangID if missing
+            // insert LangName and LangID if missing.
```
andrew-polk, line 764: [Devin] **Investigate**: Other hand-built JSON containing user text may remain elsewhere

This PR fixes the two hand-built JSON strings it found (`settings/languageData` and the Synphony LangName/LangID splice). `src/BloomExe/web/ReadersApi.cs:498-502` still contains a hand-rolled escaper (only `"`, `\r`, `\n` — notably not backslash) used for other reader-tool JSON. If any user-supplied text flows through that path, it has the same failure mode as BL-16209 and would deserve the same `JsonConvert` treatment.

*(Mirrored from Devin's review of `7ada2427`, which is not posted to GitHub automatically.)*

andrew-polk: [Claude Opus 5 (1M context) from Andrew Polk's machine during preflight] **Not an issue as described — `EscapeJsonValue` does escape the backslash**, and it does it first, which is the order that matters (`src/BloomExe/web/ReadersApi.cs:499`):

    return value
        .Replace("\\", "\\\\")   // backslash, escaped first
        .Replace("\"", "\\\"")
        .Replace("\r", "\\r")
        .Replace("\n", "\\n");

Escaping the backslash before the quote is required — doing it the other way round would re-escape the backslash that the quote rule just introduced. So the specific failure mode of BL-16209 (an unescaped `"` or `\` breaking the JSON) does not apply to this path.

The broader point is fair, though, so for the record: that escaper does not cover the *other* JSON control characters — a literal tab, or anything else in U+0000–U+001F — which RFC 8259 also requires to be escaped. That is a far narrower exposure than a quote or a backslash, the text it handles comes from book page content, and it is pre-existing and untouched by this branch. So I am leaving it alone rather than widening a targeted fix. Resolving; worth revisiting if it turns out real page text can carry a tab through that path.

src/BloomExe/web/ReadersApi.cs
andrew-polk: [Claude Opus 5 (1M context) from Andrew Polk's machine during preflight] **Correct, and intended — this is the point of the change, not a side effect.** Confirmed in the code: the file is written with `Encoding.UTF8`, which emits a BOM, while the old comparison used `Encoding.UTF8.GetBytes(newContent)`, which does not, so the lengths always differed by exactly those 3 bytes and the method could never return false. The write-avoidance that the `<Note>` on `SaveSynphonyLanguageData` describes — and with it the avoidance of a Team Collection sync — has never once worked. Making the skip path live is the whole fix.

Your framing of the risk is the useful part, so it is now covered rather than just noted: the QA test ideas on BL-16209 ask a tester to open the Decodable Reader tool in a **Team Collection** whose collection is affected, confirm the one-time rewrite settles, and confirm that subsequent opens no longer set off a collection-level sync and that nothing is reported as conflicting or lost. That is precisely the "first-time-active path in a real Team Collection" exercise you are asking for.

Worth adding for whoever reads this later: no test in the suite reaches `SaveSynphonyLanguageData`, so this path is not covered automatically and the manual Team Collection check is the real verification. The four new `ReadersApiTests` cases cover the comparison itself (BOM present, BOM absent, content changed, file missing) but not the endpoint that calls it. Resolving, with that manual test recorded on the card.

andrew-polk:
[Claude Opus 5 (1M context) from Andrew Polk's machine during preflight]

Consulted Devin on 2026-08-10 up to commit `7ada2427f032fcd285ec916dedd191606bed1ed3`.

**No bugs. Two things flagged for investigation, both now mirrored above as threads with an outcome, plus ten informational observations.**

- *Behaviour change: language-data file writes are now actually skipped* — correct, and it is the point of the change rather than a side effect. The write-avoidance in `SaveSynphonyLanguageData` had never once worked, so the skip path is live for the first time. Devin's framing of the risk was the useful part, so the QA notes on BL-16209 now ask a tester to exercise exactly that in a real Team Collection. [Thread](https://github.com/BloomBooks/BloomDesktop/pull/8186#discussion_r3754223795)
- *Other hand-built JSON containing user text may remain elsewhere* — **not an issue as described.** `EscapeJsonValue` does escape the backslash, and does it first, which is the order that matters. The residual gap is only the other JSON control characters (a literal tab and friends), which is pre-existing, much narrower, and left alone here. [Thread](https://github.com/BloomBooks/BloomDesktop/pull/8186#discussion_r3754225494)

Earlier in the same session Devin found the UTF-8 BOM problem in `ShouldWriteLanguageDataFile`, which is fixed in `6c91ee7d` and recorded in [its own thread](https://github.com/BloomBooks/BloomDesktop/pull/8186#discussion_r3754117099).

CI (`pr-automation`) passed. CodeRabbit does not review this repo (`auto_review` is disabled in `.coderabbit.yml`) and TeamCity does not build feature branches, so nothing else is pending.

One note on the full C# suite, since it is not a clean story: it **passed at this commit** (3070 passed, 12 skipped, 0 failed), but two earlier runs today aborted with the test host crashing in `HttpListener` teardown (`ObjectDisposedException` on `HttpRequestQueueV2Handle`), around 350 tests in. That looks like a pre-existing intermittent race in `BloomServer` disposal rather than anything on this branch — the same tree both aborted and passed — and it is raised for the developer as its own item.

src/BloomExe/web/ReadersApi.cs
andrew-polk: [Claude Opus 5 (1M context) from Andrew Polk's machine during preflight] **Developer decision: leave as is, and recorded in the code.**

Andrew's call on the narrower residual point (that this escaper does not cover a literal tab or the other characters below U+0020) is to leave it alone, for the reasons above: the text through that path is book page content, and changing it would alter the exact bytes of other reader-tool JSON for no known symptom.

He asked for the reasoning to live in the source rather than only here, so `EscapeJsonValue` now carries a `<remarks>` block covering both halves of this thread — why the backslash handling is correct and order-dependent, and why the control-character gap is knowingly left. Committed in `3436b257`.

(The original claim, that backslashes are unescaped, remains refuted — see above.)

src/BloomExe/web/ReadersApi.cs
andrew-polk, line 1: [Devin] **Investigate**: Pre-existing: request is both failed and succeeded when reading the posted JSON throws

`src/BloomExe/web/ReadersApi.cs:406-421` — the `catch (IOException)` around `request.RequiredPostJson()` calls `request.Failed(...)` but does not return, so execution falls through to `SaveSynphonyLanguageData(langdata)` and then `request.PostSucceeded()`. The same request is thus reported as both failed and succeeded.

*(Mirrored from Devin's review of `c0c7b2f17e`, which is not posted to GitHub automatically.)*

---

[Claude Opus 5 (1M context) from Andrew Polk's machine during preflight] **Confirmed, and the consequence is worse than the double-response.** Not fixed here — flagged to the developer.

The comment immediately above that `try` states the intent plainly: *"we will ignore failure to read all the json, and just **not update the file**."* The code does the opposite. On an `IOException`, `langdata` is still `""`, and execution continues into `SaveSynphonyLanguageData("")`, where:

- neither `Contains("\"LangName\":\"\"")` nor the `LangID` check matches, so nothing is spliced;
- `ShouldWriteLanguageDataFile(fileName, "")` compares the existing file against `""`, finds them different, and returns true;
- so the collection's `ReaderToolsWords-<lang>.json` is **overwritten with an empty file**.

That is the user's Synphony word list and grapheme data for the collection, replaced with nothing — on precisely the transient failure the comment says is expected when "switching between books fast" with "a very large block of data" (BL-4313 and friends). So the guard that exists to avoid a crash can instead quietly discard reader data.

Two notes on scope:

- **This branch neither causes nor worsens it.** Before the write-avoidance fix on this branch, `ShouldWriteLanguageDataFile` always returned true, so the empty overwrite happened just the same. The fix does not touch this path.
- The likely fix is a one-line `return` in the `catch` (which is what the comment already describes), but it sits in a data-writing error path in a file where review has now turned up three separate pre-existing problems, so it deserves its own card and its own testing rather than being folded into a fix about JSON escaping.

Leaving this thread **open** for the developer's decision.

andrew-polk: [Claude Opus 5 (1M context) from Andrew Polk's machine during preflight] **Fixed in `0b405085`** — Andrew's decision was to fix it on this branch rather than defer it.

The `catch` now returns, so on an `IOException` reading the posted JSON, Bloom logs it, reports the failure, and stops — leaving the collection's `ReaderToolsWords-<lang>.json` untouched. That is what the note above the `try` always said it intended ("just not update the file"). Both defects go with it: no empty overwrite of the user's word list, and no reporting the same request as succeeded after failing it.

I deliberately kept the existing `request.Failed(...)` call rather than converting the path to a silent success. Worth saying why, since the note's word "ignore" could be read as arguing for the latter: `Failed(...)` already runs today, so the client already sees this as a failed post and already turns it into a JS error report via `wrapAxios`. Keeping it means this fix changes exactly one thing — it stops destroying data — and leaves observable client behaviour alone. Whether this transient, expected condition *should* be reported to the client at all is a separate question, and not one to settle inside a fix about JSON escaping.

Not covered by an automated test: provoking it needs an `IOException` out of the request stream, which the test server cannot readily produce. The change is a plain early return in a path whose only previous outcome was data loss.

Full C# suite green at this commit: 3070 passed, 0 failed, 12 skipped.

andrew-polk:
[Claude Opus 5 (1M context) from Andrew Polk's machine during preflight]

Consulted Devin on 2026-08-11 up to commit `0b405085a6`.

**No bugs.** Three Investigate flags, all of which are restatements of threads already resolved above, so nothing new was mirrored:

- *Behaviour change: language-data file writes are now actually skipped* — intended; that is the fix. Covered by a manual Team Collection test on BL-16209. [Thread](https://github.com/BloomBooks/BloomDesktop/pull/8186#discussion_r3754223795)
- *EscapeJsonValue still not RFC-compliant for control characters* — the original claim (unescaped backslashes) was refuted; the narrower control-character point is a deliberate, developer-approved leave-as-is, now recorded in a `remarks` block on the method itself. [Thread](https://github.com/BloomBooks/BloomDesktop/pull/8186#discussion_r3754225494)
- *Pre-existing: request is both failed and succeeded when reading the posted JSON throws* — **this one is stale.** It was fixed in this very commit; the `catch` now returns. Verified against the pushed tree, where `ReadersApi.cs:423` is `return;` inside that `catch`. Devin has carried a superseded flag forward once before on this PR (the `LangID` splice), so this looks like the same behaviour rather than a real finding. [Thread](https://github.com/BloomBooks/BloomDesktop/pull/8186#discussion_r3761032690)

CI (`pr-automation`) passed. CodeRabbit does not review this repo (`auto_review` disabled in `.coderabbit.yml`) and TeamCity builds no feature branches, so nothing else is pending.

Full C# suite at this commit: 3070 passed, 0 failed, 12 skipped.

StephenMcConnel (approved):
> @StephenMcConnel reviewed 4 files and all commit messages, and resolved 4 discussions.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8186)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on andrew-polk).

PR #8187, opened 2026-08-10
Fix BL-16639 Extra text in "By Sentence" tooltip in disabled state on page with no text area

andrew-polk:
On a page with no text area, hovering the disabled **"By Sentence"** radio in the Talking Book tool's Advanced section opened *two* tooltips at once, and one of them gave the wrong reason.

The radio sits inside two nested `BloomTooltip`s:

| Tooltip | Text |
|---|---|
| Outer, wrapping the whole Recording Mode group | "First, put the cursor in a box which has something to record." |
| Inner, on the radio — given unconditionally | "If you want to change this mode, first use the \"Clear\" button to remove your **recording.**" |

Hover bubbles from the inner `<span>` to the outer one, so MUI opened both, overlapping at `top-start`. Whichever mounted second covered the other, and because the has-audio message is a line taller, its last line ("recording.") peeked out below the correct tooltip — the stray text in the bug report.

The fix gives the radio a disabled tip only when it is genuinely disabled *because of existing audio*. When the page has nothing recordable the tip is empty, so only the group-level tooltip shows — the same pattern the neighbouring "By Whole Text Box" tooltip already uses. When `hasRecordableDivs` is true the expression yields exactly the object it did before, so the has-audio path is unchanged.

No XLF changes: no localizable string was added or modified.

## Verification

Attached to a running Bloom over CDP on a "Just an Image" page, counting open tooltips in the DOM:

- **original code restored** → 2 tooltips, the second overlapping the first at nearly identical coordinates;
- **with the fix** → exactly 1: "First, put the cursor in a box which has something to record.";
- **page with text** → radio enabled, still shows "Record one sentence at a time…".

Ref: https://issues.bloomlibrary.org/youtrack/issue/BL-16639

[Devin review](https://devinreview.com/BloomBooks/BloomDesktop/pull/8187)

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8187)

### DIFF
--- a/src/BloomBrowserUI/bookEdit/toolbox/talkingBook/talkingBookAdvancedSection.tsx
+++ b/src/BloomBrowserUI/bookEdit/toolbox/talkingBook/talkingBookAdvancedSection.tsx
@@ -41,6 +41,15 @@ export const TalkingBookAdvancedSection: React.FunctionComponent<{
         props.recordingMode === RecordingMode.TextBox &&
         !!wholeTextBoxAudioFeatureStatus?.enabled &&
         props.hasRecordableDivs;
+    // The "By Sentence" radio and the tooltip wrapped around it must agree about when the
+    // option is unavailable. They used to test different things, so the tooltip could explain
+    // a disabled state the user was not actually in: it went by whether the page had any audio
+    // at all, while the radio went by whether there was a whole-text-box recording in the way.
+    // The radio's rule is the correct one, so both now read it from here (BL-16639).
+    const sentenceModeDisabled =
+        !props.hasRecordableDivs ||
+        // we don't allow you to go from textbox to sentence mode if you have a recording of the whole box
+        props.haveACurrentTextboxModeRecording;
     // The toolbox is currently its own iframe, so we can't spill out to the left yet.
     // Unfortunately, we can't spill out the bottom without bad side effects either (BL-12366), so we go topside.
     const commonTooltipProps = {
@@ -129,11 +138,7 @@ export const TalkingBookAdvancedSection: React.FunctionComponent<{
                                 l10nKey:
                                     "EditTab.Toolbox.TalkingBookTool.RecordingModeSentenceTip",
                             }}
-                            showDisabled={
-                                !props.hasRecordableDivs ||
-                                // we don't allow you to go from textbox to sentence mode if you have audio
-                                props.hasAudio
-                            }
+                            showDisabled={sentenceModeDisabled}
                             // When there is nothing recordable on this page, the tooltip around the
                             // whole Recording Mode group already explains that. Adding a tip here too
                             // would leave two tooltips overlapping on the same hover, and this one
@@ -148,12 +153,7 @@ export const TalkingBookAdvancedSection: React.FunctionComponent<{
                             {...commonTooltipProps}
                         >
                             <MuiRadio
-                                disabled={
-                                    !props.hasRecordableDivs ||
-                                    // we don't allow you to go from textbox to sentence mode if you have audio
-                                    //props.hasAudio
-                                    props.haveACurrentTextboxModeRecording
-                                }
+                                disabled={sentenceModeDisabled}
                                 value={RecordingMode.Sentence}
                                 label="By Sentence"
                                 l10nKey="EditTab.Toolbox.TalkingBookTool.RecordingModeSentence"
--- a/src/BloomBrowserUI/bookEdit/toolbox/talkingBook/talkingBookAdvancedSection.tsx
+++ b/src/BloomBrowserUI/bookEdit/toolbox/talkingBook/talkingBookAdvancedSection.tsx
@@ -134,9 +134,17 @@ export const TalkingBookAdvancedSection: React.FunctionComponent<{
                                 // we don't allow you to go from textbox to sentence mode if you have audio
                                 props.hasAudio
                             }
-                            tipWhenDisabled={{
-                                l10nKey: `EditTab.Toolbox.TalkingBookTool.RecordingModeDisabledBecauseHasAudioTip`,
-                            }}
+                            // When there is nothing recordable on this page, the tooltip around the
+                            // whole Recording Mode group already explains that. Adding a tip here too
+                            // would leave two tooltips overlapping on the same hover, and this one
+                            // would be telling the user the wrong reason anyway (BL-16639).
+                            tipWhenDisabled={
+                                props.hasRecordableDivs
+                                    ? {
+                                          l10nKey: `EditTab.Toolbox.TalkingBookTool.RecordingModeDisabledBecauseHasAudioTip`,
+                                      }
+                                    : ""
+                            }
                             {...commonTooltipProps}
                         >
                             <MuiRadio


src/BloomBrowserUI/bookEdit/toolbox/talkingBook/talkingBookAdvancedSection.tsx
```diff
@@ -134,9 +134,17 @@ export const TalkingBookAdvancedSection: React.FunctionComponent<{
...
+                            // whole Recording Mode group already explains that. Adding a tip here too
+                            // would leave two tooltips overlapping on the same hover, and this one
+                            // would be telling the user the wrong reason anyway (BL-16639).
+                            tipWhenDisabled={
+                                props.hasRecordableDivs
+                                    ? {
+                                          l10nKey: `EditTab.Toolbox.TalkingBookTool.RecordingModeDisabledBecauseHasAudioTip`,
+                                      }
+                                    : ""
+                            }
```
andrew-polk, line 152: [Devin] **Investigate**: No explanatory tooltip when there is audio but nothing recordable on the page

With the new conditional, if `props.hasRecordableDivs` is false the radio's disabled tip becomes `""`. The outer group tooltip (`talkingBookAdvancedSection.tsx:96-103`) only shows its "put the cursor in a box..." text when `!props.hasAudio`; when `props.hasAudio` is true it falls back to `props.tip`, which is not supplied for that outer tooltip. So in the (probably unreachable in practice) combination `hasAudio === true && hasRecordableDivs === false`, hovering the disabled radio now yields no tooltip at all, whereas before it showed the has-audio explanation. Worth a quick sanity check whether that state can occur (e.g. a page whose recordable box was just deleted while audio metadata remains).

andrew-polk: [Claude Opus 5 (1M context) during preflight] Not an issue: that combination can't happen.

`hasAudio` is `AudioRecording.haveAudio`, and the only place it becomes `true` is `audioRecording.ts:3895`, where it is set from `doesElementAudioExist` — whether the **currently highlighted element** has a recording. An element that is currently highlighted for recording is by definition one of the recordable divs, so `hasAudio === true` implies `hasRecordableDivs === true`.

The specific scenario suggested here — a page whose recordable box was deleted while audio metadata lingers — is closed off explicitly. `handleNewPageReady` resets `this.haveAudio = false` for every page (`audioRecording.ts:2665`), and when a page has no editable text at all it forces it again with a comment saying exactly why: `this.haveAudio = false; // appropriately disables some advanced controls` (`audioRecording.ts:2697`).

Confirmed in a running Bloom: on a picture-only page the **outer** group tooltip displayed, which only happens when `!hasAudio` — i.e. `hasAudio` was false there, as predicted.

So no tooltip is lost. Adding a tip for a state that cannot occur would also cut against this repo's "fail fast, don't code around impossible states" guidance in `AGENTS.md`.

andrew-polk:
[Claude Opus 5 (1M context) from Polk's machine during preflight] Consulted Devin on 2026-08-10 22:16 UTC, up to commit `db97a8baffed16301a3ad94481257d2b907f5f5c`.

Devin found **no bugs**. It raised one **Investigate** flag and one **Informational** item, both on the changed file:

- *"No explanatory tooltip when there is audio but nothing recordable on the page"* — posted as its own thread, assessed as not an issue (that combination can't occur; `haveAudio` is only ever true for a currently-highlighted recordable element, and is forced false on a page with no editable text), and resolved. Reasoning is on [the thread](https://github.com/BloomBooks/BloomDesktop/pull/8187#discussion_r3753853742).
- *"Disabled-state mismatch between tooltip and radio remains"* — informational, so not mirrored as a thread. It's a real pre-existing inconsistency (the tooltip treats the radio as disabled when there's audio, but the radio itself keys off whether there's a whole-text-box recording), which this PR does not introduce or change. Raised with the developer as a scope question rather than fixed here.

CI (`pr-automation`) is green. CodeRabbit is configured in this repo but has `auto_review.enabled: false`, so it does not review PRs here.

andrew-polk:
[Claude Opus 5 (1M context) from Polk's machine during preflight] Follow-up on the informational item in the consultation above — the developer decided to act on it, so it is no longer just a scope note.

Devin observed that the "By Sentence" radio and the tooltip wrapped around it used **different** rules for deciding the option was unavailable: the tooltip went by whether the page had any audio at all, while the radio went by whether there was a whole-text-box recording in the way. In the gap between those rules a user saw a button that behaved as clickable while its tooltip insisted they had to press Clear first.

Andrew's call was to treat the **button's** rule as the correct one and align the tooltip to it. Rather than duplicate the condition in two places (which is how they drifted apart originally), both now read a single `sentenceModeDisabled` value computed once in the component, so they cannot disagree again.

Thanks Devin — this one was worth raising.

src/BloomBrowserUI/bookEdit/toolbox/talkingBook/talkingBookAdvancedSection.tsx
```diff
@@ -41,6 +41,15 @@ export const TalkingBookAdvancedSection: React.FunctionComponent<{
...
         props.hasRecordableDivs;
+    // The "By Sentence" radio and the tooltip wrapped around it must agree about when the
+    // option is unavailable. They used to test different things, so the tooltip could explain
+    // a disabled state the user was not actually in: it went by whether the page had any audio
+    // at all, while the radio went by whether there was a whole-text-box recording in the way.
+    // The radio's rule is the correct one, so both now read it from here (BL-16639).
+    const sentenceModeDisabled =
+        !props.hasRecordableDivs ||
+        // we don't allow you to go from textbox to sentence mode if you have a recording of the whole box
+        props.haveACurrentTextboxModeRecording;
```
andrew-polk, line 52: [Devin] **Investigate**: showDisabled change alters tooltip in sentence-mode-with-audio case

Previously the inner tooltip's `showDisabled` was `!hasRecordableDivs || props.hasAudio` while the radio's `disabled` was `!hasRecordableDivs || props.haveACurrentTextboxModeRecording`. On a page that already has sentence-level audio (`hasAudio` true, `haveACurrentTextboxModeRecording` false), the old code showed the "use Clear to remove your recording" tip even though the radio was enabled. With the unified `sentenceModeDisabled`, that case now shows the normal "Record one sentence at a time…" tip. This is a behavior change beyond the two-tooltip overlap described in the PR, and appears to be the desired one, but it is worth confirming with the product owner.

andrew-polk: [Claude Opus 5 (1M context) during preflight] Confirmed intended — this is exactly the change that was asked for, not a side effect.

Devin has read the diff correctly: on a page holding ordinary sentence-by-sentence audio, "By Sentence" used to *look* blocked and claim you had to press **Clear** first, while the button underneath was in fact clickable. That contradiction is what the change removes.

Devin raised the underlying mismatch on the previous commit as an informational item, it was put to Andrew as a decision, and his call was: **treat the button's enablement as correct and align the tooltip to it.** So the "Record one sentence at a time…" tip in this case is the desired outcome — the tooltip now agrees with what the button already did.

Rather than leave the two conditions duplicated (which is how they drifted apart in the first place), both the tooltip and the radio now read a single `sentenceModeDisabled` computed once, so they cannot disagree again.

Verified in a running Bloom: on a page with text and no whole-text-box recording the radio is enabled and shows the ordinary sentence-mode tip; on a picture-only page exactly one tooltip appears, the "put the cursor in a box" one. The remaining branch — disabled because a whole-text-box recording exists — needs a microphone to reach, so it is called out for manual testing on the card.

Good catch flagging it for confirmation rather than assuming.

andrew-polk:
[Claude Opus 5 (1M context) from Polk's machine during preflight] Consulted Devin again on 2026-08-10, up to commit `2df10546d4491f2097772bd84a3bf28453822dc3` (the tooltip/radio alignment).

Devin found **no bugs**. It raised two **Investigate** flags:

- *"showDisabled change alters tooltip in sentence-mode-with-audio case"* — new, and a fair thing to ask about: it spotted the deliberate behavior change and asked for confirmation from the product owner rather than assuming. Confirmed as intended (it is the change Andrew asked for) on [its thread](https://github.com/BloomBooks/BloomDesktop/pull/8187#discussion_r3754174611), and resolved.
- *"No explanatory tooltip when there is audio but nothing recordable on the page"* — a repeat of the flag from the previous commit, already posted, answered and resolved [here](https://github.com/BloomBooks/BloomDesktop/pull/8187#discussion_r3753853742). Not posted a second time.

CI is green. Full front-end test suite green at this commit: 48 files passed / 1 skipped, 524 tests passed / 5 skipped.

JohnThomson (approved):
> @JohnThomson reviewed 1 file and all commit messages, and resolved 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8187)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on andrew-polk).

JohnThomson (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8187)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on andrew-polk).

PR #8190, opened 2026-08-10
Make PDFs work for the StoryWeaver ebook page sizes, and stop blaming memory (BL-16684)

andrew-polk:
Books at the StoryWeaver ebook page sizes could not be made into a PDF at all — which also blocked uploading, since uploading makes a PDF preview. And when it failed, Bloom told the user to turn on a setting that was removed years ago.

## The page sizes

`Ebook2x3Portrait` and `Ebook7x5Landscape` were added as page sizes in 6.4, but `BloomPdfMaker` (WebView2PdfMaker) was never taught about them. Any size Bloom does not special-case is passed straight through as `-s <name>`, and `GetPaperSize` threw for these two.

Both are now in WebView2PdfMaker's paper size list. They are the only Bloom page sizes declared in CSS pixels rather than a physical unit (deliberately, to match StoryWeaver's ePUB CSS), so they are converted at the CSS standard of 96px per inch. Like every other entry in that list they are stored as *portrait* dimensions; `Ebook7x5` is only ever offered as landscape, so Bloom passes `-O landscape` and WebView2 swaps them.

Verified against the generated PDFs' `MediaBox`:

| Page size | Expected | Actual |
| --- | --- | --- |
| `Ebook7x5` landscape | 719.3 × 517.5 pt | 719.04 × 516.96 |
| `Ebook2x3` portrait | 356.3 × 525.0 pt | 355.92 × 525.12 |
| `Device16x9` landscape (control) | 504 × 283.5 pt | 504 × 282.96 |

The sub-point deltas are WebView2 quantizing to 1/100 inch.

## The message

Whenever PDF generation failed, Bloom appended *"The book's images might have exceeded the amount of RAM memory available. Please turn on the "Use Less Memory" option…"*. That option was taken out of the UI in 5.5 (BL-11954), so for years we have been sending people to hunt for a setting that no longer exists — whatever the real cause was. That is exactly what happened on this report: the actual failure was the unknown page size, and the message sent the user looking for a memory option.

The advice is gone, along with the plumbing that only fed it. `UserPrefs.ReducePdfMemoryUse` had no way to be set once the checkbox went away, so `PdfMakingSpecs.SaveMemoryMode` was always `false` and this message was its last remaining reader. Json.NET ignores the leftover `reducePdfMemory` key in existing `userPrefs` files, so old books are unaffected.

## Tests

`MakePdf_ScreenPageSize_OutputsPdf` covers both ebook sizes with `Device16x9` as a control. Sanity-checked: with the WebView2PdfMaker change reverted, the two ebook cases fail and the control passes.

`RunMakePdf` now returns the exception `MakePdf` hands back through `DoWorkEventArgs`, so a failure reports *why* it failed rather than just "no PDF file appeared".

## Note on the branch

Targets `Version6.4` rather than `master`: the ebook page sizes landed on `Version6.4` in `1c8e24b50d` (2026-06-25), so the broken PDF ships in released 6.4, not only in the 6.5 alpha the bug was reported against.

Ref: https://issues.bloomlibrary.org/youtrack/issue/BL-16684

[Devin review](https://devinreview.com/BloomBooks/BloomDesktop/pull/8190)

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8190)

### DIFF
--- a/DistFiles/localization/en/Bloom.xlf
+++ b/DistFiles/localization/en/Bloom.xlf
@@ -4309,11 +4309,6 @@
         <note>ID: PublishTab.PDF.Error.NoPrinter</note>
         <note>Error message displayed in a message dialog box.</note>
       </trans-unit>
-      <trans-unit id="PublishTab.PDF.Error.TrySinglePage">
-        <source xml:lang="en">The book's images might have exceeded the amount of RAM memory available. Please turn on the "Use Less Memory" option which is slower but uses less memory.</source>
-        <note>ID: PublishTab.PDF.Error.TrySinglePage</note>
-        <note>Error message displayed in a message dialog box.</note>
-      </trans-unit>
       <trans-unit id="PublishTab.PdfMaker.BadPdf">
         <source xml:lang="en">Bloom had a problem making a PDF of this book. You may need technical help or to contact the developers. But here are some things you can try:</source>
         <note>ID: PublishTab.PdfMaker.BadPdf</note>
--- a/src/BloomExe/Publish/PDF/MakePdfUsingExternalPdfMakerProgram.cs
+++ b/src/BloomExe/Publish/PDF/MakePdfUsingExternalPdfMakerProgram.cs
@@ -204,27 +204,24 @@ namespace Bloom.Publish.PDF
                     @"***ERROR PDF generation failed: res.StandardError = " + res.StandardError
                 );
 
-                var msg = L10NSharp.LocalizationManager.GetString(
-                    @"PublishTab.PDF.Error.Failed",
-                    "Bloom was not able to create the PDF file ({0}).{1}{1}Details: BloomPdfMaker (command line) did not produce the expected document.",
-                    @"Error message displayed in a message dialog box. {0} is the filename, {1} is a newline character."
-                );
+                var fullMsg = GetUnsupportedPageSizeMessage(specs, res.StandardError);
+                if (fullMsg == null)
+                {
+                    var msg = L10NSharp.LocalizationManager.GetString(
+                        @"PublishTab.PDF.Error.Failed",
+                        "Bloom was not able to create the PDF file ({0}).{1}{1}Details: BloomPdfMaker (command line) did not produce the expected document.",
+                        @"Error message displayed in a message dialog box. {0} is the filename, {1} is a newline character."
+                    );
 
-                // This message string is intentionally separate because it was added after the previous string had already been localized in most languages.
-                // It's not useful to add if we're already in save memory mode.
-                var msg2 = specs.SaveMemoryMode
-                    ? ""
-                    : L10NSharp.LocalizationManager.GetString(
-                        @"PublishTab.PDF.Error.TrySinglePage",
-                        "The book's images might have exceeded the amount of RAM memory available. Please turn on the \"Use Less Memory\" option which is slower but uses less memory.",
-                        @"Error message displayed in a message dialog box"
-                    ) + Environment.NewLine;
-
-                var fullMsg =
-                    String.Format(msg, specs.OutputPdfPath, Environment.NewLine)
-                    + Environment.NewLine
-                    + msg2
-                    + res.StandardOutput;
+                    // We used to follow this with PublishTab.PDF.Error.TrySinglePage, which told the user
+                    // to turn on the "Use Less Memory" option. That option was taken out of the UI back in
+                    // 5.5 (BL-11954), so ever since then we have been sending people to hunt for a setting
+                    // that isn't there, whatever the real cause of the failure was (BL-16684).
+                    fullMsg =
+                        String.Format(msg, specs.OutputPdfPath, Environment.NewLine)
+                        + Environment.NewLine
+                        + res.StandardOutput;
+                }
 
                 var except = new ApplicationException(fullMsg);
                 // Note that if we're being run by a BackgroundWorker, it will catch the exception.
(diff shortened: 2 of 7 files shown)


DistFiles/localization/en/Bloom.xlf
```diff
@@ -4313,6 +4313,7 @@
         <source xml:lang="en">The book's images might have exceeded the amount of RAM memory available. Please turn on the "Use Less Memory" option which is slower but uses less memory.</source>
```
andrew-polk: [Devin] **Bug**: A retired translatable message was deleted outright instead of being marked obsolete

The English entry for the old "Use Less Memory" advice (`PublishTab.PDF.Error.TrySinglePage`) was removed from `en/Bloom.xlf` entirely, rather than kept and labelled as no longer used.

Why it matters: the translated language files still contain that id, so deleting it from the English catalog puts the catalog and the translations out of step — which is the situation the project's localization rules exist to prevent.

src/WebView2PdfMaker/Program.cs
```diff
@@ -261,6 +261,22 @@ private PaperSize GetPaperSize(string name)
...
+                new PaperSize(
+                    "ebook2x3",
+                    475 * kMillimetersPerCssPixel,
+                    700 * kMillimetersPerCssPixel
+                ),
+                new PaperSize(
+                    "ebook7x5",
+                    690 * kMillimetersPerCssPixel,
+                    959 * kMillimetersPerCssPixel
+                ),
```
andrew-polk, line 285: [Devin] **Investigate**: Full-bleed printing still rejects the new ebook page sizes

`GetFullBleedPageSize` has no case for `ebook2x3`/`ebook7x5`, so in principle a book at those sizes with full bleed enabled would hit the "not yet implemented" error. Devin asked us to confirm the UI actually prevents full bleed for these sizes, so the fix is complete for every publish path.

DistFiles/localization/en/Bloom.xlf
andrew-polk: [Claude Opus 5] Not an issue here — deleting it is what this file's own practice calls for, so we're leaving it deleted (`e4e17b0d4d3525233a9ad74119b4ec0cea7a20fb`).

Devin has the general convention right: a retired string normally stays in `en/Bloom.xlf` with an "obsolete as of …" note. But that note records *when* the string died, and this one died in **5.5**, when the "Use Less Memory" option was taken out of the UI (BL-11954) — not now.

And the catalog only keeps recently-retired entries. Every surviving "obsolete as of" note in this file is 6.2, 6.3 or 6.4; everything from 6.1 and earlier has already been purged. A string that has been dead since 5.5 is well past that window, so it goes rather than getting a fresh obsolete label.

(I did briefly mark it `obsolete as of Bloom 6.4` acting on this finding, then reverted that — the label would have been wrong on the facts.)

src/WebView2PdfMaker/Program.cs
andrew-polk: [Claude Opus 5] Not an issue — it is structurally impossible to reach, and the guard is the very function this flag is about.

Whether a book can print full bleed at all is decided by `Book.FullBleed` (`Book.cs:4159`), which starts with `PageSizeSupportsFullBleed()` → `SizeAndOrientation.SupportsFullBleed()` (`SizeAndOrientation.cs:93`) → **`GetFullBleedPageSize(pageSize, landscape) != null`**.

So a page size `GetFullBleedPageSize` doesn't know is reported as not supporting full bleed, `Book.FullBleed` is false, `GetPrintingWithFullBleed()` is false, and `ConfigureFullBleedPageSize` is never called. That is a code-level guarantee rather than a UI convention, and it is the same reason `Device16x9` has been fine without a case for years.

In user terms: the full-bleed option simply isn't offered for the ebook/screen sizes, so there is no path to the error.

andrew-polk:
[Claude Opus 5] Consulted Devin on 2026-08-10 up to commit `e4e17b0d4d3525233a9ad74119b4ec0cea7a20fb`.

It raised **1 Bug** and **1 Investigate** flag, plus 4 informational notes (not mirrored). Both actionable items have their own thread above with the outcome recorded:

- *A retired translatable message was deleted outright instead of being marked obsolete* — assessed as not applying here. The string has been dead since 5.5, and this catalog only keeps entries retired in the last few versions.
- *Full-bleed printing still rejects the new ebook page sizes* — not reachable; the function it worries about is itself what decides a book can't use full bleed.

Two of Devin's informational notes independently checked the arithmetic in this PR and agreed: the px→mm conversion matches the LESS definitions, and the save-memory plumbing removal is complete.

CI is green (pr-automation, track/set-waiting). CodeRabbit has `auto_review.enabled: false` repo-wide, so it does not review on push.

src/BloomExe/Publish/PDF/MakePdfUsingExternalPdfMakerProgram.cs
```diff
@@ -245,6 +242,40 @@ DoWorkEventArgs doWorkEventArgs
...
+            // on it rather than on the English sentence around it so that rewording either end
+            // can't quietly turn this back into the generic message.
+            if (standardError == null || !standardError.Contains(kUnsupportedPageSizeMarker))
+                return null;
+
+            // Deliberately not localized: this can only appear if we have shipped a page size
+            // BloomPdfMaker can't render, which is a bug on our side that we expect to fix rather
+            // than translate. English is better than nothing for the rare user who hits it, and it
+            // is the form we want quoted back to us in the problem report.
+            return $"Bloom was not able to make a PDF of this book because its page size, \"{specs.PaperSizeName}\", is one that Bloom's PDF maker does not yet support. Please report this. You can work around it by changing the book's page size.";
```
andrew-polk: [Devin] **Investigate**: New message asks the user to "report this" but the dialog it appears in has no report option

The new unsupported-page-size message ended with "Please report this." But `PublishModel.ReportPdfGenerationError` routes `ApplicationException` to `ErrorReport.NotifyUserOfProblem(error.Message)` — the overload that passes no exception — so the user is asked to report a problem from a dialog that offers no way to do it.

andrew-polk: [Claude Opus 5] Confirmed and fixed — good catch, this was real.

Traced it: `HtmlErrorReporter` decides whether to show its Report button with `AllowSendReport.AllowIfException`, so the button only appears when an exception is passed alongside the message. `ReportPdfGenerationError` passes the message alone (deliberately — the comment there says including the original leads to reports we don't want, e.g. the PDF being open in Word). So there was indeed no Report button behind that sentence.

Rather than change how that whole class of error is routed, the message now points at the affordance that does exist:

> …is one that Bloom's PDF maker does not yet support. **Please tell us using "Report a Problem..." on the Help menu.** You can work around it by changing the book's page size.

That is also the route the user in BL-16684 actually took. A comment at the call site records why it names the Help menu rather than saying "report this", so the next person doesn't 'simplify' it back.

andrew-polk:
[Claude Opus 5] Consulted Devin on 2026-08-10 up to commit `7559d4131f07a0f5b2ebadcf4f02e529f1b71960`.

This supersedes the earlier log. Devin ran once per pushed commit, three times in all. Across the run it raised **1 Bug** and **2 Investigate** flags; each has its own thread above with the outcome recorded, and all are resolved:

- *A retired translatable message was deleted outright instead of being marked obsolete* — not applicable here. The string has been dead since 5.5, and this catalog only keeps entries retired in the last few versions.
- *Full-bleed printing still rejects the new ebook page sizes* — not reachable; the function it worries about is itself what decides a book can't use full bleed.
- *New message asks the user to "report this" but the dialog it appears in has no report option* — **a real catch, and fixed in `7559d4131f`.** The message now points at "Report a Problem..." on the Help menu. Devin's final pass still reports this one, but its text quotes the old wording, which is no longer in the code.

CI green throughout. CodeRabbit has `auto_review.enabled: false` repo-wide, so it does not review on push.

andrew-polk:
[Claude Opus 5] Consulted Devin on 2026-08-11 up to commit `928283f26cff6066784abbfa27ab3fe7e79030c5` (the squashed commit).

The branch was squashed from six commits to one and rebased onto the updated `Version6.4`. The tree is unchanged apart from picking up #8189's `AGENTS.md` additions, so this re-review is a re-run over the same code rather than a review of anything new.

It returned the same three findings, all of which already have resolved threads above:

- *A retired translatable message was deleted outright* — decided: the string has been dead since 5.5 and this catalog only keeps recently-retired entries.
- *Full-bleed at the new ebook sizes still throws a developer-flavored error* (reworded from its earlier phrasing, same mechanism) — not reachable: `Book.FullBleed` is gated on `GetFullBleedPageSize(...) != null`, the very function in question, so full bleed is never offered for a size it doesn't know.
- *New message asks the user to "report this"* — fixed earlier in the run; Devin's text still quotes the old wording, which is no longer in the code.

CI green on the squashed commit. No unresolved threads.

StephenMcConnel (approved):
> @StephenMcConnel reviewed 7 files and all commit messages, and resolved 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8190)* status: all files reviewed, 1 unresolved discussion (waiting on andrew-polk).

StephenMcConnel (approved):
:lgtm:

> @StephenMcConnel made 1 comment and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8190#-:-OzrmNnZ1IS_DHk8E_gR:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on andrew-polk).

PR #8191, opened 2026-08-11
Fix BL-16682 Save page before invoking Bloom AI Image Tools

andrew-polk:
Fixes BL-16682: "Edit with AI…" opened the AI Image Tools with an empty "Image to Edit" slot when the user had just added the image.

Ref: https://issues.bloomlibrary.org/youtrack/issue/BL-16682

## Why it happened

Everything Bloom tells the AI editor about the book — the whole-book image list, and on commit each slot's current src — is read from the **saved** book DOM. An image the user has just added lives only in the live page, because `changeImage`/`changeImageByElement` deliberately don't save (BL-16330). So the clicked image matched nothing in the list and no edit target was sent.

Saving first also closes two latent hazards on the same path: a current-page commit result would describe an image the live page no longer showed, and `DeleteSupersededAiImageFiles`, which judges orphanhood from the saved DOM, could delete a file the live page still used.

## What this does

The menu command posts `aiImageEditor/saveThenLaunch`; C# saves the page with `EditingModel.SaveThen` and then opens the editor. Because saving always ends in a navigation, two decisions follow, each for its own reason:

**Where the overlay lives — the top window.** An ordinary page save replaces the page iframe, so an overlay hosted there would be torn down. This is what the three other image commands in the same menu already do; the comment on `missingMetadata` in `canvasControlRegistry.ts` says so outright. `aiEditorLauncher.ts` therefore splits along the frame boundary:

- **`aiEditorOverlay.ts`** (top window, on `workspaceBundle`) — the overlay, the postMessage handshake with the editor's iframe, and the commit.
- **`aiEditorPageCommands.ts`** (page frame, on `editablePageBundle`) — the only two things that must run where the live page is: the menu command, and applying a commit's current-page swaps.
- **`aiEditorShared.ts`** — the types and the predicates both halves must agree on.

**When it opens — after the page has come back**, via the new `EditingModel.RunAfterNextPageLoad`. `SaveThen`'s existing `doAfterSaveToDisk` looks like the right hook, but it runs immediately *before* the navigation, and that navigation is not always confined to the page iframe: `EditingView.StartNavigationToEditPage` reloads the whole workspace root once `MemoryUtils.SystemIsShortOfMemory()` — Bloom's own private bytes past ~2GB, i.e. the ordinary state of a long editing session on a big book. Opening from `doAfterSaveToDisk` there meant the page saved correctly and the editor never appeared, with no message.

Consequences worth knowing:

- The request to open is queued up front, but acted on only if the save reached disk (`doAfterSaveToDisk`) or was never attempted (`doIfNotInRightStateToSave`). Queuing up front matters because `OnTabAboutToChange` discards the queue when the user leaves the Edit tab — queuing later let that discard run first on an in-flight save, so the editor sprang open on return. Gating on the outcome is what stops a **failed** save from opening the editor on a book DOM that is still stale; `failureAction` would not do, since it isn't called when `_saveBook()` itself throws (the disk-full case) nor on the deliberate discard path.
- Opening the overlay waits for `window.workspaceBundle` rather than assuming it. C# is called the instant the page iframe reports loaded, and on the whole-workspace-reload route the root document loads in parallel, assigning that global at the very end of `workspaceRoot.ts`; calling straight in would throw inside a fire-and-forget script, i.e. silently do nothing on precisely the route this design exists to survive.
- **The deferred-save workaround in the commit handler is gone.** It existed only because saving reloaded the frame owning the overlay's close button. That deferral also meant a second commit in one session read its `oldSrc` from a still-stale saved page and matched nothing — so this fixes that too.
- The page frame is asked to apply current-page swaps only when the commit actually contains one, so an all-off-page commit is no longer failed just because that frame happens to be mid-reload.

## Commits

1. **`ca0f97d`** — the fix.
2. **`2814f90`** — moves the AI editor front end out of `bookEdit/toolbox/canvas/` (whose own `AGENTS.md` describes it as canvas UI and utilities) into `bookEdit/aiImageEditor/`, beside `copyrightAndLicense/`. Pure relocation plus import paths — `git show -M` shows 11 renames, four with no content change — plus an `AGENTS.md` recording the frame rule, which bundling does not enforce on its own.

A smaller alternative that kept the overlay in the page iframe was written and compared as #8193, and closed in favour of this one.

## Verification

- Front-end: full `vitest run --no-file-parallelism` — **687 passed, 5 skipped, 62 files**. The launcher's single suite became two, tracking the split; `aiEditorOverlay.test.ts` needs no page DOM at all.
- C#: full `BloomTests` suite — **3063 passed, 0 failed, 12 skipped**.
- `pnpm typecheck`, eslint, and the isolated production bundle build are clean.
- The 5 `bloom-exe` AI-editor Playwright tests pass against a running Bloom.
- **Verified live**, including the awkward case: with `ShouldDoFullReload()` forced true — the state that made the first attempt at this fix silently show no editor — the overlay appears.

---
[Devin review](https://app.devin.ai/review/BloomBooks/BloomDesktop/pull/8191)

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8191)

[Devin review](https://devinreview.com/BloomBooks/BloomDesktop/pull/8191)


src/BloomBrowserUI/bookEdit/toolbox/canvas/aiEditorPageCommands.ts
```diff
@@ -0,0 +1,116 @@
...
+// change that page itself (this live browser owns it), so it returns oldSrc/newSrc and we
+// use Bloom's changeImageByElement() here. Returns how many swaps landed and how many were
+// asked for, so the caller can report a partial failure honestly.
+//
+// Called from the top window, so it must not assume anything about who is calling: `results`
+// has crossed a frame boundary but is plain data, and everything it touches is this
+// document.
+export const applyAiImageEditorReplacements = (
+    results?: IAiImageEditorCommitResult[],
+): IAiImageEditorApplyOutcome => {
```
andrew-polk: [Devin] **Bug**: New helper functions use arrow-const style instead of the required function syntax

New top-level helper functions in the split-out modules are written as `const foo = () => …` (e.g. `applyAiImageEditorReplacements` at `src/BloomBrowserUI/bookEdit/toolbox/canvas/aiEditorPageCommands.ts:60`), but the front-end coding rules require plain `function` syntax for non-React functions.
Impact: Code drifts from the repository's stated TypeScript style, making the codebase less consistent for future maintainers.

<details>
<summary>Which declarations are affected and what the rule says</summary>

`src/BloomBrowserUI/AGENTS.md` states: "For functions, prefer typescript `function` syntax over `const foo = () =>` functions." The newly-added/moved declarations that use the arrow-const form are:

- `launchAiImageEditor` (`src/BloomBrowserUI/bookEdit/toolbox/canvas/aiEditorPageCommands.ts:38`)
- `asMessage` (`src/BloomBrowserUI/bookEdit/toolbox/canvas/aiEditorPageCommands.ts:50`)
- `applyAiImageEditorReplacements` (`src/BloomBrowserUI/bookEdit/toolbox/canvas/aiEditorPageCommands.ts:60`)
- `fileNameOf` (`src/BloomBrowserUI/bookEdit/toolbox/canvas/aiEditorShared.ts:48`)
- `openAiImageEditor` (`src/BloomBrowserUI/bookEdit/toolbox/canvas/aiEditorOverlay.ts:43`)

None of these are React components, so the "always use arrow functions" rule for React does not apply.

</details>

src/BloomExe/web/controllers/AiImageEditorApi.cs
```diff
@@ -215,6 +226,90 @@ private string GetAiImageEditorUrl()
...
+                {
+                    // Nothing got saved, so the book DOM we are about to enumerate may still be
+                    // missing an image the user just added. Open anyway rather than let the menu
+                    // item do nothing: the user gets the old behavior (perhaps an empty "Image to
+                    // Edit" slot) rather than no response at all. Reaching here needs a save
+                    // already in flight or a navigation under way, so it is rare.
+                    OpenEditorInBrowser(payload);
+                },
+                doAfterSaveToDisk: () => OpenEditorInBrowser(payload)
+            );
```
andrew-polk, line 341: [Devin] **Investigate**: The menu item can silently do nothing if the save fails

`HandleSaveThenLaunch` only opens the overlay from `doAfterSaveToDisk` (or from `doIfNotInRightStateToSave`). In `EditingStateMachine.DoPostSaveAction`, `doAfterSaveToDisk` is skipped when the in-flight save is being discarded (`_discardInFlightSave`), and it is never reached if `_updateBookWithPageContents` throws for a page id that already failed once (the error is swallowed and... actually it continues) or if `_saveBook()` throws (the `finally` navigates but the action was already skipped). No `failureAction` is supplied here, so in those paths the user clicks "Edit with AI…" and nothing at all happens, with no message tied to the AI editor. Supplying `failureAction` (even just to open the editor anyway, as the not-in-right-state path does) would make the command's failure mode consistent.

src/BloomBrowserUI/bookEdit/aiImageEditor/aiEditorOverlay.ts
```diff
@@ -370,38 +258,46 @@ export const launchAiImageEditor = (
...
+                                if (!pageFrame) {
+                                    throw new Error(
+                                        "the page being edited is not available",
+                                    );
+                                }
+                                const cp =
+                                    pageFrame.applyAiImageEditorReplacements(
+                                        result?.results,
+                                    );
+                                currentPageApplied = cp.applied;
```
andrew-polk, line 305: [Devin] **Investigate**: A commit landing while the page frame is reloading fails the whole commit, including work C# already persisted

When `getEditablePageBundleExports()` returns null the overlay throws and acks failure. Note the asymmetry: by that point `aiImageEditor/commit` has already applied and saved every off-page replacement in C#, and the new image files exist in the book folder — only the current-page swaps are lost. The editor is told the whole commit failed, so a retry will re-apply the off-page slots (whose `oldSrc` now no longer matches, likely producing further partial failures). Since the PR makes the save fire immediately after each commit, the page-frame-reloading window is now reachable in normal use (commit, read the error, immediately re-commit). Worth confirming the editor's retry behaviour against a partially-applied commit.

.github/skills/bloom-automation/ai-image-editor-driving.md
```diff
@@ -59,7 +59,17 @@ shell page  (top, url .../bloom/…Temp/bloomXXXX.htm)
...
+
+**The overlay does not appear on the same tick as the click.** The menu command posts
+`aiImageEditor/saveThenLaunch`, which makes C# save the current page — reloading the `page`
+frame — and only then call `workspaceBundle.openAiImageEditor(...)` in the shell (BL-16682; the
+saved book DOM has to be current or the editor opens with an empty "Image to Edit" slot). So
+**wait for the overlay**, e.g. `await page.waitForSelector("#ai-editor-overlay iframe")`, and
+don't re-use any content-frame handle taken before the click — that frame is gone.
+
+Note also that the overlay itself belongs to the **shell**, not the `page` frame, so it survives
+page reloads: a commit saves the page immediately and the ✕ keeps working across that reload.
```
andrew-polk, line 73: [Devin] **Investigate**: The CDP driver script may still assume the overlay appears synchronously after the click

The skill doc was updated to say the overlay no longer appears on the same tick as the click and that content-frame handles taken before the click are stale, but `.github/skills/bloom-automation/driveAiImageEditor.mjs` (the reusable driver the doc points at) was not touched in this PR. It resolves `contentFrame(page)` before driving the context menu and then waits for the editor frame; whether it re-resolves the `page` frame after the save-triggered reload should be checked, or the documented recipe and the script will disagree.

src/BloomBrowserUI/bookEdit/toolbox/canvas/aiEditorPageCommands.ts
andrew-polk: [Claude Opus 5 (1M context)] Fixed in `bcab9a5`. Correct catch — `src/BloomBrowserUI/AGENTS.md` does ask for `function` syntax over `const foo = () =>` for non-React functions, and all five of those top-level helpers were arrow-consts (inherited from the single file they were split out of, which had the same style). They are now `function` declarations. Nested callbacks inside them stay arrows, which is idiomatic and what the rule intends.

src/BloomExe/web/controllers/AiImageEditorApi.cs
andrew-polk: [Claude Opus 5 (1M context)] Fixed in `bcab9a5` — a real hole, thank you. `doAfterSaveToDisk` is only reached when the save succeeds, so a save that was attempted and failed left the user having clicked "Edit with AI..." with nothing happening at all and no message naming the editor. `failureAction` is now supplied too, doing the same thing as the not-in-a-state-to-save path (open the editor anyway, on the possibly-stale book, rather than no response).

Worth recording what this does *not* cover, since you named three paths: `failureAction` fires on the page-content merge failure. It is not called when `_saveBook()` itself throws, and deliberately nothing fires on the `_discardInFlightSave` path — there Bloom is intentionally throwing the user's unsaved page away because an external process rewrote the book on disk, and opening an image editor on top of that would be the wrong response.

.github/skills/bloom-automation/ai-image-editor-driving.md
andrew-polk: [Claude Opus 5 (1M context)] Checked, and no change needed — though the doc was worth updating, so thanks for pointing at the pair.

`driveAiImageEditor.mjs` already tolerates the delayed overlay: `openEditorOverlay` polls for the editor frame up to 40 x 500ms (~20s) after clicking the menu item, and it re-resolves the `page` frame on every `dumpImages` call rather than holding a handle across the reload.

On the post-commit timing, the premise doesn't quite hold: the save did *not* become immediate for the path this script drives. Before this PR, a **successful** commit called `cleanup()`, and `cleanup()` was what fired `saveChangesAndRethinkPageEvent` — so the save already happened right after a successful commit, which is why the script's existing `waitForTimeout(2500)` comment already says "commit + (current-page) save+rethink". What this PR changes is that the save also fires after a **partial** failure instead of waiting for the user to close the overlay — and the script never exercises a partial failure.

src/BloomBrowserUI/bookEdit/aiImageEditor/aiEditorOverlay.ts
andrew-polk: [Claude Opus 5 (1M context)] Partly addressed in `bcab9a5`, and the rest is going to the developer as a decision — leaving this thread open until it comes back.

What I changed: when the page frame is unreachable, the message now says the replacements on the *other* pages were made and are saved, rather than just reporting a failure. A bare "the commit failed" is what invites the blind retry you describe.

What I did not decide unilaterally, because it is a design question about the editor's protocol rather than a bug in this diff: whether a partially-applied commit should be retryable at all, and if so whether Bloom should tell the editor *which* slots already landed so a retry can skip them. As it stands a retry re-applies the off-page slots against an `oldSrc` that no longer matches, which both fails again and leaves orphaned `ai-image*` files behind.

One correction to the framing, though. This window did not become newly reachable "in normal use" — a **successful** commit already saved immediately before this PR, because `cleanup()` fired `saveChangesAndRethinkPageEvent` and a successful commit called `cleanup()`. What this PR newly does is save after a *partial failure* too, instead of holding it until the user closes the overlay. So the sequence that reaches this window is: partial failure, then re-commit without closing — narrower than commit-then-recommit generally, but real.

src/BloomExe/web/controllers/AiImageEditorApi.cs
```diff
@@ -215,6 +226,95 @@ private string GetAiImageEditorUrl()
...
+                    // already in flight or a navigation under way, so it is rare.
+                    OpenEditorInBrowser(payload);
+                },
+                // Same reasoning for a save that is attempted and fails: doAfterSaveToDisk is
+                // only reached on success, so without this the user would click "Edit with AI…"
+                // and get nothing at all — no editor and no message naming it. They have already
+                // been told the save failed by the time we get here.
+                failureAction: () => OpenEditorInBrowser(payload),
+                doAfterSaveToDisk: () => OpenEditorInBrowser(payload)
+            );
```
andrew-polk, line 341: [Devin] **Investigate**: Overlay is opened before the post-save navigation, relying on the shell surviving a page reload

`doAfterSaveToDisk` runs inside `DoPostSaveAction` *before* `ToNavigating(pageId)` (`src/BloomExe/Edit/EditingStateMachine.cs:290-303`), so `OpenEditorInBrowser` fires the `workspaceBundle.openAiImageEditor(...)` script while the page navigation is still pending. That is safe only if the navigation reloads the `page` iframe alone and never the workspace root document (the PR description asserts `switchContentPage` only sets the iframe `src` plus a `history.replaceState`). If any navigation path ever re-navigates the whole WebView2 document, the overlay created here — plus its message listener and `__bloomAiImageEditorCleanup` — would be silently destroyed just after being created, leaving the user with no editor and no message. Worth a live-run confirmation, which the author notes has not yet been done for this variant.

andrew-polk: [Claude Opus 5 (1M context)] Correct on both counts, and this is the right thing to have singled out — leaving this thread open, because the live confirmation you're asking for is genuinely still outstanding.

Verified at the code level: the post-save navigation goes through `workspaceRoot.switchContentPage`, which sets `iframe.src` on the `page` frame only, and the shell's own URL only changes via `window.history.replaceState` — no shell reload. That is also why the image-gallery and copyright/license dialogs live in the shell and survive a metadata save, which is the precedent this PR is following.

But "I read the navigation path" is not "I saw it work". The developer has confirmed the *other* commit on this branch live; this variant has not been exercised in a running Bloom, and this exact call — C# invoking `workspaceBundle.openAiImageEditor(...)` while a page navigation is pending — is the one step no unit test covers. It is the top item in the preflight report going to the developer.

andrew-polk:
[Claude Opus 5 (1M context) from Andrew Polk's machine during preflight]

Consulted Devin twice on 2026-08-10, up to commit `bcab9a5bfb647b6c9af7994d6d702f90c0f256b3`.

**First pass** (`fcec7b2`) raised 1 Bug and 3 Investigate flags. Three are dealt with and their threads are resolved:

- the new helper functions used `const foo = () =>` where `src/BloomBrowserUI/AGENTS.md` asks for `function` syntax — a fair catch, fixed;
- **the real one**: "Edit with AI…" could silently do nothing. The editor was only opened once the save had succeeded, so a save that was attempted and *failed* left the user with no editor and no message. Fixed by also opening it on the failure path;
- it suspected the `driveAiImageEditor.mjs` CDP helper still assumed the overlay appears instantly after the click. Checked and it does not need changing — it already waits up to 20s for the editor, and a successful commit already saved immediately before this PR. Reasoning is on that thread.

**Second pass** (`bcab9a5`) raised one new flag worth keeping: the overlay is created *before* the post-save page navigation, so the design depends on that navigation reloading only the `page` iframe and never the whole window. That holds when you read the navigation code, but it has not been confirmed in a running Bloom for this variant, so that thread is left **open**.

Two threads are deliberately still open for the developer: the one just described, and the question of what should happen when someone re-commits after a partially-applied commit (today's retry redoes the already-saved off-page slots and orphans their files). Both are in the preflight decision report.

Devin also raised 6 informational observations, not mirrored here as threads. Two were worth a second look and neither needs action: `saveThenLaunch` runs with `requiresSync:false`, which is fine because it only *starts* a save — the work itself lands in `editView/pageContent`, which does take the lock; and its "the editor could open twice in rare save paths" idea doesn't reproduce, because the failure path returns before the after-save action can run.

Note that its second pass re-reported the arrow-const Bug and the silent-failure flag with their **original pre-fix text** (still describing `applyAiImageEditorReplacements` as an arrow-const, and saying no `failureAction` is supplied when it is, at `AiImageEditorApi.cs:298`). Those are re-review staleness, not new findings; the fixes are in `bcab9a5` and the threads stay resolved.

CI: `pr-automation` passed. No other review bot runs on this repo — CodeRabbit is configured but disabled (`auto_review.enabled: false`).

andrew-polk:
[Claude Opus 5 (1M context)] This branch has been rewritten as a single self-contained commit (`f9cd8ba`) so it no longer stacks on approach A's commit — the two approaches are now independently comparable, with approach A at #8193. The tree is byte-identical to what was reviewed as `bcab9a5`; only the history changed. The Devin threads above anchor to the pre-rewrite commits and so show as outdated, but every finding and its outcome still stands as recorded.

src/BloomExe/Edit/EditingModel.cs
```diff
@@ -2128,13 +2133,61 @@ public UrlPathString AddWidgetFilesToBookFolder(string fullWidgetPath)
...
+            if (nowEditing)
+            {
+                // Run whatever was queued for "the browser has a page again" (see
+                // RunAfterNextPageLoad). Taken and cleared before invoking, so it fires at most
+                // once even if it throws, and so an action that queues another one works.
+                // Before AdvanceUpdatingAllPages, which may navigate straight off this page.
+                var afterPageLoad = _doAfterNextPageLoad;
+                _doAfterNextPageLoad = null;
+                afterPageLoad?.Invoke(pageId);
+            }
```
andrew-polk, line 2145: [Devin] **Bug**: A queued request to open the AI image editor can linger and pop the editor open later unexpectedly

The pending request to open the AI image editor is only discarded when the page finishing loading is accepted as the page now being edited (`if (nowEditing)` at `src/BloomExe/Edit/EditingModel.cs:2136-2145`), so a page load that is not accepted leaves the request queued and it can spring to life on a later load of the same page.
Impact: The AI image editor can suddenly open by itself, long after the user's click, while the user is doing something else on that page.

<details>
<summary>How the queued action survives an ignored page-load report</summary>

`HandleSaveThenLaunch` (`src/BloomExe/web/controllers/AiImageEditorApi.cs:294-300`) queues the launch via `RunAfterNextPageLoad` before starting the save. `HandlePageDomLoadedEvent` only takes-and-clears `_doAfterNextPageLoad` when `_stateMachine.ToEditing(pageId)` returns true. `EditingStateMachine.ToEditing` returns false whenever the current state is not `Navigating` (or the reported page id doesn't match the one being navigated to) — see `src/BloomExe/Edit/EditingStateMachine.cs:195-216`. In that case the queued action is neither run nor cleared, and it stays armed until some later successful page load. Because the queued lambda only checks `loadedPageId == pageId`, a later return to the same page will open the overlay. Only leaving the Edit tab clears it (`src/BloomExe/Edit/EditingModel.cs:387-391`).

Clearing the slot unconditionally at the top of `HandlePageDomLoadedEvent` (or clearing it whenever `ToEditing` is refused) would make "next page load" mean exactly that.

</details>

andrew-polk: [Claude Opus 5 (1M context)] Correct, and worth keeping open — but the fix you suggest has a real cost, so this is going to the developer rather than being changed here.

The `nowEditing` gate cuts both ways, and the two failure modes are mutually exclusive:

- **Keep the gate (today):** an ignored page-load report leaves the launch armed, so a later return to that same page can open the editor unprompted. Confusing, but recoverable — and the user did click the command at some point.
- **Clear unconditionally (your suggestion):** "next page load" means exactly that, but an intermediate or ignored load — an `about:blank` while the workspace root reloads, say — consumes and discards the launch, and the user gets *nothing at all* with no message. That is precisely the silent-dead-end this PR has already been fixed for twice (once for `doAfterSaveToDisk` running before the navigation, once for a missing failure path), so I'm not trading back into it without the developer's call.

Worth noting explicitly: this is not a regression from either design under review. `RunAfterNextPageLoad` is now shared by both #8191 and #8193, so whatever we decide applies to both.

A third option I considered and rejected as too risky to do blind: clear the queued launch when the *page selection* changes to a different page, which is semantically exactly right ("my launch was for page X, the user has moved to page Y"). The hazard is that the post-save re-navigation selects the *same* page, and if `SelectionChanged` fires for that too, it would drop the launch we are waiting on — reintroducing failure mode 2. That needs checking against the selection-change plumbing before anyone relies on it.

andrew-polk:
[Claude Opus 5 (1M context)] Squashed to a single commit `b3b9311` at Andrew's request. The tree is **byte-identical** to `8c4c4b7`, which is what Devin reviewed and what both full suites ran against (687 front-end, 3063 C#) — only the history changed. The Devin threads above therefore anchor to commits no longer in the branch and show as outdated; every finding and its outcome still stands as recorded, and the two threads left open are still open.

andrew-polk:
[Claude Opus 5 (1M context) from Andrew Polk's machine during preflight]

Consulted Devin five times across this branch, most recently up to commit `b3b9311d581fbc315a1fc8bd79f096303ec24540`.

What it found that was real, and is fixed:

- **"Edit with AI…" could silently do nothing.** The editor was only opened once the save had succeeded, so a save that was attempted and *failed* left the user with no editor and no message.
- **A wholly successful commit could be reported as an error.** The overlay asked the page frame for its exports before checking whether the commit had anything to do on the open page, and that frame is briefly unreachable while it reloads — a window this feature's own post-commit save opens. Users would have seen a failure for pictures that had in fact been replaced and saved.
- **An over-defensive payload parse**, against `AGENTS.md`'s explicit "don't be overly defensive / let it throw" rule.
- **A duplicated, self-contradicting paragraph** in the CDP driving guide.
- **The PR description no longer matched the code** — the comparison table still claimed this approach left `EditingModel` untouched. That one mattered: someone could have chosen between #8191 and #8193 on a table that was wrong about the main difference between them. Corrected on the PR.

Two threads are deliberately left **open** for the developer, both design questions rather than defects, and both applying to #8193 equally now that the two approaches share the `RunAfterNextPageLoad` hook:

- what a retry should do after a partially-applied commit (today it redoes the already-saved off-page slots and orphans their files);
- whether a queued launch that is never consumed should be dropped. Devin's suggested fix trades that surprise back for the silent-do-nothing failure fixed above, so it needs a human call.

A note for anyone reading Devin's own pages here: **its later passes re-report findings that were already fixed, using their original pre-fix text** — still describing helpers as arrow functions after they were converted, still saying no `failureAction` is supplied when the code deliberately no longer needs one, still describing the off-page-commit bug after it was fixed. Each was checked against the working tree rather than taken at face value; the resolved threads stay resolved. Treat a re-review's findings as describing the diff, not necessarily the current code.

CI: `pr-automation` passed. No other review bot runs on this repo (CodeRabbit is configured but disabled via `auto_review.enabled: false`).

src/BloomExe/web/controllers/AiImageEditorApi.cs
andrew-polk: [Claude Opus 5 (1M context)] Fixed in `b3b9311`, and you were right to press on it — this is the finding that changed the design.

The overlay is no longer created in `doAfterSaveToDisk`. It is opened once the browser reports a page again, via `EditingModel.RunAfterNextPageLoad`, so it no longer depends on the shell surviving the post-save navigation at all.

Which matters, because the assumption you asked me to confirm turned out to be false. `EditingView.StartNavigationToEditPage` has a third route that reloads the whole workspace root, taken once `MemoryUtils.SystemIsShortOfMemory()` — Bloom's own private bytes past ~2GB, i.e. the ordinary state of a long editing session on a big book, and exactly what that reload exists to recover from. In that state the page saved correctly and the editor never appeared, silently: `openAiImageEditor` doesn't build the overlay synchronously, it POSTs launch first and builds it in the reply, a whole round trip after the reload began, in a document being replaced.

Resolving this one. The related question of what should happen to a queued launch that is never consumed is tracked separately and is still open.

src/BloomBrowserUI/bookEdit/aiImageEditor/aiEditorOverlay.ts
andrew-polk: [Claude Opus 5 (1M context)] **Developer decision: leave as is.** Andrew has reviewed this and chosen not to act on it now.

The reasoning, for the record: this is not a defect in the diff but a gap in the agreement between Bloom and the separate `bloom-ai-image-tools` app about what a retry means, and closing it properly would mean changing the commit reply in both repos. What has been done instead is to make the message honest — when the page being edited can't be reached, it now says the replacements on the other pages were made and are saved, which is what stops someone retrying blindly and duplicating that work.

Resolving. If the two-app fix is taken up later it should get its own card rather than riding on this one.

src/BloomExe/Edit/EditingModel.cs
andrew-polk: [Claude Opus 5 (1M context)] **Developer decision: leave as is.** Andrew has reviewed this and chosen to accept the current behaviour.

The reasoning, for the record: the two candidate behaviours are mutually exclusive, and the one you propose trades this rare surprise for a rare *silent* failure — an ignored or in-between page load consuming the queued launch, so the menu item does nothing at all with no message. That failure mode has already been fixed twice on this card, and it is the worse of the two: a stray overlay is visible and closable, whereas nothing-happened leaves the user with no idea why.

Worth restating that this is not specific to the approach on this PR — `RunAfterNextPageLoad` is shared with the alternative implementation, so the behaviour would be identical either way.

Resolving.

andrew-polk (commented):
> @andrew-polk partially reviewed 12 files and all commit messages, and resolved 4 discussions.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8191)* status: all files reviewed, 2 unresolved discussions.

andrew-polk (commented):
> @andrew-polk partially reviewed 17 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8191)* status: all files reviewed, 2 unresolved discussions.

andrew-polk (commented):
> @andrew-polk reviewed 1 file and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8191)* status: all files reviewed, 2 unresolved discussions.

andrew-polk:
You probably want to review as two separate commits. The second one is just moving files.

JohnThomson (changes requested):
> @JohnThomson reviewed all commit messages, made 3 comments, and resolved 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8191)* status: all files reviewed, 3 unresolved discussions (waiting on andrew-polk).

___
*[a discussion](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8191#-OznGwNG-KaaywUlfqJh:-Ozrj00RECefAJwzpNln:bjon177) (no related file):*

You probably want to review as two separate commits. The second one is just moving files.
</blockquote></details>

I found the control for requesting this, but couldn't make it work. What it showed me at first was just the changes to editing model. Reviewing as a whole makes it a bit overwhelming, but I've tried to focus on changes that looked significant. Just a couple of questions.
___
*[`src/BloomExe/web/controllers/AiImageEditorApi.cs` line 299 at r7](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8191#-OzrhKfI05y-Mk20nqYz:-OzrhKfI05y-Mk20nqZ-:b-nj74yd) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/2814f90acb1e7c5839016d747b4d2934a0523db2/src/BloomExe/web/controllers/AiImageEditorApi.cs#L299)):*
> ```Smalltalk
> 
>             // Queue BEFORE the save: with no page loaded, SaveThen completes synchronously.
>             model.RunAfterNextPageLoad(loadedPageId =>
> ```

Things are probably just being defensive here, but the comments seem confused. A few lines above, we aborted completely if there isn't a current page being edited. Now the comment immediately above (and I think I've seen this in a few places) claims we're trying to continue even if no page is loaded.
___
*[`src/BloomExe/web/controllers/AiImageEditorApi.cs` line 318 at r7](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8191#-Ozri6am74Q_3yLupK9S:-Ozri6am74Q_3yLupK9T:b445fkz) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/2814f90acb1e7c5839016d747b4d2934a0523db2/src/BloomExe/web/controllers/AiImageEditorApi.cs#L318)):*
> ```Smalltalk
>             // save can fail still ends in EditingStateMachine navigating back to the page, so the
>             // queued launch fires anyway. That the hook covers the failure paths for free is one of
>             // the reasons for preferring it over doAfterSaveToDisk.
> ```

This is (hopefully) a very unlikely corner case, but are we sure we still WANT to open the AI image editor if something went so badly wrong that we can't even save the current page? I'm not sure what could cause such a failure, but things that come to mind are not enough disk space to update the HTML file, a badly corrupted image file, out of memory,...

andrew-polk:
[Claude Opus 5 (1M context) from Andrew Polk's machine during preflight]

Thanks John — both of your questions were right, and the second one changed the behaviour. Fixed in `e9179df` (force-pushed, so the PR is back in draft while it re-runs the checks).

**On the confusing comment at `AiImageEditorApi.cs:299`** — you read it exactly as written and it was wrong to write it that way. I had conflated two different things called "page": the guard a few lines up tests `model.CurrentPage`, the *book page selection*, whereas the comment meant `EditingStateMachine`'s `NoPage` state, i.e. no page loaded *in the browser* (the Edit tab hidden). Those are independent — you can have a `CurrentPage` while the machine sits in `NoPage`, and in that state `ToSavePending` runs synchronously, which was the real reason the request had to be made before `SaveThen`. Rather than reword it, the restructuring below removes the ordering subtlety altogether, so the comment is simply gone. (You wondered whether you'd seen the same thing elsewhere — I checked, and that was the only occurrence.)

**On whether we should open the editor at all after a save that failed** — you've convinced me, and for a reason that strengthens your case: if the page could not be saved then the saved book DOM is still stale, so the editor would open on exactly the out-of-date data this whole change exists to prevent, and a commit from there would call `book.Save()` again on top of whatever went wrong. Your examples are apt: disk full, a corrupt image, out of memory.

The earlier reasoning for opening anyway was to avoid a menu item that silently does nothing. That concern doesn't apply here, though: `EditingStateMachine` has already shown the user *"Bloom had trouble saving a page… Then quit Bloom, run it again"* before any of this, so they are told — they just shouldn't also get an image editor on top of it.

So the request to open is now made only from the two paths where the book DOM is sound:

- `doAfterSaveToDisk` — runs only after `_saveBook()` succeeded;
- `doIfNotInRightStateToSave` — no save was attempted, nothing failed, and a page load is coming from whatever refused it.

Every failing path reaches neither, so the editor stays closed. Worth noting that this is broader than the obvious fix: my first attempt was to cancel via `failureAction`, and checking it against your disk-full example showed it wouldn't have covered it — `failureAction` isn't called when `_saveBook()` itself throws. Inverting it to "only ask when we know the save landed" also covers the `_discardInFlightSave` path, where Bloom has deliberately thrown the page away and an image editor would have been the wrong response.

**On reviewing as two commits** — sorry that fought you. Noted that the tooling didn't cooperate; the split is still worth having in the history (the second commit is renames only, four files with no content change at all), but I won't claim it made your review easier this time.

andrew-polk (commented):
[Claude Opus 5 (1M context)] Sorry the two-commit split fought you rather than helping. Noted that the control didn't work — the second commit is renames only (11 files, four with no content change), so `git show -M` on it is the quickest way to skip past it if you look again.

Also my fault that you had to read one long summary instead of answers under the lines you asked about: I replied with a single top-level comment because your comments have no GitHub mirror and I had wrongly ruled out using the Reviewable CLI. Both of your code comments now have their own replies in place, so this thread doesn't need the detail.

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8191#-:-Ozs0rJRDvJvtOTiSPkc:bac0b0n)* status: 19 of 20 files reviewed, 3 unresolved discussions (waiting on andrew-polk and JohnThomson).

___
*[`src/BloomExe/web/controllers/AiImageEditorApi.cs` line 299 at r7](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8191#-OzrhKfI05y-Mk20nqYz:-Ozs0qY95axwD1t29DrL:bj8i00i) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/2814f90acb1e7c5839016d747b4d2934a0523db2/src/BloomExe/web/controllers/AiImageEditorApi.cs#L299)):*

Things are probably just being defensive here, but the comments seem confused. A few lines above, we aborted completely if there isn't a current page being edited. Now the comment immediately above (and I think I've seen this in a few places) claims we're trying to continue even if no page is loaded.
</blockquote></details>

[Claude Opus 5 (1M context)] You read it exactly as written, and it was wrong to write it that way. I had conflated two different things called "page": the guard a few lines above tests `model.CurrentPage`, the *book page selection*, whereas this comment meant `EditingStateMachine`'s `NoPage` state — no page loaded **in the browser**, e.g. the Edit tab hidden. Those are independent: you can have a `CurrentPage` while the machine sits in `NoPage`, and in that state `ToSavePending` runs `DoPostSaveAction` synchronously, which was the real reason the request had to be made before `SaveThen`.

Rather than reword it, the restructuring for your other comment removed the ordering subtlety altogether, so this comment is simply gone in `140387d`.

On your parenthetical — you wondered whether you'd seen the same thing in a few places. I grepped: that was the only occurrence in this change. The two other hits for similar wording are pre-existing and unrelated (`EditingStateMachine.cs` describing its own state, and a comment in `ImageApi.cs`).
___
*[`src/BloomExe/web/controllers/AiImageEditorApi.cs` line 318 at r7](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8191#-Ozri6am74Q_3yLupK9S:-Ozs0qw48SeIJ11xAwv2:b-rl6plb) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/2814f90acb1e7c5839016d747b4d2934a0523db2/src/BloomExe/web/controllers/AiImageEditorApi.cs#L318)):*

This is (hopefully) a very unlikely corner case, but are we sure we still WANT to open the AI image editor if something went so badly wrong that we can't even save the current page? I'm not sure what could cause such a failure, but things that come to mind are not enough disk space to update the HTML file, a badly corrupted image file, out of memory,...
</blockquote></details>

[Claude Opus 5 (1M context)] You've convinced me, and for a reason that strengthens your case: if the page could not be saved then the saved book DOM is still stale, so the editor would open on exactly the out-of-date data this whole change exists to prevent — and a commit from there would call `book.Save()` again on top of whatever went wrong. Your examples are apt.

The earlier reasoning for opening anyway was to avoid a menu item that silently does nothing. That doesn't apply here: `EditingStateMachine` has already shown the user *"Bloom had trouble saving a page… Then quit Bloom, run it again"* before this point, so they are told — they just shouldn't also get an image editor on top of it.

Fixed in `140387d`. The request to open is now queued up front but acted on only if a captured `bookDomIsSound` flag was set, from either `doAfterSaveToDisk` (the save reached disk) or `doIfNotInRightStateToSave` (no save was attempted, nothing failed, and a page load is coming anyway). Every failing path reaches neither, so the editor stays closed.

Two things I got wrong on the way there, in case the shape looks roundabout:

1. My first attempt was to cancel via `failureAction`. Checking it against your disk-full example showed it wouldn't have covered it — `failureAction` isn't called when `_saveBook()` itself throws, nor on the deliberate `_discardInFlightSave` path. Gating on "did the save land" covers all of them.
2. My second attempt only *queued* from those two callbacks rather than queueing up front and gating. Devin caught that this broke a different ordering: `OnTabAboutToChange` discards the queue when you leave the Edit tab, so on a save still in flight that discard ran first and the completing save re-armed behind it — the editor then sprang open when you returned to the page. Hence queue-early-but-gate.

JohnThomson (approved):
> @JohnThomson reviewed 1 file and all commit messages, and resolved 3 discussions.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8191)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on andrew-polk).

PR #8192, opened 2026-08-11
Fix BloomServer shutdown race that crashed the test host (BL-16667)

JohnThomson:
Running the full C# suite, roughly **one run in three** ended early — not with a test failure, but with the **test host process crashing**. Because everything that had run so far had passed, the run then printed `Passed! - Failed: 0, Passed: 348 ...` and anything judging it by that line called it green.

## The bug

Both crash faces in the card are the same lifetime problem. `RequestInfo` sent response bodies with `HttpListenerResponse.Close(buffer, willBlock: false)`. That returns as soon as the write is *issued* — which is the point, it frees the server worker thread — but the framework then finishes the send on an I/O completion callback of its own whose only handler is `catch (Win32Exception)`. `BloomServer.Dispose` joined the workers, saw nobody busy, and closed the `HttpListener` while such a send was still in flight. The send then failed with `ObjectDisposedException`, which that handler does not catch, on a thread none of our `try`/`catch` blocks cover — so it reached the runtime and took the process with it.

## The fix

Issue the same asynchronous write — same I/O completion port, same thread, no difference in when the data goes out — but supply the completion ourselves, in the new `PendingResponseWrites`. Owning that last step gives two things: a failure at the end of a send is ours to log and swallow, and we know when sends finish, so `Dispose` can wait for the ones in flight instead of closing the listener and hoping.

Three other things in the same area could also kill the process, and are fixed too:

- `RequestProcessorLoop`'s `Debug.Fail` now stays quiet during shutdown and under unit tests. There is no debugger to stop in a test host, so all it can do there is terminate the run; the error is still logged and the request still fails, so a real server bug still shows up as a failing test.
- `QueueRequest` is likewise an I/O completion callback, and was unguarded.
- `CloseListener`'s `Debug.Assert` asserted a condition the next line already handles.

## Making an aborted run report itself as a failure

Asked for alongside the fix, because the silent-truncation half of this bug is not specific to it:

- `build/agent-dotnet.{sh,ps1}` now judge a `test` run and print a verdict as their **last** line (so it survives `| tail`), exiting non-zero on crash markers, a `Failed!` summary, or a non-zero `dotnet` exit. The markers live in one shared `build/test-abort-markers.txt` so the twin scripts cannot drift. This also fixes the `PAPERCUTS.md` entry about the wrapper reporting a failing suite as a success, so that entry is removed.
- `build/Bloom.proj`'s `RunNUnit` now fails if the results file — which the logger writes only on completion — is missing, so CI cannot go green on a run that stopped early.

## Testing

`ServeFileAndShutDownTests.DisposingWhileAResponseIsStillGoingOut_DoesNotKillTheProcess` reproduces the bug **deterministically**: with the old `Close(buffer, false)` restored it kills the test host with exactly the stack from the card (`NonBlockingCloseCallback` → `EndWriteCore` → `Abort` → `ObjectDisposedException` on `HttpRequestQueueV2Handle`), and the new wrapper reported that run as `*** TEST RUN ABORTED ***`. It passes with the fix.

Also new: `PendingResponseWritesTests` (6 tests) for the counting, the timeout, and that neither a failed send nor a failed close can escape; and tests that fetch a real file, and a real 404, from a real listening server over real HTTP — which nothing covered before.

Full C# suite run four times end to end: all four completed, 0 failed.

[Devin review](https://devinreview.com/BloomBooks/BloomDesktop/pull/8192)

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8192)

### DIFF
--- a/build/agent-dotnet.ps1
+++ b/build/agent-dotnet.ps1
@@ -11,6 +11,10 @@
 # build and run unit tests while a Bloom keeps running and while other terminals do
 # their own builds.
 #
+# For `test` it also judges the run and prints a verdict as its last line -- see the
+# "Judging a test run" comment in agent-dotnet.sh for why that is not something the
+# exit code can be trusted to do on its own.
+#
 # Usage (exactly like dotnet, just build/test through this script):
 #   build/agent-dotnet.ps1 test src/BloomTests/BloomTests.csproj --filter "FullyQualifiedName~UrlPathStringTests"
 #   build/agent-dotnet.ps1 build src/BloomExe/BloomExe.csproj
@@ -32,5 +36,76 @@ Write-Host "[agent-dotnet] isolated build dir: $scratch"
 # the BloomPdfMaker.exe the PDF tests shell out to — so it now lives in
 # Directory.Build.props, which a global property would have overridden.
 $env:BLOOM_AGENT_BUILD_DIR = $scratch
-& dotnet @args "-p:OutDir=$scratch/bin/"
-exit $LASTEXITCODE
+
+# For everything except `test`, dotnet's own exit code is the whole story.
+if ($args.Count -eq 0 -or $args[0] -ne 'test') {
+    & dotnet @args "-p:OutDir=$scratch/bin/"
+    exit $LASTEXITCODE
+}
+
+$markersFile = Join-Path $PSScriptRoot 'test-abort-markers.txt'
+if (-not (Test-Path $markersFile)) {
+    Write-Host "[agent-dotnet] cannot find $markersFile, so a truncated test run could not be detected"
+    exit 1
+}
+
+$log = Join-Path ([System.IO.Path]::GetTempPath()) "agent-dotnet-test-$PID.log"
+
+# stderr is merged in because that is where the runtime prints the crash banners we look
+# for. In Windows PowerShell that redirection wraps each stderr line in an ErrorRecord,
+# which under $ErrorActionPreference='Stop' would end the script; hence Continue for the
+# duration, and the stringify step so what reaches the console and the log is plain text.
+$ErrorActionPreference = 'Continue'
+& dotnet @args "-p:OutDir=$scratch/bin/" 2>&1 |
+    ForEach-Object {
+        # An ErrorRecord here is just a line dotnet wrote to stderr. Take its message rather than
+        # letting PowerShell format it, which for some of them yields the useless string
+        # "System.Management.Automation.RemoteException" -- and a crash banner turned into that
+        # would be a marker we could no longer match.
+        if ($_ -is [System.Management.Automation.ErrorRecord]) { $_.Exception.Message } else { [string]$_ }
+    } | Tee-Object -FilePath $log
+$status = $LASTEXITCODE
+$ErrorActionPreference = 'Stop'
+
+try {
+    $abortEvidence = $null
+    foreach ($pattern in Get-Content $markersFile) {
+        if ($pattern.Trim() -eq '' -or $pattern.TrimStart().StartsWith('#')) { continue }
+        $hit = Select-String -Path $log -Pattern $pattern -List
+        if ($hit) { $abortEvidence = $hit.Line.Trim(); break }
+    }
+
+    $summaryHit = Select-String -Path $log -Pattern '^[ \t]*(Passed|Failed)!'
+    $summary = if ($summaryHit) { $summaryHit[-1].Line.Trim() } else { $null }
+
+    Write-Host ""
+    if ($abortEvidence) {
+        Write-Host "[agent-dotnet] *** TEST RUN ABORTED -- NOT a pass. ***"
+        Write-Host "[agent-dotnet] Evidence: $abortEvidence"
+        Write-Host "[agent-dotnet] Any summary above counts only the tests that ran before this, so it means nothing."
+        exit 1
+    }
+    if ($status -ne 0) {
+        $detail = if ($summary) { $summary } else { 'No summary line was printed.' }
+        Write-Host "[agent-dotnet] *** TEST RUN FAILED (dotnet exited $status). *** $detail"
+        exit $status
+    }
+    if (-not $summary) {
+        # A discovery-only run has no tests to summarize, so its silence is not evidence of anything.
+        if ($args -contains '--list-tests' -or $args -contains '-t') {
+            Write-Host "[agent-dotnet] listed tests without running them."
+            exit 0
+        }
+        Write-Host "[agent-dotnet] *** TEST RUN INCOMPLETE: it exited 0 but never printed a summary, so we do not know what ran. ***"
+        exit 1
+    }
+    if ($summary -like '*Failed!*') {
+        # dotnet said 0 but its own summary says otherwise; believe the summary.
+        Write-Host "[agent-dotnet] *** TEST RUN FAILED. *** $summary"
+        exit 1
+    }
+    Write-Host "[agent-dotnet] test run completed. $summary"
+}
+finally {
+    Remove-Item $log -ErrorAction SilentlyContinue
+}
--- a/build/agent-dotnet.ps1
+++ b/build/agent-dotnet.ps1
@@ -65,6 +65,12 @@ $ErrorActionPreference = 'Continue'
         if ($_ -is [System.Management.Automation.ErrorRecord]) { $_.Exception.Message } else { [string]$_ }
     } | Tee-Object -FilePath $log
 $status = $LASTEXITCODE
+# $LASTEXITCODE is never assigned if dotnet could not be launched at all -- not on PATH in this
+# terminal, say -- because then nothing ran to set it. That leaves $status $null, and PowerShell
+# turns `exit $null` into exit code 0: this script would announce a failed run and hand its caller
+# a success. Which is the exact hazard it exists to remove. (The bash twin cannot have this hole;
+# PIPESTATUS[0] is always a number.)
+if ($null -eq $status) { $status = 1 }
 $ErrorActionPreference = 'Stop'
 
 try {
--- a/build/agent-dotnet.ps1
+++ b/build/agent-dotnet.ps1
@@ -68,14 +68,20 @@ $status = $LASTEXITCODE
 $ErrorActionPreference = 'Stop'
 
 try {
+    # Match against the output with the indentation taken off the front of every line. That is what
+    # lets the markers file anchor its patterns with a plain ^ and mean the same thing here as it
+    # does to grep -E in the bash twin -- writing "start of line, then optional whitespace" is the
+    # one thing the two regex engines cannot be made to agree on (see the note in the markers file).
+    $trimmedOutput = Get-Content $log | ForEach-Object { $_.TrimStart() }
+
     $abortEvidence = $null
     foreach ($pattern in Get-Content $markersFile) {
         if ($pattern.Trim() -eq '' -or $pattern.TrimStart().StartsWith('#')) { continue }
-        $hit = Select-String -Path $log -Pattern $pattern -List
+        $hit = $trimmedOutput | Select-String -Pattern $pattern -List
         if ($hit) { $abortEvidence = $hit.Line.Trim(); break }
     }
 
-    $summaryHit = Select-String -Path $log -Pattern '^[ \t]*(Passed|Failed)!'
+    $summaryHit = $trimmedOutput | Select-String -Pattern '^(Passed|Failed)!'
     $summary = if ($summaryHit) { $summaryHit[-1].Line.Trim() } else { $null }
 
     Write-Host ""
--- a/build/agent-dotnet.ps1
+++ b/build/agent-dotnet.ps1
@@ -77,11 +77,15 @@ try {
     $abortEvidence = $null
     foreach ($pattern in Get-Content $markersFile) {
         if ($pattern.Trim() -eq '' -or $pattern.TrimStart().StartsWith('#')) { continue }
-        $hit = $trimmedOutput | Select-String -Pattern $pattern -List
+        # -CaseSensitive because grep -E in the bash twin is, and Select-String is not unless told.
+        # Without it the same output could be judged aborted here and passing there -- and the
+        # markers file spells "[Tt]esthost" out precisely because it expects to be matched with
+        # case respected.
+        $hit = $trimmedOutput | Select-String -Pattern $pattern -List -CaseSensitive
         if ($hit) { $abortEvidence = $hit.Line.Trim(); break }
     }
 
-    $summaryHit = $trimmedOutput | Select-String -Pattern '^(Passed|Failed)!'
+    $summaryHit = $trimmedOutput | Select-String -Pattern '^(Passed|Failed)!' -CaseSensitive
     $summary = if ($summaryHit) { $summaryHit[-1].Line.Trim() } else { $null }
 
     Write-Host ""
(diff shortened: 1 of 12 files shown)


build/agent-dotnet.ps1
```diff
@@ -32,5 +36,82 @@ Write-Host "[agent-dotnet] isolated build dir: $scratch"
...
+    $trimmedOutput = Get-Content $log | ForEach-Object { $_.TrimStart() }
+
+    $abortEvidence = $null
+    foreach ($pattern in Get-Content $markersFile) {
+        if ($pattern.Trim() -eq '' -or $pattern.TrimStart().StartsWith('#')) { continue }
+        $hit = $trimmedOutput | Select-String -Pattern $pattern -List
+        if ($hit) { $abortEvidence = $hit.Line.Trim(); break }
+    }
+
+    $summaryHit = $trimmedOutput | Select-String -Pattern '^(Passed|Failed)!'
```
JohnThomson: [Devin] **Bug**: Windows and Linux/macOS versions of the test wrapper can reach opposite verdicts on the same run

Devin spotted that the two halves of the test wrapper do not agree about letter case. PowerShell's `Select-String` ignores case unless you tell it not to; `grep -E`, which the bash twin uses, respects it. So a line the runtime wrote as `Unhandled Exception.` would be treated as a crash on Windows and as ordinary output everywhere else — the same test run called a failure in one shell and a pass in the other, which is exactly the drift the shared marker file exists to prevent.

Devin is right, and the marker file itself is the evidence: it spells one marker `[Tt]esthost` rather than relying on either engine to be lenient about case, so case-sensitive matching was always the intent.

build/agent-dotnet.sh
```diff
@@ -48,5 +51,98 @@ echo "[agent-dotnet] isolated build dir: $SCRATCH" >&2
...
+set +e
+BLOOM_AGENT_BUILD_DIR="$SCRATCH" dotnet "$@" -p:OutDir="$SCRATCH/bin/" 2>&1 | tee "$LOG"
+STATUS=${PIPESTATUS[0]}
+set -e
+
+# Match against a copy with the indentation taken off the front of every line. That is what lets
+# the markers file anchor its patterns with a plain ^ and mean the same thing here as it does to
+# Select-String in the PowerShell twin -- writing "start of line, then optional whitespace" is the
+# one thing the two regex engines cannot be made to agree on (see the note in the markers file).
+sed 's/^[[:space:]]*//' "$LOG" >"$TRIMMED"
```
JohnThomson, line 101: [Devin] **Investigate**: Trimming leading whitespace weakens the ^ anchors the markers file relies on

Both wrappers now strip the indentation off each line of output before looking for the crash markers. Devin's concern is that this lets a *test's own* output trip a marker: NUnit indents anything a test prints, so a test that printed `Unhandled exception. ...` on an indented line would, after trimming, look exactly like the runtime's own crash banner and the wrapper would wrongly announce the run as aborted.

src/BloomTests/web/ServeFileAndShutDownTests.cs
```diff
@@ -0,0 +1,215 @@
...
+                    PendingResponseWrites.WaitUntilAllHaveFinished(TimeSpan.FromSeconds(10)),
+                    Is.True,
+                    "the abandoned send should have been accounted for, not left counted forever"
+                );
+            }
+            finally
+            {
+                client.Dispose();
+            }
+        }
```
JohnThomson, line 180: [Devin] **Investigate**: New real-HTTP tests take the fixed-port lock but run alongside no explicit ordering guarantee for PendingResponseWrites state

`ServeFileAndShutDownTests` deliberately leaves one response half-sent (it asks for a large file and never reads the body) so that shutdown has something in flight to trip over. `PendingResponseWritesTests` starts by insisting nothing is in flight. Devin's concern is that these two fixtures talk to the same process-wide counter with no ordering guarantee between them, so in principle the stalled send could still be outstanding when the other fixture starts and make its setup fail.

build/agent-dotnet.ps1
JohnThomson: [Claude Opus 5 (1M context) from John Thomson's machine during preflight] Fixed in `11cfe91242`.

Both `Select-String` calls now pass `-CaseSensitive`, so the two wrappers apply the same rules to the same output, and the marker file says so — a pattern added later gets written for the right rules. Checked in PowerShell that a line reading `Unhandled Exception.` matches the marker without the switch and does not with it, which is what `grep -E` already did.

build/agent-dotnet.sh
JohnThomson: [Claude Opus 5 (1M context) from John Thomson's machine during preflight] Not a regression, but the comment it was based on was overstated, and that is now fixed in `11cfe91242`.

The patterns allowed indentation before this change too — they were written `^[ \t]*Unhandled exception\.`. So a test printing that text on an indented line was always going to match; trimming did not open that door, it only made the bash half behave the way the PowerShell half already did. (The reason for trimming at all is that `[ \t]` does not mean "space or tab" to `grep -E` — inside a bracket expression it is a literal backslash and letter t — so the two halves genuinely disagreed about tab-indented lines.)

What the anchor still buys is worth being precise about, so the marker file now says it: it rules out the words appearing mid-sentence, which is the common case, but not a test whose own indented output begins with them. If that ever happens the cost is a false "aborted" on a run that finished — the safe direction to be wrong in — and the verdict prints the exact line it matched on, so it takes one look to see what happened.

src/BloomTests/web/ServeFileAndShutDownTests.cs
JohnThomson: [Claude Opus 5 (1M context) from John Thomson's machine during preflight] Not an issue: neither fixture can leave a send outstanding for the other to trip over.

The stalled-send test does not end while its send is still in flight — its last assertion is that `WaitUntilAllHaveFinished` returns true, so the test only passes once that abandoned send has been accounted for. (Shutdown gives up waiting after about two seconds, the send then fails, and our own completion runs and stops counting it — that accounting is the thing the test is checking.) The other fixture's setup is the belt to that braces: it waits up to ten seconds and, if something really were still in flight, fails with a sentence saying so rather than producing a confusing failure somewhere else.

The coupling through process-wide state is real and deliberate — shutdown has to be able to ask about every send in the process, not one server's — and the note on `WaitUntilAllHaveFinished` says so.

build/agent-dotnet.ps1
```diff
@@ -32,5 +36,86 @@ Write-Host "[agent-dotnet] isolated build dir: $scratch"
...
+    if ($abortEvidence) {
+        Write-Host "[agent-dotnet] *** TEST RUN ABORTED -- NOT a pass. ***"
+        Write-Host "[agent-dotnet] Evidence: $abortEvidence"
+        Write-Host "[agent-dotnet] Any summary above counts only the tests that ran before this, so it means nothing."
+        exit 1
+    }
+    if ($status -ne 0) {
+        $detail = if ($summary) { $summary } else { 'No summary line was printed.' }
+        Write-Host "[agent-dotnet] *** TEST RUN FAILED (dotnet exited $status). *** $detail"
+        exit $status
```
JohnThomson, line 107: [Devin] **Bug**: Test wrapper can announce a failed run and still report success to whatever launched it

If `dotnet` cannot be launched at all — not on PATH in that terminal, say — then nothing ever runs to set `$LASTEXITCODE`, so the wrapper's `$status` is `$null`. It correctly prints `*** TEST RUN FAILED ***`, and then `exit $null`, which PowerShell turns into exit code **0**. So a caller that trusts the exit code — a CI step, a hook, an agent's background-task notification — is told the suite passed, on a run that never happened.

That is precisely the hazard this whole change exists to remove, so Devin finding it here is a fair hit. The bash twin does not have the hole, because `PIPESTATUS[0]` is always a number.

src/BloomExe/web/BloomServer.cs
```diff
@@ -2831,6 +2869,24 @@ var kvp in _workers.Where(kvp =>
...
+                        // A worker finishes with a request before the response body has actually gone
+                        // out; it hands the rest of the send off (see PendingResponseWrites) precisely
+                        // so that it can get back to work. So having joined all the workers we still
+                        // have to wait for those sends, because closing the listener underneath one
+                        // makes it fail -- which used to kill the process (BL-16667).
+                        if (
+                            !PendingResponseWrites.WaitUntilAllHaveFinished(
+                                TimeSpan.FromSeconds(secondsToWait)
+                            )
+                        )
```
JohnThomson, line 2889: [Devin] **Investigate**: The pending-write wait inherits an already-shortened timeout

Shutdown waits up to `secondsToWait` for response bodies that are still going out. But the loop just above halves `secondsToWait` every time a worker thread refuses to stop ("if one thing is broken, likely others are, so get less patient"). Devin's point is that the two are correlated in the wrong direction: sends are most likely to still be in flight exactly when workers are stuck, so with four stuck workers the wait shrinks to about an eighth of a second — largely defeating it in the one situation it is there for.

build/agent-dotnet.ps1
JohnThomson: [Claude Opus 5 (1M context) from John Thomson's machine during preflight] Fixed in `b8fdffce1d`.

`$status` is now normalised to a real failure code the moment it comes back empty, so every later use of it — the failure check and the `exit` — is safe, rather than patching the one place the empty value happened to leak out.

src/BloomExe/web/BloomServer.cs
JohnThomson: [Claude Opus 5 (1M context) from John Thomson's machine during preflight] Good catch — fixed in `b8fdffce1d`.

The wait now has its own fixed two-second budget instead of inheriting the decaying one. Reusing `secondsToWait` was a deliberate "if one thing is broken, get less patient", but Devin is right that the two are correlated the wrong way round here: workers refusing to stop is precisely the state in which sends are likely to still be going out, so the impatience bit hardest in the one case the wait exists for. The worst the fixed budget costs is a couple of seconds added to a shutdown that is already going badly, and Bloom's own force-exit timer is twenty seconds.

build/test-abort-markers.txt
```diff
@@ -0,0 +1,49 @@
...
+^Process terminated\.
+^Fatal error\.
+^Stack overflow\.
+Process is terminating due to StackOverflowException
+
+# VSTest noticing that the host it was driving went away.
+Test Run Aborted
+The active test run was aborted
+[Tt]esthost process exited with error
+[Tt]est host process crashed
```
JohnThomson, line 45: [Devin] **Bug**: One of the crash phrases the test wrapper watches for is written in the wrong capitalisation, so it never matches

Fallout from making the matching case-sensitive in the previous commit. The marker was written all lower case, but the test runner capitalises the phrase. This session's own reproduction of BL-16667 printed:

```
The active test run was aborted. Reason: Test host process crashed : Unhandled exception. System.ObjectDisposedException
```

so with case respected, that marker could never fire.

JohnThomson: [Claude Opus 5 (1M context) from John Thomson's machine during preflight] Fixed in `bdd087735d` — written `[Tt]est host process crashed`, matching the style of the marker beside it.

Worth noting for the record that the run was still caught: `Test Run Aborted` and `The active test run was aborted` both appear in that same output, and either of them fires. So this was a dead signal rather than a hole. Still worth fixing — a marker that cannot match is worse than no marker at all, because it looks like coverage.

JohnThomson:
[Claude Opus 5 (1M context) from John Thomson's machine during preflight] Consulted Devin on 2026-08-11, up to commit `bdd087735d`.

Three rounds, because each round's fixes were worth re-reviewing. Devin now considers every bug it raised fixed, and raised nothing new on the last pass.

It found four bugs across the three rounds, all in the new test-wrapper scripts rather than in the server fix:

- The two wrappers could reach opposite verdicts on the same run, because PowerShell ignores letter case unless told not to and `grep -E` does not — [thread](https://github.com/BloomBooks/BloomDesktop/pull/8192#discussion_r3759647498).
- The PowerShell wrapper could announce a failed run and still hand its caller exit code 0, if `dotnet` could not be launched at all — [thread](https://github.com/BloomBooks/BloomDesktop/pull/8192#discussion_r3759717137).
- One of the crash phrases was written in the wrong capitalisation, so after the case fix it could never match — [thread](https://github.com/BloomBooks/BloomDesktop/pull/8192#discussion_r3759772586).
- Shutdown's wait for responses still going out was reusing a timeout that gets halved for every worker thread that will not stop — so it shrank to a fraction of a second in exactly the situation it exists for — [thread](https://github.com/BloomBooks/BloomDesktop/pull/8192#discussion_r3759717300).

Two Investigate flags were assessed as not needing a change; both have the reasoning written on their threads rather than being dismissed here — [anchors and indentation](https://github.com/BloomBooks/BloomDesktop/pull/8192#discussion_r3759647612) and [test fixtures sharing process-wide state](https://github.com/BloomBooks/BloomDesktop/pull/8192#discussion_r3759647718). Thirteen informational items were not mirrored.

CI (`pr-automation`) is green, and no other review bot has posted on this PR.

build/agent-dotnet.sh
JohnThomson: [Claude Opus 5 (1M context) from John Thomson's machine during preflight] Developer decision: leave as is.

John reviewed this in the preflight report and confirmed the trade-off stands — the phrases are not being tightened. To restate why, so the record is complete: the patterns always allowed indentation, so trimming opened nothing new; a false "aborted" on a run that finished is the safe direction to be wrong in; and the verdict prints the exact line it matched, so a false alarm takes one look to diagnose. Tightening each phrase would also make it likelier to miss a real crash.

JohnThomson:
[Claude Opus 5 (1M context) from John Thomson's machine during preflight] Consulted Devin again on 2026-08-11, up to commit `fd62e496da`.

Fourth round, after the commit that stops `NoErrorForEmptyAudio` failing on unlucky process ids. **Clean** — no new bugs and no new flags. Devin now marks every bug it raised across the four rounds as resolved, and also marks the shutdown-timeout flag resolved. Sixteen informational items, not mirrored.

Every finding it raised has a thread on this PR with a documented outcome, and all of them are resolved.

andrew-polk (changes requested):
> @andrew-polk reviewed 12 files and all commit messages, made 1 comment, and resolved 6 discussions.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8192)* status: all files reviewed, 1 unresolved discussion (waiting on JohnThomson).

___
*[`src/BloomExe/web/PendingResponseWrites.cs` line 35 at r4](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8192#-OzmmiNs5jyAHJ41U5OC:-OzmmiNs5jyAHJ41U5OD:b-3q1r4) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/fd62e496da84038b864dd65c464094cd394dbec4/src/BloomExe/web/PendingResponseWrites.cs#L35)):*
> ```Smalltalk
>     /// flight instead of closing the listener and hoping.
>     /// </summary>
>     internal static class PendingResponseWrites
> ```

This feels like a really heavy solution to something we don't think is even a problem in production.
Have you considered other options? Even bandaids? 
Just one idea I haven't researched:
catch all exceptions at the place it says we currently only catch win32Exception; if we are in unit tests, handle, otherwise rethrow.

JohnThomson:
[Claude Opus 5 (1M context) from John Thomson's machine] Alternative implementation of this same fix, for comparison: #8194. It is ~20 lines of production change instead of ~300, and is faster in a throughput test — but a measurement of clients that stop reading says this PR's approach is the safer one. The numbers are in #8194's description. Exactly one of the two should be merged.

andrew-polk:
Superceded by #8196

PR #8193, opened 2026-08-11
Fix BL-16682 Save page before invoking Bloom AI Image Tools (approach A: minimal)

andrew-polk:
Fixes BL-16682: "Edit with AI…" opened the AI Image Tools with an empty "Image to Edit" slot when the user had just added the image.

Ref: https://issues.bloomlibrary.org/youtrack/issue/BL-16682

## ⚖️ One of two alternatives — please compare

This is **approach A: the minimal fix.** The alternative is **#8191**, which fixes the same bug by moving the AI editor overlay into the top window. Pick one; the other gets closed.

|  | This PR (A) | #8191 (B) |
| --- | --- | --- |
| Where the overlay lives | page iframe (unchanged) | top window |
| `EditingModel` | **+41 lines** — new `RunAfterNextPageLoad` hook | untouched |
| Post-save hook used | the new hook | `SaveThen`'s existing `doAfterSaveToDisk` |
| Save after a commit | deferred until the user closes the overlay | immediate |
| Repeat commit in one session | still broken | fixed |
| Files | 1 launcher module | 3 modules, split by frame |
| Live-verified | **yes** | not yet |
| Reviewed | **not yet** (no Devin, no preflight) | yes |

## Why the bug happened

Everything Bloom tells the AI editor about the book — the whole-book image list, and on commit each slot's current src — is read from the **saved** book DOM. But an image the user just added lives only in the live page, because `changeImage`/`changeImageByElement` deliberately don't save (BL-16330). So the clicked image matched nothing in the list and no edit target was sent.

Saving first also closes two latent hazards: a current-page commit result would describe an image the live page no longer showed, and `DeleteSupersededAiImageFiles`, which judges orphanhood from the saved DOM, could delete a file the live page still used.

## How this approach fixes it

The menu command posts `aiImageEditor/saveThenLaunch`; C# saves the page with `EditingModel.SaveThen`, and then calls the launcher back **in the reloaded page frame**.

Bloom's save strips the live page and therefore always ends by re-navigating to it, which tears the page frame down — and `SaveThen`'s own `doAfterSaveToDisk` runs *before* that navigation, so it is too early. Hence the one new piece: `EditingModel.RunAfterNextPageLoad(action)`, run and cleared in `HandlePageDomLoadedEvent` beside the existing "Update Book" per-page chain, and dropped when the Edit tab is left.

When Bloom isn't in a state to save, the queued launch is deliberately left queued: those states are heading for a page load anyway, and opening the overlay right then would let the in-flight save's reload kill the code that operates it.

## Verification

- Front-end: full `vitest run --no-file-parallelism` — 685 passed, 5 skipped, 62 files. Three new launcher tests.
- C#: `--filter "FullyQualifiedName~Editing|FullyQualifiedName~AiImageEditor"` — 71 passed.
- `pnpm typecheck`, eslint, and the isolated production bundle build all clean.
- **Confirmed working in a running Bloom by the developer.**

⚠️ This branch has **not** been through preflight — no Devin review, and the full C# suite hasn't been run against it. If it's the one we keep, it needs that first.

[Devin review](https://devinreview.com/BloomBooks/BloomDesktop/pull/8193)

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8193)

### DIFF
--- a/.github/skills/bloom-automation/ai-image-editor-driving.md
+++ b/.github/skills/bloom-automation/ai-image-editor-driving.md
@@ -61,6 +61,13 @@ Right-click the canvas element → **"Edit with AI…"**. Loading the iframe dir
 `bloom-exe-ai-editor-open.uitest.ts` smoke test does) does **not** register
 `aiEditorLauncher.ts`'s postMessage handler, so the commit + current-page save path wouldn't run.
 
+**The overlay does not appear on the same tick as the click.** The menu command posts
+`aiImageEditor/saveThenLaunch`, which makes C# save the current page — reloading the `page`
+frame — and only then call back into the reloaded page to open the overlay (BL-16682; the saved
+book DOM has to be current or the editor opens with an empty "Image to Edit" slot). So **wait for
+the overlay**, e.g. `await page.waitForSelector("#ai-editor-overlay iframe")`, and don't re-use
+any content-frame handle taken before the click — that frame is gone.
+
 **Gotcha:** a Comical `<canvas class="comical-generated">` overlay sits on top of the image and
 intercepts element-targeted clicks (Playwright reports `<canvas …> intercepts pointer events`).
 Drive with **raw mouse coordinates** from the image's bounding box instead:
--- a/src/BloomBrowserUI/bookEdit/editablePage.ts
+++ b/src/BloomBrowserUI/bookEdit/editablePage.ts
@@ -104,6 +104,7 @@ export interface IPageFrameExports {
         imageInfo: Omit<IImageInfo, "imageId">,
     ): void;
     removeImageId(imageId: string): void;
+    openAiImageEditor(target: IAiImageEditorTarget): void;
 }
 
 // This exports the functions that should be accessible from other IFrames or from C#.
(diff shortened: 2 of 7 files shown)


andrew-polk:
[Claude Opus 5 (1M context) from Andrew Polk's machine] Closing: Andrew has chosen the top-window implementation in #8191.

For anyone reading this later, the two were deliberate alternatives to the same fix, and the comparison shifted while they were open. This PR's selling point was that it was much the smaller change. #8191's was that it left `EditingModel` untouched by putting the AI editor overlay in the top window and using `SaveThen`'s existing `doAfterSaveToDisk` — but that turned out to be unsafe, because `EditingView.StartNavigationToEditPage` reloads the whole workspace root once `MemoryUtils.SystemIsShortOfMemory()` (Bloom's own private bytes past ~2GB, i.e. a long editing session on a big book). In that state the page saved correctly and the editor never appeared, silently. #8191 therefore adopted the same `RunAfterNextPageLoad` hook this PR introduced, so the two no longer differ on `EditingModel` at all.

What decided it was the rest of #8191: the overlay in the top window (matching the image-gallery and copyright/license commands), which removes the deferred-save workaround and fixes repeat commits in one session, plus the split of the launcher along the frame boundary.

The branch `BL-16682` is left in place rather than deleted, in case any of it is wanted.

PR #8194, opened 2026-08-12
Alternative to #8192: fix BL-16667 with a synchronous send

JohnThomson:
**An alternative to #8192, not an addition.** Same bug, same three guards, ~20 lines of production change instead of ~300. Opened so the two can be compared side by side; exactly one of them should be merged.

**Read the drawback first** — measurement below says #8192 is probably still the right one.

## The bug (identical in both)

`RequestInfo` sent response bodies with `Response.Close(buffer, willBlock: false)`. That returns as soon as the write is *issued* and lets the framework finish the send on a callback whose only handler is `catch (Win32Exception)`. Closing the `HttpListener` under such a send — which is what quitting Bloom does — made it fail with an `ObjectDisposedException` that handler does not catch, on a thread none of ours covers. It reached the runtime and killed the process.

In production this only happens at application exit: `BloomServer` is an application-level singleton, and `ProjectContext.Dispose` deliberately does not dispose it, so it survives collection switches.

## How this branch differs from #8192

| | #8192 | this branch |
|---|---|---|
| Approach | issue the same async write, but supply the completion ourselves so shutdown can wait for sends in flight | `willBlock: true` — the send finishes before the worker is free, so joining the workers is enough |
| Production lines changed | ~306 | ~20 |
| New types | `PendingResponseWrites` | none |

## Throughput is not the deciding factor

300 files of 60 KB over loopback, three rounds each:

| | #8192 (async) | this branch (blocking) |
|---|---|---|
| all 300 at once | 5962 / 6798 / 5690 ms | **4815 / 4585 / 4530 ms** |
| one at a time | 29978 / 30974 / 29664 ms | 29660 / 28453 / 29472 ms |

So the "makes this thread available sooner" comment this replaces was a reasonable guess that does not survive measurement.

## What *is* wrong with this branch

A client that stops reading holds its worker thread until the write drains, and enough of them starve the server. Measured with raw sockets that never read a byte, receive buffer squeezed to 512 bytes, 20 of them against 16 workers — then one ordinary small request:

| | ordinary request |
|---|---|
| **this branch** | **timed out after 15 s** (at 60 KB, 400 KB and 1.9 MB) |
| **#8192** | served in **94 / 106 / 115 ms** |

Bloom's existing worker top-up does not rescue it: a thread blocked in a socket write never calls `ReportThreadBlocking`, so `_countBlockedThreads` stays low and `EnsureAWorkerCanStillTakeWork` never fires.

Bloom's `>2MB` path already blocks this way today, so the exposure is not new — but this would extend it from large media to every asset Bloom serves.

## Shared with #8192

Independently justified, and not dependent on which send is used: the `Debug.Fail` that no longer fires during shutdown or under unit tests; the guard on `QueueRequest` (an I/O completion callback); the removed `Debug.Assert` in `CloseListener`; `ReplyWithStreamContent` now sending exactly `ContentLength64` bytes; the test-run abort reporting; and the `NoErrorForEmptyAudio` flake fix.

[Devin review](https://devinreview.com/BloomBooks/BloomDesktop/pull/8194)

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8194)

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -76,6 +76,12 @@ under `output/agent/<key>/` so your build/test never touches the locked shared o
 means you do **not** need to stop the developer's Bloom to build or run unit tests, and
 multiple terminals can build/test at once. See `Directory.Build.props` for how it works.
 
+- For `test`, the wrapper **judges the run and prints a verdict as its last line**, so
+  `[agent-dotnet] test run completed. Passed! ...` is the only thing you need to read (and it
+  survives `| tail`). Do not judge a run by the `Passed!`/`Failed!` summary above it: a run whose
+  test host is killed part way through still prints a passing summary of however many tests got to
+  run. The wrapper catches that and says `*** TEST RUN ABORTED ***`, and exits non-zero, as it
+  does for ordinary failures. The text it looks for lives in `build/test-abort-markers.txt`.
 - This wrapper is for **building and running tests only**. To *run* Bloom, still use
   `./go.sh` (see "Running Bloom" below) — the wrapper builds no `Bloom.exe` apphost.
   (`BloomPdfMaker.exe` is the one apphost it does build, because Bloom's PDF code shells
--- a/PAPERCUTS.md
+++ b/PAPERCUTS.md
@@ -166,22 +166,6 @@ House rules:
   own suite.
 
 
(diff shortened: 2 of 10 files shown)


andrew-polk:
Superceded by #8196

PR #8196, opened 2026-08-12
Alternative to #8192/#8194: fix BL-16667 by retiring test servers

JohnThomson:
**Third approach to BL-16667, alongside #8192 and #8194 — and the only one that leaves production's runtime behaviour completely alone.** Exactly one of the three should be merged.

## The bug

A response body can still be going out after the worker thread that produced it has finished: `RequestInfo` hands the tail of the send to the framework, which finishes it on a callback whose only handler is `catch (Win32Exception)`. Closing the `HttpListener` underneath such a send makes it fail with an `ObjectDisposedException` that handler does not catch, on a thread none of ours covers. It reaches the runtime and kills the process — a run stopping part way through while still printing a cheerful summary of however many tests passed before it died. Roughly one run in three.

## Where it actually hurts

Production disposes the server **only when Bloom quits**: it's an application-level singleton, and `ProjectContext.Dispose` deliberately leaves it alone, so it survives collection switches. The tests dispose **55 per run** — measured by counting binds in `EnsureListening`.

So this PR changes *when the tests close a listener*, not how anything is sent.

## What it does

- **`BloomServer.PreDispose()`** — everything `Dispose` does except closing the listener: stop accepting requests, stop and join the threads (they're *foreground* threads, so releasing them is what lets a process exit), free the image cache. `Dispose` is now `PreDispose` + `CloseListener`, same order, same effects — **Bloom behaves exactly as before**. `IsDisposed` is deliberately not set by `PreDispose`, so the real `Dispose` can still run later.
- **`RetiredTestServers.Retire(server)`** replaces `Dispose()` in the 11 fixtures that make a server listen. A retired server is PreDisposed at once; its listener is closed later, after several more have been retired behind it. By then whatever it was serving finished long ago.

## Why a queue, not "hold them all to the end"

Because a listener holds its port until closed, and `EnsureListening` tries only **20** ports before calling `ProgramExit.Exit()`. Measured: successive un-closed listeners take 8089, 8092, 8095, … so 55 would run the suite out of ports.

The limit that actually binds, though, is not this run but the others: ports are machine-wide, and we habitually run suites in several worktrees at once. With the cap at **3**, a run holds **four** ports — the three waiting to be closed, plus the one it is using — so four concurrent runs still fit in the twenty candidates. Ports are reused as listeners close, so the footprint does not creep.

(The cap was 5 when this PR was opened, which meant six ports per run and only three concurrent runs before the fourth died. Devin caught that; `dda169daf1` brought it down.) `RetiredTestServersTests.TheRealCapLeavesRoomForSeveralConcurrentTestRuns` now holds us to the multi-run arithmetic rather than the single-process version, because the failure it prevents is not a failing test — it is `ProgramExit.Exit()` with nothing in the output to say it was about ports.

## Also included

`build/agent-dotnet.{sh,ps1}` now judge a `dotnet test` run and print a verdict as their **last** line, exiting non-zero when a run was aborted instead of letting its partial `Passed!` summary stand. Markers live in one shared `build/test-abort-markers.txt` so the two scripts can't drift. `build/Bloom.proj` fails if the NUnit results file — which the logger writes only on completion — is missing. This is how we'll find out whether CI still hits the problem.

## Honest about what this is

A **mitigation, not a proof**. It doesn't make closing a listener safe; it waits until there's nothing to be unsafe about. If an aborted run appears again the reporting above will say so plainly, and the thing to look for is a test that leaves a response half-read.

Full C# suite run **six times end to end: all six completed, 0 failed** (3066 passed, 13 skipped, 3079 total).

## Compared with the other two

| | production runtime change | fixes prod crash-at-exit | starvation risk | tests exercise shipped path |
|---|---|---|---|---|
| #8192 own the completion | ~306 lines | yes | no | yes |
| #8194 blocking send | ~20 lines | yes | **yes** | yes |
| **this PR** | **none** (pure extraction) | no | no | yes |

[Devin review](https://devinreview.com/BloomBooks/BloomDesktop/pull/8196)

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8196)

### DIFF
--- a/build/agent-dotnet.sh
+++ b/build/agent-dotnet.sh
@@ -17,6 +17,9 @@
 #   the one global -p: this script appends, and apphost policy lives in that same props
 #   file because it has to vary per project).
 #
+#   For `test` it does one more thing: it judges the run itself and prints a verdict as
+#   its LAST line. See "Judging a test run" below.
+#
 # Usage (exactly like dotnet, just build/test through this script):
 #   build/agent-dotnet.sh test src/BloomTests/BloomTests.csproj --filter "FullyQualifiedName~UrlPathStringTests"
 #   build/agent-dotnet.sh build src/BloomExe/BloomExe.csproj
@@ -48,5 +51,98 @@ echo "[agent-dotnet] isolated build dir: $SCRATCH" >&2
 # be a global here too, but it has to vary per project — WebView2PdfMaker's apphost is
 # the BloomPdfMaker.exe the PDF tests shell out to — so it now lives in
 # Directory.Build.props, which a global property would have overridden.
-BLOOM_AGENT_BUILD_DIR="$SCRATCH" exec dotnet "$@" \
-    -p:OutDir="$SCRATCH/bin/"
+
+# For everything except `test`, dotnet's own exit code is the whole story, so get out of
+# the way entirely and let it have this process.
+if [ "${1:-}" != "test" ]; then
+    BLOOM_AGENT_BUILD_DIR="$SCRATCH" exec dotnet "$@" \
+        -p:OutDir="$SCRATCH/bin/"
+fi
+
+# ---------------------------------------------------------------------------
+# Judging a test run
+#
+# A `dotnet test` run can stop early — Bloom's own code taking the test host down with
+# it is the case that prompted this (BL-16667) — and when it does, VSTest still prints a
+# summary of the tests that got as far as running. Everything that ran had passed, so
+# that summary reads "Passed!". Anything that judges the run by its last lines, which is
+# what a human skimming and an agent reading `| tail` both do, calls that green.
+#
+# There is a second way to be told a lie here, recorded in PAPERCUTS.md: a run whose own
+# summary said `Failed! - Failed: 11` still arrived at the caller as exit code 0 (a pipe
+# swallows the status, and callers do pipe this).
+#
+# So: capture the output, judge it three ways (crash markers, dotnet's exit code, the
+# summary line), and print a one-line verdict LAST so it survives `| tail`. Nothing is
+# hidden — the output still streams to the terminal as it happens.
+# ---------------------------------------------------------------------------
+
+MARKERS="$SCRIPT_DIR/test-abort-markers.txt"
+if [ ! -f "$MARKERS" ]; then
+    echo "[agent-dotnet] cannot find $MARKERS, so a truncated test run could not be detected" >&2
+    exit 1
+fi
+
+LOG="$(mktemp -t agent-dotnet-test-XXXXXX)"
+TRIMMED="$LOG.trimmed"
+trap 'rm -f "$LOG" "$TRIMMED"' EXIT
+
+# stderr is merged in because that is where the runtime prints the crash banners we are
+# looking for. PIPESTATUS[0] rather than $? because $? here would be tee's.
+set +e
+BLOOM_AGENT_BUILD_DIR="$SCRATCH" dotnet "$@" -p:OutDir="$SCRATCH/bin/" 2>&1 | tee "$LOG"
+STATUS=${PIPESTATUS[0]}
+set -e
+
+# Match against a copy with the indentation taken off the front of every line. That is what lets
+# the markers file anchor its patterns with a plain ^ and mean the same thing here as it does to
+# Select-String in the PowerShell twin -- writing "start of line, then optional whitespace" is the
+# one thing the two regex engines cannot be made to agree on (see the note in the markers file).
+sed 's/^[[:space:]]*//' "$LOG" >"$TRIMMED"
+
+ABORT_EVIDENCE=""
+while IFS= read -r pattern || [ -n "$pattern" ]; do
+    # The repo is text=auto and Windows checkouts are autocrlf, so the markers file may well
+    # arrive with CRLF line endings. A trailing carriage return left on the end of a pattern
+    # would silently stop it ever matching, which is the one failure this file must not have.
+    pattern="${pattern%$'\r'}"
+    case "$pattern" in '' | '#'*) continue ;; esac
+    match="$(grep -m1 -E "$pattern" "$TRIMMED" || true)"
+    if [ -n "$match" ]; then
+        ABORT_EVIDENCE="$match"
+        break
+    fi
+done <"$MARKERS"
+
+SUMMARY="$(grep -E "^(Passed|Failed)!" "$TRIMMED" | tail -1 || true)"
+
+echo ""
+if [ -n "$ABORT_EVIDENCE" ]; then
+    echo "[agent-dotnet] *** TEST RUN ABORTED -- NOT a pass. ***"
+    echo "[agent-dotnet] Evidence: $ABORT_EVIDENCE"
+    echo "[agent-dotnet] Any summary above counts only the tests that ran before this, so it means nothing."
+    exit 1
+fi
+if [ "$STATUS" -ne 0 ]; then
+    echo "[agent-dotnet] *** TEST RUN FAILED (dotnet exited $STATUS). *** ${SUMMARY:-No summary line was printed.}"
+    exit "$STATUS"
+fi
+if [ -z "$SUMMARY" ]; then
+    # A discovery-only run has no tests to summarize, so its silence is not evidence of anything.
+    for arg in "$@"; do
+        if [ "$arg" = "--list-tests" ] || [ "$arg" = "-t" ]; then
+            echo "[agent-dotnet] listed tests without running them."
+            exit 0
+        fi
+    done
+    echo "[agent-dotnet] *** TEST RUN INCOMPLETE: it exited 0 but never printed a summary, so we do not know what ran. ***"
+    exit 1
+fi
+case "$SUMMARY" in
+    *Failed!*)
+        # dotnet said 0 but its own summary says otherwise; believe the summary.
+        echo "[agent-dotnet] *** TEST RUN FAILED. *** $SUMMARY"
+        exit 1
+        ;;
+esac
+echo "[agent-dotnet] test run completed. $SUMMARY"
--- a/build/agent-dotnet.sh
+++ b/build/agent-dotnet.sh
@@ -114,7 +114,7 @@ while IFS= read -r pattern || [ -n "$pattern" ]; do
     fi
 done <"$MARKERS"
 
-SUMMARY="$(grep -E "^(Passed|Failed)!" "$TRIMMED" | tail -1 || true)"
+SUMMARY="$(grep -E "^(Passed|Failed|Skipped)!" "$TRIMMED" | tail -1 || true)"
 
 echo ""
 if [ -n "$ABORT_EVIDENCE" ]; then
@@ -143,6 +143,13 @@ case "$SUMMARY" in
         # dotnet said 0 but its own summary says otherwise; believe the summary.
         echo "[agent-dotnet] *** TEST RUN FAILED. *** $SUMMARY"
         exit 1
+        ;;
+    Skipped!*)
+        # A healthy run in which every matched test was skipped -- e.g. the RAB test without
+        # BLOOM_RUN_RAB_MANUAL_TESTS. Not a failure, but say so plainly: "completed" on its own
+        # would read as "the tests passed", and nothing ran.
+        echo "[agent-dotnet] test run completed, but every test was skipped. $SUMMARY"
+        exit 0
         ;;
 esac
 echo "[agent-dotnet] test run completed. $SUMMARY"
--- a/src/BloomTests/RetiredTestServers.cs
+++ b/src/BloomTests/RetiredTestServers.cs
@@ -48,12 +48,21 @@ namespace BloomTests
         internal const int kPortsEnsureListeningTries = 20;
 
         /// <summary>
-        /// How many retired servers keep their listener open at once. Every one of these holds a
-        /// port, so this has to stay well under the number of ports EnsureListening will try;
-        /// bigger also means longer between a server's last request and its listener closing,
-        /// which is the safety this class is buying.
+        /// How many retired servers keep their listener open at once. Bigger means longer between
+        /// a server's last request and its listener closing, which is the safety this class buys.
+        ///
+        /// What limits it is not this run but the others: each held listener is a port, so a run
+        /// occupies this many plus the one it is actually using, out of the
+        /// <see cref="kPortsEnsureListeningTries"/> EnsureListening will try — machine-wide. We
+        /// habitually run suites in several worktrees at once, and the failure when a run cannot
+        /// find a port is not a failing test but ProgramExit.Exit, with nothing to say it was
+        /// about ports. At 3 a run holds 4, so four concurrent runs still fit; at 5 they would
+        /// not. TheRealCapLeavesRoomForSeveralConcurrentTestRuns holds us to that.
+        ///
+        /// The delay this buys is generous even so: retirements happen at fixture boundaries,
+        /// seconds apart, against a window of danger measured in milliseconds.
         /// </summary>
-        internal const int kMaxAwaitingClose = 5;
+        internal const int kMaxAwaitingClose = 3;
 
         private static readonly RetiredServerQueue _theQueue = new RetiredServerQueue(
             kMaxAwaitingClose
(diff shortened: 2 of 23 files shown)


src/BloomTests/web/RetiredTestServersTests.cs
```diff
@@ -0,0 +1,90 @@
...
+                    RetiredTestServers.CountAwaitingClose,
+                    Is.LessThanOrEqualTo(RetiredTestServers.kMaxAwaitingClose),
+                    $"after retiring {i + 1} servers, more than the cap were still holding a port; "
+                        + "a run retires about 55, so this would exhaust EnsureListening's ports"
+                );
+            }
+        }
+
+        [Test]
+        public void Retire_ThenCloseAllNow_LeavesNothingHoldingAPort()
```
JohnThomson: [Devin] **Bug**: New safety-net tests close test web servers immediately after use, re-creating the crash the change exists to avoid

The tests meant to guard this scheme were the one place in the suite reliably doing the dangerous thing.

`RetiredTestServersTests` called `RetiredTestServers.CloseAllNow()` in the middle of the run. That drains the **shared** queue, so it force-closed listeners that other fixtures had retired moments earlier — and its own servers had just served a real response, because `EnsureListening` issues one through `VerifyWeAreNowListening`. Closing a listener microseconds after it answered a request is precisely the timing BL-16667 is about. NUnit runs a fixture's tests alphabetically, so this landed immediately after the test that retires five listening servers.

src/BloomTests/RetiredTestServers.cs
```diff
@@ -0,0 +1,137 @@
...
+    /// would collide with the other test runs this team habitually has going in parallel
+    /// worktrees. Keeping a handful in the queue costs a handful of ports and still puts whole
+    /// fixtures between a server's last request and its listener closing.
+    ///
+    /// This is a mitigation, not a proof. It does not make closing a listener safe; it waits until
+    /// there is nothing to be unsafe about. If an aborted run ever shows up again -- and
+    /// build/agent-dotnet.* now says so loudly rather than printing a passing summary -- the thing
+    /// to look for is a test that leaves a response half-read.
+    /// </summary>
+    public static class RetiredTestServers
```
JohnThomson, line 45: [Devin] **Investigate**: One fixture still leaks a listening server, which permanently consumes one of the twenty ports

`FileIOApiTests` creates its `BloomServer` in `[OneTimeSetUp]` and never disposes or retires it, and its tests go through `ApiTest`, which makes the server listen. So that listener holds one of `EnsureListening`'s twenty ports for the rest of the run.

Pre-existing and harmless until now, but it eats into the port budget this change introduces, and it contradicts the rule this PR adds to `AGENTS.md` — that a fixture which calls `EnsureListening` retires its server rather than disposing it.

src/BloomTests/web/RetiredTestServersTests.cs
JohnThomson: [Claude Opus 5 (1M context) from John Thomson's machine during preflight] Good catch, and an uncomfortable one — fixed in `3625d77d66`.

The bookkeeping now lives in an instance, `RetiredServerQueue`, with `RetiredTestServers` as a thin static over one of them. The tests drive their own queue, so nothing they do can reach the one the run is using, and they retire servers that never listened — no listener to close, nothing to interrupt. The queue counts servers rather than ports, so its behaviour is identical either way.

The one thing that arrangement can no longer show directly is that the *real* cap is affordable, so that is now its own assertion: the cap must stay well under `EnsureListening`'s port budget. Those two constants live in different files and nothing else was keeping them in relation.

src/BloomTests/RetiredTestServers.cs
JohnThomson: [Claude Opus 5 (1M context) from John Thomson's machine during preflight] Confirmed and fixed in `3625d77d66`.

`FileIOApiTests` now retires its server in a `[OneTimeTearDown]`, like every other fixture that listens. You are right that it was pre-existing and harmless until this change made the port budget matter — and it would have been an odd thing to leave standing right after adding a rule to `AGENTS.md` saying fixtures that call `EnsureListening` retire their server.

build/agent-dotnet.sh
```diff
@@ -48,5 +51,105 @@ echo "[agent-dotnet] isolated build dir: $SCRATCH" >&2
...
+    pattern="${pattern%$'\r'}"
+    case "$pattern" in '' | '#'*) continue ;; esac
+    match="$(grep -m1 -E "$pattern" "$TRIMMED" || true)"
+    if [ -n "$match" ]; then
+        ABORT_EVIDENCE="$match"
+        break
+    fi
+done <"$MARKERS"
+
+SUMMARY="$(grep -E "^(Passed|Failed|Skipped)!" "$TRIMMED" | tail -1 || true)"
```
JohnThomson, line 117: [Devin] **Bug**: A test run in which every test is skipped is reported as an incomplete failure

The wrapper only recognised a summary line beginning `Passed!` or `Failed!`. VSTest also prints `Skipped!` when every test it matched was skipped, so such a run looked like a run that printed no summary at all, and the wrapper announced `*** TEST RUN INCOMPLETE ***` and exited 1 — a false red on a healthy run.

Reachable through a workflow `AGENTS.md` documents: the Reading App Builder test calls `Assert.Ignore` unless `BLOOM_RUN_RAB_MANUAL_TESTS` is set, so the filtered command `AGENTS.md` gives for it skips its only test.

JohnThomson: [Claude Opus 5 (1M context) from John Thomson's machine during preflight] Fixed in `145cda2d9e`. Reproduced it first, with exactly the command `AGENTS.md` gives:

```
Skipped! - Failed: 0, Passed: 0, Skipped: 1, Total: 1
[agent-dotnet] *** TEST RUN INCOMPLETE: it exited 0 but never printed a summary ***
```

Both wrappers now recognise `Skipped!`, and say so in as many words — "completed, but every test was skipped" — rather than a bare "completed", which would read as "the tests passed" when nothing ran. Verified in bash and PowerShell that the all-skipped run now exits 0 with that wording and an ordinary run is unaffected.

Fair catch in a change whose whole point is that a test run should report itself honestly.

src/BloomTests/RetiredTestServers.cs
```diff
@@ -0,0 +1,171 @@
...
+    {
+        /// <summary>Matches kNumberOfPortsToTry in BloomServer.EnsureListening.</summary>
+        internal const int kPortsEnsureListeningTries = 20;
+
+        /// <summary>
```
JohnThomson, line 54: [Devin] **Investigate**: Retired listeners raise the per-run port footprint from ~1 to ~6, and the failure mode is a hard process exit

`EnsureListening` scans only 20 candidate ports and, if all fail, calls `ProgramExit.Exit()`. With this change a run keeps up to 5 retired listeners plus the live server — about 6 held ports instead of 1. Safe for one run, but this PR itself notes the team habitually runs suites in parallel worktrees: three concurrent runs would hold ~18 of the 20, and the fourth wouldn't fail a test, it would take the whole run down with nothing pointing at ports. The guard test only checked the single-process relation, so nothing covered the multi-process case.

JohnThomson: [Claude Opus 5 (1M context) from John Thomson's machine during preflight] Good catch, and the right frame — it is about our habits, not this run. Fixed in `dda169daf1`.

The cap is now 3, so a run holds four ports and four concurrent runs still fit inside `EnsureListening`'s twenty candidates. The delay that buys is unchanged in kind and still generous: retirements happen at fixture boundaries, seconds apart, against a window of danger measured in milliseconds.

The guard test now checks the relation that actually matters — cap plus one, times four concurrent runs, must fit in the candidates — rather than the single-process version, which a cap of 5 would have sailed through. You were right that nothing covered the multi-process case; that was the real gap here, more than the number itself.

JohnThomson:
[Claude Opus 5 (1M context) from John Thomson's machine during preflight] Consulted Devin on 2026-08-12, up to commit `dda169daf1`.

Five rounds, because each round's fixes were worth re-reviewing. The last is clean: Devin now considers every finding it raised fixed, and raised nothing new.

It found three bugs and two flags, all in the test-side code and the wrapper scripts rather than in the production change:

- The tests written to guard this scheme were themselves the one place in the suite reliably re-creating the race — they called `CloseAllNow()` mid-run, draining the shared queue and closing listeners that had just served a response — [thread](https://github.com/BloomBooks/BloomDesktop/pull/8196#discussion_r3770429887).
- A run in which every test is skipped was reported as `*** TEST RUN INCOMPLETE ***` and exited non-zero, because the wrapper only recognised `Passed!`/`Failed!` and not `Skipped!`. Reachable through the exact Reading App Builder command `AGENTS.md` documents; reproduced before fixing — [thread](https://github.com/BloomBooks/BloomDesktop/pull/8196#discussion_r3770546043).
- Retiring listeners raised the per-run port footprint from about one to about six, so three concurrent worktree runs would have held eighteen of `EnsureListening`'s twenty candidates and the fourth would have died via `ProgramExit.Exit` with nothing naming ports. The cap came down from 5 to 3, and the guard test now checks the multi-run arithmetic rather than the single-process one — [thread](https://github.com/BloomBooks/BloomDesktop/pull/8196#discussion_r3770996411).
- `FileIOApiTests` never released its server at all, holding a port for the whole run — pre-existing, but it ate into the new budget and contradicted the rule this PR adds to `AGENTS.md` — [thread](https://github.com/BloomBooks/BloomDesktop/pull/8196#discussion_r3770430045).

Seventeen informational items were not mirrored. The in-session local review ran too and independently found the `FileIOApiTests` leak, plus one latent seam — `PreDispose` bypasses the protected virtual `Dispose(bool)`, so a future subclass would be half-invoked — which is now documented in the code rather than changed, since nothing derives from `BloomServer` today.

CI (`pr-automation`) is green, and no other review bot has posted on this PR.

src/BloomTests/RetiredTestServers.cs
```diff
@@ -0,0 +1,185 @@
...
+    /// fixtures between a server's last request and its listener closing.
+    ///
+    /// This is a mitigation, not a proof. It does not make closing a listener safe; it waits until
+    /// there is nothing to be unsafe about. If an aborted run ever shows up again -- and
+    /// build/agent-dotnet.* now says so loudly rather than printing a passing summary -- the thing
+    /// to look for is a test that leaves a response half-read.
+    /// </summary>
+    public static class RetiredTestServers
+    {
+        /// <summary>
```
JohnThomson, line 47: [Devin] **Investigate**: The port budget arithmetic is asserted against a copied constant, not the real one

`kPortsEnsureListeningTries = 20` in the test helper was a hand-copied duplicate of `kNumberOfPortsToTry` in `BloomServer.EnsureListening`. So if someone lowered the real budget, `TheRealCapLeavesRoomForSeveralConcurrentTestRuns` would keep passing while the relation it asserts no longer held — and the failure it guards against is `ProgramExit.Exit()`, which says nothing about ports on its way out.

src/BloomTests/RetiredTestServers.cs
```diff
@@ -0,0 +1,185 @@
...
+    /// the defence would be pointless until the day parallelism is actually switched on, and on
+    /// that day this note is what says where to look.
+    /// </remarks>
+    internal sealed class RetiredServerQueue
+    {
+        private readonly int _maxAwaitingClose;
+        private readonly object _lock = new object();
+        private readonly Queue<BloomServer> _awaitingClose = new Queue<BloomServer>();
+
+        internal RetiredServerQueue(int maxAwaitingClose)
```
JohnThomson, line 115: [Devin] **Investigate**: Retirement scheme depends on fixtures never overlapping, because ServerIsListening is process-global

`ServerIsListening` is static, and both `PreDispose` and the deferred `Dispose` set it false. So retiring a server fires two global writes: one for the server being retired, and one for the older server it evicts — the second triggered by an unrelated fixture's teardown.

Harmless today, because the suite runs strictly sequentially. But it is a new coupling: with assembly-level parallelism, an eviction in one fixture would clear the global "server is listening" flag out from under another fixture's live server.

src/BloomTests/RetiredTestServers.cs
JohnThomson: [Claude Opus 5 (1M context) from John Thomson's machine during preflight] Fair hit, and on a guard you asked for two rounds ago — fixed in `0040f15fbc`.

`kNumberOfPortsToTry` is now a class-level `internal const` instead of a local, and the test references it directly rather than a copy. That is the entire production change: a constant moved up a scope and made visible to the test assembly, nothing behavioural. A guard that cannot fail is worse than no guard, because it looks like cover.

src/BloomTests/RetiredTestServers.cs
JohnThomson: [Claude Opus 5 (1M context) from John Thomson's machine during preflight] Recorded rather than defended against, in `0040f15fbc`.

You are right that it is a new coupling and right that it is harmless today — `BloomTests.runsettings` sets no parallelism, so no other server is live when an eviction fires. Guarding it now would mean inventing a mechanism for a configuration nobody runs; the note on `RetiredServerQueue` instead says plainly what the scheme assumes and what would break if assembly-level parallelism were switched on, which is what someone will need on the day they try it.

The related observation about `_theOneInstance` never being cleared is pre-existing and left alone deliberately: this branch is meant to change when tests close listeners and nothing else, and we are stabilising a release.

JohnThomson:
[Claude Opus 5 (1M context) from John Thomson's machine during preflight] Consulted Devin again on 2026-08-13, up to commit `0040f15fbc`.

Two more rounds after master was merged in. No unresolved bugs. Two flags on the merged branch, both now addressed:

- The port-budget assertion compared the cap against a **hand-copied** 20 rather than `BloomServer`'s own `kNumberOfPortsToTry`, so lowering the real budget would have left the guard passing while the relation it asserts no longer held. `kNumberOfPortsToTry` is now a class-level `internal const` and the test references it — [thread](https://github.com/BloomBooks/BloomDesktop/pull/8196#discussion_r3777142347).
- Retiring a server evicts an older one, and both set the **static** `ServerIsListening` to false, so an eviction fires from an unrelated fixture's teardown. Harmless while fixtures run one at a time, which they do; recorded on the queue rather than defended against, since the defence would be inventing a mechanism for a configuration nobody runs — [thread](https://github.com/BloomBooks/BloomDesktop/pull/8196#discussion_r3777142492).

The second is the one flag Devin still raises on the latest pass. That is expected: an Investigate flag has no resolved state, and the fix for this one was a note, which is what its own text suggested. Its outcome is documented on the thread.

Across the whole preflight: five bugs and four flags, every one fixed or documented, every thread resolved. Twenty informational items not mirrored. CI green; full suite green at `0040f15fbc` (3075 passed, 13 skipped).

andrew-polk (approved):
> @andrew-polk reviewed 23 files and all commit messages, made 1 comment, and resolved 6 discussions.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8196)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on JohnThomson).

___
*[`src/BloomTests/RetiredTestServers.cs` line 107 at r4](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8196#-Ozw4EvvAVLYQnL4BNYJ:-Ozw4EvvAVLYQnL4BNYK:b-hxgpqp) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/0040f15fbcb40dd75b9c2be56d0612951bf411ce/src/BloomTests/RetiredTestServers.cs#L107)):*
> ```Smalltalk
>     /// fixture's live server. Devin raised it; recorded here rather than defended against, because
>     /// the defence would be pointless until the day parallelism is actually switched on, and on
>     /// that day this note is what says where to look.
> ```

I'm not excited about moving us further in a direction that parallelism is harder as I was hoping at some point soon to task an agent with finding out where/how we can enable parallelism. But we need a fix today...

andrew-polk (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8196)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on JohnThomson).

PR #8182, opened 2026-08-10
Fix gear icon displaced in Picture on Right page with ebook theme (BL-14901)

andrew-polk:
Fixes the 6.0 regression where the format cog (gear icon) on a "Picture on Right" page was displaced away from the book margin when the book uses an ebook page theme.

The two ebook appearance themes (Zero Margin Ebook, Rounded Border Ebook) set `--formatButton-pageNumber-dodge: 20px` so the edit-mode cog dodges the page number, which those themes normally place at the bottom left. But on landscape `pictureOnRight` pages the same themes move the page number to the right side, so the dodge just pushed the cog 20px away from the margin for nothing. Each theme now zeroes the dodge using the same `pictureOnRight` selectors it already uses to move the page number.

The fix is deliberately in the theme files rather than the shared `editMode.less` dodge rule: the EFL migration stylesheets also set a dodge but keep their page number bottom-left on `pictureOnRight` pages, so a blanket rule would have regressed BL-13206 for those books.

Verified live in Bloom (Device16x9Landscape book, Image on Right page + Basic Text & Image control page, both ebook themes): the cog's computed `margin-left` is now 3px (was 23px) on the pictureOnRight page, and remains 23px (dodge intact) on the control page.

Ref: https://issues.bloomlibrary.org/youtrack/issue/BL-14901

🤖 Generated with [Claude Code](https://claude.com/claude-code)

[Devin review](https://devinreview.com/BloomBooks/BloomDesktop/pull/8182)

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8182)

PR #8198, opened 2026-08-12
Gear icon dodge follows the page number's real position (BL-14901)

andrew-polk:
Follow-up to #8182 (the landscape Picture-on-Right fix): the tester found the gear icon (format cog) still displaced away from the margin on portrait ebook pages, and the same pointless dodge occurred whenever the Book Settings "Page Numbers" control moved the number away from the bottom left.

Two commits:

1. **Zero Margin portrait fix.** That theme only places its page number at the bottom left on landscape pages; on portrait pages the number is centered (the device-size default), yet its 20px dodge applied everywhere. The dodge is now scoped to landscape. Also gave the default theme's dodge a unit (`0` → `0px`): the unitless zero made the cog's `margin-left: calc(3px + var(...))` invalid, so in every default-theme book the cog sat at 0 instead of the intended 3px (harmless but wrong since ~6.0).

2. **The dodge now honors the Page Numbers setting.** `editMode.less` folds everything that determines where the number actually sits into one effective dodge, reading the same margin-override variables the appearance system writes for that setting (via container style queries): Center or Right ⇒ nothing to dodge, whatever the theme wanted; Left ⇒ the theme's full dodge even on pages whose theme moved the number elsewhere (e.g. pictureOnRight); Automatic ⇒ the theme's per-page decision, as before. Both ebook themes publish the new `--formatButton-pageNumber-dodge-when-forced-left`; the EFL migration books' baked-in stylesheets fall back to their ordinary dodge.

Verified live in Bloom with a 12-case matrix (both ebook themes × both orientations × all four Page Numbers settings, measuring the cog's computed margin on a Picture-on-Right page and a text-bottom control page): every case the tester reported now puts the cog at the margin, and every case where the number genuinely sits bottom-left still dodges, including landscape regression checks.

Ref: https://issues.bloomlibrary.org/youtrack/issue/BL-14901

[Devin review](https://devinreview.com/BloomBooks/BloomDesktop/pull/8198)

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8198)

### DIFF
--- a/src/BloomBrowserUI/bookEdit/css/editMode.less
+++ b/src/BloomBrowserUI/bookEdit/css/editMode.less
@@ -923,17 +923,23 @@ div.textWholePage ul {
 // Now sometimes, the format button lands under the page number.
 // Changing the z-index did not help. So we allow themes to "dodge" it.
 // Kind of a hack, I know. See BL-13206.
+// The dodge only makes sense when the page number actually sits at the bottom left,
+// which depends on two things (BL-14901):
+//  - the theme's automatic placement, expressed as --formatButton-pageNumber-dodge on
+//    the page (themes set it to 0 on pages where they move the number elsewhere,
+//    e.g. pictureOnRight);
+//  - the user's Book Settings > Page Numbers choice, which beats the theme. It reaches
+//    CSS as the pageNumber margin overrides (see AppearanceSettings.SetPageNumberProperties):
+//    Center sets both overrides to 50%; Right sets the left override to deliberately-invalid;
+//    Left sets the right override to deliberately-invalid; Automatic/Hidden set neither.
+// The rules below fold all of that into --formatButton-pageNumber-dodge-effective,
+// which is what the margin actually uses.
 @container style(--pageNumber-show-multiplicand:1) {
-    .split-pane-component.position-right
-        .split-pane-component-inner
-        > .bloom-translationGroup
-        > #formatButton,
-    .split-pane-component.position-top
-        .split-pane-component-inner
-        > .bloom-translationGroup
-        > #formatButton {
-        // don't need this dodge if the page number can't be in this text area
-        --formatButton-pageNumber-dodge: 0px;
+    #formatButton {
+        // Default: the theme's automatic-placement decision.
+        --formatButton-pageNumber-dodge-effective: var(
+            --formatButton-pageNumber-dodge
+        );
     }
     .split-pane-component.position-left
         > .split-pane-component-inner
@@ -949,7 +955,44 @@ div.textWholePage ul {
         > .bloom-translationGroup
         > #formatButton {
         // this "3px" is there from long ago, nothing to do with this nudging
-        margin-left: calc(3px + var(--formatButton-pageNumber-dodge));
+        margin-left: calc(3px + var(--formatButton-pageNumber-dodge-effective));
+    }
+}
+// The user put the page number at the center or the right: nothing to dodge,
+// whatever the theme wanted.
+@container style(--pageNumber-left-margin-override: 50%) or
+    style(--pageNumber-left-margin-override: deliberately-invalid) {
+    #formatButton {
+        --formatButton-pageNumber-dodge-effective: 0px;
+    }
+}
+// The user forced the page number to the left: dodge, even on pages where the theme's
+// automatic placement would have moved the number away (e.g. pictureOnRight). Themes
+// that waive the dodge per-page publish the distance to use in this situation as
+// --formatButton-pageNumber-dodge-when-forced-left; stylesheets that never waive it
+// per-page (e.g. the EFL migration books' customBookStyles.css) don't define it, so
+// fall back to their ordinary dodge value.
+@container style(--pageNumber-right-margin-override: deliberately-invalid) {
+    #formatButton {
+        --formatButton-pageNumber-dodge-effective: var(
+            --formatButton-pageNumber-dodge-when-forced-left,
+            var(--formatButton-pageNumber-dodge)
+        );
+    }
+}
+// These selectors are more specific than the plain #formatButton rules above, so this
+// wins regardless of the page-number position: a text pane in the right/top half of the
+// page can never sit over a bottom-left page number, so it never needs the dodge.
+@container style(--pageNumber-show-multiplicand:1) {
+    .split-pane-component.position-right
+        .split-pane-component-inner
+        > .bloom-translationGroup
+        > #formatButton,
+    .split-pane-component.position-top
+        .split-pane-component-inner
+        > .bloom-translationGroup
+        > #formatButton {
+        --formatButton-pageNumber-dodge-effective: 0px;
     }
 }
 
--- a/src/BloomBrowserUI/bookEdit/css/editMode.less
+++ b/src/BloomBrowserUI/bookEdit/css/editMode.less
@@ -936,9 +936,13 @@ div.textWholePage ul {
 // which is what the margin actually uses.
 @container style(--pageNumber-show-multiplicand:1) {
     #formatButton {
-        // Default: the theme's automatic-placement decision.
+        // Default: the theme's automatic-placement decision. The 0px fallback matters:
+        // legacy-theme books get no appearance-theme-default.css block (see
+        // AppearanceSettings.ToCss), so nothing defines the variable there, and an
+        // undefined var() would invalidate the margin calc() below entirely.
         --formatButton-pageNumber-dodge-effective: var(
-            --formatButton-pageNumber-dodge
+            --formatButton-pageNumber-dodge,
+            0px
         );
     }
     .split-pane-component.position-left
@@ -976,22 +980,21 @@ div.textWholePage ul {
     #formatButton {
         --formatButton-pageNumber-dodge-effective: var(
             --formatButton-pageNumber-dodge-when-forced-left,
-            var(--formatButton-pageNumber-dodge)
+            var(--formatButton-pageNumber-dodge, 0px)
         );
     }
 }
-// These selectors are more specific than the plain #formatButton rules above, so this
-// wins regardless of the page-number position: a text pane in the right/top half of the
-// page can never sit over a bottom-left page number, so it never needs the dodge.
+// A text box with any position-right or position-top split among its ancestors sits
+// somewhere in the right or top portion of the page, so it can never be over a
+// bottom-left page number and never needs the dodge — whatever the rules above decided.
+// Deliberately no child combinators: however deep the origami nesting, one top/right
+// ancestor is enough. (The margin rule above does use child combinators, and there they
+// matter: they anchor the dodge to a pane's own immediate text box and keep
+// canvas-element text boxes out of it.) These selectors still out-rank the plain
+// #formatButton rules above.
 @container style(--pageNumber-show-multiplicand:1) {
-    .split-pane-component.position-right
-        .split-pane-component-inner
-        > .bloom-translationGroup
-        > #formatButton,
-    .split-pane-component.position-top
-        .split-pane-component-inner
-        > .bloom-translationGroup
-        > #formatButton {
+    .split-pane-component.position-right #formatButton,
+    .split-pane-component.position-top #formatButton {
         --formatButton-pageNumber-dodge-effective: 0px;
     }
 }
--- a/src/content/appearanceThemes/appearance-theme-default.css
+++ b/src/content/appearanceThemes/appearance-theme-default.css
@@ -125,8 +125,9 @@
     --marginBox-border-width: medium; /* must not be 0px/0mm/0in to have a border. 'medium' is a keyword. */
 
     /* In some themes, at edit-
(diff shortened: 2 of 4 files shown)


src/BloomBrowserUI/bookEdit/css/editMode.less
```diff
@@ -949,7 +955,44 @@ div.textWholePage ul {
...
+// per-page (e.g. the EFL migration books' customBookStyles.css) don't define it, so
+// fall back to their ordinary dodge value.
+@container style(--pageNumber-right-margin-override: deliberately-invalid) {
+    #formatButton {
+        --formatButton-pageNumber-dodge-effective: var(
+            --formatButton-pageNumber-dodge-when-forced-left,
+            var(--formatButton-pageNumber-dodge)
+        );
+    }
+}
```
andrew-polk, line 986: [Devin] **Investigate**: Books whose theme never needed a dodge still get none when the user forces the number left

The forced-left branch falls back to `var(--formatButton-pageNumber-dodge)`, which for the default theme is now `0px`. So in a plain default-theme book where the user sets Page Numbers = Left, the number moves to the bottom left but the cog does not dodge at all. In practice the default theme puts the number 15px above the page bottom, i.e. below the marginBox (page-margin 12mm), so an actual overlap is unlikely — but it is worth confirming for page sizes with small bottom margins (Device/Ebook sizes under the *default* theme, where `--pageNumber-extra-height` is 8mm and the number sits inside the marginBox area). If an overlap does occur there, the default theme would need its own `--formatButton-pageNumber-dodge-when-forced-left`.

andrew-polk: [Claude Fable 5 during preflight] Not an issue — the default theme keeps the page number outside the text area in every case, including forced-Left, so no dodge is needed there.

The dodge exists because the two ebook themes zero `--pageNumber-extra-height` and place the number *on* the page content. The default theme does the opposite: for Device/Ebook page sizes it reserves `--pageNumber-extra-height: 8mm` (appearance-theme-default.css), which shrinks the marginBox so the number sits *below* the box — and forcing Page Numbers = Left moves the number horizontally without changing that reservation. The cog lives at the bottom-left *inside* the marginBox, so they can't overlap. For print sizes the 12mm bottom margin does the same job. So "no dodge in themes that never declared one" is the intended behavior, not a gap; a theme that ever places the number inside the text area is exactly a theme that must declare a dodge.

andrew-polk:
[Claude Fable 5 during preflight] Consulted Devin on 2026-08-12 up to commit aa7c5bcebfdff505f9d5686dc7312288c47a9eae. Review: 0 bugs, 1 Investigate flag, 3 Informational observations. The Investigate flag (default-theme books get no dodge when Page Numbers is forced Left) was posted as an inline thread, assessed as not-an-issue (the default theme reserves space so its page number sits below the text area in every case), and resolved: https://github.com/BloomBooks/BloomDesktop/pull/8198#discussion_r3770463613. The informational notes (cascade verification, the 0px unit fix affecting all default-theme books by restoring the intended 3px, and the forced-left variable being published on portrait pages) all describe intended behavior. CI green at time of writing.

JohnThomson (changes requested):
> @JohnThomson reviewed 4 files and all commit messages, made 1 comment, and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8198)* status: all files reviewed, 1 unresolved discussion (waiting on andrew-polk).

___
*[`src/BloomBrowserUI/bookEdit/css/editMode.less` line 996 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8198#-OzwaZcw9pTnJyWEYuDa:-OzwaZcx8qowbBjcZnj2:b-2krlja) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/aa7c5bcebfdff505f9d5686dc7312288c47a9eae/src/BloomBrowserUI/bookEdit/css/editMode.less#L996)):*
> ```Less
>         > #formatButton {
>         --formatButton-pageNumber-dodge-effective: 0px;
>     }
> ```

I think these rules would be better without the wedges. In more complex origami, a TG that is in the bottom left of it's immediate parent split doesn't need to dodge if any ancestor split is top or right; but the wedges will prevent this rule from fixing it.

andrew-polk (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8198)* status: 3 of 4 files reviewed, 1 unresolved discussion (waiting on JohnThomson).

___
*[`src/BloomBrowserUI/bookEdit/css/editMode.less` line 996 at r1](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8198#-OzwaZcw9pTnJyWEYuDa:-P-084CVB_G4E-lwk7_R:b7ybaqz) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/aa7c5bcebfdff505f9d5686dc7312288c47a9eae/src/BloomBrowserUI/bookEdit/css/editMode.less#L996)):*

I think these rules would be better without the wedges. In more complex origami, a TG that is in the bottom left of it's immediate parent split doesn't need to dodge if any ancestor split is top or right; but the wedges will prevent this rule from fixing it.
</blockquote></details>

[Claude Fable 5] Done in 5f91be937a — the no-dodge rule is now just `.split-pane-component.position-right #formatButton, .split-pane-component.position-top #formatButton`, no child combinators.

Worth being clear that this isn't actually fixing a behavior bug, just making the rule's intent legible. The old selector's first hop was already a descendant combinator (`.position-right .split-pane-component-inner`), so a text group at the bottom/left of its immediate split already had its dodge suppressed whenever *any* ancestor split was top or right — the trailing `inner > TG > #formatButton` part is satisfied by construction for any button the dodge margin reaches at all. I verified with DOM tests over nested origami (TG in left-of-right, bottom-of-right, left-of-top, and a three-level bottom-of-left-of-top): every ancestor-top/right case was suppressed both before and after the change, and genuine bottom-left panes still dodge. So the wedge removal changes no outcomes; it just makes "one top/right ancestor is enough" something the selector says instead of something you have to prove.

The child combinators in the margin-applying rule above stay, though — there they're load-bearing: they anchor the dodge to a pane's own immediate text box and keep canvas-element text boxes out of the dodge entirely (an all-descendant version would sweep overlay text boxes into the 20px nudge). The code comment now explains this distinction.

src/BloomBrowserUI/bookEdit/css/editMode.less
```diff
@@ -923,17 +923,27 @@ div.textWholePage ul {
...
+    #formatButton {
+        // Default: the theme's automatic-placement decision. The 0px fallback matters:
+        // legacy-theme books get no appearance-theme-default.css block (see
+        // AppearanceSettings.ToCss), so nothing defines the variable there, and an
+        // undefined var() would invalidate the margin calc() below entirely.
+        --formatButton-pageNumber-dodge-effective: var(
+            --formatButton-pageNumber-dodge,
+            0px
+        );
     }
```
andrew-polk, line 947: [Devin] **Investigate**: Effective-dodge chain still yields margin-left:0 when no stylesheet defines the base dodge variable (e.g. legacy theme)

`--formatButton-pageNumber-dodge-effective: var(--formatButton-pageNumber-dodge)` has no fallback. For books whose appearance.css does not include the default theme's variable block — `AppearanceSettings.ToCss` skips `appearance-theme-default.css` entirely when the theme name starts with "legacy", yet still emits `--pageNumber-show-multiplicand: 1` from the book settings — the `var()` is guaranteed-invalid, so the custom property is invalid at computed-value time and `margin-left: calc(3px + var(--formatButton-pageNumber-dodge-effective))` drops to 0 instead of the intended 3px. This is the same class of defect the PR fixes for the default theme via `0px`; adding a fallback (`var(--formatButton-pageNumber-dodge, 0px)`, and likewise in the forced-left chain) would make it robust for any stylesheet that omits the variable.

andrew-polk: [Claude Fable 5 during preflight] Fixed in `1b6d1d34d4`: both var() chains now carry a 0px fallback, so the margin survives any stylesheet that omits the variable. Verified the premise first: AppearanceSettings.ToCss really does skip the default-theme block for legacy themes, and appearance-theme-legacy-5-6.css never defines the variable. Note this was pre-existing rather than a regression — the old `calc(3px + var(--formatButton-pageNumber-dodge))` failed identically for legacy books — but it's the same defect class this PR fixes for the default theme, so it belongs here. Resolving.

src/BloomBrowserUI/bookEdit/css/editMode.less
```diff
@@ -949,7 +959,43 @@ div.textWholePage ul {
...
+// Deliberately no child combinators: however deep the origami nesting, one top/right
+// ancestor is enough. (The margin rule above does use child combinators, and there they
+// matter: they anchor the dodge to a pane's own immediate text box and keep
+// canvas-element text boxes out of it.) These selectors still out-rank the plain
+// #formatButton rules above.
+@container style(--pageNumber-show-multiplicand:1) {
+    .split-pane-component.position-right #formatButton,
+    .split-pane-component.position-top #formatButton {
+        --formatButton-pageNumber-dodge-effective: 0px;
     }
```
andrew-polk, line 999: [Devin] **Investigate**: Dodge cancellation now uses descendant combinators, which changes behavior for nested origami splits

The old rule cancelled the dodge only for a format button that was the immediate text box of a `position-right`/`position-top` pane (`.split-pane-component.position-right .split-pane-component-inner > .bloom-translationGroup > #formatButton`). The new rule matches any `#formatButton` anywhere under such a pane. Combined with the margin rule (which still requires the immediate pane to be `position-left`/`position-bottom`/`marginBox`), a real behavior change exists for nested layouts: a text box in a `position-bottom` sub-pane of an outer `position-top` pane, or in a `position-left` sub-pane of an outer `position-right` pane, now loses its dodge. In lopsided splits (e.g. a 90/10 horizontal split, or a 1/99 vertical split) that text box can still sit near the bottom-left of the page, where the number lives, so the cog could overlap the number again. Worth a quick sanity check on an extreme origami layout; in typical 50/50 splits the box is far from the bottom-left corner so it doesn't matter.

andrew-polk: [Claude Fable 5 during preflight] Not a behavior change — the premise misreads the old selector. Its first combinator was already a *descendant* combinator (`.split-pane-component.position-right .split-pane-component-inner`, note the space), so it never required the button to be the immediate text box of the right/top pane: any text box whose own inner sat anywhere under a right/top component was already cancelled. The cases named here (a bottom sub-pane inside a top pane, a left sub-pane inside a right pane) were verified with DOM tests against both selector versions — including a three-level nesting — and every one was cancelled identically before and after this commit. The trailing `inner > TG > #formatButton` the old rule carried is satisfied by construction for any button the dodge margin reaches, which is why removing it changes no outcomes.

The lopsided-split observation (a 90/10 split can put a "top-ancestor" text box visually near the bottom-left where the number lives) is true but pre-existing and unchanged by this PR: the dodge rules have keyed off position classes, not geometry, since BL-13206, and split percentages aren't expressed in classes for CSS to see. If that corner case ever bites a real book, it needs a different mechanism, not these selectors. Resolving.

andrew-polk:
[Claude Fable 5 during preflight] Consulted Devin on 2026-08-13 up to commit 1b6d1d34d4c574a382e92de5d31c32786e971c8f (two re-reviews this cycle, one per push). Outcomes: 0 bugs each. The legacy-theme margin flag was fixed in 1b6d1d34d4 (0px fallbacks in both var() chains) — thread https://github.com/BloomBooks/BloomDesktop/pull/8198#discussion_r3785687958. The descendant-combinator flag was assessed as not-a-behavior-change (the old selector's first hop was already a descendant combinator; DOM tests show identical outcomes before/after, and the lopsided-split geometry caveat is pre-existing, inherent to class-based rules) — thread https://github.com/BloomBooks/BloomDesktop/pull/8198#discussion_r3785787026. The forced-Left flag is a repeat of the already-resolved https://github.com/BloomBooks/BloomDesktop/pull/8198#discussion_r3770463613. Informational observations (6) all describe intended or verified behavior. Local suites at this HEAD: vitest 524 passed / 5 skipped; C# not re-run for this delta (edit-mode CSS only, unreachable by BloomTests; green at aa7c5bce earlier in this PR).

andrew-polk (commented):
> @andrew-polk reviewed 1 file and all commit messages, and resolved 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8198)* status: all files reviewed, 1 unresolved discussion (waiting on JohnThomson).

JohnThomson (approved):
> @JohnThomson reviewed 1 file and all commit messages, and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8198)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on andrew-polk).

PR #8199, opened 2026-08-12
Honor no-checkout flag (BL-16691)

andrew-polk:
Honor a new `AllowCheckouts` setting in the `.bloomCollection` file, as preparation for Cloud syncing. There is deliberately no UI for it: a collection administrator pauses checkouts by hand-editing the XML, and in a Team Collection that file syncs to the whole team.

When `AllowCheckouts` is `False`, the "Check out book" button is disabled and grayed out, and an info NoteBox below it says "The administrator of this collection has paused checkouts."

### Changes

- **`CollectionSettings`** — read `AllowCheckouts` (defaulting to `true`, so existing collections are unaffected) and write it back out. The write is load-bearing: `Save()` rebuilds the whole file, so a setting we didn't round-trip would be silently erased — quietly un-pausing checkouts for the team — the next time anything saved collection settings.
- **`TeamCollectionApi`** — report `checkoutsArePaused` in the book status JSON, and refuse in `HandleAttemptLockOfCurrentBook`. The endpoint guard is there because the panel's status can be stale (the setting arrives whenever collection files sync), and because the registration-dialog path posts the checkout separately from the button.
- **`StatusPanelCommon`** — add a `belowButton` slot, and pay off the YAGNI note on `disabledBackground`: the check-out button is the primary color, not the warning color, so a disabled one no longer renders in the check-in button's orange.
- **LESS** — dim the button's `` when disabled; MUI dims the label but not our icon, which otherwise left a bright icon beside grayed-out text.

### Testing

Added `CollectionSettingsTests` covering the default when the element is absent, reading `True`/`False`, and surviving a save-and-reload round trip. I verified the round-trip test genuinely fails if the `Save()` line is removed.

Manually verified in a real Team Collection: the disabled button and the note appear as intended.

### Note for whoever writes the admin instructions

In a Team Collection, hand-editing the **local** `.bloomCollection` while Bloom is closed does **not** work — at startup `SyncLocalAndRepoCollectionFiles` overwrites the local copy from the repo ("whatever wins in the repo wins everywhere"). The setting has to be edited either while Bloom is running (the idle watcher pushes it up to the repo) or directly in the repo's `Other Collection Files.zip`. This is pre-existing behavior for all collection-level settings, not something this PR introduces.

Ref: https://issues.bloomlibrary.org/youtrack/issue/BL-16691

[Devin review](https://devinreview.com/BloomBooks/BloomDesktop/pull/8199)

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8199)

### DIFF
--- a/src/BloomBrowserUI/teamCollection/TeamCollectionBookStatusPanel.tsx
+++ b/src/BloomBrowserUI/teamCollection/TeamCollectionBookStatusPanel.tsx
@@ -25,6 +25,7 @@ import { IBookTeamCollectionStatus } from "./teamCollectionApi";
 import { ForceUnlockDialog } from "./ForceUnlockDialog";
 import { kBloomRed } from "../utils/colorUtils";
 import { showRegistrationDialog } from "../react_components/registration/registrationDialog";
+import { NoteBox } from "../react_components/boxes";
 
 interface CheckInProgressEvent extends IBloomWebSocketEvent {
     fraction: number;
@@ -412,6 +413,10 @@ export const TeamCollectionBookStatusPanel: React.FunctionComponent<
                                     setBusy(false);
                                 },
                                 () => {
+                                    // Likewise on failure: when the server refuses because an
+                                    // administrator paused checkouts, it pushes a status refresh,
+                                    // so the panel will redraw and explain itself. We just need to
+                                    // stop looking busy, and not report an error. See BL-16691.
                                     setBusy(false);
                                 },
                             );
@@ -421,6 +426,26 @@ export const TeamCollectionBookStatusPanel: React.FunctionComponent<
                                     if (hasValidEmail)
                                         post(
                                             "teamCollection/attemptLockOfCurrentBook",
+                                            undefined,
+                                            // This needs a failure handler even though it has
+                                            // nothing to do. Without one, post() rethrows, and a
+                                            // refused checkout surfaces as Bloom's "something went
+                                            // wrong" problem report -- blaming Bloom when in fact
+                                            // an administrator paused checkouts. The server pushes
+                                            // a status refresh when it refuses, so the panel
+                                            // explains itself. See BL-16691.
+                                            //
+                                            // Note this swallows *every* failure here, not just a
+                                            // paused checkout, so a genuine problem (shared folder
+                                            // unreachable, say) now shows nothing on this path.
+                                            // Considered distinguishing the paused case by status
+                                            // code and reporting anything else; decided against it
+                                            // because the ordinary button path above has always had
+                                            // a do-nothing failure handler too, so this matches it
+                                            // rather than introducing a new kind of silence, and
+                                            // both paths refresh status so the panel ends up
+                                            // showing the real state either way.
+                                            () => {},
                                         );
                                 },
                             });
@@ -451,7 +476,20 @@ export const TeamCollectionBookStatusPanel: React.FunctionComponent<
                             "checkout-button",
                             "/bloom/teamCollection/Check Out.svg",
                             checkoutHandler,
+                            props.checkoutsArePaused,
                         )}
+                        belowButton={
+                            props.checkoutsArePaused ? (
+                                // Deliberately not localized yet: this is an
+                                // unreleased admin-only setting with no UI, so
+                                // per AGENTS.md we ship the English and add an
+                                // XLF entry once the feature stabilizes.
+                                <NoteBox>
+                                    The administrator of this collection has
+                                    paused checkouts.
+                                </NoteBox>
+                            ) : undefined
+                        }
                         menu={menu}
                     />
                 );
--- a/src/BloomBrowserUI/teamCollection/statusPanelCommon.tsx
+++ b/src/BloomBrowserUI/teamCollection/statusPanelCommon.tsx
@@ -26,6 +26,7 @@ export interface IStatusPanelProps {
     useWarningColorForButton?: boolean;
     children?: JSX.Element;
     menu?: JSX.Element; // when book is checked out, About my Avatar... and Forget Changes and Check in Book
+    belowButton?: JSX.Element; // spans the full panel width underneath the button row
     className?: string; // additional class for  root div; enables emotion CSS.
 }
 
@@ -43,9 +44,12 @@ export const StatusPanelCommon: React.FunctionComponent<IStatusPanelProps> = (
                     main: buttonColor,
                 },
                 action: {
-                    // Yagni: currently the only time we disable a button is when using the warning color
-                    // If we ever disable the other version, we probably want primary.dark here.
-                    disabledBackground: outerTheme.palette.warning.dark,
+                    // Keep the disabled background in the same family as the button's own color.
+                    // The check-in button uses the warning color; the check-out button uses primary,
+                    // and gets disabled when an administrator has paused checkouts (BL-16691).
+                    disabledBackground: props.useWarningColorForButton
+                        ? outerTheme.palette.warning.dark
+                        : outerTheme.palette.primary.dark,
                 },
             },
         }),
@@ -94,6 +98,7 @@ export const StatusPanelCommon: React.FunctionComponent<IStatusPanelProps> = (
                     </ThemeProvider>
                 </StyledEngineProvider>
             </div>
+            {props.belowButton}
         </div>
     );
 };
--- a/src/BloomExe/Collection/CollectionSettings.cs
+++ b/src/BloomExe/Collection/CollectionSettings.cs
@@ -194,6 +194,7 @@ namespace Bloom.Collection
             XMatterPackName = kDefaultXmatterName;
             Language2Tag = "en";
             AllowNewBooks = true;
+            AllowCheckouts = true;
             CollectionName = "dummy collection";
             AudioRecordingMode = TalkingBookApi.AudioRecordingMode.Sentence;
             AudioRecordingTrimEndMilliseconds = kDefaultAudioRecordingTrimEndMilliseconds;
@@ -394,6 +395,16 @@ namespace Bloom.Collection
             xml.Add(new XElement("Province", Province));
             xml.Add(new XElement("District", District));
             xml.Add(new XElement("AllowNewBooks", AllowNewBooks.ToString()));
+            // Unlike the settings around it, this one is written only when it has been set false
+            // to prevent checkouts.
+            // Save() rebuilds the whole file, so writing it unconditionally would add a line to
+            // every .bloomCollection in existence the first time its settings were saved -- and in
+            // a Team Collection that edit gets pushed to the shared folder, so the churn would
+            // spread to the whole team for a setting almost nobody uses. Omitting it is safe
+            // because a missing element already reads back as true (see Load), so a paused
+            // collection still round-trips correctly. See BL-16691.
+            if (!AllowCheckouts)
+                xml.Add(new XElement("AllowCheckouts", AllowCheckouts.ToString()));
             xml.Add(new XElement("AudioRecordingMode", AudioRecordingMode.ToString()));
             xml.Add(
                 new XElement("AudioRecordingTrimEndMilliseconds", AudioRecordingTrimEndMilliseconds)
@@ -644,6 +655,15 @@ namespace Bloom.Collection
                 Province = ReadString(xml, "Province", "");
                 District = ReadString(xml, "District", "");
                 AllowNewBooks = ReadBoolean(xml, "AllowNewBooks", true);
+                // A missing element reads back as true, which is what we want: checkouts are
+                // allowed unless someone deliberately turns them off. Note that a *malformed*
+                // value does not fall back to that default -- ReadBoolean returns whatever
+                // bool.TryParse produced, which is false -- so "no" or a typo pauses checkouts for
+                // the whole team rather than being ignored. We considered special-casing this
+                // setting to be strict and decided against it: every other boolean in this file
+                // behaves the same way, and making this the one exception would be more surprising
+                // than the behavior itself. See BL-16691.
+                AllowCheckouts = ReadBoolean(xml, "AllowCheckouts", true);
                 IsSourceCollection = ReadBoolean(xml, "IsSourceCollection", false);
 
                 string audioRecordingModeStr = ReadString(xml, "AudioRecordingMode", "Unknown");
@@ -981,6 +1001,11 @@ namespace Bloom.Collection
         public int OneTimeCheckVersionNumber { get; set; }
 
         public bool AllowNewBooks { get; set; }
+
+        /// <summary>
+        /// When false, no one may check out a book in this (Team) collection. See BL-16691.
+        /// </summary>
+        public bool AllowCheckouts { get; set; }
 
         public TalkingBookApi.AudioRecordingMode AudioRecordingMode { get; set; }
 
--- a/src/BloomExe/TeamCollection/FolderTeamCollection.cs
+++ b/src/BloomExe/TeamCollection/FolderTeamCollection.cs
@@ -4,6 +4,7 @@ using System.Diagnostics;
 using System.IO;
 using System.Linq;
 using System.Text.RegularExpressions;
+using System.Xml.Linq;
 using Bloom.Collection;
 using Bloom.CollectionCreating;
 using Bloom.MiscUI;
@@ -301,6 +302,66 @@ namespace Bloom.TeamCollection
         {
             var path = GetPathToBookFileInRepo(bookName);
             return RobustZip.GetZipEntryContent(path, fileName);
+        }
+
+        /// <summary>
+        /// Read AllowCheckouts straight out of the repo's copy of the collection settings file,
+        /// which lives inside "Other Collection Files.zip". We deliberately do not go through the
+        /// local copy: that is only refreshed from the repo when the collection is opened, and the
+        /// whole point is to notice a change made while Bloom is running. See BL-16691.
+        /// </summary>
+        public override bool? GetAllowCheckoutsFromRepo()
+        {
+            try
+            {
+                var zipPath = GetRepoProjectFilesZipPath(_repoFolderPath);
+                if (!RobustFile.Exists(zipPath))
+                    return null;
+                var entryName = GetCollectionSettingsEntryName(zipPath);
+                if (entryName == null)
+                    return null;
+                var content = RobustZip.GetZipEntryContent(zipPath, entryName);
+                if (string.IsNullOrWhiteSpace(content))
+                    return null;
+                return CollectionSettings.ReadBoolean(
+                    XElement.Parse(content),
+                    "AllowCheckouts",
+                    true
+                );
+            }
+            catch (Exception e)
+            {
+                // The zip may be part-written, or Dropbox may be midway through syncing it. There
+                // is nothing to do about it now: we'll get another notification when the file
+                // settles, and failing that we read it again at startup. Leaving the setting alone
+                // is the safe outcome either way.
+                Bloom.Utils.MiscUtils.SuppressUnusedExceptionVarWarning(e);
+                return null;
+            }
+        }
+
+        /// <summary>
+        /// The name of the .bloomCollection entry within the given zip, or null if there isn't one.
+        /// Found by extension rather than by building the name, because the collection may have
+        /// been renamed locally.
+        /// </summary>
+        private static string GetCollectionSettingsEntryName(string zipPath)
+        {
+            using (var zipFile = new ICSharpCode.SharpZipLib.Zip.ZipFile(zipPath))
+            {
+                foreach (ICSharpCode.SharpZipLib.Zip.ZipEntry entry in zipFile)
+                {
+                    if (
+                        entry.IsFile
+                        && entry.Name.EndsWith(
+                            CollectionSettings.kFileExtension,
+                            StringComparison.OrdinalIgnoreCase
+                        )
+                    )
+                        return entry.Name;
+                }
+            }
+            return null;
         }
 
         /// <summary>
--- a/src/BloomExe/TeamCollection/TeamCollection.cs
+++ b/src/BloomExe/TeamCollection/TeamCollection.cs
@@ -870,6 +870,13 @@ namespace Bloom.TeamCollection
                     else
                     {
                         CopyRepoCollectionFilesFromLocal(_localCollectionFolder);
+                        // Pick up AllowCheckouts from what we just pushed. Our own repo watcher is
+                        // deliberately suppressed while we write (see CopyRepoCollectionFilesFromLocal),
+                        // so without this the machine that made the change is the one machine that
+                        // does NOT act on it. Worse, our in-memory copy would still say checkouts are
+                        // allowed, and the next Save() would write that back over the change and
+                        // un-pause the whole team. See BL-16691.
+                        UpdateAllowCheckoutsFromRepo();
                     }
                 }
             }
@@ -1345,6 +1352,49 @@ namespace Bloom.TeamCollection
                 null,
                 null
             );
+            UpdateAllowCheckoutsFromRepo();
+        }
+
+        /// <summary>
+        /// Read the AllowCheckouts setting from the repo's copy of the collection settings, which
+        /// may be newer than the copy we read when the collection was opened. Returns null if we
+        /// can't tell (disconnected, no repo copy yet, or the file is unreadable right now).
+        /// </summary>
+        public virtual bool? GetAllowCheckoutsFromRepo()
+        {
+            return null;
+        }
+
+        /// <summary>
+        /// Collection settings otherwise take effect only when Bloom next opens the collection, but
+        /// pausing checkouts must not wait for that: Bloom may stay open for days, so an
+        /// administrator who pauses checkouts would have no effect on anyone still running. Here we
+        /// pick up that one setting whenever the repo's collection files change. Deliberately only
+        /// this one -- reloading all collection settings mid-session is not safe. See BL-16691.
+        ///
+        /// This is best-effort by nature: the change has to reach this machine through the shared
+        /// folder first, and the user may be offline. The aim is that anyone with a working
+        /// connection stops being able to check out within a few minutes of the switch being
+        /// flipped, not that it is instant.
+        /// </summary>
+        internal void UpdateAllowCheckoutsFromRepo()
+        {
+            var settings = _tcManager?.Settings;
+            if (settings == null)
+                return; // no settings to update (unit tests)
+            var repoValue = GetAllowCheckoutsFromRepo();
+            if (repoValue == null || repoValue.Value == settings.AllowCheckouts)
+                return;
+            settings.AllowCheckouts = repoValue.Value;
+            Logger.WriteEvent(
+                $"TeamCollection: AllowCheckouts changed remotely to {repoValue.Value}."
+            );
+            // Tell the browser to re-read book status, so the checkout button and its explanation
+            // update without a restart. The InvokeSelectionChanged that our caller does afterwards
+            // is NOT enough: nothing on that path sends bookTeamCollectionStatus, so the panel kept
+            // showing a live checkout button until something else forced it to refetch. Tested by
+            // pausing checkouts in the shared folder with Bloom running.
+            _tcManager.SendBookStatusReload();
         }
 
         /// <summary>
--- a/src/BloomExe/TeamCollection/TeamCollectionApi.cs
+++ b/src/BloomExe/TeamCollection/TeamCollectionApi.cs
@@ -534,6 +534,7 @@ namespace Bloom.TeamCollection
                         isNewLocalBook = true,
                         checkinMessage = "",
                         isUserAdmin = _tcManager.OkToEditCollectionSettings,
+                        checkoutsArePaused = !_settings.AllowCheckouts,
                     }
                 );
             }
@@ -620,6 +621,7 @@ namespace Bloom.TeamCollection
                     isNewLocalBook,
                     checkinMessage,
                     isUserAdmin = _tcManager.OkToEditCollectionSettings,
+                    checkoutsArePaused = !_settings.AllowCheckouts,
                 }
             );
         }
@@ -685,6 +687,23 @@ namespace Bloom.TeamCollection
             if (!_tcManager.CheckConnection())
             {
                 request.Failed();
+                return;
+            }
+
+            // The UI disables the checkout button when checkouts are paused, but it decides that
+            // from a status snapshot it fetched earlier, and not every path back here re-checks it
+            // -- the post issued after the registration dialog, for one, doesn't. This is also the
+            // side that notices an administrator pausing checkouts part way through a session (see
+            // TeamCollection.UpdateAllowCheckoutsFromRepo), so it can know before the browser does.
+            // Either way, refuse here rather than rely on the browser having current status.
+            // See BL-16691.
+            if (!_settings.AllowCheckouts)
+            {
+                // Tell the browser to re-read book status, so the panel redraws itself into the
+                // paused state and explains why the checkout didn't happen. Without this the
+                // person just sees the button stop responding with no reason given.
+                _socketServer.SendEvent("bookTeamCollectionStatus", "reload");
+                request.Failed("checkouts are paused for this collection");
                 return;
             }
 
--- a/src/BloomBrowserUI/teamCollection/statusPanelCommon.less
+++ b/src/BloomBrowserUI/teamCollection/statusPanelCommon.less
@@ -34,6 +34,11 @@
                 width: @panel-button-icon-size;
                 margin-top: -2px;
             }
+            // Material-UI dims a disabled button's text but not our image, which would
+            // leave a bright icon next to a grayed-out label. See BL-16691.
+            &:disabled img {
+                opacity: 0.4;
+            }
             & > span {
                 font-size: larger;
             }
--- a/src/BloomBrowserUI/teamCollection/teamCollectionApi.tsx
+++ b/src/BloomBrowserUI/teamCollection/teamCollectionApi.tsx
@@ -25,6 +25,7 @@ export interface IBookTeamCollectionStatus {
     error: string; // This one is not current sent from the C# side.
     checkInMessage: string;
     isUserAdmin: boolean;
+    checkoutsArePaused: boolean; // an administrator has turned off checkouts for the whole collection
 }
 
 export const initialBookStatus: IBookTeamCollectionStatus = {
@@ -45,6 +46,7 @@ export const initialBookStatus: IBookTeamCollectionStatus = {
     error: "",
     checkInMessage: "",
     isUserAdmin: false,
+    checkoutsArePaused: false,
 };
 
 export function useTColBookStatus(
--- a/src/BloomExe/TeamCollection/TeamCollectionManager.cs
+++ b/src/BloomExe/TeamCollection/TeamCollectionManager.cs
@@ -35,6 +35,14 @@ namespace Bloom.TeamCollection
         /// </summary>
         void SendBookContentReload();
 
+        /// <summary>
+        /// Sends the bookTeamCollectionStatus/reload websocket event so the book status panel
+        /// re-reads status. Call this after changing something the panel displays without the
+        /// book's own checkout status having changed -- a book-status-changed event would be the
+        /// wrong signal for that. See BL-16691.
+        /// </summary>
+        void SendBookStatusReload();
+
         // ENHANCE: Add other properties and methods as needed
     }
 
@@ -563,6 +571,12 @@ namespace Bloom.TeamCollection
         public void SendBookContentReload()
         {
             _webSocketServer.SendEvent("bookContent", "reload");
+        }
+
+        /// <inheritdoc />
+        public void SendBookStatusReload()
+        {
+            _webSocketServer.SendEvent("bookTeamCollectionStatus", "reload");
         }
 
         /// <summary>
--- a/src/BloomTests/Collection/CollectionSettingsTests.cs
+++ b/src/BloomTests/Collection/CollectionSettingsTests.cs
@@ -263,6 +263,125 @@ namespace BloomTests.Collection
             Assert.That(settings.Subscription.Code, Is.EqualTo("Foobar-123456-1234"));
         }
 
+        /// <summary>
+        /// AllowCheckouts (BL-16691) has no UI; an administrator hand-edits it into the file.
+        /// Collections that predate it, and the overwhelming majority that will never have it,
+        /// must keep allowing checkouts.
+        /// </summary>
+        [Test]
+        public void AllowCheckouts_ElementMissing_DefaultsToTrue()
+        {
+            var settings = CreateSettingsFromFileContents(
+                "allowCheckoutsMissing",
+                @"<?xml version=""1.0"" encoding=""utf-8""?>
+<Collection version=""0.2"">
+	<AllowNewBooks>True</AllowNewBooks>
+</Collection>"
+            );
+            Assert.That(settings.AllowCheckouts, Is.True);
+        }
+
+        [TestCase("False", false)]
+        [TestCase("True", true)]
+        public void AllowCheckouts_ElementPresent_IsRead(string valueInFile, bool expected)
+        {
+            var settings = CreateSettingsFromFileContents(
+                "allowCheckoutsRead" + valueInFile,
+                $@"<?xml version=""1.0"" encoding=""utf-8""?>
+<Collection version=""0.2"">
+	<AllowCheckouts>{valueInFile}</AllowCheckouts>
+</Collection>"
+            );
+            Assert.That(settings.AllowCheckouts, Is.EqualTo(expected));
+        }
+
+        /// <summary>
+        /// Save() rebuilds the whole settings file from scratch, so a setting we don't write out
+        /// would be silently erased the next time anything saves collection settings. That would
+        /// quietly un-pause checkouts for the whole team, so guard against it.
+        /// </summary>
+        [Test]
+        public void AllowCheckouts_False_SurvivesSaveAndReload()
+        {
+            const string collectionName = "allowCheckoutsRoundTrip";
+            var settings = CreateSettingsFromFileContents(
+                collectionName,
+                @"<?xml version=""1.0"" encoding=""utf-8""?>
+<Collection version=""0.2"">
+	<AllowCheckouts>False</AllowCheckouts>
+</Collection>"
+            );
+            // Sanity check: we only learn anything from the round trip if it started out false.
+            Assert.That(
+                settings.AllowCheckouts,
+                Is.False,
+                "setup failed: AllowCheckouts should have been read as false"
+            );
+
+            settings.Save();
+
+            var reloaded = CreateCollectionSettings(_folder.Path, collectionName);
+            Assert.That(reloaded.AllowCheckouts, Is.False);
+        }
+
+        /// <summary>
+        /// The setting is written only when checkouts are paused, so that the overwhelming
+        /// majority of collections -- which will never use it -- don't all gain a line the first
+        /// time their settings are saved. See BL-16691.
+        /// </summary>
+        [Test]
+        public void AllowCheckouts_True_IsNotWrittenToTheFile()
+        {
+            const string collectionName = "allowCheckoutsNotWritten";
+            var settings = CreateSettingsFromFileContents(
+                collectionName,
+                @"<?xml version=""1.0"" encoding=""utf-8""?>
+<Collection version=""0.2"">
+	<AllowCheckouts>False</AllowCheckouts>
+</Collection>"
+            );
+            // Sanity check: start from the value that IS written, so we can watch it disappear.
+            Assert.That(
+                settings.AllowCheckouts,
+                Is.False,
+                "setup failed: AllowCheckouts should have been read as false"
+            );
+            settings.Save();
+            Assert.That(
+                RobustFile.ReadAllText(settings.SettingsFilePath),
+                Does.Contain("AllowCheckouts"),
+                "setup failed: a paused collection should write the element"
+            );
+
+            settings.AllowCheckouts = true;
+            settings.Save();
+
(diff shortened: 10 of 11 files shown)


DistFiles/localization/en/Bloom.xlf
```diff
@@ -5679,6 +5679,11 @@ is mostly status, which shows on the Collection Tab -->
         <note>ID: TeamCollection.Checkout</note>
         <note>On button to check out a book</note>
       </trans-unit>
+      <trans-unit id="TeamCollection.CheckoutsPaused" sil:dynamic="true" translate="no">
+        <source xml:lang="en">The administrator of this collection has paused checkouts.</source>
+        <note>ID: TeamCollection.CheckoutsPaused</note>
+        <note>Shown under the disabled "Check out book" button when the collection's AllowCheckouts setting is False</note>
+      </trans-unit>
```
andrew-polk: [Devin] **Bug**: New "checkouts paused" message is never registered for translation and triggers a missing-string warning

*(Posted by Claude Opus 5 from Andrew Polk's machine during preflight — mirrored from Devin, which does not post to GitHub itself.)*

The new `TeamCollection.CheckoutsPaused` entry was added between the `<!--` on line 5644 and the `-->` on line 5815, so it sits **inside** the big commented-out block and is not a real element in the shipped English string list.

What that means in practice: the note still displays the right English words (the component falls back to its hard-coded text), but Bloom's localization manager never sees the id. On Alpha/Developer builds that makes a "Missing l10n" warning toast appear every time the paused-checkouts note is shown.

The reason this doesn't bite the neighbouring TeamCollection strings is that every one of them is looked up with the i18n warning explicitly suppressed, whereas this one went through `NoteBox`, which does its own lookup and has no way to pass that flag.

src/BloomBrowserUI/teamCollection/statusPanelCommon.tsx
```diff
@@ -43,9 +44,12 @@ export const StatusPanelCommon: React.FunctionComponent<IStatusPanelProps> = (
...
                 action: {
-                    // Yagni: currently the only time we disable a button is when using the warning color
-                    // If we ever disable the other version, we probably want primary.dark here.
-                    disabledBackground: outerTheme.palette.warning.dark,
+                    // Keep the disabled background in the same family as the button's own color.
+                    // The check-in button uses the warning color; the check-out button uses primary,
+                    // and gets disabled when an administrator has paused checkouts (BL-16691).
+                    disabledBackground: props.useWarningColorForButton
+                        ? outerTheme.palette.warning.dark
+                        : outerTheme.palette.primary.dark,
```
andrew-polk, line 52: [Devin] **Investigate**: disabledBackground change also affects other primary-color panel buttons

*(Posted by Claude Opus 5 from Andrew Polk's machine during preflight — mirrored from Devin, which does not post to GitHub itself.)*

Replacing the hard-coded `warning.dark` with a conditional means every non-warning button in this panel would now get `primary.dark` as its disabled background, not just the check-out button. And because the blue's auto-derived `dark` shade is fairly close to the enabled blue, a disabled check-out button is distinguished mainly by its dimmed label text — so it's worth confirming it actually reads as disabled rather than just slightly darker.

DistFiles/localization/en/Bloom.xlf
andrew-polk: [Claude Opus 5 from Andrew Polk's machine during preflight] Fixed in `a6b8086545` — though not the way the suggested edit proposed, and the difference is deliberate.

Devin suggested moving the entry out of the comment block. We want it to stay in there: that block exists precisely so TeamCollection strings aren't sent to translators yet ("Don't want these actually translated until things stabilize"), and every other string on this panel lives in it.

What actually needed fixing was the code side. The reason the neighbouring strings don't produce this warning is that they're all looked up with `temporarilyDisableI18nWarning` set. This one went through `NoteBox`, which does its own lookup and has no way to pass that flag. So the string is now looked up alongside the panel's other strings with the warning suppressed, and the result is passed to `NoteBox` as children.

Net effect: the same English text renders as before, the spurious "Missing l10n" toast on Alpha/Developer builds is gone, and the string will be picked up automatically whenever that block is uncommented.

Resolving.

src/BloomBrowserUI/teamCollection/statusPanelCommon.tsx
andrew-polk: [Claude Opus 5 from Andrew Polk's machine during preflight] Checked both halves of this; neither is a problem.

**"Affects other primary-color buttons" — not reachable.** The only other button this panel renders with `useWarningColorForButton` falsy is Reload (in the `conflictingChange` state), and it is never given a disabled state — `getBloomButton` is called for it without the `disabled` argument, so it is always enabled. A `disabledBackground` that never applies can't change how it looks. The check-out button is the only button in this panel that is ever disabled and not warning-colored, which is exactly the case this change is for.

**"Does it read as disabled?" — verified visually, and it prompted a real fix.** Running it in a Team Collection with checkouts paused showed the label greying out correctly, but the button's icon stayed bright white, so it read as only half-disabled. That's fixed in `39c6e75a3d`, which dims the icon along with the label (MUI dims a disabled button's text but not an `` we hand it). With the darker background, the greyed label, and the dimmed icon together, it reads as clearly disabled.

Resolving.

DistFiles/localization/en/Bloom.xlf
andrew-polk: [Claude Opus 5 from Andrew Polk's machine during preflight] Correction — the final fix is `8847605e28`, not the `a6b8086545` described above. Andrew reviewed the approach and preferred a simpler one, so I've replaced it.

Instead of registering the string or suppressing a warning for it, this note simply isn't localized yet, and the `.xlf` entry is gone — this branch no longer touches `Bloom.xlf` at all.

That follows the project's own rule in AGENTS.md: *"It's OK not to make XLF entries for strings only used in experimental features, as long as there is fallback English in the code that will be used."* `AllowCheckouts` is exactly that — an unreleased, admin-only setting with no UI, added in preparation for Cloud syncing. There's precedent in the UI too: `SpreadsheetExportDialog.tsx` renders a raw-English `WarningBox` for another feature that hadn't stabilized.

The underlying problem Devin identified was real and is gone either way: nothing looks up an id any more, so there's no unregistered string and no missing-l10n toast. A trans-unit can be added when the feature stabilizes and the surrounding TeamCollection block is uncommented.

andrew-polk: [Claude Opus 5 from Andrew Polk's machine during preflight] Devin's re-review of `8847605e28` still lists this (and raises a matching Investigate flag, "New XLF entry lands inside the commented-out block"), both anchored to `DistFiles/localization/en/Bloom.xlf:5682-5686`. Recording here that neither applies any more, so nobody re-opens this from the Devin page.

That trans-unit no longer exists. This branch doesn't modify `Bloom.xlf` at all — `git diff Version6.4...HEAD -- DistFiles/localization/en/Bloom.xlf` is empty, and the string `TeamCollection.CheckoutsPaused` appears nowhere in the repo. Devin appears to be carrying the finding forward from the earlier commits that did add and then remove it.

Leaving this thread resolved.

src/BloomExe/TeamCollection/TeamCollectionApi.cs
```diff
@@ -688,6 +690,18 @@ public void HandleAttemptLockOfCurrentBook(ApiRequest request)
...
+            // from a status snapshot it fetched earlier, and not every path back here re-checks it
+            // -- the post issued after the registration dialog, for one, doesn't. This is the
+            // authoritative copy of the setting, so refuse here rather than rely on the browser
+            // having current status. (The setting itself can't change mid-session: collection
+            // settings are only read as the collection opens.) See BL-16691.
+            if (!_settings.AllowCheckouts)
+            {
+                request.Failed("checkouts are paused for this collection");
+                return;
+            }
```
andrew-polk, line 708: [Devin] **Bug**: Blocked checkout can pop up an error report instead of a friendly explanation

*(Posted by Claude Opus 5 from Andrew Polk's machine during preflight — mirrored from Devin, which does not post to GitHub itself.)*

If this new guard ever refuses a checkout, the person is never told why — and on one of the two paths they get something worse than silence.

- **After the registration dialog** (`TeamCollectionBookStatusPanel.tsx`, the checkout posted once an email address is saved) the post is made with no failure handling at all. A rejected post with no handler is rethrown and ends up in Bloom's standard "something went wrong" problem-report window. So the user is told Bloom has a problem, when in fact the administrator simply paused checkouts.
- **On the ordinary button path** the failure handler only clears the busy cursor, so the click just appears to do nothing.

I've verified both: `post()` without a failure callback rethrows into `wrapAxios`, which reports the error unless told not to.

Worth saying how reachable this is, because it changes how much it matters: the button is already disabled when checkouts are paused, and the setting cannot change while Bloom is running (collection settings are only read as the collection opens). So it needs the panel to be working from a status snapshot taken before the setting was in effect. Narrow — but the registration path is exactly where a stale snapshot is most likely, since it posts without re-checking.

Left open for a decision on whether to fix now — it's in the preflight decision report as "A refused checkout tells the user nothing".

andrew-polk: [Claude Opus 5 from Andrew Polk's machine during preflight] Fixed in `c4705dc62c`. Andrew chose to handle the failure on both paths and refresh the status, so both halves of this are addressed.

**No more misleading error report.** The checkout posted after the registration dialog now has a failure handler. It does nothing, but its presence is the point: without one, `post()` rethrows and the refusal ends up in Bloom's "something went wrong" problem report, blaming Bloom for what is actually an administrator's decision.

**And the user is now told why.** As the server refuses, it sends a `bookTeamCollectionStatus` refresh, so the panel re-reads its status and redraws into the paused state — the greyed-out button with "The administrator of this collection has paused checkouts." underneath. Previously either path just left the click looking like it had done nothing.

Both comments at those sites say why they're there, so neither looks like dead code to the next reader.

Resolving.

andrew-polk: [Claude Opus 5 from Andrew Polk's machine during preflight] Devin's review of `c4705dc62c` still lists this finding. Recording that it is stale — both halves are fixed in that very commit, verified against the pushed tree:

- The registration-dialog checkout now passes a failure handler (`TeamCollectionBookStatusPanel.tsx`, the `post` at line 428), so a refusal no longer rethrows into the problem-report window.
- The endpoint sends `bookTeamCollectionStatus` / `reload` before refusing (`TeamCollectionApi.cs:704`), so the panel re-reads status and redraws into the paused state.

Devin appears to be re-reporting from the diff as a whole rather than the current state of these two lines. Leaving this resolved.

andrew-polk: [Claude Opus 5 from Andrew Polk's machine during pr-ready-for-human] The branch has been squashed to a single commit (`cceb09e064`, identical tree), and Devin re-reviewed it and re-listed this finding plus the XLF one again. Both are still stale, re-verified against the squashed commit: the branch does not modify `Bloom.xlf` at all and `TeamCollection.CheckoutsPaused` appears nowhere in the repo, and the two halves of the checkout fix are both present (the registration-path failure handler, and the status refresh sent before refusing).

Mentioning it because the Devin review page will show two bugs to anyone who opens it. Nothing outstanding.

JohnThomson (changes requested):
> @JohnThomson reviewed 8 files and all commit messages, made 3 comments, and resolved 3 discussions.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8199)* status: all files reviewed, 3 unresolved discussions (waiting on andrew-polk).

___
*[`src/BloomExe/Collection/CollectionSettings.cs` line 398 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8199#-Ozwp49r9B8CU_WiLYIy:-Ozwp49r9B8CU_WiLYIz:b-owzm3a) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/cceb09e064436a182b6c1bcd5952630ae803d6fb/src/BloomExe/Collection/CollectionSettings.cs#L398)):*
> ```Smalltalk
>             xml.Add(new XElement("District", District));
>             xml.Add(new XElement("AllowNewBooks", AllowNewBooks.ToString()));
>             // Unlike the settings around it, this one is written only when it is turned on.
> ```

The wording here is awkward. One would naturally think of "turned on" as "true", that is, permission to check out is granted. Here, however, (as explained below), "false" is the unusual case and the only one we need to actually write.
How about "...only when it has been set false to prevent checkouts"?
___
*[`src/BloomExe/Collection/CollectionSettings.cs` line 1008 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8199#-Ozwq2VI3d6RdGblgVxW:-Ozwq2VJ7ks8FQiAC2s1:bwywg8s) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/cceb09e064436a182b6c1bcd5952630ae803d6fb/src/BloomExe/Collection/CollectionSettings.cs#L1008)):*
> ```Smalltalk
>         /// no UI for this; a collection administrator pauses checkouts by hand-editing the
>         /// AllowCheckouts element of the .bloomCollection file, which then syncs to the whole team.
>         /// See BL-16691.
> ```

This is perhaps too specific about the intended use. It could be set this way; however, it's more likely that a future version of Bloom will set it as part of a process of migrating to a new implementation of team collection.
___
*[`src/BloomExe/TeamCollection/TeamCollectionApi.cs` line 698 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8199#-Ozwqrf861Zuh6zb1GWU:-Ozwqrf861Zuh6zb1GWV:bdnzjyf) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/cceb09e064436a182b6c1bcd5952630ae803d6fb/src/BloomExe/TeamCollection/TeamCollectionApi.cs#L698)):*
> ```Smalltalk
>             // authoritative copy of the setting, so refuse here rather than rely on the browser
>             // having current status. (The setting itself can't change mid-session: collection
>             // settings are only read as the collection opens.) See BL-16691.
> ```

This largely defeats the purpose of this change. Bloom can run for a long time...it may be hours or days ago that we started the session and synced and read the collection settings. If the admin really wants to stop people checking books out, and we want to force that to be so, I think we need to find a way to not only check the current content of the local collectionSettings XML file, but also notice if the shared file that contains it changes. We may already have a file watcher that covers the shared folder file; if not, I think we need to add one.
Note that in many states, we don't want to reload all collection settings; but if this one changes under us, we need to update our copy fairly soon.
(Of course things won't be instant. There will be some delay in getting the updated settings from the admin computer to this one, and this user may be offline, and there may even be some pathological situation where we can do the checkin, at least to the local dropbox folder, before we receive the collection settings change. That sort of uncertainty is inherent in the DropBox approach that we are in the process of fixing. But there should be a reasonably high probability that any user with a good, active network connection will stop being able to check out within a few minutes of the admin flipping the switch.)

src/BloomExe/TeamCollection/TeamCollection.cs
```diff
@@ -1345,6 +1345,46 @@ internal void HandleCollectionSettingsChange(RepoChangeEventArgs result)
...
+        /// administrator who pauses checkouts would have no effect on anyone still running. Here we
+        /// pick up that one setting whenever the repo's collection files change. Deliberately only
+        /// this one -- reloading all collection settings mid-session is not safe. See BL-16691.
+        ///
+        /// This is best-effort by nature: the change has to reach this machine through the shared
+        /// folder first, and the user may be offline. The aim is that anyone with a working
+        /// connection stops being able to check out within a few minutes of the switch being
+        /// flipped, not that it is instant.
+        /// </summary>
+        internal void UpdateAllowCheckoutsFromRepo()
```
andrew-polk, line 1380: [Devin] **Investigate**: In-memory pause is not written to the local settings file, so it can be lost on restart or pushed back up

*(Posted by Claude Opus 5 from Andrew Polk's machine during preflight — mirrored from Devin, which does not post to GitHub itself.)*

When we notice a remote pause we change only the in-memory setting; the local `.bloomCollection` still says checkouts are allowed. Devin's concern is that this could either be forgotten on restart, or be pushed back up to the shared folder and un-pause the whole team.

I chased both halves and neither happens:

**Lost on restart — no.** Startup runs `SyncLocalAndRepoCollectionFiles(atStartup: true)`, which copies the repo's collection files *down* over the local ones ("whatever wins in the repo wins everywhere"). So the next launch reads the administrator's paused copy and stays paused.

**Pushed back up — no, and this is the one that would have been serious.** The non-startup branch of that same method only pushes local files up when it is safe to: if the repo has been modified since our last recorded sync and the local change is an ordinary settings change, it warns the user that their settings changes will be lost and deliberately does *not* push (`TeamCollection.cs`, the `repoModTime > savedSyncTime` branch). An administrator pausing checkouts is exactly what makes the repo newer, so that guard is in force.

There is also a third path worth naming: if the user opens Collection Settings and clicks OK while paused, `Save()` writes the in-memory value, so the local file simply catches up to the repo. Consistent either way.

Resolving — but it is a fair thing to have asked, because the failure mode would have been silent and team-wide.

src/BloomExe/TeamCollection/FolderTeamCollection.cs
```diff
@@ -303,6 +304,66 @@ public override string GetRepoBookFile(string bookName, string fileName)
             return RobustZip.GetZipEntryContent(path, fileName);
         }
 
+        /// <summary>
+        /// Read AllowCheckouts straight out of the repo's copy of the collection settings file,
+        /// which lives inside "Other Collection Files.zip". We deliberately do not go through the
+        /// local copy: that is only refreshed from the repo when the collection is opened, and the
+        /// whole point is to notice a change made while Bloom is running. See BL-16691.
+        /// </summary>
+        public override bool? GetAllowCheckoutsFromRepo()
```
andrew-polk, line 313: [Devin] **Investigate**: Repo settings are opened twice per check and rely on the zip's text being BOM-free

*(Posted by Claude Opus 5 from Andrew Polk's machine during preflight — mirrored from Devin, which does not post to GitHub itself.)*

Two separate observations here.

**The BOM one was worth chasing, and it found a hole in the tests.** Bloom writes the settings file with a UTF-8 BOM — I confirmed a real `.bloomCollection` starts with `EF BB BF` — and `XElement.Parse` throws if a BOM reaches it as text. The code is in fact safe: `RobustZip.GetZipEntryContent` reads through a `StreamReader`, which detects and consumes the BOM before we parse. But every test here wrote the file by hand *without* a BOM, so nothing would have caught it if that had not been true. Fixed in `812eb33c7c`, which adds a test going through the real path (`CollectionSettings.Save()` then `CopyRepoCollectionFilesFromLocal`) and asserts the BOM is actually present before reading it back, so it can't quietly stop testing what it claims to.

**Opening the zip twice is deliberate.** Once to find the `.bloomCollection` entry by extension, once to read it. Finding it by extension rather than building the name is what makes this survive a locally renamed collection, and the alternative — hand-rolling the read to reuse the open handle — would duplicate what `RobustZip` already does robustly, including its retry. This runs at most once per change to the collection settings, so the second open is not worth that trade.

Resolving.

andrew-polk (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8199)* status: 5 of 11 files reviewed, 5 unresolved discussions (waiting on andrew-polk and JohnThomson).

___
*[`src/BloomExe/Collection/CollectionSettings.cs` line 398 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8199#-Ozwp49r9B8CU_WiLYIy:-P-0KIoJBct1fjsl6e7g:bnyux0t) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/cceb09e064436a182b6c1bcd5952630ae803d6fb/src/BloomExe/Collection/CollectionSettings.cs#L398)):*

The wording here is awkward. One would naturally think of "turned on" as "true", that is, permission to check out is granted. Here, however, (as explained below), "false" is the unusual case and the only one we need to actually write.
How about "...only when it has been set false to prevent checkouts"?
</blockquote></details>

[Claude Opus 5] Agreed, and taken as you suggested. It now reads "only when it has been set false to prevent checkouts".

You put your finger on the actual confusion: I had written "turned on" meaning "the pause is turned on", but the setting is `AllowCheckouts`, so "on" naturally reads as true/checkouts-allowed — the opposite of when we write it. Fixed in 7ceb5d76de.
___
*[`src/BloomExe/Collection/CollectionSettings.cs` line 1008 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8199#-Ozwq2VI3d6RdGblgVxW:-P-0KYn48np0RRxePGc3:bvcbgeh) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/cceb09e064436a182b6c1bcd5952630ae803d6fb/src/BloomExe/Collection/CollectionSettings.cs#L1008)):*

This is perhaps too specific about the intended use. It could be set this way; however, it's more likely that a future version of Bloom will set it as part of a process of migrating to a new implementation of team collection.
</blockquote></details>

[Claude Opus 5] Fair — and useful to know, since I had assumed hand-editing was the expected route rather than something Bloom would do itself while migrating a collection.

Andrew's call was to go further than rewording it: the comment now just states what the setting does and stops, without speculating about who sets it or why. It reads "When false, no one may check out a book in this (Team) collection. See BL-16691." Fixed in 7ceb5d76de.
___
*[`src/BloomExe/TeamCollection/TeamCollectionApi.cs` line 698 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8199#-Ozwqrf861Zuh6zb1GWU:-P-0Khaf922xgTSFos89:b-yiloxl) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/cceb09e064436a182b6c1bcd5952630ae803d6fb/src/BloomExe/TeamCollection/TeamCollectionApi.cs#L698)):*

This largely defeats the purpose of this change. Bloom can run for a long time...it may be hours or days ago that we started the session and synced and read the collection settings. If the admin really wants to stop people checking books out, and we want to force that to be so, I think we need to find a way to not only check the current content of the local collectionSettings XML file, but also notice if the shared file that contains it changes. We may already have a file watcher that covers the shared folder file; if not, I think we need to add one.
Note that in many states, we don't want to reload all collection settings; but if this one changes under us, we need to update our copy fairly soon.
(Of course things won't be instant. There will be some delay in getting the updated settings from the admin computer to this one, and this user may be offline, and there may even be some pathological situation where we can do the checkin, at least to the local dropbox folder, before we receive the collection settings change. That sort of uncertainty is inherent in the DropBox approach that we are in the process of fixing. But there should be a reasonably high probability that any user with a good, active network connection will stop being able to check out within a few minutes of the admin flipping the switch.)
</blockquote></details>

[Claude Opus 5] You're right, and thanks — this was the substantive one. My comment claimed the setting couldn't change mid-session, which was only true in the sense that we never re-read it. Since Bloom can stay open for days, an administrator pausing checkouts would have had no effect on anyone already running, which is most of the point of the feature gone.

Good news on the file watcher: we already have one. `FolderTeamCollection` watches the repo's `Other` folder — where `Other Collection Files.zip` lives — and debounces into `RepoCollectionFilesChanged`, which queues through to `HandleCollectionSettingsChange`. That handler was only writing a log message, so nothing new was needed structurally. Fixed in 7ceb5d76de:

- `TeamCollection.GetAllowCheckoutsFromRepo()` is virtual and returns null for "can't tell"; `FolderTeamCollection` reads the `.bloomCollection` entry out of the repo zip and parses it with the same `ReadBoolean` the normal load path uses, so the semantics match. It finds the entry by extension rather than building the name, so a local rename can't break it, and any failure returns null — the zip may be part-written or mid-Dropbox-sync, and leaving the setting alone is the safe outcome.
- `UpdateAllowCheckoutsFromRepo()` hangs off `HandleCollectionSettingsChange` and updates **only** this one setting, per your point that reloading all collection settings mid-session isn't safe.
- Nothing new was needed to update the UI: `HandleRemoteBookChangesOnIdle` already tells the browser to re-read book status right after that handler, so the button and its explanation update without a restart.

I took your framing on the guarantees too — the doc comment says this is best-effort, that the change has to reach the machine through the shared folder and the user may be offline, and that the aim is within a few minutes for a well-connected user rather than instant.

Two things worth flagging, since they touch your subsystem:

1. I convinced myself a stale *local* settings file can't bite us, but you'd know better. On restart, `SyncLocalAndRepoCollectionFiles(atStartup: true)` copies the repo copy down, so the pause survives. And the local-to-repo push is guarded by the `repoModTime > savedSyncTime` branch, which warns instead of pushing — and an administrator pausing checkouts is exactly what makes the repo newer. So I don't think we can push a stale "allowed" back up over their change, but please sanity-check that reasoning.
2. Six tests in `FolderTeamCollectionTests2` cover this, including one that reads the settings file as Bloom actually writes it — with a UTF-8 BOM — since the others wrote it by hand without one and so wouldn't have caught it if the BOM reached the XML parser.

src/BloomExe/Collection/CollectionSettings.cs
```diff
@@ -644,6 +655,15 @@ public void Load()
...
                 AllowNewBooks = ReadBoolean(xml, "AllowNewBooks", true);
+                // A missing element reads back as true, which is what we want: checkouts are
+                // allowed unless someone deliberately turns them off. Note that a *malformed*
+                // value does not fall back to that default -- ReadBoolean returns whatever
+                // bool.TryParse produced, which is false -- so "no" or a typo pauses checkouts for
+                // the whole team rather than being ignored. We considered special-casing this
+                // setting to be strict and decided against it: every other boolean in this file
+                // behaves the same way, and making this the one exception would be more surprising
+                // than the behavior itself. See BL-16691.
+                AllowCheckouts = ReadBoolean(xml, "AllowCheckouts", true);
```
andrew-polk, line 666: [Devin] **Bug**: A typo in the hand-edited setting silently blocks the whole team from checking out books

*(Posted by Claude Opus 5 from Andrew Polk's machine during pr-ready-for-human — mirrored from Devin, which does not post to GitHub itself.)*

`ReadBoolean` hands back whatever `bool.TryParse` produced, and a failed parse produces `false`. So a mistyped value — `yes`, `0`, a stray character — does not fall back to the documented default of "allowed". It pauses checkouts for everyone, and an administrator has no hint that the file rather than Bloom is at fault. A missing element is still safely "allowed", because there it is the default text that gets parsed.

**Decision: leave as is** (Andrew's call, recorded in the code at the read site).

The reasoning: every other true/false setting in this file has always behaved exactly this way, so making `AllowCheckouts` the single strict exception would be more surprising to the next reader than the behaviour itself. It is also not silent in the way the title suggests — the team sees the note saying an administrator has paused checkouts, so the visible outcome is "paused", not "mysteriously broken". Worth revisiting as a change to `ReadBoolean` for *all* settings if it ever bites someone, rather than as a special case here.

This is called out in the QA notes on the card so a tester confirms the behaviour instead of filing it as new.

Resolving.

andrew-polk (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8199)* status: 5 of 12 files reviewed, 6 unresolved discussions (waiting on andrew-polk and JohnThomson).

___
*[`src/BloomExe/TeamCollection/TeamCollectionApi.cs` line 698 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8199#-Ozwqrf861Zuh6zb1GWU:-P-0WHAb1G2W_N-ndzeS:b-1lgd8e) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/cceb09e064436a182b6c1bcd5952630ae803d6fb/src/BloomExe/TeamCollection/TeamCollectionApi.cs#L698)):*
<details><summary><i>Previously, andrew-polk+AI (Agent AI for andrew-polk) wrote…</i></summary><blockquote>

[Claude Opus 5] You're right, and thanks — this was the substantive one. My comment claimed the setting couldn't change mid-session, which was only true in the sense that we never re-read it. Since Bloom can stay open for days, an administrator pausing checkouts would have had no effect on anyone already running, which is most of the point of the feature gone.

Good news on the file watcher: we already have one. `FolderTeamCollection` watches the repo's `Other` folder — where `Other Collection Files.zip` lives — and debounces into `RepoCollectionFilesChanged`, which queues through to `HandleCollectionSettingsChange`. That handler was only writing a log message, so nothing new was needed structurally. Fixed in 7ceb5d76de:

- `TeamCollection.GetAllowCheckoutsFromRepo()` is virtual and returns null for "can't tell"; `FolderTeamCollection` reads the `.bloomCollection` entry out of the repo zip and parses it with the same `ReadBoolean` the normal load path uses, so the semantics match. It finds the entry by extension rather than building the name, so a local rename can't break it, and any failure returns null — the zip may be part-written or mid-Dropbox-sync, and leaving the setting alone is the safe outcome.
- `UpdateAllowCheckoutsFromRepo()` hangs off `HandleCollectionSettingsChange` and updates **only** this one setting, per your point that reloading all collection settings mid-session isn't safe.
- Nothing new was needed to update the UI: `HandleRemoteBookChangesOnIdle` already tells the browser to re-read book status right after that handler, so the button and its explanation update without a restart.

I took your framing on the guarantees too — the doc comment says this is best-effort, that the change has to reach the machine through the shared folder and the user may be offline, and that the aim is within a few minutes for a well-connected user rather than instant.

Two things worth flagging, since they touch your subsystem:

1. I convinced myself a stale *local* settings file can't bite us, but you'd know better. On restart, `SyncLocalAndRepoCollectionFiles(atStartup: true)` copies the repo copy down, so the pause survives. And the local-to-repo push is guarded by the `repoModTime > savedSyncTime` branch, which warns instead of pushing — and an administrator pausing checkouts is exactly what makes the repo newer. So I don't think we can push a stale "allowed" back up over their change, but please sanity-check that reasoning.
2. Six tests in `FolderTeamCollectionTests2` cover this, including one that reads the settings file as Bloom actually writes it — with a UTF-8 BOM — since the others wrote it by hand without one and so wouldn't have caught it if the BOM reached the XML parser.
</blockquote></details>

[Claude Opus 5] Correction to my reply above: one thing I told you was wrong.

I said "nothing new was needed to update the UI: `HandleRemoteBookChangesOnIdle` already tells the browser to re-read book status right after that handler". That isn't true, and I'd taken it from a stale comment rather than checking. Andrew tested it against a running Bloom and the panel did **not** update — the collection-level button showed "Updates Available" and the setting really had changed in memory, but the checkout button stayed live until a click forced the panel to refetch.

`InvokeSelectionChanged` doesn't send `bookTeamCollectionStatus`. Only three places do: `WorkspaceView.HandleBookStatusChange` (driven by a book-status-changed event, not a selection change), `CollectionTabView` (and only when disconnected), and the checkout endpoint itself — which is exactly why clicking appeared to fix it.

Fixed by sending it explicitly. `ITeamCollectionManager` already had `SendBookContentReload` for the analogous problem, so I added the sibling `SendBookStatusReload` rather than inventing a mechanism. I deliberately didn't raise a book-status-changed event instead, since the book's own checkout status hasn't changed and that event also drives a toast — but say the word if you'd rather it went through the existing event.

Now verified live rather than asserted: with the setting edited in the shared folder while Bloom ran, pausing and un-pausing each took effect in about 19 seconds, with no restart and no click, confirmed both in the UI and by polling `teamCollection/selectedBookStatus`.

So your comment turned out to be load-bearing twice over — not only was the setting never re-read, but even once it was, the screen wouldn't have shown it.

One housekeeping note: the branch has since been squashed, so the commit SHAs I cited in my three replies above no longer exist. Everything is now the single commit `0415a21040`.

andrew-polk (commented):
> @andrew-polk resolved 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8199)* status: 5 of 12 files reviewed, 4 unresolved discussions (waiting on andrew-polk and JohnThomson).

src/BloomExe/Collection/CollectionSettings.cs
```diff
@@ -394,6 +395,16 @@ public void Save()
...
+            // Unlike the settings around it, this one is written only when it has been set false
+            // to prevent checkouts.
+            // Save() rebuilds the whole file, so writing it unconditionally would add a line to
+            // every .bloomCollection in existence the first time its settings were saved -- and in
+            // a Team Collection that edit gets pushed to the shared folder, so the churn would
+            // spread to the whole team for a setting almost nobody uses. Omitting it is safe
+            // because a missing element already reads back as true (see Load), so a paused
+            // collection still round-trips correctly. See BL-16691.
+            if (!AllowCheckouts)
+                xml.Add(new XElement("AllowCheckouts", AllowCheckouts.ToString()));
```
andrew-polk, line 407: [Devin] **Investigate**: Hand-editing the local settings file while Bloom runs leaves the editing machine itself unpaused, and a later Save can undo the pause

*(Posted by Claude Opus 5 from Andrew Polk's machine during preflight — mirrored from Devin, which does not post to GitHub itself.)*

This was right, and it was the more dangerous half of the gap. **Fixed in `2aea26ccb1`.**

`CopyRepoCollectionFilesFromLocal` deliberately suppresses our own repo watcher while it writes, and records a fresh sync time, so `HandleCollectionSettingsChange` never fired on the machine that made the change. That meant the administrator who paused checkouts was the one person still able to check out, until they restarted.

The worse half: their in-memory `AllowCheckouts` was still `true`, so the next `Save()` — the settings dialog, or anything else that saves — would rewrite the file from that stale value. Since we deliberately omit the element when true, that erased the pause locally and pushed it back up, quietly un-pausing the whole team. The `repoModTime > savedSyncTime` guard doesn't help, because their own push had already advanced `savedSyncTime`.

The fix re-reads the setting immediately after pushing local files up, which closes both halves at once: the machine that made the change acts on it, and memory can no longer disagree with the file it is about to write.

`SyncLocalAndRepoCollectionFiles_LocalPausePushedUp_UpdatesLiveSettings` covers the whole path. I verified it fails with "the machine that made the change should be paused too" when the new call is removed, so it genuinely tests the fix rather than passing by construction.

Resolving.

src/BloomBrowserUI/teamCollection/TeamCollectionBookStatusPanel.tsx
```diff
@@ -421,6 +426,26 @@ export const TeamCollectionBookStatusPanel: React.FunctionComponent<
...
+                                            // paused checkout, so a genuine problem (shared folder
+                                            // unreachable, say) now shows nothing on this path.
+                                            // Considered distinguishing the paused case by status
+                                            // code and reporting anything else; decided against it
+                                            // because the ordinary button path above has always had
+                                            // a do-nothing failure handler too, so this matches it
+                                            // rather than introducing a new kind of silence, and
+                                            // both paths refresh status so the panel ends up
+                                            // showing the real state either way.
+                                            () => {},
```
andrew-polk, line 448: [Devin] **Bug**: Failed check-out after registering gives the user no feedback at all

*(Posted by Claude Opus 5 from Andrew Polk's machine during preflight — mirrored from Devin, which does not post to GitHub itself.)*

The checkout posted after the registration dialog has a failure handler that does nothing, so it swallows every failure on that path, not just a paused checkout. A genuine problem — the shared folder unreachable, a permissions error — now shows nothing there.

**Decision: leave as is** (Andrew's call, recorded in the code at the call site).

The handler has to exist: without one, `post()` rethrows and a refused checkout ends up in Bloom's "something went wrong" problem report, blaming Bloom for what is actually an administrator's decision. Given that, the choice was between doing nothing and adding a status-code contract between the two ends so we could stay quiet for only the paused case. We left it, because the ordinary button path has *always* had a do-nothing failure handler, so this matches it rather than introducing a new kind of silence — and both paths now refresh book status, so the panel ends up showing the real state either way.

Resolving.

src/BloomBrowserUI/teamCollection/TeamCollectionBookStatusPanel.tsx
```diff
@@ -451,7 +476,20 @@ export const TeamCollectionBookStatusPanel: React.FunctionComponent<
...
+                        belowButton={
+                            props.checkoutsArePaused ? (
+                                // Deliberately not localized yet: this is an
+                                // unreleased admin-only setting with no UI, so
+                                // per AGENTS.md we ship the English and add an
+                                // XLF entry once the feature stabilizes.
+                                <NoteBox>
+                                    The administrator of this collection has
+                                    paused checkouts.
+                                </NoteBox>
```
andrew-polk, line 490: [Devin] **Bug**: Message telling users that checkouts are paused can never appear in the user's own language

*(Posted by Claude Opus 5 from Andrew Polk's machine during preflight — mirrored from Devin, which does not post to GitHub itself.)*

Correct, and deliberate. The note renders in English in every interface language, because it has no `.xlf` entry at all.

**Decision: leave as is for now** (Andrew's call).

AGENTS.md says it is fine to skip XLF entries "for strings only used in experimental features, as long as there is fallback English in the code that will be used", which is exactly this: an unreleased, admin-only setting with no UI, added in preparation for Cloud syncing. There is precedent in the UI too — `SpreadsheetExportDialog` renders a raw-English `WarningBox` for another feature that hadn't stabilised.

It also isn't an isolated choice: every other TeamCollection string on this panel is currently inside the commented-out block in `Bloom.xlf` ("Don't want these actually translated until things stabilize"), so none of them reach translators either. A trans-unit should be added when the feature stabilises and that block is uncommented. Flagged in the QA notes on the card so a tester doesn't file the English text as a bug.

Resolving.

andrew-polk (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8199)* status: 5 of 12 files reviewed, 7 unresolved discussions (waiting on andrew-polk and JohnThomson).

___
*[`src/BloomExe/TeamCollection/TeamCollectionApi.cs` line 698 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8199#-Ozwqrf861Zuh6zb1GWU:-P-0zq8H8DTSraBeiSIh:bgjyvhp) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/cceb09e064436a182b6c1bcd5952630ae803d6fb/src/BloomExe/TeamCollection/TeamCollectionApi.cs#L698)):*
<details><summary><i>Previously, andrew-polk+AI (Agent AI for andrew-polk) wrote…</i></summary><blockquote>

[Claude Opus 5] Correction to my reply above: one thing I told you was wrong.

I said "nothing new was needed to update the UI: `HandleRemoteBookChangesOnIdle` already tells the browser to re-read book status right after that handler". That isn't true, and I'd taken it from a stale comment rather than checking. Andrew tested it against a running Bloom and the panel did **not** update — the collection-level button showed "Updates Available" and the setting really had changed in memory, but the checkout button stayed live until a click forced the panel to refetch.

`InvokeSelectionChanged` doesn't send `bookTeamCollectionStatus`. Only three places do: `WorkspaceView.HandleBookStatusChange` (driven by a book-status-changed event, not a selection change), `CollectionTabView` (and only when disconnected), and the checkout endpoint itself — which is exactly why clicking appeared to fix it.

Fixed by sending it explicitly. `ITeamCollectionManager` already had `SendBookContentReload` for the analogous problem, so I added the sibling `SendBookStatusReload` rather than inventing a mechanism. I deliberately didn't raise a book-status-changed event instead, since the book's own checkout status hasn't changed and that event also drives a toast — but say the word if you'd rather it went through the existing event.

Now verified live rather than asserted: with the setting edited in the shared folder while Bloom ran, pausing and un-pausing each took effect in about 19 seconds, with no restart and no click, confirmed both in the UI and by polling `teamCollection/selectedBookStatus`.

So your comment turned out to be load-bearing twice over — not only was the setting never re-read, but even once it was, the screen wouldn't have shown it.

One housekeeping note: the branch has since been squashed, so the commit SHAs I cited in my three replies above no longer exist. Everything is now the single commit `0415a21040`.
</blockquote></details>

[Claude Opus 5] One more thing on this thread, because it sharpens a claim I made above and you should see it before signing off.

I told you the local-to-repo push was guarded, so a stale local file couldn't be pushed back over an administrator's pause. That guard is real but **narrower than I implied**, and Devin found the hole: it doesn't cover the administrator's *own* push.

`CopyRepoCollectionFilesFromLocal` suppresses our own repo watcher while it writes and records a fresh sync time. So on the machine that made the edit, `HandleCollectionSettingsChange` never fired:

- The administrator who paused checkouts was the one person still able to check out, until they restarted.
- Worse, their in-memory `AllowCheckouts` was still `true`, so the next `Save()` — the settings dialog, or anything else — rewrote the file from that stale value. Because we only write the element while paused, that erased the pause locally and pushed it back up, quietly un-pausing the whole team. The `repoModTime > savedSyncTime` guard I cited doesn't apply, because their own push had already advanced `savedSyncTime`.

Fixed in `2aea26ccb1` by re-reading the setting straight after pushing local files up, which closes both halves: the machine that made the change acts on it, and memory can no longer disagree with the file it is about to write. `SyncLocalAndRepoCollectionFiles_LocalPausePushedUp_UpdatesLiveSettings` covers it, and I checked it fails when the new call is removed rather than passing by construction.

Two things I'd genuinely like your eye on, since this is your subsystem:

1. Re-reading after our own push is the least invasive spot I could find, but it does mean `SyncLocalAndRepoCollectionFiles` now has a side effect on `CollectionSettings`. If you'd rather that lived somewhere else — or would rather the watcher simply weren't suppressed for the settings zip — say so.
2. I've now had to correct myself twice on this thread about mechanisms I asserted from reading rather than running. The behaviour is verified end-to-end in a real Team Collection now, but treat my code-reading claims here with appropriate suspicion.

andrew-polk (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8199)* status: 5 of 12 files reviewed, 7 unresolved discussions (waiting on andrew-polk and JohnThomson).

___
*[`src/BloomExe/TeamCollection/TeamCollectionApi.cs` line 698 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8199#-Ozwqrf861Zuh6zb1GWU:-P-1By_X3NkH648CGp36:b5xybj6) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/cceb09e064436a182b6c1bcd5952630ae803d6fb/src/BloomExe/TeamCollection/TeamCollectionApi.cs#L698)):*
<details><summary><i>Previously, andrew-polk+AI (Agent AI for andrew-polk) wrote…</i></summary><blockquote>

[Claude Opus 5] One more thing on this thread, because it sharpens a claim I made above and you should see it before signing off.

I told you the local-to-repo push was guarded, so a stale local file couldn't be pushed back over an administrator's pause. That guard is real but **narrower than I implied**, and Devin found the hole: it doesn't cover the administrator's *own* push.

`CopyRepoCollectionFilesFromLocal` suppresses our own repo watcher while it writes and records a fresh sync time. So on the machine that made the edit, `HandleCollectionSettingsChange` never fired:

- The administrator who paused checkouts was the one person still able to check out, until they restarted.
- Worse, their in-memory `AllowCheckouts` was still `true`, so the next `Save()` — the settings dialog, or anything else — rewrote the file from that stale value. Because we only write the element while paused, that erased the pause locally and pushed it back up, quietly un-pausing the whole team. The `repoModTime > savedSyncTime` guard I cited doesn't apply, because their own push had already advanced `savedSyncTime`.

Fixed in `2aea26ccb1` by re-reading the setting straight after pushing local files up, which closes both halves: the machine that made the change acts on it, and memory can no longer disagree with the file it is about to write. `SyncLocalAndRepoCollectionFiles_LocalPausePushedUp_UpdatesLiveSettings` covers it, and I checked it fails when the new call is removed rather than passing by construction.

Two things I'd genuinely like your eye on, since this is your subsystem:

1. Re-reading after our own push is the least invasive spot I could find, but it does mean `SyncLocalAndRepoCollectionFiles` now has a side effect on `CollectionSettings`. If you'd rather that lived somewhere else — or would rather the watcher simply weren't suppressed for the settings zip — say so.
2. I've now had to correct myself twice on this thread about mechanisms I asserted from reading rather than running. The behaviour is verified end-to-end in a real Team Collection now, but treat my code-reading claims here with appropriate suspicion.
</blockquote></details>

[Claude Opus 5] Housekeeping so this thread stays accurate, since a fair bit has moved since my earlier replies.

**The commit SHAs I quoted above no longer exist.** The branch has been rebased onto the current Version6.4 (it had advanced with the BL-14901 work) and squashed. Everything is now the single commit `661bcc1bde`. Nothing about the content changed in that process; I checked the tree hash was identical before and after, and that my diff doesn't touch any of the BL-14901 files.

**Two things are new since I last wrote here, both from actually running it rather than reading it.**

First, the fix for your comment was incomplete when I told you it was done. The setting was being re-read, but the status panel wasn't refreshing, so the checkout button stayed live until something forced a refetch. That's covered in the correction above; it's fixed and verified.

Second, and more relevant to you: the administrator's own machine was not being paused. `CopyRepoCollectionFilesFromLocal` suppresses our own repo watcher while it writes, so on the machine that made the edit `HandleCollectionSettingsChange` never fired. Its stale in-memory value would then be written back over the pause by the next `Save()`, un-pausing the whole team. That's fixed by re-reading the setting straight after we push local files up, and it is now verified end to end in a real Team Collection by both routes: editing the shared folder, and editing the local file as an administrator actually would. Each took effect in about 20 seconds with no restart.

**On the Devin page:** it still lists five bugs and three flags, which looks alarming and isn't. Two are fixed, three record decisions Andrew made deliberately (the lenient parsing of a mistyped value, the silent failure handler on the registration path, and shipping the note in English for now), and the rest are repeats of things that no longer exist in the code. Every one has a documented outcome on a resolved PR thread, so the PR is the reliable record, not that page.

JohnThomson (approved):
> @JohnThomson reviewed 7 files and all commit messages, and resolved 4 discussions.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8199)* status: all files reviewed, 3 unresolved discussions (waiting on andrew-polk).

JohnThomson (approved):
> @JohnThomson resolved 3 discussions.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8199)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on andrew-polk).

PR #8200, opened 2026-08-13
Honor a minimum Bloom version in collection settings (BL-16690)

andrew-polk:
Honors a `MinimumBloomVersion` element in a collection's `.bloomCollection` file, in preparation for Cloud syncing. If the collection requires a newer Bloom than the one running, we refuse to open it and offer the user the two choices the card asks for: upgrade Bloom, or open a different collection.

There is no UI for setting the flag yet — it has to be added to the file by hand, as the card specifies.

### How it works

- **`CollectionSettings`** reads `MinimumBloomVersion`, and writes it back out on `Save()` when non-empty. The write-back matters: `Save()` rebuilds the file from scratch, so without it a hand-added flag would silently disappear the first time the user changed anything in Collection Settings.

- **`MinimumBloomVersionCheck`** (new) does the gating. It reads just the one element rather than loading a whole `CollectionSettings`, since we have to answer the question before committing to build a `ProjectContext`. A missing, empty, or unparseable value imposes no restriction — locking someone out of their own collection over a typo would be worse than ignoring it.

- **The gate lives in `Program.OpenProjectWindow`**, the single point all the ways of opening a collection funnel through (startup from MRU, the collection chooser, a command-line path, switching collections mid-session). Every one of those callers already responds to a `false` return by putting up the collection chooser, so "open a different collection" needed no new plumbing.

- **`HideSplashScreenForDialog`** (new, in `StartupScreenManager`) clears the splash out of the way without consuming the one-shot `DoLastOfAllAfterClosingSplashScreen`, which is what brings the main window to the front. `CloseSplashScreen()` would have spent it at the gate, so the main window would never come forward when a collection finally opened.

### Decisions worth flagging for review

- **Major.minor comparison only**, matching Bloom's other two version gates (`BookStorage`'s feature requirements and `BookDownload`'s `minVersion`). So `6.5` is satisfied by any 6.5.x or later, and a build number in the requirement is ignored. Both versions are *reported* as major.minor too, so we don't claim to enforce a build number we don't look at.

- **Channel plays no part.** A version number means the same thing on every channel, so an alpha tester running ahead of the release is correctly let in. (This was the card's open question.)

- **"Upgrade Bloom" opens the installers page and quits**, rather than doing an in-app Velopack update. Velopack reports progress solely through toasts, and `ToastHost` is mounted only in `app/App.tsx`, which does not exist before a collection opens. More decisively, Velopack only offers the latest build on the user's *own channel*, so a Release user blocked by a 6.5 requirement would be told they are up to date and left with no way forward.

- **No XLF entries** — the strings use `LocalizationManager.GetString` with English fallbacks, per the `AGENTS.md` allowance for experimental features.

### Testing

24 new tests in `MinimumBloomVersionCheckTests` cover the comparison table, reading the element, unparseable/absent files, that the setting survives a `Load`/`Save` round trip, and that the requirement is reported as major.minor. Verified in the running app that the dialog reads *'The collection "Fake TC Books" requires Bloom 6.6 or greater. You are running Bloom 6.4.'*

Note the base branch is **Version6.4**, not master.

[Devin review](https://devinreview.com/BloomBooks/BloomDesktop/pull/8200)

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8200)

### DIFF
--- a/src/BloomExe/Collection/MinimumBloomVersionCheck.cs
+++ b/src/BloomExe/Collection/MinimumBloomVersionCheck.cs
@@ -179,7 +179,7 @@ namespace Bloom.Collection
         /// </summary>
         private static void ShowDownloadsPage()
         {
-            var url = UrlLookup.LookupUrl(UrlType.LibrarySite, null) + "/installers";
+            var url = UrlLookup.LookupUrl(UrlType.LibrarySite, null) + "/downloads";
             if (SIL.PlatformUtilities.Platform.IsWindows)
                 // Let the default browser open the link.
                 ProcessExtra.SafeStartInFront(url);
--- a/src/BloomExe/Collection/MinimumBloomVersionCheck.cs
+++ b/src/BloomExe/Collection/MinimumBloomVersionCheck.cs
@@ -198,8 +198,11 @@ namespace Bloom.Collection
         /// <summary>
         /// The version of Bloom we are running. We use the assembly version rather than
         /// Application.ProductVersion because the latter can carry a non-numeric suffix.
+        /// A loaded assembly always has a version, so we don't guard against it being missing:
+        /// any fallback we picked would be a lie, and a low one would lock the user out of every
+        /// collection that declares a minimum -- the opposite of what this class is careful to do.
         /// </summary>
         private static Version RunningBloomVersion =>
-            typeof(MinimumBloomVersionCheck).Assembly?.GetName()?.Version ?? new Version(1, 0, 0);
+            typeof(MinimumBloomVersionCheck).Assembly.GetName().Version;
     }
 }
--- a/src/BloomExe/Collection/MinimumBloomVersionCheck.cs
+++ b/src/BloomExe/Collection/MinimumBloomVersionCheck.cs
@@ -26,12 +26,19 @@ namespace Bloom.Collection
         /// Decide whether this Bloom is allowed to open the given collection.
         /// </summary>
         /// <param name="settingsFilePath">Path to the .bloomCollection file</param>
-        /// <param name="minimumVersion">What the collection asked for, when we say no</param>
+        /// <param name="minimumVersion">What the collection asked for, as major.minor, when we say no</param>
         /// <returns>true if the collection demands a newer Bloom than we are</returns>
         public static bool IsThisBloomTooOld(string settingsFilePath, out string minimumVersion)
         {
-            minimumVersion = ReadMinimumBloomVersion(settingsFilePath);
-            return !IsVersionSufficient(minimumVersion, RunningBloomVersion);
+            var declaredVersion = ReadMinimumBloomVersion(settingsFilePath);
+            if (IsVersionSufficient(declaredVersion, RunningBloomVersion))
+            {
+                minimumVersion = "";
+                return false;
+            }
+            // We can only get here if it parsed, since anything we can't parse counts as satisfied.
+            minimumVersion = ToMajorMinor(Version.Parse(declaredVersion.Trim()));
+            return true;
         }
 
         /// <summary>
@@ -119,8 +126,8 @@ namespace Bloom.Collection
                 ),
                 // The message is HTML, and a collection name can perfectly well contain an ampersand.
                 System.Net.WebUtility.HtmlEncode(collectionName),
-                System.Net.WebUtility.HtmlEncode(minimumVersion),
-                RunningBloomVersion.ToString(3)
+                minimumVersion,
+                ToMajorMinor(RunningBloomVersion)
             );
             var upgradeButtonText = LocalizationManager.GetString(
                 "Collection.UpgradeBloom",
@@ -133,7 +140,13 @@ namespace Bloom.Collection
 
             var buttons = new[]
             {
-                new MessageBoxButton() { Text = upgradeButtonText, Id = kUpgradeButtonId },
+                new MessageBoxButton()
+                {
+                    Text = upgradeButtonText,
+                    Id = kUpgradeButtonId,
+                    // Warn the user that this takes them out to a web page.
+                    Icon = "launch",
+                },
                 new MessageBoxButton()
                 {
                     Text = chooseOtherButtonText,
@@ -172,6 +185,13 @@ namespace Bloom.Collection
             else
                 ProcessExtra.SafeStartInFront("xdg-open", Uri.EscapeUriString(url));
         }
+
+        /// <summary>
+        /// Format a version the way we actually compare it: major and minor only. Showing the user a
+        /// build number we don't look at would misrepresent the rule, and invites them to compare
+        /// the wrong digits ("6.5 required, but I have 6.4.900, and 900 is more than 5").
+        /// </summary>
+        private static string ToMajorMinor(Version version) => $"{version.Major}.{version.Minor}";
 
         /// <summary>
         /// The version of Bloom we are running. We use the assembly version rather than
--- a/src/BloomExe/Collection/MinimumBloomVersionCheck.cs
+++ b/src/BloomExe/Collection/MinimumBloomVersionCheck.cs
@@ -107,9 +107,9 @@ namespace Bloom.Collection
         /// upgrade, or open some other collection. If they choose to upgrade we send them to the
         /// downloads page and quit, since they can't install over a running Bloom anyway.
         /// </summary>
-        /// <remarks>If we return, the caller should abandon opening this collection; its callers
-        /// already fall back to showing the collection chooser.</remarks>
-        public static void ReportCollectionNeedsNewerBloom(
+        /// <returns>true if the user chose to upgrade, in which case we have already asked Bloom to
+        /// quit and the caller must not start any more UI.</returns>
+        public static bool ReportCollectionNeedsNewerBloom(
             string collectionName,
             string minimumVersion
         )
@@ -165,8 +165,10 @@ namespace Bloom.Collection
             {
                 ShowDownloadsPage();
                 ProgramExit.Exit();
+                return true;
             }
-            // Otherwise just return; the caller will put the collection chooser back up.
+            // Otherwise the caller will put the collection chooser back up.
+            return false;
         }
 
         private const string kUpgradeButtonId = "upgrade";
--- a/src/BloomExe/Collection/MinimumBloomVersionCheck.cs
+++ b/src/BloomExe/Collection/MinimumBloomVersionCheck.cs
@@ -1,0 +1,183 @@
+using System;
+using System.Text;
+using System.Windows.Forms;
+using System.Xml.Linq;
+using Bloom.MiscUI;
+using Bloom.ToPalaso;
+using Bloom.web;
+using L10NSharp;
+using SIL.IO;
+using SIL.Reporting;
+
+namespace Bloom.Collection
+{
+    /// <summary>
+    /// A collection can declare, via the MinimumBloomVersion element of its .bloomCollection file,
+    /// the oldest version of Bloom that is allowed to open it. This is in preparation for Cloud
+    /// syncing: once a collection has been touched by a version of Bloom that knows how to sync it,
+    /// letting an older Bloom loose on it could do real damage.
+    ///
+    /// At this point there is no UI for setting the flag; it has to be added to the file by hand.
+    /// See BL-16690.
+    /// </summary>
+    public static class MinimumBloomVersionCheck
+    {
+        /// <summary>
+        /// Decide whether this Bloom is allowed to open the given collection.
+        /// </summary>
+        /// <param name="settingsFilePath">Path to the .bloomCollection file</param>
+        /// <param name="minimumVersion">What the collection asked for, when we say no</param>
+        /// <returns>true if the collection demands a newer Bloom than we are</returns>
+        public static bool IsThisBloomTooOld(string settingsFilePath, out string minimumVersion)
+        {
+            minimumVersion = ReadMinimumBloomVersion(settingsFilePath);
+            return !IsVersionSufficient(minimumVersion, RunningBloomVersion);
+        }
+
+        /// <summary>
+        /// Compare a collection's declared minimum version against the version we are running.
+        /// We compare only major and minor, like the other version gates in Bloom
+        /// (BookStorage's feature requirements and BookDownload's minVersion), so a minimum of
+        /// "6.5" is satisfied by any 6.5.x or later, and a build number in the minimum is ignored.
+        ///
+        /// The channel (Alpha/Beta/Release) deliberately plays no part: 6.5 means the same thing on
+        /// every channel, so an alpha tester who is running ahead of the release is correctly let in.
+        /// </summary>
+        /// <remarks>Separated out from the file reading so it can be unit tested.</remarks>
+        internal static bool IsVersionSufficient(string minimumVersion, Version runningVersion)
+        {
+            if (string.IsNullOrWhiteSpace(minimumVersion))
+                return true; // the normal case: the collection doesn't care
+
+            // Version.TryParse needs at least "major.minor", so a bare "6" lands here too.
+            if (!Version.TryParse(minimumVersion.Trim(), out var requiredVersion))
+            {
+                // Someone hand-edited the file and mistyped. Locking them out of their own
+                // collection over a typo would be worse than ignoring it, so just complain to the log.
+                Logger.WriteEvent(
+                    $"Ignoring unparseable {CollectionSettings.kMinimumBloomVersionElementName} '{minimumVersion}' in collection settings."
+                );
+                return true;
+            }
+
+            if (runningVersion.Major != requiredVersion.Major)
+                return runningVersion.Major > requiredVersion.Major;
+            return runningVersion.Minor >= requiredVersion.Minor;
+        }
+
+        /// <summary>
+        /// Read just the minimum version out of the settings file. We can't use CollectionSettings
+        /// for this, because we have to answer the question before we commit to building a
+        /// ProjectContext around the collection.
+        /// </summary>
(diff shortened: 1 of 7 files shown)


src/BloomExe/Collection/MinimumBloomVersionCheck.cs
```diff
@@ -0,0 +1,208 @@
...
+                    Text = chooseOtherButtonText,
+                    Id = "chooseOther",
+                    Default = true,
+                },
+            };
+            var result = BloomMessageBox.Show(
+                null,
+                $"<strong>{header}</strong><br/><br/>{explanation}",
+                buttons,
+                MessageBoxIcon.Warning
```
andrew-polk, line 160: [Devin] **Bug**: Warning symbol is missing from the “needs a newer Bloom” dialog

The dialog asks for a warning symbol, but none appears — it shows as a plain message.

**Why:** `MessageBoxIcon.Warning` and `MessageBoxIcon.Exclamation` are the same underlying value in .NET, and `ToString()` returns the first-declared name, so what actually reaches the display side is `"exclamation"`. The Typescript only recognises `"warning"` and `"asterisk"`, so it draws nothing. This is the same trap the existing comment in `BloomMessageBox.tsx` already notes for `Information` arriving as `"Asterisk"`.

**Confirmed:** a screenshot of the running dialog during this preflight shows the text and both buttons but no warning triangle.

This is pre-existing and affects **every** `BloomMessageBox.ShowWarning` call in Bloom, not just this new dialog — so fixing it (teaching the Typescript side that `"exclamation"` means warning) would give warning icons back to warning dialogs app-wide. Because that is a visible change well beyond this card, it has been left for the developer to decide rather than changed here.

src/BloomExe/Collection/MinimumBloomVersionCheck.cs
```diff
@@ -0,0 +1,208 @@
...
+        /// every channel, so an alpha tester who is running ahead of the release is correctly let in.
+        /// </summary>
+        /// <remarks>Separated out from the file reading so it can be unit tested.</remarks>
+        internal static bool IsVersionSufficient(string minimumVersion, Version runningVersion)
+        {
+            if (string.IsNullOrWhiteSpace(minimumVersion))
+                return true; // the normal case: the collection doesn't care
+
+            // Version.TryParse needs at least "major.minor", so a bare "6" lands here too.
+            if (!Version.TryParse(minimumVersion.Trim(), out var requiredVersion))
```
andrew-polk, line 60: [Devin] **Investigate**: A bare major version like "6" silently imposes no restriction

Since the only way to set this flag is to hand-edit the collection file, someone wanting to require "Bloom 7" may well type `<MinimumBloomVersion>7</MinimumBloomVersion>`. That form needs at least a major *and* minor number to be understood, so it parses as nothing and the collection opens to everyone — the flag silently does nothing.

Erring toward letting people in is deliberate here (a typo must not lock someone out of their own collection), but a bare whole number is arguably a *correct-looking* entry rather than a typo. Normalising `7` to `7.0` before comparing would make it behave the way the person clearly meant.

Left for the developer, since it changes what an existing file would mean.

src/BloomExe/Collection/MinimumBloomVersionCheck.cs
```diff
@@ -0,0 +1,208 @@
...
+                    Icon = "launch",
+                },
+                new MessageBoxButton()
+                {
+                    Text = chooseOtherButtonText,
+                    Id = "chooseOther",
+                    Default = true,
+                },
+            };
+            var result = BloomMessageBox.Show(
```
andrew-polk, line 156: [Devin] **Investigate**: Message box is a fixed 500x200, and this message is longer than usual

Bloom's message box is a fixed 500x200 with no auto-sizing. This message is a bold heading plus a two-clause sentence, plus two fairly wide buttons — more than most.

Checked during this preflight: with the collection "Fake TC Books" it renders correctly with room to spare. The residual risk is a much longer collection name or a translated message pushing it past the fixed height.

Left as is for now (the size is a property of the shared message box, not of this dialog), but worth a look at long names.

src/BloomExe/Program.cs
```diff
@@ -1634,6 +1634,33 @@ private static bool OpenProjectWindow(string collectionPath)
                     return false;
                 }
 
+                // The collection may declare a minimum Bloom version. Check that before we go to the
+                // trouble of building a ProjectContext around it. This is the one place all the ways
+                // of opening a collection funnel through, and each caller responds to a false return
+                // by putting up the collection chooser, which is one of the two choices we offer the
+                // user here. See BL-16690.
+                if (MinimumBloomVersionCheck.IsThisBloomTooOld(path, out var minimumBloomVersion))
```
andrew-polk, line 1649: [Devin] **Investigate**: Team Collection gate reads the local settings file, which may not yet reflect the repo's requirement

The check reads the *local* copy of the collection settings, which happens before any Team Collection sync (the sync runs while the collection is being opened, which is after this point).

So if someone on a newer Bloom adds a minimum-version requirement to a Team Collection, a teammate on an older Bloom will still open it **once** — doing exactly the thing the flag exists to prevent — and only be stopped on a later start, after the settings file has come down.

This matters for the Cloud-syncing work this is preparing for, so it is flagged for the developer rather than decided here.

src/BloomExe/Program.cs
```diff
@@ -1634,6 +1634,33 @@ private static bool OpenProjectWindow(string collectionPath)
                     return false;
                 }
 
+                // The collection may declare a minimum Bloom version. Check that before we go to the
+                // trouble of building a ProjectContext around it. This is the one place all the ways
```
andrew-polk, line 1638: [Devin] **Investigate**: Only GUI collection opening is gated; the CLI/harvester/bulk-upload paths bypass the minimum version check

The gate sits in the one place all the *interactive* ways of opening a collection funnel through. But the command-line paths (create-artifacts, send-font-analytics) and bulk upload build a collection directly, without passing through it, so they will happily open — and potentially write to — a collection that declares a newer minimum.

Given the stated motivation (protecting cloud-synced collections from damage by an older Bloom), whether those paths should be gated too is a scope question for the developer.

src/BloomExe/Program.cs
```diff
@@ -1634,6 +1634,33 @@ private static bool OpenProjectWindow(string collectionPath)
...
+                // of opening a collection funnel through, and each caller responds to a false return
+                // by putting up the collection chooser, which is one of the two choices we offer the
+                // user here. See BL-16690.
+                if (MinimumBloomVersionCheck.IsThisBloomTooOld(path, out var minimumBloomVersion))
+                {
+                    // We're normally still showing the splash screen at this point, and it would sit on
+                    // top of our dialog. Also, the dialog is a ReactDialog, so it needs the server.
```
andrew-polk, line 1652: [Devin] **Investigate**: Blocked collection stays at the top of the MRU list, so the block repeats every launch

When we refuse a collection we leave it as the most-recently-used one, so starting Bloom again goes straight back to the same dialog.

That is the intended behaviour: the whole point is to tell the user their Bloom is too old, and it is not a message they should be able to lose track of by restarting. It also self-corrects — as soon as they open any other collection through "Open a Different Collection", that one becomes most-recent and the dialog stops appearing on startup.

src/BloomExe/Collection/MinimumBloomVersionCheck.cs
```diff
@@ -0,0 +1,208 @@
...
+                null,
+                $"<strong>{header}</strong><br/><br/>{explanation}",
+                buttons,
+                MessageBoxIcon.Warning
+            );
+
+            if (result == kUpgradeButtonId)
+            {
+                ShowDownloadsPage();
```
andrew-polk, line 165: [Devin] **Investigate**: Quitting after 'Upgrade Bloom' relies on Application.Exit winning the race with the queued collection chooser

Flagged against an earlier commit on this branch: after "Upgrade Bloom" we asked Bloom to quit and then still reported failure to the caller, which could put the collection chooser up *during shutdown* — a modal window that could hold the process open until the 20-second forced-shutdown watchdog fired.

src/BloomExe/Program.cs
andrew-polk: [Claude Opus 5] Not an issue — this is the intended behaviour, and it self-corrects.

Being reminded on every launch is the point: the user's Bloom really is too old for that collection, and restarting shouldn't quietly make the message go away. And it isn't a trap: the moment they open any other collection via "Open a Different Collection", that collection becomes the most-recent one and startup stops landing on the blocked one.

Resolving.

src/BloomExe/Collection/MinimumBloomVersionCheck.cs
andrew-polk: [Claude Opus 5] Already fixed, in `70fb4b31` — before this flag was raised, so the analysis describes the previous commit.

The "we're quitting" case is now reported back to the caller explicitly, and the caller returns early precisely so that nothing starts new UI on the way out. There is no longer a race to lose: the chooser is never queued on the upgrade path.

The same review pass also caught a second, worse variant of this — one caller was discarding the result entirely, which could have left Bloom running with no window at all. That is fixed in the same commit.

Resolving.

andrew-polk:
[Claude Opus 5] Consulted Devin on 2026-08-13 up to commit `a362bbec9c44babed2743656368a6b7bf262bfa1`.

It raised one bug and six things to investigate; each is its own review thread above, so they can be answered one at a time.

Two are already closed out: the collection staying at the top of the recently-used list is intended (and self-corrects once you open anything else), and the shutdown race after "Upgrade Bloom" was already fixed in `70fb4b31`, before Devin flagged it.

The other five are waiting on a developer decision — the missing warning symbol on the dialog, a bare version number like `7` silently meaning nothing, the fixed message-box size, Team Collections being judged on the local settings file before it syncs, and the command-line/bulk-upload paths not being gated.

Six further items were informational and are not mirrored here. One of those — a fallback version number that could have locked users out — was fixed in `a362bbec` (Devin's copy of that note describes the earlier commit).

CI is green, and CodeRabbit is switched off for this repo in `.coderabbit.yml`.

src/BloomExe/Collection/MinimumBloomVersionCheck.cs
andrew-polk: [Claude Opus 5] Decided: leave as is.

Someone hand-editing the file and typing a bare `7` will get no protection at all, and that is being accepted rather than special-cased. The permissive behaviour it falls into is deliberate — a value we can't read must never lock a person out of their own collection — and the flag has no user interface yet, so the audience for it is small and technical.

Worth remembering if a UI is ever added for this, since at that point the input can simply be validated instead.

src/BloomExe/Program.cs
andrew-polk: [Claude Opus 5] Not being decided here — this has been raised as an open question for the designer on the card (BL-16690).

The developer's view is that it turns on something design has to answer first: is this flag meant to *warn a person*, or to *protect the files*? As a warning, checking the local copy is fine. If the Cloud-syncing work later relies on it to keep an older Bloom from touching a collection, then the one visit it does get before the settings sync down is exactly the visit that could do the damage, and the check would have to move.

Resolving this thread; the question now lives on the card where the designer will see it.

src/BloomExe/Program.cs
andrew-polk: [Claude Opus 5] Decided: leave as is, and the reasoning is now recorded as a comment in the code next to the check.

Those tools have nobody to read a dialog or pick a different collection, so gating them would turn a silent problem into a failed overnight job. The flag today is a message to a person, and the interactive paths are where people are.

The code comment names what would have to change if that ever stops being true — if the flag needs to protect the files themselves rather than warn someone, this is the decision to revisit.

src/BloomExe/Program.cs
```diff
@@ -1634,6 +1634,42 @@ private static bool OpenProjectWindow(string collectionPath)
...
+                            Path.GetFileNameWithoutExtension(path),
+                            minimumBloomVersion
+                        )
+                    )
+                    {
+                        // The user chose to upgrade, so we've opened the downloads page and asked
+                        // Bloom to quit. Claim success, not because we opened anything, but so that
+                        // no caller puts up the collection chooser while we are shutting down. Doing
+                        // so would start a fresh modal message loop and could keep Bloom alive.
+                        // The flag lets OpenCollection tell this apart from a real open.
```
andrew-polk, line 1666: [Devin] **Bug**: A collection that refused to open is still recorded as the last-used collection when the user chooses to upgrade

To stop the collection chooser being raised while Bloom is shutting down, the "user chose to upgrade" case reports success. But the code that records the most-recently-used collection reads that as a real open, so a collection Bloom had just refused becomes the one it will try again next time.

The user-visible cost: pick a too-new collection from the chooser, click **Upgrade Bloom**, then think better of the download. Restarting the same Bloom lands you back on the blocked collection's dialog rather than the collection you were actually working in.

src/BloomExe/Program.cs
```diff
@@ -1634,6 +1634,42 @@ private static bool OpenProjectWindow(string collectionPath)
                     return false;
                 }
 
+                // The collection may declare a minimum Bloom version. Check that before we go to the
+                // trouble of building a ProjectContext around it. This is the one place all the ways
+                // of opening a collection funnel through, and each caller responds to a false return
+                // by putting up the collection chooser, which is one of the two choices we offer the
+                // user here. See BL-16690.
+                //
+                // Deliberately, this gates only the interactive ways of opening a collection. The
```
andrew-polk, line 1643: [Devin] **Investigate**: Joining a Team Collection builds a ProjectContext directly and so is not gated

Joining a Team Collection builds its own cut-down context rather than going through the gate, so an older Bloom can join a shared collection whose settings already demand a newer one.

src/BloomExe/Program.cs
andrew-polk: [Claude Opus 5] Fixed in `3f63ff9f`.

The shutdown case now carries a flag, so the code that records the most-recently-used collection can tell a real open from "handled, don't start any more UI", and leaves the list alone.

Good catch — this was a regression in an earlier fix on this branch, not in the original change.

src/BloomExe/Program.cs
andrew-polk: [Claude Opus 5] Assessed as acceptable, and consistent with the decision already recorded in the code here.

Joining does get past the gate, but the join only sets the local collection up — **opening** it goes through the normal path and is gated. So the user is still stopped, at the point where it actually matters, and with the dialog that explains why. What they lose is being told a little earlier.

This is worth distinguishing from the command-line and bulk-upload paths, which were left ungated for a different reason (no user present to answer a dialog). Here there is a user, and they do get the message — just on the next step.

Resolving. If joining should refuse outright, that belongs with the "warn a person vs. protect the files" question now on the card.

src/BloomExe/Program.cs
andrew-polk, line 1666: [Devin] **Investigate**: Startup tasks may still run after ProgramExit.Exit() in the upgrade path

Asking Bloom to quit does not stop the code that is already running. After the user picks "Upgrade Bloom" we return up through the startup machinery, and other tasks were queued behind the one that opened the collection — the registration prompt among them. If one of those opened a window after we had asked Bloom to quit, it would start a fresh modal loop and could hold Bloom open, which is the same hazard the collection chooser was just fixed for.

andrew-polk: [Claude Opus 5] Assessed as not reachable, though this one rests on reasoning rather than an observed run, so it is worth a second opinion.

Asking Bloom to quit posts a quit message to the queue. The "run a task when nothing else is happening" event that drives the startup queue only fires when that queue is *empty*, so once the quit message is sitting there the loop ends instead of idling again. By the time we return, the dialog's own modal loop has already finished, so we are back on the main loop with the quit message pending and nothing further gets picked up.

The variant that genuinely could bite — the collection chooser being raised on the way out — was real, because it was started directly on our own call stack rather than waiting for an idle moment. That is fixed in `70fb4b31`, and `3f63ff9f` cleared up a follow-on effect.

Resolving on that basis. If anyone sees Bloom linger after choosing "Upgrade Bloom", this is the first thing to re-open.

andrew-polk:
[Claude Opus 5] Consulted Devin again on 2026-08-13, up to commit `5eae3004f4130746d939b23986339272f50f468f` — the last of several rounds as the developer's decisions were carried out.

Since the previous log, Devin found one genuine regression in a fix made earlier in this same review: reporting "success" on the upgrade path meant a collection Bloom had just refused was recorded as the last-used one, so abandoning the download and restarting landed you back on the blocked collection instead of the one you were working in. Fixed in `3f63ff9f`.

It also caught two smaller things, both fixed: a fallback version number that would have blocked every flagged collection (`a362bbec`), and two test cases sharing a temp file (`5eae3004`).

Three findings were assessed and closed with the reasoning on their threads: joining a Team Collection isn't gated (but opening the result is, so the user is still stopped where it matters); the blocked collection staying top of the recently-used list (intended, and self-correcting); and whether other startup tasks can run after we ask Bloom to quit (they can't, because the quit message stops the idle processing that drives them — noted on the thread that this one rests on reasoning rather than an observed run).

Three were decided by the developer and recorded on their threads: the bare version number, the Team Collection sync ordering (now an open question for the designer on this card), and the command-line paths (left ungated, with the reasoning written into the code).

**Two threads remain open, both deliberately**: the missing warning symbol and the fixed message-box size. Both turn out to affect dialogs across Bloom rather than just this one, so the developer is deciding them separately rather than having them changed here.

CI is green. CodeRabbit is switched off for this repo in `.coderabbit.yml`.

andrew-polk:
superceded by #8206

PR #8201, opened 2026-08-13
Re-mark up the page after an Undo or Redo (BL-16558)

JohnThomson:
Follow-up to #8123: the reader and Talking Book highlights were not being adjusted after an Undo or Redo changed the text.

Since BL-16558 the tools paint their highlights as `::highlight()` pseudo-elements over live `Range` objects rather than by wrapping text in spans. An undo does not edit text in place — it writes a whole saved snapshot over the editable, rebuilding every text node in the box — so every highlight there dies at that moment: still registered, painting nothing. Only a markup pass re-creates them, and three separate things stopped one from running.

Reproduced against a running Bloom (driving the Edit tab over CDP and inspecting `CSS.highlights`, checking each Range for `collapsed` / `getClientRects()`), with the Leveled Reader tool on a page holding an over-long sentence:

| Scenario | Before |
| --- | --- |
| Top-bar **Undo** button | highlight **permanently** dead (still dead 5.8s later) |
| **Ctrl+Z** that restores a *range* selection (select a word, type over it, undo) | **permanently** dead |
| Plain **Ctrl+Z** | dead for ~550ms, then repainted |

Talking Book turned out to be largely self-healing — the `MutationObserver` added for BL-15300 repairs its current/split highlights in place — so the reader tools were where this showed. Its *markup* was equally stale after a top-bar Undo, though, since nothing re-ran it.

The three causes:

1. **No trigger.** The top-bar Undo button produces no keystroke in the page, so the keyup-driven markup update never ran at all. Same for the reader tools' own undo stack.
2. **The range-selection guard.** `selectionStateIsInvalidForMarkup` bails when the selection is not collapsed — and an undo restores whatever selection its snapshot was taken with.
3. **The longpress flag.** `jquery.longpress` sets `isLongPressEvaluating` on *every* keydown and clears it on keyup. Ctrl+Z does its work on the keydown, so an update requested from there always found the flag set and returned.

## What changed

One entry point, `updateMarkupAfterUndoOrRedo()` in `toolbox.ts`, called from every undo path we own: CKEditor's own undo/redo commands (`attachToCkEditor`), the top bar's two text-changing branches (`handleUndo`), and the reader tools' Ctrl+Z/Ctrl+Y interception.

`handlePageEditing` now takes a trigger (`"editing" | "paste" | "undoOrRedo"`) instead of a retry count. The trigger is what the four undo-specific behaviours hang off: skip the longpress flag, tolerate a range selection, run immediately instead of after the 500ms typing debounce, and wait on its own timer so the keyup that ends Ctrl+Z cannot cancel it.

## After

Top-bar Undo, Ctrl+Z, Ctrl+Z-over-a-selection and Ctrl+Y all keep the highlight alive. Sampling every 10ms found it dead only once, for ~18ms (sub-frame), on the reader-model branch. Talking Book unchanged: all four split highlights stay live through both paths.

Not unit-tested, for the reason already recorded in `handlePageEditing`: exercising it needs a live CKEditor instance on the editable plus the parent window's `page` iframe and a real `Selection`, none of which stand up in jsdom. Verified in the running app instead, as above.

Ref: https://issues.bloomlibrary.org/youtrack/issue/BL-16558

[Devin review](https://devinreview.com/BloomBooks/BloomDesktop/pull/8201)

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8201)

## This fixes a tester-reported regression on the card

The problem reported in [BL-16558 comment 102-75392](https://issues.bloomlibrary.org/youtrack/issue/BL-16558#focus=Comments-102-75392.0-0) — select highlighted text in a decodable reader, delete it, click the top-bar **Undo**, and all the highlighting is gone until you click outside the box — is exactly the top-bar-Undo path above. That report was against 6.5.1307 Alpha, and it was noted there as a regression from 6.4.107 Beta.

### DIFF
--- a/src/BloomBrowserUI/bookEdit/toolbox/toolbox.ts
+++ b/src/BloomBrowserUI/bookEdit/toolbox/toolbox.ts
@@ -1588,13 +1588,6 @@ function handlePageEditing(trigger: MarkupUpdateTrigger = "editing"): void {
             return;
         }
 
-        // Past every reason to give up, so this pass is really going to do the markup. That is
-        // what the next undo/redo has to wait behind; an attempt that bailed above (and each of
-        // its retries) did no work and must not make one wait.
-        if (isUndoOrRedo) {
-            lastUndoRedoMarkupStartTime = Date.now();
-        }
-
         // the hard thing about all this is preserving the user's insertion point while we change the actual
         // html out from under them to add/remove markup.
         // ckeditor specific discussion: http://stackoverflow.com/questions/16835365/set-cursor-to-specific-position-in-ckeditor
@@ -1627,6 +1620,13 @@ function handlePageEditing(trigger: MarkupUpdateTrigger = "editing"): void {
                 let ckeditorSelection = ckeditorOfThisBox.getSelection();
                 if (!ckeditorSelection) {
                     return; // may be changing pages?
+                }
+                // We are now certainly going to do the markup, which is the work the next
+                // undo/redo has to wait behind. Anything that gave up before this point - an
+                // unusable selection and each of its retries, no editable under the caret, a
+                // box with no ckeditor - did none of that work and must not make one wait.
+                if (isUndoOrRedo) {
+                    lastUndoRedoMarkupStartTime = Date.now();
                 }
                 // there is also createBookmarks2(), which avoids actually inserting anything. That has the
                 // advantage that changing a character in the middle of a word will allow the entire word to
--- a/src/BloomBrowserUI/bookEdit/toolbox/toolbox.ts
+++ b/src/BloomBrowserUI/bookEdit/toolbox/toolbox.ts
@@ -1479,7 +1479,7 @@ function handlePageEditing(trigger: MarkupUpdateTrigger = "editing"): void {
     // the very update the undo just asked for. These two put the choice in one place so the
     // rest of the method doesn't have to care which timer it is. Replacing whatever that same
     // timer was already waiting for is what makes a burst of events do the markup just once.
-    const cancelPendingUpdate = (): void => {
+    function cancelPendingUpdate(): void {
         if (isUndoOrRedo) {
             if (undoRedoMarkupTimer) clearTimeout(undoRedoMarkupTimer);
             undoRedoMarkupTimer = null;
@@ -1487,18 +1487,18 @@ function handlePageEditing(trigger: MarkupUpdateTrigger = "editing"): void {
         }
         if (keypressTimer) clearTimeout(keypressTimer);
         keypressTimer = null;
-    };
-    const scheduleMainTask = (
+    }
+    function scheduleMainTask(
         task: () => void,
         delayMilliseconds: number,
-    ): void => {
+    ): void {
         cancelPendingUpdate();
         if (isUndoOrRedo) {
             undoRedoMarkupTimer = setTimeout(task, delayMilliseconds);
         } else {
             keypressTimer = setTimeout(task, delayMilliseconds);
         }
-    };
+    }
     cancelPendingUpdate();
     // Not sure we need this now the method is triggered by keyup. If it is triggered by keydown,
     // we have a problem:
--- a/src/BloomBrowserUI/bookEdit/toolbox/toolbox.ts
+++ b/src/BloomBrowserUI/bookEdit/toolbox/toolbox.ts
@@ -35,6 +35,22 @@ let savedSettings: ToolboxSettings = {};
 
 let keypressTimer: ReturnType<typeof setTimeout> | null = null;
 
+// The pending markup update asked for by an undo or redo, if any. It deliberately does NOT
+// share keypressTimer: every keystroke cancels that one, and Ctrl+Z ends with a keyup, which
+// would therefore throw away the very update the undo just asked for.
+let undoRedoMarkupTimer: ReturnType<typeof setTimeout> | null = null;
+
+// When the last undo/redo markup pass began, and how close together we will let them be.
+// A single undo asks for its markup at once, because the whole point is to repaint the
+// highlights it just detached before anyone sees them missing. But holding Ctrl+Z down
+// auto-repeats the undo, and each repeat asks for another pass; with no delay to coalesce
+// them, that is a full re-markup of the page per repeat, which is what the 500ms typing
+// throttle exists to avoid. So the first one runs immediately and any that pile in behind it
+// wait their turn. (A burst is bounded anyway - ckeditor's undo stack is 20 deep - but 20
+// re-markups in half a second while ckeditor is mid-burst is still worth not doing.)
+let lastUndoRedoMarkupStartTime = 0;
+const minMillisecondsBetweenUndoRedoMarkups = 150;
+
 // This variable stores all the ids of the enabled tools, so
 // that the React toolbox settings can initially check the
 // checkboxes that correspond to the enabled tools
@@ -279,10 +295,7 @@ export class ToolBox {
                 const isPasteShortcut =
                     (event.ctrlKey || event.metaKey) && event.keyCode === 86;
                 if (isPasteShortcut) {
-                    setTimeout(
-                        () => handlePageEditing(maxPasteMarkupUpdateRetries),
-                        0,
-                    );
+                    setTimeout(() => handlePageEditing("paste"), 0);
                 }
             })
             .keyup((event) => {
@@ -1397,8 +1410,8 @@ function beginAddTool(
 }
 
 let keydownEventCounter = 0;
-const retryDelayForPasteMarkupUpdateInMilliseconds = 100;
-const maxPasteMarkupUpdateRetries = 3;
+const retryDelayForMarkupUpdateInMilliseconds = 100;
+const maxMarkupUpdateRetries = 3;
 
 export function scheduleMarkupUpdateAfterPaste(): void {
     // AI thinks we might need this to "allow the DOM to settle" even before we do the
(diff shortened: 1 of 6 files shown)


src/BloomBrowserUI/bookEdit/toolbox/toolbox.ts
```diff
@@ -1431,7 +1474,32 @@ function handlePageEditing(
...
+        task: () => void,
+        delayMilliseconds: number,
+    ): void {
+        cancelPendingUpdate();
+        if (isUndoOrRedo) {
+            undoRedoMarkupTimer = setTimeout(task, delayMilliseconds);
+        } else {
+            keypressTimer = setTimeout(task, delayMilliseconds);
+        }
+    }
```
JohnThomson, line 1501: [Devin] **Investigate**: Two independent markup timers mean two markup passes can now overlap

Previously all markup passes shared `keypressTimer`, so a new request always cancelled any pending pass and at most one pass could be pending. Now `cancelPendingUpdate()` only cancels the timer belonging to the same trigger class, so a keystroke/paste pass pending on `keypressTimer` and an undo pass pending on `undoRedoMarkupTimer` can coexist. Two consequences worth being aware of: (1) after Ctrl+Z the earlier keystroke-scheduled pass still fires ~500ms later, doing a redundant bookmark/`cleanUpNbsps`/`selectBookmarks` cycle; (2) `keydownEventCounter` is still shared and is bumped by the undo call, so an in-flight async pass (Talking Book's `updateMarkupAsync`) will find the counter changed and skip `actualUpdateFunc()` while still running the rest of its body, including restoring its own (possibly stale) bookmarks after the undo rewrote the editable's innerHTML. Neither is clearly harmful, but if cursor-jumping or "markup didn't update after paste-then-undo" reports appear, this interleaving is where to look.

---

(Claude Opus 5) Leaving this **open for John's decision** — it's in the preflight decision report.

Consequence (1) is deliberate: the extra pass is the price of repainting the highlights immediately instead of half a second later, and it is the same work the keyup pass already did before this PR.

Consequence (2) is real and is recorded as a known limitation in the commit message. Short version: on **Ctrl+Z with the Talking Book tool open**, the keyup that follows bumps the counter, so the undo pass's async markup action is discarded and the keyup's pass does the work 500ms later — i.e. exactly today's behaviour, no worse. The top-bar Undo button produces no keyup, so it does get the immediate pass. Making the undo pass immune to the counter would risk applying markup computed from *pre-keystroke* text: `getActionToMakeAudioSentenceElementsLeaf` builds its update from a `clone` taken before the await, so a stale action can overwrite a character typed in the meantime. That counter is what prevents it, which is why this wasn't "fixed" here.

src/BloomBrowserUI/bookEdit/toolbox/toolbox.ts
```diff
@@ -1468,21 +1540,29 @@ function handlePageEditing(
...
+                (selection.rangeCount === 1 &&
+                    !selection.getRangeAt(0).collapsed))
+        );
         const selectionStateIsInvalidForMarkup =
-            !active ||
-            (selection &&
-                (selection.rangeCount > 1 ||
-                    (selection.rangeCount === 1 &&
-                        !selection.getRangeAt(0).collapsed)));
+            !active || (selectionIsRange && !isUndoOrRedo);
```
JohnThomson, line 1556: [Devin] **Investigate**: Undo now marks up with a non-collapsed selection, inserting two bookmark spans

With `isUndoOrRedo`, a range selection no longer blocks the pass, so `createBookmarks(true)` inserts bookmark spans at BOTH ends of the selection before `activeTool.updateMarkup()` runs. For the reader tools that means the markup analysis sees the selected word split in two places (the documented "hous"-bookmark-"e" tradeoff, now doubled), so the decodable/leveled highlighting can be momentarily wrong across the selection until the next pass. Also note `readerToolsModel.doMarkup` is called with `createCkEditorBookMarks=false` and internally runs `doCkEditorCleanup`, which rewrites the editable's html; whether both outer bookmark spans survive that to be restored by `selectBookmarks` at the end of the pass is worth confirming for the range case specifically (the collapsed case is exercised constantly by typing).

---

(Claude Opus 5) **Checked in a running Bloom — both bookmarks do survive, and the selection is restored.** This is the scenario the change exists for, so it was one of the four cases I measured over CDP with the Leveled Reader tool active: select the last word, type a replacement, wait for the markup to settle, then Ctrl+Z. That undo restores a *range* selection (the browser reports `range:"beta."` afterwards, not a caret), which is what puts two bookmark spans into the editable. After the pass, the selection is still that same range, and the too-long-sentence highlight is alive and correctly placed — sampling the highlight every 10ms found it present continuously, never collapsed and never with zero client rects.

On the "momentarily wrong across the selection" point: the highlight measured correct in that test, and the markup is redone on the very next pass regardless, so this is the pre-existing bookmark tradeoff rather than something new — the only new part is that it can now happen once at an undo as well as while typing.

Resolving on that basis.

src/BloomBrowserUI/bookEdit/workspaceRoot.ts
```diff
@@ -107,6 +107,11 @@ export function handleUndo(): void {
     const toolboxWindow = getToolboxBundleExports();
     if (toolboxWindow && toolboxWindow.canUndo()) {
         toolboxWindow.undo();
+        // The reader tools' undo restores a saved innerHTML, which replaces the text nodes
+        // their highlights are painted over. Nothing else will notice: unlike Ctrl+Z, a click
+        // on this button produces no keystroke in the page, so the usual keyup markup update
+        // never happens and the highlights would stay dead. (BL-16558)
+        toolboxWindow.updateMarkupAfterUndoOrRedo();
         return;
```
JohnThomson, line 115: [Devin] **Investigate**: Marking up after a reader-model undo can clear the redo stack

`readerToolsModel.doMarkup()` ends by pushing onto `undoStack` and clearing `redoStack` whenever `activeElement.textContent` differs from the top of the stack (`readerToolsModel.ts:774-785`). Normally the new markup pass after an undo leaves text content untouched (highlights are ranges, not spans), so nothing is pushed. But `cleanUpNbsps()` runs first in the same pass and can convert a non-breaking space to a regular space, changing `textContent` — which would push a new undo entry and wipe the redo stack right after the user's undo, making Redo unavailable. Previously the top-bar Undo button triggered no markup pass at all, so this combination is newly reachable from that button.

---

(Claude Opus 5) Leaving this **open for John's decision** — it's in the preflight decision report.

I checked the mechanism and it is real as described: `cleanUpNbsps()` does run before `activeTool.updateMarkup()` in the same pass, and `doMarkup()` does clear `redoStack` on any `textContent` change. It needs a non-breaking space that `cleanUpNbsps` decides to convert (one not next to a regular space, not at a paragraph edge, not French-style punctuation) to be sitting in the box at the moment of the undo — narrow, but the double-space-then-delete-one case from BL-12391 produces exactly that.

Two things make me want your call rather than a fix here:

- It is confined to the reader tools' **own** undo/redo stack, not ckeditor's. That stack is only reached from the top-bar Undo button (Ctrl+Z is taken by ckeditor before the reader tool's keydown handler can claim it), and there is no Redo button in the top bar at all — so what a user would actually lose is hard to state without deciding what that stack is still for.
- That is the same question already parked on this card: whether to drop the reader tools' undo/redo now that markup no longer rewrites the text. If it goes, this dissolves with it.

If you'd rather close it here, the cheap fix is to have the undo-triggered pass tell `doMarkup()` not to push/clear — but that means threading the trigger into `readerToolsModel`, which is exactly the code that may be about to be deleted.

JohnThomson:
(Claude Opus 5) **Consulted Devin** on 2026-08-13, up to commit `3c857036a8423478da38dd724b82610a03a2fe83`.

Devin reviewed all four commits on this branch as they were pushed. Across those runs it raised one bug and three things to investigate, plus eight informational notes:

- **Bug — the two new helpers in `handlePageEditing` used `const foo = () =>` where `src/BloomBrowserUI/AGENTS.md` asks for `function` syntax.** Fixed in `3c85703`; Devin's final pass confirms it as resolved. Never needed a thread.
- **Two independent markup timers mean two markup passes can now overlap** — https://github.com/BloomBooks/BloomDesktop/pull/8201#discussion_r3778161743. Left **open** for John: part of it is a known limitation recorded in the commit message (on Ctrl+Z with Talking Book open, the keyup discards the undo pass's async result, leaving today's 500ms behaviour).
- **Marking up after a reader-model undo can clear the redo stack** — https://github.com/BloomBooks/BloomDesktop/pull/8201#discussion_r3778168535. Left **open** for John: the mechanism is real but confined to the reader tools' own undo stack, which this card already has a parked question about removing.
- **Undo now marks up with a non-collapsed selection, inserting two bookmark spans** — https://github.com/BloomBooks/BloomDesktop/pull/8201#discussion_r3778162849. Checked in a running Bloom: both bookmarks survive and the range selection is restored intact, so this is resolved.
- Two of the informational notes led to fixes anyway (`8ff980d` and `659c806`, both about when the undo throttle records a pass). The rest describe deliberate choices — e.g. that image and origami undo get no markup refresh because they don't rewrite text.

CI (`pr-automation`) is green. No comment-posting review bot is active on this repo: CodeRabbit is installed but `.coderabbit.yml` sets `auto_review.enabled: false`, and no bot has posted on any recent PR here.

src/BloomBrowserUI/bookEdit/toolbox/toolbox.ts
JohnThomson: (Claude Opus 5) **Decided: leave as is.**

John's call on the preflight report. To record the reasoning: the visible symptom this would improve is already handled — the yellow sentence highlight is repaired in place by the MutationObserver added for BL-15300, which is why Talking Book measured healthy through every undo path in testing. What the shared counter costs is only the *promptness* of Talking Book's markup refresh on Ctrl+Z, and only there: the key-up bumps the counter, the undo pass's async result is discarded, and the key-up's own pass does the work 500ms later — exactly the behaviour before this PR, so nothing regresses. The top-bar Undo button produces no key-up and does get the immediate pass.

Both ways of closing that gap mean touching a guard that exists to stop a markup action computed from a *pre-keystroke* copy of the text box from overwriting a character the user has since typed. That is a worse failure than a half-second wait, so the guard stays.

Consequence (1) in the flag — the extra ~500ms pass after Ctrl+Z — is deliberate: it is the same work the key-up pass already did before this PR, and it is the price of repainting the highlights immediately.

src/BloomBrowserUI/bookEdit/workspaceRoot.ts
JohnThomson: (Claude Opus 5) **Decided: leave as is.**

John's call on the preflight report. The mechanism is real — `cleanUpNbsps()` runs before `activeTool.updateMarkup()` in the same pass, and `doMarkup()` clears `redoStack` on any `textContent` change — but the reasoning for not acting on it here:

- It is confined to the reader tools' **own** undo list, not ckeditor's. That list is only reached from the top-bar Undo button (ckeditor takes Ctrl+Z before the reader tool's keydown handler can claim it), and there is no Redo button in the top bar at all, so it is hard to state what a user actually loses.
- It needs a non-breaking space that `cleanUpNbsps` decides to convert to be sitting in the box at the moment of the undo — reachable, but narrow.
- The fix would add plumbing to `readerToolsModel`, which is exactly the code this card has already parked a question about deleting now that markup no longer rewrites the text. If that undo/redo stack goes, this goes with it.

So it folds into that larger decision rather than being patched here.

StephenMcConnel (approved):
> @StephenMcConnel reviewed 6 files and all commit messages, and resolved 3 discussions.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8201)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on JohnThomson).

PR #8202, opened 2026-08-13
Prompt Team Collection users to sign in (BL-16692)

JohnThomson:
When Bloom opens a Team Collection while nobody is signed in to BloomLibrary.org, invite the user to sign in. Team Collection members will need an account as team collaboration moves to the web.

Per the card's answer comment, the invitation appears **every time** they open the collection until they sign in — there is deliberately no "don't show this again" setting. If the registration dialog also wants to show, both show.

### The dialog

`react_components/signInInvitationDialog.tsx`, following the mock-up on the card: title, a reassuring message, a question-mark help link to `docs.bloomlibrary.org/team-collections` (the closest page we have until a Bloom Collaboration page is written), an "I'm not able to yet" text button, and "Sign in to BloomLibrary.org" which starts the usual external-browser login. It stays open after that click and closes itself when the login state (`useLoginState`) reports an email, so finishing the login in the browser dismisses it.

### Why the front end asks instead of C# pushing

The Forum-invitation nag pushes a `LaunchDialog` websocket event from a startup action. Live testing showed that does not work here: in a Team Collection the workspace page finishes loading several seconds **after** the startup actions run (the TC sync comes first), the `collectionButtonsDrawn` milestone times out, and the pushed event is lost with nobody listening yet. So instead there is a new `GET account/signInInvitationNeeded`, which the collection tab asks for when it mounts. `AccountApi` answers "needed" at most once per run, so a page reload doesn't re-nag.

### Testing

- `AccountApiTests` — 3 tests: the invitation is offered once, and not when signed in or not a Team Collection.
- `signInInvitationDialog.spec.tsx` — 7 tests: launcher visibility and dismissal, plus the dialog's buttons and self-close.
- Live-tested in a real Team Collection with the saved login cleared: the dialog appears by itself at startup, the button opens the login page in the browser, the login callback closes the dialog and puts the avatar in the top bar, and restarting while signed in shows nothing.

Ref: https://issues.bloomlibrary.org/youtrack/issue/BL-16692

[Devin review](https://devinreview.com/BloomBooks/BloomDesktop/pull/8202)

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8202)


src/BloomExe/web/controllers/AccountApi.cs
JohnThomson, line 1: [Devin] **Investigate**: Asking the server consumes the invitation even if the answer never reaches the user

`TakeSignInInvitation` flips `_signInInvitationStillOwed` to false at the moment the GET is answered, not when the dialog is actually shown or dismissed. If the collection-tab page is torn down or reloaded between the launcher's mount-time GET and the dialog appearing (a full Vite reload in dev, or any C#-initiated reload of the collection tab shortly after startup), the invitation is silently spent and the user sees nothing for the rest of that Bloom run — which conflicts with the stated goal of nagging until they sign in. Since the invitation is deliberately per-run, an alternative would be to consume it only when the front end reports that it showed/dismissed the dialog, or simply to answer from live state (TC + not logged in) and let the front-end `closed` state suppress re-showing within a page session.

JohnThomson: [Claude Opus 5] Fixed in `a21971b8ce507d48324b8cb56e3fe9531554a5bb`.

Good catch, and it is exactly the hole it looks like: the run only ever offers one invitation, so if the page went away in the moment between the launcher asking and the dialog appearing, the user would never be invited at all that run. Now the front end posts `account/signInInvitationShown` once the dialog is actually on screen, and only that uses the invitation up. Asking is a plain read.

We did not take the second suggestion (drop the server-side memory and let the front end's own `closed` state suppress re-showing), because `App.tsx` renders `<CollectionsTabPane/>` only while the collection tab is active — visiting the Edit or Publish tab unmounts it, so any memory it kept is lost. Going back to the collection tab is not opening the collection again and must not bring the dialog back, so the server has to be the one that remembers.

Verified live in a Team Collection: the dialog comes up at startup, and once it has been shown `GET account/signInInvitationNeeded` answers `{"needed":false}`, which is what a remount reads.

DistFiles/localization/en/BloomMediumPriority.xlf
JohnThomson, line 1: [Devin] **Investigate**: Priority file choice for strings every Team Collection user sees at startup

These five strings were added to `BloomMediumPriority.xlf`. Per `.github/skills/xlf-strings/SKILL.md`, `Bloom.xlf` is for "the most important things in our UI that users must understand"; this dialog is shown to every not-signed-in Team Collection user on every startup until they sign in, which arguably makes it high priority. Worth confirming with the string owner (the skill says to ask which file to use).

JohnThomson: [Claude Opus 5] Not an issue — this was already decided by the string owner.

The `xlf-strings` skill does say to ask, and we did: before writing the strings, John was offered `Bloom.xlf` (recommended, since every not-signed-in Team Collection user meets this dialog at every startup) versus `BloomMediumPriority.xlf`, and he chose `BloomMediumPriority.xlf`. Leaving them there.

src/BloomExe/web/controllers/AccountApi.cs
JohnThomson, line 1: [Devin] **Bug**: Sign-in prompt silently skipped instead of failing when the team-collection service is missing

The team-collection manager is accessed defensively (`_tcManager?.CurrentCollectionEvenIfDisconnected` at `src/BloomExe/web/controllers/AccountApi.cs:164`) even though it is a required constructor dependency, so a wiring mistake would quietly turn the invitation off instead of surfacing an error.
Impact: If the dependency is ever not supplied, Team Collection users would silently never be asked to sign in, and nobody would learn why.

Why this conflicts with the repository's fail-fast rule: AGENTS.md states "Don't be overly defensive about error handling. If the API is used incorrectly, it's fine for it to throw an error" and "Fail Fast. Don't write code that silently works around failed dependencies... if you expect a foo to be defined, don't write 'if(foo){}'". `_tcManager` is injected by autofac (`src/BloomExe/ProjectContext.cs:238-241` registers `ITeamCollectionManager`) and is never null in production, so the null-conditional adds nothing but a silent failure mode. Note the existing code in `src/BloomExe/TeamCollection/TeamCollectionApi.cs:443` uses `_tcManager.CurrentCollectionEvenIfDisconnected != null` — the manager itself is dereferenced unconditionally.

JohnThomson: [Claude Opus 5] Fixed in `c0edab3e2f5c32a95967c6061add6ca2536aa1a4`. Agreed — the manager is always supplied by autofac, so the `?.` bought nothing but a silent way for the whole prompt to stop working. It is now a plain dereference, matching TeamCollectionApi. (The same commit also updates the `RegisterWithApiHandler` summary, which still listed only the three original `account/*` endpoints.)

JohnThomson:
[Claude Opus 5 from Thomson's machine during preflight] Consulted Devin on 2026-08-13, up to commit `c0edab3e2f5c32a95967c6061add6ca2536aa1a4`.

Three reviews ran across this branch's commits, and everything it raised has been dealt with:

- **1 Bug** — the sign-in check was defensive about the team collection manager, which autofac always supplies, so a wiring mistake would have quietly stopped inviting anyone instead of failing. Fixed; Devin's own re-review now marks it resolved. [Thread](https://github.com/BloomBooks/BloomDesktop/pull/8202#discussion_r3779215994)
- **Investigate: the invitation was used up by the question, not by the dialog** — a real hole, since the run only offers one invitation. Now the front end reports back once the dialog is actually on screen. [Thread](https://github.com/BloomBooks/BloomDesktop/pull/8202#discussion_r3779172031)
- **Investigate: which XLF priority file the new strings belong in** — already settled by the string owner (John chose `BloomMediumPriority.xlf` when asked during planning). [Thread](https://github.com/BloomBooks/BloomDesktop/pull/8202#discussion_r3779175302)

It also raised 9 informational items (not mirrored as threads, but each looked at). Two led to changes: the `RegisterWithApiHandler` comment no longer listed all the endpoints, and the new C# tests were writing fake login data into the developer's real settings — both fixed. Two were mistaken: `fail()` *is* provided to Vitest tests by `vitest.setup.ts`, and the translator note about capital letters is accurate, because dialogs use `lightTheme`, which keeps Material's uppercase buttons (only `toolboxTheme` turns that off). The rest were observations we're comfortable with: the invitation is deliberately spent once the user has seen the dialog, even if the page then reloads; the restored-login token is validated in the background; and a plain bool is enough for a per-run flag that only one page ever asks about.

CI (`pr-automation`) is green. CodeRabbit posted nothing, which on this repo means the team's free monthly allowance is used up rather than anything about this branch.

JohnThomson:
[Claude Opus 5] Consulted Devin on 2026-08-14, up to commit `bb2dd66195748b8f834eaad5fa06db5b6b392e3e` — extending the log above, which covered `c0edab3e`. Devin's review of the final commit is complete and raised nothing new: the one bug it found earlier is now marked resolved by Devin itself, and both Investigate flags have documented, resolved threads. CI (`pr-automation`) is green.

andrew-polk (changes requested):
> @andrew-polk reviewed 7 files and all commit messages, made 1 comment, and resolved 3 discussions.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8202)* status: all files reviewed, 1 unresolved discussion (waiting on JohnThomson).

___
*[`DistFiles/localization/en/BloomMediumPriority.xlf` line 1633 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8202#-P-0ETss1RMtMqUiPbXB:-P-0ETss1RMtMqUiPbXC:b-hth8sm) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/bb2dd66195748b8f834eaad5fa06db5b6b392e3e/DistFiles/localization/en/BloomMediumPriority.xlf#L1633)):*
> ```XML
>         <note>ID: SignInInvitationDialog.SignIn</note>
>         <note>Label of the main button in the dialog inviting a Team Collection user to sign in; it opens the BloomLibrary.org login page in the user's web browser. "BloomLibrary.org" is a web address and must not be translated. Shown in capital letters by the button itself, so do not capitalize it here.</note>
>       </trans-unit>
> ```

So far, we haven't localized anything for Team Collections, so it seems like we don't want to add these, especially since I think this is a one-version dialog?

JohnThomson (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8202)* status: 6 of 8 files reviewed, 1 unresolved discussion (waiting on andrew-polk).

___
*[`DistFiles/localization/en/BloomMediumPriority.xlf` line 1633 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8202#-P-0ETss1RMtMqUiPbXB:-P-0u6xH6Z3rUWAyUo8w:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/bb2dd66195748b8f834eaad5fa06db5b6b392e3e/DistFiles/localization/en/BloomMediumPriority.xlf#L1633)):*

So far, we haven't localized anything for Team Collections, so it seems like we don't want to add these, especially since I think this is a one-version dialog?
</blockquote></details>

Done.

andrew-polk (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8202)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on JohnThomson).

PR #8203, opened 2026-08-13
Stop SelectBook persisting the selected book, and save settings promptly (BL-16660)

JohnThomson:
`BookSelection.SelectBook()` did more than its name suggests: it ended by writing
`Settings.Default.CurrentBookPath` and saving. Every caller therefore wrote persisted,
process-wide user settings, whether or not that made sense:

- **Unit tests** that selected a book in a `TemporaryFolder` left later tests looking at a folder
  that no longer existed (BL-16660 documents 47 such failures, found by bisection).
- The **command-line bulk uploader** builds a throwaway `BookSelection` per book just to satisfy
  constructors, so a `bloom upload` run replaced the user's remembered book with each book it
  uploaded.
- The write was **unfiltered** while the restore path is heavily filtered, so the reader had grown
  a pile of defensive cleanup (BL-11678, BL-16327) to cope with paths that should never have been
  stored.

## What changed

Remembering the selection now belongs to `WorkspaceView`, as a handler on `BookSelection`'s
low-priority `SelectionChanged` event, guarded by the same `IsSelectedBookObsoleteOrInvalid`
predicate `SelectBookAtStartup` applies when restoring:

- tests and console mode have no `WorkspaceView`, so they no longer write settings at all;
- a book outside the current and source collections is not stored, since startup would refuse to
  restore it anyway — and an existing good value is left alone rather than overwritten, because the
  main way to reach that case is the stale cross-collection selection of BL-14313;
- a `null` selection still clears the setting, as before;
- `SaveCurrentBookPath` skips the write when the value is unchanged, so startup restore and
  re-selection no longer rewrite the settings file.

It goes on the *low*-priority list deliberately: the high-priority list exists so button
highlighting responds quickly, and this writes a file.

## Also fixed: settings changed but never saved

Several places assigned a `Settings.Default` property and never called `Save()`. They were relying
on some later, unrelated save to flush them — most often the per-selection save this PR removes,
which was never a guarantee in either direction. Our intent is to save changed settings
immediately, so each now saves its own:

- `WorkspaceView.SelectBookAtStartup`'s catch block — BL-11678's fix nulls the corrupt book path
  but never persisted it, so **Bloom would meet the same corrupt book again on the next launch**.
- `EditingView`'s book-renamed fixup (`CurrentBookPath`).
- `ExperimentalFeatures.SetValue` and `MigrateFromOldSettings` — without the latter, the
  "once and once only" migration could run again.
- `WorkspaceView`'s two "default the UI language to en" normalizations, restructured so they only
  write when the value actually changes.
- `CollectionSettingsDialog` (`AutoUpdate`), `PublishToBloomPubApi` (`PublishAndroidMethod`),
  `Shell.SetAlwaysMeasurePerformance`.

### Two places where "save immediately" is the wrong rule

Both were tried and deliberately reverted, because they run on **every** startup path — including
the command-line verbs — and so would reintroduce, in miniature, the very problem this PR fixes.
Each carries a comment saying so:

- **`Program.SetUpLocalization`** (`UserInterfaceLanguage`, `UserInterfaceLanguageSetExplicitly`).
  It runs before `Main` sets `RunningInConsoleMode` and dispatches the verbs, so saving made `bloom
  upload`/`hydrate` rewrite the user's settings file, and could fail an unattended run at startup.
  It also runs before the settings migration, whose `Reload()` discards these values anyway. Nothing
  is lost: these are derived at startup rather than chosen by the user, `GetDesiredUiLanguage`
  derives them again next launch, and `ApplyUiLanguageChange` already saves immediately when the
  user actually picks a language.
- **`ApplicationContainer` (`MruProjects`)** — the same trap: all five CLI verbs construct an
  `ApplicationContainer`, so on a profile with no MRU list yet (a harvester or service account) a
  CLI run would rewrite the whole settings file. The list is registered with the container either
  way, and the GUI saves it when it adds a path.

Three near-identical assign-save-log blocks in `WorkspaceView` now share `SaveCurrentBookPath`.

`SaveCurrentBookPath` enumerates the ways saving can fail — `IOException`,
`UnauthorizedAccessException`, `ArgumentException` (thrown when the path contains a surrogate pair,
i.e. a book folder named with an emoji, which is why the shutdown code at the end of `Program.Run`
copes with the same thing) and `ConfigurationErrorsException` (a corrupt `user.config`) — because
remembering the book must never interrupt what the user was doing. Crucially it **puts the old value
back** when the save fails: `Save()` writes the whole settings object, so leaving an unsaveable value
in it would make every later save anywhere in Bloom fail the same way, turning this into a broken
Collection Settings dialog, publish tab or rename. `PersistSelectedBookPath` guards its own body too,
since deciding whether a book is worth remembering reads collection folders from disk.

## Tests

New `BookSelectionTests.SelectBook_DoesNotChangeCurrentBookPathSetting` locks in the fix for both a
book and `null`, with sanity checks that the selection really changed. It fails against the old code.

`ExperimentalFeaturesTests` no longer assumes `EnabledExperimentalFeatures` starts out empty. That
held only because nothing persisted it; now that `SetValue` saves, a run that died part way through
(AGENTS.md documents `*** TEST RUN ABORTED ***` as a real occurrence) left a value behind that
failed the fixture on **every** later run until someone hand-edited the test host's `user.config` —
the same hazard BL-16660 removes from `SelectBook`, and it would have been reintroduced here. The
fixture now establishes its starting value and restores what it found. Verified by injecting a
leftover value into the test host config: the reworked fixture passes, the old one fails with
`But was: "leftover-from-aborted-run"`.

Full C# suite green (3093 passed, 0 failed) before these last two fixes; re-run against final HEAD
reported in the preflight summary.

Ref: https://issues.bloomlibrary.org/youtrack/issue/BL-16660

---
[Devin review](https://devinreview.com/BloomBooks/BloomDesktop/pull/8203)

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8203)

### DIFF
--- a/src/BloomExe/ApplicationContainer.cs
+++ b/src/BloomExe/ApplicationContainer.cs
@@ -41,8 +41,13 @@ namespace Bloom
 
             if (Settings.Default.MruProjects == null)
             {
+                // Deliberately not saved: every CLI verb constructs an ApplicationContainer too
+                // (see the RunningInConsoleMode test below), so saving here would let `bloom
+                // upload`, `hydrate` and friends rewrite the user's settings file on any machine
+                // whose profile has no MRU list yet - a harvester or service account, say - which
+                // is the very thing BL-16660 is about. Nothing needs it: the list is registered
+                // with the container either way, and the GUI saves it when it adds a path.
                 Settings.Default.MruProjects = new MostRecentPathsList();
-                Settings.Default.Save();
             }
             builder.RegisterInstance(Settings.Default.MruProjects).SingleInstance();
 
--- a/src/BloomExe/ApplicationContainer.cs
+++ b/src/BloomExe/ApplicationContainer.cs
@@ -42,6 +42,7 @@ namespace Bloom
             if (Settings.Default.MruProjects == null)
             {
                 Settings.Default.MruProjects = new MostRecentPathsList();
+                Settings.Default.Save();
             }
             builder.RegisterInstance(Settings.Default.MruProjects).SingleInstance();
 
--- a/src/BloomExe/Program.cs
+++ b/src/BloomExe/Program.cs
@@ -569,6 +569,7 @@ namespace Bloom
                             Settings.Default.MruProjects = new MostRecentPathsList();
                         }
 
+                        // The Save() below covers both this and the list we may just have created.
                         Settings.Default.MruProjects.AddNewPath(newCollection);
                         Settings.Default.Save();
 
@@ -2048,6 +2049,9 @@ namespace Bloom
                 );
 
                 Settings.Default.UserInterfaceLanguage = LocalizationManager.UILanguageId;
+                // Saves this and the UserInterfaceLanguageSetExplicitly set above; GetDesiredUiLanguage
+                // reads both on the next launch.
+                Settings.Default.Save();
 
                 // Per BL-6449, these two languages should try Spanish before English if a localization is missing.
                 // (If they ever get localized enough to show up in our list.)
--- a/src/BloomExe/Program.cs
+++ b/src/BloomExe/Program.cs
@@ -2050,8 +2050,13 @@ namespace Bloom
 
                 Settings.Default.UserInterfaceLanguage = LocalizationManager.UILanguageId;
                 // Saves this and the UserInterfaceLanguageSetExplicitly set above; GetDesiredUiLanguage
-                // reads both on the next launch.
-                Settings.Default.Save();
+                // reads both on the next launch. But not before Main has migrated the previous
+                // version's settings: we run well before its "if (NeedUpgrade) { Upgrade();
+                // Reload(); }" block, and Reload() would throw away what we wrote anyway, so saving
+                // now would gain nothing while writing the new version's settings file earlier than
+                // the migration expects.
+                if (!Settings.Default.NeedUpgrade)
+                    Settings.Default.Save();
 
                 // Per BL-6449, these two languages should try Spanish before English if a localization is missing.
                 // (If they ever get localized enough to show up in our list.)
--- a/src/BloomExe/Program.cs
+++ b/src/BloomExe/Program.cs
@@ -2049,14 +2049,16 @@ namespace Bloom
                 );
 
                 Settings.Default.UserInterfaceLanguage = LocalizationManager.UILanguageId;
-                // Saves this and the UserInterfaceLanguageSetExplicitly set above; GetDesiredUiLanguage
-                // reads both on the next launch. But not before Main has migrated the previous
-                // version's settings: we run well before its "if (NeedUpgrade) { Upgrade();
-                // Reload(); }" block, and Reload() would throw away what we wrote anyway, so saving
-                // now would gain nothing while writing the new version's settings file earlier than
-                // the migration expects.
-                if (!Settings.Default.NeedUpgrade)
-                    Settings.Default.Save();
+                // Deliberately NOT saved here, unlike the other settings this branch made save
+                // promptly. This runs on every startup path, and very early: before Main sets
+                // RunningInConsoleMode and dispatches the command-line verbs, so saving would make
+                // `bloom upload`, `hydrate` and friends rewrite the user's settings file - a smaller
+                // version of the very problem BL-16660 is about - and would risk failing an
+                // unattended run at startup. It is also before Main migrates the previous version's
+                // settings, whose Reload() would discard whatever we wrote anyway.
+                // Nothing is lost: this is a value derived at startup, not a user's choice, and
+                // GetDesiredUiLanguage derives it again next launch. When the user actually picks a
+                // language, WorkspaceView.ApplyUiLanguageChange saves both settings immediately.
 
                 // Per BL-6449, these two languages should try Spanish before English if a localization is missing.
                 // (If they ever get localized enough to show up in our list.)
--- a/src/BloomExe/Workspace/WorkspaceView.Designer.cs
+++ b/src/BloomExe/Workspace/WorkspaceView.Designer.cs
@@ -19,6 +19,14 @@ namespace Bloom.Workspace
 		{
 			if (disposing)
 			{
+				// _bookSelection outlives us (it is application-level, and we are recreated when
+				// the user switches collections), so stop it calling us. It matters for this
+				// subscriber in particular: deciding whether a selection is worth remembering
+				// depends on _collectionSettings, which for a disposed view describes the
+				// collection we have just left. See BL-16660.
+				if (_bookSelection != null)
+					_bookSelection.SelectionChanged -= PersistSelectedBookPath;
+
 				Current = null;
 				_editingView?.Dispose();
 				_editingView = null;
--- a/src/BloomExe/Workspace/WorkspaceView.Designer.cs
+++ b/src/BloomExe/Workspace/WorkspaceView.Designer.cs
@@ -20,12 +20,14 @@ namespace Bloom.Workspace
 			if (disposing)
 			{
 				// _bookSelection outlives us (it is application-level, and we are recreated when
-				// the user switches collections), so stop it calling us. It matters for this
-				// subscriber in particular: deciding whether a selection is worth remembering
-				// depends on _collectionSettings, which for a disposed view describes the
-				// collection we have just left. See BL-16660.
+				// the user switches collections), so stop it calling us. Both handlers depend on
+				// _collectionSettings and _model, which for a disposed view describe the collection
+				// we have just left, so neither should still be answering. See BL-16660.
 				if (_bookSelection != null)
+				{
 					_bookSelection.SelectionChanged -= PersistSelectedBookPath;
+					_bookSelection.SelectionChangedHighPriority -= HandleBookSelectionChanged;
+				}
 
 				Current = null;
 				_editingView?.Dispose();
--- a/src/BloomExe/Workspace/WorkspaceView.cs
+++ b/src/BloomExe/Workspace/WorkspaceView.cs
@@ -210,6 +210,13 @@ namespace Bloom.Workspace
             // We'll need to do something even trickier if there start to be slow things that
             // happen in response to the book selection changed websocket message.
             bookSelection.SelectionChangedHighPriority += HandleBookSelectionChanged;
+
+            // Remembering the selection for the next launch belongs here rather than in
+            // SelectBook() (BL-16660). It goes on the ordinary (low) priority list because saving
+            // settings writes a file, and per the comment above the button highlighting must not
+            // wait for that.
+            bookSelection.SelectionChanged += PersistSelectedBookPath;
+
             bookStatusChangeEvent.Subscribe(args =>
             {
                 HandleBookStatusChange(args);
@@ -316,8 +323,7 @@ namespace Bloom.Workspace
                     return;
                 }
 
-                Settings.Default.CurrentBookPath = resolvedPath;
-                Settings.Default.Save();
+                SaveCurrentBookPath(resolvedPath);
             }
             catch (Exception e) when (e is IOException || e is UnauthorizedAccessException)
             {
@@ -362,21 +368,69 @@ namespace Bloom.Workspace
             }
         }
 
+        /// <summary>
+        /// Remember the newly selected book, so that SelectBookAtStartup can restore it next time.
+        /// </summary>
+        /// <remarks>
+        /// BookSelection.SelectBook() used to do this itself, which meant that anything selecting a
+        /// book wrote global, persisted settings: unit tests left later tests looking at books in
+        /// deleted temporary folders, and the command-line bulk uploader replaced the user's
+        /// remembered book with each book it uploaded. See BL-16660. Doing it here means only the
+        /// running UI persists a selection, and only one that startup could actually make use of.
+        /// </remarks>
+        private void PersistSelectedBookPath(object sender, BookSelectionChangedEventArgs e)
+        {
+            var book = _bookSelection.CurrentSelection;
+            if (book == null)
+            {
+                // Nothing is selected (e.g. the selected book was just deleted), so there is
+                // nothing to restore next time.
+                SaveCurrentBookPath("");
+                return;
+            }
+            // SelectBookAtStartup applies this same test and refuses to restore a book that fails
+            // it, so storing such a path could only do harm: a path we can't use is a known source
+            // of trouble (BL-11678, BL-16327). We leave any good value already stored alone rather
+            // than clearing it, because the main way to get here is the stale selection left over
+            // from the previous collection while we are switching collections (BL-14313).
+            if (IsSelectedBookObsoleteOrInvalid(book.FolderPath))
+                return;
+            SaveCurrentBookPath(book.FolderPath);
+        }
+
+        /// <summary>
+        /// Store a new value for the remembered book path, doing nothing if it is already correct.
+        /// </summary>
+        private static void SaveCurrentBookPath(string path)
+        {
+            if (Settings.Default.CurrentBookPath == path)
+                return; // no change, so no need to rewrite the settings file
+            try
+            {
+                Settings.Default.CurrentBookPath = path;
+                Settings.Default.Save();
+            }
+            catch (Exception e)
+                when (e is IOException || e is UnauthorizedAccessException || e is ArgumentException
+                )
+            {
+                // Failing to remember the book is not worth interrupting the user's work over.
+                // ArgumentException belongs here because Save() throws it when the path contains a
+                // surrogate pair, i.e. the book's folder name contains an emoji or similar; the
+                // shutdown code at the end of Program.Run has to cope with the same thing. Without
+                // this, selecting such a book would throw out of SelectBook to whoever called it,
+                // and the recovery path in SelectBookAtStartup below could not do its job.
+                Logger.WriteError("Unable to save the current book path.", e);
+            }
+        }
+
         private static void ClearCurrentBookPathIfMissing()
         {
             var currentBookPath = Settings.Default.CurrentBookPath;
             if (string.IsNullOrEmpty(currentBookPath) || Directory.Exists(currentBookPath))
                 return;
 
-            try
-            {
-                Settings.Default.CurrentBookPath = null;
-                Settings.Default.Save();
-            }
-            catch (Exception e) when (e is IOException || e is UnauthorizedAccessException)
-            {
-                Logger.WriteError("Unable to clear stale current book path.", e);
-            }
+            SaveCurrentBookPath(null);
         }
 
         private static ReactControlAdditionalHtml GetWorkspaceAdditionalHtml()
@@ -564,8 +618,8 @@ window.showWorkspaceInitializationFailure = function(message) {
                 // We certainly don't want to crash because we had a problem doing so.
                 // One scenario we know of which causes this is if the book at
                 // Settings.Default.CurrentBookPath gets corrupted, such as having no .htm file.
-                // See BL-11678.
-                Settings.Default.CurrentBookPath = null;
+                // See BL-11678. Save it, or we would meet the same book again next launch.
+                SaveCurrentBookPath(null);
 
                 MiscUtils.SuppressUnusedExceptionVarWarning(e);
             }
@@ -812,7 +866,10 @@ window.showWorkspaceInitializationFailure = function(message) {
             if (onlyActiveItem)
             {
                 if (String.IsNullOrEmpty(Settings.Default.UserInterfaceLanguage))
+                {
                     Settings.Default.UserInterfaceLanguage = "en"; // See BL-13545.
+                    Settings.Default.Save();
+                }
                 items.Add(CreateLanguageItem(Settings.Default.UserInterfaceLanguage));
             }
             else
@@ -965,8 +1022,12 @@ window.showWorkspaceInitializationFailure = function(message) {
         {
             var current = Settings.Default.UserInterfaceLanguage;
             if (String.IsNullOrEmpty(current))
+            {
+                // Store the default we are falling back to, so the rest of Bloom sees it too.
                 current = "en";
-            Settings.Default.UserInterfaceLanguage = current;
+                Settings.Default.UserInterfaceLanguage = current;
+                Settings.Default.Save();
+            }
             return current;
         }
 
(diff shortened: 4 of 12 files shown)


src/BloomExe/Workspace/WorkspaceView.cs
```diff
@@ -362,23 +368,71 @@ private static bool TryGetBookIdFromFolder(string folderPath, out string bookId)
...
             {
-                Logger.WriteError("Unable to clear stale current book path.", e);
+                // Failing to remember the book is not worth interrupting the user's work over.
+                // ArgumentException belongs here because Save() throws it when the path contains a
+                // surrogate pair, i.e. the book's folder name contains an emoji or similar; the
+                // shutdown code at the end of Program.Run has to cope with the same thing. Without
+                // this, selecting such a book would throw out of SelectBook to whoever called it,
+                // and the recovery path in SelectBookAtStartup below could not do its job.
+                Logger.WriteError("Unable to save the current book path.", e);
             }
```
JohnThomson, line 462: [Devin] **Bug**: Selecting a book whose folder name contains an emoji can silently poison saved preferences and make later preference changes crash

In plain terms: if a book's folder name contains an emoji (or any other non-BMP character), selecting
it fails to save — which this code deliberately tolerates — but the unsaveable name is left sitting
in Bloom's in-memory preferences. From then on, *every* attempt to save *any* preference in that
session hits the same failure. So after clicking such a book, changing the BloomPUB publish method,
clicking OK in Collection Settings, toggling an experimental feature, or renaming a book can throw an
error dialog instead of quietly working.

This PR makes it worse rather than better: it adds several new "save immediately" calls on exactly
those ordinary user paths, so there are now more places for the poisoned value to surface.

<details>
<summary>Mechanism: the swallowed ArgumentException leaves the bad path in the in-memory settings object</summary>

`Settings.Default.Save()` throws `ArgumentException` ("surrogate") when any stored string contains a
surrogate pair — which is precisely what the shutdown code at `src/BloomExe/Program.cs:1459-1473`
exists to compensate for, by resetting `CurrentBookPath` to `""` and re-saving.

`SaveCurrentBookPath` assigns the value first and then catches the `ArgumentException` from `Save()`,
but does not restore the previous value. The in-memory settings object therefore keeps the poison
path for the rest of the session, and because `Save()` writes the whole object, every subsequent
`Settings.Default.Save()` in the process throws the same `ArgumentException`. Newly-added unguarded
saves on ordinary user paths:

- `src/BloomExe/Publish/BloomPub/PublishToBloomPubApi.cs:90` (changing the BloomPUB publish method)
- `src/BloomExe/Collection/CollectionSettingsDialog.cs:420` (OK in Collection Settings)
- `src/BloomExe/ExperimentalFeatures.cs:60` (any experimental-feature toggle)
- `src/BloomExe/Edit/EditingView.cs:174,181` (book renamed after a title change — which can itself be
  the emoji-named path)

Remembering the old value and restoring it when `Save()` fails keeps the settings object saveable.

</details>

src/BloomExe/Workspace/WorkspaceView.Designer.cs
```diff
@@ -19,6 +19,14 @@ protected override void Dispose(bool disposing)
 		{
 			if (disposing)
 			{
+				// _bookSelection outlives us (it is application-level, and we are recreated when
+				// the user switches collections), so stop it calling us. It matters for this
+				// subscriber in particular: deciding whether a selection is worth remembering
+				// depends on _collectionSettings, which for a disposed view describes the
+				// collection we have just left. See BL-16660.
+				if (_bookSelection != null)
+					_bookSelection.SelectionChanged -= PersistSelectedBookPath;
```
JohnThomson, line 28: [Devin] **Investigate**: Only the low-priority selection handler is unsubscribed on dispose; the high-priority one still fires after dispose

The new dispose code detaches `PersistSelectedBookPath` from `SelectionChanged`, and the comment
explains why that matters — a disposed view's collection settings describe the collection the user has
just left. But `HandleBookSelectionChanged`, subscribed to `SelectionChangedHighPriority` two lines
above at `src/BloomExe/Workspace/WorkspaceView.cs:212`, is still left attached. It calls
`GetCurrentSelectedBookInfo()`, which leans on the same `_collectionSettings` and
`_model.GetSourceCollectionFolders()`, and then writes to `_webSocketServer`.

That part is pre-existing, but the reasoning added here applies to it just as much, so it reads as a
half-applied fix worth completing.

src/BloomExe/Workspace/WorkspaceView.cs
JohnThomson: [Claude Opus 5 from John Thomson's machine during preflight] Good catch, and a real one — fixed in `05c2c7d5b5c55da4a3d2778f05338a0ac3c9b8c4`.

`SaveCurrentBookPath` now remembers the previous value and puts it back if `Save()` throws. That value had saved successfully before, so the in-memory settings object stays saveable and the knock-on failures in Collection Settings / publish / rename can't happen.

Worth noting the swallow itself was deliberate and is staying: `ArgumentException` is caught on purpose, because failing to remember which book was selected must not interrupt what the user was actually doing, and the recovery path in `SelectBookAtStartup` needs to survive exactly the corrupted-book case. The bug was only the missing rollback, which is what changed.

src/BloomExe/Workspace/WorkspaceView.Designer.cs
JohnThomson: [Claude Opus 5 from John Thomson's machine during preflight] Agreed — completed rather than half-applied, in `05c2c7d5b5c55da4a3d2778f05338a0ac3c9b8c4`. `Dispose` now detaches `HandleBookSelectionChanged` from `SelectionChangedHighPriority` as well, and the comment covers both handlers.

The local review raised the same asymmetry independently. In practice the stale handler was harmless today (a disposed `BloomWebSocketServer` has nulled its `_server`, so the message it sends goes nowhere), and unsubscribing on dispose matches what `CollectionTabView` already does — so this is tidying up a latent inconsistency, not fixing a live bug.

src/BloomExe/Program.cs
```diff
@@ -2048,6 +2049,9 @@ public static void SetUpLocalization()
                 );
 
                 Settings.Default.UserInterfaceLanguage = LocalizationManager.UILanguageId;
+                // Saves this and the UserInterfaceLanguageSetExplicitly set above; GetDesiredUiLanguage
+                // reads both on the next launch.
+                Settings.Default.Save();
```
JohnThomson: [Devin] **Investigate**: New Save() in SetUpLocalization runs before the settings-version upgrade

`SetUpLocalization()` is called from `Main` at `src/BloomExe/Program.cs:177`, well before the
`if (Settings.Default.NeedUpgrade) { Upgrade(); Reload(); ... }` block at `src/BloomExe/Program.cs:340-350`.
The new `Settings.Default.Save()` therefore creates/writes the *new* version's `user.config` before the
previous version's settings have been migrated.

I believe this is harmless because `LocalFileSettingsProvider.Upgrade()` writes the previous version's
values into the current file (which is why the existing `Upgrade(); Reload();` idiom works) and would
overwrite what we just wrote, and because `NeedUpgrade` itself is saved as still-true so the upgrade
block is still entered. Worth a sanity check on a real first-launch-after-upgrade, since a UI-language
value written pre-migration is exactly the kind of thing that would be silently reverted (or, if the
provider behaved differently, would suppress the migration).

src/BloomExe/Workspace/WorkspaceView.cs
```diff
@@ -362,23 +368,95 @@ private static bool TryGetBookIdFromFolder(string folderPath, out string bookId)
...
+                // collections (BL-14313).
+                if (IsSelectedBookObsoleteOrInvalid(book.FolderPath))
+                    return;
+                SaveCurrentBookPath(book.FolderPath);
+            }
+            catch (Exception error)
+                when (error is IOException || error is UnauthorizedAccessException)
+            {
+                Logger.WriteError("Unable to work out which book to remember.", error);
+            }
```
JohnThomson, line 423: [Devin] **Investigate**: Exception filter in the new selection handler is narrower than what the disk-reading predicate can throw

`PersistSelectedBookPath` catches only `IOException`/`UnauthorizedAccessException`, but the work it
guards is `IsSelectedBookObsoleteOrInvalid`, which calls `Path.GetDirectoryName` (can throw
`ArgumentException` for a malformed path) and enumerates source collection folders via
`SourceCollectionsList.GetCollectionFolders`, which resolves `.lnk` shortcuts through `Shortcut.Resolve`
— a COM call that can throw types outside the filter.

Anything escaping here now escapes out of `BookSelection.SelectBook()` to whichever caller selected the
book, which is a new failure mode for every book selection (previously only the settings write, whose
`ArgumentException` case is handled in `SaveCurrentBookPath`). Given the handler's own comment says
failing to remember the book should never interrupt the user's task, the filter arguably wants to be
broader.

JohnThomson: [Claude Opus 5 from John Thomson's machine during preflight] Agreed on the filter — widened in `0a8d5e7bfdf47e5f403cf5c559ab5a8e7bd0a109` to include `ArgumentException` (from `Path.GetDirectoryName`) and `COMException` (from resolving the `.lnk` shortcuts in the source collection list), which is what the guarded work can actually throw.

One correction to the reasoning, though: this is **not** a new failure mode for book selection. `HandleBookSelectionChanged` is subscribed to `SelectionChangedHighPriority`, and `InvokeSelectionChanged` fires the high-priority list first, so it already calls the same `IsSelectedBookObsoleteOrInvalid` predicate — unguarded — before `PersistSelectedBookPath` ever runs. A malformed path or a bad shortcut would therefore have thrown out of `SelectBook()` on master too, via `GetCurrentSelectedBookInfo()`. So the widened filter is belt-and-braces for consistency with this handler's own comment, not the thing standing between the user and a crash — the code comment now says exactly that, so nobody later mistakes it for load-bearing.

src/BloomExe/Program.cs
JohnThomson: [Claude Opus 5 from John Thomson's machine during preflight] This one is a good catch and I've taken the safe route in `0a8d5e7bfdf47e5f403cf5c559ab5a8e7bd0a109`, but I'm **leaving this thread open** for John, because the part you ask for — a sanity check on a real first-launch-after-upgrade — is not something I can do here.

What I changed: the save is now skipped while an upgrade is pending (`if (!Settings.Default.NeedUpgrade)`). On an ordinary launch it behaves as before. Reasoning: `Reload()` would discard our in-memory values anyway, so saving before the migration gained nothing even in the good case, which makes skipping it strictly better than reasoning about whether it is safe.

One thing worth flagging about your analysis: Bloom's settings do **not** use `LocalFileSettingsProvider`. `Settings.settings` specifies `Provider="SIL.Settings.CrossPlatformSettingsProvider"` for these properties, so the argument that `Upgrade()` would overwrite what we wrote rests on a provider we aren't using. I could not verify that provider's `Upgrade()` semantics from this checkout (source isn't vendored here), which is the other reason I preferred avoiding the situation over relying on it.

Residual question for John: the guard means that on a fresh install or first launch after an upgrade, the UI language derived at startup isn't persisted that session. That matches the old behaviour (the pre-existing in-memory assignment was equally discarded by `Reload()`), so nothing regresses — but it is a deliberate exception to the "save changed settings immediately" rule this branch is otherwise applying.

src/BloomExe/ApplicationContainer.cs
```diff
@@ -42,6 +42,7 @@ public ApplicationContainer()
             if (Settings.Default.MruProjects == null)
             {
                 Settings.Default.MruProjects = new MostRecentPathsList();
+                Settings.Default.Save();
```
JohnThomson: [Devin] **Bug**: Command-line Bloom runs can now rewrite the user's settings file at startup

The user's settings file is written (`Settings.Default.Save()` at `src/BloomExe/ApplicationContainer.cs:45`)
from code that also runs for command-line commands such as `upload`, `hydrate` and `createArtifacts`, so a
background/automated run can overwrite the settings a running Bloom would otherwise own.

Impact: An automated command-line run can quietly change the settings file that the interactive program
shares — the very thing this change set out to stop.

<details><summary>Mechanism: ApplicationContainer is constructed by CLI verbs too</summary>

`ApplicationContainer`'s constructor is used by every CLI command handler (`src/BloomExe/CLI/UploadCommand.cs:57`,
`src/BloomExe/CLI/CreateArtifactsCommand.cs:101`, `src/BloomExe/CLI/ChangeLayoutCommand.cs:35`,
`src/BloomExe/CLI/DownloadBookCommand.cs:24`, `src/BloomExe/CLI/SendFontAnalyticsCommand.cs:34`), not just by
the GUI path. The constructor already recognises the difference elsewhere (`if (!Program.RunningInConsoleMode)`
at `src/BloomExe/ApplicationContainer.cs:102`). The new `Save()` fires whenever `Settings.Default.MruProjects`
is null (fresh profile, harvester machine, service account), writing the whole settings object from a console
process — the same concern the PR states explicitly in the comment added at `src/BloomExe/Program.cs:2052-2061`.

Simply not saving here is safe: the in-memory list is registered with the container regardless, and the GUI
path saves the MRU list when it adds a path (`src/BloomExe/Program.cs:572-573`).

</details>

src/BloomExe/Workspace/WorkspaceView.cs
```diff
@@ -362,23 +369,104 @@ private static bool TryGetBookIdFromFolder(string folderPath, out string bookId)
...
             try
             {
-                Settings.Default.CurrentBookPath = null;
+                Settings.Default.CurrentBookPath = path;
                 Settings.Default.Save();
             }
-            catch (Exception e) when (e is IOException || e is UnauthorizedAccessException)
+            catch (Exception e)
+                when (e is IOException || e is UnauthorizedAccessException || e is ArgumentException
+                )
```
JohnThomson, line 447: [Devin] **Investigate**: Save failures other than IO/UnauthorizedAccess/Argument still escape into SelectBook's callers

`SaveCurrentBookPath` runs inside the `SelectionChanged` chain, so anything it throws surfaces to whoever called
`SelectBook`. The filter covers `IOException`, `UnauthorizedAccessException` and `ArgumentException`, but
`Settings.Default.Save()`'s other classic failure is `ConfigurationErrorsException` (corrupt or unavailable
`user.config`), which is neither of those and would propagate out of an ordinary book click.

This is not a regression (the old `SelectBook` called `Save()` unguarded too), and per AGENTS.md over-defensive
handling is discouraged, but since this code deliberately enumerates save failures it is worth deciding whether
that one belongs in the list.

src/BloomExe/ApplicationContainer.cs
JohnThomson: [Claude Opus 5 from John Thomson's machine during preflight] Confirmed and fixed in `868d350a2c640ec60d23f00f0acbd83bb9e021de` — this is the sibling of the `SetUpLocalization` save fixed one commit earlier, and I'd missed it. I verified the mechanism: all five CLI verbs do construct an `ApplicationContainer`, and the constructor's existing `if (!Program.RunningInConsoleMode)` a few lines down is a fair hint that the distinction matters there.

The `Save()` is simply gone, for the reason you give: the list is registered with the container either way, and the GUI saves it when it adds a path. The comment now records why this one deliberately does *not* follow the branch's "save changed settings immediately" rule.

src/BloomExe/Workspace/WorkspaceView.cs
JohnThomson: [Claude Opus 5 from John Thomson's machine during preflight] Agreed, and added in `868d350a2c640ec60d23f00f0acbd83bb9e021de`. You framed it exactly right: the general rule in AGENTS.md is not to be over-defensive, but this filter's whole job is to enumerate the ways saving can fail so that remembering the book never interrupts what the user was doing — and `ConfigurationErrorsException` was the one classic case missing from that list. An enumeration that omits the most likely member is worse than either a narrow or a broad catch.

Also correct that it isn't a regression: the old `SelectBook` called `Save()` unguarded, so a corrupt `user.config` would have thrown out of a book click on master too. This makes that strictly better rather than restoring parity.

src/BloomExe/Program.cs
```diff
@@ -2048,6 +2049,16 @@ public static void SetUpLocalization()
...
 
                 Settings.Default.UserInterfaceLanguage = LocalizationManager.UILanguageId;
+                // Deliberately NOT saved here, unlike the other settings this branch made save
+                // promptly. This runs on every startup path, and very early: before Main sets
+                // RunningInConsoleMode and dispatches the command-line verbs, so saving would make
+                // `bloom upload`, `hydrate` and friends rewrite the user's settings file - a smaller
+                // version of the very problem BL-16660 is about - and would risk failing an
+                // unattended run at startup. It is also before Main migrates the previous version's
+                // settings, whose Reload() would discard whatever we wrote anyway.
+                // Nothing is lost: this is a value derived at startup, not a user's choice, and
```
JohnThomson, line 2059: [Devin] **Investigate**: Console-mode runs now write the settings file at startup

The PR's goal is that non-GUI runs stop writing persisted user settings, but this new save runs on
*every* startup path, including command-line verbs: `SetUpLocalization()` is called at
`src/BloomExe/Program.cs:177`, well before the verb dispatch that returns at `src/BloomExe/Program.cs:299`.
So `bloom upload`, `hydrate`, `createArtifacts` etc. now rewrite the whole user settings file, whereas
previously a CLI verb only touched it via the BulkUploader's `SelectBook` (the very thing this PR removes)
or the GUI-only shutdown save.

Two follow-ups worth considering: (a) it re-introduces, in a smaller way, the "CLI clobbers the user's
settings file" concern BL-16660 is about; (b) it is an unguarded `Save()` in an unattended pipeline — if
the config file cannot be written the verb now throws at startup where before it did not. Guarding with
`!Program.RunningInConsoleMode` (as `ApplicationContainer` does for the exit handler) would keep the
intended benefit for the GUI without the CLI side effects.

src/BloomExe/Program.cs
```diff
@@ -2048,6 +2049,16 @@ public static void SetUpLocalization()
                 );
 
                 Settings.Default.UserInterfaceLanguage = LocalizationManager.UILanguageId;
+                // Deliberately NOT saved here, unlike the other settings this branch made save
```
JohnThomson, line 2052: [Devin] **Investigate**: PR description says SetUpLocalization now saves, but the code deliberately does not

The PR description lists `Program.SetUpLocalization` among the places given an immediate `Settings.Default.Save()`,
but the code at `src/BloomExe/Program.cs:2052` deliberately does not save, and carries a comment explaining why.
A reader going by the description would look for a save that isn't there.

src/BloomExe/Program.cs
JohnThomson: [Claude Opus 5 from John Thomson's machine during preflight] Correct, and this was the most valuable finding of the run — thank you. Fixed in `e8107a4f36f8ea111e12e1633e08c64ac0895d00`, and its sibling in `ApplicationContainer` in `868d350a2c640ec60d23f00f0acbd83bb9e021de`.

One correction: the suggested guard wouldn't have worked. `Program.RunningInConsoleMode` is not set until `src/BloomExe/Program.cs:224`, which is *after* `SetUpLocalization()` runs at line 177, so `!Program.RunningInConsoleMode` would have read false on every CLI run. There was no good later home for the save either — the settings migration a little further on calls `Reload()`, which discards these values by design.

So the save is simply gone. Nothing is lost: these are values derived at startup rather than the user's choice, `GetDesiredUiLanguage` derives them again next launch, and `ApplyUiLanguageChange` already saves both immediately when the user actually picks a language.

src/BloomExe/Program.cs
JohnThomson: [Claude Opus 5 from John Thomson's machine during preflight] Right — the description had gone stale when the save was removed. Fixed: the PR description now has a section called "Two places where 'save immediately' is the wrong rule" covering both `SetUpLocalization` and `ApplicationContainer`, with the reasoning, instead of listing them among the places that gained a save.

src/BloomExe/Program.cs
JohnThomson: [Claude Opus 5 from John Thomson's machine during preflight] Update: this is now moot, so I'm closing it rather than leaving it for John.

A later Devin pass found a bigger problem with the same save — it runs before the command-line verb dispatch, so `bloom upload`/`hydrate` were rewriting the user's settings file. The save has therefore been removed outright in `e8107a4f36f8ea111e12e1633e08c64ac0895d00`, which means there is no longer a write before the settings migration and no first-launch-after-upgrade check is needed.

Leaving the note about the provider here for the record, since it would matter to anyone adding a save early in startup later: Bloom's settings specify `Provider="SIL.Settings.CrossPlatformSettingsProvider"`, not `LocalFileSettingsProvider`, so the usual reasoning about `Upgrade()` overwriting an early write is not something to rely on here.

JohnThomson:
[Claude Opus 5 from John Thomson's machine during preflight] Consulted Devin on 2026-08-13/14 up to commit `868d350a2c640ec60d23f00f0acbd83bb9e021de`.

Four passes, because each round of fixes turned up something new. Devin now reports **both bugs resolved**, and every finding it raised has its own thread above with the outcome recorded.

What it caught that mattered:

- **The rollback bug** — swallowing the failed save left an unsaveable book path in the in-memory settings, which would have broken *every* later settings save in the session (Collection Settings, publish, rename). Devin confirms this fixed.
- **Two places where saving promptly was actively wrong** — `SetUpLocalization` and `ApplicationContainer` both run for command-line verbs, so `bloom upload`/`hydrate` had started rewriting the user's settings file. That is a miniature version of the bug this PR exists to fix, so both saves were removed. Devin confirms this fixed too.
- Smaller ones: a missing `ConfigurationErrorsException` in the save-failure list, a too-narrow guard around the selection handler, a half-applied dispose fix, and a stale claim in the PR description.

Two of its suggestions were not right and are corrected on their threads: guarding with `!Program.RunningInConsoleMode` cannot work at that point in startup (the flag is set later), and the widened exception filter is belt-and-braces rather than a crash fix, because the high-priority handler already calls the same predicate unguarded.

CI (`pr-automation`) is green. CodeRabbit is switched off for this repo in `.coderabbit.yml`, so it is not part of the panel. Full C# suite green at this commit: 3093 passed, 0 failed.

andrew-polk (approved):
> @andrew-polk reviewed 12 files and all commit messages, and resolved 8 discussions.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8203)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on JohnThomson).

andrew-polk (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8203)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on JohnThomson).

PR #8204, opened 2026-08-13
Fix collection whose name ends with a period breaking Team Collections (BL-16679)

StephenMcConnel:
Fixes the two problems reported in BL-16679, which both trace to one trigger: the reported collection's name ends with a **period**. Windows silently drops trailing periods and spaces when it creates a folder, so the folder lost the period while the `.bloomCollection` file kept it (`…نهاية\…نهاية..bloomCollection`).

### 1. The path broke FileSystemWatcher, so the collection wouldn't open

The path Bloom remembered still had the period on the folder component. Ordinary `File`/`Directory` APIs hide that — they normalize the path first — but `FileSystemWatcher` does not: .NET prefixes a path ending in a period or space with `\\?\`, which turns Windows normalization **off**, so `StartRaisingEvents` threw `Error reading the … directory`. That killed `TeamCollection.StartMonitoring`; `BookCollection` watches the collection folder the same way.

- `MiscUtils.GetPathAsOnDisk` normalizes a collection path, applied where a collection enters Bloom (`Program.OpenProjectWindow`) and before it goes to the MRU list, so a bad path already saved by an older Bloom heals.
- `CollectionSettings.GetPathForNewSettings` no longer creates such a path in the first place.

(An existing guard in `TeamCollectionManager` aimed at this could never fire — `Directory.Exists` returns true for the dotted path whenever the real folder exists — and had already been removed.)

### 2. That failure left Bloom as a zombie process

The exception escaped the `ProjectContext` constructor, which had already locked the `.bloomCollection` file. `Program` assigns `_projectContext` only *after* the constructor returns, so nobody could dispose the half-built context: it kept the collection file locked, along with the project-scoped objects it had built — the Shell and its WebView2 browsers, the file watchers. (`BloomServer` itself is application-level, resolved from the parent container, so it is not the context's to shut down — an earlier version of this description said otherwise.) Bloom could then neither reopen the collection (`being used by another process`) nor exit — `ProgramExit.Exit()`, and its 20-second force-quit net, is never reached, so it sat in a windowless `Application.Run()`.

The constructor now disposes itself on failure, and `Dispose` tolerates a partly-built context.

### 3. Three places assumed the settings file is named after its folder

False for this collection — and for any collection whose folder was renamed, of which this repo's own dev machines have several.

- `ProjectContext.GetCollectionSettings` already *intended* to fall back to whatever `.bloomCollection` is in the folder, but passed the **file** path where `TryGetSettingsFilePath` wants the **folder**, so `Directory.EnumerateFiles` threw `DirectoryNotFoundException` instead of repairing anything. This is what aborted the Team Collection startup sync (via `OkToEditCollectionSettings`), so no books were copied in.
- `SpreadsheetExportCommand` now looks for the settings file that is really in the folder instead of assuming the name. It keeps its old tolerance for a book that isn't inside a collection at all (falling back to default settings) rather than failing.
- The collection chooser's default-directory scan silently omitted such collections.

### Testing

20 new test cases across `MiscUtilsTests`, `CollectionSettingsTests`, `ProjectContextTests` (new file) and `TeamCollectionManagerTests` (previously an empty class). The `OkToEditCollectionSettings` test was written before the fix and confirmed to fail with the exact reported exception. Full C# suite green (3110 passed).

Verified by hand on the real collection: joining the Team Collection now works.

Ref: https://issues.bloomlibrary.org/youtrack/issue/BL-16679

[Devin review](https://devinreview.com/BloomBooks/BloomDesktop/pull/8204)

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8204)


src/BloomExe/Program.cs
StephenMcConnel, line 1: [Devin] **Bug**: A collection can appear twice in the Open/Create Collections list after the remembered path is corrected

(`src/BloomExe/Program.cs:1880-1883` as of `c1a7cde4a`. Posted at file level because the lines have moved since Devin reviewed that commit.)

The corrected path is added to the recently-used list (`AddNewPath(MiscUtils.GetPathAsOnDisk(path))` at `src/BloomExe/Program.cs:1883`) without removing the old, uncorrected one, so the same collection can be listed twice in the chooser forever.
Impact: A user whose collection name ends with a period or space sees that collection listed twice in the Open/Create Collections dialog, indefinitely.

<details><summary>Why the stale MRU entry survives the heal</summary>

`MostRecentPathsList.AddNewPath` de-duplicates only by exact string match (`_paths.Remove(path)` at `src/BloomExe/CollectionChoosing/MostRecentPathsList.cs:103`), and it prunes entries only when they no longer exist (`GetNonStalePaths`, `src/BloomExe/CollectionChoosing/MostRecentPathsList.cs:67-81`). Because ordinary `File`/`Directory` APIs normalize away trailing periods, `RobustFile.Exists` returns true for the old dotted path too, so that entry is never considered stale.

Result: after the first open, the list holds both `\Name.\Name..bloomCollection` (old) and `\Name\Name..bloomCollection` (normalized). `HandleGetMostRecentlyUsedCollections` (`src/BloomExe/web/controllers/CollectionChooserApi.cs:128-131`) builds one entry per MRU path, and its later de-dup against directory-scan results is also plain string `Contains`, so both entries are shown with the same title/folder; only the normalized one is flagged `isCurrentCollection`.

A fix would be to drop the un-normalized path from the MRU when normalization changes it (e.g. `RemovePath(path)` before adding the normalized one).

</details>

src/BloomExe/ProjectContext.cs
StephenMcConnel, line 1: [Devin] **Investigate**: GetCollectionSettings can now silently create a fresh default collection where it used to throw

(`src/BloomExe/ProjectContext.cs:528-545` as of `c1a7cde4a`. Posted at file level because the lines have moved since Devin reviewed that commit.)

With the folder passed correctly, `TryGetSettingsFilePath` returns false when the folder exists but holds no `.bloomCollection` at all; the method then falls through to `new CollectionSettings(projectSettingsPath)`, which (since the file does not exist) creates the folder(s) and *saves a brand-new default settings file* (`src/BloomExe/Collection/CollectionSettings.cs:246-261`). Previously this case threw `DirectoryNotFoundException`. So a Team Collection sync that deleted the local settings file without leaving a replacement will now quietly produce an empty/default collection instead of failing loudly  which is contrary to the repo's "fail fast" guidance for missing dependencies. Worth a deliberate decision.

src/BloomExe/Program.cs
StephenMcConnel: [Claude Opus 5 from Steve McConnel's machine during preflight] Fixed in `dd5d7e66c`. Confirmed the mechanism first: `MostRecentPathsList.AddNewPath` de-duplicates with `_paths.Remove(path)`, which is an exact string match, so the old dotted spelling really would have sat alongside the corrected one and shown up as a second, identical-looking entry in the dialog. `OpenCollection` now calls `RemovePath` on the old spelling whenever normalization changed it.

src/BloomExe/ProjectContext.cs
StephenMcConnel: [Claude Opus 5 from Steve McConnel's machine during preflight] Good catch, and fixed in `dd5d7e66c`. The `CollectionSettings` constructor does treat a missing file as "make me a new collection here" (it creates the folders and `Save()`s a default), so once this method started looking in the right folder it would have written a stray `.bloomCollection` into any folder that had none — including, via `SpreadsheetExportCommand`, a folder that is not a collection at all. It now throws `FileNotFoundException` naming the folder, which is what this code did before (as `DirectoryNotFoundException`), only legibly. There is a test for it: `GetCollectionSettings_NoCollectionInFolder_ThrowsRatherThanCreatingOne`.

src/BloomExe/Spreadsheet/SpreadsheetExportCommand.cs
StephenMcConnel, line 1: [Devin] **Bug**: Exporting a book that isn't inside a Bloom collection now fails instead of exporting with default settings

(`src/BloomExe/Spreadsheet/SpreadsheetExportCommand.cs:63` as of `dd5d7e66c`; posted at file level because the lines have since moved.)

Spreadsheet export now asks for the collection's settings through a lookup that refuses to invent one (`ProjectContext.GetCollectionSettings` at `src/BloomExe/Spreadsheet/SpreadsheetExportCommand.cs:68`) instead of falling back to defaults, so exporting a book that does not sit inside a real collection now aborts with an error code.

Impact: Command-line spreadsheet export of a loose book folder stops working, returning a failure instead of a spreadsheet.

<details>
<summary>Mechanism: GetCollectionSettings now throws where the CollectionSettings constructor used to create defaults</summary>

The old code called `new CollectionSettings(collectionSettingsFile)`. That constructor (`src/BloomExe/Collection/CollectionSettings.cs:246-261`) treats a non-existent path as "create it": it creates the folder if needed, saves a default `.bloomCollection`, and the export proceeds with default settings.

The new `ProjectContext.GetCollectionSettings` explicitly throws `FileNotFoundException` when no `.bloomCollection` is found in the folder (`src/BloomExe/ProjectContext.cs:546-556`). `SpreadsheetExportCommand.Handle` catches all exceptions and returns exit code 1 (`src/BloomExe/Spreadsheet/SpreadsheetExportCommand.cs:51-55`), so the export simply fails.

Also note `CollectionSettings.TryGetSettingsFilePath` calls `Directory.EnumerateFiles` without checking that the folder exists, so if the derived collection folder does not exist at all the failure is a `DirectoryNotFoundException` rather than the intended `FileNotFoundException`.

If refusing to write a stray settings file is the desired behavior, the export path probably still needs a defaults-only fallback (e.g. construct an in-memory `CollectionSettings`) rather than failing.

</details>

src/BloomExe/web/controllers/CollectionChooserApi.cs
StephenMcConnel, line 1: [Devin] **Bug**: A collection remembered with a trailing period is still listed twice in the Open/Create dialog after a normal launch

(`src/BloomExe/Program.cs:1880` as of `dd5d7e66c`; posted at file level because the lines have since moved.)

The remembered location of a collection is only corrected when it is opened from the chooser dialog (`MiscUtils.GetPathAsOnDisk` at `src/BloomExe/Program.cs:1883`), so a launch that opens the collection directly leaves the old, misspelled location in the recent list and the dialog shows the same collection twice.

Impact: Users can see one collection listed twice in the Open/Create Collections dialog, with the currently open one not highlighted on the duplicate entry.

<details>
<summary>Mechanism: MRU write sites other than OpenCollection are not normalized</summary>

`OpenCollection` normalizes before writing to the MRU list and removes the old spelling, but two other paths put an un-normalized path into the MRU and never go through it:

- Startup with an existing MRU entry: `StartUpShellBasedOnMostRecentUsedIfPossible` calls `OpenProjectWindow(path)` directly (`src/BloomExe/Program.cs:1659`), which normalizes only the path handed to the project, leaving the MRU entry as-is. `MostRecentPathsList.GetNonStalePaths` keeps it because `RobustFile.Exists` succeeds on the dotted path (Windows normalizes it).
- Double-clicking a `.bloomCollection` file: `Settings.Default.MruProjects.AddNewPath(path)` stores the raw path (`src/BloomExe/Program.cs:684`).

Meanwhile `CollectionChooserApi.HandleGetMostRecentlyUsedCollections` now also discovers such collections by scanning the default folder, and de-duplicates against the MRU with an exact string comparison (`src/BloomExe/web/controllers/CollectionChooserApi.cs:148-155`). The MRU's dotted spelling never equals the scanned (real on-disk) spelling, so both appear. For the same reason `isCurrentCollection` (compared against the normalized `CurrentCollectionSettings.SettingsFilePath`) is false for the MRU copy.

Normalizing at every MRU write site (or when reading the list) would make the heal complete.

</details>

StephenMcConnel, line 1: [Devin] **Investigate**: Directory scan can now throw where it previously just skipped a folder

(`src/BloomExe/web/controllers/CollectionChooserApi.cs:145` as of `dd5d7e66c`; posted at file level because the lines have since moved.)

`CollectionSettings.TryGetSettingsFilePath` calls `Directory.EnumerateFiles(collectionFolder, ...)` without guarding for a missing or inaccessible folder (`src/BloomExe/Collection/CollectionSettings.cs:175-177`). The old code here only did `RobustFile.Exists(path)`, which quietly returned false for any problem folder. Now a single sub-folder of `My Documents\Bloom` that the user cannot enumerate (permissions, a reparse point, a folder deleted between `GetDirectories` and the scan) will throw `UnauthorizedAccessException`/`DirectoryNotFoundException` out of the API handler, so the chooser gets an error instead of a (partial) list. Consider wrapping the per-folder lookup, or making `TryGetSettingsFilePath` return false rather than throw when the folder cannot be read.

src/BloomTests/Utils/MiscUtilsTests.cs
StephenMcConnel, line 1: [Devin] **Investigate**: New tests assume Windows path normalization and will fail on Linux

(`src/BloomTests/Utils/MiscUtilsTests.cs:38` as of `dd5d7e66c`; posted at file level because the lines have since moved.)

`GetPathAsOnDisk_FolderNameEndsWithPeriod_ReturnsFolderWindowsActuallyCreated` sanity-checks that `Directory.CreateDirectory("Collection Name.")` produced a folder named `Collection Name`. On Linux the trailing period is legal and preserved, so the sanity assert (and the later comparison) fails there; the same applies to the `GetPathForNewSettings`/`TryGetSettingsFilePath` cases that rely on Windows dropping trailing dots. If the suite is ever run on Linux/CI containers these need `[Platform("Win")]` or equivalent.

src/BloomExe/Spreadsheet/SpreadsheetExportCommand.cs
StephenMcConnel: [Claude Opus 5 from Steve McConnel's machine during preflight] You're right, and fixed in `43ac9799a`. Exporting a book that isn't inside a collection used to fall back to default settings (`new CollectionSettings` on a path that doesn't exist quietly creates one), and routing this through `ProjectContext.GetCollectionSettings` turned that into a hard failure — which was a side effect of the previous round's fix, not something this PR set out to change. It now looks for the real settings file directly and falls back to the derived name, and so to the old behaviour, when the folder holds no collection at all.

src/BloomExe/web/controllers/CollectionChooserApi.cs
StephenMcConnel: [Claude Opus 5 from Steve McConnel's machine during preflight] Confirmed and fixed in `43ac9799a`. The previous round only healed the remembered path in `OpenCollection`, and a normal launch opens the most recent collection through `OpenProjectWindow` without touching that list — so the old spelling survived, and now that the folder scan finds the same collection under its real name, the dialog listed both. The chooser now normalizes the remembered paths where it reads them, which fixes the display whatever happens to be stored.

StephenMcConnel: [Claude Opus 5 from Steve McConnel's machine during preflight] Agreed, and fixed in `43ac9799a`. This was a deliberate trade-off when the change was made (the repo's guidance is to fail fast rather than code defensively), but the failure mode is disproportionate: this is one folder among many in a list, and one unreadable or just-deleted folder would cost the user the entire Open/Create dialog — the way into the app. The lookup is now in a small helper that skips such a folder and logs it.

src/BloomTests/Utils/MiscUtilsTests.cs
StephenMcConnel: [Claude Opus 5 from Steve McConnel's machine during preflight] Not an issue here: both `BloomTests.csproj` and `BloomExe.csproj` target `net8.0-windows`, so the test project cannot be built or run on Linux at all — Bloom is Windows-only since it hosts WebView2. The reasoning about the behaviour is right, though, and `GetPathAsOnDisk`'s remarks now say it is a no-op on Linux, where such folder names are legal and FileSystemWatcher copes with them.

src/BloomExe/CollectionCreating/CollectionNameControl.cs
StephenMcConnel, line 1: [Devin] **Bug**: Collection names containing hidden control characters are accepted, then fail when the collection is created

(`src/BloomExe/CollectionCreating/CollectionNameControl.cs:89` as of `43ac9799a`; posted at file level because the lines have since moved.)

The name is checked for characters that folders can't contain only after whitespace has been stripped from it (`Text.Trim().TrimEnd('.', ' ')` at `src/BloomExe/CollectionCreating/CollectionNameControl.cs:84-92`), while the folder is still built from the name the user actually entered, so a name that ends with a pasted tab or line break is reported as fine and then fails when Bloom tries to make the folder.
Impact: A user who pastes a name with a stray tab/newline gets the Next button enabled and then an error instead of a new collection.

<details><summary>Why trimming before the invalid-character check loses coverage</summary>

`Path.GetInvalidFileNameChars()` includes the control characters `\t`, `\n`, `\v`, `\f`, `\r`. `string.Trim()` removes exactly those (they are whitespace), so `effectiveName` can no longer contain them, and the check at `src/BloomExe/CollectionCreating/CollectionNameControl.cs:89` can never see them. Before this change the check was performed on the raw `Text`, so such a name was rejected.

The path handed to the wizard is still derived from the raw text (`src/BloomExe/CollectionCreating/CollectionNameControl.cs:94-97`), and `CollectionSettings.GetPathForNewSettings` only strips trailing `'.'` and `' '` (`src/BloomExe/Collection/CollectionSettings.cs:1036`), so the tab survives into the folder name. `CollectionSettings.CreateNewCollection` then creates that directory, which Windows rejects.

Keeping the trimmed name for the emptiness/"templates" checks while also validating the raw text for invalid characters restores the old protection.

</details>

StephenMcConnel: [Claude Opus 5 from Steve McConnel's machine during preflight] Correct, and fixed in `1e2d1cb5e`. This was a regression from the previous round: moving the invalid-character check onto the trimmed name lost the control characters, because `Trim()` removes exactly the ones `Path.GetInvalidFileNameChars()` cares about, while the folder is still built from the raw text. The check is back on the raw text; the trimmed name is still what the emptiness and reserved-"templates" checks judge, which is what the earlier round was for.

src/BloomExe/Collection/CollectionSettings.cs
StephenMcConnel, line 1: [Devin] **Investigate**: Renaming a collection can still create the folder/settings-name mismatch

(`src/BloomExe/Collection/CollectionSettings.cs:1031-1036` as of `1e2d1cb5e`; posted at file level because the lines have since moved.)

The fix blocks the bad state at creation time (`GetPathForNewSettings`) and heals existing paths on open, but the rename path is untouched: `CollectionSettings.RenameCollection` (`src/BloomExe/Collection/CollectionSettings.cs:1043-1092`, invoked from `src/BloomExe/Program.cs:697` for `--rename`) moves the folder to a caller-supplied `toDirectory` and then derives the settings file name from that same untrimmed name. Renaming a collection to `Foo.` still yields folder `Foo` containing `Foo..bloomCollection`, and the path handed to `MruProjects.AddNewPath` at `src/BloomExe/Program.cs:702` still carries the trailing period. The new normalization in `OpenProjectWindow`/`OpenCollection` and the `TryGetSettingsFilePath` fallbacks now cover that state, so it should no longer be fatal  but the root cause is still reachable from rename, so it may be worth trimming there as well.

StephenMcConnel: [Claude Opus 5 from Steve McConnel's machine during preflight] Agreed, and closed in `6eac1460c`. Your read of the severity is right — the rest of the fix means that state is no longer fatal — but there was no reason to keep manufacturing it, and `Bloom --rename` was the one remaining way in. `RenameCollection` now trims the destination the same way collection creation does, so the folder it asks for is the folder Windows actually makes. Covered by `RenameCollection_NewNameEndsWithPeriod_FolderAndFileNamesAgree`.

src/BloomExe/ProjectContext.cs
StephenMcConnel, line 1: [Devin] **Investigate**: Collection lock is still created from the possibly-unrepaired settings path

(`src/BloomExe/ProjectContext.cs:529-535`.)

`_collectionLock = new CollectionLock(SettingsPath)` runs on the path handed to the constructor, before `GetCollectionSettings` performs the new folder-based repair (`src/BloomExe/ProjectContext.cs:525-559`). So in exactly the mismatch case this PR is about  a caller that derived the settings-file name from the folder name  the lock targets a file that does not exist: `CollectionLock.Lock` swallows the resulting `FileNotFoundException` in release (so the "don't rename my folder while I'm running" protection is silently lost), and rethrows it in a DEBUG build, which would abort opening the collection before the repair ever runs. This should be rare after the fix, because both `Program.OpenProjectWindow` and the collection chooser now hand out real on-disk settings paths, but the ordering means the repair does not cover the lock (and `ProjectContext.SettingsPath` itself remains the unrepaired string; only `BulkUploader.cs:370` reads it, comparing it to the path it passed in, so that comparison is unaffected).

StephenMcConnel: [Claude Opus 5 from Steve McConnel's machine during preflight] Left open deliberately: this is a real gap, but it arrived after this run had used its budget of automatic fix rounds, so it goes to the developer rather than being changed unreviewed. Summarising for whoever picks it up: the lock is taken on the path as handed in, before the folder-based repair runs, so in the mismatch case the lock silently protects nothing in a release build and aborts the open in a DEBUG one. I checked the one consumer of `ProjectContext.SettingsPath` (`BulkUploader.cs:370`) — it passes in a path it got from `TryGetSettingsFilePath`, so repairing the path before locking would make that comparison compare like with like rather than break it.

StephenMcConnel:
[Claude Opus 5 from Steve McConnel's machine during preflight] Consulted Devin on 2026-08-13, five rounds, up to commit `6eac1460c`.

It found real problems each time, and two of them were regressions from fixes for the earlier ones — worth knowing if you are deciding how much to trust a single pass. Nine findings were mirrored as review threads above, each with a reply saying what happened to it: seven fixed, one assessed as not applicable (the tests it worried about can't run on Linux — both projects target `net8.0-windows`), and one left **open** for a developer decision, because it arrived after this run had spent its budget of automatic fix rounds:

- Open: [the collection-file lock is taken before the settings-path repair](https://github.com/BloomBooks/BloomDesktop/pull/8204#discussion_r3779916527).

Devin also raised 13 informational items on the final round; those are not mirrored (low signal), but they were read. CI (`pr-automation`) is green on every commit, and CodeRabbit is switched off for this repo in `.coderabbit.yml`, so it is not a silent reviewer.

Note for anyone reading the raw Devin output: each re-review re-reported the earlier rounds' findings alongside the new ones, quoting code that had already changed. Every one was checked against the committed file before being acted on or dismissed.

src/BloomExe/ProjectContext.cs
StephenMcConnel: [Claude Opus 5 from Steve McConnel's machine during preflight] Developer decision: fix it. Done in `62aa76cac`.

The constructor now settles the real settings path before it does anything else, locks that, and builds the rest of the project from it. The lookup `GetCollectionSettings` was doing privately became a named helper the two share (`ProjectContext.GetRealSettingsPath`), and it deliberately returns the path it was given when there is no collection to find — so nothing changes for that case: `GetCollectionSettings` still reports it, and the lock still tolerates it.

Three tests cover the helper directly, including the case from this thread (a folder whose settings file has the doubled period, addressed by the folder-derived name) and the no-collection-at-all case.

StephenMcConnel:
[Claude Opus 5 from Steve McConnel's machine during preflight] Consultation log addendum, up to commit `62aa76cac`.

The one finding that had been left open — [the collection lock being taken before the settings-path repair](https://github.com/BloomBooks/BloomDesktop/pull/8204#discussion_r3779916527) — was decided by the developer in favour of fixing it, and is fixed in `62aa76cac`; that thread now records the decision and is resolved. All nine mirrored findings are therefore closed.

**Devin has no review for this commit.** Its job errored immediately, three separate triggers running: the push, a re-run of `pr-automation`, and loading the review page directly. No error message and no findings pass in any of them, and the failures are marked retryable on their side, so this looks like a Devin-side problem rather than anything about the diff. Its last completed review is `6eac1460c`; the only change since is the lock fix it asked for, plus a merge of master. Re-running preflight later would pick up a real result.

CI is green on `62aa76cac`, and the C# suite is green at 3128 tests after re-merging master.

src/BloomExe/ProjectContext.cs
StephenMcConnel, line 1: [Devin] **Bug**: Collections opened from the command line can still fail to open when their name ends with a period

(`src/BloomExe/ProjectContext.cs:539` as of `62aa76cac`; posted at file level because the lines have since moved.)

The folder spelling is only corrected in the desktop startup path, not where a collection is actually adopted (`GetRealSettingsPath` at `src/BloomExe/ProjectContext.cs:539-555`), so collections whose name ends with a period keep failing when Bloom is driven from the command line.
Impact: Command-line/automation runs (artifact creation, bulk upload, font analytics) on such a collection can still break the way the reported bug did, even though the same collection now opens fine in the app.

<details>
<summary>Why the normalization does not reach every entry point</summary>

`MiscUtils.GetPathAsOnDisk` (`src/BloomExe/Utils/MiscUtils.cs:431-434`) is applied only in `Program.OpenProjectWindow` (`src/BloomExe/Program.cs:1682-1684`) and before saving to the MRU list (`src/BloomExe/Program.cs:1884`). `ProjectContext`'s constructor instead calls `GetRealSettingsPath`, which does not normalize: for a path like `...\Foo.\Foo..bloomCollection`, `RobustFile.Exists` returns true (Windows normalizes for file APIs), so the un-normalized path is returned as-is and becomes `SettingsPath` and, via `Path.GetDirectoryName`, `editableCollectionDirectory`  which is exactly what the `FileSystemWatcher`s in `TeamCollection`/`BookCollection` choke on.

Other `ProjectContext` creators pass paths that never go through `GetPathAsOnDisk`:
- `src/BloomExe/CLI/CreateArtifactsCommand.cs:114`
- `src/BloomExe/CLI/SendFontAnalyticsCommand.cs:39`
- `src/BloomExe/WebLibraryIntegration/BulkUploader.cs:101` and `:377`

Normalizing inside `GetRealSettingsPath` (i.e. returning `MiscUtils.GetPathAsOnDisk(...)`) would cover every caller in one place, instead of relying on each entry point remembering to do it.

</details>

StephenMcConnel: [Claude Opus 5 from Steve McConnel's machine during preflight] This is the best catch of the review, and it's fixed in `5a11b5e05`. You are right that the earlier fix only covered Program's desktop startup path, and right about why nothing else noticed: `RobustFile.Exists` is perfectly happy with the un-normalized spelling because Windows normalizes for file APIs, so it survived all the way to the FileSystemWatchers — the original failure, still reachable from `CreateArtifactsCommand`, `SendFontAnalyticsCommand` and `BulkUploader`. Taken your suggested shape too: the normalization now happens inside `GetRealSettingsPath`, the one place every `ProjectContext` passes through, rather than asking each entry point to remember. New test: `GetRealSettingsPath_FolderSpelledWithTrailingPeriod_ReturnsTheFolderOnDisk`.

src/BloomExe/Collection/CollectionSettings.cs
StephenMcConnel, line 1: [Devin] **Investigate**: Renaming an existing collection to a name that only adds a trailing period now fails with a different error

(`src/BloomExe/Collection/CollectionSettings.cs:1045` as of `62aa76cac`; posted at file level because the lines have since moved.)

Because `toDirectory` is trimmed first, renaming e.g. "Foo" to "Foo." makes `toDirectory == fromDirectory`, so `Directory.Exists(toDirectory)` is true and the method throws "there is already a directory with the new name". `Program`'s `--rename` handler reports that and calls `Environment.Exit(-1)` (`src/BloomExe/Program.cs:698-717`), so Bloom quits instead of reopening. This is not a regression (the old code would have called `Directory.Move` with source and destination normalizing to the same path, which also throws and also exits), but since the user-visible name genuinely did change, the nicer behavior would be to treat a rename whose trimmed target equals the source folder as a no-op and return the existing settings path.

StephenMcConnel: [Claude Opus 5 from Steve McConnel's machine during preflight] Taken, and fixed in `5a11b5e05` — including your point that it isn't a regression, which is why the fix is the nicer behaviour rather than a restoration. A rename whose trimmed target is the folder we already have is now treated as the no-op it is, returning the existing settings path, so Bloom reopens instead of quitting with "there is already a directory with the new name". The comparison is exact rather than case-insensitive, so renaming a collection to change only the letter case is still a real rename. New test: `RenameCollection_NewNameOnlyAddsAPeriod_IsANoOpAndReturnsTheExistingSettings`.

StephenMcConnel:
[Claude Opus 5 from Steve McConnel's machine during preflight] Consulted Devin on 2026-08-17, up to commit `5a11b5e05`. **Re-review clean.**

The review of `62aa76cac` (which had errored on three triggers earlier — transient, it went through on a re-trigger) raised two genuinely new things, both now fixed in `5a11b5e05` and both mirrored above with their outcomes and resolved:

- [the normalization never reached the command-line entry points](https://github.com/BloomBooks/BloomDesktop/pull/8204#discussion_r3797481865) — worth reading, it is the sharper form of an informational item from round 1 that I had wrongly judged cosmetic;
- [renaming a collection to add only a trailing period reported "there is already a directory with the new name"](https://github.com/BloomBooks/BloomDesktop/pull/8204#discussion_r3797483647).

The review of `5a11b5e05` then found nothing new. It lists eleven findings, but every one carries an older commit in its own `head_sha` field (`62aa76cac`, `43ac9799a`, `dd5d7e66c`, `c1a7cde4a`, or none) — they are carry-forwards of findings already fixed, not observations about this commit. That per-finding `head_sha` is the field to check: the raw count on the job says nothing.

Eleven findings have been mirrored across the whole review, and all eleven threads are resolved with a documented outcome — nine fixed, one assessed as not applicable (the tests cannot run on Linux; both projects target `net8.0-windows`), and one that went to the developer for a decision and was then fixed at their direction.

CI green on `5a11b5e05`; C# suite 3130 passed, 0 failed.

StephenMcConnel:
[Claude Opus 5 from Steve McConnel's machine during preflight] Consulted Devin on 2026-08-17, up to commit `86eb72f6f`. **Re-review clean.**

Nothing new. The job lists eleven findings, but every one carries an older commit in its own `head_sha` field (or none at all) — they are carry-forwards of findings already fixed, not observations about this commit. Zero carry `86eb72f6f`.

This run existed only to close two mechanical gaps that stopped promotion: three test cases sitting uncommitted in the working tree (now `560ca4713` — they pass), and `origin/master` having moved on (now merged as `86eb72f6f`). No code behaviour changed.

CI green on `86eb72f6f`; C# suite 3133 passed, 0 failed; all eleven mirrored review threads remain resolved.

JohnThomson (changes requested):
> @JohnThomson reviewed 12 files and all commit messages, and made 3 comments.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8204)* status: all files reviewed, 14 unresolved discussions (waiting on StephenMcConnel).

___
*[`src/BloomExe/Spreadsheet/SpreadsheetExportCommand.cs` line 75 at r6](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8204#-P-Ln6qe8VMChrgjCsc8:-P-Ln6qi4cOdGqkto1pk:bvrilvd) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/86eb72f6f60a7b7352eb87f3dc1aaa883ff16599/src/BloomExe/Spreadsheet/SpreadsheetExportCommand.cs#L75)):*
> ```Smalltalk
>                 // No collection here at all -- the book may not be in one. Carry on with the derived
>                 // name, which gives us default settings, as this command has always done.
>                 collectionSettingsFile = CollectionSettings.GetSettingsFilePath(collectionFolder);
> ```

The usual relationship between a GetX function and a TryGetX function is that they return the exact same information if successful, but the latter returns false if it fails while the former throws. I'm thinking that the current GetSettingsFilePath wants to be renamed something like GetDefaultSettingsFilePath, and GetSettingFilePath (if we want such a thing) should use the Try version and throw if it returns false. A plain GetSettingsFilePath that we mostly need to remember not to use because we need the logic in the Try version feels like asking for bugs.
___
*[`src/BloomExe/Utils/MiscUtils.cs` line 431 at r6](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8204#-P-LpUEdAex-6kEPIo1C:-P-LpUEdAex-6kEPIo1D:b-a43vns) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/86eb72f6f60a7b7352eb87f3dc1aaa883ff16599/src/BloomExe/Utils/MiscUtils.cs#L431)):*
> ```Smalltalk
>         /// are therefore what is really on disk, and where FileSystemWatcher has no trouble with them.
>         /// </remarks>
>         public static string GetPathAsOnDisk(string path)
> ```

The name also suggested to me that it would fix case errors, though the remarks clarifies this. Since this is just a shorthand for GetFullPath with null-checking, maybe we could just use that name?
___
*[`src/BloomExe/web/controllers/CollectionChooserApi.cs` line 179 at r6](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8204#-P-LqS-qCegOdZP1cbw1:-P-LqS-qCegOdZP1cbw2:bi98bsz) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/86eb72f6f60a7b7352eb87f3dc1aaa883ff16599/src/BloomExe/web/controllers/CollectionChooserApi.cs#L179)):*
> ```Smalltalk
>         /// whole Open/Create dialog.
>         /// </remarks>
>         private static string FindCollectionFileIn(string folder)
> ```

What would you think of just folding the error-reporting logic into TryGetSettingsFilePath? Or if we need a version that doesn't catch these exceptions, give that a less nice name like TryGetSettingsFilePathUnprotected. I don't like having multiple methods that do almost the same thing and having to wonder which to use.

StephenMcConnel (commented):
> @StephenMcConnel reviewed 12 files and all commit messages, made 3 comments, and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8204)* status: all files reviewed, 13 unresolved discussions (waiting on JohnThomson).

___
*[`src/BloomExe/Spreadsheet/SpreadsheetExportCommand.cs` line 75 at r6](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8204#-P-Ln6qe8VMChrgjCsc8:-P-LujlhAEHVqXioC7CI:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/86eb72f6f60a7b7352eb87f3dc1aaa883ff16599/src/BloomExe/Spreadsheet/SpreadsheetExportCommand.cs#L75)):*

The usual relationship between a GetX function and a TryGetX function is that they return the exact same information if successful, but the latter returns false if it fails while the former throws. I'm thinking that the current GetSettingsFilePath wants to be renamed something like GetDefaultSettingsFilePath, and GetSettingFilePath (if we want such a thing) should use the Try version and throw if it returns false. A plain GetSettingsFilePath that we mostly need to remember not to use because we need the logic in the Try version feels like asking for bugs.
</blockquote></details>

Done.
___
*[`src/BloomExe/Utils/MiscUtils.cs` line 431 at r6](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8204#-P-LpUEdAex-6kEPIo1C:-P-LvwzHEROvnsVtm3eM:b-896fix) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/86eb72f6f60a7b7352eb87f3dc1aaa883ff16599/src/BloomExe/Utils/MiscUtils.cs#L431)):*

The name also suggested to me that it would fix case errors, though the remarks clarifies this. Since this is just a shorthand for GetFullPath with null-checking, maybe we could just use that name?
</blockquote></details>

Done.
___
*[`src/BloomExe/web/controllers/CollectionChooserApi.cs` line 179 at r6](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8204#-P-LqS-qCegOdZP1cbw1:-P-M2c201lgy0Su1FeSg:bc2a396) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/86eb72f6f60a7b7352eb87f3dc1aaa883ff16599/src/BloomExe/web/controllers/CollectionChooserApi.cs#L179)):*

What would you think of just folding the error-reporting logic into TryGetSettingsFilePath? Or if we need a version that doesn't catch these exceptions, give that a less nice name like TryGetSettingsFilePathUnprotected. I don't like having multiple methods that do almost the same thing and having to wonder which to use.
</blockquote></details>

I moved the exception catching / reporting try-catch logic to TryGetSettingsFilePath, and renamed this method to GetSettingsFilePath.  I also moved the method to CollectionSettings instead of being in CollectionChooserApi.

StephenMcConnel (commented):
> @StephenMcConnel+CLAU resolved 10 discussions.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8204)* status: 3 of 16 files reviewed, 3 unresolved discussions (waiting on JohnThomson).

StephenMcConnel (commented):
> @StephenMcConnel reviewed 13 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8204)* status: 3 of 16 files reviewed, 3 unresolved discussions (waiting on JohnThomson and StephenMcConnel+CLAU).

___
*[`src/BloomExe/Spreadsheet/SpreadsheetExportCommand.cs` line 75 at r6](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8204#-P-Ln6qe8VMChrgjCsc8:-P-MByTc2qfKF_hamBCG:bllkjxn) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/86eb72f6f60a7b7352eb87f3dc1aaa883ff16599/src/BloomExe/Spreadsheet/SpreadsheetExportCommand.cs#L75)):*

Done.
</blockquote></details>

Note that I applied that logic in renaming other methods in CollectionSettings.

StephenMcConnel:
[Claude Opus 5 from Steve McConnel's machine during preflight] Consulted Devin on 2026-08-18, up to commit `5668b56a3`. **Re-review clean** — eleven findings listed, none of them carrying this commit's sha in its own `head_sha`; all are carry-forwards of findings already fixed.

This run picked up the naming refactor John asked for (`110565ef5`), merged master, and repaired one thing that merge broke: master had added three call sites asking `GetSettingsFilePath` for the path to *write* a settings file to, which on this branch now answers with what is really in the folder — nothing, for an empty test folder — so eight Team Collection tests died on `ArgumentNullException`. Git reported no conflict; the two changes simply disagreed about what the name meant. Fixed in `5668b56a3`.

Also tidied the review queue: my mirrored Devin threads were resolved here on GitHub but still counted as unresolved in Reviewable, which was padding the reviewer's list to thirteen. Ten are now dispositioned there, leaving the three that are genuinely John's.

CI green on `5668b56a3`; C# suite 3148 passed, 0 failed (master's new tests included). The PR is back in draft, since this run added commits after it had been promoted.

JohnThomson (approved):
> @JohnThomson reviewed 13 files and all commit messages, and resolved 3 discussions.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8204)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on StephenMcConnel).

PR #8206, opened 2026-08-14
BL-16690 Honor a collection's minimum Bloom version, upgrading in place

andrew-polk:
Honors a `MinimumBloomVersion` element in a `.bloomCollection` file (set by hand today, in
preparation for Cloud syncing). If the running Bloom is older than that, it refuses to open the
collection and offers two ways forward: upgrade Bloom, or choose a different collection.

**Upgrading happens in place, through Velopack** — the same machinery behind the usual "a new
version is available" toast, driven directly so it works before any collection is open (there is no
toast host that early). We never send the user to the website: if there is nothing newer on their
channel, or the update fails, they are told so in a message box using the wording the toast would
have used, and are left to pick another collection.

An update that doesn't reach the required version is still installed — getting as far as the channel
allows is real progress, and the message says plainly that they will be asked again next launch.

**Team Collections** are locked out mid-session too: if the administrator sets a minimum version
while a teammate is working, the same modal dialog appears as soon as the change arrives, and the
only ways out are to upgrade or to switch collections. The repository's copy of the settings is read
straight out of the repo zip, since Bloom deliberately doesn't overwrite local collection settings
mid-session. A read failure never locks anyone out.

Version comparison is major.minor only, matching `BookStorage.CurrentBloomDesktopVersion` and the
book-download gate. An empty or unparseable value is treated as "no requirement" — we never lock
someone out of their own collection over a typo. Channel plays no part: a version number means the
same thing on every channel.

Caveat for the reviewer: none of the Velopack path or the Team Collection lock-out has been
exercised against a live update feed or a real Team Collection.

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8206)

[Devin review](https://devinreview.com/BloomBooks/BloomDesktop/pull/8206)

### DIFF
--- a/DistFiles/localization/en/Bloom.xlf
+++ b/DistFiles/localization/en/Bloom.xlf
@@ -314,6 +314,38 @@
         <source xml:lang="en">There was a problem while saving. Please return to the previous page and make sure it looks correct.</source>
         <note>ID: Browser.ProblemSaving</note>
       </trans-unit>
+      <trans-unit id="Collection.CollectionRequiresNewerVersion" translate="no">
+        <source xml:lang="en">The collection "{0}" requires Bloom {1} or greater. You are running Bloom {2}.</source>
+        <note>ID: Collection.CollectionRequiresNewerVersion</note>
+        <note>{0} is the name of the collection, {1} is the version it requires, {2} is the version of Bloom that is running.</note>
+      </trans-unit>
+      <trans-unit id="Collection.NewerVersionNeededHeader" translate="no">
+        <source xml:lang="en">This collection needs a newer version of Bloom.</source>
+        <note>ID: Collection.NewerVersionNeededHeader</note>
+      </trans-unit>
+      <trans-unit id="Collection.NothingNewerAvailable" translate="no">
+        <source xml:lang="en">This collection needs Bloom {0}, but there is no newer Bloom available to you yet. You are already on the newest one for the kind of Bloom you have installed.</source>
+        <note>ID: Collection.NothingNewerAvailable</note>
+        <note>{0} is the version the collection requires.</note>
+      </trans-unit>
+      <trans-unit id="Collection.OpenDifferentCollection" translate="no">
+        <source xml:lang="en">Open a Different Collection</source>
+        <note>ID: Collection.OpenDifferentCollection</note>
+      </trans-unit>
+      <trans-unit id="Collection.UpgradeBloom" translate="no">
+        <source xml:lang="en">Upgrade Bloom</source>
+        <note>ID: Collection.UpgradeBloom</note>
+      </trans-unit>
+      <trans-unit id="Collection.UpgradeDownloaded" translate="no">
+        <source xml:lang="en">Bloom {0} has been downloaded. Bloom will now close to install it; please start it again.</source>
+        <note>ID: Collection.UpgradeDownloaded</note>
+        <note>{0} is the version number of the Bloom that was downloaded.</note>
+      </trans-unit>
+      <trans-unit id="Collection.UpgradeIsAStepButNotEnough" translate="no">
+        <source xml:lang="en">Bloom {0} has been downloaded, but this collection needs {1}, so you will need to upgrade again later. Bloom will now close to install it; please start it again.</source>
+        <note>ID: Collection.UpgradeIsAStepButNotEnough</note>
+        <note>{0} is the version number that was downloaded, {1} is the version the collection requires.</note>
+      </trans-unit>
       <trans-unit id="CollectionSettingsDialog.AboutBloomSubscriptions">
         <source xml:lang="en">About Bloom Subscriptions</source>
         <note>ID: CollectionSettingsDialog.AboutBloomSubscriptions</note>
@@ -5808,6 +5840,11 @@ is mostly status, which shows on the Collection Tab -->
         <note>ID: TeamCollection.ConflictingEdit</note>
       </trans-unit>
 -->
+      <trans-unit id="TeamCollection.MinimumVersionSetRemotely" sil:dynamic="true" translate="no">
+        <source xml:lang="en">This collection now requires a newer version of Bloom than the one you are running.</source>
+        <note>ID: TeamCollection.MinimumVersionSetRemotely</note>
+        <note>Shown in the Team Collection message log when the administrator sets a minimum Bloom version that this Bloom does not meet.</note>
+      </trans-unit>
       <trans-unit id="TemplateBooks.BookName.Activity" sil:dynamic="true">
         <source xml:lang="en">Activity</source>
         <note>ID: TemplateBooks.BookName.Activity</note>
--- a/src/BloomExe/ApplicationUpdateSupport.cs
+++ b/src/BloomExe/ApplicationUpdateSupport.cs
@@ -41,6 +41,54 @@ namespace Bloom
             Verbose,
         }
 
+        /// <summary>
+        /// What came of an attempt to update Bloom without any toast UI. See
+        /// TryDownloadUpdateWithoutToasts.
+        /// </summary>
+        internal enum SilentUpdateOutcome
+        {
+            /// A newer Bloom was downloaded and will be installed when Bloom exits.
+            Downloaded,
+
+            /// We reached the update feed, and there is nothing newer on this channel.
+            NothingNewer,
+
+            /// We can't update this copy of Bloom at all: a developer build, one an administrator
+            /// manages, or one running under the debugger.
+            CannotUpdateThisBloom,
+
+            /// We tried and something went wrong -- most likely we couldn't reach the feed.
+            Failed,
+        }
+
+        /// <summary>
+        /// What happened, plus the words to say about it. Since this path has no toasts, the caller
+        /// shows Message in a message box; it is already the same wording the toast would have used,
+        /// so the user hears the same thing either way.
+        /// </summary>
+        internal class SilentUpdateResult
+        {
+            public readonly SilentUpdateOutcome Outcome;
+
+            /// Ready to show, or null when the outcome speaks for itself and the caller has
+            /// something better to say (Downloaded, NothingNewer).
+            public readonly string Message;
+
+            /// The version we downloaded; only set for Downloaded.
+            public readonly string DownloadedVersion;
+
+            public SilentUpdateResult(
+                SilentUpdateOutcome outcome,
+                string message,
+                string downloadedVersion = null
+            )
+            {
+                Outcome = outcome;
+                Message = message;
+                DownloadedVersion = downloadedVersion;
+            }
+        }
+
         enum UploadStatus
         {
             // First call this session, or previous call(s) completed and found no updates
@@ -228,6 +276,11 @@ namespace Bloom
                 // When we exit, apply the updates. (If autoupdate is false, this is still appropriate,
                 // because the user responded to the message about updates available by clicking "Update Now",
                 // so we're just completing something already approved).
+                // Check the flag as well as setting it, so the "only one exit handler" rule holds
+                // whichever path got here first.
+                if (_willInstallUpdateOnExit)
+                    return;
+                _willInstallUpdateOnExit = true;
                 Application.ApplicationExit += (sender, args) =>
                 {
                     // Write a file so that if the update fails (e.g., a running process prevents it),
@@ -252,6 +305,225 @@ namespace Bloom
                     e
                 );
             }
+#endif
+        }
+
+        /// <summary>
+        /// Download an update, if there is one, without using any toast UI.
+        ///
+        /// Everything else in this class talks to the user through toasts, and those are only
+        /// rendered by ToastHost, which lives in the main workspace. This method exists for callers
+        /// that run before a collection is open -- notably the "this collection needs a newer Bloom"
+        /// dialog -- where there is no workspace and so a toast would go nowhere at all.
+        ///
+        /// It behaves as if the user had clicked the "Update Now" toast: it finds the update and
+        /// downloads it, and the caller is expected to shut Bloom down afterwards, at which point
+        /// Velopack installs it (see ArrangeToInstallDownloadedUpdateOnExit).
+        ///
+        /// Do not call this on the UI thread without wrapping it in Task.Run: it awaits network
+        /// work, and blocking the UI thread on that would deadlock.
+        /// </summary>
+        internal static async Task<SilentUpdateResult> TryDownloadUpdateWithoutToasts()
+        {
+#if __MonoCS__
+            // Nothing in this class works off Windows.
+            return new SilentUpdateResult(
+                SilentUpdateOutcome.CannotUpdateThisBloom,
+                "Bloom can only update itself on Windows."
+            );
+#else
+            // The same three situations WorkspaceView.CheckForUpdatesImpl refuses to update in, with
+            // the same explanations, since the user is owed the same information either way.
+            if (Debugger.IsAttached)
+                return new SilentUpdateResult(
+                    SilentUpdateOutcome.CannotUpdateThisBloom,
+                    "Sorry, you cannot check for updates from the debugger."
+                );
+            if (InstallerSupport.SharedByAllUsers())
+                return new SilentUpdateResult(
+                    SilentUpdateOutcome.CannotUpdateThisBloom,
+                    LocalizationManager.GetString(
+                        "CollectionTab.AdminManagesUpdates",
+                        "Your system administrator manages Bloom updates for this computer."
+                    )
+                );
+            if (IsDev)
+                return new SilentUpdateResult(
+                    SilentUpdateOutcome.CannotUpdateThisBloom,
+                    "Checking for updates is disabled on developer builds. No relevant channel."
+                );
+
+            // Another part of Bloom may already be partway through an update effort of its own,
+            // using the very statics we are about to write. (The workspace checks for updates on a
+            // timer, so in the Team Collection lock-out case it has had a whole session to get
+            // going.) CheckForAVelopackUpdate has a switch just like this one, and for the same
+            // reason: whoever got here first owns _bloomUpdateManager and _newVersion, and
+            // overwriting them behind their back would strand the update they had already lined up.
+            switch (_status)
+            {
+                case UploadStatus.NothingKnown:
+                    break; // nothing else is going on, so do the whole job ourselves
+                case UploadStatus.FoundUpdates:
+                    // The timer already found an update and is waiting for the user to accept it.
+                    // They just have, in our dialog, so download the one it found.
+                    return await DownloadTheUpdateWeAlreadyKnowAbout();
+                case UploadStatus.DownloadedWaitingForRestart:
+                    // Already downloaded, and installing-on-exit is already arranged. Telling the
+                    // caller it is Downloaded is exactly right: quitting will install it.
+                    return new SilentUpdateResult(
+                        SilentUpdateOutcome.Downloaded,
+                        null,
+                        _newVersion?.TargetFullRelease?.Version?.ToString()
+                    );
+                case UploadStatus.LookingForUpdates:
+                case UploadStatus.Downloading:
+                    // Joining in would mean two downloads of the same thing.
+                    return new SilentUpdateResult(
+                        SilentUpdateOutcome.Failed,
+                        LocalizationManager.GetString(
+                            "CollectionTab.UpdateCheckInProgress",
+                            "Bloom is already working on checking for updates."
+                        )
+                    );
+                case UploadStatus.Failed:
+                    // The same words the normal path's error toast uses.
+                    return new SilentUpdateResult(
+                        SilentUpdateOutcome.Failed,
+                        "Restart Bloom to try checking for updates again"
+                    );
+            }
+
+            try
+            {
+                // Quiet matters: with Verbose, GetUpdateUrl reports failures by toast.
+                if (!GetUpdateUrl(BloomUpdateMessageVerbosity.Quiet, out var updateUrl))
+                {
+                    // Overwhelmingly the reason we can't work out where to look is that we can't
+                    // reach the server, which is what the normal path says here too.
+                    return new SilentUpdateResult(
+                        SilentUpdateOutcome.Failed,
+                        LocalizationManager.GetString(
+                            "CollectionTab.UnableToCheckForUpdate",
+                            "Could not connect to the server to check for an update. Are you connected to the internet?",
+                            "Shown when Bloom tries to check for an update but can't, for example because it can't connect to the internet, or a problems with our server, etc."
+                        )
+                    );
+                }
+
+                _status = UploadStatus.LookingForUpdates;
+                _bloomUpdateManager = new UpdateManager(
+                    updateUrl,
+                    new UpdateOptions { MaximumDeltasBeforeFallback = 2 }
+                );
+
+                try
+                {
+                    _newVersion = await _bloomUpdateManager.CheckForUpdatesAsync();
+                }
+                catch (Exception e)
+                {
+                    Logger.WriteError("Could not check for a Velopack update without toasts", e);
+                    _status = UploadStatus.Failed;
+                    // The same words the "check failed" toast uses.
+                    return new SilentUpdateResult(
+                        SilentUpdateOutcome.Failed,
+                        "Bloom was unable to check for updates. Restart to try again."
+                    );
+                }
+
+                if (_newVersion == null)
+                {
+                    // Nothing newer on this channel. That is a real outcome here rather than good
+                    // news: the caller only asked because this Bloom is too old for the collection.
+                    // Clearing these is safe only because the switch above established that nobody
+                    // else had an update in hand when we started.
+                    _bloomUpdateManager = null;
+                    _status = UploadStatus.NothingKnown;
+                    return new SilentUpdateResult(SilentUpdateOutcome.NothingNewer, null);
+                }
+
+                return await DownloadTheUpdateWeAlreadyKnowAbout();
+            }
+            catch (Exception e)
+            {
+                Logger.WriteError("Velopack update without toasts failed unexpectedly", e);
+                _status = UploadStatus.Failed;
+                return new SilentUpdateResult(
+                    SilentUpdateOutcome.Failed,
+                    "Bloom was unable to check for updates. Restart to try again."
+                );
+            }
+#endif
+        }
+
+#if !__MonoCS__
+        /// <summary>
+        /// Download the update we have already found (_newVersion), reporting the outcome the way
+        /// TryDownloadUpdateWithoutToasts does. Shared by the case where we found it ourselves and
+        /// the case where the workspace's timer found it and was waiting for the user to accept.
+        /// </summary>
+        private static async Task<SilentUpdateResult> DownloadTheUpdateWeAlreadyKnowAbout()
+        {
+            try
+            {
+                _status = UploadStatus.Downloading;
+                await _bloomUpdateManager.DownloadUpdatesAsync(_newVersion);
+            }
+            catch (Exception e)
+            {
+                Logger.WriteError("Could not download a Velopack update without toasts", e);
+                _status = UploadStatus.Failed;
+                // The same words the "download failed" toast uses.
+                return new SilentUpdateResult(
+                    SilentUpdateOutcome.Failed,
+                    "Bloom was unable to download and install updates. Restart to try again."
+                );
+            }
+
+            _status = UploadStatus.DownloadedWaitingForRestart;
+            return new SilentUpdateResult(
+                SilentUpdateOutcome.Downloaded,
+                null,
+                _newVersion.TargetFullRelease?.Version?.ToString()
+            );
+        }
+#endif
+
+        /// <summary>
+        /// True once something has arranged to install a downloaded update as Bloom exits. Both
+        /// the toast path and the silent path do that, and they must not both do it, or Bloom
+        /// would try to apply the same update twice on the way out.
+        /// </summary>
+        private static bool _willInstallUpdateOnExit;
+
+        /// <summary>
+        /// Arrange for an update downloaded by TryDownloadUpdateWithoutToasts to be installed when
+        /// Bloom exits. Call this on the UI thread, then shut Bloom down. Does nothing if the
+        /// normal update path has already arranged it.
+        /// </summary>
+        internal static void ArrangeToInstallDownloadedUpdateOnExit()
+        {
+#if !__MonoCS__
+            if (_willInstallUpdateOnExit)
+                return;
+            _willInstallUpdateOnExit = true;
+            Application.ApplicationExit += (sender, args) =>
+            {
+                // So that a failed install (e.g. a running process got in the way) can be spotted
+                // and reported on the next launch, exactly as the normal update path does.
+                WriteUpdateAttemptFile(_newVersion.TargetFullRelease.Version.ToString());
+                // Shutting down is not instant -- Shell.OnClosing cancels the first close while it
+                // saves -- so the restart toast can appear and be clicked in the meantime, and that
+                // has already asked for the install. Applying the same update twice is how an
+                // upgrade fails or restarts Bloom when nobody asked it to. Same guard, same reason,
+                // as the normal update path above.
+                if (_restartingAfterToastClicked)
+                    return;
+                // false = don't restart Bloom for us. The user was trying to open a collection this
+                // Bloom can't handle, so there is nothing useful to come back to until the new
+                // version is in place; they start Bloom again themselves.
+                _bloomUpdateManager.WaitExitThenApplyUpdates(null, true, false);
+            };
 #endif
         }
 
--- a/src/BloomExe/Collection/MinimumBloomVersionCheck.cs
+++ b/src/BloomExe/Collection/MinimumBloomVersionCheck.cs
@@ -1,0 +1,582 @@
+using System;
+using System.Text;
+using System.Threading.Tasks;
+using System.Windows.Forms;
+using System.Xml.Linq;
+using Bloom.MiscUI;
+using L10NSharp;
+using SIL.IO;
+using SIL.Reporting;
+
+namespace Bloom.Collection
+{
+    /// <summary>
+    /// A collection can declare, via the MinimumBloomVersion element of its .bloomCollection file,
+    /// the oldest version of Bloom that is allowed to open it. This is in preparation for Cloud
+    /// syncing: once a collection has been touched by a version of Bloom that knows how to sync it,
+    /// letting an older Bloom loose on it could do real damage.
+    ///
+    /// At this point there is no UI for setting the flag; it has to be added to the file by hand.
+    /// See BL-16690.
+    /// </summary>
+    public static class MinimumBloomVersionCheck
+    {
+        /// <summary>
+        /// Decide whether this Bloom is allowed to open the given collection.
+        /// </summary>
+        /// <param name="settingsFilePath">Path to the .bloomCollection file</param>
+        /// <param name="minimumVersion">What the collection asked for, as major.minor, when we say no</param>
+        /// <returns>true if the collection demands a newer Bloom than we are</returns>
+        public static bool IsThisBloomTooOld(string settingsFilePath, out string minimumVersion)
+        {
+            // When this is a Team Collection and we can actually read the repository, the repository
+            // decides -- including when it says there is no requirement at all. It has to be able to
+            // LIFT a requirement, not just impose one: a member who is being refused never opens the
+            // collection, so the startup sync that would refresh their own copy never runs, and an
+            // administrator's mistaken requirement would shut them out on every launch for ever,
+            // fixable only by hand-editing the file on each machine.
+            //
+            // Otherwise -- an ordinary collection, or a shared folder we cannot reach right now --
+            // we fall back to what we have: the file on this computer and anything the repository
+            // told us earlier this session, taking whichever demands more. Not being able to see the
+            // repository is not the same as the repository saying "no requirement", so in that case
+            // we keep the protection rather than quietly dropping it.
+            var declaredVersion =
+                MinimumVersionInTeamCollectionRepo(settingsFilePath)
+                ?? MoreDemandingOf(
+                    ReadMinimumBloomVersion(settingsFilePath),
+                    MinimumVersionLearnedFromRepo(settingsFilePath)
+                );
+            if (IsVersionSufficient(declaredVersion, RunningBloomVersion))
+            {
+                minimumVersion = "";
+                return false;
+            }
+            // We can only get here if it parsed, since anything we can't parse counts as satisfied.
+            minimumVersion = ToMajorMinor(Version.Parse(declaredVersion.Trim()));
+            return true;
+        }
+
+        /// <summary>
+        /// Compare a collection's declared minimum version against the version we are running.
+        /// We compare only major and minor, like the other version gates in Bloom
+        /// (BookStorage's feature requirements and BookDownload's minVersion), so a minimum of
+        /// "6.5" is satisfied by any 6.5.x or later, and a build number in the minimum is ignored.
+        ///
+        /// The channel (Alpha/Beta/Release) deliberately plays no part: 6.5 means the same thing on
+        /// every channel, so an alpha tester who is running ahead of the release is correctly let in.
+        /// </summary>
+        /// <remarks>Separated out from the file reading so it can be unit tested.</remarks>
+        internal static bool IsVersionSufficient(string minimumVersion, Version runningVersion)
+        {
+            if (string.IsNullOrWhiteSpace(minimumVersion))
+                return true; // the normal case: the collection doesn't care
+
+            // Version.TryParse needs at least "major.minor", so a bare "6" lands here too.
+            if (!Version.TryParse(minimumVersion.Trim(), out var requiredVersion))
+            {
+                // Someone hand-edited the file and mistyped. Locking them out of their own
+                // collection over a typo would be worse than ignoring it, so just complain to the log.
+                Logger.WriteEvent(
+                    $"Ignoring unparseable {CollectionSettings.kMinimumBloomVersionElementName} '{minimumVersion}' in collection settings."
+                );
+                return true;
+            }
+
+            if (runningVersion.Major != requiredVersion.Major)
+                return runningVersion.Major > requiredVersion.Major;
+            return runningVersion.Minor >= requiredVersion.Minor;
+        }
+
+        /// <summary>
+        /// What a Team Collection's repository says, read straight from the shared folder, or null
+        /// for an ordinary collection or a repository we cannot reach.
+        ///
+        /// This is what closes the first-launch gap. The local copy of a Team Collection's settings
+        /// is only refreshed from the repository later in startup, well after this gate, so a member
+        /// whose administrator set a minimum version while they were closed would otherwise be
+        /// judged on yesterday's file, let in, and only stopped the launch after that. Asking the
+        /// repository directly costs one small read, and only for collections that are actually in
+        /// a Team Collection. See BL-16690.
+        /// </summary>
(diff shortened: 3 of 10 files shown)


src/BloomExe/Collection/MinimumBloomVersionCheck.cs
andrew-polk, line 1: [Devin] **Bug**: New user-visible messages are missing from the English translation file

Eight new translatable messages are introduced (starting with the dialog heading built at `src/BloomExe/Collection/MinimumBloomVersionCheck.cs:156-178`) without being added to the English localization file, so translators never see them and the strings can never be localized.

Impact: The upgrade dialog and its follow-up messages will stay English-only for every user, and translators have no way to know they exist.

Missing ids: `Collection.NewerVersionNeededHeader`, `Collection.CollectionRequiresNewerVersion`, `Collection.UpgradeBloom`, `Collection.OpenDifferentCollection`, `Collection.UpgradeDownloaded`, `Collection.UpgradeIsAStepButNotEnough`, `Collection.NothingNewerAvailable`, `TeamCollection.MinimumVersionSetRemotely`.

src/BloomExe/ApplicationUpdateSupport.cs
andrew-polk, line 1: [Devin] **Bug**: Choosing to upgrade can cancel an already-downloaded update and crash Bloom as it closes

`TryDownloadUpdateWithoutToasts` overwrites `_status`, `_bloomUpdateManager` and `_newVersion` without the state check that guards `CheckForAVelopackUpdate` (`src/BloomExe/ApplicationUpdateSupport.cs:389-396`).

Sequence: with AutoUpdate on, the workspace's idle check downloads an update, so `_status == DownloadedWaitingForRestart` and an `Application.ApplicationExit` handler is registered that dereferences `_newVersion.TargetFullRelease` and `_bloomUpdateManager`. Later in the same session a Team Collection settings change locks the user out and they click "Upgrade Bloom". If the feed check now returns nothing, `_newVersion` and `_bloomUpdateManager` are nulled, so that earlier exit handler throws a NullReferenceException at `ApplicationExit` and the update it downloaded is never installed. If it returns something, a second download runs and two exit handlers both try to apply updates.

src/BloomExe/Collection/MinimumBloomVersionCheck.cs
andrew-polk, line 1: [Devin] **Investigate**: PR description describes a different feature (AllowCheckouts) than the code

The submitted description talks about an `AllowCheckouts` setting, `TeamCollectionApi`, `StatusPanelCommon` and LESS changes, and references BL-16691. None of that appears in this diff, which implements `MinimumBloomVersion` (BL-16690). The description is unusable as a statement of intent for this change, so reviewers should get a correct one before merging.

src/BloomExe/TeamCollection/TeamCollection.cs
andrew-polk, line 1: [Devin] **Investigate**: Collection-settings change handler now blocks the idle queue in a modal dialog

`HandleCollectionSettingsChange` previously only wrote a log message; it can now run a modal dialog and (if the user clicks Upgrade) a whole synchronous network download inside the `Application.Idle` handler. While that modal dialog runs, idle events continue to fire, so `HandleRemoteBookChangesOnIdle` itself re-enters and will dequeue and process other pending repo events (`HandleNewBook`, `HandleModifiedFile`, `HandleDeletedRepoFile`) against a collection we are in the middle of shutting down. `MinimumBloomVersionCheck._collectionBeingLockedOut` guards only against stacking up a second lock-out dialog, not against this other work.

src/BloomExe/Program.cs
andrew-polk, line 1: [Devin] **Investigate**: Startup lock-out reads the local settings file, which a Team Collection may not have synced yet

At startup the gate reads the local `.bloomCollection`, but for a Team Collection the local copy is only overwritten from the repo later, during `SyncLocalAndRepoCollectionFiles`. So a member whose repo has just gained a `MinimumBloomVersion` will pass the startup gate and only be locked out afterwards, via the repo-change handler (`CheckWhetherRepoNowRequiresANewerBloom`), i.e. after a ProjectContext and Shell have been built around a collection this Bloom is not supposed to touch. If the flag is ever meant to protect files rather than warn a person, the Team Collection sync should be consulted before the collection is opened.

src/BloomExe/Collection/MinimumBloomVersionCheck.cs
andrew-polk: [Claude Opus 5] Fixed in `65d069c707`. All seven `Collection.*` strings and `TeamCollection.MinimumVersionSetRemotely` are now in `DistFiles/localization/en/Bloom.xlf`, marked `translate="no"` as AGENTS.md requires for new entries.

One correction to the finding: not every id passed to `_tcLog.WriteMessage` is in the XLF today (`TeamCollection.SettingsModifiedRemotely` and `TeamCollection.NewBookArrived`, for instance, are not), so that one was following an existing mixed convention rather than breaking a uniform one. Added it anyway — the ones that are there, like `TeamCollection.ConflictingCheckout`, are the better example.

Andrew is reviewing the full set of new strings, including whether they should stay `translate="no"`.

src/BloomExe/ApplicationUpdateSupport.cs
andrew-polk: [Claude Opus 5] Fixed in `65d069c707` — good catch, this was a real crash-on-exit and it had been on our own list as "too architectural to fix in review". The concrete route out you described is the one we took.

`TryDownloadUpdateWithoutToasts` now switches on `_status` before touching anything, the way `CheckForAVelopackUpdate` does: an already-downloaded update is reported as our answer (no second download, nothing nulled), an update the timer found and was waiting on is downloaded, and a check already in flight is declined rather than joined. `ArrangeToInstallDownloadedUpdateOnExit` also refuses to register a second exit handler, so the same update can't be applied twice on the way out.

src/BloomExe/Collection/MinimumBloomVersionCheck.cs
andrew-polk: [Claude Opus 5] Fixed. The description had been published from the wrong file — it was another feature's text (BL-16691's `AllowCheckouts`) pasted in by mistake, exactly as you read it. The PR description now describes this change.

Worth noting the two features are genuinely related: this branch has since been merged onto BL-16691, so some `AllowCheckouts` code does now appear in the branch — but as its base, not as part of this change.

src/BloomExe/TeamCollection/TeamCollection.cs
andrew-polk: [Claude Opus 5] Fixed in `65d069c707`. `HandleRemoteBookChangesOnIdle` now guards against re-entering itself: it sets a flag around the dequeue-and-handle, so while a modal dialog is up and Idle keeps firing, nothing further comes off the queue. Whatever is queued waits for the next Idle after we return.

You were right that `_collectionBeingLockedOut` was the wrong tool for this — it only ever stopped a second lock-out dialog for the same collection.

DistFiles/localization/en/Bloom.xlf
andrew-polk, line 1: [Devin] **Bug**: New Team Collection warning text can never be translated because it was added inside a commented-out region of the English string file

The new Team Collection message text was added (`TeamCollection.MinimumVersionSetRemotely`) inside a large XML comment block that runs from `DistFiles/localization/en/Bloom.xlf:5676` to `:5847` ("Don't want these actually translated until things stabilize"), so the entry is invisible to the localization system and translators.

Impact: the "this collection now requires a newer version of Bloom" log message would stay English forever in every language.

Why it hides: at runtime `TeamCollectionMessage.TextForDisplay` calls `LocalizationManager.GetString(L10NId, RawEnglishMessageTemplate)`, so the English fallback in the code is shown and nothing visibly breaks.

src/BloomExe/Collection/MinimumBloomVersionCheck.cs
andrew-polk, line 1: [Devin] **Investigate**: Lock-out guard is never cleared, and the short-circuit falsely reports a shutdown

`_collectionBeingLockedOut` is set and never reset. Two consequences:

1. When the guard short-circuits, `LockUserOutOfOpenCollection` returns without doing anything, but `CheckWhetherRepoNowRequiresANewerBloom` still returns true, so `HandleRemoteBookChangesOnIdle` skips `BookSelection.InvokeSelectionChanged(false)` on the assumption that the collection is being torn down. That is right while the modal dialog is up, but it is also what happens for every later settings-change event for that same collection name.

2. Because the local `.bloomCollection` is deliberately not rewritten mid-session, a user who is locked out of collection A, moves to B, and then re-opens A from the toolbar in the same session will pass the startup gate in `Program.OpenProjectWindow` (it reads the stale local file), and any subsequent repo settings change will be silently ignored thanks to the guard.

andrew-polk, line 1: [Devin] **Investigate**: Blocking the UI thread on the download also blocks the embedded server and WebView content

`UpgradeBloom` deliberately blocks the UI thread (`Task.Run(...).GetAwaiter().GetResult()`), which is safe from deadlock because the awaited continuations run on the thread pool. Two side effects worth confirming with a live feed:

1. `Application.UseWaitCursor = true` normally needs a message pump turn to take effect, so the wait cursor may never actually appear during the blocked period.
2. In the mid-session lock-out case, any WebView2/BloomServer work that needs the UI thread stalls for the whole download, which can look like a hang for a large full-release download.

A modal progress dialog that pumps messages would fix both.

andrew-polk, line 1: [Devin] **Investigate**: Follow-up message boxes may be owned by the hidden splash form at startup

`ReportCollectionNeedsNewerBloom` correctly passes a null owner for the main dialog, but the follow-up messages go through `BloomMessageBox.ShowInfo`/`ShowWarning`, which use `Shell.GetShellOrOtherOpenForm()`. At startup there is no Shell, and `StartupScreenManager.HideSplashScreenForDialog` only `Hide()`s the splash form rather than closing it, so `Application.OpenForms.LastOrDefault()` can return the still-open-but-hidden splash as owner.

Worth verifying these post-upgrade messages actually appear (and are positioned sensibly) in the pre-collection case.

DistFiles/localization/en/Bloom.xlf
andrew-polk: [Claude Opus 5] Fixed in `b749bc0391` — a genuinely good catch, and one that would have hidden indefinitely for exactly the reason you gave: the code's English fallback means nothing looks wrong.

The entry has been moved below the closing `-->` at line 5847. Verified by parsing the file with an XML parser and asking for the ids, so a commented-out entry simply wouldn't appear: all eight new ids are live (the seven `Collection.*` ones were already outside the block).

src/BloomExe/Collection/MinimumBloomVersionCheck.cs
andrew-polk: [Claude Opus 5] Fixed in `b749bc0391`. Both consequences are real, and the second one is the substantive bug.

`_collectionBeingLockedOut` is now cleared whenever a collection is actually opened (`MinimumBloomVersionCheck.NoteCollectionOpened`, called from `Program.OpenProjectWindow` on success), so the A → B → A case locks out properly the second time. Clearing it inside the lock-out is the one thing we can't do — repeat notifications arrive during shutdown and would put the dialog straight back up — which is why it needed an outside event to reset it.

`LockUserOutOfOpenCollection` now also returns whether it did anything, and `CheckWhetherRepoNowRequiresANewerBloom` passes that through, so a short-circuit no longer tells the idle handler the collection is being torn down when nothing happened.

andrew-polk: [Claude Opus 5] Confirmed, not fixed — this is on the report going to the developer as a decision, along with your point about the wait cursor.

We already knew the download blocks the UI thread and noted in the code that a progress window is wanted; your first point sharpens it usefully, because if `UseWaitCursor` never gets its pump turn then the one mitigation we do have may not exist at all. Both are solved by the same thing: a modal progress dialog that pumps messages. That is new UI rather than a review fix, so it is the developer's call whether it lands here or as a follow-up.

andrew-polk: [Claude Opus 5] Noted, and on the report as something that needs live testing rather than a code fix.

You're right that `HideSplashScreenForDialog` only hides the splash, so it stays in `Application.OpenForms`. Nothing in this branch has been exercised against a live update feed yet, so which owner these follow-up boxes actually land on — and whether they appear sensibly — is one of the things that first real run has to check.

src/BloomExe/Program.cs
andrew-polk: [Claude Opus 5] Devin re-raised this on the latest commit, promoted from a flag to a bug, with a sharper account of how long the gap lasts. Recording it here since this thread is staying open for Andrew's decision.

The mid-session lock-out only fires from a repository *change notification*. A requirement the administrator set **before** this session started produces no notification, so nothing catches it: the startup gate reads the stale local `.bloomCollection` and passes, and the member works in the collection for the whole session. They are blocked at the *next* launch, once the startup sync has written the flag into the local file.

So the exposure is one full session, not a few seconds as we had assumed. The fix Devin suggests is the one on the report: consult the repo copy (the `GetRepoCollectionSettingsContent` / `IsThisBloomTooOldForSettings` pair we already use mid-session) once the Team Collection is known, or re-run the check right after the startup collection-files sync.

src/BloomExe/TeamCollection/TeamCollection.cs
andrew-polk, line 1: [Devin] **Investigate**: Detected minimum version is not stored in the in-memory settings, so a later local save can push a copy back that lacks it

`CheckWhetherRepoNowRequiresANewerBloom` reads the repo's minimum version but, unlike the sibling `UpdateAllowCheckoutsFromRepo`, it does not copy the value into `_tcManager.Settings` (`CollectionSettings.MinimumBloomVersion`).

Bloom now writes `MinimumBloomVersion` back out from memory in `CollectionSettings.Save()`, and several code paths call `Save()` mid-session (e.g. `TalkingBookApi.cs:46`, `PublishToBloomPubApi.cs:233`). Since the local settings object was loaded before the administrator's change arrived, its `MinimumBloomVersion` is empty; any such save writes a local file without the element, and the Team Collection local-file watcher then pushes collection files up, which would remove the administrator's protection from the repository.

The window is realistic: repo notifications can be minutes behind (Dropbox), and an offline user may never receive one this session.

andrew-polk: [Claude Opus 5] Fixed in `c357453f7f` — the best catch of the run, and the one with real consequences for a team.

You're right about the mechanism and about the window. The repo's value is now copied into the in-memory settings whenever we read it, mirroring the `AllowCheckouts` treatment you pointed at. Worth stressing the part that makes it dangerous: it matters **even when this Bloom is new enough to carry on working**, which is exactly the case where nothing else would ever notice — no dialog, no log entry a user would see, just the administrator's protection quietly gone from the repository, removed by the Bloom it was meant to keep out.

One thing I did differently from a straight mirror: failing to read the repo copy is treated as "we don't know" rather than "there is no minimum", so a transient read error (a part-written zip, Dropbox mid-sync) can't clear the setting either. Both cases now have tests in `FolderTeamCollectionTests2`.

src/BloomExe/Collection/MinimumBloomVersionCheck.cs
andrew-polk, line 1: [Devin] **Bug**: User can be trapped in an endless dialog with no way to close Bloom

When Bloom is locked mid-session, the warning is shown again and again in an endless loop with no option to quit, so a user who has no newer Bloom available and no other collection to open cannot get out of Bloom at all.

Impact: the user must kill Bloom from Task Manager, because the only offered ways forward (upgrade, or open a different collection) may both be impossible for them.

`ReportCollectionNeedsNewerBloom` shows a `BloomMessageBox` with `ControlBox = false` and a null owner, so it is application-modal and has no close box; the main window cannot be reached to close Bloom. If the user clicks "Upgrade Bloom" and there is nothing newer on their channel, `UpgradeBloom` reports that and returns false. If they instead click "Open a Different Collection", `Program.ChooseACollection(shell)` is called with a non-null `formToClose`, and cancelling that chooser returns false *without* exiting. Either way the loop immediately puts the modal dialog back up.

The startup path does not have this problem, because there `ChooseACollection` is called with no form to close and cancelling exits Bloom.

src/BloomExe/ApplicationUpdateSupport.cs
andrew-polk, line 1: [Devin] **Bug**: A downloaded upgrade can be applied twice if the user later clicks the restart notification

The exit-time install arranged by `ArrangeToInstallDownloadedUpdateOnExit` does not check whether the user has meanwhile clicked the "Restart Bloom to Update" notification, so the same update can be handed to the installer twice as Bloom closes.

Impact: in rare cases the upgrade install is triggered twice, which can make the upgrade fail or restart Bloom unexpectedly.

The normal path's exit handler only calls `WaitExitThenApplyUpdates` when `!_restartingAfterToastClicked`, precisely because clicking the restart toast already calls it with different arguments. After the silent path has downloaded and arranged the install, `_status` is `DownloadedWaitingForRestart`, so a subsequent update check from the workspace timer will show the restart toast; if Bloom's shutdown is delayed (`Shell.OnClosing` cancels the first close while it saves) the user can click it, and both calls then run.

src/BloomExe/Program.cs
andrew-polk, line 1: [Devin] **Investigate**: Returning true from OpenProjectWindow with a null project context leaves other startup actions running during shutdown

When the user chooses to upgrade, `OpenProjectWindow` returns true after `ProgramExit.Exit()` has been requested. `Application.Exit()` does not necessarily end things promptly (`Shell.OnClosing` cancels the first close while it saves), and the remaining queued startup actions still run on Idle — notably `CheckRegistration`, which can put up its own modal dialog while Bloom is trying to quit for the install.

Worth a look at whether the queue should be drained or disabled once we have decided to quit for an upgrade.

src/BloomExe/Collection/MinimumBloomVersionCheck.cs
andrew-polk: [Claude Opus 5] Fixed in `fbdec6e274`. This was a trap of my own making — the loop was added earlier today to stop a cancel dropping the user back into the collection they'd just been locked out of, and I did not think through the case where neither offered way out is actually available to them.

Cancelling the chooser now quits Bloom, exactly as it does on the startup path, and the loop is gone. The design intent survives — you still cannot stay in the collection — but "none of the above" is now a real answer rather than a dead end.

I considered a third **Quit Bloom** button instead. It is on the report as a decision, since Andrew specified the two-button dialog deliberately and may prefer that.

src/BloomExe/ApplicationUpdateSupport.cs
andrew-polk: [Claude Opus 5] Fixed in `fbdec6e274`. Right on both the mechanism and why the window is real: shutdown isn't instant because `Shell.OnClosing` cancels the first close while it saves, which is exactly enough time for the toast to appear and be clicked.

`ArrangeToInstallDownloadedUpdateOnExit` now checks `_restartingAfterToastClicked` before applying — the same guard, for the same reason, as the handler in `DownloadAndApplyUpdates`.

src/BloomExe/Program.cs
andrew-polk: [Claude Opus 5] Confirmed as plausible, not fixed — it is on the report for Andrew rather than something I should change unasked, and this thread stays open.

The concern is real: `CheckRegistration` raising its own modal dialog while Bloom is trying to quit for an install would be a poor experience and could conceivably interfere with the install. But the fix means changing how the startup action queue behaves once we've decided to quit, which is a startup-sequencing change well beyond a review fix, and we have also hit this run's cap on fix-and-re-review cycles. Flagging it rather than reaching for it.

src/BloomExe/Collection/MinimumBloomVersionCheck.cs
andrew-polk, line 1: [Devin] **Investigate**: Lock-out flow depends on the topmost open form actually being the Shell

`LockUserOutOfOpenCollection` passes `Shell.GetShellOrOtherOpenForm() as Shell` to `Program.ChooseACollection`. `GetShellOrOtherOpenForm` falls back to the last open form when no Shell is found, in which case the `as` yields null and `ChooseACollection(null)` takes the startup branch: on cancel it exits Bloom, and on selection it calls `OpenCollection` while `_projectContext` is still alive, which trips `Debug.Assert(_projectContext == null)` in `OpenProjectWindow` and would leave a second Shell/ProjectContext.

In practice the Shell is open whenever Team Collection monitoring runs, so this is unlikely, but passing the real Shell explicitly (or bailing if it isn't found) would make the invariant safe.

andrew-polk: [Claude Opus 5] Confirmed as a real edge, not fixed — leaving this thread open for Andrew.

The reasoning holds, and it got slightly more interesting with the change we just made: cancelling the chooser now quits Bloom, which is what the `null` branch does anyway, so half of your scenario is no longer a divergence. The remaining half stands — if the topmost form is not the Shell and the user *does* pick a collection, we would open one while a project context is still alive.

Not acting because this run has reached its cap on fix-and-re-review cycles, and because the honest fix is to decide what should happen when there is no Shell (bail? assert?) rather than to paper over the cast. Recorded on the preflight report.

andrew-polk:
[Claude Opus 5] **Consulted Devin** through this branch, up to commit `fbdec6e274`.

Six reviews ran, one per commit — Devin restarts on every push, so this is the record of the whole conversation rather than a single pass. It earned its keep: **three of its findings were real bugs that neither I nor the local review had caught**, and two more were bugs in fixes I had made earlier in the same run.

Fixed and closed on their own threads:

- **An old Bloom could delete the minimum-version setting out of the shared collection.** The worst of the run. `CollectionSettings.Save()` writes the value out from memory, and the in-memory copy predated the administrator's change, so an ordinary save would have pushed a settings file back up with the line missing — removing the protection for the whole team, silently. Fixed in `c357453f7f`.
- **A user could be trapped in the lock-out dialog with no way to quit Bloom.** The dialog has no close box by design; if nothing newer exists on your channel and you cancel the collection chooser, my loop just put the dialog back. Task Manager was the only exit. Fixed in `fbdec6e274`.
- **Choosing to upgrade could cancel an update already downloaded by the workspace timer**, leaving Bloom to throw on the way out and never install it. Fixed in `65d069c707`.
- **A new string was inert**, having landed inside a 170-line XML comment block; the code's English fallback meant nothing would ever have looked wrong. Fixed in `b749bc0391`.
- **The same upgrade could be applied twice**, if the restart notification was clicked during the pause while Bloom saves on shutdown. Fixed in `fbdec6e274`.
- Plus: the missing XLF entries, the idle-handler re-entering itself while a modal dialog was up, the lock-out guard never being cleared, and the wrong PR description.

Left open for Andrew, and on the preflight report:

- The Team Collection startup gap — a member whose repo already carries the requirement gets one whole session before it bites.
- The UI freezing during the download, and whether the wait cursor appears at all.
- Whether the follow-up messages end up owned by the hidden splash form (needs a live run to know).
- Other startup jobs still running while Bloom quits for the upgrade.

Around ten informational items were assessed and not mirrored, the substantive ones being that the command-line and bulk-upload paths remain ungated (a deliberate decision, recorded in the code), that `OpenProjectWindow` returns true with a null project context on the upgrade path (now commented for whoever adds the next caller), and that the update URL lookup uses its cached result.

CI is green. CodeRabbit is switched off on this repo (`auto_review.enabled: false`), so its silence says nothing either way.

src/BloomExe/Program.cs
andrew-polk, line 1: [Devin] **Bug**: A user shut out of a shared collection can get straight back into it by picking the same collection again

The version gate that decides whether a collection may be opened is checked against the copy of the collection's settings stored on this computer (`MinimumBloomVersionCheck.IsThisBloomTooOld`), which has not yet received the new minimum version an administrator just set for the shared collection, so the collection re-opens on the very next attempt.

Impact: someone told they cannot use a shared collection can immediately re-enter it just by choosing it again from the collection chooser, defeating the protection until they restart Bloom.

When the minimum arrives mid-session, `CheckWhetherRepoNowRequiresANewerBloom` reads the *repository's* copy and stores it only in memory, because Bloom deliberately does not overwrite local collection settings while running. `LockUserOutOfOpenCollection` then offers "Open a Different Collection". If the user selects the same collection, the gate reads the on-disk `.bloomCollection`, which still lacks `MinimumBloomVersion`, so the collection opens — and `NoteCollectionOpened()` then clears the lock-out record. The startup Team Collection sync copies the repo settings down only *after* the gate has already passed.

andrew-polk: [Claude Opus 5] Fixed in `dd3c5e08eb`. This was the hole that made the whole mid-session lock-out theatre — you could be shut out, click "Open a Different Collection", pick the same one, and be back in, with the lock-out record cleared behind you.

What the repository tells us is now remembered against the collection itself, not only in the settings object, and the gate consults that before falling back to the file. The lookup distinguishes "the repository never said anything about this collection" from "the repository says there is no requirement" — the second is a real answer and has to be able to clear a stale requirement still sitting in the local file.

I did not take your other suggested route, writing the value into the local settings file. Bloom's not rewriting local collection settings mid-session is a rule this feature should live within rather than punch through, and it would have us editing a file the administrator may be editing at that moment.

Two tests: the gate refuses when only the repository knows about the requirement, and stops refusing when the repository has withdrawn one the local file still carries.

src/BloomExe/Collection/MinimumBloomVersionCheck.cs
andrew-polk, line 1: [Devin] **Investigate**: A remembered repo value of "" completely suppresses the local file's requirement

`IsThisBloomTooOld` prefers `MinimumVersionLearnedFromRepo` over the file, and treats `""` (repo says: no requirement) as an answer that wins.

Note that `RememberRepoMinimumBloomVersion` is also called on the push-local-to-repo path in `SyncLocalAndRepoCollectionFiles`, and reads the zip we have just written. If that read happens to return settings without the element — e.g. a partially-synced or older zip that still lacks it, while the local file has just been hand-edited to add it — the remembered `""` would mask the local requirement for the rest of the session.

Not reachable in the common flows I traced, but the precedence rule is worth being deliberate about.

andrew-polk: [Claude Opus 5] Fixed in `e580da0af3`, and you were right to push on the precedence rather than only on the reachability.

The rule is now "whichever demands more" rather than "the repository wins outright". Where the two sources disagree we take the stiffer requirement, so a repository copy that is momentarily missing the element — a zip mid-sync, exactly your scenario — can no longer mask a requirement the local file is carrying.

The reasoning for choosing that direction: this is a protection, so the two ways of being wrong are not equal. Keeping someone out slightly longer than necessary is a nuisance they can clear by restarting, once the startup sync has brought the local file into line. Letting someone in who should have been kept out is the thing the feature exists to prevent. So the tie goes to whoever asks for more.

Empty and unparseable still count as no requirement at all on either side, which keeps the "nobody loses their collection to a typo" rule intact — a typo in one source just leaves the other to speak.

Tests updated to pin the new rule in both directions: the repository saying nothing no longer clears a requirement in the file, and a file asking for less does not water down a stiffer one the repository has since reported.

src/BloomExe/Program.cs
andrew-polk: [Claude Opus 5] Fixed in `7901985ea9`. Andrew's decision was to close this properly rather than accept the one-session window, so the gate now consults the repository's own copy of the settings for a Team Collection, read straight from the shared folder.

That is the earlier of the two places you suggested, and it is the one that actually closes the hole: the check has to happen before Bloom commits to opening the collection, and `SyncLocalAndRepoCollectionFiles` runs long after that. `FolderTeamCollection` grew a static reader for the purpose, taking a plain folder path, because at that moment no `TeamCollection` object exists yet.

Three details worth recording:

- Being unable to reach the shared folder means "we don't know", never "there is no requirement". An offline member is judged on their local file, rather than either shut out by a transient failure or let in by one.
- It costs one small read, and only for collections that actually carry a `TeamCollectionLink.txt`.
- It folds in through the same "whichever demands more" rule as the other two sources, so it can tighten the answer but never loosen it.

Test: `IsThisBloomTooOld_OnlyTheRepoKnowsAboutTheRequirement_StillSaysTooOld` builds a repository whose copy demands 99.0 while the local file says nothing, which is exactly the morning-after state you described. I verified it fails without the new call.

src/BloomExe/Collection/MinimumBloomVersionCheck.cs
andrew-polk: [Claude Opus 5] Investigated at Andrew's request, no code changed yet. Summary: **the wait cursor almost certainly does nothing, so "good enough" is not available**, and Bloom already has the right component to fix it properly.

**Does the wait cursor appear?** Two reasons to think not, and they are independent.

The first is the one you raised: `Application.UseWaitCursor = true` marks the open forms, but the cursor only actually changes when the window processes `WM_SETCURSOR`. We set the flag and then immediately block the UI thread, so nothing is processed until the download is over, by which time we have set it back to false.

The second is worse and applies to the startup case regardless: `Application.UseWaitCursor` acts on `Application.OpenForms`, and at the startup gate there are no open forms to act on. There is no Shell yet, and `StartupScreenManager.HideSplashScreenForDialog` only *hides* the splash. So even with a pumping message loop there would be nothing to show a wait cursor on.

I could not prove this by observation: a developer build refuses to update itself, so the live run returned from `TryDownloadUpdateWithoutToasts` immediately and there was no window during which to look. That part is analysis, and it is the part worth checking against a real download.

**What we already have.** `BrowserProgressDialog.DoWorkWithProgressDialog` (`src/BloomExe/MiscUI/BrowserProgressDialog.cs:43`) is built for exactly this shape: it runs the work on a `BackgroundWorker` while the dialog owns the UI thread, so the window keeps painting. `TeamCollection.cs:2919` is a working example — a `ReactDialog` on the `progressDialogBundle`, a `doWhat` returning bool, and a fixed size. It needs `SocketServer`, which the startup path already has, since we call `BloomServer.EnsureListening()` before showing the too-old dialog.

**Recommendation.** Replace the `Task.Run(...).GetAwaiter().GetResult()` in `MinimumBloomVersionCheck.UpgradeBloom` with `BrowserProgressDialog.DoWorkWithProgressDialog`, moving `TryDownloadUpdateWithoutToasts` into the `doWhat` callback and dropping the wait cursor entirely. That fixes the frozen window and the cursor together, and it uses a component the codebase already trusts rather than inventing one.

Two things to decide when it is written: Velopack's `DownloadUpdatesAsync` takes a progress callback, so we can report real percentages rather than a spinner; and the dialog would want `showReportButton = "never"` and no cancel, since a half-cancelled update is not a state we want to reason about.

Effort: moderate but contained — one method reworked, no new UI component, no change to what the user is told.

src/BloomExe/Program.cs
andrew-polk: [Claude Opus 5] Investigated at Andrew's request, no code changed yet.

**How the queue works.** `StartupScreenManager.StartManaging` subscribes `DoStartupAction` to `Application.Idle` (`src/BloomExe/MiscUI/StartupScreenManager.cs:59`). It runs one queued action per idle event and unsubscribes itself only when the queue is empty (`:201`). Nothing in it knows or cares that Bloom has been asked to quit, so after the user chooses to upgrade and we call `ProgramExit.Exit()`, the remaining actions keep being handed out on every idle turn while shutdown proceeds. `Application.Exit()` is not instant — `Shell.OnClosing` cancels the first close while it saves — so there is a real window, and `CheckRegistration` is the action most likely to use it to put up a dialog.

**Three options.**

1. **A quit flag consulted by the queue.** `ProgramExit` already knows: it sets `_forceShutdownThread` the moment `Exit()` is called, so it needs only `public static bool IsExiting => _forceShutdownThread != null;` and one early return at the top of `DoStartupAction`. About five lines, in the place that already owns the fact, and it covers every future path that quits during startup rather than just ours. Downside: it changes shared startup behaviour for all callers, so it wants a moment's thought about whether any queued action *should* still run while quitting (I did not find one that should).

2. **Stop the queue explicitly when we decide to upgrade.** A `StartupScreenManager.StopRunningStartupActions()` that unsubscribes and clears the list, called from `MinimumBloomVersionCheck` just before `ProgramExit.Exit()`. Around ten lines, and the blast radius is exactly our feature. Downside: it is a second mechanism for "we are quitting", and the next path that quits mid-startup has to remember to call it.

3. **Leave it.** Nobody has seen this happen; it is reasoning from the code, and the window is a second or two.

**Recommendation: option 1.** It is smaller than option 2, it puts the knowledge where it already lives, and "do not start new startup work once we have decided to quit" is a rule that is true independently of this feature. I would pair it with a quick look at the queued actions to confirm none of them needs to run during shutdown.

Not doing it in this PR unless you say so: it changes shared startup sequencing, which is a different kind of risk from the rest of this branch, and it cannot be verified here — the upgrade path that triggers it has never run on a build that can actually update itself.

src/BloomExe/Collection/MinimumBloomVersionCheck.cs
andrew-polk, line 1: [Devin] **Bug**: Team members stay shut out of a shared collection even after the administrator removes or lowers the version requirement

The stricter of the shared folder's requirement and the copy on this computer is always used (`MoreDemandingOf`), and the copy on this computer is only refreshed after the collection is allowed to open, so a requirement that the administrator has since removed or lowered keeps blocking the member forever.

Impact: a team member locked out by a mistaken version requirement can never get back into the collection from inside Bloom, even after the administrator fixes it.

1. Administrator sets `MinimumBloomVersion` to something too new; the sync writes it into the member's local `.bloomCollection`.
2. Administrator realises the mistake and removes (or lowers) the element in the repository.
3. On the member's next launch the gate runs *before* any Team Collection sync. `MinimumVersionInTeamCollectionRepo` correctly reports "" (or the lower value), but `MoreDemandingOf` keeps the stale local value, so `IsThisBloomTooOld` returns true.
4. Because the collection never opens, the startup sync that would rewrite the local file never runs — so step 3 repeats on every launch. The only escape is hand-editing the local `.bloomCollection`.

andrew-polk: [Claude Opus 5] Fixed in `f14d7ab94a`. You caught the trap the previous fix had just built, and the reasoning you demolished was mine — the comment claiming a withdrawal "is honoured at the next startup, once the sync has brought the local file into line" was exactly wrong for the only people it mattered to, because being refused is precisely what stops that sync running.

The rule is now: **when this is a Team Collection and we can read the repository, the repository decides** — including when it says there is no requirement. It has to be able to lift a requirement, not only impose one.

The earlier "whichever demands more" rule survives, but only where it was actually justified: when we cannot reach the shared folder, or it is not a Team Collection at all. Being unable to see the repository is not the same as the repository saying "no requirement", so an offline member keeps the protection rather than silently losing it, and the session-long memory still stops someone we just shut out from re-picking the same collection and walking back in past a stale file.

Worth noting the two failure modes are not symmetrical, which is what settles the direction. Wrongly keeping someone out for a moment is a nuisance. Wrongly keeping a whole team out for ever, recoverable only by hand-editing a file on every machine, is the kind of thing that makes an administrator afraid to use the feature at all.

Test: `IsThisBloomTooOld_RepoHasWithdrawnTheRequirement_LetsThemBackIn` puts a requirement in the local file that the repository no longer has, and asserts the member gets back in. The first-launch test you prompted earlier still passes alongside it, so imposing and lifting both work from the repository.

andrew-polk, line 1: [Devin] **Investigate**: Startup gate now does a synchronous read of the Team Collection shared folder before anything else

`IsThisBloomTooOld` calls `MinimumVersionInTeamCollectionRepo` on every interactive collection open, which stats the shared folder, opens `Other Collection Files.zip` and reads one entry. For a Team Collection whose shared folder is on an unresponsive network share, `Directory.Exists`/file open can block for many seconds at a point in startup where there is no UI at all (splash just hidden, no progress).

Team Collection startup already does repo work, but that happens later and under the progress dialog. Worth confirming this is acceptable, or caching/short-circuiting when the local file already declares a minimum this Bloom meets.

andrew-polk: [Claude Opus 5] Bounded in `c7a0448f12`. Fair catch on a risk my own previous commit introduced, and at the worst possible moment in startup — the splash is already hidden and there is no window, so a stalled share would look exactly like Bloom having died.

The read now runs on a background task with a five-second cap. Not answering in time is treated as "we don't know", which is a state this code already had a meaning for: fall back to the local copy and whatever the repository told us earlier in the session.

I looked at your short-circuit suggestion and could not make it work, which is worth writing down so nobody tries again. Skipping the repository when the local file declares a minimum we *meet* would reopen the first-launch gap, since the repository may have gained a stricter one. Skipping it when the local file declares one we *don't* meet would reopen the permanent lock-out, since the repository is the only thing that can lift it. Both directions genuinely need the answer, so bounding the wait was the available lever rather than avoiding it.

Five seconds is a guess at the right number: long enough that a healthy share, including a Dropbox folder that is really local disk, answers comfortably; short enough that a dead one does not read as a hang. Worth revisiting if anyone sees it fire in practice — the log records it when it does.

src/BloomExe/TeamCollection/TeamCollection.cs
andrew-polk, line 1: [Devin] **Bug**: A hand-added minimum-version requirement can be silently erased for the whole team

After pushing an administrator's newly hand-edited requirement to the shared folder, Bloom re-read that requirement *from the shared folder* rather than from the file it had just pushed, so a momentary read failure left Bloom unaware of it and the next ordinary save would delete it for everyone.

Impact: the administrator's minimum-version protection can quietly disappear from the collection, letting older Blooms back in without anyone noticing.

In the non-startup branch of `SyncLocalAndRepoCollectionFiles`, Bloom has just written the local collection files up to the repository and then called `RememberRepoMinimumBloomVersion(TryGetRepoCollectionSettingsContent())`. That reader swallows any exception and returns null, and `RememberRepoMinimumBloomVersion` deliberately does nothing when the content is null. If reading the zip failed just after we wrote it — Dropbox mid-sync, a part-written zip, exactly the transient conditions the surrounding comments cite — the in-memory `MinimumBloomVersion` stayed empty. `CollectionSettings.Save()` rebuilds the file from memory and only writes the element when non-empty, so the next ordinary save dropped it from the local file and the sync pushed that erasure to the repository: the precise failure the comment there says it is preventing.

The value is already available locally — that is what was just pushed — so reading the local settings file would not have this hole.

andrew-polk: [Claude Opus 5] Fixed in `69225b152a`, and your last sentence was the whole answer: the value was already in our hands and we went back to the network for it anyway.

It now reads the local settings file, which is by definition what we have just pushed, so there is no window in which a half-written zip can make us forget a requirement we ourselves published a moment ago.

Worth naming the irony you spotted: the code that erased the setting was the code added to *stop* the setting being erased. It guarded against the in-memory copy being stale, then took its replacement value from the one source that can transiently fail, and treated failure as "nothing to record".

Test: `SyncLocalAndRepoCollectionFiles_RepoReadFails_StillRemembersWhatWePushed` uses a Team Collection subclass whose repository read always throws, and asserts the pushed requirement is still picked up. It fails against the previous code, so it is testing the fix rather than the happy path.

src/BloomExe/Collection/MinimumBloomVersionCheck.cs
andrew-polk, line 1: [Devin] **Investigate**: Repository copy overrides a local requirement entirely, not "whichever demands more"

Note the asymmetry: when the Team Collection repository can be read, its answer is used verbatim, including an empty value, so a requirement still present in the local `.bloomCollection` is ignored. The `MoreDemandingOf` logic only applies when the repository cannot be reached.

This is deliberate (documented in the comment, and covered by `IsThisBloomTooOld_RepoHasWithdrawnTheRequirement_LetsThemBackIn`) so an administrator can undo a mistake, but it also means anyone able to write the shared folder can lift the protection instantly for every member, and that a repository zip that is momentarily complete-but-old silently wins over the local file.

Worth confirming that is the intended trust model before this is relied on for Cloud syncing.

andrew-polk: [Claude Opus 5] Confirmed as deliberate, and left open for Andrew, because the second half of your question is genuinely his to answer rather than mine.

On the mechanics you describe: yes, and it was chosen with the alternative in front of us. Making the repository authoritative is what stops a mistaken requirement locking a team out permanently — a member who is refused never opens the collection, so their own copy never refreshes, and "whichever demands more" left them stranded for ever. Between briefly trusting a stale-but-readable repository copy and permanently stranding a team, the first is much the better failure.

On the trust model, which is the part worth Andrew's eye: "anyone who can write the shared folder can lift the protection" is not really a new power this introduces. The shared folder is where an administrator *sets* the requirement in the first place, and anyone who can write there can already edit the collection settings, add and remove books, and change checkout state. The protection is against an old Bloom mangling files, not against a hostile teammate.

That said, your framing is the right one for what comes next. This is preparation for Cloud syncing, and a cloud service is a different setting: there the requirement would presumably be something the service asserts rather than a line in a file every member can rewrite. If the flag ever needs to be tamper-resistant rather than advisory, this is the decision to revisit, and the code comment already says as much for the command-line paths that are ungated for the same reason.

andrew-polk: [Claude Opus 5] Devin has now promoted this from a flag to a bug on the latest commit, which matches what we found: the wait cursor is not a mitigation, so this is a plain freeze with nothing on screen.

Andrew's decision was to investigate before coding, and that investigation is in the reply above. The short version: `Application.UseWaitCursor` cannot help here, because beyond the message-pump problem it acts on `Application.OpenForms` and at the startup gate there are none — no Shell yet, and the splash is only hidden. So there is nothing to show a wait cursor on even in principle.

The recommendation stands: `BrowserProgressDialog.DoWorkWithProgressDialog`, which already exists for exactly this shape and is used by the Team Collection sync. Leaving this thread open until Andrew decides whether it lands in this PR or a follow-up.

andrew-polk: [Claude Opus 5] **Now settled by observation, not analysis.** Andrew ran the real upgrade himself and reports: no wait cursor, and no indication of progress of any kind.

Worth being precise about how that was obtained, because a developer build normally refuses to update. `IsDev` is decided by the *path* of `Bloom.dll` — `/output/Debug/` means developer, `/Bloom<channel>/current/` means an installed channel — so this branch's binaries were laid over a copy of an installed Beta. It then reported channel `Beta`, found 6.4.105 → 6.4.108 on the real feed, and genuinely downloaded a 90 MB package. That is the first time this code path has run end to end.

So the question you raised is answered in the worst direction: during that download there is nothing on screen at all. At the startup gate no main window exists — the collection never opened — and the only form in `Application.OpenForms` is the hidden splash, so `Application.UseWaitCursor` has nothing visible to mark even before the blocked message pump is considered. The user simply sees the dialog vanish and the machine sit there.

(One correction to my earlier reply: I said there were *no* open forms at that point. Too strong — a hidden form is still in `OpenForms`. The conclusion holds for the more precise reason above.)

The recommendation is unchanged but no longer speculative: `BrowserProgressDialog.DoWorkWithProgressDialog`, which runs the work on a background worker while the dialog owns the UI thread, as `TeamCollection.cs:2919` already does. Velopack's `DownloadUpdatesAsync` takes a progress callback, so a real percentage is available rather than a spinner. Leaving this thread open until Andrew decides whether it lands in this PR or a follow-up.

andrew-polk:
[Claude Fable 5 from Andrew Polk's machine during preflight] Superseded: this commit is now the first of the two commits on #8213, which is based directly on Version6.4 — the stacked arrangement (and the BL-16690-velopack-spike branch) is no longer needed. Closing to keep one PR for BL-16690.

PR #8207, opened 2026-08-14
Fix concurrent writes of the same AI image history file (BL-16702)

JohnThomson:
Assigning one AI-generated image to two book-image slots and clicking **Replace** failed with `Failed to write host file history/<id>.png:503` and a yellow "Cannot access …" box, and nothing in the book was replaced.

### Cause

The AI image editor writes each assigned result to the host **before** posting the commit, and it does so once per slot, **concurrently** (`Promise.all`). One image in two slots therefore means two simultaneous POSTs of the same `history/<resultId>.png`.

Bloom's `aiImageEditor/file` POST handler wrote through a temp file named after the target (`<file>.tmp`), so the two requests collided on that one path: `RobustFile.Create` opens with `FileShare.None` and, unlike most of `RobustFile`, does **not** retry, so the loser threw `IOException`. `ApiRequest` turns an `IOException` into a 503 "Cannot access …" plus a NonFatalProblem, the editor reports it as "Failed to write host file", and that one failure abandoned the **whole** commit before Bloom was ever asked to change an image.

Confirmed live before fixing — two concurrent POSTs of the same file name against a running Bloom returned `200` and `503`, while one on its own returned `200`. The affected book's history PNG showed the winning write's timestamp, and the book folder contained no `ai-image*.png` at all: the commit never ran.

### Fix

Serialize changes to a given file. `HandleFile`'s POST body write moves into `WriteRequestBodyToFile`, which takes a per-path lock; the DELETE branch takes the same lock. A GET deliberately doesn't — a read can't corrupt anything, and making a response stream wait behind a multi-MB upload would cost more than it saves. Nothing changes for the AI image editor, or for single-slot commits.

The rest of the "same image twice" path was already sound: `matchReplacementsToElements` handles two current-page replacements sharing an `oldSrc`, and each slot gets its own `ai-image*` copy in the book folder.

### Tests

Three new tests on `WriteRequestBodyToFile`: body written + folder created, empty body truncates, and the BL-16702 regression — four concurrent writes of one file all succeed and leave a complete file with no temp file behind. The regression test was confirmed to **fail** with the lock defeated (the very same "used by another process" `IOException` from `RobustFile.Create`) and to pass with it.

Ref: https://issues.bloomlibrary.org/youtrack/issue/BL-16702

[Devin review](https://devinreview.com/BloomBooks/BloomDesktop/pull/8207)

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8207)

### DIFF
--- a/src/BloomTests/web/controllers/AiImageEditorApiTests.cs
+++ b/src/BloomTests/web/controllers/AiImageEditorApiTests.cs
@@ -391,6 +391,56 @@ namespace BloomTests.web.controllers
             Assert.That(new FileInfo(path).Length, Is.EqualTo(0));
         }
 
+        // A body that hands over some bytes and then fails, standing in for a client that goes
+        // away mid-upload.
+        private class FailingStream : MemoryStream
+        {
+            public FailingStream(byte[] initialBytes)
+                : base(initialBytes) { }
+
+            public override int Read(byte[] buffer, int offset, int count)
+            {
+                var read = base.Read(buffer, offset, count);
+                if (Position >= Length)
+                    throw new IOException("the client went away");
+                return read;
+            }
+        }
+
+        [Test]
+        public void WriteRequestBodyToFile_WriteFails_KeepsTheOldFileAndLeavesNoTempBehind()
+        {
+            // Writing through a temp file exists so a half-finished upload can't destroy the
+            // good file that was there. The temp itself must not survive either: nothing would
+            // serve or enumerate it, but it can be multi-MB and would sit in the book's folder
+            // for the life of the book.
+            var path = HistoryFilePath("result.png");
+            var good = new byte[] { 1, 2, 3 };
+            AiImageEditorApi.WriteRequestBodyToFile(path, new MemoryStream(good));
+            Assert.That(File.ReadAllBytes(path), Is.EqualTo(good), "setup: the good file is there");
+
+            Assert.That(
+                () =>
+                    AiImageEditorApi.WriteRequestBodyToFile(
+                        path,
+                        new FailingStream(new byte[5000])
+                    ),
+                Throws.InstanceOf<IOException>(),
+                "a failed body read must still fail the request"
+            );
+
+            Assert.That(
+                File.ReadAllBytes(path),
+                Is.EqualTo(good),
+                "the file that was there must be untouched"
+            );
+            Assert.That(
+                Directory.EnumerateFiles(Path.GetDirectoryName(path), "*.tmp"),
+                Is.Empty,
+                "the failed write's temp file must have been cleaned up"
+            );
+        }
+
         [Test]
         public void WriteRequestBodyToFile_ConcurrentWritesOfTheSameFile_AllSucceed()
         {
--- a/src/BloomTests/web/controllers/AiImageEditorApiTests.cs
+++ b/src/BloomTests/web/controllers/AiImageEditorApiTests.cs
@@ -1,8 +1,11 @@
+using System;
+using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Drawing;
 using System.Drawing.Imaging;
 using System.IO;
 using System.Linq;
+using System.Threading;
 using Bloom.Book;
 using Bloom.web.controllers;
 using NUnit.Framework;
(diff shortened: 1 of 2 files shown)


src/BloomTests/web/controllers/AiImageEditorApiTests.cs
```diff
@@ -348,6 +351,201 @@ public void ImportImageIntoBookFolder_CorruptImage_StillLandsInTheBookFolder()
...
+            // Holding the destination open blocks the delete half of the swap, so the write
+            // gets as far as the swap and then fails — which is the case we care about.
+            using (new FileStream(path, FileMode.Open, FileAccess.Read, FileShare.Read))
+            {
+                Assert.That(
+                    () => AiImageEditorApi.WriteRequestBodyToFile(path, new MemoryStream(newBytes)),
+                    Throws.InstanceOf<IOException>(),
+                    "setup: holding the file open should make the swap fail"
+                );
+            }
```
JohnThomson, line 469: [Devin] **Investigate**: Swap-failure test depends on Windows file-locking semantics

Holding the destination open with `FileShare.Read` is what makes the swap fail, which is true on Windows but not on Unix (where unlinking an open file succeeds); on a non-Windows run this test would likely fail rather than skip. Also, `RobustFile.Move` retries before giving up, so this test spends the retry window waiting. Fine if the suite is Windows-only, worth knowing if it is not.

(`src/BloomTests/web/controllers/AiImageEditorApiTests.cs:460-469`)

JohnThomson: [Claude Opus 5 from John Thomson's machine during preflight] Not an issue: this test project cannot run anywhere but Windows. `src/BloomTests/BloomTests.csproj` targets `net8.0-windows` with `<RuntimeIdentifiers>win-x64;win-arm64</RuntimeIdentifiers>`, and the fixtures reference WinForms and GDI+ throughout, so there is no non-Windows run for the file-locking behavior to differ on. Deliberately keeping the test as it is rather than guarding it with a platform attribute, which would only suggest a possibility that does not exist.

On the retry window: yes, `RobustFile.Move` retries the delete before giving up, so this test takes about two seconds. That is the price of driving the real failure path instead of mocking it, and the whole fixture still runs in single-figure seconds.

JohnThomson:
[Claude Opus 5 from John Thomson's machine during preflight] Consulted Devin on 2026-08-14 up to commit `2959246602`.

It reviewed all four pushed commits and found **no bugs** on any of them. It raised **one Investigate flag** and a run of informational notes; the notes were the most useful review of this run, and three of them became commits:

- **Interrupted upload left a multi-MB `.tmp` in the book folder forever** → fixed in `d6ff841d27`.
- **The concurrency test could pass without the writes ever overlapping** → fixed in `74996a2e04` (the writers are now released together from a barrier).
- **The cleanup above could destroy both copies of a file**, because swapping the new file in deletes the destination before moving, so a failure in between has already taken the old file — and deleting the temp as well threw away the new bytes too → fixed in `2959246602`, which also fixes two test-honesty points Devin was right about (the failing-body test never actually produced a truncated file, and the overlap assertion could not fail).
- **Investigate: the swap-failure test depends on Windows file-locking semantics** → assessed as not an issue and resolved on [its thread](https://github.com/BloomBooks/BloomDesktop/pull/8207#discussion_r3786860352): this test project targets `net8.0-windows` with Windows-only runtime identifiers, so there is no non-Windows run for the behavior to differ on.

Notes deliberately not acted on, for the record: the lock is held while the request body is read (so a second write **of the same file** waits — going to the developer as a decision, with Devin's `<file>.<guid>.tmp` alternative spelled out); a read landing exactly on the swap can 404 (as old as the endpoint, and nothing reads a file while writing it — now said in the comment); the lock table grows by one small entry per history file (documented); and the serialization is in-process only (fine, since one Bloom has a given collection open — now said in the comment too).

CI (`pr-automation`) is green on every commit. CodeRabbit is present in the repo but switched off for automatic review (`.coderabbit.yml` sets `auto_review.enabled: false`), so it posted nothing. The full C# suite is green at this commit: 3102 passed, 0 failed, 13 skipped.

JohnThomson:
[Claude Opus 5 from John Thomson's machine during pr-ready-for-human] Consulted Devin again on 2026-08-14 up to commit `af0ae41188` (the comment-only commit recording the de-duplication decision): still **0 bugs**, and its one Investigate flag is the same Windows file-locking note already [answered and resolved above](https://github.com/BloomBooks/BloomDesktop/pull/8207#discussion_r3786860352). Nothing new.

andrew-polk (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8207)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on JohnThomson).

andrew-polk (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8207)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on JohnThomson).

PR #8208, opened 2026-08-14
Fix the four XLF entries wrongly marked obsolete (BL-16686)

JohnThomson:
Resolves the four English XLF entries that `901bf1f333` ("Remove English XLF entries marked obsolete as of 6.3 or earlier") deliberately kept back because their "obsolete" note turned out to be wrong.

Three different problems were wearing the same label:

### 1. `PublishTab.Feature.Games.None` / `.Present` — the note was just false

Both were **added** already carrying `Obsolete as of 6.2`, in `d4e42040df` ("Activity -> Game in the UI"), copy-pasted from the `PublishTab.Feature.Activities.*` entries immediately above them. But they are the *replacements* for those entries — that commit deleted `noActivitiesTooltip`/`hasActivitiesTooltip` and put `noGamesTooltip`/`hasGamesTooltip` in their place — and they are still the live tooltips at `PublishFeaturesGroup.tsx:188,192`.

→ Notes removed. No code change needed.

### 2. `EditTab.Toolbox.ComicTool.Options.Style` — an unfinished migration

`c57c1a3f2c` (BL-16088) added `...Options.ElementType` and converted `CanvasToolControls.tsx`, but the declarative canvas control system — which landed six weeks earlier — kept its own copies of the old id in `canvasControlRegistry.ts` and `canvasPanelControls.tsx`.

Nothing user-visible was affected: that system's tool-panel surface isn't wired up yet (`getToolPanelControls` has no callers), so the only label on screen is already "Element Type". That's also why the automated "is this id still referenced?" check kept the entry — it can't distinguish a live reference from one in unreachable scaffolding.

→ Both sites now use `...Options.ElementType`; the old entry is deleted.

### 3. `TemplateBooks.PageLabel.Quiz Page` — genuinely superseded, and effectively invisible

Superseded in 6.2 by `TemplateBooks.PageLabel.Comprehension Quiz`. Tracing every consumer showed the translation is essentially never displayed for an ordinary book:

| Path | What actually happens |
|---|---|
| Edit tab page | localized by `editablePage.ts:374`, then hidden — `.bloom-page .pageLabel { display: none }` (BL-7925), overridden only for `contenteditable` template pages |
| Page thumbnails (`PageListApi.cs:206`), video page-range (`PublishApi.cs:737`) | `GetCaptionOrPageNumber` returns the **page number** for any `numberedPage`, so the id is `"7"`, not the label |
| ePUB TOC / progress | xmatter names plus one "Content"; `GetNumberOrLabelOfPageWhereElementLives` also returns the page number for numbered pages |
| Collection preview | `.pageLabel { display: none }` except on `.preview.template` |
| Publish "pages removed" warnings, spreadsheet export | raw English label, no localization |

The common thread: `LocalizePageLabel` is only reached for pages with **no page number** or for back matter, and a quiz page is a numbered content page.

There's also no migration — `RepairPageLabelLocalization` only *fills in* a missing `data-i18n`, so old books keep the id forever. Bloom does reach into those very pages twice (maintenance levels 6 and 10) without touching the label.

→ Entry deleted. Accepted cost: a pre-6.2 **template** book would show this one label in English. The translated XLFs keep their entries, so re-adding the id would recover them.

### The shipped content that used it

`The Moon and the Cap` sample shell had two pages on the old id. Both carry `data-pagelineage="F125A8B6-EA15-4FB7-9F8D-271D7B3C8D4D"` — the id of the template page that `simpleComprehensionQuiz.pug` now labels "Comprehension Quiz" — so that's what they become, `data-i18n` and English inner text together (the inner text is what spreadsheet export and `RepairPageLabelLocalization` read).

This differs from the card's suggestion of `TemplateBooks.PageLabel.Quiz`, which is a different, deliberately shorter label ("Used in page preview for Comprehension Question pages. Should be short."). The lineage GUID makes "Comprehension Quiz" the evidenced match.

I also re-checked all 39 ids that `901bf1f333` deleted against shipped content using the correct paths and extensions (sample shells are `.htm`, and templates now live under `src/content/`) — all clean; `Quiz Page` was the only one still in use.

Only files under `DistFiles/localization/en/` are touched.

### Noted, not fixed

`Games.html` uses `TemplateBooks.PageLabel.Drag Letters to Targets`, which `fc8ed04b98` deleted from the English XLF — the same class of mistake with the halves reversed (content updated, string list not). That label silently falls back to the English DOM text and can never be translated. Left alone as out of scope for this card.

Ref: https://issues.bloomlibrary.org/youtrack/issue/BL-16686

[Devin review](https://devinreview.com/BloomBooks/BloomDesktop/pull/8208)

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8208)


DistFiles/localization/en/Bloom.xlf
```diff
@@ -2366,12 +2366,6 @@
         <note>ID: EditTab.Toolbox.ComicTool.Options.ShowTail</note>
         <note>This checkbox indicates if a line tail should be shown or not.</note>
       </trans-unit>
-      <trans-unit id="EditTab.Toolbox.ComicTool.Options.Style" sil:dynamic="true">
```
JohnThomson, line 2369: [Devin] **Bug**: Two English text entries are deleted from the translation file instead of being marked as no-longer-used

Devin flags that this PR deletes two translated entries from `DistFiles/localization/en/Bloom.xlf` outright — `EditTab.Toolbox.ComicTool.Options.Style` (was line 2369) and `TemplateBooks.PageLabel.Quiz Page` (was line 6430) — where the repo's written rule says an unused string should be kept and annotated instead.

It quotes `src/BloomBrowserUI/AGENTS.md:65`, and the quote is accurate:

> If a string is no longer used, we do not remove it. Instead, add a note like this: `<note>Obsolete as of 6.2</note>`.

**Why it might matter:** any book or shipped Bloom version still asking for one of these ids now gets the English text rather than a translation.

**What we already know.** The deletions are deliberate and were directed by the developer during this session, with the stated reasoning that keeping dead entries costs every future translator a little work. There is precedent: commit `901bf1f333` on master deleted 39 such entries the same way. Only files under `DistFiles/localization/en/` are touched, so every translated `.xlf` keeps its entry and re-adding an id would pick its translations back up. For `Quiz Page` specifically, the one piece of shipped content that used it was migrated in this same commit, and tracing every consumer showed the label is never displayed for an ordinary book (see the PR description).

Devin lands on the real question rather than on the code:

> If the team has deliberately adopted a delete-obsolete-entries policy (as a prior commit suggests), the rule files should be updated to match.

That makes this a **policy call, not a defect**: either `AGENTS.md` (and `.github/skills/xlf-strings/SKILL.md`, which says the same for the modify case) should be updated to describe what we actually do, or these two entries should be kept with notes after all. Leaving this thread open for that decision.

<sub>Posted by Claude Opus 5 (1M context) during `preflight`.</sub>

src/BloomBrowserUI/bookEdit/toolbox/canvas/canvasControlRegistry.ts
JohnThomson, line 1: [Devin] **Investigate**: Neighbouring registry entry references a localization id that does not exist in the English XLF

While checking the id swap this PR makes, Devin looked at the entry immediately next to it in the same registry and found the same class of problem, already present on master:

`src/BloomBrowserUI/bookEdit/toolbox/canvas/canvasControlRegistry.ts:1033` asks for `EditTab.Toolbox.ComicTool.Options.OutlineColor`, but there is no such entry in `DistFiles/localization/en/`. The two that do exist are `...Options.OuterOutlineColor` (Bloom.xlf:2329) and `...Options.TextOutlineColor` (Bloom.xlf:2424).

**Verified** by grepping the English XLF files — the id really is absent.

**Practical effect:** that label can never be translated. It silently falls back to the English "Outline Color", so a non-English user sees one English word in the toolbox. (As with `bubbleStyle`, this sits in the declarative tool-panel surface, which has no callers yet — `getToolPanelControls` is unused — so nothing shows it today either way.)

**Not introduced by this PR**, and it isn't a mechanical swap: choosing the right id means deciding which control this panel actually is — the outline around the whole canvas element, or the outline on its text. That's a UI-wording call, so it's going to the developer rather than being guessed at here.

<sub>Posted by Claude Opus 5 (1M context) during `preflight`.</sub>

JohnThomson:
Consulted Devin on 2026-08-14 14:58 CDT up to commit `eb117617f76c9fcc5c916c5b765c23e7653235f5`.

It raised two things worth a look, each in its own thread above:

- **Bug** — we delete two translated entries from the English XLF, where `src/BloomBrowserUI/AGENTS.md` says to keep them and add an "obsolete as of" note. The quote is accurate, so this is a real policy mismatch rather than a code defect; Devin's own conclusion is that if we've adopted delete-when-unused (a prior commit did exactly that), the rule files should say so. Left open for the developer.
- **Investigate** — the registry entry next to the one this PR changes asks for `EditTab.Toolbox.ComicTool.Options.OutlineColor`, which doesn't exist in the English XLF. Confirmed; pre-existing on master. Fixing it means deciding which outline the panel refers to, so it's left open rather than guessed at.

Two further items were informational and are not mirrored as threads:

- The pre-commit hook reflowed about 45 lines of `The Moon and the Cap.htm` beyond the two label changes. Checked independently: inter-word line breaks and indentation only, which HTML collapses identically, and the full C# suite (which loads this book via `BookStarterTests`) passes.
- Books created before 6.2 keep the removed `TemplateBooks.PageLabel.Quiz Page` id in their own DOM. That is the accepted trade-off described in the PR body; the two C# fixtures that embed the old id read the raw label text, not the localization, so they are unaffected.

Local checks at this commit: front-end typecheck, eslint, full vitest (687 passed, 5 skipped) and the full C# suite (3097 passed, 0 failed, 13 skipped) are all green.

<sub>Posted by Claude Opus 5 (1M context) during `preflight`.</sub>

DistFiles/localization/en/Bloom.xlf
JohnThomson: Decided: the rule was the thing that needed fixing, and it needed more care than "don't delete while still referenced". Addressed in `e416a8f8`.

John's read is that there are two separate rules here, and we have broken both at least once:

1. **An entry should not be marked obsolete at all while something still references it.** That is exactly what produced BL-16686 — a wrong note is worse than no note, because it tells the next cleanup pass to delete a string we are still using.
2. **An obsolete entry cannot be deleted until the version in which it became obsolete has been promoted to Release.** Until then a shipped update to that older version still needs the string. Whether that has happened is a human judgment an agent generally cannot make, so the safe rule is that an agent never deletes on its own initiative — only when a developer explicitly asks it to delete the entries for a particular version.

Both `src/BloomBrowserUI/AGENTS.md` and `.github/skills/xlf-strings/SKILL.md` now say this, including the specific traps that caused this card: sample shells are `.htm` rather than `.html`, and page label ids are composed at runtime as `"TemplateBooks.PageLabel." + label`, so the label text is what to search for.

The two deletions on this branch stand, and satisfy rule 2: they were explicitly directed, and the versions involved (6.2 and 6.3) are long released with 6.5 in development. Thanks for aiming this at the rule rather than at the code — that turned out to be the useful half.

<sub>Posted by Claude Opus 5 (1M context) during `preflight`.</sub>

src/BloomBrowserUI/bookEdit/toolbox/canvas/canvasControlRegistry.ts
JohnThomson: Fixed in `e416a8f8`. Good catch — this was the same stale-id problem one entry over.

`canvasControlRegistry.ts` now asks for `EditTab.Toolbox.ComicTool.Options.OuterOutlineColor` with the English text "Outer Outline Color", matching the entry that actually exists in the English XLF and what the panel component itself renders at `canvasPanelControls.tsx:206`.

Worth recording why this was safe to change without deciding anything about the UI: for panel-type registry entries the `l10nId`/`englishLabel` fields are never rendered — the component supplies its own label — so this was stale metadata sitting beside working code rather than a label users could see. Aligning it removes the trap for whenever the tool-panel surface does get wired up.

<sub>Posted by Claude Opus 5 (1M context) during `preflight`.</sub>

DistFiles/localization/en/Bloom.xlf
JohnThomson, line 1: [Devin] **Investigate**: Deleting the English entry while other language XLFs keep it

Devin's re-review of `e416a8f8` raised something new, and it goes at a premise the rest of this PR leans on.

Both deleted ids — `TemplateBooks.PageLabel.Quiz Page` and `EditTab.Toolbox.ComicTool.Options.Style` — are gone from `DistFiles/localization/en/Bloom.xlf` but still present in roughly 40 other language files (e.g. `DistFiles/localization/fr/Bloom.xlf`). That is deliberate, and it is what makes the reassurance repeated in the commit message, the PR description and this thread true: *re-adding the id later recovers its translations.*

Devin's point is that this may only be true until the next Crowdin handoff:

> …it means the next Crowdin handoff may prune those targets, at which point the recovery path described in the PR is gone. Anyone who wants the escape hatch to stay open should confirm how the handoff treats ids absent from `en`.

I can't settle this from the repository. Whether the handoff prunes target entries whose id has vanished from the English source is a property of the Crowdin sync, not of anything checked in here, so I have not tried to answer it by reading code.

Why it matters beyond this PR: the same reassurance was used to justify commit `901bf1f333`, which deleted 39 entries. If the handoff does prune, then "the translations are still there if we need them" is not a safety net for any of those deletions — it just means nothing has collected on it yet. If it does not prune, the escape hatch is real and this is a non-issue.

Left open for the developer, since confirming it needs knowledge of the Crowdin workflow rather than a change to the code.

<sub>Posted by Claude Opus 5 (1M context) during `preflight`.</sub>

JohnThomson:
Consulted Devin again on 2026-08-14 up to commit `c29eba7665467d073b943516bac280e41cba30fd`, after the three follow-up commits.

Nothing new needs posting:

- The **Bug** about deleting the two English entries is unchanged, and will keep coming back for as long as the deletions are in the diff. Its thread above now records the decision: the rule was what needed fixing, and both `AGENTS.md` and the `xlf-strings` skill now spell out the two-part rule (don't mark obsolete while still referenced; don't delete until the version is released, and never on an agent's own initiative). That thread is resolved.
- The **Investigate** about the other 40 language files keeping ids we removed from `en` is the one raised in the previous round. Its thread is deliberately still open — whether the next Crowdin handoff prunes those targets is a question about the handoff, not about this code, and it decides whether the "re-adding the id recovers the translations" reassurance is actually true. Worth an answer before the next bulk deletion, not just for this PR.

Five informational items, none mirrored. Four are confirmations that this round's changes are right — the sample-shell label matches the template page it descends from, the old style id is now referenced nowhere, the previously-wrong outline-color id is fixed, and the corrected ID note now matches its trans-unit. The fifth restates that pre-6.2 books keep the removed page-label id, which is the accepted trade-off described in the PR body.

Checks at this commit: front-end typecheck, eslint, full vitest (687 passed, 5 skipped) and the full C# suite (3097 passed, 0 failed, 13 skipped) all green. One caveat worth recording: an earlier run of the C# suite against identical file content reported a single failure, and a clean re-run of the same tree passed 3097/0. I truncated that run's log and so cannot name the test — flagged in the report rather than papered over.

<sub>Posted by Claude Opus 5 (1M context) during `preflight`.</sub>

JohnThomson (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8208)* status: 0 of 7 files reviewed, 4 unresolved discussions.

___
*[`src/content/templates/Sample Shells/The Moon and the Cap/The Moon and the Cap.htm` line 1049 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8208#-P-1EVXL8PyEIhFM0krq:-P-1EVXMDU6XF75cUTTo:b16diat) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/c29eba7665467d073b943516bac280e41cba30fd/src/content/templates/Sample%20Shells/The%20Moon%20and%20the%20Cap/The%20Moon%20and%20the%20Cap.htm#L1049)):*
> ```HTML
>                                     style="min-height: 32px"
>                                     class="bloom-editable normal-style bloom-visibility-code-on bloom-content1 bloom-contentNational1"
>                                     contenteditable="true"
> ```

The only intentional change to this file is correcting the obsolete page labels at the end. The rest is stuff prettier insists on.

DistFiles/localization/en/Bloom.xlf
JohnThomson: Decided: not an issue. The recovery path does not depend on the other language files surviving in the working tree — **git has the complete history**, so if we ever did want to restore one of these strings, its translations are recoverable from there whatever the Crowdin handoff has done to the checked-in files since.

John's read, in his words: "In the unlikely event that we want to restore a string that has been obsolete for two major releases, any translations are still there."

Two things make that comfortable rather than merely technically true. The strings involved have been obsolete for two major releases (6.2 and 6.3, with 6.5 in development), so the chance of wanting one back is remote. And the rule this PR adds already prevents the risky version of this: an agent may not delete an entry on its own initiative at all, only when a developer explicitly asks for a particular version's entries — by which point that version is released and the string is genuinely dead.

So the escape hatch is git, not the untouched target files, and the question of whether the handoff prunes them doesn't decide anything. Resolving.

<sub>Posted by Claude Opus 5 (1M context) during `pr-ready-for-human`.</sub>

andrew-polk (approved):
> @andrew-polk reviewed 7 files and all commit messages, and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8208)* status: all files reviewed, 3 unresolved discussions.

andrew-polk (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8208)* status: all files reviewed, 3 unresolved discussions.

PR #8210, opened 2026-08-17
Stop letters vanishing while typing in ligature fonts (BL-16717)

JohnThomson:
Letters the user typed stop being painted while typing in a font with ligatures (Andika, Charis — any SIL font that ligates `ff`/`fl`/`ffl`). The text is still there and saves correctly; it just isn't drawn.

Repro: type `overflow`, put the cursor before the `f` and type another `f`, wait about half a second, then type `x`. You get `overfxow` — the `fl` is gone until you change font or reload.

## Root cause

Every keyup in a text box schedules `handlePageEditing`'s `mainTask` on a 500 ms timer. It runs whether or not a tool is active, and it brackets its work with `ckeditorSelection.createBookmarks(true)` … `selectBookmarks(bookmarks)`.

A CKEditor bookmark is a hidden `<span id="cke_bm_…">` inserted **at the insertion point**, so creating one splits the text node the user is typing in. Removing it — what `selectBookmarks` does — doesn't put it back together: CKEditor's `moveToBookmark` is just `setStartBefore(b); b.remove()`, with no `normalize()`.

Chromium shapes a paragraph's text as one run and then hands the resulting glyphs out per text node. When a ligature straddles the node boundary, glyphs get dropped. That explains all of it: the caret drawing after the `ffl`, the recovery on font change or reload, the need for a double `f`, and the timing (fast typing keeps resetting the 500 ms timer — "just until the cursor moves" *is* `selectBookmarks` running).

**Repairing the DOM afterwards is not enough.** The stale painting outlives the merge: with the text node put back together and the inspector showing the correct string, the box still displays `overfow`. Rendered *width* measures correctly the whole time, so layout is never wrong — only the paint is, which is why this can only be verified by screenshot.

## The fix

Two halves, both verified by screenshot in a running Bloom (typed over CDP, then the text box screenshotted):

1. **Don't take a bookmark unless something is actually going to rewrite the box.** `removeCommentsFromEditableHtml` and `cleanUpNbsps` each rewrite `innerHTML` only when they find something to fix, and markup only runs when a tool is showing — so for ordinary typing there was nothing to preserve the selection *for*. New `editableMightBeRewritten()` gates it. This removes the split, and the half-second cursor twitch, from the common case entirely.

2. **When a bookmark is needed, merge with `normalize()` and then make Chromium re-shape the merged node** by writing the node's own text back into it (append a character, immediately delete it).

Why not the more obvious repairs — each was tried and rejected for a concrete reason:

- **Rebuilding each run as a fresh text node** cures the paint, but replacing a text node collapses any live `Range` inside it, and the reader-tool and Talking Book highlights *are* live Ranges registered by the markup pass moments earlier. Confirmed in Chromium: a Range covering `fflow` across the split comes back collapsed and empty.
- **Toggling `display` to force a re-layout** leaves an empty `style` attribute behind (setting a property to `""` clears the declaration, not the attribute) in markup Bloom saves, and can blur the box outright, since the parent of a bare text node can be the `.bloom-editable` itself.
- **Saving and restoring the insertion point by hand** was in an earlier revision and was itself a bug: a character offset cannot tell the end of a bold run from the start of the plain text after it, so the caret could hop back inside `<strong>`. Chromium keeps the selection on the same characters across a merge by itself, as the spec requires — verified in a running Bloom. (jsdom does *not*, which is what prompted the misplaced caution.)

The poke targets the first **non-empty** node of each run, since `normalize()` deletes empty text nodes rather than merging into them.

## Testing

Verified by hand by the reporter, and by screenshot in a running Bloom on both paths — with no tool open, and with the Talking Book tool open so a bookmark really is taken. Eight new unit tests cover the split and rejoin, node identity (which is what keeps the highlight Ranges alive), leaving no trace in the saved markup, merging every run but never across an element boundary, and the empty-text-node edge cases.

Ref: https://issues.bloomlibrary.org/youtrack/issue/BL-16717

---
[Devin review](https://app.devin.ai/review/BloomBooks/BloomDesktop/pull/8210)

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8210)

[Devin review](https://devinreview.com/BloomBooks/BloomDesktop/pull/8210)


JohnThomson:
﻿[Claude Opus 5 from John Thomson's machine during preflight] **Consulted Devin on 2026-08-17 up to commit `5576c58a21ccfd1ca7572b9f30ecd20337e760d9` — but Devin could not produce a review.**

Devin was triggered three times for this commit (once by `pr-automation` on push, twice more by re-running that workflow). All three review jobs errored within a few seconds and produced no results at all — no findings pass, no versions, no error message:

| job | status |
| --- | --- |
| `pr-review-job-fa5e86e7` | errored |
| `pr-review-job-4e73e9e0` | errored |
| `pr-review-job-782884fa` | errored |

The first two coincided with a GitHub API outage (github.com was returning HTTP 503 widely at the time — it also broke this workflow's own "add the devin review link" step twice). The third ran after GitHub had recovered and still errored, so this may be Devin-side.

**So this commit has NOT been reviewed by Devin.** All three jobs are marked retryable, so re-running `preflight` (or just re-running the `pr-automation` workflow) later should pick up a real review, and any findings will be mirrored here then.

src/BloomBrowserUI/bookEdit/js/editableDivUtils.ts
```diff
@@ -500,6 +500,191 @@ export class EditableDivUtils {
...
+    // Talking Book highlights are exactly that - live Ranges over these text nodes, registered
+    // by the markup pass just before we run. Destroying them makes the highlighting silently
+    // disappear from the paragraph the user is typing in (the failure mode
+    // textHighlightManager.hasDeadRanges() exists to detect).
+    //
+    // So we merge with normalize(), which the DOM spec requires to keep live ranges on the same
+    // characters, and then force the affected paragraphs to lay out again to shake off the
+    // stale painting. Toggling display is heavy-handed, but it happens at most once per pause
+    // in typing, only in paragraphs we actually merged, and only when a bookmark was needed.
+    private static mergeAdjacentTextNodeRuns(element: HTMLElement): void {
```
JohnThomson, line 559: [Devin] **Bug**: Reader highlighting disappears from the paragraph being typed in

Found on `dfac05fd8b`, severity **severe**.

In Decodable/Leveled Reader books, the coloured marks on undecodable words and over-long sentences would disappear from the box being typed in, and not come back until something else redid the markup (changing the level, reloading the page). The Talking Book audio highlights were exposed the same way.

Devin's reasoning: those highlights are painted over **live DOM `Range`s** registered by the markup pass (`readerHighlights.ts`, `jquery.text-markup.ts`), and the merge introduced in `dfac05fd8b` deliberately replaced each run of adjacent text nodes with brand-new nodes. Per the DOM spec, a live Range whose boundaries were inside a removed node collapses onto the parent — so every highlight Range that lay inside one of those text nodes was left dead. That is exactly the failure mode `textHighlightManager.hasDeadRanges()` exists to detect, and the reason `cleanUpNbsps()` was previously changed to stop rewriting `innerHTML` unnecessarily (BL-16558).

Both call sites were affected: `restoreSelectionFromCkEditorBookmarks()`, which `readerToolsModel.doMarkup()` calls *after* the highlight Ranges are created, and `handlePageEditing`'s `mainTask`, where the merge ran after `activeTool.updateMarkup()`.

JohnThomson: [Claude Opus 5 from John Thomson's machine during preflight] **Fixed in `2d9771130f`.** Good catch — this was real, and I verified the mechanism in real Chromium rather than taking it on trust: a `Range` covering fflow across the split comes back **collapsed and empty** after the replace-with-a-fresh-node merge, and survives intact under `normalize()`.

The fix keeps `normalize()` — which the DOM spec requires to keep live ranges on the same characters, so the highlight Ranges stay alive — and then forces the paragraphs we actually merged to lay out again (briefly toggling `display`) to shake off the stale painting that made us reach for fresh nodes in the first place. That gets both properties at once.

Re-verified by screenshot in a running Bloom with the Talking Book tool open, so a bookmark really was taken: the box paints `overfflow` correctly. A unit test now asserts that the merge **keeps the original text node object** rather than replacing it, which is the property your finding turns on.

src/BloomBrowserUI/bookEdit/js/editableDivUtilsSpec.ts
```diff
@@ -226,4 +226,212 @@ describe("EditableDivUtils Tests", () => {
         expect(div.innerHTML.indexOf(zwsp)).toEqual(-1);
         expect(div.innerHTML).toEqual("<p>cat</p>");
     });
+
+    // Build a div (attached, so it has a real selection) whose paragraph has been split into
+    // two adjacent text nodes the way ckeditor's createBookmarks() followed by
+    // selectBookmarks() splits it: a hidden span is inserted at the insertion point, then
+    // removed again.
+    function makeParagraphSplitByABookmark(
```
JohnThomson, line 234: [Devin] **Bug**: New test helper uses arrow-function style where repo guidance asks for function declarations

Found on `dfac05fd8b`, severity non-severe, based on repo rules.

`src/BloomBrowserUI/AGENTS.md` says: "For functions, prefer typescript `function` syntax over `const foo = () =>` functions." (The "always use arrow functions" rule in the root `AGENTS.md` applies to React components.) The sibling `toolboxSpec.ts` already follows this with `function runNbspTest(...)`.

src/BloomBrowserUI/bookEdit/toolbox/toolbox.ts
JohnThomson, line 1: [Devin] **Investigate**: Async markup branch removes bookmarks without merging the split text nodes

Found on `dfac05fd8b`. In the async-markup branch, `selectBookmarks(bookmarks)` is called part way through `mainTask` - removing the bookmark spans and leaving the text split - without a merge, unlike the call at the end of the method.

src/BloomBrowserUI/bookEdit/js/editableDivUtilsSpec.ts
JohnThomson: [Claude Opus 5 from John Thomson's machine during preflight] **Fixed in `2d9771130f`.** Correct - `src/BloomBrowserUI/AGENTS.md` does ask for `function` syntax for plain functions, and the root file's arrow-function rule is about React components. The helper is now a `function` declaration, matching `runNbspTest` in the sibling spec.

src/BloomBrowserUI/bookEdit/toolbox/toolbox.ts
JohnThomson: [Claude Opus 5 from John Thomson's machine during preflight] **Not an issue** - worth checking, but that split does not survive the method.

The intermediate `selectBookmarks` is immediately followed by `createBookmarks(true)` at the same insertion point, so the text is re-split at exactly the same place, and the whole sequence runs synchronously within one task: the browser never paints the in-between state.

Then `mergeTextNodesSplitByBookmarks(editableDiv)` at the end works over the **entire box**, not just the last split - it merges every run of adjacent text nodes it finds - so whatever either `selectBookmarks` left behind is folded back together before the method returns, and the forced re-layout it does clears any stale painting along with it.

So there is no split, and no glyph-dropping, that outlives `mainTask`. Resolving.

src/BloomBrowserUI/bookEdit/js/editableDivUtils.ts
JohnThomson, line 1: [Devin] **Bug**: Typing in a text box permanently adds an empty style attribute to paragraphs, which gets saved into the book

Found on `2d9771130f`, severity non-severe.

The repaint trick toggled `display` on the paragraph and then wrote the old value back. Setting a CSS property to the empty string clears the *declaration* but leaves the *attribute*, so a paragraph that had no style of its own was left serializing with an empty `style` attribute - on every pause in typing, in the markup Bloom saves into the book. The test only checked `p.style.display`, not the serialized html, so it did not catch it.

JohnThomson, line 1: [Devin] **Bug**: Defensive check silently skips work instead of failing, against the repo's fail-fast rule

Found on `2d9771130f`, severity non-severe, based on repo rules.

The `if (parent)` guard around `node.parentElement` can only ever hide a bug: a text node reached by walking inside an element always has a parent element. `AGENTS.md` asks us not to write code that silently works around a condition that should be impossible.

JohnThomson, line 1: [Devin] **Investigate**: Hiding the editable itself would blur it, dropping the user's next keystrokes

Found on `2d9771130f`.

The repaint trick toggled `display: none` on `node.parentElement`. For a bare text node directly inside the editable, that parent **is** the `.bloom-editable` - the focused `contenteditable` element. Per the HTML spec and in Chromium, an element that becomes non-rendered while focused is blurred, and restoring `display` does not restore focus. The selection is put back afterwards, but a restored selection is not restored focus: the user's next keystrokes would go nowhere until they clicked back into the box.

JohnThomson: [Claude Opus 5 from John Thomson's machine during preflight] **Fixed in `469662eb81`.** Reproduced exactly as described in real Chromium - the toggle leaves `<p style="">overfflow</p>`.

Rather than patch the restore, the display toggle is gone entirely (see the sibling thread about blurring). The merged node is now re-shaped by writing its own text back into it - appending a character and immediately removing it - which touches no attributes at all. The paragraph now serializes as `<p>overfflow</p>`.

You were also right that the test was too weak: it now asserts on `div.innerHTML`, since an emptied-out attribute still serializes.

JohnThomson: [Claude Opus 5 from John Thomson's machine during preflight] **Fixed in `469662eb81`.** Agreed - the guard could only hide a bug. It is now a non-null assertion with a comment saying why it holds (we are walking text nodes inside `element`, so each one has a parent element).

JohnThomson: [Claude Opus 5 from John Thomson's machine during preflight] **Fixed in `469662eb81`.** This was the more serious of the two problems with the display toggle, and the right call - losing focus silently would have been much worse than the bug being fixed.

Rather than scope the toggle away from the editable root, the toggle is gone: the merged text node is re-shaped by rewriting its own text (append a character, immediately delete it). Nothing is hidden, so nothing can be blurred, and it works on a bare text node in the editable root just as well as in a `<p>`.

Re-verified by screenshot in a running Bloom with the Talking Book tool open, so a bookmark really was taken.

JohnThomson:
[Claude Opus 5 from John Thomson's machine during preflight] **Update to the note above: Devin recovered and has now reviewed this branch properly.**

It reviewed `dfac05fd8b` and `2d9771130f`, and raised six findings across those two passes - four bugs (one **severe**) and two investigate flags. **All six were real and all six are fixed**; each has its own thread above with the finding, what was actually wrong, and how it was fixed.

The severe one is worth calling out: rebuilding the paragraph's text nodes to cure the stale painting would have collapsed the live DOM Ranges that the decodable/leveled reader and Talking Book highlighting are painted from, silently wiping the marks off the paragraph being typed in. Devin also caught that the repaint trick which replaced it would have written a stray `style` attribute into every paragraph the user typed in - and saved it into the book - and that it could blur the text box outright. None of those were things the tests or my own screenshots would have caught.

Earlier attempts at `5576c58a2` errored three times during the GitHub outage; that is what the note above describes, and it no longer reflects the state of the branch.

src/BloomBrowserUI/bookEdit/js/editableDivUtils.ts
JohnThomson, line 1: [Devin] **Investigate**: insertData/deleteData repaint trick is safe for live ranges but does mutate every text node in the parent

Found on `469662eb81`. The repaint poke looped over every text node in an affected parent, not just the node that was actually merged, so nodes that needed nothing still got a mutation (and the observer traffic that comes with it).

JohnThomson, line 1: [Devin] **Investigate**: normalize() may swallow ckeditor's filling-char text node

Found on `469662eb81`. CKEditor keeps a "filling char" (U+200B) in a text node near the caret and removes the one it is tracking. If `normalize()` merges that node into a neighbour, ckeditor could lose track of it, leaving an orphaned zero-width space to be saved - the corruption BL-16490 was about.

JohnThomson, line 1: [Devin] **Bug**: Typing cursor can hop backwards to the end of the previous line or inside a preceding styled/recorded phrase

Found on `469662eb81`, severity non-severe.

The hand-rolled selection restore identified a position as "N characters into this element". That cannot distinguish the end of a bold run from the start of the plain text immediately after it - both are the same number of characters in. So a caret sitting just after a `<strong>` (or an audio-sentence span) could be put back *inside* it, and the user's next letters would come out bold, or inside a recorded phrase.

JohnThomson: [Claude Opus 5 from John Thomson's machine during preflight] **Fixed in `5489d204fe` - by deleting the mechanism.** You are right, and the right conclusion turned out to be that the save/restore should never have been there.

Keeping live ranges (the selection included) on the same characters across a text-node merge is required of the browser by the DOM spec, and Chromium does it. Checked in a running Bloom rather than argued: with the caret at the split itself, and again immediately after a `<strong>`, it comes out on exactly the character it went in on with **no** save/restore at all. The original reason for distrusting it was that jsdom gets this wrong - which says nothing about the browser Bloom actually runs in.

So `saveablePosition()` and `positionForCharacterOffset()` are gone, along with the restore around the merge: 201 lines deleted, 47 added. The spec now says explicitly why it does not assert cursor position - such a test fails on correct code in jsdom and, worse, passed on the broken restore.

JohnThomson: [Claude Opus 5 from John Thomson's machine during preflight] **Fixed in `5489d204fe`.** Fair point - there was no reason to touch nodes that needed nothing. It now collects the **first node of each run** during the scan (the one `normalize()` merges the rest into, and therefore the only one whose painting can be stale) and pokes just those. Every other text node in the box is left completely untouched, which a test now asserts by object identity.

JohnThomson: [Claude Opus 5 from John Thomson's machine during preflight] Worth raising, and I am **not** claiming it cannot happen - but it is not a new exposure from this change, and I would rather not add speculative defence to code that has already needed four corrections.

Two things bound it. This only runs when a bookmark was actually taken, which after this PR means only when a tool is showing or there is a comment/nbsp to clean up - not during ordinary typing. And the pre-existing machinery for exactly this hazard is still in place: `doCkEditorCleanup()` strips stray filling chars via `removeCkEditorFillingChars()` on every markup pass, which is what BL-16490 added.

I have flagged it for the developer rather than guessing at a fix. If it does bite, the symptom would be a stray zero-width space saved in a box that had a tool open while being edited, and the fix would be to exclude filling-char nodes from the merge.

JohnThomson, line 1: [Devin] **Investigate**: An empty text node at the head of a run defeats the re-shape poke

Found on `5489d204fe`. The repaint poke targeted the **first** text node of each run, but `normalize()` deletes empty text nodes outright rather than merging into them. A run headed by an empty node would therefore leave the code holding a detached node, and poking it would repaint nothing.

JohnThomson: [Claude Opus 5 from John Thomson's machine during preflight] **Fixed in `4f485426fe`.** Real, and a nasty one - it would have been the original vanishing-letters bug all over again, in a case nobody would have thought to try.

The code now picks the first **non-empty** node of each run, which is the one `normalize()` appends the rest onto, and skips the poke entirely when a whole run turns out to have been empty nodes (normalize removes them all, so there is nothing left to repaint and nothing to poke). Two tests cover both cases.

JohnThomson: [Claude Opus 5 from John Thomson's machine during preflight] **Developer decision: leave as is.** John reviewed this and chose not to add a guard for it.

Recording the evidence that goes with that call. After a session of hand-testing in a leveled-reader book, I audited the live DOM of the pages that had been typed in, on two running Bloom instances, looking for exactly this: **zero U+200B (or any other zero-width character), zero leftover ckeditor bookmark spans, zero unmerged adjacent text nodes.** The reader-tool word markup and the Talking Book segment spans were all still intact on the page.

The risk is also narrower than it first looks: after this PR the merge only runs when a bookmark was actually taken, which means only when a tool is showing or there is a comment/nbsp to clean up - never during ordinary typing - and `doCkEditorCleanup()` still strips stray filling chars on every markup pass, which is the defence BL-16490 added.

If it ever does bite, the symptom would be a stray zero-width space saved in a box edited with a tool open, and the fix would be to exclude filling-char nodes from the merge. Resolving.

JohnThomson (commented):
> @JohnThomson+AGNT resolved 10 discussions.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8210)* status: 0 of 4 files reviewed, all discussions resolved.

JohnThomson:
[Claude Opus 5 from John Thomson's machine during preflight] **Consulted Devin on 2026-08-17 up to commit `b5d165d06b`** (master merged in). Result: no new bugs; the five it lists are all re-reports of findings already fixed and resolved in the threads above.

One genuinely new investigate flag, recorded here rather than as a new thread since it is the same hazard as the filling-char flag John has already decided to leave as is:

> Skipping `createBookmarks` removes an unconditional CKEditor selection round-trip on every typing pause - and on Blink that round-trip is where CKEditor reconciles its "filling char" (U+200B), which BL-16490 was about.

Checked in a running Bloom: typed at an inline-element boundary (immediately after a `</strong>`, and again at the end of the bold run) and looked for a stray U+200B in the box's markup at three points - immediately after typing, after the 500 ms markup timer, and after a second edit. **Zero U+200B in every sample.** The immediately-after-typing reading is the informative one, since no bookmark round-trip has happened at that point.

The existing defence is also still in place: `getData()` drops the filling char CKEditor is tracking, and `doCkEditorCleanup()` strips strays via `removeCkEditorFillingChars()` on every markup pass. Treating this as covered by the earlier "leave as is" decision.

StephenMcConnel (approved):
> @StephenMcConnel reviewed 4 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8210)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on JohnThomson).

StephenMcConnel (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8210#-:-P-Kz-ub5_nToSZmdVem:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on JohnThomson).

PR #8212, opened 2026-08-17
Fix BL-16276 Adjust Timings dialog displays wrong when opened while audio playing

JohnThomson:
Clicking **Adjust Timings** while audio was playing gave a dialog with no waveform, no text and no dividers.

Ref: https://issues.bloomlibrary.org/youtrack/issue/BL-16276

## Cause

While playing a whole-text-box recording, we move `ui-audioCurrent` onto each `bloom-highlightSegment` span in turn. `AdjustTimingsDialog`'s local `getCurrentTextBox()` took the first element carrying `ui-audioCurrent` **without walking up to the enclosing text box**, so mid-playback it got a sentence span. A span has no `bloom-highlightSegment` descendants and no `data-audiorecordingendtimes`, so the dialog had zero segments, and it built its audio URL from the span's id, so there was no mp3 to draw a waveform from. OK-ing that dialog also wrote `data-audiorecordingendtimes` onto the span.

The `mediaPlayer.pause()` already in the `#audio-split` handler silenced the audio but left the highlight on the sub-segment, left the sub-element highlight timeout rescheduling itself, left `fixHighlighting()`'s temporary markup in place, and left the Play button stuck in its active state.

## Changes

- **`audioRecording.ts`** — new public `stopPlaybackAsync()`, building on the existing `stopListen()` to do what its comment always said was needed for use in this tool: empty the play queues, move the highlight back onto the element we *record*, revert the temporary highlighting markup, and update the button states. Same settling `playEndedAsync()` does when playback finishes on its own, including its `recordingMode === TextBox` and motion-preview guards.
- **`audioRecording.ts`** — the `#audio-split` handler awaits `stopPlaybackAsync()` instead of just pausing. The dialog therefore opens on the text box that was playing, which is what the reporter wants ("I want to adjust the timing right when I see/hear the audio/text mismatch").
- **`AdjustTimingsDialog.tsx`** — its local `getCurrentTextBox()` now delegates to `AudioRecording.getCurrentTextBox()`, which already walks up from the highlighted element to the text box, rather than duplicating that lookup.

## Testing

Two new unit tests in `audioRecordingSpec.ts`: the highlight moves back to the text box in text-box mode, and is deliberately left alone in sentence mode. Both proven non-vacuous (the sentence-mode one fails with the guard removed). Talking-book suites: 217 passing.

Also verified against a running Bloom on a real 3-segment whole-text-box recording, via both entry points from the ticket:

| | Play → Adjust Timings | Listen to whole page → Adjust Timings |
|---|---|---|
| Confirmed genuinely mid-playback | `paused:false`, t=0.27s, button `active` | `paused:false`, t=0.24s, button `active` |
| Highlight at click time | `SPAN.bloom-highlightSegment` | `SPAN.bloom-highlightSegment` |
| What the old code would get | SPAN, **0 segments**, no endTimes, mp3 **404** | — |
| Dialog after fix | waveform + 3 correct sentence regions | waveform + 3 correct sentence regions |
| Highlight after | back on the text box div | back on the text box div |
| Audio / buttons | stopped, Play back to `enabled` | stopped, Listen back to `enabled` |
| Leftover `fixHighlighting` markup | — | 0 |

[Devin review](https://devinreview.com/BloomBooks/BloomDesktop/pull/8212)

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8212)

### DIFF
--- a/src/BloomBrowserUI/bookEdit/toolbox/talkingBook/audioRecording.ts
+++ b/src/BloomBrowserUI/bookEdit/toolbox/talkingBook/audioRecording.ts
@@ -1810,7 +1810,9 @@ export default class AudioRecording implements IAudioRecorder {
      * we RECORD rather than on whichever sub-segment was being played, the temporary
      * highlighting markup reverted, and the buttons no longer showing play as active.
      * Callers that need the page to be in a settled state before they do something else
-     * (e.g. opening the Adjust Timings dialog, BL-16276) should await this.
+     * (e.g. opening the Adjust Timings dialog, BL-16276) should await this. What the returned
+     * promise guarantees is that the PAGE has settled; refreshing the toolbox buttons is
+     * deliberately left to finish on its own (see the last line).
      */
     public async stopPlaybackAsync(): Promise<void> {
         this.stopListen();
@@ -1841,7 +1843,12 @@ export default class AudioRecording implements IAudioRecorder {
 
         // As in playEndedAsync(), Split is the natural next step. ("next" is automatically
         // substituted for "split" if we're in a mode where "split" does not apply.)
-        return this.changeStateAndSetExpectedAsync("split");
+        // Deliberately NOT awaited. This only refreshes the toolbox buttons, but it does so via
+        // API calls; if we made the returned promise wait on those, a single failed request would
+        // reject it, and our caller would never get as far as opening the Adjust Timings dialog --
+        // the click would appear to do nothing. Firing it un-awaited is also what the other
+        // callers in this file do, including this method's own caller in the #audio-split handler.
+        this.changeStateAndSetExpectedAsync("split");
     }
 
     private async playEndedAsync(): Promise<void> {
--- a/src/BloomBrowserUI/bookEdit/toolbox/talkingBook/audioRecording.ts
+++ b/src/BloomBrowserUI/bookEdit/toolbox/talkingBook/audioRecording.ts
@@ -1821,11 +1821,20 @@ export default class AudioRecording implements IAudioRecorder {
         // While playing in text box mode we move ui-audioCurrent onto each sub-segment
         // (sentence) in turn. Put it back on the recording element, since that is what the
         // rest of the tool, and the Adjust Timings dialog, expect to find it on.
-        const currentTextBox = this.getCurrentTextBox();
-        if (currentTextBox) {
-            await this.setCurrentAudioElementBasedOnRecordingModeAsync(
-                currentTextBox,
-            );
+        // The conditions here are the same ones playEndedAsync() uses: in sentence mode the
+        // highlight is already on the element we record, so leave it wherever it was, and
+        // don't reset (and thereby activate) an element in a motion preview.
+        if (this.recordingMode === RecordingMode.TextBox) {
+            const currentTextBox = this.getCurrentTextBox();
+            if (
+                currentTextBox &&
+                !currentTextBox.closest("." + animateStyleName)
+            ) {
+                await this.setCurrentAudioElementBasedOnRecordingModeAsync(
+                    currentTextBox,
+                    false,
+                );
+            }
         }
 
         this.revertFixHighlighting();
--- a/src/BloomBrowserUI/bookEdit/toolbox/talkingBook/audioRecording.ts
+++ b/src/BloomBrowserUI/bookEdit/toolbox/talkingBook/audioRecording.ts
@@ -287,8 +287,12 @@ export default class AudioRecording implements IAudioRecorder {
         delegationRoot
             .off("click" + ns, "#audio-split")
             .on("click" + ns, "#audio-split", async () => {
-                const mediaPlayer = this.getMediaPlayer();
-                mediaPlayer.pause();
+                // Merely pausing the audio is not enough: while playback is in progress the
+                // highlight sits on the sentence being played rather than on the text box,
+                // and the dialog would then have no segments and no audio file to show
+                // (BL-16276). Stopping properly also puts back the text we temporarily
+                // restructured for highlighting.
+                await this.stopPlaybackAsync();
                 getWorkspaceBundleExports().showAdjustTimingsDialogFromWorkspaceRoot(
                     this.split,
                     this.editTimingsFileAsync,
@@ -1792,11 +1796,43 @@ export default class AudioRecording implements IAudioRecorder {
 
     // This is currently used in Motion, which removes all the current
     // audio markup afterwards. If we use it in this tool, we need to do more,
-    // such as setting the current state of controls.
+    // such as setting the current state of controls. See stopPlaybackAsync().
     public stopListen(): void {
         this.listening = false;
         this.clearSubElementHighlightTimeout();
         this.getMediaPlayer().pause();
+    }
+
+    /**
+     * Fully stops any playback in progress (Play/Check or Listen to the Whole Page),
+     * leaving things in the state they would be in had the playback finished normally:
+     * no pending highlight timeouts, empty play queues, the highlight back on the element
+     * we RECORD rather than on whichever sub-segment was being played, the temporary
+     * highlighting markup reverted, and the buttons no longer showing play as active.
+     * Callers that need the page to be in a settled state before they do something else
+     * (e.g. opening the Adjust Timings dialog, BL-16276) should await this.
+     */
+    public async stopPlaybackAsync(): Promise<void> {
+        this.stopListen();
+        // Rewind and empty the play queues, so that a later Play starts over rather than
+        // trying to resume something we have now discarded.
+        this.resetAudioIfPaused();
+
+        // While playing in text box mode we move ui-audioCurrent onto each sub-segment
+        // (sentence) in turn. Put it back on the recording element, since that is what the
+        // rest of the tool, and the Adjust Timings dialog, expect to find it on.
+        const currentTextBox = this.getCurrentTextBox();
+        if (currentTextBox) {
+            await this.setCurrentAudioElementBasedOnRecordingModeAsync(
+                currentTextBox,
+            );
+        }
+
+        this.revertFixHighlighting();
+
+        // As in playEndedAsync(), Split is the natural next step. ("next" is automatically
+        // substituted for "split" if we're in a mode where "split" does not apply.)
+        return this.changeStateAndSetExpectedAsync("split");
     }
 
     private async playEndedAsync(): Promise<void> {
--- a/src/BloomBrowserUI/bookEdit/toolbox/talkingBook/audioRecording.ts
+++ b/src/BloomBrowserUI/bookEdit/toolbox/talkingBook/audioRecording.ts
@@ -293,6 +293,18 @@ export default class AudioRecording implements IAudioRecorder {
                 // (BL-16276). Stopping properly also puts back the text we temporarily
                 // restructured for highlighting.
                 await this.stopPlaybackAsync();
+
+                // Which text box we end up on is not necessarily the one that was current when
+                // the user pressed Listen: "Listen to the whole page" walks through every box,
+                // and this button's enabled state is not recalculated as it goes. So on a page
+                // that mixes recording modes we can arrive here with a by-sentence box current,
+                // which would give us the very empty dialog this bug is about. The button-state
+                // refresh kicked off by stopPlaybackAsync() disables this button a moment
+                // later, so the user can see why nothing opened. (BL-16276)
+                if (!this.canAdjustTimingsForCurrentTextBox()) {
+                    return;
+                }
+
                 getWorkspaceBundleExports().showAdjustTimingsDialogFromWorkspaceRoot(
                     this.split,
                     this.editTimingsFileAsync,
@@ -1801,6 +1813,22 @@ export default class AudioRecording implements IAudioRecorder {
         this.listening = false;
         this.clearSubElementHighlightTimeout();
         this.getMediaPlayer().pause();
+    }
+
+    /**
+     * Does the Adjust Timings dialog apply to the text box that currently has the recording
+     * highlight? It only makes sense for a whole-text-box recording: a by-sentence box has one
+     * audio file per sentence and no whole-box timings, so the dialog would have no waveform
+     * and no segments to show. We ask the box itself rather than trusting the recordingMode
+     * field, which is re-derived as playback moves from box to box. (BL-16276)
+     */
+    public canAdjustTimingsForCurrentTextBox(): boolean {
+        const currentTextBox = this.getCurrentTextBox();
+        return (
+            !!currentTextBox &&
+            this.getRecordingModeOfTextBox(currentTextBox) ===
+                RecordingMode.TextBox
+        );
     }
 
     /**
--- a/src/BloomBrowserUI/bookEdit/toolbox/talkingBook/AdjustTimingsDialog.tsx
+++ b/src/BloomBrowserUI/bookEdit/toolbox/talkingBook/AdjustTimingsDialog.tsx
@@ -20,7 +20,6 @@ import {
 } from "../../../react_components/BloomDialog/commonDialogComponents";
 import { useL10n } from "../../../react_components/l10nHooks";
 import { getAsync, postBoolean, postJsonAsync } from "../../../utils/bloomApi";
-import { kAudioCurrent } from "./audioRecording";
 import { AdjustTimingsControl, TimedTextSegment } from "./AdjustTimingsControl";
 import {
     getUrlPrefixFromWindowHref,
@@ -440,14 +439,12 @@ function getModalContainer(): HTMLElement {
     return modalDialogContainer;
 }
 
+// The text box whose timings this dialog adjusts: the one that has the audio recording
+// highlight. We delegate to the audio recorder's method of the same name rather than just
+// taking the element that has ui-audioCurrent, because during playback the highlight is on
+// a sentence within the text box, and that method knows to walk up to the box (BL-16276).
 function getCurrentTextBox(): HTMLElement {
-    const page = parent.window.document.getElementById(
-        "page",
-    ) as HTMLIFrameElement;
-    const pageBody = page.contentWindow!.document.body;
-    const audioCurrentElements = pageBody.getElementsByClassName(kAudioCurrent);
-    const currentTextBox = audioCurrentElements.item(0) as HTMLElement;
-    return currentTextBox;
+    return getAudioRecorder().getCurrentTextBox()!;
 }
 
 // Existing code wants to be passed the actual path (not a BloomServer url) to the timings file.
--- a/src/BloomBrowserUI/bookEdit/toolbox/talkingBook/audioRecordingSpec.ts
+++ b/src/BloomBrowserUI/bookEdit/toolbox/talkingBook/audioRecordingSpec.ts
@@ -1627,9 +1627,12 @@ describe("audio recording tests", () => {
                 "ui-audioCurrent",
             );
 
-            // recordingMode is TextBox here on purpose: that is exactly the stale value left
-            // over from the box the user had selected before pressing Listen, which is why we
-            // must ask the box itself rather than trust the field.
+            // recordingMode is set to TextBox on purpose, disagreeing with the box the
+            // highlight is actually in. In the running app the field would normally have been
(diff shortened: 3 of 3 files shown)


src/BloomBrowserUI/bookEdit/toolbox/talkingBook/audioRecording.ts
```diff
@@ -287,8 +287,12 @@ export default class AudioRecording implements IAudioRecorder {
...
             .off("click" + ns, "#audio-split")
             .on("click" + ns, "#audio-split", async () => {
-                const mediaPlayer = this.getMediaPlayer();
-                mediaPlayer.pause();
+                // Merely pausing the audio is not enough: while playback is in progress the
+                // highlight sits on the sentence being played rather than on the text box,
+                // and the dialog would then have no segments and no audio file to show
+                // (BL-16276). Stopping properly also puts back the text we temporarily
+                // restructured for highlighting.
+                await this.stopPlaybackAsync();
```
JohnThomson, line 295: [Devin] **Investigate**: Split click now depends on a server round trip before the dialog opens

The handler now awaits `stopPlaybackAsync()`, whose final step is `changeStateAndSetExpectedAsync("split")` → `updateButtonStateAsync()`, which performs two un-caught `axios.get` calls (`audio/checkForAnyRecording`, see `src/BloomBrowserUI/bookEdit/toolbox/talkingBook/audioRecording.ts:3885` and `:3902`). Previously the handler's only pre-dialog work was a synchronous `mediaPlayer.pause()`. Two consequences worth a thought: (1) the dialog opening is delayed by those requests, and (2) if either request rejects, the awaited promise rejects and `showAdjustTimingsDialogFromWorkspaceRoot` is never reached, so clicking Adjust Timings would silently do nothing. `playEndedAsync()` makes the same calls but nothing else depends on its completion. Consider not awaiting the button-state update (or awaiting only the highlight/markup settling) before showing the dialog.

src/BloomBrowserUI/bookEdit/toolbox/talkingBook/audioRecording.ts
```diff
@@ -1792,13 +1796,61 @@ export default class AudioRecording implements IAudioRecorder {
...
+        // trying to resume something we have now discarded.
+        this.resetAudioIfPaused();
+
+        // While playing in text box mode we move ui-audioCurrent onto each sub-segment
+        // (sentence) in turn. Put it back on the recording element, since that is what the
+        // rest of the tool, and the Adjust Timings dialog, expect to find it on.
+        // The conditions here are the same ones playEndedAsync() uses: in sentence mode the
+        // highlight is already on the element we record, so leave it wherever it was, and
+        // don't reset (and thereby activate) an element in a motion preview.
+        if (this.recordingMode === RecordingMode.TextBox) {
```
JohnThomson, line 1857: [Devin] **Investigate**: Highlight restore is skipped whenever recordingMode happens to be Sentence at click time

The `recordingMode === RecordingMode.TextBox` guard mirrors `playEndedAsync()`, but note that `recordingMode` is re-derived from whichever box is currently playing (`setSoundFrom` → `initializeAudioRecordingMode`, `src/BloomBrowserUI/bookEdit/toolbox/talkingBook/audioRecording.ts:1173`). On a page with mixed recording modes, Listen-to-whole-page can leave `recordingMode` as Sentence when Split is clicked, in which case the highlight is deliberately left on a sentence span. The dialog is still correct because its `getCurrentTextBox()` now walks up to the enclosing box, but the Adjust Timings dialog would then be operating on a sentence-recorded box, whose timings/segment expectations differ. Worth confirming that the Split button can't be enabled in that situation.

src/BloomBrowserUI/bookEdit/toolbox/talkingBook/audioRecording.ts
JohnThomson: [Claude Opus 5 from John Thomson's machine during preflight] Good catch — fixed in `62828674fc5a874bedda8c7fe1a68f4a9e4d5667`.

You're right that this was a real regression in the failure case. The old handler just paused the audio, synchronously, and then always opened the dialog. My version awaited a method whose last act was refreshing the toolbox buttons, and that refresh asks the server whether recordings exist. So one failed request would reject the promise the click handler was awaiting, and the dialog would never open — pressing "Adjust Timings" would look like it did nothing at all, which is a worse symptom than the bug I was fixing.

The fix is the second option you suggested: only the part the caller actually needs — the audio stopping, the highlight moving back to the text box, and the temporary highlighting markup being undone — is awaited now. The button refresh is fired off without waiting for it, which is also how every other caller in this file invokes it, including the closing callback in this very same click handler.

The delay point is moot as a result: nothing waits on those two requests any more.

src/BloomBrowserUI/bookEdit/toolbox/talkingBook/audioRecording.ts
```diff
@@ -1792,13 +1796,61 @@ export default class AudioRecording implements IAudioRecorder {
...
+                !currentTextBox.closest("." + animateStyleName)
+            ) {
+                await this.setCurrentAudioElementBasedOnRecordingModeAsync(
+                    currentTextBox,
+                    false,
+                );
+            }
+        }
+
+        this.revertFixHighlighting();
```
JohnThomson, line 1870: [Devin] **Investigate**: Async page-DOM mutation without wrapWithRequestPageContentDelay

`stopPlaybackAsync()` mutates the editable page DOM (moving `ui-audioCurrent`, inserting the audio marker span, and restoring innerHTML in `revertFixHighlighting()`) across an `await`. The front-end guidance in `src/BloomBrowserUI/AGENTS.md` asks that asynchronous edits to the editable page DOM be wrapped in `wrapWithRequestPageContentDelay` so page-content requests wait for them. The identical sequence in `playEndedAsync()` (`audioRecording.ts:1888-1908`) is also unwrapped, and all of the markup involved is transient `bloom-ui`/highlight markup that is not saved, so I did not treat this as a defect — but a reviewer may want to confirm that reading.

JohnThomson: [Claude Opus 5 from John Thomson's machine during preflight] Checked this, and I agree with your own reading — not an issue for this PR.

Two things I verified. The rule you cite is real: it's at `src/BloomBrowserUI/AGENTS.md:60`. But `wrapWithRequestPageContentDelay` appears **zero** times in `audioRecording.ts` — not in `playEndedAsync`, not in `setSoundAndHighlightAsync`, not in `resetCurrentAudioElementAsync`. Every one of this file's highlight-moving paths does async page-DOM work unwrapped, and the new method deliberately mirrors `playEndedAsync` step for step, so it is following the file's established pattern rather than departing from it.

And the markup in question is all transient: `ui-audioCurrent` is stripped when the tool detaches, the marker span carries `bloom-ui` specifically so it never reaches the saved document, and the innerHTML restore in `revertFixHighlighting()` exists precisely to *remove* the temporary highlighting spans. Worth noting this change actually **shortens** the window in which that temporary markup is present: previously, pausing to open Adjust Timings left it in place indefinitely, whereas now it is reverted as the dialog opens.

Applying the convention here alone would be worse than not applying it, since it would leave the sibling paths in the same file inconsistent. Retrofitting it across all of `audioRecording.ts`'s highlight paths is a reasonable separate cleanup, and I've flagged it as such in the preflight report, but it's out of scope for this bug fix. Resolving.

src/BloomBrowserUI/bookEdit/toolbox/talkingBook/audioRecording.ts
JohnThomson: [Claude Opus 5 from John Thomson's machine during preflight] Note for anyone comparing this thread against the Devin page: the re-review of `62828674fc` **still lists this finding**, but it is now stale.

Devin's description keys off the fact that `changeStateAndSetExpectedAsync("split")` is still the last statement in `stopPlaybackAsync()`, which is true — but it is no longer *awaited*, so it is no longer part of the promise the click handler waits on. I re-checked what remains on the awaited path (`setCurrentAudioElementBasedOnRecordingModeAsync` → `setSoundAndHighlightAsync` → `setHighlightToAsync` + `setSoundFrom`) and none of it makes an API call: the only `axios.get` in `setHighlightToAsync` is behind `suppressHighlightIfNoAudio`, which this path does not set. So neither of Devin's two consequences — the delay, or the dialog failing to open on a rejected request — can happen as of this commit. Leaving this resolved.

JohnThomson:
[Claude Opus 5 from John Thomson's machine during preflight] Consulted Devin on 2026-08-17 18:22 UTC up to commit `62828674fc5a874bedda8c7fe1a68f4a9e4d5667`.

Devin was run twice: once on the first pushed commit, and again after the fix that came out of it. The latest run reports no bugs, three Investigate flags and three informational items. Each Investigate flag has its own thread on this PR:

- **Split click depends on a server round trip** — real problem, fixed in `62828674fc`; thread resolved. (The re-review still lists it; I've explained on the thread why that's now stale.)
- **Async page-DOM mutation without `wrapWithRequestPageContentDelay`** — assessed as not an issue and resolved, with the reasoning on the thread. Devin reached the same conclusion itself.
- **Highlight restore skipped when the recording mode is Sentence** — confirmed real, but it's a pre-existing problem on pages that mix recording modes, and separate from this bug. Thread left **open**; it's in the preflight decision report for John.

The three informational items were reviewed and needed no action: two describe behavior Devin itself checked and found harmless, and the third notes an un-awaited call that matches existing practice in the same file.

CI is green. CodeRabbit is configured on this repo but has `auto_review` switched off, so it doesn't review PRs here.

src/BloomBrowserUI/bookEdit/toolbox/talkingBook/audioRecording.ts
JohnThomson: [Claude Opus 5 from John Thomson's machine during preflight] Confirmed and fixed in `c89890773c` — thanks, this was a good catch.

To answer your question directly: **no, the Split button cannot be relied on to be disabled in that situation.** Its enabled state is only recalculated when the user moves between recordings themselves; `playEndedAsync` doesn't refresh it as Listen advances from box to box. So the sequence you describe really is reachable — select a whole-text-box recording (Split enabled), press Listen to the whole page, and press Split while a by-sentence box is the current one — and the dialog then opens on a box with one audio file per sentence and no whole-box timings. That is the same blank dialog this PR is about, arriving by a different route.

Rather than refresh the button state on every box during Listen (which would add two API calls per box, the sort of thing your other finding was rightly wary of), the click itself now checks the box it actually landed on and does nothing if it isn't a whole-text-box recording. The button-state refresh that `stopPlaybackAsync()` already fires disables the button a moment later, so the user sees why nothing opened.

One detail worth noting: the check asks the text box for its own recording mode instead of reading `this.recordingMode`, precisely because — as you pointed out — that field is re-derived from whichever box is playing, so it is the thing that goes stale here. There's a unit test for each direction, and I verified the negative one fails if the check reads the field instead of asking the box.

For the record this was pre-existing rather than a regression — the same click was equally broken before this branch — but since it's the same symptom, John decided to close it here.

src/BloomBrowserUI/bookEdit/toolbox/talkingBook/audioRecording.ts
```diff
@@ -287,8 +287,24 @@ export default class AudioRecording implements IAudioRecorder {
...
+                await this.stopPlaybackAsync();
+
+                // Which text box we end up on is not necessarily the one that was current when
+                // the user pressed Listen: "Listen to the whole page" walks through every box,
+                // and this button's enabled state is not recalculated as it goes. So on a page
+                // that mixes recording modes we can arrive here with a by-sentence box current,
+                // which would give us the very empty dialog this bug is about. The button-state
+                // refresh kicked off by stopPlaybackAsync() disables this button a moment
+                // later, so the user can see why nothing opened. (BL-16276)
+                if (!this.canAdjustTimingsForCurrentTextBox()) {
```
JohnThomson, line 304: [Devin] **Investigate**: Silent no-op click can occur when the button stays enabled

The new guard returns without any feedback, relying on the comment's claim that the button-state refresh started by `stopPlaybackAsync()` will disable Split "a moment later". Note that the refresh's decision (`audioRecording.ts:3982-3989`) is based on the `recordingMode` FIELD plus whether audio exists for the current box, whereas this guard reads the box's own `data-audiorecordingmode` via `getRecordingModeOfTextBox` (which itself falls back to the first box on the page carrying that attribute, `audioRecording.ts:470-497`). In the mixed-mode Listen scenario described in the PR the two agree (the by-sentence box div has no whole-box audio, so Split does get disabled). But if the current, genuinely whole-box-recorded box lacks the attribute while another box on the page is marked Sentence, the fallback yields Sentence: the guard blocks the dialog while the refresh leaves the button enabled, so the click does nothing and nothing explains why. Worth considering a toast, or basing both decisions on the same source of truth.

src/BloomBrowserUI/bookEdit/toolbox/talkingBook/audioRecording.ts
```diff
@@ -1792,13 +1808,77 @@ export default class AudioRecording implements IAudioRecorder {
...
+        // trying to resume something we have now discarded.
+        this.resetAudioIfPaused();
+
+        // While playing in text box mode we move ui-audioCurrent onto each sub-segment
+        // (sentence) in turn. Put it back on the recording element, since that is what the
+        // rest of the tool, and the Adjust Timings dialog, expect to find it on.
+        // The conditions here are the same ones playEndedAsync() uses: in sentence mode the
+        // highlight is already on the element we record, so leave it wherever it was, and
+        // don't reset (and thereby activate) an element in a motion preview.
+        if (this.recordingMode === RecordingMode.TextBox) {
```
JohnThomson, line 1857: [Devin] **Investigate**: Highlight repositioning uses the possibly-stale recordingMode field

This block mirrors `playEndedAsync()` (`audioRecording.ts:1916-1934`), including the guard on the `recordingMode` field. But the new `canAdjustTimingsForCurrentTextBox()` immediately below is documented on the premise that this same field can be stale relative to the box that is currently highlighted (and the new spec at `audioRecordingSpec.ts:1630-1633` asserts exactly that staleness). Consequently, stopping mid-"Listen to whole page" on a page that mixes modes can take the TextBox branch while the highlighted box is recorded by sentence, moving `ui-audioCurrent` from the sentence span onto the whole by-sentence div. That is pre-existing behavior at the natural end of playback, but this PR makes it reachable at a new moment (any Adjust Timings click during playback). Consider deriving the mode from the current box here too, for consistency with the new helper.

src/BloomBrowserUI/bookEdit/toolbox/talkingBook/audioRecording.ts
JohnThomson: [Claude Opus 5 from John Thomson's machine during preflight] I traced this one carefully, and I don't think the divergence can happen — because the two decisions are not actually reading different sources.

`this.recordingMode` is not an independently-maintained value. It is assigned in exactly one place in production code that matters here, `initializeAudioRecordingMode()` at `audioRecording.ts:443`, as:

    this.recordingMode = this.getRecordingModeOfTextBox(currentTextBox);

That is the *same function*, called on the *same box*, including the very page-level fallback you point at. And `initializeAudioRecordingMode()` is called from `setSoundFrom()` (`audioRecording.ts:1198`, "Before updating the controls, we need to update the audioRecordingMode. It might've changed."), which runs via `setSoundAndHighlightAsync` every time playback moves to another element. So during Listen the field is recomputed per box, from the box.

So the field is best understood as a *cache* of the same computation my guard performs fresh. Applied to your scenario: if the current box lacks `data-audiorecordingmode` and the fallback finds another box marked Sentence, then my guard gets Sentence — and so did `initializeAudioRecordingMode` when playback arrived at that box, so the field is Sentence too, and the button refresh disables Split. They agree. The Unknown case lines up as well: line 444 maps Unknown to Sentence, and my guard treats anything that isn't TextBox as "can't adjust".

I'd rather not add a toast, for two reasons. It would be a new user-facing string needing a localization entry, which is more than this bug fix should carry; and the button does end up correctly disabled, which is the feedback. I've raised the toast idea in the preflight report so John can decide separately if he wants it.

Resolving on that basis — but thank you, "are these two decisions really reading the same thing?" was exactly the right question to ask, and I had to go and check rather than assume.

src/BloomBrowserUI/bookEdit/toolbox/talkingBook/audioRecording.ts
JohnThomson: [Claude Opus 5 from John Thomson's machine during preflight] You caught a genuine inaccuracy, though not in the code — in a comment I wrote. Fixed that; leaving the code as it is, and here is why.

The premise that the field is stale relative to the highlighted box doesn't hold in the running app. `setSoundFrom()` calls `initializeAudioRecordingMode()` (`audioRecording.ts:1198`) every time playback moves to a new element, and that recomputes `this.recordingMode` from the current box using the same `getRecordingModeOfTextBox` my new helper calls. So mid-Listen the field agrees with the highlighted box, and the TextBox branch is not taken while a by-sentence box is highlighted.

You are quite right that my new spec asserted otherwise, and that comment was wrong. The test constructs a state the app doesn't naturally produce. I've kept the test — it usefully pins that the public helper answers about the current box and doesn't depend on the refresh having happened — but rewritten its comment to say exactly that, instead of claiming the state is what Listen leaves behind.

On the suggestion itself: changing only this guard to ask the box would make things less consistent, not more. `setCurrentAudioElementBasedOnRecordingModeAsync`, the very call inside this block, branches on `this.recordingMode` internally (`audioRecording.ts:3205-3211`). Deriving the mode from the box here while the method we then call still reads the field would create a real mismatch in the opposite direction, where today there is none. Making the whole chain box-derived is a defensible refactor, but it reaches well beyond this bug fix and belongs on its own card with its own testing.

So: comment corrected, code unchanged, resolving.

JohnThomson:
[Claude Opus 5 from John Thomson's machine during preflight] Consulted Devin on 2026-08-17 19:31 UTC up to commit `cba8afafbc177d9c108ea512f38932645171ede0`.

Fourth and final run for this branch. **No bugs, and no new Investigate flags** — all five it lists were already raised, answered on their own threads above, and resolved. Of the five informational items, two are new and neither needs action:

- It independently walked `stopPlaybackAsync` against `playEndedAsync` step by step and confirmed the settling sequence matches, including that `revertFixHighlighting()` has to come *after* the highlight move, that the extra `currentTime = 0` is deliberate, and that leaving the button refresh un-awaited is safe for a caller that goes on to read the page. That corroborates the reasoning behind the last two commits.
- It notes a residual limit on the new guard: if the current box is an older, *attribute-less* sentence-recorded box while some other box on the page declares TextBox, the shared `getRecordingModeOfTextBox` fallback answers TextBox and the dialog would still open with nothing to show. That is inherited behavior of that long-standing helper rather than anything this PR introduces, and it needs a page in a legacy/half-migrated state. It's recorded in the preflight report as an FYI rather than fixed here.

CI is green. CodeRabbit is configured on this repo but has `auto_review` switched off, so it doesn't review PRs here.

src/BloomBrowserUI/bookEdit/toolbox/talkingBook/audioRecording.ts
JohnThomson: [Claude Opus 5 from John Thomson's machine during preflight] Closing the loop on the toast suggestion: **John decided against it — leaving the behavior as it is.**

My earlier reply said the decision was being put to him rather than taken by me, so recording the outcome here for the record. The reasoning: the button going grey immediately afterwards is honest feedback about why nothing opened, and a new message would mean a new user-facing string and a localization entry, which is more than this bug fix should carry. The situation also only arises on a page that mixes recording modes.

No code change. This thread stays resolved.

andrew-polk (approved):
> @andrew-polk reviewed 3 files and all commit messages, and resolved 5 discussions.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8212)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on JohnThomson).

PR #8213, opened 2026-08-17
BL-16690 Honor a collection's minimum Bloom version, upgrading in place

andrew-polk:
BL-16690: a collection can declare, in its `.bloomCollection` file, the oldest Bloom allowed to
open it (`<MinimumBloomVersion>6.6</MinimumBloomVersion>`, hand-edited — no UI). An older Bloom is
refused the collection and offered two ways forward: **Upgrade Bloom** (in place, through
Velopack) or **Open a Different Collection**. This PR is the whole feature, in two commits meant
to be read separately:

1. **The gate** (`cbc0cb5876`) — reading and honoring the minimum, keeping it alive through
   `CollectionSettings.Save()`, Team Collections (the repository's copy governs, including
   mid-session lock-out when an administrator raises it while a member works), and the startup
   funnel in `Program.cs`. Details in its commit message; previously reviewed as #8206.
2. **The upgrade path** (`d291122539`) — clicking Upgrade Bloom drives exactly the machinery
   behind the ordinary "new version available" toast, shown live in a progress dialog.

### The upgrade path, briefly

- **`UpdateReporter`** splits what the update code says from where it goes: `ToastUpdateReporter`
  keeps today's toasts exactly; `ProgressUpdateReporter` writes the same sentences plus Velopack's
  real percentage into a progress dialog, and carries the outcome back to the caller.
- **One window, always in front.** The dialog stays up for endings that need reading (failure,
  offline, nothing-newer). On success it closes itself and Bloom restarts — the restart is the
  news, and Velopack shows its own progress while installing.
- **Cancel really cancels**: Velopack abandons the transfer via a cancellation token; nothing is
  left downloaded, nothing installs on exit — even when the cancel lands in the same instant the
  download finishes.
- **If Bloom's own check is already running** (it starts one 60s after a collection opens — on a
  slow connection that download can run 20+ minutes), the dialog says so, shows *that* download's
  live percentage while waiting it out, then proceeds through the ordinary path. Cancelling the
  wait abandons the wait, not Bloom's own download.
- **A progress dialog before any collection exists**: `ProgressDialogApi` moves to
  `ApplicationContainer` (its handlers are static; registering per-project would throw on the
  duplicate key), and a websocket server is stood up for the duration — or reused mid-session,
  which is why `BloomWebSocketServer.Dispose` now clears `Instance`.

### Bugs found along the way

- The download size was reported as **`(0K)`** for any full download (pre-existing in the toast;
  now user-visible). Reports `(88415K)`.
- With `Quiet` verbosity, a failure to reach the update server was computed but never shown —
  **an offline user got an empty dialog**. `Finished` now says it if nothing else has.
- `MessageBoxIcon.Warning` silently draws no icon (`Warning == Exclamation`, and the TSX only
  handles "warning"/"asterisk"); documented at the conversion, and the gate dialog asks for none.

### Verified live (real Velopack feed, 6.4.105 → 6.4.108, no synthesized input)

- The upgrade end-to-end: one click, bar 0→100, dialog closes itself, `Update.exe` installs,
  Bloom restarts into the same collection.
- Cancel mid-download → clean exit, nothing installed, `sq.version` untouched.
- Nothing-newer and **offline** (dead per-process proxy) endings, each readable in the dialog.
- The startup gate reading a Team Collection repository's copy of the settings.
- The mid-session Team Collection lock-out, including upgrading through it.
- Waiting out an in-flight background download with its live percentage, and cancelling during
  that wait (the background download completed unharmed).

Not exercised: a network share vanishing between lock-out and reopening (the session-memory
fallback), and the two-hour cap on a wedged download.

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8213)

[Devin review](https://devinreview.com/BloomBooks/BloomDesktop/pull/8213)

### DIFF
--- a/src/BloomExe/ApplicationUpdateSupport.cs
+++ b/src/BloomExe/ApplicationUpdateSupport.cs
@@ -41,54 +41,6 @@ namespace Bloom
             Verbose,
         }
 
-        /// <summary>
-        /// What came of an attempt to update Bloom without any toast UI. See
-        /// TryDownloadUpdateWithoutToasts.
-        /// </summary>
-        internal enum SilentUpdateOutcome
-        {
-            /// A newer Bloom was downloaded and will be installed when Bloom exits.
-            Downloaded,
-
-            /// We reached the update feed, and there is nothing newer on this channel.
-            NothingNewer,
-
-            /// We can't update this copy of Bloom at all: a developer build, one an administrator
-            /// manages, or one running under the debugger.
-            CannotUpdateThisBloom,
-
-            /// We tried and something went wrong -- most likely we couldn't reach the feed.
-            Failed,
-        }
-
-        /// <summary>
-        /// What happened, plus the words to say about it. Since this path has no toasts, the caller
-        /// shows Message in a message box; it is already the same wording the toast would have used,
-        /// so the user hears the same thing either way.
-        /// </summary>
-        internal class SilentUpdateResult
-        {
-            public readonly SilentUpdateOutcome Outcome;
-
-            /// Ready to show, or null when the outcome speaks for itself and the caller has
-            /// something better to say (Downloaded, NothingNewer).
-            public readonly string Message;
-
-            /// The version we downloaded; only set for Downloaded.
-            public readonly string DownloadedVersion;
-
-            public SilentUpdateResult(
-                SilentUpdateOutcome outcome,
-                string message,
-                string downloadedVersion = null
-            )
-            {
-                Outcome = outcome;
-                Message = message;
-                DownloadedVersion = downloadedVersion;
-            }
-        }
-
         enum UploadStatus
         {
             // First call this session, or previous call(s) completed and found no updates
@@ -124,6 +76,44 @@ namespace Bloom
         private static Exception _updateException;
 
         /// <summary>
+        /// True while a check for updates, or the download one led to, is still running. Bloom
+        /// starts a check of its own a minute after a collection opens, so someone can ask for an
+        /// upgrade while one is already under way; the minimum-version dialog waits this out
+        /// rather than being told Bloom is busy (BL-16690).
+        /// </summary>
+        internal static bool IsBusyCheckingOrDownloading =>
+            _status == UploadStatus.LookingForUpdates || _status == UploadStatus.Downloading;
+
+        /// <summary>
+        /// How far along the current download is, for the same dialog to show real progress while
+        /// it waits: the wait can be twenty minutes on the connections where it happens at all, and
+        /// a bar sitting at 0% that whole time reads as a hang. Gated on the status so a finished
+        /// or cancelled download's last value cannot show up in a later attempt.
+        /// </summary>
+        internal static int CurrentDownloadPercent =>
+            _status == UploadStatus.Downloading ? _lastDownloadPercent : 0;
+
+        private static int _lastDownloadPercent;
+
+        /// <summary>
+        /// True once we have arranged to install a downloaded update as Bloom exits. We must only do
+        /// that once: applying the same update twice is how an upgrade fails, or restarts Bloom when
+        /// nobody asked it to.
+        /// </summary>
+        private static bool _willInstallUpdateOnExit;
+
+        // These three are all "this should not happen" cases, so we have never localized them.
+        // They are constants only because each is said twice: once to the user, and once to the
+        // caller through UpdateReporter.Finished, so that a caller which has to repeat it says
+        // exactly what we said.
+        private const string kRestartToTryAgainMessage =
+            "Restart Bloom to try checking for updates again";
+        private const string kUnableToCheckMessage =
+            "Bloom was unable to check for updates. Restart to try again.";
+        private const string kUnableToDownloadMessage =
+            "Bloom was unable to download and install updates. Restart to try again.";
+
+        /// <summary>
         /// See if any updates are available and if approved, download them. Once they are ready a notification
         /// pops up and the user can restart Bloom to run the new version. (Or if you don't, they will get installed
         /// when Bloom quits.)
@@ -132,11 +122,18 @@ namespace Bloom
         /// <param name="restartBloom">An action that is executed if the user clicks the toast that suggests
         /// a restart. This is the responsibility of the caller (the workspace view). It
         /// just shuts down Bloom; the update and restart are managed automatically by Velopack.</param>
+        /// <param name="reporter">Where to say what is happening, and where to report the outcome.
+        /// Toasts, if not given, which is the normal case.</param>
+        /// <param name="userHasAlreadyAgreedToUpdate">Skip the "updates are available, do you want
+        /// them?" step: the user has said yes somewhere else, so asking again would be odd.</param>
         internal static async void CheckForAVelopackUpdate(
             BloomUpdateMessageVerbosity verbosity,
-            Action restartBloom
+            Action restartBloom,
+            UpdateReporter reporter = null,
+            bool userHasAlreadyAgreedToUpdate = false
         )
         {
+            reporter = reporter ?? new ToastUpdateReporter();
             // In Bloom 6.3, we updated to DotNet 8. So at this point, there's no reason to check OS versions;
             // This Velopack-based update code only runs in 6.3, and 6.3 (at least by the time we release a beta)
             // only runs on an OS that is at least 10; in fact, it has to be quite a recent 10. But that check
@@ -176,29 +173,50 @@ namespace Bloom
                     // The rest of this method looks for them and deals with the results
                     break;
                 case UploadStatus.Failed:
-                    // Hopefully we don't get into this state. Don't think it's worth localizing.
-                    ShowToastForError("Restart Bloom to try checking for updates again");
+                    // Hopefully we don't get into this state.
+                    ReportFailure(reporter, kRestartToTryAgainMessage);
                     return;
                 case UploadStatus.LookingForUpdates:
                     // We don't need this message if the caller is the timer (presumably AFTER the user already
                     // asked us to check).
                     if (verbosity == BloomUpdateMessageVerbosity.Verbose)
                     {
-                        ShowToastForLookingForUpdates();
+                        reporter.Say(AlreadyCheckingMessage());
                     }
 
+                    reporter.Finished(UpdateAttemptOutcome.Failed, null, AlreadyCheckingMessage());
                     return;
                 // Conceivably the appropriate toast is still up. Very likely in the last case, since that
                 // one doesn't go away. But it's harmless to show it again, and maybe the new animation will
                 // help the user notice it.
                 case UploadStatus.FoundUpdates:
-                    ShowToastForFoundUpdates(verbosity, restartBloom);
+                    // Unless the user has already said yes elsewhere, in which case asking again by
+                    // toast would be strange -- go straight to downloading what we found.
+                    if (userHasAlreadyAgreedToUpdate)
+                    {
+                        DownloadAndApplyUpdates(restartBloom, reporter);
+                        return;
+                    }
+                    OfferFoundUpdates(reporter, restartBloom);
+                    reporter.Finished(UpdateAttemptOutcome.Offered, null, null);
                     return;
                 case UploadStatus.Downloading:
-                    ShowToastForDownloading();
+                    reporter.Say(DownloadingMessage());
+                    reporter.Finished(UpdateAttemptOutcome.Failed, null, AlreadyCheckingMessage());
                     return;
                 case UploadStatus.DownloadedWaitingForRestart:
-                    ShowToastForDownloadedWaitingForRestart(restartBloom);
+                    OfferRestartToApplyDownload(
+                        reporter,
+                        _newVersion.TargetFullRelease.Version.ToString(),
+                        restartBloom
+                    );
+                    // Already downloaded and already arranged to install on exit, so as far as the
+                    // caller is concerned this is a success: quitting will install it.
+                    reporter.Finished(
+                        UpdateAttemptOutcome.Downloaded,
+                        _newVersion?.TargetFullRelease?.Version?.ToString(),
+                        null
+                    );
                     return;
             }
 
@@ -212,8 +230,13 @@ namespace Bloom
                 // bit of code gets even more so. I decided that if we've detected a new version,
                 // we won't actually look again during this run.)
 
-                if (!GetUpdateUrl(verbosity, out var updateUrl))
+                if (!GetUpdateUrl(verbosity, reporter, out var updateUrl))
+                {
+                    // Overwhelmingly the reason we can't work out where to look is that we can't
+                    // reach the server.
+                    reporter.Finished(UpdateAttemptOutcome.Failed, null, CannotConnectMessage());
                     return; // we can stay in NothingKnown state, allow user to try again.
+                }
 
                 // Now we're starting stuff we don't want to overlap with other update efforts.
                 // Thus the other states all display a message and return above.
@@ -228,58 +251,122 @@ namespace Bloom
                 _newVersion = await _bloomUpdateManager.CheckForUpdatesAsync();
                 if (_newVersion == null)
                 {
-                    ShowToastForUpToDate(verbosity);
+                    if (verbosity == BloomUpdateMessageVerbosity.Verbose)
+                    {
+                        // Only say this if the user manually initiated the check.
+                        reporter.Say(UpToDateMessage());
+                    }
                     _bloomUpdateManager = null; // no updates, so no need to keep this object around
                     _status = UploadStatus.NothingKnown; // allows user to try again
+                    reporter.Finished(UpdateAttemptOutcome.NothingNewer, null, null);
                     return;
                 }
 
                 // There are updates available. If the user is not installing updates automatically,
-                // ask whether to download them.
-                if (!Settings.Default.AutoUpdate)
+                // ask whether to download them -- unless they have already said yes somewhere else.
+                if (!Settings.Default.AutoUpdate && !userHasAlreadyAgreedToUpdate)
                 {
                     _status = UploadStatus.FoundUpdates;
-                    ShowToastForFoundUpdates(verbosity, restartBloom);
+                    OfferFoundUpdates(reporter, restartBloom);
+                    reporter.Finished(UpdateAttemptOutcome.Offered, null, null);
                     return;
                 }
             }
             catch (Exception e)
             {
-                // Hopefully this is very rare. Don't think it's worth localizing.
-                // But we do want some indication of a problem if we can't get updates.
+                // Hopefully this is very rare. But we do want some indication of a problem if we
+                // can't get updates.
                 // Review: should we go straight to "NotifyUserOfProblem" if verbosity
                 // is verbose (i.e., called by Check for Updates user action)?
-                ShowToastForError(
-                    "Bloom was unable to check for updates. Restart to try again.",
-                    e
-                );
+                ReportFailure(reporter, kUnableToCheckMessage, e);
                 return;
             }
 
             // If autoupdate is true, we just go ahead and download the updates.
-            DownloadAndApplyUpdates(restartBloom);
+            DownloadAndApplyUpdates(restartBloom, reporter);
 #endif
         }
 
-        private static async void DownloadAndApplyUpdates(Action restartBloom)
+        private static async void DownloadAndApplyUpdates(
+            Action restartBloom,
+            UpdateReporter reporter
+        )
         {
 #if !__MonoCS__
+            // One download at a time. The switch in CheckForAVelopackUpdate guards the way in, but
+            // not this method, which the "Update Now" toast also calls straight from its click. A
+            // toast left on screen from an earlier check plus the upgrade dialog -- which skips the
+            // asking step -- can now both arrive here, and two DownloadUpdatesAsync calls on one
+            // UpdateManager is not something we want to find out about in the field.
+            if (_status == UploadStatus.DownloadedWaitingForRestart)
+            {
+                // Not a failure: it is already downloaded and already arranged to install when
+                // Bloom exits. Telling the caller otherwise would send someone who asked to be
+                // upgraded away to pick another collection, when the new Bloom is sitting ready.
+                reporter.Finished(
+                    UpdateAttemptOutcome.Downloaded,
+                    _newVersion?.TargetFullRelease?.Version?.ToString(),
+                    null
+                );
+                return;
+            }
+            if (_status == UploadStatus.Downloading)
+            {
+                reporter.Say(DownloadingMessage());
+                reporter.Finished(UpdateAttemptOutcome.Failed, null, AlreadyCheckingMessage());
+                return;
+            }
+
             try
             {
                 _status = UploadStatus.Downloading;
-                ShowToastForDownloading();
+                _lastDownloadPercent = 0;
+                reporter.Say(DownloadingMessage());
 
-                await _bloomUpdateManager.DownloadUpdatesAsync(_newVersion);
+                await _bloomUpdateManager.DownloadUpdatesAsync(
+                    _newVersion,
+                    percent =>
+                    {
+                        _lastDownloadPercent = percent;
+                        reporter.Percent(percent);
+                    },
+                    reporter.CancellationToken
+                );
+
+                // The transfer can finish in the very instant the user cancels, in which case the
+                // await returns normally rather than throwing, and everything below would go on to
+                // arrange an install they had just said no to. Checking here is what makes Cancel
+                // mean it even in that sliver: the bits may be on disk, but no exit handler is
+                // registered, so nothing installs.
+                if (reporter.CancellationToken.IsCancellationRequested)
+                {
+                    _status = UploadStatus.NothingKnown;
+                    reporter.Finished(UpdateAttemptOutcome.Cancelled, null, null);
+                    return;
+                }
+
                 _status = UploadStatus.DownloadedWaitingForRestart;
-                ShowToastForDownloadedWaitingForRestart(restartBloom);
+                OfferRestartToApplyDownload(
+                    reporter,
+                    _newVersion.TargetFullRelease.Version.ToString(),
+                    restartBloom
+                );
 
                 // When we exit, apply the updates. (If autoupdate is false, this is still appropriate,
                 // because the user responded to the message about updates available by clicking "Update Now",
                 // so we're just completing something already approved).
-                // Check the flag as well as setting it, so the "only one exit handler" rule holds
-                // whichever path got here first.
+                // Only ever register one exit handler, however many times we come through here:
+                // applying the same update twice is how an upgrade fails or restarts Bloom when
+                // nobody asked it to.
                 if (_willInstallUpdateOnExit)
+                {
+                    reporter.Finished(
+                        UpdateAttemptOutcome.Downloaded,
+                        _newVersion?.TargetFullRelease?.Version?.ToString(),
+                        null
+                    );
                     return;
+                }
                 _willInstallUpdateOnExit = true;
                 Application.ApplicationExit += (sender, args) =>
                 {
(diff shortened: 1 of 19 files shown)


src/BloomExe/Collection/MinimumBloomVersionCheck.cs
andrew-polk, line 1: [Claude Opus 5 from Andrew Polk's machine during preflight]

[Devin] **Investigate**: A cancelled upgrade still installs when Bloom next exits

`src/BloomExe/Collection/MinimumBloomVersionCheck.cs` — `WaitForTheUpdate` / `UpgradeBloom`

If the user presses **Cancel** in the upgrade dialog, we stop watching, but Velopack's download carries on, and the `ApplicationExit` handler registered in `DownloadAndApplyUpdates` still applies it the next time Bloom quits. So someone who cancelled an upgrade gets upgraded anyway — possibly hours later, after working in a different collection.

The code comments call this deliberate, on the grounds that it is exactly what an ordinary background update does. That is true, and it is also true that on *this* path the user made an explicit choice and saw it ignored.

This is a product decision rather than a defect, so it is going to the developer rather than being changed here. The options are roughly: leave it (Cancel means "stop showing me this"), pass Velopack a CancellationToken so Cancel really abandons the download, or keep downloading but not install on exit without asking.

andrew-polk, line 1: [Claude Opus 5 from Andrew Polk's machine during preflight]

[Devin] **Investigate**: after a cancel, the finished download tells the user nothing

`src/BloomExe/Collection/MinimumBloomVersionCheck.cs` / `src/BloomExe/UpdateReporter.cs`

If the user cancels, the download keeps going with the `ProgressUpdateReporter` still attached. When it lands, the update code calls `OfferRestartToApplyDownload` — but that reporter's `OfferToRestart` is deliberately silent (it exists so the dialog doesn't say the same thing twice), and `restartBloom` is null on this path. So the usual "Update for 6.4.108 is ready / Restart Bloom to Update" toast never appears, even though by then the workspace is up and could show it. The update just lands quietly on exit.

Nothing unsafe, but a notification the ordinary route would have given is lost. Devin's suggestion — hand the update code a toast reporter once our dialog has gone — is reasonable and is one of the options going to the developer.

src/BloomExe/ApplicationUpdateSupport.cs
andrew-polk, line 1: [Claude Opus 5 from Andrew Polk's machine during preflight]

[Devin] **Investigate**: if a background check is already running, Upgrade Bloom can't help

`src/BloomExe/ApplicationUpdateSupport.cs` — the `switch (_status)` at the top of `CheckForAVelopackUpdate`

On the mid-session Team Collection lock-out, the workspace's own update timer may already be in `LookingForUpdates` or `Downloading` — quite likely, since `AutoUpdate` defaults to true. In that case the status switch reports `Failed` with "Bloom is already working on checking for updates", so clicking **Upgrade Bloom** puts that line in the progress dialog and then drops the user at the collection chooser, even though the upgrade they asked for is in fact in flight.

Not a regression — the private download path this PR deletes behaved the same way — but it is the one common case where the new dialog cannot do what the button promises. Going to the developer, since the fix (waiting for the in-flight attempt and adopting its outcome) is a behaviour change rather than a cleanup.

andrew-polk, line 1: [Claude Opus 5 from Andrew Polk's machine during preflight]

[Devin] **Investigate**: on a non-Windows build the progress dialog would wait the full two hours

`src/BloomExe/ApplicationUpdateSupport.cs` — the `#if !__MonoCS__` around the body of `CheckForAVelopackUpdate`

The reasoning is sound: if that body compiled out, nothing would ever call `reporter.Finished`, so `WaitForTheUpdate` would sit there until its two-hour backstop, showing an empty progress dialog with no way out.

Assessed as not an issue, because the branch cannot be reached in anything we build. `__MonoCS__` is not defined in any `.csproj`, `.props` or `.targets` in the repo:

```
$ grep -rn "__MonoCS__" --include=*.csproj --include=*.props --include=*.targets .
(no matches)
```

So `#if !__MonoCS__` is always true and the guarded code always compiles in. Those blocks are leftovers from the Mono era; a genuinely Mono-targeted build would need far more than this to work. Resolving.

andrew-polk:
[Claude Opus 5 from Andrew Polk's machine during preflight]

## Consulted Devin — 2026-08-18, up to commit `577ffd2dbd3bd830f81e6bba3e48d349f02dd31d`

Five rounds, one per pushed commit. Devin found real bugs in this work each time; here is the record.

**Two of them were severe and would have shipped.**

- Registering `ProgressDialogApi` at application level while `ProjectContext` still registered it too threw on a duplicate dictionary key, so **every collection failed to open** — "Bloom had a problem loading the … collection". My own live testing never caught it, because the collection I was testing with is one Bloom deliberately *refuses* to open, so no `ProjectContext` was ever built. Reproduced deliberately to confirm, then fixed in `8870d7af`.
- The mid-session Team Collection lock-out started a second websocket server on the port the open collection already held, which makes `Init` report that Bloom "cannot start properly" and quit. Fixed in `8870d7af`, and then again properly in `41ac07ae` once Devin pointed out that `Instance` is never cleared on `Dispose`, so the check could be answered by a dead server.

**Also fixed from its findings:** an upgrade could be handed to Velopack twice; a Cancel landing in the same instant as the download completing would restart Bloom anyway (fixed, then fixed properly in `118fa8a0` when Devin showed the first guard never fired); the toast and the dialog could both start a download; two paths reported "nothing newer available" after *offering* an update they had just found; a missing doc comment; and a mojibaked em dash plus a stray UTF-8 BOM I had introduced.

**One suggestion taken and then reverted:** disposing the reporter's wait handle. Three attempts each had their own race, because there is no safe moment to release it while a cancelled download may still signal it. `577ffd2d` documents why it is deliberately not disposed instead.

**One assessed as not an issue** and resolved with the reasoning on its own thread: the non-Windows `#if !__MonoCS__` wait, which cannot be reached because that symbol is not defined in any build config.

**Three left open for the developer**, each on its own thread above. They are all really one question — what should **Cancel** mean? Today the download continues and installs on next exit, the usual "update is ready" toast is suppressed on that path, and if a background check is already running the button cannot help at all.

Devin reports **no open bugs** at this commit; all twelve it raised across the five rounds are marked resolved, with reasons naming the fixes. Informational items not mirrored here: 27. CI (`pr-automation`, `track`) green. CodeRabbit has `auto_review.enabled: false` in this repo, so it does not review.

src/BloomExe/Collection/MinimumBloomVersionCheck.cs
andrew-polk: [Claude Opus 5 from Andrew Polk's machine during preflight]

**Fixed in `8618989c`.** Andrew's decision was unambiguous: *"Definitely can't have a button which doesn't cancel."*

So Cancel now really cancels. `ProgressUpdateReporter` owns a `CancellationTokenSource`, exposed through the base class as `UpdateReporter.CancellationToken`, and that token is handed to Velopack's `DownloadUpdatesAsync`. Pressing Cancel cancels it; Velopack abandons the transfer; `DownloadAndApplyUpdates` catches the `OperationCanceledException`, puts `_status` back to `NothingKnown` so a later attempt this session starts clean, and reports the new `UpdateAttemptOutcome.Cancelled`. The exit handler is never reached, so nothing is arranged to install.

Note this needed no new parameter on `CheckForAVelopackUpdate`: the reporter already represents whoever is watching the attempt, and cancelling is that watcher withdrawing. The toast route returns `CancellationToken.None`, so it is unaffected.

Verified against a real download: cancelled at 30%, the dialog closed in under a second, and then Bloom was exited *gracefully* so any exit handler would have run — no `Update.exe` ran, `sq.version` stayed at 6.4.105, the binaries were untouched, and no completed package was left behind. Resolving.

andrew-polk: [Claude Opus 5 from Andrew Polk's machine during preflight]

**Resolved by the fix for the sibling finding** (`8618989c`), which Andrew's answer here pointed at ("see 1").

Cancel now abandons the download outright, so there is no longer a finished download arriving after the dialog has gone, and therefore nothing for the suppressed "Update for X is ready / Restart Bloom to Update" notification to announce. The reason this flag existed — a completed download that could not tell anyone about itself — no longer occurs on this path.

`ProgressUpdateReporter.OfferToRestart` stays deliberately silent, for its original reason: the caller says what actually happens next ("Bloom will now close, install it, and start up again") on the very next line, and having both read as the same news told twice. Resolving.

src/BloomExe/ApplicationUpdateSupport.cs
andrew-polk: [Claude Opus 5 from Andrew Polk's machine during preflight]

Andrew asked for the user experience spelled out before deciding, so here it is, traced through the code rather than guessed. Leaving this thread open.

**Setting:** a Team Collection member is working, the repository starts demanding a newer Bloom, and they get the lock-out dialog. Bloom's own background update check happens to be in flight — likely, because `AutoUpdate` defaults to true.

**What they see.** They click **Upgrade Bloom** and the progress dialog opens as usual. Then, depending on how far the background check has got:

- *Still looking:* one line, in red — "Bloom is already working on checking for updates." — and a **Close** button.
- *Already downloading:* two lines — "Downloading update to 6.4.108 (88415K)", then that same red "already working on checking for updates". Which is a slightly odd pair to read together: the first says it is happening, the second reads like a refusal.

Either way they click Close and land at the collection chooser, and nothing more is said to them.

**What actually happens afterwards.** The background attempt carries on and finishes. When it does, it offers the ordinary "Update for 6.4.108 is ready / Restart Bloom to Update" toast — but they have left the workspace for the chooser, so there is no `ToastHost` to draw it and the toast goes nowhere. However, the background attempt *has* registered the install-on-exit handler, so the next time they quit Bloom the update installs, and the launch after that they are on the newer version, which may well satisfy the collection.

So: **it works, but silently, and one screen of it reads like a failure when it is not.**

**Correcting myself:** the preflight report said the in-flight check "will finish and offer the update by toast anyway, so the user does get there". That was too generous — the toast almost certainly goes nowhere, because by then they are at the chooser. The upgrade still lands, on the next quit; the notification does not.

Two things worth weighing against changing it:

- This is not a regression. The private download path this PR deletes behaved the same way.
- One part already improved: `8618989c` fixed a case where a background download that had *already finished* was reported as a failed attempt. Now that state correctly reports as downloaded and Bloom restarts into the new version. So the awkward case is narrowed to "while it is still in flight".

If you want it properly fixed, the shape is to let the dialog attach to the in-flight attempt and adopt its outcome instead of being turned away — feasible now that the reporter abstraction exists, but it is new behaviour in shared update code rather than a cleanup.

andrew-polk:
[Claude Opus 5 from Andrew Polk's machine during preflight]

## Consulted Devin again — 2026-08-19, up to commit `9f829fcf4d1a92adaee0e140689c03a39ed464a0`

Three further rounds, after Andrew's decisions came back. **No open bugs at this commit.**

The headline decision was that **Cancel must actually cancel** — "definitely can't have a button which doesn't cancel". Velopack now receives a cancellation token through the reporter and abandons the transfer. Verified against a real download: cancelled at 30%, then Bloom was exited *gracefully* so any exit handler would have run — no installer ran, `sq.version` unchanged, no completed package left behind.

Devin found three things in that work, all fixed:

- The one-download-at-a-time guard lumped "already finished downloading" in with "still downloading" and reported the finished state as a **failed** attempt, so someone asking to upgrade could be sent away to pick another collection while the new Bloom sat ready.
- A cancel landing in the same instant the transfer completed still left an install arranged for the next exit, because the `await` returned normally rather than throwing. That is now checked explicitly, which is what makes Cancel mean it even in that sliver.
- Clearing the shared websocket reference on shutdown (needed so this dialog can tell a live server from a dead one) could turn a **book upload finishing as a collection closes** into a crash. Two call sites were relying on the static pointing at a disposed-but-non-null server, where sending was harmlessly a no-op. Both now check.

Also carried out from the decision report: the `NothingNewerAvailable` message is shortened; the dead `UpdateOutcome`/`UpdateResult` are deleted; the two-hour backstop and the undisposed wait handle are kept with the decisions recorded as code comments; and the gate dialog no longer asks for a warning icon, since nothing in this ticket calls for one and no working way to request one exists — that gap is documented where the conversion happens.

Two threads above remain open by choice: the "a background check is already running" case, where Andrew asked for the user experience to be spelled out first (posted there), and verifying the offline path live, which is now a checklist item in the PR description.

The branch is squashed to a single commit, as asked. Suites: 2871 passed, 1 failed — the opt-in Reading App Builder real-build test, which needs a toolchain this machine lacks.

src/BloomExe/ApplicationUpdateSupport.cs
andrew-polk: [Claude Opus 5 from Andrew Polk's machine during preflight]

**Now implemented, in `410ae829`.** Andrew read the explanation above, asked how hard it would be to make this just work, and then asked for the fuller fix.

A second caller now **joins** the attempt already in flight instead of being turned away. `AttemptWatchers` fans everything the update code says out to every watcher; a joiner that needs it (a window that has just opened) is replayed what it missed plus the last percentage; and cancelling from any watcher cancels the download for all of them.

What decided it was not the frequency but who is affected. The window is only as long as Bloom's own check plus its download, and that check runs **once**, sixty seconds after a collection opens. On a fast machine that is about twelve seconds and this is nearly unreachable — measured here: two seconds for the check, eight to eleven for an 86 MB download. But the download time is the user's bandwidth: at 500 kbit/s it is over twenty minutes, and the people on those connections are exactly the ones a minimum-version lock-out hurts most. "Rare" was an artefact of testing on a fast desk.

Devin found seven further defects in the course of that work, which is a fair comment on how much interaction the change has. All are fixed and this commit is clean. Resolving.

src/BloomExe/Collection/MinimumBloomVersionCheck.cs
andrew-polk, line 1: [Claude Opus 5 from Andrew Polk's machine during preflight]

[Devin] **Investigate**: the mid-session lock-out blocks the UI thread in a modal progress dialog

`BrowserProgressDialog.DoWorkWithProgressDialog` calls `Form.ShowDialog`, so it must run on the UI thread, and mid-session it is reached from `TeamCollection.HandleCollectionSettingsChange` — driven by repository change notifications. Devin asked whether that route really is on the UI thread on every path, and noted the dialog now stays up for the whole download rather than a moment.

Assessed as not a defect:

- **The thread affinity is unchanged by this PR.** That path already showed a modal `BloomMessageBox` — also `ShowDialog` — before any of this work, so if the notification were arriving off the UI thread it would already have been broken. What changed is how long the modal stays up, not which thread puts it there.
- **The workspace is not frozen while it is up.** `ShowDialog` runs a nested message loop, so painting, idle work and websocket requests all continue — which is exactly why this dialog can fill in at all. The user cannot *interact* with the workspace behind it, which is intended: they are being told they may not use that collection.
- **The collection's own websocket server driving it is deliberate**, and was a bug fix: standing up a second one on the port the open collection holds is what made Bloom announce it "cannot start properly" and quit.

Resolving. Worth a note for anyone changing the Team Collection notification plumbing later: this dialog assumes the UI thread, as the message box before it did.

src/BloomExe/UpdateReporter.cs
andrew-polk, line 1: [Claude Opus 5 from Andrew Polk's machine during preflight]

[Devin] **Investigate**: cancellation registrations on the watcher list are never released

Each watcher's own token is registered against the attempt's shared source so that one watcher cancelling stops the download for all of them, and those `CancellationTokenRegistration`s are never disposed.

Assessed as not worth acting on, and consistent with a decision already recorded in this file. Both the registrations and the `ManualResetEventSlim` beside them are per-upgrade-attempt, and an attempt happens only when a user asks to be upgraded — in a process that is usually about to exit and reinstall itself. The comment in `UpdateReporter` explains why the wait handle is deliberately not disposed: three attempts at releasing it safely each had their own race, because nothing can declare the attempt over while a cancelled download may still signal it. The same reasoning covers these registrations, and Andrew reviewed and accepted it (BL-16690, 19 Aug), asking only that it be revisited if Cancel were *not* implemented. It was. Resolving.

src/BloomExe/web/BloomWebSocketServer.cs
andrew-polk, line 1: [Claude Opus 5 from Andrew Polk's machine during preflight]

[Devin] **Investigate**: clearing `Instance` on Dispose changes behaviour for callers that assume it is non-null

Fixed rather than assessed away, because Devin was right that this had teeth.

`BloomWebSocketServer.Dispose` now clears the static `Instance` when it refers to the server being disposed. That was needed so this feature can tell a live server from a dead one — without it, a collection the user had closed left `Instance` pointing at the wreck, and a progress dialog wired to it would wait forever for a reply that could never come, with no close box to escape by.

But it does change what a caller sees. Previously `Instance` kept pointing at a disposed server and sending on one was harmlessly a no-op; now it can be null. I audited every use. Most already wrote `Instance?.`; `WorkspaceView` guards with an explicit null check. Two did not, and both are now null-safe:

- `BookUpload` announced a finished upload — Devin's own example, an upload completing as a collection closes;
- `EditingView` reported a failed thumbnail download from inside a `catch`, where an exception would have replaced a handled error with an unhandled one.

`WebSocketProgress` is null-safe internally, so the places that build one over `Instance` were already fine. Losing a notification is the right failure here; a `NullReferenceException` is not. Resolving.

andrew-polk:
[Claude Opus 5 from Andrew Polk's machine during preflight]

## Consulted Devin — 2026-08-19, up to commit `410ae8291785dcd0bbd240113fe45953c647a10b`

Eight rounds on the second commit, which makes a caller **join** an update attempt already in flight rather than being told Bloom is busy. **No open bugs at this commit.**

That commit exists because of the decision recorded on the thread above: the "Bloom is already working on checking for updates" case looked like a refusal even though the upgrade was under way. What decided it was not frequency but bandwidth — Bloom's own check runs once, sixty seconds after a collection opens, so the window is that check plus its download: about twelve seconds here, but over twenty minutes at 500 kbit/s, and those users are the ones a lock-out hurts most.

It is by far the most interaction-heavy change in this PR, and Devin found **seven** defects in it, which is a fair comment on that. Two would have broken Bloom outright and both were mine, made minutes earlier:

- The dialog's own reporter was left out of the watcher list on the "an update was found earlier and is now accepted" path, so it watched a window that never moved, with a Cancel that could not stop the download.
- Then, after I simplified that away, the watcher list was **replaced** when a check turned into its download — dropping anyone who had joined during the check. That is precisely the case the commit exists to fix.

The other five, each real:

- `joinable.Add(reporter)` ran while holding the status lock, and joining replays messages to the UI — a deadlock waiting to happen, and against the rule the comment two lines above it stated. Found by reading my own diff rather than by Devin.
- The "Update Now" toast's accept action closed over the watcher list and handed it to itself, which my own self-reference guard then rejected outright, so **nothing downloaded** for anyone with automatic updating off. Fixed by never passing the list where an individual reporter belongs.
- A `CancellationTokenSource` cannot be un-cancelled, so re-arming a list for a further attempt carried an old cancel into it and abandoned the next download silently. Re-arming is gone entirely; a new attempt gets a new list.
- The restart toast could be handed a null restart action on the dialog's path, and mid-session that toast really renders.
- Replaying the backlog to a joining *toast* reporter announced a check the user never asked about. Only a watcher that wants catching up gets it now — and a Verbose manual check gets its immediate acknowledgement back, which that fix had removed.

Three further flags were assessed rather than changed, each with the reasoning on its own thread: the mid-session dialog's UI-thread affinity (unchanged by this PR — the message box before it was also modal), the undisposed cancellation registrations (the decision Andrew already recorded for the wait handle covers them), and the `Instance`-on-`Dispose` change, where I *did* act — two callers were relying on the static pointing at a disposed-but-non-null server, including a book upload finishing as a collection closes.

Verified live on each iteration: a real upgrade 6.4.105 → 6.4.108, and a cancel at 30% followed by a graceful exit which ran no installer and left `sq.version` untouched. **The joining path itself is still not exercised** — reaching it needs a Team Collection whose repository changes while a check is in flight, which the rig cannot set up.

Suites: 2871 passed, 1 failed — the opt-in Reading App Builder real-build test, which needs a toolchain this machine lacks. One run also failed `ProblemLanguages_KeepsAsteriskMatch`, a network-dependent license-checker test in a file this branch does not touch; it passes 3/3 in isolation and did not recur.

The fix-and-re-review cycle ran far past preflight's cap of four. Each round found something real, so stopping earlier would have meant shipping it.

src/BloomExe/Collection/MinimumBloomVersionCheck.cs
```diff
@@ -0,0 +1,802 @@
...
+            {
+                if (reporter.WaitForFinish(TimeSpan.FromMilliseconds(250)))
+                {
+                    // A cancel that arrived during that wait still counts: the user pressed it
+                    // before they could possibly know the download had just landed, and taking the
+                    // download's word for it here is what would restart Bloom under someone who had
+                    // said no.
+                    if (worker.CancellationPending)
+                    {
+                        reporter.CancelTheDownload();
```
andrew-polk, line 759: [Claude Fable 5 from Andrew Polk's machine during preflight] Devin flagged (non-severe): **Cancelling an upgrade at the last moment can still install the new version when Bloom is next closed.** If the user's Cancel lands in the same instant the download finishes, the code refuses to restart Bloom (the `UserCancelled` check), but by then `DownloadAndApplyUpdates` may already have registered the install-on-exit handler — so the update quietly installs whenever Bloom is eventually closed, even though the user said no.

andrew-polk: [Claude Fable 5 from Andrew Polk's machine during preflight] Assessed and deliberately left as is. This is the reviewed cancel-vs-finish sliver already documented in the code (see the comment block at the end of ProgressUpdateReporter and the UserCancelled check in UpgradeBloom): the window is the sub-second gap between the download's completion guard and the exit-handler registration. A cancel that lands there never restarts Bloom under the user -- the UserCancelled check guarantees that -- and what remains is ordinary background-update behavior: a fully downloaded update installs when Bloom eventually exits, exactly as if auto-update had fetched it. Undoing an already-registered install-on-exit would need a way to disarm the anonymous ApplicationExit handler, which is real complexity for a race the user cannot perceive reliably; the project's standing call on this feature has been simple working code over bulletproofing rare slivers. Flagged in the preflight report as an FYI in case the developer wants it hardened after all.

src/BloomExe/Collection/MinimumBloomVersionCheck.cs
```diff
@@ -0,0 +1,802 @@
...
+            // collection, so the startup sync that would refresh their own copy never runs, and an
+            // administrator's mistaken requirement would shut them out on every launch for ever,
+            // fixable only by hand-editing the file on each machine.
+            //
+            // Otherwise -- an ordinary collection, or a shared folder we cannot reach right now --
+            // we fall back to what we have: the file on this computer and anything the repository
+            // told us earlier this session, taking whichever demands more. Not being able to see the
+            // repository is not the same as the repository saying "no requirement", so in that case
+            // we keep the protection rather than quietly dropping it.
+            var declaredVersion =
```
andrew-polk, line 49: [Claude Fable 5 from Andrew Polk's machine during preflight] Devin flagged (Investigate): **A Team Collection repo copy with no MinimumBloomVersion silently cancels the local file's requirement.** When the repository's settings are readable and carry no minimum, that answer wins over the local file — so a repo zip regenerated without the element removes the protection with no trace.

andrew-polk: [Claude Fable 5 from Andrew Polk's machine during preflight] Confirmed intended, and it is the harder-won half of the design rather than an oversight. A readable repo answer has to govern in both directions because a member who is being refused never opens the collection, so the sync that would refresh their own copy never runs -- if the repo could only raise the requirement, an administrator's mistaken minimum would lock members out on every launch forever, fixable only by hand-editing each machine. The regenerated-zip case Devin describes is the same event as an administrator withdrawing the requirement: the repo's collection settings are the team's single source of truth, and whoever writes them without the element is, in effect, withdrawing it. The comment at the top of IsThisBloomTooOld records this reasoning; the commit message for the gate spells it out at more length.

andrew-polk:
[Claude Fable 5 from Andrew Polk's machine during preflight] Consulted Devin on this PR up to `4ee539da57a001b06014f081c694ab5c2ac586d1` (two rounds, this preflight run).

**Round 1** (on `d291122539`): one non-severe bug — the cancel-vs-finish sliver — mirrored and answered in [this thread](https://github.com/BloomBooks/BloomDesktop/pull/8213#discussion_r3817509706) (left as a documented, reviewed decision; also raised in the preflight report for the developer). Of six informational flags, three were acted on in the squashed commit (two stale comments fixed, a write-only property removed), one was factually mistaken (`ParseOrNull` still has a caller), and two describe deliberate choices documented in code.

**Round 2** (on the final `4ee539da57`): no bugs. One Investigate flag — the repo's empty answer lifting a local requirement — confirmed as the intended design in [this thread](https://github.com/BloomBooks/BloomDesktop/pull/8213#discussion_r3817570229). Remaining informational flags describe pre-existing behavior (a failed attempt parking update checks for the session) or rare cosmetics (the bar restarting after a waited-out download), none introduced by this PR.

CI is green on the current head; the full C# suite passes (2871, with the one known opt-in RAB environmental failure in an untouched fixture). The gate commit's content carries its byte-identical review history from #8206.

src/BloomExe/Collection/MinimumBloomVersionCheck.cs
andrew-polk: [Claude Fable 5 from Andrew Polk's machine during preflight] Developer decision recorded: leave as is. Andrew reviewed this in the preflight report and confirmed the trade-off — the sub-second window and standard install-on-exit effect don't justify machinery to disarm an already-registered exit handler.

andrew-polk (commented):
> @andrew-polk resolved 4 discussions.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8213)* status: 0 of 19 files reviewed, 5 unresolved discussions.

andrew-polk (commented):
> @andrew-polk reviewed 19 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8213)* status: 0 of 19 files reviewed, 5 unresolved discussions.

StephenMcConnel (changes requested):
One minor comment that might not be fixable.

> @StephenMcConnel reviewed 19 files and all commit messages, made 2 comments, and resolved 5 discussions.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8213#-:-P-WAr_F1ea_SNy87kff:b-bihc6g)* status: all files reviewed, 1 unresolved discussion (waiting on andrew-polk).

___
*[`-- commits` line 243 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8213#-P-VspUW7fqYDxhRtEeS:-P-VspUW7fqYDxhRtEeT:b-9i5yw):*
What does this have to do with minimum version?  Did your branch get polluted? Everything following seems unrelated.

andrew-polk (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8213)* status: all files reviewed, 1 unresolved discussion (waiting on StephenMcConnel).

___
*[`-- commits` line 243 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8213#-P-VspUW7fqYDxhRtEeS:-P-WMycM4-8WWpTgqtEH:b-ghhp21):*

What does this have to do with minimum version?  Did your branch get polluted? Everything following seems unrelated.
</blockquote></details>

Wow. Yeah, apparently it just blindly took messages from a merge or something and folded those into the squashed commit message. I'll have it fix it. 
I don't pay much attention to commit messages. I guess I should.

andrew-polk (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8213)* status: all files reviewed (commit messages unreviewed), 1 unresolved discussion (waiting on StephenMcConnel).

___
*[`-- commits` line 243 at r3](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8213#-P-VspUW7fqYDxhRtEeS:-P-WOREr1gGLkJKc97fz:b-896fix):*

Wow. Yeah, apparently it just blindly took messages from a merge or something and folded those into the squashed commit message. I'll have it fix it. 
I don't pay much attention to commit messages. I guess I should.
</blockquote></details>

Done.

StephenMcConnel (approved):
> @StephenMcConnel reviewed all commit messages and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8213)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on andrew-polk).

PR #8214, opened 2026-08-17
Fix faulty page number placement in Zero Margin portrait ebooks (BL-16695)

andrew-polk:
In the Zero Margin ebook theme, a page number the user forces to the Left or Right in Book Settings sat flush against the page edge on portrait content pages, because the forced placement uses `var(--page-margin-left/right)` and this theme sets that margin to 0 there. The text box, by contrast, keeps a thin visual margin from `--topLevel-text-padding`, so the number looked misaligned (BL-16695).

Fix:

- `AppearanceSettings.SetPageNumberProperties` now emits the forced left/right overrides with an extra theme-overridable term, `var(--pageNumber-forced-side-extra-margin, 0px)`. The `0px` fallback leaves every other theme's behavior unchanged. (A theme-only fix was tried first and does not work: the override is declared in the generated `.bloom-page { }` block, so its `var()` references are substituted at computed-value time on the page element — a pseudo-element redefinition of `--page-margin-left` comes too late.)
- The Zero Margin theme sets that variable to `var(--topLevel-text-padding)` on portrait numbered pages, so a forced side page number aligns with the text box's thin margin. Landscape is untouched: its flush corner circle is that theme's deliberate design.
- Updated the two affected `ToCss_SetsRelevantPageNumberPositionOverrides` test cases.

Verified in the running app (Ebook 2x3 Portrait, zero-margin theme): forced Left/Right insets the number 0.5em (≈7.3px at the 11pt page-number font) from the edge; Automatic placement still centers exactly.

Ref: https://issues.bloomlibrary.org/youtrack/issue/BL-16695

[Devin review](https://devinreview.com/BloomBooks/BloomDesktop/pull/8214)

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8214)

### DIFF
--- a/src/content/appearanceThemes/appearance-theme-zero-margin-ebook.css
+++ b/src/content/appearanceThemes/appearance-theme-zero-margin-ebook.css
@@ -32,6 +32,11 @@ Note that hiding the page numbers is done by a setting in appearance.json, not h
     /* this rule will apply more generally than we'd prefer, but the common situation we're improving here
     is picture on top, text on the bottom, we need to make room for the page number */
     --pageNumber-extra-height: 12mm !important;
+
+    /* The page margin is zero, so a page number the user forces to the left or right would
+       sit flush against the page edge. Inset it to line up with the thin visual margin the
+       text box gets from --topLevel-text-padding (BL-16695). */
+    --pageNumber-forced-side-extra-margin: var(--topLevel-text-padding);
 }
 .bloom-fullBleed .Ebook7x5Landscape.numberedPage,
 .bloom-fullBleed .Device16x9Landscape.numberedPage {
--- a/src/content/appearanceThemes/appearance-theme-zero-margin-ebook.css
+++ b/src/content/appearanceThemes/appearance-theme-zero-margin-ebook.css
@@ -32,10 +32,14 @@ Note that hiding the page numbers is done by a setting in appearance.json, not h
     /* this rule will apply more generally than we'd prefer, but the common situation we're improving here
     is picture on top, text on the bottom, we need to make room for the page number */
     --pageNumber-extra-height: 12mm !important;
-
+}
+.Ebook2x3Portrait.numberedPage:not(.bloom-interactive-page),
+.Device16x9Portrait.numberedPage:not(.bloom-interactive-page) {
     /* The page margin is zero, so a page number the user forces to the left or right would
        sit flush against the page edge. Inset it to line up with the thin visual margin the
-       text box gets from --topLevel-text-padding (BL-16695). */
+       text box gets from --topLevel-text-padding (BL-16695).
+       Not on interactive pages: they keep a nonzero margin (the zero-margin rule above
+       excludes them), so the forced number already sits at the page margin there. */
     --pageNumber-forced-side-extra-margin: var(--topLevel-text-padding);
 }
 .bloom-fullBleed .Ebook7x5Landscape.numberedPage,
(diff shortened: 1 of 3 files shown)


src/content/appearanceThemes/appearance-theme-zero-margin-ebook.css
```diff
@@ -32,6 +32,11 @@ Note that hiding the page numbers is done by a setting in appearance.json, not h
     /* this rule will apply more generally than we'd prefer, but the common situation we're improving here
     is picture on top, text on the bottom, we need to make room for the page number */
     --pageNumber-extra-height: 12mm !important;
+
+    /* The page margin is zero, so a page number the user forces to the left or right would
+       sit flush against the page edge. Inset it to line up with the thin visual margin the
+       text box gets from --topLevel-text-padding (BL-16695). */
+    --pageNumber-forced-side-extra-margin: var(--topLevel-text-padding);
```
andrew-polk, line 43: [Devin] **Investigate**: New theme variable also applies to interactive (game) portrait pages, where the page margin is not zero

The declaration is added to the `.Ebook2x3Portrait.numberedPage, .Device16x9Portrait.numberedPage` rule, which â€” unlike the zero-margin rule higher in this file and the padding rule below it â€” does not exclude `.bloom-interactive-page`. Games template pages are both `numberedPage` and `bloom-interactive-page`, so on such portrait pages the page margin is the theme default (non-zero) yet a forced Left/Right page number now gets an extra inset. Because the `:where(...:not(.bloom-interactive-page))` rule does not define `--topLevel-text-padding` there, the value would come from the default theme block (`--topLevel-text-padding: 1em`), i.e. roughly 1em of unintended extra inset rather than the intended 0.5em alignment. Note the pre-existing `--pageNumber-extra-height: 12mm !important` in the same rule shares the same selector scope, so this may be deliberate consistency; worth a quick check whether the exclusion should be added.

andrew-polk: [Claude Fable 5 from Andrew Polk's machine during preflight] Fixed in `a36cbc3140`: the inset now lives in its own rule scoped with `:not(.bloom-interactive-page)`, matching the theme's zero-margin and padding rules — so game pages (which keep a nonzero margin in this theme) no longer get the extra inset on a forced Left/Right page number.

andrew-polk:
[Claude Fable 5 from Andrew Polk's machine during preflight] Consulted Devin on 2026-08-17 up to commit `a36cbc3140`.

Two rounds. Round 1 (at `ad00de70`): no bugs; one Investigate flag — the new inset variable also applied to interactive (game) portrait pages, which keep a nonzero margin in this theme. Fixed in `a36cbc3140` by scoping the rule with `:not(.bloom-interactive-page)`; see the resolved thread at https://github.com/BloomBooks/BloomDesktop/pull/8214#discussion_r3799524811. Round 2 (at `a36cbc3140`): no bugs; the same flag reappeared with analysis text describing the pre-fix selector, which no longer exists at this commit — treated as stale, outcome already documented on that thread. Three informational notes (not mirrored, low signal): the em-based inset is approximate rather than exact alignment (acknowledged in the PR description); setting text padding to 0 makes the number flush again, which is correct alignment behavior in that case; and the format-button dodge queries are unaffected by the longer calc string. CI (pr-automation, track) is green.

JohnThomson (approved):
> @JohnThomson reviewed 3 files and all commit messages, and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8214)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on andrew-polk).

StephenMcConnel (approved):
> @StephenMcConnel reviewed 3 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8214)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on andrew-polk).

PR #8216, opened 2026-08-19
Make the Standard/Custom cover layout menu select instead of toggle (BL-16725)

StephenMcConnel:
The Standard/Custom cover-layout menu above a cover page drew a tick beside the layout you were on, but clicking the already-ticked item switched you to the other layout. Clicking **Custom** while already on Custom therefore threw the page back to Standard and — because that counts as a deliberate switch to Standard — **discarded the user's saved custom cover layout**, with no dialog and no undo.

Nothing compared the clicked value with the current state, and the endpoint it called was a pure toggle that ignored the choice entirely.

## What changed

- `customPageLayoutMenu.tsx` — `handleSelect` returns early (just closing the menu) when the clicked item is already the current layout. These items are a radio-style selection, not toggles.
- The API now says what it means: `editView/toggleCustomPageLayout` → **`editView/setCustomPageLayout`**, carrying the desired layout (`"standard"` | `"custom"`). The C# handler no-ops when the page is already in that layout, and sets the `bloom-customLayout` class explicitly rather than flipping it. The plain-string request-shape fallback, which no caller used, is gone.
- `setupPageLayoutMenu`'s legacy-theme force-to-standard had been exploiting toggle semantics ("I know it's custom, so toggle it"); it now asks for `"standard"` explicitly.
- New `customPageLayoutMenu.test.tsx` — covers both redundant-click cases and both real switches. Verified it fails without the guard.

## Verified in a running Bloom

Front cover of a `zero-margin-ebook` book, Enterprise subscription:

- Clicking the ticked **Custom** while on Custom → nothing changes; the custom arrangement stays intact. (This was the data-loss case.)
- Clicking the ticked **Standard** while on Standard → nothing changes.
- Standard→Custom and Custom→Standard still work, exercising both the first-time-generation (`"false"` reply) and saved-data paths.
- A plain **Standard** switch still discards the saved layout by design (`customOutsideFrontCover` data 3932 → 0 chars); **ctrl+shift+Standard** still preserves it (4017 chars survived).
- Switching the book to `legacy-5-6` while the cover was custom correctly forced it back to standard *and kept* the data — the `setupPageLayoutMenu` path through the renamed endpoint — and the menu then showed **Custom** disabled.

Ref: https://issues.bloomlibrary.org/youtrack/issue/BL-16725

[Devin review](https://devinreview.com/BloomBooks/BloomDesktop/pull/8216)

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8216)

### DIFF
--- a/src/BloomBrowserUI/bookEdit/toolbox/canvas/customPageLayoutMenu.test.tsx
+++ b/src/BloomBrowserUI/bookEdit/toolbox/canvas/customPageLayoutMenu.test.tsx
@@ -1,0 +1,137 @@
+import { act } from "react";
+import { afterEach, describe, expect, it, vi } from "vitest";
+import { renderRoot, unmountRoot } from "../../../utils/reactRender";
+import { CustomPageLayoutMenu } from "./customPageLayoutMenu";
+
+// The menu asks the server whether the CustomXMatterPage feature is available. Returning
+// undefined is what the component treats as "no information yet, so not blocked", which is the
+// state we want for these tests: both items live and clickable.
+vi.mock("../../../react_components/featureStatus", () => ({
+    useGetFeatureStatus: () => undefined,
+    useGetFeatureAvailabilityMessage: () => "",
+}));
+
+// Only the legacy-theme tooltip's link uses this; mocked so these tests don't drag in the
+// workspace frames machinery.
+vi.mock("../../js/workspaceFrames", () => ({
+    getWorkspaceBundleExports: () => ({
+        showBookSettingsDialog: () => {},
+    }),
+}));
+
+let renderedContainer: HTMLDivElement | undefined;
+
+// Renders the menu closed, then clicks its button to open it, since that is the only way to get
+// the items into the DOM. Returns the setCustom spy the items should (or should not) call.
+function renderAndOpenMenu(isCustom: boolean) {
+    const setCustom = vi.fn();
+    const container = document.createElement("div");
+    document.body.appendChild(container);
+    renderedContainer = container;
+
+    // React 18's createRoot renders asynchronously; act() flushes the render and its effects.
+    act(() => {
+        renderRoot(
+            <CustomPageLayoutMenu isCustom={isCustom} setCustom={setCustom} />,
+            container,
+        );
+    });
+
+    const button = container.querySelector("button");
+    if (!button) {
+        fail(
+            "The layout menu did not render its button, so nothing can be clicked.",
+        );
+    }
+    act(() => {
+        button.dispatchEvent(new MouseEvent("click", { bubbles: true }));
+    });
+
+    return setCustom;
+}
+
+// The MUI Menu renders into a portal on document.body, not into our container. An item's text is
+// whatever the localization lookup produced: the English label in production, but the l10n id
+// itself under the localizationManager mock in vitest.setup.ts, so accept either.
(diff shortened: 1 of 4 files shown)


StephenMcConnel:
[Claude Opus 5 (1M context) from Steve McConnel's machine during preflight]

Consulted Devin on 2026-08-19 12:10 up to commit `23e51e809`.

**No Bugs and no Investigate flags** — nothing to mirror as a review thread. CI (`pr-automation`) is green, and CodeRabbit's `auto_review` is disabled in this repo's `.coderabbit.yml`, so it isn't a reviewer here.

Devin raised six informational observations across the three commits. Two we acted on:

- *Unrecognized `layout` values silently mean "standard"* — fixed in `7124d0dc0`. That silent default was the branch that discards the saved custom layout, and `AGENTS.md` prefers throwing when our own API is misused, so the handler now rejects anything that isn't `"standard"` or `"custom"`.
- *Tick detection via "any svg in the item"* — fixed in `23e51e809`; the test now matches MUI's CheckIcon by its `data-testid`.

(Both were re-listed unchanged in the final run — Devin re-reports findings from earlier commits, so those two entries are stale.)

Four we assessed and deliberately left alone:

- *The no-op reply of `"true"` reads as success if client and server disagree about the layout* — Devin notes this isn't a regression (the old toggle flipped to the wrong layout in the same situation). Changing the reply shape would disturb the `"false"` first-time-custom protocol for a case we have no evidence occurs.
- *The menu can show a stale layout for a moment right after a switch, so an immediate click to switch straight back could be swallowed* — sub-second window, benign (the click just does nothing), the staleness is pre-existing, and the page reload immediately re-derives the menu from the real DOM.
- *When the subscription doesn't include custom covers, clicking "Custom" opens the subscription settings and leaves the layout menu open behind it* — real, but pre-existing behavior of the shared menu component and untouched by this PR. Flagged for the developer as a possible follow-up rather than changed here, since it alters what a user sees.
- *The test clicks portal nodes with raw `dispatchEvent`* — works because React registers its delegated listeners on portal containers too; if that ever changed the test would fail confusingly, which its own diagnostic message already hints at. Not worth restructuring the test for.

StephenMcConnel:
[Claude Opus 5 (1M context) from Steve McConnel's machine during preflight]

Steve reviewed the four informational items from the [consultation log above](https://github.com/BloomBooks/BloomDesktop/pull/8216#issuecomment-5346170215) and confirmed leaving all four as they are — no code changes:

- the layout menu staying open behind the subscription dialog (pre-existing, and the fix would land in a menu component shared across Bloom);
- the server's "already in that layout, nothing to do" reply being indistinguishable from a successful switch (no evidence the two ends ever disagree, and that reply also carries the "no saved custom layout yet" signal);
- the tick briefly showing the old layout in the moment before the page reloads (sub-second, and the only effect is that a click does nothing);
- the C# handler's two safeguards having no automated test (`EditingViewApi` has no fixture; both were exercised by hand in a running Bloom).

No follow-up code comments were requested. Nothing further this round — the branch stays a draft pending Steve's own review.

JohnThomson (approved):
> @JohnThomson reviewed 4 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8216)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on StephenMcConnel).

PR #8217, opened 2026-08-19
Keep typing safe around ligatures wherever the split came from (BL-16717)

JohnThomson:
## Problem

In fonts that join letters together — Andika, Bloom's default, and Charis both join `ff`, `fl`
and `ffl` — letters the user has typed sometimes stop being drawn, and the caret sometimes stops
moving as they arrow across those letters. The text is all there and saves correctly; only the
screen is wrong, and it comes back on a font change or a page reload. The first fix for this
shipped in 6.5, and the reporter still sees it on 6.5.1322, about 1% of the time when typing just
after a ligature.

## Cause

Chromium drops glyphs from a paragraph whose text sits in two adjacent text nodes: it shapes the
paragraph as a single run and then hands the glyphs out per text node, so a ligature that spans
the boundary loses glyphs — not when the split happens, but at the next edit, which for the user
is simply the next letter they type. The caret misbehaves in the same state, staying drawn in one
place for two or three arrow presses.

The first fix repaired the splits Bloom itself caused, with ckeditor bookmarks, and stopped taking
those bookmarks for ordinary typing. But bookmarks are not the only thing that splits a paragraph:
**Chromium's own backspace** in the middle of a word splits it, **long-press** composes its
character into a text node of its own, and **ckeditor** takes bookmarks of its own for Enter, paste
and the style commands. None of those were ever put back together.

## Fix

- The repair no longer depends on Bloom having caused the split:
  `mergeTextNodesSplitByBookmarks` becomes `mergeAdjacentTextNodes`, and its comment names the
  other causes.
- It runs **on keydown, before the keystroke lands**, so the character arrives in a paragraph that
  is one whole text node again. This is the part that keeps the letters on screen — the existing
  half-second sweep runs *after* the damage.
- That half-second sweep, at the end of `handlePageEditing`'s `mainTask`, now runs
  unconditionally, catching the splits no keystroke follows: a menu paste, an IME, an undo, a
  long-press the user doesn't type after.
- The merge stands down where ckeditor is holding a "filling char" — a text node ckeditor tracks
  and will come back to edit. It asks ckeditor rather than searching the text for a zero-width
  space (some scripts use one in real text), ignores a reference ckeditor has been left holding to
  a node no longer in the box, and stands down only for the paragraph the char is in.
- Both entry points are no-ops unless the box really is split: 0.056 ms to check on a
  2000-text-node box.

Ref: https://issues.bloomlibrary.org/youtrack/issue/BL-16717

[Devin review](https://devinreview.com/BloomBooks/BloomDesktop/pull/8217)

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8217)

### DIFF
--- a/src/BloomBrowserUI/bookEdit/js/editableDivUtils.ts
+++ b/src/BloomBrowserUI/bookEdit/js/editableDivUtils.ts
@@ -542,19 +542,20 @@ export class EditableDivUtils {
         if (!EditableDivUtils.hasAdjacentTextNodes(element)) {
             return; // nothing to merge; leave the box, and the selection, strictly alone
         }
-        if (EditableDivUtils.ckEditorIsHoldingAFillingChar(element)) {
-            // Leave the box strictly alone while ckeditor is holding a filling char.
-            // That char is a zero-width space in a text node of its own, which ckeditor
-            // inserts at a collapsed selection next to an inline element so the caret shows,
-            // and it remembers THAT NODE (setCustomData("cke-fillingChar", ...)) to take the
-            // char out again later. normalize() would fold the node into its neighbour, so
-            // ckeditor would be left writing to a detached node and the zero-width space
-            // would stay in the text and get saved into the book - the BL-16490 hazard.
-            // Skipping costs us nothing lasting: the filling char is transient, and the next
-            // keystroke brings us back here to do the repair.
-            return;
-        }
-        EditableDivUtils.mergeAdjacentTextNodeRuns(element);
+        // Leave ckeditor's filling char alone. That char is a zero-width space in a text node
+        // of its own, which ckeditor inserts at a collapsed selection next to an inline element
+        // so the caret shows, and it remembers THAT NODE
+        // (setCustomData("cke-fillingChar", ...)) to take the char out again later. normalize()
+        // would fold the node into its neighbour, so ckeditor would be left writing to a
+        // detached node and the zero-width space would stay in the text and get saved into the
+        // book - the BL-16490 hazard.
+        // We hand the node down rather than giving up on the whole box, so a filling char in
+        // one paragraph doesn't stop us repairing another: only the runs whose merge would
+        // reach it are skipped, and they come back to us on the next keystroke.
+        EditableDivUtils.mergeAdjacentTextNodeRuns(
+            element,
+            EditableDivUtils.ckEditorFillingCharNodeIn(element),
+        );
     }
 
     // Merge each run of adjacent sibling text nodes back into one.
@@ -581,7 +582,10 @@ export class EditableDivUtils {
     // can be the .bloom-editable itself, and hiding a focused contenteditable blurs it - the
     // user's next keystrokes would go nowhere - besides leaving a stray style="" attribute
     // behind in the saved book.
-    private static mergeAdjacentTextNodeRuns(element: HTMLElement): void {
+    private static mergeAdjacentTextNodeRuns(
+        element: HTMLElement,
+        fillingCharNodeToLeaveAlone?: Node,
+    ): void {
         const doc = element.ownerDocument;
         // The node each run will collapse into: normalize() appends a run's text onto its
         // first non-empty node and removes the others, so that is the node that survives, and
@@ -614,7 +618,17 @@ export class EditableDivUtils {
 
         survivorOfEachRun.forEach((survivor) => {
             // A text node always has a parent element here: we are walking inside element.
-            survivor.parentElement!.normalize();
+            const parent = survivor.parentElement!;
+            // normalize() is DEEP - it merges every text node under `parent`, not just this
+            // run - so the filling char is out of reach only if it is outside `parent`
+            // altogether. See mergeAdjacentTextNodes() for why we must not touch it.
+            if (
+                fillingCharNodeToLeaveAlone &&
+                parent.contains(fillingCharNodeToLeaveAlone)
+            ) {
+                return;
+            }
+            parent.normalize();
             if (!survivor.isConnected) {
                 return; // the whole run was empty nodes; normalize() removed them all
             }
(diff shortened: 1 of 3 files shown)


src/BloomBrowserUI/bookEdit/js/editableDivUtils.ts
```diff
@@ -633,6 +659,35 @@ export class EditableDivUtils {
...
+        const ckEditorOfThisBox = (
+            element as HTMLElement & {
+                bloomCkEditor?: {
+                    editable: () => { getCustomData: (key: string) => unknown };
+                };
+            }
+        ).bloomCkEditor;
+        return !!ckEditorOfThisBox
+            ?.editable()
+            ?.getCustomData("cke-fillingChar");
```
JohnThomson: [Devin] **Bug**: Glyph-dropping repair can be skipped after a reader-tool rewrite, leaving letters missing for a keystroke

The repair that rejoins split text is skipped whenever the editor still remembers a helper character node (`getCustomData("cke-fillingChar")`), **even after that node has been thrown away by a box rewrite** — so a paragraph can stay in its broken state and keep letters unpainted.

Why an orphaned filling char still blocks the merge, contrary to the comment above it: ckeditor stores the filling-char reference as custom data on the *editable element* (`ckeditor.js`: `n.setCustomData("cke-fillingChar", i)`), not on the text node. `EditableDivUtils.doCkEditorCleanup` assigns `div.innerHTML`, which detaches the tracked text node but leaves the custom data pointing at it — precisely the BL-16490 "orphaned filling char" case. The guard then returns truthy, so `mergeAdjacentTextNodes` returns early both in `restoreSelectionFromCkEditorBookmarks` and in the new keydown handler, until ckeditor's next `selectionChange` clears the key. Before this PR the merge in `restoreSelectionFromCkEditorBookmarks` always ran, so this is a behavior regression for that path.

Suggested: check that the remembered node is still connected and inside `element`, which would make the code match its own comment.

src/BloomBrowserUI/bookEdit/js/editableDivUtils.ts
```diff
@@ -523,12 +536,24 @@ export class EditableDivUtils {
...
+            // That char is a zero-width space in a text node of its own, which ckeditor
+            // inserts at a collapsed selection next to an inline element so the caret shows,
+            // and it remembers THAT NODE (setCustomData("cke-fillingChar", ...)) to take the
+            // char out again later. normalize() would fold the node into its neighbour, so
+            // ckeditor would be left writing to a detached node and the zero-width space
+            // would stay in the text and get saved into the book - the BL-16490 hazard.
+            // Skipping costs us nothing lasting: the filling char is transient, and the next
+            // keystroke brings us back here to do the repair.
+            return;
+        }
```
JohnThomson: [Devin] **Investigate**: Filling-char guard can make the per-keystroke repair a no-op in boxes full of inline elements

ckeditor (re)creates a filling char on `selectionChange` whenever the collapsed caret sits next to an inline element or before a `<br>`, and only removes it on the *second* selectionChange (`lib/ckeditor/ckeditor.js`: `b.getCustomData("ready") ? h(a) : b.setCustomData("ready", 1)`). In boxes that are dense with inline markup — Talking Book sentence spans, bold/italic runs — the caret is frequently adjacent to such an element, so the new guard may be true at keydown time on many consecutive keystrokes, turning off exactly the per-keystroke repair this PR adds. The 500ms sweep still catches it, but the user-visible glitch that the keydown path exists to prevent may still show up in those boxes. Worth checking by typing next to a Talking Book span with Andika.

src/BloomBrowserUI/bookEdit/js/editableDivUtils.ts
JohnThomson: [Claude Opus 5 from John Thomson's machine during preflight] Fixed in `aa20452a33`. You were right, and right about the mechanism: the reference lives on the editable div, so a `div.innerHTML = …` in `doCkEditorCleanup` or `cleanUpNbsps` leaves it pointing at a node that is no longer in the document, and the guard was reading that as "ckeditor is busy here, keep out". Worse for `restoreSelectionFromCkEditorBookmarks`, where the merge used to be unconditional, so it was a straight regression on the path the reader tools use.

The guard now asks for the node and skips only while it is still inside the box (`element.contains`), which is what the comment claimed all along. It also hands the node down to `mergeAdjacentTextNodeRuns` instead of giving up on the whole box, so a filling char parked in one paragraph no longer stops another paragraph being repaired — the test being whether the char is under the parent we are about to `normalize()`, since that call is deep.

Two new unit tests pin it: one where the tracked node has been detached (the repair must go ahead), and one where a filling char in one paragraph must not stop the other being merged.

src/BloomBrowserUI/bookEdit/js/editableDivUtils.ts
JohnThomson: [Claude Opus 5 from John Thomson's machine during preflight] Measured it rather than argued it, by doing exactly what you suggested — typing next to a Talking Book span in Andika in a running Bloom, over CDP, with a capture-phase keydown hook reading the same state the guard reads (a tracked node that is still inside the box).

- Talking Book sentence spans: **0 of 30 keydowns**
- a bold boundary: **0 of 30**
- plain text: **0 of 30**

Then I tried to force it, parking the caret in the five positions ckeditor's own condition describes — immediately after `</strong>`, immediately before `<strong>`, before a `<br>`, inside an empty `<strong>`, and just after an audio-sentence span — placing the selection programmatically and then pressing a real key so ckeditor ran its selection handling. It never created one in any of them. So in Bloom's WebView2 this branch appears not to be reached in practice; the guard is insurance rather than a cost.

Two things from your other finding also bound what happens when it *does* fire (`aa20452a33`): it can no longer be stuck on by an orphaned reference, and it now skips only the runs whose `normalize()` would reach the char, not the whole box. So the worst case is one paragraph's repair deferred to the next keystroke.

JohnThomson:
[Claude Opus 5 from John Thomson's machine during preflight] Consulted Devin on 2026-08-19, up to commit `d7a23b0a7f`.

Across three passes it raised one Bug and one Investigate flag, both about the filling-char guard, and both now have their own resolved thread:

- **Bug — an orphaned filling char blocked the repair.** Real, and a regression on the reader-tool path. Fixed in `aa20452a33`; [thread](https://github.com/BloomBooks/BloomDesktop/pull/8217#discussion_r3815649086).
- **Investigate — does the guard fire often enough to turn the per-keystroke repair off in inline-heavy boxes?** Measured in a running Bloom rather than argued: 0 of 90 keydowns, and not producible at all by parking the caret where ckeditor's own rule says it creates one. [thread](https://github.com/BloomBooks/BloomDesktop/pull/8217#discussion_r3815650387).

Note for anyone reading the Devin page directly: the two later passes **re-list both of those against the pre-fix line numbers** (the Bug still cites `editableDivUtils.ts:686-688`, which no longer holds that code, and carries the first pass's finding id). I checked each against the current file rather than taking the re-report at face value; both are addressed.

Eleven informational items in the final pass, none acted on. Three were new and each is a fair observation that Devin itself concludes is harmless: `normalize()` on a keystroke also nudges the Talking Book highlight-integrity observer (repair-in-place, before paint — and the reader/Talking Book highlights were verified intact by hand); `configureElementsForTools` can bind the keydown handler twice, in which case the second merge short-circuits; and the sweep can now run against a detached editable after the async-markup await, doing pointless but safe work. One of them suggested skipping the arrow keys in the keydown repair — that one is answered in a code comment (`d7a23b0a7f`), because it would give back the half of BL-16717 where the caret stops moving.

CI green (`pr-automation`). CodeRabbit is configured in this repo but with auto-review disabled, so it is not a reviewer here.

JohnThomson:
[Claude Opus 5 from John Thomson's machine during preflight] Consulted Devin again, up to commit `1cdfac7339` (a comment-only change recording why we don't turn ligatures off in the edit view).

**Nothing new.** The pass re-lists the same one Bug and one Investigate flag as before — both already have replied-to, resolved threads above — and the same 11 informational items, with no additions. As noted in the earlier log, the Bug re-report still carries the first pass's finding id and cites `editableDivUtils.ts:686-688`, which no longer holds that code; Devin's own informational item in this very pass places the guard at lines 703-723. Checked against the current file again rather than taking the re-report at face value.

CI green on this commit as well.

StephenMcConnel (approved):
> @StephenMcConnel reviewed 3 files and all commit messages, and resolved 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8217)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on JohnThomson).

PR #8218, opened 2026-08-19
Publish health signals so a Freeze Doctor can diagnose freezes, crashes and zombies (BL-16719)

JohnThomson:
## Problem

Users report that Bloom froze, and we get almost nothing to work with. Their problem report is written *after* they killed Bloom, so it describes a healthy new process, and the log only shows what Bloom managed to write before it stopped responding. Three quite different failures all arrive looking identical — the UI stops responding; Bloom exits without managing to report anything; or Bloom's window is gone while the process lives on, so the user cannot start Bloom again — and none of them currently leaves usable evidence. BL-16697 is the live example.

## Cause

Nobody is watching at the moment it happens, and the worst case cannot be watched from outside at all. A UI thread blocked in a managed wait on an STA thread keeps dispatching *sent* messages so COM still works, so the window answers probes, `IsHungAppWindow` reports it healthy and `Process.Responding` returns `true` while Bloom is completely stuck. Measured on a real Bloom: nine minutes frozen, reported responsive throughout. Since Bloom's UI thread awaits WebView2 constantly, that is likely the *common* shape of freeze rather than an exotic one.

## What this PR changes

This is Bloom's half of the work; the diagnostic tool itself lives in [BloomBooks/bloom-freeze-doctor](https://github.com/BloomBooks/bloom-freeze-doctor). Nothing here changes what Bloom does for a user, and all of it is inert on a machine where the Doctor is not installed.

- **Bloom publishes a health heartbeat** through a shared-memory page a watcher can read without Bloom's cooperation: a UI-thread beat (whose silence is the freeze signal that outside observation cannot see), a separate background beat so "the UI thread is blocked" can be told from "the whole process is wedged", what Bloom thinks it is doing, and how far shutdown has got.
- **Bloom proves its clean exits**, from a single `ProcessExit` handler. That runs for a normal return from `Main` and for `Environment.Exit`, and *not* for `FailFast`, `TerminateProcess` or an access violation — which is exactly the line worth drawing, and one no future exit path can forget to honour. A Doctor-requested exit is deliberately not recorded as clean.
- **Bloom records a session file** with the facts a watcher cannot reliably work out from outside: above all which log file this run is writing to (Bloom recreates `Log.txt` each run and falls back to a random name when another Bloom holds it, so guessing from the filesystem picks the wrong file in exactly the restart-after-a-freeze case), plus its ports, version, channel and collection.
- **Bloom tracks in-flight API requests**, so a report can say *which request* has been running for 47 seconds rather than only that a thread is waiting. Instrumented at the single inner dispatch point, with no locks and nothing that can throw.
- **Bloom starts the Doctor if it is installed**, because a diagnostic tool is no use unless it is already running when the trouble starts. No handshake: Bloom's only job is to ensure one is running. One directory check when it is not installed.
- **Bloom honours a request to exit** when its UI is gone and it is holding the single-instance token, and **asks for a dump as it crashes** — checking with a zero timeout whether anyone is listening first, so users without the Doctor never wait.
- **A freeze simulator for developer *and Alpha* builds**, inert unless `BLOOM_SIMULATE_FREEZE` is set *and* the channel is one we are willing to break deliberately, so the detection can be tested against a real Bloom instead of waiting for a real freeze. Alpha is included on purpose: reproducing a freeze usually means working with somebody who is actually having one, and those people are running Alpha, not a build from source. A stray environment variable on a Release or Beta machine does nothing at all, and three tests pin exactly which channels are allowed, so widening or narrowing that set breaks a test.
- **A check that the shared files have not drifted.** Three files here — the shared-memory layout, the session file and the named events — are copies of files in the Doctor's repository, and drift between them fails silently and expensively: Bloom writes one set of offsets, the Doctor reads another, and the resulting reports look plausible and are wrong. A PR-triggered workflow compares them against the Doctor's repository. It compares them *ignoring the namespace line and all whitespace*, because they are **not** byte-identical — the namespaces necessarily differ, and csharpier formats this copy — and it reads the Doctor's repo with a shallow clone rather than `raw.githubusercontent.com`, whose CDN serves stale content for minutes after a push.

Also: `BloomServer` gains a `BusyWorkerCount` beside the existing `BlockedWorkerCount`, and the `RobustFile` pre-commit check exempts the two of those shared files that do file I/O — they cannot call `RobustFile`, because it does not exist in the Doctor's repository. Every call in them is a best-effort diagnostic write already wrapped in a catch-everything, and the failure `RobustFile` would retry through just means "no session file this time", which the callers are built to tolerate.

## The duplication is temporary

Keeping three files in step across two repositories by hand is not the intended end state, and the drift check exists only because they are copies. **#8219 removes the duplication**: the Doctor publishes them as a NuGet package (`BloomBooks.FreezeDoctor.Protocol`) and Bloom references it, which deletes the copies, the drift check, its workflow, and the `RobustFile` exemption above. That PR is stacked on this one and can be read for shape now; it is not mergeable until the package is actually published.

Ref: https://issues.bloomlibrary.org/youtrack/issue/BL-16719

[Devin review](https://devinreview.com/BloomBooks/BloomDesktop/pull/8218)

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8218)


src/BloomExe/FreezeDoctor/DoctorChannel.cs
```diff
@@ -0,0 +1,401 @@
...
+        {
+            view.Write(DoctorChannelLayout.OffsetWriteSequence, ++_writeSequence);
+            update(view);
+            view.Write(DoctorChannelLayout.OffsetWriteSequence, ++_writeSequence);
+        }
+        catch (Exception)
+        {
+            // Never let publishing state break the thing whose state we are publishing.
+        }
+    }
```
JohnThomson, line 453: [Claude Opus 5] [Devin] **Bug**: Health data published for the diagnostic tool can become permanently unreadable when two threads report at once

Bloom's health page is updated without any locking or atomic counter (`++_writeSequence` at `src/BloomExe/FreezeDoctor/DoctorChannel.cs:385-387`) even though two different threads publish to it at the same time, so the reader can be handed mixed-up data or shut out of the page altogether.
Impact: A frozen Bloom can be diagnosed with garbled information, or the diagnostic tool may decide no health information is being published at all and fall back to guessing from outside.

<details>
<summary>Seqlock protocol broken by unsynchronised increments from the UI timer and the watchdog thread</summary>

`FreezeDoctorSupport` publishes from two threads concurrently: the WinForms timer calls `RecordUiTick()` every 500 ms on the UI thread (`src/BloomExe/FreezeDoctor/FreezeDoctorSupport.cs:104-109`, `552-559`) while the watchdog thread calls `RecordWatchdogTick()`, `SetActivity()` and `SetServerWorkerCounts()` every second (`src/BloomExe/FreezeDoctor/FreezeDoctorSupport.cs:380-444`). `SetActivity`/`SetLongOperation` are also documented as callable from any thread.

All of these funnel through `Write`, which relies on a seqlock: increment the sequence (odd = write in progress), update, increment again (even = settled). Two problems arise with concurrent callers:

1. `++_writeSequence` on a `long` field is a non-atomic read-modify-write. Lost updates can leave the resting value ODD, in which case `DoctorChannelReader.TryRead` sees `before % 2 != 0` on every attempt and returns false forever (`src/BloomExe/FreezeDoctor/DoctorChannel.cs:185-200`) — the channel appears absent for the rest of the run.
2. Even without a lost update, thread A can be mid-`update` while thread B completes its own pair of increments, so a reader sampling before/after sees the same EVEN value and accepts a state that is half A's write and half B's — exactly the torn activity string the sequence number was added to prevent.

A simple fix is to serialise the whole of `Write` under a private lock (the callers are all diagnostic and hold no other lock) and/or use `Interlocked.Increment` for the counter.

</details>

src/BloomExe/FreezeDoctor/FreezeDoctorSupport.cs
```diff
@@ -0,0 +1,570 @@
...
+        /// Computed on this thread rather than in the request path, so the cost falls on a once-a-second
+        /// thread instead of on every request.
+        /// </summary>
+        private static void PublishWhatBloomIsDoing()
+        {
+            try
+            {
+                var activity = ApiActivityTracker.DescribeCurrentActivity();
+                if (activity != null)
+                    _channel?.SetActivity(activity);
```
JohnThomson: [Claude Opus 5] [Devin] **Bug**: Diagnostic report can name a request that finished long ago as still running

The description of what Bloom is doing is only ever overwritten while a slow request is in progress (`SetActivity` only when non-null at `src/BloomExe/FreezeDoctor/FreezeDoctorSupport.cs:435-437`) and is never cleared once the work finishes, so the published state keeps naming a request that has already completed.
Impact: A freeze report can blame a piece of work that actually finished minutes earlier, sending diagnosis down the wrong path.

<details>
<summary>Activity field is write-only-when-interesting, never reset to idle</summary>

`ApiActivityTracker.DescribeCurrentActivity()` returns null when nothing has been in flight for more than 2 seconds (`src/BloomExe/FreezeDoctor/ApiActivityTracker.cs:94-111`). `PublishWhatBloomIsDoing` only publishes when the value is non-null, so the last non-null value — e.g. "api api/publish/epub/save running 47s on thread 12" — remains in the shared page indefinitely after the request completes. Nothing else in the codebase calls `FreezeDoctorSupport.SetActivity` after startup, so there is no other path that would overwrite it.

Publishing something like "idle" (or the startup string) when `DescribeCurrentActivity()` returns null would keep the field honest.

</details>

src/BloomExe/FreezeDoctor/FreezeDoctorSupport.cs
```diff
@@ -0,0 +1,570 @@
...
+                {
+                    Exit = new DoctorSessionExit
+                    {
+                        AtUtc = DateTimeOffset.UtcNow,
+                        ShutdownPhase = _shutdownPhase,
+                        BloomAlreadyReported = true,
+                        ReportedId = reportedId,
+                    },
+                };
+                DoctorSessionStore.TryWrite(_session);
```
JohnThomson: [Claude Opus 5] [Devin] **Bug**: A still-running Bloom is recorded on disk as having already ended

When the user files a problem report, Bloom writes an "how this run ended" record for itself (`Exit = new DoctorSessionExit ...` at `src/BloomExe/FreezeDoctor/FreezeDoctorSupport.cs:232-241`) even though it is still running, so anything reading that file believes the run is over.
Impact: A Bloom that later freezes or crashes may be skipped by the diagnostic tool because its file already claims the run finished.

<details>
<summary>Exit is contractually "null while Bloom is running"</summary>

The contract states `Exit` is "Null while Bloom is running — and null *after* Bloom has gone is itself the evidence that it did not shut down properly" (`src/BloomExe/FreezeDoctor/DoctorSession.cs:75-79`). `NoteBloomReportedAProblem` is called from the successful YouTrack submission path (`src/BloomExe/web/controllers/ProblemReportApi.cs:363-367`), which happens while Bloom continues to run — a user can submit a non-fatal report and carry on working. From then on the session file records an exit (with a shutdown phase that has not happened) for a live process, and `WriteSessionExit` deliberately refuses to correct it at real shutdown (`src/BloomExe/FreezeDoctor/FreezeDoctorSupport.cs:299-300`).

If the intent is only "Bloom already reported a problem", that flag belongs on the session itself rather than inside the exit record, so a live process is not described as finished.

</details>

src/BloomExe/FreezeDoctor/FreezeDoctorSupport.cs
```diff
@@ -0,0 +1,570 @@
...
+                    HttpPort = httpPort,
+                    CdpPort = httpPort > 0 ? Api.BloomServer.RemoteDebuggingPort : 0,
+                    CollectionName = SafeCollectionName(),
+                };
+                // Only touch the disk when something actually changed: this runs every ten seconds for the
+                // life of the process, and a pointless write every ten seconds is a pointless write.
+                if (refreshed == _session)
+                    return;
+                _session = refreshed;
+                DoctorSessionStore.TryWrite(_session);
```
JohnThomson: [Claude Opus 5] [Devin] **Bug**: Note that a problem was already reported can be silently lost, allowing a duplicate report

The record of what Bloom knows about itself is read, modified and written back on two different threads without coordination (`_session with { ... }` at `src/BloomExe/FreezeDoctor/FreezeDoctorSupport.cs:460-472`), so a periodic refresh can overwrite and discard the note that a problem was just reported.
Impact: The diagnostic tool may file a second report about a problem the user has already reported.

<details>
<summary>Lost update between the watchdog's RefreshSessionFile and NoteBloomReportedAProblem</summary>

`RefreshSessionFile` runs on the watchdog thread every ten seconds and does an unsynchronised read-modify-write of the static `_session` (`src/BloomExe/FreezeDoctor/FreezeDoctorSupport.cs:453-472`). `NoteBloomReportedAProblem` does the same from the request/UI thread (`src/BloomExe/FreezeDoctor/FreezeDoctorSupport.cs:226-243`). If the refresh reads `_session` before the note is applied and writes its derived copy afterwards, the `Exit`/`BloomAlreadyReported` information is dropped from both the in-memory record and the file, which is precisely the information the Doctor uses to stay quiet.

Guarding the mutation of `_session` (a small lock, or a single-threaded owner for all updates) would remove the race.

</details>

src/BloomExe/FreezeDoctor/FreezeDoctorSupport.cs
```diff
@@ -0,0 +1,570 @@
...
+            // The session file goes first, and happens even if the shared-memory channel cannot be
+            // created: it is what a Doctor reads about a Bloom that has already died, so it is the more
+            // important of the two for the case where nobody was watching at the time.
+            WriteSessionFile();
+
+            try
+            {
+                _channel = new DoctorChannelWriter(Process.GetCurrentProcess().Id);
+                if (!_channel.IsOpen)
+                    return; // Another process owns the name, or the OS refused. Nothing more to do.
```
JohnThomson, line 127: [Claude Opus 5] [Devin] **Investigate**: If the shared page cannot be created, no clean-exit record is ever written to disk either

`Start()` returns as soon as `_channel.IsOpen` is false, which skips the UI heartbeat, the watchdog thread, the session pruning, AND the `ProcessExit` hook that writes `WriteSessionExit()`. Since the session file has already been written at that point with `Exit == null`, a Doctor reading it later would conclude this run did not shut down properly — the exact false positive the design tries to avoid. The channel name embeds the pid so creation failure should be rare, but the on-disk half of the contract does not depend on shared memory and arguably should be set up (including the ProcessExit hook) before this early return.

JohnThomson: [Claude Opus 5 from Thomson's machine during preflight] Fixed in `c69b7e53ad`. This was the sharpest of the five, because the failure was a *false positive generator*: returning early meant the session file sat there for ever with no exit record, which the Doctor reads as `this run did not shut down properly` - manufacturing exactly the conclusion the clean-exit proof exists to prevent anyone drawing.

The `ProcessExit` hook is now installed before anything can return early, so the on-disk half of the contract no longer depends on the shared page being creatable.

src/BloomExe/FreezeDoctor/FreezeDoctorSupport.cs
JohnThomson: [Claude Opus 5 from Thomson's machine during preflight] Fixed in `c69b7e53ad`. All mutations of the session now compose on the current value inside a single private lock, so the ten-second refresh can no longer discard a note written a moment earlier by another thread. Same class of problem as the seqlock one above, and the same remedy.

src/BloomExe/FreezeDoctor/FreezeDoctorSupport.cs
JohnThomson: [Claude Opus 5 from Thomson's machine during preflight] Fixed in `c69b7e53ad`, and you identified the right fix - the flag belonged on the session, not inside the exit record. `BloomAlreadyReported`/`ReportedId` have moved onto `DoctorSession` itself, and `WriteSessionExit` no longer refuses to record the real exit.

You also spotted the reason it mattered: a user can file a problem report and then work on for hours, so writing an exit there described a live process as finished - which the Doctor reads as proof of an orderly shutdown, for a process that might still crash. Both copies of the contract file are updated, and there are now tests asserting that recording the note leaves `Exit` null, and that an exit the Doctor forced is marked as forced rather than looking clean.

src/BloomExe/FreezeDoctor/DoctorChannel.cs
JohnThomson: [Claude Opus 5 from Thomson's machine during preflight] Fixed in `c69b7e53ad`. You were right, and the consequence was the serious part: a lost update leaves the counter odd, every reader then reads that as `a write is in progress` and gives up, and the channel is silently dead for the rest of the run - which would disable the one signal this whole feature exists to provide, with nothing to show why.

The body of `Write` is now serialised on a private lock (only the two diagnostic callers ever take it, and neither holds another lock, so it cannot deadlock what we are diagnosing). I also went one step further than suggested: the sequence is now left even in a `finally`, so a write whose *update* throws can no longer strand it either. That turned out to matter immediately - a UTF-8 truncation fix I made in the same pass read one byte past the end of the array, and this exact failure mode took the channel out until the `finally` was added. There is a regression test for it now.

src/BloomExe/FreezeDoctor/FreezeDoctorSupport.cs
JohnThomson: [Claude Opus 5 from Thomson's machine during preflight] Fixed in `c69b7e53ad`, with your suggested change in substance: the field is now written on every pass, publishing `no request in flight` when nothing is running.

Worth saying why this one deserved fixing rather than noting: a report that blames a request which finished minutes ago is worse than a report with no lead at all, because someone will go and investigate it.

src/BloomExe/FreezeDoctor/FreezeDoctorSupport.cs
```diff
@@ -0,0 +1,598 @@
...
+        /// </summary>
+        private static void PublishWhatBloomIsDoing()
+        {
+            try
+            {
+                // Publish "idle" rather than leaving the last interesting value in place. Otherwise the
+                // field keeps naming a request that finished minutes ago, and a freeze report blames work
+                // that had already completed — a wrong lead is worse than no lead.
+                var activity = ApiActivityTracker.DescribeCurrentActivity();
+                _channel?.SetActivity(activity ?? "no request in flight");
```
JohnThomson: [Claude Opus 5] [Devin] **Bug**: Anything Bloom says about what it is doing is wiped out within a second

The description of what Bloom is doing is replaced once a second by a generic request-derived line (`SetActivity(activity ?? "no request in flight")` at `src/BloomExe/FreezeDoctor/FreezeDoctorSupport.cs:460`), so a freeze report names "no request in flight" instead of the work that was really under way.

Impact: Diagnostic reports lose the most human-readable clue about what Bloom was busy with, including during startup and during long operations that are not driven by an API request.

<details>
<summary>How the once-a-second overwrite defeats the public SetActivity entry point</summary>

`FreezeDoctorSupport.SetActivity` (`src/BloomExe/FreezeDoctor/FreezeDoctorSupport.cs:174-181`) is a public, documented way for Bloom code to publish text such as "Publishing to BloomPUB" or "Saving Foo.htm". But the watchdog thread calls `PublishWhatBloomIsDoing` every `WatchdogInterval` (1s) and unconditionally writes over the same single activity slot with either `ApiActivityTracker.DescribeCurrentActivity()` or the literal `"no request in flight"`.

Consequences:
- The `"starting up"` value written in `Start` (`src/BloomExe/FreezeDoctor/FreezeDoctorSupport.cs:117`) is replaced within one second, so a Bloom that wedges during startup publishes "no request in flight".
- Any future caller of `SetActivity` (the API exists precisely for this) has its message survive for less than a second, which makes the entry point effectively inoperative.
- `ApiActivityTracker.DescribeCurrentActivity()` also returns null until a request has been running 2 seconds (`InterestingDuration`, `src/BloomExe/FreezeDoctor/ApiActivityTracker.cs:38`), so even genuinely busy periods can read as "no request in flight".

A fix would be to keep the caller-supplied activity separately (with a timestamp or precedence rule) and only fall back to the request-derived text when there is no caller-supplied activity, rather than blindly clobbering one with the other.

</details>

src/BloomExe/FreezeDoctor/FreezeDoctorSupport.cs
```diff
@@ -0,0 +1,598 @@
...
+                // Also on disk, because the shared-memory page vanishes once the last handle closes and a
+                // Doctor may not have been watching. The file is what a Doctor started tomorrow will read.
+                WriteSessionExit();
+            }
+            catch (Exception)
+            {
+                // Losing the proof means the Doctor may report an exit that was in fact clean. Annoying,
+                // but far better than delaying or breaking Bloom's shutdown to avoid it.
+            }
+        }
```
JohnThomson, line 274: [Claude Opus 5] [Devin] **Investigate**: Bloom's own fatal-error exits are recorded as clean shutdowns

`RecordCleanExit` is wired to `AppDomain.ProcessExit`, so it also fires for `Environment.Exit(1)` calls that are NOT orderly shutdowns — e.g. `src/BloomExe/WebView2Browser.cs:145` and `:186` (WebView2 could not be initialised) and the non-message-loop branch of `src/BloomExe/FatalExceptionHandler.cs:105`. Those runs get `CleanExitRecorded` plus an on-disk `Exit` record (with `ShutdownPhase` still 0, since none of Program.Run's phase markers ran), which the Doctor reads as "this run shut down properly". Only the Doctor-requested exit is excluded via `_endedAtDoctorsRequest`. The PR states that treating every `Environment.Exit` as clean is the intended line, so this is not reported as a bug, but the recorded `ShutdownPhase` of 0 is the only hint that the shutdown sequence never ran; a reviewer may want the same kind of flag used for hard-failure exits.

src/BloomExe/FreezeDoctor/FreezeDoctorSupport.cs
JohnThomson: [Claude Opus 5 from Thomson's machine during preflight] Fixed in `01ccbd84fb`, and thank you - this was a regression I introduced in the previous round while fixing the opposite complaint (that the field was never cleared). You are right that overwriting it every second made the public `SetActivity` entry point inoperative, and right about which case that hurts most: a Bloom wedging during startup would have published `no request in flight` instead of `starting up`, which is both wrong and the least useful thing it could say.

The watchdog now composes the two sources rather than letting either clobber the other: `SetActivity` remembers what it was told, and the once-a-second publish joins that with the in-flight request text when both exist, falling back to whichever is present. Both facts are worth having and they no longer compete for the slot.

src/BloomExe/FreezeDoctor/FreezeDoctorSupport.cs
JohnThomson: [Claude Opus 5 from Thomson's machine during preflight] Fixed in `01ccbd84fb`. Filed as Investigate rather than a bug, but it was a real false negative and worth acting on: a Bloom that died because WebView2 would not initialise was being recorded as an orderly shutdown, so the tool built to notice that failure would have passed over it in silence.

Rather than flagging each hard-failure exit site, the test is now whether the shutdown sequence actually ran. `Program.Run`'s phase markers are only reached on the orderly path, so a recorded phase of zero means it was not walked - which covers `WebView2Browser`'s exits, the fatal handler's non-message-loop branch, and any hard-failure exit added later without anyone having to remember this conversation. Pruning also no longer treats a forced exit as `explained`, since a forced exit is not an explanation, it is the evidence.

JohnThomson:
[Claude Opus 5 from Thomson's machine during preflight] Consulted Devin on 2026-08-19 16:43 up to commit `01ccbd84fb`.

Three rounds. The first found **4 bugs and 1 Investigate flag**, all genuine and all in this branch's new code; the second found **1 new bug** (a regression I introduced fixing one of the first four) and **1 new Investigate flag**. Every one of those six has its own thread above with a reply saying what changed, and is resolved. Also acted on five of the informational items (UTF-8-safe truncation, a timer that could be collected before firing, the initial session write taking its own lock, pruning no longer discarding forced-exit evidence, and two wrong comments).

One thing worth knowing for anyone reading the review page: on the second and third rounds Devin **re-emitted its earlier findings verbatim** rather than re-checking them. Each was verified against the analysed commit and is fixed — the clearest case quotes `SetActivity(activity ?? "no request in flight")` at lines that, in the very commit it analysed, contain the watchdog's catch block. So the current review page overstates what is outstanding; the resolved threads above are the accurate record.

CI (`pr-automation`) passed. The full .NET suite is green at this commit (3149 passed, 13 skipped). CodeRabbit is configured on this repo but its free monthly allowance is exhausted, so its silence here is expected rather than meaningful.

src/BloomExe/FreezeDoctor/FreezeDoctorSupport.cs
```diff
@@ -0,0 +1,637 @@
...
+            // the false positive the proof exists to prevent.
+            AppDomain.CurrentDomain.ProcessExit += (sender, args) => RecordCleanExit();
+
+            try
+            {
+                _channel = new DoctorChannelWriter(Process.GetCurrentProcess().Id);
+                if (!_channel.IsOpen)
+                    return; // Another process owns the name, or the OS refused. Nothing more to do.
+
+                _channel.SetActivity("starting up");
```
JohnThomson: [Devin — Bug, current commit] **The "starting up" note is erased a second later, so a freeze during startup is reported with the least useful message.**

`Start()` writes "starting up" straight into the shared page instead of going through `SetActivity`. Only the wrapper remembers the text; the watchdog's once-a-second refresh composes from what it remembers, finds nothing, and publishes "no request in flight" instead.

Why it matters: a Bloom that wedges *during startup* is the one case where this string is the only clue there is — no request has run yet, so there is nothing else to report. That case published the least helpful thing it could.

This is the surviving half of an earlier finding on this PR. The composition was fixed; the initial value was still bypassing it, so in practice nothing had changed for the startup case.

Devin is right. Fixed in d7f5f597af.

src/BloomExe/FreezeDoctor/FreezeDoctorSupport.cs
```diff
@@ -0,0 +1,637 @@
...
+                // worse than no diagnostics at all.
+                TryLog("Freeze Doctor channel could not be opened", e);
+            }
+        }
+
+        /// <summary>
+        /// Says what Bloom is doing, in words fit to appear at the top of a bug report — "Publishing to
+        /// BloomPUB", "Saving Foo.htm". Truncated if very long; safe to call from any thread.
+        /// </summary>
+        public static void SetActivity(string activity)
```
JohnThomson, line 192: [Devin — Investigate, current commit] **`SetActivity` and `SetLongOperation` are defined but never called from Bloom.**

Confirmed, and the second one has a real cost. The Freeze Doctor reads `LongOperationInProgress` and uses it to raise its freeze threshold from one minute to five — the mechanism that stops a legitimately slow operation being filed as a freeze. Nothing in Bloom ever sets it, so it is permanently false and **that patience does not exist**: every freeze is judged at one minute.

What that does and doesn't threaten:

- Work behind a modal progress dialog is safe either way. `ShowDialog` runs a nested message loop, so the UI-thread heartbeat keeps ticking and no freeze is detected at all.
- Work that blocks the UI thread for over a minute *without* pumping messages would be filed as a freeze. That is the false positive the flag exists to prevent.

It is left deliberately unwired rather than guessed at, with the gap written up at the method so nobody assumes the grace works. Choosing which operations to mark is a judgement about how Bloom really behaves, and it cannot be inferred: "a request has been running a minute" is the *same signal* as the freeze itself, so no automatic rule can tell legitimately-slow from wedged. It also spans several progress/publish paths rather than one chokepoint.

Raised with the developer as the one open decision on this PR. Thread stays open until that comes back.

`SetActivity` having no callers is fine and needs nothing — it is a public entry point for Bloom code to use, and `Start()` is now a caller as of d7f5f597af.

src/BloomExe/FreezeDoctor/FreezeDoctorSupport.cs
```diff
@@ -0,0 +1,637 @@
...
+
+        /// <summary>
+        /// How long an unexplained session file is kept. Long enough that a Doctor installed the day after
+        /// a crash can still find out about it, short enough not to accumulate.
+        /// </summary>
+        private static readonly TimeSpan SessionRetention = TimeSpan.FromDays(7);
+
+        /// <summary>
+        /// Starts publishing. Call once, on the UI thread, immediately before entering the message loop:
+        /// the UI heartbeat is a WinForms timer, so it only ticks once messages are being pumped, which is
```
JohnThomson, line 102: [Devin — Investigate, current commit] **"If the shared page cannot be created, no clean-exit record is ever written to disk either."**

Not the case — this is the one thing the ordering here was written to guarantee, and the code does it:

- `WriteSessionFile()` runs *before* the shared page is attempted, so the on-disk record exists regardless.
- `AppDomain.CurrentDomain.ProcessExit += … RecordCleanExit` is installed *before* the `if (!_channel.IsOpen) return;` bail-out, so the early return cannot skip it.
- `RecordCleanExit()` calls `_channel?.RecordCleanExit()` — null-safe — and then `WriteSessionExit()` unconditionally.

So a Bloom that fails to create the shared page still writes its exit record. That matters precisely because getting it wrong would manufacture the false positive the proof exists to prevent: a session file with no exit record is what a Doctor reads as "this run did not shut down properly", so a Bloom that merely lost a race for the shared-memory name would have been reported as an unreported crash.

No change needed. Resolving.

src/BloomExe/FreezeDoctor/FreezeDoctorSupport.cs
```diff
@@ -0,0 +1,664 @@
...
+                if (!_channel.IsOpen)
+                    return; // Another process owns the name, or the OS refused. Nothing more to do.
+
+                // Through the static wrapper, NOT _channel directly: the wrapper is what remembers the text
+                // in _statedActivity, and the watchdog composes from that. Writing straight to the channel
+                // here left _statedActivity empty, so the first watchdog tick (one second later) replaced
+                // "starting up" with "no request in flight" — and a Bloom that wedges during startup, the one
+                // case where this string is the only clue there is, reported the least useful thing it could.
+                SetActivity("starting up");
+                // Authoritative, and worth more than the Doctor's own guess: it is why a developer
```
JohnThomson, line 136: [Devin — Bug, current commit] **The activity line shows "starting up" permanently, long after startup.**

Right, and this is the exact opposite of the finding above it — which is why this one line has now been got wrong three times in three different ways:

1. the refresh overwrote what Bloom stated, so "starting up" survived less than a second;
2. "starting up" went straight into the shared page, so there was nothing for the refresh to carry forward — same symptom, different cause;
3. it was carried forward for ever, so a freeze at 4pm was reported as a freeze during startup.

The two failures point in opposite directions, so a fix for either one walks into the other.

Bloom states "starting up" once and has no natural "finished starting" moment to clear it at — `Start()` runs immediately before `Application.Run`, and there is no later hook meaning "now interactive". Handling an API request *is* that moment: the UI is up and talking to the server. So `ApiActivityTracker` now exposes `HasHandledARequest` and the composer retires the startup note once it is true. Reading the condition rather than requiring a call means no future startup path has to remember to announce itself. Any other stated activity is left alone — a long publish must keep saying so until its caller says otherwise.

The composition is now a pure `Compose(stated, request, hasHandledARequest)` with no static state, and both directions are pinned by a test. Worth noting the first version of that test was itself unsound: it asserted whichever outcome the ambient static state happened to produce, so it would have passed whatever the rule did.

Fixed in 67b488d03b.

src/BloomExe/FreezeDoctor/DoctorChannel.cs
```diff
@@ -0,0 +1,442 @@
...
+    private void Write(Action<MemoryMappedViewAccessor> update)
+    {
+        var view = _view;
+        if (view == null)
+            return;
+        try
+        {
+            lock (_writeLock)
+            {
+                view.Write(DoctorChannelLayout.OffsetWriteSequence, ++_writeSequence);
```
JohnThomson: [Devin — Bug, current commit] **The health channel can permanently disable itself if a status write fails at the wrong moment.**

Correct, and this is the worst failure mode this file has, so it is worth spelling out.

The sequence counter was incremented in the same statement that published it, from *outside* the inner `try`:

```csharp
view.Write(OffsetWriteSequence, ++_writeSequence);
try { update(view); } finally { view.Write(OffsetWriteSequence, ++_writeSequence); }
```

`++` happens before the call, so a throw from that one write left the counter odd with no `finally` to correct it. From then on the parity was **inverted for the rest of the run**: every write published an even value while the write was in progress and came to rest on an odd one. Readers treat odd as "a write is in progress" and give up — so the channel silently switched itself off, the Doctor fell back to watching Bloom from outside, and watching from outside is blind to precisely the freeze this whole mechanism exists to catch. Nothing anywhere would have said why.

This is the second time this one counter has produced that exact outcome (the first was the non-atomic increment, fixed by `_writeLock`). The pattern worth noting: the invariant was being maintained by careful statement ordering, which is fragile in proportion to how bad breaking it is.

Now every increment is inside the `try`, and the `finally` restores parity from whatever value was actually reached rather than assuming one more increment is right. Two properties follow: no ordering of failures can leave the counter odd, and a failure that stops even the final `view.Write` is repaired by the next successful write instead of lasting for ever.

Applied identically to the copy in `BloomBooks/bloom-freeze-doctor`, where this file's source of truth lives.

**Not unit tested directly, deliberately.** Reaching it needs the memory-mapped view itself to fail mid-write, which cannot be injected without adding test surface to a file whose whole purpose is being the same in both repos. The existing tests do cover an update delegate that throws, and would fail if the parity were inverted.

Fixed in 20707c113f.

JohnThomson:
[Claude Opus 5 from Thomson's machine during preflight]

Consulted Devin up to `20707c113f` — four further rounds since the last log (`3b165a1556`, `d7f5f597af`, `67b488d03b`, `20707c113f`). The final round reports **no outstanding bugs**: all eight are marked resolved against the commits that fixed them.

**Three real bugs, all fixed, each with its own thread.** All three were in this branch's new code, and none would have been caught by testing — a freeze report would simply have said something false.

- The "starting up" note was erased a second later, so a Bloom that wedged during startup — the one case where that string is the only clue there is — reported "no request in flight". Fixed in `d7f5f597af`.
- Fixing that made the note permanent instead, so a freeze at 4pm would have been reported as a freeze during startup. Fixed in `67b488d03b`.
- The seqlock counter was incremented outside the inner `try`, so one throw inverted its parity for the rest of the run — readers then treat every resting value as "a write is in progress" and give up, silently switching the channel off. The Doctor falls back to watching from outside, which is blind to exactly the freeze this exists to catch, with nothing saying why. Fixed in `20707c113f`, and identically in `bloom-freeze-doctor`.

Two patterns in there worth a reviewer's attention, both about invariants held up by statement ordering:

- The activity line has now been got wrong **three times in three ways**, and its two failure directions are *opposites*, so a fix for either walks into the other. It is now a pure `Compose(stated, request, hasHandledARequest)` with both directions pinned. The first version of that test was itself unsound — it asserted whichever outcome the ambient static state happened to produce, so it would have passed whatever the rule did.
- The seqlock counter has now silently disabled the channel **twice** (first the non-atomic increment, fixed by `_writeLock`). Both times the invariant depended on getting statement order right, which is fragile in proportion to how badly it fails. Parity is now restored in a `finally` from whatever value was actually reached.

**One flag refuted** — "a Bloom that cannot create the shared page never writes a clean-exit record". It does: the session file is written before the shared page is attempted, and the `ProcessExit` hook is installed before the bail-out. Getting this wrong would have manufactured the very false positive the proof exists to prevent, so it is worth being sure of. Reasoning is on the thread; resolved.

**One flag open, and it is the decision on this PR** — `SetLongOperation` has no callers, so the Doctor's five-minute grace for legitimately slow work does not exist and every freeze is judged at one minute. Left open deliberately.

Finally, so the review page is not misread: **Devin re-emits earlier findings verbatim on re-review rather than re-checking them.** Several Informational flags across these rounds are demonstrably false against the very commit they were raised on — a pruning rule they describe is the opposite of what the code does, a comment they call contradictory is correct and explains itself, and a list of simulator kinds they call mismatched is complete. Each was checked individually rather than assumed. The resolved threads, not the flag count, are the accurate record.

CI (`pr-automation`) green on every round. Full .NET suite run after each, green each time; at `20707c113f`: **3153 passed, 13 skipped, 0 failed**. Freeze Doctor repo: 95 passed. CodeRabbit is silent because its free monthly allowance is exhausted, not because of a config problem.

JohnThomson:
Superseded by #8229, which contains this work in its final form.

The Freeze Doctor moved into the BloomDesktop repository rather than living in its own, and lost its separate installer — it now ships inside Bloom's installer and is switched on by a "Run Freeze Doctor" item in the collections-tab debug menu. Everything on this branch is in #8229, together with the Doctor itself.

Closing rather than merging: the two would conflict, and #8229 is the shape that was reviewed.

The review history here is worth keeping — the eight bugs Devin found on this branch, and the three-times-wrong activity line, are all recorded above.

PR #8219, opened 2026-08-20
Reference the Freeze Doctor contract as a package instead of copying it (BL-16719)

JohnThomson:
## Read this for shape, not to merge yet

**Stacked on #8218** — review that one first; this branch's base is that branch, so its diff shows only the migration.

**Nothing is published.** The package is restored from a local folder, so this branch is buildable and reviewable before anything irreversible happens. Companion PR: BloomBooks/bloom-freeze-doctor#1.

## Why

Three files describing the wire format between Bloom and the Freeze Doctor — the shared-memory layout, the session file, the named events — were **750 lines of hand-maintained duplicates** across two repos. They had already drifted, and a fix in one had to be carried across the other by hand.

Drift there fails silently and expensively: Bloom writes one set of offsets, the Doctor reads another, and the reports look plausible and are wrong.

## What this removes

`193 insertions, 1095 deletions.` Three things go away, and the third is the one worth noting:

- **The three copied files.**
- **The drift check and its workflow** — they existed only to catch divergence between copies. With one definition there is nothing to diverge.
- **The `RobustFile` pre-commit exemption.** It was only needed because the copies couldn't call `RobustFile` and still match the originals. The files have left this repo, so that shared quality gate is back to full strength with nothing excused from it — the check passes on its own merits. This is the exemption that needed approving in the last preflight; it's now moot.

## What arrives

One `PackageReference` on **`BloomBooks.FreezeDoctor.Protocol`** — with **no transitive dependencies at all**, just a 26KB assembly, which is most of what makes this reasonable to add to Bloom — two `using` lines, and temporary local-feed scaffolding (`NuGet.Config` at the root, `build/pack-freeze-doctor-protocol.ps1`, `localpackages/`) that all disappears when the package is published.

## To build it

```
build/pack-freeze-doctor-protocol.ps1
```

It packs the protocol from a clone of the Doctor's repo beside this one (or `BLOOM_FREEZE_DOCTOR_REPO`) into `localpackages/`. That project only exists on the Doctor repo's `contract-package` branch until it merges.

## Why "Protocol" and not "Contract"

These files are not a contract in any sense a reader would recognize from the word — no interface, no assertions, nothing enforced. They are the **wire format** of the shared-memory channel between two processes: byte offsets, a schema version, event names. "Protocol" says that, so the package, its namespace, its folder in the Doctor's repo, and the test fixture here all say it.

The `BloomBooks.` prefix follows the house style already in production: `BloomBooks.Velopack` is a Bloom-only package published the same way (nuget.org, then unlisted — all 11 versions currently unlisted, 0 search hits), which also settles the earlier open question about a `SIL.` prefix. There isn't one to seek.

## The pinned test, with its purpose rewritten — and a hole in it closed

`FreezeDoctorProtocolTests` (was `DoctorChannelContractTests`) used to guard against two copies diverging. That reason is gone, but it still earns its place for a different one: it pins the layout Bloom *expects* against the layout the referenced version actually *has*. A package upgrade that moved an offset would otherwise be silent — Bloom would compile, run, and publish its health where the Doctor no longer looks.

**It wasn't actually doing that.** It pinned `SchemaVersion`, `Size`, `ActivityMaxBytes` and the section name — not a single field offset. So the one failure it existed to prevent was the one it did not cover, and it couldn't have: the offsets are `internal` to the package. It now works from a `Fields` table the layout publishes about itself, and asserts every field's name, offset and size by value. Its comment says explicitly *not* to update the numbers to make it pass — a moved field means every Doctor already installed is misreading Bloom, so the numbers are the symptom.

Two tests join it. One pins the *invariant* rather than the numbers, which is what lets the first be strict without becoming a nuisance: `PayloadBytes` must be exactly the end of the last field, and at least the frozen generation-1 baseline. The other checks Bloom's extent actually reaches the shared page rather than trusting the constant.

## `PayloadBytes`, and why the version moved to 0.2.0-alpha

The protocol now records how far the writer wrote. This is Bloom's side of [bloom-freeze-doctor@aaec915](https://github.com/BloomBooks/bloom-freeze-doctor/commit/aaec915), and for Bloom it is only a reference bump — Bloom writes the same fields either way.

Worth knowing why it exists, since it changes how this format is allowed to evolve: adding a field is the likeliest change, and until now there was no way to make one. `SchemaVersion` is part of the section's *name*, so bumping it makes an old Doctor stop finding the channel altogether rather than merely ignoring a field it doesn't know. With an extent recorded, fields can be appended within a generation, and a **newer Doctor reading an older Bloom** can tell a field that Bloom never wrote from a real zero — the page is zero-filled, so without it a report would confidently say "0 blocked workers" where the truth is "this Bloom is too old to say". That direction is the common one in the field, not a curiosity: the Doctor updates itself through Velopack while Bloom versions linger, and the 6.4 backport guarantees more than one vintage writing this page.

Several offsets moved to make room for it in the header. **That was free only because nothing is published yet**, which is a reason to settle the package id and get it published rather than to keep the local-folder arrangement going.

The version bump is not cosmetic: NuGet caches by id *and* version, so re-packing `0.1.0-alpha` with a changed layout would let a machine that had already restored it keep the old assembly, with nothing looking wrong. Bump the version for any layout change, even while restoring from a local folder.

## Decided, so not up for review here

The feed: **publish to nuget.org, then unlist the version** — absent from search and browse, still restorable when a consumer pins an exact version, and no feed credentials anywhere. A private feed was considered and rejected: it hides nothing (the Doctor's repo is public MIT, so the package is a compiled copy of readable source) and would cost credentials on every dev machine, in TeamCity, and in CI. Reasoning lives in the protocol `.csproj`.

The id, likewise: `BloomBooks.FreezeDoctor.Protocol`. An id on nuget.org is claimed permanently, which is exactly why it is being reviewed before anything is pushed rather than after.

## Still open

- **Who owns the nuget.org API key** for the Doctor's repo. `BloomBooks.Velopack`'s fork uses a repo-level `BLOOMBOOKS_NUGET` secret; whether that is an org-level key we can reuse couldn't be checked from here. Note the gotcha for whoever holds it: pushing a *new* package id needs a key scoped to "Push new packages and package versions", not just new versions of existing ones.

## Verified

Bloom builds against the package with the copies deleted, and the full suite is **3155 passed, 13 skipped, 0 failed** — the two extra over the pre-package count are the new layout tests.

The Doctor-side sentinel test (the writer never touches a byte past `PayloadBytes`) was verified by mutation rather than by passing: a stray write past the extent makes it fail with the intended message, and the mutation was reverted.

[Devin review](https://devinreview.com/BloomBooks/BloomDesktop/pull/8219)

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8219)


src/BloomExe/BloomExe.csproj
```diff
@@ -234,6 +234,14 @@
...
 		<PackageReference Include="Autofac" Version="6.0.0" />
 		<PackageReference Include="AWSSDK.Core" Version="3.5.1.32" />
+		<!--
+			The protocol Bloom and the Bloom Freeze Doctor share: the shared-memory health page, the
+			session file, and the named events. This used to be three files copied into this repo by
+			hand, 750 lines of them, kept in step by a drift check because they had already drifted.
+			One package means one definition of the wire format instead of two.
+			Source: https://github.com/BloomBooks/bloom-freeze-doctor
+		-->
+		<PackageReference Include="BloomBooks.FreezeDoctor.Protocol" Version="0.1.0-alpha" />
```
JohnThomson: [Devin — Investigate, current commit] **A fresh checkout cannot build until someone runs the pack script.**

This `PackageReference` can only be restored from the `localpackages` folder that the new root `NuGet.Config` adds, and that folder is empty on a fresh clone — only `.gitkeep` is tracked. Nothing is published to nuget.org yet, so restore fails on any machine that has not run `build/pack-freeze-doctor-protocol.ps1` against a clone of the Doctor repo. That includes CI and TeamCity.

Devin's own read is that this is intentional and documented rather than a bug, and adds that a reviewer should confirm the package is published and the temporary scaffolding removed before this actually merges. Agreed on both counts — which is why this thread is being left **open**.

**Leaving open deliberately: this is a merge precondition, not a settled point.** Before #8219 merges, three things have to go: the root `NuGet.Config`, `build/pack-freeze-doctor-protocol.ps1`, and `localpackages/` — replaced by the package being restorable from nuget.org. Resolving this thread now would delete the only durable record of that.

Worth noting for anyone reading the checks: GitHub Actions does not build Bloom on PRs, so this will not appear as a red check here. It would surface in TeamCity.

build/pack-freeze-doctor-protocol.ps1
```diff
@@ -0,0 +1,46 @@
...
+    Write-Host "or point BLOOM_FREEZE_DOCTOR_REPO at an existing clone. Note that the protocol project"
+    Write-Host "only exists on that repo's contract-package branch until it is merged."
+    exit 1
+}
+
+New-Item -ItemType Directory -Force -Path $output | Out-Null
+
+Write-Host "Packing $project"
+Write-Host "     to $output"
+dotnet pack $project --configuration Release --output $output
```
JohnThomson, line 39: [Devin — Investigate, current commit] **The pinned version has to match what the Doctor's csproj emits, across two repos.**

If `BloomExe.csproj` pins a version that `dotnet pack` does not produce from the Doctor's `BloomBooks.FreezeDoctor.Protocol.csproj`, restore fails with NU1101/NU1102. Devin's point is that this coupling spans two repositories and is worth confirming.

**Checked, and they agree.** Bloom pins `0.1.0-alpha`; the Doctor's csproj declares `<Version>0.1.0-alpha</Version>`; and the package sitting in `localpackages/` is `BloomBooks.FreezeDoctor.Protocol.0.1.0-alpha.nupkg`. All three line up.

The cross-repo coupling is real but temporary: once the package is published, the version comes from the Doctor's release tag and Bloom pins a published version like any other dependency, so there is no local pack step left to disagree with.

JohnThomson: Follow-up, since this thread named specific version numbers and they have moved: the protocol is now **0.2.0-alpha**, not `0.1.0-alpha`. Re-checked, and the three still agree — Bloom pins `0.2.0-alpha`, the Doctor's csproj declares `0.2.0-alpha`, and the packed nupkg is `BloomBooks.FreezeDoctor.Protocol.0.2.0-alpha.nupkg`.

The bump was needed for a reason worth recording on this thread, because it is the same class of problem Devin was pointing at: **NuGet caches by id *and* version**, so re-packing `0.1.0-alpha` with a changed layout would have let a machine that had already restored it keep using the old assembly, with restore succeeding and nothing looking wrong. Silent-and-plausible, exactly like the drift this whole change removes. So: bump the version for any layout change, even while the package is coming from a local folder rather than nuget.org.

JohnThomson:
Superseded by #8229, and its whole approach is no longer needed.

This PR shared the wire format between Bloom and the Freeze Doctor by publishing it as a NuGet package, because the two programs lived in different repositories. In #8229 the Doctor lives in *this* repository, so both sides reference one project directly — which is what the package existed to simulate.

That removed a fair amount of machinery along with it: the publish-then-unlist flow, an nuget.org API key, a version to keep in step across two repos, and the temporary local-feed scaffolding (`NuGet.Config`, the pack script, `localpackages/`).

Worth preserving from here, since it still applies to the shared files in their new home: the growth rules. `SchemaVersion` is the compatibility generation and is part of the shared section's *name*, so bumping it makes an old Doctor stop finding the channel rather than misreading it; additive change is what `PayloadBytes` is for, so a newer Doctor reading an older Bloom can tell a field that Bloom never wrote from a real zero. Both are carried over, with the tests that pin every offset by value.

PR #8220, opened 2026-08-20
Recover cleanly when a collection fails to open (BL-16678)

StephenMcConnel:
After one collection fails to open, every later collection in the same run used to die with **`An item with the same key has already been added. Key: audio/startrecord`**. `ProjectContext` registers its ~65 project-level API endpoints on the **application-level** `BloomServer.ApiHandler`, so a project that failed part way through construction left them registered, and the next project's identical registrations hit a duplicate key.

### #8204 (BL-16679) already fixes the reported symptom

It made the constructor dispose itself on failure, and `Dispose` already called `ClearProjectLevelHandlers()`. `_scope` is assigned well before the first `RegisterWithApiHandler`, so the clear is reached on the reported path. The log attached to the card confirms the boundary: the empty `Bad Collection.bloomCollection` failed at `_scope.Resolve<CollectionSettings>()`, *before* any registration, and poisoned nothing; the BL-16679 collection failed at `_scope.Resolve<TeamCollectionApi>()`, *after*, and poisoned everything.

So this PR is the hardening: recovery no longer *depends* on `Dispose` getting far enough, and it closes two other ways a failed open left Bloom running with no window and no way to quit.

### 1. The application-level reset can no longer be skipped

`Dispose` did the reset inside a `try` whose first statement was `_scope.Resolve<BloomServer>()` and whose `catch` only tolerated `ObjectDisposedException`. If that resolve threw anything else, `ClearProjectLevelHandlers()` was skipped, the exception was swallowed by the constructor's inner catch, and the next collection was poisoned again.

The server is now kept in a field, and the reset — extracted as `ReleaseApplicationLevelProjectState` — runs from that field, before anything that touches `_scope`.

We deliberately do **not** clear the server's own `CurrentCollectionSettings`. `BloomServer` reads it unguarded (e.g. to serve `writingSystemDisplayForUI.css` and files relative to the collection folder), so between one project closing and the next opening — exactly when the Open/Create Collections dialog is up — settings describing the collection we just left are a better answer than null. The next project overwrites them.

### 2. A clean slate is guaranteed at the recovery point

`Program.HandleErrorOpeningProjectWindow` is the one place that knows we failed and are about to offer another collection, so it repeats the reset (idempotent), with a `Debug.Assert` that `Dispose` should already have done it.

Its reporting half is now separate and guarded. It runs *inside* `OpenProjectWindow`'s catch, and the callers depend on `OpenProjectWindow` returning `false` to put up the chooser — so a failure to *report* used to take the user's way back into the app with it.

### 3. One throw no longer wedges startup permanently

`StartupScreenManager.DoStartupAction` ran `_current.Task()` unguarded, and returns immediately on every later idle while `_current` is set. A throw therefore killed the queue for the rest of the run: the splash screen never closed, `ChooseACollection()` never ran, nothing called `ProgramExit.Exit()` so not even its 20-second force-quit net was armed, and `Main`'s `finally` never released the `UniqueToken` — so the **next launch** was turned away with "Bloom is already running". That is the shape the card's title describes.

`FatalExceptionHandler` cannot save this: during collection open `UseFallback` is true, so it shows a message and *returns*, letting the process carry on windowless.

### 4. `ReopenProject` gets the fallback its sibling already had

`ReopenProject` called `OpenCollection(MruProjects.Latest)` with nothing on failure, unlike `OpenCollectionChosenInDialog` — although it also runs with the Shell already closed (it is reached from a UI language change and from the Collection Settings dialog). New `ChooseACollectionOrQuit` is used at all three recovery sites, so a chooser that itself fails quits rather than idling invisibly.

### 5. Two diagnosis traps, and a null hole

- `catch (FileNotFoundException)` showed the "Bloom was not able to find all its bits" message, called `ProgramExit.Exit()`, then **fell through** into code dereferencing the `_scope` it had just failed to build — so what the user and Sentry actually saw was a `NullReferenceException` naming nothing useful.
- `catch (Exception) { return; }` inside the `BeginLifetimeScope` lambda would have abandoned every remaining registration, turning any failure there into an unrelated Autofac `ComponentNotRegisteredException`.
- The startup MRU pruning loop called `IsInvalidCollectionToEdit(path)` after `path` could have become null.

### Testing

New `BloomApiHandlerTests` (5 tests) covers the application-level/project-level split the whole recovery depends on, and nothing covered it before. It includes a sanity check that re-registering *without* the clear really does throw the card's `ArgumentException`, so it cannot pass falsely. New `StartupScreenManagerTests` (3) covers a throwing startup action not wedging the queue. `ProjectContextTests` gains 3 tests for the new reset.

Full C# suite green (3155 passed). No TypeScript in the diff, so the front-end suite was not run.

Also verified by hand in the running app: with a deliberate failure staged just after `audio/startRecord` is registered, the card's repro now recovers — error dialog, chooser, and the good collection opens — and no windowless Bloom is left behind. With the self-dispose disabled to simulate the pre-#8204 state, Bloom did not survive at all, consistent with the card's reports.

### Deliberately not done

- **The MRU never demotes a collection that failed to open**, so every launch retries it, reports, and offers the chooser. Leaving it: `MostRecentPathsList.Paths` already prunes on serialization and `AddNewPath` refuses to re-add, so demoting on one failure would drop a user's Team Collection from the Open/Create dialog after a transient network blip. Annoying but always recoverable beats silently losing a collection.
- `Program.cs` writes a double-clicked `.bloomCollection` to the MRU *before* trying to open it.
- `CollectionSettings.DoDefenderFolderProtectionCheck` calls `Environment.Exit(-1)` from inside the `CollectionSettings` constructor, skipping `ReleaseBloomToken()` — a read-only collection folder is currently unrecoverable, and the Defender-specific wording will mislead. Worth its own card.
- No `--no-collection` / Shift-at-startup escape hatch for a poisoned MRU.
- `FileSystemWatcher.EnableRaisingEvents` is called unguarded in `TeamCollection.StartMonitoring` and `FolderTeamCollection`, and nothing subscribes to `FileSystemWatcher.Error`.

Ref: https://issues.bloomlibrary.org/youtrack/issue/BL-16678

[Devin review](https://devinreview.com/BloomBooks/BloomDesktop/pull/8220)

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8220)

### DIFF
--- a/src/BloomExe/ProjectContext.cs
+++ b/src/BloomExe/ProjectContext.cs
@@ -996,13 +996,18 @@ namespace Bloom
         /// </remarks>
         internal static void ReleaseApplicationLevelProjectState(BloomServer server)
         {
+            // No server means we failed before we ever got hold of one, so we had not registered any
+            // handlers and had not advertised our collection either -- there is nothing of ours to take
+            // back. Bailing out first also keeps a second Dispose() the no-op it has always been,
+            // rather than letting it clear settings that by then belong to a live project.
+            if (server == null)
+                return;
+
             // So that common/instanceInfo doesn't report a collection we are not in. Unlike the
             // server's copy, this one is read as "which collection is current", and the collection
             // chooser uses it to decide what to highlight.
             web.controllers.CommonApi.CurrentCollectionSettings = null;
 
-            if (server == null)
-                return; // we failed before we ever got hold of it, so we registered nothing
             try
             {
                 server.ApiHandler.ClearProjectLevelHandlers();
@@ -1027,11 +1032,6 @@ namespace Bloom
         /// ------------------------------------------------------------------------------------
         public void Dispose()
         {
-            if (_collectionLock != null)
-            {
-                _collectionLock.Unlock();
-            }
-
             // Disposing ProjectContext disables api functionality and disposes WorkspaceModel/View, BloomServer, et al.,
             // so we need to resort to our fallback error handler.
             ResetToFallbackHandler();
@@ -1051,6 +1051,15 @@ namespace Bloom
             // at least we avoid mysterious errors in the API handlers due to disposed objects.
             ReleaseApplicationLevelProjectState(_server);
             _server = null;
+
+            // Unlock after that, not before: CollectionLock.Unlock rethrows in a DEBUG build, and on the
+            // ordinary collection-switch path nothing catches it -- so unlocking first meant a failing
+            // unlock could leave the project's API handlers registered, which is the state this whole
+            // change exists to prevent. (BL-16678)
+            if (_collectionLock != null)
+            {
+                _collectionLock.Unlock();
+            }
 
             // We can be disposed without a scope: our constructor disposes us if it fails, and it can
             // fail in (or even before) BuildSubContainerForThisProject. Unlocking the collection above
--- a/src/BloomExe/ProjectContext.cs
+++ b/src/BloomExe/ProjectContext.cs
@@ -45,6 +45,13 @@ namespace Bloom
         /// </summary>
         private ILifetimeScope _scope;
 
+        /// <summary>
+        /// The application-level server we registered our project-level API endpoints on. Kept so that
+        /// Dispose can take them off again without having to resolve anything. See
+        /// ReleaseApplicationLevelProjectState. (BL-16678)
+        /// </summary>
+        private BloomServer _server;
+
         //		private CommandAvailabilityPublisher _commandAvailabilityPublisher;
         public Form ProjectWindow { get; private set; }
 
@@ -68,17 +75,21 @@ namespace Bloom
         )
         {
             Logger.WriteMinorEvent("starting to construct the project context");
-            // Settle which settings file we are really opening before anything uses the path. The
-            // caller's may not exist: the name is not always the folder's (BL-16679), and the lock
-            // below would then be aimed at a file that isn't there -- silently protecting nothing in
-            // a release build, and throwing in a DEBUG one.
-            SettingsPath = GetRealSettingsPath(projectSettingsPath);
-
-            _collectionLock = new CollectionLock(SettingsPath);
-            _collectionLock.Lock();
-
             try
             {
+                // Settle which settings file we are really opening before anything uses the path. The
+                // caller's may not exist: the name is not always the folder's (BL-16679).
+                SettingsPath = GetRealSettingsPath(projectSettingsPath);
+
+                _collectionLock = new CollectionLock(SettingsPath);
+                // Only lock a file that is there to be locked. GetRealSettingsPath hands back the path
+                // it was given when the folder holds no .bloomCollection at all, and Lock() rethrows in
+                // a DEBUG build -- which would replace the clear "could not find a .bloomCollection
+                // file in ..." that GetCollectionSettings throws with a bare FileNotFoundException
+                // about the lock, sending the next developer to look in the wrong place. (BL-16678)
+                if (RobustFile.Exists(SettingsPath))
+                    _collectionLock.Lock();
+
                 // BL-8019: A couple lines down, BuildSubContainerForThisProject() starts BloomServer with the new project.
                 // While we are starting (or restarting, in the case of switching collections) BloomServer we need to use
                 // the WinFormsExceptionHandler mechanism, which doesn't use a browser.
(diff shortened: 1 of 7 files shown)


src/BloomExe/ProjectContext.cs
```diff
@@ -433,9 +444,19 @@ IContainer parentContainer
                     MessageBoxIcon.Error
                 );
                 ProgramExit.Exit();
+                // Don't fall through. This used to, straight into code that dereferences the _scope we
+                // just failed to build, so what the user and Sentry actually got was a
+                // NullReferenceException naming nothing useful instead of the missing file. Throwing
+                // lets our constructor dispose us and the real exception reach the log and the report.
+                // (BL-16678)
+                throw;
```
StephenMcConnel, line 451: [Devin] **Investigate**: FileNotFoundException path now shows two dialogs before quitting

*(mirrored by Claude Opus 5 (1M context) from Steve McConnel's machine during preflight — Devin does not post to GitHub itself)*

**What Devin flagged.** When Bloom is opening a collection and discovers that pieces of its own program are missing (a damaged install), it shows the "Bloom was not able to find all its bits… please run the installer again and choose 'Repair'" message and asks to quit. This PR stops that handler from falling through into code that dereferenced the scope it had just failed to build, and passes the real exception on instead. A consequence is that Bloom's ordinary "had a problem loading the *X* collection" report also appears, so a user with a broken install now sees two dialogs in a row rather than one.

**Why it may matter.** It is a visible change for real users, even if a rare set of them. Devin's own read is that it "appears to be an accepted trade-off, but it is a visible UX change worth confirming".

**Why the change is still an improvement.** Before this PR the fall-through threw a `NullReferenceException` on the half-built scope, so while the user saw the right *message*, the crash report that reached us named a meaningless internal error instead of the file that was actually missing. The second dialog is the price of the report naming the real cause.

This one is deliberately left open: it is a user-facing behaviour question rather than a defect, so it has gone to the developer as a decision rather than being resolved here. The alternative on the table is to suppress the second dialog by marking the error as already reported (~15 lines, but it adds machinery to a path that is hard to test).

StephenMcConnel:
[Claude Opus 5 (1M context) from Steve McConnel's machine during preflight] Consulted Devin on 2026-08-20 20:49 UTC, up to commit `e5f662dcb161917bbc99241b3d3a4434dd85125d`.

It raised one Bug and six Flags.

The Bug — the "path too long" signal being swallowed on the startup recovery path, leaving Bloom invisible — Devin itself now marks resolved. The local review had found the same thing independently, and it was fixed in `e5f662dcb` before Devin's pass finished, so there was never a thread to close.

Of the Flags, one needed investigating (the broken-install path now showing two dialogs) and is posted above as a review thread. It is deliberately left **open**: it is a user-facing judgement call that has gone to the developer as a decision rather than something to settle here.

The other five were informational and are not mirrored. All five are choices this PR makes deliberately and documents in code comments: a repeating startup action that throws is dropped rather than retried; the server's own copy of the collection settings is deliberately left alone; the collection lock is skipped when the settings file is absent; startup-task exceptions are reported rather than left fatal; and one `ObjectDisposedException` catch may now be unreachable, which is harmless.

CI is green. Greptile posted only a "trial has ended" notice, and CodeRabbit is switched off for this repo in `.coderabbit.yml`.

src/BloomExe/ProjectContext.cs
StephenMcConnel: [Claude Opus 5 (1M context) from Steve McConnel's machine during preflight]

Decided: leave as is. The developer has accepted the extra dialog.

The reasoning: a broken installation is rare, and it is already a dead end for the user whatever we show them — the value of this change is that the crash report we receive now names the file that is actually missing, instead of a `NullReferenceException` thrown on the half-built scope a moment later. That is the difference between a report we can act on and one we cannot. Suppressing the second dialog would mean threading an "already reported" flag through the reporting path, which is new machinery on a path that is hard to exercise and rarely hit — not worth it for one extra dialog on a route that ends in "reinstall Bloom" either way.

Thanks for raising it; it was the right thing to flag, since it is a genuine user-visible change rather than a pure internal one.

JohnThomson (approved):
> @JohnThomson reviewed 7 files and all commit messages, and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8220)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on StephenMcConnel).

JohnThomson (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8220)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on StephenMcConnel).

PR #8221, opened 2026-08-21
Rewrite a resized image's reference with the encoder that wrote it (BL-16669)

JohnThomson:
## Problem

Two unrelated faults, both found by Steve McConnel while testing the earlier percent-in-filename
fix in 6.5.1322.

**Pictures went missing from published books.** Publishing to BloomPUB silently dropped any
picture whose file name contained a `%`, `#`, `&` or `+` — a blank space where the picture should
be, in Bloom's own preview and in BloomPUB Viewer alike. The file in the published book had been
renamed to `Restroom%41%42_r.jpg` while the page still asked for `Restroom%2541%2542.jpg`.
Ordinary names, spaces included, came through fine, which made it look narrower than it was:
`red & green.jpg` and `one+one.jpg` were equally broken.

**A picture's click sound could not be set at all.** The menu item for the sound that plays when a
reader touches a picture had been reachable only on draggable-game pages since the canvas controls
were reworked in February, so on an ordinary page there was no way in.

## Cause

The first is an encoder mismatch. Publishing shrinks each image, which renames it, and then has to
point the page at the new file. It did that by searching the page's style for the old name and
substituting — encoding both names with a *different* encoder than the one that wrote the style.
`UrlPathString.UrlEncoded` (the writer) escapes `%`, `#`, `&` and `+`; `HttpUtility.UrlPathEncode`
(the searcher) escapes none of them. So nothing matched, and the page went on naming the file the
resize had just deleted.

The second was a one-condition oversight: `canChooseAudioForElement` was gated on the page being a
draggable game rather than on the selected thing being a picture.

## What this PR changes

- Publishing now **replaces the url outright** rather than hunting for the old name, so there is
  no old-name encoding left to get wrong. Anything else in the style is untouched.
- The two places in the same file that read a file name *out* of a page get the matching
  treatment; they used `HttpUtility.UrlDecode`, which turns a literal `+` into a space (BL-3259).
- **Any picture can take a click sound again**, on any page. Text is deliberately left alone: it
  never had a click sound, and its form of that menu item points at the Talking Book tool instead.
- Tests for both. The publish one puts a picture through ten awkward names and checks the page
  still points at a file that exists — seven of the ten fail without the fix. The menu one covers
  picture and text, on a game page and an ordinary one, plus the page's background picture.

Steve's third finding — a game sound with a `%` in its name staying silent when you *play* the
game, in the editor as well as in a published book — is bloom-player's, tracked as BL-16688, and
is deliberately not in this PR.

Ref: https://issues.bloomlibrary.org/youtrack/issue/BL-16669

[Devin review](https://devinreview.com/BloomBooks/BloomDesktop/pull/8221)

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8221)

### DIFF
--- a/src/BloomBrowserUI/bookEdit/toolbox/canvas/buildCanvasElementControlRegistryContext.ts
+++ b/src/BloomBrowserUI/bookEdit/toolbox/canvas/buildCanvasElementControlRegistryContext.ts
@@ -205,9 +205,18 @@ export const buildCanvasElementControlRegistryContext = (
             !!img &&
             !img.getAttribute("data-copyright"),
         isInDraggableGame,
+        // A click sound ("play when touched") is offered for any image, on any page, not
+        // just inside a draggable game: attaching a sound to a picture is ordinary book
+        // making, and the game-only gate was an unintended casualty of the move to this
+        // declarative registry -- it left no way at all to set one on a normal page
+        // (BL-16669 testing).
+        // Text is deliberately NOT broadened. There has never been a click sound for text;
+        // the text form of this menu item just points at the Talking Book tool, so it stays
+        // where it was.
         canChooseAudioForElement:
             elementType === "sound" ||
-            (isInDraggableGame && (hasImage || hasText)),
+            hasImage ||
+            (isInDraggableGame && hasText),
         hasCurrentImageSound,
         currentImageSoundLabel: hasCurrentImageSound
             ? dataSound.replace(/\.mp3$/, "")
--- a/src/BloomExe/Publish/BloomPub/BloomPubMaker.cs
+++ b/src/BloomExe/Publish/BloomPub/BloomPubMaker.cs
@@ -296,14 +296,10 @@ namespace Bloom.Publish.BloomPub
                             claimedFilenames
                         );
                         if (newFilename != null)
-                        {
-                            var oldEncoded = System.Web.HttpUtility.UrlPathEncode(filename);
-                            var newEncoded = System.Web.HttpUtility.UrlPathEncode(newFilename);
                             div.SetAttribute(
                                 "style",
-                                style.Replace(oldEncoded, newEncoded).Replace(filename, newFilename)
+                                ReplaceBackgroundImageFilename(style, newFilename)
                             );
-                        }
                     }
 
                     // After ConvertImagesToBackground, all content images (inside
@@ -470,18 +466,54 @@ namespace Bloom.Publish.BloomPub
                     .Cast<SafeXmlElement>()
             )
             {
-                preservedImages.Add(System.Web.HttpUtility.UrlDecode(img.GetAttribute("src")));
+                // An img @src is URL-encoded (see the encoding conventions in UrlPathString), and
+                // must be decoded the matching way; HttpUtility.UrlDecode would eat a literal '+'.
+                preservedImages.Add(
+                    UrlPathString.CreateFromUrlEncodedString(img.GetAttribute("src")).NotEncoded
+                );
             }
             return preservedImages;
         }
 
         private static string ExtractFilenameFromBackgroundImageStyleUrl(string style)
         {
-            var filename = style.Substring(
-                style.IndexOf(kBackgroundImage) + kBackgroundImage.Length
-            );
-            filename = filename.Substring(0, filename.IndexOf("'"));
-            return System.Web.HttpUtility.UrlDecode(filename);
+            // The url in the style is URL-encoded (HtmlDom.SetImageElementUrl wrote it with
+            // UrlPathString.UrlEncoded), so decode it the matching way. HttpUtility.UrlDecode is
+            // NOT the matching way: it would turn a '+' in a real file name into a space (BL-3259).
+            return UrlPathString
+                .CreateFromUrlEncodedString(ExtractEncodedUrlFromStyle(style, out _, out _))
+                .NotEncoded;
+        }
+
+        /// <summary>
+        /// Return the encoded url between the quotes of a "background-image:url('...')" style,
+        /// along with where in <paramref name="style"/> it starts and ends.
+        /// </summary>
+        private static string ExtractEncodedUrlFromStyle(string style, out int start, out int end)
+        {
+            start = style.IndexOf(kBackgroundImage) + kBackgroundImage.Length;
+            end = style.IndexOf("'", start);
+            return style.Substring(start, end - start);
+        }
+
+        /// <summary>
+        /// Point a "background-image:url('...')" style at a different file in the same folder,
+        /// leaving the rest of the style alone.
+        /// </summary>
+        /// <remarks>
+        /// This replaces the url outright rather than looking for the old name in the style and
+        /// substituting, because the search is what used to go wrong: it encoded the old and new
+        /// names with HttpUtility.UrlPathEncode, which encodes neither '%' nor '#' nor '&amp;', while
+        /// the style itself was written with UrlPathString.UrlEncoded, which encodes all three. So
+        /// for any such name nothing matched, the style went on naming the file that the resize
+        /// had just replaced, and the image vanished from the published book (BL-16669).
+        /// </remarks>
+        private static string ReplaceBackgroundImageFilename(string style, string newFilename)
+        {
+            ExtractEncodedUrlFromStyle(style, out var start, out var end);
+            return style.Substring(0, start)
+                + UrlPathString.CreateFromUnencodedString(newFilename).UrlEncoded
+                + style.Substring(end);
         }
 
         private static void CreateVersionFileWithSha(string bookFilePath, string outputDirectory)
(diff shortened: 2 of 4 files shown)


JohnThomson:
[Claude Opus 5 (1M context) from John Thomson's machine during preflight]

Consulted Devin on 2026-08-21 16:40 UTC up to commit `6c71cdcbb0b301276d3c9020af2a4316c3f41cc9`.

**No bugs and nothing to investigate.** Devin raised two informational observations, neither of which needs a change, so neither is mirrored as a review thread:

- It noticed that the helper which pulls a file name out of a `background-image` style will throw if the style has no `url('...')` in it at all — and said so itself: that shape is pre-existing and sits behind a `Debug.Fail` "should not happen" branch. This repo's convention is to fail rather than silently work around a broken assumption, so it stays as it is.
- The other is a confirmation rather than a concern: the new code re-encodes the file name with the same encoder that wrote the style, and both read paths decode with the matching one, so the round trip holds and the preserve-resolution comparison stays consistent. That is exactly the property the fix is for.

Other reviewers: the GitHub `pr-automation` check passed. Greptile posted only a "trial has ended" notice, so it did not review. The full C# suite is green locally (3182 passed, 13 skipped, 0 failed).

src/BloomBrowserUI/bookEdit/toolbox/canvas/buildCanvasElementControlRegistryContext.ts
```diff
@@ -205,9 +205,18 @@ export const buildCanvasElementControlRegistryContext = (
...
         isInDraggableGame,
+        // A click sound ("play when touched") is offered for any image, on any page, not
+        // just inside a draggable game: attaching a sound to a picture is ordinary book
+        // making, and the game-only gate was an unintended casualty of the move to this
+        // declarative registry -- it left no way at all to set one on a normal page
+        // (BL-16669 testing).
+        // Text is deliberately NOT broadened. There has never been a click sound for text;
+        // the text form of this menu item just points at the Talking Book tool, so it stays
+        // where it was.
         canChooseAudioForElement:
```
JohnThomson, line 216: [Devin] **Investigate**: Image click sound now offered for background and navigation images

`canChooseAudioForElement` now returns true for any element with an image (`hasImage`), so the menu item also appears for background images and image-bearing navigation buttons, not just ordinary pictures. The tests cover only plain image/text elements, so this wider surface is unverified.

JohnThomson: [Claude Opus 5 (1M context) from John Thomson's machine during preflight]

Half right, and the half that's right is worth a decision.

**Navigation buttons: not affected.** The three navigation-button element types list their menu sections as `["url", "image", "wholeElement"]` (plus `"text"` where they have a label) — no `"audio"` section. So the `chooseAudio` control is never in a navigation button's menu, and this flag's availability function is never consulted for one. Nothing changes there.

**The page's background picture: yes, it now offers "play when touched" on an ordinary page.** A background picture infers as element type `image`, which does carry the `audio` section, so widening the gate reaches it. Two things worth knowing before treating that as a defect:

- It was already the case *inside a draggable game*. The old condition was `isInDraggableGame && (hasImage || hasText)`, with no background-image exclusion — unlike `canToggleDraggability` right above it, which does exclude `isBackgroundImage` explicitly. So this isn't new behavior so much as the same behavior now reaching ordinary pages.
- Attaching a sound to a full-page picture is a plausible thing for a book maker to want — touch the picture, hear the word.

On the "unverified" point, fair, and now fixed: there is a test case covering the background-picture path, so whichever way we settle it, the behavior is pinned rather than incidental.

Whether a background picture *should* offer a click sound is a product call rather than a correctness one, so it goes to the developer instead of being decided here. Leaving this thread open until it comes back.

src/BloomExe/Publish/BloomPub/BloomPubMaker.cs
```diff
@@ -470,18 +466,54 @@ var img in dom.SafeSelectNodes(
                     .Cast<SafeXmlElement>()
             )
             {
-                preservedImages.Add(System.Web.HttpUtility.UrlDecode(img.GetAttribute("src")));
+                // An img @src is URL-encoded (see the encoding conventions in UrlPathString), and
+                // must be decoded the matching way; HttpUtility.UrlDecode would eat a literal '+'.
+                preservedImages.Add(
+                    UrlPathString.CreateFromUrlEncodedString(img.GetAttribute("src")).NotEncoded
```
JohnThomson, line 472: [Devin] **Investigate**: preserveResolution img src assumed URL-encoded

`FindImagesToPreserveResolution` decodes an img `@src` with `CreateFromUrlEncodedString`, assuming it is URL-encoded and non-null. `SetImageElementUrl` (`src/BloomExe/Book/HtmlDom.cs:2925-2935`) only encodes img src when `urlEncode` is true. This path rarely fires, but the assumption is worth confirming.

JohnThomson: [Claude Opus 5 (1M context) from John Thomson's machine during preflight]

Confirmed, and both halves of the assumption hold here.

**Non-null.** This is a `SafeXmlElement`, whose `GetAttribute(name)` forwards to `XmlElement.GetAttribute(name)`, which returns `String.Empty` for a missing attribute rather than null. (The base `SafeXmlNode.GetAttribute` does return null, but the loop casts to `SafeXmlElement`, so the override is what runs.) An empty string round-trips harmlessly. Worth adding that the previous code here, `HttpUtility.UrlDecode`, was null-tolerant, so this was the right thing to check.

**URL-encoded.** Only two callers pass `urlEncode: false`, and neither writes into the DOM this code reads:

- `EpubMaker` writes into the ePUB's own xhtml, a separate output tree.
- `PageListApi` writes into the thumbnail DOM it serves to the page list, not the book's saved DOM.

The BloomPUB path here walks a clone of the book, whose img `src` values are written encoded — the convention documented on `UrlPathString`. So an encoded read is correct for this DOM. (The two exceptions are a good illustration of why that convention is per-DOM rather than universal.)

One last note on direction: even in a hypothetical unencoded src, this change does not make things worse. Both the old and new call would mis-handle a literal `%41`; the difference is that the new one preserves a literal `+` where `HttpUtility.UrlDecode` turned it into a space. So this is strictly the safer of the two.

No change needed; resolving.

JohnThomson:
[Claude Opus 5 (1M context) from John Thomson's machine during preflight]

Consulted Devin on 2026-08-24 up to commit `6e557b92aa7fc5bbc7f21ddee6de158eb99164e9`.

No bugs. Two things to investigate, both mirrored as review threads above:

- **A picture's address where the publish code reads it** — resolved. Checked both halves of the assumption: the attribute cannot come back null on this element type, and the only two writers that leave a picture's address unencoded write into other documents (the ePUB, and the page-list thumbnail), not the one this code walks. Reasoning is on the thread.
- **The click sound now reaching a page's background picture** — left **open** for the developer. Devin's other half of that flag, about navigation buttons, does not apply: those menus carry no audio section at all, so this gate is never consulted for one. Recorded on the thread.

Two informational notes as well, neither needing a change; one of them is a confirmation that the encoding round trip now matches on both the read and the write side.

Local suites, on the tree with master merged in: C# 3182 passed / 13 skipped / 0 failed; front-end 729 passed / 5 skipped / 0 failed. `pr-automation` green. Greptile posts only a "trial has ended" notice, so it has not reviewed.

src/BloomBrowserUI/bookEdit/toolbox/canvas/buildCanvasElementControlRegistryContext.ts
JohnThomson: [Claude Opus 5 (1M context) from John Thomson's machine during preflight]

Developer decision: **leave as is** — a page's background picture keeps offering "play when touched".

The reasoning, for whoever reads this later: this is not new behavior so much as existing behavior reaching further. Inside a draggable game the background picture already offered a click sound, because the old condition had no background-picture exclusion (unlike the draggability check immediately above it, which excludes it explicitly). Narrowing it now would be the only change in this PR that *removes* something currently available, and attaching a sound to a full-page picture is a reasonable thing for a book maker to want — touch the picture, hear the word.

The behavior is pinned by a test case covering the background-picture path, so if we ever do decide to narrow it, that is a visible edit rather than a silent drift.

The other half of this flag, about navigation buttons, does not apply: those element types list no "audio" menu section, so this gate is never consulted for one.

Resolving.

StephenMcConnel (approved):
> @StephenMcConnel reviewed 4 files and all commit messages, and resolved 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8221)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on JohnThomson).

PR #8222, opened 2026-08-21
Remove some Experimental Flags (BL-16731)

andrew-polk:
## Problem

App Building and the AI image editor were each hidden behind an extra "Experimental Features"
checkbox in Collection Settings → Advanced. Both features are now considered ready, so that
second opt-in is just an obstacle: a Pro subscriber had to know to go find and tick a checkbox
before the Publish tab's **Apps** button or the image context menu's **Edit with AI…** command
would appear at all. The subscription tier was already doing the real gating.

## Fix

- `ExperimentalFeatures` no longer defines `app-builder` or `ai-image-editing`, and the startup
  migration clears those two tokens out of users' stored settings — the same treatment
  `webView2` got when it went mainstream.
- The two features' registry entries no longer carry an `ExperimentalFeatureToken`, so they are
  always visible and gated only by the Pro tier. The generic experimental-token mechanism stays
  in place for whatever is experimental next.
- The two checkboxes are gone from the Advanced settings panel, along with the pending-value
  plumbing that carried them between the WinForms dialog, the settings API, and the React panel.
- Their two English label strings get opposite treatment in the XLF, because only one of them
  will ever be translated. **App Builder** is marked `Obsolete as of Bloom 6.5` and kept: 6.4
  made that string translatable, so it goes to Crowdin, and removing the id from master would
  destroy those translations. **Edit Images with AI** does not exist on `Version6.4` at all, has
  always been `translate="no"`, and appears in no translated file, so it is simply deleted.
- The two tests that exercised the experimental-token mechanism were written against
  `kAppBuilder`; they now use a locally-constructed `FeatureInfo` with a token no real feature
  uses, so the mechanism stays covered. Added a test pinning that both features are now
  unconditionally visible.

**Net effect for users:** the two checkboxes disappear from Collection Settings; the Publish
tab's Apps button and the "Edit with AI…" image command show for everyone, with the existing
subscription overlays handling the Pro requirement.

Ref: https://issues.bloomlibrary.org/youtrack/issue/BL-16731

[Devin review](https://devinreview.com/BloomBooks/BloomDesktop/pull/8222)

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8222)

### DIFF
--- a/DistFiles/localization/en/BloomMediumPriority.xlf
+++ b/DistFiles/localization/en/BloomMediumPriority.xlf
@@ -38,11 +38,7 @@
         <source xml:lang="en">App Builder</source>
         <note>ID: CollectionSettingsDialog.AdvancedTab.Experimental.AppBuilder</note>
         <note>This checkbox enables the experimental publish-to-apps workflow.</note>
-      </trans-unit>
-      <trans-unit id="CollectionSettingsDialog.AdvancedTab.Experimental.EditWithAI" translate="no">
-        <source xml:lang="en">Edit Images with AI</source>
-        <note>ID: CollectionSettingsDialog.AdvancedTab.Experimental.EditWithAI</note>
-        <note>This checkbox enables the experimental "Edit with AI" image editor.</note>
+        <note>Obsolete as of Bloom 6.5</note>
       </trans-unit>
       <trans-unit id="Feature.AiImageEditing" translate="no">
         <source xml:lang="en">Edit Images with AI</source>
--- a/src/BloomBrowserUI/collection/AdvancedSettingsPanel.tsx
+++ b/src/BloomBrowserUI/collection/AdvancedSettingsPanel.tsx
@@ -19,8 +19,6 @@ interface IAdvancedSettings {
     autoUpdate?: boolean;
     showExperimentalBookSources?: boolean;
     allowTeamCollection?: boolean;
-    allowAppBuilder?: boolean;
-    allowAiImageEditing?: boolean;
     showQrCode?: boolean;
     qrcodeCaption?: string;
 }
@@ -61,14 +59,6 @@ export const AdvancedSettingsPanel: React.FunctionComponent = () => {
         "Team Collections",
         "TeamCollection.TeamCollections",
     );
-    const appBuilderLabel = useL10n(
-        "App Builder",
-        "CollectionSettingsDialog.AdvancedTab.Experimental.AppBuilder",
-    );
-    const aiImageEditingLabel = useL10n(
-        "Edit Images with AI",
-        "CollectionSettingsDialog.AdvancedTab.Experimental.EditWithAI",
-    );
     const qrCodesLabel = useL10n(
         "QR Codes",
         "CollectionSettingsDialog.AdvancedTab.QrCodes",
@@ -93,16 +83,6 @@ export const AdvancedSettingsPanel: React.FunctionComponent = () => {
     const featureStatus = useGetFeatureStatus("TeamCollection");
     const teamCollectionOptionEnabled =
         featureStatus === undefined ? true : featureStatus.enabled;
-    const appBuilderFeatureStatus = useGetFeatureStatus("AppBuilder");
-    const appBuilderOptionEnabled =
-        appBuilderFeatureStatus === undefined
-            ? false
-            : appBuilderFeatureStatus.enabled;
-    const aiImageEditingFeatureStatus = useGetFeatureStatus("AiImageEditing");
-    const aiImageEditingOptionEnabled =
-        aiImageEditingFeatureStatus === undefined
-            ? false
-            : aiImageEditingFeatureStatus.enabled;
     const canChangeTeamCollectionOption = allowTeamCollectionEnabled !== false;
 
     const normalizeConfigrSettings = React.useCallback(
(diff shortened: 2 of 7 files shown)


DistFiles/localization/en/BloomMediumPriority.xlf
```diff
@@ -34,16 +34,6 @@
...
-      <trans-unit id="CollectionSettingsDialog.AdvancedTab.Experimental.AppBuilder" translate="no">
-        <source xml:lang="en">App Builder</source>
-        <note>ID: CollectionSettingsDialog.AdvancedTab.Experimental.AppBuilder</note>
-        <note>This checkbox enables the experimental publish-to-apps workflow.</note>
-      </trans-unit>
-      <trans-unit id="CollectionSettingsDialog.AdvancedTab.Experimental.EditWithAI" translate="no">
-        <source xml:lang="en">Edit Images with AI</source>
-        <note>ID: CollectionSettingsDialog.AdvancedTab.Experimental.EditWithAI</note>
-        <note>This checkbox enables the experimental "Edit with AI" image editor.</note>
-      </trans-unit>
```
andrew-polk, line 46: [Devin] **Bug**: Unused localization strings deleted instead of marked obsolete

Two `trans-unit` entries are deleted from `DistFiles/localization/en/BloomMediumPriority.xlf`. The localization rules in `src/BloomBrowserUI/AGENTS.md` require unused strings to be kept and annotated with a `<note>Obsolete as of ...</note>` note rather than removed, and forbid deleting obsolete entries until the version ships to Release.

andrew-polk: [Claude Opus 5 (1M context)] Not an issue here — this deletion was explicitly directed by the developer, which is the exception the rule names.

The rule in `src/BloomBrowserUI/AGENTS.md` says never to delete an obsolete entry *on your own initiative*, and to "treat deletion as out of scope unless the developer explicitly asks you to delete the entries for a particular version." Andrew asked for exactly that, after we checked the specific facts:

- Both entries were born `translate="no"` in commit `8082ce981` and were never edited afterwards, on `master` or on `Version6.4`. `translate="no"` means they were never sent to translators.
- Neither id appears in any of the 57 translated language folders under `DistFiles/localization/`, on `master` or `Version6.4`.

So there are no translations to lose, which is the harm the rule exists to prevent. The "wait until the version reaches Release" clause protects a shipped update to an older version — but `Version6.4` keeps its own copy of the file, so deleting from `master` cannot affect a 6.4 maintenance release either.

Resolving.

andrew-polk:
[Claude Opus 5 (1M context)] Consulted Devin on 2026-08-21 11:55 up to commit `c969dbb162041938f060431b552f9b98ac1883ea`.

It raised one Bug — that the two XLF strings were deleted rather than marked obsolete. That is posted as its own review thread, where it is answered and resolved: the developer explicitly asked for the deletion, and neither string had ever been sent to translators or appears in any translated file.

Two Informational items (not mirrored, low signal): both were observations confirming the change is consistent — the obsolete-token cleanup in `ExperimentalFeatures.cs`, and the two features becoming unconditionally visible in `FeatureRegistry.cs`.

No Investigate flags. CI (pr-automation) is green.

DistFiles/localization/en/BloomMediumPriority.xlf
andrew-polk: [Claude Opus 5 (1M context)] Correcting one part of my reply above. The conclusion stands, but my supporting argument was wrong.

I claimed the "wait until the version reaches Release" caution didn't apply because `Version6.4` keeps its own copy of the file. That is false, and it's the exact trap the rule exists for. Per `.github/prompts/bloom-l10.prompt.md`: Crowdin syncs off **master**, so when an id disappears from master, Crowdin drops the string *and its translations* — and "the fact that its code base still has it will not be sufficient." A release branch's own copy saves nothing. Please disregard that sentence; nobody should reuse it to justify a future deletion.

What actually makes this deletion lossless is the other half, which is unchanged: both entries were born `translate="no"` in `8082ce981` and were never edited since, on `master` or `Version6.4`, and neither id appears in any of the translated language folders. They were never handed to Crowdin, so Crowdin has no translations for them to drop.

Both `src/BloomBrowserUI/AGENTS.md` and `.github/skills/xlf-strings/SKILL.md` are being updated to state the Crowdin-off-master mechanism explicitly and to warn against the release-branch argument I made, so the next agent doesn't repeat it.

andrew-polk: [Claude Opus 5 (1M context)] Final outcome on this thread: **Devin was right about App Builder**, and the code now does what it asked for.

What changed since my replies above. `CollectionSettingsDialog.AdvancedTab.Experimental.AppBuilder` was `translate="no"` when I checked it, which is what made me conclude it had no translations to lose. Version6.4 has since **made that string translatable** (`69dce9087`, merged to master in `df9992c3b`), so it now goes to Crowdin and will acquire translations — and deleting the id from master would destroy them. It is therefore restored and marked `<note>Obsolete as of Bloom 6.5</note>`, exactly as this finding asked.

`CollectionSettingsDialog.AdvancedTab.Experimental.EditWithAI` is still deleted. That one does not exist on `Version6.4` at all, has always been `translate="no"`, and appears in no translated file, so there is nothing for Crowdin to lose.

Also worth recording, since I got it wrong on this thread: my argument that a deletion was safe because "the Version6.4 branch keeps its own copy of the file" was false. Translations reach a release branch only by being picked from master, so a loss on master propagates. That reasoning is now written down once, in `DistFiles/localization/README.md` under "Why we can't just delete a string", with `src/BloomBrowserUI/AGENTS.md` and the xlf-strings skill pointing at it (PR #8223).

andrew-polk (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8222)* status: 0 of 7 files reviewed, all discussions resolved.

StephenMcConnel (approved):
> @StephenMcConnel reviewed 7 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8222)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on andrew-polk).

PR #8223, opened 2026-08-21
Consolidate the XLF/Crowdin localization rules into one home each

andrew-polk:
## Problem

The rules for editing Bloom's localizable strings (XLF entries) were stated in five documents,
each with its own partly-wrong account of *why* they exist, and none of them linking to the
others. `DistFiles/localization/README.md` is the only doc that describes the Crowdin mechanics
correctly, and nothing an agent normally loads pointed at it. The concrete cost: on BL-16731 an
agent reasoned from one of the incomplete copies and told a reviewer, wrongly, that deleting a
string from master was safe "because the Version6.4 branch keeps its own copy of the file."

## Cause

The five copies had drifted, and the one correct explanation was unreachable from always-loaded
context. In particular the widely-read `src/BloomBrowserUI/AGENTS.md` justified the
don't-delete rule with "until the version is promoted to Release" — a rule of thumb that does not
explain, and in fact contradicts, how translations actually reach a release branch.

## What this PR changes

Documentation only; no code. Each kind of content now has exactly one home, and the others cite
it rather than restating it.

- **`DistFiles/localization/README.md`** owns the reasoning. Its new "Why we can't just delete a
  string" section traces the one route translations travel — Crowdin reads its sources from
  master, translations return to master via the `l10n_master` PR, and a release branch gets them
  only by cherry-picking that merge — so an id deleted from master is gone from Crowdin and can
  never be picked into `Version6.x`. It names the trap explicitly: the release branch still
  listing the entry saves nothing. It also documents the one real exception (an entry that was
  always `translate="no"` never reached Crowdin, so has no translations to lose), and corrects
  the "Effects of English xliff file changes" bullet that described removal without saying it
  deletes translations.
- **`.github/skills/xlf-strings/SKILL.md`** owns the procedures. It gains the two commands that
  establish that exception — including the point that the evidence must go in the commit message
  and the review reply, because a reviewer reading only the rule will correctly flag the
  deletion — and a matching item in the PR-review checklist. It is also now the **only** place
  the obsolete-note text and the version source are written down.
- **`src/BloomBrowserUI/AGENTS.md`** owns the short normative rules: it drops its duplicated
  priority table and its incorrect "promoted to Release" reasoning, keeping the rules plus
  pointers to both docs above. It stays normative because review bots cite this file by name.
- **`AGENTS.md`** (root) links the README so it is reachable from always-loaded context — the
  omission that caused the original mistake — and states the two rules that hold even outside the
  skill. **`.github/prompts/bloom-l10.prompt.md`** is trimmed to its own job, making one selected
  string localizable.

Two details that would have reproduced the original failure by other routes are fixed rather than
documented: the always-`translate="no"` check uses `git log -G` (with `-S` it misses a commit that
edited only the `translate` attribute, and then looks clean), and it no longer names a release
branch that will go stale — an unfilled placeholder fails loudly instead. The prompt file also had
`useL10n`'s arguments in the wrong order, an invalid-JSX example, and a path to a folder that does
not exist.

Ref: BL-16731 (context in which the documentation gap was found; this PR is documentation cleanup, not a fix for that card)

[Devin review](https://devinreview.com/BloomBooks/BloomDesktop/pull/8223)

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8223)

### DIFF
--- a/.github/prompts/bloom-l10.prompt.md
+++ b/.github/prompts/bloom-l10.prompt.md
@@ -4,7 +4,7 @@ description: make a string localizable
 
 This project uses the following methods to make strings localizable in react components:
 - use an l10n-aware component such as <BloomButton l10nKey="myKey">My Text</BloomButton>, <H1 l10nKey="myKey">My Text</H1>, etc.
-- useL10n("myKey", "My Text")
+- useL10n("My Text", "myKey")  — note the order: the English text first, then the id
 
 There should be a selected line with a string. If there is not, stop and tell me that I have to select a line with a string to be localized.
 
@@ -12,7 +12,7 @@ There should be a selected line with a string. If there is not, stop and tell me
 If the string is already wrapped in a l10n-aware component or useL10n, go on to the XLF portion of these instructions. Otherwise, wrap it in `useL10n2()`. Or `useL10n(english, id)`.
 ```tsx
 // somewhere in Foobar dialog
-<button>{Brighten everything}</button>
+<button>Brighten everything</button>
 ```
 to this:
 ```tsx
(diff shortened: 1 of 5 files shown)


andrew-polk:
[Claude Opus 5 (1M context)] Consulted Devin on 2026-08-21 19:50 UTC, up to commit `fcb685984`.

Devin's review of this commit is clean — no bugs, no Investigate flags, and no informational items. (It was also clean on the previous commit, `478198c40`.) CI (`pr-automation`) is green. Greptile posted only a "trial has ended" notice, so it did not review; CodeRabbit's auto-review is disabled for this repo in `.coderabbit.yml`.

andrew-polk:
[Claude Opus 5 (1M context)] Consulted Devin on 2026-08-21 20:10 UTC, up to commit `510afcfb0`.

Clean again — no bugs, no Investigate flags, no informational items. (Also clean on `478198c40` and `fcb685984`.) CI (`pr-automation`) green.

PR #8224, opened 2026-08-21
Fix unresponsive "Set up levels" after turning on the reader tool (BL-16732)

andrew-polk:
## Problem

**Set Up Levels** in the Leveled Reader tool did nothing at all — no dialog, no message — and
stayed that way until Bloom was restarted. **Set Up Stages** in the Decodable Reader tool had the
same fault. Timing-dependent, so not reproducible on demand.

## Cause

The reader tools remember in module state that they have loaded the collection's reader settings,
but the `ReaderToolsModel` holding those settings is replaced when the frame that created it
reloads. The memory then outlives the data: `beginLoadSynphonySettings()` saw its flags set, found
the settings unchanged, and skipped the work, leaving `synphony` undefined for the rest of the
session. The dialog opened but died on an unguarded `synphony.source` before it could be filled
in, so nothing the user would recognize as a dialog appeared.

## Fix

- Refresh when the model we're holding has no `synphony`, instead of trusting the flags. Root
  cause; goes through the existing refresh path, so no stale word cache and no leaked folder
  watcher.
- `showSetupDialog()` loads the settings itself rather than assuming the tool did.
- Release callers in the three places that could wait forever: a failed settings request, a failed
  sample-texts fetch during refresh, and a throw while building the dialog.
- Guard the `synphony` deref so "the settings were not loaded" is what gets reported.
- `requestWordsForSelectedStage()` tolerates being called before the dialog has its data.
- Tests for the reload into a replaced model, the failed request settling, and the error.

Ref: https://issues.bloomlibrary.org/youtrack/issue/BL-16732

[Devin review](https://devinreview.com/BloomBooks/BloomDesktop/pull/8224)

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8224)

### DIFF
--- a/src/BloomBrowserUI/bookEdit/toolbox/readers/readerSetup/readerSetupDialog.ts
+++ b/src/BloomBrowserUI/bookEdit/toolbox/readers/readerSetup/readerSetupDialog.ts
@@ -10,6 +10,7 @@
 /// <reference path="../readerToolsModel.ts" />
 
 import { getTheOneReaderToolsModel } from "../readerToolsModel";
+import { beginLoadSynphonySettings } from "../readerTools";
 import theOneLocalizationManager from "../../../../lib/localizationManager/localizationManager";
 import { getWorkspaceBundleExports } from "../../../js/workspaceFrames";
 import { get, postBoolean } from "../../../../utils/bloomApi";
@@ -43,6 +44,24 @@ function settingsFrameWindow() {
 let setupDialogElement: JQuery;
 
 export function showSetupDialog(showWhat) {
+    // The dialog is useless without the collection's reader settings: as soon as its
+    // iframe loads, initializeReaderSetupDialog() posts them in, and the dialog uses
+    // them to pick the tab to show and to fill it in. The tool that owns this button
+    // normally loads them when it is activated, but we have no guarantee that has
+    // happened (or has survived) by the time the button is clicked, and if it hasn't,
+    // clicking the button appears to do nothing at all. So make sure they are loaded
+    // before we show anything. This is cheap when they already are. (BL-16732)
+    //
+    // The setTimeout builds the dialog outside the promise chain: a jQuery promise turns
+    // anything thrown in a .then() callback into a rejection nobody is listening for, so a
+    // failure in here would be swallowed and once again show up as a button that does
+    // nothing. Outside the chain, it reaches our usual error reporting.
+    beginLoadSynphonySettings().always(() =>
+        window.setTimeout(() => beginShowSetupDialog(showWhat), 0),
+    );
+}
+
+function beginShowSetupDialog(showWhat) {
     //var toolbox = window;
     theOneLocalizationManager.loadStrings(
         getSettingsDialogLocalizedStrings(),
@@ -176,14 +195,19 @@ function getSettingsDialogLocalizedStrings() {
  * Used by the settings_frame to initialize the setup dialog
  */
 export function initializeReaderSetupDialog() {
+    // Note that synphony itself is undefined until the settings load, so we have to check
+    // it before reaching for source; otherwise the intended error below is pre-empted by a
+    // bare "cannot read properties of undefined" that says nothing about what went wrong.
+    // (BL-16732)
+    const synphony = getTheOneReaderToolsModel().synphony;
     if (
-        typeof getTheOneReaderToolsModel().synphony.source === "undefined" ||
-        getTheOneReaderToolsModel().synphony.source === null
+        !synphony ||
+        synphony.source === undefined ||
+        synphony.source === null
     ) {
         throw new Error("ReaderToolsModel was not loaded with settings");
     }
-    const sourceMsg =
-        "Data\n" + JSON.stringify(getTheOneReaderToolsModel().synphony.source);
+    const sourceMsg = "Data\n" + JSON.stringify(synphony.source);
     const fontMsg = "Font\n" + getTheOneReaderToolsModel().fontName;
     const window = settingsFrameWindow();
     if (window) {
--- a/src/BloomBrowserUI/bookEdit/toolbox/readers/readerTools.ts
+++ b/src/BloomBrowserUI/bookEdit/toolbox/readers/readerTools.ts
@@ -82,15 +82,7 @@ function loadReaderSettingsWithRetry(
     onLoaded: (settings: ReaderSettings) => void,
     onFailed: () => void,
 ): void {
-    get("readers/io/readerToolSettings", (settingsFileContent) => {
-        const normalizedSettings = tryNormalizeReaderSettings(
-            settingsFileContent.data,
-        );
-        if (normalizedSettings) {
-            onLoaded(normalizedSettings);
-            return;
-        }
-
+    const retryOrFail = () => {
         if (attemptsRemaining > 1) {
             window.setTimeout(() => {
                 loadReaderSettingsWithRetry(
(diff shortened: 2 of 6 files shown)


src/BloomBrowserUI/bookEdit/toolbox/readers/readerSetup/readerSetupDialog.ts
```diff
@@ -43,6 +44,24 @@ function settingsFrameWindow() {
...
+    // normally loads them when it is activated, but we have no guarantee that has
+    // happened (or has survived) by the time the button is clicked, and if it hasn't,
+    // clicking the button appears to do nothing at all. So make sure they are loaded
+    // before we show anything. This is cheap when they already are. (BL-16732)
+    //
+    // The setTimeout builds the dialog outside the promise chain: a jQuery promise turns
+    // anything thrown in a .then() callback into a rejection nobody is listening for, so a
+    // failure in here would be swallowed and once again show up as a button that does
+    // nothing. Outside the chain, it reaches our usual error reporting.
+    beginLoadSynphonySettings().always(() =>
```
andrew-polk, line 59: [Devin] **Investigate**: Dialog never shows if settings refresh rejects

`showSetupDialog` gates the dialog on `beginLoadSynphonySettings().always(...)`. In the refresh branch the deferred resolves only in `beginRefreshEverything(...).then(success)` at `readerTools.ts:249-252`, and `beginRefreshEverything` can return an axios promise for the sample-texts list (`readerTools.ts:383-387`). If that request rejects, the deferred never settles, `.always` never fires, and the button silently does nothing again — the symptom this PR targets, reachable now that the dialog waits on the refresh.

src/BloomBrowserUI/bookEdit/toolbox/readers/readerTools.ts
```diff
@@ -103,7 +95,35 @@ function loadReaderSettingsWithRetry(
         }
 
         onFailed();
-    });
+    };
+
+    get(
```
andrew-polk, line 100: [Devin] **Investigate**: Synchronous load errors now swallowed as retries

`bloomApi.get` runs `.then(successCallback).catch(errorCallback)`, and `retryOrFail` is now the error callback. If the success handler throws — e.g. a synchronous failure inside `onLoaded`/`initializeSynphony`/`beginRefreshEverything` — the throw is caught and treated as a failed attempt, retried, then routed to `onFailed`. Previously no error callback meant such throws propagated and were reported, so this weakens the fail-fast behavior the repo prefers.

src/BloomBrowserUI/bookEdit/toolbox/readers/readerTools.ts
andrew-polk, line 1: [Devin] **Investigate**: Reset path can double-init during an in-flight load

(Raised against `readerTools.ts:218-221` as of commit `5cf59d7`; that line has since moved, hence the file-level anchor.)

The new guard at `readerTools.ts:218-221` resets the init flags whenever the model has no `synphony`. `readerToolsInitialized` is set true at `readerTools.ts:237` before the async settings load resolves, so a second call to `beginLoadSynphonySettings` during that window sees `synphony` still undefined, resets, and re-enters full init. `initializeSynphony` then starts a second `DirectoryWatcher` (`readerTools.ts:286-294`) that is never stopped. The window is narrow, but `showSetupDialog` now always calls `beginLoadSynphonySettings`, making concurrent entry newly reachable.

src/BloomBrowserUI/bookEdit/toolbox/readers/readerSetup/readerSetupDialog.ts
andrew-polk: [Claude Opus 5 from Andrew Polk's machine during preflight] Fixed in `832f224ba`. Good catch — this was the same bug as the one the PR is about, just one branch over.

`beginRefreshEverything` now resolves the deferred whether the sample-texts request succeeds or fails. When it fails we have no settings worth remembering, but the callers still get released, so the dialog opens and reports "the settings were not loaded" instead of the button doing nothing.

src/BloomBrowserUI/bookEdit/toolbox/readers/readerTools.ts
andrew-polk: [Claude Opus 5 from Andrew Polk's machine during preflight] Fixed in `832f224ba`. Correct, and this repo does explicitly prefer failing fast — the error callback was catching more than the failed request it was meant for.

The reply is now acted on from inside a `setTimeout`, which puts `onLoaded` (and everything it reaches — `initializeSynphony`, `beginRefreshEverything`) outside the request's promise chain. A genuine bug in there now propagates and gets reported, and only an actual request failure counts as a failed attempt.

src/BloomBrowserUI/bookEdit/toolbox/readers/readerTools.ts
andrew-polk: [Claude Opus 5 from Andrew Polk's machine during preflight] Already addressed in `315bc7c05`, the commit after the one this was raised against — an in-session review had reached the same conclusion, including the never-stopped `DirectoryWatcher`.

The flag-resetting version of the staleness check is gone. Instead, the "already initialized" branch forces `shouldRefresh` when the model has no `synphony`, so a reload goes through `beginRefreshEverything` — which creates no `DirectoryWatcher` at all, and does the `LanguageData`/word-cache cleanup a reload needs. A second caller arriving mid-load now behaves exactly as it did before this branch, so the double-init window is closed rather than narrowed.

src/BloomBrowserUI/bookEdit/toolbox/readers/readerTools.ts
```diff
@@ -216,16 +236,34 @@ export function beginLoadSynphonySettings(): JQueryPromise<void> {
...
+                    !getTheOneReaderToolsModel().synphony;
                 if (!shouldRefresh) {
                     result.resolve();
                     return;
                 }
-                beginRefreshEverything(normalizedSettings).then(() => {
-                    lastReaderToolSettingsContent = newSettingsContent;
-                    result.resolve();
-                });
+                beginRefreshEverything(normalizedSettings).then(
```
andrew-polk, line 256: [Devin] **Investigate**: Replaced-model refresh skips restoreState and directory watcher

On a replaced model (synphony undefined), `beginLoadSynphonySettings` now refreshes through `beginRefreshEverything` (readerTools.ts:376) rather than `initializeSynphony` (readerTools.ts:262). The refresh path sets synphony but never calls `restoreState()` or re-establishes the Sample Texts `DirectoryWatcher`, so a reloaded frame gets settings without persisted stage/level state or an active sample-file watcher. Strictly better than the pre-fix dead state, and consistent with the existing reload path, so not flagged.

andrew-polk: [Claude Opus 5 from Andrew Polk's machine during preflight] Accurate observation, and we're deliberately leaving it — Devin's own last sentence reaches the same conclusion.

Both gaps are the behavior of the pre-existing reload path, which Bloom already takes on every book switch; this change only routes one more case through it, so neither is a regression. And in practice:

- **The stage/level does get restored.** The tool's own `beginRestoreSettings` sets the level from the book's saved `leveledReaderState` (or `readers/io/defaultLevel`) right after this resolves, so `restoreState()` not running here isn't visible.
- **Re-establishing the watcher here is what we just removed.** Nothing stops the previous `DirectoryWatcher`, so creating one on this path is how you end up with orphans polling forever and firing duplicate rebuilds — the problem fixed in `315bc7c05`. Doing it safely means teaching the model to own a single watcher, which is a change to every book switch and well outside this bug.

Left alone on purpose; a reloaded frame gets working reader tools, which is the point.

andrew-polk:
[Claude Opus 5 from Andrew Polk's machine during preflight] Consulted Devin on 2026-08-21 23:50 UTC up to commit `832f224ba`.

No bugs. Four Investigate flags across the run, each with its own resolved thread above: three were real and are fixed (the dialog never showing if the settings refresh rejects; a real bug in the load handler being swallowed as a retry; the earlier flag-resetting staleness check double-initializing during an in-flight load), and the fourth — a replaced-model refresh not calling `restoreState()` or re-creating the Sample Texts watcher — we deliberately left, since both are the pre-existing reload path's behavior and re-creating the watcher is what caused the orphan we just removed. One informational item (a refresh failure resolving quietly but self-healing on the next call), not mirrored.

CI (`pr-automation`) green. Greptile is not reviewing — its trial has ended. CodeRabbit's auto-review is disabled in `.coderabbit.yml`.

StephenMcConnel (approved):
> @StephenMcConnel reviewed 6 files and all commit messages, and resolved 3 discussions.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8224)* status: all files reviewed, 1 unresolved discussion.

StephenMcConnel (approved):
> @StephenMcConnel resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8224)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on andrew-polk).

PR #8215, opened 2026-08-18
Add analytics for the features new in 6.5 (BL-16716)

JohnThomson:
## Problem

Bloom's 6.5 features shipped essentially blind. All 41 existing analytics events came from C#, so
anything implemented in React could only be measured by inventing a bespoke endpoint for it —
which in practice meant it wasn't measured. Nothing under `Publish/Rab` reported anything at all,
the AI image editor and the new image chooser reported nothing, and `Change Picture` — the one
event covering how a picture gets into a book — was bypassed by *both* new 6.5 image routes, so
it had been quietly under-counting.

A second, subtler problem made all of this hard to fix safely: DesktopAnalytics is constructed
with `allowTracking: false` in DEBUG and acts on that *inside* its own `Track` method, so a new
event was easy to write, easy to believe in, and impossible to observe without shipping to alpha.

## Fix

The test applied to every candidate was **whether the answer would change a decision.** Counting
use of a capability we already know is essential does not; nor does measuring a step whose
direction is settled. What does is data that splits a known behaviour into actionable parts.

**Plumbing**

- `POST analytics/track` plus a `trackEvent()` helper, so front-end code can report an event at
  all. C# fills in `BookId`, since React in the edit view does not know it. Not the branding: every
  event has always carried that as `BrandingProjectName`, an application property set per
  collection, and it holds the richer subscription descriptor rather than the branding-folder key.
- `BloomAnalytics` wraps DesktopAnalytics and **logs every event before handing it on** — to the
  Bloom log always, and to stderr when tracking is off, so a developer running `./go.sh` watches
  events scroll past. All 53 call sites go through it, and `build/check-csharp-analytics.sh` keeps
  it that way: a partial log would be worse than none, since a missing line would mean "not
  instrumented" as readily as "did not happen".

**The events, by what they answer**

- **Where pictures come from, and whether searches succeed** — `Image Search`, `Image Chooser
  Closed`, `Image Source Unavailable`, `Pixabay Key Saved`, `Image Preview Slow`, and `Change
  Picture` with `source`/`provider` from all four routes (repairing the under-count on the way
  past). Result counts can't answer the real question: Pixabay and Openverse nearly always return
  *some* pictures, so the failure is a full page of wrong ones, and only what the user did next
  reveals it.
- **Whether the AI editor's funnel converts, and what it costs** — open, generate, commit,
  cancel, key-saved, unavailable. Generation is reported by the editor itself over a new bridge
  message; a session that generates and then throws everything away is our clearest quality
[...]

PR #8228, opened 2026-08-24
Analytics for the features new in 6.5 (BL-16716)

JohnThomson:
## Problem

Bloom's 6.5 features shipped essentially blind. Every existing analytics event came from C#, so
anything implemented in React could only be measured by inventing a bespoke endpoint for it —
which in practice meant it wasn't measured at all. Nothing under `Publish/Rab` reported anything,
the AI Image Editor and the new image chooser reported nothing, and `Change Picture` — the one
event covering how a picture gets into a book — was bypassed by *both* new 6.5 image routes, so
even the old count was quietly low.

## Cause

A second, subtler problem made this hard to fix safely. DesktopAnalytics decides whether to send
an event *inside* its own `Track` method, and Bloom constructs it with `allowTracking: false` in
DEBUG. So a developer could not tell "fired but not sent" from "never fired" — a new event was
easy to write, easy to believe in, and impossible to observe without shipping to alpha.

## Fix

Seven events, chosen by one test: **would the answer change a decision?** Counting use of a
capability we already know is essential does not. Splitting a known behaviour into actionable
parts does.

**Plumbing**

- `POST analytics/track` plus a `trackEvent()` helper, so front-end code can report an event at
  all. C# fills in `BookId`, which React in the edit view does not know. Not the branding: every
  event already carries `BrandingProjectName`, which holds the richer subscription descriptor.
- `BloomAnalytics` wraps DesktopAnalytics and **logs every event as well as sending it**, so a
  developer running `./go.sh` watches events scroll past. Every call site goes through it, and
  `build/check-csharp-analytics.sh` (run from the existing pre-commit hook) keeps it that way: a
  partial log would be worse than none, since a missing line would read as "not instrumented" as
  readily as "did not happen".
- `BloomDebug.Fail` (new, in `Utils`) is the assertion the endpoint uses when a caller sends a
  request with no event name. `Debug.Fail` could not be: with no debugger attached it is
  `Environment.FailFast`, which kills Bloom outright, and mid-suite kills the test host. This one
  breaks into the debugger when there is one, fails the test when under test, and otherwise shows a
  dialog offering to attach a debugger. In a release build the call compiles away entirely. It is
  written for wider use later, not just this call site.

**The events, by what they answer**

- **Where pictures come from** — `Change Picture` gains a `source` field naming the route
  (`pixabay`, `openverse`, a collection slug, `local-disk`, `clipboard`, `ai-editor`), repairing
  the under-count on the way past.
- **Whether searching works** — `Image Chooser Closed`, one event per visit to the chooser. A
  result count could never answer this: Pixabay and Openverse nearly always return *something*,
  so the failure is a full page of wrong pictures, and only what the user did next reveals it.
- **What AI costs and whether it lands** — `AI Image Editor Generate` (from the ai-editor over
  the existing bridge: cost, model, retries, failures) and `AI Image Editor Closed`, one event
  per session, where an `appliedCount` of zero *is* the abandoned case.
- **A Pro feature with no usage data** — `Cover Layout Changed`, the baseline nobody has after
  two releases of charging for custom cover layouts.
- **A heuristic we cannot otherwise evaluate** — `Image Transparency Set`. Every explicit choice
  is a user telling us which way our line-art detection was wrong.
- **A path CI cannot run** — `Publish App`, with the stage as a field. CI never runs a real
  Reading App Builder build, so field telemetry is this path's only continuous coverage.

**One fix that is not analytics.** "Apply this information to all images" re-read every picture's
metadata once per picture — about 160,000 reads on a 400-picture book, appearing to freeze Bloom
for minutes. Found while instrumenting that operation. The event that found it is not being kept;
the fix is.

**No search terms, and nothing persisted.** The image gallery still hands Bloom the search term, so
revisiting that is a one-property change, but it is dropped at Bloom's boundary deliberately rather
than by omission. And this PR writes nothing to the user's disk: the durable count of image-chooser
visits it used to keep was dropped once we established that, since Bloom identifies users to Segment,
a user's earlier visits are simply their earlier events.

**One dependency bump comes with it.** `Image Chooser Closed` reads a search-report callback that
exists in `bloom-image-gallery` 0.0.4 and not in 0.0.3, so this advances the lockfile to the
gallery's current tip (`e376463b`, "Let the host see searches and where an image came from"). Six
lines. Worth a reviewer's eye only because the gallery is declared with no tag, so the lockfile is
the only thing pinning it, and pnpm holds a git dependency at its locked commit until someone runs
an explicit update — which is why master sat on 0.0.3 for five days after that gallery PR merged.

## Scope

This supersedes **#8215**, which instrumented 22 events. That set was reviewed and judged too
large; thirteen event names were dropped and four merged into two. The parked
`BL-16716-analytics` branch keeps the dropped code should any of it be wanted back, and the
[event catalogue](https://bloombooks.github.io/dev-process-artifacts/reports/BL-16716-event-catalogue.html) records
what each dropped event would have told us.

Ref: https://issues.bloomlibrary.org/youtrack/issue/BL-16716

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8228)

[Devin review](https://devinreview.com/BloomBooks/BloomDesktop/pull/8228)

### DIFF
--- a/src/BloomBrowserUI/react_components/image-gallery/ImageGalleryDialog.tsx
+++ b/src/BloomBrowserUI/react_components/image-gallery/ImageGalleryDialog.tsx
@@ -20,7 +20,6 @@ import {
 import { kBloomBlue } from "../../bloomMaterialUITheme";
 import BloomMessageBoxSupport from "../../utils/bloomMessageBoxSupport";
 import { getEditablePageBundleExports } from "../../bookEdit/js/workspaceFrames";
-import { useMountEffect } from "../../utils/useMountEffect";
 import { ShowEditViewDialog } from "../../bookEdit/workspaceRoot";
 
 // The shape of what C# returns from imageGallery/imageGalleryResult
@@ -59,9 +58,6 @@ const ImageGalleryDialog: React.FunctionComponent<{
     // Pixabay's site -- so whether this session had one is the headline number for how much of
     // an obstacle that is.
     const pixabayKeyPresentRef = useRef(false);
-    // How many times this user had opened the chooser before now. A key (or an accept) on the
-    // first visit is a speed bump; one that takes eight visits is a real barrier.
-    const priorChooserSessionsRef = useRef<number | undefined>(undefined);
     // Exactly one "Image Chooser Closed" event per dialog session.
     const closeReportedRef = useRef(false);
 
@@ -84,26 +80,6 @@ const ImageGalleryDialog: React.FunctionComponent<{
             })
             .finally(() => setKeysLoaded(true));
     }, []);
-
-    // Unlike the fetch above, this one WRITES, so it has to happen exactly once per dialog.
-    // React StrictMode runs mount effects twice in development, which would bump the durable
-    // counter by two on every visit -- skewing the very number this exists to provide, and
-    // doing it precisely when a developer is checking the feature.
-    const sessionCountedRef = useRef(false);
-    useMountEffect(() => {
-        if (sessionCountedRef.current) return;
-        sessionCountedRef.current = true;
-        getAsync("app/userSetting?settingName=ImageChooserSessionCount").then(
-            (r) => {
-                const priorSessions = (r?.data?.settingValue as number) ?? 0;
-                priorChooserSessionsRef.current = priorSessions;
-                postJsonAsync("app/userSetting", {
-                    settingName: "ImageChooserSessionCount",
-                    settingValue: priorSessions + 1,
-                });
-            },
-        );
-    });
 
     // Searches are counted, not reported one by one: how many a visit took and which sources it
     // tried are what the close event needs, and a per-query event adds nothing on top of them.
@@ -141,7 +117,6 @@ const ImageGalleryDialog: React.FunctionComponent<{
             providersTried: providersTriedRef.current.size,
             providers: [...providersTriedRef.current].join(","),
             pixabayKeyPresent: pixabayKeyPresentRef.current,
-            priorChooserSessions: priorChooserSessionsRef.current,
         });
     };
 
--- a/src/BloomBrowserUI/react_components/image-gallery/ImageGalleryDialog.tsx
+++ b/src/BloomBrowserUI/react_components/image-gallery/ImageGalleryDialog.tsx
@@ -1,7 +1,11 @@
 import { css } from "@emotion/react";
 import { ImageGallery } from "bloom-image-gallery";
-import type { IImage, IProviderKeysV1 } from "bloom-image-gallery";
-import React, { useEffect, useState } from "react";
+import type {
+    IImage,
+    IProviderKeysV1,
+    ISearchReport,
+} from "bloom-image-gallery";
+import React, { useEffect, useRef, useState } from "react";
 import {
     BloomDialog,
     DialogTitle,
@@ -11,10 +15,12 @@ import {
     getAsync,
     postJsonAsync,
     postDataWithConfigAsync,
+    trackEvent,
 } from "../../utils/bloomApi";
 import { kBloomBlue } from "../../bloomMaterialUITheme";
 import BloomMessageBoxSupport from "../../utils/bloomMessageBoxSupport";
 import { getEditablePageBundleExports } from "../../bookEdit/js/workspaceFrames";
+import { useMountEffect } from "../../utils/useMountEffect";
 import { ShowEditViewDialog } from "../../bookEdit/workspaceRoot";
 
 // The shape of what C# returns from imageGallery/imageGalleryResult
@@ -24,6 +30,9 @@ interface IImageGalleryApiResult {
     creator: string;
     license: string;
 }
+
+// The provider id the gallery stamps on an image the user opened from their own disk.
+const kLocalDiskProviderId = "local-disk";
 
 const ImageGalleryDialog: React.FunctionComponent<{
     img: HTMLElement;
@@ -37,6 +46,25 @@ const ImageGalleryDialog: React.FunctionComponent<{
     >(undefined);
     const [keysLoaded, setKeysLoaded] = useState(false);
 
+    // ----- Analytics for this visit to the chooser (BL-16716) -----
+    // The unit of analysis is the visit, not the query: people usually run several searches
+    // before accepting an image, and a search's outcome is only known later. So a visit
+    // accumulates what it did in the refs below and reports it all in one event when it closes.
+    // Refs, not state: they are written from callbacks, read at close time, and must never cause
+    // a re-render of the gallery.
+    const searchCountRef = useRef(0);
+    // In the order the user tried them, which is more telling than the bare count.
+    const providersTriedRef = useRef(new Set<string>());
+    // Pixabay is the one source a user cannot simply use -- it needs an API key fetched from
+    // Pixabay's site -- so whether this session had one is the headline number for how much of
+    // an obstacle that is.
+    const pixabayKeyPresentRef = useRef(false);
+    // How many times this user had opened the chooser before now. A key (or an accept) on the
+    // first visit is a speed bump; one that takes eight visits is a real barrier.
+    const priorChooserSessionsRef = useRef<number | undefined>(undefined);
+    // Exactly one "Image Chooser Closed" event per dialog session.
+    const closeReportedRef = useRef(false);
+
     // useEffect justified: this is a one-time async fetch that must run after mount
     // so the component can render before the network round-trip completes.
     // There are no dependencies to react to; [] is correct.
@@ -46,7 +74,9 @@ const ImageGalleryDialog: React.FunctionComponent<{
                 const json = r?.data?.settingValue as string;
                 if (json) {
                     try {
-                        setProviderKeys(JSON.parse(json) as IProviderKeysV1);
+                        const keys = JSON.parse(json) as IProviderKeysV1;
+                        setProviderKeys(keys);
+                        pixabayKeyPresentRef.current = !!keys.pixabay;
                     } catch {
                         // ignore malformed stored value
                     }
@@ -55,7 +85,68 @@ const ImageGalleryDialog: React.FunctionComponent<{
             .finally(() => setKeysLoaded(true));
     }, []);
 
+    // Unlike the fetch above, this one WRITES, so it has to happen exactly once per dialog.
+    // React StrictMode runs mount effects twice in development, which would bump the durable
+    // counter by two on every visit -- skewing the very number this exists to provide, and
+    // doing it precisely when a developer is checking the feature.
+    const sessionCountedRef = useRef(false);
+    useMountEffect(() => {
+        if (sessionCountedRef.current) return;
+        sessionCountedRef.current = true;
+        getAsync("app/userSetting?settingName=ImageChooserSessionCount").then(
+            (r) => {
+                const priorSessions = (r?.data?.settingValue as number) ?? 0;
+                priorChooserSessionsRef.current = priorSessions;
+                postJsonAsync("app/userSetting", {
+                    settingName: "ImageChooserSessionCount",
+                    settingValue: priorSessions + 1,
+                });
+            },
+        );
+    });
+
+    // Searches are counted, not reported one by one: how many a visit took and which sources it
+    // tried are what the close event needs, and a per-query event adds nothing on top of them.
+    //
+    // WE DO NOT SEND THE SEARCH TERM. It is the only free-form user text this instrumentation ever
+    // had, an earlier round of BL-16716 did send it, and we were then told not to. It is dropped
+    // here, at Bloom's boundary, deliberately rather than by omission -- `report.term` is still
+    // handed to us by the image gallery (see ISearchReport), because if the decision is revisited
+    // the change is one property and nothing in the gallery has to move.
+    //
+    // What that costs us: we cannot see WHICH subjects people search for and never find, which
+    // was the original argument for the term; and we cannot tell one idea tried in three
+    // languages from three different ideas, so searchCount counts queries and nothing finer.
+    const handleSearch = (report: ISearchReport) => {
+        searchCountRef.current++;
+        providersTriedRef.current.add(report.providerId);
+    };
+
+    // What turns a list of searches into a success rate. Note that a result count could never
+    // do this job: Pixabay and Openverse nearly always return *some* pictures, so the failure
+    // we care about isn't an empty page, it's a full page of pictures that are all wrong -- and
+    // the only reliable sign of that is what the user did next.
+    const reportChooserClosed = (
+        outcome: "accepted" | "local disk" | "cancelled",
+        acceptedProvider?: string,
+    ) => {
+        if (closeReportedRef.current) return;
+        closeReportedRef.current = true;
+        trackEvent("Image Chooser Closed", {
+            outcome,
+            acceptedProvider,
+            // No acceptedTerm: see handleSearch above. Which SOURCE satisfied the user is the part
+            // that makes this a success rate, and that is not user text.
+            searchCount: searchCountRef.current,
+            providersTried: providersTriedRef.current.size,
+            providers: [...providersTriedRef.current].join(","),
+            pixabayKeyPresent: pixabayKeyPresentRef.current,
+            priorChooserSessions: priorChooserSessionsRef.current,
+        });
+    };
+
     const handleClose = () => {
+        reportChooserClosed("cancelled");
         setOpen(false);
     };
 
@@ -86,6 +177,14 @@ const ImageGalleryDialog: React.FunctionComponent<{
                 license: result.license,
                 undoable: "true",
             });
+            // A file the user opened from disk is a different outcome from finding a picture in
+            // one of the collections we offer, even though both end in an image being used.
+            reportChooserClosed(
+                image.providerId === kLocalDiskProviderId
+                    ? "local disk"
+                    : "accepted",
+                image.providerId,
+            );
             setOpen(false);
         } catch {
             BloomMessageBoxSupport.CreateAndShowSimpleMessageBox(
@@ -116,6 +215,11 @@ const ImageGalleryDialog: React.FunctionComponent<{
             thumbnailUrl: previewUrl,
             reasonableSizeUrl: previewUrl,
             localPath: filePath,
+            // Set here as well as by the gallery, so that the "local disk" outcome and the
+            // Change Picture provider do not silently depend on the package continuing to stamp
+            // it. If it ever stopped, every disk-opened picture would be recorded as an ordinary
+            // chooser accept from an unknown provider -- the exact split this event exists for.
+            providerId: kLocalDiskProviderId,
             // C# reports the original file's dimensions and byte count. These matter because
             // previewUrl may serve a downscaled stand-in for an image too large for the
             // browser to display, and we want to report what the user actually chose.
@@ -174,12 +278,17 @@ const ImageGalleryDialog: React.FunctionComponent<{
                         lang={props.searchLang}
                         initialProviderKeys={providerKeys}
                         primaryColor={kBloomBlue}
-                        onProviderKeysChange={(keys) =>
+                        onSearch={handleSearch}
+                        onProviderKeysChange={(keys) => {
+                            // Kept up to date mid-visit as well as read at startup, so that a
+                            // key supplied while the chooser is open is reflected in what this
+                            // visit reports.
+                            pixabayKeyPresentRef.current = !!keys.pixabay;
                             postJsonAsync("app/userSetting", {
                                 settingName: "ImageGalleryProviderKeys",
                                 settingValue: JSON.stringify(keys),
-                            })
-                        }
+                            });
+                        }}
                         onLanguageChange={(lang) =>
                             postJsonAsync("app/userSetting", {
                                 settingName: "ImageSearchLanguage",
--- a/src/BloomBrowserUI/react_components/image-gallery/ImageGalleryDialog.tsx
+++ b/src/BloomBrowserUI/react_components/image-gallery/ImageGalleryDialog.tsx
@@ -70,6 +70,9 @@ const ImageGalleryDialog: React.FunctionComponent<{
                 licenseUrl: image.licenseUrl,
                 credits: image.credits,
                 creator: image.creator,
+                // Which source the picture came from, so C#'s "Change Picture" event can say
+                // where pictures actually come from rather than only how many there were.
+                provider: image.providerId,
             };
             const response = await postJsonAsync(
                 "imageGallery/imageGalleryResult",
--- a/build/check-csharp-analytics.sh
+++ b/build/check-csharp-analytics.sh
@@ -1,0 +1,65 @@
+#!/bin/sh
+# This script is run by the pre-commit hook in src/BloomBrowserUI/.vite-hooks/pre-commit. We need to get to
+# the root of the git repository, which is one level up from where this script lives.
+cd $(dirname $0)/..
+
+missing_dependencies=
+for dependency in git awk tr; do
+  if ! command -v "$dependency" >/dev/null 2>&1; then
+    missing_dependencies="$missing_dependencies $dependency"
+  fi
+done
+
+if [ -n "$missing_dependencies" ]; then
+  echo "Missing required commands for build/check-csharp-analytics.sh:$missing_dependencies"
+  echo "This hook runs in a POSIX shell environment and requires Git plus common Unix tools."
+  echo "If you are committing from Windows, make sure the hook is running under Git Bash or equivalent."
+  exit 1
+fi
+
+echo Checking that analytics events go through BloomAnalytics.
+# Why: DesktopAnalytics decides whether to send an event INSIDE Analytics.Track, and in a DEBUG
+# build it decides not to (Program.InitializeAnalytics passes allowTracking:false). An event that
+# calls it directly is therefore impossible to observe on a developer machine -- it neither sends
+# nor says it didn't. BloomAnalytics.Track logs every event before handing it on, which is the only
+# way to see one without shipping to alpha. That log is only worth reading if it is complete, so a
+# call that bypasses it stops the commit: a missing line has to mean "did not happen" rather than
+# "somebody used the other API".
+#
+# 1) Get the list of C# files that have been staged for commit (added or modified).
+# 2) Skip the wrapper itself, which necessarily calls DesktopAnalytics.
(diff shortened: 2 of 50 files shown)


JohnThomson:
[Claude Opus 5 (1M context) from John Thomson's machine during preflight]

**Consulted Devin** on this PR up to `376a5579cb04be0741b571bf558b467d2745df35`.

**No bugs and no Investigate flags, across two rounds.** So there is nothing mirrored as a review
thread here — that absence is the result, not an oversight.

It did leave **seven informational notes**, which by convention are not posted as threads. Two were
verifications of the things most likely to be wrong, and it is worth recording that it looked:
the AI-editor session-settlement logic, and this branch's new pre-commit analytics guard (its exit
code and its regex). Three more were reasoned confirmations rather than doubts — the picture-source
reporting being split between C# and the browser, the transparency-choice history being
process-lifetime but bounded, and the image-chooser session counter genuinely incrementing rather
than string-concatenating.

Two were questions, and both have been answered:

- *Generated and reused counts include replacements that failed.* Correct, and intentional; those
  figures say what the user chose and therefore paid OpenRouter for, while `appliedCount` and
  `failedCount` say what reached the book. Devin read the intent right and asked for confirmation,
  so the confirmation now lives next to the code rather than in a thread — `376a557`.
- *The analytics guard script skips only leading comments*, so a line of real code followed by an
  inline comment that happens to name `Analytics.Track(` would block a commit. A genuine, narrow
  false positive. Left as it is deliberately: the script prints exactly why it refused, so the fix
  is to rephrase the comment, and a cleverer regex risks the far worse direction — letting a real
  unwrapped call through.

Greptile also ran and posted only "Your trial has ended", so there is no Greptile review of this PR.
CI (`pr-automation`) is green, and both test suites pass locally: 740 front-end tests and 3184 C#.

JohnThomson:
[Claude Opus 5 (1M context) from John Thomson's machine during preflight]

**Consulted Devin** again, up to `510e4ceb2860612f6a0ef8fdeba57d2900049867` (the earlier log above
covered `376a557`).

**Still no bugs and no Investigate flags**, so again nothing is mirrored as a review thread. Eight
informational notes this time. The two new ones since the last log:

- *`analytics/track` is handled off the UI thread*, and `HandleTrack` reads the current book from the
  worker. Devin's own conclusion is that this is safe — the read is of a reference and the send is
  async — with the nuance that an event fired *during* a book switch attaches whichever book is
  current when it sends. Left alone: the window is a few milliseconds and the alternative is putting
  analytics on the UI thread, which is worse.
- *The analytics guard script skips only leading comments*, so a line of real code followed by an
  inline comment that happens to name `Analytics.Track(` would block a commit. A genuine but narrow
  false positive, left deliberately: the script prints exactly why it refused, so the fix is to
  rephrase, and a cleverer regex risks the far worse direction of letting a real unwrapped call
  through.

One note is **stale** and should be ignored: it describes the image-chooser visit counter and its
mount effect, which this same commit removed. Devin appears to have carried it forward from the
previous round.

Also worth recording for anyone reading this PR's history: **`pr-automation` stopped firing on
pushes part way through this run.** It ran for the first two commits and not for the two after, so
Devin only reviewed those because I re-triggered the workflow by hand. On this repo a commit whose
`pr-automation` never fires gets no Devin review at all, and the missing check looks exactly like a
passing one.

src/BloomBrowserUI/react_components/image-gallery/ImageGalleryDialog.tsx
```diff
@@ -1,7 +1,11 @@
 import { css } from "@emotion/react";
 import { ImageGallery } from "bloom-image-gallery";
-import type { IImage, IProviderKeysV1 } from "bloom-image-gallery";
-import React, { useEffect, useState } from "react";
+import type {
+    IImage,
+    IProviderKeysV1,
+    ISearchReport,
```
JohnThomson, line 6: [Claude Opus 5 (1M context) from John Thomson's machine during preflight]

**Devin (Investigate):** "The new `ISearchReport` import and `onSearch` prop require a
`bloom-image-gallery` version that exposes them. The PR description notes master's lockfile still
resolves 0.0.3, which lacks both, and the package is declared without a tag. Confirm the resolved
version is bumped so this compiles."

**Confirmed, and done in this commit.** Devin was reading an earlier draft of the PR description,
written while that was still true; the merge it is analysing is what fixed it.

`pnpm update bloom-image-gallery` moved the lockfile from 0.0.3 to **0.0.4**
(`e376463bcd21b1558750570b63269a2e133b46c0`), which is the gallery repo's current `master` tip and
the merge of the gallery PR that added these two APIs. Six lines of lockfile. The proof that it
compiles is that `pnpm typecheck` passes on this tree — it failed on exactly this import when I
first merged master while the lockfile was still on 0.0.3.

Devin is right that the underlying fragility is real, and it is worth a reader knowing: the gallery
is declared as `github:BloomBooks/bloom-image-gallery` with no tag, and pnpm holds a git dependency
at its locked commit until someone runs an explicit update. That is why master sat on 0.0.3 for five
days after the gallery PR merged, and it is why a plain `pnpm install` does *not* pick the new
version up. Tagging the gallery the way `bloom-ai-image-tools` is tagged would remove the trap, but
that is a change to how a shared dependency is managed rather than part of this card, so it is
raised for the developer rather than done here.

JohnThomson:
[Claude Opus 5 (1M context) from John Thomson's machine during preflight]

**Consulted Devin** up to `bbddb3d72b0ba8bbb7bb51bfc09b6ba7c81cddd6`.

**No bugs.** Twelve informational notes and one Investigate flag — and that flag is
[the gallery-version one already answered and resolved above](https://github.com/BloomBooks/BloomDesktop/pull/8228#discussion_r3846431316),
re-raised unchanged on this commit. Nothing to add: the lockfile bump to gallery 0.0.4 landed with
the base merge and the typecheck passes on it. Not mirrored a second time, because a duplicate
thread saying the same thing would only bury the first.

Two of the informational notes are worth a reader's attention, both left deliberately and both
consistent across runs: the pre-commit analytics guard scans the working tree rather than the
staged blob (real, but it matches its sibling `check-csharp-*.sh` scripts, so consistency argues
against diverging here), and a failed commit still contributes to `generatedCount`/`reusedCount`
— which Devin traced to the comment explaining that those count what the user chose and paid for,
not what landed.

One note is stale and can be ignored: it describes the image-chooser visit counter and its mount
effect, which were removed several commits ago. Devin has carried it forward for three runs now.

andrew-polk (changes requested):
> @andrew-polk reviewed 48 files and all commit messages, and made 5 comments.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8228)* status: all files reviewed, 6 unresolved discussions (waiting on JohnThomson).

___
*[`src/BloomExe/web/controllers/AnalyticsApi.cs` line 14 at r4](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8228#-P-tla8G9qLxsVBaoytn:-P-tla8G9qLxsVBaoyto:b-lg8kgd) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/bbddb3d721540098d5dc13148f25791721f0be00/src/BloomExe/web/controllers/AnalyticsApi.cs#L14)):*
> publish/pdf/printAnalytics did

"did" is misleading since it still does.

I see there are complications with replacing that call with our new one. I'm not necessarily advocating for trying to work through those complications and making the replacement, but it does make me wonder if there anything about the **design** of the API which we should consider changing.
I don't have anything specific in mind, it just smells that our only existing analytics endpoint can't be fully modeled by the new endpoint.
___
*[`src/BloomExe/web/controllers/AnalyticsApi.cs` line 61 at r4](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8228#-P-tmhFj5ccKXK5bDEOD:-P-tmhFj5ccKXK5bDEOE:blt8jyu) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/bbddb3d721540098d5dc13148f25791721f0be00/src/BloomExe/web/controllers/AnalyticsApi.cs#L61)):*
> this endpoint used to do it

This is a new endpoint, so this type of comment seems less than unhelpful.
___
*[`src/BloomBrowserUI/bookEdit/toolbox/canvas/canvasControlRegistry.ts` line 843 at r4](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8228#-P-tqEgFCpbYXDmaIfzh:-P-tqEgFCpbYXDmaIfzi:b-7sbswe) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/bbddb3d721540098d5dc13148f25791721f0be00/src/BloomBrowserUI/bookEdit/toolbox/canvas/canvasControlRegistry.ts#L843)):*
> ```XML
>                         // they saw and was guessing. A lot of those means the three labels do not
>                         // predict the result well enough, which is a UI problem rather than an
>                         // algorithm one.
> ```

What is the "whole sequence of choices"? It seems to be referring to something which isn't here.

Also, it is overstating its theory on what switching means. That **might** be what it means. Or they are just exploring, etc. Anyway, if this comment even fits somewhere, I would just add a "wiggle word" like might/maybe.
___
*[`src/BloomExe/web/controllers/AnalyticsApi.cs` line 53 at r4](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8228#-P-tmU2278qSxbjlvfPb:-P-tmU2278qSxbjlvfPc:b-v79dbz) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/bbddb3d721540098d5dc13148f25791721f0be00/src/BloomExe/web/controllers/AnalyticsApi.cs#L53)):*
> ```Smalltalk
>                 // quietly recording nothing, which is a very hard thing to notice.
>                 request.Failed("analytics/track requires an 'event' name");
>                 return;
> ```

Is it worth adding a heavier development guard here?
Maybe `Debug.Fail`?
___
*[`src/BloomExe/web/controllers/AnalyticsApi.cs` line 73 at r4](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8228#-P-tn5OI6hv8-1c96rOW:-P-tn5OI6hv8-1c96rOX:b-iq6l2e) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/bbddb3d721540098d5dc13148f25791721f0be00/src/BloomExe/web/controllers/AnalyticsApi.cs#L73)):*
> ```Smalltalk
>             // tracking is off, and BloomAnalytics logs only the per-event properties, so on a
>             // developer build the log line shows no branding at all. That is not evidence it is
>             // missing in production. It cost a round trip to learn; please do not re-add it.
> ```

Sorry; I'm probably being dense, but this paragraph is not clicking for me. I'm not sure what I'm supposed to not re-add.

JohnThomson (commented):
> @JohnThomson+AGNT made 5 comments.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8228)* status: 46 of 48 files reviewed, 6 unresolved discussions (waiting on andrew-polk).

___
*[`src/BloomBrowserUI/bookEdit/toolbox/canvas/canvasControlRegistry.ts` line 843 at r4](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8228#-P-tqEgFCpbYXDmaIfzh:-P-ufQr9Ab-FPy_CtRA9:bplmzmq) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/bbddb3d721540098d5dc13148f25791721f0be00/src/BloomBrowserUI/bookEdit/toolbox/canvas/canvasControlRegistry.ts#L843)):*

What is the "whole sequence of choices"? It seems to be referring to something which isn't here.

Also, it is overstating its theory on what switching means. That **might** be what it means. Or they are just exploring, etc. Anyway, if this comment even fits somewhere, I would just add a "wiggle word" like might/maybe.
</blockquote></details>

[Claude Opus 5 (1M context) from John Thomson's machine] Both fair, and the first was a real defect: the comment named a value it never showed you. `path` is every transparency choice made on that picture so far, joined with " > ", so a picture switched twice reads `auto > transparent > opaque`. The comment now says that and names `recordTransparencyChoice`, which builds it.

And you're right that I was asserting a motive I can't observe. It now reads that a long path *may* mean the labels didn't predict the result — with "or they were just exploring" stated alongside, rather than left for the reader to supply.
___
*[`src/BloomExe/web/controllers/AnalyticsApi.cs` line 14 at r4](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8228#-P-tla8G9qLxsVBaoytn:-P-ufP327Y-isPtgXZaY:brq4ofq) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/bbddb3d721540098d5dc13148f25791721f0be00/src/BloomExe/web/controllers/AnalyticsApi.cs#L14)):*

> publish/pdf/printAnalytics did

"did" is misleading since it still does.

I see there are complications with replacing that call with our new one. I'm not necessarily advocating for trying to work through those complications and making the replacement, but it does make me wonder if there anything about the **design** of the API which we should consider changing.
I don't have anything specific in mind, it just smells that our only existing analytics endpoint can't be fully modeled by the new endpoint.
</blockquote></details>

[Claude Opus 5 (1M context) from John Thomson's machine] Fixed: "still does".

On the design smell — I went and looked, and I think it dissolves. `printAnalytics` does two unrelated things:

1. `BloomAnalytics.Track("Print PDF", { BookId, Country })`
2. `CurrentBook.ReportSimplisticFontAnalytics(FontEventType.PublishPdf, "Print PDF")`

The **event** models fine. `BookId` is filled in by the new endpoint anyway, and `Country` is already an application property (`SetAnalyticsProperties` sets it), so it is redundant there exactly like the branding property would be — a front-end `trackEvent("Print PDF")` would produce the same record.

What does not model is (2). Font analytics is a separate pipeline with its own event types, and it needs the `Book`. So it is not that our one existing analytics endpoint resists the new shape; it is that this endpoint isn't only an analytics endpoint — it is one event plus a font-analytics side effect. Nothing about the generic channel needs to grow to cover it.

I have not converted it, per your "not necessarily advocating". If we ever do, the note to self is that `Country` should be dropped rather than passed.
___
*[`src/BloomExe/web/controllers/AnalyticsApi.cs` line 53 at r4](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8228#-P-tmU2278qSxbjlvfPb:-P-ufPXz2DeYVe7SCLlC:b-vesp21) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/bbddb3d721540098d5dc13148f25791721f0be00/src/BloomExe/web/controllers/AnalyticsApi.cs#L53)):*

Is it worth adding a heavier development guard here?
Maybe `Debug.Fail`?
</blockquote></details>

[Claude Opus 5 (1M context) from John Thomson's machine] Agreed on the concern, and I've made it louder — but deliberately not with `Debug.Fail`, so please push back if you disagree.

`Debug.Fail` with no debugger attached becomes `Environment.FailFast`, which takes the process down. That would mean a mis-shaped analytics call can kill Bloom in a developer or alpha build — and "recording an event must never be able to affect what it is observing" is the rule the rest of this feature is built on. It would also abort a test run: the abort markers in `build/test-abort-markers.txt` exist precisely because a `FailFast` mid-suite prints a cheerful "Passed!" for however many tests got to run.

What it does now instead is say it three ways, because each reaches a different person: `request.Failed` to the browser console, a log line to Help > Show Event Log, and standard error to the terminal a developer already has analytics scrolling past in. That covers your "hard thing to notice" without the hammer. If you'd still rather it were fatal in DEBUG, it is a one-liner guarded on `Program.RunningUnitTests` and I'll add it.
___
*[`src/BloomExe/web/controllers/AnalyticsApi.cs` line 61 at r4](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8228#-P-tmhFj5ccKXK5bDEOD:-P-ufPwg7EnRK0GN3XPL:buj4wm7) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/bbddb3d721540098d5dc13148f25791721f0be00/src/BloomExe/web/controllers/AnalyticsApi.cs#L61)):*

> this endpoint used to do it

This is a new endpoint, so this type of comment seems less than unhelpful.
</blockquote></details>

[Claude Opus 5 (1M context) from John Thomson's machine] You're right, and it's worse than unhelpful — it's untrue of the code. What "used to" carry a branding property was an earlier draft of *this PR*. That is the PR's history, not the file's, and it had no business being in a comment. Dropped.
___
*[`src/BloomExe/web/controllers/AnalyticsApi.cs` line 73 at r4](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8228#-P-tn5OI6hv8-1c96rOW:-P-ufQNA-qdflCc31pY3:bmod8fe) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/bbddb3d721540098d5dc13148f25791721f0be00/src/BloomExe/web/controllers/AnalyticsApi.cs#L73)):*

Sorry; I'm probably being dense, but this paragraph is not clicking for me. I'm not sure what I'm supposed to not re-add.
</blockquote></details>

[Claude Opus 5 (1M context) from John Thomson's machine] Not dense — the paragraph never said what "it" was. "It" was a `branding` property on this event, which an earlier draft of the PR added and we removed; three paragraphs later "do not re-add it" had lost its referent entirely.

Rewritten to open with the instruction ("Do NOT add a branding property here"), say why (every event already carries `BrandingProjectName` as an application property, and it holds the richer subscription descriptor), and keep the one fact that makes the mistake recur: on a developer build no branding appears in the log at all, because `SetAnalyticsProperties` returns early when tracking is off — which is not evidence it is missing in production. The sentence about what it cost me to find that out is gone.

andrew-polk (changes requested):
> @andrew-polk reviewed 2 files and all commit messages, made 2 comments, and resolved 3 discussions.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8228)* status: all files reviewed, 3 unresolved discussions (waiting on JohnThomson+AGNT).

___
*[`src/BloomBrowserUI/bookEdit/toolbox/canvas/canvasControlRegistry.ts` line 843 at r4](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8228#-P-tqEgFCpbYXDmaIfzh:-P-ugNu1EveKRvvyNbwF:b-eb3c2a) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/bbddb3d721540098d5dc13148f25791721f0be00/src/BloomBrowserUI/bookEdit/toolbox/canvas/canvasControlRegistry.ts#L843)):*
<details><summary><i>Previously, JohnThomson+AGNT (Agent AGNT for John Thomson) wrote…</i></summary><blockquote>

[Claude Opus 5 (1M context) from John Thomson's machine] Both fair, and the first was a real defect: the comment named a value it never showed you. `path` is every transparency choice made on that picture so far, joined with " > ", so a picture switched twice reads `auto > transparent > opaque`. The comment now says that and names `recordTransparencyChoice`, which builds it.

And you're right that I was asserting a motive I can't observe. It now reads that a long path *may* mean the labels didn't predict the result — with "or they were just exploring" stated alongside, rather than left for the reader to supply.
</blockquote></details>

Sorry to ask for another round here, but "path" is a really bad name. I assumed, and I think it will almost always be assumed to be a file path.
history? choices?
___
*[`src/BloomExe/web/controllers/AnalyticsApi.cs` line 53 at r4](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8228#-P-tmU2278qSxbjlvfPb:-P-uh1e96S_7ZL18PqWv:b-q91j0i) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/bbddb3d721540098d5dc13148f25791721f0be00/src/BloomExe/web/controllers/AnalyticsApi.cs#L53)):*
<details><summary><i>Previously, JohnThomson+AGNT (Agent AGNT for John Thomson) wrote…</i></summary><blockquote>

[Claude Opus 5 (1M context) from John Thomson's machine] Agreed on the concern, and I've made it louder — but deliberately not with `Debug.Fail`, so please push back if you disagree.

`Debug.Fail` with no debugger attached becomes `Environment.FailFast`, which takes the process down. That would mean a mis-shaped analytics call can kill Bloom in a developer or alpha build — and "recording an event must never be able to affect what it is observing" is the rule the rest of this feature is built on. It would also abort a test run: the abort markers in `build/test-abort-markers.txt` exist precisely because a `FailFast` mid-suite prints a cheerful "Passed!" for however many tests got to run.

What it does now instead is say it three ways, because each reaches a different person: `request.Failed` to the browser console, a log line to Help > Show Event Log, and standard error to the terminal a developer already has analytics scrolling past in. That covers your "hard thing to notice" without the hammer. If you'd still rather it were fatal in DEBUG, it is a one-liner guarded on `Program.RunningUnitTests` and I'll add it.
</blockquote></details>

I honestly don't think any of those would be observed by a developer who just broke it. Seems like failing hard is what we want when running in debug or tests.

andrew-polk (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8228)* status: all files reviewed, 2 unresolved discussions (waiting on JohnThomson+AGNT).

JohnThomson (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8228)* status: 47 of 50 files reviewed, 2 unresolved discussions (waiting on andrew-polk).

___
*[`src/BloomExe/web/controllers/AnalyticsApi.cs` line 53 at r4](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8228#-P-tmU2278qSxbjlvfPb:-P-uz8Ww39ezOQzgTO_Y:brzvxiz) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/bbddb3d721540098d5dc13148f25791721f0be00/src/BloomExe/web/controllers/AnalyticsApi.cs#L53)):*

I honestly don't think any of those would be observed by a developer who just broke it. Seems like failing hard is what we want when running in debug or tests.
</blockquote></details>

[Claude Opus 5 (1M context) from John Thomson's machine] You're right, and that settles it — a log line and a console message are not going to reach the developer who just broke it. John agrees, so we've done it properly rather than picking one horn of it.

New `Bloom.Utils.BloomDebug.Fail(string message = null)`, and `analytics/track` now calls it. It picks the loudest thing that suits where Bloom is running:

- **debugger attached** → `Debug.Fail`, exactly as you wanted;
- **under test** → throws, so the offending test fails and says why. This branch matters most: a modal dialog in an automated run blocks until something kills the suite. Two tests pin it;
- **otherwise** → a dialog with the message and the stack, offering to attach a debugger. `Debugger.Launch()` gives us the old `Debug.Fail` "Retry" behaviour for one line, so you get to break on the offending call after all.

It is `[Conditional("DEBUG")]` rather than `#if DEBUG` around the body — the same mechanism `Debug.Fail` itself uses, and stronger, because it removes the *call*, so a release build does not even evaluate the arguments.

It throws rather than calling `Assert.Fail` because BloomExe deliberately does not reference a test framework; an exception is how NUnit reports a failure anyway.

The log line and the failed request stay alongside it, since in a release build `BloomDebug` compiles away entirely and they are all that is left to record that a caller is broken.

John's intent is that this eventually replaces `Debug.Fail` across the codebase, with a sibling for `Debug.Assert`, which has the same FailFast problem. For now it is only used where you asked for it.

JohnThomson (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8228)* status: 46 of 50 files reviewed, 2 unresolved discussions (waiting on andrew-polk).

___
*[`src/BloomBrowserUI/bookEdit/toolbox/canvas/canvasControlRegistry.ts` line 843 at r4](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8228#-P-tqEgFCpbYXDmaIfzh:-P-v-kyCEB03uy0gAhMH:bi6zapt) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/bbddb3d721540098d5dc13148f25791721f0be00/src/BloomBrowserUI/bookEdit/toolbox/canvas/canvasControlRegistry.ts#L843)):*

Sorry to ask for another round here, but "path" is a really bad name. I assumed, and I think it will almost always be assumed to be a file path.
history? choices?
</blockquote></details>

[Claude Opus 5 (1M context) from John Thomson's machine] No apology needed — you're right, and the giveaway is that the rest of this code already agreed with you. `recordTransparencyChoice` builds it and `transparencyChoicesByImage` stores it; only the event field was called something else. So it's `choices` now, and the two comments that used "path" to mean the sequence say "sequence" and "the whole of it" instead.

Worth noting for anyone reading later: it's a stronger objection than a style nit, because a few lines up this same file really is handling image file paths (`imageFileOf`, `imageHistoryKeyOf`). The name collided with a live neighbouring concept, not just with a general expectation.

Nothing has shipped, so there's no recorded data to migrate. Catalogue updated to match: https://bloombooks.github.io/dev-process-artifacts/reports/BL-16716-event-catalogue.html

JohnThomson (commented):
> @JohnThomson reviewed 50 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8228)* status: 44 of 50 files reviewed, 2 unresolved discussions (waiting on andrew-polk).

andrew-polk (approved):
> @andrew-polk reviewed 6 files and all commit messages, and resolved 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8228)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on JohnThomson).

JohnThomson:
[Claude Opus 5 (1M context) from John Thomson's machine during preflight]

**Consulted Devin** up to `6ac6337a38c2a419ba97b215783eff66d0c8e34b`.

**No bugs.** One Investigate flag and sixteen informational notes.

The Investigate flag is the gallery-version one [already mirrored and answered above](https://github.com/BloomBooks/BloomDesktop/pull/8228#discussion_r3846431316) — the lockfile is on 0.0.4 and `pnpm typecheck` passes, so it is settled; nothing new to mirror.

Two informational notes were worth checking rather than filing, so for the record:

- Devin asked to confirm that `generatedCount`/`reusedCount` deliberately count every replacement the AI Image Editor sent, not only the ones that landed. They do, and the code says so where it happens: those two answer "what did the user choose, and therefore pay OpenRouter for", while `appliedCount` and `failedCount` are the pair that says what reached the book. Not comparable on purpose.
- One note describes an `ImageChooserSessionCount` setting round-tripping as an int. That setting no longer exists — it was removed earlier in this branch once we established that Bloom identifies users to Segment, so a user's earlier visits are simply their earlier events. Nothing to act on.

CI is green (`pr-automation` pass), the front-end suite is 753 passed / 5 skipped, and the C# suite is 3247 passed / 0 failed at this commit.

PR #8230, opened 2026-08-25
Allow image placeholders to get "edit with AI" (BL-16744)

hatton:
### Problem

"Edit with AI..." was offered only on a picture box that already held an image, so a user who
wanted the AI to make a picture for an **empty** box could not start there. When Bloom did hand the
editor a book's pictures, they arrived as an unnamed list, so on a book with several pictures the
user could not tell which was which. And an image the user accepted could not be undone.

### What the PR does

- Enables "Edit with AI..." on an image container that has no image yet.

image

- Gives every slot Bloom sends a name the user can read: "Page 3", the page's own name for front
  and back matter, "Page 3 - Canvas Background" for the picture behind a page, and
  "Page 3 - Image 2" for the pictures on top of it. The ai-image-editor labels each slot with that
  name.

image

- Identifies a slot by its position among the page's image containers, so the box the user clicked
  is the box the editor works on and fills, even on a page holding several empty ones. That index
  is the slot's whole identity, and C# and the page frame number the same containers.
- Makes accepting an image **undoable**, like Paste Image. The page is no longer saved as soon as
  the editor closes, because that save discarded the undo.
- Pins the AI image editor at `dist-v0.1.7`, the release that keeps a placeholder out of "Image to
  Edit", opens the tool that makes an image from a description, and returns the image to the box
  the user launched on.

Ref: https://issues.bloomlibrary.org/youtrack/issue/BL-16744

[Devin review](https://devinreview.com/BloomBooks/BloomDesktop/pull/8230)

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8230)


src/BloomBrowserUI/bookEdit/aiImageEditor/aiEditorOverlay.ts
```diff
@@ -121,11 +121,17 @@ export function openAiImageEditor(target: IAiImageEditorTarget): void {
...
+        // An empty placeholder slot is sent like any other (BL-16744). It used to be
+        // withheld, on the grounds that an empty slot has nothing to edit — but the AI
+        // image editor answers a missing selectedBookImageId by targeting the FIRST
+        // image of the book, which is normally the front cover. So withholding it aimed
+        // the user at the cover when they had asked for an empty slot on some other page.
+        // The editor reads isPlaceholder on the named slot and, for an empty one, puts
+        // nothing in its "Image to Edit" panel and opens its "Create an Image" tool
+        // instead; it keeps the slot so the created image can be committed straight into
+        // it. That behavior needs a bloom-ai-image-tools build newer than dist-v0.1.5,
+        // so the dep in package.json has to be bumped for this to work in a real Bloom.
```
hatton: [Devin] **Investigate**: Menu enabled on placeholder before editor dependency is bumped

[Claude Opus 5 from Hatton's machine during preflight — mirroring a Devin finding]

Bloom now enables "Edit with AI..." on an empty image placeholder and sends that slot as
`selectedBookImageId`. The editor behavior that answers this — leave "Image to Edit" empty, open
the "Create an Image" tool, keep the slot for the commit — lives in the separate
`bloom-ai-image-tools` repository. Bloom pins that editor as `dist-v0.1.5` in
`src/BloomBrowserUI/package.json`, and this PR does not bump it.

So until a new `dist-` tag exists and the dependency is bumped, a user who picks "Edit with AI..."
on an empty slot reaches an editor that treats the placeholder graphic as the picture to edit.

The editor-side work is written and its tests pass, but it is not yet committed or tagged. This
thread stays open until the tag and the bump land.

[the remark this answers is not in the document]
hatton: [Claude Opus 5 from Hatton's machine during preflight]

Correct, and deliberate. Devin raised the same point independently.

The editor half of this work is written in `bloom-ai-image-tools` and its tests pass, but it is
not committed or tagged yet. Bloom cannot bump `src/BloomBrowserUI/package.json` until a
`dist-` tag exists to bump it to.

This is now a decision item for the developer, so this thread stays open until the tag and the
dependency bump land.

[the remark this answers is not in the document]
hatton: [Claude Opus 5 from Hatton's machine during preflight]

Fixed. The editor half is released and Bloom now points at it: `bloom-ai-image-tools` 0.1.6 is tagged as `dist-v0.1.6`, and `src/BloomBrowserUI/package.json` pins that tag as of dc89512c58.

That release is what makes an empty box a real target: the editor keeps the placeholder graphic out of the "Image to Edit" panel, opens the tool that makes an image from a description, remembers the box the user launched on, and puts the created image back into it. It also shows the per-slot labels this PR sends. So the menu item and the editor now agree.

hatton: [Claude Opus 5 from Hatton's machine during preflight]

Fixed. The editor half is released and Bloom now points at it: `bloom-ai-image-tools` 0.1.6 is tagged as `dist-v0.1.6`, and `src/BloomBrowserUI/package.json` pins that tag as of dc89512c58.

That release is what makes an empty box a real target: the editor keeps the placeholder graphic out of the "Image to Edit" panel, opens the tool that makes an image from a description, remembers the box the user launched on, and puts the created image back into it. It also shows the per-slot labels this PR sends. So the menu item and the editor now agree.

src/BloomBrowserUI/bookEdit/aiImageEditor/aiEditorOverlay.ts
```diff
@@ -112,20 +112,33 @@ export function openAiImageEditor(target: IAiImageEditorTarget): void {
...
-        // Don't preload an empty placeholder slot into the edit target — there's
-        // nothing to edit, and its placeholder graphic isn't a real raster image.
-        const selectedBookImageId = clickedMatch?.isPlaceholder
-            ? undefined
-            : clickedMatch?.id;
+                : [];
+        // The fallback covers the one way the count can overshoot: C# leaves some pictures
+        // out of the book image list, so a page could hold more same-named slots than it sent.
+        const clickedMatch =
+            sameNameOnPage[target.sameNameOrdinal] ?? sameNameOnPage[0];
```
hatton: [Devin] **Bug**: Wrong empty slot selected when a page has several

[Claude Opus 5 from Hatton's machine during preflight]

Every empty picture box shows the same file, `placeHolder.png`, and the overlay found the box the user clicked by page and file name alone. On a page with two empty boxes that always found the first one. A user who chose "Edit with AI..." on the second box was therefore given the first as the target, and the image they made landed in the wrong box.

The commit path already tells same-named boxes apart by their ordinal; only the launch path did not.

hatton: [Claude Opus 5 from Hatton's machine during preflight]

Fixed in bdbb4ba884. The page frame now counts how many same-named boxes come before the one the user clicked, C# carries that count back to the overlay with the file name, and the overlay takes that one out of the same-named boxes on the page.

Counting only the same-named boxes is what makes the count safe: the extra images Bloom injects into the live page, and the pictures C# leaves out of the book image list, carry other file names, so neither can shift it. A count that overshoots falls back to the first same-named box. Four tests cover it, two on each side.

src/BloomBrowserUI/bookEdit/aiImageEditor/aiEditorPageCommands.ts
```diff
@@ -43,11 +43,62 @@ export function launchAiImageEditor(
...
+            imgContainer ?? img,
+            imageFileName,
+        ),
     });
 }
 
+// The classes C# refuses to offer the AI image editor (IsUserChangeableImageElement in
+// AiImageEditorApi.cs). An empty one of those shows placeHolder.png like any other empty slot,
+// so the count below has to skip them or it would run ahead of the list C# sent.
+const kNotUserChangeableClasses = ["branding", "licenseImage", "bloom-qrcode"];
```
hatton: [Devin] **Bug**: Wrong empty slot targeted when a page has an omitted placeholder

[Claude Opus 5 from Hatton's machine during preflight]

The count of same-named boxes ran over every picture on the live page, but the list it indexes into leaves out the branding, license, and QR-code slots. An empty one of those shows `placeHolder.png` like any other empty box, so a page holding one before the box the user clicked put the count one too high. The overlay then fell back to the first empty box, and the image the user made landed there.

hatton: [Claude Opus 5 from Hatton's machine during preflight]

Fixed in 5b116fe125. The count now skips those three classes, mirroring `IsUserChangeableImageElement`, and also skips the controls Bloom injects into the live page, which are in no saved book and so are in no list C# sends. Each side names the other in a comment, so the two lists stay together. Two tests cover it.

hatton:
[Claude Opus 5 from Hatton's machine during preflight]

Consulted Devin on 2026-08-25, up to commit 5b116fe12501a89668b471f7aa4d6aae76414f37.

It raised two bugs, both real, both about aiming at the wrong empty picture box on a page that has more than one. Each has its own thread here, and each is fixed and resolved:

- the second empty box on a page resolving to the first — https://github.com/BloomBooks/BloomDesktop/pull/8230#discussion_r3856038764
- the count running ahead of the list when a branding or licence box also shows the placeholder graphic — https://github.com/BloomBooks/BloomDesktop/pull/8230#discussion_r3856109404

Devin now marks both as fixed. Its one remaining Investigate flag was that Bloom still pinned an older AI image editor; the pin is now at `dist-v0.1.6` and that thread is answered and resolved. Five other flags were Informational and are not posted: three of them read the slot-numbering code and agree with it, one notes that a new string is not yet translated, and one restates the assumption behind the ordinal count, which the code comments already state.

CI (`pr-automation`) passes, and Greptile's one comment was the same pinned-editor point.

src/BloomBrowserUI/bookEdit/aiImageEditor/aiEditorOverlay.ts
```diff
@@ -112,20 +112,33 @@ export function openAiImageEditor(target: IAiImageEditorTarget): void {
...
+        // An empty placeholder slot is sent like any other (BL-16744). It used to be
+        // withheld, on the grounds that an empty slot has nothing to edit — but the AI
+        // image editor answers a missing selectedBookImageId by targeting the FIRST
+        // image of the book, which is normally the front cover. So withholding it aimed
+        // the user at the cover when they had asked for an empty slot on some other page.
+        // The editor reads isPlaceholder on the named slot and, for an empty one, puts
+        // nothing in its "Image to Edit" panel and opens its "Create an Image" tool
+        // instead; it keeps the slot so the created image can be committed straight into
+        // it. That behavior arrived in bloom-ai-image-tools 0.1.6 (package.json pins a
+        // later dist tag); an older pin gets the placeholder graphic as the image to edit.
```
hatton: [Devin] **Bug**: Comment names the wrong pinned editor version

[Claude Fable 5 from Hatton's machine during preflight — mirroring a Devin finding]

The note at the end of this comment said package.json pins the AI image editor as `dist-v0.1.6`, but the pin had already moved to `dist-v0.1.7`. The stated pin was wrong.

hatton: [Claude Fable 5 from Hatton's machine during preflight] Fixed in ad7a92e589: the comment now names only the version where the behavior arrived (0.1.6), not the pinned tag, so a later pin bump cannot make it stale again. Resolving.

hatton:
[Claude Fable 5 from Hatton's machine during preflight]

Consulted Devin on 2026-08-25 up to commit ad7a92e589916b8482d8dd7ccb06dced6fbd6ba6. It raised one bug on this round — a comment naming the pinned editor tag dist-v0.1.6 after the pin moved to dist-v0.1.7 — fixed in ad7a92e589, thread resolved: https://github.com/BloomBooks/BloomDesktop/pull/8230#discussion_r3858106501. Its re-review of that commit marks all three of its bugs resolved; its one standing Investigate flag (the menu enabled before the editor pin moved) was answered earlier by the pin bump. Nine informational items, none needing action. CI (pr-automation) is green.

hatton (commented):
> @hatton partially reviewed 10 files and made 3 comments.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8230)* status: 0 of 16 files reviewed, 7 unresolved discussions.

___
*[`DistFiles/localization/en/BloomMediumPriority.xlf` line 53 at r5](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8230#-P-zfsno0ibLSL3h7Rv2:-P-zfsno0ibLSL3h7Rv3:b-1m69ot) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/ad7a92e589916b8482d8dd7ccb06dced6fbd6ba6/DistFiles/localization/en/BloomMediumPriority.xlf#L53)):*
> ```XML
>         <note>Context-menu item on an image in the Edit tab; opens the AI image editor in an overlay. The "..." indicates further UI will open, as in the sibling item "Choose image from your computer...".</note>
>       </trans-unit>
>       <trans-unit id="AiImageEditor.SlotLabel.Page" translate="no">
> ```

Do we already have "Page" and "Image"? then just reuse those and append the number. Don't worry that some language might prefer to change the order or something. 
___
*[`DistFiles/localization/en/BloomMediumPriority.xlf` line 68 at r5](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8230#-P-zfkxL8DO1ULuCfqzv:-P-zfkxL8DO1ULuCfqzw:b-esidi) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/ad7a92e589916b8482d8dd7ccb06dced6fbd6ba6/DistFiles/localization/en/BloomMediumPriority.xlf#L68)):*
> ```XML
>         <note>Names one picture of a page that has several, shown over the picture in the AI image editor's strip of the book's images. {0} is replaced with the number of the picture on its page, counting from 1 and not counting the background picture. Combined with the page name by AiImageEditor.SlotLabel.PageAndSlot, giving e.g. "Page 3 - Image 2".</note>
>       </trans-unit>
>       <trans-unit id="AiImageEditor.SlotLabel.PageAndSlot" translate="no">
> ```

Let's not have this as a localizable thing. It's too confusing for the translator and not needed.
___
*[`src/BloomBrowserUI/bookEdit/aiImageEditor/aiEditorPageCommands.ts` line 60 at r5](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8230#-P-zgUxS8eNh4__pLJ7w:-P-zgUxS8eNh4__pLJ7x:b7kbsl3) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/ad7a92e589916b8482d8dd7ccb06dced6fbd6ba6/src/BloomBrowserUI/bookEdit/aiImageEditor/aiEditorPageCommands.ts#L60)):*
> ```XML
> // AiImageEditorApi.cs). An empty one of those shows placeHolder.png like any other empty slot,
> // so the count below has to skip them or it would run ahead of the list C# sent.
> const kNotUserChangeableClasses = ["branding", "licenseImage", "bloom-qrcode"];
> ```

Is this the best place to keep this list? If so, okay.
Next question on this: None of these show up in Bloom image containers, right? So if that's true, why do we need to even have this list? Could we instead limit the list of images to those that are inside of Bloom image containers?

hatton (commented):
> @hatton+JHAI made 3 comments and resolved 3 discussions.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8230)* status: 0 of 16 files reviewed, 4 unresolved discussions.

___
*[`DistFiles/localization/en/BloomMediumPriority.xlf` line 53 at r5](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8230#-P-zfsno0ibLSL3h7Rv2:-P0-J5SD86TTzqCWVLrn:bmxyc93) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/ad7a92e589916b8482d8dd7ccb06dced6fbd6ba6/DistFiles/localization/en/BloomMediumPriority.xlf#L53)):*

Do we already have "Page" and "Image"? then just reuse those and append the number. Don't worry that some language might prefer to change the order or something.
</blockquote></details>

[Claude Opus 5 from Hatton's machine during reviewable-replies]

Yes, both existed. It now reuses `ReaderSetup.PageHeader` ("Page") and `EditTab.CustomPage.Image` ("Image") and appends the number. The three new trans-units are deleted; only `AiImageEditor.SlotLabel.CanvasBackground` remains.
___
*[`DistFiles/localization/en/BloomMediumPriority.xlf` line 68 at r5](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8230#-P-zfkxL8DO1ULuCfqzv:-P0-J57aDAEH2CyPzWoW:b-grgymk) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/ad7a92e589916b8482d8dd7ccb06dced6fbd6ba6/DistFiles/localization/en/BloomMediumPriority.xlf#L68)):*

Let's not have this as a localizable thing. It's too confusing for the translator and not needed.
</blockquote></details>

[Claude Opus 5 from Hatton's machine during reviewable-replies]

Done. `AiImageEditor.SlotLabel.PageAndSlot` is gone. The C# now joins with a literal `" - "`, so no translator ever sees a format string.
___
*[`src/BloomBrowserUI/bookEdit/aiImageEditor/aiEditorPageCommands.ts` line 60 at r5](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8230#-P-zgUxS8eNh4__pLJ7w:-P0-J5mI0zZNXLfQ1nv_:b7lhlyd) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/ad7a92e589916b8482d8dd7ccb06dced6fbd6ba6/src/BloomBrowserUI/bookEdit/aiImageEditor/aiEditorPageCommands.ts#L60)):*

Is this the best place to keep this list? If so, okay.
Next question on this: None of these show up in Bloom image containers, right? So if that's true, why do we need to even have this list? Could we instead limit the list of images to those that are inside of Bloom image containers?
</blockquote></details>

[Claude Opus 5 from Hatton's machine during reviewable-replies]

You are right, and the list is gone. Your second question led to a bigger simplification of the whole slot-numbering design, which John approved for this card:

- A slot is now simply a `bloom-imageContainer` on the page. The ordered list of those containers *is* the numbering, and the index is the whole identity.
- Branding, license and QR-code images sit outside an image container, so they are excluded by construction. No class list, no filename, no fallback.
- `aiEditorSlotMatching.ts` and its test are deleted, `sameNameOrdinalOnPage` is deleted, and `IAiImageEditorTarget` is now `{ pageId, slotIndex }`.
- C# numbers the slots the same way (`SelectImageSlotsOnPage`), so the two ends agree without sharing anything but the index.

The only exclusion left on the live-page side is `.bloom-ui`, because Bloom injects temporary controls that the saved book does not have.

src/BloomBrowserUI/bookEdit/aiImageEditor/aiEditorOverlay.ts
hatton: [Claude Opus 5 from Hatton's machine during reviewable-replies]

Still fixed, but by simpler means than this reply describes. Review feedback asked why the design used a file name and an ordinal at all. It no longer does: a slot is now the *n*th `bloom-imageContainer` on the page, and that index is the whole identity, on both sides. `aiEditorSlotMatching.ts` and `sameNameOrdinalOnPage` are deleted.

src/BloomBrowserUI/bookEdit/aiImageEditor/aiEditorPageCommands.ts
hatton: [Claude Opus 5 from Hatton's machine during reviewable-replies]

Still fixed, and the class list this reply mentions is gone. Slots are now taken from the ordered list of `bloom-imageContainer` elements on the page. Branding, license and QR-code images are not inside an image container, so they are excluded by construction rather than by name. The only remaining exclusion on the live-page side is `.bloom-ui`, for the controls Bloom injects that the saved book does not have.

hatton (commented):
> @hatton+JHAI made 2 comments and resolved 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8230)* status: 0 of 16 files reviewed, 2 unresolved discussions.

src/BloomBrowserUI/bookEdit/aiImageEditor/aiImageEditorPageCommands.ts
hatton, line 1: [Devin] **Investigate**: Live-page slot filter excludes only .bloom-ui, but save strips more

[Claude Opus 5 from Hatton's machine during preflight — mirroring a Devin finding]

A picture box is now identified by its position among the page's image containers, and the live page and the saved book have to number the same boxes. The live page leaves out only the controls that carry the `bloom-ui` class, but a save also removes anything marked `editTimeOnly`, a `dragHandle`, or a jQuery resize handle. If Bloom ever injected a picture box inside one of those, the live count would be one too high and the AI editor would fill the wrong box.

hatton, line 1: [Devin] **Investigate**: Post-commit save removed; current-page swap relies on the normal page save

[Claude Opus 5 from Hatton's machine during preflight — mirroring a Devin finding]

Bloom no longer saves the page as soon as the AI image editor closes, because that save threw away the undo. The new picture is put into the page the same way Paste Image puts one there, and it is written to disk by whatever saves the page next. Devin asks whether closing Bloom straight after accepting an image, without turning to another page first, still keeps the picture.

hatton: [Claude Opus 5 from Hatton's machine during preflight]

Not an issue, and the invariant holds. `Cleanup` removes three other things besides `bloom-ui`, and none of them can hold a picture box:

- `editTimeOnly` has exactly one producer in the whole front end, `BloomSourceBubbles`, which adds an empty `<ul>` for the source-language tabs. It also carries `bloom-ui`, so it is excluded twice over.
- `dragHandle` and `ui-resizable-handle` are the small grab handles jQuery UI adds. They are empty by construction.

Bloom has no code that puts a picture box inside a control it later strips, so the two counts cannot drift. Adding those three names to the filter would put back the kind of class list this PR just removed, for a case that cannot happen.

hatton: [Claude Opus 5 from Hatton's machine during preflight]

Not an issue. Since this PR the new picture is an ordinary edit to the page, made by the same call Paste Image uses, so it is saved by exactly the same path as a pasted picture. If quitting Bloom right after accepting an image lost the picture, quitting right after pasting one would lose it too, and that would be a long-standing bug in Bloom's page saving rather than anything this branch introduced.

The save that was removed is the one that ran immediately after the editor closed. It was removed on purpose, because it discarded the undo record and so made an accepted image impossible to undo, which is what BL-16744 asks for.

hatton:
[Claude Opus 5 from Hatton's machine during preflight]

Consulted Devin on 2026-08-26, up to commit `05bc788dc1f261ec29d2cd96c0bc99b415910b97`.

It raised two things worth a thread on this round, both about the change that now identifies a picture box by its position on the page. Each has its own thread here, and I judged both to be non-issues and said why:

- whether a picture box could hide inside a control that a save strips — https://github.com/BloomBooks/BloomDesktop/pull/8230#discussion_r3867625668
- whether accepting an image and quitting Bloom straight away still keeps the picture — https://github.com/BloomBooks/BloomDesktop/pull/8230#discussion_r3867625743

Devin now marks all three of its earlier bugs as fixed, and their threads were already resolved. Its one remaining Investigate flag, that the menu was enabled before the editor pin moved, is answered by the pin now sitting at `dist-v0.1.7`; that thread is resolved too. Eight informational items needed no action.

CI (`pr-automation`) passes. The whole C# suite (3269 tests) and the whole front-end suite (781 tests) are green at this commit.

hatton (commented):
> @hatton reviewed 15 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8230)* status: 0 of 17 files reviewed, 4 unresolved discussions.

andrew-polk:
@copilot+gpt-5.4 resolve the merge conflicts in this pull request

andrew-polk (commented):
> @andrew-polk partially reviewed 24 files and all commit messages, made 1 comment, and resolved 4 discussions.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8230)* status: 23 of 24 files reviewed, 1 unresolved discussion (waiting on hatton).

___
*[`src/BloomBrowserUI/bookEdit/toolbox/canvas/canvasControlTypes.ts` line 122 at r8](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8230#-P08Y6Bn1LntIaqqEoYY:-P08Y6Bn1LntIaqqEoYZ:b-j69117) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/2537a53f7fca646d63d9f7d639ce46e7e6dcfd18/src/BloomBrowserUI/bookEdit/toolbox/canvas/canvasControlTypes.ts#L122)):*
> ```XML
>     // Distinct from !hasRealImage, which is also true for an image that failed to
>     // load: an empty slot is a normal state a user can fill, a broken image is not.
>     isPlaceholderImage: boolean;
> ```

I think this would be more correct/clear as hasPlaceholderImage.

I won't force it, but I did get confused wondering if there was a semantic meaning difference between **has**Image and **has**RealImage vs **is**PlaceholderImage.

Of course, you could then make the argument that isBackgroundImage should be hasBackgroundImage...

hatton:
> > @andrew-polk partially reviewed 24 files and all commit messages, made 1 comment, and resolved 4 discussions.
> > _[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8230)_ status: 23 of 24 files reviewed, 1 unresolved discussion (waiting on hatton).
> 
> _[`src/BloomBrowserUI/bookEdit/toolbox/canvas/canvasControlTypes.ts` line 122 at r8](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8230#-P08Y6Bn1LntIaqqEoYY:-P08Y6Bn1LntIaqqEoYZ:b-j69117) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/2537a53f7fca646d63d9f7d639ce46e7e6dcfd18/src/BloomBrowserUI/bookEdit/toolbox/canvas/canvasControlTypes.ts#L122)):_
> 
> > ```
> >     // Distinct from !hasRealImage, which is also true for an image that failed to
> >     // load: an empty slot is a normal state a user can fill, a broken image is not.
> >     isPlaceholderImage: boolean;
> > ```
> 
> I think this would be more correct/clear as hasPlaceholderImage.
> 
> I won't force it, but I did get confused wondering if there was a semantic meaning difference between **has**Image and **has**RealImage vs **is**PlaceholderImage.
> 
> Of course, you could then make the argument that isBackgroundImage should be hasBackgroundImage...

@copilot fix this

hatton:
> > > > @andrew-polk partially reviewed 24 files and all commit messages, made 1 comment, and resolved 4 discussions.
> > > > _[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8230)_ status...
> 
> Fixed in `256bbf6`. I renamed the flag to `hasPlaceholderImage` and updated its uses and tests for consistency with the other `has*Image` context fields.

@copilot why don't I see that commit in the list of commits for this pr?

hatton (commented):
> @hatton made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8230)* status: 19 of 24 files reviewed, 1 unresolved discussion (waiting on andrew-polk).

___
*[`src/BloomBrowserUI/bookEdit/toolbox/canvas/canvasControlTypes.ts` line 122 at r8](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8230#-P08Y6Bn1LntIaqqEoYY:-P0Nfn-TDqtGOImz_7MD:b-t19ajf) ([raw file](https://github.com/BloomBooks/BloomDesktop/blob/2537a53f7fca646d63d9f7d639ce46e7e6dcfd18/src/BloomBrowserUI/bookEdit/toolbox/canvas/canvasControlTypes.ts#L122)):*

@copilot fix this
</blockquote></details>

So once I saw the fix then I thought, yeah no, that in other contexts just seems to make things more complex. Since you weren't feeling super strong about it, I decided to undo the change.

hatton (commented):
> @hatton resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8230)* status: 2 of 24 files reviewed, all discussions resolved (waiting on andrew-polk).

andrew-polk:
Based on your comment, I was assuming the code is the same as last time I looked which would allow me to merge.
But things are definitely in a strange state. Reviewable is super confused with its diff. And there are now merge conflicts I thought I had resolved already.

In addition, this should target Version6.5. I wonder if that is part of the issue?

Anyway, I don't think any further chagnes were needed from my review.

PR #8231, opened 2026-08-25
BL-16690 bring the minimum-version startup dialogs to the front

andrew-polk:
**Problem.** QA on BL-16690's test channel found that when Bloom is launched by double-clicking a `.bloomCollection` file and the minimum-version gate fires, the windows that follow the gate dialog open **behind Windows Explorer** with no taskbar button: both the "Upgrade Bloom" progress dialog and the collection chooser ("Open a Different Collection"). Bloom looks hung, and there is nothing on the taskbar to find it by.

**Cause.** Those windows are shown with **no owner**: by the time they open, the gate message box that held the foreground has closed, and Windows hands the foreground back to Explorer — sometimes asynchronously, *after* the new window has already appeared. `ReactDialog` also suppresses taskbar buttons, which is right for owned dialogs but leaves an ownerless one unfindable.

**Fix.**
- A shared `Form` extension, `BringToFrontWhenShown`: on `Shown`, go topmost and `Activate()`, then drop `TopMost` 1.5 s later — the delayed drop re-asserts the window at the top of the ordinary z-band, so it wins the race against Explorer's asynchronous re-raise whichever order the raises land in. (Same idea as `Shell.ReallyComeToFront`, which solves this for the main window after the splash screen.)
- Applied to the Upgrade Bloom progress dialog and to the collection chooser when it opens ownerless at startup. The owned, mid-session paths are untouched.
- The progress dialog and any ownerless `BloomMessageBox` (the gate dialog itself) now get their own taskbar buttons, so a startup dialog that does end up behind something can always be found.

[Devin review](https://devinreview.com/BloomBooks/BloomDesktop/pull/8231)

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8231)

### DIFF
--- a/src/BloomExe/Collection/MinimumBloomVersionCheck.cs
+++ b/src/BloomExe/Collection/MinimumBloomVersionCheck.cs
@@ -625,7 +625,8 @@ namespace Bloom.Collection
             BrowserProgressDialog.DoWorkWithProgressDialog(
                 socketServer,
                 () =>
-                    new ReactDialog(
+                {
+                    var dlg = new ReactDialog(
                         "progressDialogBundle",
                         new
                         {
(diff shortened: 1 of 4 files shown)


andrew-polk:
﻿[Claude Fable 5] Consulted Devin (devinreview.com) on this PR up to `6eeceb0bd1e17d6f2c3d48cc21c338c6d09b2f74`. Result: **no bugs found**; 1 informational note (the 1.5s bring-to-front timer disposes itself only from its own Tick, so if the message pump stopped first it would be a one-shot leak — these dialogs run under the app's normal pump, so no action taken). Greptile and CI were also clean.

StephenMcConnel (approved):
> @StephenMcConnel reviewed 4 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8231)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on andrew-polk).

StephenMcConnel (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8231)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on andrew-polk).

PR #8234, opened 2026-08-25
Paste on a standard-layout xmatter page replaces its background image (BL-16542)

StephenMcConnel:
On a **Standard Layout** front cover, Ctrl+V and the top-bar **Paste** button added the clipboard
image as a *new canvas element on top of* the cover picture. Such a page cannot hold canvas
elements at all, so a paste there can only sensibly mean "replace this page's picture".

`finishPasteImageFromClipboard()` gets a third rule, after "a selected image is replaced" and
before "add a new canvas element": if the page cannot hold canvas elements — a standard-layout
xmatter page — replace its background image, even when that image is real rather than a
placeholder and nothing is selected, and never add an element. It goes through the existing
`replaceImageInCanvasElement()` helper, so the copyright/metadata button is rebuilt as it is
elsewhere (BL-16605). It is deliberately not gated on the canvas subscription feature: replacing a
cover picture is not the overlay feature.

Interior pages and custom-layout covers are unchanged — they still add an overlay, which is why
this card's earlier "nothing selected replaces the picture" rule had to be reversed for them. The
missing distinction was the page, not the selection.

The "does this page allow canvas elements?" test (an xmatter page allows them only in custom
layout) was written inline in the `becomeBackground` availability rule; it is now a shared
`pageAllowsCanvasElements()` in `canvasElementDomUtils.ts` that both callers use.

### Verified in the running app (not just unit tests)

On a Standard Layout front cover of a real book, with a real cover picture and nothing selected:

- toolbar **Paste** → cover picture replaced, canvas-element count stayed at 1, metadata button
  rebuilt;
- **Ctrl+V** → same;
- an **interior page** with a real background image and nothing selected → still adds an overlay
  (1 → 2 canvas elements, background untouched).

Ref: https://issues.bloomlibrary.org/youtrack/issue/BL-16542

[Devin review](https://devinreview.com/BloomBooks/BloomDesktop/pull/8234)

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8234)

### DIFF
--- a/src/BloomBrowserUI/bookEdit/js/canvasElementManager/CanvasElementClipboard.test.ts
+++ b/src/BloomBrowserUI/bookEdit/js/canvasElementManager/CanvasElementClipboard.test.ts
@@ -17,6 +17,11 @@ vi.mock("../bloomImages", () => ({
     isPlaceHolderImage: (src: string | null) =>
         !!src && src.toLowerCase().includes("placeholder.png"),
     SetupMetadataButton: vi.fn(),
+    // Mirrors the real helper in bloomImages.ts: the first bloom-backgroundImage in the canvas.
+    getBackgroundCanvasElementFromBloomCanvas: (bloomCanvas: HTMLElement) =>
+        bloomCanvas.getElementsByClassName(
+            "bloom-backgroundImage",
+        )[0] as HTMLElement,
 }));
 
 vi.mock("../bloomEditing", () => ({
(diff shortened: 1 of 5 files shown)


StephenMcConnel:
[Claude Opus 5 (1M context)] Consulted Devin on 2026-08-25 14:12 (local) up to commit `0fab6387b2832b262f9b52a92c886e0a12174dd6`.

Clean: no bugs and nothing flagged for investigation. It did raise two informational observations, neither of which asks for a change:

- it checked the extracted `pageAllowsCanvasElements()` helper against the inline test it replaced in the "Become Background" rule and agrees they behave identically, including for a missing page;
- it noted that on a page which cannot hold a floating picture, if the picture frame had no picture area at all, the paste would quietly do nothing. Standard front/back-matter pages always have one, so this is unreachable in practice; it is called out in the preflight report for the developer to decide on.

Greptile also finished on this commit with a summary and no findings, and CI is green.

StephenMcConnel:
[Claude Opus 5 (1M context)] Developer decisions on the two open questions from the preflight report — both **leave as is**, so there is no code change:

- **No message when a paste replaces a Standard Layout cover picture.** A tester had suggested Bloom explain that such a cover cannot hold an extra picture. Decided against for now: the cover picture visibly changes, so the result is its own feedback, and a notice on a routine action would wear thin. If the team wants the wording later it is a small, self-contained addition.
- **The "nothing to replace" edge case stays a quiet no-op** (Devin's second informational note above). If a Standard Layout front/back-matter page ever had a picture frame with no picture area in it, the paste ends without doing anything rather than adding a picture the page cannot hold. Bloom builds that picture area for every picture frame as the page loads, so this is not reachable in practice, and guarding it would cost a translatable string for a case we do not believe exists.

Also decided: the Custom Layout cover path stays covered by its unit test rather than a run-through in the app, since setting that up means converting a real book's cover.

andrew-polk (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomDesktop/8234)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on StephenMcConnel).

